PR #9407, opened 2026-02-21
[Fair Sharing] If workload fits in nominalQuota, ignore DRS checks during admission and preemption

### LINKED ISSUE #9406: Fair sharing admission ordering does not preserve nominal-first workload priority
**What happened**:

|---|---|---|---|

**Current state:**

**Environment**:

mukund-wayve:
#### What type of PR is this?

/kind bug

In a multi-flavor setup, with fair sharing enabled, we noticed that borrowing resources on one flavor affects a CQ's ability to use its nominal quota on a different flavor -- in both admission ordering and preemption.

I've described the issue in more detail in #9406 . This PR tries to handle it the same way that classical preemption does. Ensures that workloads fitting within their CQ's nominal quota are never penalized by borrowing on other flavors.

**Changes:**

- Admission ordering (fair_sharing_iterator.go): Prefer workloads that don't require borrowing before comparing DRS, so nominal-entitled workloads are scheduled first.
- Preemption (preemption.go): Add `preemptorWithinNominalForContestedResources` check before the preemption simulation. When the preemptor CQ (including the incoming workload) stays within nominal quota for the contested flavor-resources, skip the DRS strategy gate in runFirstFsStrategy and preempt unconditionally.
- Refactor: Extract preemptCandidate helper to deduplicate the repeated pop/remove/append/check-fits sequence across four different places.

Fixes #9406

This definitely changes the behaviour. If this change is accepted,
- Should we keep this under a feature gate?
- Should we declare this as a breaking change?

```release-note
[Fair Sharing] When a workload fits within its ClusterQueue's nominal quota, prioritize it in admission ordering and allow preemption regardless of the CQ's DRS. This prevents heavy borrowing on one flavor from blocking a CQ's nominal entitlement on another flavor.
```

### DIFF
--- a/pkg/scheduler/preemption/preemption_fair_test.go
+++ b/pkg/scheduler/preemption/preemption_fair_test.go
@@ -916,6 +916,10 @@ func TestFairPreemptions(t *testing.T) {
 					Preemption(kueue.ClusterQueuePreemption{
 						ReclaimWithinCohort: kueue.PreemptionPolicyAny,
 					}).
+					FlavorFungibility(kueue.FlavorFungibility{
+						WhenCanBorrow:  kueue.MayStopSearch,
+						WhenCanPreempt: kueue.MayStopSearch,
+					}).
 					Obj(),
 				utiltestingapi.MakeClusterQueue("b").
 					Cohort("all").


pkg/scheduler/preemption/preemption_fair_test.go
```diff
@@ -95,13 +95,15 @@ func TestFairPreemptions(t *testing.T) {
...
-		cohorts       []*kueue.Cohort
-		strategies    []config.PreemptionStrategy
-		admitted      []kueue.Workload
-		incoming      *kueue.Workload
-		targetCQ      kueue.ClusterQueueReference
-		wantPreempted sets.Set[string]
+		clusterQueues   []*kueue.ClusterQueue
+		cohorts         []*kueue.Cohort
+		flavors         []*kueue.ResourceFlavor
+		assignmentFlavor kueue.ResourceFlavorReference
```
mukund-wayve: Required to support multiple flavors in tests

pkg/scheduler/preemption/preemption.go
```diff
@@ -311,24 +311,44 @@ func parseStrategies(fs *config.FairSharing) []fairsharing.Strategy {
...
-			if workloadFitsForFairSharing(preemptionCtx) {
+			targets, fits = preemptCandidate(preemptionCtx, candCQ.PopWorkload(), candCQ, kueue.InClusterQueueReason, targets)
+			if fits {
+				return true, targets, nil
+			}
+			continue
+		}
+
+		if preemptorUnderNominal {
+			targets, fits = preemptCandidate(preemptionCtx, candCQ.PopWorkload(), candCQ, kueue.InCohortFairSharingReason, targets)
```
pajakd: I would use a different Reason -- perhaps InCohortReclamationReason?

pajakd:
In general I agree that its a bug. Fair sharing is about access to shared resources -- access to the nominal quota should not be blocked by borrowing in some other flavor.

This PR looks good but will work for using nominal quota of CQ. But in more complex hierarchies, there could be a nominal quota of a cohort for which the descendants of that cohort should have a preferential access, right?

CQ1 -> Cohort1 -> Root
CQ2 -> Cohort2 -> Root
In the above example if CQ1 and CQ2 were competing for nominal quota in Cohort1, then CQ1 should be preferred, regardless if it is borrowing more in some other flavor.

pkg/scheduler/preemption/preemption.go
```diff
@@ -400,10 +408,19 @@ func (p *Preemptor) fairPreemptions(preemptionCtx *preemptionCtx, strategies []f
...
 	}
 
+	// Check before simulation: would the preemptor CQ, including the
+	// incoming workload, still be within nominal quota for the contested
+	// resources?  If so, preemption should be allowed regardless of DRS
+	// (the workload is entitled to its nominal quota).
+	// This only affects runFirstFsStrategy: when true, all candidates are
+	// preempted unconditionally (bypassing the strategy check), so no
+	// retryCandidates are produced and runSecondFsStrategy has nothing to do.
+	preemptorUnderNominal := preemptorWithinNominalForContestedResources(preemptionCtx)
```
pajakd: I don't think we need the new function preemptorWithinNominalForContestedResources. In the line below we add the usage of the workload to the snapshot. So you can just use queueUnderNominalInResourcesNeedingPreemption (but in just call it below "SimulateUsageAddition").

pkg/scheduler/preemption/preemption.go
pajakd, line 426: This is not relevant for the second strategy?

mukund-wayve: The workloads are already considered in `runFirstFsStrategy`. With your suggestion to move the `preemptorWithinNominal` after simulation, I've simplified the code and moved the check to be only inside `runFirstFsStrategy` rather than the caller which is much cleaner. Thanks for the suggestion

pkg/scheduler/preemption/preemption_fair_test.go
```diff
@@ -118,7 +120,7 @@ func TestFairPreemptions(t *testing.T) {
 			},
 			incoming:      unitWl.Clone().Name("c_incoming").Obj(),
 			targetCQ:      "c",
-			wantPreempted: sets.New(targetKeyReason("/b1", kueue.InCohortFairSharingReason)),
+			wantPreempted: sets.New(targetKeyReason("/b1", kueue.InCohortReclamationReason)),
```
mukund-wayve, line 123: All of these test reason changes are because we are reclaiming nominal quota and the new code path is activated. This is a more correct reason compared to the existing behaviour.

mukund-wayve: Several of these test cases already mention that the preemptor is within nominal quota

pkg/scheduler/preemption/preemption.go
mukund-wayve: Done. This has triggered a few changes in the tests. But I agree this is the correct reason and we provide a more correct reason for those cases

pkg/scheduler/preemption/preemption.go
mukund-wayve: Thanks for the suggestion. I agree it's better to move after `SimulateUsageAddition` because it also increases usage corresponding in the parents.

But we can't use `queueUnderNominalInResourcesNeedingPreemption` because it doesn't handle the case where adding the new workload fits exactly at nominal quota. I have added a new function `queueWithinNominalInResourcesNeedingPreemption` to handle this case. Let me know what you think.

Moving after simulation has allowed making this call only in `runFirstFsStrategy` which is cleaner.

mukund-wayve (commented):
Thanks for the review @pajakd!

I've addressed your suggestions:
- Changed the reason to InCohortReclamationReason for the nominal-first preemption path.
- Moved the nominal check after SimulateUsageAddition and created queueWithinNominalInResourcesNeedingPreemption (uses <= instead of <) to be used in runFirstFsStrategy. This also sets up a cleaner path for the cohort extension — after simulation, the snapshot already has correct usage at all hierarchy levels.


On the hierarchical case: great spot! This PR only handles the CQ-level nominal case. But what we have here is already an improvement and covers what I'd assume is the most common case. I'll raise a separate issue to handle the cohort-level case and work on it as a follow-up PR. Would that be okay?

pajakd:
Yes, I think we can handle the general, hierarchical case in a separate PR -- please open an issue for that. Its not clear yet how we'll handle it. This PR is already a big step in the right direction, thanks!

/lgtm

mukund-wayve:
Thanks for the review @pajakd ! I've created #9466 as a follow-up for the hierarchical cohort case. What are the next steps to get this PR merged? This is my first contribution to kueue and I'd appreciate any guidance

mimowo:
> I've described the issue in more detail in https://github.com/kubernetes-sigs/kueue/issues/9406 . This PR tries to handle it the same way that classical preemption does. Ensures that workloads fitting within their CQ's nominal quota are never penalized by borrowing on other flavors.

This sounds like something we should already support in FlavorFungability with the `preference: PreemptionOverBorrowing` (but there is currently a bug we are fixing: https://github.com/kubernetes-sigs/kueue/issues/9462. Could you please check that feature first and see what is missing there?

mimowo:
/hold
To first determine what is missing in flavorFungability.preference=PreemptionOverBorrowing, because this sounds to me like exactly the issue we were covering.

mukund-wayve:
Hi @mimowo, thanks for your response. I'm still learning the codebase, so appreciate the patience.

TLDR:

FlavorFungibility.Preference (PreemptionOverBorrowing) is only allowed when both WhenCanBorrow and WhenCanPreempt are set to TryNextFlavor (enforced by CEL validation on the API). We use MayStopSearch, so Preference is not applicable to our setup. I've updated the test to [explicitly set MayStopSearch](https://github.com/kubernetes-sigs/kueue/pull/9407/changes#diff-aee60c971eca1e23b63bed32e1e54b3c67c4fee9f35f1ca54e5a2bb879c7683bR919) to reflect this. 

#9462 fixes the flavor assignment step, this PR fixes the fair sharing preemption step. They are independent — this PR is needed with or without #9462, whenever fair sharing is enabled.

---

They address different code paths:
When fair sharing is disabled, flavor assignment picks the right flavor (via MayStopSearch or PreemptionOverBorrowing), sets Mode=Preempt, and getTargets() calls classicalPreemptions() (preemption.go L143), which has its own nominal-aware logic and allows the preemption. #9462 is about a bug in this flavor assignment step.

When fair sharing is enabled, the flavor assignment step works the same way — it still correctly picks the flavor and sets Mode=Preempt. But getTargets() calls fairPreemptions() instead (preemption.go L141). The DRS strategy gates can then block the preemption because the preemptor's DRS is inflated by borrowing on a different flavor, even though it's within nominal quota on the contested flavor. If no targets pass the strategy gate, GetTargets() returns nil and the workload is not admitted (scheduler.go L267-281).

This PR fixes that: when the preemptor is within nominal for the contested resources, bypass the DRS strategy gate so the preemption proceeds. Without this fix, flavor assignment (whether via MayStopSearch or PreemptionOverBorrowing) picks the right flavor, but fair sharing blocks the preemption it chose.

mimowo:
> FlavorFungibility.Preference (PreemptionOverBorrowing) is only allowed when both WhenCanBorrow and WhenCanPreempt are set to TryNextFlavor (enforced by CEL validation on the API). We use MayStopSearch, so Preference is not applicable to our setup. I've updated the test to [explicitly set MayStopSearch](https://github.com/kubernetes-sigs/kueue/pull/9407/changes#diff-aee60c971eca1e23b63bed32e1e54b3c67c4fee9f35f1ca54e5a2bb879c7683bR919) to reflect this.

Yes, but the scope of Preference is restricted to `TryNextFlavor` only for the simplicity for now. If this would help your use case we could easily relax the validation. However, for what I read below it would not help anyway, because even currently the correct flavor is chosen with the "Preempt" mode:

> This PR fixes that: when the preemptor is within nominal for the contested resources, bypass the DRS strategy gate so the preemption proceeds. Without this fix, flavor assignment (whether via MayStopSearch or PreemptionOverBorrowing) picks the right flavor, but fair sharing blocks the preemption it chose.

Interesting, so in this case the correct flavor and mode is selected. The preemption is possible, but it does not happen in practice, right?

mukund-wayve:
> so in this case the correct flavor and mode is selected. The preemption is possible, but it does not happen in practice, right?

That's exactly right. The correct flavor and mode are selected, but preemption is blocked by the fair sharing DRS strategy gate. The workload ends up [hitting this path](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/scheduler/scheduler.go#L268): `"Workload requires preemption, but there are no candidate workloads allowed for preemption"` and is skipped. This PR solves for that.

This PR also handles [admission ordering](https://github.com/kubernetes-sigs/kueue/pull/9407/files#diff-8716acbdd5bd256fcf6e05caf9b5573b819e7480aca9b6154dc91c8942ae73efR173). The classical iterator already [prefers workloads that don't require borrowing](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/scheduler/scheduler.go#L733-L738), but the fair sharing iterator only uses DRS. This means a workload that fits within its CQ's nominal quota on the contested flavor can be deprioritized solely because the CQ's aggregate DRS is high from borrowing on other flavors. The PR adds the same nominal-first ordering to the fair sharing iterator

mimowo:
@mukund-wayve thanks a lot for helping me understand the PR, this looks quite reasonable. I now fully undestand this is not related to the flavor preference:
/unhold

Let me summarize my understanding of the changes:

> Admission ordering (fair_sharing_iterator.go): Prefer workloads that don't require borrowing before comparing DRS, so nominal-entitled workloads are scheduled first.

Simple bug, we should first schedule workloads which don't require borrowing. 
Also, alignment with the classical preemption.
I think no gate needed.

> Preemption (preemption.go): Add preemptorWithinNominalForContestedResources check before the preemption simulation. When the preemptor CQ (including the incoming workload) stays within nominal quota for the contested flavor-resources, skip the DRS strategy gate in runFirstFsStrategy and preempt unconditionally.

I'm not yet sure - is this related to the "Workload requires preemption, but there are no candidate workloads allowed for preemption"? Reading the code it seems like just a short circuit rather than changing the outcomes can preempt / cannot. Let's clarify here. 


> Refactor: Extract preemptCandidate helper to deduplicate the repeated pop/remove/append/check-fits sequence across four different places.

Nice refactor. 


Would it make sense to split the PR into 3 smaller to make it easier to see which change is relevant where?

The smaller changes would help us make a decision about the feature gates. I **think** what you report are bugs, but fair sharing is already with us for a while, so maybe two dedicated FGs (enabled by default) makes sense as a bailout option

mukund-wayve:
Thanks for the detailed comment @mimowo  🙏  That's a fair ask.

I've created the [first PR here](https://github.com/kubernetes-sigs/kueue/pull/9484) which is about admission ordering. No gates here. 

I'll create another one for preemption later today. Preemption code changes deviated from the PR description as I was addressing the review comments. Hence why it's confusing. Now that I know what a good implementation looks like, I'll explain it in the new PR

PR #9251, opened 2026-02-15
KEP-8902: Flavor-Aware Dominant Resource Share (DRS)

mukund-wayve:
#### What type of PR is this?

/kind feature

/kind api-change

#### What this PR does / why we need it:

Creates a KEP for https://github.com/kubernetes-sigs/kueue/issues/8902

#### Which issue(s) this PR fixes:

Fixes #

#### Special notes for your reviewer:

This is my first time writing a KEP. I've tried to follow the template and read a few different KEPs to be consistent.

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #9427, opened 2026-02-23
Narrow TAS ResourceFlavor spec immutability to topology-sensitive fields

mukund-wayve:
#### What type of PR is this?

/kind cleanup
/kind api-change

When reviewing [KEP-8902](https://github.com/kubernetes-sigs/kueue/pull/9251/changes#diff-af475a18949bdc489d3c2e61e817a3507f1c71f27cb65eb5daad2ff363782c71R257), @sohankunkerkar made a really good point that enabling TAS currently makes the entire ResourceFlavorSpec immutable. This is handled by a CEL validation rule: `!has(oldSelf.topologyName) || self == oldSelf` and was introduced as a conservative measure to protect the TAS topology tree. But it over-approximates: any new field added to the spec (e.g. `fairSharing.resourceWeights`) also becomes immutable, forcing users to delete and recreate the ResourceFlavor to update it.

This PR narrows the immutability to only the fields whose mutation would actually invalidate the TAS topology tree: `nodeLabels`, `tolerations`, and `topologyName`. Each gets its own CEL rule with `has()` guards for absent optional fields and a field-specific error message.

`nodeTaints` is excluded because it is not stored in the TAS cache and is not used to build the topology tree. It is only consumed by the flavor assigner at admission time. Please correct me if I'm wrong.

Fixes #

```release-note
TAS: Relax ResourceFlavor spec immutability to only topology-sensitive fields (nodeLabels, tolerations, topologyName), allowing nodeTaints and future spec fields to be updated in-place.
```

### DIFF
--- a/apis/kueue/v1beta1/resourceflavor_types.go
+++ b/apis/kueue/v1beta1/resourceflavor_types.go
@@ -44,8 +44,6 @@ type TopologyReference string
 
 // ResourceFlavorSpec defines the desired state of the ResourceFlavor
 // +kubebuilder:validation:XValidation:rule="!has(self.topologyName) || self.nodeLabels.size() >= 1", message="at least one nodeLabel is required when topology is set"
-// nodeLabels, tolerations, and topologyName are immutable when topologyName
-// is set because their mutation could invalidate the TAS topology tree.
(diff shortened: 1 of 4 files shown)


mimowo:
This is quite reasonable, thank you 👍 
I think making tolerations should also be quite easy. Yes, they are part of the cache, but all the already running workloads will have the tolerations already injected, and their status of matching the flavor will not change.

Would you like to follow up on this effort here wrt tolerations?

/lgtm
/approve

Issue #9428, opened 2026-02-23
Scheduler may occasionally send multiple preemption requests to the same workload (emitting multiple events)


mimowo:
**What happened**:

Scheduler may issue multiple preemption requests for workloads which it already preempted.

**What you expected to happen**:

A single preemption request is sent, and a single event emitted.

**How to reproduce it (as minimally and precisely as possible)**:

I'm not sure how common this is, but we definitely observe that in longs on some deployments. 

**Anything else we need to know?**:

I think the bug is likely around this line: https://github.com/kubernetes-sigs/kueue/blob/main/pkg/scheduler/preemption/preemption.go#L173

One idea for solving is to use the in-memory "PreemptionExpectation" which are added by scheduler when preempting and removed by workload_controller when the workload is updated ("PreemptionExpectation" is observed). Before the "PreemptionExpectation" is observed the scheduler skips sending the request and emitting the event.

PR #9437, opened 2026-02-24
Prevent duplicate preemption requests across scheduling cycles

sohankunkerkar:
Use `expectations.Store` (same pattern as pod controller) to track workloads with in-flight preemptions. The expectation is set before calling Evict and rolled back on failure. It is cleared by the workload controller when the eviction (or deletion) is observed.                                                                                                                                                                                                                                                                                                                            

<!--  Thanks for sending a pull request!  Here are some tips for you:

1. If this is your first time, please read our contributor guidelines: https://git.k8s.io/community/contributors/guide/first-contribution.md#your-first-contribution and developer guide https://git.k8s.io/community/contributors/devel/development.md#development-guide
2. Please label this pull request according to what type of issue you are addressing, especially if this is a release targeted pull request. For reference on required PR/issue labels, read here:
https://git.k8s.io/community/contributors/devel/sig-release/release.md#issuepr-kind-label
3. Ensure you have added or ran the appropriate tests for your PR: https://git.k8s.io/community/contributors/devel/sig-testing/testing.md
4. If you want *faster* PR reviews, read how: https://git.k8s.io/community/contributors/guide/pull-requests.md#best-practices-for-faster-reviews
5. If the PR is unfinished, see how to mark it: https://git.k8s.io/community/contributors/guide/pull-requests.md#marking-unfinished-pull-requests
-->

#### What type of PR is this?
/kind bug

<!--
*Automatically closes linked issue when PR is merged.
Usage: `Fixes #<issue number>`, or `Fixes (paste link of issue)`.
_If PR is about `failing-tests or flakes`, please post the related issues/tests in a comment and do not use `Fixes`_*
-->
Fixes https://github.com/kubernetes-sigs/kueue/issues/9428

I was thinking about implementing a `sync.Map` on Preemptor approach which is simpler, but entries could persist indefinitely if a workload was evicted but never targeted  again. Also required a `schedulingCycle` parameter to avoid blocking legitimate within-cycle multi-preemptor targeting of the same workload.

<!--
If no, just write "NONE" in the release-note block below.
If yes, a release note is required:
Enter your extended release note in the block below. If the PR requires additional action from users switching to the new release, include the string "action required".
-->
```release-note
Scheduling: Fix that the Kueue's scheduler could issue duplicate preemption requests and events for the same workload.
```

### DIFF
--- a/cmd/kueue/main.go
+++ b/cmd/kueue/main.go
@@ -72,8 +72,10 @@ import (
 	"sigs.k8s.io/kueue/pkg/features"
 	"sigs.k8s.io/kueue/pkg/metrics"
 	"sigs.k8s.io/kueue/pkg/scheduler"
+	preemptexpectations "sigs.k8s.io/kueue/pkg/scheduler/preemption/expectations"
 	"sigs.k8s.io/kueue/pkg/scheduler/preemption/fairsharing"
 	"sigs.k8s.io/kueue/pkg/util/cert"
+	"sigs.k8s.io/kueue/pkg/util/expectations"
 	"sigs.k8s.io/kueue/pkg/util/kubeversion"
 	utillogging "sigs.k8s.io/kueue/pkg/util/logging"
 	"sigs.k8s.io/kueue/pkg/util/roletracker"
@@ -332,7 +334,9 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err := setupControllers(ctx, mgr, cCache, queues, &cfg, serverVersionFetcher, roleTracker); err != nil {
+	preemptionExpectations := preemptexpectations.New()
+
+	if err := setupControllers(ctx, mgr, cCache, queues, &cfg, serverVersionFetcher, roleTracker, preemptionExpectations); err != nil {
 		setupLog.Error(err, "Unable to setup controllers")
 		os.Exit(1)
 	}
@@ -354,7 +358,7 @@ func main() {
 		}()
 	}
 
-	if err := setupScheduler(mgr, cCache, queues, &cfg, roleTracker); err != nil {
+	if err := setupScheduler(mgr, cCache, queues, &cfg, roleTracker, preemptionExpectations); err != nil {
 		setupLog.Error(err, "Could not setup scheduler")
 		os.Exit(1)
 	}
@@ -397,8 +401,8 @@ func setupIndexes(ctx context.Context, mgr ctrl.Manager, cfg *configapi.Configur
 	return jobframework.SetupIndexes(ctx, mgr.GetFieldIndexer(), opts...)
 }
 
-func setupControllers(ctx context.Context, mgr ctrl.Manager, cCache *schdcache.Cache, queues *qcache.Manager, cfg *configapi.Configuration, serverVersionFetcher *kubeversion.ServerVersionFetcher, roleTracker *roletracker.RoleTracker) error {
-	if failedCtrl, err := core.SetupControllers(mgr, queues, cCache, cfg, roleTracker); err != nil {
+func setupControllers(ctx context.Context, mgr ctrl.Manager, cCache *schdcache.Cache, queues *qcache.Manager, cfg *configapi.Configuration, serverVersionFetcher *kubeversion.ServerVersionFetcher, roleTracker *roletracker.RoleTracker, preemptionExpectations *expectations.Store) error {
+	if failedCtrl, err := core.SetupControllers(mgr, queues, cCache, cfg, roleTracker, preemptionExpectations); err != nil {
 		return fmt.Errorf("unable to create controller %s: %w", failedCtrl, err)
 	}
 	if features.Enabled(features.FailureRecoveryPolicy) {
@@ -519,7 +523,7 @@ func setupProbeEndpoints(mgr ctrl.Manager, certsReady <-chan struct{}) error {
 	return nil
 }
 
-func setupScheduler(mgr ctrl.Manager, cCache *schdcache.Cache, queues *qcache.Manager, cfg *configapi.Configuration, roleTracker *roletracker.RoleTracker) error {
+func setupScheduler(mgr ctrl.Manager, cCache *schdcache.Cache, queues *qcache.Manager, cfg *configapi.Configuration, roleTracker *roletracker.RoleTracker, preemptionExpectations *expectations.Store) error {
 	sched := scheduler.New(
 		queues,
 		cCache,
@@ -529,6 +533,7 @@ func setupScheduler(mgr ctrl.Manager, cCache *schdcache.Cache, queues *qcache.Ma
 		scheduler.WithFairSharing(cfg.FairSharing),
 		scheduler.WithAdmissionFairSharing(cfg.AdmissionFairSharing),
 		scheduler.WithRoleTracker(roleTracker),
+		scheduler.WithPreemptionExpectations(preemptionExpectations),
 	)
 	if err := mgr.Add(sched); err != nil {
 		return fmt.Errorf("unable to add scheduler to manager: %w", err)
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -59,6 +59,7 @@ import (
 	"sigs.k8s.io/kueue/pkg/util/admissioncheck"
 	afs "sigs.k8s.io/kueue/pkg/util/admissionfairsharing"
 	clientutil "sigs.k8s.io/kueue/pkg/util/client"
+	"sigs.k8s.io/kueue/pkg/util/expectations"
 	qutil "sigs.k8s.io/kueue/pkg/util/queue"
 	"sigs.k8s.io/kueue/pkg/util/resource"
 	"sigs.k8s.io/kueue/pkg/util/roletracker"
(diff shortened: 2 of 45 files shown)


pkg/scheduler/preemption/preemption.go
```diff
@@ -170,33 +175,47 @@ func (p *Preemptor) IssuePreemptions(ctx context.Context, preemptor *workload.In
...
+			ctx, p.client, p.recorder, wlCopy, kueue.WorkloadEvictedByPreemption, message, "", p.clock, p.roleTracker,
+			workload.WithCustomPrepare(func(wl *kueue.Workload) {
+				workload.SetPreemptedCondition(wl, p.clock.Now(), target.Reason, message)
+			}),
+			workload.EvictWithLooseOnApply(), workload.EvictWithRetryOnConflictForPatch(),
+		)
+		if err != nil {
+			if p.preemptionExpectations != nil {
+				p.preemptionExpectations.ObservedUID(log, targetKey, target.WorkloadInfo.Obj.UID)
+			}
```
mimowo: Hm, I'm thinking if we could ensure `preemptionExpectations` is always non-nil, as we assume for qManager and cCache. 

Two ideas I have:
1. put the expectations under one of the existing caches: schdcache, or queues cache
2. wrap all caches under the "cache" param. 

I'm thinking (2.) sounds better long term. The scheduler is already using both cCache and qManager so the distinction is not that clear anyway, Let me know your thoughts. I'm ok-ish for now as is, but I would like to agree on some direction wrt caches.  

cc @gabesaba @tenzen-y

pkg/scheduler/preemption/preemption.go
```diff
@@ -170,33 +175,47 @@ func (p *Preemptor) IssuePreemptions(ctx context.Context, preemptor *workload.In
...
+			errCh.SendErrorWithCancel(err, cancel)
+			preemptionErrors.Add(1)
+			return
 		}
+		log.V(3).Info("Preempted", "targetWorkload", klog.KObj(target.WorkloadInfo.Obj), "preemptingWorkload", klog.KObj(preemptor.Obj), "preemptorUID", string(preemptor.Obj.UID),
```
mwielgus, line 212: Can you include what other clusters wanted to preempt as well?

pkg/scheduler/preemption/preemption.go
sohankunkerkar: Good question. I looked at both options.
Option 1 is tempting since both the scheduler and workload controller already have the cache. But `schdcache.Cache` is fundamentally about state that gets snapshotted for scheduling decisions. Expectations are live coordination state between components, not schedulable state. Embedding them in the cache means either including them in Snapshot() (wrong, they're mutable) or excluding them (leaky, the cache becomes a mix of snapshotable and non-snapshotable fields).

Option 2 feels like the right long-term direction. As you described, the scheduler already uses both cCache and qManager, and threading more params through SetupControllers doesn't scale. But the scope is large, touching every controller setup path, so it's better as a dedicated follow-up than a side effect of this bug fix.

Happy to open a follow-up issue to track the cache consolidation discussion.

pkg/scheduler/preemption/preemption.go
sohankunkerkar: Done! I added `preemptingWorkload` to the skip log so it shows which workload triggered the duplicate attempt.

pkg/scheduler/preemption/preemption.go
mimowo: I agree, a follow up issue makes sense. 

I think because we don't have the umbrella cache concept we already put some cache concepts into the  pre-existing cache (as you point out discussing 1.). For example I'm thinking about the `finishedWorkloads` - they could probably be a sub-cache on their own - they are only used for metrics tracking currently. OTOH, the pre-existing cache is not too bad for them either.

sohankunkerkar: Good point about `finishedWorkloads`. Similar situation where shared state sits inside a cache because there wasn't a dedicated place for it. A unified cache struct would clean that up. Will open a follow-up issue for that. For now the expectations store defaults to non-nil in the constructor so there are fewer nil checks.

pkg/scheduler/preemption/preemption.go
```diff
@@ -85,16 +88,21 @@ func New(
 	enabledAfs bool,
 	clock clock.Clock,
 	tracker *roletracker.RoleTracker,
+	preemptionExpectations *expectations.Store,
 ) *Preemptor {
+	if preemptionExpectations == nil {
+		preemptionExpectations = expectations.NewStore("preemptions")
+	}
```
mimowo: Why are we initializing the new store here, this risks that there are two instances. If this is just for tests, then I would rather adjust tests.

sohankunkerkar: Ok, I will move the nil-default out of `preemption.New`. Right now the fallback sits in scheduler.New to avoid updating 29 test callsites. If you'd rather have no fallback at all, I can update the tests to pass `WithPreemptionExpectations(expectations.NewStore(...))` explicitly. Let me know which you prefer.

mimowo: Thank you for the summary - so tests only changes as I thought. Yes, I propose to cleanup the production code:
1. drop the fallbacks instantiating preemptionExpectations inside components - they risk "double cache"
2. drop nil checks inside controllers - they are risking some controllers don't have the cache initialized, don't panic, but don't work correctly

I'm ok with the test adjustments

pkg/controller/core/workload_controller.go
```diff
@@ -878,6 +887,10 @@ func (r *WorkloadReconciler) Delete(e event.TypedDeleteEvent[*kueue.Workload]) b
 		status = workload.Status(e.Object)
 	}
 	r.logger().V(2).Info("Workload delete event", "workload", klog.KObj(e.Object), "queue", e.Object.Spec.QueueName, "status", status)
+	if r.preemptionExpectations != nil && !e.DeleteStateUnknown {
```
mimowo: drop the nil checks for preemptionExpectations, they risk covering non-initialized expecations, as mentioned in the sibling comment

test/e2e/singlecluster/certs_test.go
```diff
@@ -0,0 +1,240 @@
+/*
+Copyright The Kubernetes Authors.
```
mimowo: This file was moved in my recent PR, do I dont expect it on the diff here, maybe some issue during rebase?

test/integration/multikueue/scheduler/suite_test.go
```diff
@@ -149,7 +150,7 @@ func managerSetup(ctx context.Context, mgr manager.Manager) {
 	err = reconciler.SetupWithManager(mgr)
 	gomega.Expect(err).NotTo(gomega.HaveOccurred())
 
-	sched := scheduler.New(queues, cCache, mgr.GetClient(), mgr.GetEventRecorderFor(constants.AdmissionName))
+	sched := scheduler.New(queues, cCache, mgr.GetClient(), mgr.GetEventRecorderFor(constants.AdmissionName), scheduler.WithPreemptionExpectations(expectations.NewStore("preemptions")))
```
mimowo: I have a doubt about the test, when creating controllers with `SetupControllers` and scheduler with `scheduler.New` a new instance is created - two total, so how does it work? 

Is it maybe not working, but the tests are not depending on the expecations? In that case it may take long debugging time for someone who adds such a test in the future.

test/integration/multikueue/scheduler/suite_test.go
```diff
@@ -115,7 +116,7 @@ func managerSetup(ctx context.Context, mgr manager.Manager) {
 	configuration := &config.Configuration{}
 	mgr.GetScheme().Default(configuration)
 
-	failedCtrl, err := core.SetupControllers(mgr, queues, cCache, configuration, nil)
+	failedCtrl, err := core.SetupControllers(mgr, queues, cCache, configuration, nil, expectations.NewStore("preemptions"))
```
mimowo: Should we introduce a helper like `preemption.NewExpectations` which would just call `expectations.NewStore("preemptions")` - to encapsulate the name, and avoid typos, etc.

sohankunkerkar: Done! I added preemption.NewExpectations() which encapsulates theexpectations.NewStore("preemptions") call. Note: main.go and minimalkueue can't use it due to an import cycle, so those two files use expectations.NewStore("preemptions") directly.

test/integration/multikueue/scheduler/suite_test.go
sohankunkerkar: Good catch. Fixed.

test/e2e/singlecluster/certs_test.go
sohankunkerkar: Fixed. It was from a stale rebase. The current diff no longer includes this file.

cmd/kueue/main.go
```diff
@@ -332,7 +333,9 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err := setupControllers(ctx, mgr, cCache, queues, &cfg, serverVersionFetcher, roleTracker); err != nil {
+	preemptionExpectations := expectations.NewStore("preemptions")
```
mimowo: Let's also use preemption.NewExpectations here to ensure the prod code is using it the same way as tests

sohankunkerkar: I can't use preemption.NewExpectations() in main.go due to an import cycle. I can move the helper to  `pkg/util/expectations as NewPreemptionStore()` if you'd prefer a single entry point, or keep `expectations.NewStore("preemptions")` here. Let me know which you prefer.

mimowo: I see, what about a new package `pkg/scheduler/preemption/expectations`, currently with only one file, and the helper function.

PR #9394, opened 2026-02-20
Fix: fix cohortName field for `v1beta2.ClusterQueue`

polinasand:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

For v1beta2:
When running `kubectl get clusterqueue` command, the expected output is:
```
NAME                                        COHORT   PENDING WORKLOADS
clusterqueue.kueue.x-k8s.io/cluster-queue   cohort-1         0
```
But the actual output has cohort name missing:
```
NAME                                        COHORT   PENDING WORKLOADS
clusterqueue.kueue.x-k8s.io/cluster-queue            0
```
This PR fixes this bug.

Manual testing:
Before:
Screenshot 2026-02-23 at 11 02 55

After:
Screenshot 2026-02-23 at 10 53 45

#### Which issue(s) this PR fixes:

Fixes #9281

#### Does this PR introduce a user-facing change?
```release-note
Fixed invalid field name in the `ClusterQueue` printer columns. The "Cohort" column will now correctly display the assigned cohort in kubectl, k9s, and other UI tools instead of being blank.
```

PR #9447, opened 2026-02-24
[release-0.15] Fix: fix cohortName field for v1beta2.ClusterQueue

polinasand:
This is a cherry-pick of https://github.com/kubernetes-sigs/kueue/pull/9394

```release-note
Fixed invalid field name in the `ClusterQueue` printer columns. The "Cohort" column will now correctly display the assigned cohort in kubectl, k9s, and other UI tools instead of being blank.
```

### DIFF
--- a/apis/kueue/v1beta2/clusterqueue_types.go
+++ b/apis/kueue/v1beta2/clusterqueue_types.go
@@ -256,7 +256,7 @@ type ResourceQuota struct {
 	// ClusterQueues in the cohort have enough unused quota.
 	// If null, it means that there is no borrowing limit.
 	// If not null, it must be non-negative.
-	// borrowingLimit must be null if spec.cohort is empty.
+	// borrowingLimit must be null if spec.cohortName is empty.
 	// +optional
 	BorrowingLimit *resource.Quantity `json:"borrowingLimit,omitempty"`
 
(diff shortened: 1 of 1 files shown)


kshalot:
PTAL https://github.com/kubernetes/community/blob/master/contributors/devel/sig-release/cherry-picks.md#steps

mimowo:
I have fixed the PR description manuall by adding the `release-note` section. ptal

mimowo:
Thanks, it is better to use the ./hack/cherry_pick_pull.sh script to save on some routine work like setting the categorization labels, etc.
/lgtm
/approve
/kind bug

mimowo:
/release-note-edit
```release-note
CLI: Fix the bug that cohort name wasn't actually displayed in the output when running `kubectl get clusterqueue` command.
```

tenzen-y:
> /release-note-edit
> 
> ```
> CLI: Fix the bug that cohort name wasn't actually displayed in the output when running `kubectl get clusterqueue` command.
> ```

@mimowo Actually, this is not only for CLI. This fixes the incorrect CRD, then fixes the schema marshaling problem.

mimowo:
@tenzen-y are you sure? The original PR description shows pretty clearly the issue is with Kubectl CLI:

image

tenzen-y:
> @tenzen-y are you sure? The original PR description shows pretty clearly the issue is with Kubectl CLI:
> 
> image

Sorry for the confusion. Actually it's not for marsharing. I wanted to say about kube-api-server field checking when any type of client requests with table format.

Previously, kube-apiserver checks the `.spec.cohort` field, then returns an empty field due to an invalid path.

mimowo:
> Previously, kube-apiserver checks the .spec.cohort field, then returns an empty field due to an invalid path.

Right, but kube-apiserver does it for the needs of the table format used by kubectl CLI, so it seems fair to say from the user-perspective it is CLI that is affected. However, I see that any other tool which would use the table formatting by kube-apiserver would be affected. So, we could generalize the message. 

Maybe:

kshalot:
From my perspective, sth like this would make sense:
```
Fixed invalid field name in the `ClusterQueue` printer columns. The "Cohort" column will now correctly display the assigned cohort in kubectl, k9s, and other UI tools instead of being blank.
```

PR #9460, opened 2026-02-24
Update RayService doc due to recent top level job change

### LINKED ISSUE #9059: KubeRay integration: Support RayService as a top-level Job

I would like to introduce support for RayService as a top-level Job, rather than via RayCluster.

**Why is this needed**:

To support dispatching RayService via MultiKueue, see related effort in KubeRay: https://github.com/ray-project/kuberay/issues/4486

Currently RayService can only be managed by Kueue at the RayCluster level, see https://kueue.sigs.k8s.io/docs/tasks/run/rayservices/ but this solution does not transfer fully to MultiKueue, because then the Gateway is not created on the worker cluster.

hiboyang:
#### What type of PR is this?

/kind documentation

RayService is treated a top level job now, need to update doc accordingly.

Fixes https://github.com/kubernetes-sigs/kueue/issues/9059

```release-note
NONE
```

### DIFF
--- a/site/content/en/docs/tasks/run/rayservices.md
+++ b/site/content/en/docs/tasks/run/rayservices.md
@@ -21,6 +21,11 @@ This guide is for [serving users](/docs/tasks#serving-user) that have a basic un
 
 3. See [KubeRay Installation](https://docs.ray.io/en/latest/cluster/kubernetes/getting-started/kuberay-operator-installation.html) for installation and configuration details of KubeRay.
 
+{{% alert title="Note" color="primary" %}}
+RayService is managed by Kueue through RayCluster, and in order to use RayCluster, prior to v0.17.0, you need to restart Kueue after the installation.
+You can do it by running: `kubectl delete pods -l control-plane=controller-manager -n kueue-system`.
+{{% /alert %}}
+
 ## RayService definition
 
 When running [RayService](https://docs.ray.io/en/latest/cluster/kubernetes/getting-started/rayservice-quick-start.html) on Kueue, take into consideration the following aspects:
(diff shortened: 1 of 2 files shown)


site/content/en/docs/tasks/run/rayservices.md
```diff
@@ -22,7 +22,7 @@ This guide is for [serving users](/docs/tasks#serving-user) that have a basic un
 3. See [KubeRay Installation](https://docs.ray.io/en/latest/cluster/kubernetes/getting-started/kuberay-operator-installation.html) for installation and configuration details of KubeRay.
 
 {{% alert title="Note" color="primary" %}}
-RayService is managed by Kueue through RayCluster, and in order to use RayCluster, prior to v0.8.1, you need to restart Kueue after the installation.
+RayService is managed by Kueue through RayCluster, and in order to use RayCluster, prior to v0.17.0, you need to restart Kueue after the installation.
```
mimowo: Actually, I think this is no longer needed at all - to restart Kueue after any CRD installation. We can just drop the alerts already for all CRDs I think.

hiboyang: cool, removed those contents

site/content/en/docs/tasks/run/rayservices.md
```diff
@@ -21,11 +21,6 @@ This guide is for [serving users](/docs/tasks#serving-user) that have a basic un
 
 3. See [KubeRay Installation](https://docs.ray.io/en/latest/cluster/kubernetes/getting-started/kuberay-operator-installation.html) for installation and configuration details of KubeRay.
 
-{{% alert title="Note" color="primary" %}}
```
mimowo, line 24: Let's keep this in this PR as is, we can have another PR which deletes the note for all CRDs.

hiboyang: ok, added them back

mimowo (commented):
/lgtm
/approve
Thank you 👍

site/content/en/docs/tasks/run/rayservices.md
win5923, line 68: I think there may be a misunderstanding here. Kueue only controls `RayService.Spec.RayClusterSpec.Suspend` for now.

https://github.com/kubernetes-sigs/kueue/blob/cead7dee9ef55593edace52f961e949f4ae47997/pkg/controller/jobs/rayservice/rayservice_controller.go#L134-L136

Issue #9462, opened 2026-02-24
PreemptionOverBorrowing FlavorFungibility leading to undesirable scheduling behavior


varunsyal:
**What happened**:
While moving from v0.15 to v0.16, we are trying to remove the feature gate "FlavorFungibilityImplicitPreferenceDefault" that we have been using but no longer supported since v0.16. In order to enable the same functionality without the feature flag, we have set the FlavorFungibility as the following on the Clusterqueues:
```
Flavor Fungibility:
    Preference:        PreemptionOverBorrowing
    When Can Borrow:   TryNextFlavor
    When Can Preempt:  TryNextFlavor
```
We observe that the preemption over borrowing functionality across 2 flavors is not working as expected.
In the example we run:

We have 3 Clusterqueues under the same Cohort

CQ-1:
   > FairSharing Weight = 1
    Flavor 1: Nominal Quota = 1cpu, Borrowing Limit = 4cpu
    Flavor 2: Nominal Quota = 0, Borrowing Limit = 4cpu

CQ-2:
>    FairSharing Weight = 1
    Flavor 1: Nominal Quota = 1, Borrowing Limit = 4cpu
    Flavor 2: Nominal Quota = 0, Borrowing Limit = 4cpu

CQ-3:
>    FairSharing Weight = 1
    Flavor 1: Nominal Quota = 0, Borrowing Limit = 4cpu
    Flavor 2: Nominal Quota = 2, Borrowing Limit = 4cpu

Now we submit 2 workloads requesting 1 cpu each to CQ-1
> Result: Both workloads go to Flavor-1

Next, we submit 2 workloads requesting 1 cpu each to CQ-2
>    Result: Both workloads go to Flavor-2 ---> This is not expected, since CQ-2 has nominal quota in Flavor-1, so before borrowing in Flavor-2, it should preempt in Flavor-1
    Expectation is: Both CQ-1 and CQ-2 eventually have 1 workload running in each of its flavor.

Adding YAML test files:
[Test-setup-scenario-multiflavor-fungibility.zip](https://github.com/user-attachments/files/25531385/Test-setup-scenario-multiflavor-fungibility.zip)

**Anything else we need to know?**:

**Environment**: 
- Kubernetes version (use `kubectl version`): v1.32.11
- Kueue version (use `git describe --tags --dirty --always`): v0.15.3
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
[...]

PR #9463, opened 2026-02-25
WIP: Reproduce a flavor fungibility preference bug

tenzen-y:
#### What type of PR is this?

/kind bug

#### Which issue(s) this PR fixes:

Part-of https://github.com/kubernetes-sigs/kueue/issues/9462

#### Does this PR introduce a user-facing change?

```release-note
TODO
```

PR #9459, opened 2026-02-24
Fix race in KubeConfigFSWatcher tests by making Start() blocking

IrvingMg:
#### What type of PR is this?

/kind bug
/kind flake

#### What this PR does / why we need it:

Fixes a test race in `KubeConfigFSWatcher` where the goroutine spawned by `Start()` could log to `testing.T` after the test returned. The root cause was `Start()` being non-blocking, it spawned a goroutine and returned immediately, violating the controller-runtime `Runnable` contract. Making `Start()` blocking fixes both the race and the contract compliance.

In production, no changes are required as `mgr.Add()` registers the watcher with the controller-runtime manager, which calls `Start()` internally.

#### Which issue(s) this PR fixes:

Fixes #9454

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #9464, opened 2026-02-25
Fix a flavorFungibility preference bug

tenzen-y:
#### What type of PR is this?

/kind bug

I swapped actions for `flavorFungability.preference` enum values in flavor assigner because the current enums do not match with the correct actions.

Fixes https://github.com/kubernetes-sigs/kueue/issues/9462

You can check the reported problem reproduces in https://github.com/kubernetes-sigs/kueue/pull/9463

```release-note
FlavorFungability: fix the bug that the semantics for the `flavorFungability.preference` enum values
(ie. PreemptionOverBorrowing and BorrowingOverPreemption) were swapped.
```

### DIFF
--- a/pkg/scheduler/flavorassigner/flavorassigner.go
+++ b/pkg/scheduler/flavorassigner/flavorassigner.go
@@ -412,9 +412,9 @@ func isPreferred(a, b granularMode, fungibilityConfig kueue.FlavorFungibility) b
 	if fungibilityConfig.Preference != nil {
 		switch *fungibilityConfig.Preference {
 		case kueue.BorrowingOverPreemption:
-			return preemptionOverBorrowing()
-		case kueue.PreemptionOverBorrowing:
 			return borrowingOverPreemption()
+		case kueue.PreemptionOverBorrowing:
+			return preemptionOverBorrowing()
 		}
 	}
 
--- a/pkg/scheduler/flavorassigner/flavorassigner_test.go
+++ b/pkg/scheduler/flavorassigner/flavorassigner_test.go
@@ -3910,7 +3910,7 @@ func TestIsPreferred(t *testing.T) {
 				WhenCanPreempt: kueue.TryNextFlavor,
 				Preference:     makePref(kueue.BorrowingOverPreemption),
 			},
-			wantPreferred: true,
+			wantPreferred: false,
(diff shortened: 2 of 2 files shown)


mimowo:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/9464/pull-kueue-test-unit-main/2026555898335334400
/test pull-kueue-test-unit-main
This is unrelated issue, I think already fixed here https://github.com/kubernetes-sigs/kueue/pull/9459

mimowo:
Proposal;
/release-note-edit
```release-note
FlavorFungability: fix the bug that the semantics for the `flavorFungability.preference` enum values
(ie. PreemptionOverBorrowing and BorrowingOverPreemption) were swapped.
```

pkg/scheduler/scheduler_test.go
```diff
@@ -7635,6 +7635,343 @@ func TestSchedule(t *testing.T) {
 				utiltesting.MakeEventRecord("sales", "preemptor", "Pending", corev1.EventTypeWarning).Obj(),
 			},
 		},
+		"PreemptionOverBorrowing preference: preempt in first flavor instead of borrowing in second": {
```
mimowo, line 7638: Nice tests, clearly showing the preference on minimal examples.

I believe the lack of such tests was the reason the bug slipped in. We only added the `wantPreference` tests which make it easy to miss.

mimowo:
/retitle Fix a flavorFungibility preference bug
I think this is ready to go, the change is small, and the tests are really good.
Thank you 👍 

/lgtm
/approve
/cherrypick release-0.16

mimowo:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/9464/pull-kueue-test-e2e-main-1-34/2026684207736033280
flake, will open issue

pkg/scheduler/scheduler_test.go
tenzen-y: Yes, exactly. As I checked the preference implementation PR, there are no comprehensive scheduler tests 😓 
So, I believe that this could prevent additional bugs.

PR #9484, opened 2026-02-25
[Fair Sharing] Prefer nominal-quota workloads in admission ordering

### LINKED ISSUE #9406: Fair sharing admission ordering does not preserve nominal-first workload priority

mukund-wayve:
#### What type of PR is this?

/kind bug

The fair sharing iterator's admission ordering compares workloads by DRS, Priority and then FIFO. This means a workload that fits within its CQ's nominal quota can be deprioritized because the CQ's aggregate DRS is inflated by borrowing on a different flavor.

The classical iterator already handles this — it prefers workloads with lower borrowing levels before comparing other criteria. This PR adds a similar nominal-first ordering to the fair sharing iterator: workloads that don't require borrowing are always preferred over those that do, before falling through to DRS and other comparisons. 

Not adding a feature gate since it's a simple fix. 

Part-of #9406

This is part 1/n of the split version of #9407 which attempts to resolve #9406.

```release-note
FairSharing: workloads fitting within their ClusterQueue's nominal quota are now preferred over workloads that require borrowing, preventing heavy borrowing on one flavor from deprioritizing a CQ's nominal entitlement on another flavor.
```

### DIFF
--- a/pkg/features/kube_features.go
+++ b/pkg/features/kube_features.go
@@ -291,7 +291,7 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 		{Version: version.MustParse("0.6"), Default: true, PreRelease: featuregate.Beta},
 	},
 	FairSharingPrioritizeNonBorrowing: {
-		{Version: version.MustParse("0.17"), Default: true, PreRelease: featuregate.Alpha},
+		{Version: version.MustParse("0.17"), Default: true, PreRelease: featuregate.Beta},
 	},
 	MultiKueue: {
 		{Version: version.MustParse("0.6"), Default: false, PreRelease: featuregate.Alpha},
--- a/pkg/scheduler/scheduler_test.go
+++ b/pkg/scheduler/scheduler_test.go
@@ -3091,9 +3091,9 @@ func TestSchedule(t *testing.T) {
 					Obj(),
 			},
 			wantAssignments: map[workload.Reference]kueue.Admission{
-				"eng-alpha/spot-heavy":  *utiltestingapi.MakeAdmission("fs-nom-a").PodSets(utiltestingapi.MakePodSetAssignment("one").Assignment(corev1.ResourceCPU, "spot", "80").Count(80).Obj()).Obj(),
-				"eng-beta/light-usage":  *utiltestingapi.MakeAdmission("fs-nom-b").PodSets(utiltestingapi.MakePodSetAssignment("one").Assignment(corev1.ResourceCPU, "on-demand", "5").Count(5).Obj()).Obj(),
-				"eng-alpha/nom-wl":      *utiltestingapi.MakeAdmission("fs-nom-a").PodSets(utiltestingapi.MakePodSetAssignment("one").Assignment(corev1.ResourceCPU, "on-demand", "45").Count(45).Obj()).Obj(),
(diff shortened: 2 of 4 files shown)


pkg/scheduler/fair_sharing_iterator.go
```diff
@@ -164,15 +164,24 @@ type entryComparer struct {
 }
 
 func (e *entryComparer) less(a, b *entry, parentCohort kueue.CohortReference) bool {
+	// 1: Nominal first — a workload that fits within its CQ's nominal
+	// quota is always preferred over one that requires borrowing.
+	// This prevents heavy borrowing on one flavor from eroding a
+	// CQ's nominal entitlement on another flavor.
+	aRequiresBorrowing := a.assignment.RequiresBorrowing()
```
mimowo: i think this is a nice bugfix. initially i thought this is no-op, but actually it allows to schedule workloads which fit within nominal quota even if increasing CQs DRS. I think this is ok, but may have non trivial consequences so I would propose to add a dedicated FairSharingPrioritizeNonBorrowing feature gate, as bailout option. Also, it would be great to add a unit test in schedule_test.go which demonstrates the semantic change in scheduling with fair sharing - not just orderibg fubction alone.

pkg/features/kube_features.go
```diff
@@ -283,6 +290,9 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 	PrioritySortingWithinCohort: {
 		{Version: version.MustParse("0.6"), Default: true, PreRelease: featuregate.Beta},
 	},
+	FairSharingPrioritizeNonBorrowing: {
+		{Version: version.MustParse("0.17"), Default: true, PreRelease: featuregate.Alpha},
```
mukund-wayve: I don't fully understand how the cherry-picking mechanism works. Given we are still maintaining 0.15, 0.16, is it okay to set the version to be 0.15 here?

mimowo: Yeah, there is not "best" option. What we typically do we put 0.17 here, but when preparing cherrypicks with the ./hack/cherry_pick_pull.sh we manually adjust the version to the target, so 0.15 and 0.16 respectively.

1, you can already prepare the CPs with the script
2. I would say Beta, this is most often considered as "enabled by default"

pkg/scheduler/fair_sharing_iterator.go
mukund-wayve: Done. Added a test in `scheduler_test.go` following the existing patterns

pkg/features/kube_features.go
mukund-wayve: Thanks. Made it beta so it becomes default. Will try to raise the CPs later today

mimowo:
/lgtm
/approve
Thank you 👍

pkg/features/kube_features.go
mimowo: > Will try to raise the CPs later today

Thank you, this will be great 👍

mimowo:
/release-note-edit
```release-note
FairSharing: workloads fitting within their ClusterQueue's nominal quota are now preferred over workloads that require borrowing, preventing heavy borrowing on one flavor from deprioritizing a CQ's nominal entitlement on another flavor.
```

PR #9433, opened 2026-02-23
Fix missing roleTracker in NonTasUsageReconciler

IrvingMg:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

`newNonTasUsageReconciler` doesn't receive the `roleTracker` parameter, so `r.roleTracker` is always `nil` and logs show `"replica-role": "standalone"` instead of the actual role.

#### Which issue(s) this PR fixes:

Fixes #9457 

#### Does this PR introduce a user-facing change?

```release-note
Observability: Fix missing "replica-role" in the logs from the NonTasUsageReconciler.
```

PR #9429, opened 2026-02-23
Fix stale replica-role in scheduler logs after leader election

IrvingMg:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

Moves role resolution from `Start()` to `schedule()`, resolving it dynamically each cycle.

#### Which issue(s) this PR fixes:

Fixes #9314 
Relates to #9287 

##### Verification
Running on the attached [e2e.log](https://github.com/user-attachments/files/25495571/e2e.log) 

```bash
grep -m1 '"schedulingCycle": 1,' e2e.log; grep -m1 'transitioned to leader' e2e.log; grep -m1 '"schedulingCycle": 2,' e2e.log
2026-02-23T16:17:18.612895618Z stderr F 2026-02-23T16:17:18.612459534Z  LEVEL(-3)       scheduler       queue/manager.go:683    Obtained ClusterQueue heads     {"replica-role": "follower", "schedulingCycle": 1, "count": 0}
2026-02-23T16:17:18.613480743Z stderr F 2026-02-23T16:17:18.613399034Z  INFO    setup   roletracker/tracker.go:59       RoleTracker: transitioned to leader
2026-02-23T16:17:30.485084013Z stderr F 2026-02-23T16:17:30.483801596Z  LEVEL(-3)       scheduler       queue/manager.go:683    Obtained ClusterQueue heads     {"replica-role": "leader", "schedulingCycle": 2, "count": 0}
```

#### Does this PR introduce a user-facing change?

```release-note
Observability: Fix the stale "replica-role" value in scheduler logs after leader election.
```

PR #9487, opened 2026-02-25
Fix missing replica_role=leader metrics after HA role transition

### LINKED ISSUE #9288: There is no data for the kueue_cluster_queue_nominal_quota with replica_role=leader
**What happened**:

A user reports there is no data reported for the `kueue_cluster_queue_nominal_quota` metric when `replica_role=leader` filter is present. There is data returned for `replica_role=follower` filter.

The `kueue_cluster_queue_nominal_quota` metric should also be returned for leader replica.

Forwarding issue reported by a user, but it should be just to create a deployment with two replicas, and check the metric for CQs. If this does not repro the issue we may investigate more. This issue was reported on 0.16.1.

IrvingMg:
#### What type of PR is this?

/kind bug

After an HA role transition from `follower` to `leader`, gauge metrics were still tagged with `replica_role` set to `follower` and were never re-reported with `replica_role` set to `leader`.

This change adds an `OnTransition` callback to `RoleTracker` that clears stale follower metrics and resynchronizes all metric-emitting components.

Fixes #9288 

```release-note
Observability: Fix missing replica_role=leader gauge metrics after HA role transition.
```

### DIFF
--- a/cmd/kueue/main.go
+++ b/cmd/kueue/main.go
@@ -272,6 +272,7 @@ func main() {
 	cacheOptions := []schdcache.Option{
 		schdcache.WithPodsReadyTracking(blockForPodsReady(&cfg)),
 		schdcache.WithRoleTracker(roleTracker),
+		schdcache.WithResourceMetrics(cfg.Metrics.EnableClusterQueueResources),
 	}
 	queueOptions := []qcache.Option{
 		qcache.WithPodsReadyRequeuingTimestamp(podsReadyRequeuingTimestamp(&cfg)),
@@ -332,6 +333,14 @@ func main() {
 	if err := setupProbeEndpoints(mgr, certsReady); err != nil {
 		setupLog.Error(err, "Unable to setup probe endpoints")
 		os.Exit(1)
+	}
+
+	if roleTracker != nil {
+		roleTracker.OnElected(func() {
+			metrics.ClearGaugeMetricsForRole(roletracker.RoleFollower)
+			cCache.ResyncGaugeMetrics()
+			queues.ResyncGaugeMetrics()
+		})
 	}
 
 	preemptionExpectations := preemptexpectations.New()
--- a/pkg/cache/queue/manager.go
+++ b/pkg/cache/queue/manager.go
@@ -863,6 +863,22 @@ func (m *Manager) queueSecondPass(ctx context.Context, w *kueue.Workload, iterat
 	}
 }
 
+// ResyncGaugeMetrics re-reports pending and finished workload gauge metrics.
+func (m *Manager) ResyncGaugeMetrics() {
+	m.RLock()
+	defer m.RUnlock()
+	for _, cq := range m.hm.ClusterQueues() {
+		reportCQPendingWorkloads(m, cq)
+		reportCQFinishedWorkloads(cq, m.roleTracker)
+	}
+	if features.Enabled(features.LocalQueueMetrics) {
+		for _, lq := range m.localQueues {
+			reportLQPendingWorkloads(m, lq)
+			reportLQFinishedWorkloads(m, lq)
+		}
+	}
+}
+
 type WorkloadUpdateWatcher interface {
 	NotifyWorkloadUpdate(oldWl, newWl *kueue.Workload)
 }
--- a/pkg/cache/scheduler/cache.go
+++ b/pkg/cache/scheduler/cache.go
@@ -25,6 +25,7 @@ import (
 	"sync"
 
 	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -41,6 +42,7 @@ import (
 	"sigs.k8s.io/kueue/pkg/metrics"
 	"sigs.k8s.io/kueue/pkg/resources"
 	"sigs.k8s.io/kueue/pkg/util/queue"
+	utilresource "sigs.k8s.io/kueue/pkg/util/resource"
 	"sigs.k8s.io/kueue/pkg/util/roletracker"
 	"sigs.k8s.io/kueue/pkg/workload"
 )
@@ -95,6 +97,12 @@ func WithAdmissionFairSharing(afs *config.AdmissionFairSharing) Option {
 	}
 }
 
+func WithResourceMetrics(enabled bool) Option {
+	return func(c *Cache) {
+		c.resourceMetricsEnabled = enabled
+	}
+}
+
 // WithRoleTracker sets the roleTracker for HA metrics.
 func WithRoleTracker(tracker *roletracker.RoleTracker) Option {
 	return func(c *Cache) {
@@ -107,13 +115,14 @@ type Cache struct {
 	sync.RWMutex
 	podsReadyCond sync.Cond
 
-	client               client.Client
-	resourceFlavors      map[kueue.ResourceFlavorReference]*kueue.ResourceFlavor
-	podsReadyTracking    bool
-	admissionChecks      map[kueue.AdmissionCheckReference]AdmissionCheck
-	workloadInfoOptions  []workload.InfoOption
-	fairSharingEnabled   bool
-	admissionFairSharing *config.AdmissionFairSharing
+	client                 client.Client
+	resourceFlavors        map[kueue.ResourceFlavorReference]*kueue.ResourceFlavor
+	podsReadyTracking      bool
+	admissionChecks        map[kueue.AdmissionCheckReference]AdmissionCheck
+	workloadInfoOptions    []workload.InfoOption
+	fairSharingEnabled     bool
+	admissionFairSharing   *config.AdmissionFairSharing
+	resourceMetricsEnabled bool
 	// Tracks Workload's ClusterQueue assignment throughout its presence in the cache, which is when they reserve quota (`QuotaReserved=True`).
 	workloadAssignedQueues map[workload.Reference]kueue.ClusterQueueReference
 
@@ -903,6 +912,36 @@ func (c *Cache) MatchingClusterQueues(nsLabels map[string]string) sets.Set[kueue
 		}
 	}
 	return cqs
+}
+
+// ResyncGaugeMetrics re-reports CQ/LQ status, active workload, resource, and weighted share gauge metrics.
+func (c *Cache) ResyncGaugeMetrics() {
+	c.RLock()
+	defer c.RUnlock()
+	for _, cq := range c.hm.ClusterQueues() {
+		metrics.ReportClusterQueueStatus(cq.Name, cq.Status, c.roleTracker)
+		cq.reportActiveWorkloads()
+		if c.resourceMetricsEnabled {
+			cq.reportResourceMetrics(c.fairSharingEnabled)
+		}
+		if features.Enabled(features.LocalQueueMetrics) {
+			for _, lq := range cq.localQueues {
+				lq.reportActiveWorkloads(c.roleTracker)
+				lq.reportResourceMetrics(cq.resourceNode.Quotas, c.roleTracker)
+			}
+		}
+	}
+	if c.fairSharingEnabled {
+		for _, cohort := range c.hm.Cohorts() {
+			drs := dominantResourceShare(cohort, nil)
+			metrics.ReportCohortWeightedShare(string(cohort.Name), drs.PreciseWeightedShare(), c.roleTracker)
+		}
+	}
+}
+
+func resourceFloat(name corev1.ResourceName, v int64) float64 {
+	q := resources.ResourceQuantity(name, v)
+	return utilresource.QuantityToFloat(&q)
 }
 
 // Key is the key used to index the queue.
--- a/pkg/cache/scheduler/clusterqueue.go
+++ b/pkg/cache/scheduler/clusterqueue.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"maps"
+	"math"
 	"slices"
 	"strings"
 
@@ -491,6 +492,40 @@ func (c *clusterQueue) reportActiveWorkloads() {
 	role := roletracker.GetRole(c.roleTracker)
 	metrics.AdmittedActiveWorkloads.WithLabelValues(string(c.Name), role).Set(float64(c.admittedWorkloadsCount))
 	metrics.ReservingActiveWorkloads.WithLabelValues(string(c.Name), role).Set(float64(len(c.Workloads)))
+}
+
+func (c *clusterQueue) reportResourceMetrics(fairSharingEnabled bool) {
+	var cohort kueue.CohortReference
+	if c.HasParent() {
+		cohort = c.Parent().GetName()
+	}
+	cqName := string(c.Name)
+	for fr, quota := range c.resourceNode.Quotas {
+		fName, rName := string(fr.Flavor), string(fr.Resource)
+		nominal := resourceFloat(fr.Resource, quota.Nominal)
+		var borrowing, lending float64
+		if quota.BorrowingLimit != nil {
+			borrowing = resourceFloat(fr.Resource, *quota.BorrowingLimit)
+		}
+		if quota.LendingLimit != nil {
+			lending = resourceFloat(fr.Resource, *quota.LendingLimit)
+		}
+		metrics.ReportClusterQueueQuotas(cohort, cqName, fName, rName, nominal, borrowing, lending, c.roleTracker)
+		metrics.ReportClusterQueueResourceReservations(cohort, cqName, fName, rName, resourceFloat(fr.Resource, c.resourceNode.Usage[fr]), c.roleTracker)
+		metrics.ReportClusterQueueResourceUsage(cohort, cqName, fName, rName, resourceFloat(fr.Resource, c.AdmittedUsage[fr]), c.roleTracker)
+	}
+	if fairSharingEnabled {
+		c.reportWeightedShare(cohort)
+	}
+}
+
+func (c *clusterQueue) reportWeightedShare(cohort kueue.CohortReference) {
+	drs := dominantResourceShare(c, nil)
+	weightedShare := drs.PreciseWeightedShare()
+	if weightedShare == math.Inf(1) {
+		weightedShare = math.NaN()
+	}
+	metrics.ReportClusterQueueWeightedShare(string(c.Name), string(cohort), weightedShare, c.roleTracker)
 }
 
 // updateWorkloadUsage updates the usage of the ClusterQueue for the workload
--- a/pkg/controller/core/core.go
+++ b/pkg/controller/core/core.go
@@ -100,6 +100,7 @@ func SetupControllers(mgr ctrl.Manager, qManager *qcache.Manager, cc *schdcache.
 	}
 	qManager.AddTopologyUpdateWatcher(cqRec)
 	qManager.AddWorkloadUpdateWatcher(qRec)
+
 	return "", nil
 }
 
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -879,6 +879,18 @@ func ReportCohortWeightedShare(cohort string, weightedShare float64, tracker *ro
 	CohortWeightedShare.WithLabelValues(cohort, roletracker.GetRole(tracker)).Set(weightedShare)
 }
 
+var allGaugeVecs []*prometheus.GaugeVec
+
+// ClearGaugeMetricsForRole deletes all gauge metric time series matching
+// replica_role=role. Called during HA role transitions to remove stale
+// time series reported under the old role.
+func ClearGaugeMetricsForRole(role string) {
+	lbls := prometheus.Labels{"replica_role": role}
+	for _, g := range allGaugeVecs {
+		g.DeletePartialMatch(lbls)
+	}
+}
+
 func ClearClusterQueueResourceMetrics(cqName string) {
 	lbls := prometheus.Labels{
 		"cluster_queue": cqName,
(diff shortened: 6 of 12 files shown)


pkg/controller/core/clusterqueue_controller.go
```diff
@@ -369,6 +369,31 @@ func (r *ClusterQueueReconciler) Generic(e event.TypedGenericEvent[*kueue.Cluste
 	return true
 }
 
+// ResyncMetrics re-reports CQ resource and weighted share gauge metrics.
+func (r *ClusterQueueReconciler) ResyncMetrics(ctx context.Context) {
+	if !r.reportResourceMetrics {
+		return
+	}
+	var cqList kueue.ClusterQueueList
```
mimowo: don't we have access to the list of CQs in the cache? listing is ok-ish but event handlers don't have error handling and the list may fail with error

pkg/metrics/metrics.go
```diff
@@ -847,6 +847,39 @@ func ReportCohortWeightedShare(cohort string, weightedShare float64, tracker *ro
 	CohortWeightedShare.WithLabelValues(cohort, roletracker.GetRole(tracker)).Set(weightedShare)
 }
 
+var roleTaggedGauges = []*prometheus.GaugeVec{
```
mimowo: Could we automate listing those by some fliters?

IrvingMg:
> @IrvingMg From your experience, are the logger reporting the correct role, while metrics not ?
> 
> In my case, I can always see the scheduler thread role to be follower, even in single-replica setups.

Over the past days, we’ve been working on a few fixes for the roleTracker. One of them was related to the scheduler: https://github.com/kubernetes-sigs/kueue/pull/9429, so this might be the one you’re experiencing.

We also fixed a couple of issue related to logs missing the roleTracker: https://github.com/kubernetes-sigs/kueue/pull/9445 and https://github.com/kubernetes-sigs/kueue/pull/9433.

pkg/controller/core/clusterqueue_controller.go
IrvingMg: I removed the API calls and moved resource metrics reporting into Cache.ResyncMetrics(). Thanks.

pkg/metrics/metrics.go
IrvingMg: I attempted to introduce a helper, `newRoleTaggedGaugeVec`, to add the role label and append it to the list automatically. 

However, this breaks the metrics doc generator because the helper hides `prometheus.NewGaugeVec`. 

I'd suggest to handle this in a follow-up.

test/integration/singlecluster/controller/core/clusterqueue_controller_test.go
```diff
@@ -1133,3 +1139,85 @@ var _ = ginkgo.Describe("ClusterQueue controller", ginkgo.Label("controller:clus
...
+			"cluster_queue": cq.Name,
+			"replica_role":  roletracker.RoleFollower,
+		}
+
+		ginkgo.By("Waiting for CQ to become active")
+		gomega.Eventually(func(g gomega.Gomega) {
+			var updatedCQ kueue.ClusterQueue
+			g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(cq), &updatedCQ)).To(gomega.Succeed())
+			g.Expect(apimeta.IsStatusConditionTrue(updatedCQ.Status.Conditions, kueue.ClusterQueueActive)).To(gomega.BeTrue())
+		}, util.Timeout, util.Interval).Should(gomega.Succeed())
```
mimowo: We have a helper for that. ptal, I think we have a function even which already waits for the CQ to be active when creating

pkg/metrics/metrics.go
mimowo: This is only used in roleTaggedGauges. So what I was thinking about is to list the metrics from `metrics.Registry`, and iterate over them. Then only clear those which belong to Kueue, and are guage. If this wouldn't work, then ok

pkg/cache/scheduler/cache.go
```diff
@@ -905,6 +915,78 @@ func (c *Cache) MatchingClusterQueues(nsLabels map[string]string) sets.Set[kueue
...
+		metrics.ReportClusterQueueResourceUsage(cohort, cqName, fName, rName, resourceFloat(fr.Resource, cq.AdmittedUsage[fr]), c.roleTracker)
+	}
+	if c.fairSharingEnabled {
+		drs := dominantResourceShare(cq, nil)
+		weightedShare := drs.PreciseWeightedShare()
+		if weightedShare == math.Inf(1) {
+			weightedShare = math.NaN()
+		}
+		metrics.ReportClusterQueueWeightedShare(cqName, string(cohort), weightedShare, c.roleTracker)
+	}
```
mimowo: This is a lot of code, don't we have already the code to report guage metrics we could call? I think we should try to commonize the code

test/integration/singlecluster/controller/core/clusterqueue_controller_test.go
IrvingMg: Done.

pkg/cache/scheduler/cache.go
IrvingMg: Moved reporting to `clusterQueue.reportResourceMetrics()` and `LocalQueue.reportResourceMetrics()` struct methods, following the `reportActiveWorkloads` pattern.

pkg/metrics/metrics.go
IrvingMg: The Prometheus client only allows reading metrics from the registry, so looks like it's not possible to clean up specific time series from it.

Instead, I added a `registerGaugeVecs()` function that registers each gauge and appends it to `allGaugeVecs`, ensuring it’s included in the role transition cleanup.

pkg/util/roletracker/tracker_test.go
```diff
@@ -89,6 +89,65 @@ func TestRoleTracker_StartLeaderElection(t *testing.T) {
...
+	}{
+		{
+			name:            "called on election with previous role",
+			registerCb:      true,
+			closeChannel:    true,
+			expectCalled:    true,
+			expectPrevRole:  RoleFollower,
+			expectFinalRole: RoleLeader,
+		},
+	}
```
mszadkow: This test seems to be missing some scenarios, otherwise I don't know why we need so many options

mszadkow:
overall lgtm
please check the PR with manual tests -  it can reveal some issues that can't be catch in integration

pkg/util/roletracker/tracker_test.go
IrvingMg: Done. Added a couple more scenarios.

test/integration/singlecluster/controller/core/clusterqueue_controller_test.go
```diff
@@ -1133,3 +1139,81 @@ var _ = ginkgo.Describe("ClusterQueue controller", ginkgo.Label("controller:clus
...
+	})
+
+	ginkgo.BeforeEach(func() {
+		rf = utiltestingapi.MakeResourceFlavor("ha-transition-flavor").Obj()
+		util.MustCreate(ctx, k8sClient, rf)
+
+		cq = utiltestingapi.MakeClusterQueue("ha-transition-cq").
+			ResourceGroup(*utiltestingapi.MakeFlavorQuotas("ha-transition-flavor").Resource(corev1.ResourceCPU, "10").Obj()).
+			Obj()
+		util.MustCreate(ctx, k8sClient, cq)
```
mbobrovskyi:
```suggestion
		util.CreateClusterQueuesAndWaitForActive(ctx, k8sClient, cq)
```

test/integration/singlecluster/controller/core/clusterqueue_controller_test.go
```diff
@@ -1133,3 +1139,81 @@ var _ = ginkgo.Describe("ClusterQueue controller", ginkgo.Label("controller:clus
...
+
+	ginkgo.It("Should resync metrics after role transition from follower to leader", func() {
+		followerLabels := map[string]string{
+			"cluster_queue": cq.Name,
+			"replica_role":  roletracker.RoleFollower,
+		}
+
+		ginkgo.By("Waiting for CQ to become active")
+		util.ExpectClusterQueuesToBeActive(ctx, k8sClient, cq)
+
```
mbobrovskyi:
```suggestion
```

test/integration/singlecluster/controller/core/clusterqueue_controller_test.go
```diff
@@ -1133,3 +1139,81 @@ var _ = ginkgo.Describe("ClusterQueue controller", ginkgo.Label("controller:clus
...
+var _ = ginkgo.Describe("ClusterQueue controller with RoleTracker", ginkgo.Label("controller:clusterqueue", "area:core"), ginkgo.Ordered, ginkgo.Serial, func() {
+	var (
+		electedChan   chan struct{}
+		trackerCancel context.CancelFunc
+		rf            *kueue.ResourceFlavor
+		cq            *kueue.ClusterQueue
+	)
+
+	ginkgo.BeforeAll(func() {
```
mbobrovskyi:
```suggestion
	ginkgo.BeforeEach(func() {
```
nit: because if we add one more test, electedChan will never become active.

test/integration/singlecluster/controller/core/clusterqueue_controller_test.go
```diff
@@ -1133,3 +1139,81 @@ var _ = ginkgo.Describe("ClusterQueue controller", ginkgo.Label("controller:clus
...
+		electedChan = make(chan struct{})
+		tracker := roletracker.NewRoleTracker(electedChan)
+		var trackerCtx context.Context
+		trackerCtx, trackerCancel = context.WithCancel(ctx)
+		go tracker.Start(trackerCtx, ctrl.Log)
+		fwk.StartManager(ctx, cfg, managerAndControllerSetup(nil, withRoleTracker(tracker)))
+	})
+
+	ginkgo.AfterAll(func() {
+		trackerCancel()
```
mbobrovskyi:
```suggestion
		if _, ok := <-electedChan; ok {
			close(electedChan)
		}
		trackerCancel()
```
Just to make sure that we close channel on failure.

test/integration/singlecluster/controller/core/suite_test.go
```diff
@@ -104,11 +115,20 @@ func managerAndControllerSetup(controllersCfg *config.Configuration, options ...
 
 		controllersCfg.Metrics.EnableClusterQueueResources = true
 
-		cCache := schdcache.New(mgr.GetClient())
-		queues := util.NewManagerForIntegrationTests(ctx, mgr.GetClient(), cCache)
+		cacheOpts := []schdcache.Option{
+			schdcache.WithResourceMetrics(controllersCfg.Metrics.EnableClusterQueueResources),
+		}
+		var queueOpts []qcache.Option
+		if opts.roleTracker != nil {
```
mbobrovskyi: I think we can avoid this if and propagate this options directly.

pkg/controller/core/core.go
```diff
@@ -102,6 +103,15 @@ func SetupControllers(mgr ctrl.Manager, qManager *qcache.Manager, cc *schdcache.
...
 	qManager.AddTopologyUpdateWatcher(cqRec)
 	qManager.AddWorkloadUpdateWatcher(qRec)
+
+	if roleTracker != nil {
+		roleTracker.OnTransition(func(previousRole string) {
+			metrics.ClearGaugeMetricsForRole(previousRole)
+			cc.ResyncMetrics()
+			qManager.ResyncMetrics()
+		})
+	}
```
kshalot: Nit: Should this be inside `SetupControllers` or higher, a little bit closer to where the role tracker is initialized (I guess this would be the `main` func)? `cc`, `qManager` and `roleTracker`  are all passed from main and I don't see any local var being referenced here so this is not a closure either.

pkg/util/roletracker/tracker.go
```diff
@@ -51,12 +52,22 @@ func NewFakeRoleTracker(role string) *RoleTracker {
 	return rt
 }
 
+// OnTransition registers a callback invoked after the role changes,
+// receiving the previous role. Must be called before Start.
+func (rt *RoleTracker) OnTransition(fn func(previousRole string)) {
+	rt.onTransition = fn
+}
```
kshalot: Nit: I'm wondering about the naming, because `OnTransition` is essentially `OnElected` in our case, right? From what I understand, there is no way for this to trigger with `previousRole` being anything other than "follower". And IIUC there is no way for this to be triggered more than once per replica.

I'd consider making the code a little less flexible but more explicit, unless there is some case here that I'm missing. WDYT?

test/integration/singlecluster/controller/core/clusterqueue_controller_test.go
IrvingMg: Done.

test/integration/singlecluster/controller/core/clusterqueue_controller_test.go
IrvingMg: Done.

test/integration/singlecluster/controller/core/clusterqueue_controller_test.go
IrvingMg: Done. Also updated to `AfterEach`.

test/integration/singlecluster/controller/core/clusterqueue_controller_test.go
IrvingMg: Done.

test/integration/singlecluster/controller/core/suite_test.go
IrvingMg: Done. Now always passes both options unconditionally.

pkg/controller/core/core.go
IrvingMg: Done. Moved it to main() right after queues is created.

pkg/util/roletracker/tracker.go
IrvingMg: Good idea. I renamed it to `onElected` and remove the `previousRole` param.

pkg/metrics/metrics.go
```diff
@@ -847,6 +847,18 @@ func ReportCohortWeightedShare(cohort string, weightedShare float64, tracker *ro
...
+
+// ClearGaugeMetricsForRole deletes all gauge metric time series matching
+// replica_role=role. Called during HA role transitions to remove stale
+// time series reported under the old role.
+func ClearGaugeMetricsForRole(role string) {
+	lbls := prometheus.Labels{"replica_role": role}
+	for _, g := range allGaugeVecs {
+		g.DeletePartialMatch(lbls)
+	}
+}
```
kshalot, line 892: If there are >1 followers, won't this remove the gauges for all of them? Or are they implicitly scoped to a replica?

I'm asking as a precaution, because from what I can see it was tested with 1 follower and 1 leader.

IrvingMg: This should run only on the replica that transitions. As I understand it, each replica has its own registry, so clearing them should be scoped to that replica.

kshalot: Yeah, that was my understanding, but wanted to double check

pkg/cache/scheduler/clusterqueue.go
```diff
@@ -492,6 +493,40 @@ func (c *clusterQueue) reportActiveWorkloads() {
...
+		if quota.BorrowingLimit != nil {
+			borrowing = resourceFloat(fr.Resource, *quota.BorrowingLimit)
+		}
+		if quota.LendingLimit != nil {
+			lending = resourceFloat(fr.Resource, *quota.LendingLimit)
+		}
+		metrics.ReportClusterQueueQuotas(cohort, cqName, fName, rName, nominal, borrowing, lending, c.roleTracker)
+		metrics.ReportClusterQueueResourceReservations(cohort, cqName, fName, rName, resourceFloat(fr.Resource, c.resourceNode.Usage[fr]), c.roleTracker)
+		metrics.ReportClusterQueueResourceUsage(cohort, cqName, fName, rName, resourceFloat(fr.Resource, c.AdmittedUsage[fr]), c.roleTracker)
+	}
```
kshalot, line 516: Unless I'm missing something, I see some overlap between this and existing code, like this:
https://github.com/kubernetes-sigs/kueue/blob/307e822b65f945b7f0a963d25ccf3d41c9571482/pkg/controller/core/clusterqueue_controller.go#L381-L411

Or this for LQ:
https://github.com/kubernetes-sigs/kueue/blob/307e822b65f945b7f0a963d25ccf3d41c9571482/pkg/controller/core/localqueue_controller.go#L374-L385

I suspect there are some differences in how to handle a total re-sync and updates like in those snippets, but maybe there is some way to re-use the logic. Or at least maybe gather it in some place? WDYT?

pkg/cache/scheduler/cache.go
```diff
@@ -905,6 +914,38 @@ func (c *Cache) MatchingClusterQueues(nsLabels map[string]string) sets.Set[kueue
...
+		cq.reportActiveWorkloads()
+		if c.resourceMetricsEnabled {
+			cq.reportResourceMetrics(c.fairSharingEnabled)
+		}
+		if features.Enabled(features.LocalQueueMetrics) {
+			for _, lq := range cq.localQueues {
+				lq.reportActiveWorkloads(c.roleTracker)
+				if c.resourceMetricsEnabled {
+					lq.reportResourceMetrics(cq.resourceNode.Quotas, c.roleTracker)
+				}
```
kshalot: To illustrate my point from [this comment](https://github.com/kubernetes-sigs/kueue/pull/9487/files#r2904826910) - here we will only report LQ resource metrics if `c.resourceMetricsEnabled` (which is `controllersCfg.Metrics.EnableClusterQueueResources`), but I don't think this check is performed in the other LQ metrics logic, which seems to only check `features.Enabled(features.LocalQueueMetrics)`:
https://github.com/kubernetes-sigs/kueue/blob/307e822b65f945b7f0a963d25ccf3d41c9571482/pkg/controller/core/localqueue_controller.go#L228-L230

So, unless I'm missing something, there logic diverges here.

pkg/cache/scheduler/clusterqueue.go
IrvingMg: Yeah, the common part is that they call the same `metrics.Report*` functions. However, the rest differs because they read from different data sources: the cache uses maps, while the controller uses API objects. So I don’t think they can be commonized much, or that there would be much benefit in doing so.

pkg/cache/scheduler/cache.go
IrvingMg: Good catch. Removed the `c.resourceMetricsEnabled` check, LQ resource metrics should be gated only by the `LocalQueueMetrics` feature gate.

pkg/cache/scheduler/clusterqueue.go
mimowo: I agree with @kshalot . I believe this is the same motivation and sentiment as in my comment here: https://github.com/kubernetes-sigs/kueue/pull/9487#discussion_r2865072597

Sure, the code has similar structure in both places, but I cannot imagine a reviewer would notice if someone forgot to adjust the MetricsResync when introducing a new metric. A contributor may notice by following the links (function references). The reviewer is almost certainly lost, unless she/he or checkout the code.

I agree that currently both functions use different data sources, but could we replace the existing occurrences with the new ones: https://github.com/kubernetes-sigs/kueue/blob/307e822b65f945b7f0a963d25ccf3d41c9571482/pkg/controller/core/clusterqueue_controller.go#L381-L411

I believe the correct approach is for both places to use the cache as the data source. This is something I discussed in the past with @mbobrovskyi on the occassion of a metric for finished workloads, was that our current model of building metrics based on statuses is too constraining. I prefer the status and the metrics to both have the source in cache.

wdyt @IrvingMg . If we agree it does not necessarily mean it needs to be this PR, but we could open a follow up issue to commonize the approach.

pkg/cache/queue/manager.go
```diff
@@ -855,14 +834,29 @@ func (m *Manager) queueSecondPass(ctx context.Context, w *kueue.Workload, iterat
...
-	wInfo.UpdateSchedulingHash(log)
 	wInfo.SecondPassIteration = iteration
 	if m.secondPassQueue.queue(wInfo) {
 		log.V(3).Info("Workload queued for second pass of scheduling", "workload", workload.Key(w))
 		m.Broadcast()
 	}
 }
 
+// ResyncMetrics re-reports pending and finished workload gauge metrics.
+func (m *Manager) ResyncMetrics() {
```
mimowo:
```suggestion
func (m *Manager) ResyncGuageMetrics() {
```
nit, I think this is by design only related to guage metrics, right?

pkg/cache/scheduler/cache.go
```diff
@@ -905,6 +914,36 @@ func (c *Cache) MatchingClusterQueues(nsLabels map[string]string) sets.Set[kueue
 	return cqs
 }
 
+// ResyncMetrics re-reports CQ/LQ status, active workload, resource, and weighted share gauge metrics.
+func (c *Cache) ResyncMetrics() {
```
mimowo:
```suggestion
func (c *Cache) ResyncGuageMetrics() {
```

cmd/kueue/main.go
```diff
@@ -334,6 +335,14 @@ func main() {
 		os.Exit(1)
 	}
 
+	if roleTracker != nil {
+		roleTracker.OnElected(func() {
+			metrics.ClearGaugeMetricsForRole(roletracker.RoleFollower)
+			cCache.ResyncMetrics()
+			queues.ResyncMetrics()
```
mimowo:
```suggestion
			cCache.ResyncGuageMetrics()
			queues.ResyncGuageMetrics()
```

pkg/cache/scheduler/clusterqueue.go
IrvingMg: Okay, I've created an issue for the follow-up: https://github.com/kubernetes-sigs/kueue/issues/9793

pkg/cache/queue/manager.go
IrvingMg: Right. Renamed.

pkg/cache/scheduler/cache.go
IrvingMg: Done.

cmd/kueue/main.go
IrvingMg: Done.

mimowo:
@IrvingMg please prepare cherrypicks manually in case of conflicts

PR #9493, opened 2026-02-25
Increase admission_wait_time_seconds histogram max bucket from ~2.84h to ~11h

mukund-wayve:
#### What type of PR is this?

/kind feature

image

As shown in the image, several ClusterQueues have admission_wait_time_seconds quantiles plateauing at ~2.84 hours, the highest finite bucket boundary (10,240s). Beyond this, histogram_quantile() cannot resolve actual wait times.

This PR adds one additional bucket at 40,960 (~11.3h) to the AdmissionWaitTime histogram, improving observability for clusters with multi-hour queue times.

Since each histogram bucket creates a separate time series per unique label combination, the change is intentionally scoped to a single metric and a 2 additional buckets to minimize the cardinality impact on Prometheus memory, storage, and scrape size.

Fixes #

```release-note
Observability: Increased the maximum finite bucket boundary for admission_wait_time_seconds histogram from ~2.84 hours to ~11.3 hours for better observability of long queue times.
```

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -275,7 +275,7 @@ The label 'underlying_cause' can have the following values:
 			Subsystem: constants.KueueName,
 			Name:      "admission_wait_time_seconds",
 			Help:      "The time between a workload was created or requeued until admission, per 'cluster_queue'",
-			Buckets:   generateExponentialBuckets(15),
+			Buckets:   generateExponentialBuckets(16),
 		}, []string{"cluster_queue", "priority_class", "replica_role"},
 	)
 


pkg/metrics/metrics.go
```diff
@@ -275,7 +275,7 @@ The label 'underlying_cause' can have the following values:
 			Subsystem: constants.KueueName,
 			Name:      "admission_wait_time_seconds",
 			Help:      "The time between a workload was created or requeued until admission, per 'cluster_queue'",
-			Buckets:   generateExponentialBuckets(14),
+			Buckets:   generateExponentialBuckets(15),
```
tenzen-y: @mukund-wayve I can accept this change, but I'd like to confirm if ~5.69h is enough or not.

Because this is less than business hours (9 hours). I can imagine that researchers submit Jobs in the morning, and the Jobs occipied all cluster queues, then the morning jobs finish around closing time. After that, the Jobs queued in the morning are admitted.

In this scenario, we might want to increase this bucket to `16` (~11.4 hours).

Additionally, I'd like to check whether you find this bucket-size problem in the other metrics as well.

tenzen-y: @mimowo, please let us know if expanding the bucket size is acceptable.
Indeed, this change might lead to Prometheus performance regressions in the very large cluster.

Additionally, I'm wondering if we can say this as "bug", then cherry-pick this to release branches.

tenzen-y (commented):
@mukund-wayve Thank you for reporting the real cluster problem!

pkg/metrics/metrics.go
mimowo: I'm ok with the change, I think the potential for causing a perf issue is really minimal. 

wrt cherrypicking I'm also ok if this makes life easier for the reporting user.

tenzen-y: > I'm ok with the change, I think the potential for causing a perf issue is really minimal.
> 
> wrt cherrypicking I'm also ok if this makes life easier for the reporting user.

That sounds good to me as well. Let me wait for @mukund-wayve answer for my question.

mukund-wayve: Thanks for the responses!

I can definitely see cases where admission wait times could reach 11 hours or more -- particularly during heavy backfills, or when a normally small project has a sudden spike in demand. That said, I agree we need a reasonable threshold, and ~11 hours seems like a good balance. We can always revisit if other users report similar needs.

Regarding the other histogram metrics, we don't actively use most of them and haven't encountered a need to bump their buckets. I reviewed all 12 histogram metrics in the codebase, and under normal operations with properly configured quotas, none of them should approach the current ~2.84-hour ceiling. So I'd prefer to keep this PR scoped to admission_wait_time_seconds and bump it to 16 buckets (~11 hours) as suggested.

And thanks for the suggestion to cherry-pick -- I'd definitely appreciate that. I'll update the PR label from feature to bug.

mimowo:
/release-note-edit
```release-note
Observability: Increased the maximum finite bucket boundary for admission_wait_time_seconds histogram from ~2.84 hours to ~11.3 hours for better observability of long queue times.
```

mimowo:
Thank you for the contribution and the additional investigations 👍 
/lgtm
/approve
/cherrypick release-0.16
/cherrypick release-0.15

pkg/metrics/metrics.go
mimowo: I think we can keep feature, it is ok to cherrypick a small feature I think. This sounds more intuitive categorization in this case.

tenzen-y: > Thanks for the responses!
> 
> I can definitely see cases where admission wait times could reach 11 hours or more -- particularly during heavy backfills, or when a normally small project has a sudden spike in demand. That said, I agree we need a reasonable threshold, and ~11 hours seems like a good balance. We can always revisit if other users report similar needs.
> 
> Regarding the other histogram metrics, we don't actively use most of them and haven't encountered a need to bump their buckets. I reviewed all 12 histogram metrics in the codebase, and under normal operations with properly configured quotas, none of them should approach the current ~2.84-hour ceiling. So I'd prefer to keep this PR scoped to admission_wait_time_seconds and bump it to 16 buckets (~11 hours) as suggested.
> 
> And thanks for the suggestion to cherry-pick -- I'd definitely appreciate that. I'll update the PR label from feature to bug.
> 

Thank you for answering to my questions. Those fully sound good to me.
Let's revisit here once we face the bucket size problem again, either for admission_wait_time_seconds or other metrics.

mimowo:
@mukund-wayve please submit the cherrypick PR for 0.15 using the ./hack/cherry_pick_pull.sh script

PR #9494, opened 2026-02-25
[Fair Sharing] If workload fits in nominalQuota for a flavor, ignore DRS checks during preemption

### LINKED ISSUE #9406: Fair sharing admission ordering does not preserve nominal-first workload priority

mukund-wayve:
#### What type of PR is this?

/kind bug

In a multi-flavor setup with fair sharing enabled, borrowing resources on one flavor can prevent a ClusterQueue from using its nominal quota on a different flavor — both in admission ordering and preemption. The issue is described in detail in #9406.

This PR addresses the preemption case. When a CQ is within nominal quota for the contested flavor, it should be able to preempt workloads even if its aggregate DRS is inflated by borrowing on other flavors. Without this fix, the DRS strategy gates block the preemption despite the preemptor having a legitimate nominal entitlement.

Changes:

1. Extract preemptCandidate helper to avoid duplicating 4 lines across 3 functions.
2. In runFirstFsStrategy, check if the preemptor is within nominal quota for contested resources. If so, bypass the DRS strategy gate and preempt with InCohortReclamationReason. Several existing test cases updated accordingly, as they were actually reclaiming nominal quota.
3. Add FairSharingPreemptWithinNominal feature gate (Beta, enabled by default).

Part-of #9406 

This is part 2/n of the split version of #9407  which attempts to resolve #9406 

```release-note
In fair sharing preemption, bypass DRS strategy gates when the preemptor ClusterQueue is within nominal quota for contested resources, allowing preemption even if the CQ's aggregate DRS is high due to borrowing on other flavors.
```

### DIFF
--- a/pkg/scheduler/preemption/preemption.go
+++ b/pkg/scheduler/preemption/preemption.go
@@ -317,6 +317,9 @@ func parseStrategies(fs *config.FairSharing) []fairsharing.Strategy {
 func runFirstFsStrategy(preemptionCtx *preemptionCtx, candidates []*workload.Info, strategy fairsharing.Strategy) (bool, []*Target, []*workload.Info) {
 	ordering := fairsharing.MakeClusterQueueOrdering(preemptionCtx.preemptorCQ, candidates, preemptionCtx.log, preemptionCtx.clock)
 
+	var targets []*Target
+	var retryCandidates []*workload.Info
+
 	// If the preemptor CQ stays within nominal quota for the contested
 	// resources (including the incoming workload, already simulated),
 	// preemption is allowed regardless of DRS (nominal entitlement).
@@ -325,9 +328,6 @@ func runFirstFsStrategy(preemptionCtx *preemptionCtx, candidates []*workload.Inf
 	// and runSecondFsStrategy has nothing to do.
 	preemptorWithinNominal := features.Enabled(features.FairSharingPreemptWithinNominal) &&
 		queueWithinNominalInResourcesNeedingPreemption(preemptionCtx)
-
-	var targets []*Target
-	var retryCandidates []*workload.Info
 	for candCQ := range ordering.Iter() {
 		if candCQ.InClusterQueuePreemption() {
 			candWl := candCQ.PopWorkload()
@@ -569,7 +569,7 @@ func queueUnderNominalInResourcesNeedingPreemption(preemptionCtx *preemptionCtx)
 // that this treats usage exactly equal to nominal as "within nominal."
 func queueWithinNominalInResourcesNeedingPreemption(preemptionCtx *preemptionCtx) bool {
 	for fr := range preemptionCtx.frsNeedPreemption {
-		if preemptionCtx.preemptorCQ.ResourceNode.Usage[fr] > preemptionCtx.preemptorCQ.QuotaFor(fr).Nominal {
+		if preemptionCtx.preemptorCQ.Borrowing(fr) {
 			return false
 		}
 	}


pkg/scheduler/preemption/preemption.go
```diff
@@ -311,24 +311,49 @@ func parseStrategies(fs *config.FairSharing) []fairsharing.Strategy {
 	return strategies
 }
 
+// preemptCandidate removes a workload from the snapshot and records it
+// as a preemption target. Returns the updated targets slice and whether
+// the preempting workload now fits.
+func preemptCandidate(preemptionCtx *preemptionCtx, candWl *workload.Info, candCQ *fairsharing.TargetClusterQueue, reason string, targets []*Target) ([]*Target, bool) {
```
mukund-wayve: Happy to revert this change and inline the logic if you'd prefer to keep this PR strictly behavioral. Can follow up with the refactor separately

mukund-wayve: Done. Reverted the refactor

pkg/scheduler/preemption/preemption.go
```diff
@@ -538,6 +562,20 @@ func queueUnderNominalInResourcesNeedingPreemption(preemptionCtx *preemptionCtx)
...
+// queueWithinNominalInResourcesNeedingPreemption checks whether the
+// preemptor CQ's usage is at or below nominal quota (usage <= nominal)
+// for all flavor-resources needing preemption.
+// The difference from queueUnderNominalInResourcesNeedingPreemption is
+// that this treats usage exactly equal to nominal as "within nominal."
+func queueWithinNominalInResourcesNeedingPreemption(preemptionCtx *preemptionCtx) bool {
+	for fr := range preemptionCtx.frsNeedPreemption {
+		if preemptionCtx.preemptorCQ.ResourceNode.Usage[fr] > preemptionCtx.preemptorCQ.QuotaFor(fr).Nominal {
+			return false
+		}
```
gabesaba: does `ClusterQueueSnapshot.Borrowing` work here? If not, please define a new function in that package

pkg/scheduler/preemption/preemption.go
```diff
@@ -317,6 +317,15 @@ func parseStrategies(fs *config.FairSharing) []fairsharing.Strategy {
...
 
+	// If the preemptor CQ stays within nominal quota for the contested
+	// resources (including the incoming workload, already simulated),
+	// preemption is allowed regardless of DRS (nominal entitlement).
+	// When true, all cross-CQ candidates are preempted unconditionally
+	// (bypassing the strategy check), so no retryCandidates are produced
+	// and runSecondFsStrategy has nothing to do.
+	preemptorWithinNominal := features.Enabled(features.FairSharingPreemptWithinNominal) &&
+		queueWithinNominalInResourcesNeedingPreemption(preemptionCtx)
+
```
gabesaba: nit: can you define this closer to its usage?

mukund-wayve: This needs to be outside the for-loop since this doesn't change for the candidates. But I've moved this just before the for-loop

pkg/scheduler/preemption/preemption.go
mukund-wayve: You're right. `ClusterQueueSnapshot.Borrowing` works here and I've used it. Thanks

mimowo:
@mukund-wayve please remember about the cherry-picks

PR #9506, opened 2026-02-26
feat: add multi-layer topology constraints for TAS

### LINKED ISSUE #9046: TAS: Support Multi-Layer Topology Constraints
## What would you like to be added

Support training workloads with **multi-layer topology constraints** (e.g., GB200/GB300 clusters).

[kubernetes-sigs/kueue#6554 (comment)](https://github.com/kubernetes-sigs/kueue/issues/6554#issuecomment-3179837495) cleanly illustrates the need.
Real-world data center topologies often have multiple physical levels (e.g., data center → building → row/NVLink domain/block → rack → host); to get the best performance, workloads increasingly need to express constraints at more than two of those levels simultaneously.

## Current situation and gap

TAS balanced placement ([#6851](https://github.com/kubernetes-sigs/kueue/pull/6851), v0.15) is a best-effort primitive that balances pod placement on a specified topology. However, in real-world GB200 clusters we need to **guarantee** that a training Job's Pods are co-located into NVLink domain / rack / other topology levels with specific, layered criteria.

Even if we changed semantics of TAS balanced placement to guaranteed, it still lacks the flexibility to express **multi-layer topology constraints**. For example, a 96-replica Job may require:

- all 96 Pods land in the same datacenter
- groups of 48 Pods land within the same virtual zone (so it doesn't end with landing on 3 zones, with each having 32 Pods)
- groups of 16 Pods land within the same rack
- (or even finer granularity)

## Closest existing primitive

The closest mechanism is [KEP-2724: Two-level Topology Aware Scheduling](https://github.com/kubernetes-sigs/kueue/pull/5449) (#5449 / #5596), which introduced the `podset-slice-required-topology` / `podset-slice-size` annotations. Although designed primarily for JobSet, the mechanism is generic enough to support arbitrary PodSet-based workloads.

However, it is limited to exactly **2 constraint tiers**: one for the entire PodSet and one for slices — close to what we need, but not sufficient.

## Proposed change

Extend TAS to support **multi-layer topology constraints** - up to N total constraint levels (1 podset-level + N−1 slice layers). Users specify **additional slice layers** that recursively subdivide slices into smaller groups, each constrained to a finer topology level.

**Example: 4-layer constraints on 96 pods:**

```yaml
# Existing "kueue.x-k8s.io/podset-slice-*" annotations are kept as-is for backward compatibility
annotations:
  # Layer 0 (podset): all 96 pods in the same datacenter
  kueue.x-k8s.io/podset-required-topology: "cloud.provider.com/datacenter"
  # Layer 1 (existing slice): groups of 48 in the same logical zone
  kueue.x-k8s.io/podset-slice-required-topology: "cloud.provider.com/logical-zone"
  kueue.x-k8s.io/podset-slice-size: "48"
  # Layer 2 (new): groups of 16 on the same rack
  kueue.x-k8s.io/podset-slice-required-topology-1: "cloud.provider.com/rack"
  kueue.x-k8s.io/podset-slice-size-1: "16"
  # Layer 3 (new): groups of 2 on the same host (optional: depends on GB200 setup: NVL36 vs. NVL72)
  # kueue.x-k8s.io/podset-slice-required-topology-2: "kubernetes.io/hostname"
  # kueue.x-k8s.io/podset-slice-size-2: "2"
```

Result: 96 pods → 2 logical zones → 6 racks of 16. So it ends up w/ `{3 racks, 3 racks}` which is a symmetric distribution instead of `{4 racks, 2 racks}`.

There are other variants of needing multiple layers, but not restricted to the following:

- As time goes, the cluster gets fragmented, and w/o (or we cannot) active de-fragmentation, we need more layers to achieve optimal performance upon cluster's fragmentation
- For an XL training Job, it may need 1000+ replicas and would need more layers to support it
- ...

> [!IMPORTANT]
> The concrete number of `N` can be discussed. We may start with a conservative cap of **2 additional layers** (3 slice layers total).

```diff
diff --git a/apis/kueue/v1beta2/workload_types.go b/apis/kueue/v1beta2/workload_types.go
index ce630751e..79a0d56b8 100644
--- a/apis/kueue/v1beta2/workload_types.go
+++ b/apis/kueue/v1beta2/workload_types.go
@@ -223,6 +223,33 @@ type PodSetTopologyRequest struct {
 	//
 	// +optional
 	PodSetSliceSize *int32 `json:"podSetSliceSize,omitempty"`
+
+	// additionalSliceLayers defines additional layers of recursive slice
+	// subdivision beyond the first slice layer (podSetSliceRequiredTopology /
+	// podSetSliceSize). Each layer further subdivides the parent layer's
+	// groups into smaller groups constrained to a finer topology domain.
+	// At most 2 additional layers are supported (for a total of 3 slice layers).
+	//
+	// +optional
+	// +listType=atomic
+	// +kubebuilder:validation:MaxItems=2
+	AdditionalSliceLayers []SliceLayer `json:"additionalSliceLayers,omitempty"`
+}
+
+// SliceLayer defines a single additional slice subdivision layer.
+type SliceLayer struct {
+	// topology indicates the topology level required for this slice layer.
+	//
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=63
+	Topology string `json:"topology"`
+
+	// size indicates the number of pods in each group at this slice layer.
+	//
+	// +required
+	// +kubebuilder:validation:Minimum=1
+	Size int32 `json:"size"`
 }

 type Admission struct {
```

### 1. Real data center topologies are deeper than 2 levels

Modern GPU clusters have hierarchies like **data center → building -> virtual zone -> NVLink Doamin/block/row → rack →  → host**. AI/ML training workloads need different communication granularities at each level: all-reduce within a host, ring-reduce within a rack, hierarchical all-reduce across racks within a block. Being limited to 2 constraint tiers means users can only optimize for 2 of these boundaries, leaving performance on the table.

### 2. Natural extension of existing design

The slice mechanism already supports recursive subdivision conceptually - a slice is just a fixed-size group constrained to a topology domain. Multi-layer slicing is "slices within slices," following the same pattern. The existing TAS algorithm's 2-phase traversal (greedy above slice level, parent-constrained at/below) generalizes naturally to N phases.

### 3. Avoids workarounds

Without this, users must work around the limitation by:

- Manually splitting workloads into smaller Jobs and coordinating placement externally
- Using multiple JobSets with separate topology constraints, losing the "all in one topology" guarantee
- Relying on the Kubernetes scheduler's topology spread constraints, which lack TAS's capacity-aware admission

## Completion requirements

- [ ] Design doc
- [ ] API change
- [ ] Docs update

Huang-Wei:
#### What type of PR is this?

/kind feature

Introduce the TASMultiLayerTopology feature gate (alpha) to support
up to 3 hierarchical slice layers in topology-aware scheduling.

- Add `PodsetSliceRequiredTopologyConstraints` field to
  `PodSetTopologyRequest` and the corresponding pod annotation
- Extend TAS flavor snapshot to handle multi-layer placement
- Add validation for the new field (mutual exclusivity with
  existing single-layer fields, layer ordering, size divisibility)
- Add v1beta1 <-> v1beta2 conversion support
- Add integration tests for multi-layer scheduling

Fixes #9046

> [!NOTE]
> This PR aims to impose minimum changes to the existing 2-layer annotation and impl.,
> so enabling or disabling this feature gate will have a minimal impact.

```release-note
TAS: Introduce the `TASMultiLayerTopology` feature gate (alpha) to support up to 3 hierarchical slice layers
```

### DIFF
--- a/pkg/cache/scheduler/tas_cache_test.go
+++ b/pkg/cache/scheduler/tas_cache_test.go
@@ -5630,6 +5630,118 @@ func TestFindTopologyAssignments(t *testing.T) {
 				},
 			}},
 		},
+		"multi-layer topology: host slice rounding makes rack slice impossible": {
+			// block → rack → hostname
+			//        b1
+			//        |
+			//       r1
+			//      /    \
+			//   x1(3)  x2(3)
+			enableFeatureGates: []featuregate.Feature{features.TASMultiLayerTopology},
+			nodes: []corev1.Node{
+				*testingnode.MakeNode("b1-r1-x1").
+					Label(tasBlockLabel, "b1").
+					Label(tasRackLabel, "r1").
+					Label(corev1.LabelHostname, "x1").
+					StatusAllocatable(corev1.ResourceList{
+						corev1.ResourceCPU:  resource.MustParse("3"),
+						corev1.ResourcePods: resource.MustParse("10"),
+					}).
+					Ready().
+					Obj(),
+				*testingnode.MakeNode("b1-r1-x2").
+					Label(tasBlockLabel, "b1").
+					Label(tasRackLabel, "r1").
+					Label(corev1.LabelHostname, "x2").
+					StatusAllocatable(corev1.ResourceList{
+						corev1.ResourceCPU:  resource.MustParse("3"),
+						corev1.ResourcePods: resource.MustParse("10"),
+					}).
+					Ready().
+					Obj(),
+			},
+			levels: defaultThreeLevels,
+			podSets: []PodSetTestCase{{
+				topologyRequest: &kueue.PodSetTopologyRequest{
+					Required: ptr.To(tasBlockLabel),
+					PodsetSliceRequiredTopologyConstraints: []kueue.PodsetSliceRequiredTopologyConstraint{
+						{Topology: tasRackLabel, Size: 6},
+						{Topology: corev1.LabelHostname, Size: 2},
+					},
+				},
+				requests: resources.Requests{
+					corev1.ResourceCPU: 1000,
+				},
+				count:      6,
+				wantReason: `topology "default" doesn't allow to fit any of 1 slice(s)`,
+			}},
+		},
+		"multi-layer topology: small host kills rack slices despite enough total capacity": {
+			// block → rack → hostname
+			//              b1
+			//       /             \
+			//      r1             r2
+			//   /      \        /    \
+			//  x1(7)  x2(3)  x3(7)  x4(3)
+			enableFeatureGates: []featuregate.Feature{features.TASMultiLayerTopology},
+			nodes: []corev1.Node{
+				*testingnode.MakeNode("b1-r1-x1").
+					Label(tasBlockLabel, "b1").
+					Label(tasRackLabel, "r1").
+					Label(corev1.LabelHostname, "x1").
+					StatusAllocatable(corev1.ResourceList{
+						corev1.ResourceCPU:  resource.MustParse("7"),
+						corev1.ResourcePods: resource.MustParse("10"),
+					}).
+					Ready().
+					Obj(),
+				*testingnode.MakeNode("b1-r1-x2").
+					Label(tasBlockLabel, "b1").
+					Label(tasRackLabel, "r1").
+					Label(corev1.LabelHostname, "x2").
+					StatusAllocatable(corev1.ResourceList{
+						corev1.ResourceCPU:  resource.MustParse("3"),
+						corev1.ResourcePods: resource.MustParse("10"),
+					}).
+					Ready().
+					Obj(),
+				*testingnode.MakeNode("b1-r2-x3").
+					Label(tasBlockLabel, "b1").
+					Label(tasRackLabel, "r2").
+					Label(corev1.LabelHostname, "x3").
+					StatusAllocatable(corev1.ResourceList{
+						corev1.ResourceCPU:  resource.MustParse("7"),
+						corev1.ResourcePods: resource.MustParse("10"),
+					}).
+					Ready().
+					Obj(),
+				*testingnode.MakeNode("b1-r2-x4").
+					Label(tasBlockLabel, "b1").
+					Label(tasRackLabel, "r2").
+					Label(corev1.LabelHostname, "x4").
+					StatusAllocatable(corev1.ResourceList{
+						corev1.ResourceCPU:  resource.MustParse("3"),
+						corev1.ResourcePods: resource.MustParse("10"),
+					}).
+					Ready().
+					Obj(),
+			},
+			levels: defaultThreeLevels,
+			podSets: []PodSetTestCase{{
+				topologyRequest: &kueue.PodSetTopologyRequest{
+					Required: ptr.To(tasBlockLabel),
+					PodsetSliceRequiredTopologyConstraints: []kueue.PodsetSliceRequiredTopologyConstraint{
+						{Topology: tasRackLabel, Size: 8},
+						{Topology: corev1.LabelHostname, Size: 4},
+					},
+				},
+				requests: resources.Requests{
+					corev1.ResourceCPU: 1000,
+				},
+				count:      16,
+				wantReason: `topology "default" doesn't allow to fit any of 2 slice(s)`,
+			}},
+		},
 	}
 	for name, tc := range cases {
 		t.Run(name, func(t *testing.T) {
--- a/pkg/cache/scheduler/tas_flavor_snapshot.go
+++ b/pkg/cache/scheduler/tas_flavor_snapshot.go
@@ -778,44 +778,10 @@ func (s *TASFlavorSnapshot) findTopologyAssignment(
 		return nil, fmt.Sprintf("podset slice topology %s is above the podset topology %s", sliceTopologyKey, *topologyKey)
 	}
 
-	// Build the map of slice sizes per level for multi-layer support.
-	// sliceSizeAtLevel map is keyed with topology level index and valued with the slice size that
-	// applies when distributing pods into domains at that level.
-	// All levels between the outermost slice level and the innermost additional layer
-	// must distribute in multiples of the inner layer's size to ensure
-	// clean grouping at the leaf level.
-	sliceSizeAtLevel := make(map[int]int32)
-	if features.Enabled(features.TASMultiLayerTopology) && workersTasPodSetRequests.PodSet.TopologyRequest != nil {
-		prevSize := sliceSize
-		prevLevelIdx := sliceLevelIdx
-		// TODO: once TASMultiLayerTopology graduates to beta, use "PodsetSliceRequiredTopologyConstraints" as
-		// first-class citizen to generalize the logic of handling N-layer topology constraint.
-		minusOneLayers := workersTasPodSetRequests.PodSet.TopologyRequest.PodsetSliceRequiredTopologyConstraints
-		if len(minusOneLayers) > 1 {
-			minusOneLayers = minusOneLayers[1:] // skip the first (outermost) layer, already handled above
-		} else {
-			minusOneLayers = nil
-		}
-		for _, layer := range minusOneLayers {
-			innerLevelIdx, innerFound := s.resolveLevelIdx(layer.Topology)
-			if !innerFound {
-				return nil, fmt.Sprintf("no requested topology level for additional slice layer: %s", layer.Topology)
-			}
-			if innerLevelIdx <= prevLevelIdx {
-				return nil, fmt.Sprintf("additional slice layer topology %s must be at a lower level than %s", layer.Topology, s.levelKeys[prevLevelIdx])
-			}
-			if prevSize%layer.Size != 0 {
-				return nil, fmt.Sprintf("additional slice layer size %d must evenly divide parent layer size %d", layer.Size, prevSize)
-			}
-			// Fill all levels from prevLevelIdx+1 through innerLevelIdx
-			// so that intermediate levels also distribute in multiples
-			// of this layer's size.
-			for lvl := prevLevelIdx + 1; lvl <= innerLevelIdx; lvl++ {
-				sliceSizeAtLevel[lvl] = layer.Size
-			}
-			prevSize = layer.Size
-			prevLevelIdx = innerLevelIdx
-		}
+	// for feature TASMultiLayerTopology
+	sliceSizeAtLevel, reason := s.buildSliceSizeAtLevel(workersTasPodSetRequests, sliceSize, sliceLevelIdx)
+	if len(reason) > 0 {
+		return nil, reason
 	}
 
 	var selector labels.Selector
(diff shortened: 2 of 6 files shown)


Huang-Wei:
cc @mimowo @tenzen-y @kannon92 PTAL when you get a chance, thanks!

mimowo:
> cc @mimowo @tenzen-y @kannon92 PTAL when you get a chance, thanks!

ack, I will try to get here tomorrow

pkg/controller/jobframework/tas_test.go
```diff
@@ -28,16 +28,18 @@ import (
...
 
 func TestPodSetTopologyRequestBuilder(t *testing.T) {
 	testCases := map[string]struct {
-		meta               *metav1.ObjectMeta
-		podIndexLabel      *string
-		subGroupIndexLabel *string
-		subGroupCount      *int32
-		wantReq            *kueue.PodSetTopologyRequest
-		wantErr            error
+		enableTASMultiLayer bool
```
mimowo: nit, maybe we could introduce generic `featureGates: map[featuretGate]: bool`, we do it already here: https://github.com/kubernetes-sigs/kueue/blob/5d9e98e23f51d283da2a8f30571ff48bc391b1f1/pkg/controller/jobframework/validation_test.go#L261 
and it is very convenient

pkg/controller/jobframework/tas_test.go
```diff
@@ -192,9 +194,77 @@ func TestPodSetTopologyRequestBuilder(t *testing.T) {
 			},
 			wantErr: strconv.ErrSyntax,
 		},
+		"multi-layer: three constraint layers with feature gate": {
```
mimowo, line 222: Let's also have a test case with 0 levels and 4 levels. I think having 0 levels doesn't make sense for the annotation too

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -777,6 +778,46 @@ func (s *TASFlavorSnapshot) findTopologyAssignment(
...
 	}
 
+	// Build the map of slice sizes per level for multi-layer support.
+	// sliceSizeAtLevel map is keyed with topology level index and valued with the slice size that
+	// applies when distributing pods into domains at that level.
+	// All levels between the outermost slice level and the innermost additional layer
+	// must distribute in multiples of the inner layer's size to ensure
+	// clean grouping at the leaf level.
+	sliceSizeAtLevel := make(map[int]int32)
+	if features.Enabled(features.TASMultiLayerTopology) && workersTasPodSetRequests.PodSet.TopologyRequest != nil {
```
mimowo: Extract building the map to a helper function, to offload the already big main function. The comment could follow the function.

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -777,6 +778,46 @@ func (s *TASFlavorSnapshot) findTopologyAssignment(
...
+		prevSize := sliceSize
+		prevLevelIdx := sliceLevelIdx
+		// TODO: once TASMultiLayerTopology graduates to beta, use "PodsetSliceRequiredTopologyConstraints" as
+		// first-class citizen to generalize the logic of handling N-layer topology constraint.
```
mimowo: What are actually code changes this entails? It would be good to clarify.

pkg/controller/jobframework/tas_validation.go
```diff
@@ -271,3 +291,88 @@ func topologyRequestsValid(r1, r2 *kueue.PodSetTopologyRequest) bool {
...
+	if err := json.Unmarshal([]byte(constraintsJSON), &constraints); err != nil {
+		allErrs = append(allErrs, field.Invalid(fldPath, constraintsJSON, fmt.Sprintf("must be a valid JSON array: %v", err)))
+		return allErrs
+	}
+
+	if len(constraints) == 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath, constraintsJSON, "must contain at least 1 entry"))
+		return allErrs
+	}
+	if len(constraints) > 3 {
```
mimowo: nit, but let's define a named const for that. If only used in this function, then we could scope it to the function. Still, the named const will be more readiable.

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -777,6 +778,46 @@ func (s *TASFlavorSnapshot) findTopologyAssignment(
 		return nil, fmt.Sprintf("podset slice topology %s is above the podset topology %s", sliceTopologyKey, *topologyKey)
 	}
 
+	// Build the map of slice sizes per level for multi-layer support.
+	// sliceSizeAtLevel map is keyed with topology level index and valued with the slice size that
+	// applies when distributing pods into domains at that level.
+	// All levels between the outermost slice level and the innermost additional layer
+	// must distribute in multiples of the inner layer's size to ensure
+	// clean grouping at the leaf level.
+	sliceSizeAtLevel := make(map[int]int32)
```
mimowo: Please check with some unit test coverage tool for golang if the new unit test cover 100% of the new code

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -777,6 +778,46 @@ func (s *TASFlavorSnapshot) findTopologyAssignment(
...
+	if features.Enabled(features.TASMultiLayerTopology) && workersTasPodSetRequests.PodSet.TopologyRequest != nil {
+		prevSize := sliceSize
+		prevLevelIdx := sliceLevelIdx
+		// TODO: once TASMultiLayerTopology graduates to beta, use "PodsetSliceRequiredTopologyConstraints" as
+		// first-class citizen to generalize the logic of handling N-layer topology constraint.
+		minusOneLayers := workersTasPodSetRequests.PodSet.TopologyRequest.PodsetSliceRequiredTopologyConstraints
+		if len(minusOneLayers) > 1 {
+			minusOneLayers = minusOneLayers[1:] // skip the first (outermost) layer, already handled above
+		} else {
+			minusOneLayers = nil
```
mimowo: nit, is this needed? If the slice is empty than already the iteration below will skip

pkg/cache/scheduler/tas_cache_test.go
```diff
@@ -5398,6 +5402,234 @@ func TestFindTopologyAssignments(t *testing.T) {
 			}},
 			enableFeatureGates: []featuregate.Feature{features.ElasticJobsViaWorkloadSlices, features.ElasticJobsViaWorkloadSlicesWithTAS},
 		},
+		"multi-layer topology: block required; rack slices of 4; host slices of 2; TASMultiLayerTopology": {
```
mimowo, line 5405: Please also add some negative test cases, where the Workload could fit if not for the multi-level requirements

pkg/cache/scheduler/tas_flavor_snapshot.go
Huang-Wei: When len(minusOneLayers) == 1, previously the 1-layer topology constraint has been processed (re-use old logic to strip layers[0]), so here we don't need to process it again and hence I drop it.

It's a bit nuanced I know, that's why I mentioned the alpha impl. is intentional to try not introducing massive change to existing logic (see `Special notes for your reviewer` section).

pkg/cache/scheduler/tas_flavor_snapshot.go
Huang-Wei: It's around ~86%:

```
⇒  go tool cover -func=/tmp/tas_cover.out | grep buildSliceSizeAtLevel
sigs.k8s.io/kueue/pkg/cache/scheduler/tas_flavor_snapshot.go:942:	buildSliceSizeAtLevel			86.4%
```

and missing hits are mostly about validation error paths for malformed constraint inputs (like `prevSize%layer.Size != 0`)

pkg/controller/jobframework/tas_validation.go
Huang-Wei: done via `errTopologyConstraintsLayerCount`

pkg/cache/scheduler/tas_flavor_snapshot.go
Huang-Wei: Update the comments.

pkg/cache/scheduler/tas_flavor_snapshot.go
Huang-Wei: Done.

pkg/cache/scheduler/tas_cache_test.go
Huang-Wei: Done.

pkg/controller/jobframework/tas_test.go
Huang-Wei: Done.

pkg/cache/scheduler/tas_flavor_snapshot.go
mimowo: I see, yeah lets make sure we cover 100 percent for the positive scenarios

pkg/controller/jobframework/tas_test.go
Huang-Wei: Actually I missed the logic to guard 0 and >=3, added the tests and logic.

pkg/cache/scheduler/tas_flavor_snapshot.go
mimowo: I see, sgtm

pkg/cache/scheduler/tas_cache_test.go
```diff
@@ -5630,6 +5630,118 @@ func TestFindTopologyAssignments(t *testing.T) {
...
+					PodsetSliceRequiredTopologyConstraints: []kueue.PodsetSliceRequiredTopologyConstraint{
+						{Topology: tasRackLabel, Size: 8},
+						{Topology: corev1.LabelHostname, Size: 4},
+					},
+				},
+				requests: resources.Requests{
+					corev1.ResourceCPU: 1000,
+				},
+				count:      16,
+				wantReason: `topology "default" doesn't allow to fit any of 2 slice(s)`,
```
mimowo, line 5742: I'm wondering how we could improve the message here to be useful.

Maybe we could describe the next domain which could fit some of the Pods, but not everything (next next best domain). 

So, maybe say: `topology "default" doesn't allow to fit any of 2 slice(s); the next domain "xyz" could only fit 1/2 sub-slices at the kubernetes.io/hostname level`. 

However, I'm not totally convinced this is the best approach, we may wait for the user feedback here / see what is useful for debugging.

mimowo:
/approve
Thank you @Huang-Wei, lgtm, please rebase. And let's follow up with the NodeHotSwap improvements to make sure the newly constructed assignment conforms to the restrictions.

pkg/cache/scheduler/tas_cache_test.go
Huang-Wei: Given we have a restriction on # of layers, I may lean towards dump the stats of error reasons on all layers, like:

```
topology "default" doesn't allow to fit; 1/2 slice(s) fit on level cloud.com/topology-rack; 3/4 slice(s) fit on level kubernetes.io/hostname
```

If we go w/ this, the current err msg pattern doesn't quite match b/c it's based on single-layer:

https://github.com/kubernetes-sigs/kueue/blob/2b0916e12fdcb73a63ceffaa6ec6313cd30099a8/pkg/cache/scheduler/tas_flavor_snapshot.go#L1497-L1501

so I may come up w/ a standalone err msg builder for it, and unify it when this feature reaches to beta.

I will need some time to think about it. (code is rebased btw)

PR #9516, opened 2026-02-26
Multikueue dev setup: extend SA token expiration time

### LINKED ISSUE #9280: MultiKueue SA tokens expire after 24 hours in a local development environment
When using the [`setup-kind-multikueue-tas.sh`](https://github.com/kubernetes-sigs/kueue/blob/9fd2597755e55160dc976866b1f2254f37d0c8bf/site/static/examples/multikueue/dev/setup-kind-multikueue-tas.sh) script from the site docs, MultiKueue stops working after 24 hours. This is due to the service account token being generated for 24 hours:
https://github.com/kubernetes-sigs/kueue/blob/9fd2597755e55160dc976866b1f2254f37d0c8bf/site/static/examples/multikueue/dev/setup-kind-multikueue-tas.sh#L155

The token's lifespan cannot be arbitrarily increased since it seems to be capped at 48 hours:
```sh
Warning: requested expiration of 2592000 seconds shortened to 172800 seconds
```

The token can be easily rotated with a script. It should be explicitly mentioned in the documentation that the token needs to be rotated.

**Why is this needed**:
1. To make the development experience simpler.
2. To make sure users that are testing out the capabilities of MultiKueue don't get discouraged by seemingly random failures.

polinasand:
#### What type of PR is this?

/kind cleanup

During [Multikueue dev setup](https://kueue.sigs.k8s.io/docs/tasks/dev/setup_multikueue_development_environment/#setup-multikueue-with-tas) process, a ServiceAccount is created for each of worker clusters. SA token expires in 24h, leading to need for a restart of the entire MK dev setup. To avoid this, increase expiration time to 7d.

This PR creates long-live API token for SA token during MultiKueue development setup. This approach is used in non-dev [MultiKueue setup docs](https://kueue.sigs.k8s.io/docs/tasks/manage/setup_multikueue/).

Fixes #9280

```release-note
NONE
```


site/static/examples/multikueue/dev/setup-kind-multikueue-tas.sh
```diff
@@ -142,14 +143,25 @@ for cluster in ${WORKER_CLUSTERS}; do
...
+    kubectl --context "kind-${cluster}" apply -f - <<EOF 
+apiVersion: v1
+kind: Secret
+metadata:
+  name: multikueue-sa-token
+  namespace: kueue-system
+  annotations:
+    kubernetes.io/service-account.name: "${SERVICE_ACCOUNT}"
+type: kubernetes.io/service-account-token
+EOF
```
kshalot, line 162: This is the long-lived token, right? I thought we were worried about people copying it over to their prod setup (https://github.com/kubernetes-sigs/kueue/issues/9280#issuecomment-3958816258).

kshalot (commented):
@polinasand I left one comment in the TAS setup script, so I'm thinking if we can maybe split this into two PRs? For example so that https://github.com/kubernetes-sigs/kueue/pull/9516/commits/e3f5c47897e04f54b19c66ee2c8192682db8dbcb becomes a separate PR. They are independent and we could merge that one quicker, since there's no change/discussion required.

hack/testing/multikueue/worker-cluster.kind.yaml
```diff
@@ -31,4 +31,5 @@ kubeadmConfigPatches:
     kind: JoinConfiguration
     nodeRegistration:
       kubeletExtraArgs:
-        v: "3"
\ No newline at end of file
+        v: "3"
+        service-account-max-token-expiration: "168h"
```
kshalot: This is in the wrong place, hence the test failure. You added it under `JoinConfiguration`. It should be sth like:
```
kind: ClusterConfiguration <-
apiServer:
  extraArgs:
    service-account-max-token-expiration: "168h" 
```

I guess this just doesn't parse and the join process cannot finish, that's why it's failing instead of being a no-op.

PR #9135, opened 2026-02-11
Fix flake in scale-down logic for LeaderWorkerSet

PannagaRao:
Signed-off-by: Pannaga Rao Bhoja Ramamanohara

#### What type of PR is this?
/kind bug
/kind flake

#### Which issue(s) this PR fixes:

Fixes # https://github.com/kubernetes-sigs/kueue/issues/9034

#### Special notes for your reviewer:

Ran the flaky test 300 times with the fix applied — all passed.
Reference : https://github.com/kubernetes-sigs/kueue/pull/9050

#### Does this PR introduce a user-facing change?

```release-note
LeaderWorkerSet: fix an occasional race condition resulting in workload deletion getting stuck during scale down.
```

PR #9529, opened 2026-02-26
WIP : Add extra logs for leaderworkerset_reconciler

PannagaRao:

Fixes #

```release-note

```

### DIFF
--- a/test/e2e/singlecluster/leaderworkerset_test.go
+++ b/test/e2e/singlecluster/leaderworkerset_test.go
@@ -681,7 +681,7 @@ var _ = ginkgo.Describe("LeaderWorkerSet integration", ginkgo.Label("area:single
 			})
 		})
 
-		for i := range 100 {
+		for i := range 200 {
 			ginkgo.FIt(fmt.Sprintf("Rolling update with maxSurge creates workloads for surge pods and completes successfully %d", i), func() {
 				const (
 					lwsReplicas    = 4


PannagaRao:
@mimowo I added diagnostic logging to the test's Eventually block since reconciler-side logs were getting rotated out before artifact collection with 200 iterations. Here's what I observed:                                               
                                                                                                                                                                                                                                              
The updatedReplicas < spec.Replicas check in isRollingUpdateWithSurge (added in #9135) causes a race condition with the LWS controller's rolling update.                                                                                    
   
When updatedReplicas reaches spec.Replicas, Kueue treats the rolling update as complete and deletes the surge workloads. But the LWS controller hasn't finished its own surge cleanup yet — status.Replicas is still greater than spec.Replicas. Kueue deleting the workloads causes the surge pods to terminate, which disrupts the LWS controller's in-progress cleanup and updatedReplicas regresses, creating an oscillation

```
  updatedReplicas=3, status=6 → all 6 workloads admitted    (isRollingUpdateWithSurge=true, surge kept)
  updatedReplicas=4, status=6 → workloads 4,5 deleted       (isRollingUpdateWithSurge=false, surge deleted)
                               → pods for groups 4,5 start terminating
  updatedReplicas=3, status=5 → workload 4 re-created       (regressed, 1 surge group cleaned up by LWS)
  updatedReplicas=4, status=5 → workload 4 deleted          (cycle repeats)
  updatedReplicas=3, status=4 → no surge groups remain      (2nd surge group cleaned up by LWS)
  updatedReplicas=4, status=4, ready=4 → done
```

Prow logs confirming the chain (workload deleted → pod deleting=true → LWS recreates pod → updatedReplicas regresses): https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/9529/pull-kueue-test-e2e-main-1-33/2027106281306722304

This oscillation mostly converges (each cycle the LWS controller cleans up one surge group), but convergence time is non-deterministic and can exceed the 5-min timeout under load.

Simply removing the updatedReplicas check would risk the scale-down flake that #9135 fixed. One approach I'm considering: keep the existing isRollingUpdateWithSurge logic, , but in filterWorkloads, rescue surplus workloads from deletion if their groups still have running pods .

WDYT? Open to other directions I might be missing.

PR #7927, opened 2025-11-27
Automatically synchronize FGs and documentation FGs list

vladikkuzn:
#### What type of PR is this?
/kind documentation

#### What this PR does / why we need it:
To automatically synchronize FGs and the documentation FGs list

#### Which issue(s) this PR fixes:

Fixes #5507

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #9533, opened 2026-02-26
Automated cherry pick of #9484: [Fair Sharing] Prefer nominal-quota workloads in admission ordering

mukund-wayve:
Cherry pick of #9484 on release-0.15.

#9484: [Fair Sharing] Prefer nominal-quota workloads in admission ordering

For details on the cherry pick process, see the [cherry pick requests](https://git.k8s.io/community/contributors/devel/sig-release/cherry-picks.md) page.

#### What type of PR is this?
/kind bug

```release-note
FairSharing: workloads fitting within their ClusterQueue's nominal quota are now preferred over workloads that require borrowing, preventing heavy borrowing on one flavor from deprioritizing a CQ's nominal entitlement on another flavor.
```

### DIFF
--- a/pkg/features/kube_features.go
+++ b/pkg/features/kube_features.go
@@ -288,7 +288,7 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 		{Version: version.MustParse("0.6"), Default: true, PreRelease: featuregate.Beta},
 	},
 	FairSharingPrioritizeNonBorrowing: {
-		{Version: version.MustParse("0.17"), Default: true, PreRelease: featuregate.Beta},
+		{Version: version.MustParse("0.15"), Default: true, PreRelease: featuregate.Beta},
 	},
 	MultiKueue: {
 		{Version: version.MustParse("0.6"), Default: false, PreRelease: featuregate.Alpha},


pkg/features/kube_features.go
```diff
@@ -280,6 +287,9 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 	PrioritySortingWithinCohort: {
 		{Version: version.MustParse("0.6"), Default: true, PreRelease: featuregate.Beta},
 	},
+	FairSharingPrioritizeNonBorrowing: {
+		{Version: version.MustParse("0.17"), Default: true, PreRelease: featuregate.Beta},
```
mimowo: Let's align to 0.15

mukund-wayve: Done

site/data/featuregates/versioned_feature_list.yaml
```diff
@@ -35,6 +35,12 @@
     lockToDefault: false
     preRelease: Alpha
     version: "0.15"
+- name: FairSharingPrioritizeNonBorrowing
+  versionedSpecs:
+  - default: true
+    lockToDefault: false
+    preRelease: Beta
+    version: "0.17"
```
mukund-wayve, line 43: @tenzen-y  this shouldn't have been allowed to be merged right? The source code changed the version to 0.15, but this generated file is still at 0.17. Same with the next file as well.

Should I raise a PR against `release-0.15` to fix these two files?

mimowo: TBH, this almost does not matter- the infromation about the status gates is displayed based on the "website" branch. It only may matter for some developer checking historical branches and getting surprised.

If you have a bit of time, then please align that.

I think this is likely because we didn't have some validation on 0.15. We recently did generation of the documentation for FGs automatically, so most likely this is due to subtle changes in the way we generate docs going forward.

mukund-wayve: Could it have been because of CI caching results of `pull-kueue-verify`?

mimowo: Oh maybe, the mechanism was done by @vladikkuzn in https://github.com/kubernetes-sigs/kueue/pull/7927 and it was first included in 0.16.

mimowo: > Could it have been because of CI caching results of pull-kueue-verify?

On the second thought - I don't think so. It could explain why the stuff didn't get updated locally on your machine, but in that case the merge should have been blocked on CI. 

I think the lack of generation (putting by hand) is the more likely explanation.

mukund-wayve: Got it. Thanks for the explanation! Unrelated question, how long do we maintain old versions? Is it always that we maintain 2 active versions - latest and the one before?

PR #9570, opened 2026-02-27
test(e2e): stabilize kuberay InTreeAutoscaling worker scale-up timing

### LINKED ISSUE #9523: Kuberay Should run a rayjob with InTreeAutoscaling [area:singlecluster, feature:kuberay]
**Which test is flaking?**:
] Kuberay Should run a rayjob with InTreeAutoscaling [area:singlecluster, feature:kuberay]
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/9521/pull-kueue-test-e2e-main-1-33/2027007885115920384
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/9521/pull-kueue-test-e2e-main-1-34/2027007885233360896
**Failure message or logs**:
```

End To End Suite: kindest/node:v1.33.7: [It] Kuberay Should run a rayjob with InTreeAutoscaling [area:singlecluster, feature:kuberay] expand_less	6m10s
{Timed out after 300.001s.
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/singlecluster/kuberay_test.go:334 with:
Expected exactly 5 pods with 'workers' in the name
Expected
    <[]string | len:0, cap:0>: nil
to have length 5 failed [FAILED] Timed out after 300.001s.
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/singlecluster/kuberay_test.go:334 with:
Expected exactly 5 pods with 'workers' in the name
Expected
    <[]string | len:0, cap:0>: nil
to have length 5
In [It] at: /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/singlecluster/kuberay_test.go:342 @ 02/26/26 13:33:38.392
}
```

Sebastianhayashi:
Fixes #9523.

## Summary
This addresses a flaky assertion in the Kuberay InTreeAutoscaling e2e test where worker scale-up is expected but often not observed within the prior task runtime window.

The test script currently runs parallel tasks with `sleep=8s`, which can complete before autoscaler demand is observed and worker pods are scheduled on slower CI nodes.

## Change
- In `test/e2e/singlecluster/kuberay_test.go`, increase the parallel task sleep from `8` to `30` seconds in the embedded Python snippet used to trigger autoscaling.

## Why this is safe
- Test-only change (no production/controller logic touched).
- Keeps assertion semantics intact (still validates worker scale-up), but gives autoscaling enough time to converge.

## Evidence
- Failing signal: expected 5 worker pods, observed 0 in CI for this scenario.
- Local verification for focused scenario was captured in the evidence bundle.

## Notes
I cannot run Go-based e2e locally in this shell environment because toolchain/runtime setup is constrained here; this PR relies on project CI validation.

```release-note
NONE
```

### DIFF
--- a/test/e2e/singlecluster/kuberay_test.go
+++ b/test/e2e/singlecluster/kuberay_test.go
@@ -200,9 +200,9 @@ def my_task(x, s):
 print([ray.get(my_task.remote(i, 1)) for i in range(4)])
 
 # run tasks in parallel to trigger autoscaling (scaling up)
-# Use longer sleep (8s) to give autoscaler time to detect demand,
-# create workload slices, and schedule new workers.
-print(ray.get([my_task.remote(i, 8) for i in range(16)]))
+# Use longer sleep (18s) to give autoscaler time to detect demand,
+# create workload slices, and schedule new workers without flaking.
+print(ray.get([my_task.remote(i, 18) for i in range(16)]))
 
 # run tasks in sequence to trigger scaling down
 print([ray.get(my_task.remote(i, 1)) for i in range(32)])`,


test/e2e/singlecluster/kuberay_test.go
```diff
@@ -200,9 +200,9 @@ def my_task(x, s):
 print([ray.get(my_task.remote(i, 1)) for i in range(4)])
 
 # run tasks in parallel to trigger autoscaling (scaling up)
-# Use longer sleep (8s) to give autoscaler time to detect demand,
-# create workload slices, and schedule new workers.
-print(ray.get([my_task.remote(i, 8) for i in range(16)]))
+# Use longer sleep (30s) to give autoscaler time to detect demand,
+# create workload slices, and schedule new workers without flaking.
+print(ray.get([my_task.remote(i, 30) for i in range(16)]))
```
mimowo: This bump might be quite a bit, and the test is flaking rarely. Can we try first with a smaller bump? Maybe 10s, wdyt?

Sebastianhayashi: thx for the review, make sense.

I reduced the bump to a smaller value first (+10s) as suggested, and pushed an update.
I also fixed the commit message format so it doesn't contain `#`.

take another look, plz.

sohankunkerkar: /hold
I don't think this is the right fix. It will alleviate the symptom but not the root cause.

sohankunkerkar: The leader-controller logs in [this](https://github.com/kubernetes-sigs/kueue/issues/9523) issue show a tight requeue loop with repeated `Handling job with no workload found an existing workload` messages (requeued with time.Nanosecond).

The autoscaler does fire but the logs show Updated PodSet worker count… newCount: 5. However, `newWorkloadName()` derives the workload name from the RayJob’s metadata.generation, which never changes. InTreeAutoscaling updates the RayCluster (the child), not the RayJob itself. As a result, the newly generated workload slice name collides with the existing one, leading to AlreadyExists, an infinite requeue loop, and the 5-worker workload never being created.

IMO, this looks like a controller bug in the workload-slice naming logic, not a test timing issue.

Sebastianhayashi: > The leader-controller logs in [this](https://github.com/kubernetes-sigs/kueue/issues/9523) issue show a tight requeue loop with repeated `Handling job with no workload found an existing workload` messages (requeued with time.Nanosecond).
> 
> The autoscaler does fire but the logs show Updated PodSet worker count… newCount: 5. However, `newWorkloadName()` derives the workload name from the RayJob’s metadata.generation, which never changes. InTreeAutoscaling updates the RayCluster (the child), not the RayJob itself. As a result, the newly generated workload slice name collides with the existing one, leading to AlreadyExists, an infinite requeue loop, and the 5-worker workload never being created.
> 
> IMO, this looks like a controller bug in the workload-slice naming logic, not a test timing issue.

thx for review, acknowledged. I agree this needs a root-cause fix, not a symptom-only change.

mimowo: I see, yes I think @sohankunkerkar is likely correct the issue is that the issue is with generation number not changing for RayJob. 

Maybe to solve that problem we shouldn't need to use generation, but rather `ResourceVersion` when creating the name for the new workload slice. 

Then, we would have a small controller which would set JSON annotation with the sizes of the RayCluster PodSets after the scaling- thus bumping the RayJob's ResourceVersion (but not generation). We could name it something like `kueue.x-k8s.io/raycluster-replica-sizes`. 

Then:
1. the code generating the WorkloadSlice will see a change (ResourceVersion bumped), solving the flake
2. the code for `PodSets()` in RayJob and RayService could be simplified to only check the annotation values -  not need to lookup the RayCluster itself, just rely on the annotation. 
3. Such an annotation is also useful for MultiKueue for Autoscaling support, so that we can observe the change of size on the worker cluster just by watching the RayJob object. Then, the manager cluster could see the change of size, and create the WorkloadSlice.

cc @sohankunkerkar @hiboyang wdyt?

hiboyang: Yes,  adding `kueue.x-k8s.io/raycluster-replica-sizes` annotation is a good idea, so the top level job (RayJob) could have information of underlying RayCluster replica change. I am experimenting it. Will come out a PR.

A slight suggestion is to make the annotation name more generic, like `kueue.x-k8s.io/podset-replica-sizes`?

mimowo: Maybe, I guess it depends if our MultiKueue code could be generic to read it. My intuition is that this is KubeRay specific, but Im open to discuss

hiboyang: Here is PR to add that annotation: https://github.com/kubernetes-sigs/kueue/pull/9726 , would like to hear feedbacks

mimowo:
/close
I think the stability issue is resolved by now

PR #9579, opened 2026-02-27
update release template post release 0.15.5/0.16.2

gabesaba:
#### What type of PR is this?
/kind documentation

Clarify release process with issues ran into during latest release

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/ISSUE_TEMPLATE/NEW_RELEASE.md
+++ b/.github/ISSUE_TEMPLATE/NEW_RELEASE.md
@@ -53,9 +53,9 @@ Please do not remove items from the checklist
 - [ ] Run the [openvex action](https://github.com/kubernetes-sigs/kueue/actions/workflows/openvex.yaml) to generate openvex data. The action will add the file to the release artifacts.
 - [ ] Run the [SBOM action](https://github.com/kubernetes-sigs/kueue/actions/workflows/sbom.yaml) to generate the SBOM and add it to the release.
 - [ ] Update the `main` branch :
-  - [ ] for both releases:
-    - [ ] manually update changelog
-  - [ ] for the latest release:
+  - [ ] for each release:
+    - [ ] update the changelog (example #9578)
+  - [ ] for only the latest release:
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/NEW_RELEASE.md
```diff
@@ -50,9 +53,12 @@ Please do not remove items from the checklist
 - [ ] Run the [openvex action](https://github.com/kubernetes-sigs/kueue/actions/workflows/openvex.yaml) to generate openvex data. The action will add the file to the release artifacts.
 - [ ] Run the [SBOM action](https://github.com/kubernetes-sigs/kueue/actions/workflows/sbom.yaml) to generate the SBOM and add it to the release.
 - [ ] Update the `main` branch :
-  - [ ] Run `./hack/releasing/prepare_pull.sh --target main $VERSION`
-  - [ ] Submit a pull request with the changes: <!-- example #3007 -->
-  - [ ] Cherry-pick the pull request onto the `website` branch
+  - [ ] for both releases:
+    - [ ] manually update changelog
```
mimowo: Hm , this is curious. I didn't follow the latest patch releases closely, but I without the context I wouldn't know what is behind the step. Please try to rephrase to make it clear.

Is eh "both releases" referring to the both patch releases being done at the same time, like "0.15.5", and "0.16.2"? I think this steps should abstract it out. We typically do them both together, but it does not need to always be the case.

gabesaba: I clarified a bit. Let me know if you have something else in mind. For the non-latest patch release, it seems the only step is to update the changelog (and this is underspecified in current template)

mimowo: oh I see now. I didn't read it this way before, so this is cleaner for sure. Ideally the update of the changelog could also be done by the script, but for that we can improve the issue once we have it.

Issue #9301, opened 2026-02-17
[Flaky E2E] ManageJobsWithoutQueueName when manageJobsWithoutQueueName=true and ManagedJobsNamespaceSelectorAlwaysRespected=false should not suspend Jobs from unmanaged JobSe


mbobrovskyi:
**Which test is flaking?**:
End To End Custom Configs handling Suite: kindest/node:v1.35.0: [It] ManageJobsWithoutQueueName when manageJobsWithoutQueueName=true and ManagedJobsNamespaceSelectorAlwaysRespected=false should not suspend Jobs from unmanaged JobSet 

**Link to failed CI job or steps to reproduce locally**:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/periodic-kueue-test-e2e-customconfigs-release-0-15/2023574537966194688

**Failure message or logs**:
```
{Expected success, but got an error:
    <*errors.StatusError | 0xc00068ea00>: 
    conversion webhook for kueue.x-k8s.io/v1beta1, Kind=LocalQueue failed: Post "https://kueue-webhook-service.kueue-system.svc:443/convert?timeout=30s": EOF
    {
        ErrStatus: {
            TypeMeta: {Kind: "", APIVersion: ""},
            ListMeta: {
                SelfLink: "",
                ResourceVersion: "",
                Continue: "",
                RemainingItemCount: nil,
            },
            Status: "Failure",
            Message: "conversion webhook for kueue.x-k8s.io/v1beta1, Kind=LocalQueue failed: Post \"https://kueue-webhook-service.kueue-system.svc:443/convert?timeout=30s\": EOF",
            Reason: "",
            Details: nil,
            Code: 500,
        },
    } failed [FAILED] Expected success, but got an error:
    <*errors.StatusError | 0xc00068ea00>: 
    conversion webhook for kueue.x-k8s.io/v1beta1, Kind=LocalQueue failed: Post "https://kueue-webhook-service.kueue-system.svc:443/convert?timeout=30s": EOF
    {
        ErrStatus: {
            TypeMeta: {Kind: "", APIVersion: ""},
            ListMeta: {
                SelfLink: "",
                ResourceVersion: "",
                Continue: "",
                RemainingItemCount: nil,
            },
            Status: "Failure",
[...]

PR #9580, opened 2026-02-27
test(e2e): retry LocalQueue create around webhook restart race

Sebastianhayashi:
#### What type of PR is this?

/kind cleanup

The `ManageJobsWithoutQueueName` e2e flow can fail right after a controller/webhook restart with transient webhook EOFs when creating `LocalQueue`.

This change wraps `LocalQueue` creation in `Eventually` and retries `Create` until success, while tolerating `already exists` when an earlier create succeeded but response handling raced.

Related to #9301

- Test-only change in `test/e2e/customconfigs/managejobswithoutqueuename_test.go`.
- No production controller logic changed.

```release-note
NONE
```

### DIFF
--- a/test/e2e/customconfigs/managejobswithoutqueuename_test.go
+++ b/test/e2e/customconfigs/managejobswithoutqueuename_test.go
@@ -74,13 +74,14 @@ var _ = ginkgo.Describe("ManageJobsWithoutQueueName", ginkgo.Ordered, func() {
 		ns = util.CreateNamespaceFromPrefixWithLog(ctx, k8sClient, "e2e-")
 		defaultRf = utiltestingapi.MakeResourceFlavor("default").Obj()
 		util.MustCreate(ctx, k8sClient, defaultRf)
-		clusterQueue = utiltestingapi.MakeClusterQueue("cluster-queue").
+		clusterQueueName := "cluster-queue-" + ns.Name
+		clusterQueue = utiltestingapi.MakeClusterQueue(clusterQueueName).
 			ResourceGroup(
 				*utiltestingapi.MakeFlavorQuotas(defaultRf.Name).
 					Resource(corev1.ResourceCPU, "2").
 					Resource(corev1.ResourceMemory, "2G").Obj()).Obj()
 		util.CreateClusterQueuesAndWaitForActive(ctx, k8sClient, clusterQueue)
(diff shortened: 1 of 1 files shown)


test/e2e/customconfigs/managejobswithoutqueuename_test.go
```diff
@@ -81,7 +82,13 @@ var _ = ginkgo.Describe("ManageJobsWithoutQueueName", ginkgo.Ordered, func() {
...
 		util.CreateClusterQueuesAndWaitForActive(ctx, k8sClient, clusterQueue)
 		localQueue = utiltestingapi.MakeLocalQueue("main", ns.Name).ClusterQueue("cluster-queue").Obj()
-		util.CreateLocalQueuesAndWaitForActive(ctx, k8sClient, localQueue)
+		gomega.Eventually(func(g gomega.Gomega) {
+			err := k8sClient.Create(ctx, localQueue)
+			if err != nil && !strings.Contains(err.Error(), "already exists") {
+				g.Expect(err).NotTo(gomega.HaveOccurred())
+			}
+		}, util.Timeout, util.Interval).Should(gomega.Succeed())
+		util.ExpectLocalQueuesToBeActive(ctx, k8sClient, localQueue)
```
mimowo: /hold 
I don't think we can rely on the resources from previous tests. They may have some properties not good for the new test. What we could do it to make sure all the resources created by each test have dedicated prefix / suffix hash, to make sure they don't conflict even in case of some orphaned resoruces by previous tests.

mimowo: PTAL how we do it in the regular e2e test suite for singlecluster. We do it there to have parallelism, but we could reuse the same approach here.

mimowo:
Opened the flake issue: https://github.com/kubernetes-sigs/kueue/issues/9582

test/e2e/customconfigs/managejobswithoutqueuename_test.go
Sebastianhayashi: I reworked the test to avoid relying on resources from previous runs by isolating queue names per test namespace.
Take another look, plz.

mimowo: /unhold
Thank you 👍

mimowo:
/close
Closing stale tickets which haven't re-occur in 2 months, because maybe some of them are already fixed along the way. We will re-open when they re-occur again.

Issue #9591, opened 2026-02-27
Scheduler when Requeueing Inadmissible Workloads Should collapse requeue requests to ClusterQueue


mimowo:
**Which test is flaking?**:
 Scheduler when Requeueing Inadmissible Workloads Should collapse requeue requests to ClusterQueue
**Link to failed CI job or steps to reproduce locally**:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/9528/pull-kueue-test-integration-baseline-release-0-15/2027424372553158656
**Failure message or logs**:
```
Scheduler Inadmissible Requeueing Suite: [It] Scheduler when Requeueing Inadmissible Workloads Should collapse requeue requests to ClusterQueue expand_less	15s
{Timed out after 10.001s.
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/kueue/test/util/util.go:577 with:
pending_workloads with status=inadmissible
Expected
    <int>: 3
to be <=
    <int>: 2 failed [FAILED] Timed out after 10.001s.
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/kueue/test/util/util.go:577 with:
pending_workloads with status=inadmissible
Expected
    <int>: 3
to be <=
    <int>: 2
In [It] at: /home/prow/go/src/sigs.k8s.io/kueue/test/integration/singlecluster/scheduler/inadmissible/requeueing_test.go:83 @ 02/27/26 16:59:55.868
}
```

PR #9232, opened 2026-02-13
Asynchronous Inadmissible Workload Requeueing

gabesaba:
#### What type of PR is this?
/kind bug

#### What this PR does / why we need it:
We decouple requesting inadmissible workloads to be reprocessed,
from the processing of these requests where we move the workloads.

This has a few main advantages
1) Allows batching requeues, avoiding the starvation noted [here](https://github.com/kubernetes-sigs/kueue/issues/8095#issuecomment-3860607333)
2) Reduces lock contention
3) Avoid spamming requeues 20+ times per second.

We accomplish this by defining the `inadmissibleWorkloadRequeuer`,
which implements the `requeueInadmissibleListener` interface.

Any requests to requeue inadmisisble workloads must go through this interface.
Then, the requeuer will process these requests in batches. The requeuer
deduplicates requests to the same ClusterQueue/Root Cohort, further reducing
duplicate reprocessing.

#### Which issue(s) this PR fixes:
Fixes #8095

#### Special notes for your reviewer:
* one preparatory PR merges this diff will be smaller #9224,
* see note in 9224 about exporting these test factories #9224
* see comment about eventCh buffering - need some feedback here

#### Does this PR introduce a user-facing change?
```release-note
Scheduling: Fix the bug where inadmissible workloads would be re-queued too frequently at scale.
This resulted in excessive processing, lock contention, and starvation of workloads deeper in the queue.
The fix is to throttle the process with a batch period of 1s per CQ or Cohort.
```

PR #9598, opened 2026-02-28
Skip PushOrUpdate for inflight workloads to prevent spurious scheduling cycles

sohankunkerkar:
When a workload is popped by the scheduler (inflight), a concurrent PushOrUpdate from the workload controller can push it onto the heap while the scheduler also places it in the inadmissible set. This causes an extra scheduling cycle.

#### What type of PR is this?
/kind bug

Fixes https://github.com/kubernetes-sigs/kueue/issues/9591

I ran the failed test 100 times in a loop and it looks fine.

```release-note
Scheduling: Fixed a race condition where a workload could simultaneously exist in the scheduler's heap
and the "inadmissible workloads" list. This fix prevents unnecessary scheduler cycles and prevents temporary 
double counting for the metric of pending workloads.
```

### DIFF
--- a/pkg/cache/queue/cluster_queue.go
+++ b/pkg/cache/queue/cluster_queue.go
@@ -252,7 +252,11 @@ func (c *ClusterQueue) PushOrUpdate(wInfo *workload.Info) {
 	c.rwm.Lock()
 	defer c.rwm.Unlock()
 	key := workload.Key(wInfo.Obj)
-	c.forgetInflightByKey(key)
+	// Skip if the scheduler is actively processing this workload.
+	// RequeueWorkload will handle placement with the latest version.
+	if c.inflight != nil && workload.Key(c.inflight.Obj) == key {
+		return
+	}
 	if oldInfo := c.inadmissibleWorkloads.get(key); oldInfo != nil {
 		// update in place if the workload was inadmissible and didn't change
 		// to potentially become admissible, unless the Eviction status changed
--- a/pkg/cache/queue/cluster_queue_test.go
+++ b/pkg/cache/queue/cluster_queue_test.go
@@ -212,6 +212,37 @@ func Test_Pop(t *testing.T) {
 	}
 }
 
+func TestPushOrUpdateSkipsInflightWorkload(t *testing.T) {
+	ctx, _ := utiltesting.ContextWithLog(t)
+	now := time.Now()
+	cq := newClusterQueueImpl(ctx, nil, defaultOrdering, testingclock.NewFakeClock(now))
+
+	wl := utiltestingapi.MakeWorkload("workload-1", defaultNamespace).Creation(now).Obj()
+	cq.PushOrUpdate(workload.NewInfo(wl))
+
+	// Pop makes the workload inflight.
+	head := cq.Pop()
+	if head == nil {
+		t.Fatal("expected to pop workload")
+	}
+
+	// Simulate a concurrent PushOrUpdate while the workload is inflight.
+	updatedWl := utiltestingapi.MakeWorkload("workload-1", defaultNamespace).
+		Creation(now).ResourceVersion("1").Obj()
+	cq.PushOrUpdate(workload.NewInfo(updatedWl))
+
+	// The workload should not be on the heap or in inadmissible.
+	activeWorkloads, _ := cq.Dump()
+	if len(activeWorkloads) != 0 {
+		t.Errorf("expected empty heap while workload is inflight, got %v", activeWorkloads)
+	}
+
+	inadmissibleWorkloads, _ := cq.DumpInadmissible()
+	if len(inadmissibleWorkloads) != 0 {
+		t.Errorf("expected no inadmissible workloads while workload is inflight, got %v", inadmissibleWorkloads)
+	}
+}
+
 func Test_Delete(t *testing.T) {
 	ctx, log := utiltesting.ContextWithLog(t)
 	cq := newClusterQueueImpl(ctx, nil, defaultOrdering, testingclock.NewFakeClock(time.Now()))


pkg/cache/queue/cluster_queue.go
```diff
@@ -252,6 +252,11 @@ func (c *ClusterQueue) PushOrUpdate(wInfo *workload.Info) {
 	c.rwm.Lock()
 	defer c.rwm.Unlock()
 	key := workload.Key(wInfo.Obj)
+	// Skip if the scheduler is actively processing this workload.
+	// RequeueWorkload will handle placement with the latest version.
+	if c.inflight != nil && workload.Key(c.inflight.Obj) == key {
```
mimowo, line 257: This looks like a good change, so that we have only one place where the workload is requeued to: inadmissibke or heap. 

iiuc this could also happen before but we didnt have a test surfaving the issue?

sohankunkerkar: I don't think so. Actually, before #9232, requeueing was synchronous. There was no `AddAfter` delay, so when a CQ update triggered a requeue, the inadmissible workloads were moved to the heap immediately within the same `UpdateClusterQueue` call, under the manager lock. The scheduler couldn't be processing (inflight) a workload while the requeue happened because both operations were serialized by the manager lock.            

With async requeueing (#9232), the requeue goes through a `workqueue.AddAfter` with a batch delay. The requeue fires asynchronously in a separate goroutine, which means the scheduler can be actively processing a workload (inflight) when the workload controller's PushOrUpdate is called that's the window where the race occurs.

mimowo: > There was no AddAfter delay, so when a CQ update triggered a requeue, the inadmissible workloads were moved to the heap immediately within the same UpdateClusterQueue call, under the manager lock.

I see, but before https://github.com/kubernetes-sigs/kueue/pull/9232 it was possible that the workload_controller could put it into inadmissible due to another issue, say due temporarily inactive LQ, while the scheduler's cycle is "in progress", and puts it back to heap if the LQ state change was very short. It was unlikely, but I don't see what would prevent the phenomenon before. wdyt?

mimowo: Also, another question. Could this issue cause double-counting of workload "pending" metric, while the workload was both on heap and in inadmissible?

sohankunkerkar: >I see, but before https://github.com/kubernetes-sigs/kueue/pull/9232 it was possible that the workload_controller could put it into inadmissible due to another issue, say due temporarily inactive LQ, while the scheduler's cycle is "in progress", and puts it back to heap if the LQ state change was very short. It was unlikely, but I don't see what would prevent the phenomenon before. wdyt?

:+1: Yeah, this could be a possibility but just much harder to hit.

sohankunkerkar: > Also, another question. Could this issue cause double-counting of workload "pending" metric, while the workload was both on heap and in inadmissible?

Yup, exactly.

mimowo: Ok, thank you for the responses. So, I think we should count it as a pre-existing bug, even if unlikely. The proposed fix makes perfect sense to me, I don't see any drawbacks.

sohankunkerkar: @mimowo do you want me to change the release note?

mimowo:
/release-note-edit
```release-note
Scheduling: Fixed a race condition where a workload could simultaneously exist in the scheduler's heap
and the "inadmissible workloads" list. This fix prevents unnecessary scheduler cycles and prevents temporary 
double counting for the metric of pending workloads.
```

pkg/cache/queue/cluster_queue.go
mimowo: Let's clarify a bit, proposed: https://github.com/kubernetes-sigs/kueue/pull/9598#issuecomment-3985987910

pkg/cache/queue/cluster_queue.go
```diff
@@ -252,6 +252,11 @@ func (c *ClusterQueue) PushOrUpdate(wInfo *workload.Info) {
...
+		return
+	}
 	c.forgetInflightByKey(key)
```
mimowo: I think after the new code added the `forgetInflightByKey` below is no-op, right?

pkg/cache/queue/cluster_queue.go
mimowo, line 257: nit, maybe we could introduce a tiny helper like `matchesInflight` which would be reused in `forgetInflightByKey`, wdyt? Maybe we could also consider a tiny struct wrapper as we have for the stickyWorkload. Not needed, just exploring your opinion here

pkg/cache/queue/cluster_queue.go
sohankunkerkar: Good catch, yes it's a no-op now. I'll remove it.

pkg/cache/queue/cluster_queue.go
sohankunkerkar: I think the struct wrapper doesn't fit well here. `stickyWorkload` holds a `workload.Reference`, but inflight stores the full `*workload.Info` because pendingActive, pendingActiveInLocalQueue, and totalElements all need to read from it. A struct would just be a wrapper around a nullable pointer with getters.

For matchesInflight, after removing `forgetInflightByKey` from `PushOrUpdate`, the inline check only appears in the new guard + inside `forgetInflightByKey` itself. Since `forgetInflightByKey` already serves as the named abstraction for check+clear, extracting just the boolean part for one other call site feels like over-abstraction.

PR #6905, opened 2025-09-17
Add RunningWorkloads metric

helayoty:
#### What type of PR is this?
Add a guage metric to track the current number of workloads with `PodsReady` state. Two metrics have been added for ClusterQueue `kueue_running_workloads` and LocalQueue `kueue_local_queue_running_workloads`

/kind feature

#### What this PR does / why we need it:

To inform users how many workloads are actually making progress, since having a metric for "admitted" workloads is not necessarily reflecting that.

#### Which issue(s) this PR fixes:

Fixes #6791 

#### Does this PR introduce a user-facing change?

```release-note
Add new metrics for running workloads, `kueue_running_workloads` and `kueue_local_queue_running_workloads`
```

PR #9626, opened 2026-03-02
Feature/pods ready workloads metrics

### LINKED ISSUE #6791: Introduce a Guage metric to track the current number of "Running" workloads

Add a guage metric to track the current number of workloads with "PodsReady" - running and making progress.

**Why is this needed**:

To inform users how many workloads are actually making progress. Just "admitted" is not necessarily reflecting that.

jimangel:
#### What type of PR is this?

Add a guage metric to track the current number of workloads with PodsReady state.

| Metric | Scope | Labels |
|--------|-------|--------|
| `kueue_pods_ready_workloads` | ClusterQueue | `cluster_queue`, `replica_role` |
| `kueue_local_queue_pods_ready_workloads` | LocalQueue | `name`, `namespace`, `replica_role` |

/kind feature

To inform users how many pods are making progress (Ready), since the metric for Admitted pods may not reflect being ready; such as waiting for `ImagePull`.

Fixes #6791

The current approach is edge-based (intentionally based on [feedback](https://github.com/kubernetes-sigs/kueue/pull/6905#discussion_r2367372297) from [#6905](https://github.com/kubernetes-sigs/kueue/pull/6905)), which may lead to inconsistency if any count operation fails. It also doesn't account for the controller restarting (count reset) while workloads are running. I'm not sure there's an alternative given the constraints.

---

#### Testing:

I wrote a [script](https://gist.github.com/jimangel/9976fac4fb02e473a2df1fdaf5995512#file-gen-with-lq-cq-sh) that submits 1,000 jobs randomly between a Cluster Queue and a Local Queue. The metrics are exported to my local Prometheus / Grafana stack.

Here's my first run:

image

You'll see a spike in Pods Pendings (green) and a compounding backoff of Pods Ready (red) before catching back up with the queue. I had a bunch of runs to tune this but ultimately after:

- Increasing the CQ nominal quota (from 9 -> 56 CPUs)
- Increasing the KinD max pods per node (from 110 to 250)
- Increasing the KinD nodes (from 1 to 3)
- Updating the script to use less CPU (random 10-250m to random 10-100m)
- Updating the script to use less memory (random 10-150Mi to static 10Mi)

I now get way better, theoretical, utilization (and no pending):

image

The ~750 ready max matches my 3 nodes * 250 pods per node. I would assume adding 2 more nodes would fix the small peak Reserving.

**Note:** I also validated the absence of metrics with LocalQueue featuregate off and with WaitForPodsReady disabled

Special thanks @helayoty for starting this convo / work!

---

```release-note
Add new metrics for ready pods (running workloads), `kueue_pods_ready_workloads` and `kueue_local_queue_pods_ready_workloads`. Requires `WaitForPodsReady` to be enabled.
```

/cc @mimowo


pkg/controller/jobframework/reconciler.go
```diff
@@ -476,6 +487,17 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
 			if err != nil && !apierrors.IsNotFound(err) {
 				return ctrl.Result{}, err
 			}
+			// Decrement pods-ready gauge if finishing while PodsReady=True
```
mimowo: I think jobframework is not the best place to put the code because:
1. all the information only depends on the Workload's object state (no Job needed)
2. JobFramework reconcilers are meant lean because some users may need to maintain their own implementations for their in-house Job CRDs
3. it will not work OOTB for some internal workloads, like StatefulSet or LeaderWorkerSet which currently don't use jobframework reconciler (IIRC)

So, I would propose moving the code to the workload_controller, and react in the Create/Update/Delete event handlers. This will also allow to account for already running Workloads on restart, because every workload on restart goes via the Create event.

jimangel: Thanks for the quick feedback, let me take a look there and see what I can do!

jimangel: Would this approach still be gated by WaitForPodsReady?

mimowo: Good question, can you remind me if setting the PodsReady conditions is also guarded by the WaitForPodsReady configuration? If so, then it would only be a micro optimization. If not, then it might be surprising to the users to see the PodsReady=true condition, but not count such a workload.

jimangel: I'm still ramping on kueue, but after poking a bit, it does [seem like it is](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/jobframework/reconciler.go#L148-L150):

>// WithWaitForPodsReady indicates if the controller should add the PodsReady
>// condition to the workload when the corresponding job has all pods ready
>// or succeeded.

Can you explain more about it being a micro optimization? I was mainly following @PBundyra's [comment](https://github.com/kubernetes-sigs/kueue/pull/6905#discussion_r2367372297) initially.

Let me know the best path forward!

mimowo: > I'm still ramping on kueue, but after poking a bit, it does seem like it is:

Thank you, this is what I was looking for.

> Can you explain more about it being a micro optimization?

1. If waitForPodsReady is enable this does not matter (most installations enable it AFAIK)
2. If waitForPodsReady is disabled, then still we would only react to transitions true -> false, false -> true, but those will be absent (because it will be constantly "false"). Detecting change transitions does not require lock, so I assume it is very lightweight

> I was mainly following @PBundyra's comment initially.

Sure, we can go with this suggestion more from the "code consistency" and "conceptual clarity", ie. anything that builds on top of WaitForPodsReady (like the metric) is disabled when the feature is disabled. Then, someone who reads the code does not need to think about the performance implications, etc.

pkg/controller/core/workload_controller.go
PBundyra, line 1: Would it be possible to move this logic into jobframework reconciler https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/jobframework/reconciler.go#L511-L543 or there are some blockers?

We keep the rest of WaitForPodsReady metrics there, so extending the logic there would be more consistent and easier to maintain

PBundyra (commented):
Thanks for working on that @jimangel!

pkg/controller/core/workload_controller.go
jimangel: That's where it originally started and there was feedback to move to workload: https://github.com/kubernetes-sigs/kueue/pull/9626#discussion_r2873360777 or are you saying something else?

I'm running tests / local validation with prometheus to see if the metric is performing as expected. So far I haven't been able to match my earlier performance but there's a lot of factors at play 😅

image

jimangel: Hey @PBundyra any thoughts? Happy to meet about this too!

Issue #9483, opened 2026-02-25
LeaderWorkerSet integration when LeaderWorkerSet created Rolling update with maxSurge creates workloads for surge pods and completes successfully [area:singlecluster, feature:leaderworkerset]


mimowo:
**Which test is flaking?**:

LeaderWorkerSet integration when LeaderWorkerSet created Rolling update with maxSurge creates workloads for surge pods and completes successfully [area:singlecluster, feature:leaderworkerset]

**Link to failed CI job or steps to reproduce locally**:

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/9464/pull-kueue-test-e2e-main-1-34/2026684207736033280

**Failure message or logs**:
```
End To End Suite: kindest/node:v1.34.3: [It] LeaderWorkerSet integration when LeaderWorkerSet created Rolling update with maxSurge creates workloads for surge pods and completes successfully [area:singlecluster, feature:leaderworkerset] expand_less	5m7s
{Timed out after 300.002s.
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/singlecluster/leaderworkerset_test.go:750 with:
Expected
    <int32>: 3
to equal
    <int32>: 4 failed [FAILED] Timed out after 300.002s.
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/singlecluster/leaderworkerset_test.go:750 with:
Expected
    <int32>: 3
to equal
    <int32>: 4
In [It] at: /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/singlecluster/leaderworkerset_test.go:761 @ 02/25/26 16:08:04.057
}
```

mimowo:
I'm not sure this is related, but maybe the code around maxSurge support was modified recently: https://github.com/kubernetes-sigs/kueue/pull/9135

mimowo:
cc @PannagaRao @sohankunkerkar

sohankunkerkar:
I looked into this. The kueue controller logs for the test window are lost. The Certs test (Serial) scales the kueue deployment to 0 right after the maxSurge test times out, so the exported logs are from replacement pods with no entries for the test namespace. Without logs I can't determine the root cause. If this flakes again, the key thing to check is whether Kueue deleted workloads for the surge indices (4, 5) while the rollout was still in progress.

[...]

PR #8670, opened 2026-01-19
Add MultiKueue support for LeaderWorkerSet

IrvingMg:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:
Adds MultiKueue support for LeaderWorkerSet workloads.

- LeaderWorkerSet creates one workload per replica (e.g., an LWS with replicas: 3 creates 3 workloads). Workloads that share the same LWS owner form a workload group. All workload group members must be dispatched to the same worker cluster atomically. To achieve this:

  1. Wait for all workload group members to exist on the manager cluster.
  2. The primary workload (index 0) nominates the target worker cluster and creates the remote workload there.
  3. Once the remote workload is admitted on the worker, the primary’s ClusterName is set.
  4. All other workload group members observe the ClusterName and follow to the same worker cluster.
  5. The LWS is created on the worker cluster only after a cluster has been selected.

  The `WorkloadGroupIndexAnnotation` stores the numeric index for each workload, ensuring correct primary selection.

- The `MultiKueueOriginOwnerUIDAnnotation` annotation has been introduced because LWS creates workloads with UID-based names. Since copying the LWS to the remote generates a new UID for the remote LWS, we need to know the UID of the origin LWS (the one in the manager) in order to properly synchronize workloads from the remote with the manager. This allows the remote LWS reconciler to calculate workload names that match the prebuilt workloads from the manager.

  The `PrebuiltWorkloadLabel` is not used in this case because LWS creates one workload per replica. As a result, the LWS reconciler would be able to find only a single workload from the manager, but not the workloads corresponding to individual replicas. Using `PrebuiltWorkloadLabel` would require extending its design to support multiple replicas, which in turn would require including information about all replicas in the LWS labels.
- The `JobOwnerGVKAnnotation` and `JobOwnerNameAnnotation` are needed to ensure correct adapter lookup when the manager LWS is deleted. LWS workloads have multiple owners (LWS + Pods) because LWS uses SetOwnerReference instead of SetControllerReference to support the pod group pattern. When the manager LWS is deleted, Kubernetes GC removes only the LWS owner reference—the workload survives because the Pod owners still exist.

  The MultiKueue `adapter` function then falls back to refs[0], which is now a Pod, returning the Pod adapter instead of the LWS adapter. The Pod adapter’s SyncJob() cannot delete the remote LWS, leaving it orphaned on the worker cluster.
  
[...]

PR #9631, opened 2026-03-02
Fix lws reconciler

PannagaRao:
#### What type of PR is this?
/kind bug
/kind flake

https://github.com/kubernetes-sigs/kueue/issues/9483#issuecomment-3981396852

Fixes # https://github.com/kubernetes-sigs/kueue/issues/9483

I have verified the fix by running the flaking e2e test 200 times. These are the links to the job runs.

- https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/9631/pull-kueue-test-e2e-main-1-32/2029279361147867136
- https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/9631/pull-kueue-test-e2e-main-1-33/2029279362003505152
- https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/9631/pull-kueue-test-e2e-main-1-34/2029279362796228608
- https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/9631/pull-kueue-test-e2e-main-1-35/2029279363651866624

```release-note
LeaderWorkerSet: fix workload recreation delay during rolling updates by watching for workload deletions.
```

### DIFF
--- a/pkg/controller/jobs/leaderworkerset/leaderworkerset_reconciler.go
+++ b/pkg/controller/jobs/leaderworkerset/leaderworkerset_reconciler.go
@@ -28,14 +28,18 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	leaderworkersetv1 "sigs.k8s.io/lws/api/leaderworkerset/v1"
@@ -98,9 +102,9 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 	ctrl.Log.V(3).Info("Setting up LeaderWorkerSet reconciler")
 
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&leaderworkersetv1.LeaderWorkerSet{}).
+		For(&leaderworkersetv1.LeaderWorkerSet{}, builder.WithPredicates(r)).
 		Named(controllerName).
-		WithEventFilter(r).
+		Watches(&kueue.Workload{}, &lwsWorkloadHandler{}).
 		WithOptions(controller.Options{
 			LogConstructor: roletracker.NewLogConstructor(r.roleTracker, controllerName),
 		}).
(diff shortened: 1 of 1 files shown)


pkg/controller/jobs/leaderworkerset/event_handlers.go
```diff
@@ -0,0 +1,76 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+	"sigs.k8s.io/kueue/pkg/controller/constants"
+)
+
+// lwsWorkloadHandler watches for workload deletions and triggers reconciliation
+// of the owning LeaderWorkerSet. This ensures that during rolling updates, when
+// workloads are deleted, the LWS reconciler is triggered to recreate them.
+type lwsWorkloadHandler struct{}
```
sohankunkerkar: I think you can fold this code into an exisiting `leaderworkerset_reconciler.go`.
You can find some examples there.

PannagaRao: I have tried moving the event handler logic inline. PTAL!

pkg/controller/jobs/leaderworkerset/leaderworkerset_reconciler.go
```diff
@@ -98,9 +102,37 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			DeleteFunc: func(ctx context.Context, e event.DeleteEvent, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+				wl, ok := e.Object.(*kueue.Workload)
+				if !ok {
+					return
+				}
+
+				log := ctrl.LoggerFrom(ctx).WithValues("workload", klog.KObj(wl))
+
+				ownerGVK, hasGVK := wl.Annotations[constants.JobOwnerGVKAnnotation]
+				if !hasGVK || ownerGVK != gvk.String() {
```
mwielgus: Pardon my ignorance, where is gvk defined?

mszadkow: I had the same question, found it in line 34 of leaderworkerset_controller.go

pkg/controller/jobs/leaderworkerset/leaderworkerset_reconciler.go
```diff
@@ -98,9 +102,37 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		Named(controllerName).
-		WithEventFilter(r).
+		Watches(&kueue.Workload{}, handler.Funcs{
+			DeleteFunc: func(ctx context.Context, e event.DeleteEvent, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+				wl, ok := e.Object.(*kueue.Workload)
+				if !ok {
+					return
+				}
+
+				log := ctrl.LoggerFrom(ctx).WithValues("workload", klog.KObj(wl))
```
mszadkow: nit: we could move the log definition to line 125

PannagaRao: Sure

pkg/controller/jobs/leaderworkerset/event_handlers.go
sohankunkerkar: I would lean towards the struct-based handler pattern which is more idiomatic for controller-runtime.
```go
type workloadHandler struct{}

func (h *workloadHandler) Create(_ context.Context, _ event.CreateEvent, _ workqueue.TypedRateLimitingInterface[reconcile.Request]) {
}

func (h *workloadHandler) Update(_ context.Context, _ event.UpdateEvent, _ workqueue.TypedRateLimitingInterface[reconcile.Request]) {
}

func (h *workloadHandler) Delete(_ context.Context, e event.DeleteEvent, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
	wl, ok := e.Object.(*kueue.Workload)
	if !ok {
		return
	}
	ownerName, ok := wl.Annotations[constants.JobOwnerNameAnnotation]
	if !ok {
		return
	}
	ownerGVK, ok := wl.Annotations[constants.JobOwnerGVKAnnotation]
	if !ok || ownerGVK != gvk.String() {
		return
	}
	q.Add(reconcile.Request{
		NamespacedName: types.NamespacedName{
			Name:      ownerName,
			Namespace: wl.Namespace,
		},
	})
}

func (h *workloadHandler) Generic(_ context.Context, _ event.GenericEvent, _ workqueue.TypedRateLimitingInterface[reconcile.Request]) {
}
```

pkg/controller/jobs/leaderworkerset/leaderworkerset_reconciler.go
```diff
@@ -98,9 +102,36 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+				ownerGVK, hasGVK := wl.Annotations[constants.JobOwnerGVKAnnotation]
+				if !hasGVK || ownerGVK != gvk.String() {
+					return
+				}
+
+				ownerName, hasName := wl.Annotations[constants.JobOwnerNameAnnotation]
+				if !hasName {
+					return
+				}
```
mimowo: Instead of using the annotations could we just verify the ownerReferences, if contain LWS and get the name of the owner for the found ownerReference.

The reason is prectical to enable easy backporting to 0.15 which does not yet have the annotations introduced for MultiKueue case, where the PrebuildWorkload may not yet have the LWS created. However, the machinery does not seem needed here.

The PR for reference: https://github.com/kubernetes-sigs/kueue/pull/8670 is only 0.16+. Also, we may revisit mutability of the annotations. IIRC they are mutable what makes them potentially tricky in some corner cases.

sohankunkerkar: Makes sense! The ownerReferences are simpler, immutable, and backportable to 0.15. The workload always has one pointing to the LWS in this flow since the reconciler sets it  via SetControllerReference.

PannagaRao: I'll switch to using ownerReferences instead of annotations. Thanks for pointing this out.

sohankunkerkar: @mimowo While I support this change, just wondering if we really need to fix it bc we are anyway gonna deprecate 0.15 with 0.17 releasing next week, right?

mimowo: Well, "really need" is a hard statement. We could be elastic, but we still have some users on 0.15 so if not much effort, then it is preferred. 

Basically, the idea is to release the patches for 0.15 with all patches just before dropping the support, so that users who need to stay on 0.15 for a little longer will have all the fixes we know about.

PR #9647, opened 2026-03-03
Add an option to E2E_MODE=dev to install a released version of Kueue

### LINKED ISSUE #9415: Add an option to E2E_MODE=dev to install a released version of Kueue

I would like to be able to skip rebuilding of Kueue and install it from one of the released versions.

**Why is this needed**:

To support cases like this: https://github.com/kubernetes-sigs/kueue/pull/9308#issuecomment-3932748704

vladikkuzn:
#### What type of PR is this?
/kind documentation

Fixes #9415

```release-note
NONE
```

### DIFF
--- a/hack/testing/e2e-common.sh
+++ b/hack/testing/e2e-common.sh
@@ -374,6 +374,15 @@ function prepare_docker_images {
     docker tag "$E2E_TEST_AGNHOST_IMAGE_OLD_WITH_SHA" "$E2E_TEST_AGNHOST_IMAGE_OLD"
     docker tag "$E2E_TEST_AGNHOST_IMAGE_WITH_SHA" "$E2E_TEST_AGNHOST_IMAGE"
 
+    # When using a pre-built Kueue image (released or staging), ensure it's available for kind load.
+    # Check remote first; if not found remotely, use local image if present; otherwise error.
+    if docker manifest inspect "$IMAGE_TAG" >/dev/null 2>&1; then
+        docker pull "$IMAGE_TAG"
+    elif ! docker image inspect "$IMAGE_TAG" >/dev/null 2>&1; then
+        echo "ERROR: Image '$IMAGE_TAG' not found remotely or locally." >&2
+        return 1
+    fi
+
     if [[ -n ${APPWRAPPER_VERSION:-} && ("$GINKGO_ARGS" =~ feature:(appwrapper|managejobswithoutqueuename) || ! "$GINKGO_ARGS" =~ "--label-filter") ]]; then
         docker pull "${APPWRAPPER_IMAGE}"
     fi
@@ -937,7 +946,7 @@ function install_sparkoperator {
             else
                 echo "Spark operator already present; upgrading."
             fi
-        else 
+        else
             echo "Spark operator helm release not found; installing."
         fi
     fi
--- a/site/content/en/docs/contribution_guidelines/testing.md
+++ b/site/content/en/docs/contribution_guidelines/testing.md
@@ -138,6 +138,24 @@ E2E_MODE=dev E2E_SKIP_REINSTALL=true make kind-image-build test-e2e
 E2E_MODE=dev E2E_SKIP_REINSTALL=true make kind-image-build test-multikueue-e2e
 ```
 
+To use a **released** or **staging** Kueue image instead of building from source (no `kind-image-build` needed), pass `IMAGE_TAG` with the desired image:
+
+```shell
+# Released version
+E2E_MODE=dev IMAGE_TAG=registry.k8s.io/kueue/kueue:v0.16.0 make test-e2e
+E2E_MODE=dev IMAGE_TAG=registry.k8s.io/kueue/kueue:v0.16.0 make test-multikueue-e2e
+
+# Staging image (e.g. from a PR or nightly)
+E2E_MODE=dev IMAGE_TAG=us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue:main make test-e2e
+```
+
+**Using a released version with matching manifests:** The e2e framework deploys CRDs and other resources from the repo's config and overrides only the controller image via `IMAGE_TAG`. To run e2e against a specific release with manifests that match that image:
+
+1. Check out that version's tag (e.g. `git checkout v0.16.0`). The CRD and deployment config in the repo are committed at each release, so no `make manifests` step is needed.
+2. Run the command above with the same image tag, e.g. `E2E_MODE=dev IMAGE_TAG=registry.k8s.io/kueue/kueue:v0.16.0 make test-e2e`.
+
+This is useful to reproduce issues on a specific released version (e.g. for on-call debugging). For installing a released version into a real cluster (not e2e), see [Install a released version](/docs/installation/#install-a-released-version).
+
 {{% alert title="Note" color="primary" %}}
 When reusing a kept cluster in `E2E_MODE=dev`, external operators (MPI, KubeRay, etc.) are installed only once.
 To force re-installing them on every run, set `E2E_ENFORCE_OPERATOR_UPDATE=true`.
--- a/site/content/en/docs/tasks/dev/setup_multikueue_development_environment.md
+++ b/site/content/en/docs/tasks/dev/setup_multikueue_development_environment.md
@@ -20,7 +20,13 @@ For example:
 E2E_MODE=dev make kind-image-build test-multikueue-e2e
 ```
 
-For more information about the DEV mode, refer to the testing documentation.
+To use a staging Kueue image without building (no `kind-image-build` needed), pass `IMAGE_TAG` with the `main` tag:
+
+```sh
+E2E_MODE=dev IMAGE_TAG=us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue:main make test-multikueue-e2e
+```
+
+For using a released version (with matching manifests) and for more information about the DEV mode, refer to the [testing documentation](/docs/contribution_guidelines/testing/#dev-mode-recommended).
 
 ## Setup MultiKueue with TAS
 
--- a/site/content/zh-CN/docs/contribution_guidelines/testing.md
+++ b/site/content/zh-CN/docs/contribution_guidelines/testing.md
@@ -125,14 +125,32 @@ func TestValidateClusterQueue(t *testing.T) {
 
 ```shell
 # 若不存在则创建；若已存在则复用。构建镜像、运行测试，并保留集群。
-E2E_MODE=dev make test-e2e
+E2E_MODE=dev make kind-image-build test-e2e
 
 # MultiKueue 的 dev 模式
-E2E_MODE=dev make test-multikueue-e2e
+E2E_MODE=dev make kind-image-build test-multikueue-e2e
 
 # 循环运行（直到失败），同时保留集群
-E2E_MODE=dev GINKGO_ARGS="--until-it-fails" make test-e2e
+E2E_MODE=dev GINKGO_ARGS="--until-it-fails" make kind-image-build  test-e2e
 ```
+
+若希望使用**已发布**或**预发布（staging）**的 Kueue 镜像而非从源码构建（无需执行 `kind-image-build`），可传入 `IMAGE_TAG` 并指定所需镜像：
+
+```shell
+# 已发布版本
+E2E_MODE=dev IMAGE_TAG=registry.k8s.io/kueue/kueue:v0.16.0 make test-e2e
+E2E_MODE=dev IMAGE_TAG=registry.k8s.io/kueue/kueue:v0.16.0 make test-multikueue-e2e
+
+# 预发布镜像（例如来自 PR 或每日构建）
+E2E_MODE=dev IMAGE_TAG=us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue:main make test-e2e
+```
+
+**使用与 manifest 匹配的已发布版本：** e2e 框架从仓库的 config 部署 CRD 等资源，仅通过 `IMAGE_TAG` 覆盖控制器镜像。若要对某一发布版本运行 e2e 且使用与该镜像匹配的 manifest：
+
+1. 检出该版本的 tag（例如 `git checkout v0.16.0`）。仓库中的 CRD 与部署配置在每个版本中均已提交，无需执行 `make manifests`。
+2. 使用相同镜像 tag 运行上述命令，例如 `E2E_MODE=dev IMAGE_TAG=registry.k8s.io/kueue/kueue:v0.16.0 make test-e2e`。
+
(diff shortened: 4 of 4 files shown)


Makefile-test.mk
```diff
@@ -51,6 +51,11 @@ PROMETHEUS_OPERATOR_VERSION ?= v0.89.0
 # When truthy, force re-installing external operators (MPI, Ray, etc.) on each run, even in E2E_MODE=dev.
 E2E_ENFORCE_OPERATOR_UPDATE ?= false
 
+# When set (e.g. v0.16.0), use a released Kueue image instead of building. Skip kind-image-build when using this.
+ifdef E2E_KUEUE_VERSION
+IMAGE_TAG := registry.k8s.io/kueue/kueue:$(E2E_KUEUE_VERSION)
+endif
```
mimowo: I'm wondering wouldn't this already work if someone passed the `IMAGE_TAG` parameter as `registry.k8s.io/kueue/kueue:v0.16.0`? If this already would work, maybe we don't need a new option, just a bit more of "education" how to pass existing params, wdyt?

cc @mbobrovskyi

mbobrovskyi: Yeah, I think this is a good idea. But what if we want to pull staging image – for example, to test a specific commit?
I’m wondering if we could check whether the image already exists localy and only try to pull it if it doesn’t. Something like:

```
if ! docker image inspect "$IMAGE_TAG" > /dev/null 2>&1; then
    docker pull "$IMAGE_TAG"
fi
```

@mimowo @vladikkuzn WDYT?

mimowo: > Yeah, I think this is a good idea. 

Cool, so if that works, let's try to focus on updating the Contribution guide with the usage example.

> I’m wondering if we could check whether the image already exists localy and only try to pull it if it doesn’t. Something like:

Maybe, but wouldn't the docker pull command already check that internally? This seems like unrelated improvement.

hack/testing/e2e-common.sh
```diff
@@ -364,6 +364,13 @@ function prepare_docker_images {
     docker tag "$E2E_TEST_AGNHOST_IMAGE_OLD_WITH_SHA" "$E2E_TEST_AGNHOST_IMAGE_OLD"
     docker tag "$E2E_TEST_AGNHOST_IMAGE_WITH_SHA" "$E2E_TEST_AGNHOST_IMAGE"
 
+    # When using a released Kueue image, pull it so it's available for kind load (only if not already present locally).
+    if [[ "${IMAGE_TAG}" == registry.k8s.io/kueue/kueue:* ]]; then
```
mbobrovskyi: Can we avoid this if? I’m wondering if we can use `us-central1-docker.pkg.dev/k8s-staging-images/kueue` as well.

mimowo: The ability of using both released and staging would be great

mbobrovskyi: For example, 

IMAGE_TAG=us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue:main

vladikkuzn: Are staging images "us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue" changing? I mean can someone retag them? Is `docker image inspect` needed?

mimowo: > Is docker image inspect needed?

Yeah, I see no need, the docker pull should already check if present or not.

mbobrovskyi: > Are staging images "us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue" changing? I mean can someone retag them?

Yes, the main tag is retagged every time, so we do need docker pull.

The problem is that IMAGE_TAG is used locally after `kueue-image-build` (e.g., `us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue:v0.17.0-devel-310-g5fa614825-dirty`) and this tag no need to pull.

Maybe we can just run docker pull and ignore the error if the image doesn’t exist.

mimowo: is it the logic we want to have? We can probably improve it by skipping the remote check for staging tags. 

```sh
echo "Checking remote image: $IMAGE_TAG"

# Check if the image exists remotely
if docker manifest inspect "$IMAGE_TAG" >/dev/null 2>&1; then
  echo "Remote image found. Pulling..."
  docker pull "$IMAGE_TAG"
  exit 0
fi

echo "Remote image not found. Checking local image..."

# Check if the image exists locally
if docker image inspect "$IMAGE_TAG" >/dev/null 2>&1; then
  echo "Using local image: $IMAGE_TAG"
  exit 0
fi

echo "ERROR: Image '$IMAGE_TAG' not found remotely or locally."
```

vladikkuzn: Does this look good? Or is if (`if [[ "${IMAGE_TAG}" == registry.k8s.io/kueue/kueue:* ]]`) unnecessary?
```sh
if [[ "${IMAGE_TAG}" == registry.k8s.io/kueue/kueue:* ]]; then
        if docker manifest inspect "$IMAGE_TAG" >/dev/null 2>&1; then
            docker pull "$IMAGE_TAG"
        elif ! docker image inspect "$IMAGE_TAG" >/dev/null 2>&1; then
            echo "ERROR: Image '$IMAGE_TAG' not found remotely or locally." >&2
            return 1
        fi
    fi
```

mbobrovskyi: Yes, it’s unnecessary because we also need to cover `us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue`.

site/content/en/docs/contribution_guidelines/testing.md
```diff
@@ -134,6 +134,13 @@ E2E_MODE=dev make kind-image-build test-multikueue-e2e
 E2E_MODE=dev GINKGO_ARGS="--until-it-fails" make kind-image-build  test-e2e
 ```
 
+To use a **released** Kueue version instead of building from source (no `kind-image-build` needed), pass `IMAGE_TAG` with the desired image (e.g. `registry.k8s.io/kueue/kueue:v0.16.0`):
+
+```shell
+E2E_MODE=dev IMAGE_TAG=registry.k8s.io/kueue/kueue:v0.16.0 make test-e2e
+E2E_MODE=dev IMAGE_TAG=registry.k8s.io/kueue/kueue:v0.16.0 make test-multikueue-e2e
```
mbobrovskyi, line 146:
```suggestion
E2E_MODE=dev IMAGE_TAG=us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue:v0.17.0-devel-332-g701268f80 make test-multikueue-e2e
```
Let's add an example with staging image.

site/content/zh-CN/docs/contribution_guidelines/testing.md
```diff
@@ -134,6 +134,13 @@ E2E_MODE=dev make test-multikueue-e2e
 E2E_MODE=dev GINKGO_ARGS="--until-it-fails" make test-e2e
 ```
 
+若希望使用**已发布**的 Kueue 版本而非从源码构建（无需执行 `kind-image-build`），可传入 `IMAGE_TAG` 并指定所需镜像（例如 `registry.k8s.io/kueue/kueue:v0.16.0`）：
+
+```shell
+E2E_MODE=dev IMAGE_TAG=registry.k8s.io/kueue/kueue:v0.16.0 make test-e2e
+E2E_MODE=dev IMAGE_TAG=registry.k8s.io/kueue/kueue:v0.16.0 make test-multikueue-e2e
```
mbobrovskyi, line 142:
```suggestion
E2E_MODE=dev IMAGE_TAG=us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue:v0.17.0-devel-332-g701268f80 make test-multikueue-e2e
```

mbobrovskyi (commented):
Overall LGTM.

site/content/en/docs/contribution_guidelines/testing.md
mimowo: Hm, why staging example? I think this option will mostly be used with released images actually. Also, for released images we can easier bump the version in the `prepare-release-branch` script by simple string replacements. If we use staging we will need to maintain manually. Unless using staging which points to "main" tag.

site/content/zh-CN/docs/contribution_guidelines/testing.md
mimowo: Same comment as in https://github.com/kubernetes-sigs/kueue/pull/9647#discussion_r2888560095, I would rather use released image but fix the `prepare-release-branch` script.

site/content/en/docs/contribution_guidelines/testing.md
mbobrovskyi: > Hm, why staging example? I think this option will mostly be used with released images actually. 

Hmm. I think it's good just for example. What if we set `MAGE_TAG=us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue:main make test-multikueue-e2e`? WDYT?

mbobrovskyi: > bump the version in the prepare-release-branch script by simple string replacements

This is good point! I think no need to bump in prepare-release-branch. We can use this syntax:

```
E2E_MODE=dev IMAGE_TAG=registry.k8s.io/kueue/kueue:{{< param "version" >}} make test-multikueue-e2e
```

mimowo: I'm good with the example using main. 

Ah, I just realized the PR will not actually support well older released versions - because we would still use the lastest artifacts, right? 

The idea for supporting released images is that some people are involved in oncall, and need to reproduce problems reported by users. Users report problems for the released version, not main, and these sometimes are very different. So being able to spin up quickly a cluster which matches the version used by a user is nice,  but for that we need more - also the correct manifests. 

For that I guess we need to still checkout the tagged version, build manifests, and only then follow the E2E_MODE command, right? It would be nice to also describe this in the guide.

site/content/en/docs/contribution_guidelines/testing.md
```diff
@@ -134,6 +134,21 @@ E2E_MODE=dev make kind-image-build test-multikueue-e2e
...
 
+To use a **released** or **staging** Kueue image instead of building from source (no `kind-image-build` needed), pass `IMAGE_TAG` with the desired image:
+
+```shell
+# Released version
+E2E_MODE=dev IMAGE_TAG=registry.k8s.io/kueue/kueue:v0.16.0 make test-e2e
+E2E_MODE=dev IMAGE_TAG=registry.k8s.io/kueue/kueue:v0.16.0 make test-multikueue-e2e
+
+# Staging image (e.g. from a PR or nightly)
+E2E_MODE=dev IMAGE_TAG=us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue:v0.17.0-devel-332-g701268f80 make test-e2e
```
mbobrovskyi:
```suggestion
E2E_MODE=dev IMAGE_TAG=us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue:main make test-e2e
```

As we discussed here: https://github.com/kubernetes-sigs/kueue/pull/9647#discussion_r2888579353, we decided to use the main tag. Could you please address it?

mbobrovskyi:
@vladikkuzn could you please address https://github.com/kubernetes-sigs/kueue/pull/9647#discussion_r2888609060?

site/content/en/docs/tasks/dev/setup_multikueue_development_environment.md
```diff
@@ -20,7 +20,15 @@ For example:
 E2E_MODE=dev make kind-image-build test-multikueue-e2e
 ```
 
-For more information about the DEV mode, refer to the testing documentation.
+To use a released or staging Kueue image without building (no `kind-image-build` needed), pass `IMAGE_TAG`:
+```sh
+# Released
+E2E_MODE=dev IMAGE_TAG=registry.k8s.io/kueue/kueue:v0.16.0 make test-multikueue-e2e
+# Staging (e.g. from a PR or nightly)
+E2E_MODE=dev IMAGE_TAG=us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue:v0.17.0-devel-332-g701268f80 make test-multikueue-e2e
```
mimowo:
```suggestion
E2E_MODE=dev IMAGE_TAG=us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue:main make test-multikueue-e2e
```
Let's build examples using main, using some custom builds will not age well.

site/content/zh-CN/docs/contribution_guidelines/testing.md
```diff
@@ -134,6 +134,21 @@ E2E_MODE=dev make test-multikueue-e2e
 E2E_MODE=dev GINKGO_ARGS="--until-it-fails" make test-e2e
 ```
 
+若希望使用**已发布**或**预发布（staging）**的 Kueue 镜像而非从源码构建（无需执行 `kind-image-build`），可传入 `IMAGE_TAG` 并指定所需镜像：
```
mimowo, line 137: It looks out of sync. Please drop, we have @windsonsea @samzong and @my-git9 who will be happy to translate

site/content/en/docs/tasks/dev/setup_multikueue_development_environment.md
mimowo: Maybe let's only mention staging "main" here, using released version is more involving becuase of the manifests. let's link to the contribution guide

site/content/en/docs/contribution_guidelines/testing.md
```diff
@@ -132,12 +132,26 @@ E2E_MODE=dev make kind-image-build test-multikueue-e2e
...
-E2E_MODE=dev E2E_SKIP_REINSTALL=true make kind-image-build test-e2e
-E2E_MODE=dev E2E_SKIP_REINSTALL=true make kind-image-build test-multikueue-e2e
+# Staging image (e.g. from a PR or nightly)
+E2E_MODE=dev IMAGE_TAG=us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue:main make test-e2e
 ```
 
+**Using a released version with matching manifests:** The e2e framework deploys CRDs and other resources from the repo's config and overrides only the controller image via `IMAGE_TAG`. To run e2e against a specific release with manifests that match that image:
+
+1. Check out that version's tag (e.g. `git checkout v0.16.0`).
+2. Run the command above with the same image tag, e.g. `E2E_MODE=dev IMAGE_TAG=registry.k8s.io/kueue/kueue:v0.16.0 make test-e2e`.
```
mimowo: is it enough to just checkout? I think we should also call `make manifests`? please check, but it might be needed, because iirc the final manifests might not be git versioned, but maybe kustomize is used?

vladikkuzn: The CRD and deployment config are committed at each release, so no make manifests step is needed

mimowo:
/approve
/cherrypick website
Thank you 👍

PR #9651, opened 2026-03-03
Forcefully remove stuck pods in the failure recovery controller

### LINKED ISSUE #9649: FailureRecovery: Pod deletion gets stuck even if FailureRecovery is used
**What happened**:

We have a case when Pod FailureRecovery is not helping, and Pod deletion is stuck. This is the order of events:

1. kubelet is disabled (eg. node crashed)
2. Job controller sends a graceful Delete to the Pod
3. Kueue's FailureRecovery controller sets status.phase=Failed
Issue: the Pod remains on the API server indefinitely. 

In particular, this blocks deletion of the Job in cascade=foreground mode which is used by JobSet: https://github.com/kubernetes-sigs/jobset/blob/e38bc265102d7dd9a68de38ff909246e261d479d/pkg/controllers/jobset_controller.go#L695-L703

I expect the FailureRecovery controller to handle the Pod in such a way that the Job's deletion with the cascade=foreground mode not to be blocked by such a Pod.

I think force deleting the Pod is reasonable, and this is what the k8s GC controller is doing: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/podgc/gc_controller.go#L365

1. setup a kind cluster with 2 nodes
2. create a Job like this:
```yaml
apiVersion: batch/v1
kind: Job
metadata:
  generateName: sample-job-
  namespace: default
  labels:
    kueue.x-k8s.io/queue-name: user-queue
spec:
  parallelism: 1
  completions: 1
  podReplacementPolicy: Failed # Replace pods only if they are terminated
  template:
    metadata:
      annotations:
        kueue.x-k8s.io/safe-to-forcefully-terminate: "true" # Opt-in to failure recovery
    spec:
      affinity:
        podAntiAffinity:
          # Schedule on a different node than kueue-controller-manager
          preferredDuringSchedulingIgnoredDuringExecution:
          - weight: 1
            podAffinityTerm:
              labelSelector:
                matchExpressions:
                - key: app.kubernetes.io/name
                  operator: In
                  values:
                  - kueue
              topologyKey: kubernetes.io/hostname
              namespaceSelector: {}
      containers:
      - name: dummy-job
        image: registry.k8s.io/e2e-test-images/agnhost:2.53
        command: [ "/bin/sh" ]
        args: [ "-c", "sleep 300" ]
        resources:
          requests:
            cpu: "1"
            memory: "200Mi"
      restartPolicy: Never
```
3. stop kubelet on the node by `docker exec -it kind-worker systemctl stop kubelet`
4. delete manually the Pod, by `kubectl delete pod/<pod name>`
5. delete the Job with `kubectl delete job/<job-name> --cascade=foreground` 
Issue: the job deletion is blocked on the Pod's deletion.

kshalot:
#### What type of PR is this?
/kind bug

Fixes #9649

```release-note
FailureRecoveryPolicy: forcefully delete stuck pods (without grace period) in addition to transitioning them
to the `Failed` phase. This fixes a scenario where foreground propagating deletions were blocked by a stuck pod.
```

### DIFF
--- a/test/e2e/customconfigs/failure_recovery_policy_test.go
+++ b/test/e2e/customconfigs/failure_recovery_policy_test.go
@@ -117,47 +117,58 @@ var _ = ginkgo.Describe("Failure Recovery Policy", ginkgo.Ordered, ginkgo.Contin
 		)
 
 		ginkgo.BeforeEach(func() {
-			cq = utiltestingapi.MakeClusterQueue("cq").
-				ResourceGroup(*utiltestingapi.MakeFlavorQuotas(rf.Name).
-					Resource(corev1.ResourceCPU, "8").
-					Resource(corev1.ResourceMemory, "36G").
-					Obj()).
-				Obj()
-			util.CreateClusterQueuesAndWaitForActive(ctx, k8sClient, cq)
+			ginkgo.By("creating the cluster queue and local queue", func() {
+				cq = utiltestingapi.MakeClusterQueue("cq").
+					ResourceGroup(*utiltestingapi.MakeFlavorQuotas(rf.Name).
+						Resource(corev1.ResourceCPU, "8").
+						Resource(corev1.ResourceMemory, "36G").
+						Obj()).
+					Obj()
+				util.CreateClusterQueuesAndWaitForActive(ctx, k8sClient, cq)
 
-			lq = utiltestingapi.MakeLocalQueue("lq", ns.Name).
-				ClusterQueue(cq.Name).
-				Obj()
-			util.CreateLocalQueuesAndWaitForActive(ctx, k8sClient, lq)
+				lq = utiltestingapi.MakeLocalQueue("lq", ns.Name).
+					ClusterQueue(cq.Name).
+					Obj()
+				util.CreateLocalQueuesAndWaitForActive(ctx, k8sClient, lq)
+			})
 
-			util.MustCreate(ctx, k8sClient, job)
-			gomega.Eventually(func(g gomega.Gomega) {
-				g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(job), job)).To(gomega.Succeed())
-				g.Expect(*job.Spec.Suspend).To(gomega.BeFalse())
-				g.Expect(job.Status.Active).To(gomega.Equal(int32(1)))
-				g.Expect(job.Status.Ready).To(gomega.Equal(ptr.To(int32(1))))
-			}, util.LongTimeout, util.Interval).Should(gomega.Succeed())
+			ginkgo.By("creating the job and waiting for it to start", func() {
+				util.MustCreate(ctx, k8sClient, job)
+				gomega.Eventually(func(g gomega.Gomega) {
+					g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(job), job)).To(gomega.Succeed())
+					g.Expect(*job.Spec.Suspend).To(gomega.BeFalse())
+					g.Expect(job.Status.Active).To(gomega.Equal(int32(1)))
+					g.Expect(job.Status.Ready).To(gomega.Equal(ptr.To(int32(1))))
+				}, util.LongTimeout, util.Interval).Should(gomega.Succeed())
+			})
 
-			gomega.Eventually(func(g gomega.Gomega) {
-				pods := &corev1.PodList{}
-				g.Expect(k8sClient.List(ctx, pods, client.InNamespace(ns.Name), client.MatchingLabels(job.Spec.Selector.MatchLabels))).To(gomega.Succeed())
-				g.Expect(pods.Items).To(gomega.HaveLen(1))
+			ginkgo.By("ensuring the pod is scheduled on a worker node", func() {
+				gomega.Eventually(func(g gomega.Gomega) {
+					pods := &corev1.PodList{}
+					g.Expect(k8sClient.List(ctx, pods, client.InNamespace(ns.Name), client.MatchingLabels(job.Spec.Selector.MatchLabels))).To(gomega.Succeed())
+					g.Expect(pods.Items).To(gomega.HaveLen(1))
 
-				pod = &pods.Items[0]
-				nodeName = pod.Spec.NodeName
-				g.Expect(nodeName).ToNot(gomega.BeEmpty())
-			}, util.Timeout, util.Interval).Should(gomega.Succeed())
+					pod = &pods.Items[0]
+					nodeName = pod.Spec.NodeName
+					g.Expect(nodeName).ToNot(gomega.BeEmpty())
+				}, util.Timeout, util.Interval).Should(gomega.Succeed())
+			})
 
-			// Stop kubelet on the node
-			cmd := exec.Command("docker", "exec", nodeName, "systemctl", "stop", "kubelet")
-			gomega.Expect(cmd.Run()).To(gomega.Succeed())
+			ginkgo.By("stopping the kubelet on the node running the pod", func() {
+				cmd := exec.Command("docker", "exec", nodeName, "systemctl", "stop", "kubelet")
+				gomega.Expect(cmd.Run()).To(gomega.Succeed())
+			})
 		})
 
 		ginkgo.AfterEach(func() {
-			cmd := exec.Command("docker", "exec", nodeName, "systemctl", "start", "kubelet")
-			gomega.Expect(cmd.Run()).To(gomega.Succeed())
+			ginkgo.By("starting the kubelet on the node running the pod", func() {
+				cmd := exec.Command("docker", "exec", nodeName, "systemctl", "start", "kubelet")
+				gomega.Expect(cmd.Run()).To(gomega.Succeed())
+			})
 
-			util.ExpectObjectToBeDeleted(ctx, k8sClient, cq, true)
+			ginkgo.By("deleting the cluster queue", func() {
+				util.ExpectObjectToBeDeleted(ctx, k8sClient, cq, true)
+			})
 		})
 
 		ginkgo.It("should delete pods running on an unreachable node", func() {


pkg/controller/failurerecovery/pod_termination_controller.go
```diff
@@ -191,6 +191,11 @@ func (r *TerminatingPodReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 	r.recorder.Event(pod, corev1.EventTypeWarning, KueueForcefulTerminationReason, eventMessage)
 
+	// Forcefully delete the pod object
+	if err = r.client.Delete(ctx, pod, &client.DeleteOptions{GracePeriodSeconds: ptr.To(int64(0))}); err != nil {
```
mimowo, line 195: I think this is very unlikely, but if this request fails, say temporarily unavailable control plane, then we should re-try. The Reconcile will be retried, but it will skip currently on the `podEligibleForTermination` check. So, I think I may need to either relax the check or delete. 

This is my thinking:
1. relaxing should be safe, the Pods have the `kueue.x-k8s.io/safe-to-forcefully-terminate` annotation anyway
2. we could say 

```golang
	if utilpod.IsTerminated(p) && (objectMeta.GetDeletionGracePeriodSeconds() == nil || *objectMeta.GetDeletionGracePeriodSeconds() == 0).{
		return false
	}
```
following the check in API server code https://github.com/kubernetes/kubernetes/blob/b9e0f9d10ae94665d6fc1817f7310672a905ffd0/staging/src/k8s.io/apiserver/pkg/registry/rest/delete.go#L120-L122, because in this case API server will delete the Pod, even if Kubelet is down.

I'm not sure the optimization is worth the complication though.

pkg/controller/failurerecovery/pod_termination_controller.go
```diff
@@ -42,7 +42,7 @@ import (
 	utiltaints "sigs.k8s.io/kueue/pkg/util/taints"
 )
 
-// +kubebuilder:rbac:groups="",resources=pods,verbs=get;list;watch
+// +kubebuilder:rbac:groups="",resources=pods,verbs=get;list;watch;delete
```
mimowo, line 45: Probably need to re-generate manifests

kshalot: Actually, I thought I missed some command but it seems not.

pkg/controller/failurerecovery/pod_termination_controller.go
kshalot: I opted for a simple change and allowed `Failed` pods to be reconciled. In practice, it shouldn't be a big deal since we expect the pods to be deleted anyway.

mimowo: sgtm

test/e2e/customconfigs/failure_recovery_policy_test.go
```diff
@@ -0,0 +1,166 @@
...
+	"sigs.k8s.io/kueue/pkg/controller/constants"
+	"sigs.k8s.io/kueue/pkg/features"
+	utiltestingapi "sigs.k8s.io/kueue/pkg/util/testing/v1beta2"
+	testingjob "sigs.k8s.io/kueue/pkg/util/testingjobs/job"
+	"sigs.k8s.io/kueue/test/util"
+)
+
+const (
+	// `node-monitor-grace-period` + `podToleration` + `deletionGracePeriodSeconds` + `forcefulTerminationGracePeriod` + buffer
+	forcefulTerminationCheckTimeout = (50 + 1 + 30 + 60 + 10) * time.Second
```
mimowo: nit, let's compute the base value here, and the buffer let's do as forcefulTerminationCheckTimeout + util.LongTimeout.

test/e2e/customconfigs/failure_recovery_policy_test.go
```diff
@@ -0,0 +1,166 @@
...
+
+			// Stop kubelet on the node
+			cmd := exec.Command("docker", "exec", nodeName, "systemctl", "stop", "kubelet")
+			gomega.Expect(cmd.Run()).To(gomega.Succeed())
+
+			// Defer restart of kubelet
+			ginkgo.DeferCleanup(func() {
+				cmd := exec.Command("docker", "exec", nodeName, "systemctl", "start", "kubelet")
+				gomega.Expect(cmd.Run()).To(gomega.Succeed())
+			})
```
mimowo: nit, I would rather keep the nodeName as a var, as lq, cq, and then restart kubelet in the AfterEech. Similarly as we do for LQ / CQ. this will avoid repeating he pattern. Disabling kubelet will be the common denominator of the test cases here.

test/e2e/customconfigs/failure_recovery_policy_test.go
```diff
@@ -0,0 +1,166 @@
...
+			gomega.Eventually(func(g gomega.Gomega) {
+				g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(job), job)).To(gomega.Succeed())
+				g.Expect(job.Status.Active).To(gomega.Equal(int32(1)))
+				g.Expect(job.Status.Ready).To(gomega.Equal(ptr.To(int32(1))))
+			}, util.LongTimeout, util.Interval).Should(gomega.Succeed())
+
+			// Get the pod
+			pods := &corev1.PodList{}
+			gomega.Expect(k8sClient.List(ctx, pods, client.InNamespace(ns.Name), client.MatchingLabels(job.Spec.Selector.MatchLabels))).To(gomega.Succeed())
+			gomega.Expect(pods.Items).To(gomega.HaveLen(1))
```
mimowo: wrap this in Eventually. Otherwise it may flake, because there is no guarantee in k8s on the delivery order of "events" across different Kinds. Even if Pod is created first, and later the Job status is updated, it is possible the informer in the test code client will get notification about the Pod before the Job status update.

test/e2e/customconfigs/failure_recovery_policy_test.go
kshalot: I changed so that it stops the kubelet in`BeforeEach` and starts it in `AfterEach`. made the most sense to me. WDYT?

mimowo: sgtm

test/e2e/customconfigs/failure_recovery_policy_test.go
```diff
@@ -0,0 +1,173 @@
...
+	testingjob "sigs.k8s.io/kueue/pkg/util/testingjobs/job"
+	"sigs.k8s.io/kueue/test/util"
+)
+
+const (
+	// `node-monitor-grace-period` + `podToleration` + `deletionGracePeriodSeconds` + `forcefulTerminationGracePeriod`
+	forcefulTerminationCheckTimeout = (50 + 1 + 30 + 60) * time.Second
+)
+
+var _ = ginkgo.FDescribe("Failure Recovery Policy", ginkgo.Ordered, ginkgo.ContinueOnFailure, func() {
```
tenzen-y:
```suggestion
var _ = ginkgo.Describe("Failure Recovery Policy", ginkgo.Ordered, ginkgo.ContinueOnFailure, func() {
```

nit

test/e2e/customconfigs/failure_recovery_policy_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+	ginkgo.AfterEach(func() {
+		util.ExpectAllPodsInNamespaceDeleted(ctx, k8sClient, ns)
+	})
+
+	ginkgo.AfterAll(func() {
+		util.ExpectObjectToBeDeleted(ctx, k8sClient, rf, true)
+	})
+
+	ginkgo.When("the kubelet on a node goes down goes down", func() {
```
tenzen-y:
```suggestion
	ginkgo.When("the kubelet on a node goes down", func() {
```

nits

pkg/controller/failurerecovery/pod_termination_controller.go
```diff
@@ -191,6 +190,11 @@ func (r *TerminatingPodReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 	r.recorder.Event(pod, corev1.EventTypeWarning, KueueForcefulTerminationReason, eventMessage)
 
+	// Forcefully delete the pod object
+	if err = r.client.Delete(ctx, pod, &client.DeleteOptions{GracePeriodSeconds: ptr.To(int64(0))}); err != nil {
+		return ctrl.Result{}, err
```
tenzen-y:
```suggestion
		return ctrl.Result{}, client.IgnoreNotFound(err)
```

Should we ignore not found error? Because this Pod might be deleted automatically during this reconciliation.

tenzen-y (commented):
I left the first comments.

pkg/controller/failurerecovery/pod_termination_controller.go
```diff
@@ -204,7 +208,9 @@ func podEligibleForTermination(p *corev1.Pod) bool {
 		return false
 	}
 
-	if utilpod.IsTerminated(p) {
+	// Do not filter out `corev1.PodFailed` to handle partial failure of the controller
+	// (pod marked as failed, but not deleted).
+	if p.Status.Phase == corev1.PodSucceeded {
```
tenzen-y: In addition, I think that we should check if this Pod has `KueueFailureRecovery` condition.
Why check only for the Pod Phase?

tenzen-y: To clarify the above my comment, my proposal was the following Pod status checking:

```go
	if <Pod doesn't have KueueFailureRecovery=True condition> {
		if utilpod.IsTerminated(p) {
			return false
		}
	} else {
		if p.Status.Phase == corev1.PodSucceeded { 
			return false
		}
	}
```

pkg/controller/failurerecovery/pod_termination_controller.go
kshalot: I'd say yes - applied.

pkg/controller/failurerecovery/pod_termination_controller.go
kshalot: We can do this. I opted for something pretty equivalent, PTAL.

The difference between the current version and your proposal is the lack of `p.Status.Phase == corev1.PodSucceeded` - I'm not sure what this case would mean to be honest (partial forceful deletion + pod succeeded). I just returned `true` if the condition is present.

pkg/util/pod/pod.go
```diff
@@ -35,6 +35,17 @@ func HasGate(pod *corev1.Pod, gateName string) bool {
...
 
+// HasCondition checks if there is a condition in the Pod's status
+// with exactly the same Type and Status.
+func HasCondition(p *corev1.Pod, cond *corev1.PodCondition) bool {
+	for _, statusCond := range p.Status.Conditions {
+		if statusCond.Type == cond.Type && statusCond.Status == cond.Status {
+			return true
+		}
+	}
+	return false
```
mimowo: nit, maybe let's modernize and use slices.ContainsFunc ?

pkg/controller/failurerecovery/pod_termination_controller.go
mimowo: Oh, I see, pods/delete was already there: https://github.com/kubernetes-sigs/kueue/blob/main/config/components/rbac/role.yaml#L31

This is good news for the ease of deploying as patch release.

test/e2e/customconfigs/failure_recovery_policy_test.go
```diff
@@ -0,0 +1,175 @@
...
+	testingjob "sigs.k8s.io/kueue/pkg/util/testingjobs/job"
+	"sigs.k8s.io/kueue/test/util"
+)
+
+const (
+	// Unresponsive node is marked as unreachable after the grace period.
+	nodeMonitorGracePeriod = 50 * time.Second
+
+	// `podToleration` + `deletionGracePeriodSeconds` + `forcefulTerminationGracePeriod`
+	unhealthyNodeforcefulTerminationCheckTimeout = (1 + 30 + 60) * time.Second
```
mimowo: Is the `deletionGracePeriodSeconds` relevant for the duration of the wait? I would suppose anyway the graceful termination is not happening in the tests

pkg/util/pod/pod.go
kshalot: Agreed. I copied the implementation from our other condition checks in the utils, but we can make this one better.

test/e2e/customconfigs/failure_recovery_policy_test.go
kshalot: The failure recovery controller relies on the `deletionGracePeriodSeconds`:
https://github.com/kubernetes-sigs/kueue/blob/116095b2e505386bf9837abaed36b89e899e51a6/pkg/controller/failurerecovery/pod_termination_controller.go#L153

The forceful termination threshold is relative to `DeletionTimestamp` which takes the grace period into account.

mimowo: i see, rhank you for explaining. However we should be able to shorten this to just 1s to make the tedt run faster, because kubelet is stopped anyway

mimowo: Also, I would prefer to move the constant down, as a var in the BeforeEach section. As it depends on the Pod's graceful termination period.

test/e2e/customconfigs/failure_recovery_policy_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+			util.ExpectObjectToBeDeleted(ctx, k8sClient, cq, true)
+		})
+
+		ginkgo.It("should delete pods running on an unreachable node", func() {
+			util.ExpectObjectToBeDeletedWithTimeout(ctx, k8sClient, pod, false, nodeMonitorGracePeriod+unhealthyNodeforcefulTerminationCheckTimeout+util.LongTimeout)
+		})
+
+		ginkgo.It("should unblock the stuck pod's parents that are being deleted with foreground propagation", func() {
+			// Pod is being terminated after toleration elapses
```
mimowo: nit, wrap the steps in ginkgo.It rather than adding comments, then the comments are also present in the test output when debugging.

test/e2e/customconfigs/failure_recovery_policy_test.go
kshalot: 👍

test/e2e/customconfigs/failure_recovery_policy_test.go
kshalot: I assume you meant `gingko.By`? I used that.

mimowo: yes, gingko.By

pkg/controller/failurerecovery/pod_termination_controller.go
```diff
@@ -191,24 +191,35 @@ func (r *TerminatingPodReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 		return false
 	}
 
-	if utilpod.IsTerminated(p) {
-		return false
-	}
+	// Do not filter out partially terminated (condition set, but not deleted) pods to handle
+	// partial executions of the controller caused by errors (for example network errors).
+	return isPodPartiallyForcefullyTerminated(p) || !utilpod.IsTerminated(p)
+}
```
tenzen-y, line 215: Yeah, this is what I imagined, thank you 👍

keps/6757-failure-recovery/README.md
tenzen-y, line 1: @kshalot Could you record the annotation name replacement in the `Implementation History` section?

kshalot: Yes, good point.

https://github.com/kubernetes-sigs/kueue/pull/9675

mimowo:
@kshalot could you prepare cherrypick for website?

PR #8761, opened 2026-01-23
[WIP] Handle `--kubeconfig` flag in visibility server

olekzabl:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

It makes the `--kubeconfig` flag, when passed into the Kueue manager binary, respected by the visibility server.

This addresses at least a part of #8606.
(I'm not sure at this point if there's more to do for that bug).

#### Does this PR introduce a user-facing change?

```release-note
OnDemandVisibility: Fix the bug that when running Kueue with the custom `--kubeconfig` flag the visibility server
fails to initialize, because the custom value of the flag is not propagated to it, leading to errors such as:
"Unable to create and start visibility server","error":"unable to apply VisibilityServerOptions: failed to get delegated authentication kubeconfig:  failed to get delegated authentication kubeconfig: ..."
```

PR #9652, opened 2026-03-03
Allow passing distinct api server flags

### LINKED ISSUE #9650: Visibility Server ignores APIServer flags, breaking out-of-cluster execution and custom port binding
**What happened**:

When attempting to run the `kueue-controller-manager` with the `VisibilityOnDemand` feature enabled and using a custom kubeconfig via the `--kubeconfig` flag, the visibility server fails to start.

It throws the following error:

```
Unable to create and start visibility server","error":"unable to apply VisibilityServerOptions: failed to get delegated authentication kubeconfig: failed to get delegated authentication kubeconfig: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
```

If KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT are mocked, the error changes to:

```
open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory
Because the visibility server does not parse the controller's flags or expose its own genericapiserver flags, parameters like the kubeconfig path for delegated authentication and the bind port (8082) are completely ignored or hardcoded.
```

The embedded visibility server should respect the flags passed to the controller manager (or expose standard APIServer flags like `--authentication-kubeconfig` and `--secure-port`). This would allow running the controller out-of-cluster for development/testing, and allow administrators to customize the visibility server's bind port.

Enable the `VisibilityOnDemand` feature gate.

Attempt to run the `kueue-controller-manager` locally (out-of-cluster) using a custom kubeconfig:

```Bash
./kueue-controller-manager --kubeconfig=/path/to/my/kubeconfig
```
Observe the startup crash from the visibility server attempting to load in-cluster credentials.

Following discussions in [#8761](https://github.com/kubernetes-sigs/kueue/pull/8761#issuecomment-3983891127) reusing `--kubeconfig` by default seems reasonable for simplicity of users, which is done in PR #9619. For other flags, a more systematic approach is necessary, which is what this bug is intended for.

Nilsachy:
#### What type of PR is this?

/kind bug

Allow to pass APIServer flags in the Visibility Server

Fixes #9650 

```release-note
VisibilityOnDemand: Introduce a new Kueue deployment argument, --visibility-server, which allows passing custom flags (such as a custom --secure-port) when starting the visibility server.
```

### DIFF
--- a/site/content/zh-CN/docs/tasks/manage/monitor_pending_workloads/pending_workloads_on_demand.md
+++ b/site/content/zh-CN/docs/tasks/manage/monitor_pending_workloads/pending_workloads_on_demand.md
@@ -76,7 +76,7 @@ If you disable the feature, you also need to remove the associated `APIService` 
 
 If you need to pass specific flags to the embedded visibility API server, you can use the `--visibility-server` flag when running the Kueue controller manager. For example: `--visibility-server "--secure-port=8443 --authentication-config=/path/to/kubeconfig"`. 
 
-Note that the arguments to the embedded visibility API server (e.g. the kubeconfig) do not support normal shell-like parsing and cannot contain quotes or whitespaces. Since internal quotes are not stripped, a value like `--authentication-config='/path/to/kubeconfig'` will result in an actual path of `'/path/to/kubeconfig'` (including the single quotes), which is not a typical file path format. Similarly, including whitespaces like `--authentication-config="/path/ with_whitespace /to/kubeconfig"` will cause the argument to be incorrectly split into multiple parts.
+Note that the arguments to the embedded visibility API server (e.g. the kubeconfig) do not support normal shell-like parsing and cannot contain quotes or whitespaces. Since internal quotes are not stripped, a value like `--authentication-config='/path/to/kubeconfig'` will result in an actual path of `'/path/to/kubeconfig'` (including the single quotes), which is not a typical file path format. Similarly, including whitespaces like `--authentication-config=/path/ with_whitespace /to/kubeconfig` will cause the argument to be incorrectly split into multiple parts.
 
 For a full list of available flags, please see the [kube-apiserver documentation](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/).
 
--- a/site/content/en/docs/tasks/manage/monitor_pending_workloads/pending_workloads_on_demand.md
+++ b/site/content/en/docs/tasks/manage/monitor_pending_workloads/pending_workloads_on_demand.md
@@ -76,7 +76,7 @@ If you disable the feature, you also need to remove the associated `APIService` 
 
 If you need to pass specific flags to the embedded visibility API server, you can use the `--visibility-server` flag when running the Kueue controller manager. For example: `--visibility-server "--secure-port=8443 --authentication-config=/path/to/kubeconfig"`. 
 
-Note that the arguments to the embedded visibility API server (e.g. the kubeconfig) do not support normal shell-like parsing and cannot contain quotes or whitespaces. Since internal quotes are not stripped, a value like `--authentication-config='/path/to/kubeconfig'` will result in an actual path of `'/path/to/kubeconfig'` (including the single quotes), which is not a typical file path format. Similarly, including whitespaces like `--authentication-config="/path/ with_whitespace /to/kubeconfig"` will cause the argument to be incorrectly split into multiple parts.
+Note that the arguments to the embedded visibility API server (e.g. the kubeconfig) do not support normal shell-like parsing and cannot contain quotes or whitespaces. Since internal quotes are not stripped, a value like `--authentication-config='/path/to/kubeconfig'` will result in an actual path of `'/path/to/kubeconfig'` (including the single quotes), which is not a typical file path format. Similarly, including whitespaces like `--authentication-config=/path/ with_whitespace /to/kubeconfig` will cause the argument to be incorrectly split into multiple parts.
 
 For a full list of available flags, please see the [kube-apiserver documentation](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/).
 


pkg/visibility/server.go
```diff
@@ -96,13 +98,24 @@ func CreateAndStartVisibilityServer(ctx context.Context, kueueMgr *qcache.Manage
...
-func applyVisibilityServerOptions(config *genericapiserver.RecommendedConfig, enableInternalCertManagement bool, tlsOpts *tlsconfig.TLS) error {
+func applyVisibilityServerOptions(config *genericapiserver.RecommendedConfig, enableInternalCertManagement bool, tlsOpts *tlsconfig.TLS, visibilityServerFlags string) error {
 	o := genericoptions.NewRecommendedOptions("", codecs.LegacyCodec(
 		visibilityv1beta2.SchemeGroupVersion,
 		visibilityv1beta1.SchemeGroupVersion,
 	))
 	o.Etcd = nil
 	o.SecureServing.BindPort = 8082
+
+	if visibilityServerFlags != "" {
```
dkaluza: I wonder if this should be rather just before `o.ApplyTo(config)` ...

Currently some other options like `tlsOpts` are taking precedence than this `visibilityServerFlags`. Are we sure this is the desired behavior?
As `visibilityServerFlags` are specific to the visibility server (more specialized) for me it makes sense to let them overwrite the common ones.

cmd/kueue/main.go
```diff
@@ -164,6 +162,11 @@ func main() {
 		}
 	}
 
+	var visibilityServerFlags string
+	flag.StringVar(&visibilityServerFlags, "visibility-server-flags", "",
+		"A space-separated list of flags to pass to the embedded visibility API server "+
```
dkaluza: Maybe we should have some reference to the available set of options and their documentation?

pkg/visibility/server.go
Nilsachy: Good point, probably better right before applying the config

cmd/kueue/main.go
dkaluza: I had in mind rather actual flags list and related docs - not the options that they modify.

dkaluza:
@Nilsachy 
16th March is current planned date AFAIK see https://github.com/kubernetes-sigs/kueue/issues/8829#issue-3861260393

pkg/visibility/server.go
```diff
@@ -115,6 +118,22 @@ func applyVisibilityServerOptions(config *genericapiserver.RecommendedConfig, en
...
+		"A space-separated list of flags to pass to the embedded visibility API server. "+
+			"(e.g., '--secure-port=8443 --authentication-kubeconfig=/path/to/kubeconfig'). "+
+			"For a full list of available options, please see https://pkg.go.dev/k8s.io/apiserver/pkg/server/options#RecommendedOptions.")
+
+	if visibilityServerFlags != "" {
+		fs := pflag.NewFlagSet("visibility-server", pflag.ContinueOnError)
+		o.AddFlags(fs)
+
+		args := strings.Fields(visibilityServerFlags)
+		if err := fs.Parse(args); err != nil {
```
kshalot, line 137: Since one of the flags is a path (`--authentication-kubeconfig=/path/to/kubeconfig`), there is a non-zero chance that it will contain whitespace (which is a valid path). This would lead to something like this: https://go.dev/play/p/pwK9lhVL8Fv, so it wouldn't parse.

I think you can define a `StringArray` or `StringSlice` flag instead of `StringVar`. Then `fs.Parse` should do the splitting for you (https://pkg.go.dev/github.com/spf13/pflag#StringArray).

Nilsachy: This would mean using `StringArrayVar ` from the `pflag` library, which, IIUC, would require changing how all the flags are being parsed in `main.go`:

```
pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
pflag.Parse()
```

According to @olekzabl's comment [here](https://github.com/kubernetes-sigs/kueue/pull/8761#issuecomment-3823888714), this could cause behavioural changes which we don't yet understand.

I believe one way to deal with the whitespaces without converting all flags to pflags, would be to use `flag.Var` and create a custom string slice type like this for example:

```
type stringSlice []string

func (s *stringSlice) String() string {
    return fmt.Sprintf("%v", *s)
}

func (s *stringSlice) Set(value string) error {
    *s = append(*s, value)
    return nil
}

var visibilityServerFlags stringSlice
```

What do you think @kshalot ?

kshalot: SGTM 👍

Nilsachy: As mentioned [here](https://github.com/kubernetes-sigs/kueue/pull/9652#issuecomment-4023892058) we won't support whitespaces in paths for now, and a later fix can be applied with backward compatibility.

site/content/zh-CN/docs/tasks/manage/monitor_pending_workloads/pending_workloads_on_demand.md
```diff
@@ -74,6 +74,8 @@ You can disable it by setting the `VisibilityOnDemand` feature gate. Check the [
 
 If you disable the feature, you also need to remove the associated `APIService` from your cluster by doing `kubectl delete APIService v1beta2.visibility.kueue.x-k8s.io`.
 
+If you need to pass specific flags to the embedded visibility API server, you can use the `--visibility-server` flag when running the Kueue controller manager. For example: `--visibility-server "--secure-port=8443 --authentication-config=/path/to/kubeconfig"`. Note that the paths (e.g., to the kubeconfig) cannot contain whitespaces or be surrounded by single quotes. For a full list of available flags, please see the [kube-apiserver documentation](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/).
```
dkaluza: > or be surrounded by single quotes

Why? we are not splitting by quotes AFAIK. The quotes should work normally as in the shell in this case, or am I missing something?

pkg/visibility/server.go
```diff
@@ -60,10 +62,15 @@ var (
 		validatingwebhook.PluginName,
 		mutatingwebhook.PluginName,
 	}
-	certDir = "/visibility"
+	certDir               = "/visibility"
+	visibilityServerFlags string
```
dkaluza: Any reason to make this a global? Cannot it be just local in init?

pkg/visibility/server.go
```diff
@@ -103,18 +110,37 @@ func applyVisibilityServerOptions(config *genericapiserver.RecommendedConfig, en
...
+		o.CoreAPI.CoreAPIKubeconfigPath = kubeConfigPath
+	}
+
 	o.Admission.DisablePlugins = disabledPlugins
 	if err := o.SecureServing.MaybeDefaultWithSelfSignedCerts("localhost", nil, []net.IP{net.ParseIP("127.0.0.1")}); err != nil {
 		return fmt.Errorf("error creating self-signed certificates: %v", err)
 	}
 
+	if visibilityServerFlags != "" {
+		fs := pflag.NewFlagSet("visibility-server", pflag.ContinueOnError)
```
dkaluza, line 133: Can we add a small comment here that we are using pflag to be compatible with apiserver and to be able to utilize all of the flags available there?

It is quite not obvious without reading the PRs why we are mixing 2 flag libraries.

site/content/zh-CN/docs/tasks/manage/monitor_pending_workloads/pending_workloads_on_demand.md
Nilsachy: I believe if we use single quotes in the file path like `--authentication-config='/path/to/kubeconfig'` the path will be extracted with the single quotes included, resulting in `'/path/to/kubeconfig'`.

This will likely cause the application to fail to find the file, as it will look for a file literally named `'/path/to/kubeconfig'` (including the punctuation) rather than `/path/to/kubeconfig`.

That is because `strings.Fields` does not parse quotes.

This will be fixed when we modify the logic to use a lexer library like [github.com/google/shlex](https://github.com/google/shlex).

pkg/visibility/server.go
Nilsachy: We are referencing the `visibilityServerFlags` variable in `applyVisibilityServerOptions` as well, if it is just a local in init we won't be able to access it from there

pkg/visibility/server.go
Nilsachy: Done.

site/content/zh-CN/docs/tasks/manage/monitor_pending_workloads/pending_workloads_on_demand.md
kshalot: But that would apply to double-quotes as well, no? You can pass:
```
--visibility-server='--secure-port=8443 --authentication-config="/path/to/kubeconfig"'
```

This is also not specific to the path, since doing `foo=123` and `foo="123"` is the same thing in a shell. So anything can be quoted.

I'd change the wording to say that the values have to be unquoted and give an example of valid/invalid args.

pkg/visibility/server.go
dkaluza: Sorry, I don't know why I haven't spot it in next round of review that those are separate functions, which obviously makes perfect sense. Thanks for pointing this out!

dkaluza: Though IMO it would be a little bit cleaner to add this flag in the `main.go` and pass to create server instead of declaring as global. In this way flag declarations are not spread across multiple files.

Nilsachy: You're right that's probably cleaner. Done.

site/content/zh-CN/docs/tasks/manage/monitor_pending_workloads/pending_workloads_on_demand.md
Nilsachy: But still there is something I don't get. Let's say the (whole) flag is in that format you mentioned:

> ```--visibility-server='--secure-port=8443 --authentication-config="/path/to/kubeconfig"'```

This will extract the authentication-config path as `"/path/to/kubeconfig"` rather than `/path/to/kubeconfig`, isn't it? And therefore it will try to look for a file with a path containing the punctuation which is wrong no?

Nilsachy: No sorry you're right, I think quotes don't matter here. Only whitespace is not allowed.

kshalot: Hm, quotes being disallowed makes sense to me, since we are essentially splitting on whitespace. So I'd say you are correct in your assessment that the quotes will become part of the "value".

My point was that the documentation explicitly mentions **single** quotes:
> cannot contain whitespaces or be surrounded by single quotes.

while the same logic applies to double quotes.

This was just an example of an "invalid" argument:
> `--visibility-server='--secure-port=8443 --authentication-config="/path/to/kubeconfig"'`

kshalot: My second point was that for some reason this limitation is said to only apply to paths:

> Note that the paths (e.g., to the kubeconfig) cannot contain...

While it would apply to all arguments.

Nilsachy: Correct, we should specify no whitespace in all arguments. When it comes to quotes, I am still unsure if they are allowed or not. I can test it manually.

kshalot: > while `--visibility-server "--authentication-config='/path/without_whitespace/to/kubeconfig'"` will work as expected.

I see you specified them as allowed. But I'm not sure... My intuition would be that they will become part of the path/argument. Like you said:

> I believe if we use single quotes in the file path like --authentication-config='/path/to/kubeconfig' the path will be extracted with the single quotes included, resulting in '/path/to/kubeconfig'.

This makes sense to me. But you can check.

Nilsachy: Tested manually, quotes are not allowed. Updated the doc. Please let me know if it is clear now @kshalot @dkaluza

Nilsachy: Done.

test/e2e/customconfigs/visibility_kubeconfig_test.go
```diff
@@ -151,11 +164,20 @@ var _ = ginkgo.Describe("Visibility Server KubeConfig flag with RBAC", func() {
...
+				container.Args = append(c.Args, "--kubeconfig=/etc/kubeconfig/config", "--visibility-server=--secure-port=9444")
 			}
 		}
 
 		gomega.Expect(k8sClient.Update(ctx, patchedDeployment)).To(gomega.Succeed())
+
+		patchedService := originalService.DeepCopy()
+		for i, p := range patchedService.Spec.Ports {
+			if p.Name == "https" {
+				patchedService.Spec.Ports[i].TargetPort = intstr.FromInt32(9444)
```
dkaluza: nit: it would be nice to use some const for port to show connection with container args setting

Nilsachy: Done.

site/content/zh-CN/docs/tasks/manage/monitor_pending_workloads/pending_workloads_on_demand.md
```diff
@@ -74,6 +74,12 @@ You can disable it by setting the `VisibilityOnDemand` feature gate. Check the [
 
 If you disable the feature, you also need to remove the associated `APIService` from your cluster by doing `kubectl delete APIService v1beta2.visibility.kueue.x-k8s.io`.
 
+If you need to pass specific flags to the embedded visibility API server, you can use the `--visibility-server` flag when running the Kueue controller manager. For example: `--visibility-server "--secure-port=8443 --authentication-config=/path/to/kubeconfig"`. 
+
+Note that the arguments to the embedded visibility API server (e.g. the kubeconfig) do not support normal shell-like parsing and cannot contain quotes or whitespaces. Since internal quotes are not stripped, a value like `--authentication-config='/path/to/kubeconfig'` will result in an actual path of `'/path/to/kubeconfig'` (including the single quotes), which is not a typical file path format. Similarly, including whitespaces like `--authentication-config=/path/ with_whitespace /to/kubeconfig` will cause the argument to be incorrectly split into multiple parts.
+
+For a full list of available flags, please see the [kube-apiserver documentation](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/).
```
mimowo, line 81: I would suggest moving this to a dedicated PR one we know which version actually supports the feature. Then we can update documetation with the precise version information.

cmd/kueue/main.go
```diff
@@ -129,6 +129,12 @@ func main() {
 	var featureGates string
 	flag.StringVar(&featureGates, "feature-gates", "", "A set of key=value pairs that describe feature gates for alpha/experimental features.")
 
+	var visibilityServerFlags string
+	flag.StringVar(&visibilityServerFlags, "visibility-server", "",
+		"A space-separated list of flags to pass to the embedded visibility API server. "+
+			"(e.g., '--secure-port=8443 --authentication-config=/path/to/kubeconfig'). "+
+			"For a description of the available flags, please see https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/")
+
```
tenzen-y, line 137: I think that command line option flag is less UX than configurations because users need to consider the command position when overriding the kueue manifests and there are no rich comments and schema.

So, shouldn't we move this to Kueue Configuration?
https://github.com/kubernetes-sigs/kueue/blob/0a58afe5f795f5412ec88f6d23c510e80b005941/apis/config/v1beta2/configuration_types.go

mimowo: yeah, ideally, this is the long term goal, but I think the team wanted to have something which could be cherrypicked. @kshalot @dkaluza @Nilsachy let me know if I'm correct. In that case I feel like a flag is more lightweight than Configuration API.

mimowo: If we don't need to cherrypick and can quickly draft a KEP for 0.17 then we could try that too, wdyt @tenzen-y ?

tenzen-y: If your team want to ship this as soon as possible by backport, I'm ok with adding this flag as temporary approach. In that case, we should remove this flag after configurations are supported.

Please let me know if your team wants to backport this to the released branches.

mimowo: > we should remove this flag after configurations are supported.

This might be very hard if we start to use flags, or at least it will require some graceful period.

Note that we support both `--feature-gates` as deployment param and `featureGates` as a field. Do you think this is much maintenance cost long term?

tenzen-y: > This might be very hard if we start to use flags, or at least it will require some graceful period.

The graceful period is also acceptable when we decide to migrate it to configurations.

> Note that we support both --feature-gates as deployment param and featureGates as a field. Do you think this is much maintenance cost long term?

I don't remember the reason why we didn't remove the feature gate flag, but we used to drop command-line flags based on growing configurations previously.

Once we accept the command-line flag addition, we will add flags more and more in the future because that is easy way, but not good UX.

By increasing command-line flags, the cluster admins' bad UX is accelerating.
So, regardless of a single command-line flag, I'd like to block the addition.

mimowo: ack, let me hear from the team

tenzen-y: thank you!

Nilsachy: So far it seems that for our use case, we only need to be able to change the port number of the visibility server (`--secure-port=xxxx`). We do need this to be backported asap. If the command-line flag addition is to be anyway removed in the long term, maybe we can add a custom flag only for the port number of the visibility server, like `--visibility-secure-port` which can be more easily removed in the future when moving to the Configuration API? Only downside is that we aren't yet fully sure if we won't be needing more configuration from the apiserver, but it seems unlikely for now. What do you think? @tenzen-y @mimowo

mimowo: +1 for me for `--visibility-secure-port` rather than generic --visibility-flags to minimize the command line surface, then it is easier to deprecate and drop. 
Let me ask: do we need it backported to 0.16?
Also, could we already start the KEP effort with something like "Configuration API for the visibility server"? It could be just the securePort for now, but maybe we could grow it in the future. If this this just one field we should be quick to approve the KEP .

Nilsachy: Ideally yes we would need this backported to 0.16

Nilsachy: This is a new Draft PR: #9850 for the Configuration API changes

PR #9102, opened 2026-02-10
KubeRay: Add RayService controller as top level job

hiboyang:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

This is to support [KubeRay integration: Support RayService as a top-level Job](https://github.com/kubernetes-sigs/kueue/issues/9059).

Also add E2E test for RayService.

#### Which issue(s) this PR fixes:

Fixes https://github.com/kubernetes-sigs/kueue/issues/9059

#### Does this PR introduce a user-facing change?
```release-note
KubeRay integration: native support for managing RayService via Kueue.
```

PR #9660, opened 2026-03-03
Update RayService E2E test to check http endpoint inside the RayService

hiboyang:
#### What type of PR is this?

/kind cleanup

This is a follow-up PR for https://github.com/kubernetes-sigs/kueue/pull/9102, to improve the test.

After this PR change, we will add RayService autoscaling test, thus need a way to hit the endpoint inside RayService to trigger autoscaling.

```release-note
NONE
```

### DIFF
--- a/test/util/e2e.go
+++ b/test/util/e2e.go
@@ -22,7 +22,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"net"
 	"net/http"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


sohankunkerkar:
> > Fixes #9059
> 
> This looks like the umbrella task. Do you maybe make the changes to fix the autoscaling flaky test: #9523?
> 
> cc @sohankunkerkar do you think this has a potential as a fix there?

This PR doesn't address #9523. Regarding this PR, the KPortForward utility has a TOCTOU port race (finds a free port, closes listener, then portforward library re-listens and then another process could grab it in between). If we do need this utility later, it should use `0:remotePort` with `GetPorts()` instead.

hiboyang:
> > > Fixes #9059
> > 
> > 
> > This looks like the umbrella task. Do you maybe make the changes to fix the autoscaling flaky test: #9523?
> > cc @sohankunkerkar do you think this has a potential as a fix there?
> 
> This PR doesn't address #9523. Regarding this PR, the KPortForward utility has a TOCTOU port race (finds a free port, closes listener, then portforward library re-listens and then another process could grab it in between). If we do need this utility later, it should use `0:remotePort` with `GetPorts()` instead.

Good suggestion, let me trying something like `0:remotePort`

PR #9661, opened 2026-03-03
Fix inadmissible workload stuck after priority update

### LINKED ISSUE #9621: Kueue when Creating a Job With Queueing Should allow updating the workload's priority through the job
**Which test is flaking?**:
Kueue when Creating a Job With Queueing Should allow updating the workload's priority through the job
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/9617/pull-kueue-test-e2e-main-1-33/2028383367677874176
**Failure message or logs**:
```
End To End Suite: kindest/node:v1.33.7: [It] Kueue when Creating a Job With Queueing Should allow updating the workload's priority through the job [area:singlecluster, feature:job] expand_less	14s
{Timed out after 10.001s.
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/singlecluster/job_test.go:534 with:
Expected
    <bool>: true
to be false failed [FAILED] Timed out after 10.001s.
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/singlecluster/job_test.go:534 with:
Expected
    <bool>: true
to be false
In [It] at: /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/singlecluster/job_test.go:535 @ 03/02/26 08:40:19.903
}
```

sohankunkerkar:
#### What type of PR is this?
/kind bug

Fixes #9621 

When a workload's spec changes while the scheduler is evaluating it, `RequeueWorkload.info.Update` refreshes the inadmissible entry to the latest version, causing `PushOrUpdate`'s spec comparison to see no change and leaving the workload stuck in inadmissible. Track the evaluated generation at Pop() time so `PushOrUpdate` detects spec changes under the CQ lock regardless of caller freshness.

I ran the failed test with this change in a loop (200 times) and it works fine.

```release-note
Scheduling: Fix a race where updated workload priority could remain stuck in the inadmissible queue and delay rescheduling.
```

### DIFF
--- a/pkg/cache/queue/cluster_queue.go
+++ b/pkg/cache/queue/cluster_queue.go
@@ -258,10 +258,12 @@ func (c *ClusterQueue) PushOrUpdate(wInfo *workload.Info) {
 		return
 	}
 	if oldInfo := c.inadmissibleWorkloads.get(key); oldInfo != nil {
-		// update in place if the workload was inadmissible and didn't change
-		// to potentially become admissible, unless the Eviction status changed
-		// which can affect the workloads order in the queue.
-		if equality.Semantic.DeepEqual(oldInfo.Obj.Spec, wInfo.Obj.Spec) &&
+		specChangedSinceEval := oldInfo.LastEvaluatedGeneration != 0 &&
+			wInfo.Obj.Generation != oldInfo.LastEvaluatedGeneration
+
+		// Update in place if the workload didn't change to potentially become admissible.
+		if !specChangedSinceEval &&
+			equality.Semantic.DeepEqual(oldInfo.Obj.Spec, wInfo.Obj.Spec) &&
 			equality.Semantic.DeepEqual(oldInfo.Obj.Status.ReclaimablePods, wInfo.Obj.Status.ReclaimablePods) &&
 			equality.Semantic.DeepEqual(apimeta.FindStatusCondition(oldInfo.Obj.Status.Conditions, kueue.WorkloadEvicted),
 				apimeta.FindStatusCondition(wInfo.Obj.Status.Conditions, kueue.WorkloadEvicted)) &&
@@ -468,6 +470,7 @@ func (c *ClusterQueue) Pop() *workload.Info {
 		return nil
 	}
 	c.inflight = c.heap.Pop()
+	c.inflight.LastEvaluatedGeneration = c.inflight.Obj.Generation
 	return c.inflight
 }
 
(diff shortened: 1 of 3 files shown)


pkg/cache/queue/cluster_queue.go
```diff
@@ -109,6 +109,10 @@ type ClusterQueue struct {
 	// inflight indicates the workload that was last popped by scheduler.
 	inflight *workload.Info
 
+	// lastEvaluatedGeneration stores each workload's Generation at Pop time,
+	// so PushOrUpdate can detect spec changes masked by RequeueWorkload's info.Update.
+	lastEvaluatedGeneration map[workload.Reference]int64
```
mwielgus: Why not put it into https://github.com/kubernetes-sigs/kueue/blob/main/pkg/workload/workload.go#L187 ?

pkg/cache/queue/cluster_queue_test.go
```diff
@@ -242,6 +242,50 @@ func TestPushOrUpdateSkipsInflightWorkload(t *testing.T) {
 	}
 }
 
+func TestPushOrUpdateMovesInadmissibleWhenGenerationChanged(t *testing.T) {
```
mwielgus: Is it possible to have tests with backoff as well?

pkg/cache/queue/cluster_queue.go
sohankunkerkar: Yeah, I think it would be cleaner. will fix that.

pkg/cache/queue/cluster_queue_test.go
```diff
@@ -242,6 +242,84 @@ func TestPushOrUpdateSkipsInflightWorkload(t *testing.T) {
...
+
+	if active, _ := cq.Dump(); len(active) != 1 {
+		t.Errorf("expected workload on heap after generation change, got %d", len(active))
+	}
+	if inadmissible, _ := cq.DumpInadmissible(); len(inadmissible) != 0 {
+		t.Errorf("expected no inadmissible workloads after generation change, got %d", len(inadmissible))
+	}
+}
+
+func TestPushOrUpdateWithBackoffKeepsInadmissibleWhenGenerationChanged(t *testing.T) {
```
mimowo: There is a signifficant overlap between the unit tests in setup, and asserts. Would it be feasible to combine them both into table-driven tests?

sohankunkerkar: Fixed!

mimowo:
Thank you 👍 
/lgtm
/approve
/cherrypick release-0.16
/cherrypick release-0.15

PR #9670, opened 2026-03-04
Sync local MK job when evicted on Manager due to eviction on admitted Worker.

### LINKED ISSUE #9338: MultiKueue "leaks" admission on the manager when a workload is evicted on a worker
# Background

The scenario considered here is when MultiKueue admits a workload on a worker but then it gets preempted there.

Before #8477, the workload would just get stuck, as described in #8302.
After that fix, with the `MultiKueueRedoAdmissionOnEvictionInWorker` gate enabled, the behavior is better:

- the manager recognizes that a worker has evicted the workload
- the manager triggers re-execution of the admission checks, including MultiKueue AC
- the workload is dispatched to some (likely other) workers, giving it a chance to run again

That's all fine, and covered by tests.

# The issue

The issue is that the workload remains `Admitted` on the manager during that whole procedure.
This becomes particularly problematic if the workload is not going to become re-scheduled for a long time.
In such case, from the perspective of the manager cluster, `Admitted` suggests that the workload is running on some worker, while it's not the case.

I'd rather ensure that `Admitted` on a manager _implies_ `Admitted` on a worker. 
(To be strict: except the short periods when the workload lost admission on a worker but the manager hasn't yet "realized" that fact).

# How to reproduce

1. In a clone of Kueue repo, run:
   ```
   cd kueue/examples/multikueue/dev
   ./setup-kind-multikueue-tas.sh --version v0.16.0
   ```
   (setting this version is important to have #8477 on board).

2. In any directory of your choice, run [this script](https://github.com/olekzabl/kueue/blob/repro-4/hack/bug-repro/test.sh).
   (Warning: it depends on the other files in the same folder).
   The script contains comments about the specific test scenario.

The script output is [here](https://github.com/olekzabl/kueue/blob/repro-4/hack/bug-repro/output/out.log). The Kueue logs from all clusters are placed in the same folder. Notable observations:

* `job1-7` has preempted `job1-2` on `worker1`.
* `manager` noticed this and dispatched `job1-2` again to `worker2` (where it could not get quota).
* `job1-2` status is correctly reflected on `worker1` (no longer admitted, no quota reserved) and on `worker2` (no quota reserved).
* However, on `manager`, `job1-2` is still displayed as `admitted`.
* The mechanism from #8477 has been invoked; this can be seen e.g. [here](  https://github.com/olekzabl/kueue/blob/77347bc316f7494d65c4bb19ca39a71791027575/hack/bug-repro/output/manager.log#L324) in the manager logs.
  Also, the status of the MultiKueue AdmissionCheck was switched to false, as logs show [here](  https://github.com/olekzabl/kueue/blob/77347bc316f7494d65c4bb19ca39a71791027575/hack/bug-repro/output/manager.log#L326).
  Given that, it feels suspicious that the workload remained admitted.

# Idea for fixing

When the manager detects a worker-side eviction:

* Retain the `QuotaReserved` condition for the workload on the manager, to keep it actively considered by MultiKueue controller.
* However, remove the `Admitted` condition. (Bring it back when the workload gets admitted on some worker).

In my understanding, this will make the overall workload status analogous as if it's been just dispatched to the workers for the first time. (With the only difference that one of the workers has "somehow managed" to evict it). This *seems resonable at the first glance*.

However, obviously we'd **need to double-check** if this doesn't bring any trouble in practice.

Not sure if this warrants a feature gate. (If we hear no opinion on this, I'll assume "yes" for safety. Though maybe it'd make most sense to re-use the `MultiKueueRedoAdmissionOnEvictionInWorker` gate introduced in #8477).

Singularity23x0:
#### What type of PR is this?
/kind bug
/area multikueue

Addressing the issue described in #9338.

The bug occurs because jobs are not synced (updating their State field) when evicted on a remote, failing to acknowledge the removal of pods, leading to remain "Active" and causing the job reconciler not to roll the workload back.

To address that, this PR ensures that the MK Reconciler always syncs the local job to its remote instance if evicted on the Manager and on the Worker. Now we will perform this operation when the eviction was originally caused by the worker. This ensures that upon eviction on the manager, if the job was already dispatched, it will be rolled back as its remote instance is.

Fixes #9338

NONE

```release-note
MultiKueue: Fixes the bug where a job, after being dispatched to a worker, would not sync correctly after being evicted there. This would also cause its workload to be incorrectly labeled as admitted.

Now the workload and the manager job instance will correctly reflect the evicted state and MultiKueue will perform a fallback, then dispatch remote workloads to all eligible workers again after being evicted from the Worker it was successfully admitted to before. An example of such a case is if the remote instance got preempted on the worker.
```

### DIFF
--- a/test/e2e/multikueue/e2e_test.go
+++ b/test/e2e/multikueue/e2e_test.go
@@ -2487,7 +2487,6 @@ app = HelloWorld.bind()`,
 					g.Expect(unaffectedWorkerClient.Get(ctx, unaffectedWlKey, unaffectedWl)).To(gomega.Succeed())
 					g.Expect(unaffectedWl.Status.Conditions).To(utiltesting.HaveConditionStatusTrue(kueue.WorkloadAdmitted))
 					g.Expect(workload.IsEvicted(unaffectedWl)).To(gomega.BeFalse())
-
 				}, util.ConsistentDuration, util.Interval).Should(gomega.Succeed())
 			})
 
@@ -2507,7 +2506,6 @@ app = HelloWorld.bind()`,
 					// Evicted workload is requeued and pending on worker 2
 					g.Expect(k8sWorker2Client.Get(ctx, evictedWlKey, evictedWl)).To(gomega.Succeed())
 					g.Expect(evictedWl.Status.Conditions).To(utiltesting.HaveConditionStatusFalse(kueue.WorkloadQuotaReserved))
-
 				}, util.ConsistentDuration, util.Interval).Should(gomega.Succeed())
 			})
 		})


test/e2e/multikueue/e2e_test.go
```diff
@@ -853,6 +856,115 @@ var _ = ginkgo.Describe("MultiKueue", func() {
...
 		})
 
+		ginkgo.It("Should preempt a running low-priority workload on a worker when a high-priority workload is admitted to it by manager", func() {
+			ginkgo.By("Ensuring worker2 does not have enough quota to handle created workloads", func() {
+				gomega.Eventually(func(g gomega.Gomega) {
+					g.Expect(k8sWorker2Client.Get(ctx, client.ObjectKeyFromObject(worker2Cq), worker2Cq)).To(gomega.Succeed())
+					worker2Cq.Spec.ResourceGroups[0].Flavors[0].Resources[0].NominalQuota = resource.MustParse("100m")
+					g.Expect(k8sWorker2Client.Update(ctx, worker2Cq)).To(gomega.Succeed())
+				}, util.Timeout, util.Interval).Should(gomega.Succeed())
+			})
```
olekzabl: AFAICS worker1 CQ has 2 CPUs, worker2 CQ has 1.2 CPU.
So - you shouldn't need this step if you raise lowJob to 1.5 CPU?

test/e2e/multikueue/e2e_test.go
```diff
@@ -853,6 +856,115 @@ var _ = ginkgo.Describe("MultiKueue", func() {
...
+
+			ginkgo.By("Checking that the low-priority workload was evicted on the manager cluster", func() {
+				gomega.Eventually(func(g gomega.Gomega) {
+					g.Expect(k8sManagerClient.Get(ctx, lowWlKey, managerLowWl)).To(gomega.Succeed())
+					g.Expect(workload.IsEvicted(managerLowWl)).To(gomega.BeTrue())
+					g.Expect(managerLowWl.Status.Conditions).To(utiltesting.HaveConditionStatusFalse(kueue.WorkloadAdmitted))
+				}, util.Timeout, util.Interval).Should(gomega.Succeed())
+			})
+
+			ginkgo.By("Checking that the low-priority job is not active", func() {
```
olekzabl, line 946: Can you also, just for completeness, check that the Workload (on the manager) is not Admitted?
Because this is the thing we wanted to fix.

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -378,7 +378,7 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
 		ctx = ctrl.LoggerInto(ctx, log)
 
 		// workload evicted on manager cluster
```
olekzabl, line 381: I think this comment needs updating.

test/e2e/multikueue/e2e_test.go
```diff
@@ -853,6 +856,115 @@ var _ = ginkgo.Describe("MultiKueue", func() {
 			})
 		})
 
+		ginkgo.It("Should preempt a running low-priority workload on a worker when a high-priority workload is admitted to it by manager", func() {
```
olekzabl, line 870: What is the value of the `MultiKueueRedoAdmissionOnEvictionInWorker` FG in this test?
I'm guessing it's `true` (which is [the default](https://github.com/kubernetes-sigs/kueue/blob/5da81fd0b798e446a7ef6fc1fb2e0c068be6c551/site/data/featuregates/versioned_feature_list.yaml#L168))?

It'd be nice to verify how this PR behaves also when this FG is false.
I'm not sure if necessarily with a test like this (though this feels best) - but _at least_ verifying in _any_ way.

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -378,7 +378,7 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
...
-		if remoteEvictCond.Reason == workload.ReasonWithCause(kueue.WorkloadDeactivated, kueue.WorkloadEvictedOnManagerCluster) {
+		if workload.IsEvicted(group.local) {
```
mbobrovskyi, line 382: Should we add an eviction reason like WorkloadEvictedOnWorkerCluster?

@mimowo @olekzabl WDYT?

test/e2e/multikueue/e2e_test.go
Singularity23x0: I need both to fit on the Manager CQ, which also has 2 CPU. Either way I would have to update a cq somewhere. I can change to that approach if you feel strongly abt it as both would work just as well.

test/e2e/multikueue/e2e_test.go
Singularity23x0: Already done in the block above, where we check for eviction:
					`g.Expect(managerLowWl.Status.Conditions).To(utiltesting.HaveConditionStatusFalse(kueue.WorkloadAdmitted))`

pkg/controller/admissionchecks/multikueue/workload.go
Singularity23x0: Not really, it is still true - we essentially will work in 2 cycles, 1 will set the check to retry. Workload controller will process this and evict the workload. Only then will we proceed to 2 where we are evicted in both places and will sync the job. So essentially - this case stays the same, we just make sure this logic will now ALSO be performed after we evict the workload on the manager by evicting on the worker.

pkg/controller/admissionchecks/multikueue/workload.go
Singularity23x0: WorkloadEvictedOnWorkerCluster feels inaccurate, as the precise reason for eviction is "AdmissionCheck no longer ready". We could ensure that that eviction reason has the cause for "MK AC" when applicable and populates it further with details like "Lost connection to worker" or "Evicted on worker" etc.. 

Either way this feels like a wider change for which I propose making a separate issue.

pkg/controller/admissionchecks/multikueue/workload.go
olekzabl: Ah, ok, the current phrasing captures the new condition equally well.

👍

test/e2e/multikueue/e2e_test.go
olekzabl: I see. we have:

* `managerCq`: 2 CPU / 4G RAM
* `worker1Cq`: 2 CPU / 1G RAM
* `worker2Cq`: 1.2 CPU / 4G RAM

Then, how about:

* `lowJob`: 1 CPU / 2G RAM
* `highJob`: 1 CPU / 2G RAM

Then: both together fit on manager; neither fits on w1; either fits on w2 but not together.
(Which is I guess what you need; just with the game moved from w1 to w2).

pkg/controller/admissionchecks/multikueue/workload.go
olekzabl: I wouldn't add such eviction reason (due to relative complexity of adding it, as follows from explanation by @Singularity23x0 ) unless we need it.
And I don't see how we'd need it currently.

@mbobrovskyi  As I understand, your concern is that there might exist some "Third Eviction Case" ("some remote is evicted & local is evicted but the remote eviction reason is neither EvictedOnManager nor EvictedOnWorker"), and in that scenario executing the block below could be incorrect?
* Whether a "Third Eviction Case" may exist - I don't know, maybe.
* Whether for a "Third Eviction Case" executing the block below could be problematic - I don't think so because:
  1. Doing a single sync doesn't harm.
      (What could harm is a loop of unnecessary syncs - but see below).
  2. Returning early does not seem to interfere with the rest of this method, as I look at it.
      (I mean, the remainder of this method anyway operates under different assumptions).
  3. Whenever "local is evicted" happens, I'd expect remotes to be removed soon.
      (Specifically - by step 2 of this `reconcileGroup` method).
  4. By (iii), any case of "local is evicted & some remote is evicted" should be short-lived.
      So we shouldn't get a loop of unnecessary syncs.

But I'm not 100% sure. If you have any concrete concerns, please share them.

test/e2e/multikueue/e2e_test.go
Singularity23x0: As per disussion in a comment below - should not be necessary.

test/e2e/multikueue/e2e_test.go
```diff
@@ -853,6 +853,116 @@ var _ = ginkgo.Describe("MultiKueue", func() {
...
```
mimowo, line 870: The test is pretty long - can you point which assert(s) fail without the change?

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -378,7 +378,7 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
 		ctx = ctrl.LoggerInto(ctx, log)
 
 		// workload evicted on manager cluster
-		if remoteEvictCond.Reason == workload.ReasonWithCause(kueue.WorkloadDeactivated, kueue.WorkloadEvictedOnManagerCluster) {
```
mimowo, line 382: think we added this more involving check to make sure there are no preemption/eviction cycles. I remmeber @mbobrovskyi was reporting some possible, but I cannot recall details now. @mbobrovskyi can you double check if simplifying this check is safe?

test/e2e/multikueue/e2e_test.go
```diff
@@ -853,6 +853,116 @@ var _ = ginkgo.Describe("MultiKueue", func() {
...
+				}, util.Timeout, util.Interval).Should(gomega.Succeed())
+			})
+
+			ginkgo.By("Checking that the low-priority job is not active", func() {
+				gomega.Eventually(func(g gomega.Gomega) {
+					admittedJob := &batchv1.Job{}
+					g.Expect(k8sManagerClient.Get(ctx, client.ObjectKeyFromObject(lowJob), admittedJob)).To(gomega.Succeed())
+					g.Expect(admittedJob.Status.Active).To(gomega.Equal(int32(0)))
+				}, util.Timeout, util.Interval).Should(gomega.Succeed())
+			})
```
Singularity23x0: These two checks are what would fail without the change.
Specifically:
* low-job would remain Active,
* its workload would keep the Admitted Condition set to True.

test/e2e/multikueue/e2e_test.go
Singularity23x0: Added as a separate comment.

test/e2e/multikueue/e2e_test.go
mimowo: Hm, why do we expect the low priority Job to get deactivated? Could you check / assert what is the reason for the Deactivated condition?

Singularity23x0: We expect the workload to be evicted on the manager after being evicted on the nominated worker.

To that end we roll back the ready ac to retry then pending. This should evict the workload on the manager, but that logic is handled by the generic ob reconciler and will happen only once the job reports IsActive as false. The introduced change ensure the local job is synced to the job on the preempted worker - the jobs state then correctly reflects the fact that no pods are running and the job reports IsActive as false. Previously it would claim that there are pods running since its last update was done before the preemption, hence we would never fall back via an eviction and the manager workload remained "Admitted".

Singularity23x0: In short: we do not expect the job to be deactivated, we know it happens, bu the local job does not reflect that as it does not sync when necessary.

mimowo:
/release-note-edit
```release-note
MultiKueue: Fixes the bug where a job, after being dispatched to a worker, would not sync correctly after being evicted there. This would also cause its workload to be incorrectly labeled as admitted.

Now the workload and the manager job instance will correctly reflect the evicted state and MultiKueue will perform a fallback, then dispatch remote workloads to all eligible workers again after being evicted from the Worker it was successfully admitted to before. An example of such a case is if the remote instance got preempted on the worker.
```

PR #9674, opened 2026-03-04
Small clarifications in scheduling code

olekzabl:
#### What type of PR is this?

/kind cleanup

It's a random set of small clarifications following my recent reading of the code (partially with @Singularity23x0 ).

This includes:

- A few changes to the comments which were previously misleading.
- In [`RequeueIfNotPresent`](https://github.com/kubernetes-sigs/kueue/blob/48a2c7d3dca3f347268566c76608f712ebe41a7d/pkg/cache/queue/cluster_queue.go#L564-L592), a refactor to indicate more clearly what is "the boolean" mentioned in the comment. (Which otherwise feels highly confusing, as "the boolean" is neither the argument to the function nor its return value).
- In `queueInadmissibleWorkloads`, a fix to the log, which in its current form can be misleading in some cases.

(none)

The log fix is, strictly speaking, user-facing. So I drafted a release note. Though it's such a minor thing that maybe `NONE` would fit better? I'm not sure about this.

```release-note
NONE
```

### DIFF
--- a/pkg/cache/queue/cluster_queue.go
+++ b/pkg/cache/queue/cluster_queue.go
@@ -558,12 +558,10 @@ func (c *ClusterQueue) Active() bool {
 }
 
 // RequeueIfNotPresent inserts a workload that was not
-// admitted back into the ClusterQueue. If the boolean is true,
-// the workloads should be put back in the queue immediately,
-// because we couldn't determine if the workload was admissible
-// in the last cycle. If the boolean is false, the implementation might
-// choose to keep it in temporary placeholder stage where it doesn't
-// compete with other workloads, until cluster events free up quota.
+// admitted back into the ClusterQueue.
+// This may be done either "immediately" or "non-immediately";
+// in the latter case the implementation may choose to keep the workload in "inadmissible workloads"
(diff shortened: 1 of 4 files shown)


sohankunkerkar (commented):
Changes LGTM
nit: IMO, this change doesn't required a release note. No user would notice this unless they're running with very verbose logging.

pkg/cache/queue/cluster_queue.go
```diff
@@ -582,13 +576,15 @@ func (c *ClusterQueue) RequeueIfNotPresent(ctx context.Context, wInfo *workload.
...
-	}
-	return c.requeueIfNotPresent(log, wInfo,
-		reason == RequeueReasonFailedAfterNomination ||
+		immediate = reason != RequeueReasonNamespaceMismatch
+	} else {
+		immediate = reason == RequeueReasonFailedAfterNomination ||
 			reason == RequeueReasonPendingPreemption ||
-			reason == RequeueReasonPreemptionFailed)
+			reason == RequeueReasonPreemptionFailed
+	}
```
mimowo, line 586: +1, much more readable

PR #9675, opened 2026-03-04
Update KEP-6757 history to include changes after foreground propagated deletion fix

kshalot:
#### What type of PR is this?
/kind kep

The KEPs history should be updated to match content updates (https://github.com/kubernetes-sigs/kueue/pull/9651).

Fixes #

```release-note
NONE
```

### DIFF
--- a/keps/6757-failure-recovery/README.md
+++ b/keps/6757-failure-recovery/README.md
@@ -275,7 +275,7 @@ Existing integration tests should prove that this feature does not impact Kueue 
 2026-03-04: The controller is [changed](https://github.com/kubernetes-sigs/kueue/pull/9651) to delete pods in addition to marking them as failed,
 to handle [foreground propagated deletion](https://github.com/kubernetes-sigs/kueue/issues/9649).
 Additionally, the annotation used by the feature was renamed from `kueue.x-k8s.io/safe-to-forcefully-terminate` to `kueue.x-k8s.io/safe-to-forcefully-delete`
-in versions `>=0.15.6`, `>=0.16.3` and `>=0.17.0`. Previous releases retain the old annotation name.
+in versions `>=0.15.6`, `>=0.16.3` and `>=0.17.0`. Previous releases retain the old behaviour and annotation name.
 
 ## Drawbacks
 


keps/6757-failure-recovery/README.md
```diff
@@ -272,6 +269,9 @@ Existing integration tests should prove that this feature does not impact Kueue
 
 2025-10-17: First draft of the KEP.
 
+2026-03-04: The controller is [changed](https://github.com/kubernetes-sigs/kueue/pull/9651) to delete pods in addition to marking them as failed,
+to handle [foreground propagated deletion](https://github.com/kubernetes-sigs/kueue/issues/9649).
+
```
tenzen-y, line 279: Additionally, could you mention about annotation renaminig? (`kueue.x-k8s.io/safe-to-forcefully-terminate` -> `kueue.x-k8s.io/safe-to-forcefully-delete`)

tenzen-y: Additionally, I'd like to mention which Kueue version uses the old annotation, and which Kueue version starts introducing the new annotation.

tenzen-y: My proposal, but feel free to refine it following my proposal.

```suggestion
2026-03-04: The controller is [changed](https://github.com/kubernetes-sigs/kueue/pull/9651) to delete pods in addition to marking them as failed,
to handle [foreground propagated deletion](https://github.com/kubernetes-sigs/kueue/issues/9649).
Alongside the implementation change, we renamed the annotation name `kueue.x-k8s.io/safe-to-forcefully-terminate` (until Kueue v0.16) with `kueue.x-k8s.io/safe-to-forcefully-delete` (since Kueue v0.17).
```

kshalot: Ah yes, I missed the sentence about the renaming in the original commit, thanks for pointing it out.

When it comes to versions, it's not that simple - it would be sth like `0.15.0 - 0.15.5` and `0.16.0 - 0.16.2` are using the old one, `>0.15.5` and `>0.16.2` are using the new one. I'll think how to phrase this.

kshalot: I went for something simple - PTAL

PR #9684, opened 2026-03-05
kueueviz: add token-based authentication for backend and frontend

### LINKED ISSUE #5993: KueueViz without authorized at all

Kubeviz current without add authorized by api call to get kueue resource.

suppport need add it?

**Why is this needed**:

If we want open the dashboard for more user, we may need to consider security.

**Completion requirements**:

- [ ] Design doc
- [x] API change
- [x] Docs update

samzong:
#### What type of PR is this?

/kind feature

Add token-based authentication to KueueViz

Fixes #5993

```release-note
KueueViz: Add token-based authentication as opt-in.
```

### DIFF
--- a/charts/kueue/templates/kueueviz/clusterrole.yaml
+++ b/charts/kueue/templates/kueueviz/clusterrole.yaml
@@ -32,4 +32,9 @@ rules:
   - apiGroups: ["kueue.x-k8s.io"]
     resources: ["workloadpriorityclass"]
     verbs: ["get", "list", "watch"]
+{{- if ne .Values.kueueViz.backend.auth.mode "Disabled" }}
+  - apiGroups: ["authentication.k8s.io"]
+    resources: ["tokenreviews"]
+    verbs: ["create"]
+{{- end }}
 {{- end }}
--- a/cmd/kueueviz/backend/middleware/auth.go
+++ b/cmd/kueueviz/backend/middleware/auth.go
@@ -1,0 +1,239 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package middleware
+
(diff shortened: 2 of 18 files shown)


charts/kueue/templates/kueueviz/clusterrole.yaml
```diff
@@ -32,4 +32,9 @@ rules:
   - apiGroups: ["kueue.x-k8s.io"]
     resources: ["workloadpriorityclass"]
     verbs: ["get", "list", "watch"]
+  {{- if .Values.kueueViz.backend.auth.enabled }}
```
mwielgus: The values.yaml file uses a mode field (defaulting to "Disabled") rather than an enabled boolean.

cmd/kueueviz/backend/middleware/auth.go
```diff
@@ -0,0 +1,207 @@
...
+		return cacheEntry{}, false
+	}
+	if now.Before(entry.expiresAt) {
+		return entry, true
+	}
+
+	c.mu.Lock()
+	entry, ok = c.entries[key]
+	if ok && !now.Before(entry.expiresAt) {
+		delete(c.entries, key)
```
mwielgus, line 229: Expired entries are only deleted when a request for the same token key is made. In a high-churn environment or under a hack-attempt scenario, the entries map will grow indefinitely as expired tokens that are never reused stay in memory

cmd/kueueviz/frontend/src/useWebSocket.js
```diff
@@ -16,75 +16,54 @@ limitations under the License.
 
 import { useEffect, useState } from 'react';
 import { buildWebSocketUrl } from './utils/urlHelper';
+import { useAuth } from './AuthContext';
+
+const WS_BASE_PROTOCOL = 'kueueviz.v1';
+const WS_TOKEN_PROTOCOL_PREFIX = 'kueueviz.auth.';
+
+const encodeTokenForProtocol = (token) =>
```
mwielgus, line 25: Do we really need to encode token? Isn't it in base64/jwt already?

cmd/kueueviz/backend/middleware/auth.go
samzong: fixed.

charts/kueue/templates/kueueviz/clusterrole.yaml
samzong: thanks, I use right way to regen the chart files.

cmd/kueueviz/frontend/src/useWebSocket.js
samzong: In RFC 6455 don't suppor `.`, so it's need to encode.

PR #2102, opened 2024-04-30
run schedule at most once every 10ms

gabesaba:
#### What type of PR is this?

/kind bug
/kind failing-test

#### What this PR does / why we need it:
In the test environment, we observed many logs due to completing the scheduling loop very fast and entering it again immediately. Now, we will wait 10ms. Additionally, we switch to [NonSlidingUntilWithContext](https://pkg.go.dev/k8s.io/apimachinery@v0.29.4/pkg/util/wait#NonSlidingUntilWithContext), so that we don't waste additional time waiting when under load.

#### Which issue(s) this PR fixes:
Fixes #2097

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #9697, opened 2026-03-06
Scheduling: reduce the maximal backoff time from 100ms to 10ms

### LINKED ISSUE #9716: Reduce the sleep time between scheduling cycles when no admission
**What happened**:

When Kueue is processing multiple inadmissble workloads it needs to "drill" the CQ in the search of admissible workloads. However, the 100ms takes a substantial part of the 1s window we have for the drilling. 

The original increase to 100ms was dictated by our CI rather than users, see: https://github.com/kubernetes-sigs/kueue/pull/2102. From our experience StrictFIFO is not really used on production systems. Also, the amount of logs at 10ms backoff does not seem to grow dramatically anyway from my experiments in https://github.com/kubernetes-sigs/kueue/pull/9697

I would propose to reduce the max backoff slip to 10ms. I think this is still safe from the excessive logging perspective, while reducing the sleep to only 1% of the time between workload requeues (1s with the rate limiting done in https://github.com/kubernetes-sigs/kueue/issues/8095

The sleep time between scheduling cycles should be reduced.

mimowo:
#### What type of PR is this?

/kind cleanup

We need to sleep less to enable good performance for BestEffortFIFO

Fixes #9716

Experiment for now to test log sizes, because this used to be some issue fixed by: https://github.com/kubernetes-sigs/kueue/pull/2102

I think reducing even below 10ms for BestEffortFIFO makes sense, but maybe could cause excessive logging for StrictFIFO, so I'm think we can defer it for later , based on the observations of some deployments.

I checked the log sizes for the integration tests before and after the change:
`pull-kueue-test-integration-baseline-main` suite:
- before: 21518306, [example build](https://gcsweb.k8s.io/gcs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/9619/pull-kueue-test-integration-baseline-main/2029979609021288448/artifacts/)
- after: 24467290, [example build](https://gcsweb.k8s.io/gcs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/9697/pull-kueue-test-integration-baseline-main/2029983361132400640/artifacts/ )
- change (24467290-21518306)/21518306 = 13%

`pull-kueue-test-integration-extended-main` suite:
- before: 17971802, [example build](https://gcsweb.k8s.io/gcs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/9619/pull-kueue-test-integration-extended-main/2029979609134534656/artifacts/)
- after: 19414231, [example build](https://gcsweb.k8s.io/gcs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/9697/pull-kueue-test-integration-extended-main/2029983361195315200/artifacts/ )
- change (19414231-17971802)/17971802 = 8%

Still, they are far from the observed problematic threshold of 100Mb.

```release-note
Scheduling: Reduced the maximum sleep time between scheduling cycles from 100ms to 10ms.
This change fixes a bug where the 100ms delay was excessive on busy systems, in which completed
workloads can trigger requeue events every second. In such cases, the scheduler could spend up to 10%
of the time between requeue events sleeping. Reducing the delay allows the scheduler to spend more time
progressing through the ClusterQueue heap between requeue events.
```

### DIFF
--- a/pkg/util/wait/backoff.go
+++ b/pkg/util/wait/backoff.go
@@ -86,9 +86,16 @@ const (
 	KeepGoing SpeedSignal = true
 	// SlowDown signals to backoff.
 	SlowDown SpeedSignal = false
+)
 
+// Constants configuring the scheduler's backoff between cycles.
+// This mechanism prevents busy-looping and excessive logging when no progress is made.
+const (
+	// initialBackoff is the base backoff reset after a successful cycle.
 	initialBackoff = time.Millisecond
-	maxBackoff     = time.Millisecond * 10
+
+	// maxBackoff is the maximum backoff used when repeated cycles make no progress.
+	maxBackoff = 10 * time.Millisecond
 )
 
 func (s *speedyBackoffManager) toggleBackoff(speedSignal SpeedSignal) {


pkg/util/wait/backoff_test.go
```diff
@@ -105,7 +105,7 @@ func TestUntilWithBackoff(t *testing.T) {
 				i++
 				return signal
 			}
-			untilWithBackoff(ctx, f, timer)
+			untilWithBackoff(ctx, 100*time.Millisecond, f, timer)
```
gabesaba: can we just update the tests instead?

mimowo: Sure, I can. 

I didn't do it initially, because I was thinking:
1. we may need to have a feature gate, and then test both variants. 
2. have a more sophisticated logic in the future with the maxBackoff depending on the presence of StrictFIFO, etc

I'm totally ok to just simplify for now.

mimowo: Done, ptal

pkg/util/wait/backoff.go
```diff
@@ -88,7 +88,7 @@ const (
 	SlowDown SpeedSignal = false
 
 	initialBackoff = time.Millisecond
```
mwielgus, line 95: Add some comments about what these backoff mean and why they are there.

mimowo: sure, added some

PR #9698, opened 2026-03-06
Skip equivalent inadmissible workloads in BestEffortFIFO scheduling [SchedulingEquivalenceHashing]

### LINKED ISSUE #9694: Skip identical workloads in CQ processing

A Hash based blacklist of already processed scheduling constraints. 

I imagine that this could work in the following way:

* Every incoming workload will generate a strict, deterministic hash based on its exact resource requests, node selectors, affinity rules, tolerations, and any other fields that impact scheduling.
* During a CQ evaluation iteration, if a workload is deemed inadmissible, its hash is immediately added to an in-memory, volatile blacklist hash set.
* As the queue is processed, the scheduler checks the hash of each subsequent workload; if it matches a blacklisted hash, the workload is bypassed without undergoing a full evaluation.
* The blacklist hash set is cleared and rebuilt whenever the inadmissible workloads are "added back" to CQ for reevaluation.

**Why is this needed**:

In best effort FIFO CQ with high churn and mixed workloads (like combination of CPU and GPU), Kueue has problems in reaching deep enough into the CQ before reevaluation is triggered. As the result, some workloads may not get admitted while there is enough capacity for them in the cluster. 

The number of different scheduling requirements (pod count, pod requirements, node selector) is usually significantly smaller than the number of workloads. Not doing unnecessary evaluations by skipping the already processed requirements will allow to reach much deeper into Best Effort FIFO CQs and improve cluster utilization.

**Completion requirements**:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

sohankunkerkar:
When the scheduler evaluates a workload and gets NoFit from FlavorAssigner, bulk-move all heap workloads with the same scheduling hash to inadmissible. This reduces scheduling cost from O(workloads) to O(equivalence classes), allowing the scheduler to reach schedulable workloads deep in the queue.

#### What type of PR is this?
/kind feature

Fixes #9694 

```release-note
Scheduling: Fix a BestEffortFIFO performance issue where many equivalent workloads could
prevent the scheduler from reaching schedulable workloads deeper in the queue. Kueue now
skips redundant evaluation by bulk-moving same-hash workloads to inadmissible when one
representative is categorized as NoFit.
```

### DIFF
--- a/pkg/cache/queue/cluster_queue.go
+++ b/pkg/cache/queue/cluster_queue.go
@@ -99,6 +99,10 @@ type ClusterQueue struct {
 	// inadmissibleWorkloads are workloads that have been tried at least once and couldn't be admitted.
 	inadmissibleWorkloads inadmissibleWorkloads
 
+	// noFitSchedulingHashes tracks scheduling equivalence classes that received NoFit.
+	// Cleared when queueInadmissibleWorkloads runs.
+	noFitSchedulingHashes sets.Set[string]
+
 	finishedWorkloads sets.Set[workload.Reference]
 
 	// popCycle identifies the last call to Pop. It's incremented when calling Pop.
@@ -201,6 +205,7 @@ func newClusterQueueImpl(ctx context.Context, client client.Client, wo workload.
 	return &ClusterQueue{
 		heap:                      *heap.New(workloadKey, lessFunc),
 		inadmissibleWorkloads:     make(inadmissibleWorkloads),
+		noFitSchedulingHashes:     sets.New[string](),
 		finishedWorkloads:         sets.New[workload.Reference](),
 		queueInadmissibleCycle:    -1,
 		compareFunc:               compareFunc,
@@ -276,6 +281,12 @@ func (c *ClusterQueue) PushOrUpdate(wInfo *workload.Info) {
 		c.inadmissibleWorkloads.delete(key)
 	}
 	if c.heap.GetByKey(key) == nil && !c.backoffWaitingTimeExpired(wInfo) {
+		c.inadmissibleWorkloads.insert(key, wInfo)
+		return
+	}
+	// Skip to inadmissible if the workload's equivalence class is already known to be NoFit
+	// (only for BestEffortFIFO; StrictFIFO preserves strict ordering).
+	if c.queueingStrategy == kueue.BestEffortFIFO && c.heap.GetByKey(key) == nil && wInfo.SchedulingHash != workload.SchedulingHashUnknown && c.noFitSchedulingHashes.Has(wInfo.SchedulingHash) {
 		c.inadmissibleWorkloads.insert(key, wInfo)
 		return
 	}
@@ -387,6 +398,29 @@ func (c *ClusterQueue) forgetInflightByKey(key workload.Reference) {
 	if c.inflight != nil && workload.Key(c.inflight.Obj) == key {
 		c.inflight = nil
 	}
+}
+
+// handleInadmissibleHash bulk-moves all heap workloads matching the given
+// scheduling hash to inadmissibleWorkloads. Returns the number moved.
+// Only applies to BestEffortFIFO queues; in StrictFIFO the head workload
+// stays in the heap and must not cause equivalent workloads to be skipped.
+func (c *ClusterQueue) handleInadmissibleHash(hash string) int {
+	c.rwm.Lock()
+	defer c.rwm.Unlock()
+	if c.queueingStrategy != kueue.BestEffortFIFO {
+		return 0
+	}
+	c.noFitSchedulingHashes.Insert(hash)
+	moved := 0
+	for _, wInfo := range c.heap.List() {
+		if wInfo.SchedulingHash == hash {
+			key := workloadKey(wInfo)
+			c.heap.Delete(key)
+			c.inadmissibleWorkloads.insert(key, wInfo)
+			moved++
+		}
+	}
+	return moved
 }
 
 // PendingTotal returns the total number of pending workloads.
--- a/pkg/cache/queue/manager.go
+++ b/pkg/cache/queue/manager.go
@@ -585,6 +585,23 @@ func (m *Manager) RequeueWorkload(ctx context.Context, info *workload.Info, reas
 	return added
 }
 
+// HandleInadmissibleHash bulk-moves all workloads in the ClusterQueue's heap
+// that share the given scheduling hash to inadmissibleWorkloads.
+func (m *Manager) HandleInadmissibleHash(cqName kueue.ClusterQueueReference, hash string) int {
+	m.RLock()
+	defer m.RUnlock()
+	cq := m.hm.ClusterQueue(cqName)
+	if cq == nil {
+		return 0
+	}
+	moved := cq.handleInadmissibleHash(hash)
+	if moved > 0 {
+		// Update pending metrics for the CQ and all its LocalQueues.
+		reportPendingWorkloads(m, cqName)
+	}
+	return moved
+}
+
 // Delete the workload from queue or cluster queue.
 // Does not remove the queue assignment caching.
 func (m *Manager) DeleteWorkload(log logr.Logger, wlKey workload.Reference) {
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -270,6 +270,12 @@ func (s *Scheduler) schedule(ctx context.Context) wait.SpeedSignal {
 
 		if mode == flavorassigner.NoFit {
 			log.V(3).Info("Skipping workload as FlavorAssigner assigned NoFit mode")
+			if features.Enabled(features.SchedulingEquivalenceHashing) && e.SchedulingHash != workload.SchedulingHashUnknown {
+				if moved := s.queues.HandleInadmissibleHash(e.ClusterQueue, e.SchedulingHash); moved > 0 {
+					log.V(2).Info("Bulk-moved equivalent workloads to inadmissible",
+						"hash", e.SchedulingHash, "movedCount", moved)
+				}
+			}
 			continue
 		}
 
--- a/pkg/workload/workload.go
+++ b/pkg/workload/workload.go
@@ -18,6 +18,8 @@ package workload
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/json"
 	"fmt"
 	"maps"
 	"slices"
@@ -48,6 +50,7 @@ import (
 	"sigs.k8s.io/kueue/pkg/resources"
 	"sigs.k8s.io/kueue/pkg/util/api"
 	clientutil "sigs.k8s.io/kueue/pkg/util/client"
+	utilpod "sigs.k8s.io/kueue/pkg/util/pod"
 	"sigs.k8s.io/kueue/pkg/util/podset"
 	"sigs.k8s.io/kueue/pkg/util/priority"
 	utilptr "sigs.k8s.io/kueue/pkg/util/ptr"
@@ -64,6 +67,9 @@ const (
 	StatusQuotaReserved = "quotaReserved"
 	StatusAdmitted      = "admitted"
 	StatusFinished      = "finished"
+
+	// SchedulingHashUnknown indicates the scheduling hash could not be computed.
+	SchedulingHashUnknown = "unknown"
 )
 
 var (
@@ -204,6 +210,11 @@ type Info struct {
 	// popped this workload for evaluation. Used by PushOrUpdate to detect spec
 	// changes masked by RequeueWorkload's info.Update.
 	LastEvaluatedGeneration int64
+
+	// SchedulingHash identifies the workload's scheduling equivalence class.
+	// Workloads with the same hash have identical scheduling-relevant shape
+	// and will receive the same FlavorAssigner result given the same cluster state.
+	SchedulingHash string
 }
 
 type PodSetResources struct {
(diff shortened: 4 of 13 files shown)


pkg/workload/workload.go
```diff
@@ -277,13 +285,44 @@ func NewInfo(w *kueue.Workload, opts ...InfoOption) *Info {
...
+		podSetShapes = append(podSetShapes, map[string]any{
+			"name":            ps.Name,
+			"spec":            utilpod.SpecShape(&ps.Template.Spec),
+			"count":           effectiveCount,
+			"minCount":        ps.MinCount,
+			"topologyRequest": ps.TopologyRequest,
+		})
+	}
+	shape := map[string]any{
+		"podSets": podSetShapes,
```
mimowo: We need to ensure "priority" is part of the hash. Please also make sure we have a test for that.

site/content/en/docs/reference/metrics.md
```diff
@@ -36,6 +36,7 @@ Use the following metrics to monitor the status of your ClusterQueues:
 | --- | --- | --- | --- |
 | `kueue_admission_checks_wait_time_seconds` | Histogram | The time from when a workload got the quota reservation until admission, per 'cluster_queue' | `cluster_queue`: the name of the ClusterQueue<br> `priority_class`: the priority class name<br> `replica_role`: one of `leader`, `follower`, or `standalone` |
 | `kueue_admission_cycle_preemption_skips` | Gauge | The number of Workloads in the ClusterQueue that got preemption candidates but had to be skipped because other ClusterQueues needed the same resources in the same cycle | `cluster_queue`: the name of the ClusterQueue<br> `replica_role`: one of `leader`, `follower`, or `standalone` |
+| `kueue_admission_equivalence_skipped_workloads_total` | Counter | Total number of workloads skipped from scheduling evaluation because an equivalent workload was already inadmissible | `cluster_queue`: name of the ClusterQueue<br> `replica_role`: one of `leader`, `follower`, or `standalone` |
```
mimowo: No strong view here, the metric is very important, but maybe we work on it in a follow up - so that we can focus on reviewing the core changes. The follow up could be merged soon after, wdyt?

mimowo: Also, thinking about metrics I think we should have a guage metric to tell the number of equivalence cases, wdyt? Maybe on some environments there are many

pkg/workload/workload.go
sohankunkerkar: Priority is already included via `utilpod.SpecShape()` which hashes `podSpec.Priority`. Let me add a test case  confirming workloads with different priorities produce different hashes.

site/content/en/docs/reference/metrics.md
sohankunkerkar: Makes sense, removed the metric from this PR. Will add it in a follow-up.

pkg/workload/workload.go
mimowo: hm but Workload level priority is independent of PodSets. It can be set but the kueue.k8s-x.io/priority-class label and sits in spec.priority

sohankunkerkar: Good point, will add `wl.Spec.Priority` to the hash shape and a test for it

pkg/cache/queue/cluster_queue.go
```diff
@@ -279,6 +284,11 @@ func (c *ClusterQueue) PushOrUpdate(wInfo *workload.Info) {
 		c.inadmissibleWorkloads.insert(key, wInfo)
 		return
 	}
+	// Skip to inadmissible if the workload's equivalence class is already known to be NoFit.
+	if c.heap.GetByKey(key) == nil && wInfo.SchedulingHash != "" && c.noFitSchedulingHashes.Has(wInfo.SchedulingHash) {
```
mimowo: Why we would check ` wInfo.SchedulingHash != ""`. I think the has should be assigned to every Info, or is this a sanity check?

pkg/cache/queue/cluster_queue.go
```diff
@@ -389,6 +399,27 @@ func (c *ClusterQueue) forgetInflightByKey(key workload.Reference) {
 	}
 }
 
+// recordInadmissibleHash bulk-moves all heap workloads matching the given
+// scheduling hash to inadmissibleWorkloads. Returns the number moved.
+func (c *ClusterQueue) recordInadmissibleHash(hash string) int {
```
mimowo: Rename to `handleInadmissibleHash`  to reflect this is more than just recording.

pkg/cache/queue/manager.go
```diff
@@ -584,6 +584,23 @@ func (m *Manager) RequeueWorkload(ctx context.Context, info *workload.Info, reas
 	return added
 }
 
+// RecordInadmissibleHash bulk-moves all workloads in the ClusterQueue's heap
+// that share the given scheduling hash to inadmissibleWorkloads.
+func (m *Manager) RecordInadmissibleHash(cqName kueue.ClusterQueueReference, hash string) int {
```
mimowo:
```suggestion
func (m *Manager) HandleInadmissibleHash(cqName kueue.ClusterQueueReference, hash string) int {
```
as in the sibling function, because the primary role of the function is the bulk move

pkg/scheduler/scheduler.go
```diff
@@ -270,6 +270,12 @@ func (s *Scheduler) schedule(ctx context.Context) wait.SpeedSignal {
 
 		if mode == flavorassigner.NoFit {
 			log.V(3).Info("Skipping workload as FlavorAssigner assigned NoFit mode")
+			if e.SchedulingHash != "" {
```
mimowo: Do we need this check? I think the hash will be computed for all workloads, right?

I would prefer not to use it as "just in case of a bug check", because ii pollutes code long term.

If we need a safety mechanism we could re-evaluate a feature gate.

pkg/workload/workload.go
```diff
@@ -277,13 +285,45 @@ func NewInfo(w *kueue.Workload, opts ...InfoOption) *Info {
...
 		info.TotalRequests = totalRequestsFromPodSets(w, &options)
 	}
+	info.SchedulingHash = computeSchedulingHash(w, info.TotalRequests)
 	return info
 }
 
+// Update refreshes the object reference without recomputing derived fields.
+// Spec changes are handled by creating a fresh Info via NewInfo.
 func (i *Info) Update(wl *kueue.Workload) {
 	i.Obj = wl
```
mimowo: Hm, I would also expect the hash to be re-computed here. What if there is say a priority change on update?

Maybe we could trigger the re-compute only on Generation change, but changing the Workload's priority is pretty important as some users already use it to boost some workloads which don't schedule for long.

pkg/workload/workload.go
```diff
@@ -277,13 +285,45 @@ func NewInfo(w *kueue.Workload, opts ...InfoOption) *Info {
...
+			"topologyRequest": ps.TopologyRequest,
+		})
+	}
+	shape := map[string]any{
+		"podSets":  podSetShapes,
+		"priority": wl.Spec.Priority,
+	}
+	shapeJSON, err := json.Marshal(shape)
+	if err != nil {
+		return ""
```
mimowo: Oh I see this is the situation why are the skips for empty hash. I would propose to introduce a constant (maybe "ErrorOnHashComputation", or "HashUnknown") and check it in the places, rather than the cryptic "".

pkg/scheduler/scheduler.go
mimowo: Ok, I see what this is for, we could introduce a const for that: https://github.com/kubernetes-sigs/kueue/pull/9698/changes#r2894312925

pkg/workload/workload.go
mimowo: Please make sure the error is logged

pkg/cache/queue/manager.go
```diff
@@ -584,6 +584,23 @@ func (m *Manager) RequeueWorkload(ctx context.Context, info *workload.Info, reas
...
+	m.RLock()
+	defer m.RUnlock()
+	cq := m.hm.ClusterQueue(cqName)
+	if cq == nil {
+		return 0
+	}
+	moved := cq.recordInadmissibleHash(hash)
+	if moved > 0 {
```
mwielgus, line 598: Why is this needed? Numer of pending workloads shouldn't change.

pkg/workload/workload.go
```diff
@@ -277,13 +285,45 @@ func NewInfo(w *kueue.Workload, opts ...InfoOption) *Info {
...
+			"count":           effectiveCount,
+			"minCount":        ps.MinCount,
+			"topologyRequest": ps.TopologyRequest,
+		})
+	}
+	shape := map[string]any{
+		"podSets":  podSetShapes,
+		"priority": wl.Spec.Priority,
+	}
+	shapeJSON, err := json.Marshal(shape)
```
mwielgus, line 328: Can you log shapeJson and end hash at some veeeery high debug level?

mwielgus (commented):
Can we put this entire thing behind a feature gate? Just in case the hashing was blacklisting too much.

pkg/cache/queue/manager.go
sohankunkerkar: Actually, The total pending doesn't change, but the active/inadmissible breakdown does and both are reported as separate gauge values. Without this, the active gauge stays stale until the next Pop().

pkg/workload/workload.go
sohankunkerkar: ok, added a V(5) log of the shape JSON and computed hash.

pkg/workload/workload.go
sohankunkerkar: ok

pkg/workload/workload.go
```diff
@@ -277,11 +288,47 @@ func NewInfo(w *kueue.Workload, opts ...InfoOption) *Info {
...
+	if err != nil {
+		log.Log.Error(err, "Failed to compute scheduling hash", "workload", klog.KObj(wl))
```
mimowo: Could we pass the contextualized logger?

pkg/workload/workload.go
```diff
@@ -277,11 +288,47 @@ func NewInfo(w *kueue.Workload, opts ...InfoOption) *Info {
...
+		return SchedulingHashUnknown
+	}
+	hash := fmt.Sprintf("%x", sha256.Sum256(shapeJSON))[:16]
+	log.Log.V(5).Info("Computed scheduling hash", "workload", klog.KObj(wl), "hash", hash)
```
mimowo: Log the Json also. I think it might be safer to check the logging level first by` log.V(5).Enabled()`

pkg/workload/workload.go
```diff
@@ -277,11 +288,47 @@ func NewInfo(w *kueue.Workload, opts ...InfoOption) *Info {
...
+// to reflect any changes (e.g., priority updates).
 func (i *Info) Update(wl *kueue.Workload) {
 	i.Obj = wl
+	i.SchedulingHash = computeSchedulingHash(wl, i.TotalRequests)
+}
+
+// computeSchedulingHash returns a deterministic hash of the workload's
+// scheduling-relevant shape: workload priority, pod spec (via SpecShape),
+// effective count, minCount, and topologyRequest per PodSet.
+func computeSchedulingHash(wl *kueue.Workload, totalRequests []PodSetResources) string {
```
mimowo: short circuit the function if SchedulingEquivalenceHashing is disabled.

pkg/workload/workload.go
sohankunkerkar: `NewInfo()` and `Update()` don't currently take a context or logger and plumbing one through would change the function signature across many call sites. The V(5) log already includes `klog.KObj(wl)` for workload identity. Happy to add it if you think the signature change is worth it, but leaning toward keeping it simple with the global logger for now.

mimowo: Could you submit a preparatory PR which propagetes the contextual logger? We could log whatever at V(5). For example, Update() could log that the Workload info is updated.

pkg/workload/workload.go
```diff
@@ -277,12 +288,53 @@ func NewInfo(w *kueue.Workload, opts ...InfoOption) *Info {
...
+	info.SchedulingHash = computeSchedulingHash(w, info.TotalRequests)
 	return info
 }
 
+// Update refreshes the object reference and recomputes the scheduling hash
+// to reflect any changes (e.g., priority updates).
 func (i *Info) Update(log logr.Logger, wl *kueue.Workload) {
 	log.V(5).Info("Workload info updated", "workload", klog.KObj(wl))
 	i.Obj = wl
+	i.SchedulingHash = computeSchedulingHash(wl, i.TotalRequests)
```
mimowo: could you pass the logger already here? Then you could probably temporarily pass the ctrl.Log from `NewInfo`, until passed in the other PR?

sohankunkerkar: ack, will add that

pkg/workload/workload.go
```diff
@@ -277,12 +288,53 @@ func NewInfo(w *kueue.Workload, opts ...InfoOption) *Info {
 	} else {
 		info.TotalRequests = totalRequestsFromPodSets(w, &options)
 	}
+	info.SchedulingHash = computeSchedulingHash(log.Log, w, info.TotalRequests)
```
mimowo, line 291: Please follow up with passing the contextual logger

mimowo (commented):
Thank you, looks great 👍 
/lgtm
/approve

Please prepare the cherrupicks, adjusting the Version for the feature gate per branch.

pkg/workload/workload.go
```diff
@@ -277,12 +288,53 @@ func NewInfo(w *kueue.Workload, opts ...InfoOption) *Info {
...
+		return SchedulingHashUnknown
+	}
+	podSetShapes := make([]map[string]any, 0, len(wl.Spec.PodSets))
+	for i, ps := range wl.Spec.PodSets {
+		effectiveCount := ps.Count
+		if i < len(totalRequests) {
+			effectiveCount = totalRequests[i].Count
+		}
+		podSetShapes = append(podSetShapes, map[string]any{
+			"name":            ps.Name,
```
olekzabl, line 317: @sohankunkerkar Apologies for asking after this was merged, but I'm very curious: 
Why does a PodSet name contribute to the hash?

sohankunkerkar: Good question! The PodSet name doesn't affect FlavorAssigner's Fit/NoFit decision, so including it here is conservative, it won't cause incorrect bulk-moves but does make the hash slightly more specific than strictly necessary. In practice it's a no-op since job frameworks use deterministic names, so same job type always produces the same hash.

mimowo:
/retitle Skip equivalent inadmissible workloads in BestEffortFIFO scheduling [SchedulingEquivalenceHashing]

PR #9615, opened 2026-03-02
Fix stuck pods bug for node replacement

j-skiba:
#### What type of PR is this?
/kind bug

#### What this PR does / why we need it:

https://github.com/kubernetes-sigs/kueue/pull/8997#discussion_r2863877179

After pods are assigned to some node by topology assignment and before are ungated (and given nodeSelector for that specific node) it might happen that in between these two events a node can be tainted or become NotReady. In such scenario these pods would be stuck in Pending state as they are bound to a node by nodeSelector but will not be scheduled on a node because of a taint or NotReady condition.

Pods would be stuck like this and replacement pods would not be created because the existing pending pods are still waiting to be scheduled.

This PR introduces mechanism of waiting for all pods expected to be on a node. Only then when all pods are there we proceed with unhealthy node handling.

#### Which issue(s) this PR fixes:
Fixes #9616

#### Does this PR introduce a user-facing change?

```release-note
TAS: Fixed a bug where pods could become stuck in a `Pending` state during node replacement.
This may occur when a node gets tainted or `NotReady` after the topology assignment phase, but before
the pods are ungated.
```

PR #9710, opened 2026-03-06
KEP-2724 - Update handling of pending pods for node replacement feature

### LINKED ISSUE #9616: Terminate forever pending pods for node replacement
**What happened**:

After pods are assigned to some node by topology assignment and before are ungated (and given `nodeSelector` for that specific node) it might happen that in between these two events a node can be tainted or become `NotReady`. In such scenario these pods would be stuck in `Pending` state as they are bound to a node by `nodeSelector` but will not be scheduled on a node because of a taint or `NotReady` condition.

 Pods would be stuck like this and replacement pods would not be created because the existing pending pods are still waiting to be scheduled.

That is necessary for node replacement mechanism to work as expected. Currently topology assignment for a workload will be updated but the pods will stay `Pending` forever until deleted manually.

We should handle such pods as described here - https://github.com/kubernetes-sigs/kueue/blob/main/keps/2724-topology-aware-scheduling/README.md#tainted-nodes-treatment

https://github.com/kubernetes-sigs/kueue/pull/9615 - shown here as an e2e test

---

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

j-skiba:
The current implementation of the node replacement feature in TAS contains several bugs and race conditions that might cause problems with node replacement:

- Stuck Pending Pods: When a workload is assigned a specific topology, and a node within that topology becomes tainted or enters a NotReady state, the resulting pods are still created with node selectors pointing to that node. Consequently, the Kubernetes scheduler rightfully refuses to schedule them. This leaves the pods stuck in a `Pending` state indefinitely unless manually intervened.

- Race Conditions in Node Failure Detection: The node failure controller currently reacts to node taints or `NotReady` condition without ensuring all of a workload's pods have actually been created. Evaluating early against an empty pod list leads to falsely concluding that all pods are terminated, prematurely marking the node as unhealthy. To act predictably, the controller must first wait for a stable state where all pods have been created and settled (accounting for aspects like `tolerationSeconds`).

- Event-Driven Correctness over Polling: While establishing an initial stable state is necessary, relying on polling to manage that state is fundamentally flawed because underlying controllers act instantaneously. For example: suppose a stable state is reached where one pod is stuck in `Pending` due to a taint. The controller transitions that pod to `Failed` and proceeds to mark the node as unhealthy. Immediately, the Job controller creates a new replacement pod. Before the node replacement is fully completed, this new pod can also get stuck in `Pending` on the same tainted node. A polling-based approach will likely miss this newly created pod, leaving it orphaned and permanently "out of watch". By transitioning to an event-driven model, the controller reacts continuously to Pod events, instantly catching and handling these fast-arriving replacement pods.

- Expanded Condition Support: The original design heavily assumed these edge-cases only occurred with `NoSchedule` taints. In reality, these issues persist equally with NoExecute taints and `NotReady` states. The event-driven implementation unifies the handling of all three node conditions.
    
Fixes #9616
    
```release-note
NONE
```

### DIFF
--- a/keps/2724-topology-aware-scheduling/README.md
+++ b/keps/2724-topology-aware-scheduling/README.md
@@ -1406,27 +1406,25 @@ We introduce the `TASReplaceNodeOnNodeTaints` feature gate from v0.17 as Beta, a
 When enabled, Kueue treats tainted nodes as unhealthy. This applies to nodes with `NoExecute` taint,
 or nodes with `NoSchedule` taint where all pods of the workload running on that node are failing, terminating, or in unscheduled state.
 
-- **NoExecute**: Nodes with the `NoExecute` effect taint, that is not tolerated by the workload, are considered unhealthy.
+- **NoExecute**: Nodes with the `NoExecute` taint, that is not tolerated by the workload, are considered unhealthy.
 The pods on such nodes are expected to be terminated by the node controller. Once terminated, Kueue will attempt
 to replace the node if `TASFailedNodeReplacement` is enabled, and evict the workload if no replacement is possible.
 If `tolerationSeconds` is specified, Kueue waits for the duration before treating the node as unhealthy.
-- **NoSchedule**: Nodes with the `NoSchedule` effect taint, that is not tolerated by the workload, are considered unhealthy
+- **NoSchedule**: Nodes with the `NoSchedule` taint, that is not tolerated by the workload, are considered unhealthy
 only if all pods of the workload that have topology assignment to that node are terminating, in the failed state,
 or if they are unscheduled. In this case, Kueue can trigger node replacement.
 
-While `NoSchedule` taint does not evict running pods (unlike `NoExecute`), Kueue triggers recovery for unscheduled or failed pods.
-Kueue also performs node replacement if pods fail or terminate after `NoSchedule` is assigned, ensuring the workload is not blocked by unschedulable Pods which are assigned to a tainted node.
+While the `NoSchedule` taint does not evict running pods (unlike `NoExecute`), Kueue triggers recovery for unscheduled or failed pods.
+Kueue also performs node replacement if pods fail or terminate after the `NoSchedule` taint is assigned, ensuring the workload is not blocked by unschedulable Pods which are assigned to a tainted node.
 
-Notably the analogous problem exists not just for Pods assigned to nodes tainted with  `NoSchedule` , but also nodes with the untolerated `NoExecute` taint, or with `NotReady` state. 
+Notably the analogous problem exists not just for Pods assigned to nodes tainted with the `NoSchedule` taint, but also nodes with the untolerated `NoExecute` taint, or in the `NotReady` state. 
 
-In order to remediate the problem, for workloads where single Node replacement is possible, the node failure controller transitions these `Pending` Pods to the `Failed` phase so that replacement Pods are created.
-
-Kueue will not terminate the Pods if there is more than one node requiring replacement, because this scenario is handled by full workload eviction.
+In order to remediate the problem, for workloads where a single Node replacement is possible, the node failure controller transitions these `Pending` Pods to the `Failed` phase so that replacement Pods are created.
 
 This ensures that the pods are re-created by the Job controller for placement on other nodes, while keeping the original
-pods in Failed state for debuggability. Without this step, the pending pods would block the creation of replacement pods.
+pods in the `Failed` state for debuggability. Without this step, the pending pods would block the creation of replacement pods.
 
-When transitioning these Pods, the controller appends the following condition:
+When transitioning the `Pending` Pods, the controller appends the following condition:
 ```yaml
 type: TerminatedByKueue
 status: True
@@ -1434,9 +1432,11 @@ reason: UnschedulableOnAssignedNode
 message: "..."
 ```
 
-In addition, Kueue emits a Normal event with reason `PodTerminated` on the Pod to inform about the termination.
+In addition, Kueue emits a Normal event with the reason `PodTerminated` on Pod termination.
 
-Nodes with `.spec.unschedulable` set to true are treated as having `NoSchedule` taint.
+Kueue will not terminate a `Pending` Pod if there is more than one node requiring replacement, because this scenario is handled by a full workload eviction.
+
+Nodes with `.spec.unschedulable` set to true are treated as having the `NoSchedule` taint.
 
 ##### User stories
 


keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1415,25 +1414,28 @@ If `tolerationSeconds` is specified, Kueue waits for the duration before treatin
...
-  status: True
-  reason: UnschedulableOnAssignedNode
-  message: "..."
-  ```
-  This ensures that the pods are re-created by the Job controller for placement on other nodes, while keeping the original
-  pods in Failed state for debuggability. Without this step, the pending pods would block the creation of replacement pods.
+For workloads where single Node replacement is possible, the controller remediates this scenario by reacting to Node and Pod events:
+
+- **On Node events:** When a node becomes unhealthy (due to a taint or the `NotReady` state), the controller evaluates the Pods on the node. If their status indicates an issue, the node is marked as unhealthy.
+- **On Pod events:** When a node is already tainted or `NotReady`, a change in a Pod's status can cause the node to fulfill the unhealthy requirements, allowing the controller to mark the node as unhealthy. Crucially, reacting to Pod events enables the controller to easily filter out Pods that were created too slowly or became stuck in the `Pending` phase. The controller transitions these Pods to the `Failed` phase so that replacement Pods are created—an event-driven approach that completely avoids polling.
```
mimowo: I think we don't need to describe here the details "how" the controller does things. 

In paricular, I'm not sure we need to describe reacting to Pod / Node events with such detail. Such things are likely to change / evolve. In paricular, I don't see reacting to Pod events in the implementation: https://github.com/kubernetes-sigs/kueue/pull/9615.

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1415,25 +1414,28 @@ If `tolerationSeconds` is specified, Kueue waits for the duration before treatin
...
 
-  In addition, Kueue emits a Normal event with reason `PodTerminated` on the Pod to inform about the termination.
+When transitioning these Pods, the controller appends the following condition:
+```yaml
+type: TerminatedByKueue
+status: True
+reason: UnschedulableOnAssignedNode
+message: "..."
+```
+This ensures that the pods are re-created by the Job controller for placement on other nodes, while keeping the original
```
olekzabl: I think I'm not following. Wouldn't just marking the old Pods `Failed` trigger some from of "recovery", in particular creating new Pods which will then get assigned to some other nodes?

Why is this the `TerminatedByKueue` condition needed, actually?

keps/2724-topology-aware-scheduling/README.md
olekzabl: > to easily filter out

Maybe this doesn't need expanding in this PR, but at least for my better understanding: what's this "easy" way of filtering out? Checking the timestamp when the node was marked unhealthy, and filtering out Pods whose Conditions changed after that?

j-skiba: I removed the implementation details

j-skiba: Currently the node failure controller is node event driven and it makes scenarios when pods are created too slowly hard to handle - we have to requeue after some period and it's hard to tell when to stop doing requeues. 

Nevertheless I removed that part because these are basically implementation details

keps/2724-topology-aware-scheduling/README.md
j-skiba: Yes, marking them as failed triggers creation of a replacement pod. The condition is added to make the termination of the pods traceable.

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1415,25 +1414,25 @@ If `tolerationSeconds` is specified, Kueue waits for the duration before treatin
 only if all pods of the workload that have topology assignment to that node are terminating, in the failed state,
 or if they are unscheduled. In this case, Kueue can trigger node replacement.
 
-  While `NoSchedule` taint does not evict running pods (unlike `NoExecute`), Kueue triggers recovery for unscheduled or failed pods.
-  If a workload is assigned to a node that is tainted with `NoSchedule`, the pods will remain in a pending state because
-  the Kubernetes scheduler will not schedule them on that node.
+While `NoSchedule` taint does not evict running pods (unlike `NoExecute`), Kueue triggers recovery for unscheduled or failed pods.
+Kueue also performs node replacement if pods fail or terminate after `NoSchedule` is assigned. Since the taint may indicate
+an underlying node problem, replacement ensures the workload is not blocked by the unavailable node.
```
mimowo:
```suggestion
Kueue also performs node replacement if pods fail or terminate after `NoSchedule` is assigned, ensuring the workload is not blocked by unschedulable Pods which are assigned to a tainted node.
```
to avoid saying "unavoidable" node which may be misiterpreted.

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1415,25 +1414,25 @@ If `tolerationSeconds` is specified, Kueue waits for the duration before treatin
...
 
-  Kueue also performs node replacement if pods fail or terminate after `NoSchedule` is assigned. Since the taint may indicate
-  an underlying node problem, replacement ensures the workload is not blocked by the unavailable node.
+Pods belonging to a workload might remain in the `Pending` phase if Kueue calculates a topology assignment that includes a specific node, but that node subsequently becomes tainted with `NoSchedule` or `NoExecute`, or enters the `NotReady` state. If a Pod is created or updated with a node selector pointing to this node, the Kubernetes scheduler is unable to schedule it due to the node's condition. Consequently, the Pod remains bound to an unavailable node by its topology assignment.
```
mimowo:
```suggestion
Notably the analogous problem exists not just Pods assigned to nodes tainted with  `NoSchedule` , but also nodes with the untolerated `NoExecute` taint, or with `NotReady` state. 
```

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1415,25 +1414,25 @@ If `tolerationSeconds` is specified, Kueue waits for the duration before treatin
...
-because they cannot run due to a taint, Kueue marks the pods as `Failed` and adds the following condition to the pods:
-  ```yaml
-  type: TerminatedByKueue
-  status: True
-  reason: UnschedulableOnAssignedNode
-  message: "..."
-  ```
-  This ensures that the pods are re-created by the Job controller for placement on other nodes, while keeping the original
-  pods in Failed state for debuggability. Without this step, the pending pods would block the creation of replacement pods.
+For workloads where single Node replacement is possible, the node failure controller remediates this scenario by waiting for all pods that are expected to be on the node to be created, and then running the unhealthy handling procedures. The controller transitions these `Pending` Pods to the `Failed` phase so that replacement Pods are created.
```
mimowo:
```suggestion
In order to remediate the problem, for workloads where single Node replacement is possible, the node failure controller transitions these `Pending` Pods to the `Failed` phase so that replacement Pods are created.
```

mimowo: I think here (or below) is also the right place to mention, that Kueue will not terminate the Pods if there is more than one node requiring replacement, because this scenario is handled by full workload eviction.

keps/2724-topology-aware-scheduling/README.md
olekzabl: Ok, this makes sense. Then I think I was misled by the last sentence here:

> _Without this step, the pending pods would block the creation of replacement pods._

I interpreted "this step" as "adding the `TerminatedByKueue` condition". As I understand now, this is not the intended meaning.

Could it be clearer if you moved that sentence higher up, placing it right after this one:

> _The controller transitions these `Pending` Pods to the `Failed` phase so that replacement Pods are created. Without this step, the pending pods would block the creation of replacement pods._

?

j-skiba: Yes, that sentence was misleading. I moved it as you said

keps/2724-topology-aware-scheduling/README.md
j-skiba: done

keps/2724-topology-aware-scheduling/README.md
j-skiba: updated

keps/2724-topology-aware-scheduling/README.md
j-skiba: updated

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1415,25 +1414,27 @@ If `tolerationSeconds` is specified, Kueue waits for the duration before treatin
...
-  type: TerminatedByKueue
-  status: True
-  reason: UnschedulableOnAssignedNode
-  message: "..."
-  ```
-  This ensures that the pods are re-created by the Job controller for placement on other nodes, while keeping the original
-  pods in Failed state for debuggability. Without this step, the pending pods would block the creation of replacement pods.
+In order to remediate the problem, for workloads where single Node replacement is possible, the node failure controller transitions these `Pending` Pods to the `Failed` phase so that replacement Pods are created.
+
+Kueue will not terminate the Pods if there is more than one node requiring replacement, because this scenario is handled by full workload eviction.
```
olekzabl: Now I think this should be moved - but it's tricky where to :)
Perhaps to the very bottom of your change?
Right now, it seems to poorly fit in various other places because:

- the next paragraph starts with "this", which does not relate to _this_ paragraph, but to previous one
- the following paragraph says "these Pods", and again it's not about _these pods_ from this pragraph
- the subsequent paragraph speaks about "the Pod", and again it's not about Pods from this paragraph

Or maybe you can make some other rewording; just please consider how it reads "end to end".

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1407,35 +1406,37 @@ We introduce the `TASReplaceNodeOnNodeTaints` feature gate from v0.17 as Beta, a
...
-because they cannot run due to a taint, Kueue marks the pods as `Failed` and adds the following condition to the pods:
-  ```yaml
-  type: TerminatedByKueue
-  status: True
-  reason: UnschedulableOnAssignedNode
-  message: "..."
-  ```
-  This ensures that the pods are re-created by the Job controller for placement on other nodes, while keeping the original
-  pods in Failed state for debuggability. Without this step, the pending pods would block the creation of replacement pods.
+In order to remediate the problem, for workloads where a single Node replacement is possible, the node failure controller transitions these `Pending` Pods to the `Failed` phase so that replacement Pods are created.
```
tenzen-y, line 1422: @j-skiba @mimowo Can we ignore these Kueue triggering Failures by the Pod Failure Policy?
In general, the Job can ignore such Node Disruption, then the Job controller (KCM) doesn't count the failure and doesn't affect the backoffLimit.

mimowo: Yes, but we would leave it to the user. A user can specify the following rule:

```yaml
podFailurePolicy:
  rules:
  - action: Ignore
    onPodConditions:
    - type: TerminatedByKueue
      status: true
```
Still, some users may prefer to increase backoffLimit actually. I don't think this is something we should be adjusting in Kueue. Maybe we can document.

tenzen-y: > Still, some users may prefer to increase backoffLimit actually. I don't think this is something we should be adjusting in Kueue. Maybe we can document.

yes, I agree with you. I think we should documentize how to ignore node failure in Job FailurePoliicy when they use TAS Node Failure controller both in this KEP and user documentations.

Because users expect to ignore node disruptions by the following podFailurePolicy and no way to find that they should specify `TerminatedByKueue` unless they read Kueue implementations.

```yaml
podFailurePolicy:
  rules:
  - action: Ignore
    onPodConditions:
    - type: DisruptionTarget
```

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1407,35 +1406,37 @@ We introduce the `TASReplaceNodeOnNodeTaints` feature gate from v0.17 as Beta, a
...
+message: "..."
+```
+
+In addition, Kueue emits a Normal event with the reason `PodTerminated` on Pod termination.
 
-  In addition, Kueue emits a Normal event with reason `PodTerminated` on the Pod to inform about the termination.
+Kueue will not terminate a `Pending` Pod if there is more than one node requiring replacement, because this scenario is handled by a full workload eviction.
 
-Nodes with `.spec.unschedulable` set to true are treated as having `NoSchedule` taint.
+Nodes with `.spec.unschedulable` set to true are treated as having the `NoSchedule` taint.
```
tenzen-y, line 1440: I added my suggestions in the following:

```suggestion

The forced transition from `Pending` to `Failed` blocks to ignore PodFailurePolicy  (`type: DisruptionTarget`)in the Job, even though the Pod disruption might be node failures. So, batch users should include (`type: TerminatedByKueue `) in their Job PodFailurePolicy if they want to ignore those failures.
```

Additionally, could you add user documentation to https://kueue.sigs.k8s.io/docs/concepts/topology_aware_scheduling/?

tenzen-y: Feel free to refine my suggestion.

PR #9712, opened 2026-03-06
TAS: Improve scheduling performance by caching nodes

### LINKED ISSUE #9337: Improve scheduling performance by caching nodes

I ran performance tests a few times with bigger node sets than the [current one](https://github.com/kubernetes-sigs/kueue/blob/main/test/performance/scheduler/configs/tas/generator.yaml) as currently we support 60k nodes in TAS.

It seems like for bigger clusters the biggest bottleneck for the scheduling are GC procedures and [node listing](https://github.com/kubernetes-sigs/kueue/blob/321e6e909eaa095e65906653396f22ccad2b2033/pkg/cache/scheduler/tas_flavor.go#L117). As we approach bigger number of nodes the GC procedures consume more time (pictures attached) given complexity of node object.

We could improve scheduling time by caching nodes. It seems like we utilize just a few fields from the node object: `Name`, `Allocatable`, `Labels` and `Taints` except for [here](https://github.com/kubernetes-sigs/kueue/blob/ffe623479f67eb689af8e16e61f979e8ae42f8db/pkg/cache/scheduler/tas_flavor_snapshot.go#L1398) where we match node affinity and we need a node object - here maybe we could "recreate" a node object manually based on the stored `Name` and `Labels` and use it for comparison, I'm not sure if that's fine.

so I would propose a cached node object to look like this:
```go
type NodeInfo struct {
	// Name holds the node's name, used to evaluate node affinity.
	Name string

	// Allocatable capacity from Status.Allocatable.
	Allocatable corev1.ResourceList

	// Labels are used to match Topology levels and NodeSelectors.
	Labels map[string]string

	// Taints are used to check tolerations.
	Taints []corev1.Taint
}
```

I ran 4 scenarios with cpu and heap profiling:

[the current configuration](https://github.com/kubernetes-sigs/kueue/blob/main/test/performance/scheduler/configs/tas/generator.yaml):
- cpu
Image
- heap alloc_objects
Image

[2k nodes configuration](https://gist.github.com/j-skiba/0764c6d1cec7f32bc0d08b88ccaffd31):
- cpu
Image
- heap alloc_objects
Image

[20k nodes configuration](https://gist.github.com/j-skiba/f53a74e94aafd67421134721f7e62ff4):
- cpu
Image
- heap alloc_objects
Image

[60k nodes configuration](https://gist.github.com/j-skiba/6b397714dc445ecb3ef99157dc7230f1):
- cpu
Image
- heap alloc_objects
Image

Similar issues: https://github.com/kubernetes-sigs/kueue/issues/8449

**Why is this needed**:

Improvement of scheduling performance for bigger scale clusters.

**Completion requirements**:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

mbobrovskyi:
#### What type of PR is this?
/kind feature

Improve scheduling performance by caching nodes.

Fixes #9337

```release-note
TAS: Fix performance bug where snapshotting would take very long due to List and DeepCopy
of all Nodes. Now the cached set of nodes is maintained in event-driven fashion.
```

### DIFF
--- a/pkg/cache/scheduler/tas_cache.go
+++ b/pkg/cache/scheduler/tas_cache.go
@@ -33,18 +33,17 @@ import (
 type tasCache struct {
 	sync.RWMutex
 	client      client.Client
-	nodes       map[string]*nodeInfo
 	flavors     map[kueue.ResourceFlavorReference]flavorInformation
 	topologies  map[kueue.TopologyReference]topologyInformation
 	flavorCache map[kueue.ResourceFlavorReference]*TASFlavorCache
 
 	nonTasUsageCache *nonTasUsageCache
+	nodesCache       *nodesCache
 }
 
 func NewTASCache(client client.Client) tasCache {
 	return tasCache{
 		client:      client,
-		nodes:       make(map[string]*nodeInfo),
 		flavors:     make(map[kueue.ResourceFlavorReference]flavorInformation),
 		topologies:  make(map[kueue.TopologyReference]topologyInformation),
 		flavorCache: make(map[kueue.ResourceFlavorReference]*TASFlavorCache),
@@ -52,6 +51,7 @@ func NewTASCache(client client.Client) tasCache {
 			podUsage: make(map[types.NamespacedName]podUsageValue),
 			lock:     sync.RWMutex{},
 		},
+		nodesCache: newNodesCache(),
 	}
 }
 
@@ -130,22 +130,10 @@ func (t *tasCache) DeletePodByKey(key client.ObjectKey) {
 	t.nonTasUsageCache.delete(key)
 }
 
-func (t *tasCache) AddOrUpdateNode(node *corev1.Node) {
-	t.Lock()
-	defer t.Unlock()
-	if !node.Spec.Unschedulable && utiltas.IsNodeStatusConditionTrue(node.Status.Conditions, corev1.NodeReady) {
-		t.nodes[node.Name] = newNodeInfo(node)
-	} else {
-		t.deleteNodeWithoutLock(node.Name)
-	}
+func (t *tasCache) SyncNode(node *corev1.Node) {
+	t.nodesCache.sync(node)
 }
 
-func (t *tasCache) DeleteNode(nodeName string) {
-	t.Lock()
-	defer t.Unlock()
-	t.deleteNodeWithoutLock(nodeName)
-}
-
-func (t *tasCache) deleteNodeWithoutLock(nodeName string) {
-	delete(t.nodes, nodeName)
+func (t *tasCache) DeleteNodeByName(nodeName string) {
+	t.nodesCache.delete(nodeName)
 }
--- a/pkg/cache/scheduler/tas_flavor.go
+++ b/pkg/cache/scheduler/tas_flavor.go
@@ -23,8 +23,6 @@ import (
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/selection"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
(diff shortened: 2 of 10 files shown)


pkg/controller/tas/resource_flavor.go
```diff
@@ -97,49 +96,42 @@ func (r *rfReconciler) setupWithManager(mgr ctrl.Manager, cache *schdcache.Cache
...
+	if oldNode != nil && newNode != nil && checkNodeSchedulingPropertiesChanged(newNode, oldNode) == nodeUnchanged {
+		r.logger().V(5).Info("Skipping node update as scheduling properties are unchanged", "node", newNode.Name)
+		return
+	}
+	if oldNode != nil {
+		r.nodeUpdateCh <- event.GenericEvent{Object: oldNode}
+	}
+	if newNode != nil {
+		r.nodeUpdateCh <- event.GenericEvent{Object: newNode}
+	}
```
vladikkuzn, line 112:
```go
	select {
        case r.nodeUpdateCh <- event.GenericEvent{Object: newNode}:
    default:
        r.logger().V(3).Info("Node update channel full, dropping new node event", "node", newNode.Name)
    }
```

mbobrovskyi: What does it mean? I didn't got.

mbobrovskyi:
/retest

Know issue https://github.com/kubernetes-sigs/kueue/issues/9752.

pkg/controller/tas/resource_flavor.go
olekzabl, line 112: I'm assuming that, if `newNode == oldNode`, we rely on something inside kube-controller library to deduplicate the reconciliations (at least usually), is that right?

pkg/cache/scheduler/tas_cache.go
```diff
@@ -127,3 +129,23 @@ func (t *tasCache) Update(pod *corev1.Pod, log logr.Logger) {
...
 	t.nonTasUsageCache.delete(key)
 }
+
+func (t *tasCache) AddOrUpdateNode(node *corev1.Node) {
+	t.Lock()
+	defer t.Unlock()
+	if !node.Spec.Unschedulable && utiltas.IsNodeStatusConditionTrue(node.Status.Conditions, corev1.NodeReady) {
+		t.nodes[node.Name] = newNodeInfo(node)
+	} else {
+		t.deleteNodeWithoutLock(node.Name)
```
olekzabl: Nit: it's a bit counter-intuitive that `AddOrUpdateX` may also delete X.
Though I don't have ideas how to rename this method... so feel free to ignore.

pkg/controller/tas/resource_flavor.go
mbobrovskyi: This is just an event. We handle it inside [nodeHandler](https://github.com/epam/kubernetes-kueue/blob/4718ce7071570b6180fc601638fc8bba853be4b2/pkg/controller/tas/resource_flavor.go#L130-L143). [Here](https://github.com/epam/kubernetes-kueue/blob/4718ce7071570b6180fc601638fc8bba853be4b2/pkg/controller/tas/resource_flavor.go#L138-L140), we are adding it to the workqueue.
And yes, controller-runtime should execute it only once, even if we have multiple events.

pkg/cache/scheduler/tas_cache.go
mbobrovskyi: Maybe SyncNode()? WDYT?

olekzabl: Sounds good. In particular, `SyncAdmittedCondition` in `scheduler.go` sounds like a "precedent for this convention".

Thanks.

pkg/controller/tas/resource_flavor.go
mbobrovskyi: Done

olekzabl: I _guess_ what @vladikkuzn meant was to explicitly handle the case when the channel buffer is full (i.e. there are 10 pending events). Then, your current code would wait until the buffer becomes not full. The proposed code would instead explicitly discard the change - in order not to block.

(In particular, I _don't think_ it's a proposal to no longer consider `oldNode`).

---

However, I don't see this practice applied anywhere else, so I personally _don't support_ this comment.

pkg/cache/scheduler/tas_flavor.go
```diff
@@ -136,10 +117,44 @@ func (c *TASFlavorCache) TopologyLevels() []string {
...
+			reqs = append(reqs, *req)
+		}
+	}
+
+	// 3. Create a single selector from all combined requirements.
+	selector := labels.NewSelector().Add(reqs...)
+
+	// 4. Filter the nodes.
+	filteredNodes := make([]*nodeInfo, 0, len(nodes))
+	for _, node := range nodes {
```
mimowo: This access to nodes is not protected by the read lock, right? 

I would rather introduce a dedicacted struct protected by R/W lock for nodes. Rather than relying on higher level locks.

pkg/cache/scheduler/tas_cache.go
```diff
@@ -127,3 +130,35 @@ func (t *tasCache) Update(pod *corev1.Pod, log logr.Logger) {
...
+func (t *tasCache) findNodes(nodeLabels map[string]string, levels []string) []*nodeInfo {
+	t.RLock()
+	defer t.RUnlock()
+	filteredNodes := make([]*nodeInfo, 0, len(t.nodes))
+	for _, node := range t.nodes {
+		if utilnode.NodeMatchesFlavor(node.Labels, nodeLabels, levels) {
+			filteredNodes = append(filteredNodes, node)
+		}
+	}
+	return filteredNodes
```
mimowo: nit, but I'm wondering if we could first clone the nodes inside the lock, then use slices.DeleteFunc outside the lock. It would move the match checking outside the lock.

We could also just move the nodes under a dedicated thread safe struct which has Clone

pkg/cache/scheduler/tas_cache.go
```diff
@@ -127,3 +130,35 @@ func (t *tasCache) Update(pod *corev1.Pod, log logr.Logger) {
 func (t *tasCache) DeletePodByKey(key client.ObjectKey) {
 	t.nonTasUsageCache.delete(key)
 }
+
+func (t *tasCache) SyncNode(node *corev1.Node) {
+	t.Lock()
+	defer t.Unlock()
+	if !node.Spec.Unschedulable && utiltas.IsNodeStatusConditionTrue(node.Status.Conditions, corev1.NodeReady) {
```
mimowo: nit, this line does not need to be under lock. However, we may keep it if it is tricky to refactor, ptal

pkg/util/tas/node.go
```diff
@@ -0,0 +1,33 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package node
+
+// NodeMatchesFlavor checks if a node's labels match the required labels
+// and contains all required topology levels. Returns true if matches.
+func NodeMatchesFlavor(nodeLabels map[string]string, requiredLabels map[string]string, requiredLevels []string) bool {
```
mimowo, line 21: We already have this capability in the `nodeBelongsToFlavor` in the tas controllers package. 

I would propose de-duplicate in this PR  by using the new function from the old place. 


I'm just thinking `pkg/util/tas/` might be a better location for the file, because it belongs there functionally, and so people will search it there. "Node" itself does not know about "Flavor" concept, only TAS.

mbobrovskyi: > We already have this capability in the nodeBelongsToFlavor in the tas controllers package.

Yeah, I moved this function (https://github.com/kubernetes-sigs/kueue/pull/9712/changes#diff-7f688e0706bb6b6775d35b320ee0d496f92ed367d87cb845979731898ea9d1eaL208-L221).

> I'm just thinking pkg/util/tas/ might be a better location for the file, because it belongs there functionally, and so people will search it there. "Node" itself does not know about "Flavor" concept, only TAS.

Good point. Moved.

pkg/cache/scheduler/tas_cache.go
mbobrovskyi: Done

pkg/cache/scheduler/tas_flavor.go
mbobrovskyi: Done

mimowo:
Thank you 👍 
/lgtm
/approve
/cherrypick release-0.16
/cherrypick release-0.15

mimowo:
@mbobrovskyi Please prepare the cherrypicks manually in case of conflicts

PR #9721, opened 2026-03-06
Implement MultiKueue single-cluster preemption guarantee

### LINKED ISSUE #8303: MultiKueue: prevent starting preemptions in multiple worker clusters

A mechanism that prevents starting preemptions for the same workload in multiple clusters in the same time. 

**Why is this needed**:

A submission of a big, high-priority workload may trigger cluster-wide preemptions in multiple clusters. All but one will be disrupting other workloads for nothing.

**Completion requirements**:

- [x] Design doc
- [x] API change (maybe)
- [ ] Docs update

kshalot:
#### What type of PR is this?
/kind feature
/area multikueue

Fixes #8303

```release-note
MultiKueue: Add the  `MultiKueueOrchestratedPreemption` alpha feature gate proposed in KEP-8303, which enables the orchestration of preemptions within MultiKueue worker clusters, guaranteeing that the preemptions are executed sequentially.
Introduce a new `PreemptionGate` API to `Workload`, which governs whether a workload can execute preemptions.
```

### DIFF
--- a/pkg/scheduler/scheduler_test.go
+++ b/pkg/scheduler/scheduler_test.go
@@ -8122,7 +8122,7 @@ func TestSchedule(t *testing.T) {
 				utiltesting.MakeEventRecord("default", "fs-high-pob", "Pending", corev1.EventTypeWarning).Obj(),
 			},
 		},
-		"block preemptions when a preemption gate is present": {
+		"block preemptions and signal `BlockedOnPreemptionGates` when a preemption gate is present": {
 			enableMultiKueueOrchestratedPreemption: true,
 			workloads: []kueue.Workload{
 				*utiltestingapi.MakeWorkload("preemptor", "eng-beta").
@@ -8184,6 +8184,123 @@ func TestSchedule(t *testing.T) {
 			},
 			wantAssignments: map[workload.Reference]kueue.Admission{
 				"eng-beta/low-priority": {
+					ClusterQueue: "eng-beta",
+					PodSetAssignments: []kueue.PodSetAssignment{
+						utiltestingapi.MakePodSetAssignment("main").
+							Assignment("example.com/gpu", "model-a", "20").
+							Obj(),
+					},
+				},
+			},
+			wantLeft: map[kueue.ClusterQueueReference][]workload.Reference{
+				"eng-beta": {"eng-beta/preemptor"},
+			},
+		},
+		"do not signal `BlockedOnPreemptionGates` when a preemption gate is present, but the workload fits without preemption": {
+			enableMultiKueueOrchestratedPreemption: true,
+			workloads: []kueue.Workload{
+				*utiltestingapi.MakeWorkload("preemptor", "eng-beta").
+					UID("wl-preemptor").
+					JobUID("job-preemptor").
+					Queue("main").
+					Request("example.com/gpu", "20").
+					PreemptionGates(kueue.PreemptionGate{Name: "gate"}).
+					Obj(),
+			},
+			wantWorkloads: []kueue.Workload{
+				*utiltestingapi.MakeWorkload("preemptor", "eng-beta").
+					UID("wl-preemptor").
+					JobUID("job-preemptor").
+					Queue("main").
+					Request("example.com/gpu", "20").
+					PreemptionGates(kueue.PreemptionGate{Name: "gate"}).
+					ReserveQuotaAt(utiltestingapi.MakeAdmission("eng-beta").
+						PodSets(utiltestingapi.MakePodSetAssignment(kueue.DefaultPodSetName).
+							Assignment("example.com/gpu", "model-a", "20").
+							Obj()).
+						Obj(), now).
+					Condition(metav1.Condition{
+						Type:               kueue.WorkloadAdmitted,
+						Status:             metav1.ConditionTrue,
+						Reason:             "Admitted",
+						Message:            "The workload is admitted",
+						LastTransitionTime: metav1.NewTime(now),
+					}).
+					Condition(metav1.Condition{
+						Type:               kueue.WorkloadQuotaReserved,
+						Status:             metav1.ConditionTrue,
+						Reason:             "QuotaReserved",
+						Message:            "Quota reserved in ClusterQueue eng-beta",
+						LastTransitionTime: metav1.NewTime(now),
+					}).
+					Obj(),
+			},
+			wantAssignments: map[workload.Reference]kueue.Admission{
+				"eng-beta/preemptor": {
+					ClusterQueue: "eng-beta",
+					PodSetAssignments: []kueue.PodSetAssignment{
+						utiltestingapi.MakePodSetAssignment("main").
+							Assignment("example.com/gpu", "model-a", "20").
+							Obj(),
+					},
+				},
+			},
+		},
+		"do not signal `BlockedOnPreemptionGates` when a preemption gate is present, but the workload had nothing to preempt": {
+			enableMultiKueueOrchestratedPreemption: true,
+			workloads: []kueue.Workload{
+				*utiltestingapi.MakeWorkload("preemptor", "eng-beta").
+					UID("wl-preemptor").
+					JobUID("job-preemptor").
+					Queue("main").
+					Request("example.com/gpu", "20").
+					PreemptionGates(kueue.PreemptionGate{Name: "gate"}).
+					Obj(),
+				*utiltestingapi.MakeWorkload("high-priority", "eng-beta").
+					UID("high-priority").
+					Priority(1).
+					Request("example.com/gpu", "20").
+					ReserveQuotaAt(utiltestingapi.MakeAdmission("eng-beta").
+						PodSets(utiltestingapi.MakePodSetAssignment(kueue.DefaultPodSetName).
+							Assignment("example.com/gpu", "model-a", "20").
+							Obj()).
+						Obj(), now).
+					Obj(),
+			},
+			wantWorkloads: []kueue.Workload{
+				*utiltestingapi.MakeWorkload("high-priority", "eng-beta").
+					UID("high-priority").
+					Priority(1).
+					Request("example.com/gpu", "20").
+					ReserveQuotaAt(utiltestingapi.MakeAdmission("eng-beta").
+						PodSets(utiltestingapi.MakePodSetAssignment(kueue.DefaultPodSetName).
+							Assignment("example.com/gpu", "model-a", "20").
+							Obj()).
+						Obj(), now).
+					Obj(),
+				*utiltestingapi.MakeWorkload("preemptor", "eng-beta").
+					UID("wl-preemptor").
+					JobUID("job-preemptor").
+					Queue("main").
+					Request("example.com/gpu", "20").
+					PreemptionGates(kueue.PreemptionGate{Name: "gate"}).
+					Condition(metav1.Condition{
+						Type:               kueue.WorkloadQuotaReserved,
+						Status:             metav1.ConditionFalse,
+						Reason:             "Pending",
+						Message:            "couldn't assign flavors to pod set main: insufficient unused quota for example.com/gpu in flavor model-a, 20 more needed",
+						LastTransitionTime: metav1.NewTime(now),
+					}).
+					ResourceRequests(kueue.PodSetRequest{
+						Name: "main",
+						Resources: corev1.ResourceList{
+							"example.com/gpu": resource.MustParse("20"),
+						},
+					}).
+					Obj(),
+			},
+			wantAssignments: map[workload.Reference]kueue.Admission{
+				"eng-beta/high-priority": {
 					ClusterQueue: "eng-beta",
 					PodSetAssignments: []kueue.PodSetAssignment{
 						utiltestingapi.MakePodSetAssignment("main").


pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -501,11 +511,24 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
...
+			}
+		}
 	}
 
 	return w.nominateAndSynchronizeWorkers(ctx, group)
 }
 
+func isRemoteSpecOutOfSync(local, remote kueue.WorkloadSpec) bool {
+	remote.PreemptionGates = nil
+	return !equality.Semantic.DeepEqual(local, remote)
```
kshalot, line 542: I also considered doing `DeepDerivative` here which would not report cases like:
```
local.field = nil
remote.field = "value"
```
as diffs. But that could have an impact on the rest of the app.

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -501,11 +511,25 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
...
+				return reconcile.Result{}, fmt.Errorf("failed to update remote workload: %w", err)
+			}
+		}
 	}
 
 	return w.nominateAndSynchronizeWorkers(ctx, group)
 }
 
+func isRemoteSpecOutOfSync(local, remote kueue.WorkloadSpec) bool {
+	remote.PreemptionGates = nil
```
olekzabl, line 541: I don't like the side-effect on `remote`.
I'd consider restoring the previous state before returning - even if it doesn't directly simplify your current caller code.

pkg/controller/admissionchecks/multikueue/workload.go
olekzabl, line 426: If you follow my other comment (on restoring `remote` inside `isRemoteSpecOutOfSync`), you'll be able to do the whole "saving of gates" just around this line. Making the code "more locally self-contained".

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -990,13 +1016,59 @@ func updateDelayedTopologyRequest(local, remote *kueue.Workload) {
...
+		} else {
+			openMkGateState := wl.Status.PreemptionGates[openMkGateStateIdx]
+			gateOpenedTime := openMkGateState.LastTransitionTime
+			if previousUngateTime == nil || gateOpenedTime.After(previousUngateTime.Time) {
+				previousUngateTime = &gateOpenedTime
+			}
+		}
+	}
+
+	// If the timeout elapsed or no workload was ungated yet, return the workload with he oldest signal.
```
olekzabl:
```suggestion
	// If the timeout elapsed or no workload was ungated yet, return the workload with the oldest signal.
```

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -501,11 +511,25 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
 		} else {
 			return reconcile.Result{}, w.updateACS(ctx, group.local, acs, kueue.CheckStateRetry, "Reserving remote lost")
 		}
+	} else if features.Enabled(features.MultiKueueOrchestratedPreemption) {
+		remWl, remClient := w.workloadToOpenPreemptionGate(group)
+
+		if remWl != nil && remClient != nil {
+			workload.SetPreemptionGateState(remWl, constants.MultiKueuePreemptionGate, kueue.GateStateOpen, metav1.NewTime(w.clock.Now()))
```
olekzabl: This implements step 5 from the KEP procedure:

> _5. Mark the `kueue.x-k8s.io/multikueue` gate of the found workload as `Open`._

Then, how about the last step:

> _6. Schedule a reconciliation in `SingleClusterPreemptionTimeout`._

? Are we guaranteed to have this reconciler invoked again after that timeout?

pkg/controller/core/workload_controller.go
```diff
@@ -678,6 +688,40 @@ func (r *WorkloadReconciler) reconcileSyncAdmissionChecks(ctx context.Context, w
 	return false, nil
 }
 
+func (r *WorkloadReconciler) syncPreemptionGateStates(wl *kueue.Workload) bool {
+	changed := false
```
olekzabl, line 704: Thinking loud: what if we extracted this logic to a helper, called from here but also from `SetPreemptionGateState` and maybe also from `HasClosedPreemptionGate`?

(Then you could remove "edge case handling" from those 2 methods)

The not-so-nice part would be that `HasClosedPreemptionGate` would do the "syncing" but then drop it.

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -781,21 +805,23 @@ func (w *wlReconciler) Generic(_ event.GenericEvent) bool {
 func newWlReconciler(c client.Client, helper *admissioncheck.MultiKueueStoreHelper, cRec *clustersReconciler, origin string,
 	recorder record.EventRecorder, workerLostTimeout, eventsBatchPeriod time.Duration,
 	adapters map[string]jobframework.MultiKueueAdapter, dispatcherName string, roleTracker *roletracker.RoleTracker,
+	singleClusterPreemptionTimeout time.Duration,
```
olekzabl: This arg list is long, and you've just added a 2nd `time.Duration` to it :(
I'd consider refactoring, e.g. into a named-field struct.
(Pretty much copying the signature of `wlReconciler`, just with public fields through which we can pass the initial values to the current private fields).
Unless you have another idea.

pkg/controller/admissionchecks/multikueue/workload_test.go
```diff
@@ -1521,6 +1523,243 @@ func TestWlReconcile(t *testing.T) {
...
+						Type:               kueue.WorkloadPreemptionBlocked,
+						Status:             metav1.ConditionTrue,
+						Reason:             kueue.PreemptionGated,
+						Message:            "Preemption gate closed",
+						LastTransitionTime: metav1.NewTime(earlier),
+					}).
+					Obj(),
+			},
+		},
+		"not opening preemption gate if timeout did not elapse": {
```
olekzabl: How about a test case for timeout-driven opening (for the 2nd worker)?

test/integration/multikueue/scheduler/scheduler_test.go
```diff
@@ -523,4 +534,122 @@ var _ = ginkgo.Describe("MultiKueue with scheduler", ginkgo.Label("area:multikue
...
+		workerHighWorkload := &kueue.Workload{}
+
+		ginkgo.By("Checking that the high-priority workload is created in the manager cluster", func() {
+			gomega.Eventually(func(g gomega.Gomega) {
+				g.Expect(managerTestCluster.client.Get(managerTestCluster.ctx, highWlKey, managerHighWl)).To(gomega.Succeed())
+				g.Expect(managerHighWl.Spec.QueueName).To(gomega.BeEquivalentTo(managerLq.Name))
+			}, util.Timeout, util.Interval).Should(gomega.Succeed())
+		})
+
+		ginkgo.By("Checking that the high-priority workload is admitted in exactly one of the workers", func() {
```
olekzabl: IIUC this would be true even without your feature. Even with multiple preemptions, as soon as one of the remotes is admitted, the other remotes are removed.
So I'm not sure if this is worth checking here.

test/integration/multikueue/scheduler/scheduler_test.go
```diff
@@ -523,4 +534,122 @@ var _ = ginkgo.Describe("MultiKueue with scheduler", ginkgo.Label("area:multikue
...
+				g.Expect(w1Requeued != w2Requeued).To(gomega.BeTrue())
+
+				g.Expect(workload.IsEvicted(workerLowW1) != workload.IsEvicted(workerLowW2)).To(gomega.BeTrue())
+				var evictedWorkload *kueue.Workload
+				if workload.IsEvicted(workerLowW1) {
+					evictedWorkload = workerLowW1
+				} else {
+					evictedWorkload = workerLowW2
+				}
+				g.Expect(evictedWorkload.Status.Conditions).To(testing.HaveConditionStatusTrue(kueue.WorkloadPreempted))
```
olekzabl: Nit: the structure of this is a bit surprising. You check for
- exactly one "requeued"
- (independently) exactly one "evicted"
- whether the "evicted" one is also "preempted".

Not sure why some things are independent, others are not.

pkg/controller/admissionchecks/multikueue/workload.go
kshalot: It's pass-by-value though, no? I could pass the reference here, but I figured it's nicer to have this guarantee that it won't change.

test/integration/multikueue/scheduler/scheduler_test.go
kshalot: This is equivalent to saying "workload is admitted". I just wanted to have a step for this to make sure this happens before the test continues. In your opinion, would just saying "Checking that the high-priority workload is admitted" be better? I guess it would. It doesn't promise some "exclusive" admission which is just the way Kueue always works.

pkg/controller/admissionchecks/multikueue/workload.go
kshalot: Like I said in that comment - there is no side effect. I restore the `remote` as low as possible intentionally - I wanted to avoid a situation where someone adds a new case without adjusting the preemption gates.

This way it's clear that **whatever you update**, the preemption gates shouldn't be overriden, rather than suggesting that this is some case specific to the elastic workload.

WDYT?

test/integration/multikueue/scheduler/scheduler_test.go
kshalot: This is a fair point, but there was some reasoning behind checking "requeued".

I wanted to have a condition which will always fail if **both** workers get preempted. Just checking evicted won't cut it, because one of the workers will get re-admitted. So it will be requeued, but will lose the evicted condition.

I restructured this expectation a little bit - I now find one evicted workload and then check that it has the required conditions, while the other workload doesn't have them. Maybe that is clearer?

olekzabl: Yes, this is what I hoped to see. Thanks!

pkg/controller/admissionchecks/multikueue/workload.go
olekzabl: Ah, indeed. 
I think I just didn't expect copying of a full struct...
Well, it's not _huge_: 7 fields, none of them of a sub-struct type.
Looking quickly through our code, I think the standard is rather to take a pointer, though some exceptions already exist (e.g. [here](https://github.com/kubernetes-sigs/kueue/blob/0d4b20498109b6cc3360f366db31e64c2a9e28c8/pkg/workloadslicing/workloadslicing.go#L361)).
So I'm on the edge. No strong opinion.

pkg/controller/admissionchecks/multikueue/workload.go
olekzabl: Makes sense.
(I somehow disliked the fact that we're introducing a special treatment for this very field, so I wanted to scope this "specialty" down as much as possible. But I now think you're right. This special treatment is sometimes needed, and given that, let's apply it broadly, for future safety).

pkg/controller/admissionchecks/multikueue/workload.go
kshalot: Two relatively small structs allocated on the stack seem like a low price to pay to avoid unsetting/resetting in the logic or doing an explicit DeepCopy.

But if it's confusing or goes against some convention, I'm not going to argue.

Also it has the pod sets from what I can see, which are not the smallest thing. It's at most 8 though

pkg/controller/admissionchecks/multikueue/workload_test.go
kshalot: Yep. Good point. Added.

pkg/controller/admissionchecks/multikueue/workload.go
kshalot: Indeed, adding another argument here felt very bad when I was writing this, but I also saw that this is a private function that is called in just a handful of places I think. So I passed the arg and went on with my life.

If you feel strongly about it, then I can consider refactoring this in this PR, but we can also revisit this later (to not extend the diff even further).

pkg/controller/admissionchecks/multikueue/workload.go
kshalot: Indeed - PTAL @ https://github.com/kubernetes-sigs/kueue/pull/9721/commits/7ceb7a59a102d9d04bfa6678664ebb65844c26d7.

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -503,7 +513,31 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
...
+				return reconcile.Result{}, fmt.Errorf("failed to update remote workload: %w", err)
+			}
+
+			// Ensure there is a reconcile that considers the next preemption gate.
+			requeueAfterSynchronize = w.singleClusterPreemptionTimeout
+		}
+	}
+
+	res, err := w.nominateAndSynchronizeWorkers(ctx, group)
+	if err == nil && requeueAfterSynchronize < res.RequeueAfter {
```
olekzabl: AFAICS `nominateASW` can often return `Result{}`, in which `Requeue` will be false and `RequeueAfter` will be 0. Then, this condition will not match, so you'll not schedule a re-run. Is that WAI?

kshalot: Oops, good catch. I added a `res.RequeueAfter != 0` but it's gone, I must have removed it by mistake.

I re-added it, thanks!

olekzabl: Wait... I think [your last commit](https://github.com/kubernetes-sigs/kueue/pull/9721/commits/29b0e76ec93cf22825cef971bc4d82421ca4cea1) doesn't change anything?

If `res.RequeueAfter` is larger than the timeout, then certainly it's `!= 0`, so `&&`-ing this is a no-op.

However, I'd like to schedule a re-run also in the case when `res.RequeueAfter == 0`.

pkg/controller/admissionchecks/multikueue/workload.go
olekzabl: > _But if it's confusing or goes against some convention, I'm not going to argue._

In my feeling, it goes against the dominant trend, though exceptions exist.
Leaving the choice to you.

> _Also it has the pod sets from what I can see, which are not the smallest thing. It's at most 8 though_

FWIW, pod sets are a slice, and my understanding of [this](https://go.dev/tour/moretypes/8) is that value-passing a slice will not lead to copying the actual elements. (If I'm wrong, I'll be happy to be educated).

kshalot: > FWIW, pod sets are a slice, and my understanding of [this](https://go.dev/tour/moretypes/8) is that value-passing a slice will not lead to copying the actual elements. (If I'm wrong, I'll be happy to be educated).

Yes, obviously only the header of the slice will be copied. I went too far there.

> In my feeling, it goes against the dominant trend, though exceptions exist.
Leaving the choice to you.

I prefer leaving it as is, i.e. passing the struct and not the pointer. Mainly because:

> I'd consider restoring the previous state before returning

This doesn't feel clean to me. Another alternative is doing sth like `remote.DeepCopy()`, which beats the purpose completely.

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -410,7 +411,13 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
...
 	// - workloads for which workload priority has changed
 	for rem, remWl := range group.remotes {
-		if remWl != nil && !equality.Semantic.DeepEqual(group.local.Spec, remWl.Spec) {
+		if remWl == nil {
+			continue
+		}
+
+		if isRemoteSpecOutOfSync(group.local.Spec, remWl.Spec) {
+			remotePreemptionGates := remWl.Spec.PreemptionGates
+
```
Singularity23x0: Why split it like this? If you want to keep the nested if you can do the if as before.
If you would like to remove the indentation and use "continue" you could do it as "remWl == nil || isRemoteSpecOutOfSync(...)".
No point in mixing both approaches as their benefits are not having the downsides of the other, while here we have downsides of both (readability wise).

kshalot: Likely leftover from some work-in-progress logic. I inlined it.

pkg/scheduler/scheduler.go
```diff
@@ -279,21 +287,30 @@ func (s *Scheduler) schedule(ctx context.Context) wait.SpeedSignal {
...
-			// later. Otherwise, we allow other workloads
-			// in the Cohort to borrow this capacity,
-			// confident we can reclaim it later.
-			if !preemption.CanAlwaysReclaim(cq) {
-				// reserve capacity up to the
-				// borrowing limit, so that
-				// lower-priority workloads in another
-				// Cohort cannot admit before us.
-				cq.AddUsage(resourcesToReserve(e, cq))
+		if mode == flavorassigner.Preempt {
```
kshalot, line 312: @mimowo We discussed how the scheduling hashes impact this feature. AFAIR we wanted to make `HasClosedPreemptionGate` part of the hash and assign a `NoFit`. Now that I think about it, it might be a bit more complex than that.

If we have two equivalent workloads A and B, both of them having a closed preemption gate, if we start scheduling A and we want to skip B, we'll have to update the `PreemptionBlocked` condition for A and B, otherwise B might not get ungated. But if B gets ungated, it's possible that it won't be scheduled in this cluster for a long time, while it could be scheduled in another one with a preemption. So the B workload could be prevented from making progress for at most `SingleClusterPreemptionTimeout` (because the gate was opened in a cluster where it won't become the head).

We could keep it as is, where the mode is still `Preempt` instead of `NoFit`. I think in terms of semantics, this case is closer to `preemptionTargets == 0` than a regular `NoFit`.

test/e2e/multikueue/e2e_test.go
```diff
@@ -2104,24 +2113,152 @@ var _ = ginkgo.Describe("MultiKueue", func() {
...
+			ginkgo.By("Checking that the first low-priority workload is created in worker1 and not in worker2", func() {
+				gomega.Eventually(func(g gomega.Gomega) {
+					g.Expect(k8sWorker1Client.Get(ctx, lowWlKey1, workerLowW1)).To(gomega.Succeed())
+					g.Expect(workload.IsAdmitted(workerLowW1)).To(gomega.BeTrue())
+
+					g.Expect(k8sWorker2Client.Get(ctx, lowWlKey1, &kueue.Workload{})).To(utiltesting.BeNotFoundError())
+				}, util.Timeout, util.Interval).Should(gomega.Succeed())
+			})
+
+			// Fits only in worker2
```
olekzabl, line 2166: Nit:
AFAICS this generally fits on both workers.
(In particular, it has the same requests as `highJob` below which you described as fitting both).
Though I understand it fits only worker2 _right now_ because worker1's ephemeral storage is currently fully consumed.

One easy way to make it unambiguous is to raise RAM request 0.5G -> 1.5G.

pkg/scheduler/scheduler.go
mimowo: Thank you for raising that. I will probably be able to look deeper into that only on Friday after the patch releases are out. 

In the meanwhile - is it feasible to design an integration/e2e test which would confirm the behavior matches our desired semantics? That would be valuable for this implementation, and also in case we make a mistake and need to refactor.

test/e2e/multikueue/e2e_test.go
kshalot: I changed it to 1.5G but also lowered the ram and CPU for other workloads so the preemption doesn't happen on the manager.

pkg/scheduler/scheduler.go
olekzabl: +1 for keeping it as is. Then, it's rather clear that gated workloads are seen as `Preempt`, not `NoFit`, so they don't "participate in the hashing game", and various edge cases are avoided.

We might reconsider it later, if we see that hashing brings significant gains which suggest "joining the game". But the current code feels like a safe start.

mimowo: I think the hashing is rather a distraction at this point, we can adjust it if needed 

The fundamental question is: do we want the semantics of "Preempt" or "NoFit"? Preempt is actively blocking the head of the queue, while NoFit gives other workloads a chance to schedule. 

At this point I'm not sure yet which semantics fits better the problem we have.

kshalot: I'm writing a test for this, but I think what's missing in this PR is adding a new requeue reason:
```go
RequeueReasonPreemptionGated       RequeueReason = "PreemptionGated"
```

This way the workload will be moved back to the inadmissible list, which IMO is the desired semantics here.

kshalot: FYI I added code that would make this work: https://github.com/kubernetes-sigs/kueue/pull/9721/commits/e5513fddb9421dfc24ea401745436422f03c8249.

As far as tests go, I tried to come up with some MultiKueue-specific way of testing this, but I don't think it's easily doable. I opted for a singlecluster test for the "stickiness" issue: https://github.com/kubernetes-sigs/kueue/pull/9721/commits/a7050fe879d67169d917fabb24bd99b9af679cad

Since the scheduler does not differentiate between a "multikueue gate" and any other gate, this test is even better than a MK-specific one.

test/integration/multikueue/scheduler/scheduler_test.go
```diff
@@ -534,125 +534,137 @@ var _ = ginkgo.Describe("MultiKueue with scheduler", ginkgo.Label("area:multikue
...
+			lowWlKey1     types.NamespacedName
+			managerLowWl1 *kueue.Workload
+			workerLowWl1  *kueue.Workload
+
+			lowWlKey2     types.NamespacedName
+			managerLowWl2 *kueue.Workload
+			workerLowWl2  *kueue.Workload
+		)
+
+		ginkgo.BeforeEach(func() {
```
olekzabl, line 538: Let me just double-check: given how much of the "scenario" you pulled into this `BeforeEach`, is it really intended that you still have a single `It` inside this `When`?

When reading this refactor, I assumed it's purposed for introducing some other test scenario (a similar one so that `BeforeEach` would capture "the common prep") - but this didn't happen.

pkg/controller/core/workload_controller.go
```diff
@@ -1006,6 +1049,13 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 		// and are not supposed to actually change anything.
 		r.cache.AddOrUpdateWorkload(log, wlCopy)
 	}
+
+	if features.Enabled(features.MultiKueueOrchestratedPreemption) {
+		if workload.HasClosedPreemptionGate(e.ObjectOld) && !workload.HasClosedPreemptionGate(e.ObjectNew) {
+			r.queues.QueueAssociatedInadmissibleWorkloadsAfter(ctx, wlKey, nil)
```
olekzabl: LGTM but I wonder why not queue only the workload under concern.
(At the first glance, it sounds like achieving same effect but with less work on the scheduler side).

However, I'm fine if you postpone considering this to after this PR.

test/integration/multikueue/scheduler/scheduler_test.go
kshalot: I was working on a new integration test but dropped it. I then remembered that I already committed an identical structure of the test (`When(is enabled)`) in the e2e test, so I figured I'll leave it for consistency. Since for now there is only one test, we can drop it in both cases.

pkg/controller/core/workload_controller.go
kshalot: Actually, this might not be needed at all... I'm triple checking.

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -503,7 +509,30 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
 		}
 	}
```
Singularity23x0, line 516: Similar to how other steps are numbered and have comments to describe them - this seems like a good place to add another one of those, e.g.:

7. Open preemption gate if orchestrated preemptions are enabled.

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -503,7 +509,30 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
...
+			workload.SetPreemptionGateState(remWl, constants.MultiKueuePreemptionGate, kueue.GateStateOpen, metav1.NewTime(w.clock.Now()))
+			if err := remClient.client.Status().Update(ctx, remWl); err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to update remote workload: %w", err)
+			}
+		}
+		// Ensure there is a reconcile that considers the next preemption gate.
+		requeueAfterSynchronize = requeueIn
+	}
+
+	res, err := w.nominateAndSynchronizeWorkers(ctx, group)
```
Singularity23x0, line 533: Does nominateAndSynchronizeWorkers ever return a non empty res? If not you can just make it return err and set requeueAfter without unnecessary ifs.

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -990,13 +1021,70 @@ func updateDelayedTopologyRequest(local, remote *kueue.Workload) {
...
+		}
+	}
+
+	var timeSinceUngate time.Duration
+	if previousUngateTime == nil {
+		timeSinceUngate = 0
+	} else {
+		timeSinceUngate = w.clock.Now().Sub(previousUngateTime.Time)
+	}
+	timeLeftInTimeout := w.singleClusterPreemptionTimeout - timeSinceUngate
```
Singularity23x0: Nit: a little shorter (left timeSinceUngate in a separate line for readability - up to u if u prefer to roll it into the line below).
```suggestion
	timeLeftInTimeout := w.singleClusterPreemptionTimeout
	if previousUngateTime != nil {
	    timeSinceUngate := w.clock.Now().Sub(previousUngateTime.Time)
		timeLeftInTimeout -= timeSinceUngate
	}
```

pkg/controller/admissionchecks/multikueue/workload.go
kshalot: Right now I think it doesn't, but I wanted to be defensive about it and handle this according to the return type. I don't think it's a huge deal either way.

pkg/controller/admissionchecks/multikueue/workload.go
kshalot: I'm not sure I prefer this actually. I like the explicit `timeSinceUngate = 0` in the original implementation. I like that it's not doing `-=`. Maybe someone else has an opinion, but this is a tiny implementation detail anyway.

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -990,13 +1021,70 @@ func updateDelayedTopologyRequest(local, remote *kueue.Workload) {
 	}
 }
 
-func cloneForCreate(orig *kueue.Workload, origin string) *kueue.Workload {
+func (w *wlReconciler) workloadToOpenPreemptionGate(group *wlGroup) (*kueue.Workload, *remoteClient, time.Duration) {
```
Singularity23x0: Any point to returning both the workload and client here? You can just return the "name" and get the wl and client as needed outside of the "workloadToOpenPreemptionGate" function.
It improves readability as you dont have to do the mental gymnastics of "what combination of nil and ptr can the result of this func have" - just return the name and leave the nil checking to the function calling this one.

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -990,13 +1021,70 @@ func updateDelayedTopologyRequest(local, remote *kueue.Workload) {
...
+	// If the timeout elapsed or no workload was ungated yet, return the workload with the oldest signal.
+	if previousUngateTime == nil || timeLeftInTimeout <= 0 {
+		var rescheduleIn time.Duration
+		if workloadToUngate != nil {
+			rescheduleIn = timeLeftInTimeout
+		}
+		return workloadToUngate, remote, rescheduleIn
+	}
+
+	return nil, nil, timeLeftInTimeout
```
Singularity23x0: Nit: readability impr proposal.
```suggestion
	if previousUngateTime != nil && timeLeftInTimeout > 0 {
	    return nil, nil, timeLeftInTimeout
	}
	if workloadToUngate == nil {
	    return nil, nil, 0
	}
	return workloadToUngate, remote, timeLeftInTimeout
```

kshalot:
> @kshalot sanity-check question (a bit late post KEP merging, but it bothers me): once MultiKueueOrchestratedPreemption graduates to Beta, this would effectively make single preemption the default behavior on all clusters?
> 
> That sounds reasonable to me, but it means semantic change, so I want to make sure we have thought through the consequences and we are comfortable assuming there are no legitimate use cases for concurrent preemptions.
> 
> Asking, because I’m wondering whether, this feature gate should evolve into an explicit opt-in knob such as `PreemptionsMode`: `Concurrent/AtMostOne`, and config like `MaxConcurrentPreemptionsCount: 1/2`.
> 
> I'm also ok to defer the decision to when we graduate to Beta, but I would like to know what is our current stance on that, wdyt?

Thanks for raising that, this should be mentioned in the KEP explicitly but it wasn't. I agree that the feature gate should be replaced with an API in the beta, so there is an explicit opt-in. I'm not sure which mode would be the default yet. Right now, obviously,  the feature gate serves as the binary opt-in mechanism.

I'll update the graduation criteria in the KEP and create an issue.

apis/config/v1beta1/configuration_types.go
```diff
@@ -280,6 +280,11 @@ type MultiKueue struct {
 	// +optional
 	WorkerLostTimeout *metav1.Duration `json:"workerLostTimeout,omitempty"`
 
+	// The timeout after which another worker cluster replica can be ungated, measured since the previous time a replica was ungated.
+	// Defaults to 5 minutes.
+	// +optional
+	SingleClusterPreemptionTimeout *metav1.Duration `json:"singleClusterPreemptionTimeout,omitempty"`
```
mimowo: Maybe we could refrain from making it configuratble in Alpha. The reason is that I foresee when graduating to Beta we will need to introduce more preemption knobs, as discussed in https://github.com/kubernetes-sigs/kueue/pull/9721#issuecomment-4054172653

So, I foresee in the long run we will want to have some grouping of the configuration options, like

```yaml
multiKueue:
  preemptionControl:
    mode: Sequential # or Concurrent
    singleClusterTimeout: 5min
```
wdyt? Do we foresee the need to move the timeout in alpha? I would prefer to avoid the need to restructure the API later on.

kshalot: Yes, in the context of the beta graduation discussion, dropping this configuration option makes perfect sense.

mimowo: ok, please submit a small KEP update then and also simplify the impl pr

pkg/controller/core/workload_controller.go
kshalot: https://github.com/kubernetes-sigs/kueue/pull/9721/commits/28fd78dd497fbe726e3bd174c98149f7c0d048b8 - I thought it wasn't working, that's why I queued the whole queue. But I found that I was just hitting a filter.

test/integration/multikueue/scheduler/scheduler_test.go
kshalot: Actually, I reverted this change. So this is addressed, thanks

pkg/controller/admissionchecks/multikueue/workload.go
kshalot: Possibly a leftover, but I don't remember exactly. I changed it to return the cluster name.

pkg/controller/admissionchecks/multikueue/workload.go
kshalot: Done.

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -990,13 +1021,71 @@ func updateDelayedTopologyRequest(local, remote *kueue.Workload) {
 	}
 }
 
-func cloneForCreate(orig *kueue.Workload, origin string) *kueue.Workload {
+func (w *wlReconciler) workloadToOpenPreemptionGate(group *wlGroup) (*string, time.Duration) {
```
mimowo: The logic here is tricky so please add some logging to at V4 to document the decision making. It will also serve as natural comments.

pkg/scheduler/scheduler.go
```diff
@@ -300,21 +309,30 @@ func (s *Scheduler) schedule(ctx context.Context) wait.SpeedSignal {
...
+					cq.AddUsage(resourcesToReserve(e, cq))
+				}
+				continue
+			}
+
+			if features.Enabled(features.MultiKueueOrchestratedPreemption) && workload.HasClosedPreemptionGate(e.Obj) {
+				gatedMsg := "Workload requires preemption, but it's gated"
+				log.V(3).Info(gatedMsg)
+				setPreemptionGated(e, gatedMsg)
+				continue
```
mimowo, line 334: I like how short the solution is. And I see we only advertise opening the gate if actual preemption is possible. 

Do we have a test case to show that the condition is not added if there are no preemption candidates?

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -410,7 +411,9 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
 	// - elastic workloads which have been scaled down
 	// - workloads for which workload priority has changed
 	for rem, remWl := range group.remotes {
-		if remWl != nil && !equality.Semantic.DeepEqual(group.local.Spec, remWl.Spec) {
+		if remWl != nil && isRemoteSpecOutOfSync(group.local.Spec, remWl.Spec) {
+			remotePreemptionGates := remWl.Spec.PreemptionGates
```
mimowo: nit but wrap with the FG

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -429,6 +432,9 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
 			}
 
 			if updateRemote {
+				// Make sure to not overwrite preemption gates which are managed by the worker.
+				remWl.Spec.PreemptionGates = remotePreemptionGates
```
mimowo: nit, but wrap with FG

pkg/controller/admissionchecks/multikueue/workload.go
mimowo: Also, is this function covered 100% by unit tests? Please check with some golang coverage tool

mimowo (commented):
LGTM, just minor comments to make sure the code is well isolated without side effects, double check test coverage, and to improve logging.

pkg/controller/admissionchecks/multikueue/workload.go
kshalot: This function was 91% covered, I added another test to test in 95%:
```
sigs.k8s.io/kueue/pkg/controller/admissionchecks/multikueue/workload.go:1029:           workloadToOpenPreemptionGate            95.3%
```

I think we won't get to 100%, because the only line which is not "covered" is:
```go
if !wlHasOlderPreemptionSignal { <- covered
    remoteWlLog.V(4).Info("Remote workload was not the first to require preemption") <- not covered
    continue <- not covered
} <- covered
```
We have a test for this (`opening preemption gate for first workload requiring preemption`), but AFAIK sometimes it just won't hit this if-statement because map iteration order is randomized.

pkg/scheduler/scheduler.go
kshalot: Nice spot, I added more cases: https://github.com/kubernetes-sigs/kueue/pull/9721/commits/af344ea93cbd2d021eea86cfa804367931dfd2a4

Specifically:
1. That this does not signal if you fit without preemptions.
2. That this does not signal if you do not fit.

Do we need another case for when there are no preemption targets? So this line specifically:
https://github.com/kubernetes-sigs/kueue/blob/755f97a6382127c3b6952798bdea68a2cdf57504/pkg/scheduler/scheduler.go#L303

I don't know if we have the setup conditions in the unit tests already, but I'll look.

mimowo: > Do we need another case for when there are no preemption targets? So this line specifically:

Yes, I asked the comment to test this line specifically :) 

> I don't know if we have the setup conditions in the unit tests already, but I'll look.

I think it should be possible in scheduler_test.go as a unit test, but please double check

kshalot: Nevermind, for some reason at the moment of writing the previous comment I thought we'll need some more complex setup, but the `Preempt` + `len(preemptionTargets) == 0` is just is for example the case where the preemption target is higher priority.

So `do not signal `BlockedOnPreemptionGates` when a preemption gate is present, but the workload had nothing to preempt` is already testing it 👍

test/e2e/multikueue/e2e_test.go
```diff
@@ -2094,6 +2103,156 @@ var _ = ginkgo.Describe("MultiKueue", func() {
...
+
+					var evictedWorkload *kueue.Workload
+					var otherWorkload *kueue.Workload
+					if workload.IsEvicted(workerLowW1) {
+						evictedWorkload = workerLowW1
+						otherWorkload = workerLowW2
+					} else {
+						evictedWorkload = workerLowW2
+						otherWorkload = workerLowW1
+					}
```
mimowo, line 2244: just a question at this point, if we conclude we could clean up it can be a follow up: could we make the choice of the workload to evict deterministic to avoid that?

kshalot: I thought about it as well, but I did not come up with anything useful.

The high priority workload has to fit in both clusters (so we can check that the preemption is not triggered in both of them). In order for this to be deterministic, we'd have to guarantee that the workload is scheduled, gated and ungated in one cluster before the other.

One direction that we could investigate is changing the `Status.NominatedClusterNames` of the local workload to include only one cluster at first and then the second one. I'm not sure whether that will work cleanly though.

mimowo: I see, no idea for now, but I expect more tests on preemption gates to also have this non-determinism, so I would suggest introducing some helper which would allow to choose the evicted, and non-evicted workloads easier. 

I'm not committed on that, but maybe two 2-fold:
1. first helper to return the clusterName for the cluster on which the preemption happened
2. select workload for the clusterName

mimowo: Let's consider this as a follow up.

PR #9726, opened 2026-03-06
Add podset-replica-sizes annotation on RayJob

hiboyang:
#### What type of PR is this?

/kind feature

This is from discussion in https://github.com/kubernetes-sigs/kueue/pull/9570, add a new annotation to serve following purpose:

1. Update RayJob resourceVersion, which can be used to compose workload name to avoid name conflict
2. For MultiKueue to sync the information back to Management cluster and handle autoscaling scenario

```release-note
NONE
```


pkg/controller/jobs/rayjob/rayjob_controller.go
```diff
@@ -172,15 +178,91 @@ func (j *RayJob) PodSets(ctx context.Context) ([]kueue.PodSet, error) {
...
+			patch := client.MergeFrom(objCopy.DeepCopyObject().(client.Object))
+			annotations := objCopy.GetAnnotations()
+			if annotations == nil {
+				annotations = make(map[string]string)
+			}
+			annotations[PodsetReplicaSizesAnnotation] = string(podSetsJSON)
+			objCopy.SetAnnotations(annotations)
+			if err := reconciler.client.Patch(ctx, objCopy, patch); err != nil {
+				return nil, fmt.Errorf("failed to patch RayJob annotations: %w", err)
+			}
```
mimowo: Instead of that use our helper in pkg/util/client/client.go

pkg/controller/jobs/rayjob/rayjob_controller.go
```diff
@@ -172,15 +178,91 @@ func (j *RayJob) PodSets(ctx context.Context) ([]kueue.PodSet, error) {
...
+	if annotations == nil {
+		return false
+	}
+	existing, ok := annotations[PodsetReplicaSizesAnnotation]
+	if !ok {
+		return false
+	}
+	var existingPodSets []podSetReplicaSize
+	if err := json.Unmarshal([]byte(existing), &existingPodSets); err != nil {
+		return false
```
mimowo: Don't silence parsing errors, I think we should propagate them up

pkg/controller/jobs/rayjob/rayjob_controller.go
```diff
@@ -172,15 +178,91 @@ func (j *RayJob) PodSets(ctx context.Context) ([]kueue.PodSet, error) {
 		return nil, err
 	}
 
+	originalCounts := make(map[kueue.PodSetReference]int32, len(podSets))
+	for _, ps := range podSets {
+		originalCounts[ps.Name] = ps.Count
+	}
```
mimowo: Do we need the originalCounts? I think we could direclty compare the sizes from `podSets` with the annotation, wdyt?

pkg/controller/jobs/rayjob/rayjob_controller.go
```diff
@@ -52,6 +53,11 @@ const (
 	headGroupPodSetName    = "head"
 	submitterJobPodSetName = "submitter"
 	FrameworkName          = "ray.io/rayjob"
+
+	// PodsetReplicaSizesAnnotation is set on the RayJob when autoscaling causes
+	// PodSet replica sizes to differ from the original spec. The value is a JSON
+	// array compatible with []kueue.PodSet, containing only the changed PodSets.
+	PodsetReplicaSizesAnnotation = "kueue.x-k8s.io/podset-replica-sizes"
```
mimowo: We will also need some unit and (integration tests or e2e tests) for the annotation

pkg/controller/jobs/rayjob/rayjob_controller.go
hiboyang: We need originalCounts to remember the counts from RayCluster, compare them with counts from underlying RayCluster, and only add annotation when there is count change.

pkg/controller/jobs/rayjob/rayjob_controller.go
hiboyang: updated

pkg/controller/jobs/rayjob/rayjob_controller.go
hiboyang: good suggestion, updated

pkg/controller/jobs/rayjob/rayjob_controller.go
hiboyang: Yes, added tests

pkg/controller/jobs/rayjob/rayjob_controller.go
mimowo: Hm, let's consider:
1. RayJob size by user was 4, RayJobSpec=4, RayCluster=4, RayJobAnnotation (absent), 
2. Autoscaler scaled up to 6, so RayJobSpec=4, RayCluster=6 -> RayJobAnnotation=6,
3. Autoscaler scales down to 4, so  RayJobSpec=4, RayCluster=4  -> so no "update" so annotation is stale at 6

I think we should instead compare the value of the annotation if changed or not.

hiboyang: Good example, you are right! We need to update annotation to handle this scenario, let me update the code.

mimowo: The code still looks rather complex. I'm wondering that we could:
Retrieve two []podSetReplicaSize - one from annotation (parsed), and another based on `podSets` - then compare. If different, then update the annotation. In this flow we don't need `originalCounts`

hiboyang: This means we always add the annotation, e.g. add the annotation initially before scale-up happening. I can do that. Let me update the code.

mimowo: Indeed, but this is ok, I think.

pkg/controller/jobs/rayjob/rayjob_controller.go
```diff
@@ -178,9 +185,72 @@ func (j *RayJob) PodSets(ctx context.Context) ([]kueue.PodSet, error) {
...
+	previousCounts := make(map[kueue.PodSetReference]int32)
+	if j.Annotations[PodsetReplicaSizesAnnotation] != "" {
+		var annotatedPodSets []podSetReplicaSize
+		if err := json.Unmarshal([]byte(j.Annotations[PodsetReplicaSizesAnnotation]), &annotatedPodSets); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal existing %s annotation: %w", PodsetReplicaSizesAnnotation, err)
+		}
+		for _, ps := range annotatedPodSets {
+			previousCounts[ps.Name] = ps.Count
+		}
+	}
```
mimowo: nit, Move it to a helper function to offload the main function, eg deserializePodSetCounts

pkg/controller/jobs/rayjob/rayjob_controller.go
```diff
@@ -178,9 +185,72 @@ func (j *RayJob) PodSets(ctx context.Context) ([]kueue.PodSet, error) {
...
+
+	// Compare current counts against previous annotation. If any differ, update the annotation
+	// with ALL current podSet counts (not just the changed ones) to avoid infinite patch loops.
+	updatedPodSets := comparePodSetCounts(podSets, previousCounts)
+	if len(updatedPodSets) > 0 && reconciler.client != nil {
```
mimowo: Why checking if `reconciler.client != nil`, this should never happen that the client is nil, afaik

pkg/controller/jobs/rayjob/rayjob_controller.go
```diff
@@ -178,9 +185,72 @@ func (j *RayJob) PodSets(ctx context.Context) ([]kueue.PodSet, error) {
...
+
+// comparePodSetCounts returns the PodSets whose counts differ from referenceCounts.
+func comparePodSetCounts(podSets []kueue.PodSet, referenceCounts map[kueue.PodSetReference]int32) []podSetReplicaSize {
+	var changed []podSetReplicaSize
+	for _, ps := range podSets {
+		if refCount, ok := referenceCounts[ps.Name]; !ok || ps.Count != refCount {
+			changed = append(changed, podSetReplicaSize{
+				Name:  ps.Name,
+				Count: ps.Count,
+			})
```
mimowo: just return true, no need for the extra infomation - it is not used

pkg/controller/jobs/rayjob/rayjob_controller.go
```diff
@@ -178,9 +185,72 @@ func (j *RayJob) PodSets(ctx context.Context) ([]kueue.PodSet, error) {
...
+	updatedPodSets := comparePodSetCounts(podSets, previousCounts)
+	if len(updatedPodSets) > 0 && reconciler.client != nil {
+		allPodSets := make([]podSetReplicaSize, len(podSets))
+		for i, ps := range podSets {
+			allPodSets[i] = podSetReplicaSize{Name: ps.Name, Count: ps.Count}
+		}
+		podSetsJSON, err := json.Marshal(allPodSets)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal updated podsets: %w", err)
+		}
```
mimowo: Extract to helper, like serializePodSetCounts

hiboyang: updated

pkg/controller/jobs/rayjob/rayjob_controller.go
hiboyang: updated

pkg/controller/jobs/rayjob/rayjob_controller.go
hiboyang: removed nil check

pkg/controller/jobs/rayjob/rayjob_controller.go
hiboyang: updated to helper function

pkg/controller/jobs/rayjob/rayjob_controller.go
```diff
@@ -178,9 +185,83 @@ func (j *RayJob) PodSets(ctx context.Context) ([]kueue.PodSet, error) {
...
+
+// comparePodSetCounts returns true if any PodSet count differs from referenceCounts.
+func comparePodSetCounts(podSets []kueue.PodSet, referenceCounts map[kueue.PodSetReference]int32) bool {
+	for _, ps := range podSets {
+		if refCount, ok := referenceCounts[ps.Name]; !ok || ps.Count != refCount {
+			return true
+		}
+	}
+	return false
+}
```
mimowo, line 205: I think there is a subtle bug that it may say not changed when recefrence PodSets have extra elements. To fix we could check also the len() is the same.

However, I'm wondering - maybe the better approach is to use `[]podSetReplicaSize` for both. Then we could simply check `slices.Equal`. wdyt?

hiboyang: Let me add length check. If use `[]podSetReplicaSize`, then will need to consider element ordering, or sort the slice. Length check seems simpler change here.

pkg/controller/jobs/rayjob/rayjob_webhook_test.go
```diff
@@ -574,6 +574,14 @@ func TestValidateCreate(t *testing.T) {
 			features.SetFeatureGateDuringTest(t, features.TopologyAwareScheduling, tc.topologyAwareScheduling)
 			features.SetFeatureGateDuringTest(t, features.ElasticJobsViaWorkloadSlices, tc.elasticJobsViaWorkloadSlices)
 			ctx, _ := utiltesting.ContextWithLog(t)
+
+			fakeClient := utiltesting.NewClientBuilder(rayv1.AddToScheme).
+				WithObjects(tc.job).
+				Build()
+			oldReconciler := reconciler
+			reconciler = rayJobReconciler{client: fakeClient}
+			t.Cleanup(func() { reconciler = oldReconciler })
```
mimowo: Why do we do this? The unit tests for webhook should not depend on the reconciler.

pkg/controller/jobs/rayjob/rayjob_webhook_test.go
```diff
@@ -669,6 +677,14 @@ func TestValidateUpdate(t *testing.T) {
 				manageJobsWithoutQueueName: tc.manageAll,
 			}
 			ctx, _ := utiltesting.ContextWithLog(t)
+
+			fakeClient := utiltesting.NewClientBuilder(rayv1.AddToScheme).
+				WithObjects(tc.newJob).
+				Build()
+			oldReconciler := reconciler
+			reconciler = rayJobReconciler{client: fakeClient}
+			t.Cleanup(func() { reconciler = oldReconciler })
```
mimowo: Same here, it looks weird that unit test for webhooks would depend on reconciler

pkg/controller/jobs/rayjob/rayjob_webhook_test.go
hiboyang: It is because RayJob will need `reconciler.client` to update the annotation. The test code here needs to prepare that client object, otherwise, we will have `reconciler.client` nil error inside RayJob.

pkg/controller/jobs/rayjob/rayjob_webhook_test.go
hiboyang: Similar like previous comment: It is because RayJob will need reconciler.client to update the annotation. The test code here needs to prepare that client object, otherwise, we will have reconciler.client nil error inside RayJob.

pkg/controller/jobs/rayjob/rayjob_webhook_test.go
mimowo: Ahhh, I see, because the validation code calls PodSets, and this code sends update(), we need to change that. The update could be submitted only from the Reconciler. Webhooks cannot have side effects

pkg/controller/jobs/rayjob/rayjob_controller.go
```diff
@@ -178,9 +186,88 @@ func (j *RayJob) PodSets(ctx context.Context) ([]kueue.PodSet, error) {
...
+	if changed {
+		podSetsJSON, err := serializePodSetCounts(podSets)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal updated podsets: %w", err)
+		}
+
+		// Patch the job with the annotation, use a copy of the job object to avoid
+		// it being overwritten by the content returned by the Server.
+		objCopy := j.Object().DeepCopyObject().(client.Object)
+		err = utilclient.Patch(ctx, reconciler.client, objCopy, func() (bool, error) {
```
mimowo: I somehow missed that, but we cannot call the Patch from within PodSets(), because PodSets() is called from webhooks for validation and defaulting. Those need to be super fast, so we cannot afford sending requests. Please move updating the annotation to the Reconcile code only.

hiboyang: Sounds good, this is good suggestion, let me update the code.

pkg/controller/jobs/rayjob/rayjob_webhook_test.go
mimowo: yes, feel free to mark this as resolved, it is the same underlying comment

pkg/controller/jobs/rayjob/rayjob_controller.go
hiboyang: updated the code to move patch outside of PodSets()

pkg/controller/jobs/rayjob/rayjob_webhook_test.go
hiboyang: updated code

pkg/controller/jobframework/reconciler.go
```diff
@@ -868,7 +873,7 @@ func (r *JobReconciler) ensureOneWorkload(ctx context.Context, job GenericJob, o
 
 	// If workload slicing is enabled for this job, use the slice-based processing path.
 	if WorkloadSliceEnabled(job) {
-		podSets, err := JobPodSets(ctx, job)
+		podSets, err := GetJobPodSetsWithUpdateJob(ctx, job, r.client)
```
mimowo: I don't think this is good place to inject it. First this is only needed for RayJob, second PodSets() don't need to be 1:1 the same as JobSet necessarily anyway.

Let me think. I would rather register a new controller for this functionality using NewAdditionalReconcilers. Otherwise, it would need to be opt-in by some dedicated interface like `UpdateAnnotations` maybe.

hiboyang: Remove job patching code from here, added `JobWithCustomAnnotations` interface and RayJob implementation for that interface

pkg/controller/jobframework/reconciler.go
```diff
@@ -68,6 +70,11 @@ import (
 const (
 	FailedToStartFinishedReason = "FailedToStart"
 	managedOwnersChainLimit     = 10
+
+	// PodsetReplicaSizesAnnotation is set on the job when autoscaling causes
+	// PodSet replica sizes to differ from the original spec. The value is a JSON
+	// array compatible with []kueue.PodSet, containing only the changed PodSets.
+	PodsetReplicaSizesAnnotation = "kueue.x-k8s.io/podset-replica-sizes"
```
mimowo, line 77: Since this annotation is only needed for ElasticJobs for now. lets guard it by the feature gate. This way we dont need to commit to maintaining it lomg term. Please also extend the comment saying "This annotation is alpha-level enabled by the ElasticJobsViaWorkloadSlices", ir sonething along the lines

pkg/controller/jobframework/reconciler.go
```diff
@@ -868,11 +875,23 @@ func (r *JobReconciler) ensureOneWorkload(ctx context.Context, job GenericJob, o
...
 		}
 
+		if ca, implements := job.(JobWithCustomAnnotations); implements {
+			if !apiequality.Semantic.DeepEqual(job.Object().GetAnnotations(), oldAnnotations) {
+				err = ca.UpdateAnnotations(ctx, r.client)
+				if err != nil {
+					return nil, fmt.Errorf("failed to update annotations on object %s: %w", object.GetName(), err)
+				}
+			}
+		}
```
mimowo: Since the interface name is generic I would expect the update to also be in the generic place: https://github.com/kubernetes-sigs/kueue/blob/2dc064137bf04449bfc25a14f5dec629dfcc21a5/pkg/controller/jobframework/reconciler.go#L419-L442

Not under the `WorkloadSliceEnabled`. I think inside the helper we should check the feature gate enablement. 

I would also prefer to follow the same pattern as for `CustomWorkloadConditions`, so make the function non-changing. Just informative. 

So I expect something like:

```golang
		if annotations := job.CustomAnnotations(); len(annotations) > 0 {
           // patch using clientutil.Patch()
		}
```
let me know what you think

pkg/controller/jobframework/reconciler.go
```diff
@@ -868,11 +875,23 @@ func (r *JobReconciler) ensureOneWorkload(ctx context.Context, job GenericJob, o
 
 	// If workload slicing is enabled for this job, use the slice-based processing path.
 	if WorkloadSliceEnabled(job) {
+		oldAnnotations := make(map[string]string, len(job.Object().GetAnnotations()))
+		maps.Copy(oldAnnotations, job.Object().GetAnnotations())
```
mimowo: This takes copy of the annotation even though we may not need to update them, I would like to avoid that

pkg/controller/jobframework/reconciler.go
hiboyang: It is doable, but will need to add some state job object (RayJob) to track the annotation change, guess such state will be used by `job.CustomAnnotations()` in your code example. Then there is chance that the job has this state (changed annotations), but the UpdateAnnotations fail and the state becomes stale. Another potential issue is the job is actually `rayv1.RayJob`, we need to figure out a proper way to add such state on `rayv1.RayJob`.

pkg/controller/jobframework/reconciler.go
hiboyang: Updated

pkg/controller/jobframework/reconciler.go
hiboyang: Good suggestion, updated code

pkg/controller/jobframework/reconciler.go
hiboyang: Another alternative is that we keep a copy of job annotations in the beginning of `JobReconciler.ReconcileGenericJob` method, then later near the end of the method, check whether job annotations change, if change, call `UpdateAnnotations`, how do you like this option?

pkg/controller/jobframework/reconciler.go
```diff
@@ -886,6 +880,26 @@ func (r *JobReconciler) ensureOneWorkload(ctx context.Context, job GenericJob, o
...
+					annotations := job.Object().GetAnnotations()
+					if annotations == nil {
+						annotations = make(map[string]string)
+					}
+					maps.Copy(annotations, customAnnotations)
+					job.Object().SetAnnotations(annotations)
+					return true, nil
+				}); err != nil {
+					return nil, fmt.Errorf("failed to get update custom annotations on job %s: %w", job.Object().GetName(), err)
+				}
```
mimowo, line 894: I would suggest to extract this to a helper `mergeAnnotations` which returns if the update is needed, so something like:

```golang
if newAnnotations, updated := mergeAnnotations(job, customAnnotations); updated {
    job.Object().SetAnnotations(newAnnotations)
    return true, nil
   }); err != nil {
     return nil, fmt.Errorf("failed to get update custom annotations on job %s: %w", job.Object().GetName(), err)
  }
} 
```
I know you already added the check inside `GetCustomAnnotations`, but it will be easier to use for also other users of the mechanism, and safer that someone would forget, wdyt?

pkg/controller/jobframework/reconciler.go
```diff
@@ -886,6 +880,26 @@ func (r *JobReconciler) ensureOneWorkload(ctx context.Context, job GenericJob, o
 			return nil, fmt.Errorf("failed to retrieve pod sets from job: %w", err)
 		}
 
+		if jobWithCustomAnnotations, ok := job.(JobWithCustomAnnotations); ok {
```
mimowo, line 883: Could you also add unit tests for reconciler.go in reconciler_test.go. I think this should be possible, based on the mocks.

pkg/controller/jobframework/reconciler.go
hiboyang: sounds good, updated

pkg/controller/jobframework/reconciler.go
hiboyang: yes, added unit test

pkg/controller/jobframework/reconciler.go
```diff
@@ -1648,3 +1656,17 @@ func WorkloadSliceEnabled(job GenericJob) bool {
 	}
 	return workloadslicing.Enabled(jobObject)
 }
+
+// mergeAnnotations merges customAnnotations into the job's existing annotations.
+// It returns a new merged annotation map and true if there are changes to apply.
+// The job's existing annotations are not modified.
+func mergeAnnotations(job GenericJob, customAnnotations map[string]string) (map[string]string, bool) {
```
mimowo, line 1663: This looks mostly OK, but I have one concern here: mergeAnnotations() currently returns true whenever customAnnotations is non-empty, even if all of those annotations already match the existing values. That could lead to unnecessary update requests on reconcile loops. I think it would be better to return true only when there is an effective change to apply. Something like:

```golang
func mergeAnnotations(job GenericJob, customAnnotations map[string]string) (map[string]string, bool) {
	if len(customAnnotations) == 0 {
		return nil, false
	}

	existing := job.Object().GetAnnotations()

	var merged map[string]string
	changed := false

	for k, v := range customAnnotations {
		if cur, ok := existing[k]; !ok || cur != v {
			if !changed {
				merged = maps.Clone(existing)
				if merged == nil {
					merged = make(map[string]string, len(customAnnotations))
				}
				changed = true
			}
			merged[k] = v
		}
	}
	return merged, changed
}
```
please also make sure the core is covered 100% with the unit tests in reconciler_test.go

hiboyang: updated

PR #9731, opened 2026-03-06
Propagate contextual logger to scheduling hash computation via UpdateSchedulingHash

sohankunkerkar:
Add `UpdateSchedulingHash(log)` method to workload.Info so production callers can compute the scheduling hash with a contextual logger after `NewInfo` construction. NewInfo signature stays unchanged.

#### What type of PR is this?
/kind cleanup

Fixes https://github.com/kubernetes-sigs/kueue/pull/9698#discussion_r2897610397

```release-note
NONE
```

### DIFF
--- a/pkg/cache/queue/manager.go
+++ b/pkg/cache/queue/manager.go
@@ -400,17 +400,19 @@ func (m *Manager) AddLocalQueue(ctx context.Context, q *kueue.LocalQueue) error 
 			continue
 		}
 
+		log := ctrl.LoggerFrom(ctx).WithValues("workload", klog.KObj(&w))
 		if features.Enabled(features.DynamicResourceAllocation) && workload.HasDRA(&w) {
 			if m.draReconcileChannel != nil {
 				m.draReconcileChannel <- event.TypedGenericEvent[*kueue.Workload]{Object: &w}
-				log := ctrl.LoggerFrom(ctx).WithValues("workload", klog.KObj(&w))
 				log.V(4).Info("Sent DRA workload to reconcile channel due to LocalQueue creation")
 			}
 			continue
 		}
 
 		workload.AdjustResources(ctx, m.client, &w)
-		qImpl.AddOrUpdate(workload.NewInfo(&w, m.workloadInfoOptions...))
+		wInfo := workload.NewInfo(&w, m.workloadInfoOptions...)
+		wInfo.UpdateSchedulingHash(log)
+		qImpl.AddOrUpdate(wInfo)
 	}
 
 	if cq != nil && cq.AddFromLocalQueue(qImpl, m.roleTracker) {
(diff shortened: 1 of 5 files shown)


pkg/cache/scheduler/clusterqueue.go
```diff
@@ -456,7 +456,7 @@ func (c *clusterQueue) addOrUpdateWorkload(log logr.Logger, w *kueue.Workload) {
 	if _, exist := c.Workloads[k]; exist {
 		c.deleteWorkload(log, k)
 	}
-	wi := workload.NewInfo(w, c.workloadInfoOptions...)
+	wi := workload.NewInfo(w, append(c.workloadInfoOptions, workload.WithLogger(log))...)
```
mimowo: I would rather consider a logger as a required paramer - first on the list rather than optional. Making it optional risks that someone will forget to pass.

sohankunkerkar: ack

pkg/cache/queue/cluster_queue_test.go
```diff
@@ -191,8 +192,8 @@ func Test_Pop(t *testing.T) {
 	ctx, _ := utiltesting.ContextWithLog(t)
 	now := time.Now()
 	cq := newClusterQueueImpl(ctx, nil, defaultOrdering, testingclock.NewFakeClock(now))
-	wl1 := workload.NewInfo(utiltestingapi.MakeWorkload("workload-1", defaultNamespace).Creation(now).Obj())
-	wl2 := workload.NewInfo(utiltestingapi.MakeWorkload("workload-2", defaultNamespace).Creation(now.Add(time.Second)).Obj())
+	wl1 := workload.NewInfo(logr.Discard(), utiltestingapi.MakeWorkload("workload-1", defaultNamespace).Creation(now).Obj())
```
mimowo: Rather than discard try to use the logger obtained from `utiltesting.ContextWithLog(t)` (applies here, but also to other spots.

pkg/cache/queue/manager.go
```diff
@@ -410,7 +410,9 @@ func (m *Manager) AddLocalQueue(ctx context.Context, q *kueue.LocalQueue) error
 		}
 
 		workload.AdjustResources(ctx, m.client, &w)
-		qImpl.AddOrUpdate(workload.NewInfo(&w, m.workloadInfoOptions...))
+		wInfo := workload.NewInfo(&w, m.workloadInfoOptions...)
+		wInfo.UpdateSchedulingHash(ctrl.LoggerFrom(ctx))
```
mimowo: This logger should be in the workload context already. We create such logger already a couple of lines above, so I would propose to move it up just above L403.

mimowo (commented):
LGTM, just one nit to make sure the logger has the workload context already

pkg/cache/queue/manager.go
sohankunkerkar: Ack

mimowo:
/cherrypick release-0.15
maybe now it will work

PR #9733, opened 2026-03-06
Fix NodeHotSwap for multi-layer TAS topology constraints

jzhaojieh:
#### What type of PR is this?
/kind bug 
/area tas

This fixes two issues with NodeHotSwap replacement in multi-layer topology:

1. Domain Selection (requiredReplacementDomain): When multiple slice
   constraints are configured (e.g., [{rack: 8}, {switch: 4}, {hostname: 2}]),
   the code was using only the outermost constraint to find the replacement
   domain. This could place replacements in the wrong sub-domain.

   Fix: Iterate from innermost to outermost constraint and use the first
   (innermost) broken constraint's domain to confine the search.

2. Slice Size (findReplacementAssignment): After clearing
   PodsetSliceRequiredTopologyConstraints, setting PodSetSliceSize alone
   was a dead write because getSliceSizeWithSinglePodAsDefault() requires
   PodSetSliceRequiredTopology to be set. This caused sliceSize=1, scattering
   replacement pods across hosts and violating hostname grouping constraints.

   Fix: Find the innermost satisfied constraint and set both
   PodSetSliceRequiredTopology and PodSetSliceSize to preserve leaf-level
   grouping.

Added test cases:
- 3-layer topology: validates innermost broken domain selection
- 2-layer topology: validates sliceSize prevents scattered placement

Fixes # 

Addresses the NodeHotSwap followup highlighted in #9506 

```release-note
NONE
```

### DIFF
--- a/pkg/cache/scheduler/tas_cache_test.go
+++ b/pkg/cache/scheduler/tas_cache_test.go
@@ -5825,13 +5825,15 @@ func TestFindTopologyAssignments(t *testing.T) {
 
 func TestFindTopologyAssignmentsMultiLayerReplacement(t *testing.T) {
 	const (
-		tasBlockLabel = "cloud.com/topology-block"
-		tasRackLabel  = "cloud.com/topology-rack"
+		tasBlockLabel  = "cloud.com/topology-block"
+		tasRackLabel   = "cloud.com/topology-rack"
+		tasSwitchLabel = "cloud.com/topology-switch"
 	)
-	levels := []string{tasBlockLabel, tasRackLabel, corev1.LabelHostname}
+	defaultLevels := []string{tasBlockLabel, tasRackLabel, corev1.LabelHostname}
 	podSetName := kueue.PodSetReference("main")
 
 	cases := map[string]struct {
+		levels          []string
 		nodes           []corev1.Node
 		pods            []corev1.Pod
 		existingTA      *kueue.TopologyAssignment
@@ -5848,6 +5850,7 @@ func TestFindTopologyAssignmentsMultiLayerReplacement(t *testing.T) {
 			//  r1        r2
 			//  /  \    /  |  \
 			// x1  x2  x3  x4 x5
+			//          ^(NotReady)
 			nodes: []corev1.Node{
 				*testingnode.MakeNode("b1-r1-x1").
 					Label(tasBlockLabel, "b1").Label(tasRackLabel, "r1").Label(corev1.LabelHostname, "x1").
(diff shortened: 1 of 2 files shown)


pkg/cache/scheduler/tas_cache_test.go
```diff
@@ -5958,6 +5960,214 @@ func TestFindTopologyAssignmentsMultiLayerReplacement(t *testing.T) {
...
+				*testingnode.MakeNode("b1-r2-s4-x7").
+					Label(tasBlockLabel, "b1").Label(tasRackLabel, "r2").Label(tasSwitchLabel, "s4").Label(corev1.LabelHostname, "x7").
+					StatusAllocatable(corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("4"), corev1.ResourcePods: resource.MustParse("10")}).
+					Ready().Obj(),
+				*testingnode.MakeNode("b1-r2-s4-x8").
+					Label(tasBlockLabel, "b1").Label(tasRackLabel, "r2").Label(tasSwitchLabel, "s4").Label(corev1.LabelHostname, "x8").
+					StatusAllocatable(corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("4"), corev1.ResourcePods: resource.MustParse("10")}).
+					Ready().Obj(),
+			},
+			pods: []corev1.Pod{
```
mimowo: What is the role of the Pods actually here? Maybe we could just drop them to simplify (same for the other test case). 

IIRC the intention of the "pods" param is to account for non-TAS Pods. So I think we could cover the code equally well without them.

pkg/cache/scheduler/tas_cache_test.go
```diff
@@ -5958,6 +5960,214 @@ func TestFindTopologyAssignmentsMultiLayerReplacement(t *testing.T) {
...
+			},
+		},
+		"2-layer: sliceSize=2 prevents scattered single-pod placement across hosts": {
+			// Topology: block → rack → hostname
+			//       b1
+			//   /        \
+			//  r1         r2
+			//  /  \    /  |  \
+			// x1  x2  x3 x4  x5
+			//          ^(NotReady)
```
mimowo, line 6050: I think this is not something specific to the PR, but could you refresh my memory what happens when multiple Nodes are NotReady? 

I know we don't support that, but maybe worth documenting the behavior with a test case. cc @PBundyra who likely has fresher memory here.

jzhaojieh: It looks like the hot swap features only supports replacing 1 node and Tte node failure controller will evict the workload if it requires replacement for more than 1 node https://github.com/kubernetes-sigs/kueue/blob/f4e1c4c15560f8b8ba26dcd138325253b019628d/pkg/controller/tas/node_failure_controller.go#L299-L302

I can update this with a test case for this as well

mimowo: > It looks like the hot swap features only supports replacing 1 node and Tte node failure controller will evict the workload if it requires replacement for more than 1 node

Yes, this matches my memory. Still, it would be good to cover that with a unit test. I don't recall if we have a unit test for the NHS for single-level slices, but we could add such a test case drive-by.

pkg/cache/scheduler/tas_cache_test.go
jzhaojieh: The pods were just used to consume resources on the nodes and influence the placement, but it looks like this is adjustable via the allocatable values directly on the nodes. Updated these tests to remove pods and just rely on the allocatable values

pkg/cache/scheduler/tas_cache_test.go
jzhaojieh: Taking another look it seems like there is a test already covering this at the controller layer which should cover this case, so the scheduler shouldnt be attempting to find replacement for more than one node. Let me know if theres any other gap that you think we should add additional unit tests for

https://github.com/kubernetes-sigs/kueue/blob/0d4b20498109b6cc3360f366db31e64c2a9e28c8/pkg/controller/tas/node_failure_controller_test.go#L246-L247

Issue #9734, opened 2026-03-07
Historical Usage Fair Sharing: half-life based score for cohort-level scheduling and preemption


mukund-wayve:
**What would you like to be added**:

An alternative fair sharing mode that uses half-life decayed historical resource usage for cohort-level admission ordering and preemption. This extends the mechanism from Admission Fair Sharing (KEP-4136), currently scoped to ordering LocalQueues within a single ClusterQueue to both ordering and preempting ClusterQueues within a cohort. It coexists alongside the current Dominant Resource Share (DRS) from KEP-1714 as a configurable option.

Where DRS answers "who is borrowing the most right now?", historical usage answers "who has consumed the most over time?" — giving the system memory of past consumption. The mode is configured globally in the Kueue FairSharing config, with observability through CQ status fields and Prometheus metrics.

**Why is this needed**:

DRS is point-in-time and has no memory of past consumption. This creates problems for organizations where teams budget compute as a percentage of the cluster over a longer period (e.g., a week) but have bursty workloads:

1. A team idle for six days that bursts on day seven is immediately penalized by DRS, even though their cumulative usage is well within their fair share.
2. A CQ that just finished consuming most of the cohort for a week gets DRS = 0 the moment its workloads complete — identical to a CQ that has been idle for months. Both get full scheduling and preemption power immediately.
3. Two CQs with identical current borrowing are treated equally by DRS, even if one has been borrowing for a week and the other started five minutes ago.

KEP-1714 explicitly listed historical data as a non-goal but noted the design should be "expandable to support history-based Fair Sharing without major redesign." KEP-4136 introduced half-life decay for intra-CQ ordering but acknowledged cohort-level scope as future work. This feature closes that gap and I believe will be a significant improvement in the fair sharing offering of Kueue.

**Completion requirements**:

This enhancement requires the following artifacts:

- [x] Design doc
- [x] API change
- [x] Docs update

The artifacts should be linked in subsequent comments.

PR #8646, opened 2026-01-16
KEP-8522: Add time-based same-priority preemption proposal

sohankunkerkar:
#### What type of PR is this?
/kind documentation
/kind feature

#### Which issue(s) this PR fixes:

Related to #8522 

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #9737, opened 2026-03-07
KEP-9734: Historical Borrowing based DRS Fair Sharing

mukund-wayve:
#### What type of PR is this?

/kind kep

Creates a KEP for #9734 

Part of  #9734 

```release-note
NONE
```

### DIFF
--- a/keps/9734-historical-usage-fair-sharing/README.md
+++ b/keps/9734-historical-usage-fair-sharing/README.md
@@ -42,8 +42,8 @@ This KEP introduces an alternative strategy for Dominant Resource Share
 from KEP-4136 (Admission Fair Sharing) to cohort-level scheduling and
 preemption, while keeping the same DRS aggregation from KEP-1714.
 
-Where DRS with borrowing input answers "who is borrowing the most right now?",
-DRS with historical borrowing input answers "who has borrowed the most over
+Where DRS with borrowing strategy answers "who is borrowing the most right now?",
+DRS with historical borrowing strategy answers "who has borrowed the most over
 time?" — giving the system memory of past borrowing so that recent heavy
 consumption of shared capacity is appropriately penalised.
 
@@ -101,7 +101,7 @@ ClusterQueues within a cohort for either admission ordering or preemption.
 We introduce a new DRS strategy — `HistoricalBorrowing` — that replaces
 instantaneous borrowing with half-life decayed borrowing as input to the
 existing DRS aggregation (max ratio of borrowing to lendable capacity). The
-aggregation and all downstream behavior remain unchanged.
+aggregation and all downstream behavior — including fair sharing weights — remain unchanged.
 
 ### Historical borrowing computation
 
--- a/keps/9734-historical-usage-fair-sharing/README.md
+++ b/keps/9734-historical-usage-fair-sharing/README.md
@@ -20,9 +20,9 @@
     - [Kueue configuration](#kueue-configuration)
     - [ClusterQueue status](#clusterqueue-status)
   - [DominantResourceShare with strategy delegation](#dominantresourceshare-with-strategy-delegation)
+  - [CQ-level borrowing tracking](#cq-level-borrowing-tracking)
   - [Preemption behavior](#preemption-behavior)
   - [Preemption frequency bounds](#preemption-frequency-bounds)
-  - [CQ-level borrowing tracking](#cq-level-borrowing-tracking)
   - [Test Plan](#test-plan)
     - [Unit Tests](#unit-tests)
     - [Integration tests](#integration-tests)
@@ -291,7 +291,7 @@ Validation rules:
 
 * `minAdmitDuration` is **required** when `Strategy = HistoricalBorrowing`
   (see [Preemption frequency bounds](#preemption-frequency-bounds)).
-  `minAdmitDuration >= 10m` is enforced as a floor.
+  `minAdmitDuration >= 1m` is enforced as a floor.
 
 * When `Strategy = Borrowing` (default), `minAdmitDuration` is optional.
   Instantaneous DRS is loop-free and does not require a minimum runtime guard,
@@ -402,7 +402,7 @@ In particular:
 * `minAdmitDuration` filtering: workloads admitted less than `minAdmitDuration`
   ago are skipped as fair-sharing preemption candidates.
 * Validation: `minAdmitDuration` required when `Strategy = HistoricalBorrowing`,
-  floor of 10 minutes enforced.
+  floor of 1 minute enforced.
 * Borrowing filter behavior with historical borrowing strategy.
 
 #### Integration tests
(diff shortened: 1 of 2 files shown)


keps/9734-historical-usage-fair-sharing/README.md
```diff
@@ -0,0 +1,423 @@
...
+
+### Historical borrowing computation
+
+For each ClusterQueue, Kueue tracks decayed **borrowing** — the amount of
+resource consumption beyond the CQ's own quota — using the same exponential
+moving average as KEP-4136:
+
+```
+borrowing(r) = max(0, usage(r) - quota(r))
+borrowing_sum(r) = (1 - α) × previous_borrowing_sum(r) + α × borrowing(r)
```
pajakd, line 116: For α = 1 this is exactly the current behavior?

keps/9734-historical-usage-fair-sharing/README.md
```diff
@@ -0,0 +1,423 @@
...
+
+* Allow cluster admins to configure fair sharing so that historical borrowing
+  — not just instantaneous borrowing — determines scheduling and preemption
+  priority across ClusterQueues within a cohort.
+
+* Apply the same historical borrowing score to both admission ordering and
+  preemption target selection, so that the two mechanisms are consistent.
+
+* Preserve the preemption loop-free property that the current DRS-based fair
+  sharing guarantees.
```
pajakd: For the current DRS-based fair sharing we have a proof that it is loop-free https://kueue.sigs.k8s.io/docs/concepts/fair_sharing/#proof-that-two-workloads-wont-preempt-each-other 

Here it'll be a little more tricky as the DRS will change over time without any new admission. So the proof will not hold (I think). 

Lets say we have a setup like this:

-  cohort (1 quota):
- CQ1 (no quota), workload `wl1` requiring 1 quota
- CQ2 (no quota), workload `wl2` requiring 1 quota

Lets say these are workloads that take a long time to finish. 

I think in your proposal `wl1` and `wl2` will keep preempting each other, right? With `wl1` admitted the new DRS will keep increasing until eventually it will be preempted by `wl2` and so on.

Its not necessarily a bad thing if the preemptions will not be too frequent -- it might be even desirable, as this is more fair to do such "multiplexing" instead of allowing `wl1` to run until completion. But this is a different property than for the current algorithm.

mukund-wayve: Great spot!
I think it relates to #8646 . I'm leaning towards accepting the loop, but just that we slow down the frequency by setting a sufficiently high `minAdmitDuration`

mukund-wayve: Addressed in [Preemption frequency bounds](https://github.com/kubernetes-sigs/kueue/pull/9737/changes#diff-94b336b2c98abe776c79d5159b9e533dc8abbcab64dfc0eda7028edbf4614b1aR362) section. Let me know what you think

mukund-wayve: Raised a PR to add [minAdmitDuration](https://github.com/kubernetes-sigs/kueue/pull/9770) as a generic FairSharingConfig

keps/9734-historical-usage-fair-sharing/README.md
```diff
@@ -0,0 +1,478 @@
...
+weightedScore = score / fairWeight
+```
+
+The aggregation method is unchanged — only the **input** to DRS changes from
+instantaneous borrowing to half-life decayed historical borrowing. This is
+deliberate: users want "DRS but with memory," and reusing max-ratio
+aggregation means zero-config adoption (no `resourceWeights` to tune) with
+built-in scarcity normalization via `lendable`.
+
+### Entry penalty
```
mwielgus, line 140: I would consider increasing entry penalty scope to more than I sampling interval. 

One idea would be to check what would historical DRS look like in, say, 15 minutes from now if:
* the preemption mas made to allocate pending workload
* the preemption was not made and the workload was not admitted. 

Something like this should stabilize the algorithm and reduce the number of oscillations caused by back-and-forth preemptions due to DRS flapping. 

In control theory there is a thing called PID controller (see https://en.wikipedia.org/wiki/Proportional%E2%80%93integral%E2%80%93derivative_controller). You may think about the old preemption based FS as a PID with just current value. You are introducing integral fraction. In this comment I'm proposing adding gradient/differential part as well, for completeness.

keps/9734-historical-usage-fair-sharing/README.md
```diff
@@ -0,0 +1,478 @@
...
+  floor of 1 minute enforced.
+* Borrowing filter behavior with historical borrowing strategy.
+
+#### Integration tests
+
+They will mainly focus on cohort-level scheduling with historical borrowing
+and interactions with existing fair sharing mechanisms.
+
+In particular:
+* Cohort-level admission ordering and preemption targeting based on historical
```
mukund-wayve, line 501: When multiple teams are involved, does preempting happen evenly across workloads from all other teams that are over their fair share?

keps/9734-historical-usage-fair-sharing/README.md
mukund-wayve: Thanks for the suggestion. I've added this to the KEP. Let me know if I've captured your suggestion correctly

keps/9734-historical-usage-fair-sharing/README.md
mukund-wayve: This is already handled by existing DRS tests. As long as we can confirm the DRS scoring is working well, we are good

amy:
@mukund-wayve @mimowo @mwielgus Wanted to check on the status of this Kep. I'd love for it to get merged. Happy to help contribute PRs to make this happen! 

Really need admission fairsharing at the cohort level for use cases where preemption is less stringent than preemption fairsharing.

keps/9734-historical-usage-fair-sharing/README.md
```diff
@@ -0,0 +1,582 @@
...
+## Proposal
+
+We introduce a new DRS strategy — `HistoricalBorrowing` — that replaces
+instantaneous borrowing with half-life decayed borrowing as input to the
+existing DRS aggregation (max ratio of borrowing to lendable capacity). The
+aggregation and all downstream behavior — including fair sharing weights — remain unchanged.
+
+### Historical borrowing computation
+
+For each ClusterQueue, Kueue tracks decayed **borrowing** — the amount of
```
gabesaba, line 110: How about for Cohorts? I would like this to be compatible with Hierarchical Cohorts, similar to how the current Fair Sharing works.

amy: +1

mukund-wayve: That makes sense, I'll change the KEP to support cohorts as well

PR #9738, opened 2026-03-07
Decouple borrowing state from DRS ratio value

mukund-wayve:
#### What type of PR is this?

/kind cleanup

The preemption pruning logic in ordering.go uses `drs.IsZero()` as a proxy for "not currently borrowing." This conflates the numerical DRS ratio with the real-time borrowing state. With historical-borrowing-based DRS (#9737), the ratio may be non-zero even when a node is not actively borrowing, breaking this assumption.

This PR adds an explicit `IsBorrowing()` method to DRS, set from the existing borrowing computation in `dominantResourceShare()`. The two pruning sites in ordering.go now use `!drs.IsBorrowing()` instead of `drs.IsZero()`.

No behavioral change with the current DRS computation.

Part-of #9734 

```release-note
NONE
```


pkg/cache/scheduler/fair_sharing_test.go
```diff
@@ -783,3 +831,47 @@ func TestDominantResourceShare(t *testing.T) {
 		})
 	}
 }
+
+func TestDRS_IsBorrowing(t *testing.T) {
```
pajakd: Right now the `IsBorrowing` and `IsZero` functions are so simple that they do not need tests imo (unless I'm missing something).

pkg/cache/scheduler/fair_sharing.go
```diff
@@ -44,20 +44,27 @@ type DRS struct {
...
 
 // NegativeDRS is used as a starting point for comparisons.
 func NegativeDRS() DRS {
 	return DRS{unweightedRatio: -1, dominantResource: "", fairWeight: defaultWeight}
 }
 
-// IsZero returns whether the DRS is 0. In other words,
-// the node for which this function was called is not
-// borrowing any resources.
+// IsZero returns whether the DRS ratio is 0.
```
pajakd: It would be nice to actually have some comments that facilitate the understanding of fair sharing -- especially that its going to be enriched with new features.

When https://github.com/kubernetes-sigs/kueue/pull/9737 meges (and the implementation) we'll have:
- if DRSStrategy = "Borrowing" then DRS = 0 iff the node is not borrowing any resources
- if DRSStrategy = "HistoricalBorrowing" then DRS = 0 iff the node has never borrowed any resources

Right? Lets keep in mind to add a comment with the correct interpretation when the PRs with HistoricalBorrowing merge.

mukund-wayve: Sounds good. Will make this change. Thanks for the review. 👍  Since you now have context of historical borrowing, can I add you as a reviewer to #9737 ?

pajakd: Sure

PR #9739, opened 2026-03-08
feat: enhance error message for multi-layer topology constraints

Huang-Wei:
#### What type of PR is this?

/kind feature
/area tas

Add detailed noFit messages for multi-layer TAS topology that report per-level slice fit counts (e.g., "1/2 slice(s) fit on level block; 3/4 slice(s) fit on rack; 6/8 slice(s) fit on hostname").

Follow-up of #9506

xref #9046

```release-note
NONE
```

### DIFF
--- a/pkg/cache/scheduler/tas_cache_test.go
+++ b/pkg/cache/scheduler/tas_cache_test.go
@@ -5752,6 +5752,62 @@ func TestFindTopologyAssignments(t *testing.T) {
 				wantReason: `topology "default" doesn't allow to fit; 1/2 slice(s) fit on level cloud.com/topology-rack; 3/4 slice(s) fit on level kubernetes.io/hostname`,
 			}},
 		},
+		"multi-layer topology: enough hostname slices but not enough rack slices": {
+			// block → rack → hostname
+			//            b1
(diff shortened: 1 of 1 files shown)


pkg/cache/scheduler/tas_cache_test.go
```diff
@@ -5673,7 +5683,7 @@ func TestFindTopologyAssignments(t *testing.T) {
 					corev1.ResourceCPU: 1000,
 				},
 				count:      6,
-				wantReason: `topology "default" doesn't allow to fit any of 1 slice(s)`,
+				wantReason: `topology "default" doesn't allow to fit; 0/1 slice(s) fit on level cloud.com/topology-rack; 2/3 slice(s) fit on level kubernetes.io/hostname. Total nodes: 3; excluded: resource "cpu": 1`,
```
mimowo, line 5686: Nice, I would like to also have a test case where there is enough capacity at the lowest level, but not enough at one of the higher levels.

Huang-Wei: Good idea. Added a test case.

PR #8734, opened 2026-01-22
KEP-2941: Add extended resources design for DRA integration

sohankunkerkar:
#### What type of PR is this?
/kind documentation

#### Which issue(s) this PR fixes:

Part of  https://github.com/kubernetes-sigs/kueue/issues/9868

#### Special notes for your reviewer:

Please review https://github.com/kubernetes-sigs/kueue/pull/8597 for implementation.

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #9742, opened 2026-03-08
Support CEL expressions in DRA ResourceClaimTemplates

### LINKED ISSUE #7298: Support CEL expressions in the DRA support
Upstream k8s allows [downstream components to evaluate CEL expressions](https://kubernetes.slack.com/archives/C0409NGC1TK/p1760556948412779?thread_ts=1760556528.438029&cid=C0409NGC1TK), so this should enable `kueue` to allow support for CEL expressions in the `ResourceClaimTemplates` 

**Why is this needed**:
CEL expression allow efficient filtering 

**Completion requirements**:
Kueue should be able to process `ResourceClaimTemplate` with the CEL expression 

- [ ] Design doc
- [ ] API change
- [ ] Docs update

kannon92:
#### What type of PR is this?

/kind feature

Removes the restriction that rejected CEL selectors in DRA `ResourceClaimTemplate` device requests. CEL selectors are filters that narrow which devices satisfy a request but don't change how many devices are requested. Kueue only needs the device count and class name for quota management, so CEL selectors can be safely accepted without evaluation. The Kubernetes scheduler handles actual device matching at scheduling time.

This enables use cases like:
- Partitionable devices (MIG GPUs) via [KEP-4815](https://github.com/kubernetes/enhancements/blob/master/keps/sig-scheduling/4815-dra-partitionable-devices/README.md)
- Network device filtering via [DRANET](https://dranet.dev/docs/user/nvidia-dranet/)
- Any DRA driver that publishes device attributes for CEL-based selection

Fixes #7298

This PR was developed with the assistance of Claude Code (AI).

The upstream issue [kubernetes/kubernetes#137490](https://github.com/kubernetes/kubernetes/issues/137490) tracks improving error feedback when CEL selectors don't match any device at scheduling time — that is a separate scheduler-side concern.

```release-note
Support CEL expressions in DRA ResourceClaimTemplates. Kueue no longer rejects workloads that use CEL device selectors for filtering devices in resource claims.
```


sohankunkerkar (commented):
I spent some time looking at this from both sides i.e. the Kueue DRA pipeline and the upstream K8s allocator and I think there's more to this than removing the CEL rejection.
                                                                                                                                                                                                                                                                                                                                                                             
The problem is that `CEL expressions in DRA` covers two very different things. For simple filtering (DRANET, selecting by driver name), CEL just narrows which devices match and the cost is still "1 device" and removing the rejection works fine. But for partitionable devices (KEP-4815), which was the primary motivation for this issue, two workloads can request the same DeviceClass with count=1, but one picks a small MIG partition consuming ~5Gi of 80Gi GPU memory while the other picks a full partition consuming the entire 80Gi. The upstream scheduler handles this through SharedCounters/ConsumesCounters on ResourceSlices but if Kueue just counts devices, both get the same quota cost and the quota becomes meaningless.

We know Kueue shouldn't evaluate CEL. But we do need some way to determine the actual cost in terms of counter consumption rather than device count; whether through admin config that assigns cost per DeviceClass, user-declared consumption, or ResourceSlice observation.

~Removing the CEL rejection is a reasonable first step for simple filtering use cases~, but I don't think we should close [this](https://github.com/kubernetes-sigs/kueue/issues/7298) issue until we've addressed the counter-based quota problem for partitionable devices.

**Edit:** Beyond the partitionable devices problem, there's an even more immediate concern with simply removing the rejection. CEL evaluation happens in the scheduler's Filter phase against real ResourceSlice data well beyond Kueue's reach. By the time the scheduler runs, Kueue has already admitted and charged quota. If a user submits a workload with a CEL selector that no device actually matches (wrong attribute name, impossible filter, typo), Kueue charges quota, the scheduler can't place the pod, and that quota is permanently consumed. This blocks other legitimate workloads from getting through. @harche reproduced this on a Kind cluster with the DRA example driver. Without `WaitForPodsReady` configured (which is optional), there's no self-correction at all.

kannon92:
Yea this is sorta a problem with Kueue generally.

If a user requests a bogus gpu and sets ClusterQueue for it (or say they want a pod requiesting 256 gpus). I think Kueue will check against quota.

But it won't check the scheduler if it can actually be placed.

Kueue's TAS does incorporate scheduling capacity and becomes its own scheduler.
So this problem becomes critical to solve for DRA + TAS but we haven't yet designed or worked on that.

For kueue as a quota manager I guess I'm not sure if we need CEL expressions to be evaulated at quota time. We can certainly do it and its required for TAS. But I don't know of anyone working on Kueue + DRA + TAS and incorportating DRA scheduling into TAS.

pkg/dra/claims.go
```diff
@@ -182,3 +207,143 @@ func GetResourceRequestsForResourceClaimTemplates(
...
+// If fewer devices match than requested, it returns field errors indicating the
+// workload is unsatisfiable, preventing quota from being consumed by workloads
+// whose pods can never be scheduled.
+func validateCELSelectorsAgainstDevices(ctx context.Context, cl client.Client, claimSpec *resourcev1.ResourceClaimSpec, basePath *field.Path) field.ErrorList {
+	celReqs := extractCELRequests(claimSpec)
+	if len(celReqs) == 0 {
+		return nil
+	}
+
+	var sliceList resourcev1.ResourceSliceList
```
kannon92, line 410: I think we may have to add rbac here for ResourceSliceList.

pkg/dra/claims.go
```diff
@@ -182,3 +207,143 @@ func GetResourceRequestsForResourceClaimTemplates(
...
+	for _, cr := range celReqs {
+		// Compile all CEL selectors for this request.
+		var compiled []dracel.CompilationResult
+		for _, sel := range cr.selectors {
+			if sel.CEL == nil {
+				continue
+			}
+			result := compiler.CompileCELExpression(sel.CEL.Expression, dracel.Options{})
+			if result.Error != nil {
+				// Already validated by validateCELSelectors, skip.
```
rphillips: Capturing the error in AllErrs would be good here to catch invalid selectors.

pkg/dra/claims.go
```diff
@@ -182,3 +207,143 @@ func GetResourceRequestsForResourceClaimTemplates(
...
+		result := compiler.CompileCELExpression(sel.CEL.Expression, dracel.Options{})
+		if result.Error != nil {
+			allErrs = append(allErrs, field.Invalid(
+				fldPath.Index(i).Child("cel", "expression"),
+				sel.CEL.Expression,
+				fmt.Sprintf("CEL compilation failed: %s", result.Error.Detail),
+			))
+		}
+	}
+	return allErrs
```
rphillips: errors.Join() might be useful here so the function can return just `error`.

pkg/dra/claims.go
kannon92: done.

pkg/dra/claims.go
kannon92: done.

kannon92:
I think this is ready for review.

pkg/dra/claims.go
```diff
@@ -182,3 +208,147 @@ func GetResourceRequestsForResourceClaimTemplates(
...
+	}
+
+	var sliceList resourcev1.ResourceSliceList
+	if err := cl.List(ctx, &sliceList); err != nil {
+		return field.ErrorList{field.InternalError(basePath, fmt.Errorf("failed to list ResourceSlices: %w", err))}
+	}
+
+	// Build CEL device representations from ResourceSlices.
+	type clusterDevice struct {
+		driver string
```
harche: nit: Is the `driver` field used? I could only see it being set but not read anywhere. `device.Driver` seems to already carry the same value.

test/integration/singlecluster/controller/dra/dra_test.go
```diff
@@ -526,7 +527,31 @@ var _ = ginkgo.Describe("DRA Integration", ginkgo.Ordered, ginkgo.ContinueOnFail
 			}, util.LongTimeout, util.Interval).Should(gomega.Succeed())
 		})
 
-		ginkgo.It("Should reject workload with CEL selectors", func() {
+		ginkgo.It("Should admit workload with CEL selectors", func() {
```
harche, line 545: Minor: The old integration test covered the CEL rejection path through the controller. Now that it's replaced with the happy path, the rejection flow is only covered at unit and E2E level. Might be worth adding an integration test that creates a ResourceSlice with non-matching devices and verifies the inadmissible condition, but not a blocker since the existing coverage across unit + E2E is solid.

harche (commented):
Nice work addressing the quota leak issue with CEL selectors. The approach of fetching ResourceSlices and evaluating CEL via `DeviceMatches` is solid. Two minor comments below.

pkg/dra/claims.go
```diff
@@ -162,6 +174,20 @@ func GetResourceRequestsForResourceClaimTemplates(
 				return nil, allErrs
 			}
 
+			// Validate CEL selectors against actual devices in the cluster.
+			celBasePath := field.NewPath("spec", "podSets").Index(i).Child("template", "spec", "resourceClaims").Index(j)
+			if celErrs := validateCELSelectorsAgainstDevices(ctx, cl, spec, celBasePath); len(celErrs) > 0 {
```
sohankunkerkar, line 195: I liked this idea but per @mimowo's suggestion, we should evaluate CEL to avoid admitting workloads that can never be placed. My concern is the liveness of this check. This validation runs in the workload controller's Reconcile, before the scheduler ever sees the workload. When the scheduler marks a workload inadmissible (e.g., insufficient quota), `queueInadmissibleWorkloads` periodically re-evaluates it. But workloads rejected here sit with an error condition on the workload status indefinitely and nothing re-triggers reconciliation when ResourceSlices change.

So if no devices match right now but a new GPU node joins or a DRA driver restarts, the workload stays stuck. Users would need to manually touch the workload to force re-evaluation.

  Two possible paths:
  1. A ResourceSlice informer that re-enqueues affected workloads when device inventory changes
  2. Move this check to the scheduler's admission cycle where the periodic re-evaluation already exists (similar to how TAS does topology checks during snapshot/nomination,
  not in the workload controller)

pkg/dra/claims.go
kannon92: done.

test/integration/singlecluster/controller/dra/dra_test.go
kannon92: added.

pkg/dra/claims.go
```diff
@@ -182,3 +208,140 @@ func GetResourceRequestsForResourceClaimTemplates(
...
+			}
+			compiled = append(compiled, result)
+		}
+		if len(compiled) == 0 {
+			continue
+		}
+
+		// Evaluate against all cluster devices.
+		var matchCount int64
+		for _, dev := range clusterDevices {
```
sohankunkerkar: The evaluation loop iterates all devices from all ResourceSlices regardless of DeviceClassName. If the request is for `nvidia.com/gpu` devices, the CEL still runs against NICs, FPGAs, and everything else in the cluster. I think the performance concern is real as evaluating CEL against a bunch of devices when only a few belong to the requested class wastes cycles.
                                                                                                                                                                          

The request already carries `req.Exactly.DeviceClassName` which maps to a specific driver via the `DeviceClass` object. We could resolve that (one GET, reusable across requests) and pre-filter the device list by driver before running CEL. Cuts evaluation scope and makes the count accurate.

pkg/dra/claims.go
```diff
@@ -182,3 +208,140 @@ func GetResourceRequestsForResourceClaimTemplates(
...
+			allMatch := true
+			for _, comp := range compiled {
+				matches, _, err := comp.DeviceMatches(ctx, dev)
```
sohankunkerkar: Errors from DeviceMatches are silently treated as `no match`. This makes sense for some errors (device doesn't have the attribute the CEL references) but not for others (context cancelled, CEL runtime type mismatch).

If a CEL expression errors on every device, the user sees `CEL selectors match only 0 device(s)` which is misleading bc the real issue is evaluation failure. Could we log the first error at V(3), or surface it in the field error message?

pkg/dra/claims.go
```diff
@@ -182,3 +208,140 @@ func GetResourceRequestsForResourceClaimTemplates(
...
+			dev := &slice.Spec.Devices[j]
+			clusterDevices = append(clusterDevices, dracel.Device{
+				Driver:     slice.Spec.Driver,
+				Attributes: dev.Attributes,
+				Capacity:   dev.Capacity,
+			})
+		}
+	}
+
+	compiler := dracel.GetCompiler(dracel.Features{})
```
sohankunkerkar: nit: the CEL expressions are compiled twice, once in `validateCELSelectors`  and again here for evaluation. Since the first check already passes before we reach this point, the compilation error handling at https://github.com/kubernetes-sigs/kueue/pull/9742/changes#diff-1817bbb9cab28ee6c2fb856446a0164731ee7fb63b3620c069d31c439750e2d1R306-R312 is dead code.

go.mod
```diff
@@ -34,6 +34,7 @@ require (
 	k8s.io/client-go v0.35.2
 	k8s.io/component-base v0.35.2
 	k8s.io/component-helpers v0.35.2
+	k8s.io/dynamic-resource-allocation v0.35.2
```
sohankunkerkar: Quick check: in PR #8734, @mimowo previously flagged concerns about importing from `k8s.io/dynamic-resource-allocation`. Just want to make sure we're aligned on taking this dependency.

test/integration/singlecluster/controller/dra/dra_test.go
sohankunkerkar: +1

pkg/dra/claims.go
kannon92: I'm not sure if this has to be solved here.

I guess your point is valid with DeviceClasses also we should watch all deviceClasses and ResourceSlices for any changes. Both of these objects are usually added by an admin when setting up a cluster.

2) Seems like we have to do this when we do DRA + TAS so I wonder if we can tackle this at that point.

go.mod
mimowo: I'm ok with the library, initially I thought this is part of the core k8s repo, but I see the library is extracted to dedicated repo

pkg/dra/claims.go
kannon92: updated.

pkg/dra/claims.go
```diff
@@ -182,3 +209,195 @@ func GetResourceRequestsForResourceClaimTemplates(
...
+					allMatch = false
+					break
+				}
+			}
+			if allMatch {
+				matchCount++
+			}
+		}
+
+		if matchCount < cr.count {
```
alaypatel07, line 454: This if check can be inverted and put inside the loop, effectively, if the matchCount >= quota requested, exit the loop early, improving the performance from what is suggested here: https://github.com/kubernetes-sigs/kueue/pull/9742/changes#r2921055482

kannon92:
@mimowo @tenzen-y @alaypatel07 

PR changed a bit. Could you all review? 

@PannagaRao work on partionable devices is blocked by this due to the need to support CEL.

pkg/dra/claims.go
kannon92: I added something like that on line 415. Going to resolve.

pkg/dra/claims.go
kannon92:
```
			allErrs = append(allErrs, field.Invalid(
				basePath.Child("devices", "requests").Index(cr.index).Child("exactly", "selectors"),
				nil,
				fmt.Sprintf("insufficient matching devices for CEL selector in DeviceClass %s: %d device(s) match in the cluster but %d requested",
					cr.deviceClassName, matchCount, cr.count),
			))
```

I added this below which I think is sufficient.

kannon92:
/hold cancel

KEP was merged.

/cc @tenzen-y @mimowo @sohankunkerkar @alaypatel07

mimowo:
Thank you @kannon92 for the update on the PR. I would love to have the first pass done by @sohankunkerkar or @alaypatel07, do you folks have cycles?

pkg/dra/claims.go
sohankunkerkar: The KEP says `queueInadmissibleWorkloads` re-evaluates these workloads, but they never enter the ClusterQueue as errors from here flow to https://github.com/kubernetes-sigs/kueue/pull/9742/changes#diff-60dd240c20adbd6a189d018d1c216c2d296730f446c341d8bf449fa6657964ffR320, so re-evaluation is controller-runtime's error retry. Beta criteria list event-driven DeviceClass tracking and ResourceSlice indexing for performance, but not event-driven ResourceSlice tracking for re-evaluating stuck CEL workloads. Consider adding that as a Beta criterion and fixing the `queueInadmissibleWorkloads` reference.

pkg/dra/claims.go
sohankunkerkar: The error message reports the count, but the concern was about distinguishing `evaluated false on all devices` from `evaluation failed on all devices.` Both produce `matchCount=0` but have different root causes. Could we log the first DeviceMatches error at V(3) so users can tell the difference in controller logs?

pkg/dra/claims.go
kannon92: https://github.com/kubernetes-sigs/kueue/pull/10961

pkg/dra/claims.go
sohankunkerkar: @kannon92 any inputs here?

kannon92: sure. sorry I missed this. I'll get to this today.

kannon92: Addressed in 1897f743e — the first DeviceMatches evaluation error is now logged at V(3) so users can distinguish "no devices matched" from "CEL evaluation failed."

pkg/dra/claims.go
kannon92: Addressed in 1897f743e — removed the redundant compilation error handling since validateCELSelectors already catches these errors before validateCELSelectorsAgainstDevices is called.

kannon92:
@sohankunkerkar @alaypatel07 
I believe I have addressed all your concerns.

Please take a look.

pkg/dra/claims.go
```diff
@@ -192,3 +224,213 @@ func GetResourceRequestsForResourceClaimTemplates(
...
+// validateCELSelectorsAgainstDevices lists all ResourceSlices in the cluster and
+// evaluates the CEL selectors from each request against the actual devices.
+// For each request, it resolves the DeviceClassName to its DeviceClass and uses
+// the class selectors to pre-filter devices before evaluating the request's own
+// CEL selectors. This avoids running CEL against devices that belong to
+// unrelated drivers or classes.
+// If fewer devices match than requested, it returns field errors indicating the
+// workload is unsatisfiable, preventing quota from being consumed by workloads
+// whose pods can never be scheduled.
+func validateCELSelectorsAgainstDevices(ctx context.Context, cl client.Client, claimSpec *resourcev1.ResourceClaimSpec, basePath *field.Path) field.ErrorList {
```
mimowo, line 404: Please break this spaghetti into smaller functional pieces :) This seems to combine a couple of stages that would be better to reflect in helper functions, like getting the clusterDevices, then classSelectors, then compiling Cels, the actual validation.

pkg/dra/claims.go
```diff
@@ -192,3 +224,213 @@ func GetResourceRequestsForResourceClaimTemplates(
...
+		if req.Exactly == nil || len(req.Exactly.Selectors) == 0 {
+			continue
+		}
+		hasCEL := false
+		for _, sel := range req.Exactly.Selectors {
+			if sel.CEL != nil {
+				hasCEL = true
+				break
+			}
+		}
```
mimowo: Please use built-in golang helpers for stuff like that, here slices.ContainsFunc seems to be enough

test/e2e/dra/dra_test.go
```diff
@@ -282,6 +282,117 @@ var _ = ginkgo.Describe("DRA", func() {
...
+				Name:      workloadjob.GetWorkloadNameForJob(legitimateJob.Name, legitimateJob.UID),
+				Namespace: ns.Name,
+			}
+
+			ginkgo.By("Verifying legitimate workload is admitted - quota was not leaked by unmatchable job")
+			legitimateWl := &kueue.Workload{}
+			gomega.Eventually(func(g gomega.Gomega) {
+				g.Expect(k8sClient.Get(ctx, legitimateWlKey, legitimateWl)).To(gomega.Succeed())
+				g.Expect(workload.HasQuotaReservation(legitimateWl)).To(gomega.BeTrue())
+			}, util.Timeout, util.Interval).Should(gomega.Succeed())
```
mimowo: We already have helpers for checking if a workload is admitted, eg ExpectWorkloadsToBeAdmitted

pkg/dra/claims.go
kannon92: This is good feedback.

kannon92: Did some refactors. I think this satifies your ask.

pkg/dra/claims.go
kannon92: done.

test/e2e/dra/dra_test.go
kannon92: updated.

kannon92:
@sohankunkerkar @mimowo 

anything outstanding here?

pkg/dra/claims.go
```diff
@@ -192,3 +225,251 @@ func GetResourceRequestsForResourceClaimTemplates(
...
+		result := celCache.GetOrCompile(sel.CEL.Expression)
+		if result.Error != nil {
+			errs = append(errs, field.Invalid(
+				fldPath.Index(i).Child("cel", "expression"),
+				sel.CEL.Expression,
+				fmt.Sprintf("CEL compilation failed: %s", result.Error.Detail),
+			))
+		}
+	}
+	return errors.Join(errs...)
```
mimowo: This code is duplicated with compileCELSelectors. I think we could just call `compileCELSelectors`, and return the errors, no?

pkg/dra/claims.go
```diff
@@ -192,3 +225,251 @@ func GetResourceRequestsForResourceClaimTemplates(
...
+		cache[className] = dc
+	}
+
+	compiled, errs := compileCELSelectors(
+		dc.Spec.Selectors,
+		basePath.Child("devices", "requests").Index(reqIndex),
+		fmt.Sprintf("DeviceClass %s has invalid CEL selector", className),
+	)
+	if len(errs) > 0 {
+		return nil, errs[0] // Return first error
```
mimowo: I think the comment "// Return first error" is not needed - this is self explanatory. The useful comment would be to explain **why** we return the first item instead of aggregating the errors into one.

pkg/dra/claims.go
```diff
@@ -192,3 +225,251 @@ func GetResourceRequestsForResourceClaimTemplates(
...
+
+	for devIdx, dev := range devices {
+		if consumed[devIdx] {
+			continue
+		}
+
+		// First check class selectors to skip devices that don't belong to this class.
+		classMatch, err := evaluateSelectorsOnDevice(ctx, classSelectors, dev)
+		if err != nil || !classMatch {
+			continue
```
mimowo, line 375: Why don't just return error here early?

pkg/dra/claims.go
```diff
@@ -192,3 +225,251 @@ func GetResourceRequestsForResourceClaimTemplates(
...
+		}
+
+		// Then evaluate the request's own selectors.
+		allMatch, err := evaluateSelectorsOnDevice(ctx, requestSelectors, dev)
+		if err != nil {
+			if firstEvalErr == nil {
+				firstEvalErr = err
+			}
+			continue
```
mimowo: Why don't just return error here early?

pkg/dra/claims.go
```diff
@@ -192,3 +225,251 @@ func GetResourceRequestsForResourceClaimTemplates(
...
+				ctrl.LoggerFrom(ctx).V(3).Info("CEL evaluation error encountered while matching devices",
+					"deviceClassName", cr.deviceClassName, "error", firstEvalErr)
+			}
+			allErrs = append(allErrs, field.Invalid(
+				basePath.Child("devices", "requests").Index(cr.index).Child("exactly", "selectors"),
+				nil,
+				fmt.Sprintf("insufficient matching devices for CEL selector in DeviceClass %s: %d device(s) match in the cluster but %d requested",
+					cr.deviceClassName, matchCount, cr.count),
+			))
+		}
```
mimowo, line 461: Shouldn't we first fail if error, and only if not error, then inspect the results?

test/integration/singlecluster/controller/dra/dra_test.go
```diff
@@ -551,7 +584,67 @@ var _ = ginkgo.Describe("DRA Integration", ginkgo.Ordered, ginkgo.ContinueOnFail
...
+				DeviceRequest("device-request", "foo.example.com", 2).
+				WithCELSelectors("device.driver == \"nonexistent-driver\"").
+				Obj()
+			gomega.Expect(k8sClient.Create(ctx, rct)).To(gomega.Succeed())
+
+			ginkgo.By("Creating a workload with unsatisfiable CEL selectors")
+			wl := utiltestingapi.MakeWorkload("test-wl-cel-reject", ns.Name).
+				Queue("test-lq").
+				Obj()
+			wl.Spec.PodSets[0].Template.Spec.ResourceClaims = []corev1.PodResourceClaim{
```
mimowo: nit, use helpers like PodSets when constructing the workload, instead of mutating in place

test/integration/singlecluster/controller/dra/dra_test.go
```diff
@@ -528,7 +537,31 @@ var _ = ginkgo.Describe("DRA Integration", ginkgo.Ordered, ginkgo.ContinueOnFail
...
+						ResourceSliceCount: 1,
+					},
+					NodeName: new("fake-node"),
+					Devices: []resourcev1.Device{
+						{Name: "dev-0"},
+						{Name: "dev-1"},
+					},
+				},
+			}
+			gomega.Expect(k8sClient.Create(ctx, slice)).To(gomega.Succeed())
```
mimowo: use the helpers that we have: util.MustCreate

test/integration/singlecluster/controller/dra/dra_test.go
```diff
@@ -551,7 +584,67 @@ var _ = ginkgo.Describe("DRA Integration", ginkgo.Ordered, ginkgo.ContinueOnFail
...
+						Generation:         1,
+						ResourceSliceCount: 1,
+					},
+					NodeName: new("fake-node"),
+					Devices: []resourcev1.Device{
+						{Name: "dev-0"},
+					},
+				},
+			}
+			gomega.Expect(k8sClient.Create(ctx, slice)).To(gomega.Succeed())
```
mimowo: use the helpers that we have: util.MustCreate

test/integration/singlecluster/controller/dra/dra_test.go
```diff
@@ -551,7 +584,67 @@ var _ = ginkgo.Describe("DRA Integration", ginkgo.Ordered, ginkgo.ContinueOnFail
...
+			wl.Spec.PodSets[0].Template.Spec.ResourceClaims = []corev1.PodResourceClaim{
+				{
+					Name:                      "device-template",
+					ResourceClaimTemplateName: new("cel-reject-template"),
+				},
+			}
+			wl.Spec.PodSets[0].Template.Spec.Containers[0].Resources.Claims = []corev1.ResourceClaim{
+				{Name: "device-template"},
+			}
+			gomega.Expect(k8sClient.Create(ctx, wl)).To(gomega.Succeed())
```
mimowo: use the helpers that we have: util.MustCreate

pkg/dra/claims.go
```diff
@@ -192,3 +225,251 @@ func GetResourceRequestsForResourceClaimTemplates(
...
+		allMatch, err := evaluateSelectorsOnDevice(ctx, requestSelectors, dev)
+		if err != nil {
+			if firstEvalErr == nil {
+				firstEvalErr = err
+			}
+			continue
+		}
+
+		if allMatch {
+			consumed[devIdx] = true
```
mimowo, line 384: What if the device is used by not fully? Couldn't a large device be able to accomodate more than one Pod?

pkg/dra/claims.go
```diff
@@ -192,3 +225,251 @@ func GetResourceRequestsForResourceClaimTemplates(
...
+// validateCELSelectorsAgainstDevices lists all ResourceSlices in the cluster and
+// evaluates the CEL selectors from each request against the actual devices.
+// For each request, it resolves the DeviceClassName to its DeviceClass and uses
+// the class selectors to pre-filter devices before evaluating the request's own
+// CEL selectors. This avoids running CEL against devices that belong to
+// unrelated drivers or classes.
+// If fewer devices match than requested, it returns field errors indicating the
+// workload is unsatisfiable, preventing quota from being consumed by workloads
+// whose pods can never be scheduled.
+func validateCELSelectorsAgainstDevices(ctx context.Context, cl client.Client, claimSpec *resourcev1.ResourceClaimSpec, basePath *field.Path) field.ErrorList {
```
mimowo, line 404: What is the test coverage of the new code? What are the lines which are not covered?

test/integration/singlecluster/controller/dra/dra_test.go
```diff
@@ -528,7 +537,31 @@ var _ = ginkgo.Describe("DRA Integration", ginkgo.Ordered, ginkgo.ContinueOnFail
...
+					NodeName: new("fake-node"),
+					Devices: []resourcev1.Device{
+						{Name: "dev-0"},
+						{Name: "dev-1"},
+					},
+				},
+			}
+			gomega.Expect(k8sClient.Create(ctx, slice)).To(gomega.Succeed())
+			defer func() {
+				gomega.Expect(k8sClient.Delete(ctx, slice)).To(gomega.Succeed())
```
mimowo: Instead of defer, can we delete the slices in AfterEach / AfterAll as we normally do for other resources?

pkg/dra/claims.go
kannon92: updated.

pkg/dra/claims.go
kannon92: updated.

pkg/dra/claims.go
kannon92: updated.

test/integration/singlecluster/controller/dra/dra_test.go
kannon92: updated.

test/integration/singlecluster/controller/dra/dra_test.go
kannon92: updated.

test/integration/singlecluster/controller/dra/dra_test.go
kannon92: updated.

test/integration/singlecluster/controller/dra/dra_test.go
kannon92: updated.

pkg/dra/claims.go
kannon92: Did some updates. The only areas that are not covered are a few of the continue loops. In this function, the failure to list ResourceSlices, cel compilation failure, and the logging of the cel failure. The coverage of the dra folder is 84%.

test/integration/singlecluster/controller/dra/dra_test.go
kannon92: updated.

pkg/dra/claims.go
mimowo: yes, looks great, left another round of comments, but this one is addressed

pkg/dra/claims.go
kannon92: I'm not exactly following you here. MIG/Partiionable devices would be supported via https://github.com/kubernetes-sigs/kueue/pull/10283.

Maybe the variable name is unfortunate because this is not consumable capacity.

kannon92: This CEL work is mainly about selecting the right device and I think more than 1 pod could bind to that device. 

I'm not actually sure how DRA limits the number of users that can request a certain device. @johnbelamaric any idea here?

mimowo: > I'm not exactly following you here. MIG/Partiionable devices would be supported via https://github.com/kubernetes-sigs/kueue/pull/10283.

I'm not an expert on DRA so I might be messing some concepts here, sorry if this is the case here.

My attention was drawn by the boolean map "consumed" which seems to assume the consumption of devices is binary, whereas IIUC some devices can be shared across multiple Pods or Workloads. IIUC this may happen even without Partiionable devices - if a single node is represented as a device it may still host multiple workloads.

If this is possible, then it would be great to add a unit or integration test to confirm sharing devices works well. 

Let me also cc @sohankunkerkar to review this bit who certainly understands it better.

kannon92: As of now I think each workload must use a ResourceClaimTemplate (so each pod would have a separate request). 

Resource Sharing can be done via a ResourceClaim but kueue does not support this at the moment. This would be how multiple pods could share the same request.

Right now, each RCT would be quoted separately. CEL basically evaluates the scheduling decision but I think the quota management would work fine. That is essentially what all of our integration tests are confirming.

sohankunkerkar: Adding to what @kannon92, the consumed map prevents two requests in the same claim from matching the same device entry. Devices in ResourceSlices are GPUs/accelerators, not nodes. For shared/partitionable device semantics, that's covered by #10283, which uses counter-based quota instead of binary device consumption

pkg/dra/claims.go
kannon92: good call. fixed.

pkg/dra/claims.go
kannon92: fixed.

pkg/dra/claims.go
```diff
@@ -66,10 +82,12 @@ func countDevicesPerClass(claimSpec *resourcev1.ResourceClaimSpec) (resources.Re
 			return nil, allErrs
 		}
 
-		switch {
-		case len(req.Exactly.Selectors) > 0:
-			allErrs = append(allErrs, field.Invalid(field.NewPath("devices", "requests").Index(i).Child("exactly", "selectors"), nil, "CEL selectors are not supported"))
+		if err := validateCELSelectors(req.Exactly.Selectors, field.NewPath("devices", "requests").Index(i).Child("exactly", "selectors")); err != nil {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("devices", "requests").Index(i).Child("exactly", "selectors"), nil, err.Error()))
```
mimowo, line 86: Precompute the path `field.Invalid(field.NewPath("devices", "requests")` as var `devicesRequestsPath`

Then compute `devicesRequestsPath.Index(i).Child("exactly", "selectors")`.` before calling validateCELSelectors avoid computing twice.

This comment is not about performance, but readability by avoiding code duplication

pkg/dra/claims.go
```diff
@@ -192,3 +224,242 @@ func GetResourceRequestsForResourceClaimTemplates(
...
+	}
+
+	var sliceList resourcev1.ResourceSliceList
+	if err := cl.List(ctx, &sliceList); err != nil {
+		return field.ErrorList{field.InternalError(basePath, fmt.Errorf("failed to list ResourceSlices: %w", err))}
+	}
+
+	clusterDevices := buildDeviceListFromSlices(sliceList.Items)
+	classCache := make(map[string]*resourcev1.DeviceClass)
+	consumed := make(map[int]bool)
```
mimowo, line 417: nit, it I think using Set[int] would better capture the intention, as here the map[]->bool is basically as set.

mimowo (commented):
/lgtm
/approve
Thank you 👍 The two remaining comments from me are non-blocking and can be addressed post merge.

pkg/dra/claims.go
mimowo: Thanks for explaining. So iiuc currently we only support `ResourceClaimTemplate` and so we can assume the requests are separate.

pkg/dra/claims.go
kannon92: https://github.com/kubernetes-sigs/kueue/pull/11142

pkg/dra/claims.go
kannon92: https://github.com/kubernetes-sigs/kueue/pull/11142

pkg/dra/claims.go
kannon92: The name choice of `consumed` was unforunate. I opened up https://github.com/kubernetes-sigs/kueue/pull/11142 to rename to make it more clear of the intention.

PR #9746, opened 2026-03-09
kueue-populator: support init different cq in single ns

### LINKED ISSUE #9745: Request kueue-populator support multi-team use different ClusterQueues in single namespace.

Add a `useClusterQueueName` as a option to `kueue-populator` that uses the
ClusterQueue name as the LocalQueue name.

Currently, when multiple ClusterQueues match a namespace via `namespaceSelector`,
only the first ClusterQueue gets a LocalQueue created, subsequent ones are skipped
with a warning event.

With `useClusterQueueName: true`, each matching ClusterQueue creates a LocalQueue
named after itself. Since ClusterQueue names are cluster-unique, there are no naming
conflicts within a namespace.

**Why is this needed**:

In multi-team / multi-tenant clusters, different teams use different ClusterQueues.
A single namespace can legitimately need access to more than one ClusterQueue (e.g.
a shared namespace that both `team-a` and `team-b` can submit workloads to).

**Completion requirements**:

- [ ] Design doc
- [x] API change
- [x] Docs update

samzong:
#### What type of PR is this?

/kind feature

Adds a `useClusterQueueName` config option to kueue-populator.
When enabled, the controller uses each ClusterQueue's name as the LocalQueue name.
This lets multiple ClusterQueues each have their own LocalQueue within the same namespace.

The change is fully backward-compatible: the default behavior (fixed name "default")
is unchanged. Set `useClusterQueueName: true` in the config file to opt in.

Fixes #9745

- Zero breaking changes — `useClusterQueueName` defaults to `false`
- Existing unit tests pass unchanged
- Config is file-based (YAML), consistent with how `localQueueName` and
  `managedJobsNamespaceSelector` are already configured

```release-note
Kueue-populator: Support creating LocalQueue instances named as their ClusterQueue target.
```


cmd/experimental/kueue-populator/pkg/config/config.go
```diff
@@ -29,6 +29,9 @@ type Configuration struct {
 	// LocalQueueName is the name of the LocalQueue to create by default.
 	LocalQueueName string `json:"localQueueName,omitempty"`
 
+	// UseClusterQueueName uses the ClusterQueue name as the LocalQueue name.
+	UseClusterQueueName bool `json:"useClusterQueueName,omitempty"`
```
mimowo: This bool is a bit inflexible. Maybe we could introduce an enum `LocalQueueNameMode` with values "Static" (Default), or "AsClusterQueue". Then the PR would only introduce the new mode "AsClusterQueue". Then `LocalQueueName` couldn't be specified when `LocalQueueNameMode`: `AsClusterQueue` is set.

wdyt?

mimowo (commented):
Please also extend the testing, at least with unit tests, and the KEP.

samzong:
@mimowo it's all fix and upgrade KEP doc.

cmd/experimental/kueue-populator/pkg/controller/controller_test.go
```diff
@@ -485,6 +593,24 @@ func TestKueuePopulatorReconciler(t *testing.T) {
...
 			}
 
+			slices.SortFunc(gotEvents, func(a, b Event) int {
+				if a.Message < b.Message {
+					return -1
+				}
+				if a.Message > b.Message {
+					return 1
+				}
+				return 0
```
mimowo, line 603: nit, use strings.Compare function iirc, same below

mimowo (commented):
Thanks 👍 I left some comment, but it can be addressed in a follow up
/lgtm
/approve

mimowo:
/release-note-edit
```release-note
Kueue-populator: Support creating LocalQueue instances named as their ClusterQueue target.
```

PR #9749, opened 2026-03-09
Bump ShortConsistentDuration.

mbobrovskyi:
#### What type of PR is this?
/kind cleanup

Remove unnecessary ShortConsistentDuration.

Fixes #

```release-note
NONE
```

### DIFF
--- a/test/util/constants.go
+++ b/test/util/constants.go
@@ -43,8 +43,8 @@ const (
 	// need started and the time it takes for a change in ready probe response triggers
 	// a change in the deployment status.
 	StartUpTimeout          = 5 * time.Minute
-	ShortConsistentDuration = 10 * time.Millisecond
 	ConsistentDuration      = 1 * time.Second
+	ShortConsistentDuration = 100 * time.Millisecond
 	ShortInterval           = 10 * time.Millisecond
 	Interval                = time.Millisecond * 250
 	LongInterval            = time.Second * 1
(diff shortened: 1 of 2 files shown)


test/util/constants.go
```diff
@@ -42,12 +42,11 @@ const (
...
 	// need started and the time it takes for a change in ready probe response triggers
 	// a change in the deployment status.
-	StartUpTimeout          = 5 * time.Minute
-	ShortConsistentDuration = 10 * time.Millisecond
-	ConsistentDuration      = 1 * time.Second
-	ShortInterval           = 10 * time.Millisecond
-	Interval                = time.Millisecond * 250
-	LongInterval            = time.Second * 1
+	StartUpTimeout     = 5 * time.Minute
+	ConsistentDuration = 10 * time.Millisecond
```
mimowo: I would suggest to make it a bit longer actually. 10ms is enough in some cases but not all. Maybe 50ms? Then we will do 5 checks with the 10ms interval

mbobrovskyi: Done

test/util/constants.go
```diff
@@ -43,8 +43,8 @@ const (
 	// need started and the time it takes for a change in ready probe response triggers
 	// a change in the deployment status.
 	StartUpTimeout          = 5 * time.Minute
-	ShortConsistentDuration = 10 * time.Millisecond
 	ConsistentDuration      = 1 * time.Second
```
mimowo, line 46: Could we maybe actually combine ConsistentDuration=1s and ShortConsistentDuration into 100ms? :)

mbobrovskyi: Done

mimowo, line 46: Do you think we could replace all uses of ConsistentDuration with `ShortConsistentDuration`? It could be a follow up, but it feels we don't need 1s to wait, wdyt?

PR #9753, opened 2026-03-09
Add skip_reinstall flag to omit redeploy of kueue to e2e cluster.

### LINKED ISSUE #9414: Add an option to E2E_MODE=dev to skip rebuild (and re-install) of Kueue

Add the SKIP_REBUILD=true option which would skip re-building and re-installing of Kueue.

Maybe it could be also called SKIP_REINSTALL actually.

**Why is this needed**:

This will be useful when we are done with code and need to repeat multiple times the tests only. 

This will save time in that case.

reruno:
#### What type of PR is this?
/kind cleanup
/area testing

Add skip_reinstall flag to omit redeploy of kueue to e2e cluster.

Fixes #9414 

```release-note
NONE
```


hack/testing/e2e-common.sh
```diff
@@ -248,7 +251,7 @@ function ensure_kind_cluster {
             return 0
         fi
 
-        echo "Creating kind cluster (E2E_MODE=dev): $name"
+        echo "Creating kind cluster x (E2E_MODE=dev): $name"
```
mszadkow: x?

hack/testing/e2e-common.sh
```diff
@@ -266,6 +269,15 @@ function e2e_should_delete_cluster {
...
 
+# returns
+# 0 (true) if E2E_SKIP_REINSTALL = true
+# 1 (false) if E2E_SKIP_REINSTALL = false
+function e2e_should_skip_reinstall {
+    # inverted because e2e_is_truthy returns 0 (success code) on E2E_SKIP_REINSTALL=false 
+    # and 1 (error code) on E2E_SKIP_REINSTALL=true
+    ! e2e_is_truthy "${E2E_SKIP_REINSTALL}"; 
+}
+
```
mszadkow: maybe we could do it with less comments and more transparency?

mbobrovskyi: I would propose avoiding this function entirely. I think it’s clearer to check `E2E_SKIP_REINSTALL = true` directly instead.

hack/testing/e2e-common.sh
```diff
@@ -33,6 +34,9 @@ fi
 # When set (truthy), external operators are forcefully re-installed_version even in E2E_MODE=dev.
 export E2E_ENFORCE_OPERATOR_UPDATE="${E2E_ENFORCE_OPERATOR_UPDATE:-false}"
 
+# When set to value "true", skip Kueue re-install if controller deployment already exists.
+export E2E_SKIP_REINSTALL="${E2E_SKIP_REINSTALL:-false}"
```
mbobrovskyi, line 38: Please propagate this variable from the Makefile-test.mk.

hack/testing/e2e-common.sh
```diff
@@ -543,6 +547,15 @@ function cluster_kueue_deploy {
         upgrade_test_flow "$1"
         return
     fi
+
+    if [[ "${E2E_SKIP_REINSTALL}" == "true" ]]; then
+        if e2e_deployment_exists "$1" "${KUEUE_NAMESPACE}" "${KUEUE_CONTROLLER_DEPLOYMENT}"; then
+            echo "Kueue controller already exists in namespace '${KUEUE_NAMESPACE}', skipping reinstall (E2E_SKIP_REINSTALL=${E2E_SKIP_REINSTALL})."
```
mbobrovskyi:
```suggestion
            echo "Kueue controller already exists in namespace '${KUEUE_NAMESPACE}', skipping reinstall"
```

hack/testing/e2e-common.sh
```diff
@@ -20,6 +20,7 @@ export KIND="$ROOT_DIR"/bin/kind
 export YQ="$ROOT_DIR"/bin/yq
 export HELM="$ROOT_DIR"/bin/helm
 export KUEUE_NAMESPACE="${KUEUE_NAMESPACE:-kueue-system}"
+export KUEUE_CONTROLLER_DEPLOYMENT="${KUEUE_CONTROLLER_DEPLOYMENT:-kueue-controller-manager}"
```
mbobrovskyi:
```suggestion
export KUEUE_CONTROLLER_NAME="${KUEUE_CONTROLLER_NAME:-kueue-controller-manager}"
```
WDYT?

reruno: I think it is good

I added deployment suffix because it underscores that it is k8s deployment to make it more clear, but I think _NAME is good too, I can rename it if you want

hack/testing/e2e-common.sh
```diff
@@ -570,7 +583,7 @@ function helm_install {
       --set "controllerManager.manager.image.repository=${IMAGE_TAG%:*}" \
       --set "controllerManager.manager.image.tag=${IMAGE_TAG##*:}" \
       --create-namespace \
-      --namespace kueue-system \
+      --namespace "${KUEUE_NAMESPACE}" \
```
mbobrovskyi: Let's move KUEUE_NAMESPACE changes to separate PR.

hack/testing/e2e-common.sh
reruno: Removed I was debugging stuff :)

hack/testing/e2e-common.sh
reruno: Fixed

hack/testing/e2e-common.sh
```diff
@@ -215,7 +219,7 @@ function cluster_collect_artifacts {
 
     kubectl config ${kubectl_args[@]+"${kubectl_args[@]}"} use-context "kind-${name}" || true
     $KIND export logs "$ARTIFACTS" --name "$name" || true
-    kubectl describe pods ${kubectl_args[@]+"${kubectl_args[@]}"} -n kueue-system > "$ARTIFACTS/${name}-kueue-system-pods.log" || true
+    kubectl describe pods ${kubectl_args[@]+"${kubectl_args[@]}"} -n "${KUEUE_NAMESPACE}" > "$ARTIFACTS/${name}-${KUEUE_NAMESPACE}-pods.log" || true
```
mbobrovskyi: Also this please revert.

hack/testing/e2e-common.sh
```diff
@@ -1127,7 +1140,7 @@ function upgrade_test_flow {
       sed 's|imagePullPolicy: Always|imagePullPolicy: IfNotPresent|g' | \
       kubectl apply --server-side -f -
 
-    kubectl wait --for=condition=available --timeout=180s deployment/kueue-controller-manager -n kueue-system
+    kubectl wait --for=condition=available --timeout=180s deployment/"${KUEUE_DEPLOYMENT_NAME}" -n "${KUEUE_NAMESPACE}"
```
mbobrovskyi:
```suggestion
    kubectl wait --for=condition=available --timeout=180s deployment/"${KUEUE_DEPLOYMENT_NAME}" -n kueue-system
```
Here also please revert for now.

mbobrovskyi (commented):
It would be also great to update contribution guidelines https://github.com/epam/kubernetes-kueue/blob/06486c559b20851f992d6a52df2cf6db0ba1cc20/site/content/en/docs/contribution_guidelines/testing.md?plain=1#L123.

hack/testing/e2e-common.sh
```diff
@@ -20,6 +20,7 @@ export KIND="$ROOT_DIR"/bin/kind
 export YQ="$ROOT_DIR"/bin/yq
 export HELM="$ROOT_DIR"/bin/helm
 export KUEUE_NAMESPACE="${KUEUE_NAMESPACE:-kueue-system}"
+export KUEUE_DEPLOYMENT_NAME="${KUEUE_DEPLOYMENT_NAME:-kueue-controller-manager}"
```
mimowo: I'm not sure we need to parametrize that. IIRC we can paramterize with Helm the namespace, but not the controlle-name. So, I think we don't need to parametrize it here.

mbobrovskyi: This is just a constant to avoid duplicating the name. I also thought that maybe we don’t need to set a default value, but it doesn’t seem harmful.

reruno: I will change value to string like that, to avoid confusion 
`export KUEUE_DEPLOYMENT_NAME="kueue-controller-manager"`

mbobrovskyi: Actually, I tried changing the chart name in the Helm charts, and the deployment name changed as well. So maybe this isn’t a bad idea.

mimowo: I'm not sure we are really changing or parameterizing all places with this change. This looks like half-done work (or 10% done). For example, I quickly found many occurrences of the hardcoded string, say in e2e_test.go, some kueueviz files, etc. 

So, this will not work anyway I think. I would do this when we have the objective to support it.

mbobrovskyi: OK. Let's keep it as constant for now.

mbobrovskyi:
@reruno could you please prepare cherry-picks?

reruno:
@mbobrovskyi I think 0.17 doesn't need cherry pick, because the changes were included in 0.17 release

PR #9755, opened 2026-03-09
LeaderWorkerSet integration when LeaderWorkerSet created with RestartPolicy should only recreate the deleted pod when policy is set to NoneRestartPolicy

### LINKED ISSUE #9750: LeaderWorkerSet integration when LeaderWorkerSet created with Restart Policy should only recreate the deleted pod when policy is set to NoneRestartPolicy
**Which test is flaking?**:
LeaderWorkerSet integration when LeaderWorkerSet created with Restart Policy should only recreate the deleted pod when policy is set to NoneRestartPolicy

https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/periodic-kueue-test-e2e-release-0-16-1-34/2030157647042318336
**Failure message or logs**:
```

End To End Suite: kindest/node:v1.34.3: [It] LeaderWorkerSet integration when LeaderWorkerSet created with Restart Policy should only recreate the deleted pod when policy is set to NoneRestartPolicy [area:singlecluster, feature:leaderworkerset] expand_less	1m24s
{Timed out after 45.001s.
The function passed to Eventually failed at /home/prow/go/src/kubernetes-sigs/kueue/test/e2e/singlecluster/leaderworkerset_test.go:1217 with:
only the deleted pod should be recreated with NoneRestartPolicy
Expected
    <int>: 0
to equal
    <int>: 1 failed [FAILED] Timed out after 45.001s.
The function passed to Eventually failed at /home/prow/go/src/kubernetes-sigs/kueue/test/e2e/singlecluster/leaderworkerset_test.go:1217 with:
only the deleted pod should be recreated with NoneRestartPolicy
Expected
    <int>: 0
to equal
    <int>: 1
In [It] at: /home/prow/go/src/kubernetes-sigs/kueue/test/e2e/singlecluster/leaderworkerset_test.go:1219 @ 03/07/26 06:03:28.417
}
```

vladikkuzn:
#### What type of PR is this?
/kind flake

The old assertion identified the recreated pod by comparing current pods against the original map keyed by pod name. That is brittle because after deletion/recreation, the replacement can be observed in a different way than the test assumes, so the check can temporarily see no “same-name, new-UID” pod and incorrectly conclude that nothing was recreated.

I changed the test to assert the actual behavior more directly:
exactly 3 pods exist again,
exactly 1 pod has a new UID,
2 original pod UIDs are still present,
the newly created pod corresponds to the deleted worker index.
So the problem was: the test was using an unstable way to detect pod replacement, which made it report 0 recreated pods even when the system behavior was valid.

Fixes #9750

```release-note
NONE
```

### DIFF
--- a/test/e2e/singlecluster/leaderworkerset_test.go
+++ b/test/e2e/singlecluster/leaderworkerset_test.go
@@ -27,6 +27,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	leaderworkersetv1 "sigs.k8s.io/lws/api/leaderworkerset/v1"
@@ -1037,7 +1038,7 @@ var _ = ginkgo.Describe("LeaderWorkerSet integration", ginkgo.Label("area:single
 			})
 
 			var podToDelete *corev1.Pod
-			originalPodUIDs := make(map[string]types.UID)
+			originalPodUIDSet := sets.New[types.UID]()
 			ginkgo.By("Select a worker pod to delete", func() {
 				pods := &corev1.PodList{}
 				gomega.Expect(k8sClient.List(ctx, pods, client.MatchingLabels{
@@ -1046,7 +1047,7 @@ var _ = ginkgo.Describe("LeaderWorkerSet integration", ginkgo.Label("area:single
 				gomega.Expect(pods.Items).To(gomega.HaveLen(3))
 
 				for i, pod := range pods.Items {
-					originalPodUIDs[pod.Name] = pod.UID
+					originalPodUIDSet.Insert(pod.UID)
 					if pod.Labels[leaderworkersetv1.WorkerIndexLabelKey] != "0" {
 						podToDelete = &pods.Items[i]
 					}
@@ -1054,8 +1055,18 @@ var _ = ginkgo.Describe("LeaderWorkerSet integration", ginkgo.Label("area:single
 				gomega.Expect(podToDelete).NotTo(gomega.BeNil(), "Couldn't find a worker pod to delete")
 			})
 
+			deletedPodUID := podToDelete.UID
+			deletedPodKey := client.ObjectKeyFromObject(podToDelete)
 			ginkgo.By("Delete the worker pod", func() {
 				gomega.Expect(k8sClient.Delete(ctx, podToDelete)).Should(gomega.Succeed())
+			})
+
+			ginkgo.By("Wait for the deleted pod to be recreated with a new UID", func() {
+				gomega.Eventually(func(g gomega.Gomega) {
+					pod := &corev1.Pod{}
+					g.Expect(k8sClient.Get(ctx, deletedPodKey, pod)).To(gomega.Succeed())
+					g.Expect(pod.UID).NotTo(gomega.Equal(deletedPodUID), "pod should be recreated with a new UID")
+				}, util.LongTimeout, util.Interval).Should(gomega.Succeed())
 			})
 
 			ginkgo.By("Wait for all pods to be recreated with new UIDs", func() {
@@ -1065,14 +1076,12 @@ var _ = ginkgo.Describe("LeaderWorkerSet integration", ginkgo.Label("area:single
 						leaderworkersetv1.SetNameLabelKey: lws.Name,
 					}, client.InNamespace(lws.Namespace))).Should(gomega.Succeed())
 
-					recreatedCount := 0
+					currentUIDs := sets.New[types.UID]()
 					for _, pod := range pods.Items {
-						originalUID, existed := originalPodUIDs[pod.Name]
-						if existed && pod.UID != originalUID {
-							recreatedCount++
-						}
+						currentUIDs.Insert(pod.UID)
 					}
-					g.Expect(recreatedCount).To(gomega.Equal(3), "all pods should be recreated with new UIDs due to RecreateGroupOnPodRestart policy")
+					survived := originalPodUIDSet.Intersection(currentUIDs)
+					g.Expect(survived).To(gomega.BeEmpty(), "all pods should be recreated with new UIDs due to RecreateGroupOnPodRestart policy")
 				}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
 			})
 
@@ -1120,7 +1129,7 @@ var _ = ginkgo.Describe("LeaderWorkerSet integration", ginkgo.Label("area:single
 			})
 
 			var podToDelete *corev1.Pod
-			originalPodUIDs := make(map[string]types.UID)
+			originalPodUIDSet := sets.New[types.UID]()
 			ginkgo.By("Select a worker pod to delete", func() {
 				pods := &corev1.PodList{}
 				gomega.Expect(k8sClient.List(ctx, pods, client.MatchingLabels{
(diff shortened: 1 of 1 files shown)


test/e2e/singlecluster/leaderworkerset_test.go
```diff
@@ -1120,7 +1120,7 @@ var _ = ginkgo.Describe("LeaderWorkerSet integration", ginkgo.Label("area:single
 			})
 
 			var podToDelete *corev1.Pod
-			originalPodUIDs := make(map[string]types.UID)
+			originalPodUIDSet := make(map[types.UID]struct{})
```
mimowo: why map not just set?

test/e2e/singlecluster/leaderworkerset_test.go
```diff
@@ -1148,22 +1150,26 @@ var _ = ginkgo.Describe("LeaderWorkerSet integration", ginkgo.Label("area:single
 					g.Expect(k8sClient.List(ctx, pods, client.MatchingLabels{
 						leaderworkersetv1.SetNameLabelKey: lws.Name,
 					}, client.InNamespace(lws.Namespace))).Should(gomega.Succeed())
+					g.Expect(pods.Items).To(gomega.HaveLen(3))
 
+					remainingOriginalUIDs := 0
```
mimowo: Instead just count, could we make it a set too? So that the failure message will contain the UIDs if test does not pass on the assert

test/e2e/singlecluster/leaderworkerset_test.go
```diff
@@ -1148,22 +1150,32 @@ var _ = ginkgo.Describe("LeaderWorkerSet integration", ginkgo.Label("area:single
...
+						g.Expect(pod.Labels[leaderworkersetv1.WorkerIndexLabelKey]).To(
+							gomega.Equal(deletedWorkerIndex),
+							"only the deleted worker slot should be recreated",
+						)
 					}
+
 					g.Expect(recreatedCount).To(gomega.Equal(1), "only the deleted pod should be recreated with NoneRestartPolicy")
-					g.Expect(unchangedCount).To(gomega.Equal(2), "other pods should NOT be recreated with NoneRestartPolicy")
+					g.Expect(remainingOriginalUIDs.Len()).To(
+						gomega.Equal(2),
```
mimowo: Use gomaga.HaveLen instead. This will elimianat the need for `remainingOriginalUIDs.UnsortedList(),` later

mimowo:
> Added explanation to the description

I still don't understand what was the issue with the previous code. Sure, I'm +1 to improve the test code to use sets rather than counters for the ease of investigations, but I don't see where was the bug. Could you maybe detail the explanation by tracking the events and correlating them with kubelet logs? This would be very helpful I think.

test/e2e/singlecluster/leaderworkerset_test.go
sohankunkerkar: Could you switch to `Switch to sets.New[types.UID]()` here as well?

test/e2e/singlecluster/leaderworkerset_test.go
```diff
@@ -1129,41 +1140,58 @@ var _ = ginkgo.Describe("LeaderWorkerSet integration", ginkgo.Label("area:single
...
+
 			ginkgo.By("Verify if only the deleted pod was recreated", func() {
 				gomega.Eventually(func(g gomega.Gomega) {
 					pods := &corev1.PodList{}
 					g.Expect(k8sClient.List(ctx, pods, client.MatchingLabels{
 						leaderworkersetv1.SetNameLabelKey: lws.Name,
 					}, client.InNamespace(lws.Namespace))).Should(gomega.Succeed())
+					g.Expect(pods.Items).To(gomega.HaveLen(3))
 
+					remainingOriginalUIDs := sets.New[types.UID]()
```
sohankunkerkar: The sets package already has Intersection and Difference. So you can replace the entire loop + counter with:

```go
g.Expect(pods.Items).To(gomega.HaveLen(3))                                                                                                                                 
              
  currentUIDs := sets.New[types.UID]()                                                                                                                                       
  for _, pod := range pods.Items {                                                                                                                                           
      currentUIDs.Insert(pod.UID)                                                                                                                                            
  }           

  survived := originalPodUIDSet.Intersection(currentUIDs)
  newUIDs := currentUIDs.Difference(originalPodUIDSet)

  g.Expect(survived).To(gomega.HaveLen(2), "2 original pods should survive")
  g.Expect(survived.Has(deletedPodUID)).To(gomega.BeFalse(), "deleted pod should not survive")
  g.Expect(newUIDs).To(gomega.HaveLen(1), "exactly 1 replacement pod")

  var replacement *corev1.Pod
  for i := range pods.Items {
      if newUIDs.Has(pods.Items[i].UID) {
          replacement = &pods.Items[i]
          break
      }
  }
  g.Expect(replacement).NotTo(gomega.BeNil())
  g.Expect(replacement.Labels[leaderworkersetv1.WorkerIndexLabelKey]).To(
      gomega.Equal(deletedWorkerIndex),
      "only the deleted worker slot should be recreated",
  )

```

test/e2e/singlecluster/leaderworkerset_test.go
```diff
@@ -1058,6 +1060,15 @@ var _ = ginkgo.Describe("LeaderWorkerSet integration", ginkgo.Label("area:single
 				gomega.Expect(k8sClient.Delete(ctx, podToDelete)).Should(gomega.Succeed())
 			})
 
+			ginkgo.By("Wait for the deleted pod to be terminated", func() {
+				deletedPodKey := client.ObjectKeyFromObject(podToDelete)
+				gomega.Eventually(func(g gomega.Gomega) {
+					pod := &corev1.Pod{}
+					err := k8sClient.Get(ctx, deletedPodKey, pod)
+					g.Expect(apierrors.IsNotFound(err)).To(gomega.BeTrue(), "deleted pod should be removed from the API")
```
PannagaRao: `IsNotFound` may not always work here — LWS uses a StatefulSet which recreates the pod with the same name, and the replacement could be created before we even check for deletion. So `Get(ctx, deletedPodKey, pod)` will always find a pod and never return `NotFound`.                 

Maybe checking for a UID change would be more reliable since that confirms recreation regardless of timing?
                                                                                                                                                                                                                    
```                                                                                                                                                                                                        
  ginkgo.By("Wait for the deleted pod to be recreated with a new UID", func() {
          pod := &corev1.Pod{}
          g.Expect(k8sClient.Get(ctx, deletedPodKey, pod)).To(gomega.Succeed())
          g.Expect(pod.UID).NotTo(gomega.Equal(deletedPodUID),
              "pod should be recreated with a new UID")
      }, util.LongTimeout, util.Interval).Should(gomega.Succeed())
  })
```

hack/testing/e2e-test.sh
```diff
@@ -68,7 +68,11 @@ if [ "$E2E_RUN_ONLY_ENV" = "true" ]; then
...
-# Quote GINKGO_ARGS so values with spaces (e.g. --focus='should ensure the eviction') are not
-# word-split; otherwise fragments get interpreted as packages and cause "flag after package list".
-$GINKGO ${GINKGO_ARGS:+"$GINKGO_ARGS"} --junit-report=junit.xml --json-report=e2e.json --output-dir="$ARTIFACTS" -v ./test/e2e/"${E2E_TARGET_FOLDER}"/...
+# Echo command to be executed, which is useful for debugging and understanding the test execution flow.
+echo "Running Ginkgo with the following command:"
+echo "$GINKGO ${GINKGO_ARGS:+$GINKGO_ARGS} --junit-report=junit.xml --json-report=e2e.json --output-dir=\"$ARTIFACTS\" -v ./test/e2e/\"${E2E_TARGET_FOLDER}\"/..."
+
+# Pass GINKGO_ARGS unquoted so multiple flags become separate arguments (avoids "invalid value for flag -procs").
+# Note: values with spaces (e.g. --focus='some spec') will need to be passed as a single arg from the caller.
+$GINKGO ${GINKGO_ARGS:+$GINKGO_ARGS} --junit-report=junit.xml --json-report=e2e.json --output-dir="$ARTIFACTS" -v ./test/e2e/"${E2E_TARGET_FOLDER}"/...
```
mbobrovskyi: Please revert this.

PR #8551, opened 2026-01-12
[KEP] Add a mechanism to consider preemption cost when finding preemption candidates

vladikkuzn:
#### What type of PR is this?
/kind documentation

#### Which issue(s) this PR fixes:

Part of #7990

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #9770, opened 2026-03-09
[FairSharing] Workloads that were recently admitted are protected from fair sharing preemption for at least minAdmitDuration

mukund-wayve:
#### What type of PR is this?

/kind feature

/kind api-change

Adds `minAdmitDuration` to the FairSharing configuration, gated behind the `FairSharingMinAdmitDuration` feature gate. The feature is alpha and disabled by default. When enabled and configured, workloads admitted less than `minAdmitDuration` ago are excluded from fair sharing preemption.

This helps prevent preemption oscillation, where competing ClusterQueues repeatedly evict each other’s workloads before they make meaningful progress. The issue became more visible as I was working with the historical borrowing strategy KEP in #9737, but this configuration is generic and expected to provide value across Fair Sharing strategies. The behavior is analogous to Slurm’s `PreemptExemptTime`.

This change is conceptually similar to #8646, but operates at a different level. #8646 applies within a single ClusterQueue, whereas `minAdmitDuration` applies across ClusterQueues during fair sharing. The two mechanisms are complementary. This implementation reuses the minimum value of one minute for `minAdmitDuration`, as suggested in #8646.

Part-of #9734 

```release-note
[FairSharing] Add minAdmitDuration to the FairSharing configuration. When enabled, workloads that were recently admitted are protected from fair sharing preemption for a configurable minimum duration. This feature is gated behind the FairSharingMinAdmitDuration feature gate and is in alpha.
```


pkg/features/kube_features.go
```diff
@@ -282,6 +282,13 @@ const (
 	// issue: https://github.com/kubernetes-sigs/kueue/issues/9694
 	// Skip equivalent inadmissible workloads in BestEffortFIFO scheduling.
 	SchedulingEquivalenceHashing featuregate.Feature = "SchedulingEquivalenceHashing"
+
+	// owner: @mukund-wayve
```
kannon92, line 286: I would suggest a KEP first to discuss.

apis/config/v1beta2/configuration_types.go
```diff
@@ -541,6 +541,14 @@ type FairSharing struct {
 	//   As a result, the strategy chooses to preempt workloads with the lowest priority and
 	//   newest start time first.
 	PreemptionStrategies []PreemptionStrategy `json:"preemptionStrategies"`
+
+	// minAdmitDuration is the minimum time a workload must be admitted before
```
kannon92, line 545: I don't love this as a global configuration. 

Will one setting work for all Kueue users or should there be workload tunability for this?

I guess one good thing about configuration is that it is set by an admin but I imagine someone using autoscaling may want a different value than someone not.

Or large scale jobs want to run for longer. anyway I don't think this is trivial so prefer a KEP to weigh pro/cons

pkg/features/kube_features.go
kannon92: https://github.com/kubernetes-sigs/kueue/pull/9770/changes#r2907362145

apis/config/v1beta2/configuration_types.go
mukund-wayve: Sounds good. I will raise a KEP and add you as a reviewer. But a quick response. I will add these to the KEP as well.

> workload tunability for this

related

> Or large scale jobs want to run for longer

Potentially. but I'm just cautious that we do not want to allow users to game the system. To prevent abuse, we may want policy controls which limit values which complicates this further. Then we'll have to let admins control the min and max values or validate this via policy engines like kyverno or other external controllers.

[Preemption costs](https://github.com/kubernetes-sigs/kueue/pull/8551/changes) is one lever. 


> autoscaling may want a different value than someone not.

Just to make sure I understand this, is it because the QuotaReservedTime does not capture the time it takes to bring up a node?

kannon92: > Just to make sure I understand this, is it because the QuotaReservedTime does not capture the time it takes to bring up a node?

I'm not actually sure here. I'm only starting to get environments supporting autoscaling.
I know that workloads get admited into ClusterQueue and that triggers creation of nodes. I don't know if that means we reserved quota.

Do you know @mimowo?

mimowo: QuotaReserved does not entail the Nodes are Provisioned yet.

I wouldn't go into tunability per-workload, because we don't want users to get the power to say "you cannot preempt me for the next 100 years" (or whatever). 

We could consider tunability at the CQ level, but this could create assymetry between CQs within cohort, so I'm not sure this is a good idea.

Global configuration sounds reasonable to me. If someone needs tunability, they could set 10min just to prevent preemption after the admission, then they could use "priority-boost". Yes, it does not protect against preemption indefinitely, but this is "by design" - we want eventual consistency in case of  FairSharing.

mukund-wayve: I've raised [a KEP](https://github.com/kubernetes-sigs/kueue/pull/9877) for this. Please review when you can

PR #9225, opened 2026-02-13
Add KEP-7066: Custom metadata labels for Kueue metrics

IrvingMg:
#### What type of PR is this?

/kind documentation

#### What this PR does / why we need it:

KEP proposing a configuration option to promote Kubernetes labels from ClusterQueue and LocalQueue objects into Prometheus metric labels.

#### Which issue(s) this PR fixes:

KEP for #7066 

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #9774, opened 2026-03-09
Add custom metadata labels for metrics

### LINKED ISSUE #7066: Additional metadata label for Kueue metrics

Add ability to add metadata based on Clusterqueue labels/annotations for prometheus metrics like: `kueue_cluster_queue_resource_usage`. These can be added for any `kueue_cluster_queue_*` or even extended to other metrics if useful.

**Why is this needed**:

It is useful to get custom metadata into the Clusterqueue metrics as labels which could make filtering easier for a variety of use cases. 

As an example, let us say I want to calculate sum of usage for clusterqueues aggregated over some custom label, example all clusterqueues by parentCohort, or clusterqueues belonging to one team/organization that is available as a label in Clusterqueue metadata.

**Completion requirements**:
`kueue_cluster_queue_*` metrics including `kueue_cluster_queue_resource_usage` have a label which could have values injected from clusterqueue metadata.

- [ ] Design doc
- [ ] API change
- [x ] Docs update

IrvingMg:
#### What type of PR is this?

/kind feature

Implements KEP-7066 (#9225)

Fixes #7066 

```release-note
Observability: Add custom Prometheus metric labels from CQ/LQ/Cohort metadata via metrics.customLabels config.
```


apis/config/v1beta2/configuration_types.go
```diff
@@ -176,6 +176,33 @@ type ControllerMetrics struct {
 	// metrics will be reported.
 	// +optional
 	EnableClusterQueueResources bool `json:"enableClusterQueueResources,omitempty"`
+
+	// CustomLabels is a list of entries whose values will be added as extra
+	// Prometheus labels (with a "custom_" prefix) on ClusterQueue, LocalQueue,
```
kannon92: These comments confuse me. Do you expect an admin to title the labels as custom_label_a? Or you will prepend custom_ to the labels?

IrvingMg: Kueue will prepend `custom_` to the labels. I've updated the comment on the `Name` field to make it more explicit.

IrvingMg:

Looks like infra issue: 

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/9774/pull-kueue-test-e2e-multikueue-main/2031316771612397568

apis/config/v1beta2/configuration_types.go
```diff
@@ -176,6 +176,33 @@ type ControllerMetrics struct {
...
+	// Prometheus labels on ClusterQueue, LocalQueue, and Cohort metrics.
+	// Requires the CustomMetricLabels feature gate.
+	// +optional
+	// +kubebuilder:validation:MaxItems=8
+	CustomLabels []ControllerMetricsCustomLabel `json:"customLabels,omitempty"`
+}
+
+// ControllerMetricsCustomLabel defines a Kubernetes label or annotation to promote
+// as a Prometheus metric label with a "custom_" prefix.
+type ControllerMetricsCustomLabel struct {
```
kannon92, line 190: Can you add information about what requirements we have on these strings?

Looks like it is a qualified name.

pkg/cache/scheduler/clusterqueue.go
```diff
@@ -96,6 +96,8 @@ type clusterQueue struct {
 	AdmissionScope *kueue.AdmissionScope
 
 	roleTracker *roletracker.RoleTracker
+
+	customLabelValues []string
```
kannon92: Our API supports annotations and labels. Is this just labels?

site/content/en/docs/reference/kueue-config.v1beta2.md
```diff
@@ -334,6 +334,61 @@ It can be set to &quot;0&quot; to disable the metrics serving.</p>
 metrics will be reported.</p>
 </td>
 </tr>
+<tr><td><code>customLabels</code><br/>
+<a href="#config-kueue-x-k8s-io-v1beta2-ControllerMetricsCustomLabel"><code>[]ControllerMetricsCustomLabel</code></a>
+</td>
+<td>
+   <p>CustomLabels is a list of entries whose values will be added as extra
```
kannon92, line 341: Can we also add some documentation for this for our site?

Not sure where it should be yet but we should document this feature.

test/integration/singlecluster/scheduler/fairsharing/fair_sharing_test.go
```diff
@@ -411,7 +411,8 @@ var _ = ginkgo.Describe("Scheduler", ginkgo.Label("feature:fairsharing"), func()
 
 			ginkgo.By("checking the weighted share metric")
```
kannon92, line 412: There needs to be a lot more testing of this feature.

Can we add a custom config test?

It would be good to have some negative testing of the feature gate at the integration level.

pkg/workload/workload.go
```diff
@@ -1610,8 +1615,8 @@ func reportEvictedWorkload(recorder record.EventRecorder, wl *kueue.Workload, cq
 	recorder.Event(wl, corev1.EventTypeNormal, eventReason, message)
 }
 
-func ReportPreemption(preemptingCqName kueue.ClusterQueueReference, preemptingReason string, targetCqName kueue.ClusterQueueReference, tracker *roletracker.RoleTracker) {
-	metrics.ReportPreemption(preemptingCqName, preemptingReason, targetCqName, tracker)
+func ReportPreemption(preemptingCqName kueue.ClusterQueueReference, preemptingReason string, tracker *roletracker.RoleTracker) {
```
kannon92: nit: can you do this cleanup separately?

pkg/scheduler/scheduler.go
```diff
@@ -878,15 +879,19 @@ func (s *Scheduler) recordWorkloadAdmissionEvents(newWorkload, originalWorkload
 	s.recorder.Eventf(newWorkload, corev1.EventTypeNormal, "Admitted", "Admitted by ClusterQueue %v, wait time since reservation was 0s", admission.ClusterQueue)
 
 	priorityClassName := workload.PriorityClassName(newWorkload)
-	metrics.AdmittedWorkload(admission.ClusterQueue, priorityClassName, waitTime, s.roleTracker)
+	cqCustomLabels := metrics.GetCQCustomLabelValues(admission.ClusterQueue)
+	metrics.AdmittedWorkload(admission.ClusterQueue, priorityClassName, waitTime, cqCustomLabels, s.roleTracker)
 	if features.Enabled(features.LocalQueueMetrics) {
```
kannon92, line 893: Can we verify in testing that this also works with LocalQueueMetrics?

Integration would be great here.

kannon92 (changes requested):
I see some testing gaps but high level it looks good.

I haven't done a deep review of the metrics code yet.

apis/config/v1beta2/configuration_types.go
IrvingMg: Done. Added string requirements to all three fields.

pkg/cache/scheduler/clusterqueue.go
IrvingMg: Renamed the field to `customMetricLabelValues` and added a comment clarifying it covers both.

site/content/en/docs/reference/kueue-config.v1beta2.md
IrvingMg: Done. Added a new page at `site/content/en/docs/tasks/manage/observability/configure_custom_metric_labels.md`.

pkg/workload/workload.go
IrvingMg: Done. Reverted.

test/integration/singlecluster/scheduler/fairsharing/fair_sharing_test.go
IrvingMg: Done. Added integration tests at `test/integration/singlecluster/controller/core/custom_labels_test.go`.

pkg/scheduler/scheduler.go
IrvingMg: Done.

pkg/cache/queue/manager_test.go
```diff
@@ -197,7 +197,7 @@ func TestUpdateClusterQueue(t *testing.T) {
 
 	// Put cq2 in the same cohort as cq1.
 	clusterQueues[1].Spec.CohortName = clusterQueues[0].Spec.CohortName
-	if err := manager.UpdateClusterQueue(ctx, clusterQueues[1], true); err != nil {
+	if err := manager.UpdateClusterQueue(ctx, clusterQueues[1], true, false); err != nil {
```
kannon92, line 202: Can you add a test case for testing the true case for UpdateClusterQueue?

pkg/metrics/reassign_vectors.go
```diff
@@ -0,0 +1,384 @@
...
+
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+
+	"sigs.k8s.io/kueue/pkg/constants"
+)
+
+// reassignMetricVectors recreates each metric variable with custom labels
```
kannon92: I really dislike this code.

I worry that we will not remember to update both reassign_vectors and metrics.go everytime we update or add a new metric.

kannon92: Cam we reuse the definitions in metrics.go somewhow?

pkg/controller/core/localqueue_controller.go
```diff
@@ -253,6 +264,13 @@ func (r *LocalQueueReconciler) Update(e event.TypedUpdateEvent[*kueue.LocalQueue
 	log := r.logger().WithValues("localQueue", klog.KObj(e.ObjectNew))
 	log.V(2).Info("Queue update event")
 
+	if metrics.LQCustomLabels.Store(string(utilqueue.Key(e.ObjectNew)), metrics.ExtractCustomLabelValues(e.ObjectNew.GetLabels(), e.ObjectNew.GetAnnotations())) {
```
mykysha: I wonder if it would be possible to commonize the label change check and clearing process of lq and cq metrics and call that function where needed

pkg/cache/queue/manager_test.go
IrvingMg: Done. I've added `TestUpdateClusterQueueLabelsUpdated`.

pkg/metrics/reassign_vectors.go
IrvingMg: The main constraint for extracting shared variables is that the metrics doc generator parses the metric definitions by looking for inline `prometheus.New*Vec(...)` declarations. If we refactor the metric definitions to avoid duplication, the generator will skip those metrics. We ran into the same constraint in this PR: https://github.com/kubernetes-sigs/kueue/pull/9487#discussion_r2859785784. 

But I’ll check if we can extend the doc generator.

IrvingMg: I've opened #9802, which extends the metricsdoc generator.

With that PR, I think we can consolidate `metrics.go` and `reassign_vectors.go` into a single `initMetrics(extraLabels)` function, so that metrics are constructed once with the correct labels from the start, eliminating the duplication entirely.

pkg/controller/core/localqueue_controller.go
IrvingMg: Done. I've added `StoreAndClear` method to commonize the pattern.

pkg/cache/queue/manager.go
```diff
@@ -304,7 +304,7 @@ func (m *Manager) AddClusterQueue(ctx context.Context, cq *kueue.ClusterQueue) e
 	return nil
 }
 
-func (m *Manager) UpdateClusterQueue(ctx context.Context, cq *kueue.ClusterQueue, specUpdated bool) error {
+func (m *Manager) UpdateClusterQueue(ctx context.Context, cq *kueue.ClusterQueue, specUpdated, labelsUpdated bool) error {
```
mimowo, line 316: Could you explain me what is the concept of updatingLabels, and why this is needed? I thought the labels are in the global configuration and so they cannot change during runtime. Is it just for writing tests?

IrvingMg: What is static (set once from the global config) is the list of label names to extract (e.g., extract `team` and `cost_center`). However, the actual values those labels resolve to are dynamic.

`labelsUpdated` is needed because custom label values come from each ClusterQueue’s labels/annotations and can change. When that happens, the old metric series must be cleared and re-reported with the new values.

This scenario is covered by the integration test `"should clean up stale series on label value change"`:
https://github.com/epam/kubernetes-kueue/blob/2b720f0d70b6420a7f012b9d4339be8549a1ab41/test/integration/singlecluster/controller/core/custom_labels_test.go#L169

and the unit test `TestUpdateClusterQueueLabelsUpdated`:
https://github.com/epam/kubernetes-kueue/blob/2b720f0d70b6420a7f012b9d4339be8549a1ab41/pkg/cache/queue/manager_test.go#L239

mimowo: > labelsUpdated is needed because custom label values come from each ClusterQueue’s labels/annotations and can change. When that happens, the old metric series must be cleared and re-reported with the new values.

Hm, but I still don't understand why we need to store the set of available label values. For example, we don't store the set of available label values for `priority_class`, and the set of values for "priority_class" is resolved at runtime, based on the classes created by an admin etc.

pkg/metrics/reassign_vectors.go
IrvingMg: I've eliminated `reassign_vectors.go` entirely by consolidating all metric initialization into a single `initMetricVectors(extraLabels)` function in `metrics.go`.

pkg/cache/queue/manager.go
IrvingMg: `priority_class` appears in counters, where having multiple series per CQ is expected, since different workloads have different priorities and all those series are valid.

Custom labels appear in gauges like `PendingWorkloads`. A CQ only has one `team` value at a time. If it changes from `alpha` to `beta`, the old gauge `PendingWorkloads{custom_team="alpha"} = 5` is never updated again. It persists indefinitely with stale data because it is a different Prometheus series. We need to explicitly delete it, which is what `StoreAndClear` does.

hack/tools/metricsdoc/main.go
```diff
@@ -243,15 +243,20 @@ func metricFromCall(expr ast.Expr) (Metric, bool) {
...
 		return labels
+	case *ast.CallExpr:
+		// Extract only the static labels from append(base, dynamic...) calls.
+		if ident, ok := v.Fun.(*ast.Ident); ok && ident.Name == "append" && len(v.Args) >= 1 {
+			return parseLabels(v.Args[0])
+		}
 	}
-	for _, elt := range cl.Elts {
-		labels = append(labels, stringLiteral(elt))
-	}
```
mimowo: Can you send a preparatory PR to offload it? It is already massive

pkg/metrics/custom_labels.go
```diff
@@ -0,0 +1,128 @@
...
+	if len(customLabelValues) == 0 {
+		return append(base, make([]string, len(customLabelNames))...)
+	}
+	return append(base, customLabelValues...)
+}
+
+// CustomLabelStore is a type-safe wrapper around SyncMap for storing custom
+// label values per object, used for stale series cleanup.
+type CustomLabelStore[K ~string] struct {
+	m *utilmaps.SyncMap[string, []string]
```
mimowo: what is the key here? What if two objects of different types have the same name, ie. LocalQueue and ClusterQueue - that is a common naming pattern,

mimowo: ah, I see you instantiate a separete store per predefined type, ok

pkg/metrics/custom_labels.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// Delete removes the stored custom label values for the given key.
+func (s CustomLabelStore[K]) Delete(key K) {
+	s.m.Delete(string(key))
+}
+
+var (
+	CQCustomLabels     = newCustomLabelStore[kueue.ClusterQueueReference]()
+	LQCustomLabels     = newCustomLabelStore[string]()
```
mimowo: why not using LocalQueue reference here?

pkg/metrics/custom_labels.go
```diff
@@ -0,0 +1,128 @@
...
+	"slices"
+
+	configapi "sigs.k8s.io/kueue/apis/config/v1beta2"
+	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+	utilmaps "sigs.k8s.io/kueue/pkg/util/maps"
+)
+
+// Set once via InitCustomLabels, then read-only.
+var customLabelNames []string
+var customLabelEntries []configapi.ControllerMetricsCustomLabel
```
mimowo: Can we avoid the global state? THis is problematic for MultiKueue integration tests, but also makes the code tricky to evolve.

pkg/cache/queue/manager.go
mimowo: Interesting, so this mechanism we could also use instead of the ResyncGuageMetrics we did to fix the ReplicaRole tracking? Because in this case also the replica_role must have at most one value. I'm wondering if the mechanism should be generic then for all guages.

Also, is it only useful / applied for guage metrics?

hack/tools/metricsdoc/main.go
IrvingMg: Opened PR: https://github.com/kubernetes-sigs/kueue/pull/9834

pkg/cache/scheduler/cache.go
```diff
@@ -564,6 +567,18 @@ func (c *Cache) ClusterQueueUsesAdmissionFairSharing(cqName kueue.ClusterQueueRe
...
+		c.RLock()
+		defer c.RUnlock()
+		cq := c.hm.ClusterQueue(newQ.Spec.ClusterQueue)
+		if cq != nil {
+			if lq := cq.localQueues[queue.Key(newQ)]; lq != nil {
+				lq.Lock()
+				if features.Enabled(features.CustomMetricLabels) {
+					lq.customMetricLabelValues = metrics.ExtractCustomLabelValues(newQ.Labels, newQ.Annotations)
+				}
+				lq.Unlock()
```
mimowo: The lock take/release should be certainly just undre the FG, but probably the entire green block of code here, right? There is also no need taking `c.RLock()` if the FG is disabled.

pkg/metrics/custom_labels.go
IrvingMg: Done. Changed to LocalQueueReference

pkg/cache/scheduler/cache.go
IrvingMg: Right, done.

pkg/metrics/custom_labels.go
IrvingMg: Done. I've added a CustomLabels struct.

pkg/cache/queue/manager.go
IrvingMg: Yes, they only apply to gauges, and both solve the same problem but at different scopes: `ResyncGaugeMetrics` does it for all objects at once, and `StoreAndClear` does it per object. 

But that’s a good point. If that’s not blocking, I can create an issue to explore how to commonize or align the patterns in a follow-up.

apis/config/v1beta1/configuration_conversion.go
```diff
@@ -92,6 +92,11 @@ func Convert_v1beta2_WaitForPodsReady_To_v1beta1_WaitForPodsReady(in *v1beta2.Wa
 	return nil
 }
 
+func Convert_v1beta2_ControllerMetrics_To_v1beta1_ControllerMetrics(in *v1beta2.ControllerMetrics, out *ControllerMetrics, s conversionapi.Scope) error {
+	// CustomLabels is dropped (not present in v1beta1)
```
mbobrovskyi: Can we add to v1beta1 as well?

pkg/metrics/custom_labels.go
```diff
@@ -0,0 +1,223 @@
...
+import (
+	"slices"
+
+	configapi "sigs.k8s.io/kueue/apis/config/v1beta2"
+	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+	utilmaps "sigs.k8s.io/kueue/pkg/util/maps"
+	utilqueue "sigs.k8s.io/kueue/pkg/util/queue"
+)
+
+func appendCustomLabels(base, customLabelValues []string) []string {
```
mbobrovskyi:
```suggestion
func appendCustomLabels(customLabelValues []string, baseLables ...string) []string {
```
I'm wondering if we can swap this arguments.

apis/config/v1beta1/configuration_conversion.go
IrvingMg: Done.

pkg/metrics/custom_labels.go
IrvingMg: I dropped the function entirely, as it was only doing a check `len() != 0` before the append and seemed unnecessary.

pkg/cache/queue/manager.go
IrvingMg: I've opened a new issue to explore consolidating this logic: https://github.com/kubernetes-sigs/kueue/issues/9854

mimowo: Yes, non-blocking, just waned to clarify. We will revisit how to fix the problem long term.

mimowo:
/lgtm
/approve
Thank you for implementing the important feature 👍

Issue #6675, opened 2025-08-26
Replace sort.Slice() with slices.Sort()


tenzen-y:
**What would you like to be cleaned**:

I would propose to replace `sort.Slice()` with `slices.Sort()` everywhere.

**Why is this needed**:

Both are implemented with the QuickSort base algorithm. But the new `slices.Sort()` function is faster than `sort.Slice()` in many situations as described in Go doc:

> Note: in many situations, the newer [slices.IsSortedFunc](https://pkg.go.dev/slices#IsSortedFunc) function is more ergonomic and runs faster.

https://pkg.go.dev/sort#Slice

PR #9781, opened 2026-03-10
chore: Use slices instead of sort

alexandear:
#### What type of PR is this?

/kind cleanup

Replace `sort.Sort` and `sort.Strings` with `slices.Sort`.

Enable `revive.use-slices-sort` and extend `forbidigo.pattern` to prevent regressions.

Updates #6675

In tests, I use `cmpopts.SortSlices`.

```release-note
NONE
```

### DIFF
--- a/cmd/kueueviz/backend/handlers/namespaces.go
+++ b/cmd/kueueviz/backend/handlers/namespaces.go
@@ -18,9 +18,9 @@ package handlers
 
 import (
 	"context"
-	"sort"
 
 	"github.com/gin-gonic/gin"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	kueueapi "sigs.k8s.io/kueue/apis/kueue/v1beta2"
 )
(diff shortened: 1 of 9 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -802,6 +802,7 @@ func (co *classicalIterator) pop() *entry {
 }
 
 func makeClassicalIterator(entries []entry, workloadOrdering workload.Ordering) *classicalIterator {
+	//nolint:forbidigo,revive // sort.Sort is more readable here
```
mimowo: might be, but i think it is preferred not to maintain exceptions. Maintaining the comment might be more text after all :)

cmd/kueueviz/backend/handlers/namespaces.go
mbobrovskyi: I'm wondering if we can use `sets.NewString()` here.

cmd/kueueviz/backend/handlers/namespaces.go
```diff
@@ -56,14 +57,7 @@ func (h *Handlers) fetchNamespaces(ctx context.Context) (any, error) {
...
 
-	// Convert namespace set to a slice of strings (just the names)
-	var namespaceNames []string
-	for namespaceName := range namespaceSet {
-		namespaceNames = append(namespaceNames, namespaceName)
-	}
-
-	// Sort namespaces alphabetically for consistent ordering
-	sort.Strings(namespaceNames)
+	namespaceNames := slices.Sorted(maps.Keys(namespaceSet))
```
mbobrovskyi: And now we can use:

```suggestion
	namespaceNames := namespaceSet.List()
```

cmd/kueueviz/backend/handlers/namespaces.go
mbobrovskyi: Or actually add it directly here:

```suggestion
		"namespaces": namespaceSet.List(namespaceSet),
```

pkg/controller/admissionchecks/multikueue/workload_test.go
```diff
@@ -20,7 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"sort"
+	stdslices "slices"
```
mbobrovskyi: We are trying to avoid std aliases.

pkg/controller/admissionchecks/multikueue/workload_test.go
mbobrovskyi:
```suggestion
	utilslices "sigs.k8s.io/kueue/pkg/util/slices"
```

Please rename this one.

PR #9791, opened 2026-03-10
TAS: improve performance of the addNode function.

### LINKED ISSUE #9782: TAS: improve performance of the addNode function

The addNode function is pretty simple https://github.com/kubernetes-sigs/kueue/blob/5fd2aa1291aa02429b2770a3dd474bf4455325af/pkg/cache/scheduler/tas_flavor_snapshot.go#L156-L177

Yet, the function shows up on the pprof graphs, see:
- https://github.com/kubernetes-sigs/kueue/issues/9337
- https://github.com/kubernetes-sigs/kueue/pull/9712#issuecomment-4023895532

I think we could easily optimize the case when `kubernetes.io/hostname` is the lowest level. This would help to reduce the time spent 2x based on : 

Image

**Why is this needed**:

To improve TAS performance for snapshot building.

mbobrovskyi:
#### What type of PR is this?
/kind cleanup
/area tas

TAS: improve performance of the addNode function.

Fixes #9782

```release-note
NONE
```

### DIFF
--- a/pkg/cache/scheduler/tas_flavor_snapshot.go
+++ b/pkg/cache/scheduler/tas_flavor_snapshot.go
@@ -131,6 +131,9 @@ type TASFlavorSnapshot struct {
 
 	// tolerations represents the list of tolerations defined for the resource flavor
 	tolerations []corev1.Toleration
+
+	// isLowestLevelNode indicates if kubernetes.io/hostname is the lowest topology level
+	isLowestLevelNode bool
 }
 
 func newTASFlavorSnapshot(log logr.Logger, topologyName kueue.TopologyReference,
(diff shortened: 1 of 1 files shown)


pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -155,9 +155,15 @@ func newTASFlavorSnapshot(log logr.Logger, topologyName kueue.TopologyReference,
 
 func (s *TASFlavorSnapshot) addNode(node *nodeInfo) utiltas.TopologyDomainID {
 	levelValues := utiltas.LevelValues(s.levelKeys, node.Labels)
-	domainID := utiltas.DomainID(levelValues)
+
+	var domainID utiltas.TopologyDomainID
```
sohankunkerkar, line 162: Thanks for this refactor. I was curious if you could share some benchmark numbers showing the improvement

mimowo: +1 on benchmarking, see the issue: https://github.com/kubernetes-sigs/kueue/issues/9782
The contribution from addNode shows up consistently on pprof charts:
- https://github.com/kubernetes-sigs/kueue/pull/9712#issuecomment-4023895532
- https://github.com/kubernetes-sigs/kueue/issues/9337

but indeed, it would be great to repeat the experiment to also show **after**

mbobrovskyi: https://github.com/kubernetes-sigs/kueue/pull/9791#issuecomment-4036493743

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -141,32 +144,47 @@ func newTASFlavorSnapshot(log logr.Logger, topologyName kueue.TopologyReference,
...
 }
 
 func (s *TASFlavorSnapshot) addNode(node *nodeInfo) utiltas.TopologyDomainID {
-	levelValues := utiltas.LevelValues(s.levelKeys, node.Labels)
-	domainID := utiltas.DomainID(levelValues)
-	if s.isLowestLevelNode() {
-		domainID = utiltas.DomainID(levelValues[len(levelValues)-1:])
+	var levelValues []string
+	var domainID utiltas.TopologyDomainID
+	var foundLeave bool
```
mimowo:
```suggestion
	var leafFound bool
```
nit

mbobrovskyi: Thank you!

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -131,6 +131,9 @@ type TASFlavorSnapshot struct {
 
 	// tolerations represents the list of tolerations defined for the resource flavor
 	tolerations []corev1.Toleration
+
+	// hostnameIsLowestLevel indicates if kubernetes.io/hostname is the lowest topology level
+	hostnameIsLowestLevel bool
```
mimowo:
```suggestion
	// isLowestLevelNode indicates if kubernetes.io/hostname is the lowest topology level
	isLowestLevelNode bool
```
I think it makes the core more readbable. We care about Nodes, because only they have taints, labels etc. 

Yes, we detect that by the `kubernetes.io/hostname` currently, but this should probably be configurable, because the label can be overriten by kubelet. We don't support it currently.

mbobrovskyi: Done

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -141,32 +144,47 @@ func newTASFlavorSnapshot(log logr.Logger, topologyName kueue.TopologyReference,
...
+		_, foundLeave = s.leaves[domainID]
+		// Only compute levelValues when we actually need to create a new leafDomain.
+		if !foundLeave {
+			levelValues = utiltas.LevelValues(s.levelKeys, node.Labels)
+		}
+	} else {
+		// Compute full level values and domain ID.
+		levelValues = utiltas.LevelValues(s.levelKeys, node.Labels)
+		domainID = utiltas.DomainID(levelValues)
+		_, foundLeave = s.leaves[domainID]
```
mimowo, line 178: it is not that important, but here we could also compute `levelValues`, only if creating a new node, righ? I l wondering computation of the `levelValues` can be delayed in both cases.

mbobrovskyi: Yeah, but we need `levelValues` to calculate `domainID`, which we return and use to add capacity.

mimowo: I see, makes sense, I missed that

Issue #9730, opened 2026-03-06
Scheduling: improve logging for start/end of each phase of scheduling cycle


mimowo:
**What would you like to be added**:

 I would like to log boundaries between each phase of scheduling cycle to make it easy to infer from logs how long they took, in particular:
- obtaining heads
- nomination
- snapshoting (can be long for TAS)
- workload processing 

Finally, in defer it would be nice to log the end. This way we can also easily tell how long scheduler was sleeping between cycles.

**Why is this needed**:

There are no clear logs indicating boundaries between phases of each scheduling cycle takes.
Estimation of the times is possible by side effects or internal logs at higher levels, but we should make it convenient.

The motivation was investigation for https://github.com/kubernetes-sigs/kueue/issues/9715

Issue #9715, opened 2026-03-06
☂️ Solve the problem of non-admissible workloads in BestEffortFIFO despite enough capacity


mimowo:
**What would you like to be added**:

This is an umbrella issue as fixing the problem with using BestEffortFIFO at scale, because currently the issues are spread.

Consider you have 10k workloads in a BestEffortCQ. Often they would target different flavors. If one flavor is capped at full capacity (say GPU), then the inadmissible list it used to put the "NoFit" workloads aside to "drill" the CQ searching for admissible workloads in the consecutive scheduling cycles.

Now, when there are 10k worklaods (or more) the admissible workload might be at position 1k. This means that to schedule the workload Kueue needs to "drill". If requeue happens in the meanwhile, the drilling starts over again. In the result Kueue scheduler ends up processing the same set of unschedulable workloads over and over again, not making progress, while admissible workload (say CPU) are waiting.

Originally Kueue would requeue all workloads any time some workload is finished. This was very unpredictible, and on a large system it would be all the time (too frequent). With the recent improvement https://github.com/kubernetes-sigs/kueue/issues/8095 to rate limit the requeue of inadmissible workloads we improved the situation, but the problem exists, especially for TAS where each scheduling cycle is long.

For TAS each scheduling cycle takes a long time, say 500ms. So, essentially, Kueue can "drill" only 2-3 workloads within the second. Good luck scheduling something at position 1k :).

To fix the issue we need improvements in multiple areas:
- predictable rate limiting for the requeue events (done already by 1s)
- performance of the single scheduling cycle (especially for TAS)
- smaller gaps between scheduling cycles
- batch drilling by skipping of workloads with equivalent scheduling directives 

Sub-issues:
- https://github.com/kubernetes-sigs/kueue/issues/8095
- https://github.com/kubernetes-sigs/kueue/issues/9337
- https://github.com/kubernetes-sigs/kueue/issues/9716
- https://github.com/kubernetes-sigs/kueue/issues/9694
- https://github.com/kubernetes-sigs/kueue/issues/9730
- https://github.com/kubernetes-sigs/kueue/issues/9782 
- https://github.com/kubernetes-sigs/kueue/issues/9799

**Why is this needed**:

Currently this is the primary problem users report when using BestEffortFIFO at scale.

PR #9795, opened 2026-03-10
Add V(2) logging for scheduling cycle phase boundaries

sohankunkerkar:
Log the duration of each scheduling phase (heads wait, snapshot, nomination, workload processing) at V(2) to make it easy to identify which phase is slow. A deferred log at cycle end also makes inter-cycle  sleep time visible from log timestamps.

#### What type of PR is this?
/kind feature

Fixes https://github.com/kubernetes-sigs/kueue/issues/9730

```release-note
Observability: Add scheduler logs for the scheduling cycle phase boundaries.
```

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -209,6 +209,11 @@ func (s *Scheduler) schedule(ctx context.Context) wait.SpeedSignal {
 	s.schedulingCycle++
 	log := roletracker.WithReplicaRole(ctrl.LoggerFrom(ctx), s.roleTracker).WithValues("schedulingCycle", s.schedulingCycle)
 	ctx = ctrl.LoggerInto(ctx, log)
+	cycleStartTime := s.clock.Now()
+	log.V(2).Info("Scheduling cycle starts")
+	defer func() {
+		log.V(2).Info("Scheduling cycle complete", "duration", s.clock.Since(cycleStartTime))
+	}()
 
 	// 1. Get the heads from the queues, including their desired clusterQueue.
 	// This operation blocks while the queues are empty.
@@ -218,6 +223,7 @@ func (s *Scheduler) schedule(ctx context.Context) wait.SpeedSignal {
 		return wait.KeepGoing
 	}
 	startTime := s.clock.Now()
+	log.V(2).Info("Obtained heads", "headCount", len(headWorkloads), "waitDuration", startTime.Sub(cycleStartTime))
 
 	// 2. Take a snapshot of the cache.
 	var snapshotOpts []schdcache.SnapshotOption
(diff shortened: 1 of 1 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -209,6 +209,11 @@ func (s *Scheduler) schedule(ctx context.Context) wait.SpeedSignal {
 	s.schedulingCycle++
 	log := roletracker.WithReplicaRole(ctrl.LoggerFrom(ctx), s.roleTracker).WithValues("schedulingCycle", s.schedulingCycle)
 	ctx = ctrl.LoggerInto(ctx, log)
+	cycleStartTime := s.clock.Now()
+	log.V(2).Info("Scheduling cycle starts")
```
mimowo, line 213: I'm wondering what is the desired log level here, 2 or 3. I would say it is probably 3, otoh we already have log level 2 for `log.V(2).Info("Attempting to schedule workload")`, and the "moving workloads to inadmissible queues". 

My intuition is to put at v2 the key scheduler phases, like in this PR (so no change needed), but put to v3 "Attempting to schedule workload" and the moving of workloads (small follow up).

I would like to hear what @tenzen-y and @gabesaba think here.

sohankunkerkar: I agreed on V(2) for phase boundaries. Happy to create the follow-up moving per-workload logs ("Attempting to schedule workload", "Workload re-queued", bulk-move) to V(3). That's a behavioral change at the default `--zap-log-level=2` though.

sohankunkerkar: I checked the integration test artifact sizes per @mimowo's request:                                                                                                                               
              
  | Job                  | Baseline  | This PR | Delta |                                                                                                           
  |----------------------|---------------------|---------|-------|
  | integration-baseline | 50.3 MB             | 54.1 MB | +7.5% |                                                                                                           
  | integration-extended | 38.9 MB             | 42.2 MB | +8.5% |

  ~8% growth from 5 extra V(2) lines per cycle. Seems reasonable for the debugging value.

gabesaba: I imagine these logs will be quite valuable and V(2) seems justified. And I also agree that we can clean those other logs as a follow-up

Defining what each logging level means would be a valuable contribution for both developers and users of Kueue. I created an issue to track it here: https://github.com/kubernetes-sigs/kueue/issues/9801

mimowo: Right, using  https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md  as a basis it sepcifies v2 as a reasonable level for scheduler logging, I think especially for phases. I'm less sure about logging particular workloads per level. Here the existing definitions become vague, and judgement based. In my judgment scheduler logic for top-level phases is v2, but per workload is v3 rather.

tenzen-y: I think that this probably should be (3) since in no problem situation, these logs are too verbose.
As we received feedback in TAS logs, the default logs (V2) should be minimized as much as possible. 

These logs seem helpful to identify the performance issue, but this is for debug. I can imagine the situations for performance issue debugging in the following step:

1. Admin finds the performance regressions by Workload admission Prometheus metrics
2. Admin increases the log level to 3 or higher.
3. Admin identifies which step consumes time by logs.
4. Admin revert log level to 2

mimowo: I'm ok to put it as v3, but in that case we should probably put as v3 also other logs: ("Attempting to schedule workload", "Workload re-queued", bulk-move) to V(3)?

tenzen-y: I think so too.
Cluster admin can change the log level based on their situation.

mimowo: Well, yes and no. The problem is that OnCallers on cloud are sometimes asked to understand some issue on the previous day based on the logs. The issue is already gone, because there was a spike of 10k jobs. Now the jobs are no longer there - processed. Still, the user complains processing took too long, but we have no logs for the time window to analyze, wdyt?

mimowo: Such scenarios happen, actually such OnCall episode triggered our work on the umbrella problem https://github.com/kubernetes-sigs/kueue/issues/9715

tenzen-y: The primary point is whether or not we should default this login in all clusters.
In some clusters (very large clusters), I think they should always use higher log levels like 6 because such large clusters always encounter problems.

But, if you think that in almost all clusters, these logs should be defaulted, I don't have any objections for V(2) because I don't have a strong preference here.

As a side note, my Kueue always uses V(8) log level.

mimowo: If you dont have objections I would leave it as v2 then. I think knowing where scheduler spends time will be very useful for debugging issues post factum based on logs. 

Im cautious not to make logs too verbose, but I think in this case it is acceptable.

tenzen-y: Alright, sounds good to me.
Let's continue the standardized log-level discussions in https://github.com/kubernetes-sigs/kueue/issues/9801.

PR #9796, opened 2026-03-10
cleanup: use Job instead of JobSet in TAS rank ordering case

### LINKED ISSUE #9284: Simplify test case for rank-based ordering to use Job rather than JobSet

Use Job rather than JobSet for the test case https://github.com/kubernetes-sigs/kueue/pull/9211/changes#diff-95ef2b4868044a4fbcc96fcda6df858f3b79548d2e4548fafcd9280114928dfcR1685

**Why is this needed**:

- To clearly show JobSet is not relevant in this scenario.
- JobSet CRD is not installed for the test so it could be a bit misleading (not possible outside of integration test)

0xlen:
#### What type of PR is this?
/kind cleanup
/area tas
/area testing

This cleanup removes the JobSet-specific subgroup labels/config from the test, since the scenario (`should fallback to greedy assignment when replacement pod conflicts with already running pod`) only relies on rank ordering via `batch.kubernetes.io/job-completion-index`.

It does not create an actual Job object, but it aligns the test with Job semantics and avoids implying that JobSet is relevant here.

Fixes #9284

```release-note
NONE
```

### DIFF
--- a/test/integration/singlecluster/tas/tas_test.go
+++ b/test/integration/singlecluster/tas/tas_test.go
@@ -35,7 +35,6 @@ import (
 	autoscaling "k8s.io/autoscaler/cluster-autoscaler/apis/provisioningrequest/autoscaling.x-k8s.io/v1"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


sohankunkerkar (approved):
/lgtm

PR #9808, opened 2026-03-11
Add support for quotaCheckStrategy

### LINKED ISSUE #7513: Provide a way to specify in the configuration what resources kueue will manage

Add a configuration field that will be an allow list of resources that kueue can manage.

This is similar to excludeResourcePrefixes but we want something like includeResourcePrefixes.

This can be a list of allowed resources that Kueue can manage and anything else would not be counted in admission of the workload (everything else is excluded from quota counts).

**Why is this needed**:

I was discussing #5800 with @amy  and she mentioned a separate problem with the excludeResourcePrefixes.

It seems that most people are really using Kueue to manage a resources like cpu, memory and GPUs while any field that gets added to the pod spec has to keep being added to this excludeResourcePrefixes.

If someone adopts a new webhook or a controller adds a new resource to requests/limits then kueue will not allow anything to be scheduled.

So what happens is that admins keep adding fields to this list to prevent workloads from being scheduled.

**Completion requirements**:

- [x] Design doc
- [x] API change
- [x] Docs update

MaysaMacedo:
#### What type of PR is this?
/kind feature

Introduces a new configuration API which allows to restrict the
quota check to only resources which are declared in the `ClusterQueue`.
Other resources requested by the Pod are skipped from the quota checks.

Fixes #7513

This PR was developed with the assistance of Claude.

```release-note
Add a new configuration named `quotaCheckStrategy` to be able to restrict the quota check to only resources which are declared in the `ClusterQueue`.
```


pkg/config/validation.go
```diff
@@ -535,6 +551,12 @@ func ValidateFeatureGates(featureGateCLI string, featureGateMap map[string]bool)
 		}
 	}
 
+	if !strings.Contains(featureGateCLI, string(features.QuotaCheckStrategy)+"=true") && !featureGateMap[string(features.QuotaCheckStrategy)] {
```
sohankunkerkar: The KEP [says](https://github.com/kubernetes-sigs/kueue/tree/main/keps/7513-quota-check-strategy#feature-gate) when the gate is disabled, `quotaCheckStrategy` should be ignored (fallback to BlockUndeclared), but this rejects the config and prevents Kueue from starting. Remove this block and instead add `features.Enabled(features.QuotaCheckStrategy)` to `quotaCheckStrategy()` in main.go, so the field is silently ignored when the gate is off. Also strings.Contains on the CLI string is fragile.

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -637,10 +644,18 @@ func (a *FlavorAssigner) assignFlavors(log logr.Logger, counts []int32) Assignme
...
-			// Skip zero-quantity requests for resources not defined in the ClusterQueue (#8079).
-			if quantity == 0 && a.cq.RGByResource(resName) == nil {
-				continue
+			// Skip zero-quantity requests for resources not defined in the ClusterQueue (#8079) or
+			// If quotaCheckStrategy is IgnoreUndeclared, skip resources not declared in the ClusterQueue.
+			if a.cq.RGByResource(resName) == nil {
+				if quantity == 0 {
+					continue
+				}
+				if a.quotaCheckStrategy == configapi.QuotaCheckIgnoreUndeclared {
```
sohankunkerkar: I guess undeclared resources are correctly skipped from flavor assignment here, but check how AFS entry penalties are computed. The penalty calculation still sees the full unfiltered resource set from the workload spec, not the filtered admission. This means undeclared resources influence admission ordering even though quota ignores them.

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -210,6 +211,10 @@ func (a *Assignment) TotalRequestsFor(wl *workload.Info) resources.FlavorResourc
 			if q == 0 {
 				continue
 			}
+			if a.PodSets[i].Flavors[res] == nil {
+				log.Info("Skipping usage count for resource with undefined flavor", "res", res)
```
sohankunkerkar: Use `log.V(3).Info`

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -637,10 +644,18 @@ func (a *FlavorAssigner) assignFlavors(log logr.Logger, counts []int32) Assignme
...
-			if quantity == 0 && a.cq.RGByResource(resName) == nil {
-				continue
+			// Skip zero-quantity requests for resources not defined in the ClusterQueue (#8079) or
+			// If quotaCheckStrategy is IgnoreUndeclared, skip resources not declared in the ClusterQueue.
+			if a.cq.RGByResource(resName) == nil {
+				if quantity == 0 {
+					continue
+				}
+				if a.quotaCheckStrategy == configapi.QuotaCheckIgnoreUndeclared {
+					log.Info("Skipping resource not declared in the ClusterQueue", "res", resName)
```
sohankunkerkar: same: https://github.com/kubernetes-sigs/kueue/pull/9808/changes#r2984447414

test/e2e/sequential/baseline/quota_check_strategy_test.go
```diff
@@ -0,0 +1,229 @@
...
+
+			localQueue = utiltestingapi.MakeLocalQueue("", ns.Name).
+				GeneratedName("test-lq-ignoreundeclared-").
+				ClusterQueue(clusterQueue.Name).
+				Obj()
+			util.CreateLocalQueuesAndWaitForActive(ctx, k8sClient, localQueue)
+		})
+
+		ginkgo.AfterEach(func() {
+			gomega.Expect(util.DeleteNamespace(ctx, k8sClient, ns)).To(gomega.Succeed())
```
sohankunkerkar, line 131: You might need to add `metricsReaderClusterRoleBinding` and `curlpod` are created in kueue-system, not in ns. AfterEach only deletes ns, so these leak across test runs.

sohankunkerkar (commented):
Thanks! I think it would be helpful if you add an integration test for this.

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -210,6 +211,10 @@ func (a *Assignment) TotalRequestsFor(wl *workload.Info) resources.FlavorResourc
 			if q == 0 {
 				continue
 			}
+			if a.PodSets[i].Flavors[res] == nil {
+				log.Info("Skipping usage count for resource with undefined flavor", "res", res)
+				continue
+			}
```
mimowo: This code should be under feature gate and the check for quota strategy?

pkg/config/validation.go
```diff
@@ -535,6 +551,12 @@ func ValidateFeatureGates(featureGateCLI string, featureGateMap map[string]bool)
...
+		if cfg.Resources != nil && cfg.Resources.QuotaCheckStrategy != nil {
+			allErrs = append(allErrs, field.Invalid(resourceQuotaCheckStrategyPath, featureGateMap[string(features.QuotaCheckStrategy)], "quotaCheckStrategy is not allowed when feature gate QuotaCheckStrategy is disabled"))
+		}
+	}
```
mimowo: Move this validation to later stage when the feature gates are loaded so that you can just check if the FG is enabled or not without this strings.Contains. Note that we also have two ways of setting feature gates - cli and configuration.

mimowo: Actually, I would probably allow to set the ignoreUndeclared just ignore iff FG is disabled.

pkg/scheduler/flavorassigner/flavorassigner.go
mimowo: Check also the FG here

pkg/scheduler/flavorassigner/flavorassigner.go
MaysaMacedo: @mimowo I was actually blocking the startup of Kueue at the validation stage, if the featureGate is not set, but the config is. That is why I didn't add it here. We prefer the featuregate check to be actually done at the stage the config is used? Maybe that makes more sense to allow other features to run.

mimowo: This code runs on flavor assignment stage, so I see no relationship with the startup.

MaysaMacedo: It would block Kueue from starting when validating the featuregates here https://github.com/MaysaMacedo/kueue/blob/add-support-quota-check-strategy/cmd/kueue/main.go#L155-L158

pkg/scheduler/flavorassigner/flavorassigner.go
MaysaMacedo: Ok, I will check this scenario.

MaysaMacedo: FYI I could check that all resources are being used and not just the filtered ones, so I worked on the fix. Now, I'm trying to figure out how to test admission ordering depending on the entry penalties.

pkg/scheduler/flavorassigner/flavorassigner.go
mimowo: Maybe I wasn't clear, what I'm saying is that the new code should be under the feature gate

```
if a.PodSets[i].Flavors[res] == nil {
  log.Info("Skipping usage count for resource with undefined flavor", "res", res)
   continue
}
```

mimowo: > It would block Kueue from starting when validating the featuregates here https://github.com/MaysaMacedo/kueue/blob/add-support-quota-check-strategy/cmd/kueue/main.go#L155-L158

This code fails when feature gates are set both by CLI and configMap, I don't see how this is related to the code

MaysaMacedo: This is fixed now:
https://github.com/kubernetes-sigs/kueue/pull/9808/changes#diff-0870ec2c0ea330a7d087a6ef599905bf9b7164a0dd889d1b00ba17c31a1bfd05R214

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -225,6 +226,10 @@ func (a *Assignment) TotalRequestsFor(wl *workload.Info) resources.FlavorResourc
 			if q == 0 {
 				continue
 			}
+			if features.Enabled(features.QuotaCheckStrategy) && a.PodSets[i].Flavors[res] == nil {
```
sohankunkerkar: Right now this skips nil flavors whenever the feature gate is on, regardless of which strategy is set. That means under `BlockUndeclared` a nil flavor gets quietly ignored instead of being caught as a bug. Add `&& a.quotaCheckStrategy == configapi.QuotaCheckIgnoreUndeclared` so it only skips when that's actually the intended behavior.

pkg/controller/core/workload_controller.go
```diff
@@ -1168,7 +1177,13 @@ func (r *WorkloadReconciler) Generic(e event.TypedGenericEvent[*kueue.Workload])
 
 func (r *WorkloadReconciler) updateAfsConsumedUsage(log logr.Logger, wl *kueue.Workload) {
 	lqKey := qutil.KeyFromWorkload(wl)
-	penalty := afs.CalculateEntryPenalty(workload.NewInfo(wl).SumTotalRequests(), r.admissionFSConfig)
+	totalRequests := workload.NewInfo(wl).SumTotalRequests()
+	if features.Enabled(features.QuotaCheckStrategy) && r.quotaCheckStrategy == config.QuotaCheckIgnoreUndeclared {
```
sohankunkerkar: Same: https://github.com/kubernetes-sigs/kueue/pull/9808/changes#r3035133993

pkg/scheduler/scheduler.go
```diff
@@ -963,10 +973,22 @@ const (
...
+	return covered
+}
+
 func (s *Scheduler) updateEntryPenalty(log logr.Logger, e *entry, op usageOp) {
 	lqKey := utilqueue.NewLocalQueueReference(e.Obj.Namespace, e.Obj.Spec.QueueName)
 	lqObjRef := klog.KRef(e.Obj.Namespace, string(e.Obj.Spec.QueueName))
-	penalty := afs.CalculateEntryPenalty(e.SumTotalRequests(), s.admissionFairSharing)
+	totalRequests := e.SumTotalRequests()
+	if features.Enabled(features.QuotaCheckStrategy) && s.quotaCheckStrategy == config.QuotaCheckIgnoreUndeclared {
+		totalRequests = utilresource.FilterByNames(totalRequests, allCoveredResources(e.clusterQueueSnapshot.ResourceGroups))
```
sohankunkerkar: hmm...The penalty is computed independently at push and sub time using the CQ's current covered resources. If the CQ changes in between, the two won't cancel out cleanly. A safer approach would be to store the exact penalty that was pushed and subtract that stored value later, instead of recomputing from mutable CQ state. Not blocking for Alpha but worth considering for Beta.

site/content/en/docs/tasks/manage/administer_cluster_quotas.md
```diff
@@ -465,3 +465,21 @@ The Workload obtains an effective resource requests for quota purposes:
         example.com/gpu-memory: 30Gi
         example.com/credits: 61
 ```
+
+## Change quota check strategy for admission
+
+By default, administrators must specify all resources required by workloads in the ClusterQueues `.spec.resourceGroups[*]`.
+If you want relax that requirement and make resources declared on the workload, but not declared on the ClusterQueues to be skipped from quota management and admission process,
```
sohankunkerkar:
```suggestion
If you want to relax that requirement and make resources declared on the workload, but not declared on the ClusterQueues to be skipped from quota management and admission process, 
```

site/content/en/docs/tasks/manage/administer_cluster_quotas.md
```diff
@@ -465,3 +465,21 @@ The Workload obtains an effective resource requests for quota purposes:
         example.com/gpu-memory: 30Gi
         example.com/credits: 61
 ```
+
+## Change quota check strategy for admission
+
+By default, administrators must specify all resources required by workloads in the ClusterQueues `.spec.resourceGroups[*]`.
```
sohankunkerkar: `s/ClusterQueues/ClusterQueue's/g`

test/e2e/customconfigs/quota_check_strategy_test.go
```diff
@@ -0,0 +1,231 @@
...
+var _ = ginkgo.Describe("QuotaCheckStrategy", ginkgo.Label("feature:quota-check-strategy"), ginkgo.Ordered, ginkgo.ContinueOnFailure, func() {
+	var (
+		ns             *corev1.Namespace
+		resourceFlavor *kueue.ResourceFlavor
+		clusterQueue   *kueue.ClusterQueue
+	)
+
+	ginkgo.BeforeAll(func() {
+		util.UpdateKueueConfigurationAndRestart(ctx, k8sClient, defaultKueueCfg, kindClusterName, func(cfg *configapi.Configuration) {
+			cfg.FeatureGates = map[string]bool{string(features.QuotaCheckStrategy): true}
```
sohankunkerkar: I think this will replace the other feature gates set in the default config. Should this be additive instead?

pkg/config/validation_test.go
```diff
@@ -1005,6 +1005,22 @@ func TestValidate(t *testing.T) {
 				},
 			},
 		},
+		"quotaCheckStrategy with value ignoreundeclared not allowed with excludeResourcePrefixes": {
```
sohankunkerkar: It would be good if you could also cover the following scenarios here:
- IgnoreUndeclared without prefixes: should pass
- BlockUndeclared with prefixes: should pass
- Invalid enum value: should fail with NotSupported

pkg/scheduler/scheduler.go
MaysaMacedo: So, your thinking is that when the second CalculateEntryPenalty happens we could just drop the additional resource that might exist on the AfsEntryPenalties, given the output from the first round of SumTotalRequests() would be different from the second given in the second the workload is already admitted?

pkg/scheduler/flavorassigner/flavorassigner.go
MaysaMacedo: I skip the flavor assignment only when the `IgnoreUndeclared` is configured here https://github.com/kubernetes-sigs/kueue/pull/9808/changes#diff-0870ec2c0ea330a7d087a6ef599905bf9b7164a0dd889d1b00ba17c31a1bfd05R668
That is why I thought it wouldn't be necessary to add this check for `QuotaCheckIgnoreUndeclared`. Let me know if you still would like me to add it and I will need to add a quotaCheckStrategy to the Assignment object, so I can check its value here.

sohankunkerkar: Yup, you are right there. But it's still worth adding as defensive code. If someone later adds a new caller to               `TotalRequestsFor` or changes the assignment flow, the nil flavor would silently undercount preemption usage instead of being caught.

pkg/scheduler/scheduler.go
sohankunkerkar: Not exactly. `SumTotalRequests()` on the sub side already excludes undeclared resources because the workload is admitted and NewInfo reads from the admission status, not the pod spec. The real risk is the `FilterByNames` call that re-filters that result against the current CQ, so if the CQ changed, you get a mismatch. Since the admission status already handles the filtering, you can just remove the FilterByNames block from `updateAfsConsumedUsage` IMO.

kannon92:
Based on https://github.com/kubernetes-sigs/kueue/pull/9808#issuecomment-4120578353 it seems that this PR is not WIP anymore.

/retitle Add support for quotaCheckStrategy

mimowo:
@MaysaMacedo what is the status here? 

I'm also wondering if we should re-consider putting this quotaCheckStrategy at the CQ level. Maybe it makes sense for alpha to skip the API update, and just rely on the FG. Then we would introduce the API, either global or CQ-scoped before Beta, wdyt? This would also simplify that PR, and allow us to validate the idea without API cc @tenzen-y @kannon92 @sohankunkerkar , wdyt?

MaysaMacedo:
@mimowo I still need to address @sohankunkerkar 's review points on the PR.
About the API, just my 2 cents, given it's already implemented in the PR, maybe we could keep it?

kannon92:
> I'm also wondering if we should re-consider putting this quotaCheckStrategy at the CQ level. Maybe it makes sense for alpha to skip the API update, and just rely on the FG. Then we would introduce the API, either global or CQ-scoped before Beta, wdyt? This would also simplify that PR, and allow us to validate the idea without API cc @tenzen-y @kannon92 @sohankunkerkar , wdyt?

cc @amy 

I discussed the original feature with Amy. Based on what we thought I think a configuration makes more sense than at the CQ level. Our goal is to limit the operational burden of admins to remember to update `excludeResourcePrefixes` so I worry putting this information at the CQ level still makes it an operational issue to remember to add this information to each CQ.

Putting it as an alpha feature gate makes it difficult for us to offer this to customers. I'd like alpha to match what we eventually want to offer customers so I also prefer the PR as is. (minus the pending updates to address reviews).

We also agreed on the design on the [KEP](https://github.com/kubernetes-sigs/kueue/tree/main/keps/7513-quota-check-strategy).

mimowo:
> I discussed the original feature with Amy. Based on what we thought I think a configuration makes more sense than at the CQ level. Our goal is to limit the operational burden of admins to remember to update excludeResourcePrefixes so I worry putting this information at the CQ level still makes it an operational issue to remember to add this information to each CQ.

Good point, making it global makes it easier for admins. OTOH, maybe for some clusters a CQ requires another setting.
Still, moving this to CQ level could be another extension of the feature later on, rather than replacement.

> Putting it as an alpha feature gate makes it difficult for us to offer this to customers. I'd like alpha to match what we eventually want to offer customers so I also prefer the PR as is. (minus the pending updates to address reviews).

That makes sense, I just want to clarify that this is going to be behind an alpha feature for the first release anyway. 

> We also agreed on the design on the KEP.

Sure, I'm just exploring your view on that. It is not unheard of that some designs change during the implementation phase. Still, if you think global is better for your use case I'm all for it, we can revisit the CQ-level override for that setting later on.

pkg/features/kube_features.go
```diff
@@ -543,6 +549,9 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 	FastQuotaReleaseInPodIntegration: {
 		{Version: version.MustParse("0.17"), Default: false, PreRelease: featuregate.Alpha},
 	},
+	QuotaCheckStrategy: {
+		{Version: version.MustParse("0.17"), Default: false, PreRelease: featuregate.Alpha},
```
kannon92:
```suggestion
		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Alpha},
```

MaysaMacedo:
The PR is ready for another look.

MaysaMacedo:
@tenzen-y @mimowo This PR is ready for a review whenever you can. :pray:

pkg/config/validation.go
```diff
@@ -87,6 +88,32 @@ func validate(c *configapi.Configuration, scheme *runtime.Scheme) field.ErrorLis
 	allErrs = append(allErrs, validateTLS(c)...)
 	allErrs = append(allErrs, validateVisibilityServer(c)...)
 	allErrs = append(allErrs, validateCustomLabels(c)...)
+	allErrs = append(allErrs, validateQuotaCheckStrategy(c)...)
+	return allErrs
+}
+
+func validateQuotaCheckStrategy(c *configapi.Configuration) field.ErrorList {
```
kannon92, line 95: We should check feature gate here.

pkg/util/resource/resource.go
```diff
@@ -100,6 +101,18 @@ func MulByFloat(q corev1.ResourceList, f float64) corev1.ResourceList {
 	return ret
 }
 
+// FilterByNames returns a new ResourceList containing only resources whose names
+// are in the allowed set.
+func FilterByNames(requests corev1.ResourceList, allowed sets.Set[corev1.ResourceName]) corev1.ResourceList {
```
kannon92, line 106: nit: If we are exporting this we should have some unit tests.

kannon92 (changes requested):
At a high level the code looks good to me.

We should validate the feature gate in the configuration also.

I also think we need some integration tests verifying this feature gate on/off in the scheduler.

mimowo:
cc @sohankunkerkar could you also give it a pass?

pkg/config/validation.go
MaysaMacedo: @kannon92 I removed the validation I had here for feature gates because @sohankunkerkar brought up an interesting point https://github.com/kubernetes-sigs/kueue/pull/9808#discussion_r2984396041
But I just noticed I'm missing checking the featuregates at quotaCheckStrategy() as he recommended, I will add it.

pkg/config/validation.go
MaysaMacedo: @kannon92 FYI

pkg/config/validation.go
kannon92: Sounds good. Yea the main point of feature gates is to not have any impacts of this feature if you turn off the gate. In this case, you are still validating quota check strategy even if its disabled.

MaysaMacedo:
/hold
I'll unhold when  I address all the comments from the review

pkg/util/resource/resource.go
MaysaMacedo: I updated it to not export given it's not needed anymore.

pkg/config/validation.go
MaysaMacedo: added the check.

cmd/kueue/main.go
```diff
@@ -171,6 +171,12 @@ func main() {
 		}
 	}
 
+	// Validates the configuration after it has been loaded and feature gates have been set.
```
kannon92, line 174: I don't really see how this is related to this feature.

Can you do this as a separate PR?

pkg/config/validation_test.go
```diff
@@ -1005,11 +1006,76 @@ func TestValidate(t *testing.T) {
...
```
kannon92:
```suggestion
		"quotaCheckStrategy with value ignoreUndeclared not allowed with excludeResourcePrefixes": {
```

site/content/en/docs/tasks/manage/administer_cluster_quotas.md
```diff
@@ -465,3 +465,21 @@ The Workload obtains an effective resource requests for quota purposes:
         example.com/gpu-memory: 30Gi
         example.com/credits: 61
 ```
+
+## Change quota check strategy for admission
```
kannon92, line 469: Can you also mention that this an alpha level feature?

test/e2e/sequential/baseline/quota_check_strategy_test.go
```diff
@@ -0,0 +1,234 @@
...
+			util.ExpectObjectToBeDeletedWithTimeout(ctx, k8sClient, highWPC, true, util.MediumTimeout)
+			util.ExpectObjectToBeDeletedWithTimeout(ctx, k8sClient, lowWPC, true, util.MediumTimeout)
+		})
+
+		ginkgo.It("should preempt workloads that have lower priority", func() {
```
kannon92: Could you include that this is testing undeclared resource in the It block?

That way we can see what feature this is testing in the junit.

test/e2e/sequential/baseline/quota_check_strategy_test.go
```diff
@@ -0,0 +1,234 @@
+/*
+Copyright The Kubernetes Authors.
```
kannon92, line 2: @mimowo 

I'm not entirely sure what defines baseline versus extended.

Should we move alpha features to extended?

kannon92:
Minor items.

/lgtm

/assign @sohankunkerkar @mimowo

cmd/kueue/main.go
MaysaMacedo: This is needed in this PR because when validating the quotacheckstrategy configs we also verify if the feature gate is enabled. And it's only possible to verify if the feature gate is enabled when they are actually set, so I moved the config validation.

kannon92: Is this the bug behind https://github.com/kubernetes-sigs/kueue/issues/10866?

I'm just wondering if we need to cherry-pick this change separate from this feature?

test/e2e/sequential/baseline/quota_check_strategy_test.go
MaysaMacedo: I'm also not sure. I saw another feature which is alpha and is on baseline package, the `FailureRecoveryPolicy`.

cmd/kueue/main.go
MaysaMacedo: No, the bug you linked is for validation of feature gates.
The work I did here is for validation of configMap, before this PR there was no check of feature gates within the validate of config Map, but now there is and that is why we started needing it.

kannon92 (commented):
/lgtm

test/e2e/sequential/baseline/quota_check_strategy_test.go
sohankunkerkar: IIRC, the baseline/extended split is about external operator dependencies (JobSet, LWS, Spark etc.), not feature maturity. This test only needs core Kueue + batch Jobs. If future tests need integration with external operators, those would go in extended.

mimowo: Yes, as @sohankunkerkar mentioned they are currently split on the use of external integrations, not feature maturity. I think this feature could be tested using built-in integrations like Jobs or Pods, so it fits baseline.

cmd/kueue/main.go
sohankunkerkar: nit(non-blocking): The `config.Validate()` move breaks Load()'s previous invariant of returning a validated config. Even with one caller today, I'd prefer either keeping Load() semantics unchanged with a small post-gate validation pass or splitting the contract change into a separate PR.

pkg/config/validation.go
```diff
@@ -87,6 +89,35 @@ func validate(c *configapi.Configuration, scheme *runtime.Scheme) field.ErrorLis
...
+	var allErrs field.ErrorList
+	if !features.Enabled(features.QuotaCheckStrategy) {
+		return allErrs
+	}
+	if c.Resources != nil && c.Resources.QuotaCheckStrategy != nil {
+		strategy := *c.Resources.QuotaCheckStrategy
+		if len(c.Resources.ExcludeResourcePrefixes) > 0 && strategy == configapi.QuotaCheckIgnoreUndeclared {
+			allErrs = append(allErrs, field.Invalid(
```
sohankunkerkar, line 104: nit: `field.Invalid()` here passes the pointer `c.Resources.QuotaCheckStrategy` as the bad value, while `field.NotSupported()` below correctly uses the dereferenced strategy. Should be strategy in both places? otherwise the error message renders a pointer instead of the enum string.

sohankunkerkar (commented):
Two minor nits but overall LGTM. Thanks!

cmd/kueue/main.go
sohankunkerkar: We discussed this offline and agreed to stick with the current approach to avoid unnecessary complications. With this PR, Load() no longer performs validation, so the config is loaded but not validated at that stage. Validation now happens later in main(), after the feature gates are set, so [this](https://github.com/kubernetes-sigs/kueue/pull/9808/changes#diff-64a1e0bd9f87bdadd12179b8e48ace63c6aa3ace3f7694375a8a0dd0e566e25dR657) change makes it clear.

sohankunkerkar (commented):
/lgtm

test/integration/singlecluster/scheduler/quotacheckstrategy/quota_check_strategy_test.go
```diff
@@ -0,0 +1,195 @@
...
+				Request("example.com/gpu", "100").
+				Obj()
+			util.MustCreate(ctx, k8sClient, wl)
+
+			wlKey := types.NamespacedName{Name: wl.Name, Namespace: ns.Name}
+
+			gomega.Consistently(func(g gomega.Gomega) {
+				g.Expect(k8sClient.Get(ctx, wlKey, wl)).To(gomega.Succeed())
+				g.Expect(workload.IsAdmitted(wl)).To(gomega.BeFalse())
+			}, util.ConsistentDuration, util.Interval).Should(gomega.Succeed())
```
mimowo: nit, but use util.ExpectPendingWorkloadsMetric to avoid waiting time on consistency. We know that when the metric is bumped, then the workload is already considered by scheduler.

mimowo (commented):
LGTM, one nit from me  👍

mimowo:
/lgtm
/approve
Thank you 👍

Issue #9799, opened 2026-03-11
Introduce temporary feature gates for mitigating BestEffortFIFO issues at scale


mimowo:
**What would you like to be added**:

I propose introducing two feature gates, temporarily. We will drop them at 0.20 if no one needs them as knobs / configuration.

- SchedulerLongRequeueInterval - increases the 1s base interval to 10s
- SchedulerTimestampPreemptionBuffer - introduces 5min buffer so that workloads which with the scheduling timestamps within 5min buffer don't preempt each other on the LowerPriorityOrNewer basis

Maybe long term instead of increasing the base interval we could introduce exponential backoff to 16s or so. The feature gates are meant to collect feedback. They are part of the umbrella issue: https://github.com/kubernetes-sigs/kueue/issues/9715. Basically, we introduced a number of improvements, and it is hard to tell if they are enough. The feature gates would help us to experiment with alternative values if the improvements "don't help". If the feature gates are not needed to be used we will drop them.

**Why is this needed**:

- 1s requeue rate limiting parameter is arbitrary, on very large environments this may still not be enough
- the higher the requeue parameter the more likely the workloads will get admitted "out of order" risking preemptions

mwielgus:
SchedulerTimestampPreemptionBuffer - What about hiding it entirely behind a FeatureGate for 0.17 and 0.16 cherry-pick? We will evaluate the feature in the upcoming ~2 months and two things will happen:
* We will get a confirmation that this is needed and the thing will get a proper API.
* We will drop it as not needed.

mimowo:
> We have continued to discuss the preemption loop (preempt a->b->c->a...). I'd like to consider the comprehensive approach both for classic preemption and fair sharing.

This feature gate is not meant to cover preemption loops. It is just to protect against too many preemptions in case of "out of order admission", when `LowerOrNewerEqualPriority` is used. The idea is to say "NewerWithBuffer", so that we don't preempt if the difference in creation timestamps is just 1min, for example.

Issue #9713, opened 2026-03-06
Pod groups when Single CQ Failed Pod can be replaced in group


mimowo:
**Which test is flaking?**:

Pod groups when Single CQ Failed Pod can be replaced in group

**Link to failed CI job or steps to reproduce locally**:

https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/periodic-kueue-test-e2e-release-0-16-1-35/2029796010150072320

**Failure message or logs**:
```

End To End Suite: kindest/node:v1.35.0: [It] Pod groups when Single CQ Failed Pod can be replaced in group [area:singlecluster, feature:pod] expand_less	1m16s
{Timed out after 45.001s.
The function passed to Eventually failed at /home/prow/go/src/kubernetes-sigs/kueue/test/e2e/singlecluster/pod_test.go:230 with:
Expected
    <v1.PodPhase>: Running
to equal
    <v1.PodPhase>: Failed failed [FAILED] Timed out after 45.001s.
The function passed to Eventually failed at /home/prow/go/src/kubernetes-sigs/kueue/test/e2e/singlecluster/pod_test.go:230 with:
Expected
    <v1.PodPhase>: Running
to equal
    <v1.PodPhase>: Failed
In [It] at: /home/prow/go/src/kubernetes-sigs/kueue/test/e2e/singlecluster/pod_test.go:231 @ 03/06/26 06:09:06.067
}
```

**Anything else we need to know?**:

Same build as https://github.com/kubernetes-sigs/kueue/issues/9714

PR #9812, opened 2026-03-11
Introduce SchedulerLongRequeueInterval feature gate.

mbobrovskyi:
#### What type of PR is this?
/kind feature

Introduce SchedulerLongRequeueInterval feature gate.

Part of #9799

```release-note
Scheduling: Add the alpha SchedulerLongRequeueInterval feature gate (disabled by default) to increase the 
inadmissible workload requeue interval from 1s to 10s. This may help to mitigate, on large environments with 
many pending workloads, issues with frequent re-queues that prevent the scheduler from reaching schedulable 
workloads deeper in the queue and result in constant re-evaluation of the same top workloads.
```

### DIFF
--- a/pkg/cache/queue/inadmissible_workloads.go
+++ b/pkg/cache/queue/inadmissible_workloads.go
@@ -30,12 +30,33 @@ import (
 
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
 	"sigs.k8s.io/kueue/pkg/cache/hierarchy"
+	"sigs.k8s.io/kueue/pkg/features"
 	"sigs.k8s.io/kueue/pkg/workload"
 )
 
 const (
-	RequeueBatchPeriodProd = 1 * time.Second
+	requeueBatchPeriod     = 1 * time.Second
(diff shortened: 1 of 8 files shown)


pkg/cache/queue/inadmissible_workloads.go
```diff
@@ -30,13 +30,34 @@ import (
...
+	requeueBatchPeriod     = 1 * time.Second
+	requeueLongBatchPeriod = 10 * time.Second
 )
 
+func getRequeueBatchPeriod() time.Duration {
+	if !features.Enabled(features.SchedulerLongRequeueInterval) {
+		return requeueLongBatchPeriod
+	}
+	return requeueBatchPeriod
+}
```
mimowo, line 47: Could we add some unit tests for the feature gate, I think a unit test for this helper would be enough probably

mbobrovskyi: Done

mbobrovskyi:
/retest

Know flake https://github.com/kubernetes-sigs/kueue/issues/9713.

PR #9814, opened 2026-03-11
Temporarily restore the FlavorFungibilityImplicitPreferenceDefault feature gate to allow upgrades 

### LINKED ISSUE #9790: Temporarily restore the FlavorFungibilityImplicitPreferenceDefault feature gate to allow upgrades

I would like to temporarily restore the  FlavorFungibilityImplicitPreferenceDefault feature gate to allow easier upgrade to 0.16 without changing in modelling.

**Why is this needed**:

To allow users an upgrade to 0.16 asap, and only later to drop the feature gate.

vladikkuzn:
#### What type of PR is this?
/kind cleanup

To allow users an upgrade to 0.16 asap, and only later to drop the feature gate.

Fixes #9790

```release-note
NONE
```

### DIFF
--- a/pkg/features/kube_features.go
+++ b/pkg/features/kube_features.go
@@ -165,6 +165,15 @@ const (
 	ManagedJobsNamespaceSelectorAlwaysRespected featuregate.Feature = "ManagedJobsNamespaceSelectorAlwaysRespected"
 
 	// owner: @pajakd
+	// kep: https://github.com/kubernetes-sigs/kueue/tree/main/keps/582-preempt-based-on-flavor-order
+	//
+	// In flavor fungibility, infer the preference between borrowing and preemption
+	// from the legacy policies when .spec.flavorFungibility.preference is unset.
+	//
+	// Deprecated: planned to be removed in 0.18.
+	FlavorFungibilityImplicitPreferenceDefault featuregate.Feature = "FlavorFungibilityImplicitPreferenceDefault"
+
+	// owner: @pajakd
 	// kep: https://github.com/kubernetes-sigs/kueue/tree/main/keps/2724-topology-aware-scheduling
 	//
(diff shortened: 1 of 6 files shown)


pkg/features/kube_features.go
```diff
@@ -380,6 +388,11 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 		{Version: version.MustParse("0.13"), Default: false, PreRelease: featuregate.Alpha},
 		{Version: version.MustParse("0.15"), Default: true, PreRelease: featuregate.Beta},
 	},
+	FlavorFungibilityImplicitPreferenceDefault: {
+		{Version: version.MustParse("0.13"), Default: false, PreRelease: featuregate.Alpha},
+		{Version: version.MustParse("0.15"), Default: true, PreRelease: featuregate.Beta},
+		{Version: version.MustParse("0.16"), Default: true, PreRelease: featuregate.Deprecated},
```
vladikkuzn: idk if I should update it to 0.17

mimowo: We can keep as is.

pkg/features/kube_features.go
```diff
@@ -380,6 +388,11 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 		{Version: version.MustParse("0.13"), Default: false, PreRelease: featuregate.Alpha},
 		{Version: version.MustParse("0.15"), Default: true, PreRelease: featuregate.Beta},
 	},
+	FlavorFungibilityImplicitPreferenceDefault: {
+		{Version: version.MustParse("0.13"), Default: false, PreRelease: featuregate.Alpha},
+		{Version: version.MustParse("0.15"), Default: true, PreRelease: featuregate.Beta},
```
mimowo: It was never enabled by default (never beta). I think we should drop this line.

Issue #7539, opened 2025-11-05
Add Prometheus metrics for Cohort to report resource usage, quota, workload counts


varunsyal:
**What would you like to be added**:
Add metrics for cohort similar to clusterqueue:
- kueue_cohort_resource_usage (similar to kueue_cluster_queue_resource_usage)
- kueue_cohort_resource_reservation (similar to kueue_cluster_queue_resource_usage)
- kueue_cohort_nominal_quota (similar to kueue_cluster_queue_nominal_quota)
- kueue_cohort_borrowing_limit (similar to kueue_cluster_queue_borrowing_limit)
- kueue_cohort_admitted_workloads_total (similar to kueue_admitted_workloads_total with clusterqueue label)
- kueue_cohort_admitted_active_workloads (similar to kueue_admitted_active_workloads with clusterqueue label)

**Why is this needed**:
This is needed to report usage, quota and workload counts at the cohort level and not just at the clusterqueue level.

**Completion requirements**:

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] API change
- [ x] Docs update

The artifacts should be linked in subsequent comments.

mimowo:
cc @mwysokin @amy

mimowo:
/priority important-soon
Since we have a user interested in that

mimowo:
@varunsyal the existing metrics, like `cluster_queue_resource_usage` already have a label for "cohort". So IIUC the intention for this proposal is to have the usage aggregated for deeper cohort hierarchies, right?

tenzen-y:
> [@varunsyal](https://github.com/varunsyal) the existing metrics, like `cluster_queue_resource_usage` already have a label for "cohort". So IIUC the intention for this proposal is to have the usage aggregated for deeper cohort hierarchies, right?

As another usecase (problem), we want to track how many Cohort resources (nominalQuota) are consumed.
Currently, we don't have a Cohort resource-specific usage because nominalQuotas defined in the Cohort resource are shared resources for ClusterQueue. Thus, there is no way to determine how many Cohort-defined nominalQuota consumed in child ClusterQueues.

mimowo:
> As another usecase (problem), we want to track how many Cohort resources (nominalQuota) are consumed.

For sure, I think this could be answered with `kueue_cohort_resource_usage` which will aggregate the usage for the subtree.

mimowo:
/assign @mszadkow 
as we already synced on slack

[...]

PR #9347, opened 2026-02-18
Provide a metric for cohort borrowing limit

mszadkow:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
Extend metrics to cohort level.

Add metrics for borrowing limit.

#### Which issue(s) this PR fixes:

Relates to #7539

#### Special notes for your reviewer:
Rebased on https://github.com/kubernetes-sigs/kueue/pull/9132

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #9833, opened 2026-03-12
Provide a metric for cohort resource reservations

mszadkow:
#### What type of PR is this?
/kind feature

Add metrics for cohort's resource reservations.

Relates to #7539

```release-note
Observability:  Introduce the cohort resource reservations metric.
```


pkg/cache/scheduler/cache.go
```diff
@@ -133,6 +133,20 @@ type Cache struct {
 	roleTracker *roletracker.RoleTracker
 }
 
+func (c *Cache) GetCohortNameForWorkload(wl *kueue.Workload) kueue.CohortReference {
```
pajakd: perhaps this should be called `GetDirectCohortNameForWorkload`, to make it clear what does it return for multi-level hierarchies?

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
```diff
@@ -425,5 +443,187 @@ var _ = ginkgo.Describe("Cohorts", func() {
...
+
+		ginkgo.It("reports CohortSubtreeResourceUsage across child-parent propagation, reduced spill, release, and topology recompute", func() {
+			var (
+				wlCh1a *kueue.Workload
+				wlCh1b *kueue.Workload
+				wlCh1c *kueue.Workload
+				wlCh2  *kueue.Workload
+			)
+
+			ginkgo.By("Create root and children cohorts with explicit quotas/lending limits to reduce spillage", func() {
```
pajakd: I'm not familiar with term "spill/spillage". In resource_node.go https://github.com/kubernetes-sigs/kueue/blob/baf8a94160fe65077c17a1387c57055c6309ac01/pkg/cache/scheduler/resource_node.go#L95-L99 the capacity is "stored" at the parent. Is "spill/spillage" only used here? Please verify.

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
```diff
@@ -425,5 +443,187 @@ var _ = ginkgo.Describe("Cohorts", func() {
...
+				// ch2 local quota before spilling to parent is 6-2=4 CPU.
+				createCohort(utiltestingapi.MakeCohort("ch2").
+					Parent("root").
+					ResourceGroup(
+						withLending(kueue.ResourceFlavorReference(defaultFlavor.Name), "6", "2"),
+					).Obj())
+
+				util.ExpectCohortSubtreeResourceUsageGaugeMetricCleaned("ch1", defaultFlavor.Name, corev1.ResourceCPU.String())
+				util.ExpectCohortSubtreeResourceUsageGaugeMetricCleaned("ch2", defaultFlavor.Name, corev1.ResourceCPU.String())
+				util.ExpectCohortSubtreeResourceUsageGaugeMetricCleaned("root", defaultFlavor.Name, corev1.ResourceCPU.String())
```
pajakd: we don't check `SubtreeQuotaGaugeMetric` -- is it intentional?

pkg/cache/scheduler/cohort_metrics.go
```diff
@@ -73,21 +74,38 @@ func (c *Cache) collectCohortMetricPoints(cohortName kueue.CohortReference, simu
 	}
 
 	removedQuota := ch.resourceNode.SubtreeQuota
+	removedUsage := ch.resourceNode.Usage
+
 	var points []cohortMetricPoint
 	for ancestor := range ch.PathSelfToRoot() {
 		quotas := ancestor.resourceNode.SubtreeQuota
+		usages := ancestor.resourceNode.Usage
```
pajakd: Perhaps a dumb question but `resourceNode.Usage` is the usage "visible" in this node (exceeding the nominal quotas, restricted by lending limits). The objective of this metric isn't to track the total usage? Looking at this comment https://github.com/kubernetes-sigs/kueue/issues/7539#issuecomment-3819010176 it seems like we want to surface the usage including the nominalQuota consumption (which might not be included in resourceNode.Usage).

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
mszadkow: Yes, it's the same, I might accidentally introduce the new word (it was more convincing metaphor to me), I will make sure it's "storing" or "using".

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
mszadkow: Yes, we check it once, when all the objects are present.
This usage check could also land one step below, maybe it has more sense to keep it in one step to check initial state.

pkg/cache/scheduler/cohort_metrics.go
mszadkow: No dumb questions.
I think you are correct, I have missed that, due to focusing on nominal quota first.
Thanks!
I will correct that

pkg/cache/scheduler/cache_test.go
```diff
@@ -3753,3 +3753,64 @@ func TestGetWorkloadFromCache(t *testing.T) {
 		})
 	}
 }
+
+func TestGetCohortNameForWorkload(t *testing.T) {
```
vladikkuzn: NIT:
```suggestion
func TestGetDirectCohortNameForWorkload(t *testing.T) {
```
And update the error message:

```
			if diff := cmp.Diff(tc.wantCohort, got); diff != "" {
				t.Errorf("GetDirectCohortNameForWorkload() mismatch (-want,+got):\n%s", diff)
```

pkg/metrics/metrics_test.go
```diff
@@ -298,9 +298,25 @@ func TestCohortMetrics(t *testing.T) {
...
+	ReportCohortSubtreeResourceUsage("cohort", "flavor", "res", 5, leaderTracker)
+	expectFilteredMetricsCount(t, CohortSubtreeResourceUsage, 1, "cohort", "cohort", "replica_role", "leader")
+
+	ReportCohortSubtreeResourceUsage("cohort", "flavor", "res", 3, followerTracker)
+	expectFilteredMetricsCount(t, CohortSubtreeResourceUsage, 1, "cohort", "cohort", "replica_role", "follower")
+
+	ReportCohortSubtreeResourceUsage("cohort_two", "flavor", "res", 3, leaderTracker)
+	expectFilteredMetricsCount(t, CohortSubtreeResourceUsage, 1, "cohort", "cohort_two", "replica_role", "leader")
+
+	ClearCohortSubtreeResourceUsage("cohort", "", "")
```
vladikkuzn: What about cohort?
NIT:
```suggestion
	ClearCohortSubtreeResourceUsage("cohort", "", "")
	expectFilteredMetricsCount(t, CohortSubtreeResourceUsage, 0, "cohort", "cohort")
```

pkg/cache/scheduler/cache_test.go
```diff
@@ -3753,3 +3753,64 @@ func TestGetWorkloadFromCache(t *testing.T) {
 		})
 	}
 }
+
+func TestGetDirectCohortNameForWorkload(t *testing.T) {
+	ctx, _ := utiltesting.ContextWithLog(t)
```
mbobrovskyi: Please move it inside t.Run().

pkg/controller/core/workload_controller.go
```diff
@@ -1452,3 +1453,23 @@ func (h *draEventHandler) Generic(ctx context.Context, e event.TypedGenericEvent
 func (r *WorkloadReconciler) GetDRAReconcileChannel() chan<- event.TypedGenericEvent[*kueue.Workload] {
 	return r.draReconcileChannel
 }
+
+func (r *WorkloadReconciler) addOrUpdateWorkload(log logr.Logger, wl *kueue.Workload) bool {
+	updated := r.cache.AddOrUpdateWorkload(log, wl)
+	if wl != nil && workload.HasQuotaReservation(wl) {
+		r.cache.RecordCohortMetrics(log, r.cache.GetDirectCohortNameForWorkload(wl))
```
mbobrovskyi: Can we move this inside AddOrUpdateWorkload()?

pkg/controller/core/workload_controller.go
```diff
@@ -1452,3 +1453,23 @@ func (h *draEventHandler) Generic(ctx context.Context, e event.TypedGenericEvent
...
+}
+
+func (r *WorkloadReconciler) deleteWorkload(log logr.Logger, wlKey workload.Reference, oldWl *kueue.Workload) error {
+	wl := r.cache.GetWorkloadFromCache(wlKey)
+	cohortName := r.cache.GetDirectCohortNameForWorkload(wl)
+	if cohortName == "" {
+		cohortName = r.cache.GetDirectCohortNameForWorkload(oldWl)
+	}
+	err := r.cache.DeleteWorkload(log, wlKey)
+	r.cache.RecordCohortMetrics(log, cohortName)
```
mbobrovskyi: And this inside r.cache.DeleteWorkload()?

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
```diff
@@ -106,8 +122,8 @@ var _ = ginkgo.Describe("Cohorts", func() {
 		})
 
 		ginkgo.It("follows reporting correct metrics", func() {
-			ginkgo.By("Creating initial ClusterQueue cqa, with no cohort", func() {
-				createQueue(utiltestingapi.MakeClusterQueue("cqa").
+			ginkgo.By("Creating initial ClusterQueue cq1, with no cohort", func() {
```
mbobrovskyi: Why this renaming?

pkg/controller/core/workload_controller.go
mszadkow: We could, the thing that stopped me from doing this was to be consistent on which "level" we record the metrics.
I wanted it to be on the controller level, it was achieved by adding this wrapper.
I think it's a matter of preference

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
mszadkow: reverted

pkg/controller/core/workload_controller.go
mszadkow: I mean it's this type of metrics, the usage, because it is recorded with the same compute function -  RecordCohortMetrics
But sure it can be done the other way

mszadkow: Also another point we avoid double locking.
One lock in the `AddOrUpdateWorkload` and the other in the `RecordCohortMetrics -> collectCohortMetricPoints`.

But obviously this also can be overcome.

pkg/cache/scheduler/cache.go
```diff
@@ -141,6 +141,20 @@ type Cache struct {
 	customLabels *metrics.CustomLabels
 }
 
+func (c *Cache) GetDirectCohortNameForWorkload(wl *kueue.Workload) kueue.CohortReference {
```
```suggestion
func (c *Cache) CohortForWorkload(wl *kueue.Workload) kueue.CohortReference {
```
nit

pkg/cache/scheduler/cache.go
```diff
@@ -141,6 +141,20 @@ type Cache struct {
...
+	if wl == nil || wl.Status.Admission == nil {
+		return ""
+	}
+	c.RLock()
+	defer c.RUnlock()
+	cqName := wl.Status.Admission.ClusterQueue
+	cq := c.hm.ClusterQueue(cqName)
```
```suggestion
	cq := c.hm.ClusterQueue(cq := c.hm.ClusterQueue(wl.Status.Admission.ClusterQueue))
```

pkg/cache/scheduler/cohort_metrics.go
```diff
@@ -99,19 +130,38 @@ func (c *Cache) withCohortLogger(log logr.Logger, cohortName kueue.CohortReferen
...
 func (c *Cache) applyCohortMetricPoint(p cohortMetricPoint) {
-	if p.qty <= 0 {
-		metrics.ClearCohortSubtreeQuota(
-			p.cohortName,
-			string(p.flavorResource.Flavor),
-			string(p.flavorResource.Resource),
-		)
-		return
+	flavor := string(p.flavorResource.Flavor)
+	resource := string(p.flavorResource.Resource)
```
mbobrovskyi: I’m wondering if we can send it as is and convert it to a string inside the report function.

mszadkow: Yes, I have this thing done already in the other PR: https://github.com/kubernetes-sigs/kueue/pull/9347/changes#diff-4bc91a74bc47b8467cdcd25a7e3fe1651dbe44907f478377eca31b982f444f23
I will apply it here aswell

pkg/cache/scheduler/cohort_metrics.go
```diff
@@ -99,19 +130,38 @@ func (c *Cache) withCohortLogger(log logr.Logger, cohortName kueue.CohortReferen
...
+}
+
+// totalSubtreeUsage returns the sum of *actual* resource usage in the subtree rooted at given cohort
+func totalSubtreeUsage(ch *cohort) resources.FlavorResourceQuantities {
+	total := make(resources.FlavorResourceQuantities)
+
+	for _, cq := range ch.ChildCQs() {
+		for fr, qty := range cq.getResourceNode().Usage {
+			total[fr] += qty
+		}
```
mbobrovskyi: Looks like duplication, right? Can we make a helper?

pkg/cache/scheduler/cache.go
mszadkow: I think that this name reflects the intent, also mentioned here: https://github.com/kubernetes-sigs/kueue/pull/9833#discussion_r2938482239

pkg/cache/scheduler/cohort_metrics.go
```diff
@@ -99,19 +130,38 @@ func (c *Cache) withCohortLogger(log logr.Logger, cohortName kueue.CohortReferen
...
+	}
+
+	for _, child := range ch.ChildCohorts() {
+		childTotal := totalSubtreeUsage(child)
```
mbobrovskyi: I’m wondering if we can convert it to an iterative approach.

pkg/cache/scheduler/cohort_metrics.go
```diff
@@ -73,21 +74,46 @@ func (c *Cache) collectCohortMetricPoints(cohortName kueue.CohortReference, simu
...
+
+		var baselineUsage resources.FlavorResourceQuantities
 		if simulateRemoval {
 			quotas = removedQuota
+			usages = removedUsage
+			baselineUsage = getTotalUsage(ancestor)
 		}
 
-		for flr, qty := range quotas {
+		keys := make(map[resources.FlavorResource]struct{}, len(quotas)+len(usages))
```
mbobrovskyi: Maybe use set?

mbobrovskyi:
```suggestion
		keys := sets.New[resources.FlavorResource]()
		keys.Insert(slices.Collect(maps.Keys(quotas))...)
		keys.Insert(slices.Collect(maps.Keys(usages))...)
```

pkg/util/metrics/metrics.go
mbobrovskyi, line 1: Actually, maybe we can keep this helper in pkg/metrics? It looks like it is used only in this package, so there is no need to expose it.

mszadkow: Yes, we can do that too

pkg/controller/core/workload_controller.go
```diff
@@ -1452,3 +1453,23 @@ func (h *draEventHandler) Generic(ctx context.Context, e event.TypedGenericEvent
...
+	updated := r.cache.AddOrUpdateWorkload(log, wl)
+	if wl != nil && workload.HasQuotaReservation(wl) {
+		r.cache.RecordCohortMetrics(log, r.cache.GetDirectCohortNameForWorkload(wl))
+	}
+
+	return updated
+}
+
+func (r *WorkloadReconciler) deleteWorkload(log logr.Logger, wlKey workload.Reference, oldWl *kueue.Workload) error {
+	wl := r.cache.GetWorkloadFromCache(wlKey)
```
mbobrovskyi: Can we propagate newWl? In that case we don't need to get it from cache.

pkg/controller/core/workload_controller.go
```diff
@@ -1452,3 +1453,23 @@ func (h *draEventHandler) Generic(ctx context.Context, e event.TypedGenericEvent
...
+	cohortName := r.cache.GetDirectCohortNameForWorkload(wl)
+	if cohortName == "" {
+		cohortName = r.cache.GetDirectCohortNameForWorkload(oldWl)
```
mbobrovskyi: Is it possible this case?

mszadkow: you mean that the cohort name couldn't be retrieved from the "wlKey" workload but from the "oldWl"?
it should not be possible, correct

mszadkow: no need to pass OldWl or repeat `GetDirectCohortNameForWorkload`, because it's the same wl

pkg/controller/core/workload_controller.go
mszadkow: It make sense to pass only WlKey and get wl from the cache here.

if wl is not found in the cache, wl will be nil, then cohort will be empty,
then no metric will be recorded 
also `r.cache.DeleteWorkload` handles situations when no wl of wlKey is found

pkg/cache/scheduler/cohort_metrics.go
```diff
@@ -73,21 +78,42 @@ func (c *Cache) collectCohortMetricPoints(cohortName kueue.CohortReference, simu
 	}
 
 	removedQuota := ch.resourceNode.SubtreeQuota
+
+	// Cache total subtree usage calls during this run
+	usageCache := make(map[*cohort]resources.FlavorResourceQuantities)
+	getTotalUsage := func(co *cohort) resources.FlavorResourceQuantities {
+		return totalSubtreeUsageWithCache(co, usageCache)
+	}
```
mbobrovskyi: Do we need this function? It looks like we can use totalSubtreeUsageWithCache(co, usageCache).

pkg/controller/core/workload_controller.go
```diff
@@ -599,15 +599,16 @@ func (r *WorkloadReconciler) deleteWorkloadFromCaches(ctx context.Context, names
 
 	// If workload cached by scheduler points to a different queue:
 	// notify watchers to ensure all identified queues remain up-to-date.
-	if wlInSchedulerCache := r.cache.GetWorkloadFromCache(wlRef); workload.GetLocalQueue(wlInSchedulerCache) != workload.GetLocalQueue(wlInQueuesCache) {
+	wlInSchedulerCache := r.cache.GetWorkloadFromCache(wlRef)
+	if workload.GetLocalQueue(wlInSchedulerCache) != workload.GetLocalQueue(wlInQueuesCache) {
```
mbobrovskyi: Please revert.

pkg/controller/core/workload_controller.go
```diff
@@ -1452,3 +1453,20 @@ func (h *draEventHandler) Generic(ctx context.Context, e event.TypedGenericEvent
...
+	}
+
+	return updated
+}
+
+func (r *WorkloadReconciler) deleteWorkload(log logr.Logger, wlKey workload.Reference) error {
+	wl := r.cache.GetWorkloadFromCache(wlKey)
+	cohortName := r.cache.GetDirectCohortNameForWorkload(wl)
+	err := r.cache.DeleteWorkload(log, wlKey)
+	r.cache.RecordCohortMetrics(log, cohortName)
```
mbobrovskyi: Do we need to record on error?

pkg/controller/core/workload_controller.go
mbobrovskyi: OK. Can we propagate wlKey inside GetDirectCohortNameForWorkload() then?

mbobrovskyi: In that case we don't need to double lock/unlock cache.

pkg/controller/core/workload_controller.go
mszadkow: Yes, correct, we shouldn't confuse user if cq can not be found and wl deleted.

pkg/metrics/metrics.go
```diff
@@ -728,6 +733,14 @@ If the Cohort has a weight of zero and is borrowing, this will return NaN.`,
 			Help:      `Reports the cohort's nominal quota aggregated within the cohort's subtree. The values are reported per resource and flavor`,
 		}, []string{"cohort", "flavor", "resource", "replica_role"},
 	))
+
+	CohortSubtreeResourceUsage = trackGaugeVec(prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: constants.KueueName,
+			Name:      "cohort_subtree_resource_usage",
+			Help:      `Reports the cohort's resource usage aggregated within the cohort's subtree. The values are reported per resource and flavor`,
```
mimowo: Is this usage for QuotaReserved worklaods, or Admitted? Let's clarify please also double check which usage is counted by the analogous metrics for CLusterQueue.

mimowo: I see this is also not clarified in Help for `cluster_queue_resource_usage`, so please investigate by code and align. Please also make it clear in the description.

pkg/cache/scheduler/cache.go
```diff
@@ -141,6 +141,19 @@ type Cache struct {
 	customLabels *metrics.CustomLabels
 }
 
+func (c *Cache) GetDirectCohortNameForWorkload(wl *kueue.Workload) kueue.CohortReference {
+	if wl == nil || wl.Status.Admission == nil {
+		return ""
+	}
```
mimowo: This check makes the function confusing, because it only restricts the workload set to those with reservation (admission), the function name does not reflect that at all. 

So, we need to either:
1. make the function generic by lookup via LocalQueue -> ClusterQueue-> Cohort, I thought we already have a helper LocalQueue -> ClusterQueue
2. reflect in the name that it only works for Worklaods with amission

I would probably be leaning towards (1.), iirc we already have such helpers in cache, but let me know if I'm missing something.

pkg/cache/scheduler/cohort_metrics.go
```diff
@@ -73,21 +78,42 @@ func (c *Cache) collectCohortMetricPoints(cohortName kueue.CohortReference, simu
...
+	removedUsage := getTotalUsage(ch)
```
mimowo: What is removedUsage? I'm very confused by the variable name computed by `getTotalUsage`. Please explain what it is doing, let's add some comments to the functions and variables, because they are named very counter-intuitive imo - I cannot deduce by anemes what is going on.

pkg/controller/core/workload_controller.go
```diff
@@ -1452,3 +1453,20 @@ func (h *draEventHandler) Generic(ctx context.Context, e event.TypedGenericEvent
...
+	if wl != nil && workload.HasQuotaReservation(wl) {
+		r.cache.RecordCohortMetrics(log, r.cache.GetDirectCohortNameForWorkload(wl))
+	}
+
+	return updated
+}
+
+func (r *WorkloadReconciler) deleteWorkload(log logr.Logger, wlKey workload.Reference) error {
+	wl := r.cache.GetWorkloadFromCache(wlKey)
+	cohortName := r.cache.GetDirectCohortNameForWorkload(wl)
```
mimowo: Are we sure wl cannot be nil at this point?

pkg/metrics/metrics.go
mszadkow: Well, yes there is a difference here.
```
CohortSubtreeResourceUsage -> reservation-side aggregate (not admitted-only).
ClusterQueueResourceReservations -> reservation-side.
ClusterQueueResourceUsage -> admitted-only.
```
That means we need some correction here.
Let me add `kueue_cohort_resource_reservation` already here and I will do this:
```
CohortSubtreeResourceUsage -> aggregation of admitted-only childCQs.
CohortSubtreeResourceReservation -> reservation-side aggregate (not admitted-only).
ClusterQueueResourceReservations -> reservation-side.
ClusterQueueResourceUsage -> admitted-only.
```
And yes it will be reflected in the metrics' help.

wdyt?

pkg/controller/core/workload_controller.go
mszadkow: This goes like this:
if wl is nil, there will be empty cohortName.
for nil wl delete won't happen and for empty cohortName metrics recording won't happen.
This is ensured as first step in those functions, so I didn't want to repeat it here.
I hope it's ok

pkg/metrics/metrics.go
mimowo: This sgtm, let's align the metrics, but TBH I don't like ClusterQueueResourceUsage and CohortSubtreeResourceUsage to be admitted-only. 

I think it would be better to have `ClusterQueueResourceAdmitted` and `ClusterQueueResourceReservations`. Unfortunately we already have the confusing "Usage". 

I propose to revisit this later, for now I would just suggest to introduce the most helpful  `ClusterQueueResourceReservations`  as in `ClusterQueueResourceReservations`

Let me revisit the topic of `ClusterQueueResourceUsage` on another PR (less priority IMO).

mszadkow: Great, thanks for explaining!

pkg/controller/core/workload_controller.go
mszadkow: can be achieved, yes
anyway inside GetDirectCohortNameForWorkload we will have to call `GetWorkloadFromCache`, but thar's ok

pkg/metrics/metrics.go
mszadkow: Done, the metric is now called `CohortSubtreeResourceReservations` and description updated

pkg/cache/scheduler/cache.go
mszadkow: Updated

pkg/cache/scheduler/cohort_metrics.go
mszadkow: There is now more comments and names were updated

pkg/controller/core/workload_controller.go
```diff
@@ -1452,3 +1452,22 @@ func (h *draEventHandler) Generic(ctx context.Context, e event.TypedGenericEvent
 func (r *WorkloadReconciler) GetDRAReconcileChannel() chan<- event.TypedGenericEvent[*kueue.Workload] {
 	return r.draReconcileChannel
 }
+
+func (r *WorkloadReconciler) addOrUpdateWorkload(log logr.Logger, wl *kueue.Workload) bool {
+	updated := r.cache.AddOrUpdateWorkload(log, wl)
+	if wl != nil && workload.HasQuotaReservation(wl) {
+		r.cache.RecordCohortMetrics(log, r.cache.GetDirectCohortNameForWorkload(workload.Key(wl)))
```
mimowo: I'm actually concerned about this change because it calls the heavy cohort metrics computation on the event handlers which should be very fast. 

I don't understand why do we need to react to workload events to update resource usage for cohort metrics, I would assume we could do it from the places where we update the usage for "nominal" subtree quota, I'm really tripped this injection point

mimowo: If we really need to react to workloads (but again I have no idea why we would) maybe we could at least move the code to the Reconcile. However, let's first settle the main conundrum.

mszadkow: Ok, after talk with @mimowo.
We checked that `NotifyWorkloadUpdate` triggers Reconcile in ClusterQueue where computation is usually done.
Then no change in workload_controller is needed.

pkg/cache/scheduler/cache.go
```diff
@@ -141,6 +141,21 @@ type Cache struct {
 	customLabels *metrics.CustomLabels
 }
 
+func (c *Cache) GetDirectCohortNameForWorkload(wlKey workload.Reference) kueue.CohortReference {
```
mimowo: So this funcion is unused now, right? Let's drop

pkg/cache/scheduler/cohort_metrics.go
```diff
@@ -60,6 +65,9 @@ func (c *Cache) ClearCohortMetrics(log logr.Logger, cohortName kueue.CohortRefer
 	}
 }
 
+// collectCohortMetricPoints prepares metric values for the target cohort and each ancestor.
+// When simulateRemoval=true, it computes the remaining ancestor subtree values after
+// subtracting the target cohort subtree quotas and reservations.
```
mimowo: Is it the intention?

// collectCohortMetricPoints prepares subtree metric points for the target cohort
// and all cohorts on the path from the target to the root.
// When simulateRemoval=true, it computes post-removal subtree values by subtracting
// the target cohort's subtree contribution from each cohort's current subtree totals.
// This is used when clearing metrics so ancestor subtree gauges are updated too,
// rather than left with stale values after the target cohort is removed.

pkg/cache/scheduler/cohort_metrics.go
```diff
@@ -72,22 +80,43 @@ func (c *Cache) collectCohortMetricPoints(cohortName kueue.CohortReference, simu
...
 			}
 			points = append(points, cohortMetricPoint{
-				cohortName:     ancestor.Name,
-				flavorResource: flr,
-				qty:            qty,
+				cohortName:      ancestor.Name,
+				flavorResource:  fr,
+				quotaQty:        quotaQty,
+				reservationsQty: reservationsQty,
 			})
```
mimowo: The code is quite complex I would say. Maybe worth considering a helper to Subtract the usage maps represtented as ResourceFlavorQuantity maps to hide technicalities behind an abstraction and avoid some weird tricks with pinning / unpining the cohort variables.

I imagine the flow could be overall somthing like :

```golang
currentCohortQuota := ch.resourceNode.SubtreeQuota
for ancestor := range ch.PathSelfToRoot() {
  ancestorCohortQuota := ancestor.resourceNode.SubtreeQuota
  if simulateRemoval {
    currentCohortQuota.Subtract(ancestorCohortQuota) // or currentCohortQuotaWithRemove = currentCohortQuota.Sub(ancestorCohortQuota) if we need to also keep previous value
  }
```
This would iiuc allow us later to skip the `simulateRemoval` handling in the flow.

Let me know if I missing something or your encounter some obstacles.

pkg/cache/scheduler/cohort_metrics.go
```diff
@@ -72,22 +80,43 @@ func (c *Cache) collectCohortMetricPoints(cohortName kueue.CohortReference, simu
 		return nil
 	}
 
-	removedQuota := ch.resourceNode.SubtreeQuota
+	removedSubtreeQuota := ch.resourceNode.SubtreeQuota
```
mimowo: Let's also add some unit tests for the function which would allow quick verification of correctness with debugger without running the heavy machinery of integration tests.

pkg/cache/scheduler/cohort_metrics.go
mszadkow: Yes, that's the point of this function

pkg/cache/scheduler/cohort_metrics.go
mszadkow: Done

pkg/cache/scheduler/cohort_metrics.go
mszadkow: I have simplified it significantly

pkg/cache/scheduler/cohort_metrics_test.go
```diff
@@ -0,0 +1,468 @@
...
+							ResourceGroup(*utiltestingapi.MakeFlavorQuotas("flavor1").
+								Resource(corev1.ResourceCPU, "5").
+								Obj()).
+							Obj(),
+					},
+				)
+
+				// reservationMemo := subtreeReservationMemo{cache: make(map[*cohort]resources.FlavorResourceQuantities)}
+				// leftReservations := reservationMemo.total(cache.hm.Cohort(cohortLeft))[fr]
+				// rootReservations := reservationMemo.total(cache.hm.Cohort(cohortRoot))[fr]
```
pajakd: This seems to be unfinished.

mszadkow: thanks, I should have removed it

pkg/resources/resource.go
```diff
@@ -51,3 +51,12 @@ func (frq FlavorResourceQuantities) FlattenFlavors() Requests {
 	}
 	return result
 }
+
+// SubNonNegative returns max(frq-other, 0) for keys present in frq.
+func (frq FlavorResourceQuantities) SubNonNegative(other FlavorResourceQuantities) FlavorResourceQuantities {
```
mimowo: can we do just Sub() ? It is simpler semantic and later I see `quotaQty <= 0` so it seems we are handling negative values as 0 anyway

mszadkow: sure

PR #9841, opened 2026-03-12
Update DRA KEP to explain CEL handling

kannon92:
#### What type of PR is this?
/kind documentation

Describe design of #7298 

Fixes #

```release-note
NONE
```

### DIFF
--- a/keps/2941-DRA/README.md
+++ b/keps/2941-DRA/README.md
@@ -40,6 +40,7 @@ tags, and then generate with `hack/update-toc.sh`.
   - [Device Class Resolution and Conflict Prevention](#device-class-resolution-and-conflict-prevention)
     - [Device Class Mapping Uniqueness](#device-class-mapping-uniqueness)
   - [RBAC Requirements](#rbac-requirements)
+  - [CEL Expression Validation](#cel-expression-validation)
   - [Workloads](#workloads)
     - [DRA-Specific Workload Processing](#dra-specific-workload-processing)
     - [Workload Processing Flow](#workload-processing-flow)
@@ -266,6 +267,14 @@ administrator configures DeviceClasses with `extendedResourceName`.
   `extendedResourceName`, Kueue unifies quota** using the `deviceClassMappings` logical
   name as the quota key for both paths, preventing over-allocation.
 
+- CEL selectors in ResourceClaimTemplates are validated against cluster devices (ResourceSlices) at quota reservation
+  time on a best-effort basis. Workloads with CEL selectors that match fewer devices than requested are rejected
+  to prevent quota leaks. This validation uses the upstream DRA CEL compiler from [`k8s.io/dynamic-resource-allocation/cel`](https://github.com/kubernetes/dynamic-resource-allocation/tree/master/cel).
+  On the other hand, devices can be allocated between Kueue's check and scheduling, and new ResourceSlices published after
+  validation can make previously-unsatisfiable workloads satisfiable without triggering re-evaluation until the
+  next periodic `queueInadmissibleWorkloads` cycle. `WaitForPodsReady` serves as the safety net for cases where
+  the validation state diverges from actual device availability at scheduling time.
+
 ### Risks and Mitigations
 
 With DRAAdminAccess and DRAPrioritizedLists (both beta, default enabled in K8s 1.35), there is a
@@ -483,19 +492,70 @@ rules:
   resources: ["resourceclaimtemplates"]
   verbs: ["get", "list", "watch"]
 - apiGroups: ["resource.k8s.io"]
-  resources: ["deviceclasses"]
+  resources: ["resourceslices", "deviceclasses"]
   verbs: ["get", "list", "watch"]
 ```
 
 **Required Permissions:**
 - `resourceclaims`: Read access to validate ResourceClaim references (though not supported for quota)
 - `resourceclaimtemplates`: Read access to process ResourceClaimTemplates and extract device class information
-- `deviceclasses`: Read access to look up `extendedResourceName` for Extended Resources
+- `resourceslices`: Read access to list cluster devices for CEL selector validation
+- `deviceclasses`: Read access to resolve DeviceClass selectors for device pre-filtering during CEL evaluation
 
 **Security Considerations:**
 - Kueue only requires read permissions - no create, update, or delete access to DRA resources
 - Permissions are cluster-scoped to allow processing workloads across all namespaces
 - No elevated privileges required beyond standard Kueue controller permissions
+
+### CEL Expression Validation
+
+ResourceClaimTemplates may include CEL (Common Expression Language) selectors that constrain which devices
+can satisfy a request. Kueue validates these CEL selectors before admitting a workload to prevent quota from
+being consumed by workloads whose pods can never be scheduled.
+
+The validation has two stages:
+
+1. **CEL Compilation**: Each CEL expression in the request's selectors is compiled using the upstream DRA CEL
+   compiler ([`k8s.io/dynamic-resource-allocation/cel`](https://github.com/kubernetes/dynamic-resource-allocation/tree/master/cel)). This catches syntax errors, type errors, and other
+   compilation issues before quota reservation.
+
+2. **CEL Evaluation Against Cluster Devices**: Kueue lists all ResourceSlices in the cluster and evaluates
+   the compiled CEL selectors against actual devices. For each request:
+   - The DeviceClass is resolved and its selectors are compiled to pre-filter devices by class, avoiding
+     CEL evaluation against unrelated devices (e.g., NICs when requesting GPUs).
+   - The request's CEL selectors are evaluated against matching devices.
+   - If fewer devices match than the requested count, the workload is marked inadmissible with a descriptive
+     error indicating that no matching devices exist in the cluster, preventing quota consumption for
+     unsatisfiable requests. The `QuotaReserved` condition message clearly distinguishes between device-based inadmissibility
+     (e.g., "insufficient matching devices for CEL selector") and quota-based inadmissibility so that users
+     know whether they need to adjust their CEL selectors / request admin hardware changes, or wait for
+     quota to become available.
+
+**Example**: A ResourceClaimTemplate requesting 2 GPUs with `device.capacity["gpu.example.com"].memory.compareTo(quantity("80Gi")) >= 0`
+will be checked against actual devices in the cluster. If only 1 device matches, the workload is rejected
+before consuming quota.
+
+```yaml
+apiVersion: resource.k8s.io/v1
+kind: ResourceClaimTemplate
+metadata:
+  name: large-gpu
+spec:
+  spec:
+    devices:
+      requests:
+      - name: gpu
+        exactly:
+          deviceClassName: gpu.example.com
+          count: 2
+          selectors:
+          - cel:
+              expression: 'device.capacity["gpu.example.com"].memory.compareTo(quantity("80Gi")) >= 0'
+```
+
+**Dependencies**: This feature imports [`k8s.io/dynamic-resource-allocation/cel`](https://github.com/kubernetes/dynamic-resource-allocation/tree/master/cel) for CEL compilation and
+evaluation. This package provides the same CEL environment used by the Kubernetes scheduler for DRA device
+matching.
 
 ### Workloads
 
@@ -531,12 +591,16 @@ When a user submits a workload and DynamicResourceAllocation feature gate is ena
    - Read the ResourceClaimTemplate from the same namespace as the workload
    - Extract deviceClassName from each request in the template spec
    - Look up the corresponding resource name using the device class mappings from the Configuration API
-5. Resource Preprocessing:
+5. CEL Selector Validation: For requests with CEL selectors:
+   - Compile CEL expressions and reject workloads with invalid syntax
+   - Evaluate CEL selectors against actual devices from ResourceSlices, pre-filtering by DeviceClass
+   - Reject workloads where fewer devices match than requested count
+6. Resource Preprocessing:
    - Calculate total device count per device class across all containers and init containers
    - Map device classes to resource names using the configuration
    - Generate preprocessed resource requests for queue admission
-6. Queue Admission: Add the workload to the queue with preprocessed DRA resources.
-7. Status Update: Once admitted, the workload status reflects the assigned flavors and resource usage including DRA resources.
+7. Queue Reservation: Add the workload to the queue with preprocessed DRA resources.
+8. Status Update: Once the quota is reserved, the workload status reflects the assigned flavors and resource usage, including DRA resources.
 
 Note: The flow above applies to the ResourceClaimTemplate path (`DynamicResourceAllocation` gate).
 When the `DRAExtendedResources` gate is also enabled, workloads with extended resources in
@@ -918,6 +982,8 @@ using mock ResourceClaimTemplates and DeviceClasses to simulate DRA workloads. K
 - Queue integration: Testing workload admission with preprocessed DRA resources
 - Extended Resources: Testing extended resource detection, DeviceClass lookup, and resource translation
 - Late DeviceClass creation: Testing workload inadmissibility when DeviceClass does not exist
+- CEL validation: Testing CEL compilation errors, evaluation against ResourceSlice devices, and rejection
+  of workloads with unsatisfiable CEL selectors
 
 #### E2E Test
 
@@ -958,6 +1024,7 @@ Use existing dra-example-driver or Kubernetes test driver for e2e testing.
 - Extended Resources implementation: January 2026
 - Integration with Admission Fair Sharing: April 2026 — added integration tests and documentation
   confirming DRA logical resources work with existing `AdmissionFairSharing.ResourceWeights`
+- CEL expression validation support added: April 2026 by @kannon92
 
 **Key Design Evolution:**
 - **Original Design**: Standalone DynamicResourceAllocationConfig CRD with runtime ambiguity resolution
--- a/keps/2941-DRA/README.md
+++ b/keps/2941-DRA/README.md
@@ -41,6 +41,7 @@ tags, and then generate with `hack/update-toc.sh`.
     - [Device Class Mapping Uniqueness](#device-class-mapping-uniqueness)
   - [RBAC Requirements](#rbac-requirements)
   - [CEL Expression Validation](#cel-expression-validation)
+    - [Performance Implications](#performance-implications)
   - [Workloads](#workloads)
     - [DRA-Specific Workload Processing](#dra-specific-workload-processing)
     - [Workload Processing Flow](#workload-processing-flow)
@@ -557,6 +558,13 @@ spec:
 evaluation. This package provides the same CEL environment used by the Kubernetes scheduler for DRA device
 matching.
 
+#### Performance Implications
+
+In clusters with large number of ResourceSlices, it may be necessary to index the slices based on DeviceNames.
+This work will be deferred to beta.
+
+Its not entirely clear if this is a performance bottleneck at this time due to the number of ResourceSlices being small.
+
 ### Workloads
 
 #### DRA-Specific Workload Processing
@@ -1007,7 +1015,7 @@ Use existing dra-example-driver or Kubernetes test driver for e2e testing.
 - TAS + DRA testing and support as a graduation requirement
 - re-evaluate event-driven DeviceClass tracking for late DeviceClass creation
 - re-evaluate post-scheduling quota reconciliation for DeviceClass drift
-- re-evaluate the support for TopologyAwareScheduling (might be moved to GA)
+- re-evaluate the need for indexing of resourceSlices for CEL performance lookups
 
 #### GA
 


keps/2941-DRA/README.md
```diff
@@ -481,6 +486,9 @@ rules:
 - apiGroups: ["resource.k8s.io"]
   resources: ["resourceclaimtemplates"]
   verbs: ["get", "list", "watch"]
+- apiGroups: ["resource.k8s.io"]
```
sohankunkerkar, line 495: nit: I think you can combine both `resourceslices` + `deviceclasses` into one.

keps/2941-DRA/README.md
```diff
@@ -489,13 +497,60 @@ rules:
...
+
+2. **CEL Evaluation Against Cluster Devices**: Kueue lists all ResourceSlices in the cluster and evaluates
+   the compiled CEL selectors against actual devices. For each request:
+   - The DeviceClass is resolved and its selectors are compiled to pre-filter devices by class, avoiding
+     CEL evaluation against unrelated devices (e.g., NICs when requesting GPUs).
+   - The request's CEL selectors are evaluated against matching devices.
+   - If fewer devices match than the requested count, the workload is marked inadmissible with a descriptive
+     error, preventing quota consumption for unsatisfiable requests.
+
+**Example**: A ResourceClaimTemplate requesting 2 GPUs with `device.attributes["memory"].compareTo(quantity("80Gi")) >= 0`
```
sohankunkerkar: I think memory is a capacity and not an attribute. Per the upstream k8s docs, it should be
```sh
device.capacity["driver.example.com"].memory.compareTo(quantity("80Gi")) >= 0 
```

keps/2941-DRA/README.md
```diff
@@ -265,6 +266,10 @@ administrator configures DeviceClasses with `extendedResourceName`.
   `extendedResourceName`, Kueue unifies quota** using the `deviceClassMappings` logical
   name as the quota key for both paths, preventing over-allocation.
 
+- CEL selectors in ResourceClaimTemplates are validated against cluster devices (ResourceSlices) at admission
```
sohankunkerkar: This reads like a guarantee but the validation is best-effort. Devices can be allocated between Kueue's check and scheduling. Also the reverse: new ResourceSlices published after validation can make previously-unsatisfiable workloads satisfiable, but nothing triggers re-evaluation until the next periodic queueInadmissibleWorkloads cycle. Worth adding a sentence about `WaitForPodsReady` being the safety net and the liveness gap for newly added devices

keps/2941-DRA/README.md
```diff
@@ -489,13 +497,60 @@ rules:
...
 - Permissions are cluster-scoped to allow processing workloads across all namespaces
 - No elevated privileges required beyond standard Kueue controller permissions
 
+### CEL Expression Validation
+
+ResourceClaimTemplates may include CEL (Common Expression Language) selectors that constrain which devices
+can satisfy a request. Kueue validates these CEL selectors before admitting a workload to prevent quota from
+being consumed by workloads whose pods can never be scheduled.
+
+The validation has two stages:
```
sohankunkerkar, line 517: Just a note: before this change, inadmissible meant "not enough quota" and users knew to wait. Now it can also mean "no matching devices exist in the cluster" which is a fundamentally different problem. Users need to change their CEL or an admin needs to add hardware. The condition message should clearly distinguish between quota-based and device-based inadmissibility so users know what action to take.

sohankunkerkar (commented):
Thanks for the write up! I have added a few comments.

keps/2941-DRA/README.md
kannon92: updated.

keps/2941-DRA/README.md
kannon92: updated.

keps/2941-DRA/README.md
kannon92: thanks for that. updated.

keps/2941-DRA/README.md
kannon92: good idea. I added a note here.

kannon92:
@alaypatel07 @tenzen-y could you take a look at this?

keps/2941-DRA/README.md
```diff
@@ -482,20 +491,71 @@ rules:
...
+      - name: gpu
+        exactly:
+          deviceClassName: gpu.example.com
+          count: 2
+          selectors:
+          - cel:
+              expression: 'device.capacity["memory"].compareTo(quantity("80Gi")) >= 0'
+```
+
+**Dependencies**: This feature imports `k8s.io/dynamic-resource-allocation/cel` for CEL compilation and
```
```suggestion
**Dependencies**: This feature imports [`k8s.io/dynamic-resource-allocation/cel`](https://github.com/kubernetes/dynamic-resource-allocation/tree/master/cel) for CEL compilation and
```

This link could help to understand what is this library.

keps/2941-DRA/README.md
```diff
@@ -265,6 +266,14 @@ administrator configures DeviceClasses with `extendedResourceName`.
   `extendedResourceName`, Kueue unifies quota** using the `deviceClassMappings` logical
   name as the quota key for both paths, preventing over-allocation.
 
+- CEL selectors in ResourceClaimTemplates are validated against cluster devices (ResourceSlices) at admission
+  time on a best-effort basis. Workloads with CEL selectors that match fewer devices than requested are rejected
```
```suggestion
- CEL selectors in ResourceClaimTemplates are validated against cluster devices (ResourceSlices) at qtuota reservation
  time on a best-effort basis. Workloads with CEL selectors that match fewer devices than requested are rejected
```

Does this mean `QuotaReservation` phase?

keps/2941-DRA/README.md
```diff
@@ -265,6 +266,14 @@ administrator configures DeviceClasses with `extendedResourceName`.
...
+  to prevent quota leaks. This validation uses the upstream DRA CEL compiler from `k8s.io/dynamic-resource-allocation/cel`.
```
```suggestion
  to prevent quota leaks. This validation uses the upstream DRA CEL compiler from [`k8s.io/dynamic-resource-allocation/cel`](https://github.com/kubernetes/dynamic-resource-allocation/tree/master/cel).
```

keps/2941-DRA/README.md
```diff
@@ -265,6 +266,14 @@ administrator configures DeviceClasses with `extendedResourceName`.
...
+  Note that devices can be allocated between Kueue's check and scheduling, and new ResourceSlices published after
```
```suggestion
  On the other hand, devices can be allocated between Kueue's check and scheduling, and new ResourceSlices published after
```

This is the NOTE section.

keps/2941-DRA/README.md
```diff
@@ -482,20 +491,71 @@ rules:
...
+### CEL Expression Validation
+
+ResourceClaimTemplates may include CEL (Common Expression Language) selectors that constrain which devices
+can satisfy a request. Kueue validates these CEL selectors before admitting a workload to prevent quota from
+being consumed by workloads whose pods can never be scheduled.
+
+The validation has two stages:
+
+1. **CEL Compilation**: Each CEL expression in the request's selectors is compiled using the upstream DRA CEL
+   compiler (`k8s.io/dynamic-resource-allocation/cel`). This catches syntax errors, type errors, and other
```
```suggestion
   compiler ([`k8s.io/dynamic-resource-allocation/cel`](https://github.com/kubernetes/dynamic-resource-allocation/tree/master/cel)). This catches syntax errors, type errors, and other
```

keps/2941-DRA/README.md
```diff
@@ -482,20 +491,71 @@ rules:
...
+   compilation issues before quota admission.
```
```suggestion
   compilation issues before quota reservation.
```

I'd like to distinguish between admission and reservation.

keps/2941-DRA/README.md
```diff
@@ -482,20 +491,71 @@ rules:
...
+   the compiled CEL selectors against actual devices. For each request:
+   - The DeviceClass is resolved and its selectors are compiled to pre-filter devices by class, avoiding
+     CEL evaluation against unrelated devices (e.g., NICs when requesting GPUs).
+   - The request's CEL selectors are evaluated against matching devices.
+   - If fewer devices match than the requested count, the workload is marked inadmissible with a descriptive
+     error indicating that no matching devices exist in the cluster, preventing quota consumption for
+     unsatisfiable requests. The condition message clearly distinguishes between device-based inadmissibility
+     (e.g., "insufficient matching devices for CEL selector") and quota-based inadmissibility so that users
+     know whether they need to adjust their CEL selectors / request admin hardware changes, or wait for
+     quota to become available.
```
tenzen-y: Which Workload condition type does this indicate? QuotaReserved?

keps/2941-DRA/README.md
```diff
@@ -530,12 +590,16 @@ When a user submits a workload and DynamicResourceAllocation feature gate is ena
...
+   - Compile CEL expressions and reject workloads with invalid syntax
+   - Evaluate CEL selectors against actual devices from ResourceSlices, pre-filtering by DeviceClass
+   - Reject workloads where fewer devices match than requested count
+6. Resource Preprocessing:
    - Calculate total device count per device class across all containers and init containers
    - Map device classes to resource names using the configuration
    - Generate preprocessed resource requests for queue admission
-6. Queue Admission: Add the workload to the queue with preprocessed DRA resources.
-7. Status Update: Once admitted, the workload status reflects the assigned flavors and resource usage including DRA resources.
+7. Queue Admission: Add the workload to the queue with preprocessed DRA resources.
```
```suggestion
7. Queue Reservation: Add the workload to the queue with preprocessed DRA resources.
```

keps/2941-DRA/README.md
```diff
@@ -265,6 +266,14 @@ administrator configures DeviceClasses with `extendedResourceName`.
   `extendedResourceName`, Kueue unifies quota** using the `deviceClassMappings` logical
   name as the quota key for both paths, preventing over-allocation.
 
+- CEL selectors in ResourceClaimTemplates are validated against cluster devices (ResourceSlices) at admission
+  time on a best-effort basis. Workloads with CEL selectors that match fewer devices than requested are rejected
+  to prevent quota leaks. This validation uses the upstream DRA CEL compiler from `k8s.io/dynamic-resource-allocation/cel`.
+  Note that devices can be allocated between Kueue's check and scheduling, and new ResourceSlices published after
+  validation can make previously-unsatisfiable workloads satisfiable without triggering re-evaluation until the
+  next periodic `queueInadmissibleWorkloads` cycle. `WaitForPodsReady` serves as the safety net for cases where
+  the validation state diverges from actual device availability at scheduling time.
```
tenzen-y, line 276: Yes, correct. We can strengthen validations by TAS in the future.
But, it's out of scope at this time.

keps/2941-DRA/README.md
```diff
@@ -482,20 +491,71 @@ rules:
...
+  spec:
+    devices:
+      requests:
+      - name: gpu
+        exactly:
+          deviceClassName: gpu.example.com
+          count: 2
+          selectors:
+          - cel:
+              expression: 'device.capacity["memory"].compareTo(quantity("80Gi")) >= 0'
```
```suggestion
              expression: 'device.capacity["gpu.example.com"].memory.compareTo(quantity("80Gi")) >= 0'
```

Is this example really correct? As I can see https://github.com/kubernetes/kubernetes/blob/03779bbd00da25c7fcd03711ddfe466a3322e1d7/pkg/apis/resource/types.go#L1241, my suggestion is correct.

keps/2941-DRA/README.md
```diff
@@ -482,20 +491,71 @@ rules:
...
+     CEL evaluation against unrelated devices (e.g., NICs when requesting GPUs).
+   - The request's CEL selectors are evaluated against matching devices.
+   - If fewer devices match than the requested count, the workload is marked inadmissible with a descriptive
+     error indicating that no matching devices exist in the cluster, preventing quota consumption for
+     unsatisfiable requests. The condition message clearly distinguishes between device-based inadmissibility
+     (e.g., "insufficient matching devices for CEL selector") and quota-based inadmissibility so that users
+     know whether they need to adjust their CEL selectors / request admin hardware changes, or wait for
+     quota to become available.
+
+**Example**: A ResourceClaimTemplate requesting 2 GPUs with `device.capacity["memory"].compareTo(quantity("80Gi")) >= 0`
```
```suggestion
**Example**: A ResourceClaimTemplate requesting 2 GPUs with `device.capacity["gpu.example.com"].memory.compareTo(quantity("80Gi")) >= 0`
```

ditto

keps/2941-DRA/README.md
```diff
@@ -928,6 +994,7 @@ Use existing dra-example-driver or Kubernetes test driver for e2e testing.
 - Alpha implementation completed: December 2024
 - KEP updated to reflect actual implementation: September 2025 by @alaypatel07
 - Extended Resources implementation: January 2026
+- CEL expression validation support added: March 2026 by @kannon92
```
```suggestion
- CEL expression validation support added: April 2026 by @kannon92
```

Let's update this after v0.17 release.

keps/2941-DRA/README.md
```diff
@@ -482,20 +491,71 @@ rules:
...
+    devices:
+      requests:
+      - name: gpu
+        exactly:
+          deviceClassName: gpu.example.com
+          count: 2
+          selectors:
+          - cel:
+              expression: 'device.capacity["memory"].compareTo(quantity("80Gi")) >= 0'
+```
```
tenzen-y, line 554: In this example, how many quotas will be consumed in ClusterQueue?
Let's say that we have the following ClusterQueue, then all the ResourceSlices have only GPUs with 100Gi VRAM. In that case, will ClusterQueue consume only `example.com/gpu: 1` even if it takes larger GPU resources compared with the minimum request size (80Gi)?

```yaml
apiVersion: config.kueue.x-k8s.io/v1beta2
kind: Configuration
...
resources:
  deviceClassMappings:
  - name: example.com/gpu
    deviceClassNames:
    - gpu.example.com
---
kind: ClusterQueue
...
spec:
  resourceGroups:
  - coveredResources: ["cpu", "memory", "example.com/gpu"]
    flavors:
    - name: "default-flavor"
      resources:
      - name: "cpu"
        nominalQuota: 40
      - name: "memory"
        nominalQuota: 200Gi
      - name: "example.com/gpu"
        nominalQuota: 8
```

tenzen-y (commented):
Thank you for updating this KEP with CEL support!
I left some comments.

tenzen-y:
@kannon92 can you also resolve the conflicts?

keps/2941-DRA/README.md
kannon92: This can be somewhat related to @PannagaRao and @sohankunkerkar work on partionable devices.

PannagaRao: With deviceClassMappings alone, Kueue charges device count — `example.com/gpu: 2` in this example. 
                                                                                                                                                                                                                    
With counterMappings from [#10320](https://github.com/kubernetes-sigs/kueue/pull/10320), Kueue reads consumesCounters from matched devices and charges based on actual counter values. The charge is MAX(counter across all matched devices) × count. In this example, if all GPUs have 100Gi memory, the charge would be gpu.memory: 100Gi × 2 = 200Gi

kannon92: For Kueue could we even do quota on VRAM for GPU? 

I think the resources has to point to the deviceClass.

keps/2941-DRA/README.md
kannon92: Yes, eventually this ends up in QuotaReserved.

keps/2941-DRA/README.md
kannon92: updated.

keps/2941-DRA/README.md
kannon92: yes.

kannon92:
@tenzen-y 
this is ready for a review again. I believe I addressed your concerns.

keps/2941-DRA/README.md
```diff
@@ -266,6 +267,14 @@ administrator configures DeviceClasses with `extendedResourceName`.
   `extendedResourceName`, Kueue unifies quota** using the `deviceClassMappings` logical
   name as the quota key for both paths, preventing over-allocation.
 
+- CEL selectors in ResourceClaimTemplates are validated against cluster devices (ResourceSlices) at qtuota reservation
```
```suggestion
- CEL selectors in ResourceClaimTemplates are validated against cluster devices (ResourceSlices) at quota reservation
```

nit: typo

keps/2941-DRA/README.md
```diff
@@ -531,12 +591,16 @@ When a user submits a workload and DynamicResourceAllocation feature gate is ena
...
+   - Evaluate CEL selectors against actual devices from ResourceSlices, pre-filtering by DeviceClass
+   - Reject workloads where fewer devices match than requested count
+6. Resource Preprocessing:
    - Calculate total device count per device class across all containers and init containers
    - Map device classes to resource names using the configuration
    - Generate preprocessed resource requests for queue admission
-6. Queue Admission: Add the workload to the queue with preprocessed DRA resources.
-7. Status Update: Once admitted, the workload status reflects the assigned flavors and resource usage including DRA resources.
+7. Queue Reservation: Add the workload to the queue with preprocessed DRA resources.
+8. Status Update: Once admitted, the workload status reflects the assigned flavors and resource usage including DRA resources.
```
```suggestion
8. Status Update: Once the quota is reserved, the workload status reflects the assigned flavors and resource usage, including DRA resources.
```

nit: admission -> quota reservation

keps/2941-DRA/README.md
tenzen-y: In that case, can you mention the `QuotaReserved` in this message like the following:

```suggestion
     unsatisfiable requests. The `QuotaReserved` condition message clearly distinguishes between device-based inadmissibility
     (e.g., "insufficient matching devices for CEL selector") and quota-based inadmissibility so that users
     know whether they need to adjust their CEL selectors / request admin hardware changes, or wait for
     quota to become available.
```

(I just aded `QuotaReseerved` term)

keps/2941-DRA/README.md
tenzen-y: @PannagaRao @kannon92 Thank you for describing that. I understand what happens here.
I'm wondering if we should `counterMappings` somehow mechanism as part of this CEL evaluation mechanism? Otherwise, there will be large gaps between actual consuming device capacities and consumed quotas.

keps/2941-DRA/README.md
```diff
@@ -483,20 +492,71 @@ rules:
...
+being consumed by workloads whose pods can never be scheduled.
+
+The validation has two stages:
+
+1. **CEL Compilation**: Each CEL expression in the request's selectors is compiled using the upstream DRA CEL
+   compiler ([`k8s.io/dynamic-resource-allocation/cel`](https://github.com/kubernetes/dynamic-resource-allocation/tree/master/cel)). This catches syntax errors, type errors, and other
+   compilation issues before quota reservation.
+
+2. **CEL Evaluation Against Cluster Devices**: Kueue lists all ResourceSlices in the cluster and evaluates
+   the compiled CEL selectors against actual devices. For each request:
```
tenzen-y, line 523: I'm just wondering if we can filter out the obtaining ResourceSlices instead of obtaining all ResourceSlices to improve the Kueue performance.

@kannon92 Are we able that?

tenzen-y (commented):
Thank you for updating that.

tenzen-y:
Also, I'm wondering if we should introduce a dedicated FeatureGate, something like DRACELExpressions.

keps/2941-DRA/README.md
kannon92: I think there is plans to use this PR for the PartionableDevices work.

There is a separate KEP for that. I think we should keep this separate for now.

CounterMappings is only relevant if you are using PartitionableDevices but the cel expression is a GA feature for structured parameters.

keps/2941-DRA/README.md
kannon92: Maybe. It is indexed by driverName but resource slices can be at the node level so I'm not sure how much that would gain us.

kannon92:
> Also, I'm wondering if we should introduce a dedicated FeatureGate, something like DRACELExpressions.

CEL is critical for DRA so I think we should not go with a feature gate. We have the DRA feature gate which is alpha and CEL is a GA feature of DRA so I think its fine to guard this under the existing feature gate.

keps/2941-DRA/README.md
tenzen-y: My motivation for the filtering of ResourceSlice is optimizing performance both for Kueue cache and the kube-apiserver load. I'm ok with postponing the optimizations to Beta or GA graduation.

Can you add such optimization to the graduation criteria?

keps/2941-DRA/README.md
tenzen-y: > I think there is plans to use this PR for the PartionableDevices work.
> 
> There is a separate KEP for that. I think we should keep this separate for now.
> 
> CounterMappings is only relevant if you are using PartitionableDevices but the cel expression is a GA feature for structured parameters.

The key point is `Otherwise, there will be large gaps between actual consuming device capacities and consumed quotas.`.

So, if we don't introduce the `counterMappings` somehow mechanism for DRA CEL impls, we should restrict DRA CEL usage only when they use DRA CEL with DRA PartitionableDevices.

Otherwise, this DRA CEL introduces critical Quota violations because, as I mentioned, DRA CEL w/o `counterMappings` can not guarantee how many actual resources are consumed.

tenzen-y:
> > Also, I'm wondering if we should introduce a dedicated FeatureGate, something like DRACELExpressions.
> 
> CEL is critical for DRA so I think we should not go with a feature gate. We have the DRA feature gate which is alpha and CEL is a GA feature of DRA so I think its fine to guard this under the existing feature gate.

I will decide FG based on the final https://github.com/kubernetes-sigs/kueue/pull/9841#discussion_r3112398693 decision.

keps/2941-DRA/README.md
kannon92: Done.

keps/2941-DRA/README.md
kannon92: I'm not really following you here.

If I have a request with CEL asking for > 80 GB of memory, it will get the device and then count the request to that device.

CEL is more about getting the right device class and not about enforcing GPU memory. With GPUs there isn't really a way to enforce memory as you would with cgroups. 

Maybe I focused too much on CEL for memory in this update but CEL is basically a way to get any kind of attribute. You could use it to say "I want a specific device" ex: https://github.com/kubernetes-sigs/dra-example-driver/blob/main/demo/gpu-test8.yaml.

For DRANET, you can see other examples: https://dranet.dev/docs/user/mpi-operator/.

In these examples we are grabbing a certain device with specific attributes. The hope of adding CEL is to support things all the ways CEL can be used with DRA.

I don't think CEL evaulation needs to tie into Partionable devices. Happy to discuss tomorrow if you have more questions.

mimowo: Hm, I'm also not following where is the concern, but I'm maybe missing some of the DRA context. 

Let me ask about this specific scenario - in my ResourceFlavor, I have 
- 2 devices with 1 GPU, 80Gi mem 
- 2 devices with 1 GPU, 40Gi mem

Now, with the configurations as above I have workloads 
- (1.) matching >=80Gi devices, and requesting 1 device, so it consumes `example.com/gpu: 1` - admitted by quota, binds to one of the machines; 3 GPU of quota left 
- (2.) matching >=80Gi devices, and requesting 1 device, so it consumes `example.com/gpu: 1` - admitted by quota, binds to one of the machines; 2 GPU of quota left
- (3.) matching >=80Gi devices, and requesting 1 device, so it consumes `example.com/gpu: 1` - admitted by quota, but cannot find the machines because the only two 80Gi machines are already taken; It cannot schedule using 2 machines with 40Gi each

So (3.) means the workload remains pending, which is a right risk to call out. However:
1. I think this risk always existed in quota-only Kueue, also even without DRA Kueue could admit workload requesting 8GPU in one Pod, even though there all machines have only 1GPU
2. In my mind this is where Kueue needs to know "infra", and for that we will have the DRA-TAS integration via scheduler-library which will fix such issues
3. for now this would be the risk we need to document, but I don't think this should prevent the CEL support completely (unless I'm missing something)

mimowo: In other words, I think the lack of machine / infra awareness is indeed a risk, but it is not new, and in Kueue the right layer to fix problems like this is TAS. So, we would still consume the quota the same, but block admission of workload (3.) based on TAS+DRA constraints rather than quota.

tenzen-y: @mimowo Thank you for getting back comments here.

As I chatted with @mimowo in Slack, I almost agreed with this fragmentation between Node-level resources and quota. Certainly, we have some risks when cluster admins misconfigure too flexible CEL parameters like `'device.capacity["memory"].compareTo(quantity("16Gi")) >= 0'`. In this scenario, if the cluster has only accelerators with over 141 GiB VRAM, it looks like a very soft restriction.

But, the same as before TAS, we had similar problems about fragmentations in non-DRA environments as well. We can recommend not specifying too flexible CEL parameters to avoid much fragmentation between Node allocatable resources and nominalQuota in our user documentation.

And, in the future, we can mitigate these risks in TAS+DRA functionalities.

tenzen-y (commented):
Based on https://github.com/kubernetes-sigs/kueue/pull/9841#discussion_r3112398693 discussion, I decided not to request adding mitigation for fragmentation between Node allocatable resources with DRA CEL and ClusterQueue quotas.

And, I don't request additional FG for that.

@kannon92 Thank you!
/lgtm
/approve

/hold
for @mimowo

mimowo:
/unhold
I'm good to give it a go as Alpha, thank you 👍

PR #9850, opened 2026-03-13
feat: Enable configuration of visibility server bind address and port…

Nilsachy:
What type of PR is this?
/kind feature

Enable configuration of visibility server bind address and port through the Configuration API

Which issue(s) this PR fixes:
Fixes #

```release-note
VisibilityOnDemand: Introduce a new configuration type: "VisibilityServer" which allows passing custom "VisibilityServerConfiguration" limited to BindAddress and BindPort as a first iteration. Removing the deprecated "visibility-server-port" flag.
```


pkg/visibility/server.go
```diff
@@ -97,14 +98,25 @@ func CreateAndStartVisibilityServer(ctx context.Context, kueueMgr *qcache.Manage
...
 }
 
-func applyVisibilityServerOptions(config *genericapiserver.RecommendedConfig, enableInternalCertManagement bool, tlsOpts *tlsconfig.TLS) error {
+func applyVisibilityServerOptions(config *genericapiserver.RecommendedConfig, cfg *configapi.Configuration, tlsOpts *tlsconfig.TLS) error {
 	o := genericoptions.NewRecommendedOptions("", codecs.LegacyCodec(
 		visibilityv1beta2.SchemeGroupVersion,
 		visibilityv1beta1.SchemeGroupVersion,
 	))
 	o.Etcd = nil
 	o.SecureServing.BindPort = 8082
```
kshalot: Maybe this could become an explicit default value in the API instead of being hard-coded here?

pkg/visibility/server.go
```diff
@@ -97,14 +98,25 @@ func CreateAndStartVisibilityServer(ctx context.Context, kueueMgr *qcache.Manage
...
-	if enableInternalCertManagement {
+
+	// Apply visibility overrides from Configuration API if present
+	if cfg.Visibility != nil {
+		if cfg.Visibility.BindPort != nil {
+			o.SecureServing.BindPort = int(*cfg.Visibility.BindPort)
+		}
+		if cfg.Visibility.BindAddress != nil && *cfg.Visibility.BindAddress != "" {
+			o.SecureServing.BindAddress = net.ParseIP(*cfg.Visibility.BindAddress)
```
kshalot: I'm not 100% sure the other `BindAddress` fields in our config are validated, but if the address is not parseable, we'll end up with a no-op (I think since it will assign `nil`, but you can verify the impact of that).

Maybe we could use `net.ParseIP` to validate the addresses when the config is being loaded.

pkg/visibility/server.go
Nilsachy: Good point, done.

pkg/visibility/server.go
Nilsachy: Do you mean creating a new `func validateBindAddresses(c *configapi.Configuration)` for all TCP and IP address fields defined in the config?

kshalot: I think I'd start by validating only the visibility bind address, to keep the scope of this PR tighter. `validateVisibility` or something along those lines.

Nilsachy: Done.

apis/config/v1beta2/configuration_types.go
```diff
@@ -588,3 +592,14 @@ type WorkloadRetentionPolicy struct {
 	// +optional
 	AfterDeactivatedByKueue *metav1.Duration `json:"afterDeactivatedByKueue,omitempty"`
 }
+
+type VisibilityServerConfiguration struct {
+	// BindAddress is the IP address on which to listen for the visibility server.
```
kshalot: Nit: "listen for" sounds a bit backwards IMO, because it's the visibility server that is doing the listening. I'd propose:
```suggestion
	// BindAddress is the IP address the visibility server listens on.
```

Same for the port.

apis/config/v1beta2/configuration_types.go
```diff
@@ -105,6 +105,10 @@ type Configuration struct {
 	// of Kueue-managed objects. A nil value disables all automatic deletions.
 	// +optional
 	ObjectRetentionPolicies *ObjectRetentionPolicies `json:"objectRetentionPolicies,omitempty"`
+
+	// Visibility configures the visibility server.
+	// +optional
+	Visibility *VisibilityServerConfiguration `json:"visibility,omitempty"`
```
kshalot: Nit: I'd consider naming this `visibilityServer`. I feel like it's a bit less ambiguous? Because we are configuring the server specifically, not some general notion of "visibility" in Kueue.

keps/168-2-pending-workloads-visibility/README.md
```diff
@@ -227,6 +232,13 @@ The introduced API uses mechanism of subresources. It means, that in the future
 GET /apis/visibility.kueue.x-k8s.io/VERSION/clusterqueues/CQ_NAME/admitted_workloads?offset=0&limit=1000
 ```
 
+### Technical Details
+
+The Visibility server can be configured in the `kueueconfigurations` API under `visibility`. The following options are available:
+
+- `bindAddress`: The IP address on which to listen for the visibility server. If not specified, the server will listen on all available network interfaces.
+- `bindPort`: The port on which to listen for the visibility server. Defaults to `8082`.
```
kshalot, line 240: Similarly to my comment about the docstring, I'd reword this "listen for" to "listen on"

Nilsachy: Done.

apis/config/v1beta2/configuration_types.go
Nilsachy: Done.

apis/config/v1beta2/configuration_types.go
Nilsachy: Done.

pkg/config/validation.go
```diff
@@ -568,6 +571,17 @@ func validateTLS(c *configapi.Configuration) field.ErrorList {
 	return allErrs
 }
 
+func validateVisibilityServer(c *configapi.Configuration) field.ErrorList {
+	var allErrs field.ErrorList
+	if c.VisibilityServer == nil || c.VisibilityServer.BindAddress == nil || *c.VisibilityServer.BindAddress == "" {
```
Nilsachy: I am not sure if an empty string for BindAddress is a valid case

apis/config/v1beta2/configuration_types.go
```diff
@@ -618,3 +622,14 @@ type WorkloadRetentionPolicy struct {
 	// +optional
 	AfterDeactivatedByKueue *metav1.Duration `json:"afterDeactivatedByKueue,omitempty"`
 }
+
+type VisibilityServerConfiguration struct {
+	// BindAddress is the IP address the visibility server listens on.
```
dkaluza, line 627: It would be good to specify what is the default behavior here also.

Is this all network interfaces? Or just localhost? Or something else?

pkg/config/config_test.go
```diff
@@ -1104,7 +1122,7 @@ webhook:
 
 			// Compare the loaded configuration
 			configCmpOpts := cmp.Options{
-				cmpopts.IgnoreFields(configapi.Configuration{}, "ControllerManager"),
+				cmpopts.IgnoreFields(configapi.Configuration{}, "ControllerManager", "VisibilityServer"),
```
dkaluza, line 1125: Why are we adding the "VisibilityServer" entries above if we are exludeing them from comparison anyway?

For me cleanest solution would be to have actually just a separate test for it, but I know test tables are the preferred solution in golang. The disadvantage is that we are doing here like 20 modifications and I do not have a feeling those actually test anything more than the first addition in this table (even if we enable cmp of those).

pkg/config/validation.go
```diff
@@ -568,6 +571,17 @@ func validateTLS(c *configapi.Configuration) field.ErrorList {
 	return allErrs
 }
 
+func validateVisibilityServer(c *configapi.Configuration) field.ErrorList {
```
dkaluza, line 575: What about port validation?

pkg/config/validation.go
mimowo: No strong view, but I would reject against such a value. Otherwise user may be thinking hey did something useful but Kueue is not reacting. It is always easier to relax validation later (than to add validation)

test/e2e/customconfigs/visibility_server_test.go
```diff
@@ -229,17 +241,6 @@ var _ = ginkgo.Describe("Visibility Server", func() {
...
-		patchedDeployment := originalDeployment.DeepCopy()
-		for i, c := range patchedDeployment.Spec.Template.Spec.Containers {
-			if c.Name == "manager" {
-				container := &patchedDeployment.Spec.Template.Spec.Containers[i]
-				container.Args = append(c.Args, fmt.Sprintf("--visibility-server-port=%d", customVisibilityPort))
-			}
-		}
-		gomega.Expect(k8sClient.Update(ctx, patchedDeployment)).To(gomega.Succeed())
-		util.WaitForKueueAvailabilityNoRestartCountCheck(ctx, k8sClient)
-
```
mimowo, line 242: Why removing the test? Let's keep both tests for as long both mechanisms are supported

Nilsachy: Oh I understood your previous comment wrongly. We want to keep both mechanisms for now? So we should still allow users to define the visibility port through a flag?

cmd/kueue/main.go
```diff
@@ -129,9 +129,6 @@ func main() {
 	var featureGates string
 	flag.StringVar(&featureGates, "feature-gates", "", "A set of key=value pairs that describe feature gates for alpha/experimental features.")
 
-	var visibilityServerPort int
-	flag.IntVar(&visibilityServerPort, "visibility-server-port", 8082, "The port the visibility server binds to.")
```
mimowo, line 133: Oh I thought we are just deprecating - not removing completely in this release, but I'm ok to drop it actually if we don't have a use case. The only use case I see is to make the migration smooth - admin can upgrade Kueue first, and only then change configuration. If we just remove in 0.17 then user needs to upgrade and change the configuration as a single step.

However, if you don't see this a problem, let's drop for simplicity.

Nilsachy: I see. So you would prefer keeping both options for now? And if that is the case, which one takes precedence on the other? I assume configuration API should override the flag?

test/e2e/customconfigs/visibility_server_test.go
mimowo: Yeah this was the comment by @tenzen-y: https://github.com/kubernetes-sigs/kueue/pull/9863#discussion_r2947704051 "Let's try to expand Configuration field for the 0.17 release, then remove this flag in the future releaase (might be 0.18 or 0.19)"

mimowo: So if not too much work I would keep both for 0.17. Configuration wins if specified. The flag is only respected if the configuration is absent.

cmd/kueue/main.go
mimowo: > I see. So you would prefer keeping both options for now?

Yeah, again, no strong view, tbut I think this is more aligned with the deprecation policy - we keep both options supported for at least one release for the ease of migration.

>  And if that is the case, which one takes precedence on the other? I assume configuration API should override the flag?

I would say configuration wins, so that user can do the following upgrade steps:
1. set the flag in 0.16
2. upgrade to 0.17.0 - all continues to work
3. modify configuration - all continues to work
4. drop feature gate - all continues to work

pkg/config/validation.go
mimowo: +1

pkg/config/config_test.go
Nilsachy: If you look closely, the VisibilityServer entries are actually not being added to the wantConfiguration structs inside `TestTLSOptionsFeatureGate`. They are only added for the structs in `TestLoad` which tests that the configuration file is parsed correctly across all possible fields, including defaults.

pkg/config/validation.go
Nilsachy: Done.

apis/config/v1beta2/configuration_types.go
Nilsachy: This defaults to all network interfaces: https://github.com/kubernetes/apiserver/blob/master/pkg/server/options/serving.go#L120. Added the comment.

pkg/config/validation.go
Nilsachy: Done.

cmd/kueue/main.go
Nilsachy: Done. Please let me know if the documentation is clear enough.

test/e2e/customconfigs/visibility_server_test.go
Nilsachy: Done.

site/content/en/docs/tasks/manage/monitor_pending_workloads/pending_workloads_on_demand.md
```diff
@@ -74,7 +74,9 @@ You can disable it by setting the `VisibilityOnDemand` feature gate. Check the [
 
 If you disable the feature, you also need to remove the associated `APIService` from your cluster by doing `kubectl delete APIService v1beta2.visibility.kueue.x-k8s.io`.
 
-If you need the embedded visibility server to run on a different port, you can use the `--visibility-server-port` flag when running the Kueue controller manager. For example: `--visibility-server-port=8443`.
+If you need the embedded visibility server to run on a different port, you can use the Configuration API `visibilityServer.bindPort`.
+
+Alternatively, you can use the `--visibility-server-port` flag when running the Kueue controller manager. For example: `--visibility-server-port=8443`. Keep in mind, this option will be deprecated soon.
```
mimowo, line 79: I think we should make it deprecated day one, so that we can remove in 0.18 or 0.19

mimowo: So I would say Alternatively, you can use the `--visibility-server-port` flag when running the Kueue controller manager. For example: `--visibility-server-port=8443`.However, this option is deprecated as of 0.17 and will be removed in the future.

Nilsachy: Thanks, I will open a PR for the updates of the docs.

PR #9863, opened 2026-03-13
Allow custom port for visibility server

Nilsachy:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it:**
Allow to pass custom port to the visibility server

**Which issue(s) this PR fixes:**
Fixes https://github.com/kubernetes-sigs/kueue/issues/9650

**Does this PR introduce a user-facing change?**

```release-note
VisibilityOnDemand: Introduce a new Kueue deployment argument, --visibility-server-port, which allows passing custom port when starting the visibility server.
```

### DIFF
--- a/test/e2e/customconfigs/visibility_server_test.go
+++ b/test/e2e/customconfigs/visibility_server_test.go
@@ -55,6 +55,7 @@ const (
 var _ = ginkgo.Describe("Visibility Server", func() {
 	var originalDeployment appsv1.Deployment
 	var originalService corev1.Service
+	var cq *kueue.ClusterQueue
 
 	ginkgo.BeforeEach(func() {
 		err := k8sClient.Get(ctx, types.NamespacedName{Name: kueueManagerName, Namespace: kueueNS}, &originalDeployment)
@@ -64,29 +65,29 @@ var _ = ginkgo.Describe("Visibility Server", func() {
 		gomega.Expect(err).NotTo(gomega.HaveOccurred())
 
 		ginkgo.By("Creating a ClusterQueue")
-		cq := &kueue.ClusterQueue{
+		cq = &kueue.ClusterQueue{
 			ObjectMeta: metav1.ObjectMeta{Name: cqName},
 		}
 		util.CreateClusterQueuesAndWaitForActive(ctx, k8sClient, cq)
+	})
 
-		ginkgo.DeferCleanup(func() {
-			ginkgo.By("Restoring the original deployment")
-			latestDeployment := &appsv1.Deployment{}
-			gomega.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: kueueManagerName, Namespace: kueueNS}, latestDeployment)).To(gomega.Succeed())
-			latestDeployment.Spec = originalDeployment.Spec
-			gomega.Expect(k8sClient.Update(ctx, latestDeployment)).To(gomega.Succeed())
+	ginkgo.AfterEach(func() {
+		ginkgo.By("Restoring the original deployment")
+		latestDeployment := &appsv1.Deployment{}
+		gomega.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: kueueManagerName, Namespace: kueueNS}, latestDeployment)).To(gomega.Succeed())
+		latestDeployment.Spec = originalDeployment.Spec
+		gomega.Expect(k8sClient.Update(ctx, latestDeployment)).To(gomega.Succeed())
 
-			ginkgo.By("Restoring the original service")
-			latestService := &corev1.Service{}
-			gomega.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: kueueVisibilityServerName, Namespace: kueueNS}, latestService)).To(gomega.Succeed())
-			latestService.Spec.Ports = originalService.Spec.Ports
-			gomega.Expect(k8sClient.Update(ctx, latestService)).To(gomega.Succeed())
+		ginkgo.By("Restoring the original service")
+		latestService := &corev1.Service{}
+		gomega.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: kueueVisibilityServerName, Namespace: kueueNS}, latestService)).To(gomega.Succeed())
+		latestService.Spec.Ports = originalService.Spec.Ports
+		gomega.Expect(k8sClient.Update(ctx, latestService)).To(gomega.Succeed())
 
-			util.WaitForKueueAvailabilityNoRestartCountCheck(ctx, k8sClient)
+		util.WaitForKueueAvailabilityNoRestartCountCheck(ctx, k8sClient)
 
-			ginkgo.By("Cleaning up cluster queue")
-			util.ExpectObjectToBeDeleted(ctx, k8sClient, cq, true)
-		})
+		ginkgo.By("Cleaning up cluster queue")
+		util.ExpectObjectToBeDeleted(ctx, k8sClient, cq, true)
 	})
 
 	ginkgo.It("Should use the RBAC identity from the provided kubeconfig", func() {


test/e2e/customconfigs/visibility_kubeconfig_test.go
```diff
@@ -35,26 +37,32 @@ import (
...
+	kueueVisibilityServerName = "kueue-visibility-server"
+	configMapName             = "visibility-kubeconfig-test"
+	customSAName              = "visibility-test-sa"
+	customSecretName          = "visibility-test-sa-token"
+	testLabelKey              = "visibility-test-rbac"
+	testLabelValue            = "true"
+	kubeconfigVolName         = "kubeconfig-vol"
+	customSAVolName           = "custom-sa-vol"
+	customSAMountPath         = "/etc/custom-sa"
+	visibilityServerPort      = 9444
```
kannon92: I would like to see us keep the default and maybe have this change under a separate test case.

We now are no longer testing the default case with this change.

Nilsachy: Refactored current test setup and add separate test.

cmd/kueue/main.go
```diff
@@ -129,6 +129,9 @@ func main() {
 	var featureGates string
 	flag.StringVar(&featureGates, "feature-gates", "", "A set of key=value pairs that describe feature gates for alpha/experimental features.")
 
+	var visibilitySecurePort int
```
kshalot: Nit: Maybe `visibilityServerSecurePort`? Also maybe you could drop "secure", if we want to be specific then it would be `HttpsPort`. But this is super nitty. Or just `visibilityServerPort` (because there is no such thing as a "secure" port).

test/e2e/customconfigs/visibility_server_test.go
```diff
@@ -35,29 +37,65 @@ import (
...
+		err = k8sClient.Get(ctx, types.NamespacedName{Name: kueueVisibilityServerName, Namespace: kueueNS}, &originalService)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+		ginkgo.By("Creating a ClusterQueue")
+		cq := &kueue.ClusterQueue{
+			ObjectMeta: metav1.ObjectMeta{Name: cqName},
+		}
+		util.CreateClusterQueuesAndWaitForActive(ctx, k8sClient, cq)
+
+		ginkgo.DeferCleanup(func() {
```
mimowo: Why DeferCleanup for BeforeEach, I think it is more idiomatic to cleanup in AfterEach for symmetry

test/e2e/customconfigs/visibility_server_test.go
```diff
@@ -195,6 +217,35 @@ var _ = ginkgo.Describe("Visibility Server KubeConfig flag with RBAC", func() {
...
+		ginkgo.By("Updating the visibility-server service's targetPort")
+		patchedService := originalService.DeepCopy()
+		for i, p := range patchedService.Spec.Ports {
+			if p.Name == "https" {
+				patchedService.Spec.Ports[i].TargetPort = intstr.FromInt32(customVisibilityPort)
+			}
+		}
+		gomega.Expect(k8sClient.Update(ctx, patchedService)).To(gomega.Succeed())
+
+		patchedDeployment := originalDeployment.DeepCopy()
```
mimowo, line 233: Add ginkgo.By for the step of patching the Deployment

site/content/en/docs/tasks/manage/monitor_pending_workloads/pending_workloads_on_demand.md
```diff
@@ -74,6 +74,8 @@ You can disable it by setting the `VisibilityOnDemand` feature gate. Check the [
 
 If you disable the feature, you also need to remove the associated `APIService` from your cluster by doing `kubectl delete APIService v1beta2.visibility.kueue.x-k8s.io`.
 
+If you need the embedded visibility server to run on a different port, you can use the `--visibility-secure-port` flag when running the Kueue controller manager. For example: `--visibility-secure-port=8443`.
```
kshalot: QQ: Since the dedicated API is coming (as far as I understand), do we want to "advertise" the flag in the documentation? In terms of reduction of potential integrations using this temporary solution.

mimowo: I'm ok to mention it, people who need it, they will find it anyway. With the docs we can already have a spot to update when the Configuration APi is ready.

test/e2e/customconfigs/visibility_server_test.go
Nilsachy: `AfterEach` does provide visual symmetry, but `DeferCleanup` seems to be a better choice here:

**Scoping & Encapsulation:** It allows us to keep variables (like cq, sa, etc.) locally scoped within the block. `AfterEach` would require us to move these to the outer Describe scope, defining variables used only for cleanup.

**Registration Safety:** `DeferCleanup` is only registered if the code execution actually reaches that line. This prevents the test from attempting to clean up resources that were never created due to a mid-setup failure.

**Automatic LIFO:** It naturally follows Last-In-First-Out (LIFO) order, which IMO is safer for handling dependent resources (e.g., deleting a Pod before its Namespace).

Does that make sense or do you still prefer the use of `AfterEach` block instead?

test/e2e/customconfigs/visibility_server_test.go
Nilsachy: Done.

cmd/kueue/main.go
Nilsachy: Used visibilityServerPort instead

test/e2e/customconfigs/visibility_server_test.go
mimowo: > Scoping & Encapsulation: It allows us to keep variables (like cq, sa, etc.) locally scoped within the block. AfterEach would require us to move these to the outer Describe scope, defining variables used only for cleanup.

That is a good point, nd I'm all for the encapsulation, but I feel some disconnect as the code declares many consts in the global scope, eg `customVisibilityPort`, or `kueueVisibilityServerName`. So it feels like a half-done effort. Let me know if I'm missing something.

AI: please double check all global consts if you can restrict their scope after the change.

> Registration Safety: DeferCleanup is only registered if the code execution actually reaches that line. This prevents the test from attempting to clean up resources that were never created due to a mid-setup failure.

I don't buy this. AfterEach also executes regardless of failure / success.

> Automatic LIFO: It naturally follows Last-In-First-Out (LIFO) order, which IMO is safer for handling dependent resources (e.g., deleting a Pod before its Namespace).

Sort of, but it rarely makes a difference. If the order of deletion makes a difference it is usually a sign of some bugs. For Deleting Pods and Namespaces it makes a difference, but we have a helper for that `util.DeleteNamespace` (which is probably badly named).

Nilsachy: But IIUC correctly, `AfterEach` is always executed (unlike `DeferCleanup` which is only executed if we reach that line). So would you then prefer the use of `AfterEach` with non-nil checks?

Something like this:
```
        if cq != nil {
			ginkgo.By("Cleaning up cluster queue")
			util.ExpectObjectToBeDeleted(ctx, k8sClient, cq, true)
			cq = nil
		}
		if sa != nil {
			ginkgo.By("Cleaning up custom SA")
			util.ExpectObjectToBeDeleted(ctx, k8sClient, sa, true)
			sa = nil
		}
		if secret != nil {
			ginkgo.By("Cleaning up secret")
			util.ExpectObjectToBeDeleted(ctx, k8sClient, secret, true)
			secret = nil
		}
		if cm != nil {
			ginkgo.By("Cleaning up config map")
			util.ExpectObjectToBeDeleted(ctx, k8sClient, cm, true)
			cm = nil
		}
```

IMO it is less readable, but let me know and of course we can update this.

mimowo: I don't like the checks, but I was pretty sure the checks are done internally https://github.com/kubernetes-sigs/kueue/blob/612c05549532ebe17d1b7b9c098f798064a497a3/test/util/util.go#L145-L147

Nilsachy: Done. In fact custom SA, secret, and configmaps are only specific to the RBAC (previous) test so this can be cleaned up separately.

cmd/kueue/main.go
```diff
@@ -129,6 +129,9 @@ func main() {
 	var featureGates string
 	flag.StringVar(&featureGates, "feature-gates", "", "A set of key=value pairs that describe feature gates for alpha/experimental features.")
 
+	var visibilityServerPort int
```
tenzen-y, line 132: Port-specific flag sounds good to me as well.
Let's try to expand Configuration field for the 0.17 release, then remove this flag in the future releaase (might be 0.18 or 0.19)

Nilsachy: Once this is merged I will pick up #9850

Issue #9876, opened 2026-03-14
Support guaranteed minimum runtime before fair sharing preemption


mukund-wayve:
**What would you like to be added**:

A minAdmitDuration configuration option for fair sharing that prevents recently admitted workloads from being preempted by workloads in other ClusterQueues. Workloads admitted less than minAdmitDuration ago would be excluded from cross-CQ preemption candidates.

**Why is this needed**:

When the owner ClusterQueue wants its nominal quota back, the scheduler may preempt workloads before they have had enough time to make meaningful progress. There is currently no way to guarantee a minimum runtime for workloads against cross-CQ fair sharing preemption.

A configurable minimum admit duration would ensure all workloads make at least some progress before becoming eligible for fair sharing preemption.

This is complementary to [#8522](https://github.com/kubernetes-sigs/kueue/issues/8522), which addresses time-based preemption protection within a single ClusterQueue. This issue covers the cross-ClusterQueue case during fair sharing rebalancing.

User Story

As a cluster administrator, I want to configure a minimum duration that workloads must be admitted before they can be preempted by fair sharing, so that workloads borrowing resources from other ClusterQueues are guaranteed enough time to make meaningful progress.

**Completion requirements**:

This enhancement requires the following artifacts:

- [x] Design doc
- [x] API change
- [x] Docs update

The artifacts should be linked in subsequent comments.

PR #9877, opened 2026-03-15
KEP-9876: Cross ClusterQueue preemption guaranteed runtime

mukund-wayve:
#### What type of PR is this?
/kind kep

Part-of #9876 

```release-note
NONE
```

### DIFF
--- a/keps/9876-cross-cq-guaranteed-runtime/README.md
+++ b/keps/9876-cross-cq-guaranteed-runtime/README.md
@@ -1,0 +1,450 @@
+# KEP-9876: Cross-ClusterQueue Minimum Admit Duration
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: Guaranteeing Minimum Runtime for GPU Training Jobs](#story-1-guaranteeing-minimum-runtime-for-gpu-training-jobs)
+    - [Story 2: Ensuring Checkpoint Completion Before Preemption](#story-2-ensuring-checkpoint-completion-before-preemption)
+    - [Story 3: Grace Period for Nominal Quota Reclaim](#story-3-grace-period-for-nominal-quota-reclaim)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API Changes](#api-changes)
+  - [Preemption Eligibility](#preemption-eligibility)
+  - [Future Extensibility](#future-extensibility)
+  - [Validation](#validation)
+  - [Test Plan](#test-plan)
+    - [Unit Tests](#unit-tests)
+    - [Integration tests](#integration-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [GA](#ga)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Rely on Within-ClusterQueue Time-Based Preemption Alone](#rely-on-within-clusterqueue-time-based-preemption-alone)
+  - [Per-ClusterQueue Configuration](#per-clusterqueue-configuration)
+  - [Separate Duration for Each Reclaim Type](#separate-duration-for-each-reclaim-type)
+<!-- /toc -->
+
+## Summary
+
+Cross-ClusterQueue preemption can cause workloads to be preempted before they
+have had enough time to make meaningful progress. This KEP introduces two
+configuration fields that allow administrators to guarantee a minimum runtime
+before workloads become eligible for cross-CQ preemption:
+
+- **`fairSharing.minAdmitDuration`**: protects workloads from fair sharing
+  rebalancing (`InCohortFairSharing`). Lives in the `FairSharing`
+  configuration because it only applies when fair sharing is enabled.
+- **`reclaimMinAdmitDuration`**: protects workloads from cross-CQ reclaim
+  (`InCohortReclamation` and `InCohortReclaimWhileBorrowing`). This is a
+  top-level `Configuration` field because reclaim occurs in both classical
+  and fair sharing preemption modes.
+
+When configured, a workload whose `QuotaReserved` condition was set less than
+the applicable duration ago is skipped as a preemption candidate.
+
+**Terminology**: Throughout this KEP, "quota reservation time" refers to the
+`LastTransitionTime` of the workload's `QuotaReserved` condition (when
+`Status=True`). This is the moment when quota was reserved for the workload.
+
+## Motivation
+
+Cross-ClusterQueue preemption in Kueue can happen for several reasons with
+fundamentally different semantics:
+
+- **Fair sharing rebalancing** (`InCohortFairSharing`): When fair sharing is
+  enabled, Kueue rebalances surplus resources across ClusterQueues within a
+  cohort hierarchy by preempting workloads from ClusterQueues that are
+  consuming more than their fair share. Neither side has a stronger
+  entitlement — this is a fairness negotiation over shared surplus.
+
+- **Nominal quota reclaim** (`InCohortReclamation`): A ClusterQueue reclaims
+  resources it owns (its nominal quota) from other ClusterQueues that
+  borrowed them. The owner has a clear entitlement to these resources. This
+  applies regardless of whether fair sharing is enabled.
+
+- **Reclaim while borrowing** (`InCohortReclaimWhileBorrowing`): A
+  ClusterQueue that itself needs to borrow preempts workloads from other
+  borrowing ClusterQueues via the `borrowWithinCohort` policy. This applies
+  in classical preemption mode.
+
+In all cases, a workload that was just admitted may be preempted before it
+has had enough time to make meaningful progress — for example, before reaching
+a useful checkpoint or completing a meaningful unit of work.
+
+Administrators need a way to guarantee a minimum runtime for workloads so that
+preemption only targets workloads that have already had a reasonable opportunity
+to make progress. Because the preemption types carry different strength of
+claim, administrators may want different thresholds — for example, a longer
+protection window for fair sharing (where neither side has priority) and a
+shorter one for reclaim (where the owner has a legitimate entitlement). This
+is analogous to Slurm's `PreemptExemptTime`, which provides a guaranteed
+minimum runtime before a job becomes preemptible.
+
+### Goals
+
+- Allow administrators to configure a global minimum runtime guarantee for
+  workloads before they become eligible for fair sharing preemption, so that
+  workloads can reach a useful checkpoint or complete a meaningful unit of work
+- Allow administrators to independently configure a global minimum runtime
+  guarantee for workloads before they become eligible for cross-CQ reclaim
+  preemption (nominal quota reclaim and reclaim while borrowing)
+- Maintain backward compatibility: existing configurations without these
+  settings (i.e., both `nil`) behave exactly as before
+
+### Non-Goals
+
+- Within-ClusterQueue time-based preemption (addressed by KEP-8522)
+- Per-CQ overrides for either duration (deferred to future iterations)
+- Per-workload minimum runtime overrides
+- Workload priority that decays over time based on runtime
+- Precise CPU/GPU time accounting
+
+## Proposal
+
+Cross-CQ preemption candidates are tagged with a preemption reason:
+`InCohortFairSharing` for fair sharing rebalancing, `InCohortReclamation`
+for nominal quota reclaim, or `InCohortReclaimWhileBorrowing` for reclaim
+while borrowing. This KEP adds two optional configuration fields that
+together cover all cross-CQ preemption types:
+
+- **`fairSharing.minAdmitDuration`**: protects workloads from fair sharing
+  rebalancing preemption. A workload admitted less than this duration ago is
+  excluded from `InCohortFairSharing` preemption candidates. Lives in the
+  `FairSharing` configuration because it only applies when fair sharing is
+  enabled.
+- **`reclaimMinAdmitDuration`**: protects workloads from cross-CQ reclaim
+  preemption. A workload admitted less than this duration ago is excluded
+  from both `InCohortReclamation` and `InCohortReclaimWhileBorrowing`
+  preemption candidates. This is a top-level `Configuration` field because
+  these reclaim types apply in both classical and fair sharing preemption
+  modes.
+
+Both fields default to `nil` (no protection), and each can be set
+independently. Having separate fields lets administrators set appropriate
+thresholds for each type — for example, a longer window for fair sharing
+(where neither side has priority) and a shorter one (or none) for quota
+reclaim (where the owner has a stronger entitlement).
+
+Both fields are gated behind the `CrossCQMinAdmitDuration` feature gate.
+
+### User Stories
+
+#### Story 1: Guaranteeing Minimum Runtime for GPU Training Jobs
+
+As a cluster administrator managing GPU resources shared across multiple teams
+via ClusterQueues in a cohort, I want to ensure that when fair sharing
+preemption rebalances resources between borrowing ClusterQueues, the affected
+workloads have had enough time to make meaningful progress.
+
+By setting `minAdmitDuration: 2h` in the fair sharing configuration, workloads
+are guaranteed at least 2 hours of runtime before they can be preempted by
+another ClusterQueue rebalancing its fair share.
+
+#### Story 2: Ensuring Checkpoint Completion Before Preemption
+
+As a platform team running ML training workloads across multiple teams, I want
+workloads to have enough time to reach a checkpoint before fair sharing
+preemption can reclaim resources. This avoids wasting the compute that was
+already invested in the current training step.
+
+By setting `minAdmitDuration: 30m`, workloads get at least 30 minutes of
+uninterrupted runtime, which is typically enough time to complete a training
+step and write a checkpoint.
+
+#### Story 3: Grace Period for Nominal Quota Reclaim
+
+As a cluster administrator, I want to give borrowing workloads a short grace
+period before a ClusterQueue reclaims its own nominal quota, so that workloads
+are not killed immediately when the owner's demand increases.
+
+By setting `reclaimMinAdmitDuration: 10m`, borrowing workloads get at least
+10 minutes to make progress before they can be reclaimed. This is shorter than
+the fair sharing `minAdmitDuration` because the owner has a legitimate
+entitlement to its quota.
+
+### Notes/Constraints/Caveats (Optional)
+
+Each field filters preemption candidates based on their preemption reason.
+`fairSharing.minAdmitDuration` applies only to candidates tagged as
+`InCohortFairSharing` (fair sharing path only), while
+`reclaimMinAdmitDuration` applies to candidates tagged as
+`InCohortReclamation` or `InCohortReclaimWhileBorrowing` in both classical
+and fair sharing preemption paths. Together, the two fields cover all
+cross-CQ preemption types. Within-CQ candidates (`InClusterQueue`) are
+unaffected by either field.
+
+This feature is complementary to KEP-8522 (time-based same-priority
+preemption). KEP-8522 addresses time-sharing among workloads with identical
+priority within the same ClusterQueue, while this feature addresses minimum
+runtime guarantees for workloads across ClusterQueues. The two mechanisms
+operate independently and do not conflict.
+
+**Interaction with Fair Sharing strategies**: Workloads that have not yet
+reached their `minAdmitDuration` are removed from the preemption candidate set.
+This applies regardless of which fair sharing strategy is configured.
+
+**Re-admission**: Preempted workloads follow the existing eviction and
+re-admission path. When a workload is re-admitted, its `QuotaReserved`
+timestamp resets, restarting the applicable duration timer.
+
+### Risks and Mitigations
+
+**Duration set too low**: If either duration is set too low, workloads might
+be preempted before making meaningful progress. Mitigated by validation
+requiring a minimum of 1 minute for both fields.
+
+**Checkpoint Dependency**: Workloads that don't implement checkpointing lose
+progress when preempted. This is not new; workloads already need to handle
+preemption gracefully.
+
+**All candidates protected (fair sharing)**: If all fair sharing preemption
+candidates in a cohort are still within their `minAdmitDuration` window, fair
+sharing preemption is blocked entirely until at least one candidate's duration
+expires. This is the intended behavior — the administrator has configured a
+minimum runtime guarantee and the system respects it.
+
+**All candidates protected (reclaim)**: Similarly, if all reclaim candidates
+are within their `reclaimMinAdmitDuration` window, an owner CQ is temporarily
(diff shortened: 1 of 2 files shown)


keps/9876-cross-cq-guaranteed-runtime/README.md
```diff
@@ -0,0 +1,450 @@
...
+- **`fairSharing.minAdmitDuration`**: protects workloads from fair sharing
+  rebalancing (`InCohortFairSharing`). Lives in the `FairSharing`
+  configuration because it only applies when fair sharing is enabled.
+- **`reclaimMinAdmitDuration`**: protects workloads from cross-CQ reclaim
+  (`InCohortReclamation` and `InCohortReclaimWhileBorrowing`). This is a
+  top-level `Configuration` field because reclaim occurs in both classical
+  and fair sharing preemption modes.
+
+When configured, a workload whose `QuotaReserved` condition was set less than
+the applicable duration ago is skipped as a preemption candidate.
```
mimowo, line 51: qq why checking QuotaReserved not Admitted? Admitted seems to fit better the minAdmitDuration. The difference is in case of 2-phase admission, I'm wondering if we really should protect by this mechanism workloads which are not running yet, wdyt?

tenzen-y:
> @tenzen-y Could I get your review in this KEP? We had briefly discussed about this during KubeCon

Thank you for updating that! Sure! let me add this to my KEP review list. (but my head of review queue already occupied by another KEP)

Anyway, I can support this for Kueue v0.18.

keps/9876-cross-cq-guaranteed-runtime/kep.yaml
```diff
@@ -0,0 +1,43 @@
...
+  - "/keps/1714-fair-sharing"
+  - "/keps/83-workload-preemption"
+
+# The target maturity stage in the current dev cycle for this KEP.
+stage: alpha
+
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. This can be the current (upcoming) milestone, if it is being actively
+# worked on.
+latest-milestone: "v0.17"
```
mimowo, line 26:
```suggestion
latest-milestone: "v0.18"
```

keps/9876-cross-cq-guaranteed-runtime/kep.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  alpha: "v0.17"
+  beta: "v0.18"
+  stable: "v0.20"
```
mimowo, line 32: bump them all by 1

keps/9876-cross-cq-guaranteed-runtime/README.md
```diff
@@ -0,0 +1,450 @@
...
+
+- **Nominal quota reclaim** (`InCohortReclamation`): A ClusterQueue reclaims
+  resources it owns (its nominal quota) from other ClusterQueues that
+  borrowed them. The owner has a clear entitlement to these resources. This
+  applies regardless of whether fair sharing is enabled.
+
+- **Reclaim while borrowing** (`InCohortReclaimWhileBorrowing`): A
+  ClusterQueue that itself needs to borrow preempts workloads from other
+  borrowing ClusterQueues via the `borrowWithinCohort` policy. This applies
+  in classical preemption mode.
```
mimowo, line 76: This seems more like a background information, maybe it should go to the Notes section. It does not seem to impact the motivation in any way, the next sentence anyway says "In all cases ...". So I propose to skip it from here and consider moving to a better section.

keps/9876-cross-cq-guaranteed-runtime/README.md
```diff
@@ -0,0 +1,450 @@
...
+configuration fields that allow administrators to guarantee a minimum runtime
+before workloads become eligible for cross-CQ preemption:
+
+- **`fairSharing.minAdmitDuration`**: protects workloads from fair sharing
+  rebalancing (`InCohortFairSharing`). Lives in the `FairSharing`
+  configuration because it only applies when fair sharing is enabled.
+- **`reclaimMinAdmitDuration`**: protects workloads from cross-CQ reclaim
+  (`InCohortReclamation` and `InCohortReclaimWhileBorrowing`). This is a
+  top-level `Configuration` field because reclaim occurs in both classical
+  and fair sharing preemption modes.
```
mimowo, line 48: The Summary feels too detailed for this section, getting down to the specifics of the API shape, that should be in the Design Details section to avoid changing too many places when the API decisions change, maybe just 

This KEP introduces two configuration fields that allow administrators to guarantee a minimum runtime before workloads become eligible for cross-CQ preemption, both as a result of fair-sharing rebalancing and reclamation.

keps/9876-cross-cq-guaranteed-runtime/README.md
```diff
@@ -0,0 +1,450 @@
...
+  rebalancing preemption. A workload admitted less than this duration ago is
+  excluded from `InCohortFairSharing` preemption candidates. Lives in the
+  `FairSharing` configuration because it only applies when fair sharing is
+  enabled.
+- **`reclaimMinAdmitDuration`**: protects workloads from cross-CQ reclaim
+  preemption. A workload admitted less than this duration ago is excluded
+  from both `InCohortReclamation` and `InCohortReclaimWhileBorrowing`
+  preemption candidates. This is a top-level `Configuration` field because
+  these reclaim types apply in both classical and fair sharing preemption
+  modes.
```
mimowo, line 128: The Proposal section feels a bit too detailed about the API-specific naming. I think those details belong in Design Details, so that we do not need to update multiple sections if we later decide on a different API shape.

keps/9876-cross-cq-guaranteed-runtime/README.md
```diff
@@ -0,0 +1,450 @@
...
+independently. Having separate fields lets administrators set appropriate
+thresholds for each type — for example, a longer window for fair sharing
+(where neither side has priority) and a shorter one (or none) for quota
+reclaim (where the owner has a stronger entitlement).
+
+Both fields are gated behind the `CrossCQMinAdmitDuration` feature gate.
+
+### User Stories
+
+#### Story 1: Guaranteeing Minimum Runtime for GPU Training Jobs
```
mimowo, line 140: Actually, since `fairSharing.minAdmitDuration` is proposed to be a global setting you cannot realized both Stories 1 & 2 in the same clusters. May I get clarification which one is your intended use-case actually, so that we can focus on this one. 

Also, I would like to get clarification if you intend to configure the `fairSharing.minAdmitDuration` indeed different than reclaimMinAdmitDuration?

keps/9876-cross-cq-guaranteed-runtime/README.md
```diff
@@ -0,0 +1,450 @@
...
+
+**Re-admission**: Preempted workloads follow the existing eviction and
+re-admission path. When a workload is re-admitted, its `QuotaReserved`
+timestamp resets, restarting the applicable duration timer.
+
+### Risks and Mitigations
+
+**Duration set too low**: If either duration is set too low, workloads might
+be preempted before making meaningful progress. Mitigated by validation
+requiring a minimum of 1 minute for both fields.
```
mimowo, line 202: I'm not sure this is any new risk. Currently we don't have any knob for that, so effectively the minAdmitDuration=0s, so it fits the validation here.

Also, blocking setting the duration less than 1min may complicate writing integration and e2e tests. 

IMO this is very similar to waitForPodsReady where we don't have such validation.

keps/9876-cross-cq-guaranteed-runtime/README.md
```diff
@@ -0,0 +1,450 @@
...
+
+**Checkpoint Dependency**: Workloads that don't implement checkpointing lose
+progress when preempted. This is not new; workloads already need to handle
+preemption gracefully.
```
mimowo, line 206: Similarly, this is not a new risk, we may mention about it in Notes section. Let's focus here on the new risks introduced by the enhancement specifically.

keps/9876-cross-cq-guaranteed-runtime/README.md
```diff
@@ -0,0 +1,450 @@
...
+**All candidates protected (reclaim)**: Similarly, if all reclaim candidates
+are within their `reclaimMinAdmitDuration` window, an owner CQ is temporarily
+unable to reclaim its own quota. Administrators should set
+`reclaimMinAdmitDuration` conservatively (or leave it `nil`) to avoid
+delaying reclaim of owned resources.
+
+**Controller Restart**: All timing is based on the persisted `QuotaReserved`
+condition timestamps stored in workload status, so workloads maintain their
+accurate admission time across controller restarts. No special handling is
+required.
```
mimowo, line 223: Since this is not a risk, I would skip this point, or move to design details in some form.

keps/9876-cross-cq-guaranteed-runtime/README.md
```diff
@@ -0,0 +1,450 @@
...
+
+## Design Details
+
+### API Changes
+
+**`minAdmitDuration` in `FairSharing`**: Add a new optional field to the
+`FairSharing` struct in `apis/config/v1beta2/configuration_types.go` (and the
+corresponding v1beta1 type):
```
mimowo, line 231: nit, No need to detail the filename.

keps/9876-cross-cq-guaranteed-runtime/README.md
```diff
@@ -0,0 +1,450 @@
...
+ClusterQueues have different protection windows.
+
+**Per-workload overrides**: A per-workload `minAdmitDuration` override (e.g.,
+via an annotation) was considered but deferred. Allowing individual users to
+set their own minimum duration introduces the risk of gaming the system by
+requesting arbitrarily long protection windows. Mitigating this would require
+additional policy controls — such as administrator-defined bounds on allowed
+values or delegation to external policy engines like Kyverno — adding
+significant complexity. This may be revisited in future iterations if demand
+arises.
```
mimowo, line 325: Yes, this would be opening a pandora box. We can move the section to the Alternatives to offload the main both of the text.

keps/9876-cross-cq-guaranteed-runtime/README.md
```diff
@@ -0,0 +1,450 @@
...
+### Future Extensibility
+
+**Per-CQ reclaim overrides**: A per-CQ `reclaimMinAdmitDuration` on
+`ClusterQueuePreemption` (next to `reclaimWithinCohort`) would let each quota
+owner control how long borrowers are protected before reclaim of its resources.
+The global `reclaimMinAdmitDuration` would serve as the cluster-wide default,
+with a per-CQ override taking precedence. This is deferred to a future
+iteration to keep the initial implementation simple and to avoid making it
+difficult for administrators to reason about why workloads in different
+ClusterQueues have different protection windows.
```
mimowo, line 316: Let's move it to the Alternatives section.

keps/9876-cross-cq-guaranteed-runtime/README.md
```diff
@@ -0,0 +1,450 @@
...
+delaying reclaim of owned resources.
+
+**Controller Restart**: All timing is based on the persisted `QuotaReserved`
+condition timestamps stored in workload status, so workloads maintain their
+accurate admission time across controller restarts. No special handling is
+required.
+
+## Design Details
+
+### API Changes
```
mimowo, line 227: Actually, I'm thinking about an alternative API which groups the "preemption-protection" rules, because:
- fair sharing has a strong overlap with `preemption.reclaimWithinCohort` - if that is disabled then also candidates filtering does not work in the [findCandidates](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/scheduler/preemption/preemption.go#L578), and so no FairSharing
- the rules are grouped together rather than scattered across configuration
- it is clear where to add future extensions, for example to protect Workloads within ClusterQueue

So, I'm thinking about something like:

Option A (just grouping):
```yaml
preemptionProtectionRules:   # or just preemption
  fairSharing:
    minAdmitDuration: 1h
  reclaim:
    minAdmitDuration: 30min
```
Option B (more flexible):
```yaml
preemptionProtectionRules:
  - reason: InCohortFairSharing   
    minAdmitDuration: 1h
  - reason: InCohortReclaim
    minAdmitDuration: 30min
```
I'm also open to some tweaks in the proposed options, but I would like to know your thoughts about the direction so that we can align on the details.

mimowo (commented):
Thank you for the KEP 👍 I'm supportive for the improvement, my main consideration is about the shape of the API to make it clear and extensible for the future.

keps/9876-cross-cq-guaranteed-runtime/README.md
```diff
@@ -0,0 +1,450 @@
...
+#### Story 2: Ensuring Checkpoint Completion Before Preemption
+
+As a platform team running ML training workloads across multiple teams, I want
+workloads to have enough time to reach a checkpoint before fair sharing
+preemption can reclaim resources. This avoids wasting the compute that was
+already invested in the current training step.
+
+By setting `minAdmitDuration: 30m`, workloads get at least 30 minutes of
+uninterrupted runtime, which is typically enough time to complete a training
+step and write a checkpoint.
```
tenzen-y, line 160: Why is `gracefulTerminationPeriod` not effective for this story?
I guess that we can leverage it instead of `minAdmitDuration`.

keps/9876-cross-cq-guaranteed-runtime/README.md
tenzen-y: I agree with @mimowo.
In some cases, AdmissionCheck will take a longer time, which can consume all the minAdmitDuration budget. So, I would recommend counting the duration from the Admitted phase.

As a concrete example, Cluster Autoscaler (on top of the ProvReq AC) often takes longer to provision new capacity.

Also, in case of 2-pass scheduling mechanism (NOT 2-phase admission mechanism), the Workload can obtain QuotaReserved in 1st pass, then the Workload gets Admitted state based on topology assignment in the 2nd pass. I think that this scheduling pattern is also a problem for the current QuotaReserved condition time-based minimum runtime calculation.

keps/9876-cross-cq-guaranteed-runtime/README.md
tenzen-y: I think that both minAdmitDuration and reclaimMinAdmitDuration concepts sound reasonable because they sometimes want to specify the different preemption protection rules.

In general, borrowed resources should be returned (reclaimation) to the original cq nominal quota as soon as possible, but the own nominal quota usage could be more flexible (more protection time budget).

But, I agree that we should organize the API structure for a more comprehensive one. Let me leave my API structure proposal to APIs section.

keps/9876-cross-cq-guaranteed-runtime/README.md
tenzen-y: I agree with @mimowo, I also don't think that we should introduce 1min validation rules.

keps/9876-cross-cq-guaranteed-runtime/README.md
tenzen-y: I fully agree with @mimowo. I also want to discuss where the best place is for these `preemptionProtectionRules` APIs. I'm wondering if we can put it in the Configuration field because I don't see the strong stories to justify placing it on the ClusterQueue level. @mimowo Which API resource (Configuration vs ClusterQueue) did you assume?

Additionally, I would recommend the following refined APIs in the following.

If we want to select "The pattern of ClusterQueue" based on user stories, I'm fine with that.
But, in that case, we should introduce the guard mechanism for reclaimation to specify how long the guaranteed runtime the lender allows borrowers so that the lender can prevent starvation. This is similar to the correspondence between borrowingLimit and lendingLimit.

The pattern of Configuration:

```yaml
fairSharing:
  preemptionStrategies: [] # existing field.
  preemptionProtectionStrategies:
    minAdmissionDuration: 1h
classicPreemption:
  protectionStrategies:
    reclaimWithinCohort: 
      minAdmissionDuration: 30min
```

The pattern of ClusterQueue:

In the case of your Option A and Option B, classic and fairSharing protection rules are placed in the same category, but I want to propose the following separate pattern because we already have a fairSharing dedicated field.

```yaml
spec:
  preemptionProtectionStrategies:
    reclaimWithinCohort:
      minAdmissionDuration: 30min
  fairSharing:
    weight: "" # existing field
    fairPreemptionProtectionStrategies:
      minAdmissionDuration: 1h
# We potentially might want to dynamically reduce the min runtime based on used resources in the future.
#     resourceWeights:
#       example.com/gpu: 5
```

keps/9876-cross-cq-guaranteed-runtime/kep.yaml
```diff
@@ -0,0 +1,43 @@
...
+authors:
+  - "@mukund-wayve"
+status: provisional
+creation-date: 2026-03-15
+reviewers:
+  - "@sohankunkerkar"
+  - "@tenzen-y"
+  - "@mimowo"
+approvers:
+  - "@sohankunkerkar"
```
tenzen-y, line 12:
```suggestion
```

At this time, unfortunately, @sohankunkerkar can not approve any KEP. 
But, I really would represent appreciation @sohankunkerkar 's KEP review 🙏

tenzen-y (commented):
Thank you for working on that!
I left some comments and responses to @mimowo !

keps/9876-cross-cq-guaranteed-runtime/kep.yaml
michael-pryor: Ok - i have changed the reviewers [here](https://github.com/kubernetes-sigs/kueue/pull/12791/changes#diff-6f20f045bbb226b1850bc71f4867226cb90adabe8d970ef8cda0c20ebad76989R13) to include yourself and mimowo

keps/9876-cross-cq-guaranteed-runtime/README.md
michael-pryor: Ok I have made the summary more concise [here](https://github.com/kubernetes-sigs/kueue/pull/12791/changes#diff-33743edd88f4408f1917d5998a2992353bd667f73d528bddcaa65ddfaa40f92aR47)

keps/9876-cross-cq-guaranteed-runtime/README.md
michael-pryor: Ok - i have changed to be Admitted time and added a section about it [here](https://github.com/kubernetes-sigs/kueue/pull/12791/changes#diff-33743edd88f4408f1917d5998a2992353bd667f73d528bddcaa65ddfaa40f92aR314)

keps/9876-cross-cq-guaranteed-runtime/README.md
michael-pryor: I've moved it into the Design Section under its own header [here](https://github.com/kubernetes-sigs/kueue/pull/12791/changes#diff-33743edd88f4408f1917d5998a2992353bd667f73d528bddcaa65ddfaa40f92aR206)

keps/9876-cross-cq-guaranteed-runtime/README.md
michael-pryor: I've moved it into the Design Section under its own header [here](https://github.com/kubernetes-sigs/kueue/pull/12791/changes#diff-33743edd88f4408f1917d5998a2992353bd667f73d528bddcaa65ddfaa40f92aR350)

keps/9876-cross-cq-guaranteed-runtime/README.md
michael-pryor: I do still think that its valid to have different configs for each because one is about reclaiming ones own quota and the other is about making borrowed quota more fair. 

I have tried to improve the stories [here](https://github.com/kubernetes-sigs/kueue/pull/12791/changes#diff-33743edd88f4408f1917d5998a2992353bd667f73d528bddcaa65ddfaa40f92aR131)

keps/9876-cross-cq-guaranteed-runtime/README.md
michael-pryor: I think the gracefulTerminationPeriod is time between sigterm and more aggressive killing. The sigterm will still cause interruption immediately. I added a section [here](https://github.com/kubernetes-sigs/kueue/pull/12791/changes#diff-33743edd88f4408f1917d5998a2992353bd667f73d528bddcaa65ddfaa40f92aR672).

keps/9876-cross-cq-guaranteed-runtime/README.md
michael-pryor: Ok I have removed this - see [here](https://github.com/kubernetes-sigs/kueue/pull/12791/changes#diff-33743edd88f4408f1917d5998a2992353bd667f73d528bddcaa65ddfaa40f92aR178)

keps/9876-cross-cq-guaranteed-runtime/README.md
michael-pryor: I have also removed this from [here](https://github.com/kubernetes-sigs/kueue/pull/12791/changes#diff-33743edd88f4408f1917d5998a2992353bd667f73d528bddcaa65ddfaa40f92aR178)

keps/9876-cross-cq-guaranteed-runtime/README.md
michael-pryor: I have included this in the new Design Details / Admission Time section [here](https://github.com/kubernetes-sigs/kueue/pull/12791/changes#diff-33743edd88f4408f1917d5998a2992353bd667f73d528bddcaa65ddfaa40f92aR314)

keps/9876-cross-cq-guaranteed-runtime/README.md
michael-pryor: I have gone with option A [here](https://github.com/kubernetes-sigs/kueue/pull/12791/changes#diff-33743edd88f4408f1917d5998a2992353bd667f73d528bddcaa65ddfaa40f92aR237):

```preemptionProtection:
  fairSharing:
    minAdmitDuration: 1h
  reclaimWithinCohort:
    minAdmitDuration: 10m
```


I have also talked about option B in alternatives considered [here](https://github.com/kubernetes-sigs/kueue/pull/12791/changes#diff-33743edd88f4408f1917d5998a2992353bd667f73d528bddcaa65ddfaa40f92aR693).

keps/9876-cross-cq-guaranteed-runtime/README.md
michael-pryor: I have moved this to the alternatives section [here](https://github.com/kubernetes-sigs/kueue/pull/12791/changes#diff-33743edd88f4408f1917d5998a2992353bd667f73d528bddcaa65ddfaa40f92aR685)

keps/9876-cross-cq-guaranteed-runtime/README.md
michael-pryor: I have moved to the Alternatives section [here](https://github.com/kubernetes-sigs/kueue/pull/12791/changes#diff-33743edd88f4408f1917d5998a2992353bd667f73d528bddcaa65ddfaa40f92aR632)

keps/9876-cross-cq-guaranteed-runtime/README.md
michael-pryor: done

michael-pryor:
hi @mimowo @tenzen-y I work with Mukund directly and am taking over getting this KEP and corresponding changes implemented. I have created a new KEP [here](https://github.com/kubernetes-sigs/kueue/pull/12791/changes) based on your feedback on the original (this PR). 

Mukund will close this next week, and we can continue on the new PR.

kannon92:
> hi @mimowo @tenzen-y I work with Mukund directly and am taking over getting this KEP and corresponding changes implemented. I have created a new KEP [here](https://github.com/kubernetes-sigs/kueue/pull/12791/changes) based on your feedback on the original (this PR).
> 
> Mukund will close this next week, and we can continue on the new PR.

/close

PR #9894, opened 2026-03-16
Fix limits only issue code

### LINKED ISSUE #9889: Kueue may hit inifinite requeue loop for workloads which only specify limits without requests
**What happened**:

Kueue may hit inifinte requeue loop for workloads which only specify limits without requests.

Such workloads loop with the `Workload no longer fits after processing another workload` even if there is enough capacity

Workloads should be able to schedule as soon as there is enough capacty.

The reason is that requeueAndUpdate does not use AdjustResources which is responsible for defaulting requests based on limits

Specifying only limits is pretty popular patter for accelarator workloads.

mimowo:
#### What type of PR is this?

/kind bug

Fixes #9889

See PR reproducing the problem: https://github.com/kubernetes-sigs/kueue/pull/9893

Example failure: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/9893/pull-kueue-test-integration-baseline-main/2033626544118697984

```release-note
Scheduling: fix the issue that scheduler could indefinitely try re-queueing a workload which was once 
inadmissible, but is admissible after an update. The issue affected workloads which don't specify 
resource requests explicitly, but rely on defaulting based on limits.
```

### DIFF
--- a/test/integration/singlecluster/tas/tas_test.go
+++ b/test/integration/singlecluster/tas/tas_test.go
@@ -557,11 +557,10 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
 				clusterQueue = utiltestingapi.MakeClusterQueue("cluster-queue").
 					ResourceGroup(*utiltestingapi.MakeFlavorQuotas(tasFlavor.Name).Resource(corev1.ResourceCPU, "5").Obj()).
 					Obj()
-				util.MustCreate(ctx, k8sClient, clusterQueue)
-				util.ExpectClusterQueuesToBeActive(ctx, k8sClient, clusterQueue)
+				util.CreateClusterQueuesAndWaitForActive(ctx, k8sClient, clusterQueue)
 
 				localQueue = utiltestingapi.MakeLocalQueue("local-queue", ns.Name).ClusterQueue(clusterQueue.Name).Obj()
-				util.MustCreate(ctx, k8sClient, localQueue)
+				util.CreateLocalQueuesAndWaitForActive(ctx, k8sClient, localQueue)
 			})
 
 			ginkgo.AfterEach(func() {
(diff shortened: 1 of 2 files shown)


test/integration/singlecluster/tas/tas_test.go
```diff
@@ -576,6 +575,37 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
...
+					Queue(kueue.LocalQueueName(localQueue.Name)).PodSets(*utiltestingapi.MakePodSet("worker", 4).
+					PreferredTopologyRequest(utiltesting.DefaultBlockTopologyLevel).
+					Obj()).Request(corev1.ResourceCPU, "1").Obj()
+				util.MustCreate(ctx, k8sClient, blocker)
+
+				util.ExpectWorkloadsToHaveQuotaReservation(ctx, k8sClient, clusterQueue.Name, blocker)
+				util.ExpectPendingWorkloadsMetric(clusterQueue, 0, 0)
+
+				ginkgo.By("Creating a second workload with only GPU limits so it becomes inadmissible while the blocker holds quota")
+				limitsOnly := utiltestingapi.MakeWorkload("limits-pnly", ns.Name).
```
sohankunkerkar: `s/limits-pnly/limits-only`

test/integration/singlecluster/tas/tas_test.go
```diff
@@ -576,6 +575,37 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
...
+				blocker := utiltestingapi.MakeWorkload("blocker", ns.Name).
+					Queue(kueue.LocalQueueName(localQueue.Name)).PodSets(*utiltestingapi.MakePodSet("worker", 4).
+					PreferredTopologyRequest(utiltesting.DefaultBlockTopologyLevel).
+					Obj()).Request(corev1.ResourceCPU, "1").Obj()
+				util.MustCreate(ctx, k8sClient, blocker)
+
+				util.ExpectWorkloadsToHaveQuotaReservation(ctx, k8sClient, clusterQueue.Name, blocker)
+				util.ExpectPendingWorkloadsMetric(clusterQueue, 0, 0)
+
+				ginkgo.By("Creating a second workload with only GPU limits so it becomes inadmissible while the blocker holds quota")
```
sohankunkerkar: Not a blocker but it shouldn't use GPU in the description as it uses `corev1.ResourceCPU`

mimowo: Yeah, I was thinking about that, but the mechanics is the same, so I went for reusing the pre-existing `ginkgo.When`. I'm happy to update if you recommend so, just explaining my reasoning.

test/integration/singlecluster/tas/tas_test.go
mimowo: done

test/integration/singlecluster/tas/tas_test.go
mimowo: For now I just removed "GPU" from the ginkgo.By

Issue #9096, opened 2026-02-10
position within clusterqueues is sorted randomly


mmolisch:
**What happened**:
Workloads within cluster-queue are sorted "randomly" each watch call -> nonstop changing `positionInClusterQueue`. No other workloads, in different queues, are waiting.
**What you expected to happen**:
workload in cluster-queue are not "randomly" sorted

**How to reproduce it (as minimally and precisely as possible)**:
`kubectl get --raw /apis/visibility.kueue.x-k8s.io/v1beta2/clusterqueues/cq-liveness/pendingworkloads \
| jq -r ".items
  | sort_by(.positionInClusterQueue)
  | (\"CLUSTER_POS\tLOCAL_POS\tNAMESPACE\tWORKLOAD\"),
    (.[] | \"\(.positionInClusterQueue)\t\(.positionInLocalQueue)\t\(.metadata.namespace)\t\(.metadata.name)\")"
'`
every second the workloads are changing positions

**Anything else we need to know?**:
We have 0 nominalQuota for each resource in team's clusterQueues and resources are defined in shared clusterQueue.

Helm values:
```enableKueueViz: false
controllerManager:
  nodeSelector:
    node-role.kubernetes.io/master: "true"
  replicas: 3
  featureGates:
    - name: VisibilityOnDemand
      enabled: true
enablePrometheus: true
enableCertManager: true
metrics:
  prometheusNamespace: observability
managerConfig:
  controllerManagerConfigYaml: |-
    apiVersion: config.kueue.x-k8s.io/v1beta2
    kind: Configuration

    admissionFairSharing:
      usageHalfLifeTime: "168h"
      usageSamplingInterval: "5m"
      resourceWeights:
        nvidia.com/gpu: 2.0
    health:
      healthProbeBindAddress: :8081
    metrics:
      bindAddress: :8443
      enableClusterQueueResources: true
    webhook:
      port: 9443
    leaderElection:
      leaderElect: true
      resourceName: c1f6bfd2.kueue.x-k8s.io
    controller:
      groupKindConcurrency:
        Job.batch: 5
        Pod: 5
        Workload.kueue.x-k8s.io: 5
        LocalQueue.kueue.x-k8s.io: 1
        ClusterQueue.kueue.x-k8s.io: 1
        ResourceFlavor.kueue.x-k8s.io: 1
    clientConnection:
      qps: 100
      burst: 200
    #pprofBindAddress: :8082
    #waitForPodsReady:
[...]

PR #9899, opened 2026-03-16
Fix Snapshot() sort inconsistency with UsageBasedAdmissionFairSharing

sohankunkerkar:
#### What type of PR is this?
/kind bug

Fixes https://github.com/kubernetes-sigs/kueue/issues/9096

Snapshot() reads live AFS state (penalties, consumed resources) on every sort comparison. Concurrent scheduler updates cause the comparator to return inconsistent results within a single sort, violating transitivity.

```release-note
VisibilityOnDemand: Fix non-deterministic workload ordering with UsageBasedAdmissionFairSharing enabled.
```

### DIFF
--- a/pkg/cache/queue/cluster_queue.go
+++ b/pkg/cache/queue/cluster_queue.go
@@ -118,7 +118,8 @@ type ClusterQueue struct {
 	// QueueInadmissibleWorkloads is called.
 	queueInadmissibleCycle int64
 
-	compareFunc func(a, b *workload.Info) int
+	compareFunc  func(a, b *workload.Info) int
+	snapshotSort func(elements []*workload.Info)
 
 	queueingStrategy kueue.QueueingStrategy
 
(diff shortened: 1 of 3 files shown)


pkg/cache/queue/cluster_queue.go
```diff
@@ -560,13 +569,85 @@ func (c *ClusterQueue) DumpInadmissible() ([]workload.Reference, bool) {
...
+		}
+	}
+
+	getLQWeight := func(lqKey utilqueue.LocalQueueReference) (float64, bool) {
+		if cl == nil {
+			return 1, true
+		}
+		ns, name := utilqueue.MustParseLocalQueueReference(lqKey)
+		var lq kueue.LocalQueue
+		if err := cl.Get(context.Background(), client.ObjectKey{Namespace: ns, Name: string(name)}, &lq); err != nil {
```
mimowo: Can we avoid using Background context? Please propagate the real one.

mimowo: Also, don't silence errors. 

OTOH, I'm wondering if it is much effort to just track the LQ weigths in cache, this would avoid the need for Get.
 wdyt?

sohankunkerkar: >OTOH, I'm wondering if it is much effort to just track the LQ weigths in cache, this would avoid the need for Get.
wdyt?

I can take that as a follow up. This would require plumbing the weight through, which is a separate concern from this bug fix.

mimowo:
/approve
Leaving lgtm to @PBundyra 
/cherrypick release-0.16

PBundyra:
> Good point on the heap path. In practice I think the heap is safe today as lessFunc is only called during PushOrUpdate/Pop, which run in the scheduler goroutine sequentially with penalty Push/Sub. The penalty updates happen after Pop returns, never during a sift.

We also call `PushOrUpdate` from the [localqueue_controller here](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/core/localqueue_controller.go#L225) which is not synchronized with the scheduler goroutine. 

Secondly, the risk comes from the fact that the comparison function uses the LQ object from the controller runtime informer which we cannot prevent from mutating in the middle of heap sort. E.g. what could happen:

1. `PushOrUpdate` function is called from whatever location
2.  Heap makes some comparisons
3. There is and UPDATE call (e.g. sent by localqueue_controller) that updates the LQ's consumedResources
4. The rest of the sorting/heapify algorithm uses a different value than initially that can lead to inconsistencies

PR #9910, opened 2026-03-16
Add e2e test for StatefulSet queue label change after scale-up from zero

### LINKED ISSUE #9830: Kueue cannot handle queue label changes on StatefulSets created with replica=0
<!-- Please use this template while reporting a bug and provide as much info as possible. Not doing so may result in your bug not being addressed in a timely manner. Thanks!

If the matter is security related, please disclose it privately via https://kubernetes.io/security/
-->


**What happened**:
                                                                                                                                                                                                                                                           
When a StatefulSet is created with replicas: 0 and deliberately pointed to a non-existent queue (e.g., baz) and then updated to replicas: 1 (two-phase pattern), Kueue's mutating webhook injects kueue.x-k8s.io/pod-group-total-count as a pod template annotation, changing it from "0" to "1". This causes a controller revision mismatch (currentRevision != updateRevision) because the pod is SchedulingGated and never becomes Ready, so the StatefulSet controller never promotes the new revision.           
                                                                                                                                                                                                                                                               
In this state, patching kueue.x-k8s.io/queue-name to a valid queue (e.g., from baz to bar) has no effect. The patch succeeds on the StatefulSet object, but:
  - The pod keeps the old queue=baz label
  - The workload stays inadmissible on baz
  - Kueue's StatefulSet reconciler never triggers the suspend → delete → recreate path

After patching kueue.x-k8s.io/queue-name from baz to bar, the pod should be recreated with queue-name=bar and admitted by the bar LocalQueue — the same behavior observed when the StatefulSet is created directly with replicas: 1 (single-phase).

Prerequisites: Kubernetes cluster with Kueue installed, a LocalQueue named bar pointing to a valid ClusterQueue.

1. Create a StatefulSet with replicas: 0 and label kueue.x-k8s.io/queue-name: baz (where baz does not exist):

```
cat <<'EOF' | kubectl apply -f -
  apiVersion: apps/v1
  kind: StatefulSet
  metadata:
    name: web
    labels:
      kueue.x-k8s.io/queue-name: baz
  spec:
    serviceName: web
    replicas: 0
    selector:
      matchLabels:
        app: web
    template:
      metadata:
        labels:
          app: web
      spec:
        containers:
          - name: web
            image: nginx:latest
            ports:
              - containerPort: 80
  EOF
```

2. Verify revisions match: 
```kubectl get statefulset web -o jsonpath='currentRev={.status.currentRevision} updateRev={.status.updateRevision}'```

3. Update to replicas: 1:
```kubectl patch statefulset web --type='json' -p='[{"op":"replace","path":"/spec/replicas","value":1}]'```

4. Verify revision mismatch and pod is gated:
```
kubectl get statefulset web -o jsonpath='currentRev={.status.currentRevision} updateRev={.status.updateRevision}'
# currentRev=web-AAAAA updateRev=web-BBBBB (mismatch)

kubectl get pod web-0
# STATUS: SchedulingGated
```

5. Patch the queue label to a valid queue:
``` 
kubectl patch statefulset web --type='json' -p='[{"op":"replace","path":"/metadata/labels/kueue.x-k8s.io~1queue-name","value":"bar"}]'
```

6. Observe the bug:
```
kubectl get pod web-0 -o jsonpath='queue={.metadata.labels.kueue\.x-k8s\.io/queue-name}'
# queue=baz  ← NOT UPDATED
kubectl get workloads.kueue.x-k8s.io -o custom-columns='QUEUE:.spec.queueName'
# baz  ← STILL INADMISSIBLE
```

**Environment**:
- Kubernetes version: OpenShift 4.19.16, Kubernetes v1.32.9
- Kueue version: 0.14 
- Cloud provider or hardware configuration: OpenShift

### PR DESCRIPTION
**Pull request by kannon92**
#### What type of PR is this?

/kind bug
/area testing

Adds an e2e regression test covering the two-phase StatefulSet pattern
described in issue 9830. The test:

1. Creates a StatefulSet with `replicas=0` and a non-existent queue
2. Scales up to 1 replica (triggering webhook annotation injection and
   a controller revision mismatch)
3. Changes the queue label to a valid LocalQueue
4. Verifies the pod queue label is updated, the workload queue name is
   updated, the replica becomes ready, and the workload is admitted

Fixes #9830

This test is expected to fail until the underlying bug is fixed. It
reproduces the exact scenario from the issue where the pod keeps the
old queue label after a queue change on a StatefulSet that was scaled
up from zero replicas.

```release-note
NONE
```

---

### DIFF
--- a/test/e2e/singlecluster/statefulset_test.go
+++ b/test/e2e/singlecluster/statefulset_test.go
@@ -384,6 +384,94 @@ var _ = ginkgo.Describe("StatefulSet integration", ginkgo.Label("area:singleclus
 			})
 		})
 
+		// Regression test for https://github.com/kubernetes-sigs/kueue/issues/9830
+		// When a StatefulSet is created with replicas=0 and an invalid queue,
+		// then scaled up (causing a controller revision mismatch due to webhook
+		// annotation injection), changing the queue label should still propagate
+		// to the pods and allow admission on the new queue.
+		ginkgo.It("should allow to change queue name after scale up from zero (two-phase pattern)", func() {
+			invalidQueueName := fmt.Sprintf("%s-nonexistent", localQueueName)
+			statefulSet := statefulsettesting.MakeStatefulSet("sts", ns.Name).
+				Image(util.GetAgnHostImage(), util.BehaviorWaitForDeletion).
+				RequestAndLimit(corev1.ResourceCPU, "200m").
+				TerminationGracePeriod(1).
+				Replicas(0).
+				Queue(invalidQueueName).
+				Obj()
+
+			ginkgo.By("Create StatefulSet with replicas=0 and invalid queue", func() {
+				util.MustCreate(ctx, k8sClient, statefulSet)
+			})
+
+			ginkgo.By("Scale up to 1 replica", func() {
+				gomega.Eventually(func(g gomega.Gomega) {
+					createdStatefulSet := &appsv1.StatefulSet{}
+					g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(statefulSet), createdStatefulSet)).To(gomega.Succeed())
+					createdStatefulSet.Spec.Replicas = ptr.To[int32](1)
+					g.Expect(k8sClient.Update(ctx, createdStatefulSet)).To(gomega.Succeed())
+				}, util.Timeout, util.Interval).Should(gomega.Succeed())
+			})
+
+			ginkgo.By("Verify pod is created and scheduling-gated with invalid queue", func() {
+				gomega.Eventually(func(g gomega.Gomega) {
+					pods := &corev1.PodList{}
+					g.Expect(k8sClient.List(ctx, pods, client.InNamespace(ns.Name))).To(gomega.Succeed())
+					g.Expect(pods.Items).To(gomega.HaveLen(1))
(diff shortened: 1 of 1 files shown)


test/e2e/singlecluster/statefulset_test.go
```diff
@@ -384,6 +380,96 @@ var _ = ginkgo.Describe("StatefulSet integration", ginkgo.Label("area:singleclus
...
+					createdStatefulSet := &appsv1.StatefulSet{}
+					g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(statefulSet), createdStatefulSet)).To(gomega.Succeed())
+					g.Expect(createdStatefulSet.Status.ReadyReplicas).To(gomega.Equal(int32(1)))
+				}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
+			})
+
+			ginkgo.By("Check workload is admitted", func() {
+				createdWorkload := &kueue.Workload{}
+				gomega.Expect(k8sClient.Get(ctx, wlLookupKey, createdWorkload)).To(gomega.Succeed())
+				util.ExpectWorkloadsToBeAdmitted(ctx, k8sClient, createdWorkload)
```
mbobrovskyi:
```suggestion
				util.ExpectWorkloadsToBeAdmittedByKeys(ctx, k8sClient, wlLookupKey)
```

PR #9371, opened 2026-02-19
feat: add local queue metrics config

mykysha:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
Add configuration for the localqueue metrics visibility

#### Which issue(s) this PR fixes:

Part of #8830

#### Does this PR introduce a user-facing change?

```release-note
Observability: added LocalQueueMetrics configuration to control which metrics should be displayed.
```

Issue #8830, opened 2026-01-27
Graduate LocalQueue metrics to Beta


mimowo:
**What would you like to be added**:

We would like to move forward with the important feature.

The main blocking point is API to ensure users can deal with the metric cardinality. Adding namespace selector seems reasonable.

**Why is this needed**:

This is important feature which we observe is already in use even though it is still alpha.

**Completion requirements**:

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

PR #9935, opened 2026-03-17
cleanup: move metric reporting from workload finish function

mykysha:
#### What type of PR is this?
/kind cleanup

Move metrics reporting from Evict and Finish functions to simplify the configuration introduction in #9371

Part of #8830

```release-note
NONE
```

### DIFF
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -938,6 +938,11 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 		if !active {
 			log.V(2).Info("Workload will not be queued because the workload is not active")
 		}
+
+		if status == workload.StatusFinished && prevStatus != workload.StatusFinished {
+			r.reportFinishedWorkload(wlCopy)
+		}
+
 		// The workload could have been in the queues if we missed an event.
 		r.queues.DeleteWorkload(log, wlKey)
 		r.queues.AddFinishedWorkload(wlCopy)
@@ -1061,6 +1066,16 @@ func (r *WorkloadReconciler) notifyWatchers(oldWl, newWl *kueue.Workload) {
 	}
 	for _, w := range r.watchers {
 		w.NotifyWorkloadUpdate(oldWl, newWl)
+	}
+}
+
+func (r *WorkloadReconciler) reportFinishedWorkload(wl *kueue.Workload) {
+	priorityClassName := workload.PriorityClassName(wl)
+	cqName := ptr.Deref(wl.Status.Admission, kueue.Admission{}).ClusterQueue
+	metrics.IncrementFinishedWorkloadTotal(cqName, priorityClassName, r.customLabels.CQGet(cqName), r.roleTracker)
+	if features.Enabled(features.LocalQueueMetrics) {
+		lqRef := metrics.LQRefFromWorkload(wl)
+		metrics.IncrementLocalQueueFinishedWorkloadTotal(lqRef, priorityClassName, r.customLabels.LQGet(qutil.KeyFromWorkload(wl)), r.roleTracker)
 	}
 }
 
--- a/pkg/controller/admissionchecks/multikueue/workload.go
+++ b/pkg/controller/admissionchecks/multikueue/workload.go
@@ -365,7 +365,7 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
 		}
 
 		// finish workload and copy the status to the local one
-		return reconcile.Result{}, workload.Finish(ctx, w.client, group.local, remoteFinishedCond.Reason, remoteFinishedCond.Message, w.clock, w.roleTracker, nil)
+		return reconcile.Result{}, workload.Finish(ctx, w.client, group.local, remoteFinishedCond.Reason, remoteFinishedCond.Message, w.clock)
 	}
 
 	// 4. Handle workload eviction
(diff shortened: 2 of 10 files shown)


mszadkow:
Looks like you covered all the places for Finish and Evict.
/lgtm

pkg/controller/core/workload_controller.go
```diff
@@ -705,7 +708,11 @@ func (r *WorkloadReconciler) reconcileOnLocalQueueActiveState(ctx context.Contex
 		}
 		log.V(3).Info("Workload is evicted because the LocalQueue is stopped", "localQueue", klog.KRef(wl.Namespace, string(wl.Spec.QueueName)))
 		err := workload.Evict(ctx, r.client, r.recorder, wl, kueue.WorkloadEvictedByLocalQueueStopped, "The LocalQueue is stopped", "", r.clock, r.roleTracker, r.customLabels)
-		return true, err
+		if err != nil {
+			return true, err
+		}
+		workload.ReportEvictionMetrics(wl, kueue.WorkloadEvictedByLocalQueueStopped, "The LocalQueue is stopped", "", r.roleTracker, r.customLabels)
```
mimowo: I'm not sure I understand the approach, I was expecting changing the metrics on the event handlers for workloads, rather than post every Evict call. 

wdyt @mbobrovskyi ?

mykysha: For the Evict metrics it was not possible to do that way, as we have underlyingCause which is not easily retrieved in the handler. However, the approach was possible for the Finish function. Readjusting the PR now

pkg/controller/core/workload_controller.go
```diff
@@ -938,6 +938,11 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 		if !active {
 			log.V(2).Info("Workload will not be queued because the workload is not active")
 		}
+
+		if status == workload.StatusFinished && prevStatus != workload.StatusFinished {
+			r.reportFinishedWorkload(wlCopy)
+		}
```
mbobrovskyi, line 944: Yeah, this approach is much better.

pkg/controller/core/workload_controller.go
mimowo: hm, I thought we could read the underlying cause feom the secondary condition, no? 

For example I thought we would have the PodsReady condition where we can find details

mykysha: We can, but the reason in condition which contains underlying cause stores it in the form of ReasonDueToUnderlyingCause. So it is possible to obtain, but we would need to split the condition in question. Would that be ok?

mimowo: Can you provide an example (yaml snippet) of the relevant conditions? I thought no parsing would be involved.

mykysha: Not yaml, but from the test, as an example:
```go
metav1.Condition{
	Type:    kueue.WorkloadEvicted,
	Status:  metav1.ConditionTrue,
	Reason:  "DeactivatedDueToRequeuingLimitExceeded",
	Message: "The workload is deactivated due to exceeding the maximum number of re-queuing retries",
}
```

Here the Reason would be actual reason (Deactivated) and underlying cause (RequeuingLimitExceeded) merged together

mimowo: Is there no "PodsReady" condition wit the info?
give me all conditions

mykysha: Surely, again, conditions from the same test:
```go
metav1.Condition{
	Type:    kueue.WorkloadPodsReady,
	Status:  metav1.ConditionFalse,
	Reason:  kueue.WorkloadWaitForStart,
	Message: "Not all pods are ready or succeeded",
},
metav1.Condition{
	Type:    kueue.WorkloadQuotaReserved,
	Status:  metav1.ConditionFalse,
	Reason:  "Pending",
	Message: "The workload is deactivated due to exceeding the maximum number of re-queuing retries",
},
metav1.Condition{
	Type:    kueue.WorkloadEvicted,
	Status:  metav1.ConditionTrue,
	Reason:  "DeactivatedDueToRequeuingLimitExceeded",
	Message: "The workload is deactivated due to exceeding the maximum number of re-queuing retries",
},
metav1.Condition{
	Type:    kueue.WorkloadAdmitted,
	Status:  metav1.ConditionFalse,
	Reason:  "NoReservation",
	Message: "The workload has no reservation",
},
metav1.Condition{
	Type:    podcontroller.WorkloadWaitingForReplacementPods,
	Status:  metav1.ConditionTrue,
	Reason:  "DeactivatedDueToRequeuingLimitExceeded",
	Message: "The workload is deactivated due to exceeding the maximum number of re-queuing retries",
}
```
Here, underlyingCause being RequeuingLimitExceeded is not present separately

mimowo: Thank you, ok, so I see in that case the underlyingCause is `RequeuingLimitExceeded`, not `WaitForStart` as I initially thought based on metrics: https://github.com/kubernetes-sigs/kueue/blob/612c05549532ebe17d1b7b9c098f798064a497a3/pkg/metrics/metrics.go#L405

mimowo: Ahhh, I see, this is because we don't have the "Deactivated" condition. I think we should have it unlimately.

However, as we don't have it currently I think it is ok to strip the prefix `DeactivatedDueTo` and use the suffix as the underlying cause in this case. Not very pretty, but I think it's better than duplicating the code.

mykysha: I suppose it might be better to split by DueTo once, with the first part being the reason and the second underlying cause, not to hard-fix on the Deactivated field

mimowo: sgtm, just split by the first DueTo, to make sure not to split if the reason has DueTo inside

mimowo: Ok, let's move refactor of Evict to separete PR

mimowo: This PR already looks good

PR #9943, opened 2026-03-17
cleanup: move workload metrics from evict function

mykysha:
#### What type of PR is this?
/kind cleanup

Move workload metrics out of Evict function

Part of #8830

```release-note
NONE
```

### DIFF
--- a/pkg/workload/workload.go
+++ b/pkg/workload/workload.go
@@ -32,7 +32,6 @@ import (
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
-	"k8s.io/client-go/tools/record"
 	resourcehelpers "k8s.io/component-helpers/resource"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/clock"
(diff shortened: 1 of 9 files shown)


pkg/workload/workload_test.go
```diff
@@ -2628,3 +2628,53 @@ func TestSchedulingHash(t *testing.T) {
...
+			wantReason:      "",
+			wantCause:       kueue.WorkloadDeactivated,
+		},
+		"reason with cause": {
+			reasonWithCause: ReasonWithCause(kueue.WorkloadEvicted, kueue.WorkloadDeactivated),
+			wantReason:      kueue.WorkloadEvicted,
+			wantCause:       kueue.WorkloadDeactivated,
+		},
+		"reason with cause containing DueTo": {
+			reasonWithCause: ReasonWithCause(kueue.WorkloadEvicted, ReasonWithCause(kueue.WorkloadDeactivated, kueue.WorkloadRequeuingLimitExceeded)),
```
mbobrovskyi: Let’s write it directly. I think it will be easier to see what we have.

pkg/workload/workload.go
```diff
@@ -771,7 +771,7 @@ func QueuedWaitTime(wl *kueue.Workload, clock clock.Clock) time.Duration {
 
 // workloadsWithPodsReadyToEvictedTime is the amount of time it takes a workload's pods running to getting evicted.
```
mbobrovskyi:
```suggestion
// WorkloadsWithPodsReadyToEvictedTime is the amount of time it takes a workload's pods running to getting evicted.
```

pkg/workload/workload.go
```diff
@@ -1524,10 +1523,7 @@ func Evict(ctx context.Context, c client.Client, recorder record.EventRecorder,
 		log.V(3).Info("WARNING: unexpected eviction of workload without status.Admission", "workload", klog.KObj(wl))
 		return nil
 	}
-	reportEvictedWorkload(recorder, wl, wl.Status.Admission.ClusterQueue, reason, msg, underlyingCause, tracker, cl)
-	if reportWorkloadEvictedOnce {
-		metrics.ReportEvictedWorkloadsOnce(wl.Status.Admission.ClusterQueue, reason, string(underlyingCause), PriorityClassName(wl), cl.CQGet(wl.Status.Admission.ClusterQueue), tracker)
-	}
+	recordEvictedWorkloadEvent(recorder, wl, reason, msg, underlyingCause)
```
mbobrovskyi: I'm wondering if we can move this as well. WDYT?

pkg/workload/workload.go
mykysha: Updated

pkg/workload/workload_test.go
mykysha: Changed the tests here

mimowo:
/close
We synced on a meeting that this is not needed for now. We will re-open if we decide it is a good idea.

PR #9944, opened 2026-03-17
set log level via helm

### LINKED ISSUE #9448: Allow configuring Kueue's controller loglevel in the helm chart
An ability to change the desired loglevel for the Kueue controller while using the helm chart.

**Why is this needed**:
Users would like an ability to configure the loglevel in the helm chart as otherwise it requires a manual kubectl intervention.

**Completion requirements**:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

gabesaba:
#### What type of PR is this?
/kind feature

Fixes #9448

```release-note
Helm: Allow setting log level
```


mimowo:
Let's do it, we did that for many other helm chart improvements

PR #9958, opened 2026-03-17
Add podset-replica-sizes annotation on RayService

hiboyang:
#### What type of PR is this?

/kind feature

This is a follow-up PR for previous PR [Add podset-replica-sizes annotation on RayJob](https://github.com/kubernetes-sigs/kueue/pull/9726). Implement similar logic for RayService. Also refactor code for better code reuse.

```release-note
NONE
```

### DIFF
--- a/pkg/controller/jobframework/reconciler.go
+++ b/pkg/controller/jobframework/reconciler.go
@@ -71,12 +71,6 @@ import (
 const (
 	FailedToStartFinishedReason = "FailedToStart"
 	managedOwnersChainLimit     = 10
-
-	// PodsetReplicaSizesAnnotation is set on the job when autoscaling causes
-	// PodSet replica sizes to differ from the original spec. The value is a JSON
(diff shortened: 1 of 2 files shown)


pkg/controller/jobframework/reconciler.go
```diff
@@ -88,6 +88,13 @@ var (
 	ErrPrebuiltWorkloadNotFound       = errors.New("prebuilt workload not found")
 )
 
+// PodSetReplicaSize is a minimal representation of a PodSet for the
+// PodsetReplicaSizesAnnotation, containing only name and count.
+type PodSetReplicaSize struct {
+	Name  kueue.PodSetReference `json:"name"`
+	Count int32                 `json:"count"`
+}
```
mimowo: nit, but let's move the struct to the utils - closer to where it is used

hiboyang: Good suggestion, updated code

Issue #8522, opened 2026-01-10
[Time Sharing]  Support Time-Sharing in ClusterQueue


dddwsd:
**What would you like to be added**:

A same-priority preemption mechanism based on a minimum admitted runtime (minAdmitDuration) to prevent starvation and improve fairness among workloads in a ClusterQueue.

Scenario
* A ClusterQueue has a fixed quota of 8 GPUs.
*  Three users submit workloads A, B, and C with:
    * the same ClusterQueue and WorkloadPriorityClass
    * unknown execution durations
    * each workload requests 8 GPUs
* Workload A is admitted first and starts running.
* Shortly after, workloads B and C are queued.

Current behavior (without this proposal)
* Workload A continues running until completion.
* Workloads B and C remain pending, potentially for hours or days.
* Even though all workloads have the same priority, A monopolizes the ClusterQueue quota, causing excessive starvation.

Behavior with minAdmitDuration-based same-priority preemption
* Workload A is guaranteed to run for at least minAdmitDuration = 30 minutes after being admitted.
* After 30 minutes, if workloads B or C are still pending:
* Kueue may preempt workload A
* A is re-queued to the back of the ClusterQueue
* Workload B is then admitted and runs for at least 30 minutes.
* The same process repeats for C, enabling round-robin–style fairness at the admission level.

**Why is this needed**:

In environments where workloads of the same priority with unknown execution durations are continuously submitted by multiple users, this mechanism ensures fair resource allocation while guaranteeing a minimum resource holding time so that workloads are not preempted too aggressively.

**Completion requirements**:

This enhancement requires the following artifacts:

- [v] Design doc
- [v] API change
- [v] Docs update

The artifacts should be linked in subsequent comments.

sohankunkerkar:
Yeah, this looks like a valid feature request. I looked into this. Today, withinClusterQueue: `LowerOrNewerEqualPriority` allows older waiting workloads to preempt newer admitted ones at the same priority, and Admission Fair Sharing provides usage-based ordering but operates at the LocalQueue level rather than individual workloads. Neither provides a minimum runtime guarantee before a workload becomes preemptible. I'm guessing this matters for workloads that checkpoint periodically and need minimum runtime to save progress or interactive sessions like Jupyter notebooks that need guaranteed allocation time.

I'm happy to drive this feature if @mimowo or @tenzen-y  feels okay with it.

mimowo:
[...]

Issue #7990, opened 2025-11-28
Add a mechanism to consider preemption cost when finding preemption candidates


mwysokin:
**What would you like to be added**:

A mechanism to consider preemption cost when looking for and sorting candidates for preemption. The field would be set by some external controller (with a referential implementation of such controller included as part of the feature). Kueue should include a signal from that controller and should try to minimize the preemption cost.

Something that we need some feedback for would be whether the preemption cost parameter should be used only to sort/compare workloads with the same priority (e.g. 3 workloads A, B, C have workload priority medium, but A: has high preemption cost, B has medium preemption cost and C has low preemption cost, in this case the order of considering them for preemption should be: C, B, A) or whether it should be a combination of workload priority and the preemption cost (e.g. when having a low priority workload D with a critical preemption cost and a medium priority workload E having low preemption cost the order of preemptions should be: E, D).

**Why is this needed**:

Cost of preemption is not equal across pods. Pods which have a relatively long initialization time (minutes) are cheap to preempt if the initialization didn't finish but expensive to preempt when they already started carrying out some work. The same applies to checkpointing. If a pod carried out some work already but hasn't created a checkpoint yet it's costly to preempt it and its effort would be wasted, meanwhile a pod that just finished checkpointing is relatively cheap to preempt.

Since access to accelerators is so expensive and the accelerators are so scarce I think Kueue should try to care about the preemption cost and should use additional signals when looking for and sorting candidates for preemptions.

Also I don't think that Kueue itself should include some well defined policies how to calculate the cost of preemption but rather should allow external controllers to populate this information and Kueue should consider it during identifying and sorting the preemption candidates. We could offer a reference implementation of such controller that'd be a good example for folks trying to implement their own accordingly to the policies at their organizations and their business goals. I think a good referential example would be a controller that increases the preemption cost for every preemption that already happened. We store that number in the `schedulingStats` (thanks @PBundyra !) so it should be easy to use and react to this field to calculate the referential preemption cost field.

**Completion requirements**:

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

mwysokin:
CC @mimowo @tenzen-y @kannon92 @mwielgus @PBundyra @gabesaba @pajakd @amy @ichekrygin

sanposhiho:
This would be really a valuable feature for users to extend the preemptions.

---

[...]

PR #9959, opened 2026-03-17
Kueue priority booster

vladikkuzn:
#### What type of PR is this?
/kind feature

Part of #7990 #8522

```release-note
Examples: Added a priority booster controller example that adjusts Workload priority
using the `kueue.x-k8s.io/priority-boost` annotation, supporting time sharing in a
ClusterQueue between Workloads with the same priority.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,11 +40,13 @@ IMAGE_REPO := $(IMAGE_REGISTRY)/kueue
 IMAGE_REPO_KUEUEVIZ_BACKEND := $(IMAGE_REGISTRY)/kueueviz-backend
 IMAGE_REPO_KUEUEVIZ_FRONTEND := $(IMAGE_REGISTRY)/kueueviz-frontend
 IMAGE_REPO_KUEUE_POPULATOR := $(IMAGE_REGISTRY)/kueue-populator
+IMAGE_REPO_KUEUE_PRIORITY_BOOSTER := $(IMAGE_REGISTRY)/kueue-priority-booster
 
 IMAGE_TAG := $(IMAGE_REPO):$(GIT_TAG)
 IMAGE_TAG_KUEUEVIZ_BACKEND := $(IMAGE_REPO_KUEUEVIZ_BACKEND):$(GIT_TAG)
 IMAGE_TAG_KUEUEVIZ_FRONTEND := $(IMAGE_REPO_KUEUEVIZ_FRONTEND):$(GIT_TAG)
 IMAGE_TAG_KUEUE_POPULATOR := $(IMAGE_REPO_KUEUE_POPULATOR):$(GIT_TAG)
+IMAGE_TAG_KUEUE_PRIORITY_BOOSTER := $(IMAGE_REPO_KUEUE_PRIORITY_BOOSTER):$(GIT_TAG)
 
 CLUSTERPROFILE_PLUGIN_IMAGE_VERSION ?= 0.0.1
 
@@ -123,6 +125,7 @@ include Makefile-deps.mk
 include Makefile-test.mk
 
 include Makefile-kueue-populator.mk
+include Makefile-kueue-priority-booster.mk
 
 # Repo-wide verification is defined in a separate fragment so it can be read/maintained
 # independently of build/test logic. See `Makefile-verify.mk` for what `make verify` runs.
@@ -285,7 +288,9 @@ clean-manifests = \
   		$(KUSTOMIZE) edit set image backend=$(STAGING_IMAGE_REGISTRY)/kueueviz-backend:$(RELEASE_BRANCH) && \
   		$(KUSTOMIZE) edit set image frontend=$(STAGING_IMAGE_REGISTRY)/kueueviz-frontend:$(RELEASE_BRANCH)) && \
 	(cd cmd/experimental/kueue-populator/config && \
-    	$(KUSTOMIZE) edit set image controller=$(STAGING_IMAGE_REGISTRY)/kueue-populator:$(RELEASE_BRANCH))
+    	$(KUSTOMIZE) edit set image controller=$(STAGING_IMAGE_REGISTRY)/kueue-populator:$(RELEASE_BRANCH)) && \
+	(cd cmd/experimental/kueue-priority-booster/config && \
+    	$(KUSTOMIZE) edit set image controller=$(STAGING_IMAGE_REGISTRY)/kueue-priority-booster:$(RELEASE_BRANCH))
 
 .PHONY: install
 install: compile-crd-manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
@@ -327,6 +332,7 @@ prepare-manifests:
 	cd config/components/kueueviz && $(KUSTOMIZE) edit set image backend=$(IMAGE_TAG_KUEUEVIZ_BACKEND)
 	cd config/components/kueueviz && $(KUSTOMIZE) edit set image frontend=$(IMAGE_TAG_KUEUEVIZ_FRONTEND)
 	cd cmd/experimental/kueue-populator/config && $(KUSTOMIZE) edit set image controller=$(IMAGE_TAG_KUEUE_POPULATOR)
+	cd cmd/experimental/kueue-priority-booster/config && $(KUSTOMIZE) edit set image controller=$(IMAGE_TAG_KUEUE_PRIORITY_BOOSTER)
 
 .PHONY: artifacts
 artifacts: DEST_CHART_DIR="$(ARTIFACTS)"
@@ -338,6 +344,7 @@ artifacts: clean-artifacts kustomize helm-chart-package prepare-manifests ## Gen
 	$(KUSTOMIZE) build config/visibility-apf -o $(ARTIFACTS)/visibility-apf.yaml
 	$(KUSTOMIZE) build config/kueueviz -o $(ARTIFACTS)/kueueviz.yaml
 	$(KUSTOMIZE) build cmd/experimental/kueue-populator/config -o $(ARTIFACTS)/kueue-populator.yaml
+	$(KUSTOMIZE) build cmd/experimental/kueue-priority-booster/config -o $(ARTIFACTS)/kueue-priority-booster.yaml
 	@$(call clean-manifests)
 	CGO_ENABLED=$(CGO_ENABLED) GO_CMD="$(GO_CMD)" LD_FLAGS="$(LD_FLAGS)" BUILD_PATH="$(ARTIFACTS)" BUILD_NAME=kubectl-kueue PLATFORMS="$(CLI_PLATFORMS)" ./hack/multiplatform-build.sh ./cmd/kueuectl/main.go
 
@@ -361,6 +368,11 @@ prepare-release-branch: yq kustomize ## Prepare the release branch with the rele
 	$(YQ) e '.appVersion = "$(RELEASE_VERSION)" | .version = "$(APP_VERSION)" | .dependencies[0].version = "~$(APP_VERSION)"' -i cmd/experimental/kueue-populator/charts/kueue-populator/Chart.yaml
 	$(YQ) e '.kueuePopulator.image.tag = "$(RELEASE_BRANCH)"' -i cmd/experimental/kueue-populator/charts/kueue-populator/values.yaml
 	$(SED) -r 's/[0-9]+\.[0-9]+\.[0-9]+/$(APP_VERSION)/g' -i cmd/experimental/kueue-populator/README.md -i cmd/experimental/kueue-populator/charts/kueue-populator/README.md
+	# Update kueue-priority-booster chart version and image tag
+	$(YQ) e '.appVersion = "$(RELEASE_VERSION)" | .version = "$(APP_VERSION)" | .dependencies[0].version = "~$(APP_VERSION)"' -i cmd/experimental/kueue-priority-booster/charts/kueue-priority-booster/Chart.yaml
+	$(YQ) e '.kueuePriorityBooster.image.tag = "$(RELEASE_BRANCH)"' -i cmd/experimental/kueue-priority-booster/charts/kueue-priority-booster/values.yaml
+	$(SED) -r 's/[0-9]+\.[0-9]+\.[0-9]+/$(APP_VERSION)/g' -i cmd/experimental/kueue-priority-booster/README.md
+
 	$(MAKE) generate-helm-docs
 
 .PHONY: update-security-insights
@@ -467,6 +479,29 @@ kueue-populator-image-push: kueue-populator-image-build
 kueue-populator-image: PLATFORMS=$(HOST_IMAGE_PLATFORM)
 kueue-populator-image: PUSH=--load
 kueue-populator-image: kueue-populator-image-build
+
+# Build the kueue-priority-booster image
+.PHONY: kueue-priority-booster-image-build
+kueue-priority-booster-image-build:
+	$(MAKE) -C cmd/experimental/kueue-priority-booster image-build \
+		IMAGE_REGISTRY=$(IMAGE_REGISTRY) \
+		IMAGE_TAG=$(IMAGE_TAG_KUEUE_PRIORITY_BOOSTER) \
+		PLATFORMS="$(PLATFORMS)" \
+		BASE_IMAGE=$(BASE_IMAGE) \
+		BUILDER_IMAGE=$(BUILDER_IMAGE) \
+		CGO_ENABLED=$(CGO_ENABLED) \
+		PUSH=$(PUSH) \
+		IMAGE_BUILD_EXTRA_OPTS="$(IMAGE_BUILD_EXTRA_OPTS) -t $(IMAGE_REPO_KUEUE_PRIORITY_BOOSTER):$(RELEASE_BRANCH)"
+
+.PHONY: kueue-priority-booster-image-push
+kueue-priority-booster-image-push: PUSH=--push
+kueue-priority-booster-image-push: kueue-priority-booster-image-build
+
+# Build a docker local us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue-priority-booster image
+.PHONY: kueue-priority-booster-image
+kueue-priority-booster-image: PLATFORMS=$(HOST_IMAGE_PLATFORM)
+kueue-priority-booster-image: PUSH=--load
+kueue-priority-booster-image: kueue-priority-booster-image-build
 
 .PHONY: kueuectl
 kueuectl:
--- a/cmd/experimental/README.md
+++ b/cmd/experimental/README.md
@@ -12,6 +12,7 @@ where they will be maintained long term.
 ### Existing integrations
 
 - `kueue-populator`
+- `kueue-priority-booster`
 
 ### Adding new integrations
 
--- a/cmd/experimental/kueue-priority-booster/README.md
+++ b/cmd/experimental/kueue-priority-booster/README.md
@@ -1,0 +1,121 @@
+# Kueue priority booster
+
+The `kueue-priority-booster` is an experimental controller that implements a
+time-sharing fairness mechanism for Kueue Workloads by decreasing their
+effective priority after a configurable amount of time using the
+`kueue.x-k8s.io/priority-boost` annotation. The decrease can lead to preemption
+issued by pending Workloads with the same priority. See KEP-7990 for the design
+rationale.
+
+Concretely: during an initial admit window the controller leaves
+`kueue.x-k8s.io/priority-boost` **unset**, then applies a **negative** boost
+after that window while the workload remains admitted, so same–base-priority
+waiters can preempt under **`withinClusterQueue: LowerPriority`**.
+
+## Purpose
+
+In a `ClusterQueue` with a fixed quota, multiple workloads of the same priority
+and unknown duration can starve peers. With **LowerPriority**, workloads with the
+same effective priority do not preempt each other. During `timeSharingWindow` the
+controller leaves the annotation unset, so the admitted workload keeps its base
+effective priority. After the window, the controller sets a negative annotation
+so same–base-priority pending workloads can preempt, producing round-robin style
+fairness at admission time.
+
+**Why not `LowerOrNewerEqualPriority` or `Any`?** Those policies allow preemption
+against equal or higher effective priority in ways that do not match this
+“defer signal until after a minimum admit time” mechanism. This controller is
+documented for **`withinClusterQueue: LowerPriority`**.
+
+This component demonstrates an external policy controller for KEP-7990 (priority
+boost signal). It is not part of the main Kueue binary and is deployed separately.
+
+Treat it as a **plugin-style integration**: only one controller should reconcile
+`kueue.x-k8s.io/priority-boost` for a given cluster; additional controllers would
+fight over the same annotation.
+
+**Requires**: the `PriorityBoost` feature gate enabled in Kueue, and the
+`ClusterQueue` must use `withinClusterQueue: LowerPriority`.
+
+## How It Works
+
+```
+Workload admitted
+      │
+      ▼
+no annotation               ← same effective priority as peers; no same-priority
+  (timeSharingWindow)            preemption under LowerPriority
+      │
+  [window elapses, still admitted]
+      │
+      ▼
+annotation = -negativeBoostValue    ← post-window preemption signal (magnitude = negativeBoostValue)
+      │
+  [workload no longer admitted]
+      │
+      ▼
+annotation cleared
+```
+
+`EffectivePriority = basePriority + boost`. With no annotation, boost is 0.
+After the window, boost is `-negativeBoostValue`, lowering effective priority so peers
+can preempt.
+
+## Build
+
+This directory is part of the root `sigs.k8s.io/kueue` Go module (same layout as `kueue-populator`); there is no separate `go.mod` here.
+
+```bash
+make build
+```
+
+```bash
+make test
+```
+
+Unit tests only (`./pkg/...`). Integration tests use envtest and Ginkgo:
+
+```bash
+make test-integration
+```
+
+Requires compiled CRDs from repo root first: `make compile-crd-manifests` (the `test-integration` target runs this automatically).
+
+```bash
+make image-build
+```
+
+## Deploy
+
+```bash
+kubectl apply -k config
+```
+
+## Configuration
+
+The controller reads `--config` YAML:
+
+| Field | Type | Default (in code) | Description |
+|-------|------|-------------------|-------------|
+| `timeSharing.duration` | duration | `"0"` | How long after admission to leave **no** annotation. After that, while still admitted, set **negative** boost. `"0"` disables the controller behavior. |
+| `timeSharing.negativeBoost` | int32 | `100000` | Magnitude of the **negative** boost after the window (stored as `-timeSharing.negativeBoost` in the annotation). |
+| `workloadSelector` | `LabelSelector` | omit | If set, only workloads whose labels match are managed; others have any boost annotation **removed**. |
+| `maxWorkloadPriority` | int32 | omit | If set, workloads with `spec.priority` **greater** than this are out of scope (unset `spec.priority` is treated as `0`). |
+
+Example manifests: `config/manager/controller_config_map.yaml`.
+
+### Example
+
+```yaml
+timeSharing:
+  duration: "30m"
+  negativeBoost: 100000
+# workloadSelector:
+#   matchLabels:
+#     tier: batch
+# maxWorkloadPriority: 1000
+```
+
+## Helm
+
+Chart path: `charts/kueue-priority-booster`
--- a/cmd/experimental/kueue-priority-booster/charts/kueue-priority-booster/templates/manager.yaml
+++ b/cmd/experimental/kueue-priority-booster/charts/kueue-priority-booster/templates/manager.yaml
@@ -1,0 +1,131 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Release.Name }}-kueue-priority-booster
+  namespace: {{ .Release.Namespace }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ .Release.Name }}-kueue-priority-booster-role
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - events
+  verbs:
+  - create
+  - patch
+- apiGroups:
+  - kueue.x-k8s.io
+  resources:
+  - workloads
+  verbs:
+  - get
+  - list
+  - watch
+  - patch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ .Release.Name }}-kueue-priority-booster-role-binding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ .Release.Name }}-kueue-priority-booster-role
+subjects:
+- kind: ServiceAccount
+  name: {{ .Release.Name }}-kueue-priority-booster
+  namespace: {{ .Release.Namespace }}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Release.Name }}-kueue-priority-booster-config
+  namespace: {{ .Release.Namespace }}
+data:
+  controller_manager_config.yaml: |
+    timeSharing:
+      interval: {{ .Values.kueuePriorityBooster.config.timeSharing.interval | quote }}
+      negativeBoost: {{ .Values.kueuePriorityBooster.config.timeSharing.negativeBoost }}
+{{- if .Values.kueuePriorityBooster.config.workloadSelector }}
+    workloadSelector:
+{{ toYaml .Values.kueuePriorityBooster.config.workloadSelector | nindent 6 }}
+{{- end }}
+{{- if hasKey .Values.kueuePriorityBooster.config "maxWorkloadPriority" }}
+    maxWorkloadPriority: {{ index .Values.kueuePriorityBooster.config "maxWorkloadPriority" }}
+{{- end }}
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ .Release.Name }}-kueue-priority-booster
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: kueue-priority-booster
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    control-plane: kueue-priority-booster
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      control-plane: kueue-priority-booster
+      app.kubernetes.io/name: kueue-priority-booster
+      app.kubernetes.io/instance: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
+        control-plane: kueue-priority-booster
+        app.kubernetes.io/name: kueue-priority-booster
+        app.kubernetes.io/instance: {{ .Release.Name }}
+    spec:
+      serviceAccountName: {{ .Release.Name }}-kueue-priority-booster
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
+      containers:
+      - args:
+        - "--config=/etc/kueue/controller_manager_config.yaml"
+        - "--zap-log-level=2"
+        image: "{{ .Values.kueuePriorityBooster.image.repository }}:{{ .Values.kueuePriorityBooster.image.tag }}"
+        imagePullPolicy: {{ .Values.kueuePriorityBooster.image.pullPolicy }}
+        name: manager
+        securityContext:
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+          capabilities:
+            drop:
+              - ALL
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 8081
+          initialDelaySeconds: 15
+          periodSeconds: 20
+        readinessProbe:
+          httpGet:
+            path: /readyz
+            port: 8081
+          initialDelaySeconds: 5
+          periodSeconds: 10
+        resources:
+          limits:
+            cpu: 100m
+            memory: 100Mi
+          requests:
+            cpu: 100m
+            memory: 100Mi
+        volumeMounts:
+        - mountPath: /etc/kueue
+          name: manager-config
+      {{- with .Values.kueuePriorityBooster.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      terminationGracePeriodSeconds: 10
+      volumes:
+      - configMap:
+          name: {{ .Release.Name }}-kueue-priority-booster-config
+        name: manager-config
--- a/cmd/experimental/kueue-priority-booster/config/manager/controller_config_map.yaml
+++ b/cmd/experimental/kueue-priority-booster/config/manager/controller_config_map.yaml
@@ -1,0 +1,18 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kueue-priority-booster-config
+  namespace: kueue-system
+data:
+  controller_manager_config.yaml: |
+    # timeSharing groups parameters of the time-sharing mode:
+    #   interval:      no annotation while admitted for this interval; after that,
+    #                  a negative priority-boost is set while still admitted
+    #                  (LowerPriority preemption). Set to "0" to disable.
+    #                  Examples: "30m", "1h".
+    #   negativeBoost: magnitude of the negative boost after the interval
+    #                  (annotation value is -negativeBoost).
+    timeSharing:
+      interval: "30m"
+      negativeBoost: 100000
+    # Optional workloadSelector (metav1.LabelSelector) and maxWorkloadPriority — see README.
--- a/cmd/experimental/kueue-priority-booster/pkg/config/config.go
+++ b/cmd/experimental/kueue-priority-booster/pkg/config/config.go
@@ -1,0 +1,121 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+import (
+	"fmt"
+	"os"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/yaml"
+)
+
+// TimeSharing groups the parameters that control the time-sharing mode
+// (no annotation while admitted for Interval, then a negative boost of
+// magnitude NegativeBoost). Grouping leaves room for future modes of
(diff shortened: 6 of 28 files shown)


pkg/controller/jobframework/interface.go
vladikkuzn, line 1: revert

pkg/scheduler/scheduler.go
vladikkuzn, line 1: revert

cmd/experimental/priority-boost-controller/charts/priority-boost-controller/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v2
+name: priority-boost-controller
+description: A Helm chart for deploying the experimental Kueue priority-boost controller.
+type: application
+version: 0.16.2
+appVersion: "v0.16.2"
+dependencies:
+  - name: kueue
+    version: "~0.16.2"
```
vladikkuzn:
```suggestion
version: 0.17.0
appVersion: "v0.17.0"
dependencies:
  - name: kueue
    version: "~0.17.0"
```

cmd/experimental/kueue-priority-booster/pkg/controller/controller.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// computeBoost returns the boost value to apply to the workload and how long
+// until the controller should re-evaluate.
+//
+// When minAdmitDuration is configured and the workload has been admitted for
+// less than that duration, the boost is set to boostValue so that same-priority
+// pending workloads cannot preempt this workload. Once the window expires, the
+// boost is removed (0) and requeueAfter is 0.
+func (r *PriorityBoostReconciler) computeBoost(wl *kueue.Workload) (boost int32, requeueAfter time.Duration) {
```
Ladicle, line 265: WDYT about adding a feature of filtering out the target workload? In our case, we'd like to enable same-priority preemption only for workloads with low/lowest priority. 
This might be out of scope for the example controller, but it'd be helpful if it could be configured to filter pods by priority. (or using label selector for more genetic use-cases)

cmd/experimental/kueue-priority-booster/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+This component demonstrates an external policy controller for KEP-7990
+(priority boost signal). It is not part of the main Kueue binary and is
+intended to be built and deployed independently.
+
+**Requires**: the `PriorityBoost` feature gate must be enabled in Kueue, and
+the ClusterQueue must have `withinClusterQueue: LowerOrNewerEqualPriority` (or
+`Any`) preemption configured.
+
+## How It Works
```
Ladicle, line 40: I'm a bit worried that the workload could become flaky if it's preempted before the race between the Kueue and this controller boosts it. Would `withinClusterQueue: LowerPriority` and applying a negative boost after the `minDuration` window expires be more stable?

vladikkuzn: Yeah, current implementation only solve the problem of some workload being not preempted during minDuration. But that doesn't mean controller cannot choose it to stay more. To have some fairness here(and round robin between workloads), I can propose the following:
controller manages three states for each admitted workload via the `kueue.x-k8s.io/priority-boost` annotation:

| Phase | Condition | Boost Value | Effect |
|-------|-----------|-------------|--------|
| **Protected** | Admitted < minAdmitDuration | `+boostValue` (e.g. +100000) | Cannot be preempted by same-priority peers (`effectivePriority = base + boost`) |
| **Preemptable** | Admitted >= minAdmitDuration | `-boostValue` (e.g. -100000) | Lower effective priority than pending peers; CAN be preempted via `LowerPriority` policy |
| **Neutral** | Not admitted (pending/evicted) | annotation removed | Normal priority in queue ordering |

### Why Three Phases Are Necessary

A naive two-phase approach (boost during window, remove after) fails because:

1. **`LowerOrNewerEqualPriority`** uses `CreationTimestamp` for ordering (`workload.go:1263`). A workload created before its peers can never be preempted by them via this policy, regardless of runtime.
2. **`Any`** policy bypasses priority entirely (`preemption_policy.go:49`), so positive boost has no protective effect during the window.
3. **`LowerPriority`** requires strictly higher effective priority — if boost is simply removed (both at 0), same-priority workloads can't preempt each other.

The **negative boost** after window expiry solves this: it makes the workload appear as `effectivePriority = base - boostValue`, which is strictly lower than any same-base-priority pending workload (`effectivePriority = base`), enabling preemption via `LowerPriority`.

### Prerequisite ClusterQueue Configuration

```yaml
kind: ClusterQueue
spec:
  preemption:
    withinClusterQueue: LowerPriority  # required for negative boost to trigger preemption
```

The `PriorityBoost` feature gate must be enabled in Kueue.

### End-to-End Round-Robin Trace

Setup: CQ with 8 GPU quota, `withinClusterQueue: LowerPriority`, `basePriority = 0` for all workloads.

1. **T=0**: A created, admitted. Controller sets `boost = +100000` → `effectivePriority = 100000`
2. **T=1**: B created, queued. No annotation → `effectivePriority = 0`
3. **T=2**: C created, queued. No annotation → `effectivePriority = 0`
4. **T=30m**: Window expires. Controller sets `boost = -100000` on A → `effectivePriority = -100000`
5. **Scheduler**: B (priority 0) wants to preempt A (priority -100000). `SatisfiesPreemptionPolicy(B, A, LowerPriority)`: `0 > -100000 = true` → **B preempts A**
6. **A re-queued**: Controller clears boost (not admitted). A returns to queue at `effectivePriority = 0`.
7. **T=31m**: B admitted, controller sets `boost = +100000` on B.
8. **T=61m**: B's window expires → `boost = -100000`. C preempts B.
9. **Cycle continues**: A → B → C → A → ...

Ladicle: Thanks for detailing this out! I just wanted to double-check if I followed correctly: is the reason for combining positive and negative boosts to ensure the implementation works for all three `withinClusterQueue` policies?

vladikkuzn: Not quite — the **positive + negative** split is not aimed at supporting all three policies with one mechanism.

- **`LowerPriority`** is the one this design **relies on** for handoff after `minAdmitDuration`: the negative boost makes the admitted workload strictly lower than same–base-priority pending work so preemption can happen; that’s what the plan’s trace assumes.

- **`LowerOrNewerEqualPriority`**: ordering is largely driven by **creation time** for same priority, so the negative-boost trick doesn’t fix “older job never yields to newer peers” in the way we need for round-robin.

- **`Any`**: preemption **doesn’t use effective priority** the same way, so a **positive** boost during the window **doesn’t** give “protection” there — the plan calls that out explicitly.

So: **positive** (while admitted & inside the window) + **negative** (after the window, still admitted) + **clear when not admitted** is about getting a **stable** story under **`LowerPriority`**: protect during the window, become strictly preemptable after, then reset when re-queued — not about unifying behavior across `Any`, `LowerOrNewerEqualPriority`, and `LowerPriority`.

Ladicle: I might be missing something, but could we achieve the same goal with just the negative boost? 
Since same-priority workloads are already protected under the `LowerPriority` policy, it seems like applying only the negative boost after the window would be enough. In that case, the positive boost might not be necessary.

vladikkuzn: Kueue only lets a workload preempt another if the preemptor’s effective priority is strictly higher than the victim’s. If they’re equal, preemption doesn’t happen. So while our workload is still at base priority (no annotation) and waiting peers are also at base, they can’t preempt it under that policy — including workloads that don’t use the priority-boost controller at all. They’re just tied at the same priority.

What we really need for the handoff after minAdmitDuration is to make the running workload strictly lower than same-priority waiters. That’s what the negative boost does. The positive boost isn’t doing extra work for that “same priority can’t preempt same priority” rule under LowerPriority.

A few things to keep in mind if we drop the positive phase:

This reasoning is really about LowerPriority. Other preemption modes behave differently, so we shouldn’t assume the same story everywhere.
Effective priority is also used in a few ordering paths (not only this preemption check). A big positive boost changes those tie-breaks; no positive boost avoids that side effect — worth a quick pass to make sure we still like the behavior.
Waiters with higher priority could always preempt anyway; that doesn’t depend on us adding a positive boost.
So I’m aligned: we can likely get the same behavior with no boost during the window, negative after it expires, clear when not admitted — and it might even be simpler (no race to slap on a positive annotation). The three-phase model can stay as one option if we want an obvious “protected” signal in the API, but it doesn’t have to be the only design.

Thanks again for pushing on this.

Ladicle: Thanks for thinking this through. That makes sense to me.

cmd/experimental/priority-boost-controller/charts/priority-boost-controller/README.md
```diff
@@ -0,0 +1,22 @@
...
+  --create-namespace
+```
+
+## Values
+
+Main values are under `priorityBoostController`:
+
+- `image.repository`, `image.tag`, `image.pullPolicy`, `imagePullSecrets`
+- `config.minAdmitDuration` — duration with no annotation after admission; then negative boost while admitted
+- `config.boostValue` — magnitude of the negative `kueue.x-k8s.io/priority-boost` after the window
```
PBundyra, line 19: Please give a concrete example of what's the effect on Workload's priority, if it's negative, maybe we should rename it

cmd/experimental/priority-boost-controller/charts/priority-boost-controller/README.md
```diff
@@ -0,0 +1,22 @@
...
+  --namespace kueue-system \
+  --create-namespace
+```
+
+## Values
+
+Main values are under `priorityBoostController`:
+
+- `image.repository`, `image.tag`, `image.pullPolicy`, `imagePullSecrets`
+- `config.minAdmitDuration` — duration with no annotation after admission; then negative boost while admitted
```
PBundyra, line 18: Does the annotation change every N mins, or it's a one time setting after N mins?

cmd/experimental/priority-boost-controller/pkg/controller/controller.go
```diff
@@ -0,0 +1,218 @@
...
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+	if !wl.DeletionTimestamp.IsZero() {
+		return ctrl.Result{}, nil
+	}
+
+	var newBoost int32
+	var requeueAfter time.Duration
+	if !r.workloadInScope(&wl) {
+		newBoost, requeueAfter = 0, 0
```
PBundyra, line 126: Does the controller updates Workloads that are not in scope? If not, maybe it would be cleaner to return here

cmd/experimental/priority-boost-controller/pkg/controller/controller.go
```diff
@@ -0,0 +1,218 @@
...
+	return ctrl.Result{RequeueAfter: requeueAfter}, nil
+}
+
+func (r *PriorityBoostReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return builder.TypedControllerManagedBy[reconcile.Request](mgr).
+		For(&kueue.Workload{}).
+		Complete(r)
+}
+
+func (r *PriorityBoostReconciler) workloadInScope(wl *kueue.Workload) bool {
```
PBundyra, line 179: Can we move filtering Workloads to EventHandlers?

cmd/experimental/priority-boost-controller/charts/priority-boost-controller/README.md
vladikkuzn: Documented that Kueue uses effective priority = spec.priority + annotation (pkg/util/priority.ParseEffectivePriority), with a numeric example (1000, boostValue 100000 → -100000 on the workload → -99000 effective), and that this makes the workload easier to preempt. 
No rename: the annotation stays the upstream name

cmd/experimental/priority-boost-controller/charts/priority-boost-controller/README.md
vladikkuzn: During the window the controller requeues with RequeueAfter and does not tick the annotation every N minutes; after the window, kueue.x-k8s.io/priority-boost is set once to -boostValue and stays stable until admission/selector/max-priority or clearing logic applies.

cmd/experimental/priority-boost-controller/pkg/config/config.go
```diff
@@ -0,0 +1,109 @@
...
+			return nil, fmt.Errorf("invalid minAdmitDuration %q: %w", cfg.MinAdmitDuration, err)
+		}
+		minAdmitDuration = d
+	}
+
+	out := &ParsedConfiguration{
+		MinAdmitDuration:    minAdmitDuration,
+		BoostValue:          cfg.BoostValue,
+		MaxWorkloadPriority: cfg.MaxWorkloadPriority,
+	}
```
PBundyra:
```suggestion
	out := &ParsedConfiguration{
		BoostValue:          cfg.BoostValue,
		MaxWorkloadPriority: cfg.MaxWorkloadPriority,
	}

		if cfg.MinAdmitDuration != "" && cfg.MinAdmitDuration != "0" {
		d, err := time.ParseDuration(cfg.MinAdmitDuration)
		if err != nil {
			return nil, fmt.Errorf("invalid minAdmitDuration %q: %w", cfg.MinAdmitDuration, err)
		}
		out.MinAdmitDuration = d
	}
```

cmd/experimental/priority-boost-controller/README.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+## Configuration
+
+The controller reads `--config` YAML:
+
+| Field | Type | Default (in code) | Description |
+|-------|------|-------------------|-------------|
+| `minAdmitDuration` | duration | `"0"` | How long after admission to leave **no** annotation. After that, while still admitted, set **negative** boost. `"0"` disables the controller behavior. |
+| `boostValue` | int32 | `100000` | Magnitude of the **negative** boost after the window (stored as `-boostValue` in the annotation). |
```
PBundyra: It would be more intuitive if the name of this field already reflected the boost is negative. Please consider changing to `negativeBoostValue`

cmd/experimental/priority-boost-controller/pkg/controller/controller.go
```diff
@@ -0,0 +1,255 @@
...
+func (r *PriorityBoostReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := ctrl.LoggerFrom(ctx)
+
+	var wl kueue.Workload
+	if err := r.client.Get(ctx, req.NamespacedName, &wl); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+	if !wl.DeletionTimestamp.IsZero() {
+		return ctrl.Result{}, nil
+	}
```
PBundyra: Duplicate from event filters

cmd/experimental/priority-boost-controller/pkg/controller/controller.go
```diff
@@ -0,0 +1,255 @@
...
+	if currentBoostValue == desiredBoostStr {
+		return ctrl.Result{RequeueAfter: requeueAfter}, nil
+	}
+
+	patch := client.MergeFrom(wl.DeepCopy())
+	if newBoost != 0 {
+		if wl.Annotations == nil {
+			wl.Annotations = make(map[string]string)
+		}
+		wl.Annotations[constants.PriorityBoostAnnotationKey] = desiredBoostStr
```
PBundyra: Please use a helper func for that

cmd/experimental/priority-boost-controller/pkg/controller/controller.go
```diff
@@ -0,0 +1,255 @@
...
+	currentBoostValue := wl.Annotations[constants.PriorityBoostAnnotationKey]
+
+	var desiredBoostStr string
+	if newBoost != 0 {
+		desiredBoostStr = strconv.FormatInt(int64(newBoost), 10)
+	}
+
+	if currentBoostValue == desiredBoostStr {
+		return ctrl.Result{RequeueAfter: requeueAfter}, nil
+	}
```
PBundyra: Please move to a helper function `hasDesiredBoostValue`

cmd/experimental/priority-boost-controller/pkg/controller/controller.go
```diff
@@ -0,0 +1,255 @@
...
+
+	patch := client.MergeFrom(wl.DeepCopy())
+	if newBoost != 0 {
+		if wl.Annotations == nil {
+			wl.Annotations = make(map[string]string)
+		}
+		wl.Annotations[constants.PriorityBoostAnnotationKey] = desiredBoostStr
+		log.Info("Setting priority-boost annotation (post-window preemption signal)",
```
PBundyra: What does `post-window preemption signal` mean?

cmd/experimental/priority-boost-controller/pkg/controller/controller.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	var newBoost int32
+	var requeueAfter time.Duration
+	if !r.workloadInScope(&wl) {
+		// Out of scope: only patch when a stale annotation must be removed.
+		if wl.Annotations[constants.PriorityBoostAnnotationKey] == "" {
+			return ctrl.Result{}, nil
+		}
+		newBoost, requeueAfter = 0, 0
```
PBundyra: For the ease of reading, please consider clearing the annotation here with a helper function and existing returning early

cmd/experimental/priority-boost-controller/pkg/controller/controller.go
```diff
@@ -0,0 +1,255 @@
...
+			"Set %s=%s (post-window preemption signal)",
+			constants.PriorityBoostAnnotationKey, desiredBoostStr)
+	} else {
+		delete(wl.Annotations, constants.PriorityBoostAnnotationKey)
+		log.Info("Removing priority-boost annotation",
+			"workload", klog.KObj(&wl),
+			"annotationKey", constants.PriorityBoostAnnotationKey)
+		r.recorder.Eventf(&wl, corev1.EventTypeNormal, "PriorityBoostCleared",
+			"Cleared %s (workload not in scope, not admitted, or feature disabled)",
+			constants.PriorityBoostAnnotationKey)
```
PBundyra: Please wrap in the helper function

cmd/experimental/kueue-priority-booster/pkg/controller/controller.go
```diff
@@ -0,0 +1,255 @@
...
+			},
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				return r.shouldEnqueueWorkload(e.ObjectOld) || r.shouldEnqueueWorkload(e.ObjectNew)
+			},
+			DeleteFunc: func(e event.DeleteEvent) bool {
+				return false
+			},
+			GenericFunc: func(e event.GenericEvent) bool {
+				return r.shouldEnqueueWorkload(e.Object)
+			},
```
PBundyra, line 221: Do we have any use case of a Generic Event upon which the controller should reconcile a Workload? If not please, return false similarly to DeleteEvent

cmd/experimental/priority-boost-controller/README.md
```diff
@@ -0,0 +1,100 @@
+# Priority Boost Controller
+
+The `priority-boost-controller` is an experimental controller that implements
+time-sharing fairness for Kueue workloads by **not** setting
+`kueue.x-k8s.io/priority-boost` during an initial admit window, then applying a
+**negative** boost after that window while the workload remains admitted. That
+lets same–base-priority waiters preempt under **`withinClusterQueue: LowerPriority`**.
```
PBundyra:
```suggestion
time-sharing fairness mechanism for Kueue Workloads by decreasing their effective priority after configurable amount of time using the `kueue.x-k8s.io/priority-boost` annotation. The decrease can lead to preemption issued by pending Workloads with the same priority. Please find more information about the priority boost feature in the KEP [insert the link].```

File: cmd/experimental/priority-boost-controller/README.md [hunk 18]
```diff
@@ -0,0 +1,100 @@
+# Priority Boost Controller
```
PBundyra: Priority Boost Controller is a very generic name that doesn't reflect the behavior of this particular controller well. Can we consider renaming it? My proposal is `Time-Slice Admission Controller`

PBundyra: Or even better, `Time-Sharing Priority Boost Controller`

cmd/experimental/priority-boost-controller/README.md
PBundyra: Or maybe something along the lines: `priorityReductionValue`

cmd/experimental/priority-boost-controller/README.md
```diff
@@ -0,0 +1,100 @@
...
+kubectl apply -k config
+```
+
+## Configuration
+
+The controller reads `--config` YAML:
+
+| Field | Type | Default (in code) | Description |
+|-------|------|-------------------|-------------|
+| `minAdmitDuration` | duration | `"0"` | How long after admission to leave **no** annotation. After that, while still admitted, set **negative** boost. `"0"` disables the controller behavior. |
```
PBundyra: Maybe we could rename it to `timeSharingInterval` or `timeSharingWindow` - something that better reflects the functionality of the controller and the role of the field in it

cmd/experimental/priority-boost-controller/pkg/controller/controller.go
vladikkuzn: From the readme of controller:
```
Workload admitted
      │
      ▼
no annotation               ← same effective priority as peers; no same-priority
  (timeSharingWindow)            preemption under LowerPriority
      │
  [window elapses, still admitted]
      │
      ▼
annotation = -negativeBoostValue    ← post-window preemption signal (magnitude = negativeBoostValue)
      │
  [workload no longer admitted]
      │
      ▼
annotation cleared
```

cmd/experimental/time-sharing-priority-boost-controller/test/integration/controller_test.go
PBundyra, line 1: Can we tests that use ginkgo framework similarly to core Kueue, and have similar structure?

PBundyra: Can we also use wrappers?

cmd/experimental/kueue-priority-booster/pkg/controller/controller_test.go
PBundyra, line 1: Can we merge those tests into a single suite, following conventions in core Kueue?

cmd/experimental/kueue-priority-booster/pkg/controller/controller.go
```diff
@@ -0,0 +1,271 @@
...
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	if !r.workloadInScope(&wl) {
+		return ctrl.Result{}, r.clearBoostAnnotationIfPresent(ctx, &wl, log)
+	}
+
+	newBoost, requeueAfter := r.computeBoost(&wl)
+	if r.hasDesiredNegativeBoostValue(&wl, newBoost) {
+		return ctrl.Result{RequeueAfter: requeueAfter}, nil
```
PBundyra, line 134: Why do we schedule another Reconcile if the value is already as desired?

cmd/experimental/kueue-priority-booster/pkg/controller/controller.go
```diff
@@ -0,0 +1,271 @@
...
+	if !ok || wl == nil {
+		return false
+	}
+	if !wl.DeletionTimestamp.IsZero() {
+		return false
+	}
+	if r.workloadInScope(wl) {
+		return true
+	}
+	return wl.Annotations[constants.PriorityBoostAnnotationKey] != ""
```
PBundyra, line 240: What's the rationale behind this check?

cmd/experimental/kueue-priority-booster/pkg/controller/controller.go
vladikkuzn: While admitted and still inside timeSharingWindow, computeBoost returns (0, remaining) so the annotation should stay unset until the window ends. There is no watch on a clock, so a delayed requeue is the way to run again when remaining hits zero and apply the negative boost.

cmd/experimental/kueue-priority-booster/pkg/controller/controller.go
vladikkuzn: Out-of-scope workloads are normally ignored so the informer does not churn. If this controller previously set kueue.x-k8s.io/priority-boost, we must enqueue when labels/priority/config move the workload out of scope so Reconcile can run clearBoostAnnotationIfPresent and remove a stale annotation. If we never enqueued those objects, the annotation could remain until some unrelated change happened. So the check is: “still needs cleanup.”

cmd/experimental/kueue-priority-booster/pkg/controller/controller.go
PBundyra: I see, in that case can we just check the if the requeueAfter is non zero right after calling a computeBoost and return ctrl.Result{RequeueAfter: requeueAfter} without calling `hasDesiredNegativeBoostValue`. It's more explicit and easier to read that way

cmd/experimental/kueue-priority-booster/pkg/controller/controller.go
PBundyra: Got it, thanks

cmd/experimental/kueue-priority-booster/pkg/controller/controller.go
vladikkuzn: `hasDesiredNegativeBoostValue` is not only an optimization to skip patches when the annotation already matches; it’s also what lets us still Patch in the `requeueAfter > 0` case when the annotation is wrong (e.g. still set from a previous cycle while we’re back inside the time-sharing window with `newBoost == 0`). If we return `ctrl.Result{RequeueAfter: requeueAfter}` solely based on `requeueAfter != 0` before checking desired state, we can skip a needed clearBoostAnnotation + Patch.

I’m happy to restructure for readability (comments and/or splitting the condition) while keeping that equivalence

cmd/experimental/README.md
```diff
@@ -12,6 +12,7 @@ where they will be maintained long term.
 ### Existing integrations
 
 - `kueue-populator`
+- `time-sharing-priority-boost-controller`
```
mimowo: I would suggest to name the controller just: `kueue-priority-boost-controller`, even though "time sharing" is the first use case, because:
1. This is likely not the last use case sfor example, avoiding multiple preemptions was another discussed. In that case I wouldn't like to start another entirely new sub-project in Kueue. We can just add configuration to this one.
2. The current design of the preemption boost annotation anyway does not allow for multiple controllers working, so I think the design is monolitic anyway. Creating multiple controllers we could make impression that users should follow the same pattern, but users should also have only one controller, which might be mixing contributions.

The alternative name I think about is just `kueue-priority-booster` to align the naming even better with `kueue-populator`. The word "controller" is kind of technical detail. The project name `kueue-priority-booster` makes more sense to me.

mimowo: cc @PBundyra

cmd/experimental/time-sharing-priority-boost-controller/test/integration/priority_boost_controller_test.go
```diff
@@ -0,0 +1,74 @@
...
+			NamespacedName: types.NamespacedName{Name: wl.Name, Namespace: wl.Namespace},
+		}
+		result, err := reconciler.Reconcile(ctx, req)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+		gomega.Expect(result.RequeueAfter).To(gomega.BeNumerically(">", 0))
+
+		var updated kueue.Workload
+		gomega.Expect(cl.Get(ctx, types.NamespacedName{Name: wl.Name, Namespace: wl.Namespace}, &updated)).To(gomega.Succeed())
+		_, hasAnn := updated.Annotations[constants.PriorityBoostAnnotationKey]
+		gomega.Expect(hasAnn).To(gomega.BeFalse(), "no annotation should be set during the window")
```
mimowo: Should we rather wrap the asserts with Eventually as we do in Kueue to avoid potential flakes when an operation takes time?

cmd/experimental/README.md
PBundyra: After syncing offline I agree we should prompt users this is a plugin pattern, as there should be only one controller that handles the priority boost annotation. Hence I agree with the suggested name `kueue-priority-booster`

cmd/experimental/kueue-priority-booster/config/manager/controller_config_map.yaml
```diff
@@ -0,0 +1,14 @@
...
+  namespace: kueue-system
+data:
+  controller_manager_config.yaml: |
+    # timeSharingWindow: no annotation while admitted for this duration; after that,
+    # a negative priority-boost is set while still admitted (LowerPriority preemption).
+    # Set to "0" to disable. Examples: "30m", "1h".
+    timeSharingWindow: "30m"
+    # negativeBoostValue: magnitude of the negative boost after the window (annotation value is -negativeBoostValue).
+    negativeBoostValue: 100000
+    # Optional workloadSelector (metav1.LabelSelector) and maxWorkloadPriority — see README.
```
mimowo: Let's embed the configuation under `timeSharing`, something like this:

```yaml
timeSharing:
  duration: "30m"
  negativeBoost: 100000
```
This way we introduce a natural place to potentially extend the confiugration for other modes of operation than just time-sharing.

cmd/experimental/kueue-priority-booster/pkg/config/config.go
```diff
@@ -0,0 +1,107 @@
...
+// Configuration defines the configuration for the kueue-priority-booster.
+type Configuration struct {
+	// TimeSharingWindow is the minimum time a workload must be admitted before
+	// a negative priority-boost is applied to enable preemption by same-priority
+	// peers. Set to "0" to disable. Examples: "30m", "1h".
+	TimeSharingWindow string `json:"timeSharingWindow,omitempty"`
+
+	// NegativeBoostValue is the magnitude of the negative priority-boost applied
+	// after TimeSharingWindow while the workload remains admitted (see README).
+	NegativeBoostValue int32 `json:"negativeBoostValue,omitempty"`
```
mimowo: I would wrap those under `timeSharing`, as suggested here: https://github.com/kubernetes-sigs/kueue/pull/9959/changes#r3259807735

WorkloadSelector seems generic enough to be kept on the top level.

cmd/experimental/kueue-priority-booster/pkg/config/config.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// WorkloadSelector, if set, limits which workloads the controller manages.
+	// Workloads whose labels do not match are left with no annotation (any
+	// existing annotation is cleared).
+	WorkloadSelector *metav1.LabelSelector `json:"workloadSelector,omitempty"`
+
+	// MaxWorkloadPriority, if set, excludes workloads with Spec.Priority greater
+	// than this value (nil Spec.Priority is treated as 0). Excluded workloads
+	// have any priority-boost annotation cleared.
+	MaxWorkloadPriority *int32 `json:"maxWorkloadPriority,omitempty"`
```
mimowo, line 60: Do we have some use case for this parameter mentioned?

cmd/experimental/kueue-priority-booster/test/integration/priority_boost_envtest_test.go
```diff
@@ -0,0 +1,108 @@
...
+		gomega.Expect(k8sClient.Get(ctx, admittedWorkloadKey(wl.Name), wl)).To(gomega.Succeed())
+		setAdmittedCondition(wl, metav1.NewTime(time.Now().Add(-5*time.Second)))
+		gomega.Expect(k8sClient.Status().Update(ctx, wl)).To(gomega.Succeed())
+
+		gomega.Consistently(func(g gomega.Gomega) {
+			var got kueue.Workload
+			g.Expect(k8sClient.Get(ctx, admittedWorkloadKey("wl-in-window"), &got)).To(gomega.Succeed())
+			_, has := got.Annotations[constants.PriorityBoostAnnotationKey]
+			g.Expect(has).To(gomega.BeFalse())
+		}, 2*time.Second, 200*time.Millisecond).Should(gomega.Succeed())
```
mimowo: Avoid magic constants iike this, reuse constants from the core kueue

Makefile
```diff
@@ -468,6 +480,29 @@ kueue-populator-image: PLATFORMS=$(HOST_IMAGE_PLATFORM)
...
+
+.PHONY: kueue-priority-booster-image-push
+kueue-priority-booster-image-push: PUSH=--push
+kueue-priority-booster-image-push: kueue-priority-booster-image-build
+
+# Build a docker local us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue-priority-booster image
+.PHONY: kueue-priority-booster-image
+kueue-priority-booster-image: PLATFORMS=$(HOST_IMAGE_PLATFORM)
+kueue-priority-booster-image: PUSH=--load
+kueue-priority-booster-image: kueue-priority-booster-image-build
```
mimowo, line 504: Let's add CI jobs for these in a follow up.

mimowo:
LGTM, left minor comments

Makefile
vladikkuzn: SGTM

cmd/experimental/kueue-priority-booster/pkg/config/config.go
vladikkuzn: Yes — maxWorkloadPriority is the safety valve for clusters that mix tenant workloads with critical/system pods that also flow through Kueue. Without it, a workloadSelector is the only way to opt critical workloads out, which forces every operator to label their high-priority workloads correctly. The numeric cutoff lets cluster admins keep e.g. anything above spec.priority > 1000 (system-tier) out of the time-sharing rotation by default, regardless of labels

cmd/experimental/kueue-priority-booster/test/integration/priority_boost_envtest_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	ginkgo.It("sets negative boost annotation after time-sharing window", func() {
+		wl := utiltestingapi.MakeWorkload("wl-post-window", ns.Name).
+			Request(corev1.ResourceCPU, "1").
+			Obj()
+		gomega.Expect(k8sClient.Create(ctx, wl)).To(gomega.Succeed())
+
+		gomega.Expect(k8sClient.Get(ctx, admittedWorkloadKey(wl.Name), wl)).To(gomega.Succeed())
+		setAdmittedCondition(wl, metav1.NewTime(time.Now().Add(-integrationTimeSharingWindow-10*time.Second)))
+		gomega.Expect(k8sClient.Status().Update(ctx, wl)).To(gomega.Succeed())
```
mimowo: Wrap that and use Eventually to retry on failures

cmd/experimental/kueue-priority-booster/test/integration/priority_boost_envtest_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	ginkgo.It("does not set boost annotation during time-sharing window", func() {
+		wl := utiltestingapi.MakeWorkload("wl-in-window", ns.Name).
+			Request(corev1.ResourceCPU, "1").
+			Obj()
+		gomega.Expect(k8sClient.Create(ctx, wl)).To(gomega.Succeed())
+
+		gomega.Expect(k8sClient.Get(ctx, admittedWorkloadKey(wl.Name), wl)).To(gomega.Succeed())
+		setAdmittedCondition(wl, metav1.NewTime(time.Now().Add(-5*time.Second)))
+		gomega.Expect(k8sClient.Status().Update(ctx, wl)).To(gomega.Succeed())
```
mimowo: Wrap that and use Eventually to retry on failures

cmd/experimental/kueue-priority-booster/test/integration/priority_boost_envtest_test.go
```diff
@@ -0,0 +1,108 @@
...
+	utiltestingapi "sigs.k8s.io/kueue/pkg/util/testing/v1beta2"
+	"sigs.k8s.io/kueue/test/util"
+)
+
+var _ = ginkgo.Describe("PriorityBoostReconciler envtest", ginkgo.Serial, func() {
+	var ns *corev1.Namespace
+
+	ginkgo.BeforeEach(func() {
+		fwk.StopManager(ctx)
+		fwk.StartManager(ctx, cfg, managerSetup)
```
mimowo: Instead, call Startmanager in BeforeAll, and StopManager in AfterAll

cmd/experimental/kueue-priority-booster/pkg/controller/controller_test.go
```diff
@@ -0,0 +1,449 @@
...
+	neg := defaultNegativeBoost
+	baseR := &PriorityBoostReconciler{
+		timeSharingWindow:  window,
+		negativeBoostValue: neg,
+	}
+
+	cases := map[string]struct {
+		r   *PriorityBoostReconciler
+		wl  *kueue.Workload
+		val func(t *testing.T, boost int32, requeueAfter time.Duration)
```
mimowo: Instead of procedural test, introduce static expectations, `wantBoost`, `wantRequeueAfter`, etc

cmd/experimental/kueue-priority-booster/pkg/config/config.go
```diff
@@ -0,0 +1,118 @@
...
+	TimeSharing TimeSharing `json:"timeSharing,omitempty"`
+
+	// WorkloadSelector, if set, limits which workloads the controller manages.
+	// Workloads whose labels do not match are left with no annotation (any
+	// existing annotation is cleared).
+	WorkloadSelector *metav1.LabelSelector `json:"workloadSelector,omitempty"`
+
+	// MaxWorkloadPriority, if set, excludes workloads with Spec.Priority greater
+	// than this value (nil Spec.Priority is treated as 0). Excluded workloads
+	// have any priority-boost annotation cleared.
```
mimowo: Drop "Excluded workloads have any priority-boost annotation cleared."  - and adjust code.

Do not clear priority from manually boosted workload, even if they are above the "MaxWorkloadPriority"

cmd/experimental/kueue-priority-booster/pkg/config/config.go
```diff
@@ -0,0 +1,118 @@
...
+// operation as sibling top-level keys.
+type TimeSharing struct {
+	// Duration is the minimum time a workload must be admitted before a
+	// negative priority-boost is applied to enable preemption by same-priority
+	// peers. Set to "0" to disable. Examples: "30m", "1h".
+	Duration string `json:"duration,omitempty"`
+
+	// NegativeBoost is the magnitude of the negative priority-boost applied
+	// after Duration while the workload remains admitted (see README).
+	NegativeBoost int32 `json:"negativeBoost,omitempty"`
```
mimowo, line 42: What is the default value here? I think we could have something like 1, that would be good enough for the use case

cmd/experimental/kueue-priority-booster/pkg/config/config.go
```diff
@@ -0,0 +1,118 @@
...
+
+// TimeSharing groups the parameters that control the time-sharing mode
+// (no annotation while admitted for Duration, then a negative boost of
+// magnitude NegativeBoost). Grouping leaves room for future modes of
+// operation as sibling top-level keys.
+type TimeSharing struct {
+	// Duration is the minimum time a workload must be admitted before a
+	// negative priority-boost is applied to enable preemption by same-priority
+	// peers. Set to "0" to disable. Examples: "30m", "1h".
+	Duration string `json:"duration,omitempty"`
```
mimowo: Let's maybe call it "Interval" instead to better reflect on the repeating nature.

Then also call `TimeSharingInterval` the variable used internally.

cmd/experimental/kueue-priority-booster/pkg/config/config.go
```diff
@@ -0,0 +1,118 @@
...
+
+// Configuration defines the configuration for the kueue-priority-booster.
+type Configuration struct {
+	// TimeSharing holds the parameters of the time-sharing mode.
+	TimeSharing TimeSharing `json:"timeSharing,omitempty"`
+
+	// WorkloadSelector, if set, limits which workloads the controller manages.
+	// Workloads whose labels do not match are left with no annotation (any
+	// existing annotation is cleared).
+	WorkloadSelector *metav1.LabelSelector `json:"workloadSelector,omitempty"`
```
mimowo, line 54: Does the default selector match all workloads? Let's clarify that. I think it should be matching all workloads by default.

cmd/experimental/kueue-priority-booster/pkg/config/config.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// Configuration defines the configuration for the kueue-priority-booster.
+type Configuration struct {
+	// TimeSharing holds the parameters of the time-sharing mode.
+	TimeSharing TimeSharing `json:"timeSharing,omitempty"`
+
+	// WorkloadSelector, if set, limits which workloads the controller manages.
+	// Workloads whose labels do not match are left with no annotation (any
+	// existing annotation is cleared).
```
mimowo: Do not clean the priority-boost from workloads not matching the selector, maybe they were adjusted manually by the admin, we are ok with that. Please adjust the comment and the implementation.

cmd/experimental/kueue-priority-booster/pkg/config/config.go
vladikkuzn: Updated the comment to make it explicit: "An empty or nil selector matches all workloads."

cmd/experimental/kueue-priority-booster/pkg/config/config.go
vladikkuzn: Defaults to 100000
It should be some big number, so that the workload remains admitted during the inteval

mimowo: ok, but let's clarify what is the default value in the comment for the field

cmd/experimental/kueue-priority-booster/charts/kueue-priority-booster/templates/manager.yaml
```diff
@@ -0,0 +1,131 @@
...
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Release.Name }}-kueue-priority-booster-config
+  namespace: {{ .Release.Namespace }}
+data:
+  controller_manager_config.yaml: |
+    timeSharing:
+      duration: {{ .Values.kueuePriorityBooster.config.timeSharing.duration | quote }}
```
mimowo: Now this is called interval, align all places

mimowo (commented):
LGTM, only nits at this point

cmd/experimental/kueue-priority-booster/charts/kueue-priority-booster/templates/manager.yaml
vladikkuzn: ty

mimowo:
/lgtm
/approve
Thank you for delivering the experimental project for 0.18. Let's see how people are going to use it. We will welcome future extensions and variations. Please follow up with the CI jobs for it.


cc @Ladicle @dddwsd @sohankunkerkar

tenzen-y:
Due to this PR, the main branch CI keeps failing:

https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-kueue-push-images/2056750186897608704

```
Error: open cmd/experimental/priority-boost-controller/charts/priority-boost-controller/values.yaml: no such file or directory
```

tenzen-y:
@vladikkuzn Could you check and fix that?

mimowo:
Opened issue: https://github.com/kubernetes-sigs/kueue/issues/11330

mimowo:
/release-note-edit
```release-note
Examples: Added a priority booster controller example that adjusts Workload priority
using the `kueue.x-k8s.io/priority-boost` annotation, supporting time sharing in a
ClusterQueue between Workloads with the same priority.
```

Issue #77, opened 2022-02-26
Support dynamically sized (elastic) jobs


ahg-g:
We should have a clear path towards support spark and other dynamically sized jobs. Another example of this is Ray. 

One related aspect is to support dynamically updating the resource requirements of a workload, we can probably limit that to support changing the count of a PodSet in QueuedWorkload (in Spark, the number of workers could change during the runtime of the job, but not the resource requirements of a worker). 

One idea is to model it in a way similar to "in-place update to pod resources" [1], but in our case it would be the count that is mutable. The driver pod in spark would be watching for the corresponding QueuedWorkload instance and adjusts the number of workers when the new count is admitted.

[1] [https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources](https://www.google.com/url?q=https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources&sa=D&source=docs&ust=1645888625262668&usg=AOvVaw3AgbHiaK920CYg_pqAxRMn)

alculquicondor:
This will be easier to support with https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/3521-pod-scheduling-readiness

tenzen-y:
> I am interested in working on this -- this probably needs some sort of design doc, will work with @alculquicondor and see if I can put something together in the next few weeks
> 
> /assign

Hi @andrewsykim! Is there any progress?

andrewsykim:
@tenzen-y I was planning to work on this in a couple weeks during the holiday season, but feel free to start working on this if you're interested.

andrewsykim:
FYI @vicentefb and I are working on a proposal in a google doc, we will share it here soon when it's ready

troychiu:
Hi I would love to work on this issue. Would resuming https://github.com/kubernetes-sigs/kueue/pull/1852 be good starting point?

mimowo:
> Hi I would love to work on this issue, especially the ray autoscaling support. Would resuming https://github.com/kubernetes-sigs/kueue/pull/1852 be a good starting point?

@troychiu sorry for the late reply. You are super welcome to work on that, but be aware the KEP and PR are old and some aspects have changed. Also, this is one of the more challenging KEPs. I think resuming the PR might be a good idea, but it requires a careful check.

mimowo:
[...]

PR #9960, opened 2026-03-17
Generate elastic workload name based on raycluster generation

hiboyang:
#### What type of PR is this?

/kind bug

Fixes autoscaling issue where workload name is not generated properly
Related to this issue (but this PR not fully fix it): https://github.com/kubernetes-sigs/kueue/issues/77

Fixes: https://github.com/kubernetes-sigs/kueue/issues/9523

```release-note
RayJob integration: fix the autosaling scenarios when using ElasticJobsViaWorkloadSlices. In particular when
two consecutive scale ups happen.
```

### DIFF
--- a/test/e2e/singlecluster/kuberay_test.go
+++ b/test/e2e/singlecluster/kuberay_test.go
@@ -525,11 +525,6 @@ print([ray.get(my_task.remote(i, 1)) for i in range(32)])`,
 			gomega.Expect(k8sClient.Create(ctx, rayJob)).Should(gomega.Succeed())
 		})
 
-		// Variable to store initial pod names for verification during scaling
-		var initialPodNames []string
-		// Variable to store scaled-up pod names for verification during scaling down
-		var scaledUpPodNames []string
-
 		ginkgo.By("Checking one workload is created", func() {
 			gomega.Eventually(func(g gomega.Gomega) {
 				workloadList := &kueue.WorkloadList{}
@@ -564,9 +559,6 @@ print([ray.get(my_task.remote(i, 1)) for i in range(32)])`,
 				// Get worker pod names and check count
 				workerPodNames := getRunningWorkerPodNames(podList)
 				g.Expect(workerPodNames).To(gomega.HaveLen(1), "Expected exactly 1 pod with 'workers' in the name")
-
-				// Store initial pod names for later verification
-				initialPodNames = workerPodNames
 			}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
 		})
 
@@ -587,13 +579,6 @@ print([ray.get(my_task.remote(i, 1)) for i in range(32)])`,
 				// Get worker pod names and check count
 				currentPodNames := getRunningWorkerPodNames(podList)
 				g.Expect(currentPodNames).To(gomega.HaveLen(5), "Expected exactly 5 pods with 'workers' in the name after second scale-up")
-
-				// Verify that the current pod names are a superset of the initial pod names
-				g.Expect(currentPodNames).To(gomega.ContainElements(initialPodNames),
-					"Current worker pod names should be a superset of initial pod names")
-
-				// Store scaled-up pod names for later verification during scaling down
-				scaledUpPodNames = currentPodNames
 			}, util.VeryLongTimeout, util.Interval).Should(gomega.Succeed())
 		})
 
@@ -626,10 +611,6 @@ print([ray.get(my_task.remote(i, 1)) for i in range(32)])`,
 				// Get worker pod names and check count
 				currentPodNames := getRunningWorkerPodNames(podList)
 				g.Expect(currentPodNames).To(gomega.HaveLen(1), "Expected exactly 1 pods with 'workers' in the name")
-
-				// Verify that the previous scaled-up pod names are a superset of the current pod names
-				g.Expect(scaledUpPodNames).To(gomega.ContainElements(currentPodNames),
-					"Previous scaled-up worker pod names should be a superset of current pod names")
 			}, util.VeryLongTimeout, util.Interval).Should(gomega.Succeed())
 		})
 


pkg/controller/jobframework/workload_names.go
```diff
@@ -34,22 +33,37 @@ const (
...
+	return generateWorkloadNameWithExtra(ownerName, ownerUID, ownerGVK, "")
 }
 
-func GetWorkloadNameForOwnerWithGVKAndGeneration(ownerName string, ownerUID types.UID, ownerGVK schema.GroupVersionKind, generation int64) string {
-	return generateWorkloadName(ownerName, ownerUID, ownerGVK, &generation)
+// ElasticWorkloadNameProvider interface contains methods to provide extra information to build workload name for elastic job
+type ElasticWorkloadNameProvider interface {
+	GetGeneration() int64
+	GetAnnotations() map[string]string
 }
```
mimowo: Maybe instead of that we just use as extra we use `generation + "_" + podSetSizes` annoation. I'm thinking it may happen that if we do scale up -> scale down, then we still can have a conflict of identifiers.

Maybe a better approach would be to change the `podset-sizes-annotation` as `raycluster-info`, making it more specific, and include there: podSetSizes + rayClusterGenearation. Then, for the generation of the workload name we would use:
"RayJobGeneration+_+RayClusterGenearation", wdyt?

Also, could you explain why using `ResourceVersion` does not work well? Is it because we require stability for the hash?

hiboyang: Yeah, we hit workload name conflict issue again when scale up -> scale down -> scale up. Using `ResourceVersion` might work, but will break a lot of tests. I am trying to see find some backward compatible ways to do this. Let me experiment more regarding your suggestions here.

pkg/controller/jobframework/workload_names.go
```diff
@@ -39,13 +39,18 @@ func GetWorkloadNameForOwnerWithGVK(ownerName string, ownerUID types.UID, ownerG
...
 func GetElasticWorkloadName(ownerName string, ownerUID types.UID, ownerGVK schema.GroupVersionKind, elasticWorkloadNameProvider ElasticWorkloadNameProvider) string {
-	extra := elasticWorkloadNameProvider.GetAnnotations()[PodsetReplicaSizesAnnotation]
-	if extra == "" {
+	extra := ""
+	if elasticWorkloadNameProvider.GetAnnotations()[PodsetReplicaSizesAnnotation] == "" {
+		// Use generation for workload name to keep backward compatible
 		extra = strconv.FormatInt(elasticWorkloadNameProvider.GetGeneration(), 10)
+	} else {
+		// Use resourceVersion to avoid workload name conflict
+		extra = elasticWorkloadNameProvider.GetResourceVersion()
```
sohankunkerkar: I think this is non-deterministic. You might want to change this to Raycluster generation per @mimowo's suggestion.

pkg/controller/jobs/job/job_multikueue_adapter.go
```diff
@@ -162,7 +162,7 @@ func needElasticJobSync(log logr.Logger, workloadName string, localJob, remoteJo
 	}
 	oldParallelism := ptr.Deref(remoteJob.Spec.Parallelism, 0)
 	newParallelism := ptr.Deref(localJob.Spec.Parallelism, 0)
-	newWorkloadName := jobframework.GetWorkloadNameForOwnerWithGVKAndGeneration(localJob.GetName(), localJob.GetUID(), gvk, localJob.GetGeneration())
+	newWorkloadName := jobframework.GetElasticWorkloadName(localJob.GetName(), localJob.GetUID(), gvk, &localJob.ObjectMeta)
```
sohankunkerkar: For regular Jobs, this also uses ResourceVersion when the annotation is present. Same instability issue applies here, not just RayJobs.

pkg/controller/jobframework/workload_names.go
mimowo: I think instead of this board interface maybe it is better to just have `getExtraForHash` function which would already compute the value. Then it could also be used in a backwards compatible fashion, doing the same fallback for all other CRDs .

Maybe we could consider introducing another annotation, like `raycluster-generation` produced by the CustomAnnotations, and would be used in the `getExtraForHash` function.

pkg/controller/jobframework/workload_names.go
hiboyang: Originally I was thinking making it deterministic, then I find actually it does not need to be deterministic, the only place it matters is test code, since in some test code, we verify workload by a given name. The code in the PR is backward compatible with those tests.

pkg/controller/jobs/job/job_multikueue_adapter.go
hiboyang: See previous comment, determinstic/stability is actually not required.

pkg/controller/jobframework/workload_names.go
hiboyang: I am trying to make workload name change:
1. Backward compatible so old test with workload name checking still pass
2. Avoid workload name conflict by using resourceVersion
Thus it needs these three:
```
	GetGeneration() int64
	GetResourceVersion() string
	GetAnnotations() map[string]string
```

Using annotation `raycluster-generation` could be another solution, but still try to explore other options before adding the new annotation :)

pkg/controller/jobframework/utils.go
```diff
@@ -143,7 +148,8 @@ func GetWorkloadslicingCustomAnnotations(object client.Object, podSets []kueue.P
 				return nil, fmt.Errorf("failed to marshal updated podsets: %w", err)
 			}
 			return map[string]string{
-				PodsetReplicaSizesAnnotation: string(podSetsJSON),
+				RayClusterPodsetReplicaSizesAnnotation: string(podSetsJSON),
+				RayClusterGenerationAnnotation:         strconv.FormatInt(object.GetGeneration(), 10),
```
sohankunkerkar: I'm wondering if we need to pass the Raycluster's generation here? IIUC, this is passing the RayJob's generation.

hiboyang: yes, need to pass the RayCluster's generation, updated the PR to fix the issue

pkg/controller/jobs/raycluster/common.go
```diff
@@ -252,3 +267,91 @@ func BuildPodSetAnnotationsPathByNameMap(rayClusterSpec *rayv1.RayClusterSpec, h
...
+			log.Error(err, "Failed to get ray cluster ", "rayCluster", rayClusterName)
+		} else {
+			rayClusterGeneration = strconv.FormatInt(rayClusterObj.GetGeneration(), 10)
+		}
+
+		// Compare current counts against previous annotation. If any differ, update the in-memory
+		// annotation with ALL current podSet counts (not just the changed ones). The actual API server
+		// patch is handled by the reconciler.
+		changed := ComparePodSetCounts(podSets, previousCounts)
+		if changed {
```
mimowo: I think we don't need to detect the change here. It is actually tricky to do right. We can rely on the change detection outside of this function - in the mergeAnnotations. So here I think we can unconditionally return the new values.

pkg/controller/jobs/raycluster/common.go
```diff
@@ -252,3 +267,91 @@ func BuildPodSetAnnotationsPathByNameMap(rayClusterSpec *rayv1.RayClusterSpec, h
...
+
+		rayClusterGeneration := ""
+
+		var rayClusterObj rayv1.RayCluster
+		err = c.Get(ctx, types.NamespacedName{
+			Namespace: jobObject.GetNamespace(),
+			Name:      rayClusterName,
+		}, &rayClusterObj)
+		if err != nil {
+			log.Error(err, "Failed to get ray cluster ", "rayCluster", rayClusterName)
```
mimowo: If this is NotFound, then just log and ignore. If any other error then return from the function

pkg/controller/jobs/raycluster/common.go
hiboyang: updated

pkg/controller/jobs/raycluster/common.go
hiboyang: updated

test/e2e/singlecluster/kuberay_test.go
```diff
@@ -405,9 +404,265 @@ print([ray.get(my_task.remote(i, 1)) for i in range(32)])`,
...
+			}, util.VeryLongTimeout, util.Interval).Should(gomega.Succeed())
+		})
+
+		ginkgo.By("Waiting for at least 2 total workloads due to first scale-up", func() {
+			gomega.Eventually(func(g gomega.Gomega) {
+				workloadList := &kueue.WorkloadList{}
+				g.Expect(k8sClient.List(ctx, workloadList, client.InNamespace(ns.Name))).To(gomega.Succeed())
+				g.Expect(len(workloadList.Items)).To(gomega.BeNumerically(">=", 2), "Expected at least 2 workloads after first scale-up")
+			}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
+		})
```
mimowo: I think we could skip these asserts. I worry they might be flaky, and the test would fail without them, and without the fix, anyway

pkg/controller/jobframework/workload_names_test.go
```diff
@@ -0,0 +1,150 @@
...
+		if elasticName == staticName {
+			t.Errorf("elastic and static names should differ, both are %q", elasticName)
+		}
+	})
+}
+
+func TestGenerateWorkloadNameWithGeneration(t *testing.T) {
+	gvk := schema.GroupVersionKind{Group: "batch", Version: "v1", Kind: "Job"}
+
+	t.Run("different generations produce different names", func(t *testing.T) {
```
mimowo: ni, turn the test to use table-driven test cases

pkg/controller/jobframework/workload_names_test.go
```diff
@@ -0,0 +1,150 @@
...
+			// Name should have prefix + "-" + 5-char hash
+			prefix := GenerateWorkloadNamePrefix(tc.ownerName, tc.ownerUID, tc.ownerGVK)
+			if len(got) != len(prefix)+1+hashLength {
+				t.Errorf("GenerateWorkloadNameWithExtra() length = %d, want %d (prefix=%d + 1 + hash=%d)", len(got), len(prefix)+1+hashLength, len(prefix), hashLength)
+			}
+		})
+	}
+
+	// Verify that different extra values produce different names.
+	t.Run("different extra values produce different names", func(t *testing.T) {
```
mimowo: Make the asserts table driven

mimowo:
/release-note-edit
```release-note
RayJob integration: fix the autosaling scenarios when using ElasticJobsViaWorkloadSlices. In particular when
two consecutive scale ups happen. 
```

test/e2e/singlecluster/kuberay_test.go
hiboyang: Good suggestion, updated

pkg/controller/jobframework/workload_names_test.go
hiboyang: updated

pkg/controller/jobframework/workload_names_test.go
hiboyang: updated

test/e2e/singlecluster/kuberay_test.go
```diff
@@ -405,9 +404,242 @@ print([ray.get(my_task.remote(i, 1)) for i in range(32)])`,
...
+		ginkgo.By("Waiting for second scale-up to 5 workers due to high parallelism tasks", func() {
+			gomega.Eventually(func(g gomega.Gomega) {
+				podList := &corev1.PodList{}
+				g.Expect(k8sClient.List(ctx, podList, client.InNamespace(ns.Name))).To(gomega.Succeed())
+				// Get worker pod names and check count
+				currentPodNames := getRunningWorkerPodNames(podList)
+				g.Expect(currentPodNames).To(gomega.HaveLen(5), "Expected exactly 5 pods with 'workers' in the name after second scale-up")
+
+				// Verify that the current pod names are a superset of the initial pod names
+				g.Expect(currentPodNames).To(gomega.ContainElements(initialPodNames),
```
sohankunkerkar: I think If a worker gets replaced for any reason (OOM, node pressure), this fails even when the workload-slice fix is working correctly. Asserting on worker count alone is sufficient to verify the fix

test/e2e/singlecluster/kuberay_test.go
```diff
@@ -405,9 +404,242 @@ print([ray.get(my_task.remote(i, 1)) for i in range(32)])`,
...
+		ginkgo.By("Waiting for workers reduced to 1 due to scaling down", func() {
+			gomega.Eventually(func(g gomega.Gomega) {
+				podList := &corev1.PodList{}
+				g.Expect(k8sClient.List(ctx, podList, client.InNamespace(ns.Name))).To(gomega.Succeed())
+				// Get worker pod names and check count
+				currentPodNames := getRunningWorkerPodNames(podList)
+				g.Expect(currentPodNames).To(gomega.HaveLen(1), "Expected exactly 1 pods with 'workers' in the name")
+
+				// Verify that the previous scaled-up pod names are a superset of the current pod names
+				g.Expect(scaledUpPodNames).To(gomega.ContainElements(currentPodNames),
```
sohankunkerkar: https://github.com/kubernetes-sigs/kueue/pull/9960/changes#r2997543952

pkg/controller/jobs/rayservice/rayservice_controller.go
```diff
@@ -218,7 +219,11 @@ func (j *RayService) PodsReady(ctx context.Context) bool {
 }
 
 func (j *RayService) GetCustomAnnotations(ctx context.Context, c client.Client, podSets []kueue.PodSet) (map[string]string, error) {
-	return jobframework.GetWorkloadslicingCustomAnnotations(j.Object(), podSets)
+	return raycluster.GetWorkloadslicingRayClusterCustomAnnotations(ctx, c, j.Object(), podSets, j.Status.ActiveServiceStatus.RayClusterName)
```
sohankunkerkar, line 222: Hmm...The fix was generalized to RayService here which is good, but the new multi-scale-up e2e only adds the RayJob path. Might be worth adding at least a unit test for RayService autoscaling to make sure this path doesn't quietly break.

test/e2e/singlecluster/kuberay_test.go
hiboyang: good suggestion, updated

test/e2e/singlecluster/kuberay_test.go
hiboyang: updated

pkg/controller/jobs/rayservice/rayservice_controller.go
hiboyang: There is unit test for `raycluster.GetWorkloadslicingRayClusterCustomAnnotations`. For RayService autoscaling e2e test, it is pretty complicated, how about I work on a follow-up PR after this one? Will need more time on that.

test/e2e/singlecluster/kuberay_test.go
```diff
@@ -405,9 +404,223 @@ print([ray.get(my_task.remote(i, 1)) for i in range(32)])`,
...
+		ginkgo.By("Waiting for at least 3 total workloads due to multiple scale-ups", func() {
+			// Use >= 3 since multiple scale-up events should create additional workload slices.
+			gomega.Eventually(func(g gomega.Gomega) {
+				workloadList := &kueue.WorkloadList{}
+				g.Expect(k8sClient.List(ctx, workloadList, client.InNamespace(ns.Name))).To(gomega.Succeed())
+				g.Expect(len(workloadList.Items)).To(gomega.BeNumerically(">=", 3), "Expected at least 3 workloads due to multiple scale-ups")
+			}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
+		})
+
+		ginkgo.By("Waiting for workers reduced to 1 due to scaling down", func() {
```
mimowo, line 607: nit, I think this part could be out of scope for the test case which aims to simulate multiple scale ups

mimowo: We could also drop this part:
```
print([ray.get(my_task.remote(i, 1)) for i in range(32)])`,
```

pkg/controller/jobs/rayservice/rayservice_controller.go
mimowo: I'm ok to add the e2e test for autoscaling RayService as a follow up. That could not be cherrypicked also.

test/e2e/singlecluster/kuberay_test.go
mimowo: I think we can follow up on that

PR #9973, opened 2026-03-18
Change workload name generation to ensure max 63 characters.

### LINKED ISSUE #9872: Change workload name generation to ensure max 63 characters (fitting as label)
I would like to change the workload name generation to cap them at 63 characters. 

I think this is not really breaking, because in Kueue we don't rely on the algorithm. We will just cut them shorter. 

Here is the algorithm generation: https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/jobframework/workload_names.go#L30-L34

For safety we should do it behind a feature gate.

**Why is this needed**:

Some users need to use workload names in external systems as labels.

mbobrovskyi:
#### What type of PR is this?
/kind bug

Change workload name generation to ensure max 63 characters.

Fixes #9872

```release-note
Introduced the `WorkloadNameShorten` feature gate to ensure generated Workload names do not exceed 63 characters. This prevents issues where Workload labels were invalid due to length. When enabled, the owner-based prefix is truncated to fit within the limit while maintaining uniqueness via a hash suffix.
```

### DIFF
--- a/pkg/controller/jobframework/workload_names.go
+++ b/pkg/controller/jobframework/workload_names.go
@@ -31,8 +31,13 @@ import (
 )
 
 const (
-	hashLength                 = 5
-	longMaxWorkloadNameLength  = 253
+	hashLength                = 5
+	longMaxWorkloadNameLength = 253
+	// shortMaxWorkloadNameLength is set to 63 to comply with Kubernetes label value length constraints.
+	// Kubernetes label values must be 63 characters or less per the Kubernetes specification.
+	// Since workload names are used as label values, they must respect this limit.
+	// See https://github.com/kubernetes-sigs/kueue/issues/9872 and
+	// https://github.com/kubernetes-sigs/kueue/issues/10098 for more details.
 	shortMaxWorkloadNameLength = 63
 )
 


pkg/controller/jobframework/workload_names.go
```diff
@@ -19,20 +19,44 @@ package jobframework
...
+
+	"sigs.k8s.io/kueue/pkg/features"
 )
 
 const (
-	hashLength = 5
-	// 253 is the maximal length for a CRD name. We need to subtract one for '-', and the hash length.
-	maxPrefixLength = 252 - hashLength
+	hashLength                   = 5
+	defaultMaxWorkloadNameLength = 253
```
mimowo:
```suggestion
	longMaxWorkloadNameLength = 253
```

pkg/features/kube_features.go
```diff
@@ -328,6 +328,14 @@ const (
...
 	// Enables gating the admission of workloads based on annotations.
 	AdmissionGatedBy featuregate.Feature = "AdmissionGatedBy"
+
+	// owner: @mbobrovskyi
+	//
+	// issue: https://github.com/kubernetes-sigs/kueue/issues/9872
+	//
+	// WorkloadNameShorten ensures that generated Workload names do not exceed
+	// 63 characters, making them compatible with Kubernetes label value limits.
+	WorkloadNameShorten featuregate.Feature = "WorkloadNameShorten"
```
```suggestion
	ShortWorkloadNames featuregate.Feature = "ShortWorkloadNames"
```
nit

pkg/features/kube_features.go
```diff
@@ -509,6 +517,9 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 	AdmissionGatedBy: {
 		{Version: version.MustParse("0.17"), Default: false, PreRelease: featuregate.Alpha},
 	},
+	WorkloadNameShorten: {
+		{Version: version.MustParse("0.17"), Default: false, PreRelease: featuregate.Alpha},
```
mimowo, line 540: I would prefer to make it default, but let me check with @tenzen-y and @gabesaba 

In any case please open a separete experiment PR to check if all tests pass if we enable it by default

pkg/features/kube_features.go
mbobrovskyi: Done

pkg/controller/jobframework/workload_names.go
mbobrovskyi: Done

pkg/controller/jobframework/workload_names.go
```diff
@@ -42,11 +66,11 @@ func GetWorkloadNameForOwnerWithGVKAndGeneration(ownerName string, ownerUID type
...
 
 func generateWorkloadName(ownerName string, ownerUID types.UID, ownerGVK schema.GroupVersionKind, generation *int64) string {
-	prefixedName := strings.ToLower(ownerGVK.Kind) + "-" + ownerName
-	if len(prefixedName) > maxPrefixLength {
-		prefixedName = prefixedName[:maxPrefixLength]
-	}
-	return prefixedName + "-" + getHash(ownerName, ownerUID, ownerGVK, generation)[:hashLength]
+	return fmt.Sprintf(
+		"%s-%s",
+		truncate(fmt.Sprintf("%s-%s", strings.ToLower(ownerGVK.Kind), ownerName), maxPrefixLength()),
```
wkpan1023, line 76: For prefix, should we consider group and version as well to avoid Kind name collision?

mbobrovskyi: We are generating a hash to avoid collisions:

https://github.com/kubernetes-sigs/kueue/blob/c1a0fe53970dfdc45d2ddc3821cb275f8a5c09b4/pkg/controller/jobframework/workload_names.go#L52-L66

wkpan1023: looks good. thanks.

pkg/features/kube_features.go
mimowo: lets keep alpha for now and discuss about the future later

PR #9974, opened 2026-03-18
Skip creation staging directory when execute generate-featuregates target.

mbobrovskyi:
#### What type of PR is this?
/kind cleanup

Skip creation staging directory when execute generate-featuregates target.

Fixes #

```release-note
NONE
```

### DIFF
--- a/hack/tools/compatibility-lifecycle/generate.sh
+++ b/hack/tools/compatibility-lifecycle/generate.sh
@@ -66,7 +66,6 @@ if [[ "${need_build}" == "true" ]]; then
   echo -n "${K8S_REPO_REF}" > "${COMPAT_LIFECYCLE_BIN_STAMP}"
 fi
 
-mkdir -p "${REPO_ROOT}/staging"
 pushd "${REPO_ROOT}" >/dev/null
 "${COMPAT_LIFECYCLE_BIN}" feature-gates update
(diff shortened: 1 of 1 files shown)


hack/tools/compatibility-lifecycle/generate.sh
```diff
@@ -66,7 +66,6 @@ if [[ "${need_build}" == "true" ]]; then
   echo -n "${K8S_REPO_REF}" > "${COMPAT_LIFECYCLE_BIN_STAMP}"
 fi
 
-mkdir -p "${REPO_ROOT}/staging"
```
mimowo, line 69: Why can we safely skip? Is it not needed? How do we test that it works without it?

mbobrovskyi: I ran `make generate-featuregates`, and it works fine without it. I’m not sure why we are creating this directory, to be honest.

vladikkuzn: Please try to remove feature/remove generated data and run `make generate-featuregates`; if that works without any problems, I'm good to merge it

mbobrovskyi: Yeah, it works.

PR #9983, opened 2026-03-18
Fix the bulk-scheduling based on scheduling hashes

### LINKED ISSUE #9694: Skip identical workloads in CQ processing

mimowo:
#### What type of PR is this?

/kind bug

Fixes #9694

Follow up to https://github.com/kubernetes-sigs/kueue/pull/9698 to cover the Preempt && noPreemptionCandidates case

```release-note
Scheduling: fix the bulk-scheduling mechanism to also cover workloads which are inadmissible as they are
in the "Preempt" mode without preemption candidates.
```

### DIFF
--- a/pkg/cache/queue/cluster_queue.go
+++ b/pkg/cache/queue/cluster_queue.go
@@ -413,6 +413,12 @@ func (c *ClusterQueue) requeueIfNotPresent(log logr.Logger, wInfo *workload.Info
 	}
 	log.V(2).Info(logMsg, "clusterQueue", c.name, "workload", key)
 
+	if features.Enabled(features.SchedulingEquivalenceHashing) && wInfo.SchedulingHash != workload.SchedulingHashUnknown && !immediate {
+		if moved := c.handleInadmissibleHash(wInfo.SchedulingHash); moved > 0 {
+			log.V(2).Info("Bulk-moved equivalent workloads to inadmissible", "hash", wInfo.SchedulingHash, "movedCount", moved)
+		}
+	}
+
 	return true
 }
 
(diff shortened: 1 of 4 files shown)


pkg/cache/queue/cluster_queue.go
```diff
@@ -427,8 +433,6 @@ func (c *ClusterQueue) forgetInflightByKey(key workload.Reference) {
 // Only applies to BestEffortFIFO queues; in StrictFIFO the head workload
 // stays in the heap and must not cause equivalent workloads to be skipped.
 func (c *ClusterQueue) handleInadmissibleHash(hash string) int {
```
sohankunkerkar, line 435: I think [HandleInadmissibleHash](https://github.com/kubernetes-sigs/kueue/blob/5a35cebf790dda2c6a0b69a07bc97d7dacd1c82c/pkg/cache/queue/manager.go#L606) becomes the dead code now  without the CQ lock.

sohankunkerkar (commented):
After this LQ metrics for bulk-moved workloads might change.

pkg/cache/queue/cluster_queue.go
mimowo: Yeah, let me see, we should probably call it still

PR #9984, opened 2026-03-18
helm: support custom cert-manager issuerRef

### LINKED ISSUE #9965: Helm chart: Support custom cert-manager Issuer/ClusterIssuer
### What would you like to be added?

When `enableCertManager: true` is set in the Helm chart, the chart always creates a self-signed `Issuer` and hardcodes the `issuerRef` in all three `Certificate` resources (webhook, visibility server, metrics). There is no way to reference a pre-existing `Issuer` or `ClusterIssuer`.

Many organizations manage certificates centrally using a shared `ClusterIssuer` backed by an internal CA, HashiCorp Vault, AWS PCA, or similar. Currently, the only workaround is to fork the chart or use a Helm post-renderer to patch the generated resources.

### Proposed change

Add an optional `certManager.issuerRef` value that, when set, is used in place of the auto-generated self-signed issuer. When not set, the current behavior is preserved.

**values.yaml:**

```yaml
enableCertManager: false
certManager:
  # -- Override the default self-signed issuer. When set, the chart
  #    skips creating its own Issuer and uses this reference instead.
  # issuerRef:
  #   name: "my-cluster-issuer"
  #   kind: "ClusterIssuer"
  #   group: "cert-manager.io"
```

**Template logic (certificate.yaml and others):**

```yaml
{{- if .Values.enableCertManager }}
{{- if not .Values.certManager.issuerRef }}
# Only create the self-signed Issuer when no custom issuerRef is provided
apiVersion: cert-manager.io/v1
kind: Issuer
metadata:
  name: '{{ include "kueue.fullname" . }}-selfsigned-issuer'
  namespace: '{{ .Release.Namespace }}'
spec:
  selfSigned: {}
{{- end }}
---
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  name: '{{ include "kueue.fullname" . }}-serving-cert'
  namespace: '{{ .Release.Namespace }}'
spec:
  dnsNames:
    - '{{ include "kueue.fullname" . }}-webhook-service.{{ .Release.Namespace }}.svc'
    - '{{ include "kueue.fullname" . }}-webhook-service.{{ .Release.Namespace }}.svc.{{ .Values.kubernetesClusterDomain }}'
  issuerRef:
    {{- if .Values.certManager.issuerRef }}
    {{- toYaml .Values.certManager.issuerRef | nindent 4 }}
    {{- else }}
    kind: Issuer
    name: '{{ include "kueue.fullname" . }}-selfsigned-issuer'
    {{- end }}
  secretName: '{{ include "kueue.fullname" . }}-webhook-server-cert'
{{- end }}
```

The same pattern would apply to `certificate-visibility-server.yaml` and `certificate-metrics.yaml`.

### Why is this needed?

- Self-signed certificates are often not acceptable in production environments with strict PKI requirements.
- Users who already have a cert-manager `ClusterIssuer` configured should be able to reuse it without forking the chart.
- This is a non-breaking, additive change — the default behavior remains unchanged.

### Prior art

Other Kubernetes Helm charts already support this pattern:

- **kube-prometheus-stack**: Supports an optional [`issuerRef`](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus-operator/certmanager.yaml) under `prometheusOperator.admissionWebhooks.certManager`. When set, the chart skips self-signed issuer creation and uses the provided reference. ([values.yaml](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml), [related issue](https://github.com/prometheus-community/helm-charts/issues/6017))

- **Rancher**: Supports custom issuers via [`ingress.tls.source`](https://github.com/rancher/server-chart/blob/master/rancher/templates/ingress.yaml) and `ingress.tls.issuerName`, allowing users to specify a `ClusterIssuer` or namespace-scoped `Issuer`. ([related issue](https://github.com/rancher/rancher/issues/16178))

### Related issues

- #6152 — Kueue Controller-Manager and client certs provided by apiserver (related but focuses on trusting external CAs rather than configuring the issuer)

MatteoFari:
#### What type of PR is this?
/kind feature
/area integrations

This PR lets the Helm chart use an existing cert-manager `Issuer` or `ClusterIssuer` with `certManager.issuerRef`.

Before this change, turning on `enableCertManager=true` always made the chart create and use its own self-signed `Issuer` making it hard to use a cluster that already has a shared issuer. Users can now point the webhook, metrics, and visibility certificates to an existing issuer instead.

Fixes #9965

```release-note
Helm: Add support to pass in custom issuerRef to allow for configuration of Issuers.
```


kannon92:
ping @mimowo @tenzen-y

PR #9985, opened 2026-03-18
Commonize gauge metric reporting, create clusterqueue_metrics in …

### LINKED ISSUE #9793: Commonize gauge metric reporting to use cache as single data source

Commonize gauge metric reporting so that the cache becomes the single source of truth. In #9487, gauge metrics are currently reported from two different places using different data sources.

1. One path is the cache resync (`ResyncGaugeMetrics()` in `pkg/cache/scheduler/cache.go` and `pkg/cache/queue/manager.go`), which reports metrics by iterating over the cache state. These two methods follow the same pattern and are always called together. 
2. The other path is the controller reconcilers (`pkg/controller/core/clusterqueue_controller.go` and `pkg/controller/core/localqueue_controller.go`), which report metrics based on status updates during reconciliation.

**Why is this needed**:
As discussed in https://github.com/kubernetes-sigs/kueue/pull/9487#discussion_r2904826910, the intended direction is for both status updates and metrics to rely on the cache as the data source, instead of building metrics from controller statuses. 

Moving metric reporting fully to the cache would remove the duplication between cache resync and controller reconcilers and reduce the risk of inconsistencies when new metrics are introduced.

reruno:
…pkg/cache/scheduler/ package

#### What type of PR is this?
/kind cleanup

Commonize gauge metric reporting so that the cache becomes the single source of truth.

Fixes #9793 

```release-note
NONE
```

### DIFF
--- a/pkg/cache/scheduler/clusterqueue_metrics.go
+++ b/pkg/cache/scheduler/clusterqueue_metrics.go
@@ -1,0 +1,79 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package scheduler
+
+import (
+	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+	"sigs.k8s.io/kueue/pkg/metrics"
+	"sigs.k8s.io/kueue/pkg/resources"
+)
+
+func (c *Cache) RecordClusterQueueResourceMetrics(log logr.Logger, cqName kueue.ClusterQueueReference) {
+	log.V(4).Info("Recording resource metrics for ClusterQueue")
+
+	c.RLock()
+	defer c.RUnlock()
+
+	cq := c.hm.ClusterQueue(cqName)
+	if cq == nil {
+		return
+	}
+
+	cq.reportResourceMetrics(c.fairSharingEnabled)
+}
+
+func (c *Cache) ClearClusterQueueOldResourceMetrics(log logr.Logger, oldCq *kueue.ClusterQueue) {
+	c.RLock()
+	defer c.RUnlock()
+
+	cqName := kueue.ClusterQueueReference(oldCq.Name)
+	log.V(4).Info("Clearing old resource metrics for ClusterQueue")
+
+	newCq := c.hm.ClusterQueue(cqName)
+	if newCq == nil {
+		return
+	}
+
+	for rgi := range oldCq.Spec.ResourceGroups {
+		oldRg := &oldCq.Spec.ResourceGroups[rgi]
+		newFlavorSet := sets.New[kueue.ResourceFlavorReference]()
+		if rgi < len(newCq.ResourceGroups) && len(newCq.ResourceGroups[rgi].Flavors) > 0 {
+			newFlavorSet.Insert(newCq.ResourceGroups[rgi].Flavors...)
+		}
+
+		for fi := range oldRg.Flavors {
+			flavor := &oldRg.Flavors[fi]
+			if !newFlavorSet.Has(flavor.Name) {
+				metrics.ClearClusterQueueResourceQuotas(oldCq.Name, string(flavor.Name), "")
+				metrics.ClearClusterQueueResourceReservations(oldCq.Name, string(flavor.Name), "")
+				metrics.ClearClusterQueueResourceUsage(oldCq.Name, string(flavor.Name), "")
+			} else {
+				for ri := range flavor.Resources {
+					fr := resources.FlavorResource{Flavor: flavor.Name, Resource: flavor.Resources[ri].Name}
+					if _, found := newCq.resourceNode.Quotas[fr]; !found {
+						metrics.ClearClusterQueueResourceQuotas(oldCq.Name, string(fr.Flavor), string(fr.Resource))
+						metrics.ClearClusterQueueResourceReservations(oldCq.Name, string(fr.Flavor), string(fr.Resource))
+						metrics.ClearClusterQueueResourceUsage(oldCq.Name, string(fr.Flavor), string(fr.Resource))
+					}
+				}
+			}
+		}
+	}
+}
--- a/pkg/cache/scheduler/localqueue_metrics.go
+++ b/pkg/cache/scheduler/localqueue_metrics.go
@@ -1,0 +1,43 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package scheduler
+
+import (
+	"github.com/go-logr/logr"
+
+	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+	"sigs.k8s.io/kueue/pkg/util/queue"
+)
+
+func (c *Cache) RecordLocalQueueResourceMetrics(log logr.Logger, cqName kueue.ClusterQueueReference, lqKey queue.LocalQueueReference) {
+	log.V(4).Info("Recording resource metrics for LocalQueue")
+
+	c.RLock()
+	defer c.RUnlock()
+
+	cq := c.hm.ClusterQueue(cqName)
+	if cq == nil {
+		return
+	}
+
+	lq, ok := cq.localQueues[lqKey]
+	if !ok {
+		return
+	}
+
+	lq.reportResourceMetrics(cq.resourceNode.Quotas, c.roleTracker)
+}
--- a/pkg/controller/core/clusterqueue_controller.go
+++ b/pkg/controller/core/clusterqueue_controller.go
@@ -51,9 +51,7 @@ import (
 	"sigs.k8s.io/kueue/pkg/constants"
 	"sigs.k8s.io/kueue/pkg/features"
 	"sigs.k8s.io/kueue/pkg/metrics"
-	"sigs.k8s.io/kueue/pkg/util/resource"
 	"sigs.k8s.io/kueue/pkg/util/roletracker"
-	utilslices "sigs.k8s.io/kueue/pkg/util/slices"
 	"sigs.k8s.io/kueue/pkg/workload"
 )
 
@@ -349,7 +347,7 @@ func (r *ClusterQueueReconciler) Create(e event.TypedCreateEvent[*kueue.ClusterQ
 	}
 
 	if r.reportResourceMetrics {
-		recordResourceMetrics(e.Object, r.roleTracker, r.customLabels)
+		r.cache.RecordClusterQueueResourceMetrics(log, kueue.ClusterQueueReference(e.Object.Name))
 	}
 
 	return true
@@ -411,7 +409,7 @@ func (r *ClusterQueueReconciler) Update(e event.TypedUpdateEvent[*kueue.ClusterQ
 	}
 
 	if r.reportResourceMetrics {
-		updateResourceMetrics(e.ObjectOld, e.ObjectNew, r.roleTracker, r.customLabels)
+		r.updateResourceMetrics(log, e.ObjectOld, e.ObjectNew)
 	}
 	return true
 }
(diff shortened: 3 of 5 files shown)


pkg/cache/scheduler/clusterqueue_metrics.go
```diff
@@ -0,0 +1,105 @@
...
+
+	cq := c.hm.ClusterQueue(cqName)
+	if cq == nil {
+		return
+	}
+
+	cq.reportResourceMetrics(fairSharingEnabled)
+}
+
+func (c *Cache) ClearClusterQueueResourceMetrics(log logr.Logger, cqName kueue.ClusterQueueReference) {
```
mszadkow: does it make sense to keep this separate?
Maybe we could clear before we record and just have one method?

pkg/controller/core/localqueue_controller.go
```diff
@@ -293,7 +293,7 @@ func (r *LocalQueueReconciler) Update(e event.TypedUpdateEvent[*kueue.LocalQueue
 	}
 
 	if features.Enabled(features.LocalQueueMetrics) {
-		updateLocalQueueResourceMetrics(e.ObjectNew, r.roleTracker, r.customLabels)
+		updateLocalQueueResourceMetrics(r.logger(), r.cache, e.ObjectNew)
```
mszadkow: the log is already here in line 281

pkg/controller/core/clusterqueue_controller.go
```diff
@@ -411,7 +409,7 @@ func (r *ClusterQueueReconciler) Update(e event.TypedUpdateEvent[*kueue.ClusterQ
 	}
 
 	if r.reportResourceMetrics {
-		updateResourceMetrics(e.ObjectOld, e.ObjectNew, r.roleTracker, r.customLabels)
+		updateResourceMetrics(r.logger(), r.cache, e.ObjectOld, e.ObjectNew, r.fairSharingEnabled)
```
mszadkow: log is already here up in line 375

pkg/cache/scheduler/clusterqueue_metrics.go
reruno: I created method for cache to replace metrics.ClearClusterQueueResourceMetrics call here 
pkg/controller/core/clusterqueue_controller.go:425

On the second thought I think I should revert this change and keep metrics call as it was. 

> Maybe we could clear before we record and just have one method?

Why should we clear metrics before record? I think these metrics are rewritten by reportResourceMetrics anyway

pkg/cache/scheduler/clusterqueue_metrics.go
```diff
@@ -0,0 +1,73 @@
...
+			for ri := range flavor.Resources {
+				fr := resources.FlavorResource{Flavor: flavor.Name, Resource: flavor.Resources[ri].Name}
+				if _, found := newCq.resourceNode.Quotas[fr]; !found {
+					metrics.ClearClusterQueueResourceQuotas(oldCq.Name, string(fr.Flavor), string(fr.Resource))
+					metrics.ClearClusterQueueResourceReservations(oldCq.Name, string(fr.Flavor), string(fr.Resource))
+					metrics.ClearClusterQueueResourceUsage(oldCq.Name, string(fr.Flavor), string(fr.Resource))
+				}
+			}
+		}
+	}
```
mszadkow: This logic seems to be "simplified", what happened to the comparison between old and new values.
We don't need to do that anymore?

reruno: If I understand it correctly `oldCq.Status.FlavorsUsage` and `oldCq.Status.FlavorReservation` are derived from `oldCq.Spec.ResourceGroups`, and there is no case, I think, when there is quota metric, but no reservation and usage metrics at the same time. 

Also cache clusterqueue doesn't have `.Status.FlavorReservation` and `.Spec.ResourceGroups`, so we can check only `Spec.ResourceGroups` from cache clusterqueue

reruno: @mbobrovskyi what do you think?

pkg/cache/scheduler/localqueue_metrics.go
```diff
@@ -0,0 +1,53 @@
...
+	if !ok {
+		return
+	}
+
+	lq.reportResourceMetrics(cq.resourceNode.Quotas, c.roleTracker)
+}
+
+func (c *Cache) withLocalQueueLogger(log logr.Logger, lqKey queue.LocalQueueReference) logr.Logger {
+	return log.WithValues("localQueue", lqKey)
+}
```
mbobrovskyi: What the idea of this method? Why not use log.WithValues("localQueue", lqKey) directly?

pkg/cache/scheduler/clusterqueue_metrics.go
```diff
@@ -0,0 +1,86 @@
...
+					}
+				}
+			}
+		}
+	}
+}
+
+func (c *Cache) withClusterQueueLogger(log logr.Logger, cqName kueue.ClusterQueueReference) logr.Logger {
+	return log.WithValues("clusterQueue", cqName)
+}
```
mbobrovskyi: Also here.

pkg/cache/scheduler/localqueue_metrics.go
reruno: I copied this function from file, to make new code similar to existing code
./pkg/cache/scheduler/cohort_metrics.go:115
```
func (c *Cache) withCohortLogger(log logr.Logger, cohortName kueue.CohortReference) logr.Logger {
	return log.WithValues("cohort", cohortName)
}
```

I can add log to ClearClusterQueueOldResourceQuotas, that will reuse withClusterQueueLogger, and about LocalQueue I can remove it because it is used only once, wdyt?

mbobrovskyi: I don't think so we need this redundant method.

mbobrovskyi: I think log.V(4).Info("Recording resource metrics for ClusterQueue", "clusterQueue", cqName) is much easier.

mbobrovskyi: Actually, we already set clusterQueue outside this function.

pkg/controller/core/clusterqueue_controller.go
```diff
@@ -421,118 +419,15 @@ func (r *ClusterQueueReconciler) Generic(e event.TypedGenericEvent[*kueue.Cluste
...
-		fu := &cq.Status.FlavorsUsage[fui]
-		for ri := range fu.Resources {
-			r := &fu.Resources[ri]
-			metrics.ReportClusterQueueResourceUsage(cq.Spec.CohortName, cq.Name, string(fu.Name), string(r.Name), resource.QuantityToFloat(&r.Total), cqCustomLabels, tracker)
-		}
-	}
-}
-
-func updateResourceMetrics(oldCq, newCq *kueue.ClusterQueue, tracker *roletracker.RoleTracker, customLabels *metrics.CustomLabels) {
+func updateResourceMetrics(log logr.Logger, cache *schdcache.Cache, oldCq, newCq *kueue.ClusterQueue, fairSharingEnabled bool) {
```
mbobrovskyi:
```suggestion
func (r *ClusterQueueReconciler) updateResourceMetrics(log logr.Logger, oldCq, newCq *kueue.ClusterQueue) {
```
Can we make `updateResourceMetrics` part of `ClusterQueueReconciler`?

Also, we already have the `fairSharingEnabled` parameter in the cache.

reruno: https://github.com/kubernetes-sigs/kueue/pull/9985/changes#diff-e7aa8a319ba9b1411da177aa70d1fb4d1a62e4172c84a443850090d9daf0e525R603

We would need to add ClusterQueueReconciler in the unit test, if we change it from function to method

mbobrovskyi: Is it problematic?

reruno: No just wanted to point that out

pkg/cache/scheduler/clusterqueue_metrics.go
```diff
@@ -0,0 +1,84 @@
...
+	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+	"sigs.k8s.io/kueue/pkg/metrics"
+	"sigs.k8s.io/kueue/pkg/resources"
+)
+
+func (c *Cache) RecordClusterQueueResourceMetrics(log logr.Logger, cqName kueue.ClusterQueueReference, fairSharingEnabled bool) {
+	if cqName == "" {
+		log.V(4).Info("ClusterQueue name is empty, skipping resource metrics recording")
+		return
+	}
```
mbobrovskyi: Is it possible this case?

pkg/cache/scheduler/localqueue_metrics.go
```diff
@@ -0,0 +1,48 @@
...
+
+	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+	"sigs.k8s.io/kueue/pkg/util/queue"
+)
+
+func (c *Cache) RecordLocalQueueResourceMetrics(log logr.Logger, cqName kueue.ClusterQueueReference, lqKey queue.LocalQueueReference) {
+	if lqKey == "" {
+		log.V(4).Info("LocalQueue key is empty, skipping resource metrics recording")
+		return
+	}
```
mbobrovskyi: The same question here.

pkg/cache/scheduler/clusterqueue_metrics.go
reruno: I think it is not possible, we extract clusterQueue name from objects that reconciler receives. It acts as a guard, if bad arg comes, we log it and exit. 

https://github.com/kubernetes-sigs/kueue/pull/9985/changes#diff-54fb2e1e38c1d823a077207b574c6dc9f21d0b354584becd6940b26282aec41fR430

https://github.com/kubernetes-sigs/kueue/pull/9985/changes#diff-54fb2e1e38c1d823a077207b574c6dc9f21d0b354584becd6940b26282aec41fR350

pkg/cache/scheduler/localqueue_metrics.go
reruno: Same guard, I will add the second one for cqName arg

https://github.com/kubernetes-sigs/kueue/pull/9985/changes#diff-9177ca87118e92b78007b8e08ffc92f1a18afa6fd6b4168b464e497e09ec53ecR424
https://github.com/kubernetes-sigs/kueue/pull/9985/changes#diff-9177ca87118e92b78007b8e08ffc92f1a18afa6fd6b4168b464e497e09ec53ecR261

pkg/cache/scheduler/clusterqueue_metrics.go
mbobrovskyi: If it’s not possible, I suggest removing it. We don’t need this guard – even if it’s empty, the next steps will handle it.

pkg/cache/scheduler/clusterqueue_metrics.go
```diff
@@ -0,0 +1,84 @@
...
+
+	cq := c.hm.ClusterQueue(cqName)
+	if cq == nil {
+		return
+	}
+
+	cq.reportResourceMetrics(fairSharingEnabled)
+}
+
+func (c *Cache) ClearClusterQueueOldResourceQuotas(log logr.Logger, oldCq *kueue.ClusterQueue) {
```
mbobrovskyi:
```suggestion
func (c *Cache) ClearClusterQueueResourceMetrics(log logr.Logger, oldCq *kueue.ClusterQueue) {
```

reruno: I would name it `ClearClusterQueueOldResourceMetrics`

Because the method clears old resource metrics that are not present in current  clusterqueue (newCq)

pkg/cache/scheduler/clusterqueue_metrics.go
```diff
@@ -0,0 +1,84 @@
...
+		if rgi < len(newCq.ResourceGroups) && len(newCq.ResourceGroups[rgi].Flavors) > 0 {
+			newFlavorSet.Insert(newCq.ResourceGroups[rgi].Flavors...)
+		}
+
+		for fi := range oldRg.Flavors {
+			flavor := &oldRg.Flavors[fi]
+			if !newFlavorSet.Has(flavor.Name) {
+				metrics.ClearClusterQueueResourceQuotas(oldCq.Name, string(flavor.Name), "")
+				metrics.ClearClusterQueueResourceReservations(oldCq.Name, string(flavor.Name), "")
+				metrics.ClearClusterQueueResourceUsage(oldCq.Name, string(flavor.Name), "")
```
mbobrovskyi, line 66: I’m wondering if we can keep only this.

reruno: No, when flavor is present in new cq and old cq, we need to check if old resource in existing flavor is present in new resources (`newCq.resourceNode.Quotas[fr]`), if old resource is not present it will be cleared.

We can clear all flavors regardless if it is present in new cq or not, but it will be a little bit less efficient

pkg/controller/core/clusterqueue_controller_test.go
```diff
@@ -556,14 +576,32 @@ func TestRecordResourceMetrics(t *testing.T) {
...
 			}
 
 			if tc.updatedQueue != nil {
-				updateResourceMetrics(tc.queue, tc.updatedQueue, nil, nil)
+				wl := workloadForReservation("name", tc.updatedQueue.Status.FlavorsReservation, time.Now())
+				cqCache.AddOrUpdateWorkload(log, wl)
+				if err := cqCache.UpdateClusterQueue(log, tc.updatedQueue); err != nil {
+					t.Fatalf("Updating clusterQueue in cache: %v", err)
+				}
+				r.updateResourceMetrics(ctrl.Log, tc.queue, tc.updatedQueue)
```
mbobrovskyi:
```suggestion
				r.updateResourceMetrics(log, tc.queue, tc.updatedQueue)
```

pkg/controller/core/clusterqueue_controller_test.go
```diff
@@ -335,6 +335,23 @@ func resourceDataPoint(cohort, name, flavor, res string, v float64) testingmetri
 	}
 }
 
+func workloadForReservation(cqName string, reservation []kueue.FlavorUsage, t time.Time) *kueue.Workload {
+	psa := utiltestingapi.MakePodSetAssignment(kueue.DefaultPodSetName)
```
mbobrovskyi:
```suggestion
func workloadForReservation(cqName string, reservation []kueue.FlavorUsage) *kueue.Workload {
	now := time.Now()
	psa := utiltestingapi.MakePodSetAssignment(kueue.DefaultPodSetName)
```
IIUC, we don’t need to propagate time – we can just use now inside function.

pkg/controller/core/clusterqueue_controller_test.go
```diff
@@ -556,14 +576,32 @@ func TestRecordResourceMetrics(t *testing.T) {
...
 		t.Run(name, func(t *testing.T) {
-			recordResourceMetrics(tc.queue, nil, nil)
+			ctx, log := utiltesting.ContextWithLog(t)
+
+			cl := utiltesting.NewClientBuilder().Build()
+			cqCache := schdcache.New(cl)
+			r := &ClusterQueueReconciler{cache: cqCache}
+			err := cqCache.AddClusterQueue(ctx, tc.queue.DeepCopy())
+			if err != nil {
+				t.Errorf("Unexpected error err=%s", err.Error())
```
mbobrovskyi:
```suggestion
				t.Fatalf("Unexpected error: %v", err)
```

pkg/controller/core/clusterqueue_controller_test.go
```diff
@@ -556,14 +576,32 @@ func TestRecordResourceMetrics(t *testing.T) {
 
 	for name, tc := range testCases {
 		t.Run(name, func(t *testing.T) {
-			recordResourceMetrics(tc.queue, nil, nil)
+			ctx, log := utiltesting.ContextWithLog(t)
+
+			cl := utiltesting.NewClientBuilder().Build()
+			cqCache := schdcache.New(cl)
+			r := &ClusterQueueReconciler{cache: cqCache}
+			err := cqCache.AddClusterQueue(ctx, tc.queue.DeepCopy())
```
mbobrovskyi:
```suggestion
			err := cqCache.AddClusterQueue(ctx, tc.queue)
```
?

pkg/controller/core/localqueue_controller.go
```diff
@@ -259,13 +258,16 @@ func (r *LocalQueueReconciler) Create(e event.TypedCreateEvent[*kueue.LocalQueue
...
-		recordLocalQueueUsageMetrics(e.Object, r.roleTracker, r.customLabels)
+		r.cache.RecordLocalQueueResourceMetrics(log, e.Object.Spec.ClusterQueue, utilqueue.Key(e.Object))
 	}
 
 	return true
 }
 
 func (r *LocalQueueReconciler) Delete(e event.TypedDeleteEvent[*kueue.LocalQueue]) bool {
+	log := r.logger().WithValues("localQueue", klog.KObj(e.Object))
+	log.V(2).Info("LocalQueue delete event")
```
mbobrovskyi: Is this required changes?

pkg/controller/core/clusterqueue_controller_test.go
```diff
@@ -556,14 +577,32 @@ func TestRecordResourceMetrics(t *testing.T) {
...
 		t.Run(name, func(t *testing.T) {
-			recordResourceMetrics(tc.queue, nil, nil)
+			ctx, log := utiltesting.ContextWithLog(t)
+
+			cl := utiltesting.NewClientBuilder().Build()
+			cqCache := schdcache.New(cl)
+			r := &ClusterQueueReconciler{cache: cqCache}
+			err := cqCache.AddClusterQueue(ctx, tc.queue)
+			if err != nil {
+				t.Fatalf("Unexpected error err=%v", err)
```
mbobrovskyi:
```suggestion
				t.Fatalf("Unexpected error: %v", err)
```
For consistency with other places.

reruno:
@mimowo I received LGTM from the team, do you have any additional comments for the PR?

mimowo:
cc @IrvingMg @mbobrovskyi could you take another pass after the rebasse?

mimowo:
/approve
Thank you 👍 
/hold
Let's retry 3 times to see if not related flakes to metrics. Feel free to release the hold once tested
/retest 
(1/3)

mimowo (commented):
Thank you 👍 
/lgtm
/approve
/unhold

reruno:
/retest 

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/9985/pull-kueue-test-e2e-main-1-35/2051647188466929664
error: timed out waiting for the condition on deployments/prometheus-operator

mimowo:
/unhold

> https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/9985/pull-kueue-test-e2e-main-1-35/2051647188466929664
> error: timed out waiting for the condition on deployments/prometheus-operator

@reruno could you open a flake issue for that?

PR #4561, opened 2025-03-12
Cohort Fair Sharing Status and Metrics.

mbobrovskyi:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
Cohort Fair Sharing Status and Metrics

#### Which issue(s) this PR fixes:

Fixes #4554

#### Does this PR introduce a user-facing change?

```release-note
Added Cohort Fair Sharing Status and Metrics
```

PR #9999, opened 2026-03-19
Introduce kueue_cohort_admitted_workloads_total metric.

mbobrovskyi:
#### What type of PR is this?
/kind feature

Introduce kueue_cohort_admitted_workloads_total metric.

Part of #7539

```release-note
Observability: Introduce kueue_cohort_admitted_workloads_total metric
```

### DIFF
--- a/pkg/cache/scheduler/cache.go
+++ b/pkg/cache/scheduler/cache.go
@@ -784,11 +784,44 @@ func (c *Cache) ClusterQueueAncestors(cqObj *kueue.ClusterQueue) ([]kueue.Cohort
 	c.RLock()
 	defer c.RUnlock()
 
-	if cqObj.Spec.CohortName == "" {
+	ancestors, err := c.ancestors(cqObj.Spec.CohortName)
+	if err != nil {
+		return nil, err
+	}
+
+	// Exclude the root cohort
+	if len(ancestors) > 0 {
+		ancestors = ancestors[:len(ancestors)-1]
+	}
+
+	return ancestors, nil
+}
+
+// workloadAncestors retrieves all ancestor Cohorts for a given Workload.
+// The caller MUST hold either c.RLock() or c.Lock() before calling this function.
+// This function does not acquire any locks internally and assumes the cache is already locked.
+func (c *Cache) workloadAncestors(wl *kueue.Workload) ([]kueue.CohortReference, error) {
+	if wl == nil || wl.Status.Admission == nil {
 		return nil, nil
 	}
 
-	cohort := c.hm.Cohort(cqObj.Spec.CohortName)
+	cq := c.hm.ClusterQueue(wl.Status.Admission.ClusterQueue)
+	if cq == nil || !cq.HasParent() {
+		return nil, nil
+	}
+
+	return c.ancestors(cq.Parent().Name)
+}
+
+// ancestors retrieves all ancestor Cohorts for a given Cohort, starting with the Cohort itself and ending at the root.
+// The caller MUST hold either c.RLock() or c.Lock() before calling this function.
+// This function does not acquire any locks internally and assumes the cache is already locked.
+func (c *Cache) ancestors(cohortName kueue.CohortReference) ([]kueue.CohortReference, error) {
+	if cohortName == "" {
+		return nil, nil
+	}
+
+	cohort := c.hm.Cohort(cohortName)
 	if cohort == nil {
 		return nil, nil
 	}
@@ -802,7 +835,7 @@ func (c *Cache) ClusterQueueAncestors(cqObj *kueue.ClusterQueue) ([]kueue.Cohort
 		ancestors = append(ancestors, ancestor.Name)
 	}
 
-	return ancestors[:len(ancestors)-1], nil
+	return ancestors, nil
 }
 
 func getUsage(frq resources.FlavorResourceQuantities, cq *clusterQueue) []kueue.FlavorUsage {
--- a/pkg/cache/scheduler/cohort_metrics.go
+++ b/pkg/cache/scheduler/cohort_metrics.go
@@ -18,11 +18,13 @@ package scheduler
 
 import (
 	"github.com/go-logr/logr"
+	"k8s.io/klog/v2"
 
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
 	"sigs.k8s.io/kueue/pkg/cache/hierarchy"
 	"sigs.k8s.io/kueue/pkg/metrics"
 	"sigs.k8s.io/kueue/pkg/resources"
+	"sigs.k8s.io/kueue/pkg/workload"
 )
 
 type cohortMetricPoint struct {
@@ -114,4 +116,24 @@ func (c *Cache) applyCohortMetricPoint(p cohortMetricPoint) {
 		float64(p.qty),
 		c.roleTracker,
 	)
+}
+
+func (c *Cache) ReportCohortSubtreeAdmittedWorkload(log logr.Logger, wl *kueue.Workload) {
+	c.RLock()
+	defer c.RUnlock()
+
+	ancestors, err := c.workloadAncestors(wl)
+	if err != nil {
+		log.Error(err, "Failed getting ancestors for workload", "workload", klog.KObj(wl))
+		return
+	}
+
+	for _, ancestor := range ancestors {
+		metrics.ReportCohortSubtreeAdmittedWorkload(
+			ancestor,
+			workload.PriorityClassName(wl),
+			c.customLabels.CohortGet(ancestor),
+			c.roleTracker,
+		)
+	}
 }
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -254,6 +254,10 @@ var (
 	// +metricsdoc:group=cohort
 	// +metricsdoc:labels=cohort="the name of the Cohort",flavor="the resource flavor name",resource="the resource name",replica_role="one of `leader`, `follower`, or `standalone`"
 	CohortSubtreeQuota *prometheus.GaugeVec
+
+	// +metricsdoc:group=cohort
+	// +metricsdoc:labels=cohort="the name of the Cohort",priority_class="the priority class name",replica_role="one of `leader`, `follower`, or `standalone`"
+	CohortSubtreeAdmittedWorkloadsTotal *prometheus.CounterVec
 )
 
 func trackGaugeVec(g *prometheus.GaugeVec) *prometheus.GaugeVec {
@@ -728,6 +732,14 @@ If the Cohort has a weight of zero and is borrowing, this will return NaN.`,
 			Help:      `Reports the cohort's nominal quota aggregated within the cohort's subtree. The values are reported per resource and flavor`,
 		}, []string{"cohort", "flavor", "resource", "replica_role"},
 	))
+
+	CohortSubtreeAdmittedWorkloadsTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: constants.KueueName,
+			Name:      "cohort_subtree_admitted_workloads_total",
+			Help:      "The total number of admitted workloads per cohort's subtree",
+		}, append([]string{"cohort", "priority_class", "replica_role"}, extraLabels...),
+	)
 }
 
 func init() {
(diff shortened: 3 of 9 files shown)


test/integration/singlecluster/scheduler/fairsharing/fair_sharing_test.go
```diff
@@ -368,6 +368,7 @@ var _ = ginkgo.Describe("Scheduler", ginkgo.Label("feature:fairsharing"), func()
 			ginkgo.By("Creating two workloads in cqA")
 			wlA1 := createWorkloadWithPriority(cqA.Name, "4", 9001)
 			wlA2 := createWorkloadWithPriority(cqA.Name, "4", 100)
+			util.ExpectCohortSubtreeAdmittedWorkloadsTotalMetric("top-cohort", "", 2)
```
mszadkow: We should try to keep all metrics relevant tests in `test/integration/singlecluster/scheduler/fairsharing/metrics_test.go`

pkg/cache/scheduler/cohort_metrics.go
```diff
@@ -115,3 +116,23 @@ func (c *Cache) applyCohortMetricPoint(p cohortMetricPoint) {
...
+	defer c.RUnlock()
+
+	ancestors, err := c.workloadAncestors(wl)
+	if err != nil {
+		log.Error(err, "Failed getting ancestors for workload", "workload", wl.Name)
+		return
+	}
+
+	for _, ancestor := range ancestors {
+		metrics.ReportCohortSubtreeAdmittedWorkload(
```
mszadkow, line 132: So, if I get this right.
For the workload in the given clusterqueue, we check the chain up to the ancestor and we increment counter for each ancestor.

pkg/cache/scheduler/cache.go
```diff
@@ -784,14 +784,42 @@ func (c *Cache) ClusterQueueAncestors(cqObj *kueue.ClusterQueue) ([]kueue.Cohort
...
+	// Exclude the root cohort
+	if len(ancestors) > 0 {
+		ancestors = ancestors[:len(ancestors)-1]
+	}
+
+	return ancestors, nil
+}
+
+func (c *Cache) workloadAncestors(wl *kueue.Workload) ([]kueue.CohortReference, error) {
+	if wl == nil || wl.Status.Admission == nil {
```
mimowo, line 804: I think this is checking more that admitted - also quota reserved and evicted before the Admission is deleted. Please also double check what workloads are we counting in admitted_workloads_total.

I think for reserving workloads we should have `cohort_subtree_reserving_workloads`

pkg/cache/scheduler/cache.go
```diff
@@ -784,14 +784,42 @@ func (c *Cache) ClusterQueueAncestors(cqObj *kueue.ClusterQueue) ([]kueue.Cohort
...
-	if cqObj.Spec.CohortName == "" {
+	ancestors, err := c.ancestors(cqObj.Spec.CohortName)
+	if err != nil {
+		return nil, err
+	}
+
+	// Exclude the root cohort
+	if len(ancestors) > 0 {
+		ancestors = ancestors[:len(ancestors)-1]
+	}
```
mimowo, line 795: Actually, why excluding the root cohort?

pkg/cache/scheduler/cohort_metrics.go
```diff
@@ -115,3 +116,23 @@ func (c *Cache) applyCohortMetricPoint(p cohortMetricPoint) {
 		c.roleTracker,
 	)
 }
+
+func (c *Cache) ReportCohortSubtreeAdmittedWorkload(log logr.Logger, wl *kueue.Workload) {
+	c.RLock()
+	defer c.RUnlock()
+
+	ancestors, err := c.workloadAncestors(wl)
+	if err != nil {
```
mimowo, line 126: ErrCohortHasCycle is the only possible error here. I would suggest not to return the error at all, and log it when we detect the cycle, because now we have the errors propagated unnecessarily. We don't support cycles anyway,.

pkg/cache/scheduler/cache.go
mbobrovskyi: Yeah, we take a decision to skip this in https://github.com/kubernetes-sigs/kueue/pull/4561#discussion_r1993495434.

mimowo: I see, the decision certainly makes sense for fair sharing, because the root cohort has no other cohort to share resources with.

Here for the root cohort I'm less sure. Let's say you have two root cohorts (independent trees) and you want to know the number of workloads in each of them, then the user would need to aggregate non-trivialy.

mbobrovskyi: I created a separate helper that returns a list of ancestors for workloads, including the root, to avoid duplicating code. But to keep the existing logic, we need to skip the root cohort in ClusterQueueAncestors(). That’s why I added this if. I’m not using ClusterQueueAncestors() in the report.

mimowo: >  But to keep the existing logic, we need to skip the root cohort in ClusterQueueAncestors().

What do you mean by the existing logic, can you provide a code pointer?

pkg/cache/scheduler/cache.go
mbobrovskyi: We have an admitted check in workload_controller where we execute the report function. But here, I don’t think we need to add this logic, since this is just a helper used in ReportCohortSubtreeAdmittedWorkload().

pkg/cache/scheduler/cache.go
mbobrovskyi: Yes, here is one:

https://github.com/kubernetes-sigs/kueue/blob/2bef19904ac7426af01d74a5d439052c7bff9a1e/pkg/cache/scheduler/cache.go#L805

pkg/cache/scheduler/cache.go
mimowo: Let me maybe clarify / rephrase the question:
1. is the pre-existing metric AdmittedWorkloadsTotal (for ClusterQueue) counting workloads which have QuotaReserved=true, but Admitted=False
2. is the new metric `CohortSubtreeAdmittedWorkload` counting workloads which have QuotaReserved=true, but Admitted=False

Do we have tests which demonstrate if such workloads are included or excluded from the metric?

mbobrovskyi: > is the pre-existing metric AdmittedWorkloadsTotal (for ClusterQueue) counting workloads which have QuotaReserved=true, but Admitted=False

Now we report AdmittedWorkloadsTotal only if Admitted=True.

https://github.com/epam/kubernetes-kueue/blob/4b862c0e50080e7eaa5e004e69e636b43c9186be/pkg/controller/core/workload_controller.go#L494-L500

https://github.com/epam/kubernetes-kueue/blob/4b862c0e50080e7eaa5e004e69e636b43c9186be/pkg/scheduler/scheduler.go#L907-L915

> is the new metric CohortSubtreeAdmittedWorkload counting workloads which have QuotaReserved=true, but Admitted=False

No, we report CohortSubtreeAdmittedWorkload in the same place.

mimowo: ok do we ha e an integration test to confiem that? lets make sure we have

pkg/metrics/metrics.go
```diff
@@ -728,6 +732,14 @@ If the Cohort has a weight of zero and is borrowing, this will return NaN.`,
 			Help:      `Reports the cohort's nominal quota aggregated within the cohort's subtree. The values are reported per resource and flavor`,
 		}, []string{"cohort", "flavor", "resource", "replica_role"},
 	))
+
+	CohortSubtreeAdmittedWorkloadsTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: constants.KueueName,
+			Name:      "cohort_subtree_admitted_workloads_total ",
```
mwielgus: Remove trailing space.

site/content/en/docs/reference/metrics.md
```diff
@@ -82,6 +82,7 @@ The following metrics are available only if `LocalQueueMetrics` feature gate is
 <!-- BEGIN GENERATED TABLE: cohort -->
 | Metric name | Type | Description | Labels |
 | --- | --- | --- | --- |
+| `kueue_cohort_subtree_admitted_workloads_total ` | Counter | The total number of admitted workloads per cohort's subtree | `cohort`: the name of the Cohort<br> `priority_class`: the priority class name<br> `replica_role`: one of `leader`, `follower`, or `standalone` |
```
mwielgus: Trailing space.

pkg/cache/scheduler/cohort_metrics.go
```diff
@@ -115,3 +116,23 @@ func (c *Cache) applyCohortMetricPoint(p cohortMetricPoint) {
...
+		log.Error(err, "Failed getting ancestors for workload", "workload", wl.Name)
```
mwielgus: Use kobj(w) so that the namespace is included as well.

pkg/cache/scheduler/cache.go
```diff
@@ -784,14 +784,42 @@ func (c *Cache) ClusterQueueAncestors(cqObj *kueue.ClusterQueue) ([]kueue.Cohort
...
+
+	return ancestors, nil
+}
+
+func (c *Cache) workloadAncestors(wl *kueue.Workload) ([]kueue.CohortReference, error) {
```
mwielgus, line 803: What is the locking contract here?

pkg/cache/scheduler/cache.go
```diff
@@ -784,14 +784,42 @@ func (c *Cache) ClusterQueueAncestors(cqObj *kueue.ClusterQueue) ([]kueue.Cohort
...
-	cohort := c.hm.Cohort(cqObj.Spec.CohortName)
+	cq := c.hm.ClusterQueue(wl.Status.Admission.ClusterQueue)
+	if cq == nil || !cq.HasParent() {
+		return nil, nil
+	}
+
+	return c.ancestors(cq.Parent().Name)
+}
+
+func (c *Cache) ancestors(cohortName kueue.CohortReference) ([]kueue.CohortReference, error) {
```
mwielgus, line 819: Document the locking contract here.

pkg/metrics/metrics.go
mbobrovskyi: Thank you!

pkg/cache/scheduler/cohort_metrics.go
mbobrovskyi: Done

pkg/cache/scheduler/cache.go
mbobrovskyi: Added.

pkg/cache/scheduler/cache.go
mbobrovskyi: Added.

test/integration/singlecluster/scheduler/fairsharing/fair_sharing_test.go
mbobrovskyi: I’m actually thinking that for this metric it’s better to keep it here instead of creating more tests, to keep it close to `ExpectAdmittedWorkloadsTotalMetric()`.

pkg/cache/scheduler/cohort_metrics.go
mbobrovskyi: Yes, exactly.

pkg/cache/scheduler/cohort_metrics.go
mbobrovskyi: Do you propose propagating the logger to the ancestors() function or just to workloadAncestors()? Because if it’s ancestors(), we would also need to change the current logic.

pkg/cache/scheduler/cache.go
mbobrovskyi: Added test case.

test/integration/singlecluster/scheduler/fairsharing/fair_sharing_test.go
mbobrovskyi: Also added separate test case with admission check.

mbobrovskyi: OK. Moved all tests to metrics.go.

pkg/cache/scheduler/cache.go
mimowo: sgtm, I see the metric is generated for the roots: https://github.com/kubernetes-sigs/kueue/pull/9999/changes#diff-119d4450e4fb8bfe8786ea5c0ff81ee57e31eea554356d79a7a16dd4e1e83d9fR617-R618

Yeah, I don't want to change the logic for the other metric, it is good that we skip the root there. 

For this metric knowing the number of workloads in the top cohort makes sense.

PR #10001, opened 2026-03-19
Fix scheduling hashes and disable temporarily

mimowo:
#### What type of PR is this?

/kind bug

To fix the problem observed on large deployments with FairSharing enabled that the hash-based bulk scheduling was no-op, because in that case allmost all inadmissible workloads are categorized as "Preempt", but with noPreemptionCandidates.

This change uses the bulk-scheduling if and only if a pair of workloads have identical scheduling hash, and the "Popped" workload is considered "inadmissible", regardless of the categorization.

At the same time, since this change has a bigger impact, we need more time to:
1. properly investigate the performance regression reported by our testing framework, this is probably just an artifact of the testing approach, but we should be careful
2. Fix the compatibility with MultiKueue preemption gates, cc @kshalot 
3. Adjust many unit and integration tests carefully

So, I propose to disable the feature gate by default.

```release-note
Scheduling: fixed SchedulingEquivalenceHashing so equivalent workloads that become inadmissible through
the preemption path with no candidates are also covered by the mechanism. 

As a safety measure while the broader fix is validated, the beta SchedulingEquivalenceHashing feature gate
is temporarily disabled by default. 
```

### DIFF
--- a/pkg/cache/queue/cluster_queue.go
+++ b/pkg/cache/queue/cluster_queue.go
@@ -413,6 +413,12 @@ func (c *ClusterQueue) requeueIfNotPresent(log logr.Logger, wInfo *workload.Info
 	}
 	log.V(2).Info(logMsg, "clusterQueue", c.name, "workload", key)
 
+	if features.Enabled(features.SchedulingEquivalenceHashing) && wInfo.SchedulingHash != workload.SchedulingHashUnknown && !immediate {
+		if moved := c.handleInadmissibleHash(wInfo.SchedulingHash); moved > 0 {
+			log.V(2).Info("Bulk-moved equivalent workloads to inadmissible", "hash", wInfo.SchedulingHash, "movedCount", moved)
+		}
+	}
+
 	return true
 }
 
@@ -427,8 +433,6 @@ func (c *ClusterQueue) forgetInflightByKey(key workload.Reference) {
 // Only applies to BestEffortFIFO queues; in StrictFIFO the head workload
 // stays in the heap and must not cause equivalent workloads to be skipped.
 func (c *ClusterQueue) handleInadmissibleHash(hash string) int {
-	c.rwm.Lock()
-	defer c.rwm.Unlock()
 	if c.queueingStrategy != kueue.BestEffortFIFO {
 		return 0
 	}
--- a/pkg/cache/queue/manager.go
+++ b/pkg/cache/queue/manager.go
@@ -601,23 +601,6 @@ func (m *Manager) RequeueWorkload(ctx context.Context, info *workload.Info, reas
 	return added
 }
 
-// HandleInadmissibleHash bulk-moves all workloads in the ClusterQueue's heap
-// that share the given scheduling hash to inadmissibleWorkloads.
-func (m *Manager) HandleInadmissibleHash(cqName kueue.ClusterQueueReference, hash string) int {
-	m.RLock()
-	defer m.RUnlock()
-	cq := m.hm.ClusterQueue(cqName)
-	if cq == nil {
-		return 0
-	}
-	moved := cq.handleInadmissibleHash(hash)
-	if moved > 0 {
-		// Update pending metrics for the CQ and all its LocalQueues.
-		reportPendingWorkloads(m, cqName)
-	}
-	return moved
-}
-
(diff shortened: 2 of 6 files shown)


mimowo:
cc @mukund-wayve @sohankunkerkar @tenzen-y @gabesaba

pkg/cache/queue/manager.go
```diff
@@ -601,23 +601,6 @@ func (m *Manager) RequeueWorkload(ctx context.Context, info *workload.Info, reas
...
-	m.RLock()
-	defer m.RUnlock()
-	cq := m.hm.ClusterQueue(cqName)
-	if cq == nil {
-		return 0
-	}
-	moved := cq.handleInadmissibleHash(hash)
-	if moved > 0 {
-		// Update pending metrics for the CQ and all its LocalQueues.
-		reportPendingWorkloads(m, cqName)
```
PBundyra, line 616: Do we intentionally want remove reporting?

pkg/cache/queue/cluster_queue.go
```diff
@@ -413,6 +413,12 @@ func (c *ClusterQueue) requeueIfNotPresent(log logr.Logger, wInfo *workload.Info
 	}
 	log.V(2).Info(logMsg, "clusterQueue", c.name, "workload", key)
 
+	if features.Enabled(features.SchedulingEquivalenceHashing) && wInfo.SchedulingHash != workload.SchedulingHashUnknown && !immediate {
```
kshalot, line 416: The reason why this is not working with preemption gates is because when a preemption gate is hit, the requeue reason is set to `RequeueReasonPreemptionGated`:
https://github.com/kubernetes-sigs/kueue/blob/5c3974bcdbcc5b1f476ce2dfe4dea6860f0adcc5/pkg/scheduler/scheduler.go#L215
which uses `immediate=False` when requeueing because we don't want the gated workloads to block others while waiting for the gate to be released. But now this makes the hash end up in `noFitSchedulingHashes`:
https://github.com/kubernetes-sigs/kueue/blob/5c3974bcdbcc5b1f476ce2dfe4dea6860f0adcc5/pkg/cache/queue/cluster_queue.go#L435

Before this PR we only executed `handleInadmissibleHash` when the mode was `NoFit`, so this made sense. But now we schedule it when the requeueing mode  is `immediate=False`. So the `NoFit` and `immediate` got coupled.

I think if we want to hack it for the release, we could just pass some extra boolean instead of re-using `immediate`.

pkg/cache/queue/manager.go
mimowo: I think this reporting here is not needed, because we already call reportCQPendingWorkloads from within `RequeueWorkload`, just after calling `RequeueIfNotPresent`. I think it might have been needed before when the change was in a different place. I would revisit the merics carefully as part of the enablement plan to Beta.

pkg/cache/queue/cluster_queue.go
mimowo: Is an alernative fix to simply add the `preemptionGates` to the scheduling hash? Then, the next workload will be evaluated still, if the state of the gates is different.

mimowo: Still, the MultiKueue preemption gates are still Alpha and 0.17+, so I think we have time to fix it as a follow up

kshalot: Now that I think of it, it's going to be more subtle then that. Adding the preemption gate to the hash would cause problems with signaling, because we will bulk move gated workloads which need the scheduler to pick them up and add the gated condition. In theory if we use the states of the gates that include the timestamps then it would "work", but it's hacky.

kshalot: > Still, the MultiKueue preemption gates are still Alpha and 0.17+, so I think we have time to fix it as a follow up

Yes, I think we should handle it as a follow-up, because it's delicate.

pkg/cache/queue/manager.go
PBundyra: I see, thanks

mimowo:
I opened the issue to follow up with graduation: https://github.com/kubernetes-sigs/kueue/issues/10005

PR #9132, opened 2026-02-11
Provide a metric for cohort nominal quota

mszadkow:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
Extend metrics to cohort level.

Add metrics for nominal quota.

#### Which issue(s) this PR fixes:

Relates to #7539

#### Does this PR introduce a user-facing change?

```release-note
Observability:  Introduce the cohort nominal quota metric.
```

PR #10004, opened 2026-03-19
Provide a metric for cohort and clusterqueue info

mszadkow:
#### What type of PR is this?
/kind feature

Add metrics for cohort's and clusterqueue's info.

Relates #7539

```release-note
Observability:  Introduce the cohort info metrics and clusterqueue info metric.
```


pkg/cache/scheduler/cache.go
```diff
@@ -520,19 +522,52 @@ func (c *Cache) AddOrUpdateCohort(apiCohort *kueue.Cohort) error {
...
+		metrics.ClearCohortInfo(oldCohort.Name)
+	}
+}
+
+func (c *Cache) reportInfoMetricsSubtree(root *cohort) {
+	if root == nil {
+		return
+	}
+
+	stack := []*cohort{root}
```
mwielgus: What if there is a cycle in cohort tree?

pkg/cache/scheduler/cache.go
```diff
@@ -520,19 +522,52 @@ func (c *Cache) AddOrUpdateCohort(apiCohort *kueue.Cohort) error {
 	cohort := c.hm.Cohort(cohortName)
 	oldParent := cohort.Parent()
 	c.hm.UpdateCohortEdge(cohortName, apiCohort.Spec.ParentName)
-	return cohort.updateCohort(apiCohort, oldParent)
+	if err := cohort.updateCohort(apiCohort, oldParent); err != nil {
+		return err
+	}
+	c.reportInfoMetricsSubtree(cohort)
```
mwielgus: If apiCohort.Spec.ParentName refers to a cohort that doesn't exist yet, UpdateCohortEdge will implicitly create it. However reportInfoMetricsSubtree is only called on the child cohort, meaning the newly created implicit parent cohort will never have its cohort_info metric reported until something else explicitly triggers it. You probably should report the metric for the parent as well

pkg/cache/scheduler/cache.go
```diff
@@ -520,19 +522,52 @@ func (c *Cache) AddOrUpdateCohort(apiCohort *kueue.Cohort) error {
...
+	return nil
```
mwielgus, line 617: If oldParent != nil and cohort was its final remaining child, UpdateCohortEdge will delete oldParent (since it was an implicit cohort and is now empty). When this happens, nothing triggers a metric cleanup for it, causing oldParent's cohort_info metric to leak permanently.

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
```diff
@@ -406,6 +438,9 @@ var _ = ginkgo.Describe("Cohorts", func() {
 				// cleared up values for ch0 with cq1 removed
 				util.ExpectCohortSubtreeQuotaGaugeMetricCleaned("ch0", flavor1.Name, corev1.ResourceCPU.String())
 				util.ExpectCohortSubtreeQuotaGaugeMetricCleaned("ch0", flavor1.Name, "nvidia.com/gpu")
+
+				util.ExpectClusterQueueInfoMetric("cq1", "ch1", "ch1")
+				util.ExpectClusterQueueInfoMetricCleaned("cq1", "ch0", "ch0")
```
mwielgus: Add a test assertion here for util.ExpectCohortInfoMetric("ch1", "", "ch1"). Because cq1 was moved away from ch1, we expect ch1 to still be intact.

mszadkow: we moved cq1 from ch0 to ch1, so maybe we check if ch0 was removed as it was implicit:
`util.ExpectCohortInfoMetricCleaned("ch0", "", "ch0")`

We later re-assign cq1 from ch1 to ch2, so yes we can check ch1 as well, but it was explicit cohort, so should be fine

pkg/cache/scheduler/cache.go
mszadkow: worth to check, I agree

pkg/cache/scheduler/cache.go
mszadkow: I will make sure it's reporting for the implicit cohort creation too, also the test coverage

pkg/cache/scheduler/cache.go
mszadkow: ok, so we talk about the situation when cq gets "reparented" to another cohort, but the old cohort was implicit and now should be gone, thus the metric should be cleared too

pkg/cache/scheduler/cache.go
mszadkow: Actually not a chance to get here if there is a cycle, as its called only in AddOrUpdateCohort or DeleteCohort, but I can provide an extra check, so even in the future we are certain

pkg/cache/scheduler/cache.go
mszadkow: done

pkg/cache/scheduler/cache.go
mszadkow: Added a test to cover both situations:
TestAddOrUpdateCohort_ImplicitParentMetricsReporting

pkg/cache/scheduler/cache.go
mszadkow: check provided

test/util/metrics.go
```diff
@@ -324,7 +324,14 @@ func ExpectCohortSubtreeResourceReservationsGaugeMetric(cohortName string, flavo
 	expectGaugeMetric(metrics.CohortSubtreeResourceReservations, lvs, gomega.Equal(count))
 }
 
-func ExpectCohortSubtreeResourceReservationsGaugeMetricCleaned(cohortName, flavor, resource string) {
+func ExpectCohortInfoMetric(cohortName, parentCohort, rootCohort string, count float64) {
 	ginkgo.GinkgoHelper()
-	ExpectCohortSubtreeResourceReservationsGaugeMetric(cohortName, flavor, resource, 0)
```
mbobrovskyi: I’m not sure why we removed this. Is it required for this issue?

pkg/metrics/metrics.go
```diff
@@ -939,6 +963,7 @@ func ClearLocalQueueMetrics(lq LocalQueueReference) {
 }
 
 func ClearCohortMetrics(cohortName kueue.CohortReference) {
+	CohortInfo.DeletePartialMatch(prometheus.Labels{"cohort": string(cohortName)})
```
mbobrovskyi: I’m not sure this is the correct place. As we discussed in https://github.com/kubernetes-sigs/kueue/pull/10088#issuecomment-4124852542, we should delete this kind of metric only on the delete event. Otherwise, we might lose it when custom labels change.

mbobrovskyi: Please also add this check if the `CustomLabels` feature is enabled.

pkg/cache/scheduler/cache.go
```diff
@@ -530,19 +532,67 @@ func (c *Cache) AddOrUpdateCohort(apiCohort *kueue.Cohort) error {
...
 	oldParent := cohort.Parent()
 	c.hm.UpdateCohortEdge(cohortName, apiCohort.Spec.ParentName)
-	return cohort.updateCohort(apiCohort, oldParent)
+	if err := cohort.updateCohort(apiCohort, oldParent); err != nil {
+		return err
+	}
+
+	// If reparenting removed an implicit old parent, there should be no cohort present in the cache,
+	// clear leaked info metric.
+	if oldParent != nil && oldParent != cohort.Parent() && c.hm.Cohort(oldParent.Name) == nil {
```
mbobrovskyi: Should we compare parent name?

mbobrovskyi: I’m wondering if we could return a bool (indicating whether it was updated) from updateCohort().

test/util/metrics.go
mszadkow: we don't need a separate helper with `Cleaned` postfix.
It can be done with `ExpectCohortSubtreeResourceReservationsGaugeMetric` instead.

mbobrovskyi: Yeah, I got but is it required for this PR?

pkg/metrics/metrics.go
mszadkow: ahh custom_labels, I didn't investigate this from that point of view, I will make sure it's safe

pkg/cache/scheduler/cache.go
mszadkow: the update itself is not enough, though might be valuable to have this info
This situation is very specific, as we changed the parent but it was implicit cohort, so won't be reconciled 
we make sure the parent was removed from the cache and we should trigger metric cleanup for it, just to avoid leaving stale metric

test/util/metrics.go
mszadkow: The removal of the helper?
I will extract it

pkg/metrics/metrics.go
mszadkow: done

test/util/metrics.go
mszadkow: done

pkg/cache/scheduler/cache.go
```diff
@@ -541,13 +559,44 @@ func (c *Cache) DeleteCohort(cohortName kueue.CohortReference) {
...
+	if oldCohort != nil {
+		metrics.ClearCohortInfo(oldCohort.Name)
+	}
+}
+
+func (c *Cache) reportInfoMetricsSubtree(root *cohort) {
+	if root == nil {
+		return
+	}
+	if hierarchy.HasCycle(root) {
```
```suggestion
	if root == nil || hierarchy.HasCycle(root) {
```
nit

pkg/cache/scheduler/cache.go
```diff
@@ -1018,14 +1073,15 @@ func (c *Cache) ResyncGaugeMetrics() {
 			}
 		}
 	}
+
+	for _, cohort := range c.hm.Cohorts() {
+		cohort.reportInfoMetric(c.customLabels.CohortGet(cohort.Name), c.roleTracker)
+	}
```
mbobrovskyi: Hmm, it looks like we have duplicate loops. Can we merge this with the one below?

test/util/metrics.go
```diff
@@ -340,3 +340,27 @@ func ExpectCohortSubtreeAdmittedActiveWorkloadsGaugeMetric(cohortName kueue.Coho
...
+	ginkgo.GinkgoHelper()
+	lvs := append([]string{cohortName, parentCohort, rootCohort, roletracker.RoleStandalone}, customLabels...)
+	expectGaugeMetric(metrics.CohortInfo, lvs, gomega.Equal(float64(0)))
+}
+
+func ExpectClusterQueueInfoMetricCleaned(cqName, parentCohort, rootCohort string, customLabels ...string) {
+	ginkgo.GinkgoHelper()
+	lvs := append([]string{cqName, parentCohort, rootCohort, roletracker.RoleStandalone}, customLabels...)
+	expectGaugeMetric(metrics.ClusterQueueInfo, lvs, gomega.Equal(float64(0)))
+}
```
mbobrovskyi: nit: do we need this?

mszadkow: can be dropped, yes

pkg/cache/scheduler/cache.go
mszadkow: ahh right, I see it

mbobrovskyi (commented):
/lgtm

Thank you!

pkg/cache/scheduler/cache.go
```diff
@@ -530,7 +532,23 @@ func (c *Cache) AddOrUpdateCohort(apiCohort *kueue.Cohort) error {
...
+	if oldParent != nil && oldParent != cohort.Parent() && c.hm.Cohort(oldParent.Name) == nil {
+		metrics.ClearCohortInfo(oldParent.Name)
+	}
+
+	// cover implicit cohort creation with potential children transfer
+	reportRoot := cohort
+	if parent := cohort.Parent(); parent != nil {
+		reportRoot = parent
+	}
+	c.reportInfoMetricsSubtree(reportRoot)
```
mimowo: Could we use the PathToRoot here?

mszadkow: This covers a situation when the implicit parent cohort just got created, I think we just need to trigger metric for one level above like mentioned here: https://github.com/kubernetes-sigs/kueue/pull/10004#discussion_r2974177932
However if we are interested in the some optimisation, there could be additional version of `reportInfoMetricsSubtree` like `reportInfoMetricsSubtreeResolved` where we pass the root.
Also I think the naming could be actually better, this is the `root` of this subtree that will be recalculated, the real root of the whole tree is retrieved later.
So maybe `subtreeRoot`, wdyt?

pkg/cache/scheduler/cache.go
```diff
@@ -530,7 +532,23 @@ func (c *Cache) AddOrUpdateCohort(apiCohort *kueue.Cohort) error {
...
+	return nil
```
mimowo, line 617: I 'm actually concerned about the changes as they seem to go up and down the scheduling tree from within cache which should be very efficient. It is hard to infer from the code, so let me ask, what is the performance of the code, assymptotic? 

Also, why this isn't done from the cohort reconciler as other cohort metrics you did in the previous PRs, it seems much better fit for the problem than cache.

mimowo: /hold
To make sure this change is safe from the performance PoV we need:
1. performance analysis of the operations when called from within cache
2. convincing argument why not doing from within cohort reconciler how the previous metrics where done

mimowo: I mean specifically these two PRs:
- https://github.com/kubernetes-sigs/kueue/pull/9833
- https://github.com/kubernetes-sigs/kueue/pull/9132
if we could follow the same pattern which is already exercised

mszadkow:
@mimowo I think it's ready to be considered again
@mbobrovskyi ptal also

It's different approach and it has extra commit from: https://github.com/kubernetes-sigs/kueue/pull/10279

mimowo:
cc @gabesaba ptal

gabesaba (commented):
I wonder if we can encapsulate these metrics in the `hierarchy` package, both as it has the best picture of what is going on, and so details don't leak around the code base.

Do you see any downside to this approach, besides the fact that we currently have two instantiations - one for `cache/scheduler`, and another for `cache/queue`.

mimowo:
> I wonder if we can encapsulate these metrics in the hierarchy package, both as it has the best picture of what is going on, and so details don't leak around the code base.

Yeah I think for this specific metric is might be sensible, but I was also thinking about this in the context of other metrics - for the guage for admitted workloads in a cohort subtree: https://github.com/kubernetes-sigs/kueue/pull/10013. 

In that case the hierarchy package does not have the full context so it does not seem like a univesal approach. 

Also, since the package is reused by both "scheduler cache" and "queue cache" we may need to be careful to avoid updating the metrics twice for a single change, this is harmless by itself, but might be wasted work. 

I'm not saying this is bad idea, but just something to consider.

mszadkow:
> I wonder if we can encapsulate these metrics in the `hierarchy` package, both as it has the best picture of what is going on, and so details don't leak around the code base.
> 
> Do you see any downside to this approach, besides the fact that we currently have two instantiations - one for `cache/scheduler`, and another for `cache/queue`.



> I wonder if we can encapsulate these metrics in the `hierarchy` package, both as it has the best picture of what is going on, and so details don't leak around the code base.
> 
> Do you see any downside to this approach, besides the fact that we currently have two instantiations - one for `cache/scheduler`, and another for `cache/queue`.

We can definitively consider moving more relevant parts to hierarchy package, like parent/root derivation and subtree traversal.
However let's keep metric reporting in scheduler cache to separate concerns.

mszadkow:
@gabesaba ptal

pkg/cache/scheduler/cache.go
```diff
@@ -1005,12 +1024,38 @@ func (c *Cache) MatchingClusterQueues(nsLabels map[string]string) sets.Set[kueue
...
+	defer c.RUnlock()
+	cq := c.hm.ClusterQueue(cqName)
+	if cq == nil || !cq.HasParent() {
+		return "", ""
+	}
+	parent := cq.Parent()
+	parentCohort = parent.GetName()
+	rootCohort = parent.getRootUnsafe().Name
+
+	return parentCohort, rootCohort
```
gabesaba: can this be a query on `HierarchyManager` to return `InfoPoint`?

pkg/cache/scheduler/cache.go
```diff
@@ -524,20 +529,33 @@ func (c *Cache) DeleteClusterQueue(cq *kueue.ClusterQueue) {
 	}
 }
 
-func (c *Cache) AddOrUpdateCohort(apiCohort *kueue.Cohort) error {
+// AddOrUpdateCohort adds a cohort to the cache or updates its hierarchy and properties if it already exists.
+// Returns a boolean indicating if the cohort hierarchy has changed, and an error if any issue occurs during the update.
+func (c *Cache) AddOrUpdateCohort(apiCohort *kueue.Cohort) (bool, error) {
```
gabesaba: is it possible to do this without signature update, as it looks like we only have one site using this change, and the rest setting _

gabesaba (commented):
Can we simplify this by plugging in these changes at call sites to `updateCohortTreeResourcesIfNoCycle`? That is a well tested path which covers all Cohort mutation cases. This should cover all the cases which results in mutation of these info metrics as well.

Something like this:

```go
func updateCohortResourcesAndMetrics() {
  if cycle {
    return
  } 
  updateCohortTreeResourcesIfNoCycle()
  updateInfoMetrics()
}

func updateInfoMetrics() {
  // 1) delete orphaned implicit cohort metrics
  // 2) update metric recursively for entire cohort tree
}
```

pkg/cache/scheduler/cache.go
mszadkow: yes, especially if we go with plugging into updateCohort

mszadkow:
@gabesaba I did centralize cohort-tree resource and info metrics updates - ptal

pkg/cache/scheduler/cohort_metrics.go
```diff
@@ -31,7 +31,7 @@ import (
 	"sigs.k8s.io/kueue/pkg/workload"
 )
 
-type cohortMetricPoint struct {
+type cohortResourceMetricPoint struct {
```
mszadkow: I will revert this name change

mszadkow:
@gabesaba @vladikkuzn @mbobrovskyi ptal

pkg/cache/scheduler/cache.go
```diff
@@ -467,6 +467,13 @@ func (c *Cache) AddClusterQueue(ctx context.Context, cq *kueue.ClusterQueue) err
 		}
 	}
 
+	var parentCohort, rootCohort kueue.CohortReference
```
mimowo: Is the new code in this file fully covered by unit tests?

pkg/controller/core/cohort_controller.go
```diff
@@ -198,13 +200,14 @@ func (r *CohortReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	}
 
 	log.V(2).Info("Cohort is being created or updated", "resources", cohort.Spec.ResourceGroups)
-	if err := r.cache.AddOrUpdateCohort(&cohort); err != nil {
+	err := r.cache.AddOrUpdateCohort(&cohort)
```
mimowo: why is this changed? this seems to be just reformatting the code, same below

vladikkuzn: Could this leave the system in an inconsistent state?
If there's a cycle error, the cache won't update correctly
But qManager.AddOrUpdateCohort will still be called, potentially with stale/invalid data
The original error is lost, so the controller won't retry for cache failures

```
 log.V(2).Info("Cohort is being created or updated", "resources", cohort.Spec.ResourceGroups)
-err := r.cache.AddOrUpdateCohort(&cohort)
-if err != nil {
+if err := r.cache.AddOrUpdateCohort(&cohort); err != nil {
     log.V(2).Error(err, "Error adding or updating cohort in the cache")
+	return ctrl.Result{}, err
 }
```

mszadkow: I will revert

vladikkuzn: > I will revert

The importand line here is `return ctrl.Result{}, err`, which is not present in the code (existing one)

mszadkow: Ahh got you, yes, that is important to cover

pkg/cache/scheduler/cache.go
mszadkow: updated, should be fully covered now

mszadkow:
@vladikkuzn you can have another look

pkg/cache/scheduler/cache.go
```diff
@@ -467,6 +467,13 @@ func (c *Cache) AddClusterQueue(ctx context.Context, cq *kueue.ClusterQueue) err
...
+	if cqImpl.HasParent() {
+		parentCohort = cqImpl.Parent().GetName()
+		rootCohort = cqImpl.Parent().getRootUnsafe().Name
```
mimowo: What happens in case there is an infinite cycle, do we have a test for that? Seems like there is a potential for a severe bug calling `getRootUnsafe`.

pkg/cache/scheduler/cache.go
```diff
@@ -467,6 +467,13 @@ func (c *Cache) AddClusterQueue(ctx context.Context, cq *kueue.ClusterQueue) err
...
+	}
```
mimowo: Could this code be moved under `c.recordCQInfo`?

pkg/cache/scheduler/cache.go
```diff
@@ -492,6 +499,18 @@ func (c *Cache) UpdateClusterQueue(log logr.Logger, cq *kueue.ClusterQueue) erro
 		}
 		qImpl.resetFlavorsAndResources(cqImpl.resourceNode.Usage, cqImpl.AdmittedUsage)
 	}
+
+	var parentCohort, rootCohort kueue.CohortReference
+	if cqImpl.HasParent() {
+		parentCohort = cqImpl.Parent().GetName()
+		rootCohort = cqImpl.Parent().getRootUnsafe().Name
+	}
```
mimowo, line 515: This looks like a duplication of https://github.com/kubernetes-sigs/kueue/pull/10004/changes#diff-9d50921591b2d9622bc605a02ee4015136a176c9b9f8886ca5b10fd58d0c06dfR470-R474 with the analogous concerns:
1. inifinite loop risk
2. could we move the code under `recordCQInfoIfChanged`
3. could we deduplicate as this looks like copy-paste?

pkg/cache/scheduler/cache.go
```diff
@@ -492,6 +499,18 @@ func (c *Cache) UpdateClusterQueue(log logr.Logger, cq *kueue.ClusterQueue) erro
...
+	c.recordCQInfoIfChanged(cqImpl, parentCohort, rootCohort)
+
+	// If old parent was implicit and got removed, clear its info metric.
+	if oldParent != nil && c.hm.Cohort(oldParent.Name) == nil {
+		metrics.ClearCohortInfo(oldParent.Name)
+	}
```
mimowo: IIRC we already have places in code responsible for clearing the deletion of the implicit parent that @mbobrovskyi worked on, could we move the code there to avoid scattering the code across the codebase? Ptal with @mbobrovskyi

pkg/cache/scheduler/cache.go
```diff
@@ -514,14 +533,25 @@ func (c *Cache) DeleteClusterQueue(cq *kueue.ClusterQueue) {
 	}
 
 	parent := curCq.Parent()
+	var parentName kueue.CohortReference
+	if parent != nil {
+		parentName = parent.Name
+	}
```
mimowo: Move it where actually it is used. Also I don't understand why computing `parentName` here, and later using `if parentName != "" {`. Why not just use `if parent != nil {` below and defer computing `parantName when` needed

pkg/cache/scheduler/cache.go
```diff
@@ -514,14 +533,25 @@ func (c *Cache) DeleteClusterQueue(cq *kueue.ClusterQueue) {
...
+
+	if parentName != "" {
+		if updatedParent := c.hm.Cohort(parentName); updatedParent != nil {
+			c.updateCohortTreeAndInfoMetricsIfNoCycle(updatedParent)
+			c.handleParentUpdate(updatedParent)
+		} else {
+			// Parent cohort could be implicit and removed when this CQ was deleted.
+			// Ensure no stale info/admitted metrics remain for removed cohorts.
+			metrics.ClearCohortInfo(parentName)
+			c.handleParentUpdate(parent)
```
mimowo: Isn't `parent == nil` here?

pkg/cache/scheduler/cache.go
```diff
@@ -1034,8 +1075,19 @@ func (c *Cache) MatchingClusterQueues(nsLabels map[string]string) sets.Set[kueue
...
+	metrics.ClusterQueueInfo.Reset()
+	metrics.CohortInfo.Reset()
+
 	for _, cq := range c.hm.ClusterQueues() {
 		metrics.ReportClusterQueueStatus(cq.Name, cq.Status, cq.customMetricLabelValues, c.roleTracker)
+		var parentCohort, rootCohort kueue.CohortReference
+		if cq.HasParent() {
+			parentCohort = cq.Parent().GetName()
+			rootCohort = cq.Parent().getRootUnsafe().Name
+		}
```
mimowo: This looks again like copy-paste of https://github.com/kubernetes-sigs/kueue/pull/10004/changes#diff-9d50921591b2d9622bc605a02ee4015136a176c9b9f8886ca5b10fd58d0c06dfR503-R507
why not deduplicate?

pkg/cache/scheduler/cache.go
```diff
@@ -1047,14 +1099,16 @@ func (c *Cache) ResyncGaugeMetrics() {
...
+			metrics.ReportCohortInfo(cohort.Name, parentName, cohort.getRootUnsafe().Name, c.customLabels.CohortGet(cohort.Name), c.roleTracker)
+		}
+		if c.fairSharingEnabled {
 			drs := dominantResourceShare(cohort, nil)
-			var customLabelValues []string
-			if features.Enabled(features.CustomMetricLabels) {
-				customLabelValues = c.customLabels.CohortGet(cohort.Name)
-			}
-			metrics.ReportCohortWeightedShare(cohort.Name, drs.PreciseWeightedShare(), customLabelValues, c.roleTracker)
+			metrics.ReportCohortWeightedShare(cohort.Name, drs.PreciseWeightedShare(), c.customLabels.CohortGet(cohort.Name), c.roleTracker)
```
mimowo: What has happened to checking the `CustomMetricLabels` feature gate?

pkg/cache/scheduler/cache.go
mszadkow: Test added

pkg/cache/scheduler/cache.go
mszadkow: Not really, `recordCQInfoIfChanged` calls `recordCQInfo`, passing the already computed values through
But I can provide a helper that extracts parent and root safely before calling `recordCQInfo`, that should be ok

pkg/cache/scheduler/cache.go
mszadkow: yes, it will be handled in the helper

pkg/cache/scheduler/cache.go
mszadkow: no, it's actually correct how it is, but I agree this could be simplified and maybe more readable in turn

mimowo: I see, yeah, let's try to make it more readable, I could figure out if `parent == nil` despite a couple of minutes staring at the code. Do we have a unit test which shows we step into this block with `parent != nil`?

pkg/cache/scheduler/cache.go
mszadkow: brought back

pkg/cache/scheduler/cache.go
mszadkow: I brought back `parent != nil` condition for better readibility

pkg/cache/scheduler/resource_node.go
```diff
@@ -242,3 +242,40 @@ func IsWithinNominalInResources(node flatResourceNode, frs sets.Set[resources.Fl
...
+	}
+}
+
+// updateCohortTreeAndInfoMetricsIfNoCycle updates subtree resources and metrics
+// only if there's no cycle.
+func (c *Cache) updateCohortTreeAndInfoMetricsIfNoCycle(cohort *cohort) {
+	if !hierarchy.HasCycle(cohort) {
+		root := cohort.getRootUnsafe()
+		c.updateCohortResourceAndInfoMetrics(root, root)
+	}
```
mimowo: These helpers seem to only be doing metrics, could we move them to the dedicated file for metrics? I think we have the cohort_metrics.go for that, wdyt?

mszadkow: yes, it's good place to move it

pkg/cache/scheduler/cache.go
mszadkow: I made it incorporated into `handleParentUpdate` as a best fit

pkg/cache/scheduler/clusterqueue.go
```diff
@@ -104,12 +104,26 @@ type clusterQueue struct {
 	customMetricLabelValues []string
 
 	lqMetrics *metrics.LocalQueueMetricsConfig
+
+	// Cached hierarchy labels from last metric recording.
+	lastRecordedParent            kueue.CohortReference
+	lastRecordedRoot              kueue.CohortReference
+	lastRecordedCustomLabelValues []string
```
mimowo: Why do we have the fields? Are they used to clear the guage metric when something is changed? If so then why not to use the same mechanism as for CustomLabels? It was recently generalized here: https://github.com/kubernetes-sigs/kueue/pull/10552
It would be nice if we could reuse the generic mechanism rather than cooking a new solution each time, but let me know if this cannot be used in this case for some reason, or is hard.

mimowo: cc @MatteoFari ptal as you are probably the most familiar currently with the mechanism

mszadkow: I understand the concern, I am for reusability too.
For CQ I admit `customMetricLabelValues` can be used and it should be doable fairly easy.
For Cohort we don't have similar field, so it would have to be introduced first.
Then this part could be used instead of `lastRecordedCustomLabelValues`.

For `lastRecordedParent` and `lastRecordedRoot`, they have no equivalent in the existing mechanism. Parent/root changes aren't captured by the CQ's own labels.
They reflect external structural changes in the cohort tree.
These two fields are genuinely new state.

MatteoFari: I agree `parent_cohort` and `root_cohort` are not covered by the custom-label store.

But I don’t think we need `lastRecordedParent` / `lastRecordedRoot` for correctness. `ClearClusterQueueInfo` deletes by `cluster_queue`, and `ClearCohortInfo` deletes by `cohort`, so stale series can be removed without knowing the old parent/root values.

That matches #10552: clear by stable identity, then report again from current cache state. For Cohort reparenting, the affected subtree needs to be cleared/resynced because descendants can get a new `root_cohort`.

If `lastRecorded*` is only to skip extra delete/report calls while walking the subtree, then it is an optimization. I’d prefer avoiding that extra cached metric state unless we really need it.

For custom labels, `lastRecordedCustomLabelValues` can also be avoided: CQ already has `customMetricLabelValues`, and Cohort can use `customLabels.CohortGet(...)`.

mimowo: I haven't yet read in full detail, but this is certainly looking suspicious that we introduce the fields despite already having the mechanism generalized. If the issue remains non-resolvable for a while I will try to look closer towards the EOW, but I'm currently under too heavy review load to investigate this very deeply so help from other contributors like @MatteoFari is greatly appreciated. Thank you 👍

mszadkow: `lastRecordedParent` / `lastRecordedRoot` is used for optimization to not re-record existing metrics.
Also we don't have generalized mechanism for that, we store customLabels, I don't think we can reuse the same thing for parent/root.

@mimowo IIRC potential performance degradation was a big concern all throughout this PR.

I agree this fields are not required for correctness, but for optimization only.
If introducing them pose a problem in reviewers' opinion we can live without them.

mimowo: Could you summarize what is the performance complexity with and without them?  So that we can make more informed decision. Saying just "it is optimization" is unclear to me how much benefit we get. I'm trying to balance between premature optimization and performance hits. 

The cohort metrics are too complex it seems over the last months to stay ungated anyway. So we need a FG to guard the code.

mszadkow: Also as a mid-ground solution I can drop it for now and introduce as optimization in the follow-up, whenever the cohort metrics will be kept behind the feature gate

mimowo: We n Ed the FG so maybe let's start from this PR as prep, say Metrics for Cohorts to be beta, then I will feel more confident accepting this PR without the optimization.

I'm still surprised it is not possible to optimize the code without the cars, but I really haven't got time to think about it.. Maybe @mbobrovskyi could also help here

pkg/controller/core/cohort_controller.go
reruno, line 200: Should we move this to func (c *Cache) AddOrUpdateCohort(apiCohort *kueue.Cohort) error {
 
to have unified metrics collection

mszadkow: This cleanup can be done in the followup, as for now `labelsUpdated` is used for recording or resyncing other cohort metrics

pkg/cache/scheduler/cohort_metrics.go
```diff
@@ -195,3 +196,62 @@ func (c *Cache) ReportCohortSubtreeAdmittedWorkload(log logr.Logger, wl *kueue.W
...
+	if !features.Enabled(features.MetricsForCohorts) {
+		return
+	}
+	customLabels := c.customLabels.CQGet(cq.Name)
+	metrics.ClearClusterQueueInfo(cq.Name)
+	metrics.ReportClusterQueueInfo(cq.Name, parentCohort, rootCohort, customLabels, c.roleTracker)
+}
+
+// updateCohortResourceAndInfoMetrics updates subtree resources and metrics in one pass.
+func (c *Cache) updateCohortResourceAndInfoMetrics(cohort *cohort, root *cohort) {
```
mimowo, line 225: This feels like a potentially heavy function.  It also has non trivial logic.

Let's skip it when `MetricsForCohorts` is disabled so that we can have some kind of bailout mechanism.

Please follow up with that.

mimowo:
/lgtm
/approve
Merging as the new metrics lgtm, but let's make sure the biggest new function is isolated from the pre-existing code: https://github.com/kubernetes-sigs/kueue/pull/10004#discussion_r3306754223

Please also address https://github.com/kubernetes-sigs/kueue/pull/10004#discussion_r3297626270 in a follow up

pkg/cache/scheduler/cohort_metrics.go
mszadkow: Ack

PR #8862, opened 2026-01-28
LeaderWorkerSet: Fix bug that doesn't allow to delete Pod after LeaderWorkerSet delete

mbobrovskyi:
#### What type of PR is this?
/kind bug

#### What this PR does / why we need it:
LeaderWorkerSet: Fix bug that doesn't allow to delete Pod after LeaderWorkerSet delete

#### Which issue(s) this PR fixes:

Fixes #8733
Fixes #8752

#### Does this PR introduce a user-facing change?

```release-note
LeaderWorkerSet: Fixed bug that doesn't allow to delete Pod after LeaderWorkerSet delete
```

PR #10009, opened 2026-03-19
Fix webhook denial when ancestor owner is not found during GC teardown

iaalm:
#### What type of PR is this?

/kind bug
/area integrations

When the Garbage Collector PATCHes a child object to remove the foregroundDeletion finalizer, Kueue's mutating webhooks invoke WorkloadShouldBeSuspended → FindAncestorJobManagedByKueue to walk the ownerReference chain. If any owner in that chain has already been deleted (e.g. via background/helm uninstall while a foreground deletion is in progress on a child), c.Get returns NotFound and the function returns ErrWorkloadOwnerNotFound. This error propagates up and causes the webhook to deny the PATCH, permanently blocking the GC from removing the finalizer and leaving objects stuck in Terminating state.

This is a generalisation of the problem partially addressed in #8862. That fix added a managedByAnotherFramework() guard based on the kueue.x-k8s.io/pod-suspending-parent annotation, but objects created directly by controllers (e.g. StatefulSets created by the LWS controller without going through Kueue's webhook path) never receive this annotation, so the guard does not help them.

Fix: treat ErrWorkloadOwnerNotFound as "no Kueue-managed ancestor" (return false, nil) inside WorkloadShouldBeSuspended. A missing owner during a webhook admission call means the ownership chain is being garbage collected — there is nothing to suspend, and the operation should be allowed.

The fix is applied at two call sites:

- WorkloadShouldBeSuspended in pkg/controller/jobframework/defaults.go — covers all webhook Default() / ValidateUpdate() paths, plus the StatefulSet and LeaderWorkerSet reconcilers' use of this function.
- The pod webhook's direct call to FindAncestorJobManagedByKueue in pkg/controller/jobs/pod/pod_webhook.go.

The main reconciler's direct call to FindAncestorJobManagedByKueue in reconciler.go is intentionally left unchanged: requeuing on ErrWorkloadOwnerNotFound there is correct, since during normal operation a parent may transiently be absent due to creation ordering.

Fixes # n/a

The scenario that triggers this bug requires a mix of foreground and background deletion in the same ownership chain (e.g. LWS → Parent STS → Leader Pod → Child STS → Worker Pods where a pod restart causes foreground deletion of the leader pod cascading foregroundDeletion finalizers down to the child STS, while helm uninstall concurrently removes the parent STS via background deletion). With a purely foreground or purely background chain the deadlock does not occur.

Reproducible evidence from a live cluster:

```
$ kubectl patch sts <name> -n <ns> --type=merge -p '{"metadata":{"finalizers":null}}' --dry-run=server
Error: admission webhook "mstatefulset.kb.io" denied the request: workload owner not found
```

no

```release-note
None
```

### DIFF
--- a/pkg/controller/jobframework/defaults.go
+++ b/pkg/controller/jobframework/defaults.go
@@ -35,6 +35,10 @@ import (
 
 func ApplyDefaultForSuspend(ctx context.Context, job GenericJob, k8sClient client.Client,
 	manageJobsWithoutQueueName bool, managedJobsNamespaceSelector labels.Selector) error {
+	// Do not default suspend an object that is already being deleted (e.g. during GC teardown).
+	if job.Object().GetDeletionTimestamp() != nil {
+		return nil
+	}
 	suspend, err := WorkloadShouldBeSuspended(ctx, job.Object(), k8sClient, manageJobsWithoutQueueName, managedJobsNamespaceSelector)
 	if err != nil {
 		return err
--- a/pkg/controller/jobs/pod/pod_webhook.go
+++ b/pkg/controller/jobs/pod/pod_webhook.go
@@ -148,6 +148,12 @@ func (w *PodWebhook) Default(ctx context.Context, obj *corev1.Pod) error {
 			return err
 		}
 
+		// During GC teardown the pod may be terminating with its parent already gone;
+		// skip further mutation to avoid adding finalizers to a deleting pod.
+		if pod.pod.DeletionTimestamp != nil {
+			return nil
+		}
+
 		// Local queue defaulting
 		if jobframework.QueueNameForObject(pod.Object()) == "" &&
 			w.queues.DefaultLocalQueueExist(pod.pod.GetNamespace()) {
--- a/pkg/controller/jobs/pod/pod_webhook_test.go
+++ b/pkg/controller/jobs/pod/pod_webhook_test.go
@@ -17,7 +17,9 @@ limitations under the License.
 package pod
 
 import (
+	"errors"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
@@ -209,6 +211,33 @@ func TestDefault(t *testing.T) {
 				Queue("test-queue").
 				OwnerReference("parent-job", batchv1.SchemeGroupVersion.WithKind("Job")).
 				Obj(),
+		},
+		"pod with ownerReference to a deleted known parent while terminating (GC teardown)": {
+			initObjects:       []client.Object{defaultNamespace},
+			podSelector:       &metav1.LabelSelector{},
+			namespaceSelector: defaultNamespaceSelector,
+			pod: testingpod.MakePod("test-pod", defaultNamespace.Name).
+				Queue("test-queue").
+				OwnerReference("deleted-parent-job", batchv1.SchemeGroupVersion.WithKind("Job")).
+				DeletionTimestamp(time.Now()).
+				Obj(),
+			enableIntegrations: []string{"batch/job"},
+			want: testingpod.MakePod("test-pod", defaultNamespace.Name).
+				Queue("test-queue").
+				OwnerReference("deleted-parent-job", batchv1.SchemeGroupVersion.WithKind("Job")).
+				DeletionTimestamp(time.Now()).
+				Obj(),
+		},
+		"pod with ownerReference to a missing parent while not terminating (cache lag, should error)": {
+			initObjects:       []client.Object{defaultNamespace},
+			podSelector:       &metav1.LabelSelector{},
+			namespaceSelector: defaultNamespaceSelector,
+			pod: testingpod.MakePod("test-pod", defaultNamespace.Name).
+				Queue("test-queue").
+				OwnerReference("deleted-parent-job", batchv1.SchemeGroupVersion.WithKind("Job")).
+				Obj(),
+			enableIntegrations: []string{"batch/job"},
+			wantErr:            errors.New("owner not found"),
 		},
 		"pod with owner managed by kueue (RayCluster)": {
 			initObjects: []client.Object{
@@ -583,8 +612,12 @@ func TestDefault(t *testing.T) {
 				podSelector:                  tc.podSelector,
 			}
 
-			if err := w.Default(ctx, tc.pod); err != nil {
-				t.Errorf("failed to set defaults for v1/pod: %s", err)
+			gotErr := w.Default(ctx, tc.pod)
+			if (gotErr != nil) != (tc.wantErr != nil) {
+				t.Errorf("Default() error = %v, wantErr %v", gotErr, tc.wantErr)
+			}
+			if gotErr != nil {
+				return
 			}
 
 			if diff := cmp.Diff(tc.want, tc.pod); len(diff) != 0 {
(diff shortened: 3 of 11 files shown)


mimowo:
Actually, I'm thinking about it. What if you use `managedByAnotherFramework` and set on such Pods `kueue.x-k8s.io/suspended-by-parent: custom-framework`. I think then Kueue should skip processing them, and skip the ownership traversal. Have you tried that? IIRC one of the teams we collab with does something like this,

pkg/controller/jobframework/defaults.go
```diff
@@ -48,6 +48,11 @@ func ApplyDefaultForSuspend(ctx context.Context, job GenericJob, k8sClient clien
 // WorkloadShouldBeSuspended determines whether jobObj should be default suspended on creation
 func WorkloadShouldBeSuspended(ctx context.Context, jobObj client.Object, k8sClient client.Client,
 	manageJobsWithoutQueueName bool, managedJobsNamespaceSelector labels.Selector) (bool, error) {
+	// Do not default suspend an object that is already being deleted (e.g. during GC teardown).
```
sohankunkerkar: This early exit affects more than webhooks. `WorkloadShouldBeSuspended` is also called from the LWS and StatefulSet reconcilers. A terminating managed object would skip the ancestor check in those paths too. Might be better to handle this at the webhook call sites instead of the shared function.

pkg/controller/jobs/pod/pod_webhook_test.go
```diff
@@ -210,6 +212,33 @@ func TestDefault(t *testing.T) {
 				OwnerReference("parent-job", batchv1.SchemeGroupVersion.WithKind("Job")).
 				Obj(),
 		},
+		"pod with ownerReference to a deleted known parent while terminating (GC teardown)": {
```
sohankunkerkar, line 215: The two new test cases have inconsistent indentation relative to the surrounding map entries.

pkg/controller/jobframework/defaults.go
iaalm: Good point, thanks! Moved the DeletionTimestamp guard out of WorkloadShouldBeSuspended and into ApplyDefaultForSuspend (covers all base-webhook callers) and each direct webhook call site (LWS, StatefulSet, TrainJob, Deployment). Note that FindAncestorJobManagedByKueue already handles the not-found-parent case for terminating objects gracefully, so the shared function itself doesn't need the early exit.

pkg/controller/jobs/pod/pod_webhook_test.go
iaalm: Should be default.go? updated

pkg/controller/jobframework/defaults.go
```diff
@@ -35,6 +35,10 @@ import (
 
 func ApplyDefaultForSuspend(ctx context.Context, job GenericJob, k8sClient client.Client,
 	manageJobsWithoutQueueName bool, managedJobsNamespaceSelector labels.Selector) error {
+	// Do not default suspend an object that is already being deleted (e.g. during GC teardown).
+	if !job.Object().GetDeletionTimestamp().IsZero() {
```
mwielgus: Unify DeletionTimestamp checks. In some places you use !job.Object().GetDeletionTimestamp().IsZero() and in others obj.GetDeletionTimestamp() != nil.

iaalm: Thanks, fixed.

iaalm:
> @iaalm is it feasible to have an integration or e2e test which fails without the fix? At least flakes

Hi @mimowo , added a new test in statefulset_webhook_test.go

test/integration/singlecluster/webhook/jobs/statefulset_webhook_test.go
```diff
@@ -93,5 +94,69 @@ var _ = ginkgo.Describe("StatefulSet Webhook", func() {
 				}, util.Timeout, util.Interval).Should(gomega.Succeed())
 			})
 		})
+
+		// Regression test for GC teardown deadlock:
+		// When foreground and background deletion are mixed in the same ownership chain,
+		// a child STS can get stuck in Terminating because the webhook denied the GC's
+		// PATCH to remove the foregroundDeletion finalizer (parent was already gone).
+		ginkgo.When("a child StatefulSet is terminating with its parent already deleted", func() {
```
mimowo, line 102: Add ginkgo.By steps for the consecutive steps so that it is easier to correlate the logs from the test runner and the kueue logs when debugging - based on the timestamps of the build test logs recorded for each ginkgo.By

test/integration/singlecluster/webhook/jobs/statefulset_webhook_test.go
```diff
@@ -93,5 +94,69 @@ var _ = ginkgo.Describe("StatefulSet Webhook", func() {
...
+			ginkgo.It("Should allow removing foregroundDeletion finalizer", func() {
```
mimowo, line 103: Could you point out which assert specifically flakes in this test and how often?

pkg/controller/jobs/trainjob/trainjob_webhook.go
```diff
@@ -76,29 +75,22 @@ func (w *TrainJobWebhook) Default(ctx context.Context, obj *kftrainerapi.TrainJo
 
 	jobframework.ApplyDefaultLocalQueue(trainJob.Object(), w.queues.DefaultLocalQueueExist)
 	jobframework.ApplyDefaultForManagedBy(trainJob, w.queues, w.cache, log)
+	if obj.GetDeletionTimestamp() != nil {
```
mimowo, line 78: Should we check that only when no "queue-name" after the LocalQueue defauting, to make the check more selective? 

Also, should we commonize by moving this check as the first step inside WorkloadShouldBeSuspended? Currently there is a signifficant duplication and only one code path is covered

pkg/controller/jobs/pod/pod_webhook.go
```diff
@@ -148,6 +148,12 @@ func (w *PodWebhook) Default(ctx context.Context, obj *corev1.Pod) error {
 			return err
 		}
 
+		// During GC teardown the pod may be terminating with its parent already gone;
+		// skip further mutation to avoid adding finalizers to a deleting pod.
+		if pod.pod.DeletionTimestamp != nil {
+			return nil
+		}
```
mimowo, line 155: Maybe this is corner case, but in other cases this check is after LocalQueueDefaulting. THis is related to other comment to try to commonize the approach.

mimowo:
I would like to also consider a feature gate for it so that we can mitigate environments if anyone is affected, say `SkipAncestorCheckForDeletedWorkloads`. This could go directly as Beta (enabled by default), but would give use a bailout tool in case of issues, wdyt?

tomsen02:
Hi @iaalm — the bug here still reproduces on current `main` (your integration test fails unchanged with `admission webhook "mstatefulset.kb.io" denied the request: StatefulSet.apps "parent-sts" not found`). Since this has been idle for a while, I took the liberty of picking it up and finishing it in #13857: rebased onto main, implemented the `SkipAncestorCheckForDeletedWorkloads` gate suggested in the review (Beta, default on, with a gate-off test restoring the previous behavior), and kept your commit authorship. Happy to hand it back if you want to resume — otherwise hope this gets your fix over the line. cc @mimowo

(AI tooling assisted the port; details in the new PR.)

mimowo:
/close
As this is now replaced by https://github.com/kubernetes-sigs/kueue/pull/13857

PR #10013, opened 2026-03-19
Introduce kueue_cohort_subtree_admitted_active_workloads metric.

mbobrovskyi:
#### What type of PR is this?
/kind feature

Introduce kueue_cohort_subtree_admitted_active_workloads metric.

Part of #7539

```release-note
Observability: Introduce kueue_cohort_subtree_admitted_active_workloads metric
```

### DIFF
--- a/pkg/cache/scheduler/clusterqueue.go
+++ b/pkg/cache/scheduler/clusterqueue.go
@@ -495,7 +495,7 @@ func (c *clusterQueue) deleteWorkload(log logr.Logger, wlKey workload.Reference)
 
 func (c *clusterQueue) reportActiveWorkloads() {
 	for ancestor := range c.Parent().PathSelfToRoot() {
-		metrics.ReportCohortSubtreeAdmittedActiveWorkloads(ancestor.Name, c.admittedWorkloadsCount, c.customMetricLabelValues, c.roleTracker)
+		metrics.ReportCohortSubtreeAdmittedActiveWorkloads(ancestor.Name, ancestor.admittedWorkloadsCount, c.customMetricLabelValues, c.roleTracker)
 	}
 	metrics.ReportAdmittedActiveWorkloads(c.Name, c.admittedWorkloadsCount, c.customMetricLabelValues, c.roleTracker)
 	metrics.ReportReservingActiveWorkloads(c.Name, len(c.Workloads), c.customMetricLabelValues, c.roleTracker)
--- a/test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
+++ b/test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
@@ -531,6 +531,10 @@ var _ = ginkgo.Describe("Cohorts", func() {
 				util.ExpectCohortSubtreeAdmittedWorkloadsTotalMetric("root", "", numWorkloadsForCQ2)
 				util.ExpectCohortSubtreeAdmittedWorkloadsTotalMetric("ch1", "", 0)
 				util.ExpectCohortSubtreeAdmittedWorkloadsTotalMetric("ch2", "", numWorkloadsForCQ2)
+
+				util.ExpectCohortSubtreeAdmittedActiveWorkloadsGaugeMetric("root", numWorkloadsForCQ2)
+				util.ExpectCohortSubtreeAdmittedActiveWorkloadsGaugeMetric("ch1", 0)
+				util.ExpectCohortSubtreeAdmittedActiveWorkloadsGaugeMetric("ch2", numWorkloadsForCQ2)
 			})
 
 			ginkgo.By("Marking the checks as passed", func() {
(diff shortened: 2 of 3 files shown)


site/content/en/docs/reference/metrics.md
```diff
@@ -82,6 +82,7 @@ The following metrics are available only if `LocalQueueMetrics` feature gate is
 <!-- BEGIN GENERATED TABLE: cohort -->
 | Metric name | Type | Description | Labels |
 | --- | --- | --- | --- |
+| `kueue_cohort_subtree_admitted_active_workloads` | Gauge | The number of admitted Workloads that are active (unsuspended and not finished), per cohort's subtree | `cohort`: the name of the Cohort<br> `replica_role`: one of `leader`, `follower`, or `standalone` |
```
mimowo: Do we actually check they are unsuspended? I don't see that in code, maybe better to say "The number of admitted Workloads that are active (ie. not finished), per cohort's subtree"

mbobrovskyi: Hmm. I think you right. Should we also update other places? WDYT?

mimowo: We could yes, but in a dedicated PR so that we can cherrypick easily, I would not mix the changes to pre-existing metrics in this PR

mbobrovskyi: Yeah, that what I mean.

pkg/cache/scheduler/cache.go
```diff
@@ -536,6 +536,7 @@ func (c *Cache) AddOrUpdateCohort(apiCohort *kueue.Cohort) error {
 func (c *Cache) DeleteCohort(cohortName kueue.CohortReference) {
 	c.Lock()
 	defer c.Unlock()
+	c.hm.Cohort(cohortName).removeAdmittedWorkloadUsage()
```
mimowo: This seems conceptually part of `c.hm.DeleteCohort(cohortName)` rather than a separate function on its own

pkg/cache/scheduler/clusterqueue.go
```diff
@@ -548,6 +551,7 @@ func (c *clusterQueue) updateWorkloadUsage(log logr.Logger, wi *workload.Info, o
 	c.updateWorkloadTASUsage(log, wi, op)
 	if admitted {
 		updateFlavorUsage(frUsage, c.AdmittedUsage, op)
+		c.Parent().applyAdmittedWorkloadUsageDelta(op)
 		c.admittedWorkloadsCount += op.asSignedOne()
```
mimowo: Hm, I think we could just handle the `admittedWorkloadsCount` inside c.updateAdmittedWorkloadsCount(op) - both for `c`, and the path.

pkg/cache/scheduler/cohort.go
```diff
@@ -96,3 +98,22 @@ func (c *cohort) PathSelfToRoot() iter.Seq[*cohort] {
...
+		return
+	}
+	if hierarchy.HasCycle(c) {
+		return
+	}
+	count := c.admittedWorkloadsCount
+	for ancestor := range c.PathSelfToRoot() {
+		ancestor.admittedWorkloadsCount -= count
+	}
+}
```
mimowo: These two helpers could use one helper like `updateAdmittedWorkloadCount(`)` which accepts the delta count.

site/content/en/docs/reference/metrics.md
mbobrovskyi: https://github.com/kubernetes-sigs/kueue/pull/10152

pkg/cache/scheduler/clusterqueue.go
mbobrovskyi: But c is a `CQ`, and `Parent()` returns a cohort.

pkg/cache/scheduler/cache.go
mbobrovskyi: Agree. But c.hm.DeleteCohort(cohortName) only works with C and CQ interfaces, so we can’t access admittedWorkloadsCount.

pkg/cache/scheduler/cohort.go
mbobrovskyi: Done

pkg/cache/scheduler/clusterqueue.go
mimowo: ok, so the helper function could be accepting CQ, and basically wrap updating for both CQ and the parent cohorts

pkg/cache/scheduler/cache.go
mimowo: I see, ok, we may revisit this in the future, does not look bad now

pkg/cache/scheduler/clusterqueue.go
mimowo: I'm ok keeping as is

PR #10018, opened 2026-03-19
chore: Fix exptostd and usetesting lint issues

alexandear:
#### What type of PR is this?

/kind cleanup

/area testing

Enable `exptostd` and `usetesting` linters. Fix appeared lint issues.

N/A

We need to explicitly enable a few checks from `usetesting` (`os-temp-dir`, `context-background`, `context-todo`) because they are `false` by default.

```release-note
NONE
```

### DIFF
--- a/pkg/cache/queue/cluster_queue_test.go
+++ b/pkg/cache/queue/cluster_queue_test.go
@@ -17,7 +17,6 @@ limitations under the License.
 package queue
 
 import (
-	"context"
 	"slices"
 	"testing"
 	"time"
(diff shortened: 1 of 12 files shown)


pkg/cache/queue/cluster_queue_test.go
```diff
@@ -1270,7 +1269,7 @@ func TestFsAdmission(t *testing.T) {
 				builder = builder.WithObjects(&lq)
 			}
 			client := builder.Build()
-			ctx := context.Background()
+			ctx := t.Context()
```
mbobrovskyi: It looks like this variable is used only once. Can we set it directly?

pkg/controller/jobs/raycluster/common_test.go
```diff
@@ -360,7 +360,7 @@ func TestUpdatePodSets(t *testing.T) {
 	for name, tc := range testCases {
 		t.Run(name, func(t *testing.T) {
 			features.SetFeatureGateDuringTest(t, features.ElasticJobsViaWorkloadSlices, true)
-			ctx := context.Background()
+			ctx := t.Context()
```
mbobrovskyi: Also here.

pkg/cache/queue/cluster_queue_test.go
alexandear: Yes, we can. Done.

Initially, I removed this variable but then decided to return back to minimize changes.

mimowo:
/approve
Thank you 👍

Issue #9439, opened 2026-02-24
MultiKueue: support RayService integration


mimowo:
**What would you like to be added**:

Add RayService support for MultiKueue.

This requires the pre-requisite at the KubeRay side: https://github.com/ray-project/kuberay/issues/4486

However, we probably start prototyping early using the main image, or release candidate of KubeRay.

**Why is this needed**:

To allow scheduling RayService objects via MultiKueue.

mimowo:
cc @hiboyang

NarayanaSabari:
Hi, I'd like to work on this. Here's my understanding of the current state and proposed approach:

**Current state:**
- KubeRay side: `spec.managedBy` on RayService has been merged ([kuberay#4491](https://github.com/ray-project/kuberay/pull/4491)), release planned for March 15
- Kueue side: RayService as a top-level job was merged via [#9102](https://github.com/kubernetes-sigs/kueue/pull/9102) (controller, webhook, RBAC, E2E), but without MultiKueue support
- The vendored KubeRay API in Kueue doesn't yet have `ManagedBy` on `RayServiceSpec`

**Proposed implementation plan:**

1. **Bump the vendored KubeRay dependency** to pick up the `ManagedBy` field on `RayServiceSpec` (either from `main` or an RC once available)
2. **Implement `JobWithManagedBy` interface** on `RayService` in `rayservice_controller.go` - adding `CanDefaultManagedBy()`, `ManagedBy()`, `SetManagedBy()` methods, following the RayJob pattern
3. **Create `rayservice_multikueue_adapter.go`** implementing `MultiKueueAdapter` and `MultiKueueWatcher` - following the existing RayJob adapter pattern (`SyncJob`, `DeleteRemoteObject`, `IsJobManagedByKueue`, `GVK`, `GetEmptyList`, `WorkloadKeysFor`)
4. **Register the adapter** in `IntegrationCallbacks` in the `init()` function
5. **Add unit tests** in `rayservice_multikueue_adapter_test.go`
6. **Add MultiKueue E2E tests** for RayService

One thing to note: RayService is long-running (`Finished()` always returns `false`), so the adapter will continuously sync status rather than waiting for completion.

I plan to raise a PR for this. Let me know if there are any concerns or if I'm missing anything.

mimowo:
Thank you @NarayanaSabari ! This all sounds reasonable. 

[...]

PR #9449, opened 2026-02-24
Add MultiKueue support for RayService

NarayanaSabari:
## What type of PR is this?

/kind feature

## What this PR does / why we need it

Adds MultiKueue support for RayService, enabling RayService workloads to be dispatched across multiple Kueue-managed clusters.

This PR depends on:
- #9885 (refactoring: shared generic Ray MultiKueue adapter)
- #9886 (dependency: kuberay v1.6.0-rc.0 bump)

### Changes in this PR (RayService-specific only):
- `rayservice_multikueue_adapter.go` — RayService MultiKueue adapter using the shared `ray.NewMKAdapter()` (~51 lines)
- `rayservice_multikueue_adapter_test.go` — 7 test cases covering all adapter operations
- `rayservice_controller.go` — implements `JobWithManagedBy` interface (`CanDefaultManagedBy`, `ManagedBy`, `SetManagedBy`) + registers MultiKueue adapter in `init()`
- `wrappers.go` — test helpers (`ManagedBy()`, `StatusConditions()`)

### Key design decisions:
- RayService is **long-running** — `Finished()` always returns `false`, so the MultiKueue adapter continuously syncs status
- Suspend is via `Spec.RayClusterSpec.Suspend` (embedded), not a top-level field
- `ManagedBy` is at the top level of `RayServiceSpec` (like RayJob)

E2E tests will be in a separate follow-up PR (agreed with @hiboyang).

## Which issue(s) this PR fixes

Fixes #9439

Depends on #9885
Depends on #9886

## Does this PR introduce a user-facing change?

```release-note
Add MultiKueue support for RayService, enabling RayService workloads to be dispatched across multiple Kueue-managed clusters.
```

PR #10025, opened 2026-03-19
Add MultiKueue support for RayService

hiboyang:
#### What type of PR is this?

/kind feature

This is the continue work for PR https://github.com/kubernetes-sigs/kueue/pull/9449, to support Kueue 0.17 release.

It is to add MultiKueue support for RayService.

Fixes https://github.com/kubernetes-sigs/kueue/issues/9439

Depends on https://github.com/kubernetes-sigs/kueue/pull/9885
Depends on https://github.com/kubernetes-sigs/kueue/pull/9886

```release-note
MultiKueue: Enabling RayService workloads to be dispatched across multiple Kueue-managed clusters.
```

### DIFF
--- a/test/e2e/multikueue/e2e_test.go
+++ b/test/e2e/multikueue/e2e_test.go
@@ -218,6 +218,17 @@ var _ = ginkgo.Describe("MultiKueue", func() {
 	})
 
 	ginkgo.AfterEach(func() {
+		// Clean up resources created by the RayService test on all clusters.
+		rayServiceConfigMap := &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: "rayservice-hello", Namespace: managerNs.Name}}
+		gomega.Expect(client.IgnoreNotFound(k8sManagerClient.Delete(ctx, rayServiceConfigMap))).To(gomega.Succeed())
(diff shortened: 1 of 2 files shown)


mimowo (commented):
/lgtm
/approve
Thank you 👍

test/e2e/multikueue/e2e_test.go
```diff
@@ -1514,6 +1516,119 @@ var _ = ginkgo.Describe("MultiKueue", func() {
 			})
 		})
 
+		ginkgo.It("Should run a RayService on worker if admitted", func() {
+			kuberayTestImage := util.GetKuberayTestImage()
+
+			// Create ConfigMap with a simple Ray Serve application
+			configMap := &corev1.ConfigMap{
```
mimowo, line 1534: Actually, is the ConfigMap getting deleted at the end of the test? I don't see namespace deleted here: https://github.com/kubernetes-sigs/kueue/blob/main/test/util/util.go#L157-L203

mimowo: So I suspect it might be keeping the deletion of the namespace from succeeding. If this is the case I think we could add the ConfigMaps to the DeleteNamespace resources.

mimowo: I think it looks like we are also missing deletion of RayService

hiboyang: yeah, let me check

PR #10035, opened 2026-03-19
Make RayJob/RayCluster reconcile on pod event

### LINKED ISSUE #9879: Restarted pods remain schedule-gated
**What happened**:
When Kueue with elastic jobs gate is deployed with KubeRay, and a Ray Serve workload with elastic job annotation is created, after killing a Ray worker pod, the restarted pod remains permanently gated with Kueue's elastic job gate. 

  - This is readily reproducible with Kueue 0.15. Seen to happen when the worker pod is scheduled on GPU instance types, but not CPU instance types. Happens when the pod is non-force-deleted, but not when it is force-deleted. 

The restarted pod should get its scheduling gate removed and run normally.

 - Deploy Kueue 0.15 with ElasticJobsViaWorkloadSlices=true feature gate
 - Deploy KubeRay
 - Submit a Ray Serve workload whose Ray Cluster has the annotation and kueue.x-k8s.io/elastic-job: "true" and spec.enableInTreeAutoscaling=true
 - [In our environment, this happens on GPU instances, but not CPU instances. As explained below, this may be incidental.]
 - Kill a worker pod without force deletion. 

The pod remains gated due to the race described below. I have a partial fix for this race. I request community input for this partial fix, and discussions for a full fix. 

### Problem
During a restart scenario, here are the main events.

1. The current pod A gets a deletion timestamp and eventually terminates.
2. KubeRay sees the pod count change and creates a new pod. The new pod has the Kueue schedule gate by default.
3. KubeRay updates Ray Cluster CR status. The following steps happen concurrently.
    1. The Ray cluster reconciliation is triggered in Kueue.
        1. [A] Kueue’s Ray Cluster reconcile loop looks up the pods for this Ray cluster and ungates them.
    2. The informer cache in Kueue gets updated that pod A is deleted.
        1. [B] The old pod A is removed from Kueue’s workload slice index.
    3. The informer cache in Kueue gets updated that pod B is added.
        1. [C] The new pod B gets added to Kueue’s workload slice index.

There is no further update of the Ray Cluster CR, and so no further triggers of Kueue’s reconcile loop.

There are 6 possible sequences in which the 3 events may occur: ABC, ACB, BAC, BCA, CAB, CBA. This creates a race condition, as shown by the following sequence of events.

 * Problem path: If A happens before C, the new pod does not get ungated. That’s because the reconcile loop does not see the new pod, and the reconciler is not triggered by any later event.  The event sequences ABC, ACB and BAC fall under this sequence.
 * Happy path: If C happens before A, the new pod gets ungated. All good. The possible event sequences are CAB, CBA and BCA. 

If a pod for a workload is being deleted, there are 2 sets of scenarios, and any proposed fix should address both:
 * Restart scenarios: The pod crashed, was evicted or was manually deleted. A new equivalent pod will eventually replace it.
 * Downscale due to autoscaling: no new pod will replace it.

### Partial fix
Among the problematic event sequences — ABC, ACB and BAC — the first two can be addressed with a simple fix: when a pod associated with a workload has a deletion timestamp, defer reconciliation for a short time, such as D=1000 milliseconds.

* If this were a pod restart scenario, this gives time for a new pod to be added to the indexer. If this were a downscale scenario, the creation of the new workload to reflect the downscaling may be a little delayed but will happen eventually.
* Caveats with Fix: 
    * Pods may occasionally get stuck during termination and may remain indefinitely in deleting state. For example, finalizers not getting removed, volume unmount failed, etc. In that corner case, Ray Cluster reconciliation may get postponed indefinitely.
    * The fix does not address the sequence BAC. So, after the deferral by D milliseconds, the next reconcile loop may see that B has already happened and C has not. That would correspond to the BAC sequence. So, this issue would recur.

This fix is seen to work in our environment for Ray Serve workloads with D=1000 milliseconds.

### Full fix is difficult
To handle the event sequence BAC, we note that during Ray Cluster reconciliation in Kueue, the old pod A has already been removed from the index, while the new pod has not been added yet. 

This situation can be detected by comparing the total number of pods when the workload was admitted to the currently available set of pods. In that situation, we could defer reconciliation by D seconds, like the partial fix above.

However, this will fail the downscaling scenario — because the pod count is intentionally lower, this change will never create a new workload.

### Workaround for Recovery
The pod remains gated because no further reconcile loop is triggered for that Ray Cluster. We can trigger it manually by applying a dummy annotation to the Ray Cluster. That is seen to ungate the pod. 

**Environment**:
- Kubernetes version (use `kubectl version`): 1.35 (and 1.29)
- Kueue version (use `git describe --tags --dirty --always`): 0.15 
- Cloud provider or hardware configuration: AWS

### LINKED ISSUE #9986: Deleted/restarted pod for autoscaling RayJob stuck in scheduling gated status
**What happened**:

During a RayJob with [autoscaling enabled](https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/configuring-autoscaling.html) running time, if kill a worker pod, a new pod will be started but stuck in scheduling gate status. The new pod does not change to running state.

After the new pod starts, Kueue should ungate it and switch it to running state.

1. Start a RayJob with autoscaling
2. Trigger autoscaling (e.g. scaling-up)
3. Delete a worker pod using kubectl delete
4. Watch the new worker pod created with scheduling gated
5. Keep watching the new worker pod, will see it stuck in scheduling gated, not change to running

RayService seems have similar issue

**Environment**:
- Kubernetes version (use `kubectl version`): any
- Kueue version (use `git describe --tags --dirty --always`): 0.15, probably also in 0.16 and latest
- Cloud provider or hardware configuration: any
- OS (e.g: `cat /etc/os-release`): any
- Kernel (e.g. `uname -a`): any
- Install tools: KubeRay
- Others:

hiboyang:
#### What type of PR is this?

/kind bug

During a RayJob with [autoscaling enabled](https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/configuring-autoscaling.html) running time, if kill a worker pod, a new pod will be started but stuck in scheduling gate status. The new pod does not change to running state due to reconciling never run again.

This PR adds pod level watch for RayJob to trigger reconcile and ungate pods.

RayService also has similar issue.

Fixes #9986
Fixes #9879

```release-note
NONE
```


mimowo:
I would like to take a step back here. The proposed approach adds new interface which wouldn't be needed if solved properly, so I'm hesitant to merge it.

Let's establish first:
1. why the Reconcile for the RayCluster is not triggered when pods fail, or if triggered, why they don't call StartWorkloadSlicePods
2.does the issue also affect other Job integrations - namely k8s batch Job integration?

To determine (1.) I would really like to see the report from the watch on such a RayCluster, by `kubectl get -w --output-watch-events -oyaml`, while the Pods are being replaced.

Now, as mentioned in another comment I believe the proper way of fixing is a new Pod controller filtering Pods by the scheduling gate for elastic Jobs. When it observes such a Pod it triggers reconcile for the associated WorkloadSlice, and the Reconcile function ungates the Pods if the latest Workload in the slice in admitted.

This is basically the same pattern as for TopologyUngater for TAS.

Now, I understand this is much more work to refactor, and the feature is Alpha, and the bug is important. 

So, if we need a quick fix I would rather add 10s RetryAfter which is independent of the Job type, whenever an ElasticJob is running. However, I would then add a graduation criteria for Beta to refactor the ungating model.

cc @sohankunkerkar

pkg/controller/jobs/raycluster/pod_handler.go
```diff
@@ -0,0 +1,104 @@
...
+	// Only process pods managed by Kueue.
+	if pod.Annotations[kueue.WorkloadAnnotation] == "" {
+		return
+	}
+
+	// Only process pods belonging to a RayCluster.
+	if pod.Labels[rayutils.RayClusterLabelKey] == "" {
+		return
+	}
+
```
mimowo, line 94: filter also only Pods with the job-slice scheduling gate

pkg/controller/jobs/raycluster/pod_handler.go
```diff
@@ -0,0 +1,104 @@
...
+	}, &cluster); err != nil {
+		return
+	}
+
+	parentRef := metav1.GetControllerOf(&cluster)
+	if parentRef == nil || parentRef.Kind != h.ParentKind {
+		return
+	}
+
+	q.Add(reconcile.Request{
```
mimowo: Use AddAfter with 1s delay to batch the Reconcile events over time

pkg/controller/jobs/raycluster/pod_handler.go
```diff
@@ -0,0 +1,104 @@
...
+		return
+	}
+
+	// Fetch the RayCluster to find the parent owner.
+	var cluster rayv1.RayCluster
+	if err := h.Client.Get(context.Background(), types.NamespacedName{
+		Namespace: pod.Namespace,
+		Name:      controllerRef.Name,
+	}, &cluster); err != nil {
+		return
```
mimowo, line 122: Log if error

pkg/controller/jobs/raycluster/pod_handler.go
```diff
@@ -0,0 +1,104 @@
...
+	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+)
+
+var _ handler.EventHandler = (*PodHandler)(nil)
+
+// PodHandler maps Pod events to the grandparent owner of a RayCluster by walking the
+// owner chain: Pod -> RayCluster (owner ref) -> parent (owner ref).
+// The parentKind parameter specifies the expected Kind of the RayCluster's controller
+// owner (e.g. "RayJob" or "RayService").
+type PodHandler struct {
```
mimowo, line 46: @hiboyang could you add an e2e test for the issue? It would be great to have a reproducible scenario which triggers the issue first. Then it would also help me better understand the interactions, and would prevent refressions.

pkg/controller/jobs/raycluster/pod_handler.go
hiboyang: good suggestion, added

pkg/controller/jobs/raycluster/pod_handler.go
hiboyang: yes, updated

pkg/controller/jobs/raycluster/pod_handler.go
hiboyang: added

pkg/controller/jobs/raycluster/pod_handler.go
hiboyang: yes, let me add e2e test and see whether it can reproduce the issue

pkg/controller/jobs/raycluster/pod_handler.go
```diff
@@ -0,0 +1,113 @@
...
+func (h *PodHandler) handle(obj client.Object, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	pod, ok := obj.(*corev1.Pod)
+	if !ok {
+		return
+	}
+
+	// Only process pods managed by Kueue.
+	if pod.Annotations[kueue.WorkloadAnnotation] == "" {
+		return
+	}
```
j-skiba, line 78: if we check `WorkloadAnnotation` and then we check
```go
	// Only process pods that have the elastic-job scheduling gate.
	if !utilpod.HasGate(pod, kueue.ElasticJobSchedulingGate) {
		return
	}
```
then shouldn't we check `WorkloadSliceNameAnnotation` instead of `WorkloadAnnotation`?

hiboyang: Hmm... we probably can check both `WorkloadSliceNameAnnotation` and `WorkloadAnnotation`. There are a little difference, `WorkloadSliceNameAnnotation` is for workload slicing enabled pod, `WorkloadAnnotation` is for kueue managed pod. Let me update code to check both.

test/e2e/singlecluster/kuberay_test.go
```diff
@@ -222,12 +222,12 @@ def my_task(x, s):
...
 # run tasks in parallel to trigger autoscaling (scaling up)
-# Use longer sleep (8s) to give autoscaler time to detect demand,
+# Use longer sleep (16s) to give autoscaler time to detect demand,
 # create workload slices, and schedule new workers.
-print(ray.get([my_task.remote(i, 8) for i in range(16)]))
+print(ray.get([my_task.remote(i, 16) for i in range(16)]))
 
 # run tasks in sequence to trigger scaling down
-print([ray.get(my_task.remote(i, 1)) for i in range(32)])`,
+print([ray.get(my_task.remote(i, 1)) for i in range(60)])`,
```
mimowo, line 230: why are these numbers bumped? They will make the test run longer

pkg/controller/jobs/raycluster/pod_handler_test.go
```diff
@@ -0,0 +1,271 @@
...
+					item, shutdown := q.Get()
+					if shutdown {
+						t.Fatal("Queue shut down unexpectedly")
+					}
+					gotRequests = append(gotRequests, item)
+					q.Done(item)
+				}
+			} else {
+				// Confirm nothing was enqueued after the batch period elapses.
+				time.Sleep(waitDuration)
```
mimowo: Can we use fakeClock instead of real clock for that?

pkg/controller/jobs/raycluster/pod_handler.go
```diff
@@ -0,0 +1,118 @@
...
+
+func (h *PodHandler) Delete(_ context.Context, e event.DeleteEvent, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	h.handle(e.Object, q)
+}
+
+func (h *PodHandler) Generic(context.Context, event.GenericEvent, workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+}
+
+func (h *PodHandler) handle(obj client.Object, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	pod, ok := obj.(*corev1.Pod)
```
mimowo: exit the function early if the feature gate ElasticJobsViaWorkloadSlices is disabled

pkg/controller/jobs/raycluster/pod_handler.go
```diff
@@ -0,0 +1,118 @@
...
+
+	parentRef := metav1.GetControllerOf(&cluster)
+	if parentRef == nil || parentRef.Kind != h.ParentKind {
+		return
+	}
+
+	// Only process pods that have the elastic-job scheduling gate.
+	if !utilpod.HasGate(pod, kueue.ElasticJobSchedulingGate) {
+		return
+	}
```
mimowo: This check can go before the RayCluster instance

pkg/controller/jobs/raycluster/pod_handler_test.go
mimowo: Actually, I would rather decompose the code, and add function `ShouldTriggerReconcileForRayClusterOwner(pod) *string` which returns the name of the RayCluster owner if should reconcile. And we could cover this function with unit tests. Then we don't need the complex time manipulation. The e2e test will verify this is all wired properly

pkg/controller/jobs/rayservice/rayservice_controller.go
```diff
@@ -86,7 +87,9 @@ func newJob() jobframework.GenericJob {
 }
 
 func setup(b *builder.Builder, c client.Client) *builder.Builder {
-	return b.Watches(&rayv1.RayCluster{}, handler.EnqueueRequestForOwner(c.Scheme(), c.RESTMapper(), &rayv1.RayService{}, handler.OnlyControllerOwner()))
+	return b.
+		Watches(&rayv1.RayCluster{}, handler.EnqueueRequestForOwner(c.Scheme(), c.RESTMapper(), &rayv1.RayService{}, handler.OnlyControllerOwner())).
+		Watches(&corev1.Pod{}, &raycluster.PodHandler{Client: c, ParentKind: "RayService"})
```
mimowo: Could we register the watch only if ElasticJobs feature gate is enabled?

pkg/controller/jobs/rayjob/rayjob_controller.go
```diff
@@ -88,7 +88,9 @@ func newJob() jobframework.GenericJob {
 }
 
 func setup(b *builder.Builder, c client.Client) *builder.Builder {
-	return b.Watches(&rayv1.RayCluster{}, handler.EnqueueRequestForOwner(c.Scheme(), c.RESTMapper(), &rayv1.RayJob{}, handler.OnlyControllerOwner()))
+	return b.
+		Watches(&rayv1.RayCluster{}, handler.EnqueueRequestForOwner(c.Scheme(), c.RESTMapper(), &rayv1.RayJob{}, handler.OnlyControllerOwner())).
+		Watches(&corev1.Pod{}, &raycluster.PodHandler{Client: c, ParentKind: "RayJob"})
```
mimowo: Could we register the watch only if ElasticJobs feature gate is enabled?

mimowo:
I'm ok with the PR as a quick fix, but I consider this direction as better long term: https://github.com/kubernetes-sigs/kueue/issues/10258

pkg/controller/jobs/rayjob/rayjob_controller.go
hiboyang: yes, updated code

pkg/controller/jobs/rayservice/rayservice_controller.go
hiboyang: yes, updated code

pkg/controller/jobs/raycluster/pod_handler.go
hiboyang: updated

pkg/controller/jobs/raycluster/pod_handler.go
hiboyang: updated

pkg/controller/jobs/raycluster/pod_handler_test.go
hiboyang: Good suggestion, updated

test/e2e/singlecluster/kuberay_test.go
hiboyang: Bumped these numbers to give RayJob enough running time for the extra steps we added: deleting pod and waiting for new pods coming up.

mimowo:
@hiboyang as we merged https://github.com/kubernetes-sigs/kueue/pull/10272 I think this will also solve the problem which this PR is trying to solve. Please verify manually if this is true, if so then I think we can close the PR.

hiboyang:
Yes, https://github.com/kubernetes-sigs/kueue/pull/10272 solved the same problem, thanks! Let me close this one.

PR #8308, opened 2025-12-17
Remove IsSuspended check from CRD MultiKueue adapters to enable status sync during suspension

IrvingMg:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

PR #3685 introduced the `IsSuspended` check for `batch/Job` to prevent a race condition where syncing `status.startTime` before the job was unsuspended would cause spec update failures. This happens because Kubernetes makes `spec.template` immutable when [`status.startTime`](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/job-v1/#JobStatus) is set and the job is not suspended.

PR #4070 copied this check to all CRD-based adapters. However, CRDs like JobSet don't have this Kubernetes core validation. In fact, the [JobSet webhook explicitly allows pod template mutations during suspension for Kueue integration](https://github.com/kubernetes-sigs/jobset/blob/ef5c09e40fd8f1a2576d4ae4c9d1e1402c66a1f5/pkg/webhooks/jobset_webhook.go#L287-L300). Therefore, the check is unnecessary for CRD types and prevents status sync while the job is suspended.
  
Note: `batch/Job` behavior is unchanged, status sync remains blocked while suspended to prevent the `startTime` race condition, regardless of the `MultiKueueBatchJobWithManagedBy` feature gate setting.

#### Which issue(s) this PR fixes:

Fixes #8162

#### Does this PR introduce a user-facing change?

```release-note
MultiKueue: Fixed status sync for CRD-based jobs (JobSet, Kubeflow, Ray, etc.) that was blocked while the local job was suspended.
```

Issue #8162, opened 2025-12-10
MultiKueue doesn't convey full up-to-date status upon workload suspension


mwielgus:
**What happened**:

When a MK jobset workload is suspended, some of the status information may be not up to date:

```
spec:
  suspend: true
status:
  replicatedJobsStatus:
  - active: 1
    ready: 1
```

**What you expected to happen**:

There are no active and ready Jobs in Jobset when the whole thing is suspended.

**Environment**:

- Kueue version: 0.14.4

PR #10041, opened 2026-03-20
Add WithRetryOnConflict to startJob and stopJob Patch calls

### LINKED ISSUE #10040: startJob/stopJob Patch fails permanently on conflict with external controllers
## What happened

In a MultiKueue setup with KubeRay, jobs submitted to worker clusters get permanently stuck in suspended state. The worker's kueue controller fails to unsuspend the job because `startJob()` uses `clientutil.Patch` with strict resourceVersion but **without `WithRetryOnConflict()`**.

When KubeRay or the MultiKueue SyncJob modifies the job object between kueue's read and patch, the patch fails with a conflict error and the job stays suspended forever.

### Production impact

We observed this on a production cluster running kueue v0.16.4 with KubeRay v1.5.0:
- **31 evictions** over 89 minutes on a single worker cluster
- All RayJobs stuck in suspended state indefinitely
- Required rollback to v0.13.3 to restore functionality

`startJob` should retry the patch on conflict (re-read the latest resourceVersion and try again), the same way the scheduler and pod controller already do.

## How to reproduce

The race occurs when three controllers write to the same job object concurrently during the admission-to-unsuspend window on the worker cluster:

1. **KubeRay operator** — adds finalizer, transitions New→Initializing→Suspending→Suspended (4 API writes, each bumping resourceVersion)
2. **MultiKueue SyncJob** (manager) — copies status to the remote job (writes during suspension since #8308 removed the `IsSuspended` guard)
3. **Local kueue** (worker) — calls `startJob()` → `Patch` with stale resourceVersion → **conflict error, no retry**

Timeline:
```
t0: Manager creates remote RayJob (suspended=true)    [rv: 1]
t1: KubeRay adds finalizer                            [rv: 2]
t2: Manager SyncJob copies status                      [rv: 3]
t3: KubeRay state transitions                          [rv: 4,5,6]
t4: Local kueue startJob() → Patch(rv=1)               ← CONFLICT (no retry)
```

In v0.13.3, the `IsSuspended` guard in SyncJob prevented the third writer, making the two-way race (kueue vs KubeRay) narrow enough to rarely trigger. After #8308 correctly removed that guard (fixing #8162), the three-way race is nearly guaranteed.

## Environment

- Kueue version: v0.16.4
- Kubernetes version: v1.31
- KubeRay version: v1.5.0
- MultiKueue: enabled (1 manager, 4 workers)

## Root cause

`startJob` ([reconciler.go:1275](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/jobframework/reconciler.go#L1275)) and `stopJob` ([reconciler.go:1319](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/jobframework/reconciler.go#L1319)) both use `clientutil.Patch` without `WithRetryOnConflict()`:

```go
// startJob — no retry
    return true, job.RunWithPodSetsInfo(ctx, info)
}); err != nil {
    return err
}
```

`WithRetryOnConflict()` is already used in 6 other call sites (scheduler workload updates, pod controller finalizer removal) but was never added to `startJob`/`stopJob`.

## Proposed fix

Add `clientutil.WithRetryOnConflict()` to both `startJob` and `stopJob` Patch calls. This is a 2-line change that uses existing infrastructure. PR incoming.

/kind bug
/area multikueue

trilamsr:
## What type of PR is this?

/kind bug
/area multikueue

`startJob()` and `stopJob()` use `clientutil.Patch` with strict resourceVersion but without `WithRetryOnConflict()`. When external controllers (KubeRay, MultiKueue SyncJob) modify the job object concurrently, the patch fails with a conflict error and the job remains stuck permanently.

This is a 2-line fix that adds `clientutil.WithRetryOnConflict()` — already used in 6 other call sites — to both functions.

Fixes #10040

```release-note
Fix startJob and stopJob to retry on conflict when external controllers (e.g., KubeRay) concurrently modify the job object, preventing jobs from getting permanently stuck in suspended state.
```

## Additional context

The conflict was made much more likely after #8308 removed the `IsSuspended` guard from CRD MultiKueue adapters (correctly fixing #8162), which allowed SyncJob to write during the admission-to-unsuspend window. In MultiKueue + KubeRay setups, three controllers race on the same object:

1. KubeRay operator (finalizer, state transitions — 4 writes per new RayJob)
2. MultiKueue SyncJob (status sync from manager)
3. Local kueue controller (unsuspend via startJob)

Without retry, `startJob`'s patch is nearly guaranteed to hit a stale resourceVersion.

In v0.13.3, the `IsSuspended` guard incidentally prevented the third writer, making the two-way race narrow enough to rarely trigger. The guard removal was correct (fixing stale status), but exposed this latent fragility.

The fix is orthogonal — it doesn't revert #8308, it makes the unsuspend/suspend path resilient to concurrent writes regardless of the source.

---

**Note:** The companion test PR #10058 adds an integration test that reproduces this bug using a client interceptor. Once this fix lands, the test assertion should be updated from `Consistently(Suspend=true)` to `Eventually(Suspend=false)` to verify prompt unsuspension.

### DIFF
--- a/pkg/controller/jobframework/reconciler_test.go
+++ b/pkg/controller/jobframework/reconciler_test.go
@@ -33,6 +33,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/clock"
 	testingclock "k8s.io/utils/clock/testing"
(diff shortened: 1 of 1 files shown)


pkg/controller/jobframework/reconciler_test.go
```diff
@@ -1236,3 +1237,79 @@ func TestReconcileGenericJobWithWaitForPodsReady(t *testing.T) {
 		})
 	}
 }
+
+// TestStartJobRetriesOnConflict verifies that startJob retries when a concurrent
+// modification causes a resourceVersion conflict on the job object.
+// Regression test for https://github.com/kubernetes-sigs/kueue/issues/10040.
+func TestStartJobRetriesOnConflict(t *testing.T) {
```
mimowo, line 1244: /hold
Let's align on the approach for fixing first: https://github.com/kubernetes-sigs/kueue/issues/10040#issuecomment-4096522241

Also, I would like to start with a TDD fashion here by an integration or e2e test which flakes often enough so that we could study the logs and debug,

trilamsr:
Closing this — per @mimowo's feedback, controllers should rely on re-reconcile rather than `WithRetryOnConflict`. After investigating our production logs more carefully, I confirmed that re-reconcile handles startJob conflicts correctly (17 conflicts over 4 hours, all self-healed in 29–152ms on v0.13.3). See updated issue description on #10040 for details.

PR #10042, opened 2026-03-20
TAS NodeHotSwap: Fix performance issue due to listing all Workloads in Reconcile

j-skiba:
#### What type of PR is this?
/kind bug
/area tas

Improves performance of `Reconcile` in node_controller by adding an indexer for workload listing

Fixes ##10037

```release-note
TAS: Improved the performance of the node_controller Reconcile loop by introducing a new field indexer for Workloads.
```

### DIFF
--- a/pkg/controller/tas/indexer/indexer.go
+++ b/pkg/controller/tas/indexer/indexer.go
@@ -35,7 +35,7 @@ const (
 	PodNodeSelectorHostnameKey    = "spec.nodeSelector.hostname"
 	PodNodeNameKey                = "spec.nodeName"
 	ResourceFlavorTopologyNameKey = "spec.topologyName"
-	WorkloadTASNodeKey            = "status.admission.podSetAssignments.topologyAssignment.nodeName"
+	AdmittedWorkloadNodesKey      = "metadata.admittedWorkloadNodes"
 )
 
 func indexPodTAS(o client.Object) []string {
@@ -87,9 +87,9 @@ func indexResourceFlavorTopologyName(o client.Object) []string {
 	return []string{string(*flavor.Spec.TopologyName)}
 }
 
-func indexWorkloadTASNode(o client.Object) []string {
+func indexAdmittedWorkloadNodes(o client.Object) []string {
 	wl, ok := o.(*kueue.Workload)
-	if !ok {
+	if !ok || workload.IsFinished(wl) || workload.IsEvicted(wl) {
 		return nil
 	}
 
@@ -115,8 +115,8 @@ func SetupIndexes(ctx context.Context, indexer client.FieldIndexer) error {
 	if err := indexer.IndexField(ctx, &kueue.ResourceFlavor{}, ResourceFlavorTopologyNameKey, indexResourceFlavorTopologyName); err != nil {
 		return fmt.Errorf("setting index resource flavor topology name: %w", err)
 	}
-	if err := indexer.IndexField(ctx, &kueue.Workload{}, WorkloadTASNodeKey, indexWorkloadTASNode); err != nil {
-		return fmt.Errorf("setting index workload TAS node: %w", err)
+	if err := indexer.IndexField(ctx, &kueue.Workload{}, AdmittedWorkloadNodesKey, indexAdmittedWorkloadNodes); err != nil {
+		return fmt.Errorf("setting index admitted workload nodes: %w", err)
 	}
 	return nil
 }
--- a/pkg/controller/tas/node_controller.go
+++ b/pkg/controller/tas/node_controller.go
@@ -309,18 +309,16 @@ func groupPodsByWorkload(pods []corev1.Pod) map[types.NamespacedName][]*corev1.P
 // or have "late" pods assigned to this node via nodeSelector.
 func (r *nodeReconciler) getWorkloadsOnNode(ctx context.Context, nodeName string, nodeSelectorPodsByWorkload map[types.NamespacedName][]*corev1.Pod) (sets.Set[types.NamespacedName], error) {
 	var workloadsOnNode kueue.WorkloadList
-	if err := r.client.List(ctx, &workloadsOnNode, client.MatchingFields{indexer.WorkloadTASNodeKey: nodeName}); err != nil {
+	if err := r.client.List(ctx, &workloadsOnNode, client.MatchingFields{indexer.AdmittedWorkloadNodesKey: nodeName}); err != nil {
 		return nil, fmt.Errorf("failed to list workloads: %w", err)
 	}
 	tasWorkloadsOnNode := sets.New[types.NamespacedName]()
 	for i := range workloadsOnNode.Items {
 		wl := &workloadsOnNode.Items[i]
-		if workload.IsFinished(wl) || workload.IsEvicted(wl) {
-			continue
-		}
 		tasWorkloadsOnNode.Insert(types.NamespacedName{Name: wl.Name, Namespace: wl.Namespace})
 	}
 
+	logger := r.logger().V(4).WithValues("node", nodeName)
 	// Also find workloads from any pods that are assigned to this node by TopologyAssignment
 	// but not yet bound. These might be stale "late" pods for a workload that has already
 	// been reassigned to another node.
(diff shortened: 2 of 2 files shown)


pkg/controller/tas/indexer/indexer.go
```diff
@@ -85,6 +86,25 @@ func indexResourceFlavorTopologyName(o client.Object) []string {
 	return []string{string(*flavor.Spec.TopologyName)}
 }
 
+func indexWorkloadTASNode(o client.Object) []string {
+	wl, ok := o.(*kueue.Workload)
+	if !ok || wl.Status.Admission == nil {
+		return nil
+	}
+	var nodes []string
```
mimowo: I think this should be set which is later flattened, because multiple PodSets could overlap on the same nodes. Please add a unit test for Reconcile in such case

pkg/controller/tas/indexer/indexer.go
```diff
@@ -85,6 +86,25 @@ func indexResourceFlavorTopologyName(o client.Object) []string {
...
+		if psa.TopologyAssignment == nil || !utiltas.IsLowestLevelHostname(psa.TopologyAssignment.Levels) {
+			continue
+		}
+		for domain := range utiltas.InternalSeqFrom(psa.TopologyAssignment) {
+			if len(domain.Values) > 0 && domain.Count > 0 {
+				nodes = append(nodes, domain.Values[len(domain.Values)-1])
+			}
+		}
+	}
+	return nodes
```
mimowo: I would suggest to extract this to a helper in the utils, and make sure it is covered with unit tests

pkg/controller/tas/indexer/indexer.go
mimowo: The indexer function is called many times, so I would suggest to do a read-only check if there is TAS used first. So that we allocate the memory (for set or slices) only for TAS workloads. IIRC we already have helpers to determine if TAS is used by workload.

pkg/controller/tas/indexer/indexer.go
```diff
@@ -88,21 +89,11 @@ func indexResourceFlavorTopologyName(o client.Object) []string {
 
 func indexWorkloadTASNode(o client.Object) []string {
 	wl, ok := o.(*kueue.Workload)
-	if !ok || wl.Status.Admission == nil {
+	if !ok || !workload.IsAdmittedByTAS(wl) {
 		return nil
 	}
```
mimowo: I think it makes sense to move the check under `NodeNamesInPodSetAssignments`, wdyt

pkg/controller/tas/indexer/indexer.go
j-skiba: changed to
```go
	if !ok || !workload.IsAdmittedByTAS(wl) {
		return nil
	}
```

---
I also changed nodes slice to a set

pkg/controller/tas/indexer/indexer.go
j-skiba: it would be import cycle. workloads util already uses util tas

pkg/controller/tas/indexer/indexer.go
j-skiba: done

pkg/controller/tas/indexer/indexer.go
mimowo: I see, so put the function `NodeNamesInPodSetAssignments` under workload package, and call the TAS. This will allow us to strengthen testing by unit tests.

mimowo: Actually, there is a nuance `IsAdmittedByTAS` only checks if "Admitted" which is a stronger condition than currently where we check that the workload has status.admission != nil. I think the intention is to run NodeHotSwap only for Admitted, so it would be good to cover such corner cases with unit tests.

pkg/util/tas/tas_assignment.go
```diff
@@ -313,6 +314,22 @@ func ComputeUsagePerDomain(ta *TopologyAssignment, singlePodRequests resources.R
 	return usage
 }
 
+// NodeNamesInPodSetAssignments extracts the unique set of node names that a Workload is assigned to.
+func NodeNamesInPodSetAssignments(psas []kueue.PodSetAssignment) []string {
```
mimowo: Make sure this is covered 100% by unit tests

pkg/controller/tas/indexer/indexer.go
j-skiba: ok, that sounds better :+1:

I moved that to workload package and also renamed to `TASAssignedNodeNames`.

pkg/util/tas/tas_assignment.go
j-skiba:
```bash
$ go test ./pkg/workload/ -coverprofile=cover.out && go tool cover -func=cover.out | grep TASAssignedNodeNames
ok      sigs.k8s.io/kueue/pkg/workload  0.322s  coverage: 55.2% of statements
sigs.k8s.io/kueue/pkg/workload/workload.go:1814:        TASAssignedNodeNames                    100.0%
```

pkg/controller/tas/indexer/indexer.go
j-skiba: > Actually, there is a nuance IsAdmittedByTAS only checks if "Admitted" which is a stronger condition than currently where we check that the workload has status.admission != nil. I think the intention is to run NodeHotSwap only for Admitted, so it would be good to cover such corner cases with unit tests.

done

pkg/workload/workload_test.go
```diff
@@ -2654,3 +2654,185 @@ func TestSchedulingHash(t *testing.T) {
...
+				Status: kueue.WorkloadStatus{
+					Admission: nil,
+				},
+			},
+			want: nil,
+		},
+		"admitted without tas": {
+			wl: &kueue.Workload{
+				Status: kueue.WorkloadStatus{
+					Conditions: []metav1.Condition{{Type: kueue.WorkloadAdmitted, Status: metav1.ConditionTrue}},
```
mimowo, line 2736: Add a test case for Workload with QuotaReserved condtion, but with "Assignment" already - waiting for AdmissionCheck

j-skiba: added 👍

pkg/controller/tas/node_controller.go
```diff
@@ -308,27 +308,31 @@ func groupPodsByWorkload(pods []corev1.Pod) map[types.NamespacedName][]*corev1.P
...
 		}
-		wlKey := types.NamespacedName{Name: wl.Name, Namespace: wl.Namespace}
-		if wl.Status.Admission != nil && utiltas.HasTASAssignmentOnNode(wl.Status.Admission.PodSetAssignments, nodeName) {
-			tasWorkloadsOnNode.Insert(wlKey)
+		tasWorkloadsOnNode.Insert(types.NamespacedName{Name: wl.Name, Namespace: wl.Namespace})
+	}
+
+	// Also find workloads from any pods that are assigned to this node by TopologyAssignment
+	// but not yet bound. These might be stale "late" pods for a workload that has already
+	// been reassigned to another node.
```
mimowo, line 324: Do we have a unit test in node_controller_test.go for the Reconcile covering such cases?

j-skiba: we didn't have such case in unit tests, it was tested just in integration tests.

I added that and also added two test cases for evicted and finished workloads - https://github.com/kubernetes-sigs/kueue/pull/10042/changes#diff-4a0cb79ce927ca48906417d823ad3303285b2c08b5a74c0a66d66214578eb654L319

pkg/controller/tas/indexer/indexer.go
```diff
@@ -34,6 +35,7 @@ const (
 	PodNodeSelectorHostnameKey    = "spec.nodeSelector.hostname"
 	PodNodeNameKey                = "spec.nodeName"
 	ResourceFlavorTopologyNameKey = "spec.topologyName"
+	WorkloadTASNodeKey            = "status.admission.podSetAssignments.topologyAssignment.nodeName"
```
mimowo: I would propose to make it a bit more scoped, like AdmittedWorkloadNodesKey with `metadata.admittedWorkloadNodes`, because this is not so generic really. And I would already include here the filtering which excludes Finished and Evicted workloads. Keeping all Finished workloads may impact memory signifficantly actually

pkg/controller/tas/node_controller.go
```diff
@@ -308,27 +308,31 @@ func groupPodsByWorkload(pods []corev1.Pod) map[types.NamespacedName][]*corev1.P
...
 			continue
 		}
-
-		// Also find workloads from any pods that are assigned to this node by TopologyAssignment
-		// but not yet bound. These might be stale "late" pods for a workload that has already
-		// been reassigned to another node.
-		if len(nodeSelectorPodsByWorkload[wlKey]) > 0 {
-			tasWorkloadsOnNode.Insert(wlKey)
+		var wl kueue.Workload
+		if err := r.client.Get(ctx, wlKey, &wl); err == nil {
```
mimowo: Do not silence the error, completely, log it at least

j-skiba: added v4 log on error

pkg/controller/tas/indexer/indexer.go
j-skiba: I added IsEvicted` and `IsFinished` checks to the indexer, also renamed to `metadata.admittedWorkloadNodes`

PR #10058, opened 2026-03-20
Add integration test reproducing startJob Patch conflict race

trilamsr:
## What type of PR is this?

/kind bug
/area jobs

Adds tests that reproduce the startJob Patch conflict race condition described in #10040. Both tests **fail without the fix**, serving as regression tests.

**Unit test** (`TestStartJobRetriesOnConflict`): Injects a single Patch conflict via interceptor and asserts that `ReconcileGenericJob` retries and succeeds.

**Integration test** (`Job controller startJob conflict retry`): Injects 12 consecutive Patch conflicts for RayJob objects via a manager-level client interceptor. After admitting a workload, asserts the job gets unsuspended (`Eventually(Suspend=false)`) within the test timeout. Without retry-on-conflict, exponential backoff accumulates to ~20s, exceeding the 10s timeout.

Ref #10040

```release-note
NONE
```

## Additional context

Companion to #10041 (which contains the fix). The tests assert the correct behavior — both fail on current main, demonstrating the bug.

The conflict race occurs in MultiKueue + KubeRay deployments where three controllers write to the same job object during the admission-to-unsuspend window, causing `clientutil.Patch` to fail with 409 Conflict. Without retry, each failure triggers controller-runtime exponential backoff (5ms base, doubling), keeping the job suspended for extended periods.

### DIFF
--- a/pkg/controller/jobframework/reconciler_test.go
+++ b/pkg/controller/jobframework/reconciler_test.go
@@ -33,6 +33,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/clock"
 	testingclock "k8s.io/utils/clock/testing"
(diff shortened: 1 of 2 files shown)


mimowo (commented):
/ok-to-test

trilamsr:
Closing — the tests were built around the `WithRetryOnConflict` approach which isn't the right fix (see #10040 comment). The startJob conflict is handled correctly by re-reconcile already.

Issue #9821, opened 2026-03-12
integration tests fail to start due to "failed to start the controlplane. retried 5 times: exec: "etcd": executable file not found in $PATH"


mimowo:
**Which test is flaking?**:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/periodic-kueue-test-integration-release-0-15/2031910147336441856
**Link to failed CI job or steps to reproduce locally**:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/periodic-kueue-test-integration-release-0-15/2031910147336441856
**Failure message or logs**:
```
  github.com/onsi/ginkgo/v2/internal.(*Suite).runNode.func3
  	/home/prow/go/src/kubernetes-sigs/kueue/vendor/github.com/onsi/ginkgo/v2/internal/suite.go:942
  [FAILED] in [BeforeSuite] - /home/prow/go/src/kubernetes-sigs/kueue/vendor/github.com/onsi/ginkgo/v2/internal/suite.go:338 @ 03/12/26 01:55:37.419
  << Timeline
  [FAILED] Unexpected error:
      <*fmt.wrapError | 0xc0002c06c0>: 
      unable to start control plane itself: failed to start the controlplane. retried 5 times: exec: "etcd": executable file not found in $PATH
      {
          msg: "unable to start control plane itself: failed to start the controlplane. retried 5 times: exec: \"etcd\": executable file not found in $PATH",
          err: <*fmt.wrapError | 0xc0002c06a0>{
              msg: "failed to start the controlplane. retried 5 times: exec: \"etcd\": executable file not found in $PATH",
              err: <*exec.Error | 0xc0002c0640>{
                  Name: "etcd",
                  Err: <*errors.errorString | 0x4d76dc0>{
                      s: "executable file not found in $PATH",
                  },
              },
          },
      }
  occurred
```

**Anything else we need to know?**:

would it help if we increased the number of retries to 10? Maybe we could experiment lowering the number of retries to see if then the error is more common.

mimowo:
/close
Closing stale tickets which haven't re-occur in 2 months, because maybe some of them are already fixed along the way. We will re-open when they re-occur again.

PR #10059, opened 2026-03-20
Fail fast for empty kubebuilder assets

PannagaRao:

#### What type of PR is this?
/kind bug
/kind flake

Fail fast when setup-envtest fails to download binaries instead of running all test suites only to fail in BeforeSuite with etcd not found in $PATH

Fixes # https://github.com/kubernetes-sigs/kueue/issues/9821

```release-note
NONE
```

### DIFF
--- a/Makefile-test.mk
+++ b/Makefile-test.mk
@@ -20,6 +20,7 @@ GO_TEST_FLAGS ?= -race
 
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION ?= 1.35
+KUBEBUILDER_ASSETS = $(or $(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path),$(error setup-envtest failed to download binaries. KUBEBUILDER_ASSETS is empty))
 
 TEST_LOG_LEVEL ?= -3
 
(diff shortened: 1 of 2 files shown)


Makefile-test.mk
```diff
@@ -95,7 +95,7 @@ test: gotestsum ## Run tests.
 
 .PHONY: test-integration
 test-integration: compile-crd-manifests envtest ginkgo dep-crds kueuectl ginkgo-top ## Run integration tests for all singlecluster suites.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" \
+	KUBEBUILDER_ASSETS="$(or $(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path),$(error setup-envtest failed to download binaries. KUBEBUILDER_ASSETS is empty))" \
```
mbobrovskyi: It looks like we are executing the same logic in many targets. Could we refactor this to avoid calling `$(shell ...)` inside the target? I think it would be cleaner to move the ENVTEST path validation to a separate variable. WDYT?

PannagaRao: Good call! Moved the validation into a single variable that all targets reference. Also applied the same change to the kueue-populator Makefile. PTAL!

PR #8189, opened 2025-12-11
Fix MultiKueue admission check stuck pending for pod groups

IrvingMg:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

Fixes MultiKueue admission check remaining stuck in Pending state for pod groups. The pod's `KeepAdmissionCheckPending()` was returning true, preventing the admission check from transitioning to Ready when the workload got reservation on a worker cluster. This caused the workload to never show Admitted: True.

#### Which issue(s) this PR fixes:

Fixes #5891 

#### Does this PR introduce a user-facing change?

```release-note
MultiKueue: fix the bug that for Pod integration the AdmissionCheck status would be kept Pending indefinitely,
even when the Pods are already running.

The analogous fix is also done for the batch/Job when the MultiKueueBatchJobWithManagedBy feature gate  is disabled.
```

PR #8708, opened 2026-01-20
Add log level override for concurrent modification errors

dkaluza:
#### What type of PR is this?
/kind bug

#### What this PR does / why we need it:
Users are flooded with reconciler errors that cannot be easily hidden with configuration. This PR overrides the log level of reconciler error and sets it to warning by using a custom zap.Core wrapper. 

This PR also introduces logs capturing to integration test logger, so they can be easily used in assertions. These capturing is used in newly proposed integration test checking that the issue is resolved.

#### Which issue(s) this PR fixes:
Fixes #8559

#### Special notes for your reviewer:
The fact that reconciler errors consistently are thrown in the proposed integration test might be a coincidence. A better test candidate idea is welcomed (I still lack knowledge about the system to judge that, but I guess it will not be a problem to move it when it becomes flaky). 

Maybe it is also worth it to disable stack trace for those errors?

#### Does this PR introduce a user-facing change?
```release-note
Observability: do not use "error" for logging when an update request fails because the underlying object was modified
concurrently. Now we use V3 logging level for such cases and don't print stracktrace. 

This is an example log message: "Operation cannot be fulfilled on podtemplates \"job\": the object has been modified;
please apply your changes to the latest version and try again"
```

PR #10061, opened 2026-03-20
Apply ShouldSkipLocalExecution to all job types, not just ComposableJob

### LINKED ISSUE #10040: startJob/stopJob Patch fails permanently on conflict with external controllers

trilamsr:
## What type of PR is this?

/kind bug
/area multikueue

`ShouldSkipLocalExecution` in `startJob` was gated behind `isComposable`, so non-ComposableJob types like RayJob always ran `startJob` on the manager cluster — even when the workload had a MultiKueue admission check.

Since RayJobs on the manager have `managedBy=kueue.x-k8s.io/multikueue` set (KubeRay skips them), the only concurrent writers on the local RayJob were `startJob` and `SyncJob` — a 2-way race. Each write bumps `resourceVersion`, causing the other's strict MergePatch to fail with 409 Conflict.

The conflict self-resolves (startJob stops once it succeeds), but the error storm (133 startJob conflicts + 264 multikueue errors in 2.5h) makes the system look broken and can trigger operational responses (controller restarts) that cause real damage — restarting the controller drops remote cluster connections, and workloads whose AC was already `CheckStateReady` get evicted after `workerLostTimeout` (15 min) with "Reserving remote lost".

Beyond the conflicts, startJob running on the manager is architecturally wrong — the job should only be unsuspended on the worker.

### The fix

Remove the `isComposable` gate so `ShouldSkipLocalExecution` applies to all job types with a MultiKueue admission check. This eliminates startJob on the manager entirely — zero conflicts, zero error noise.

### Why removing the gate entirely

The gate was introduced in #8189 as `isComposable || isBatchJobWithoutManagedBy`. The idea was that `managedBy` on the job already prevents external controllers from touching it, making `ShouldSkipLocalExecution` redundant.

When `MultiKueueBatchJobWithManagedBy` went GA, `isBatchJobWithoutManagedBy` was always false and got removed, leaving only `isComposable` — which covers only Pod (the sole ComposableJob implementer). All other 17 job types with MultiKueue adapters (batch Job, RayJob, MPIJob, PyTorchJob, JobSet, etc.) were unprotected.

I considered three approaches:

1. **Remove the gate entirely** (this PR) — `ShouldSkipLocalExecution` applies to all job types
2. **Add RayJob explicitly** — fragile, every new job type risks the same bug
3. **Check if `managedBy` is set, fall back otherwise** — more complex, `GenericJob` doesn't expose `managedBy`

Option 1 is the right choice because:

- `ShouldSkipLocalExecution` is cheap (one `client.Get` on the AdmissionCheck, served from informer cache)
- `managedBy` and `ShouldSkipLocalExecution` protect against different things: `managedBy` stops external controllers, `ShouldSkipLocalExecution` stops the jobframework reconciler. Defense in depth
- No future job type integration can regress this path

### Production evidence

MultiKueue setup (1 manager, 4 workers) with KubeRay v1.5.0, v0.16.4 upgrade (2.5h window before rollback):

| Metric | v0.13.3 (4 hours) | v0.16.4 (2.5 hours) |
|--------|-------------------|---------------------|
| startJob conflicts | 17 (workers only, self-healed) | 133 (on manager — shouldn't happen) |
| MultiKueue reconciler errors | ~20 | 264 |
| Evictions | 0 | 23 (from controller restarts during debugging) |

The evictions were caused by controller restarts dropping remote connections, not directly by the conflicts. But the error storm from the bug is what prompted the restarts.

Fixes #10040

```release-note
Fixed ShouldSkipLocalExecution to apply to all job types in MultiKueue, not just ComposableJob. Previously, non-ComposableJob types (e.g., RayJob) always ran startJob on the manager, causing Patch conflicts with SyncJob and error noise that could lead to operational issues.
```

### DIFF
--- a/pkg/controller/jobframework/reconciler.go
+++ b/pkg/controller/jobframework/reconciler.go
@@ -1252,18 +1252,14 @@ func (r *JobReconciler) startJob(ctx context.Context, job GenericJob, object cli
 
 	log := ctrl.LoggerFrom(ctx)
 	if features.Enabled(features.MultiKueue) {
-		_, isComposable := job.(ComposableJob)
-
-		if isComposable {
-			skip, err := admissioncheck.ShouldSkipLocalExecution(ctx, r.client, wl)
-			if err != nil {
-				log.V(3).Info("Failed to check for MultiKueue admission check", "workload", klog.KObj(wl), "error", err)
-				return err
-			}
-			if skip {
-				log.V(3).Info("Workload has MultiKueue admission check, skipping local execution", "workload", klog.KObj(wl))
-				return nil
-			}
+		skip, err := admissioncheck.ShouldSkipLocalExecution(ctx, r.client, wl)
+		if err != nil {
+			log.V(3).Info("Failed to check for MultiKueue admission check", "workload", klog.KObj(wl), "error", err)
+			return err
+		}
+		if skip {
+			log.V(3).Info("Workload has MultiKueue admission check, skipping local execution", "workload", klog.KObj(wl))
+			return nil
 		}
 	}
 
--- a/pkg/controller/jobframework/reconciler_test.go
+++ b/pkg/controller/jobframework/reconciler_test.go
@@ -1236,3 +1236,80 @@ func TestReconcileGenericJobWithWaitForPodsReady(t *testing.T) {
 		})
 	}
 }
+
+// Regression test for https://github.com/kubernetes-sigs/kueue/issues/10040.
+// Verifies that ShouldSkipLocalExecution applies to all job types, not just ComposableJob.
+// Without the fix, startJob runs on the manager for non-ComposableJob types (e.g., RayJob),
+// creating a 2-way race between startJob and SyncJob on the local job object.
+func TestStartJobSkippedForMultiKueueNonComposableJob(t *testing.T) {
+	var (
+		testJobName        = "test-job"
+		testLocalQueueName = kueue.LocalQueueName("test-lq")
+		testGVK            = batchv1.SchemeGroupVersion.WithKind("Job")
+		testACName         = "multikueue-check"
+	)
+
+	features.SetFeatureGateDuringTest(t, features.MultiKueue, true)
+
+	testJob := testingjob.MakeJob(testJobName, metav1.NamespaceDefault).
+		UID(testJobName).Queue(testLocalQueueName).Suspend(true).Obj()
+
+	ac := utiltestingapi.MakeAdmissionCheck(testACName).
+		ControllerName(kueue.MultiKueueControllerName).
+		Obj()
+
+	now := time.Now()
+	admission := utiltestingapi.MakeAdmission("test-cq").Obj()
+	wl := utiltestingapi.MakeWorkload("job-test-job", metav1.NamespaceDefault).
+		ResourceVersion("1").
+		Finalizers(kueue.ResourceInUseFinalizerName).
+		Label(constants.JobUIDLabel, testJobName).
+		ControllerReference(testGVK, testJobName, testJobName).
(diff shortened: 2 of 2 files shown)


trilamsr:
Thanks @mimowo, that makes sense. We'll close this PR — the `suspend=false` on the manager being by design tracks with what we saw on closer inspection: the 409 conflicts self-resolved within a few reconcile cycles, and the actual evictions were caused by us restarting the controller repeatedly during debugging (which dropped remote watches and triggered workerLostTimeout).

We'll upgrade our setup to v0.16.4 again and monitor without intervening. If the race turns out to be genuinely problematic at our scale we'll open a new issue with the data. A cherry-pick of #8708 would be nice to have for the log noise but isn't blocking.

Thanks for the quick engagement on this.

PR #10077, opened 2026-03-23
Add documentation for DRA integration (KEP-2941)

sohankunkerkar:
#### What type of PR is this?
/kind documentation

Fixes https://github.com/kubernetes-sigs/kueue/issues/9868

```release-note
NONE
```

### DIFF
--- a/site/content/en/docs/concepts/dynamic_resource_allocation.md
+++ b/site/content/en/docs/concepts/dynamic_resource_allocation.md
@@ -1,0 +1,151 @@
+---
+title: "Dynamic Resource Allocation"
+date: 2026-03-22
+weight: 7
+description: >
+  Quota management for workloads using Kubernetes Dynamic Resource Allocation (DRA).
+---
+
+## Dynamic Resource Allocation
+
+[Dynamic Resource Allocation (DRA)](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/)
+is a Kubernetes API for requesting and managing hardware devices such as GPUs,
+FPGAs, and network adapters. Kueue can account for DRA devices in quota
+management through two paths:
+
+1. **ResourceClaimTemplate path**: Pods explicitly reference a
+   `ResourceClaimTemplate` that specifies a device request. Kueue maps each
+   `DeviceClass` referenced by the claim to a logical resource name using
+   `deviceClassMappings` in the Kueue Configuration.
+
+2. **Extended resource path**: Pods request DRA devices using the traditional
+   `resources.requests` syntax (e.g., `nvidia.com/gpu: 1`). When the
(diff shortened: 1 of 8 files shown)


site/content/en/docs/tasks/run/dra.md
```diff
@@ -0,0 +1,230 @@
...
+weight: 7
+description: >
+  Run workloads that use Dynamic Resource Allocation with Kueue quota management.
+---
+
+This page shows you how to run workloads that use
+[Dynamic Resource Allocation (DRA)](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/)
+with Kueue for quota management.
+
+This guide is for [batch users](/docs/tasks#batch-user) and
```
kannon92: this doesn't really match the rest of the docs under tasks/run.
This was supposed to focus on batch user persona while the manage section was more on admins.

I think we should keep this separation here. Let's highlight DRA as an admin.

And then we can focus on DRA as a user in this page.

site/content/en/docs/concepts/dynamic_resource_allocation.md
```diff
@@ -0,0 +1,146 @@
+---
+title: "Dynamic Resource Allocation"
+date: 2026-03-22
+weight: 7
+description: >
+  Quota management for workloads using Kubernetes Dynamic Resource Allocation (DRA).
+---
+
+## Overview
```
kannon92:
```suggestion
## Dynamic Resource Allocation
```

Following other pages we don't start with overview.

site/content/en/docs/concepts/dynamic_resource_allocation.md
```diff
@@ -0,0 +1,146 @@
...
+`deviceClassMappings`. This mapping tells Kueue which logical resource name
+to charge quota against. The number of units charged is determined by the
+`count` field in the device request (default 1).
+
+Only the `ExactCount` allocation mode is supported. CEL selectors and the
+`All` allocation mode are not supported in alpha.
+
+For setup instructions, see
+[Set Up Dynamic Resource Allocation](/docs/tasks/manage/setup_dra).
+
```
kannon92, line 53: Should we have an alert here that this feature gate is only available on k8s clusters where DRA is v1?

I think 1.34 and on right?

site/content/en/docs/tasks/run/dra.md
kannon92: This was done. Thanks for updating that.

site/content/en/docs/concepts/dynamic_resource_allocation.md
sohankunkerkar: Added here: https://github.com/kubernetes-sigs/kueue/pull/10077/changes#diff-ece880f7573221f2308a19af603904198effaa0b793d4f8517e88e85ccd511bdR28

PR #8342, opened 2025-12-18
[TAS] Do not reconcile on Nodes heartbeats

PBundyra:
#### What type of PR is this?
/kind bug

#### Which issue(s) this PR fixes:

Fixes #8340

#### Special notes for your reviewer:
Tested the changes manually on kind cluster:
Before the change on each heartbeat Kueue logged `Reconcile TAS Resource Flavor` (every 5 mins per node)
After the change there's no more reconciles

#### Does this PR introduce a user-facing change?

```release-note
TAS: fix a performance bug that continues reconciles of TAS ResourceFlavor (and related ClusterQueues) 
were triggered by updates to Nodes' heartbeat times.
```

PR #10078, opened 2026-03-24
Strip managedFields from informer cache to reduce memory footprint

jzhaojieh:
#### What type of PR is this?
/kind bug

Sets `DefaultTransform` to `TransformStripManagedFields()` on the
controller-runtime cache, stripping `metadata.managedFields` from all
objects before they enter the informer store.

Audit that outside of unit tests managedFields is [not used](https://github.com/search?q=repo%3Akubernetes-sigs%2Fkueue+managedFields+language%3AGo&type=code&l=Go). There was one other attempt to [strip these fields](https://github.com/kubernetes-sigs/kueue/pull/8342/changes) when comparing whether Node objects have changed 

```release-note
Strip managedFields from informer cache via DefaultTransform to reduce memory footprint on large clusters.
```

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -380,6 +380,9 @@ objectRetentionPolicies:
 		cmpopts.IgnoreUnexported(net.ListenConfig{}),
 		cmpopts.IgnoreFields(ctrl.Options{}, "Scheme", "Logger"),
 		cmpopts.IgnoreFields(webhook.Options{}, "TLSOpts"),
+		// DefaultTransform is a function and cannot be compared with cmp.Diff;
+		// it is tested separately in TestDefaultTransformStripsManagedFields.
+		cmpopts.IgnoreFields(ctrlcache.Options{}, "DefaultTransform"),
 	}
 
 	// Ignore the controller manager section since it's side effect is checked against
@@ -860,6 +863,9 @@ objectRetentionPolicies:
 				}
 				if diff := cmp.Diff(tc.wantOptions, options, ctrlOptsCmpOpts...); diff != "" {
 					t.Errorf("Unexpected options (-want +got):\n%s", diff)
+				}
+				if options.Cache.DefaultTransform == nil {
+					t.Error("Expected DefaultTransform to be set, got nil")
 				}
 			} else {
 				if diff := cmp.Diff(tc.wantError.Error(), err.Error()); diff != "" {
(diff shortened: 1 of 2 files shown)


pkg/config/config_test.go
```diff
@@ -1400,3 +1400,147 @@ func TestConfigureClusterProfileCache(t *testing.T) {
...
+	cfg := &configapi.Configuration{
+		Namespace: ptr.To(configapi.DefaultNamespace),
+	}
+	addCacheByObjectTo(o, cfg)
+
+	if o.Cache.DefaultTransform == nil {
+		t.Fatal("DefaultTransform is nil, cannot test transform behavior")
+	}
+
+	t.Run("strips managedFields and preserves object data", func(t *testing.T) {
```
mimowo: nit, combine the two test cases into using table test cases, use pod and wantPod parameters to avoid checking field by field, but use cmp.Diff to check that the diff iis empty,

pkg/config/config_test.go
```diff
@@ -1400,3 +1400,147 @@ func TestConfigureClusterProfileCache(t *testing.T) {
...
+		wantTransform bool
+		wantByObject  bool
+	}{
+		"valid namespace sets DefaultTransform and ByObject": {
+			namespace:     ptr.To(configapi.DefaultNamespace),
+			wantTransform: true,
+			wantByObject:  true,
+		},
+		"nil namespace is noop": {
+			namespace:     nil,
```
mimowo: nit, but this is a test case which cannot happen. Maybe we could move the testing to the TestLoad function which is the main entry point?

mimowo: Since it happens in all test cases we could I think just add an extra assert in the  TestLoad, something like

```golang
				if o.Cache.DefaultTransform == nil {
					t.Error("Expected DefaultTransform to be set, got nil")
				}
```

pkg/config/config_test.go
```diff
@@ -1400,3 +1400,147 @@ func TestConfigureClusterProfileCache(t *testing.T) {
...
+func TestDefaultTransformStripsManagedFields(t *testing.T) {
+	o := &ctrl.Options{}
+	cfg := &configapi.Configuration{
+		Namespace: ptr.To(configapi.DefaultNamespace),
+	}
+	addCacheByObjectTo(o, cfg)
+
+	if o.Cache.DefaultTransform == nil {
+		t.Fatal("DefaultTransform is nil, cannot test transform behavior")
+	}
```
mimowo: nit, but I would rather go via entire Load or at least addTo

PR #10080, opened 2026-03-24
Clear metrics after deleting an implicit Cohort.

### LINKED ISSUE #10057: Cohorts when creating, modifying and removing correctly handles cohort metrics when workload admitted with admission check
**Which test is flaking?**:
Cohorts when creating, modifying and removing correctly handles cohort metrics when workload admitted with admission check
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10035/pull-kueue-test-integration-baseline-main/2035039297333104640
**Failure message or logs**:
```
{Timed out after 10.001s.
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/kueue/test/util/metrics.go:230 with:
Expected
    <int>: 5
to equal
    <int>: 2 failed [FAILED] Timed out after 10.001s.
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/kueue/test/util/metrics.go:230 with:
Expected
    <int>: 5
to equal
    <int>: 2
In [It] at: /home/prow/go/src/sigs.k8s.io/kueue/test/integration/singlecluster/scheduler/fairsharing/metrics_test.go:524 @ 03/20/26 17:20:10
}
```

mbobrovskyi:
#### What type of PR is this?
/kind bug
/kind failing-test
/kind flake

Clear metrics after deleting a Cohort.

Fixes #10057

```release-note
Observability: Fix a bug where kueue_cohort_subtree_admitted_workloads_total and kueue_cohort_subtree_admitted_active_workloads metrics could include results for an implicit root Cohort after deletion of a child Cohort or ClusterQueue.
```

### DIFF
--- a/pkg/cache/scheduler/cache.go
+++ b/pkg/cache/scheduler/cache.go
@@ -484,6 +484,7 @@ func (c *Cache) UpdateClusterQueue(log logr.Logger, cq *kueue.ClusterQueue) erro
 	if err := cqImpl.updateClusterQueue(log, cq, c.resourceFlavors, c.admissionChecks, oldParent); err != nil {
 		return err
 	}
+	c.handleParentUpdate(oldParent)
 	for _, qImpl := range cqImpl.localQueues {
 		if qImpl == nil {
 			return errQNotFound
@@ -516,9 +517,10 @@ func (c *Cache) DeleteClusterQueue(cq *kueue.ClusterQueue) {
 	c.hm.DeleteClusterQueue(cqName)
 	metrics.ClearCacheMetrics(cq.Name)
 
-	// Update cohort resources after deletion
 	if parent != nil {
+		// Update cohort resources after deletion
 		updateCohortTreeResourcesIfNoCycle(parent)
+		c.handleParentUpdate(parent)
 	}
 }
 
@@ -530,18 +532,27 @@ func (c *Cache) AddOrUpdateCohort(apiCohort *kueue.Cohort) error {
 	cohort := c.hm.Cohort(cohortName)
 	oldParent := cohort.Parent()
 	c.hm.UpdateCohortEdge(cohortName, apiCohort.Spec.ParentName)
-	return cohort.updateCohort(apiCohort, oldParent)
+	err := cohort.updateCohort(apiCohort, oldParent)
+	if err != nil {
+		return err
+	}
+	c.handleParentUpdate(oldParent)
+	return nil
 }
 
 func (c *Cache) DeleteCohort(cohortName kueue.CohortReference) {
 	c.Lock()
 	defer c.Unlock()
 
+	var parent *cohort
 	if cohort := c.hm.Cohort(cohortName); cohort != nil {
 		cohort.updateAdmittedWorkloadsCount(-cohort.admittedWorkloadsCount)
+		metrics.ClearCohortAdmittedWorkloadsMetrics(cohort.Name)
+		parent = cohort.Parent()
 	}
 
 	c.hm.DeleteCohort(cohortName)
+	c.handleParentUpdate(parent)
 
 	// If the cohort still exists after deletion, it means
 	// that it has one or more children referencing it.
@@ -549,6 +560,21 @@ func (c *Cache) DeleteCohort(cohortName kueue.CohortReference) {
 	if cohort := c.hm.Cohort(cohortName); cohort != nil {
 		updateCohortResourceNode(cohort)
 	}
+}
+
+func (c *Cache) handleParentUpdate(cachedParent *cohort) {
+	if cachedParent == nil {
+		return
+	}
+	if updatedParent := c.hm.Cohort(cachedParent.Name); updatedParent != nil {
+		if updatedParent.IsExplicit() {
+			return
+		}
+		if len(updatedParent.ChildCohorts()) > 0 || len(updatedParent.ChildCQs()) > 0 {
+			return
+		}
+	}
+	metrics.ClearCohortAdmittedWorkloadsMetrics(cachedParent.Name)
 }
 
 func (c *Cache) AddLocalQueue(q *kueue.LocalQueue) error {
--- a/test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
+++ b/test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
@@ -28,7 +28,6 @@ import (
 
 	config "sigs.k8s.io/kueue/apis/config/v1beta2"
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
-	"sigs.k8s.io/kueue/pkg/metrics"
 	utiltestingapi "sigs.k8s.io/kueue/pkg/util/testing/v1beta2"
 	"sigs.k8s.io/kueue/pkg/workload"
 	"sigs.k8s.io/kueue/test/util"
@@ -73,8 +72,8 @@ var _ = ginkgo.Describe("Cohorts", func() {
 	}
 
 	var createWorkload = func(wl *kueue.Workload) *kueue.Workload {
-		wls[wl.Name] = wl
 		util.MustCreate(ctx, k8sClient, wl)
+		wls[wl.Name] = wl
 		return wl
 	}
 
@@ -125,7 +124,6 @@ var _ = ginkgo.Describe("Cohorts", func() {
 				util.ExpectObjectToBeDeleted(ctx, k8sClient, cq, true)
 			}
 			for _, cohort := range cohorts {
-				metrics.ClearCohortMetrics(kueue.CohortReference(cohort.Name))
 				util.ExpectObjectToBeDeleted(ctx, k8sClient, cohort, true)
 			}
 			util.ExpectObjectToBeDeleted(ctx, k8sClient, defaultFlavor, true)
@@ -458,7 +456,7 @@ var _ = ginkgo.Describe("Cohorts", func() {
 			})
 		})
 
-		ginkgo.XIt("correctly handles cohort metrics when workload admitted with admission check", func() {
+		ginkgo.It("correctly handles cohort metrics when workload admitted with admission check", func() {
 			const (
 				numWorkloadsForCQ1 = 5
 				numWorkloadsForCQ2 = 2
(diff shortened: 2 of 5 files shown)


test/integration/singlecluster/scheduler/fairsharing/fair_sharing_test.go
```diff
@@ -120,6 +120,7 @@ var _ = ginkgo.Describe("Scheduler", ginkgo.Label("feature:fairsharing"), func()
 		}
 		for _, cohort := range cohorts {
 			util.ExpectObjectToBeDeleted(ctx, k8sClient, cohort, true)
+			metrics.ClearCohortMetrics(kueue.CohortReference(cohort.Name))
```
mszadkow: the assumption is that we reuse the metrics registry?
I think we needed this line 👍

Makefile-test.mk
```diff
@@ -98,6 +98,7 @@ test: gotestsum ## Run tests.
 test-integration: compile-crd-manifests envtest ginkgo dep-crds kueuectl ginkgo-top ## Run integration tests for all singlecluster suites.
 	KUBEBUILDER_ASSETS="$(KUBEBUILDER_ASSETS)" \
 	PROJECT_DIR=$(PROJECT_DIR)/ \
+	GINKGO_ARGS=--repeat=200 \
```
mimowo: is this PR already fixing the issue? if so then we need to revert that

mbobrovskyi: Done

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
```diff
@@ -125,8 +125,8 @@ var _ = ginkgo.Describe("Cohorts", func() {
 				util.ExpectObjectToBeDeleted(ctx, k8sClient, cq, true)
 			}
 			for _, cohort := range cohorts {
-				metrics.ClearCohortMetrics(kueue.CohortReference(cohort.Name))
 				util.ExpectObjectToBeDeleted(ctx, k8sClient, cohort, true)
```
mwielgus, line 127: Maybe we should expand ExpectObjectToBeDeleted to clear up the metrics associated with the specific passed object? Right now it is easy to forget what needs to be cleared when.

mbobrovskyi: Done

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
```diff
@@ -628,6 +626,52 @@ var _ = ginkgo.Describe("Cohorts", func() {
...
+			ginkgo.By("Checking that only cq2 Workloads admitted", func() {
+				util.ExpectCohortSubtreeAdmittedWorkloadsTotalMetric("root", "", 5)
+				util.ExpectCohortSubtreeAdmittedWorkloadsTotalMetric("ch1", "", 5)
+			})
+
+			ginkgo.By("Deleting Cohorts except root", func() {
+				for _, cohort := range cohorts {
+					util.ExpectObjectToBeDeleted(ctx, k8sClient, cohort, true)
+					util.ExpectCohortSubtreeAdmittedWorkloadsTotalMetric("ch1", "", 0)
+					util.ExpectCohortSubtreeAdmittedWorkloadsTotalMetric("root", "", 5)
```
mbobrovskyi: Oh, I think I found the problem. If the root cohort has an object, we delete it and clear the metrics (https://github.com/kubernetes-sigs/kueue/blob/816af5a4b23203e9e860bfe713e3bdfbf5ef22a0/pkg/controller/core/cohort_controller.go#L187). However, if it doesn’t have one, the metrics are still not cleared.

To be honest, I’m not sure what the correct behavior should be. Maybe we should check whether there’s no reference and clear the metrics in that case.

@mimowo WDYT?

mimowo: IIUC we are discussing a case of childless Cohort - should we delete the metrics or keep? 

I'm not sure, bit it is a corner case with little practical impact, I would not complicate the code for it. 

I would rather make sure the test reflects realistic and important cases:
1. where a new CQ is added or deleted, then update the metrics for the cohort
2. when the childless cohort is removed then the metrics are gone too, so that we don't leak metrics

mimowo: if not much complication then probably clearing metrics for childless parent is ok, but then we have another question if the childless cohort has a parent, then should we clear for the parent, probably. but im a little bit afraid of complicatong code for such cases, so I think it is ok to keep the behavior underspecified for the corner cases as long as we dont leak metrics after cohort removal

mimowo: Thinking about the semantics, I think CohortSubtreeAdmittedWorkloadsTotalMetric metric should be cleared for cohorts which doesn't have any CQs in the subtree, because if there are no CQs, then also no workloads.
However, again, I'm not sure worth the complication, I would probably need to see code to assess.

Also, note that it is not universal rule for all metrics. Metric for nominal quota for cohorts should probably still report even if no CQs.

 So, I'm also ok to temporarily just adjust the test so that it does not delete the last CQ. I'm not concerned with the corner case for 0.17.0, but we may return to it for sure.

mbobrovskyi: > Thinking about the semantics, I think CohortSubtreeAdmittedWorkloadsTotalMetric metric should be cleared for cohorts which doesn't have any CQs in the subtree, because if there are no CQs, then also no workloads.
However, again, I'm not sure worth the complication, I would probably need to see code to assess.

Yeah, that’s probably the best approach. Or even better, we could check whether the Cohort has an object: if it does, keep the metrics; if it doesn’t and there are no CQs in the cohort, then clear the metrics.

mimowo: Something like this but in that scenario:

`root cohort -> sub cohort` // no CQs, just to  cohorts linked

we would clear the metric for admitted workloads for both "root cohort" and "sub cohort", right?

None of them can have any workloads admitted, because none of them have any CQ, either transitive or direct.

So in this case:

`root cohort -> sub cohort -> CQ1`

the metric is accounted both for   "root cohort" and "sub cohort".

Makefile-test.mk
```diff
@@ -107,6 +107,7 @@ test-integration: compile-crd-manifests envtest ginkgo dep-crds kueuectl ginkgo-
 
 .PHONY: test-integration-baseline
 test-integration-baseline: INTEGRATION_FILTERS= --label-filter="!slow && !redundant"
+test-integration-baseline: GINKGO_ARGS= --repeat=200
```
mbobrovskyi: /hold for test

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
```diff
@@ -458,7 +456,7 @@ var _ = ginkgo.Describe("Cohorts", func() {
 			})
 		})
 
-		ginkgo.XIt("correctly handles cohort metrics when workload admitted with admission check", func() {
+		ginkgo.It("correctly handles cohort metrics when workload admitted with admission check", func() {
```
mimowo, line 459: Have you been able to confirm this fix works by looping the test?

pkg/cache/scheduler/cohort.go
```diff
@@ -107,3 +107,15 @@ func (c *cohort) updateAdmittedWorkloadsCount(delta int) {
 		ancestor.admittedWorkloadsCount += delta
 	}
 }
+
+func (c *cohort) hasChildren() bool {
```
mimowo:
```suggestion
func (c *cohort) hasCQInSubtree() bool {
```
this is the intention. hasChildren I find misleading, because it says "false" if child Cohort exists.

pkg/cache/scheduler/cache.go
```diff
@@ -516,9 +516,16 @@ func (c *Cache) DeleteClusterQueue(cq *kueue.ClusterQueue) {
...
 
-	// Update cohort resources after deletion
 	if parent != nil {
+		// Update cohort resources after deletion.
 		updateCohortTreeResourcesIfNoCycle(parent)
+		// Clear metrics for the cohort and its ancestors if no children.
+		for ancestor := range parent.PathSelfToRoot() {
+			if ancestor.hasChildren() {
+				break
+			}
```
mimowo: This seems like quadratic time complexity with the depth of the hierarchy tree. I'm a bit concerned about that. Also, we might be calling ClearCohortAdmittedWorkloadsMetrics for the same

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
mbobrovskyi: Yes, we have loop https://github.com/kubernetes-sigs/kueue/pull/10080#discussion_r3009159215.

pkg/cache/scheduler/cache.go
mbobrovskyi: Added cache.

Makefile-test.mk
mszadkow: `--until-it-fails` is also good one ;)

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
```diff
@@ -649,6 +647,129 @@ var _ = ginkgo.Describe("Cohorts", func() {
 			})
 		})
 
+		ginkgo.It("should clear metrics for root Cohort after deletion child Cohort", func() {
```
mszadkow: the important mention here it is implicit cohort, so no API resource, simply cache object

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
```diff
@@ -125,7 +124,6 @@ var _ = ginkgo.Describe("Cohorts", func() {
 				util.ExpectObjectToBeDeleted(ctx, k8sClient, cq, true)
 			}
 			for _, cohort := range cohorts {
-				metrics.ClearCohortMetrics(kueue.CohortReference(cohort.Name))
```
mszadkow, line 128: no need to clear that before each test?

Makefile-test.mk
mbobrovskyi: Not exactly – it will run indefinitely.

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
mbobrovskyi: Now no need. We are clearing it after deletion.

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
mbobrovskyi: Done

pkg/cache/scheduler/cohort.go
```diff
@@ -107,3 +111,35 @@ func (c *cohort) updateAdmittedWorkloadsCount(delta int) {
...
+		if childCohort.hasCQInSubtree() {
+			return true
+		}
+	}
+	return false
+}
+
+// invalidateHasCQInSubtreeCache invalidates the cache for this cohort and all ancestors
+func (c *cohort) invalidateHasCQInSubtreeCache() {
+	if hierarchy.HasCycle(c) {
```
mszadkow: That's a lot of checks for HasCycle, can we reduce them?

mbobrovskyi:
> @mbobrovskyi what is the status of the PR?

Waiting for your review.

mbobrovskyi:
> Finally, @mbobrovskyi why is this on hold if ready for review?

Oh, it’s because I added changes for test (https://github.com/kubernetes-sigs/kueue/pull/10080#discussion_r3009159215). I marked it as “hold” so I wouldn’t forget to revert it. I’ve now reverted it (https://github.com/kubernetes-sigs/kueue/pull/10080/changes/52d6e9b3a2b96fdb28619da46482f16c3437177f) since all tests are passing.

/unhold

test/integration/singlecluster/scheduler/fairsharing/fair_sharing_test.go
```diff
@@ -1441,6 +1443,7 @@ var _ = ginkgo.Describe("Scheduler with AdmissionFairSharing = nil", ginkgo.Labe
...
```
mimowo: This looks like a useful fix to the tests not really related to the PR, should we submit a dedicated PR?

mimowo:
Also, could you explain how this fix relates to the disabled test "correctly handles cohort metrics when workload admitted with admission check"? I don't see this test removing a cohort, but maybe I'm missing something.

test/integration/singlecluster/scheduler/fairsharing/fair_sharing_test.go
mbobrovskyi: Reverted

mbobrovskyi:
> Do we now also clear the counter metric: kueue_cohort_subtree_admitted_workloads_total?
As a consistency check could you please check we also clear the kueue_cohort_subtree_admitted_workloads_total metric when the ClusterQueue is deleted (on the main branch) ? I'm feeling we don't necessarily need to clean counter metrics

Yes, I added this check on tests.

pkg/cache/scheduler/cache.go
```diff
@@ -183,6 +183,9 @@ func (c *Cache) newClusterQueue(log logr.Logger, cq *kueue.ClusterQueue) (*clust
 	}
 	c.hm.AddClusterQueue(cqImpl)
 	c.hm.UpdateClusterQueueEdge(kueue.ClusterQueueReference(cq.Name), cq.Spec.CohortName)
+	if cqImpl.HasParent() {
+		cqImpl.Parent().invalidateHasCQInSubtreeCache()
+	}
```
mimowo: This should be responsibility of `UpdateClusterQueueEdge`. I don't like the idea of the cache invalidation cache scattered across the codebase. I think this is not maintainable

pkg/cache/scheduler/cohort.go
```diff
@@ -33,6 +33,10 @@ type cohort struct {
 	FairWeight float64
 
 	admittedWorkloadsCount int
+
+	// hasCQInSubtreeCache caches the result of hasCQInSubtree to avoid O(N) traversal.
+	// nil = not cached, true = has CQ in subtree, false = no CQ in subtree
+	hasCQInSubtreeCache *bool
```
mimowo: Use synchronized types, either atomic.Pointer to bool, or encapsulate with R/W lock.

mimowo:
> On release-0.16 we clearing it with other metrics here https://github.com/kubernetes-sigs/kueue/blob/release-0.16/pkg/metrics/metrics.go#L759 which we are using in https://github.com/kubernetes-sigs/kueue/blob/release-0.16/pkg/cache/queue/metrics.go#L89-L91 and https://github.com/kubernetes-sigs/kueue/blob/release-0.16/pkg/cache/queue/manager.go#L371.
> It was added here https://github.com/kubernetes-sigs/kueue/pull/6882.

Interesting, because those are visualized as "deltas" rather than "raw". Most important is clearing guage, because those are visualized often in the "raw" form.

Anyway, I'm ok to clear counters too, especially if we already have that practice for long, and we have helper functions to that.

pkg/cache/scheduler/cohort.go
mbobrovskyi: Done

pkg/cache/scheduler/cache.go
mbobrovskyi: Yes, I agree. But in that case, we should extend the C interface.

mimowo: I'm not sure why this requires interface changes, the cache should be technical detail for the implementation of the interface. The same interface could be implemented without cache at all.

mbobrovskyi: I need to use `hasCQInSubtree()` and `invalidateHasCQInSubtreeCache()`, which are currently implemented on the `cohort`. If we move them to `hm`, we’ll need to extend the `C` interface so they can be used outside in the cache. That’s doable, but it would require some refactoring. We would also need to move the PathSelfToRoot() helper (and possibly more logic), or alternatively reimplement this functionality natively.

mimowo: > If we move them to hm, we’ll need to extend the C interface so they can be used outside in the cache.

How do you want / need to extend the interface? I think the `invalidateHasCQInSubtreeCache` should not be part of the interface

mbobrovskyi: We definitely need hasCQInSubtree() to check its value externally when clearing metrics. invalidateHasCQInSubtreeCache() is used internally, but we still need PathSelfToRoot() to traverse to the root and implement the CycleCheckable interface to detect cycles.

mimowo: I'm still confused why we would need `hasCQInSubtree` in the interface. I would assume this is all just implementation details for the metrics management. It seems I'm missing some implementation nuances, but not clear to me what.

mimowo: Let me look deeper into the code to think about alternative ways of implementing that.

mimowo: Sorry, I'm not sure I will find time to guide the implementation this week, but I would like to still re-consider the implementation here. 

I think this is anti-pattern if the caller needs to know the internal structure of the library, to invalidate the cache by an external call. Cache management and implementation should be "technical detail" which does not require leaking functions like invalidateHasCQInSubtreeCache to outside world.

I appreciate it may not be easy task given the current code structure, but I would like to at least see some effort in this direction.

mimowo: Actually I raised the code complication as a concern already in https://github.com/kubernetes-sigs/kueue/pull/10080#discussion_r2986560162. I would rather: adjust the existing tests so that they don't remove the last CQ. Open an issue to track this bug as a low priority item.  Also make sure there is a test which shows then when such a Cohort (without CQ) is deleted then the metrics are deleted too. In other words the metrics don't leak.

So, I admit there is a bug, but I don't think we need to rush with complex and tricky implementation to cover this case. Cohorts left without any CQs are really cluster misconfigs. As long as we make sure the metrics don't leak (see above) I think it is ok for now.

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
```diff
@@ -649,6 +647,168 @@ var _ = ginkgo.Describe("Cohorts", func() {
...
+			})
+
+			ginkgo.By("Checking that metrics cleared", func() {
+				util.ExpectAdmittedWorkloadsTotalMetric(cq1, "", 5)
+				util.ExpectAdmittedActiveWorkloadsGaugeMetric("cq1", 5)
+				util.ExpectCohortSubtreeAdmittedWorkloadsTotalMetric("ch1", "", 0)
+				util.ExpectCohortSubtreeAdmittedWorkloadsTotalMetric("root", "", 0)
+				util.ExpectCohortSubtreeAdmittedActiveWorkloadsGaugeMetric("root", 0)
+				util.ExpectCohortSubtreeAdmittedActiveWorkloadsGaugeMetric("ch1", 0)
+			})
```
mimowo, line 706: Could you add a comment or update PR description to indicate which assert specifically fails on the "main" branch, which is indicative of the leak for memory for metrics

mbobrovskyi: This one:

https://github.com/kubernetes-sigs/kueue/blob/2b07fac8fd8f4916bfc1d31161ba068a88029f11/test/integration/singlecluster/scheduler/fairsharing/metrics_test.go#L461-L570

mimowo: but as we sync on slack IIUC this test would fail if the changes are reverted, I'm trying to figure out which assert in this test "should clear metrics for an implicit root Cohort after deleting a child Cohort" would fail specifically

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
```diff
@@ -649,6 +647,230 @@ var _ = ginkgo.Describe("Cohorts", func() {
...
+				}
+			})
+
+			ginkgo.By("Checking that metrics cleared", func() {
+				util.ExpectAdmittedWorkloadsTotalMetric(cq1, "", 0)
+				util.ExpectAdmittedActiveWorkloadsGaugeMetric("cq1", 0)
+			})
+		})
+
+		ginkgo.It("shouldn't clear metrics for an implicit root Cohort after deleting the ClusterQueue if it has children Cohorts", func() {
```
mimowo: I'm trying to understand the intention of the tests, this one is particularly confusing "shouldn't clear metrics for an implicit root Cohort after deleting the ClusterQueue if it has children Cohorts". 

This PR adds new three test cases which looks ok-ish, but I would rather rephrase them:

1. "should clear metrics for the implicit root Cohort after deleting ClusterQueues then Cohort"
2. "should clear metrics for the implicit root Cohort after deleting Cohort then ClusterQueues"
3. "should clear metrics for the implicit root Cohort after deleting ClusterQueues"

In (2.) the results of https://github.com/kubernetes-sigs/kueue/pull/10080/changes#diff-119d4450e4fb8bfe8786ea5c0ff81ee57e31eea554356d79a7a16dd4e1e83d9fR804-R809 aren't that relevant. I would rather check what happens after clearing ch1.

mbobrovskyi: Done

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
mbobrovskyi: https://github.com/kubernetes-sigs/kueue/pull/10354

mimowo:

unrelated

pkg/cache/scheduler/cache.go
```diff
@@ -530,18 +534,31 @@ func (c *Cache) AddOrUpdateCohort(apiCohort *kueue.Cohort) error {
...
 	if cohort := c.hm.Cohort(cohortName); cohort != nil {
 		cohort.updateAdmittedWorkloadsCount(-cohort.admittedWorkloadsCount)
+		metrics.ClearCohortAdmittedWorkloadsMetrics(cohort.Name)
+		parent = cohort.Parent()
 	}
 
 	c.hm.DeleteCohort(cohortName)
+	if parent != nil {
+		c.clearMetricsForImplicitCohort(parent.Name)
+	}
```
mimowo: Why don't do that from within `if cohort := c.hm.Cohort(cohortName); cohort != nil {` - it could save some lines

mbobrovskyi: We need to clear the old parent. After calling c.hm.DeleteCohort(cohortName), we could lose the reference to it.

pkg/cache/scheduler/cache.go
```diff
@@ -551,6 +568,15 @@ func (c *Cache) DeleteCohort(cohortName kueue.CohortReference) {
...
 }
 
+func (c *Cache) clearMetricsForImplicitCohort(parentName kueue.CohortReference) {
+	parent := c.hm.Cohort(parentName)
+	if parent != nil && (parent.HasParent() || parent.IsExplicit() ||
+		len(parent.ChildCohorts()) > 0 || len(parent.ChildCQs()) > 0) {
+		return
+	}
+	metrics.ClearCohortAdmittedWorkloadsMetrics(parentName)
+}
```
mimowo, line 578: I think this name is misleading, it does not necessarily clear the metrics. Also implic cohort seems like a very technical detail.



Could we generalize the function to say:
`handleParentChange(oldParent, newParent)`. For now it would just clear the metrics, but it would allow to save lines for example in UpdateClusterQueue: from
```golang
	if oldParent != nil && oldParent != cqImpl.Parent() {
		c.clearMetricsForImplicitCohort(oldParent.Name)
	}
```
to just `c.handleparentChange(oldParent, cqImpl.Parent())`. In case of parent deletion we could just set nil as the newParent. I think it also reflects the intention much better, because it does not say "clear" as the clear may not need to happen (as mentioned above)

pkg/cache/scheduler/cache.go
mbobrovskyi: And also we need to clear it after c.hm.DeleteCohort(cohortName).

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
```diff
@@ -649,6 +647,257 @@ var _ = ginkgo.Describe("Cohorts", func() {
...
+				}
+			})
+
+			ginkgo.By("Deleting LocalQueues", func() {
+				for _, lq := range lqs {
+					util.ExpectObjectToBeDeleted(ctx, k8sClient, lq, true)
+				}
+			})
+
+			gomega.Expect(util.DeleteNamespace(ctx, k8sClient, ns)).To(gomega.Succeed())
```
mimowo: Why do we do that during the test, it seems completely irrelevant

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
```diff
@@ -649,6 +647,257 @@ var _ = ginkgo.Describe("Cohorts", func() {
...
```
mimowo: Similarly, I think we can drop this code

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
```diff
@@ -649,6 +647,257 @@ var _ = ginkgo.Describe("Cohorts", func() {
...
```
mimowo: Again, let's drop, it should be AfterEach that cleans this up

mimowo (commented):
LGTM overall, thank you 👍

mimowo:
Let's squash commits

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
mbobrovskyi: Done

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
mbobrovskyi: Done

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
mbobrovskyi: Done

pkg/cache/scheduler/cache.go
mbobrovskyi: I moved the logic for fetching `newParent` inside the function.

pkg/cache/scheduler/cache.go
mimowo: > And also we need to clear it after c.hm.DeleteCohort(cohortName).

I dont see why, could you explain? Also do we have a test which would fail if we call in different order?

pkg/cache/scheduler/cache.go
```diff
@@ -551,6 +564,16 @@ func (c *Cache) DeleteCohort(cohortName kueue.CohortReference) {
 	}
 }
 
+func (c *Cache) handleParentChange(oldParent *cohort) {
+	if oldParent == nil {
+		return
+	}
+	if newParent := c.hm.Cohort(oldParent.Name); newParent == nil ||
+		(!newParent.HasParent() && !newParent.IsExplicit() && len(newParent.ChildCohorts()) == 0 && len(newParent.ChildCQs()) == 0) {
```
mimowo: This condition is super long, could you instead check that inside and escape with return?

pkg/cache/scheduler/cache.go
mbobrovskyi: Only after deleting the Cohort can we determine whether the parent has any child ClusterQueues or child Cohorts remaining.

mimowo: This sounds reasonable, let's ensure we have a test which would pick that, iiuc when we switch the cohort's parent from one root to another

pkg/cache/scheduler/cache.go
mimowo: sgtm

pkg/cache/scheduler/cache.go
```diff
@@ -530,18 +532,29 @@ func (c *Cache) AddOrUpdateCohort(apiCohort *kueue.Cohort) error {
...
 	if cohort := c.hm.Cohort(cohortName); cohort != nil {
 		cohort.updateAdmittedWorkloadsCount(-cohort.admittedWorkloadsCount)
+		metrics.ClearCohortAdmittedWorkloadsMetrics(cohort.Name)
+		parent = cohort.Parent()
 	}
 
 	c.hm.DeleteCohort(cohortName)
+	if parent != nil {
+		c.handleParentChange(parent)
+	}
```
mimowo: The check for nillness is already inside handleParentChange

mbobrovskyi: Good catch! Thank you!

pkg/cache/scheduler/cache.go
mbobrovskyi: Done

pkg/cache/scheduler/cache.go
```diff
@@ -551,6 +564,21 @@ func (c *Cache) DeleteCohort(cohortName kueue.CohortReference) {
...
+	if oldParent == nil {
+		return
+	}
+	if newParent := c.hm.Cohort(oldParent.Name); newParent != nil {
+		if newParent.HasParent() || newParent.IsExplicit() {
+			return
+		}
+		if len(newParent.ChildCohorts()) > 0 || len(newParent.ChildCQs()) > 0 {
+			return
+		}
```
mimowo: Hm, I'm not sure why the cleanup of metrics for `oldParent` depends on the fact if the "newParent" is Explicit or not, or the newParent children. I think we should rather just check if the old parent has children or not. I'm a bit confused here.

pkg/cache/scheduler/cache.go
mbobrovskyi: I moved this logic and discovered an error in the test case "Cohorts when creating, modifying and removing should clear metrics for the implicit root Cohort after deleting a Cohort, then ClusterQueues." So this case is now covered.

pkg/cache/scheduler/cache.go
mimowo: I actually sense a bug here in the scenario when we have:
CQ1  -> Cohort X (explicit) -> Cohort "root1" (implicit)
CQ2 -> Cohort "root2"

now we switch the parent for Cohort X from Cohort "root1" to Cohort "root2" which is "explicit, then we will not clear the metrics, but IIUC we should.

mbobrovskyi: > Hm, I'm not sure why the cleanup of metrics for oldParent depends on the fact if the "newParent" is Explicit or not, or the newParent children. I think we should rather just check if the old parent has children or not. I'm a bit confused here.

We don’t need to cover explicit Cohorts. For explicit Cohorts, we clean up metrics when the Cohort is deleted. The issue only affects implicit Cohorts.

pkg/cache/scheduler/cache.go
```diff
@@ -551,6 +562,21 @@ func (c *Cache) DeleteCohort(cohortName kueue.CohortReference) {
 	}
 }
 
+func (c *Cache) handleParentChange(oldParent *cohort) {
```
mimowo: I think we should rather change to `handleParentUpdate`, because change may suggest we are changing from "root1" to "root2". Here We are only checking the updated value for the parent.

mimowo:
```suggestion
func (c *Cache) handleParentChange(cachedParent *cohort) {
```
Just to reflect his one is cached version, "old" might be confused with the "previous" value for spec.CohortName - that was my mistake reading the code at first

pkg/cache/scheduler/cache.go
```diff
@@ -551,6 +562,21 @@ func (c *Cache) DeleteCohort(cohortName kueue.CohortReference) {
...
+	if oldParent == nil {
+		return
+	}
+	if newParent := c.hm.Cohort(oldParent.Name); newParent != nil {
```
mimowo:
```suggestion
	if updatedParent := c.hm.Cohort(oldParent.Name); newParent != nil {
```
To make it more clear this is not the new value for "spec.CohortName"

pkg/cache/scheduler/cache.go
```diff
@@ -551,6 +562,21 @@ func (c *Cache) DeleteCohort(cohortName kueue.CohortReference) {
...
+		if newParent.HasParent() || newParent.IsExplicit() {
```
mimowo: I think we can simplify here:

if `newParent.HasParent() ` then it implies `newParent.IsExplicit()`, so we only need to check  `updatedParent.IsExplicit()`. Also this check we could do using `cachedParent` actualy before calling  c.hm.Cohort( becuase that would not change

pkg/cache/scheduler/cache.go
mimowo: Ok, I think this is ok, I was just confused by the nameing thinking that 
oldParent=root1
newParent=root2
based on the spec.CohortName.

We synced on slack, and now I understand both oldParent and newParent point to "root1" ,  just different momemnts in time. I proposed some renames to make it more clear in the follow up comments.

pkg/cache/scheduler/cache.go
mbobrovskyi: Done

mbobrovskyi: Done

pkg/cache/scheduler/cache.go
mbobrovskyi: Done

pkg/cache/scheduler/cache.go
mbobrovskyi: Done

mimowo:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10080/pull-kueue-test-integration-extended-main/2042568960041291776
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10080/pull-kueue-test-integration-baseline-main/2042568959944822784

ptal, the first is likely related, the other may be just a flake, if so then open another Issue

mbobrovskyi:
The https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10080/pull-kueue-test-integration-extended-main/2042568960041291776 should be fixed by https://github.com/kubernetes-sigs/kueue/pull/10381.

mbobrovskyi:
> https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10080/pull-kueue-test-integration-baseline-main/2042568959944822784

Known issue https://github.com/kubernetes-sigs/kueue/issues/10321.

mimowo:
> The https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10080/pull-kueue-test-integration-extended-main/2042568960041291776 should be fixed by #10381.

Great
/test pull-kueue-test-integration-baseline-main
/test pull-kueue-test-integration-extended-main

mimowo:
Hmm this seems to fail often could you double check by looping the PR locally 50 times on the cohort metrics tests?

mbobrovskyi:
> Hmm this seems to fail often could you double check by looping the PR locally 50 times on the cohort metrics tests?

Yes, it's because https://github.com/kubernetes-sigs/kueue/pull/10381 just now merged. Let's try again.

/test pull-kueue-test-integration-baseline-main
/test pull-kueue-test-integration-extended-main

Hmm. I have only one question why baseline and extended running the same tests?

mimowo:
yeah I observed the same on another PR, maybe some CI bug in Makefile. Could you open issue?

mbobrovskyi:
> yeah I observed the same on another PR, maybe some CI bug in Makefile. Could you open issue?

Created PR https://github.com/kubernetes-sigs/kueue/pull/10389

Issue #9779, opened 2026-03-10
Visibility server returns 500 Internal Server Error when lacking SubjectAccessReview permissions


Nilsachy:
**What happened**:

When the Kueue visibility server is configured with a custom kubeconfig and the associated identity lacks the `system:auth-delegator` ClusterRole (which grants permission to create `SubjectAccessReview` resources), requests to visibility endpoints result in an `Internal Server Error (500)`.

The error message emitted is: 

```
"Internal Server Error: \"/apis/visibility.kueue.x-k8s.io/v1beta2/clusterqueues/test-kubeconfig-cq/pendingworkloads\": subjectaccessreviews.authorization.k8s.io is forbidden: User \"system:serviceaccount:kueue-system:visibility-test-sa\" cannot create resource \"subjectaccessreviews\" in API group \"authorization.k8s.io\" at the cluster scope"
```

**What you expected to happen**:

The visibility server should handle the case where it cannot perform delegated authorization and return a `Forbidden (403)` or `Unauthorized (401)` error to the client, rather than a `500 Internal Server Error`.

**How to reproduce it (as minimally and precisely as possible)**:

1. Start Kueue with the visibility server enabled.
2. Configure the visibility server to use a custom ServiceAccount via a kubeconfig.
3. Ensure the custom ServiceAccount does not have the `system:auth-delegator` ClusterRole.
4. Attempt to access a visibility API endpoint (e.g., `/apis/visibility.kueue.x-k8s.io/v1beta2/clusterqueues/<name>/pendingworkloads`).
5. Observe the 500 error response.

**Anything else we need to know?**:

This behavior was identified during E2E testing of the RBAC identity delegation in the visibility server (see `test/e2e/customconfigs/visibility_kubeconfig_test.go`). Ideally, the authorization delegation logic should be configured to map these "forbidden to check" errors to a 403 response.

PR #10081, opened 2026-03-24
test/e2e: expect internal error for visibility delegated authz failure

MatteoFari:
Tighten the e2e to isolate delegated authorization failure from delegated authentication.

#### What type of PR is this?
/kind cleanup
/area testing

This PR keeps the default apiserver behavior for delegated authorization failures and tightens the existing custom kubeconfig e2e.

The negative case now isolates delegated authorization from delegated authentication and expects `InternalError`, which matches the server-side failure we see in practice when the visibility server cannot create `SubjectAccessReview` objects.

Related to #9779

```release-note
NONE
```

### DIFF
--- a/test/e2e/customconfigs/visibility_server_test.go
+++ b/test/e2e/customconfigs/visibility_server_test.go
@@ -52,6 +52,8 @@ const (
 	customSAMountPath         = "/etc/custom-sa"
 	cqName                    = "test-kubeconfig-cq"
 	customVisibilityPort      = 9444
+	tokenReviewerRoleName     = "visibility-test-token-reviewer"
+	tokenReviewerBindingName  = "visibility-test-token-reviewer"
 )
 
 var _ = ginkgo.Describe("Visibility Server", func() {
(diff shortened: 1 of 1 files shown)


pkg/visibility/server.go
```diff
@@ -168,6 +173,44 @@ func newVisibilityServerConfig(kubeConfig *rest.Config) *genericapiserver.Recomm
...
+	return authorizer.AuthorizerFunc(func(ctx context.Context, attrs authorizer.Attributes) (authorizer.Decision, string, error) {
+		decision, reason, err := base.Authorize(ctx, attrs)
+		if decision == authorizer.DecisionAllow || err == nil {
+			return decision, reason, err
+		}
+		if containsForbiddenAuthorizationError(err) {
+			if reason == "" {
+				reason = err.Error()
+			}
+			return authorizer.DecisionDeny, reason, nil
```
mwielgus: Why error is set to nil here?

pkg/visibility/server_test.go
```diff
@@ -0,0 +1,116 @@
...
+	internalErr := errors.New("internal error")
+
+	tests := map[string]struct {
+		baseDecision authorizer.Decision
+		baseReason   string
+		baseErr      error
+		wantDecision authorizer.Decision
+		wantReason   string
+		wantErr      error
+	}{
```
mwielgus: Add a test case for an Aggregate error that contains both a Forbidden error and an InternalServer error to verify that the "Deny" logic correctly prioritizes the security failure.

test/e2e/customconfigs/visibility_server_test.go
```diff
@@ -155,6 +157,30 @@ var _ = ginkgo.Describe("Visibility Server", func() {
...
+				APIGroups: []string{"authentication.k8s.io"},
+				Resources: []string{"tokenreviews"},
+				Verbs:     []string{"create"},
+			}},
+		}
+		util.MustCreate(ctx, k8sClient, tokenReviewerRole)
+
+		tokenReviewerBinding := &rbacv1.ClusterRoleBinding{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:   "visibility-test-token-reviewer",
```
mwielgus: Use the constant.

Issue #6929, opened 2025-09-20
Workload scheduling failure when preemption (reclaimWithinCohort) is required and cluster queue has NoFit workload


ichekrygin:
**What happened**:
When a NoFit workload exists in a cohort, other pending workloads that should be schedulable via preemption are never admitted. Instead, periodic preemption and rescheduling cycles occur without progress.

Job's workload is not admitted.

**What you expected to happen**:
Job's workload is admitted.

**How to reproduce it (as minimally and precisely as possible)**:

### Setup

**Cohort**: `demo`

**ClusterQueues**

*CQ1*

```yaml
cohort: demo
reclaimWithinCohort: Any
resourceGroups:
- coveredResources: [cpu]
  flavors:
  - name: default-flavor
    resources:
    - name: cpu
      nominalQuota: 3
      borrowingLimit: 0  # nothing borrowed
```

*CQ2*

```yaml
cohort: demo
reclaimWithinCohort: Any
resourceGroups:
- coveredResources: [cpu]
  flavors:
  - name: default-flavor
    resources:
    - name: cpu
      nominalQuota: 0
      borrowingLimit: 3  # everything is borrowed
```

**LocalQueues**: `lq1` (CQ1), `lq2` (CQ2)

**Workloads**

* `job-1`: queue=`lq2`, request `cpu=1`
* `job-2`: queue=`lq2`, request `cpu=1`
* `job-3`: queue=`lq1`, request `cpu=4` *(unsatisfiable, >3 capacity)*
* `job-4`: queue=`lq1`, request `cpu=3`

### Repro Steps

Submit jobs in this order:

1. `job-1` → admitted
2. `job-2` → admitted
3. `job-3` → pending *(insufficient quota, 4 > 3)*
4. `job-4` → pending *(insufficient unused quota, needs 2 more CPU)*

At this point, periodic preemption cycles repeatedly evict `job-1` and `job-2`. However, `job-4` is **never admitted**.

#### Workaround

Deleting `job-3` (the `NoFit` workload) restores expected behavior:

* `job-1` and `job-2` are preempted
* `job-4` is admitted

### No Repro Cases

**Case 1: Submit `job-3` then `job-4`**

* `job-3`: pending (expected)
* `job-4`: admitted (expected)

**Case 2: Submit `job-1`, `job-2`, `job-4`**

* `job-1`: admitted
* `job-2`: admitted
* `job-4`: admitted, with `job-1` and `job-2` preempted (expected)

[...]

PR #10082, opened 2026-03-24
Prioritize workload priority over sticky workload status

### LINKED ISSUE #7301: Higher-priority workloads are blocked while a “sticky” low-priority workload is pinned to the head of the queue
**Related issues and PRs:**

* Root cause discussion: [#6929](https://github.com/kubernetes-sigs/kueue/issues/6929)
* Sticky workload introduction: [#7157](https://github.com/kubernetes-sigs/kueue/pull/7157)

### Summary

After the “sticky pending workload” fix introduced in [#7157](https://github.com/kubernetes-sigs/kueue/pull/7157), higher-priority workloads may become blocked behind a lower-priority workload that is pinned to the head of the queue.

The intent of the sticky behavior was to prevent starvation by keeping the workload that initiated preemption at the queue head until its preemption sequence completes. However, this also prevents newly added, higher-priority workloads from being scheduled while the sticky workload is still waiting for preemption to finish.

### Setup

The setup is similar to [#6929](https://github.com/kubernetes-sigs/kueue/issues/6929).

**ClusterQueues:**

* `cq1`:

    * nominal CPU = 3
    * borrowingLimit = 0
    * preemption:

      ```yaml
      preemption:
        reclaimWithinCohort: Any
        withinClusterQueue: LowerPriority
      ```
* `cq2`: nominal CPU = 0, borrowingLimit = 3

**LocalQueues:**

* `lq1` and `lq2`, corresponding to `cq1` and `cq2`

**PriorityClasses:**

* `high` = 10000
* `low` = 1000

**Jobs:**

| Job | CPU | Priority  | ClusterQueue |
| --- | --- | --------- | ------------ |
| j01 | 2   | (default) | cq2          |
| j02 | 4   | low       | cq1          |
| j03 | 2   | low       | cq1         |
| j04 | 2   | high      | cq1          |

---

### Repro Steps

1. Submit `j01` → **admitted**
2. Submit `j02` → **pending** (exceeds available CPU)
3. Submit `j03`, then `j04` in order

**Observed behavior:**

* `j03` enters scheduling cycle first and triggers preemption of `j01`
* Due to sticky logic, `j03` is pinned at the head of the queue until preemption completes
* Meanwhile, `j04` (higher priority) cannot proceed
* After several scheduling cycles `j03` is eventually admitted
* When `j04` runs next, it preempts `j03`

### Expected behavior

When a higher-priority workload (`j04`) enters the queue, it should preempt or take precedence over lower-priority workloads, even if one of them is currently “sticky.” The sticky behavior should not override priority-based scheduling order.

### Impact

This can lead to:

* Unintended scheduling delays for higher-priority workloads
* Inversion of expected priority ordering
* Reduced responsiveness of the scheduler under preemption-heavy workloads
* **Unordered scheduling may result in preemption churn**, as illustrated in the example where `j03` blocks `j04`, with eventual `j04` preemption of `j03`

j-skiba:
#### What type of PR is this?
/kind bug

This PR modifies the workload ordering logic in the ClusterQueue to ensure that workload priority takes precedence over sticky status.

Fixes #7301

```release-note
Fixed an issue in ClusterQueue workload ordering where workloads performing preemption could incorrectly block higher-priority workloads from being admitted. Priority now takes precedence over preemption status in the scheduling queue.
```

### DIFF
--- a/pkg/cache/queue/cluster_queue.go
+++ b/pkg/cache/queue/cluster_queue.go
@@ -732,6 +732,13 @@ func (c *ClusterQueue) RequeueIfNotPresent(ctx context.Context, wInfo *workload.
 // baseCompareFunc orders workloads by sticky status, priority, timestamp, and UID.
 func baseCompareFunc(log logr.Logger, wo workload.Ordering, sw *stickyWorkload) func(a, b *workload.Info) int {
 	return func(a, b *workload.Info) int {
+		p1 := utilpriority.EffectivePriority(log, a.Obj)
+		p2 := utilpriority.EffectivePriority(log, b.Obj)
+		// Higher priority comes first (reverse order).
+		if cmpResult := cmp.Compare(p2, p1); cmpResult != 0 {
+			return cmpResult
+		}
+
 		aSticky := sw.matches(workload.Key(a.Obj))
 		bSticky := sw.matches(workload.Key(b.Obj))
 		if aSticky != bSticky {
@@ -741,13 +748,6 @@ func baseCompareFunc(log logr.Logger, wo workload.Ordering, sw *stickyWorkload) 
 			}
 			logStickyWorkloadSelectionIfVerbose(log, b.Obj)
 			return 1
-		}
-
-		p1 := utilpriority.EffectivePriority(log, a.Obj)
-		p2 := utilpriority.EffectivePriority(log, b.Obj)
-		// Higher priority comes first (reverse order).
-		if cmpResult := cmp.Compare(p2, p1); cmpResult != 0 {
-			return cmpResult
 		}
 
 		tA := wo.GetQueueOrderTimestamp(a.Obj)
--- a/test/integration/singlecluster/scheduler/fairsharing/fair_sharing_test.go
+++ b/test/integration/singlecluster/scheduler/fairsharing/fair_sharing_test.go
@@ -863,6 +863,36 @@ var _ = ginkgo.Describe("Scheduler", ginkgo.Label("feature:fairsharing"), func()
 			util.ExpectClusterQueueWeightedShareMetric(cq2, 0.0)
 		})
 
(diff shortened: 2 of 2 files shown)


mwielgus (commented):
https://github.com/kubernetes-sigs/kueue/issues/6929 was (in theory) fixed a while ago. What is the real world scenario that was not fixed then?

j-skiba:
I synced with @gabesaba to discuss the problem around inadmissible workloads and how we can handle that more efficiently.

Instead of relying on immediate retries or a generic "sticky" retry mechanism, we believe the best approach is to make the requeuing of inadmissible workloads conditional, triggered only by specific relevant events in the cluster.

We can categorize the triggering events into distinct types and map them directly to the workload's "requeue reason".
1. "QuotaReleasedEvent" - triggered when another workload releases its quota. This should requeue workloads that became inadmissible due to:
- nominated but failed final admission
- pending preemption
- generic inadmissible reasons
2. "PreemptedWorkloadEvent" - triggered specifically when a workload is evicted due to preemption. This should requeue workloads that became inadmissible due to any preemption related reasons.
3. "CapacityChangeEvent" - triggered by cluster capacity changes (e.g. ClusterQueue updates, ResourceFlavor changes or node updates). This should trigger a bulk requeue of all inadmissible workloads, as the scheduling landscape has changed.

By introducing this granular mapping, we avoid wasteful scheduling cycles and ensure workloads are only retried when a relevant cluster event has actually occurred to improve their chance of admission.

### Alternative approach considered

As an alternative, I was also thinking about introducing resource reservations for preemptors. In that model, when a workload triggers a preemption, we would temporarily reserve those resources specifically for that preemptor. This would prevent other incoming workloads from "stealing" the newly freed capacity before the preemptor actually gets admitted. However, the event-driven requeuing approach seems more cohesive and less disruptive to existing scheduling queue and cache architecture.

PR #10085, opened 2026-03-24
Add label to filter e2e tests for customconfigs

MaysaMacedo:
#### What type of PR is this?
/kind cleanup

This PR adds filters to the customconfigs e2e tests and also allow to only install the dependencies of each tests cases when are meant to run.

Fixes #

```release-note
NONE
```

### DIFF
--- a/hack/testing/e2e-common.sh
+++ b/hack/testing/e2e-common.sh
@@ -89,12 +89,12 @@ function e2e_crd_exists {
     kubectl ${kubectl_args[@]+"${kubectl_args[@]}"} get crd "${crd}" >/dev/null 2>&1
 }
 
-if [[ -n ${APPWRAPPER_VERSION:-} && ("$GINKGO_ARGS" =~ feature:appwrapper || ! "$GINKGO_ARGS" =~ "--label-filter") ]]; then
+if [[ -n ${APPWRAPPER_VERSION:-} && ("$GINKGO_ARGS" =~ feature:(appwrapper|managejobswithoutqueuename) || ! "$GINKGO_ARGS" =~ "--label-filter") ]]; then
     export APPWRAPPER_MANIFEST=${ROOT_DIR}/dep-crds/appwrapper/config/default
     export APPWRAPPER_IMAGE=quay.io/ibm/appwrapper:${APPWRAPPER_VERSION}
 fi
 
-if [[ -n ${JOBSET_VERSION:-} && ("$GINKGO_ARGS" =~ feature:(jobset|tas|trainjob) || ! "$GINKGO_ARGS" =~ "--label-filter") ]]; then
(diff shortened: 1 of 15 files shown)


test/e2e/customconfigs/suite_test.go
```diff
@@ -64,10 +64,14 @@ var _ = ginkgo.BeforeSuite(func() {
...
-	util.WaitForLeaderWorkerSetAvailability(ctx, k8sClient)
-	util.WaitForSparkOperatorAvailability(ctx, k8sClient)
+	if ginkgo.Label("feature:managejobswithoutqueuename").MatchesLabelFilter(ginkgo.GinkgoLabelFilter()) {
+		util.WaitForJobSetAvailability(ctx, k8sClient)
+		util.WaitForAppWrapperAvailability(ctx, k8sClient)
+		util.WaitForLeaderWorkerSetAvailability(ctx, k8sClient)
+	}
+	if ginkgo.Label("feature:spark").MatchesLabelFilter(ginkgo.GinkgoLabelFilter()) {
+		util.WaitForSparkOperatorAvailability(ctx, k8sClient)
+	}
```
j-skiba, line 74: similar to this https://github.com/kubernetes-sigs/kueue/pull/10123/changes#r2999320668. Shouldn't we check `area:customconfigs` there?

hack/testing/e2e-common.sh
```diff
@@ -89,12 +89,12 @@ function e2e_crd_exists {
     kubectl ${kubectl_args[@]+"${kubectl_args[@]}"} get crd "${crd}" >/dev/null 2>&1
 }
 
-if [[ -n ${APPWRAPPER_VERSION:-} && ("$GINKGO_ARGS" =~ feature:appwrapper || ! "$GINKGO_ARGS" =~ "--label-filter") ]]; then
+if [[ -n ${APPWRAPPER_VERSION:-} && ("$GINKGO_ARGS" =~ feature:(appwrapper|managejobswithoutqueuename) || ! "$GINKGO_ARGS" =~ "--label-filter") ]]; then
```
j-skiba, line 92: will the images get uploaded when only `area:customconfigs` label filter is used?

j-skiba: actually I'm wondering what's the use case of `area:customconfig`? If it's not used in any scripts and not used in [test/e2e/customconfigs/suite_test.go](https://github.com/kubernetes-sigs/kueue/pull/10085/files/97b3c57a5d7040c2002d78fbbca579c31842c240#diff-377a81b9875abab18f7fb6aca730e602231f184e9a1ec45111897073e4231ae3) for waiting for deployments then I'm not sure if it's necessary. Also these tests are run with `make test-e2e-customconfig` which already targets specific paths.

The same applies to https://github.com/kubernetes-sigs/kueue/pull/10123

MaysaMacedo: good point, I'm removing the area label.

PR #10086, opened 2026-03-25
WIP: Fix flaky pod group e2e test by bumping ExpectWorkloadToFinish timeout

PannagaRao:
#### What type of PR is this?
/kind bug
/kind flake

Fixes #

```release-note
NONE
```

### DIFF
--- a/test/e2e/singlecluster/pod_test.go
+++ b/test/e2e/singlecluster/pod_test.go
@@ -139,45 +139,47 @@ var _ = ginkgo.Describe("Pod groups", ginkgo.Label("area:singlecluster", "featur
 			})
 		})
 
-		ginkgo.It("Should only admit a complete group", func() {
-			group := podtesting.MakePod("group", ns.Name).
-				Image(util.GetAgnHostImage(), util.BehaviorExitFast).
-				Queue(lq.Name).
-				RequestAndLimit(corev1.ResourceCPU, "1").
-				MakeGroup(3)
+		for i := range 200 {
+			ginkgo.FIt(fmt.Sprintf("Should only admit a complete group %d", i), func() {
+				group := podtesting.MakePod("group", ns.Name).
+					Image(util.GetAgnHostImage(), util.BehaviorExitFast).
+					Queue(lq.Name).
+					RequestAndLimit(corev1.ResourceCPU, "1").
+					MakeGroup(3)
 
-			ginkgo.By("Incomplete group should not start", func() {
-				// Create incomplete group.
-				for _, p := range group[:2] {
-					util.MustCreate(ctx, k8sClient, p.DeepCopy())
-				}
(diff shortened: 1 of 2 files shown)


test/e2e/singlecluster/pod_test.go
```diff
@@ -139,45 +139,47 @@ var _ = ginkgo.Describe("Pod groups", ginkgo.Label("area:singlecluster", "featur
 			})
 		})
 
-		ginkgo.It("Should only admit a complete group", func() {
-			group := podtesting.MakePod("group", ns.Name).
-				Image(util.GetAgnHostImage(), util.BehaviorExitFast).
-				Queue(lq.Name).
-				RequestAndLimit(corev1.ResourceCPU, "1").
-				MakeGroup(3)
+		for i := range 200 {
```
mimowo, line 142: is it enough to repro the issue? I would exoect the issue is so rare and only when CI is loaded that it might be hard to catch this way anyway

test/util/util.go
```diff
@@ -482,7 +482,7 @@ func ExpectWorkloadToFinish(ctx context.Context, k8sClient client.Client, wlKey
 		var wl kueue.Workload
 		g.Expect(k8sClient.Get(ctx, wlKey, &wl)).To(gomega.Succeed())
 		g.Expect(wl.Status.Conditions).To(utiltesting.HaveConditionStatusTrue(kueue.WorkloadFinished), "it's finished")
-	}, MediumTimeout, Interval).Should(gomega.Succeed())
```
mimowo: This change may be affecting many tests which are not relying on pod termination. Maybe we could add an extra wait using LongTimeout just for the affected tests? Otherwise we may consider ExpectWorkloadToFinishWithTimeout

test/e2e/singlecluster/pod_test.go
mszadkow: I would rather use `--until-it-fails` flag
also increase log level for the PR

mszadkow: later when you get it fail, I think it worth to download logs locally for more convenient analysis:
`hack/testing/get-build-logs.sh`

mbobrovskyi: > I would rather use --until-it-fails flag

Oh, this will run indefinitely. Maybe --repeat=200 would be better?

PannagaRao: Agreed, --repeat=200 is cleaner. The loop is temporary and will be removed before merge — I don't expect a failure since the fix is already included.

test/util/util.go
PannagaRao: Good point, will create ExpectWorkloadToFinishWithTimeout and use it only for the affected test to avoid impacting other callers.

PR #10090, opened 2026-03-25
featureGates unit tests cleanup

### LINKED ISSUE #9957: featureGates unit tests cleanup
Use `featureGates map[featuregate.Feature]bool` as a standard way for enabling/disabling features in unit tests
```go
// ✅ GOOD
featureGates map[featuregate.Feature]bool

// ❌ BAD — separate booleans per gate
enablePriorityBoost           bool
admissionFairSharingEnabled   bool
```

**Why is this needed**:
To have a standard way for enabling/disabling features in unit tests

Ref: https://github.com/kubernetes-sigs/kueue/pull/9120#discussion_r2939419215

ivnovakov:
#### What type of PR is this?

/kind cleanup
/area testing

This PR refactors unit tests to use a standardized featureGates map[featuregate.Feature]bool instead of individual boolean fields per feature gate.
This approach simplifies test setup, improves readability, and ensures consistency across the codebase when enabling or disabling feature gates.

Fixes #9957

```release-note
NONE
```

### DIFF
--- a/pkg/cache/scheduler/tas_cache_test.go
+++ b/pkg/cache/scheduler/tas_cache_test.go
@@ -381,12 +381,12 @@ func TestFindTopologyAssignments(t *testing.T) {
 	}
 
 	cases := map[string]struct {
-		enableFeatureGates []featuregate.Feature
-		nodes              []corev1.Node
-		pods               []corev1.Pod
-		levels             []string
-		nodeLabels         map[string]string
-		podSets            []PodSetTestCase
+		featureGates map[featuregate.Feature]bool
+		nodes        []corev1.Node
+		pods         []corev1.Pod
+		levels       []string
+		nodeLabels   map[string]string
+		podSets      []PodSetTestCase
 	}{
 		"minimize the number of used racks before optimizing the number of nodes; BestFit": {
 			// Solution by optimizing the number of racks then nodes: [r3]: [x1,x6,x2,x4]
@@ -499,7 +499,6 @@ func TestFindTopologyAssignments(t *testing.T) {
 					},
 				},
 			}},
-			enableFeatureGates: []featuregate.Feature{},
 		},
 		"choose the node that can accommodate all Pods": {
 			//        b1
@@ -651,7 +650,7 @@ func TestFindTopologyAssignments(t *testing.T) {
 					},
 				},
 			}},
-			enableFeatureGates: []featuregate.Feature{features.TASProfileMixed},
+			featureGates: map[featuregate.Feature]bool{features.TASProfileMixed: true},
 		},
 		"block required; 4 pods fit into one host each; BestFit": {
 			nodes:  binaryTreesNodes,
@@ -742,7 +741,7 @@ func TestFindTopologyAssignments(t *testing.T) {
 					},
 				},
 			}},
-			enableFeatureGates: []featuregate.Feature{features.TASProfileMixed},
+			featureGates: map[featuregate.Feature]bool{features.TASProfileMixed: true},
 		},
 		"host preferred; single Pod fits in the host; BestFit; TASProfileMixed": {
 			nodes:  defaultNodes,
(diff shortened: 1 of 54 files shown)


pkg/controller/core/workload_controller_test.go
```diff
@@ -2793,8 +2810,9 @@ func TestReconcile(t *testing.T) {
 	for name, tc := range cases {
 		for _, enabled := range []bool{false, true} {
 			t.Run(fmt.Sprintf("%s WorkloadRequestUseMergePatch enabled: %t", name, enabled), func(t *testing.T) {
-				features.SetFeatureGateDuringTest(t, features.DynamicResourceAllocation, tc.enableDRAFeature)
-				features.SetFeatureGateDuringTest(t, features.MultiKueueOrchestratedPreemption, tc.enableMKOrchestratedPreemptionFeature)
+				for fg, enable := range tc.featureGates {
+					features.SetFeatureGateDuringTest(t, fg, enable)
+				}
```
vladikkuzn: NIT: this can also be extracted to helper:

```suggestion
features.SetFeatureGatesDuringTest(t, tc.featureGates)
```

pkg/controller/core/workload_controller_test.go
```diff
@@ -2793,8 +2810,9 @@ func TestReconcile(t *testing.T) {
...
 				features.SetFeatureGateDuringTest(t, features.WorkloadRequestUseMergePatch, enabled)
 				features.SetFeatureGateDuringTest(t, features.AdmissionGatedBy, true)
```
vladikkuzn, line 2815: Shouldn't this be moved to featureGates?

pkg/controller/core/workload_controller_test.go
ivnovakov: I think it's a right call. This will enforce map usage.

pkg/controller/core/workload_controller_test.go
ivnovakov: Yep, we could.

> `features.SetFeatureGateDuringTest(t, features.WorkloadRequestUseMergePatch, enabled)`

I left this one as because each test is run twice with feature on and off and It wouldn't make sense to create twice as much test cases because of this feature gate. I can suggest replacing this line with this.

> `tc.featureGates[features.WorkloadRequestUseMergePatch] = enabled`

> `features.SetFeatureGateDuringTest(t, features.AdmissionGatedBy, true)`

With this one we can alter each test case and add `features.AdmissionGatedBy` to `featureGates` field or use the same approach suggested above.

Let me know what do you think.

mimowo: > With this one we can alter each test case and add features.AdmissionGatedBy to featureGates field or use the same approach suggested above.

The proposal sgtm

pkg/controller/core/workload_controller_test.go
mimowo: sgtm

pkg/cache/scheduler/tas_cache_test.go
```diff
@@ -5949,9 +5949,7 @@ func TestFindTopologyAssignments(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 			ctx, log := utiltesting.ContextWithLog(t)
 			// TODO: remove after dropping the TAS profiles feature gates
```
mbobrovskyi: nit: do we need this?

pkg/controller/core/workload_controller_test.go
vladikkuzn: > I left this one as because each test is run twice with feature on and off

Ah, ok, looks good to me then

pkg/controller/core/workload_controller_test.go
ivnovakov: Added `features.SetFeatureGatesDuringTest` which accept `featureGates` map.

pkg/controller/core/workload_controller_test.go
ivnovakov: So, leaving it as is then.

pkg/cache/scheduler/tas_cache_test.go
ivnovakov: After discussion with @vladikkuzn, removed it because it's outdated.

PR #10096, opened 2026-03-25
[Cleanup] Use fresh gauge metric for fetch

mszadkow:
#### What type of PR is this?
/kind cleanup

Use fresh gauge metric for fetch instead of reusing the same one, which may have stale value.

```release-note
NONE
```


test/util/metrics.go
```diff
@@ -293,19 +328,22 @@ func expectHistogramMetric(metric *prometheus.HistogramVec, matcher gomegatypes.
 	}, Timeout, Interval).Should(gomega.Succeed())
 }
 
-func expectGaugeMetric(metric prometheus.Gauge, matcher gomegatypes.GomegaMatcher, msgAndArgs ...any) {
+func expectGaugeMetric(metricFn func() prometheus.Gauge, matcher gomegatypes.GomegaMatcher, msgAndArgs ...any) {
```
mykysha: Can we use a slice of the label values instead of the function here?

mszadkow: Updated

PR #10097, opened 2026-03-25
feat: clean local queue metrics based on label change if required

mykysha:
#### What type of PR is this?
/kind feature

Clean LocalQueue metrics on label change if the new labels are not included in the config.

Part of #8830

```release-note
Observability: clean local queue metrics if the labels are not matching with the configuration after update.
```

### DIFF
--- a/pkg/cache/scheduler/cache.go
+++ b/pkg/cache/scheduler/cache.go
@@ -588,7 +588,7 @@ func (c *Cache) GetLocalQueueLabels(cqName kueue.ClusterQueueReference, lqKey qu
 	if err != nil {
 		return nil, err
 	}
-	return lq.labels, nil
+	return lq.GetLabels(), nil
 }
 
 func (c *Cache) ClusterQueueUsesAdmissionFairSharing(cqName kueue.ClusterQueueReference) bool {
@@ -603,18 +603,7 @@ func (c *Cache) ClusterQueueUsesAdmissionFairSharing(cqName kueue.ClusterQueueRe
 
 func (c *Cache) UpdateLocalQueue(oldQ, newQ *kueue.LocalQueue) error {
 	if oldQ.Spec.ClusterQueue == newQ.Spec.ClusterQueue {
-		if features.Enabled(features.CustomMetricLabels) {
-			c.RLock()
-			defer c.RUnlock()
-			cq := c.hm.ClusterQueue(newQ.Spec.ClusterQueue)
-			if cq != nil {
-				if lq := cq.localQueues[queue.Key(newQ)]; lq != nil {
-					lq.Lock()
-					lq.customMetricLabelValues = c.customLabels.ExtractValues(newQ.Labels, newQ.Annotations)
-					lq.Unlock()
-				}
-			}
-		}
+		c.updateLqMetricLabels(newQ)
 		return nil
 	}
 	c.Lock()
@@ -632,6 +621,19 @@ func (c *Cache) UpdateLocalQueue(oldQ, newQ *kueue.LocalQueue) error {
 		return cq.addLocalQueue(newQ, customLabelValues)
 	}
 	return nil
+}
+
+func (c *Cache) updateLqMetricLabels(newLq *kueue.LocalQueue) {
+	cachedLq, err := c.GetCacheLocalQueue(newLq.Spec.ClusterQueue, queue.Key(newLq))
+	if err != nil {
+		return
+	}
+	cachedLq.Lock()
+	defer cachedLq.Unlock()
+	cachedLq.labels = newLq.GetLabels()
+	if features.Enabled(features.CustomMetricLabels) {
+		cachedLq.customMetricLabelValues = c.customLabels.ExtractValues(newLq.Labels, newLq.Annotations)
+	}
 }
 
 func (c *Cache) AddOrUpdateWorkload(log logr.Logger, w *kueue.Workload) bool {
@@ -1013,7 +1015,7 @@ func (c *Cache) ResyncGaugeMetrics() {
 			cq.reportResourceMetrics(c.fairSharingEnabled)
 		}
 		for _, lq := range cq.localQueues {
-			if c.lqMetrics.ShouldExposeLocalQueueMetrics(lq.labels) {
+			if lq.shouldExposeMetrics(c.lqMetrics) {
 				lq.reportActiveWorkloads(c.roleTracker)
 				lq.reportResourceMetrics(cq.resourceNode.Quotas, c.roleTracker)
 			}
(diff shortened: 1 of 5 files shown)


pkg/cache/scheduler/cache.go
```diff
@@ -603,16 +603,17 @@ func (c *Cache) ClusterQueueUsesAdmissionFairSharing(cqName kueue.ClusterQueueRe
...
-			cq := c.hm.ClusterQueue(newQ.Spec.ClusterQueue)
-			if cq != nil {
-				if lq := cq.localQueues[queue.Key(newQ)]; lq != nil {
-					lq.Lock()
+		c.RLock()
+		defer c.RUnlock()
+		cq := c.hm.ClusterQueue(newQ.Spec.ClusterQueue)
+		if cq != nil {
+			if lq := cq.localQueues[queue.Key(newQ)]; lq != nil {
+				lq.Lock()
```
mszadkow: maybe defer the unlock

mykysha: Sure, done

pkg/cache/scheduler/cache.go
```diff
@@ -603,15 +605,16 @@ func (c *Cache) ClusterQueueUsesAdmissionFairSharing(cqName kueue.ClusterQueueRe
...
+				defer lq.Unlock()
```
mimowo: What is the scope of the defer here actually? Maybe it is safer to extract that code to a helper function to make it super clear the lock is released when the function exists

pkg/cache/scheduler/cache.go
```diff
@@ -584,6 +584,8 @@ func (c *Cache) GetCacheLocalQueue(cqName kueue.ClusterQueueReference, lqKey que
 }
 
 func (c *Cache) GetLocalQueueLabels(cqName kueue.ClusterQueueReference, lqKey queue.LocalQueueReference) (map[string]string, error) {
+	c.RLock()
+	defer c.RUnlock()
```
mimowo: Why this lock was not needed before but is needed now? Maybe this was a bug that we didn't have it?

pkg/cache/scheduler/cache.go
mykysha: Yeah, I guess the scope of this check is too nested either way, will introduce a helper

pkg/cache/scheduler/cache.go
mykysha: Yes, a test caught the race here so I added the lock to that read as well

mykysha: Can extract to a separate fix PR if needed

pkg/cache/scheduler/cache.go
mykysha: Done

pkg/cache/scheduler/cache.go
```diff
@@ -634,6 +625,23 @@ func (c *Cache) UpdateLocalQueue(oldQ, newQ *kueue.LocalQueue) error {
...
+	cq := c.hm.ClusterQueue(newLq.Spec.ClusterQueue)
+	if cq != nil {
+		if cachedLq := cq.localQueues[queue.Key(newLq)]; cachedLq != nil {
+			cachedLq.Lock()
+			defer cachedLq.Unlock()
+			cachedLq.labels = newLq.GetLabels()
+			if features.Enabled(features.CustomMetricLabels) {
+				cachedLq.customMetricLabelValues = c.customLabels.ExtractValues(newLq.Labels, newLq.Annotations)
+			}
```
mbobrovskyi: Oh, this logic is deeply nested. Could we split it into multiple if statements?

pkg/cache/scheduler/cache.go
```diff
@@ -634,6 +625,23 @@ func (c *Cache) UpdateLocalQueue(oldQ, newQ *kueue.LocalQueue) error {
...
+		}
+	}
```
```suggestion
	if cq == nil {
		return nil
	}

	lq := cq.localQueues[queue.Key(newLq)]
	if lq == nil {
		return nil
	}

	lq.Lock()
	defer lq.Unlock()

	lq.labels = newLq.GetLabels()
	if features.Enabled(features.CustomMetricLabels) {
		lq.customMetricLabelValues = c.customLabels.ExtractValues(newLq.Labels, newLq.Annotations)
	}
```

test/integration/singlecluster/controller/core/localqueue_controller_test.go
```diff
@@ -690,4 +690,47 @@ var _ = ginkgo.Describe("Queue controller metrics filtering", ginkgo.Label("cont
...
+				updated.Labels = make(map[string]string)
+			}
+			updated.Labels["metrics-test"] = "false"
+			g.Expect(k8sClient.Update(ctx, &updated)).To(gomega.Succeed())
+		}, util.Timeout, util.Interval).Should(gomega.Succeed())
+
+		ginkgo.By("Verifying metrics are cleared after labels are updated to non-matching")
+		gomega.Eventually(func(g gomega.Gomega) {
+			util.ExpectLQPendingWorkloadsMetric(queueDeleteMetrics, 0, 0)
+		}, util.LongTimeout, util.Interval).Should(gomega.Succeed())
```
mbobrovskyi:
```suggestion
		}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
```

pkg/cache/scheduler/cache.go
mykysha: Done

test/integration/singlecluster/controller/core/localqueue_controller_test.go
mykysha: I think that the regular timeout would be sufficient here

mbobrovskyi: Are you sure it's not enough 45s?

mykysha: I mean just the util.Timeout is good, no need for the Medium or Long one

pkg/cache/scheduler/cache.go
```diff
@@ -634,6 +625,25 @@ func (c *Cache) UpdateLocalQueue(oldQ, newQ *kueue.LocalQueue) error {
 	return nil
 }
 
+func (c *Cache) updateLqMetricLabels(newLq *kueue.LocalQueue) {
+	c.RLock()
+	defer c.RUnlock()
+	cq := c.hm.ClusterQueue(newLq.Spec.ClusterQueue)
+	if cq == nil {
+		return
+	}
```
mimowo: The lock on c seems to have a longer lock than needed. I would rather have two helpers: getClusterQueueByName() which takes read lock for c, and getLocalQueueByName which takes the logkc on ClusterQueue. Then we perform the update under lock for cachedLq as currently

mykysha: We have `GetCacheLocalQueue` which seems to fit this usecase perfectly, have updated the PR to use it instead and then lock on the lq for the update

PR #10120, opened 2026-03-26
Wait for the node to be ready in FailureRecoveryPolicy e2e test

kshalot:
#### What type of PR is this?
/kind cleanup
/kind flake

The test is flaking. I ran it in a loop and it did not fail, but I'm not 100% convinced this is going to fix it, but this will give more visibility into what is actually flaking in prow.

Fixes #

```release-note
NONE
```

### DIFF
--- a/test/e2e/customconfigs/failure_recovery_policy_test.go
+++ b/test/e2e/customconfigs/failure_recovery_policy_test.go
@@ -32,7 +32,6 @@ import (
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
 	"sigs.k8s.io/kueue/pkg/controller/constants"
 	"sigs.k8s.io/kueue/pkg/features"
-	"sigs.k8s.io/kueue/pkg/util/tas"
 	utiltestingapi "sigs.k8s.io/kueue/pkg/util/testing/v1beta2"
 	testingjob "sigs.k8s.io/kueue/pkg/util/testingjobs/job"
 	"sigs.k8s.io/kueue/test/util"
@@ -161,20 +160,18 @@ var _ = ginkgo.Describe("Failure Recovery Policy", ginkgo.Ordered, ginkgo.Contin
 			})
 		})
 
-		ginkgo.AfterEach(func() {
+		ginkgo.JustAfterEach(func() {
 			ginkgo.By("starting the kubelet on the node running the pod", func() {
 				cmd := exec.Command("docker", "exec", nodeName, "systemctl", "start", "kubelet")
 				gomega.Expect(cmd.Run()).To(gomega.Succeed())
 			})
 
 			ginkgo.By("waiting for the node to be ready again", func() {
-				gomega.Eventually(func(g gomega.Gomega) {
-					node := &corev1.Node{}
-					g.Expect(k8sClient.Get(ctx, client.ObjectKey{Name: nodeName, Namespace: ns.Name}, node)).To(gomega.Succeed())
-					g.Expect(tas.IsNodeStatusConditionTrue(node.Status.Conditions, corev1.NodeReady)).To(gomega.BeTrue())
-				}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
+				util.ExpectNodeToBecomeReady(ctx, k8sClient, nodeName, lq)
 			})
+		})
 
+		ginkgo.AfterEach(func() {
 			ginkgo.By("deleting the cluster queue", func() {
 				util.ExpectObjectToBeDeleted(ctx, k8sClient, cq, true)
 			})
--- a/test/e2e/tas/hotswap_test.go
+++ b/test/e2e/tas/hotswap_test.go
@@ -22,7 +22,6 @@ import (
 	"os/exec"
 	"slices"
 
-	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
 	batchv1 "k8s.io/api/batch/v1"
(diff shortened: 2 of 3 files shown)


test/e2e/customconfigs/failure_recovery_policy_test.go
```diff
@@ -166,6 +167,14 @@ var _ = ginkgo.Describe("Failure Recovery Policy", ginkgo.Ordered, ginkgo.Contin
 				gomega.Expect(cmd.Run()).To(gomega.Succeed())
 			})
 
+			ginkgo.By("waiting for the node to be ready again", func() {
```
mwielgus, line 169: As you are not 100% sure that it will help maybe consider adding some more logs to the tests so we can track what is actually failing?

mszadkow: also you can increase log verbosity and repeat tests couple times

mimowo: Oh indeed, disabling kubelet on a node may require waiting time to re-enable the node. Even if we wait for the node to be Ready in the test code it does not necessarily mean that all informers already consider the Node as ready. So, it is possible that Kueue scheduler may still see the Node notReady, and avoid scheduling here, which may result in some flakes also in other tests.

We solved this problem for node HotSwap by running a dummy workload (which requires the temporarily disabled node specifically) after the re-enablement of the Kubelet on the node. So I propose to extract this code https://github.com/kubernetes-sigs/kueue/blob/main/test/e2e/tas/hotswap_test.go#L93-L120 as a helper and use it here

test/e2e/customconfigs/failure_recovery_policy_test.go
```diff
@@ -166,6 +167,14 @@ var _ = ginkgo.Describe("Failure Recovery Policy", ginkgo.Ordered, ginkgo.Contin
...
+				gomega.Eventually(func(g gomega.Gomega) {
+					node := &corev1.Node{}
+					g.Expect(k8sClient.Get(ctx, client.ObjectKey{Name: nodeName, Namespace: ns.Name}, node)).To(gomega.Succeed())
+					g.Expect(tas.IsNodeStatusConditionTrue(node.Status.Conditions, corev1.NodeReady)).To(gomega.BeTrue())
+				}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
+			})
```
mszadkow: I think this is the right approach, checking if the node became ready again.
From the logs we can see that pod was created
```
Mar 09 02:03:00 kind-worker kubelet[1632]: I0309 02:03:00.870202    1632 kubelet.go:2588] "SyncLoop ADD" source="api" pods=["frp-plpm8/test-job-dqw79"]
```
but immediately evicted due to unreachable taint:
```
Mar 09 02:03:03 kind-worker kubelet[1632]: I0309 02:03:03.660364    1632 status_manager.go:1065] "Patch status for pod" pod="frp-plpm8/test-job-dqw79" podUID="1e5729a4-5dfc-4215-a236-2929863f4545" patch="{\"metadata\":{\"uid\":\"1e5729a4-5dfc-4215-a236-2929863f4545\"},\
...[{"type":"DisruptionTarget","observedGeneration":1,"status":"True","lastProbeTime":null,"lastTransitionTime":"2026-03-09T02:03:01Z","reason":"DeletionByTaintManager","message":"Taint manager: deleting due to NoExecute taint"},
```

test/e2e/customconfigs/failure_recovery_policy_test.go
kshalot: Since yesterday, I'm running the code locally on a loop and it does not flake (the original), that's why I'm not 100% sure whether this will fix it.

My hunch is that the node going back to ready can introduce some more delay when waiting for the job to be active, which can cause timeouts. I simulated something similar by reducing the timeout locally, but again - the exact sequence is not yet 100% clear to me.

The job has preferred anti affinity, so I'd assume it would try to get scheduled on some other "healthy" node if the node is not ready, but maybe there's still some delay introduced when it's ready but "warming up". I considered changing the anti affinity to required (it would make the test more predictable IMO), but I don't see how it would affect the flakiness, so I'd rather do it in a follow-up.

I'll look into https://github.com/kubernetes-sigs/kueue/blob/main/test/e2e/tas/hotswap_test.go#L93-L120.

test/e2e/customconfigs/failure_recovery_policy_test.go
mszadkow: But as @mimowo mentioned, we might require extra wait for dummy workload still.

test/e2e/customconfigs/failure_recovery_policy_test.go
mimowo: I see, this issue might be super hard to repro locally, because we usually work on more powerful boxes, so the delay between calling systemctl start kubelet , and the actualy NodeReady is very small. The boxes used for CI testing are often under heavy stress. Additionally, you probably didn't deploy other controllers while testing like Kubeflow, KubeRay, Prometheus etc. When we do e2e testing on CI they are all deployed on the cluster, which adds a lot more of variance to the timings. This may be considered both downside or upside, because thanks to the load and variance we sometimes catch race conditions in production code which we would not catch locally.

kshalot: Yeah, I'm aware and I agree.

I'm wondering whether we need all the restoring logic from hotswap though. Why not just extract `waitForDummyWorkloadToRunOnNode`, we are not deleting/restoring the Node object itself.

mimowo: Indeed, probably just this part should be in common https://github.com/kubernetes-sigs/kueue/blob/main/test/e2e/tas/hotswap_test.go#L112-L118

test/e2e/customconfigs/failure_recovery_policy_test.go
kshalot: Yeah. What I'm wondering about though is why the pod was scheduled there in the first place if the node wasn't ready. It's as if the node no longer has the `NoSchedule` taint while having the `NoExecute` one.

test/e2e/customconfigs/failure_recovery_policy_test.go
kshalot: I'm not 100% convinced yet that this is required here, but it certainly won't hurt. I extracted the logic to a helper.

PR #10142, opened 2026-03-27
cleanup: move ShouldExposeLocalQueueMetricsForWorkload to cache

mykysha:
#### What type of PR is this?
/kind cleanup

Cleanup ShouldExposeLocalQueueMetricsForWorkload

Part of #8830

```release-note
NONE
```

### DIFF
--- a/pkg/cache/scheduler/cache.go
+++ b/pkg/cache/scheduler/cache.go
@@ -583,14 +583,6 @@ func (c *Cache) GetCacheLocalQueue(cqName kueue.ClusterQueueReference, lqKey que
 	return nil, errQNotFound
 }
 
-func (c *Cache) GetLocalQueueLabels(cqName kueue.ClusterQueueReference, lqKey queue.LocalQueueReference) (map[string]string, error) {
-	lq, err := c.GetCacheLocalQueue(cqName, lqKey)
-	if err != nil {
-		return nil, err
-	}
-	return lq.GetLabels(), nil
-}
-
 func (c *Cache) ClusterQueueUsesAdmissionFairSharing(cqName kueue.ClusterQueueReference) bool {
 	c.RLock()
 	defer c.RUnlock()
@@ -1041,4 +1033,21 @@ func resourceFloat(name corev1.ResourceName, v int64) float64 {
 // Key is the key used to index the queue.
 func queueKey(q *kueue.LocalQueue) queue.LocalQueueReference {
 	return queue.NewLocalQueueReference(q.Namespace, kueue.LocalQueueName(q.Name))
+}
+
+// ShouldExposeLocalQueueMetricsForWorkload determines if LocalQueue metric reporting should be made for the associated LocalQueue.
+func (c *Cache) ShouldExposeLocalQueueMetricsForWorkload(log logr.Logger, wl *kueue.Workload) bool {
+	if !c.lqMetrics.IsEnabled() {
+		return false
+	}
+	if wl.Status.Admission == nil {
+		log.V(5).Info("Skip the update for local queue metrics for a workload without admission", "workload", klog.KObj(wl))
+		return false
+	}
+	lq, err := c.GetCacheLocalQueue(wl.Status.Admission.ClusterQueue, queue.KeyFromWorkload(wl))
+	if err != nil {
+		log.Error(err, "Failed to get LocalQueue for metrics", "localQueue", klog.KRef(wl.Namespace, string(wl.Spec.QueueName)))
+		return false
+	}
+	return lq.shouldExposeMetrics(c.lqMetrics)
 }
--- a/cmd/kueue/main.go
+++ b/cmd/kueue/main.go
@@ -362,7 +362,7 @@ func main() {
 
 	preemptionExpectations := preemptexpectations.New()
 
-	if err := setupControllers(ctx, mgr, cCache, queues, &cfg, serverVersionFetcher, roleTracker, preemptionExpectations, customLabels, lqMetrics); err != nil {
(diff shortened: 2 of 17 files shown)


pkg/cache/scheduler/cache.go
```diff
@@ -1042,3 +1034,20 @@ func resourceFloat(name corev1.ResourceName, v int64) float64 {
...
 	return queue.NewLocalQueueReference(q.Namespace, kueue.LocalQueueName(q.Name))
 }
+
+// ShouldExposeLocalQueueMetricsForWorkload detemines if LocalQueue metric reporting should be made for the associated LocalQueue.
+func (c *Cache) ShouldExposeLocalQueueMetricsForWorkload(log logr.Logger, wl *kueue.Workload) bool {
+	if !c.lqMetrics.IsEnabled() {
+		return false
+	}
+	if wl.Status.Admission == nil {
+		log.V(5).Info("WARNING: cannot expose local queue metrics of workload without status.Admission", "workload", wl)
```
mimowo: I know this is not related to PR. It might be ok for now, but this is not universally true. For example we may have some metrics which are also reported for pending workloads.

When we have that then we may need to be able to get the LocalQueue just by `queue.NewLocalQueueReference(wl.Namespace, wl.Spec.QueueName)`. This is technically speaking enough to uniquely identify each LocalQueue. I think we only use `wl.Status.Admission.ClusterQueue` because we don't have a flattened map of all LocalQueues, right?

pkg/cache/scheduler/cache.go
```diff
@@ -1042,3 +1034,20 @@ func resourceFloat(name corev1.ResourceName, v int64) float64 {
 func queueKey(q *kueue.LocalQueue) queue.LocalQueueReference {
 	return queue.NewLocalQueueReference(q.Namespace, kueue.LocalQueueName(q.Name))
 }
+
+// ShouldExposeLocalQueueMetricsForWorkload detemines if LocalQueue metric reporting should be made for the associated LocalQueue.
```
mwielgus: Determines.

pkg/cache/scheduler/cache.go
mwielgus: Wouldn't it log the entire workload object?

mwielgus: WARNING, info and V(5) seem inconsistent.

pkg/cache/scheduler/cache.go
```diff
@@ -1042,3 +1034,20 @@ func resourceFloat(name corev1.ResourceName, v int64) float64 {
...
+func (c *Cache) ShouldExposeLocalQueueMetricsForWorkload(log logr.Logger, wl *kueue.Workload) bool {
+	if !c.lqMetrics.IsEnabled() {
+		return false
+	}
+	if wl.Status.Admission == nil {
+		log.V(5).Info("WARNING: cannot expose local queue metrics of workload without status.Admission", "workload", wl)
+		return false
+	}
+	lq, err := c.GetCacheLocalQueue(wl.Status.Admission.ClusterQueue, queue.NewLocalQueueReference(wl.Namespace, wl.Spec.QueueName))
```
mwielgus: Would queue.KeyFromWorkload(wl) fit here?

pkg/cache/scheduler/cache.go
```diff
@@ -1042,3 +1034,20 @@ func resourceFloat(name corev1.ResourceName, v int64) float64 {
...
+	if err != nil {
+		log.V(5).Error(err, "Failed to get LocalQueue for metrics", "localQueue", klog.KRef(wl.Namespace, string(wl.Spec.QueueName)))
```
mwielgus: If this is an actionable error, consider using log.Error without .V(5). If this is just an expected trace or debug information, consider using log.V(5).Info("Failed to get LocalQueue for metrics", "err", err, ...) instead.

pkg/cache/scheduler/cache.go
mykysha: Yes, in this case we get lq instance from cache, and to get it we use cluster queue reference stored in admission. So we might need to update the cache to have a more direct access

mimowo: Ok, we can re-work that when needed. 

For now let's just adjust the comment because I think this is misleading. First of all: is this function called for Pending workloads? If so this is misleading because it is normal that Pending workloads don't have Admission.

So maybe it should rather be:

// We can skip the update for LocalQueue metrics for a pending workload as currently all our 
// LocalQueue metrics for workloads only record for workloads with admission.
// We may need to re-work this in the  future when such metrics are also recorded.
log.V(5).Info("Skip the update for local queue metrics for a workload without admission", "workload", klog.KObj(wl))

mykysha: Updated with the proposed message

pkg/cache/scheduler/cache.go
mykysha: Fixed

pkg/cache/scheduler/cache.go
mykysha: Fixed, now it logs just the reference to the workload

mykysha: Fixed via changing the message to the one proposed in another comment chain

pkg/cache/scheduler/cache.go
mykysha: It would, have changed now

pkg/cache/scheduler/cache.go
mykysha: I think this would be an actionable error explaining why the metrics are not appearing, so removed V(5) from the log

PR #10149, opened 2026-03-27
[FairSharing] Use flavor based borrowing information in entire root cohort - CQ tree when deciding if a workload needs borrowing

### LINKED ISSUE #10126: [Fair Sharing] Incorrect admission ordering with FairSharingPrioritizeNonBorrowing
**What happened**:

Suppose the following structure

```
- Root-Cohort
  - Cohort-A (10 CPUs)
    - CQ-A
  - Cohort-B
    - CQ-B (5 CPUs)
    - CQ-C
```

Then:

```
- submit 10cpu workload to CQ-C
- workload admits to CQ-C
- submit 5 cpu workload to each of CQ-A and CQ-B
- workload from CQ-B admits
```

workload from CQ-A should admit, as Cohort-A is not borrowing, while Cohort-B is.

See (LLM generated) test case: https://github.com/kubernetes-sigs/kueue/commit/be3e92c984ce93b772f557200d6e1e36ae84d45b

```
go test ./pkg/scheduler/ -run "TestSchedule/with_fair_sharing:_hierarchical_nominal-first" -count=1
```

This test passes with flag disabled.

See #9484 for definition of this flag

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

mukund-wayve:
#### What type of PR is this?

/kind bug

Fixes `FairSharingPrioritizeNonBorrowing` in hierarchical cohorts by checking borrowing at every level of the cohort tree (CQ to root), not just at the CQ level. The check is per-flavor: a subtree borrowing on one flavor does not penalize workloads requesting a different flavor where the subtree has ample quota. The DRS struct now exposes which FlavorResources are borrowed, enabling callers to query borrowing against any set of flavors.

Fixes #10126 

```release-note
FairSharing: Fix `FairSharingPrioritizeNonBorrowing` to check per-flavor borrowing at every hierarchy level in hierarchical cohorts, not just at the ClusterQueue level.
```

### DIFF
--- a/keps/1714-fair-sharing/README.md
+++ b/keps/1714-fair-sharing/README.md
@@ -11,6 +11,7 @@
     - [Story 3](#story-3)
   - [Risks and Mitigations](#risks-and-mitigations)
 - [Design Details](#design-details)
+  - [Nominal-first admission ordering](#nominal-first-admission-ordering)
   - [Share value function and weights](#share-value-function-and-weights)
   - [Preemption algorithm](#preemption-algorithm)
     - [Choosing ClusterQueues to preempt from](#choosing-clusterqueues-to-preempt-from)
@@ -200,6 +201,20 @@ The algorithms take a decision from root-to-leaves. This favors admission from
 suborganizations that, as a whole, still have unused resources, or favors preemptions 
 from suborganizations that, as a whole, are borrowing too much.
 
+### Nominal-first admission ordering
+
+When the `FairSharingPrioritizeNonBorrowing` feature gate is enabled,
+a workload whose subtree is not borrowing is preferred over one whose
+subtree is borrowing, before falling through to the DRS comparison.
+
+This check is **per-flavor**: a subtree that borrows on flavor X but has
+ample quota on flavor Y does not penalize workloads requesting flavor Y.
+
+In hierarchical cohorts, the check is applied at every level of the
+hierarchy (CQ to root). A subtree that fits within its quota at every
+ancestor is preferred over one that borrows at any ancestor, even if the
+borrowing occurs at a higher cohort level rather than the CQ itself.
+
 ### Share value function and weights
 
 The value function is a variation of DRF (see
@@ -374,6 +389,10 @@ metrics
 * Scalability scenarios
 
 ## Implementation History
+
+* 2024.03.18 - KEP created (#1773)
+* 2024.04.24 - Make dominant resource share flavor aware (#2037)
+* 2026.03.27 - Fix nominal-first to check per-flavor borrowing at every hierarchy level (#10149)
 
 ## Drawbacks
 
--- a/pkg/features/kube_features.go
+++ b/pkg/features/kube_features.go
@@ -62,8 +62,9 @@ const (
 	// owner: @mukund-wayve
 	// kep: https://github.com/kubernetes-sigs/kueue/issues/9406
 	//
-	// In fair sharing admission ordering, prefer workloads that fit
-	// within their CQ's nominal quota over workloads that require borrowing.
+	// In fair sharing admission ordering, prefer workloads whose subtree
+	// is not borrowing on the workload's requested flavors, checked at
+	// every level of the cohort hierarchy.
 	FairSharingPrioritizeNonBorrowing featuregate.Feature = "FairSharingPrioritizeNonBorrowing"
 
 	// owner: @trasc


pkg/cache/scheduler/fair_sharing.go
```diff
@@ -47,13 +47,21 @@ type DRS struct {
...
+	// currently borrowing on.
+	borrowedFRs []resources.FlavorResource
 }
 
 // NegativeDRS is used as a starting point for comparisons.
 func NegativeDRS() DRS {
 	return DRS{unweightedRatio: -1, dominantResource: "", fairWeight: defaultWeight}
 }
 
+// BorrowingDRS returns a DRS that is borrowing on the given FlavorResources.
```
mwielgus: This seems to be a test utility. Consider moving it to some test utility file.

pkg/scheduler/fair_sharing_iterator.go
```diff
@@ -161,27 +162,35 @@ type drsKey struct {
...
-	// CQ's nominal entitlement on another flavor.
+	aDrs := e.drsValues[drsKey{parentCohort: parentCohort, workloadKey: workload.Key(a.Obj)}]
+	bDrs := e.drsValues[drsKey{parentCohort: parentCohort, workloadKey: workload.Key(b.Obj)}]
+
 	if features.Enabled(features.FairSharingPrioritizeNonBorrowing) {
-		aRequiresBorrowing := a.assignment.RequiresBorrowing()
-		bRequiresBorrowing := b.assignment.RequiresBorrowing()
-		if aRequiresBorrowing != bRequiresBorrowing {
-			return !aRequiresBorrowing
+		// 1. Per-flavor subtree non-borrowing: prefer workloads whose
```
mwielgus: Maybe update KEP/docs as well?

pkg/scheduler/scheduler_test.go
mwielgus, line 10968: @mimowo @gabesaba - this file is growing way too large.

mimowo: agree, opened the issue https://github.com/kubernetes-sigs/kueue/issues/10166 to track it independently

mimowo:
cc @pajakd ptal

pkg/scheduler/scheduler_test.go
```diff
@@ -3110,6 +3110,653 @@ func TestSchedule(t *testing.T) {
...
+			},
+			wantLeft: map[kueue.ClusterQueueReference][]workload.Reference{
+				"cq-b": {"eng-alpha/wl-b"},
+			},
+		},
+		//            ROOT
+		//       /            \
+		// COHORT-A (6/18)    COHORT-B (20/18)
+		//   /         \        /         \
+		// cq-a1 (0/5) cq-a2 (6/3)  cq-b (0/5) cq-c (20/3)
```
pajakd: Add some whitespaces to fix the picture (looks as if `cq-a2` belonged to both cohorts :) ).

pkg/scheduler/scheduler_test.go
```diff
@@ -3110,6 +3110,653 @@ func TestSchedule(t *testing.T) {
...
+		//       admitted: 5 spot           admitted: 8 on-demand
+		//
+		// Shown as (usage/subtreeQuota). Both cohorts borrow from
+		// ROOT, but on different flavors.
+		// After pending workloads (both request on-demand):
+		// wl-a (cq-a, 3 on-demand) -> COHORT-A on-demand (3/10), not borrowing
+		// wl-b (cq-b, 3 on-demand) -> COHORT-B on-demand (11/5), borrowing
+		//
+		// Per-flavor gate prefers wl-a: COHORT-A is not borrowing
+		// on-demand despite borrowing spot.
```
pajakd: I don't see what is this test testing. We have 7 units of on-demand available and two workloads asking for 3 each. Both of them get in. Even if the preference was the other way around (wl-b preferred over wl-a), the outcome would be the same, no? I think two workload asking for 4 units each would test what we want (see that wl-a gets in and wl-b stays pending).

pkg/cache/scheduler/fair_sharing_test.go
```diff
@@ -831,3 +831,100 @@ func TestDominantResourceShare(t *testing.T) {
...
+		usage                    resources.FlavorResourceQuantities
+		requestedFRs             resources.FlavorResourceQuantities
+		wantBorrowingOnRequested bool
+		wantBorrowing            bool
+	}{
+		"borrows on requested flavor": {
+			usage:                    resources.FlavorResourceQuantities{cpuDefault: 3_000},
+			requestedFRs:             resources.FlavorResourceQuantities{cpuDefault: 1_000},
+			wantBorrowingOnRequested: true,
+			wantBorrowing:            true,
```
pajakd, line 851: perhaps give context (quota values) in a comment somewhere higher to make it easier to understand the testcases.

pkg/scheduler/fair_sharing_iterator.go
```diff
@@ -208,22 +217,27 @@ func (e *entryComparer) less(a, b *entry, parentCohort kueue.CohortReference) bo
...
 			continue
 		}
+		usage := entry.assignmentUsage()
 		// We add workload's usage to CQ, so that all
 		// subsequent DRS include the admission of workload.
-		revert := cq.SimulateUsageAddition(entry.assignmentUsage())
+		revert := cq.SimulateUsageAddition(usage)
+
+		wlKey := workload.Key(entry.Obj)
+		e.requestedFRs[wlKey] = usage.Quota
```
pajakd: Shouldn't this be behind the featuregate? (I know that reading of this is so perhaps writing should be too)

pkg/scheduler/scheduler_test.go
mukund-wayve: Good catch. I've change it to 4 and now we can see what wl-b is pending while wl-a gets in

pkg/cache/scheduler/fair_sharing_test.go
mukund-wayve: Done

pkg/scheduler/fair_sharing_iterator.go
mukund-wayve: Done

pkg/cache/scheduler/fair_sharing.go
mukund-wayve: done

pkg/scheduler/fair_sharing_iterator.go
mimowo: ai think we could update the fair sharing KEP with handling the special case

mukund-wayve: Done

PR #10158, opened 2026-03-27
Fix handling workload with limits-only in case of Kueue restart

mimowo:
#### What type of PR is this?

/kind bug

Follow up to https://github.com/kubernetes-sigs/kueue/pull/9939

```release-note
TAS: fix the bug that workloads which only specify resource limits, without requests, are not able to perform 
the second-pass scheduling correctly, after Kueue restart, responsible for NodeHotSwap and ProvisioningRequests. 
```

### DIFF
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -1035,7 +1035,7 @@ func (r *WorkloadReconciler) Create(e event.TypedCreateEvent[*kueue.Workload]) b
 	if !r.cache.AddOrUpdateWorkload(log, wlCopy) {
 		log.V(2).Info("ClusterQueue for workload didn't exist; ignored for now")
 	}
-	r.queues.QueueSecondPassIfNeeded(ctx, e.Object, 0)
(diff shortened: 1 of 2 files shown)


pkg/controller/core/workload_controller.go
```diff
@@ -1035,7 +1035,7 @@ func (r *WorkloadReconciler) Create(e event.TypedCreateEvent[*kueue.Workload]) b
 	if !r.cache.AddOrUpdateWorkload(log, wlCopy) {
 		log.V(2).Info("ClusterQueue for workload didn't exist; ignored for now")
 	}
-	r.queues.QueueSecondPassIfNeeded(ctx, e.Object, 0)
+	r.queues.QueueSecondPassIfNeeded(ctx, wlCopy, 0)
```
mimowo, line 1038: We need to use wlCopy, because only wlCopy contains defaulting for resources when only-limits are specified. This fix is analogous to https://github.com/kubernetes-sigs/kueue/pull/9939/changes#diff-60dd240c20adbd6a189d018d1c216c2d296730f446c341d8bf449fa6657964ffR1069, but covers Create events responsible for handling Kueue's restart

PR #10169, opened 2026-03-29
[trainjob] use wrappers pattern for the RuntimePatches

### LINKED ISSUE #10148: TrainJob tests: use wrappers pattern for the RuntimePatches

I would like to use the wrapper pattern for the RuntimePatches,

**Why is this needed**:

Because the tests are very verbose currently, example https://github.com/kubernetes-sigs/kueue/pull/10127/changes#diff-55c150e33811e00097a8c38c6ec4bc21f90572fc6dfcb29d14fbf9a796630572R233-R261

ShaanveerS:
#### What type of PR is this?
/kind cleanup
/area testing

Refactors TrainJob tests to use shared `RuntimePatch` wrappers and helpers instead of verbose inline patch literals and repeated manual lookup logic.

Fixes #10148

```release-note
NONE
```

### DIFF
--- a/pkg/controller/jobs/trainjob/trainjob_controller_test.go
+++ b/pkg/controller/jobs/trainjob/trainjob_controller_test.go
@@ -93,15 +93,11 @@ func TestRunWithPodsetsInfo(t *testing.T) {
 	}{
 		"should add to the TrainJob the config specified in the PodSet info": {
 			trainJob: testTrainJob.Clone().
-				RuntimePatches([]kftrainerapi.RuntimePatch{{
-					Manager: runtimePatchManagerName,
-					TrainingRuntimeSpec: &kftrainerapi.TrainingRuntimeSpecPatch{
-						Template: &kftrainerapi.JobSetTemplatePatch{
-							Metadata: &metav1.ObjectMeta{},
-							Spec:     &kftrainerapi.JobSetSpecPatch{},
-						},
-					},
-				}}).Obj(),
+				RuntimePatches([]kftrainerapi.RuntimePatch{
+					testingtrainjob.MakeRuntimePatchWrapper(runtimePatchManagerName).
+						EmptyMetadata().
+						Obj(),
+				}).Obj(),
 			podsetsInfo: []podset.PodSetInfo{
 				{
 					Name: "node",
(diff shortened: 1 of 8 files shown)


pkg/controller/jobs/trainjob/trainjob_controller_test.go
```diff
@@ -160,36 +136,18 @@ func TestRunWithPodsetsInfo(t *testing.T) {
...
-					{
-						Manager: runtimePatchManagerName,
-						TrainingRuntimeSpec: &kftrainerapi.TrainingRuntimeSpecPatch{
-							Template: &kftrainerapi.JobSetTemplatePatch{
-								Metadata: &metav1.ObjectMeta{},
-								Spec:     &kftrainerapi.JobSetSpecPatch{},
-							},
-						},
-					},
+					testingtrainjob.MakeRuntimePatchWrapper("user-manager").
```
mimowo, line 139: Let's use manager which contains the domain, like "example.com/user-manager". I think we discussed that in the TrainJob repo, not sure if the validation is added there though.

pkg/controller/jobs/trainjob/trainjob_controller_test.go
```diff
@@ -93,15 +93,11 @@ func TestRunWithPodsetsInfo(t *testing.T) {
...
-					Manager: runtimePatchManagerName,
-					TrainingRuntimeSpec: &kftrainerapi.TrainingRuntimeSpecPatch{
-						Template: &kftrainerapi.JobSetTemplatePatch{
-							Metadata: &metav1.ObjectMeta{},
-							Spec:     &kftrainerapi.JobSetSpecPatch{},
-						},
-					},
-				}}).Obj(),
+				RuntimePatches([]kftrainerapi.RuntimePatch{
+					testingtrainjob.MakeRuntimePatchWrapper(runtimePatchManagerName).
```
mimowo, line 97: No need to add the "Wrapper" word at the end, just MakeRuntimePatch. Same for other wrappers.

pkg/controller/jobs/trainjob/trainjob_controller_test.go
mimowo: cc @kaisoz @tenzen-y @andreyvelich are we going to have validation in the trainer that the manager-name contains the domain field? I think it would be very good, otherwise conflicts are possible in the future.

andreyvelich: Yes, we should introduce this validation in the Trainer webhook.
@astefanutti We haven't implemented this yet, right?

astefanutti: I think it was already there thanks to @kaisoz and been ported to RuntimePatches: https://github.com/kubeflow/trainer/pull/3309/changes#diff-8caf1d3ba7f5fba94f9dd16accf9a2861237950ff4b889319b83812d10f52a07R128

andreyvelich: Do we know why `trainjob_controller_test` in Kueue didn't fail in that case?

PR #10185, opened 2026-03-30
Fix preemption expectations

### LINKED ISSUE #10181: Preemptions can get stuck on preemption expectations
**What happened**:

Preemptions can occassionally get stuck due to unsatisfied preemption expectations, see https://github.com/kubernetes-sigs/kueue/pull/10179

* preemptions never get stuck on preemptionExpectations

test: https://github.com/kubernetes-sigs/kueue/pull/10179

The root cause is that the workload_controller removes the expctations before the scheduler sees the Evicted=true condition. As a result scheduler may repeat the request, and repeat adding the preemptionExpectation. The repeated expectation is never removed because the victim workload has already transitioned to Evicted.

mimowo:
#### What type of PR is this?
/kind bug
/kind regression

Fixes #10181

```release-note
Scheduling: fix the bug that scheduler could get stuck trying to preempt a workload due to the corruption of the
in-memory state tracking the pending preemptions (called preemptionExpectations).
```

### DIFF
--- a/test/integration/singlecluster/controller/jobs/job/suite_test.go
+++ b/test/integration/singlecluster/controller/jobs/job/suite_test.go
@@ -113,7 +113,10 @@ func managerAndControllersSetup(
 
 		cCache := schdcache.New(mgr.GetClient(), schdcache.WithLocalQueueMetrics(lqMetrics))
 		preemptionExpectations := preemptexpectations.New()
-		queueOptions := []qcache.Option{qcache.WithPreemptionExpectations(preemptionExpectations)}
+		queueOptions := []qcache.Option{
+			qcache.WithPreemptionExpectations(preemptionExpectations),
+			qcache.WithLocalQueueMetrics(lqMetrics),
+		}
 		queues := util.NewManagerForIntegrationTests(ctx, mgr.GetClient(), cCache, queueOptions...)
 
 		opts = append(opts, jobframework.WithCache(cCache))
--- a/test/integration/singlecluster/webhook/core/suite_test.go
+++ b/test/integration/singlecluster/webhook/core/suite_test.go
@@ -73,14 +73,14 @@ func managerSetup(ctx context.Context, mgr manager.Manager) {
 	gomega.Expect(err).NotTo(gomega.HaveOccurred())
 
 	cCache := schdcache.New(mgr.GetClient())
-	preemptexpectations := preemptexpectations.New()
-	queueOptions := []qcache.Option{qcache.WithPreemptionExpectations(preemptexpectations)}
+	preemptionExpectations := preemptexpectations.New()
+	queueOptions := []qcache.Option{qcache.WithPreemptionExpectations(preemptionExpectations)}
 	queues := util.NewManagerForIntegrationTests(ctx, mgr.GetClient(), cCache, queueOptions...)
 
 	configuration := &config.Configuration{}
 	mgr.GetScheme().Default(configuration)
 
-	failedCtrl, err := core.SetupControllers(mgr, queues, cCache, configuration, nil, preemptexpectations, nil)
+	failedCtrl, err := core.SetupControllers(mgr, queues, cCache, configuration, nil, preemptionExpectations, nil)
 	gomega.Expect(err).ToNot(gomega.HaveOccurred(), "controller", failedCtrl)
 
 	failedWebhook, err := webhooks.Setup(mgr, nil)


cmd/kueue/main.go
```diff
@@ -326,6 +326,8 @@ func main() {
 		os.Exit(1)
 	}
 
+	preemptionExpectations := preemptexpectations.New()
+	// queueOptions = append(queueOptions, qcache.WithPreemptionExpectations(preemptionExpectations))
```
mwielgus: Why this is commented out? Uncomment or remove.

pkg/cache/queue/manager.go
```diff
@@ -564,6 +573,7 @@ func (m *Manager) AddOrUpdateWorkloadWithoutLock(log logr.Logger, w *kueue.Workl
 		return ErrClusterQueueDoesNotExist
 	}
 	cq.PushOrUpdate(wInfo)
+	m.preemptionExpectations.ObservedUID(log, client.ObjectKeyFromObject(w), w.UID)
```
mwielgus, line 579: is it guaranteed that m.preemptionExpectations is not nil?

pkg/cache/queue/manager.go
```diff
@@ -156,6 +163,8 @@ type Manager struct {
 	lqMetrics    *metrics.LocalQueueMetricsConfig
 
 	requeuer inadmissibleRequeuer
+
+	preemptionExpectations *expectations.Store
```
mwielgus, line 171: Please add some documentation what is the intended use of this field.

pkg/controller/core/workload_controller.go
```diff
@@ -1068,11 +1066,6 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 	}
 	log.V(2).Info("Workload update event")
 
-	if workload.IsEvicted(e.ObjectNew) && !workload.IsEvicted(e.ObjectOld) {
```
mwielgus, line 1071: Where is the tracking done now?

test/integration/singlecluster/scheduler/preemption_test.go
```diff
@@ -237,6 +239,97 @@ var _ = ginkgo.Describe("Preemption", func() {
 			util.ExpectWorkloadsToBePending(ctx, k8sClient, wl4)
 		})
 
+		for i := range 20 {
```
mwielgus: Why 20?

mimowo: The test would only fail 1/2 of the time at the main branch. So repeating makes it more likely to catch the issue. I will remove the looping when ready to merge

pkg/controller/core/workload_controller.go
mimowo: Here: https://github.com/kubernetes-sigs/kueue/pull/10185/changes#diff-c7ed976ece2ea8cfd72847c76a214be5920b16aa7770e4d3cef7d8c974a36028R575
whenever we add it to the queue.

cmd/kueue/main.go
mimowo: uncommented, this is important

pkg/cache/queue/manager.go
mimowo: In the production code it is non-nil because it is wired in main.go when creating the single instance of the queue manager, via [NewManager](https://github.com/kubernetes-sigs/kueue/blob/2a9f6f00cae15104197f2c4a5e592f755c188799/pkg/cache/queue/manager.go#L164).  After that the  queue manager is not changed.

However, it might be nil for the unit and integration tests which don't use it. Those tests which use the field started to fail so I had to plumb many of them.

test/integration/singlecluster/scheduler/preemption_test.go
mimowo: Removed now

pkg/cache/queue/manager.go
mimowo: added a comment, ptal

cmd/experimental/kueue-populator/go.mod
```diff
@@ -94,14 +95,16 @@ require (
...
-	sigs.k8s.io/lws v0.7.0 // indirect
+	sigs.k8s.io/lws v0.8.0 // indirect
 	sigs.k8s.io/randfill v1.0.0 // indirect
-	sigs.k8s.io/structured-merge-diff/v6 v6.3.1 // indirect
-	volcano.sh/apis v1.13.1-0.20251028070205-46d20c0699e7 // indirect
+	sigs.k8s.io/structured-merge-diff/v6 v6.3.2 // indirect
+	volcano.sh/apis v1.14.0 // indirect
 )
+
+replace sigs.k8s.io/kueue => ../../../
```
mimowo, line 110: This and the related changes are needed so that populator is compiled using the "local" version of Kueue. This is needed, because we change the code to setup the controllers, and the populator is setting up the controllers for integration tests. 

cc @PBundyra @sohankunkerkar

test/integration/singlecluster/controller/jobs/job/suite_test.go
```diff
@@ -112,12 +112,14 @@ func managerAndControllersSetup(
 		lqMetrics := metrics.NewLocalQueueMetricsConfig(configuration.Metrics.LocalQueueMetrics)
 
 		cCache := schdcache.New(mgr.GetClient(), schdcache.WithLocalQueueMetrics(lqMetrics))
-		queues := util.NewManagerForIntegrationTests(ctx, mgr.GetClient(), cCache, qcache.WithLocalQueueMetrics(lqMetrics))
+		preemptionExpectations := preemptexpectations.New()
```
sohankunkerkar, line 115: Since we dropped `qcache.WithLocalQueueMetrics(lqMetrics)` from the queue manager, does this break LQ metric reporting in DeleteLocalQueue and Requeue code paths?

test/integration/singlecluster/webhook/core/suite_test.go
```diff
@@ -72,12 +73,14 @@ func managerSetup(ctx context.Context, mgr manager.Manager) {
 	gomega.Expect(err).NotTo(gomega.HaveOccurred())
 
 	cCache := schdcache.New(mgr.GetClient())
-	queues := util.NewManagerForIntegrationTests(ctx, mgr.GetClient(), cCache)
+	preemptexpectations := preemptexpectations.New()
```
sohankunkerkar: The variable `preemptexpectations` shadows the package alias `preemptexpectations`.

mimowo: fixed

test/integration/singlecluster/controller/jobs/job/suite_test.go
mimowo: fixed

test/integration/singlecluster/scheduler/inadmissible/suite_test.go
```diff
@@ -81,12 +82,14 @@ func managerAndSchedulerSetup(ctx context.Context, mgr manager.Manager) {
 	// Use a longer batch period so that
 	// requeue notifications are reliably collapsed in the test.
 	batchPeriod := 2 * time.Second
-	queues := util.NewManagerForIntegrationTestsWithBatchPeriod(ctx, mgr.GetClient(), cCache, batchPeriod)
+	preemptionExpectations := preemptexpectations.New()
```
sohankunkerkar, line 85: Is this suite_test using the shared-store wiring as other setups?
The queue manager gets `qcache.WithPreemptionExpectations(preemptionExpectations)`, and controllers get the same `preemptionExpectations`, but the scheduler here is still created without scheduler.WithPreemptionExpectations(preemptionExpectations)

mimowo: Indeed, I think this `preemptionExpectations` is nil for scheduler in this test suite, but it also was nil before so no regression here. Ideally we wire scheduler too, but we can do that when some tests in test/integration/singlecluster/scheduler/inadmissible/requeueing_test.go 
 start to require it.

PR #10193, opened 2026-03-30
Fix TAS preemption with multiple resources

### LINKED ISSUE #10089: Classic preemption fails when both quota-level and topology-level preemption are required
## What happened

With the following node configuration, CQ/cohort settings, and deployments:

**node**:

|*node*|cpu|gpu|
|---|---|---|
|node-1|8|8|
|node-2|8|8|

**cohort**:

|*cohort*|cpu quota|gpu quota|
|---|---|---|
|cohort|16|16|

**CQ**:

|*CQ*|parent cohort|cpu borrowing limit|gpu borrowing limit|
|---|---|---|---|
|low|cohort|16|16|
|standard|cohort|16|16|

```
preemption:
  reclaimWithinCohort: Any
  withinClusterQueue: LowerPriority
  borrowWithinCohort:
    policy: LowerPriority
```

**deployment**:

|*deployment*|CQ|priority|replicas|requests/cpu|requests/gpu|
|---|---|---|---|---|---|
|low-cpu|low|-1000|2|5|-|
|low-gpu|low|-1000|2|10m|8|
|standard|standard|0|1|5|8|

When a user submits the workloads in the order `low-cpu` -> `low-gpu` -> `standard`, Kueue does not admit `standard`, even though it has higher priority.

This seems to happen only when admitting `standard` would require both quota-level preemption and topology-level preemption. In the above example:
- Quota-level preemption is required since `low-gpu` consumes all of the cohort's nominal gpu quota. 
- Topology-level preemption is required since each node can take at most one pod from `low-cpu` or `standard`.

By contrast, `low-cpu` -> `standard` works as expected. `low-gpu` -> `standard` also works as expected.
In those cases, `standard` is admitted by preempting lower-priority workloads as needed.

The higher-priority workload `standard` is admitted by preempting lower-priority workloads as needed.

## How to reproduce it (as minimally and precisely as possible)

Reproduction code is available in: https://github.com/wafrelka/bugbox/tree/kueue-260325

To reproduce:

1. Install required components: `kind`, `kubectl`, `kustomize`, `make`
2. Clone the repository above and switch to the `kueue-260325` branch.
3. Run `make`
4. Observe the behavior.

Kueue Config:

- https://github.com/wafrelka/bugbox/blob/f4353ccf1954188df7863d2a2e06af25943a2c4d/manifests/kueue-system/kustomization.yaml

Node:

- https://github.com/wafrelka/bugbox/blob/f4353ccf1954188df7863d2a2e06af25943a2c4d/manifests/kind-node/node.yaml
- https://github.com/wafrelka/bugbox/blob/f4353ccf1954188df7863d2a2e06af25943a2c4d/manifests/kind-node/patches/node.yaml

Cohort:

- https://github.com/wafrelka/bugbox/blob/f4353ccf1954188df7863d2a2e06af25943a2c4d/manifests/kueue-user/cohort.yaml

CQ:

- https://github.com/wafrelka/bugbox/blob/f4353ccf1954188df7863d2a2e06af25943a2c4d/manifests/kueue-user/clusterqueue.yaml

Workload:

- https://github.com/wafrelka/bugbox/blob/f4353ccf1954188df7863d2a2e06af25943a2c4d/tests/low.yaml
- https://github.com/wafrelka/bugbox/blob/f4353ccf1954188df7863d2a2e06af25943a2c4d/tests/standard.yaml

## Anything else we need to know?

Additional observations:

- The Kueue log contains `Workload requires preemption, but there are no candidate workloads allowed for preemption`.
- The workload event shows `couldn't assign flavors to pod set main: insufficient unused quota for nvidia.com/gpu in flavor alpha, 8 more needed`

## Environment
 
- Kubernetes version (use `kubectl version`): v1.35.1
- Kueue version (use `git describe --tags --dirty --always`): 0.16.4
- Cloud provider or hardware configuration: kind + kwok
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

pajakd:
#### What type of PR is this?
/kind bug
/area tas

As described in the issue, the problem occurs if the flavor assigner assigns mixed modes `(Preempt, Fit)`, and the `Fit` mode has to be updated to `Preempt` because of TAS rules (fragmentation). 

In this PR if the RepresentativeMode is `Preempt`, we update the mode to `Preempt` in all PodSets and all Resources of this workload. This ensures that we consider for preemption, the candidates that only use the resource that originally was assigned the `Fit` mode.

Fixes #10089

```release-note
TAS: Fix a bug where preemption with multiple resources sometimes fails
```

### DIFF
--- a/pkg/scheduler/flavorassigner/flavorassigner.go
+++ b/pkg/scheduler/flavorassigner/flavorassigner.go
@@ -708,6 +708,18 @@ func (a *FlavorAssigner) assignFlavors(log logr.Logger, counts []int32) Assignme
 				// update the mode for all flavors and the representative mode
 				psAssignment.updateMode(NoFit)
 				assignment.representativeMode = ptr.To(NoFit)
+			} else {
+				// Update TAS-related assignments to Preempt because preemptions might be needed
+				// in resources in which total unused quota is sufficient (Fit), but the
+				// quota is fragmented.
+				for _, reqs := range tasRequests {
+					for _, req := range reqs {
+						if psAssignment := assignment.podSetAssignmentByName(req.PodSet.Name); psAssignment != nil {
+							psAssignment.updateMode(Preempt)
+						}
+					}
+				}
+				assignment.representativeMode = ptr.To(Preempt)
 			}
 		}
 	}
--- a/pkg/scheduler/scheduler_tas_test.go
+++ b/pkg/scheduler/scheduler_tas_test.go
@@ -4490,6 +4490,7 @@ func TestScheduleForTASCohorts(t *testing.T) {
 		nodes           []corev1.Node
 		pods            []corev1.Pod
 		topologies      []kueue.Topology
+		cohorts         []kueue.Cohort
 		resourceFlavors []kueue.ResourceFlavor
 		clusterQueues   []kueue.ClusterQueue
 		workloads       []kueue.Workload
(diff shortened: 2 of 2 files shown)


pkg/scheduler/scheduler_tas_test.go
```diff
@@ -6255,6 +6256,238 @@ func TestScheduleForTASCohorts(t *testing.T) {
...
+			cohorts: []kueue.Cohort{
+				*utiltestingapi.MakeCohort("tas-cohort-main").
+					ResourceGroup(
+						*utiltestingapi.MakeFlavorQuotas("tas-default").
+							Resource(corev1.ResourceCPU, "16").
+							Resource("example.com/gpu", "16").
+							Obj(),
+					).
+					Obj(),
+			},
```
PBundyra, line 6291: Are cohorts necessary to test this scenario?

pkg/scheduler/scheduler_tas_test.go
PBundyra, line 1: Can we aim for the minimal repro example? IIUC, one CQ should be enough

pkg/scheduler/scheduler_tas_test.go
mimowo: Even if they are not necessary I would still keep a test which corresponds to the reported issue. Still it might be worth thinking about "minimal" case.

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -708,6 +708,18 @@ func (a *FlavorAssigner) assignFlavors(log logr.Logger, counts []int32) Assignme
...
+				// in resources in which total unused quota is sufficient (Fit), but the
+				// quota is fragmented.
+				for _, reqs := range tasRequests {
+					for _, req := range reqs {
+						if psAssignment := assignment.podSetAssignmentByName(req.PodSet.Name); psAssignment != nil {
+							psAssignment.updateMode(Preempt)
+						}
+					}
+				}
+				assignment.representativeMode = ptr.To(Preempt)
```
mimowo, line 722: I have a feeling this should be encapsulated within a helper which sets the representativeMode, and while doing so it also lowers the mode for PsAssignments, wdyt? At the same time I'm ok to leave it for a follow up if we discover the need to deduplicate the pattern.

pkg/scheduler/scheduler_tas_test.go
mimowo: +1, we synced that this is going to be a follow up

PR #10199, opened 2026-03-30
Fix adding a trailing space to `GINKGO_ARGS`

kshalot:
#### What type of PR is this?
/kind bug
/area testing

`+=` adds an implicit space in Makefile. So invocations like:
```
GINKGO_ARGS="--repeat=10" make test-multikueue-e2e
```
fail with:
```
invalid value "10 " for flag -repeat: parse error
```

Fixes #

Found by @Singularity23x0, but I'm also checking out a flake and I saw the issue so I'm posting a fix.

We can also consider using `strip`.

```release-note
NONE
```

### DIFF
--- a/Makefile-test.mk
+++ b/Makefile-test.mk
@@ -60,7 +60,12 @@ KIND_CLUSTER_NAME ?= kind
 # Number of processes to use during e2e tests.
 E2E_NPROCS ?= 1
 
-GINKGO_ARGS += $(if $(filter-out 1,$(E2E_NPROCS)),-procs=$(E2E_NPROCS))
+# Inject the number of procs into GINKGO_ARGS
+GINKGO_PROC_ARG := $(if $(filter-out 1,$(E2E_NPROCS)),-procs=$(E2E_NPROCS))
+# Check empty string to avoid adding a trailing space
+ifneq ($(GINKGO_PROC_ARG),)
+    GINKGO_ARGS += $(GINKGO_PROC_ARG)
+endif
 
 # For restricting to a specific directory
 GO_TEST_TARGET ?= .


Makefile-test.mk
```diff
@@ -60,7 +60,12 @@ KIND_CLUSTER_NAME ?= kind
 # Number of processes to use during e2e tests.
 E2E_NPROCS ?= 1
 
-GINKGO_ARGS += $(if $(filter-out 1,$(E2E_NPROCS)),-procs=$(E2E_NPROCS))
+# Inject the number of procs into GINKGO_ARGS
+GINKGO_PROC_ARG := $(if $(filter-out 1,$(E2E_NPROCS)),-procs=$(E2E_NPROCS))
+# Check empty string to avoid adding a trailing space
+ifneq ($(EXTRA_PROCS),)
```
mbobrovskyi: What does it mean EXTRA_PROCS?

kshalot: Oh no, I was changing the name of the arg and I guess it did not auto update here or I just missed this. Thanks

kshalot:
> Can we do like this:
> 
> ```
> GINKGO_ARGS += $(strip $(if $(filter-out 1,$(E2E_NPROCS)),-procs=$(E2E_NPROCS)))
> ```
> 
> ?

IIUC no, because the problem is doing `+=` with an empty string, not that the right hand side is a space. The problem is we are doing this, which also adds a space:
```Makefile
GINKGO_ARGS += ""
```

Makefile-test.mk
```diff
@@ -60,7 +60,12 @@ KIND_CLUSTER_NAME ?= kind
 # Number of processes to use during e2e tests.
 E2E_NPROCS ?= 1
 
-GINKGO_ARGS += $(if $(filter-out 1,$(E2E_NPROCS)),-procs=$(E2E_NPROCS))
+# Inject the number of procs into GINKGO_ARGS
+GINKGO_PROC_ARG := $(if $(filter-out 1,$(E2E_NPROCS)),-procs=$(E2E_NPROCS))
+# Check empty string to avoid adding a trailing space
+ifneq ($(GINKGO_PROC_ARG),)
+    GINKGO_ARGS += $(GINKGO_PROC_ARG)
+endif
```
Singularity23x0, line 68: Nit: code clarity; non mandatory suggestion
```suggestion
# Read the declared number of procs
GINKGO_PROC_ARG := $(if $(filter-out 1,$(E2E_NPROCS)),-procs=$(E2E_NPROCS))
# Inject the number of procs into GINKGO_ARGS if declared
ifneq ($(GINKGO_PROC_ARG),)
    GINKGO_ARGS += $(GINKGO_PROC_ARG)
endif
```

PR #10211, opened 2026-03-31
Add borrowWithinCohort LowerPriorityBorrowersOnly policy

### LINKED ISSUE #10171: borrowWithinCohort: Add LowerPriorityBorrowersOnly policy to protect nominal quota during preemption

A new policy value `LowerPriorityBorrowersOnly` for `borrowWithinCohort.policy` (Classical Preemption only).

Today `borrowWithinCohort` supports `Never` and `LowerPriority`. `LowerPriority` allows a borrowing preemptor to preempt any lower-priority workload in the cohort, including workloads that are running within their CQ's nominal quota. 

The proposed `LowerPriorityBorrowersOnly` policy adds a constraint: a workload is only a valid preemption target if **the target CQ is borrowing** for the contested resources, and **removing that specific workload would not take the target CQ's usage below its nominal quota**.

**Concrete example:**

CQ-A (preemptor) needs to borrow 4 GPUs. CQ-B has nominal=10 GPUs, current usage=13 GPUs (borrowing 3):
- Workload W1 in CQ-B uses 2 GPUs → **valid target** (13-2=11, still ≥ nominal 10)
- Workload W2 in CQ-B uses 5 GPUs → **not a valid target** (13-5=8, below nominal 10)

CQ-C has nominal=10 GPUs, current usage=9 GPUs (not borrowing):
- No workloads in CQ-C are candidates, regardless of priority.

`maxPriorityThreshold` composes with this: both constraints must be satisfied (target must be lower priority, at or below the threshold, AND must be a borrower whose removal keeps the CQ at/above nominal).

**Why is this needed**:

In multi-tenant clusters, teams are allocated nominal quota as a guarantee. When a CQ borrows unused capacity from the cohort, it's understood that this capacity is best-effort and may be reclaimed. However, the current `LowerPriority` policy doesn't distinguish between borrowed and guaranteed capacity, meaning a borrowing preemptor can evict workloads that are running entirely within a CQ's guaranteed allocation.

`LowerPriorityBorrowersOnly` provides a middle ground between `Never` (no preemption while borrowing) and `LowerPriority` (unrestricted preemption of lower-priority workloads): borrowing preemptors can reclaim borrowed capacity from other CQs, but guaranteed/nominal capacity is protected. This aligns with the principle that nominal quota should be a hard guarantee.

isumitsolanki:
Introduce `BorrowWithinCohortPolicyLowerPriorityBorrowersOnly` for classical preemption. When the preemptor is borrowing, a workload in another `ClusterQueue` is only a candidate if removing it leaves that queue at or above nominal quota for every contested flavor-resource.

- Extend v1beta1/v1beta2 `ClusterQueue` API and regenerate CRDs, Helm chart, and client apply configs.
- Add `ClusterQueueSnapshot.UsageRemainsAtOrAboveNominalForFlavorResources` and use it in hierarchical preemption classification.
- Add scheduler unit tests and a webhook integration case.
- Add KEP-10171 and link it from KEP-1337 (README + `kep.yaml`).

#### What type of PR is this?

/kind feature
/kind api-change
/kind kep

Adds `LowerPriorityBorrowersOnly` to `spec.preemption.borrowWithinCohort.policy` so clusters can let borrowing preemptors reclaim cohort capacity from **other** queues without evicting workloads that would push the victim **below** nominal on the resources involved in preemption. This matches multi-tenant expectations that nominal quota stays protected while borrowed capacity remains reclaimable.

`maxPriorityThreshold` still applies together with this policy. Behavior is **classical preemption only** (unchanged from existing `borrowWithinCohort` scope; Fair Sharing is unaffected).

Fixes #10171

- KEP: `keps/10171-borrow-within-cohort-lower-priority-borrowers-only/`; cross-links added under KEP-1337.
- Implementation: candidates that fail the nominal-remainder check are dropped entirely (not only in the borrowing pass), so nominal protection is not bypassed when the scheduler tries a non-borrowing preemption attempt.
- Please confirm wording for user-facing docs (e.g. `site/` follow-up) if you want this PR to include them; this change set focuses on API, codegen, scheduler, tests, and KEP.

```release-note
ClusterQueue: new value `LowerPriorityBorrowersOnly` for `spec.preemption.borrowWithinCohort.policy`. When a workload needs to borrow, it may preempt lower-priority workloads in other cohort queues only if removing each candidate keeps that queue at or above nominal quota for the contested resources. Composes with `maxPriorityThreshold`. Classical preemption only.****

### DIFF
--- a/apis/kueue/v1beta1/clusterqueue_types.go
+++ b/apis/kueue/v1beta1/clusterqueue_types.go
@@ -507,8 +507,9 @@ type ClusterQueuePreemption struct {
 type BorrowWithinCohortPolicy string
 
 const (
-	BorrowWithinCohortPolicyNever         BorrowWithinCohortPolicy = "Never"
-	BorrowWithinCohortPolicyLowerPriority BorrowWithinCohortPolicy = "LowerPriority"
+	BorrowWithinCohortPolicyNever                      BorrowWithinCohortPolicy = "Never"
+	BorrowWithinCohortPolicyLowerPriority              BorrowWithinCohortPolicy = "LowerPriority"
+	BorrowWithinCohortPolicyLowerPriorityBorrowersOnly BorrowWithinCohortPolicy = "LowerPriorityBorrowersOnly"
 )
 
 // BorrowWithinCohort contains configuration which allows to preempt workloads
@@ -522,9 +523,14 @@ type BorrowWithinCohort struct {
 	// - `LowerPriority`: allow preemption, in other ClusterQueues
 	//    within the cohort, for a borrowing workload, but only if
 	//    the preempted workloads are of lower priority.
+	// - `LowerPriorityBorrowersOnly`: like `LowerPriority`, but a workload in
+	//    another ClusterQueue is a valid target only if that ClusterQueue is
+	//    borrowing for the contested resources and removing the target workload
+	//    would not reduce that ClusterQueue's usage below its nominal quota
+	//    for those resources.
 	//
 	// +kubebuilder:default=Never
-	// +kubebuilder:validation:Enum=Never;LowerPriority
+	// +kubebuilder:validation:Enum=Never;LowerPriority;LowerPriorityBorrowersOnly
 	Policy BorrowWithinCohortPolicy `json:"policy,omitempty"`
 
 	// maxPriorityThreshold allows to restrict the set of workloads which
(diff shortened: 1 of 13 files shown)


amy:
/close 

the originally linked issue was based on a false premise. Please reopen if still relevant and add clarifications in comments. Thank you for the PR!

PR #10229, opened 2026-03-31
multikueue: requeue workloads on AdmissionCheck config replacement

### LINKED ISSUE #10122: MultiKueue: workload reconciler overlooks MultiKueueConfig replacements
### Background

In MultiKueue, the [workload reconciler](https://github.com/kubernetes-sigs/kueue/blob/ac4afcf77713d48e2faa67e53c522709ecc8c7ed/pkg/controller/admissionchecks/multikueue/workload.go) is set up [to watch changes in MultiKueueConfig objects](https://github.com/kubernetes-sigs/kueue/blob/ac4afcf77713d48e2faa67e53c522709ecc8c7ed/pkg/controller/admissionchecks/multikueue/workload.go#L915).

Thanks to this, if new clusters are added ([here](https://github.com/kubernetes-sigs/kueue/blob/ac4afcf77713d48e2faa67e53c522709ecc8c7ed/apis/kueue/v1beta2/multikueue_types.go#L150)), the watch handler [will notice this](https://github.com/kubernetes-sigs/kueue/blob/ac4afcf77713d48e2faa67e53c522709ecc8c7ed/pkg/controller/admissionchecks/multikueue/workload.go#L857) and [trigger reconciliation](https://github.com/kubernetes-sigs/kueue/blob/ac4afcf77713d48e2faa67e53c522709ecc8c7ed/pkg/controller/admissionchecks/multikueue/workload.go#L860) for all potentially affected workloads.

In particular, some pending workloads may be _immediately_ dispatched and scheduled on the new workers.

### The issue

The issue is that we watch for changes only in *one place* in the following diagram:

```
ClusterQueue object ---> AdmissionCheck object ---> MultiKueueConfig object ---> .Clusters field
```

Specifically:

* We will detect when `.Clusters` inside a single MultiKueueConfig is *changed*.

* We **will not** detect when the whole MultiKueueConfig is **replaced** in the AdmissionCheck.

* We will detect when the whole AdmissionCheck is replaced in the ClusterQueue. \
  (A bit around: replacing the AdmissionCheck will trigger an update of the manager workload in the Kueue core workload controller, and that will trigger a reconcile in the MultiKueue workload controller).

As a result, replacing MultiKueueConfig inside the AdmissionCheck may leave MultiKueue workloads pending, despite new capacity being available.

I guess the intended behavior is to handle this in the same way as for an updated MultiKueueConfig.

### Real demo

[This branch](https://github.com/olekzabl/kueue/tree/repro-5/hack/bug-repro) contains scripts which all test the following scenario:

0. (Not covered by the script) Run our [standard "set up MultiKueue with TAS" script](https://github.com/kubernetes-sigs/kueue/blob/3f0c4b2884fe5577d7a7ae4c3579a49718077be3/site/static/examples/multikueue/dev/setup-kind-multikueue-tas.sh).
1. Set up MultiKueue with manager quota of 3 CPU, and worker quotas of 1 CPU.
2. Initially have just 1 worker.
3. Submit 2 Jobs, 1 CPU each. (Obviously only one will run).
4. *Add* another worker.
5. Wait for the second Job to start.
6. If the second Job is still pending, submit a third one. (To verify that we successfully added a worker).

The key difference is *how* we add the second worker in step 4:

* [`test-update-mkconfig.sh`](https://github.com/olekzabl/kueue/blob/repro-5/hack/bug-repro/test-update-mkconfig.sh) updates the MultiKueueConfig;
* [`test-update-ac.sh`](https://github.com/olekzabl/kueue/blob/repro-5/hack/bug-repro/test-update-ac.sh) replaces the MultiKueueConfig referenced in the AdmissionCheck;
* [`test-update-cq.sh`](https://github.com/olekzabl/kueue/blob/repro-5/hack/bug-repro/test-update-cq.sh) replaces the AdmissionCheck referenced in the ClusterQueue.

The outputs of these scenarios can be viewed in the corresponding `*.out` files. Basically:

* In the "update MKC" and "update CQ" scenarios, the second Job runs immediately.
* In the "update AC" scenario, the second Job does not start while the third Job does.

(Note: we deliberately set a small `workerLostTimeout` and wait longer than that. See "Remarks" below for why that duration might matter).

### Remarks

1. Whenever the MK workload reconciler is triggered for a particular manager workload, it will learn the new MultiKueueConfig due to fetching the whole chain of dependencies ([here](https://github.com/kubernetes-sigs/kueue/blob/ac4afcf77713d48e2faa67e53c522709ecc8c7ed/pkg/controller/admissionchecks/multikueue/workload.go#L298)). 
   So the issue is only about triggering such reconcile.

2. While above we discussed *adding* workers, the story for *removing* workers is a bit different:

   1. It's not clear what the desired outcome is.

      * (A) Should the manager immediately detect workloads left on the removed remotes, and mark them pending from MultiKueue perspective (leading to re-dispatching and re-scheduling, if possible)?

      * (B) Or rather let the running remotes run, as they've already started, and avoid duplication of work on the "still registered" workers?

   2. The *actual* behavior is also a bit subtler.

      * Even if the worker removal happens via a change of `.Clusters` in a single MultiKueueConfig (and hence is detected by existing watches), the effect is that the MK workload reconciler - seeing no running remote among the still-registered workers - [considers](https://github.com/kubernetes-sigs/kueue/blob/ac4afcf77713d48e2faa67e53c522709ecc8c7ed/pkg/controller/admissionchecks/multikueue/workload.go#L506-L507) this as a "lost remote" case, and plans for re-dispatching only after `workerLostTimeout` passes.

      * OTOH, if the worker removal happens via a replacement of MultiKueueConfig ("undetected" by existing watches), the MK workload reconciler is *still likely* to react in the same way after `workerLostTimeout`, due to [this](https://github.com/kubernetes-sigs/kueue/blob/ac4afcf77713d48e2faa67e53c522709ecc8c7ed/pkg/controller/admissionchecks/multikueue/workload.go#L504). In practice, the reaction will just happen *somewhat later* because of the moment when we "started the watch" ([admission time](https://github.com/kubernetes-sigs/kueue/blob/ac4afcf77713d48e2faa67e53c522709ecc8c7ed/pkg/controller/admissionchecks/multikueue/workload.go#L508) for in-MKC updates vs. last reconcile time for MKC replacements).

   Therefore, for workload removals, I could agree that the current behavior is already acceptable.
   (Still, adding the missing watch should not hurt).

MatteoFari:
#### What type of PR is this?
/kind bug
/area multikueue

This fixes a missing update trigger in MultiKueue.

If a `MultiKueueConfig` is changed in place, workloads get rechecked. But if an `AdmissionCheck` is changed to point to a different `MultiKueueConfig`, that recheck does not happen. This PR adds that missing trigger.

Fixes #10122

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Workloads now react to admission check changes, so related items are re-queued automatically when needed.
  * Updates to shared configuration now also trigger the same workload refresh behavior across all linked admission checks.

* **Bug Fixes**
  * Improved handling so only workloads tied to the relevant admission check are reprocessed, reducing unnecessary queueing.
  * Added coverage for create and update events to ensure the behavior stays consistent.

### DIFF
--- a/pkg/controller/admissionchecks/multikueue/workload.go
+++ b/pkg/controller/admissionchecks/multikueue/workload.go
@@ -936,6 +936,11 @@ type configHandler struct {
 	eventsBatchPeriod time.Duration
 }
 
+type admissionCheckHandler struct {
+	client            client.Client
+	eventsBatchPeriod time.Duration
+}
+
 func (c *configHandler) Create(context.Context, event.CreateEvent, workqueue.TypedRateLimitingInterface[reconcile.Request]) {
 	// no-op as we don't need to react to new configs
 }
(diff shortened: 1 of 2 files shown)


pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -892,18 +897,69 @@ func (c *configHandler) queueWorkloadsForConfig(ctx context.Context, configName
...
-		}
-		for _, workload := range workloads.Items {
-			q.AddAfter(reconcile.Request{NamespacedName: client.ObjectKeyFromObject(&workload)}, c.eventsBatchPeriod)
 		}
 	}
 	return errors.Join(errs...)
 }
 
+func (a *admissionCheckHandler) Create(context.Context, event.CreateEvent, workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	// no-op as we don't need to react to new admission checks
```
olekzabl: I'm not sure we don't need to react to creation.

Imagine a set of objects (workload, CQ, AC, MKConfig etc.) placed in a single YAML file, created "all together". There is no guaranteed order of creation. AC could be created last, and it even could so happen that all the other objects have been processed by their controllers before the AC has been created.

If so, it seems we need this particular handler to wire things together? Or am I missing sth?

pkg/controller/admissionchecks/multikueue/workload_test.go
```diff
@@ -2363,6 +2363,256 @@ func TestConfigHandlerUpdate(t *testing.T) {
...
+				Parameters(kueue.GroupVersion.Group, "MultiKueueConfig", "config1").
+				Obj(),
+			newAdmissionCheck: func() *kueue.AdmissionCheck {
+				ac := utiltestingapi.MakeAdmissionCheck("ac1").
+					ControllerName(kueue.MultiKueueControllerName).
+					Parameters(kueue.GroupVersion.Group, "MultiKueueConfig", "config1").
+					Obj()
+				ac.Labels = map[string]string{"updated": "true"}
+				return ac
+			}(),
```
olekzabl: Nit: this test case would be equally strong if you e.g. bumped `Generation` instead of modifying `Labels`.
Which would be more convenient as we already have the `.Generation()` wrapper method.
So no need for the `func() *AC { ... }()` trick.

olekzabl (commented):
Ouch, please accept my HUGE apologies for missing this PR for so long!

I like it. In particular, thanks a lot for the solid test coverage!

I have just 2 comments - one is a nit which could be ignored; the other one is more about the matter. PTAL.

pkg/controller/admissionchecks/multikueue/workload.go
MatteoFari: Thanks for the feedback, I didn't think about that.

olekzabl:
/lgtm
/approve

Thank you!

PR #10237, opened 2026-03-31
Enforce line length limits using golines formatter.

### LINKED ISSUE #10231: Enforce line length limits using lll linter

I would like to integrate and enforce the [lll](https://golangci-lint.run/docs/linters/configuration/#lll) (long line linter).

**Why is this needed**:

Long lines make code reviews more difficult, especially in side-by-side diffs, as they require horizontal scrolling and obscure the context of changes.

mbobrovskyi:
#### What type of PR is this?
/kind cleanup

Enforce line length limits using golines formatter.

Fixes #10231

```release-note
NONE
```

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -119,6 +119,7 @@ issues:
 formatters:
   enable:
     - gci
+    - golines
   settings:
     gci:
       sections:
@@ -127,6 +128,9 @@ formatters:
         - prefix(sigs.k8s.io/kueue) # Custom section: groups all imports with the specified Prefix.
         - blank # Blank section: contains all blank imports. This section is not present unless explicitly enabled.
         - dot # Dot section: contains all dot imports.
+    golines:
+      max-len: 200
+      reformat-tags: false
   exclusions:
     generated: lax
     paths:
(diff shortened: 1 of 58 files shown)


.golangci.yaml
```diff
@@ -127,6 +128,9 @@ formatters:
         - prefix(sigs.k8s.io/kueue) # Custom section: groups all imports with the specified Prefix.
         - blank # Blank section: contains all blank imports. This section is not present unless explicitly enabled.
         - dot # Dot section: contains all dot imports.
+    golines:
+      max-len: 200
```
mimowo, line 132: Note for posterity as we synced on slack: this PR originally started with 120 characters but Kueue is already a large codebase, and the PR was too large, so this would substantially make git blame harder to use. Also I actually didn't like some of the reformatting as too aggressive.

At 200 characters:
- the formatting is not so aggressive and seems to capture only the legitimate cases
- this should be enough on large modern days screens, not always enough when working on a laptop for side-by-side review,  but that's fine
- this seems to be the approach other large code bases take, using LLM I found other projects using 200:  external-secrets/external-secrets, openshift/cert-manager-operator, openshift/external-secrets-operator, stolostron/multicluster-observability-operator, microsoft/fabric-terraform-quickstart.

mimowo:
Thank you 👍 
/lgtm
/approve

This is probably going to fail, so please prepare manually, ideally today because I expect many conflicts on other branches 
/cherrypick release-0.17
/cherrypick release-0.16

Issue #10093, opened 2026-03-25
Make MultiKueue usage observable for MK manager clusters


polinasand:
**What would you like to be added**:
Ensure that MK controller places MultiKueue specific labels on manager Pods.
This issue is focused on MK **manager** clusters. For MK worker clusters, see #10083.

**Why is this needed**:
The goal is to make it observable whether a given Kueue controller deployment is using MultiKueue (on the manager side). For this, we'd like to put some specific label (e.g. `kueue.x-k8s.io/multikueue-manager-cluster=true`) on its Pods.

**Design considerations**
There are few approaches on how the label can be placed/removed:
1. Add the label if there is a MK cluster with AC attached, never remove it.
+ adv: the simplest one
2. Add the label if there is a MK cluster with AC attached to a "healthy" CQ ("can admit workloads"). Otherwise remove the label.
3. Add the label if there is a MK cluster with AC and some workloads are actually admitted through it. Otherwise remove the label.
+ adv: most consistent with the worker side

Possible options how to detect if AC is running/changed/not running:
Use MK AC controller.
1. Create and keep the set of admitted workload. Update it based on workload change events in Workload Reconciler. Track this set to update labels on `kueue-controller-manager` pod.
2. Use information about number of Admitted Workloads from scheduler cache. The issue here is that MK controller has no direct access to it. Possible solution is calculating whether there are admitted workloads inside [setupControllers](https://github.com/kubernetes-sigs/kueue/blob/766f0232b868cc1ce204d15fc13a7f89f7ff082c/cmd/kueue/main.go#L426), and then passing this info to Workload Reconciler through [multikueue.SetupControllers](https://github.com/kubernetes-sigs/kueue/blob/766f0232b868cc1ce204d15fc13a7f89f7ff082c/cmd/kueue/main.go#L470).

**Related issues**:
#10083

PR #10242, opened 2026-03-31
Add MK origin label for worker Pods

### LINKED ISSUE #10083: Make MultiKueue usage observable for MK worker clusters
Ensure that MK controller places MultiKueue specific labels on worker Pods running under MK Workloads.
This issue is focused on MK **worker** clusters. For MK manager clusters, see #10093.

**Why is this needed**:
The goal is to make MultiKueue usage observable. Initial research shows that there are no existing MK-specific labels, unlike, for example, for Kueue. Currently, the Kueue usage can be observed thanks to placing Kueue-specific labels on the Pods, for example `kueue.x-k8s.io/podset=main`. Similar labels can be placed on worker Pods for MK, for example: `kueue.x-k8s.io/multikueue-worker-workload=true`.

**Related issues**:
#10093

polinasand:
#### What type of PR is this?

/kind feature

Add MultiKueue-specific label (`kueue.x-k8s.io/multikueue-worker-workload-pod=true`) on worker Pods when run job in MK.

Demo: 
Setup MK and run 1 sample job. Then run `kubectl describe job` command.
This is how Job PodTemplate running on worker cluster looks like:
```
Name:             sample-job-nwgjc
...
Pod Template:
  Labels:       batch.kubernetes.io/controller-uid=1b9e01b9-80fe-4466-abec-3296cb75d660
                batch.kubernetes.io/job-name=sample-job-nwgjc
                controller-uid=1b9e01b9-80fe-4466-abec-3296cb75d660
                job-name=sample-job-nwgjc
                kueue.x-k8s.io/multikueue-worker-workload-pod=true
                kueue.x-k8s.io/podset=main
...
```
Then run `kubectl describe pod` command to see the corresponding Pod(s):
```
Name:             sample-job-nwgjc-sw52h
...
Labels:           batch.kubernetes.io/controller-uid=1b9e01b9-80fe-4466-abec-3296cb75d660
                  batch.kubernetes.io/job-name=sample-job-nwgjc
                  controller-uid=1b9e01b9-80fe-4466-abec-3296cb75d660
                  job-name=sample-job-nwgjc
                  kueue.x-k8s.io/multikueue-worker-workload-pod=true
                  kueue.x-k8s.io/podset=main
...
```

Fixes #10083 

```release-note
MultiKueue: Add the `kueue.x-k8s.io/multikueue-worker-workload-pod=true` label to workload pods running on worker clusters.
```


apis/kueue/v1beta2/multikueue_types.go
```diff
@@ -35,6 +35,8 @@ const (
 	// MultiKueueControllerName is the name used by the MultiKueue
 	// admission check controller.
 	MultiKueueControllerName = "kueue.x-k8s.io/multikueue"
+
+	MultiKueueWorkerWorkloadLabel = "kueue.x-k8s.io/multikueue-worker-workload"
```
Singularity23x0: Given that this label is added to the resources themselves, not the workloads representing them I propose renaming.

Looking how it is handled in case of the ["ManagedByKueue" label](https://github.com/kubernetes-sigs/kueue/blob/be29927bec3dd45788b1a5c8e36e4cf5f1a4df7a/pkg/constants/constants.go#L42) - i suggest following suit:

```suggestion
	ManagedByMultiKueueLabelKey   = "kueue.x-k8s.io/multikueue"
	ManagedByMultiKueueLabelValue = "true"
```

polinasand: In our case we want to distinguish workers and managers, so I would propose next labels:
`"kueue.x-k8s.io/multikueue-worker=true"`
`"kueue.x-k8s.io/multikueue-manager=true"`

pkg/controller/jobs/job/job_multikueue_adapter.go
```diff
@@ -105,6 +105,12 @@ func (b *multiKueueAdapter) SyncJob(ctx context.Context, localClient client.Clie
 	// clear the managedBy enables the batch/Job controller to take over
 	remoteJob.Spec.ManagedBy = nil
 
+	if remoteJob.Spec.Template.Labels == nil {
+		remoteJob.Spec.Template.Labels = map[string]string{}
+	}
+
+	remoteJob.Spec.Template.Labels[kueue.ManagedByMultiKueueLabelKey] = kueue.ManagedByMultiKueueLabelValue
```
kshalot: I'd like to at least consider some way of doing this more generically.

For example, couldn't we label the "local" object as well? Then the label would be copied automatically by `SyncJob`. The difference is that this annotation would also exist on the manager cluster in this case, so it could skew the results in some cases. AFAIR this is not acceptable for us?

kshalot: Or maybe add some new mechanism that can maintain additional details on the job object, in case we need to add more labels/annotations/anything in the future.

A very hacky example - pass some `func` to the `SyncJob` 🤷 Not the ideal solution, but maybe this will help us consider whether this is the right approach.

apis/kueue/v1beta2/multikueue_types.go
```diff
@@ -35,6 +35,10 @@ const (
 	// MultiKueueControllerName is the name used by the MultiKueue
 	// admission check controller.
 	MultiKueueControllerName = "kueue.x-k8s.io/multikueue"
+
+	ManagedByMultiKueueLabelKey = "kueue.x-k8s.io/multikueue"
```
kshalot: I noticed that we already have a `MultiKueueOriginLabel` added by each adapter. Maybe that is enough or do we need a new one?

pkg/controller/jobs/job/job_multikueue_adapter.go
kshalot: It seems that doing this in each adapter independently is the current status quo - https://github.com/search?q=repo%3Akubernetes-sigs%2Fkueue%20MultiKueueOriginLabel&type=code.

But this begs the question - https://github.com/kubernetes-sigs/kueue/pull/10242/files#r3016194946

Singularity23x0: Yeah but look at the Job - we can add the label to the local and it will be copied. But it will not propagate to the pods created by the Job. And different job types will require different approaches still, ranging from similar to completely different. The reason for implementing it inside the adapter is the same as for why the adapter exists in the first place. 

Creating a generic solution is in my opinion the matter of updating the [generic adapter](https://github.com/kubernetes-sigs/kueue/blob/51930db170e499df7a0385b530f250127b4e9b54/pkg/controller/admissionchecks/multikueue/externalframeworks/adapter.go#L17) and something much more complex than updating the existing job-specific adapters.

kshalot: I meant adding a label to the pod template of the local job, but that's not a very clean approach IMO.

We could imagine something like `extraLabels` and `podExtraLabels` (provisional naming) to handle the current `origin`, `workloadName` and the feature in this PR. This way it's more baked into the contract of `SyncJob` and not an implementation detail.

apis/kueue/v1beta2/multikueue_types.go
kshalot: OK, `MultiKueueOriginLabel` will only be propagated to the job itself, not the pods. Also it's being used by the watcher to filter the relevant objects, so we don't want to propagate it to the pods anyway.

pkg/controller/jobs/job/job_multikueue_adapter.go
kshalot: Or we can just re-use the `origin` argument instead of using `"true"` - it seems to do the same thing and potentially more, since it can distinguish between origins.

```go
remoteJob.Spec.Template.Labels[kueue.ManagedByMultiKueueLabelKey] = origin
```

kshalot: But I'm fine with the current approach of using a "true" as well.

apis/kueue/v1beta2/multikueue_types.go
kshalot: There is a mismatch between the issue/description and the key of this label:
`multikueue-worker-workload` vs `multikueue`.

Since this label is supposed to specifically track the pods of **workloads** in MK, maybe this should be more explicit. `multikueue` could mean either a workload pod or a controller manager pod.

For example: `multikueue-workload-pod` or `multikueue-worker-workload-pod`. I'm not sure we have to specify `worker` though - it can be assumed implicitly that if this is a workload pod then it's running on the worker.

kshalot (commented):
Given that the changes are quite predictable, I think you can change all the adapters in a single PR. It's going to be easier to spot whether we missed some adapter and will require less work overall IMO (fewer reviews etc.)

kshalot:
> > I assume you mean labeling the "local" Job that is being synced? I proposed this as well ([#10242 (comment)](https://github.com/kubernetes-sigs/kueue/pull/10242#discussion_r3016130440)), but I think there's one case that can be tricky here - some workloads create the pods on manager as well, which might introduce "false positives" into the observability.
> 
> What I'm proposing is to inject the new label MultiKueueOriginPodLabel for Pods into the Job template as we do for "PodSetLabel", we could inject it based on the presence of the `MultiKueueOriginLabel` label on the Job itself. This way we will not label the pods created on the management cluster - only on the workers, as only jobs on the workers will have MultiKueueOriginLabel

Okay, understood. When I considered this, I thought it would be slightly cleaner for all the MK related labels to be injected by the manager - then the workers are more agnostic of the fact that they are part of an MK cluster since the specs and metadata are prepared on the manager.

olekzabl:
> What I'm proposing is to inject the new label MultiKueueOriginPodLabel for Pods into the Job template as we do for "PodSetLabel", we could inject it based on the presence of the `MultiKueueOriginLabel` label on the Job itself. This way we will not label the pods created on the management cluster - only on the workers, as only jobs on the workers will have MultiKueueOriginLabel

The plan for injecting sounds nice, thanks!

To clarify, I understand that:

1. We'd set it near [here](https://github.com/kubernetes-sigs/kueue/blob/2b07fac8fd8f4916bfc1d31161ba068a88029f11/pkg/controller/jobframework/reconciler.go#L1528).
2. We'd set it based on whether `MultiKueueOriginLabel` is present on the _Workload_ object (rather than "Job itself"). \
   That's because the Workload object is already available, resource-kind-agnostic, and it also should have that label (when on a worker), populated [here](https://github.com/kubernetes-sigs/kueue/blob/2b07fac8fd8f4916bfc1d31161ba068a88029f11/pkg/controller/admissionchecks/multikueue/workload.go#L1098).

> _When I considered this, I thought it would be slightly cleaner for all the MK related labels to be injected by the manager - then the workers are more agnostic of the fact that they are part of an MK cluster since the specs and metadata are prepared on the manager._

I understand, though MK awareness is baked around anyway, e.g. [here](https://github.com/kubernetes-sigs/kueue/blob/2b07fac8fd8f4916bfc1d31161ba068a88029f11/pkg/controller/jobframework/reconciler.go#L1257-L1271) (which is one of the callers of what's discussed just above).

I even considered if we couldn't reuse that existing MK awareness for our manager/worker distinction, instead of looking at the `MultiKueueOriginLabel`. However, looking at the label is much simpler conceptually, and can be done regardless of which function was the caller. So I think I'd stick with it.

mimowo:
> To clarify, I understand that:
> 
> 1. We'd set it near [here](https://github.com/kubernetes-sigs/kueue/blob/2b07fac8fd8f4916bfc1d31161ba068a88029f11/pkg/controller/jobframework/reconciler.go#L1528).
> 2. We'd set it based on whether `MultiKueueOriginLabel` is present on the _Workload_ object (rather than "Job itself"). 
>    That's because the Workload object is already available, resource-kind-agnostic, and it also should have that label (when on a worker), populated [here](https://github.com/kubernetes-sigs/kueue/blob/2b07fac8fd8f4916bfc1d31161ba068a88029f11/pkg/controller/admissionchecks/multikueue/workload.go#L1098).

exactly 👍 

> I understand, though MK awareness is baked around anyway, e.g. [here](https://github.com/kubernetes-sigs/kueue/blob/2b07fac8fd8f4916bfc1d31161ba068a88029f11/pkg/controller/jobframework/reconciler.go#L1257-L1271) (which is one of the callers of what's discussed just above).

Yes, at this point MultiKueue isn't an easily separable "addon" anyway.

olekzabl:
> (b) make it possible to correlate the Pod with the "MutiKueue" manager instance if we have multiple of those in the future - this is the reason for the "origin" idea, rather than just boolean

That alone LGTM; yet, I'm concerned if it won't lead to confusion with the manager side naming.
(Please keep in mind that we ultimately want to convey the manager/worker distinction; see #10083 vs. #10093).

On the manager side, we want to label Pods of the Kueue controller, for which the concept of "MultiKueue origin" also generally makes sense (as it's [a part of manager-cluster-wide MK config](https://github.com/kubernetes-sigs/kueue/blob/2b07fac8fd8f4916bfc1d31161ba068a88029f11/cmd/kueue/main.go#L478)).

Then, how clear would such combination be?

```
kueue.x-k8s.io/multikueue-origin=abcdef   // on worker workload Pods
kueue.x-k8s.io/multikueue-manager=true   // on manager Kueue controller Pod(s)
```

(Technically, we could put `abcdef` as value also on the manager side, just to allow "even better observability" - but then I've no idea how to name those labels to avoid confusion in interpreting them...)

mimowo:
> Then, how clear would such combination be?

This soubds reasonable, but what about more generic: kueue.x-k8s.io/multikueue-cluster-role=manager/worker

olekzabl:
> what about more generic: kueue.x-k8s.io/multikueue-cluster-role=manager/worker

This I wouldn't like because one of the known use cases of this effort is to scan Pods with a tool (which we don't own) which, sadly, looks at Pod labels' keys and values somewhat separately. Thus, scattering knowledge between keys & values (like `...-multikueue-...=worker`) would make the concept of "MK worker" _less visible_ for that tool.

So I'd prefer to stay with

```
kueue.x-k8s.io/multikueue-origin=abcdef   // on worker workload Pods
kueue.x-k8s.io/multikueue-manager=true   // on manager Kueue controller Pod(s)
```

pkg/controller/jobframework/reconciler.go
```diff
@@ -1527,6 +1527,10 @@ func getPodSetsInfoFromStatus(ctx context.Context, c client.Client, w *kueue.Wor
 
 		info.Labels[constants.PodSetLabel] = string(psAssignment.Name)
 
+		if _, exists := w.Labels[kueue.MultiKueueOriginLabel]; exists {
+			info.Labels[kueue.MultiKueueOriginLabel] = string(w.Name)
+		}
```
kshalot, line 1532: Shouldn't this be the value of `w.Labels[kueue.MultiKueueOriginLabel]` instead of `string(w.Name)`? So it points to the name of the MultiKueue manager deployment.

polinasand: Doesn't `w.Labels[kueue.MultiKueueOriginLabel] ==  'multikueue'` in this case?

kshalot: It does by default, but you can change it:
https://github.com/kubernetes-sigs/kueue/blob/b08ac3784e89e1e00394d493a3a95e616bb8a0aa/cmd/kueue/main.go#L479

kshalot:
/release-note-edit
```release-note
MultiKueue: Add the `kueue.x-k8s.io/multikueue-origin` label to workload pods running on worker clusters.
```

Feel free to tweak further.

pkg/controller/jobframework/reconciler.go
```diff
@@ -1527,6 +1527,10 @@ func getPodSetsInfoFromStatus(ctx context.Context, c client.Client, w *kueue.Wor
 
 		info.Labels[constants.PodSetLabel] = string(psAssignment.Name)
 
+		if val, exists := w.Labels[kueue.MultiKueueOriginLabel]; exists {
+			info.Labels[kueue.MultiKueueOriginLabel] = val
+		}
```
kshalot: I'd consider adding a test for this behavior.

Singularity23x0: +1

pkg/controller/jobframework/reconciler.go
```diff
@@ -1527,6 +1527,10 @@ func getPodSetsInfoFromStatus(ctx context.Context, c client.Client, w *kueue.Wor
 
 		info.Labels[constants.PodSetLabel] = string(psAssignment.Name)
 
+		if val, exists := w.Labels[kueue.MultiKueueOriginLabel]; exists {
+			info.Labels[kueue.MultiKueueOriginLabel] = val
```
kshalot: Actually, should we re-use this label here?

It is used to setup a watch for the remote jobs:
https://github.com/kubernetes-sigs/kueue/blob/ea6e1e2621e80e7d705166904f78bdba94ba782e/pkg/controller/admissionchecks/multikueue/multikueuecluster.go#L184-L190
https://github.com/kubernetes-sigs/kueue/blob/ea6e1e2621e80e7d705166904f78bdba94ba782e/pkg/controller/admissionchecks/multikueue/multikueuecluster.go#L208

There is an adapter for plain pods, so if you have it enabled then won't we start triggering reconciles for those pods? Which at best would be a wasteful no-op.

CC @olekzabl

kshalot: What I'm saying is - maybe we should add a new label 🆕

olekzabl: Good point.

I _think_ we won't go as far as triggering reconciles for those Pods, given this:

https://github.com/kubernetes-sigs/kueue/blob/ea6e1e2621e80e7d705166904f78bdba94ba782e/pkg/controller/jobs/pod/pod_multikueue_adapter.go#L109-L112

then blocking us here:

https://github.com/kubernetes-sigs/kueue/blob/ea6e1e2621e80e7d705166904f78bdba94ba782e/pkg/controller/admissionchecks/multikueue/multikueuecluster.go#L232-L235

But it feels an implementation detail; I don't feel safe relying on this.

So I would use a new label; still thinking how to name it nicely...

olekzabl: FTR my favorite is
```
multikueue-worker-workload-pod=true
```
Giving up the idea of passing the "MK origin" there. 
For the user, the MK origin will be discoverable via the parent generic-job anyway.
And I just don't see how to pass it in a way which is not perplexing to read.

kshalot: I'm okay with that.

pkg/controller/jobframework/reconciler.go
```diff
@@ -1527,6 +1527,10 @@ func getPodSetsInfoFromStatus(ctx context.Context, c client.Client, w *kueue.Wor
 
 		info.Labels[constants.PodSetLabel] = string(psAssignment.Name)
 
+		if _, exists := w.Labels[kueue.MultiKueueOriginLabel]; exists {
+			info.Labels[kueue.MultiKueueWorkerWorkloadPodLabel] = "true"
```
kshalot, line 1531: Since the value is always the same maybe we could extract it to a constant next to the `MultiKueueWorkerWorkloadPodLabel` definition.

pkg/controller/admissionchecks/multikueue/workload_test.go
```diff
@@ -1956,7 +1963,7 @@ func TestWlReconcile(t *testing.T) {
 				cRec := newClustersReconciler(managerClient, TestNamespace, 0, defaultOrigin, nil, adapters, nil, nil)
 
 				worker1Client := getClientBuilder(ctx).
-					WithLists(&kueue.WorkloadList{Items: tc.worker1Workloads}, &batchv1.JobList{Items: tc.worker1Jobs}).
+					WithLists(&kueue.WorkloadList{Items: tc.worker1Workloads}, &batchv1.JobList{Items: tc.worker1Jobs}, &corev1.PodList{Items: tc.wantWorker1Pods}).
```
kshalot, line 1966: I don't think we should seed the client with `wantWorker1Pods`, since that wouldn't be a useful test.

This might not be the best place to test this, as I'm not sure this will actually create the pods. But you can check.

kshalot: > This might not be the best place to test this, as I'm not sure this will actually create the pods. But you can check.

My point is - this test will probably just patch the job object, but not create the pods themself. Because it's just running the reconcile function of the MK workload controller.

So we can either:
1. Check whether the Pod template is updated for the job.
2. Test it for real in an integration or e2e test.

olekzabl: I vote for 1.
(I think propagating labels from Pod template to Pods is not sth we need to test here).

pkg/controller/jobframework/reconciler_test.go
```diff
@@ -329,6 +330,68 @@ func TestReconcileGenericJob(t *testing.T) {
 			},
 			wantEvents: nil,
 		},
+		"job has MultiKueue origin label in PodTemplate if workload has Multikueue origin label": {
```
kshalot: I think this name is a little bit confusing. Also, we are not testing whether the job is labeled, but that its pods are labeled.

I'd propose sth like:

```suggestion
		"workload's MultiKueue origin label is propagated to the job's pods": {
```

or smth like this.

pkg/controller/jobframework/reconciler_test.go
```diff
@@ -342,6 +405,27 @@ func TestReconcileGenericJob(t *testing.T) {
...
+						kueue.MultiKueueWorkerWorkloadPodLabel: kueue.MultiKueueWorkerWorkloadPodValue,
+						kueueconstants.PodSetLabel:             "main",
+					},
+					Affinity:        nil,
+					NodeSelector:    map[string]string{},
+					Tolerations:     nil,
+					SchedulingGates: nil,
+				},
+			},
+			).Return(nil).AnyTimes()
```
kshalot: Is this an expectation for all the cases? Shouldn't we only expect the pod label to be present when the origin label is present on the workload? At first glance it looks like the expectation is moved from the test case to the runner.

Can't we do sth like `wantPodSets`?

kshalot (changes requested):

pkg/controller/jobframework/reconciler_test.go
kshalot: Ok, I must have missed the "in PodTemplate" part on my initial reading :) The name is fine as is. You can optionally consider the one I proposed, but I'm fine with both

pkg/controller/jobframework/reconciler_test.go
```diff
@@ -329,6 +331,88 @@ func TestReconcileGenericJob(t *testing.T) {
 			},
 			wantEvents: nil,
 		},
+		"MultiKueue worker pod origin label is propagated to PodTemplate if workload has Multikueue origin label": {
```
olekzabl: Nit: our label now has little to do with "origin"; I'd remove that word.

olekzabl (approved):

mimowo:
Reading the PR now it looks ok, I'm hesitating with additional comments, will follow soon if any. 

However, @polinasand please address all comments either by:
1. following suggestion, and resolving, especially if simple remarks
2. following suggestion, answering, but keeping unresolved if you think keeping them open is helpful for other reviewers / posterity
3. answer why you don't agree with the suggestion, or partly agree so that the discussion can continue - maybe the reviewer missed something, or more eyes / ideas are needed

pkg/controller/jobframework/reconciler.go
polinasand: Outdated

pkg/controller/jobframework/reconciler_test.go
```diff
@@ -329,6 +331,88 @@ func TestReconcileGenericJob(t *testing.T) {
 			},
 			wantEvents: nil,
 		},
+		"MultiKueue worker pod label is propagated to PodTemplate if workload has Multikueue origin label": {
+			featureGates: map[featuregate.Feature]bool{features.ElasticJobsViaWorkloadSlices: true},
```
mimowo: Why enabling elasticJobs, this has nothing to do with the new feature.

mimowo (commented):
LGTM, just one nit
/approve
I'm also ok with lgtm from @olekzabl if addressed in case I'm on other stuff

PR #10244, opened 2026-03-31
Add admissionFairSharing to fairSharing instead of spec.admissionScope

### LINKED ISSUE #7710: API Inconsistency in Fair Sharing and Admission Fair Sharing Configuration

To improve API consistency, we propose consolidating the configuration for both Fair Sharing (FS) and Admission Fair Sharing (AFS) under a single, unified fairSharing object in the ClusterQueueSpec.

Currently, Fair Sharing is configured in `spec.fairSharing`, while Admission Fair Sharing is configured separately in `spec.admissionScope`. We recommend moving the AFS configuration to be nested within the fairSharing object.

Here is an example of the proposed structure:
```
...
fairSharing:
  weight: "1"
  admissionFairSharing: # new substruct for AFS
    mode: "UsageBasedAdmissionFairSharing"
...
```

**Why is this needed**:
There is an inconsistency between how fair sharing status is reported and how it is configured.

The status for both FS and AFS is combined within the FairSharingStatus object (`ClusterQueueStatus.fairSharing`). However, their configurations are split across two different fields in the ClusterQueue specification. This separation is unintuitive and makes the API harder to use.

Unifying the configuration to mirror the status structure would provide several benefits:

* Improved Consistency: The configuration and status APIs would align, creating a more predictable user experience.

* Better Discoverability: Users would find all fair sharing-related settings in a single, logical location.

* Enhanced Readability: The relationship between Fair Sharing and Admission Fair Sharing would be much clearer in the spec.

**Completion requirements**:

- [ ] Design doc
- [x] API change
- [ ] Docs update

andrewseif:
#### What type of PR is this?
/kind feature
/kind api-change

We are moving Admission Fair Sharing to be nested in Fair Sharing instead of ClusterQueue Spec.
The status for both FS and AFS is combined within the FairSharingStatus object (ClusterQueueStatus.fairSharing). However, their configurations are split across two different fields in the ClusterQueue specification
Fixes #7710 

Yes

```release-note
Migrated Admission Fair Sharing to be nested in Fair Sharing instead of ClusterQueue Spec.
```


apis/kueue/v1beta2/clusterqueue_types.go
```diff
@@ -138,10 +138,6 @@ type ClusterQueueSpec struct {
 	// if FairSharing is enabled in the Kueue configuration.
 	// +optional
 	FairSharing *FairSharing `json:"fairSharing,omitempty"`
-
-	// admissionScope indicates whether ClusterQueue uses the Admission Fair Sharing
```
kannon92, line 142: This is a breaking change.

IMO this work should be done on promotion from v1beta2 to v1.

/hold

kannon92 (commented):
/hold

https://github.com/kubernetes-sigs/kueue/pull/10244/changes#r3016139897

We should never remove API fields without a new API promotion. This goes against Kubernetes api practices for beta fields.

apis/kueue/v1beta2/clusterqueue_types.go
andrewseif: for some reason my brain glitched and thought we are removing it, my apologies, I have fixed it 😥

kannon92: github won't let me resolve but this is better now. thanks for that.

apis/kueue/v1beta1/fairsharing_types.go
```diff
@@ -38,6 +38,23 @@ type FairSharing struct {
 	// When not 0, Weight must be greater than 10^-9.
 	// +kubebuilder:default=1
 	Weight *resource.Quantity `json:"weight,omitempty"`
+
+	// admissionFairSharing contains the properties of the ClusterQueue
```
kannon92: I'm not sure if we need to add this to v1beta1 anymore.

kannon92 (changes requested):
1) We should update our documentation/site to reflect the new changes.

2) What is our behavior if someone uses the deprecated field and the new field? I didn't see anything around validation to avoid this.

3) For existing CQ that have this field, should we default to new field so that over time the deprecated field gets no longer used?

Few questions for high level review.

apis/kueue/v1beta1/fairsharing_types.go
andrewseif: sure I can remove it, I just added it for completion's sake.

pkg/cache/queue/cluster_queue.go
```diff
@@ -183,6 +183,16 @@ func withAfsConsumedResources(consumed *queueafs.AfsConsumedResources) clusterQu
 	}
 }
 
+// admissionFairSharingFromSpec returns the effective AdmissionFairSharing for a ClusterQueue,
```
kannon92, line 194: **AI Review**

This does look a little suspect.  

  1. Inconsistent backward-compatibility fallback (Bug)                                                                                                                                                                                                                                   
  
  The scheduler cache in pkg/cache/scheduler/clusterqueue.go:189-196 correctly falls back to the deprecated AdmissionScope:                                                                                                                                                               
                              
```  
switch {                                                                                                                                                                                                                                                                                
  case in.Spec.FairSharing != nil && in.Spec.FairSharing.AdmissionFairSharing != nil:
      c.AdmissionFairSharing = in.Spec.FairSharing.AdmissionFairSharing                                                                                                                                                                                                                   
  case in.Spec.AdmissionScope != nil:                                                                                                                                                                                                                                                     
      c.AdmissionFairSharing = &kueue.AdmissionFairSharing{Mode: in.Spec.AdmissionScope.AdmissionMode}                                                                                                                                                                                    
  default:                                                                                                                                                                                                                                                                                
      c.AdmissionFairSharing = nil
  }
```                                                                                                                                                                                                                                                                                       
                              
  But pkg/cache/queue/cluster_queue.go:189-193 does not fall back:                                                                                                                                                                                                                        
  
```
  func admissionFairSharingFromSpec(spec *kueue.ClusterQueueSpec) *kueue.AdmissionFairSharing {                                                                                                                                                                                           
      if spec.FairSharing != nil && spec.FairSharing.AdmissionFairSharing != nil {                                                                                                                                                                                                        
          return spec.FairSharing.AdmissionFairSharing                                                                                                                                                                                                                                    
      }                                                                                                                                                                                                                                                                                   
      return nil  // <-- silently ignores spec.AdmissionScope                                                                                                                                                                                                                             
  }
```                                                                                                                                                                                                                                                                                       
  
  This means existing ClusterQueues using only admissionScope will lose AFS functionality in the queue cache path while still appearing to work in the scheduler cache path. This should use the same fallback logic.

kannon92:
> What is our behavior if someone uses the deprecated field and the new field? I didn't see anything around validation to avoid this.

Thinking about this, I guess the deprecated field is in the v1beta1 API (so its already deprecated). There is a conversion webhook to convert this to v1beta2.

I'm not sure this comment is valid anymore.

apis/kueue/v1beta2/fairsharing_types.go
```diff
@@ -37,6 +39,24 @@ type FairSharing struct {
...
+// when participating in AdmissionFairSharing.
+type AdmissionFairSharing struct {
+	// mode indicates which mode for AdmissionFairSharing should be used.
+	// Possible values are:
+	// - UsageBasedAdmissionFairSharing
+	// - NoAdmissionFairSharing
+	//
+	// +kubebuilder:validation:Enum=UsageBasedAdmissionFairSharing;NoAdmissionFairSharing
+	// +required
+	Mode AdmissionMode `json:"mode,omitempty"`
```
kannon92, line 59:
```suggestion
	Mode AdmissionMode `json:"mode"`
```

andrewseif: I think the linter is failing when adding this change.
apis/kueue/v1beta2/fairsharing_types.go:59:2: requiredfields: field Mode should have the omitempty tag. (kubeapilinter)
	Mode AdmissionMode `json:"mode"`

pkg/cache/queue/cluster_queue.go
andrewseif: fixed, for some reason I didn't copy that branch.

kannon92 (commented):
/lgtm

/assign @mimowo @tenzen-y

kannon92:
can you squash your commits

mimowo:
I think we need KEP for that first, or at least KEP updates for the AdmissionFairSharing KEP and FairSharing.

mimowo:
Let's bring this topic to one of the wg-batch meetings

andrewseif:
@mimowo & @kannon92 I added the KEP, can you review again and confirm if this is fitting?

kannon92:
@andrewseif 
Thank you for your work on this. You did a great job on this PR.

https://github.com/kubernetes-sigs/kueue/issues/7710#issuecomment-4509488680

Unfortunately, we will not proceed with this PR at this time due comment above.

/close

PR #10255, opened 2026-04-01
Add fallback directory, when ARTIFACTS is not set, for xml junit repo…

### LINKED ISSUE #10243: when tests are run inside vs-code then xml files are created in the git versions area
**What happened**:

when tests are run inside vs-code then xml files are created in the git versions area

no xml files are created in the git versions area when running tests

run tests from vs-code

reruno:
…rt files

#### What type of PR is this?
/kind bug

Fixes #10243

```release-note
NONE
```

### DIFF
--- a/test/util/util.go
+++ b/test/util/util.go
@@ -18,6 +18,7 @@ package util
 
 import (
 	"bytes"
+	"cmp"
 	"context"
 	"errors"
 	"fmt"
@@ -115,10 +116,11 @@ func ConfigureSuiteReporting(report ginkgo.Report) error {
 		OmitFailureMessageAttr: true,
 	}
 	suiteName := uniqueSuiteName(report.SuitePath)
-	reportPath := "junit-" + suiteName + ".xml"
-	if dir := os.Getenv("ARTIFACTS"); dir != "" {
-		reportPath = filepath.Join(dir, reportPath)
+	reportDir := cmp.Or(os.Getenv("ARTIFACTS"), BinDir)
+	if err := os.MkdirAll(reportDir, 0o755); err != nil {
+		return fmt.Errorf("cannot create suite report directory %q: %w", reportDir, err)
 	}
+	reportPath := filepath.Join(reportDir, "junit-"+suiteName+".xml")
 	ginkgo.GinkgoLogr.Info(
 		"Generating JUnit report",
 		"path", reportPath,
--- a/test/util/constants.go
+++ b/test/util/constants.go
@@ -58,6 +58,7 @@ var (
 
 var (
 	ProjectBaseDir           = getProjectBaseDir()
+	BinDir                   = filepath.Join(ProjectBaseDir, "bin")
 	AutoscalerCrds           = filepath.Join(ProjectBaseDir, "dep-crds", "cluster-autoscaler")
 	JobsetCrds               = filepath.Join(ProjectBaseDir, "dep-crds", "jobset-operator")
 	TrainingOperatorCrds     = filepath.Join(ProjectBaseDir, "dep-crds", "training-operator-crds")


test/util/util.go
```diff
@@ -115,10 +115,16 @@ func ConfigureSuiteReporting(report ginkgo.Report) error {
 		OmitFailureMessageAttr: true,
 	}
 	suiteName := uniqueSuiteName(report.SuitePath)
-	reportPath := "junit-" + suiteName + ".xml"
+	reportDir := os.TempDir()
 	if dir := os.Getenv("ARTIFACTS"); dir != "" {
-		reportPath = filepath.Join(dir, reportPath)
+		reportDir = dir
+	} else if projectDir := getProjectBaseDir(); projectDir != "" {
```
mbobrovskyi: I think it’s unlikely that ProjectBaseDir is empty.

https://github.com/kubernetes-sigs/kueue/blob/f96c22bdb9a3f23dfb8bf2da90803bc9f5a5a03e/test/util/util.go#L1091-L1094

You can use the ProjectBaseDir variable in the else branch.

https://github.com/kubernetes-sigs/kueue/blob/f96c22bdb9a3f23dfb8bf2da90803bc9f5a5a03e/test/util/constants.go#L60

test/util/util.go
```diff
@@ -115,10 +115,14 @@ func ConfigureSuiteReporting(report ginkgo.Report) error {
 		OmitFailureMessageAttr: true,
 	}
 	suiteName := uniqueSuiteName(report.SuitePath)
-	reportPath := "junit-" + suiteName + ".xml"
+	reportDir := filepath.Join(ProjectBaseDir, "bin")
```
mbobrovskyi: I'm wondering if we can create similar variable (e.g. BinDir) in constants.go. And here just use it with cmp.Or().

test/util/util.go
```diff
@@ -115,10 +115,14 @@ func ConfigureSuiteReporting(report ginkgo.Report) error {
...
 	if dir := os.Getenv("ARTIFACTS"); dir != "" {
```
mimowo: What is the ARTIFACTS actually resolved to normally? is it the "artifacts/*"? If so then I would like to move out the files somewhere else, because we use the artifacts for releases currently, so that we don't mix. We could maybe consider two folders: "release-artifacts" and "artifacts" cc @mbobrovskyi

mimowo: The split into RELEASE_ARTIFACTS and ARTIFACTS is a potential follow up, don't do it here. Here just put it under "bin" probably.

mbobrovskyi: > What is the ARTIFACTS actually resolved to normally? is it the "artifacts/*"? If so then I would like to move out the files somewhere else, because we use the artifacts for releases currently, so that we don't mix. We could maybe consider two folders: "release-artifacts" and "artifacts" cc @mbobrovskyi

Yes, in tests prow use the artifacts/* directory. This is different from the release setup. Locally, it should be the bin/* directory.

mimowo: I see, so it would be great to align CI and local behavior long term, and put such auxiliary artifacts not in the artifacts/* folder. Let me open an issue for that.

reruno: In in local dev env ARTIFACTS is BIN_DIR
./kueue/Makefile
```
BIN_DIR ?= $(PROJECT_DIR)/bin
ARTIFACTS ?= $(BIN_DIR)
```
I assume kubernetes CI CD pipelines uses it for prow report generation, I will double check it

mbobrovskyi: No. Prow uses the ARTIFACTS variable internally to collect all results and copy them to the artifacts directory:
https://gcsweb.k8s.io/gcs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10255/pull-kueue-test-e2e-main-1-33/2039248765931491328/

I think we can’t change this behavior.

mbobrovskyi: But locally, yes – ARTIFACTS and BIN_DIR refer to the bin/* directory.

mimowo: https://github.com/kubernetes-sigs/kueue/issues/10256 - but for now bin/ sounds reasonable

mbobrovskyi:
> /milestone v0.17

Actually, wait – shouldn’t it be 0.18?

Issue #10189, opened 2026-03-30
MultiKueue Global Status - condition


Singularity23x0:
**What would you like to be added**:
A new condition type to be added to the WorkloadStatus of the Manager Cluster Workloads denoting and summarizing their status based on what is happening to them in the broader MultiKueue context (outside of just the Manager Cluster and its Cluster Queue).

**Why is this needed**:
The Workload status in queue in the Management Cluster must reflect the true state (Pending, Admitted, Running, Failed, Succeeded) derived from the remote cluster, including a human-readable message explaining the state (e.g., "Waiting for quota in cluster X").

**Completion requirements**:

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] Docs update

The artifacts will be linked in subsequent comments.

PR #10266, opened 2026-04-01
KEP-10189: MultiKueue Global Status Condition

Singularity23x0:
#### What type of PR is this?
/kind documentation

KEP for #10189.

It introduces the implementation proposal for the MultiKueue Global Status Condition to track the true status of a Manage Workload in the MultiKueue environment throughout its lifetime.

Addresses #10189

NONE

```release-note
NONE
```


keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,293 @@
...
+* SUCCESS: `Workload has finished successfully on worker cluster: <worker cluster reference>.`
+* FAILED: `Workload failed after admission on worker cluster: <worker cluster reference>.`
+* INACTIVE: `Workload inactive: <reason>.`
+* RUNNING: `Workload admitted on worker cluster: <worker cluster reference>.`
+* WORKER_SELECTED: `Workload received quota reservation on worker cluster: <worker cluster reference>. <number of ready admission checks>/<number of all admission checks> Admission Checks Ready.`
+* WAITING_FOR_WORKER:
+  * Default:
+  `Workload awaiting admission on one of the registered workers. <number of remotes> remotes created.`
+  * Non primary component workload:
+  `Component Workload awaiting admission on worker cluster: <worker nominated by the primary local workload>, nominated by <primary local workload reference>.`
```
kshalot: Should we make this distinction? Do the users even know what a "component workload" is?

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,293 @@
...
+Note: Local workload can be admitted (have the admitted condition), as described in #9338 (when MultiKueue admits a workload on a worker but then it gets preempted there with the MultiKueueRedoAdmissionOnEvictionInWorker gate enabled).
+* **WAITING_FOR_WORKER_NOMINATION** - specific to a non-primary component workload in the multi-workload-resource handling scenario;  quota reserved on the component workload; the component workload is waiting for the primary to nominate a worker to create a remote on,
+* **WAITING_FOR_QUOTA** - local workload is waiting to be granted a quota reservation on the Manager Cluster.
+
+For each MultiKueueGlobalStatus a message - **MultiKueueGlobalStatusMessage** - will be defined:
+* SUCCESS: `Workload has finished successfully on worker cluster: <worker cluster reference>.`
+* FAILED: `Workload failed after admission on worker cluster: <worker cluster reference>.`
+* INACTIVE: `Workload inactive: <reason>.`
+* RUNNING: `Workload admitted on worker cluster: <worker cluster reference>.`
```
kshalot: Doesn't this duplicate information that already exists?

Which makes me think if a condition like `AdmittedInWorker` that has the information about the MK replication process wouldn't "fit in" to the existing workload state rather than being an independent condition that contains everything.

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: We can omit the "component" keyword, but this distinction is important as the underlying logic is different to regular MK Workloads. We could name the job type instead of saying "compoinenet", e.g.
`LeaderWorkerSet Workload awaiting admission... nominated by Primary LeaderWorkerSet Workload <id>.`

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: Then, once this condition is no longer needed, we loose the information that this workload is an MK one.

The idea behind this condition is to make the workload status always be able to quickly identify an MK Workload and clearly relay what is happening to it. As running, failed and success are fundamentally different from what they are for an individual workload (e.g. finished-success for individual workload means the underlying job is finished, but it does not say whether it was finished locally or on a remote worker, nor which one specifically). The idea behind the differences of each of those statuses is expanded upon on the "Consolidated Workload State" section, focusing on contextual differences of each of those states.

While I agree that the "INACTIVE" condition could be contested, other ones do always carry different information to what it "duplicates", as the context of MK in and of itself expands the meaning of these states/messages.

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,293 @@
...
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Global Status Summary](#global-status-summary)
+  - [Visibility API](#visibility-api)
+  - [Consolidated Workload State](#consolidated-workload-state)
+<!-- /toc -->
+
+## Summary
+
+The Workload status in queue in the MultiKueue Manager Cluster must reflect the true state of the workload derived from the remote cluster,
```
olekzabl:
```suggestion
The Workload status in the MultiKueue Manager Cluster must reflect the true state of the workload derived from the Worker Clusters,
```

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,293 @@
...
+The condition:
+1. Will be populated for workloads created on the Manager Cluster.
+1. Will provide a high-level, human-readable message explaining the state of the Manager Workload.
+1. Will provide insights into the Workload's progress throughout its **whole lifecycle**.
+1. Will aggregate the information from all the Remote Workloads dispatched by MultiKueue to Worker Clusters for the subject Manager Workload.
+
+## Motivation
+
+Currently, the Workload Status subresource is missing key information aggregating data from across its remote counterparts on Worker Clusters.
+The existing conditions do not track the process of dispatching Remote Workloads to Workers, obstructing the details of MultiKueue's most critical part of the workload's execution.
```
olekzabl:
```suggestion
The existing conditions do not track the process of dispatching Remote Workloads to Workers, obfuscating the details of MultiKueue's most critical part of the workload's execution.
```

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,293 @@
...
+This forces users to search across all Worker Clusters to be able to see the big picture.
+
+Moreover, it does not natively support a contract defining a human-readable execution status.
+It instead relies on a list of Conditions and support information provided inside the WorkloadStatus field for the user to piece the actual global state of the underlying job's execution together on their own.
+
+To amend this, we need a way to present the user with a clearly defined, user-readable summary of the Global State, which aggregates information from across all the clusters of the MultiKueue environment.
+
+If nothing is implemented, the users will remain forced to rely on:
+* the conditions of the Manager Workload, which are misaligned and non-representative of all the extensive logic happening in MultiKueue,
+* manually querying and aggregating the conditions of Remote Workload across all Registered Workers; this covers a lot of distributed data; users are missing the core information of which Workers are eligible for distribution by virtue of being put forward by the Distribution Strategy.
```
olekzabl: What is "Distribution Strategy"? Did you mean "Dispatching"?

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Proposal
+
+The lifecycle of the Manager Workload will be split into the following **MultiKueueGlobalStatus** enumeration:
+* **SUCCESS** - workload finished successfully,
+* **FAILED** - workload finished with failed state,
+* **INACTIVE** - Manager Workload marked as inactive, preventing it from being scheduled; we can enter this state from any other as the workload can be deactivated both by Kueue and by the user,
+* **RUNNING** - local workload is admitted (has the admitted condition); a worker was nominated; the remote is admitted (has the admitted condition); the underlying job will attempt to execute; if it finishes we transition into the SUCCESS/FAILED state, otherwise we back-off,
+* **WORKER_SELECTED** - local workload is admitted (has the admitted condition); a single worker was nominated; the remote has currently received quota but is not admitted yet;
+this is possible if the MultiKueueWaitForWorkloadAdmitted feature is disabled; it can also occur as a consequence of graduating from the WAITING_FOR_WORKER state when the behavior described in  #9338  occurs,
```
olekzabl:
```suggestion
this is possible if the MultiKueueWaitForWorkloadAdmitted feature is disabled; it can also occur as a consequence of graduating from the WAITING_FOR_WORKER state when the behavior described in [#9338](https://github.com/kubernetes-sigs/kueue/issues/9338) occurs,
```

AFAICS (in the preview) these #ids don't become links automatically in KEPs.

And same below.

keps/10189-multikueue-global-status-condition/README.md
olekzabl: Nit: I'm feeling that this paragraph just repeats what has been already said above in the "Motivation" section. How much is it needed?

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,293 @@
...
+* Defining Status Summary describing data structures beyond what's necessary for implementing the new Workload Status Condition.
+
+## Proposal
+
+The lifecycle of the Manager Workload will be split into the following **MultiKueueGlobalStatus** enumeration:
+* **SUCCESS** - workload finished successfully,
+* **FAILED** - workload finished with failed state,
+* **INACTIVE** - Manager Workload marked as inactive, preventing it from being scheduled; we can enter this state from any other as the workload can be deactivated both by Kueue and by the user,
+* **RUNNING** - local workload is admitted (has the admitted condition); a worker was nominated; the remote is admitted (has the admitted condition); the underlying job will attempt to execute; if it finishes we transition into the SUCCESS/FAILED state, otherwise we back-off,
+* **WORKER_SELECTED** - local workload is admitted (has the admitted condition); a single worker was nominated; the remote has currently received quota but is not admitted yet;
```
olekzabl: Maybe it would be clearer to explicitly note here that our ultimate goal is to make this never happen?
(Yet, it'll continue to be possible for some time, due to bug fixes rolling out behind feature gates).

(Or do you think such an open claim is too strong / inaccurate?)

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,293 @@
...
+
+* Defining Status Summary describing data structures beyond what's necessary for implementing the new Workload Status Condition.
+
+## Proposal
+
+The lifecycle of the Manager Workload will be split into the following **MultiKueueGlobalStatus** enumeration:
+* **SUCCESS** - workload finished successfully,
+* **FAILED** - workload finished with failed state,
+* **INACTIVE** - Manager Workload marked as inactive, preventing it from being scheduled; we can enter this state from any other as the workload can be deactivated both by Kueue and by the user,
+* **RUNNING** - local workload is admitted (has the admitted condition); a worker was nominated; the remote is admitted (has the admitted condition); the underlying job will attempt to execute; if it finishes we transition into the SUCCESS/FAILED state, otherwise we back-off,
```
olekzabl: The phrase "admitted (has the admitted condition)" appears here in a few places, which I find surprising.
It seems you wish to avoid some ambiguity, but what kind of? And why with so many words?

How about just "local workload is Admitted"? By capitalizing, you'd explicitly refer to the standard Kueue concept.

keps/10189-multikueue-global-status-condition/README.md
kshalot: We wouldn't lose the `AdmittedInWorker` condition, it would simply remain `True` with the proposed "Workload admitted on worker cluster: <worker cluster reference>." message.

And for "success" - how does it differ from `WorkloadFinished`? It should be kept in sync between the local workload and the running remote so there's not really a distinction of "whether it was finished locally or on a remote worker", no?

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,293 @@
...
+* **WORKER_SELECTED** - local workload is admitted (has the admitted condition); a single worker was nominated; the remote has currently received quota but is not admitted yet;
+this is possible if the MultiKueueWaitForWorkloadAdmitted feature is disabled; it can also occur as a consequence of graduating from the WAITING_FOR_WORKER state when the behavior described in  #9338  occurs,
+* **WAITING_FOR_WORKER** -  quota reserved on the local workload; dispatching remotes to eligible workers and waiting  to nominate one of them; a worker will be nominated once the remote achieves a state allowing it to graduate to either the WORKER_SELECTED or the RUNNING state,
+Note: Local workload can be admitted (have the admitted condition), as described in #9338 (when MultiKueue admits a workload on a worker but then it gets preempted there with the MultiKueueRedoAdmissionOnEvictionInWorker gate enabled).
+* **WAITING_FOR_WORKER_NOMINATION** - specific to a non-primary component workload in the multi-workload-resource handling scenario;  quota reserved on the component workload; the component workload is waiting for the primary to nominate a worker to create a remote on,
```
olekzabl: Following up on [this comment](https://github.com/kubernetes-sigs/kueue/pull/10266/changes#r3049824311), "component workload" might be unclear even for Kueue developers; e.g. it seems not to appear in any earlier KEP. 🫤 

To clarify this, I'd propose to link e.g. [here](https://github.com/kubernetes-sigs/kueue/pull/8670#discussion_r2745488245); at least that's the nicest target I could find.

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,293 @@
...
+* **FAILED** - workload finished with failed state,
+* **INACTIVE** - Manager Workload marked as inactive, preventing it from being scheduled; we can enter this state from any other as the workload can be deactivated both by Kueue and by the user,
+* **RUNNING** - local workload is admitted (has the admitted condition); a worker was nominated; the remote is admitted (has the admitted condition); the underlying job will attempt to execute; if it finishes we transition into the SUCCESS/FAILED state, otherwise we back-off,
+* **WORKER_SELECTED** - local workload is admitted (has the admitted condition); a single worker was nominated; the remote has currently received quota but is not admitted yet;
+this is possible if the MultiKueueWaitForWorkloadAdmitted feature is disabled; it can also occur as a consequence of graduating from the WAITING_FOR_WORKER state when the behavior described in  #9338  occurs,
+* **WAITING_FOR_WORKER** -  quota reserved on the local workload; dispatching remotes to eligible workers and waiting  to nominate one of them; a worker will be nominated once the remote achieves a state allowing it to graduate to either the WORKER_SELECTED or the RUNNING state,
```
olekzabl: "and waiting to nominate one of them" - is this accurate?

* If it is, then we have a problem with names of states (because `WAITING_FOR_WORKER_NOMINATION` is another state, not this one)
* However, I think it isn't accurate. I think we just hit a _clash of notions_; see below.

AFAICS the concept of "nominating" is used in Kueue in **two different meanings**:

* One meaning refers to MK dispatcher _nominating_ worker clusters for dispatching remote workloads (see [MK KEP](https://github.com/kubernetes-sigs/kueue/tree/main/keps/693-multikueue); also [KEP-8303](https://github.com/kubernetes-sigs/kueue/tree/main/keps/8303-multikueue-orchestrated-preemption)).
  In that sense, we're no longer waiting for nomination here; nomination already happened.
  I think this is the meaning you used in the names of proposed states.
  I also think this is the dominant meaning, at least in recent months.

* The other meaning is mostly found in the scheduler code (e.g. [here](https://github.com/kubernetes-sigs/kueue/blob/b62584120f7f94cc01a2ed8d3afc5b340ec03a19/pkg/scheduler/scheduler.go#L478)) and also mentioned in some KEPs but just briefly (AFAICS [here](https://github.com/kubernetes-sigs/kueue/blob/b62584120f7f94cc01a2ed8d3afc5b340ec03a19/keps/349-all-or-nothing/README.md?plain=1#L534) and [here](https://github.com/kubernetes-sigs/kueue/blob/b62584120f7f94cc01a2ed8d3afc5b340ec03a19/keps/420-partial-admission/README.md?plain=1#L86)).
  I think this is the meaning you used when writing the descriptions of states here. (Not only this one; look around).

Mixing both meanings leads to confusion.
I propose to:

* Remove all usages of "nominate" in "the scheduler meaning".
* Link to MK KEP (e.g. [here](https://github.com/kubernetes-sigs/kueue/tree/main/keps/693-multikueue#multikueue-dispatcher-api)) to clarify what "nomination" means in this KEP.

In particular, for this bullet, I'd propose this:

```suggestion
* **WAITING_FOR_WORKER** -  quota reserved on the local workload; dispatching remotes to eligible workers and waiting for one of them to receive quota (upon which we'll graduate to either the WORKER_SELECTED or the RUNNING state),
```

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,293 @@
...
+Note: Local workload can be admitted (have the admitted condition), as described in #9338 (when MultiKueue admits a workload on a worker but then it gets preempted there with the MultiKueueRedoAdmissionOnEvictionInWorker gate enabled).
+* **WAITING_FOR_WORKER_NOMINATION** - specific to a non-primary component workload in the multi-workload-resource handling scenario;  quota reserved on the component workload; the component workload is waiting for the primary to nominate a worker to create a remote on,
+* **WAITING_FOR_QUOTA** - local workload is waiting to be granted a quota reservation on the Manager Cluster.
```
olekzabl: Looking at it now, wouldn't `WAITING_FOR_MANAGER_QUOTA` be clearer?

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,293 @@
...
+|Message|MultiKueueGlobalStatusMessage|
+
+The condition will be gated behind the **MultiKueueGlobalStatusCondition** feature gate.
+
+### User Stories
+
+#### Story 1
+
+I want to get an idea of when my job will be admitted. Has it already received quota on the manager?
+Is it being dispatched? If yes - what are the workers has it been dispatched to, so that I can check the status of some or all of them for how my job is doing there?
```
olekzabl:
```suggestion
Is it being dispatched? If yes - what are the workers it has been dispatched to, so that I can check the status of some or all of them for how my job is doing there?
```

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,293 @@
...
+#### Story 2
+
+My job hasn’t been admitted for a long time. I would like to verify why that is.
+Are remotes being created? Have they been dispatched to all workers? What is their status there? Are they waiting for quota? Are they waiting for admission checks?
+
+### Notes/Constraints/Caveats
+
+This proposal is limited to only providing the user with a high-level summary.
+This is a minimum value proposition, lacking in a few key areas:
+* In the WAITING_FOR_WORKER state we only provide a miniscule summary of what is happening on the Workers. In reality, this state is much more complex than the messages we propose can describe and could greatly benefiot form being accompanied by a set of aggregations describing in greater detail what is the status of each Remote Workload.
```
olekzabl:
```suggestion
* In the WAITING_FOR_WORKER state we only provide a miniscule summary of what is happening on the Workers. In reality, this state is much more complex than the messages we propose can describe and could greatly benefit form being accompanied by a set of aggregations describing in greater detail what is the status of each Remote Workload.
```

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,293 @@
...
+* In an effort ot keep the condition message concise, we ware limited in how much data relevant to the user we can provide.
```
olekzabl:
```suggestion
* In an effort to keep the condition message concise, we are limited in how much data relevant to the user we can provide.
```

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: Still AdmittedInWorker would only cover the "Running" status.

While I agree that removing the INACTIVE, SUCCESS and FAILED would not make us loose data, we still need the previous state descriptions to remain in place. With that in mind, the RUNNING status could be renamed to ADMITTED_ON_WORKER.

One caveat here is that once we set the Condition to ADMITTED_ON_WORKER, the user does not immediately know if the workload has finished or not, requiring them to seek for that information further.

As for removing the INACTIVE status - what would you propose happen to this Condition when the workload becomes inactive? Should we remove it? Should we set it to false? In either case we would have to rename the condition. And renaming it to AdmittedInWorker would not work because of the other statuses to account for (ones preceding the RUNNING status).

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: Yes, Dispatch Strategy, correcting.

Singularity23x0: I'd say this portion is more direct in describing the struggles implied by the previous motivation section contents. It does not simply reiterate, but rather clarifies the iplications.

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: The ambiguity is that, from the implementation's perspective, QuotaReserved means being granted an admission with a flavor and a reservation, whereas being Admitted is further specified by bing granted the admitted condition. I wanted to mention this explicitly to avoid confusion between having an admission and having the admitted condition.

Singularity23x0: I am uncertain if using the capitalized "Admitted" is enough, but if the sentiment is shared by others I can introduce the change.

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: I can't find any information on whether/when the MultiKueueWaitForWorkloadAdmitted feature gate will be phased out.
9338 has been addressed with a PR incorporated into releases 0.16 and 0.17, so this part of the paragraph can be removed.

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: Switched from using "nominated" to "selected".

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: For now opted to use the "multi-workload resource type name" Workload (e.g. LeaderWorkerSet Workload) naming scheme - please let me know your thoughts.

keps/10189-multikueue-global-status-condition/README.md
kshalot: `AdmittedInWorker` can cover statuses other than running, because it could be set to `False` with a the reasons proposed in this KEP (I'm just spitballing here).

As for inactive, what happens in the current implementation? Because we also have the `WorkloadDeactivated` condition.

Singularity23x0: Yes, but then we give the AdmittedInWorker=False a different meaning then when AdmittedInWorker is not present (i.e. mk workload vs non-mk workload). I would prefer keeping "missing" to mean "false".

Still the main issue with using AdmittedInWorker is immediate readability:
|MultiKueueWorkload|AdmittedInWorker|
|---|---|
|Immediately lets the user know the workload is an MK one|Requires user to know the context behind the condition|
|Allows identifying the state of the workload immediately|Requires checking other conditions to determine if wl has finished and with what result|
|Has only two legal states: present (true) or missing (false)|Has 3 states: present (true), present (false) and missing (not even an MK workload)|

kshalot: > Requires user to know the context behind the condition

Even if this is an issue, it could be solved with proper naming, reasons and messages.

---

> Requires checking other conditions to determine if wl has finished and with what result

My point is that you could just as well check the `Finished` condition, which would read sth like:
```
    Message:               Reached expected number of succeeded pods
    Reason:                Succeeded
    Status:                True
    Type:                  Finished
```
I'm not sure how this is less user friendly than reading the `MultiKueueWorkload` condition. If anything it's less convenient, because you don't have a single place where this information is exposed, but two. Which are also dependent on whether this is a MK workload or not.

---

> Has 3 states: present (true), present (false) and missing (not even an MK workload)

I'm not yet convinced this is an issue, but I'll think on it.

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,293 @@
...
+The condition:
+1. Will be populated for workloads created on the Manager Cluster.
+2. Will provide a high-level, human-readable message explaining the state of the Manager Workload.
+3. Will provide insights into the Workload's progress throughout its **whole lifecycle**.
+4. Will aggregate information from all the Remote Workloads dispatched by MultiKueue to Worker Clusters for the subject Manager Workload.
+
+## Motivation
+
+Currently, the Workload status subresource is missing key information aggregating data from its remote counterparts on Worker Clusters.
+The existing conditions do not track the process of dispatching Remote Workloads to Workers, obfuscating the details of the most critical part of MultiKueue wWrkload's execution.
```
olekzabl: typo: wWrkload

keps/10189-multikueue-global-status-condition/README.md
olekzabl: So if the second part is clearer than the first one (which I'd agree with), then maybe we don't need the first one?

How about e.g. this (which is just reshuffling of some of your sentences):

_Currently, the Workload status subresource is missing key information aggregating data from its remote counterparts on Worker Clusters. The users are forced to rely on:_

* _the conditions of the Manager Workload, which are misaligned and non-representative of the extensive logic in MultiKueue,_
* _manually querying and aggregating the conditions of Remote Workloads across all registered workers; this covers a lot of distributed data, and users lack core information about which workers are eligible for dispatch by virtue of being put forward by the dispatch strategy._

_To address this, we need a way to present users with a clearly defined, human-readable summary of the global state, which aggregates information from all clusters in the MultiKueue environment._

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: 1. Agreed.
2. Id argue that in that case it is better to have redundancy making both Conditions mention the workload is MK and Finished rather than modifying the MK Cond.

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,293 @@
...
+  // MultiKueue is currently dispatching remote workloads to eligible Workers.
+  WaitingForWorker = "WAITING_FOR_WORKER"
+
+  // WaitingForWorkerNomination state is specific to a non-primary component workload in the multi-workload-resource handling scenario.
+  // It means the component workload has received quota reservation on the Manager
+  // and is waiting for the primary component workload to select a worker to dispatch a remote to.
+  WaitingForWorkerNomination = "WAITING_FOR_WORKER_NOMINATION"
+
+  // WaitingForQuota state means the workload is currently in the "Pending" state on the Manager Cluster (does not currently hold any quota reservations).
+  WaitingForQuota = "WAITING_FOR_MANAGER_QUOTA"
```
olekzabl:
```suggestion
  // WaitingForManagerQuota state means the workload is currently in the "Pending" state on the Manager Cluster (does not currently hold any quota reservations).
  WaitingForManagerQuota = "WAITING_FOR_MANAGER_QUOTA"
```

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: Mentioned in an alternative - ptal.

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,326 @@
...
+| WAITING FOR WORKER NOMINATION | Local workload is a non-primary workload in a group of composite workloads. Local has got an admission but not the admitted condition. Primary has not selected a worker yet. | — | WAITING_FOR_WORKER_NOMINATION | — |
+| PENDING | Local workload does not have an admission. No remotes exist. | Workload does not have an admission. | WAITING_FOR_MANAGER_QUOTA | Pending |
+
+This consolidated set of statuses could then be used to populate a new, stand-alone field in the WorkloadStatus, allowing for a generic solution providing a high-level status summary for both individual and MultiKueue workloads.
+
+The main caveat here is that this approach risks being confusing to users, as the Workload States shift meanings significantly depending on whether the Workload is a Manager Workload or an Individual (remote or non-MultiKueue) Workload.
+
+### Granular Conditions
+
+Instead of defining a single Condition, we could provide one or more Granular Conditions, each servicing the purpose of supplying data missing from the WorkloadStatus, allowing to piece the Global Status of a MultiKueue workload by the user on their own.
```
olekzabl:
```suggestion
Instead of defining a single Condition, we could provide one or more Granular Conditions, each serving the purpose of supplying data missing from the WorkloadStatus, allowing to piece the Global Status of a MultiKueue workload by the user on their own.
```

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,326 @@
...
+* QuotaReservedInWorker - set to true when in the WORKER_SELECTED status. The message would identify the worker.
+* WorkerNominated - set to False when no worker was nominated for dispatch yet. Set to True once a single worker is nominated. This can mean entering the RUNNING or WORKER_SELECTED states, as well as cover exiting the WAITING_FOR_WORKER_NOMINATION state in case of non-primary composite workloads. The reason and message would identify which of these 3 cases we encountered.
+* MultiKueueManagerWorkload - present and set to True only for MultiKueue Manager Workloads, ensuring we are always aware if the workload in question is a Manager Workload or not.
+
+#### Disadvantages
+
+The main issue here, is that while we ensure the status covers all multi-kueue cases, reading the state of the job's execution remains complex. We still have to read an understand a multitude of conditions and status information to piece together the inner workings of MultiKueue, isntead of being provided with a clear, human-readable summary. This is especially apparrent when considering the latter variant, where the number of added conditions expands to 3 or 4.
```
olekzabl:
```suggestion
The main issue here, is that while we ensure the status covers all MultiKueue cases, reading the state of the job's execution remains complex. We still have to read an understand a multitude of conditions and status information to piece together the inner workings of MultiKueue, isntead of being provided with a clear, human-readable summary. This is especially apparrent when considering the latter variant, where the number of added conditions expands to 3 or 4.
```

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,326 @@
...
+
+Unit tests will be added for the methods identifying the MultiKueueGlobalStatus and calculating the condition as part of the [MultiKueue Workload Controller](https://github.com/Singularity23x0/kueue/blob/66c2cf1a93ee3f19d8b29309d27832a4595d4106/pkg/controller/admissionchecks/multikueue/workload_test.go#L17) test suite.
+
+#### Integration tests
+
+Integration tests for assigning the expected condition correctly for each Global Status will be added in a new file as part of the [MultiKueue Integration Test Suite](https://github.com/Singularity23x0/kueue/blob/66c2cf1a93ee3f19d8b29309d27832a4595d4106/test/integration/multikueue/suite_test.go#L93).
+
+#### e2e tests
+
+Integration tests for assigning the expected condition correctly for each Global Status will be added in a new file as part of the [MultiKueue E2E Test Suite](https://github.com/Singularity23x0/kueue/blob/66c2cf1a93ee3f19d8b29309d27832a4595d4106/test/e2e/multikueue/suite_test.go#L62).
```
mszadkow: `Integration` -> `End to end`

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -290,4 +287,40 @@ The proposed set of states would be as follows:
...
+|Corresponding MultiKueueGlobalStatus|AdmittedInWorker Condition Value|
+|---|---|
+|SUCCESS|nil|
+|FAILED|nil|
+|INACTIVE|**State**: False, **Reason**: WorkloadInactive|
+|RUNNING|**State**: True|
+|WORKER_SELECTED|**State**: False|
+|WAITING_FOR_WORKER|**State**: False|
+|WAITING_FOR_WORKER_NOMINATION|**State**: False|
+|WAITING_FOR_MANAGER_QUOTA|**State**: False|
```
olekzabl, line 309: There's one thing in this table that makes me confused: 

Would you set "Reason" _only_ in the INACTIVE case?

My understanding from our discussion was that (while we may choose to provide the Reason in Inactive case, just for clarity) the other "False" cases (Worker Selected, Waiting for Worker etc.) would need the Reason _even more_, as these are the cases that can't be distinguished based on currently existing manager-side Conditions. Or am I getting it wrong?

Also, the text below "Reason and Message values can be copied directly in all cases but the INACTIVE state one" doesn't help me. The table currently seems to copy Reason for INACTIVE, not for others :)

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,326 @@
...
+* QuotaReservedInWorker - set to true when in the WORKER_SELECTED status. The message would identify the worker.
+* WorkerNominated - set to False when no worker was nominated for dispatch yet. Set to True once a single worker is nominated. This can mean entering the RUNNING or WORKER_SELECTED states, as well as cover exiting the WAITING_FOR_WORKER_NOMINATION state in case of non-primary composite workloads. The reason and message would identify which of these 3 cases we encountered.
+* MultiKueueManagerWorkload - present and set to True only for MultiKueue Manager Workloads, ensuring we are always aware if the workload in question is a Manager Workload or not.
+
+#### Disadvantages
+
+The main issue here, is that while we ensure the status covers all MultiKueue cases, reading the state of the job's execution remains complex. We still have to read an understand a multitude of conditions and status information to piece together the inner workings of MultiKueue, isntead of being provided with a clear, human-readable summary. This is especially apparrent when considering the latter variant, where the number of added conditions expands to 3 or 4.
```
olekzabl, line 326:
```suggestion
The main issue here is that, while we ensure the status covers all MultiKueue cases, reading the state of the job's execution remains complex. We still have to read and understand a multitude of conditions and status information to piece together the inner workings of MultiKueue, isntead of being provided with a clear, human-readable summary. This is especially apparent when considering the latter variant, where the number of added conditions expands to 3 or 4.
```

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: The idea is that if reason an message are not provided in the table, they are the same as in the core proposal.

worktree1
kshalot, line 1: Committed by mistake.

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,328 @@
...
+
+The condition will be gated behind the **MultiKueueGlobalStatusCondition** feature gate.
+
+### User Stories
+
+#### Story 1
+
+I want to get an idea of when my job will be admitted. Has it already received quota on the manager?
+Is it being dispatched? If yes - what are the workers it has been dispatched to, so that I can check the status of some or all of them for how my job is doing there?
+Or maybe it is already running on a specific worker? If yes - which one?
```
kshalot: I think all those questions can be answered already, even without the proposed condition. So maybe this story could put more emphasis on the centralization of this information, if that's the goal.

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,328 @@
...
+
+### Test Plan
+
+[ ] I/we understand the owners of the involved components may require updates to
+existing tests to make this code solid enough prior to committing the changes necessary
+to implement this enhancement.
+
+#### Unit tests
+
+Unit tests will be added for the methods identifying the MultiKueueGlobalStatus and calculating the condition as part of the [MultiKueue Workload Controller](https://github.com/Singularity23x0/kueue/blob/66c2cf1a93ee3f19d8b29309d27832a4595d4106/pkg/controller/admissionchecks/multikueue/workload_test.go#L17) test suite.
```
kshalot: You linked to your fork repo. I'd change it to the kubernetes-sigs one.

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: Expanded - ptal.

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,333 @@
...
+The main risk of the feature is its effect on the MultiKueue Reconciler's performance and code complexity.
+Implementing the condition calculating logic efficiently will require careful consideration of where the Reconciler can reliably determine the state of the Manager Workload.
+That may necessitate distributing pieces of logic across functions, complicating the already convoluted logic. Mitigating that will require careful naming and code grouping in the implementation phase.
+
+Another issue stems from the MultiKueue's complex logic itself. It may prove difficult to reliably track the actual state due to the branching nature of the functions.
+Countering that will require thorough test coverage, with a special emphasis on implementing a wide coverage in E2E tests.
+
+## Design Details
+
+The new condition type will be defined alongside existing ones in the [workload types file](https://github.com/kubernetes-sigs/kueue/blob/9e20f44bb05758976744a5ee6aa77309677bb22f/apis/kueue/v1beta1/workload_types.go#L628).
```
tenzen-y: Could you replace this link with v1beta2?
If you are aiming to add these APIs only to v1beta1, please let me know. In that case, we should discuss API migration and graduation.

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,333 @@
...
+  // - WAITING_FOR_WORKER,
+  // - WAITING_FOR_WORKER_NOMINATION,
+  // - WAITING_FOR_MANAGER_QUOTA.
+  MultiKueueWorkload = "MultiKueueWorkload"
+
+  // ...
+)
+```
+
+The MultiKueueGlobalStatus enumeration will be defined in the [MultiKueue types file](https://github.com/kubernetes-sigs/kueue/blob/9e20f44bb05758976744a5ee6aa77309677bb22f/apis/kueue/v1beta1/multikueue_types.go#L17).
```
tenzen-y: ditto about link

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,333 @@
...
+  WaitingForWorkerNomination = "WAITING_FOR_WORKER_NOMINATION"
+
+  // WaitingForManagerQuota state means the workload is currently in the "Pending" state on the Manager Cluster (does not currently hold any quota reservations).
+  WaitingForManagerQuota = "WAITING_FOR_MANAGER_QUOTA"
+)
+
+```
+
+The condition will be populated inside the [Reconciler of the MultiKueue Core Controller](https://github.com/kubernetes-sigs/kueue/blob/9e20f44bb05758976744a5ee6aa77309677bb22f/pkg/controller/admissionchecks/multikueue/workload.go#L158).
+This reconciler already gathers all the necessary data in the form of the [Workload Group](https://github.com/kubernetes-sigs/kueue/blob/9e20f44bb05758976744a5ee6aa77309677bb22f/pkg/controller/admissionchecks/multikueue/workload.go#L83) internal structure.
```
tenzen-y:
```suggestion
The condition will be populated inside the Reconciler of the MultiKueue Core Controller.
This reconciler already gathers all the necessary data in the form of the Workload Group internal structure.
```

Could you avoid attaching the code base? Because the implementations will be drastically changed in the future, but this documentation will be kept even in the future.

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,333 @@
...
+
+### Test Plan
+
+[ ] I/we understand the owners of the involved components may require updates to
+existing tests to make this code solid enough prior to committing the changes necessary
+to implement this enhancement.
+
+#### Unit tests
+
+Unit tests will be added for the methods identifying the MultiKueueGlobalStatus and calculating the condition as part of the [MultiKueue Workload Controller](https://github.com/kubernetes-sigs/kueue/blob/9e20f44bb05758976744a5ee6aa77309677bb22f/pkg/controller/admissionchecks/multikueue/workload_test.go#L17) test suite.
```
tenzen-y:
```suggestion
Unit tests will be added for the methods identifying the MultiKueueGlobalStatus and calculating the condition as part of the MultiKueue Workload Controller test suite.
```

ditto about code attachment

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,333 @@
...
+
+#### Integration tests
+
+Integration tests for assigning the expected condition correctly for each Global Status will be added in a new file as part of the [MultiKueue Integration Test Suite](https://github.com/kubernetes-sigs/kueue/blob/9e20f44bb05758976744a5ee6aa77309677bb22f/test/integration/multikueue/suite_test.go#L93).
```
tenzen-y:
```suggestion
Integration tests for assigning the expected condition correctly for each Global Status will be added in a new file as part of the MultiKueue Integration Test Suite.
```

ditto about code attachment

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,333 @@
...
+
+#### e2e tests
+
+E2E Tests for assigning the expected condition correctly for each Global Status will be added in a new file as part of the [MultiKueue E2E Test Suite](https://github.com/kubernetes-sigs/kueue/blob/9e20f44bb05758976744a5ee6aa77309677bb22f/test/e2e/multikueue/suite_test.go#L62).
```
tenzen-y:
```suggestion
E2E Tests for assigning the expected condition correctly for each Global Status will be added in a new file as part of the MultiKueue E2E Test Suite.
```

ditto about code attachment.

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,333 @@
...
+
+### Non-Goals
+
+* Defining Status Summary describing data structures beyond what's necessary for implementing the new Workload Status Condition.
+
+## Proposal
+
+The lifecycle of the Manager Workload will be split into the following **MultiKueueGlobalStatus** enumeration:
+* **SUCCESS** - workload finished successfully,
+* **FAILED** - workload finished with failed state,
```
tenzen-y, line 67: 1. Why do we need these reasons even though the Workload already has the duplicated Finished Workload condition?
2. How to keep consistency between the Finished condition type and the new ones?

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,333 @@
...
+* **INACTIVE** - Manager Workload marked as inactive, preventing it from being scheduled; we can enter this state from any other as the workload can be deactivated both by Kueue and by the user,
+* **RUNNING** - local workload is admitted (has the admitted condition); a single worker was selected; the remote is admitted (has the admitted condition); the underlying job will attempt to execute; if it finishes we transition into the SUCCESS/FAILED state, otherwise we back-off,
```
tenzen-y, line 69: This intermediate condition is very dangerous because it could cause the statue handling bug.
I strongly wouldn't recommend this reason. Especially, it's very challenging to define what " Running " is.

Additionally, what is the difference between "Admitted" and "Running" condition on the manager cluster?
AFAIK, the Admitted condition on the worker-cluster will be populated to the manager cluster, then the manager cluster will add "Admitted" condition to the Workload on the manager cluster after AdmissionCheck=Ready state.

I understand that this doesn't indicate the underlying Pods are running in the nominated worker cluster.
But, in general, the "Running" is estimated to indicate that actual application processes are executed.

I see, this indicates the consolidated condition reason based on Admitted + `clusterName != ""`.
So, instead of "Running", I would propose "AdmittedOnWorkerCluster" as an alternative.

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,333 @@
...
+* **WORKER_SELECTED** - local workload is admitted (has the admitted condition); a single worker was selected; the remote has currently received quota but is not admitted yet;
```
tenzen-y, line 70: Why do we need this? This is obviously duplicated with `.status.clusterName`.

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,333 @@
...
+this is possible if the MultiKueueWaitForWorkloadAdmitted feature is disabled,
+* **WAITING_FOR_WORKER** -  quota reserved on the local workload; dispatching remotes to nominated workers and waiting for one of them to be selected - a worker will be selected once the remote achieves a state allowing it to graduate to either the WORKER_SELECTED (receiving quota reservation, MultiKueueWaitForWorkloadAdmitted feature gate disabled) or the RUNNING (workload receives the admitted condition set to true) state,
```
tenzen-y, line 72: Why do we need this? This is obviously deplicated with `len(.status.nominatedClusterNames) != 0 && .status.clusterName == ""`.

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,333 @@
...
+* **WAITING_FOR_WORKER_NOMINATION** - specific to a non-primary component workload in the multi-workload-resource handling scenario;  quota reserved on the component workload; the component workload is waiting for the primary to select a worker to create a remote on,
```
tenzen-y, line 73: Why do we need this? Isn't this duplicated with `len(.status.nominatedClusterNames) == 0 `?
Also, what is "component workload" ?

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,333 @@
...
+  // - WAITING_FOR_MANAGER_QUOTA.
+  MultiKueueWorkload = "MultiKueueWorkload"
+
+  // ...
+)
+```
+
+The MultiKueueGlobalStatus enumeration will be defined in the [MultiKueue types file](https://github.com/kubernetes-sigs/kueue/blob/9e20f44bb05758976744a5ee6aa77309677bb22f/apis/kueue/v1beta1/multikueue_types.go#L17).
+
+```go
```
tenzen-y, line 228: When we define reason, we should use CamelCase. For example, the "WORKER_SELECTED" should be "WorkerSelected".

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,333 @@
...
+* **WORKER_SELECTED** - local workload is admitted (has the admitted condition); a single worker was selected; the remote has currently received quota but is not admitted yet;
+this is possible if the MultiKueueWaitForWorkloadAdmitted feature is disabled,
+* **WAITING_FOR_WORKER** -  quota reserved on the local workload; dispatching remotes to nominated workers and waiting for one of them to be selected - a worker will be selected once the remote achieves a state allowing it to graduate to either the WORKER_SELECTED (receiving quota reservation, MultiKueueWaitForWorkloadAdmitted feature gate disabled) or the RUNNING (workload receives the admitted condition set to true) state,
+* **WAITING_FOR_WORKER_NOMINATION** - specific to a non-primary component workload in the multi-workload-resource handling scenario;  quota reserved on the component workload; the component workload is waiting for the primary to select a worker to create a remote on,
+* **WAITING_FOR_MANAGER_QUOTA** - local workload is waiting to be granted a quota reservation on the Manager Cluster.
```
tenzen-y, line 74: Why do we need this? It's duplicated with "QuotaReserved" condition type.
What is the difference?

keps/10189-multikueue-global-status-condition/README.md
```diff
@@ -0,0 +1,333 @@
...
+### Risks and Mitigations
+
+The main risk of the feature is its effect on the MultiKueue Reconciler's performance and code complexity.
+Implementing the condition calculating logic efficiently will require careful consideration of where the Reconciler can reliably determine the state of the Manager Workload.
+That may necessitate distributing pieces of logic across functions, complicating the already convoluted logic. Mitigating that will require careful naming and code grouping in the implementation phase.
+
+Another issue stems from the MultiKueue's complex logic itself. It may prove difficult to reliably track the actual state due to the branching nature of the functions.
+Countering that will require thorough test coverage, with a special emphasis on implementing a wide coverage in E2E tests.
+
+## Design Details
```
tenzen-y, line 202: @Singularity23x0 I would request to add the state transition diagram either by image or a marmaid graph.
Because we should evaluate Reasons if there are any buggy paths or duplicated and unnecessary reasons.

keps/10189-multikueue-global-status-condition/kep.yaml
```diff
@@ -1,20 +1,13 @@
...
 status: implementable
-creation-date: 2026-01-21
+creation-date: 2026-03-30
 reviewers:
-  - "@mimowo"
-  - "@ichekrygin"
-  - "@pajakd"
-  - "@tenzen-y"
-  - "@olekzabl"
+  - TBD
```
tenzen-y:
```suggestion
  - @kshalot 
  - @olekzabl 
  - @tenzen-y 
```

keps/10189-multikueue-global-status-condition/kep.yaml
```diff
@@ -1,20 +1,13 @@
...
 approvers:
-  - "@mimowo"
-  - "@tenzen-y"
-  
+  - TBD
```
tenzen-y:
```suggestion
  - @tenzen-y 
```

keps/10189-multikueue-global-status-condition/kep.yaml
```diff
@@ -27,11 +20,15 @@ latest-milestone: "v0.18"
...
+  beta: "v0.19"
+  stable: "v0.20"
 
 # The following PRR answers are required at alpha release
 # List the feature gate name and the components for which it must be enabled
 feature-gates:
-  - name: ConcurrentAdmission
-  
\ No newline at end of file
+  - name: MultiKueueGlobalStatusCondition
```
tenzen-y, line 31: What is the primary responsibility for this FG?
If we disable this FG, what happens?

keps/10189-multikueue-global-status-condition/kep.yaml
```diff
@@ -27,11 +20,15 @@ latest-milestone: "v0.18"
 # The milestone at which this feature was, or is targeted to be, at each stage.
 milestone:
   alpha: "v0.18"
-#   beta: "v0.3"
-#   stable: "v0.5"
+  beta: "v0.19"
```
tenzen-y, line 25: I'm wondering if we might need to start from Beat because APIs are exposed and cannot be guarded by FGs.

But, let's discuss after we finalize almost all the designs.

tenzen-y (commented):
Thank you for working on this KEP. I left the first comments.

keps/10189-multikueue-global-status-condition/kep.yaml
Singularity23x0: When disabled, the condition will not be calculated by the MK Wl Reconciler, nor populated for any Workload.

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: Added - ptal.

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: 1. I initially addressed this matter in the "Granular Conditions" alternatives section. TLDR: the idea is to provide this information via a unified, human-readable message.
2. The new condition will be populated in the MultiKueue Core Workload Reconciler - there we can ensure that if the finished condition is present and true, our new condition is compliant.

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: The way the "Admitted" Condition works on the Manager is not based on the Worker Workload directly.
The manager workload sets Admitted=True when it gets quota reservation on the manager and all the ACs are Ready. The MultiKueue AC will be Ready either when we get Admitted or get QuotaReserved on one of the Workers, depending on whether MultiKueueWaitForWorkloadAdmitted is enabled.

I wanted to avoid using "ADMITTED" so as not to confuse it with the Admitted=True condition.

The split between WORKER_SELECTED and RUNNING stems form the MultiKueueWaitForWorkloadAdmitted feature, where we can have Admitted=True and `clusterName != ""`, but:
* MultiKueueWaitForWorkloadAdmitted DISABLED - the Remote is not yet admitted, so we say WORKER_SELECTED,
* MultiKueueWaitForWorkloadAdmitted ENABLED or we were in WORKER_SELECTED  and now the Remote Workload is actually Admitted on the Worker.

Renaming to "ADMITTED_ON_WORKER_CLUSTER" or "ADMITTED_ON_WORKER" sounds like an ok alternative.

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: To differentiate between being Pending, with Quota Reserved or Admitted on the worker cluster.
When clusterName is set:
* WAITING_FOR_WORKER means we are still pending on the worker,
* WORKER_SELECTED means we have quota but are not yet Admitted there,
* RUNNING means we are admitted on the Worker.

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: To expand on 1: While the data is available, it requires a lot of analysis to gather into a simple, readable message. The core idea behind this Condition is to provide users with such a simple, readable message to quickly summarize the state of the workload.

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: Aside from that, the core idea of providing a unified, human readable summary of the status applies here.
See: https://github.com/kubernetes-sigs/kueue/pull/10266#discussion_r3117908206

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: See: https://github.com/kubernetes-sigs/kueue/pull/10266#discussion_r3117908206

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: Why do we need this: see https://github.com/kubernetes-sigs/kueue/pull/10266#discussion_r3117908206

Component workloads denotes all workloads created by multiworkload resources (for example LeaderWorkerSet). They are handled by a MultiWorkloadAdapted in MultiKueue. TLDR:
* multiple workloads are created for such a multiworklaod resource,
* one of the workloads is denoted as Primary,
* the primary works like any other mk workload,
* other component workloads will not have any workers nominated; instead, once the primary finds a suitable Worker, that worker will be nominated for all other component workloads.

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: Core reason: https://github.com/kubernetes-sigs/kueue/pull/10266#discussion_r3117908206
This status unifies the cases of "quota reserved" being either missing or set to false.

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: For added generic context, please refer to the Motivation section.

tenzen-y: The Workload condition is responsible for the Workload condition (phase) atomically.
I'm worried about adding a new condition type only for MultiKueue because Workloads could be in an unstable phase, which means the Workload terminal state will be split into 2 condition types.

For example, as a first step, the Workload gets the `type: MultiKueueWorkload, reason: SUCCESS`, but the Workload doesn't have `type: Finished` condition even though the Workload is actually in the terminal phase.

Also, this proposed pattern loses the status field (`conditions[*].status: True|False`) responsibility.
For example, what does `type: MultiKueueWorkload, reason: SUCCESS, status: False` mean? I imagine that this state doesn't have any meaningful information.

So, I'm wondering if (1. we can have a MultiKueue specific status instead of conditions or (2. Expand existing condition reasons like `{type: Finished: reason: SucceededOnMultiKueueWorkerCluster}`.

I guess that 1 could align with your use cases (emphasize MultiKueue phase <status>). But, (1. has a risk where we have to imagine the future API design and decide the best API at this time.

The safest approach is introducing (1 as a first step, then in the second step, we can move (1) to a dedicated API (2) based on feedback. But I want to decide if we should select this step-by-step approach based on MultiKueue roadmaps. So, could you share the MK roadmap if you can?

Singularity23x0: > For example, as a first step, the Workload gets the type: MultiKueueWorkload, reason: SUCCESS, but the Workload doesn't have type: Finished condition even though the Workload is actually in the terminal phase. 

The intent is to update the MultiKueueWorkload condition on every MK Reconcile, which would be triggered on a change to the Finished condition. This would ensure the MK Condition is always synced.

> Also, this proposed pattern loses the status field (conditions[*].status: True|False) responsibility. For example, what does type: MultiKueueWorkload, reason: SUCCESS, status: False mean? I imagine that this state doesn't have any meaningful information.

Indeed, that status would not be utilized. This is one of the downsides of this solution as opposed to encoding the aggregated MK status somewhere else.

> So, I'm wondering if (1. we can have a MultiKueue specific status instead of conditions or (2. Expand existing condition reasons like {type: Finished: reason: SucceededOnMultiKueueWorkerCluster}.

Re. 1. Do you mean to add a new field to the Wl Status field? Or placing it in a new resource (similar to the Alt. 1 proposal)?

> The safest approach is introducing (1 as a first step, then in the second step, we can move (1) to a dedicated API (2) based on feedback. But I want to decide if we should select this step-by-step approach based on MultiKueue roadmaps. So, could you share the MK roadmap if you can?

The plan was to:
A. Introduce the MK Status Summary (current proposal was to do it as a Condition to make the implementation simpler, without adding the need to modify the resources and deciding on the API already).
B. At a later stage - reevaluate the need for introducing a more robust summary for MK as mentioned in the Alternatives section, potentially moving the MultiKueueGlobalStatus form the conditions into there alongside a more expansive aggregation of data from across the workers.

tenzen-y: > > So, I'm wondering if (1. we can have a MultiKueue specific status instead of conditions or (2. Expand existing condition reasons like {type: Finished: reason: SucceededOnMultiKueueWorkerCluster}.
> 
> Re. 1. Do you mean to add a new field to the Wl Status field? Or placing it in a new resource (similar to the Alt. 1 proposal)?

I meant that (1 indicates the Workload status field has a new MultiKueue specific state field.

> > The safest approach is introducing (1 as a first step, then in the second step, we can move (1) to a dedicated API (2) based on feedback. But I want to decide if we should select this step-by-step approach based on MultiKueue roadmaps. So, could you share the MK roadmap if you can?
> 
> The plan was to:
> A. Introduce the MK Status Summary (current proposal was to do it as a Condition to make the implementation simpler, without adding the need to modify the resources and deciding on the API already).
> B. At a later stage - reevaluate the need for introducing a more robust summary for MK as mentioned in the Alternatives section, potentially moving the MultiKueueGlobalStatus form the conditions into there alongside a more expansive aggregation of data from across the workers.

Thank you for sharing that. In that case, if you want to accelerate this proposal for 0.18 release, I'm ok with condition approach.
But, before beta graduation, we have to add the MultiKueue-specific status field in the Workload Custom Resource which means new MultiKueue specific Workload conditions added at this time have to be removed in second alpha stage.

In other words, if we don't move MultiKueue specific Workload state, we never graduate this KEP to Beta.

The following is the proposed timeline:

Timeline (A):

1st alpha (0.18): Add algorithms and condition types and reasons.
2nd alpha (0.19 or 0.20): Add MultiKueue-specific workload state field.
beta (0.20 or 0.21): Enabling state operation by default.
ga ....

Timeline (B):

alpha (0.18 or 0.19): Add MultiKueue-specific workload state field (we never introduce conditions).
beta (0.19 or 0.20): Enabling state operation by default.
ga ....

@Singularity23x0 Which timeline do you prefer?

Singularity23x0: Proceeding with Timeline (A) - amending the KEP.

Singularity23x0: Updated - PTAL.

keps/10189-multikueue-global-status-condition/README.md
Singularity23x0: WORKER_SELECTED is no longer a concern.
ADMITTED_ON_WORKER added.

kannon92:
Is this still relevant?

tenzen-y:
> Is this still relevant?
 
@kannon92 IIUC, this will be replaced by DQO effectiveQuotas capability.
@Singularity23x0 Please correct me if you are still thinkig this proposal.

Issue #10258, opened 2026-04-01
Refactor ungaing of Pods for the ElasticJobs to use dedicated job-agnostic controller


mimowo:
**What would you like to be cleaned**:

I would like to use the same pattern for ungating Pods for ElasticJobs as in [TopologyUngater](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/tas/topology_ungater.go) for TAS. 

Currently the ungating is triggered by the job Reconciler by calling [StartWorkloadSlicePods](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/workloadslicing/workloadslicing.go)

By analogy to TopologyUngater we could call the controller ElasticJobUngater.

**Why is this needed**:

1. single path for handling Failed Pods and Pods created for the first time, avoiding bugs like https://github.com/kubernetes-sigs/kueue/issues/9879
2. code which works for all CRDs OOTB, current short-term fix is only for RayCluster: https://github.com/kubernetes-sigs/kueue/pull/10035
3. no direct dependency from jobframework to elasticworkloads directory, so easier package-level dependency management
4. the same pattern used for similar cases (TAS & ElasticJobs)

PR #10272, opened 2026-04-01
Refactor elastic job pod ungating to dedicated controller

sohankunkerkar:
#### What type of PR is this?
/kind bug

Fixes https://github.com/kubernetes-sigs/kueue/issues/10258

- I dropped the `OwnerReference` fallback from the pod listing. The old `StartWorkloadSlicePods` had this for pods created before the `WorkloadSliceNameAnnotation` was introduced. The new controller doesn't need it as any pods that pre-date the annotation were already ungated by the old code path before this refactor takes effect
- This PR addresses all the motivations from [#10258](https://github.com/kubernetes-sigs/kueue/issues/10258) except 3, which is partially addressed. The ungating path is decoupled from jobframework, but jobframework still depends on workloadslicing for slice lifecycle management. I can handle a full package decoupling as a separate refactor. The existing `StartWorkloadSlicePods` helper is kept for safety and should be deprecated and will be removed in a follow-up once the controller path is validated in CI.

```release-note
ElasticJobs: Fix a bug where pods stay gated after scale-up by allowing finished workloads to ungate their own pods.
```

### DIFF
--- a/test/e2e/singlecluster/kuberay_test.go
+++ b/test/e2e/singlecluster/kuberay_test.go
@@ -386,6 +386,45 @@ print([ray.get(my_task.remote(i, 1)) for i in range(32)])`,
 			}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
 		})
 
+		ginkgo.By("Waiting for all 5 worker pods to be running", func() {
+			gomega.Eventually(func(g gomega.Gomega) {
+				podList := &corev1.PodList{}
+				g.Expect(k8sClient.List(ctx, podList, client.InNamespace(ns.Name))).To(gomega.Succeed())
+				runningWorkers := getRunningWorkerPodNames(podList)
+				g.Expect(runningWorkers).To(gomega.HaveLen(5), "Expected 5 running worker pods")
+			}, util.VeryLongTimeout, util.Interval).Should(gomega.Succeed())
+		})
+
+		var deletedPodName string
+		ginkgo.By("Deleting one worker pod", func() {
+			podList := &corev1.PodList{}
+			gomega.Expect(k8sClient.List(ctx, podList, client.InNamespace(ns.Name))).To(gomega.Succeed())
+			runningWorkers := getRunningWorkerPodNames(podList)
+			gomega.Expect(runningWorkers).NotTo(gomega.BeEmpty())
+			deletedPodName = runningWorkers[0]
+			pod := &corev1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      deletedPodName,
+					Namespace: ns.Name,
+				},
+			}
+			gomega.Expect(k8sClient.Delete(ctx, pod)).To(gomega.Succeed())
+		})
+
+		ginkgo.By("Waiting for a new worker pod to replace the deleted one", func() {
+			gomega.Eventually(func(g gomega.Gomega) {
+				podList := &corev1.PodList{}
+				g.Expect(k8sClient.List(ctx, podList, client.InNamespace(ns.Name))).To(gomega.Succeed())
+				runningWorkers := getRunningWorkerPodNames(podList)
+				g.Expect(runningWorkers).To(gomega.HaveLen(5), "Expected 5 running worker pods after replacement")
+				g.Expect(runningWorkers).NotTo(gomega.ContainElement(deletedPodName),
+					"Deleted pod should not be present among running workers")
+				// Update scaledUpPodNames to include the replacement pod,
+				// so the later scale-down superset check accounts for it.
+				scaledUpPodNames = runningWorkers
+			}, util.VeryLongTimeout, util.Interval).Should(gomega.Succeed())
+		})
+
 		ginkgo.By("Waiting for workers reduced to 1 due to scaling down", func() {
 			gomega.Eventually(func(g gomega.Gomega) {
 				podList := &corev1.PodList{}


pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -0,0 +1,229 @@
...
+		return reconcile.Result{}, nil
+	}
+	if !r.expectationsStore.Satisfied(log, req.NamespacedName) {
+		return reconcile.Result{}, errPendingUngateOps
+	}
+	if !workload.IsAdmitted(wl) || !workloadslicing.IsElasticWorkload(wl) {
+		return reconcile.Result{}, nil
+	}
+
+	pods, err := r.listGatedPods(ctx, wl)
```
hiboyang, line 119: Curious about performance and memory impact here if calling `listGatedPods(ctx, wl)` frequently. If there is a large number of pods for many workloads, the processing here may be delayed, then pending reconciliation events can be accumulated and cause pressure to internal event queue?

sohankunkerkar: The `listGatedPods` call uses the `WorkloadSliceNameKey` field index, so it's an indexed lookup scoped to a single workload slice, not a full pod list. The workload predicate also filters to only admitted elastic workloads, and the expectations pattern prevents re-processing once pods are ungated. This is the same pattern TopologyUngater uses for TAS. (no additional performance overhead compared to what's already in production)

hiboyang: Cool, thanks for the clarification!

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -0,0 +1,229 @@
...
+
+func (h *elasticPodHandler) Generic(context.Context, event.GenericEvent, workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+}
+
+func (h *elasticPodHandler) queueReconcileForPod(ctx context.Context, object client.Object, deleted bool, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	pod, isPod := object.(*corev1.Pod)
+	if !isPod {
+		return
+	}
+	wlName, found := pod.Annotations[kueue.WorkloadAnnotation]
```
mimowo, line 224: I think this is not fully correct, because this will not catch the case when a Pod is re-created for a Workload which no longer exists. So, I'm thinking the unit of reconciliation should rather be the "WorkloadSlice" rather than a specific workload. Once we get the slice then we can check the status of the "last not finished workload in the slice". 

I will add a separate comment to cover this in elastic_job_ungater_test.go

pkg/controller/elasticjobs/elastic_job_ungater_test.go
```diff
@@ -0,0 +1,361 @@
...
+	now := time.Now().Truncate(time.Second)
+
+	testCases := map[string]struct {
+		expectUIDs []types.UID
+		workloads  []kueue.Workload
+		pods       []corev1.Pod
+		wantPods   []corev1.Pod
+		wantErr    error
+	}{
+		"ungate single pod": {
```
mimowo, line 60: Please add a test case for scenario when WorkloadSliceNameAnnotation!= WorkloadAnnotation, because the workload was already scaled up.

pkg/controller/elasticjobs/elastic_job_ungater.go
mimowo: Ahhh, ok this is less of an issue that I initially thought, because the replacement Pods will already have the Workload annotation pointing to the last admitted workload (still worth a unit test).

So the only small gap we could close by using the sliceName annotation here is when there is a Pod created, the Workload is scaled up before ungating, and the old Workload Slice is quickly deleted. This is pretty unlikely, but might be good to cover nonetheless. Let me know what you think.

sohankunkerkar: Thanks, I agree the original concern is smaller than it first looked, since replacement Pods already point via `WorkloadAnnotation` to the last admitted Workload. The remaining gap is the narrow race you described: Pod created, Workload scales up before ungating, old WorkloadSliceNameAnnotation doesn't match the new slice name. I'll add the unit test for `WorkloadSliceNameAnnotation != WorkloadAnnotation`. If the test shows the current code misses those pods, I'll fix it in the smallest way possible here. Otherwise I'll keep slice-based reconciliation for a follow-up, thoughts?

mimowo: Unless good reasons I would prefer to contain the fix for this scenario in this PR, because I consider this a bugfix and I would like to cherrypick. If it is not included then we may risk releasing 0.17.1 with the gap.

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -0,0 +1,252 @@
...
+			}
+			return ungated, nil
+		})
+		if e != nil {
+			r.expectationsStore.ObservedUID(log, req.NamespacedName, pod.UID)
+			log.Error(e, "failed ungating elastic pod", "pod", klog.KObj(pod))
+		}
+		if !ungated {
+			r.expectationsStore.ObservedUID(log, req.NamespacedName, pod.UID)
+		}
```
mimowo: I'm not sure I understand this code, why not just:

```golang
		if e != nil {
			r.expectationsStore.ObservedUID(log, req.NamespacedName, pod.UID)
			log.Error(e, "failed ungating elastic pod", "pod", klog.KObj(pod))
		}
		return e
```
current code looks like may call `r.expectationsStore.ObservedUID1` twice on error, which seems harmless, but wrong

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -0,0 +1,252 @@
...
+	sliceName := workloadslicing.SliceName(wl)
+	var slicePods corev1.PodList
+	if err := r.client.List(ctx, &slicePods,
+		client.InNamespace(wl.Namespace),
+		client.MatchingFields{indexer.WorkloadSliceNameKey: sliceName},
+	); err != nil {
+		return nil, fmt.Errorf("listing pods for workload slice: %w", err)
+	}
+
+	// After scale-up, pods may still carry the old slice annotation while
```
mimowo: IIUC this is just for legacy workloads before the introduction of the Workload SliceName annotation, right? 

I would like to make that comment explicit.

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -0,0 +1,252 @@
...
+		}
+		if !ungated {
+			r.expectationsStore.ObservedUID(log, req.NamespacedName, pod.UID)
+		}
+		return e
+	})
+	return reconcile.Result{}, err
+}
+
+func (r *elasticJobUngater) listGatedPods(ctx context.Context, wl *kueue.Workload) ([]*corev1.Pod, error) {
```
mimowo, line 153: I would like to commonize this function with StartWorkloadSlicePods because they actually solve the same problem. For that we should decouple into two stages: findGatedPods, and actual ungating,

On that note, do we still need `StartWorkloadSlicePods` at all?

pkg/controller/elasticjobs/elastic_job_ungater.go
sohankunkerkar: Good catch! the `!ungated` case is effectively unreachable here since `listGatedPods` already filters for pods with the gate.

pkg/controller/elasticjobs/elastic_job_ungater.go
sohankunkerkar: Not just legacy but non-elastic workloads (regular Jobs with Kueue management but without `ElasticJobsViaWorkloadSlices`) only have the `WorkloadAnnotation`, never the        `WorkloadSliceNameAnnotation`. TAS also uses this index for those pods. I'll update the comment to make both cases explicit

pkg/controller/elasticjobs/elastic_job_ungater.go
sohankunkerkar: `StartWorkloadSlicePods` is no longer called from `ReconcileGenericJob` (removed in this PR), so yes, it's effectively dead code now. I kept it to avoid removing it in the same PR as the refactor, per your earlier comment about keeping this minimal for cherrypick. I can remove it and commonize in a follow-up, or if you prefer I can do it here. Let me know your thoughts.

pkg/controller/elasticjobs/elastic_job_ungater.go
mimowo: Hold on, this code is only used for ElasticWorkloads even per the file name.

sohankunkerkar: Yeah sorry, wrong context. This query is here for the scale-up mismatch you asked to fix earlier. After scale-up, pods still have the old slice annotation but already point to the new workload via WorkloadAnnotation. The slice name lookup alone misses them.

mimowo: > After scale-up, pods still have the old slice annotation but already point to the new workload via WorkloadAnnotation. The slice name lookup alone misses them.

This actually might be a bug here, because when the Pods are listed by WorkloadSliceNameKey (above), they may be listed for the new Workload which is not amitted yet. If the new workload is not admitted then the Pods should stay gated. 

Actually, I would like to add an e2e test case for this scenario. We attempt to scale up, but there is not enough qouta, so the new Workload stays pending - then verify the newly created Pods remain gated.

pkg/controller/elasticjobs/elastic_job_ungater.go
mimowo: I'm ok with both approaches, let's focus on the remaining issues with the controller itself.

test/e2e/singlecluster/kuberay_test.go
```diff
@@ -222,12 +222,12 @@ def my_task(x, s):
...
 # run tasks in parallel to trigger autoscaling (scaling up)
-# Use longer sleep (8s) to give autoscaler time to detect demand,
+# Use longer sleep (16s) to give autoscaler time to detect demand,
 # create workload slices, and schedule new workers.
-print(ray.get([my_task.remote(i, 8) for i in range(16)]))
+print(ray.get([my_task.remote(i, 16) for i in range(16)]))
 
 # run tasks in sequence to trigger scaling down
-print([ray.get(my_task.remote(i, 1)) for i in range(32)])`,
+print([ray.get(my_task.remote(i, 1)) for i in range(256)])`,
```
mimowo: I think these are just moved from the other PR, can we revert? I don't see a need for that here - actually it was also my comment in the previous PR.

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -0,0 +1,250 @@
...
+		client.InNamespace(wl.Namespace),
+		client.MatchingFields{indexer.WorkloadSliceNameKey: sliceName},
+	); err != nil {
+		return nil, fmt.Errorf("listing pods for workload slice: %w", err)
+	}
+
+	// After scale-up, replacement pods created by the operator (e.g. KubeRay)
+	// may carry the parent's WorkloadSliceNameAnnotation while their
+	// WorkloadAnnotation already points to the current admitted workload.
+	// Query by workload name to find these pods.
```
mimowo: I'm actually unclear for the need for this. I thought this is just for backwards compatibility for Workloads where pods may not have the WorkloadSliceName annotation, but for an Alpha feature this is a secondary concern.

Assuming the WorkloadSliceNameAnnotation is populated to the Pods, they should all be found by the previous check.

mimowo: OTOH, now I'm wondering the opposite - we probably could fully rely on the WorkloadNameAnnotation - could you remind me why we actually check the WorkloadSliceName also here?

I think it was needed on the previous approach when the ungating was triggered by the Job reconciler to make sure all Pods are ungated no matter which Workload they belong to.

Now, if we have a mix of Pods belonging to different Workloads, then we will schedule two reconcile events, and they will both ungate their subset of Pods, if the workload is admitted. So in the new approach I actually don't see a need currently to use WorkloadSliceName.

pkg/controller/elasticjobs/elastic_job_ungater.go
sohankunkerkar: Addressed. `listGatedPods` now filters pods through `eligibleWorkloadNames`, which walks the replacement chain and only includes workload names in the admitted chain. Pods from a pending replacement workload are excluded. Added unit test to cover this case

pkg/controller/elasticjobs/elastic_job_ungater.go
sohankunkerkar: After scale-up the parent workload is finished, so its reconcile returns early at `IsAdmitted`. Parent pods with `WorkloadAnnotation: "wl"` would never get ungated if we only query by `WorkloadAnnotation` as no one reconciles wl as admitted anymore. The slice name query lets the child workload find all pods in the chain with one List, then `eligibleWorkloadNames` filters out pods from non-admitted workloads.

sohankunkerkar: As discussed in DM with @mimowo , I simplified the ungating approach: each workload now handles its own pods (filtered by WorkloadAnnotation), and finished workloads that were previously admitted can still ungate their pods. This removes the chain-walking logic and keeps the controller lean.

mimowo:
/cherrypick release-0.17
@hiboyang I merged the PR as this LGTM, but an extra round of manual testing before the release would be great

tenzen-y:
/release-note-edit
```release-note
ElasticJobs: Fix a bug where pods stay gated after scale-up by allowing finished workloads to ungate their own pods.                                                                                                                                                                                      
```

PR #10274, opened 2026-04-02
Fix stale workload accumulation after PodsReady timeout eviction

### LINKED ISSUE #1789: Workloads corresponding to Jobs deleted with --cascede=orphan still consume resources
**What happened**:

Job deleted with --cascade=orphan continues to reserve ClusterQueue resources.

Note this is a "known" issue, which is a follow up to https://github.com/kubernetes-sigs/kueue/issues/1726

Jobs deleted with --cascade=orphan should free the cluster resources.

1. Create a job with
```yaml
apiVersion: batch/v1

kind: Job
metadata:
  name: sample-job
  namespace: default
  labels:
    kueue.x-k8s.io/queue-name: user-queue
spec:
  parallelism: 3
  completions: 3
  suspend: true
  template:
    spec:
      containers:
      - name: dummy-job
        image: gcr.io/k8s-staging-perf-tests/sleep:v0.0.3
        args: ["1800s"]
        resources:
          requests:
            cpu: "1"
            memory: "200Mi"
      restartPolicy: Never
```
2. Delete the job with `kubectl delete job/sample-job --cascade=orphan`

Issue: The deleted job continues to reserve cluster queue resources:

`kubectl get clusterqueue -oyaml` returns:

```yaml
...
  status:
    admittedWorkloads: 1
    conditions:
    - lastTransitionTime: "2024-03-04T10:41:42Z"
      message: Can admit new workloads
      reason: Ready
      status: "True"
      type: Active
    flavorsReservation:
    - name: default-flavor
      resources:
      - borrowed: "0"
        name: cpu
        total: "3"
      - borrowed: "0"
        name: memory
        total: 600Mi
    flavorsUsage:
    - name: default-flavor
      resources:
      - borrowed: "0"
        name: cpu
        total: "3"
      - borrowed: "0"
        name: memory
        total: 600Mi
    pendingWorkloads: 0
    reservingWorkloads: 1
```

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

sebest:
## Summary

This PR fixes two related issues with orphaned workloads holding ClusterQueue quota:

### 1. Stale workload accumulation after PodsReady timeout eviction

When a pod workload is evicted due to `waitForPodsReady` timeout, Kueue deletes the pod via `Stop()` and requeues the workload with exponential backoff. If the pod is owned by a Deployment, the Deployment creates a replacement pod with a new workload. The old evicted workload keeps getting requeued despite its owning pod being deleted, causing unbounded stale workload accumulation.

**Fix:** Add a `FinishOrphanedWorkloads` feature gate (Beta in 0.18, default enabled) that detects when a workload's controller owner no longer exists (`isControllerOwnerGone` check in the workload reconcile path) and finishes the workload with reason `OwnerNotFound`.

### 2. Jobs deleted with `--cascade=orphan` hold quota indefinitely (#1789)

When a Job is deleted, the jobframework reconciler's `dropFinalizers` path removes the workload's finalizer but does not finish it. With `--cascade=orphan`, the workload remains admitted and holds quota indefinitely.

**Fix:** Call `workload.Finish` with `OwnerNotFound` in the `dropFinalizers` path before removing the finalizer, so the workload releases quota immediately.

### Changes

- `apis/kueue/v1beta2/workload_types.go` — Added `WorkloadFinishedReasonOwnerNotFound` constant
- `pkg/features/kube_features.go` — Added `FinishOrphanedWorkloads` feature gate (Beta in 0.18)
- `pkg/controller/core/workload_controller.go` — Added `isControllerOwnerGone` orphan detection in the general reconcile path
- `pkg/controller/jobframework/reconciler.go` — Finish admitted workloads in `dropFinalizers` path (#1789)
- `pkg/controller/core/workload_controller_test.go` — Unit tests for orphan detection
- `test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go` — Integration test for PodsReady eviction scenario
- `test/integration/singlecluster/controller/jobs/job/job_controller_test.go` — Integration test for #1789

## Test plan

- [x] New unit tests pass (orphan detection with/without feature gate, existing/gone owner)
- [x] New pod controller integration test passes (PodsReady eviction scenario)
- [x] New job controller integration test passes (#1789 scenario)
- [x] Full job controller integration suite passes (66/66)
- [x] Full pod controller integration suite passes (27/27)
- [x] All controller unit tests pass
- [ ] CI passes

Fixes #1789

```release-note
Fix handling of orphaned workloads which could result in the accumulation of stale workloads
after PodsReady timeout eviction for Deployment-owned pods.
```


Huang-Wei:
cc @ichekrygin @amy

mimowo:
@sebest I injected release note into the PR description to make the robot happy, and also to use for the release notes summary. PTAL

pkg/controller/core/workload_controller.go
```diff
@@ -380,6 +380,31 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+					log.V(3).Info("Pending workload is being garbage collected, skipping requeue")
+					return ctrl.Result{}, nil
+				}
+
+				// Before re-enqueuing, check if the workload's controller owner
+				// still exists. If the owner (e.g., a pod deleted by eviction)
+				// is gone, the workload is stale and should be finished rather
+				// than re-enqueued. This prevents unbounded stale workload
+				// accumulation when Deployment-owned pods are evicted and replaced.
+				if ownerGone, err := r.isControllerOwnerGone(ctx, &wl); err != nil {
```
mimowo: I like the approach, but I'm wondering if we should move the fix from here to somewhere around there: https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/core/workload_controller.go#L203-L207

This way we could also fix https://github.com/kubernetes-sigs/kueue/issues/1789 with the same code.

Let me know what are your thoughts, I know this is a bit extending the scope, but I think such handling of orphan Workloads could be done also in other scenarios - not just eviction.

To stay on the safe side we could then protect the fix by a feature gate (already in Beta), like FinishOrphanedWorkloads.

Let me know your thoughts.

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
```diff
@@ -2659,3 +2662,168 @@ var _ = ginkgo.Describe("Pod controller with TASReplaceNodeOnPodTermination", gi
...
+			g.Expect(wl2.Status.Conditions).To(gomega.ContainElements(
+				gomega.BeComparableTo(metav1.Condition{
+					Type:   kueue.WorkloadQuotaReserved,
+					Status: metav1.ConditionTrue,
+				}, wlConditionCmpOpts...),
+				gomega.BeComparableTo(metav1.Condition{
+					Type:   kueue.WorkloadAdmitted,
+					Status: metav1.ConditionTrue,
+				}, wlConditionCmpOpts...),
+			))
```
mimowo: Let's use the helpers like ExpectWorkloadsToBeAdmitted

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
```diff
@@ -2659,3 +2662,168 @@ var _ = ginkgo.Describe("Pod controller with TASReplaceNodeOnPodTermination", gi
...
+			Obj()
+		util.MustCreate(ctx, k8sClient, pod1)
+
+		ginkgo.By("waiting for pod-1 to get a workload and be admitted")
+		createdPod1 := &corev1.Pod{}
+		gomega.Eventually(func(g gomega.Gomega) {
+			g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(pod1), createdPod1)).To(gomega.Succeed())
+			for _, gate := range createdPod1.Spec.SchedulingGates {
+				g.Expect(gate.Name).ShouldNot(gomega.Equal("kueue.x-k8s.io/admission"))
+			}
```
mimowo: IIUC waiting for the actual schedulingGates deletion is not really needed for the minimal reproducible scenario, so we can skip. If there is a reason for checking, then assert that the set of gates is empty, it will be fewer lines.

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
```diff
@@ -2659,3 +2662,168 @@ var _ = ginkgo.Describe("Pod controller with TASReplaceNodeOnPodTermination", gi
...
+				gomega.BeComparableTo(metav1.Condition{
+					Type:   kueue.WorkloadQuotaReserved,
+					Status: metav1.ConditionTrue,
+				}, wlConditionCmpOpts...),
+				gomega.BeComparableTo(metav1.Condition{
+					Type:   kueue.WorkloadAdmitted,
+					Status: metav1.ConditionTrue,
+				}, wlConditionCmpOpts...),
+			))
+		}, util.Timeout, util.Interval).Should(gomega.Succeed())
```
mimowo: Use the helper for checking that the workload is admitted, like ExpectWorkloadsToBeAdmitted

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
```diff
@@ -2659,3 +2662,168 @@ var _ = ginkgo.Describe("Pod controller with TASReplaceNodeOnPodTermination", gi
...
+
+		ginkgo.By("waiting for pod-1 to be deleted by the pod controller (Stop)")
+		gomega.Eventually(func(g gomega.Gomega) {
+			err := k8sClient.Get(ctx, client.ObjectKeyFromObject(pod1), createdPod1)
+			if apierrors.IsNotFound(err) {
+				return // pod is gone, success
+			}
+			g.Expect(err).NotTo(gomega.HaveOccurred())
+			g.Expect(createdPod1.DeletionTimestamp).ShouldNot(gomega.BeZero())
+		}, util.Timeout, util.Interval).Should(gomega.Succeed())
```
mimowo: Can we use ExpectPodsFinalizedOrGone instead?

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
```diff
@@ -2659,3 +2662,168 @@ var _ = ginkgo.Describe("Pod controller with TASReplaceNodeOnPodTermination", gi
...
+			Obj()
+		util.MustCreate(ctx, k8sClient, pod2)
+
+		ginkgo.By("waiting for pod-2 to get a workload and be admitted")
+		createdPod2 := &corev1.Pod{}
+		gomega.Eventually(func(g gomega.Gomega) {
+			g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(pod2), createdPod2)).To(gomega.Succeed())
+			for _, gate := range createdPod2.Spec.SchedulingGates {
+				g.Expect(gate.Name).ShouldNot(gomega.Equal("kueue.x-k8s.io/admission"))
+			}
```
mimowo: As for pod1, I think this check isn't really necessary, and the check for the state of the workload should go first, because first workload gets admitted, and only later the scheduling gates get removed.

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
```diff
@@ -2659,3 +2662,168 @@ var _ = ginkgo.Describe("Pod controller with TASReplaceNodeOnPodTermination", gi
...
+		}, util.Timeout, util.Interval).Should(gomega.Succeed())
+
+		ginkgo.By("asserting workload-1 is finished as stale (owner pod-1 was deleted)")
+		gomega.Eventually(func(g gomega.Gomega) {
+			g.Expect(k8sClient.Get(ctx, wl1Key, wl1)).To(gomega.Succeed())
+			g.Expect(workload.IsFinished(wl1)).Should(gomega.BeTrue(),
+				"stale workload-1 should be finished after its owner pod was deleted")
+			finishedCond := apimeta.FindStatusCondition(wl1.Status.Conditions, kueue.WorkloadFinished)
+			g.Expect(finishedCond).ShouldNot(gomega.BeNil())
+			g.Expect(finishedCond.Reason).Should(gomega.Equal("OwnerNotFound"))
```
mimowo: nit, here I would actually suggest using g.Expect(wl1.Status.Conditions).To(gomega.ContainElements( for readability of the assert

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
```diff
@@ -2659,3 +2662,168 @@ var _ = ginkgo.Describe("Pod controller with TASReplaceNodeOnPodTermination", gi
...
+	var (
+		ns *corev1.Namespace
+		fl *kueue.ResourceFlavor
+		cq *kueue.ClusterQueue
+		lq *kueue.LocalQueue
+	)
+
+	ginkgo.BeforeAll(func() {
+		waitForPodsReady := &configapi.WaitForPodsReady{
+			Timeout:        metav1.Duration{Duration: 3 * time.Second},
```
mimowo, line 2674: Is there any reason for making the timeout that long, I'm wondering the issue should be reproducible also using util.TinyTimeout, please check, if this is, then  we can safe some waiting time.

pkg/controller/core/workload_controller.go
sebest: Hey @mimowo , thanks for the review!
I will look into it in a bit and see if I can address https://github.com/kubernetes-sigs/kueue/issues/1789 with the same code.

sebest: Done. Moved the orphan check to the general reconcile path (after L207), gated behind `FinishOrphanedWorkloads` (Beta in 0.18). The `isControllerOwnerGone` check detects when the ownerReference points to a non-existent object and finishes the workload.

For #1789 specifically, also added `workload.Finish` in the jobframework reconciler's `dropFinalizers` path — this finishes admitted workloads before removing their finalizer when the owner Job is deleted, so quota is released immediately regardless of cascade policy.

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
sebest: Switched to `util.TinyTimeout` — works fine (in unit tests)

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
sebest: Removed. Kept only the `Get` to obtain the UID for workload name computation.

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
sebest: Done — using `util.ExpectWorkloadsToBeAdmittedByKeys`.

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
sebest: Done.

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
sebest: Removed.

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
sebest: Done — using `util.ExpectWorkloadsToBeAdmittedByKeys`.

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
sebest: Used `utiltesting.HaveConditionStatusTrueAndReason` instead — even more concise and idiomatic to the codebase. Also extracted `"OwnerNotFound"` to `kueue.WorkloadFinishedReasonOwnerNotFound` constant alongside the existing `WorkloadFinishedReason*` constants.

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
sebest: Update: tried both `util.TinyTimeout` and `util.ShortTimeout` — neither works. With very short timeouts, the PodsReady eviction fires before the test can observe the admitted state (the queuing delay counts toward the timeout). Kept the original 3s which gives enough window for the test to poll the admitted state before eviction.

pkg/controller/jobframework/reconciler.go
```diff
@@ -319,6 +319,19 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
...
+			if workload.IsAdmitted(&workloads.Items[i]) {
+				err := workload.Finish(ctx, r.client, &workloads.Items[i],
+					kueue.WorkloadFinishedReasonOwnerNotFound,
+					"The workload's owner no longer exists",
+					r.clock)
+				if err != nil && !apierrors.IsNotFound(err) {
+					log.Error(err, "Finishing orphaned workload")
+					return ctrl.Result{}, err
+				}
+			}
```
mimowo: I think this is not the good place to fix the --cascade=orphan and likely is the source of the issue actually.

I guess I understand why the code added to workload_controller Reconcile in itself does not work - because there is nothing that would trigger the Reconcile on Job deletion, is that assumption correct?

I suppose this code added here might be messing with the tests as it does not have all the checks which the code in workload_controller has.

So, if this is the case we could solve it from the JobReconciler too, but would need to repeat exactly the same conditions as for Workload. So extract the new code in workload_controller to a helper like `workload.FinishedOrphanedWorkload`, and call it from both places.

At the same time seeing the complication I'm also ok to put it out of scope for the PR. Up to you, but I think it makes sense to explore the problem deeper now.

sebest: I'll try to address it in a separate PR on top of this one.

mimowo: sgtm

mimowo:
I think can skip the info about the FG in release note.
/release-note-edit
```release-note
Fix handling of orphaned workloads which could result in the accumulation of stale workloads
after PodsReady timeout eviction for Deployment-owned pods. 
```

mimowo (commented):
Let's try to include in the patch release today. Could you please prepare the CPs manually, and adjust the FG on the patch releases to 0.16 and 0.17 respectively?

/lgtm
/approve

mimowo:
Ok I opened: 
- 0.16 https://github.com/kubernetes-sigs/kueue/pull/10520 
- 0.17  https://github.com/kubernetes-sigs/kueue/pull/10516

mimowo:
cc @sebest we are downgrading the FG, and likely will need to drop it completely in future releases (TBD). 
We need to rethink how to approach the problem this PR attempted to fix.

PR #10275, opened 2026-04-02
LeaderWorkerSet: Relax PodSpec validation to allow changing nodeSelector.

### LINKED ISSUE #10178: Allow to mutate nodeSelector for LWS

Currently we have the [webook  ](https://github.com/kubernetes-sigs/kueue/blob/v0.15.3/pkg/controller/jobs/leaderworkerset/leaderworkerset_webhook.go#L243)
which does not allow to mutate the LWS. 

As a quick fix workaround if would be useful to investigate if the webhook could be disabled entirely on a cluster.

**Why is this needed**:

Some users would like to mutate the nodeSelector while the LWS running so that they can re-locate the LWS. It is ok if the LWS workloads are evicted and re-admitted.

mbobrovskyi:
LeaderWorkerSet: Relax PodSpec validation to allow changing nodeSelector.

Fixes #10178

```release-note
NONE
```

### DIFF
--- a/pkg/controller/jobframework/reconciler.go
+++ b/pkg/controller/jobframework/reconciler.go
@@ -1207,7 +1207,8 @@ func EquivalentToWorkload(ctx context.Context, c client.Client, job GenericJob, 
 	}
 	jobPodSets := clearMinCountsIfFeatureDisabled(getPodSets)
 
-	opts := make([]equality.ComparePodSetsOption, 0, 1)
+	opts := make([]equality.ComparePodSetsOption, 0, 2)
+	opts = append(opts, equality.WithIgnoreNodeSelector())
 	if workload.IsAdmitted(wl) {
 		opts = append(opts, equality.WithIgnoreTolerations())
 	}
--- a/pkg/util/equality/podset_test.go
+++ b/pkg/util/equality/podset_test.go
@@ -28,10 +28,11 @@ import (
 
 func TestComparePodSetSlices(t *testing.T) {
 	cases := map[string]struct {
-		a                 []kueue.PodSet
-		b                 []kueue.PodSet
-		ignoreTolerations bool
-		wantEquivalent    bool
+		a                  []kueue.PodSet
+		b                  []kueue.PodSet
+		ignoreTolerations  bool
+		ignoreNodeSelector bool
+		wantEquivalent     bool
 	}{
 		"different name": {
 			a:              []kueue.PodSet{*utiltestingapi.MakePodSet("ps", 10).SetMinimumCount(5).Obj()},
@@ -46,7 +47,13 @@ func TestComparePodSetSlices(t *testing.T) {
 		"different node selector": {
 			a:              []kueue.PodSet{*utiltestingapi.MakePodSet("ps", 10).SetMinimumCount(5).Obj()},
 			b:              []kueue.PodSet{*utiltestingapi.MakePodSet("ps", 10).SetMinimumCount(5).NodeSelector(map[string]string{"key": "val"}).Obj()},
-			wantEquivalent: true,
+			wantEquivalent: false,
+		},
+		"different node selector with ignoreNodeSelector": {
+			a:                  []kueue.PodSet{*utiltestingapi.MakePodSet("ps", 10).SetMinimumCount(5).Obj()},
+			b:                  []kueue.PodSet{*utiltestingapi.MakePodSet("ps", 10).SetMinimumCount(5).NodeSelector(map[string]string{"key": "val"}).Obj()},
+			ignoreNodeSelector: true,
+			wantEquivalent:     true,
 		},
 		"different requests": {
 			a:              []kueue.PodSet{*utiltestingapi.MakePodSet("ps", 10).SetMinimumCount(5).Request("res", "1").Obj()},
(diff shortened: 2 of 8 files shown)


pkg/controller/jobframework/reconciler.go
```diff
@@ -1212,18 +1222,23 @@ func EquivalentToWorkload(ctx context.Context, c client.Client, job GenericJob,
 	}
 	jobPodSets := clearMinCountsIfFeatureDisabled(getPodSets)
 
+	opts := make([]equality.ComparePodSetSlicesOption, 0, 2)
+	if workload.IsAdmitted(wl) {
+		opts = append(opts, equality.WithIgnoreTolerations(), equality.WithIgnoreNodeSelector())
```
kannon92: If we are adding tolerations also can we add a e2e test for that also?

pkg/util/equality/podset_test.go
```diff
@@ -28,10 +28,11 @@ import (
 
 func TestComparePodSetSlices(t *testing.T) {
 	cases := map[string]struct {
-		a                 []kueue.PodSet
-		b                 []kueue.PodSet
-		ignoreTolerations bool
-		wantEquivalent    bool
+		a                  []kueue.PodSet
+		b                  []kueue.PodSet
+		ignoreTolerations  bool
```
kannon92, line 33: Can you add a unit test for using this field? I think its unusued right now.

mbobrovskyi: We already have it.

pkg/controller/jobframework/reconciler.go
mbobrovskyi: Actually, we added nodeSelector.

pkg/util/equality/podset_test.go
kannon92: Ah sorry git diffs confused me here.

pkg/controller/jobframework/reconciler.go
kannon92: Ah but I guess we don't have a e2e test for mutating tolerations. 

This can be done as a followup.

pkg/controller/jobframework/reconciler.go
```diff
@@ -1207,7 +1207,8 @@ func EquivalentToWorkload(ctx context.Context, c client.Client, job GenericJob,
 	}
 	jobPodSets := clearMinCountsIfFeatureDisabled(getPodSets)
 
-	opts := make([]equality.ComparePodSetsOption, 0, 1)
+	opts := make([]equality.ComparePodSetsOption, 0, 2)
+	opts = append(opts, equality.WithIgnoreNodeSelector())
```
mimowo, line 1211: Can you elaborate why / if this is needed? This seems surprising to me,  because:
1. JobReconciler in the jobframework iiuc is not responsible for creating or managing workloads for LWS
2. it affects other Jobs and so the consequences are unclear to me

So my suspicion is that this is not necessary / unrelated change.

mbobrovskyi: Yes, exactly. But if we don’t add WithIgnoreNodeSelector() to the jobframework, the behavior will change and we’ll end up comparing nodeSelectors as well. We don’t need to make any changes to the jobframework for this fix.

mbobrovskyi: WithIgnoreNodeSelector() ignores the behavior I added in comparePodTemplate().

mimowo: Ahh, I see because `IgnoreNodeSelector` was previously the default (there was not code to verify that). THis makes sense.

test/e2e/singlecluster/leaderworkerset_test.go
```diff
@@ -135,6 +135,55 @@ var _ = ginkgo.Describe("LeaderWorkerSet integration", ginkgo.Label("area:single
...
+					g.Expect(k8sClient.Update(ctx, createdLeaderWorkerSet)).To(gomega.Succeed())
+				}, util.Timeout, util.Interval).Should(gomega.Succeed())
+			})
+
+			ginkgo.By("Checking that workload recreated", func() {
+				gomega.Eventually(func(g gomega.Gomega) {
+					g.Expect(k8sClient.Get(ctx, util.WorkloadKeyForLeaderWorkerSet(lws, "0"), createdWorkload)).To(gomega.Succeed())
+					g.Expect(createdWorkload.UID).ToNot(gomega.Equal(oldUID))
+				}, util.Timeout, util.Interval).Should(gomega.Succeed())
+			})
```
mimowo, line 185: Let's also verify the Pods (replicas) get running (Ready) after the change.

pkg/controller/jobframework/reconciler.go
mimowo: lgtm

test/e2e/singlecluster/leaderworkerset_test.go
```diff
@@ -135,6 +135,63 @@ var _ = ginkgo.Describe("LeaderWorkerSet integration", ginkgo.Label("area:single
...
+			ginkgo.By("Updating nodeSelector", func() {
+				createdLeaderWorkerSet := &leaderworkersetv1.LeaderWorkerSet{}
+				gomega.Eventually(func(g gomega.Gomega) {
+					g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(lws), createdLeaderWorkerSet)).To(gomega.Succeed())
+					createdLeaderWorkerSet.Spec.LeaderWorkerTemplate.WorkerTemplate.Spec.NodeSelector = map[string]string{"instance-type": "on-demand"}
+					g.Expect(k8sClient.Update(ctx, createdLeaderWorkerSet)).To(gomega.Succeed())
+				}, util.Timeout, util.Interval).Should(gomega.Succeed())
+			})
+
+			ginkgo.By("Checking that workload recreated", func() {
```
mimowo, line 180: Let's also await that the newly craeted workload is admitted. This checks the admission worked properly, and also gives more time before the ReadyReplicas check so that it would go down if the Pods were not restared.

test/e2e/singlecluster/leaderworkerset_test.go
mbobrovskyi: Done

test/e2e/singlecluster/leaderworkerset_test.go
mbobrovskyi: Done

mimowo:
/lgtm
/approve
Thank you 👍

mimowo (commented):
Let me cherrypick as the diff is pretty small, and I've got an ask from a user.
/cherrypick release-0.17
/cherrypick release-0.16

mimowo:
cc @mbobrovskyi could you prepare the cherrypicks here?

mimowo:
/release-note-edit
```release-note
NONE
```
We decided to revert in https://github.com/kubernetes-sigs/kueue/pull/11214 due to complications encountered during manual testing. We may return to it later.

Issue #9854, opened 2026-03-13
Consolidate gauge metric cleanup mechanisms


IrvingMg:
**What would you like to be cleaned**:
There are two different mechanisms used to clean up stale Prometheus gauge series when a label value changes.

1. `StoreAndClear` (`pkg/metrics/custom_labels.go`). This was introduced in #9774 for custom metric labels. Each controller (CQ, LQ, Cohort) tracks its current label values. When a value changes, for example, when a CQ's team annotation changes from `alpha` to `beta`, the controller deletes the old metric series for that object. During the next update cycle, the metrics are created again with the new label value.
2. `ResyncGaugeMetrics` (`pkg/cache/scheduler/cache.go`, `pkg/cache/queue/manager.go`)
This was introduced in #9487 for the `replica_role` label during HA leader election. When a follower becomes the `leader`, it deletes all gauge series that have the old role label. Then it immediately reports all gauge metrics again for every object, using the new role label.

Both approaches solve the same problem: a gauge series with an outdated label value remains in Prometheus unless it is explicitly deleted. However, they use different methods to handle this.

**Why is this needed**:

Aligning these two patterns could reduce duplicated logic and make it easier to manage the lifecycle of gauge metrics in one place.

See https://github.com/kubernetes-sigs/kueue/pull/9774#discussion_r2924664021.

PR #10276, opened 2026-04-02
metrics: consolidate gauge cleanup scopes

MatteoFari:
#### What type of PR is this?
/kind cleanup

This PR cleans up how Kueue removes old Prometheus gauge metrics.

Right now, Kueue has two different ways to clean up stale gauge metrics. This PR keeps both flows working the same way, but makes them use the same cleanup code.

Partially addresses #9854

```release-note
NONE
```

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -269,13 +269,32 @@ var (
 	CohortSubtreeAdmittedActiveWorkloads *prometheus.GaugeVec
 )
 
-func trackGaugeVec(g *prometheus.GaugeVec) *prometheus.GaugeVec {
-	allGaugeVecs = append(allGaugeVecs, g)
+type gaugeCleanupScope uint8
+
+const (
+	gaugeCleanupScopeRole gaugeCleanupScope = iota
+	gaugeCleanupScopeClusterQueue
+	gaugeCleanupScopeClusterQueueLabelChange
+	gaugeCleanupScopeClusterQueueCache
+	gaugeCleanupScopeClusterQueueResource
+	gaugeCleanupScopeLocalQueue
+	gaugeCleanupScopeLocalQueueCache
+	gaugeCleanupScopeLocalQueueResource
+	gaugeCleanupScopeCohort
+)
+
+var gaugeVecsByScope map[gaugeCleanupScope][]*prometheus.GaugeVec
+
+func trackGaugeVec(g *prometheus.GaugeVec, scopes ...gaugeCleanupScope) *prometheus.GaugeVec {
+	gaugeVecsByScope[gaugeCleanupScopeRole] = append(gaugeVecsByScope[gaugeCleanupScopeRole], g)
+	for _, scope := range scopes {
+		gaugeVecsByScope[scope] = append(gaugeVecsByScope[scope], g)
+	}
 	return g
 }
 
 func InitMetricVectors(extraLabels []string) {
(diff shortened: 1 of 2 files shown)


pkg/metrics/metrics.go
```diff
@@ -300,14 +319,15 @@ The label 'result' can have the following values:
 		}, []string{"result", "replica_role"},
 	)
 
-	AdmissionCyclePreemptionSkips = trackGaugeVec(prometheus.NewGaugeVec(
+	AdmissionCyclePreemptionSkips = prometheus.NewGaugeVec(
```
IrvingMg, line 322: Can you check that `make generate-metrics-tables` produces no diff? This change updates metric initialization, and the doc generator parses these AST patterns.

MatteoFari: yes, no diff

MatteoFari:
@IrvingMg do you prefer a separate PR, or would you rather keep everything in this PR?

pkg/metrics/metrics.go
```diff
@@ -933,27 +973,25 @@ func ClearClusterQueueMetrics(cq kueue.ClusterQueueReference) {
...
+	clearScopedGaugeMetrics(gaugeCleanupScopeLocalQueue, lbls)
+	LocalQueueQuotaReservedWorkloadsTotal.DeletePartialMatch(lbls)
+	LocalQueueQuotaReservedWaitTime.DeletePartialMatch(lbls)
+	LocalQueueFinishedWorkloadsTotal.DeletePartialMatch(lbls)
+	LocalQueueAdmittedWorkloadsTotal.DeletePartialMatch(lbls)
+	LocalQueueAdmissionWaitTime.DeletePartialMatch(lbls)
+	LocalQueueAdmissionChecksWaitTime.DeletePartialMatch(lbls)
+	LocalQueueQueuedUntilReadyWaitTime.DeletePartialMatch(lbls)
+	LocalQueueAdmittedUntilReadyWaitTime.DeletePartialMatch(lbls)
+	LocalQueueEvictedWorkloadsTotal.DeletePartialMatch(lbls)
```
mimowo, line 990: Actually, this is an interesting approach. I'm also interested if this is actually a bugfix, because some of the guage metrics were not cleaned here before, for example LocalQueueReservingActiveWorkloads, wdyt? Maybe this is a corner case, but IIUC if LQ would be deleted then LocalQueueReservingActiveWorkloads, wdyt?

If this is the case then I think it makes sense to actually move forward the PR as an independent body of work / bugfix even if this does not fully cover the original intent of the issue. cc @IrvingMg

IrvingMg: I'm not sure this is actually fixing a bug. `LocalQueueReservingActiveWorkloads` is already cleaned up in `ClearLocalQueueCacheMetrics`, which is called on LocalQueue deletion:
https://github.com/kubernetes-sigs/kueue/blob/9494dcc26a3907a3ec2efd9f2dcdca0b2e0d2720/pkg/cache/scheduler/clusterqueue.go#L640-L650

But the refactoring looks good, so I think we can keep this as a separate preparatory PR.

mimowo: I see, thanks for clarifying, in that case let me merge purely as a cleanup. I also think it is better to split into smaller PRs.

mimowo:
Thank you 👍 
/lgtm
/approve

PR #10279, opened 2026-04-02
[Cleanup] Return boolean for hierarchy modification in AddOrUpdateCohort

mszadkow:
#### What type of PR is this?
/kind cleanup

Prep for [Provide a metric for cohort and clusterqueue info](https://github.com/kubernetes-sigs/kueue/pull/10004)

Relates #7539

```release-note
NONE
```

### DIFF
--- a/pkg/cache/scheduler/cache.go
+++ b/pkg/cache/scheduler/cache.go
@@ -524,20 +524,27 @@ func (c *Cache) DeleteClusterQueue(cq *kueue.ClusterQueue) {
 	}
 }
 
-func (c *Cache) AddOrUpdateCohort(apiCohort *kueue.Cohort) error {
+// AddOrUpdateCohort adds a cohort to the cache or updates its hierarchy and properties if it already exists.
+// Returns a boolean indicating if the cohort hierarchy has changed, and an error if any issue occurs during the update.
+func (c *Cache) AddOrUpdateCohort(apiCohort *kueue.Cohort) (bool, error) {
 	c.Lock()
 	defer c.Unlock()
 	cohortName := kueue.CohortReference(apiCohort.Name)
 	c.hm.AddCohort(cohortName)
 	cohort := c.hm.Cohort(cohortName)
 	oldParent := cohort.Parent()
+
 	c.hm.UpdateCohortEdge(cohortName, apiCohort.Spec.ParentName)
-	err := cohort.updateCohort(apiCohort, oldParent)
-	if err != nil {
-		return err
+	if err := cohort.updateCohort(apiCohort, oldParent); err != nil {
+		return false, err
 	}
 	c.handleParentUpdate(oldParent)
-	return nil
+
+	var oldParentName kueue.CohortReference
+	if oldParent != nil {
+		oldParentName = oldParent.Name
+	}
+	return oldParentName != apiCohort.Spec.ParentName, nil
 }
 
 func (c *Cache) DeleteCohort(cohortName kueue.CohortReference) {
--- a/pkg/cache/scheduler/cache_test.go
+++ b/pkg/cache/scheduler/cache_test.go
@@ -1011,7 +1011,8 @@ func TestCacheClusterQueueOperations(t *testing.T) {
 			name: "create cohort",
 			operation: func(log logr.Logger, cache *Cache) error {
 				cohort := utiltestingapi.MakeCohort("cohort").Obj()
-				return cache.AddOrUpdateCohort(cohort)
+				_, err := cache.AddOrUpdateCohort(cohort)
+				return err
 			},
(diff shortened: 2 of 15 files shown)


pkg/cache/scheduler/cache.go
```diff
@@ -522,15 +522,32 @@ func (c *Cache) DeleteClusterQueue(cq *kueue.ClusterQueue) {
 	}
 }
 
-func (c *Cache) AddOrUpdateCohort(apiCohort *kueue.Cohort) error {
+// AddOrUpdateCohort returns (hierarchyChanged, error), hierarchyChanged is true
+// when the cohort's parent or root cohort changed.
```
```suggestion
// AddOrUpdateCohort adds a cohort to the cache or updates its hierarchy and properties if it already exists.
// Returns a boolean indicating if the cohort hierarchy has changed, and an error if any issue occurs during the update.
```

pkg/cache/scheduler/cache.go
```diff
@@ -522,15 +522,32 @@ func (c *Cache) DeleteClusterQueue(cq *kueue.ClusterQueue) {
...
+	}
+
 	c.hm.UpdateCohortEdge(cohortName, apiCohort.Spec.ParentName)
-	return cohort.updateCohort(apiCohort, oldParent)
+	if err := cohort.updateCohort(apiCohort, oldParent); err != nil {
+		return false, err
+	}
+
+	newParentName := apiCohort.Spec.ParentName
+	newRootName := cohort.getRootUnsafe().Name
```
mbobrovskyi: Actually, I’m not sure we need to check the root. I think checking whether the parent changed is enough.

mszadkow: Maybe here it's not visible why, but we will need it as this is trigger to update the hierarchy info in the metrics

pkg/cache/scheduler/cache.go
```diff
@@ -522,15 +522,32 @@ func (c *Cache) DeleteClusterQueue(cq *kueue.ClusterQueue) {
...
+	hierarchyChanged := oldParentName != newParentName || oldRootName != newRootName
```
vladikkuzn: Stupid question: can we say that there's a change in the hierarchy if parent hasn't changed, root hasn't changed, but there're changes in another brach of the hierarchy?

mszadkow: Our goal here is to detect hierarchy changed that refers to the cohort.
The concern is either parent or root has changed.

pkg/cache/scheduler/cache.go
mszadkow: root is important for the reparenting with root change

mbobrovskyi: We update apiCohort, right? After that, we check newParentName. If it changed, then we return true. But I still don’t understand how it’s possible for the root to change if the parent didn’t.

mszadkow: I thought it will be in the scenario when we do the reparenting on the higher levels, but after some investigation I think you are correct.
The root should be already updated and we can't detect it here.

mszadkow: @mbobrovskyi updated

pkg/cache/scheduler/cache.go
```diff
@@ -522,15 +522,28 @@ func (c *Cache) DeleteClusterQueue(cq *kueue.ClusterQueue) {
...
+	}
+
 	c.hm.UpdateCohortEdge(cohortName, apiCohort.Spec.ParentName)
-	return cohort.updateCohort(apiCohort, oldParent)
+	if err := cohort.updateCohort(apiCohort, oldParent); err != nil {
+		return false, err
+	}
+
+	newParentName := apiCohort.Spec.ParentName
+	hierarchyChanged := oldParentName != newParentName
```
```suggestion
	hierarchyChanged := oldParentName != apiCohort.Spec.ParentName
```
nit

pkg/cache/scheduler/cache.go
```diff
@@ -522,15 +522,28 @@ func (c *Cache) DeleteClusterQueue(cq *kueue.ClusterQueue) {
...
 	c.Lock()
 	defer c.Unlock()
 	cohortName := kueue.CohortReference(apiCohort.Name)
 	c.hm.AddCohort(cohortName)
 	cohort := c.hm.Cohort(cohortName)
 	oldParent := cohort.Parent()
+	var oldParentName kueue.CohortReference
+	if oldParent != nil {
+		oldParentName = oldParent.Name
+	}
```
mbobrovskyi: nit: can we move it after updateCohort?

mbobrovskyi (commented):
Overall LGTM.

mbobrovskyi (commented):
/lgtm

Thank you!

PR #10280, opened 2026-04-02
Minor fixes and extra logs around fair sharing

mwielgus:
#### What type of PR is this?

/kind cleanup

Minor fixes and extra logs around Fair Sharing.

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/preemption/fairsharing/ordering.go
+++ b/pkg/scheduler/preemption/fairsharing/ordering.go
@@ -152,6 +152,9 @@ func (t *TargetClusterQueueOrdering) nextTarget(cohort *schdcache.CohortSnapshot
 		switch {
 		case (!drs.IsBorrowing() && cq != t.preemptorCq) || !t.hasWorkload(cq):
 			t.prunedClusterQueues.Insert(cq)
+			if logV := t.log.V(6); logV.Enabled() {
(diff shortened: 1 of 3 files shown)


pkg/scheduler/preemption/fairsharing/ordering.go
```diff
@@ -152,6 +152,9 @@ func (t *TargetClusterQueueOrdering) nextTarget(cohort *schdcache.CohortSnapshot
 		switch {
 		case (!drs.IsBorrowing() && cq != t.preemptorCq) || !t.hasWorkload(cq):
 			t.prunedClusterQueues.Insert(cq)
+			if logV := t.log.V(6); logV.Enabled() {
+				logV.Info("Pruning ClusterQueue during Target ordering", "clusterQueue", cq.GetName(), "drs", drs)
+			}
```
mbobrovskyi, line 157: Why we are checking is it enabled? Can we just do like this?

```suggestion
			t.log.V(6).Info("Pruning ClusterQueue during Target ordering", "clusterQueue", cq.GetName(), "drs", drs)
```

gabesaba: it avoids function argument evaluation, which is useful in hotspots. I'm not convinced that this is a hotspot, but there's no harm in this pattern besides a few extra lines of code (and this ensures the change wouldnt introduce a hotspot in a critical path)

PR #10283, opened 2026-04-03
KEP-2941: Add partitionable devices support

### LINKED ISSUE #10108: Support DRA Partitionable Devices (KEP-4815) in Kueue
**What would you like to be added:**
When a DRA driver publishes devices with `SharedCounters` and `ConsumesCounters` fields ([KEP-4815](https://github.com/kubernetes/enhancements/issues/4815)), the kube-scheduler tracks counter consumption to prevent conflicting device allocations. Kueue needs counter-based quota management for these partitionable devices instead of device-count quota. For example, an NVIDIA H100 GPU can be partitioned into MIG profiles (1g.10gb, 2g.20gb, 7g.80gb) that consume different amounts of GPU memory and compute slices. Device-count quota treats all profiles equally, while counter-based quota accurately reflects that a 7g.80gb profile consumes 8x more GPU memory than a 1g.10gb profile.

**Why is this needed:**

GPU partitioning via MIG is a key use case for maximizing GPU utilization. Without counter-based quota, admins cannot enforce fair sharing across teams using different MIG profile sizes. A team using four 1g.10gb profiles (40Gi total memory) would consume the same quota as a team using four 7g.80gb profiles (320Gi total memory), even though the latter uses 8x more physical resources.
  
**This enhancement requires the following artifacts:**

  - [x] Design doc: https://github.com/kubernetes-sigs/kueue/pull/10283
  - [x] Implementation: https://github.com/kubernetes-sigs/kueue/pull/10320
  - [ ] Docs update: https://github.com/kubernetes-sigs/kueue/pull/11844

  Depends on #9742 (CEL expression support)
  Part of #8243

sohankunkerkar:
#### What type of PR is this?
/kind documentation

Partially fixes https://github.com/kubernetes-sigs/kueue/issues/10108

<details>
<summary>1. ResourceSlice published by NVIDIA DRA driver</summary>

```sh
$ kubectl get resourceslice gcp-a100-test-gfn26-worker-f-gz764-gpu.nvidia.com-tt5gs \
    -o jsonpath='{.spec.sharedCounters[0]}'
gpu-0-counter-set: memory=40320Mi

$ kubectl get resourceslice gcp-a100-test-gfn26-worker-f-gz764-gpu.nvidia.com-tt5gs \
    -o jsonpath='{range .spec.devices[*]}{.name}: {.consumesCounters[0].counters.memory.value}{"\n"}{end}' \
    | grep 'mig-1g\|mig-7g'
gpu-0-mig-1g10gb-15-0: 9984Mi
gpu-0-mig-7g40gb-0-0: 40320Mi
```

</details>

<details>
<summary>2. Kueue config</summary>

```yaml
counterMappings:
- name: gpu.memory
  driver: gpu.nvidia.com
  counterName: memory
```

</details>

<details>
<summary>3. ClusterQueue with counter quota</summary>

```sh
$ kubectl get clusterqueue pd-test-cq \
    -o jsonpath='{range .spec.resourceGroups[0].flavors[0].resources[*]}{.name}: {.nominalQuota}{"\n"}{end}'
cpu: 100
memory: 200Gi
gpu.memory: 40Gi
```

</details>

<details>
<summary>4. Workload admitted with counter-based charge</summary>

```sh
$ kubectl get workload job-mig-1g-test-8be29 -n pd-test \
    -o jsonpath='{.status.admission.podSetAssignments[0].resourceUsage}'
{"cpu":"100m","gpu.memory":"9984Mi","memory":"256Mi"}
```

`1g.10gb` charges `9984Mi` and `7g.40gb` would charge `40320Mi`. Without counter-based quota, both would charge "1 GPU" via `deviceClassMappings` despite `7g.40gb` using 4x the memory.

</details>

```release-note
NONE
```

### DIFF
--- a/keps/2941-DRA/README.md
+++ b/keps/2941-DRA/README.md
@@ -33,6 +33,7 @@ tags, and then generate with `hack/update-toc.sh`.
     - [Story 2](#story-2)
     - [Story 3](#story-3)
     - [Story 4](#story-4)
+    - [Story 5](#story-5)
   - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
   - [Risks and Mitigations](#risks-and-mitigations)
 - [Design Details](#design-details)
@@ -54,6 +55,14 @@ tags, and then generate with `hack/update-toc.sh`.
     - [DeviceClass Resolution via Field Indexer](#deviceclass-resolution-via-field-indexer)
     - [DeviceClass Lifecycle Scenarios](#deviceclass-lifecycle-scenarios)
     - [Late DeviceClass Creation](#late-deviceclass-creation)
+  - [Partitionable Devices](#partitionable-devices)
+    - [ResourceSlice Structure](#resourceslice-structure)
+    - [User Workload](#user-workload)
+    - [Configuration](#configuration-1)
+    - [Processing Flow](#processing-flow-1)
+    - [Path Interactions](#path-interactions)
+    - [Counter Lifecycle Scenarios](#counter-lifecycle-scenarios)
+    - [Validation](#validation)
   - [Architecture Details](#architecture-details)
     - [Queue Manager Extensions](#queue-manager-extensions)
   - [Integration with Admission Fair Sharing](#integration-with-admission-fair-sharing)
@@ -67,6 +76,7 @@ tags, and then generate with `hack/update-toc.sh`.
     - [Alpha](#alpha)
       - [KueueDRAIntegration (v0.14)](#kueuedraintegration-v014)
       - [KueueDRAIntegrationExtendedResource (v0.17)](#kueuedraintegrationextendedresource-v017)
+      - [KueueDRAIntegrationPartitionableDevices (v0.18)](#kueuedraintegrationpartitionabledevices-v018)
     - [Beta](#beta)
       - [KueueDRAIntegration (v0.18)](#kueuedraintegration-v018)
       - [KueueDRAIntegrationExtendedResource](#kueuedraintegrationextendedresource)
@@ -82,6 +92,12 @@ tags, and then generate with `hack/update-toc.sh`.
   - [Defining DeviceClass mapping in ClusterQuota](#defining-deviceclass-mapping-in-clusterquota)
   - [Using ResourceFlavor for DeviceClass Mapping](#using-resourceflavor-for-deviceclass-mapping)
   - [Creating a new CRD for device class mapping](#creating-a-new-crd-for-device-class-mapping)
+  - [User Annotation as Primary Counter Consumption Mechanism](#user-annotation-as-primary-counter-consumption-mechanism)
+  - [Separate counterMappings Struct](#separate-countermappings-struct)
+  - [Device-Count Quota with Dual Tracking](#device-count-quota-with-dual-tracking)
+  - [Auto-discovery of Counters Without Configuration](#auto-discovery-of-counters-without-configuration)
+- [Appendix](#appendix)
+  - [Consumable Capacity Compatibility](#consumable-capacity-compatibility)
 <!-- /toc -->
 
 ## Summary
@@ -89,9 +105,13 @@ tags, and then generate with `hack/update-toc.sh`.
 [Dynamic Resource Allocation (DRA)](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/)
 is a major effort to improve device support in Kubernetes. It changes how one can request resources in a myriad of ways.
 
-This KEP supports two approaches for DRA integration with Kueue:
+This KEP supports three approaches for DRA integration with Kueue:
 1. **ResourceClaimTemplates**: Pods explicitly reference ResourceClaimTemplates that specify device requests.
 2. **Extended Resources**: Pods request DRA devices via standard `resources.requests` (e.g., `example.com/gpu: 1`), and kube-scheduler automatically creates ResourceClaims when the DeviceClass has an `extendedResourceName` field set.
+3. **Partitionable Devices**: Counter-based quota for devices that can be dynamically
+   partitioned (e.g., NVIDIA MIG). Instead of counting devices, Kueue tracks counter
+   consumption (e.g., GPU memory) from the `SharedCounters` and `ConsumesCounters` fields
+   defined by [KEP-4815](https://github.com/kubernetes/enhancements/issues/4815).
 
 ## Motivation
 
@@ -191,13 +211,22 @@ a simple device class named `gpu.example.com`. This will be the way to enforce q
 - Users can submit workloads using extended resource requests (e.g., `example.com/gpu: 1`) and
   Kueue can account for quota when the DeviceClass has `extendedResourceName` set.
 - Admins can enforce the quota for number of devices for a given DeviceClass.
+- Admins can enforce counter-based quota for partitionable devices (e.g., GPU memory quota
+  instead of device count quota for MIG profiles).
 
 ### Non-Goals
 
 - Quota-aware handling of DRAAdminAccess and DRAPrioritizedLists (beta, default enabled in K8s 1.35)
   is not included in Kueue's alpha. See [Risks and Mitigations](#risks-and-mitigations) for the
   planned approach.
-- Support for alpha DRA features like DRADeviceTaints and DRAPartitionableDevices will not be included.
+- Support for DRA features like DRADeviceTaints is not included.
+- Multi-host partitionable devices (e.g., NVLink fabrics spanning multiple nodes) are not
+  supported.
+- Support for DRAConsumableCapacity (alpha in K8s 1.34, beta targeting K8s 1.36) is not
+  included. Consumable capacity enables software-level device sharing (e.g., MPS, time-slicing)
+  where multiple claims share one device. This requires different quota semantics as Kueue
+  would need to read the cost from the claim's capacity request rather than the device's
+  counters. No GPU driver ships consumable capacity yet.
 - This design does not work with Topology Aware Scheduling feature of Kueue. It is a significant
   amount of work, will be addressed in the future with a separate body of work.
 
@@ -214,6 +243,10 @@ scheduling. This includes:
 4. Allowing admins to define quota for DRA resources in ClusterQueues using the logical resource names from device class
    mappings
 5. Implementing validation to prevent device class conflicts and ensure predictable quota behavior
+6. Extending `deviceClassMappings` with a `sources` field to support counter-based quota for
+   partitionable devices (requires Kubernetes `DRAPartitionableDevices` feature gate, beta in
+   K8s 1.36). This builds on CEL expression support which adds ResourceSlice access and
+   device matching to Kueue.
 
 More details are documented in [Design Details](#design-details)
 
@@ -239,6 +272,13 @@ As a Kueue user, I want to request DRA devices using standard resource requests 
 instead of ResourceClaimTemplates, so my existing workloads can benefit from DRA without modification when the cluster
 administrator configures DeviceClasses with `extendedResourceName`.
 
+#### Story 5
+
+As a cluster administrator, I want to enforce GPU memory quota for MIG partitions so that teams
+sharing a pool of partitionable GPUs get fair access based on counter consumption, not
+just device counts. A team requesting a 1g.10gb MIG profile should consume about 9856Mi of
+GPU memory quota, while a team requesting a 7g.80gb profile should consume 80Gi.
+
 ### Notes/Constraints/Caveats (Optional)
 
 - The `ResourceClaims` and `ResourceClaimTemplates` APIs for DRA in k8s are immutable.
@@ -247,9 +287,13 @@ administrator configures DeviceClasses with `extendedResourceName`.
 - Device class uniqueness is enforced - each device class can only map to one resource name to prevent quota ambiguity.
 - Configuration-based approach - device class mappings are configured through the Kueue Configuration API
 - This design does not work with Kueue's Topology Aware Scheduling feature and will be addressed in future work.
+- DRA resource preprocessing is not scoped by ResourceFlavor node constraints. Counter
+  charges and device matching are computed globally before flavor assignment.
 - Quota-aware handling of DRAAdminAccess and DRAPrioritizedLists (beta in K8s 1.35) is deferred
-  to beta graduation. Alpha DRA features like DRADeviceTaints and DRAPartitionableDevices are
-  not supported.
+  to beta graduation. DRADeviceTaints is not supported in alpha.
+- **Single-node partitionable devices (e.g., MIG) are supported** via counter-based
+  quota. See [Partitionable Devices](#partitionable-devices). Multi-host partitionable
+  devices are not supported.
 - **Extended Resources** requires DeviceClasses to have `spec.extendedResourceName` set.
   This depends on the Kubernetes `DRAExtendedResource` feature gate (alpha in k8s 1.35).
   When enabled, kube-scheduler automatically creates ResourceClaims for pods requesting extended resources.
@@ -266,7 +310,6 @@ administrator configures DeviceClasses with `extendedResourceName`.
   With structured parameters, GPU sharing is supported via ResourceClaimTemplates where
   containers within the same pod share a GPU. Cross-pod sharing via direct ResourceClaims
   is not supported.
-
 - **Kueue does not validate DeviceClass existence at config load time.** Admins should
   create DeviceClasses before submitting workloads but strict ordering is not enforced.
 
@@ -278,12 +321,12 @@ administrator configures DeviceClasses with `extendedResourceName`.
   time on a best-effort basis. Workloads with CEL selectors that match fewer devices than requested are rejected
   to prevent quota leaks. This validation uses the upstream DRA CEL compiler from [`k8s.io/dynamic-resource-allocation/cel`](https://github.com/kubernetes/dynamic-resource-allocation/tree/master/cel).
   On the other hand, devices can be allocated between Kueue's check and scheduling, and new ResourceSlices published after
-  validation can make previously-unsatisfiable workloads. `WaitForPodsReady` serves as the safety net for cases where
-  the validation state diverges from actual device availability at scheduling time.
-
-- **Kueue does not have any informers on ResourceSlice changes.** CEL expression evalulation
-requires that the ResourceSlices are fetched from the apiserver. If a driver publishes new capacity and the workload was rejected, there is no
-resubmission of the workload once ResourceSlices changes. Evaluating this for beta is important.
+  validation can make previously-unsatisfiable workloads satisfiable. Kueue does not
+  currently have a ResourceSlice informer. Inadmissible workloads are only re-evaluated
+  when the ClusterQueue is notified through other events such as quota changes. Adding
+  event-driven requeuing on ResourceSlice changes is an Alpha graduation criterion.
+  `WaitForPodsReady` serves as the safety net for cases where the validation state
+  diverges from actual device availability at scheduling time.
 
 ### Risks and Mitigations
 
@@ -295,9 +338,8 @@ See [Workload Rejection When DRA Is Disabled](#workload-rejection-when-dra-is-di
 
 With DRAAdminAccess and DRAPrioritizedLists (both beta, default enabled in K8s 1.35), there is a
 risk that effective tallying of resources will not be available until after allocation of these
-resources by kube-scheduler.
-
-In order to mitigate this risk, Kueue can take the following approach:
+resources by kube-scheduler. Support for these is deferred to Beta but the mitigation approach
+is documented here:
 1. For DRAPrioritizedLists: all the mentioned device classes in the request will be counted against the quota
 2. For DRAAdminAccess: This feature can only be enabled in admin namespace, therefore it should be skipped for being
    counted against ClusterQuota. This is a different stance than Kubernetes ResourceQuota because kubernetes
@@ -336,6 +378,10 @@ Three feature gates control DRA support in Kueue:
 - `DRAExtendedResources`: gates extended resources support, including DeviceClass
   auto-discovery via `extendedResourceName`. Does not use `deviceClassMappings`.
   Requires `DynamicResourceAllocation` to also be enabled.
+- `KueueDRAIntegrationPartitionableDevices`: gates counter-based quota for partitionable
+  devices. Enables the `sources` field on `deviceClassMappings` entries. Requires
+  `DynamicResourceAllocation` to also be enabled. Also requires the Kubernetes
+  `DRAPartitionableDevices` feature gate (beta in K8s 1.36).
 - `KueueDRARejectWorkloadsWhenDRADisabled` (default: enabled, Beta since v0.18): rejects workloads that
   use DRA resources (ResourceClaimTemplates or ResourceClaims) when the
   `DynamicResourceAllocation` feature gate is disabled. Without this gate, DRA workloads
@@ -437,7 +483,7 @@ spec:
 
 The above ClusterQueue is an example configuration of a queue, with half quota configured for single allocation of
 example
-GPUs, and half quota configured for GPUs that are shared by workloads. Similarly, when DRAPartitionableDevices feature
+GPUs, and half quota configured for GPUs that are shared by workloads. Similarly, when KueueDRAIntegrationPartitionableDevices feature
 is supported in kubernetes, GPUs partitions can be represented by a single device class.
 
 ### Device Class Resolution and Conflict Prevention
@@ -521,7 +567,8 @@ rules:
 **Required Permissions:**
 - `resourceclaims`: Read access to validate ResourceClaim references (though not supported for quota)
 - `resourceclaimtemplates`: Read access to process ResourceClaimTemplates and extract device class information
-- `resourceslices`: Read access to list cluster devices for CEL selector validation
+- `resourceslices`: Read access to list cluster devices for CEL selector validation and `consumesCounters` reading.
+  Kueue only reads device attributes for CEL matching and counter values for quota.
 - `deviceclasses`: Read access to resolve DeviceClass selectors for device pre-filtering during CEL evaluation
 
 **Security Considerations:**
@@ -774,7 +821,9 @@ An extended resource can be identified by verifying that qualified resource name
 
 The extended resources path does not require `deviceClassMappings`. Kueue auto-discovers
 DeviceClasses via a field indexer on `spec.extendedResourceName` and uses the
-`extendedResourceName` directly as the quota key in ClusterQueue.
+`extendedResourceName` as the default quota key. If the DeviceClass is also in
+`deviceClassMappings`, Kueue uses the mapped logical name instead to unify quota
+with the ResourceClaimTemplate path.
 
 DeviceClass with `extendedResourceName` (DeviceClass API is v1/GA, but the `extendedResourceName`
 field requires the Kubernetes `DRAExtendedResource` feature gate):
@@ -809,14 +858,16 @@ continues to use `deviceClassMappings`.
 
 #### Path Separation
 
-The two DRA paths have independent quota resolution:
+The two DRA paths resolve quota independently:
 1. **Extended resources** (`DRAExtendedResources` gate): auto-discovers DeviceClass via
-   field indexer, uses `extendedResourceName` as quota key. No `deviceClassMappings` needed.
+   field indexer. Uses `extendedResourceName` as the default quota key. If the resolved
+   DeviceClass is also present in `deviceClassMappings`, Kueue uses the mapped logical
+   name instead to unify quota with the ResourceClaimTemplate path. If the mapping has
+   counter sources configured, the workload is marked inadmissible because extended resources
+   do not carry the profile-level information needed for counter-based charging.
 2. **ResourceClaimTemplates** (`DynamicResourceAllocation` gate): uses `deviceClassMappings`
-   to map DeviceClass names to logical resource names. No auto-discovery.
-
-There is no precedence or fallback between the two paths. Each path has exactly one resolution
-mechanism.
+   to map DeviceClass names to logical resource names. When the mapping has counter sources
+   configured, charges counter units instead of device count.
 
 #### Processing Flow
 
@@ -824,10 +875,14 @@ mechanism.
 2. Looks up DeviceClass by `extendedResourceName` by field indexer
 3. If no matching DeviceClass is found, the resource is not DRA-backed and Kueue
    processes it through the standard resource quota path (counted via `node.Status.Allocatable`)
-4. If a matching DeviceClass is found, uses the `extendedResourceName` as the quota key
-5. Removes original extended resource from the workload's effective resource requests
+4. If a matching DeviceClass is found, resolves the quota key. If the DeviceClass is also
+   in `deviceClassMappings`, uses the mapped logical name. Otherwise uses `extendedResourceName`.
+5. If the mapping has counter sources configured, the workload is marked inadmissible.
+   Extended resources do not carry profile-level information for counter-based charging.
+   Otherwise charges device count.
+6. Removes original extended resource from the workload's effective resource requests
    (tracked internally per PodSet) to avoid double-counting
-6. Admits workload against the quota for the `extendedResourceName`
+7. Admits workload against the resolved quota key
 
 The extended resource translation reads directly from the workload spec before
 `excludeResourcePrefixes` filtering is applied. The processing order:
@@ -913,6 +968,382 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation 
 (e.g., triggered by other workload or queue events) re-evaluates the workload after the
 DeviceClass is created. Event-driven re-reconciliation on DeviceClass creation will be
 evaluated as a Beta graduation criterion.
+
+### Partitionable Devices
+
+This section is gated behind the `KueueDRAIntegrationPartitionableDevices` Kueue feature gate.
+
+Kueue supports counter-based quota for partitionable DRA devices as defined by
+Kubernetes [KEP-4815](https://github.com/kubernetes/enhancements/issues/4815). Instead of
+counting devices, Kueue tracks counter consumption (e.g., GPU memory) from the
+`SharedCounters` and `ConsumesCounters` fields on ResourceSlices.
+
+Counter-based resources fit into Kueue's existing (Flavor, Resource) quota model.
+Borrowing, lending, cohorts, preemption, and fair sharing work with counter resources.
+The `deviceSelector` ensures accurate charging by narrowing the accounting
+domain. See [Processing Flow](#processing-flow-1) for details.
+
+#### ResourceSlice Structure
+
+Starting in K8s 1.35, `SharedCounters` and `Devices` are mutually exclusive in a single
+ResourceSlice (`+zeroOrOneOf=ResourceSliceType`). Drivers must split them into separate
+slices in the same pool. On K8s 1.34 (where partitionable devices are alpha) this
+validation does not apply and some drivers put both in one slice
+(`resourceSliceCount: 1`). Pool completeness checks `len(slices) == resourceSliceCount`
+so both layouts work.
+
+The example below shows the separate-slice layout. Only the `memory` counter is shown.
+The driver also
+publishes `multiprocessors` and `memory-slice-0` through `memory-slice-7` which the
+kube-scheduler uses for MIG placement but Kueue does not need for quota:
+
+```yaml
+# ResourceSlice 1: SharedCounters (total capacity for this GPU)
+spec:
+  driver: gpu.nvidia.com
+  pool:
+    name: node1-gpu0
+    generation: 1
+    resourceSliceCount: 2
+  sharedCounters:
+  - name: gpu-0-counter-set
+    counters:
+      memory:
+        value: 80Gi
+      # Driver also publishes multiprocessors, memory-slice-0 through
+      # memory-slice-7 - used by kube-scheduler, not by Kueue.
+---
+# ResourceSlice 2: Devices with ConsumesCounters
+spec:
+  driver: gpu.nvidia.com
+  pool:
+    name: node1-gpu0
+    generation: 1
+    resourceSliceCount: 2
+  nodeName: node1
+  devices:
+  - name: gpu-0-mig-1g.10gb-0
+    attributes:
+      gpu.nvidia.com/profile:
+        string: "1g.10gb"
+    consumesCounters:
+    - counterSet: gpu-0-counter-set
+      counters:
+        memory:
+          value: 9856Mi
+  - name: gpu-0-mig-7g.80gb-0
+    attributes:
+      gpu.nvidia.com/profile:
+        string: "7g.80gb"
+    consumesCounters:
+    - counterSet: gpu-0-counter-set
+      counters:
+        memory:
+          value: 80Gi
+```
+
+#### User Workload
+
+Users request MIG profiles via CEL selectors on ResourceClaimTemplates. Counter tracking
+is transparent to the user:
+
+```yaml
+apiVersion: resource.k8s.io/v1
+kind: ResourceClaimTemplate
+metadata:
+  name: mig-1g-10gb
+spec:
+  spec:
+    devices:
+      requests:
+      - name: gpu
+        exactly:
+          deviceClassName: mig.nvidia.com
+          selectors:
+          - cel:
+              expression: "device.attributes['gpu.nvidia.com'].profile == '1g.10gb'"
+```
+
+#### Configuration
+
+The existing `DeviceClassMapping` struct is extended with an optional `sources` field.
+When a counter source is present, Kueue tracks quota in counter units (e.g., GPU memory)
+instead of device count. This can allow whole-GPU and MIG DeviceClasses to share a single
+quota pool. See [Path Interactions](#path-interactions) for caveats on unified pool charging.
+
+The `DeviceClassMapping` struct is extended:
+
+```golang
+type DeviceClassMapping struct {
+    // ...existing fields (Name, DeviceClassNames)...
+
+    // Sources configures resource accounting sources for this mapping.
+    // Each source defines how quota is tracked for this DeviceClass.
+    // Currently only counter sources are supported (for partitionable devices).
+    // Extended resource requests that resolve to a DeviceClass with sources
+    // configured are marked inadmissible.
+    // Requires the KueueDRAIntegrationPartitionableDevices feature gate.
+    // +optional
+    Sources []DeviceClassSourceConfig `json:"sources,omitempty"`
+}
+
+// DeviceClassSourceConfig defines a resource accounting source for a DeviceClassMapping.
+// Exactly one of the source types must be set.
+type DeviceClassSourceConfig struct {
+    // Counter configures counter-based quota for partitionable devices.
+    // Maps a DRA driver counter to the parent DeviceClassMapping's Kueue quota resource.
+    // +optional
+    Counter *DeviceClassCounterSource `json:"counter,omitempty"`
+}
+
+// DeviceClassCounterSource identifies where to read counter data from and which counter to track.
+type DeviceClassCounterSource struct {
+    // Name is the counter name within the device's consumesCounters
+    // entries to track for quota. Must match a counter name published by
+    // the driver in ResourceSlice devices' consumesCounters field.
+    // Counter set names are per-device identifiers (e.g., gpu-0-counter-set,
+    // gpu-1-counter-set), so name matches across all counter sets
+    // for a given driver without requiring one mapping per device.
+    // +required
+    Name string `json:"name"`
+
+    // Driver is the DRA driver name used to filter relevant ResourceSlices.
+    // Must match the spec.driver field on ResourceSlice objects.
+    // +required
+    Driver string `json:"driver"`
+
+    // DeviceSelector scopes which devices are eligible for counter-based
+    // quota accounting. Typically matches a GPU model (e.g., productName)
+    // so all partition profiles on that model share one quota pool.
+    // Per-workload charging is determined by the workload's own
+    // ResourceClaimTemplate selector, which narrows to the requested profile.
+    // The selector is compiled at config load time using the upstream dracel
+    // compiler.
+    // +required
+    DeviceSelector resourcev1.DeviceSelector `json:"deviceSelector"`
+}
+```
+
+Multi-profile MIG setup sharing a single `gpu.memory` quota pool. The
+`deviceSelector` scopes the accounting domain to devices from the configured
+driver. Per-workload charging comes from the workload's own ResourceClaimTemplate selector, which narrows
+to the requested profile:
+
+```yaml
+# Kueue Configuration
+apiVersion: config.kueue.x-k8s.io/v1beta2
+kind: Configuration
+featureGates:
+  KueueDRAIntegration: true
+  KueueDRAIntegrationPartitionableDevices: true
+resources:
+  deviceClassMappings:
+  - name: gpu.memory
+    deviceClassNames: [mig.nvidia.com]
+    sources:
+    - counter:
+        name: memory
+        driver: gpu.nvidia.com
+        deviceSelector:
+          cel:
+            expression: "device.driver == 'gpu.nvidia.com'"
+---
+# ClusterQueue: 10 A100 GPUs worth of memory
+apiVersion: kueue.x-k8s.io/v1beta2
+kind: ClusterQueue
+metadata:
+  name: gpu-queue
+spec:
+  resourceGroups:
+  - coveredResources: ["gpu.memory"]
+    flavors:
+    - name: a100-pool
+      resources:
+      - name: gpu.memory
+        nominalQuota: "800Gi"
+```
+
+Workloads requesting different MIG profiles share the same `gpu.memory` quota. Kueue
+matches devices using both the DeviceClass selectors and the workload's ResourceClaimTemplate selectors
+(step 4 in [Processing Flow](#processing-flow-1)), then reads `consumesCounters.memory`
+from the matched devices:
+
+```yaml
+# Workload A: requests 1g.10gb profile
+apiVersion: resource.k8s.io/v1
+kind: ResourceClaimTemplate
+metadata:
+  name: mig-small
+spec:
+  spec:
+    devices:
+      requests:
+      - name: gpu
+        exactly:
+          deviceClassName: mig.nvidia.com
+          count: 1
+          selectors:
+          - cel:
+              expression: "device.attributes['gpu.nvidia.com'].profile == '1g.10gb'"
+---
+# Workload B: requests 7g.80gb profile
+apiVersion: resource.k8s.io/v1
+kind: ResourceClaimTemplate
+metadata:
+  name: mig-large
+spec:
+  spec:
+    devices:
+      requests:
+      - name: gpu
+        exactly:
+          deviceClassName: mig.nvidia.com
+          count: 1
+          selectors:
+          - cel:
+              expression: "device.attributes['gpu.nvidia.com'].profile == '7g.80gb'"
+```
+
+Resulting quota usage against `gpu.memory: 800Gi`:
+
+| Workload | ResourceClaimTemplate selector | consumesCounters.memory | Charge |
+|----------|-------------|------------------------|--------|
+| A | `profile == '1g.10gb'` | 9856Mi | `gpu.memory: 9856Mi` |
+| B | `profile == '7g.80gb'` | 80Gi | `gpu.memory: 80Gi` |
+
+The `deviceSelector` does not select the MIG profile. It scopes which devices
+are eligible for counter-based accounting. The workload's own ResourceClaimTemplate selector narrows to the
+requested profile. Different profiles produce different charges against the same quota
+because Kueue reads the actual `consumesCounters` value from the matched devices. When a
+workload uses a broad ResourceClaimTemplate selector matching multiple profiles, Kueue charges conservatively
+using the maximum `consumesCounters` value across matched devices.
+
+`deviceSelector` is required when a counter source is configured. Single-profile
+DeviceClasses that match devices with identical counter values do not need counter-based
+quota and can use device-count quota instead.
+
+When `sources` is absent on a mapping, the mapping behaves exactly as today (device-count
+quota). This is backward compatible and covers non-GPU DRA devices where device count is
+the appropriate unit.
+
+Counter names are driver-specific, so the counter source's `name` maps them to the Kueue
+quota resource name. The admin chooses which counters to track for quota.
+
+#### Processing Flow
+
+Partitionable devices reuses the upstream `dracel` compiler from the CEL expression
+support but performs its own
+ResourceSlice listing with pool-aware processing, since the CEL validation path does not
+return matched device objects or do pool grouping:
+
+1. Kueue looks up the ResourceClaimTemplate's DeviceClass in `deviceClassMappings`
+2. If the mapping has counter sources configured, enters the counter-based path
+3. Lists ResourceSlices for the configured driver, groups them by pool and checks completeness
+4. Filters devices by `deviceSelector` to narrow the candidate pool
+5. From filtered devices, matches using DeviceClass selectors and the workload's
+   ResourceClaimTemplate request selectors using the `dracel` compiler
+6. For each matched device, resolves the counter charge:
+   - Device has `consumesCounters` containing the configured `name`: uses the
+     actual value (e.g., 9856Mi for a 1g.10gb MIG profile)
+   - Device has `consumesCounters` but `name` is not found: workload is
+     marked inadmissible
+   - Device has no `consumesCounters`: workload is marked inadmissible
+7. Uses the maximum consumption across matched devices as the per-device counter charge
+8. For requests with `count > 1`, multiplies per-device consumption by count
+
+Counter resources are injected through the existing `WithPreprocessedDRAResources` path.
+
+In Alpha, partitionable devices performs its own ResourceSlice listing independently from
+the CEL validation path because the two have different requirements: CEL validation only
+needs a match count, while counter processing needs matched device objects and pool-aware
+grouping. In Beta, these can be consolidated into a shared ResourceSlice listing layer.
+
+The workload is marked inadmissible if the matching device count is less than the
+requested count, or if ResourceSlice data is unavailable. Workloads are never admitted
+with zero counter charge when counter sources are configured. For each pool, Kueue only
+considers ResourceSlices with the latest generation. If a pool's slice count is less
+than its `resourceSliceCount`, the pool is incomplete and its devices are excluded from
+matching.
+
+For `count > 1`, the per-device consumption is multiplied by count.
+
+The Kubernetes API limits each device to 2 `consumesCounters` entries
+(`ResourceSliceMaxDeviceCounterConsumptionsPerDevice`), each referencing a different counter
+set. In Alpha, only single-node partitionable devices are supported where each device
+consumes from a single local counter set. Multi-counter-set semantics (e.g., a device
+consuming from both a local GPU pool and a shared NVLink fabric pool) will be addressed
+with multi-host support in future work.
+
+#### Path Interactions
+
+A `deviceClassMappings` entry uses either device-count or source-based quota, determined
+by whether `sources` is set. A DeviceClass appears in exactly one mapping entry.
+
+When a counter source is set, the charge comes from the matched device's `consumesCounters`.
+If the device has no `consumesCounters`, the workload is marked inadmissible.
+
+**Extended resources and counter sources are not supported together:**
+
+If an extended resource request resolves to a DeviceClass whose mapping has counter sources
+configured, the workload is marked inadmissible. Extended resources carry only a device
+count (e.g., `nvidia.com/gpu: 1`) without profile-level CEL selectors, so Kueue cannot
+determine an accurate counter charge. Workloads requiring counter-based quota should use
+ResourceClaimTemplates with CEL selectors.
+
+**Unified quota pools:**
+
+Whole-GPU and MIG DeviceClasses can share one mapping entry with counter sources configured.
+The `deviceSelector` must be compatible with both request shapes for unified
+charging to work. Workloads that need both a whole GPU and a MIG slice from the same
+counter pool should use ResourceClaimTemplates for both.
+
+**Cohort and borrowing:**
+
+Counter resources participate in Cohort borrowing and lending like any other resource.
+`nominalQuota`, `borrowingLimit`, and `lendingLimit` are all in counter units. A
+ClusterQueue with `nominalQuota: "0"` borrows counter capacity from other ClusterQueues
+or the Cohort. Counter resources use a distinct resource name (e.g., `gpu.memory`) so
+borrowing only operates within the same resource.
+
+#### Counter Lifecycle Scenarios
+
+1. **No counter data on matched devices**: the driver published devices without
+   `consumesCounters` entries. The workload is marked inadmissible. Drivers using
+   partitionable devices must publish `consumesCounters` on all devices for counter-based
+   quota to work.
+
+2. **Non-existent counter name**: the configured `name` does not match any entry in
+   matched devices' `consumesCounters`. The workload is marked inadmissible.
+
+3. **ResourceSlice changes after admission**: Kueue does not re-evaluate admitted workloads.
+   If ResourceSlices change and the scheduler cannot find a matching device, the pod stays
+   pending and `waitForPodsReady` evicts the workload. If the scheduler allocates a smaller
+   partition than what Kueue charged, the pod runs fine but quota stays over-reserved until
+   the workload finishes.
+
+4. **Driver restart**: ResourceSlices may temporarily disappear. Workloads submitted during
+   this window are marked inadmissible. Kueue does not currently have a ResourceSlice
+   informer, so inadmissible workloads are only re-evaluated when the ClusterQueue is
+   notified through other events. Event-driven requeuing on ResourceSlice changes is an
+   Alpha graduation criterion.
+
+#### Validation
+
+- At config load time: when `sources` is present on a `deviceClassMappings` entry, the
+  `KueueDRAIntegrationPartitionableDevices` feature gate must be enabled. `sources` must have exactly
+  one entry in Alpha, and it must be a counter source. Each counter source is validated for
+  required fields (`name`, `driver`, `deviceSelector`). The `deviceSelector` CEL expression
+  is compiled at config load time using the upstream `dracel` compiler to catch syntax
+  and type errors early. Exactly one source type must be set per entry.
+- Duplicate `(driver, name)` tuples within a single mapping's `sources` are
+  rejected. Across different mappings, the same `(driver, name)` is allowed
+  since DeviceClass uniqueness already prevents double-counting. This supports separate
+  quota for GPU models that share a driver and counter name (e.g., A100 vs H100).
+- At runtime: no cross-validation between `name` and actual ResourceSlice counter
+  names. This is consistent with how `deviceClassMappings` does not validate DeviceClass
+  existence at config load time.
+- `KueueDRAIntegrationPartitionableDevices` requires `KueueDRAIntegration` to be enabled. Validated
+  at startup in `pkg/config/validation.go`.
 
 ### Architecture Details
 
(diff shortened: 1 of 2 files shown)


keps/2941-DRA/README.md
```diff
@@ -182,13 +198,17 @@ a simple device class named `gpu.example.com`. This will be the way to enforce q
...
 
 ### Non-Goals
 
 - Quota-aware handling of DRAAdminAccess and DRAPrioritizedLists (beta, default enabled in K8s 1.35)
   is not included in Kueue's alpha. See [Risks and Mitigations](#risks-and-mitigations) for the
   planned approach.
-- Support for alpha DRA features like DRADeviceTaints and DRAPartitionableDevices will not be included.
+- Support for DRADeviceTaints will not be included in alpha.
+- Multi-host partitionable devices (e.g., NVLink fabrics spanning multiple nodes) are not
+  supported. These require TAS + DRA integration which will be addressed in future work.
```
kannon92:
```suggestion
  supported.
```

I am not sure we will even support this with TAS + DRA. MultiHost should probably go through the WAS work. I think let's just say its not supported and not hold us to an implementation that may or may not happen.

keps/2941-DRA/README.md
```diff
@@ -73,16 +82,23 @@ tags, and then generate with `hack/update-toc.sh`.
...
 ## Summary
 
 [Dynamic Resource Allocation (DRA)](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/)
 is a major effort to improve device support in Kubernetes. It changes how one can request resources in a myriad of ways.
 
-This KEP supports two approaches for DRA integration with Kueue:
+This KEP supports three approaches for DRA integration with Kueue:
 1. **ResourceClaimTemplates**: Pods explicitly reference ResourceClaimTemplates that specify device requests.
 2. **Extended Resources**: Pods request DRA devices via standard `resources.requests` (e.g., `example.com/gpu: 1`), and kube-scheduler automatically creates ResourceClaims when the DeviceClass has an `extendedResourceName` field set.
+3. **Partitionable Devices**: Counter-based quota for devices that can be dynamically
```
kannon92, line 111: Reading this I am wondering about our support for https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/5075-dra-consumable-capacity#proposal.

kannon92: Maybe for now we should call that consumable capacity devices are not yet supported? 
@alaypatel07 any idea how these two features are supposed to work together?

sohankunkerkar: Sure, I can add a short blurb about that. So the consumable capacity devices feature will require a separate effort as Kueue would need to read the cost from the claim's capacity request, which is a different data path.
NVIDIA’s DRA driver doesn’t support consumable capacity yet, even though the API is moving to beta in Kubernetes 1.36. The underlying GPU mechanisms (MPS and time-slicing) still have key limitations.
MPS lacks isolation i.e. one crashing CUDA process can affect all workloads on the GPU. Time-slicing has no memory isolation, which likely explains why these haven't integrated into their DRA driver yet.

kannon92 (commented):
LGTM!

/assign @mimowo @tenzen-y

keps/2941-DRA/README.md
```diff
@@ -800,6 +850,268 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+```yaml
+apiVersion: config.kueue.x-k8s.io/v1beta2
+kind: Configuration
+featureGates:
+  DynamicResourceAllocation: true
+  DRAPartitionableDevices: true
+resources:
+  counterMappings:
+  - name: gpu.memory
+    driver: gpu.nvidia.com
```
alaypatel07: do you mean `mig.nvidia.com` here?

alaypatel07: actually nevermind, I got confused with device class name, but upon reading back the reference here is the driver which is correctly referenced to gpu.nvidia.com

tenzen-y:
@sohankunkerkar can you resolve conflicts?

keps/2941-DRA/README.md
```diff
@@ -800,6 +850,288 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+
+When no devices match or ResourceSlice data is unavailable, the workload is marked
+inadmissible. Workloads are never admitted with zero counter charge when `counterMappings`
+are configured. Kueue only considers ResourceSlices from the pool with the highest
+generation. If the slice count is less than `resourceSliceCount`, the pool is incomplete
+and the workload is marked inadmissible.
+
+When CEL selectors match devices with different profiles (e.g., both 1g.10gb and 7g.80gb),
+Kueue charges the maximum consumption. Users should write specific CEL selectors that
+narrow to a single profile for accurate quota.
```
PannagaRao: When count > 1, the per-device charge (MAX) is multiplied by count. This can overcharge when matched devices have different counter values — for example, if a CEL selector matches devices with memory >= 80Gi  and 3 devices match (80Gi, 90Gi, 100Gi) with count=2, the charge is 100Gi × 2 = 200Gi, whereas the actual worst-case allocation is 100Gi + 90Gi = 190Gi. Should we add this as an explicit alpha limitation?

kannon92:
> @sohankunkerkar can you resolve conflicts?


@PannagaRao can address the feedback in the implementation. If we need to get the KEP merged I guess @PannagaRao can take over the KEP if you think its necessary.

keps/2941-DRA/README.md
sohankunkerkar: I added explicit note for the count > 1 case. The alpha behavior is intentionally conservative: max * count rather than top-N summation. In practice, users write profile-specific selectors so all matched devices have the same counter values and max * count equals actual consumption.

tenzen-y:
I'd like to hold this KEP until we reach an agreement on the https://github.com/kubernetes-sigs/kueue/pull/9841.
Because this seems to depend on the DRA CEL mechanism.

/hold

keps/2941-DRA/README.md
```diff
@@ -873,6 +922,293 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+The `Resources` struct is extended with a `CounterMappings` field:
+
+```golang
+type Resources struct {
+    // ...existing fields...
+
+    // CounterMappings defines mappings from DRA driver counters to logical resources
+    // for partitionable device quota management.
+    // +optional
+    CounterMappings []CounterMapping `json:"counterMappings,omitempty"`
```
alaypatel07: While this API is a good start for counter-based quota, I think it falls short for the case where an admin wants a single physical GPU pool to serve both whole-device and partitioned-device requests.

With the proposed API, the ClusterQueue ends up with two independent quota dimensions, one for whole devices and one for partitioned devices:
```
kind: ClusterQueue
metadata:
  name: mig-queue
spec:
  resourceGroups:
  - coveredResources: ["gpu.memory", "whole-gpus"]
    flavors:
    - name: partitionable-device
      resources:
      - name: gpu.memory
        nominalQuota: "320Gi"
    - name: whole-device
       resources:
       - name: whole-gups
          nominalQuota: "10"
```

There is no mechanism to express that these quotas are backed by the same physical devices, or that admitting a whole-GPU workload should reduce the available partitionable-device counter budget, and vice versa.

That seems to bring us back toward the device-plugin model, where the admin decides ahead of time how much capacity is exposed as “whole GPUs” versus “MIG/time-sliced/MPS-style” capacity. The DRA model should let the user request drive the shape of the allocation through the ResourceClaim, with Kueue accounting against the shared underlying device capacity.

keps/2941-DRA/README.md
```diff
@@ -873,6 +922,293 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+}
+```
+
+```yaml
+apiVersion: config.kueue.x-k8s.io/v1beta2
+kind: Configuration
+featureGates:
+  DynamicResourceAllocation: true
+  DRAPartitionableDevices: true
+resources:
```
alaypatel07, line 1139: Could we explore an alternative API shape where counters are attached to the `deviceClassMappings` entry instead of being exposed as a separate top-level quota resource?

For example:
```
resources:
  deviceClassMappings:
  - name: gpus
    deviceClassNames:
    - gpu.example.com
    - partitionable-device-class.example.com
    counters:
    - driverName: gpu.example.com
      name: memory
---
kind: ClusterQueue
metadata:
  name: gpu-queue
spec:
  resourceGroups:
  - coveredResources: ["gpus"]
    flavors:
    - name: whole-along-with-partitions-gpus
      resources:
      - name: gpus
         nominalQuota: "10"
```

With this shape, the cluster admin only sets quota on the parent resource, gpus. Kueue can then derive the available counter budget for each configured counter dimension from ResourceSlices, for example 10 * sharedCounters["memory"] for the selected flavor/pool.

A whole-device allocation would consume one unit of gpus and the full relevant counter capacity for the allocated device. A partitioned allocation would consume from the same gpus counter budget using the device’s consumesCounters values. This keeps whole-device and partitioned-device requests tied to the same underlying physical capacity, while still letting the user’s ResourceClaim determine the actual device shape.

alaypatel07 (changes requested):
Did a first pass of the proposed API, added some high level thoughts and an alternative.

keps/2941-DRA/README.md
sohankunkerkar: You're right, the separate struct creates two disconnected quota pools for the same hardware. I will rework it, so counters can be a field on `deviceClassMappings`. Both whole-GPU and MIG go in one entry, one pool.

keps/2941-DRA/README.md
sohankunkerkar: Yes, it makes sense to think from the admin mental model here, so this approach fits well. I kept your idea of putting counters inside `deviceClassMappings`, but used counter units directly as the quota value.                                                                                                                                                                          

The issue with a device-count model plus a derived counter budget is that Kueue would need to track two coupled dimensions (device count and counter budget), handle fractional devices for MIG workloads, and compute budgets dynamically from ResourceSlice data , none of which exist in Kueue today.  For example, say an admin sets nominalQuota: "10" (devices) and Kueue derives a counter budget from sharedCounters. A whole-GPU workload is easy as  it consumes 1 device. But a MIG workload consumes a counter value (the driver publishes 9856Mi for 1g.10gb). Converting that back to a device fraction requires knowing the GPU's total capacity, and that ratio changes across GPU models and counter dimensions. There's no clean single number, so Kueue would need to track both dimensions. Using counter units keeps things simple since they behave like normal resources such as CPU or memory. The admin UX tradeoff is fair, but admins already need to know GPU specs while setting up flavors, so this doesn't feel like a huge additional burden.

alaypatel07: > The issue with a device-count model plus a derived counter budget is that Kueue would need to track two coupled dimensions (device count and counter budget), handle fractional devices for MIG workloads, and compute budgets dynamically from ResourceSlice data , none of which exist in Kueue today. For example, say an admin sets nominalQuota: "10" (devices) and Kueue derives a counter budget from sharedCounters. A whole-GPU workload is easy as it consumes 1 device. But a MIG workload consumes a counter value (the driver publishes 9856Mi for 1g.10gb). 

I don’t think we need to convert counter usage back into fractional GPUs. That is not the model I am proposing.

The model I have in mind is that a physical GPU has two coupled accounting dimensions:

1. whole-device availability: can this physical GPU still satisfy a whole-GPU request?
2. partitionable counter usage: how much of each configured counter, such as memory, has been consumed?

Once Kueue admits a partitioned allocation on a physical GPU, that GPU should no longer be counted as available for whole-GPU admission. Separately, Kueue should track the configured counter vectors, such as memory, so additional partitioned allocations can still be admitted against the remaining counter capacity on that same physical GPU.


For example, with:

```yaml
resources:
  deviceClassMappings:
  - name: gpus
    deviceClassNames:
    - gpu.example.com
    - mig.nvidia.com
    counters:
    - driverName: gpu.nvidia.com
      counterName: memory
```

If an admin configures quota in whole-device units:

```yaml
kind: ClusterQueue
..
..
      resources:
      - name: gpus
        nominalQuota: "10"
```

then the counter dimension should become active only when Kueue admits a partitioned allocation. At that point Kueue can use ResourceSlice data for the actual selected/opened GPU:
- mark one physical GPU as unavailable for whole-GPU admission
- add that GPU’s actual sharedCounters.memory value to the gpus.counters.memory budget
- charge the claim’s actual consumesCounters.memory value against that budget

Conceptually, after admitting a 1g.10gb claim on a GPU whose ResourceSlice reports sharedCounters.memory = 188928Mi and whose selected device reports consumesCounters.memory = 9984Mi, the accounting could look like:

```
status:
  flavorsUsage:
  - name: gpu-pool
    resources:
    - name: gpus
      nominalQuota: "10"
      used: "1" # one physical GPU is no longer available as a whole GPU
    - name: gpus.counters.memory
      nominalQuota: "188928Mi" # actual opened GPU counter capacity
      used: "9984Mi"           # actual partition consumption
```

with the above, I dont think the `defaultConsumption` field is needed.

sohankunkerkar: Thanks for the detailed model, I think I understand it now. What you're describing is the precise physical model: when a MIG allocation lands on a GPU, that GPU’s specific counter capacity gets reduced, and future allocations must fit within the remaining capacity of that same physical GPU.

The challenge is that Kueue quota tracking today is purely aggregate `(Flavor, Resource) -> Quantity`. Everything in the quota path works on totals, not on which physical GPU contributed the capacity. Supporting per-GPU counter budgets would require a new dimension in the quota model.

What I'm proposing for Alpha is an aggregate approximation: admins set `nominalQuota` in counter units, and MIG workloads charge their `consumesCounters` values against that pool. This does not enforce per-GPU limits directly, but the scheduler already handles placement while Kueue tracks overall consumption using the existing borrowing/preemption model unchanged.

This is similar to how Kueue already treats non-DRA resources today: quota is an approximation of physical capacity, and `waitForPodsReady` is the safety net when admission and scheduling diverge. I think this aggregate model is reasonable for Alpha, while proper per-GPU tracking feels more aligned with the TAS + DRA integration work already planned for graduation.

alaypatel07: > Thanks for the detailed model, I think I understand it now. What you're describing is the precise physical model: when a MIG allocation lands on a GPU, that GPU’s specific counter capacity gets reduced, and future allocations must fit within the remaining capacity of that same physical GPU.

I am not asking Kueue to track the exact GPU that kube-scheduler will later allocate. I agree that exact per-GPU placement/reservation is a larger problem and is out of scope for this KEP.

What I am asking is whether Kueue can derive the whole-device counter charge from ResourceSlices instead of requiring the admin to hard-code it via `defaultConsumption`.

For example, if the admin configures parent device quota:

```yaml
resources:
  deviceClassMappings:
  - name: gpus
    deviceClassNames:
    - gpu.example.com
    - mig.nvidia.com
    counters:
    - driverName: gpu.nvidia.com
      counterName: memory
      selector:
        cel:
          expression: "device.attributes['gpu.nvidia.com/product'].string == 'A100'"
```

then Kueue could use ResourceSlice data matching the driver/counter/selector to derive the effective whole-device counter consumption.

sohankunkerkar: ah, I see what you are saying. That's a cleaner approach. Using a CEL selector on the counter config to derive the whole-device charge from `ResourceSlices` instead of hard-coding defaultConsumption. The selector solves the `which pool's sharedCounters?` question that made auto-derivation hard without it.

For Alpha, would it be acceptable to keep `defaultConsumption` as the simpler path and add the selector-based derivation as a follow-up? The CEL selector on the counter config adds another compilation + evaluation path, and I'd like to get the core counter-based quota working first. The selector approach is the right direction for removing the admin burden. I just want to stage the complexity

alaypatel07: > For Alpha, would it be acceptable to keep defaultConsumption as the simpler path and add the selector-based derivation as a follow-up? The CEL selector on the counter config adds another compilation + evaluation path, and I'd like to get the core counter-based quota working first. The selector approach is the right direction for removing the admin burden. I just want to stage the complexity

I would prefer not to make `defaultConsumption` the primary Alpha API if we agree selector-based derivation is the right direction.

A few reasons:

1. This is static Kueue configuration, so CEL compilation can happen at startup/config load time. The compilation cost should not be on the admission hot path.
2. Kueue already has CEL evaluation machinery from https://github.com/kubernetes-sigs/kueue/pull/9742, so this would reuse the same direction.
3. For Alpha, the selector can be intentionally constrained. For example, we could only allow matching ResourceSlice/device attributes, and even start with a simple equality match on one attribute. That is enough for the main use case: letting admins identify a homogeneous device shape, such as product name.

The key concept of the selector here is to give admins the ability to define devices with similar shape, other ideas that solve the same issue but avoid CEL compilation+evaluation could also be explored.

keps/2941-DRA/README.md
```diff
@@ -873,6 +923,327 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+presence of `counters` on the mapping entry selects the path. Within the counter path,
+the charge per device is determined by data: `consumesCounters` on the matched device
+selects between the actual counter value and `defaultConsumption`.
+
+- `counters` and `extendedResourceNames` are mutually exclusive on the same mapping
+  entry. This is validated at config load time.
+- Using `counters` and `extendedResourceName` (the DeviceClass API field) for the same
+  DeviceClass is not supported in Alpha. This combination produces ambiguous quota
+  semantics depending on whether the workload arrives via extended resources or
+  ResourceClaimTemplates. Runtime validation will be evaluated for Beta.
```
PannagaRao: Maybe we can consider supporting counters alongside extendedResourceName instead of treating them as mutually exclusive. When an extended resource request (e.g., nvidia.com/gpu: 1) resolves to a DeviceClass whose mapping has counters configured, the extended resources path can detect this and charge the derived counter value instead of device count. If the mapping has no counters, the extended resource is charged as device count — same as today.                                                                                                                                                                                  
   
This way both paths -  nvidia.com/gpu: 1 via extended resources and a MIG RCT for mig.nvidia.com would both charge against gpu.memory in counter units. WDYT?

keps/2941-DRA/README.md
sohankunkerkar: I would like to provide some findings here. I looked at the NVIDIA DRA driver code to understand the actual need. In partitionable mode (DynamicMIG), the driver publishes consumesCounters on both whole GPUs and MIG devices and whole GPUs consume the full counter set. So the primary path (consumesCounters → actual value) covers everything without needing deviceSelector.

That said, this is driver-specific behavior. The K8s API allows devices without consumesCounters in pools that have sharedCounters, so other drivers may not follow this pattern. Including deviceSelector as a driver-agnostic fallback for Alpha makes sense. I will update the KEP to frame it as a fallback for drivers that don't publish consumesCounters on non-partitioned devices, not as the primary path.

keps/2941-DRA/README.md
sohankunkerkar: ~Yup, agreed! will update the KEP to reflect that.~

Good point about unified charging. The use case you're describing-> whole GPU + MIG from the same counter quota pool is valid. The issue is that extended resources only carry a quantity (nvidia.com/gpu: 1) with no profile-level information. The K8s scheduler auto-creates a ResourceClaim with just DeviceClassName + count, no CEL selectors or capacity requests. So Kueue can't determine the accurate counter charge. It would always be the max across all devices matching the DeviceClass. 

For workloads that need both a whole GPU and a MIG slice from the same counter pool, both should use RCT with CEL selectors. The whole-GPU RCT matches devices with consumesCounters: {memory: 80Gi}, the MIG RCT matches with consumesCounters: {memory: 9856Mi}. Both charge against gpu.memory accurately. Extended resources are better suited for device-count quota where profile specificity isn't needed.

keps/2941-DRA/README.md
alaypatel07: > I would like to provide some findings here. I looked at the NVIDIA DRA driver code to understand the actual need. In partitionable mode (DynamicMIG), the driver publishes consumesCounters on both whole GPUs and MIG devices and whole GPUs consume the full counter set. So the primary path (consumesCounters → actual value) covers everything without needing deviceSelector.

I am not sure I understand, the driver providing consumesCounter for the whole devices is for the accounting of resources at the DRA level (between scheduler and driver).

This thread is about the right UX for admins to configure the quota for whole devices as well as partitionable devices.  IMO here are the usecases that should all work:
1. admin should be able to set up quota for whole devices (this is the alpha functionility).
1. admin should be able to set up quota for whole device and partitionable devices i.e. 
   1.  when users use partitionable devices to request quota from whole devices, it should work the same as users requesting whole devices. Kueue should proper accounting.
1. admin should be able to set up quota exclusively for partitionable devices

for requirement (2) either there are two alternatives:
1. the admin can define a rough estimate of how much memory each devices consumes, or
2. the admin provides the accelerator name, that kueue can look up the memory advertised by the DRA driver

I prefer option (2) but I will leave it up to other reviewers to share their thoughts. cc @kannon92 @tenzen-y @mimowo

keps/2941-DRA/README.md
sohankunkerkar: @alaypatel07 As discussed over the call, I went ahead and looked into the Consumable Capacity API (KEP-5075) to check how the current config shape holds up. The core difference between Partitionable Devices and Consumable Capacity is the charge source. PD reads from the device (`consumesCounters`) and CC reads from the ResourceClaimTemplate's device request (`capacity.requests`). But from Kueue's perspective, both produce the same output: a resource.Quantity that gets charged against the (Flavor, Resource) → int64 quota model. The config just needs to identify which device resource dimension maps to which Kueue quota resource, that works for both. The implementation would add a new code path that reads `capacity.requests` from the RCT instead of matching devices in ResourceSlices, but the quota tracking stays the same.

alaypatel07: Thats amazing, this is correct as per my understanding. Thanks @sohankunkerkar

alaypatel07: @sohankunkerkar for posterity can you add this as an Appendix?

sohankunkerkar: Yup, I initially added that under `Notes` but moved it under `Appendix` now.

keps/2941-DRA/README.md
```diff
@@ -503,7 +549,9 @@ rules:
 **Required Permissions:**
 - `resourceclaims`: Read access to validate ResourceClaim references (though not supported for quota)
 - `resourceclaimtemplates`: Read access to process ResourceClaimTemplates and extract device class information
-- `resourceslices`: Read access to list cluster devices for CEL selector validation
+- `resourceslices`: Read access to list cluster devices for CEL selector validation and `consumesCounters` reading.
+  Kueue only reads device attributes for CEL matching and counter values for quota. It does
+  not store or expose device-level details such as UUIDs or firmware versions.
```
alaypatel07: > It does not store or expose device-level details such as UUIDs or firmware versions.

nit: this part is obvious and can be removed

keps/2941-DRA/README.md
```diff
@@ -771,25 +821,31 @@ continues to use `deviceClassMappings`.
 
 #### Path Separation
 
-The two DRA paths have independent quota resolution:
+The two DRA paths resolve quota independently by default:
 1. **Extended resources** (`DRAExtendedResources` gate): auto-discovers DeviceClass via
-   field indexer, uses `extendedResourceName` as quota key. No `deviceClassMappings` needed.
+   field indexer. Uses `extendedResourceName` as the default quota key. If the resolved
+   DeviceClass is also present in `deviceClassMappings`, Kueue uses the mapped logical
+   name instead to unify quota with the ResourceClaimTemplate path. When the mapping has
```
alaypatel07: From what I remember, based on our last meeting, extendedResources and partitionable devices should not be supported. This is an open question that needs to be addressed later on. Given that can we make the language here reflect that if an extendedResourceName uses a counter, the workload will be rejected?

keps/2941-DRA/README.md
```diff
@@ -876,6 +932,367 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+
+The `DeviceClassMapping` struct is extended:
+
+```golang
+type DeviceClassMapping struct {
+    // ...existing fields (Name, DeviceClassNames)...
+
+    // Counters configures counter-based quota for partitionable devices.
+    // When present, quota is tracked in counter units instead of device count.
+    // When the extended resources path resolves a DeviceClass whose mapping
```
alaypatel07: same as above, drop extended resources with counters, not supported

keps/2941-DRA/README.md
```diff
@@ -876,6 +932,367 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+
+    // DeviceSelector identifies the device shape for deriving the whole-device
+    // counter charge from ResourceSlices. When a matched device has no
+    // consumesCounters (e.g., a whole GPU), Kueue uses the selector to find
+    // pools with matching devices and reads sharedCounters[counterName] as the
+    // charge. The selector is compiled at config load time using the upstream
+    // dracel compiler. When omitted, devices without consumesCounters are
+    // marked inadmissible.
+    // +optional
+    DeviceSelector *resourcev1.DeviceSelector `json:"deviceSelector,omitempty"`
```
alaypatel07: resourcev1.DeviceSelector is very broad, do we need all of its functionality?

Is there a way to limit the CEL expression such that it is computationally cheap?

sohankunkerkar:
@alaypatel07 @mimowo could you PTAL?

keps/2941-DRA/README.md
```diff
@@ -876,6 +932,367 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+    // gpu-1-counter-set), so counterName matches across all counter sets
+    // for a given driver without requiring one mapping per GPU.
+    // +required
+    // +kubebuilder:validation:Required
+    // +kubebuilder:validation:MaxLength=63
+    CounterName string `json:"counterName"`
+
+    // DeviceSelector identifies the device shape for deriving the whole-device
+    // counter charge from ResourceSlices. When a matched device has no
+    // consumesCounters (e.g., a whole GPU), Kueue uses the selector to find
```
alaypatel07: What happens if kueue finds multiple devices matching the selector? Which devices Kueue uses to charge the counter?

keps/2941-DRA/README.md
alaypatel07: should it rather be named `sharedCounterDeviceSelector`?

keps/2941-DRA/README.md
```diff
@@ -876,6 +932,367 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+3. Lists ResourceSlices for the configured driver, groups them by pool and checks completeness
+4. Compiles DeviceClass selectors (pre-filter) and RCT request selectors using the
+   `dracel` compiler, evaluates both against devices in complete pools to produce
+   matched devices
+5. For each matched device, resolves the counter charge (in order):
+   - Device has `consumesCounters` containing the configured `counterName`: uses the
+     actual value (e.g., 9856Mi for a 1g.10gb MIG profile)
+   - Device has `consumesCounters` but `counterName` is not found: workload is
+     marked inadmissible
+   - Device has no `consumesCounters` and `deviceSelector` is configured: Kueue
```
alaypatel07: This is not what I expressed in https://github.com/kubernetes-sigs/kueue/pull/10283#discussion_r3235129534

There are two distinct problems:
1. how does admin configure quota for partitionable device?
2. how does kueue charge the quota for a partitionable device?

The KEP currently reflects deviceSelector as mechanism for resolving 2nd option. In my comment here: https://github.com/kubernetes-sigs/kueue/pull/10283#discussion_r3235129534, I was suggesting using device selector for first problem.

keps/2941-DRA/README.md
```diff
@@ -876,6 +932,367 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+metadata:
+  name: gpu-queue
+spec:
+  resourceGroups:
+  - coveredResources: ["gpu.memory"]
+    flavors:
+    - name: a100-pool
+      resources:
+      - name: gpu.memory
+        nominalQuota: "800Gi"   # 10 GPUs x 80Gi, in counter units
```
alaypatel07: Why can we not have the following?
```yaml
resources:
  deviceClassMappings:
  - name: gpu.memory
    deviceClassNames: [gpu-whole, mig.example.com]
    counters:
    - driverName: gpu.example.com
      counterName: memory
      deviceSelector:
        cel:
          expression: "device.attributes['gpu.example.com'].productName == 'ExampleGPU'"
---
kind: ClusterQueue
metadata:
  name: gpu-queue
spec:
  resourceGroups:
  - coveredResources: ["gpu.memory"]
    flavors:
    - name: a100-pool
      resources:
      - name: gpu.memory
        nominalQuota: "10"   # 10 GPUs, each ExampleGPU has 80GB, so 800GB memory is available
```

alaypatel07 (changes requested):
Added comments inline PTAL

keps/2941-DRA/README.md
sohankunkerkar: Makes sense. I initially added counter-aware extended resources to address Pannaga's [point](https://github.com/kubernetes-sigs/kueue/pull/10283#discussion_r3230166089) about unified charging for workloads using both nvidia.com/gpu: 1 and MIG RCTs. But you're right Extended Resources don't carry profile-level specificity, so the charge would always be the max across all matched devices. That defeats the  purpose of counter-based quota. Will revert to rejecting workloads where extended resources resolve to a DeviceClass with counters configured.

keps/2941-DRA/README.md
sohankunkerkar: We don't need the full functionality,  a simple attribute match covers the use case. The DRA CEL environment is already constrained: it only accesses `device.driver`, `device.attributes`, and `device.capacity`. Kubernetes enforces a runtime cost limit (`CELSelectorExpressionMaxCost = 1000000`) on every evaluation, and device attributes are bounded at 32 per device. Using DeviceSelector avoids defining a custom type and reuses the syntax admins already know from DeviceClass selectors. The derivation result will be cached so the evaluation doesn't repeat per workload.

keps/2941-DRA/README.md
sohankunkerkar: I think the the selector is for quota derivation, not per-workload charging. If multiple pools match, Kueue uses the maximum `sharedCounters[counterName]` value to derive the per-device capacity. Within a `ResourceFlavor`, GPUs are expected to be homogeneous so all values should be the same. The max is a safe default for edge cases.  Per-workload charges always come from `consumesCounters` on the matched device.

keps/2941-DRA/README.md
```diff
@@ -876,6 +931,342 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+    - name: a100-pool
+      resources:
+      - name: gpu.memory
+        nominalQuota: "10"   # 10 GPUs; Kueue derives 10 * 80Gi = 800Gi
+```
+
+When `sharedCounterDeviceSelector` is configured, Kueue evaluates the selector against
+devices across pools for the configured driver, reads `sharedCounters[counterName]` from
+matching pools to determine the per-device counter capacity, and multiplies by
+`nominalQuota` to derive the effective counter budget. If the selector matches devices
```
mimowo: This seems very suspicious for Kueue resource model tbh, because it seems to completely ignore resource borroting and COhorts. In Kueue we have prod users operating on ClusterQueues:
- which have `nominalQuota: 0`, but the `nominalQuota` is defined in Cohort
- have `nominalQuota: 0` but the capacity is borrowed from other CLusterQueues

sohankunkerkar: @alaypatel07 I think @mimowo raised a valid point about Cohorts and borrowing. The derivation model (nominalQuota * perDeviceCapacity) doesn't work when nominalQuota: 0, which is a real production pattern with Cohort-based capacity.
 
I have a proposal for Alpha: keep `nominalQuota` in counter units for correctness with Cohorts/borrowing. The `sharedCounterDeviceSelector` can serve as a validation and discovery tool. Kueue evaluates it at config time and surfaces the per-device counter capacity in logs or status, helping the admin compute the right nominalQuota value. The admin sets nominalQuota: "800Gi" (10 * 80Gi) directly. No runtime derivation, Cohorts work unchanged.                                                                               

For Beta, we can explore a kueuectl helper or ClusterQueue status field that shows the equivalent device count (e.g., "800Gi ≈ 10 A100-80GB GPUs") based on the selector, maybe?

Thoughts?

alaypatel07: > @alaypatel07 I think @mimowo raised a valid point about Cohorts and borrowing. The derivation model (nominalQuota * perDeviceCapacity) doesn't work when nominalQuota: 0, which is a real production pattern with Cohort-based capacity.

I dont completely follow why this will be problematic. With partitionable devices, kueue allows a change in resource model which can allow users to configure along different dimensions of counters (memory, compute etc). If counters are configured, they should only be able to borrow capacity from other counters, it does not make sense to allow a counter to borrow memory (a dimension) from a pool which does not have dimensions.

 

> I have a proposal for Alpha: keep nominalQuota in counter units for correctness with Cohorts/borrowing. The sharedCounterDeviceSelector can serve as a validation and discovery tool. Kueue evaluates it at config time and surfaces the per-device counter capacity in logs or status, helping the admin compute the right nominalQuota value. The admin sets nominalQuota: "800Gi" (10 * 80Gi) directly. No runtime derivation, Cohorts work unchanged.

With this approach my concern is that each new device in the cluster will require admin to monitor and audit the quotas. I am not convinced pre-declaring static quota is a user experience kueue should be providing to admins, when DRA provides ability to partition the device on-the-fly.

sohankunkerkar: @alaypatel07 Thanks for the inputs. You're right on both points. The derivation math works with Cohorts. IIUC, @mimowo's concern is not just about implementation risk but about changing Kueue's resource model. Today the resource model is pure: nominalQuota is what the admin sets, stored directly, used directly, no transformations, no external dependencies. With derivation, the stored quota value is not what the admin configured but it's transformed based on external ResourceSlice data. The cache layer would depend on DRA runtime state to compute quota values, which is a fundamental shift from "quota is what you see" to "quota is derived from external data." That changes the model's contract, not just the code.

The question is whether this model change is the right direction, or whether counter-unit nominalQuota (admin computes once, no transformation) is more aligned with how Kueue's resource model works today.

alaypatel07: > Today the resource model is pure: nominalQuota is what the admin sets, stored directly, used directly, no transformations, no external dependencies. With derivation, the stored quota value is not what the admin configured but it's transformed based on external ResourceSlice data. The cache layer would depend on DRA runtime state to compute quota values, which is a fundamental shift from "quota is what you see" to "quota is derived from external data." That changes the model's contract, not just the code.

I agree with this. The indirection comes with a cost. However, the resource model "quota is what you see" is fundamentally different from what partitionable devices API allows. With partitionable devices API, users defer the decision of which device gets allocated all the way till scheduling time, and that is the core advantage of feature. 

So the challenge here is to find the middle ground where both the models can converge safely.

If you want a statically defined quota, the correct model is to:
1. use device plugins
2. configure the nodes with the right shapes of the device
3. configure the quota in terms of memory using the device class mapping

If folks feel that for alpha it is okay to not find the right middle ground and we should start somewhere, its fine to me, but this thought should be captured.

sohankunkerkar: >If you want a statically defined quota, the correct model is to:

>use device plugins
>configure the nodes with the right shapes of the device
>configure the quota in terms of memory using the device class mapping

I would like to clarify here: the per-workload charge is fully dynamic. When a workload arrives, Kueue reads `consumesCounters` from actual ResourceSlice data at admission time. A 1g.10gb is charged 9856Mi, a 7g.80gb is charged 80Gi and all from live driver data. The only static part is the total budget (nominalQuota: "800Gi"), which the admin sets once. This is different from the device plugin model where both the capacity and the per-workload charge are statically pre-configured.

That said, I agree the static budget is a limitation. When GPUs are added or removed, the admin needs to update nominalQuota. Happy to capture this as an Alpha limitation.

For Beta, one approach that avoids changing the resource model: the `sharedCounterDeviceSelector` derivation can happen in `createResourceQuotas` at parse time (the same boundary where CPU millicores conversion already happens via ResourceValue). The admin sets `nominalQuota: "10" (devices)`, Kueue multiplies by the per-device capacity from sharedCounters before storing. By the time values reach the cache layer, they're plain counter units. The cache layer, Cohorts, borrowing, hot path, all unchanged. The
multiplication runs once per CQ reconcile, not per workload admission. This keeps the resource model intact while giving the admin device-unit UX. 
@alaypatel07 @mimowo Thoughts?

alaypatel07: > I would like to clarify here: the per-workload charge is fully dynamic. When a workload arrives, Kueue reads consumesCounters from actual ResourceSlice data at admission time. A 1g.10gb is charged 9856Mi, a 7g.80gb is charged 80Gi and all from live driver data. The only static part is the total budget (nominalQuota: "800Gi"), which the admin sets once. This is different from the device plugin model where both the capacity and the per-workload charge are statically pre-configured.

 don’t think this is 100% accurate. At admission time, the workload may request any available partition, and its CEL selectors may match more than one device shape.

For example, suppose the matching devices have:
- dev-a: consumesCounters.memory = 10Gi
- dev-b: consumesCounters.memory = 20Gi
- dev-c: consumesCounters.memory = 40Gi

The NVIDIA MIG DeviceClass is an example of this pattern: it selects MIG-capable devices, but does not necessarily select one exact profile:
https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/3687fd54d337b35f37ade08389cabdc380b734a9/deployments/helm/dra-driver-nvidia-gpu/templates/deviceclass-mig.yaml#L10

In that case, Kueue does not know which device kube-scheduler will allocate. The conservative admission-time charge would be:
```
per-workload charge = max(consumesCounters[counterName]) across all matching devices
```

That avoids undercharging, but it means Kueue may overcharge quota.

This is why I think a selector in the Kueue counter configuration is useful. It lets the admin define a homogeneous accounting domain, for example “this quota is for A100 1g.10gb-shaped partitions,” where all matching devices are expected to have the same counter consumption. If the workload request is not compatible with that accounting domain, Kueue can reject it or fall back to conservative max charging.

If adding the selector is too disruptive for Alpha, then we can keep both sides static: the admin configures the total counter capacity and the default/maximum per-workload charge. But that takes me back to my original concern: both the capacity declaration and the quota accounting become admin-maintained approximations as expressed here: https://github.com/kubernetes-sigs/kueue/pull/10283/changes#r3243697874

alaypatel07: It sounds like we have three possible directions. Let me try to summarize them so we can decide explicitly.

### Option 1: Counter-unit quota only for Alpha

Admins configure quota directly in counter units:

```yaml
resources:
- name: gpu.memory
  nominalQuota: "800Gi"
```

Kueue charges workloads using consumesCounters from ResourceSlices. For broad selectors that match multiple device shapes, Kueue charges conservatively, for example max(consumesCounters.memory) across matching devices.

Pros:

- Fits Kueue’s current aggregate (Flavor, Resource) -> Quantity model.
- Cohorts, borrowing, lending, and preemption work unchanged.
- Smallest implementation surface for Alpha.

Cons:

- Admins must compute and maintain the aggregate counter budget manually.
- GPU capacity changes require quota updates.
- Broad selectors may lead to overcharging.
- Does not give admins a device-count UX for partitionable devices.

### Option 2: Counter-unit quota plus config selector for homogeneous accounting domains

```yaml
resources:
  deviceClassMappings:
  - name: gpu.memory
    deviceClassNames: [mig.nvidia.com]
    counters:
    - driverName: gpu.nvidia.com
      counterName: memory
      selector:
        cel:
          expression: "device.attributes['gpu.nvidia.com'].profile == '1g.10gb'"
---
resources:
- name: gpu.memory
  nominalQuota: "800GiB"
```

Pros:

- Still preserves Kueue’s current quota model.
- Lets admins define homogeneous device/counter domains.
- Avoids ambiguous charging when broad DeviceClasses match multiple profiles.
- Can reject incompatible workloads or fall back to max charging.

Cons: 
- Adds CEL config surface and validation/evaluation work.
- Workload selectors and config selectors need clear compatibility semantics.
- Admins still configure total counter quota manually.

### Option 3: Device-unit UX with derived effective counter quota
Admins configure quota in device units, and Kueue derives the effective counter quota from ResourceSlices using a selector:
```yaml
counters:
- driverName: gpu.nvidia.com
  counterName: memory
  sharedCounterDeviceSelector:
    cel:
      expression: "device.attributes['gpu.nvidia.com'].productName == 'A100'"
---
resources:
- name: gpu.memory
  nominalQuota: "10" # device units; Kueue derives effective counter quota
```
Pros:

- Best admin UX when admins think in “number of GPUs.”
- Avoids duplicating GPU memory sizes in Kueue config.
- Uses ResourceSlices as the source of truth for counter capacity.

Cons:
- Changes/extends the quota model because configured quota and effective quota have different units.
- Needs explicit semantics for ClusterQueue and Cohort resourceGroups.
- Must apply consistently to nominalQuota, borrowingLimit, and lendingLimit.
- Needs status/debuggability for configured vs effective quota.
- Needs recomputation when relevant ResourceSlices change.
- Probably too large for Alpha.


My preference order is: option 2, option 1, option 3.  But I am okay if option 1 is preferable as long as the pros and cons are agreed upon and documented.

sohankunkerkar: I think we should use Option 1 for alpha because it preserves Kueue’s resource model and gives us a correct, composable base API. Then we can add higher-level UX improvements later, either through tooling/status or a carefully designed selector-based enhancement, without committing alpha to quota derivation semantics that are still unsettled.

sohankunkerkar: To clarify my earlier comment, I meant Option 1's quota model (counter-unit nominalQuota) as the base, which Option 2 also uses and adds a selector for accurate per-workload charging on top of the same quota model. Since the selector lives in the DRA processing path and doesn't touch the quota model or cache layer, I'm open to including it in Alpha alongside counter-unit nominalQuota. This gives unified pool + accurate charging.

alaypatel07: > Since the selector lives in the DRA processing path and doesn't touch the quota model or cache layer, I'm open to including it in Alpha alongside counter-unit nominalQuota.

+1, This is the primary reason for my preference of option 2. Will let others chime in on the API, but lgtm from my side with this.

keps/2941-DRA/README.md
```diff
@@ -876,6 +928,346 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+
+    // Selector narrows which devices determine the per-workload counter
+    // charge. When configured, only devices matching this selector are
+    // considered for reading consumesCounters. If a workload's matched
+    // devices include devices outside the selector scope, the workload is
+    // rejected to prevent undercharging. Accurate charging depends on the
+    // selected devices having the same counter value. The selector is
+    // compiled at config load time using the upstream dracel compiler.
+    // The DRA CEL environment is constrained
+    // to device properties only (driver, attributes, capacity) with a
```
alaypatel07: For this field, Kueue only needs `attributes.profile`

Is there any value is keeping this field optional?

keps/2941-DRA/README.md
```diff
@@ -876,6 +928,346 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+    // runtime cost limit enforced by Kubernetes. When omitted, all matched
+    // devices are used and the maximum consumesCounters value is charged.
+    // +optional
+    Selector *resourcev1.DeviceSelector `json:"selector,omitempty"`
```
alaypatel07: s/selector/consumesCounterSelector/g

alaypatel07 (commented):
Added a couple of questions inline, but those are not blockers.

After the design reviews:
1. The API has been changed to allow admins to manage partitionable devices alongside whole devices.
2. The quota charging mechanism has been changed from being implementation logic to an API field configurable by admins allowing for narrower/stricter quota management.

Considering the above changes, LGTM. Thanks @sohankunkerkar !!

/lgtm

keps/2941-DRA/README.md
sohankunkerkar: Keeping it optional covers two cases: single-profile DeviceClasses where all matched devices already have the same counter value, and broad DeviceClasses where the admin accepts conservative max-charging. Making it required would add config overhead for the simple case with no accuracy benefit.

sohankunkerkar: Discussed this offline. I think it make sense to keep it required as Single profile DeviceClasses are essentially whole-devices.

keps/2941-DRA/README.md
```diff
@@ -876,6 +928,340 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+resources:
+  deviceClassMappings:
+  - name: gpu.memory
+    deviceClassNames: [mig.nvidia.com]
+    counters:
+    - driverName: gpu.nvidia.com
+      counterName: memory
+      consumesCounterSelector:
+        cel:
+          expression: "device.attributes['gpu.nvidia.com'].profile == '1g.10gb'"
```
mimowo: I’m having a bit of trouble following how this scales beyond a single profile, or beyond a single resource.

In the example, the Kueue resource name is generic (`gpu.memory`), but the `consumesCounterSelector` is profile-specific (`profile == '1g.10gb'`). Because of that, it is not clear to me whether this configuration is intended to support only one MIG profile, or whether `1g.10gb` was chosen only for brevity.

For this KEP, I think it is important to include an example that demonstrates the common multi-profile case: a single GPU memory quota pool shared by workloads requesting different profiles, for example `1g.10gb` and another larger profile (say `2g.10gb`). In particular, it would be useful to see the ConfigMap, ClusterQueue, and workloads together, and how the resulting usage is aggregated against the same `gpu.memory` quota.

If the examples are too large for Design Details, I’d be fine with moving the larger examples to the Appendix.

keps/2941-DRA/README.md
```diff
@@ -876,6 +928,340 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+    name: node1-gpu0
+    generation: 1
+    resourceSliceCount: 2
+  sharedCounters:
+  - name: gpu-0-counter-set
+    counters:
+      memory:
+        value: 80Gi
+      # Driver also publishes multiprocessors, memory-slice-0 through
+      # memory-slice-7 - used by kube-scheduler, not by Kueue.
```
mimowo, line 1014: The KEP mentions that drivers may publish additional counters such as `multiprocessors` and `memory-slice-*`, while the alpha validation section says that `counters` must have exactly one entry.

Could we clarify whether this is only an alpha scoping / implementation-complexity limitation, or whether there is a more fundamental design limitation that would require expanding the API?

If it is the latter, then I think we should reason a bit more about the future API shape now, so that the alpha API does not require a breaking change later.

keps/2941-DRA/README.md
sohankunkerkar: The `1g.10gb` example was picked for brevity. The design handles multiple profiles sharing one quota pool. I will update the KEP with a multi-profile example showing Configuration + ClusterQueue + two workloads (1g.10gb and 7g.80gb) aggregating against the same `gpu.memory` quota.

keps/2941-DRA/README.md
sohankunkerkar: This is an alpha scoping choice. The counters field is already a slice ([]CounterConfig), so no API change needed to support multiple entries later. The Beta graduation criteria covers this under `re-evaluate multi-counter tracking.` Should have called that out more clearly in the validation section, will make sure it reads better.

mimowo: > The counters field is already a slice ([]CounterConfig), so no API change needed to support multiple entries later.

I see, so the scoping is to de-risk the initial implementation, and make it simpler, the API-layer is ready, right?

keps/2941-DRA/README.md
mimowo: > The 1g.10gb example was picked for brevity. 

This is what I thought, but I couldn't wrap my head around how this can be extended to handle multiple profiles, seeing the examples will help me a lot. Also, the generic naming of `gpu.memory`, and scoped naming of profiles made it unclear to me.

> I will update the KEP with a multi-profile example showing Configuration + ClusterQueue + two workloads (1g.10gb and 7g.80gb) aggregating against the same gpu.memory quota.

Thank you 👍

keps/2941-DRA/README.md
```diff
@@ -876,6 +928,379 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+return matched device objects or do pool grouping:
+
+1. Kueue looks up the RCT's DeviceClass in `deviceClassMappings`
+2. If the mapping has `counters` configured, enters the counter-based path
+3. Lists ResourceSlices for the configured driver, groups them by pool and checks completeness
+4. Compiles DeviceClass selectors (pre-filter) and RCT request selectors using the
+   `dracel` compiler, evaluates both against devices in complete pools to produce
+   matched devices
+5. Checks that all matched devices fall within the `consumesCounterSelector` scope.
+   If any matched device is outside the scope, the workload is marked inadmissible.
```
PannagaRao: If consumesCounterSelector is set to a narrow expression like `device.attributes['gpu.nvidia.com'].profile == '1g.10gb'`, and a workload's RCT selector matches devices across multiple profiles (e.g., `device.capacity['gpu.nvidia.com'].memory.compareTo(quantity('30Gi')) <= 0`), the matched devices could include both 1g.10gb (9856Mi) and 3g.20gb (20Gi) devices. 

Step 5 would reject this workload because the 3g.20gb device is outside the consumesCounterSelector scope — but the workload can still be satisfied by the 1g.10gb device. 

Should step 5 be relaxed , or should the consumesCounterSelector be applied as a filter before the workload's RCT selectors rather than as a post-match validation?

keps/2941-DRA/README.md
sohankunkerkar: Yes, exactly. The Alpha just validates `len(counters) == 1`

keps/2941-DRA/README.md
mimowo: I checked the new examples and they answer my questions. Specifically the global configuration no longer hardcodes a specific profile. Re-reading the KEP with the clarfiied examples will be much simpler 👍

keps/2941-DRA/README.md
```diff
@@ -876,6 +928,375 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+}
+
+// CounterConfig maps a DRA driver counter to the parent DeviceClassMapping's
+// Kueue quota resource.
+type CounterConfig struct {
+    // DriverName is the DRA driver name used to filter relevant ResourceSlices.
+    // Must match the spec.driver field on ResourceSlice objects.
+    // +required
+    // +kubebuilder:validation:Required
+    // +kubebuilder:validation:MaxLength=253
```
```suggestion
```

We can not use marker validations in Config API.

keps/2941-DRA/README.md
```diff
@@ -876,6 +928,375 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+
+    // CounterName is the counter name within the device's consumesCounters
+    // entries to track for quota. Must match a counter name published by
+    // the driver in ResourceSlice devices' consumesCounters field.
+    // Counter set names are per-GPU identifiers (e.g., gpu-0-counter-set,
+    // gpu-1-counter-set), so counterName matches across all counter sets
+    // for a given driver without requiring one mapping per GPU.
+    // +required
+    // +kubebuilder:validation:Required
+    // +kubebuilder:validation:MaxLength=63
```
tenzen-y: ditto

keps/2941-DRA/README.md
```diff
@@ -876,6 +928,375 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+    // selector is compiled at config load time using the upstream dracel
+    // compiler.
+    // +required
+    ConsumesCounterSelector resourcev1.DeviceSelector `json:"consumesCounterSelector"`
+}
+```
+
+Multi-profile MIG setup sharing a single `gpu.memory` quota pool. The
+`consumesCounterSelector` scopes the accounting domain to devices from the configured
+driver. Per-workload charging comes from the workload's own RCT selector, which narrows
```
```suggestion
driver. Per-workload charging comes from the workload's own ResourceClaimTemplate selector, which narrows
```

Is this `ResourceClaimTemplate`? If yes, can you keep the fully name for readers and reviewers.

keps/2941-DRA/README.md
```diff
@@ -876,6 +928,375 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+    // gpu-1-counter-set), so counterName matches across all counter sets
+    // for a given driver without requiring one mapping per GPU.
+    // +required
+    // +kubebuilder:validation:Required
+    // +kubebuilder:validation:MaxLength=63
+    CounterName string `json:"counterName"`
+
+    // ConsumesCounterSelector scopes which devices are eligible for
+    // counter-based quota accounting. Typically matches a GPU model
+    // (e.g., productName) so all MIG profiles on that model share one
```
```suggestion
    // (e.g., productName) so all partition profiles on that model share one
```

Can you avoid using a vendor-specific name?

keps/2941-DRA/README.md
```diff
@@ -876,6 +928,375 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+    // quota pool. Per-workload charging is determined by the workload's
+    // own RCT selector, which narrows to the requested profile. The
```
```suggestion
    // own ResourceClaimTemplate selector, which narrows to the requested profile. The
```

keps/2941-DRA/README.md
```diff
@@ -876,6 +928,375 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+    ConsumesCounterSelector resourcev1.DeviceSelector `json:"consumesCounterSelector"`
+}
+```
+
+Multi-profile MIG setup sharing a single `gpu.memory` quota pool. The
+`consumesCounterSelector` scopes the accounting domain to devices from the configured
+driver. Per-workload charging comes from the workload's own RCT selector, which narrows
+to the requested profile:
+
+```yaml
```
tenzen-y, line 1132: I was not sure the reason why we should have both `driverName` and `deviceClassNames` in the API.
Isn't the only `deviceClassNames` enough?

alaypatel07: It is not enough. The driver name is needed for Kueue to find out which the right counter on the resource slice. The counter configuration is driver specific while device class is a higher level user abstract for admin

keps/2941-DRA/README.md
```diff
@@ -185,13 +204,22 @@ a simple device class named `gpu.example.com`. This will be the way to enforce q
...
+- Admins can enforce counter-based quota for partitionable devices (e.g., GPU memory quota
+  instead of device count quota for MIG profiles).
 
 ### Non-Goals
 
 - Quota-aware handling of DRAAdminAccess and DRAPrioritizedLists (beta, default enabled in K8s 1.35)
   is not included in Kueue's alpha. See [Risks and Mitigations](#risks-and-mitigations) for the
   planned approach.
-- Support for alpha DRA features like DRADeviceTaints and DRAPartitionableDevices will not be included.
+- Support for DRADeviceTaints will not be included in alpha.
```
mimowo: I think the message of the original point was "Support for alpha DRA features like DRADeviceTaints.".

sohankunkerkar: good catch, will fix this part

keps/2941-DRA/README.md
sohankunkerkar: Yeah, will address this

keps/2941-DRA/README.md
```diff
@@ -876,6 +928,375 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+name. The admin chooses which counters to track for quota.
+
+#### Processing Flow
+
+Partitionable devices reuses the upstream `dracel` compiler from the CEL expression
+support but performs its own
+ResourceSlice listing with pool-aware processing, since the CEL validation path does not
+return matched device objects or do pool grouping:
+
+1. Kueue looks up the RCT's DeviceClass in `deviceClassMappings`
```
mimowo: This question might be going beyond this KEP extension (still catching up with DRA), but how are ResourceFlavors taken into account in this flow?

IIUC this processing happens in the context of a currently considered ResourceFlavor. ResourceFlavors may restrict nodes via nodeLabels / nodeSelector-like constraints, and therefore also restrict the set of devices. For example, a ResourceFlavor may select only nodes in `topology.kubernetes.io/zone=europe-west-4`.

In the current processing flow, we seem to list/filter ResourceSlices without mentioning the ResourceFlavor. If so, we may account quota for devices that are attached to nodes from other zones, or otherwise outside of the currently considered flavor.

Maybe the ResourceFlavor constraint is assumed implicitly here, but it is not obvious to me. Should the flow explicitly say that the ResourceSlice/device matching is scoped to the current ResourceFlavor?

keps/2941-DRA/README.md
```diff
@@ -876,6 +928,375 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+
+Counter resources are injected through the existing `WithPreprocessedDRAResources` path.
+
+In Alpha, partitionable devices performs its own ResourceSlice listing independently from
+the CEL validation path because the two have different requirements: CEL validation only
+needs a match count, while counter processing needs matched device objects and pool-aware
+grouping. In Beta, these can be consolidated into a shared ResourceSlice listing layer.
+
+When no devices match or ResourceSlice data is unavailable, the workload is marked
+inadmissible. Workloads are never admitted with zero counter charge when `counters` is
```
mimowo: What if there are some devices or ResourceSlice matching , but not enough of them to admit the workload? I think in this case we should also mark inadmissible, because saying "no matching" is ad-hoc, if there is one matching but does not have enough capacity. wdyt?

mimowo:
After understanding how multiple profiles are handled, the KEP PR looks good overall. 

At this point, my biggest remaining question is how ResourceFlavors are taken into account in quota management: https://github.com/kubernetes-sigs/kueue/pull/10283#discussion_r3269316633

Given the complexity of the feature, I’m pretty sure we will need to iterate on this in following releases, but clarifying the meaning of ResourceFlavors seems critical before merging.

If this is already covered well, then I’m happy to LGTM, modulo some minor remarks.

keps/2941-DRA/README.md
sohankunkerkar: Good point. Counter processing runs in the workload controller during DRA preprocessing, before the workload reaches the FlavorAssigner. The charge (for example, `gpu.memory: 9856Mi`) is computed once globally and injected into the workload’s Requests as a regular resource. The FlavorAssigner then picks a flavor based on quota availability, just as it does for CPU or memory.

So ResourceSlice and device matching are not scoped by ResourceFlavor. The `consumesCounterSelector` scopes by GPU model (for example, `productName == 'A100'`), which keeps the charge correct because the same model produces the same counter values on any node. But if a flavor restricts scheduling to specific nodes and the relevant devices exist only elsewhere, Kueue may still admit the workload based on quota even though it cannot actually run on nodes compatible with that flavor. `WaitForPodsReady` handles this by evicting workloads that do not become ready within the configured timeout.

I’ll add a note to the KEP documenting this as an alpha limitation. Pool-aware flavor assignment is already listed in the Beta graduation criteria.

keps/2941-DRA/README.md
sohankunkerkar: Yup, I agree. When fewer devices match than the requested count, the workload should be marked inadmissible at admission time rather than deferring to scheduling. The existing CEL validation path already does this, so the counter path should be consistent. I'll update the KEP.

keps/2941-DRA/README.md
sohankunkerkar: Fixed

keps/2941-DRA/README.md
sohankunkerkar: Addressed.

keps/2941-DRA/README.md
tenzen-y: > The counter configuration is driver specific while device class is a higher level user abstract for admin

AFAIK, the `counter` is placed in KueueConfiguration which is for cluster admins.
Any reason to decouple driver admin (`driverName`) and cluster admin (`deviceClassName`)?

keps/2941-DRA/README.md
mimowo: I see, thank you for explaining. So, if there is a "a100" ResourceFlavor restricting nodes to `topology.kubernetes.io/zone=europe-west-4`, but there are no a100 machines in the admin needs to set the quota to 0. There is no automation like we have for TAS, at the same time this is the same basic quota model there is already for cpu / memory. sgtm.

keps/2941-DRA/README.md
mimowo: sgtm, yes, the "When no devices match or ResourceSlice data is unavailable" seems ad hoc, because no matter if zero devices, one or two matches the workload will remain inadmissible if it requires five devices.

The revised text lgtm

keps/2941-DRA/README.md
alaypatel07: The same driver name can have multiple device classes, the DRA API expects this. Because of this decoupling is preferred as it's hard to predict what device classes admins will come up with.

keps/2941-DRA/README.md
```diff
@@ -318,12 +359,16 @@ In order to mitigate this risk, Kueue can take the following approach:
 
 ## Design Details
```
mimowo, line 373: I think I know the answer but we may also clarify about preemption. In case of preemption the device mappings and counters are not considered, the workloads are preempted to make enough room for the "gpu.memory" resource. We know how much reasource is released for each workload based on the value recorded in the admission status. is that correct?

keps/2941-DRA/README.md
alaypatel07: Another concrete example is a device class can also cover multiple drivers names for example, give me a device that has 80GB of accelerator memory regardless of driver name, it could be a GPU or my own accenerator. Decoupling helps in making sure the API is forward looking for those usecases

keps/2941-DRA/README.md
```diff
@@ -272,20 +314,19 @@ administrator configures DeviceClasses with `extendedResourceName`.
...
-  the validation state diverges from actual device availability at scheduling time.
-
-- **Kueue does not have any informers on ResourceSlice changes.** CEL expression evalulation
-requires that the ResourceSlices are fetched from the apiserver. If a driver publishes new capacity and the workload was rejected, there is no
-resubmission of the workload once ResourceSlices changes. Evaluating this for beta is important.
+  validation can make previously-unsatisfiable workloads satisfiable. Kueue does not
+  currently have a ResourceSlice informer. Inadmissible workloads are only re-evaluated
+  when the ClusterQueue is notified through other events such as quota changes. Adding
+  event-driven requeuing on ResourceSlice changes is an Alpha graduation criterion.
+  `WaitForPodsReady` serves as the safety net for cases where the validation state
```
mimowo, line 328: A little off topic, but we also find waitForPodsReady find super useful as a safety net and recommend it to all our users.  We had multiple incidents where wait for pods ready could have mitigated the issue. 

It may require v1beta3 (I'm not sure totally, maybe with a long timeout like 1h it wouldn't really be breaking to anyone), but it is something I wanted to raise for a long time, because this should really be default. 

Let me open an issue: https://github.com/kubernetes-sigs/kueue/issues/11366

mimowo:
LGTM, thank you for working on the DRA integration, and this proposal in particular. Closing the feature gap is really exciting for Kueue. It would also be great (long term) to marry DRA with TAS.

I will wait with approval for additional comments from @tenzen-y

keps/2941-DRA/README.md
tenzen-y: In case of Kueue extended resource name supports, we just ignore the situation where multiple device classes point to the same extended resource name: https://github.com/kubernetes-sigs/kueue/pull/8734#discussion_r2926775361

I think this relationship between device class name and driver name is similar.
Any reason or real story to support the situation where a device class covers multiple driver names?

I understand that technically, DRA covers lots of matrices, but I don't think that we should consider all matrices (patterns) at this first step.

tenzen-y:
> LGTM, thank you for working on the DRA integration, and this proposal in particular. Closing the feature gap is really exciting for Kueue. It would also be great (long term) to marry DRA with TAS.
> 
> I will wait with approval for additional comments from @tenzen-y

Just for reference: https://github.com/kubernetes-sigs/kueue/issues/10548

keps/2941-DRA/README.md
mimowo: Iiuc Kueue may be installed on a cluster with already configured DRA which does not necessarily name driver name and device name the same.

Also decoupling is also more intuitive and readable to me, colliding device name and driver name as the same seems potentially confusing (abstracting out from real use cases).

sohankunkerkar: Just to add here: From an implementation perspective Kueue needs the driver name to filter ResourceSlices efficiently without an extra DeviceClass lookup per workload.

keps/2941-DRA/README.md
sohankunkerkar: Yes, exactly. Preemption works on the admitted `gpu.memory` value in the workload's admission status. No counter re-evaluation or device matching involved. It's the same resource arithmetic as CPU or memory.

keps/2941-DRA/README.md
```diff
@@ -881,6 +936,372 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+    // Extended resource requests that resolve to a DeviceClass with counters
+    // are marked inadmissible.
+    // Requires the KueueDRAIntegrationPartitionableDevices feature gate.
+    // +optional
+    Counters []CounterConfig `json:"counters,omitempty"`
+}
+
+// CounterConfig maps a DRA driver counter to the parent DeviceClassMapping's
+// Kueue quota resource.
+type CounterConfig struct {
```
```suggestion
    Counters []DeviceClassCounterConfig `json:"counters,omitempty"`
}

// DeviceClassCounterConfig maps a DRA driver counter to the parent DeviceClassMapping's
// Kueue quota resource.
type DeviceClassCounterConfig struct {
```

I would keep the space for `CounterConfig` for everything, which means that I don't want to reserve `CounterConfig` only for DeviceClass.

keps/2941-DRA/README.md
tenzen-y: > Iiuc Kueue may be installed on a cluster with already configured DRA which does not necessarily name driver name and device name the same.
> 
> Also decoupling is also more intuitive and readable to me, colliding device name and driver name as the same seems potentially confusing (abstracting out from real use cases).
> 

I see, thank you for describing that. In that case, I'm ok with keeping both driver name and device class name here.

keps/2941-DRA/README.md
```diff
@@ -914,6 +969,372 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+    // counter-based quota accounting. Typically matches a GPU model
+    // (e.g., productName) so all partition profiles on that model share one
+    // quota pool. Per-workload charging is determined by the workload's
+    // own ResourceClaimTemplate selector, which narrows to the requested profile. The
+    // selector is compiled at config load time using the upstream dracel
+    // compiler.
+    // +required
+    ConsumesCounterSelector resourcev1.DeviceSelector `json:"consumesCounterSelector"`
+}
+```
```
tenzen-y: I'd like to propose a different API view based on the following 3 reasons

1. The current API's `consumesCounterSelector` sounds confusing because `consumeCounter` is also used in ResourceSlice `.spec.devices[*].consumesCounters`.

But, in ResourceSlice `consumesCounters` indicates which ResourceSlices are consumed in this ResourceSlice. In other words, the consumers and producers are the same API resource (ResourceSlice). 

OTOH, in this Kueue Config `consumesCounters`, the producer is DRA driver (ResourceSlice) and the consumer is Kueue.

So, I think that we should avoid introducing the same `consume` model into Kueue Configurations.

2. The current API has duplicated namespaces like `counters.counterName` which means `counters` namespace has also nested `counter` semantics. It is duplicated and we should avoid that.

3. This reason is only for `driverName` field. In kubernetes, we use `Name` suffix for resourceName like `resourceFlavorName` and `cohortName`, but `driver` is not Kubernetes API Resource ,IIUC.
So, I'd propose using `driver` instead of `drivername`.

For clarification, I attached my proposed API in the thread.

tenzen-y:
```yaml
apiVersion: config.kueue.x-k8s.io/v1beta2
kind: Configuration
resources:
  deviceClassMappings:
  - name: "gpu.memory"
    deviceClassNames: [mig.nvidia.com]
    counters:
    - name: "memory"
      source:
        driver: "gpu.nvidia.com"
        deviceSelector:
          cel:
            expression: "device.attributes['gpu.nvidia.com'].productName == 'A100'"
      # poolSelector is not implemeted for now, but it is good future enhancement when we find real use case.
      # poolSelector:                                       # Just for future reference, I don't say that we should introduce `poolSelector` at this time.
      #   matchLabels: { tier: production }
```

tenzen-y (commented):
@sohankunkerkar Thank you for updating that. I left my final comment.

keps/2941-DRA/README.md
mimowo: +1, especially on the `consumesCounterSelector` rename to `deviceSelector`.

sohankunkerkar: Yeah, the source grouping also helps future ConsuambleCapacity extensibility.

keps/2941-DRA/README.md
```diff
@@ -914,6 +969,380 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+  deviceClassMappings:
+  - name: gpu.memory
+    deviceClassNames: [mig.nvidia.com]
+    counters:
+    - name: memory
+      source:
+        driver: gpu.nvidia.com
+        deviceSelector:
+          cel:
+            expression: "device.driver == 'gpu.nvidia.com'"
```
tenzen-y, line 1149: Based on additional offline discussion with @sohankunkerkar, I'd propose the following API pattern:

```suggestion
apiVersion: config.kueue.x-k8s.io/v1beta2
kind: Configuration
featureGates:
  KueueDRAIntegration: true
  KueueDRAIntegrationPartitionableDevices: true
resources:
  deviceClassMappings:
  - name: gpu.memory
    deviceClassNames: [mig.nvidia.com]
    sources:
    - counter:
         name: memory
         driver: gpu.nvidia.com
         deviceSelector:
           cel:
            expression: "device.driver == 'gpu.nvidia.com'"
    - capacity:
         name: memory
         driver: gpu.nvidia.com
         { ... }
```

DRA driver can support both counter and capacity pattern, but Kueue admins should select which pattern should be tied to the counter name.

In other words, If installing DRA driver support only counters, they specify counter. If installing the DRA driver supports only consumable capacities, they specify capacities in Kueue. If both are exposed under the DRA driver, they can select the preferred approach.

keps/2941-DRA/README.md
```diff
@@ -914,6 +969,382 @@ is treated as a normal (non-DRA) extended resource. A subsequent reconciliation
...
+// Exactly one of the source types must be set.
+type DeviceClassSourceConfig struct {
+    // Counter configures counter-based quota for partitionable devices.
+    // Maps a DRA driver counter to the parent DeviceClassMapping's Kueue quota resource.
+    // +optional
+    Counter *CounterSource `json:"counter,omitempty"`
+}
+
+// CounterSource identifies where to read counter data from and which counter to track.
+type CounterSource struct {
```
```suggestion
type DeciceClassCounterSource struct {
```

To avoid conflict within APIs package.

tenzen-y (commented):
I left a small comment.

keps/2941-DRA/README.md
sohankunkerkar: Fixed

tenzen-y (commented):
Thank you!
/lgtm
/approve

/hold cancel

PR #10294, opened 2026-04-05
cleanup: migrate to Status().Apply() with ApplyConfiguration for SSA

### LINKED ISSUE #8767: Use of Status.Patch(...patch. Apply) is deprecated in controller-runtime v0.23.0
**What happened**:

During the manual fix to [bump controller-runtime to v0.23.0](https://github.com/kubernetes-sigs/kueue/pull/8746) we stumbled upon a deprecated `Status.Patch(...patch. Apply)`.
New version requires to use new method `Status().Apply()`.
However the method accepts `WorkloadApplyConfiguration` instead of regular `Workload`.
[The structure](https://github.com/kubernetes-sigs/kueue/blob/v0.15.3/client-go/applyconfiguration/kueue/v1beta2/workloadstatus.go) is a mirror of regular `Workload` just of a different types.

IMHO there are 3 options:
1. Write a massive converter to go from `Workload` to this `WorkloadApplyConfiguration`.
2. Update each PatchStatus with alternative updateApply func.
3. Switch to Patch Merge as we plan in https://github.com/kubernetes-sigs/kueue/issues/7035.

Remove linter exception from golangci.yaml to see the problem.

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

falconlee236:
#### What type of PR is this?

/kind cleanup
/kind bug

This PR replaces the deprecated `client.Status().Patch(..., client.Apply)` method with the newly standard `client.Status().Apply(...)` method, using typed `ApplyConfiguration` to ensure compatibility with `controller-runtime` v0.23.0 updates.

Previously, the status patch logic relied on passing a full or dummy object to the generic patch method. This approach lacked explicit Field Management intent and has been deprecated. This PR introduces a robust Server-Side Apply (SSA) architecture by migrating to the typed builder pattern, which securely asserts Field Ownership and prevents potential overwrite conflicts with other controllers.

**Key changes include:**
1. **Workload Refactoring (`pkg/workload/workload.go`):** - Implemented `BaseWorkloadApplyConfiguration` to create a safe skeleton for SSA.
   - Introduced a converter (`ToWorkloadStatusApplyConfiguration`) to accurately map `WorkloadStatus` fields into the `WorkloadApplyConfiguration` builder.
2. **Pod Controller Refactoring (`pkg/controller/jobs/pod/pod_controller.go`):** - Replaced the dummy `pCopy` Pod object with the `corev1ac.Pod()` builder to clearly assert Field Ownership over the `TerminationTarget` condition upon workload eviction/stop.

Fixes #8767

@mimowo @tenzen-y 
Following the discussion in the issue, the "massive converter" approach was taken for the `Workload` status to properly construct the `WorkloadApplyConfiguration`. The logic for extracting and mapping the status fields has been isolated into helper functions to keep the `patchStatus` block clean. 

All existing tests pass locally. Please let me know if any field mappings need further adjustment!

/assign

```release-note
NONE
```

## Summary

- Replaced deprecated `Status().Patch(..., client.Apply)` calls with typed `Status().Apply()` calls.
- Added workload status apply conversion and preserved the applied status in the caller object.
- Updated Pod termination condition handling to use `PodApplyConfiguration`.
- Updated fake-client and integration tests for subresource apply behavior.

## Release note assessment

Ready. `NONE` accurately describes this compatibility change because it has no user-facing behavior change.


pkg/controller/jobs/pod/pod_controller.go
```diff
@@ -489,29 +481,21 @@ func (p *Pod) Stop(ctx context.Context, c client.Client, _ []podset.PodSetInfo,
...
-					},
-				},
-			},
-		}
-		//nolint:staticcheck //SA1019: client.Apply is deprecated
-		if err := c.Status().Patch(ctx, pCopy, client.Apply, client.FieldOwner(constants.KueueName)); err != nil && !apierrors.IsNotFound(err) {
+		podApplyConfig := corev1ac.Pod(podInGroup.pod.Name, podInGroup.pod.Namespace).
+			WithUID(podInGroup.pod.UID).
+			WithAPIVersion("v1").
+			WithKind("Pod").
```
mbobrovskyi: Do we need this?

pkg/controller/jobs/pod/pod_controller.go
```diff
@@ -489,29 +481,21 @@ func (p *Pod) Stop(ctx context.Context, c client.Client, _ []podset.PodSetInfo,
...
+			WithStatus(corev1ac.PodStatus().
+				WithConditions(corev1ac.PodCondition().
+					WithType(corev1.PodConditionType(ConditionTypeTerminationTarget)).
+					WithStatus(corev1.ConditionTrue).
+					WithLastTransitionTime(metav1.Time{Time: p.clock.Now()}).
```
mbobrovskyi:
```suggestion
					WithLastTransitionTime(metav1.NewTime(p.clock.Now())).
```

pkg/workload/workload.go
```diff
@@ -835,6 +818,118 @@ func BaseSSAWorkload(w *kueue.Workload, strict bool) *kueue.Workload {
...
+		applyConfig.WithAPIVersion(kueue.GroupVersion.String())
+	} else {
+		applyConfig.WithAPIVersion(w.APIVersion)
+	}
+
+	if w.Kind == "" {
+		applyConfig.WithKind("Workload")
+	} else {
+		applyConfig.WithKind(w.Kind)
+	}
```
mbobrovskyi: Do we need this?

pkg/controller/jobs/pod/pod_controller.go
falconlee236: what is difference NewTime with Time in metav1 ??

pkg/controller/jobs/pod/pod_controller.go
falconlee236: Calling `corev1ac.Pod` function is already included this line. i was fixed.

mimowo:
@falconlee236 what is the status here? Are there any blockers identified or we could move forward, let us know if you don't have cycles to work on it.

falconlee236:
@mimowo 
Thanks for reaching out! I’m still actively working on this.

I've identified the root cause, but my current fix is triggering failures in other unrelated tests. I’m spending time understanding the dependencies and the state machine to fix these side effects without breaking other features.

It’s taking a bit more time as I learn the internals, but I’m committed to resolving it. If there is an urgent timeline, please let me know and I will speed up the process as much as possible.

mimowo:
@falconlee236 thanks for reporting the progress and willingness to continue. I think it is ok if you continue on the task at the same time I would like to engage someone with already long expertise in Kueue so that maybe the failing tests can be understood easier. 

So, I would like to ask @mbobrovskyi to start looking into the PR already to see if this is going in the right direction or where possible issues are.

mimowo:
cc @mbobrovskyi would you find some cycles to guide the PR?

pkg/workload/admissionchecks.go
```diff
@@ -136,6 +134,7 @@ func SetAdmissionCheckState(checks *[]kueue.AdmissionCheckState, newCheck kueue.
 	existingCondition.Message = newCheck.Message
 	existingCondition.PodSetUpdates = newCheck.PodSetUpdates
 	existingCondition.RequeueAfterSeconds = newCheck.RequeueAfterSeconds
+	existingCondition.RetryCount = newCheck.RetryCount
```
mbobrovskyi: Why this changes?

pkg/controller/jobs/pod/pod_controller.go
```diff
@@ -482,41 +482,28 @@ func (p *Pod) Stop(ctx context.Context, c client.Client, _ []podset.PodSetInfo,
 
 	stoppedNow := make([]client.Object, 0)
 	for i := range podsInGroup {
-		// If the workload is being deleted, delete even finished Pods.
```
mbobrovskyi, line 511: Why removed comments?

pkg/controller/jobs/pod/pod_controller.go
```diff
@@ -470,7 +471,6 @@ func (p *Pod) GVK() schema.GroupVersionKind {
 func (p *Pod) PodLabelSelector() string {
 	return fmt.Sprintf("%s=%s", podconstants.GroupNameLabel, p.pod.Labels[podconstants.GroupNameLabel])
 }
-
```
mbobrovskyi, line 499: Please revert.

pkg/controller/jobs/pod/pod_controller.go
```diff
@@ -482,41 +482,28 @@ func (p *Pod) Stop(ctx context.Context, c client.Client, _ []podset.PodSetInfo,
...
+				WithConditions(corev1ac.PodCondition().
+					WithType(corev1.PodConditionType(ConditionTypeTerminationTarget)).
+					WithStatus(corev1.ConditionTrue).
+					WithLastTransitionTime(metav1.NewTime(p.clock.Now())).
+					WithReason(string(stopReason)).
+					WithMessage(eventMsg),
+				),
+			)
+
+		err := c.Status().Apply(ctx, podApplyConfig, client.FieldOwner(constants.KueueName), client.ForceOwnership)
```
mbobrovskyi, line 530: OK, I see these changes are more complex than they initially seemed and require updates to our unit tests as well, especially where we use interceptors.

https://github.com/kubernetes-sigs/kueue/blob/167001f17838d601a88309d4b55fd0858ac6da31/pkg/controller/jobs/pod/pod_controller_test.go#L6282-L6296

We need to use SubResourceApply here instead of SubResourcePatch.

I suggest moving these changes to a separate PR for easier review and fixes.

@falconlee236, could you please create a separate PR to address this?

pkg/controller/jobs/pod/pod_controller.go
```diff
@@ -482,41 +482,28 @@ func (p *Pod) Stop(ctx context.Context, c client.Client, _ []podset.PodSetInfo,
...
 		if client.IgnoreNotFound(err) != nil {
```
mbobrovskyi, line 531: @mimowo I took a deep dive into this case and found an interesting issue. Apply does not return a NotFound error – instead, it recreates the Pod if it doesn’t exist. I don’t think we can use it, as this behavior isn’t correct for our use case.

mimowo: Oh I see, but IIUC we used apply for long without any issues so far, so what has changed?

mimowo: in other words, do we already have a bug?

mbobrovskyi: It looks like the behavior of Apply() has changed. Previously, it returned a NotFound error if the Pod didn’t exist. We may need to migrate all cases to Merge Patch.

mimowo: Weird, do you mean the behavior changed for unit tests, or production behavior? Can we check how the Apply behaves on a real cluster? 

This remains unclear to me:
1. do we have currently a bug that on a real cluster this line could re-create Pod
2. does the PR introduce a bug that on a real cluster this line coule re-create the Pod
3. is it just about unit test behavior change?

mbobrovskyi: We are migrating from `client.Status().Patch(..., client.Apply)` (which is deprecated) to the newer standard `client.Status().Apply(...)`. With `client.Status().Patch(..., client.Apply)`, a NotFound error is returned if the Pod doesn’t exist, whereas `client.Status().Apply(...)` creates the Pod instead.

mbobrovskyi: Currently, we don’t have a bug because client.Status().Patch(..., client.Apply) returns a NotFound error. However, we can’t migrate to client.Status().Apply(...) without handling the NotFound case.

mimowo: > Currently, we don’t have a bug because client.Status().Patch(..., client.Apply) returns a NotFound error. However, we can’t migrate to client.Status().Apply(...) without handling the NotFound case.

Have you confirmed both behaviors- Pod recreation (after change) and non-recreation (before change)  on a test in a real cluster?

mbobrovskyi: Yeah, you’re right. I tested it on a real cluster, and it works – it returns a NotFound error. The question is, how do we test this in unit tests?

mbobrovskyi: I created experiment PR to test it https://github.com/kubernetes-sigs/kueue/pull/10686.

mimowo: Maybe we could registed a dedicated interceptor for unit tests which would mimic the real cluster behavior and respond with NotFound. 

Also do you suspect this is a bug in the library? In that case we shoukd open an issue to leaen from the maintainers, maybe this is overlooked, wdyt?

mimowo: basically when the fake client fails to mimic the real server it suggests to me a bug in the fake client, maybe easy to fix, but this requires a bit more investigation to properly locate where the bug is, and to open the issue

pkg/controller/jobs/pod/pod_controller.go
falconlee236: oh, i work this issue today, but you already merged this PR..
I'm so sorry about late addressing your request

https://github.com/kubernetes-sigs/kueue/pull/10749

falconlee236:
@mimowo, Thanks for your understanding and for asking @mbobrovskyi to join!

@mbobrovskyi, I just pushed my latest code. I've managed to reduce the number of failing tests significantly, and this current commit has the fewest failures so far. I would really appreciate it if you could start your review from this code. Your expertise will be super helpful for me to understand and fix the remaining side effects!

pkg/controller/jobs/pod/pod_controller.go
mimowo: No need to be sorry @falconlee236. AFAIK this PR only addresses a part of the issue, and the full migration still requires extra work. Let us know if you want to carry that work, and where you face the issues. 

I would love it if you continue working on the issue, at the same time I need to make sure we are not getting stuck.

mbobrovskyi:
To fix unit tests instead of 

```
WithInterceptorFuncs(interceptor.Funcs{SubResourcePatch: utiltesting.TreatSSAAsStrategicMerge})
```

use 

```
WithInterceptorFuncs(interceptor.Funcs{SubResourceApply: utiltesting.TreatSSAAsStrategicMergeForApplyConfiguration})
```

I added this interceptor in https://github.com/kubernetes-sigs/kueue/pull/10749. You can see how I fixed it for Pods.

mbobrovskyi:
To fix verify use `make lint-fix`.

falconlee236:
> To fix unit tests instead of
> 
> ```
> WithInterceptorFuncs(interceptor.Funcs{SubResourcePatch: utiltesting.TreatSSAAsStrategicMerge})
> ```
> 
> use
> 
> ```
> WithInterceptorFuncs(interceptor.Funcs{SubResourceApply: utiltesting.TreatSSAAsStrategicMergeForApplyConfiguration})
> ```
> 
> I added this interceptor in #10749. You can see how I fixed it for Pods.

Thank you for guide me. i try to fix SubResourcePatch to SubResourceApply some test codes step by step.
I read where to error occured in test failed log, i appreciated it.
please wait when i fix all unit test code. Thank you!

pkg/scheduler/scheduler_tas_test.go
```diff
@@ -6908,20 +6908,20 @@ func TestScheduleForTASWhenWorkloadModifiedConcurrently(t *testing.T) {
...
+							if obj.GetObjectKind().GroupVersionKind().Kind == "Workload" && subResourceName == "status" && !patched {
 								patched = true
 								// Simulate concurrent modification by another controller
 								wlCopy := tc.workload.DeepCopy()
 								if wlCopy.Labels == nil {
 									wlCopy.Labels = make(map[string]string, 1)
 								}
 								wlCopy.Labels["test.kueue.x-k8s.io/timestamp"] = time.Now().String()
-								if err := c.Update(ctx, wlCopy); err != nil {
+								if err := client.Update(ctx, wlCopy); err != nil {
```
mbobrovskyi:
```suggestion
								if err := c.Update(ctx, wlCopy); err != nil {
```
nit: could you please revert renaming?

falconlee236: OK I will write client vars into c after commits

mimowo:
@falconlee236 @mbobrovskyi what is the status here? Is there a chance to split the big PR into smaller pieces? We have the 0.18 release approaching EOW, and it would be great to use the new API. If we cannot then fine, but I'm trying to make sure we are trying to make progress.

falconlee236:
Hi @mimowo  @mbobrovskyi,

I thought it would be best to fix the unit tests first. By utilizing the test wrapping tools you provided, I've been actively modifying the code and managed to reduce the number of failing unit tests from 50 down to just 1.

Next up are the integration tests. From a quick look, it seems to be a metric-related issue, and I haven't quite figured out how to resolve it yet.

Regarding splitting the PR: the files that need to be changed for the SSA migration have very strong dependencies on each other, so I'm not entirely sure how to safely break this down into smaller pieces. I would really appreciate your thoughts or suggestions on this.

Realistically, since I am working on this alongside my full-time job, it will be difficult to finish everything by the end of this week to meet the 0.18 release deadline.

If this is extremely urgent for the release, you are more than welcome to take over and build upon my current code. However, I still feel a strong sense of responsibility for this issue and would love to see this PR through to completion myself if time allows.

Thanks!

mbobrovskyi:
It looks like a new unit test was added to main. Please rebase and apply the changes for:

https://github.com/kubernetes-sigs/kueue/blob/d18417383abac529413ead27d3b5bae594b21257/pkg/scheduler/scheduler_test.go#L7904-L7909

mbobrovskyi:
To fix `pull-kueue-verify-main` use `make lint-fix`.

pkg/workload/workload.go
```diff
@@ -1249,26 +1500,52 @@ func patchStatus(ctx context.Context, c client.Client, wl *kueue.Workload, owner
...
+
+		if err := c.Status().Apply(ctx, applyConfig, owner, client.ForceOwnership); err != nil {
+			return err
+		}
+		if err := c.Get(ctx, client.ObjectKeyFromObject(wl), wl); err != nil {
+			return err
 		}
-		return update(wl)
+		// wl.Status = wlCopy.Status
+		return nil
```
mbobrovskyi:
```suggestion

		return c.Get(ctx, client.ObjectKeyFromObject(wl), wl)
```

pkg/workload/workload.go
```diff
@@ -1249,26 +1500,52 @@ func patchStatus(ctx context.Context, c client.Client, wl *kueue.Workload, owner
...
+
+		if err := c.Status().Apply(ctx, applyConfig, client.FieldOwner(constants.AdmissionName), client.ForceOwnership); err != nil {
+			return err
 		}
-		return true, nil
+		if err := c.Get(ctx, client.ObjectKeyFromObject(wl), wl); err != nil {
+			return err
+		}
+		// wl.Status = wlCopy.Status
+		return nil
```
mbobrovskyi:
```suggestion
		return c.Get(ctx, client.ObjectKeyFromObject(wl), wl)
```

pkg/workload/workload.go
```diff
@@ -1304,37 +1581,30 @@ func HasQuotaReservation(w *kueue.Workload) bool {
...
+		applyConfig := BaseWorkloadApplyConfiguration(wlPatch, opts.StrictApply)
+		applyConfig.WithStatus(ToWorkloadStatusApplyConfiguration(&wlPatch.Status))
+		if err := c.Status().Apply(ctx, applyConfig, client.FieldOwner(constants.ReclaimablePodsMgr), client.ForceOwnership); err != nil {
+			return err
+		}
+		if err := c.Get(ctx, client.ObjectKeyFromObject(wl), wl); err != nil {
+			return err
+		}
+		// wl.Status.ReclaimablePods = reclaimablePods
+		return nil
```
mbobrovskyi:
```suggestion
		return c.Get(ctx, client.ObjectKeyFromObject(wl), wl)
```

mbobrovskyi: Actually, PatchStatus() should already handle it. Why are we handling it separately?

mbobrovskyi:

Due to https://github.com/kubernetes-sigs/kueue/issues/11631

mbobrovskyi:
@falconlee236 could you please squash commits?

falconlee236:
> @falconlee236 could you please squash commits?

your mean is "Squash the last commits into one" and force pushed?

mbobrovskyi:
> your mean is "Squash the last commits into one" and force pushed?

Right.

pkg/workload/workload.go
```diff
@@ -1210,6 +1213,317 @@ func WithForceApply() PatchStatusOption {
...
+		applyConfig.WithLabels(w.Labels)
+	}
+	if len(w.Annotations) > 0 {
+		applyConfig.WithAnnotations(w.Annotations)
+	}
+
+	return applyConfig
+}
+
+func ToWorkloadStatusApplyConfiguration(status *kueue.WorkloadStatus) *kueueac.WorkloadStatusApplyConfiguration {
```
mbobrovskyi: It looks like this function is very large. I'm wondering if we can decompose some parts into separate functions.

pkg/workload/workload.go
```diff
@@ -1252,26 +1566,155 @@ func patchStatus(ctx context.Context, c client.Client, wl *kueue.Workload, owner
...
 		}
-		return update(wl)
+		// Refresh wl so the caller sees the post-apply resourceVersion (required for
+		// follow-up writes like RemoveFinalizer to not fail with "object was
+		// modified"). Then overlay the just-applied status changes in case the
+		// cached read has not yet observed the server-side merge.
+		if err := c.Get(ctx, client.ObjectKeyFromObject(wl), wl); err != nil {
+			return err
+		}
+		mergeAppliedStatus(&wl.Status, &wlCopy.Status)
```
mbobrovskyi: Why we merging after get?

pkg/workload/workload.go
```diff
@@ -1252,26 +1566,155 @@ func patchStatus(ctx context.Context, c client.Client, wl *kueue.Workload, owner
...
+
+		if err := c.Status().Apply(ctx, applyConfig, client.FieldOwner(constants.AdmissionName), client.ForceOwnership); err != nil {
+			return err
 		}
-		return true, nil
+		// Reflect the just-applied changes locally (see PatchStatus for rationale).
+		if err := c.Get(ctx, client.ObjectKeyFromObject(wl), wl); err != nil {
+			return err
+		}
+		mergeAppliedStatus(&wl.Status, &wlCopy.Status)
```
mbobrovskyi: I don't think this is correct. We shouldn't merge the old status; we need to fetch the new one.

pkg/workload/workload.go
falconlee236: ah yeah it became too long. let me split it per field in next commit

pkg/workload/workload.go
falconlee236: the new c.Status().Apply doesn't write server response back into input like the old c.Status().Patch(client.Apply) did. and c.Get goes through cache which can be stale right after Apply.

without this merge, provisioning controller in same reconcile reads ac.State as Pending even though we just set it to Retry, and creates extra ProvisioningRequest. that was the cause of the 5 retry tests failure.

is my understanding correct ?? I will add comment to explain

pkg/workload/workload.go
falconlee236: sorry, small thing - wlCopy.Status is not the old one. BaseSSAWorkload() makes it empty first, and update() only fills the fields we want to apply. so it has only the new values that we just sent.

but I think I understand your concern. is it better to use mgr.GetAPIReader() instead? it would bypass cache and we don't need the merge. only thing is I need to add APIReader parameter to PatchStatus / PatchAdmissionStatus / UpdateReclaimablePods.

which way do you prefer ??

pkg/workload/workload.go
mbobrovskyi: > the new c.Status().Apply doesn't write server response back into input like the old c.Status().Patch(client.Apply) did. and c.Get goes through cache which can be stale right after Apply.

Yes, exactly. That's why you are getting object from server, right? So why you are setting it manually? It was not updated?

falconlee236: Two clarifications, then a question:

1) wlCopy.Status is not the old status. BaseSSAWorkload() strips it to an empty skeleton first, and update() only fills the fields we intend to apply. So mergeAppliedStatus overlays only what we just sent — it never restores stale values, and fields we didn't touch (owned by other controllers) are left exactly as the cached Get returned them.
2) Why the overlay exists at all: c.Status().Apply doesn't write the server response back the way the old Status().Patch(client.Apply) did, and c.Get reads from the informer cache, which can lag right after Apply. In the provisioning controller this caused a same-reconcile stale read (ac.State was Retry on the server but read as Pending), which created a premature ProvisioningRequest — that was the root cause of the 5 retry test failures.

That said, I agree the hand-rolled merge is a maintenance burden (it has to know every status field). The clean alternative is to bypass the cache with mgr.GetAPIReader():
```go
 c.Status().Apply(...)
return apiReader.Get(ctx, key, wl)   // linearizable read, no merge needed
```

Cost: adding an apiReader client.Reader param to PatchStatus / PatchAdmissionStatus / UpdateReclaimablePods and threading it through ~57 call sites (we don't use APIReader anywhere yet), plus one apiserver round-trip per patch.

Which do you prefer — keep cached Get + overlay, or move to APIReader and drop mergeAppliedStatus entirely? Happy to do the APIReader refactor if that's the direction.

mimowo:
/retest
retest to see if the PR is failing due to recent makefile changes or there are issues.

mbobrovskyi:
@falconlee236 could you please resolve conflicts?

falconlee236:
@mbobrovskyi friendly ping on the open question here 🙂

**1. Direction decision (waiting on your call):** for the post-`Apply` refresh, do you prefer keeping the cached `Get` + `mergeAppliedStatus` overlay, or switching to an uncached read via `mgr.GetAPIReader()` and dropping `mergeAppliedStatus` entirely? CodeRabbit raised the same point. The APIReader route removes the hand-rolled merge but threads an `apiReader client.Reader` through `PatchStatus`/`PatchAdmissionStatus`/`UpdateReclaimablePods` (~57 call sites) plus one apiserver round-trip per patch. Happy to do the APIReader refactor if that's the direction — just want your preference before I commit.

**2. Rebase heads-up:** #12577 moved patching into `pkg/workload/patching/patching.go`, where `patchStatus` still uses the deprecated `c.Status().Patch(…, client.Apply, …)`. I'll re-port this SSA migration onto that location; since the refresh approach in (1) changes how I write it there, I'd like to settle (1) first.

Thanks!

mbobrovskyi:
> 1. Direction decision (waiting on your call): for the post-Apply refresh, do you prefer keeping the cached Get + mergeAppliedStatus overlay, or switching to an uncached read via mgr.GetAPIReader() and dropping mergeAppliedStatus entirely? CodeRabbit raised the same point. The APIReader route removes the hand-rolled merge but threads an apiReader client.Reader through PatchStatus/PatchAdmissionStatus/UpdateReclaimablePods (~57 call sites) plus one apiserver round-trip per patch. Happy to do the APIReader refactor if that's the direction — just want your preference before I commit.

Yeah, I think we should switch to mgr.GetAPIReader() and drop mergeAppliedStatus entirely to fetch a fresh object.

> 2. Rebase heads-up: https://github.com/kubernetes-sigs/kueue/pull/12577 moved patching into pkg/workload/patching/patching.go, where patchStatus still uses the deprecated c.Status().Patch(…, client.Apply, …). I'll re-port this SSA migration onto that location; since the refresh approach in (1) changes how I write it there, I'd like to settle (1) first.

Yes, please move this changes to `patching.go`.

falconlee236:
Hi @mbobrovskyi, I need your advice on something.

Like you said, I changed `patchStatus` to use `c.Status().Apply(...)` and added `apiReader.Get(...)` after apply to get the fresh object, and removed `mergeAppliedStatus`.

Most tests pass, but 2 concurrent admission tests fail:
- `concurrent_admission_test.go:187` (migrate to preferred flavor when it becomes available)
- `concurrent_admission_test.go:311` (only migrate to flavors at or above min)

The problem is when a better flavor becomes free, the workload is evicted but it never gets admitted again. `IsAdmitted` stays false until timeout. The controller keeps evicting again and again (`EvictedDueToFlavorMigration`), and the target variant stays `Pending the migration of 1 workload(s)`.

I checked `main` and it passes these 2 tests (main still uses old `c.Status().Patch(client.Apply)`). so my `Apply` + `apiReader.Get` change is the cause.

I also tried `apiReader.Get` + copy the applied status back to wl, but it still fails the same way. so i think it is not about what the caller sees, the server state just doesn't converge.

I couldn't find the exact reason yet. for concurrent admission, do you still want pure `apiReader.Get` without merge, or should this path keep some merge?? if you have any idea i can dig more. thanks!

mimowo:
@falconlee236 @mszadkow @mbobrovskyi @ivnovakov could you please brainstorm how we could move forward with this? Maybe instead of one big bang PR we could divide this into smaller delivarable and reviewable pieces?

mimowo:
I think we should resume the work on the PR. Most of the complications it seems to me was due to introducing the Get after apply to get the new object.

However, after https://github.com/kubernetes-sigs/controller-runtime/pull/3475 controller runtime started to return that in the response 🥇 That PR is already available since v0.24.0 of controller runtime, and I think it wasn't available prior to that. So, I think the task is much simpler now. 

@falconlee236 let me know if you want to continue

falconlee236:
Hi @mimowo,

It sounds good to getting simple this migration works.

I want to continue this work, 
Is it ok for proceeding this PR?

mimowo:
Yes, please continue, but I think you may rather take a backup branch and re-start with clean history, because most of the code changes will now need to be reverted.

falconlee236:
Are you saying close this PR and open new PR about this migration?

mimowo:
> Are you saying close this PR and open new PR about this migration?

Up to you. I meant that once you have something working feel free to force push overriding the history, because we will no longer need it.

mimowo:
@falconlee236 please sign CLA

Now the PR looks much better, thank you 👍 

cc @mbobrovskyi could you please give it a pass?

cmd/importer/pod/import_test.go
```diff
@@ -410,7 +410,10 @@ func TestImportNamespace(t *testing.T) {
 			pcList := schedulingv1.PriorityClassList{Items: tc.priorityClasses}
 
 			builder := utiltesting.NewClientBuilder().
-				WithInterceptorFuncs(interceptor.Funcs{SubResourcePatch: utiltesting.TreatSSAAsStrategicMerge}).WithStatusSubresource(&kueue.Workload{}).
+				WithInterceptorFuncs(interceptor.Funcs{
+					SubResourcePatch: utiltesting.TreatSSAAsStrategicMerge,
```
mbobrovskyi: Do we need this?

mbobrovskyi: Also in other places.

mbobrovskyi:
@falconlee236 please sign CLA

cmd/importer/pod/import_test.go
falconlee236: Is `SubResourcePatch` not nessesary  because We remove all deprecated Client side apply?

mbobrovskyi: We don't use `SubResourcePatch` anymore, right?

mbobrovskyi: I think we can drop `utiltesting.TreatSSAAsStrategicMerge` at all. It should work with new `utiltesting.TreatSSAAsStrategicMergeForApplyConfiguration`.

mimowo:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10294/pull-kueue-test-unit-shard-0-main/2097863798609154048

Let me check if the issue continues
/test pull-kueue-test-unit-shard-0-main

cmd/importer/pod/import_test.go
falconlee236: I was drop `TeatSSAAsaStrategicMerge` in kueue codebase. Is this change applying your request?

mbobrovskyi: Yes, that's exactly what I mean. Thank you!

mbobrovskyi:
@falconlee236 could you please rebase?

mbobrovskyi:
Thank you!

/lgtm

mimowo:
LGTM overall, but we still need a fix for https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10294/pull-kueue-test-unit-shard-0-main/2097989933653299200 

and rebase. Let's move it to the finish line

falconlee236:
@mimowo @mbobrovskyi 
Applyed your comments, Please check my commits. 
Let's go to this PR to finish line!!!!

And Please recommend another PR or issue that i can solve? i want to consist contribute this Project.

mimowo:
/approve
Thank you 👍 It is always refreshing when we stop using deprecated functions !

mimowo (commented):
/lgtm

mbobrovskyi:
Awesome! Thank you!

/lgtm

Issue #8586, opened 2026-01-14
Logging: some logging does not conform to JSON lines format


mimowo:
**What happened**:

Some logging does not conform to JSON format. 

**What you expected to happen**:

I would like all the logging, at least down to V6, to be consistent with JSON formatting.

This is so that we can use standard JSON tooling for parsing. 

**How to reproduce it (as minimally and precisely as possible)**:

What I did I looked at the logs from integration tests at v3, eg: https://storage.googleapis.com/kubernetes-ci-logs/logs/periodic-kueue-test-integration-main/2011334372992487424/build-log.txt

Then I found with some scripting problematic lines, examples:

```
2026-01-14T07:18:19.675323116Z	DEBUG	events	recorder/recorder.go:104	Quota reserved in ClusterQueue prod-cq, wait time since queued was 1s; Flavors considered: main: spot-tainted(NoFit;untolerated taint {cloud.provider.com/instance spot-tainted NoSchedule <nil>} in flavor spot-tainted)	{"type": "Normal", "object": {"kind":"Workload","namespace":"core-tl7wn","name":"job-prod-job1-e9458","uid":"61f0e5dc-2adc-43a4-a37b-ab4eaad0d455","apiVersion":"kueue.x-k8s.io/v1beta2","resourceVersion":"831"}, "reason": "QuotaReserved"}
```
However, this is just for v3, we should repeat this experiment for v6.

**Anything else we need to know?**:

This is the python script I used to process the log:

```python
import sys
import re

def process_logs():
    # ^\s* allows for any amount of leading whitespace
    # \d{4}-\d{2}-\d{2} matches the YYYY-MM-DD date format
    date_pattern = re.compile(r"^\s*\d{4}-\d{2}-\d{2}")
    for line in sys.stdin:
        clean_line = line.rstrip()      
        if not clean_line:
            continue
        if not date_pattern.match(clean_line):
            continue
        index = clean_line.find('{')
        
        if index != -1:
            json_payload = clean_line[index:]
            print(json_payload)

if __name__ == "__main__":
    try:
        process_logs()
    except (KeyboardInterrupt, BrokenPipeError):
        sys.stderr.close()
        sys.exit(0)
```
Then I used https://jsonlines.org/validator/ to find examples of invalid JSON lines.

mimowo:
cc @mwielgus @PBundyra

Piyushkhobragade:
Hi @mimowo,

Thanks for the detailed report and examples.

[...]

PR #10295, opened 2026-04-05
fix(scheduler): use context-aware structured logging for cohort-less ClusterQueue path

Piyushkhobragade:
Switch to context-aware structured logging for the cohort-less ClusterQueue path.

This preserves controller-runtime logging context while ensuring JSON-lines compliant output.

Part of #8586.

Scope is intentionally minimal.


mbobrovskyi (commented):
I didn’t quite get the difference. Could you please show a before-and-after example?

mimowo:
/close
Seems like we don't need it. Please re-open if you see some use-case

PR #10320, opened 2026-04-08
Implement support for partitionable devices

PannagaRao:
#### What type of PR is this?
/kind feature

Fixes # https://github.com/kubernetes-sigs/kueue/issues/10108

```release-note
DRA: Add counter-based quota support for partitionable DRA devices (e.g., NVIDIA MIG) via counterMappings configuration.
```


kannon92:
/assign

@PannagaRao has mentioned to me that this is ready for review so going to put this on my plate this week.

apis/config/v1beta2/configuration_types.go
```diff
@@ -569,6 +574,37 @@ type DeviceClassMapping struct {
...
+	// Name is the Kueue resource name used in ClusterQueue coveredResources
+	// and Workload resourceUsage.
+	// Must be a valid fully qualified name consisting of an optional DNS subdomain prefix
+	// followed by a slash and a DNS label, or just a DNS label.
+	// DNS labels consist of lower-case alphanumeric characters or hyphens,
+	// and must start and end with an alphanumeric character.
+	// DNS subdomain prefixes follow the same rules as DNS labels but can contain periods.
+	// The total length must not exceed 253 characters.
+	// +required
+	// +kubebuilder:validation:MaxLength=253
```
kannon92: I don't think these kubebuilder validations do anything for config apis. These are not CRDs so I don't think there is any enforcement here.

cmd/experimental/kueue-populator/go.mod
```diff
@@ -18,7 +18,9 @@ require (
 )
 
 require (
```
kannon92: I don't really get why you changed this file.

cmd/kueue/main.go
```diff
@@ -303,8 +303,9 @@ func main() {
 		cacheOptions = append(cacheOptions, schdcache.WithResourceTransformations(cfg.Resources.Transformations))
 		queueOptions = append(queueOptions, qcache.WithResourceTransformations(cfg.Resources.Transformations))
 	}
-	if features.Enabled(features.DynamicResourceAllocation) && cfg.Resources != nil && len(cfg.Resources.DeviceClassMappings) > 0 {
-		if err := dra.CreateMapperFromConfiguration(cfg.Resources.DeviceClassMappings); err != nil {
+	if features.Enabled(features.DynamicResourceAllocation) && cfg.Resources != nil &&
```
kannon92, line 307: This feature gate is not complete.

I think you also need PartitionableDevices in addition to DRA.

pkg/dra/claims.go
```diff
@@ -182,3 +215,209 @@ func GetResourceRequestsForResourceClaimTemplates(
...
+				count:           req.Exactly.Count,
+				deviceClassName: req.Exactly.DeviceClassName,
+				selectors:       req.Exactly.Selectors,
+			})
+		}
+	}
+	return result
+}
+
+// validateCELSelectorsAgainstDevices lists all ResourceSlices in the cluster and
```
kannon92, line 287: This reads like my work in https://github.com/kubernetes-sigs/kueue/pull/9742.

Are you waiting on this PR to merge? Or do you want to close mine?

kannon92 (changes requested):
There are a few items that need to be addressed (feature gates especially).

I am also curious on the dependency on CEL https://github.com/kubernetes-sigs/kueue/pull/9742 and do we have a dependency we should call out or are you aiming to merge this?

cmd/experimental/kueue-populator/go.mod
PannagaRao: The populator verify CI (pull-kueue-populator-verify-main) failed without it as the populator go.mod/go.sum were out of sync.

kannon92: cc @mbobrovskyi 

Is this expected?

pkg/dra/claims.go
PannagaRao: This PR has a dependency on https://github.com/kubernetes-sigs/kueue/pull/9742. So I'm waiting for it to merge.

PannagaRao: Added it in the PR description.

apis/config/v1beta2/configuration_types.go
PannagaRao: Got it. I'll remove them

cmd/kueue/main.go
PannagaRao: Good catch. Gated counterMappings behind DRAPartitionableDevices — when the gate is off, nil is passed to CreateMapperFromConfiguration which already handles nil.

apis/config/v1beta2/configuration_types.go
```diff
@@ -520,6 +520,11 @@ type Resources struct {
 	// for Dynamic Resource Allocation support.
 	// +optional
 	DeviceClassMappings []DeviceClassMapping `json:"deviceClassMappings,omitempty"`
+
+	// CounterMappings defines mappings from DRA driver counters to logical resources
```
kannon92, line 524: Usually when we add a new API based on a feature gate, we also want to make sure that if this feature gate is disabled this API is not valid.

apis/config/v1beta2/configuration_types.go
```diff
@@ -520,6 +520,11 @@ type Resources struct {
...
+	// for partitionable device quota management.
```
kannon92, line 525: We should add a comment that this field is under a feature gate.

cmd/kueue/main.go
kannon92: I think we need to change this a bit. This logic is a bit confusing.

Should we fail validation if CounterMappings is set with feature gate off?

We should handle this in our default logic for pkg/config.

ie if feature is off, disable this field.

PannagaRao: DeviceClassMappings doesn't reject when DynamicResourceAllocation is off either — failing the validation for CounterMappings alone would be inconsistent.

PannagaRao: We could nil out CounterMappings in pkg/config/config.go after config loading when the gate is off. WDYT?

kannon92: That sounds good to me.

kannon92: > DeviceClassMappings doesn't reject when DynamicResourceAllocation is off either — failing the validation for CounterMappings alone would be inconsistent.

@mimowo @tenzen-y how should we handle API changes if feature gate is disabled?

I think validation errors in config or defaulting are both good options. I'd prefer failing validation honestly over defaulting.

PannagaRao: > I think we need to change this a bit. This logic is a bit confusing.
> 
> Should we fail validation if CounterMappings is set with feature gate off?
> 
> We should handle this in our default logic for pkg/config.
> 
> ie if feature is off, disable this field.

@kannon92 Moved the counterMappings nil-out to config.go using cfg.FeatureGates.

Wanted to discuss one case though — if an admin sets DRAPartitionableDevices via --feature-gates CLI flag but has counterMappings in the config file, cfg.FeatureGates won't have the gate and counterMappings gets dropped. Is this a case we don't have to worry about or would it be safer to put this check in main.go after feature gates are fully resolved from both sources?

kannon92: I guess I'd like to simplify the code so that we don't need to check counterMappings plus feature gate in a few places.

We really want to encourage the use of feature gates in the config rather than the CLI.

I wonder if we should announce deprecation of cli feature gates? WDYT @mimowo @tenzen-y ?

cmd/kueue/main.go
```diff
@@ -304,8 +304,9 @@ func main() {
...
+		(len(cfg.Resources.DeviceClassMappings) > 0 || len(cfg.Resources.CounterMappings) > 0) {
```
kannon92: This logic seems way more complicated than it needs to be. Plus we aren't testing any of the edge cases.

Could you refactor this?

Maybe if DRA is true, call dra.CreateMapperFromConfiguration.

And in that code you can handle checking for nils of the resources values. And that way you can add some test cases for this.

PannagaRao: Simplified — removed the length checks from main.go. CreateMapperFromConfiguration already handles nil inputs internally. 

I have added test cases for CreateMapperFromConfiguration covering: both nil, deviceClassMappings only, counterMappings only, and both configured. 

PTAL!

cmd/experimental/kueue-populator/go.mod
kannon92: This is no longer relevant. Please resolve.

cmd/kueue/main.go
kannon92: Thank you for fixing this.

kannon92:
> This PR has a dependency on https://github.com/kubernetes-sigs/kueue/pull/9742

CEL work has merged so this is ready to go. I guess we still need the KEP merged though.

PannagaRao:
I verified the current implementation on Nvidia A100 GPU with MIGenabled

Configuration:                                                                                                                                                                                                                                                                                                                                                                            
```
  resources:                                                                                                                                                                                                                                                                                                                                                                                
    deviceClassMappings:                                                                                                                                                                                                                                                                                                                                                                    
    - name: gpu.memory                                                                                                                                                                                                                                                                                                                                                                      
      deviceClassNames:                                                                                                                                                                                                                                                                                                                                                                     
      - gpu.nvidia.com                                                                                                                                                                                                                                                                                                                                                                      
      - mig.nvidia.com                                                                                                                                                                                                                                                                                                                                                                      
      counters:                                                                                                                                                                                                                                                                                                                                                                             
      - driverName: gpu.nvidia.com                                                                                                                                                                                                                                                                                                                                                          
        counterName: memory                                                                                                                                                                                                                                                                                                                                                                 
        defaultConsumption: "40Gi"
```
        
  
Mixed workload — whole GPU + MIG in one pod:                                                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                                                                            
  RCT 1 (whole GPU, no CEL):                                                                                                                                                                                                                                                                                                                                                                
```
  deviceClassName: gpu.nvidia.com                                                                                                                                                                                                                                                                                                                                                           
  count: 1  
```                                                                                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                                                            
  RCT 2 (MIG 1g.5gb, with CEL):
```
  deviceClassName: mig.nvidia.com                                                                                                                                                                                                                                                                                                                                                           
  count: 1                                                                                                                                                                                                                                                                                                                                                                                
  selectors:                                                                                                                                                                                                                                                                                                                                                                                
  - cel:          
      expression: "device.attributes['gpu.nvidia.com'].profile == '1g.5gb'" 
```                                                                                                                                                                                                                                                                                                                
                                                                           
  Pod spec:                                                                                                                                                                                                                                                                                                                                                                                 
```
  resourceClaims:                                                                                                                                                                                                                                                                                                                                                                           
  - name: whole                                                                                                                                                                                                                                                                                                                                                                             
    resourceClaimTemplateName: whole-gpu-rct                                                                                                                                                                                                                                                                                                                                                
  - name: mig                               
    resourceClaimTemplateName: mig-rct
```
    
 
ClusterQueue quota reservation:                                                                                                                                                                                                                                                                                                                                                           
```
  flavorsReservation:                                                                                                                                                                                                                                                                                                                                                                       
  - name: gpu-flavor                                                                                                                                                                                                                                                                                                                                                                        
    resources:                                                                                                                                                                                                                                                                                                                                                                              
    - name: cpu                                                                                                                                                                                                                                                                                                                                                                             
      total: 100m                                                                                                                                                                                                                                                                                                                                                                           
    - name: gpu.memory                                                                                                                                                                                                                                                                                                                                                                      
      total: "48049946624"   # 40Gi (whole GPU) + 4864Mi (MIG 1g.5gb) = 44864Mi                                                                                                                                                                                                                                                                                                           
    - name: memory                                                             
      total: 256Mi
```                                                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                                             Both charges against the same gpu.memory quota pool — unified accounting for whole-GPU and MIG workloads.

pkg/config/config.go
```diff
@@ -225,6 +225,11 @@ func Load(scheme *runtime.Scheme, configFile string) (ctrl.Options, configapi.Co
 			return options, cfg, err
 		}
 	}
+	if cfg.Resources != nil && !cfg.FeatureGates[string(features.KueueDRAIntegrationPartitionableDevices)] {
```
sohankunkerkar: Here it strips Counters from `deviceClassMappings` when the gate is not in cfg.FeatureGates. But `cmd/kueue/main.go` applies CLI feature gates after Load() returns `(Set(featureGates)`. So `--feature-gates=KueueDRAIntegrationPartitionableDevices=true` won't work as counters are already gone before the CLI flag is processed. This gate-dependent stripping needs to happen after feature gates are finalized, or move out of Load().

pkg/config/validation.go
```diff
@@ -503,6 +503,28 @@ func validateDeviceClassMappings(c *configapi.Configuration) field.ErrorList {
...
+					allErrs = append(allErrs, field.Required(counterPath.Child("driverName"), ""))
+				} else if len(counter.DriverName) > 253 {
+					allErrs = append(allErrs, field.Invalid(counterPath.Child("driverName"), counter.DriverName, "must not exceed 253 characters"))
+				}
+				if counter.CounterName == "" {
+					allErrs = append(allErrs, field.Required(counterPath.Child("counterName"), ""))
+				} else if len(counter.CounterName) > 63 {
+					allErrs = append(allErrs, field.Invalid(counterPath.Child("counterName"), counter.CounterName, "must not exceed 63 characters"))
+				}
+				if counter.ConsumesCounterSelector.CEL == nil || counter.ConsumesCounterSelector.CEL.Expression == "" {
```
sohankunkerkar: Here we check that the CEL expression is non-empty, but not that it compiles. A bad expression here won't be caught until a workload hits this mapping. KEP talks about `compiled at config load time.` Could we compile/validate the expression during config validation so invalid config fails early?

pkg/dra/counters.go
```diff
@@ -0,0 +1,279 @@
...
+		}
+		devices := pool.collectDevices()
+		for i := range devices {
+			dev := &devices[i]
+			celDev := dracel.Device{
+				Driver:     driver,
+				Attributes: dev.Attributes,
+				Capacity:   dev.Capacity,
+			}
+			if selectorMatch, _ := evaluateSelectorsOnDevice(ctx, selectorSelectors, celDev); !selectorMatch {
```
sohankunkerkar: Should we log the error here with V(3) instead of discarding it altogether?

pkg/dra/counters.go
```diff
@@ -0,0 +1,279 @@
...
+				continue
+			}
+			if classMatch, _ := evaluateSelectorsOnDevice(ctx, classSelectors, celDev); !classMatch {
```
sohankunkerkar: https://github.com/kubernetes-sigs/kueue/pull/10320/changes#r3263580105

pkg/dra/counters.go
```diff
@@ -0,0 +1,279 @@
...
+				continue
+			}
+			if requestMatch, _ := evaluateSelectorsOnDevice(ctx, requestSelectors, celDev); requestMatch {
```
sohankunkerkar: https://github.com/kubernetes-sigs/kueue/pull/10320/changes#r3263580105

pkg/controller/core/resourceslice_controller.go
```diff
@@ -0,0 +1,100 @@
...
+	return &ResourceSliceReconciler{
+		qManager: qManager,
+	}
+}
+
+func (r *ResourceSliceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	return ctrl.Result{}, nil
+}
+
+func (r *ResourceSliceReconciler) notifyAll() {
```
sohankunkerkar: nit: `notifyAll()` requeues inadmissible workloads for all ClusterQueues on every ResourceSlice event, regardless of driver. Not a blocker for alpha, but in clusters with unrelated DRA drivers this is broader than necessary.

sohankunkerkar (changes requested):

pkg/dra/counters.go
PannagaRao: Done

pkg/config/validation.go
PannagaRao: I have added the validation. PTAL!

pkg/config/config.go
PannagaRao: Good catch — I raised this same concern [here](https://github.com/kubernetes-sigs/kueue/pull/10320#discussion_r3090729350). The stripping was in Load() which runs before CLI feature gates are applied via LoadAndValidateFeatureGates. Kevin suggested deprecating CLI feature gates but nothing was decided. Should I move the stripping to main.go after both config and CLI gates are finalized?

sohankunkerkar: @mimowo do you have any insights here?

mimowo: > Kevin suggested deprecating CLI feature gates but nothing was decided. 

We cannot deprecate them as part of this PR. Make the code work with both ways of setting the FGs.

PannagaRao: Done.

pkg/dra/claims.go
```diff
@@ -217,6 +217,9 @@ func GetResourceRequestsForResourceClaimTemplates(
 					))
 					return nil, allErrs
 				}
+				if Mapper().getCounterConfig(dc) != nil {
```
sohankunkerkar:
```suggestion
				if features.Enabled(features.KueueDRAIntegrationPartitionableDevices) && Mapper().getCounterConfig(dc) != nil { 
```
When the `KueueDRAIntegrationPartitionableDevices` gate is off, this continue skips device-count charging but `GetCounterResourcesForWorkload` is never called either,  resulting in zero charge for the DeviceClass.

apis/config/v1beta2/configuration_types.go
```diff
@@ -586,6 +587,53 @@ type DeviceClassMapping struct {
...
+	Sources []DeviceClassSourceConfig `json:"sources,omitempty"`
+}
+
+// DeviceClassSourceConfig defines a resource accounting source for a DeviceClassMapping.
+// Exactly one of the source types must be set.
+type DeviceClassSourceConfig struct {
+	// Counter configures counter-based quota for partitionable devices.
+	// Maps a DRA driver counter to the parent DeviceClassMapping's Kueue quota resource.
+	// +optional
+	Counter *CounterSource `json:"counter,omitempty"`
```
sohankunkerkar:
```suggestion
	Counter *DeviceClassCounterSource `json:"counter,omitempty"`
```

apis/config/v1beta2/configuration_types.go
```diff
@@ -586,6 +587,53 @@ type DeviceClassMapping struct {
...
+}
+
+// CounterSource identifies where to read counter data from and which counter to track.
```
sohankunkerkar:
```suggestion
// DeviceClassCounterSource identifies where to read counter data from and which counter to track.
```

apis/config/v1beta2/configuration_types.go
```diff
@@ -586,6 +587,53 @@ type DeviceClassMapping struct {
...
+type CounterSource struct {
```
sohankunkerkar:
```suggestion
type DeviceClassCounterSource struct {
```

pkg/config/validation.go
```diff
@@ -503,6 +504,44 @@ func validateDeviceClassMappings(c *configapi.Configuration) field.ErrorList {
...
+				if source.Counter.Driver == "" {
+					allErrs = append(allErrs, field.Required(counterPath.Child("driver"), ""))
+				} else if len(source.Counter.Driver) > 253 {
+					allErrs = append(allErrs, field.Invalid(counterPath.Child("driver"), source.Counter.Driver, "must not exceed 253 characters"))
+				}
+				selectorPath := counterPath.Child("deviceSelector", "cel", "expression")
+				if source.Counter.DeviceSelector.CEL == nil || source.Counter.DeviceSelector.CEL.Expression == "" {
+					allErrs = append(allErrs, field.Required(selectorPath, ""))
+				} else {
+					cache := dracel.NewCache(1, dracel.Features{})
```
sohankunkerkar: This creates a new cache inside the loop for every source entry. Move it before https://github.com/kubernetes-sigs/kueue/pull/10320/changes#diff-3e0b3707cc5d712d06d1eeb6a67c10b45a7ba0b27e972924d71df3c8e539f23bR512 loop and reuse.

test/integration/singlecluster/controller/dra/dra_test.go
```diff
@@ -1254,3 +1255,842 @@ var _ = ginkgo.Describe("DRA Integration", ginkgo.Ordered, ginkgo.ContinueOnFail
 		})
 	})
 })
+
+var _ = ginkgo.Describe("DRA Partitionable Devices Integration", ginkgo.Ordered, ginkgo.ContinueOnFailure, func() {
```
sohankunkerkar: I think we should extract this test into a separate file, maybe?

pkg/dra/claims.go
PannagaRao: The guard in main.go (after LoadAndValidateFeatureGates) nils out Sources from all DeviceClassMappings when the gate is off, before CreateMapperFromConfiguration is called. So getCounterConfig() can never return non-nil when KueueDRAIntegrationPartitionableDevices is disabled.

I didn't add the gate check here to keep the feature gate enforcement at a single point (main.go), per the [earlier discussion](https://github.com/kubernetes-sigs/kueue/pull/10320#discussion_r3094440434). Happy to add it if you see a code path where getCounterConfig() could return non-nil with the gate off.

sohankunkerkar: ah, right! I  think this makes sense. Thanks for pointing this out.

sohankunkerkar:
Just a note: e2e tests have not been added to this PR to keep the PR size manageable for reviewers. You can take a look here: https://github.com/kubernetes-sigs/kueue/pull/11420

pkg/controller/core/resourceslice_controller.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func (r *ResourceSliceReconciler) Generic(event.TypedGenericEvent[*resourcev1.ResourceSlice]) bool {
+	return false
+}
+
+// +kubebuilder:rbac:groups=resource.k8s.io,resources=resourceslices,verbs=get;list;watch
+
+func (r *ResourceSliceReconciler) SetupWithManager(mgr ctrl.Manager, cfg *configapi.Configuration) error {
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &resourcev1.ResourceSlice{},
```
mimowo: Don\t use background context, pass the main context

pkg/controller/core/resourceslice_controller.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func (r *ResourceSliceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	return ctrl.Result{}, nil
+}
+
+func (r *ResourceSliceReconciler) notifyAll() {
+	cqNames := sets.New(r.qManager.GetClusterQueueNames()...)
+	qcache.NotifyRetryInadmissible(r.qManager, cqNames)
+}
```
mimowo, line 70: This looks like having a potential perf issue.

I think `NotifyRetryInadmissible` should be called from Reconcile. 

Only relevent changes to the ResourceSlice should be triggering reconcile

pkg/controller/core/resourceslice_controller.go
```diff
@@ -0,0 +1,100 @@
...
+var _ reconcile.Reconciler = (*ResourceSliceReconciler)(nil)
+
+func NewResourceSliceReconciler(qManager *qcache.Manager) *ResourceSliceReconciler {
+	return &ResourceSliceReconciler{
+		qManager: qManager,
+	}
+}
+
+func (r *ResourceSliceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	return ctrl.Result{}, nil
```
mimowo: wire logger with roleTracker and log the reconcile events with V3

pkg/dra/claims.go
mimowo: Wrap in FG

pkg/config/validation.go
```diff
@@ -503,6 +504,44 @@ func validateDeviceClassMappings(c *configapi.Configuration) field.ErrorList {
 				deviceClassToResource[deviceClass] = mapping.Name
 			}
 		}
+
+		if len(mapping.Sources) > 0 {
```
mimowo: check FG

cmd/kueue/main.go
```diff
@@ -155,6 +155,12 @@ func main() {
 		os.Exit(1)
 	}
 
+	if cfg.Resources != nil && !features.Enabled(features.KueueDRAIntegrationPartitionableDevices) {
+		for i := range cfg.Resources.DeviceClassMappings {
+			cfg.Resources.DeviceClassMappings[i].Sources = nil
+		}
+	}
```
mimowo: Hm, why clearing the field rather than just ignore based on the FG from within the code? I mean this approach works too, but ignoring depending on the FG is more consistent with the rest of Kueue code.

test/integration/singlecluster/controller/dra/dra_test.go
```diff
@@ -1254,3 +1255,842 @@ var _ = ginkgo.Describe("DRA Integration", ginkgo.Ordered, ginkgo.ContinueOnFail
...
+							Attributes: map[resourcev1.QualifiedName]resourcev1.DeviceAttribute{
+								"type":    {StringValue: new("mig")},
+								"profile": {StringValue: new("1g.5gb")},
+							},
+							ConsumesCounters: []resourcev1.DeviceCounterConsumption{
+								{CounterSet: "gpu-0-counter-set", Counters: map[string]resourcev1.Counter{"memory": {Value: resource.MustParse("4864Mi")}}},
+							},
+						},
+					},
+				},
```
mimowo: We need to introduce the Wrappers pattern for this

test/integration/singlecluster/controller/dra/dra_test.go
```diff
@@ -1254,3 +1255,842 @@ var _ = ginkgo.Describe("DRA Integration", ginkgo.Ordered, ginkgo.ContinueOnFail
...
+								DeviceSelector: resourcev1.DeviceSelector{
+									CEL: &resourcev1.CELDeviceSelector{
+										Expression: "device.driver == 'gpu.example.com'",
+									},
+								},
+							}},
+						},
+					},
+				})
+			}()
```
mimowo: oh this is very heavy to read. First of all don't use defer func in ginkgo test.

But I'm also not clear what is the role of the code. Is it to revert, then use AfterEach / After All

pkg/dra/mapper.go
```diff
@@ -18,10 +18,18 @@ package dra
...
 
+// deviceClassCounterConfig holds counter configuration for a specific DeviceClass.
+type deviceClassCounterConfig struct {
+	driver         string
+	counterName    string
+	deviceSelector resourcev1.DeviceSelector
+}
+
 var (
 	globalMapper *ResourceMapper
```
mimowo: It is not this PR, but I think we need to get rid of the global state for DRA.

This will be problematic for integration tests for MultiKueue, but the global state is generally making the code hard to follow. 

I think we could move the cache under `/pkg/cache/dra/`

cmd/kueue/main.go
PannagaRao: We discussed this approach [here](https://github.com/kubernetes-sigs/kueue/pull/10320#discussion_r3094440434) - centralized stripping in main.go after both config and CLI gates are finalized to keep the FG enforcement in one place. Downstream code (claims.go, validation.go, counters.go) just sees getCounterConfig() == nil when the gate is off, avoiding the risk of a missed FG check.

Happy to switch to distributed FG checks if you prefer consistency with the rest of Kueue.

pkg/dra/claims.go
PannagaRao: This is guarded by the stripping in main.go. getCounterConfig() can't return non-nil when the gate is off because Sources are nilled out before the mapper is created.

pkg/controller/core/resourceslice_controller.go
PannagaRao: Moved the field indexer into a standalone SetupResourceSliceIndexer(ctx, indexer) function, called from setupIndexes in main.go where the main context is available to match the existing pattern.

pkg/controller/core/resourceslice_controller.go
PannagaRao: Moved NotifyRetryInadmissible into Reconcile. The predicates now filter by driver (a drivers set built at startup) and only return true for ResourceSlices matching those drivers.

test/integration/singlecluster/controller/dra/dra_test.go
PannagaRao: Done

test/integration/singlecluster/controller/dra/dra_test.go
PannagaRao: Replaced all defer func with ginkgo.DeferCleanup.

pkg/config/validation.go
PannagaRao: Same as above. This is also  guarded by the stripping in main.go. When the gate is off because Sources are nilled out before the mapper is created.

cmd/kueue/main.go
PannagaRao: @mimowo I believe I have addressed all the other comments and would appreciate your thoughts on this one. Could you PTAL?

mimowo: > Downstream code (claims.go, validation.go, counters.go) just sees getCounterConfig() == nil when the gate is off, avoiding the risk of a missed FG check.

Let's add the FG checks to all places in the downsteam code. Making the check explicit is actually also good for searching of the logic injection. Otherwise it is really hard to find which logic relies on the cleaning. With FG check I can easily find all the logic injection points by simple search.

Sure, there is a risk of missing a check, but that is still better for code readabilty.

> Happy to switch to distributed FG checks if you prefer consistency with the rest of Kueue.

Yes, let's do that

mimowo:
This flake should be already fixed

cmd/kueue/main.go
PannagaRao: Done. Added FG checks and removed the stripping of Sources code.

pkg/dra/counters.go
```diff
@@ -0,0 +1,291 @@
...
+					log.V(3).Info("CEL evaluation error in DeviceClass selector", "device", dev.Name, "error", err)
+				}
+				continue
+			}
+			if requestMatch, err := evaluateSelectorsOnDevice(ctx, requestSelectors, celDev); requestMatch {
+				allMatched = append(allMatched, *dev)
+			} else if err != nil {
+				log.V(3).Info("CEL evaluation error in request selector", "device", dev.Name, "error", err)
+			}
+		}
```
mimowo, line 197: I'm wonering when the CEL evaluation errors may happen, but it seems to me they should not be silenced, but propagated up. Silencing errors like this is usually as sign of potential problems downsteam.

If we need to silence for some reason then it warrants comments,

mimowo: cc @sohankunkerkar wdyt?

pkg/dra/counters.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+
+	return nil, field.ErrorList{field.Invalid(
+		reqPath,
+		nil,
+		fmt.Sprintf("matched devices have no consumesCounters entry for counter %q", counterConfig.counterName),
+	)}
+}
+
+func matchDevicesWithSelectors(
```
mimowo, line 158: I think this function is complex enough to also have some logging in positive scenarios, probably at v3 or v4.

pkg/dra/counters.go
```diff
@@ -0,0 +1,291 @@
...
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	dracel "k8s.io/dynamic-resource-allocation/cel"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+)
+
+func GetCounterResourcesForWorkload(
```
mimowo, line 34: I think this function is complex enough to also have some logging, probably at v3 or v4.

pkg/dra/counters.go
```diff
@@ -0,0 +1,291 @@
...
+
+		if len(aggregated) > 0 {
+			perPodSet[ps.Name] = aggregated
+		}
+	}
+
+	return perPodSet, nil
+}
+
+func processCounterCharge(
```
mimowo, line 101: This function is certainly complex enough to have logging at v3 or v4 at least

pkg/dra/counters.go
sohankunkerkar: Yeah, I second this opinion.

mimowo (commented):
/lgtm
/approve
Since this is alpha feature I mostly focused on reviewing the isolation of the new code from the old code. 

In terms of the logic I rely mostly on the reviews from the DRA experts like @sohankunkerkar and @kannon92 . 

Still, I skimmed through and nothing really stood out. Proposed some logging improvements as follow ups, but nothing is blocking here for me.

Thank  you for the PR for DRA 👍

tenzen-y:
/release-note-edit

```release-note
DRA: Add counter-based quota support for partitionable DRA devices (e.g., NVIDIA MIG) via counterMappings configuration.  
```

Issue #10124, opened 2026-03-26
Add metrics for age of pending workloads and time to preemption


ns-sundar:
**What would you like to be added**:
This is a request to add 2 metrics:
* Age of pending workloads, by clusterqueue. Gauge metric, presentable as a time series.
* Time from decision to preempt a workload till the time the preemption completes (latency of preemption), by clusterqueue. 

**Why is this needed**:
We are facing issues in production where some jobs remain pending for a long time. So, we want the average/max times that jobs are waiting in each clusterqueue. The `kueue_admission_wait_time_seconds` gives a somewhat similar metric but only for admitted workloads, not pending ones. Also, it is a histogram, not a time series. To make things simpler, this could be the time till job admission from either job creation or eviction. 

Secondly, another issue we have seen is that, once Kueue decides to preempt a workload, it may take a while for the workload to terminate. That's because it may have finalizers, long pod grace period, volume unmount issues, etc.  It would be good to measure the 'preemption latency', i.e., the time from the decision to preempt to the time preemption completes (i.e., the other workload is gone). 

**Completion requirements**:
Have the metrics visible in Prometheus.

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] API change
- [ x ] Docs update

The artifacts should be linked in subsequent comments.

mimowo:
@ns-sundar certainly the proposed metrics are valuable for observability. At the same I'm wondering if a single metrics like `pending_workloads_wait_seconds{quantile=0.5|0.95|0.99}` could effectively cover the same use cases. With the quantile 0.5 capturing the median which is close to mean, and with 0.99 quantile we could get close to max. The two approaches are also not mutually exclusive, just wondering how flexible are the expectations here.

amy:
@ns-sundar I have a clarifying question for this metric. 

There's 2 approaches to implement `Age of pending workloads, by clusterqueue. Gauge metric, presentable as a time series.`:
1. A monotonically increasing live update of the age of pending workloads
2. The metric is captured when pending workload changes state to admitted or preempted once for you to retroactively analyze

Do you need 1? Or is 2 sufficient?

ns-sundar:
Hi Amy, #1 is better. Our intended use case is this: we see issues where some jobs are stuck without getting admitted. To catch that, we want to set an alert on the max age of pending workloads. That means the metric needs to be updated even when the workload is not yet admitted. 
For #2, we have kueue_admission_wait_time_seconds metric already.

[...]

Issue #8249, opened 2025-12-15
Add priority class label to resource-usage and workload metrics


dominikgrewe:
**What would you like to be added**:

I would like have an additional label for some metrics that specifies the priority class. In particular I would like the following metrics to have the additional priority class label:

- kueue_local_queue_resource_reservation
- kueue_local_queue_resource_usage
- kueue_local_queue_pending_workloads
- kueue_cluster_queue_resource_reservation
- kueue_cluster_queue_resource_usage

**Why is this needed**:

This helps us understand how resources are used across priority classes. For example, when looking into fair sharing, we would like to understand how fairly resources are distributed within a priority class. Or when trying to understand how busy the queues are it's helpful to know what priorities pending and active workloads have.

**Completion requirements**:

The metrics listed above have an additional label indicating the priority class.

kannon92:
Have you checked recent releases of Kueue?

We did have https://github.com/kubernetes-sigs/kueue/issues/5989 and I believe we addressed most of these asks.

dominikgrewe:
Yes, the ask is very similar but none of the metrics above were included in #5989 from what I can see.

kannon92:
cc @mbobrovskyi

mukund-murali:
For the reason described in the PR, adding `priority_class` to the below metrics would also be super helpful:
```
kueue_pending_workloads
kueue_reserving_active_workloads
kueue_admitted_active_workloads

kueue_local_queue_pending_workloads
kueue_local_queue_reserving_active_workloads
kueue_local_queue_admitted_active_workloads
```

andrewseif:
I would like to work on this feature/task 😃

andrewseif:
**LocalQueue Status Metrics:**

- [ ] #8426

mbobrovskyi:
This doesn’t look easy to implement. We are currently using the LocalQueue status as the source of metrics:

https://github.com/kubernetes-sigs/kueue/blob/1d5cae5e227e117dd331993f22b3d961afae6458/pkg/controller/core/localqueue_controller.go#L369-L380

We’ll probably need to make additional changes to implement this. We need to carefully consider the approach, as changes to the API are not easy.

mimowo:
Hi folks @andrewseif @mbobrovskyi what is the status here? Can we make some progress with some of the metrics? It is not clear to me which are hard and which are easy.

mimowo:
[...]

PR #10323, opened 2026-04-08
Add WorkloadEvictionLatencySeconds metric

vladikkuzn:
#### What type of PR is this?
/kind feature

Adds WorkloadEvictionLatencySeconds metric

Part of #10124

```release-note
Observability: Introduced the workload_eviction_latency_seconds histogram metric, which records the time from when an eviction starts to when it is finalized.
```


pkg/controller/core/workload_controller.go
```diff
@@ -1038,9 +1038,27 @@ func (r *WorkloadReconciler) Delete(e event.TypedDeleteEvent[*kueue.Workload]) b
 	r.logger().V(2).Info("Workload delete event", "workload", klog.KObj(e.Object), "queue", e.Object.Spec.QueueName, "status", status)
 	r.preemptionExpectations.ObservedUID(r.logger(),
 		client.ObjectKeyFromObject(e.Object), e.Object.UID)
+	if !e.DeleteStateUnknown {
```
mwielgus: Is it strictly necessary to exclude cases where DeleteStateUnknown is true? In controller-runtime, when an informer misses a deletion, e.Object still contains the last known state of the object, which might be perfectly valid and sufficient for extracting the preemption completion metric.

pkg/controller/core/workload_controller.go
```diff
@@ -1038,9 +1038,25 @@ func (r *WorkloadReconciler) Delete(e event.TypedDeleteEvent[*kueue.Workload]) b
 	r.logger().V(2).Info("Workload delete event", "workload", klog.KObj(e.Object), "queue", e.Object.Spec.QueueName, "status", status)
 	r.preemptionExpectations.ObservedUID(r.logger(),
 		client.ObjectKeyFromObject(e.Object), e.Object.UID)
+	r.observePreemptionCompletionMetrics(e.Object)
```
mwielgus: Why is it updated on workload delete? Preemption doesn't delete the Workload.

pkg/controller/core/workload_controller.go
```diff
@@ -1038,9 +1038,25 @@ func (r *WorkloadReconciler) Delete(e event.TypedDeleteEvent[*kueue.Workload]) b
...
+func (r *WorkloadReconciler) observePreemptionCompletionMetrics(wl *kueue.Workload) {
+	c := apimeta.FindStatusCondition(wl.Status.Conditions, kueue.WorkloadEvicted)
+	if c == nil || c.Status != metav1.ConditionTrue || c.Reason != kueue.WorkloadEvictedByPreemption {
+		return
+	}
+	var cqRef kueue.ClusterQueueReference
+	if wl.Status.Admission != nil && wl.Status.Admission.ClusterQueue != "" {
+		cqRef = wl.Status.Admission.ClusterQueue
+	} else {
+		cqRef = "unknown"
```
mwielgus: If wl.Status.Admission is nil, cqRef defaults to "unknown". With the cleanup of 
clears Status.Admission when a workload is preempted, would  hisresult in many
"unknown" entries in the metrics?

vladikkuzn: Agree, imo it should be read from oldWl.Status.Admission instead

mykysha:
Nit: I believe we can use doc comments in some places instead of named returns

mszadkow:
looks pretty good, were you able to test it manually?

pkg/cache/queue/cluster_queue.go
```diff
@@ -482,6 +482,23 @@ func (c *ClusterQueue) Pending() (int, int) {
...
+	active = append(active, c.heap.List()...)
+	if c.inflight != nil {
+		active = append(active, c.inflight)
+	}
+	inadmissible = make([]*workload.Info, 0, c.inadmissibleWorkloads.len())
+	for _, wi := range c.inadmissibleWorkloads {
+		inadmissible = append(inadmissible, wi)
+	}
+	return active, inadmissible
+}
```
mbobrovskyi:
```suggestion
func (c *ClusterQueue) pendingWorkloadInfos() ([]*workload.Info, []*workload.Info) {
	c.rwm.RLock()
	defer c.rwm.RUnlock()
	active := c.heap.List()
	if c.inflight != nil {
		active = append(active, c.inflight)
	}
	return active, slices.Collect(maps.Values(c.inadmissibleWorkloads))
}
```

pkg/cache/queue/metrics.go
```diff
@@ -43,12 +48,36 @@ func reportPendingWorkloads(m *Manager, cqRef kueue.ClusterQueueReference) {
...
+	active, inadmissible := len(activeInfos), len(inadmissibleInfos)
 	if m.statusChecker != nil && !m.statusChecker.ClusterQueueActive(cq.name) {
-		inadmissible += active
+		inadmissibleInfos = append(inadmissibleInfos, activeInfos...)
+		activeInfos = nil
+		inadmissible = len(inadmissibleInfos)
 		active = 0
 	}
-	metrics.ReportPendingWorkloads(cq.name, active, inadmissible, m.customLabels.CQGet(cq.name), m.roleTracker)
+	cl := m.clock
```
mbobrovskyi: Looks like redundant variable.

pkg/cache/queue/metrics.go
```diff
@@ -43,12 +48,36 @@ func reportPendingWorkloads(m *Manager, cqRef kueue.ClusterQueueReference) {
...
+	custom := m.customLabels.CQGet(cq.name)
+	metrics.ReportPendingWorkloads(cq.name, active, inadmissible, custom, m.roleTracker)
+	activeMax, activeMean := pendingWaitStats(activeInfos, cl)
+	inadmissibleMax, inadmissibleMean := pendingWaitStats(inadmissibleInfos, cl)
+	metrics.ReportPendingWorkloadWaitTimes(cq.name, activeMax, activeMean, inadmissibleMax, inadmissibleMean, custom, m.roleTracker)
+}
+
+func pendingWaitStats(infos []*workload.Info, cl clock.Clock) (maxSeconds, meanSeconds float64) {
```
mbobrovskyi:
```suggestion
func pendingWaitStats(infos []*workload.Info, cl clock.Clock) (float64, float64) {
```

pkg/cache/queue/metrics.go
```diff
@@ -43,12 +48,36 @@ func reportPendingWorkloads(m *Manager, cqRef kueue.ClusterQueueReference) {
 }
 
 func reportCQPendingWorkloads(m *Manager, cq *ClusterQueue) {
-	active, inadmissible := cq.Pending()
+	activeInfos, inadmissibleInfos := cq.PendingWorkloadInfos()
+	active, inadmissible := len(activeInfos), len(inadmissibleInfos)
```
mbobrovskyi: I think we can calculate it after if.

pkg/cache/queue/metrics.go
```diff
@@ -43,12 +48,36 @@ func reportPendingWorkloads(m *Manager, cqRef kueue.ClusterQueueReference) {
...
-		inadmissible += active
+		inadmissibleInfos = append(inadmissibleInfos, activeInfos...)
+		activeInfos = nil
+		inadmissible = len(inadmissibleInfos)
 		active = 0
 	}
-	metrics.ReportPendingWorkloads(cq.name, active, inadmissible, m.customLabels.CQGet(cq.name), m.roleTracker)
+	cl := m.clock
+	custom := m.customLabels.CQGet(cq.name)
+	metrics.ReportPendingWorkloads(cq.name, active, inadmissible, custom, m.roleTracker)
```
mbobrovskyi:
```suggestion
	metrics.ReportPendingWorkloads(cq.name, len(activeInfos), len(inadmissibleInfos), custom, m.roleTracker)
```

pkg/metrics/metrics.go
```diff
@@ -913,6 +956,33 @@ func ReportPendingWorkloads(cqName kueue.ClusterQueueReference, active, inadmiss
...
+	PendingWorkloadMeanWaitTimeSeconds.WithLabelValues(inadmissibleLabels...).Set(inadmissibleMeanWait)
+}
+
+// ReportPreemptionEvictionToPendingTime records latency from preemption eviction until the preemptee workload returns to Pending (quota released).
+func ReportPreemptionEvictionToPendingTime(cqName kueue.ClusterQueueReference, latency time.Duration, customLabelValues []string, tracker *roletracker.RoleTracker) {
+	labels := append([]string{string(cqName), roletracker.GetRole(tracker)}, customLabelValues...)
+	seconds := latency.Seconds()
+	if seconds < 0 {
+		seconds = 0
+	}
```
mbobrovskyi, line 985: Is it possible?

vladikkuzn: Added a comment:
```
// Negative duration is possible if the condition's LastTransitionTime is after the
// controller clock (clock skew, delayed watches). Clamp to zero for the histogram.
```

mbobrovskyi: Yeah, but is it possible for LastTransitionTime to be later than clock.Now()?

pkg/cache/queue/metrics.go
```diff
@@ -43,12 +48,32 @@ func reportPendingWorkloads(m *Manager, cqRef kueue.ClusterQueueReference) {
...
+		return 0, 0
+	}
+	var maxD time.Duration
+	var sum time.Duration
+	for _, wi := range infos {
+		d := workload.QueuedWaitTime(wi.Obj, cl)
+		sum += d
+		if d > maxD {
+			maxD = d
+		}
```
mbobrovskyi:
```suggestion
		maxD = max(d, maxD)
```

pkg/cache/queue/cluster_queue.go
```diff
@@ -482,6 +483,18 @@ func (c *ClusterQueue) Pending() (int, int) {
 	return c.pendingActive(), c.pendingInadmissible()
 }
 
+// PendingWorkloadInfos returns pending workload infos split into active (heap + inflight, no double-count)
+// and inadmissible, matching Pending() counts. Callers must not mutate returned *workload.Info or Obj.
+func (c *ClusterQueue) PendingWorkloadInfos() ([]*workload.Info, []*workload.Info) {
```
mbobrovskyi:
```suggestion
func (c *ClusterQueue) pendingWorkloadInfos() ([]*workload.Info, []*workload.Info) {
```

pkg/cache/queue/cluster_queue.go
```diff
@@ -482,6 +483,18 @@ func (c *ClusterQueue) Pending() (int, int) {
 	return c.pendingActive(), c.pendingInadmissible()
```
mbobrovskyi: It looks like we’re duplicating logic. Can we reuse pendingWorkloadInfos() instead?

```suggestion
	activeInfos, inadmissibleInfos := c.pendingWorkloadInfos()
	return len(activeInfos), len(inadmissibleInfos)
```

In that case we don't need pendingActive() and pendingInadmissible().

WDYT?

pkg/controller/core/workload_controller.go
```diff
@@ -1090,6 +1083,28 @@ func (r *WorkloadReconciler) Delete(e event.TypedDeleteEvent[*kueue.Workload]) b
...
+// preemptionEvictionToPendingObserveInput determines whether to record preemption eviction→pending latency.
+// cluster_queue is taken from oldWl's admission; latency is now minus the WorkloadEvicted condition's
+// LastTransitionTime on newWl (preemption reason).
+func preemptionEvictionToPendingObserveInput(oldWl, newWl *kueue.Workload, prevStatus, newStatus string, now time.Time) (cq kueue.ClusterQueueReference, latency time.Duration, ok bool) {
+	if prevStatus != workload.StatusQuotaReserved && prevStatus != workload.StatusAdmitted {
+		return "", 0, false
+	}
+	if newStatus != workload.StatusPending {
+		return "", 0, false
+	}
```
mbobrovskyi:
```suggestion
```

Do we need to double check this?

pkg/controller/core/workload_controller.go
```diff
@@ -1090,6 +1083,28 @@ func (r *WorkloadReconciler) Delete(e event.TypedDeleteEvent[*kueue.Workload]) b
 	return true
 }
 
+// preemptionEvictionToPendingObserveInput determines whether to record preemption eviction→pending latency.
+// cluster_queue is taken from oldWl's admission; latency is now minus the WorkloadEvicted condition's
+// LastTransitionTime on newWl (preemption reason).
+func preemptionEvictionToPendingObserveInput(oldWl, newWl *kueue.Workload, prevStatus, newStatus string, now time.Time) (cq kueue.ClusterQueueReference, latency time.Duration, ok bool) {
```
mbobrovskyi:
```suggestion
func preemptionEvictionToPendingObserveInput(oldWl, newWl *kueue.Workload now time.Time) (kueue.ClusterQueueReference, time.Duration, bool) {
```

pkg/controller/core/workload_controller.go
```diff
@@ -1090,6 +1083,28 @@ func (r *WorkloadReconciler) Delete(e event.TypedDeleteEvent[*kueue.Workload]) b
...
+	}
+	c := apimeta.FindStatusCondition(newWl.Status.Conditions, kueue.WorkloadEvicted)
+	if c == nil || c.Status != metav1.ConditionTrue || c.Reason != kueue.WorkloadEvictedByPreemption {
+		return "", 0, false
+	}
+	if oldWl.Status.Admission != nil && oldWl.Status.Admission.ClusterQueue != "" {
+		cq = oldWl.Status.Admission.ClusterQueue
+	} else {
+		cq = "unknown"
+	}
```
mbobrovskyi: If we don't have cq I think we should skip report.

mbobrovskyi: I think this function does too much. It would be helpful to add a workload.WorkloadClusterQueue(oldWl) helper and use it outside this function.

test/e2e/singlecluster/metrics_test.go
mbobrovskyi, line 1: Could you please add integration tests?

vladikkuzn: Is it worth putting those under `feature:prometheus` test tag?

vladikkuzn:
Metrics looks like this: 
image

pkg/metrics/metrics.go
vladikkuzn: Yes, it can happen in real clusters, because:
- LastTransitionTime is set using the apiserver’s notion of time when the object is written.
- r.clock.Now() is the controller’s clock.
- Skew between control-plane and node can make now.Sub(LastTransitionTime) negative.

pkg/controller/core/workload_controller.go
vladikkuzn: I would leave it if somebody uses this function in different place

test/integration/singlecluster/controller/core/workload_preemption_eviction_metrics_test.go
vladikkuzn, line 1: Is it better to put this test closer to metrics related tests?

pkg/metrics/metrics.go
mbobrovskyi: OK, I think adding an additional check wouldn’t hurt.

pkg/controller/core/workload_controller.go
mbobrovskyi: Can we move `(prevStatus == workload.StatusQuotaReserved || prevStatus == workload.StatusAdmitted) && status == workload.StatusPending` to some helper? We can reuse it in both places.

test/e2e/singlecluster/metrics_test.go
mbobrovskyi: Not sure – I don’t see this label in the integration tests.

vladikkuzn: Oh, yeah, good point, it's only for e2e

pkg/cache/queue/cluster_queue.go
```diff
@@ -479,24 +480,21 @@ func (c *ClusterQueue) PendingTotal() int {
 func (c *ClusterQueue) Pending() (int, int) {
 	c.rwm.RLock()
 	defer c.rwm.RUnlock()
-	return c.pendingActive(), c.pendingInadmissible()
+	active, inadmissible := c.pendingWorkloadInfos()
```
dkaluza: nit: we are increasing the memory usage with the new implementation, previous one used constant memory IIUC

pkg/metrics/metrics.go
```diff
@@ -88,6 +88,14 @@ var (
 	// +metricsdoc:labels=cluster_queue="the name of the ClusterQueue",status="status label (varies by metric)",replica_role="one of `leader`, `follower`, or `standalone`"
 	PendingWorkloads *prometheus.GaugeVec
 
+	// +metricsdoc:group=clusterqueue
+	// +metricsdoc:labels=cluster_queue="the name of the ClusterQueue",status="one of `active` or `inadmissible` (same as kueue_pending_workloads)",replica_role="one of `leader`, `follower`, or `standalone`"
+	PendingWorkloadMaxWaitTimeSeconds *prometheus.GaugeVec
+
+	// +metricsdoc:group=clusterqueue
+	// +metricsdoc:labels=cluster_queue="the name of the ClusterQueue",status="one of `active` or `inadmissible` (same as kueue_pending_workloads)",replica_role="one of `leader`, `follower`, or `standalone`"
+	PendingWorkloadMeanWaitTimeSeconds *prometheus.GaugeVec
```
mimowo, line 97: Instead of the two metrics could we rather publish the Historgram metric, then we could explain users how to compute, mean/max/median/95% etc using PromQL in here https://kueue.sigs.k8s.io/docs/tasks/manage/observability/common_grafana_queries/

mimowo: Ah, I think I see, the metric is just for those currently Pending, so it is "guage" in nature, rather than cummulative like histogram or counter. 

I see the case, still wondering if people solve similar problems somewhat more generically, basically I would love to avoid soon introducing "95%" metric, but for now no better idea from me.

vladikkuzn: Yeah, I though to introduce such metrics per WL, but this will blow up quickly I think

vladikkuzn: (cardinality explosion)

pkg/controller/core/workload_controller.go
```diff
@@ -1093,6 +1093,26 @@ func (r *WorkloadReconciler) Delete(e event.TypedDeleteEvent[*kueue.Workload]) b
 	return true
 }
 
+// preemptionEvictionToPendingObserveInput determines whether to record preemption eviction→pending latency.
+// Latency is now minus the WorkloadEvicted condition's LastTransitionTime on newWl (preemption reason).
+// Returns ok when status.admission.cluster_queue on oldWl is set and non-empty (see workload.WorkloadClusterQueue).
+func preemptionEvictionToPendingObserveInput(oldWl, newWl *kueue.Workload, now time.Time) (kueue.ClusterQueueReference, time.Duration, bool) {
```
dkaluza: nit: this function documentation is unclear. It looks like it has multiple responsibilities - it is computing the time difference and returning from which clusterqueue wl was evicted.

Consider separating those to make the code simpler.
We can just have a function that returns boolean if workload was evicted and transformed to pending state and easily compute the sub and retrieve the cq in the calling code.

```
func isPreemptionEvictionToPending(oldWl, newWl *kueue.Workload) bool {
	prevStatus := workload.Status(oldWl)
	newStatus := workload.Status(newWl)
	if !workload.FromQuotaReservedOrAdmittedToPending(prevStatus, newStatus) {
		return false
	}
	c := apimeta.FindStatusCondition(newWl.Status.Conditions, kueue.WorkloadEvicted)
	if c == nil || c.Status != metav1.ConditionTrue || c.Reason != kueue.WorkloadEvictedByPreemption {
		return  false
	}
	cq, found := workload.WorkloadClusterQueue(oldWl)
	if !found {
		return  false
	}
	return true
```

And then in the calling place:
```
if cq, found := workload.WorkloadClusterQueue(e.ObjectOld); found && preemptionEvictionToPendingObserveInput(e.ObjectOld, e.ObjectNew) {
            latency := r.clock.Now().Sub(c.LastTransitionTime.Time)
			metrics.ReportPreemptionEvictionToPendingTime(cq, latency, r.customLabels.CQGet(cq), r.roleTracker)
		}
```

pkg/controller/core/preemption_eviction_to_pending_test.go
```diff
@@ -0,0 +1,222 @@
...
+				Status: kueue.WorkloadStatus{
+					Conditions: []metav1.Condition{preemptedEvicted},
+				},
+			},
+			wantOK:      true,
+			wantCQ:      "cq-a",
+			wantLatency: 30 * time.Second,
+		},
+		{
+			name: "emit quota reserved to pending",
```
dkaluza: Is preempted evicted the only reason of workload being in pending status?

If i understand correctly there can be also other reasons why workload is pending, e.g. admission check failing.

Maybe it is worth to be more accurate in the name to avoid confusion.

pkg/metrics/metrics.go
```diff
@@ -628,6 +660,17 @@ The label 'reason' can have the following values:
 		}, append([]string{"preempting_cluster_queue", "reason", "replica_role"}, extraLabels...),
 	)
 
+	PreemptionEvictionToPendingSeconds = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: constants.KueueName,
+			Name:      "preemption_eviction_to_pending_seconds",
+			Help: `The time from preemption eviction (WorkloadEvicted with reason Preempted) until the preemptee workload returns to Pending (quota released).
+Observed on status transition from admitted or quota-reserved to pending while still evicted by preemption.
+Uses the eviction condition LastTransitionTime on the updated object as start; cluster_queue is taken from status.admission.cluster_queue on the pre-update object when set and non-empty (otherwise the histogram is not recorded).`,
```
dkaluza:
```suggestion
Uses the eviction condition LastTransitionTime on the updated object as start; cluster_queue is taken from status.admission.cluster_queue on the pre-update object when set and non-empty (otherwise the entry is not recorded).`,
```

nit: A single emission from one workload is not a hist, it is just a latency value. They create a histogram after aggregation with other entries.

pkg/cache/queue/metrics.go
```diff
@@ -43,12 +48,30 @@ func reportPendingWorkloads(m *Manager, cqRef kueue.ClusterQueueReference) {
...
-		inadmissible += active
-		active = 0
+		inadmissibleInfos = append(inadmissibleInfos, activeInfos...)
+		activeInfos = nil
+	}
+	custom := m.customLabels.CQGet(cq.name)
+	metrics.ReportPendingWorkloads(cq.name, len(activeInfos), len(inadmissibleInfos), custom, m.roleTracker)
+	activeMax, activeMean := pendingWaitStats(activeInfos, m.clock)
+	inadmissibleMax, inadmissibleMean := pendingWaitStats(inadmissibleInfos, m.clock)
+	metrics.ReportPendingWorkloadWaitTimes(cq.name, activeMax, activeMean, inadmissibleMax, inadmissibleMean, custom, m.roleTracker)
```
dkaluza: Reporting 0 mean and 0 max when there are no workloads for me is highly misleading.  IMO it would be best to not report metric entries in this case to indicate no data - this can be probably easily fixed if we switch to hist based pending wait times.

pkg/metrics/metrics.go
```diff
@@ -968,6 +1041,7 @@ func ClearClusterQueueMetrics(cq kueue.ClusterQueueReference) {
 	EvictedWorkloadsTotal.DeletePartialMatch(prometheus.Labels{"cluster_queue": cqName})
 	EvictedWorkloadsOnceTotal.DeletePartialMatch(prometheus.Labels{"cluster_queue": cqName})
 	PreemptedWorkloadsTotal.DeletePartialMatch(prometheus.Labels{"preempting_cluster_queue": cqName})
+	PreemptionEvictionToPendingSeconds.DeletePartialMatch(prometheus.Labels{"cluster_queue": cqName})
```
dkaluza, line 999: Why are we not clearing other pending workload metrics for cq here?

test/e2e/singlecluster/metrics_test.go
```diff
@@ -180,6 +182,8 @@ var _ = ginkgo.Describe("Metrics", ginkgo.Label("area:singlecluster", "feature:m
 
 			deletedMetrics := [][]string{
 				{"kueue_pending_workloads", clusterQueue.Name},
+				{"kueue_pending_workload_max_wait_time_seconds", clusterQueue.Name},
+				{"kueue_pending_workload_mean_wait_time_seconds", clusterQueue.Name},
```
dkaluza: What about preemption metric?

site/content/en/docs/reference/metrics.md
```diff
@@ -45,9 +45,12 @@ Use the following metrics to monitor the status of your ClusterQueues:
...
+| `kueue_preemption_eviction_to_pending_seconds` | Histogram | The time from preemption eviction (WorkloadEvicted with reason Preempted) until the preemptee workload returns to Pending (quota released).<br>Observed on status transition from admitted or quota-reserved to pending while still evicted by preemption.<br>Uses the eviction condition LastTransitionTime on the updated object as start; cluster_queue is taken from status.admission.cluster_queue on the pre-update object when set and non-empty (otherwise the histogram is not recorded). | `cluster_queue`: the preempted workload's ClusterQueue from status.admission before quota was released, or `unknown` if unset<br> `replica_role`: one of `leader`, `follower`, or `standalone` |
```
dkaluza: Can the cq really be `unknown` here? Aren't we checking that cq exists with every emission?

pkg/metrics/metrics.go
dkaluza: But cannot gauge histogram be easily achieved by histogram and rate expression?

Gauge histograms are [mentioned](https://prometheus.io/docs/concepts/metric_types/#histogram) in Prometheus as achievable this way. And functions like [quantiles](https://prometheus.io/docs/prometheus/latest/querying/functions/#histogram_quantile) support that IIUC.

However getting an accurate max value can be tricky in this way

pkg/cache/queue/cluster_queue.go
```diff
@@ -479,24 +480,21 @@ func (c *ClusterQueue) PendingTotal() int {
...
-// workloads that are in the admission queue.
-func (c *ClusterQueue) pendingActive() int {
-	result := c.heap.Len()
+// pendingWorkloadInfos returns pending workload infos split into active (heap + inflight, no double-count)
+// and inadmissible.
+// Callers must not mutate returned *workload.Info or Obj.
+func (c *ClusterQueue) pendingWorkloadInfos() ([]*workload.Info, []*workload.Info) {
+	c.rwm.RLock()
+	defer c.rwm.RUnlock()
+	active := c.heap.List()
```
mimowo: Actually, I'm concerned about the performance impact of this change, because this is something we call on every scheduling cycle via reportCQPendingWorkloads. So, effectively it makes the use of heap useless, because we do  Pop head with O(1) just in the next function call to list all workloads .
I'm not sure if this would show up in production use, but this is risky change actually.

I'm wondering a better design for such heavy metrics (and maybe actually all metrics) would be to trigger their computation asynchronously, similarly as we trigger the requeue asynchronously, using the workqueue. Then here we would just say enqueue the computation task, that would also update the metric at the at. For reference check how we do it with workload [requeuer](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/cache/queue/inadmissible_workloads.go#L243).

This is ok, because Prometheus metrics are scraped only with 15s intervals, so users anyway need to understand they might be lagging a little bit.

pkg/metrics/metrics.go
mimowo: > Gauge histograms are mentioned in Prometheus as achievable this way. And functions like quantiles support that IIUC.

If there is something standard then great, that should be investigated more.

Otherwise I'm ok to consolidate the two metrics into `PendingWorkloadsWaitSeconds{quantile: 0.5|0.95|0.99}`.
Then PendingWorkloadsWaitSeconds with 0.5 quantile would effectively be covering the Average use case, while PendingWorkloadsWaitSeconds 0.99 would cover the  PendingWorkloadMaxWaitTimeSeconds. 

> However getting an accurate max value can be tricky in this way

I wouldn't worry about the max metrics, I think it can be approximated pretty well in the standard way by 0.99 or 0.999 percentile, effectively.

dkaluza: > I wouldn't worry about the max metrics, I think it can be approximated pretty well in the standard way by 0.99 or 0.999 percentile, effectively.

Fully agree

pkg/cache/queue/cluster_queue.go
dkaluza: From my point of view, the problem is not with metrics in general, but with metrics that are describing a currently changing state like "live latency" of pending workloads. The fact that those are live requires recomputation here as some time delta occurred.

I wouldn't generalize this to all of the metrics as in many situations we have the exact value in hand - e.g. when workload transitions the state from pending to admitted and we are reporting number of pending workloads. Or when we are reporting latency on status change.
Going through historic status changes after the event IMO woukd lead only to unnecessary additional computation.

Here though a job measuring the values every X seconds sounds like a good solution.

The other that I see is to remove the "live" factor from them and report them on status change, in this case they will no longer change, and therefore no iteration is needrd. However it has a downside of larger lag and lack of visibility if they actually never change the status (besides count).

3rd approach is to try to optimize this with some dedicated structure, however I do not think it is worth it.

mimowo: > I wouldn't generalize this to all of the metrics as in many situations we have the exact value in hand - e.g. when workload transitions the state from pending to admitted and we are reporting number of pending workloads. Or when we are reporting latency on status change.
Going through historic status changes after the event IMO woukd lead only to unnecessary additional computation.

Agree, it may not be silver bullet, but probably there are more metrics which would fit into this category that delayed computation is ok, for example the pending_workloads guage metric. 

> Here though a job measuring the values every X seconds sounds like a good solution.

Indeed, this is were I'm leaning to now, so thank you for confirming. 

Also, we recently wanted to add priority_class to pending_workloads guage metric with @mbobrovskyi and this turned out very tricky to make performance reasons - in the previous "sync" approach it requires predefined tricky structures bucketing by priority, and so complexity is big. With this new async approach we could have all the time in the world to aggregate all the listed workloads per priority.

For reference: https://github.com/kubernetes-sigs/kueue/issues/8249

> The other that I see is to remove the "live" factor from them and report them on status change, in this case they will no longer change, and therefore no iteration is needrd. However it has a downside of larger lag and lack of visibility if they actually never change the status (besides count).

Maybe this would work sensibly in practice, but it would not give good results when there is no status change on CQ for 30min, because all workloads continue computations, and no new workloads are submitted. Then we would show the stale pending_wait_time, while the waiting time actually increased by 30min. It would be worth-considering alternative if we couldn't just make it async for some reason.

pkg/cache/queue/cluster_queue.go
```diff
@@ -479,24 +480,21 @@ func (c *ClusterQueue) PendingTotal() int {
...
-// pendingActive returns the number of active pending workloads,
-// workloads that are in the admission queue.
-func (c *ClusterQueue) pendingActive() int {
-	result := c.heap.Len()
+// pendingWorkloadInfos returns pending workload infos split into active (heap + inflight, no double-count)
+// and inadmissible.
+// Callers must not mutate returned *workload.Info or Obj.
+func (c *ClusterQueue) pendingWorkloadInfos() ([]*workload.Info, []*workload.Info) {
+	c.rwm.RLock()
+	defer c.rwm.RUnlock()
```
mbobrovskyi: Why are we locking it again in this function? We already have a lock in the calling functions.

pkg/metrics/metrics.go
```diff
@@ -913,6 +928,19 @@ func ReportPendingWorkloads(cqName kueue.ClusterQueueReference, active, inadmiss
...
+// ReportPreemptionEvictionToPendingTime records latency from preemption eviction until the preemptee workload returns to Pending (quota released).
+func ReportPreemptionEvictionToPendingTime(cqName kueue.ClusterQueueReference, latency time.Duration, customLabelValues []string, tracker *roletracker.RoleTracker) {
+	labels := append([]string{string(cqName), roletracker.GetRole(tracker)}, customLabelValues...)
+	seconds := latency.Seconds()
+	// LastTransitionTime comes from the API object (apiserver clock). This path compares it to the
+	// controller's clock.Now(), so skew between apiserver and controller (or delayed watch delivery)
+	// can make the transition time appear after "now" and yield a negative duration. Clamp to zero.
+	if seconds < 0 {
+		seconds = 0
+	}
```
mimowo: nit, just inline `seconds = max(0, latency.Seconds())`

mimowo:
Nice one, thank you 👍 

cc @ns-sundar @amy

pkg/cache/queue/cluster_queue.go
vladikkuzn: This metric is moved to a separate PR: https://github.com/kubernetes-sigs/kueue/pull/10633

pkg/cache/queue/metrics.go
vladikkuzn: This metric is moved to a separate PR: https://github.com/kubernetes-sigs/kueue/pull/10633

pkg/cache/queue/cluster_queue.go
vladikkuzn: This metric is moved to a separate PR: https://github.com/kubernetes-sigs/kueue/pull/10633

pkg/metrics/metrics.go
vladikkuzn: Isn't it covered by clearScopedGaugeMetrics already, or am I missing something?

pkg/metrics/metrics.go
```diff
@@ -628,6 +632,18 @@ The label 'reason' can have the following values:
 		}, append([]string{"preempting_cluster_queue", "reason", "replica_role"}, extraLabels...),
 	)
 
+	PreemptionEvictionToPendingSeconds = prometheus.NewHistogramVec(
```
mimowo, line 635: Instead of introducing preemption-oriented metric I would prefer to generalize and add the "reason" label as we do for workload_evictions_total metric, and other eviction-related metrics.

pkg/metrics/metrics.go
```diff
@@ -628,6 +632,18 @@ The label 'reason' can have the following values:
...
+		prometheus.HistogramOpts{
+			Subsystem: constants.KueueName,
+			Name:      "preemption_eviction_to_pending_seconds",
```
mimowo, line 638: Can we maybe name it `workload_eviction_latency_seconds`, because:
- I think this would be more descriptive to users rather than "eviction_to_pending"
- generalize the "preemption (also another comment)

pkg/metrics/metrics.go
vladikkuzn: Do you mean to remove this filter, and add reason as a label: `c.Reason != kueue.WorkloadEvictedByPreemption`?

mimowo: Exactly, and record the reason as a label, check as we do for https://github.com/kubernetes-sigs/kueue/blob/main/pkg/metrics/metrics.go#L545-L564

pkg/metrics/metrics.go
dkaluza: It is, thanks for pointing this out.

pkg/controller/core/workload_controller.go
```diff
@@ -1174,7 +1181,12 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 		if afs.Enabled(r.admissionFSConfig) && status == workload.StatusAdmitted && r.cache.ClusterQueueUsesAdmissionFairSharing(wlCopy.Status.Admission.ClusterQueue) {
 			r.updateAfsConsumedUsage(log, wlCopy)
 		}
-	case (prevStatus == workload.StatusQuotaReserved || prevStatus == workload.StatusAdmitted) && status == workload.StatusPending:
+	case workload.FromQuotaReservedOrAdmittedToPending(prevStatus, status):
+		if c := evictionQuotaReleaseToPendingCondition(e.ObjectOld, e.ObjectNew); c != nil {
+			cq, _ := workload.WorkloadClusterQueue(e.ObjectOld)
```
mimowo: Only return one value from WorkloadClusterQueue, if the second is dumped anyway.

pkg/controller/core/workload_controller.go
```diff
@@ -1174,7 +1181,12 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 		if afs.Enabled(r.admissionFSConfig) && status == workload.StatusAdmitted && r.cache.ClusterQueueUsesAdmissionFairSharing(wlCopy.Status.Admission.ClusterQueue) {
 			r.updateAfsConsumedUsage(log, wlCopy)
 		}
-	case (prevStatus == workload.StatusQuotaReserved || prevStatus == workload.StatusAdmitted) && status == workload.StatusPending:
+	case workload.FromQuotaReservedOrAdmittedToPending(prevStatus, status):
+		if c := evictionQuotaReleaseToPendingCondition(e.ObjectOld, e.ObjectNew); c != nil {
```
mimowo: I would rather rename the function to evictionLatency, and include the latency calculation. It will be easier to unit test.

pkg/controller/core/workload_eviction_latency_test.go
```diff
@@ -0,0 +1,284 @@
...
+	"testing"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+	"sigs.k8s.io/kueue/pkg/workload"
+)
+
+func TestEvictionQuotaReleaseToPendingCondition(t *testing.T) {
```
mimowo: Rather than asserting on the contiion I would rather assert on the latency itself, asked the comment above also

pkg/controller/core/workload_eviction_latency_test.go
```diff
@@ -0,0 +1,284 @@
...
+				t.Fatalf("condition present=%v want=%v", gotOK, tc.wantOK)
+			}
+			if tc.wantOK && got.Reason != tc.wantReason {
+				t.Fatalf("reason: got %q want %q", got.Reason, tc.wantReason)
+			}
+		})
+	}
+}
+
+func TestWorkloadEvictionLatencyAndClusterQueue(t *testing.T) {
```
mimowo: I think we can drop this test once we change the function to return the latency directly

pkg/workload/workload.go
```diff
@@ -1280,6 +1285,18 @@ func HasQuotaReservation(w *kueue.Workload) bool {
...
+func WorkloadClusterQueue(wl *kueue.Workload) (kueue.ClusterQueueReference, bool) {
+	if wl == nil || wl.Status.Admission == nil {
+		return "", false
+	}
+	cq := wl.Status.Admission.ClusterQueue
+	if cq == "" {
+		return "", false
+	}
+	return cq, true
+}
```
mimowo: I think this could be part of the function computing the latency

amy:
Thank you!

dkaluza (approved):
lgtm

mimowo:
Metrics require release notes for sure
/release-note-edit
```release-note
Observability: introduce the workload_eviction_latency_seconds metric which is a histogram recording the 
the time from when an eviction is started to when it is finalized.
```

mimowo:
/lgtm
/approve
Thank you 👍

mimowo:
/release-note-edit
```release-note
Observability: Introduced the workload_eviction_latency_seconds histogram metric, which records the time from when an eviction starts to when it is finalized.
```

Issue #10200, opened 2026-03-30
MultiKueue e2e tests are running too long


mimowo:
**What would you like to be cleaned**:

One idea is to split the Job into parallel and sequential (disruptive) suites

I like (2.), because some tests are disruptive and cannot run in parallel. However, I think most e2e tests for could run in parallel using the namespaces, and namespace suffix for cluster-scoped objects as we do for regular e2e tests.

**Why is this needed**:

Such long build impacts release process, and development process.

mimowo:
oh yes, my initial description had point (1.) to bump resources (CPU and memory) for the CI jobs. We can do that too

mszadkow:
As of now we don't utilize ginkgo ability to parallelize, e2e MK tests are run for 1 proc/node:
```
E2E_KIND_VERSION="kindest/node:v1.35.0" KIND_CLUSTER_NAME=kind \
	ARTIFACTS="/logs/artifacts/run-test-multikueue-e2e-1.35.0" IMAGE_TAG=us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue:v0.18.0-devel-6-gf973c9168 GINKGO_ARGS="" \
	E2E_MODE=ci \
	E2E_SKIP_REINSTALL=false \
	E2E_ENFORCE_OPERATOR_UPDATE=false \
	APPWRAPPER_VERSION=v1.2.0 \
	JOBSET_VERSION=v0.11.1 KUBEFLOW_VERSION=v1.9.3 \
	KUBEFLOW_MPI_VERSION=v0.8.0 \
	KUBERAY_VERSION=v1.6.0 RAY_VERSION=2.41.0 RAYMINI_VERSION=0.0.1 USE_RAY_FOR_TESTS="raymini" \
	KUBEFLOW_TRAINER_VERSION=v2.2.0 \
	LEADERWORKERSET_VERSION=v0.8.0 \
	CLUSTERPROFILE_VERSION=v0.0.0-20251124125836-445319b6307a \
	CLUSTERPROFILE_PLUGIN_IMAGE_VERSION=0.0.1 \
	TEST_LOG_LEVEL=-3 \
	E2E_RUN_ONLY_ENV=false \
	E2E_USE_HELM=false \
	./hack/testing/e2e-multikueue-test.sh
```

```
[...]

PR #10356, opened 2026-04-09
Move disruptive MultiKueue tests to separate suite

mszadkow:
#### What type of PR is this?
/kind cleanup
/area testing

Separate disruptive tests to allow for parallel run of other tests.

Relates to #10200

```release-note
NONE
```


test/e2e/multikueue/e2e_test.go
mimowo, line 1789: IIUC this is the pre-requistite step to enable running MK tests in parallel, those which aren't disruptive.
So, there will be a follow up work to use the namespace-suffix for the Cluster-scoped resources here, right @mszadkow ?

mimowo:
cc @kshalot @olekzabl

Makefile-test.mk
```diff
@@ -198,6 +201,10 @@ test-e2e-dra: setup-e2e-env run-test-e2e-dra-$(E2E_KIND_VERSION:kindest/node:v%=
 .PHONY: test-e2e-multikueue-dra
 test-e2e-multikueue-dra: setup-e2e-env run-test-e2e-multikueue-dra-$(E2E_KIND_VERSION:kindest/node:v%=%)
 
+.PHONY: test-e2e-multikueue-disruptive
+test-e2e-multikueue-disruptive: setup-e2e-env run-test-e2e-multikueue-disruptive-$(E2E_KIND_VERSION:kindest/node:v%=%)
```
olekzabl: Duplicate of lines 161-162?

test/e2e/multikueue-sequential/e2e_test.go
```diff
@@ -0,0 +1,917 @@
...
+	client.Object
+	*T
+}
+
+func expectObjectToBeDeletedOnWorkerClusters[PtrT objAsPtr[T], T any](ctx context.Context, obj PtrT) {
+	gomega.EventuallyWithOffset(1, func(g gomega.Gomega) {
+		util.ExpectObjectToBeDeleted(ctx, k8sWorker1Client, obj, false)
+		util.ExpectObjectToBeDeleted(ctx, k8sWorker2Client, obj, false)
+	}, util.Timeout, util.Interval).Should(gomega.Succeed())
+}
```
olekzabl, line 913: AFAICS these helpers got duplicated - should we move them to some shared util file?

test/e2e/multikueue-disruptive/e2e_test.go
```diff
@@ -0,0 +1,917 @@
...
+	})
+
+	ginkgo.When("Worker cluster is unreliable", ginkgo.Ordered, func() {
+		ginkgo.It("Should update the cluster status to reflect the connection state", func() {
+			regularFlavor := utiltestingapi.MakeResourceFlavor("default-regular").Obj()
+			util.MustCreate(ctx, k8sWorker1Client, regularFlavor)
+
+			worker1Cq2 := utiltestingapi.MakeClusterQueue("q2").
+				ResourceGroup(
+					*utiltestingapi.MakeFlavorQuotas(regularFlavor.Name).
```
olekzabl: This part is not a pure move - you introduced `regularFlavor` (and also changed the test name).
And - if I'm not mistaken - these are your additions, not some older version sneaking in.
Out of curiosity, why?

test/e2e/config/multikueue-disruptive/controller_manager_config.yaml
```diff
@@ -0,0 +1,24 @@
...
+clientConnection:
+  qps: 50
+  burst: 100
+integrations:
+  frameworks:
+  - "batch/job"
+  - "pod"
+featureGates:
+  LocalQueueMetrics: true
+  ElasticJobsViaWorkloadSlices: true
```
olekzabl: Nit: no newline at end of file

olekzabl (approved):

test/e2e/multikueue-sequential/e2e_test.go
mszadkow: Yes, I plan a follow-up  or maybe I could provide prep for that functions, I didn't want to mix those.

Makefile-test.mk
```diff
@@ -153,7 +153,13 @@ test-multikueue-e2e-parallel-builds:
...
 
 .PHONY: test-multikueue-e2e
-test-multikueue-e2e: setup-e2e-env test-multikueue-e2e-parallel-builds run-test-multikueue-e2e-$(E2E_KIND_VERSION:kindest/node:v%=%)
+test-multikueue-e2e: test-multikueue-e2e-main test-multikueue-e2e-disruptive
+
+.PHONY: test-multikueue-e2e-main
+test-multikueue-e2e-main: setup-e2e-env test-multikueue-e2e-parallel-builds run-test-multikueue-e2e-$(E2E_KIND_VERSION:kindest/node:v%=%)
+
+.PHONY: test-multikueue-e2e-disruptive
+test-multikueue-e2e-disruptive: setup-e2e-env run-test-e2e-multikueue-disruptive-$(E2E_KIND_VERSION:kindest/node:v%=%)
```
kshalot: I realize that the name "disruptive" comes from the original issue, but wouldn't "sequential" be more self explanatory?

When I'm adding a test, I think it's more natural for me to notice that my test needs to run sequentially rather than considering them to be somehow disruptive. It's also not clear what is being disrupted by the tests without prior context.

WDYT?

test/e2e/multikueue/e2e_test.go
mszadkow: Yes, that's correct

Makefile-test.mk
mszadkow: Thanks, I couldn't find a better name, this sounds much more descriptive, without introducing this "worrying" scent ;)

mimowo: Yes, I like sequential as well. Maybe in the future we will also have tests which need to run sequentiall but are not disruptive, for example check some metrics which are global and can be affected by any test.

test/e2e/multikueue-disruptive/e2e_test.go
mszadkow: My bad, I have mistakenly added the flavor here.
The name is a different thing, that left after initial approach with labels, cuz I wanted to group them under one test container.
I will amke sure there are no more name changes, thanks for spotting this.

Makefile-test.mk
olekzabl: For me, "sequential" also would be likely unclear without prior context. I think I'd then imagine some "sequence" _inside_ the test, while (IIUC) here we're talking about _external_ arrangement, whether a particular test can execute in parallel with some others.

That said, I'm not contesting the name, I like it. (Especially in contrast with "parallel", it just feels natural).
I'm just thinking, maybe you could add some comment (e.g. on top of `multikueue-sequential/e2e_test.go`, or wherever feels more appropriate) to clarify the meaning?

kshalot: We can also consider being more verbose and use `sequentially-executed`, but I'm not sure it's really that much better. I agree with the remark about adding a comment.

mimowo: I think sequential is preferred as it is short. 

btw, we also have similar pattern for singlecluster, we call them customconfigs, but I think we could also rename to sequential, because there also some tests aren't really testing different configurations, but for example cert management. So "customconfigs", similarly as "disruptive" was the idea to name them after the main use case, but since the primary use case is not the only after a while, I think it makes sense to call more generically, reflecting what is actually happening. The name "sequential" would also fit both "multikueue" and "singlecluster" test.

mimowo (commented):
Thank you 👍 
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16

mimowo:
seems like the robot cannot do the CPs, @mszadkow could you prepare manually?

mimowo:
Please later also follow up with the test-infra changes.

PR #10357, opened 2026-04-09
feat: add job_to_workload_latency_seconds metric

Nilsachy:
#### What type of PR is this?

/kind feature

This PR introduces a new Prometheus histogram metric, workload_creation_latency_seconds, to track the latency between when a job-like resource is created and when Kueue creates its corresponding Workload.

This metric is needed to help operators monitor the responsiveness of Kueue in acknowledging new jobs and to detect potential bottlenecks or delays in the reconciliation queue.

Fixes #

- The metric is implemented as a Histogram to allow calculating percentiles (p50, p95, p99).
- It uses job_kind and replica_role as default labels, and also supports any user-configured extraLabels.

```release-note
Observability: Added a new Prometheus metric `workload_creation_latency_seconds` to track the time elapsed between a Job's creation and the creation of its corresponding Workload by Kueue.
```


pkg/controller/jobframework/reconciler.go
```diff
@@ -1588,6 +1588,17 @@ func (r *JobReconciler) handleJobWithNoWorkload(ctx context.Context, job Generic
 	}
 	r.record.Eventf(object, corev1.EventTypeNormal, ReasonCreatedWorkload,
 		"Created Workload: %v", workload.Key(wl))
+
+	creationTime := wl.CreationTimestamp.Time
+	if creationTime.IsZero() {
+		creationTime = r.clock.Now()
+	}
```
kshalot: this is unused if we don't enter the next `if`. I'd nest it there.

pkg/controller/jobframework/reconciler.go
```diff
@@ -1588,6 +1588,17 @@ func (r *JobReconciler) handleJobWithNoWorkload(ctx context.Context, job Generic
...
+	if !job.Object().GetCreationTimestamp().Time.IsZero() {
```
kshalot: You can consider moving `job.Object().GetCreationTimestamp().Time` to a variable like `jobCreationTime` because this chain is called twice. It's not a big overhead, but it would potentially reduce some visual noise.

Then you can rename the existing `creationTime` var to `wlCreationTime`. And latency would become:
```go
latency := wlCreationTime.Sub(jobCreationTime)
```
which IMO is slightly more readable than:
```go
latency := creationTime.Sub(job.Object().GetCreationTimestamp().Time)
```
because it's clearer what you are subtracting from what.

pkg/controller/jobframework/reconciler.go
Nilsachy: Done.

pkg/controller/jobframework/reconciler.go
Nilsachy: Good point! Done.

pkg/controller/jobframework/reconciler.go
```diff
@@ -1588,6 +1588,18 @@ func (r *JobReconciler) handleJobWithNoWorkload(ctx context.Context, job Generic
...
+	jobCreationTime := job.Object().GetCreationTimestamp().Time
+	if !jobCreationTime.IsZero() {
+		wlCreationTime := wl.CreationTimestamp.Time
+		if wlCreationTime.IsZero() {
+			wlCreationTime = r.clock.Now()
+		}
+		latency := wlCreationTime.Sub(jobCreationTime)
+		customLabelValues := r.customLabels.LQGet(utilqueue.KeyFromWorkload(wl))
+		metrics.ReportJobToWorkloadLatency(job.GVK().Kind, latency, customLabelValues, r.RoleTracker())
+	}
```
kshalot: Is there actually a case where the `CreationTimestamp` is zero? By this point both objects should be created and `r.client.Create` should update the newly created `wl` object as well ([code](https://github.com/kubernetes-sigs/controller-runtime/blob/6210f847b2c1df3f28e5be34a4b1458f03896c73/pkg/client/interfaces.go#L67-L69)).

So neither should be zero. If you want to be super safe this handling would be better IMO:
```suggestion
	jobCreationTime := job.Object().GetCreationTimestamp().Time
	wlCreationTime := wl.CreationTimestamp.Time
	if !jobCreationTime.IsZero() && !wlCreationTime.IsZero() {
		latency := wlCreationTime.Sub(jobCreationTime)
		customLabelValues := r.customLabels.LQGet(utilqueue.KeyFromWorkload(wl))
		metrics.ReportJobToWorkloadLatency(job.GVK().Kind, latency, customLabelValues, r.RoleTracker())
	}
	// Potentially log that one of the timestamps was zero? But it seems redundant.
```

Or we could ditch the zero handling entirely and just assume that the timestamps are not zero.

kshalot: Also, maybe this could be extracted to a helper function for readability, like `reportJobToWorkloadLatency`

pkg/metrics/metrics.go
```diff
@@ -187,6 +187,10 @@ var (
 	// +metricsdoc:labels=preempting_cluster_queue="the ClusterQueue executing preemption",reason="eviction or preemption reason",replica_role="one of `leader`, `follower`, or `standalone`"
 	PreemptedWorkloadsTotal *prometheus.CounterVec
 
+	// +metricsdoc:group=job
+	// +metricsdoc:labels=job_kind="the kind of the job",replica_role="one of `leader`, `follower`, or `standalone`"
```
kshalot: Please run `make generate` to populate the metrics tables in the docs:
https://github.com/kubernetes-sigs/kueue/blob/9494dcc26a3907a3ec2efd9f2dcdca0b2e0d2720/site/content/en/docs/reference/metrics.md?plain=1#L17

If this doesn't update it, then you likely have to add the corresponding section and:
```
<!-- BEGIN GENERATED TABLE: job -->
<!-- END GENERATED TABLE: job -->
```
manually before running `make generate`, I'm not sure what the process is.

pkg/metrics/metrics.go
```diff
@@ -509,6 +513,15 @@ The label 'underlying_cause' can have the following values:
 		}, append([]string{"name", "namespace", "priority_class", "replica_role"}, extraLabels...),
 	)
 
+	JobToWorkloadLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: constants.KueueName,
+			Name:      "job_to_workload_latency_seconds",
```
kshalot, line 554: This time is probably much lower than a second - why use seconds? Won't you just bucket everything into the first bucket - "1"?

kshalot: By looking at the generated buckets:
```go
func generateExponentialBuckets(count int) []float64 {
	return append([]float64{1}, prometheus.ExponentialBuckets(2.5, 2, count-1)...)
}
```
The first bucket would represent 1 second, so everything would fall there, no?

kshalot: From what I can see, the analogous admission attempt latency ("admission_attempt_duration_seconds") uses the default buckets (does not set them explicitly):
```go
// DefBuckets are the default Histogram buckets. The default buckets are
// tailored to broadly measure the response time (in seconds) of a network
// service. Most likely, however, you will be required to define buckets
// customized to your use case.
var DefBuckets = []float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10}
```

kshalot: So just to be clear here - using "seconds" is fine, since it's a float64 and we already have precedence in the form of `admission_attempt_duration_seconds`. But the buckets have to be adjusted from what I can see.

pkg/metrics/metrics.go
kshalot: I'm also wondering whether this shouldn't be grouped under "health" instead of a new group.

We already have a "admission_attempt_duration_seconds" metric defined under "health".

pkg/metrics/metrics.go
Nilsachy: I see, thanks for catching that. IIUC, we should be seeing a heavy right-skewed distribution as most of the Workloads are created almost instantly, correct? In this case, would `DefBuckets` be good enough? Or should we introduce more granularity?

pkg/controller/jobframework/reconciler.go
Nilsachy: Removed zero check and moved to helper func.

pkg/metrics/metrics.go
Nilsachy: Done.

pkg/metrics/metrics.go
kshalot: Good question. I'm not sure how useful a bucket like `.005` would be, my intuition says that it's too low - event propagation, cache updating, workload construction etc. will likely always be over 5ms. I guess it can be there as a safety margin, doesn't really hurt. Since we are already using them for that metric, we could use them here as well.

OTOH we probably want to get the most resolution in the "critical zone", which is probably around 100ms - 1000ms? I think the goal of this metric is to catch certain controllers underperforming. Having more buckets in that zone would potentially help catch some more linear increase in latency.

Nilsachy: I see, are you then imagining to keep the defaults but expand the buckets in the range 100ms - 1000ms? Something like this?

```
Buckets: []float64{
                .005, .01, .025, .05,
                .1, .2, .3, .4, .5, .6, .7, .8, .9, 1.0,
                2.5, 5.0, 10.0,
            },
```

Nilsachy: I am just afraid that having too many entries in the bucket list increases the cardinality and might lead to scraping overhead. Isn't that a concern to keep in mind?

Nilsachy: The prometheus default is 11 buckets, while this for example would be 17.

kshalot: Hmmm, the cardinality is pretty low in this metric, no? It's limited by the number of job kinds we support.

Maybe we don't need all those buckets, though. For example, we could have `[..., .5, .75, 1.0, ...]` instead of linear.

Nilsachy: I have updated it to `[]float64{0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 0.75, 1, 1.5, 2, 3, 5, 10}`. Let me know if that makes sense.

kshalot: You can also consider `prometheus.ExponentialBuckets(0.01, 2, 11)` instead of hard-coding. It would also generate a sensible range (..., 80ms, 160ms, 320ms, 640ms, ...). But it also seems very close to the `DefBuckets` which I previously criticized. Sorry for the lack of strong opinion on this :P I think your proposed range is fine as well (expect maybe the 0.005 bucket which seems very low as I mentioned previously).

But let's keep this thread open, maybe the approvers will have a stronger opinion.

pkg/controller/jobframework/reconciler_test.go
```diff
@@ -1236,3 +1238,57 @@ func TestReconcileGenericJobWithWaitForPodsReady(t *testing.T) {
...
+	rec := NewReconciler(cl, recorder)
+
+	_, err := rec.ReconcileGenericJob(ctx, controllerruntime.Request{
+		NamespacedName: types.NamespacedName{Name: testJobName, Namespace: metav1.NamespaceDefault},
+	}, mgj)
+	if err != nil {
+		t.Fatalf("Failed to Reconcile GenericJob: %v", err)
+	}
+
+	all := testingmetrics.CollectFilteredGaugeVec(metrics.JobToWorkloadLatency, map[string]string{"job_kind": uniqueJobKind})
```
kshalot: Even if this somehow works, the fact that we are using a function called "CollectFiltered**Gauge**Vec" on a histogram sounds very suspicious. Maybe a new function could be useful here, one that more explicitly captures the intent for a histogram.

pkg/metrics/metrics.go
Nilsachy: I have moved this to `prometheus.ExponentialBuckets(0.01, 2, 11)`, will leave it opened for other reviewers to give their opinion.

pkg/controller/jobframework/reconciler_test.go
Nilsachy: I agree it is very misleading, but the function actually handles several cases already:

```
if dtoMetric.Gauge != nil {
   dp.Value = *dtoMetric.Gauge.Value
}
if dtoMetric.Counter != nil {
   dp.Value = *dtoMetric.Counter.Value
}
...
```

I believe the function itself should be renamed to something more generic

Nilsachy: We could rename it for example: `func CollectFilteredDataPoints(...)`

pkg/metrics/metrics.go
```diff
@@ -187,6 +187,10 @@ var (
 	// +metricsdoc:labels=preempting_cluster_queue="the ClusterQueue executing preemption",reason="eviction or preemption reason",replica_role="one of `leader`, `follower`, or `standalone`"
 	PreemptedWorkloadsTotal *prometheus.CounterVec
 
+	// +metricsdoc:group=health
+	// +metricsdoc:labels=job_kind="the kind of the job",replica_role="one of `leader`, `follower`, or `standalone`"
+	JobToWorkloadLatency *prometheus.HistogramVec
```
kshalot, line 195: This is now nested under "// Metrics tied to the queue system." (line 80). It should probably be it's own thing, like "// Metrics tied to the job framework".

pkg/controller/jobframework/reconciler_test.go
kshalot: My intuition says that it's different for histogram, because it's multi dimensional. Gauges and counters are pretty much a single number, but a histogram data point has 3 dimensions I think. 

Can't we use something that exists already, like `GetHistogramMetricCount`? Won't that fit better here anyway? You are currently taking the `SampleSum`.

pkg/metrics/metrics.go
Nilsachy: Done.

pkg/controller/jobframework/reconciler_test.go
Nilsachy: Right that's a good point, we can just use `GetHistogramMetricCount` here and it's probably better, thanks for the catch!

kshalot:
/lgtm

💪 I think all my comments were addressed, thanks for that. I'll add LGTM to signal that someone with approval powers can start looking at this.

Nilsachy:
@mimowo Could you PTAL at this PR when you have a moment

mimowo (commented):
lgtm, but the PR requires rebase

pkg/controller/jobframework/reconciler_test.go
```diff
@@ -1033,3 +1036,61 @@ func TestReconcileGenericJobWithWaitForPodsReady(t *testing.T) {
 		})
 	}
 }
+
+func TestReconcileJobToWorkloadLatencyMetric(t *testing.T) {
```
mimowo: Let's convert this test  into table-driven with two test cases so that it is easier to add future test cases. 
I think we could check different buckets, by different latency, or replica-role: leader/ follower.

Nilsachy: I have parameterized the test with two different test cases (latency of 2s and 10s). Was that what you had in mind?

pkg/controller/jobframework/reconciler.go
```diff
@@ -1598,9 +1598,20 @@ func (r *JobReconciler) handleJobWithNoWorkload(ctx context.Context, job Generic
...
+	r.reportJobToWorkloadLatency(job, wl)
+
 	return nil
 }
 
+func (r *JobReconciler) reportJobToWorkloadLatency(job GenericJob, wl *kueue.Workload) {
+	jobCreationTime := job.Object().GetCreationTimestamp().Time
+	wlCreationTime := wl.CreationTimestamp.Time
+	latency := wlCreationTime.Sub(jobCreationTime)
+	customLabelValues := r.customLabels.LQGet(utilqueue.KeyFromWorkload(wl))
```
mimowo: This is behind Alpha FG for custom labels

pkg/controller/jobframework/reconciler_test.go
mimowo: This looks ok, but I'm not sure if I understand the asserts, because the assert is on exact latency, but the histogram is bucketing the results. So I would exect the assert "wantCountsForBuckets", or wantNonEmptyBuckets. I'm not sure how to interpret `wantLatency: 10.0`, is the the left boundary of the non-empty bucket, or something else? This needs at least clarifying.

pkg/metrics/metrics.go
```diff
@@ -192,6 +192,12 @@ var (
 	// +metricsdoc:labels=cluster_queue="the evicted workload's ClusterQueue from status.admission on the workload before quota was released (only present when the metric records a sample)",reason="eviction or preemption reason (same values as evicted_workloads_total)",replica_role="one of `leader`, `follower`, or `standalone`"
 	WorkloadEvictionLatencySeconds *prometheus.HistogramVec
 
+	// Metrics tied to the job framework
+
+	// +metricsdoc:group=health
+	// +metricsdoc:labels=job_kind="the kind of the job",replica_role="one of `leader`, `follower`, or `standalone`"
+	JobToWorkloadLatency *prometheus.HistogramVec
```
mimowo: Wdyt about `WorkloadCreationLatency` as the alternative name of the metric? That seems to capture the intention more directly.

 cc @kshalot @olekzabl

pkg/controller/jobframework/reconciler.go
```diff
@@ -1598,9 +1598,20 @@ func (r *JobReconciler) handleJobWithNoWorkload(ctx context.Context, job Generic
 	}
 	r.record.Eventf(object, corev1.EventTypeNormal, ReasonCreatedWorkload,
 		"Created Workload: %v", workload.Key(wl))
+
+	r.reportJobToWorkloadLatency(job, wl)
```
mimowo: I think this place in code only covers the Jobs represented by the GenericJob interface, but we also create workloads for StatefulSets and LEaderWorkerSet workloads which are not managed this way. So, I think to make the metric complete we also need to cover the other workload creation places.

pkg/controller/jobframework/reconciler_test.go
Nilsachy: The thing is I tried to write a test that checks the logic of `reportJobToWorkloadLatency` and for that I want to make sure that the metric is exactly recorded as it should be. If we assert on buckets, there is a loss of precision because the buckets are created using `prometheus.ExponentialBuckets(0.01, 2, 11)`. So let's say the `reportJobToWorkloadLatency` doesn't work properly and it records 1.5 seconds instead of 2, the test will not fail because it still falls under the `1.28 < x < 2.56` bucket. Does that make sense?

That's why I only assert for a single sample (count == 1) and `GetHistogramMetricValue` is simply returning my single value as it uses `GetSampleSum()` under the hood.

mimowo: Ah, I see, so despite the fact that the prometheus API exposes the bucketed metric the code-level API still provides a helper to get the underlying "sampleSum", and in this case the same is just a single element, so we get its value. 

Now I understand how it works.

 In that case I'm ok, but I would rename the param to `wantSampleSum` to make this direct. I don't think we need to check the count for the bucket is 1. In fact it might be interesting to have multiple jobs in the test, but we can leave that for later if needed.

Nilsachy: Good point, the assert on the count is probably redundant. Will update this and rename to `wantSampleSum`, thanks!

Nilsachy: Done.

pkg/controller/jobframework/reconciler.go
Nilsachy: Mmmh it would seem that [customLabels.LQGet](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/metrics/custom_labels.go#L176) handles that gracefully. It returns `nil` if cl is `nil`. In `ReportJobToWorkloadLatency` we receives it as a slice (which can be `nil` as well).

pkg/controller/jobframework/reconciler.go
Nilsachy: That's right thank you for noticing that! I will move `ReportJobToWorkloadLatency` to a generic function which can be reused in other reconcilers.

pkg/metrics/metrics.go
Nilsachy: I renamed the metric and functions to `WorkloadCreationLatency` and `RecordWorkloadCreationLatency`.

pkg/controller/jobframework/reconciler.go
Nilsachy: I moved the logic to `pkg/controller/jobframework/utils.go`. The test is now parameterized per job types (GenericJob, StatefulSet, and LWS). I have also simplified the test as `RecordWorkloadCreationLatency` is a straightforward helper that accepts concrete Kubernetes objects (client.Object, *kueue.Workload). Lmk if the test coverage is enough.

pkg/controller/jobframework/utils_test.go
```diff
@@ -95,3 +101,51 @@ func TestSanitizePodSets(t *testing.T) {
...
+		},
+	}
+
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			metrics.WorkloadCreationLatency.Reset()
+
+			baseTime := time.Now().Truncate(time.Second)
+			jobCreationTime := baseTime.Add(-5 * time.Second)
+			job := testingjob.MakeJob(testJobName, metav1.NamespaceDefault).
```
Nilsachy: I made a mistake as this doesn't actually test different kind of jobs (it passes a generic job for every test case). I need to update this.

Nilsachy: Done.

pkg/metrics/metrics.go
```diff
@@ -192,6 +192,12 @@ var (
 	// +metricsdoc:labels=cluster_queue="the evicted workload's ClusterQueue from status.admission on the workload before quota was released (only present when the metric records a sample)",reason="eviction or preemption reason (same values as evicted_workloads_total)",replica_role="one of `leader`, `follower`, or `standalone`"
 	WorkloadEvictionLatencySeconds *prometheus.HistogramVec
 
+	// Metrics tied to the job framework
+
+	// +metricsdoc:group=health
+	// +metricsdoc:labels=job_kind="the kind of the job",replica_role="one of `leader`, `follower`, or `standalone`"
+	WorkloadCreationLatency *prometheus.HistogramVec
```
mimowo, line 199: Hmmm, I'm still not sure about this metric. It seems it will have undesired results in some scenarios:
1. use of ElasticJobsViaWorkloadSlices - this feature assumes at its core creation of new "replacement" Workloads when the Job is resized
2. LWS workloads, here we also create new workloads as the LWS is resized, because we create a new workload per LWS group

So, we need to fix the issues which increases the effort/gain from the metric. 

To fix (1.) we could probably just exclude "replacement" workloads and only measure for the "first" workloads. 
For (2.) we should probably just measure the time to create the workload for 0 replicas? I'm not sure how to solve it well actually.

So I think we need to rethink how useful the metric really is expected to be. 

cc @dkaluza

Nilsachy: > use of ElasticJobsViaWorkloadSlices - this feature assumes at its core creation of new "replacement" Workloads when the Job is resized

For this use case, wouldn't it make sense to still report the metric but instead of measuring from CreationTimestamp we use the timestamp at the time of resize? Not sure yet how we get this information tho (if at all possible).

mimowo: Yeah, conceptually probably yes, but that is the challenge - we don't really have the time of resize readily available.

dkaluza: From my point of view if we create many workloads for LWS and ElasticJobsViaWorkloadSlices , measurement of time from resize to the creation of new workload sounds as conceptually correct solution.

Here probably [lastTransitionTime](https://github.com/kubernetes-sigs/lws/blob/main/config/crd/bases/leaderworkerset.x-k8s.io_leaderworkersets.yaml#L17823) on appropriate condition could be used, but I'm not even sure if those condidtions exist for for all of the needed types.

Edit: it doesn't look like this in k8 batch job api [source](https://github.com/kubernetes/kubernetes/blob/a11a1367fb1a1434bcc6c2d731b8b90722953607/api/openapi-spec/v3/apis__batch__v1_openapi.json#L268).

mimowo: Yeah that is why I'm thinking maybe we could exclude resize based workloads, just measure time to first workload to avoid the issue with defining the metric.

mimowo: There is also a case when a user manually deleted a workload which may be messing with the stats

Nilsachy: Just a thought, could we use the time when the reconciler detects that a new workload is needed? That's an approximation but IIUC this metric is mainly useful to detect the workload creation time, regardless of the job creation time (especially considering the edge cases you mentioned). So if we stay consistent, this can still provide value, wdyt?

dkaluza: Correct, though I'm not sure we can detect deleted cases without introduction of some additional state like registering seen jobs or annotating the resources that we already created workload for them. 
And if we need to save additional state to cover those I'm actually not sure it is worth it.
Unless there is some annotation/label that we can reuse?

What if we:
1. Exclude LWS and ElasticWorkloads, and explicitly describe that those are for now not covered by this metric since they create many workloads.
2. Live for now with the fact that users can delete a workload (I think this is not a highly probable situation) and mess with the stats.

Unless we have some reliable way to track that this will be first metric emission for LWS/Elastic.

Annotation that may be added to job on metric emission comes into my mind, but this will trigger the watches, which might have a negative impact for overall performance.

Nilsachy: I'm still thinking that even in the typical case, we don't necessarily care at what moment the job was created. The metric is rather trying to capture the workload creation time. Isn't the timestamp at which the reconciler decides we need the workload be enough then? This way we don't really need to handle any of the edge cases like you mentioned above

dkaluza: >I'm still thinking that even in the typical case, we don't necessarily care at what moment the job was created. The metric is rather trying to capture the workload creation time. Isn't the timestamp at which the reconciler decides we need the workload be enough then? This way we don't really need to handle any of the edge cases like you mentioned above

Not exactly, this metric is trying to capture the latency, between job creation and workload creation. In this way we can spot performance degradation, when Kueue cannot catch up with new workloads demand.

Nilsachy: Of course, but my point is that if we deconstruct the latency in two phases:

1. Job creation -> Reconciler detection
2. Reconciler detection -> workload creation

IIUC it is in that second phase that the latency might be high. Especially when the degradation is related to high number of requested workloads, isn't it?

Nilsachy: And if we realize that phase 1 (from job creation to reconciler detection) is also important to record, we can always have that as a separate metric. This way the separation between job creation and job resizing is handled out of the box. 

This is just to introduce a different perspective to the problem, I don't think we necessarily need the full pipeline (phase 1 and 2) together to evaluate kueue's health

Edit : with a second look, not sure if phase 1 can be implemented in its own metric without also differentiating between an initial creation and a replacement

dkaluza: This is a little more complex AFAIK. 
Generally controller runtime keeps items to reconcile in a workqueue. Then they are taken by an appropriate controller and reconciled according to permitted concurrency. 

So the items may stay for some time in this queue, and we are generally interested in that. Because if they are tending to stay longer and longer, we will have a performance issue.
There are metrics for that in the controller-runtime that we can utilize instead - https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/internal/metrics/workqueue.go#L69

Generally the latency from the worqueue is linked to the number of items in the workqueue, concurrency and processing time. But it is also tracked by appropriate metric in controller-runtime so there is no need to compute it. (As well as workDuration)

So those metrics are in a way already tracking this in parts.
However, metric that shows workload creation timestamp - job timestamp is more aligned with our business perspective in my opinion - it represents better how much time user needs to wait for a workload to appear.
The predefined metrics also do not track in any way the time from job creation, they rather track "when job was added to the workqueue", so they will not spot the problem that controller is actually receiving items with huge delay , for example because of networking issues.

> Edit : with a second look, not sure if phase 1 can be implemented in its own metric without also differentiating between an initial creation and a replacement

Yes, generally those metrics do not take this into the account, neither  the proposed division.

Having that said, if we cannot easily implement the metric that makes sense from our business perspective  they can be used as a substitutions for performance tracking. However, as mentioned I rather prefer `WorkloadCreationLatency` if it is achievable (even with some exclusions, we should anyway also track the workqueue to detect other type of performance issues).

mimowo: Myabe a sensible and simple alternative is WorkloadCreationLatency that would only account for Workloads created while the owner generation is 1? This woukd ofc have a downside of not counting  workloads for resizes, or if the owner is mutated very quickly after crewation, but such cases seem rare, and we would already get a relevant signal. Or alternatively have a label with owner generation. Then users coukd filter the metric by generation 1. Brainstorming.

mimowo: Or another alternative: only count the metrics for workloads when the owner is "new", say its first 5min since creation.

dkaluza: Generation 1 might work, I would have to look into the details if mutations in webhooks count into this
(I would guess no, but I'm not sure).

Or maybe some combined heuristic? Require owner generation 1 in case of LWS and ElasticJobsViaWorkloadSlices as those have multiple iterations but do not require generation 1 for the rest as those shouldn't be recreated in normal circumstances? 

>  Or alternatively have a label with owner generation

You have the metric label in mind, yes? Not the job label? This will increase the cardinality significantly (though we should be fine if no new labels are added in the future...).

mimowo: > Or maybe some combined heuristic? Require owner generation 1 in case of LWS and ElasticJobsViaWorkloadSlices as those have multiple iterations but do not require generation 1 for the rest as those shouldn't be recreated in normal circumstances?

But this still leaves the case for manual workload deletion / recration which could mess up the stats.

> You have the metric label in mind, yes? Not the job label? This will increase the cardinality significantly (though we should be fine if no new labels are added in the future...).

Yes, metric label. Yeah, cardinality would be bad for the label, unless we could bucket by "generation=new/not_new".

dkaluza: > But this still leaves the case for manual workload deletion / recration which could mess up the stats.

True. However, I'm still not sure there is any real use case for that. If this happens I would be curious to hear from the users why are deleting them. But, I agree that it would be best to have a metric that can filter those out.

> Yes, metric label. Yeah, cardinality would be bad for the label, unless we could bucket by "generation=new/not_new".

What if we go one step further, bucket the generation number to new/not_new and add the workload source type label? Cardinality will still be manageable ~50 and we will be able to use the combination to select of filter out the metric? In this way it will be up to the metric user if they want just the first one for non LWS/Elastic workloads or all (And in many situations they have the knowledge if they deleted the workloads, how many and when...).  WDYT?

mimowo: > True. However, I'm still not sure there is any real use case for that. If this happens I would be curious to hear from the users why are deleting them. But, I agree that it would be best to have a metric that can filter those out.

Actually, this sometimes happens - it is the easiest way for a user to manually suspend a job, so that it goes via the re-admission cycle, on-demand. 

> What if we go one step further, bucket the generation number to new/not_new and add the workload source type label? Cardinality will still be manageable ~50 and we will be able to use the combination to select of filter out the metric? In this way it will be up to the metric user if they want just the first one for non LWS/Elastic workloads or all (And in many situations they have the knowledge if they deleted the workloads, how many and when...). WDYT?

Maybe but this again feels like "day two" metric when the "day one" is not sufficient. The day one go-to metric for me would be to just check the time for creating workloads when the generation of the owner is 1. Unless we can find some counter examples when this is skewed severely. Yes, it can mean that we will not report some workloads if the owner is mutated immediately after creation, but that feels like rare, and even if this happens it should not skew to much the overall "latency" picture. OTOH, including workloads created due to deletion, or re-creation may skew the results severely ( at least this is my intuition).

However, I just realized another corner cases: what if the user deletes / recreates workload while the owner generation is still 1. This is so unlikely that we may probably not care about it.

I'm still wondering if instead of looking at the generation we could somehow tell "FirstWorkloadCreationLatency". This could probably be possible if we stored the information in cache that for a given owner a workload already existed.

dkaluza: > Maybe but this again feels like "day two" metric when the "day one" is not sufficient.... 

I agree, the additional label in my mind were just for additional investigation purposes. If you think that it will be rarely used maybe it is not worth the additional storage costs on the metric sink side. On the other hand users can easily  drop the labels with [relabel-config](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config).

> However, I just realized another corner cases: what if the user deletes / recreates workload while the owner generation is still 1. This is so unlikely that we may probably not care about it.

Yea, I wouldn't bother about it right now, this sounds like highly unlikely scenario, so it will not influence p95 normally. And again in many situations users know actions that they performed so they will be able to interpret such outliers. 

> I'm still wondering if instead of looking at the generation we could somehow tell "FirstWorkloadCreationLatency". This could probably be possible if we stored the information in cache that for a given owner a workload already existed.

I tried to touch on this in one of the first comments "though I'm not sure we can detect deleted cases without introduction of some additional state like registering seen jobs ".
Personally I'm not convinced it is worth the effort. Maintaining additional state of size O(workloads) just for the purpose of slightly better metric reporting sounds like an overkill.

Nilsachy: Food for thought: it feels to me like we are trying to have one metric to capture a plural mechanism. Since workload creation is not necessarily related to job creation (although job creation is directly related to workload creation), we don't have this double implication, so can't we have two separate metrics for this?

> However, metric that shows workload creation timestamp - job timestamp is more aligned with our business perspective in my opinion 

I agree with this point. I'm just wondering if this has to be one-metric-captures-it-all, or if we can benefit from a separation of concerns. I actually see benefits in splitting the metrics into:

Metric 1: job creation -> reconciler detection 
Metric 2: reconciler detection -> workload creation

This way, the user gets a more fine grained explanation for performance degradation. Metric 1 would inform about latency in the workqueue, while metric 2 would inform about latency in the controller. 

Metric 2 (reconciler detection -> workload creation) is very easy to implement. Only minor changes to the current implementation.

> The day one go-to metric for me would be to just check the time for creating workloads when the generation of the owner is 1. Unless we can find some counter examples when this is skewed severely. Yes, it can mean that we will not report some workloads if the owner is mutated immediately after creation, but that feels like rare, and even if this happens it should not skew to much the overall "latency" picture. OTOH, including workloads created due to deletion, or re-creation may skew the results severely ( at least this is my intuition).

We can use this idea of owner generation of 1 for the first metric (job creation -> reconciler detection), but we still need to solve the problem with workloads created due to deletion.

I'm probably missing something more here but it feels to me like two metrics is still a better approach, even from a user feedback perspective. Wdyt?

mimowo: I think the users are not interested in the distinction job creation -> reconciler detection. Or reconcler detection -> workload creation. 

Maybe such metrics could be interesting to some hardcore admins, but:
1. the first metric -> time from update to reconcile seems like could already exposed in a generic way by the controller-runtime logs (I'm not sure, but I know that controller-runtime exposes some metrics on reconcilers). 
2. the second metric could already be approximated very closely the api-server latency metrics for requests, because from reconcile to update 99% of the time will be the request itself.

dkaluza: @Nilsachy 
As mentioned from my point of view partial metrics are already covered by the controller runtime built-in metrics. There is no need to repeat those, users can access  them just fine AFAIK. Here it also depends what do you mean by "reconciler detection".

[workDuration](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/internal/metrics/workqueue.go#L69) corresponds to "how long" does the reconciliation take, sure it does not align perfectly to metric 2 mentioned by you - because some reconciliations do not end up with workload creation, but from the performance tracking perspective we also would like to track those. 

[latency](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/internal/metrics/workqueue.go#L59) is an approximation of mentioned Metric 1. If we treat by detection when item is added to workqueue, then we might not be able to implement this at all as we are not in control of this part of code. This queuing is implemented in controller-runtime library, and there might be no entrypoint for additional metric emission there.

If you had in mind start of "reconciliation", then sure one can try to implement this, but here the problems are even larger from my point of view. What if this reconciliation doesn't result in workload creation at all? Because for example workload already exists should you report it or not? What if workload is not yet created because of some other reason and is requeued (not sure if this can be a real case, would have to investigate further)...
Generally from my perspective meaning of such metric would be a somehow unclear, we would have to define it more strictly to make it understandable.

Nilsachy: I see, thanks for the explanation. Is the consensus then to record the metric only for jobs with owner generation of 1? And we accept to have some skewed results if the user deletes the workload?

mimowo: before we conclude that, could you prototype that and see if this approach works well in practice? I'm not sure, maybe there is something about generations that could still surprise us.

Nilsachy: Sure!

pkg/controller/jobframework/utils.go
```diff
@@ -74,3 +78,15 @@ func sanitizeContainer(container *corev1.Container) {
 		container.Env = append(container.Env, envVars[len(envVars)-1])
 	}
 }
+
+// RecordWorkloadCreationLatency records the latency between job creation and workload creation.
+func RecordWorkloadCreationLatency(job client.Object, jobKind string, wl *kueue.Workload, customLabels *metrics.CustomLabels, tracker *roletracker.RoleTracker) {
+	if job.GetGeneration() > 1 {
```
Nilsachy, line 89: @mimowo @dkaluza am I missing something else here?

Nilsachy: What about when the mutating webhook adds `spec.suspend: true` to the original job? Wouldn't that increment the Job's Generation to 2?

dkaluza: I tested this locally, mutation webhooks do not change the generation number during creation. IIUC this is the case because they intercept the request before the object is created/written to etcd for the first time as they are run before validation webhooks.

Looks fine to me

pkg/metrics/metrics.go
```diff
@@ -546,6 +552,15 @@ The label 'underlying_cause' can have the following values:
 		}, append([]string{"name", "namespace", "priority_class", "replica_role"}, extraLabels...),
 	)
 
+	WorkloadCreationLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: constants.KueueName,
+			Name:      "workload_creation_latency_seconds",
+			Help:      "The time between a job was created until its workload was created, per 'job_kind'",
```
dkaluza: Can we add information about the limitation to help that entries are only recorded for objects with generation 1.

pkg/controller/jobframework/utils_test.go
```diff
@@ -87,11 +97,83 @@ func TestSanitizePodSets(t *testing.T) {
...
+			}
+
+			job.SetGeneration(1)
+			jobframework.RecordWorkloadCreationLatency(job, tc.jobKind, wl, nil, nil)
+
+			val, err := testutil.GetHistogramMetricValue(metrics.WorkloadCreationLatency.WithLabelValues(tc.jobKind, roletracker.RoleStandalone))
+			if err != nil {
+				t.Fatalf("Failed to get histogram metric value: %v", err)
+			}
+			if val != 5.0 {
```
mimowo: Do not check the magical value inside the test code, instead move the expected value to the test cases. This will also require naming the paramter so easier to understand.

pkg/controller/jobframework/utils.go
```diff
@@ -74,3 +78,15 @@ func sanitizeContainer(container *corev1.Container) {
...
+		return
```
mimowo, line 91: Let's log that information at v(4), say skip recording the workload creation metrics as the owner generation is already xyz, or something along the lines.

pkg/controller/jobframework/utils.go
Nilsachy: Resolving.

pkg/controller/jobframework/utils_test.go
Nilsachy: Done.

pkg/controller/jobframework/utils.go
Nilsachy: Done.

pkg/metrics/metrics.go
Nilsachy: Done.

mimowo (commented):
Given the uncertinity about the final shape of the metric I would like to introduce a feature gate, can be Beta in 0.18, to indicate the metric may still evolve. Let's call it MetricForWorkloadCreationLatency. Please add a comment to GA the metric in 0.21, we should have enough feedback on the metric by then.

Nilsachy:
> Given the uncertinity about the final shape of the metric I would like to introduce a feature gate, can be Beta in 0.18, to indicate the metric may still evolve. Let's call it MetricForWorkloadCreationLatency. Please add a comment to GA the metric in 0.21, we should have enough feedback on the metric by then.

Makes sense, I have added the flag (enabled by default) with comment to GA in 0.21.

mimowo (commented):
/lgtm
/approve
Thank you 👍

mimowo:
looks like unrelated flake, opened: https://github.com/kubernetes-sigs/kueue/issues/11115

tenzen-y:
/release-note-edit

```release-note
Observability: Added a new Prometheus metric `workload_creation_latency_seconds` to track the time elapsed between a Job's creation and the creation of its corresponding Workload by Kueue.
```

PR #10358, opened 2026-04-09
[Cleanup] Set policy rules based on the enabled integration

mszadkow:
#### What type of PR is this?
/kind cleanup
/area testing

Unify the approach to set RBAC rules for test based on the enabled integrations.

Relates #10200

```release-note
NONE
```


mimowo (commented):
I'm wondering if this is it just to avoid deploying RBAC which is not necessary, or it realistically improves performance or fixes running some tests?

mszadkow:
> I'm wondering if this is it just to avoid deploying RBAC which is not necessary, or it realistically improves performance or fixes running some tests?

The later, I needed another set of PolicyRules for different test, which differs just by one integration.

mimowo:
I see ok, we can CP that
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16

Issue #10269, opened 2026-04-01
Large heap retained by zap logger clones in reconcile loops (19K-pod cluster)


ikchifo:
**What happened**:

On a large cluster (19K pods, 1571 nodes, 277 Kueue Workloads), the kueue-controller-manager holds ~190 MB of heap from per-reconcile zap logger allocations. Top inuse allocators from `go tool pprof`:

| inuse_space | Function |
|-------------|----------|
| 60.56 MB | `go.uber.org/zap/internal/bufferpool.init.NewPool.func1` |
| 44.54 MB | `go.uber.org/zap/zapcore.(*jsonEncoder).clone` |
| 8.50 MB | `go.uber.org/zap.(*Logger).clone` |
| 77.06 MB cum | `github.com/go-logr/zapr.(*zapLogger).WithValues` |

Each reconcile calls `WithValues` which clones the logger, core, and encoder, drawing a new buffer from `bufferpool`. Under sustained throughput (~thousands of reconciles/sec across all controllers), the sync.Pool refills faster than GC drains it. The per-reconcile `reconcileID` prevents caching.

The call chain per reconcile:

```
reconcileHandler -> NewLogConstructor.func1
  -> zapr.WithValues -> zap.Logger.With -> ioCore.With
    -> jsonEncoder.Clone -> bufferpool.Get  (x2 per reconcile)
```

The default `LogConstructor` chains two `WithValues` calls (controller name, then namespace/name/reconcileID), producing two full clone-and-encode cycles per reconcile.

**What you expected to happen**:

Per-reconcile logger allocations should not dominate the heap profile (~190 MB at this scale).

**How to reproduce it (as minimally and precisely as possible)**:

On a cluster with >= 15K pods and Kueue managing >= 200 workloads:

1. Enable pprof (`pprofBindAddress: :8083` in the ConfigMap)
2. At steady state, capture the heap profile:
   ```
   kubectl port-forward -n kueue-system pod/<kueue-pod> 8083:8083
   go tool pprof -inuse_space http://localhost:8083/debug/pprof/heap
   ```
3. Run `top 20`. The top allocators are `bufferpool`, `jsonEncoder.clone`, and `zapr.WithValues`.

**Proposed fix: cache the base logger per controller**

Override `LogConstructor` to cache the controller-level logger at setup and call `WithValues` once per reconcile instead of twice:

```go
controllerLogger := baseLogger.WithValues("controller", name) // once at setup
[...]

PR #10365, opened 2026-04-10
fix: optimize NewLogConstructor logging

MatteoFari:
#### What type of PR is this?
/kind bug

Improves roletracker.NewLogConstructor by creating the named controller logger once and reusing it for later reconcile calls.

It's still safe because replica-role is still updated on each call, namespace and name are still added for each reconcile request.

This is the first part of the work discussed in #10269 

Partially addresses #10269

```release-note
Observability: Fix excessive memory overhead in hot code paths by reusing the named logger in NewLogConstructor and avoiding unnecessary logger cloning.
```

### DIFF
--- a/pkg/util/roletracker/logger_test.go
+++ b/pkg/util/roletracker/logger_test.go
@@ -17,10 +17,15 @@ limitations under the License.
 package roletracker
 
 import (
+	"encoding/json"
 	"testing"
 
 	"github.com/go-logr/logr"
 	"github.com/go-logr/logr/funcr"
+	"github.com/google/go-cmp/cmp"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 func TestWithReplicaRole(t *testing.T) {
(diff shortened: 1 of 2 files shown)


mimowo:
/ok-to-test
Thanks 👍

pkg/util/roletracker/logger_test.go
```diff
@@ -53,58 +57,142 @@ func TestWithReplicaRole(t *testing.T) {
...
-			expected: RoleLeader,
-		},
-		{
-			name:     "follower tracker",
-			tracker:  NewFakeRoleTracker(RoleFollower),
-			expected: RoleFollower,
-		},
-	}
+	capture, logger := newJSONLoggerCapture()
+	setControllerLogger(t, logger)
```
mimowo: I would prefer to keep this test as table driven, this allows to avoid setup duplication, and calling `capture.reset()`. Then the helper function I think we could inline. 

IIUC this is done mostly for this test case:
```golang
		constructor(nil).Info("before role transition")
		tracker.role.Store(RoleLeader)
		constructor(nil).Info("after role transition")
```
I think for that test we may just generalize the function used for testing, by having a param like `action function(roleTracker, constructor)` which would encapsulate that, wdyt?

MatteoFari: Thanks for the feedback. I agree.

So the test is table-driven and added `action` function for the role-change case.

pkg/util/roletracker/logger_test.go
```diff
@@ -53,58 +57,157 @@ func TestWithReplicaRole(t *testing.T) {
...
 
 func TestNewLogConstructor(t *testing.T) {
 	tests := []struct {
-		name     string
-		tracker  *RoleTracker
-		expected string
+		name    string
+		tracker *RoleTracker
+		action  func(*RoleTracker, func(*reconcile.Request) logr.Logger)
+		check   func(*testing.T, *jsonLogCapture)
```
mimowo: Hm, I think this does not need to be a function, just a static list of expected log entries is enough, then we could assert by cmp.Diff by comparing the list of entries

mimowo:
Thank you 👍 
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16
/release-note-edit
```release-note
Optimized memory consumption by removing an unnecessary copy of the logger, which previously caused excessive memory usage in code hot paths.
```

mimowo:
/release-note-edit
```release-note
Fix excessive memory overhead in hot code paths by reusing the named logger in NewLogConstructor and avoiding unnecessary logger cloning.
```

tenzen-y:
/release-note-edit
```release-note
Observability: Fix excessive memory overhead in hot code paths by reusing the named logger in NewLogConstructor and avoiding unnecessary logger cloning.
```

PR #10366, opened 2026-04-10
TAS: incremental per-node usage aggregation in nonTasUsageCache

jzhaojieh:
#### What type of PR is this?

Add one of the following kinds:
/kind feature

Please also consider setting the area:
/area tas

Replace the full scan in usagePerNode() with an incrementally maintained nodeUsage map. update() and delete() now maintain pre-aggregated per-node totals, reducing usagePerNode() to O(nodes) making the TAS snapshot cheaper for clusters with large number of non TAS pods per node

N/A

```release-note
TAS: optimize performance of building the snapshot by pre-aggregating the node usage coming from non-TAS Pods.
```

### DIFF
--- a/pkg/cache/scheduler/tas_non_tas_pod_cache.go
+++ b/pkg/cache/scheduler/tas_non_tas_pod_cache.go
@@ -31,8 +31,9 @@ import (
 // nonTasUsageCache caches pod usage, to avoid
 // the hot path documented in kueue#8449.
 type nonTasUsageCache struct {
-	podUsage map[types.NamespacedName]podUsageValue
-	lock     sync.RWMutex
+	podUsage  map[types.NamespacedName]podUsageValue
+	nodeUsage map[string]resources.Requests // pre-aggregated per-node totals
+	lock      sync.RWMutex
 }
 
 type podUsageValue struct {
@@ -46,37 +47,74 @@ func (n *nonTasUsageCache) update(pod *corev1.Pod, log logr.Logger) {
 	n.lock.Lock()
 	defer n.lock.Unlock()
 
+	key := client.ObjectKeyFromObject(pod)
+
 	// delete terminated pods as they no longer use any capacity.
 	if utilpod.IsTerminated(pod) {
 		log.V(5).Info("Deleting terminated pod from the cache")
-		delete(n.podUsage, client.ObjectKeyFromObject(pod))
+		if old, found := n.podUsage[key]; found {
+			n.removeNodeUsage(old.node, old.usage, log)
+		}
+		delete(n.podUsage, key)
 		return
+	}
+
+	// Remove old entry if pod already exists (handles node migration, resource resize).
+	if old, found := n.podUsage[key]; found {
+		n.removeNodeUsage(old.node, old.usage, log)
 	}
 
 	log.V(5).Info("Adding non-TAS pod to the cache")
 	requests := resources.NewRequestsFromPodSpec(&pod.Spec)
-	n.podUsage[client.ObjectKeyFromObject(pod)] = podUsageValue{
+	n.podUsage[key] = podUsageValue{
 		node:  pod.Spec.NodeName,
 		usage: requests,
 	}
+	n.addNodeUsage(pod.Spec.NodeName, requests)
 }
 
-func (n *nonTasUsageCache) delete(key client.ObjectKey) {
+func (n *nonTasUsageCache) delete(key client.ObjectKey, log logr.Logger) {
 	n.lock.Lock()
 	defer n.lock.Unlock()
+	if old, found := n.podUsage[key]; found {
+		n.removeNodeUsage(old.node, old.usage, log)
+	}
 	delete(n.podUsage, key)
 }
 
 func (n *nonTasUsageCache) usagePerNode() map[string]resources.Requests {
 	n.lock.RLock()
 	defer n.lock.RUnlock()
-	usage := make(map[string]resources.Requests)
-	for _, podUsage := range n.podUsage {
-		if _, found := usage[podUsage.node]; !found {
-			usage[podUsage.node] = resources.Requests{}
-		}
-		usage[podUsage.node].Add(podUsage.usage)
-		usage[podUsage.node].Add(resources.Requests{corev1.ResourcePods: 1})
+	usage := make(map[string]resources.Requests, len(n.nodeUsage))
+	for node, reqs := range n.nodeUsage {
+		usage[node] = reqs.Clone()
 	}
 	return usage
+}
+
+// addNodeUsage increments the pre-aggregated per-node usage.
+// Must be called under write lock.
+func (n *nonTasUsageCache) addNodeUsage(node string, usage resources.Requests) {
+	if _, found := n.nodeUsage[node]; !found {
+		n.nodeUsage[node] = resources.Requests{}
+	}
+	n.nodeUsage[node].Add(usage)
+	n.nodeUsage[node][corev1.ResourcePods]++
+}
+
+// removeNodeUsage decrements the pre-aggregated per-node usage.
+// Must be called under write lock.
+func (n *nonTasUsageCache) removeNodeUsage(node string, usage resources.Requests, log logr.Logger) {
+	existing, found := n.nodeUsage[node]
+	if !found {
+		return
+	}
+	existing.Sub(usage)
+	existing[corev1.ResourcePods]--
+	if pods := existing[corev1.ResourcePods]; pods <= 0 {
+		if pods < 0 {
+			log.V(0).Info("Unexpected negative pod count in nodeUsage", "node", node, "podCount", pods)
+		}
+		delete(n.nodeUsage, node)
+	}
 }
--- a/pkg/cache/scheduler/tas_non_tas_pod_cache_test.go
+++ b/pkg/cache/scheduler/tas_non_tas_pod_cache_test.go
@@ -1,0 +1,199 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


mimowo:
/ok-to-test
I think this is a nice performance improvement for TAS. I think we could even consider cherrypicking as there are no API changes, and we could consider that as a performance bugfix.

Let me cc @gabesaba and @sohankunkerkar for review

pkg/cache/scheduler/tas_non_tas_pod_cache.go
```diff
@@ -46,37 +47,71 @@ func (n *nonTasUsageCache) update(pod *corev1.Pod, log logr.Logger) {
...
+// Must be called under write lock.
+func (n *nonTasUsageCache) removeNodeUsage(node string, usage resources.Requests) {
+	existing, found := n.nodeUsage[node]
+	if !found {
+		return
+	}
+	existing.Sub(usage)
+	existing[corev1.ResourcePods]--
+	if existing[corev1.ResourcePods] <= 0 {
+		delete(n.nodeUsage, node)
```
mwielgus, line 118: Can we log cases where pod count goes below 0 (this may indicate some problems)?

jzhaojieh: Added logging for this case

pkg/cache/scheduler/tas_non_tas_pod_cache.go
```diff
@@ -46,37 +47,74 @@ func (n *nonTasUsageCache) update(pod *corev1.Pod, log logr.Logger) {
...
+func (n *nonTasUsageCache) removeNodeUsage(node string, usage resources.Requests, log logr.Logger) {
+	existing, found := n.nodeUsage[node]
+	if !found {
+		return
+	}
+	existing.Sub(usage)
+	existing[corev1.ResourcePods]--
+	if pods := existing[corev1.ResourcePods]; pods <= 0 {
+		if pods < 0 {
+			log.V(5).Info("Negative pod count", "node", node)
```
mimowo: IIUC that should be V2 at least, since this is an anomaly (sign of a bug)

jzhaojieh: Bumped to logging level to v2

pkg/cache/scheduler/tas_non_tas_pod_cache.go
```diff
@@ -46,37 +47,74 @@ func (n *nonTasUsageCache) update(pod *corev1.Pod, log logr.Logger) {
...
+func (n *nonTasUsageCache) removeNodeUsage(node string, usage resources.Requests, log logr.Logger) {
+	existing, found := n.nodeUsage[node]
+	if !found {
+		return
+	}
+	existing.Sub(usage)
+	existing[corev1.ResourcePods]--
+	if pods := existing[corev1.ResourcePods]; pods <= 0 {
+		if pods < 0 {
+			log.V(2).Info("Negative pod count", "node", node)
```
sohankunkerkar: nit: consider adding `log.Error` or `V(0)` instead of `V(2)` so it surfaces in default logs. Also worth logging the actual count for debugging.

pkg/cache/scheduler/tas_non_tas_pod_cache_test.go
```diff
@@ -0,0 +1,195 @@
...
+					Requests: corev1.ResourceList{
+						corev1.ResourceCPU: resource.MustParse(cpu),
+					},
+				},
+			}},
+		},
+	}
+}
+
+func TestNonTasUsageCacheIncrementalUpdates(t *testing.T) {
```
sohankunkerkar, line 67: Good coverage! Does it make sense to add a test for a pod's resource requests changing in-place?

sohankunkerkar (commented):
Nice optimization! Two non-blocking nits but overall LGTM!

pkg/cache/scheduler/tas_non_tas_pod_cache.go
jzhaojieh: Makes sense, changed to `V(0)` to always log and included the count

pkg/cache/scheduler/tas_non_tas_pod_cache_test.go
jzhaojieh: Added additional test

pkg/cache/scheduler/tas_non_tas_pod_cache_test.go
```diff
@@ -0,0 +1,222 @@
...
+	newCache := func() *nonTasUsageCache {
+		return &nonTasUsageCache{
+			podUsage:  make(map[types.NamespacedName]podUsageValue),
+			nodeUsage: make(map[string]resources.Requests),
+		}
+	}
+
+	t.Run("add then delete same pod", func(t *testing.T) {
```
mimowo: nit, could you organize the tests in a table-driven fashion to avoid duplicating the setup and asserts code? For example the lines like `_, log := utiltesting.ContextWithLog(t)` or `cache := newCache()`. 

If we need to have some more open ended changes to pods, then I would introduce a param function like `updatePods func([]*corev1.Pod)`, or something along the lines. Another param could be `podsToDelete []typed.NamespacedName`.

jzhaojieh: Refactored into TDD format

mimowo: Actually, this is not exactly addressing my comment.

I would prefer to have testCases table, ptal for [example](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/util/logging/logging_test.go#L56-L73) or [here](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/cache/scheduler/tas_nodes_cache_test.go#L31-L46)

I think the params would be useful like:
initialPods  []corev1.Pod
podsUpdate func(pods []*corev1.Pod) {}
wantNodeUsage map[string]Usage

jzhaojieh: Done — refactored as a testCases table

mimowo:
/release-note-edit
```release-note
TAS: optimize performance of building the snapshot by pre-aggregating the node usage coming from non-TAS Pods.
```

pkg/cache/scheduler/tas_non_tas_pod_cache.go
```diff
@@ -46,37 +47,74 @@ func (n *nonTasUsageCache) update(pod *corev1.Pod, log logr.Logger) {
...
-	usage := make(map[string]resources.Requests)
-	for _, podUsage := range n.podUsage {
-		if _, found := usage[podUsage.node]; !found {
-			usage[podUsage.node] = resources.Requests{}
-		}
-		usage[podUsage.node].Add(podUsage.usage)
-		usage[podUsage.node].Add(resources.Requests{corev1.ResourcePods: 1})
+	usage := make(map[string]resources.Requests, len(n.nodeUsage))
+	for node, reqs := range n.nodeUsage {
+		usage[node] = reqs.Clone()
```
mimowo, line 90: Isn't this equivalent to just `usage := maps.Clone(n.nodeUsage)`?

pkg/cache/scheduler/tas_non_tas_pod_cache.go
```diff
@@ -46,37 +47,74 @@ func (n *nonTasUsageCache) update(pod *corev1.Pod, log logr.Logger) {
...
+			n.removeNodeUsage(old.node, old.usage, log)
+		}
+		delete(n.podUsage, key)
 		return
 	}
 
+	// Remove old entry if pod already exists (handles node migration, resource resize).
+	if old, found := n.podUsage[key]; found {
+		n.removeNodeUsage(old.node, old.usage, log)
+	}
```
mimowo, line 65: Do we have a test case covering the Pod resize actually?

pkg/cache/scheduler/tas_non_tas_pod_cache.go
```diff
@@ -46,37 +47,74 @@ func (n *nonTasUsageCache) update(pod *corev1.Pod, log logr.Logger) {
...
+	if !found {
+		return
+	}
+	existing.Sub(usage)
+	existing[corev1.ResourcePods]--
+	if pods := existing[corev1.ResourcePods]; pods <= 0 {
+		if pods < 0 {
+			log.V(0).Info("Unexpected negative pod count in nodeUsage", "node", node, "podCount", pods)
+		}
+		delete(n.nodeUsage, node)
```
mimowo, line 118: Do we have a test case which would fail if not for this line - to ensure we don't have a memory leak here? Asking cause some test asserts may not distinuguish nil from empty map.

mimowo (commented):
LGTM overall, thanks 👍

mimowo:
@jzhaojieh the PR basically lgtm, just raised some nits, let me know folks if you have cycles to address them. cc @Huang-Wei

jzhaojieh:
> @jzhaojieh the PR basically lgtm, just raised some nits, let me know folks if you have cycles to address them. cc @Huang-Wei

Sorry for the delay, will get these addressed today

pkg/cache/scheduler/tas_non_tas_pod_cache.go
jzhaojieh: Added explicit check on the map size in unit test

pkg/cache/scheduler/tas_non_tas_pod_cache.go
jzhaojieh: Added test case for pod resize

pkg/cache/scheduler/tas_non_tas_pod_cache.go
jzhaojieh: Not quite, i think `maps.Clone(n.nodeUsage)` would result in a shallow copy of values, here its explicitly doing a deep copy to maintain the existing behavior

mimowo (commented):
Thank you 👍 
/lgtm
/approve
I feel good about the change to cherrypick as we did historically with other TAS performance improvements. 
/cherrypick release-0.17
/cherrypick release-0.16

mimowo:
@jzhaojieh please prepare CPs in case the robot fails due to conficts

PR #10381, opened 2026-04-10
Use fresh counter metric for fetch.

mbobrovskyi:
#### What type of PR is this?
/kind cleanup

Use fresh counter metric for fetch.

Fixes #

We already had a similar issue with gauge metrics: https://github.com/kubernetes-sigs/kueue/pull/10096.

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10080/pull-kueue-test-integration-extended-main/2042568960041291776

```release-note
NONE
```

### DIFF
--- a/test/util/metrics.go
+++ b/test/util/metrics.go
@@ -66,15 +66,14 @@ func ExpectLQReservingActiveWorkloadsMetric(lq *kueue.LocalQueue, value int) {
 
 func ExpectLQAdmittedWorkloadsTotalMetric(lq *kueue.LocalQueue, priorityClass string, value int, customLabels ...string) {
 	ginkgo.GinkgoHelper()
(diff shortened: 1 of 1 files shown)


test/util/metrics.go
```diff
@@ -220,10 +217,10 @@ func ExpectLQFinishedWorkloadsGaugeMetric(lq *kueue.LocalQueue, count int) {
 	expectGaugeMetric(metrics.LocalQueueFinishedWorkloads, lvs, gomega.Equal(float64(count)))
 }
 
-func expectCounterMetric(metric prometheus.Counter, count int) {
+func expectCounterMetric(metric *prometheus.CounterVec, lvs []string, count int) {
```
mykysha: What do you think about changing it to `func expectCounterMetric(metric *prometheus.CounterVec, count int, lvs ...string)`?

mbobrovskyi: Done

PR #10384, opened 2026-04-10
Add feature gate for invalid ClusterQueue onFlavors updates

ShaanveerS:
#### What type of PR is this?
/kind feature
Introduce the `RejectUpdatesToCQWithInvalidOnFlavors` feature gate.
When the feature gate is disabled, updates to legacy ClusterQueues that already contain invalid `AdmissionCheckStrategy.OnFlavors` references are still allowed as long as those entries are unchanged.
When enabled, updates to such ClusterQueues are rejected until the invalid flavor references are fixed.

Follow up to https://github.com/kubernetes-sigs/kueue/pull/10336

```release-note
AdmissionChecks: Add the alpha `RejectUpdatesToCQWithInvalidOnFlavors` feature gate (disabled by default) to reject updates to existing ClusterQueues with invalid `AdmissionCheckStrategy.OnFlavors` references. 
action required: when enabling this feature gate, fix any existing invalid `OnFlavors` references before updating the affected ClusterQueues.
```

### DIFF
--- a/pkg/features/kube_features.go
+++ b/pkg/features/kube_features.go
@@ -352,6 +352,13 @@ const (
 	// Pod integration's IsActive() check, allowing quota to be released immediately
 	// when preempted pods begin terminating rather than waiting for the grace period.
 	FastQuotaReleaseInPodIntegration featuregate.Feature = "FastQuotaReleaseInPodIntegration"
+
+	// owner: @ShaanveerS
+	//
+	// issue: https://github.com/kubernetes-sigs/kueue/issues/7259
+	// Enables rejecting updates to ClusterQueues with invalid
+	// AdmissionCheckStrategy.OnFlavors references.
+	RejectUpdatesToCQWithInvalidOnFlavors featuregate.Feature = "RejectUpdatesToCQWithInvalidOnFlavors"
 )
 
 func init() {
@@ -542,6 +549,9 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 	},
 	FastQuotaReleaseInPodIntegration: {
 		{Version: version.MustParse("0.17"), Default: false, PreRelease: featuregate.Alpha},
+	},
+	RejectUpdatesToCQWithInvalidOnFlavors: {
+		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Alpha},
 	},
 }
 
--- a/pkg/webhooks/clusterqueue_webhook.go
+++ b/pkg/webhooks/clusterqueue_webhook.go
@@ -31,6 +31,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+	"sigs.k8s.io/kueue/pkg/features"
 	utilqueue "sigs.k8s.io/kueue/pkg/util/queue"
 	"sigs.k8s.io/kueue/pkg/util/roletracker"
 )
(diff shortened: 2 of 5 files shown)


mimowo (commented):
Thank you 👍 
/lgtm
/approve

pkg/features/kube_features.go
```diff
@@ -352,6 +352,13 @@ const (
 	// Pod integration's IsActive() check, allowing quota to be released immediately
 	// when preempted pods begin terminating rather than waiting for the grace period.
 	FastQuotaReleaseInPodIntegration featuregate.Feature = "FastQuotaReleaseInPodIntegration"
+
+	// owner: @ShaanveerS
+	//
+	// pr: https://github.com/kubernetes-sigs/kueue/pull/10336
```
tenzen-y:
```suggestion
	// issue: https://github.com/kubernetes-sigs/kueue/issues/7259
```

Could you use the tracking issue?

tenzen-y (commented):
Otherwise LGTM, thank you!

/hold

pkg/features/kube_features.go
ShaanveerS: done!

mimowo:
/lgtm
/unhold
as the remaining comment is now addressed

pkg/features/kube_features.go
```diff
@@ -543,6 +550,9 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 	FastQuotaReleaseInPodIntegration: {
 		{Version: version.MustParse("0.17"), Default: false, PreRelease: featuregate.Alpha},
 	},
+	RejectUpdatesToCQWithInvalidOnFlavors: {
+		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Alpha},
+	},
```
tenzen-y, line 555:
```suggestion
	RejectUpdatesToCQWithInvalidOnFlavors: {
		{Version: version.MustParse("0.16"), Default: false, PreRelease: featuregate.Alpha},
	},
```

@mimowo Isn't this `0.16`? Because we will cherry-pick this to released branch, right?

mimowo: we are not going to cherrypick the FG I think, should we?

mimowo: But even if we decide to cherrypick, Im ok with that I would prepare a dedicated PR for each branch setting the version as the branhc. This way the decision is decoupled. So on 0.17 release branch it would be 0.17.

tenzen-y: > we are not going to cherrypick the FG I think, should we?

Right? IIUC, we introduced this FG to avoid breaking their production just by updating Kueue version.
So, cherry-picking as alpha would be worth it.

If we fully align with K8S compatible rules, we can not make validations strong, but we will ignore the rule here. So, at least, I think we should disable it by default for a while.

mimowo: > Right? IIUC, we introduced this FG to avoid breaking their production just by updating Kueue version.
So, cherry-picking as alpha would be worth it.

But the default behavior already cherrypicked to 0.16 and 0.17 is to ignore the pre-existing validation errors, so as RejectUpdatesToCQWithInvalidOnFlavors=false. This means the user cannot say RejectUpdatesToCQWithInvalidOnFlavorz=true, but this is probably ok. So, I'm not clear what cherrypicking would change there.

> If we fully align with K8S compatible rules, we can not make validations strong, but we will ignore the rule here. So, at least, I think we should disable it by default for a while.

Sure, so for 0.18 the feature gate is planned as Alpha, and we may consider the FG as Beta in 0.20 or later.

tenzen-y: > But the default behavior already cherrypicked to 0.16 and 0.17 is to ignore the pre-existing validation errors, so as RejectUpdatesToCQWithInvalidOnFlavors=false. 

But the commits are not released yet. So, I still think that we should cherry pick this one as alpha to released branches.

> Sure, so for 0.18 the feature gate is planned as Alpha, and we may consider the FG as Beta in 0.20 or later.

I think so too.

mimowo: > But the commits are not released yet. So, I still think that we should cherry pick this one as alpha to released branches.

I'm ok with that, I just wanted to clarify about the argument: "we introduced this FG to avoid breaking their production just by updating Kueue version." is not fully accurate. Avoiding the production breaking is already the default behavior that was cherrypicked. 

The feature gate is introduced to allow simplifying the code in the future and to drop the code responsible for graceful handling of invalid CQs.

tenzen-y: > Avoiding the production breaking is already the default behavior that was cherrypicked.

I am missing something here. Do you indicate https://github.com/kubernetes-sigs/kueue/pull/10336?
If so, what "the default behavior" does that mean?

10336 has not been released yet, so no default behavior yet.
If we didn't cherry-pick this FG PR, the ClusterQueue updates sometimes will fail on an invalid ClusterQueue user env. Surely, the invalid ClusterQueue parameters are bad, but from the user PoV, they can not operate on what they have been proceeding with before Kueue update.

Isn't the situation "breaking"?

PR #10406, opened 2026-04-11
TAS: Use hierarchical LeastFreeCapacity for unconstrained workloads

abhi179:
### **TL;DR**
Fixes rack fragmentation for unconstrained workloads by making LeastFreeCapacity topology-aware. Instead of scheduling at node level only, the algorithm now packs workloads top-down (rack → node), preserving empty racks for large jobs (e.g., NVL72).

**What changed**

When TASProfileMixed is enabled and a workload uses unconstrained topology, LeastFreeCapacity now evaluates capacity hierarchically:

- Previously: scheduling was done only at node level (lowest level)
- Now: scheduling is done top-down across topology levels
    - Select the domain with least free capacity at the highest level (e.g., rack)
    - Recursively select the tightest-fitting child domain (e.g., node)

### **Detailed Description:**
**Title: LeastFreeCapacity for unconstrained workloads should consider higher topology levels to prevent rack fragmentation**

**Issue Description :**
On a large GPU cluster with GB200 NVL72 racks (~33 racks, ~18 nodes per rack, 4 GPUs per node), single-GPU inference workloads using unconstrained-topology: "true" are being scattered across many partially-filled racks instead of being consolidated into the fewest racks possible.

The cluster topology has two levels:

 kind: Topology
 metadata:
   name: rack-topology
 spec:
   levels:
   - nodeLabel: nvidia.com/gpu.clique   # rack
   - nodeLabel: kubernetes.io/hostname   # node

With TASProfileMixed enabled (default since v0.15), unconstrained workloads use LeastFreeCapacity. However, the algorithm only operates at the hostname (lowest) level — it picks the globally busiest node without considering rack-level grouping. This causes
single-GPU pods to land on partially-used nodes across many different racks, creating widespread rack fragmentation.

**What should have been the expected behavior:**

LeastFreeCapacity should traverse the topology tree top-down for unconstrained workloads: first select the busiest rack (least free capacity at the rack level), then select the busiest node within that rack. This would consolidate single-GPU workloads into the fewest racks possible.

Why this matters (GB200/NVL72 use case):

GB200 NVL72 workloads require an entire rack (72 GPUs across 18 nodes connected via NVLink). When single-GPU inference pods (image-encoding, proxy, etc.) are scattered across many racks — each consuming just 1 GPU per rack — it becomes impossible to find a fully empty rack for NVL72 jobs. The goal is to pack unconstrained single-GPU workloads into the fewest racks possible, keeping maximum racks entirely free for large topology-constrained jobs.

**Reproduction:**

 1. Create a topology with two levels (rack → hostname)
 2. Deploy multiple single-GPU StatefulSets with kueue.x-k8s.io/podset-unconstrained-topology: "true"
 3. Observe pods scattered across many racks instead of consolidating into the fewest racks

**Evidence from production (Kueue v0.16.1):**

 - Pod deleted from a node landed back on the same rack instead of going to the busiest rack (LeastFree)
 - TopologyAssignment shows only levels: ["kubernetes.io/hostname"] — rack level absent
 - Rack utilization ranged from 1–72 GPUs across 33 racks, confirming no rack-level consolidation

**Root cause in code:**

pkg/cache/scheduler/tas_flavor_snapshot.go:

 1. levelKey() (line 1105): For unconstrained, returns s.lowestLevel() = kubernetes.io/hostname, skipping the rack level entirely
 2. findLevelWithFitDomains() (line 1250): Upward traversal to rack level is blocked by !unconstrained guard: if levelIdx > 0 && !unconstrained {
      return s.findLevelWithFitDomains(levelIdx-1, ...)
  }
 3. LeastFreeCapacity sorts hostname-level domains only — it never groups or scores by rack

Proposed behavior:
For unconstrained + LeastFreeCapacity with multi-level topologies, the algorithm should:

 1. Score capacity at each topology level (rack, then hostname)
 2. Select the rack with the least free capacity that can fit the workload
 3. Within that rack, select the node with the least free capacity

This preserves the "no hard placement constraint" semantics of unconstrained while enabling topology-aware bin-packing to reduce fragmentation.

Environment:

 - Kueue version: v0.16.1
 - Feature gates: TASProfileMixed=true (default)
 - Kubernetes: AKS with unmanaged GPU nodes
 - Topology: 2 levels (nvidia.com/gpu.clique → kubernetes.io/hostname)

```release-note
TAS: Introduce hierarchical LeastFreeCapacity for unconstrained workloads, gated by the TASHierarchicalUnconstrained feature gate (alpha, disabled by default). When enabled alongside TASProfileMixed, the algorithm resolves topology starting from the highest level (e.g., rack) instead of the node level, packing pods into the busiest racks first to reduce fragmentation.
```

### DIFF
--- a/keps/2724-topology-aware-scheduling/README.md
+++ b/keps/2724-topology-aware-scheduling/README.md
@@ -1561,8 +1561,9 @@ Without `TASHierarchicalUnconstrained`, the existing behavior is preserved: the 
 level for unconstrained workloads is resolved at the node level, and `LeastFreeCapacity`
 sorting occurs only within that level.
 
-This feature gate is alpha and disabled by default. It is not mutually exclusive with
-`TASProfileMixed` — it augments the unconstrained scheduling path when both are enabled.
+This feature gate is Beta and enabled by default in v0.18. It is planned to graduate to
+GA in v0.21. It is not mutually exclusive with `TASProfileMixed` — it augments the
+unconstrained scheduling path when both are enabled.
 
 ### Two-level Topology Aware scheduling
 In consideration of a [Story 5](#story-5) a two-level scheduling is introduced.
--- a/keps/2724-topology-aware-scheduling/README.md
+++ b/keps/2724-topology-aware-scheduling/README.md
@@ -53,6 +53,7 @@
     - [Selecting the algorithm](#selecting-the-algorithm)
       - [Until v0.14](#until-v014)
       - [Since v0.15](#since-v015)
+      - [Hierarchical LeastFreeCapacity for unconstrained workloads](#hierarchical-leastfreecapacity-for-unconstrained-workloads)
   - [Two-level Topology Aware scheduling](#two-level-topology-aware-scheduling)
     - [Example](#example-1)
   - [Multi-level Topology Aware scheduling](#multi-level-topology-aware-scheduling)
@@ -1541,10 +1542,27 @@ Since v0.15, the available feature gates are as follows:
 | TASProfileBestFit (deprecated)                   | BestFit           | BestFit           | BestFit           |
 | TASProfileLeastFreeCapacity (removed in v0.17)   | LeastFreeCapacity | LeastFreeCapacity | LeastFreeCapacity |
 | TASBalancedPlacement                             | BalancedPlacement | BestFit           | LeastFreeCapacity |
+| TASHierarchicalUnconstrained                     | BestFit           | BestFit           | LeastFreeCapacity (hierarchical) |
 
 Based on the user feedback, we decided to make `TASProfileMixed` default starting in v0.15. (The corresponding feature gate is hence obsolete; we formally keep it "deprecated" for backwards compatibility). It differs from the previous default only in the `unconstrained` case - in which Kueue should prioritize minimizing fragmentation which is provided by the `LeastFreeCapacity` algorithm.
 
 For users still preferring the "always BestFit" profile, we introduce the `TASProfileBestFit` feature gate, marking it as deprecated. We will remove it in v0.17 if we see no report indicating a need for that configuration.
+
+##### Hierarchical LeastFreeCapacity for unconstrained workloads
+
+When `TASHierarchicalUnconstrained` is enabled (alongside the default `TASProfileMixed`
+profile), the `LeastFreeCapacity` algorithm for `unconstrained` workloads resolves the
+topology level starting from the **highest level** in the hierarchy (e.g., rack) rather
+than the lowest level (host). This means the algorithm first identifies the busiest
+higher-level domain and then packs pods into the busiest nodes within that domain,
+resulting in hierarchical packing behavior.
+
+Without `TASHierarchicalUnconstrained`, the existing behavior is preserved: the topology
+level for unconstrained workloads is resolved at the node level, and `LeastFreeCapacity`
+sorting occurs only within that level.
+
+This feature gate is alpha and disabled by default. It is not mutually exclusive with
+`TASProfileMixed` — it augments the unconstrained scheduling path when both are enabled.
 
 ### Two-level Topology Aware scheduling
 In consideration of a [Story 5](#story-5) a two-level scheduling is introduced.
--- a/pkg/cache/scheduler/tas_cache_test.go
+++ b/pkg/cache/scheduler/tas_cache_test.go
@@ -595,11 +595,68 @@ func TestFindTopologyAssignments(t *testing.T) {
 					Levels: defaultOneLevel,
 					Domains: []tas.TopologyDomainAssignment{
 						{Count: 1, Values: []string{"x1"}},
+						{Count: 1, Values: []string{"x3"}},
+						{Count: 1, Values: []string{"x5"}},
+						{Count: 1, Values: []string{"x2"}},
+						{Count: 2, Values: []string{"x6"}},
+					},
+				},
+			}},
+		},
+		// Topology: scatteredNodes (block → rack → host)
+		//   b1/r1: x3(4cpu), x5(1cpu), x1(1cpu) = 6cpu total
+		//   b2/r1: x6(2cpu), x2(1cpu) = 3cpu total
+		// With TASHierarchicalUnconstrained, LeastFreeCapacity sorts top-down:
+		// block b1 has more used capacity than b2, so b1 is chosen first, packing
+		// 6 pods into b1's nodes (x3:4, x5:1, x1:1) instead of scattering across blocks.
+		"unconstrained; 6 pods fit into hosts packed within rack; TASHierarchicalUnconstrained": {
+			nodes:  scatteredNodes,
+			levels: defaultThreeLevels,
+			podSets: []PodSetTestCase{{
+				topologyRequest: &kueue.PodSetTopologyRequest{
+					Unconstrained: ptr.To(true),
+				},
+				requests: resources.Requests{
+					corev1.ResourceCPU: 1000,
+				},
+				count: 6,
+				wantAssignment: &tas.TopologyAssignment{
+					Levels: defaultOneLevel,
+					Domains: []tas.TopologyDomainAssignment{
+						{Count: 1, Values: []string{"x1"}},
 						{Count: 4, Values: []string{"x3"}},
 						{Count: 1, Values: []string{"x5"}},
 					},
 				},
 			}},
+			featureGates: map[featuregate.Feature]bool{features.TASHierarchicalUnconstrained: true},
+		},
+		// Same topology as above but with TASHierarchicalUnconstrained disabled.
+		// Without hierarchical packing, pods are scattered across all nodes sorted
+		// by least free capacity at the node level only (no rack awareness).
+		"unconstrained; 6 pods fit into hosts scattered across blocks; TASHierarchicalUnconstrained disabled": {
+			nodes:  scatteredNodes,
+			levels: defaultThreeLevels,
+			podSets: []PodSetTestCase{{
+				topologyRequest: &kueue.PodSetTopologyRequest{
+					Unconstrained: ptr.To(true),
+				},
+				requests: resources.Requests{
+					corev1.ResourceCPU: 1000,
+				},
+				count: 6,
+				wantAssignment: &tas.TopologyAssignment{
+					Levels: defaultOneLevel,
+					Domains: []tas.TopologyDomainAssignment{
+						{Count: 1, Values: []string{"x1"}},
+						{Count: 1, Values: []string{"x3"}},
+						{Count: 1, Values: []string{"x5"}},
+						{Count: 1, Values: []string{"x2"}},
+						{Count: 2, Values: []string{"x6"}},
+					},
+				},
+			}},
+			featureGates: map[featuregate.Feature]bool{features.TASHierarchicalUnconstrained: false},
 		},
 		"unconstrained; a single pod fits into each host; BestFit": {
 			nodes:  defaultNodes,
@@ -618,14 +675,21 @@ func TestFindTopologyAssignments(t *testing.T) {
 						{
 							Count: 1,
 							Values: []string{
-								"x2",
+								"x3",
 							},
 						},
 					},
 				},
 			}},
 		},
-		"unconstrained; a single pod fits into each host; LeastFreeCapacity; TASProfileMixed": {
+		// Topology: defaultNodes (block → rack → host), all 1cpu except x4(2cpu)
+		//   b1/r1: x3(1cpu); b1/r2: x5(1cpu), x1(1cpu), x6(1cpu)
+		//   b2/r1: x2(1cpu); b2/r2: x4(2cpu)
+		// With TASHierarchicalUnconstrained enabled, the algorithm picks the
+		// tightest-fitting block first (b2: 3cpu total vs b1: 4cpu total), then
+		// the tightest host within b2 → x2(1cpu). Without this gate, the
+		// node-level-only sort picks x3 (least free at node level globally).
+		"unconstrained; a single pod fits into each host; LeastFreeCapacity; TASHierarchicalUnconstrained": {
 			nodes:  defaultNodes,
 			levels: defaultThreeLevels,
 			podSets: []PodSetTestCase{{
@@ -648,7 +712,32 @@ func TestFindTopologyAssignments(t *testing.T) {
 					},
 				},
 			}},
-			featureGates: map[featuregate.Feature]bool{features.TASProfileMixed: true},
+			featureGates: map[featuregate.Feature]bool{features.TASHierarchicalUnconstrained: true},
+		},
+		"unconstrained; a single pod fits into each host; LeastFreeCapacity; TASProfileMixed and TASHierarchicalUnconstrained": {
+			nodes:  defaultNodes,
+			levels: defaultThreeLevels,
+			podSets: []PodSetTestCase{{
+				topologyRequest: &kueue.PodSetTopologyRequest{
+					Unconstrained: ptr.To(true),
+				},
+				requests: resources.Requests{
+					corev1.ResourceCPU: 1000,
+				},
+				count: 1,
+				wantAssignment: &tas.TopologyAssignment{
+					Levels: defaultOneLevel,
+					Domains: []tas.TopologyDomainAssignment{
+						{
+							Count: 1,
+							Values: []string{
+								"x2",
+							},
+						},
+					},
+				},
+			}},
+			featureGates: map[featuregate.Feature]bool{features.TASProfileMixed: true, features.TASHierarchicalUnconstrained: true},
 		},
 		"block required; 4 pods fit into one host each; BestFit": {
 			nodes:  binaryTreesNodes,
--- a/pkg/cache/scheduler/tas_cache_test.go
+++ b/pkg/cache/scheduler/tas_cache_test.go
@@ -602,6 +602,7 @@ func TestFindTopologyAssignments(t *testing.T) {
 					},
 				},
 			}},
+			featureGates: map[featuregate.Feature]bool{features.TASHierarchicalUnconstrained: false},
 		},
 		// Topology: scatteredNodes (block → rack → host)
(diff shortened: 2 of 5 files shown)


pkg/cache/scheduler/tas_cache_test.go
```diff
@@ -595,10 +595,8 @@ func TestFindTopologyAssignments(t *testing.T) {
 					Levels: defaultOneLevel,
 					Domains: []tas.TopologyDomainAssignment{
 						{Count: 1, Values: []string{"x1"}},
-						{Count: 1, Values: []string{"x3"}},
+						{Count: 4, Values: []string{"x3"}},
 						{Count: 1, Values: []string{"x5"}},
-						{Count: 1, Values: []string{"x2"}},
-						{Count: 2, Values: []string{"x6"}},
 					},
```
mszadkow, line 629: Those test changes are valid, but we might want to consider to add a separate test for the un-gated scenario if it should remain supported.
wdyt @PBundyra ?

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1103,6 +1103,9 @@ func (s *TASFlavorSnapshot) levelKey(topologyRequest *kueue.PodSetTopologyReques
 	case isSliceTopologyOnlyRequest(topologyRequest):
 		return ptr.To(s.highestLevel())
 	case ptr.Deref(topologyRequest.Unconstrained, false):
+		if features.Enabled(features.TASProfileMixed) {
+			return ptr.To(s.highestLevel())
+		}
```
mimowo: A couple of thoughts here:
1. I know the idea of using the lowest level for "unconstrained" was to minimize strictly the fragmentation at the node level, which is not guaranteed in the new hierarchical mode, OTOH, as pointed out in this issue this leads to increased fragmentation at the rack level. I think it makes sense that for topologies we optimize fragmentation at different levels. Additionally, since this is "unconstrained" mode, Kueue has freedom of choice for the exact algo, from this perspective tweaking the algo as proposed sgtm
2. since TASProfileMixed is default behavior, and this PR is changing it, so we may consider a feature gate to ensure users can bailout, something like TASHierarchicalUnconstrained
3. we may also consider the "unconstrained" annotation to indicate the level for the top-level LeastFreeCapacity behavior, similarly as "preferred" and "required" point to a level, but I'm not clear we have enough motivation for that currently. 

So I'm considering the approaches:
A. Do as proposed in the PR, but use a feature gate TASHierarchicalUnconstrained for bailout option
B. Allow the "unconstrained" annotation to indicate the level at each it starts, default is "node" which is used for just "true" value

I'm leaning towards starting with A for simplicity, even as a bugfix, and only then do B if this is proven not enough.

cc @PBundyra who worked on the initial version of the algo.

abhi179: good idea, let me introduce it under a feature gate TASHierarchicalUnconstrained

pkg/cache/scheduler/tas_cache_test.go
```diff
@@ -603,6 +603,28 @@ func TestFindTopologyAssignments(t *testing.T) {
 				},
 			}},
 		},
+		"unconstrained; 6 pods fit into hosts packed within rack; TASHierarchicalUnconstrained": {
```
mwielgus, line 612: Consider adding a brief comment explaining the test, topology and domains. See above.

pkg/cache/scheduler/tas_cache_test.go
```diff
@@ -627,7 +649,7 @@ func TestFindTopologyAssignments(t *testing.T) {
 				},
 			}},
 		},
-		"unconstrained; a single pod fits into each host; LeastFreeCapacity; TASProfileMixed": {
+		"unconstrained; a single pod fits into each host; LeastFreeCapacity; TASHierarchicalUnconstrained": {
```
mwielgus, line 692: Was TasProvileMixed removed?

pkg/cache/scheduler/tas_cache_test.go
```diff
@@ -603,6 +603,28 @@ func TestFindTopologyAssignments(t *testing.T) {
...
+				wantAssignment: &tas.TopologyAssignment{
+					Levels: defaultOneLevel,
+					Domains: []tas.TopologyDomainAssignment{
+						{Count: 1, Values: []string{"x1"}},
+						{Count: 4, Values: []string{"x3"}},
+						{Count: 1, Values: []string{"x5"}},
+					},
+				},
+			}},
+			featureGates: map[featuregate.Feature]bool{features.TASHierarchicalUnconstrained: true},
```
mwielgus, line 632: To ensure robust coverage of the feature flag logic, consider adding a corresponding test case with the exact same setup but with  features.TASHierarchicalUnconstrained: false to verify the fallback behavior explicitly.

pkg/cache/scheduler/tas_flavor_snapshot.go
mwielgus, line 1095: Do we log what key is eventually used?

mwielgus:
And please sign the CLA.

pkg/cache/scheduler/tas_flavor_snapshot.go
abhi179: Added a V(3) log line in findTopologyAssignment right after the levelKey is resolved: s.log.V(3).Info("resolved topology level key", "topologyKey", *topologyKey, "unconstrained", unconstrained). This logs which key the scheduler selected for every assignment.

pkg/cache/scheduler/tas_cache_test.go
abhi179: Done — added "unconstrained; 6 pods fit into hosts scattered across blocks; TASHierarchicalUnconstrained disabled" with the same scatteredNodes setup but gate=false. It verifies the fallback: pods scatter across all nodes (matching the existing BestFit behavior).

pkg/cache/scheduler/tas_cache_test.go
abhi179: No, TASProfileMixed was not removed. It's still the default-on gate that enables LeastFreeCapacity sorting for unconstrained workloads. This PR introduces a separate gate TASHierarchicalUnconstrained (per mimowo's suggestion) that, when enabled alongside
 TASProfileMixed, makes LeastFreeCapacity start from the highest topology level instead of just the node level. The test was renamed from TASProfileMixed → TASHierarchicalUnconstrained because the behavioral change being tested is specifically gated by the new flag.

pkg/cache/scheduler/tas_cache_test.go
abhi179: Added in the second commit — test case "unconstrained; 6 pods fit into hosts scattered across blocks; TASHierarchicalUnconstrained disabled" verifies the un-gated fallback where pods scatter across all nodes sorted by least free capacity at node level only (no rack awareness).

abhi179:
> And please sign the CLA.

done

abhi179:
@mimowo @kannon92 quick ping on this PR, I have addressed all the feedback. it doesn't change the current behavior, and the new flag is gated.

mimowo:
> @mimowo @kannon92 quick ping on this PR, I have addressed all the feedback. it doesn't change the current behavior, and the new flag is gated.

Great, Adding to my ToDo and will try to review EoW. In the meanwhile it would be great if @PBundyra could give it a pass who has most context here as worked on the initial version. 

If we agree about changing the behavior then I'm thinking this FG could be Beta in 0.18, and cherrypicked as Alpha down to 0.16 and 0.17. wdyt?

Please also:
1. add a release in the release-note box in the PR description, I just injected the box with some template, feel free to adjust
2. adjust the TAS KEP where it mentions the available algorithms

pkg/cache/scheduler/tas_cache_test.go
PBundyra: Could you elaborate on the un-gated scenario?

mimowo:
@abhi179 please also make sure all comments are addressed clearly one way or another - either confirm that you applied the suggestion (then feel free to mark the comment as resolved), or explain why you didn't, so that we don't need to reverse engineer your actions and motivations. For example this comment does not have any answer: https://github.com/kubernetes-sigs/kueue/pull/10406#discussion_r3078880555

pkg/cache/scheduler/tas_cache_test.go
abhi179: Added layout information + per-test capacity explanations

pkg/cache/scheduler/tas_cache_test.go
abhi179: The un-gated scenario is the current (existing) behavior: when TASHierarchicalUnconstrained is disabled, workloads with Unconstrained: true resolve their topology level at the node level and sort by
 LeastFreeCapacity only within that level. 
 
 When the gate is enabled, the topology level is resolved starting from the highest level (e.g., rack), so LeastFreeCapacity sorting first picks the busiest rack, then finds the busiest node within it —
 giving a hierarchical packing behavior.
 
 The gate-off test (TASHierarchicalUnconstrained disabled) validates that the existing node-level behavior is preserved when the feature gate is off.

abhi179:
> > @mimowo @kannon92 quick ping on this PR, I have addressed all the feedback. it doesn't change the current behavior, and the new flag is gated.
> 
> Great, Adding to my ToDo and will try to review EoW. In the meanwhile it would be great if @PBundyra could give it a pass who has most context here as worked on the initial version.
> 
> If we agree about changing the behavior then I'm thinking this FG could be Beta in 0.18, and cherrypicked as Alpha down to 0.16 and 0.17. wdyt?
> 
> Please also:
> 
> 1. add a release in the release-note box in the PR description, I just injected the box with some template, feel free to adjust
> 2. adjust the TAS KEP where it mentions the available algorithms

@mimowo updated release note and TAS KEP documentation.

@mimowo That sounds like a good plan. Starting as Alpha cherry-picked to v0.16/v0.17 gives existing users access to the fix sooner, and promoting to Beta in v0.18 makes sense once it's had some validation. Happy to help with the cherry-picks once this lands on main.

pkg/cache/scheduler/tas_cache_test.go
```diff
@@ -627,7 +682,14 @@ func TestFindTopologyAssignments(t *testing.T) {
 				},
 			}},
 		},
-		"unconstrained; a single pod fits into each host; LeastFreeCapacity; TASProfileMixed": {
```
PBundyra, line 630: Can we keep the old test as well to showcase how the algorithm acts without `TASHierarchicalUnconstrained`?

PBundyra:
Overall lgtm just a small comment regarding tests

pkg/features/kube_features.go
```diff
@@ -431,6 +439,9 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 		{Version: version.MustParse("0.10"), Default: false, PreRelease: featuregate.Alpha},
 		{Version: version.MustParse("0.15"), Default: true, PreRelease: featuregate.Beta},
 	},
+	TASHierarchicalUnconstrained: {
+		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Alpha},
```
mimowo: Let's make it Beta in 0.18, and backport it as alpha only in a small follow up. I think this new approach makes more sense actually as default. Please also add a note to graduate to GA in 0.21 so that we have enough time to collect feedback in case it was breaking for someone,

abhi179: Done — promoted to Beta (default: true) in v0.18. Added GA graduation note for v0.21 in the feature gate comment, KEP, and versioned_feature_list.yaml.

pkg/cache/scheduler/tas_cache_test.go
abhi179: Done. Added "unconstrained; a single pod fits into each host; LeastFreeCapacity; TASProfileMixed" (gate-off) alongside the existing gate-on test. The gate-off test shows node-level-only sorting (picks
 x3), while the gate-on test shows hierarchical rack→node sorting (picks x2).

tenzen-y:
I land here right now

abhi179:
Thanks @mimowo and @PBundyra for the thorough review! All feedback has been addressed. Let me know if I missed anything.

mimowo:
/lgtm
/approve
Thank you 👍 
/hold
in case @tenzen-y has extra comments

mwysokin:
@abhi179 Thank you for submitting the improvement 🙏 Since you have some real-life work experience with prod workloads on GB200 let me pick your brain about something.

I think in general our strategy going forward should be to have configurable placement policies e.g. in heterogenous clusters there should be an option to set default placement policy at the topology level (e.g. GB200 vs H100) but also a user should be able to override it if they feel the need to. Would this be a useful in your system? 

I'm anticipating many more placement policies in the future and treating them individually, hiding them under feature gates and exposing them as pre-configured packages of placement policies doesn't seem efficient to me. That's my thinking.

mimowo:
> hiding them under feature gates and exposing them as pre-configured packages of placement policies doesn't seem efficient to me. That's my thinking.

Yes, for sure, [feature gates](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/feature-gates.md#feature-gates-in-kubernetes) are transitive in nature, quoting from the guide:

```
Feature "gates" are how Kubernetes manages behavioral changes as they progress through the development lifecycle. This gives cluster operators a way to try new features or to disable them when something goes wrong.
```

Feature gates are useful to start with a speculative feature, or risky feature, but once we collect user feedback we should either deprecate feature gates, or graduate to GA. The most direct translation of a feature gate is global configuration knob, but in some cases we may replace them with CRD-level knobs.

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1541,11 +1542,29 @@ Since v0.15, the available feature gates are as follows:
...
 For users still preferring the "always BestFit" profile, we introduce the `TASProfileBestFit` feature gate, marking it as deprecated. We will remove it in v0.17 if we see no report indicating a need for that configuration.
 
+##### Hierarchical LeastFreeCapacity for unconstrained workloads
+
+When `TASHierarchicalUnconstrained` is enabled (alongside the default `TASProfileMixed`
+profile), the `LeastFreeCapacity` algorithm for `unconstrained` workloads resolves the
+topology level starting from the **highest level** in the hierarchy (e.g., rack) rather
+than the lowest level (host). This means the algorithm first identifies the busiest
+higher-level domain and then packs pods into the busiest nodes within that domain,
+resulting in hierarchical packing behavior.
```
mimowo, line 1558: Actually, looking at this PR again I'm not sure we want to start from the highest level. Assume we have the following levels:
- block
- rack
- hostname (node)

And we have many blocks that could fully accomodate the workload, so we would choose the random block, and commit to place the workload in a block which is not fragmented internally.

So, I'm wondering the better strategy might be to actually start from the highest level at whcih the workload cannot fully fit. Wdyt?

pkg/cache/scheduler/tas_cache_test.go
```diff
@@ -602,6 +602,62 @@ func TestFindTopologyAssignments(t *testing.T) {
...
 			}},
+			featureGates: map[featuregate.Feature]bool{features.TASHierarchicalUnconstrained: false},
+		},
+		// Topology: scatteredNodes (block → rack → host)
+		//   b1/r1: x3(4cpu), x5(1cpu), x1(1cpu) = 6cpu total
+		//   b2/r1: x6(2cpu), x2(1cpu) = 3cpu total
+		// With TASHierarchicalUnconstrained, LeastFreeCapacity sorts top-down:
+		// block b1 has more used capacity than b2, so b1 is chosen first, packing
+		// 6 pods into b1's nodes (x3:4, x5:1, x1:1) instead of scattering across blocks.
+		"unconstrained; 6 pods fit into hosts packed within rack; TASHierarchicalUnconstrained": {
```
mimowo, line 613: Let's add a test cases for the scanario when we have multiple blocks, and each of them fully accomodates the workload, but choosing the random block may result in placing the workload in a block which is not much fragmented, as described in https://github.com/kubernetes-sigs/kueue/pull/10406/changes#r3189306058. I'm wondering if my understanding is correct or not.

pkg/features/kube_features.go
```diff
@@ -431,6 +440,9 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 		{Version: version.MustParse("0.10"), Default: false, PreRelease: featuregate.Alpha},
 		{Version: version.MustParse("0.15"), Default: true, PreRelease: featuregate.Beta},
 	},
+	TASHierarchicalUnconstrained: {
+		{Version: version.MustParse("0.18"), Default: true, PreRelease: featuregate.Beta},
```
mimowo, line 444: After the second thought, we may need to keep it in Alpha for 0.18, especially if I'm correct about the issue in the current version: https://github.com/kubernetes-sigs/kueue/pull/10406/changes#r3189316475

mimowo:
I read the PR again, and I think likely it is going in the right direction, but I think I found a flaw (or at least I'm not sure yet about it), so let me cancel approval for now to avoid premature merging.
/approve cancel

Also, I would like to explore more how we could confirm how it performs in practice, so that we can have some better graduation criteria. Intuitively I think the approach is good (with the caveat mentioned), but I'm wondering if we can some better experimental support for it, for example;
1. are you going to use it on prod, how would you measure the improvement?
2. do you have maybe some ideas for "simulated" measuring of the improvement?

I would also be happy to learn more about your use case, so that we can better learn how people use "unconstrained" in practice, for example:
a) what is the percentage of "unconstrained" / "preferred" / "required" workloads
b) what do you use "unconstrained" for - training / inference / auxiliary jobs
c) what is the relative priority of the "unconstained" jobs to other workloads.

The practice is super relevant here, because the algos BestFit / LeastFreeCapactiy, etc are heuristics, which we thought make sense, but proving anything is very hard.

The forums that come to my mind:
1. feel free to reach out on slack (kubernetes workspace), either DM, and wg-batch
2. maybe you could present the intended improvement on  [wg-batch meetings](https://github.com/kubernetes/community/tree/main/wg-batch)

tenzen-y:
/lgtm cancel

based on https://github.com/kubernetes-sigs/kueue/pull/10406#issuecomment-4380420346

PR #9344, opened 2026-02-18
Fix StatefulSet MultiKueue adapter selection by creating workload in STS reconciler

IrvingMg:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

Currently, the pod framework creates the workload with the pod as the controller owner (and without the `JobOwnerGVKAnnotation`), causing the MK workload controller's `adapter()` function to select the pod adapter instead of the STS adapter.

This PR addresses the issue by having the STS reconciler create the workload (following the LWS pattern) with `JobOwnerGVKAnnotation` set at creation time, ensuring that MK always uses the STS adapter.

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note
MultiKueue: for the StatefulSet integration copy the entire StatefulSet onto the worker clusters. This allows
for proper management (and replacements) of Pods on the worker clusters.
```

PR #10435, opened 2026-04-13
tas: strip stale topology nodeSelector keys on workload re-evaluation

abhi179:
When a TAS workload is evicted (e.g., due to PodsReadyTimeout after a partial pod deletion by a cluster health manager), the surviving pods retain the kubernetes.io/hostname nodeSelector injected during the prior admission by the topology ungater. When the workload is re-queued, the pod group controller reconstructs the workload from the surviving pod's spec via DeepCopy, carrying the stale nodeSelector into the new PodSet template.

During TAS topology assignment evaluation, findTopologyAssignment() merges PodSet.NodeSelector with AdmissionCheck PodSetUpdates. The stale kubernetes.io/hostname constrains node filtering to a single node (the old assignment), causing the workload to be permanently stuck in Pending — it can never be re-admitted because the target node may no longer exist or have capacity.

Fix: After merging the combined nodeSelector, strip any keys that match TAS topology level keys (e.g., kubernetes.io/hostname, topology.kubernetes.io/zone). These keys are always recomputed by TAS during topology assignment, so prior values must not constrain node filtering. This allows evicted workloads to be re-admitted to any valid topology domain.

Observed in production: CHM NodeStatusUnknown repair action on GB200 clusters force-deleted individual pods from multi-pod GPU inference StatefulSets. 86% of affected workloads (6/7) became permanently stuck due to this bug.

#### What type of PR is this?

 /kind bug
 /area tas

When a TAS workload is evicted (e.g., due to PodsReadyTimeout after a partial pod deletion by a cluster health manager), the surviving pods retain the kubernetes.io/hostname nodeSelector injected during
prior admission by the topology ungater. When the workload is re-queued, the pod group controller reconstructs the workload from the surviving pod's spec via DeepCopy, carrying the stale nodeSelector into
the new PodSet template.

During TAS topology assignment evaluation, findTopologyAssignment() merges PodSet.NodeSelector with AdmissionCheck PodSetUpdates. The stale kubernetes.io/hostname constrains node filtering to a single node
(the old assignment), causing the workload to be permanently stuck in Pending — it can never be re-admitted because the target node may no longer exist or have capacity.

Fix: After merging the combined nodeSelector in findTopologyAssignment(), strip any keys that match TAS topology level keys (e.g., kubernetes.io/hostname, topology.kubernetes.io/zone). These keys are
always recomputed by TAS during topology assignment, so prior values must not constrain node filtering.

(No upstream issue yet — discovered in production. Can file one if needed.)

 - Observed in production: CHM NodeStatusUnknown repair on GPU clusters force-deleted individual pods from multi-pod GPU inference StatefulSets managed by Kueue TAS. 86% of affected workloads (6/7) 
became permanently stuck.
 - The fix is minimal and safe: topology level keys are always recomputed by TAS during findTopologyAssignment(), so stripping them before node filtering has no effect on fresh admissions — only on 
re-admissions where stale data survived eviction.
 - Surviving ungated pods that are already running still retain the old nodeSelector — this is non-blocking since they're already scheduled and running on the correct node.

 TAS: Fix permanently stuck workloads after eviction by stripping stale topology nodeSelector keys (e.g., kubernetes.io/hostname) from PodSet templates during re-evaluation, allowing evicted workloads to 
be re-admitted to any valid topology domain.
```

### DIFF
--- a/pkg/cache/scheduler/tas_cache_test.go
+++ b/pkg/cache/scheduler/tas_cache_test.go
@@ -1655,6 +1655,36 @@ func TestFindTopologyAssignments(t *testing.T) {
 				wantReason: `topology "default" doesn't allow to fit any of 1 pod(s). Total nodes: 4; excluded: nodeSelector: 2, resource "cpu": 1, taint "key=value:NoSchedule": 1`,
 			}},
 		},
+		"stale topology nodeSelector from prior admission is stripped; pod still schedules": {
+			nodes: defaultNodes,
+			levels: defaultThreeLevels,
+			podSets: []PodSetTestCase{{
+				topologyRequest: &kueue.PodSetTopologyRequest{
+					Required: ptr.To(corev1.LabelHostname),
+				},
+				requests: resources.Requests{
+					corev1.ResourceCPU: 1000,
+				},
+				// Simulate stale nodeSelector from a prior TAS admission that
+				// injected kubernetes.io/hostname. This should be stripped by the
+				// scheduler so it does not constrain node filtering.
+				nodeSelector: map[string]string{
+					corev1.LabelHostname: "deleted-node",
+				},
+				count: 1,
+				wantAssignment: &tas.TopologyAssignment{
+					Levels: defaultOneLevel,
+					Domains: []tas.TopologyDomainAssignment{
+						{
+							Count: 1,
+							Values: []string{
+								"x3",
+							},
+						},
+					},
+				},
+			}},
+		},
 		"resource exclusion picks most restrictive resource": {
 			nodes: []corev1.Node{
 				*testingnode.MakeNode("dual-shortage").
--- a/pkg/cache/scheduler/tas_flavor_snapshot.go
+++ b/pkg/cache/scheduler/tas_flavor_snapshot.go
@@ -792,6 +792,17 @@ func (s *TASFlavorSnapshot) findTopologyAssignment(
 	}
 	podSetTolerations := info.Tolerations
 	podSetNodeSelectors := info.NodeSelector
+
+	// Strip any nodeSelector keys that match topology level keys. These may
+	// be stale values from a prior TAS admission (e.g., kubernetes.io/hostname
+	// injected by the topology ungater) that survived eviction because the
+	// PodSet template was rebuilt from live pods carrying the old nodeSelector.
+	// TAS will compute its own topology assignment, so these keys must not
+	// constrain node filtering.
+	for _, levelKey := range s.levelKeys {
+		delete(podSetNodeSelectors, levelKey)
+	}
+
 	count := workersTasPodSetRequests.Count
 
 	// If slice topology is not requested then we can assume that slice is a single pod


mimowo:
@abhi179 thank you for the PR, I admit there is likely a bug, but let me ask what is the integration for which you experienced the issue: Job, JobSet, LWS? It may matter because different jobs have different lifecycle of Pods.

I'm asking because the proposed fix has a drawback iiuc, that we would ignore the original nodeSelector if this targets a specific node. And I already saw on prod examples of such jobs.

So, the proper solution I'm thinking about, but not sure if feasible, is to fully revert the Job PodTemplate on eviction, rather than leaving it stale, and dealing with the stale state later.

So I think we should start by a repro based on a specific integration, and figure out why the nodeSelector does not get restored.

abhi179:
> @abhi179 thank you for the PR, I admit there is likely a bug, but let me ask what is the integration for which you experienced the issue: Job, JobSet, LWS? It may matter because different jobs have different lifecycle of Pods.
> 
> I'm asking because the proposed fix has a drawback iiuc, that we would ignore the original nodeSelector if this targets a specific node. And I already saw on prod examples of such jobs.
> 
> So, the proper solution I'm thinking about, but not sure if feasible, is to fully revert the Job PodTemplate on eviction, rather than leaving it stale, and dealing with the stale state later.
> 
> So I think we should start by a repro based on a specific integration, and figure out why the nodeSelector does not get restored.



> @abhi179 thank you for the PR, I admit there is likely a bug, but let me ask what is the integration for which you experienced the issue: Job, JobSet, LWS? It may matter because different jobs have different lifecycle of Pods.
> 
> I'm asking because the proposed fix has a drawback iiuc, that we would ignore the original nodeSelector if this targets a specific node. And I already saw on prod examples of such jobs.
> 
> So, the proper solution I'm thinking about, but not sure if feasible, is to fully revert the Job PodTemplate on eviction, rather than leaving it stale, and dealing with the stale state later.
> 
> So I think we should start by a repro based on a specific integration, and figure out why the nodeSelector does not get restored.

@mimowo Thanks for the detailed feedback. Let me provide the full context:
 
 Integration: StatefulSet (GPU inference workloads with TAS topology-aware scheduling). Each pod is a full-node pod (one pod per GPU node).
 
 Repro scenario:
 
 Consider a StatefulSet with 2 replicas using TAS with kubernetes.io/hostname as the lowest topology level:
 
  //After initial admission, topology ungater (topology_ungater.go:267-270) injects per-pod hostname:
  Pod-0: nodeSelector = { "kubernetes.io/hostname": "node-A" }   # full GPU node
  Pod-1: nodeSelector = { "kubernetes.io/hostname": "node-B" }   # full GPU node
 
 
  1. An external component (CHM - Cluster Health Manager) force-deletes Pod-0 due to a node issue.
  2. Pod-1 survives — still has kubernetes.io/hostname: node-B in its nodeSelector (injected by ungater at admission).
  3. PodsReadyTimeout triggers eviction → workload is re-queued.
  4. On workload reconstruction, constructPodSet() (pod_controller.go:704-709) does *p.Spec.DeepCopy() from the surviving Pod-1.
  5. The reconstructed PodSet template now has kubernetes.io/hostname: node-B baked in — for all pods.
  6. On re-admission, TAS tries to schedule both pods constrained to node-B, but these are full-node pods — only 1 fits → permanently stuck.
 
 Why nodeSelector doesn't get restored: The topology ungater writes directly into pod.Spec.NodeSelector via maps.Copy() at admission time. There is no cleanup path that strips these TAS-injected keys from surviving pods on eviction. And SanitizePodSet() only
 sanitizes env vars, not nodeSelector — so stale labels survive into the new workload.

mimowo:
I see, thank you for explaining 👍 

So my first instinct is to try solving this at "(3.) PodsReadyTimeout triggers eviction → workload is re-queued." and revert the nodeSelector as it was before, because at this point the workload should still have the "clean" PodSets without that selector, right? So I expect the bug is likely somewhere around RestorePodSetsInfo that for the Pod integration it does not restore the NodeSelector properly.

The second alternative that comes to my mind is that our StatefulSet integration could trigger deletion of the stale Pods on Workload eviction, to make sure the re-admission will happen on new Pods. We could probably narrow this down to the case "when needed" - only when TAS is used, and we see the workload contains already some NodeSelector that could remain stale.

mimowo:
@abhi179 something just occurred to me, could you double check if the problem exists on a version which includes this PR: https://github.com/kubernetes-sigs/kueue/pull/9344, ie 0.17.0, or v0.16.2+. 

The reason I think this PR is likely to fix the problem is because since this PR we create the Workload for StatefulSet based of the StatefulSet PodTemplate rather than the Pods themselves, the relevant change would be here: https://github.com/kubernetes-sigs/kueue/pull/9344/changes#diff-0872cab5dfa1c3ae9de3b77cfdf379a370a7fb28ff114e38ccf9d22e411105f3R242-R260

It would be great if that helps.

abhi179:
> @abhi179 something just occurred to me, could you double check if the problem exists on a version which includes this PR: #9344, ie 0.17.0, or v0.16.2+.
> 
> The reason I think this PR is likely to fix the problem is because since this PR we create the Workload for StatefulSet based of the StatefulSet PodTemplate rather than the Pods themselves, the relevant change would be here: https://github.com/kubernetes-sigs/kueue/pull/9344/changes#diff-0872cab5dfa1c3ae9de3b77cfdf379a370a7fb28ff114e38ccf9d22e411105f3R242-R260

Thankyou @mimowo  ! I think this fix very likely resolves the rootcause of the issue. Recreating from original PodTemplate rather then copying from other live pod replica. Thanks. i will close this issue as resolved. i am assuming version 0.17.0 has the fix.

PR #10463, opened 2026-04-14
Trigger failure recovery reconciliations for unreachable nodes

kshalot:
#### What type of PR is this?
/kind bug

If the node is not tainted at the moment the timeout elapses, the pod won't be forcefully terminated and the reconcile will never be queued again, leading to the pod being stuck.

Fixes #

```release-note
FailureRecoveryPolicy: Fixed an issue where pods could remain stuck terminating if their node became unreachable only after the force-termination timeout had already elapsed.
```

### DIFF
--- a/pkg/controller/failurerecovery/pod_termination_controller.go
+++ b/pkg/controller/failurerecovery/pod_termination_controller.go
@@ -241,6 +241,9 @@ func (r *TerminatingPodReconciler) mapNodeToPods(ctx context.Context, node *core
 			})
 		}
 	}
+
+	log.V(4).Info("Reconciling pods affected by unreachable node", "pod_count", len(requests))
+
 	return requests
 }
 
--- a/pkg/controller/failurerecovery/pod_termination_controller.go
+++ b/pkg/controller/failurerecovery/pod_termination_controller.go
@@ -242,7 +242,9 @@ func (r *TerminatingPodReconciler) mapNodeToPods(ctx context.Context, node *core
 		}
 	}
 
-	log.V(4).Info("Reconciling pods affected by unreachable node", "pod_count", len(requests))
+	if len(requests) > 0 {
+		log.V(4).Info("Reconciling pods affected by unreachable node", "pod_count", len(requests))
+	}
 
 	return requests
 }
--- a/pkg/controller/failurerecovery/pod_termination_controller.go
+++ b/pkg/controller/failurerecovery/pod_termination_controller.go
@@ -250,7 +250,8 @@ func (r *TerminatingPodReconciler) nodeEventsPredicate() predicate.TypedPredicat
 			return utiltaints.TaintKeyExists(e.Object.Spec.Taints, corev1.TaintNodeUnreachable)
 		},
 		UpdateFunc: func(e event.TypedUpdateEvent[*corev1.Node]) bool {
-			return utiltaints.TaintKeyExists(e.ObjectNew.Spec.Taints, corev1.TaintNodeUnreachable)
+			return !utiltaints.TaintKeyExists(e.ObjectOld.Spec.Taints, corev1.TaintNodeUnreachable) &&
+				utiltaints.TaintKeyExists(e.ObjectNew.Spec.Taints, corev1.TaintNodeUnreachable)
 		},
 		DeleteFunc: func(e event.TypedDeleteEvent[*corev1.Node]) bool {
 			return false
--- a/pkg/controller/failurerecovery/pod_termination_controller.go
+++ b/pkg/controller/failurerecovery/pod_termination_controller.go
@@ -33,11 +33,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	configapi "sigs.k8s.io/kueue/apis/config/v1beta2"
 	"sigs.k8s.io/kueue/pkg/controller/constants"
 	"sigs.k8s.io/kueue/pkg/controller/core"
+	"sigs.k8s.io/kueue/pkg/controller/tas/indexer"
 	utilclient "sigs.k8s.io/kueue/pkg/util/client"
 	utilpod "sigs.k8s.io/kueue/pkg/util/pod"
 	"sigs.k8s.io/kueue/pkg/util/roletracker"
@@ -123,16 +125,7 @@ func (r *TerminatingPodReconciler) Create(e event.TypedCreateEvent[*corev1.Pod])
 }
 
 func (r *TerminatingPodReconciler) Update(u event.TypedUpdateEvent[*corev1.Pod]) bool {
-	if !podEligibleForTermination(u.ObjectNew) {
-		return false
-	}
-
-	// Pod was not marked for deletion in the update
-	if !u.ObjectOld.DeletionTimestamp.IsZero() || u.ObjectNew.DeletionTimestamp.IsZero() {
-		return false
-	}
-
-	return true
+	return podEligibleForTermination(u.ObjectNew)
 }
 
 func (r *TerminatingPodReconciler) Delete(event.TypedDeleteEvent[*corev1.Pod]) bool {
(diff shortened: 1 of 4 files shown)


mimowo:
/release-note-edit
```release-note
FailureRecoveryPolicy: Fixed an issue where pods could remain stuck terminating if their node became unreachable only after the force-termination timeout had already elapsed.
```

pkg/controller/failurerecovery/pod_termination_controller.go
```diff
@@ -228,6 +221,46 @@ func isPodPartiallyForcefullyTerminated(p *corev1.Pod) bool {
...
+	return requests
+}
+
+func (r *TerminatingPodReconciler) nodeEventsPredicate() predicate.TypedPredicate[*corev1.Node] {
+	return predicate.TypedFuncs[*corev1.Node]{
+		CreateFunc: func(e event.TypedCreateEvent[*corev1.Node]) bool {
+			return utiltaints.TaintKeyExists(e.Object.Spec.Taints, corev1.TaintNodeUnreachable)
+		},
+		UpdateFunc: func(e event.TypedUpdateEvent[*corev1.Node]) bool {
+			return utiltaints.TaintKeyExists(e.ObjectNew.Spec.Taints, corev1.TaintNodeUnreachable)
```
mimowo: Here we only need to trigger the reconcile when the Node is actually transitioning to Unreachable, right?
So `utiltaints.TaintKeyExists(e.ObjectOld.Spec.Taints, corev1.TaintNodeUnreachable) &&`, wdyt?

test/integration/singlecluster/controller/failurerecovery/pod_termination_controller_test.go
```diff
@@ -72,6 +72,39 @@ var _ = ginkgo.Describe("Pod termination controller", ginkgo.Ordered, ginkgo.Con
 		}, forcefulTerminationCheckTimeout, util.Interval).Should(gomega.Succeed())
 	})
 
+	ginkgo.It("triggers reconciliation when the node becomes unreachable", func() {
```
mimowo:
```suggestion
	ginkgo.It("should forcefufully delete a pod once the hosting node becomes unreachable", func() {
```
to make it more intent-oriented, rather than technical

test/integration/singlecluster/controller/failurerecovery/pod_termination_controller_test.go
```diff
@@ -72,6 +72,39 @@ var _ = ginkgo.Describe("Pod termination controller", ginkgo.Ordered, ginkgo.Con
...
+				g.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: pod.Name, Namespace: pod.Namespace}, pod)).
+					To(utiltesting.BeNotFoundError())
+			}, forcefulTerminationCheckTimeout, util.Interval).Should(gomega.Succeed())
+		})
+		ginkgo.By("removing the unreachable taint from the node", func() {
+			node := &corev1.Node{}
+			gomega.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: reachableNodeName}, node)).To(gomega.Succeed())
+			node.Spec.Taints = nil
+			gomega.Expect(k8sClient.Update(ctx, node)).To(gomega.Succeed())
+		})
```
mimowo: Actually, instead of using BeforeAll and reverting the nodes I would prefer to create those which are relevant per scenario, and then just delete. This makes the tests easier to run in parallel, which is gain for our runtime. 

Since this is pre-existing I'm fine as is and refactor later, at the same time it does not seem like much work yet, so I'm considering it could go in this PR. Let me know what you think.

test/integration/singlecluster/controller/failurerecovery/pod_termination_controller_test.go
```diff
@@ -72,6 +72,39 @@ var _ = ginkgo.Describe("Pod termination controller", ginkgo.Ordered, ginkgo.Con
...
+			node := &corev1.Node{}
+			gomega.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: reachableNodeName}, node)).To(gomega.Succeed())
+			node.Spec.Taints = append(node.Spec.Taints, corev1.Taint{Key: corev1.TaintNodeUnreachable, Effect: corev1.TaintEffectNoSchedule})
+			gomega.Expect(k8sClient.Update(ctx, node)).To(gomega.Succeed())
+		})
+		ginkgo.By("verifying that the pod is forcefully terminated after the timeout elapses", func() {
+			gomega.Eventually(func(g gomega.Gomega) {
+				g.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: pod.Name, Namespace: pod.Namespace}, pod)).
+					To(utiltesting.BeNotFoundError())
+			}, forcefulTerminationCheckTimeout, util.Interval).Should(gomega.Succeed())
```
mimowo: iiuc this timeout is not relevant here, we should just check using util.Timeout. It was relevant for the Consistency only - waited above.

pkg/controller/failurerecovery/pod_termination_controller.go
```diff
@@ -228,6 +221,46 @@ func isPodPartiallyForcefullyTerminated(p *corev1.Pod) bool {
 	})
 }
 
+func (r *TerminatingPodReconciler) mapNodeToPods(ctx context.Context, node *corev1.Node) []ctrl.Request {
+	log := log.FromContext(ctx)
```
mimowo, line 225: Either move under the List for error handling, or use to log the number of Pods eligible for termination. No strong view, but I would probably log the number of pods / requests as V4

mimowo (commented):
LGTM overall, thanks 👍

pkg/controller/failurerecovery/pod_termination_controller.go
kshalot: This is what I wrote initially, but then I thought that it's the same case as this check which we decided we want to get rid of:
```go
	// Pod was not marked for deletion in the update
	if !u.ObjectOld.DeletionTimestamp.IsZero() || u.ObjectNew.DeletionTimestamp.IsZero() {
		return false
	}
```
I was on the fence about adding it, so since you are mentioning it I'll add it back for now. I'll revert if we decide otherwise.

mimowo: I see, my main issue with the snippet above is readability. I wouldn't mind `!podEligible(oldPod) && podEligible(newPod)` this makes it very clear we are checking the transition.

test/integration/singlecluster/controller/failurerecovery/pod_termination_controller_test.go
kshalot: We can reduce the timeout here, yes.

pkg/controller/failurerecovery/pod_termination_controller.go
kshalot: I added the log. It's always logged atm, but I'm wondering whether we want to chech len() > 0

test/integration/singlecluster/controller/failurerecovery/pod_termination_controller_test.go
kshalot: I see I see. I can add the change in this PR later today, but if it happens to get merged in the meantime I'll make it a followup.

mimowo: Nah, I will merge tomorrow anyway

pkg/controller/failurerecovery/pod_termination_controller.go
mimowo: Yes, we can check len() > 0

pkg/controller/failurerecovery/pod_termination_controller.go
```diff
@@ -228,6 +221,50 @@ func isPodPartiallyForcefullyTerminated(p *corev1.Pod) bool {
...
+
+	log.V(4).Info("Reconciling pods affected by unreachable node", "pod_count", len(requests))
+
+	return requests
+}
+
+func (r *TerminatingPodReconciler) nodeEventsPredicate() predicate.TypedPredicate[*corev1.Node] {
+	return predicate.TypedFuncs[*corev1.Node]{
+		CreateFunc: func(e event.TypedCreateEvent[*corev1.Node]) bool {
+			return utiltaints.TaintKeyExists(e.Object.Spec.Taints, corev1.TaintNodeUnreachable)
```
mimowo, line 255: I think the code here is correct, but not covered with a test and so it could be overlooked one day as not relevant, as seemingly this scenario is impossible, but in fact it may happen if Kueue is restarted in the meanwhile. I'm wondering we it would be good to add an integration tests, where we restart Kueue manager during the test, I think this could be just a slight modification of the new test case.

kshalot: I agree with you, I'll see how to test it without limiting our ability to run the test in parallel. I think we can just label it with `ginkgo.Serial`.

test/integration/singlecluster/controller/failurerecovery/pod_termination_controller_test.go
```diff
@@ -43,26 +41,24 @@ var _ = ginkgo.Describe("Pod termination controller", ginkgo.Ordered, ginkgo.Con
...
 			StatusPhase(corev1.PodPending).
 			TerminationGracePeriod(1).
-			NodeName(unreachableNodeName).
 			Annotation(constants.SafeToForcefullyDeleteAnnotationKey, constants.SafeToForcefullyDeleteAnnotationValue)
 	})
 
-	ginkgo.It("forcefully terminates pods that opt-in, scheduled on unreachable nodes", func() {
-		matchingPod := matchingPodWrapper.Clone().Obj()
+	ginkgo.It("should forcefully terminate pods that opt-in, scheduled on unreachable nodes", func() {
+		unreachableNode := testingnode.MakeNode("unreachable-node").
```
mimowo, line 53: Can you also add the ns suffix, and remove the ` ginkgo.Ordered,` from the describe?

pkg/controller/failurerecovery/pod_termination_controller.go
mimowo: Ah good point, I'm not sure about running such a test in parallel, requires a check, but maybe not needed, because for integration tests IIRC each set of tests will run on a dedicated API server, and dedicated Kueue manager.

test/integration/singlecluster/controller/failurerecovery/pod_termination_controller_test.go
mimowo: Ah, this is related to https://github.com/kubernetes-sigs/kueue/pull/10463#discussion_r3085272581, and what I suggested may actually not be needed to run the tests in parallel, because iirc the tests will anyway have their own API server and Kueue manager, but this is worth double-checking.

kshalot: Good point. I already made the change, but I'll double check and revert if it's not needed, no point in keeping it.

kshalot: Yeah it should be isolated anyway. I removed the node namespacing to reduce noise.

pkg/controller/failurerecovery/pod_termination_controller.go
kshalot: https://github.com/kubernetes-sigs/kueue/pull/10463/commits/2774c2b5ab274e2b03728ce1296101e19364ae8f

pkg/controller/failurerecovery/pod_termination_controller.go
```diff
@@ -123,16 +125,7 @@ func (r *TerminatingPodReconciler) Create(e event.TypedCreateEvent[*corev1.Pod])
...
-		return false
-	}
-
-	// Pod was not marked for deletion in the update
-	if !u.ObjectOld.DeletionTimestamp.IsZero() || u.ObjectNew.DeletionTimestamp.IsZero() {
-		return false
-	}
-
-	return true
+	return podEligibleForTermination(u.ObjectNew)
```
mimowo: Should we actually check for the transition  ` podEligibleForTermination(u.ObjectNew) && !podEligibleForTermination(u.ObjectOld)`, wdyt? This seems natural and I cannot see any counter example.

mimowo:
Thanks 👍 
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16

kshalot:
@mimowo This also warrants a KEP update, I'll post a PR later today.

PR #10478, opened 2026-04-14
Fix TAS empty slices for count=0 podSets causing infinite scheduling loop

jzhaojieh:

When a workload with completed podSets (count=0 after reclaim) is preempted and re-queued, FindTopologyAssignmentsForFlavor produces a non-nil TopologyAssignment with empty Domains for the count=0 podSets. This converts to empty Slices in the API format, which CRD validation rejects with "topologyAssignment.slices: Required value", causing an infinite scheduling loop with no backoff.

Skip count=0 podSets in WorkloadsTopologyRequests. PodSets with 0 pods need no topology assignment

N/A

```release-note
TAS: Fix empty slices for count=0 podSets causing infinite scheduling loop
```

### DIFF
--- a/pkg/scheduler/flavorassigner/tas_flavorassigner.go
+++ b/pkg/scheduler/flavorassigner/tas_flavorassigner.go
@@ -23,6 +23,7 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/utils/ptr"
@@ -36,13 +37,17 @@ import (
 )
 
 // WorkloadsTopologyRequests - returns the TopologyRequests of the workload
-func (a *Assignment) WorkloadsTopologyRequests(wl *workload.Info, cq *schdcache.ClusterQueueSnapshot) schdcache.WorkloadTASRequests {
+func (a *Assignment) WorkloadsTopologyRequests(log logr.Logger, wl *workload.Info, cq *schdcache.ClusterQueueSnapshot) schdcache.WorkloadTASRequests {
 	tasRequests := make(schdcache.WorkloadTASRequests)
 	for i, podSet := range wl.Obj.Spec.PodSets {
 		if isTASRequested(&podSet, cq) {
 			psAssignment := a.podSetAssignmentByName(podSet.Name)
 			if psAssignment.Status.IsError() {
 				// There is no resource quota assignment for the PodSet - no need to check TAS.
+				continue
+			}
+			if psAssignment.Count == 0 {
+				log.V(3).Info("Skipping TAS for count=0 podSet", "podSet", podSet.Name)
 				continue
 			}
 			if psAssignment.TopologyAssignment != nil && !psAssignment.HasUnhealthyNode(wl) {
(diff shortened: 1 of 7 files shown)


pkg/scheduler/flavorassigner/tas_flavorassigner.go
```diff
@@ -45,6 +45,11 @@ func (a *Assignment) WorkloadsTopologyRequests(wl *workload.Info, cq *schdcache.
 				// There is no resource quota assignment for the PodSet - no need to check TAS.
 				continue
 			}
+			if psAssignment.Count == 0 {
+				// Skip podSets with count=0 (fully reclaimable/completed).
+				// No TAS assignment is needed for 0 pods.
+				continue
+			}
```
mimowo, line 52: LGTM, but as this function is growing I think it would be good to propagate the logger and call it here, wdyt? It can then be also useful for other places (but we can keep out of scope for the PR).

jzhaojieh: Makes sense, plumbed the logger here

mimowo:
> When a workload with completed podSets (count=0 after reclaim) is preempted and re-queued, FindTopologyAssignmentsForFlavor produces a non-nil TopologyAssignment with empty Domains for the count=0 podSets. This converts to empty Slices in the API format, which CRD validation rejects with "topologyAssignment.slices: Required value", causing an infinite scheduling loop with no backoff.

That feels like a complex scenario, which Im not fully following. Questions that I have
1. is the workload created with podSets.count=0?
2. is the preemption / reclamation relevant here? What makes the difference between the first admission and the repeated admission
3. how does the infinite loop break now, does the workload get re admitted with podSets.count=0?

Would you like to put together an integration test for the scenario? It would be great

jzhaojieh:
> 1. is the workload created with podSets.count=0?

No the count is decremented to 0 when 1 of the replicated jobs completes

> 2. is the preemption / reclamation relevant here? What makes the difference between the first admission and the repeated admission

yea preemption triggers the rescheduling based on the state where some of the jobs have completed already to trigger the bug

> 3. how does the infinite loop break now, does the workload get re admitted with podSets.count=0?

Manual remediation is needed to cleanup the workload 

> Would you like to put together an integration test for the scenario? It would be great

Ive added an integration test for this scenario, the exact case that TAS doesnt handle well today is

1. JobSet is created with 2 replicated jobs (A,B)
2. A completes but B is still running
3. JobSet is preempted and TAS attempts to reschedule
4. Since A has completed, TAS generates an empty topology assignment which fails validation against the API server
5. The reschedule and validation failure loop continues indefinitely until the JobSet is deleted

mimowo:
/kind bug
based on the release note and the description

mimowo (commented):
Thanks 👍 
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16

mimowo:
Can you also prepare a CP for 0.16?

tenzen-y:
/release-note-edit

```release-note
TAS: Fix empty slices for count=0 podSets causing infinite scheduling loop
```

mimowo:
Opened that PR  for 0.16 https://github.com/kubernetes-sigs/kueue/pull/10510

PR #10480, opened 2026-04-14
TAS: Handle cordoned nodes as immediate failures for workload replacement

tg123:
#### What type of PR is this?

/kind feature
/area tas

This PR enhances TAS node failure handling to immediately mark cordoned (unschedulable) nodes as failed, triggering workload replacement without waiting for the `NodeFailureDelay` timer.

Changes:
- add an early check for `node.Spec.Unschedulable` before the not-ready timer logic. When a node is cordoned, it is immediately treated as unhealthy and affected workloads are handled right away.

Previously, cordoning a node would only be detected if the node also became not-ready, and even then it would be subject to the `NodeFailureDelay` wait. With this change, cordoned nodes are handled as an explicit, immediate failure signal.

Fixes #

The unschedulable check is placed before the not-ready timer block so that a cordoned-but-not-ready node doesn't unnecessarily wait for the delay to expire.

Let me know if you think this feature is ok to go or need more discussion — I will add tests and documentation after proposal accepted.

```release-note
TAS: Immediately mark cordoned (unschedulable) nodes as failed and trigger workload replacement without waiting for NodeFailureDelay.
```

### DIFF
--- a/pkg/controller/tas/node_controller.go
+++ b/pkg/controller/tas/node_controller.go
@@ -145,6 +145,21 @@ func (r *nodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		readyCondition = utiltas.GetNodeCondition(&node, corev1.NodeReady)
 	}
 
+	// Cordoned (Unschedulable) — mark unhealthy immediately.
+	if nodeExists && node.Spec.Unschedulable {
+		log.V(3).Info("Node is cordoned (Unschedulable), marking as failed")
+		nodeSelectorPodsByWorkload, err := r.listPodsAssignedByNodeSelector(ctx, req.Name)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		affectedWorkloads, err := r.getWorkloadsOnNode(ctx, req.Name, nodeSelectorPodsByWorkload)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		_, err = r.handleUnhealthyNode(ctx, req.Name, affectedWorkloads)
+		return ctrl.Result{}, err
+	}
+
 	var timerExpired bool
 	if readyCondition != nil && readyCondition.Status != corev1.ConditionTrue {
 		timeSinceNotReady := r.clock.Now().Sub(readyCondition.LastTransitionTime.Time)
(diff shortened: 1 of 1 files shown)


tg123:
This PR is more of a proposal. It seems the current implementation may end up draining the node, which is not the intended behavior.

j-skiba:
This has already been defined and written down in a KEP. Here's the KEP for this feature - https://github.com/kubernetes-sigs/kueue/tree/main/keps/2724-topology-aware-scheduling#tainted-nodes-treatment - which states:
```
Nodes with .spec.unschedulable set to true are treated as having the NoSchedule taint.
```
IMO it is more consistent to treat the `unschedulable` nodes the same way as nodes with `NoSchedule` taint as the pods might still be healthy. Also kubernetes core doesn't evict such pods from nodes so keeping the workload would be consistent with kubernetes core behavior.

cc @mimowo

mimowo:
> IMO it is more consistent to treat the unschedulable nodes the same way as nodes with NoSchedule taint as the pods might still be healthy. 

Yes, this matches my understanding - the handling path should be the same as for NoSchedule, because k8s core does not evict Pods. Also, IIUC cordoning a node already sets both spec.schedulable=false, and adds the NoSchedule taint.

 @tg123  Could you check if this matches your expectations using the 0.17+ version, or enabling the `TASReplaceNodeOnNodeTaints` feature gate on 0.15+.

tg123:
Thanks i am on 0.16
The scenario for us is that we found some bad node and cordoned them, so wish the pod to be re-scheduled somewhere else asap. 

however, by deleting pod, it does not help, tas keep setting the nodeselector to old bad node until node restart and node became notready

tg123:
@mimowo @j-skiba may i know what is the policy?
during my test, seems 2nd bad node arrives before the 1st replacement completes would trigger `Full workload eviction` `EvictedDueToNodeFailures`?

mimowo:
> @mimowo @j-skiba may i know what is the policy?
> during my test, seems 2nd bad node arrives before the 1st replacement completes would trigger `Full workload eviction` `EvictedDueToNodeFailures`?

Yes, this is how Node Hot Swap currently works - we only support replacement for a single Node. When two nodes fail at the same time, when we trigger eviction of the entire workload. 

However, from the beginning we considered multiple node replacements, and thus we introduced the generic API already: UnhealthyNodes. So, I'm totally ok to try replacing more than one node, but it would need to be careful, because I remember that was tricky algorithmically.

tg123:
a bit confused, (sorry did not read into code), why not just do 1-node host-swap at a time ? no fancy algo required any more

PR #10482, opened 2026-04-14
add aggregate-to-view label to read-only cluster roles

### LINKED ISSUE #10481: Add aggregate-to-view label to Kueue Read-Only ClusterRoles
Add `rbac.authorization.k8s.io/aggregate-to-view: "true"` label to Kueue ClusterRoles. This allows Kueue CRDs to be aggregated to the existing k8s `view` role. 

**Why is this needed**:
Getting access for Kueue CRDs is out of band vs just aggregating it to the existing view only k8s role that comes with each k8s release: https://github.com/kubernetes/kubernetes/blob/v1.35.3/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/testdata/cluster-roles.yaml#L1428

You can also see other sig projects with similar practices: https://github.com/kubernetes-sigs/metrics-server/blob/master/charts/metrics-server/templates/clusterrole-aggregated-reader.yaml

amy:
#### What type of PR is this?
/kind cleanup 

Add 
- `rbac.authorization.k8s.io/aggregate-to-view: "true"`
- `rbac.authorization.k8s.io/aggregate-to-admin: "true"`
- `rbac.authorization.k8s.io/aggregate-to-edit: "true"` 

labels to Kueue ClusterRoles. This allows Kueue CRDs read-only permissions to be aggregated to the default k8s `view`, `admin`, and `edit` roles that's provided with each k8s release.

Getting access for Kueue CRDs is out of band vs just aggregating it to the existing view only k8s role that comes with each k8s release: https://github.com/kubernetes/kubernetes/blob/v1.35.3/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/testdata/cluster-roles.yaml#L1428

You can also see other sig projects with similar practices: https://github.com/kubernetes-sigs/metrics-server/blob/master/charts/metrics-server/templates/clusterrole-aggregated-reader.yaml

Fixes #10481

```release-note
Aggregate Kueue CRD read-only clusterRoles to k8s default view clusterRole
```

### DIFF
--- a/charts/kueue/templates/rbac/clusterqueue_viewer_role.yaml
+++ b/charts/kueue/templates/rbac/clusterqueue_viewer_role.yaml
@@ -23,6 +23,9 @@ metadata:
   labels:
   {{- include "kueue.labels" . | nindent 4 }}
     rbac.kueue.x-k8s.io/batch-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
 rules:
   - apiGroups:
       - kueue.x-k8s.io
--- a/charts/kueue/templates/rbac/cohort_viewer_role.yaml
+++ b/charts/kueue/templates/rbac/cohort_viewer_role.yaml
@@ -23,6 +23,9 @@ metadata:
   labels:
   {{- include "kueue.labels" . | nindent 4 }}
     rbac.kueue.x-k8s.io/batch-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
 rules:
   - apiGroups:
       - kueue.x-k8s.io
--- a/charts/kueue/templates/rbac/resourceflavor_viewer_role.yaml
+++ b/charts/kueue/templates/rbac/resourceflavor_viewer_role.yaml
@@ -23,6 +23,9 @@ metadata:
   labels:
   {{- include "kueue.labels" . | nindent 4 }}
     rbac.kueue.x-k8s.io/batch-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
 rules:
   - apiGroups:
       - kueue.x-k8s.io
--- a/charts/kueue/templates/rbac/topology_viewer_role.yaml
+++ b/charts/kueue/templates/rbac/topology_viewer_role.yaml
@@ -23,6 +23,9 @@ metadata:
   labels:
   {{- include "kueue.labels" . | nindent 4 }}
     rbac.kueue.x-k8s.io/batch-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
 rules:
   - apiGroups:
       - kueue.x-k8s.io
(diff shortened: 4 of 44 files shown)


kannon92 (commented):
/lgtm

tenzen-y (commented):
Thank you!
/approve

charts/kueue/templates/rbac/clusterqueue_viewer_role.yaml
```diff
@@ -23,6 +23,9 @@ metadata:
   labels:
   {{- include "kueue.labels" . | nindent 4 }}
     rbac.kueue.x-k8s.io/batch-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
```
tenzen-y, line 28:
```suggestion
    rbac.authorization.k8s.io/aggregate-to-admin: "true"
```

Uhm, after I check this PR again, we probably should drop the read capability from the viewer and editor in cluster-scoped resources like ClusterQueue / Cohort / Topology / ResourceFlavors.

cc @mimowo

charts/kueue/templates/rbac/cohort_viewer_role.yaml
```diff
@@ -23,6 +23,9 @@ metadata:
   labels:
   {{- include "kueue.labels" . | nindent 4 }}
     rbac.kueue.x-k8s.io/batch-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
```
tenzen-y, line 28:
```suggestion
    rbac.authorization.k8s.io/aggregate-to-admin: "true"
```

charts/kueue/templates/rbac/resourceflavor_viewer_role.yaml
```diff
@@ -23,6 +23,9 @@ metadata:
   labels:
   {{- include "kueue.labels" . | nindent 4 }}
     rbac.kueue.x-k8s.io/batch-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
```
tenzen-y, line 28:
```suggestion
    rbac.authorization.k8s.io/aggregate-to-admin: "true"
```

charts/kueue/templates/rbac/topology_viewer_role.yaml
```diff
@@ -23,6 +23,9 @@ metadata:
   labels:
   {{- include "kueue.labels" . | nindent 4 }}
     rbac.kueue.x-k8s.io/batch-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
```
tenzen-y, line 28:
```suggestion
    rbac.authorization.k8s.io/aggregate-to-admin: "true"
```

charts/kueue/templates/rbac/clusterqueue_viewer_role.yaml
mimowo: Indeed, we intentionally don't give viewer rights in our current RBAC configuration for Cluster-scoped objects. 

The most sensitive are ClusterQueues, because they contain quota configuration in multi-tenant environments, and some platforms / admins may prefer to keep it confidential how much quota the other tenant has.

The Topology and ResourceFlavor are less sensitive, but still I would keep them hidden by default.

Also, for "edit" the situation might be less clear, because currently we only distinguish batch user and batch admin. I'm not sure if "edit" is more like "user" or "admin", wdyt @tenzen-y @amy ?

In any case, to be "safe by default" I would suggest to just remove the view and edit aggregations.

cc @amy

tenzen-y: That sounds reasonable to me. @mimowo 's policy is exactly the same as mine.
I opened PR: https://github.com/kubernetes-sigs/kueue/pull/10519

amy: ah yeah. Keeping it to just the admin aggregation makes sense. Thanks for making the PR!

PR #10485, opened 2026-04-15
Pending workload resource metric

### LINKED ISSUE #9741: Pending workload resource metric
Similar to `kueue_cluster_queue_resource_reservation`, but I want to track the resources pending in the CQ. Today we just count the number of pending workloads.

Call it `kueue_cluster_queue_resource_pending`.

**Why is this needed**:
Helps with instrumenting Kueue to alert when something is broken. You can take the diff of Pending - Unallocated Nominal Quota to roughly gauge whether there's a scheduling issue and investigation needs to happen. (ex: fragmentation, scheduling gates not initiated by Kueue, bugs, etc.) If there's enough unallocated nominal quota to admit the pending workloads, but you see that diff for a prolonged period of time (ie not a spike shape), there's potentially an issue.

**Completion requirements**:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

amy:
#### What type of PR is this?
/kind feature

Similar to kueue_cluster_queue_resource_reservation, but I want to track the resources pending in the CQ. Today we just count the number of pending workloads. Call it kueue_cluster_queue_resource_pending.

Helps with instrumenting Kueue to alert when something is broken. You can take the diff of Pending - Unallocated Nominal Quota to roughly gauge whether there's a scheduling issue and investigation needs to happen. (ex: fragmentation, scheduling gates not initiated by Kueue, bugs, etc.) If there's enough unallocated nominal quota to admit the pending workloads, but you see that diff for a prolonged period of time (ie not a spike shape), there's potentially an issue.

Fixes #9741

```release-note
New kueue_cluster_queue_resource_pending metric to show total resources pending.
```

---
> AI Assistance Disclosure: This pull request was created with the assistance of AI


pkg/cache/queue/cluster_queue.go
```diff
@@ -469,6 +490,36 @@ func (c *ClusterQueue) handleInadmissibleHash(hash string) int {
...
+		}
+	}
+	for _, wl := range c.inadmissibleWorkloads {
+		for _, ps := range wl.TotalRequests {
+			for name, q := range ps.Requests {
+				total[name] += q
+			}
+		}
+	}
+	if c.inflight != nil {
```
amy: Is this part correct? Is inflight considered as part of pending?

amy:
cc/ @mbobrovskyi could you take a review pass? 🙏

pkg/cache/queue/cluster_queue.go
```diff
@@ -136,6 +137,19 @@ type ClusterQueue struct {
...
+	// in heap and inadmissibleWorkloads (not inflight). Updated at each mutation site so
+	// PendingResources() is O(1) rather than O(N).
+	// Configured resources are seeded at 0 by Update() so they appear in metrics
+	// even when no workloads are pending; stale zero entries are pruned on Update().
+	pendingResourcesTotal map[corev1.ResourceName]int64
+
+	// inflightResources is a snapshot of the current inflight workload's TotalRequests,
+	// taken at Pop time under the exclusive lock before the scheduler receives the pointer.
+	// Avoids a data race with flavorassigner.assignFlavors, which writes to the same maps.
+	// Nil when no workload is inflight.
```
mbobrovskyi, line 151:
```suggestion
```

mbobrovskyi: Or should we remove the initialization of this map in newClusterQueueImpl()?

mbobrovskyi: I'm actually can't see where we checking for nil.

pkg/cache/queue/metrics.go
```diff
@@ -48,7 +50,21 @@ func reportCQPendingWorkloads(m *Manager, cq *ClusterQueue) {
...
+	metrics.ReportPendingWorkloads(cq.name, active, inadmissible, cqCustomLabels, m.roleTracker)
+
+	var cohort kueue.CohortReference
+	if cq.HasParent() {
+		cohort = cq.Parent().GetName()
+	}
+	// pendingResourcesTotal carries 0 entries for configured resources (seeded by
+	// Update), so iterating it once covers both the zero-series and actual pending.
+	pendingResources := cq.PendingResources()
+	metrics.ClearClusterQueueResourcePendingMetrics(string(cq.name))
```
mbobrovskyi: Actually, I’m not sure we need to clear it here. I think we should clear it if the cohort or CQ is deleted, right?

pkg/cache/queue/cluster_queue.go
amy: Removed the initialization of inflightResources. Keeping the Nil comment because that reflects current inflightresources behavior

InflightResources is used here: https://github.com/kubernetes-sigs/kueue/pull/10485/changes#diff-04ff27d9f9b97758172eb0674ca8040fe5ec7c2c3c427e68aadc679b075e2697R564
So if nil, the resources aren't counted.

mbobrovskyi:
Thank you!

/lgtm

amy:
PTAL @mimowo when you have time. PR is ready! 🙏

pkg/cache/queue/cluster_queue.go
```diff
@@ -250,6 +265,27 @@ func (c *ClusterQueue) Update(apiCQ *kueue.ClusterQueue) error {
...
+				}
+			}
+		}
+	}
+	// Prune zero entries for resources that were removed from the spec.
+	for r, v := range c.pendingResourcesTotal {
+		if v == 0 && !newConfigured.Has(r) {
+			delete(c.pendingResourcesTotal, r)
+		}
+	}
```
mimowo: Let's extract this code to a helper to limit the scope of the `newConfigured` variable.

pkg/metrics/metrics.go
```diff
@@ -693,6 +697,15 @@ For a LocalQueue, the metric only reports a value of 1 for one of the statuses.`
...
 	trackGaugeVec(ClusterQueueResourceReservations, gaugeCleanupScopeClusterQueueResource)
 
+	ClusterQueueResourcePending = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: constants.KueueName,
+			Name:      "cluster_queue_resource_pending",
+			Help:      `Reports the cluster_queue's total pending resource requests. Unlike resource_reservation, pending workloads have not yet been assigned to flavors.`,
+		}, append([]string{"cohort", "cluster_queue", "resource", "replica_role"}, extraLabels...),
+	)
+	trackGaugeVec(ClusterQueueResourcePending, gaugeCleanupScopeClusterQueue, gaugeCleanupScopeClusterQueueResource)
```
mimowo, line 731: Let's split this metric for only ClusterQueue first, and later we can follow up with the extension for the Cohorts. This is for the consistency of the approach, because other ClusterQueue metrics don't report the cohort label. The reason is that it is unclear if "cohort" means immediate parent or higher level ancestor. This feature is tracked here: https://github.com/kubernetes-sigs/kueue/issues/7539

pkg/cache/queue/cluster_queue.go
```diff
@@ -302,20 +339,29 @@ func (c *ClusterQueue) PushOrUpdate(wInfo *workload.Info) {
...
 	// (only for BestEffortFIFO; StrictFIFO preserves strict ordering).
 	if c.queueingStrategy == kueue.BestEffortFIFO && c.heap.GetByKey(key) == nil && wInfo.SchedulingHash != workload.SchedulingHashUnknown && c.noFitSchedulingHashes.Has(wInfo.SchedulingHash) {
 		c.inadmissibleWorkloads.insert(key, wInfo)
+		c.addPendingResources(wInfo)
 		return
 	}
+	// Subtract the old entry's resources before overwriting; requests may have changed.
+	if oldHeapInfo := c.heap.GetByKey(key); oldHeapInfo != nil {
+		c.subtractPendingResources(oldHeapInfo)
+	}
```
mimowo, line 378: Do we have an integration test covering this update scenario?

pkg/cache/queue/cluster_queue.go
```diff
@@ -302,20 +339,29 @@ func (c *ClusterQueue) PushOrUpdate(wInfo *workload.Info) {
 			c.inadmissibleWorkloads.insert(key, wInfo)
 			return
 		}
+		// Workload is leaving inadmissible; account for its resources before moving.
+		c.subtractPendingResources(oldInfo)
 		// otherwise move or update in place in the queue.
 		c.inadmissibleWorkloads.delete(key)
```
mimowo: I think it would be good to combine the two lines which always go together:
- `c.subtractPendingResources` with `c.inadmissibleWorkloads.delete`, this should also remove from the heap
- `c.addPendingResources` with `c.inadmissibleWorkloads.insert` (or heap.PushOrUpdate)

This way we can avoid duplication of injecting the pending resources management next to adding / deleting pending workloads.

pkg/cache/queue/cluster_queue.go
```diff
@@ -469,6 +559,19 @@ func (c *ClusterQueue) handleInadmissibleHash(hash string) int {
 	return moved
 }
 
+// PendingResources returns the total resources requested by all pending workloads,
+// aggregated by resource name. Pending workloads have not yet been assigned to flavors.
+func (c *ClusterQueue) PendingResources() map[corev1.ResourceName]int64 {
+	c.rwm.RLock()
+	defer c.rwm.RUnlock()
+	result := make(map[corev1.ResourceName]int64, len(c.pendingResourcesTotal)+len(c.inflightResources))
+	maps.Copy(result, c.pendingResourcesTotal)
```
mimowo: Why not just Clone the map of the pendingResourcesTotal? In most cases the resources in pendingResourcesTotal and inflightResources will be the same, so pre-allocation of the large (2x)  map feels weird.

pkg/cache/queue/cluster_queue.go
```diff
@@ -136,6 +137,19 @@ type ClusterQueue struct {
...
+	// PendingResources() is O(1) rather than O(N).
+	// Configured resources are seeded at 0 by Update() so they appear in metrics
+	// even when no workloads are pending; stale zero entries are pruned on Update().
+	pendingResourcesTotal map[corev1.ResourceName]int64
+
+	// inflightResources is a snapshot of the current inflight workload's TotalRequests,
+	// taken at Pop time under the exclusive lock before the scheduler receives the pointer.
+	// Avoids a data race with flavorassigner.assignFlavors, which writes to the same maps.
+	// Nil when no workload is inflight.
+	inflightResources map[corev1.ResourceName]int64
```
mimowo: I'm not sure why we need that. I can totally understand the need to aggregate the pending resources in pendingResourcesTotal, but since there is only one "inflight" workload it should be O(1) to get the pending resources for the workload without the explicit field.

I see this comment "Avoids a data race with flavorassigner.assignFlavors, which writes to the same maps. Nil when no workload is inflight.", but I don't understand the race at the moment, could you clarify step by step under what circumstances this may happen? 

My instinct is rather that if we have only one field "inflight" then it is less prone to race conditions as this is single source of truth, whereas with the new field we need to think: what if inflight=nil, but inflightResources is not nil. 

I'm not clear if such scenarios may happen and what they entail. Let's clarify that.

pkg/cache/queue/manager_test.go
```diff
@@ -307,6 +305,91 @@ func TestUpdateClusterQueueLabelsUpdated(t *testing.T) {
 	}
 }
 
+func TestPendingResourceMetrics(t *testing.T) {
```
mimowo, line 308: Instead of the single spagetti test let's refactor to have it table-driven test cases so that we can test multiple scenarios easily.

pkg/cache/queue/cluster_queue.go
```diff
@@ -445,6 +534,7 @@ func (c *ClusterQueue) requeueIfNotPresent(log logr.Logger, wInfo *workload.Info
 func (c *ClusterQueue) forgetInflightByKey(key workload.Reference) {
 	if c.inflight != nil && workload.Key(c.inflight.Obj) == key {
 		c.inflight = nil
+		c.inflightResources = nil
```
mimowo: As I mentioned in the other comment I don't understand currently why we introduce `inflightResources`, but if needed, then I'm wondering that we lack some kind of structure to encapsulate both `inflight` and `inflightResources`, to ensure they are both set and cleaned in-sync.

mimowo:
Thank you, I can totally see this metric helpful, but we need to clean up some details.

pkg/metrics/metrics.go
mimowo, line 731: Is this metric controller  (enabled/ disabled|) by [EnableClusterQueueResources](https://github.com/kubernetes-sigs/kueue/blob/main/apis/config/v1beta2/configuration_types.go#L184C2-L184C29). I think it should be, but it does not clearly follow from the code, so let;s ensure this is the case.

pkg/cache/queue/cluster_queue.go
amy: For a previous commit that's gone due to rebase + git push -f, I had a race detection failure for tests, so I added the snapshot as a reaction to that.

Goroutine A, scheduler (no lock)

Pop() returns c.inflight.info ([cluster_queue.go:669](vscode-webview://0oeb79kq8q1qn4q4pkdrkdrf9fpoas98r5f79a306j4av2l5rf40/pkg/cache/queue/cluster_queue.go#L669)), a *workload.Info. When assignFlavors runs with this pointer, requests = a.wl.TotalRequests ([flavorassigner.go:585](vscode-webview://0oeb79kq8q1qn4q4pkdrkdrf9fpoas98r5f79a306j4av2l5rf40/pkg/scheduler/flavorassigner/flavorassigner.go#L585)) shares the same underlying slice. The range loop then writes into podSet.Requests ([flavorassigner.go:608](vscode-webview://0oeb79kq8q1qn4q4pkdrkdrf9fpoas98r5f79a306j4av2l5rf40/pkg/scheduler/flavorassigner/flavorassigner.go#L608)), which is the same map as c.inflight.info.TotalRequests[i].Requests.

Goroutine B, metrics collector (under rwm.RLock())

Without the snapshot, PendingResources() would iterate c.inflight.info.TotalRequests[i].Requests directly — the same map Goroutine A is writing to, with no lock on either side. The race detector fires here.

pkg/cache/queue/cluster_queue.go
amy: done: https://github.com/kubernetes-sigs/kueue/pull/10485/changes/de69bccf2e355ebdda03d8e4aa1fce39af49ec3d#diff-04ff27d9f9b97758172eb0674ca8040fe5ec7c2c3c427e68aadc679b075e2697R279

pkg/metrics/metrics.go
amy: done: https://github.com/kubernetes-sigs/kueue/pull/10485/changes/de69bccf2e355ebdda03d8e4aa1fce39af49ec3d#diff-4bc91a74bc47b8467cdcd25a7e3fe1651dbe44907f478377eca31b982f444f23R705

pkg/cache/queue/cluster_queue.go
amy: done: https://github.com/kubernetes-sigs/kueue/pull/10485/changes/de69bccf2e355ebdda03d8e4aa1fce39af49ec3d#diff-10a8c388e1887a008412f124e62a51a46fbd935fcc5b35f0314d633e0be99353R446

pkg/cache/queue/cluster_queue.go
amy: done: https://github.com/kubernetes-sigs/kueue/pull/10485/changes/de69bccf2e355ebdda03d8e4aa1fce39af49ec3d#diff-04ff27d9f9b97758172eb0674ca8040fe5ec7c2c3c427e68aadc679b075e2697R427

Also, working on a review skill that generalizes your comments on my PRs: https://github.com/kubernetes-sigs/kueue/pull/10819

pkg/cache/queue/cluster_queue.go
amy: done: https://github.com/kubernetes-sigs/kueue/pull/10485/changes/de69bccf2e355ebdda03d8e4aa1fce39af49ec3d#diff-04ff27d9f9b97758172eb0674ca8040fe5ec7c2c3c427e68aadc679b075e2697R577

pkg/cache/queue/manager_test.go
amy: 😂🍝 done: https://github.com/kubernetes-sigs/kueue/pull/10485/changes/de69bccf2e355ebdda03d8e4aa1fce39af49ec3d#diff-2e1eeb5b2bed8f8a36ecb7f9245f11a4ef7046516076b03adf94f57a9dbcf86eR308

pkg/cache/queue/cluster_queue.go
amy: see: https://github.com/kubernetes-sigs/kueue/pull/10485#discussion_r3158311135

pkg/metrics/metrics.go
amy: done: https://github.com/kubernetes-sigs/kueue/pull/10485/changes/de69bccf2e355ebdda03d8e4aa1fce39af49ec3d#diff-6bbb2b918012d698df1926d0b61f656e6137f3ab991da863daaaf76824a5c4a8R56

amy:
@mimowo ready for another pass 🙇🏻‍♀️

ns-sundar:
Hi all, I have a basic question about the metric and its usage/usefulness.

The PR defines "pendingResourcesTotal is the incremental sum of TotalRequests across workloads".
Let's take an example. Say the queue's nominal quota for a resource is 100, resource reservation is 20 and so the unallocated ("remaining nominal") amount is 80. Now, say there are 3 jobs A, B and C with resource needs of 50, 40 and 10, resp.

So, pendingResourcesTotal is 100 (A+B+C). It is more than the unallocated amount of 80. 

So, if we are monitoring the condition `if (pendingResourcesTotal <= remaining nominal) then /* some scheduling issue */`, that would not trigger, and things would seem normal.

However, job A can be admitted. If job A is not admitted for a while, that could be a scheduling problem, but that is not flagged because the metric is a sum across pending workloads rather than per workload.

IOW, is it useful to be looking only at the aggregated resource count?
@mbobrovskyi @Amy @mimowo

amy:
@ns-sundar 
> is it useful to be looking only at the aggregated resource count

yes I've seen this used pretty successfully. 

That being said. I think another metric that can be added with more granularity that I mentioned to you is something like: "Head of queue pending resource". So that way you can do: workload at head of CQ is <= remaining nominal quota

This would definitely tell you if something is incorrectly "stuck" (and again short of fragmentation issues)

ns-sundar:
@Amy

> yes I've seen this used pretty successfully.
Could you clarify how this works? In the example I have above, the alert gets raised if A alone is pending. The moment users submit more jobs whereby the pending resource total exceeds the quota, the alert will stop. So, how it is a stable signal of scheduling problems?

amy:
@ns-sundar 

Moving the convo here: https://github.com/kubernetes-sigs/kueue/issues/10897
Made a new github issue for you. 

A clarification: 
1. I think you want single head of CQ workload <= remaining nominal: (**new issue**) https://github.com/kubernetes-sigs/kueue/issues/10897
    - This offers the granularity you're looking for in your example
2. this PR is for total pending resource capacity. And yeah you're right that it doesn't offer the specific granularity youre pointing out. However, I think it can offer a view into the scale of the issue. So... "because scheduling is failing, how many total resources should be scheduled but isn't"

But yeah lets move convo to: https://github.com/kubernetes-sigs/kueue/issues/10897
(sorry for bouncing around. Want to keep this PR to PR comments. Your point is really useful!)

pkg/cache/queue/cluster_queue.go
```diff
@@ -493,6 +592,20 @@ func (c *ClusterQueue) handleInadmissibleHash(hash string) int {
 	return moved
 }
 
+// PendingResources returns the total resources requested by all pending workloads,
+// aggregated by resource name. Pending workloads have not yet been assigned to flavors.
+func (c *ClusterQueue) PendingResources() map[corev1.ResourceName]int64 {
```
mimowo: nit, this does not need to be exported, as all uses seem scoped to the pkg/cache/queue

pkg/cache/queue/cluster_queue.go
mimowo: Oh I see, thank for for the detailed analysis.  I think the analysis is correct, but I'm wondering there is another way of fixing this which is less code, simply by making the Clone [here](https://github.com/kubernetes-sigs/kueue/blob/5252f419ec2610443a8f8af67bdb3af84b172094/pkg/scheduler/flavorassigner/flavorassigner.go#L585-L592). I think the fact that we are using the same pointer looks like a premature optimization to me, with this fishy update. Also, see that in the other code branch [here](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/scheduler/flavorassigner/flavorassigner.go#L588-L591) we would allocate a new array anyway.

No one could think that with the maps.Clone() wrapping `requests = a.wl.TotalRequests` in `flavorassigner.go` it runs slower - maybe compared against the main yet, but compared to the new proposal we need to make the clone anyway. We just make it explicit outside of the Info structure. 

Let me know if I'm missing something, but I think just wrapping `a.wl.TotalRequests` with maps.Clone could also solve the race, and the eventual performance would be the same.

The only potential problem is if we have somewhere else code which may be relying on the side effect of updating the pointer in place, but that is very tricky if this is true.

pkg/cache/queue/cluster_queue.go
amy: done

pkg/cache/queue/cluster_queue.go
amy: done! Lemme turn this into a review skill too.

amy:
@mimowo ready for a another pass! 
Changes from latest code review in latest commit here: https://github.com/kubernetes-sigs/kueue/pull/10485/commits/9e4efed3ab7c2b6c3e03c573d865886e4882a691

pkg/cache/queue/cluster_queue.go
```diff
@@ -140,6 +142,12 @@ type ClusterQueue struct {
 	sw *stickyWorkload
 
 	ConcurrentAdmissionPolicy *kueue.ConcurrentAdmissionPolicy
+	// pendingResourcesTotal is the incremental sum of TotalRequests across workloads
+	// in heap and inadmissibleWorkloads (not inflight). Updated at each mutation site so
+	// pendingResources() is O(1) rather than O(N).
+	// Configured resources are seeded at 0 by Update() so they appear in metrics
+	// even when no workloads are pending; stale zero entries are pruned on Update().
+	pendingResourcesTotal map[corev1.ResourceName]int64
```
mimowo, line 150: Actually, I have a question, why does the pendingResourcesTotal does not include the "inflight"? Because of that we need to subtract on pop, and add on aggregation. This is not a big issue, but I'm wondering if we could simplify even more.

amy: @mimowo 
`pendingResources()` gets simpler, but `delete()` and `wasTracked` both get new inflight cases. 
Its also the same pattern as `pendingActive()` and `totalElements()`

mimowo: > Its also the same pattern as pendingActive() and totalElements()

Yes, it makes sense to follow the same pattern here too.  I just hesitated for a while, because in other cases we only had to add one item, here we copy the resouce map, which is more work. OTOH, our scheduling cycle algorithm already has at least O(resource_count*flavors*pod_sets*...), so adding the clone here will not be a problem, I think.

mimowo:
/lgtm
/approve
Thank you 🌷

PR #10487, opened 2026-04-15
Fix Job and co-scheduling incompatibility after partial success on readmission 

jzhaojieh:
#### What type of PR is this?

/kind bug
/area integrations

When nodes fail, containers catching SIGTERM may exit 0, causing kubelet to report pods
as Succeeded. The k8s Job controller persists these indices in `CompletedIndexes`, which
survives suspend/resume. On re-admission after eviction, the Job controller skips
"completed" indices, creating fewer pods than the gang requires — leading to an infinite
PodsReadyTimeout loop since all pods are not created.

**Root cause sequence:**
1. JobSet with gang indexed Job (parallelism=completions=10) is running, all 10 pods active
2. Pods at indices 0,1 finish successfully — k8s Job controller sets
   `CompletedIndexes="0,1"` and `Succeeded=2` on the child Job
3. Higher-priority workload arrives → Kueue preempts and evicts this workload
4. Kueue suspends the JobSet → child Job suspended, but `CompletedIndexes="0,1"`
   and `Succeeded=2` persist 
5. Kueue re-admits workload → child Job unsuspended → k8s Job controller reads
   CompletedIndexes, creates only 8 pods (indices 2-9), skips 0 and 1
6. Co-scheduling plugin needs 10 pods (minMember=10), sees only 8 →
   PodsReadyTimeout → evict → re-admit → same 8 pods → infinite loop

During eviction (`stopJob` path), list child batch Jobs owned by the parent
(e.g., JobSet) and delete any gang indexed Jobs with stale `CompletedIndexes`
(`Succeeded > 0` with `parallelism == completions`). To cleanup any stale completed indexes from the previous attempt
and ensure all pods are created after readmission

N/A

-->
```release-note
Fix Job and co-scheduling incompatibility after partial success on readmission 
```

### DIFF
--- a/pkg/controller/jobframework/reconciler.go
+++ b/pkg/controller/jobframework/reconciler.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
@@ -52,6 +53,7 @@ import (
 	qcache "sigs.k8s.io/kueue/pkg/cache/queue"
 	schdcache "sigs.k8s.io/kueue/pkg/cache/scheduler"
 	"sigs.k8s.io/kueue/pkg/constants"
+	"sigs.k8s.io/kueue/pkg/controller/core/indexer"
 	controllerconsts "sigs.k8s.io/kueue/pkg/controller/constants"
 	"sigs.k8s.io/kueue/pkg/features"
 	"sigs.k8s.io/kueue/pkg/metrics"
@@ -583,6 +585,13 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
 			return ctrl.Result{}, err
 		}
 		if workload.HasQuotaReservation(wl) {
+			// Delete any child gang indexed Jobs with stale CompletedIndexes so
+			// the parent controller (e.g., JobSet) recreates them with clean state
+			// on re-admission. Gated on HasQuotaReservation to avoid redundant
+			// List calls on subsequent reconciles after admission is cleared.
+			if err := r.deleteStaleGangChildJobs(ctx, object); err != nil {
+				return ctrl.Result{}, err
+			}
 			if !job.IsActive() {
 				log.V(6).Info("The job is no longer active, clear the workloads admission")
 				err := workload.PatchAdmissionStatus(ctx, r.client, wl, r.clock, func(wl *kueue.Workload) (bool, error) {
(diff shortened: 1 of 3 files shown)


mimowo:
> When nodes fail, containers catching SIGTERM may exit 0, causing kubelet to report pods
as Succeeded. The k8s Job controller persists these indices in CompletedIndexes, which
survives suspend/resume. On re-admission after eviction, the Job controller skips
"completed" indices, creating fewer pods than the gang requires — leading to an infinite
PodsReadyTimeout loop since all pods are not created.

I can see the problem and I agree this needs a fix. At the same time I have two thoughts:
1. I'm ok with this behavior as default, but we should also support the previous by some JobSet annotations, because maybe for some use cases with embarrassingly parallel jobs it makes sense not to repeat the partially successful computations
2. this seems like something that could entirely be fixed on the JobSet level, with JobSet respecting some kind of annotation like "jobset/delete-jobs-on-suspend", or vice versa "jobset/keep-jobs-on-suspend", depending on the preferred default. I opened the issue there: https://github.com/kubernetes-sigs/jobset/issues/1208

cc @kannon92 @GiuseppeTT

pkg/controller/jobframework/reconciler.go
```diff
@@ -1771,3 +1780,59 @@ func mergeAnnotations(job GenericJob, customAnnotations map[string]string) (map[
...
+// ensure stale Succeeded pods are deleted before the Job object is removed,
+// preventing the k8s Job controller from re-adopting them on recreation.
+//
+// This runs for all GenericJob types during eviction. For non-parent types
+// (e.g., standalone batch/v1 Job), the List returns empty and is a no-op.
+// For parent types with children (JobSet, MPIJob, etc.), only gang indexed
+// child Jobs with Succeeded > 0 are deleted — non-gang and non-indexed
+// children are unaffected.
+func (r *JobReconciler) deleteStaleGangChildJobs(ctx context.Context, parent client.Object) error {
+	log := ctrl.LoggerFrom(ctx)
```
mimowo, line 1816: Would it be possible to de-couple the code from the JobSet and move into the JobSet_controller inside Kueue, but using the GenericJob interface? I imagine there is a new optional interface like CleanupStaleChild jobs which is called by the reconciler, and delegates the steps to the JobSet integration.

pkg/controller/jobframework/reconciler.go
```diff
@@ -1771,3 +1780,59 @@ func mergeAnnotations(job GenericJob, customAnnotations map[string]string) (map[
...
+// recreate them with clean state on re-admission. Uses foreground propagation to
+// ensure stale Succeeded pods are deleted before the Job object is removed,
+// preventing the k8s Job controller from re-adopting them on recreation.
+//
+// This runs for all GenericJob types during eviction. For non-parent types
+// (e.g., standalone batch/v1 Job), the List returns empty and is a no-op.
+// For parent types with children (JobSet, MPIJob, etc.), only gang indexed
+// child Jobs with Succeeded > 0 are deleted — non-gang and non-indexed
+// children are unaffected.
+func (r *JobReconciler) deleteStaleGangChildJobs(ctx context.Context, parent client.Object) error {
```
mimowo, line 1815: Also, I think this code should be behind a feature gate, and ulimately behind an annotation, as this is not a clear cut bug. Some emabrassinly parallel workloads may benefit from not retrying all indexes, and described in https://github.com/kubernetes-sigs/kueue/pull/10487#issuecomment-4249966002

pkg/controller/jobframework/reconciler.go
```diff
@@ -1771,3 +1780,59 @@ func mergeAnnotations(job GenericJob, customAnnotations map[string]string) (map[
...
+	log := ctrl.LoggerFrom(ctx)
+	var childJobs batchv1.JobList
```
Huang-Wei, line 1817: Can the list type be generic? i.e., to fit other Job-ish workloads.

mwielgus (commented):
In principle, Kueue should not interact too much with the internal mechanics of jobset, in particular trigger internal state cleanup via deletions. This behavior doesn't seam part of the official API and should rather be handled via https://github.com/kubernetes-sigs/jobset/issues/1208 rather than in Kueue. Please try proceeding with the Jobset community and let us know if you hit some pushback - we will evaluate other options then.

PR #8484, opened 2026-01-09
[TAS] Cache non-TAS pod usage

gabesaba:
#### What type of PR is this?
/kind bug

#### What this PR does / why we need it:
Improve performance of TAS by caching pods used to compute non-TAS usage.

#### Which issue(s) this PR fixes:
Fixes #8449

See https://github.com/kubernetes-sigs/kueue/issues/8449#issuecomment-3725934428 for alternatives considered

#### Special notes for your reviewer:
Considered WIP until I add an integration test (or verify that one already exists)

#### Does this PR introduce a user-facing change?
```release-note
TAS: significantly improves scheduling performance by replacing Pod listing with an event-driven
cache for non-TAS Pods, thereby avoiding expensive DeepCopy operations during each scheduling cycle.
```

PR #10488, opened 2026-04-15
TAS: filter irrelevant non-TAS pod updates

### LINKED ISSUE #10269: Large heap retained by zap logger clones in reconcile loops (19K-pod cluster)

MatteoFari:
#### What type of PR is this?
/kind bug
/area tas

Filters Pod updates that do not matter in NonTasUsageReconciler.

Before this change, the controller reconciled every update for scheduled non-TAS Pods. Now it only reconciles when a non-TAS Pod gets scheduled, moves to a different node, or finishes. 

Fixes #10269

```release-note
TAS: reduce the churn on the TAS-enabled controller, called NonTasUsageReconciler, by skipping triggering
of the Reconcile on Pod changes which are irrelevant from the controller point-of-view.
```

### DIFF
--- a/pkg/controller/tas/non_tas_usage_controller.go
+++ b/pkg/controller/tas/non_tas_usage_controller.go
@@ -32,6 +32,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	schdcache "sigs.k8s.io/kueue/pkg/cache/scheduler"
+	utilpod "sigs.k8s.io/kueue/pkg/util/pod"
 	"sigs.k8s.io/kueue/pkg/util/roletracker"
 	utiltas "sigs.k8s.io/kueue/pkg/util/tas"
 )
@@ -71,30 +72,45 @@ func (r *NonTasUsageReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, nil
 	}
 
-	r.cache.TASCache().Update(&pod, log)
+	if belongsToNonTASCache(&pod) {
+		r.cache.TASCache().Update(&pod, log)
+	} else {
+		r.cache.TASCache().DeletePodByKey(req.NamespacedName)
+	}
 	return ctrl.Result{}, nil
 }
 
-func filterPod(pod *corev1.Pod) bool {
+func belongsToNonTASCache(pod *corev1.Pod) bool {
+	if pod == nil {
+		return false
+	}
 	if utiltas.IsTAS(pod) {
 		return false
-	} else if len(pod.Spec.NodeName) == 0 {
-		// skip unscheduled pods as they don't use any capacity.
+	}
+	if len(pod.Spec.NodeName) == 0 {
+		// Skip unscheduled pods as they don't use any capacity.
+		return false
+	}
+	if utilpod.IsTerminated(pod) {
 		return false
 	}
 	return true
 }
 
 func (r *NonTasUsageReconciler) Create(e event.TypedCreateEvent[*corev1.Pod]) bool {
-	return filterPod(e.Object)
+	return belongsToNonTASCache(e.Object)
+}
+
+func shouldReconcilePodUpdate(oldPod, newPod *corev1.Pod) bool {
+	return belongsToNonTASCache(oldPod) != belongsToNonTASCache(newPod)
 }
 
 func (r *NonTasUsageReconciler) Update(e event.TypedUpdateEvent[*corev1.Pod]) bool {
-	return filterPod(e.ObjectNew)
+	return shouldReconcilePodUpdate(e.ObjectOld, e.ObjectNew)
 }
 
 func (r *NonTasUsageReconciler) Delete(e event.TypedDeleteEvent[*corev1.Pod]) bool {
-	return filterPod(e.Object)
+	return belongsToNonTASCache(e.Object)
 }
 
 func (r *NonTasUsageReconciler) Generic(event.TypedGenericEvent[*corev1.Pod]) bool {
--- a/pkg/controller/tas/non_tas_usage_controller_test.go
+++ b/pkg/controller/tas/non_tas_usage_controller_test.go
@@ -1,0 +1,225 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tas
+
+import (
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+
+	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+	testingpod "sigs.k8s.io/kueue/pkg/util/testingjobs/pod"
+)
+
+func TestBelongsToNonTASCache(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pkg/controller/tas/non_tas_usage_controller.go
```diff
@@ -89,8 +90,21 @@ func (r *NonTasUsageReconciler) Create(e event.TypedCreateEvent[*corev1.Pod]) bo
...
+		return false
+	}
+	if oldPod.Spec.NodeName != newPod.Spec.NodeName {
+		return true
+	}
+	if !utilpod.IsTerminated(oldPod) && utilpod.IsTerminated(newPod) {
+		return true
+	}
+	return false
+}
```
mimowo, line 106: I think this works well, so I can merge, but let me ask you an opinion about the alternative. I think the alternative will be easier to follow:

```golang
func belongsToNonTASCache(pod *corev1.Pod) bool {
	if utiltas.IsTAS(pod) {
		return false
	} else if len(pod.Spec.NodeName) == 0 {
		return false
	} else if util.IsTerminated(pod) {
        return false
    }
	return true
}
```

Then:

Create: `return belongsToNonTASCache(newPod)`
Update: `return belongsToNonTASCache(oldPod) !=  belongsToNonTASCache(newPod)`
Delete: `return belongsToNonTASCache(oldPod)`

So basically in this way the Create and Delete are just special cases for Update, because in case of Create we can assume `belongsToNonTASCache(oldPod=nil) = false`, and in case of Delete analogously `belongsToNonTASCache(newPod=nil) = false`, so the other needs to be true.

cc let me also cc @gabesaba who worked on the original impl.

mimowo:
/test pull-kueue-test-integration-baseline-main
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10488/pull-kueue-test-integration-baseline-main/2044336373787267072 looks like a known unrelated flake:
https://github.com/kubernetes-sigs/kueue/issues/10321

pkg/controller/tas/non_tas_usage_controller.go
MatteoFari: I think the alternative is easier to follow too.

One thing I am a bit unsure about is the case where the Pod changes from belonging to the non-TAS cache to not belonging anymore, while the new Pod is still not terminated. Right now `Reconcile` always fetches the Pod and calls `TASCache().Update(&pod, log)`, and that cache update does not enforce `belongsToNonTASCache` on its own.

So I think it is safe only if we also handle that case by deleting the Pod from the cache instead of updating it. Otherwise a transition like scheduled non-TAS -> scheduled TAS could still reconcile, but then add/update the new Pod in the cache instead of removing it.

gabesaba: We had this discussion during the original implementation, and decided that we don't need to handle the non-TAS to TAS case: https://github.com/kubernetes-sigs/kueue/pull/8484#discussion_r2690452272

mimowo: Indeed, but "don't need to handle" does not imply "don't handle", and my attention back then wasn't focused on the corner case.

Now looking at this it seems support that only requires  `if belongsToNonTASCache() Update else DeletePodByKey()` inside the Reconcile.

And at the expense of the single "if" we also express the intention better, so that the reader of the Reconcile function does not need to worry what is the transitioning.

gabesaba: This sounds good to me. It mitigates cache corruption risk by explicitly handling all cases (even ones that require unexpected/adversarial user intervention)

mimowo: @MatteoFari let's update as per https://github.com/kubernetes-sigs/kueue/pull/10488#discussion_r3085253452 + the extra "if" in the Reconcile. Feel free to still keep ShouldReconcilePodUpdate as the wrapper for the transition detection check.

mimowo (commented):
Thank you 👍 
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16

mimowo:
/release-note-edit
```release-note
TAS: reduce the churn on the TAS-enabled controller, called NonTasUsageReconciler, by skipping triggering
of the Reconcile on Pod changes which are irrelevant from the controller point-of-view.
```

PR #10492, opened 2026-04-15
[Fix] Avoid multiprocess run on MacOS

mszadkow:
#### What type of PR is this?
/kind bug
/area testing

We tracked the crash to gsutil running with Python on macOS in multiprocess mode, where forked child processes can segfault in system networking/framework code.
To prevent that we force GSUtil:parallel_process_count=1 on macOS, so downloads still use threads without spawning extra worker processes.
An opt-in override remains: set ENABLE_MULTIPROCESSING=1 if you explicitly want the old multiprocess behavior.

```release-note
NONE
```


mimowo:
sgtm, cc @mbobrovskyi

mbobrovskyi (commented):
/lgtm

Thank you!

PR #8655, opened 2026-01-19
Workload controller refactor - predicates cleanup - Delete event

Singularity23x0:
#### What type of PR is this?
/kind cleanup

#### What this PR does / why we need it:
The PR moves the logic previously set in the predicates to more appropriate places. 
Deletion now happens in the Reconcile method proper.
Notifying the watchers (other reconcilers, specifically queue reconcilers) has been moved to the Controller update handlers.
The two described changes aim to make Workload Controller and Reconciler compliant with general Kubernetes guidelines in terms of predicate methods only performing the check on whether the reconcile loop should be performed for a given event.

#### Which issue(s) this PR fixes:

Addresses the Delete method of the Workload Reconciler portion of the issue #5310.

#### Special notes for your reviewer:
NONE
#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #10496, opened 2026-04-15
Fix stale follower workload cache cleanup on NotFound reconciles

### LINKED ISSUE #10495: Follower can keep stale workload cache after leader failover, causing preemption target overlap and "workload not found" errors
**What happened**:

In the HA setup, we observed that pods remained `SchedulingGated` and workloads were pending after a leader failover. The pod had warning events indicating `Workload has overlapping preemption targets with another workload`. Also, kueue logs show a lot `workloads.kueue.x-k8s.io "<name>" not found` errors.

As additional context, we also observed that the gap between leader and follower metrics became much larger after the upgrade to `v0.17.0`, which might suggest cache divergence across replicas.

After a leader failover, the new leader works properly.

We do not yet have a minimal standalone reproducer, but the problem appears to require all of the following:

1. Run Kueue v0.17.0 in HA mode with at least 2 controller-manager replicas.
3. Let the follower observe workload updates while it is not the active leader.
4. Delete workloads that may still exist in the follower-side scheduler / queue caches.
5. Force a leader failover (in our case, the leader was OOM-killed).
6. Let the former follower become leader and start scheduling.

After failover, the new leader may keep stale preemption targets for deleted workloads and start emitting:

- `workloads.kueue.x-k8s.io "<name>" not found`
- `Workload has overlapping preemption targets with another workload`

We believe this may be a regression related to the workload delete handling refactor, specifically because:

- workload cache cleanup on delete appears to have moved to the `Reconcile` path on `Get(...)=NotFound`
- however, on follower replicas, `leaderAwareReconciler` returns early on `NotFound` and does not delegate to the underlying reconciler
- this seems to leave stale queue/scheduler cache state on followers
- when such a follower later becomes leader, it schedules using stale workload state

**Environment**:
- Kubernetes version (use `kubectl version`): 1.35
- Kueue version (use `git describe --tags --dirty --always`): 0.17.0
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Ladicle:
#### What type of PR is this?

/kind bug
/kind regression

This PR fixes a stale cache issue on follower replicas in HA controller-manager deployments.

A recent workload delete handling change #8655 relies on the delegated workload reconciler to clean up queue and scheduler caches when `Reconcile` sees `Get(...)=NotFound`. However, `leaderAwareReconciler` returned early on `NotFound` for follower replicas and did not delegate the request.

Fixes https://github.com/kubernetes-sigs/kueue/issues/10495

```release-note
Fixed a regression in HA controller-manager deployments where a follower replica could keep stale workload cache state after workload deletion.
```

### DIFF
--- a/pkg/controller/core/leader_aware_reconciler.go
+++ b/pkg/controller/core/leader_aware_reconciler.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"time"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -77,8 +78,12 @@ func (r *leaderAwareReconciler) Reconcile(ctx context.Context, request reconcile
 		return r.delegate.Reconcile(ctx, request)
 	default:
 		if err := r.client.Get(ctx, request.NamespacedName, r.object); err != nil {
-			// Discard request if not found, to prevent from re-enqueueing indefinitely.
-			return ctrl.Result{}, client.IgnoreNotFound(err)
+			if apierrors.IsNotFound(err) {
+				// Delegate not found requests so reconcilers can clean up in-memory state
+				// after delete events while this replica is still a follower.
+				return r.delegate.Reconcile(ctx, request)
+			}
+			return ctrl.Result{}, err
 		}
 		// The manager hasn't been elected leader yet, requeue the reconciliation request
 		// to prevent against any missed / discarded events over the period it takes
--- a/pkg/controller/core/leader_aware_reconciler_test.go
+++ b/pkg/controller/core/leader_aware_reconciler_test.go
@@ -1,0 +1,144 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package core
+
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+	qcache "sigs.k8s.io/kueue/pkg/cache/queue"
+	schdcache "sigs.k8s.io/kueue/pkg/cache/scheduler"
+	preemptexpectations "sigs.k8s.io/kueue/pkg/scheduler/preemption/expectations"
+	utiltesting "sigs.k8s.io/kueue/pkg/util/testing"
+	utiltestingapi "sigs.k8s.io/kueue/pkg/util/testing/v1beta2"
+	"sigs.k8s.io/kueue/pkg/workload"
+)
+
+type stubReconciler struct {
+	calls  int
+	result ctrl.Result
+	err    error
+}
+
+func (r *stubReconciler) Reconcile(context.Context, ctrl.Request) (ctrl.Result, error) {
+	r.calls++
+	return r.result, r.err
+}
+
+func newFollowerLeaderAwareReconciler(cl client.Client, delegate reconcile.Reconciler, requeueAfter time.Duration) *leaderAwareReconciler {
+	return &leaderAwareReconciler{
+		elected:         make(chan struct{}),
+		client:          cl,
+		delegate:        delegate,
+		object:          &kueue.Workload{},
+		requeueDuration: requeueAfter,
+	}
+}
+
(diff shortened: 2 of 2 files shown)


pkg/controller/core/leader_aware_reconciler.go
```diff
@@ -77,8 +78,12 @@ func (r *leaderAwareReconciler) Reconcile(ctx context.Context, request reconcile
 		return r.delegate.Reconcile(ctx, request)
 	default:
 		if err := r.client.Get(ctx, request.NamespacedName, r.object); err != nil {
-			// Discard request if not found, to prevent from re-enqueueing indefinitely.
-			return ctrl.Result{}, client.IgnoreNotFound(err)
+			if apierrors.IsNotFound(err) {
+				// Delegate not found requests so reconcilers can clean up in-memory state
+				// after delete events while this replica is still a follower.
+				return r.delegate.Reconcile(ctx, request)
```
mimowo, line 84: I think there is actually a subtle problem here, because this risks that we trigger Reconcile from follower, and it actually performs some work, like sending a request if another workload with the same name is created in the meanwhile. This sounds very unlikely, but possible, especially as the naming for workload is based on deterministic hashes - when a workload is deleted by a user it is almost immediately re-created by Kueue.

So I can see three possibilities:
1. accept the change as is and accept the risk of followers "doing some work on reconcile"
2. rollback https://github.com/kubernetes-sigs/kueue/pull/8655 as AFAIK it does not have any user-observable benefits, and it brings this regression. This was supposed to be a step in the right direction, but we didn't foresee the problem. cc @Singularity23x0 @PBundyra 
3. introduce the optional interface like `ReconcilerWithCleanup` which would be implemented by the workload reconciler, and called from here, to avoid calling the entire Reconcile

Actually, I'm thinking that 2 > 3 > 1, because:
* (1.) is unlikely , but possible, we will always need to think "maybe there was some request sent by a follower", and follower logs are often filtered out based on the replica-role
* (3.) would work, but requires non-trivial thinking about the mechanics
* (2.) was with us for very long, and the refactoring brought by the PR was not really required (I think)

Let me also know @Ladicle what you think.

And let me cc @tenzen-y and @gabesaba as this may have some non-trivial design decision about handling HA.

EDIT: at the same time both 2 and 3 are totally ok to me, the interface could be useful in other contexts maybe. I don't have a strong view here by any means.

tenzen-y: @mimowo Thank you for summarizing some options. I strongly recommend (2) for now.

For a long-term approach, instead of (3), I'm wondering if we can introduce SourceEventType to the Generic reconcile Request (`reconcile.TypedRequest[T]`) on the Workload controller, which could allow us to identify which event triggers the reconciliation.

The current Workload Reconciler mechanism estimates the source event, but it seems that not 100% sure. I guess that's the root cause.

When we face a similar problem even after we introduce SourceEventType, I want to consider (3).
Additionally, I guess that SourceEventType could reduce the risk of bugs rather than directly introducing (3).

Ladicle: Thanks. I missed the consideration of workloads with the same name. I don't fully realize the impact of the refactoring, but if it's reversible, I'm for (2) as the fastest way to resolve this.

mimowo: Would you have cycles to try reverting?

Ladicle: Sure, I'll give it a try.

mimowo: Thank you, if simple enough we may still include it in the patch today, cc @tenzen-y @gabesaba. Otherwise next patch

gabesaba: I agree with #2 as a temporary measure.

I discussed this briefly with @mimowo, and I propose the following long-term solution:

1) start serving visibility API from only leader
2) use standard list/watch pattern in follower -> leader upon failover
3) delete this leader_aware_reconciler (this seems quite brittle, and I can imagine seeing other cache corruption issues in the future)

gabesaba: > I discussed this briefly with [@mimowo](https://github.com/mimowo), and I propose the following long-term solution:

Though @tenzen-y notes potential performance issues with this approach, in both scheduling and visibility, due to lock contention

tenzen-y: > I discussed this briefly with @mimowo, and I propose the following long-term solution:
> 
> 1) start serving visibility API from only leader
> 2) use standard list/watch pattern in follower -> leader upon failover
> 3) delete this leader_aware_reconciler (this seems quite brittle, and I can imagine seeing other cache corruption issues in the future)

I synced this approach with @gabesaba, internally.
This has significant performance problems from workload scheduling and visibility endpoint PoV.
The OnDemandVisibility feature depends on internal queue information and takes read locks, which are shared across the scheduler and controllers.

Once we host the OnDemandVisibility endpoint only in the leader replicas, all user requests are accommodated only by the leader replicas, and then the read locks will be so busy.

As a result, the OnDemandVisility response time and workload scheduling performance will decline.

Hence, I'm still wondering if we can take either my approach or the (3) approach.

mimowo: > For a long-term approach, instead of (3), I'm wondering if we can introduce SourceEventType to the Generic reconcile Request (reconcile.TypedRequest[T]) on the Workload controller, which could allow us to identify which event triggers the reconciliation.
> The current Workload Reconciler mechanism estimates the source event, but it seems that not 100% sure. I guess that's the root cause.

I'm not sure how SourceEventType would fix the issue. The problem with https://github.com/kubernetes-sigs/kueue/pull/8655 is that Delete would only trigger Reconcile, and assume it is Reconcile to clean the cache for a deleted workload. 

However, Reconcile is silenced by `leaderAwareReconciler` on followers, and so even when a follower becomes a leader there is no mechanism to ever clean the stale workloads. So, I'm not clear how this would be related to the types of the objects triggering the Reconcile.

mimowo: Let me close the PR as the issue is already addressed, and we can continue the discussion in a dedicated issue: https://github.com/kubernetes-sigs/kueue/issues/10553

Issue #8036, opened 2025-12-02
[TAS] Cleanup `findTopologyAssignment` and related functions


PBundyra:
**What would you like to be cleaned**:
This list is not exhaustive, please see if you can see any room for cleanups yourself following best coding practices:

- Check if `fillInCountsHelper` needs to return 5 ints or maybe the code could rely on the `domain` argument: https://github.com/kubernetes-sigs/kueue/blob/main/pkg/cache/scheduler/tas_flavor_snapshot.go#L1397
- Add a new struct that would store the state of the algorithm so we can put all [this code](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/cache/scheduler/tas_flavor_snapshot.go#L667-L734) to some kind of initialization function and avoid passing e.g. 9 parameters to `fillInCounts` func. Add helper function if needed.
- Group all functions that take `*kueue.PodSetTopologyRequest` as a parameter together so they're not scattered across the whole file
- See if we reduce the number of [parameters in the domain struct](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/cache/scheduler/tas_flavor_snapshot.go#L71-L84) so the functions are a bit cleaner

**Why is this needed**:

kshalot:
Of the four things explicitly mentioned in the issue:
1. Addressed in #8491.
2. This change is possible and easy, but seemed low impact to me at this moment. Yes, it will reduce the amount of arguments to one function call, but won't help other functions in further phases in the algorithm (for example [this](https://github.com/kubernetes-sigs/kueue/blob/e83fce915d00d098c5b1280aa19e9bf03fafc131/pkg/cache/scheduler/tas_flavor_snapshot.go#L807)). It's a very simple change and can be readily done, but I think it would be better to come back to this in the context of a broader refactor. Wrapping the initialization in a separate function might be good for readability, OTOH the state is only used in this one function, so it naturally reads top to bottom when inlined. Potentially, creating a high level function that does `phase2(phase1(initialize(...)))` could be worthwhile, but might require more changes than just moving code around.
3. Addressed in #8448.
4. I took some time to understand how the algorithm works, but I was hesitant to introduce changes here as the `domain` struct fields are all used and in many many places in the code so a refactor is tricky, as it has to cover a lot cases.

[...]

PR #10506, opened 2026-04-16
TAS: group fillInCounts state

ShaanveerS:
#### What type of PR is this?
/kind cleanup
/area tas

Add a new struct that stores the state of the algorithm and avoids passing e.g. 9 parameters to `fillInCounts`.

Part of #8036

```release-note
NONE
```

### DIFF
--- a/pkg/cache/scheduler/tas_flavor_snapshot.go
+++ b/pkg/cache/scheduler/tas_flavor_snapshot.go
@@ -429,6 +429,29 @@ type ExclusionStats struct {
 	TotalNodes     int
 }
 
+// findTopologyAssignmentState stores derived inputs for a single run of the
+// TAS assignment algorithm.
+type findTopologyAssignmentState struct {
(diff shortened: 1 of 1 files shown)


mimowo (commented):
Nice one, I imagine in a follow up we could also use the new struct and pass it to the algos actually, ie functions like findBestDomainsForBalancedPlacement etc, wdyt?

/lgtm
/approve
Thank you 👍

ShaanveerS:
> Nice one, I imagine in a follow up we could also use the new struct and pass it to the algos actually, ie functions like findBestDomainsForBalancedPlacement etc, wdyt?

Thanks, 
SGTM, will followup in another pr.

mimowo:
/cherrypick release-0.17
/cherrypick release-0.16
As the basis for the actual fix: https://github.com/kubernetes-sigs/kueue/pull/10783

PR #10515, opened 2026-04-16
Finish orphaned workloads using stored owner metadata

sebest:
## Summary
Addresses #1789 by making orphan cleanup rely on workload-stored owner metadata instead of Job-side orphan handling.

- record owner GVK, owner name, and owner UID on created and prebuilt workloads
- finish orphaned workloads from the workload controller after orphan propagation removes owner references, including the window where the owner object is already deleting
- update job and pod controller expectations and add focused unit/integration coverage for orphaned and stale workload behavior

## Testing
- go test ./pkg/controller/core -run TestReconcile -count=1
- go test ./pkg/controller/jobframework -run 'TestReconcileGenericJob|TestReconcileGenericJobWithCustomWorkloadActivation|TestReconcileGenericJobWithWaitForPodsReady' -count=1
- go test ./pkg/controller/jobs/job -run TestReconcile -count=1
- go test ./pkg/controller/jobs/pod -run TestReconcile -count=1
- go test ./test/integration/singlecluster/controller/jobs/job -run TestAPIs -ginkgo.focus='orphaned workload cleanup' -count=1
- go test ./test/integration/singlecluster/controller/jobs/pod -run TestAPIs -ginkgo.focus='should not requeue stale workloads after pod is deleted and replaced' -count=1

### DIFF
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -210,11 +210,12 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	// Finish orphaned workloads whose controller owner no longer exists.
-	// The ownerReference still points to a non-existent object (e.g., a
-	// Deployment-owned pod deleted after PodsReady timeout eviction).
+	// Finish orphaned workloads whose owner no longer exists.
+	// Prefer the controller ownerReference when it is still present, and fall
+	// back to the stored owner metadata after orphan propagation removes owner
+	// references from the workload.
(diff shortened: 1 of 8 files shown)


mimowo:
@sebest I think the main complication here is how to distiguish a Workload which lost a parent due to --cascade=orphan from a workload which is prebuilt - like in the context of MultiKueue. 

I think we can safely delete the Workloads which had the Job in the past and lost it. So, we would need some kind of annotation to annotate they are adopted which outlives the ownerReference loss. Maybe something like kueue.x-k8s.io/owner-reference-uid: <>`. This could be populated either when creating a new Workload, or when adopting a worokload (setting owner reference).

Alternative is to use some pre-existing annotation/label to determine the distinction, controllerconsts.JobUIDLabel seems like a good fit actually.

PR #10518, opened 2026-04-16
Revert "Workload controller refactor - predicates cleanup - Delete event (#8655)"

### LINKED ISSUE #10495: Follower can keep stale workload cache after leader failover, causing preemption target overlap and "workload not found" errors

Ladicle:
#### What type of PR is this?

/bug
/kind regression

This PR reverts #8655 as a quick workaround to resolve #10495. (https://github.com/kubernetes-sigs/kueue/pull/10496#discussion_r3087674111)

Fixes https://github.com/kubernetes-sigs/kueue/issues/10495

```release-note
Fix a bug in HA mode that caused follower replicas to retain stale workload caches after deletion.
```


pkg/controller/core/workload_controller.go
```diff
@@ -1098,13 +1047,31 @@ func (r *WorkloadReconciler) Create(e event.TypedCreateEvent[*kueue.Workload]) b
...
 	status := "unknown"
 	if !e.DeleteStateUnknown {
 		status = workload.Status(e.Object)
 	}
-	r.logger().V(2).Info("Workload delete event", "workload", klog.KObj(e.Object), "queue", e.Object.Spec.QueueName, "status", status)
-	r.preemptionExpectations.ObservedUID(r.logger(),
+	log := r.logger().WithValues("workload", klog.KObj(e.Object), "queue", e.Object.Spec.QueueName, "status", status)
+	log.V(2).Info("Workload delete event")
+	r.preemptionExpectations.ObservedUID(log,
 		client.ObjectKeyFromObject(e.Object), e.Object.UID)
```
Ladicle, line 1058: Most of the conflicts were in the tests originally added in #8655, so I've simply removed them. Aside from the tests, this was the only conflict in the logic.

mimowo:
Can you already also prepare the CP for 0.17 using `./hack/cherry_pick_pull.sh release-0.17 10518` ?

pkg/controller/core/workload_controller.go
mimowo: Thank you for the report and the PR 👍

tenzen-y (commented):
I carefully checked this PR, then confirmed that the conflicts are mostly only TestWorkloadDeletion and TestGetWorkloadFromCache.

Thank you!
/lgtm
/approve

tenzen-y:
/release-note-edit
```release-note
Fix a bug in HA mode that caused follower replicas to retain stale workload caches after deletion.
```

PR #10527, opened 2026-04-16
Clean deleted workloads from non-leader cache

strigazi:
Ensure caches of non-leaders are cleaned on delete events.
With the current behaviour a delete event cleans only the leader's cache.

To reproduce the issue:
* create a workload and delete it before finishing
* check the cluster queue as documented in: docs/tasks/manage/monitor_pending_workloads/pending_workloads_on_demand/#cluster-queue-visibility-via-curl

/kind bug

```release-note
Clean deleted workloads from non-leader cache.
```

### DIFF
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -1084,9 +1084,15 @@ func (r *WorkloadReconciler) Delete(e event.TypedDeleteEvent[*kueue.Workload]) b
 	if !e.DeleteStateUnknown {
 		status = workload.Status(e.Object)
 	}
-	r.logger().V(2).Info("Workload delete event", "workload", klog.KObj(e.Object), "queue", e.Object.Spec.QueueName, "status", status)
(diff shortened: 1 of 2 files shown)


kshalot:
Hey, thanks for the contribution @strigazi. At first glance this looks like the merged https://github.com/kubernetes-sigs/kueue/pull/10518.

strigazi:
a bit late https://github.com/kubernetes-sigs/kueue/pull/10518 reverted https://github.com/kubernetes-sigs/kueue/pull/8655

mimowo:
No worries, but I think we need some kind of e2e tests to make sure this does not return as a regression. Let me open an issue for that, then if you want you could grab it

strigazi:
Thank you for the quick reply @kshalot and @mimowo , I realized that it was fixed when rebasing.

mimowo:
ptal: https://github.com/kubernetes-sigs/kueue/issues/10529

PR #10537, opened 2026-04-16
TAS: pass state to placement algos

ShaanveerS:
#### What type of PR is this?

/kind cleanup
/area tas

Use the state struct added in #10506 in the placement algos too
Followup #10506

```release-note
NONE
```

### DIFF
--- a/pkg/cache/scheduler/tas_flavor_snapshot.go
+++ b/pkg/cache/scheduler/tas_flavor_snapshot.go
@@ -891,10 +891,10 @@ func (s *TASFlavorSnapshot) findTopologyAssignment(
 	var useBalancedPlacement bool
 	if features.Enabled(features.TASBalancedPlacement) && !state.required && !state.unconstrained {
 		var bestThreshold int32
-		currFitDomain, bestThreshold = findBestDomainsForBalancedPlacement(s, levelIdx, state.sliceLevelIdx, state.count, state.leaderCount, state.sliceSize)
+		currFitDomain, bestThreshold = findBestDomainsForBalancedPlacement(s, state)
 		useBalancedPlacement = bestThreshold > 0
 		if useBalancedPlacement {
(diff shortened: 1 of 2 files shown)


pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -891,27 +891,18 @@ func (s *TASFlavorSnapshot) findTopologyAssignment(
...
-			levelIdx,
-			state.required,
-			state.count,
-			state.leaderCount,
-			state.sliceSize,
-			state.unconstrained,
-			state.stats,
-			state.multiLayerConstraints,
-		)
+		fitLevelIdx, currFitDomain, reason = s.findLevelWithFitDomains(state.levelIdx, state)
```
pajakd, line 905: nit: It is easy to get confused with all the `levelIdx` variables. The levelIdx in state struct is the requested topology level so perhaps should be renamed to `requestedLevelIdx` or sth (can be in the followup).

pajakd:
This is a much needed refactor! We could even consider breaking the `findTopologyAssignmentState` into some smaller structs like `podRequirements` (with requests, selectors, tolerations etc) and `tasParameters` (with sliceSize, required etc). The idea is that `podRequirements` will only be needed up to `fillInCounts`. When the `state` fields are filled, we should not pass these arguments to functions like `findLevelWithFitDomains`.

This is another idea for the followup, WDYT?

ShaanveerS:
> This is a much needed refactor! We could even consider breaking the `findTopologyAssignmentState` into some smaller structs like `podRequirements` (with requests, selectors, tolerations etc) and `tasParameters` (with sliceSize, required etc). The idea is that `podRequirements` will only be needed up to `fillInCounts`. When the `state` fields are filled, we should not pass these arguments to functions like `findLevelWithFitDomains`.
> 
> This is another idea for the followup, WDYT?

Sounds good, will do that in a follow-up.

mimowo:
/approve
Thanks !

mimowo:
/cherrypick release-0.17
As the basis for the actual fix: #10783

PR #10552, opened 2026-04-17
metrics: unify stale gauge cleanup and resync

### LINKED ISSUE #9854: Consolidate gauge metric cleanup mechanisms

MatteoFari:
#### What type of PR is this?
/kind cleanup

Makes custom metric label changes follow the same cleanup and resync flow as `ResyncGaugeMetrics`.

When metric labels change on a CQ, LQ or Cohort, Kueue now removes the old gauge metrics and reports them again with the new label values.

Also fixes the last cohort subtree case, old `cohort_subtree_admitted_active_workloads` metrics are removed when a CQ changes both its cohort and its custom label at the same time.

Fixes #9854

PR builds on #10276, which was merged as the 1st step

```release-note
NONE
```

### DIFF
--- a/pkg/cache/scheduler/cache.go
+++ b/pkg/cache/scheduler/cache.go
@@ -495,6 +495,66 @@ func (c *Cache) UpdateClusterQueue(log logr.Logger, cq *kueue.ClusterQueue) erro
 	return nil
 }
 
+func (c *Cache) resyncClusterQueueGaugeMetricsLocked(cq *clusterQueue) {
+	if cq == nil {
+		return
+	}
+	metrics.ReportClusterQueueStatus(cq.Name, cq.Status, cq.customMetricLabelValues, c.roleTracker)
+	cq.reportActiveWorkloads()
+	if c.resourceMetricsEnabled {
+		cq.reportResourceMetrics(c.fairSharingEnabled)
+	}
+	for _, lq := range cq.localQueues {
+		c.resyncLocalQueueGaugeMetricsLocked(cq, lq)
+	}
+}
+
+func (c *Cache) ResyncClusterQueueGaugeMetrics(cqName kueue.ClusterQueueReference) {
+	c.RLock()
+	defer c.RUnlock()
+	c.resyncClusterQueueGaugeMetricsLocked(c.hm.ClusterQueue(cqName))
+}
+
+func (c *Cache) resyncLocalQueueGaugeMetricsLocked(cq *clusterQueue, lq *LocalQueue) {
+	if cq == nil || lq == nil || !lq.shouldExposeMetrics(c.lqMetrics) {
+		return
+	}
+	lq.reportActiveWorkloads(c.roleTracker)
+	lq.reportResourceMetrics(cq.resourceNode.Quotas, c.roleTracker)
+}
+
+func (c *Cache) ResyncLocalQueueGaugeMetrics(cqName kueue.ClusterQueueReference, lqRef queue.LocalQueueReference) {
+	c.RLock()
+	defer c.RUnlock()
+	cq := c.hm.ClusterQueue(cqName)
+	if cq == nil {
+		return
+	}
+	lq, ok := cq.localQueues[lqRef]
+	if !ok {
+		return
+	}
+	c.resyncLocalQueueGaugeMetricsLocked(cq, lq)
+}
+
+func (c *Cache) ResyncCohortGaugeMetrics(log logr.Logger, cohortName kueue.CohortReference) {
+	c.RecordCohortMetrics(log, cohortName)
+	c.RLock()
+	defer c.RUnlock()
+	cohort := c.hm.Cohort(cohortName)
+	if cohort == nil || hierarchy.HasCycle(cohort) {
+		return
+	}
+	if c.fairSharingEnabled {
+		drs := dominantResourceShare(cohort, nil)
+		var customLabelValues []string
+		if features.Enabled(features.CustomMetricLabels) {
+			customLabelValues = c.customLabels.CohortGet(cohort.Name)
+		}
+		metrics.ReportCohortWeightedShare(cohort.Name, drs.PreciseWeightedShare(), customLabelValues, c.roleTracker)
+	}
+}
+
 func (c *Cache) DeleteClusterQueue(cq *kueue.ClusterQueue) {
 	c.Lock()
 	defer c.Unlock()
@@ -1031,31 +1091,25 @@ func (c *Cache) MatchingClusterQueues(nsLabels map[string]string) sets.Set[kueue
 }
 
 // ResyncGaugeMetrics re-reports CQ/LQ status, active workload, resource, and weighted share gauge metrics.
-func (c *Cache) ResyncGaugeMetrics() {
+func (c *Cache) ResyncGaugeMetrics(log logr.Logger) {
 	c.RLock()
-	defer c.RUnlock()
-	for _, cq := range c.hm.ClusterQueues() {
-		metrics.ReportClusterQueueStatus(cq.Name, cq.Status, cq.customMetricLabelValues, c.roleTracker)
-		cq.reportActiveWorkloads()
-		if c.resourceMetricsEnabled {
-			cq.reportResourceMetrics(c.fairSharingEnabled)
-		}
-		for _, lq := range cq.localQueues {
-			if lq.shouldExposeMetrics(c.lqMetrics) {
-				lq.reportActiveWorkloads(c.roleTracker)
-				lq.reportResourceMetrics(cq.resourceNode.Quotas, c.roleTracker)
-			}
-		}
+	cqs := c.hm.ClusterQueues()
+	cqNames := make([]kueue.ClusterQueueReference, 0, len(cqs))
+	for _, cq := range cqs {
+		cqNames = append(cqNames, cq.Name)
 	}
-	if c.fairSharingEnabled {
-		for _, cohort := range c.hm.Cohorts() {
-			drs := dominantResourceShare(cohort, nil)
-			var customLabelValues []string
-			if features.Enabled(features.CustomMetricLabels) {
-				customLabelValues = c.customLabels.CohortGet(cohort.Name)
-			}
-			metrics.ReportCohortWeightedShare(cohort.Name, drs.PreciseWeightedShare(), customLabelValues, c.roleTracker)
-		}
+	cohorts := c.hm.Cohorts()
+	cohortNames := make([]kueue.CohortReference, 0, len(cohorts))
+	for _, cohort := range cohorts {
+		cohortNames = append(cohortNames, cohort.Name)
+	}
+	c.RUnlock()
+
+	for _, cqName := range cqNames {
+		c.ResyncClusterQueueGaugeMetrics(cqName)
+	}
+	for _, cohortName := range cohortNames {
+		c.ResyncCohortGaugeMetrics(log, cohortName)
 	}
 }
 
--- a/pkg/controller/core/cohort_controller.go
+++ b/pkg/controller/core/cohort_controller.go
@@ -18,6 +18,7 @@ package core
 
 import (
 	"context"
+	"slices"
 
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/api/equality"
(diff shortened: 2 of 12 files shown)


mimowo:
/ok-to-test
cc @IrvingMg

IrvingMg (commented):
Thanks for this, @MatteoFari! 

I gave it a quick pass, one thing I’d suggest before continuing: could we separate the fix for the `cohort_subtree_admitted_active_workloads` metrics issue? It would be helpful to have more context on the bug and address it in a separate PR so it can be easily cherry-picked if needed.

pkg/controller/core/localqueue_controller.go
```diff
@@ -316,14 +312,23 @@ func (r *LocalQueueReconciler) Update(e event.TypedUpdateEvent[*kueue.LocalQueue
...
 			log.Error(err, "Failed to update localQueue in the cache")
 		}
+		if labelsUpdated {
+			r.resyncLocalQueueGaugeMetrics(e.ObjectNew)
+		}
 		return true
 	}
 
 	if newStopPolicy == kueue.None {
+		if err := r.cache.UpdateLocalQueue(e.ObjectOld, e.ObjectNew); err != nil {
```
IrvingMg, line 322: Why do we need this? Does it fix an existing bug?

pkg/cache/scheduler/cache.go
```diff
@@ -494,6 +494,59 @@ func (c *Cache) UpdateClusterQueue(log logr.Logger, cq *kueue.ClusterQueue) erro
...
+
+func (c *Cache) resyncLocalQueueGaugeMetricsLocked(cq *clusterQueue, lq *LocalQueue) {
+	if cq == nil || lq == nil || !lq.shouldExposeMetrics(c.lqMetrics) {
+		return
+	}
+	lq.reportActiveWorkloads(c.roleTracker)
+	lq.reportResourceMetrics(cq.resourceNode.Quotas, c.roleTracker)
+}
+
+func (c *Cache) ResyncLocalQueueGaugeMetrics(lqRef queue.LocalQueueReference) {
```
IrvingMg, line 525: I think we can also pass `cqName kueue.ClusterQueueReference` to avoid the linear search below, as done in `GetCacheLocalQueue`.

pkg/cache/scheduler/cache.go
```diff
@@ -1008,29 +1061,21 @@ func (c *Cache) MatchingClusterQueues(nsLabels map[string]string) sets.Set[kueue
 // ResyncGaugeMetrics re-reports CQ/LQ status, active workload, resource, and weighted share gauge metrics.
 func (c *Cache) ResyncGaugeMetrics() {
 	c.RLock()
-	defer c.RUnlock()
+	cqNames := make([]kueue.ClusterQueueReference, 0, len(c.hm.ClusterQueues()))
```
IrvingMg, line 1064: Note that c.hm.ClusterQueues() clones the entire map under the hood, so it may be better to call it once (e.g., `cqs := c.hm.ClusterQueues()`) and reuse it for all other operations.

pkg/controller/core/localqueue_controller.go
```diff
@@ -316,14 +312,23 @@ func (r *LocalQueueReconciler) Update(e event.TypedUpdateEvent[*kueue.LocalQueue
 		if err := r.cache.UpdateLocalQueue(e.ObjectOld, e.ObjectNew); err != nil {
 			log.Error(err, "Failed to update localQueue in the cache")
 		}
+		if labelsUpdated {
+			r.resyncLocalQueueGaugeMetrics(e.ObjectNew)
+		}
```
IrvingMg, line 317: Looks like this check is duplicated, maybe we can move it to the end of the function?

pkg/cache/scheduler/resync_metrics_test.go
IrvingMg, line 1: nit: Should we rename this to cache_resync_metrics_test.go to match the format of the other test files?

pkg/cache/scheduler/cache.go
```diff
@@ -494,6 +494,59 @@ func (c *Cache) UpdateClusterQueue(log logr.Logger, cq *kueue.ClusterQueue) erro
...
+	c.RecordCohortMetrics(logr.Discard(), cohortName)
+	c.RLock()
+	defer c.RUnlock()
+	cohort := c.hm.Cohort(cohortName)
+	if cohort == nil || hierarchy.HasCycle(cohort) {
+		return
+	}
+	if c.fairSharingEnabled {
+		drs := dominantResourceShare(cohort, nil)
+		metrics.ReportCohortWeightedShare(cohort.Name, drs.PreciseWeightedShare(), c.customLabels.CohortGet(cohort.Name), c.roleTracker)
```
IrvingMg, line 546: The old `ResyncGaugeMetrics` had an `if features.Enabled(features.CustomMetricLabels)` check around `CohortGet`. I think it should work without it as well, but the check makes the intent explicit, so it might be worth keeping it.

IrvingMg (changes requested):

pkg/controller/core/localqueue_controller.go
MatteoFari: It's only needed when the LocalQueue changes labels or moves to a different ClusterQueue while switching back to `StopPolicy=None`.

Without that cache update, the targeted resync would read stale LocalQueue state from the scheduler cache. I narrowed it to only those cases.

mimowo (commented):
cc @IrvingMg do you have cycles to finalize the review here? Maybe also @mszadkow could take a look as I know he recently worked on metrics

test/integration/singlecluster/controller/core/custom_labels_test.go
```diff
@@ -134,6 +135,18 @@ var _ = ginkgo.Describe("CustomMetricLabels", ginkgo.Label("controller:clusterqu
...
 			ginkgo.By("verifying metric with custom_team=alpha")
 			util.ExpectPendingWorkloadsMetric(cq, 1, 0, "alpha")
+			gomega.Eventually(func() int {
+				return len(testingmetrics.CollectFilteredGaugeVec(metrics.ClusterQueueByStatus, map[string]string{
+					"cluster_queue": cq.Name,
+					"custom_team":   "alpha",
+				}))
+			}, util.Timeout, util.Interval).Should(gomega.Equal(len(metrics.CQStatuses)))
+			gomega.Eventually(func() int {
+				return len(testingmetrics.CollectFilteredGaugeVec(metrics.ClusterQueueResourceNominalQuota, map[string]string{
```
mszadkow: Maybe we could extend `ExpectClusterQueueStatusMetric` with custom labels and use it here

pkg/controller/core/localqueue_controller.go
```diff
@@ -290,24 +286,26 @@ func (r *LocalQueueReconciler) Update(e event.TypedUpdateEvent[*kueue.LocalQueue
...
 		clearLocalQueueMetrics(e.ObjectOld)
 	}
 
 	oldStopPolicy := ptr.Deref(e.ObjectOld.Spec.StopPolicy, kueue.None)
 	newStopPolicy := ptr.Deref(e.ObjectNew.Spec.StopPolicy, kueue.None)
+	shouldResyncGauges := false
 
-	if newStopPolicy == oldStopPolicy {
+	switch newStopPolicy {
+	case oldStopPolicy:
```
mszadkow: it is readably suspicious, even if it may be functionally correct
wdyt @mbobrovskyi?

pkg/controller/core/cohort_controller.go
```diff
@@ -143,12 +144,10 @@ func (r *CohortReconciler) Update(e event.TypedUpdateEvent[*kueue.Cohort]) bool
...
-		customLabelsChanged = r.customLabels.CohortStoreAndClear(
-			kueue.CohortReference(e.ObjectNew.GetName()),
-			e.ObjectNew.GetLabels(), e.ObjectNew.GetAnnotations(),
-			func() {
-				metrics.ClearCohortMetrics(kueue.CohortReference(e.ObjectNew.GetName()))
-			})
+		customLabelsChanged = !slices.Equal(
+			r.customLabels.CohortGet(kueue.CohortReference(e.ObjectNew.GetName())),
+			r.customLabels.ExtractValues(e.ObjectNew.GetLabels(), e.ObjectNew.GetAnnotations()),
+		)
```
mszadkow, line 150: we stopped to store the labels here in favor of doing that in the Reconcile, is that correct?

pkg/controller/core/localqueue_controller.go
```diff
@@ -185,14 +185,10 @@ func (r *LocalQueueReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			utilqueue.Key(&queueObj),
 			queueObj.GetLabels(), queueObj.GetAnnotations(),
-			func() {
-				lqRef := localQueueReferenceFromLocalQueue(&queueObj)
-				metrics.ClearLocalQueueMetrics(lqRef)
-				metrics.ClearLocalQueueCacheMetrics(lqRef)
-				metrics.ClearLocalQueueResourceMetrics(lqRef)
-			})
+		)
 	}
```
mszadkow, line 191: maybe we could move the actual store to the Reconcile, while checking it here only, like in cq

pkg/cache/scheduler/cache.go
```diff
@@ -1008,29 +1068,23 @@ func (c *Cache) MatchingClusterQueues(nsLabels map[string]string) sets.Set[kueue
...
+		cohortNames = append(cohortNames, cohort.Name)
+	}
+	c.RUnlock()
+
+	for _, cqName := range cqNames {
+		c.ResyncClusterQueueGaugeMetrics(cqName)
+	}
+	for _, cohortName := range cohortNames {
+		c.ResyncCohortGaugeMetrics(cohortName)
 	}
```
mszadkow, line 1113: maybe it's worth to use/create locked versions for that part too and then just defer the lock, thinking out loud

pkg/controller/core/localqueue_controller.go
MatteoFari: changed so each branch reads as explicit condition.

pkg/controller/core/cohort_controller.go
MatteoFari: exactly

pkg/controller/core/localqueue_controller.go
MatteoFari: I kept this as-is because resync needs the new LQ labels. If we save labels in `Reconcile`, should resync move there too?

pkg/cache/scheduler/cache.go
MatteoFari: It reuses the same resync helpers as the custom-label path. Let me know if a single lock-unlock would be better

pkg/cache/scheduler/cache.go
```diff
@@ -494,6 +494,66 @@ func (c *Cache) UpdateClusterQueue(log logr.Logger, cq *kueue.ClusterQueue) erro
...
+	}
+	lq, ok := cq.localQueues[lqRef]
+	if !ok {
+		return
+	}
+	c.resyncLocalQueueGaugeMetricsLocked(cq, lq)
+}
+
+func (c *Cache) ResyncCohortGaugeMetrics(cohortName kueue.CohortReference) {
+	c.RecordCohortMetrics(logr.Discard(), cohortName)
```
IrvingMg: Why are we discarding the logs? Can we pass the logger somehow? I see `log.Discard()` is being introduced in other places as well.

pkg/controller/core/localqueue_controller.go
```diff
@@ -290,24 +286,26 @@ func (r *LocalQueueReconciler) Update(e event.TypedUpdateEvent[*kueue.LocalQueue
 	log := r.logger().WithValues("localQueue", klog.KObj(e.ObjectNew))
 	log.V(2).Info("Queue update event")
 
+	var labelsUpdated bool
```
IrvingMg: `labelsUpdated` feels like it’s doing a lot here: it’s gating the metrics update or clear, the cache refresh, and the resync at the end.

Also, `shouldResyncGauges` doesn’t seem to add much. It’s basically just `labelsUpdated` in the non-default cases and `false` otherwise. Maybe we could drop it?

Not sure what the best change is for `labelsUpdated`, but the current logic is a bit hard to follow. Maybe there’s a way to make this logic clearer?

MatteoFari: Removed `shouldResyncGauges` and split the logic into clearer flags. Resync happens once at the end.

mimowo:
cc @mszadkow @IrvingMg @mbobrovskyi ptal as the PR looks almost ready to my eyes, but would be good to have someone else to look at it

mimowo:
/restest
to double check no flakes

pkg/cache/scheduler/cache.go
mszadkow: it's fine

pkg/controller/core/localqueue_controller.go
mszadkow: ok, understood, no need to change that

mszadkow:
looks good to me

mimowo:
Thanks 👍 
/lgtm
/approve
I checked the PR briefly and looks reasonable to me,  getting additional support from lgtm in https://github.com/kubernetes-sigs/kueue/pull/10552#issuecomment-4462716389

PR #5985, opened 2025-07-15
Incorrect admission due to wrong borrowable resource when a cluster queue in cohort is deleted

amy:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:
Fixes https://github.com/kubernetes-sigs/kueue/issues/5937
updates cohort tree resources in cache on clusterQueue deletion event

#### Which issue(s) this PR fixes:

Fixes #5937

#### Does this PR introduce a user-facing change?

```release-note
Fix incorrect workload admission after CQ is deleted in a cohort reducing the amount of available quota. The culprit of the issue was that the cached amount of quota was not updated on CQ deletion.
```

PR #10579, opened 2026-04-17
Update SubtreeQuota post cohort deletion

### LINKED ISSUE #10417: flaky test: Cohorts when creating, modifying and removing follows reporting correct metrics
**Which test is flaking?**:
Cohorts when creating, modifying and removing follows reporting correct metrics
https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/periodic-kueue-test-integration-main/2042688894889103360
**Failure message or logs**:
```
Scheduler Fair Sharing Suite: [It] Cohorts when creating, modifying and removing follows reporting correct metrics expand_less	17s
{ failed [FAILED] Timed out after 10.007s.
The function passed to Eventually failed at /home/prow/go/src/kubernetes-sigs/kueue/test/util/metrics.go:296 with:
Expected
    <float64>: 50000
to equal
    <float64>: 40000
In [It] at: /home/prow/go/src/kubernetes-sigs/kueue/test/integration/singlecluster/scheduler/fairsharing/metrics_test.go:342 @ 04/10/26 20:04:50.752
}
```

The same test, but different assert was recently fixed: https://github.com/kubernetes-sigs/kueue/issues/10057

mszadkow:
#### What type of PR is this?
/kind bug

Update the SubtreeQuota of all ancestors to reflect the removal of the deleted cohort. 
Without this, stale SubtreeQuota values would cause any subsequent RecordCohortMetrics call for ancestor cohorts to re-publish the pre-deletion metric values.

Fixes #10417 

```release-note
NONE
```

### DIFF
--- a/pkg/cache/scheduler/cache.go
+++ b/pkg/cache/scheduler/cache.go
@@ -540,6 +540,8 @@ func (c *Cache) AddOrUpdateCohort(apiCohort *kueue.Cohort) error {
 	return nil
 }
 
+// DeleteCohort removes the cohort from the cache and updates the SubtreeQuota
+// of ancestor cohorts to reflect the removal.
 func (c *Cache) DeleteCohort(cohortName kueue.CohortReference) {
 	c.Lock()
 	defer c.Unlock()
@@ -552,7 +554,6 @@ func (c *Cache) DeleteCohort(cohortName kueue.CohortReference) {
 	}
 
 	c.hm.DeleteCohort(cohortName)
-	c.handleParentUpdate(parent)
 
 	// If the cohort still exists after deletion, it means
 	// that it has one or more children referencing it.
@@ -560,6 +561,12 @@ func (c *Cache) DeleteCohort(cohortName kueue.CohortReference) {
 	if cohort := c.hm.Cohort(cohortName); cohort != nil {
 		updateCohortResourceNode(cohort)
 	}
+
+	if parent != nil {
+		updateCohortTreeResourcesIfNoCycle(parent)
+	}
+
+	c.handleParentUpdate(parent)
 }
 
 func (c *Cache) handleParentUpdate(cachedParent *cohort) {
--- a/pkg/cache/scheduler/cache_test.go
+++ b/pkg/cache/scheduler/cache_test.go
@@ -3570,148 +3570,125 @@ func TestCohortCycles(t *testing.T) {
 }
 
 func TestDeleteCohortUpdatesAncestorSubtreeQuota(t *testing.T) {
-	t.Run("deleting child cohort updates parent SubtreeQuota", func(t *testing.T) {
-		cache := New(utiltesting.NewFakeClient())
-
-		root := utiltestingapi.MakeCohort("root").
-			ResourceGroup(*utiltestingapi.MakeFlavorQuotas("default").Resource(corev1.ResourceCPU, "10").Obj()).
-			Obj()
-		if err := cache.AddOrUpdateCohort(root); err != nil {
+	mustAddCohort := func(t *testing.T, cache *Cache, cohort *kueue.Cohort) {
+		t.Helper()
+		if err := cache.AddOrUpdateCohort(cohort); err != nil {
 			t.Fatal(err)
 		}
+	}
 
-		child := utiltestingapi.MakeCohort("child").
-			Parent("root").
-			ResourceGroup(*utiltestingapi.MakeFlavorQuotas("default").Resource(corev1.ResourceCPU, "5").Obj()).
-			Obj()
-		if err := cache.AddOrUpdateCohort(child); err != nil {
-			t.Fatal(err)
-		}
+	testCases := map[string]struct {
+		setup      func(*testing.T, *Cache)
+		deleteName kueue.CohortReference
+		wantBefore map[kueue.CohortReference]resources.FlavorResourceQuantities
+		wantAfter  map[kueue.CohortReference]resources.FlavorResourceQuantities
+	}{
+		"deleting child cohort updates parent SubtreeQuota": {
+			setup: func(t *testing.T, cache *Cache) {
+				mustAddCohort(t, cache, utiltestingapi.MakeCohort("root").
+					ResourceGroup(*utiltestingapi.MakeFlavorQuotas("default").Resource(corev1.ResourceCPU, "10").Obj()).
+					Obj())
+				mustAddCohort(t, cache, utiltestingapi.MakeCohort("child").
+					Parent("root").
+					ResourceGroup(*utiltestingapi.MakeFlavorQuotas("default").Resource(corev1.ResourceCPU, "5").Obj()).
+					Obj())
+			},
+			deleteName: "child",
+			wantBefore: map[kueue.CohortReference]resources.FlavorResourceQuantities{
+				"root": {
+					{Flavor: "default", Resource: corev1.ResourceCPU}: 15_000,
+				},
+			},
+			wantAfter: map[kueue.CohortReference]resources.FlavorResourceQuantities{
+				"root": {
+					{Flavor: "default", Resource: corev1.ResourceCPU}: 10_000,
+				},
+			},
+		},
+		"deleting grandchild cohort updates all ancestors SubtreeQuota": {
+			setup: func(t *testing.T, cache *Cache) {
+				mustAddCohort(t, cache, utiltestingapi.MakeCohort("root").
+					ResourceGroup(*utiltestingapi.MakeFlavorQuotas("default").Resource(corev1.ResourceCPU, "10").Obj()).
+					Obj())
+				mustAddCohort(t, cache, utiltestingapi.MakeCohort("mid").
+					Parent("root").
+					ResourceGroup(*utiltestingapi.MakeFlavorQuotas("default").Resource(corev1.ResourceCPU, "5").Obj()).
+					Obj())
+				mustAddCohort(t, cache, utiltestingapi.MakeCohort("leaf").
+					Parent("mid").
+					ResourceGroup(*utiltestingapi.MakeFlavorQuotas("default").Resource(corev1.ResourceCPU, "3").Obj()).
+					Obj())
+			},
+			deleteName: "leaf",
+			wantBefore: map[kueue.CohortReference]resources.FlavorResourceQuantities{
+				"mid": {
+					{Flavor: "default", Resource: corev1.ResourceCPU}: 8_000,
+				},
+				"root": {
+					{Flavor: "default", Resource: corev1.ResourceCPU}: 18_000,
+				},
+			},
+			wantAfter: map[kueue.CohortReference]resources.FlavorResourceQuantities{
+				"mid": {
+					{Flavor: "default", Resource: corev1.ResourceCPU}: 5_000,
+				},
+				"root": {
+					{Flavor: "default", Resource: corev1.ResourceCPU}: 15_000,
+				},
+			},
+		},
+		"deleting cohort with children updates parent SubtreeQuota": {
+			setup: func(t *testing.T, cache *Cache) {
+				ctx, _ := utiltesting.ContextWithLog(t)
+				mustAddCohort(t, cache, utiltestingapi.MakeCohort("root").
+					ResourceGroup(*utiltestingapi.MakeFlavorQuotas("default").Resource(corev1.ResourceCPU, "10").Obj()).
+					Obj())
+				mustAddCohort(t, cache, utiltestingapi.MakeCohort("mid").
+					Parent("root").
+					ResourceGroup(*utiltestingapi.MakeFlavorQuotas("default").Resource(corev1.ResourceCPU, "5").Obj()).
+					Obj())
 
-		// before deletion: root SubtreeQuota = 10 (own) + 5 (child) = 15
-		wantBefore := resources.FlavorResourceQuantities{
-			{Flavor: "default", Resource: corev1.ResourceCPU}: 15_000,
-		}
-		if diff := cmp.Diff(wantBefore, cache.hm.Cohort("root").getResourceNode().SubtreeQuota); diff != "" {
-			t.Errorf("before deletion, unexpected SubtreeQuota (-want,+got):\n%s", diff)
-		}
+				if err := cache.AddClusterQueue(ctx, utiltestingapi.MakeClusterQueue("cq").
+					Cohort("mid").
+					ResourceGroup(*utiltestingapi.MakeFlavorQuotas("default").Resource(corev1.ResourceCPU, "7").Obj()).
+					Obj()); err != nil {
+					t.Fatal(err)
+				}
+			},
+			deleteName: "mid",
+			wantBefore: map[kueue.CohortReference]resources.FlavorResourceQuantities{
+				"root": {
+					{Flavor: "default", Resource: corev1.ResourceCPU}: 22_000,
+				},
+			},
+			wantAfter: map[kueue.CohortReference]resources.FlavorResourceQuantities{
+				"root": {
+					{Flavor: "default", Resource: corev1.ResourceCPU}: 10_000,
+				},
+			},
+		},
+	}
 
-		cache.DeleteCohort("child")
+	for name, tc := range testCases {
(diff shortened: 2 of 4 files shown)


pkg/cache/scheduler/cache.go
```diff
@@ -560,6 +562,14 @@ func (c *Cache) DeleteCohort(cohortName kueue.CohortReference) {
 	if cohort := c.hm.Cohort(cohortName); cohort != nil {
 		updateCohortResourceNode(cohort)
 	}
+
+	// Update the SubtreeQuota of all ancestors to reflect the removal of
```
mimowo, line 566: Could you maybe open two other experiment PRs:
1. without the change, just loop the failing test 200 times to see the rate of failures
2. with the change looped 200 times to see no flake

mszadkow: Yes, good idea, will follow on that

mszadkow: https://github.com/kubernetes-sigs/kueue/pull/10584
https://github.com/kubernetes-sigs/kueue/pull/10585

mszadkow:
The tests have shown that it's not easy to reproduce the flake, both runs timeout-ed after 1h5m, without single failed test.
However I still think this PR is worthy bugfix.

mimowo:
> The tests have shown that it's not easy to reproduce the flake, both runs timeout-ed after 1h5m, without single failed test.

Is there a way to design a dedicated test which shows the issue? Is the bug showing by just running one test, or do to leaking some stuff between tests. I'm trying to understand why looping all cohorts test just for this one test.

> However I still think this PR is worthy bugfix.

Maybe, but I still don't understand what is fixed, and the lack of the release note "Does this PR introduce a user-facing change?" is not helping. 

Finally it triggers calling `updateCohortTreeResourcesIfNoCycle`, so IIUC this goes beyond just fixing the metrics. 
Basically, yes, it is likely a worthy bugfix, it must be clear what we are fixing and how.

mszadkow:
> > The tests have shown that it's not easy to reproduce the flake, both runs timeout-ed after 1h5m, without single failed test.
> 
> Is there a way to design a dedicated test which shows the issue? Is the bug showing by just running one test, or do to leaking some stuff between tests. I'm trying to understand why looping all cohorts test just for this one test.
> 
> > However I still think this PR is worthy bugfix.
> 
> Maybe, but I still don't understand what is fixed, and the lack of the release note "Does this PR introduce a user-facing change?" is not helping.
> 
> Finally it triggers calling `updateCohortTreeResourcesIfNoCycle`, so IIUC this goes beyond just fixing the metrics. Basically, yes, it is likely a worthy bugfix, it must be clear what we are fixing and how.

I actually have an idea to modify the AFS time, so it happens more frequently and we could see it triggers recording metrics on stale cache.
Will update if successful

mimowo:
> I actually have an idea to modify the AFS time, so it happens more frequently and we could see it triggers recording metrics on stale cache.

Maybe, but I'm still not understanding if we have a prod bug or not, and tweaking AFS setting my just swipe the bug under the rug. So before we do one or another change I would love to know:
1. what is the scenario when that happens, can we have a minimal reproducible scenario?
2. is this an issue with metrics, or the metrics test just uncovers a bigger problem?
3. how is that related to AFS

mszadkow:
> > I actually have an idea to modify the AFS time, so it happens more frequently and we could see it triggers recording metrics on stale cache.
> 
> Maybe, but I'm still not understanding if we have a prod bug or not, and tweaking AFS setting my just swipe the bug under the rug. So before we do one or another change I would love to know:
> 
> 1. what is the scenario when that happens, can we have a minimal reproducible scenario?
> 2. is this an issue with metrics, or the metrics test just uncovers a bigger problem?
> 3. how is that related to AFS

1. The scenario is:
After `DeleteCohort`, the hierarchy relation is updated, but the ancestor `SubtreeQuota` in cache used to stay stale.
`ClearCohortMetrics `could still publish the correct post-delete metric value, because it simulates the removal.
Later, another `RecordCohortMetrics` call would read the stale cache and publish the old value again.

I also have a minimal repro in mind at the cache/metrics layer, without relying on timing:
clear metrics for the deleted cohort, delete it, then trigger recording from a sibling/child cohort again and check that root goes back to the old quota pre-fix.

2. The metrics test is mostly uncovering `resourceNode` cache consistency issue, not just a metrics-only problem.

3. AFS is not the root cause, it just happens to be around for the integration tests, because we cloned the setup from fair_sharing.
It just increases the chance of reproducing it, because it causes frequent queue rebuild / reconcile activity, and that leads to more RecordCohortMetrics calls against stale cache

mszadkow:
OK, maybe you could ask so why is this not more frequent in the tests.
Because we verified the metrics and not what is the real value of the SubtreeQuota.
Usually metrics check happens fast enough that we don't see any issue.
And the next test step in the integration test updates the subtreeQuota by modifying cohort.

mszadkow:
This should show the issue now:
https://github.com/kubernetes-sigs/kueue/pull/10585
https://github.com/kubernetes-sigs/kueue/pull/10584

mimowo:
> The metrics test is mostly uncovering resourceNode cache consistency issue, not just a metrics-only problem.

Yeah, so I'm wondering if we could design a test which exposes the issue, and potentially even abstracts out metrics. Maybe this issue could even lead to invalid scheduling decisions is the quota for parent Cohort is not decreased after child removal, wdyt?

mszadkow:
> > The metrics test is mostly uncovering resourceNode cache consistency issue, not just a metrics-only problem.
> 
> Yeah, so I'm wondering if we could design a test which exposes the issue, and potentially even abstracts out metrics. Maybe this issue could even lead to invalid scheduling decisions is the quota for parent Cohort is not decreased after child removal, wdyt?

Sure, I will work on such test now

mszadkow:
It hardly impossible to reproduce the issue, I have tried several approaches, but no luck.
The real bug lives in the cache layer's operation ordering during DeleteCohort, which is why the focused unit test `TestDeleteCohort_DoesNotRepublishStaleAncestorQuotaMetrics` fails deterministically without the fix and catches it cleanly. 

Regarding 
> a test which exposes the issue, and potentially even abstracts out metrics

test named `"Deleting child Cohort should update borrowable resources for parent Cohort"` is provided.

mimowo:
> It hardly impossible to reproduce the issue, I have tried several approaches, but no luck.

This is surprising to me, I thought the issue would be deterministic at the quota / scheduling level, and the new test would fail regularly. 

I found similar issue fix before https://github.com/kubernetes-sigs/kueue/pull/5985/, and there we added this test: https://github.com/kubernetes-sigs/kueue/pull/5985/changes#diff-8fc55df1e8578a6eb073960491a7de3ee974e604745ac3c93a99a068176b9c8bR2687 

In any case I will try to understand what is going on with debugger today, but I'm not sure. It would be great if the problem has another pair of eyes who can investigate what is going on. Maybe @mbobrovskyi could help here.

pkg/cache/scheduler/cache_test.go
```diff
@@ -3569,6 +3569,150 @@ func TestCohortCycles(t *testing.T) {
 	})
 }
 
+func TestDeleteCohortUpdatesAncestorSubtreeQuota(t *testing.T) {
+	t.Run("deleting child cohort updates parent SubtreeQuota", func(t *testing.T) {
```
mimowo: Could you change the tests into table-driven to avoid duplicating setups and assserts? it is also more idiomatic for golang

pkg/cache/scheduler/cohort_metrics_test.go
```diff
@@ -462,3 +462,138 @@ func expectGaugeValue(t *testing.T, collector prometheus.Collector, labels map[s
 		t.Fatalf("unexpected metric value for labels %v: got=%v want=%v", labels, dps[0].Value, want)
 	}
 }
+
+func TestDeleteCohort_DoesNotRepublishStaleAncestorQuotaMetrics(t *testing.T) {
+	fixture := newCohortMetricsFixture(t)
```
mimowo: Let's make the test also table-driven to be more idiomatic for golan.

test/integration/singlecluster/scheduler/scheduler_test.go
```diff
@@ -2936,6 +2936,104 @@ var _ = ginkgo.Describe("Scheduler", func() {
...
+			chLeft  *kueue.Cohort
+			chRight *kueue.Cohort
+		)
+
+		ginkgo.AfterEach(func() {
+			util.ExpectObjectToBeDeleted(ctx, k8sClient, root, true)
+			util.ExpectObjectToBeDeleted(ctx, k8sClient, chLeft, true)
+		})
+
+		ginkgo.It("Should not admit from stale subtree quota after deleting sibling cohort", func() {
```
mimowo, line 2951: Thank you for the test. A test like this I was imagining and I found similar PR with similar test https://github.com/kubernetes-sigs/kueue/pull/5985/changes#diff-8fc55df1e8578a6eb073960491a7de3ee974e604745ac3c93a99a068176b9c8

So the test is passing after and before changes?

mimowo:
Also is this correct "Does this PR introduce a user-facing change?"? I'm surprised seeing NONE, my undestanding it was that the issue was in cache corruption that could lead to scheduling too many workloads. Please revisit if this is correct.

mimowo:
> > It hardly impossible to reproduce the issue, I have tried several approaches, but no luck.
> 
> This is surprising to me, I thought the issue would be deterministic at the quota / scheduling level, and the new test would fail regularly.
> 

Ok, I got a bit of time to run the test locally and it failed on the first attempt:

```
  [38;5;243m<< Timeline[0m

  [38;5;9m[FAILED] Timed out after 10.001s.
  Unexpected workloads are pending
      <*v1beta2.Workload | 0x18d681a28b48>: metadata:
            creationTimestamp: "2026-04-24T07:55:37Z"
            generation: 1
            managedFields:
            - apiVersion: kueue.x-k8s.io/v1beta2
              fieldsType: FieldsV1
              fieldsV1:
                f:status:
                  f:admission:
                    f:clusterQueue: {}
                    f:podSetAssignments:
                      k:{"name":"main"}:
                        .: {}
                        f:count: {}
                        f:flavors:
                          f:cpu: {}
                        f:name: {}
                        f:resourceUsage:
                          f:cpu: {}
                  f:conditions:
                    k:{"type":"Admitted"}:
                      .: {}
                      f:lastTransitionTime: {}
                      f:message: {}
                      f:observedGeneration: {}
                      f:reason: {}
                      f:status: {}
                      f:type: {}
                    k:{"type":"QuotaReserved"}:
                      .: {}
                      f:lastTransitionTime: {}
                      f:message: {}
                      f:observedGeneration: {}
                      f:reason: {}
                      f:status: {}
                      f:type: {}
              manager: kueue-admission
              operation: Apply
              subresource: status
              time: "2026-04-24T07:55:37Z"
            - apiVersion: kueue.x-k8s.io/v1beta2
              fieldsType: FieldsV1
              fieldsV1:
                f:spec:
                  .: {}
                  f:active: {}
                  f:podSets:
                    .: {}
                    k:{"name":"main"}:
                      .: {}
                      f:count: {}
                      f:name: {}
                      f:template:
                        .: {}
                        f:metadata: {}
                        f:spec:
                          .: {}
                          f:containers:
                            .: {}
                            k:{"name":"c"}:
                              .: {}
                              f:name: {}
                              f:resources:
                                .: {}
                                f:requests:
                                  .: {}
                                  f:cpu: {}
                          f:restartPolicy: {}
                  f:queueName: {}
              manager: scheduler.test
              operation: Update
              time: "2026-04-24T07:55:37Z"
            name: wl-after-delete
            namespace: core-85mxb
            resourceVersion: "263"
            uid: fc0fa817-22aa-49fb-92d3-c22e20161cfc
          spec:
            active: true
            podSets:
            - count: 1
              name: main
              template:
                metadata: {}
                spec:
                  containers:
                  - name: c
                    resources:
                      requests:
                        cpu: "8"
                  restartPolicy: Never
            queueName: cq-a
          status:
            admission:
              clusterQueue: cq-a
              podSetAssignments:
              - count: 1
                flavors:
                  cpu: on-demand
                name: main
                resourceUsage:
                  cpu: "8"
            conditions:
            - lastTransitionTime: "2026-04-24T07:55:37Z"
              message: Quota reserved in ClusterQueue cq-a
              observedGeneration: 1
              reason: QuotaReserved
              status: "True"
              type: QuotaReserved
            - lastTransitionTime: "2026-04-24T07:55:37Z"
              message: The workload is admitted
              observedGeneration: 1
              reason: Admitted
              status: "True"
              type: Admitted
          

  The function passed to Eventually failed at /usr/local/google/home/michalwozniak/go/src/sigs.k8s.io/kueue/test/util/util.go:484 with:
  Expected
      <[]types.NamespacedName | len:0, cap:1>: []
  to equal
      <[]types.NamespacedName | len:1, cap:1>: [
          {
              Namespace: "core-85mxb",
              Name: "wl-after-delete",
          },
      ][0m
  [38;5;9mIn [1m[It][0m[38;5;9m at: [1m/usr/local/google/home/michalwozniak/go/src/sigs.k8s.io/kueue/test/integration/singlecluster/scheduler/scheduler_test.go:3034[0m [38;5;243m@ 04/24/26 07:55:47.485[0m
[38;5;243m------------------------------[0m
[38;5;14mS[0m[38;5;14mS[0m[38;5;14mS[0m[38;5;14mS[0m[38;5;14mS[0m[38;5;14mS[0m[38;5;14mS[0m[38;5;14mS[0m[38;5;14mS[0m[38;5;14mS[0m[38;5;14mS[0m[38;5;14mS[0m[38;5;14mS[0m[38;5;14mS[0m[38;5;14mS[0m

[38;5;9m[1mSummarizing 1 Failure:[0m
  [38;5;9m[FAIL][0m [0mScheduler [38;5;243mwhen Deleting child Cohort should update borrowable resources for parent Cohort [38;5;9m[1m[It] Should not admit from stale subtree quota after deleting sibling cohort[0m
  [38;5;243m/usr/local/google/home/michalwozniak/go/src/sigs.k8s.io/kueue/test/integration/singlecluster/scheduler/scheduler_test.go:3034[0m

```

EDIT: I mean it failed on the first attempt with the fix reverted, when the fix is applied it passes, so all seems in great.

mszadkow:
> rising to me, I thought the issue would be deterministic at the quota / scheduling level, and the new test would fail regularly.
> 
> I found similar issue fix before #5985, and there we added this test: https://github.com/kubernetes-sigs/kueue/pull/5985/changes#diff-8fc55df1e8578a6eb073960491a7de3ee974e604745ac3c93a99a068176b9c8bR2687
> 
> In any case I will try to understand what is going on with debugger today, but I'm not sure. It would be great if the problem has another pair of eyes who can investigate what is going on. Maybe @mbobrovskyi could help here.

As I have already mentioned offline, my bad I didn't said exactly which test I had in mind.
I referenced the original flay test, not the one that was provided with the PR.
So we are on the same page now.

test/integration/singlecluster/scheduler/scheduler_test.go
mszadkow: No, that PR you linked is regarding ClusterQueue, current one is about Cohort.

pkg/cache/scheduler/cohort_metrics_test.go
mszadkow: I did that, but also changing this one test into table driven didn't bring any value, so I decided to combine the tests with similar(quite similar) setups into one

mimowo (commented):
Thank you 👍 The new version looks great, and I'm happy we have a reliable repro 👍 
/lgtm
/approve
/cherrypick release-0.17
/unhold

mimowo:
/release-note-edit
```release-note
Fixed a bug in Kueue's cache that could leave stale SubtreeQuota values in ancestor cohorts after a child Cohort
was deleted, leading to potential over-admission of workloads and incorrect metrics reporting.
```

mimowo:
/release-note-none
As this is getting reverted in https://github.com/kubernetes-sigs/kueue/pull/10776. Let's set the release note in the follow up PR once fixed

mimowo:
/release-note-edit
```release-note
NONE
```

PR #10587, opened 2026-04-17
Reduce informer cache memory by stripping non-scheduling fields from Workloads and TAS node labels

shshr:
#### What type of PR is this?

/kind feature
/area tas

This PR reduces the memory footprint of Kueue's informer caches through two complementary optimizations:

**1. Strip non-scheduling fields from cached Workloads**

Registers a `TransformFunc` on the Workload informer that removes PodTemplateSpec fields Kueue never reads at runtime. For each PodSet template, only scheduling-relevant fields are retained:

- **ObjectMeta**: Labels, Annotations
- **PodSpec**: NodeSelector, Affinity, Tolerations, SchedulingGates, RuntimeClassName, Overhead, ResourceClaims, TopologySpreadConstraints, Priority
- **Containers/InitContainers**: Name, Resources, Ports only

Fields like Image, Command, Args, Env, VolumeMounts, and Volumes are stripped since Kueue only needs resource requests and scheduling constraints for admission and preemption decisions.

The PodSet equality comparison (`ComparePodSets`) is updated to compare only container Resources instead of full container specs, consistent with the stripped cache entries.

**2. Configurable node label prefix filtering for TAS node cache**

Adds a `Resources.ExcludeNodeLabelPrefixes` configuration field that strips infrastructure labels from cached node objects in the TAS node cache. Nodes in large clusters can carry dozens of labels set by cloud providers and infrastructure controllers (e.g., `kubectl.kubernetes.io/`, `cloud.google.com/`, `eks.amazonaws.com/`) that are irrelevant for topology, flavor, or workload scheduling decisions.

A sensible set of default prefixes is provided. Operators can override the list via the Kueue configuration to match their environment.

In large clusters with thousands of Workloads and hundreds of nodes, informer caches become the dominant memory consumer. A significant portion of the cached data consists of fields Kueue never inspects. Stripping these fields at ingestion time reduces per-object memory without any behavioral change.

NONE

The two optimizations are independent and target different cache domains:
- Commit 1 targets the Workload informer cache (`pkg/config/config.go`, `pkg/util/equality/podset.go`)
- Commit 2 targets the TAS node cache (`pkg/cache/scheduler/tas_flavor.go`, `apis/config/v1beta2/`)

They can be reviewed and tested independently. Happy to split into separate PRs if preferred.

```release-note
Added Resources.ExcludeNodeLabelPrefixes configuration field to strip infrastructure labels from the TAS node cache, reducing memory usage. A default set of common infrastructure label prefixes is provided. Added informer TransformFunc to strip non-scheduling PodTemplateSpec fields from cached Workloads, reducing per-workload memory footprint.
```


apis/config/v1beta2/configuration_types.go
```diff
@@ -512,6 +512,18 @@ type Resources struct {
...
+	// stripped from cached node objects in the Topology Aware Scheduling
+	// (TAS) node cache to reduce memory usage. Any node label whose key
+	// starts with one of these prefixes is dropped when the node is
+	// stored in the TAS cache. Labels needed for topology levels, flavor
+	// node selectors, workload node selectors, and node affinity should
+	// NOT be listed here.
+	// Defaults to a set of common infrastructure labels that are not
+	// relevant for scheduling decisions (see defaults.go).
+	// +optional
+	ExcludeNodeLabelPrefixes []string `json:"excludeNodeLabelPrefixes,omitempty"`
```
mimowo, line 525: The API changes should first go via KEP design phase. Maybe it is worth de-coupling some changes which don't necessarily need API-level changes?

shshr: Sure that's a good point. I can send out separate PRs

shshr:
Closing this PR to split into two separate PRs per reviewer feedback: one for the non-API cache optimization (stripping non-scheduling PodTemplateSpec fields via TransformFunc), and one for the KEP + API change (ExcludeNodeLabelPrefixes).

PR #10590, opened 2026-04-17
Strip non-scheduling PodTemplateSpec fields from Workload informer cache

shshr:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

Registers a `TransformFunc` on the Workload informer that removes PodTemplateSpec fields Kueue never reads at runtime, reducing per-Workload memory in the informer cache.

For each PodSet template, only scheduling-relevant fields are retained:

- **ObjectMeta**: Labels, Annotations
- **PodSpec**: NodeSelector, Affinity, Tolerations, SchedulingGates, RuntimeClassName, Overhead, ResourceClaims, TopologySpreadConstraints, Priority
- **Containers/InitContainers**: Name, Resources, Ports only

Fields like Image, Command, Args, Env, VolumeMounts, and Volumes are stripped since Kueue only needs resource requests and scheduling constraints for admission and preemption decisions.

The PodSet equality comparison (`ComparePodSets`) is updated to compare only container Resources instead of full container specs, consistent with the stripped cache entries.

In large clusters with thousands of Workloads, informer caches become a dominant memory consumer. A significant portion of the cached PodTemplateSpec consists of fields Kueue never inspects. Stripping these fields at ingestion time reduces per-object memory without any behavioral change.

This PR contains **no API changes**—only internal cache optimization.

#### Which issue(s) this PR fixes:

NONE

#### Special notes for your reviewer:

This is split from #10587 per reviewer feedback to decouple non-API changes from API-level changes that require a KEP. The companion KEP + API change (ExcludeNodeLabelPrefixes for TAS node cache) will be submitted separately.

Changes are in 4 files:
- `cmd/kueue/main.go` – register TransformFunc on Workload informer
- `pkg/config/config.go` – implement `StripWorkloadTransform`
- `pkg/util/equality/podset.go` – update `comparePodTemplate` to compare only container Resources
- `test/util/equality/podset_test.go` – updated tests

#### Does this PR introduce a user-facing change?

```release-note
Added informer TransformFunc to strip non-scheduling PodTemplateSpec fields (Image, Command, Args, Env, VolumeMounts, Volumes, etc.) from cached Workloads, reducing per-workload memory footprint in large clusters.
```

PR #10591, opened 2026-04-17
KEP-10587: Configurable node label prefix filtering for TAS cache

shshr:
#### What type of PR is this?

/kind feature
/kind kep
/area tas

This PR proposes **KEP-10587** and includes its reference implementation for a new `Resources.ExcludeNodeLabelPrefixes` configuration field. The field controls which node label key prefixes are stripped from cached node objects in the Topology Aware Scheduling (TAS) node cache, reducing memory usage.

Nodes in large clusters carry 30–80+ labels injected by cloud providers and infrastructure controllers (e.g., `kubectl.kubernetes.io/`, `cloud.google.com/`, `eks.amazonaws.com/`). These labels are irrelevant for topology, flavor, or workload scheduling decisions but consume significant memory in the TAS node cache.

**What's included:**

1. **KEP-10587** (`keps/10587-exclude-node-label-prefixes/`) — full design proposal with motivation, API design, defaults, graduation criteria (alpha → beta → GA), alternatives analysis
2. **Implementation** — adds `ExcludeNodeLabelPrefixes` to `Resources` in `apis/config/v1beta2/configuration_types.go`, default prefix list in `defaults.go`, and filtering logic in `pkg/cache/scheduler/tas_flavor.go`

The companion non-API change (stripping non-scheduling PodTemplateSpec fields from Workload cache) is submitted separately in #10590.

NONE

This is split from #10587 per reviewer feedback to decouple API changes (which need KEP design review) from non-API optimizations.

The KEP follows the pattern established by the existing `ExcludeResourcePrefixes` field and proposes alpha → beta → GA graduation with a feature gate (`ExcludeNodeLabelPrefixes`). At beta, startup validation will warn when excluded prefixes overlap with configured topology level labels.

Happy to iterate on the KEP design before merging the implementation.

```release-note
KEP-10587: Proposed Resources.ExcludeNodeLabelPrefixes configuration field to strip infrastructure labels from the TAS node cache, reducing per-node memory usage. A default set of common infrastructure label prefixes (kubectl.kubernetes.io/, cloud.google.com/, eks.amazonaws.com/, etc.) is provided.
```


keps/10587-exclude-node-label-prefixes/README.md
```diff
@@ -0,0 +1,318 @@
...
+affecting scheduling correctness, because the excluded labels are infrastructure
+metadata that Kueue never uses for topology levels, flavor node selectors, or
+workload scheduling decisions.
+
+## Motivation
+
+In large clusters (hundreds to thousands of nodes), each node can carry 30–80+
+labels injected by cloud providers and infrastructure controllers. Examples
+include `kubectl.kubernetes.io/`, `cloud.google.com/`, `eks.amazonaws.com/`,
+and `node.cluster.x-k8s.io/` prefixes. The TAS node cache stores all labels on
```
mwielgus, line 50: Not sure about other cloud providers but on GCP topological labels are prefixed with cloud.google.com (exmple: "cloud.google.com/gce-topology-block"). So using prefixes, it may be difficult to tell them from labels that are not important for TAS.

keps/10587-exclude-node-label-prefixes/README.md
```diff
@@ -0,0 +1,318 @@
...
+Combined with a complementary Workload cache optimization (stripping
+non-scheduling PodTemplateSpec fields), excluding infrastructure node labels
+can reduce controller-manager RSS by ~10%.
+
+This is analogous to the existing `Resources.ExcludeResourcePrefixes` field,
+which strips irrelevant resource types from quota calculations. The same
+pattern—a denylist of key prefixes with sensible defaults—is applied here to
+node labels.
+
+### Goals
```
mwielgus, line 64: What will happen if a workload has antiaffinity on a excluded label? How can this be prevented or at least explicitely reported?

keps/10587-exclude-node-label-prefixes/README.md
```diff
@@ -0,0 +1,318 @@
...
+  to the default prefix list; an explicit empty list (`[]`) disables
+  filtering entirely.
+
+### Non-Goals
+
+* Filtering labels from node objects outside the TAS cache (e.g., in the
+  Kubernetes API server or other Kueue caches).
+* Allowlist-based filtering (only keep certain prefixes). This could be a
+  future enhancement if needed.
+* Filtering annotations or taints from cached nodes.
```
mwielgus, line 82: Why not filter annotations as well?

mwielgus:
And please sign the CLA.

shshr:
For some reason, CLA signing is not reflecting here. I have signed it a couple times.

kannon92:
Check your commits. You have two authors on your first one.

keps/10587-exclude-node-label-prefixes/README.md
shshr: @mwielgus Good point. Maybe prefix-based exclusion is too coarse.

Two mitigations worth considering:

- **Support exact label names in addition to prefixes** - so operators can exclude `cloud.google.com/machine-family` without affecting `cloud.google.com/gce-topology-block`.

- **Add an allowlist option (`includeNodeLabelPrefixes`)** - operators declare which label prefixes TAS should keep rather than trying to enumerate what to exclude. Since operators already know which topology keys they configure in their Topology objects, an allowlist is more natural and less error-prone.

keps/10587-exclude-node-label-prefixes/README.md
shshr: Thinking of 2 approaches here:

- **Admission-time validation**: when a workload is submitted, check whether any affinity/anti-affinity `topologyKey` values match an excluded prefix. If so, reject with a clear error message like `topologyKey X matches excluded node label prefix Y`. This gives operators immediate, actionable feedback.
 
- **An allowlist approach** (`includeNodeLabelPrefixes`) should largely address this by design - operators explicitly declare which label prefixes to keep, and they'd include any prefixes used as topology keys. If an operator omits a prefix their workloads reference, admission-time validation can catch it: check whether any affinity/anti-affinity `topologyKey` values are absent from the included prefixes and reject with a clear error like `topologyKey X does not match any included node label prefix`.

This approach seems safer than the exclude-list approach where operators might not realize they've excluded something their workloads need.

keps/10587-exclude-node-label-prefixes/README.md
shshr: Good question. They could be stripped unconditionally from cached nodes without any prefix configuration needed. I'll expand the scope of this KEP to include annotation stripping on cached nodes as part of the same feature.

keps/10587-exclude-node-label-prefixes/README.md
shshr: @mwielgus Would like to know your thoughts on this before updating the KEP

PR #10594, opened 2026-04-17
fix cel expressions for v1beta1 if priorityClassRef doesn't exist

### LINKED ISSUE #10593: v1beta1 CEL validation fails on v1beta2-stored workloads: no such key: priorityClassSource
**What happened**:

Patching a Kueue Workload via the `v1beta1` API fails with a CEL validation error on clusters where the Workload CRD storage version is `v1beta2`.

The error:
```
The Workload "<name>" is invalid: <nil>: Invalid value: "object": no such key: priorityClassSource evaluating rule: priorityClassSource is immutable while workload quota reserved
```

This affects any PATCH or UPDATE to an admitted workload (one with `QuotaReserved: True`) when accessed through the `v1beta1` API endpoint. The same operation via `v1beta2` succeeds.

Patching a workload via the `v1beta1` API (e.g., toggling `/spec/active`) should succeed, even when the workload is stored as `v1beta2`. The CEL validation rules on `v1beta1` should account for the fact that `spec.priorityClassSource` may not exist on objects originally stored as `v1beta2` (where it was replaced by `spec.priorityClassRef`).

1. Deploy a cluster with Kueue 0.16 or 0.17 (v1beta2 storage version)
2. Create a ResourceFlavor, ClusterQueue, and LocalQueue
3. Submit a workload (e.g., a Job) with a `kueue.x-k8s.io/queue-name` label
4. Wait for the workload to be admitted (`QuotaReserved: True`)
5. Attempt to patch the workload via the v1beta1 API:

```bash
kubectl patch workloads.v1beta1.kueue.x-k8s.io <workload-name> -n <namespace> \
  --type='json' \
  -p='[{"op":"replace","path":"/spec/active","value":false}]' \
  --dry-run=server
```

This fails. The same patch via v1beta2 succeeds:

```bash
kubectl patch workloads.v1beta2.kueue.x-k8s.io <workload-name> -n <namespace> \
  --type='json' \
  -p='[{"op":"replace","path":"/spec/active","value":false}]' \
  --dry-run=server
```

**Root cause**:

The v1beta1 `Workload` type has a CEL validation rule at [`apis/kueue/v1beta1/workload_types.go:790`](https://github.com/kubernetes-sigs/kueue/blob/main/apis/kueue/v1beta1/workload_types.go#L790):

```
rule: (has(oldSelf.status) && has(oldSelf.status.conditions) && oldSelf.status.conditions.exists(c, c.type == 'QuotaReserved' && c.status == 'True')) ? (oldSelf.spec.priorityClassSource == self.spec.priorityClassSource) : true
message: priorityClassSource is immutable while workload quota reserved
```

This rule unconditionally accesses `oldSelf.spec.priorityClassSource` and `self.spec.priorityClassSource` when `QuotaReserved` is `True`. However, workloads stored as `v1beta2` don't have `spec.priorityClassSource` (it was replaced by `spec.priorityClassRef`). When the API server converts the stored v1beta2 object to v1beta1 for validation, the `priorityClassSource` field may not be present, causing the CEL evaluation to fail with `no such key`.

The same issue likely affects the adjacent CEL rule at line 791 which references `spec.priorityClassName` (also removed in v1beta2, replaced by `spec.priorityClassRef.name`).

**Suggested fix**:

Guard the field accesses with `has()` checks:

```
rule: (has(oldSelf.status) && has(oldSelf.status.conditions) && oldSelf.status.conditions.exists(c, c.type == 'QuotaReserved' && c.status == 'True') && has(oldSelf.spec.priorityClassSource) && has(self.spec.priorityClassSource)) ? (oldSelf.spec.priorityClassSource == self.spec.priorityClassSource) : true
```

**Environment**:
- Kueue version: 0.16.5 (Red Hat Build of Kueue)
- Kubernetes version: 4.x (OpenShift)
- CRD storage version: v1beta2

kannon92:
#### What type of PR is this?
/kind bug

Fix CEL expressions in workload api to check for nillness before doing the equality check. 

Fixes #10593 

```release-note
Fixed a bug where admitted Workloads could fail to patch through the v1beta1 API due to CEL validation of the `priorityClassSource` immutability rule.
```


kannon92:
/assign @mimowo @tenzen-y 

We got this from a user so I'd like to cherry-pick this.

test/integration/singlecluster/webhook/core/workload_v1beta1_cel_test.go
tenzen-y, line 1: can you move this tests to https://github.com/kubernetes-sigs/kueue/blob/52b6722eedf549171280a544ddd20d5e939d4d34/test/integration/singlecluster/webhook/core/workload_test.go?

We actually have both webhooks and CEL tests in a single RESOURCE_NAME_test.go file.

tenzen-y (commented):
Basically, LGTM
Thank you!

test/integration/singlecluster/webhook/core/workload_v1beta1_cel_test.go
kannon92: updated.

tenzen-y (commented):
Thank you!
/lgtm
/approve

mimowo:
Thanks 👍

mimowo:
/release-note-edit
```release-note
Fixed a bug where admitted Workloads could fail to patch through the v1beta1 API due to CEL validation of the `priorityClassSource` immutability rule.
```

Issue #6587, opened 2025-08-15
Wall time limits for ClusterQueues


kannon92:
**What would you like to be added**:

I would like to add a way to specify ClusterQueues and set a limit on wall time.

It would be ideal to be able to limit wall time per resourceFlavor in the ClusterQueue. ClusterQueues will then track the usage of walltime.

```yaml
type ClusterQueueSpec struct {
...
        // wallTimePolicy defines the wallTimePolicy for the ClusterQueue.
        // +optional
        WallTimePolicy *WallTimePolicy `json:"wallTimePolicy,omitempty"`
}
...
type WallTimePolicy struct {
        // WallTimeFlavors describes a group of resources that this wall time limits applies to.
        // flavors is the list of flavors that provide the resources of this group.
        // Typically, different flavors represent different hardware models
        // (e.g., gpu models, cpu architectures) or pricing models (on-demand vs spot
        // cpus).
        // Each flavor MUST list all the resources listed for this group in the same
        // order as the .resources field.
        // The list cannot be empty and it can contain up to 16 flavors.
        // +listType=map
        // +listMapKey=name
        // +kubebuilder:validation:MinItems=1
        // +kubebuilder:validation:MaxItems=16
        WallTimeFlavors []WallTimeFlavor `json:"wallTimeFlavors"`
}

type WallTimeFlavor struct {
        // name of this flavor. The name should match the .metadata.name of a
        // ResourceFlavor. If a matching ResourceFlavor does not exist, the
        // ClusterQueue will have an Active condition set to False.
        Name ResourceFlavorReference `json:"name"`

        // wallTimeAllocatedHours is the number of hours that this wall time quota applies to.
        // +kubebuilder:validation:Minimum=1
        WallTimeAllocatedHours int32 `json:"wallTimeAllocatedHours"`

        // actionWhenWallTimeExhausted defines the action to take when the budget is exhausted.
        // The possible values are:
        // +kubebuilder:validation:Enum=Hold;HoldAndDrain
        // +kubebuilder:default="Hold"
        ActionWhenWallTimeExhausted StopPolicy `json:"actionWhenWallTimeExhausted,omitempty"`
}

type ClusterQueueStatus struct{
...
        // wallTimeFlavorUsage contains the current wall time usage for this ClusterQueue.
        // +optional
        // +listType=map
        // +listMapKey=name
        // +kubebuilder:validation:MaxItems=16
        // +optional
[...]

PR #10479, opened 2026-04-14
Add ClusterQueue default maximum execution time (KEP-7029)

kannon92:
#### What type of PR is this?

/kind feature
/kind api-change
/kind kep

#### What this PR does / why we need it:

Implement KEP-7029.
#### Which issue(s) this PR fixes:

Implements https://github.com/kubernetes-sigs/kueue/issues/7029

#### Does this PR introduce a user-facing change?

```release-note
Add `spec.workloadDefaults.maximumExecutionTimeSeconds` field to ClusterQueue, allowing administrators to set a default maximum execution time for workloads submitted to a ClusterQueue.
```

---
> **AI Assistance Disclosure**: This pull request was created with the assistance of AI (Claude Code by Anthropic).

## Summary by CodeRabbit

* **New Features**
  * ClusterQueues can define a default maximum execution time for workloads.
  * Defaults apply when enabled and are used only when a workload does not specify its own value.
  * Workload-specific labels override ClusterQueue defaults.
  * Workloads exceeding the configured limit are prevented from admission.
  * Added the alpha `ClusterQueueMaxExecutionTime` feature gate, disabled by default.

* **Documentation**
  * Updated API reference and feature-gate documentation for ClusterQueue workload defaults.

PR #10596, opened 2026-04-17
KEP-7029: ClusterQueue Default Maximum Execution Time

### LINKED ISSUE #7029: [Feature]: Provide a default timeout on a job per local queue

I would like the ability to add add a timeout to jobs if they submit to certain local queues.

We would set https://kueue.sigs.k8s.io/docs/concepts/workload/#maximum-execution-time based on the local queue.

I think we could add a timeout section on the LocalQueue API and that can automatically set a timeout. If the timeout is specified in the job, we can ignore this.

**Why is this needed**:

Requested in two separate issues.

- https://github.com/kubernetes-sigs/kueue/discussions/6684

- https://github.com/kubernetes-sigs/kueue/issues/6587#issuecomment-3342233971

**Completion requirements**:

- [x] Design doc
- [x] API change
- [x] Docs update

kannon92:
#### What type of PR is this?

/kind kep

Adds KEP-7029 proposing the ability to configure a default maximum execution time on a ClusterQueue. When a job is submitted to a LocalQueue backed by a ClusterQueue with a default maximum execution time configured, the resulting workload will automatically have its `maximumExecutionTimeSeconds` field set, unless the job already specifies a value via the `kueue.x-k8s.io/max-exec-time-seconds` label.

This builds on KEP-3125 (maximum execution time) and addresses community requests in:
- https://github.com/kubernetes-sigs/kueue/discussions/6684
- https://github.com/kubernetes-sigs/kueue/issues/6587#issuecomment-3342233971

Fixes #7029

This is the KEP only. Implementation will follow in a separate PR.

The original proposal placed the default at the LocalQueue level. Based on review feedback (most configuration lives at the ClusterQueue level, and LocalQueue acts primarily as a pointer), the design was moved to the ClusterQueue level. LocalQueue-level defaults are documented as a future extension.

```release-note
NONE
```

## Summary by CodeRabbit

* **Documentation**
  * Added KEP-7029 documentation for ClusterQueue maximum execution time, including defaulting, precedence rules, and non-retroactive behavior.
  * Documented admission enforcement when job-specified timeouts exceed ClusterQueue limits, including dedicated quota-related messaging and inadmissible handling to prevent retry loops.
  * Added KEP metadata for the alpha feature gate `ClusterQueueMaxExecutionTime`, including stage details and graduation criteria.

### DIFF
--- a/keps/7029-clusterqueue-default-execution-time/README.md
+++ b/keps/7029-clusterqueue-default-execution-time/README.md
@@ -1,0 +1,513 @@
+# KEP-7029: ClusterQueue Maximum Execution Time
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+    - [Story 3](#story-3)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API](#api)
+    - [ClusterQueue Spec](#clusterqueue-spec)
+  - [Controller](#controller)
+    - [Workload Creation](#workload-creation)
+    - [Admission Enforcement](#admission-enforcement)
+    - [Workload Equivalency Check](#workload-equivalency-check)
+  - [Precedence Rules](#precedence-rules)
+  - [Why Workload, Not Job](#why-workload-not-job)
+  - [Specific Field vs. Generic Defaulting](#specific-field-vs-generic-defaulting)
+  - [Test Plan](#test-plan)
+    - [Prerequisite testing updates](#prerequisite-testing-updates)
+    - [Unit Tests](#unit-tests)
+    - [Integration tests](#integration-tests)
+      - [controller/jobs/job](#controllerjobsjob)
+      - [webhook/core/clusterqueue](#webhookcoreclusterqueue)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Future Extensions](#future-extensions)
+  - [LocalQueue-Level Defaults](#localqueue-level-defaults)
+  - [Default execution time](#default-execution-time)
+<!-- /toc -->
+
+## Summary
+
+Add the ability to configure a maximum execution time on a ClusterQueue that
+serves as both a **default** and an **upper bound** for workloads submitted to
+LocalQueues backed by that ClusterQueue.
+
+When a job is submitted without a `kueue.x-k8s.io/max-exec-time-seconds` label,
+the ClusterQueue's value is applied as the workload's
+`maximumExecutionTimeSeconds`. When a job *does* specify its own value via the
+label, the ClusterQueue's value acts as a ceiling: if the job's value exceeds
+the ClusterQueue's limit, the workload is kept in a pending state and is not
+admitted.
+
+This is consistent with how Slurm enforces partition-level time limits as upper
+bounds that cannot be exceeded by individual jobs ([Slurm Resource Limits](https://slurm.schedmd.com/resource_limits.html)).
+
+This builds on the existing maximum execution time feature introduced in
+[KEP-3125](../3125-maximum-execution-time/README.md).
+
+## Motivation
+
+Cluster administrators often want to enforce time limits on jobs submitted to
+specific queues. Today, setting the maximum execution time requires each job to
+carry the `kueue.x-k8s.io/max-exec-time-seconds` label. This places the burden
+on individual users to remember to set the label, and there is no way for an
+administrator to enforce a default or maximum timeout at the queue level.
+
+Most Kueue configuration lives at the ClusterQueue level, and administrators
+typically manage policies there. Adding the maximum execution time to the
+ClusterQueue is consistent with this pattern and addresses two common cases:
+ensuring all jobs have a timeout (default), and preventing users from requesting
+excessive runtimes (upper bound).
+
+This has been requested by the community in multiple discussions:
+
+- https://github.com/kubernetes-sigs/kueue/discussions/6684
+- https://github.com/kubernetes-sigs/kueue/issues/6587#issuecomment-3342233971
+
+### Goals
+
+- Allow cluster administrators to configure a maximum execution time on a
+  ClusterQueue via a new field in `ClusterQueueSpec`.
+- Automatically apply this value as a default to workloads created for jobs
+  submitted to any LocalQueue backed by the ClusterQueue, when the job does not
+  specify its own maximum execution time.
+- Enforce this value as an upper bound: workloads whose job-specified maximum
+  execution time exceeds the ClusterQueue's limit are not admitted and remain
+  pending.
+- Preserve backward compatibility: when the ClusterQueue does not set this field,
+  behavior is unchanged.
+
+### Non-Goals
+
+- Overriding (silently reducing) a job-level timeout to match the ClusterQueue
+  limit. When the job exceeds the limit, the workload is kept pending rather
+  than having its timeout silently modified.
+- Changing the scheduling behavior based on the maximum execution time.
+- Adding the default timeout at the LocalQueue level. This may be considered in
+  a future enhancement (see [Future Extensions](#future-extensions)).
+
+## Proposal
+
+Add a new optional field `maximumExecutionTimeSeconds` inside a
+`workloadDefaults` struct on `ClusterQueueSpec`. This value serves two purposes:
+
+1. **Default**: When a workload is created for a job that does not have the
+   `kueue.x-k8s.io/max-exec-time-seconds` label, the workload's
+   `spec.maximumExecutionTimeSeconds` is populated with the ClusterQueue's value.
+
+2. **Upper bound**: When a job specifies a maximum execution time via the label
+   that exceeds the ClusterQueue's value, the workload is not admitted and
+   remains pending. An event and condition are set on the workload to inform the
+   user that the requested execution time exceeds the ClusterQueue's limit.
+
+The existing enforcement mechanism in the workload controller
+(`reconcileMaxExecutionTime`) handles timeout expiry — no changes are needed
+there.
+
+### User Stories
+
+#### Story 1
+
+As a cluster administrator, I want to set a default maximum execution time of
+1 hour on a ClusterQueue used for interactive workloads, so that forgotten or
+runaway jobs are automatically cleaned up without requiring each user to set
+the timeout label on their jobs.
+
+#### Story 2
+
+As a batch user submitting jobs to a queue backed by a ClusterQueue with a
+maximum execution time of 1 hour, I want to set
+`kueue.x-k8s.io/max-exec-time-seconds` to 1800 (30 minutes) on my job, so that
+my job is cleaned up sooner than the queue default.
+
+#### Story 3
+
+As a cluster administrator, I need to enforce a maximum execution time limit on jobs. When a user specifies`kueue.x-k8s.io/max-exec-time-seconds` (e.g., 7200 for 2 hours) on a job submitted to a queue limited to 1 hour, the workload must remain pending, accompanied by a clear message guiding the user to correct their request.
+
+### Notes/Constraints/Caveats
+
+- The default is applied at workload creation time. Existing workloads are not
+  retroactively updated in the following scenarios:
+
+  - **ClusterQueue's timeout field is modified**: If an administrator changes the
+    `maximumExecutionTimeSeconds` value on a ClusterQueue, existing workloads keep
+    their original timeout. Only newly created workloads pick up the new value.
+
+  - **Job moves to a different LocalQueue**: If a user changes the
+    `kueue.x-k8s.io/queue-name` label on a Job to point to a different
+    LocalQueue (potentially backed by a different ClusterQueue), the workload's
+    `QueueName` is updated in-place, but its `MaximumExecutionTimeSeconds` is not
+    recalculated from the new ClusterQueue. This is intentional: the timeout was
+    a default applied at creation, and changing it mid-flight could be disruptive
+    (e.g., a workload admitted with a 1-hour timeout should not suddenly receive
+    a 15-minute timeout from the new queue). This is consistent with existing
+    behavior where the queue change path only updates the `QueueName` field on
+    the workload. If the workload is recreated (e.g., due to job spec changes
+    that trigger non-equivalency), the new ClusterQueue's default would apply.
+
+  - **Readmission**: Changing the queue label does not trigger readmission today.
+    This is existing behavior, and this KEP does not change it. The timeout
+    inherited at creation time persists through queue changes.
+
+- Prebuilt workloads are not affected. They are expected to be fully specified
+  externally.
+
+### Risks and Mitigations
+
+- **Risk**: An API call to look up the ClusterQueue during workload
+  creation adds latency to the reconcile loop.
+  - **Mitigation**: The lookup is a single GET call that only happens once per
+    workload creation. The ClusterQueue is likely cached by the
+    controller-runtime client cache. The performance impact is negligible.
+
+## Design Details
+
+### API
+
+#### ClusterQueue Spec
+
+Add a new optional `workloadDefaults` field to `ClusterQueueSpec`.
+This struct groups all default values that a
+ClusterQueue can apply to workloads, making it straightforward to add future
+defaults (e.g., priority class, TAS configuration) without cluttering the
+top-level spec:
+
+```go
+// ClusterQueueSpec defines the desired state of ClusterQueue
+type ClusterQueueSpec struct {
+    // ...existing fields...
+
+    // workloadDefaults defines default values that are applied to workloads
+    // submitted to LocalQueues backed by this ClusterQueue when the workload
+    // does not already specify them.
+    // +optional
+    WorkloadDefaults *ClusterQueueWorkloadDefaults `json:"workloadDefaults,omitempty"`
+}
+
+// ClusterQueueWorkloadDefaults defines default values that are applied to
+// workloads submitted to LocalQueues backed by this ClusterQueue when the
+// workload does not already specify them.
+type ClusterQueueWorkloadDefaults struct {
+    // maximumExecutionTimeSeconds if provided, determines the default maximum
+    // time, in seconds, for workloads submitted to LocalQueues backed by this
+    // ClusterQueue.
+    // This value is used when the job does not already specify a maximum execution
+    // time via the kueue.x-k8s.io/max-exec-time-seconds label.
+    //
+    // +optional
+    // +kubebuilder:validation:Minimum=1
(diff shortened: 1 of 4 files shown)


keps/7029-localqueue-default-execution-time/README.md
```diff
@@ -0,0 +1,427 @@
...
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Future Extensions](#future-extensions)
+  - [Consolidating LocalQueue Defaults](#consolidating-localqueue-defaults)
+<!-- /toc -->
+
+## Summary
+
+Add the ability to configure a default maximum execution time on a LocalQueue.
```
mwielgus, line 40: We have most of the configuration on the CQ level and very very few on LQ, which acts mainly as a pointer to CQ. Why do we need to have it on LQ? How often do you face use cases where multiple LQ pointing to the same CQ would have significantly different max exec time?

kannon92: Good question!

For timelimits I could see value in having it in both LQ/CQ.

Other values, like workload priority class or TAS settings, I'm not sure they should be at the CQ level.

Drawing parallels to slurm, it looks like they have time limits at the partition level (I think thats CQ) but you can also set timelimits for users I think.

WDYT @tenzen-y @mimowo @gabesaba?

kannon92: @mukund-wayve 

I saw you +1 this PR so curious if you have any thoughts on where this config should live.

kannon92: Updated the KEP to be for CQ instead of LQ for now.

keps/7029-clusterqueue-default-execution-time/kep.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+see-also:
+  - "/keps/3125-maximum-execution-time"
+
+stage: alpha
+
+latest-milestone: "v0.18"
+
+milestone:
+  alpha: "v0.18"
```
kannon92:
```suggestion
  alpha: "v0.19"
```

keps/7029-clusterqueue-default-execution-time/kep.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - TBD
+approvers:
+  - TBD
+
+see-also:
+  - "/keps/3125-maximum-execution-time"
+
+stage: alpha
+
+latest-milestone: "v0.18"
```
kannon92:
```suggestion
latest-milestone: "v0.19"
```

kannon92:
friendly ping for review: @tenzen-y @mimowo @mwielgus

kannon92:
cc @sohankunkerkar @amy 

anyone able to help with review?

kannon92:
/hold

Looking at Slurm, it does turn out that they enforce that users can't go over the CQ timelimit.

> Note: The precedence order specified above is respected except for the following limits: Max[Time|Wall], [Min|Max]Nodes. For these limits, even if the job is enforced with QOS and/or Association limits, it can't go over the limit imposed at Partition level, even if it listed at the bottom. So the default for these 3 types of limits is that they are upper bound by the Partition one.

ref: https://slurm.schedmd.com/resource_limits.html#:~:text=Note%3A%20The%20precedence%20order%20specified,it%20listed%20at%20the%20bottom.&text=This%20Partition%20level%20bound%20can,level%20respecting%20the%20order%20above.

I think this makes sense to me. The change would be that a user can specify a time limit on their workload but it cannot be less than ClusterQueue timelimit. If it is, then the workload will not be admitted. We can't really enforce this at creation time so I think keeping the workload in pending makes sense.

kannon92:
Anyone up for a review?

keps/7029-clusterqueue-default-execution-time/README.md
```diff
@@ -0,0 +1,456 @@
...
+   indicating that the requested execution time exceeds the ClusterQueue's limit
+   (e.g., "requested maximumExecutionTimeSeconds 7200 exceeds ClusterQueue limit
+   of 3600").
+3. An event is emitted on the workload to surface the issue to the user.
+
+This check cannot be enforced at workload creation time because the ClusterQueue
+assignment may not yet be resolved. The scheduler is the appropriate place
+because it already has the ClusterQueue context and performs admission checks.
+
+If the user corrects the job (e.g., by reducing or removing the label), the
```
ArangoGutierrez: This says a stuck workload recovers when the user corrects the job by reducing or removing the label, but the equivalency fix described below only compares `maximumExecutionTimeSeconds` when the label is present - and after removal it is absent. So removing the label never triggers recreation and the workload stays pending with the stale over-limit value; worth specifying how label removal forces re-evaluation.

keps/7029-clusterqueue-default-execution-time/README.md
```diff
@@ -0,0 +1,456 @@
...
+1. The workload remains in the queue in a pending state.
+2. A condition of type `Inadmissible` is set on the workload with a message
+   indicating that the requested execution time exceeds the ClusterQueue's limit
+   (e.g., "requested maximumExecutionTimeSeconds 7200 exceeds ClusterQueue limit
+   of 3600").
+3. An event is emitted on the workload to surface the issue to the user.
+
+This check cannot be enforced at workload creation time because the ClusterQueue
+assignment may not yet be resolved. The scheduler is the appropriate place
+because it already has the ClusterQueue context and performs admission checks.
```
ArangoGutierrez, line 279: The scheduler's view of a ClusterQueue is the cache snapshot, which doesn't carry `workloadDefaults` today, so this assumption doesn't hold yet. A short cache/snapshot plumbing note, plus which requeue class an over-limit workload lands in (inadmissible rather than backoff retry, to avoid a hot requeue loop), would make this section implementable.

keps/7029-clusterqueue-default-execution-time/README.md
```diff
@@ -0,0 +1,456 @@
...
+
+#### Admission Enforcement
+
+During the admission cycle, the scheduler checks whether the workload's
+`maximumExecutionTimeSeconds` exceeds the ClusterQueue's
+`workloadDefaults.maximumExecutionTimeSeconds`. If it does, the workload is not
+admitted. Instead:
+
+1. The workload remains in the queue in a pending state.
+2. A condition of type `Inadmissible` is set on the workload with a message
```
ArangoGutierrez: `Inadmissible` exists today as a Reason on the `QuotaReserved=False` condition, documented as LocalQueue/ClusterQueue missing or inactive - a different failure mode than exceeding a time limit. Suggest a new granular `QuotaReserved=False` reason (e.g. `MaximumExecutionTimeExceeded`) following the KEP-10852 pattern, and rewording this bullet since there is no `Inadmissible` condition type.

keps/7029-clusterqueue-default-execution-time/README.md
```diff
@@ -0,0 +1,456 @@
...
+    // does not already specify them.
+    // +optional
+    WorkloadDefaults *ClusterQueueWorkloadDefaults `json:"workloadDefaults,omitempty"`
+}
+
+// ClusterQueueWorkloadDefaults defines default values that are applied to
+// workloads submitted to LocalQueues backed by this ClusterQueue when the
+// workload does not already specify them.
+type ClusterQueueWorkloadDefaults struct {
+    // maximumExecutionTimeSeconds if provided, determines the default maximum
```
ArangoGutierrez, line 202: The comment documents only the defaulting behavior, but the same field is also a hard admission ceiling that can leave a workload permanently pending - that belongs in the godoc, since it ships in the CRD schema and `kubectl explain`. Also worth noting the cited Slurm model separates `DefaultTime` and `MaxTime`; a single field forces default == max, so a two-field split deserves an entry in Alternatives.

keps/7029-clusterqueue-default-execution-time/README.md
```diff
@@ -0,0 +1,456 @@
...
+    controller-runtime client cache. The performance impact is negligible.
+
+## Design Details
+
+### API
+
+#### ClusterQueue Spec
+
+Add a new optional `workloadDefaults` field to `ClusterQueueSpec` in both
+`v1beta2` and `v1beta1`. This struct groups all default values that a
```
ArangoGutierrez: Recent ClusterQueueSpec additions have been v1beta2-only since the v1beta1 deprecation (concurrent admission, MultiKueue manager quota automation), so adding this field to v1beta1 would deviate from that convention. Suggest scoping the API change to v1beta2 unless there's a specific backport reason - the companion implementation PR #10479 would need the matching update.

keps/7029-clusterqueue-default-execution-time/README.md
```diff
@@ -0,0 +1,456 @@
...
+
+Add "An existing workload is not affected when the ClusterQueue's timeout is modified"
+
+##### webhook/core/clusterqueue
+
+Add validation tests for the new `maximumExecutionTimeSeconds` field (must be >= 1).
+
+### Graduation Criteria
+
+This feature is gated behind the `ClusterQueueMaxExecutionTime` feature gate.
```
ArangoGutierrez, line 409: Recent KEPs (9270, 9988, 10765, 10852) spell out per-stage Alpha/Beta/GA criteria - feature gate default state, test coverage, a feedback checkpoint before beta. Understood this KEP is still provisional, but this section needs that structure before moving to implementable.

keps/7029-clusterqueue-default-execution-time/kep.yaml
```diff
@@ -0,0 +1,26 @@
+title: ClusterQueue Maximum Execution Time
+kep-number: 7029
+authors:
+  - "@kannon92"
+status: provisional
+creation-date: 2026-04-09
+reviewers:
+  - TBD
```
ArangoGutierrez: Every merged kep.yaml names real reviewers and approvers even at provisional stage - `TBD` is the template placeholder. KEP-3125's reviewers would be natural candidates given this builds directly on it.

ArangoGutierrez (commented):
Found 7 issues. The two load-bearing ones: the advertised recovery path for over-limit workloads breaks in the remove-the-label case, and the admission-enforcement design assumes scheduler access to `workloadDefaults` that doesn't exist in the cache snapshot today. The rest are API-doc/convention items. Overall the ClusterQueue-level default + ceiling direction looks sound and well-motivated - these are asks to make the KEP implementable, not objections to the approach.

This review was prepared with AI assistance, disclosed per the Kubernetes AI tool usage policy.

1. This says a stuck workload recovers when the user corrects the job by reducing or removing the label, but the equivalency fix described below only compares `maximumExecutionTimeSeconds` when the label is present - and after removal it is absent. So removing the label never triggers recreation and the workload stays pending with the stale over-limit value; worth specifying how label removal forces re-evaluation. (keps/7029-clusterqueue-default-execution-time/README.md:261)
2. The scheduler's view of a ClusterQueue is the cache snapshot, which doesn't carry `workloadDefaults` today, so this assumption doesn't hold yet. A short cache/snapshot plumbing note, plus which requeue class an over-limit workload lands in (inadmissible rather than backoff retry, to avoid a hot requeue loop), would make this section implementable. (keps/7029-clusterqueue-default-execution-time/README.md:259)
3. `Inadmissible` exists today as a Reason on the `QuotaReserved=False` condition, documented as LocalQueue/ClusterQueue missing or inactive - a different failure mode than exceeding a time limit. Suggest a new granular `QuotaReserved=False` reason (e.g. `MaximumExecutionTimeExceeded`) following the KEP-10852 pattern, and rewording this bullet since there is no `Inadmissible` condition type. (keps/7029-clusterqueue-default-execution-time/README.md:251)
4. The comment documents only the defaulting behavior, but the same field is also a hard admission ceiling that can leave a workload permanently pending - that belongs in the godoc, since it ships in the CRD schema and `kubectl explain`. Also worth noting the cited Slurm model separates `DefaultTime` and `MaxTime`; a single field forces default == max, so a two-field split deserves an entry in Alternatives. (keps/7029-clusterqueue-default-execution-time/README.md:201)
5. Recent ClusterQueueSpec additions have been v1beta2-only since the v1beta1 deprecation (concurrent admission, MultiKueue manager quota automation), so adding this field to v1beta1 would deviate from that convention. Suggest scoping the API change to v1beta2 unless there's a specific backport reason - the companion implementation PR #10479 would need the matching update. (keps/7029-clusterqueue-default-execution-time/README.md:180)
6. Recent KEPs (9270, 9988, 10765, 10852) spell out per-stage Alpha/Beta/GA criteria - feature gate default state, test coverage, a feedback checkpoint before beta. Understood this KEP is still provisional, but this section needs that structure before moving to implementable. (keps/7029-clusterqueue-default-execution-time/README.md:378)
7. Every merged kep.yaml names real reviewers and approvers even at provisional stage - `TBD` is the template placeholder. KEP-3125's reviewers would be natural candidates given this builds directly on it. (keps/7029-clusterqueue-default-execution-time/kep.yaml:8)

keps/7029-clusterqueue-default-execution-time/README.md
kannon92: Interesting. What do you think we should do here?

https://slurm.schedmd.com/sched_config.html#backfill

keps/7029-clusterqueue-default-execution-time/README.md
```diff
@@ -0,0 +1,504 @@
...
+
+5. **Default only (no upper bound)**: The ClusterQueue value could serve purely
+   as a default, allowing jobs to specify any value (including longer timeouts)
+   without restriction. This was rejected after reviewing how Slurm handles
+   partition-level time limits: Slurm enforces the partition limit as an upper
+   bound that cannot be exceeded, even by jobs with QOS or association limits.
+   This prevents runaway resource consumption and is the expected behavior for
+   administrators setting queue-level time limits.
+
+6. **Silently cap (reduce) the job's timeout**: Instead of rejecting workloads
```
kaushik229, line 480: nit: "rejecting" here reads as slightly inconsistent with the Non-Goals, which describe the workload as "kept pending rather than … modified." Should this say "keeping the workload pending"?

keps/7029-clusterqueue-default-execution-time/README.md
```diff
@@ -0,0 +1,504 @@
...
+
+1. The workload is not requeued for immediate retry. It is treated the same way
+   as other structurally-inadmissible workloads (e.g., a missing LocalQueue or
+   ClusterQueue): it is moved to the ClusterQueue's inadmissible set
+   (`queue.RequeueReasonGeneric` is not used) so it is not re-evaluated on every
+   scheduling cycle, avoiding a hot retry loop. It is requeued for
+   re-evaluation only on relevant events — e.g., the ClusterQueue's
+   `workloadDefaults` changing, or the workload itself being updated/recreated.
+2. The workload's `QuotaReserved` condition is set to `False` with a new,
+   dedicated reason — e.g., `MaximumExecutionTimeExceeded` — following the same
```
kaushik229, line 262: The proposed QuotaReserved=False reason MaximumExecutionTimeExceeded is already an existing constant used for a different event: [WorkloadMaximumExecutionTimeExceeded = "MaximumExecutionTimeExceeded"](https://github.com/kubernetes-sigs/kueue/blob/main/apis/kueue/v1beta2/workload_types.go#L1144), emitted when a workload is evicted after actually running past [its limit](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/core/workload_controller.go#L866-L872).

Reusing the same string means operators can't distinguish "rejected up front for requesting too long" from  "evicted after running too long" — the same concern the KEP raises for avoiding Inadmissible. Suggest a distinct reason such as MaximumExecutionTimeLimitExceeded, and calling out the existing constant so the collision is a conscious choice.

keps/7029-clusterqueue-default-execution-time/README.md
```diff
@@ -0,0 +1,504 @@
...
+A new helper method on `JobReconciler` performs a GET on the LocalQueue to
+resolve the ClusterQueue name, then a GET on the ClusterQueue to retrieve its
+`WorkloadDefaults`. The `MaximumExecutionTimeSeconds` is copied into the
+workload spec if set.
+
+If either lookup fails (e.g., the queue or ClusterQueue does not exist yet), the
+error is logged but workload creation proceeds without the default. This is
+consistent with existing behavior where a job can reference a non-existent queue.
+
+#### Admission Enforcement
```
kaushik229, line 240: This section doesn't mention how the ceiling interacts with admission checks, but the timing is worth stating explicitly — especially for ProvisioningRequest users:
- The max-exec-time clock starts at WorkloadAdmitted=True, which is after all admission checks resolve. So time a workload spends waiting on a ProvisioningRequest to provision nodes is not counted against its budget. Stating this avoids the impression that a long-pending workload is burning its timeout when it isn't.
- The upper-bound rejection sets QuotaReserved=False, which fires before quota is reserved — and therefore before any ProvisioningRequest is created — so over-limit workloads never trigger provisioning. Worth calling out as a positive property.
A short "Notes/Constraints" paragraph making this timing explicit would preempt confusion

keps/7029-clusterqueue-default-execution-time/README.md
```diff
@@ -0,0 +1,504 @@
...
+
+This check cannot be enforced at workload creation time because the ClusterQueue
+assignment may not yet be resolved. The scheduler is the appropriate place
+because it already has the ClusterQueue context and performs admission checks.
+
+If the user corrects the job by lowering the label's value to within the
+ClusterQueue's limit, the workload is recreated through the equivalency check
+mechanism described below, and the new workload can be admitted.
+
+Removing the label entirely does **not** recover the workload: because the
```
kaushik229, line 285: Nice that this edge case is documented. It's a genuine UX trap, though: removing the label leaves the over-limit workload stuck, with "delete the workload/job" as the only recovery.
@kannon92 — is "delete to recover" the intended Alpha behavior? If so, it'd be worth surfacing the required action directly in the condition message so users aren't left guessing.

kaushik229 (commented):
Nice KEP! A few small inline notes — the main one is a possible reason-name overlap worth a look (MaximumExecutionTimeExceeded looks like it's already used as an eviction reason), plus a couple of clarity suggestions and a wording nit.

keps/7029-clusterqueue-default-execution-time/kep.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+see-also:
+  - "/keps/3125-maximum-execution-time"
+
+stage: alpha
+
+latest-milestone: "v0.19"
+
+milestone:
+  alpha: "v0.19"
```
mimowo: it is a little late in the cycle for 0.19, especially as it includes API changes, we plan the release for next Wed.  Let us know if 0.20 is ok for you.

kannon92: yea thats fine.

keps/7029-clusterqueue-default-execution-time/kep.yaml
```diff
@@ -0,0 +1,27 @@
...
+  - "@tenzen-y"
+approvers:
+  - "@mimowo"
+
+see-also:
+  - "/keps/3125-maximum-execution-time"
+
+stage: alpha
+
+latest-milestone: "v0.19"
```
kannon92:
```suggestion
latest-milestone: "v0.20"
```

keps/7029-clusterqueue-default-execution-time/kep.yaml
kannon92:
```suggestion
  alpha: "v0.20"
```

kannon92:
/cc @mimowo @tenzen-y @mwielgus 

This has been opened since April 17th without much reviews. We are getting use cases for this now so I'd like to see how we can proceed this feature.

PR #10623, opened 2026-04-20
Improve eviction message for AdmissionChecks in Retry state to includ…

### LINKED ISSUE #10618: Improve the message for AdmissionChecks in Retry state

I would like to make the eviction message and event more informative exposing the underlying cause for setting Retry by the external controller.

Currently we only say "At least one admission check is false" [here](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/core/workload_controller.go#L748) which does not allow to surface the underlying cause for the Retry.

I would like to have a message like:
"There are 2 admission check(s) in Retry state; "ac1" infrastructure not prepared; "ac2" budget exhausted".

Or something along the lines. 

This is related to https://github.com/kubernetes-sigs/kueue/issues/10620, but only focuses on messages so that we can easily cherrypick. I think the extension of the metric will be tricky to CP as it requires (IIUC) a new API field.

**Why is this needed**:

In some cases the external controller may set "Retry" due to different reasons. We would like to get visibility into the underlying reasons.

reruno:
…e per-check name and reason

#### What type of PR is this?
/kind feature

Improve eviction message for AdmissionChecks in Retry state to include per-check name and reason

Fixes #10618 

```release-note
Improve eviction message for AdmissionChecks in Retry state to include per-check name and reason
```

### DIFF
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -720,6 +720,27 @@ func (r *WorkloadReconciler) reconcileMaxExecutionTime(ctx context.Context, wl *
 	return 0, nil
 }
 
+// buildAdmissionChecksMessage formats a human-readable message
+// describing the list of admission checks in the given state.
+func buildAdmissionChecksMessage(checks []kueue.AdmissionCheckState, state kueue.CheckState) string {
+	slices.SortFunc(checks, func(a, b kueue.AdmissionCheckState) int {
+		return cmp.Compare(a.Name, b.Name)
+	})
+	parts := make([]string, 0, len(checks))
+	for _, ac := range checks {
+		if ac.Message != "" {
+			parts = append(parts, fmt.Sprintf("%q (%s)", ac.Name, ac.Message))
+		} else {
+			parts = append(parts, fmt.Sprintf("%q", ac.Name))
+		}
+	}
+	noun := "AdmissionCheck"
+	if len(checks) > 1 {
+		noun = "AdmissionChecks"
+	}
+	return fmt.Sprintf("%s in %s state: %s", noun, state, stringsutils.Join(parts, "; "))
+}
+
 // reconcileCheckBasedEviction evicts or deactivates the given Workload if any admission checks have failed.
 // Returns true if the Workload was rejected or deactivated, and false otherwise.
 func (r *WorkloadReconciler) reconcileCheckBasedEviction(ctx context.Context, wl *kueue.Workload) (bool, error) {
@@ -729,23 +750,21 @@ func (r *WorkloadReconciler) reconcileCheckBasedEviction(ctx context.Context, wl
 	log := ctrl.LoggerFrom(ctx)
 	log.V(3).Info("Workload is evicted due to admission checks")
 	if workload.HasRejectedChecks(wl) {
-		var rejectedCheckNames []kueue.AdmissionCheckReference
-		for _, check := range workload.RejectedChecks(wl) {
-			rejectedCheckNames = append(rejectedCheckNames, check.Name)
-		}
+		rejectedChecks := workload.RejectedChecks(wl)
+		message := buildAdmissionChecksMessage(rejectedChecks, kueue.CheckStateRejected)
 		err := workload.PatchAdmissionStatus(ctx, r.client, wl, r.clock, func(wl *kueue.Workload) (bool, error) {
-			return workload.SetDeactivationTarget(wl, kueue.WorkloadEvictedByAdmissionCheck, fmt.Sprintf("Admission check(s): %v, were rejected", stringsutils.Join(rejectedCheckNames, ","))), nil
+			return workload.SetDeactivationTarget(wl, kueue.WorkloadEvictedByAdmissionCheck, message), nil
 		})
 		if err != nil {
 			return false, err
 		}
-		log.V(3).Info("Workload is evicted due to rejected admission checks", "workload", klog.KObj(wl), "rejectedChecks", rejectedCheckNames)
-		rejectedCheck := workload.RejectedChecks(wl)[0]
-		r.recorder.Eventf(wl, corev1.EventTypeWarning, "AdmissionCheckRejected", "Deactivating workload because AdmissionCheck for %v was Rejected: %s", rejectedCheck.Name, rejectedCheck.Message)
+		log.V(3).Info("Workload is deactivated due to rejected admission checks", "workload", klog.KObj(wl), "rejectedChecks", rejectedChecks)
+		r.recorder.Event(wl, corev1.EventTypeWarning, "AdmissionCheckRejected", fmt.Sprintf("Deactivated due to %s", message))
 		return true, nil
 	}
 	// at this point we know a Workload has at least one Retry AdmissionCheck
-	message := "At least one admission check is false"
+	retryChecks := workload.RetryChecks(wl)
+	message := fmt.Sprintf("Evicted due to %s", buildAdmissionChecksMessage(retryChecks, kueue.CheckStateRetry))
 	exposeLqMetrics := r.cache.ShouldExposeLocalQueueMetricsForWorkload(log, wl)
 	if err := workload.Evict(ctx, r.client, r.recorder, wl, kueue.WorkloadEvictedByAdmissionCheck, message, "", r.clock, exposeLqMetrics, r.roleTracker, r.customLabels); err != nil {
 		return false, err
--- a/pkg/controller/core/workload_controller_test.go
+++ b/pkg/controller/core/workload_controller_test.go
@@ -800,7 +800,7 @@ func TestReconcile(t *testing.T) {
 						Type:    kueue.WorkloadDeactivationTarget,
 						Status:  metav1.ConditionTrue,
 						Reason:  "AdmissionCheck",
-						Message: "Admission check(s): check, were rejected",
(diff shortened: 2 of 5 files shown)


pkg/workload/admissionchecks.go
```diff
@@ -151,6 +151,18 @@ func RejectedChecks(wl *kueue.Workload) []kueue.AdmissionCheckState {
...
+// RetryChecks returns the list of Retry admission checks
+func RetryChecks(wl *kueue.Workload) []kueue.AdmissionCheckState {
+	retryChecks := make([]kueue.AdmissionCheckState, 0, len(wl.Status.AdmissionChecks))
+	for i := range wl.Status.AdmissionChecks {
+		ac := wl.Status.AdmissionChecks[i]
+		if ac.State == kueue.CheckStateRetry {
+			retryChecks = append(retryChecks, ac)
+		}
+	}
+	return retryChecks
```
kshalot: Nit: Since this is pretty much identical to the existing `RejectedChecks`, we could de-duplicate with a helper: 
```go
func matchingChecks(wl *kueue.Workload, s CheckState) []kueue.AdmissionCheckState
```

pkg/controller/core/workload_controller.go
```diff
@@ -745,7 +745,19 @@ func (r *WorkloadReconciler) reconcileCheckBasedEviction(ctx context.Context, wl
...
+	})
+	parts := make([]string, 0, len(retryChecks))
+	for _, ac := range retryChecks {
+		if ac.Message != "" {
+			parts = append(parts, fmt.Sprintf("%q %s", ac.Name, ac.Message))
+		} else {
+			parts = append(parts, fmt.Sprintf("%q", ac.Name))
+		}
+	}
+	message := fmt.Sprintf("There are %d admission check(s) in Retry state: %s", len(retryChecks), stringsutils.Join(parts, "; "))
```
kshalot: Nit: I realize this format is taken from the issue description, but I'm wondering whether sth like:
```
Retrying %d admission check(s): %q (%s), %q (%s), %q (%s)
```
wouldn't sound more natural and be a little easier to parse with your eyes. But I don't think we are using some strict convention for messages like these. I took a quick look at the [api-conventions](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md), but I could only find some for error messages.

pkg/controller/core/workload_controller.go
```diff
@@ -745,7 +745,19 @@ func (r *WorkloadReconciler) reconcileCheckBasedEviction(ctx context.Context, wl
...
+	})
+	parts := make([]string, 0, len(retryChecks))
+	for _, ac := range retryChecks {
+		if ac.Message != "" {
+			parts = append(parts, fmt.Sprintf("%q (%s)", ac.Name, ac.Message))
+		} else {
+			parts = append(parts, fmt.Sprintf("%q", ac.Name))
+		}
+	}
+	message := fmt.Sprintf("There are %d admission check(s) in Retry state: %s", len(retryChecks), stringsutils.Join(parts, ", "))
```
mimowo: Wrap it into a helper like buildRetryMessage, 

Also, I would like commonize the message building for Rejected (above), so that there we also have more details.

Maybe something like "Deactivating workload, because there are %d admission check(s) in Rejected state: %s".

pkg/controller/core/workload_controller.go
```diff
@@ -729,23 +746,21 @@ func (r *WorkloadReconciler) reconcileCheckBasedEviction(ctx context.Context, wl
...
+			return workload.SetDeactivationTarget(wl, kueue.WorkloadEvictedByAdmissionCheck, message), nil
 		})
 		if err != nil {
 			return false, err
 		}
-		log.V(3).Info("Workload is evicted due to rejected admission checks", "workload", klog.KObj(wl), "rejectedChecks", rejectedCheckNames)
-		rejectedCheck := workload.RejectedChecks(wl)[0]
-		r.recorder.Eventf(wl, corev1.EventTypeWarning, "AdmissionCheckRejected", "Deactivating workload because AdmissionCheck for %v was Rejected: %s", rejectedCheck.Name, rejectedCheck.Message)
+		log.V(3).Info("Workload is evicted due to rejected admission checks", "workload", klog.KObj(wl), "rejectedChecks", rejectedChecks)
+		r.recorder.Eventf(wl, corev1.EventTypeWarning, "AdmissionCheckRejected", "Deactivating workload, because there are %s", message)
```
mimowo: Let's reuse the same detailed message for the Event and the condition, for consistency, similar to what workload.Evict(...) does. 

I would also prefer to:
- drop the explicit count, as in 99% there will be only one, and they are all listed anyway
- differentiate based on singular / plural so that it reads more naturally, in 99% there will be only one

*Rejected*: 
singular: Deactivated due to AdmissionCheck in Rejected state: (...)
plural: Deactivated due to AdmissionChecks in Rejected state: (...)
* Retry*: 
singular: Evicted due to AdmissionCheck in Retry state: (...)
plural: Evicted due to AdmissionChecks in Retry state: (...)

pkg/controller/core/workload_controller_test.go
```diff
@@ -819,17 +819,19 @@ func TestReconcile(t *testing.T) {
 				AdmittedAt(true, now).
 				ControllerReference(batchv1.SchemeGroupVersion.WithKind("Job"), "ownername", "owneruid").
 				AdmissionCheck(kueue.AdmissionCheckState{
-					Name:  "check",
-					State: kueue.CheckStateRejected,
+					Name:    "check",
+					State:   kueue.CheckStateRejected,
+					Message: "quota exceeded",
```
mimowo, line 872: Let's add a new test case "with message" rather than modidying the existing one.

pkg/controller/core/workload_controller_test.go
mimowo, line 1021: Let's add a new test case rather than modifying the existing one

pkg/controller/core/workload_controller.go
```diff
@@ -465,7 +466,14 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
 				underlyingCause = kueue.EvictionUnderlyingCause(dtCond.Reason)
-				message = fmt.Sprintf("%s due to %s", message, dtCond.Message)
+				if strings.HasPrefix(dtCond.Message, "Deactivated due to ") {
+					// dtCond.Message is already a complete sentence (e.g. produced by
+					// buildAdmissionChecksMessage); avoid duplicating the "deactivated due to"
+					// fragment by prepending "The workload is " and lowercasing the first letter.
+					message = "The workload is " + strings.ToLower(dtCond.Message[:1]) + dtCond.Message[1:]
+				} else {
+					message = fmt.Sprintf("%s due to %s", message, dtCond.Message)
+				}
```
mimowo: I'm not sure why this is changed. I thought the message would be taken from the DeactivationTarget which has the complete message. 

Can you provide what would be the message without / with the change in the discussion thread.

mimowo: I basically don't like the string manupulations like that `"The workload is " + strings.ToLower(dtCond.Message[:1]) + dtCond.Message[1:]`. I think we need to rethink how to approach it. I gave some proposals for the messages, but if they require this kind of manipulations I'm ok to adjust them, but I need to understand why is happening.

reruno: Without this change message in condition will look like this 
`The workload is deactivated due to Deactivated due to AdmissionCheck in Rejected state: "check"`

I think it is best to revert this change completely and make different message in condition and event in `buildAdmissionChecksMessage`, I am working on it.

mimowo: Maybe the message for conditions could be just `AdmissionCheck in Rejected state: "check"`, and `AdmissionCheck in Retry state: "check"`, because `Evicted` / `Deactivated` is anyway redundant given the condition type.

mimowo:
LGTM overall, let me also ask @pajakd for another pass, especially as he is working also on observability improvements in the area.

pkg/controller/core/workload_controller.go
```diff
@@ -729,23 +750,21 @@ func (r *WorkloadReconciler) reconcileCheckBasedEviction(ctx context.Context, wl
...
-			return workload.SetDeactivationTarget(wl, kueue.WorkloadEvictedByAdmissionCheck, fmt.Sprintf("Admission check(s): %v, were rejected", stringsutils.Join(rejectedCheckNames, ","))), nil
+			return workload.SetDeactivationTarget(wl, kueue.WorkloadEvictedByAdmissionCheck, message), nil
 		})
 		if err != nil {
 			return false, err
 		}
-		log.V(3).Info("Workload is evicted due to rejected admission checks", "workload", klog.KObj(wl), "rejectedChecks", rejectedCheckNames)
-		rejectedCheck := workload.RejectedChecks(wl)[0]
-		r.recorder.Eventf(wl, corev1.EventTypeWarning, "AdmissionCheckRejected", "Deactivating workload because AdmissionCheck for %v was Rejected: %s", rejectedCheck.Name, rejectedCheck.Message)
+		log.V(3).Info("Workload is evicted due to rejected admission checks", "workload", klog.KObj(wl), "rejectedChecks", rejectedChecks)
```
pajakd: The log should probably say "Workload is deactivated", right?

mimowo (commented):
Thanks 👍 
/approve
/cherrypick release-0.16
/cherrypick release-0.17

PR #10633, opened 2026-04-20
Add metrics for time to preemption

### LINKED ISSUE #10124: Add metrics for age of pending workloads and time to preemption

vladikkuzn:

Part of #10124

```release-note
Observability: introduce a guage metric, called pending_workload_wait_time_seconds, which allows to track
the current waiting time for the pending workloads. This metric is currently available behind the 
PendingWorkloadsMetrics feature gate (alpha, disabled by default).
```


pkg/cache/queue/pending_wait_metrics.go
```diff
@@ -0,0 +1,219 @@
...
+	"sigs.k8s.io/kueue/pkg/metrics"
+	"sigs.k8s.io/kueue/pkg/workload"
+)
+
+const (
+	pendingWaitMetricsBatchPeriod = time.Second
+	pendingWaitMetricsTickPeriod  = 15 * time.Second
+)
+
+type queuedWaitAgg struct {
```
mwielgus: Can you add some commends here and below?

pkg/cache/queue/pending_wait_metrics.go
```diff
@@ -0,0 +1,219 @@
...
+		item, shutdown := w.queue.Get()
+		if shutdown {
+			return nil
+		}
+		w.reconcile(ctx, item)
+		w.queue.Done(item)
+	}
+}
+
+func (w *PendingWaitMetricsWorker) enqueueAllClusterQueues() {
```
mwielgus, line 208: Can you introduce some minor jitter to evaluation, so not all CQ are processed at the very same time?

pkg/cache/queue/pending_wait_metrics.go
```diff
@@ -0,0 +1,219 @@
...
+		return
+	}
+	cq := m.getClusterQueue(cqName)
+	if cq == nil {
+		metrics.ClearPendingWorkloadWaitTimes(cqName)
+		return
+	}
+	cqSchedulingActive := m.statusChecker == nil || m.statusChecker.ClusterQueueActive(cqName)
+	activeMax, activeMean, activeN, inadmMax, inadmMean, inadmN := cq.pendingWaitAggregates(m.clock, cqSchedulingActive)
+	metrics.ReportPendingWorkloadWaitTimes(cqName, activeMax, activeMean, activeN, inadmMax, inadmMean, inadmN, m.customLabels.CQGet(cq.name), m.roleTracker)
```
mwielgus: Is cq.name the same as cqName?

pkg/metrics/metrics.go
```diff
@@ -352,6 +360,30 @@ The label 'result' can have the following values:
...
+		prometheus.GaugeOpts{
+			Subsystem: constants.KueueName,
+			Name:      "pending_workload_max_wait_time_seconds",
+			Help: `The maximum queued wait time in seconds among pending workloads in this ClusterQueue bucket,
+per 'cluster_queue' and 'status'. Queued wait time is from workload creation or last requeue until now.
+'status' matches kueue_pending_workloads: "active" for workloads in the admission heap (plus inflight scheduling),
+"inadmissible" for workloads waiting for cluster conditions after a failed admission attempt.
+When the ClusterQueue is inactive for scheduling, active workloads are folded into the inadmissible bucket.
+No time series is reported for a bucket with zero pending workloads.`,
+		}, append([]string{"cluster_queue", "status", "replica_role"}, extraLabels...),
```
mimowo: I think for this metric it would be very helpful to include the longest pending workload, so that the admin can react. Maybe a dedicated label like "workload", which would be namespace/name format.

pkg/metrics/metrics.go
```diff
@@ -352,6 +360,30 @@ The label 'result' can have the following values:
...
+
+	PendingWorkloadMeanWaitTimeSeconds = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: constants.KueueName,
+			Name:      "pending_workload_mean_wait_time_seconds",
+			Help: `The mean queued wait time in seconds among pending workloads in this ClusterQueue bucket,
+per 'cluster_queue' and 'status'. Semantics match kueue_pending_workload_max_wait_time_seconds.`,
+		}, append([]string{"cluster_queue", "status", "replica_role"}, extraLabels...),
+	)
+	trackGaugeVec(PendingWorkloadMeanWaitTimeSeconds, gaugeCleanupScopeClusterQueue)
```
mimowo: For this metric I would rather have `PendingWorkloadWaitTimeSeconds{quantile=0.5|0.95|0.99}`, I suppose this will not complicate the code that much.

pkg/metrics/metrics.go
```diff
@@ -352,6 +360,30 @@ The label 'result' can have the following values:
 	)
 	trackGaugeVec(PendingWorkloads, gaugeCleanupScopeClusterQueue)
 
+	PendingWorkloadMaxWaitTimeSeconds = prometheus.NewGaugeVec(
```
mimowo: Let's keep both metrics under a feature gate, like PendingWorkloadsMetrics, because they introduce new mechanism which may have non trivial consequences. I think we could start with Alpha in 0.18

pkg/metrics/metrics.go
vladikkuzn: > I think for this metric it would be very helpful to include the longest pending workload, so that the admin can react.

Following the same pattern as [kube_pod_start_time](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/metrics/workload/pod-metrics.md), age is computed in PromQL (time() - metric), I think we can add:
```
# HELP kueue_pending_workload_oldest_start_time_seconds Unix timestamp of the oldest pending workload in a ClusterQueue, by status.
# TYPE kueue_pending_workload_oldest_start_time_seconds gauge
kueue_pending_workload_oldest_start_time_seconds{cluster_queue, status, replica_role}
```
And the same for LQ:
```
kueue_local_queue_pending_workload_oldest_start_time_seconds{name, namespace, status, replica_role}
```
This could be potential alerts based on those:
```
- alert: KueueWorkloadPendingTooLong
  expr: time() - kueue_pending_workload_oldest_start_time_seconds > 1800
  for: 5m
  labels:
    severity: warning
  annotations:
    summary: "Workload pending >30m in ClusterQueue {{ $labels.cluster_queue }} (status={{ $labels.status }})"

- alert: KueueWorkloadInadmissibleTooLong
  expr: time() - kueue_pending_workload_oldest_start_time_seconds{status="inadmissible"} > 3600
  for: 10m
  ```

> Maybe a dedicated label like "workload", which would be namespace/name format.

We need to avoid workload and priority_class labels and as it blow-ups cardinality, or make it opt-in

Comment (vladikkuzn) [reply pkg/metrics/metrics.go]: I can also propose to add traces per-workload via OpenTelemetry as a longer term goal. This would enable true per-workload analytics for power users

Comment (vladikkuzn) [reply pkg/metrics/metrics.go]: @mimowo quick demo is available here: https://github.com/kubernetes-sigs/kueue/pull/10818

**Comment by mbobrovskyi** [comment]
@vladikkuzn could you please rebase?

File: pkg/metrics/metrics.go (hunk 4 above)
Comment (mimowo) [reply pkg/metrics/metrics.go]: This is really cool, but let's discuss that on the PR. Let's discuss the OpenTelemetry stuff under the PR/issue.

As for the metric for the longest waiting workload, if we could include the workload name / namespace in the guage metric it would be very useful for admins to setup alerting on, and look at such workloads. We can discuss details in the follow up PR.

**Comment by mimowo** [comment]
/release-note-edit
```release-note
Observability: introduce a guage metric, called pending_workload_wait_time_seconds, which allows to track
the current waiting time for the pending workloads. This metric is currently available behind the 
PendingWorkloadsMetrics feature gate (alpha, disabled by default).
```

pkg/cache/queue/pending_wait_metrics.go
```diff
@@ -0,0 +1,233 @@
...
+
+// quantileFromSorted returns the q-th quantile from a pre-sorted slice
+// using the nearest-rank (ceiling) method.
+func quantileFromSorted(sorted []float64, q float64) float64 {
+	n := len(sorted)
+	if n == 0 {
+		return 0
+	}
+	// Nearest-rank: rank = ceil(q * n) - 1 (0-indexed).
+	rank := max(int(q*float64(n)+0.5)-1, 0)
```
mimowo, line 76: The comment does not match the  implementation. The comment says ceiling while the impl is using rounding as I see int(  x +0.5). I think this looks correct `idx := int(math.Ceil(q*float64(len(sorted)))) - 1`

pkg/metrics/metrics.go
```diff
@@ -352,6 +356,21 @@ The label 'result' can have the following values:
...
+			Subsystem: constants.KueueName,
+			Name:      "pending_workload_wait_time_seconds",
+			Help: `Queued wait time quantiles in seconds among pending workloads in this ClusterQueue bucket,
+per 'cluster_queue', 'status', and 'quantile'. Queued wait time is from workload creation or last requeue until now.
+'status' matches kueue_pending_workloads: "active" for workloads in the admission heap (plus inflight scheduling),
+"inadmissible" for workloads waiting for cluster conditions after a failed admission attempt.
+When the ClusterQueue is inactive for scheduling, active workloads are folded into the inadmissible bucket.
+No time series is reported for a bucket with zero pending workloads.
+Requires feature gate PendingWorkloadsMetrics (alpha in 0.18).`,
+		}, append([]string{"cluster_queue", "status", "quantile", "replica_role"}, extraLabels...),
```
mimowo, line 370: I would suggest the status to a follow up PR, because:

For the foundation PR, I think the metric should answer the core request from the issue: “what is the pending wait time distribution for this ClusterQueue?” Splitting active vs inadmissible may be useful, but it is not requested in the issue and it complicates both the implementation and the metric surface.

Since this is alpha, we can still evolve it. If we add breakdowns later, I would prefer to design that deliberately around multiple possible dimensions, for example `status` and `priority_class`, rather than hard-coding the implementation around only the active/inadmissible split.

So my preference would be: one aggregated vector for now, and discuss status/priority breakdowns as a follow-up.

mimowo:
cc @amy would you like to give it a pass?

pkg/cache/queue/pending_wait_metrics.go
```diff
@@ -0,0 +1,233 @@
...
+		return
+	}
+	maxJitter := int64(w.tickPeriod / 2)
+	for _, name := range w.manager.GetClusterQueueNames() {
+		jitter := time.Duration(rand.Int64N(maxJitter))
+		w.queue.AddAfter(name, w.batchPeriod+jitter)
+	}
+}
+
+func (w *PendingWaitMetricsWorker) reconcile(ctx context.Context, cqName kueue.ClusterQueueReference) {
```
mimowo, line 219: Seems like ctx is unused, let's use it for logging at V4

pkg/cache/queue/manager.go
```diff
@@ -200,10 +209,20 @@ func NewManager(client client.Client, checker StatusChecker, requeuer inadmissib
 		option(m)
 	}
 	m.requeuer.setManager(m)
+	if m.pendingWaitMetrics != nil {
+		m.pendingWaitMetrics.setManager(m)
+	}
```
mimowo, line 214: hide it behind the FG, just in case

pkg/cache/queue/manager.go
```diff
@@ -200,10 +209,20 @@ func NewManager(client client.Client, checker StatusChecker, requeuer inadmissib
...
 	m.cond.L = &m.RWMutex
 	return m
 }
 
+func (m *Manager) notifyPendingWaitMetrics(cqName kueue.ClusterQueueReference) {
+	if m.pendingWaitMetrics == nil {
```
mimowo, line 220: short cirtuit if the FG is disabled

pkg/cache/queue/pending_wait_metrics_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+	testingclock "k8s.io/utils/clock/testing"
+
+	"sigs.k8s.io/kueue/pkg/metrics"
+	utiltesting "sigs.k8s.io/kueue/pkg/util/testing"
+	utiltestingapi "sigs.k8s.io/kueue/pkg/util/testing/v1beta2"
+	"sigs.k8s.io/kueue/pkg/workload"
+)
+
+func TestPendingWaitQuantiles_ActiveHeap(t *testing.T) {
```
mimowo, line 31: Instead of adding many tests, could you consolidate with fewer, but table-driven?

This will save us code on the setup / asserting which repeats across some of the tests

pkg/cache/queue/pending_wait_metrics.go
```diff
@@ -0,0 +1,233 @@
...
+// Start implements manager.Runnable.
+func (w *PendingWaitMetricsWorker) Start(ctx context.Context) error {
+	log := ctrl.LoggerFrom(ctx).WithName("pending_wait_metrics_worker")
+	ctx = ctrl.LoggerInto(ctx, log)
+	go func() {
+		<-ctx.Done()
+		w.queue.ShutDown()
+	}()
+
+	w.enqueueAllClusterQueues()
```
mimowo, line 180: Why is this added? Asking cause this should also happen inside the goroutine.

site/content/en/docs/reference/metrics.md
```diff
@@ -45,6 +45,7 @@ Use the following metrics to monitor the status of your ClusterQueues:
...
+| `kueue_pending_workload_wait_time_seconds` | Gauge | Queued wait time quantiles in seconds among pending workloads in this ClusterQueue bucket,<br>per 'cluster_queue', 'status', and 'quantile'. Queued wait time is from workload creation or last requeue until now.<br>'status' matches kueue_pending_workloads: "active" for workloads in the admission heap (plus inflight scheduling),<br>"inadmissible" for workloads waiting for cluster conditions after a failed admission attempt.<br>When the ClusterQueue is inactive for scheduling, active workloads are folded into the inadmissible bucket.<br>No time series is reported for a bucket with zero pending workloads.<br>Requires feature gate PendingWorkloadsMetrics (alpha in 0.18). | `cluster_queue`: the name of the ClusterQueue<br> `status`: one of `active` or `inadmissible` (same as kueue_pending_workloads)<br> `quantile`: one of `0.5`, `0.95`, or `0.99`<br> `replica_role`: one of `leader`, `follower`, or `standalone` |
```
amy, line 48: Queued wait time is from workload creation or last requeue until now. -> the "now" word should be changed to be more clear. 

Queued wait time is from workload creation or last requeue until ... something about while the workload is still pending.

When the ClusterQueue is inactive for scheduling, active workloads are folded into the inadmissible bucket. -> 
Users dont need to know what inadmissible is. What does this mean for the metric.

pkg/cache/queue/pending_wait_metrics.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+// notify schedules a single ClusterQueue for metrics re-evaluation after the batch period.
+// Used on event-driven paths (workload add/remove); no jitter is applied here.
+func (w *PendingWaitMetricsWorker) notify(cqName kueue.ClusterQueueReference) {
+	w.queue.AddAfter(cqName, w.batchPeriod)
+}
+
+// Start implements manager.Runnable.
+func (w *PendingWaitMetricsWorker) Start(ctx context.Context) error {
```
amy, line 172: Sorry to make this annoying comment bc I think it would fundamentally change your entire PR... but... 

Do you need pending to update every second? Like I think people use this retroactively. To analyze how long a workload was pending for. So you can just capture the pending time when it changes state to something like admitted.

amy: Clarifying with the issue creator here: https://github.com/kubernetes-sigs/kueue/issues/10124#issuecomment-4354985487

pkg/metrics/metrics.go
```diff
@@ -1229,6 +1293,7 @@ func Register() {
 		admissionAttemptDuration,
 		AdmissionCyclePreemptionSkips,
 		PendingWorkloads,
+		PendingWorkloadWaitTimeSeconds,
```
amy, line 1296: You should copy how localQueue registers if the featuregate is on only. See: https://github.com/epam/kubernetes-kueue/blob/bda4d000ac50942fe845aa3f1710b122b959bc6c/pkg/metrics/metrics.go#L1325

pkg/cache/queue/pending_wait_metrics.go
mimowo: That is a very good point, I was also thinking about this, because making it retroactively is so much easier to implement. 

At the same time it feels that this specific metric maybe be very helpful for alerting ASAP.

In particular, we had incidents where users were calling "why my job hasn't scheduled for 2 or 3h". At that point seeing the increase in the waiting time gives admins a potential for faster alerting, before a user calls. 

Making the metric retrospective may not show any increase for a long time if there is an outage actually ongoing.

wdyt?

pkg/metrics/metrics.go
vladikkuzn: My initial design intended to hide reporting behind FG, the metric is registered but not reported, but I guess I can hide it as well

pkg/cache/queue/pending_wait_metrics.go
amy: @mimowo Alternatively you could alert on "guarantee violation" using: https://github.com/kubernetes-sigs/kueue/pull/10485

Its a different alert responsibility. But... alerting per workload on pending time is maybe the wrong layer to alert on. Like what if all the workloads that are pending are supposed to be pending. That seems like a noisy alert to proactively page on. When actually you sorta just need point in time debugging on whether or not that workload is supposed to be pending or not.

I'd want to know what rough criteria you're filtering out the noise to make this sorta thing useful vs. just "workloads that exceed 2hrs pending". (not opposed & willing to change my mind. Just slightly skeptical about the practically of hooking this up to a page.)

pkg/metrics/metrics.go
amy: Yeah its just nice if it follows existing patterns.

pkg/cache/queue/pending_wait_metrics.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+	return sorted[rank]
+}
+
+// pendingWaitQuantiles computes quantile-based queued wait times for active vs inadmissible pending workloads.
+// When cqSchedulingActive is false (ClusterQueue inactive for scheduling), active workloads are folded into the
+// inadmissible bucket — matching ReportPendingWorkloads counts semantics.
+//
+// Caller must NOT hold ClusterQueue locks; this method acquires RLock internally.
+func (c *ClusterQueue) pendingWaitQuantiles(clk clock.Clock, cqSchedulingActive bool) (active, inadm metrics.PendingWorkloadWaitQuantiles) {
```
amy, line 88: the integration test only covers the happy-path active bucket. Missing coverage for the inadmissible bucket, inadmissible→active transition, and the inactive CQ folding behavior.

pkg/cache/queue/pending_wait_metrics.go
mimowo: Good point I'm not sure either this should page by itself, probably not. 

However I had a time in the past I was paged exactly like this: my job didn't schedule for 3h.

So my first thought: is the user outlier who is just unhappy or in 30min, or I will soonget a wave of many more pages.

I was back then thinking such a metric would tell me on the spot about the scale of the issue.

Still I'm not the creator of the issue, and it is right thing to ask, but seeing the issue I just correlated that with my past experience and thought this metric might be useful if done live, but this interpretation is not necessarily MVP for the issue. 

Because the waiting time is very much correlated with ux.

amy: Waiting for sundar to get back. He said he would think about it over dm: https://github.com/kubernetes-sigs/kueue/issues/10124#issuecomment-4356310720

> So my first thought: is the user outlier who is just unhappy or in 30min, or I will soon get a wave of many more pages. I was back then thinking such a metric would tell me on the spot about the scale of the issue.

yeah... chicken and egg. Also I think that it depends on per ClusterQueue usage patterns. Regardless, I'm not blocking if y'all end up deciding you want to merge.

amy: Update: spoke with sundar offline. He said that he agrees that age of pending workload is maybe the wrong metric.

I'll leave it to him to comment on the original issue with a clarification.

I think guarantee violation is the direction we should go in. (Total Pending <= remaining nominal) https://github.com/kubernetes-sigs/kueue/pull/10485

For more granularity... we could also emit the specific quota for the head of queue workload. Where in grafana you could do a query on if capacity for head of queue workload is <= remaining nominal.

mimowo:
/hold
Thank you for the PR - it is close to being ready, and it shows us what is the complexity added by the metric - it is significant, but not a blocker if we have a really good use case. 

However, I'm still not clear if the use cases can or cannot be covered by the more idiomatic approaches first, posted here: https://github.com/kubernetes-sigs/kueue/issues/10124#issuecomment-4372642534. Let's discuss to determine that first.

amy:
@vladikkuzn just checking do we want to keep this PR open?

PR #10634, opened 2026-04-20
Detect and log missing scheduling gates

### LINKED ISSUE #10167: Kueue sometimes does not inject scheduling gates for elastic jobs
**What happened**:
Environment:
Kueue 0.15.4 is deployed with ElasticJobsViaWorkloadSlices feature gate. Ray Services are submitted to one or more cluster queues, all in the same namespace, and all of them have kueue.x-k8s.io/elastic-job annotation. All of them have the Redis-cleanup finalizer. The Ray Cluster (RC) webhook is enabled.

Problem:
However, some Ray Clusters (RCs), but not all, are missing the pod scheduling gates in head/worker pod templates that Kueue is supposed to inject. They are also missing the tolerations defined in the Resource Flavor  that Kueue picked for the RC. At this point, there are 74 RCs in this state in our environment.

This is a big problem because it causes lots of downstream impact:
  * When the Ray Service is terminated, after the Redis-cleanup job has run, KubeRay will try to remove the finalizer from the RC.
  * This will fail, because Kueue checks to see whether the scheduler gates are present before allowing mutation. This check was introduced in the [commit "Support in-tree RayAutoscaler for Elastic RayCluster objects (#6662)" ](https://github.pie.apple.com/apple-batch/kueue/commit/6d7eac119f7e0bbe80a34e6df572e4598c1b9c9a). It is [still present today](https://github.pie.apple.com/apple-batch/kueue/blob/main-apple/pkg/controller/jobs/raycluster/raycluster_webhook.go#L180-L190).
  * As a result, the Ray Cluster will remain in zombie state -- head/worker pods and redis-cleanup pods are gone, but the Ray Cluster alone remains.
  * But this means the Kueue workload also remains. So, resource quota, including GPU quota, is held by Kueue. This is a big problem.
  * Another problem is that, Kueue may select this Ray Service to be preempted when another job is submitted. But the preemption will fail because the Ray Cluster and the Kueue workload remain forever. The preempting job's workload has a message that "Pending the preemption of 1 workload(s)" and it remains stuck in that state.

We have observed all these problems in production. At this point, 74 Ray clusters are stuck due to this issue. 

We expect that the pod scheduling gates are present in all RCs with the elastic job annotation when Kueue is deployed with the elastic job feature gate.

Deploy lots of Ray Services with Redis-cleanup enabled. This happens frequently in our environment. 

My analysis: there's a potential code path that leads to this issue. 
 * The Ray Cluster webhook does not apply the pod gates if [ApplyDefaultForSuspend() returns error](https://github.pie.apple.com/apple-batch/kueue/blob/main-apple/pkg/controller/jobs/raycluster/raycluster_webhook.go#L93).
 * ApplyDefaultForSuspend() would return an error if [WorkloadShouldBeSuspended() returns error](https://github.pie.apple.com/apple-batch/kueue/blob/main-apple/pkg/controller/jobframework/defaults.go#L40).
 * WorkloadShouldBeSuspended() returns error if [FindAncestorJobManagedByKueue() returns error](https://github.pie.apple.com/apple-batch/kueue/blob/main-apple/pkg/controller/jobframework/defaults.go#L54). The next check in WorkloadShouldBeSuspended for the queue name will always succeed, because we always have the queue name annotation.
 * FindAncestorJobManagedByKueue() can return error for many reasons. However, the cyclic dependency and the long chain of ownership are unlikely reasons. But the [attempt to get the owner Ray Service may fail](https://github.pie.apple.com/apple-batch/kueue/blob/main-apple/pkg/controller/jobframework/reconciler.go#L838). For example, the API server may be overloaded or have rate limits. So, this Get may fail randomly.

In short, a Get call that fails randomly can cause pod gates to not be added to RCs.

Potential solution:  The Ray Cluster webhook should not return early if the suspend call fails. Instead, we could go ahead and apply the gates anyway, and then return a combined error at the bottom of the function. 

**Environment**:
- Kubernetes version (use `kubectl version`): 1.29
- Kueue version (use `git describe --tags --dirty --always`): 0.15.4 (but may apply to main branch too)
 - KubeRay

ns-sundar:
#### What type of PR is this?
/kind feature 
/area integrations

As noted in [Issue 10167](https://github.com/kubernetes-sigs/kueue/issues/10167), sometimes we see Ray Clusters without scheduling gates for elastic jobs. This is a big problem when it happens with a Ray Service with GCS FT enabled, because the Ray Cluster has a finalizer in that situation, and KubeRay cannot remove that finalizer -- because Kueue rejects the finalizer removal if the gates are missing. As a result, Ray Clusters remain in zombie state but that means Kueue workloads also remain. Another consequence is that, if a Ray Cluster needs to be preempted by another job A, and the deletion fails due to missing gates, the preemption awaits. 

This PR logs an error in the generic job reconcile loop when we detect Ray Clusters which are elastic jobs but are missing scheduling gates. It also logs an error in the RC webhook if ApplyDefaultForSuspend() fails, which would cause the gates to not be applied.  

Fixes #10167

```release-note
NONE
```

### DIFF
--- a/pkg/controller/jobframework/reconciler.go
+++ b/pkg/controller/jobframework/reconciler.go
@@ -70,8 +70,9 @@ import (
 )
 
 const (
-	FailedToStartFinishedReason = "FailedToStart"
-	managedOwnersChainLimit     = 10
+	FailedToStartFinishedReason         = "FailedToStart"
+	managedOwnersChainLimit             = 10
+	schedulingGateDetectionRequeueDelay = 2 * time.Second
 )
 
 var (
@@ -430,6 +431,17 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
 	wl, err := r.ensureOneWorkload(ctx, job, object)
 	if err != nil {
 		return ctrl.Result{}, err
+	}
+
+	// 1.1 Catchall detection for missing scheduling gates (for elastic jobs like Ray Clusters)
+	if jobWithGateDetection, implements := job.(JobWithSchedulingGateDetection); implements {
+		gatesMissing, err := jobWithGateDetection.DetectMissingSchedulingGates(ctx, r.client)
+		// Log the error but continue with the rest of the reconcile loop
+		if err != nil {
+			log.Error(err, "Failed to detect if scheduling gates are missing")
+		} else if gatesMissing {
+			log.Error(nil, "Missing scheduling gates detected, but not injecting them")
+		}
 	}
 
 	// Update workload conditions if implemented JobWithCustomWorkloadConditions interface.
--- a/pkg/controller/jobs/raycluster/raycluster_controller.go
+++ b/pkg/controller/jobs/raycluster/raycluster_controller.go
@@ -19,14 +19,17 @@ package raycluster
 import (
 	"context"
 	"fmt"
+	"slices"
 
 	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
 	rayutils "github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/utils/ptr"
+	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
@@ -77,6 +80,7 @@ type RayCluster rayv1.RayCluster
 
 var _ jobframework.GenericJob = (*RayCluster)(nil)
 var _ jobframework.JobWithManagedBy = (*RayCluster)(nil)
+var _ jobframework.JobWithSchedulingGateDetection = (*RayCluster)(nil)
 
 func (j *RayCluster) Object() client.Object {
 	return (*rayv1.RayCluster)(j)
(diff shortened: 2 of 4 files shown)


kannon92 (commented):
/ok-to-test

kannon92:
Given that this is a bug fix I think you need a release note.

I don't have a lot of context here but I'm curious if we should consider a feature gate for this "bug fix". It seems a little risky.

Has there been any reproducer? Or even a mock integration test that can catch this error?

kannon92 (changes requested):
Given if there isn't a reproducer you should add some integration tests for RayCluster jobs.

pkg/controller/jobframework/reconciler.go
```diff
@@ -70,8 +70,9 @@ import (
 )
 
 const (
-	FailedToStartFinishedReason = "FailedToStart"
-	managedOwnersChainLimit     = 10
+	FailedToStartFinishedReason         = "FailedToStart"
+	managedOwnersChainLimit             = 10
+	schedulingGateDetectionRequeueDelay = 2 * time.Second
```
kannon92, line 75: This reads a little like a code smell. 

I think maybe this is something we should consider configurable.

kannon92: Actually this variable doesn't seem to be used. Can you remove it?

pkg/controller/jobframework/reconciler.go
```diff
@@ -432,6 +433,17 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{}, err
 	}
 
+	// 1.1 Catchall detection for missing scheduling gates (for elastic jobs like Ray Clusters)
```
kannon92, line 436:
```suggestion
	// 1.1 Catch all detection for missing scheduling gates (for elastic jobs like Ray Clusters)
```

kannon92: wow. TIL that catch-all or catchall are the correct spellings for this.

sorry, what you have is correct. silly english.

kannon92: please disregard as I can't resolve.
can't resolve. :(

mimowo:
Actually, I would rather consider this PR as a last resort option. Let's try to first go to the bottom of the bug for https://github.com/kubernetes-sigs/kueue/issues/10167. Notably, this is using old version of Kueue (0.15.4), and I think some of the issues may stem from that. 

Also for RayService we recently fixed a bug https://github.com/ray-project/kuberay/issues/4686.

So, I think the next steps should be to try to repro the issue using latest Kueue and the bugfixed KubeRay. I think @yaroslava-serdiuk is looking into that.

mimowo:
/hold
For now, and let's try to understand the issue first, and see if we can repro on the latest versions of the projects. 
cc @yaroslava-serdiuk @hiboyang

pkg/controller/jobframework/reconciler.go
ns-sundar: Agreed, let me remove it.

ns-sundar:
> Given if there isn't a reproducer you should add some integration tests for RayCluster jobs.

This is difficult to reproduce. I have tried by forcing various suspected error paths. Could you clarify what you expect in the integration test?

pkg/controller/jobs/raycluster/raycluster_controller.go
```diff
@@ -207,3 +211,59 @@ func (j *RayCluster) ManagedBy() *string {
...
+		return false, nil
+	}
+
+	// Log the detection of missing gates
+	log.Error(nil, "Detected missing ElasticJobSchedulingGate in Ray Cluster",
+		"rayCluster", client.ObjectKeyFromObject(j.Object()),
+		"missingInHead", missingGatesInHead,
+		"missingInWorkers", missingGatesInWorkers)
+
+	// Return true to indicate gates are missing
```
mwielgus, line 247: What happens now? Can we make some corrective action?

ns-sundar:
Difficult to reproduce.

PR #10636, opened 2026-04-21
fix: downgrade TAS flavor mismatch log from Error to V(3)

### LINKED ISSUE #10611: TAS: flavorassigner logs errors prematurely for any flavors non-matching to the PodSet
**What happened**:

We observe a lot of ERROR level logs due to flavors not matching to PodSet for TAS due to 
- "Flavor "xyz" does not contain the requested level"
- "Flavor "xyz" does not support TopologyAwareScheduling"

Logged from https://github.com/kubernetes-sigs/kueue/blob/main/pkg/scheduler/flavorassigner/flavorassigner.go#L944

Such a non-compatible flavor should not be reported as an error. It might be a useful V3 log for debugging. It only means that the PodSet needs to be matched against another flavor.

For example the unit test "workload requiring TAS skips the non-TAS flavor"

maishivamhoo123:
**What this PR does / why we need it**:
This PR fixes an issue where `flavorassigner` was prematurely logging an `ERROR` whenever a flavor did not match a PodSet for TopologyAwareScheduling (TAS). 

Since a non-matching flavor is a normal occurrence during scheduling (meaning the PodSet simply needs to be matched against a different flavor), this should not trigger an error alert. The log level has been downgraded to `V(3).Info` so it is still available for debugging without polluting the main logs.

**Which issue(s) this PR fixes**:
Fixes #10611

**Special notes for your reviewer**:
Tested locally via `go test -v ./pkg/scheduler/flavorassigner` and confirmed the `ERROR` log no longer appears for non-TAS flavors. Full `make test` suite passes successfully.

```release-note
Observability: downgrade the non-compatible flavor error logs to Info level (v3).
```

### DIFF
--- a/pkg/scheduler/flavorassigner/flavorassigner.go
+++ b/pkg/scheduler/flavorassigner/flavorassigner.go
@@ -941,7 +941,7 @@ func (a *FlavorAssigner) checkFlavorForPodSets(
 		if features.Enabled(features.TopologyAwareScheduling) {
 			ps := &a.wl.Obj.Spec.PodSets[psID]
 			if message := checkPodSetAndFlavorMatchForTAS(a.cq, ps, flavor, rg); message != nil {
-				log.Error(nil, *message)
+				log.V(3).Info(*message)
 				status.appendf("%s", *message)
 				return status
 			}


tenzen-y:
@maishivamhoo123, for the next contribution, could you keep the PR template? Because we have some of automations for release.

tenzen-y:
I refined the PR description manually.

tenzen-y (commented):
Thank you!
/lgtm
/approve

PR #7289, opened 2025-10-16
Support mutating workload priority class.

mbobrovskyi:
#### What type of PR is this?
Support mutating workload priority class.

#### What this PR does / why we need it:
/kind cleanup

#### Which issue(s) this PR fixes:

Fixes #7138

#### Does this PR introduce a user-facing change?

```release-note
Scheduling: support mutating the "kueue.x-k8s.io/workloadpriorityclass" label for Jobs with reserved quota.
```

PR #10643, opened 2026-04-21
doc: update workload-priority-class documentation to clarify that WorkloadPriorityClassese is mutable

atosatto:
#### What type of PR is this?

/kind documentation

Update WorkloadPriorityClass concept page and task guide to clarify that `priorityClassSource` and `priorityClassName` are mutable while the Workload is pending, and only become immutable once the `QuotaReserved` condition is `True`.

The previous wording stated these fields are unconditionally immutable, which does not match the current implementation as of https://github.com/kubernetes-sigs/kueue/pull/7289 

```release-note
NONE
```

### DIFF
--- a/site/content/en/docs/concepts/workload_priority_class.md
+++ b/site/content/en/docs/concepts/workload_priority_class.md
@@ -101,12 +101,17 @@ The priority of workloads is used for:
 - Sorting the workloads in the ClusterQueues.
 - Determining whether a workload can preempt others.
 
-## Workload's priority values are always mutable
+## Mutability of priority fields
 
-The `Workload`'s `Priority` field is always mutable.
-If a `Workload` has been pending for a while, you can consider updating its priority to execute it earlier,
-based on your own policies.
-Workload's `PriorityClassSource` and `PriorityClassName` fields are immutable.
+The `kueue.x-k8s.io/priority-class` label on the Job can be changed while the Job is suspended.
+When updated, Kueue reconciles the Workload's priority fields accordingly.
+
+On the Workload, `Priority` is always mutable.
+`priorityClassRef` (and its `group`/`kind`) is mutable while the Workload is pending,
+and becomes immutable once the `QuotaReserved` condition is `True`.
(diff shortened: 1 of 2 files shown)


site/content/en/docs/concepts/workload_priority_class.md
```diff
@@ -101,12 +101,14 @@ The priority of workloads is used for:
...
-The `Workload`'s `Priority` field is always mutable.
-If a `Workload` has been pending for a while, you can consider updating its priority to execute it earlier,
-based on your own policies.
-Workload's `PriorityClassSource` and `PriorityClassName` fields are immutable.
+The `kueue.x-k8s.io/priority-class` label on the Job can be changed while the Job is suspended.
+When updated, Kueue reconciles the Workload's priority fields accordingly.
+
+On the Workload, `Priority` is always mutable.
+`PriorityClassSource` and `PriorityClassName` can be updated while the Workload is pending,
+but become immutable once the `QuotaReserved` condition is `True`.
```
tenzen-y:
```suggestion
On the Workload, `Priority` is always mutable. priorityClassRef (and its group/kind) is mutable while 
the Workload is pending, and becomes immutable once the `QuotaReserved` condition is
`True`. `priorityClassRef.name` follows the same rule except when `priorityClassRef.kind` is 
`WorkloadPriorityClass`. In that case, only the `.priorityClassRef.name` can still be updated after
the `QuotaReserved` condition is `True`.
```

I think the new sentence is also slightly different from the Priority mutability specification.

tenzen-y (commented):
@atosatto Thank you for clarifying this documentation!
I left a small comment.

site/content/en/docs/concepts/workload_priority_class.md
atosatto: Thank you @tenzen-y. I have reflected your suggestion across both the changed files.

tenzen-y (commented):
Thank you!
/lgtm
/approve

PR #8876, opened 2026-01-29
promote PropagateBatchJobLabelsToWorkload to stable

kannon92:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
PropagateBatchJobLabelsToWorkload was requested to be promoted to GA.
#### Which issue(s) this PR fixes:

Fixes #

#### Special notes for your reviewer:
https://github.com/kubernetes-sigs/kueue/issues/8855#issue-3865341428
#### Does this PR introduce a user-facing change?

```release-note
promote PropagateBatchJobLabelsToWorkload to stable
```

PR #10645, opened 2026-04-21
Promote SkipFinalizersForPodsSuspendedByParent to stable.

mbobrovskyi:
#### What type of PR is this?
/kind feature

Promote SkipFinalizersForPodsSuspendedByParent to stable.

Fixes #

```release-note
Promote SkipFinalizersForPodsSuspendedByParent to stable.
```

### DIFF
--- a/pkg/controller/jobs/pod/pod_webhook.go
+++ b/pkg/controller/jobs/pod/pod_webhook.go
@@ -167,7 +167,7 @@ func (w *PodWebhook) Default(ctx context.Context, obj *corev1.Pod) error {
 	}
 
 	if suspend {
-		if !features.Enabled(features.SkipFinalizersForPodsSuspendedByParent) || !suspendByParent {
+		if !suspendByParent {
 			controllerutil.AddFinalizer(pod.Object(), podconstants.PodFinalizer)
 		}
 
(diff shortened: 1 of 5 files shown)


pkg/controller/jobs/pod/pod_webhook.go
```diff
@@ -167,7 +167,7 @@ func (w *PodWebhook) Default(ctx context.Context, obj *corev1.Pod) error {
 	}
 
 	if suspend {
-		if !features.Enabled(features.SkipFinalizersForPodsSuspendedByParent) || !suspendByParent {
```
mszadkow, line 170: I am not sure about this part, it was not removed yet, so maybe it should stay here?

mbobrovskyi: But we have LockToDefault, so the user can’t change it. I’m also looking at this PR: https://github.com/kubernetes-sigs/kueue/pull/8876 – it looks like we’re removing it.

mszadkow: ah right, then all good I assume

mimowo (commented):
/lgtm
/approve
Thanks 👍

Issue #10660, opened 2026-04-21
TAS: admitted workload with UnhealthyNodes is not evicted due to AdmissionCheck Retry or PodsReadyTimeout


mimowo:
**What happened**:

We have a case when a failing workload with unhealthy nodes is not evicted due to AdmissionCheck=Retry, not PodsReadyTimeout exceeded.

This is the snippet of the workload state:

```yaml
admission_checks:
  - name: "some-ac"
    requeue_after_seconds: 10
    state: "Retry"
conditions:
  - message: "At least one pod has failed, waiting for recovery"
    reason: "WaitForRecovery"
    status: "False"
    type_: "PodsReady"
  - message: "Quota reserved in ClusterQueue cq"
    reason: "QuotaReserved"
    status: "True"
    type_: "QuotaReserved"
  - message: "Previously: Workload was evicted as there was no replacement for unhealthy node(s): xyz"
    reason: "QuotaReserved"
    status: "False"
    type_: "Evicted"
  - message: "The workload is admitted"
    reason: "Admitted"
    status: "True"
    type_: "Admitted"
  - message: "Workload was evicted as there was no replacement for unhealthy node(s): xyz"
    reason: "NodeFailures"
    status: "True"
    type_: "Requeued"
unhealthy_nodes:
  - name: "xyz"
```

This is likely because of this check which short-circuits the additional downstream checks: https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/core/workload_controller.go#L263-L266

This kept the workload "admitted" for unnecessarily long.

A quick experiment shows that all unit tests pass if the check is removed, so maybe we can drop the "if" completely - I see no benefit. It was introduced in https://github.com/kubernetes-sigs/kueue/pull/6648

**What you expected to happen**:

The workload with AdmissionCheck = Retry, or exceeding the PodsReadyTimeout is evicted.

**How to reproduce it (as minimally and precisely as possible)**:

mimowo:
cc @pajakd

vamsikrishna-siddu:
Hi @mimowo  I would like to work on this

mimowo:
/close 
done by https://github.com/kubernetes-sigs/kueue/pull/10666

PR #10666, opened 2026-04-22
tas: allow admitted workloads with unhealthy nodes to be evicted

vamsikrishna-siddu:
#### What type of PR is this?
/kind bug
/kind regression
/area tas

Suggested PR text:

This PR fixes a bug where an admitted workload with `UnhealthyNodes` could skip normal reconciliation entirely due to an early return in the workload controller.

Because of that skip, workloads could remain admitted longer than expected and miss downstream eviction paths, including:
- eviction when an AdmissionCheck is in `Retry`
- eviction when `PodsReady` recovery timeout is exceeded

This change removes that early return so admitted workloads with unhealthy nodes still go through the normal eviction and timeout reconciliation logic.

It also adds regression tests covering:
- admitted workload with unhealthy nodes and AdmissionCheck `Retry`
- admitted workload with unhealthy nodes and `PodsReady=False` with `WaitForRecovery`

Fixes # https://github.com/kubernetes-sigs/kueue/issues/10660

```release-note
TAS: Fix a bug where admitted workloads with unhealthy nodes were not evicted when an AdmissionCheck entered Retry or when the PodsReady recovery timeout was exceeded.
```

### DIFF
--- a/pkg/controller/core/workload_controller_test.go
+++ b/pkg/controller/core/workload_controller_test.go
@@ -1065,6 +1065,105 @@ func TestReconcile(t *testing.T) {
 				},
 			},
 		},
+		"workload with retry checks and unhealthy nodes should be evicted and checks should be pending": {
+			workload: utiltestingapi.MakeWorkload("wl", "ns").
+				ReserveQuotaAt(utiltestingapi.MakeAdmission("q1").Obj(), now).
+				ControllerReference(batchv1.SchemeGroupVersion.WithKind("Job"), "ownername", "owneruid").
+				AdmissionChecks(kueue.AdmissionCheckState{
+					Name:  "check-1",
+					State: kueue.CheckStateRetry,
+				}, kueue.AdmissionCheckState{
+					Name:  "check-2",
+					State: kueue.CheckStateReady,
+				}).
+				Condition(metav1.Condition{
+					Type:               kueue.WorkloadAdmitted,
+					Status:             metav1.ConditionTrue,
+					LastTransitionTime: metav1.NewTime(now),
+					Reason:             "ByTest",
+					Message:            "The workload is admitted",
+				}).
+				AdmittedAt(true, now).
+				UnhealthyNodes("xyz").
+				Obj(),
+			wantWorkload: utiltestingapi.MakeWorkload("wl", "ns").
+				ReserveQuotaAt(utiltestingapi.MakeAdmission("q1").Obj(), now).
+				ControllerReference(batchv1.SchemeGroupVersion.WithKind("Job"), "ownername", "owneruid").
+				AdmissionChecks(kueue.AdmissionCheckState{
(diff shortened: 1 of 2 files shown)


mimowo:
/release-note-edit
```release-note
TAS: fix the bug in NodeHotSwap handling that a workload running on an unhealthy node may not be evicted
immediately even if its AdmissionChecks is flipped to Retry, or it crosses the PodsReady timeout.
```

mimowo:
/release-note-edit
```release-note
TAS: Fix a bug where admitted workloads with unhealthy nodes were not evicted when an AdmissionCheck entered Retry or when the PodsReady recovery timeout was exceeded.
```

pkg/controller/core/workload_controller_test.go
```diff
@@ -1384,6 +1483,117 @@ func TestReconcile(t *testing.T) {
...
+				}).
+				Condition(metav1.Condition{
+					Type:               kueue.WorkloadPodsReady,
+					Status:             metav1.ConditionFalse,
+					LastTransitionTime: metav1.NewTime(now),
+					Reason:             kueue.WorkloadWaitForRecovery,
+					Message:            "At least one pod has failed, waiting for recovery",
+				}).
+				AdmittedAt(true, now).
+				UnhealthyNodes("xyz").
```
mimowo, line 1545: The fact that we now have the workload Evicted looks great, but it seems there might be another bug here, as the Workload continues to have `UnhealthyNodes`. These should be removed atomically during eviction by this code: https://github.com/kubernetes-sigs/kueue/blob/d35e637e7e4dcfc970f6c4b93d3f19366f2a7d23/pkg/workload/workload.go#L1710

cc @pajakd

mimowo: Ah, this is likely some weirdness for how we mimic SSA in unit tests. We also have the safety clean for the field in scheduler: https://github.com/kubernetes-sigs/kueue/blob/main/pkg/scheduler/scheduler.go#L713-L716

mimowo: yeah, I checked with debugger we call TreatSSAAsStrategicMerge with correct workload.

mimowo:
Thank you 👍 As a follow up I think we also need an integration test as a follow up. I think this could land around here: https://github.com/kubernetes-sigs/kueue/blob/1518ce12363a997e0133c93be196f8ae0ddb3308/test/integration/singlecluster/tas/tas_test.go#L3446

/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16

mimowo:
@vamsikrishna-siddu let us know if you would have cycles to work on the integration test as this is a priority fix for us. Ideally we include it in 0.17.2

mimowo:
We need cherrypicks, can you take them @vamsikrishna-siddu or @pajakd ?

Issue #10626, opened 2026-04-20
MultiKueue when Preemption with a multikueue admission check Should re-do admission process when workload gets evicted in the worker


mimowo:
**Which test is flaking?**:

MultiKueue when Preemption with a multikueue admission check Should re-do admission process when workload gets evicted in the worker 

**Link to failed CI job or steps to reproduce locally**:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10616/pull-kueue-test-e2e-multikueue-main/2046209479061540864
**Failure message or logs**:
```
End To End MultiKueue Suite: kindest/node:v1.35.0: [It] MultiKueue when Preemption with a multikueue admission check Should re-do admission process when workload gets evicted in the worker expand_less	51s
{ failed [FAILED] Timed out after 45.000s.
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/multikueue/e2e_test.go:1437 with:
Expected
    <*string | 0x0>: nil
not to be <nil>
In [It] at: /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/multikueue/e2e_test.go:1438 @ 04/20/26 13:10:01.35
}
```

/area multikueue

mimowo:
cc @kshalot

kshalot:
Maybe this was introduced in #9670, since it dealt with worker evictions?

@Singularity23x0 can you take a look?

mimowo:
btw, maybe if the issue is tricky we could start with appending the workload objects to the test failure output? It would be super helpful for debugging. WDYT? For that we have the assertMsg helper.

mimowo:
Another occurence with more data: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10984/pull-kueue-test-e2e-multikueue-release-0-17/2051950681895473152
(full workload yaml this time):

```
End To End MultiKueue Suite: kindest/node:v1.35.0: [It] MultiKueue when Preemption with a multikueue admission check Should re-do admission process when workload gets evicted in the worker expand_less	58s
{ failed [FAILED] Timed out after 45.001s.
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/multikueue/e2e_test.go:1437 with:
Manager Workload is not admitted in worker2
[...]

PR #10649, opened 2026-04-21
Export AssertMsg and AssertMsgObjList

reruno:
#### What type of PR is this?
/kind cleanup

#### What this PR does / why we need it:
Prep PR for cleanup/assert-msg-kuberay-test https://github.com/kubernetes-sigs/kueue/pull/10650

#### Which issue(s) this PR fixes:

Fixes #

#### Special notes for your reviewer:
https://github.com/kubernetes-sigs/kueue/issues/10642#issuecomment-4288028366

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #10684, opened 2026-04-22
Increase flaky pull-kueue-test-e2e-multikueue-main verbosity.

Singularity23x0:
#### What type of PR is this?
/kind flake
/area multikueue
/area testing

Increases the verbosity of a flaky test in the pull-kueue-test-e2e-multikueue-main suite.
Test: MultiKueue when Preemption with a multikueue admission check [It] **Should re-do admission process when workload gets evicted in the worker**
Appends the failing workload to test result message.

While the test is flaky, it fails rarely. To be able to monitor it we want to increase the test's verbosity to be able to investigate future incidents.

Progresses #10626

NONE

```release-note
NONE
```


test/e2e/multikueue/e2e_test.go
```diff
@@ -1434,17 +1434,17 @@ app = HelloWorld.bind()`,
...
-					g.Expect(managerWl.Status.ClusterName).To(gomega.HaveValue(gomega.Equal(workerCluster2.Name)))
+					g.Expect(managerWl.Status.ClusterName).To(gomega.HaveValue(gomega.Equal(workerCluster2.Name)), util.AssertMsg("Manager Workload is not admitted in worker2", managerWl))
 				}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
 			})
 
 			ginkgo.By("Checking that the workload is created in worker2", func() {
 				gomega.Eventually(func(g gomega.Gomega) {
 					g.Expect(k8sWorker2Client.Get(ctx, wlKey, workerWorkload)).To(gomega.Succeed())
-					g.Expect(workload.IsAdmitted(workerWorkload)).To(gomega.BeTrue())
+					g.Expect(workload.IsAdmitted(workerWorkload)).To(gomega.BeTrue(), util.AssertMsg("Worklaod not Admitted in worker2", workerWorkload))
```
kshalot:
```suggestion
					g.Expect(workload.IsAdmitted(workerWorkload)).To(gomega.BeTrue(), util.AssertMsg("Workload not Admitted in worker2", workerWorkload))
```

test/e2e/multikueue/e2e_test.go
```diff
@@ -1434,17 +1434,17 @@ app = HelloWorld.bind()`,
...
 					g.Expect(workerWorkload.Spec).To(gomega.BeComparableTo(managerWl.Spec))
 
-					g.Expect(k8sWorker1Client.Get(ctx, wlKey, workerWorkload)).To(utiltesting.BeNotFoundError())
+					g.Expect(k8sWorker1Client.Get(ctx, wlKey, workerWorkload)).To(utiltesting.BeNotFoundError(), util.AssertMsg("Worklaod present in worker1", workerWorkload))
```
kshalot:
```suggestion
					g.Expect(k8sWorker1Client.Get(ctx, wlKey, workerWorkload)).To(utiltesting.BeNotFoundError(), util.AssertMsg("Workload present in worker1", workerWorkload))
```

test/util/util.go
kshalot, line 172: This was already exported in https://github.com/kubernetes-sigs/kueue/pull/10649

kshalot:
/remove-kind bug

I feel that we should save the `kind/bug` label for bugs in logic. WDYT?

test/util/util.go
Singularity23x0: Synced to main.

Singularity23x0:
> /remove-kind bug
> 
> I feel that we should save the `kind/bug` label for bugs in logic. WDYT?

SGTM

test/e2e/multikueue/e2e_test.go
```diff
@@ -1389,17 +1389,17 @@ app = HelloWorld.bind()`,
 			ginkgo.By("Checking that the workload is created and admitted in the manager cluster", func() {
 				gomega.Eventually(func(g gomega.Gomega) {
 					g.Expect(k8sManagerClient.Get(ctx, wlKey, managerWl)).To(gomega.Succeed())
-					g.Expect(workload.IsAdmitted(managerWl)).To(gomega.BeTrue())
+					g.Expect(workload.IsAdmitted(managerWl)).To(gomega.BeTrue(), util.AssertMsg("Workload not admitted in manager", managerWl))
```
mimowo, line 1392: I'm wondering if we could also include in the messages all 3 related workloads: manager and the 2 workloads, wdyt? 

I'm also good to start as is - just the checked workload, and we extend if we see this is still not enough to debug

mbobrovskyi, line 1392: Please move this as a part of eventually instead. Otherwise ginkgo will execute AssertMsg() in each fail.

kshalot: I was thinking about this as well, but on the other hand it might introduce noise since this test is re-using variables like `managerWl` that contain data from previous `By()` blocks, if the test fails on the `Get` we'll print stale data which might confuse someone. But you'll still be able to see where it failed, so you should be able to catch that when analyzing carefully. I'm on the fence

mimowo:
I think https://github.com/kubernetes-sigs/kueue/pull/10684#discussion_r3123316717 seems reasonable, I'm totally ok either way, and I don't have a strong view here, but let me merge so that we can catch the next failure already. 

/lgtm
/approve 
/cherrypick release-0.17
/cherrypick release-0.16

mimowo:
@Singularity23x0 could you prep a PR for 0.16?

PR #10694, opened 2026-04-22
[refactor] scheduler entry handling

gabesaba:
#### What type of PR is this?
/kind cleanup

Break out scheduler entry handling into its own method,
and further break this into functions/methods for readability.

Fixes #

- I recommend reviewing 1-commit at a time
- Change is LLM generated; I reviewed myself first and endorse them.
- Change is mostly a pure move refactor, except:
  - some changes to comments
  - guard clause addition in `waitForPodsReadyIfBlocked`
  - reuse `mode` variable, rather than calling `RepresentativeMode()` again

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -286,73 +286,7 @@ func (s *Scheduler) schedule(ctx context.Context) wait.SpeedSignal {
 	preemptedWorkloads := make(preemption.PreemptedWorkloads)
 	skippedPreemptions := make(map[kueue.ClusterQueueReference]int)
 	for iterator.hasNext() {
-		e := iterator.pop()
-
-		cq := snapshot.ClusterQueue(e.ClusterQueue)
-		log := log.WithValues("workload", klog.KObj(e.Obj), "clusterQueue", klog.KRef("", string(e.ClusterQueue)))
-		if cq.HasParent() {
-			log = log.WithValues("parentCohort", klog.KRef("", string(cq.Parent().GetName())), "rootCohort", klog.KRef("", string(cq.Parent().Root().GetName())))
-		}
-		ctx := ctrl.LoggerInto(ctx, log)
-		log.V(2).Info("Attempting to schedule workload")
-
-		mode := e.assignment.RepresentativeMode()
-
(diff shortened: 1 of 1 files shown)


gabesaba:
/retest

flake (appeared in 2 other recent PRs): https://github.com/kubernetes-sigs/kueue/issues/10699

pkg/scheduler/scheduler.go
```diff
@@ -448,6 +315,168 @@ func (s *Scheduler) schedule(ctx context.Context) wait.SpeedSignal {
...
+		return
+	}
+	e.status = evicted
+}
+
+// checkPreemptPreconditions handles the two early-exit cases for an entry whose
+// representative mode is Preempt: no candidate targets (optionally reserving
+// capacity), and a closed MultiKueue preemption gate. Returns false when the
+// caller should stop processing this entry.
+func (s *Scheduler) checkPreemptPreconditions(log logr.Logger, e *entry, cq *schdcache.ClusterQueueSnapshot) bool {
```
mimowo: Something feels off with this function, because it is named "check", but has side effects.
I would suggest to either change the name to remove this stateless suggestion, or somehow make the split into functions different.

mimowo (commented):
Thanks this function needed some refactor for long. Could you also list / point out where are the logging improvements. in that case we may consider cherrypicking so that we can already make debugging easier on some evns.

mimowo:
>  Could you also list / point out where are the logging improvements.

Ah sorry, IIUC there are no changes to the logging improvements, I misread the PR description point "some changes to comments" for changes to logs.

pkg/scheduler/scheduler.go
mimowo: Let me know what you think here, in any case this is not a blocker to me, just something that caught attention. The PR already makes the code more readable than it was.

gabesaba: I agree with you. The LLM split this into two blocks, and I think it reads better.

mimowo: Indeed, thanks for following up on the comment

mimowo:
Thank you 👍 
/lgtm

mimowo (commented):
/cherrypick release-0.17
Trying to cherrypick the refactor to make easier cherrypick of the bugfix: https://github.com/kubernetes-sigs/kueue/pull/12419

Issue #10642, opened 2026-04-21
Flaky E2E: Kuberay Should run a rayjob with InTreeAutoscaling


mimowo:
**Which test is flaking?**:

 Kuberay Should run a rayjob with InTreeAutoscaling

**Link to failed CI job or steps to reproduce locally**:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10172/pull-kueue-test-e2e-main-1-35/2046520282943852544
**Failure message or logs**:
```
End To End Suite: kindest/node:v1.35.0: [It] Kuberay Should run a rayjob with InTreeAutoscaling [area:singlecluster, feature:kuberay] expand_less	9m5s
{ failed [FAILED] Timed out after 300.001s.
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/singlecluster/kuberay_test.go:374 with:
Expected
    <v1.JobDeploymentStatus>: Running
to equal
    <v1.JobDeploymentStatus>: Complete
In [It] at: /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/singlecluster/kuberay_test.go:376 @ 04/21/26 09:47:52.909
}
```

mimowo:
cc @yaroslava-serdiuk @sohankunkerkar ptal

mbobrovskyi:
/retitle Flaky E2E: Kuberay Should run a rayjob with InTreeAutoscaling

mimowo:
We could also start here with a prep PR to record the entire state of the RayCluster when that happens using assertMsg. cc @reruno 
I'm wondering maybe the pods completed but just later than usual, maybe we can consider optimizing the subsequent steps that they are more predictible.

cc @sohankunkerkar @yaroslava-serdiuk @hiboyang

mbobrovskyi:
One more https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10656/pull-kueue-test-e2e-main-1-33/2046621701176823808.

sohankunkerkar:
> One more https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10656/pull-kueue-test-e2e-main-1-33/2046621701176823808.

[...]

PR #10707, opened 2026-04-22
Bump Ray version from 2.41.0 to 2.53.0 to fix opencensus SIGABRT

sohankunkerkar:
#### What type of PR is this?
/kind cleanup

Addressing https://github.com/kubernetes-sigs/kueue/issues/10642#issuecomment-4296226760

```release-note
NONE
```


sohankunkerkar:
Need this fix in place: https://github.com/ray-project/kuberay/issues/4760

mimowo:
> Need this fix in place: https://github.com/ray-project/kuberay/issues/4760

Thank you for the investigation. I have some questions:
1. do you know already why the submitter job is failing in new Ray - what is the root cause so that we can help to guide the community there, or even fix by ourselves?
2. does it fail deterministically or occasionally
3. the looks bad, and I'm wondering if/how new KubeRay works on real / not e2e clusters
4. can we maybe attempt the upgrade in the meanwhile to a Ray version which solves the stability issues, but does not run into the newly discovered issue, wdyt?

sohankunkerkar:
>do you know already why the submitter job is failing in new Ray - what is the root cause so that we can help to guide the community there, or even fix by ourselves?

The submitter Job is failing because KubeRay's RayJob controller gates submission using the cached DashboardURL and the deprecated `RayCluster.Status.State`, instead of rechecking whether the head is actually ready to serve requests. Once the dashboard URL is cached, the controller can proceed without verifying `HeadPodReady=True`. With Ray 2.53.0, head startup is slower in our CI environment, so the submitter starts retrying against a dashboard endpoint that is not serving yet, eventually hitting BackoffLimitExceeded.

>does it fail deterministically or occasionally

In CI, this failure mode is reproducible deterministically with Ray 2.53.0. The older Ray 2.41.0 issue is different and appears only occasionally. That older flake is the `opencensus DeltaProducer::RunHarvesterLoop()` crash.

>the looks bad, and I'm wondering if/how new KubeRay works on real / not e2e clusters

I do not think this means new KubeRay is broadly broken on real clusters. On clusters with enough CPU, the head usually comes up quickly enough that the race is masked.

> can we maybe attempt the upgrade in the meanwhile to a Ray version which solves the stability issues, but does not run into the newly discovered issue, wdyt?

Unfortunately I do not see a good intermediate Ray version. The crashing opencensus stats path remains present through the 2.42.x-2.52.x line and is only fully removed in 2.53.0.

mimowo:
Thank you for such a deep analysis.

so iiuc the issue could be fixed in the KubeRay to properly recognize if Head is running or not,  rather than relying on some proxies.

cc @andrewsykim who may help us with promoting the fix inside KubeRay, if feasible

 In the meanwhile Im wondering if we could mitigate the issue by 2x/4x cpu for the head Pod?

sohankunkerkar:
>In the meanwhile Im wondering if we could mitigate the issue by 2x/4x cpu for the head Pod?

Sure, per https://github.com/ray-project/kuberay/blob/v1.6.0/ray-operator/config/samples/ray-cluster.complete.yaml#L54-L56, we can bump it to `500`

pkg/util/testingjobs/rayjob/wrappers.go
```diff
@@ -339,7 +339,7 @@ func (j *JobWrapper) Annotation(key string, value string) *JobWrapper {
 func (j *JobWrapper) EnableInTreeAutoscaling() *JobWrapper {
 	enable := true
 	aggressive := rayv1.UpscalingMode("Aggressive")
-	idleTimeoutSeconds := int32(5)
+	idleTimeoutSeconds := int32(60)
```
sohankunkerkar, line 347: Moving to 60s aligns with KubeRay's standard samples and removes the timing sensitivity:                                                                                                                                                               
                                                                                                                                                                           
  - https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-cluster.autoscaler.yaml#L20-L21                                                       
  - https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-cluster.autoscaler-v2.yaml#L13

test/e2e/singlecluster/kuberay_test.go
```diff
@@ -570,16 +570,6 @@ print([ray.get(my_task.remote(i, 1)) for i in range(32)])`,
 			}, util.MediumTimeout, util.Interval).Should(gomega.Succeed(), util.AssertMsgObjList("Did not observe >=3 workloads from multiple scale-ups", workloadList))
 		})
 
-		ginkgo.By("Waiting for workers reduced to 1 due to scaling down", func() {
```
sohankunkerkar, line 583: This step checked for 1 Running worker pod, but breaks when pods transition to Succeeded after job completion (returns 0 instead of 1). The next step already checks the podset-replica-sizes annotation for count = 1, which is set by Kueue's reconciler and doesn't depend on pod phase.

sohankunkerkar:
@mimowo Based on https://github.com/ray-project/kuberay/issues/4760#issuecomment-4367798220, I think we need the CPU bump regardless.

mimowo:
> @mimowo Based on https://github.com/ray-project/kuberay/issues/4760#issuecomment-4367798220, I think we need the CPU bump regardless.

Awesome ! Thank you for moving this forward 👍 I think we should be good to set the cpu to 500m for the submitter Job in our e2e tests. The only potential case we may not want that is if we are testing defaulting for the k8s submitter Job, but maybe for that we can rely on unit and integration tests.

sohankunkerkar:
> > @mimowo Based on [ray-project/kuberay#4760 (comment)](https://github.com/ray-project/kuberay/issues/4760#issuecomment-4367798220), I think we need the CPU bump regardless.
> 
> Awesome ! Thank you for moving this forward 👍 I think we should be good to set the cpu to 500m for the submitter Job in our e2e tests. The only potential case we may not want that is if we are testing defaulting for the k8s submitter Job, but maybe for that we can rely on unit and integration tests.

@mimowo  500m works locally but not in CI.  I initially tried 500m (CI run failed with 3 test [failures](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10707/pull-kueue-test-e2e-main-1-33/2047774443152871424)). Only 1 CPU made the tests stable. I'm assuming CI kind clusters have more CPU contention than local machines, so the head pod needs more    headroom.

mimowo (commented):
Thank you 👍 
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16

tenzen-y:
@sohankunkerkar Could you manually create cherry pick PRs to release-0.16 and release-0.17 branches?

PR #10723, opened 2026-04-23
hack/releasing: add upgrade note to patch releases

### LINKED ISSUE #10489: Extend docs on installation to include the upgrade policy
**What would you like to be documented or improved**:

I would like to add a section in the Installation page about upgrade policy. 

Sometimes users when upgrading from 0.15.x to 0.16.y may only check the upgrade steps for  0.16.y  skipping the upgrade steps for 0.16.0. 

We should explicitly state in the section, that when upgrading to any patch release > .0 the admin should also check the upgrade steps to the .0 release. 

It would be helpful to even give an example with auto-generated links. 

For example, in order to upgrade to the latest 0.17.1 please also follow the upgrade steps for 0.17.0 (with the appropriate link). 

We could also give this information in the release notes itself.

**Location** (URL, file path, or section if applicable):

https://kueue.sigs.k8s.io/docs/installation/

ShaanveerS:
#### What type of PR is this?
/kind documentation

Add patch upgrade policy to `hack/releasing/sync-notes.sh` for patch releases under "Urgent Upgrade Notes" section.

Before:
```md
## v0.17.1

Changes since `v0.17.0`:

## Urgent Upgrade Notes

### (No, really, you MUST read this before you upgrade)

- AdmissionChecks: Add the alpha `RejectUpdatesToCQWithInvalidOnFlavors` ...

## Changes by Kind
```
  After:
```md
## v0.17.1

Changes since `v0.17.0`:

## Actions Required Before Upgrading

### (No, really, you MUST read this before you upgrade)

- **Minor releases:** Review the `.0` release notes for each new minor version you cross; see: [`v0.16.0`](https://github.com/kubernetes-sigs/kueue/releases/tag/v0.16.0), [`v0.17.0`](https://github.com/kubernetes-sigs/kueue/releases/tag/v0.17.0)
- AdmissionChecks: Add the alpha `RejectUpdatesToCQWithInvalidOnFlavors` ...

## Changes by Kind
```

Fixes #10489 (second part)

```release-note
NONE
```

### DIFF
--- a/hack/releasing/sync-notes.sh
+++ b/hack/releasing/sync-notes.sh
@@ -101,6 +101,89 @@ function find_head_branch() {
   echo "$head_branch"
 }
 
+function release_note_links() {
+  local release_prefix="$1"
+  local start="$2"
+  local end="$3"
+  local release_suffix="$4"
+  local joined=""
+  local current release_version
+
+  for ((current = start; current <= end; current++)); do
+    release_version="${release_prefix}${current}${release_suffix}"
+    if [[ -n "${joined}" ]]; then
+      joined+=", "
+    fi
+    joined+="[\`${release_version}\`](https://github.com/${MAIN_REPO_ORG}/${MAIN_REPO_NAME}/releases/tag/${release_version})"
+  done
+  printf "%s" "${joined}"
+}
+
+# Returns links to the two latest minor `.0` release notes.
+function minor_upgrade_release_links() {
+  local release_version="$1"
+  local major minor patch
+  IFS='.' read -r major minor patch <<< "${release_version#v}"
+
+  local end_minor=$((patch == 0 ? minor - 1 : minor))
+  local start_minor=$((end_minor > 0 ? end_minor - 1 : 0))
+
+  release_note_links "v${major}." "${start_minor}" "${end_minor}" ".0"
+}
+
+# Returns links to earlier patch release notes in the same minor line.
+function patch_upgrade_release_links() {
+  local release_version="$1"
+  local major minor patch
+  IFS='.' read -r major minor patch <<< "${release_version#v}"
+
+  release_note_links "v${major}.${minor}." 1 "$((patch - 1))" ""
+}
+
+function add_upgrade_notes() {
+  local input_file="$1"
+  local output_file="$2"
+
+  # Add the upgrade notes under the upgrade actions section,
+  # or create that section immediately before "## Changes by Kind" if it is missing.
+  awk -v minor_release_links="$3" \
+      -v patch_release_links="$4" '
+    function print_upgrade_notes() {
+      if (minor_release_links != "") {
+        printf "- **Minor releases:** Review the `.0` release notes for each new minor version you cross; see: %s.\n", minor_release_links
+      }
+      if (patch_release_links != "") {
+        printf "- **Patch releases:** Review the patch release notes leading up to this version, but *only* within this minor release line; see: %s.\n", patch_release_links
+      }
+    }
+
+    /^## Urgent Upgrade Notes[[:space:]]*$/ {
+      print "## Actions Required Before Upgrading"
+      next
+    }
+
+    /^### \(No, really, you MUST read this before you upgrade\)[[:space:]]*$/ {
+      print
+      print ""
+      print_upgrade_notes()
+      inserted = 1
+      next
+    }
+
+    /^## Changes by Kind[[:space:]]*$/ && !inserted {
+      print "## Actions Required Before Upgrading"
+      print ""
+      print "### (No, really, you MUST read this before you upgrade)"
+      print ""
+      print_upgrade_notes()
+      print ""
+      inserted = 1
+    }
+
+    { print }
+  ' "$input_file" > "$output_file"
+}
+
 HEAD_BRANCH=$(find_head_branch "$RELEASE_VERSION")
 declare HEAD_BRANCH
 echo "+++ HEAD_BRANCH=$HEAD_BRANCH"
@@ -117,6 +200,9 @@ declare END_SHA
 
 CHANGELOG_FILE=$(mktemp)
 declare CHANGELOG_FILE
+
+FINAL_CHANGELOG_FILE="$CHANGELOG_FILE"
+declare FINAL_CHANGELOG_FILE
 
 echo "+++ Running release-notes:"
 echo
@@ -141,7 +227,18 @@ GITHUB_TOKEN=${GITHUB_TOKEN} release-notes --org "${MAIN_REPO_ORG}" \
               --required-author="" \
               --output="${CHANGELOG_FILE}"
 
-cat "$CHANGELOG_FILE"
+MINOR_UPGRADE_RELEASE_LINKS=$(minor_upgrade_release_links "${RELEASE_VERSION}")
+declare MINOR_UPGRADE_RELEASE_LINKS
+
+PATCH_UPGRADE_RELEASE_LINKS=$(patch_upgrade_release_links "${RELEASE_VERSION}")
+declare PATCH_UPGRADE_RELEASE_LINKS
+
+if [[ -n "${MINOR_UPGRADE_RELEASE_LINKS}${PATCH_UPGRADE_RELEASE_LINKS}" ]]; then
+  FINAL_CHANGELOG_FILE=$(mktemp)
+  add_upgrade_notes "${CHANGELOG_FILE}" "${FINAL_CHANGELOG_FILE}" "${MINOR_UPGRADE_RELEASE_LINKS}" "${PATCH_UPGRADE_RELEASE_LINKS}"
+fi
+
+cat "$FINAL_CHANGELOG_FILE"
 
 echo
 echo
@@ -168,7 +265,7 @@ fi
 
 RELEASE_ISSUE_BODY=$(echo "${RELEASE_ISSUE}" | jq -r '.body')
 
-NEW_RELEASE_ISSUE_BODY=$(awk -v previous_version="$PREVIOUS_VERSION" -v changelog_file="$CHANGELOG_FILE" '
+NEW_RELEASE_ISSUE_BODY=$(awk -v previous_version="$PREVIOUS_VERSION" -v changelog_file="$FINAL_CHANGELOG_FILE" '
   BEGIN {
     in_block=0
     while ((getline line < changelog_file) > 0) {


hack/releasing/sync-notes.sh
```diff
@@ -101,6 +101,40 @@ function find_head_branch() {
   echo "$head_branch"
 }
 
+function add_patch_upgrade_note() {
+  local input_file="$1"
+  local output_file="$2"
+
+  awk -v current_release="${3}" \
```
mimowo: This awk is a bit tricky to follow, could you explain how it works? 

Also, provide some examples for the release notes adjusted by the script for 0.17.1 and 0.17.0. I think for 0.17.0 we don't really need to inject the text.

ShaanveerS: added a comment and updated the description with an example. Thanks!

mimowo (commented):
LGTM, thank you 👍 

but before we merge let me folks who might be interested in the change, maybe there are some good ideas how to improve it even more.

cc @mwysokin @tenzen-y @gabesaba @kannon92 @amy

hack/releasing/sync-notes.sh
mimowo: lgtm, thanks 👍

amy:
To clarify. Lets say I'm upgrading from 0.17.0 to 0.17.3, I should look at the release notes for ..0, ..1, ..2, and ..3 right? (actually asking. I think... that should be the case but should confirm.) 
If this generates that, then LGTM!

mimowo:
> To clarify. Lets say I'm upgrading from 0.17.0 to 0.17.3, I should look at the release notes for ..0, ..1, ..2, and ..3 right? (actually asking. I think... that should be the case but should confirm.)

Yeah, this is a tricky question. To know all differences, yes, but from the upgrade PoV upgrading 0.17.0 -> 0.17.3 should not require release note reading assuming we don't break compatibilities, which we very occasionally do, so for safety we may include the interim releases too. 

This is currently not included in the message.

wdyt?

mimowo:
cc @mwysokin

amy:
> Yeah, this is a tricky question. To know all differences, yes, but from the upgrade PoV upgrading 0.17.0 -> 0.17.3 should not require release note reading assuming we don't break compatibilities, which we very occasionally do, so for safety we may include the interim releases too.

Ah okay. If thats the case then current message LGTM!

tenzen-y:
> > To clarify. Lets say I'm upgrading from 0.17.0 to 0.17.3, I should look at the release notes for ..0, ..1, ..2, and ..3 right? (actually asking. I think... that should be the case but should confirm.)
> 
> Yeah, this is a tricky question. To know all differences, yes, but from the upgrade PoV upgrading 0.17.0 -> 0.17.3 should not require release note reading assuming we don't break compatibilities, which we very occasionally do, so for safety we may include the interim releases too.
> 
> This is currently not included in the message.
> 
> wdyt?

I absolutely agree that we should recommend users should check all patch release notes. We sometimes refine some specifications. Especially, we often aggressively break alpha feature functionalities even in patch version.

mimowo:
> I absolutely agree that we should recommend users should check all patch release notes. We sometimes refine some specifications. Especially, we often aggressively break alpha feature functionalities even in patch version.

Thank you, in that case I will propose small tweaks to this PR and the other PRs to capture that explicitly👍

mimowo:
Ok, after the next round of thinking I would propose:

## v0.17.0

Changes since `v0.16.0`:

## Urgent Upgrade Notes

### (No, really, you MUST read this before you upgrade)

- When upgrading from an earlier minor release, review the upgrade notes for each
  minor `.0` release after your current version, up to and including this minor
  release; see: 0.15.0, 0.16.0
- ...

-------

## Changes by Kind

## v0.17.1

Changes since `v0.17.1`:

## Urgent Upgrade Notes

### (No, really, you MUST read this before you upgrade)

- When upgrading from an earlier minor release, review the upgrade notes for each
  minor `.0` release after your current version, up to and including this minor
  release; see: 0.15.0, 0.16.0, 0.17.0
- ...

## Changes by Kind

-------

## v0.17.2

Changes since `v0.17.2`:

## Urgent Upgrade Notes

### (No, really, you MUST read this before you upgrade)

- When upgrading from an earlier minor release, review the upgrade notes for each
  minor `.0` release after your current version, up to and including this minor
  release; see: 0.16.0, 0.17.0
- When upgrading to this patch release, also review the upgrade notes for earlier
  patch releases in this minor line; see: 0.17.1.
- ...

## Changes by Kind
-----

Here the versions after "see" would be links actually.

mimowo:
@ShaanveerS how do you test the code? Are you able to maybe test it on your for e2e, when the "v0.18.0" release issue is already open?

ShaanveerS:
Tested in https://github.com/ShaanveerS/kueue/issues/1 (0.18.0) and https://github.com/ShaanveerS/kueue/issues/2 (0.17.2)

mimowo:
Thank you, one more remark, might be a follow up, I would like to rename the title "Urgent Upgrade Notes" to  "Upgrade Notes" or "Important Upgrade Notes".

 I'm not clear how they are "Urgent", that suggests to me "upgrade as soon as possible because there is a critical bug (say security)" . 
 
 However, the way we are using this section is more like "When you decide to upgrade please be aware of this potential extra steps or complications", but we don't say "you need to upgrade today".
 
 wdyt @tenzen-y @amy @mwysokin ?

mimowo:
Or more generally I think we could split this section "Urgent Upgrade Steps" into:
1. "New Upgrade Steps"  - here goes what was so far under "Urgent Upgrade Steps"
2. "Upgrade Prerequisites" - here go the new points about checking the previous versions

tenzen-y:
> Thank you, one more remark, might be a follow up, I would like to rename the title "Urgent Upgrade Notes" to "Upgrade Notes" or "Important Upgrade Notes".
> 
> I'm not clear how they are "Urgent", that suggests to me "upgrade as soon as possible because there is a critical bug (say security)" .
> 
> However, the way we are using this section is more like "When you decide to upgrade please be aware of this potential extra steps or complications", but we don't say "you need to upgrade today".
> 
> wdyt @tenzen-y @amy @mwysokin ?

I think that "Urgent" is not weird. That's the same pattern as Kubernetes: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.36.md#urgent-upgrade-notes

mimowo:
> > Thank you, one more remark, might be a follow up, I would like to rename the title "Urgent Upgrade Notes" to "Upgrade Notes" or "Important Upgrade Notes".
> > I'm not clear how they are "Urgent", that suggests to me "upgrade as soon as possible because there is a critical bug (say security)" .
> > However, the way we are using this section is more like "When you decide to upgrade please be aware of this potential extra steps or complications", but we don't say "you need to upgrade today".
> > wdyt @tenzen-y @amy @mwysokin ?
> 
> I think that "Urgent" is not weird. That's the same pattern as Kubernetes: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.36.md#urgent-upgrade-notes

I know it is the same wording in k/k because the same script is used underneath. To my reading:
1. "urgent steps" suggests to me: "do these steps "ASAP". 
2. "Important steps" means: "don't forget about this step". 
And we use it for (2.), conceptually.

However, if you prefer to keep "urgent" for consistency with k/k then I'm totally ok with it.

EDIT: maybe I could also open an issue for the repo hosting the script so that we can discuss the reasons for "urgent" vs "important" there and hear from maintainers, and for now stick to what we have. So that the decisions are de-coupled.

hack/releasing/sync-notes.sh
```diff
@@ -101,6 +101,84 @@ function find_head_branch() {
...
+  local input_file="$1"
+  local output_file="$2"
+
+  # Add the upgrade notes under the existing "Urgent Upgrade Notes" section,
+  # or create that section immediately before "## Changes by Kind" if it is missing.
+  awk -v minor_release_links="$3" \
+      -v patch_release_links="$4" '
+    function print_upgrade_notes() {
+      if (minor_release_links != "") {
+        printf "- When upgrading from an earlier minor release, review the upgrade notes for each minor `.0` release after your current version, up to and including this minor release; see: %s.\n", minor_release_links
```
mwysokin: I'd try to make it simpler and more direct. Something like in the suggestion but don't feel to required to apply the suggestion. Just an idea.

```suggestion
        printf "- **Minor releases:** Review the `.0` release notes for each new minor version you cross; see: %s.\n", minor_release_links
```

hack/releasing/sync-notes.sh
```diff
@@ -101,6 +101,84 @@ function find_head_branch() {
...
+      }
+      if (patch_release_links != "") {
+        printf "- When upgrading to this patch release, also review the upgrade notes for earlier patch releases in this minor line; see: %s.\n", patch_release_links
```
mwysokin: Same as above. Just a suggestion to make it simpler and consistent with the language used in the documentation.

```suggestion
        printf "- **Patch releases:** Review the patch release notes leading up to this version, but *only* within this minor release line; see: %s.\n", patch_release_links
```

hack/releasing/sync-notes.sh
```diff
@@ -101,6 +101,84 @@ function find_head_branch() {
...
+    joined+="[\`${release_version}\`](https://github.com/${MAIN_REPO_ORG}/${MAIN_REPO_NAME}/releases/tag/${release_version})"
+  done
+  printf "%s" "${joined}"
+}
+
+# Returns links to the two latest minor `.0` release notes.
+function minor_upgrade_release_links() {
+  local release_version="$1"
+  local major minor patch
+  IFS='.' read -r major minor patch <<< "${release_version#v}"
```
mwysokin, line 126: Great logic here, but there is a bug that will crash the script during pre-releases.

Justification: If we ever generate release notes for a tag with a suffix (e.g., `v0.8.2-rc.1`), the current `IFS` parsing will assign `2-rc.1` to the `patch` variable. When bash tries to evaluate `$((patch == 0))`, it will throw a syntax error (`invalid arithmetic operator`). We need to strip out any pre-release suffix before performing math operations on the patch number.

```suggestion
  local major minor patch raw_patch
  IFS='.' read -r major minor raw_patch <<< "${release_version#v}"
  patch="${raw_patch%%-*}"
```

hack/releasing/sync-notes.sh
```diff
@@ -101,6 +101,84 @@ function find_head_branch() {
...
+  local start_minor=$((end_minor > 0 ? end_minor - 1 : 0))
+
+  release_note_links "v${major}." "${start_minor}" "${end_minor}" ".0"
+}
+
+# Returns links to earlier patch release notes in the same minor line.
+function patch_upgrade_release_links() {
+  local release_version="$1"
+  local major minor patch
+  IFS='.' read -r major minor patch <<< "${release_version#v}"
```
mwysokin, line 138: Same minor bug as above.

Justification: We need to apply the exact same suffix-stripping logic here to ensure that `patch` is strictly a number. Otherwise, generating notes for alphas, betas, or RCs will result in a bash syntax crash on line 140 when evaluating `$((patch - 1))`.

```suggestion
  local major minor patch raw_patch
  IFS='.' read -r major minor raw_patch <<< "${release_version#v}"
  patch="${raw_patch%%-*}"
```

mwysokin (commented):
Submitted 4 comments: 2 are regarding bugs which I think need need to be fixed and 2 are suggestions to make the content more concise.

mwysokin:
> > > Thank you, one more remark, might be a follow up, I would like to rename the title "Urgent Upgrade Notes" to "Upgrade Notes" or "Important Upgrade Notes".
> > > I'm not clear how they are "Urgent", that suggests to me "upgrade as soon as possible because there is a critical bug (say security)" .
> > > However, the way we are using this section is more like "When you decide to upgrade please be aware of this potential extra steps or complications", but we don't say "you need to upgrade today".
> > > wdyt @tenzen-y @amy @mwysokin ?
> > 
> > 
> > I think that "Urgent" is not weird. That's the same pattern as Kubernetes: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.36.md#urgent-upgrade-notes
> 
> I know it is the same wording in k/k because the same script is used underneath. To my reading:
> 
> 1. "urgent steps" suggests to me: "do these steps "ASAP".
> 2. "Important steps" means: "don't forget about this step".
>    And we use it for (2.), conceptually.
> 
> However, if you prefer to keep "urgent" for consistency with k/k then I'm totally ok with it.
> 
> EDIT: maybe I could also open an issue for the repo hosting the script so that we can discuss the reasons for "urgent" vs "important" there and hear from maintainers, and for now stick to what we have. So that the decisions are de-coupled.

+1 I don't think urgent fits here. Urgent gives some sense of urgency here we should give some sense of importance. I'd propose either:
- **Action Required Before Upgrading**: I think it is the most common phras I've seen on Github
- **Breaking Changes and Upgrade Instructions**: Plain, simple and direct

@mimowo @ShaanveerS WDYT?

mimowo:
I like both proposals, but slightly leaning to "Actions Required Before Upgrading" - this calls for action and doesn't have the negative connotation of "Breaking Changes", even if the actions are due to breaking changes :)  

It also feels better to describe the section in its current form.

hack/releasing/sync-notes.sh
ShaanveerS: Thanks for catching this. I think this bug is not reachable today since pre-releases are rejected before we parse the version:
https://github.com/kubernetes-sigs/kueue/blob/ba7c512d557dd2571025dad68983d606f180934b/hack/releasing/sync-notes.sh#L52-L54

ShaanveerS:
Done, switched to `Actions Required Before Upgrading`. Thanks both!
[v.0.18.0 test](https://github.com/ShaanveerS/kueue/issues/1)

hack/releasing/sync-notes.sh
ShaanveerS: done!

hack/releasing/sync-notes.sh
ShaanveerS: done!

mimowo:
LGTM, thank you 👍

mimowo:
> I think that "Urgent" is not weird. That's the same pattern as Kubernetes: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.36.md#urgent-upgrade-notes

@tenzen-y are you ok with the latest proposal, and "Action Required Before Upgrading" for the section name?

mimowo:
@ShaanveerS could you update the PR description with the latest version of the release-notes?

tenzen-y:
> > I think that "Urgent" is not weird. That's the same pattern as Kubernetes: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.36.md#urgent-upgrade-notes
> 
> @tenzen-y are you ok with the latest proposal, and "Action Required Before Upgrading" for the section name?

I'm fine with that 👍 Thank you for discussing some representations!

mimowo:
Thank you, let's see how this works in practice
/lgtm
/approve

PR #10725, opened 2026-04-23
Propagate accumulated past execution time in MultiKueue

### LINKED ISSUE #10437: Missing Workload status(accumulatedPastExecutionTimeSeconds)
**What happened**:
In a MultiKueue environment, when a workload is executed on a worker cluster, 
the workload.status.accumulatedPastExecutionTimeSeconds field is not being populated/updated on the Workload object. This field remains unset or at its initial value (0 or nil) even after the workload has completed execution on the worker cluster.

The workload.status.accumulatedPastExecutionTimeSeconds field should be updated with the actual execution time when a workload completes on a worker cluster in a MultiKueue setup. This field is critical for tracking workload re-execution and implementing features based on past execution history.

1. Set up a MultiKueue environment with a manager cluster and at least one worker cluster
2. Create a workload that gets admitted to the worker cluster
3. Wait for the workload to complete execution on the worker cluster
4. Check the workload status on the manager cluster:
$ kubectl get workload <workload-name> -n <namespace> -o jsonpath='{.status.accumulatedPastExecutionTimeSeconds}'
5. Observe that the field is not populated (returns empty, 0, or nil)

**Environment**:
- Kubernetes version (use `kubectl version`): v1.31.14
- Kueue version (use `git describe --tags --dirty --always`): v0.16.4
- Cloud provider or hardware configuration: HP A100 GPU Server 
- OS (e.g: `cat /etc/os-release`): Ubuntu 24.04.3
- Kernel (e.g. `uname -a`): Linux 6.8.0-90-generic
- Install tools: Helm

MatteoFari:
#### What type of PR is this?
/kind bug
/area multikueue

This PR fixes a MultiKueue bug where the manager workload could miss `status.accumulatedPastExecutionTimeSeconds` after a remote workload finished.

When a remote workload finishes, the manager now copies the past execution time and updates the local workload before marking it finished.

Fixes #10437

```release-note
Fixed MultiKueue bug where manager-side Workloads could miss
`status.accumulatedPastExecutionTimeSeconds` after the worker-side Workload finished.
```

### DIFF
--- a/pkg/controller/admissionchecks/multikueue/workload.go
+++ b/pkg/controller/admissionchecks/multikueue/workload.go
@@ -128,6 +128,32 @@ func (g *wlGroup) bestMatchByCondition(conditionType string) (*metav1.Condition,
 	return bestMatchCond, bestMatchRemote
 }
 
+// finishedRemotePastAdmittedTime returns the execution time to copy to the manager workload.
+// It includes the time already accumulated by the remote workload and, when the remote
+// workload is still admitted, the current interval from its Admitted condition to its
+// Finished condition. It returns nil when the inputs are missing or the remote workload
+// has no accumulated time and no admitted interval to derive from.
+func finishedRemotePastAdmittedTime(remoteWl *kueue.Workload, remoteFinishedCond *metav1.Condition) *int32 {
+	if remoteWl == nil || remoteFinishedCond == nil {
+		return nil
+	}
+
+	pastAdmittedTime := ptr.Deref(remoteWl.Status.AccumulatedPastExecutionTimeSeconds, 0)
+	hasPastAdmittedTime := remoteWl.Status.AccumulatedPastExecutionTimeSeconds != nil
+
+	if remoteAdmittedCond := apimeta.FindStatusCondition(remoteWl.Status.Conditions, kueue.WorkloadAdmitted); remoteAdmittedCond != nil && remoteAdmittedCond.Status == metav1.ConditionTrue {
+		currentAdmittedTime := max(int32(0), int32(remoteFinishedCond.LastTransitionTime.Sub(remoteAdmittedCond.LastTransitionTime.Time).Seconds()))
+		pastAdmittedTime += currentAdmittedTime
+		hasPastAdmittedTime = true
+	}
+
+	if !hasPastAdmittedTime {
+		return nil
+	}
+
+	return new(pastAdmittedTime)
+}
+
 // RemoveRemoteObjects deletes the remote controller object and workload for a cluster.
 // The controller object is deleted first to handle cases where GC has already removed
 // the remote workload.
@@ -363,6 +389,18 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
 			}
 		} else {
 			log.V(3).Info("Group with no adapter, skip owner status copy", "workerCluster", remote)
+		}
+
+		if remotePastAdmittedTime := finishedRemotePastAdmittedTime(group.remotes[remote], remoteFinishedCond); remotePastAdmittedTime != nil {
+			if err := workload.PatchAdmissionStatus(ctx, w.client, group.local, w.clock, func(wl *kueue.Workload) (bool, error) {
+				if wl.Status.AccumulatedPastExecutionTimeSeconds != nil && *wl.Status.AccumulatedPastExecutionTimeSeconds == *remotePastAdmittedTime {
+					return false, nil
+				}
+				wl.Status.AccumulatedPastExecutionTimeSeconds = new(*remotePastAdmittedTime)
+				return true, nil
+			}); err != nil {
+				return reconcile.Result{}, err
+			}
 		}
 
 		// finish workload and copy the status to the local one
--- a/pkg/controller/admissionchecks/multikueue/workload_test.go
+++ b/pkg/controller/admissionchecks/multikueue/workload_test.go
@@ -714,6 +714,135 @@ func TestWlReconcile(t *testing.T) {
 					Obj(),
 			},
 		},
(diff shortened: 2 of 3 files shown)


mimowo:
cc @kshalot @olekzabl

kannon92:
It would be good to have an integration test in test/integration/multikueue reproducing this bug.

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -128,6 +128,27 @@ func (g *wlGroup) bestMatchByCondition(conditionType string) (*metav1.Condition,
 	return bestMatchCond, bestMatchRemote
 }
 
+func finishedRemotePastAdmittedTime(remoteWl *kueue.Workload, remoteFinishedCond *metav1.Condition) *int32 {
```
mwielgus, line 136: Add a comment of what exactly is returned and when the output is nil.

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -128,6 +128,32 @@ func (g *wlGroup) bestMatchByCondition(conditionType string) (*metav1.Condition,
...
+// It includes the time already accumulated by the remote workload and, when the remote
+// workload is still admitted, the current interval from its Admitted condition to its
+// Finished condition. It returns nil when the inputs are missing or the remote workload
+// has no accumulated time and no admitted interval to derive from.
+func finishedRemotePastAdmittedTime(remoteWl *kueue.Workload, remoteFinishedCond *metav1.Condition) *int32 {
+	if remoteWl == nil || remoteFinishedCond == nil {
+		return nil
+	}
+
+	pastAdmittedTime := ptr.Deref(remoteWl.Status.AccumulatedPastExecutionTimeSeconds, 0)
```
kshalot, line 141: I don't think this accumulation will work in the MultiKueue case, because evictions remove the workload object on the remotes. So this fix would make the final `AccumulatedPastExecutionTimeSeconds` equal to the execution time in the **last** admit-finish cycle, not the accumulated one.

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -365,6 +391,18 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
 			log.V(3).Info("Group with no adapter, skip owner status copy", "workerCluster", remote)
 		}
 
+		if remotePastAdmittedTime := finishedRemotePastAdmittedTime(group.remotes[remote], remoteFinishedCond); remotePastAdmittedTime != nil {
+			if err := workload.PatchAdmissionStatus(ctx, w.client, group.local, w.clock, func(wl *kueue.Workload) (bool, error) {
+				if wl.Status.AccumulatedPastExecutionTimeSeconds != nil && *wl.Status.AccumulatedPastExecutionTimeSeconds == *remotePastAdmittedTime {
+					return false, nil
+				}
+				wl.Status.AccumulatedPastExecutionTimeSeconds = new(*remotePastAdmittedTime)
```
kshalot, line 399: Hm, I don't understand why this is needed to be honest... Isn't`AccumulatedPastExecutionTimeSeconds` managed on the manager workload independently? This could lead to double-counting in my opinion, since the local workload is still subject to reconciliations from the core controllers (for example the controller that checks `AccumulatedPastExecutionTimeSeconds` and the max execution time).

Am I missing something?

kshalot (commented):
Here's my understanding - Kueue calls `SyncAdmittedCondition` upon eviction:
https://github.com/kubernetes-sigs/kueue/blob/78109e3799c9ff98ada439b99fb92f00c70c634c/pkg/controller/core/workload_controller.go#L559
and `SyncAdmittedCondition` is what updates `AccumulatedPastExecutionTimeSeconds`:
https://github.com/kubernetes-sigs/kueue/blob/78109e3799c9ff98ada439b99fb92f00c70c634c/pkg/workload/admissionchecks.go#L75-L87

So we won't see `AccumulatedPastExecutionTimeSeconds` increase if we do not evict. Eviction also wasn't included in the reproduction steps.

@MatteoFari Could you confirm that if the reproduction steps involve an eviction, the time is increased?

pkg/controller/admissionchecks/multikueue/workload.go
MatteoFari: Yeah, remote workloads can be recreated without prior status, so using the remote value can lose earlier cycles. Keeping the manager workload as the source of truth seems better.

pkg/controller/admissionchecks/multikueue/workload.go
MatteoFari: I think that’s right, I was missing that the manager workload already has the state needed for this. Should I rework this so the manager folds in its own current admitted interval before finish, instead of copying the remote value?

MatteoFari:
@kshalot Yes, if the manager workload is evicted, the manager-side time should increase through SyncAdmittedCondition.

The missing case is a normal successful finish without eviction. I think the best fix is to keep the manager workload as the source of truth and add its current admitted time before marking it Finished.

kshalot:
> I think the best fix is to keep the manager workload as the source of truth and add its current admitted time before marking it Finished.

The fact that the time of the current cycle is not added to `AccumulatedPastExecutionTimeSeconds` is not MultiKueue specific. It's an accounting field designed to hold the time of the past executions:
https://github.com/kubernetes-sigs/kueue/blob/705b942982b7177656235f8038504985deba4aa8/apis/kueue/v1beta2/workload_types.go#L675-L679

So this wouldn't really be a "fix" but a change in the established semantics of this field, which might be breaking to somebody. I opened a discussion about this in the original issue: https://github.com/kubernetes-sigs/kueue/issues/10437#issuecomment-4314261301

PR #10742, opened 2026-04-23
Update kuberay to 1.6.1

hiboyang:
#### What type of PR is this?

/kind bug

Update to KubeRay 1.6.1 which has fix for https://github.com/ray-project/kuberay/issues/4686

Fixes KubeRay issue https://github.com/ray-project/kuberay/issues/4686

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/prometheus/client_golang v1.23.2
 	github.com/prometheus/client_model v0.6.2
 	github.com/prometheus/common v0.67.5
-	github.com/ray-project/kuberay/ray-operator v1.6.0
+	github.com/ray-project/kuberay/ray-operator v1.6.1
 	github.com/spf13/cobra v1.10.2
 	github.com/spf13/pflag v1.0.10
 	go.uber.org/mock v0.6.0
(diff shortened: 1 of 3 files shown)


kannon92:
Do you think this should be cherry-pick?

kannon92:
Looks like it would be somewhat easy to carry for 0.17 (0.17 is on ray 1.6.0).

kannon92 (commented):
/lgtm
/assign @mimowo @tenzen-y

hiboyang:
> Do you think this should be cherry-pick?

Yes, better to cherry-pick to 0.17 and 0.16, so we can port RayService autoscaling e2e test there as well. I remember there is some bot to automatcially cherry-pick, which we might try first before doing manual cherry-pick.

hiboyang:
@kannon92 I just noticed kueue 0.16 still use [KubeRay 1.5.1.](https://github.com/kubernetes-sigs/kueue/blob/e5bea04a9b47ff2241234f99e20dd22884444e26/go.mod#L20). Seems we should keep it as is, and do not cherry-pick this PR to kueue 0.16.

PR #10747, opened 2026-04-24
fix(metrics): convert milliCPU to CPU units

### LINKED ISSUE #10746: kueue_cohort_subtree_quota reports CPU as 1000x the actual value
**What happened**:
When exporting the cpu metric from kueue_cohort_subtree_quota, it shows up as 1000x the actual value
i.e. 96 cpu in a resource flavor in a cohort should report as 96 cpu not 96000

curl metrics endpoint and grep for subtree
`curl https://localhost:8080/metrics | grep subtree`

**Environment**:
- Kubernetes version (use `kubectl version`): 1.33
- Kueue version (use `git describe --tags --dirty --always`): 0.17.1
- Cloud provider or hardware configuration: EKS
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

baoalvin1:
#### What type of PR is this?
/kind bug

in cohort subtree metrics `kueue_cohort_subtree_quota` and `kueue_cohort_subtree_resource_reservations` were reporting raw milliCPU values (e.g. 30000) instead of CPU units (e.g. 30.0), unlike the equivalent ClusterQueue metrics which correctly use resourceFloat(). Apply the same conversion in applyCohortMetricPoint and update the metric function signatures to accept float64.

Fixes https://github.com/kubernetes-sigs/kueue/issues/10746

```release-note
Observability: Fix kueue_cohort_subtree_quota and kueue_cohort_subtree_resource_reservations metrics incorrectly reporting raw milliCPU values instead of CPU units for CPU resources.
```

### DIFF
--- a/pkg/cache/scheduler/cohort_metrics_test.go
+++ b/pkg/cache/scheduler/cohort_metrics_test.go
@@ -113,12 +113,12 @@ func TestRecordCohortMetrics_Guards(t *testing.T) {
 					{
 						cohort: cohortLeft,
 						fr:     fr,
-						quota:  float64(cache.hm.Cohort(cohortLeft).resourceNode.SubtreeQuota[fr]),
+						quota:  resourceFloat(fr.Resource, cache.hm.Cohort(cohortLeft).resourceNode.SubtreeQuota[fr]),
 					},
 					{
 						cohort: cohortRoot,
 						fr:     fr,
-						quota:  float64(cache.hm.Cohort(cohortRoot).resourceNode.SubtreeQuota[fr]),
+						quota:  resourceFloat(fr.Resource, cache.hm.Cohort(cohortRoot).resourceNode.SubtreeQuota[fr]),
 					},
 				}
 			},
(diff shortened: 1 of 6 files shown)


pkg/cache/scheduler/cohort_metrics_test.go
```diff
@@ -352,18 +352,18 @@ func TestClearCohortMetrics_ClearsTargetAndAncestors(t *testing.T) {
...
-	expectGaugeValue(t, kueuemetrics.CohortSubtreeResourceReservations, cohortMetricLabels("root", frDefaultCPU), 13_000)
+	expectGaugeValue(t, kueuemetrics.CohortSubtreeQuota, cohortMetricLabels("ch1", frDefaultCPU), 18)
+	expectGaugeValue(t, kueuemetrics.CohortSubtreeQuota, cohortMetricLabels("ch2", frDefaultCPU), 14)
+	expectGaugeValue(t, kueuemetrics.CohortSubtreeQuota, cohortMetricLabels("root", frDefaultCPU), 62)
+	expectGaugeValue(t, kueuemetrics.CohortSubtreeResourceReservations, cohortMetricLabels("root", frDefaultCPU), 13)
 
 	cache.ClearCohortMetrics(log, "ch1")
 	expectGaugeCount(t, kueuemetrics.CohortSubtreeQuota, 0, cohortMetricLabels("ch1", frDefaultCPU))
 	expectGaugeCount(t, kueuemetrics.CohortSubtreeResourceReservations, 0, cohortMetricLabels("ch1", frDefaultCPU))
-	expectGaugeValue(t, kueuemetrics.CohortSubtreeQuota, cohortMetricLabels("ch2", frDefaultCPU), 14_000)
```
kannon92, line 363: Hmm. I wonder if this was by design considering we have tests covering this.

kannon92 (commented):
Seeing the tests I'm wondering if this is by design. What would show up for CPU if it's actually using milliCPUs. 500 milli?

kannon92:
Can you add a release note? This is a user facing bug.

baoalvin1:
@kannon92 no it shows up as an int value, I think the tests are incorrect because the clusterqueue metrics are correct

baoalvin1:
```
kueue_cohort_subtree_quota{cohort="cohort",flavor="flavor1",replica_role="leader",resource="cpu"} 96000
```
is what the /metrics endpoint outputs even though I only have 96 cpus in the flavor

pkg/cache/scheduler/cohort_metrics_test.go
kannon92: cc @mimowo @tenzen-y 

Curious if you expect this behavior? Just surprised to see tests expecting this.

mimowo: Thanks for calling this out. I think this is a bug that just slipped reviewer attention.

cc @mszadkow in case he has some more thoughts here, but Im pretty sure a bug

mszadkow: fully agree, now as you pointed that out, I think it's a bug too

mimowo:
Thanks 👍 
/lgtm
/approve
/cherrypick release-0.17

tenzen-y:
/release-note-edit

```release-note
Observability: Fix kueue_cohort_subtree_quota and kueue_cohort_subtree_resource_reservations metrics incorrectly reporting raw milliCPU values instead of CPU units for CPU resources.
```

PR #10752, opened 2026-04-24
Fix TopologyAwareScheduling when Creating a JobSet requesting TAS should admit a JobSet via TAS test.

### LINKED ISSUE #10751: Flaky E2E: TopologyAwareScheduling when Creating a JobSet requesting TAS should admit a JobSet via TAS [area:singlecluster, feature:tas]
**Which test is flaking?**:
] TopologyAwareScheduling when Creating a JobSet requesting TAS should admit a JobSet via TAS [area:singlecluster, feature:tas]

**Failure message or logs**:
```
End To End Suite: kindest/node:v1.35.0: [It] TopologyAwareScheduling when Creating a JobSet requesting TAS should admit a JobSet via TAS [area:singlecluster, feature:tas] expand_less	49s
{ failed [FAILED] Timed out after 45.000s.
The function passed to Eventually failed at /home/prow/go/src/kubernetes-sigs/kueue/test/e2e/singlecluster/tas_test.go:277 with:
Expected
    <[]v1.Condition | len:2, cap:2>: [
        {
            Type: "QuotaReserved",
            Status: "True",
            ObservedGeneration: 1,
            LastTransitionTime: {
                Time: 2026-04-24T02:57:58Z,
            },
            Reason: "QuotaReserved",
            Message: "Quota reserved in ClusterQueue cluster-queue-e2e-tas-9prcq",
        },
        {
            Type: "Admitted",
            Status: "True",
            ObservedGeneration: 1,
            LastTransitionTime: {
                Time: 2026-04-24T02:57:58Z,
            },
            Reason: "Admitted",
            Message: "The workload is admitted",
        },
    ]
to have condition type Finished and status True
In [It] at: /home/prow/go/src/kubernetes-sigs/kueue/test/e2e/singlecluster/tas_test.go:278 @ 04/24/26 02:58:44.063
}
```

We are using the MediumTimeout awaiting for Pod termination, that should be changed to LongTimeout, plus we have already some helper functions to await for workload termination in e2e tests.

mbobrovskyi:
#### What type of PR is this?
/kind bug
/kind failing-test
/kind flake

Fix TopologyAwareScheduling when Creating a JobSet requesting TAS should admit a JobSet via TAS test.

Fixes #10751

```release-note
NONE
```

### DIFF
--- a/test/e2e/singlecluster/job_test.go
+++ b/test/e2e/singlecluster/job_test.go
@@ -17,13 +17,13 @@ limitations under the License.
 package singlecluster
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -211,18 +211,16 @@ var _ = ginkgo.Describe("Kueue", ginkgo.Label("area:singlecluster", "feature:job
 			sampleJob = (&testingjob.JobWrapper{Job: *sampleJob}).Image(util.GetAgnHostImage(), util.BehaviorExitFast).Obj()
 			util.MustCreate(ctx, k8sClient, sampleJob)
 
-			createdWorkload := &kueue.Workload{}
-
 			// The job might have finished at this point. That shouldn't be a problem for the purpose of this test
 			util.ExpectJobUnsuspendedWithNodeSelectors(ctx, k8sClient, jobKey, map[string]string{
 				"instance-type": "on-demand",
 			})
+
 			wlLookupKey := types.NamespacedName{Name: workloadjob.GetWorkloadNameForJob(sampleJob.Name, sampleJob.UID), Namespace: ns.Name}
-			gomega.Eventually(func(g gomega.Gomega) {
-				g.Expect(k8sClient.Get(ctx, wlLookupKey, createdWorkload)).Should(gomega.Succeed())
-				g.Expect(workload.HasQuotaReservation(createdWorkload)).Should(gomega.BeTrue())
-				g.Expect(createdWorkload.Status.Conditions).Should(utiltesting.HaveConditionStatusTrue(kueue.WorkloadFinished))
-			}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
+
+			ginkgo.By(fmt.Sprintf("Wait for the workload %q to finish", wlLookupKey), func() {
+				util.ExpectWorkloadToFinishWithTimeout(ctx, k8sClient, wlLookupKey, util.LongTimeout)
+			})
 		})
 
 		ginkgo.It("Should run with prebuilt workload", func() {
@@ -294,11 +292,11 @@ var _ = ginkgo.Describe("Kueue", ginkgo.Label("area:singlecluster", "feature:job
 			})
 
 			ginkgo.By("Await for jobs completion", func() {
-				gomega.Eventually(func(g gomega.Gomega) {
-					g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(wl), createdWorkload)).To(gomega.Succeed())
-					g.Expect(createdWorkload.Finalizers).NotTo(gomega.ContainElement(kueue.ResourceInUseFinalizerName))
-					g.Expect(createdWorkload.Status.Conditions).To(utiltesting.HaveConditionStatusTrueAndReason(kueue.WorkloadFinished, kueue.WorkloadFinishedReasonFailed))
-				}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
+				util.ExpectWorkloadToFinishWithTimeout(ctx, k8sClient, wlLookupKey, util.LongTimeout)
+			})
+
+			ginkgo.By("Await workload finalized", func() {
+				util.ExpectWorkloadsFinalizedOrGone(ctx, k8sClient, wlLookupKey)
 			})
 		})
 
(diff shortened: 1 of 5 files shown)


test/e2e/singlecluster/tas_test.go
```diff
@@ -270,11 +270,7 @@ var _ = ginkgo.Describe("TopologyAwareScheduling", ginkgo.Label("area:singleclus
 			})
 
 			ginkgo.By(fmt.Sprintf("verify the workload %q gets finished", wlLookupKey), func() {
-				gomega.Eventually(func(g gomega.Gomega) {
-					g.Expect(k8sClient.Get(ctx, wlLookupKey, createdWorkload)).Should(gomega.Succeed())
-					g.Expect(workload.HasQuotaReservation(createdWorkload)).Should(gomega.BeTrue())
-					g.Expect(createdWorkload.Status.Conditions).Should(utiltesting.HaveConditionStatusTrue(kueue.WorkloadFinished))
-				}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
+				util.ExpectWorkloadToFinishWithTimeout(ctx, k8sClient, wlLookupKey, util.LongTimeout)
```
mimowo, line 267: Should we use the same helper also in other places? I would assume so, or this spot is somehow special

mszadkow: I wonder if checking quota reservation along side finished state brings any value?
If not, then we might do what @mimowo mentioned and replace those check pairs with the wrapper

mimowo: > I wonder if checking quota reservation along side finished state brings any value?

It could be useful if this was a separate assert, separete ginkgo.By. So we could keep it as part of the previous ginkgo.By: ginkgo.By(fmt.Sprintf("await for admission of workload %q and verify TopologyAssignment", wlLookupKey), func() {

mimowo: > If not, then we might do what @mimowo mentioned and replace those check pairs with the wrapper

I don't have a strong view about the QuotaReservation check - eiither a separate ginkgo.By, or merge with previous step. 

Still, I think we should use the extended timeout for all places where we are waiting for workload finished in e2e tests.

We don't need to cover all tests in this PR, but as I checked there are more similar checks, so I'm wondering if we could extend the scope of the PR slightly, at least for the analogous cases.

mbobrovskyi: > I wonder if checking quota reservation along side finished state brings any value?

I'm not sure we need to check this here. We already know the workload is admitted in the "waiting for the JobSet to be unsuspended" case.

mimowo: Exactly, I'm ok to drop this check. I'm just thinking it might be a good idea to also cover other places doing the smilar check in this file

test/util/util.go
```diff
@@ -569,6 +570,15 @@ func expectWorkloadsWithPriority(
 	}, Timeout, Interval).Should(gomega.Succeed(), AssertMsg("Workload priority does not match expected", createdWl))
 }
 
+func ExpectWorkloadFinalized(ctx context.Context, k8sClient client.Client, wlKey client.ObjectKey) {
```
mimowo: It does not seem used

mbobrovskyi: Removed

test/e2e/singlecluster/job_test.go
```diff
@@ -402,18 +400,10 @@ var _ = ginkgo.Describe("Kueue", ginkgo.Label("area:singlecluster", "feature:job
...
-						return
-					}
-					g.Expect(err).To(gomega.Not(gomega.HaveOccurred()))
-					g.Expect(workload.HasQuotaReservation(createdWorkload)).Should(gomega.BeTrue())
-					g.Expect(createdWorkload.Status.Conditions).Should(utiltesting.HaveConditionStatusTrue(kueue.WorkloadFinished))
-				}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
+			wlLookupKey := types.NamespacedName{Name: workloadjob.GetWorkloadNameForJob(job.Name, job.UID), Namespace: ns.Name}
+
+			ginkgo.By(fmt.Sprintf("wait for the workload %q to finish", wlLookupKey), func() {
+				util.ExpectWorkloadToFinishOrGoneWithTimeout(ctx, k8sClient, wlLookupKey, util.LongTimeout)
```
mimowo: I see why you add the function to reflect the previous code, but do we really need to pass the test when the workload is deleted? I'm not sure what could delete the workload, iiuc the workload should exist after the job is finshed, so we could simplify this case. Maybe we should check which PRs added this code, but this looks unnecessary, wdyt?

mbobrovskyi: Actually, here we are checking that the finalizers are empty after the Job finishes. I think we should keep this to cover that case. We may not need to check for Gone, but we should still verify that the finalizers are empty.

mbobrovskyi: Ah, wait. We are talking about another case. Yes, you right here we can simplify.

test/e2e/singlecluster/job_test.go
```diff
@@ -294,11 +292,11 @@ var _ = ginkgo.Describe("Kueue", ginkgo.Label("area:singlecluster", "feature:job
 			})
 
 			ginkgo.By("Await for jobs completion", func() {
-				gomega.Eventually(func(g gomega.Gomega) {
-					g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(wl), createdWorkload)).To(gomega.Succeed())
-					g.Expect(createdWorkload.Finalizers).NotTo(gomega.ContainElement(kueue.ResourceInUseFinalizerName))
```
mbobrovskyi, line 299: I told about this case (https://github.com/kubernetes-sigs/kueue/pull/10752/changes#r3148888419).

mimowo: Hmm, this assert is very weird actually, because it is checking that the workload is finished, but if the workload is deleted quickly, then the assert will fail. So IIUC it is passing only because typically the GC collector takes time. IIUC, the workload gets deleted eventually by GC anyway.

If this is true, when I think we should not be really checking if the workload is finsished. I would rather just check `ExpectWorkloadsFinalizedOrGone` - I think checking if finished is flaky,

mbobrovskyi: I think we should first check for the finished status, and then verify the finalizers or whether the workload is already gone. This is what I’ve already implemented.

mimowo: I know you have implemented it, and i know this is how it was.

However I think this is not correct and may flake. If the workload is GCed immediately after finishing then we may not have time to obsrrve it is finished. 

Sure GC works in 20s interval so probably this is very unlikely but I think possible

Let me know what im missing

mimowo: ok in any case we can revisit this as a follow up, this is not a regression, just something that caught my attention

mimowo:
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16
Thank you

test/e2e/singlecluster/job_test.go
mbobrovskyi: So, are you propose to check `OR finished OR finalized OR gone` together? Because currently we have:

```
			ginkgo.By("Await for jobs completion", func() {
				util.ExpectWorkloadToFinishWithTimeout(ctx, k8sClient, wlLookupKey, util.LongTimeout)
			})

			ginkgo.By("Await workload finalized", func() {
				util.ExpectWorkloadsFinalizedOrGone(ctx, k8sClient, wlLookupKey)
			})
```

This means we first wait for the workload to finish, and only then check whether it’s finalized or already gone. I think this approach is correct and should work as expected.

PR #10775, opened 2026-04-24
Fix Dockerfile cross-compilation for multi-arch builds

carterpewpew:
#### What type of PR is this?

/kind bug
/area dashboard

The runtime `FROM` stage in three Dockerfiles (`cmd/importer/Dockerfile`, `cmd/experimental/kueue-populator/Dockerfile`, and `cmd/kueueviz/backend/Dockerfile`) uses `--platform=${BUILDPLATFORM}`, which pins the base image to the build host's architecture instead of the target's. Additionally, `cmd/kueueviz/backend/Dockerfile` was missing `ARG TARGETARCH` and `GOARCH=${TARGETARCH}`, so `go build` would produce a binary for the build host architecture rather than the target.

Together, these issues cause cross-platform builds (e.g. building `linux/arm64` images on an `amd64` host) to produce containers that fail at runtime with `exec format error`.

This PR:
- Removes `--platform=${BUILDPLATFORM}` from the runtime `FROM` stage in all three Dockerfiles, so Docker defaults to `${TARGETPLATFORM}` (matching the convention in the root `Dockerfile`).
- Adds `ARG TARGETARCH` and passes `GOOS=linux GOARCH=${TARGETARCH}` to `go build` in `cmd/kueueviz/backend/Dockerfile`.

None

The root `Dockerfile` already follows the correct pattern and documents it explicitly:

```
# final image, implicitly --platform=${TARGETPLATFORM}
FROM ${BASE_IMAGE}
```

The importer and kueue-populator Dockerfiles already had the `GOARCH=${TARGETARCH}` cross-compilation in their build steps, but the runtime stage was pulling the wrong platform's base image. The kueueviz backend was missing both fixes.

```release-note
Fixed multi-arch image builds for importer, kueue-populator, and kueueviz backend images so runtime images
and binaries are built for the target platform, preventing wrong-architecture containers and exec format error
for non-amd64 target platforms, such as arm64, ppc64le, and s390x.
```

### DIFF
--- a/cmd/experimental/kueue-populator/Dockerfile
+++ b/cmd/experimental/kueue-populator/Dockerfile
@@ -19,7 +19,7 @@ WORKDIR /workspace/cmd/experimental/kueue-populator
 RUN go mod download
 RUN make build GO_BUILD_ENV='CGO_ENABLED=${CGO_ENABLED} GOOS=linux GOARCH=${TARGETARCH}'
 
-FROM --platform=${BUILDPLATFORM} ${BASE_IMAGE}
+FROM ${BASE_IMAGE}
 WORKDIR /
 COPY --from=builder /workspace/cmd/experimental/kueue-populator/bin/manager .
 USER 65532:65532
--- a/cmd/importer/Dockerfile
+++ b/cmd/importer/Dockerfile
@@ -14,7 +14,7 @@ COPY . .
 # Build
 RUN make importer-build GO_BUILD_ENV='CGO_ENABLED=${CGO_ENABLED} GOOS=linux GOARCH=${TARGETARCH}'
 
-FROM --platform=${BUILDPLATFORM} ${BASE_IMAGE}
+FROM ${BASE_IMAGE}
 WORKDIR /
 COPY --from=builder /workspace/bin/importer .
 USER 65532:65532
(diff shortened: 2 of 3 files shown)


mimowo:
/ok-to-test
cc @mbobrovskyi ptal

mbobrovskyi:
Thank you!

/lgtm

mimowo:
@carterpewpew @mbobrovskyi  should we set a release note and cherrypick? It seems like this bug prevents using the runtime images on some platforms, wdyt?

carterpewpew:
> @carterpewpew @mbobrovskyi should we set a release note and cherrypick? It seems like this bug prevents using the runtime images on some platforms, wdyt?

Thank you @mimowo @mbobrovskyi  Sounds good to me, I'll go ahead and add a release note.

mimowo:
/release-note-edit
```release-note
Fixed multi-arch image builds for importer, kueue-populator, and kueueviz backend images so runtime images
and binaries are built for the target platform, preventing wrong-architecture containers and exec format error
for non-amd64 target platforms, such as arm64, ppc64le, and s390x.
```

mimowo:
Thank you 👍 
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16

Issue #10168, opened 2026-03-29
Add support for kueue simulator


kimminw00:
**What would you like to be added**:

Add a lightweight simulator for Kueue. kube-scheduler integration can be considered later as an optional extension, but it should not block the initial simulator.

**Why is this needed**:

The main goal of this simulator is to **simulate a large number of workloads virtually and cheaply without building a real Kubernetes cluster(ex. kwok)**.

Today, it is expensive and slow to validate Kueue behavior with real workloads because it usually requires setting up a real cluster and running actual workloads.

A lightweight simulator would make it much easier to:

- test Kueue policy and admission behavior
- compare queueing / preemption / fair sharing configurations
- reproduce workload waiting / admission / eviction scenarios
- run large-scale what-if analysis with low cost
- do regression testing for Kueue changes

For the MVP, it is more important to model **Kueue decision-making** accurately than to model full Kubernetes end-to-end execution.

**Completion requirements**:

This enhancement requires the following artifacts:

- [x] Design doc
- [ ] API change
- [x] Docs update

The artifacts should be linked in subsequent comments.

wzshiming:
https://github.com/kubernetes-sigs/kwok/pull/1513
https://github.com/kubernetes-sigs/kwok/pull/1516

kueue/kueueviz is already integrated in kwokctl. Its use is like `kwokctl create cluster --runtime=docker --enable kueue`, it will be included in the next release.

zhitaoli:
+100 for this idea

mimowo:
FYI we have a PR which introduces higher scale for performance testing which may also interest folks who are looking into kwok simulations: https://github.com/kubernetes-sigs/kueue/pull/10780

PR #10780, opened 2026-04-25
add more performance tests for throughput and load

### LINKED ISSUE #10773: Expand/Add On Performance Tests for Load & Throughput
Load and throughput tests with much higher scale than the current tests provide

**Why is this needed**:
We run Kueue at a much larger scale than the performance tests cover and on our side we have to run load tests manually for every upgrade we perform, ideally this would be something covered in the upstream repo

**Completion requirements**:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

baoalvin1:
#### What type of PR is this?
/kind feature
/area testing

Adds more large scale performance tests to Kueue simulating hundreds of CQs and tens of thousands of workloads

Fixes https://github.com/kubernetes-sigs/kueue/issues/10773

```release-note
NONE
```

### DIFF
--- a/Makefile-test.mk
+++ b/Makefile-test.mk
@@ -472,7 +472,7 @@ test-performance-scheduler-once: gotestsum run-performance-scheduler
 PERFORMANCE_RETRY_COUNT?=2
 .PHONY: test-performance-scheduler
 test-performance-scheduler:
-	ARTIFACTS="$(ARTIFACTS)/$@" ./hack/testing/performance-test.sh $(PERFORMANCE_RETRY_COUNT) test-large-scale-performance-scheduler-once
+	ARTIFACTS="$(ARTIFACTS)/$@" ./hack/testing/performance-test.sh $(PERFORMANCE_RETRY_COUNT) test-performance-scheduler-once
 
 .PHONY: run-performance-scheduler-in-cluster
 run-performance-scheduler-in-cluster: envtest performance-scheduler-runner
--- a/Makefile-test.mk
+++ b/Makefile-test.mk
@@ -472,7 +472,7 @@ test-performance-scheduler-once: gotestsum run-performance-scheduler
 PERFORMANCE_RETRY_COUNT?=2
 .PHONY: test-performance-scheduler
 test-performance-scheduler:
-	ARTIFACTS="$(ARTIFACTS)/$@" ./hack/testing/performance-test.sh $(PERFORMANCE_RETRY_COUNT) test-performance-scheduler-once
+	ARTIFACTS="$(ARTIFACTS)/$@" ./hack/testing/performance-test.sh $(PERFORMANCE_RETRY_COUNT) test-large-scale-performance-scheduler-once
 
 .PHONY: run-performance-scheduler-in-cluster
 run-performance-scheduler-in-cluster: envtest performance-scheduler-runner
--- a/Makefile-test.mk
+++ b/Makefile-test.mk
@@ -520,6 +520,41 @@ run-tas-performance-scheduler-in-cluster: envtest performance-scheduler-runner
 		--enableTAS=true \
 		--qps=1000 --burst=2000 --timeout=25m $(SCALABILITY_SCRAPE_ARGS)
 
+##@ Scheduler Performance Testing - Large Scale
+
+SCALABILITY_LARGE_SCALE_GENERATOR_CONFIG ?= $(PROJECT_DIR)/test/performance/scheduler/configs/large-scale/generator.yaml
+SCALABILITY_LARGE_SCALE_RANGE_FILE ?= $(PROJECT_DIR)/test/performance/scheduler/configs/large-scale/rangespec.yaml
+
+.PHONY: run-large-scale-performance-scheduler
+run-large-scale-performance-scheduler: envtest performance-scheduler-runner minimalkueue
+	mkdir -p "$(ARTIFACTS)/$@"
+	KUBEBUILDER_ASSETS="$(KUBEBUILDER_ASSETS)" \
+	$(SCALABILITY_RUNNER) \
+		--o "$(ARTIFACTS)/$@" \
+		--crds=$(PROJECT_DIR)/config/components/crd/bases \
+		--generatorConfig=$(SCALABILITY_LARGE_SCALE_GENERATOR_CONFIG) \
+		--minimalKueue=$(MINIMALKUEUE_RUNNER) --timeout=30m $(SCALABILITY_EXTRA_ARGS) $(SCALABILITY_SCRAPE_ARGS)
+
+.PHONY: test-large-scale-performance-scheduler-once
+test-large-scale-performance-scheduler-once: gotestsum run-large-scale-performance-scheduler
+	$(GOTESTSUM) --junitfile $(ARTIFACTS)/junit.xml -- $(GO_TEST_FLAGS) ./test/performance/scheduler/checker  \
+		--summary=$(ARTIFACTS)/run-large-scale-performance-scheduler/summary.yaml \
+		--cmdStats=$(ARTIFACTS)/run-large-scale-performance-scheduler/minimalkueue.stats.yaml \
+		--range=$(SCALABILITY_LARGE_SCALE_RANGE_FILE)
+
+.PHONY: test-large-scale-performance-scheduler
+test-large-scale-performance-scheduler:
+	ARTIFACTS="$(ARTIFACTS)/$@" ./hack/testing/performance-test.sh $(PERFORMANCE_RETRY_COUNT) test-large-scale-performance-scheduler-once
+
+.PHONY: run-large-scale-performance-scheduler-in-cluster
+run-large-scale-performance-scheduler-in-cluster: envtest performance-scheduler-runner
+	mkdir -p "$(ARTIFACTS)/$@"
+	KUBEBUILDER_ASSETS="$(KUBEBUILDER_ASSETS)" \
+	$(SCALABILITY_RUNNER) \
+		--o "$(ARTIFACTS)/$@" \
+		--generatorConfig=$(SCALABILITY_LARGE_SCALE_GENERATOR_CONFIG) \
+		--qps=1000 --burst=2000 --timeout=30m $(SCALABILITY_SCRAPE_ARGS)
+
 .PHONY: ginkgo-top
 ginkgo-top:
 	cd $(TOOLS_DIR) && go mod download && \
--- a/Makefile-test.mk
+++ b/Makefile-test.mk
@@ -533,7 +533,8 @@ run-large-scale-performance-scheduler: envtest performance-scheduler-runner mini
 		--o "$(ARTIFACTS)/$@" \
 		--crds=$(PROJECT_DIR)/config/components/crd/bases \
 		--generatorConfig=$(SCALABILITY_LARGE_SCALE_GENERATOR_CONFIG) \
-		--minimalKueue=$(MINIMALKUEUE_RUNNER) --timeout=30m $(SCALABILITY_EXTRA_ARGS) $(SCALABILITY_SCRAPE_ARGS)
+		--minimalKueue=$(MINIMALKUEUE_RUNNER) \
+		--timeout=30m $(SCALABILITY_EXTRA_ARGS) $(SCALABILITY_SCRAPE_ARGS)
 
 .PHONY: test-large-scale-performance-scheduler-once
 test-large-scale-performance-scheduler-once: gotestsum run-large-scale-performance-scheduler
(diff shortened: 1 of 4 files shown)


mimowo:
I added manually the release note as NONE to make the robot happy

mimowo:
/ok-to-test
Thank you for the test suite 👍 

I have two questions to better explore the community needs:
1. Do you see this scale as sufficient for you or just a compromise for the CI time? 
2. Do you see some use cases for closer to e2e tests using kwok, as proposed here: https://github.com/kubernetes-sigs/kueue/issues/10168?

baoalvin1:
@mimowo 

1. I think the scale is sufficient but I would be curious to see how far we could push it but I do want to be mindful of CI times
2. The overlap I see from that issue is `test Kueue policy and admission behavior` and `do regression testing for Kueue changes`. For us I don't think the other 3 items are as crucial since we have our own ways of testing these and we'd want it to be closer to a "real" env for us.

mimowo:
> I think the scale is sufficient but I would be curious to see how far we could push it but I do want to be mindful of CI times

Awesome. CI times are hard to predict, the best is to just try. I would just temporarily swap the make file goal called by our CI which is specified here: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/kueue/kueue-presubmits-main.yaml#L738C17-L738C43

Then we can see how long it takes before the merge, and potentially adjust some parameters.

EDIT: I think this could be achieved by swapping in this [line](https://github.com/kubernetes-sigs/kueue/pull/10780/changes#diff-23f8123955f3fa9646deaef30d6463b18df3abe77e5a0dfc5ffedea069ce193dR475) `test-performance-scheduler-once` with `test-large-scale-performance-scheduler-once`

Makefile-test.mk
```diff
@@ -472,7 +472,7 @@ test-performance-scheduler-once: gotestsum run-performance-scheduler
 PERFORMANCE_RETRY_COUNT?=2
 .PHONY: test-performance-scheduler
 test-performance-scheduler:
-	ARTIFACTS="$(ARTIFACTS)/$@" ./hack/testing/performance-test.sh $(PERFORMANCE_RETRY_COUNT) test-performance-scheduler-once
+	ARTIFACTS="$(ARTIFACTS)/$@" ./hack/testing/performance-test.sh $(PERFORMANCE_RETRY_COUNT) test-large-scale-performance-scheduler-once
```
baoalvin1: @mimowo I changed this, the link from your comment didn't seem to load anything

baoalvin1: from the logs
```
ARTIFACTS="/logs/artifacts/test-performance-scheduler" ./hack/testing/performance-test.sh 2 test-large-scale-performance-scheduler-once
```

mimowo: /hold 
to avoid merging while we are performing the experiments. Feel free to unhold when we have the experiment data and the changes are reverted

baoalvin1: without the max-requests-inflight modification it takes 45 minutes and I don't think the results get properly recorded, it was trying to do a 2nd attempt

mimowo: I suppose this is because the test exceeded the timeout of 15min. We may need to increase the timeout, but:
1. I wouldn't like to cross the time it takes for any other suite which is currently around 25min
2. running tests locally which take long is also inconvenient

So I would say 20-24min should be the timeout still.  To fit within the timeout we may need to:
a) compromise on some dimentions of the scale
b) increase the machines in test-infra, currently they have 6CPUs, I would be ok with 8CPU
c) increase the infligts as proposed

To guide the decision about b) please also check the monitoring: https://monitoring-eks.prow.k8s.io/d/96Q8oOOZk/builds?orgId=1&from=1777307966363&to=1777308587547&var-org=kubernetes-sigs&var-repo=kueue&var-job=periodic-kueue-test-scheduling-perf-main&var-build=All

kshalot: If we have to compromise on the scale too much in order to preserve velocity, we can consider using this large scale test in periodics only. I think right now they run every 12h, but we could run this one once every 24h if it's very heavy.

One disadvantage is that it would make it harder to track which exact change caused a regression, but it shouldn't be that hard to pin down in practice.

test/performance/scheduler/runner/main.go
```diff
@@ -127,8 +127,8 @@ func main() {
 		// resulting in "http2: client connection lost" errors when pending requests queue up
 		// and connections time out.
 		testEnv.ControlPlane.GetAPIServer().Configure().
-			Append("max-requests-inflight", "800").
-			Append("max-mutating-requests-inflight", "400")
+			Append("max-requests-inflight", "2000").
+			Append("max-mutating-requests-inflight", "1000")
```
kshalot: At first glance - won't we potentially have even 2000 mutating requests inflight in the proposed scenario? Since we'll start a workload generator for each CQ:
https://github.com/kubernetes-sigs/kueue/blob/32c4babf8f793c3e19415e17fffabc2155d5d037/test/performance/scheduler/runner/generator/generator.go#L227-L229

I'm worried we'll choke on workload creation instead of scheduling. There are some "http2: client connection lost" errors in the test log

baoalvin1: yeah we would, can adjust

Makefile-test.mk
baoalvin1: looks like we're OOMing after I bumped the inflight requests, is there a way to adjust it just to validate this scenario? i.e. would I have to modify https://github.com/kubernetes/test-infra/blob/9666851ae895800be1424b19f2399ab376c39424/config/jobs/kubernetes-sigs/kueue/kueue-presubmits-main.yaml#L742-L748

baoalvin1: I'm going to adjust down to 1k CQs and 45k workloads to see if it finishes

baoalvin1: even on this lower scale it still OOMs @mimowo any way we could get more memory for this?

mimowo: I think so, but before that, let's do the sanity checks:
1. Could you provide a link to the artifacts indicating the oom kill? Is it api-server that is oom'ed or kueue?
2. Could you bump locally and run the build with SCALABILITY_MEM_PROFILE, and share the mem profile key stats / image - just so we verify there are no nasty memory leaks which could be covered by more mem.

baoalvin1: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10780/pull-kueue-test-scheduling-perf-main/2051336321665863680

baoalvin1: I'm not really sure from `Job pod was OOM killed by the cluster.` is there a better way to tell?

mimowo: Oh I see, in this case it seems the API server broke, because Kueue process continued logging like this:

```
2026-05-04T16:31:47.248348062Z	ERROR	controller/controller.go:495	Reconciler error	{"controller": "workload", "controllerGroup": "kueue.x-k8s.io", "controllerKind": "Workload", "Workload": {"name":"small-0-2-0","namespace":"cq-0-7-cohort-33"}, "namespace": "cq-0-7-cohort-33", "name": "small-0-2-0", "reconcileID": "574256da-0ef9-4e9e-adfa-76e742eae61d", "error": "Patch \"https://127.0.0.1:44423/apis/kueue.x-k8s.io/v1beta2/namespaces/cq-0-7-cohort-33/workloads/small-0-2-0/status?fieldManager=kueue-job-controller-Finished&force=true\": http2: client connection lost"}
```
So it looks like API server was down and was not responding. In that case it really suggests not enough memory at the api-server side rather than Kueue leak.

mimowo: Aside from running locally, could you try to get the max scale supported by the current mem (9Gi), so that we can roughly know where we are, and assuming linear scaling we can tell how much % we need to bump the memory.

baoalvin1: Screenshot 2026-05-04 at 12 13 30 PM
for the last job I ran, 500 CQs and 25k workloads

mimowo: I see, so at the scale of 25k workloads we are pretty much capped by 9Gi, and we want roughly 2x that. CPU is probably not a blocker, the test will just take longer. I looked at test-infra, and there is a number of jobs using 16Gi, particularly scale tests here:  https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/kops/kops-presubmits-scale.yaml so I think we could follow this pattern 6CPU / 16 Gi for scale tests.

 So, I would propose:
 1. introduce the new suite at the level of 400CQ, 20k to leave some room and make sure we are stable
 2. add test infra job for the new test using 6CPU / 16 Gi (periodic and presubmits)
 3. test on another branch in Kueue how much we can increase the scale to still fit within 20min of runtime for the test, it might be 30k, maybe 45k, we will see.

baoalvin1: just so I understand, the next steps would be
1. merge this pr but without running it for an existing test (i.e. don't replace the existing perf test)
2. add test infra job to be able to run the merged code
3. create new branch and target the new test infra job created

mimowo: Yes, but we do in the lockstep:
1. we merge to Kueue but only at the scale that we know will pass (safely) on the regular scaling of the machines (as for other perf tests, ie. 7 cpu)
2. we add the test infra job at the increased scale to run the code
3. we add another PR to Kueue to increase the scale

In the meanwhile let me ask at the k8s-infra slack channel if there are any recommendations about the machine scaling for (2.)

mimowo: Asked here: https://kubernetes.slack.com/archives/C09QZ4DQB/p1778053793108319

mimowo: Ok, I know more, based on the discussion here: https://kubernetes.slack.com/archives/CCK68P2Q2/p1778053935373549

So at (2.) I would suggest using 7cpu/26Gi to basically cap a single machine,wdyt?

baoalvin1: yeah that sounds good to me, I can work on these!

baoalvin1: @mimowo opened https://github.com/kubernetes/test-infra/pull/36969

baoalvin1: Screenshot 2026-05-06 at 7 49 38 PM
last run

mimowo: Hm so the run took around 25min. That is on the edge of what I consider acceptable, and also I'm not sure from a single sample if this was the lucky fast or unlucky long.

To be on the safe side let's reduce the scale to 15k before merge, so that we can be closer to 20min, or ideally below.

I know we are going to increase the scale once we have more mem, but I want to make sure that it does not impact negatively daily work by other developers in the meanwhile. Alose the final scale is unclear at the moment.

mimowo: Another question, do you know why the CPU usage drops to 0 basically in the middle of the run?

baoalvin1: @mimowo I believe it's because the test isn't calibrated for the avg cq usage so it fails the requirements, I think it hovers around 40% so it takes 25 minutes since it does 2 attempts, I can fix it to track the right cq utilization

baoalvin1: so I think 2 runs takes 25 minutes so the average will probably be 12 minutes, I can fix the assertions

baoalvin1:
```
=== Failed
=== FAIL: test/performance/scheduler/checker TestScalability/ClusterQueueClasses/cq (0.00s)
    checker_test.go:98: Usage 40.50% is less then expected 55.00%
=== FAIL: test/performance/scheduler/checker TestScalability/ClusterQueueClasses (0.00s)
=== FAIL: test/performance/scheduler/checker TestScalability (0.01s)
DONE 8 tests, 3 failures in 0.301s
make[1]: *** [Makefile-test.mk:541: test-large-scale-performance-scheduler-once] Error 1
make[1]: Leaving directory '/home/prow/go/src/sigs.k8s.io/kueue'
Try 2
```

logs from test, it runs a 2nd try

mimowo: I see, cool. So instead of lowering the scale, let's make sure it is passing, and performs only a single run

baoalvin1: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10780/pull-kueue-test-scheduling-perf-main/2052294546481483776 passes and only does a single run

mimowo: Awesome, let's revert then https://github.com/kubernetes-sigs/kueue/pull/10780/changes#diff-23f8123955f3fa9646deaef30d6463b18df3abe77e5a0dfc5ffedea069ce193dR475  to prepare to merge

baoalvin1: done!

mimowo:
Thank you!
/lgtm
/approve

PR #10782, opened 2026-04-26
add unit tests for indexer

gyliu513:
#### What type of PR is this?

/kind testing

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/controller/core/indexer/indexer_test.go
+++ b/pkg/controller/core/indexer/indexer_test.go
@@ -1,0 +1,650 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package indexer
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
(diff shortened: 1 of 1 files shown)


mimowo:
cc @mykysha @mszadkow ptal

pkg/controller/core/indexer/indexer_test.go
```diff
@@ -0,0 +1,650 @@
...
+	}
+}
+
+func TestOwnerReferenceIndexFieldMatcher(t *testing.T) {
+	got := OwnerReferenceIndexFieldMatcher(batchJobGVK, "my-job")
+
+	want := client.MatchingFields{".metadata.ownerReferences[batch.Job]": "my-job"}
+	if diff := cmp.Diff(want, got); diff != "" {
+		t.Errorf("MatchingFields mismatch (-want +got):\n%s", diff)
+	}
```
mszadkow, line 101: nit:
could be in the table-driven form, even though it's a single test

mszadkow:
one nit, other than that lgtm
/lgtm

pkg/controller/core/indexer/indexer_test.go
```diff
@@ -0,0 +1,650 @@
...
+
+	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+)
+
+// helpers to build test objects without importing any package that transitively
+// imports this (indexer) package, which would create an import cycle.
+
+func makeWorkload(name, ns string) *kueue.Workload {
+	return &kueue.Workload{ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: ns}}
+}
```
mbobrovskyi, line 38: Why this helper? We already have WorkloadWrapper.

mbobrovskyi: Ah, I see – it’s due to an import cycle.

pkg/controller/core/indexer/indexer_test.go
mbobrovskyi: I'm actually not sure about this. If we have only one case I think it's ok to have as it is or add more cases.

mbobrovskyi:
Thank you!

/approve

PR #10598, opened 2026-04-20
TAS: split topology assignment state

ShaanveerS:
#### What type of PR is this?
/kind cleanup
/area tas

#### What this PR does / why we need it:
Split `findTopologyAssignmentState` into `topologyAssignmentPodRequirements` and `topologyAssignmentTASParameters`, so pod requirement inputs are only used up to `fillInCounts` and the placement algos only receive the TAS placement state.

#### Which issue(s) this PR fixes:
Followup #10537

#### Does this PR introduce a user-facing change?
```release-note
NONE
```

PR #10783, opened 2026-04-26
fix: Fix TAS stateWithLeader aggregation for grouped podsets

### LINKED ISSUE #10778: [TAS] stateWithLeader over-estimation from empty children causes infinite scheduling loop
**What happened**:

TAS workloads with PodSetGrouping enter an infinite scheduling loop when a spine has exactly 1 free node and the rest are reserved. The scheduler repeatedly fails CRD validation (`topologyAssignment.slices: Required value`) and retries with no backoff.

The bug triggers when a spine has exactly 1 free node and the leader pod's resource request is large enough that placing it on the free node leaves insufficient capacity for any workers. The over-estimated `stateWithLeader` equals `state` (worker capacity without leader), so the bug causes failure when the free node can fit all the workers WITHOUT the leader but CANNOT fit them WITH the leader. In other words: when the leader consumes enough of the node that workers no longer fit.

This is not limited to 1 leader + 1 worker. It also triggers with mixed sizes, e.g., master requesting 2500m CPU + 2 workers requesting 1250m each: the free node fits 2 workers (state=2) but after the leader takes 2500m only 300m remains (stateWithLeader=0). The buggy aggregation inflates stateWithLeader to 2, matching sliceCount=2, so the spine is selected and fails.

Filing separately from #7446 because this is a specific, reproducible trigger with a targeted fix, while #7446 tracks the broader phase 1/phase 2 inconsistency.

The spine should be skipped. The workload should be placed on a larger spine with sufficient capacity.

Topology: 3 levels (region → spine → hostname)
- spine-x: 4 nodes, spine-y: 2 nodes
- Each node: 4 CPUs, ~1.2 CPU used by system pods (~2800m allocatable)
- Required topology level: spine
- Each pod requests 2500m CPU so that at most 1 pod fits per node

**Test 1 — Bug triggers (equal sizes):**
1. Submit a filler job (2500m CPU) with a `nodeSelector` targeting spine-y, leaving 1 free node
2. Submit a PyTorchJob with Master (1 replica) + Worker (1 replica), same `podSetGroupName`, each requesting 2500m CPU
3. Result: Stuck. Logs show `unexpected remainingCount` / `code assumptions violated` and `topologyAssignment.slices: Required value` in a tight loop

**Test 2 — Bug triggers (mixed sizes):**
1. Submit a filler job (2500m CPU) with a `nodeSelector` targeting spine-y, leaving 1 free node
2. Submit a PyTorchJob with Master (1 replica, 2500m CPU) + Worker (2 replicas, 500m CPU each), same `podSetGroupName`
3. Result: Same failure. Master consumes the node, leaving 295m — not enough for any 500m worker.

**Test 3 — No bug (total fits on one node):**
1. Same spine-y setup (1 free node)
2. Submit a PyTorchJob with Master (1 replica, 1000m CPU) + Worker (1 replica, 1000m CPU), same `podSetGroupName`
3. Result: Admitted. Both fit on the single free node (total 2000m < 2795m available).

**Trigger conditions** (validated across 11 test cases, all on spine-y with 1 free node at 2795m available):

The bug triggers when ALL THREE hold for the single free node:
1. Leader fits on the node: `leaderRequest <= nodeCapacity`
2. All workers fit on the node (without leader): `workerCount * workerRequest <= nodeCapacity`
3. Leader + all workers DON'T fit together: `leaderRequest + workerCount * workerRequest > nodeCapacity`

| # | Master | Workers | Worker CPU | Total | Cond 1 | Cond 2 | Cond 3 | All met? | Result |
|---|--------|---------|------------|-------|--------|--------|--------|----------|--------|
| 1 | 2500m | 1 | 1250m | 3750m | ✓ | ✓ | ✓ | Yes | STUCK |
| 2 | 2500m | 2 | 500m | 3500m | ✓ | ✓ | ✓ | Yes | STUCK |
| 3 | 2500m | 1 | 2500m | 5000m | ✓ | ✓ | ✓ | Yes | STUCK |
| 4 | 1500m | 2 | 1000m | 3500m | ✓ | ✓ | ✓ | Yes | STUCK |
| 5 | 2000m | 3 | 500m | 3500m | ✓ | ✓ | ✓ | Yes | STUCK |
| 6 | 1000m | 1 | 1000m | 2000m | ✓ | ✓ | ✗ | No | ADMITTED |
| 7 | 2500m | 3 | 1250m | 6250m | ✓ | ✗ | ✓ | No | ADMITTED |
| 8 | 2500m | 2 | 2500m | 7500m | ✓ | ✗ | ✓ | No | ADMITTED |
| 9 | 2500m | 1 | 100m | 2600m | ✓ | ✓ | ✗ | No | ADMITTED |
| 10 | 1250m | 1 | 1250m | 2500m | ✓ | ✓ | ✗ | No | ADMITTED |
| 11 | 1300m | 1 | 1300m | 2600m | ✓ | ✓ | ✗ | No | ADMITTED |

### Root cause

`fillInCountsHelper` in `pkg/cache/scheduler/tas_flavor_snapshot.go` ([buggy line in v0.17.1](https://github.com/kubernetes-sigs/kueue/blob/v0.17.1/pkg/cache/scheduler/tas_flavor_snapshot.go#L1680)):

```go
minStateWithLeaderDifference = min(childState-childStateWithLeader, minStateWithLeaderDifference)
```

This iterates ALL children including those with `leaderState=0` (reserved nodes that cannot host a leader). Children with `state=0, stateWithLeader=0` contribute `penalty = 0 - 0 = 0`, which becomes the minimum. The parent spine's `stateWithLeader` is computed as `childrenCapacity - 0 = childrenCapacity`, making it equal to `state`. This makes the scheduler believe the spine can fit both the leader and the worker, when in reality only one pod can be placed.

Worked example for spine-y (1 free node, 1 reserved node):
```
Free node:     state=1, stateWithLeader=0, leaderState=1, penalty=1
Reserved node: state=0, stateWithLeader=0, leaderState=0, penalty=0

spine-y: stateWithLeader = 1 - min(1,0) = 1 - 0 = 1   ← WRONG, should be 0
```

Best-fit selects spine-y (smallest fitting domain). `updateCountsToMinimumGeneric` at the hostname level places the leader on the only free node, exhausts capacity, cannot place the worker, and returns `nil`. Both assignments get empty slices via `buildAssignment(nil)` → `singleCompactSliceEncoding` → `Slices: []` → dropped by `omitempty` → CRD rejects.

### Proposed fix

Only consider leader-capable children in the penalty computation:

```go
// Current (buggy):
minStateWithLeaderDifference = min(childState-childStateWithLeader, minStateWithLeaderDifference)
minSliceStateWithLeaderDifference = min(child.sliceState-child.sliceStateWithLeader, minSliceStateWithLeaderDifference)

// Fix:
if child.leaderState > 0 {
    minStateWithLeaderDifference = min(childState-childStateWithLeader, minStateWithLeaderDifference)
    minSliceStateWithLeaderDifference = min(child.sliceState-child.sliceStateWithLeader, minSliceStateWithLeaderDifference)
}

// Guard when no child can host a leader (minStateWithLeaderDifference stays MaxInt32):
if minStateWithLeaderDifference == math.MaxInt32 {
    domain.stateWithLeader = 0
} else {
    domain.stateWithLeader = childrenCapacity - minStateWithLeaderDifference
}
// Same for sliceStateWithLeader
```

This fixes the empty-children trigger but not the general #7446 inconsistency (phase 2 greedy assignment may pick a different child than phase 1 assumed).

This fix also incidentally corrects `stateWithLeader` for workloads with `count=0` podSets in a group, though #10478 addresses that case more directly.

### Secondary issue

`updateCountsToMinimumGeneric` returns `nil` on failure without propagating a failure reason. This causes malformed `TopologyAssignment` instead of clean rejection. Propagating the failure would at least prevent the infinite retry loop. This should be addressed as a follow-up.

Bug exists in all versions with PodSetGrouping — confirmed by code inspection in v0.15.4, v0.17.1, and mainline.

**NOTE**: Our production is running v0.15.4 and this issue blocks the queue completely. However I was not able to reproduce this locally. However, locally I am running a much smaller cluster, 10s of nodes and nodes vs 1000s of nodes and workloads. Not sure if that can lead to starvation and kueue never reaching to the good job which fits.

Related: #7446

**Environment**:
- Kubernetes version: v1.32
- Kueue version: v0.15.4 (confirmed in v0.17.1 and mainline by code inspection)
- Cloud provider or hardware configuration: AWS EKS
- Install tools: Helm via Argo

[tas-repro-setup.yaml](https://github.com/user-attachments/files/27068345/tas-repro-setup.yaml)

yuluo-yx:
## What this PR does

  For grouped podSets with a leader, parent topology domains were computing the `stateWithLeader` penalty across all children, including children that cannot host the leader. Empty or fully occupied children have `state=0`,
  `stateWithLeader=0`, and `leaderState=0`, so they contributed a zero penalty. This could make the parent domain report `stateWithLeader == state`, even when the only leader-capable child could not also fit the workers after placing
  the leader.

  In practice, this caused the scheduler to select a smaller topology domain that looked like it could fit the grouped workload during phase 1, then fail during assignment generation and produce an empty/malformed
  `TopologyAssignment`.

  This PR changes the parent aggregation logic so that, when a leader is required, only leader-capable children contribute to the `stateWithLeader` and `sliceStateWithLeader` penalty calculation. If no child can host the leader, the
  parent with-leader capacity is set to zero.

  The balanced placement pruning path was updated to preserve the same leader-aware aggregation behavior when recomputing domain counts.

  ## Tests

  Added regression coverage in `TestFindTopologyAssignments` for:
  - equal-size grouped leader/worker podSets where the small domain must be skipped
  - mixed-size grouped podSets where workers fit without the leader but not after leader placement
  - a positive case where leader and workers do fit together on the tight domain and should still be assigned there

  Fixes #10778.
  
  ```release-note
TAS: Fix handling of PodSet groups which could lead in some scenarios to empty topologyAssignment.
```

### DIFF
--- a/pkg/cache/scheduler/tas_flavor_snapshot.go
+++ b/pkg/cache/scheduler/tas_flavor_snapshot.go
@@ -1699,16 +1699,14 @@ func (s *TASFlavorSnapshot) fillInCountsHelper(domain *domain, sliceSize int32, 
 		leaderState = max(child.leaderState, leaderState)
 	}
 	domain.state = childrenCapacity
+	sliceStateWithLeader := int32(0)
 	if minStateWithLeaderDifference == math.MaxInt32 {
 		domain.stateWithLeader = 0
 	} else {
 		domain.stateWithLeader = childrenCapacity - minStateWithLeaderDifference
-	}
-	domain.leaderState = leaderState
-	sliceStateWithLeader := int32(0)
-	if minSliceStateWithLeaderDifference != math.MaxInt32 {
 		sliceStateWithLeader = sliceCapacity - minSliceStateWithLeaderDifference
 	}
+	domain.leaderState = leaderState
 
 	if level == sliceLevelIdx {
 		// initialize the sliceState for the requested slice level.
--- a/pkg/cache/scheduler/tas_flavor_snapshot.go
+++ b/pkg/cache/scheduler/tas_flavor_snapshot.go
@@ -1668,7 +1668,7 @@ func (s *TASFlavorSnapshot) fillInCountsHelper(domain *domain, sliceSize int32, 
 	// logic for a parent
 	childrenCapacity := int32(0)
 	sliceCapacity := int32(0)
-	hasWithLeaderCapacityContributor = false
+	hasWithLeaderCapacityContributor := false
 	minStateWithLeaderDifference := int32(math.MaxInt32)
 	minSliceStateWithLeaderDifference := int32(math.MaxInt32)
 	leaderState := int32(0)
--- a/pkg/cache/scheduler/tas_flavor_snapshot.go
+++ b/pkg/cache/scheduler/tas_flavor_snapshot.go
@@ -1668,7 +1668,7 @@ func (s *TASFlavorSnapshot) fillInCountsHelper(domain *domain, sliceSize int32, 
 	// logic for a parent
 	childrenCapacity := int32(0)
 	sliceCapacity := int32(0)
-
+	hasWithLeaderCapacityContributor = false
 	minStateWithLeaderDifference := int32(math.MaxInt32)
 	minSliceStateWithLeaderDifference := int32(math.MaxInt32)
 	leaderState := int32(0)
@@ -1693,6 +1693,7 @@ func (s *TASFlavorSnapshot) fillInCountsHelper(domain *domain, sliceSize int32, 
 		childrenCapacity += childState
 		sliceCapacity += child.sliceState
 		if !leaderRequired || child.leaderState > 0 {
+			hasWithLeaderCapacityContributor = true
 			minStateWithLeaderDifference = min(childState-childStateWithLeader, minStateWithLeaderDifference)
 			minSliceStateWithLeaderDifference = min(child.sliceState-child.sliceStateWithLeader, minSliceStateWithLeaderDifference)
 		}
@@ -1700,11 +1701,11 @@ func (s *TASFlavorSnapshot) fillInCountsHelper(domain *domain, sliceSize int32, 
 	}
 	domain.state = childrenCapacity
 	sliceStateWithLeader := int32(0)
-	if minStateWithLeaderDifference == math.MaxInt32 {
-		domain.stateWithLeader = 0
-	} else {
+	if hasWithLeaderCapacityContributor {
 		domain.stateWithLeader = childrenCapacity - minStateWithLeaderDifference
 		sliceStateWithLeader = sliceCapacity - minSliceStateWithLeaderDifference
+	} else {
+		domain.stateWithLeader = 0
 	}
 	domain.leaderState = leaderState
 
--- a/pkg/cache/scheduler/tas_flavor_snapshot.go
+++ b/pkg/cache/scheduler/tas_flavor_snapshot.go
@@ -1642,7 +1642,7 @@ func (s *TASFlavorSnapshot) fillInCounts(requirements *topologyAssignmentPodRequ
 		leaf.stateWithLeader = requirements.requests.CountIn(remainingCapacity)
 	}
 	for _, root := range s.roots {
-		s.fillInCountsHelper(root, state.sliceSize, state.sliceLevelIdx, 0, state.sliceSizeAtLevel)
+		s.fillInCountsHelper(root, state.sliceSize, state.sliceLevelIdx, 0, state.sliceSizeAtLevel, state.leaderCount > 0)
 	}
 }
 
@@ -1655,7 +1655,7 @@ func belongsToRequiredDomain(leaf *leafDomain, requiredReplacementDomain utiltas
 	return strings.HasPrefix(string(utiltas.DomainID(leaf.levelValues)), string(requiredReplacementDomain))
 }
 
-func (s *TASFlavorSnapshot) fillInCountsHelper(domain *domain, sliceSize int32, sliceLevelIdx int, level int, sliceSizeAtLevel map[int]int32) {
+func (s *TASFlavorSnapshot) fillInCountsHelper(domain *domain, sliceSize int32, sliceLevelIdx int, level int, sliceSizeAtLevel map[int]int32, leaderRequired bool) {
 	// logic for a leaf
 	if len(domain.children) == 0 {
 		if level == sliceLevelIdx {
@@ -1681,7 +1681,7 @@ func (s *TASFlavorSnapshot) fillInCountsHelper(domain *domain, sliceSize int32, 
 	innerSize, hasInnerConstraint := sliceSizeAtLevel[childLevel]
 
 	for _, child := range domain.children {
-		s.fillInCountsHelper(child, sliceSize, sliceLevelIdx, childLevel, sliceSizeAtLevel)
+		s.fillInCountsHelper(child, sliceSize, sliceLevelIdx, childLevel, sliceSizeAtLevel, leaderRequired)
 
 		childState := child.state
 		childStateWithLeader := child.stateWithLeader
@@ -1692,14 +1692,23 @@ func (s *TASFlavorSnapshot) fillInCountsHelper(domain *domain, sliceSize int32, 
 
 		childrenCapacity += childState
 		sliceCapacity += child.sliceState
-		minStateWithLeaderDifference = min(childState-childStateWithLeader, minStateWithLeaderDifference)
-		minSliceStateWithLeaderDifference = min(child.sliceState-child.sliceStateWithLeader, minSliceStateWithLeaderDifference)
+		if !leaderRequired || child.leaderState > 0 {
+			minStateWithLeaderDifference = min(childState-childStateWithLeader, minStateWithLeaderDifference)
+			minSliceStateWithLeaderDifference = min(child.sliceState-child.sliceStateWithLeader, minSliceStateWithLeaderDifference)
+		}
 		leaderState = max(child.leaderState, leaderState)
 	}
 	domain.state = childrenCapacity
-	domain.stateWithLeader = childrenCapacity - minStateWithLeaderDifference
+	if minStateWithLeaderDifference == math.MaxInt32 {
+		domain.stateWithLeader = 0
+	} else {
+		domain.stateWithLeader = childrenCapacity - minStateWithLeaderDifference
+	}
 	domain.leaderState = leaderState
-	sliceStateWithLeader := sliceCapacity - minSliceStateWithLeaderDifference
+	sliceStateWithLeader := int32(0)
+	if minSliceStateWithLeaderDifference != math.MaxInt32 {
+		sliceStateWithLeader = sliceCapacity - minSliceStateWithLeaderDifference
+	}
 
 	if level == sliceLevelIdx {
 		// initialize the sliceState for the requested slice level.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by mimowo** [comment]
I injected a tentative release note, @yuluo-yx ptal. Next time please keep the release-note in PR template as our automation is using the field.\

> In practice, this caused the scheduler to select a smaller topology domain that looked like it could fit the grouped workload during phase 1, then fail during assignment generation and produce an empty/malformed
TopologyAssignment.

@yuluo-yx question here: what would happen in this scenario: the workload would be admitted with empty topologyAssignment, or admission is rejected?

**Comment by mimowo** [comment]
Let me see if we can have reviews from folks familiar with this code, cc @pajakd @PBundyra

File: pkg/cache/scheduler/tas_flavor_snapshot.go [hunk 1]
```diff
@@ -1692,14 +1692,23 @@ func (s *TASFlavorSnapshot) fillInCountsHelper(domain *domain, sliceSize int32,
...
+		domain.stateWithLeader = 0
+	} else {
+		domain.stateWithLeader = childrenCapacity - minStateWithLeaderDifference
+	}
 	domain.leaderState = leaderState
-	sliceStateWithLeader := sliceCapacity - minSliceStateWithLeaderDifference
+	sliceStateWithLeader := int32(0)
+	if minSliceStateWithLeaderDifference != math.MaxInt32 {
+		sliceStateWithLeader = sliceCapacity - minSliceStateWithLeaderDifference
+	}
```
pajakd: Cannot we merge the two ifs into one? The condition is the same and the line in between is not interfering, right?

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1692,14 +1692,23 @@ func (s *TASFlavorSnapshot) fillInCountsHelper(domain *domain, sliceSize int32,
 
 		childrenCapacity += childState
 		sliceCapacity += child.sliceState
-		minStateWithLeaderDifference = min(childState-childStateWithLeader, minStateWithLeaderDifference)
-		minSliceStateWithLeaderDifference = min(child.sliceState-child.sliceStateWithLeader, minSliceStateWithLeaderDifference)
+		if !leaderRequired || child.leaderState > 0 {
```
pajakd, line 1695: it is counter-intuitive that we enter this code branch if the leader is not required. Why do we have to do it?

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1692,14 +1692,23 @@ func (s *TASFlavorSnapshot) fillInCountsHelper(domain *domain, sliceSize int32,
...
+			minStateWithLeaderDifference = min(childState-childStateWithLeader, minStateWithLeaderDifference)
+			minSliceStateWithLeaderDifference = min(child.sliceState-child.sliceStateWithLeader, minSliceStateWithLeaderDifference)
+		}
 		leaderState = max(child.leaderState, leaderState)
 	}
 	domain.state = childrenCapacity
-	domain.stateWithLeader = childrenCapacity - minStateWithLeaderDifference
+	if minStateWithLeaderDifference == math.MaxInt32 {
```
pajakd: This works now, but looks brittle (what if we change `minStateWithLeaderDifference` type or initial value). I would prefer an explicit flag.

pkg/cache/scheduler/tas_flavor_snapshot.go
yuluo-yx: yep, done

pkg/cache/scheduler/tas_flavor_snapshot.go
yuluo-yx: In PodSetGrouping with a leader, `stateWithLeader` indicates "how many workers can be added to a domain after adding the leader," so only children with `child.leaderState > 0` can participate in penalty calculation. Otherwise, the occupied child will only have a penalty of 0, as mentioned in the bug report.

However, in a normal TAS without a leader, `stateWithLeader` will still be read by subsequent general paths. In this case, it should degenerate into normal worker capacity, meaning all children participate in the calculation. If this branch is not entered when `!leaderRequired` is called, `minStateWithLeaderDifference` will remain `math.MaxInt32`, the parent domain's `stateWithLeader` will be set to 0, and worker-only scheduling will be judged as unavailable.

pkg/cache/scheduler/tas_flavor_snapshot.go
yuluo-yx: How about using bool instead? 👀 

such as:

```shell
hasWithLeaderCapacityContributor = false

if !leaderRequired || child.leaderState > 0 {
	hasWithLeaderCapacityContributor = true
	.....
}

if hasWithLeaderCapacityContributor {
	// do
}
```

pajakd: Yes, exactly

pkg/cache/scheduler/tas_flavor_snapshot.go
pajakd: OK, Makes sense. It would be nice (in a future PR) to decouple so that TAS without leader does not depend on any of the "withLeader" fields or variables.

pkg/cache/scheduler/tas_flavor_snapshot.go
yuluo-yx: done

pkg/cache/scheduler/tas_flavor_snapshot.go
yuluo-yx: got it. I created #10812 to trac it.

mimowo (commented):
Thank you 👍 
/lgtm
/approve

mimowo:
@yuluo-yx could you check if we could cherrypick without conflicts if we cherrypicked first https://github.com/kubernetes-sigs/kueue/pull/10506 and https://github.com/kubernetes-sigs/kueue/pull/10537? 

If this is the case then I'm happy to CP then. Otherwise we can use ./hack/cherry_pick_pull.sh to prepare the CPs.

yuluo-yx:
> @yuluo-yx could you check if we could cherrypick without conflicts if we cherrypicked first #10506 and #10537?
> 
> If this is the case then I'm happy to CP then. Otherwise we can use ./hack/cherry_pick_pull.sh to prepare the CPs.

got it, I have verified this locally:

For `release-0.17`, even after cherry-picking #10506 and #10537, #10783 still results in a conflict within `pkg/cache/scheduler/tas_balanced_placement.go`. The reason is that #10783 depends on the splitting of the "topology assignment state" introduced in #10598. By applying the patches in the sequence #10506 -> #10537 -> #10598 -> #10783, #10783 can be applied to `release-0.17` without any conflicts.

For `release-0.16`, #10506 itself generates a conflict in `pkg/cache/scheduler/tas_flavor_snapshot.go`; therefore, simply backporting #10506 and #10537 beforehand does not yield a conflict-free path.

The conclusion is: this approach works for `release-0.17` only if #10598 is also backported; for `release-0.16`, manual preparation is still required to perform the cherry-pick.

mimowo:
@yuluo-yx thank you for the summary. Okay let me proceed with 0.17 path. All of the PRs are straighforward cleanups.

> For release-0.16, https://github.com/kubernetes-sigs/kueue/pull/10506 itself generates a conflict in pkg/cache/scheduler/tas_flavor_snapshot.go; therefore, simply backporting https://github.com/kubernetes-sigs/kueue/pull/10506 and https://github.com/kubernetes-sigs/kueue/pull/10537 beforehand does not yield a conflict-free path.

would it be feasible to backport this one manually, to unblock backporting this PR easier for 0.16?

I'm thinking for 0.16 we could put all the chain as a single PR, just different commits, wdyt?

mimowo:
/cherrypick release-0.17
This should be clean now

Edit: failed, ptal

mimowo:
/release-note-edit
```release-note
TAS: Fix handling of PodSet groups which could lead in some scenarios to empty topologyAssignment.
```

Issue #10766, opened 2026-04-24
[MultiKueue] Composable dispatcher KEP


amy:
**What would you like to be added**:
We need a composable dispatcher that allows for different routing policies. The overall goal is to treat clusters like nodes with concepts like filtering and scoring, and also node anti/affinity. 

**Why is this needed**:
MultiKueue is a powerful concept but today it has major shortcomings. Today its limited to submitting the workload to all worker clusters and allowing the first one that admits it to pick up the workload. At most you can define some limited set of clusters to submit to. 

I need stronger policy setting for training clusters. Users want to be able to do things like :
1. Target highest quality of quota (ex: target guarantees or best burst, decreasing probability of preemption)
2. Have affinity to things like data locality or specific services installed on the cluster, etc. 

(These two are out of scope and require separate keps that I will link. But you need the foundation for adding these sort of plugins.)

**Completion requirements**:

This enhancement requires the following artifacts:

- [x] Design doc: https://github.com/kubernetes-sigs/kueue/pull/10784
- [x] API change
- [x] Docs update

The artifacts should be linked in subsequent comments.

amy:
fyi/ @olekzabl for visibility

maishivamhoo123:
@amy Is there an existing design doc I should align with before I start the framework implementation?

amy:
@maishivamhoo123 yeah there's context you're missing. Please join batch meeting group meeting.

And re: this particular issue, Happy to collaborate. See my above assign comment: https://github.com/kubernetes-sigs/kueue/issues/10766#issuecomment-4314179592

olekzabl:
Let me link to [this comment](https://github.com/kubernetes-sigs/kueue/pull/10145#discussion_r3140639040) as I feel it may be relevant here.

In short: any functionality like this feels great - but risky for scalability.
I think we should keep this in mind and invest in benchmarks to research how "smart" dispatching we can afford.

I'll comment on #10784 as well.

amy:
> In short: any functionality like this feels great - but risky for scalability. I think we should keep this in mind and invest in benchmarks to research how "smart" dispatching we can afford.

@olekzabl Agree on some baseline of scalability. But also note that people are willing to take more time for placement decisions if the placement is "better". The bottleneck that would be bad is if placement decisions affected allocation on the worker, so that's baseline.

[...]

Issue #10105, opened 2026-03-25
Support worker resource stats visibility in MultiKueue manager cluster


olekzabl:
**What would you like to be added**:

A ClusterQueue on a MultiKueue manager cluster should track the **resource usage** (on the worker side) of workers submitted to it, and present a summary of those to the user through the manager cluster API, without a need for the user to log in to all the workers.

A nice addition would be also to track the **quotas** on the worker side, to let users compare the usage against the quotas.
(Note: this part is related to #9988, though distinct from it in *not* involving a customizable multiplier).

**Why is this needed**:

MultiKueue users often would like to treat the manager cluster as a single "management panel", without needing to log in to the worker cluster (except for when troubleshooting). On the other hand, understanding the picture of resource usage vs. capacity is central to using Kueue.

**Completion requirements**:

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

amy:
Tracking this one as I think its a prerequisite for this: https://github.com/kubernetes-sigs/kueue/issues/10766

olekzabl:
/unassign
as it's no longer among my priorities.

olekzabl:
> Tracking this one as I think its a prerequisite for this: [#10766](https://github.com/kubernetes-sigs/kueue/issues/10766)

FTR I think it's not really a prerequisite for that; details are discussed in [this comment](https://github.com/kubernetes-sigs/kueue/pull/10784#discussion_r3362595775) on the KEP for #10766.

Issue #8585, opened 2026-01-14
MultiKueue: workloads from worker clusters are deleted prematurely


mimowo:
**What happened**:

NOTE: this is based on static code analysis for now, specifically of this code: https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/admissionchecks/multikueue/workload.go#L409-L420

This is quite important for the use-case of searching for capacity with ProvisioningRequests. Otherwise the we only run one ProvisiniongRequest at the time.

**What you expected to happen**:

Deletion of the other workers should be deferred only after one is admitted.

**Anything else we need to know?**:

Since this issue existed for long we should introduce a feature gate for safety.

I consider it a bug, becuase it was stated in the [KEP](https://github.com/kubernetes-sigs/kueue/tree/main/keps/693-multikueue#proposal) that we should wait until the workload is admitted, rather than quota reserving, sepecifically:

If a remote workload is **admitted** first, the job will be created in the remote cluster with a kueue.x-k8s.io/prebuilt-workload-name label pointing to that clone. Then it will remove the workloads from the remaining worker clusters and allow the single instance of the job to proceed. The workload will be also admitted in the management cluster.

mimowo:
/assign @mbobrovskyi 
tentatively
cc @mszadkow @olekzabl @mwielgus @yaroslava-serdiuk

mbobrovskyi:
Should this feature be in Alpha stage?

mimowo:
I think this is a bug. IIUC the current implementation is inconsistent with the KEP, and I don't see a need for new API. Let me know if I'm missing something.

Since this behavior was for long we can use a feature gate for safety.

mimowo:
Using Admitted rather than QuotaReserved should not have a difference for clusters which are not using custom admission checks.  

However, the current behavior of looking into QuotaReserved is currently broken when using provisioning requests, because we prefer to look for the capacity at the same time on all clusters, still first one wins. If one can limit the number of clusters to be checked in parallel, they can use the status.nominatedClusterNames.

IrvingMg:
I think the question for Alpha was in regards to the feature gate. 

IIUC, we need to add a new feature gate, e.g.`MultiKueueWaitForWorkloadAdmitted`, which changes the condition to check from QuotaReserved to Admitted before cleaning up non-selected worker workloads. Should this start as Alpha?

mimowo:
I see, but I  think with Beta, unless the. implementation proves to be very complex and risky

PR #1681, opened 2024-02-02
[multikueue] Manage worker cluster unavailability

trasc:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

- Keep the AdmissionCheck Active even if not all it's clusters are usable.
- Add a mechanism which keeps a workload's MultiKueue AdmissionCheckState as `Ready`  for a configured time even if the manager cannot connect to its reserving worker.

#### Which issue(s) this PR fixes:

Relates to #693

#### Does this PR introduce a user-facing change?

```release-note
MuliKueue  - Manage worker cluster unavailability
```

PR #10784, opened 2026-04-26
WIP: [KEP] composable multikueue dispatcher

amy:
#### What type of PR is this?
THIS IS VERY MUCH A WIP. Made this PR so that people know that its being worked on. 

 /kind kep

We need a composable dispatcher that allows for different routing policies. The overall goal is to treat clusters like nodes with concepts like filtering and scoring, and also node anti/affinity.

#### Which issue(s) this PR partially fixes:

Part of: https://github.com/kubernetes-sigs/kueue/issues/10766

its a WIP

```release-note
NONE
```

### DIFF
--- a/keps/8692-composable-multikueue-dispatcher/README.md
+++ b/keps/8692-composable-multikueue-dispatcher/README.md
@@ -1,0 +1,791 @@
+# KEP-8692: Composable MultiKueue Dispatcher
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: Region-constrained GPU jobs](#story-1-region-constrained-gpu-jobs)
+    - [Story 2: Cost-aware hardware-tier dispatching](#story-2-cost-aware-hardware-tier-dispatching)
+    - [Story 3: Cluster taints for maintenance](#story-3-cluster-taints-for-maintenance)
+    - [Story 4: Data-locality-aware dispatching for ML training](#story-4-data-locality-aware-dispatching-for-ml-training)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Mental Model: Clusters as Nodes](#mental-model-clusters-as-nodes)
+  - [Dispatch Pipeline](#dispatch-pipeline)
+  - [API: MultiKueueCluster — Taints and Labels](#api-multikueuecluster--taints-and-labels)
+  - [API: WorkloadSpec — Cluster Placement Fields](#api-workloadspec--cluster-placement-fields)
+  - [API: ClusterDispatcherProfile CRD](#api-clusterdispatcherprofile-crd)
+  - [API: ClusterQueue Changes](#api-clusterqueue-changes)
+  - [Workload Annotation: Dispatch Mode](#workload-annotation-dispatch-mode)
+  - [Plugin Interfaces](#plugin-interfaces)
+  - [Built-in Plugins](#built-in-plugins)
+    - [Filter Plugins](#filter-plugins)
+    - [Score Plugins](#score-plugins)
+  - [Dispatch Algorithm](#dispatch-algorithm)
+  - [Interaction with Existing Dispatchers](#interaction-with-existing-dispatchers)
+  - [Validation](#validation)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [Stable](#Stable)
+  - [Observability](#observability)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Future Work](#future-work)
+<!-- /toc -->
+
+## Summary
+
+This KEP introduces a composable, plugin-based dispatcher for MultiKueue that treats child
+clusters as schedulable units — directly analogous to how the Kubernetes scheduler treats
+nodes. Workloads express cluster placement requirements (`clusterSelector`, `clusterAffinity`,
+`clusterTolerations`) directly in `WorkloadSpec`, mirroring how pods express node placement
+via `nodeSelector`, `nodeAffinity`, and `tolerations`. Operators configure a
+`Filter → Score → Normalize → Select` pipeline through a single new CRD:
+`ClusterDispatcherProfile` (operator-level plugin selection and scoring weights). The
+dispatcher nominates exactly one cluster at a time, offering a new dispatcher alongside the 
+existing AllAtOnce and Incremental dispatchers.
+
+## Motivation
+
+MultiKueue today provides two dispatching strategies:
+
+- **AllAtOnce**: nominates every cluster in the `MultiKueueConfig` simultaneously; the first
+  cluster to admit the workload wins. All other nominated remote workloads are then cleaned
+  up via sequential deletion. This is wasteful — O(N) remote objects are created and O(N-1)
+  are deleted per admission — and admission order is non-deterministic.
+
+- **Incremental**: nominates clusters in alphabetical batches of three, with a five-minute
+  round timeout. Cluster order is unrelated to any workload or cluster property, so
+  placement is essentially arbitrary.
+
+Neither strategy allows workloads to express placement requirements ("run only on GPU clusters
+in us-west-2") or operators to rank clusters by preference ("prefer A100 clusters before H100
+clusters"). The only filtering today is cluster health (the `Active` condition), and that
+filtering is implicit and hard-coded.
+
+As MultiKueue deployments grow — spanning regions, cloud providers, availability zones, and
+hardware tiers — both workload owners and operators need a principled way to control where
+workloads land. This KEP provides that control through a workload-driven placement model
+paired with a declarative operator plugin pipeline.
+
+### Goals
+
+- Introduce a `Filter` extension point: plugins that eliminate ineligible clusters before
+  scoring (hard constraints).
+- Introduce a `Score` extension point: plugins that rank remaining clusters by preference
+  (soft constraints), with operator-configurable weights.
+- Add `spec.clusterSelector`, `spec.clusterAffinity`, and `spec.clusterTolerations` fields
+  to `WorkloadSpec`: workload-owned placement requirements, analogous to pod `nodeSelector`,
+  `nodeAffinity`, and `tolerations`.
+- Implement `ClusterAffinity` filter: evaluates `WorkloadSpec.ClusterSelector` and
+  `WorkloadSpec.ClusterAffinity.Required` against `MultiKueueCluster` labels.
+- Implement `ClusterTaints` filter: rejects clusters whose `NoSchedule` taints are not
+  tolerated by `WorkloadSpec.ClusterTolerations`.
+- Implement `ClusterHealth` filter: rejects clusters whose `Active` condition is not `True`.
+- Implement `ClusterAffinityScore`: scores clusters by `WorkloadSpec.ClusterAffinity.Preferred`
+  terms, analogous to Kubernetes preferred node affinity scoring.
+- Implement `ClusterTaintScore`: reduces scores for clusters with untolerated
+  `PreferNoSchedule` taints.
+- Introduce `ClusterDispatcherProfile` CRD: operator-controlled plugin list and scoring
+  weights.
+- Top-1 nomination model (compare to AllAtOnce/Incremental).
+- Gate all new behavior behind a `ComposableDispatcher` feature gate.
+- Implement `ClusterFeasibility` filter: rejects clusters that cannot admit the workload
+  based on quota headroom visible in `MultiKueueCluster.Status`. Blocked on
+  [kubernetes-sigs/kueue#10105](https://github.com/kubernetes-sigs/kueue/issues/10105)
+  (worker resource stats visibility in the manager cluster); the interface is defined at
+  alpha and the implementation ships at beta.
+
+### Non-Goals
+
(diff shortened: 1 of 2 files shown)


keps/8692-composable-multikueue-dispatcher/README.md
```diff
@@ -0,0 +1,791 @@
...
+- **AllAtOnce**: nominates every cluster in the `MultiKueueConfig` simultaneously; the first
+  cluster to admit the workload wins. All other nominated remote workloads are then cleaned
+  up via sequential deletion. This is wasteful — O(N) remote objects are created and O(N-1)
+  are deleted per admission — and admission order is non-deterministic.
+
+- **Incremental**: nominates clusters in alphabetical batches of three, with a five-minute
+  round timeout. Cluster order is unrelated to any workload or cluster property, so
+  placement is essentially arbitrary.
+
+Neither strategy allows workloads to express placement requirements ("run only on GPU clusters
```
mwielgus, line 66: Playing a devil's advocate a bit.  In theory there can be a completely custom, user-provided dispatcher. I'm wondering why do we need fine grain extensions points which would require some controller anyway. Maybe we should just provide a reference implementation and encourage users to fork it?

amy: I need a cloud agnostic framework that users use across clouds. And then be able to have custom configuration recommendations. The first pass of this would be a dispatcher that's in tree anyways. See: https://github.com/kubernetes-sigs/kueue/pull/10784/changes/BASE..4ec7a758f51f1527cfda7c9e527cc2f307db2f3b#diff-ab13a7d3e9220a7d4c936bdf44adc4aaffdd91eecc1d524da90bac251b82ba5aR106

Like I need to be able to have filtering and scoring semantics for basic affinity/anti-affinity stuff. So sure you could have people write a new one that has no policy configuration but have the policy thats embedded into the code. But that... is a bad experience and not reusable 

This is just another dispatcher option that sits alongside Incremental and AllAtOnce as a third option.

keps/8692-composable-multikueue-dispatcher/README.md
```diff
@@ -0,0 +1,791 @@
...
+
+`WorkloadSpec.ClusterSelector`, `ClusterAffinity`, and `ClusterTolerations` are populated
+by Kueue's job framework controllers from well-known annotations on the owner job. This
+mirrors how pod placement fields are set in a job's pod template:
+
+| WorkloadSpec field | Source annotation on owner job |
+|---|---|
+| `clusterSelector` | `kueue.x-k8s.io/cluster-selector` (JSON `map[string]string`) |
+| `clusterAffinity` | `kueue.x-k8s.io/cluster-affinity` (JSON-encoded `ClusterAffinity`) |
+| `clusterTolerations` | `kueue.x-k8s.io/cluster-tolerations` (JSON array of `ClusterToleration`) |
```
olekzabl, line 408: I'm anxious about these JSON encodings.
AFAICS JSON encoding does not happen for "pod placement fields", nor elsewhere in Kueue; even in K8s ecosystem it seems rather rare (and, when present, not fully standardized).

For the user, this would be typo-prone. (Cf. item 1 in your Risks, but not just label values, also JSON syntax).
I understand it's webhook-enforceable, still unpleasant.

I do see where it comes from - we want these specified for the original jobs (not Workloads), which can have various types; hence annotations; hence flat string values.

Still, maybe there are alternatives?
E.g. define a new structured CRD to hold all these fields and just refer an instance of this CRD in a job annotation value?
(We could also accept annotations for Alpha, for more elastic experimenting, and then plan for standardizing into a CRD for Beta).
This could win some type safety, but at the cost of inconvenience of spreading spec across resources.

WDYT?

keps/8692-composable-multikueue-dispatcher/README.md
```diff
@@ -0,0 +1,791 @@
...
+- `default` profile installed by Kueue.
+- Top-1 nomination with retry.
+- Validation webhook.
+- Unit and integration tests passing.
+
+#### Beta
+
+- Feature gate enabled by default.
+- `ClusterFeasibility` filter fully implemented (requires [kubernetes-sigs/kueue#10105](https://github.com/kubernetes-sigs/kueue/issues/10105)).
+- `CapacityScore` plugin implemented (also requires [kubernetes-sigs/kueue#10105](https://github.com/kubernetes-sigs/kueue/issues/10105)).
```
olekzabl, line 727: How do these 2 differ?
Is it _filter_ ("required") vs. _score_ ("preferred")? Or sth more?

keps/8692-composable-multikueue-dispatcher/README.md
```diff
@@ -0,0 +1,791 @@
...
+type WorkloadSpec struct {
+    // ... existing fields unchanged ...
+
+    // ClusterSelector is a map of label key/value pairs. The target MultiKueueCluster
+    // must have all listed labels to be eligible for dispatch.
+    // Analogous to pod.spec.nodeSelector.
+    // Only evaluated when a MultiKueue AdmissionCheck is active on the workload's
+    // ClusterQueue.
+    // +optional
+    ClusterSelector map[string]string `json:"clusterSelector,omitempty"`
```
olekzabl, line 324: Nit: do I understand correctly that the value (even after JSON-decoding) will *not* be of the `ClusterSelector` type defined below? (Instead, it'll be a standard label selector)?
If so, this may be confusing.

keps/8692-composable-multikueue-dispatcher/README.md
```diff
@@ -0,0 +1,791 @@
...
+  terms, analogous to Kubernetes preferred node affinity scoring.
+- Implement `ClusterTaintScore`: reduces scores for clusters with untolerated
+  `PreferNoSchedule` taints.
+- Introduce `ClusterDispatcherProfile` CRD: operator-controlled plugin list and scoring
+  weights.
+- Top-1 nomination model (compare to AllAtOnce/Incremental).
+- Gate all new behavior behind a `ComposableDispatcher` feature gate.
+- Implement `ClusterFeasibility` filter: rejects clusters that cannot admit the workload
+  based on quota headroom visible in `MultiKueueCluster.Status`. Blocked on
+  [kubernetes-sigs/kueue#10105](https://github.com/kubernetes-sigs/kueue/issues/10105)
```
olekzabl, line 100: What is meant by "quota headroom" - is it "unreserved capacity"?
If so, defining a strict filter based on that may be too strict - even if a workload "seems not to fit", it could still borrow or preempt.

I imagine we could _strictly reject_ a workload if its requests exceed _worker (whole) capacity_. 
"Unreserved capacity" feels like valuable information - though ideally only for _soft scoring_.

keps/8692-composable-multikueue-dispatcher/README.md
```diff
@@ -0,0 +1,791 @@
...
+    // +optional
+    DispatcherProfileRef *DispatcherProfileReference `json:"dispatcherProfileRef,omitempty"`
+}
+
+type DispatcherProfileReference struct {
+    Name string `json:"name"`
+}
+```
+
+### Workload Annotation: Dispatch Mode
```
olekzabl, line 486: I can't understand the rationale for `BestEffort`.
I'm not sure when the misunderstanding is - so let me try nail it down with the following questions / remarks:

1. In `BestEffort`, when a workload is dispatched to a worker and the AC is marked `Ready`, what will be the workload status on manager & selected worker? 

   A: `Admitted` on manager / `QuotaReserved` on worker?
   B:`Admitted` on both (even though actually not running yet on the worker)?

   I assumed A but my AI friend claimed it's B. Hence asking.

2. If it's 1B ("`Admitted` on both"), doesn't it break some contracts?
   What if a worker has its own AdmissionChecks - would we just magically skip them?

3. _If the remote cluster later evicts the workload, the eviction propagates back to the manager_
   You say this for `BestEffort` - but why wouldn't it hold also for `MustBeAdmitted`?

4. _If the remote does not admit within `workerLostTimeout`, the dispatcher retries with the next-best cluster._
   You say this for `MustBeAdmitted` - but why wouldn't it hold also for `BestEffort`?
   (At least assuming 1A = "`QuotaReserved` on worker" - this could stay for long, even in `BestEffort`).

5. You mention `workerLostTimeout` but I'm feeling it doesn't fit in this story.
   * IIUC the intent behind `workerLostTimeout` is to handle cases when we _lost connection_ to a worker cluster on which the remote workload _had already started_; see [this comment](https://github.com/kubernetes-sigs/kueue/blob/4e440ec3f781c2e86c0b0041d89589995ff80f8e/pkg/controller/admissionchecks/multikueue/workload.go#L505-L506) and the description of #1681.
     In that case, we want to _speculatively assume_ for some time that the workload is still running there - to avoid re-dispatching which could later turn out to have been wasteful.
   * The case in this KEP seems very different. Here (IIUC) we still _can see_ the worker cluster but the workload _failed to start_ on it.
     This feels more like `waitForPodsReady` (though admittedly can't be directly expressed by that).

6. I can't help the feeling that the choice between `BestEffort` / `MustBeAdmitted`:
   * will not affect the _actual status_ (running / not running) of the workload on the 1st selected worker, or the timeline of that status
   * ideally should not affect the timeline of manager's reaction to any worker-side complications
   * and hence, by induction, ideally should have no effect on "placement latency" at all
   * by definition, it affects local workload statuses - but this feels just abstract labeling?

7. Historically, the early MultiKueue behavior was "mark the workload `Admitted` on the manager as soon as it got `QuotaReserved` on some worker" - but then, it was considered imperfect and evolved towards "let manager-side admission follow worker-side admission" (#8585).
   In this context, introducing `BestEffort` seems to be a step in the opposite direction (and even more so making it default).
   So I'd really like to understand the reason for this.

keps/8692-composable-multikueue-dispatcher/README.md
```diff
@@ -0,0 +1,791 @@
...
+**In-tree plugins require Kueue code changes for new scoring logic.** Operators cannot add
+custom cluster scoring without building a fork or waiting for upstream acceptance. The
+out-of-tree extension path is deferred, not excluded.
+
+## Future Work
+
+- **Top-K nomination.** The current dispatcher nominates exactly one cluster per cycle
+  (top-1). A future KEP can add parallel nomination of K clusters simultaneously as an
+  opt-in `nominationMode: TopK` field on `ClusterDispatcherProfile`, reducing tail latency
+  when the top-scoring cluster is slow to admit. Top-1 remains the default.
```
olekzabl, line 781: Or, without introducing "nomination modes", just expose an `*int` for K, where `nil` means 1.
Cf. [KEP-9270](https://github.com/kubernetes-sigs/kueue/pull/10877) which proposes a similar thing for the incremental dispatcher.

BTW I'd reconsider if adding this wouldn't be valuable right away.
How well can we predict if a worker will actually run the workload?
I think not too well, given preemptions, borrowing, ProvReqs etc.
(I mean, even with #10105 in place. That feature is just about some rough quota / usage stats; it won't solve any of the aspects mentioned above).

olekzabl (commented):
Hello @amy , my apologies for reading it so late.

I generally like the ideas here, and I think the topic is essential for MultiKueue.

I didn't manage to read very deeply - but let me leave a few early questions.

amy:
> Hello @amy , my apologies for reading it so late.
> I generally like the ideas here, and I think the topic is essential for MultiKueue.
> I didn't manage to read very deeply - but let me leave a few early questions.

@olekzabl No worries! Just plopped it here even though it's very much a wip bc I wanted to let people know I'm thinking about it and really need it. I'll probably engage with this deeply towards the end of 0.19 / start of 0.20 release.

Also, tbh I need to do more work on understanding the gaps that need to be filled in order for this kep to be possible.

keps/8692-composable-multikueue-dispatcher/README.md
```diff
@@ -0,0 +1,791 @@
...
+  `PreferNoSchedule` taints.
+- Introduce `ClusterDispatcherProfile` CRD: operator-controlled plugin list and scoring
+  weights.
+- Top-1 nomination model (compare to AllAtOnce/Incremental).
+- Gate all new behavior behind a `ComposableDispatcher` feature gate.
+- Implement `ClusterFeasibility` filter: rejects clusters that cannot admit the workload
+  based on quota headroom visible in `MultiKueueCluster.Status`. Blocked on
+  [kubernetes-sigs/kueue#10105](https://github.com/kubernetes-sigs/kueue/issues/10105)
+  (worker resource stats visibility in the manager cluster); the interface is defined at
+  alpha and the implementation ships at beta.
```
olekzabl, line 102: I think [#10105](https://github.com/kubernetes-sigs/kueue/issues/10105) (which we kind of shelved for now) is actually **not a blocker** here.

* The key blockers for the progress of [#10105](https://github.com/kubernetes-sigs/kueue/issues/10105) are:
  * how to efficiently measure the "given manager CQ -> given worker CQ used volume" (which in general requires looking at individual workloads due to many-to-many CQ-to-CQ relation)
  * how to nicely surface the resource stats in the API

  which are both *irrelevant* for the purposes of this KEP.

* On the other hand, what this KEP needs is the awareness of total / unused workers capacity on the manager side, possibly broken down per flavor.
  That I'd say is *almost* achieved in [#11141](https://github.com/kubernetes-sigs/kueue/pull/11141); specifically:
  * manager CQ is now aware of worker CQ quotas
  * the missing bits are awareness of "unusued quotas" and breakdown per flavor, however:
    * that's easier than [#10105](https://github.com/kubernetes-sigs/kueue/issues/10105) in that it doesn't touch individual workloads, or API surface
    * [#11141](https://github.com/kubernetes-sigs/kueue/pull/11141) introduced caching of remote CQs (& LQs) on the manager, so now the missing bits boil down to using [these fields](https://github.com/kubernetes-sigs/kueue/blob/1744552f93d8a886c7265dd81aee3198d79424b6/apis/kueue/v1beta2/clusterqueue_types.go#L362-L377) from the already-fetched remote CQ resources in the computations.

PR #10785, opened 2026-04-26
fix(queue): avoid blocking send on DRA reconcile channel

gyliu513:
#### What type of PR is this?

The original blocking send was inside the manager lock. If the channel buffer (size=10) filled up, AddLocalQueue would hold the manager lock for an extended period, blocking all other manager operations (scheduling, workload admission, etc.) until the channel drained.

/kind bug

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/cache/queue/manager.go
+++ b/pkg/cache/queue/manager.go
@@ -450,13 +450,13 @@ func (m *Manager) DefaultLocalQueueExist(namespace string) bool {
 // addLocalQueueLocked registers the local queue under the manager lock and
 // returns the DRA reconcile channel and any workloads that need DRA reconciling.
 // Callers must send events to the channel after this returns (outside any lock).
-func (m *Manager) addLocalQueueLocked(ctx context.Context, q *kueue.LocalQueue) (chan<- event.TypedGenericEvent[*kueue.Workload], []*kueue.Workload, error) {
+func (m *Manager) addLocalQueueLocked(ctx context.Context, q *kueue.LocalQueue) ([]*kueue.Workload, error) {
 	m.Lock()
 	defer m.Unlock()
 
 	key := queue.Key(q)
 	if _, ok := m.localQueues[key]; ok {
-		return nil, nil, fmt.Errorf("queue %q already exists", q.Name)
+		return nil, fmt.Errorf("queue %q already exists", q.Name)
 	}
 	qImpl := newLocalQueue(q)
 	m.localQueues[key] = qImpl
@@ -470,7 +470,7 @@ func (m *Manager) addLocalQueueLocked(ctx context.Context, q *kueue.LocalQueue) 
 	// queue might have been added earlier.
 	var workloads kueue.WorkloadList
 	if err := m.client.List(ctx, &workloads, client.MatchingFields{utilindexer.WorkloadQueueKey: q.Name}, client.InNamespace(q.Namespace)); err != nil {
-		return nil, nil, fmt.Errorf("listing workloads that match the queue: %w", err)
+		return nil, fmt.Errorf("listing workloads that match the queue: %w", err)
 	}
 	var draWorkloads []*kueue.Workload
 	for _, w := range workloads.Items {
@@ -486,11 +486,9 @@ func (m *Manager) addLocalQueueLocked(ctx context.Context, q *kueue.LocalQueue) 
 
 		log := ctrl.LoggerFrom(ctx).WithValues("workload", klog.KObj(&w))
 		if dra.NeedsDRAReconcile(&w) {
-			if m.draReconcileChannel != nil {
-				// Collect DRA workloads to send outside the lock; DeepCopy keeps a
-				// stable pointer since the range variable is reused each iteration.
-				draWorkloads = append(draWorkloads, w.DeepCopy())
-			}
+			// Collect DRA workloads to send outside the lock; DeepCopy keeps a
+			// stable pointer since the range variable is reused each iteration.
+			draWorkloads = append(draWorkloads, w.DeepCopy())
 			continue
 		}
 
@@ -504,18 +502,22 @@ func (m *Manager) addLocalQueueLocked(ctx context.Context, q *kueue.LocalQueue) 
 		m.Broadcast()
 	}
 
-	return m.draReconcileChannel, draWorkloads, nil
+	return draWorkloads, nil
 }
 
 func (m *Manager) AddLocalQueue(ctx context.Context, q *kueue.LocalQueue) error {
-	ch, draWorkloads, err := m.addLocalQueueLocked(ctx, q)
+	draWorkloads, err := m.addLocalQueueLocked(ctx, q)
 	if err != nil {
 		return err
 	}
 
+	if !features.Enabled(features.DynamicResourceAllocation) {
+		return nil
+	}
+
 	for _, wl := range draWorkloads {
 		log := ctrl.LoggerFrom(ctx).WithValues("workload", klog.KObj(wl))
-		ch <- event.TypedGenericEvent[*kueue.Workload]{Object: wl}
+		m.draReconcileChannel <- event.TypedGenericEvent[*kueue.Workload]{Object: wl}
 		log.V(4).Info("Sent DRA workload to reconcile channel due to LocalQueue creation")
 	}
 
--- a/pkg/cache/queue/manager.go
+++ b/pkg/cache/queue/manager.go
@@ -449,10 +449,10 @@ func (m *Manager) DefaultLocalQueueExist(namespace string) bool {
 
 func (m *Manager) AddLocalQueue(ctx context.Context, q *kueue.LocalQueue) error {
 	m.Lock()
-	defer m.Unlock()
 
 	key := queue.Key(q)
 	if _, ok := m.localQueues[key]; ok {
+		m.Unlock()
 		return fmt.Errorf("queue %q already exists", q.Name)
 	}
 	qImpl := newLocalQueue(q)
@@ -467,8 +467,10 @@ func (m *Manager) AddLocalQueue(ctx context.Context, q *kueue.LocalQueue) error 
 	// queue might have been added earlier.
 	var workloads kueue.WorkloadList
 	if err := m.client.List(ctx, &workloads, client.MatchingFields{utilindexer.WorkloadQueueKey: q.Name}, client.InNamespace(q.Namespace)); err != nil {
+		m.Unlock()
 		return fmt.Errorf("listing workloads that match the queue: %w", err)
 	}
+	var draWorkloadsToReconcile []*kueue.Workload
 	for _, w := range workloads.Items {
 		m.assignWorkload(workload.Key(&w), qImpl.Key)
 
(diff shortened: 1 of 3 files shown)


pkg/cache/queue/manager.go
```diff
@@ -435,8 +435,14 @@ func (m *Manager) AddLocalQueue(ctx context.Context, q *kueue.LocalQueue) error
...
 		if dra.NeedsDRAReconcile(&w) {
 			if m.draReconcileChannel != nil {
-				m.draReconcileChannel <- event.TypedGenericEvent[*kueue.Workload]{Object: &w}
-				log.V(4).Info("Sent DRA workload to reconcile channel due to LocalQueue creation")
+				select {
+				case m.draReconcileChannel <- event.TypedGenericEvent[*kueue.Workload]{Object: &w}:
+					log.V(4).Info("Sent DRA workload to reconcile channel due to LocalQueue creation")
+				default:
+					// Best-effort fast-path only: dropping is safe because the workload reconciler
+					// will eventually observe the object via watches and reconcile.
```
mimowo: I'm not totally sure this is safe, because what if the channel is full with events for other workloads, then by dropping this even we may drop all events for the workload.

cc @sohankunkerkar @kannon92 who are also recently looking into DRA

sohankunkerkar: I agree with @mimowo here. Dropping isn't safe. For DRA workloads discovered during `AddLocalQueue`, this channel send is the handoff that triggers workload reconcile and a workload that already existed before the LocalQueue was created may not get any later workload event to re-fire it. Ideally, we should collect DRA workloads inside the lock, release it, then send to `draReconcileChannel`, that preserves guaranteed delivery without a blocking send under the queue manager lock.

sohankunkerkar: Also, @gyliu513 could you provide some context around this? like how did you encounter this issue?

gyliu513: @sohankunkerkar @mimowo Thanks for the detailed review, you are correct.

My original assumption that "the workload reconciler will eventually observe  the object via watches" is wrong in this scenario. These are pre-existing  workloads that were already reconciled before the LocalQueue was created, the channel send in AddLocalQueue is the *only* trigger that re-fires their  reconciliation. Dropping it would leave them stuck.

I'll implement the fix you proposed: collect the DRA workloads inside the  lock, release it, then do a blocking send to draReconcileChannel to guarantee  delivery without holding the lock.

This is actually a performance issue here, we should never have a blocking send in a critical section while holding the manager lock, as it can stall all other manager operations (scheduling, workload admission, etc.) if the channel buffer (size=10) fills up. Like when LoacalQueue recovered or created, `AddLocalQueue` may detect bunch of workloads which will block this function.

pkg/cache/queue/manager.go
```diff
@@ -442,10 +442,10 @@ func (m *Manager) DefaultLocalQueueExist(namespace string) bool {
 
 func (m *Manager) AddLocalQueue(ctx context.Context, q *kueue.LocalQueue) error {
 	m.Lock()
-	defer m.Unlock()
 
 	key := queue.Key(q)
 	if _, ok := m.localQueues[key]; ok {
+		m.Unlock()
```
sohankunkerkar: The fix looks right. The manual unlock is error-prone if someone adds an early return later; consider extracting the locked section into a helper that returns the DRA workload slice.

gyliu513: Done

sohankunkerkar (approved):
/lgtm

pkg/cache/queue/manager.go
```diff
@@ -498,6 +503,22 @@ func (m *Manager) AddLocalQueue(ctx context.Context, q *kueue.LocalQueue) error
...
+
+func (m *Manager) AddLocalQueue(ctx context.Context, q *kueue.LocalQueue) error {
+	ch, draWorkloads, err := m.addLocalQueueLocked(ctx, q)
+	if err != nil {
+		return err
+	}
+
+	for _, wl := range draWorkloads {
+		log := ctrl.LoggerFrom(ctx).WithValues("workload", klog.KObj(wl))
+		ch <- event.TypedGenericEvent[*kueue.Workload]{Object: wl}
```
mimowo: Actually, could this `ch` be ever nil when DRA is used? The check above suggests it could, but I'm not sure: https://github.com/kubernetes-sigs/kueue/pull/10785/changes#diff-c7ed976ece2ea8cfd72847c76a214be5920b16aa7770e4d3cef7d8c974a36028R489

pkg/cache/queue/manager.go
```diff
@@ -498,6 +503,22 @@ func (m *Manager) AddLocalQueue(ctx context.Context, q *kueue.LocalQueue) error
 	if cq != nil && cq.AddFromLocalQueue(qImpl, m.roleTracker, m.customLabels) {
 		m.Broadcast()
 	}
+
+	return m.draReconcileChannel, draWorkloads, nil
+}
+
+func (m *Manager) AddLocalQueue(ctx context.Context, q *kueue.LocalQueue) error {
+	ch, draWorkloads, err := m.addLocalQueueLocked(ctx, q)
```
mimowo: Why do we return / pass the channel if this is a field on  manager, so we could just reference it as `m.draReconcileChannel`, wdyt?

pkg/cache/queue/manager.go
```diff
@@ -498,6 +503,22 @@ func (m *Manager) AddLocalQueue(ctx context.Context, q *kueue.LocalQueue) error
...
+	if err != nil {
+		return err
+	}
+
+	for _, wl := range draWorkloads {
+		log := ctrl.LoggerFrom(ctx).WithValues("workload", klog.KObj(wl))
+		ch <- event.TypedGenericEvent[*kueue.Workload]{Object: wl}
+		log.V(4).Info("Sent DRA workload to reconcile channel due to LocalQueue creation")
+	}
+
```
mimowo, line 523: Wrap that piece of code by checking the DRA feature gate. I know `draWorkloads` is empty if the FG is disabled, but this is primarly to make all the injection about the DRA code feature easy to discover by the FG search.

pkg/cache/queue/manager_test.go
```diff
@@ -91,6 +93,63 @@ func TestAddLocalQueueOrphans(t *testing.T) {
...
+		Count: 1,
+		Template: corev1.PodTemplateSpec{
+			Spec: corev1.PodSpec{
+				ResourceClaims: []corev1.PodResourceClaim{{
+					Name:                      "rc",
+					ResourceClaimTemplateName: &tmplName,
+				}},
+			},
+		},
+	}}
```
mimowo: Use the WorkloadWrapper to construct the object

pkg/cache/queue/manager_test.go
```diff
@@ -91,6 +93,63 @@ func TestAddLocalQueueOrphans(t *testing.T) {
...
+	done := make(chan error, 1)
+	go func() {
+		done <- manager.AddLocalQueue(ctx, utiltestingapi.MakeLocalQueue("foo", "earth").ClusterQueue("cq").Obj())
+	}()
+
+	// Drain the pre-filled event to unblock AddLocalQueue's blocking send.
+	// The send must happen outside the manager lock, so draining here cannot deadlock.
+	select {
+	case <-ch:
+	case <-time.After(5 * time.Second):
```
mimowo: maybe let's util.Timeout, which is 10s by default, for standradizing the timeout use

pkg/cache/queue/manager.go
sohankunkerkar: I was planning to provide that feedback but ch can't be nil here  as the guard at `if m.draReconcileChannel != nil. It  prevents appending to draWorkloads when the channel is nil, so if we reach this loop, ch is guaranteed non-nil.

mimowo: I see, but that makes the code a bit fragile. Do we have cases when DRA is enabled but m.draChannel is nil? Maybe the other check above under https://github.com/kubernetes-sigs/kueue/pull/10785/changes#diff-c7ed976ece2ea8cfd72847c76a214be5920b16aa7770e4d3cef7d8c974a36028R489  is unnecessarily confusing.

sohankunkerkar: Yup, it looks like a leftover defensive guard. You approach sounds good!

gyliu513: Good catch. Agreed! The inner if m.draReconcileChannel != nil guard is redundant since draWorkloads is only populated and iterated when m.draReconcileChannel != nil is already guaranteed. I'll remove the redundant check to make the invariant more explicit.

pkg/cache/queue/manager.go
gyliu513: Yes, the channel is already a field on the manager, so passing it through the return value is unnecessary indirection. I'll simplify the signature to return only ([]*kueue.Workload, error) and access m.draReconcileChannel directly in AddLocalQueue.

pkg/cache/queue/manager.go
gyliu513: Makes sense. Wrapping with the DRA feature gate makes it easy to discover all DRA-related code paths via a feature gate search, even if the behavior is already correct without it. I'll add the check.

pkg/cache/queue/manager_test.go
gyliu513: Yes, I'll use MakePodSet(...).PodSpec(...) via the WorkloadWrapper instead of directly mutating the struct.

pkg/cache/queue/manager_test.go
gyliu513: test/util imports pkg/cache/queue, which creates a cycle when imported from this package's test. I've matched the value (10s) directly to align with util.Timeout. 

If you'd prefer, we could move the constant to pkg/util/testing where it's importable from here. But I'd prefer handle this in a separate PR.

gyliu513 (commented):
Thanks @mimowo @sohankunkerkar , comments are addressed

pkg/cache/queue/manager_test.go
mimowo: I see, in that case let's just keep it at 10s for now as is - hardcoded.

sohankunkerkar (approved):
/lgtm

/assign @mimowo

pkg/cache/queue/manager.go
```diff
@@ -498,6 +501,26 @@ func (m *Manager) AddLocalQueue(ctx context.Context, q *kueue.LocalQueue) error
...
+	if !features.Enabled(features.DynamicResourceAllocation) {
+		return nil
+	}
+
+	for _, wl := range draWorkloads {
+		log := ctrl.LoggerFrom(ctx).WithValues("workload", klog.KObj(wl))
+		m.draReconcileChannel <- event.TypedGenericEvent[*kueue.Workload]{Object: wl}
+		log.V(4).Info("Sent DRA workload to reconcile channel due to LocalQueue creation")
+	}
```
mimowo, line 522: nit, but I consider this pattern more robust:

```golang
if features.Enabled(features.DynamicResourceAllocation) {
  for _, wl := range draWorkloads {
		log := ctrl.LoggerFrom(ctx).WithValues("workload", klog.KObj(wl))
		m.draReconcileChannel <- event.TypedGenericEvent[*kueue.Workload]{Object: wl}
		log.V(4).Info("Sent DRA workload to reconcile channel due to LocalQueue creation")
	}
```
The reason is that if DynamicResourceAllocation is beta, then likely someone will add "non FG" code at the bottom without realizing that can get switched off when DynamicResourceAllocation FG is disabled.

mimowo (commented):
/lgtm
/approve
Let me merge as this is already solving the issue with prolonged locking, but please follow up on the remaining nit comment: https://github.com/kubernetes-sigs/kueue/pull/10785#discussion_r3256675595

mimowo:
This is a known flake: https://github.com/kubernetes-sigs/kueue/issues/11062

Issue #10765, opened 2026-04-24
Support a default WorkloadPriorityClass (analogous to PriorityClass globalDefault)


atosatto:
**What would you like to be added**:

A mechanism to designate a WorkloadPriorityClass as the cluster-wide default, so that workloads without an explicit `kueue.x-k8s.io/priority-class` label automatically receive its priority value instead of falling through to the Kubernetes `scheduling.k8s.io/v1` PriorityClass chain.

This could take the form of a `globalDefault` boolean field on WorkloadPriorityClass (mirroring the Kubernetes PriorityClass API), or a configuration option in the Kueue Configuration API. The behaviour should be opt-in and preserve the current fall back mechanism.

**Why is this needed**:

In clusters where only a subset of workloads are managed by Kueue, admins need a way to assign a `default` WorkloadPriorityClass to Kueue-managed workloads without requiring job authors to explicitly set the `kueue.x-k8s.io/priority-class` label on every job.

Today, if a Kueue-managed job omits that label, `ExtractPriority()` falls through to the Kubernetes `scheduling.k8s.io/v1 PriorityClass` chain — either from the PodSpec, the `globalDefault` PriorityClass, or a hardcoded 0. This conflates pod scheduling priority with queueing priority, which are intentionally separate concerns in clusters that use WorkloadPriorityClass.

The only workarounds are:

1. Require every job author to set the label — which is fragile and hard to enforce across teams.
2. Deploy a custom mutating webhook — which adds operational overhead just to inject a default value that Kueue could handle natively.

A default WorkloadPriorityClass would let Kueue set the priority automatically at admission time for workloads that don't specify one, keeping all Kueue-managed workloads in a consistent priority domain without burdening users or requiring external mutation infrastructure.

**Completion requirements**:

This enhancement requires the following artifacts:

- [ ] Design doc
- [x] API change
- [x] Docs update

The artifacts should be linked in subsequent comments.

mimowo:
I'm wondering if we could make it simpler by just using the name 'default' to indicate the global default as we did for LocalQueue defaulting? I would go with that simple solution. I guess they couldnt in the core k8a due to backwards compatibility, but for Kueue I think we can a bit relax this using feature gates for the transition period. The nice thing is that name provides automatic global validation that this is the only default

atosatto:
[...]

PR #10796, opened 2026-04-27
KEP-10765: WorkloadPriorityClass defaulting

atosatto:
#### What type of PR is this?

/kind kep

Adds KEP-10765 for WorkloadPriorityClassDefaulting, a new feature gate that automatically assigns a WorkloadPriorityClass named default to workloads that don't specify one via the `kueue.x-k8s.io/priority-class` label.

In clusters where only a subset of workloads are managed by Kueue, this avoids conflating pod scheduling priority with workload queueing priority for jobs that omit the label, without requiring custom mutating webhooks or per-job labeling by users.

The design follows the same convention-based pattern as LocalQueueDefaulting.

Related #10765 

```release-note
NONE
```

### DIFF
--- a/keps/10765-workload-priority-class-defaulting/README.md
+++ b/keps/10765-workload-priority-class-defaulting/README.md
@@ -18,7 +18,6 @@
     - [Integration Tests](#integration-tests)
   - [Graduation Criteria](#graduation-criteria)
 - [Alternatives](#alternatives)
-  - [Configuration-based default](#configuration-based-default)
   - [globalDefault field on WorkloadPriorityClass](#globaldefault-field-on-workloadpriorityclass)
 <!-- /toc -->
 
@@ -143,20 +142,26 @@ necessary to implement this enhancement.
 
 #### Unit Tests
 
-- `pkg/util/priority/priority_test.go`:
-  - Feature gate disabled: `default` WPC is not consulted.
-  - Feature gate enabled, `default` WPC exists: unlabeled workloads get its
-    priority.
-  - Feature gate enabled, `default` WPC does not exist: falls through to
-    Kubernetes PriorityClass chain.
-  - Explicit `kueue.x-k8s.io/priority-class` label takes precedence over the
-    `default` WPC.
+- `pkg/util/priority/priority_test.go` — `DefaultWorkloadPriorityClassExist`:
+  - `default` WorkloadPriorityClass exists: returns `true`.
+  - `default` WorkloadPriorityClass does not exist (different name): returns
+    `false`.
+  - No WorkloadPriorityClasses exist: returns `false`.
+
+- `pkg/controller/jobframework/defaults_test.go` —
+  `ApplyDefaultWorkloadPriorityClass`:
+  - Feature gate enabled, no label, `default` WPC exists: label
+    `kueue.x-k8s.io/priority-class: default` is set on the job.
+  - Feature gate disabled: label is not set.
+  - Label already set: existing label is preserved unchanged.
+  - `default` WPC does not exist: label is not set.
+  - Job's owner is managed by Kueue: label is not set.
 
 #### Integration Tests
 
 - Webhook/controller integration tests verifying that workloads created from
-  Jobs without a `WorkloadPriorityClass` label receive the `default` WPC
-  priority when the feature gate is enabled.
+  Jobs without a `kueue.x-k8s.io/priority-class` label receive the `default`
+  WPC priority when the feature gate is enabled.
 
 ### Graduation Criteria
 
--- a/keps/10765-workload-priority-class-defaulting/README.md
+++ b/keps/10765-workload-priority-class-defaulting/README.md
@@ -19,6 +19,7 @@
   - [Graduation Criteria](#graduation-criteria)
 - [Alternatives](#alternatives)
   - [globalDefault field on WorkloadPriorityClass](#globaldefault-field-on-workloadpriorityclass)
+  - [Defaulting in the Workload rather than in the Job webhook](#defaulting-in-the-workload-rather-than-in-the-job-webhook)
 <!-- /toc -->
 
 ## Summary
@@ -172,6 +173,8 @@ necessary to implement this enhancement.
 **Beta:**
 - Feature gate enabled by default.
 - Address feedback from Alpha usage.
+- Re-evaluate whether defaulting should move from the Job webhook to the
+  Workload level based on user feedback and integration complexity.
 
 **GA:**
 - Feature gate locked to true.
@@ -185,3 +188,23 @@ mirroring Kubernetes' `PriorityClass.GlobalDefault`. This was rejected in favor
 of the simpler convention-based approach for consistency with
 `LocalQueueDefaulting`. The convention-based approach requires no API changes and
 is simpler for both implementation and users to understand.
+
+### Defaulting in the Workload rather than in the Job webhook
+
+Instead of defaulting the `kueue.x-k8s.io/priority-class` label in the Job
+mutating webhook, the default `WorkloadPriorityClass` could be resolved at the
+time of Workload creation in the reconciler.
+
+This approach has some advantages: it would be implemented in a single place,
+agnostic of the actual Job type, avoiding code duplication across all job-type
+webhooks. It also avoids the webhook hot path and makes it straightforward to
+emit an event for debuggability when the default is applied.
+
+However, the webhook approach was chosen for Alpha to maintain consistency with
+`LocalQueueDefaulting`, which already sets the `kueue.x-k8s.io/queue-name`
+label via `ApplyDefaultLocalQueue` in the webhook. The webhook implementation is
+simple, already exercised by the existing `LocalQueueDefaulting` pattern, and
+provides immediate visibility to the user through the label on the Job object.
+
+This decision will be re-evaluated before Beta based on user feedback and the
+integration complexity observed during Alpha.
--- a/keps/10765-workload-priority-class-defaulting/README.md
+++ b/keps/10765-workload-priority-class-defaulting/README.md
@@ -126,9 +126,16 @@ object.
 
 This follows the same pattern as `LocalQueueDefaulting`, which sets
 `kueue.x-k8s.io/queue-name: default` via `ApplyDefaultLocalQueue` in the
-webhook. The new `ApplyDefaultWorkloadPriorityClass` function is called from all
-job-type webhooks (both `BaseWebhook` and custom webhooks) alongside
-`ApplyDefaultLocalQueue`.
+webhook. The new `ApplyDefaultWorkloadPriorityClass` function is called
+alongside `ApplyDefaultLocalQueue` from every job-type webhook:
+
+- Integrations that embed the shared `BaseWebhook` (e.g. `AppWrapper`,
+  Kubeflow's `JAXJob`/`PaddleJob`/`PyTorchJob`/`TFJob`/`XGBoostJob`)
+  inherit the call from `BaseWebhook.Default`.
+- Integrations with a custom webhook (`Pod`, `Deployment`, `Job`, `JobSet`,
+  `LeaderWorkerSet`, `MPIJob`, `RayCluster`, `RayJob`, `RayService`,
+  `SparkApplication`, `StatefulSet`, `TrainJob`) invoke it explicitly from
+  their own `Default` method.
 
 The label is set on the job at admission time, so the existing priority
 resolution chain in `ExtractPriority()` picks it up naturally during workload
--- a/keps/10765-workload-priority-class-defaulting/README.md
+++ b/keps/10765-workload-priority-class-defaulting/README.md
@@ -20,6 +20,7 @@
 - [Alternatives](#alternatives)
   - [globalDefault field on WorkloadPriorityClass](#globaldefault-field-on-workloadpriorityclass)
   - [Defaulting in the Workload rather than in the Job webhook](#defaulting-in-the-workload-rather-than-in-the-job-webhook)
+  - [Per-ClusterQueue default WorkloadPriorityClass](#per-clusterqueue-default-workloadpriorityclass)
 <!-- /toc -->
 
 ## Summary
@@ -227,3 +228,47 @@ provides immediate visibility to the user through the label on the Job object.
 
 This decision will be re-evaluated before Beta based on user feedback and the
 integration complexity observed during Alpha.
+
+### Per-ClusterQueue default WorkloadPriorityClass
+
+Instead of resolving defaulting through a cluster-scoped `WorkloadPriorityClass`
+named `default`, the default could be configured per-`ClusterQueue` via a new
+optional field on `ClusterQueueSpec`:
+
+```yaml
+apiVersion: kueue.x-k8s.io/v1beta2
+kind: ClusterQueue
+spec:
+  defaultWorkloadPriorityClass: gold
+```
+
+Resolution would still happen in the Job mutating webhook, so the defaulted
+priority remains visible on the Job object. Given the `kueue.x-k8s.io/queue-name`
+label (set by the user or by `ApplyDefaultLocalQueue`), the webhook would:
+
+1. Look up the `LocalQueue` and its bound `ClusterQueue` via the queue cache
+   (`Manager.ClusterQueueFromLocalQueue`, already used by
+   `ApplyDefaultForManagedBy`).
+2. Read `ClusterQueue.Spec.DefaultWorkloadPriorityClass`.
+3. If set and the Job has no `kueue.x-k8s.io/priority-class` label, stamp the
+   label with that value.
+
+**Advantages over the cluster-wide `default` WPC:**
+
+- Different `ClusterQueue`s can have different defaults (e.g., `interactive`
+  vs `batch`), aligning with how admins typically configure per-partition QoS
+  in systems like Slurm.
+- Operators can change the default `WorkloadPriorityClass` one `ClusterQueue`
+  at a time, so Jobs already stamped with the old default keep resolving
+  successfully while the new default is rolled out. With the cluster-wide
+  `default` WPC convention, deleting it leaves every already-stamped Job
+  unable to produce a Workload until the old WPC is restored or the label is
+  stripped.
+- Defaulting is opt-in per `ClusterQueue`. Operators can enable WPC defaulting
+  on some queues while other queues continue to fall through to
+  `PodSpec.priorityClassName` and the Kubernetes `PriorityClass` chain.
+
+**Disadvantages:**
+
+- Admins who want a single default for every `ClusterQueue` must set the
+  field on each one (or use their templating / policy tooling).
--- a/keps/10765-workload-priority-class-defaulting/README.md
+++ b/keps/10765-workload-priority-class-defaulting/README.md
@@ -1,0 +1,182 @@
+# KEP-10765: WorkloadPriorityClass defaulting
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Defaulting mechanism](#defaulting-mechanism)
+  - [Test Plan](#test-plan)
+    - [Unit Tests](#unit-tests)
+    - [Integration Tests](#integration-tests)
+  - [Graduation Criteria](#graduation-criteria)
+- [Alternatives](#alternatives)
+  - [Configuration-based default](#configuration-based-default)
+  - [globalDefault field on WorkloadPriorityClass](#globaldefault-field-on-workloadpriorityclass)
+<!-- /toc -->
+
+## Summary
+
+This KEP introduces a `WorkloadPriorityClassDefaulting` feature gate. When
+enabled, workloads that do not specify a `WorkloadPriorityClass` via the
+`kueue.x-k8s.io/priority-class` label automatically receive the priority value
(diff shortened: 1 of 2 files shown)


keps/10765-workload-priority-class-defaulting/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Proposal
+
+When the `WorkloadPriorityClassDefaulting` feature gate is enabled and a workload
+does not specify a `WorkloadPriorityClass`, the priority resolution chain looks
+for a `WorkloadPriorityClass` named `default` before falling through to the
+Kubernetes PriorityClass chain. This follows the same convention-based pattern as
+`LocalQueueDefaulting`, which looks for a LocalQueue named `default`.
+
+The `default` WorkloadPriorityClass itself must be created manually by the
```
mwielgus, line 74: What happens if the default WorkloadPriorityClass is not there?

keps/10765-workload-priority-class-defaulting/README.md
```diff
@@ -0,0 +1,179 @@
...
+The current priority resolution chain in `ExtractPriority()` is:
+
+1. Explicit `WorkloadPriorityClass` via `kueue.x-k8s.io/priority-class` label
+2. Custom priority class from the job framework (e.g., MPIJob's `schedulingPolicy.priorityClass`)
+3. Kubernetes PodSpec `priorityClassName`
+4. Kubernetes `globalDefault` PriorityClass
+5. Hardcoded `0`
+
+With this KEP, a new step is inserted between step 1 and step 2 when the
+`WorkloadPriorityClassDefaulting` feature gate is enabled:
```
mwielgus: Does Kueue provide any visibility for the user how the workload ended up with a specific priority? If not, can we have it added?

keps/10765-workload-priority-class-defaulting/README.md
atosatto: In that case, we will fall-back to the Kubernetes PodSpec `priorityClassName`.
The intended behaviour is captured in the " Design Details" section.

keps/10765-workload-priority-class-defaulting/README.md
atosatto: Similar to LocalQueueDefaulting the idea is to set the `kueue.x-k8s.io/priority-class` on the Job if the `default` WorkloadPriorityClass exists and the label has not been set by the Job owner. Is there any other any other mechanism you would suggest adopting in this case?

atosatto: I have pushed an update that hopefully should capture better the intended behaviour in the KEP document.

keps/10765-workload-priority-class-defaulting/README.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Design Details
+
+### Defaulting mechanism
+
+When the `WorkloadPriorityClassDefaulting` feature gate is enabled and a job is
+submitted without a `kueue.x-k8s.io/priority-class` label, the mutating webhook
+checks whether a `WorkloadPriorityClass` named `default` exists. If it does, the
+webhook sets the label `kueue.x-k8s.io/priority-class: default` on the job
+object.
```
mimowo, line 151: I'm not sure which approach is better, but the alternative one is to default the WPC at the moment of Workload creation. I remember this was discussed at the LocalQeueDefaulting, and we didn't do that for some reasons I cannot now recall. 

The good things about workload-level defaulting:
1. one place agnostic of the actual Job
2. not a hot path like webhooks, so we can easily afford to send an event for debuggability

Let me know your thoughts. Maybe for Alpha we could start at workload level and re-evaluate if we need webhooks.

cc @yaroslava-serdiuk do you maybe remember why LocalQueueDefaulting was done at the Job level?

atosatto: > I'm not sure which approach is better, but the alternative one is to default the WPC at the moment of Workload creation. I remember this was discussed at the LocalQeueDefaulting, and we didn't do that for some reasons I cannot now recall.
> Let me know your thoughts. Maybe for Alpha we could start at workload level and re-evaluate if we need webhooks.

I think workload level defaulting is a good alternative, but we would be changing the behaviour from what we have done with LocalQueue. Is there any specific concern with the webhook? Is this about scalability?

mimowo: > I think workload level defaulting is a good alternative, but we would be changing the behaviour from what we have done with LocalQueue. Is there any specific concern with the webhook? Is this about scalability?

Scalability (performance) is not my concern, because this check will be just cache-based. Rather the amount of testing and code duplication to ensure all integrations have it.

atosatto: There is an implementation here already - I feel it is reasonably self-contained which makes me comfortable with the webhook approach being ok.

https://github.com/kubernetes-sigs/kueue/pull/10798

Let me know what you think!

mimowo: lgtm thank you. We could also consider as a follow up 

```golang
jobframework.ApplyDefaultLocalQueue(job.Object(), w.queues.DefaultLocalQueueExist)
jobframework.ApplyDefaultWorkloadPriorityClass(ctx, w.client, job.Object())
```
In that case I'm ok with this approach, because the code is simple, already exercised by LocalQueueDefaulting.

atosatto: Are you suggesting any specific change to the proposed PR? Feel free to comment there whenever you get the chance. I will take care of implementing them as I rebase the code.

Also @mimowo could it be ok for me to add you as reviewer and approver for this?

mimowo: let's mention the alternative in the Alternatives section and add apoint to graduation criteria to re evaluate the approach before going to beta based on user feedback. So that we make it clear it may change likely it will not.

yes add me as an approver. add as reviewers other folks participating in to review too

atosatto: Thanks! Added the changes we discussed in e62c558. Let me know if you have any other feedback 🙏

mimowo:
Checking the latest version now

keps/10765-workload-priority-class-defaulting/README.md
```diff
@@ -0,0 +1,210 @@
...
+works. Administrators who want a descriptive name can create a
+`WorkloadPriorityClass` with their preferred name and instruct users to reference
+it via label; this feature only covers the case where no label is set at all.
+
+### Risks and Mitigations
+
+Enabling this feature changes the priority of workloads that previously fell
+through to the Kubernetes PriorityClass default. Admins should set the `default`
+WorkloadPriorityClass value carefully before enabling the feature gate to avoid
+unexpected changes in queueing and preemption behavior.
```
mimowo: Maybe let's rephase that slightly to emphasize:
- the semantics is changed in case the environment already contains WPC named "default" - other environemnts are not affected
To mitigate:
- feature gate will allow admins to disable the mechanism while in Beta in case their environments are affected negatively
- we will add information about this change in the release notes

mimowo (commented):
LGTM thank you 👍 I will give a bit more time to review by @tenzen-y, but green light from me. We have exercised the same pattern with LocalQueue defaulting and haven't encountered user issues.

keps/10765-workload-priority-class-defaulting/README.md
atosatto: Sounds good, addressed this in c56cc9695

keps/10765-workload-priority-class-defaulting/README.md
```diff
@@ -0,0 +1,222 @@
...
+When the `WorkloadPriorityClassDefaulting` feature gate is enabled and a job is
+submitted without a `kueue.x-k8s.io/priority-class` label, the mutating webhook
+checks whether a `WorkloadPriorityClass` named `default` exists. If it does, the
+webhook sets the label `kueue.x-k8s.io/priority-class: default` on the job
+object.
+
+This follows the same pattern as `LocalQueueDefaulting`, which sets
+`kueue.x-k8s.io/queue-name: default` via `ApplyDefaultLocalQueue` in the
+webhook. The new `ApplyDefaultWorkloadPriorityClass` function is called from all
+job-type webhooks (both `BaseWebhook` and custom webhooks) alongside
```
tenzen-y: Just checking about `custom webhooks`. Do we call this new `ApplyDefaultWorkloadPriorityClass` in Pod integration as well? That's why I am asking this is that Pod integration doesn't use `BaseWebhook`, AFAIK.

keps/10765-workload-priority-class-defaulting/README.md
```diff
@@ -0,0 +1,222 @@
...
+- The job already has a `kueue.x-k8s.io/priority-class` label.
+- The job's owner is already managed by Kueue (to avoid double-defaulting).
+- No `WorkloadPriorityClass` named `default` exists.
+
+When the `default` WorkloadPriorityClass does not exist, the webhook makes no
+changes to the job and the existing priority resolution chain applies: the
+workload priority falls through to the Kubernetes PodSpec `priorityClassName`,
+then to the Kubernetes `globalDefault` PriorityClass, and finally to the
+hardcoded value `0`. This is the same behavior as when the feature gate is
+disabled.
```
tenzen-y, line 181: AFAIK, admins can remove WorkloadPriorityClass (WPC) even when any Workload points to the WPC.

It's ok to align with K8S PriorityClass behavior, but I want to check if Workload creation will fail in the same way as today when `default` WPC doesn't exist during Workload creation (default WPC is removed after being stamped by webhooks).

keps/10765-workload-priority-class-defaulting/README.md
```diff
@@ -0,0 +1,222 @@
...
+
+## Design Details
+
+### Defaulting mechanism
+
+When the `WorkloadPriorityClassDefaulting` feature gate is enabled and a job is
+submitted without a `kueue.x-k8s.io/priority-class` label, the mutating webhook
+checks whether a `WorkloadPriorityClass` named `default` exists. If it does, the
+webhook sets the label `kueue.x-k8s.io/priority-class: default` on the job
+object.
```
tenzen-y, line 151: What if Job podTemplateSpec has Pod PriorityClass?

tenzen-y, line 151: Do we stamp `kueue.x-k8s.io/queue-name: default` label only in managed namespaces or all namespaces? And does this WorkloadPriorityClass defaulting have compatibility with manageJobsWithoutQueueName?

keps/10765-workload-priority-class-defaulting/kep.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - "@atosatto"
+status: implementable
+creation-date: 2026-04-27
+reviewers:
+  - "@vladikkuzn"
+  - "@mwielgus"
+  - "@tenzen-y"
+approvers:
+  - "@mimowo"
+
```
tenzen-y, line 14:
```suggestion
  - "@mimowo"
  - "@tenzen-y"
  
```

Just adding me to hew :)

tenzen-y (commented):
@atosatto Thank you for working on this KEP, and apologies for delaying my review.
I left some comments to clarify this feature specifications.

keps/10765-workload-priority-class-defaulting/README.md
```diff
@@ -0,0 +1,222 @@
...
+**Beta:**
+- Feature gate enabled by default.
+- Address feedback from Alpha usage.
+- Re-evaluate whether defaulting should move from the Job webhook to the
+  Workload level based on user feedback and integration complexity.
+
+**GA:**
+- Feature gate locked to true.
+
+## Alternatives
```
kannon92, line 227: In Slurm, they have a default setting at the partition level (ClusterQueue equivalent).

https://slurm.schedmd.com/qos.html#priority

I've been interested in having Kueue maybe consider ClusterQueue defaults or LocalQueue defaults for some of these values. Slurm has a concept of QoS which sets resources, priority, wall time limits.

Have we considered this in this design?

https://github.com/kubernetes-sigs/kueue/issues/7029

keps/10765-workload-priority-class-defaulting/README.md
atosatto: Good catch, thanks! For the Pod integration the call to `ApplyDefaultWorkloadPriorityClass` is wired explicitly from `PodWebhook.Default`, same as the other custom-webhook integrations (Deployment, Job, JobSet, LeaderWorkerSet, MPIJob, RayCluster, RayJob, RayService, SparkApplication, StatefulSet, TrainJob), as done already for `LocalQueueDefaulting`.

I've updated the KEP to enumerate which integrations inherit the call via `BaseWebhook.Default` and which invoke it from their own `Default` method, so it's explicit for future readers.

keps/10765-workload-priority-class-defaulting/README.md
atosatto: Yes, this aligns with today's behavior. The `default` WorkloadPriorityClass label is resolved at Job-creation time. When default WPC doesn't exist during Workload creation, the Get returns NotFound, the error propagates up, and the reconciler fails to create the Workload (and retries until the WPC is recreated or the label is removed).

That's identical to what happens today if a user sets `kueue.x-k8s.io/priority-class: foo` manually and an admin later deletes `foo`. This means that removing any WorkloadPriorityClass, including `default`, has the risk of impacting multiple Workloads within a cluster. In order to do this operation safely the operator would have to first disable the feature-flag and then remove the `default` WPC. This does not seem great if we intend to remove this flag ad part of the GA promotion criteria.

Following up on @kannon92 's comment here - one of the alternatives we might want to consider is switching the defaulting to the `ClusterQueue`. Will add this to the alternatives considered so that we can explore whether it's a much more desirable approach.

keps/10765-workload-priority-class-defaulting/README.md
atosatto: That's a great feedback! I have added this to the alternatives considered for now (see 40a6a07d7) but I feel this could be the way we deliver this feature.

@kannon92  / @tenzen-y  / @mimowo any feedback on making defaulting at the ClusterQueue vs LocalQueue level? I can see merits for both approaches. I am curious on whether there are specific use-cases that would be supported by one approach but not the other that you might be aware of.

keps/10765-workload-priority-class-defaulting/README.md
atosatto: Yes, when defaulting is enabled and a `default` WPC exists the webhook still stamps the `kueue.x-k8s.io/priority-class: default` label. Defaulting at the ClusterQueue or LocalQueue (as described in 40a6a07d7) gives us a more flexible approach to enable this behaviour on a more granular basis vs cluster-wide.

keps/10765-workload-priority-class-defaulting/kep.yaml
atosatto: Done in e7f121cab! 🙇

keps/10765-workload-priority-class-defaulting/README.md
atosatto: > Do we stamp kueue.x-k8s.io/queue-name: default label only in managed namespaces or all namespaces?

The current LocalQueue defaulting logic is set independently from the `managedJobsNamespaceSelector` as the label is set in the webhook before before `ApplyDefaultForSuspend`, which is where the namespace-selector check happens. For WorkloadPriorityClass defaulting we are following the same pattern.

> And does this WorkloadPriorityClass defaulting have compatibility with manageJobsWithoutQueueName?

The default WorkloadPriorityClass is set regardless on the `kueue.x-k8s.io/queue-name` label. so Jobs admitted under `manageJobsWithoutQueueName=true` should be defaulted as expected.

keps/10765-workload-priority-class-defaulting/README.md
kannon92: Thanks for adding that!

atosatto: @kannon92  do you have any suggestion on whether this should be a LocalQueue or ClusterQueue level setting?

kannon92: Let me ask the slurm folks. cc @catblade

kannon92: I'm not sure if priority should be namespaced setting or a global setting. I think slurm provides options for both but not sure what is most commonly used.

Based on feedback I got for time-limits I think ClusterQueue as that is where we put most of our configuration for ClusterQueues. LocalQueues are primarly a link to a CQ.

But I want to get some input on slurm community what they find the most used/useful.

mimowo: I think both global and CQ-level have some use-cases and are not mutually exclusive. 

Global default can be helpful for the stories which we already have, and corresponds to the `globalDefault` setting in the core k8s. 

Also, doing it global is no API commitment, whereas CQ-level entails new APIs. In other words: if we make global setting as Alpha, we can easily change our approach in the next release. However, if we choose to introduce new CQ level API fields, then the cost of deprecating the approach is much higher as it requires v1beta3 to drop the fields.

So I would suggest to follow YAGNI and start with global for now, then only work on CQ-level when we discover that this is needed.

mimowo:
LGTM, let me give some time for other reviewers in case they have additional comments

keps/10765-workload-priority-class-defaulting/README.md
atosatto: > Also, doing it global is no API commitment, whereas CQ-level entails new APIs. In other words: if we make global setting as Alpha, we can easily change our approach in the next release. However, if we choose to introduce new CQ level API fields, then the cost of deprecating the approach is much higher as it requires v1beta3 to drop the fields.

That's reasonable. I can definitely see us doing this at the CQ-level eventually. However I do agree that a less involved alpha phase will provide us with the use-cases and understanding to design that API properly, including the decision on whether we surface this at the CQ or LQ level.

tenzen-y:
> LGTM, let me give some time for other reviewers in case they have additional comments

ACK

keps/10765-workload-priority-class-defaulting/README.md
tenzen-y: The updated description looks great, thank you!

keps/10765-workload-priority-class-defaulting/README.md
tenzen-y: That sounds good, let's see a more comprehensive or complementary way in the next phase (beta).

keps/10765-workload-priority-class-defaulting/README.md
tenzen-y: I think that the core problem is that users typically have permissions to specify Pod PriorityClass.
So, I'm wondering if we should stamp `kueue.x-k8s.io/priority-class: default` only when Jobs don't have Pod PriorityClass.

But, I can understand that admins sometimes hope to enforce the default "WPC" at all.
So, can you add a new alternative section to describe the enforcement mode which has "strong" enforcement (your demand) and soft enforcement (only when the Job doesn't have Pod PriorityClass).

I don't request to describe a concrete design for the mode at this time, but I want to leave consideration to this KEP even at this time.

tenzen-y: That sounds good to me, can you add explanations for that to this KEP.
I think we can expand `Notes/Constraints/Caveats` section.

keps/10765-workload-priority-class-defaulting/README.md
tenzen-y: +1 on @mimowo 
We can revisit here at another time.

tenzen-y:
@atosatto Thank you for updating this KEP!
The continuing threads are https://github.com/kubernetes-sigs/kueue/pull/10796#discussion_r3228120659 and https://github.com/kubernetes-sigs/kueue/pull/10796#discussion_r3228141281

keps/10765-workload-priority-class-defaulting/README.md
kannon92: Yea, I'm happy with this for alpha or even GA. Just wanted to add an alternative.

keps/10765-workload-priority-class-defaulting/README.md
atosatto: > I think that the core problem is that users typically have permissions to specify Pod PriorityClass.
So, I'm wondering if we should stamp kueue.x-k8s.io/priority-class: default only when Jobs don't have Pod PriorityClass.

Added a new "Soft enforcement when a Pod PriorityClass is set" section under Alternatives that describes both enforcement modes and the rationale for starting with strong enforcement: in clusters with a mix of batch and non-batch workloads, Pod PriorityClass is typically used cluster-wide for quota and scheduling priority, and batch Jobs often still set one to participate in that system — those are exactly the Jobs where decoupling Kueue queueing priority from scheduler priority matters most, so skipping defaulting for them would undermine the feature.

I think this is totally something we should revisit based on user feedback as part of the Beta graduation criteria.

atosatto: Done. I have expanded Notes/Constraints/Caveats with two new subsections:

- Interaction with `managedJobsNamespaceSelector` — capturing the fact that the webhook sets the `kueue.x-k8s.io/priority-class: default` label on every Job that passes through it (including Jobs in excluded namespaces), mirroring `LocalQueueDefaulting` which runs at the same point in `BaseWebhook.Default` (before `ApplyDefaultForSuspend`).
- Compatibility with `manageJobsWithoutQueueName` — clarifying that defaulting does not depend on the `kueue.x-k8s.io/queue-name` label, so Jobs admitted under `manageJobsWithoutQueueName=true` receive the default WPC similarly to any other Kueue-managed Job.

atosatto:
> @atosatto Thank you for updating this KEP! The continuing threads are [#10796 (comment)](https://github.com/kubernetes-sigs/kueue/pull/10796#discussion_r3228120659) and [#10796 (comment)](https://github.com/kubernetes-sigs/kueue/pull/10796#discussion_r3228141281)

Thank you @tenzen-y for your review! I should have addressed all of the remaining feedback in the KEP. 🤞

tenzen-y (commented):
@atosatto Thank you for finalizing this KEP! This looks really great!
/lgtm
/approve

PR #10799, opened 2026-04-27
Document installation upgrade policy

### LINKED ISSUE #10489: Extend docs on installation to include the upgrade policy

VPC-byte:
Fixes #10489.

This adds an installation-page note explaining that patch-release upgrades should also review the corresponding .0 minor release upgrade notes, with a v0.16.x to v0.17.1 example.

### DIFF
--- a/site/content/en/docs/installation/_index.md
+++ b/site/content/en/docs/installation/_index.md
@@ -12,6 +12,7 @@ description: >
 - [Install a released version](#install-a-released-version)
   - [Install by kubectl](#install-by-kubectl)
   - [Install by Helm](#install-by-helm)
+  - [Upgrade policy](#upgrade-policy)
   - [Add metrics scraping for prometheus-operator](#add-metrics-scraping-for-prometheus-operator)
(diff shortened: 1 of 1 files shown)


mimowo:
/close
We don't accept PRs by AI agents directly, see: https://www.kubernetes.dev/docs/guide/pull-requests/#ai-guidance

PR #10797, opened 2026-04-27
Update SubtreeQuota post cohort deletion

mszadkow:
#### What type of PR is this?
/kind bug

#### What this PR does / why we need it:
Update the SubtreeQuota of all ancestors to reflect the removal of the deleted cohort. 
Without this, stale SubtreeQuota values would cause any subsequent RecordCohortMetrics call for ancestor cohorts to re-publish the pre-deletion metric values.

#### Which issue(s) this PR fixes:

Fixes #10417 #10771

#### Does this PR introduce a user-facing change?

```release-note
Scheduling: Fixed a bug in Kueue's cache that could leave stale SubtreeQuota values in ancestor cohorts after a child Cohort
was deleted, leading to potential over-admission of workloads and incorrect metrics reporting.
```

PR #10805, opened 2026-04-27
fix: add delay to fix flaky Cohort metrics test (#10417)

Bvq629:
This PR fixes the flaky test issue #10417 by adding a 100ms delay before metrics assertions in both cohort_metrics_test.go and cohort_controller_test.go. This ensures the scheduler has time to update metrics before we check them, eliminating the race condition.


mimowo:
/close
This is being actively worked on here: https://github.com/kubernetes-sigs/kueue/pull/10797.

PR #10806, opened 2026-04-27
Metrics on topological domain usage

### LINKED ISSUE #10505: Metrics on topological node assignment
I would like to emit a workload metric that tells me which nodes workloads are assigned to so that I can query something like "how many 64 GPU sized slots are there?" from Kueue's perspective per topology domain.

It could look something like:
kueue_tas_domain_usage{flavor="flavor-123", domain_id="host-123", resource="gpu"} 8

**Why is this needed**:
To understand Topology fragmentation, there's 2 tiers: 1.) k8s scheduler view and 2.) Kueue controller view

From the k8s scheduler view, you can sort of derive what Kueue may be seeing by mapping created pods to nodes and construct a query that looks something like: "how many 4 GPU sized slots are there?"

However there's potentially a difference between what Kueue sees and what the k8s scheduler sees. 
- Kueue could get to QuotaReserved state without admitting the workload. 
- People may also be confused from the k8s view where you see that there's slots available, but Kueue is still not admitting workloads because the quota has already been reserved (just the pods arent created yet)

I think this could be complicated to determine these slots when workloads within a flavor may satisfy different hierarchy levels. But I welcome discussion/brainstorming on how to get this overall result. 

**Completion requirements**:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

amy:
#### What type of PR is this?

/kind feature
/area tas

Metrics on domain usage. Ex:

kueue_tas_domain_usage{flavor="flavor-123", domain="hostname", domain_id="host-123", resource="gpu"} 8

Callout. I need to count both Kueue and non-Kueue workloads. The main purpose is to understand what Kueue is counting for resource usage (not what the k8s scheduler is thinking.)

Fixes #10505

### TAS Workload Domain lookup logic
When a workload is admitted, TAS writes a TopologyAssignment into wl.Status.Admission.PodSetAssignments[*].TopologyAssignment. This contains the topology Levels (the label keys, e.g. ["block", "rack", "hostname"]) and per-domain Values — but as a storage optimization, when hostname is the lowest level only the hostname value is stored in Values, omitting ancestor values.

At metrics time, the code takes that hostname value, looks it up in the in-memory node informer cache, and reads the full set of ancestor label values off the node object — reconstructing the complete Values slice to match all Levels. 

With the full path, it fans out one Prometheus gauge increment per topology level, so a single node assignment contributes a series at each of hostname, rack, and block.

### Track Non-TAS Usage Changes
(I need both TAS and non-TAS workloads aggregated into this metric so that we have a full view of what Kueue thinks is being used by the domain.)

- Pre-compute topology coordinates at pod add time (tas_non_tas_pod_cache.go, tas_cache.go): When a non-TAS pod is first added to the cache, the matching flavor/topology coordinates (flavor name, topology levels, node label values) are computed and stored alongside the pod entry. This decouples metric subtraction from the current state of nodesCache — if a node is marked unschedulable and evicted from the cache between pod add and pod delete (e.g. by a cluster autoscaler), the subtract still fires correctly.

- Fix Delete predicate for terminal-phase pods (non_tas_usage_controller.go): The previous predicate called belongsToNonTASCache, which returns false for Succeeded/Failed pods. If the informer skipped the Running→Succeeded Update event and delivered only the Delete event with the pod already in a terminal phase, the predicate would drop it and the pod's usage would never be subtracted. The new predicate filters only on NodeName and TAS annotation, so terminal-phase deletes are always reconciled. DeletePodByKey is idempotent, so double-removal is safe. (case caught during e2e)

```release-note
Added kueue_tas_domain_usage gauge metric (Alpha) to track TAS workloads and non-Kueue workload resource usage per topology domain
```
> AI Assistance Disclosure: This pull request was created with the assistance of AI


pkg/cache/scheduler/tas_nodes_cache.go
```diff
@@ -59,6 +59,14 @@ func (t *nodesCache) deleteWithoutLock(nodeName string) {
 	delete(t.nodes, nodeName)
 }
 
+// findByHostname returns the nodeInfo for the given kubernetes.io/hostname label value.
+// The nodes map is keyed by node name, which Kubernetes guarantees equals the hostname label.
```
amy, line 63: Review note: I assume this is true? Its true for our stuff and for kind.

pkg/workload/workload.go
```diff
@@ -650,9 +650,21 @@ func totalRequestsFromAdmission(wl *kueue.Workload) []PodSetResources {
...
+			// TASUsage() computes domain totals as SinglePodRequests × Count, so without
+			// this adjustment the TAS usage would ignore the reclaim reduction.
+			// Each domain count is scaled by the same reclaim ratio (countAfterReclaim /
+			// originalCount) as the overall pod set count above.
+			if setRes.TopologyRequest != nil {
+				for i := range setRes.TopologyRequest.DomainRequests {
+					dr := &setRes.TopologyRequest.DomainRequests[i]
+					dr.Count = int32(int64(dr.Count) * int64(countAfterReclaim) / int64(originalCount))
+				}
+			}
```
amy, line 679: Found this issue due to integration testing failures.

amy:
@mbobrovskyi @mimowo PTAL! This is ready for review 🙇🏻‍♀️

As discussed from wg-batch meeting, I am emitting the metric in an event driven way. The metric that is emitted is also respective to only the workload's topology and not the entire node cache.

mwielgus (commented):
How do I configure which levels of topology should have metrics and which have to large cardinality for that (nodes)?

site/data/featuregates/versioned_feature_list.yaml
```diff
@@ -369,6 +369,12 @@
     lockToDefault: false
```
mwielgus, line 381: Also update the metrics list in the documentation.

amy: done: https://github.com/kubernetes-sigs/kueue/pull/10806/changes#diff-4588beb5f417cee7a2f87e79b5afd2de6ab13a98a03fdafc3288838bb8532eb4R121

amy:
> How do I configure which levels of topology should have metrics and which have to large cardinality for that (nodes)?

@mwielgus Done: https://github.com/kubernetes-sigs/kueue/pull/10806/changes#diff-d3e4256c738cb14b6f16ee4d436a823060d522641f51017f10775d3cc4f725a8R238

test/integration/singlecluster/controller/core/workload_preemption_eviction_metrics_test.go
```diff
@@ -81,6 +81,7 @@ var _ = ginkgo.Describe("Workload eviction to pending metrics", ginkgo.Label("co
 			Request(corev1.ResourceCPU, "1").
 			Obj()
 		util.MustCreate(ctx, k8sClient, lowWl)
+		util.ExpectWorkloadsToHaveQuotaReservation(ctx, k8sClient, cq.Name, lowWl)
```
amy, line 84: Will remove/rebase this once this is merged: https://github.com/kubernetes-sigs/kueue/pull/10879

It was causing test failures due to incorrect ordering for quota reservation.

amy:
cc/ @mbobrovskyi if you have time 🙏 
Its a beefy one.

pkg/controller/tas/non_tas_usage_controller.go
```diff
@@ -109,7 +108,11 @@ func (r *NonTasUsageReconciler) Update(e event.TypedUpdateEvent[*corev1.Pod]) bo
 }
 
 func (r *NonTasUsageReconciler) Delete(e event.TypedDeleteEvent[*corev1.Pod]) bool {
-	return belongsToNonTASCache(e.Object)
+	// Don't filter on terminal phase: if the informer skips the Running→Terminated
+	// Update and delivers only the Delete event with the pod already Succeeded/Failed,
+	// the pod's usage would never be removed from the cache. DeletePodByKey is
+	// idempotent, so double-removal is safe.
+	return len(e.Object.Spec.NodeName) > 0 && !utiltas.IsTAS(e.Object)
```
mimowo: Hold on, this seems totally unrelated. We may consider separately if this is a good or bad change in another PR.

apis/config/v1beta2/configuration_types.go
```diff
@@ -229,6 +234,16 @@ type LocalQueueMetrics struct {
...
 
+// TASMetrics defines configuration options for TAS domain usage metrics.
+type TASMetrics struct {
+	// ExcludedTopologyLevels lists topology level label keys to exclude from
+	// kueue_tas_domain_usage metrics (e.g. "kubernetes.io/hostname" to avoid
+	// high-cardinality node-level series in large clusters).
+	// +optional
+	// +listType=set
+	ExcludedTopologyLevels []string `json:"excludedTopologyLevels,omitempty"`
+}
```
mimowo, line 245: Since this introduces API level changes (config API) I would like to have a KEP for that.

amy: @mimowo ah shoot. Alrighty. Was hoping to get this in for the upcoming release. Let's still try anyways? Will write a kep asap

pkg/controller/tas/non_tas_usage_controller.go
amy: I specifically need this to catch the event changes for non-tas pods. Without it, we dont catch non-tas pod deletion events in e2e tests.

apis/config/v1beta2/configuration_types.go
amy: @mimowo PTAL KEP: https://github.com/kubernetes-sigs/kueue/pull/10998

pkg/controller/tas/non_tas_usage_controller.go
amy: @mimowo ok separated PTAL: https://github.com/kubernetes-sigs/kueue/pull/11033

I'll remove it from this PR after that one is merged because its required for tests to pass for this PR.

Issue #8848, opened 2026-01-28
Add support for OTLP metrics export


kimminw00:
**What would you like to be added**:
I would like to request support for exporting Kueue metrics via the **OpenTelemetry Protocol (OTLP)**.

Currently, Kueue only supports exposing metrics via the Prometheus format (pull model). I propose adding an option to configure an OTLP exporter so that Kueue can directly **push** metrics to an OpenTelemetry Collector or other OTLP-compatible backends.

**Why is this needed**:

While Prometheus is widely used, relying solely on the pull-based model has limitations in certain environments. OTLP support is needed for the following reasons:

1.  **Push-based Architecture**: In environments with strict network policies (e.g., firewalls preventing inbound scraping) or complex mesh setups, pushing metrics via OTLP is often easier to manage than configuring Prometheus scrapers.
2.  **Ecosystem Compatibility**: OpenTelemetry is the de-facto standard for observability. OTLP support allows users to integrate Kueue with a wide range of tools (Datadog, New Relic, Dynatrace, etc.) and OTel Collectors.

**Completion requirements**:

This enhancement requires the following artifacts:

- [ ] Design doc
- [X] API change
- [X] Docs update

The artifacts should be linked in subsequent comments.

mimowo:
This sounds great @kimminw00 ! Do you know if there are some projects in the k8s ecosystem which already support OTLP we could use for guiding the decisions for implementation? For example what libraries to use, what are the useful config params etc.

yashnib:
@mimowo One Kubernetes-native prior art we can use is **Kubernetes system component tracing**. K8s components emit telemetry via **OTLP over gRPC** and recommend routing through an **OpenTelemetry Collector** (Collector as the aggregation point, rather than pushing directly to arbitrary vendor backends). This is traces rather than metrics, but it’s a strong reference for the *shape* of configuration we could follow in Kueue: specify an OTLP endpoint and choose secure transport (TLS vs insecure). 

[...]

PR #10818, opened 2026-04-29
[Expiriment] Tracing

vladikkuzn:
This pull request adds **optional OpenTelemetry tracing** for the **Workload** controller so operators and developers can see **reconcile spans** and **high-level workload status transitions** (pending → quotaReserved → admitted → finished, etc.) in a backend such as **Jaeger**, instead of inferring behavior only from logs and metrics.
- **Configuration:** `tracing.enable` and optional `tracing.samplingRatio` on the Kueue `Configuration` CRD (controller manager section). When tracing is off, behavior stays **no-op** (no exporter overhead).
- **Export:** OTLP over **gRPC** using standard **`OTEL_EXPORTER_OTLP_*`** environment variables on the controller Deployment.
- **Correlation:** When sampling, the controller may persist **W3C trace context** and **last observed status** on **Workload** annotations so traces can continue across reconciles and record time spent in each summary status.
- **Docs and hacks:** English and Chinese reference pages under **OpenTelemetry tracing**, plus **`hack/tracing/`** (local Jaeger via Docker Compose, patch script for kind/dev clusters, optional demo manifests).
This is intentionally **experimental / demonstration** quality: it improves **visibility into status transition timing** for Workloads; design and scope can be refined in follow-ups.

Fixes #

- Some ideas and direction draw on exploration in [**agent-sandbox**](https://github.com/kubernetes-sigs/agent-sandbox); this PR is a **Kueue-specific** slice for Workload tracing only.
- **Draft / WIP:** not proposing this as the final observability model; feedback welcome on **API shape**, **annotation semantics**, **sampling defaults**, and **e2e/dev workflows**.

**Try tracing locally (kind + e2e-style run)**

1. Start Jaeger (OTLP gRPC **4317**, UI **http://localhost:16686**):

   ```bash
   docker compose -f hack/tracing/docker-compose.yaml up -d
   ```

2. Build the controller image, deploy into the test cluster, and run e2e with the OTLP patch and dev mode:

   ```bash
   export KUEUE_OTLP_PATCH=true
   export E2E_MODE=dev
   make kind-image-build test-e2e
   ```

   `KUEUE_OTLP_PATCH=true` runs `hack/tracing/patch-kueue-for-local-jaeger.sh` after deploy (enables tracing in manager config and sets `OTEL_*` on the controller; default endpoint targets **`host.docker.internal:4317`** where applicable). See **`hack/tracing/README.md`** and **`site/content/en/docs/reference/tracing.md`** for networking caveats (pod loopback vs host, kind/Docker Desktop).
   
   image

```release-note

```


amy:
This is cool. Would be interested in seeing any examples of this helping debug something or coming to some analysis conclusion.

mimowo:
@amy @vladikkuzn @kimminw00 how is this related to the issue we have: https://github.com/kubernetes-sigs/kueue/issues/8848? Is it addressing the issue requirements?

vladikkuzn:
> @amy @vladikkuzn @kimminw00 how is this related to the issue we have: https://github.com/kubernetes-sigs/kueue/issues/8848? Is it addressing the issue requirements?

@mimowo The https://github.com/kubernetes-sigs/kueue/issues/8848 is about usage of different protocol for metrics (OTLP instead of Prometheus), this PR is about adding traces using OTEL which is slightly covered by https://github.com/kubernetes-sigs/kueue/issues/8848

My intention is to give a way to answer a question "How much time workload spend in some state, and when and why it was transitioned to the other status", also it's useful for debugging if you want to understand lifecycle of some workload

mimowo:
This sounds like what also @amy is looking into

amy:
> This sounds like what also @amy is looking into

This could be cool! I'm less familiar with Jaeger and OTLP. So would love if people shared their findings so I can copy the methodology.

vladikkuzn:
I didn't know k8s has traces already, so I think we can go this direction as well as it's much easier to debug then

On the other hand consumers must be cost aware as it costs much more then metrics

PR #10821, opened 2026-04-29
Fix ClusterQueue finalizer after admission failure

### LINKED ISSUE #10759: CQ may stay undeleted due to a finalizer after deletion of all workloads and LQs
**What happened**:

See details in the issue  https://github.com/kubernetes-sigs/kueue/issues/10321 and the PR analysis: https://github.com/kubernetes-sigs/kueue/pull/10753

Deletion of CQ is not blocked if all workloads and LQs are deleted.

One need to delete all workloads, while they are evicting, and delete the LQ soon after. 

See more details in the analysis in https://github.com/kubernetes-sigs/kueue/pull/10753

ShaanveerS:
#### What type of PR is this?
/kind bug

When admission fails, the scheduler already removes the assumed Workload from its cache. The fix is just to also notify the ClusterQueue reconciler after that cleanup, so it can recheck the ClusterQueue and remove the finalizer once it's empty.

Fixes #10759

```release-note
Fix a race-condition bug that a deleted ClusterQueue may be kept by a finalizer, even after deletion of all workloads and LQs.
```

### DIFF
--- a/pkg/controller/core/core.go
+++ b/pkg/controller/core/core.go
@@ -109,6 +109,7 @@ func SetupControllers(mgr ctrl.Manager, qManager *qcache.Manager, cc *schdcache.
 	}
 	qManager.AddTopologyUpdateWatcher(cqRec)
 	qManager.AddWorkloadUpdateWatcher(qRec)
+	qManager.AddWorkloadUpdateWatcher(cqRec)
 
 	return "", nil
 }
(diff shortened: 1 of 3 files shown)


mimowo:
/release-note-edit
```release-note
Fix a race-condition bug that a deleted ClusterQueue may be kept by a finalizer, even after deletion of all workloads and LQs.
```
Nice one, thank you 👍

PR #10829, opened 2026-04-29
[TrainJob] Don't patch the TrainingRuntime with the queue label in th…

### LINKED ISSUE #10099: TrainJob integration: investigate  if we still need setting queue-name for JobSet (and cleanup if we can)

I would like to cleanup this code https://github.com/kubernetes-sigs/kueue/blob/bff02117c208902c363479cef3f42db1abbc9b3b/pkg/controller/jobs/trainjob/trainjob_webhook.go#L84-L89

If the code is needed we need at least some commentary explaining why this is needed.

**Why is this needed**:

I suspect this code is unnecessary because we don't create Workload at the JobSet level, but at the TrainJob level. 

I might be legacy code if we created Workloads at the JobSet level.

kaisoz:
#### What type of PR is this?

/kind cleanup
/area integrations

The trainjob mutating webhook was unnecessarily patching the TrainingRuntime with the queue label. That's not necessary anymore now that the integration is not accessing the underlying Jobset aymore

Fixes #10099

```release-note
NONE
```

### DIFF
--- a/pkg/controller/jobs/trainjob/trainjob_webhook.go
+++ b/pkg/controller/jobs/trainjob/trainjob_webhook.go
@@ -20,7 +20,6 @@ import (
 	"context"
 
 	kftrainerapi "github.com/kubeflow/trainer/v2/pkg/apis/trainer/v1alpha1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


mimowo:
/lgtm
/approve
Thank you!

PR #10853, opened 2026-04-30
Forcefully delete Failed/Succeeded pods that are stuck on unreachable nodes

### LINKED ISSUE #10847: PodFailureRecovery does not delete Pods which are already in Failed which blocks JobSet's Job deletion
**What happened**:

JobSet's Job foreground deletion was blocked by the stuck "Failed" Pod.

Forcefully delete such a pod when the mechanism kueue.x-k8s.io/safe-to-forcefully-delete is enabled.

This seems like a tricky race condition. 

In our environment the Kubelet was unhealty. It seems it transitioned the Pod into the Failed phase, but didn't finalize with the forceful delete request at the end.

kshalot:
#### What type of PR is this?
/kind bug

This handles the case where the pod is already in a final phase when the kubelet goes down. This unblocks JobSet's that delete Job's with foreground cascade.

Fixes #10847

I think this will also be an issue when the pod is in Succeeded phase (which IMO can happen), so I handled both.

**AI Assisted:** I made the changes to tests with Gemini.

```release-note
FailureRecovery: Forcefully delete pods that are Failed/Succeeded and scheduled on unreachable nodes.
This unblocks cases like a JobSet deleting a Job with foreground cascade being stuck because a pod in a terminal phase exists on one of the unhealthy nodes.
```

### DIFF
--- a/pkg/controller/failurerecovery/pod_termination_controller.go
+++ b/pkg/controller/failurerecovery/pod_termination_controller.go
@@ -41,7 +41,6 @@ import (
 	"sigs.k8s.io/kueue/pkg/controller/core"
 	"sigs.k8s.io/kueue/pkg/controller/tas/indexer"
 	utilclient "sigs.k8s.io/kueue/pkg/util/client"
-	utilpod "sigs.k8s.io/kueue/pkg/util/pod"
 	"sigs.k8s.io/kueue/pkg/util/roletracker"
 	utiltaints "sigs.k8s.io/kueue/pkg/util/taints"
 )
@@ -204,21 +203,7 @@ func podEligibleForTermination(p *corev1.Pod) bool {
 		return false
 	}
 
-	if p.DeletionTimestamp.IsZero() {
-		return false
-	}
-
-	// Do not filter out partially terminated (condition set, but not deleted) pods to handle
-	// partial executions of the controller caused by errors (for example network errors).
-	return isPodPartiallyForcefullyTerminated(p) || !utilpod.IsTerminated(p)
-}
-
-func isPodPartiallyForcefullyTerminated(p *corev1.Pod) bool {
-	return utilpod.HasCondition(p, &corev1.PodCondition{
-		Type:   KueueFailureRecoveryConditionType,
-		Reason: KueueForcefulTerminationReason,
-		Status: corev1.ConditionTrue,
-	})
+	return !p.DeletionTimestamp.IsZero()
 }
 
 func (r *TerminatingPodReconciler) mapNodeToPods(ctx context.Context, node *corev1.Node) []ctrl.Request {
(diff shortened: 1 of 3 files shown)


pkg/controller/failurerecovery/pod_termination_controller_test.go
```diff
@@ -323,9 +336,22 @@ func TestReconciler(t *testing.T) {
 			wantErr:    nil,
 			wantPod: podToForcefullyDelete.
 				Clone().
-				StatusPhase(corev1.PodSucceeded).
+				StatusPhase(corev1.PodFailed).
```
kshalot: Ok, I just noticed that this is probably incorrect. When the pod is Succeeded, we should delete but not change the phase.

mimowo: Yes, let's don't flip the phase.

pkg/controller/failurerecovery/pod_termination_controller.go
```diff
@@ -174,7 +174,9 @@ func (r *TerminatingPodReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	)
 
 	err := utilclient.PatchStatus(ctx, r.client, pod, func() (bool, error) {
-		pod.Status.Phase = corev1.PodFailed
+		if !utilpod.IsTerminated(pod) {
+			pod.Status.Phase = corev1.PodFailed
+		}
```
kshalot, line 179: @mimowo Like I said, I wanted to add handling for Succeeded which to me sounds analogous, so this check will prevent us from overwriting the final phase of the pod (in practice overwriting Succeeded with Failed).

But this also makes me think about the naming of the conditions and events - since sometimes we forcefully terminate + delete, sometimes only forcefully delete. Maybe we can revisit that in a follow-up

kshalot: My point is - if the pod is already in terminal phase and we add an event that it was "forcefully terminated" it might mislead someone when debugging into thinking that it was moved to this terminal phase by this controller.

mimowo: What the controller does is: "ensure terminated, then forcefully delete when the node is unreachable". So, part of the story is "ensuring termination". However, since the annotation is named now forece-delete, then we may also put more emphasizis on the "deletion" step in the condition/ event and controller naming.

mimowo: I propose to indeed make such a follow up. Since the controller is behind Alpha feature gate we don't have any k8s policy obligations on backwards compatibility, and to my knowledge changing the condition name is not something our users rely on. So, I'm ok to do this for 0.18+

mimowo (commented):
Thank you 👍 
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16

PR #10869, opened 2026-04-30
Fix prepare_pull.sh for patch releases.

### LINKED ISSUE #9867: Release automation: submit the PR to update only changelog for the N-2 patch

I would like to extend our release automation to generate and submit PR for the N-2 patch release.

Basically a PR like this https://github.com/kubernetes-sigs/kueue/pull/9865

And reference in the release issue 

Image

Maybe it could be `./hack/releasing/prepare_pull.sh --target changelog $VERSION`

Or we could also rename `--target` to `--mode`, but no need.

**Why is this needed**:

To save some time when doing the N-2 release.

mbobrovskyi:
#### What type of PR is this?
/kind bug

Fix prepare_pull.sh for patch releases.

Fixes #9867

I tested it in local branch:

1. Release 0.16.8:
https://github.com/mbobrovskyi/kueue/issues/182
https://github.com/mbobrovskyi/kueue/pull/244
https://github.com/mbobrovskyi/kueue/pull/245

2. Release 0.17.3:
https://github.com/mbobrovskyi/kueue/issues/183
https://github.com/mbobrovskyi/kueue/pull/246
https://github.com/mbobrovskyi/kueue/pull/247
--skip-version-updates:
https://github.com/mbobrovskyi/kueue/pull/250
https://github.com/mbobrovskyi/kueue/pull/251

3. Release 0.18.0:
https://github.com/mbobrovskyi/kueue/issues/181
https://github.com/mbobrovskyi/kueue/pull/248
https://github.com/mbobrovskyi/kueue/pull/249

```release-note
NONE
```

### DIFF
--- a/.github/ISSUE_TEMPLATE/NEW_RELEASE.md
+++ b/.github/ISSUE_TEMPLATE/NEW_RELEASE.md
@@ -22,7 +22,7 @@ Please do not remove items from the checklist
 - [ ] Update the release branch:
   - [ ] Ensure there are no unstaged changes in your directory (the script adds everything)
   - [ ] Run `./hack/releasing/prepare_pull.sh --target release $VERSION`
-  - [ ] Wait for this PR to merge <!-- PREPARE_PULL --> <!-- example #211 #214 -->
+  - [ ] Wait for this PR to merge <!-- PREPARE_PULL_RELEASE --> <!-- example #211 -->
 - [ ] An OWNER creates a signed tag
   - [ ] pull the release branch after PR from previous step merged
   - [ ] run
@@ -51,12 +51,10 @@ Please do not remove items from the checklist
 - [ ] Run the [openvex action](https://github.com/kubernetes-sigs/kueue/actions/workflows/openvex.yaml) to generate openvex data. The action will add the file to the release artifacts.
 - [ ] Run the [SBOM action](https://github.com/kubernetes-sigs/kueue/actions/workflows/sbom.yaml) to generate the SBOM and add it to the release.
 - [ ] Update the `main` branch :
-  - [ ] for each release:
-    - [ ] update the changelog <!-- example #9578 -->
-  - [ ] for only the latest release:
-    - [ ] Run `./hack/releasing/prepare_pull.sh --target main $VERSION`
-    - [ ] Submit a pull request with the changes: <!-- example #3007 -->
-    - [ ] Cherry-pick the pull request onto the `website` branch
+  - [ ] Run `./hack/releasing/prepare_pull.sh --target main $VERSION`
+        Note: Add --skip-version-updates if a newer minor or major version is already out.
+  - [ ] Wait for this PR to merge <!-- PREPARE_PULL_MAIN --> <!-- example #214 -->
+  - [ ] Cherry-pick the pull request onto the `website` branch
 - [ ] For major or minor releases, merge the `main` branch into the `website` branch to publish the updated documentation.
 - [ ] Send an announcement email to `sig-scheduling@kubernetes.io` and `wg-batch@kubernetes.io` with the subject `[ANNOUNCE] kueue $VERSION is released`.   <!--Link: example https://groups.google.com/a/kubernetes.io/g/wg-batch/c/-gZOrSnwDV4 -->
 - [ ] For a major or minor release, prepare the repo for the next version:
--- a/hack/releasing/prepare_pull.sh
+++ b/hack/releasing/prepare_pull.sh
@@ -32,15 +32,22 @@ FORK_REMOTE=${FORK_REMOTE:-origin}
 # shellcheck source=hack/utils.sh
 source "${REPO_ROOT}/hack/utils.sh"
 
-MAIN_REPO_ORG=${MAIN_REPO_ORG:-$(get_repo_org "$(git remote get-url "$UPSTREAM_REMOTE")")}
-MAIN_REPO_NAME=${MAIN_REPO_NAME:-$(get_repo_name "$(git remote get-url "$UPSTREAM_REMOTE")")}
+MAIN_REMOTE_URL="$(git remote get-url "$UPSTREAM_REMOTE")"
+MAIN_REPO_ORG=${MAIN_REPO_ORG:-$(get_repo_org "$MAIN_REMOTE_URL")}
+MAIN_REPO_NAME=${MAIN_REPO_NAME:-$(get_repo_name "$MAIN_REMOTE_URL")}
 
 TARGET=${TARGET:-all}
 POSITIONAL_ARGS=()
+SKIP_VERSION_UPDATES=false
 
 # Parse flags (can appear anywhere in argument list)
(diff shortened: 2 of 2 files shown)


mimowo:
Thank you for the work, but the current output does not look correct, because https://github.com/mbobrovskyi/kueue/pull/207 does not include the update to website etc. This is actually the "latest" release currently. v0.18.0 is future release.

I'm wondering it might be easier to just have a param like "--latest" so that we can indicate the latest manually so I could call `./hack/releasing/prepare_pull.sh --target main $VERSION` for 0.16.7, or `./hack/releasing/prepare_pull.sh --target main --latest $VERSION` for 0.17.2

I'm ok with the script to be smart, but it seems like the current conditions are not good.

For titles I would suggest:
- Update changelog for v0.16.7
- Update main with the latest v0.17.3

mbobrovskyi:
> Thank you for the work, but the current output does not look correct, because https://github.com/mbobrovskyi/kueue/pull/207 does not include the update to website etc. This is actually the "latest" release currently. v0.18.0 is future release.

Yeah, you right. Fixed this.

mimowo (commented):
Thank you 👍 This will be very useful to save time while doing releases
/lgtm
/approve
in case @tenzen-y has additional comments
/hold

hack/releasing/prepare_pull.sh
```diff
@@ -37,10 +37,15 @@ MAIN_REPO_NAME=${MAIN_REPO_NAME:-$(get_repo_name "$(git remote get-url "$UPSTREA
...
 POSITIONAL_ARGS=()
+LATEST=false
 
 # Parse flags (can appear anywhere in argument list)
 while [[ $# -gt 0 ]]; do
   case "$1" in
+    --latest)
+      LATEST=true
+      shift
+      ;;
```
tenzen-y: Instead of this flag, can we identify the latest release with `gh release list` command?
In that case, we can keep `./hack/releasing/prepare_pull.sh --target main $VERSION` both for the latest and non latest versions.

```shell
$ gh release list --repo https://github.com/kubernetes-sigs/kueue --json name,isLatest
[
  {
    "isLatest": true,
    "name": "v0.17.2"
  },
  {
    "isLatest": false,
    "name": "v0.16.7"
  },
  {
    "isLatest": false,
    "name": "v0.18.0-rc.0"
  },
  {
    "isLatest": false,
    "name": "v0.17.1"
  },
  ...
]
```

.github/ISSUE_TEMPLATE/NEW_RELEASE.md
```diff
@@ -51,12 +51,9 @@ Please do not remove items from the checklist
 - [ ] Run the [openvex action](https://github.com/kubernetes-sigs/kueue/actions/workflows/openvex.yaml) to generate openvex data. The action will add the file to the release artifacts.
 - [ ] Run the [SBOM action](https://github.com/kubernetes-sigs/kueue/actions/workflows/sbom.yaml) to generate the SBOM and add it to the release.
 - [ ] Update the `main` branch :
-  - [ ] for each release:
-    - [ ] update the changelog <!-- example #9578 -->
-  - [ ] for only the latest release:
-    - [ ] Run `./hack/releasing/prepare_pull.sh --target main $VERSION`
-    - [ ] Submit a pull request with the changes: <!-- example #3007 -->
-    - [ ] Cherry-pick the pull request onto the `website` branch
```
tenzen-y, line 59: Auto latest detection also allows us to keep these 2-phase TODO lists.

tenzen-y (commented):
@mbobrovskyi Thank you for moving this forward!

hack/releasing/prepare_pull.sh
mbobrovskyi: Will this work with the v0.18.0 release? When we create a new release, will it be considered the latest? I’m just worried that we might create multiple releases at the same time (e.g., v0.17.3 and v0.18.0), and I’m not sure how that will behave.

mimowo: I think in most cases the automated way will work well. Actually, it nicely skips  "v0.18.0-rc.0".

However, I'm not sure it will work well in all cases. For example, when we are releasing 0.17.6 and 0.18.0 on the same day, then the automated selection for 0.17.6 may choose "true", because it is currently "latest", but we know that 30min later we will release 0.18.0. So probably we can just generate the changelog for 0.17.6.

So, I'm wondering if we could use the automated way by default, but just have a param like `--force-latest y/n` true/false that would allow us to force which generation option to use.

Then we don't need to even put the param in the "new release notes", but keep it as an option.

mbobrovskyi: Actually, I think we may need to disable the "latest" flag. If we have v0.18.0 and v0.17.3 at the same time, we should explicitly set v0.18.0 as latest and disable it for v0.17.3. So I’m not sure automation is the better approach here.

mimowo: > Actually, I think we may need to disable the "latest" flag.

What do you mean by disabling the flag - I'm not following here.

> . If we have v0.18.0 and v0.17.3 at the same time, we should explicitly set v0.18.0 as latest and disable it for v0.17.3.

That should work well based on the API proposed in https://github.com/kubernetes-sigs/kueue/pull/10869#discussion_r3172456460, because at the moment we execute `./hack/releasing/prepare_pull.sh --target main $VERSION` then the "latest" release will be already published. The only exception I can see is if we are releasing 0.17.x, and then soon later 0.18.0, and so 0.18.0 is not published yet. In this case maybe we want to skip updating main with 0.17.x, because anyway soon (say 30min) later we will update with 0.18.0.

So, I was saying maybe `--force-latest` is useful for this case. However, I don't have a strong view here.

mbobrovskyi: > The only exception I can see is if we are releasing 0.17.x, and then soon later 0.18.0, and so 0.18.0 is not published yet. In this case maybe we want to skip updating main with 0.17.x, because anyway soon (say 30min) later we will update with 0.18.0.

Yes, that's my case. Do you mean that `--force-latest` skips version updates? If so, maybe we should rename this flag?

mimowo: I was thinking about saying y/n to the param, but maybe this is confusing. Maybe we just name it '--skip-version-updates'. So by default we follow the latest detection but this param allows to skip if we anticipate another release soon.

mbobrovskyi: Done

mbobrovskyi: Also updated PR description.

.github/ISSUE_TEMPLATE/NEW_RELEASE.md
mbobrovskyi: I’m not entirely sure about this. The algorithm is the same; the only difference is the --skip-version-updates flag. We probably just need to add a note about it.

mimowo (commented):
/lgtm
Thank you 👍

PR #10874, opened 2026-04-30
Handle NotFound errors during kueue preemption gracefully

jzhaojieh:
#### What type of PR is this?

/kind bug

When the controller restarts, the cache is rebuilt by listing all Workloads from the API server. Some entries returned by that List may be deleted shortly after by external controllers (job controllers, deactivation, etc.), so the cache can transiently contain Workloads that no longer exist on the API server. 

When IssuePreemptions selected such a Workload as a preemption target, the eviction patch returned 404 NotFound. The deleted Workload remained in Cache.workloadAssignedQueues, so on the next scheduling cycle it was re-selected as a target and re-failed in the same way — an infinite loop that pinned quota and starved real work.

This change treats NotFound as a terminal, successful preemption inside IssuePreemptions:

```release-note
Handle NotFound errors during kueue preemption gracefully
```

### DIFF
--- a/pkg/scheduler/preemption/preemption.go
+++ b/pkg/scheduler/preemption/preemption.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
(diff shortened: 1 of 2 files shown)


mimowo:
> When IssuePreemptions selected such a Workload as a preemption target, the eviction patch returned 404 NotFound. The deleted Workload remained in Cache.workloadAssignedQueues, so on the next scheduling cycle it was re-selected as a target and re-failed in the same way — an infinite loop that pinned quota and starved real work.

I'm wondering why the loop was infinite. I would assume a deleted workload would trigger a Delete event which would make the workload_controller to delete the workload from cache, and so it would stop being a target in consecutive scheduling cycles.

I'm wondering if maybe you hit also another issue where the Delete event was skipped due to replica failover: https://github.com/kubernetes-sigs/kueue/issues/10495,. This issue was fixed here: 

It does not indicate the PR is not useful, but maybe the severity of the issue would be smaller on the version fixing  wdyt?

jzhaojieh:
> I'm wondering if maybe you hit also another issue where the Delete event was skipped due to replica failover: #10495,. This issue was fixed here:

Thanks for sharing that bug fix — it may very well be the cause here since this loop only happened during leader failover during kueue-controller restarts and it was a bit unclear why it never received the delete event to clear the workload.

> It does not indicate the PR is not useful, but maybe the severity of the issue would be smaller on the version fixing wdyt?

Yup agreed, with delete event propagation fixed, this PR would just expedite the unblocking of the scheduling loop since it would eventually unblock once the delete event is reconciled. The other thing this catches is that NotFound on one preemption target currently cancels the parallel evictions for the other targets in the same IssuePreemptions call which would have succeeded otherwise

mimowo:
@jzhaojieh could you rebase the PR as it improves the handling of scheduling in case of deleted workloads. This can be particularly useful when a user may deleted a Deployment - and thus many workload objects almost at the same time.

pkg/scheduler/preemption/preemption.go
```diff
@@ -232,6 +233,25 @@ func (p *Preemptor) IssuePreemptions(
 		)
 		if err != nil {
 			p.preemptionExpectations.ObservedUID(log, targetKey, target.WorkloadInfo.Obj.UID)
+			if apierrors.IsNotFound(err) {
```
mimowo, line 236: We probably also need to handle the error here: https://github.com/kubernetes-sigs/kueue/blob/762bba717f69b4621a973318846133bee2df7990/pkg/scheduler/scheduler.go#L857-L859

Issue #8607, opened 2026-01-15
MultiKueue: invesigate alternative API for dispatcher API which does not suffer the SSA issues


mimowo:
**What would you like to be added**:

To investigate if we can design (or re-design the API) for the Dispatcher API which wouldn't suffer the issues for MultiKueue dispatcher API, that the external controllers need to use SSA with the `kueue-admission` as manager.

**Why is this needed**:

To simplify using MultiKueue Dispatcher API, see the docs: https://kueue.sigs.k8s.io/docs/concepts/multikueue/#external-custom-implementation

tenzen-y:
AFAIK, eliminating the limitation of the `kueue-admission` field manager is one of our motivations for MergePatch instead of SSA in Kueue Admission, right?

Why do we need to investigate alternatives?

mimowo:
> AFAIK, eliminating the limitation of the kueue-admission field manager is one of our motivations for MergePatch instead of SSA in Kueue Admission, right?

Yes, that's right. 

> Why do we need to investigate alternatives?

Because if we manage to redesign the API which eliminates the problem, then we don't have currently any other motivation to use Patch, see: https://github.com/kubernetes-sigs/kueue/issues/7035#issuecomment-3754250002.

tenzen-y:
> > AFAIK, eliminating the limitation of the kueue-admission field manager is one of our motivations for MergePatch instead of SSA in Kueue Admission, right?
> 
> Yes, that's right.
> 
> > Why do we need to investigate alternatives?
> 
> Because if we manage to redesign the API which eliminates the problem, then we don't have currently any other motivation to use Patch, see: [#7035 (comment)](https://github.com/kubernetes-sigs/kueue/issues/7035#issuecomment-3754250002).

Thank you for updating that. As I commented on #7035, I'm with you.

PR #10877, opened 2026-05-01
KEP-9270: Introduce configuration for the MultiKueue Incremental Dispatcher

### LINKED ISSUE #9270: MultiKueue: Support priority-based / ordered cluster selection
Hi Team,

  Support for priority/weight-based cluster selection in MultiKueue, so that workloads
  are preferentially dispatched to higher-priority worker clusters before falling back
  to lower-priority ones.

  ## Why is this needed?

  Currently, MultiKueue uses a race model — workloads are dispatched to **all** worker
  clusters listed in `MultiKueueConfig.spec.clusters` simultaneously, and the first
  cluster to admit wins. There is no mechanism to express cluster preference or ordering.

  ### Use case

  We run 3 worker clusters across regions (e.g., JP, US, SG) for GPU workloads. We want
  to prefer scheduling on our primary cluster (cheaper/closer), and only overflow to
  secondary and tertiary clusters when the primary lacks resources. 

 ## Proposed solution

  Add a `priority` or `weight` field to `MultiKueueConfig.spec.clusters`:

  ```yaml
  kind: MultiKueueConfig
  metadata:
    name: multikueue-default
  spec:
    clusters:
      - name: "jp-osa"
        priority: 1        # highest priority, try first
      - name: "us-002"
        priority: 2        # try if jp-osa is full
      - name: "sg-001"
        priority: 3        # last resort
```

Expected behavior with ordered dispatch

  1. Submit workload to the highest-priority cluster only
  2. If that cluster's ClusterQueue cannot admit (quota full / pending) within a configurable timeout, try the next cluster
  3. Clean up remote workloads on clusters that lost
  4. Fall back to race-all if all clusters are above threshold

  Environment

  - Kueue version: v0.16.1
  - Kubernetes version: 1.33
  - MultiKueue with 3 worker clusters 

If there are already workaround, please suggest.

Thank You,
Sanjay

### PR DESCRIPTION
**Pull request by Mostafahassen1**
#### What type of PR is this?

/kind kep
/area multikueue

This PR proposes a KEP to introduce the `stepSize` configuration parameter for the MultiKueue Incremental Dispatcher. 

Currently, the Incremental Dispatcher can either query worker clusters one-by-one (which is slow) or all-at-once (which creates a heavy API load and risks race conditions). This KEP outlines a design to allow administrators to batch these queries by specifying a `stepSize`, finding the optimal balance between dispatch speed and system stability.

Fixes #9270

This PR contains the initial KEP document (`README.md` and `kep.yaml`) for the feature proposed in #9270. 

Once this KEP is reviewed and approved (status: implementable), I will open follow-up PRs containing the API additions to `MultiKueueConfig` and the core logic updates to the Workload Controller.

```release-note
NONE

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Configurable batch size for the incremental dispatcher via a new stepSize setting (default 3, min 1), letting admins control how many worker clusters are dispatched per batch while preserving all-at-once behavior.
  * Added alpha-stage metadata for the incremental dispatcher configuration.

* **Documentation**
  * Added a comprehensive design proposal covering configuration, batching behavior, tests, graduation criteria, risks/mitigations, implementation notes, and follow-ups.
  * Chart docs updated to v0.18.1 and Helm prerequisites clarified to v4+.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### CODE REVIEW

**Comment by Mostafahassen1** [comment]
CC /  @mimowo , @olekzabl

File: keps/9270-multikueue-incremental-step-size/kep.yaml [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. Kueue typically uses 0.x versioning.
+latest-milestone: "v0.9"
+
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  alpha: "v0.9"
+  beta: "v0.10"
+  stable: "v1.0"
```
olekzabl: This will need updating. I guess alpha 0.19, beta 0.20, stable 0.21?

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -0,0 +1,119 @@
...
+- [Drawbacks](#drawbacks)
+<!-- /toc -->
+
+## Summary
+
+This proposal introduces a new configuration field called `stepSize` to the MultiKueue Incremental Dispatcher. This setting will control the number of worker clusters that the Manager Cluster will attempt to dispatch a workload to concurrently.
+
+## Motivation
+
+Currently, when MultiKueue attempts to schedule a workload across multiple worker clusters, the Incremental Dispatcher has a limitation. If it checks clusters one by one, the process is too slow and delays job execution. If it attempts to dispatch to all clusters simultaneously, it generates unnecessary API traffic, increases the load on the Manager Cluster, and can lead to race conditions where multiple clusters try to admit the same job at the same time.
```
olekzabl: Nit: I feel this paragraph is somewhat confusing.

I think you captured the _tradeoff_ well (between slow dispatching & high resource usage).
However, that tradeoff _does not yet exist_, as we currently hard-code `batchSize := 3`.
I think this paragraph sounds as if this was already variable.

Also, when you write "the Incremental Dispatcher has a limitation", it remains unclear what's the "limitation" specifically.

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+Introducing a `stepSize` allows cluster administrators to configure a "batch size" for this process, finding the perfect balance between speed and system stability.
+
+### Goals
+
+* Add a `stepSize` parameter to the `MultiKueueConfig` API.
+* Update the Workload Controller's dispatching loop to process worker clusters in batches defined by the `stepSize`.
+* Ensure that if a workload is admitted in a batch, the remaining pending proxy workloads in that batch are cleanly removed.
```
olekzabl: 1. Why only in that batch?
2. Isn't this already done, [here](https://github.com/kubernetes-sigs/kueue/blob/3a3d405ff0596aa4d3f433c90a73b570b59b2e94/pkg/controller/admissionchecks/multikueue/workload.go#L460-L469)?

(I'm not sure if it's worth mentioning as a goal).

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Proposal
+
+We propose updating the `MultiKueueConfig` API to accept an optional integer for the step size, and updating the Workload Controller to respect this batch limit when iterating through the list of available worker clusters.
+
+
+### Risks and Mitigations
+
+**Risk:** An administrator might set the `stepSize` to an extremely high number (e.g., equal to the total number of clusters), which effectively recreates the heavy load issues of the `AllAtOnce` dispatcher and increases the chance of race conditions where multiple clusters admit the job at the exact same time.
+**Mitigation:** Provide clear API documentation regarding best practices. The Workload Controller's conflict resolution logic (which deletes duplicate proxy workloads once one is admitted) will naturally handle race conditions, just as it already does for the `AllAtOnce` strategy.
```
olekzabl: Nit: please separate them by an empty line, for better rendering.

olekzabl: > (which deletes duplicate proxy workloads once one is admitted)

Nit: I'd remove this part because this is not the only race condition threatening us.
See e.g. [KEP-8303](https://github.com/kubernetes-sigs/kueue/tree/main/keps/8303-multikueue-orchestrated-preemption) which deals with another kind of race condition.

(Still, the rest of the sentence makes sense: whatever prevention mechanisms we have, they will still act).

olekzabl: > regarding best practices

Out of curiosity, what "best practices" do you have in mind?
Is there any number (constant, or dependent on N = number of clusters) which you wouldn't recommend exceeding?

Asking because I don't know. Maybe N/2 is still reasonable for some use cases? Maybe even more? (Presumably, some users are happy even with `AllAtOnce`).

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Risks and Mitigations
+
+**Risk:** An administrator might set the `stepSize` to an extremely high number (e.g., equal to the total number of clusters), which effectively recreates the heavy load issues of the `AllAtOnce` dispatcher and increases the chance of race conditions where multiple clusters admit the job at the exact same time.
+**Mitigation:** Provide clear API documentation regarding best practices. The Workload Controller's conflict resolution logic (which deletes duplicate proxy workloads once one is admitted) will naturally handle race conditions, just as it already does for the `AllAtOnce` strategy. 
+
+## Design Details
+
+### API Changes
+We will update the `MultiKueueConfigSpec` struct to include the new field. This change will be backwards compatible by defaulting the step size to `1` (which matches the current one-by-one behavior).
```
olekzabl: > defaulting the step size to `1` (which matches the current one-by-one behavior).

Hmm, is the current behavior one-by-one?

https://github.com/kubernetes-sigs/kueue/blob/3a3d405ff0596aa4d3f433c90a73b570b59b2e94/pkg/controller/workloaddispatcher/incrementaldispatcher.go#L180

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -0,0 +1,119 @@
...
+    // Minimum value is 1. If not set, it defaults to 1.
+    // +optional
+    // +kubebuilder:default=1
+    // +kubebuilder:validation:Minimum=1
+    StepSize *int32 `json:"stepSize,omitempty"`
+}
+```
+
+### Execution Loop
+The core logic change will take place inside the MultiKueue Workload Controller's reconciliation loop:
```
olekzabl: Hmm, it seems that we need less changes than that.
IIUC the batching logic already exists in [this file](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/workloaddispatcher/incrementaldispatcher.go); we just need to replace the hard-coded `batchSize := 3` with your proposed custom parameter.

Unless I'm confusing things.

olekzabl (changes requested):
Hi, thank you for your PR!

This is my first round of comments; I plan to come back soon.

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -0,0 +1,119 @@
...
+    * Feature has been in Beta for at least one release cycle with no major bugs reported.
+
+## Implementation History
+
+- 2026-05-01: KEP proposed and initial draft created.
+
+## Drawbacks
+
+This adds slight code complexity to the existing Incremental Dispatcher loop and introduces one additional configuration parameter for administrators to manage and tune.
+```
```
olekzabl:
```suggestion
```

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+- `TestIncrementalDispatcherStepSize`: Simulate a Manager Cluster with 5 Worker Clusters. Set `stepSize` to `2`. Submit a workload and verify via the API server that exactly 2 proxy workloads are created in the first step. Simulate a rejection on the first 2 clusters, and verify that the controller correctly creates the next 2 proxy workloads in the second step.
+
+#### e2e tests
+
+Standard E2E multi-cluster tests will be updated to include an iteration where `MultiKueueConfig` is deployed with a `stepSize > 1` to ensure workloads execute successfully end-to-end without duplication.
+
+### Graduation Criteria
+
+* **Alpha:** * API fields added.
```
olekzabl:
```suggestion
* **Alpha:**
    * API fields added.
```

and same for "Beta" below.

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -0,0 +1,119 @@
...
+ I/we understand the owners of the involved components may require updates to existing tests to make this code solid enough prior to committing the changes necessary to implement this enhancement.
+
+#### Prerequisite testing updates
+
+No major prerequisite testing updates are required. Current mock worker cluster frameworks in EnvTest are sufficient.
+
+#### Unit tests
+
+- `pkg/controller/core/workload_controller_test.go`: `<date>` - Verify that the worker cluster list is correctly divided into chunks matching the `stepSize`.
+- `pkg/controller/core/workload_controller_test.go`: `<date>` - Verify that the default value remains `1` if the user does not specify a `stepSize`.
```
olekzabl:
```suggestion
- `pkg/controller/core/workload_controller_test.go`: Verify that the worker cluster list is correctly divided into chunks matching the `stepSize`.
- `pkg/controller/core/workload_controller_test.go`: Verify that the default value remains `1` if the user does not specify a `stepSize`.
```

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -0,0 +1,119 @@
...
+type MultiKueueConfigSpec struct {
+    // Other existing fields...
+
+    // StepSize defines the number of worker clusters the Incremental 
+    // Dispatcher will query at the same time. 
+    // Minimum value is 1. If not set, it defaults to 1.
+    // +optional
+    // +kubebuilder:default=1
+    // +kubebuilder:validation:Minimum=1
+    StepSize *int32 `json:"stepSize,omitempty"`
```
olekzabl: With this proposal, if a user wants to use incremental dispatcher with step 4, they'll have to say:

```yaml
# in the global Configuration object for Kueue
multiKueue:
  dispatcherName: kueue.x-k8s.io/multikueue-dispatcher-incremental

# in the MultiKueueConfig (referenced by the MK AC, referenced by the manager CQ)
spec:
  stepSize: 4
```

This _might_ be best - but I'd like to discuss:

1. Do we really want step size configured in a very distant place from the choice of the dispatcher?
   (This would be justified if we needed to support different step sizes for different CQs.
   However, note that the choice "Incremental or AllAtOnce or Custom" is already global.
   So maybe we have no reason to stand out?)

2. Regardless of the answer to (1), I think I'd prefer the field name to be more self-explaining.
   E.g. `IncrementalDispatcherStepSize`.

BTW, if we choose to keep both fields next to each other, I'd also add validation that "step size" can be set _only if_ "incremental" is actually chosen. (This can be done with kubebuilder rules).

keps/9270-multikueue-incremental-step-size/README.md
Mostafahassen1: > Out of curiosity, what "best practices" do you have in mind?

Since you are providing them with a new configuration, they need to know how to use it safely. The 'best practices' that you will document will typically include the following:

**Avoiding extreme values:**  Warning them against setting the `stepSize` to a massive number (such as the total number of clusters they have). You will clarify that doing so defeats the purpose of this feature and recreates the exact issues of high API load and race conditions

keps/9270-multikueue-incremental-step-size/README.md
Mostafahassen1: You're right. I knew it was currently 3, but I incorrectly assumed we should just default to the minimum value. Sticking with 3 makes much more sense so we maintain the exact same behavior for users when they upgrade.

keps/9270-multikueue-incremental-step-size/README.md
Mostafahassen1: I agree with these points. Sticking the step size next to the dispatcher globally makes much more sense.

I'll update the KEP to:

1-Move the field to the global Configuration struct.

2-Rename it to `IncrementalDispatcherStepSize`.

3-Add kubebuilder validation so it can only be used with the incremental dispatcher.

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -0,0 +1,127 @@
...
+* We will not alter the `AllAtOnce` dispatcher behavior, as this proposal strictly targets the Incremental Dispatcher.
+
+## Proposal
+
+We propose updating the `MultiKueueConfig` API to accept an optional integer for the step size, and updating the Workload Controller to respect this batch limit when iterating through the list of available worker clusters.
+
+
+### Risks and Mitigations
+
+**Risk:**An administrator might set the `stepSize` to an extremely high number (e.g., equal to the total number of clusters), which effectively recreates the heavy load issues of the `AllAtOnce` dispatcher and increases the chance of race conditions where multiple clusters admit the job at the exact same time.
```
olekzabl: `**Risk**` misrenders, probably due to a missing space after that.

keps/9270-multikueue-incremental-step-size/README.md
olekzabl: Thanks for fixing here - but it's still there for "Beta".

olekzabl:
1. Once the plan gets approved (which I hope to happen soon), we'll probably need to update the "main" KEP for MultiKueue, KEP-693. (Search for "incremental" there, and you'll see).
  OTOH, AFAIU, the good practice is to keep such old KEPs universally valid for all versions - which is achieved by saying "until v..." or "since v..." (for inspiration, search for "since v" in KEP-693, or e.g. in KEP-2724).

2. Given the above - my apologies for coming to this thought so late! - I'm now not sure if this new KEP should even exist.
  So far, this feature feels quite simple, and not involving any major controversy. So maybe just updating KEP-693 could be enough.
  OTOH, Kueue already has some tiny KEPs (e.g. 2936, or 3122). So perhaps having this new one won't hurt :)

@mimowo You know the customs better; what would you say?
Should KEP-693 be updated? Should it happen in this PR? Should we still keep this new KEP?

_(P.S. I think a separate KEP makes much sense **if** we want to have a new dedicated feature gate - but do we? see [below](https://github.com/kubernetes-sigs/kueue/pull/10877#issuecomment-4383899103))_

keps/9270-multikueue-incremental-step-size/kep.yaml
```diff
@@ -0,0 +1,35 @@
...
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  alpha: "v0.19"
+  beta: "v0.20"
+  stable: "v0.21"
+
+# The following PRR answers are required at alpha release
+# List the feature gate name and the components for which it must be enabled
+feature-gates:
+  - name: MultiKueueStepSize
```
olekzabl: I'm not sure if it's strictly required, but how about explicitly mentioning this FG anywhere in the KEP text?

Perhaps the best place would be the `//` comment above your new field - disclaiming that the field is going to be ignored (right?) when the FG is disabled.

Mostafahassen1:
@olekzabl 
>Given the above - my apologies for coming to this thought so late! - I'm now not sure if this new KEP should even exist.
So far, this feature feels quite simple,

No problem at all! I’m completely open to following the Kueue standards here. If you think it’s better to update KEP-693 instead of having a new one

olekzabl:
> @olekzabl
> 
> > Given the above - my apologies for coming to this thought so late! - I'm now not sure if this new KEP should even exist.
> > So far, this feature feels quite simple,
> 
> No problem at all! I’m completely open to following the Kueue standards here. If you think it’s better to update KEP-693 instead of having a new one

I'm sorry, I'm puzzled by this question (i.e. "update old or write new or both").
I'm increasingly feeling that the answer depends on whether we want to have a dedicated feature gate for this - which is also not obvious. (After all, your only API change is introducing a new field, which does not change _much_ in the implementation when set, and should change _nothing_ in it when unset).

For me, right now, this feels a delicate judgement call... so for now I'd keep it as is and wait to see what the other reviewers will say.

keps/9270-multikueue-incremental-step-size/kep.yaml
```diff
@@ -0,0 +1,35 @@
...
+    components:
+      - kueue-controller-manager
+disable-supported: true
+
+# The following PRR answers are required at beta release
+metrics:
+  - kueue_multikueue_dispatched_workloads_total
```
olekzabl: I'm not sure if we need it; AFAICS the last KEP featuring a metric here was KEP-1282 (>30 KEPs ago).

(If, however, you claim we should have it, then I think it should be somehow described in the KEP - which I don't see)

keps/9270-multikueue-incremental-step-size/kep.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - TBD
+approvers:
+  - TBD
+
+# The target maturity stage in the current dev cycle for this KEP.
+stage: alpha
+
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. Kueue typically uses 0.x versioning.
+latest-milestone: "v0.9"
```
olekzabl: This probably needs an update.

keps/9270-multikueue-incremental-step-size/kep.yaml
```diff
@@ -0,0 +1,35 @@
+title: MultiKueue Incremental Dispatcher Step Size
+kep-number: 9270
+authors:
+  - "@Mostafahassen1"
+status: provisional
+creation-date: "2026-05-01"
+reviewers:
+  - TBD
```
olekzabl:
```suggestion
  - olekzabl
```

:)

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Design Details
+
+### API Changes
+
+We will update the global `Configuration` API (specifically the `MultiKueue` struct) to include the new field, rather than placing it in `MultiKueueConfigSpec`. This keeps the step size configuration co-located with the `DispatcherName` selection.
+
+The field will be named `IncrementalDispatcherStepSize` for clarity. To ensure backwards compatibility and preserve existing system behavior, the default value will be `3`.
+
+We will also use Kubebuilder validation markers (e.g., `CEL` rules or `+union` discriminators depending on the struct setup) to enforce that this field can only be populated if `DispatcherName` is set to the incremental dispatcher.
```
olekzabl: Nit: I think the "struct setup" is already known.

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -0,0 +1,128 @@
...
+ I/we understand the owners of the involved components may require updates to existing tests to make this code solid enough prior to committing the changes necessary to implement this enhancement.
+
+#### Prerequisite testing updates
+
+No major prerequisite testing updates are required. Current mock worker cluster frameworks in EnvTest are sufficient.
+
+#### Unit tests
+
+- `pkg/controller/core/workload_controller_test.go`: Verify that the worker cluster list is correctly divided into chunks matching the `stepSize`.
+- `pkg/controller/core/workload_controller_test.go`: Verify that the default value remains `1` if the user does not specify a `stepSize`.
```
olekzabl:
```suggestion
- `pkg/controller/core/workload_controller_test.go`: Verify that the default value remains `3` if the user does not specify a `stepSize`.
```

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+Standard E2E multi-cluster tests will be updated to include an iteration where `MultiKueueConfig` is deployed with a `stepSize > 1` to ensure workloads execute successfully end-to-end without duplication.
+
+### Graduation Criteria
+
+* **Alpha:**
+    * API fields added.
+    * Workload Controller updated to support batching.
+    * Unit tests and EnvTest integration tests passing.
+* **Beta:** * Gather feedback from the community.
```
olekzabl:
```suggestion
* **Beta:**
    * Gather feedback from the community.
```

olekzabl (commented):
LGTM; I just have one more round of tiny comments.

Now I'm feeling we're just blocked by [this thread](https://github.com/kubernetes-sigs/kueue/pull/10877#issuecomment-4383767843).

tenzen-y:
@olekzabl @mimowo I am just wondering if we should add new capabilities to the dispatcher mechanism since we were planning to replace it with another approach in https://github.com/kubernetes-sigs/kueue/issues/8607

But if you want to postpone https://github.com/kubernetes-sigs/kueue/issues/8607 to a later date for now, I'm ok with that.

mimowo:
@tenzen-y my intention in https://github.com/kubernetes-sigs/kueue/issues/8607 is not to replace or decommission the built-in dispatches that we have (AllAtOnce, Incremental). The intention of that Issue is to make it easier to work with external dispatchers - currently they need to pretend they are core Kueue when sending requests.

So regardless if we do https://github.com/kubernetes-sigs/kueue/issues/8607 it is useful to add new capabilities to the built-in dispatchers.

mimowo:
As discussed with @olekzabl today - I would prefer to generalize the KEP as I think there is a need to extend the configuration in the future - to add other fields like "timeout", or maybe "stepSizeAsPercent". So, I would suggest to name the KEP as Introduce configuration for the MultiKueue Incremental Dispatcher. For now the configuration will be just one field, but it will also give a paved way for future configuration of built-in dispatchers, I'm thinking about

```yaml
multiKueue:
  dispatcherName: "kueue.x-k8s.io/multikueue-dispatcher-incremental"
  inrementalDispatcherConfig:
    stepSize: 1
```
cc @olekzabl @tenzen-y @Mostafahassen1

tenzen-y:
> @tenzen-y my intention in #8607 is not to replace or decommission the built-in dispatches that we have (AllAtOnce, Incremental). The intention of that Issue is to make it easier to work with external dispatchers - currently they need to pretend they are core Kueue when sending requests.
> 
> So regardless if we do #8607 it is useful to add new capabilities to the built-in dispatchers.

That sounds good to me, thank you. In that case, let's see this KEP.

olekzabl:
> As discussed with @olekzabl today - I would prefer to generalize the KEP as I think there is a need to extend the configuration in the future - to add other fields like "timeout", or maybe "stepSizeAsPercent".

+1 (and thank you for posting this info before I managed to do it).

Additional points:

1. We think your new feature gate can stay, particularly given the Beta graduation criterion on gathering user feedback.

2. This KEP can also stay as it is.
   Personally, I'd be happy if you add a short "Possible Follow-ups" section, covering @mimowo 's ideas listed above (timeout, stepSizeAsPercent). This way we'll ensure good ideas aren't lost. This may also facilitate receiving feedback.

3. In KEP-693, currently mentioning "set of 3 clusters", let's make sure this doesn't become incorrect.
   This can be a minimal change, e.g. at the bottom of the description of the incremental dispatcher, add 1 sentence:
   "There are plans to make this behavior configurable; see [a link to your new KEP]".

   * FYI, one clarification I got from @mimowo : our custom is to allow KEPs get out of date once their feature becomes stable. However, this is not the case for KEP-693, where MultiKueue is still in Beta. So, AFAIU, it's still "fresh enough" to be kept accurate.

keps/9270-multikueue-incremental-step-size/README.md
olekzabl: This still misrenders; perhaps you have some auto-formatter doing it?
(If so, consider adding blank lines to prevent this).

keps/9270-multikueue-incremental-step-size/kep.yaml
```diff
@@ -0,0 +1,32 @@
+title: MultiKueue Incremental Dispatcher Step Size
```
mimowo: Align the title with KEP title

keps/9270-multikueue-incremental-step-size/kep.yaml
```diff
@@ -0,0 +1,32 @@
+title: MultiKueue Incremental Dispatcher Step Size
+kep-number: 9270
+authors:
+  - "@Mostafahassen1"
+status: provisional
+creation-date: "2026-05-01"
+reviewers:
+  - olekzabl
+approvers:
+  - TBD
```
mimowo: add me here

keps/9270-multikueue-incremental-step-size/kep.yaml
```diff
@@ -0,0 +1,32 @@
...
+  - olekzabl
+approvers:
+  - TBD
+
+# The target maturity stage in the current dev cycle for this KEP.
+stage: alpha
+
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. Kueue typically uses 0.x versioning.
+latest-milestone: "v0.17"
```
mimowo: 0.18 or 0.19

keps/9270-multikueue-incremental-step-size/kep.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  alpha: "v0.19"
```
mimowo: align with latest-version

keps/9270-multikueue-incremental-step-size/kep.yaml
```diff
@@ -0,0 +1,32 @@
...
+  beta: "v0.20"
+  stable: "v0.21"
```
mimowo: replace with TBD for now

keps/9270-multikueue-incremental-step-size/kep.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+# The following PRR answers are required at alpha release
+# List the feature gate name and the components for which it must be enabled
+feature-gates:
+  - name: MultiKueueStepSize
```
mimowo: Update the FG name to MultiKueueIncrementalDispatcherConfig

mimowo:
/retitle KEP-9270: Introduce configuration for the MultiKueue Incremental Dispatcher

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -54,27 +54,33 @@ We propose updating the `MultiKueueConfig` API to accept an optional integer for
...
-    IncrementalDispatcherStepSize *int32 `json:"incrementalDispatcherStepSize,omitempty"`
+// Other existing fields...
+// DispatcherName defines the type of dispatcher to use.
+DispatcherName *string `json:"dispatcherName,omitempty"`
+
+// IncrementalDispatcherConfig contains the configuration for the incremental dispatcher.
+// This field is only valid when DispatcherName is set to the incremental dispatcher.
+// Note: This field is going to be ignored when the MultiKueueStepSize feature gate is disabled.
+// +optional
+IncrementalDispatcherConfig *IncrementalDispatcherConfig `json:"incrementalDispatcherConfig,omitempty"`
```
olekzabl: Nit: restore the indent

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -54,27 +54,33 @@ We propose updating the `MultiKueueConfig` API to accept an optional integer for
...
-    // +kubebuilder:default=3
-    // +kubebuilder:validation:Minimum=1
-    IncrementalDispatcherStepSize *int32 `json:"incrementalDispatcherStepSize,omitempty"`
+// Other existing fields...
+// DispatcherName defines the type of dispatcher to use.
+DispatcherName *string `json:"dispatcherName,omitempty"`
+
+// IncrementalDispatcherConfig contains the configuration for the incremental dispatcher.
+// This field is only valid when DispatcherName is set to the incremental dispatcher.
+// Note: This field is going to be ignored when the MultiKueueStepSize feature gate is disabled.
```
olekzabl: Don't forget to update the FG name also here

olekzabl (approved):
LGTM!

keps/9270-multikueue-incremental-step-size/kep.yaml
```diff
@@ -8,7 +8,8 @@ reviewers:
   - olekzabl
   - mimowo
 approvers:
-  - TBD
+  - mimowo
+  - olekzabl
```
olekzabl: Thank you :) but IIUC I'm not an Approver in this repo, so I should be listed only among `reviewers`.

keps/9270-multikueue-incremental-step-size/README.md
olekzabl: Nit: this line still mis-renders.
(Perhaps you have some auto-formatter doing it? If so, consider adding blank lines to prevent it).

olekzabl (approved):

keps/9270-multikueue-incremental-step-size/kep.yaml
```diff
@@ -0,0 +1,32 @@
...
+# The target maturity stage in the current dev cycle for this KEP.
+stage: alpha
+
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. Kueue typically uses 0.x versioning.
+latest-milestone: "v0.19"
+
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  alpha: "v0.19"
```
mimowo: So I skimmed the PR and [impl](https://github.com/kubernetes-sigs/kueue/pull/11208.) and both LGTM overall 

If we have time we may consider including this for 0.18. TBD as this is not top priority, but seems like a nice and simple feature

cc @tenzen-y

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -0,0 +1,145 @@
...
+      IncrementalDispatcherConfig *IncrementalDispatcherConfig `json:"incrementalDispatcherConfig,omitempty"`
+}
+
+type IncrementalDispatcherConfig struct {
+      // StepSize defines the number of worker clusters the Incremental 
+      // Dispatcher will query simultaneously. 
+      // Minimum value is 1. If not set, it defaults to 3.
+      // +optional
+      // +kubebuilder:default=3
+      // +kubebuilder:validation:Minimum=1
```
mimowo: nit, but the kubebuilder markers don't matter for ConfigAPI , because the kubebuilder markers are only used for schema generation, but we don't generate the schema for the configmap api.,

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -0,0 +1,145 @@
...
+```go
+type MultiKueue struct {
+       // Other existing fields...
+
+      // DispatcherName defines the type of dispatcher to use.
+      DispatcherName *string `json:"dispatcherName,omitempty"`
+
+      // IncrementalDispatcherConfig contains the configuration for the incremental dispatcher.
+      // This field is only valid when DispatcherName is set to the incremental dispatcher.
+      // Note: This field is going to be ignored when the MultiKueueIncrementalDispatcher feature gate is disabled.
```
mimowo:
```suggestion
      // Note: This field is ignored when the MultiKueueIncrementalDispatcher feature gate is disabled.
```
nit, assume the reader is a future user or developer, so "going to" will not be relevant soon.

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -0,0 +1,145 @@
...
+- `TestIncrementalDispatcherStepSize`: Simulate a Manager Cluster with 5 Worker Clusters. Set `stepSize` to `2`. Submit a workload and verify via the API server that exactly 2 proxy workloads are created in the first step. Simulate a rejection on the first 2 clusters, and verify that the controller correctly creates the next 2 proxy workloads in the second step.
+
+#### e2e tests
+
+Standard E2E multi-cluster tests will be updated to include an iteration where `MultiKueueConfig` is deployed with a `stepSize > 1` to ensure workloads execute successfully end-to-end without duplication.
+
+### Graduation Criteria
+
+* **Alpha:**
+    * API fields added.
```
mimowo: nit, add that "MultiKueueIncrementalDispatcherConfig" is alpha (disabled by default)

for beta: "MultiKueueIncrementalDispatcherConfig" is beta (enabled by default)

for Stabke: "MultiKueueIncrementalDispatcherConfig" is locked (will be removed in N+2)

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -0,0 +1,145 @@
...
+```
+
+### Execution Loop
+The core logic change is minimal because the batching mechanism is already implemented in the Incremental Dispatcher (`pkg/controller/workloaddispatcher/incrementaldispatcher.go`).
+
+The update will simply modify the existing logic to:
+1. Read the `stepSize` value from the `MultiKueueConfig`.
+2. Replace the hardcoded `batchSize := 3` variable with this dynamically configured `stepSize` parameter.
+
+The existing loop and conflict resolution mechanisms will naturally handle the rest of the execution based on this new size, meaning no major architectural changes are required to the dispatching loop itself.
```
mimowo:
```suggestion
### Implementation overview

The core logic change is minimal because the batching mechanism is already implemented in the Incremental Dispatcher, assuming the hard-coded value of stepSize=3.  
No architectural changes are required to the dispatching loop itself.
```
nit, but I think this section is too verbose, proposing to make it more concise.

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Summary
+
+This proposal introduces a new configuration structure for the MultiKueue Incremental Dispatcher, starting with a stepSize field. This setting will control the number of worker clusters that the Manager Cluster will attempt to dispatch a workload to concurrently.
+
+## Motivation
+
+Currently, the MultiKueue Incremental Dispatcher has a hardcoded batch size `batchSize := 3` when attempting to schedule a workload across multiple worker clusters. The primary limitation is that this value is not configurable. For administrators managing a massive fleet of worker clusters, a fixed batch size of 3 is too slow and delays job execution. Conversely, if administrators were to theoretically increase this to query all clusters simultaneously to speed things up, it would generate heavy API traffic and increase the risk of race conditions. By making this step size configurable, administrators can find the optimal balance between dispatch speed and Manager Cluster resource usage for their specific environment.
+Introducing a `stepSize` allows cluster administrators to configure a "batch size" for this process, finding the perfect balance between speed and system stability.
```
mimowo: ### 
```suggestion
Currently, the MultiKueue Incremental Dispatcher has a hardcoded batch size `batchSize := 3` when attempting to schedule a workload across multiple worker clusters. 

Sometimes administrators need to configure the parameter in order to:
1. avoid unnecessary job execution delays when there are many clusters (by increasing the stepSize)
2. use more granular prioritization of the clusters (by setting stepSize=1)
3. tune the api-server traffic and load on the manager cluster
```
nit, in order to:
1. make it less verbose and to-the-point
2. avoid the overselling statements like " finding the perfect balance"
3. add the use case for more granular prioritization of the clusters
4. rephrase the api-server traffiic point slightly

mimowo:
LGTM overall, left mostly editorial comments

mimowo:
/lgtm
/approve
In case @tenzen-y would like to also review
/hold

charts/kueue/README.md
```diff
@@ -1,6 +1,6 @@
 # kueue
 
-![Version: 0.17.2](https://img.shields.io/badge/Version-0.17.2-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.17.2](https://img.shields.io/badge/AppVersion-v0.17.2-informational?style=flat-square)
+![Version: 0.18.1](https://img.shields.io/badge/Version-0.18.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.18.1](https://img.shields.io/badge/AppVersion-v0.18.1-informational?style=flat-square)
```
mimowo, line 3: Can you rebase the PR, this looks like applying changes to the old version before rebase

Mostafahassen1: Hi  @mimowo, this file is now not part of my PR

tenzen-y:
/release-note-edit

```release-note
NONE
```

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+The field will be named `IncrementalDispatcherConfig` and will contain a `StepSize` field. To ensure backwards compatibility and preserve existing system behavior, the default value for `StepSize` will be `3`.
+
+```go
+type MultiKueue struct {
+       // Other existing fields...
+
+      // DispatcherName defines the type of dispatcher to use.
+      DispatcherName *string `json:"dispatcherName,omitempty"`
+
```
tenzen-y:
```suggestion
```

Can you exclude already existing fields from API changes?

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -0,0 +1,143 @@
...
+      // IncrementalDispatcherConfig contains the configuration for the incremental dispatcher.
+      // This field is only valid when DispatcherName is set to the incremental dispatcher.
+      // Note: This field is ignored when the MultiKueueIncrementalDispatcher feature gate is disabled.
+      // +optional
```
tenzen-y: This is just a side note. This API is a configuration. So, these markers don't work for anything. Then we have to implement validations and defaulting manually.

keps/9270-multikueue-incremental-step-size/README.md
```diff
@@ -0,0 +1,143 @@
...
+      IncrementalDispatcherConfig *IncrementalDispatcherConfig `json:"incrementalDispatcherConfig,omitempty"`
+}
+
+type IncrementalDispatcherConfig struct {
+      // StepSize defines the number of worker clusters the Incremental 
+      // Dispatcher will query simultaneously. 
+      // Minimum value is 1. If not set, it defaults to 3.
```
tenzen-y: Also, this is just a side note. This defaulting should be implemented by manual defaulter implementations: https://github.com/kubernetes-sigs/kueue/blob/5446b2a38eeec3615d95b3e06ecd0ec311ce7d38/apis/config/v1beta2/defaults.go#L68

keps/9270-multikueue-incremental-step-size/kep.yaml
```diff
@@ -0,0 +1,32 @@
...
+kep-number: 9270
+authors:
+  - "@Mostafahassen1"
+status: provisional
+creation-date: "2026-05-01"
+reviewers:
+  - olekzabl
+  - mimowo
+approvers:
+  - mimowo
```
tenzen-y:
```suggestion
reviewers:
  - @olekzabl
  - @mimowo
approvers:
  - @mimowo
  - @tenzen-y
```

keps/9270-multikueue-incremental-step-size/kep.yaml
tenzen-y: We might be able to start from Beta because we add a dedicated field which is optional and has the same default value as the current mechanism.

Additionally, once we add any kind of field to API, we can not drop those features unless API version changes. Hence, starting Alpha may have fewer advantages.

@mimowo Any thoughts?

tenzen-y (commented):
Thank you for keeping this latest!
I left some comments, thank you!

keps/9270-multikueue-incremental-step-size/kep.yaml
mimowo: +1 I think sometimes it makes sense to start with Alpha even if there is API, because the feature is making changes to the core scheduling logic like TAS or Concurrent admission. However in case of this feature the logic changes are very narrow so I'm not worried about breaking core code. Let's start with Beta here.

mimowo:
@Mostafahassen1 Do you have cycles to address the remaining comments?

 I somehow assumed the KEP is already merged and merged the implementation in https://github.com/kubernetes-sigs/kueue/issues/9270. Seems like the remaining KEP comments are mostly editorial, but it would be great to address them.

Mostafahassen1:
Hi @mimowo, I’ve been working on it since this morning. I’ll finish it in 15 minutes at most.

Mostafahassen1:
@mimowo  I  finished now from KEP

mimowo:
/lgtm
/unhold
Thank you, I checked https://github.com/kubernetes-sigs/kueue/pull/10877/changes/96ce49af42a8696920ce7ec00bdedcfdf59b9685 to verify if the remaining comments are addressed

PR #10878, opened 2026-05-01
Reduce LFC fragmentation for single pod workloads

jzhaojieh:
#### What type of PR is this?

/kind feature
/area tas

When the LeastFreeCapacity places a single pod workload that fits inside one host, the existing sort keys operate only at the level being sorted without visibility into partially full nodes, so TAS may spread single pod workloads across domains based on least free capacity rather than stacking them on the same nodes, fragmenting capacity 

Adds one new sort key minNonZeroLeafState which represents the smallest free node capacity among leaves in this subtree, such that TAS will pick the subtree containing the tightest non empty node

No 

```release-note
Reduce LFC fragmentation for single pod workloads
```

### DIFF
--- a/test/integration/singlecluster/tas/tas_test.go
+++ b/test/integration/singlecluster/tas/tas_test.go
@@ -5964,6 +5964,111 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
 			})
 		})
 	})
+
+	ginkgo.When("LeastFreeCapacity profile prefers the globally tightest non-empty leaf", framework.SlowSpec, func() {
+		var (
+			topology     *kueue.Topology
+			tasFlavor    *kueue.ResourceFlavor
+			clusterQueue *kueue.ClusterQueue
+			localQueue   *kueue.LocalQueue
+			nodes        []corev1.Node
+		)
+
+		// Topology (block → rack → host), with asymmetric per-rack/per-host
+		// capacity:
+		//
+		//   block-a / rack-a1 / host-a1   capacity 1   ← globally tightest leaf
+		//   block-a / rack-a2 / host-a2   capacity 8
+		//   block-b / rack-b1 / host-b1   capacity 4
+		//   block-b / rack-b2 / host-b2   capacity 4
+		//
+		// Aggregate sliceState (counted as host count × per-host capacity):
+		//   block-a = 9, block-b = 8.
+		//
+		// Without the new minNonZeroLeafState sort key, LFC at the block
+		// level prefers the smaller aggregate (block-b, 8 < 9) and the
+		// workload lands on a host inside block-b. With the new key,
+		// block-a wins (minNZL = 1 < 4) and the workload lands on host-a1
+		// — the globally tightest non-empty leaf.
+		ginkgo.BeforeEach(func() {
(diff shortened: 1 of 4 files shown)


test/integration/singlecluster/tas/tas_test.go
```diff
@@ -5964,6 +5964,111 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
 			})
 		})
 	})
+
+	ginkgo.When("LeastFreeCapacity profile prefers the globally tightest non-empty leaf", framework.SlowSpec, func() {
```
mimowo, line 5968: Interesting, I was assuming that the current LeastFreeCapacity is already looking for the globally thightest node, regardless of the hierarchy, and that this is the subject for the TASHierarchicalUnconstrained PR here: https://github.com/kubernetes-sigs/kueue/pull/10406

Or is this behavior because you are using the slices here with PodSetSliceSize? However, in that case I'm not sure why to use slices.

mimowo (commented):
cc @abhi179 as the author of the related PR https://github.com/kubernetes-sigs/kueue/pull/10406/ so that we make sure the PRs a compatible.

cc @PBundyra

test/integration/singlecluster/tas/tas_test.go
jzhaojieh: > Interesting, I was assuming that the current LeastFreeCapacity is already looking for the globally thightest node

Good callout, using `unconstrained: true` today picks the globally tightest node since it only looks at the lowest topology level, and the issue from this PR is indeed due to using podset slice.                                                                                                                                                     
                                                                                                                                                                         
However PR #10406, which changes the unconstrained behavior to start at the highest level, exposes the issue described here that's currently only present when using podset slice for unconstrained as well. If the globally tightest node gets pruned during the top-down traversal, unconstrained may inadvertently miss the tightest placement.

mimowo: > (...), and the issue from this PR is indeed due to using podset slice.

Can you elaborate why using podset slices is changing the results.

Also, why do you actually use podset slices for single Pod workloads? 

Is it just to illustrate the example, but your actual use case is different? If so then I would like to learn more about your actual use case so that we can be sure the PR is needed, and it is satisfying the use case. 

Overall the idea seems ok, but it is a bit hard for me to assess the various proposals to the unconstrained annotation without a broarder context. 

Also asked a question in similar spirit to the "sibling" PR: https://github.com/kubernetes-sigs/kueue/pull/10406#issuecomment-4380420346

jzhaojieh: > Is it just to illustrate the example, but your actual use case is different? 

So the usecase is the same cluster is used for both training and inference workloads. Training workloads are mostly full-node and specify topology constraints at the NVSO domain layer. Inference workloads may be stackable or full-node but don't require topology constraints. To reduce fragmentation there are two goals:

1. minimize NVSO domain fragmentation to keep as many free domains for training workloads 
2. minimize node level fragmentation to avoid partially filled nodes 

> Also, why do you actually use podset slices for single Pod workloads?

There is not a single TAS constraint we can apply for the workloads without topology constraints to achieve objectives. Using podset slice will help with 1. but doesnt account for 2. While unconstrained topology would help with 2 but not 1. With this PR, using podset slice annotation will achieve both 1 and 2

PR #10890, opened 2026-05-01
remove flavorless admissions handling

omerap12:
#### What type of PR is this?
/kind cleanup

Eliminates the test-only branch in `admissionChecksForAdmission` that treated flavorless admissions as satisfying all admission checks. That branch was introduced as a workaround so a handful of tests that created admissions without any flavor assignments would keep passing, even though a flavorless admission is never a legal state in real use cases.

Fixes #10316 

```release-note
NONE
```

### DIFF
--- a/pkg/workload/workload.go
+++ b/pkg/workload/workload.go
@@ -1530,23 +1530,11 @@ func AdmissionChecksForWorkload(log logr.Logger, wl *kueue.Workload, cq *kueue.C
 	return checksForAllFlavors
 }
 
-func admissionChecksForAdmission(log logr.Logger, acs AdmissionChecks, admission kueue.Admission) sets.Set[kueue.AdmissionCheckReference] {
+func admissionChecksForAdmission(_ logr.Logger, acs AdmissionChecks, admission kueue.Admission) sets.Set[kueue.AdmissionCheckReference] {
 	admissionFlavors := findAdmissionFlavors(admission)
-	if len(admissionFlavors) > 0 {
-		return filterChecks(acs, func(acFlavors flavorSet) bool {
-			return admissionFlavors.Intersection(acFlavors).Len() > 0
-		})
-	}
-
-	// Some tests allow for admissions not to be assigned any flavors.
-	// To ensure those tests work as before: flavorless Admissions are assigned/expected to fulfill all Admission Checks.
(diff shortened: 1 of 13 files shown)


pkg/workload/workload.go
```diff
@@ -1575,23 +1530,11 @@ func AdmissionChecksForWorkload(log logr.Logger, wl *kueue.Workload, cq *kueue.C
 	return checksForAllFlavors
 }
 
-func admissionChecksForAdmission(log logr.Logger, acs AdmissionChecks, admission kueue.Admission) sets.Set[kueue.AdmissionCheckReference] {
+func admissionChecksForAdmission(_ logr.Logger, acs AdmissionChecks, admission kueue.Admission) sets.Set[kueue.AdmissionCheckReference] {
```
omerap12, line 1533: Should I remove the `logr.Logger` entirely or putting `_` is good enough?

test/integration/multikueue/dispatcher_test.go
```diff
@@ -40,6 +40,23 @@ import (
...
+}
+
+// multikueueTestPodSet builds a PodSetAssignment for the multikueueTestFlavor.
+// Used in multikueue integration tests where admissions need a flavor so that
+// the MultiKueue admission check applies to the workload.
+func multikueueTestPodSet(name kueue.PodSetReference) *utiltestingapi.PodSetAssignmentWrapper {
+	return utiltestingapi.MakePodSetAssignment(name).
+		Flavor(corev1.ResourceCPU, multikueueTestFlavor)
+}
+
```
omerap12: Also, I am not sure if this is the right place for these.

omerap12:
@mimowo , maybe you could take a look?

test/integration/multikueue/dispatcher_test.go
mimowo: I would just inline them, so that we can easily see what is the test setup without jumping between files

omerap12: Got it. makes sense. (although the code repetition for this will be high).

mimowo: im ok with that, the code is declarative in nature which makes it easy to follow

omerap12: done.

mimowo (commented):
/lgtm
/approve
Thank you 👍

PR #10891, opened 2026-05-01
fix: Avoid nil pointer panic for fairshare ordering

gyliu513:
#### What type of PR is this?
/kind bug

### Root cause (English)

The `Iter()` implementation has a special-case branch for a preemptor `ClusterQueue` **without a parent Cohort** (`!t.preemptorCq.HasParent()`).

In that scenario, `t.preemptorCq.Parent()` is **nil**. If the function does not **return after finishing** the no-parent branch, control will fall through to the general path and execute:

- `root := t.preemptorCq.Parent().Root()`

That dereferences a nil pointer and will **panic**.

Fixes #

```release-note
FairSharing: fix the bug which can cause nil-pointer panic when FairSharing is enabled on clusters
with ClusterQueues without Cohorts.
```

### DIFF
--- a/pkg/scheduler/preemption/fairsharing/ordering.go
+++ b/pkg/scheduler/preemption/fairsharing/ordering.go
@@ -100,6 +100,7 @@ func (t *TargetClusterQueueOrdering) Iter() iter.Seq[*TargetClusterQueue] {
 					return
 				}
 			}
+			return
 		}
 
 		root := t.preemptorCq.Parent().Root()
--- a/pkg/scheduler/preemption/fairsharing/ordering_test.go
+++ b/pkg/scheduler/preemption/fairsharing/ordering_test.go
@@ -50,7 +50,7 @@ func TestMakeClusterQueueOrdering(t *testing.T) {
 		actions   []string
 		wantOrder []kueue.ClusterQueueReference
 	}{
-		"no cohort: preemptor CQ yielded for in-CQ preemption": {
+		"no cohort: preemptor CQ yielded for in-CQ preemption; repro for nil pointer panic issue": {
 			clusterQueues: []*kueue.ClusterQueue{
 				utiltestingapi.MakeClusterQueue("preemptor").
 					ResourceGroup(*utiltestingapi.MakeFlavorQuotas("default").
@@ -249,8 +249,7 @@ func TestMakeClusterQueueOrdering(t *testing.T) {
 				}
 				actionIdx++
 				if len(gotOrder) > 50 {
-					t.Fatal("exceeded 50 iterations, likely an infinite loop")
-					break
+					t.Error("exceeded 50 iterations, likely an infinite loop")
 				}
 			}
 
--- a/pkg/scheduler/preemption/fairsharing/ordering_test.go
+++ b/pkg/scheduler/preemption/fairsharing/ordering_test.go
@@ -1,0 +1,262 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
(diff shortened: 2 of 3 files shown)


kannon92:
I feel like we need more details on this issue.

Did you discover this issue in production?

gyliu513:
@kannon92 just by reviewing code here, I think this is a coding error, and there is no unit test for this file as well.

pkg/scheduler/preemption/fairsharing/ordering_test.go
```diff
@@ -0,0 +1,262 @@
...
+				gotOrder = append(gotOrder, target.GetTargetCq().GetName())
+				if actionIdx < len(tc.actions) && tc.actions[actionIdx] == "drop" {
+					ordering.DropQueue(target)
+				} else {
+					target.PopWorkload()
+				}
+				actionIdx++
+				if len(gotOrder) > 50 {
+					t.Fatal("exceeded 50 iterations, likely an infinite loop")
+					break
```
mbobrovskyi:
```suggestion
```

pkg/scheduler/preemption/fairsharing/ordering_test.go
```diff
@@ -0,0 +1,262 @@
...
+			for target := range ordering.Iter() {
+				gotOrder = append(gotOrder, target.GetTargetCq().GetName())
+				if actionIdx < len(tc.actions) && tc.actions[actionIdx] == "drop" {
+					ordering.DropQueue(target)
+				} else {
+					target.PopWorkload()
+				}
+				actionIdx++
+				if len(gotOrder) > 50 {
+					t.Fatal("exceeded 50 iterations, likely an infinite loop")
```
mbobrovskyi: Or actually, maybe it’s much better to use t.Error?

mbobrovskyi (commented):
Overall LGTM.

pkg/scheduler/preemption/fairsharing/ordering_test.go
```diff
@@ -0,0 +1,262 @@
...
+		// The same workloads are used as preemption candidates.
+		admitted    []kueue.Workload
+		preemptorCQ kueue.ClusterQueueReference
+		// candidateCQs restricts which admitted workloads become candidates (by CQ name).
+		candidateCQs []kueue.ClusterQueueReference
+		// actions controls per-iteration behavior: "drop" calls DropQueue, anything else calls PopWorkload.
+		actions   []string
+		wantOrder []kueue.ClusterQueueReference
+	}{
+		"no cohort: preemptor CQ yielded for in-CQ preemption": {
```
mimowo: IIUC only this case is reproducing the problem, because the other test cases have cohort, and the panic branch is only for CQs without a parent. So, let's emphasize that by adding `; repro for nil pointer panic issue`.


```suggestion
		"no cohort: preemptor CQ yielded for in-CQ preemption; repro for nil pointer panic issue": {
```

mimowo:
IIUC this the problem: 
/release-note-edit
```release-note
FairSharing: fix the bug which can cause nil-pointer panic when FairSharing is enabled on clusters
with ClusterQueues without Cohorts.
```

pkg/scheduler/preemption/fairsharing/ordering.go
```diff
@@ -100,6 +100,7 @@ func (t *TargetClusterQueueOrdering) Iter() iter.Seq[*TargetClusterQueue] {
 					return
 				}
 			}
+			return
```
mimowo, line 103: Could you also add a unit test in scheduler_fs_test.go - it is not clear to me if the issue exists in real scheduling, or just in unit test which calls this iterator  - that information is also needed for the release notes, I proposed some based on my understanding, but such a test could confirm this is correct.

gyliu513: Done

pkg/scheduler/preemption/fairsharing/ordering_test.go
gyliu513: Done

pkg/scheduler/preemption/fairsharing/ordering_test.go
gyliu513: Done

pkg/scheduler/preemption/fairsharing/ordering_test.go
gyliu513: Done

mimowo (commented):
Thank you 👍 
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16

Issue #10902, opened 2026-05-04
Add preferredDuringSchedulingIgnoredDuringExecution node affinity support in TAS


j-skiba:
**What would you like to be added**:
I would like `preferredDuringSchedulingIgnoredDuringExecution` node affinity to not be ignored in TAS

**Why is this needed**:
Improve scheduling configurability

Domains sorting: https://github.com/kubernetes-sigs/kueue/blob/3af386cd515a18d8dfffad4b227d247a6ebf43f8/pkg/cache/scheduler/tas_flavor_snapshot.go#L1548 and with leader: https://github.com/kubernetes-sigs/kueue/blob/3af386cd515a18d8dfffad4b227d247a6ebf43f8/pkg/cache/scheduler/tas_flavor_snapshot.go#L1511

**Concerns**:
Simply adding preferred affinity check at the beginning of `sortedDomains` method changes default behavior of TAS, on the other hand, if you don't define preferred affinity (which wasn't handled before) you will get the same behavior before handling of preferred affinity was introduced.

So currently we have the following order for sorting domains:
- slice-level capacity check
- pod-level capacity check
- lexicographical check

where slice-level and pod-level are configurable with `TASProfileMixed` feature gate.

The easiest solution would be to add affinity check above all of them but the questions is if we need order of these checks to be configurable, and if yes then is it just 1 -> 2, or 2 -> 1 option or something more sophisticated like reordering of these. I'm not sure if there might be added any more checks in future.

-------

feature gate for choosing sorting criteria https://github.com/kubernetes-sigs/kueue/blob/3af386cd515a18d8dfffad4b227d247a6ebf43f8/pkg/cache/scheduler/tas_flavor_snapshot.go#L1330 when a workload has unconstrained topology.

**Completion requirements**:

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

j-skiba:
kube-scheduler does it using scheduling plugins - https://kubernetes.io/docs/reference/scheduling/config/#scheduling-plugins

Each plugin calculates score of a node, normalizes the result ([0, 100]), multiplies by weight and sums up all scores

j-skiba:
[...]

PR #10903, opened 2026-05-04
Add prefferedDuringScheduling node affinity support for TAS

j-skiba:
#### What type of PR is this?
/kind feature
/area tas

This PR adds support for `preferredDuringSchedulingIgnoredDuringExecution` node affinity in TAS.
The preferred node affinity takes precedence over capacity fitness.
The feature is hidden behind `TASPreferredSchedulingAffinity` feature gate.

Part of #10902

```release-note
TAS: Added support for `preferredDuringSchedulingIgnoredDuringExecution` node affinity. This feature is currently in alpha and is guarded by the `TASPreferredSchedulingAffinity` feature gate.
```


pkg/features/kube_features.go
```diff
@@ -382,6 +382,12 @@ const (
 
 	// Enable recording of WorkloadCreationLatency metric.
 	MetricForWorkloadCreationLatency featuregate.Feature = "MetricForWorkloadCreationLatency"
+
+	// owner: @j-skiba
+	// issue: https://github.com/kubernetes-sigs/kueue/issues/10902
+	// Enable evaluation of preferredDuringSchedulingIgnoredDuringExecution node affinities during scheduling in TAS.
+	// The preferred affinity will take precedence over capacity fitness of a domain.
+	TASPreferredSchedulingAffinity featuregate.Feature = "TASPreferredSchedulingAffinity"
```
mimowo: As I synced with @j-skiba and @mwielgus this solution might be in the future generalized by the score mixing coming from the domain affinity and placement policy.

We are currently designing the more generic solution, and this FG is the easy approach to collect early feedback from users.

I would like to have this FG reflected in the KEP and the plan about its graduation or decomissioning

j-skiba: updated the KEP

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1574,7 +1588,31 @@ func (s *TASFlavorSnapshot) fillInCounts(requirements *topologyAssignmentPodRequ
...
+	var compiledTerms []compiledPreferredTerm
+	if features.Enabled(features.TASPreferredSchedulingAffinity) {
+		compiledTerms = make([]compiledPreferredTerm, 0, len(requirements.preferredNodeAffinity))
+		for _, term := range requirements.preferredNodeAffinity {
+			if term.Preference.MatchExpressions == nil && term.Preference.MatchFields == nil {
+				continue
+			}
+			if nodeSelector, err := nodeaffinity.NewNodeSelector(&corev1.NodeSelector{
+				NodeSelectorTerms: []corev1.NodeSelectorTerm{term.Preference},
+			}); err == nil {
```
mimowo: Do not silence the compilation errors, preferably return them up, or at least log

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1574,7 +1588,31 @@ func (s *TASFlavorSnapshot) fillInCounts(requirements *topologyAssignmentPodRequ
 		domain.sliceState = 0
 		domain.sliceStateWithLeader = 0
 		domain.leaderState = 0
+		domain.affinityScore = 0
 	}
+
+	type compiledPreferredTerm struct {
+		selector *nodeaffinity.NodeSelector
+		weight   int64
+	}
```
mimowo: Is there no structure in the core k8s that could be used to represent the terms with weights?

mimowo: If not then at least move the struct decalation out of the function

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1574,7 +1588,31 @@ func (s *TASFlavorSnapshot) fillInCounts(requirements *topologyAssignmentPodRequ
...
+			}
+			if nodeSelector, err := nodeaffinity.NewNodeSelector(&corev1.NodeSelector{
+				NodeSelectorTerms: []corev1.NodeSelectorTerm{term.Preference},
+			}); err == nil {
+				compiledTerms = append(compiledTerms, compiledPreferredTerm{
+					selector: nodeSelector,
+					weight:   int64(term.Weight),
+				})
+			}
+		}
```
mimowo: Extract to a helper function to offload the main function with technical details

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1603,14 +1641,22 @@ func (s *TASFlavorSnapshot) fillInCounts(requirements *topologyAssignmentPodRequ
...
 				state.stats.Affinity++
 				continue
 			}
+
+			// 4. Calculate Affinity Score
+			for _, term := range compiledTerms {
+				if term.selector.Match(nodeObj) {
+					leaf.affinityScore += term.weight
+				}
+			}
```
mimowo, line 1671: wrap that in the new feature gate

pkg/cache/scheduler/tas_flavor_snapshot_test.go
```diff
@@ -737,3 +740,460 @@ func TestTruncateAssignment(t *testing.T) {
...
+			}
+
+			if diff := cmp.Diff(tc.wantOrder, gotValues); diff != "" {
+				t.Errorf("unexpected sorted domains (-want,+got): %s", diff)
+			}
+		})
+	}
+}
+
+func TestTASFindTopologyAssignments(t *testing.T) {
```
mimowo: Any reason to add the new test instead of extending the pre-existing one which it seems could cover that feature as well ;TestFindTopologyAssignments in tas_cache_test.go

pkg/cache/scheduler/tas_flavor_snapshot_test.go
```diff
@@ -737,3 +740,460 @@ func TestTruncateAssignment(t *testing.T) {
...
+			},
+			podSet: kueue.PodSet{
+				Name: "main",
+				TopologyRequest: &kueue.PodSetTopologyRequest{
+					Preferred: new("rack"),
+				},
+				Template: corev1.PodTemplateSpec{
+					Spec: corev1.PodSpec{
+						Affinity: &corev1.Affinity{
+							NodeAffinity: &corev1.NodeAffinity{
```
mimowo: Try to avoid such verbose declarations with the wrappers

pkg/cache/scheduler/tas_flavor_snapshot.go
j-skiba: added v(5) error log

pkg/cache/scheduler/tas_flavor_snapshot.go
j-skiba: changed to `*nodeaffinity.PreferredSchedulingTerms`

pkg/cache/scheduler/tas_flavor_snapshot.go
j-skiba: removed completely after moving to `*nodeaffinity.PreferredSchedulingTerms`

pkg/cache/scheduler/tas_flavor_snapshot.go
j-skiba: done

pkg/cache/scheduler/tas_flavor_snapshot_test.go
j-skiba: oh, I thought just to simply have it there as changes were made in tas_flavor_snapshot.go.

I moved the tests to tas_cache_test.go now

pkg/cache/scheduler/tas_flavor_snapshot_test.go
j-skiba: done

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1747,6 +1749,25 @@ For a 3-level topology (block, rack, hostname), in the TopologyRequest, the user
...
 | [[15], [15], [15, 15]] | 25|5 |  [[0], [0], [15, 10]] | `podset-slice-required-topology = hostname`
+
+### Support for Preferred Node Affinity
+
+TAS supports evaluating `preferredDuringSchedulingIgnoredDuringExecution`
+node affinities during the scheduling cycle. When the
+`TASPreferredSchedulingAffinity` feature gate is enabled, the preferred
+affinity scores are evaluated when selecting candidate topology domains,
+allowing affinity preferences to guide placement decisions. Preferred affinity
+takes precedence over capacity fitness of a domain.
```
mimowo:
```suggestion
In 0.18 we introduce the pilot support for domain affinity based on `preferredDuringSchedulingIgnoredDuringExecution`, behind the 
`TASRespectNodeAffinityPreferred` feature gate (alpha).

When the feature gate is enabled, we compute the "domain affinity scores" at all levels by
aggregating (summing) the scores from child topology domains.

Then, the "domain affinity scores" takes precedence over standard placement ordering of domains (based on BestFit or LeastFreeCapacity policies).

This feature only works when the lowest topology level specified in the Topology CRD is `kubernetes.io/hostname`.
```
To include the key informations:
- propose to rename as TASRespectNodeAffinityPreferred to reflect we are "respecting" api which is already there in API
- the version when the new semantic is introduced
- the mechanism of aggregation and evaluation in a bit more detail so that someone can imagine impl just by text
- make it clear the feature only works when "hostname" is the lowest level.

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1747,6 +1749,25 @@ For a 3-level topology (block, rack, hostname), in the TopologyRequest, the user
...
+
+#### Future Vision
+The long-term goal is to decommission this strict precedence model and its
+associated `TASPreferredSchedulingAffinity` feature gate in favor of a generic
+weighted scoring model of domains. Similar to the kube-scheduler scheduling
+plugins scoring model, candidate topology domains will be evaluated using
+a configurable list of individual scoring components (such as affinity
+preferences and capacity packing efficiency) that contribute to a normalized,
+unified final score to determine the globally optimal placement.
```
mimowo, line 1782:
```suggestion
#### Future Vision
One drawback of the proposed solution is that it introduces a strict precedence between the 
"domain affinity score" and the "placement policy". 

A more generic approach using weighted scoring model of domains is considered, similar to
the kube-scheduler scheduling plugins scoring model, candidate topology domains will be
evaluated using a configurable list of individual scoring components (such as affinity
preferences and capacity packing efficiency) that contribute to a normalized,
unified final score to determine the globally optimal placement.

We are going to re-evaluate the next steps based on the user and community feedback.
```

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1916,6 +1937,9 @@ Consider the following improvements and implement if feasible:
 - perform full scheduling simulation rather than just capacity counting
  (including pod affinities and anti-affinities)
 - drop `TASLeastAllocated` feature gate
+- drop `TASPreferredSchedulingAffinity` feature gate upon decommissioning the
+ strict precedence affinity model in favor of the Weighted Scoring Model of
+ domains (see [Future Vision](#future-vision))
```
mimowo:
```suggestion
- re-evaluate the status of the `TASPreferredSchedulingAffinity` feature gate - either graduate, or decommission with the more generic model of weighted scores (see [Future Vision](#future-vision))
```

keps/2724-topology-aware-scheduling/kep.yaml
```diff
@@ -47,6 +47,7 @@ feature-gates:
   - name: TASBalancedPlacement
   - name: TASReplaceNodeOnNodeTaints
   - name: TASMultiLayerTopology
+  - name: TASPreferredSchedulingAffinity
```
mimowo: I would propose TASRespectNodeAffinityPreferred to reflect the fact this is focused on respecting the exting k8s api, rather than entirely new feature  / api.

test/integration/singlecluster/tas/tas_test.go
```diff
@@ -3586,6 +3586,129 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
...
+							Obj(),
+					}
+					util.CreateNodesWithStatus(ctx, k8sClient, nodes)
+				})
+
+				ginkgo.AfterEach(func() {
+					for _, node := range nodes {
+						util.ExpectObjectToBeDeleted(ctx, k8sClient, &node, true)
+					}
+				})
```
mimowo: Hm, I remember some of the When / Descibe sections already had the code to cleanup the nodes created by the suite. Ptal maybe the nodes are already cleared at the higher level

pkg/features/kube_features.go
```diff
@@ -382,6 +382,12 @@ const (
 
 	// Enable recording of WorkloadCreationLatency metric.
 	MetricForWorkloadCreationLatency featuregate.Feature = "MetricForWorkloadCreationLatency"
+
+	// owner: @j-skiba
+	// issue: https://github.com/kubernetes-sigs/kueue/issues/10902
+	// Enable evaluation of preferredDuringSchedulingIgnoredDuringExecution node affinities during scheduling in TAS.
+	// The preferred affinity will take precedence over capacity fitness of a domain.
```
```suggestion
	// The preferred affinity will take precedence over the placement policy (eg. BestFit , LeastFreeCapacity).
```

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1365,7 +1365,7 @@ The annotation contains the name of (a single) failed node to replace.
 We propose to remove the Annotation and replace it with new field in Workload status that will it will contain
 a list of failed nodes:
 
-```golang
+```golangS
```
mimowo: typo to add "S"?

keps/2724-topology-aware-scheduling/README.md
j-skiba: done

keps/2724-topology-aware-scheduling/README.md
j-skiba: done

keps/2724-topology-aware-scheduling/README.md
j-skiba: done

keps/2724-topology-aware-scheduling/kep.yaml
j-skiba: renamed to `TASRespectNodeAffinityPreferred`

test/integration/singlecluster/tas/tas_test.go
j-skiba: now I moved the whole `When` block outside of `When("Node is mutated during test cases")` as it was not the best place for the tests.

Also added richer `AfterEach` cleanup procedures (the same as other `When` blocks) and added creation of topology, rf, cq etc to `BeforeEach` as it was necessary after moving out `When("Node is mutated during test cases")` block.

pkg/features/kube_features.go
j-skiba: done

keps/2724-topology-aware-scheduling/README.md
j-skiba: removed

pkg/cache/scheduler/tas_cache_test.go
```diff
@@ -40,16 +40,18 @@ import (
...
-	wantAssignment     *tas.TopologyAssignment
-	wantReason         string
+	podSetName                string
+	topologyRequest           *kueue.PodSetTopologyRequest
+	requests                  resources.Requests
+	count                     int32
+	tolerations               []corev1.Toleration
+	nodeSelector              map[string]string
+	requiredNodeSelectorTerms []corev1.NodeSelectorTerm
+	preferredSchedulingTerms  []corev1.PreferredSchedulingTerm
```
mimowo: nit, maybe we could encapsulate both just by `nodeAffinity`?

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -894,6 +899,15 @@ func (s *TASFlavorSnapshot) findTopologyAssignment(
 			}
 			requirements.affinitySelector = affinitySelector
 		}
+		preferredAffinity := info.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution
+		if features.Enabled(features.TASRespectPreferredSchedulingAffinity) && len(preferredAffinity) > 0 {
+			prefTerms, err := nodeaffinity.NewPreferredSchedulingTerms(preferredAffinity)
+			if err != nil {
+				s.log.V(5).Info("ignoring invalid preferred node affinity terms", "err", err)
```
mimowo: Let's rather `return` here as for the errors when parsing `requiredAffinity`.

mimowo:
LGTM, left remaining comments. 

cc @tenzen-y in case you would like to review the KEP updates here

pkg/cache/scheduler/tas_flavor_snapshot.go
j-skiba: done

pkg/cache/scheduler/tas_cache_test.go
j-skiba: yeah, it looks better. Done

keps/2724-topology-aware-scheduling/kep.yaml
```diff
@@ -47,6 +47,7 @@ feature-gates:
   - name: TASBalancedPlacement
   - name: TASReplaceNodeOnNodeTaints
   - name: TASMultiLayerTopology
+  - name: TASRespectNodeAffinityPreferred
```
tenzen-y, line 50: Which FG names `TASRespectPreferredSchedulingAffinity` vs `TASRespectNodeAffinityPreferred` are correct?

I see that you use separate FG names in KEP and implementation.

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1512,6 +1527,10 @@ func (s *TASFlavorSnapshot) sortedDomainsWithLeader(domains []*domain, unconstra
 	isLeastFreeCapacity := useLeastFreeCapacityAlgorithm(unconstrained)
 	result := slices.Clone(domains)
 	slices.SortFunc(result, func(a, b *domain) int {
+		if features.Enabled(features.TASRespectPreferredSchedulingAffinity) && a.affinityScore != b.affinityScore {
+			return cmp.Compare(b.affinityScore, a.affinityScore)
+		}
```
tenzen-y: Why do we need to use the preferred affinity score evaluation as a first-step tiebreak?
I don't see any context in KEP.

Doesn't this break the existing topology assignment for the leader on multiple domains?

I'm wondering if we should use this as the second tiebreak.

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -894,6 +899,16 @@ func (s *TASFlavorSnapshot) findTopologyAssignment(
 			}
 			requirements.affinitySelector = affinitySelector
 		}
+		preferredAffinity := info.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution
+		if features.Enabled(features.TASRespectPreferredSchedulingAffinity) && len(preferredAffinity) > 0 {
+			prefTerms, err := nodeaffinity.NewPreferredSchedulingTerms(preferredAffinity)
+			if err != nil {
+				s.log.V(5).Info("ignoring invalid preferred node affinity terms", "err", err)
```
mimowo: Let's drop the log - it should be logged already in the higher levels as we are returning error.

tenzen-y (commented):
Thank you for working on that.
I left a small but critical algorithm change comment.

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -894,6 +899,16 @@ func (s *TASFlavorSnapshot) findTopologyAssignment(
 			}
 			requirements.affinitySelector = affinitySelector
 		}
+		preferredAffinity := info.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution
```
mimowo: Do not read it outside the FG, we can do 

```golang
if features.Enabled(features.TASRespectPreferredSchedulingAffinity) {
  preferredAffinity := info.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution
  if len(preferredAffinity) == 0 {
    continue
  }
  ...
}
```
just to make it very clear to users that we are not getting any nil pointer

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1512,6 +1527,10 @@ func (s *TASFlavorSnapshot) sortedDomainsWithLeader(domains []*domain, unconstra
 	isLeastFreeCapacity := useLeastFreeCapacityAlgorithm(unconstrained)
 	result := slices.Clone(domains)
 	slices.SortFunc(result, func(a, b *domain) int {
+		if features.Enabled(features.TASRespectPreferredSchedulingAffinity) && a.affinityScore != b.affinityScore {
+			return cmp.Compare(b.affinityScore, a.affinityScore)
+		}
+
```
mimowo: I would like to also make sure we have tests for the scenario (maybe we already have):
`shoud choose the second domain (ordering by  `affinityScore`), because the first domain does not satisfy the "required" constraint`; 

This is in order to make sure the code does not give up on the search too early if the first domain does not satisfy the request. 

Now the largest (and only domain satisfying the "required" constaint) may not be first.

Historically the first was the largest, so I think we might have done such an optimization. For example we have some special hanling about the topDomain here: https://github.com/kubernetes-sigs/kueue/blob/2edaef59111e7a06a5b55488a9c8d15e39eae668/pkg/cache/scheduler/tas_flavor_snapshot.go#L1246

pkg/cache/scheduler/tas_flavor_snapshot.go
mimowo: First step tie-break is what we need for our use case.

Basically we have a performance costly procedure for preparing infrastructure (in an AdmissionCheck) on some "racks", and so we want to prioritize the racks with already prepared infrastucture as much as possible.

If we cannot schedule on the racks with "hot" infrastructure we may fallback to another "rack" and warm it up, but this is costly, so is avoided, even at the cost of fragmentation.

Stiil, in a follow up we may need a more weighted approach when we can say that nodeAffinity is 90% important, but fragmentation 10%. However, we are not able to deliver that for 0.18.  

This is partly already mentioned in the KEP, but we can extend the "User stories" section with this case, where the node preference is much more important.

tenzen-y: I see. In that case, I'm ok with using affinity as the first tiebreak.
OTOH, this will change LWS TAS behavior. Before this feature, the leader would be placed in the best topology even though the scheduled place didn't match the affinity.

But, after this feature, the leader doesn't get an assignment when the assigned topology doesn't match the affinity.

So, I would propose mentioning this behavior change in the KEP and user documentation.
Also, I'm wondering if we can expand notFitReason so that we can emphasize affinity is the no fit reason.

mimowo: Let me think tomorrow, maybe there are such cases

mimowo: > OTOH, this will change LWS TAS behavior. Before this feature, the leader would be placed in the best topology even though the scheduled place didn't match the affinity.
> But, after this feature, the leader doesn't get an assignment when the assigned topology doesn't match the affinity.

After reading this again, I'm not sure I understand the concern. Yes, the domain for the leader might be different than before, but:
1. users who previously set ` nodeAffinity.preferred` it was completely ignored for them so equally well they could just skip it
2. users who don't use nodeAffinity.preferred are not affected
3. for users who now use `nodeAffinity.preferred` the domain for leader may change, but if the domain could be found before, then it will also be found now. We only change soft preference. So, the feature should not result in `noFit` now.

> So, I would propose mentioning this behavior change in the KEP and user documentation.
Also, I'm wondering if we can expand notFitReason so that we can emphasize affinity is the no fit reason.

Given my point above (3.) this should never happen with this feature that we found a domain before, but cannot find any domain now. We may just find an alternative / different placement.

So, I think we may:
1. clarify that in the text of the KEP
2. extend testing to demonstrate the tricky cases are covered, which is also my suggestion here: https://github.com/kubernetes-sigs/kueue/pull/10903#discussion_r3235820573

@j-skiba correct me if I'm missing something.

@tenzen-y let me also know if you meant something different, cause I'm not totally sure if I captured your concern well

j-skiba: yes, I agree with these statements

pkg/cache/scheduler/tas_flavor_snapshot.go
j-skiba: removed

pkg/cache/scheduler/tas_flavor_snapshot.go
j-skiba: :+1: done like this
```go
if features.Enabled(features.TASRespectPreferredSchedulingAffinity) {
  preferredAffinity := info.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution
  if len(preferredAffinity) > 0 {
    ...
  }
}
```
because there's no loop

pkg/cache/scheduler/tas_flavor_snapshot.go
j-skiba: added two unit test cases:
- `affinity ordering with preferred inner topology level bubbles up to parent level to avoid splitting across parent domains`
- `affinity ordering with required topology level chooses secondary domain if primary is too small`

and I also had to adjust the code a little to handle them

keps/2724-topology-aware-scheduling/kep.yaml
j-skiba: good catch, changed to `TASRespectNodeAffinityPreferred` everywhere

pkg/cache/scheduler/tas_flavor_snapshot.go
mimowo: I would like to add another test case - looking at the code I'm not sure if this works well. 


1.  flat topology (only hostname), workload with the "preferred" mode, size 5 pods x 1
- preferred nodes with sizes: [2, 1]
- non-preferred nodes with sizes: [2, 1]
2.  flat topology (only hostname), workload with the "unconstrained" mode, size 5 pods x 1:
- prefferred nodes with sizes: [2, 1]
- non-preferred nodes with sizes: [2, 1, 1]
3.  flat topology (only hostname), workload with the "preferred" mode, size 2 pods x 2
- preferred nodes with sizes: [2, 1]
- non-preferred nodes with sizes: [2, 2, 1]
4.  flat topology (only hostname), workload with the "unconstrained" mode, size 2 pods x 2
- preferred nodes with sizes: [2, 1]
- non-preferred nodes with sizes: [2, 2, 1]

My expecations:
- in (1.) we choose use two node from preferred, and one node from non-preferred of size 2
- in (2.) we should use two nodes from preferred, and two of size "1" from the non-preferred
- in (3.) we should use one node from preferred, and one of size "2" from the non-preferred
- in (4.) we should use one node from preferred, and one of size "2" from the non-preferred

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1270,8 +1286,17 @@ func (s *TASFlavorSnapshot) findLevelWithFitDomains(
...
 	if topDomain.sliceStateWithLeader < sliceCount || topDomain.leaderState < state.leaderCount {
 		if state.required {
+			// Scan remaining domains to support preferred affinity before failing
+			for i := 1; i < len(sortedDomain); i++ {
+				d := sortedDomain[i]
+				if d.sliceStateWithLeader >= sliceCount && d.leaderState >= state.leaderCount {
+					candidates := topAffinityTierDomains(sortedDomain[i:])
+					return searchLevelIdx, []*domain{findBestFitDomainForSlices(candidates, sliceCount, state.leaderCount)}, ""
+				}
+			}
```
mimowo, line 1297: Is the new code fully covered by unit tests?

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1270,8 +1286,17 @@ func (s *TASFlavorSnapshot) findLevelWithFitDomains(
 	}
 	if topDomain.sliceStateWithLeader < sliceCount || topDomain.leaderState < state.leaderCount {
 		if state.required {
+			// Scan remaining domains to support preferred affinity before failing
```
mimowo, line 1289: Wrap the code in the FG

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1270,8 +1286,17 @@ func (s *TASFlavorSnapshot) findLevelWithFitDomains(
...
+			for i := 1; i < len(sortedDomain); i++ {
+				d := sortedDomain[i]
+				if d.sliceStateWithLeader >= sliceCount && d.leaderState >= state.leaderCount {
+					candidates := topAffinityTierDomains(sortedDomain[i:])
+					return searchLevelIdx, []*domain{findBestFitDomainForSlices(candidates, sliceCount, state.leaderCount)}, ""
+				}
+			}
 			return 0, nil, notFitReason(topDomain.sliceState, sliceCount)
 		}
+
```
mimowo: nit, revert line changes

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -86,6 +86,10 @@ type domain struct {
 	// levelValues stores the mapping from domain ID back to the
 	// ordered list of values
 	levelValues []string
+
+	// affinityScore is the sum of weights of all preferred affinity terms that match the node.
+	// For non-leaf domains, it is the sum of affinity scores of all children.
+	affinityScore int64
```
mimowo, line 92: Check code coverage with some tooling for hew newly added code to confirm 100% coverage

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1288,7 +1313,8 @@ func (s *TASFlavorSnapshot) findLevelWithFitDomains(
 			domain := sortedDomain[idx]
 			if useBestFitAlgorithm(state.unconstrained) && sortedDomain[idx].sliceStateWithLeader >= remainingSliceCount {
 				// optimize the last domain
-				domain = findBestFitDomainForSlices(sortedDomain[idx:], remainingSliceCount, remainingLeaderCount)
+				candidates := topAffinityTierDomains(sortedDomain[idx:])
+				domain = findBestFitDomainForSlices(candidates, remainingSliceCount, remainingLeaderCount)
 			}
```
j-skiba, line 1317: this part doesn't have any coverage and I think it's dead code.

here 
https://github.com/kubernetes-sigs/kueue/pull/10903/changes#diff-015ac0ccc6a319e922161cdd677487236e5d4e7edc11d0376fd59aebf14ae148L1271
let's say we have workload with `leaderCount = 1` and `sliceCount = 3` and `topDomain` can fit the leader pod but doing so leaves enough space for only 2 worker slices.

then we move here
https://github.com/kubernetes-sigs/kueue/pull/10903/changes#diff-015ac0ccc6a319e922161cdd677487236e5d4e7edc11d0376fd59aebf14ae148L1279
and we assign
```go
remainingSliceCount := 3
remainingLeaderCount := 1 // this is always 1 or 0
```

then we enter `for` loop
https://github.com/kubernetes-sigs/kueue/pull/10903/changes#diff-015ac0ccc6a319e922161cdd677487236e5d4e7edc11d0376fd59aebf14ae148L1287
and this `sortedDomain[idx].sliceStateWithLeader >= remainingSliceCount` will always be false because the first iteration always has `idx = 0` so we use `sortedDomains[0]` but here
https://github.com/kubernetes-sigs/kueue/pull/10903/changes#diff-015ac0ccc6a319e922161cdd677487236e5d4e7edc11d0376fd59aebf14ae148L1271
we entered the condition because of the exact opposite reason (`topDomain.sliceStateWithLeader < sliceCount`) so it evaluates to false.

Note: If `topDomain.leaderState` is 0 here
https://github.com/kubernetes-sigs/kueue/pull/10903/changes#diff-015ac0ccc6a319e922161cdd677487236e5d4e7edc11d0376fd59aebf14ae148L1271
then `for` loop condition fails to even enter for first iteration

Then 
https://github.com/kubernetes-sigs/kueue/pull/10903/changes#diff-015ac0ccc6a319e922161cdd677487236e5d4e7edc11d0376fd59aebf14ae148L1295
here we subtract and get 0 and the loop ends

Am I right?

j-skiba: cc @mimowo

pkg/cache/scheduler/tas_flavor_snapshot.go
j-skiba: done

pkg/cache/scheduler/tas_flavor_snapshot.go
j-skiba: done

pkg/cache/scheduler/tas_flavor_snapshot.go
j-skiba: done, passes without additional changes

pkg/cache/scheduler/tas_flavor_snapshot.go
j-skiba: yes except for this https://github.com/kubernetes-sigs/kueue/pull/10903#discussion_r3240806355 but it wasn't covered even before

pkg/cache/scheduler/tas_flavor_snapshot.go
mimowo: Let me check with @pajakd first

pajakd: I synced with @j-skiba and I also think its dead code (at least now when the number of leaders is at most 1).

https://github.com/kubernetes-sigs/kueue/blob/b2a2dd8336aed252bc9fea7a0fbec3c4366fe8c5/pkg/cache/scheduler/tas_flavor_snapshot.go#L1287-L1301

 - `sortedDomain[idx].leaderState > 0` so in each iteration of this loop we only consider domains that can host a leader.
 - `remainingLeaderCount -= domain.leaderState` so in each iteration of this loop we decrease the `remainingLeaderCount`.
 - we stop the loop when  `remainingLeaderCount` is 0 so this loop can have at most one iteration.
 - if in this first iteration we would have `sortedDomain[idx].sliceStateWithLeader >= remainingSliceCount`, this would mean that this single domain can host the leader and all the slices. This cannot happen as we are in the if `topDomain.sliceStateWithLeader < sliceCount || topDomain.leaderState < state.leaderCount` so the topDomain cannot fit everything.

So, basically I would consider removing this code (in a separate PR).

pajakd: But, this is a complicated logic. So I would like to see some tests. For example. If the preferred affinity would move the domain that cannot host a leader to be the first domain, we will not execute this loop at all.

mimowo: Thank you for the investigation folks 👍 

Let's leave this PR as is, but maybe open another PR which turns the code into os.Exit(-1) to test if this is maybe tested in integration /e2e tests. If we don't have any coverage signal I'm ok to drop in a follow up PR.

pkg/cache/scheduler/tas_flavor_snapshot.go
j-skiba: apart from [this](https://github.com/kubernetes-sigs/kueue/pull/10903#discussion_r3240806355) and - https://github.com/kubernetes-sigs/kueue/pull/10903/changes#diff-015ac0ccc6a319e922161cdd677487236e5d4e7edc11d0376fd59aebf14ae148R1415  - (which doesn't have coverage too) everything is covered by tests.

Should I add short unit test to just cover this `else` path?

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1306,7 +1333,8 @@ func (s *TASFlavorSnapshot) findLevelWithFitDomains(
 			domain := sortedDomain[idx]
 			if useBestFitAlgorithm(state.unconstrained) && sortedDomain[idx].sliceState >= remainingSliceCount {
 				// optimize the last domain
-				domain = findBestFitDomainForSlices(sortedDomain[idx:], remainingSliceCount, 0)
+				candidates := topAffinityTierDomains(sortedDomain[idx:])
+				domain = findBestFitDomainForSlices(candidates, remainingSliceCount, 0)
```
mimowo: One more question, these two lines `topAffinityTierDomains` and `findBestFitDomainForSlices` seem to always go together, could we put `topAffinityTierDomains` inside `findBestFitDomainForSlices` to avoid duplicating the code?

mimowo: If there are some counter-examples them let's make them explictig here, and let's ponder if we have enough of them to justify the duplication, or we could maybe add a param to indicate if `topAffinityTierDomains` should be used

pkg/cache/scheduler/tas_flavor_snapshot.go
mimowo: > Should I add short unit test to just cover this else path?

It would be great as you are now fresh with understanding of the code.

pkg/cache/scheduler/tas_flavor_snapshot.go
tenzen-y: @j-skiba @mimowo Let's imagine the following Cross PodSchedulig LWS and cluster capacities.

```
- us-west
  - number of host: 8
  - each host allocatable resources: CPU=8, memory=32Gi, GPU=8
- us-east
  - number of host: 4
  - each host allocatable resources: CPU=32, memory=128Gi, GPU=8
```

If we respect worker preferred affinity by this FG enabling, a whole of LWS will not be scheduled because they specify us-west as preferred affinity.

In algorithm PoV, before this PR, we could select us-east because the `leaderState` is satisfied.
But, after this PR, we can not select us-east due to affinityScore even though leader can not be accommodated by us-west.

As a result, a whole LWS doesn't get any assignment due to PodSetGroup constraints.

```
- topolotgy: us-west
    leaderStete: 0 # because each us-west machine has only 8 core / 32 Gi mem.
    affinityScore: 10
- topology: us-east
    leaderState: 1
    affinityScore: 0
```

I think that this is critical behavior changes. In this situation, the worker's preferred affinity works as strong constraints even though it is just preference which means we will face no-fit.

```yaml
apiVersion: leaderworkerset.x-k8s.io/v1
kind: LeaderWorkerSet
metadata:
  name: llm-serving
  labels:
    kueue.x-k8s.io/queue-name: tas-queue
spec:
  replicas: 1
  leaderWorkerTemplate:
    size: 9
    leaderTemplate:
      metadata:
        annotations:
          kueue.x-k8s.io/podset-group-name: "llm-group"
          kueue.x-k8s.io/podset-preferred-topology: "topology.kubernetes.io/zone"
      spec:
        containers:
        - name: router
          resources:
            requests:
              cpu: "16"
              memory: "64Gi"
    workerTemplate:
      metadata:
        annotations:
          kueue.x-k8s.io/podset-group-name: "llm-group"
          kueue.x-k8s.io/podset-preferred-topology: "topology.kubernetes.io/zone"
      spec:
        affinity:
          nodeAffinity:
            preferredDuringSchedulingIgnoredDuringExecution:
            - weight: 10
              preference:
                matchExpressions:
                - key: topology.kubernetes.io/zone
                  operator: In
                  values: ["us-west"]
        containers:
        - name: model-shard
          resources:
            requests:
              cpu: "4"
              memory: "32Gi"
              nvidia.com/gpu: "8"
```

mimowo: @tenzen-y  I see, this is an excellent example 👍 

This shows that indeed we may turn "fit" into "nofit" with `nodeAffinity.preferredDuringSchedulingIgnoredDuringExecution` + `kueue.x-k8s.io/podset-group-name` used.

In that case, the approaches I can see:
1. just keep as is, but document the behavior, because this is a very special case, users just need to know the risk
2. ignore  `nodeAffinity.preferredDuringSchedulingIgnoredDuringExecution `  when `kueue.x-k8s.io/podset-group-name`  is used (as before basically, just respect `nodeAffinity.preferredDuringSchedulingIgnoredDuringExecution ` for no-group scenarios
3. if we hit "noFit" then re-run the whole algorithm ignoring  `nodeAffinity.preferredDuringSchedulingIgnoredDuringExecution` , so that the "fit" solution is found
4. add validation against using `nodeAffinity.preferredDuringSchedulingIgnoredDuringExecution ` with `kueue.x-k8s.io/podset-group-name` 

I'm leaning towards (1.), because:
- the case is rather specially crafted, typically the leader will have the same resources( is one of workers), or small (just small CPU to coordinate), so this case is rather theoretical IMO
- (2.) may restrict user options - if they want to ignore, then they can just not set the nodeAffinity.preferred
- (3.) may mean excessive performance hit, and perforamnce of TAS is very important
- (4.) is sensible, but still restricting the use-case for the case which many users will consider theoretical
- note that even in the baseline TAS we sometimes say "nofit" even though we could say "fit" if we considered PodSets in a different order (even without `kueue.x-k8s.io/podset-group-name`), so the algorithm is intrinsically heuristic anyway already

So, I would suggest (1.), just make sure the documentation is mentioning this possible case.

j-skiba: ok so actually moving the affinity check to the second position (after `leaderState` check) should be ok even for https://github.com/kubernetes-sigs/kueue/pull/10903#discussion_r3235862873 this scenario I think.

The `leaderState` is a binary value so even when domains are sorted then the affinity tie-breaker will guarantee the leader-capable domains sublist of all domains will be sorted by their affinity score.

This way we have the scenario @tenzen-y described handled as before and it also satisfies the scenario from the comment https://github.com/kubernetes-sigs/kueue/pull/10903#discussion_r3235862873

mimowo: cool in that case lets try to fix this scenario

tenzen-y: @mimowo @j-skiba Thank you for moving this discussion forward.
@jackfrancis 's new approach sounds good.

As I check the algorithm again, even when LWS leader replicas are multiple, the `leaderSate` will be binary.
So, I believe that both my scenario and @mimowo 's customer request (https://github.com/kubernetes-sigs/kueue/pull/10903#discussion_r3235862873) will be satisfied.

Awesome, thank you.

j-skiba: I moved the affinity score check as the second fallback, also removed section from README added before

pkg/cache/scheduler/tas_flavor_snapshot.go
j-skiba: done

pkg/cache/scheduler/tas_flavor_snapshot.go
j-skiba: `findBestFitDomainForSlices` and `findBestFitDomain` call `findBestFitDomainBy` inside. I moved the `topAffinityTierDomains` into `findBestFitDomainBy`

mimowo (commented):
/lgtm
/approve

Thank you 👍 
Leaving on hold in case @tenzen-y has extra comments
/hold

pkg/cache/scheduler/tas_flavor_snapshot_test.go
tenzen-y, line 438: Can we add `TestSrotedDomains(t *testing.T)` which has affinity cases same as `TestSortedDomainsWithLeader`?

test/integration/singlecluster/tas/tas_test.go
```diff
@@ -3416,6 +3416,152 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
...
+						Label("region", "us-west").
+						Label("zone", "eu").
+						StatusAllocatable(corev1.ResourceList{
+							corev1.ResourceCPU:    resource.MustParse("1"),
+							corev1.ResourceMemory: resource.MustParse("1Gi"),
+							corev1.ResourcePods:   resource.MustParse("10"),
+						}).
+						Ready().
+						Obj(),
+				}
```
tenzen-y, line 3470: Currently, we have "node-best", "node-good", and "node-bad". The only "node-best" and "node-bad" have "region=us-west," which is the preferred affinity in the "should admit workload to the node matching preferred affinity" case.

So, both "node-best" and "node-bad" match with the preferred affinity, then "node-bad" is selected due to alphabetical order.

It would be better to prepare Nodes so that exactly one Node matches the preferred affinity.
So, I would recommend using the `region=us-east` in `node-bad`.

tenzen-y (commented):
Almost LGTM, thank you!

test/integration/singlecluster/tas/tas_test.go
j-skiba: sounds good, done

pkg/cache/scheduler/tas_flavor_snapshot_test.go
j-skiba: done

tenzen-y (commented):
Awesome! Thank you!
/lgtm
/approve

/hold cancel

PR #10917, opened 2026-05-04
[ConcurrentAdmission] Skip less favorable Variants from being admitted

### LINKED ISSUE #10913: [Concurrent Admission] Scheduler can admit less favorable Variant while more favorable Variant is running
**What happened**:
With the Concurrent Admission feature, the less favorable Variants are deactivated by the concurrentadmission controller when a more favorable Variant is admitted. The controller runs in parallel to Kueue scheduler and there might be an interleaving where:
1) Scheduler admits more favorable Variant
2) Scheduler admits less favorable Variant
3) ConcurrentAdmission controller deactivates less favorable Variant

There's eventual correctness, but we want to eliminate step 2) by adding some checks to scheduler logic that will make sure that if there's more favorable Variant running, scheduler doesn't admit the less favorable one. 

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PBundyra:
#### What type of PR is this?
/kind bug

Skips less favorable Variants from being admitted, if the more favorable Variant is running. Eventually the less favorable Variant is deactivated by the concurrentadmission controller.

Fixes #10913

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -383,14 +383,8 @@ func (s *Scheduler) processEntry(
 	ctx = ctrl.LoggerInto(ctx, log)
 	log.V(2).Info("Attempting to schedule workload")
 
-	if features.Enabled(features.ConcurrentAdmission) {
-		moreFavorableSibling, err := s.findAdmittedMoreFavorableSibling(&e.Info, snapshot)
-		if err != nil {
-			log.Error(err, "Failed to check for more favorable sibling")
-			e.markSkipped(fmt.Sprintf("Error checking for more favorable sibling: %v", err))
-			return
-		}
-		if moreFavorableSibling != nil {
+	if features.Enabled(features.ConcurrentAdmission) && concurrentadmission.IsVariant(e.Obj) {
+		if moreFavorableSibling := s.findAdmittedMoreFavorableSibling(&e.Info, snapshot); moreFavorableSibling != nil {
 			log.V(3).Info("Skipping workload as a more favorable variant is already admitted", "moreFavorableVariant", klog.KObj(moreFavorableSibling.Obj))
 			e.markSkipped("A more favorable variant is already admitted")
 			return
@@ -459,14 +453,8 @@ func (s *Scheduler) processEntry(
 		}
 	}
 
-	if features.Enabled(features.ConcurrentAdmission) {
-		lessFavorableSibling, err := s.findAdmittedLessFavorableSibling(&e.Info, snapshot)
-		if err != nil {
-			log.Error(err, "Failed to check for less favorable sibling")
-			e.markSkipped(fmt.Sprintf("Error checking for less favorable sibling: %v", err))
-			return
-		}
-		if lessFavorableSibling != nil {
+	if features.Enabled(features.ConcurrentAdmission) && concurrentadmission.IsVariant(e.Obj) {
+		if lessFavorableSibling := s.findAdmittedLessFavorableSibling(&e.Info, snapshot); lessFavorableSibling != nil {
 			s.issueMigration(ctx, log, e, lessFavorableSibling)
 			return
 		}
@@ -1122,7 +1110,7 @@ func (s *Scheduler) updateEntryPenalty(log logr.Logger, e *entry, op usageOp) {
 
 // findAdmittedLessFavorableSibling returns an admitted Sibling Workload that has a
 // less favorable Resource Flavor assignment than the provided Workload.
-func (s *Scheduler) findAdmittedLessFavorableSibling(wl *workload.Info, snap *schdcache.Snapshot) (*workload.Info, error) {
+func (s *Scheduler) findAdmittedLessFavorableSibling(wl *workload.Info, snap *schdcache.Snapshot) *workload.Info {
 	return s.findAdmittedSibling(wl, snap, func(siblingIdx, wlIdx int) bool {
 		return siblingIdx > wlIdx // larger index means less favorable
 	})
@@ -1130,7 +1118,7 @@ func (s *Scheduler) findAdmittedLessFavorableSibling(wl *workload.Info, snap *sc
 
 // findAdmittedMoreFavorableSibling returns an admitted Sibling Workload that has a
 // more favorable Resource Flavor assignment than the provided Workload.
-func (s *Scheduler) findAdmittedMoreFavorableSibling(wl *workload.Info, snap *schdcache.Snapshot) (*workload.Info, error) {
+func (s *Scheduler) findAdmittedMoreFavorableSibling(wl *workload.Info, snap *schdcache.Snapshot) *workload.Info {
 	return s.findAdmittedSibling(wl, snap, func(siblingIdx, wlIdx int) bool {
 		return wlIdx > siblingIdx
 	})
@@ -1141,36 +1129,33 @@ func (s *Scheduler) findAdmittedMoreFavorableSibling(wl *workload.Info, snap *sc
 // A Sibling is a Workload belonging to the same Parent Workload (part of the
 // Concurrent Admission feature). Favorability is determined by the order of
 // Resource Flavors defined in the ClusterQueue
-func (s *Scheduler) findAdmittedSibling(wl *workload.Info, snap *schdcache.Snapshot, compareFunc func(siblingIdx, wlIdx int) bool) (*workload.Info, error) {
+func (s *Scheduler) findAdmittedSibling(wl *workload.Info, snap *schdcache.Snapshot, compareFunc func(siblingIdx, wlIdx int) bool) *workload.Info {
 	if !features.Enabled(features.ConcurrentAdmission) {
-		return nil, nil
+		return nil
 	}
 	parentUID := concurrentadmission.GetParentWorkloadUID(wl.Obj)
 	if parentUID == "" {
-		return nil, nil
+		return nil
 	}
 	cq := snap.ClusterQueue(wl.ClusterQueue)
-	if cq == nil {
-		return nil, fmt.Errorf("cluster queue %s not found", wl.ClusterQueue)
-	}
-	if len(cq.ResourceGroups) == 0 {
-		return nil, nil
+	if cq == nil || len(cq.ResourceGroups) == 0 {
+		return nil
 	}
 	flavors := cq.ResourceGroups[0].Flavors
 	wlFlavorIdx, err := getFlavorIndex(wl, flavors)
 	if err != nil {
-		return nil, err
+		return nil
 	}
-	return s.findAdmittedSiblingMatching(wl, cq, parentUID, func(sibling *workload.Info) (bool, error) {
+	return s.findAdmittedSiblingMatching(wl, cq, parentUID, func(sibling *workload.Info) bool {
 		siblingFlavorIdx, err := getFlavorIndex(sibling, flavors)
 		if err != nil {
-			return false, err
+			return false
 		}
-		return compareFunc(siblingFlavorIdx, wlFlavorIdx), nil
+		return compareFunc(siblingFlavorIdx, wlFlavorIdx)
 	})
 }
 
-func (s *Scheduler) findAdmittedSiblingMatching(wl *workload.Info, cq *schdcache.ClusterQueueSnapshot, parentUID types.UID, match func(sibling *workload.Info) (bool, error)) (*workload.Info, error) {
+func (s *Scheduler) findAdmittedSiblingMatching(wl *workload.Info, cq *schdcache.ClusterQueueSnapshot, parentUID types.UID, match func(sibling *workload.Info) bool) *workload.Info {
 	for _, otherWl := range cq.Workloads {
 		if otherWl.Obj.UID == wl.Obj.UID {
 			continue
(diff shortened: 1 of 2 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -1092,18 +1100,32 @@ func (s *Scheduler) getLessFavorableSibling(log logr.Logger, wl *workload.Info,
...
-				}
+			if siblingFlavor != "" && matchFunc(candidateFlavor, siblingFlavor, flavors) {
+				return otherWl
 			}
 		}
 	}
 	return nil
 }
 
+func (s *Scheduler) getLessFavorableSibling(log logr.Logger, wl *workload.Info, snap *schdcache.Snapshot) *preemption.Target {
```
mimowo: Please order the functions so that the caller is always higher, it is currently very jumpy when reading the code. I think `getSiblingVariant` should be below `getLessFavorableSibling`.

I know that some languages have the opposite listing convention, just trying to make sure there is some natural order.

pkg/scheduler/scheduler.go
```diff
@@ -1092,18 +1100,32 @@ func (s *Scheduler) getLessFavorableSibling(log logr.Logger, wl *workload.Info,
...
+		return nil
+	}
+	return &preemption.Target{
+		WorkloadInfo: sibling,
+		WorkloadCq:   snap.ClusterQueue(wl.ClusterQueue),
+		Reason:       kueue.WorkloadEvictedByFlavorMigration,
+	}
+}
+
+func (s *Scheduler) getMoreFavorableSibling(log logr.Logger, wl *workload.Info, snap *schdcache.Snapshot) *workload.Info {
```
mimowo: The semantics of this function is unclear to me, same for getLessFavorableSibling, because there might be more than one more/less preferreable sibling. Are you choosing the "most preffrable" / least preferrable? If so then let's capture the intention in naming. If we choose random then it is a code smell. Let's also add comments documenting the semantics for the input and ouput of these functions.

PBundyra: The assumption of this function is that there's only one more/less favorable sibling admitted. In those functions we iterate over Workloads stored in Snapshot, which iirc stores only admitted Workloads. Let me clarify that in naming/comments

pkg/scheduler/scheduler.go
PBundyra: Done

pkg/scheduler/scheduler.go
PBundyra: Renamed and added comments

mimowo: > The assumption of this function is that there's only one more/less favorable sibling admitted.

Hm, if we look for "admitted", then shouldn't we guarantee there is only one globally anyway? IN that case we could have a single function findAdmittedSiblling / findAdmittedVariant?

pkg/scheduler/scheduler.go
```diff
@@ -1066,7 +1074,39 @@ func (s *Scheduler) updateEntryPenalty(log logr.Logger, e *entry, op usageOp) {
...
 }
 
-func (s *Scheduler) getLessFavorableSibling(log logr.Logger, wl *workload.Info, snap *schdcache.Snapshot) *preemption.Target {
+// getAdmittedLessFavorableSibling returns an admitted Sibling Workload that has a
+// less favorable Resource Flavor assignment than the provided Workload.
+//
+// A Sibling is a Workload belonging to the same Parent Workload (part of the
+// Concurrent Admission feature). Favorability is determined by the order of
+// Resource Flavors defined in the ClusterQueue.
+func (s *Scheduler) getAdmittedLessFavorableSibling(log logr.Logger, wl *workload.Info, snap *schdcache.Snapshot) *preemption.Target {
```
mimowo: nit, let's use "find" prefixes, get suggests a simple get operation as if the value was preconputed. This is typical prefix used in the getter/setter pattern. Here we do some computations, so better to reflect that.

pkg/scheduler/scheduler.go
```diff
@@ -1066,7 +1074,39 @@ func (s *Scheduler) updateEntryPenalty(log logr.Logger, e *entry, op usageOp) {
...
+// Resource Flavors defined in the ClusterQueue.
+func (s *Scheduler) getAdmittedMoreFavorableSibling(log logr.Logger, wl *workload.Info, snap *schdcache.Snapshot) *workload.Info {
+	return s.getSiblingVariant(wl, snap, func(candidate, sibling kueue.ResourceFlavorReference, flavors []kueue.ResourceFlavorReference) bool {
+		// the order of the arguments is reversed in comparsinon to getAdmittedLessFavorableSibling
+		return isLessFavorable(sibling, candidate, flavors)
+	})
+}
+
+func (s *Scheduler) getSiblingVariant(wl *workload.Info, snap *schdcache.Snapshot,
+	matchFunc func(candidate, sibling kueue.ResourceFlavorReference, flavors []kueue.ResourceFlavorReference) bool) *workload.Info {
```
mimowo: Can we rename `matchFunc` to `lessFunc`? It seems to play this role, it would be much more flamilar naming. `match` to me suggests as "are some criteria matching", here we use it for comparision.

For comparing there are two popular patterns: `lessFunc` , or `compareFunc`. Compare returns  -1/0/1.

pkg/scheduler/scheduler.go
```diff
@@ -1066,7 +1074,39 @@ func (s *Scheduler) updateEntryPenalty(log logr.Logger, e *entry, op usageOp) {
...
+// Concurrent Admission feature). Favorability is determined by the order of
+// Resource Flavors defined in the ClusterQueue.
+func (s *Scheduler) getAdmittedMoreFavorableSibling(log logr.Logger, wl *workload.Info, snap *schdcache.Snapshot) *workload.Info {
+	return s.getSiblingVariant(wl, snap, func(candidate, sibling kueue.ResourceFlavorReference, flavors []kueue.ResourceFlavorReference) bool {
+		// the order of the arguments is reversed in comparsinon to getAdmittedLessFavorableSibling
+		return isLessFavorable(sibling, candidate, flavors)
+	})
+}
+
+func (s *Scheduler) getSiblingVariant(wl *workload.Info, snap *schdcache.Snapshot,
```
pajakd: The name of this function should perhaps specify that it returns the admitted sibling variant.

pkg/scheduler/scheduler.go
```diff
@@ -372,6 +372,14 @@ func (s *Scheduler) processEntry(
 	ctx = ctrl.LoggerInto(ctx, log)
 	log.V(2).Info("Attempting to schedule workload")
 
+	if features.Enabled(features.ConcurrentAdmission) {
+		if moreFavorableSibling := s.getAdmittedMoreFavorableSibling(log, &e.Info, snapshot); moreFavorableSibling != nil {
+			log.V(3).Info("Skipping workload as a more favorable variant is already admitted", "moreFavorableVariant", klog.KObj(moreFavorableSibling.Obj))
```
pajakd, line 388: A general question: can there happen that there is one variant running and then multiple more favorable siblings get admitted in parallel?

getAdmittedMoreFavorableSibling / getAdmittedLessFavorableSibling assume that there is only a single one so I'm wondering if there is anything blocking that from happening.

PBundyra: The goal of this change is to make only one Variant running.

Before there was a short window between admission of Variant-A and deactivation of Variant-B when both of them could be running. However if we prevented admission in scheduler of Variant-B (before it gets deactivated) then we have guarantee only one Variant is running

pajakd: Ok, if Variant-A is running, it can happen that more favorable Variant-B gets admitted and start to replace Variant-A? Its possible, right? I was just wondering if at some point of this process an even more favorable Variant-C gets admitted -- will it cause some problems or will it be prevented at some point?

PBundyra: The hierarchy I had in mind is that Variant-A > Variant-B > Variant-C, with Variant-A as the most favorable Variant. 

To answer your question, yes more favorable Variant can evict less favorable one,  but during this eviction process more favorable Variant waits until the less favorable one finishes eviction admitting, so there's no risk of having two Variants running at the same time because of that

PBundyra:
/hold
Until https://github.com/kubernetes-sigs/kueue/pull/10954 is merged

pkg/scheduler/scheduler.go
PBundyra: Done

pkg/scheduler/scheduler.go
PBundyra: Done

PBundyra:
Made some refactor, let me know wdyt @pajakd @mimowo

pkg/scheduler/scheduler.go
```diff
@@ -1066,45 +1087,90 @@ func (s *Scheduler) updateEntryPenalty(log logr.Logger, e *entry, op usageOp) {
...
 }
 
-func (s *Scheduler) getLessFavorableSibling(wl *workload.Info, snap *schdcache.Snapshot) *workload.Info {
+// findAdmittedLessFavorableSibling returns an admitted Sibling Workload that has a
+// less favorable Resource Flavor assignment than the provided Workload.
+//
+// A Sibling is a Workload belonging to the same Parent Workload (part of the
+// Concurrent Admission feature). Favorability is determined by the order of
+// Resource Flavors defined in the ClusterQueue
+func (s *Scheduler) findAdmittedLessFavorableSibling(log logr.Logger, wl *workload.Info, snap *schdcache.Snapshot) (*workload.Info, error) {
```
pajakd: log <- unused variable?

pkg/scheduler/scheduler.go
```diff
@@ -1066,45 +1087,90 @@ func (s *Scheduler) updateEntryPenalty(log logr.Logger, e *entry, op usageOp) {
...
+		return siblingIdx > wlIdx // larger index means less favorable
+	})
+}
+
+// findAdmittedMoreFavorableSibling returns an admitted Sibling Workload that has a
+// more favorable Resource Flavor assignment than the provided Workload.
+//
+// A Sibling is a Workload belonging to the same Parent Workload (part of the
+// Concurrent Admission feature). Favorability is determined by the order of
+// Resource Flavors defined in the ClusterQueue
```
pajakd: This part of the comment is repeated above the `findAdmittedLessFavorableSibling` function. 

Perhaps we can just have it only once above`findAdmittedSibling`?

pkg/scheduler/scheduler.go
```diff
@@ -1066,45 +1087,90 @@ func (s *Scheduler) updateEntryPenalty(log logr.Logger, e *entry, op usageOp) {
...
+	return s.findAdmittedSibling(wl, snap, func(siblingIdx, wlIdx int) bool {
+		return wlIdx > siblingIdx
+	})
+}
+
+// findAdmittedSibling is a helper that finds an admitted sibling workload matching the comparison criteria.
+func (s *Scheduler) findAdmittedSibling(wl *workload.Info, snap *schdcache.Snapshot, compareFunc func(siblingIdx, wlIdx int) bool) (*workload.Info, error) {
 	if !features.Enabled(features.ConcurrentAdmission) {
-		return nil
+		return nil, nil
```
pajakd: (At least in the current implementation) we always check that this featuregate is on before calling findAdmittedMoreFavorableSibling/findAdmittedLessFavorableSibling so perhaps this is not needed. Is it left intentionally for safety?

PBundyra: Yes I left it intentionally in case in the future somebody called it without checking the fg

pkg/scheduler/scheduler.go
PBundyra: Fixed

pkg/scheduler/scheduler.go
PBundyra: Done

mimowo:
> /hold

@PBundyra why is that put on hold?

PBundyra:
/hold cancel
I put it on hold to slightly refactor it after out offline discussion. Now it's ready for review

pkg/scheduler/scheduler.go
```diff
@@ -1099,45 +1108,80 @@ func (s *Scheduler) updateEntryPenalty(log logr.Logger, e *entry, op usageOp) {
...
+		siblingFlavorIdx, err := getFlavorIndex(sibling, flavors)
+		if err != nil {
+			return false
+		}
+		return compareFunc(siblingFlavorIdx, wlFlavorIdx)
+	})
+}
 
+func (s *Scheduler) findAdmittedSiblingMatching(wl *workload.Info, cq *schdcache.ClusterQueueSnapshot, parentUID types.UID, match func(sibling *workload.Info) bool) *workload.Info {
 	for _, otherWl := range cq.Workloads {
```
mimowo, line 1159: This is pre-existing code, but it is a range over all Workloads in CQ, so I think we should optimize that, is this on the list of follow ups?

pkg/scheduler/scheduler.go
```diff
@@ -1099,45 +1108,80 @@ func (s *Scheduler) updateEntryPenalty(log logr.Logger, e *entry, op usageOp) {
 	}
 }
 
-func (s *Scheduler) getLessFavorableSibling(wl *workload.Info, snap *schdcache.Snapshot) *workload.Info {
+// findAdmittedLessFavorableSibling returns an admitted Sibling Workload that has a
+// less favorable Resource Flavor assignment than the provided Workload.
+func (s *Scheduler) findAdmittedLessFavorableSibling(wl *workload.Info, snap *schdcache.Snapshot) *workload.Info {
```
mimowo, line 1113: Something to consider, outside the PR - I'm wondering if we should move such helper functions to dedicated files, like scheduler_concurrent_admission.go, next to scheduler.go, just to offload the reading of the main scheduler.go. 

We have done that with tas_flavorassigner.go, but not everywhere, and I'm wondering if this is a good pattern or not. In any case, this is for follow up, just exploring what is your view here.

mimowo (commented):
/lgtm
/approve
Thank you 👍

PR #10931, opened 2026-05-04
Ensure featureGates passed by the user are validated

### LINKED ISSUE #10866: FeatureGates enabled with configMap or CLI are not being validated
**What happened**:

The validation of the FeatureGates happens before the values specified either with ConfigMap or CLI are actually set on the `DefaultMutableFeatureGate`. So, the validation is actually using the default values instead of the ones defined by the user.

```
$ kubectl get po kueue-controller-manager-856c45b9dd-qwdbq -n kueue-system -o yaml |grep -e DRAExtendedResources
    - --feature-gates=DRAExtendedResources=true
```

```
kubectl logs -f kueue-controller-manager-856c45b9dd-qwdbq -n kueue-system |grep -e DRAExtendedResources -e DynamicResourceAllocation
{"level":"Level(-2)","ts":"2026-04-30T15:17:38.011102194Z","logger":"setup","caller":"features/kube_features.go:591","msg":"Loaded feature gates","featureGates":{"AdmissionFairSharing":true,"AdmissionGatedBy":false,"AssignQueueLabelsForPods":true,"CustomMetricLabels":false,"DRAExtendedResources":true,"DynamicResourceAllocation":false,
```
According to this code block, the validation should have mentioned `DRAExtendedResources` when enabled should have `DynamicResourceAllocation` enabled as well.

https://github.com/kubernetes-sigs/kueue/blob/f26d35cb3d8a39967839909c66899c86399ce226/pkg/config/validation.go#L532-L536

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

MaysaMacedo:
#### What type of PR is this?
/kind bug

This commit ensures the validation of the feature gates happens
after the feature gates defined by the user are actually
enforced.

Fixes #10866

This PR was developed with the assistance of Claude.

```release-note
FeatureGates: Fixed a bug that user-specified feature gate parameters are not verified.
```

### DIFF
--- a/cmd/kueue/main.go
+++ b/cmd/kueue/main.go
@@ -34,7 +34,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
-	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	autoscaling "k8s.io/autoscaler/cluster-autoscaler/apis/provisioningrequest/autoscaling.x-k8s.io/v1"
 	"k8s.io/client-go/discovery"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
@@ -154,21 +153,9 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err := config.ValidateFeatureGates(featureGates, cfg.FeatureGates).ToAggregate(); err != nil {
+	if err := config.LoadAndValidateFeatureGates(featureGates, cfg.FeatureGates).ToAggregate(); err != nil {
 		setupLog.Error(err, "conflicting feature gates detected")
 		os.Exit(1)
-	}
-
-	if featureGates != "" {
-		if err := utilfeature.DefaultMutableFeatureGate.Set(featureGates); err != nil {
-			setupLog.Error(err, "Unable to set flag gates for known features")
-			os.Exit(1)
-		}
-	} else {
-		if err := utilfeature.DefaultMutableFeatureGate.SetFromMap(cfg.FeatureGates); err != nil {
-			setupLog.Error(err, "Unable to set flag gates for known features")
-			os.Exit(1)
-		}
 	}
 
 	// Validates the configuration after it has been loaded and feature gates have been set.
--- a/pkg/config/validation.go
+++ b/pkg/config/validation.go
@@ -34,6 +34,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	apimachineryutilvalidation "k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	utilfeature "k8s.io/apiserver/pkg/util/feature"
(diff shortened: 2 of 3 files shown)


test/e2e/sequential/baseline/feature_gates_validation_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+				rsList := &appsv1.ReplicaSetList{}
+				g.Expect(k8sClient.List(ctx, rsList, client.InNamespace(kueueNS))).To(gomega.Succeed())
+
+				var newestRS *appsv1.ReplicaSet
+				for i := range rsList.Items {
+					rs := &rsList.Items[i]
+					for _, ref := range rs.OwnerReferences {
+						if ref.UID == deployment.UID {
+							if newestRS == nil || rs.CreationTimestamp.After(newestRS.CreationTimestamp.Time) {
```
mimowo: The timestamp resolution in k8s in only 1s, so if two RS are created within the same seconds this may not be reliable and result in flakes (potentially). 

So, I'm wondering if we could use a more reliable way of choosing the latest replicaSet, maybe corresponding to the current version based on `deployment-kubernetes.io/revision`, wdyt?

mimowo (commented):
LGTM, just one nit to make sure the test works reliably

test/e2e/sequential/baseline/feature_gates_validation_test.go
MaysaMacedo: Good idea. Makes sense to me.

MaysaMacedo:
@mimowo This PR is ready for another look, whenever you can. Thanks

kannon92 (commented):
/lgtm

/assign @mimowo @gabesaba @tenzen-y

MaysaMacedo:
@mimowo this is ready for another look. Thank you!

test/e2e/sequential/baseline/feature_gates_validation_test.go
```diff
@@ -0,0 +1,90 @@
+/*
+Copyright The Kubernetes Authors.
```
mimowo: The e2e test here works, but it feels overkill really. I don't think we should be adding a new e2e test every such validation done in main.go, this will blow testing time very quickly.

So, I'm wondering if instead we could extract the code in main.go, for example: `loadConfigurationAndValidate` that would encapsulate the operations, and we then could move the testing to unit tests. 

The downside of e2e test is that if we also decide to drop `ElasticJobsViaWorkloadSlices` (say GA), we may need to think how to replace the testing.

mimowo:
Left the comment to try to extract the load and validate as a function in main go which would allow us more lightweight testing: https://github.com/kubernetes-sigs/kueue/pull/10931#discussion_r3239786729

mimowo:
Now this looks great, thank you for the bugfix, and making sure we have a lightweight testing strategy 👍 
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16

mimowo:
Please prepare manual CP in case the robot fails  on conflicts

tenzen-y:
/release-note-edit

```release-note
FeatureGates: Fixed a bug that user-specified feature gate parameters are not verified.
```

Issue #10861, opened 2026-04-30
Release v0.18.0


mimowo:
## Release Checklist

- [x] [OWNERS](https://github.com/kubernetes-sigs/kueue/blob/main/OWNERS) must LGTM the release proposal.
  At least two for minor or major releases. At least one for a patch release.
- [x] Verify that the changelog in this issue and the CHANGELOG folder is up-to-date
  - [x] Run `./hack/releasing/sync-notes.sh $VERSION` to generate and publish the release notes
- [x] For major or minor releases (v$MAJ.$MIN.0), create a new release branch.
  - [x] An OWNER creates a vanilla release branch with
        `git branch release-$MAJ.$MIN main`
  - [x] An OWNER pushes the new release branch with
        `git push upstream release-$MAJ.$MIN`
- [x] Update the release branch:
  - [x] Ensure there are no unstaged changes in your directory (the script adds everything)
  - [x] Run `./hack/releasing/prepare_pull.sh --target release $VERSION`
  - [x] Wait for this PR to merge #11818 
- [x] An OWNER creates a signed tag
  - [x] pull the release branch after PR from previous step merged
  - [x] run
     `git tag -s $VERSION`
      and inserts the changelog into the tag description.
      To perform this step, you need [a PGP key registered on github](https://docs.github.com/en/authentication/managing-commit-signature-verification/checking-for-existing-gpg-keys).
- [x] An OWNER pushes the tag with
      `git push upstream $VERSION`
  - Triggers prow to build and publish a staging container image
      `us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue:$VERSION`
- [x] An OWNER [prepares a draft release](https://github.com/kubernetes-sigs/kueue/releases)
  - [x] Create the draft release pointing out to the created tag.
  - [x] Write the change log into the draft release.
  - [x] Run
      `make release-artifacts IMAGE_REGISTRY=registry.k8s.io/kueue GIT_TAG=$VERSION`
      to generate the artifacts in the `release-artifacts` folder.
  - [x] Upload the files in the `release-artifacts` folder to the draft release - either
      via UI or `gh release --repo kubernetes-sigs/kueue upload $VERSION release-artifacts/*`.
- [x] Promote images and Helm Charts to production:
  - [x] Run `./hack/releasing/wait_for_images.sh $VERSION` to await for the staging images.
  - [x] Run `./hack/releasing/promote_pull.sh $VERSION` to submit the promotion PR
[...]

PR #10936, opened 2026-05-04
Update k8s 1 36

### LINKED ISSUE #11121: Update kueue to use 0.36 apis
Kueue should be updated to use 0.36 apis.

I have a PR up but there seems to be a few issues we need to resolve so filing a tracker issue.

kannon92:
#### What type of PR is this?
/kind cleanup

Update k8s dependencies to 0.36

Fixes #11121

Trainer has some compilation issues with 0.36 and I'm working on that in parallel.

```
# github.com/kubeflow/trainer/v2/pkg/apis/trainer/v1alpha1
vendor/github.com/kubeflow/trainer/v2/pkg/apis/trainer/v1alpha1/zz_generated.openapi.go:353:10: undefined: corev1.WorkloadReference
vendor/github.com/kubeflow/trainer/v2/pkg/apis/trainer/v1alpha1/zz_generated.openapi.go:13403:32: undefined: corev1.WorkloadReference
vendor/github.com/kubeflow/trainer/v2/pkg/apis/trainer/v1alpha1/zz_generated.openapi.go:13411: undefined: corev1.WorkloadReference
```

I opened up https://github.com/kubeflow/trainer/pull/3470 to hopefully resolve trainer failures.

We will need a release of trainer to fix this though. cc @tenzen-y @andreyvelich 

```release-note
NONE
```


tenzen-y:
> I opened up https://github.com/kubeflow/trainer/pull/3470 to hopefully resolve trainer failures.
> 
> We will need a release of trainer to fix this though. cc @tenzen-y @andreyvelich 

I left a comment for the Trainer your PR.

mbobrovskyi (commented):
Also please update Makefile-test.mk: https://github.com/kubernetes-sigs/kueue/blob/d72b32a94d24bda77ddf1cb5f4e7f5e627df12d7/Makefile-test.mk#L41.

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/json-iterator/go v1.1.12
 	github.com/kubeflow/mpi-operator v0.8.0
 	github.com/kubeflow/spark-operator/v2 v2.5.0
-	github.com/kubeflow/trainer/v2 v2.2.0
+	github.com/kubeflow/trainer/v2 v2.0.0-20260512164132-81803e638df3
```
kannon92: @andreyvelich @tenzen-y
I had to use main for trainer to fix some build issues when updating to Kubernetes 1.36.

@mimowo @tenzen-y I'm not sure if this needs to block kueue from merging but we will need a minor release of trainer with the 1.36 bump if we want to use release tags.

/hold

kannon92: https://github.com/kubeflow/trainer/issues/3513

kannon92:
> Also please update Makefile-test.mk:
> 
> https://github.com/kubernetes-sigs/kueue/blob/d72b32a94d24bda77ddf1cb5f4e7f5e627df12d7/Makefile-test.mk#L41
> 
> .

https://github.com/kubernetes-sigs/kueue/pull/10936/changes/a8e46400665839f565d87234c9af60827d6e23cc

updated.

Makefile-test.mk
```diff
@@ -38,8 +38,8 @@ INTEGRATION_API_LOG_LEVEL ?= 0
 
 # Folder where the e2e tests are located.
 E2E_TARGET ?= ./test/e2e/...
-E2E_K8S_VERSIONS ?= 1.33.7 1.34.3 1.35.0
-E2E_K8S_VERSION ?= 1.35
+E2E_K8S_VERSIONS ?= 1.34.7 1.35.4 1.36.0
```
kannon92:
```suggestion
E2E_K8S_VERSIONS ?= 1.34.7 1.35.0 1.36.0
```

Makefile-test.mk
```diff
@@ -38,8 +38,8 @@ INTEGRATION_API_LOG_LEVEL ?= 0
 
 # Folder where the e2e tests are located.
 E2E_TARGET ?= ./test/e2e/...
-E2E_K8S_VERSIONS ?= 1.33.7 1.34.3 1.35.0
-E2E_K8S_VERSION ?= 1.35
+E2E_K8S_VERSIONS ?= 1.34.7 1.35.0 1.36.0
```
kannon92:
```suggestion
E2E_K8S_VERSIONS ?= 1.34.3 1.35.0 1.36.0
```

go.mod
mimowo: cc @kshalot

kannon92: We followed off online so going to resolve.

kshalot is aware of this for the scheduler-library work.

kannon92:
> **Note:** This review was generated with the assistance of AI (Claude).

## K8s 1.36 Bump Review (non-vendor changes only)

### Summary
The non-vendor changes are clean. No bugs introduced. The changes fall into three categories: (1) a new functional change (PreStart hook), (2) test fixes to accommodate upstream API changes, and (3) generated code / version bumps.

### PreStart Hook — the only real logic change

**Files:** `pkg/controller/jobframework/interface.go`, `reconciler.go`, `pkg/controller/jobs/job/job_controller.go`

K8s 1.36 requires the `JobSuspended` status condition to be set before allowing pod template mutations on suspended jobs. The implementation looks correct:

- **No stale-object conflict**: `PreStart` does `c.Status().Update()` which updates the in-memory object's `ResourceVersion`. Since `job.Object()` returns a pointer to the same underlying struct as `object`, the subsequent `clientutil.Patch` in `startJob` will use the updated ResourceVersion.
- **Idempotent**: Checks if the condition already exists before adding it. Safe for retries.
- **Early-return guard**: Correctly skips if the job isn't suspended.
- **Placement**: Runs before the MultiKueue skip-local-execution check — harmless but unnecessary for MultiKueue jobs that won't execute locally. Not a bug, just a minor inefficiency.
- **Only batch/v1 Job**: Only the `Job` type implements `JobWithPreStartHook`. Other job types (RayJob, MPIJob, etc.) don't get this hook, which is correct since the `JobSuspended` condition is specific to batch/v1.

### Test fixes — all appropriate

- **`base_webhook_test.go`**: Ignoring `BadValue` on `field.Error` — upstream changed how `BadValue` is populated. Tests still validate error type, field path, and message.
- **`delete_workload_test.go`**: Ignoring `ResourceVersion` on `ListMeta` — the fake client now populates this field.
- **`websocket_handler_test.go`**: Simplified `DoneChecker` mock to match a changed interface.
- **`job_controller_test.go`**: All start/unsuspend test cases correctly updated to expect the new `JobSuspended` condition. `LastTransitionTime` correctly ignored in comparisons.

### Generated / mechanical changes — all expected

- **CRDs** (charts + config): Upstream schema changes (schedulingGroup replaces workloadRef, ProcMountType/UserNamespacesSupport feature gate notes removed, ShardedListAndWatch additions).
- **OpenAPI**: New `ShardInfo` type and `shardSelector` field.
- **Informers** (`client-go/informers/`): Generated code from the new informer codegen.
- **`doc.go` files**: `+k8s:defaulter-gen=TypeMeta` — required by the new code generator.
- **`.golangci.yaml`**: Suppressing deprecation warnings for `GetEventRecorderFor` and `scheme.Builder` — pragmatic for a dep bump, but these should eventually be migrated.

### Version matrix

Dropping K8s 1.33 and adding 1.36 for E2E tests. Supported range is now **1.34, 1.35, 1.36** (N-2 skew policy). Correct.

### No issues found

hack/tools/go.mod
```diff
@@ -499,17 +499,17 @@ require (
...
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	helm.sh/helm/v3 v3.20.2 // indirect
 	honnef.co/go/tools v0.7.0 // indirect
 	howett.net/plist v1.0.1 // indirect
 	k8s.io/api v0.35.5 // indirect
-	k8s.io/apiextensions-apiserver v0.35.5 // indirect
-	k8s.io/apimachinery v0.35.5 // indirect
-	k8s.io/apiserver v0.35.5 // indirect
+	k8s.io/apiextensions-apiserver v0.35.3 // indirect
```
mbobrovskyi: Hmm. Why we downgrade?

pkg/controller/jobs/job/job_controller.go
```diff
@@ -162,6 +163,29 @@ func (j *Job) Suspend() {
...
+	}
+	for _, cond := range j.Status.Conditions {
+		if cond.Type == batchv1.JobSuspended && cond.Status == corev1.ConditionTrue {
+			return nil
+		}
+	}
+	j.Status.Conditions = append(j.Status.Conditions, batchv1.JobCondition{
+		Type:               batchv1.JobSuspended,
+		Status:             corev1.ConditionTrue,
+		LastTransitionTime: metav1.Now(),
```
mbobrovskyi: Please use clock.

pkg/controller/jobframework/interface.go
```diff
@@ -169,6 +169,14 @@ type JobWithManagedBy interface {
 	SetManagedBy(*string)
 }
 
+// JobWithPreStartHook interface should be implemented by generic jobs
```
mbobrovskyi:
```suggestion
// JobWithPreStartHook is an optional interface should be implemented by generic jobs
```

pkg/controller/jobs/job/job_controller.go
```diff
@@ -162,6 +163,29 @@ func (j *Job) Suspend() {
 	j.Spec.Suspend = new(true)
 }
 
+// PreStart ensures the JobSuspended status condition is set before the job is
+// unsuspended. K8s 1.36+ requires this condition to allow pod template mutations
+// on suspended jobs. In envtest the K8s Job controller doesn't run, and in
+// production there can be a race, so Kueue sets it defensively.
```
mbobrovskyi: Actually, maybe we should wait for this condition instead of set manually?

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/json-iterator/go v1.1.12
 	github.com/kubeflow/mpi-operator v0.8.0
 	github.com/kubeflow/spark-operator/v2 v2.5.0
-	github.com/kubeflow/trainer/v2 v2.2.0
+	github.com/kubeflow/trainer/v2 v2.2.1-0.20260515134202-448c432a7073
```
tenzen-y, line 13: The discussion point is here.

cc @mimowo @sohankunkerkar @kannon92

mimowo: I think it would be great to use github trainer  v2.2.1, but I wouldn't hold the upgrade to k8s 1.36 on this. 

The reason is that the integration is still early days, and I think getting the core kueue to the latest k8s dependencies is far more important.

mimowo: cc @andreyvelich who is considering the 2.2.1 release, but again, I wouldn't block this PR on that one, certainly.

hack/tools/go.mod
kannon92: Not really sure here.

Makefile-test.mk
```diff
@@ -43,8 +43,8 @@ INTEGRATION_API_LOG_LEVEL ?= 0
 
 # Folder where the e2e tests are located.
 E2E_TARGET ?= ./test/e2e/...
-E2E_K8S_VERSIONS ?= 1.33.7 1.34.3 1.35.0
-E2E_K8S_VERSION ?= 1.35
+E2E_K8S_VERSIONS ?= 1.34.3 1.35.0 1.36.0
```
mbobrovskyi:
```suggestion
E2E_K8S_VERSIONS ?= 1.34.8 1.35.5 1.36.1
```

mimowo: Is it something we could bump in a small preparatory pr?

Asking cause this feels independent of the upgrade

mbobrovskyi: We can bump it in follow-up.

go.mod
tenzen-y: I discussed with @mimowo offline.
Then we agreed to include the Trainer release branch commit in the next Kueue 0.18.

pkg/controller/jobs/job/job_controller.go
mimowo: Actually, do we really need this PreStart? Are there any tests failing without it?

Asking because if the tests are passing I would leave it as a follow up, the PR is already massive, and I don't think the currently implemented approach is correct. Kueue should not be setting the Job's conditions really.

kannon92: if I don't include this the job integration tests fail.

mbobrovskyi: Can we add this condition manually in integration tests?

tenzen-y:
@kannon92 When you rebase this PR, please care about https://github.com/kubernetes-sigs/kueue/pull/11481#discussion_r3289640732

cmd/kueueviz/backend/handlers/websocket_handler_test.go
```diff
@@ -36,29 +36,18 @@ import (
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+type alwaysDone struct{}
+
+func (alwaysDone) Name() string          { return "mock" }
+func (alwaysDone) Done() <-chan struct{} { ch := make(chan struct{}); close(ch); return ch }
```
mbobrovskyi: Not sure about this. We are creating new channel for each execution.

mbobrovskyi: Can we create a variable with already closed channel? Maybe like this?

```
var closedChan = func() <-chan struct{} {
	ch := make(chan struct{})
	close(ch)
	return ch
}()
```

hack/tools/go.mod
mbobrovskyi: Can we bump to v0.36.x?

pkg/controller/jobs/job/job_controller.go
mimowo: Ok, I synced with @kannon92 on slack. The reason for the changes is that k8s actually introduced a breaking change in 1.36 introducing this change to Beta: https://github.com/kubernetes/enhancements/issues/5440

Because previously Job's PodTemplate could be modified before job is started, and not later. After the enhancement it can be modified whenever the Job is suspeded, i.e. JobSuspended=True.

However, this is a breaking change, because a just created job, without JobSuspenced=True not cannot be mutated in 1.36. This hits in the Kueue integration tests because we are not running the Job_controller there to set JobSuspended=True condition for the newly created Jobs. At the same time we are running api-server in integration tests, and so we are running validation.

THis is not a production issue for Kueue, detectable in e2e tests, because in e2e scenario the Job controller sets the condition and so the PodTemplate can be mutated.

So, I think the action items here:
1. open issue in the core k8s to fix the new enhancement so that we don't have the regression
2. revert the PreStart changes in the Kueue production code - they are not needed
3. plumb integration tests by simulating the Job controller behavior and set JobSuspended=True from the integration tests

I will do (1.) and open issue in the core.

hack/tools/go.mod
```diff
@@ -499,17 +499,17 @@ require (
...
+	k8s.io/apimachinery v0.36.0 // indirect
+	k8s.io/apiserver v0.35.3 // indirect
```
mbobrovskyi: Also here and other places.

pkg/controller/jobs/job/job_controller.go
mimowo: I synced with @kannon92 who gave me write access to the branch - I will work together with Kevin to prepare the changes .

Makefile-test.mk
mimowo: I think we can bump that as a prep actually, testing here: https://github.com/kubernetes-sigs/kueue/pull/11542

pkg/controller/jobframework/interface.go
mimowo: Dropped the function as per https://github.com/kubernetes-sigs/kueue/pull/10936#discussion_r3289665769

pkg/controller/jobs/job/job_controller.go
mimowo: Dropped the function as per Dropped the function as per https://github.com/kubernetes-sigs/kueue/pull/10936#discussion_r3289665769

hack/tools/go.mod
mimowo: These changes are reverted in the latest diff

mimowo:
/unhold
unholding the hold from https://github.com/kubernetes-sigs/kueue/pull/10936#discussion_r3230467743
As we discussed in this thread https://github.com/kubernetes-sigs/kueue/pull/10936#discussion_r3285653329 the consensus is that using the commit for kubeflow trainer is not a blocker for the upgrade of k8s dependencies.

cmd/kueueviz/backend/handlers/websocket_handler_test.go
mimowo: Done in the latest push

go.mod
andreyvelich: Since this PR is almost ready, I think we can release Trainer v2.2.1 next week: https://github.com/kubeflow/trainer/pull/3536
@mimowo @tenzen-y What is timeline for Kueue 0.18?
cc @Krishna-kg732

kannon92: https://github.com/kubernetes-sigs/kueue/issues/10861#issuecomment-4500678084

Hopefully we release on May 28th.

mimowo:
FYI: the main issues with the upgrade are due to the regression in the core k8s: https://github.com/kubernetes/kubernetes/issues/139281

This specifically makes currently MultiKueue not working with Kubernetes 1.36: https://github.com/kubernetes-sigs/kueue/pull/11576

I'm currently finalizing the workaround inside Kueue by the "miniJobController" which is needed to update the status of the JobSuspended condition following the spec.suspend field.

mimowo:
/test pull-kueue-test-integration-multikueue-main
the failure does not look related, retrying, I'm not sure: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10936/pull-kueue-test-integration-multikueue-main/2059718671579746304

sohankunkerkar:
@mimowo as discussed, I tested your proposed fix here: https://github.com/kubernetes-sigs/kueue/pull/11699/changes/491dcd208bd206313e50b581f92766896efd79d9 but that alone wasn't sufficient to avoid the OOM since k8s 1.36 envtest kube-apiserver uses roughly 100MB more RSS per instance, and with 3 procs running 9 API servers the baseline exceeds what the 10Gi Prow limit can handle. I also reduced `INTEGRATION_NPROCS_MULTIKUEUE` from 3 to 2 and with both changes together tests passed.

mimowo:
i see in that case I would try bumping the mwmory available for the suite from 9Gi to 12Gi. We could do that as prep task. wdyt?

sohankunkerkar:
> i see in that case I would try bumping the mwmory available for the suite from 9Gi to 12Gi. We could do that as prep task. wdyt?

That's even better! I thought most of the jobs were taking around 9Gi, so I didn't recommend that since I felt this might be an outlier.

sohankunkerkar:
@mimowo PTAL https://github.com/kubernetes/test-infra/pull/37119

mimowo:
It makes sense. Actually maybe even more than 12 Gi. I never fully realized how heavy the suite is -9 API servers running. I think this is the reasons the tests were also failing at 4 parallel suite runs we tried initially. Because using 3 parallel suites wasn't our go to approach, we just didn't understand back then why we need to reduce 4 to 3.

Having said that let's stay wit 3 parallel suites to avoid running into flakes, at least not in this task.

sohankunkerkar:
> It makes sense. Actually maybe even more than 12 Gi. I never fully realized how heavy the suite is -9 API servers running. I think this is the reasons the tests were also failing at 4 parallel suite runs we tried initially. Because using 3 parallel suites wasn't our go to approach, we just didn't understand back then why we need to reduce 4 to 3

Yup, makes sense. I can bump it to 14Gi across all three configs to account for future k8s version growth and leave headroom for potential 4-proc runs.

mimowo:
I'm good with 14Gi to leave that space, it can be a follow up pr once we confirm this is helping to stabilize the test

sohankunkerkar:
Alright, the test-infra PR has merged. Let me check [here](https://github.com/kubernetes-sigs/kueue/pull/11699) to see if that resolves the issue.

sohankunkerkar:
> Alright, the test-infra PR has merged. Let me check [here](https://github.com/kubernetes-sigs/kueue/pull/11699) to see if that resolves the issue.

Yup, it worked: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/11699/pull-kueue-test-integration-multikueue-main/2059848276890357760

mimowo:
/test all
stability test

kannon92:
@mimowo 
thanks for taking this up!

mimowo:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10936/pull-kueue-test-integration-multikueue-main/2060013791545921536
investigating

mimowo:
Oh I know, the bug is in the assert we use `gomega.Expect(createdJob.Spec.Suspend).To(gomega.Equal(new(false)))` instead of `g.Expect(createdJob.Spec.Suspend).To(gomega.Equal(new(false)))`. In that case the assert is not waiting 45s, but fails immediately.  Which is also consistent with the fact the test didn't report any waiting time:
```
MultiKueue Suite: [It] MultiKueue Should run a job on worker if admitted (ManagedBy) [area:multikueue, feature:multikueue] expand_less	5s
{ failed [FAILED] Expected
    <*bool | 0xc00cc00a0f>: true
to equal
    <*bool | 0xc00cc00a2f>: false
In [It] at: /home/prow/go/src/sigs.k8s.io/kueue/test/integration/multikueue/jobs_test.go:355 @ 05/28/26 15:09:49.992
}
```
I will submit a separate PR as the test bug was always there, but the longer "no sync" window makes it now exposed.

mimowo:
Oh, this is new code added, I will update https://github.com/kubernetes-sigs/kueue/pull/11730

EDIT: updated. Cherrypicked the fix on top.

mimowo:
@olekzabl @sohankunkerkar @kannon92 ptal. It would be nice to include it today so that we have some buffer of testing before tomorrow's release.

sohankunkerkar:
I will add a label once CI is green

sohankunkerkar (approved):
/lgtm
Thanks for handling this!

tenzen-y (commented):
I confirmed that the primary change is a workaround for https://github.com/kubernetes/kubernetes/issues/139281

Thanks 👍 

/lgtm
/approve

mimowo:
Right, opened the issue https://github.com/kubernetes-sigs/kueue/issues/11756 to track the removal of the hack.

olekzabl:
#11863 tracks another hack cleanup once https://github.com/kubernetes/kubernetes/issues/139281 is fixed.

PR #10937, opened 2026-05-04
Move AllAtOnce MultiKueue dispatcher to a dedicated controller

### LINKED ISSUE #6803: Move AllAtOnce MultiKueue dispatcher to a dedicated controller

To move the code of the AllAtOnce MK dispatcher to a dedicated controller, similarly as done for the incremental dispatcher in https://github.com/kubernetes-sigs/kueue/pull/6601

**Why is this needed**:

To decouple responsibility of MultiKueue dispatcher, and the core MultiKueue controllers.

This is a follow up suggested in https://github.com/kubernetes-sigs/kueue/issues/6179#issuecomment-3283450927

andrewseif:
#### What type of PR is this?
/kind cleanup
/area multikueue

Move AllAtOnce MultiKueue dispatcher to a dedicated controller 

Fixes #6803

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added an external all-at-once workload dispatcher controller for MultiKueue “AllAtOnce” mode.
  * Introduced the `MultiKueueAllAtOnceExternal` feature gate (Beta, default enabled).
  * Dispatcher selection now supports incremental and all-at-once modes.
* **Bug Fixes**
  * Preserves evicted remote workloads during remote synchronization to allow eviction recovery.
  * Maintains legacy all-at-once nomination behavior when the external dispatcher feature is disabled.
* **Tests**
  * Expanded reconciliation, worker nomination, feature-gate, and integration setup coverage.


pkg/controller/workloaddispatcher/allatoncedispatcher_test.go
```diff
@@ -0,0 +1,245 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
```
mszadkow, line 5: filename type

pkg/controller/workloaddispatcher/allatonedispatcher.go
```diff
@@ -0,0 +1,131 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
```
mszadkow: filename typo

mimowo (commented):
Thank you for the effort 👍

pkg/controller/workloaddispatcher/allatonedispatcher.go
andrewseif: nice catch, I would have never seen that missing "c" 😃

andrewseif:
I had to add some logic to watch the config and clusters, as these were free in the inline version, but the logic itself is mostly similar

mimowo:
cc @Singularity23x0 @olekzabl @kshalot ptal

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -788,6 +775,18 @@ func (w *wlReconciler) nominateAndSynchronizeWorkers(ctx context.Context, group
...
+			// propagate the remote job's termination back to the manager Job.
+			// Deleting it here breaks the eviction-recovery flow because
+			// bestMatchByCondition(WorkloadEvicted) becomes nil, SyncJob is no
+			// longer called, and the manager Job's Status.Active is never updated.
+			// The remote will be cleaned up later, once the local workload is
+			// re-admitted (or finishes / loses its quota reservation).
+			if workload.IsEvicted(remoteWl) {
+				log.V(3).Info("Preserving evicted remote workload to allow eviction-recovery sync", "remote", rem)
+				continue
+			}
```
mimowo: Interesting does it mean we have a bug for the Increamental dispatcher which is already extracted? If this is the case maybe we could start by showing the bug, and fixing it here. It will also make the PRs more dijestable by the split.

pkg/controller/workloaddispatcher/incrementaldispatcher.go
```diff
@@ -123,6 +123,16 @@ func (r *IncrementalDispatcherReconciler) Reconcile(ctx context.Context, req ctr
...
 
+	// The workload is being evicted; let the core eviction flow complete (the Job
+	// reconciler will UnsetQuotaReservation once the job is no longer active, and
+	// the scheduler will requeue it) before re-nominating clusters. Re-nominating
+	// during eviction races with the post-eviction cleanup and prevents the
+	// workload from re-entering the queue.
+	if workload.IsEvicted(wl) {
+		log.V(3).Info("Workload is being evicted, skip the reconciliation")
+		return reconcile.Result{}, nil
+	}
```
mimowo: Whys we need the change in the incremental dispatcher? If it requires a bugfix, let's decouple that change as a preparatory bugfix.

mimowo (commented):
cc @mszadkow who worked before on extrating / adding he incremental dispatcher. Ptal

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -757,21 +756,9 @@ func (w *wlReconciler) nominateAndSynchronizeWorkers(ctx context.Context, group
...
-		}
-		if group.local.Status.ClusterName == nil && !equality.Semantic.DeepEqual(group.local.Status.NominatedClusterNames, nominatedWorkers) {
-			if err := workload.PatchAdmissionStatus(ctx, w.client, group.local, w.clock, func(wl *kueue.Workload) (bool, error) {
-				wl.Status.NominatedClusterNames = nominatedWorkers
-				return true, nil
-			}); err != nil {
-				log.V(2).Error(err, "Failed to patch nominated clusters", "workload", klog.KObj(group.local))
-				return reconcile.Result{}, err
-			}
-		}
```
mimowo, line 917: The AllAtOnce dispatcher is default, and critical currently. Could we add a safety FG like MultiKueueAllAtOnceExternal, which would be Beta in 0.18, and we would make it stable in 0.19. 
That would guard the removal of the code here, and registration of the controller, wdyt?

andrewseif: I agree, I believe a FG would be beneficial here.

pkg/controller/admissionchecks/multikueue/workload.go
andrewseif: I wouldn't say it was a bug, but there was a race window that was apparent, when I extracted the AllAtOnce MK dispatcher, ill make sure to post the findings for verification in wg-batch for more experienced eyes to verify my findings

pkg/controller/workloaddispatcher/incrementaldispatcher.go
andrewseif: It was to avoid the race window, and to also mirror the guard in AllAtOnce Dispatcher, so they are both similar.

mimowo: What are the implications of the race? Are there any user observable?

andrewseif: I don't think they were user observable, but now that I am preparing the write up, I don't see any logs that indicate incremental dispatcher was the problem, I think I added this since, I was missing this guard in my dispatcher and the original function had it (and it made the tests pass, which means original functionality is unaffected), so added to both dispatchers, thinking it might be the problem.

I'll revert the change and see if the tests fails, and that should prove if there's a bug in this specific dispatcher or not.

andrewseif: I have ran the tests and I don't think we need to put this guard here anymore, it passed.

we can keep it if we want symmetry between dispatchers (since the original method had it, but I don't think its needed)

mimowo:
@andrewseif I 'm planning to conclude the review next week, overall it looks great, but another pair of eyes from MK experts (@olekzabl or @kshalot ) would be great. Thank you for the effort once again 👍

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -757,7 +757,12 @@ func (w *wlReconciler) nominateAndSynchronizeWorkers(ctx context.Context, group
 	// supporting preferred or required placement constraints.
 	if clusterName := workload.ClusterName(group.local); group.IsElasticWorkload() && clusterName != "" {
 		nominatedWorkers = []string{clusterName}
-	} else if w.dispatcherName == config.MultiKueueDispatcherModeAllAtOnce {
+	} else if !features.Enabled(features.MultiKueueAllAtOnceExternal) && w.dispatcherName == config.MultiKueueDispatcherModeAllAtOnce {
+		// Legacy inline AllAtOnce nomination path. Kept under feature gate so
+		// operators can roll back to the pre-#6803 behavior if the dedicated
```
olekzabl: Nit: for me it feels more natural to use the # of the fixing PR (= this one), not of the issue.
```suggestion
		// operators can roll back to the pre-#10937 behavior if the dedicated
```

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -771,7 +776,9 @@ func (w *wlReconciler) nominateAndSynchronizeWorkers(ctx context.Context, group
 			}
 		}
 	} else {
-		// Incremental dispatcher and External dispatcher path
+		// Dispatcher (AllAtOnce when MultiKueueAllAtOnceExternal=true,
+		// Incremental, or External) is responsible for populating
```
olekzabl: Another nit, feel free to ignore
```suggestion
		// A dispatcher placed outside this file (AllAtOnce when MultiKueueAllAtOnceExternal=true,
		// Incremental, or External) is responsible for populating
```

pkg/controller/admissionchecks/multikueue/workload.go
olekzabl: I have trouble following the issue description in the comment here.

Mostly because, in the handling of Evicted condition which you mentioned (I suppose, [here](https://github.com/kubernetes-sigs/kueue/blob/698e60c9e09a3bd587edb9a713592a650e4c81ff/pkg/controller/admissionchecks/multikueue/workload.go#L372-L407)), the only explicit call to `SyncJob` is [here](https://github.com/kubernetes-sigs/kueue/blob/698e60c9e09a3bd587edb9a713592a650e4c81ff/pkg/controller/admissionchecks/multikueue/workload.go#L383), i.e. in the case of a *manager-originating* eviction, while your case seems to be the other one (the *worker-originating* eviction, dealt with [here](https://github.com/kubernetes-sigs/kueue/blob/698e60c9e09a3bd587edb9a713592a650e4c81ff/pkg/controller/admissionchecks/multikueue/workload.go#L392-L405)), given that you care whether the manager will notice.

This might be not yet contradictory; maybe you've found a longer path (a ping-pong across a few reconcilers?) leading to calling `SyncJob` also in the worker-originating case?

But anyway, my bottom line is: 
* +1 to documenting this as a separate issue #N
* and then, instead of summarizing that issue in a comment here, I'd just leave a link to #N, because:
  * even a several-lines summary can be hard to follow (as I'm right now experiencing)
  * an even longer summary does not feel fit in this place
  * #N will act as a place where we can further discuss (while such comments are more "frozen")

pkg/controller/workloaddispatcher/allatoncedispatcher.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+
+	mkAc, err := admissioncheck.GetMultiKueueAdmissionCheck(ctx, r.client, wl)
+	if err != nil {
+		log.Error(err, "Can not get MultiKueue AdmissionCheckState")
+		return reconcile.Result{}, err
+	}
+
+	if mkAc == nil || mkAc.State != kueue.CheckStatePending {
+		log.V(3).Info("AdmissionCheckState is not in Pending, skip the reconciliation")
```
olekzabl:
```suggestion
		log.V(3).Info("MultiKueue AdmissionCheckState is not Pending, skip the reconciliation")
```

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -788,6 +795,18 @@ func (w *wlReconciler) nominateAndSynchronizeWorkers(ctx context.Context, group
 				}
 			}
 		} else if remoteWl != nil {
+			// Preserve a remote workload that is currently Evicted: reconcileGroup's
+			// needs it (with WorkloadEvicted=True) to call SyncJob and
```
olekzabl:
```suggestion
			// Preserve a remote workload that is currently Evicted: reconcileGroup
			// needs it (with WorkloadEvicted=True) to call SyncJob and
```

pkg/controller/workloaddispatcher/allatoncedispatcher.go
```diff
@@ -0,0 +1,317 @@
...
+		return reconcile.Result{}, nil
+	}
+
+	// The workload is already assigned to a cluster, no need to nominate workers.
```
olekzabl: Nit: this comment feels not very useful, given that it's duplicated in the log text just below.
(Though I'm aware that it looks so also in `incrementaldispatcher.go`).

pkg/controller/workloaddispatcher/allatoncedispatcher.go
```diff
@@ -0,0 +1,317 @@
...
+			}
+			// Missing cluster: skip.
+			continue
+		}
+		if apimeta.IsStatusConditionTrue(cluster.Status.Conditions, kueue.MultiKueueClusterActive) {
+			active.Insert(clusterName)
+		}
+	}
+	return active, nil
+}
```
olekzabl, line 161: This whole `filterActiveClusters` seems to be mostly an optimization that you've added "by the way"?

Looking at the original logic, IIUC, the set of nominated workers was based just on `group.remotes` ([here](https://github.com/kubernetes-sigs/kueue/blob/698e60c9e09a3bd587edb9a713592a650e4c81ff/pkg/controller/admissionchecks/multikueue/workload.go#L761)), that in turn based on [this call](https://github.com/kubernetes-sigs/kueue/blob/698e60c9e09a3bd587edb9a713592a650e4c81ff/pkg/controller/admissionchecks/multikueue/workload.go#L298). Digging deeper, I didn't see anything like checking `MultiKueueClusterActive`. I guess it makes sense, true, but I'd vote for separating optimizations from refactors (I mean, into separate PRs).

For this refactoring PR, I'd consider ways to inject `wlReconciler` into this reconciler and just call its `remoteClientsForAc` method. (There are precedents, e.g. `wlReconciler` knows the `clustersReconciler`, [here](https://github.com/kubernetes-sigs/kueue/blob/698e60c9e09a3bd587edb9a713592a650e4c81ff/pkg/controller/admissionchecks/multikueue/workload.go#L69)). Not necessarily strictly this way, but sth like this, to reduce duplication of code.

Then, in a follow-up PR, you're welcome to add this smarter filtering. But maybe not only here? Maybe the other dispatchers could also benefit from that?

WDYT?

pkg/controller/workloaddispatcher/allatoncedispatcher.go
```diff
@@ -0,0 +1,317 @@
...
+	return &AllAtOnceDispatcherReconciler{
+		client:      c,
+		helper:      helper,
+		clock:       realClock,
+		roleTracker: roleTracker,
+	}
+}
+
+func (r *AllAtOnceDispatcherReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := ctrl.LoggerFrom(ctx)
```
olekzabl, line 82: AFAICS the first 40 lines of this method are _almost_ identical as in `incrementaldispatcher.go`.
This raises 2 questions:

1. Can this be unified?
   The differences seem not very blocking - they're only about `r.clearRoundStartTime` in the incr dispatcher; this could be passed as a callback "what should the common code do with a troubling error".

2. The first real difference is your newly-added special handling of eviction.
   Though then - maybe it'd make sense to add it to the incremental dispatcher as well?
   (Hence, again, I'd prefer to deal with it in a dedicated issue, and a dedicated fixing PR, separate from refactoring).

pkg/controller/workloaddispatcher/allatoncedispatcher.go
```diff
@@ -0,0 +1,317 @@
...
+		if apimeta.IsStatusConditionTrue(cluster.Status.Conditions, kueue.MultiKueueClusterActive) {
+			active.Insert(clusterName)
+		}
+	}
+	return active, nil
+}
+
+func (r *AllAtOnceDispatcherReconciler) nominateWorkers(ctx context.Context, wl *kueue.Workload, remoteClusters sets.Set[string], log logr.Logger) (reconcile.Result, error) {
+	nominatedWorkers := remoteClusters.UnsortedList()
+	slices.Sort(nominatedWorkers)
```
olekzabl: Would this work?

```suggestion
	nominatedWorkers := sets.List(remoteClusters)
```

olekzabl (commented):
First of all, thank you for doing this!

Then, even before going into detailed comments, I'm feeling I should start from a "provoking" question:
The incremental dispatcher is currently being parametrized with step size (#10877).
Given that, what if we "implemented" the "externalized" AllAtOnce just as a special case of that? 
`standaloneAllAtOnce := incrementalDispatcher{stepSize: 1000000}`

You could say that's too simplistic, wasting performance etc. Because incremental dispatcher contains some bits of logic that we don't need. Perhaps.

But even if so, I'd like to ask how much of them we can have in common. Maybe extracting some shared pieces. Or maybe a central shared entry point with injectable per-case callbacks. IDK yet.

I'm just intuitively afraid of nearly-duplicating ~200 lines of code which may then diverge without a good reason. (And, in my eyes, this "unjustified divergence" shows up already in this PR. See my detailed comments).

I haven't yet read everything but must pause now. Will come back later.

pkg/controller/workloaddispatcher/allatoncedispatcher.go
andrewseif: This was not intended as smart filtering.

it was added because the integration test test/integration/multikueue/setup_test.go:L752 was failing on the refactor branch. That test ("Should properly detect insecure kubeconfig of MultiKueueClusters and remove remote client") explicitly asserts that an inactive cluster does not appear in Status.NominatedClusterNames. 

The old code passed this test structurally because nominations came from group.remotes (in-process map maintained by clustersReconciler), which never contains a disconnected cluster. The new dispatcher pulls from admissioncheck.GetRemoteClusters() which returns configured clusters regardless of activity, so without the filter that test fails.

pkg/controller/admissionchecks/multikueue/workload.go
andrewseif: I have sent you my investigation findings, maybe you can verify them, I believe moving the AllAtOnce dispatcher created this, and it should be part of the PR, as without it the program won't function properly.

mszadkow:
@andrewseif can you share the failing scenario? I guess it was not caught by any of the existing one, right?

mimowo:
@andrewseif just checking where are with the PR - do you need some review / another pair of eyes? I'm wondering if this PR is still feasible for 0.18

andrewseif:
@mimowo I sent my investigation to both @olekzabl and @mszadkow, but I haven't gotten any review/feedback from them yet.

And I am not sure what would be missing here, if any. 
the current branch is just pending rebase, and that's it.

olekzabl:
I apologize @andrewseif , I must declare bankruptcy on this PR, at least until Wednesday.
I haven't managed to look at your investigation yet.

> And I am not sure what would be missing here, if any.
the current branch is just pending rebase, and that's it.

Well... there still are some comments from me which you haven't responded to?

pkg/controller/admissionchecks/multikueue/workload.go
mimowo: I'm wondering if this is related to this problem which is also attempted to be fixed here:  https://github.com/kubernetes-sigs/kueue/pull/11378

cc @mszadkow wdyt?

mimowo:
/test pull-kueue-priority-booster-test-integration-main
/test pull-kueue-verify-main
Checking if these are flakes, but pull-kueue-verify-main seems like permanent

andrewseif:
[pull-kueue-verify-main] is a small linter fix I need to deploy, for the open comments, I think the investigation answers most of them, except the architectural one, specifically this

> AFAICS the first 40 lines of this method are almost identical as in incrementaldispatcher.go.
> This raises 2 questions:
> 
> Can this be unified?
> The differences seem not very blocking - they're only about r.clearRoundStartTime in the incr dispatcher; this could be passed as a callback "what should the common code do with a troubling error".
> 
> The first real difference is your newly-added special handling of eviction.
> Though then - maybe it'd make sense to add it to the incremental dispatcher as well?
> (Hence, again, I'd prefer to deal with it in a dedicated issue, and a dedicated fixing PR, separate from refactoring).

I think this might need to be discussed in our wg-batch meeting.

I can answer from a software design perspective, but I am not sure I can answer from a kueue architecture direction, which I think @olekzabl is referring to

pkg/controller/admissionchecks/multikueue/workload.go
andrewseif: I have looked at what @mszadkow did, it's in the same test suite, I think they are both targeting the same race condition, and it was showing in two different tests. 

edit: I think it should solve #11115 aswell?

olekzabl:
> I can answer from a software design perspective, but I am not sure I can answer from a kueue architecture direction, which I think @olekzabl is referring to

My intent basically is to reduce the divergences between AllAtOnce and Incremental, because such divergences - especially if not clear at the first glance - feel like a risk of having some issues on one of the sides.
(See [this comment](https://github.com/kubernetes-sigs/kueue/pull/10937#pullrequestreview-4255743785). 
BTW I still think it could be valuable to start writing "external AllAtOnce" by taking "Incremental with N = 1000000" at least as a starting point. Or, if it won't work because Incremental has its own issues, it'll be great to know that).

I consider it as a software engineering healthy practice, rather than "Kueue architecture".

pkg/controller/admissionchecks/multikueue/workload.go
mimowo: >  I wouldn't say it was a bug, but there was a race window that was apparent, when I extracted the AllAtOnce MK dispatcher, ill make sure to post the findings for verification in wg-batch for more experienced eyes to verify my findings

@andrewseif actually most of "races" are bugs, so I would like to understand which test and how is failing. It very well might be that you have discovered a bug we should extract to a separate preparatory bugfix PR we should cherrypick. 

In order to let us understand what is the race I would recommend that you temporarily revert (or comment out the code) so that we can see what is the failure, and analyze it. Then we can make an informed decision if this is a separate bugfix or part of this PR.

mimowo:
> > I can answer from a software design perspective, but I am not sure I can answer from a kueue architecture direction, which I think @olekzabl is referring to
> 
> My intent basically is to reduce the divergences between AllAtOnce and Incremental, because such divergences - especially if not clear at the first glance - feel like a risk of having some issues on one of the sides. (See [this comment](https://github.com/kubernetes-sigs/kueue/pull/10937#pullrequestreview-4255743785). BTW I still think it could be valuable to start writing "external AllAtOnce" by taking "Incremental with N = 1000000" at least as a starting point. Or, if it won't work because Incremental has its own issues, it'll be great to know that).
> 
> I consider it as a software engineering healthy practice, rather than "Kueue architecture".

I totally agree there is a lot of duplication between the two dispatchers, and we should commonize the code. Additionally, the new dispatcher seems to do a much better job by tracking changes to MultiKueueConfig, which the previous didn't do - and I think this is a bug in the incrementaldispatcher which just didn't notice. We could get the bug for incremental dispatcher fixed for free if we commonized the code.

On "how to" commonize the code I'm wondering architecturally what is better - and I'm not sure:
1. single "GeneticDispatcher" with two modes AllAtOnce and Incremental (single Reconciler)
2. two dispatchers which call a commonized function like GenericReconcile (two Reconcilers)

I'm leaning towards (1.) as this avoids duplication for the event handlers. I thought an argument for (2.) could be load separation, so similarly as @olekzabl suggested, but I just wouldn't say "Incremental with N = 1000000", but GenericDispatcher which supports both modes.

Let me know @andrewseif if this makes sense.

pkg/controller/workloaddispatcher/allatoncedispatcher.go
mszadkow: That's correct, it only proves how embedded AllAtOnce was, good catch @andrewseif

mszadkow:
@andrewseif I read through your investigation, it appears to be correct.
I don't think it's a bug necessarily, but rather the effect of how embedded the AllAtOnce is into the MK Workload Reconciler.

andrewseif:
I understand the idea about commonizing the code, and personally I agree with that approac, specifically this:

> single "GenericDispatcher" with two modes AllAtOnce and Incremental (single Reconciler)

From a software engineering/architecture perspective, it reduces a lot of the common code (which is mainly guards and checks).

However that leaves us with 2 realistic options:

1. Merge this PR, and create another one to refactor both dispatchers into one (ill take that issue and work on it)
2. Close this PR, and the referenced ticket and open a new one where we refactor both dispatchers at once.

There is a third option, which I wouldn't recommend since it will not be related to the issue in this PR, which is just refactor this PR to merge the 2 dispatchers into GenericDispatcher.

I am more leaning towards option (1) as it is more incremental in the changes, with smaller blast radius.

What do you guys think?

For the GenericDispatcher I think we can say it defaults to AllAtOnce dispatcher if any of the following match:
1. Step N = 0
2. Step N = 1000000
3. Step == nil

pkg/controller/admissionchecks/multikueue/workload.go
andrewseif: @mimowo I can tell you for sure, there was a race condition, and in my investigation I removed the piece of code that was preventing the bug, then reviewed the logs and I could see it, I can share my findings with you on slack(I didn't earlier because I know you are busy) or here if u want.

> Then we can make an informed decision if this is a separate bugfix or part of this PR.

The fix needs to be part of the PR.

mimowo:
Thank you for the summary 👍 

> From a software engineering/architecture perspective, it reduces a lot of the common code (which is mainly guards and checks).

Yes, but the guards can also be commonized by calling functions, similarly as we do here: https://github.com/kubernetes-sigs/kueue/pull/11598 so I think the option (2.) is on the table as well, especially as an interim step.

> However that leaves us with 2 realistic options:
> 
> 1. Merge this PR, and create another one to refactor both dispatchers into one (ill take that issue and work on it)
> 2. Close this PR, and the referenced ticket and open a new one where we refactor both dispatchers at once.

I think this makes sense to split the task into these two steps. At the first step I would just try to align both implementations by calling common functions, similarly as in the previously referenced PR, wdyt @olekzabl ?

andrewseif:
I honestly took a look at this PR #11208 , and I don't think we really need this PR anymore, since I think we don't need to move AllAtOnce at all, but remove it and have it a subset of the incremental (if stepSize = 100000+ (or so), then just dispatch to all clusters at once).

So I don't know why would we need this PR at all, other than just removing AllAtOnce.

mimowo:
Even with that PR we need to maintain the AllAtOnce as UX, because:
1. UX convenience option for OOTB no-config option
2. it expresses the user intention much better than Increamental with step 100000
3. backwards compatibility, while possible for Beta features (like MK), deprecating (and dropping) the default AllAtOnce dispatcher will be months of time and deprecation notices, community discussions etc

Having said that we have totally freedom to drop AllAtOnce and implement it using the Incremental dispatcher.

andrewseif:
Thank you for the explanation, Ill follow your lead on whatever decision you guys make 😃

PR #10940, opened 2026-05-05
fix: inject TAS workload annotation and podset label for LWS pods in webhook

islewis:
EDIT: prematurely opened, closing


islewis:
prematurely opened, closing

PR #10948, opened 2026-05-05
tas: improve LWS balanced placement

ShaanveerS:
#### What type of PR is this?
/kind cleanup
/area tas

Improves TAS Balanced Placement for LeaderWorkerSet by avoiding leader reservation on each node.
Domains that cannot meet the threshold while hosting the leader now lose only leader capacity and stay usable, instead of being pruned from worker placement.

For example with `threshold=5` and `leaderRequired=true`:
before pruning:
```
# [state, sliceState, stateWithLeader, sliceStateWithLeader, leaderState]
leader-domain      [6, 6, 5, 5, 1]
worker-only-domain [5, 5, 4, 4, 1]
  ```
after pruning (old):
  ```
leader-domain      [6, 6, 5, 5, 1]
worker-only-domain [0, 0, 0, 0, 0]  ← pruned, even though sliceState=5 meets threshold
parent-domain      [6, 6, 5, 5, 1]
  ```
after pruning (new):
  ```
leader-domain      [6, 6, 5, 5, 1]
worker-only-domain [5, 5, 0, 0, 0]  ← kept for workers with leader capacity cleared 
parent-domain      [11, 11, 10, 10, 1]
  ```
Part of #7494

```release-note
Improved TAS balanced placement for LeaderWorkerSet: domains that meet the worker threshold stay usable even when they can't host the leader.
```

### DIFF
--- a/pkg/cache/scheduler/tas_balanced_placement.go
+++ b/pkg/cache/scheduler/tas_balanced_placement.go
@@ -69,12 +69,7 @@ func balanceThresholdValue(sliceCount int32, selectedDomainsCount int32, lastDom
 		threshold = min(threshold, lastDomainWithLeader.sliceStateWithLeader)
 	}
 	if lastDomain != nil {
-		// TODO: https://github.com/kubernetes-sigs/kueue/issues/7494
-		// we don't have min(threshold, lastDomain.sliceState) here because
-		// later we prune all nodes with sliceStateWithLeader < threshold
-		// while pruning we don't know which node will host the leader
-		// so we have to leave space on each node
-		threshold = min(threshold, lastDomain.sliceStateWithLeader)
+		threshold = min(threshold, lastDomain.sliceState)
 	}
 	return threshold
 }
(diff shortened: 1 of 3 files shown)


mimowo (commented):
cc @pajakd @mwysokin

pkg/cache/scheduler/tas_balanced_placement.go
```diff
@@ -318,18 +317,52 @@ func clearState(d *domain) {
...
+	clone := *d
+	clone.parent = parent
+	clone.children = make([]*domain, len(d.children))
+	for i, child := range d.children {
+		clone.children[i] = cloneDomain(child, &clone)
+	}
+	return &clone
+}
+
+func pruneDomainBelowThreshold(d *domain, threshold int32, leaderRequired bool) {
```
pajakd: Lets give this function a different name so that its not confused with the one below.

This function is pruning a single node, while the one below is going over a list of domains.

ShaanveerS: changed to `pruneDomainNodeBelowThreshold`. Thanks!

mwielgus (commented):
It might be hard to understand what this change really does. Could you please provide more information? Something like state before/ state after would be great.

pkg/cache/scheduler/tas_balanced_placement.go
```diff
@@ -69,12 +69,7 @@ func balanceThresholdValue(sliceCount int32, selectedDomainsCount int32, lastDom
 		threshold = min(threshold, lastDomainWithLeader.sliceStateWithLeader)
 	}
 	if lastDomain != nil {
-		// TODO: https://github.com/kubernetes-sigs/kueue/issues/7494
-		// we don't have min(threshold, lastDomain.sliceState) here because
-		// later we prune all nodes with sliceStateWithLeader < threshold
-		// while pruning we don't know which node will host the leader
-		// so we have to leave space on each node
-		threshold = min(threshold, lastDomain.sliceStateWithLeader)
+		threshold = min(threshold, lastDomain.sliceState)
```
pajakd, line 72: the PR is looking good

But we need some more testing. Currently it only has unit tests for the pruning but since its changing a some important evaluation I would like to see some (integration?) tests for the LWS with balanced placement.

pkg/cache/scheduler/tas_balanced_placement.go
```diff
@@ -257,19 +252,23 @@ func findBestDomainsForBalancedPlacement(s *TASFlavorSnapshot, params *topologyA
...
 		if !fits {
 			continue
 		}
 		threshold := balanceThresholdValue(sliceCount, selectedDomainsCount, lastDomainWithLeader, lastDomain)
 		if threshold >= bestThreshold {
-			s.pruneDomainsBelowThreshold(requestedLevelSiblingDomains, threshold, params.sliceSize, params.sliceLevelIdx, params.requestedLevelIdx, params.leaderCount > 0)
-			_, requestedLevelDomainCount, _, _ := evaluateGreedyAssignment(s, requestedLevelSiblingDomains, sliceCount, params.leaderCount)
+			s.pruneDomainsBelowThreshold(candidateDomains, threshold, params.sliceSize, params.sliceLevelIdx, params.requestedLevelIdx, params.leaderCount > 0)
+			fitsAfterPruning, requestedLevelDomainCount, _, _ := evaluateGreedyAssignment(s, candidateDomains, sliceCount, params.leaderCount)
+			if !fitsAfterPruning {
```
pajakd, line 279: Please add some testcase that enters the "!fitsAfterPruning" code branch. Also, I thought, that if we enter this branch -- meaning that we don't fit after pruning, we should also consider the lower value of the threshold (like calculated with the old code). Isn't it needed?

pkg/cache/scheduler/tas_balanced_placement.go
ShaanveerS: Done. added a cache-level test.

pkg/cache/scheduler/tas_balanced_placement.go
ShaanveerS: sorry, missed that. added the fallback and tests.

ShaanveerS:
> It might be hard to understand what this change really does. Could you please provide more information? Something like state before/ state after would be great.

Sure! For example with `threshold=5` and `leaderRequired=true`:
before pruning:
```
# [state, sliceState, stateWithLeader, sliceStateWithLeader, leaderState]
leader-domain      [6, 6, 5, 5, 1]
worker-only-domain [5, 5, 4, 4, 1]
  ```
after pruning (old):
  ```
leader-domain      [6, 6, 5, 5, 1]
worker-only-domain [0, 0, 0, 0, 0]  ← pruned, even though sliceState=5 meets threshold
parent-domain      [6, 6, 5, 5, 1]
  ```
after pruning (new):
  ```
leader-domain      [6, 6, 5, 5, 1]
worker-only-domain [5, 5, 0, 0, 0]  ← kept for workers with leader capacity cleared 
parent-domain      [11, 11, 10, 10, 1]
  ```

File: pkg/cache/scheduler/tas_balanced_placement_test.go [hunk 4]
```diff
@@ -170,3 +170,110 @@ func TestPlaceSlicesOnDomainsBalanced(t *testing.T) {
...
+		"parentDomain":     {11, 11, 10, 10, 1},
+		"workerOnlyDomain": {5, 5, 0, 0, 0},
+		"workerOnlyLeaf":   {5, 5, 0, 0, 0},
+	}
+	if diff := cmp.Diff(want, got); diff != "" {
+		t.Errorf("Unexpected domain state (-want,+got):\n%s", diff)
+	}
+}
+
+func TestFindBestDomainsForBalancedPlacementFallsBackAfterPruning(t *testing.T) {
```
pajakd: Lets refactor these testcases to align with the rest of the codebase. Define `TestFindBestDomainsForBalancedPlacement` in which we define testcases and then run them in a loop (just check other files in our repo). It'll be easier to add more tests.

EDIT: same for `pruneDomainsBelowThreshold`

ShaanveerS: done!

mimowo:
@ShaanveerS @pajakd should we categorize the issue as a bugfix or feature? The release note is there because there is a user-observable impact, but the PR categorization is cleanup. So, I suspect this should be categorized as bugfix? But the user impact remains unclear to me, on the surface - I would need to check the code and tests. Please think if we could formulate the release-note to indicate the user-observable change.

ShaanveerS:
i am leaning feature rather than bug. the old behavior was conservative, but not necessarily a correctness bug imo.
maybe something like:
```
TAS Balanced Placement for LeaderWorkerSet can now handle placements where only the leader domain has extra capacity and the other selected domains have just enough capacity for workers.
``` 
wdyt?

mimowo (commented):
Ack, thank you 👍 
/lgtm
/approve

PR #10949, opened 2026-05-05
[TAS] fix: Prevent negative state in topology assignment

tg123:
#### What type of PR is this?

/kind bug
/area tas

Adds regression test coverage for a TAS scheduling bug where `TopologyDomainAssignment.Count` could be emitted with a negative value, causing the apiserver to reject the Workload admission update with a CRD validation error (`podCounts.individual[*]` requires `>= 1`).

The bug occurs in over-subscribed leaf domains: when `tasUsage` (or `tasUsage + assumedUsage`) exceeds `freeCapacity`, `CountInWithLimitingResource` returns a negative `int32`. That value was previously stored verbatim on `leaf.state` and propagated through `fillInCountsHelper` (`childrenCapacity += child.state`) into the emitted `TopologyAssignment`.

This PR adds two regression tests in `pkg/cache/scheduler/tas_flavor_snapshot_test.go`:

1. `TestBuildTopologyAssignmentForLevels_DropsNonPositiveStateDomains` — asserts the output-side filter in `buildTopologyAssignmentForLevels` skips any domain whose `state <= 0`, so a non-positive `Count` is never emitted.
2. `TestFillInCounts_ClampsNegativeRemainingCapacity` — asserts the source-side fix: `fillInCounts` must clamp `leaf.state` to `0` when remaining capacity goes negative, covering both the `tasUsage` over-subscription case and the `assumedUsage` (preempt-mode) case.

Both tests currently fail on `main`, demonstrating the bug. The fix will land in a follow-up PR.

Fixes #

_Guessing reason_: during preempt-mode admission while the lower-priority workload is still being evicted

Reproduced in production with the following scheduler error (names redacted):

```
Could not admit Workload and assign flavors in apiserver
  workload:     <workload>/<namespace>
  clusterQueue: <clusterQueue>
  parentCohort: <cohort>
  error: Workload.kueue.x-k8s.io "<workload>" is invalid:
    [status.admission.podSetAssignments[0].topologyAssignment
       .slices[0].podCounts.individual[168]:
       Invalid value: -1: ... should be greater than or equal to 1,
     status.admission.podSetAssignments[0].topologyAssignment
       .slices[0].podCounts.individual[190]:
       Invalid value: -1: ... should be greater than or equal to 1]
Workload re-queued
  requeueReason: FailedAfterNomination
```

The negative `-1` count values in `podCounts.individual[168]` and `podCounts.individual[190]` are exactly the symptom these tests guard against.

```release-note
NONE
```


pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1475,8 +1475,9 @@ func (s *TASFlavorSnapshot) buildTopologyAssignmentForLevels(domains []*domain,
 	}
 	assignment.Levels = s.levelKeys[levelIdx:]
 	for _, domain := range domains {
-		if domain.state == 0 {
-			// It may happen when PodSet count is 0 or when using LeastFreeCapacity algorithm.
+		if domain.state <= 0 {
+			// == 0: It may happen when PodSet count is 0 or when using LeastFreeCapacity algorithm.
+			// <  0: Defensive clamp: negative state would cause apiserver CRD validation to reject
```
mimowo: I'm ok with the defensive approach here, but we need to log that, probably as V(0) as this should really not happen. Do you see that happening on a real cluster?

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1626,6 +1627,10 @@ func (s *TASFlavorSnapshot) fillInCounts(requirements *topologyAssignmentPodRequ
 		}
 		var limitingRes corev1.ResourceName
 		leaf.state, limitingRes = requirements.requests.CountInWithLimitingResource(remainingCapacity)
+		// apiserver CRD validation silently rejects (Individual >= 1).
+		if leaf.state < 0 {
+			leaf.state = 0
```
mimowo: Here too, I'm ok-ish with the defensive approach, but we need to log this high. Also, I'm curious if you see this at real clusters

mimowo (commented):
/ok-to-test

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1475,8 +1475,9 @@ func (s *TASFlavorSnapshot) buildTopologyAssignmentForLevels(domains []*domain,
 	}
 	assignment.Levels = s.levelKeys[levelIdx:]
 	for _, domain := range domains {
-		if domain.state == 0 {
-			// It may happen when PodSet count is 0 or when using LeastFreeCapacity algorithm.
+		if domain.state <= 0 {
```
mwielgus: This is a corrective step. Do you know HOW Kueue gets into this state? Can the problem be intercepted/logged earlier?

tg123:
@mimowo @mwielgus i do not fully understand the issue too, but i will dig into full logs again

some info 
our cluster has 100+ jobs and 300+ GB200 racks. 
running 1.16.6 with a customized plugin, should not be related to plugin because it just manages the pods lifecycle

this happened twice and likely due to pre-check cordon the nodes and some internal state messed up
i have tried restarting kueue to clean up tas cache, but it did not help

mimowo:
> @mimowo @mwielgus i do not fully understand the issue too, but i will dig into full logs again

I see, in that case maybe instead of making the corrective actions let's start with improved logging to provide details that will allow us to underestand the issue when that repeats.

> our cluster has 100+ jobs and 300+ GB200 racks.
> running 1.16.6 with a customized plugin, should not be related to plugin because it just manages the pods lifecycle

Let me ask some questions:
- what is the version of Kueue you are using?
- do you have all ResourceFlavors defined on disjoint sets of nodes, or there might be some overlap (a node matching two TAS flavors)?

tg123:
what is the version of Kueue you are using?
sorry typo, 0.16.6


do you have all ResourceFlavors defined on disjoint sets of nodes, or there might be some overlap (a node matching two TAS flavors)?
no, very simple rf 
```
kind: ResourceFlavor
metadata:
  ..............
spec:
  nodeLabels:
    ..................
  tolerations:
  - effect: NoSchedule
    key: nvidia.com/gpu
    operator: Exists
  topologyName: rack-topology
```

```
kind: Topology
metadata:
  ......
spec:
  levels:
  - nodeLabel: nvidia.com/gpu.clique
  - nodeLabel: kubernetes.io/hostname
  ```



also some interesting findings in log: 
kueue logs errCodeAssumptionsViolated "unexpected remainingCount" at line 1462 40 ms before the apiserver rejection.

in updateCountsToMinimumGeneric pkg/cache/scheduler/tas_flavor_snapshot.go


```
 01:00:45.103  creates Workload (208 pods × 4 GPU = 832 GPU)
 01:00:45.160  scheduler cycle 14889 starts
 01:00:45.204  Snapshot taken (43.9 ms)
 01:00:45.207   tas_flavor_snapshot.go:1462 ERROR errCodeAssumptionsViolated
               "unexpected remainingCount" remainingCount=2 sliceSize=1
               leaves={..qr..qs..qt..qu..qv..qw..qx..qy..qz..r0..r1..r2..r3..r4..r5..r6..r7..}
 01:00:45.208  Workload assumed in cache
 01:00:45.247  scheduler.go:695 ERROR podCounts.individual[96]:-1, [168]:-1, [190]:-1
 01:00:45.247  Workload re-queued requeueReason=FailedAfterNomination
 01:00:45.250  cycle 14890 — same outcome … repeats every 50 ms for 3.5 hours
 ```

mimowo:
Thank you for the quick response:
1. do your workloads have multiple Podsets or just one?
2. what is the CRD you are using for the integration?
3. what is the mode you are using "required" / "preferred" / "unconstrained"?
4. Are you able to share the anonimized workload object when that happens?

tg123:
1 do your workloads have multiple Podsets or just one?
just one podset

2 what is the CRD you are using for the integration?
the CRD defines:
 * Count -> map to jobset.Count
 * Pod Template -> map to jobset.Template
 * PodSetSliceSize -> map into  jobset.Template (template.Annotations[kueue.PodSetSliceSizeAnnotation])

3 when kwl get admitted, kueue  jobframework.ComposableJob.Run will signal our customized operator to create pod by writing to CRD

4 what is the mode you are using "required" / "preferred" / "unconstrained"?
required

5 Are you able to share the anonimized workload object when that happens?
hmmm, already patched our cluster with this defensive approach. i did not save the bug kwl

mimowo:
I see, one more question, do you maybe have a lot of small Pods on a node, or rather few big. I'm wondering that maybe `CountInWithLimitingResource` returns negative due to exceeding the cap on the number of Pods (110 iirc by default) - this part is relatively weakly tested.

tg123:
no, our jobs are big and always 1 pod per node

mimowo:
I see, ok I'm not sure what is the root cause, but:
1. I think the change in [here](https://github.com/kubernetes-sigs/kueue/pull/10949/changes#diff-015ac0ccc6a319e922161cdd677487236e5d4e7edc11d0376fd59aebf14ae148R1478-R1480) would not be necessary if the leaf.state is non-negative, so we correct the other place
2. I think instead of correcting outside of CountInWithLimitingResource, this function should never produce negative results - the responsibilty of this function is to give "count" of pods which can schedule. Negative count does not make sense.
3. I think it is technically possible (not sure if this is what happened in your env) that the CountInWithLimitingResource produce negative results in some cases when the Node resources "allocatable" minus pods is negative - that could happen for example:
a) if a Node driver reduced the Node allocatable - then for a short while the node could be running over-booked. 
b) if we have a non-TAS Pod scheduled by kube-scheduler and accounted before the TAS workload is fully evicted and releases the quota. For example, a TAS workload spans 1000 Nodes, it will release quota and domains only ones all Pods terminate, but this creates a time windows when a non-TAS Pod may schedule on the Node1, and so the Node1 will be running overbooked - and so CountInWithLimitingResource will produce currently negative results.

Again, I'm not sure if this is what happened for you, but this is possible in such scenarios, so I would propose:
1. make the correction inside CountInWithLimitingResource
2. add logging at V(2) around here if you observe the remainingCapacity is negative: https://github.com/kubernetes-sigs/kueue/pull/10949/changes#diff-015ac0ccc6a319e922161cdd677487236e5d4e7edc11d0376fd59aebf14ae148R1628-R1629 for any resource
3. drop the corrective actions outside of calling CountInWithLimitingResource

wdyt?

tg123:
update PR to clamp in CountInWithLimitingResource only
this approach is more clean

totally agree we'd better to understand `why` instead of just patch it

i did noticed Terminating pods and weird nvidia dev plugin state: reporting 0 gpu count
let me dig more and update

mimowo:
> i did noticed Terminating pods and weird nvidia dev plugin state: reporting 0 gpu count
let me dig more and update

Nice this can suggest malfunction of the gpu driver, and so an issue like (3a) from https://github.com/kubernetes-sigs/kueue/pull/10949#issuecomment-4387896844.

tg123:
thanks @mimowo you are right about 3a

found the root cause, Allocatable was set to 0 by dev plugin.

```
 gpuAlloc=0, gpuCap=4, Ready=True, unsched=false. 
``` 


The most extreme example is <nodeA> (one specific GB200 node): 8 consecutive snapshot samples over ~2 hours (2026-05-04 21:41–23:47 UTC) with gpuAlloc=0, gpuCap=4, Ready=True, unsched=false. 

During that window kueue's TAS accounting still sees 1 pods * 4 GPU = 4 GPU "in use" on a node whose *Allocatable* is 0, so remainingCapacity for that leaf is 0 - 4 =-4. With rValue = 4 per pod, cap/rValue = -1

that is where -1 came from

related kubelet log

```
Kernel + kubelet logs from <nodeA> around the trigger:

 21:34:38.933  kernel    NVRM: Xid (PCI:0008:01:00): 145, ...
 21:34:38.933  kernel    NVRM: Xid (PCI:0008:01:00): 45, pid=…, channel 0x… caused by previous Xid 145
 21:34:39.xxx  kernel    (storm of Xid 45 across 30+ channels on the same GPU)
 ...
 21:35:05.181  kubelet   setters.go:393 "Updated allocatable" device="nvidia.com/gpu" allocatable=0
 21:35:15.310  kubelet   setters.go:393 "Updated allocatable" device="nvidia.com/gpu" allocatable=0
 21:35:25.362  kubelet   setters.go:393 "Updated allocatable" device="nvidia.com/gpu" allocatable=0
```

tg123:
follow up question for us is that why the node kept in this state for long, our node system should detect it and kick it out
however, that is not kueue related.

the only thing i think kueue should take care of is `allocatable=0 `

pkg/resources/requests.go
```diff
@@ -193,7 +193,12 @@ func (r Requests) CountInWithLimitingResource(capacity Requests) (int32, corev1.
 		if rValue == 0 {
 			count = int32(math.MaxInt32)
 		} else {
-			count = int32(cap / rValue)
+			// Clamp to 0: when capacity has been over-subscribed (e.g. by
```
mimowo: Let's also add an integration test in test/integration/singlecluster/tas/tas_test.go which would fail if not for the fix. The test should simulate the node driver failing and setting gpu / or custom resource to 0 for allocatable while the Pods are running.

mimowo:
SGTM, I also propose to add an integration test here: https://github.com/kubernetes-sigs/kueue/pull/10949#discussion_r3199123272. Let me know if there are any complications with that.

tg123:
@mimowo
e2e test added
and you can repro by just picking the tas_test.go to main: `apiserver rejects podCounts.individual[0]: Invalid value: -4` without `Clamp to 0`

test/integration/singlecluster/tas/tas_test.go
```diff
@@ -3655,6 +3655,173 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
...
+						utiltas.V1Beta2From(&utiltas.TopologyAssignment{
+							Levels: []string{corev1.LabelHostname},
+							Domains: []utiltas.TopologyDomainAssignment{
+								{Count: 4, Values: []string{"x2"}},
+							},
+						}),
+					))
+				})
+
+				ginkgo.By("verifying no Workload's TopologyAssignment carries a negative Count", func() {
```
mimowo: Well, that would not happen anyway due to the validation rule that requires the count to be >0, right?
So IIUC , the previous assert is enough to discriminate the fixed / broken versions, right?

tg123: correct, let me clean it up, 
this wont happen if wl2 is admitted

test/integration/singlecluster/tas/tas_test.go
```diff
@@ -3655,6 +3655,151 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
...
+
+				ginkgo.By("creating running pods so cache observes the GPU usage on x1", func() {
+					createPodsForWorkload(wl1, ns.Name, false, true)
+				})
+
+				ginkgo.By("simulating a device-plugin failure: drop x1 nvidia.com/gpu allocatable to 0", func() {
+					nodeToUpdate := &corev1.Node{}
+					gomega.Expect(k8sClient.Get(ctx, apitypes.NamespacedName{Name: "x1"}, nodeToUpdate)).Should(gomega.Succeed())
+					nodeToUpdate.Status.Allocatable["nvidia.com/gpu"] = resource.MustParse("0")
+					gomega.Expect(k8sClient.Status().Update(ctx, nodeToUpdate)).Should(gomega.Succeed())
```
mimowo: wrap than in Eventually block to make sure the update is retried if it fails on the first attempt

mimowo: Actually, in integration tests there is no kubelet, so I don't expect any other controller would modify that concurrently

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16
Thank you 👍

mimowo:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10949/pull-kueue-test-e2e-sequential-baseline-main/2052452824444309504
looks like flake, opened: https://github.com/kubernetes-sigs/kueue/issues/11029

mimowo:
@tg123 please prepare the CPs manually in case the robot fails due to conflicts

/cherrypick release-0.17
/cherrypick release-0.16

PR #10951, opened 2026-05-05
[ConcurrentAdmission] Emit events on Variant creation/activation/deactivation

PBundyra:
…tivation

#### What type of PR is this?
/kind bug

ConcurrentAdmission controller emits events on Variant creation/activation/deactivation

Fixes #

```release-note
NONE
```


pkg/controller/concurrentadmission/controller.go
```diff
@@ -246,7 +250,7 @@ func (r *variantReconciler) createVariants(ctx context.Context, parent *kueue.Wo
 			return err
 		}
 		log.V(3).Info("Variant created", "variant", klog.KObj(variant), "flavor", flavor)
-		// TODO: Emit event here
+		r.recorder.Eventf(parent, corev1.EventTypeNormal, ReasonCreatedVariant, "Variant Workload %s created", variant.Name)
```
mbobrovskyi: Can we add namespace?

mbobrovskyi: Also, maybe we should mention the parent?

pkg/controller/concurrentadmission/controller.go
```diff
@@ -454,28 +459,36 @@ func (r *variantReconciler) activateVariants(ctx context.Context, parent *kueue.
...
+func (r *variantReconciler) deactivateWl(ctx context.Context, wl *kueue.Workload, message string) error {
 	if wl == nil || !workload.IsActive(wl) {
 		return nil
 	}
 	wl.Spec.Active = new(false)
-	return c.Update(ctx, wl)
+	if err := r.client.Update(ctx, wl); err != nil {
+		return err
+	}
+	r.recorder.Eventf(wl, corev1.EventTypeNormal, ReasonDeactivatedVariant, "Deactivated Workload: %s", message)
```
mbobrovskyi:
```suggestion
	r.recorder.Eventf(wl, corev1.EventTypeNormal, ReasonDeactivatedVariant, "Workload %s/%s is deactivated due to %s", wl.Namespace, wl.Name, reason)
```

pkg/controller/concurrentadmission/controller.go
```diff
@@ -365,7 +369,7 @@ func (r *variantReconciler) deactivateVariants(
 		log.V(2).Info("Parent is not active, deactivating all variants", "parent", klog.KObj(parent))
 		for i := range variants {
 			v := &variants[i]
-			if err := r.deactivateVariant(ctx, v); err != nil {
+			if err := r.deactivateVariant(ctx, v, fmt.Sprintf("Parent Workload: %s/%s not active", parent.Namespace, parent.Name)); err != nil {
```
mbobrovskyi:
```suggestion
			if err := r.deactivateVariant(ctx, v, fmt.Sprintf("parent Workload %s/%s is not active", parent.Namespace, parent.Name)); err != nil {
```

pkg/controller/concurrentadmission/controller.go
```diff
@@ -392,7 +396,7 @@ func (r *variantReconciler) deactivateVariants(
 			if flavorOrder[concurrentadmission.GetVariantFlavor(v)] > flavorOrder[*minPreferredFlavor] {
 				log.V(2).
 					Info("Deactivating variant because it is below the minPreferredFlavor", "variant", klog.KObj(v), "flavor", concurrentadmission.GetVariantFlavor(v), "minPreferredFlavor", *minPreferredFlavor)
-				if err := r.deactivateVariant(ctx, v); err != nil {
+				if err := r.deactivateVariant(ctx, v, fmt.Sprintf("Variant %s/%s below minPreferredFlavor: %s", v.Namespace, v.Name, *minPreferredFlavor)); err != nil {
```
mbobrovskyi:
```suggestion
				if err := r.deactivateVariant(ctx, v, fmt.Sprintf("variant %s/%s below minPreferredFlavor: %s", v.Namespace, v.Name, *minPreferredFlavor)); err != nil {
```

pkg/controller/concurrentadmission/controller.go
```diff
@@ -454,28 +459,36 @@ func (r *variantReconciler) activateVariants(ctx context.Context, parent *kueue.
...
+func (r *variantReconciler) activateWl(ctx context.Context, wl *kueue.Workload, message string) error {
 	if wl == nil || workload.IsActive(wl) {
 		return nil
 	}
 	wl.Spec.Active = new(true)
-	return c.Update(ctx, wl)
+	if err := r.client.Update(ctx, wl); err != nil {
+		return err
+	}
+	r.recorder.Eventf(wl, corev1.EventTypeNormal, ReasonActivatedVariant, "Activated Workload: %s", message)
```
mbobrovskyi: Also here the same.

PBundyra: Added

pkg/controller/concurrentadmission/controller.go
PBundyra: Done

pkg/controller/concurrentadmission/controller.go
PBundyra: Good point about a namespace. I skipped the parent name as it's already present in the event object

pkg/controller/concurrentadmission/controller.go
PBundyra: I think it's better to keep the camel case to be consitent with other places in code and documentation

pkg/controller/concurrentadmission/controller.go
PBundyra: I think it's better to keep the camel case to be consitent with other places in code and documentation

pkg/controller/concurrentadmission/controller.go
PBundyra: After second consideration I didn't include Workload name here to not duplicate information already present in the Event

pkg/controller/concurrentadmission/controller.go
PBundyra: Deleted the colon

pkg/controller/concurrentadmission/controller.go
```diff
@@ -246,7 +250,7 @@ func (r *variantReconciler) createVariants(ctx context.Context, parent *kueue.Wo
 			return err
 		}
 		log.V(3).Info("Variant created", "variant", klog.KObj(variant), "flavor", flavor)
-		// TODO: Emit event here
+		r.recorder.Eventf(parent, corev1.EventTypeNormal, ReasonCreatedVariant, "Variant Workload %s created", klog.KObj(variant))
```
mbobrovskyi: Hmm. Can we use klog.KObj() in all cases for consistency?

mbobrovskyi:
```suggestion
		r.recorder.Eventf(parent, corev1.EventTypeNormal, ReasonCreatedVariant, "Variant Workload %q created", klog.KObj(variant))
```
Maybe use %q?

PBundyra: Done

PBundyra: Good point, fixed

mbobrovskyi (approved):
Thank you!

/lgtm

pkg/controller/concurrentadmission/controller_test.go
```diff
@@ -570,6 +608,14 @@ func TestReconcile(t *testing.T) {
 					Active(false).
 					Obj(),
 			},
+			wantEvents: []utiltesting.EventRecord{
+				{
+					Key:       types.NamespacedName{Namespace: "default", Name: "wl-variant-spot"},
+					EventType: corev1.EventTypeNormal,
+					Reason:    ReasonDeactivatedVariant,
+					Message:   "Variant Workload deactivated due to Variant \"default/wl-variant-spot\" below minPreferredFlavor: \"reservation\"",
```
mimowo: 1. the workload name inside the message looks redundant, because the workload is already the "Key" for the event.
2. "due to Variant \"default/wl-variant-spot\" below minPreferredFlavor: \"reservation\"" - I think this does not convey the entire reason. IIUC we deactivate because the Workload is below the `minPreferredFlavor` AND another variant is admitted first, (we may include the name of the variant actually admitted as first)

wdyt?

pkg/controller/concurrentadmission/controller_test.go
```diff
@@ -414,6 +436,14 @@ func TestReconcile(t *testing.T) {
 					Active(true).
 					Obj(),
 			},
+			wantEvents: []utiltesting.EventRecord{
+				{
+					Key:       types.NamespacedName{Namespace: "default", Name: "wl-variant-on-demand"},
+					EventType: corev1.EventTypeNormal,
+					Reason:    ReasonActivatedVariant,
+					Message:   "Variant Workload activated due to Variant \"default/wl-variant-on-demand\" no admitted variant",
```
mimowo: 1. the workload name looks redundant here as it is already the "Key" for the event
2. I don't understand the message "Variant Workload activated due to Variant \"default/wl-variant-on-demand\" no admitted variant"

Can you clarify what are the scenarios leading to the activation of workload, and their corresponding messages?

PBundyra: Removed the name and clarified the message

pkg/controller/concurrentadmission/controller_test.go
PBundyra: Removed the name and clarified the message

pkg/controller/concurrentadmission/controller_test.go
PBundyra: Activation of a Variant may happen when it was previously has been deactivated due to admission of another Variant, and the running Variant has been evicted

mimowo (commented):
/lgtm
/approve
Thank you 👍

PR #10964, opened 2026-05-05
Reject DRA workloads when DynamicResourceAllocation feature gate is disabled

### LINKED ISSUE #10504: [DRA] Workloads admitted when deviceClassMappings is removed from Kueue configuration
**What happened**:

In a Kueue + DRA setup, when `deviceClassMappings` is removed from the Kueue configuration (e.g., via the Kueue CR in OpenShift), Kueue loses all awareness of DRA-managed GPU resources. However, it continues to admit workloads that reference DRA `ResourceClaimTemplates` and unsuspends Jobs as if everything is properly configured.

This means:
  - DRA workloads are admitted and Jobs unsuspended without any warning that `deviceClassMappings` is not configured
  - The `ResourceClaim` is allocated directly by the Kubernetes DRA scheduler, bypassing Kueue's quota entirely
  - The ClusterQueue's GPU resource usage remains at `0` despite a physical GPU being consumed
  - Since Kueue always sees DRA resource usage as `0`, it will admit an unlimited number of DRA workloads regardless of the `nominalQuota` value.

When a DRA workload references a `DeviceClass` that has no corresponding `deviceClassMapping` configured, Kueue could either:
  - Reject admission with a message like: `"Workload references DeviceClass 'gpu.nvidia.com' but no deviceClassMapping is configured"`
  - Or set a warning condition on the Workload indicating the DRA resource is not being quota-managed

  1. Remove `deviceClassMappings` from the Kueue configuration and wait for reconciliation
  2. Verify that `deviceClassMapping` was removed from the kueue-manager-config ConfigMap
  3. Create Kueue resources with a GPU-related `coveredResource` (ResourceFlavor, ClusterQueue, Namespace, and LocalQueue)
  4. Submit a DRA workload referencing a `ResourceClaimTemplate`
  5. Observe the workload is admitted, the Job is unsuspended, and the GPU is consumed outside Kueue's control via DRA
  6. Check `kubectl describe clusterqueue` — GPU usage shows `0` despite physical GPU being consumed through DRA

```
Notice below that `spec.config` has no `deviceClassMappings` — this is the root cause:
$ k get kueue cluster -o json | jq '.spec'
{
  "config": {
    "integrations": {
      "frameworks": [
        "BatchJob",
        "Pod",
        "Deployment",
        "StatefulSet"
      ]
    }
  },
  "logLevel": "Normal",
  "managementState": "Managed",
  "operatorLogLevel": "Normal"
}

❯ k get clusterqueue gpu-cluster-queue -o json | jq '.spec.resourceGroups'
[
  {
    "coveredResources": [
      "cpu",
      "memory",
      "nvidia.com/gpu"
    ],
    "flavors": [
      {
        "name": "gpu-nodes",
        "resources": [
          {
            "name": "cpu",
            "nominalQuota": "1"
          },
          {
            "name": "memory",
            "nominalQuota": "1Gi"
          },
          {
            "name": "nvidia.com/gpu",
            "nominalQuota": "1"
          }

3 workloads admitted, each requesting 1 GPU — but Kueue reports `nvidia.com/gpu` usage as `0`:
$ k get workloads
NAME                 QUEUE        RESERVED IN         ADMITTED   FINISHED   AGE
job-job-dra-ac709    test-queue   gpu-cluster-queue   True                  34m
job-job-dra2-6ab6b   test-queue   gpu-cluster-queue   True                  33m
job-job-dra3-3f6c9   test-queue   gpu-cluster-queue   True                  33m


$ k get clusterqueue gpu-cluster-queue -o json | jq '.status.flavorsUsage'
  [
    {
      "name": "gpu-nodes",
      "resources": [
        {
          "borrowed": "0",
          "name": "cpu",
          "total": "300m"           # 3 jobs x 100m — tracked correctly
        },
        {
          "borrowed": "0",
          "name": "memory",
          "total": "300Mi"          # 3 jobs x 100Mi — tracked correctly
        },
        {
          "borrowed": "0",
          "name": "nvidia.com/gpu",
          "total": "0"              # 3 jobs x 1 GPU — NOT tracked (nominalQuota is 1, but 3 admitted)
        }
      ]
    }
  ]

  $ oc get resourceclaim
  NAME                       STATE                AGE
  job-dra-fxz4b-gpu-5tqwg    allocated,reserved   57m
  job-dra2-w8944-gpu-5wk44   pending              56m
  job-dra3-48cp8-gpu-wd7lt   pending              56m
```

This silently bypasses quota enforcement. Since Kueue never accounts for DRA resource usage, it will admit an unlimited number of DRA workloads regardless of `nominalQuota`. There is no error, warning, or condition indicating the misconfiguration — making it easy to over-commit physical GPUs without any visibility.

**Environment**:
  - Kubernetes version (use `kubectl version`): v1.31
  - Kueue version (use `git describe --tags --dirty --always`): v0.17.0 (Kueue Operator v1.4.0)
  - Cloud provider or hardware configuration: OpenShift cluster with NVIDIA GPUs
  - Kernel (e.g. `uname -a`):
  - Install tools: Kueue Operator
  - Others: DRA feature gates enabled, NVIDIA GPU Operator with DRA Driver

@kannon92

kannon92:
/kind feature
/kind kep

Adds a new `RejectDRAWhenDisabled` feature gate (default: enabled, Beta) that rejects
workloads using DRA resources (ResourceClaimTemplates or ResourceClaims) when the
`DynamicResourceAllocation` feature gate is disabled.

Without this, DRA workloads submitted while `DynamicResourceAllocation` is off are
silently admitted with zero device resource usage, bypassing quota enforcement entirely.
The Kubernetes DRA scheduler then allocates devices directly, allowing unlimited GPU
consumption outside Kueue's control.

The rejection is enforced at two levels:
- **Event handlers**: DRA workloads are not added to the scheduler queue
- **Reconcile loop**: workloads are marked `WorkloadInadmissible` with a clear message

Fixes #10504

The guard is needed in multiple places in workload_controller.go because there are
several code paths that can add a workload to the scheduler queue (Create handler,
Update handler, resource updates handler, immediate requeue path). All must skip DRA
workloads to prevent the scheduler from admitting them before the reconciler can reject.

```release-note
DRA: Add KueueDRARejectWorkloadsWhenDRADisabled feature gate (default: enabled, Beta) that rejects
workloads using DRA resources when KueueDRAIntegration is disabled, preventing silent quota bypass.
```

---

### DIFF
--- a/keps/2941-DRA/README.md
+++ b/keps/2941-DRA/README.md
@@ -655,7 +655,7 @@ The rejection is enforced in the Reconcile loop: workloads are marked with
 with a message indicating that the `DynamicResourceAllocation` feature gate is not enabled.
 
 Administrators who intentionally want to admit DRA workloads without Kueue quota
-management can disable `KueueDRARejectWorkloadsWhenDRADisabled` to restore the previous behavior.
+management can disable `KueueDRARejectWorkloadsWhenDRADisabled` and `DynamicResourceAllocation` to restore the previous behavior.
 
 The steps above are reflected in the complete configuration and workload example below:
 
--- a/keps/2941-DRA/README.md
+++ b/keps/2941-DRA/README.md
@@ -45,6 +45,7 @@ tags, and then generate with `hack/update-toc.sh`.
   - [Workloads](#workloads)
     - [DRA-Specific Workload Processing](#dra-specific-workload-processing)
     - [Workload Processing Flow](#workload-processing-flow)
+    - [Workload Rejection When DRA Is Disabled](#workload-rejection-when-dra-is-disabled)
   - [Extended Resources](#extended-resources)
     - [Configuration](#configuration)
     - [Path Separation](#path-separation)
@@ -286,6 +287,12 @@ resubmission of the workload once ResourceSlices changes. Evaluating this for be
 
 ### Risks and Mitigations
 
+**Silent quota bypass when DRA is disabled**: When the `DynamicResourceAllocation` feature
+gate is disabled, DRA workloads are admitted without any device resource accounting, allowing
+unlimited GPU consumption outside Kueue's control. The `KueueDRARejectWorkloadsWhenDRADisabled` feature gate
+(default: enabled, Beta) mitigates this by rejecting DRA workloads when the DRA feature is off.
+See [Workload Rejection When DRA Is Disabled](#workload-rejection-when-dra-is-disabled).
+
 With DRAAdminAccess and DRAPrioritizedLists (both beta, default enabled in K8s 1.35), there is a
 risk that effective tallying of resources will not be available until after allocation of these
 resources by kube-scheduler.
@@ -323,12 +330,18 @@ In order to mitigate this risk, Kueue can take the following approach:
 
 ## Design Details
 
-Two feature gates control DRA support in Kueue:
+Three feature gates control DRA support in Kueue:
 - `DynamicResourceAllocation`: gates ResourceClaimTemplate-based DRA quota accounting.
   Uses `deviceClassMappings` for DeviceClass-to-quota-resource mapping.
 - `DRAExtendedResources`: gates extended resources support, including DeviceClass
   auto-discovery via `extendedResourceName`. Does not use `deviceClassMappings`.
   Requires `DynamicResourceAllocation` to also be enabled.
+- `KueueDRARejectWorkloadsWhenDRADisabled` (default: enabled, Beta since v0.18): rejects workloads that
+  use DRA resources (ResourceClaimTemplates or ResourceClaims) when the
+  `DynamicResourceAllocation` feature gate is disabled. Without this gate, DRA workloads
+  submitted while `DynamicResourceAllocation` is off are silently admitted with zero
+  device resource usage, bypassing quota enforcement entirely. See
+  [Workload Rejection When DRA Is Disabled](#workload-rejection-when-dra-is-disabled).
 
 The following sections will explain the design in detail.
 
@@ -623,6 +636,26 @@ When the `DRAExtendedResources` gate is also enabled, workloads with extended re
 `resources.requests` follow a separate resolution path through the ExtendedResourceCache.
 See [Extended Resources](#extended-resources) for details. Both paths can be active simultaneously
 for workloads that use both ResourceClaimTemplates and extended resources.
+
+#### Workload Rejection When DRA Is Disabled
+
+When the `DynamicResourceAllocation` feature gate is disabled, the DRA processing pipeline
+is skipped entirely. Without additional safeguards, workloads that reference
+ResourceClaimTemplates or ResourceClaims are silently admitted based on CPU/memory only,
+with zero device resource usage recorded. This allows unlimited DRA workloads to bypass
+quota enforcement, since the Kubernetes DRA scheduler still allocates devices directly.
+
+The `KueueDRARejectWorkloadsWhenDRADisabled` feature gate (default: enabled, Beta) closes this gap. When
+enabled and `DynamicResourceAllocation` is disabled, Kueue detects workloads with DRA
+resources (via `HasDRA()` which checks for `ResourceClaimTemplateName` or
+`ResourceClaimName` in any PodSet) and rejects them as inadmissible.
+
+The rejection is enforced in the Reconcile loop: workloads are marked with
+`WorkloadQuotaReserved=False` (reason: `WorkloadInadmissible`) and `WorkloadRequeued=False`,
+with a message indicating that the `DynamicResourceAllocation` feature gate is not enabled.
+
+Administrators who intentionally want to admit DRA workloads without Kueue quota
+management can disable `KueueDRARejectWorkloadsWhenDRADisabled` to restore the previous behavior.
 
 The steps above are reflected in the complete configuration and workload example below:
 
@@ -1000,6 +1033,9 @@ using mock ResourceClaimTemplates and DeviceClasses to simulate DRA workloads. K
 - Late DeviceClass creation: Testing workload inadmissibility when DeviceClass does not exist
 - CEL validation: Testing CEL compilation errors, evaluation against ResourceSlice devices, and rejection
   of workloads with unsatisfiable CEL selectors
+- DRA disabled rejection: Testing that workloads with ResourceClaimTemplates or ResourceClaims are
+  rejected as inadmissible when `DynamicResourceAllocation` is off and `KueueDRARejectWorkloadsWhenDRADisabled` is on,
+  and that non-DRA workloads are still admitted normally
 
 #### E2E Test
 
@@ -1057,6 +1093,9 @@ Use existing dra-example-driver or Kubernetes test driver for e2e testing.
   confirming DRA logical resources work with existing `AdmissionFairSharing.ResourceWeights`
 - CEL expression validation support added: April 2026 by @kannon92
 - Promoted KueueDRAIntegration to Beta: May 2026 by @sohankunkerkar
+- `KueueDRARejectWorkloadsWhenDRADisabled` feature gate added: May 2026 by @kannon92 — rejects DRA workloads
+  when the `DynamicResourceAllocation` feature gate is disabled to prevent silent quota bypass
+  (see [#10504](https://github.com/kubernetes-sigs/kueue/issues/10504))
 
 **Key Design Evolution:**
 - **Original Design**: Standalone DynamicResourceAllocationConfig CRD with runtime ambiguity resolution
--- a/keps/2941-DRA/kep.yaml
+++ b/keps/2941-DRA/kep.yaml
@@ -37,6 +37,7 @@ milestone:
 feature-gates:
   - name: KueueDRAIntegration
   - name: KueueDRAIntegrationExtendedResource
+  - name: KueueDRARejectWorkloadsWhenDRADisabled
 disable-supported: true
 
 # The following PRR answers are required at beta release
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -293,6 +293,26 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		})
 		return reconcile.Result{}, client.IgnoreNotFound(err)
 	}
+	if workload.Status(&wl) == workload.StatusPending &&
+		!features.Enabled(features.KueueDRAIntegration) &&
+		features.Enabled(features.KueueDRARejectWorkloadsWhenDRADisabled) &&
+		workload.HasDRA(&wl) {
+		log.V(3).Info("Rejecting workload that uses DRA resources because KueueDRAIntegration feature gate is disabled")
+		err := workload.PatchAdmissionStatus(ctx, r.client, &wl, r.clock, func(wl *kueue.Workload) (bool, error) {
+			updated := workload.UnsetQuotaReservationWithCondition(wl, kueue.WorkloadInadmissible,
+				"Workload uses DRA resources but the KueueDRAIntegration feature gate is not enabled",
+				r.clock.Now())
+			if workload.SetRequeuedCondition(wl, kueue.WorkloadInadmissible,
+				"Workload uses DRA resources but the KueueDRAIntegration feature gate is not enabled", false) {
+				updated = true
+			}
+			return updated, nil
+		})
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to update workload status for DRA rejection: %w", err)
+		}
+		return ctrl.Result{}, nil
+	}
 	if workload.Status(&wl) == workload.StatusPending && dra.NeedsDRAReconcile(&wl) {
 		workload.AdjustResources(ctx, r.client, &wl)
 		if workload.HasResourceClaim(&wl) {
@@ -1203,9 +1223,10 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 			}
 		})
 	case prevStatus == workload.StatusPending && status == workload.StatusPending:
-		if dra.NeedsDRAReconcile(e.ObjectNew) {
+		switch {
+		case dra.NeedsDRAReconcile(e.ObjectNew):
 			log.V(2).Info("Skipping queue update for DRA workload - handled in Reconcile")
-		} else {
+		default:
 			if err := r.queues.AddOrUpdateWorkload(log, wlCopy); err != nil {
 				log.V(2).Info("ignored an error for now", "error", err)
 			}
@@ -1242,9 +1263,10 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 			// AddOrUpdateWorkload is only called once. When moving it to the main
 			// reconciler, we would execute it on every run, which might mess up the state.
 			if immediate {
-				if dra.NeedsDRAReconcile(e.ObjectNew) {
+				switch {
+				case dra.NeedsDRAReconcile(e.ObjectNew):
 					log.V(2).Info("Skipping immediate requeue for DRA workload - handled in Reconcile")
-				} else {
+				default:
 					if err := r.queues.AddOrUpdateWorkloadWithoutLock(log, wlCopy); err != nil {
 						log.V(2).Info("ignored an error for now", "error", err)
 					}
--- a/test/integration/singlecluster/controller/dra/dra_test.go
+++ b/test/integration/singlecluster/controller/dra/dra_test.go
@@ -1132,4 +1132,125 @@ var _ = ginkgo.Describe("DRA Integration", ginkgo.Ordered, ginkgo.ContinueOnFail
 			}, util.Timeout, util.Interval).Should(gomega.Succeed())
 		})
 	})
+
+	ginkgo.When("KueueDRAIntegration disabled with KueueDRARejectWorkloadsWhenDRADisabled enabled", func() {
+		var (
+			ns             *corev1.Namespace
+			resourceFlavor *kueue.ResourceFlavor
+			clusterQueue   *kueue.ClusterQueue
+			localQueue     *kueue.LocalQueue
+		)
+
+		ginkgo.BeforeEach(func() {
+			features.SetFeatureGateDuringTest(ginkgo.GinkgoTB(), features.KueueDRAIntegration, false)
+			features.SetFeatureGateDuringTest(ginkgo.GinkgoTB(), features.KueueDRARejectWorkloadsWhenDRADisabled, true)
+			ns = &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: "dra-reject-",
+				},
+			}
+			gomega.Expect(k8sClient.Create(ctx, ns)).To(gomega.Succeed())
+
+			resourceFlavor = utiltestingapi.MakeResourceFlavor("").Obj()
+			resourceFlavor.GenerateName = "rf-reject-"
+			gomega.Expect(k8sClient.Create(ctx, resourceFlavor)).To(gomega.Succeed())
+
+			clusterQueue = &kueue.ClusterQueue{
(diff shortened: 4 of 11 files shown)


kannon92:
/hold

@sohankunkerkar and I have been discussing this offline. Not sure if we should merge this yet.

mimowo:
I will review this one this week. Rejecting DRA workloads seems to me like a better choice than rushing DRA graduation

keps/2941-DRA/kep.yaml
```diff
@@ -37,6 +37,7 @@ milestone:
 feature-gates:
   - name: DynamicResourceAllocation
   - name: DRAExtendedResources
+  - name: RejectDRAWhenDisabled
```
mimowo: I would propose to add the "DRA" prefix, and also add "Workloads" to make it clear what is rejected, so maybe : DRARejectWorkloadsWhenDRADisabled

pkg/controller/core/workload_controller.go
```diff
@@ -282,6 +282,26 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
 		return reconcile.Result{}, client.IgnoreNotFound(err)
 	}
+	if workload.Status(&wl) == workload.StatusPending &&
+		!features.Enabled(features.DynamicResourceAllocation) &&
+		features.Enabled(features.RejectDRAWhenDisabled) &&
+		workload.HasDRA(&wl) {
+		log.V(3).Info("Rejecting workload that uses DRA resources because DynamicResourceAllocation feature gate is disabled")
+		err := workload.PatchAdmissionStatus(ctx, r.client, &wl, r.clock, func(wl *kueue.Workload) (bool, error) {
+			updated := workload.UnsetQuotaReservationWithCondition(wl, kueue.WorkloadInadmissible,
+				"Workload uses DRA resources but the DynamicResourceAllocation feature gate is not enabled",
```
mimowo: Let's maybe use the same message in both cases, say:

msg := "Workload uses DRA resources but the DynamicResourceAllocation feature gate is not enabled"

and use it both for UnsetQuotaReservationWithCondition, and SetRequeuedCondition

pkg/controller/core/workload_controller.go
```diff
@@ -1087,6 +1107,13 @@ func (r *WorkloadReconciler) Create(e event.TypedCreateEvent[*kueue.Workload]) b
 		return true
 	}
 
+	if !features.Enabled(features.DynamicResourceAllocation) &&
+		features.Enabled(features.RejectDRAWhenDisabled) &&
+		workload.HasDRA(e.Object) {
+		log.V(2).Info("Skipping DRA workload in Create event - DRA feature gate is disabled, will be rejected in Reconcile")
+		return true
+	}
```
mimowo: Is this really needed? We already handle them in Reconcile

pkg/controller/core/workload_controller.go
```diff
@@ -1182,9 +1209,14 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
...
 	case prevStatus == workload.StatusPending && status == workload.StatusPending:
-		if dra.NeedsDRAReconcile(e.ObjectNew) {
+		switch {
+		case dra.NeedsDRAReconcile(e.ObjectNew):
 			log.V(2).Info("Skipping queue update for DRA workload - handled in Reconcile")
-		} else {
+		case !features.Enabled(features.DynamicResourceAllocation) &&
+			features.Enabled(features.RejectDRAWhenDisabled) &&
+			workload.HasDRA(e.ObjectNew):
+			log.V(2).Info("Skipping queue update for DRA workload - DRA feature gate is disabled, will be rejected in Reconcile")
```
mimowo: I think these new code injections pollute the Event handlers code, so I think we could drop them, and only handle in Reconcile. If we skip in all event handlers, then the code in Reconcile is not needed. 

Choosing between the two I would prefer to only cover in Reconcile

test/integration/singlecluster/controller/dra/dra_test.go
```diff
@@ -1046,4 +1047,140 @@ var _ = ginkgo.Describe("DRA Integration", ginkgo.Ordered, ginkgo.ContinueOnFail
...
+			clusterQueue   *kueue.ClusterQueue
+			localQueue     *kueue.LocalQueue
+		)
+
+		ginkgo.BeforeAll(func() {
+			gomega.Expect(utilfeature.DefaultMutableFeatureGate.SetFromMap(map[string]bool{
+				string(features.DynamicResourceAllocation): false,
+				string(features.RejectDRAWhenDisabled):     true,
+			})).Should(gomega.Succeed())
+			fwk.StopManager(ctx)
```
mimowo: I'm puzzled why the StopManager is called here in BeforeAll. Checking other places in code we don't do that, for example here: https://github.com/kubernetes-sigs/kueue/blob/main/test/integration/singlecluster/controller/jobs/job/job_controller_test.go#L3576-L3582

sohankunkerkar:
> I will review this one this week. Rejecting DRA workloads seems to me like a better choice than rushing DRA graduation

My 2 cents: the disabled state should fail safe, agreed. But that should be the permanent ungated behavior, not a separate feature gate with its own lifecycle. I also don’t think it should block beta graduation. Moving DRA to beta removes the quota-bypass path for the default configuration, which is the main safety concern. We can still land the fail-safe behavior as a fast follow in the same release for the explicit-disabled case.

mimowo:
> My 2 cents: the disabled state should fail safe, agreed. But that should be the permanent ungated behavior, not a separate feature gate with its own lifecycle. 

Right it should be enabled by default.The only reason for the FG here is to safeguard addition of the new code. New code is rarely added without FG to a prod grade project so that users can mitigate bugs in the nee code by disabling FG rather than by downgrade or waiting ob patch version.

mimowo:
> Moving DRA to beta removes the quota-bypass path for the default configuration, which is the main safety concern. We can still land the fail-safe behavior as a fast follow in the same release for the explicit-disabled case.

Yes moving DRA to Beta could make this FG unnecessary. At the same time DRA integration has a lot of new code being added just now, so my concern about DRA beta is about maturity, so that we dont break prod users who dont use dra by enabling this new code by default. I havent made my mind here yet. This PR allows to derisk Kueue if we decide to delay graduation of DRA  to next release

mimowo:
> > My 2 cents: the disabled state should fail safe, agreed. But that should be the permanent ungated behavior, not a separate feature gate with its own lifecycle. 
> 
> Right it should be enabled by default.The only reason for the FG here is to safeguard addition of the new code. New code is rarely added without FG to a prod grade project so that users can mitigate bugs in the nee code by disabling FG rather than by downgrade or waiting ob patch version.

 Actually, I take it back. When DRA is stable then we dont need the Reject code, right? in that case it is better to have the rejection code behind FG yet for another reason - so that we can safely decommission the code when DRA becones stable

kannon92:
> Actually, I take it back. When DRA is stable then we dont need the Reject code, right? in that case it is better to have the rejection code behind FG yet for another reason - so that we can safely decommission the code when DRA becones stable

You could argue that if DRA is beta on we could remove the reject feature gate. If someone is disabling the beta feature gate then they can deal with the downsides of that.

keps/2941-DRA/kep.yaml
kannon92: updated.

pkg/controller/core/workload_controller.go
kannon92: removed. you are correct.

test/integration/singlecluster/controller/dra/dra_test.go
kannon92: updated.

kannon92:
@mimowo I believe I have addressed all your outstanding requests.

Following offline, it seems we still should go with a feature gate to protect production users and it is not yet set in stone that DRA will go beta in 0.18.

pkg/controller/core/workload_controller.go
kannon92: Fixed this.

pkg/controller/core/workload_controller.go
kannon92: addressed

mimowo (commented):
/lgtm
/approve
Thanks 👍

sohankunkerkar:
@alaypatel07 is planning to review this PR

keps/2941-DRA/kep.yaml
```diff
@@ -37,6 +37,7 @@ milestone:
 feature-gates:
   - name: DynamicResourceAllocation
   - name: DRAExtendedResources
+  - name: DRARejectWorkloadsWhenDRADisabled
```
alaypatel07: I am unsure if this is a feature gate or a configuration option with DRA. I wonder if it would be more appropriate to have this in Kueue Configuration?

pkg/controller/core/workload_controller.go
```diff
@@ -1183,9 +1203,10 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 			}
 		})
 	case prevStatus == workload.StatusPending && status == workload.StatusPending:
```
alaypatel07, line 1225: What happens in the following case:
1. DRA feature gate is on, workload is admitted 
2. DRA feature gate is turned off

Do we expect admin to clean up the workload or should it somehow be evicted in the reconcile loop?

alaypatel07 (commented):
Overall the concept here makes sense to me, added couple of questions in-line, will do another pass tomorrow morning.

pkg/controller/core/workload_controller.go
kannon92: I'm not sure what our project does in these cases. Feature gates are usually applied on startup or upgrades.

alaypatel07: I'm not opinionated one of the pros of config change is that it can be back-ported as bug fix is needed. Can't think of any other benefit

kannon92: Feature gate or config are both fields in the configuration so its really the same operational aspect.

keps/2941-DRA/kep.yaml
kannon92: The reason for the feature gate is temporary. When DRA feature gate graduates we would remove it. 

If we add this to a configuration this would end up becoming a dead API field. I personally think a feature gate is better in this scenario.

mimowo: Yes, this is not a config field, because its lifetime is limited. The lifetime of config api is infinite, or at least until v1beta3

pkg/controller/core/workload_controller.go
alaypatel07: I think its okay for admins to clean it up in case feature gate is turned off. As long as it is documented.

pkg/controller/core/workload_controller.go
```diff
@@ -1183,9 +1203,10 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
...
-		if dra.NeedsDRAReconcile(e.ObjectNew) {
+		switch {
```
alaypatel07, line 1226: Out of curiosity, is this change improving the readability? Same question for the switch case below

alaypatel07 (approved):
Added a question on an syntactic change.

Apart from that the functionality itself looks good to me.

pkg/controller/core/workload_controller.go
kannon92: I have if, else if, else if, but linter said to switch to switch.

kannon92: > Key gocritic switch-related checkers include:
ifElseChain: Suggests rewriting if-else chains with 3 or more blocks into a switch statement for better readability.

alaypatel07: ahh good to know, thanks

pkg/controller/core/workload_controller.go
kannon92: I'm not even sure if we should recommend this. If someone changes Kueue behavior do you really recommend that admin should delete all the jobs that use DRA? 

The workload object is tied to the generic job object (Jobset, Job, etc) so admin should remove the true workload. But that seems disruptive to users so I'm not sure we should add this docs.

Maybe we should just add to documentation that if the feature gate is disabled, Kueue willl not monitor quota from DRA resources?

alaypatel07: My concern is that those jobs will keep hogging quota and could block other jobs waiting in-line.  Best for admin to decide on case-by-case basis?

kannon92: This feels a bit of an rare edge case though. I guess you could make a case that if you disable the feature gate, then you should remove the jobs that use DRA + Kueue together. 

I'm not sure we have much testing/support features like this in Kubernetes since the upgrade process is usually manual. 

@mimowo @tenzen-y WDYT of this?

kannon92:
@tenzen-y 

Do you want to take a quick look? Otherwise I will remove the hold now that we have the labels.

kannon92:
/hold cancel

This got enough approvals and I don't see anything blocking merge.

tenzen-y:
> @tenzen-y 
> 
> 
> 
> Do you want to take a quick look? Otherwise I will remove the hold now that we have the labels.

I will check this in the early next week.

/hold

pkg/controller/core/workload_controller.go
mimowo: > What happens in the following case:
> 
> 1. DRA feature gate is on, workload is admitted
> 2. DRA feature gate is turned off
> 
> Do we expect admin to clean up the workload or should it somehow be evicted in the reconcile loop?

IIUC: When DRA feature gate gets disabled, then Kueue restart is required (only this way FG gets applied). At this point the workload_controller will receive "ADDED" events for all workloads via "Create" even handler, triggering "Reconcile". Now the new code will evict all the workloads so that they stop using the quota, which is good. I don't think we expect admins to cleanup the workloads.

However, we may test e2e on a running cluster if then the evicted workloads are not getting re-admitted, because that could put then on an infinite admit / evict loop. Could you check that @kannon92 ?

tenzen-y:
I am landing here and checking this now.
@kannon92 can you rebase this PR on top of the latest main branch?

tenzen-y (commented):
@kannon92 Thank you for moving this forward. Almost LGTM
But can you add FG dependencies validations to https://github.com/kubernetes-sigs/kueue/blob/91387fe1625d1c4250555497bde1add4b40f04d0/pkg/config/validation.go#L579?

kannon92:
> @kannon92 Thank you for moving this forward. Almost LGTM But can you add FG dependencies validations to
> 
> https://github.com/kubernetes-sigs/kueue/blob/91387fe1625d1c4250555497bde1add4b40f04d0/pkg/config/validation.go#L579
> 
> ?

edit: sorry I think I get this. I should not allow DRA and RejectDRA feature gates to be both false.

pkg/controller/core/workload_controller.go
kannon92: I'm at a conference this week. I'll check this once I am back to work.

Can we leave this a followup?

tenzen-y:
> > @kannon92 Thank you for moving this forward. Almost LGTM But can you add FG dependencies validations to
> > https://github.com/kubernetes-sigs/kueue/blob/91387fe1625d1c4250555497bde1add4b40f04d0/pkg/config/validation.go#L579
> > 
> > ?
> 
> I'm not really sure what I should put here.
> 
> There isn't really any dependency validation here. I have the code so that if DRA feature gate is enabled this can also be enabled.

Sorry for the confusion. I wanted to restrict enabling RejectDRAWhenDisabled only when DynamicResourceAllocation FG is enabled.

kannon92:
> Sorry for the confusion. I wanted to restrict enabling RejectDRAWhenDisabled only when DynamicResourceAllocation FG is enabled.

I don't think we actually need this. I did add a validation to make sure that rejectDRA can not disabled if DRA is disabled. 

My code covers the case if RejectDRA is enabled and DRA is enabled. We only check the reject path if DRA is disabled. 

So there isn't really any reason to validate that.

tenzen-y:
> > Sorry for the confusion. I wanted to restrict enabling RejectDRAWhenDisabled only when DynamicResourceAllocation FG is enabled.
> 
> I don't think we actually need this. I did add a validation to make sure that rejectDRA can not disabled if DRA is disabled.
> 
> My code covers the case if RejectDRA is enabled and DRA is enabled. We only check the reject path if DRA is disabled.
> 
> So there isn't really any reason to validate that.

Yes, behavior is correct. But, validation can be helpful to notice misconfigurations.
We should tell misconfigurations to users when RejectDRAWhenDisabled is enabled, but DynamicResourceAllocation FG is disabled.

kannon92:
> We should tell misconfigurations to users when RejectDRAWhenDisabled is enabled, but DynamicResourceAllocation FG is disabled.

This is not a misconfiguration but the happy path for this feature.

Since DRA is beta I guess this PR may not be needed actually. But the main goal was to not allow kueue users to submit DRA workloads when DRA FG is disabled. And so we would reject those workloads for admission.

PTAL at what I have and see if this is fine. (Do not allow rejectDRA false and DRA false). That would be the case where DRA workloads can be ignored by Kueue and go directly to scheduler. That is the main configuration I want to fail on validation for.

keps/2941-DRA/README.md
```diff
@@ -624,6 +637,26 @@ When the `DRAExtendedResources` gate is also enabled, workloads with extended re
...
+enabled and `DynamicResourceAllocation` is disabled, Kueue detects workloads with DRA
+resources (via `HasDRA()` which checks for `ResourceClaimTemplateName` or
+`ResourceClaimName` in any PodSet) and rejects them as inadmissible.
+
+The rejection is enforced in the Reconcile loop: workloads are marked with
+`WorkloadQuotaReserved=False` (reason: `WorkloadInadmissible`) and `WorkloadRequeued=False`,
+with a message indicating that the `DynamicResourceAllocation` feature gate is not enabled.
+
+Administrators who intentionally want to admit DRA workloads without Kueue quota
+management can disable `KueueDRARejectWorkloadsWhenDRADisabled` to restore the previous behavior.
```
```suggestion
```

I guess that this is already stale.

tenzen-y:
> > We should tell misconfigurations to users when RejectDRAWhenDisabled is enabled, but DynamicResourceAllocation FG is disabled.
> 
> This is not a misconfiguration but the happy path for this feature.
> 
> Since DRA is beta I guess this PR may not be needed actually. But the main goal was to not allow kueue users to submit DRA workloads when DRA FG is disabled. And so we would reject those workloads for admission.
> 
> PTAL at what I have and see if this is fine. (Do not allow rejectDRA false and DRA false). That would be the case where DRA workloads can be ignored by Kueue and go directly to scheduler. That is the main configuration I want to fail on validation for.

Oh, I see. I might have been misunderstanding. In that case, I'm fine with the current validations.

keps/2941-DRA/README.md
kannon92: Not quite. I guess technically they would need to disable Reject DRA and DRA features gates together.

tenzen-y: We actually don't allow that here: https://github.com/kannon92/kueue/blob/6995a69791d3c988dee2e2049734aa97d664423c/pkg/config/validation.go#L586-L593

kannon92: Hmm so this does mean there isn't really a good way to opt out of this behavior with the validation.

Should we just remove the validation then?

tenzen-y: Based on discussion offline, we decided to remove validation. 

Thanks Kevin.

kannon92: Removed.

kannon92:
/hold cancel

Based on latest feedback I think this can merge once @tenzen-y gives the lgtm.

tenzen-y (commented):
@kannon92 Thank you!
/lgtm
/approve

/hold cancel

mimowo:
/release-note-edit
```release-note
DRA: Add KueueDRARejectWorkloadsWhenDRADisabled feature gate (default: enabled, Beta) that rejects
workloads using DRA resources when KueueDRAIntegration is disabled, preventing silent quota bypass.
```

PR #10966, opened 2026-05-05
Simplify  TAS validation webhook

### LINKED ISSUE #7297: Simplify TAS-related validation based on the pattern in Kubeflow

We have rather complex validation steps for TAS, see here: https://github.com/kubernetes-sigs/kueue/blob/a02abfed9173e74dcf5fc074e68d8d0cb7e6a09f/pkg/controller/jobs/jobset/jobset_webhook.go#L142-L168

I think the major issue was to know the validation path per PodSet, but with the recent changes in https://github.com/kubernetes-sigs/kueue/pull/7061 we can now follow this much simpler pattern:
https://github.com/kubernetes-sigs/kueue/pull/7249/files#diff-5e480b6f7a8d2f0786c69430f298412c1045bdda45790b865c7be63e55b1441cR145-R162

or at least investigate that deeper.

**Why is this needed**:

To simplify the validation code.

kaisoz:
#### What type of PR is this?

/kind cleanup

TAS validation in the job's webhook is too complex. This PR simplifies it based on the implementation in the TrainJob webhook

Fixes #7297

```release-note
NONE
```

### DIFF
--- a/pkg/controller/jobs/mpijob/mpijob_webhook.go
+++ b/pkg/controller/jobs/mpijob/mpijob_webhook.go
@@ -35,17 +35,20 @@ import (
 	"sigs.k8s.io/kueue/pkg/controller/jobframework"
 	"sigs.k8s.io/kueue/pkg/features"
 	"sigs.k8s.io/kueue/pkg/util/kubeversion"
-	"sigs.k8s.io/kueue/pkg/util/podset"
 	"sigs.k8s.io/kueue/pkg/util/webhook"
 )
 
 var (
-	mpiReplicaSpecsPath         = field.NewPath("spec", "mpiReplicaSpecs")
-	launcherAnnotationsPath     = mpiReplicaSpecsPath.Key(string(v2beta1.MPIReplicaTypeLauncher)).Child("template", "metadata", "annotations")
-	workerAnnotationsPath       = mpiReplicaSpecsPath.Key(string(v2beta1.MPIReplicaTypeWorker)).Child("template", "metadata", "annotations")
+	mpiReplicaSpecsPath      = field.NewPath("spec", "mpiReplicaSpecs")
+	launcherMetadataPath     = mpiReplicaSpecsPath.Key(string(v2beta1.MPIReplicaTypeLauncher)).Child("template", "metadata")
+	workerMetadataPath       = mpiReplicaSpecsPath.Key(string(v2beta1.MPIReplicaTypeWorker)).Child("template", "metadata")
+	podSetMetadataPathByName = map[kueue.PodSetReference]*field.Path{
+		kueue.NewPodSetReference(string(v2beta1.MPIReplicaTypeLauncher)): launcherMetadataPath,
+		kueue.NewPodSetReference(string(v2beta1.MPIReplicaTypeWorker)):   workerMetadataPath,
+	}
 	podSetAnnotationsPathByName = map[kueue.PodSetReference]*field.Path{
-		kueue.NewPodSetReference(string(v2beta1.MPIReplicaTypeLauncher)): launcherAnnotationsPath,
-		kueue.NewPodSetReference(string(v2beta1.MPIReplicaTypeWorker)):   workerAnnotationsPath,
+		kueue.NewPodSetReference(string(v2beta1.MPIReplicaTypeLauncher)): launcherMetadataPath.Child("annotations"),
+		kueue.NewPodSetReference(string(v2beta1.MPIReplicaTypeWorker)):   workerMetadataPath.Child("annotations"),
 	}
 )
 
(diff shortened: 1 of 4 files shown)


pkg/controller/jobs/mpijob/mpijob_webhook.go
```diff
@@ -175,21 +178,13 @@ func (w *MpiJobWebhook) validateTopologyRequest(ctx context.Context, mpiJob *MPI
...
-	}
-
-	for replicaType, replicaSpec := range mpiJob.Spec.MPIReplicaSpecs {
-		replicaMetaPath := mpiReplicaSpecsPath.Key(string(replicaType)).Child("template", "metadata")
-		allErrs = append(allErrs, jobframework.ValidateTASPodSetRequest(replicaMetaPath, &replicaSpec.Template.ObjectMeta)...)
-
-		if podSetsErr != nil {
-			continue
+		for _, p := range podSets {
+			replicaMetaPath := podSetMetadataPathByName[p.Name]
```
mimowo, line 183: Can we be sure there are no other types? I'm wondering if we could generalize that so that we would not panic if another worker type is introduced or used

mimowo (commented):
LGTM, thanks 👍 I have a small doubt about the mpi operator if we can assume the set of replica types is closed.

pkg/controller/jobs/mpijob/mpijob_webhook.go
kaisoz: The replica types are defined by the Kubeflow MPI operator API and currently [those are the only ones available](https://pkg.go.dev/github.com/kubeflow/mpi-operator@v0.8.0/pkg/apis/kubeflow/v2beta1#MPIReplicaType) and [the only ones we consider](https://github.com/kubernetes-sigs/kueue/blob/b4deeaf4ed3e4816c733039a233761b21a0d63f5/pkg/controller/jobs/mpijob/mpijob_controller.go#L221). Adding a new replica type would definitely require some changes to the controller and therefore to the webhook, as issues would likely surface during development.

mimowo: Yes, we certainly don't support other replicaTypes, but I'm wondering more about creating maliciously mpi job with a custom replica type, rather than about proper support.

Just for sanity check as we are on this - could you check experimentally what would happen on a real cluster on a release, and on this PR if one tries to create an MPI job with an unrecognized replicaType. I'm wondering if this could be a vector of attack to make Kueue panic. 

If this panic does not exist - great. If it existed prior to the PR, then also ok, but I would like to follow up with a bugfix to eliminate this volnurability.

kaisoz: I just tested with an unknown `ReplicaType` and it worked properly. Let me explain de code flow:

1. [The MPI job TAS validation](https://github.com/kaisoz/kueue/blob/c56c5a63c778f6ddcbb5c72c6613882df190afaa/pkg/controller/jobs/mpijob/mpijob_webhook.go#L177) is run on each validation webhook event
2. The first thing the validator does is [call the ](https://github.com/kaisoz/kueue/blob/c56c5a63c778f6ddcbb5c72c6613882df190afaa/pkg/controller/jobs/mpijob/mpijob_webhook.go#L180)Podsets[ function of the controller](https://github.com/kaisoz/kueue/blob/c56c5a63c778f6ddcbb5c72c6613882df190afaa/pkg/controller/jobs/mpijob/mpijob_webhook.go#L180) to get the podsets
3. The PodSets function implementation for MPI jobs [first extracts the available ReplicaTypes](https://github.com/kaisoz/kueue/blob/c56c5a63c778f6ddcbb5c72c6613882df190afaa/pkg/controller/jobs/mpijob/mpijob_controller.go#L113). A here's where the safety net is. Because the `MPIjob.Spec.MPIReplicaSpecs` is a map, the function  checks and returns only supported `ReplicaTypes`. This means that if you add a malicious `ReplicaType`, the controller just ignores it and doesn't create a `PodSet` for it (this is what I tested). The returned `PodSets` list will always be valid (in that sense)
4. And coming back to the webhook. Since [it loops the returned `PodSets` list](https://github.com/kaisoz/kueue/blob/c56c5a63c778f6ddcbb5c72c6613882df190afaa/pkg/controller/jobs/mpijob/mpijob_webhook.go#L182) and they are already validated, it won't break

I've added a test to confirm this 👍🏻

mimowo (commented):
/lgtm
/approve
Thank you 👍

pkg/controller/jobs/mpijob/mpijob_webhook.go
mimowo: Thank you for investigating that 👍

PR #10252, opened 2026-04-01
Add RayService e2e autoscaling test

hiboyang:
#### What type of PR is this?

/area testing

#### What this PR does / why we need it:

This is to add e2e testing for RayService autoscaling.

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #10967, opened 2026-05-05
Add RayService e2e autoscaling test (#10252)

hiboyang:
#### What type of PR is this?

/kind cleanup

Cherry pick https://github.com/kubernetes-sigs/kueue/pull/10252 to 0.17

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/util/testingjobs/rayjob/wrappers.go
+++ b/pkg/util/testingjobs/rayjob/wrappers.go
@@ -344,7 +344,7 @@ func (j *JobWrapper) Annotation(key string, value string) *JobWrapper {
 func (j *JobWrapper) EnableInTreeAutoscaling() *JobWrapper {
 	enable := true
 	aggressive := rayv1.UpscalingMode("Aggressive")
-	idleTimeoutSeconds := int32(60)
+	idleTimeoutSeconds := int32(10)
 	j.Spec.RayClusterSpec.EnableInTreeAutoscaling = &enable
 	j.Spec.RayClusterSpec.AutoscalerOptions = &rayv1.AutoscalerOptions{
 		UpscalingMode:      &aggressive,
--- a/test/e2e/singlecluster/kuberay_test.go
+++ b/test/e2e/singlecluster/kuberay_test.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
@@ -77,15 +78,33 @@ var _ = ginkgo.Describe("Kuberay", ginkgo.Label("area:singlecluster", "feature:k
 		localQueueName     string
 	)
 
-	// getRunningWorkerPodNames returns the names of running pods that have "workers" in their name
-	getRunningWorkerPodNames := func(podList *corev1.PodList) []string {
+	getRunningRayWorkerPodNames := func(g gomega.Gomega) []string {
+		pods := &corev1.PodList{}
+		g.Expect(k8sClient.List(ctx, pods,
+			client.InNamespace(ns.Name),
+			client.MatchingLabels{
+				"ray.io/node-type": "worker",
+			},
+		)).To(gomega.Succeed())
 		var podNames []string
-		for _, pod := range podList.Items {
+		for _, pod := range pods.Items {
 			if strings.Contains(pod.Name, "workers") && pod.Status.Phase == corev1.PodRunning {
 				podNames = append(podNames, pod.Name)
 			}
 		}
 		return podNames
+	}
+
+	getRayHeadPod := func(g gomega.Gomega) corev1.Pod {
+		pods := &corev1.PodList{}
+		g.Expect(k8sClient.List(ctx, pods,
+			client.InNamespace(ns.Name),
+			client.MatchingLabels{
+				"ray.io/node-type": "head",
+			},
+		)).To(gomega.Succeed())
+		g.Expect(pods.Items).NotTo(gomega.BeEmpty())
+		return pods.Items[0]
 	}
 
 	ginkgo.BeforeEach(func() {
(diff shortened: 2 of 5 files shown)


pkg/util/testingjobs/rayjob/wrappers.go
```diff
@@ -344,7 +344,7 @@ func (j *JobWrapper) Annotation(key string, value string) *JobWrapper {
 func (j *JobWrapper) EnableInTreeAutoscaling() *JobWrapper {
 	enable := true
 	aggressive := rayv1.UpscalingMode("Aggressive")
-	idleTimeoutSeconds := int32(60)
+	idleTimeoutSeconds := int32(10)
```
mimowo, line 347: Is this changed intentionally or during conflict resolution?

hiboyang: This is changed intentially due to flaky test failure. I still get other test failure in this branch, while not get them in main branch, still investigating.

mimowo: hm, I thought the failures are due to the change actually

maybe @sohankunkerkar has some ideas here as he recently bumped this parameter

sohankunkerkar: https://github.com/kubernetes-sigs/kueue/pull/10707#discussion_r3141549657

test/e2e/singlecluster/kuberay_test.go
```diff
@@ -420,7 +434,7 @@ print(ray.get([my_task.remote(i, 8) for i in range(4)]))
 print(ray.get([my_task.remote(i, 8) for i in range(16)]))
 
 # run tasks in sequence to trigger scaling down
-print([ray.get(my_task.remote(i, 1)) for i in range(32)])`,
+print([ray.get(my_task.remote(i, 1)) for i in range(64)])`,
```
sohankunkerkar, line 437: any reason to change this to 64? it works fine on the main branch.

test/e2e/singlecluster/kuberay_test.go
```diff
@@ -570,6 +581,14 @@ print([ray.get(my_task.remote(i, 1)) for i in range(32)])`,
 			}, util.MediumTimeout, util.Interval).Should(gomega.Succeed(), util.AssertMsgObjList("Did not observe >=3 workloads from multiple scale-ups", workloadList))
 		})
 
+		ginkgo.By("Waiting for workers reduced to 1 due to scaling down", func() {
```
sohankunkerkar: I don't think we need this block anymore.
xref: https://github.com/kubernetes-sigs/kueue/pull/10707#discussion_r3141553209

test/e2e/singlecluster/kuberay_test.go
hiboyang: I got random test failure due to the rayjob not staying in the scaling down state longer enough, thus changed this to make the test more stable.

test/e2e/singlecluster/kuberay_test.go
hiboyang: This step verifies there is really only 1 worker after scaling down, should still be a good check?

pkg/util/testingjobs/rayjob/wrappers.go
hiboyang: Let me switch back to `idleTimeoutSeconds := int32(60)` and run another round of test.

hiboyang: `idleTimeoutSeconds := int32(60)` caused test fail, reverted to `idleTimeoutSeconds := int32(10)` to make test pass. The reason to use 10 seconds here is the scale-down happens faster, so later scale-down verification (1 worker) can pass.

sohankunkerkar: I think you need to remove-> `Waiting for workers reduced to 1 due to scaling down` test. Otherwise, I'm afraid we will re-introduced the same flake that was closed with https://github.com/kubernetes-sigs/kueue/pull/10707
If you are going ahead with this value, then make sure to run the tests in loop for at least 200 times to check for flakiness.

hiboyang: cool, is there anyway to run the test 200 times in PR check? I started something like `GINKGO_ARGS="--repeat=200" make kind-image-build test-e2e` on my local machine, still better to have some way to run in PR check environment.

sohankunkerkar: You can do something like this: https://github.com/kubernetes-sigs/kueue/pull/9050/changes#diff-8be708e6a6046713666b607889f5a2fd5f69f817a0a2815c62b03f725f6ab34fR411

hiboyang: sounds good, thanks for the example! let me try it

mimowo: So, are we sure reverting this to 10s is a good idea here? In that case, could we also align the main branch?

mimowo:
@hiboyang I see numerous updates to the PR, could you summarize what are the actual problems which requires the manual bugfix.

hiboyang:
> @hiboyang I see numerous updates to the PR, could you summarize what are the actual problems which requires the manual bugfix.

Yes, there was RayJob autoscaling test failure previously in the step verifying 1 worker after scale-down, the root cause was RayJob idleTimeoutSeconds too long (60 seconds) which did not trigger scale-down fast enough. I reduced that to 10 seconds. Then tried to rerun test with a loop to see whether it is stable. Now reverted that loop testing. The code should be good now.

pkg/util/testingjobs/rayjob/wrappers.go
hiboyang: I did loop test with 50 times, it was good. When testing with 300 times or 100 times, it exceeded test suite timeout limit. I think the code should be good now, reverted those loop test code.

mimowo: I see, thank you, but do we stll need setting `idleTimeoutSeconds := int32(10)`?

mimowo: If so, then let's also open PR against main so that we have the same value everywhere

hiboyang: Yes, 10 seconds is better if it does not cause other issue, since it can trigger scale-down faster than previous value (60 seconds).

hiboyang: yes, will open another PR again main after this PR merged.

mimowo: I see, awesome

test/e2e/singlecluster/kuberay_test.go
mimowo: Hm, let's drop it as this wasn't added in the original PR. We may consider re-adding, but as @sohankunkerkar if scale down was broken the following assert would fail already: `ginkgo.By("Checking podset-replica-sizes annotation updated after scaling down", func() {`

mimowo: Basically I want to make sure the testing we have is consident across release branches

hiboyang: sounds good, let me remove "Waiting for workers reduced to 1 due to scaling down" step

mimowo:
/lgtm
/approve
Thank you 👍 Please also try to CP to 0.16 if feasible

hiboyang:
@mimowo this is PR in main to reduce to 10 seconds: https://github.com/kubernetes-sigs/kueue/pull/11065

PR #10973, opened 2026-05-06
Add event-driven DeviceClass tracking for extended resources

PannagaRao:

#### What type of PR is this?
/kind feature

Adds event-driven DeviceClass tracking with a DRABackedResources cache and workload index by extended resource names, ensuring only DRA-backed workloads enter the DRA path and only affected workloads are requeued on DeviceClass changes

Fixes #https://github.com/kubernetes-sigs/kueue/pull/11552

```release-note
DRA: Fix the performance of determining if a Workload should be handled using DRA Extended Resources, 
using an event-driven maintained cache of DRA Extended Resources.
```

### DIFF
--- a/pkg/cache/queue/manager.go
+++ b/pkg/cache/queue/manager.go
@@ -125,6 +125,12 @@ func WithResourceMetrics(enabled bool) Option {
 	}
 }
 
+func WithDRABackedResources(cache *dra.ExtendedResourceCache) Option {
+	return func(m *Manager) {
+		m.draBackedResources = cache
+	}
+}
+
 // SetDRAReconcileChannel sets the DRA reconcile channel after manager creation.
 func (m *Manager) SetDRAReconcileChannel(ch chan<- event.TypedGenericEvent[*kueue.Workload]) {
 	m.draReconcileChannel = ch
@@ -165,6 +171,7 @@ type Manager struct {
 	workloadUpdateWatchers []WorkloadUpdateWatcher
 
 	draReconcileChannel chan<- event.TypedGenericEvent[*kueue.Workload]
+	draBackedResources  *dra.ExtendedResourceCache
 
 	roleTracker            *roletracker.RoleTracker
 	customLabels           *metrics.CustomLabels
@@ -485,7 +492,7 @@ func (m *Manager) addLocalQueueLocked(ctx context.Context, q *kueue.LocalQueue) 
 		}
 
 		log := ctrl.LoggerFrom(ctx).WithValues("workload", klog.KObj(&w))
-		if dra.NeedsDRAReconcile(&w) {
+		if dra.NeedsDRAReconcile(&w, m.draBackedResources) {
 			// Collect DRA workloads to send outside the lock; DeepCopy keeps a
 			// stable pointer since the range variable is reused each iteration.
 			draWorkloads = append(draWorkloads, w.DeepCopy())
--- a/pkg/controller/core/indexer/indexer.go
+++ b/pkg/controller/core/indexer/indexer.go
@@ -30,6 +30,7 @@ import (
 
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
 	"sigs.k8s.io/kueue/pkg/features"
+	utilresource "sigs.k8s.io/kueue/pkg/util/resource"
 	"sigs.k8s.io/kueue/pkg/util/slices"
 )
 
@@ -44,6 +45,7 @@ const (
 	WorkloadAdmissionCheckKey            = "status.admissionChecks"
 	WorkloadPriorityClassKey             = "spec.priorityClassRef"
 	DeviceClassExtendedResourceNameIndex = "spec.extendedResourceName"
+	WorkloadExtendedResourceKey          = "spec.extendedResources"
 	// WorkloadSliceNameKey is an index for pods by their workload slice name annotation.
 	// Used to find pods belonging to an elastic workload slice chain.
 	WorkloadSliceNameKey = "metadata.workloadSliceName"
@@ -214,6 +216,32 @@ func IndexWorkloadPriorityClass(obj client.Object) []string {
 	return []string{wl.Spec.PriorityClassRef.Name}
 }
 
+// IndexWorkloadExtendedResources indexes Workloads by the extended resource names
+// in their container requests. Used by the DeviceClass handler to find workloads
+// affected by a specific DeviceClass change.
+func IndexWorkloadExtendedResources(obj client.Object) []string {
+	wl, ok := obj.(*kueue.Workload)
+	if !ok {
+		return nil
+	}
+	resNames := sets.New[string]()
+	for _, ps := range wl.Spec.PodSets {
+		for _, containers := range [][]corev1.Container{ps.Template.Spec.InitContainers, ps.Template.Spec.Containers} {
+			for _, c := range containers {
+				for name, qty := range c.Resources.Requests {
+					if !qty.IsZero() && utilresource.IsExtendedResourceName(name) {
+						resNames.Insert(string(name))
+					}
+				}
+			}
+		}
+	}
+	if resNames.Len() > 0 {
+		return resNames.UnsortedList()
+	}
+	return nil
+}
+
 // IndexDeviceClassExtendedResourceName indexes DeviceClasses by spec.extendedResourceName.
 func IndexDeviceClassExtendedResourceName(obj client.Object) []string {
 	dc, ok := obj.(*resourceapi.DeviceClass)
@@ -267,6 +295,9 @@ func Setup(ctx context.Context, indexer client.FieldIndexer) error {
 	if features.Enabled(features.KueueDRAIntegrationExtendedResource) {
 		if err := indexer.IndexField(ctx, &resourceapi.DeviceClass{}, DeviceClassExtendedResourceNameIndex, IndexDeviceClassExtendedResourceName); err != nil && !apimeta.IsNoMatchError(err) {
 			return fmt.Errorf("setting index on extendedResourceName for DeviceClass: %w", err)
+		}
+		if err := indexer.IndexField(ctx, &kueue.Workload{}, WorkloadExtendedResourceKey, IndexWorkloadExtendedResources); err != nil {
+			return fmt.Errorf("setting index on extended resources for Workload: %w", err)
 		}
 	}
 	return nil
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -27,6 +27,7 @@ import (
 	gocmp "github.com/google/go-cmp/cmp"
 	corev1 "k8s.io/api/core/v1"
 	nodev1 "k8s.io/api/node/v1"
+	resourcev1 "k8s.io/api/resource/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
@@ -147,6 +148,12 @@ func WithDRAMapper(value *dra.ResourceMapper) Option {
 	}
 }
 
+func WithDRABackedResources(value *dra.ExtendedResourceCache) Option {
+	return func(r *WorkloadReconciler) {
+		r.draBackedResources = value
+	}
+}
+
 type WorkloadUpdateWatcher interface {
 	NotifyWorkloadUpdate(oldWl, newWl *kueue.Workload)
 }
@@ -164,6 +171,7 @@ type WorkloadReconciler struct {
 	workloadRetention      *workloadRetentionConfig
 	draReconcileChannel    chan event.TypedGenericEvent[*kueue.Workload]
 	draMapper              *dra.ResourceMapper
+	draBackedResources     *dra.ExtendedResourceCache
 	admissionFSConfig      *config.AdmissionFairSharing
 	roleTracker            *roletracker.RoleTracker
 	preemptionExpectations *expectations.Store
@@ -302,7 +310,7 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		}
 		return ctrl.Result{}, nil
 	}
-	if workload.Status(&wl) == workload.StatusPending && dra.NeedsDRAReconcile(&wl) {
+	if workload.Status(&wl) == workload.StatusPending && dra.NeedsDRAReconcile(&wl, r.draBackedResources) {
 		workload.AdjustResources(ctx, r.client, &wl)
 		if workload.HasResourceClaim(&wl) {
 			log.V(3).Info("Workload is inadmissible because it uses resource claims which is not supported")
@@ -1135,7 +1143,7 @@ func (r *WorkloadReconciler) Create(e event.TypedCreateEvent[*kueue.Workload]) b
 	wlCopy := e.Object.DeepCopy()
 	workload.AdjustResources(ctx, r.client, wlCopy)
 
-	if dra.NeedsDRAReconcile(e.Object) {
+	if dra.NeedsDRAReconcile(e.Object, r.draBackedResources) {
 		log.V(2).Info("Skipping DRA workload in Create event - will be handled in Reconcile")
 		return true
 	}
@@ -1236,7 +1244,7 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 		})
 	case prevStatus == workload.StatusPending && status == workload.StatusPending:
 		switch {
-		case dra.NeedsDRAReconcile(e.ObjectNew):
+		case dra.NeedsDRAReconcile(e.ObjectNew, r.draBackedResources):
 			log.V(2).Info("Skipping queue update for DRA workload - handled in Reconcile")
 		default:
 			if err := r.queues.AddOrUpdateWorkload(log, wlCopy); err != nil {
@@ -1276,7 +1284,7 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 			// reconciler, we would execute it on every run, which might mess up the state.
 			if immediate {
 				switch {
-				case dra.NeedsDRAReconcile(e.ObjectNew):
+				case dra.NeedsDRAReconcile(e.ObjectNew, r.draBackedResources):
 					log.V(2).Info("Skipping immediate requeue for DRA workload - handled in Reconcile")
 				default:
 					if err := r.queues.AddOrUpdateWorkloadWithoutLock(log, wlCopy); err != nil {
@@ -1361,7 +1369,8 @@ func (r *WorkloadReconciler) SetupWithManager(mgr ctrl.Manager, cfg *config.Conf
 	ruh := &resourceUpdatesHandler{r: r}
 	wqh := &workloadQueueHandler{r: r}
 	deh := &draEventHandler{}
-	return builder.TypedControllerManagedBy[reconcile.Request](mgr).
+	dch := &deviceClassHandler{r: r}
+	bld := builder.TypedControllerManagedBy[reconcile.Request](mgr).
 		Named("workload_controller").
 		WatchesRawSource(source.TypedKind(
 			mgr.GetCache(),
@@ -1378,8 +1387,17 @@ func (r *WorkloadReconciler) SetupWithManager(mgr ctrl.Manager, cfg *config.Conf
 		Watches(&corev1.LimitRange{}, ruh).
 		Watches(&nodev1.RuntimeClass{}, ruh).
 		Watches(&kueue.ClusterQueue{}, wqh).
-		Watches(&kueue.LocalQueue{}, wqh).
-		Complete(WithLeadingManager(mgr, r, &kueue.Workload{}, cfg))
+		Watches(&kueue.LocalQueue{}, wqh)
+	if features.Enabled(features.KueueDRAIntegrationExtendedResource) {
+		if _, err := mgr.GetRESTMapper().RESTMapping(resourcev1.SchemeGroupVersion.WithKind("DeviceClass").GroupKind()); err != nil && apimeta.IsNoMatchError(err) {
+			r.logger().V(2).Info("DeviceClass API not available, skipping DeviceClass watcher")
+		} else if err != nil {
+			return err
+		} else {
+			bld = bld.Watches(&resourcev1.DeviceClass{}, dch)
+		}
+	}
+	return bld.Complete(WithLeadingManager(mgr, r, &kueue.Workload{}, cfg))
 }
 
 // admittedNotReadyWorkload returns the underlying cause and remaining time for
@@ -1472,7 +1490,7 @@ func (h *resourceUpdatesHandler) queueReconcileForPending(ctx context.Context, q
 		log.V(5).Info("Queue reconcile for")
 		workload.AdjustResources(ctrl.LoggerInto(ctx, log), h.r.client, wlCopy)
 
-		if dra.NeedsDRAReconcile(wlCopy) {
+		if dra.NeedsDRAReconcile(wlCopy, h.r.draBackedResources) {
 			req := reconcile.Request{
 				NamespacedName: types.NamespacedName{
 					Name:      wlCopy.Name,
@@ -1624,4 +1642,76 @@ func (h *draEventHandler) Generic(ctx context.Context, e event.TypedGenericEvent
 // GetDRAReconcileChannel returns the DRA reconcile channel for connecting to the queue manager.
 func (r *WorkloadReconciler) GetDRAReconcileChannel() chan<- event.TypedGenericEvent[*kueue.Workload] {
 	return r.draReconcileChannel
+}
+
+type deviceClassHandler struct {
+	r *WorkloadReconciler
+}
+
+func (h *deviceClassHandler) Create(ctx context.Context, e event.CreateEvent, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	dc := e.Object.(*resourcev1.DeviceClass)
+	if ern := extendedResourceName(dc); ern != "" {
+		h.r.draBackedResources.Add(corev1.ResourceName(ern), dc.Name)
+		h.reconcileWorkloads(ctx, q, ern)
+	}
+}
+
+func (h *deviceClassHandler) Update(ctx context.Context, e event.UpdateEvent, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	oldDC := e.ObjectOld.(*resourcev1.DeviceClass)
+	newDC := e.ObjectNew.(*resourcev1.DeviceClass)
+	if oldERN := extendedResourceName(oldDC); oldERN != "" {
+		h.r.draBackedResources.Remove(corev1.ResourceName(oldERN), oldDC.Name)
+	}
+	if newERN := extendedResourceName(newDC); newERN != "" {
+		h.r.draBackedResources.Add(corev1.ResourceName(newERN), newDC.Name)
+	}
+	h.reconcileWorkloads(ctx, q, extendedResourceName(oldDC), extendedResourceName(newDC))
+}
+
+func (h *deviceClassHandler) Delete(ctx context.Context, e event.DeleteEvent, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	dc := e.Object.(*resourcev1.DeviceClass)
+	if ern := extendedResourceName(dc); ern != "" {
+		h.r.draBackedResources.Remove(corev1.ResourceName(ern), dc.Name)
+		h.reconcileWorkloads(ctx, q, ern)
+	}
+}
+
+func (h *deviceClassHandler) Generic(_ context.Context, _ event.GenericEvent, _ workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+}
+
+func extendedResourceName(dc *resourcev1.DeviceClass) string {
+	if dc.Spec.ExtendedResourceName != nil {
+		return *dc.Spec.ExtendedResourceName
+	}
+	return ""
+}
+
+func (h *deviceClassHandler) reconcileWorkloads(ctx context.Context, q workqueue.TypedRateLimitingInterface[reconcile.Request], resourceNames ...string) {
+	log := h.r.logger()
+
+	// Requeue only workloads that request the affected extended resources.
+	for _, name := range resourceNames {
+		if name == "" {
+			continue
+		}
+		lst := kueue.WorkloadList{}
+		err := h.r.client.List(ctx, &lst,
+			client.MatchingFields{
+				indexer.WorkloadExtendedResourceKey: name,
+				indexer.WorkloadQuotaReservedKey:    string(metav1.ConditionFalse),
+			},
+		)
+		if err != nil {
+			log.Error(err, "Could not list workloads for extended resource", "resource", name)
+			continue
+		}
+		for _, w := range lst.Items {
+			q.AddAfter(reconcile.Request{
+				NamespacedName: types.NamespacedName{
+					Name:      w.Name,
+					Namespace: w.Namespace,
+				},
+			}, time.Second)
+		}
+	}
 }
--- a/pkg/dra/extended_resource_cache.go
+++ b/pkg/dra/extended_resource_cache.go
@@ -1,0 +1,70 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package dra
+
+import (
+	"sync"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// ExtendedResourceCache tracks which extended resource names are backed by DRA
+// DeviceClasses. Maps each resource name to the set of DeviceClass names that
+// declare it, so that deleting one DeviceClass doesn't remove the resource if
+// another DeviceClass still backs it.
+type ExtendedResourceCache struct {
+	lock      sync.RWMutex
+	resources map[corev1.ResourceName]sets.Set[string]
+}
+
+// NewExtendedResourceCache creates a new ExtendedResourceCache.
+func NewExtendedResourceCache() *ExtendedResourceCache {
+	return &ExtendedResourceCache{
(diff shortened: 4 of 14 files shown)


pkg/features/kube_features.go
```diff
@@ -480,6 +480,7 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 	},
 	DRAExtendedResources: {
 		{Version: version.MustParse("0.17"), Default: false, PreRelease: featuregate.Alpha},
+		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Beta},
```
kannon92: Its a little odd to me to graduate DRAExtendedResources to beta while we leave DynamicResourceAllocation in alpha.

sohankunkerkar: https://github.com/kubernetes-sigs/kueue/pull/10996

pkg/features/kube_features.go
```diff
@@ -486,6 +480,7 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
...
```
sohankunkerkar:
```suggestion
		{Version: version.MustParse("0.18"), Default: true, PreRelease: featuregate.Beta},
```

keps/2941-DRA/README.md
```diff
@@ -1008,12 +1016,25 @@ Use existing dra-example-driver or Kubernetes test driver for e2e testing.
 
 #### Beta
 
+##### DRAExtendedResources (v0.18)
+
+- the `DRAExtendedResources` feature gate in Beta
+- event-driven DeviceClass tracking for late DeviceClass creation: implemented via
```
sohankunkerkar: This criterion was moved from the DynamicResourceAllocation beta section to DRAExtendedResources beta. But the DeviceClass watcher uses `NeedsDRAReconcile` which catches both RCT  workloads (`HasDRA`) and extended resource workloads. When a DeviceClass is created, it re-reconciles workloads from both paths. This is general DRA infrastructure being pulled into an extended-resources-specific graduation. If the watcher is gated behind `DRAExtendedResources` only, RCT-only users don't benefit from event-driven DeviceClass tracking. I think it should this stay under `DynamicResourceAllocation` beta criteria instead?

pkg/controller/core/workload_controller.go
```diff
@@ -1598,3 +1603,45 @@ func (h *draEventHandler) Generic(ctx context.Context, e event.TypedGenericEvent
...
+	log := ctrl.LoggerFrom(ctx)
+	lst := kueue.WorkloadList{}
+	err := h.r.client.List(ctx, &lst, client.MatchingFields{indexer.WorkloadQuotaReservedKey: string(metav1.ConditionFalse)})
+	if err != nil {
+		log.Error(err, "Could not list pending workloads for DeviceClass event")
+		return
+	}
+	var count int
+	for _, w := range lst.Items {
+		if dra.NeedsDRAReconcile(&w) {
```
sohankunkerkar: nit: There's no startup validation that `DRAExtendedResources` requires `DynamicResourceAllocation`. If both default to true at beta this is less of a concern, but for correctness consider adding the dependency in `config/validation.go` (same pattern as `ElasticJobsViaWorkloadSlicesWithTAS`).

test/integration/singlecluster/controller/dra/dra_test.go
```diff
@@ -1046,4 +1034,248 @@ var _ = ginkgo.Describe("DRA Integration", ginkgo.Ordered, ginkgo.ContinueOnFail
...
+			gomega.Eventually(func(g gomega.Gomega) {
+				var updatedWl kueue.Workload
+				g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(wl), &updatedWl)).To(gomega.Succeed())
+				g.Expect(workload.HasQuotaReservation(&updatedWl)).To(gomega.BeTrue())
+				g.Expect(updatedWl.Status.Admission).NotTo(gomega.BeNil())
+				g.Expect(updatedWl.Status.Admission.PodSetAssignments).To(gomega.HaveLen(1))
+
+				assignment := updatedWl.Status.Admission.PodSetAssignments[0]
+				g.Expect(assignment.ResourceUsage).To(gomega.HaveKey(corev1.ResourceName(logicalName)))
+			}, util.Timeout, util.Interval).Should(gomega.Succeed())
```
sohankunkerkar, line 1359: Have you tried using `util.ShortInterval` instead Interval? did you see any deviations?

pkg/controller/core/workload_controller.go
PannagaRao: There is validation present https://github.com/kubernetes-sigs/kueue/blob/main/pkg/config/validation.go#L563. Is this what you were referring or am I missing something here?

test/integration/singlecluster/controller/dra/dra_test.go
PannagaRao: ShortInterval works fine. Updated it

kannon92:
@sohankunkerkar @PannagaRao 

should we close https://github.com/kubernetes-sigs/kueue/pull/10996?

It seems you carried that commit into this PR.

PannagaRao:
> @sohankunkerkar @PannagaRao
> 
> should we close #10996?
> 
> It seems you carried that commit into this PR.

Maybe I could add #10996 as a dependency for this PR.  Just to make sure #10996 isn't blocked coz of the changes in this PR.

sohankunkerkar:
JFYI: https://github.com/kubernetes-sigs/kueue/pull/10996#issuecomment-4427350328

keps/2941-DRA/README.md
```diff
@@ -1030,13 +1036,19 @@ Use existing dra-example-driver or Kubernetes test driver for e2e testing.
...
-- re-evaluate post-scheduling quota reconciliation for DeviceClass drift
+- feature gate enabled by default
+- re-evaluated post-scheduling quota reconciliation for DeviceClass drift: not applicable
+  for extended resources as the quota key is `extendedResourceName` itself, making
+  DeviceClass-level drift invisible to quota accounting
 - re-evaluate the need for indexing of resourceSlices for CEL performance lookups
 - re-evaluate event-driven ResourceSlice tracking for late ResourceSlice creation
+- integration and e2e tests for DeviceClass lifecycle scenarios
 
 #### GA
```
sohankunkerkar, line 1541: Can you add some GA criteria here like:
  - the feature gate in stable                                                                                                                                               
  - user adoption feedback confirms stability                                                                                                                                
  - re-evaluate DeviceClass watcher performance at scale

pkg/cache/queue/cluster_queue.go
```diff
@@ -363,6 +363,14 @@ func (c *ClusterQueue) PushOrUpdate(wInfo *workload.Info) {
 		}
 		// Workload is leaving inadmissible; account for its resources before moving.
 		c.removeFromInadmissible(key, oldInfo)
+		// DRA translation changes TotalRequests without changing the pod spec,
```
sohankunkerkar: Good catch! This is important now that `SchedulingEquivalenceHashing` is default-true in 0.18. Without this, a DRA workload that gets NoFit before DeviceClass creation would be permanently stuck and the equivalence hash persists in noFitSchedulingHashes and sends it straight back to inadmissible even after DRA translation changes TotalRequests.

pkg/controller/core/workload_controller.go
```diff
@@ -1344,8 +1353,13 @@ func (r *WorkloadReconciler) SetupWithManager(mgr ctrl.Manager, cfg *config.Conf
 		Watches(&corev1.LimitRange{}, ruh).
 		Watches(&nodev1.RuntimeClass{}, ruh).
 		Watches(&kueue.ClusterQueue{}, wqh).
-		Watches(&kueue.LocalQueue{}, wqh).
-		Complete(WithLeadingManager(mgr, r, &kueue.Workload{}, cfg))
+		Watches(&kueue.LocalQueue{}, wqh)
+	if features.Enabled(features.KueueDRAIntegrationExtendedResource) {
+		if _, err := mgr.GetRESTMapper().RESTMapping(resourcev1.SchemeGroupVersion.WithKind("DeviceClass").GroupKind()); err == nil {
```
sohankunkerkar: The indexer fix in this same PR  uses `!apimeta.IsNoMatchError(err)` to distinguish "API doesn't exist" from "API server error." Here it silently disables the DeviceClass watcher for any error, including transient API failures. Could we use the same `IsNoMatchError` pattern for consistency?

keps/2941-DRA/README.md
PannagaRao: Added

pkg/controller/core/workload_controller.go
PannagaRao: Done

pkg/cache/queue/cluster_queue.go
mimowo: Hmm, but the fix looks suspicious. I have some questions: 
1, When we see `draRequestsChanged`, do we have a new schedulingHash basically, so that we could compute `UpdateSchedulingHash`?
2. Why not rely on updating the scheduling hash in some already existing code paths, I'm not clear what makes DRA special here

pkg/cache/queue/manager.go
```diff
@@ -485,7 +485,11 @@ func (m *Manager) addLocalQueueLocked(ctx context.Context, q *kueue.LocalQueue)
 		}
 
 		log := ctrl.LoggerFrom(ctx).WithValues("workload", klog.KObj(&w))
-		if dra.NeedsDRAReconcile(&w) {
+		// Only defer DRA workloads to the reconcile channel if the channel
+		// is set up. Without this check, workloads would be silently dropped
+		// from the queue when the channel is not configured (e.g., in unit
+		// tests that don't run the workload controller).
+		if dra.NeedsDRAReconcile(&w) && m.draReconcileChannel != nil {
```
mimowo: I would rather register the channel in unit tests rather than add the non-trivial skip code. It is rally tricky to evaluate correctness of the skip.

pkg/controller/core/workload_controller.go
```diff
@@ -368,21 +369,28 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
 		}
 
-		quotaReservedCond := apimeta.FindStatusCondition(wl.Status.Conditions, kueue.WorkloadQuotaReserved)
-		requeuedCond := apimeta.FindStatusCondition(wl.Status.Conditions, kueue.WorkloadRequeued)
+		// Only clear stale conditions when DRA processing actually resolved
+		// resources. Workloads that enter this block due to domain-qualified
+		// resource names but have no matching DeviceClass produce empty
+		// draResources and should not have their conditions cleared, as
+		// downstream requeue logic needs Requeued=False to set Requeued=True.
+		if len(draResources) > 0 {
```
mimowo: Is this a bugfix? If so could we make it as a separate PR to make that clear? 

Cc @sohankunkerkar

pkg/controller/core/indexer/indexer.go
```diff
@@ -268,8 +268,9 @@ func Setup(ctx context.Context, indexer client.FieldIndexer) error {
 		}
 	}
 	// Index DeviceClasses by extendedResourceName for fast lookup during extended resource translation.
+	// The index is skipped if the DeviceClass API is not available (e.g., k8s < 1.35).
 	if features.Enabled(features.KueueDRAIntegrationExtendedResource) {
-		if err := indexer.IndexField(ctx, &resourceapi.DeviceClass{}, DeviceClassExtendedResourceNameIndex, IndexDeviceClassExtendedResourceName); err != nil {
+		if err := indexer.IndexField(ctx, &resourceapi.DeviceClass{}, DeviceClassExtendedResourceNameIndex, IndexDeviceClassExtendedResourceName); err != nil && !apimeta.IsNoMatchError(err) {
```
mimowo: Please submit this as a separate bugfix PR

pkg/controller/core/workload_controller.go
```diff
@@ -1619,3 +1637,45 @@ func (h *draEventHandler) Generic(ctx context.Context, e event.TypedGenericEvent
...
+}
+
+func (h *deviceClassHandler) queueReconcileForDRAWorkloads(ctx context.Context, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	log := ctrl.LoggerFrom(ctx)
+	lst := kueue.WorkloadList{}
+	err := h.r.client.List(ctx, &lst, client.MatchingFields{indexer.WorkloadQuotaReservedKey: string(metav1.ConditionFalse)})
+	if err != nil {
+		log.Error(err, "Could not list pending workloads for DeviceClass event")
+		return
+	}
```
mimowo: Listing all workloads unconditionally (or just by the generic key) seems quite excessive, especially if we are making this feature Beta. I think we should at least restrict that to relevant changes in the DeviceClass. Is there a way to find quickly the workloads which are affected by deviceClass?

I think we could make a more narrowing indexer basically.

mimowo: IIUC the issue is that by looking at the workload status we cannot tell if it was using DRA, right? I"m wondering maybe we should have an annotation or status field in the status.Admission for the fast lookup of the affected workloads.

pkg/cache/queue/cluster_queue.go
sohankunkerkar: >1, When we see draRequestsChanged, do we have a new schedulingHash basically, so that we could compute UpdateSchedulingHash?

IIUC, `UpdateSchedulingHash` would produce the same hash. It's computed from pod spec, count, and priority, and DRA translation doesn't change any of those. It modifies `TotalRequests` on workload.Info, which isn't part of the hash.

> Why not rely on updating the scheduling hash in some already existing code paths, I'm not clear what makes DRA special here

That's what makes DRA different from other state changes as spec edits and priority updates naturally change the hash because they modify what the hash is computed from. DRA changes what the workload actually needs for quota without touching the spec. So clearing the hash here is the way to force re-evaluation.

mimowo: Hm, what about then including `TotalRequests` in the hash computation? IIUC this would also fix the issue, and keep the model consistent.

pkg/controller/core/workload_controller.go
sohankunkerkar: This isn't a pre-existing bug. It only surfaces when the gate is default-true because `NeedsDRAReconcile` starts matching any workload with domain-qualified resources. In alpha (gate off by default), only explicit DRA workloads hit this path. Thoughts?

mimowo: Does this mean that wit this change we run the DRA logic for any users using nvidia.com/gpu, or google.com/tpu, even if they don't actually use DRA? 

I think this might be pretty risky, and I'm wondering how we could mitigate. One approach could be to quickly excape the DRA Reconcile if we confirm the extended resources aren't used.

Another approach I'm thinking about is to have some cache for knowing if a resource is baked by DRA or not.

pkg/controller/core/workload_controller.go
sohankunkerkar: Agreed, the full workload scan is too broad. We can follow the RuntimeClass handler pattern i.e. add a workload index keyed by requested extended resource names, then on DeviceClass events only requeue workloads that actually request the affected resource. For updates we'd need to check both old and new `extendedResourceName` since the name itself can change.

pkg/cache/queue/cluster_queue.go
sohankunkerkar: Yeah,I agreed! Including `TotalRequests` in the hash works. Resource transformations and excluded prefixes are cluster config so they won't cause hash divergence between same-spec workloads and the only realistic case where TotalRequests differs is DRA translation, which is exactly the case we need to handle.

pkg/controller/core/workload_controller.go
sohankunkerkar: Good call! I think cache is the cleanest approach as it resolves both channel routing and DeviceClass handler narrowing problems.

pkg/cache/queue/manager.go
PannagaRao: Addressed by the DRABackedResources cache. NeedsDRAReconcile now only matches workloads whose extended resources are backed by an actual DeviceClass. In unit tests (no DeviceClasses), workloads with   example.com/gpu go through normal processing. The nil channel skip has been removed.

pkg/controller/core/workload_controller.go
PannagaRao: Added the cache implementation. NeedsDRAReconcile now checks the cache (populated from DeviceClass objects) instead of the broad IsExtendedResourceName format check. Non-DRA workloads with domain-qualified resources no longer enter the DRA reconcile path — they go through normal scheduling. Only workloads whose extended resources match an actual DeviceClass with extendedResourceName are processed through DRA.

pkg/controller/core/workload_controller.go
PannagaRao: Added a workload index keyed by extended resource names. The DeviceClass handler now extracts the extendedResourceName from the event and queries only workloads requesting that specific resource, following the RuntimeClass handler pattern. PTAL!

pkg/controller/core/indexer/indexer.go
PannagaRao: https://github.com/kubernetes-sigs/kueue/pull/11405

pkg/cache/queue/cluster_queue.go
sohankunkerkar: https://github.com/kubernetes-sigs/kueue/pull/11399

mimowo:
@PannagaRao @sohankunkerkar please de-couple the ER mechanism changes from the actual graduation. As I propose here: https://github.com/kubernetes-sigs/kueue/pull/10996#issuecomment-4519109275 we may consider staying with the ER in Alpha for 0.18, at the same time we would like to merge the mechanism improvements.

Staying in Alpha for 0.18 I consider is still not determined, but I want to basically de-couple the two decisions.

tenzen-y (commented):
I'm with @mimowo based on his https://github.com/kubernetes-sigs/kueue/pull/10996#issuecomment-4519109275 comment and discussions in this PR.

The previous impls (just confirming resource name) were very aggressive which is not align with today's kube-scheduler ER support. Today's kube-scheduler properly confirms if the detected ER is truly backed by the DRA driver. OTOH, the previous this PR impls don't have that.

Surely, the latest impls properly check if the detected ER name is tied to the DRA driver.
But, this is completely new way which should not be coupled with Beta graduation.

And also the current impls are premature. We should mature the detailed logic before Beta graduation.

So, I'd like to consider this new mechanism as a second alpha for ER, then exclude ER support from Kueue DRA integration Beta criteria.

PannagaRao:
@mimowo  @tenzen-y Thanks for the feedback. I've decoupled the two — this PR now contains only the mechanism improvements and the  feature gate stays at Alpha (default false).

  - DeviceClass watcher on the workload controller for create/update/delete events
  - DRABackedResources cache — ensures workloads with domain-qualified resources that aren't DRA-backed (no corresponding DeviceClass) skip DRA processing entirely
  - Workload index by extended resource names — targeted requeue of only affected workloads on DeviceClass changes, instead of listing all pending workloads

mimowo:
cc @sohankunkerkar ptal

pkg/controller/core/workload_controller.go
```diff
@@ -1595,3 +1652,69 @@ func (h *draEventHandler) Generic(ctx context.Context, e event.TypedGenericEvent
...
+	dra.DRABackedResources().Sync(dcList.Items)
+
+	// Requeue only workloads that request the affected extended resources.
+	for _, name := range resourceNames {
+		if name == "" {
+			continue
+		}
+		lst := kueue.WorkloadList{}
+		err := h.r.client.List(ctx, &lst,
+			client.MatchingFields{indexer.WorkloadExtendedResourceKey: name},
```
sohankunkerkar: I suspect the second `MatchingFields` might overwrites the first. `MatchingFields.ApplyToList()` does a direct assignment (`opts.FieldSelector = sel`), not a merge, unlike `MatchingLabels` which ANDs requirements together. So only the `QuotaReserved=False` filter applies; the extended resource name filter will silently drop here.

sohankunkerkar (commented):
Overall this looks good to me. I have one comment.

pkg/controller/core/workload_controller.go
PannagaRao: I think you are right. I combined both fields into a single MatchingFields map. Thoughts?

pkg/dra/extended_resources.go
```diff
@@ -38,9 +38,6 @@ import (
 // Fast in-memory check with no API calls. Uses a broad format-based filter
```
sohankunkerkar: The godoc still says "may over-trigger for non-DRA extended resources" but with the `DRABackedResources().Has(name)`  check, it no longer over-triggers. The comment should reflect the new behavior.

PannagaRao: Fixed

pkg/controller/core/workload_controller.go
sohankunkerkar: this looks better now.

sohankunkerkar (approved):
/lgtm

pkg/controller/core/indexer/indexer.go
```diff
@@ -268,6 +296,9 @@ func Setup(ctx context.Context, indexer client.FieldIndexer) error {
 		if err := indexer.IndexField(ctx, &resourceapi.DeviceClass{}, DeviceClassExtendedResourceNameIndex, IndexDeviceClassExtendedResourceName); err != nil && !apimeta.IsNoMatchError(err) {
 			return fmt.Errorf("setting index on extendedResourceName for DeviceClass: %w", err)
 		}
+		if err := indexer.IndexField(ctx, &kueue.Workload{}, WorkloadExtendedResourceKey, IndexWorkloadExtendedResources); err != nil {
+			return fmt.Errorf("setting index on extended resources for Workload: %w", err)
+		}
```
mimowo, line 301: Wrap the indexing under the FG

pkg/controller/core/workload_controller.go
```diff
@@ -1348,8 +1350,17 @@ func (r *WorkloadReconciler) SetupWithManager(mgr ctrl.Manager, cfg *config.Conf
 		Watches(&corev1.LimitRange{}, ruh).
 		Watches(&nodev1.RuntimeClass{}, ruh).
 		Watches(&kueue.ClusterQueue{}, wqh).
-		Watches(&kueue.LocalQueue{}, wqh).
-		Complete(WithLeadingManager(mgr, r, &kueue.Workload{}, cfg))
+		Watches(&kueue.LocalQueue{}, wqh)
+	if features.Enabled(features.KueueDRAIntegrationExtendedResource) {
+		if _, err := mgr.GetRESTMapper().RESTMapping(resourcev1.SchemeGroupVersion.WithKind("DeviceClass").GroupKind()); err != nil && apimeta.IsNoMatchError(err) {
+			ctrl.Log.V(2).Info("DeviceClass API not available, skipping DeviceClass watcher")
```
mimowo: We should use roleTracker for all logging

pkg/controller/core/workload_controller.go
```diff
@@ -1595,3 +1606,71 @@ func (h *draEventHandler) Generic(ctx context.Context, e event.TypedGenericEvent
...
+				indexer.WorkloadExtendedResourceKey: name,
+				indexer.WorkloadQuotaReservedKey:    string(metav1.ConditionFalse),
+			},
+		)
+		if err != nil {
+			log.Error(err, "Could not list workloads for extended resource", "resource", name)
+			continue
+		}
+		for _, w := range lst.Items {
+			q.Add(reconcile.Request{
```
mimowo: Use AddAfter with 1s delay

pkg/controller/core/workload_controller.go
```diff
@@ -1595,3 +1606,71 @@ func (h *draEventHandler) Generic(ctx context.Context, e event.TypedGenericEvent
...
+func (h *deviceClassHandler) syncCacheAndReconcileWorkloads(ctx context.Context, q workqueue.TypedRateLimitingInterface[reconcile.Request], resourceNames ...string) {
+	log := ctrl.LoggerFrom(ctx)
+
+	// Rebuild the DRA-backed resource cache from current DeviceClasses.
+	var dcList resourcev1.DeviceClassList
+	if err := h.r.client.List(ctx, &dcList); err != nil {
+		log.Error(err, "Could not list DeviceClasses to rebuild cache")
+		return
+	}
+	dra.DRABackedResources().Sync(dcList.Items)
```
mimowo: I don't understand, why calling Sync based on the entire list. we have the modified DeviceClass at hand in the update.

pkg/controller/core/indexer/indexer.go
sohankunkerkar: The workload index registration is already inside the `KueueDRAIntegrationExtendedResource` gate block. Both the DeviceClass index and the workload index  are gated together. Is there a different scope you had in mind?

pkg/controller/core/workload_controller.go
sohankunkerkar: IIUC, K8s permits multiple DeviceClasses with the same extendedResourceName. On Delete, incremental removal could incorrectly drop the name if another DeviceClass still has it. The full rebuild avoids this.

pkg/dra/extended_resource_cache.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+var globalERCache = &extendedResourceCache{
+	resources: sets.New[corev1.ResourceName](),
+}
+
+// DRABackedResources returns the global extended resource cache.
+func DRABackedResources() *extendedResourceCache {
+	return globalERCache
+}
```
mimowo: I think this global state proliferation is getting unconfortably large for DRA, we need to try get rid of it, either this PR or follow ups, opened: https://github.com/kubernetes-sigs/kueue/issues/11660

pkg/controller/core/workload_controller.go
```diff
@@ -1595,3 +1606,71 @@ func (h *draEventHandler) Generic(ctx context.Context, e event.TypedGenericEvent
...
+		if name == "" {
+			continue
+		}
+		lst := kueue.WorkloadList{}
+		err := h.r.client.List(ctx, &lst,
+			client.MatchingFields{
+				indexer.WorkloadExtendedResourceKey: name,
+				indexer.WorkloadQuotaReservedKey:    string(metav1.ConditionFalse),
+			},
+		)
```
mimowo, line 1703: I think this feels like should be part of the manager cache, where the not reserving workloads are placed . This could be part of the follow up to get rid of the global state.

pkg/controller/core/workload_controller.go
PannagaRao: Fixed — using r.logger() which wraps with roleTracker

pkg/controller/core/indexer/indexer.go
```diff
@@ -214,6 +216,32 @@ func IndexWorkloadPriorityClass(obj client.Object) []string {
...
 
+// IndexWorkloadExtendedResources indexes Workloads by the extended resource names
+// in their container requests. Used by the DeviceClass handler to find workloads
+// affected by a specific DeviceClass change.
+func IndexWorkloadExtendedResources(obj client.Object) []string {
+	wl, ok := obj.(*kueue.Workload)
+	if !ok {
+		return nil
+	}
+	set := sets.New[string]()
```
tenzen-y:
```suggestion
	resNames := sets.New[string]()
```

nit: We should avoid using a typed name as a variable name

pkg/dra/extended_resource_cache.go
PannagaRao: Removed the global cache approach. Using dependency injection instead.

pkg/controller/core/workload_controller.go
```diff
@@ -1625,3 +1643,71 @@ func (h *draEventHandler) Generic(ctx context.Context, e event.TypedGenericEvent
...
+
+func extendedResourceName(dc *resourcev1.DeviceClass) string {
+	if dc.Spec.ExtendedResourceName != nil {
+		return *dc.Spec.ExtendedResourceName
+	}
+	return ""
+}
+
+func (h *deviceClassHandler) syncCacheAndReconcileWorkloads(ctx context.Context, q workqueue.TypedRateLimitingInterface[reconcile.Request], resourceNames ...string) {
+	log := ctrl.LoggerFrom(ctx)
```
mimowo: make sure the roleTracker is injected into the logger

pkg/controller/core/workload_controller.go
mimowo: Hm, Yes, so this is the reason why DRABackedResources would not be a simple set, but it could be a map: ResourceName -> set(DeviceClass), then removing a deviceClass would remove an element from the set, witbout empting the map.

test/integration/singlecluster/controller/dra/dra_test.go
```diff
@@ -1253,4 +1253,248 @@ var _ = ginkgo.Describe("DRA Integration", ginkgo.Ordered, ginkgo.ContinueOnFail
...
+				Queue("dc-tracking-lq").
+				Request(corev1.ResourceName(extendedResourceName), "2").
+				Obj()
+			gomega.Expect(k8sClient.Create(ctx, wl2)).To(gomega.Succeed())
+
+			gomega.Consistently(func(g gomega.Gomega) {
+				var updatedWl kueue.Workload
+				g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(wl2), &updatedWl)).To(gomega.Succeed())
+				g.Expect(workload.HasQuotaReservation(&updatedWl)).To(gomega.BeFalse())
+			}, util.ConsistentDuration, util.ShortInterval).Should(gomega.Succeed())
```
mimowo: nit. but instead of using consistently could you check the kueue_pending_workloads metric? This way we could save on empty waiting.

test/e2e/dra/baseline/dra_test.go
```diff
@@ -709,4 +709,102 @@ var _ = ginkgo.Describe("DRA", func() {
...
+				Name:      workloadjob.GetWorkloadNameForJob(job.Name, job.UID),
+				Namespace: ns.Name,
+			}
+
+			ginkgo.By("Verifying job remains suspended (no DeviceClass, no translation)")
+			createdJob := &batchv1.Job{}
+			gomega.Consistently(func(g gomega.Gomega) {
+				g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(job), createdJob)).To(gomega.Succeed())
+				g.Expect(createdJob.Spec.Suspend).To(gomega.Equal(new(true)))
+			}, util.ConsistentDuration, util.ShortInterval).Should(gomega.Succeed())
```
mimowo: probably using ShortConsistentDuration would be enougb, wdyt?

test/integration/singlecluster/controller/dra/dra_test.go
```diff
@@ -1253,4 +1253,248 @@ var _ = ginkgo.Describe("DRA Integration", ginkgo.Ordered, ginkgo.ContinueOnFail
...
+				Request(corev1.ResourceName(extendedResourceName), "2").
+				Obj()
+			gomega.Expect(k8sClient.Create(ctx, wl)).To(gomega.Succeed())
+
+			ginkgo.By("Verifying workload is not admitted (no DeviceClass, no translation)")
+			gomega.Consistently(func(g gomega.Gomega) {
+				var updatedWl kueue.Workload
+				g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(wl), &updatedWl)).To(gomega.Succeed())
+				g.Expect(workload.HasQuotaReservation(&updatedWl)).To(gomega.BeFalse())
+			}, util.ConsistentDuration, util.ShortInterval).Should(gomega.Succeed())
```
mimowo: let's check the metric to make the test faster

test/integration/singlecluster/controller/dra/dra_test.go
```diff
@@ -1253,4 +1253,248 @@ var _ = ginkgo.Describe("DRA Integration", ginkgo.Ordered, ginkgo.ContinueOnFail
...
+				Queue("dc-tracking-lq").
+				Request(corev1.ResourceName(newExtendedResourceName), "2").
+				Obj()
+			gomega.Expect(k8sClient.Create(ctx, wl)).To(gomega.Succeed())
+
+			ginkgo.By("Verifying workload is not admitted")
+			gomega.Consistently(func(g gomega.Gomega) {
+				var updatedWl kueue.Workload
+				g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(wl), &updatedWl)).To(gomega.Succeed())
+				g.Expect(workload.HasQuotaReservation(&updatedWl)).To(gomega.BeFalse())
```
mimowo: use the metric for pending workloads

test/integration/singlecluster/controller/dra/dra_test.go
```diff
@@ -1253,4 +1253,248 @@ var _ = ginkgo.Describe("DRA Integration", ginkgo.Ordered, ginkgo.ContinueOnFail
...
+				Request(corev1.ResourceName(extendedResourceName), "2").
+				Obj()
+			gomega.Expect(k8sClient.Create(ctx, wl)).To(gomega.Succeed())
+
+			ginkgo.By("Verifying workload is not admitted")
+			gomega.Consistently(func(g gomega.Gomega) {
+				var updatedWl kueue.Workload
+				g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(wl), &updatedWl)).To(gomega.Succeed())
+				g.Expect(workload.HasQuotaReservation(&updatedWl)).To(gomega.BeFalse())
+			}, util.ConsistentDuration, util.ShortInterval).Should(gomega.Succeed())
```
mimowo: use the metric for pending workloads

pkg/controller/core/indexer/indexer.go
mimowo: > 
The workload index registration is already inside the KueueDRAIntegrationExtendedResource gate

I missed that

>  Is there a different scope you had in mind?

no, the code lgtm

mimowo (commented):
LGTM, I left only small comments (I think), thank you for the PR 👍

pkg/controller/core/workload_controller.go
PannagaRao: Done

test/integration/singlecluster/controller/dra/dra_test.go
PannagaRao: Makes sense. Switched to use the metric instead

mimowo:
/test pull-kueue-test-e2e-extended-main-1-36

that is unrelated https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10973/pull-kueue-test-e2e-extended-main-1-36/2059712946472226816

I will open issue

pkg/controller/core/workload_controller.go
PannagaRao: Done. Using map instead of set

PannagaRao:
> LGTM, I left only small comments (I think), thank you for the PR 👍

I believe to have addressed all of them. PTAL! Thanks!

mimowo (commented):
/lgtm
/approve
Thank you, this looks pretty well 👍

mimowo:
/remove-kind cleanup
/kind bug
I think the previous design was actually a bug

mimowo:
/release-note-edit
```release-note
DRA: Fix the performance of determining if a Workload should be handled using DRA Extended Resources, 
using an event-driven maintained cache of DRA Extended Resources.
```

PR #10989, opened 2026-05-06
WIP: LeaderWorkerSet: Fix nodeSelector mutation.

### LINKED ISSUE #10178: Allow to mutate nodeSelector for LWS

mbobrovskyi:
#### What type of PR is this?
/kind bug

LeaderWorkerSet: Fix nodeSelector mutation.

Fixes #10178

```release-note
NONE
```

### DIFF
--- a/pkg/controller/jobs/leaderworkerset/leaderworkerset_reconciler.go
+++ b/pkg/controller/jobs/leaderworkerset/leaderworkerset_reconciler.go
@@ -77,6 +77,11 @@ type workloadToCreate struct {
 	index int
 }
 
+type workloadToUpdate struct {
+	wl    *kueue.Workload
+	index int
+}
+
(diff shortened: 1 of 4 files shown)


mbobrovskyi:
Close for now. Need to find another solution how to fix nodeSelector mutation.

PR #10996, opened 2026-05-06
dra: graduate KueueDRAIntegration to beta

### LINKED ISSUE #8243: Tracking DRA Support for Beta
## Context

  DRA support in Kueue is currently Alpha, supporting ResourceClaimTemplates with `AllocationMode=ExactCount` and DeviceClass mapping.

  This issue is to discuss what's needed for Beta, aligned with upstream Kubernetes DRA development.

  KEP: [keps/2941-DRA](../keps/2941-DRA/README.md)

  ---

  ## Gaps in Current Implementation

  DRA Structured Parameters is GA in Kubernetes 1.34.

  ### Affects Quota (Needs Implementation)

  | Capability | Description |
  |------------|-------------|
  | AllocationMode=All | Request all devices on a node |
  | Direct ResourceClaims | Reference pre-created claims |
  | CEL Selectors | Filter devices by attributes (only with AllocationMode=All) |

  ### Does Not Affect Quota (Remove Rejection)

  | Capability | Description |
  |------------|-------------|
  | Device Constraints | Co-location requirements - pass through to scheduler |
  | Device Config | Driver configuration - pass through to scheduler |

  ---

  ## Upstream Features to Track

  | Feature | Kubernetes Status | Feature Gate |
  |---------|-------------------|--------------|
  | FirstAvailable | Beta | `DRAPrioritizedList` |
  | Extended Resources | Alpha | `DRAExtendedResource` |
  | Partitionable Devices | Alpha | `DRAPartitionableDevices` |
  | Consumable Capacity | Alpha | `DRAConsumableCapacity` |
  | Admin Access | Alpha | `DRAAdminAccess` |
  | Device Taints | Alpha | `DRADeviceTaints` |

  ---

  ## Discussion

  1. Which capabilities are required for Kueue DRA Beta?
  2. CEL Selectors: conservative counting or CEL evaluation?
  3. Should we include k8s Beta features (FirstAvailable)?
  4. Any Alpha features we should wait for?

sohankunkerkar:
Promote the KueueDRAIntegration feature gate from alpha to beta, enabling it by default in v0.18. This covers ResourceClaimTemplate-based DRA quota accounting with deviceClassMappings.

#### What type of PR is this?
/kind feature

Fixes https://github.com/kubernetes-sigs/kueue/issues/8243

Before we consider this PR, we need to land the Partitionable Devices feature first.

```release-note
DRA: Promote KueueDRAIntegration feature gate to beta
```

### DIFF
--- a/keps/2941-DRA/README.md
+++ b/keps/2941-DRA/README.md
@@ -64,8 +64,13 @@ tags, and then generate with `hack/update-toc.sh`.
     - [E2E Test](#e2e-test)
   - [Graduation Criteria](#graduation-criteria)
     - [Alpha](#alpha)
+      - [KueueDRAIntegration (v0.14)](#kueuedraintegration-v014)
+      - [KueueDRAIntegrationExtendedResource (v0.17)](#kueuedraintegrationextendedresource-v017)
     - [Beta](#beta)
+      - [KueueDRAIntegration (v0.18)](#kueuedraintegration-v018)
+      - [KueueDRAIntegrationExtendedResource](#kueuedraintegrationextendedresource)
     - [GA](#ga)
+      - [KueueDRAIntegration](#kueuedraintegration)
 - [Implementation History](#implementation-history)
 - [Drawbacks](#drawbacks)
 - [Alternatives](#alternatives)
@@ -1004,24 +1009,38 @@ Use existing dra-example-driver or Kubernetes test driver for e2e testing.
 
 #### Alpha
 
-- the implementation behind the feature gate flag in alpha
+##### KueueDRAIntegration (v0.14)
+
+- ResourceClaimTemplate-based DRA quota accounting
 - support v1 API of DRA in core k8s
 - initial e2e tests for baseline scenario
-- support for Extended Resources (alpha in k8s 1.35)
+
+##### KueueDRAIntegrationExtendedResource (v0.17)
+
+- DeviceClass auto-discovery via field indexer on `extendedResourceName`
+- extended resource detection and resource translation
+- double-counting prevention with `deviceClassMappings`
 
 #### Beta
 
-- the feature gate in Beta
-- all known bugs are fixed
+##### KueueDRAIntegration (v0.18)
+
+- feature gate enabled by default
 - support integration with MultiKueue
 - e2e tests
-- TAS + DRA testing and support as a graduation requirement
+- CEL expression validation against ResourceSlice devices
+
+##### KueueDRAIntegrationExtendedResource
+
+- user adoption feedback
 - re-evaluate event-driven DeviceClass tracking for late DeviceClass creation
 - re-evaluate post-scheduling quota reconciliation for DeviceClass drift
 - re-evaluate the need for indexing of resourceSlices for CEL performance lookups
 - re-evaluate event-driven ResourceSlice tracking for late ResourceSlice creation
 
 #### GA
+
+##### KueueDRAIntegration
 
 - the feature gate in stable
 - integration with TopologyAwareScheduling
@@ -1037,6 +1056,7 @@ Use existing dra-example-driver or Kubernetes test driver for e2e testing.
 - Integration with Admission Fair Sharing: April 2026 — added integration tests and documentation
   confirming DRA logical resources work with existing `AdmissionFairSharing.ResourceWeights`
 - CEL expression validation support added: April 2026 by @kannon92
+- Promoted KueueDRAIntegration to Beta: May 2026 by @sohankunkerkar
 
 **Key Design Evolution:**
 - **Original Design**: Standalone DynamicResourceAllocationConfig CRD with runtime ambiguity resolution
--- a/keps/2941-DRA/kep.yaml
+++ b/keps/2941-DRA/kep.yaml
@@ -19,17 +19,17 @@ replaces:
   - "https://github.com/kubernetes-sigs/kueue/pull/3071"
 
 # The target maturity stage in the current dev cycle for this KEP.
-stage: alpha
+stage: beta
 
 # The most recent milestone for which work toward delivery of this KEP has been
 # done. This can be the current (upcoming) milestone, if it is being actively
 # worked on.
-latest-milestone: "v0.17"
+latest-milestone: "v0.18"
 
 # The milestone at which this feature was, or is targeted to be, at each stage.
 milestone:
-  alpha: "v0.13"
-  beta:
+  alpha: "v0.14"
+  beta: "v0.18"
   stable:
 
 # The following PRR answers are required at alpha release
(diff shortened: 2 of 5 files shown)


keps/2941-DRA/kep.yaml
```diff
@@ -19,17 +19,17 @@ replaces:
...
 
 # The most recent milestone for which work toward delivery of this KEP has been
 # done. This can be the current (upcoming) milestone, if it is being actively
 # worked on.
-latest-milestone: "v0.17"
+latest-milestone: "v0.18"
 
 # The milestone at which this feature was, or is targeted to be, at each stage.
 milestone:
   alpha: "v0.13"
```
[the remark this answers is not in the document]
kannon92: Good bot. I guess this should be updated.

kannon92:
> Before we consider this PR, we need to land the Partitionable Devices feature first.

I don't under this. PartionableDevices should be a separate feature gate and it will be alpha in 0.18.

Why block the beta promotion on an alpha feature?

sohankunkerkar:
> > Before we consider this PR, we need to land the Partitionable Devices feature first.
> 
> I don't under this. PartionableDevices should be a separate feature gate and it will be alpha in 0.18.
> 
> Why block the beta promotion on an alpha feature?

Actually, we called that out as one of the dependencies [here](https://github.com/kubernetes-sigs/kueue/issues/8243) for the beta promotion. Having said that, this is not a hard requirement but good to have.

sohankunkerkar:
@mimowo @tenzen-y could you please take a look at it?

keps/2941-DRA/README.md
```diff
@@ -1001,24 +1006,38 @@ Use existing dra-example-driver or Kubernetes test driver for e2e testing.
...
 
-- the feature gate in Beta
-- all known bugs are fixed
+##### DynamicResourceAllocation (v0.18)
+
+- feature gate enabled by default
 - support integration with MultiKueue
 - e2e tests
-- TAS + DRA testing and support as a graduation requirement
+- CEL expression validation against ResourceSlice devices
```
kannon92, line 1031: https://github.com/kubernetes-sigs/kueue/pull/9742

This would need to merge for this KEP to be true.

kannon92 (commented):
> Important
The DeviceClass must exist before users submit workloads. There is no DeviceClass watcher in alpha. If a DeviceClass is created after a workload was marked inadmissible, the workload will not be re-evaluated until the next cluster event that triggers inadmissible workload requeuing (e.g., another workload completes or quota changes).

Should we update this text in the DRA section of our website?

sohankunkerkar:
> > Important
> > The DeviceClass must exist before users submit workloads. There is no DeviceClass watcher in alpha. If a DeviceClass is created after a workload was marked inadmissible, the workload will not be re-evaluated until the next cluster event that triggers inadmissible workload requeuing (e.g., another workload completes or quota changes).
> 
> Should we update this text in the DRA section of our website?

Yes, PR #10973 adds the DeviceClass watcher for beta, so this limitation no longer applies. We should update the website once this PR + #10973 merges.

sohankunkerkar:
@mimowo @tenzen-y  PTAL

test/integration/multikueue/dra_test.go
```diff
@@ -73,8 +72,6 @@ var _ = ginkgo.Describe("MultiKueue with DRA", ginkgo.Label("area:multikueue", "
 	})
 
 	ginkgo.BeforeEach(func() {
-		features.SetFeatureGateDuringTest(ginkgo.GinkgoTB(), features.KueueDRAIntegration, true)
```
mimowo, line 76: let's keep that. This makes the diff smaller and allows to easly flip the FG for local testing while in Beta.

mimowo:
/lgtm
/approve
Thank you for working towards graduating DRA integration, this is certainly a strategically important feature
/hold
in case @tenzen-y has extra comments

tenzen-y:
> /lgtm /approve Thank you for working towards graduating DRA integration, this is certainly a strategically important feature /hold in case @tenzen-y has extra comments

ACK, let me check this from now

tenzen-y (commented):
Mostly LGTM
My question is whether we should merge this before Extended Resource support graduates to beta. Because the Kueue DRA beta criteria contain it.

tenzen-y (commented):
/lgtm
/approve

This actually depands on https://github.com/kubernetes-sigs/kueue/pull/10973, but https://github.com/kubernetes-sigs/kueue/pull/10973 PR depends on this PR as commit.

So, I merge this one first, then if we find any critical problem in DRA Extended Resource supports, we will revert this PR.

mimowo:
> So, I merge this one first, then if we find any critical problem in DRA Extended Resource supports, we will revert this PR.

I have actually conerns about going Beta with "Extended Resource" for 0.18, because the currently implementation assumes that Extended Resource is any resource with "/". So, we start running new code paths for any users declaring nvidia.com/gpu. I proposed another solution when we first scrape the resources which are extended, not rely on strings. However, the new approach will only be provided now, so we don't have enough testing time to claim Beta. 

So, I would rather say that PD and ER stay Alpha in 0.18, and we move them both to Beta in 0.19 when we gain more confidence. I think this is perfectly valid approach to first graduate to Beta the "core" FG, and only later the secondary FGs. We took that approach with TAS where we graduated the main FG to Beta, while still keeping some FGs in Alpha for a while, incubating. The same approach is taken by the core k8s where the "core" FG is leading the graduation, while the secondary FGs graduate slower. 

In other words, I would like to still keep PD and ER in Alpha, while already move to Beta with the core FG. Let me know if this is ok for you.

tenzen-y:
> > So, I merge this one first, then if we find any critical problem in DRA Extended Resource supports, we will revert this PR.
> 
> I have actually conerns about going Beta with "Extended Resource" for 0.18, because the currently implementation assumes that Extended Resource is any resource with "/". So, we start running new code paths for any users declaring nvidia.com/gpu. I proposed another solution when we first scrape the resources which are extended, not rely on strings. However, the new approach will only be provided now, so we don't have enough testing time to claim Beta.
> 
> So, I would rather say that PD and ER stay Alpha in 0.18, and we move them both to Beta in 0.19 when we gain more confidence. I think this is perfectly valid approach to first graduate to Beta the "core" FG, and only later the secondary FGs. We took that approach with TAS where we graduated the main FG to Beta, while still keeping some FGs in Alpha for a while, incubating. The same approach is taken by the core k8s where the "core" FG is leading the graduation, while the secondary FGs graduate slower.
> 
> In other words, I would like to still keep PD and ER in Alpha, while already move to Beta with the core FG. Let me know if this is ok for you.

My comment: https://github.com/kubernetes-sigs/kueue/pull/10973#pullrequestreview-4347075423

PR #10999, opened 2026-05-07
Improve runtime performance for QuotaCheckStrategy test

mimowo:
#### What type of PR is this?

/kind bug
/kind cleanup

To improve testing time

```release-note
NONE
```

### DIFF
--- a/test/e2e/sequential/baseline/quota_check_strategy_test.go
+++ b/test/e2e/sequential/baseline/quota_check_strategy_test.go
@@ -76,7 +76,6 @@ var _ = ginkgo.Describe("QuotaCheckStrategy", ginkgo.Label("feature:quotacheckst
 	})
 
 	ginkgo.AfterAll(func() {
-		util.UpdateKueueConfigurationAndRestart(ctx, k8sClient, defaultKueueCfg, kindClusterName)
 		util.ExpectObjectToBeDeletedWithTimeout(ctx, k8sClient, clusterQueue, true, util.MediumTimeout)
 		util.ExpectObjectToBeDeletedWithTimeout(ctx, k8sClient, resourceFlavor, true, util.MediumTimeout)
 	})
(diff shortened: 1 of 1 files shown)


mimowo:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10999/pull-kueue-verify-main/2052265488649555968

I'm not sure if this is flake 

I will open issue if flake, retesting for now

mimowo:
Yeah, now passed, opened: https://github.com/kubernetes-sigs/kueue/issues/11000

test/e2e/sequential/baseline/quota_check_strategy_test.go
```diff
@@ -76,7 +76,6 @@ var _ = ginkgo.Describe("QuotaCheckStrategy", ginkgo.Label("feature:quotacheckst
 	})
 
 	ginkgo.AfterAll(func() {
-		util.UpdateKueueConfigurationAndRestart(ctx, k8sClient, defaultKueueCfg, kindClusterName)
```
mimowo, line 79: This is not needed, because in the suite we have the convention that each tests sets the configuration it wants in BeforeAll, so no need to revert twice. Maybe we could optimize `UpdateKueueConfigurationAndRestart` to skip the update if the configuration is the same, then it would be harmless.

test/e2e/sequential/baseline/quota_check_strategy_test.go
```diff
@@ -138,6 +137,7 @@ var _ = ginkgo.Describe("QuotaCheckStrategy", ginkgo.Label("feature:quotacheckst
 				createdJob = testingjob.MakeJob("ignoreundeclared-job", ns.Name).
 					Queue(kueue.LocalQueueName(localQueue.Name)).
 					Image(util.GetAgnHostImage(), util.BehaviorWaitForDeletion).
+					TerminationGracePeriod(1).
```
mimowo, line 140: Using short grace period to make sure the eviction is fast. Also using agnhost images, because the default is pause.

mbobrovskyi (commented):
Thank you!

/lgtm

PR #10950, opened 2026-05-05
Use LongTimeout for worker1 reconnect check after apiserver restart

reruno:
#### What type of PR is this?
/kind flake

#### Which issue(s) this PR fixes:

Fixes #10926 

#### Special notes for your reviewer:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10921/pull-kueue-test-e2e-multikueue-sequential-main/2051295881319157760

The PR fixes "MultiKueue Sequential when The connection to a worker cluster is unreliable Should update the cluster status to reflect the connection state" test, because after the test fail, the failure propagated into other tests, so fixing this will resolve issues with other tests as well. 

Test times out at 13:55:12, because manager cluster failed to connect to worker1 cluster apiserver  in MediumTimeout. 

It fails to connect at 13:54:37, but at 13:55:17 it succeeded. 
```
# ./artifacts/run-test-e2e-multikueue-sequential-1.35.0/kind-manager-worker/pods/kueue-system_kueue-controller-manager-7b5ff6c954-hnf2x_5d0dd006-e7ff-4107-ab42-e8c32aeaf91f/manager/0.log
2026-05-04T13:54:37.648166952Z stderr F 2026-05-04T13:54:37.64795336Z	ERROR	multikueue-cluster	multikueue/multikueuecluster.go:467	setting client config	{"replica-role": "leader", "name": "worker1", "reconcileID": "817ac025-61d4-46a6-8dcc-9b151a93140d", "retryAfter": "40s", "error": "unable to retrieve the complete list of server APIs: kueue.x-k8s.io/v1beta2: no matches for kueue.x-k8s.io/v1beta2, Resource="}
<...>
2026-05-04T13:55:17.673297679Z stderr F 2026-05-04T13:55:17.673085757Z	LEVEL(-2)	multikueue/multikueuecluster.go:216	Starting watch	{"clusterName": "worker1", "watchKind": "Workload.kueue.x-k8s.io"}
```

By increasing the timeout, we increase stability of the test, under such conditions. 

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

Issue #10926, opened 2026-05-04
MultiKueue Sequential when The connection to a worker cluster is unreliable Should update the cluster status to reflect the connection state


mimowo:
**Which test is flaking?**:
MultiKueue Sequential when The connection to a worker cluster is unreliable Should update the cluster status to reflect the connection state

**Link to failed CI job or steps to reproduce locally**:

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10921/pull-kueue-test-e2e-multikueue-sequential-main/2051295881319157760
**Failure message or logs**:
```
End To End MultiKueue Suite: kindest/node:v1.35.0: [It] MultiKueue Sequential when The connection to a worker cluster is unreliable Should update the cluster status to reflect the connection state expand_less	2m4s
{ failed [FAILED] Timed out after 45.000s.
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/multikueuesequential/e2e_test.go:346 with:
Expected
    <[]v1.Condition | len:1, cap:1>: [
        {
            Type: "Active",
            Status: "False",
            ObservedGeneration: 1,
            LastTransitionTime: {
                Time: 2026-05-04T13:54:27Z,
            },
            Reason: "ClientConnectionFailed",
            Message: "unable to retrieve the complete list of server APIs: kueue.x-k8s.io/v1beta2: no matches for kueue.x-k8s.io/v1beta2, Resource=",
        },
    ]
to contain element matching
    <*matchers.BeComparableToMatcher | 0x14f2ccdc5800>: {
        Expected: <v1.Condition>{
            Type: "Active",
            Status: "True",
            ObservedGeneration: 0,
            LastTransitionTime: {
                Time: 0001-01-01T00:00:00Z,
            },
            Reason: "Active",
            Message: "Connected",
        },
        Options: [
            <*cmp.pathFilter | 0x14f2cc25c0d8>{
                core: {},
                fnc: 0x651740,
                opt: <cmp.ignore>{core: {}},
            },
        ],
    }
In [It] at: /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/multikueuesequential/e2e_test.go:354 @ 05/04/26 13:55:12.876

```

**Anything else we need to know?**:

PR #11001, opened 2026-05-07
Only reconcile cluster updates on spec changes or deletion

reruno:
#### What type of PR is this?
/kind cleanup
/kind bug

Only reconcile multikueue cluster updates on spec changes or deletion

Discovered when working on #10926
Related to #10950

```release-note
MultiKueue: improved performance by avoiding unnecessary MultiKueueCluster reconciliations.
MultiKueueCluster updates now trigger reconciliation only when the spec changes or the object is deleted.
```

### DIFF
--- a/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
+++ b/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
@@ -841,7 +841,10 @@ func (c *clustersReconciler) Update(e event.UpdateEvent) bool {
 			log.Error(err, "AddOrUpdate FS watch")
 		}
 	}
-	return true
+	if !clusterNew.DeletionTimestamp.IsZero() {
+		return true
+	}
+	return clusterOld.Generation != clusterNew.Generation
 }
 
 func (c *clustersReconciler) Delete(e event.DeleteEvent) bool {
--- a/pkg/controller/admissionchecks/multikueue/multikueuecluster_test.go
+++ b/pkg/controller/admissionchecks/multikueue/multikueuecluster_test.go
@@ -1077,7 +1077,7 @@ func TestClustersReconcilerEventFilters(t *testing.T) {
 			},
 			wantReconcile: true,
 		},
-		"update cluster status only change does not skip reconcile": {
+		"update cluster status only change skips reconcile": {
 			invoke: func(p predicate.Predicate) bool {
 				return p.Update(event.UpdateEvent{
 					ObjectOld: baseCluster,
@@ -1087,7 +1087,7 @@ func TestClustersReconcilerEventFilters(t *testing.T) {
 						Obj(),
 				})
 			},
-			wantReconcile: true,
+			wantReconcile: false,
 		},
 		"update cluster deletion timestamp triggers reconcile": {
 			invoke: func(p predicate.Predicate) bool {
@@ -1098,20 +1098,22 @@ func TestClustersReconcilerEventFilters(t *testing.T) {
 			},
 			wantReconcile: true,
 		},
-		"update cluster no change does not skip reconcile": {
+		"update cluster no change skips reconcile": {
 			invoke: func(p predicate.Predicate) bool {
 				return p.Update(event.UpdateEvent{
 					ObjectOld: baseCluster,
 					ObjectNew: baseCluster.DeepCopy(),
 				})
 			},
-			wantReconcile: true,
+			wantReconcile: false,
 		},
 		"update cluster path location changed to secret triggers reconcile": {
 			invoke: func(p predicate.Predicate) bool {
+				objNew := baseCluster.DeepCopy()
+				objNew.SetGeneration(2)
 				return p.Update(event.UpdateEvent{
 					ObjectOld: baseClusterWithPath,
-					ObjectNew: baseCluster,
+					ObjectNew: objNew,
 				})
 			},
 			wantReconcile: true,


pkg/controller/admissionchecks/multikueue/multikueuecluster_test.go
```diff
@@ -925,3 +926,116 @@ func TestValidateKubeconfig(t *testing.T) {
 		})
 	}
 }
+
+func TestReconcileEventFilters(t *testing.T) {
```
mimowo: I like this way of testing, could we maybe also use to consolidate TestNonTasUsageReconcilerCreate, TestNonTasUsageReconcilerDelete and TestNonTasUsageReconcilerUpdate

Also, to improve testing in TestUpdateEventFilter, TestNodeEventsPredicateUpdate in pod_termination_controller_test.go

I will open an issue

mimowo: Opened: https://github.com/kubernetes-sigs/kueue/issues/11154 for discussion

pkg/controller/admissionchecks/multikueue/multikueuecluster_test.go
```diff
@@ -925,3 +926,116 @@ func TestValidateKubeconfig(t *testing.T) {
...
+
+	deletingCluster := baseCluster.DeepCopy()
+	now := metav1.Now()
+	deletingCluster.DeletionTimestamp = &now
+
+	cases := map[string]struct {
+		createEvent   *event.CreateEvent
+		updateEvent   *event.UpdateEvent
+		deleteEvent   *event.DeleteEvent
+		genericEvent  *event.GenericEvent
```
kshalot: I'm not necessarily saying it's much better, but here's an alternative that I think should work:

```suggestion
		invoke        func(p predicate.Predicate) bool
```

It would look like this in the test case:
```go
invoke: func(p predicate.Predicate) bool {
		return p.Create(event.CreateEvent{Object: baseCluster})
}
```

This avoids the switch-statement below and you can't accidentally pass two events in one case... but I think it's less "declarative".

I'm leaving this comment just as a food for thought since this might be re-used in the codebase (#11154).

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
```diff
@@ -714,10 +717,15 @@ func newClustersReconciler(
 		fsWatcher:           fsWatcher,
 		adapters:            adapters,
 		clusterProfileCreds: cpCreds,
+		logName:             "localqueue-reconciler",
```
kshalot: copy error?

kshalot:
What is the difference between this and #11153?

reruno:
> What is the difference between this and #11153?

[Only reconcile cluster updates on spec changes or deletion](https://github.com/kubernetes-sigs/kueue/pull/11001/changes/7e8988ac8ae646c2b6faa032e58e7593689f8db1)
This pr has this commit, that actually changes behavior of update event filter

kshalot:
Ok I see 7e8988ac8ae646c2b6faa032e58e7593689f8db1 is added here, I got confused because the diffs look pretty much identical

pkg/controller/admissionchecks/multikueue/multikueuecluster_test.go
mimowo: Yes, this is less declarative. I'm wondering if we could have some solution which is both less verbose, yet declarative.

reruno: https://github.com/kubernetes-sigs/kueue/pull/11153

proposed changes were implemented in this PR

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
reruno: https://github.com/kubernetes-sigs/kueue/pull/11153

proposed changes were implemented in this PR

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
```diff
@@ -841,7 +841,10 @@ func (c *clustersReconciler) Update(e event.UpdateEvent) bool {
 			log.Error(err, "AddOrUpdate FS watch")
 		}
 	}
-	return true
+	if !clusterNew.DeletionTimestamp.IsZero() {
+		return true
+	}
+	return !equality.Semantic.DeepEqual(clusterOld.Spec, clusterNew.Spec)
```
mimowo: Hm, I'm wondering if instead of comparing entire specs we could achieve the same with lightweight generation check, wdyt?

mimowo:
/release-note-edit
```release-note
MultiKueue: improved performance by avoiding unnecessary MultiKueueCluster reconciliations.
MultiKueueCluster updates now trigger reconciliation only when the spec changes or the object is deleted.
```

mimowo:
/lgtm
/approve
Thank you 👍

reruno:
/retest 

docker: Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint kind-worker2-control-plane (2ff96e372f59cafa0af0622028977edf479895ec721a7673eca5683a6bcfc762): Bind for 127.0.0.1:41347 failed: port is already allocated

Created the issue https://github.com/kubernetes-sigs/kueue/issues/11278

Issue #10901, opened 2026-05-04
Workloads corresponding to Jobs/Jobsets may be marked finished immediately after creation


mimowo:
**What happened**:

Workloads corresponding to JobSets got marked finished immediately after creation.

We observe in the logs that
1. the JobReconciler created a Workload for a JobSet
2. within the same millisecond the workload_controller marked the Workload as finished with the reason OwnerNotFound

This is because the cache inside Kueue for the JobReconciler already saw the owner JobSet, but the cache for the Workload_controller still didn't see the JobSet.

**Additionally**, we observed that the workload which was marked as finished was requeued via `Manager.RequeueWorkload `. this is not the root cause, but we should strenghten the check manager.go to do not requeue Finished workloads (use !workload.IsAdmissible(&w) ), because the workload which was requeued kept winning the fair sharing tournament in the consecutive scheduling cycles, thus blocking other workloads from making progress. If it was not requeued it would at least not cause problems to other workloads.

**What you expected to happen**:

Workloads should not be marked as Finished if the owner still exists, as in this case. 

**How to reproduce it (as minimally and precisely as possible)**:

This is very hard to repro, but we observed that on a busy large scale cluster.

**Anything else we need to know?**:

This makes the fix for Requeue of evicted Deployment workloads buggy https://github.com/kubernetes-sigs/kueue/pull/10274.

So we need search for a better fix, some options to consider:
1. mark the workloads as finished from within JobReconciler rather than workload_controller, but this would still leave some window open for the workload controller to put such a workload in the queue
2. (preferred) add an additional check in the workload_controller to skip adding a workload which has no owner onto the heap. Once the workload has the owner then it will be put back onto the heap.

mimowo:
I think we should do the  following:
1. strenghten the guard inside RequeueWorkload to prevent requeue of workloads which are inadmissible
2. mark  FinishOrphanedWorkloads as alpha
2. fix https://github.com/kubernetes-sigs/kueue/pull/10274 properly, if the proper fix is not finishing the workloads, then deprecate FinishOrphanedWorkloads so that it can be dropped in the next release

mveroone:
We've noticed a similar behavior on native K8s Jobs, which happens randomly time-wise but seem to affect several Workloads/Jobs at the same time over a few seconds.

As far as I can tell it didn't happen previously when we were running 0.16.1, but I cannot be sure as we have been using kueue on and off during the last few monthes.

Currently running 0.17.2

[...]

PR #11014, opened 2026-05-07
Prevent requeuing inadmissible workloads.

mbobrovskyi:
#### What type of PR is this?
/kind bug

Prevent requeuing inadmissible workloads.

Part of #10901

```release-note
Scheduling: Fixed a bug where in-flight workloads that were concurrently marked as finished (`Finished=True`) or deactivated could be requeued by Kueue's scheduler, causing re-scheduling attempts which were interfering with the scheduling of other workloads.
```

### DIFF
--- a/pkg/cache/queue/manager_test.go
+++ b/pkg/cache/queue/manager_test.go
@@ -969,57 +969,94 @@ func TestStatus(t *testing.T) {
 }
 
 func TestRequeueWorkloadStrictFIFO(t *testing.T) {
+	now := time.Now()
 	cq := utiltestingapi.MakeClusterQueue("cq").Obj()
 	queues := []*kueue.LocalQueue{
 		utiltestingapi.MakeLocalQueue("foo", "").ClusterQueue("cq").Obj(),
 		utiltestingapi.MakeLocalQueue("bar", "").Obj(),
 	}
-	cases := []struct {
+	cases := map[string]struct {
 		workload     *kueue.Workload
 		inClient     bool
 		inQueue      bool
 		wantRequeued bool
 	}{
-		{
-			workload: utiltestingapi.MakeWorkload("existing_queue_and_obj", "").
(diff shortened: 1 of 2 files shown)


pkg/cache/queue/manager_test.go
```diff
@@ -974,52 +974,64 @@ func TestRequeueWorkloadStrictFIFO(t *testing.T) {
...
+		"not in client": {
+			workload: utiltestingapi.MakeWorkload("wl", "").
 				Queue("foo").
 				Obj(),
+			inClient: false,
+			inQueue:  false,
 		},
-		{
-			workload: utiltestingapi.MakeWorkload("already_in_queue", "").
+		"finished": {
```
mimowo, line 1022: do we also have a test case for deactivated? Lets add that too

mbobrovskyi: Done

mimowo:
I think "inadmissible" workloads is not well defined currently in Kueue. We have different meanings of that - for example workload with Inadmissible reason for QuotaReserved=False are not counted in the kueue_pending_workloads metric, so we need to first clean this up before using in messages to users.
/release-note-edit
```release-note
Fixed a bug where in-flight workloads that were concurrently marked as finished (`Finished=True`) or deactivated 
could be requeued by Kueue's scheduler, causing re-scheduling attempts which were interfering with scheduling
of other workloads.
```

mimowo:
Thank you 👍 
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16

tenzen-y:
/release-note-edit
```release-note
Scheduling: Fixed a bug where in-flight workloads that were concurrently marked as finished (`Finished=True`) or deactivated could be requeued by Kueue's scheduler, causing re-scheduling attempts which were interfering with the scheduling of other workloads.
```

PR #11023, opened 2026-05-07
fix: Use non exist image as default image for e2e tests

### LINKED ISSUE #11003: Do not use "pause" as default image for e2e tests

I would like to make sure we are not using the "pause" as default image for e2e tests in functions like MakeJob.

I would propose to just replace ["pause"](https://github.com/kubernetes-sigs/kueue/blob/3b4f8ffe1ee9b0fab6f1c6d680e8f37b90bacd20/pkg/util/testingjobs/tfjob/wrappers_tfjob.go#L83) with some image which does not exist, like "dummy_non_existing_image". Then e2e tests would fail fast. For integration and unit tests it will not matter what is there.

**Why is this needed**:

Using "pause" "works", but it requires pulling the image, and we should be using k8s stack, ie using "agnhost" for testing whenever possible. Because it "works" it is very easy to miss that on review.  One place was recently fixed here: https://github.com/kubernetes-sigs/kueue/pull/10999

gyliu513:
#### What type of PR is this?
/kind bug
/kind cleanup
/area testing

/cc @mimowo 

Fixes #11003 

```release-note
NONE
```

### DIFF
--- a/pkg/controller/jobs/leaderworkerset/leaderworkerset_reconciler_test.go
+++ b/pkg/controller/jobs/leaderworkerset/leaderworkerset_reconciler_test.go
@@ -40,6 +40,7 @@ import (
 	"sigs.k8s.io/kueue/pkg/features"
 	utiltesting "sigs.k8s.io/kueue/pkg/util/testing"
 	utiltestingapi "sigs.k8s.io/kueue/pkg/util/testing/v1beta2"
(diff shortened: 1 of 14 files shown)


mimowo:
Thank you 👍 
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16

mimowo:
@gyliu513 please prepare the cherrycpiks manually

PR #11031, opened 2026-05-07
perf explore reduce cqs

baoalvin1:
#### What type of PR is this?
/area testing
/kind feature

Updates the large scale performance test to test higher load and throughput

Fixes #
https://github.com/kubernetes-sigs/kueue/issues/10773

```release-note
NONE
```

### DIFF
--- a/test/performance/scheduler/configs/large-scale/generator.yaml
+++ b/test/performance/scheduler/configs/large-scale/generator.yaml
@@ -1,10 +1,11 @@
 # Scheduler Performance Test Configuration - Large Scale
-# Tests scheduler throughput at high CQ count: 20 cohorts × 100 queues = 2000 cluster queues
-# Workloads: 30 per queue × 2000 queues = 60,000 total (1 pod each)
+# Tests scheduler throughput at moderate CQ count: 10 cohorts × 100 queues = 1000 cluster queues
+# Workloads: 50 per queue × 1000 queues = 50,000 total (1 pod each)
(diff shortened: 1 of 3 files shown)


baoalvin1:
@mimowo I'd be okay with this I think

baoalvin1:
I can modify the wall time to be like 21/22 minutes if that's fine with you, the first test run looks like it took just over 20 minutes

mimowo:
> I can modify the wall time to be like 21/22 minutes if that's fine with you, the first test run looks like it took just over 20 minutes

Let's try to fit under 20, maybe that would not make much difference for you, but will make me happy that we can reach the goal of all ci jobs below 20min eventually.

Also, I'm ok to have the job running as periodic, then we could have it even 30min.

baoalvin1:
@mimowo walltime is under 20 minutes but it looks like the test finished after 25? I'm fine with only doing a periodic if you feel that would be better, I think we just need some form of test running at higher scale

baoalvin1:
do you want me to remove the test infra pr job then and just keep the periodic?

mimowo:
/test pull-kueue-test-large-scale-scheduling-perf-main
checking stability

mimowo:
> do you want me to remove the test infra pr job then and just keep the periodic?

Yes, let's do that. We will get to know about the failures the next day if any, because we have an alerting setup for all CI jobs.  Also, you can setup your own alerting based on the testgrid.

mimowo:
Thank you 👍 
/lgtm
/approve

PR #11033, opened 2026-05-07
TAS: fix delete predicate to reconcile terminal pods

amy:
The NonTasUsageReconciler.Delete predicate was delegating to belongsToNonTASCache, which filters out terminated pods (Succeeded/Failed). This created a correctness hole: if the informer skips the Running→Terminated Update event and delivers only the Delete event with the pod already in a terminal phase, the predicate returns false and no reconciliation is triggered — leaving the pod's resource usage permanently stuck in the non-TAS cache.

#### The Fix:
Broaden the Delete predicate to pass through any scheduled non-TAS pod regardless of phase. 
The Create predicate is intentionally left unchanged — there's no value in tracking a pod that was already terminal when the cache first saw it.

Related to get tests passing for: https://github.com/kubernetes-sigs/kueue/pull/10806

Fixes #

```release-note
TAS: Fixed cache cleanup for non-TAS Pods that reach a terminal phase without Kueue observing the expected status update, preventing stale Pod usage from remaining in the TAS cache.
```

### DIFF
--- a/pkg/controller/tas/non_tas_usage_controller.go
+++ b/pkg/controller/tas/non_tas_usage_controller.go
@@ -109,7 +109,11 @@ func (r *NonTasUsageReconciler) Update(e event.TypedUpdateEvent[*corev1.Pod]) bo
 }
 
 func (r *NonTasUsageReconciler) Delete(e event.TypedDeleteEvent[*corev1.Pod]) bool {
-	return belongsToNonTASCache(e.Object)
+	// Don't filter on terminal phase: if the informer skips the Running→Terminated
+	// Update and delivers only the Delete event with the pod already Succeeded/Failed,
+	// the pod's usage would never be removed from the cache. DeletePodByKey is
+	// idempotent, so double-removal is safe.
+	return len(e.Object.Spec.NodeName) > 0 && !utiltas.IsTAS(e.Object)
 }
 
 func (r *NonTasUsageReconciler) Generic(event.TypedGenericEvent[*corev1.Pod]) bool {
--- a/pkg/controller/tas/non_tas_usage_controller_test.go
+++ b/pkg/controller/tas/non_tas_usage_controller_test.go
@@ -76,7 +76,7 @@ func TestBelongsToNonTASCache(t *testing.T) {
 	}
 }
 
-func TestNonTasUsageReconcilerCreateDelete(t *testing.T) {
+func TestNonTasUsageReconcilerCreate(t *testing.T) {
 	reconciler := &NonTasUsageReconciler{}
 
 	tests := []struct {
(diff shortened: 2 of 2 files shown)


pkg/controller/tas/non_tas_usage_controller.go
```diff
@@ -109,7 +109,11 @@ func (r *NonTasUsageReconciler) Update(e event.TypedUpdateEvent[*corev1.Pod]) bo
 }
 
 func (r *NonTasUsageReconciler) Delete(e event.TypedDeleteEvent[*corev1.Pod]) bool {
-	return belongsToNonTASCache(e.Object)
+	// Don't filter on terminal phase: if the informer skips the Running→Terminated
+	// Update and delivers only the Delete event with the pod already Succeeded/Failed,
+	// the pod's usage would never be removed from the cache. DeletePodByKey is
+	// idempotent, so double-removal is safe.
+	return len(e.Object.Spec.NodeName) > 0 && !utiltas.IsTAS(e.Object)
```
mimowo, line 116: Question: how have you observed the api-server skipping the Running->Terminated transition. I'm wondering how is that possible. Surely, we could say this is possible in case of a bug in controller-runtime or api-server, fine, so we have a guardrail, but I'm wondering if there is also a no-bug scenario for such a skip possible.

cc @kshalot with whom I recently investigated a case which might look like such a skip.

In any case this looks good to me, but we need to add a release note for the PR, I think this is a bugfix essentially, just trying to think what is the scenario.

amy: @mimowo So I made this change because of failures for https://github.com/kubernetes-sigs/kueue/pull/10806. If a Delete event arrives with a Succeeded pod, it should still be processed as a delete event. In the unit test you can see that if you attempt to delete while the pod is marked as succeeded, it doesn't trigger a delete.

```
{
	name: "terminated scheduled non-TAS pod",
	pod:  testingpod.MakePod("pod", "ns").NodeName("node-a").StatusPhase(corev1.PodSucceeded).Obj(),
	want: true,
},

...
...
if got := reconciler.Delete(event.TypedDeleteEvent[*corev1.Pod]{Object: tc.pod}); got != tc.want {
	t.Errorf("Delete() = %v, want %v", got, tc.want)
}
```

The cases are theoretical. So alternatively we can also drop this test case if it seems impractical:
- watch disconnects & relist when the pod succeeds/deletes in the gap
- job pod cleanup after ttl: job completes, pod succeeds, job operator triggers deletion after pod success
- ... there's more theoretical cases

mimowo: I think this is ok as a guard check, because the downside is very small one more Reconcile, at the same time the risk of cache corruption is much worse - requires Kueue restart.

In particular the documentation on DeleteStateUnknown suggests we might be missing some events: https://github.com/kubernetes-sigs/controller-runtime/blob/v0.24.0/pkg/event/event.go#L64-L66

Anyways, https://github.com/kubernetes-sigs/kueue/pull/10806 should not rely on that -  it should focus on testing the "no-bug" scenarios, rather than this very tricky one.

mimowo:
/release-note-edit
```release-note
TAS: introduce a safeguard check in the cache for non-TAS Pods to ensure the cache is cleaned when 
the Pods transitioned to the terminal (Failed or Succeeded state), but Kueue hasn't received the Running -> Failed
Update event.
```

mimowo:
/release-note-edit
```release-note
TAS: Fixed cache cleanup for non-TAS Pods that reach a terminal phase without Kueue observing the expected status update, preventing stale Pod usage from remaining in the TAS cache.
```

Issue #11029, opened 2026-05-07
WaitForPodsReady with default Timeout and a long RecoveryTimeout should continue running workload if pod recovers before recoveryTimeout


mimowo:
**Which test is flaking?**:

WaitForPodsReady with default Timeout and a long RecoveryTimeout should continue running workload if pod recovers before recoveryTimeout
**Link to failed CI job or steps to reproduce locally**:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10949/pull-kueue-test-e2e-sequential-baseline-main/2052452824444309504

**Failure message or logs**:
```
End To End Sequential Baseline Suite: kindest/node:v1.35.0: [It] WaitForPodsReady with default Timeout and a long RecoveryTimeout should continue running workload if pod recovers before recoveryTimeout expand_less	2m3s
{ failed [FAILED] Timed out after 45.012s.
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/sequential/baseline/waitforpodsready_test.go:478 with:
Expected
    <[]v1.Condition | len:3, cap:4>: [
        {
            Type: "PodsReady",
            Status: "False",
            ObservedGeneration: 1,
            LastTransitionTime: {
                Time: 2026-05-07T18:28:37Z,
            },
            Reason: "WaitForRecovery",
            Message: "At least one pod has failed, waiting for recovery",
        },
        {
            Type: "QuotaReserved",
            Status: "True",
            ObservedGeneration: 1,
            LastTransitionTime: {
                Time: 2026-05-07T18:28:14Z,
            },
            Reason: "QuotaReserved",
            Message: "Quota reserved in ClusterQueue cq",
        },
        {
            Type: "Admitted",
            Status: "True",
            ObservedGeneration: 1,
            LastTransitionTime: {
                Time: 2026-05-07T18:28:14Z,
            },
            Reason: "Admitted",
            Message: "The workload is admitted",
        },
    ]
to have condition type PodsReady, status True and reason Recovered
In [It] at: /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/sequential/baseline/waitforpodsready_test.go:479 @ 05/07/26 18:29:05.228
}
[...]

PR #11054, opened 2026-05-08
AgentSkills: add kueue-flake-debugger skill

mimowo:
#### What type of PR is this?

/kind documentation
/kind feature

To speed up flake analysis with agent. 

It was successfully tested in https://github.com/kubernetes-sigs/kueue/issues/11029

We have the idea for follow up to extend the skill with comparing against a successful run: https://github.com/kubernetes-sigs/kueue/issues/11055

```release-note
AgentSkills: Add an agent skill for debugging Kueue CI flakes.
```

### DIFF
--- a/cmd/experimental/agent/skills/kueue-flake-debugger.md
+++ b/cmd/experimental/agent/skills/kueue-flake-debugger.md
@@ -18,7 +18,8 @@ You are an expert in Kueue which is the project for Workload orchestration.
 When asked to debug a flake with the given link to Github issue ABCD
 then identify the prow link to the failure.
 
-Please report the list of all prow links, for example:
+Please report the list of all prow links. Try using `gh` first, if available,
+otherwise fallback to other tools you have, for example curl:
 
 ```sh
 # Note: If the issue context is already provided or visible, you can skip this automated curl and manually identify the Prow link from the issue description.
--- a/cmd/experimental/agent/skills/kueue-flake-debugger.md
+++ b/cmd/experimental/agent/skills/kueue-flake-debugger.md
@@ -15,8 +15,7 @@ You are an expert in Kueue which is the project for Workload orchestration.
 
 ### Step one - initial build-log analysis
 
-When asked to debug a flake with the given link to Github issue ABCD
-then identify the prow link to the failure.
+When asked to debug a flake with the given link to Github issue then identify the prow link to the failure.
 
 Please report the list of all prow links. Try using `gh` first, if available,
 otherwise fallback to other tools you have, for example curl:
--- a/cmd/experimental/agent/skills/kueue-flake-debugger.md
+++ b/cmd/experimental/agent/skills/kueue-flake-debugger.md
@@ -1,0 +1,118 @@
+---
+name: kueue-flake-debugger
+description: Expertise in debugging Kueue flakes
+---
+
+# Kueue flake debugger
+
+You are an expert in Kueue which is the project for Workload orchestration.
+
+## Expertise
+
+- Deep understanding of Kueue.
+
+## Flake debugging
+
+### Step one - initial build-log analysis
+
+When asked to debug a flake with the given link to Github issue ABCD
+then identify the prow link to the failure.
+
+Please report the list of all prow links, for example:
+
+```sh
+# Note: If the issue context is already provided or visible, you can skip this automated curl and manually identify the Prow link from the issue description.
+curl -Lv https://github.com/kubernetes-sigs/kueue/issues/ABCD 2>/dev/null | grep -e "href=\"https://prow\.k8s\.io/view/gs/kubernetes-ci-logs/pr-logs/pull"
+```
+
+Then, extract the links.
+
+Choose the first, let call it PROW_LOG, eg:
+
+```sh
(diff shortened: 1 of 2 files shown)


mimowo:
cc @amy @kannon92 @reruno @mbobrovskyi

mimowo:
cc @PBundyra

cmd/experimental/agent/skills/kueue-flake-debugger.md
```diff
@@ -0,0 +1,118 @@
...
+### Step one - initial build-log analysis
+
+When asked to debug a flake with the given link to Github issue ABCD
+then identify the prow link to the failure.
+
+Please report the list of all prow links, for example:
+
+```sh
+# Note: If the issue context is already provided or visible, you can skip this automated curl and manually identify the Prow link from the issue description.
+curl -Lv https://github.com/kubernetes-sigs/kueue/issues/ABCD 2>/dev/null | grep -e "href=\"https://prow\.k8s\.io/view/gs/kubernetes-ci-logs/pr-logs/pull"
```
kshalot, line 25: We could try using the `gh` CLI, like:
```
gh issue view <issue-number> --repo kubernetes-sigs/kueue --json body --jq '.body'
```
and fallback to curl if it's not available. This would be more token efficient, since gives you the description only and not the HTML page.

cmd/experimental/agent/skills/kueue-flake-debugger.md
```diff
@@ -0,0 +1,118 @@
...
+
+Then, extract the links.
+
+Choose the first, let call it PROW_LOG, eg:
+
+```sh
+BASE_PROW_LOG=https://gcsweb.k8s.io/gcs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/9617/pull-kueue-test-e2e-main-1-33/2028383367677874176
+```
+
+Then appenend "/build-log.txt", down load using curl, say
```
kshalot, line 36:
```suggestion
Then appenend "/build-log.txt", download using curl, say
```

cmd/experimental/agent/skills/kueue-flake-debugger.md
```diff
@@ -0,0 +1,118 @@
...
+description: Expertise in debugging Kueue flakes
+---
+
+# Kueue flake debugger
+
+You are an expert in Kueue which is the project for Workload orchestration.
+
+## Expertise
+
+- Deep understanding of Kueue.
```
kshalot, line 12: Wouldn't it better to focus more on the CI/CD Kubernetes-native systems expert persona rather than Kueue?

cmd/experimental/agent/skills/kueue-flake-debugger.md
kshalot: We use `gh` in other tools anyway.

cmd/experimental/agent/skills/kueue-flake-debugger.md
mimowo: I guess it doesn't really matter for the better models they can translate the skills to large extend.

cmd/experimental/agent/skills/kueue-flake-debugger.md
kshalot: I mean - it's piped immediately to grep so token-wise should be identical. But maybe it's a bit cleaner.

Sometimes there are comments under issues, like: "Another occurrence: <prow_link>". So it depends whether we want to catch those as well.

mimowo: sounds like a good idea, let me try. I think this could be a skill on its own to use gh, rather than curl

mimowo: That being said both gemini and chatgpt didn't have issue following the skill in the current form using curl, see: https://github.com/kubernetes-sigs/kueue/issues/11029#issuecomment-4404864868

let me see if I can convert that to use gh/ gcs tooling

kannon92: Yea gh is useful. I find I don't really even need to mention how to use it for gemini/claude.

mimowo: Pushed this updated, ptal: https://github.com/kubernetes-sigs/kueue/pull/11054/changes/38c28bbbad9f3052dfb93d89f70ff3d6b7b905dc

cmd/experimental/agent/skills/kueue-flake-debugger.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Flake debugging
+
+### Step one - initial build-log analysis
+
+When asked to debug a flake with the given link to Github issue ABCD
```
amy: nit: you can remove ABCD

cmd/experimental/agent/skills/kueue-flake-debugger.md
amy: bump it still says down <space> load

cmd/experimental/agent/skills/kueue-flake-debugger.md
mimowo: done

amy:
/lgtm 

exciting!

Issue #11024, opened 2026-05-07
Add first-class “workload/job dependencies” support to Kueue


gyliu513:
**What would you like to be added**:

Add **first-class “workload/job dependencies”** support to Kueue so users can express *“B must not be admitted until A finishes (Succeeded/Completed)”* as a built-in capability, instead of implementing bespoke external controllers.

Concretely, introduce a **built-in dependency gate** integrated with Kueue’s admission flow, so dependent workloads remain queued but **not admitted** until their prerequisites are satisfied.

Possible scope for the built-in feature (community can pick the exact shape during design):

- A standardized way to declare dependencies (e.g., via `Workload`-level fields or well-defined annotations).
- A built-in “Dependency” admission check (or equivalent mechanism) that:
  - Evaluates prerequisite completion (Succeeded/Finished).
  - Provides clear status/messages on the blocked workload (`Pending` / `Retry` / `Rejected`).
  - Supports basic patterns: 1→1 dependency, fan-in, optionally cross-namespace policy.
  - Avoids deadlocks and minimizes extra reconcile churn.

**Why is this needed**:

Many batch/ML pipelines have strict ordering constraints (ETL → training → evaluation, preprocessing → simulation → aggregation). Today Kueue users typically solve this with external workflow engines (Argo/Tekton) or by writing their own `AdmissionCheck` controller, but that has drawbacks:

- **Common need, repeated re-implementation**: dependency gating is a frequent ask; every organization ends up building a similar controller with slightly different semantics and bugs.
- **Harder for users to get right**: a custom check controller must handle watch patterns, reconcile triggers, failure modes, backoff, race conditions (objects created out of order), and consistent status reporting—non-trivial for new users.
- **Inconsistent UX & operability**: different implementations produce different statuses/events/metrics, making debugging “why isn’t my workload running?” harder across clusters.
- **Tight integration benefits**: Kueue can implement this more reliably by integrating with its existing queuing model, admission checks state machine, and workload lifecycle (finished/evicted/deactivated), and by defining supported semantics.
- **Lower adoption friction**: a built-in solution reduces the barrier for users who want Kueue *without* adopting a full workflow system.
- **Community-maintained correctness**: when dependency gating is built-in, the community can ensure correctness across features like preemption, fair-sharing, stop policies, and future changes to Workload lifecycle.

(Users can still build custom controllers for advanced DAG logic, but a built-in “simple dependency gating” covers the 80% case and provides a standard foundation.)

[...]

PR #11064, opened 2026-05-08
kep: Design proposal for workloaddependency

gyliu513:
#### What type of PR is this?
/kind feature
/kind kep

/cc @mimowo @kannon92 @mwielgus 

I will try to bring this to bi-weekly meeting as well, but would love to get some early comments, thanks!

Kep for https://github.com/kubernetes-sigs/kueue/issues/11024

```release-note
NONE
```

### DIFF
--- a/keps/11024-workload-dependency/README.md
+++ b/keps/11024-workload-dependency/README.md
@@ -1,0 +1,760 @@
+# KEP-11024: Workload Dependency Admission Check
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: Sequential ML Pipeline](#story-1-sequential-ml-pipeline)
+    - [Story 2: JobSet Manager-Worker Ordering](#story-2-jobset-manager-worker-ordering)
+    - [Story 3: Fan-in Aggregation Job](#story-3-fan-in-aggregation-job)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Overview](#overview)
+  - [End-to-End Walkthrough](#end-to-end-walkthrough)
+    - [Step 1: Cluster Administrator Setup](#step-1-cluster-administrator-setup)
+    - [Step 2: User Submits Jobs](#step-2-user-submits-jobs)
+    - [Step 3: Kueue Creates Workloads](#step-3-kueue-creates-workloads)
+    - [Step 4: Prerequisite Job Runs and Completes](#step-4-prerequisite-job-runs-and-completes)
+    - [Step 5: Dependent Job Is Admitted](#step-5-dependent-job-is-admitted)
+  - [Sequence Diagram](#sequence-diagram)
+  - [Edge Case Walkthrough: Prerequisite Failure](#edge-case-walkthrough-prerequisite-failure)
+  - [Edge Case Walkthrough: Fan-in with Partial Completion](#edge-case-walkthrough-fan-in-with-partial-completion)
+  - [API Changes](#api-changes)
+    - [Dependency Declaration via Annotation](#dependency-declaration-via-annotation)
+    - [WorkloadDependencyConfig CRD](#workloaddependencyconfig-crd)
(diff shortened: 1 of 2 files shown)


tenzen-y:
/hold 
https://github.com/kubernetes-sigs/kueue/issues/11024#issuecomment-4408387022

keps/11024-workload-dependency/README.md
```diff
@@ -0,0 +1,760 @@
...
+    ▼
+Kueue Job Integration Layer
+    │  propagates dependency info to Workload annotation
+    ▼
+Workload (status.admissionChecks[workload-dependency] = Pending)
+    │
+    ▼
+WorkloadDependency Controller (built-in)
+    │  watches prerequisite Workloads for status changes
+    │  sets AdmissionCheckState → Ready / Pending / Rejected
```
mimowo, line 214: I'm wondering if you considered using the AdmissionGatedBy annotation as an alternative for the mechanism. One benefit is that with the annotation you don't need to reserve quota for all Jobs waiting on the AdmissionChecks and reserving quota. The downside is that it may create a delay between the consecutive steps, so there is a trade-off, but I'm asking in case maybe you didn't consider the alternative.

keps/11024-workload-dependency/README.md
```diff
@@ -0,0 +1,760 @@
...
+User Job (annotation: kueue.x-k8s.io/depends-on)
+    │
+    ▼
+Kueue Job Integration Layer
+    │  propagates dependency info to Workload annotation
+    ▼
+Workload (status.admissionChecks[workload-dependency] = Pending)
+    │
+    ▼
+WorkloadDependency Controller (built-in)
```
mimowo, line 212: Given that the solution can be completely decoupled from the core Kueue (either if this is AdmissionCheck, or external controller managing AdmissionGatedBy annotation), I would like to consider introducing this as an experimental sub-project within Kueue, under `cmd/experimental/dependencycontroller`.

I hope this could mitigate the concerns raised by @tenzen-y and @kannon92

keps/11024-workload-dependency/README.md
```diff
@@ -0,0 +1,760 @@
...
+Each time another shard completes, the controller reconciles and updates the message. Once all three are `Succeeded`, the check transitions to `Ready` and the `aggregation` workload is admitted.
+
+### API Changes
+
+#### Dependency Declaration via Annotation
+
+A new well-known annotation is introduced on user-facing Job objects:
+
+```
+kueue.x-k8s.io/depends-on
```
mimowo, line 546: Do we put some constraints on the Jobs managed by the mechanism, for example:
- should the dependency have the same GVK as the waiting job?
- should the dependency have the same owner as the waiting Job?

In particular if we relax on (1.) then I think we need to enrich the API in the annotation. Maybe for alpha it is better to put the constraints, wdyt?

mimowo (commented):
@tenzen-y @kannon92 I left the initial comments as I was reading if the proposal is reasonable, and I think it is. In particular, I hope the use of an external cmd/experimental controller could mitigate your concerns about complicating the core Kueue. I think it is a good idea to present and discuss further the proposal on the wg-batch meetings.

keps/11024-workload-dependency/README.md
```diff
@@ -0,0 +1,760 @@
...
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Alternative 1: Dependency fields directly on Workload spec](#alternative-1-dependency-fields-directly-on-workload-spec)
+  - [Alternative 2: External-only custom AdmissionCheck controller](#alternative-2-external-only-custom-admissioncheck-controller)
+  - [Alternative 3: Integration with Argo Workflows / Tekton only](#alternative-3-integration-with-argo-workflows--tekton-only)
+<!-- /toc -->
+
+## Summary
+
+This KEP proposes a built-in **WorkloadDependency** AdmissionCheck controller for Kueue that allows users to express ordering constraints between workloads: *"workload B must not be admitted until workload A has completed successfully."* The dependency is declared via a well-defined annotation on the user-facing Job object (BatchJob, JobSet, etc.) and evaluated as a standard Kueue AdmissionCheck. This is not a workflow engine — it covers the common "sequential admission gating" case without DAG scheduling, data flow, or retry orchestration.
```
mwielgus, line 50: This should probably be not implemented as an AdmissionCheck. Admission check are evaluated once the quota is prebooked, and in this case we could be hours (if not days) before all dependencies are completed. In the meantime, the other workloads should be allowed to execute.

PR #10819, opened 2026-04-29
[AGENT] Reviewer skill

amy:
#### What type of PR is this?
adds a new code review skill

/kind cleanup

#### What this PR does / why we need it:
adds a new code review skill 

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #11088, opened 2026-05-09
AgentSkills: Add code-eval skill, that evaluates diff code based on defined rules …

reruno:
…and gives a score from 0 to 100

I added code evaluation skills with defined rules that was created based on mimowo comments on PRs. The main purpose of the skill is evaluating code quality and representing code quality in numeric form. This will allow us to determine what LLM prompt techniques improve LLMs coding abilities and what do not. Code evaluation skill also could act as a testing tool for future SKILL that focuses on increasing performance of LLM coding abilities.     

tool: claude code
model: Opus 4.7 (1M context) 

### Example 1 

Prompt
```
evaluate code between these commits 4af7becae5c3d6786d112ba023e572ae3113607c c91286c0aaf45d003ec3c8e93b03c3ab28de5f25
```

Output
```
Code Evaluation Report

  Commits: 4af7beca..c91286c0
  Files changed:
  - Makefile-test.mk (+10/0)
  - hack/e2e-common.sh (+395/-33)
  - site/content/en/docs/contribution_guidelines/testing.md (+5/0)
  - site/content/zh-CN/docs/contribution_guidelines/testing.md (+3/0)
  
  Note: Higher score = worse code. 0 means no findings in that domain.

  ---
  Skill: Architectural Decisions (weight 1.5)

  Domain score: 67.5 pt
  Findings:
  - High | +22.5 pt | Massive duplicated install_* preamble: same ~15-line "deployment exists → image → tag → version compare → log" block copied across 9
  installers (appwrapper, jobset, kubeflow, kubeflow_trainer, mpi, kuberay, lws, cert_manager, dra).
  - Medium | +10.5 pt | Scope creep — MPI manifest curl|yq patching: unrelated SSA aggregated-ClusterRole workaround bundled into an idempotency refactor.
  - Medium | +10.5 pt | Scope creep — DRA image tag derivation: changes tag from local to a version-derived string in all modes, not just the dev-reuse path the
  PR is scoped to.
  - Medium | +10.5 pt | install_multicluster uses CRD-existence-only check while seven peers do full version comparison — asymmetric abstraction.
  - Low | +4.5 pt | Pointless rename of positional args in kind_load (e2e_cluster_name/e2e_kubeconfig).
  - Low | +4.5 pt | kubectl_args array reconstructed in three places instead of a thin e2e_kubectl wrapper.
  - Low | +4.5 pt | Cert-manager dual-container-name fallback inline rather than in a shared image-tag helper.

  Skill: Code Style (weight 1.0)

  Domain score: 37 pt
  Findings:
  - High | +15 pt | e2e_is_truthy returns 1 (failure) for truthy inputs — opposite of its name and of bash convention; every call site reads the inverse of what
  it says. 
  - Medium | +7 pt | *"/lws:"* || *"/lws@"* || *"/lws"* — third pattern subsumes first two and also matches unrelated images (e.g., foo/lws-debug).
  - Low | +3 pt | Comment typo "re-installed_version" on hack/e2e-common.sh:86.
  - Low | +3 pt | e2e_image_ref_get_tag violates verb-first convention used by sibling helpers (e2e_deployment_exists, e2e_versions_match, e2e_crd_exists).
  - Low | +3 pt | e2e_cluster_name/e2e_kubeconfig prefix drift in kind_load vs name/kubeconfig everywhere else.
  - Low | +3 pt | local kubeconfig=${2:-} for required args silences arg-missing bugs; inconsistent with local name=$1.
  - Low | +3 pt | force_reinstall carried as the literal strings "true"/"false" and routed through the broken truthy helper rather than a plain [[ ... == "true" 
  ]].

  Skill: Buggy Behavior (weight 2.0)

  Domain score: 54 pt
  Findings:
  - Medium | +14 pt | install_kuberay delete guard is effectively always-true: if e2e_is_truthy "${E2E_ENFORCE_OPERATOR_UPDATE}" || e2e_is_truthy 
  "${force_reinstall}" — given the inverted helper, at least one operand is always falsy, so the delete runs in every code path. --ignore-not-found=true masks
  the mistake on fresh clusters; on a non-fresh CI cluster it would tear down kuberay unnecessarily.
  - Medium | +14 pt | LWS image match too broad: third pattern picks up any image containing /lws, including foo/lwsxyz, and could pin installed_version to the
  wrong container's tag.
  - Medium | +14 pt | Inconsistent kubectl wait … || true: appwrapper/jobset/kubeflow/kubeflow_trainer fail loud on wait timeouts;
  mpi/kuberay/lws/cert-manager-sidecars swallow them. A non-ready operator now produces a confusing downstream test failure.
  - Low | +6 pt | install_mpi adds curl|yq|kubectl pipeline without ensuring set -o pipefail in callers; a curl failure could feed an empty manifest to kubectl.
  - Low | +6 pt | dra_image_tag silently changes from constant "local" to a version-derived slug whenever DRA_EXAMPLE_DRIVER_VERSION is set — tooling/runbooks
  grepping for the historical tag will silently miss the image.

  Skill: Comments (weight 0.5)

  Domain score: 4.5 pt
  Findings:
  - Low | +1.5 pt | Doc claim "external operators are installed only once" is too strong; the code reinstalls on version mismatch.
  - Low | +1.5 pt | install_kuberay NOTE leads with reused-cluster background rather than the actual why (create -k non-idempotency).
  - Low | +1.5 pt | No comment documenting the deliberately-inverted e2e_is_truthy return semantics — exactly the kind of non-obvious behavior that warrants a
  why-comment.

  Skill: Security (weight 2.0)

  Domain score: 0 pt
  Findings:
  - No findings (dev-only test harness; curl|yq|kubectl uses TLS-verified HTTPS to a pinned ref, array expansions are quoted correctly, no credential/secret/exec
   surface introduced).

  ---
  Final Score: 163 pt

  ---
  Recommendations

  Recommendation 1: Collapse the duplicated install_* idempotency preamble

  Severity: High
  Domain: Architectural Decisions
  Findings: Massive duplicated install_* preamble.
  Problem: Nine install_* functions repeat a near-identical ~15-line block (early-exit on mode/flag, deployment exists, fetch container image, extract tag,
  version compare, branch log).
  Reason: Every future operator copies the same boilerplate. The diff already shows drift between copies — different container-name selectors, "upgrading" vs.
  "reinstalling" wording, the cert-manager dual-container fallback, the /lws triple-pattern. Bug fixes will not propagate uniformly.
  Solution: Add one helper, e.g. e2e_skip_if_installed label kubeconfig ns deployment expected_version selector, returning 0 to short-circuit the caller:
  function install_jobset {
      local name=$1 kubeconfig=${2:-}
      e2e_skip_if_installed "JobSet" "$kubeconfig" "${JOBSET_DEPLOYMENT_NAMESPACE:-jobset-system}" \
          "${JOBSET_CONTROLLER_MANAGER_DEPLOYMENT:-jobset-controller-manager}" \
          "${JOBSET_VERSION:-}" "image~jobset/jobset" && return 0
      cluster_kind_load_image "$name" "$JOBSET_IMAGE"
      kubectl apply --kubeconfig="$kubeconfig" --server-side -f "$JOBSET_MANIFEST"
      kubectl wait --kubeconfig="$kubeconfig" deploy/... --for=condition=available --timeout=5m
  }
  Locations:
  - hack/e2e-common.sh:213-231 (install_appwrapper)
  - hack/e2e-common.sh:250-275 (install_jobset)
  - hack/e2e-common.sh:293-312 (install_kubeflow)
  - hack/e2e-common.sh:329-348 (install_kubeflow_trainer)
  - hack/e2e-common.sh:379-404 (install_mpi)
  - hack/e2e-common.sh:437-457 (install_kuberay)
  - hack/e2e-common.sh:484-509 (install_lws)
  - hack/e2e-common.sh:524-548 (install_cert_manager)
  - hack/e2e-common.sh:577-597 (install_dra_example_driver)

  Recommendation 2: Fix the inverted e2e_is_truthy return convention

  Severity: High
  Domain: Code Style
  Findings: e2e_is_truthy returns 1 (failure) for truthy inputs.
  Problem: The function returns 1 (shell-failure) for 1|true|yes|on|... and 0 (success) for everything else. if e2e_is_truthy "$x"; then ... therefore runs only
  when x is not truthy.
  Reason: A predicate that lies about its sign is the textbook "name lies" defect; every caller has to be read backwards, and at least one already gets it
  functionally wrong (Rec. 3).
  Solution:
  function e2e_is_truthy {
      case "${1:-}" in
          1|true|TRUE|True|yes|YES|Yes|y|Y|on|ON|On) return 0 ;;
          *) return 1 ;;
      esac
  }
  Then flip every call: && e2e_is_truthy X (intended "and is truthy") that currently means "and is NOT truthy" becomes && ! e2e_is_truthy X, and so on. Re-audit
  the kuberay double-call after fixing.
  Locations:
  - hack/e2e-common.sh:91-96 (definition)
  - hack/e2e-common.sh:213, 250, 293, 329, 379, 437, 466 (×2), 484, 524, 560, 577 (call sites)

  Recommendation 3: Fix the install_kuberay delete guard

  Severity: Medium
  Domain: Buggy Behavior
  Findings: install_kuberay delete guard is effectively always-true.
  Problem: if e2e_is_truthy "${E2E_ENFORCE_OPERATOR_UPDATE}" || e2e_is_truthy "${force_reinstall}"; then kubectl delete -k ... ; fi runs the delete whenever
  either var is falsy. Because force_reinstall is the literal string "false" by default and only flips to "true" inside the dev/non-enforce branch, at least one
  operand is always falsy and the delete always runs.
  Reason: The destructive delete is intended to gate on "force reinstall requested"; today it fires unconditionally. --ignore-not-found=true masks the mistake on
   fresh clusters but does an unnecessary teardown elsewhere, and the next reader will be misled about the actual semantics.
  Solution: After Rec. 2, write the condition straight:
  if e2e_is_truthy "${E2E_ENFORCE_OPERATOR_UPDATE}" || [[ "${force_reinstall}" == "true" ]]; then
      kubectl ${kubectl_args[@]+"${kubectl_args[@]}"} delete -k "${KUBERAY_MANIFEST}" --ignore-not-found=true
  fi
  Locations:
  - hack/e2e-common.sh:466-468
  
  Recommendation 4: Tighten the LWS image-match patterns

  Severity: Medium
  Domain: Buggy Behavior (also Code Style)
  Findings: LWS image match too broad; third pattern subsumes first two.
  Problem: if [[ "${img}" == *"/lws:"* || "${img}" == *"/lws@"* || "${img}" == *"/lws"* ]] is equivalent to a single overly broad *"/lws"* test, which matches
  images like foo/lws-debug:1.0.
  Reason: Mismatching the wrong container's image silently records a wrong installed_version, then bypasses install or upgrades when it shouldn't.
  Solution:
  if [[ "${img}" == *"/lws:"* || "${img}" == *"/lws@"* ]]; then
  Or anchor to the canonical published image name (e.g. *"sigs.k8s.io/lws/lws-controller-manager"*).
  Locations:
  - hack/e2e-common.sh:492
  
  Recommendation 5: Stop masking kubectl wait failures inconsistently

  Severity: Medium
  Domain: Buggy Behavior
  Findings: Inconsistent kubectl wait … || true.
  Problem: install_mpi, install_kuberay, install_lws, and the cert-manager webhook/cainjector waits append || true while install_appwrapper, install_jobset,
  install_kubeflow, install_kubeflow_trainer do not.
  Reason: Silently continuing past an unhealthy operator turns "MPI operator never became Available" into a far-away test failure. The inconsistency also leaves
  readers unsure which controllers are required.
  Solution: Remove || true from all kubectl wait --for=condition=available calls unless a documented reason exists; if so, add an inline comment and echo a
  warning.
  Locations:
  - hack/e2e-common.sh:416 (install_mpi)
  - hack/e2e-common.sh:470 (install_kuberay)
  - hack/e2e-common.sh:513 (install_lws)
  - hack/e2e-common.sh:552-553 (cert-manager webhook/cainjector)
  
  Recommendation 6: Split the MPI manifest patching out of this PR

  Severity: Medium
  Domain: Architectural Decisions
  Findings: Scope creep — MPI manifest curl|yq patching.
  Problem: install_mpi adds a curl | yq | kubectl pipeline that drops empty rules: [] from aggregated ClusterRoles. The block has a clear NOTE, but it is
  unrelated to "make installs idempotent in dev mode" and changes CI behavior too.
  Reason: Bundling an unrelated bugfix obscures both changes in git blame, complicates reverting either independently, and makes reviewers focused on idempotency
   skim past a SSA workaround.
  Solution: Extract into its own commit/PR; or, if it must ship together, gate the transformation on E2E_MODE=dev so CI behavior is unchanged.
  Locations:
  - hack/e2e-common.sh:407-415

  Recommendation 7: Hoist the DRA image-tag change out of the idempotency refactor

  Severity: Medium
  Domain: Architectural Decisions
  Findings: Scope creep — DRA image tag derivation.
  Problem: dra_image_tag="${expected_version#v}"; dra_image_tag="${dra_image_tag//\//-}" changes the image tag from the constant "local" to a version-derived
  slug in all modes.
  Reason: The PR documentation, env var name, and Makefile additions all scope the work to dev-mode idempotency. Future bisects of DRA failures will skip this
  commit.
  Solution: Split into a separate commit with a message explaining why version-based tagging is required (presumably to allow the new version-compare check to
  work), or gate the new tag on the same E2E_MODE=dev condition.
  Locations:
  - hack/e2e-common.sh:907-911

  Recommendation 8: Align install_multicluster with the rest of the installers

  Severity: Medium
  Domain: Architectural Decisions
  Findings: Inconsistent abstraction in install_multicluster.
  Problem: install_multicluster checks only CRD existence; the other seven installers do deployment + image-tag version comparison.
  Reason: A future reader copying the wrong template adds a silent stale-CRD risk (or over-engineers a fresh installer). The asymmetry is a maintenance hazard.
  Solution: Either accept "object exists, skip" as the canonical pattern and simplify the others (preferred for a dev harness), or pull the version-compare flow
  into a CRD-aware helper that compares metadata.annotations or stored schema version.
  Locations:
  - hack/e2e-common.sh:559-567

  Recommendation 9: Fix typo "re-installed_version"

  Severity: Low
  Domain: Code Style
  Findings: Comment typo "re-installed_version".
  Problem: # When set (truthy), external operators are forcefully re-installed_version even in E2E_MODE=dev.
  Reason: Ungrammatical leftover from a rename; sits directly above the new env var and confuses readers about whether _version is meaningful.
  Solution: # When set (truthy), external operators are forcefully re-installed on every run, even in E2E_MODE=dev.
  Locations:
  - hack/e2e-common.sh:86
  
  Recommendation 10: Rename e2e_image_ref_get_tag to verb-first form

  Severity: Low
  Domain: Code Style
  Findings: e2e_image_ref_get_tag violates the file's verb-first naming convention.
  Problem: Siblings are e2e_deployment_exists, e2e_versions_match, e2e_crd_exists.
  Reason: Convention drift inside the same commit reduces grep-ability.
  Solution: Rename to e2e_get_image_tag (or e2e_image_tag).
  Locations:
  - hack/e2e-common.sh:111 (def); 219, 259, 299, 335, 388, 443, 493, 535, 586 (call sites)

  Recommendation 11: Drop the e2e_ prefix on kind_load locals

  Severity: Low
  Domain: Code Style
  Findings: e2e_cluster_name/e2e_kubeconfig convention drift.
  Problem: kind_load introduces e2e_cluster_name and e2e_kubeconfig while every callee uses bare name / kubeconfig.
  Reason: The prefix adds nothing (entire file is e2e), and the rename inflates the diff. Reverting also restores the original "$1" "$2" call sites unchanged.
  Solution: Revert the rename — call install_X "$1" "$2" as before, or use local cluster_name=$1; local kubeconfig=$2.
  Locations:
  - hack/e2e-common.sh:148-149, 156, 160, 167, 172, 177, 181, 185, 189, 193, 197
  
  Recommendation 12: Don't default required positional args

  Severity: Low
  Domain: Code Style
  Findings: ${2:-} on required positional args.
  Problem: Inconsistent use of local kubeconfig=${2:-} (default-empty) for an argument the function cannot operate without.
  Reason: Hides a caller's bug behind a downstream kubectl error.
  Solution: Use local kubeconfig=$2 for required args, or : "${kubeconfig:?install_*: kubeconfig required}". Reserve ${X:-} for truly optional inputs.
  Locations:
  - hack/e2e-common.sh:208, 245, 288, 324, 374, 427, 479, 519, 574
  
  Recommendation 13: Replace e2e_is_truthy "${force_reinstall}" with a plain compare

  Severity: Low
  Domain: Code Style
  Findings: force_reinstall as string boolean.
  Problem: force_reinstall is set internally to literal strings; routing it through the broken truthy helper compounds Rec. 3.
  Reason: Internal flags should use bash booleans directly; the helper is for user-supplied env vars.
  Solution: if [[ "${force_reinstall}" == "true" ]]; then ...
  Locations:
  - hack/e2e-common.sh:431, 453, 466

  Recommendation 14: Encapsulate kubectl_args construction

  Severity: Low
  Domain: Architectural Decisions
  Findings: kubectl_args rebuilt in three places.
  Problem: e2e_deployment_exists, e2e_crd_exists, and install_kuberay each rebuild the same array preamble.
  Reason: Trivial duplication that will multiply with the next caller.
  Solution:
  function e2e_kubectl { local kc=$1; shift; [[ -n "$kc" ]] && set -- --kubeconfig="$kc" "$@"; kubectl "$@"; }
  Locations:
  - hack/e2e-common.sh:103-108, 131-138, 432-435
  
  Recommendation 15: Move cert-manager's container-name fallback into the shared helper

  Severity: Low
  Domain: Architectural Decisions
  Findings: Cert-manager dual-container-name fallback inline.
  Problem: install_cert_manager hand-rolls a second jsonpath when container name cert-manager returns empty.
  Reason: After Rec. 1 lands, this dual lookup belongs inside the shared image-tag helper, accepting either a container name or an image substring selector.
  Solution: Extend e2e_get_image_tag kubeconfig ns deployment selector... to iterate selectors until one yields non-empty.
  Locations:
  - hack/e2e-common.sh:529-534

  Recommendation 16: Correct the dev-mode docs note

  Severity: Low
  Domain: Comments
  Findings: Doc claim "installed only once" is too strong.
  Problem: The English and Chinese contributor docs say operators "are installed only once" in E2E_MODE=dev; the code in fact upgrades when the installed version
   differs.
  Reason: Contributors who bump KUBEFLOW_MPI_VERSION etc. will be surprised by upgrades despite the doc's promise.
  Solution:
  When reusing a kept cluster in `E2E_MODE=dev`, external operators (MPI, KubeRay, etc.)
  are reinstalled only when their installed version no longer matches the requested one.
  To force reinstalling on every run regardless of version, set
  `E2E_ENFORCE_OPERATOR_UPDATE=true`.
  Locations:
  - site/content/en/docs/contribution_guidelines/testing.md:137-141
  - site/content/zh-CN/docs/contribution_guidelines/testing.md:137-138
  
  Recommendation 17: Lead the install_kuberay NOTE with the actual reason

  Severity: Low
  Domain: Comments
  Findings: install_kuberay NOTE wording.
  Problem: The comment opens with reused-cluster background instead of the load-bearing why (create -k non-idempotency).
  Reason: Why-comments should put the reason for this code first.
  Solution:
  # `kubectl create -k` (used instead of apply per https://github.com/ray-project/kuberay/issues/504)
  # is not idempotent: it exits non-zero if any object already exists. In E2E_MODE=dev we reuse
  # the kind cluster between runs, so we must explicitly delete the previous manifest first
  # whenever an upgrade or forced reinstall is requested.
  Locations:
  - hack/e2e-common.sh:461-464
  
  Recommendation 18: Document or remove the inverted e2e_is_truthy semantics

  Severity: Low
  Domain: Comments
  Findings: No comment explaining the inverted return convention.
  Problem: The function deliberately inverts bash predicate semantics and has no comment saying so.
  Reason: Either fix the function (Rec. 2) — preferred — or, if kept, the inversion deserves an explicit why-comment so the next contributor doesn't reintroduce
  the kuberay-style bug.
  Solution: Fold into Rec. 2 by flipping the return values. If that is rejected, add:
  # Returns 1 (failure) when $1 is truthy, 0 (success) when falsy.
  # Inversion is deliberate so that `cond && e2e_is_truthy x` runs the body when x is NOT truthy.
  Locations:
  - hack/e2e-common.sh:91-96
```

### Example 2 
Prompt
```
evaluate code between these commits c43e788acc787db54809ec01954d6441c80ace5a 0b2c1d0f7675448fc2455904e0d62b4a8130a51d
```

Output 
```
Code Evaluation Report

  Commits: c43e788acc787db54809ec01954d6441c80ace5a..0b2c1d0f7675448fc2455904e0d62b4a8130a51d
  Files changed:
  - pkg/controller/jobframework/reconciler.go (+2/-1)
  - pkg/controller/jobframework/validation.go (+10/-1)
  - pkg/controller/jobframework/validation_test.go (+1/-7)
  - pkg/controller/jobs/leaderworkerset/leaderworkerset_reconciler.go (+30/-15)
  - pkg/controller/jobs/leaderworkerset/leaderworkerset_webhook_test.go (+32)
  - pkg/util/equality/podset.go (+11/-1)
  - pkg/util/equality/podset_test.go (+32/-6)
  - test/e2e/singlecluster/leaderworkerset_test.go (+63)

  Note: Higher score = worse code. 0 means no findings in that domain.

  ---
  Skill: Architectural Decisions (weight 1.5)

  Domain score: 25.5 pt
  Findings:
  - Medium | +10.5 pt | Scope creep into shared validation: ValidateImmutablePodGroupPodSpec is permanently nodeSelector-blind, changing the contract for every
  caller (notably StatefulSet at statefulset_webhook.go:173) for an LWS-only feature.
  - Medium | +10.5 pt | Misleading helper name updateWorkload: the new method (leaderworkerset_reconciler.go:336-350) may instead delete the workload when
  podSets differ; the name promises an in-place update.
  - Low | +4.5 pt | Rename createPrebuiltWorkload → createWorkload drops the meaningful "Prebuilt" qualifier that distinguished this controller's
  prebuilt-workload path from the framework's standard construction.

  Skill: Code Style (weight 1.0)

  Domain score: 6 pt
  Findings:
  - Low | +3 pt | Imprecise helper name ignoreNodeSelector: signature reads like a pure filter but mutates the input map in place (validation.go:250-253).
  - Low | +3 pt | Misleading test case name "different requests in node selector with ignoreNodeSelector" — node selectors have no "requests"; the sibling case
  at line 134 carries the same wording (podset_test.go:128-160).

  Skill: Buggy Behavior (weight 2.0)

  Domain score: 64 pt
  Findings:
  - High | +30 pt | Webhook now allows StatefulSet nodeSelector changes but the StatefulSet reconciler does not resync the Workload PodSet — Workload accounting
  drifts from the actual Pod nodeSelector (validation.go:207-212 is shared with statefulset_webhook.go:173; statefulset_reconciler.go:182-247 has no recreate
  path).
  - Medium | +14 pt | EquivalentToWorkload unconditionally appends WithIgnoreNodeSelector() (reconciler.go:1210-1211), flipping equivalence semantics for every
  framework. Tolerations are guarded by IsAdmitted(wl) — nodeSelector should follow the same convention.
  - Medium | +14 pt | LWS PodSet diff triggers hard-delete instead of workload.Finish(..., WorkloadFinishedReasonOutOfSync, ...). Operators lose the audit
  trail/event that the generic prebuilt path emits on out-of-sync recreation (leaderworkerset_reconciler.go:343-345 vs. reconciler.go:1202-1203).
  - Low | +6 pt | Test rename "change nodeTemplate" → "change nodeSelector" drops the only negative-error assertion in the table without replacement; the
  immutability contract for other fields (containers, affinity) loses coverage in this table (validation_test.go:152-156).

  Skill: Comments (weight 0.5)

  Domain score: 8 pt
  Findings:
  - Medium | +3.5 pt | Godoc on ValidateImmutablePodGroupPodSpec still says "no changes are allowed to the PodSpec except fields that required for role-hash
  generation," but nodeSelector is now also permitted (validation.go:207).
  - Low | +1.5 pt | ignoreNodeSelector helper has no "why" comment explaining the LWS recreate contract (validation.go:250).
  - Low | +1.5 pt | New exported WithIgnoreNodeSelector lacks godoc describing the caller contract that the workload must be recreated (podset.go:40).
  - Low | +1.5 pt | TODO above comparePodTemplate mentioning nodeSelectors/tolerations as future work is now stale — both are checked by default with explicit
  opt-out (podset.go:42-44).

  Skill: Security (weight 2.0)

  Domain score: 6 pt
  Findings:
  - Low | +6 pt | Mild DoS amplification: a tenant churning nodeSelector forces the controller into delete+recreate loops on each mutation. Admission is re-run
  each cycle so there is no quota bypass, but the tenant has a cheap multiplier on controller/API-server work (leaderworkerset_reconciler.go:336-350).

  ---
  Final Score: 109.5 pt

  ---
  Recommendations

  Recommendation 1: Make StatefulSet reconciler sync PodSets when nodeSelector changes (or scope the relaxation to LWS)

  Severity: High
  Domain: Buggy Behavior
  Findings: Webhook now allows StatefulSet nodeSelector changes but the StatefulSet reconciler does not resync the Workload PodSet
  Problem: ValidateImmutablePodGroupPodSpec is shared by StatefulSet and LWS. After relaxation, a user can mutate a suspended StatefulSet's nodeSelector, but the
   StatefulSet reconciler does not detect the diff or recreate the Workload — Pods schedule with the new nodeSelector while the Workload's PodSet still carries
  the old one.
  Reason: Kueue's accounting (quota, flavor matching, admission checks) assumes the admitted PodSet matches the runtime PodSpec. Allowing those to drift defeats
  the immutability invariant and produces subtle scheduling mismatches that are hard to triage in production.
  Solution: Make the relaxation opt-in. Add a validation option and pass it only from the LWS webhook:
  // validation.go
  func ValidateImmutablePodGroupPodSpec(new, old *corev1.PodSpec, fp *field.Path, opts ...ValidationOption) field.ErrorList { ... }

  // leaderworkerset webhook
  jobframework.ValidateImmutablePodGroupPodSpec(..., jobframework.WithAllowNodeSelectorMutation())
  Alternative: replicate the LWS delete-and-recreate logic in statefulset_reconciler.reconcileWorkload — but the opt-in option has the smaller blast radius.
  Locations:
  - pkg/controller/jobframework/validation.go:207-212
  - pkg/controller/jobs/statefulset/statefulset_webhook.go:173
  - pkg/controller/jobs/statefulset/statefulset_reconciler.go:182-247
  
  Recommendation 2: Scope WithIgnoreNodeSelector in EquivalentToWorkload to admitted workloads

  Severity: Medium
  Domain: Buggy Behavior
  Findings: EquivalentToWorkload unconditionally appends WithIgnoreNodeSelector()
  Problem: reconciler.go:1210-1211 ignores nodeSelector for every framework. Tolerations are guarded by IsAdmitted(wl) because scheduler-injected fields only
  appear post-admission; nodeSelector should follow the same convention.
  Reason: Before admission, a nodeSelector difference is a real user change and should retire the prebuilt workload via the standard OutOfSync path. Silently
  treating it as equivalent masks intent.
  Solution:
  opts := make([]equality.ComparePodSetsOption, 0, 2)
  if workload.IsAdmitted(wl) {
      opts = append(opts, equality.WithIgnoreTolerations(), equality.WithIgnoreNodeSelector())
  }
  Locations:
  - pkg/controller/jobframework/reconciler.go:1210-1211
  
  Recommendation 3: Use workload.Finish(..., OutOfSync, ...) instead of hard-delete on PodSet drift

  Severity: Medium
  Domain: Buggy Behavior
  Findings: LWS PodSet diff triggers hard-delete instead of workload.Finish
  Problem: The LWS updateWorkload path removes the finalizer and deletes the Workload directly when PodSets diverge, skipping the standard
  WorkloadFinishedReasonOutOfSync event/condition emitted by the generic prebuilt-workload sync.
  Reason: Operators triaging "why did my pods just get evicted" expect a discoverable OutOfSync reason in the Workload status/events. Hard-delete erases that
  audit trail.
  Solution:
  return workload.Finish(ctx, r.client, wl, kueue.WorkloadFinishedReasonOutOfSync,
      "LeaderWorkerSet PodSpec changed; recreating workload", r.clock)
  Keep deleteWorkload for the genuine toDelete (stale workload cleanup) branch only.
  Locations:
  - pkg/controller/jobs/leaderworkerset/leaderworkerset_reconciler.go:343-345
  
  Recommendation 4: Limit the validation relaxation to LWS (architectural scope)

  Severity: Medium
  Domain: Architectural Decisions
  Findings: Scope creep into shared validation
  Problem: The PR is framed as an LWS-only behavior change, but ValidateImmutablePodGroupPodSpec is shared API. Every current and future caller silently inherits
   "nodeSelector is mutable."
  Reason: The contract of an immutability validator should be set by the caller, not by an internal helper's choice of fields. Hidden contract changes propagate
  to integrations that haven't designed for them.
  Solution: Move the nodeSelector masking behind an option (see Recommendation 1's WithAllowNodeSelectorMutation()) and pass it explicitly from the LWS webhook.
  This pairs cleanly with the StatefulSet fix above.
  Locations:
  - pkg/controller/jobframework/validation.go:207-212

  Recommendation 5: Rename updateWorkload to reflect its conditional delete

  Severity: Medium
  Domain: Architectural Decisions
  Findings: Misleading helper name updateWorkload
  Problem: r.updateWorkload can call r.deleteWorkload when PodSets diverge — the name promises in-place update.
  Reason: Hiding a destructive branch behind a benign verb is a maintenance hazard. Future callers will compose it expecting idempotent semantics and may regress
   the recreation path without noticing.
  Solution: Lift the recreation decision into the reconcile loop:
  if !equality.ComparePodSetSlices(podSets, toUpdate[i].Spec.PodSets) {
      return r.deleteWorkload(ctx, toUpdate[i])
  }   
  return r.updateWorkload(ctx, lws, toUpdate[i])
  Or rename to syncWorkload / reconcileWorkloadUpdate to telegraph that it may recreate.
  Locations:
  - pkg/controller/jobs/leaderworkerset/leaderworkerset_reconciler.go:336-350
  - pkg/controller/jobs/leaderworkerset/leaderworkerset_reconciler.go:151-153

  Recommendation 6: Fix inaccurate godoc on ValidateImmutablePodGroupPodSpec

  Severity: Medium
  Domain: Comments
  Findings: Godoc on ValidateImmutablePodGroupPodSpec is now wrong
  Problem: The comment claims the function disallows all PodSpec changes "except fields that required for role-hash generation," but nodeSelector is now also
  masked before comparison.
  Reason: Inaccurate comments are worse than missing ones — a future maintainer will add new "allowed" fields inconsistently with the true contract. The grammar
  ("fields that required") is also broken.
  Solution:
  // ValidateImmutablePodGroupPodSpec validates that the PodSpec is immutable for
  // serving workloads, except for fields used in role-hash generation and
  // nodeSelector (which the LWS reconciler handles by recreating the workload).
  Locations:
  - pkg/controller/jobframework/validation.go:207
  
  Recommendation 7: Restore the rename Prebuilt qualifier (or rename the trio consistently)

  Severity: Low
  Domain: Architectural Decisions
  Findings: Rename createPrebuiltWorkload → createWorkload loses the "Prebuilt" descriptor
  Problem: "Prebuilt" was a useful pointer that this controller constructs workloads outside the framework's standard path. The rename was justified only by
  symmetry with updateWorkload/deleteWorkload.
  Reason: Symmetry that erases meaningful distinctions has negative value for readers.
  Solution: Keep createPrebuiltWorkload, or rename the trio (createPrebuiltWorkload/updatePrebuiltWorkload/deletePrebuiltWorkload).
  Locations:
  - pkg/controller/jobs/leaderworkerset/leaderworkerset_reconciler.go:144,219

  Recommendation 8: Make ignoreNodeSelector non-mutating (and clearer named)

  Severity: Low
  Domain: Code Style
  Findings: Imprecise helper name ignoreNodeSelector
  Problem: The signature func(shape map[string]any) map[string]any reads like a pure transformation, but the function mutates the caller's map.
  Reason: Safe today only because SpecShape returns fresh maps. A future caller passing a shared/cached map will silently corrupt it, and a nil map will panic.
  Solution:
  func withoutNodeSelector(shape map[string]any) map[string]any {
      out := maps.Clone(shape)
      delete(out, "nodeSelector")
      return out
  }
  Or inline the two assignments directly in the caller — locality is obvious.
  Locations:
  - pkg/controller/jobframework/validation.go:250-253
  
  Recommendation 9: Fix misleading test case names referring to "requests in node selector"

  Severity: Low
  Domain: Code Style
  Findings: Misleading test case name "different requests in node selector with ignoreNodeSelector"
  Problem: Both the new case and its sibling at line 134 say "different requests in node selector" — node selectors have no "requests"; the cases differ only in
  NodeSelector values.
  Reason: Test names are scenario documentation. Inaccurate names mislead future readers, especially since the sibling now expects wantEquivalent: false and the
  wording is doubly stale.
  Solution:
  "different node selector across pod sets":                          { ... wantEquivalent: false },
  "different node selector across pod sets with ignoreNodeSelector":  { ... ignoreNodeSelector: true, wantEquivalent: true },
  Locations:
  - pkg/util/equality/podset_test.go:128-144
  - pkg/util/equality/podset_test.go:145-160
  
  Recommendation 10: Restore negative immutability coverage in validation_test.go

  Severity: Low
  Domain: Buggy Behavior
  Findings: Test rename drops the only negative-error assertion in the table
  Problem: The diff converts "change nodeTemplate" (asserting an error) into "change nodeSelector" (asserting no error) and leaves the table without any other
  negative case.
  Reason: No remaining guard against a future change that inadvertently allows containers, affinity, or other immutable fields to mutate.
  Solution: Keep the new positive case and add at least one negative case (e.g., "change affinity" or "change container resources" expecting an error).
  Locations:
  - pkg/controller/jobframework/validation_test.go:152-156

  Recommendation 11: Add a "why" comment on ignoreNodeSelector

  Severity: Low
  Domain: Comments
  Findings: Missing why on ignoreNodeSelector
  Problem: Helper has no comment explaining why nodeSelector is special-cased.
  Reason: Readers must trace to the LWS reconciler's updateWorkload to learn that nodeSelector mutation is permitted because the controller recreates the
  workload. That's exactly the non-obvious-why that justifies a comment.
  Solution:
  // ignoreNodeSelector masks nodeSelector from the immutability check so LWS
  // template updates to nodeSelector are permitted; the reconciler responds by
  // recreating the workload (see leaderworkerset_reconciler.updateWorkload).
  Locations:
  - pkg/controller/jobframework/validation.go:250

  Recommendation 12: Document WithIgnoreNodeSelector caller contract

  Severity: Low
  Domain: Comments
  Findings: Missing godoc on WithIgnoreNodeSelector
  Problem: Exported option has no comment; its caller contract (recreate workload on diff) is non-obvious.
  Reason: Exported API surface should explain when callers should opt in, especially when the option has a paired behavioral obligation.
  Solution:
  // WithIgnoreNodeSelector skips nodeSelector when comparing PodSpecs. Use only
  // when the caller will recreate the workload on nodeSelector change (e.g., the
  // LeaderWorkerSet reconciler).
  func WithIgnoreNodeSelector() ComparePodSetsOption { ... }
  Locations:
  - pkg/util/equality/podset.go:40

  Recommendation 13: Refresh the stale TODO above comparePodTemplate

  Severity: Low
  Domain: Comments
  Findings: Stale TODO mentioning nodeSelectors/tolerations as future work
  Problem: TODO says we should "extend the check to … nodeSelectors(when suspended), tolerations". Both are now checked by default with explicit opt-out.
  Reason: Stale TODOs mislead reviewers about which fields are enforced and what work remains.
  Solution: Either delete or rewrite to point at remaining gaps (priority, affinity, topologySpreadConstraints, …).
  Locations:
  - pkg/util/equality/podset.go:42-44

  Recommendation 14: Emit an event on tenant-driven workload recreation

  Severity: Low
  Domain: Security
  Findings: Reconciler DoS amplification on nodeSelector churn
  Problem: A tenant repeatedly mutating LWS nodeSelector forces delete+recreate cycles; while each cycle re-runs admission (so no quota bypass), the multiplier
  is uncapped.
  Reason: Observability of recreation makes abuse detectable and lets controller-runtime's workqueue rate-limiter and operators react. Each recreate also kills
  running pods — operators should be able to find a breadcrumb.
  Solution:
  if !equality.ComparePodSetSlices(podSets, wl.Spec.PodSets) {
      r.record.Eventf(lws, corev1.EventTypeWarning, "WorkloadRecreated",
          "PodSet template changed; recreating workload")
      return r.deleteWorkload(ctx, wl)
  }
  (Subsumed by Recommendation 3 if you switch to workload.Finish(..., OutOfSync, ...), which already emits the appropriate condition.)
  Locations:
  - pkg/controller/jobs/leaderworkerset/leaderworkerset_reconciler.go:336-350
```

```release-note
NONE
```

### DIFF
--- a/cmd/experimental/skills/reviewer/code-eval/SKILL.md
+++ b/cmd/experimental/skills/reviewer/code-eval/SKILL.md
@@ -1,0 +1,191 @@
+---
+name: code-eval
+description: Evaluate the quality of a git diff between two commits by delegating per-skill analysis (Code Style, Buggy Behavior, Comments, Architectural Decisions, Security) to subagents, then aggregating their findings and recommendations into a scored report.
+argument-hint: <BaseCommit> <HeadCommit>
+license: Apache-2.0
+metadata:
+  copyright: The Kubernetes Authors
+---
+
+# Code Evaluation Skill
+
+Evaluate the quality of a git diff between two commits. Each evaluation skill listed
+below contributes findings and recommendations for one domain of concern; this skill's
+job is to fan out to each one in parallel, then assemble the results into a single
+scored report.
+
+Use this when the user wants a code-quality evaluation. If the user expresses the need
+without providing **BaseCommit** and **HeadCommit**, ask them for both before
+proceeding.
+
+## Arguments
+
+The user invoked this with: $ARGUMENTS
+
+Parse two positional arguments from `$ARGUMENTS`:
+- **BaseCommit** — the base (older) commit
+- **HeadCommit** — the head (newer) commit
+
+## Principles
+
+- Ground every penalty in a specific line or pattern from the diff. No vague criticism.
+- Do not penalize stylistic choices that are consistent with the surrounding code, even if you personally prefer something else.
+- Do not invent bugs that aren't there. If the code looks correct, say so.
+- Reward simplicity. If the diff is clean and well-structured, say so explicitly — a clean diff scores 0 in every domain.
+- Severity matters: backwards-compatibility violations and logic errors on the hot path are more serious than naming nits. Reflect this in point weights.
+- **Higher score = worse code.** Every finding adds points; nothing subtracts. There is no upper cap.
+
+## Step 1 — Retrieve the diff
+
+Run:
+```
+git diff <BaseCommit> <HeadCommit>
+```
+
+Read the full output. This is the **Diff Code** that the subagent evaluators will work
+against.
+
+Also run:
+```
+git diff <BaseCommit> <HeadCommit> --stat
+```
+
+to get a high-level overview of changed files for the final report.
+
+## Step 2 — Fan out to per-domain subagents
+
+For each **domain** in the table below, spawn one subagent **in parallel** (single
+message, multiple Agent tool calls). One subagent per domain — not one per sub-skill.
+Each domain SKILL.md is a table-of-contents over many small rule sub-skills and
+inlines them via `@<rule>/SKILL.md` references. Each subagent must:
+
+1. Read the domain SKILL.md end-to-end. That file pulls in every rule sub-skill in the
+   domain plus the shared "How to report" instructions.
+2. Read the diff between `<BaseCommit>` and `<HeadCommit>` and load enough surrounding
+   context (unchanged lines, nearby functions, imports, existing helpers, log-verbosity
+   conventions in the package) to make grounded judgments.
+3. Identify every violation of any rule sub-skill in the domain.
+4. Classify each violation as **high**, **medium**, or **low**.
+5. Return findings AND recommendations in the format defined in the domain SKILL.md.
+
+Domains:
+| Domain | Domain weight |
+|---|---|
+| [../architectural-decisions/SKILL.md](../architectural-decisions/SKILL.md) | 1.5 |
+| [../code-style/SKILL.md](../code-style/SKILL.md) | 1.0 |
+| [../buggy-behavior/SKILL.md](../buggy-behavior/SKILL.md) | 2.0 |
+| [../comments/SKILL.md](../comments/SKILL.md) | 0.5 |
+| [../security/SKILL.md](../security/SKILL.md) | 2.0 |
+
+The domain weight is a multiplier applied to every finding in that domain — bugs and
+security defects cost more than style issues, comments cost less.
+
+This skill does **not** perform the rule-checking itself. Its job is to dispatch,
+gather, score, and format.
+
+## Step 3 — Reclassify and Score each skill
+
+Each skill starts at **0** points. Every finding **adds** points based on its severity,
+multiplied by the domain weight. There is no cap — the more findings (and the worse
+they are), the higher the score. **A higher score means worse code quality.** A clean
+domain with no findings scores 0.
+
+Base severity weights (before domain multiplier):
+
+- **High** = 15 pt
+- **Medium** = 4 pt
+- **Low** = 1 pt
+
+Per-finding points = base severity weight × domain weight.
+
+Per-domain score = sum of points across all findings in that domain.
+
+Final score = sum of all per-domain scores. No floor, no ceiling.
+
+Use these severity definitions when reviewing or reclassifying subagent findings:
+
+- **High** — the finding describes a concrete defect or design problem with serious impact,
+  such as a hot-path panic, data corruption or loss, broken backward compatibility,
+  incorrect admission or scheduling behavior, security boundary violation, or a public/API
+  semantic mismatch that is likely to mislead maintainers or users.
+- **Medium** — the finding describes a real issue with meaningful but narrower impact,
+  such as a realistic edge-case bug, feature-gate behavior leak, maintainability problem
+  likely to cause follow-up bugs, wrong high-volume log verbosity, or misleading naming or
+  comments that materially reduce readability.
+- **Low** — the finding describes a real but minor issue with limited impact, such as a
+  small convention drift, low-risk structural noise, isolated typo in important text, or
+  minor comment/style issue that is still worth cleaning up.
+
+Each finding could be mislabeled or misclassified, so you must reclassify each finding
+based on the actual impact shown by the diff and surrounding code, not only on the label
+returned by the subagent.
+
+## Step 4 — Produce the report
+
+Output the evaluation in this exact structure (skills and scores are examples; the set
+of skills is whatever appears in the table above):
+
+```
+## Code Evaluation Report
+
+**Commits**: <BaseCommit>..<HeadCommit>
+**Files changed**: (from --stat output)
+
+**Note**: Higher score = worse code. 0 means no findings in that domain.
+
+---
+
+### Skill: Code Style (weight <weight defined in Skills table>)
+**Domain score**: X pt  (sum of all findings below)
+**Findings**: (bullet list of specific observations, in accordance with the rules in the skill file)
+  - <High/Medium/Low> | +x pt | <Finding Title>: <Explanation>
+
+### Skill: Buggy Behavior (weight <weight defined in Skills table>)
+**Domain score**: 0 pt
+**Findings**:
+  - No findings
+
+### Skill: Comments (weight <weight defined in Skills table>)
+**Domain score**: 0 pt
+**Findings**:
+  - No findings
+
+### Skill: Architectural Decisions (weight <weight defined in Skills table>)
+**Domain score**: X pt
+**Findings**:
+  - <High/Medium/Low> | +x pt | <Finding Title>: <Explanation>
+
+### Skill: Security (weight <weight defined in Skills table>)
+**Domain score**: X pt
+**Findings**:
+  - <High/Medium/Low> | +x pt | <Finding Title>: <Explanation>
+
+---
+
+### Final Score: X pt
+(sum of all domain scores — higher is worse; 0 is a clean diff)
+
+---
+
+## Recommendations
+
+(One recommendation per issue found. Only include recommendations whose corresponding finding contributed points. Each recommendation must have all sections below. Recommendations MUST be ordered by severity: all **High** first, then all **Medium**, then all **Low**. Within the same severity, order by the skill order in the table above. Number recommendations sequentially across all skills after sorting by severity.)
+
+### Recommendation N: <short title>
+
+**Severity**: High | Medium | Low
+**Domain**: The skill domain this recommendation came from (e.g., Code Style, Buggy Behavior, Comments, Architectural Decisions, Security).
+**Findings**: The exact finding title(s) this recommendation addresses, copied from the Findings section above.
+**Problem**: Describe the specific issue in the diff.
+**Reason**: Explain why this is a problem — what goes wrong or degrades over time.
+**Solution**: Give a concrete, actionable fix. Where possible, show a before/after code snippet.
+**Locations**: List every place this issue occurs, one per line, in `filepath:line` form.
+```
+
+Rules:
+- The report format above MUST NOT be broken. Headings, ordering, and the four-section
+  recommendation layout are normative.
+- Recommendations come from the subagents — this skill aggregates them, renumbers them
+  globally, and omits any whose corresponding finding added 0 points.
+- Code evaluation MUST promote these principles: **Maintainability, Simplicity,
+  Backward Compatibility**.
--- a/cmd/experimental/skills/reviewer/code-style/SKILL.md
+++ b/cmd/experimental/skills/reviewer/code-style/SKILL.md
@@ -1,0 +1,80 @@
+---
+name: code-style
+description: Review a diff for naming precision, convention drift, reinvented helpers, log-verbosity mistakes, misaligned test names, and typos.
+license: Apache-2.0
+metadata:
+  copyright: The Kubernetes Authors
+---
+
+# Skill: Code Style
+
+Use this when reviewing a diff for naming, convention, and stylistic violations against
+the surrounding code. The goal is consistency with the existing package, not the
+reviewer's personal taste.
+
+## Rules
+
+Each rule in this domain is its own sub-skill. Scan the diff for violations of every
+rule below.
+
+| Rule | Triggers on |
+|---|---|
+| [imprecise-names](./imprecise-names/SKILL.md) | identifier whose name does not describe exactly what it contains |
+| [convention-drift](./convention-drift/SKILL.md) | new code that breaks naming conventions already established in the same file or package |
+| [reinvented-helpers](./reinvented-helpers/SKILL.md) | logic that duplicates an existing util/helper function instead of reusing it |
+| [wrong-log-verbosity](./wrong-log-verbosity/SKILL.md) | per-reconcile-cycle log lines emitted at `V(2)` |
+| [misaligned-test-names](./misaligned-test-names/SKILL.md) | test function names that do not reflect the function or behavior under test |
+| [code-style-typos](./code-style-typos/SKILL.md) | typos in identifiers, strings, or any text introduced by the diff |
+
+@./imprecise-names/SKILL.md
+@./convention-drift/SKILL.md
+@./reinvented-helpers/SKILL.md
+@./wrong-log-verbosity/SKILL.md
+@./misaligned-test-names/SKILL.md
+@./code-style-typos/SKILL.md
+
+## What not to include
+
+Do not report personal preferences or taste-based objections. This skill is about
+consistency, precision, and established local conventions, not about the reviewer's
+favorite naming style or formatting habits.
+
+In particular, do not flag:
+
+- code that is consistent with the surrounding file or package, even if you would have
+  named or structured it differently;
+- short helpers, direct expressions, or naming choices that are reasonable and accurate,
+  even if you would personally prefer a different phrasing;
+- harmless wording differences, minor prose preferences, or cosmetic style opinions with
+  no maintenance or readability impact.
+
+## How to report
+
+Cite the exact file and line, name the convention or rule being violated, and classify
+severity (high / medium / low). Naming-precision issues that introduce semantic
+ambiguity (a name that lies about its contents) are high. Convention drift and
+reinvented helpers are typically medium. Stray typos and minor wording choices are low.
+
+Return two sections:
+
+### Findings
+
+A bullet list, one per violation, in this format:
+
+```
+- <High/Medium/Low> | <Finding Title>: <one-sentence explanation grounded in the diff>
+```
+
+### Recommendations
+
+One recommendation per finding above, numbered sequentially within this skill. Each
+recommendation MUST contain all four sections below.
+
+```
+### Recommendation N: <short title>
+
+**Problem**: Describe the specific issue in the diff.
+**Reason**: Explain why this is a problem — what goes wrong or degrades over time.
+**Solution**: Give a concrete, actionable fix. Where possible, show a before/after code snippet.
+**Locations**: List every place this issue occurs, one per line, in `filepath:line` form.
+```
--- a/cmd/experimental/skills/reviewer/security/SKILL.md
+++ b/cmd/experimental/skills/reviewer/security/SKILL.md
@@ -1,0 +1,81 @@
+---
+name: security
+description: Review a diff for security issues exploitable by a tenant, neighbouring controller, or malicious workload spec — input validation, injection, DoS, nil-safety, authz, info disclosure, supply chain, webhook safety.
+license: Apache-2.0
+metadata:
+  copyright: The Kubernetes Authors
+---
+
+# Skill: Security
+
+Use this when reviewing a diff for code patterns exploitable by an authenticated
+low-privileged tenant, a compromised neighbouring controller, or a malicious workload
+spec. Kueue is a privileged in-cluster controller with broad watch permissions that
+also sits in the admission path via webhooks — one insecure pattern can crash-loop the
+batch-scheduling surface, leak tenant data, allow quota escape, or enable arbitrary
+code execution.
+
+## Rules
+
+Each rule in this domain is its own sub-skill. Scan the diff for violations of every
+rule below.
+
+| Rule | Triggers on |
+|---|---|
+| [input-validation](./input-validation/SKILL.md) | missing validation on user-settable CR fields, webhook payloads, labels, annotations, or untrusted deserialization |
+| [injection](./injection/SKILL.md) | exec/SSRF/template/query/selector built from user-supplied strings |
+| [path-traversal](./path-traversal/SKILL.md) | file paths or archive extraction without root containment |
+| [resource-bounds-dos](./resource-bounds-dos/SKILL.md) | uncapped loops/allocations, missing timeouts, reconciler-wedging input, unbounded metric labels |
+| [nil-safety](./nil-safety/SKILL.md) | unguarded dereferences from malformed CRs; goroutines without recover |
+| [authn-authz-relaxation](./authn-authz-relaxation/SKILL.md) | webhook token skipping, failurePolicy Ignore, wildcard RBAC, privileged pod specs |
+| [information-disclosure](./information-disclosure/SKILL.md) | full-object logging, request echoing, credentials in status/annotations, kubeconfig logging |
+| [supply-chain-hygiene](./supply-chain-hygiene/SKILL.md) | unpinned images, TLS bypass, unjustified replace directives, curl\|sh, moving-tag Actions |
+| [annotation-label-namespace-abuse](./annotation-label-namespace-abuse/SKILL.md) | generic annotation/label loops that copy or forward values |
+| [feature-gated-insecure-paths](./feature-gated-insecure-paths/SKILL.md) | gated behavior whose security assumptions fail when accidentally enabled |
+| [integration-adapter-trust-boundary](./integration-adapter-trust-boundary/SKILL.md) | adapters reading credential-like fields from third-party CRDs without treating them as untrusted |
+| [webhook-safety-regressions](./webhook-safety-regressions/SKILL.md) | loosened failurePolicy, shortened timeouts, TLS bypass, non-idempotent mutating webhooks |
+
+@./input-validation/SKILL.md
+@./injection/SKILL.md
+@./path-traversal/SKILL.md
+@./resource-bounds-dos/SKILL.md
+@./nil-safety/SKILL.md
+@./authn-authz-relaxation/SKILL.md
+@./information-disclosure/SKILL.md
+@./supply-chain-hygiene/SKILL.md
+@./annotation-label-namespace-abuse/SKILL.md
+@./feature-gated-insecure-paths/SKILL.md
+@./integration-adapter-trust-boundary/SKILL.md
+@./webhook-safety-regressions/SKILL.md
+
+## How to report
+
+Cite the exact file and line, name the rule violated and the relevant CWE/CVE where
+applicable, and classify severity (high / medium / low). Anything that crashes the
+controller, leaks credentials, bypasses authz, or enables injection is high. Missing
+caps, unbounded cardinality, or weakened supply-chain controls are typically medium.
+Stylistic security smells with no concrete exploit path are low.
+
+Return two sections:
+
+### Findings
+
+A bullet list, one per violation, in this format:
+
+```
+- <High/Medium/Low> | <Finding Title>: <one-sentence explanation grounded in the diff, with CWE/CVE where applicable>
+```
+
+### Recommendations
+
+One recommendation per finding above, numbered sequentially within this skill. Each
+recommendation MUST contain all four sections below.
+
+```
+### Recommendation N: <short title>
+
+**Problem**: Describe the specific issue in the diff.
+**Reason**: Explain why this is a problem — what goes wrong or degrades over time.
+**Solution**: Give a concrete, actionable fix. Where possible, show a before/after code snippet.
+**Locations**: List every place this issue occurs, one per line, in `filepath:line` form.
+```
--- a/cmd/experimental/skills/README.md
+++ b/cmd/experimental/skills/README.md
@@ -6,7 +6,7 @@
 | [kueue-lineage](kueue-lineage/SKILL.md) | "what pods are running for my workload", "trace workload to pods", "show me the jobs for this workload", lineage/ownership questions |
 | [kueue-flake-debugger](kueue-flake-debugger/SKILL.md) | "debug a flake", "investigate test failure", "test timed out", "CI flake" |
 | [kueue-release-notes](kueue-release-notes/SKILL.md) | Review or propose a concise PR release note focused on the user-observable change. |
-| [reviewer/](reviewer/README.md) | "review this PR", "review this code", any code review of kueue changes |
+| [reviewer/](reviewer/README.md) | "review this PR", "review this code", "evaluate this diff", "score these commits", "review the changes between", "code quality evaluation", any code review of kueue changes |
 
 ---
 
--- a/cmd/experimental/skills/reviewer/README.md
+++ b/cmd/experimental/skills/reviewer/README.md
@@ -23,6 +23,46 @@ Each skill is independent and can be applied in parallel with the others. When r
 | [api-field-comments](api-field-comments/SKILL.md) | new CRD field missing godoc comment or enum value descriptions |
 | [race-conditions](race-conditions/SKILL.md) | snapshot/mirror field added to protect a reader from a concurrent writer |
 | [visibility-exports](visibility-exports/SKILL.md) | new exported function or method with no callers outside the package |
+| [code-eval](code-eval/SKILL.md) | runs code evaluation and produces a scored report |
+| [architectural-decisions](architectural-decisions/SKILL.md) | domain — unjustified complexity, duplicated logic across types/adapters, scope creep, or misplaced code (see ToC inside) |
+| [buggy-behavior](buggy-behavior/SKILL.md) | domain — logic errors, broken edge cases, feature-gate bugs, deletions that break backwards compatibility during rolling upgrades (see ToC inside) |
+| [code-style](code-style/SKILL.md) | domain — imprecise names, convention drift, reinvented helpers, wrong log verbosity, misaligned test names, typos (see ToC inside) |
+| [comments](comments/SKILL.md) | domain — over-commenting, what-not-why narration, inaccurate comments, missing TODOs on compatibility shims (see ToC inside) |
+| [security](security/SKILL.md) | domain — input-validation gaps, injection, DoS, nil-safety crashes, authz relaxation, info disclosure, supply-chain weakening, webhook safety regressions (see ToC inside) |
+| [imprecise-names](code-style/imprecise-names/SKILL.md) | identifier whose name does not describe exactly what it contains (code-style) |
+| [convention-drift](code-style/convention-drift/SKILL.md) | new code that breaks naming conventions already established in the same file or package (code-style) |
+| [reinvented-helpers](code-style/reinvented-helpers/SKILL.md) | logic that duplicates an existing util/helper function instead of reusing it (code-style) |
+| [wrong-log-verbosity](code-style/wrong-log-verbosity/SKILL.md) | per-reconcile-cycle log lines emitted at `V(2)` (code-style) |
+| [misaligned-test-names](code-style/misaligned-test-names/SKILL.md) | test function names that do not reflect the function or behavior under test (code-style) |
+| [code-style-typos](code-style/code-style-typos/SKILL.md) | typos in identifiers, strings, or any text introduced by the diff (code-style) |
+| [illogical-structure](architectural-decisions/illogical-structure/SKILL.md) | code a future maintainer will struggle to follow, modify, or extend (architectural-decisions) |
+| [nonsensical-decisions](architectural-decisions/nonsensical-decisions/SKILL.md) | unnecessary indirection, mismatched abstractions, confusing data flow (architectural-decisions) |
+| [avoidable-complexity](architectural-decisions/avoidable-complexity/SKILL.md) | solutions more elaborate than the problem requires (architectural-decisions) |
+| [pointless-intermediate-variables](architectural-decisions/pointless-intermediate-variables/SKILL.md) | redundant local variables that add noise without clarity (architectural-decisions) |
+| [duplicated-logic](architectural-decisions/duplicated-logic/SKILL.md) | identical blocks across types/adapters/call sites that should be shared — or new helpers with one caller (architectural-decisions) |
+| [scope-creep](architectural-decisions/scope-creep/SKILL.md) | diffs bundling a bugfix with an unrelated refactor, or over-generalizing a change (architectural-decisions) |
+| [misplaced-logic](architectural-decisions/misplaced-logic/SKILL.md) | code placed somewhere a reader would not expect to find it (architectural-decisions) |
+| [logic-errors](buggy-behavior/logic-errors/SKILL.md) | incorrect conditionals, inverted checks, off-by-one, unhandled edge cases, races, mishandled errors (buggy-behavior) |
+| [deleted-backwards-compatibility-code](buggy-behavior/deleted-backwards-compatibility-code/SKILL.md) | removal or relocation of code that reconciles state owned by previous controller versions (buggy-behavior) |
+| [feature-gate-interaction-bugs](buggy-behavior/feature-gate-interaction-bugs/SKILL.md) | gated behavior whose gate-off path is untested or incorrect (buggy-behavior) |
+| [unnecessary-guard-conditions](buggy-behavior/unnecessary-guard-conditions/SKILL.md) | extra checks that are logically unreachable at the call site (buggy-behavior) |
+| [over-commenting](comments/over-commenting/SKILL.md) | comments that explain what self-documenting code already says (comments) |
+| [wrong-kind-of-comment](comments/wrong-kind-of-comment/SKILL.md) | comments describing *what* instead of *why* (comments) |
+| [inaccurate-comments](comments/inaccurate-comments/SKILL.md) | comments or docstrings that no longer match the code (comments) |
+| [missing-deferred-removal-markers](comments/missing-deferred-removal-markers/SKILL.md) | compatibility shim kept without a deferred-removal comment (comments) |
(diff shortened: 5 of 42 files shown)


cmd/experimental/agent/skills/code-eval.md
```diff
@@ -0,0 +1,163 @@
+# Skill: Code Evaluation
+
+Evaluate the quality of a git diff between two commits. Scores the changes across four domains — Code Style, Buggy Behavior, Comments, and Small Architectural Decisions — then provides actionable recommendations for improvement.
+
+Use this when a user asks to evaluate, score, or review a diff between two commits.
+
+## Arguments
```
amy: Add a prompt where when the user says something like "perform a code eval", that the agent tells you what arguments it needs.

cmd/experimental/agent/skills/code-eval.md
```diff
@@ -0,0 +1,163 @@
...
+- Log verbosity conventions (what V-level is used for recurring vs lifecycle events)
+- Whether any deleted code might be needed during a rolling upgrade or version skew
+
+## Step 3 — Evaluate across domains
+
+Score the Diff Code on each domain below. Be strict but fair. Every point deduction must be noted so the user understands what happened.
+
+---
+
+### Domain 1: Code Style (0–50 pts)
```
amy: My overall feedback is that you need to be aware of:
1. context size
2. parallelizable tasks

So for instance, each review should be in its own file. As an example: https://github.com/kubernetes-sigs/kueue/pull/10819/changes -> I'm going to make a review folder, then dump each number into its own file. This ensures that the agent can spin up in parallel & close each context when its done with it.

amy: So here, each domain can be a file that is a table of contents that points to each rule file.

cmd/experimental/agent/skills/code-eval.md
```diff
@@ -0,0 +1,163 @@
...
+(One recommendation per issue found. Only include recommendations where the score was reduced. Each recommendation must have all three sections below.)
+
+### Recommendation N: <short title>
+
+**Problem**: Describe the specific issue in the diff.
+**Reason**: Explain why this is a problem — what goes wrong or degrades over time.
+**Solution**: Give a concrete, actionable fix. Where possible, show a before/after code snippet.
+```
+
+## Principles
```
amy: I'd move this to the top.

amy:
@reruno This is really cool! Thank you! I would be interested in.... how scoring works / how to trust the heuristic like if we're going to use it to evaluate skills. How do ml scientists usually do evals. Maybe there's room for collaboration there.

reruno:
@amy Thanks for the feedback, I really appreciate it! 

> how scoring works / how to trust the heuristic like if we're going to use it to evaluate skills. How do ml scientists usually do evals. Maybe there's room for collaboration there.

https://www.anthropic.com/engineering/demystifying-evals-for-ai-agents
Answering you question, this is very good blog post about evaluating ai agents performance. 

IIUC to grade LLM output we have 3 options
1. Code-based graders (go lint, make verify)
2. Model-based graders (LLM evaluates LLM)
3. Human graders (actual code reviewers)

As of now IMO large chunk of the code is written by LLMs and changed a bit by PR authors to be less ChatGPTish, thus we can think of PR code as LLM output. Code Eval is Model-based grader, which takes LLM output (Diff Code) and grades it according to the rules in the SKILL. Code-based graders are already implemented (make verify), Human graders also present (code reviewers), the only thing missing right now is Model-based grader. 

I am still new to these concepts and I will definitely want to deepen my knowledge how to make grader properly, but I think code-eval is a move in the right direction.

cmd/experimental/agent/skills/code-eval.md
```diff
@@ -0,0 +1,163 @@
...
+## Code Evaluation Report
+
+**Commits**: <BaseCommit>..<HeadCommit>
+**Files changed**: (from --stat output)
+
+---
+
+### Domain 1: Code Style
+**Score**: X/50
+**Findings**: (bullet list of specific observations, in accordance with the specified rules for findings)
```
kshalot: This is likely model and chain-of-thought dependent, but I think it might be generally better to first output the findings and then output the score, to make sure that the score follows from the findings and not the other way around.

reruno: I think state of the art models with thinking chains are not that vulnerable to this kind of errors, also model calculates the Score based on findings. Model assumes diff code is already at max score, and when models find a violation of rule it deducts from the score defined amount of points, so score pretty much depends on number and severity of findings.  Probably for more budget models it will be good change. 

Also IMO having score at the top is just more convenient for an eye  :)

cmd/experimental/agent/skills/code-eval.md
reruno: https://github.com/kubernetes-sigs/kueue/pull/10819#issuecomment-4415936010

As here was mentioned, I will split domains into different files, when the PR is merged

cmd/experimental/agent/skills/code-eval.md
```diff
@@ -0,0 +1,163 @@
+# Skill: Code Evaluation
+
+Evaluate the quality of a git diff between two commits. Scores the changes across four domains — Code Style, Buggy Behavior, Comments, and Small Architectural Decisions — then provides actionable recommendations for improvement.
+
+Use this when a user asks to evaluate, score, or review a diff between two commits.
```
j-skiba: I think we should use frontmatter here - https://agentskills.io/specification#frontmatter

reruno:
New runs after force push are in special notes to reviewer

mimowo:
ping @amy or @PBundyra ptal

amy:
> ping @/amy or @/PBundyra ptal

@mimowo, @reruno and I discussed that we will wait for this one to merge first: https://github.com/kubernetes-sigs/kueue/pull/10819

And then reruno will break up files and use similar structure.

reruno:
I recommend to read this blog post about AI and cyber security https://www.jefftk.com/p/ai-is-breaking-two-vulnerability-cultures

TLDR: AI managed to find zero days vulnerabilities in Linux kernel. Taking that into account I decided to add security.md domain. AI can pick up obvious bugs and inconsistencies, and that could help fix potential vulnerabilities and help with hardening at PR stage.

@amy I split up the files, can you take a look at the new security.md file also pls

amy:
@reruno 

Oooo security skill is exciting! Will take a pass. 

To clarify more on directory structure. Your code eval skill is sort of stand alone. Where each rule for instance can also be its own standalone skill to parallelize. I would also add instructions to the skill to parallelize. (maybe add a cap for the number of threads)

What we could do is this directory structure:
reviewer: 
- eval 
  - README.md
  - SKILL.md -> points to architectural/code style/comments README.md
- architectural
  - README.md  -> table of contents for all the skills in architectural
  - <rule 1 name>
    - SKILL.md 
- code style
- comments
- ...etc some other category

Example for table of contents: https://github.com/kubernetes-sigs/kueue/pull/10819/changes#diff-8d980b2fb04049af1d5327d8042808ecc7922e559b3d893e54fe9b615e4a6c64

cmd/experimental/skills/reviewer/code-style/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+
+Use this when reviewing a diff for naming, convention, and stylistic violations against
+the surrounding code. The goal is consistency with the existing package, not the
+reviewer's personal taste.
+
+## What to look for
+
+Scan the diff and flag any of the following:
+
+- **Imprecise names** — every new identifier must describe exactly what it contains. A
```
amy: Can you split these out into skill files too? 

So if you look at the current reviewer folder, you'll see that each review bullet is its own skill. 

So as an example for code-style, you'll want... 
Code-style README to point to a table of contents of each reviewer skill (so imprecise names is its own skill, wrapper method is its own skill, etc.)

Let me know if that makes sense!

cmd/experimental/skills/reviewer/code-eval/SKILL.md
```diff
@@ -0,0 +1,172 @@
...
+5. Return findings AND recommendations in the format defined in its skill file.
+
+Skills:
+| Skill | Max score |
+|---|---|
+| [../architectural-decisions/SKILL.md](../architectural-decisions/SKILL.md) | 20 |
+| [../code-style/SKILL.md](../code-style/SKILL.md) | 40 |
+| [../buggy-behavior/SKILL.md](../buggy-behavior/SKILL.md) | 20 |
+| [../comments/SKILL.md](../comments/SKILL.md) | 5 |
+| [../security/SKILL.md](../security/SKILL.md) | 20 |
```
amy: These add up to 105 not 100. Maybe say that you're normalizing the percentage.

cmd/experimental/skills/reviewer/code-eval/SKILL.md
```diff
@@ -0,0 +1,172 @@
...
+
+This skill does **not** perform the rule-checking itself. Its job is to dispatch,
+gather, score, and format.
+
+## Step 3 — Reclasify and Score each skill
+
+Each skill starts at its max score. Deduct based on the severity of each finding:
```
amy: You might want to make the deductions proporional instead of these flat points. Ex: a Low in Comments hurts the final score as much as a Low in Code Style

cmd/experimental/skills/reviewer/security/SKILL.md
```diff
@@ -0,0 +1,167 @@
...
+- New direct dependencies without a one-line justification.
+
+### 9. Annotation/label namespace abuse
+- Generic loops over `obj.Annotations` that copy, forward, or echo values — these
+  allow tenant injection into the reserved `kueue.x-k8s.io/` namespace and leak
+  operator-set values across boundaries. Forwarding between CRs must go through an
+  explicit allowlist.
+
+### 10. Feature-gated insecure paths
+- Behavior added behind a [[feature-gates|feature gate]] whose security assumptions
```
amy: I don't think this is markdown: `[[feature-gates|feature gate]]`

cmd/experimental/skills/reviewer/security/SKILL.md
```diff
@@ -0,0 +1,167 @@
...
+  (`/etc/kubernetes`, `/var/run/docker.sock`), or dropping `runAsNonRoot: true`.
+
+### 7. Information disclosure (CWE-200, CWE-522, CWE-532, CWE-552)
+- Full Kubernetes objects logged with `%+v` / `%#v`. Required: `klog.KObj(obj)` plus
+  structured key/value pairs.
+- Webhook responses, events, or status fields that echo request bodies, stack traces,
+  or credential-bearing fields.
+- Credentials or tokens persisted in annotations, labels, env vars, or status —
+  credentials belong in mounted Secrets.
+- [[multikueue]] kubeconfigs logged, echoed in status, or written with
```
amy: I don't think this is markdown: `[[multikueue]] `

cmd/experimental/skills/reviewer/code-eval/SKILL.md
reruno: I specifically make this not to add up to 100 because I calculate percentage in case if we would want to add new domain with 20pt

cmd/experimental/skills/reviewer/code-eval/SKILL.md
reruno: I am thinking about system where you collect points for each violation, and more points means worse findings

cmd/experimental/skills/reviewer/code-style/SKILL.md
reruno: I split all the files into each skill, but I left domain skills like code-style, security, that gather all skills related to code-style or security. 

And code evaluation collects all domains and assigns multiplier to point system. 

wdyt?

amy: @reruno I think we're getting close. Can you put the code-style stuff into one directory? 
So top level categories are:

- Eval/ <- references security, style, etc.
- Security/
- Style/
  - top level style skill
  - typo dir/
  - comments dir/ 
  - etc.  
- Architectural/
- etc... 

After that. Good to approve! The reason I want it organized like this is so that it makes future PRs easier to organize in the correct sections. Also this way your eval skill can always be updated with the full breadth of new PRs to each section.

reruno: Yeah I misinterpreted, updated the file structure

amy: noice

amy:
/approve
/lgtm
/hold 

@kannon92 @PBundyra Looks good to me if y'all wanna take a pass. I'll remove the hold wed (5/27) next week though if no comments. Wanna make sure we merge this because its been sitting for a bit.

amy:
Alrighty. @reruno after you rebase ready to release hold and approve! thanks for your patience.

reruno:
@amy rebased on upstream, let me know if you have any additional comments

PR #11095, opened 2026-05-10
fix: Add e2e test for simple migration scenarios

### LINKED ISSUE #10988: [ConcurrentAdmission] Add E2E tests for simple migration scenarios
E2E tests that will check simple paths of the Concurrent Admission feature - Job creates a correct number of Workloads and migration from less favorable to more favorable Variant works as expected

**Why is this needed**:

**Completion requirements**:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

gyliu513:
#### What type of PR is this?
/kind bug
/area testing

/cc @PBundyra 

Fixes https://github.com/kubernetes-sigs/kueue/issues/10988

```release-note
NONE
```

### DIFF
--- a/test/e2e/singlecluster/concurrent_admission_test.go
+++ b/test/e2e/singlecluster/concurrent_admission_test.go
@@ -1,0 +1,300 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package singlecluster
+
+import (
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+	controllerconstants "sigs.k8s.io/kueue/pkg/controller/constants"
+	workloadjob "sigs.k8s.io/kueue/pkg/controller/jobs/job"
+	utiltestingapi "sigs.k8s.io/kueue/pkg/util/testing/v1beta2"
+	testingjob "sigs.k8s.io/kueue/pkg/util/testingjobs/job"
+	"sigs.k8s.io/kueue/pkg/workload"
+	"sigs.k8s.io/kueue/test/util"
+)
+
+var _ = ginkgo.Describe("ConcurrentAdmission", ginkgo.Label("area:singlecluster", "feature:concurrent-admission"), func() {
+	var ns *corev1.Namespace
+
+	ginkgo.BeforeEach(func() {
+		ns = util.CreateNamespaceFromPrefixWithLog(ctx, k8sClient, "e2e-concurrent-")
+	})
+	ginkgo.AfterEach(func() {
+		gomega.Expect(util.DeleteNamespace(ctx, k8sClient, ns)).To(gomega.Succeed())
+		util.ExpectAllPodsInNamespaceDeleted(ctx, k8sClient, ns)
+	})
+
+	ginkgo.When("Job is submitted to a ClusterQueue with ConcurrentAdmissionPolicy", func() {
+		var (
+			reservationRF     *kueue.ResourceFlavor
+			spotRF            *kueue.ResourceFlavor
+			cq                *kueue.ClusterQueue
+			lq                *kueue.LocalQueue
+			reservationFlavor string
+			spotFlavor        string
+		)
+
+		ginkgo.BeforeEach(func() {
+			reservationFlavor = "reservation-" + ns.Name
+			spotFlavor = "spot-" + ns.Name
+
+			reservationRF = utiltestingapi.MakeResourceFlavor(reservationFlavor).
+				NodeLabel("instance-type", "reservation").Obj()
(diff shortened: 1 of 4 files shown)


mimowo:
Thanks 👍

test/e2e/singlecluster/concurrent_admission_test.go
```diff
@@ -0,0 +1,300 @@
...
+							}
+						}
+					}
+					g.Expect(variantCount).To(gomega.Equal(2))
+				}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
+			})
+		})
+	})
+
+	ginkgo.When("Migration from less favorable to more favorable flavor", func() {
```
PBundyra: Could we also check if Pods have the correct `instance-type` node selector? Just to ensure it works e2e

PBundyra: Does this `When` block differ from the first one when it comes to created objects? If not is there any blocker from merging it into one?

PBundyra (commented):
Thanks @gyliu513 it's heading towards the right direction!

test/e2e/singlecluster/concurrent_admission_test.go
gyliu513: Good point! Added a step to verify that Pods have the correct `instance-type` node selector matching the admitted flavor — dynamically derived from the parent workload's admission status so the check is not hardcoded to a specific flavor.

gyliu513: Good suggestion! The two `When` blocks differed only in their initial CQ quota — the migration test needed reservation to start at 0 to force initial admission on spot. I've merged them into a single `When` block sharing the same `BeforeEach`/`AfterEach`, and moved the quota reduction to the beginning of the migration `It` test as its own `By` step.

test/e2e/singlecluster/concurrent_admission_test.go
```diff
@@ -0,0 +1,281 @@
...
+					for _, wl := range wlList.Items {
+						if wl.Name == parentWl.Name {
+							continue
+						}
+						for _, ref := range wl.OwnerReferences {
+							if ref.UID == parentWl.UID {
+								variantCount++
+							}
+						}
+					}
```
mimowo: Please introduce a helper function like `findVariantsForParent()` which returns the list of variants. Then assert the list has the specified length with gomega.HasLen. THis produces much more informative output when the test fails.

test/e2e/singlecluster/concurrent_admission_test.go
```diff
@@ -0,0 +1,281 @@
...
+					g.Expect(k8sClient.Get(ctx, parentWlKey, &parentWl)).To(gomega.Succeed())
+					g.Expect(workload.IsAdmitted(&parentWl)).To(gomega.BeTrue())
+					g.Expect(parentWl.Status.Admission).ToNot(gomega.BeNil())
+					g.Expect(parentWl.Status.Admission.PodSetAssignments).ToNot(gomega.BeEmpty())
+					admittedFlavor := string(parentWl.Status.Admission.PodSetAssignments[0].Flavors[corev1.ResourceCPU])
+
+					expectedInstanceType := "reservation"
+					if admittedFlavor == spotFlavor {
+						expectedInstanceType = "spot"
+					}
```
mimowo: Is the admittedFlavor non determininitic here? It would really be great to write the tests in such a fashion that the migration is deterministic, otherwise we may get many flakes. 
We may consider non-deterministic tests "day two" , but they go beyond "simple migration scenarios"

test/e2e/singlecluster/concurrent_admission_test.go
```diff
@@ -0,0 +1,281 @@
...
+					spotVariant := e2eGetVariantByFlavor(wlList.Items, parentWlKey.Name, spotFlavor)
+					g.Expect(spotVariant).ToNot(gomega.BeNil(), "spot variant not found")
+					g.Expect(ptr.Deref(spotVariant.Spec.Active, true)).To(gomega.BeFalse(), "spot variant should be deactivated after migration")
+				}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
+			})
+		})
+	})
+})
+
+func e2eGetVariantByFlavor(workloads []kueue.Workload, parentName string, flavor string) *kueue.Workload {
```
mimowo: no need for the `e2e` prefix - for one we know we are in e2e tests, second there is nothing specific about e2e tests in the body of this function

test/e2e/singlecluster/concurrent_admission_test.go
gyliu513: Good point,  I'll add a findVariantsForParent() helper and update the assertion.

test/e2e/singlecluster/concurrent_admission_test.go
gyliu513: Yes, the admittedFlavor in that check is non-deterministic since both flavors have equal quota and either could be admitted first. I'll remove the pod node-selector check from this first test (its purpose is just to verify variants are created, not to verify migration behavior). 

The deterministic migration scenario is already covered by the second It block which explicitly forces spot admission by zeroing reservation quota.

test/e2e/singlecluster/concurrent_admission_test.go
gyliu513: Done

test/e2e/singlecluster/concurrent_admission_test.go
mimowo: I see, but quite clearly the non-determinism is not the intention of the test by the name "Verifying one Variant Workload is created per flavor (2 variants + 1 parent = 3 total)"

That could be asserted with determinism, or without any admission even. So we need to either change the name of the test to express the scenario intention, or find a way to test the intention without non-determinism.

mimowo: but I would generally suggest to scope this PR to only deterministic scenarios as the name says "simple migration scenarios"

mimowo:
/test pull-kueue-test-e2e-sequential-baseline-main
retry to get more evidence that there are no flakes

test/e2e/singlecluster/concurrent_admission_test.go
PBundyra: Could you share a link to the step you mentioned?

mimowo:
/test pull-kueue-test-e2e-sequential-baseline-main
retry 2

mimowo:
/lgtm
/approve
Looks like the tests are stable, let me merge 👍

PR #11097, opened 2026-05-11
Graduate SchedulingEquivalenceHashing to beta (default to true)

### LINKED ISSUE #10005: Graduate SchedulingEquivalenceHashing to Beta

Graduate SchedulingEquivalenceHashing to Beta. 

That was originally the plan for the bugfix in https://github.com/kubernetes-sigs/kueue/pull/10001. but we encountered some complications with :
1. MultiKueue preemption gates mechanism requiring adjustment
2. Performance regression tests requiring adjustment (lowering the requeue interval to 100ms helps)
3. many unit tests failing requiring adjustments

**Why is this needed**:

This is a very useful performance bugfix for large scale systems, where currently Kueue scheduler wastes
a lot of cycles processing identical (from the scheduling perspective) workloads.

sohankunkerkar:
Narrow equivalence hashing to NoFit-only by introducing explicit requeue reasons and deriving a hashable flag from the reason instead of the broad non-immediate condition.

#### What type of PR is this?

/kind feature

Fixes https://github.com/kubernetes-sigs/kueue/issues/10005

```release-note
Scheduling: re-enable SchedulingEquivalenceHashing by default, narrowed to NoFit-only to avoid incorrectly bulk-moving namespace-mismatched or preemption-gated workloads.
```

### DIFF
--- a/pkg/cache/queue/cluster_queue.go
+++ b/pkg/cache/queue/cluster_queue.go
@@ -55,13 +55,15 @@ import (
 type RequeueReason string
 
 const (
-	RequeueReasonFailedAfterNomination RequeueReason = "FailedAfterNomination"
-	RequeueReasonNamespaceMismatch     RequeueReason = "NamespaceMismatch"
-	RequeueReasonGeneric               RequeueReason = ""
-	RequeueReasonPreemptionGated       RequeueReason = "PreemptionGated"
-	RequeueReasonPendingPreemption     RequeueReason = "PendingPreemption"
-	RequeueReasonPendingMigration      RequeueReason = "PendingMigration"
-	RequeueReasonPreemptionFailed      RequeueReason = "PreemptionFailed"
+	RequeueReasonFailedAfterNomination  RequeueReason = "FailedAfterNomination"
+	RequeueReasonNamespaceMismatch      RequeueReason = "NamespaceMismatch"
+	RequeueReasonGeneric                RequeueReason = ""
+	RequeueReasonPreemptionGated        RequeueReason = "PreemptionGated"
+	RequeueReasonPendingPreemption      RequeueReason = "PendingPreemption"
+	RequeueReasonPendingMigration       RequeueReason = "PendingMigration"
+	RequeueReasonPreemptionFailed       RequeueReason = "PreemptionFailed"
+	RequeueReasonNoFit                  RequeueReason = "NoFit"
+	RequeueReasonPreemptionNoCandidates RequeueReason = "PreemptionNoCandidates"
 )
 
 var (
@@ -106,8 +108,8 @@ type ClusterQueue struct {
 	// inadmissibleWorkloads are workloads that have been tried at least once and couldn't be admitted.
 	inadmissibleWorkloads inadmissibleWorkloads
 
-	// noFitSchedulingHashes tracks scheduling equivalence classes that received NoFit.
-	// Cleared when queueInadmissibleWorkloads runs.
+	// noFitSchedulingHashes tracks scheduling equivalence classes that received
+	// NoFit or PreemptionNoCandidates. Cleared when queueInadmissibleWorkloads runs.
 	noFitSchedulingHashes sets.Set[string]
 
 	finishedWorkloads sets.Set[workload.Reference]
@@ -494,7 +496,10 @@ func (c *ClusterQueue) DeleteFromLocalQueue(log logr.Logger, q *LocalQueue, role
 // or if there was a call to QueueInadmissibleWorkloads after a call to Pop,
 // the workload will be pushed back to heap directly. Otherwise, the workload
 // will be put into the inadmissibleWorkloads.
-func (c *ClusterQueue) requeueIfNotPresent(log logr.Logger, wInfo *workload.Info, immediate bool) bool {
+// When SchedulingEquivalenceHashing is enabled and the reason is NoFit or
+// PreemptionNoCandidates, equivalent workloads in the heap are bulk-moved
+// to inadmissible.
+func (c *ClusterQueue) requeueIfNotPresent(log logr.Logger, wInfo *workload.Info, immediate bool, reason RequeueReason) bool {
 	c.rwm.Lock()
 	defer c.rwm.Unlock()
 	key := workload.Key(wInfo.Obj)
@@ -537,7 +542,8 @@ func (c *ClusterQueue) requeueIfNotPresent(log logr.Logger, wInfo *workload.Info
 	}
 	log.V(2).Info(logMsg, "clusterQueue", c.name, "workload", key)
 
-	if features.Enabled(features.SchedulingEquivalenceHashing) && wInfo.SchedulingHash != workload.SchedulingHashUnknown && !immediate {
+	if features.Enabled(features.SchedulingEquivalenceHashing) && wInfo.SchedulingHash != workload.SchedulingHashUnknown &&
+		(reason == RequeueReasonNoFit || reason == RequeueReasonPreemptionNoCandidates) {
 		if moved := c.handleInadmissibleHash(wInfo.SchedulingHash); moved > 0 {
 			log.V(2).Info("Bulk-moved equivalent workloads to inadmissible", "hash", wInfo.SchedulingHash, "movedCount", moved)
 		}
@@ -869,7 +875,7 @@ func (c *ClusterQueue) RequeueIfNotPresent(ctx context.Context, wInfo *workload.
 			reason == RequeueReasonPendingMigration ||
 			reason == RequeueReasonPreemptionFailed
 	}
-	return c.requeueIfNotPresent(log, wInfo, immediate)
+	return c.requeueIfNotPresent(log, wInfo, immediate, reason)
 }
 
 // baseCompareFunc orders workloads by sticky status, priority, timestamp, and UID.
--- a/test/integration/singlecluster/scheduler/inadmissible/equivalence_test.go
+++ b/test/integration/singlecluster/scheduler/inadmissible/equivalence_test.go
@@ -22,6 +22,7 @@ import (
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
 	utiltestingapi "sigs.k8s.io/kueue/pkg/util/testing/v1beta2"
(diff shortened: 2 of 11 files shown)


pkg/cache/queue/cluster_queue.go
```diff
@@ -494,7 +496,7 @@ func (c *ClusterQueue) DeleteFromLocalQueue(log logr.Logger, q *LocalQueue, role
 // or if there was a call to QueueInadmissibleWorkloads after a call to Pop,
 // the workload will be pushed back to heap directly. Otherwise, the workload
 // will be put into the inadmissibleWorkloads.
-func (c *ClusterQueue) requeueIfNotPresent(log logr.Logger, wInfo *workload.Info, immediate bool) bool {
+func (c *ClusterQueue) requeueIfNotPresent(log logr.Logger, wInfo *workload.Info, immediate bool, recordNoFitHash bool) bool {
```
mimowo: /hold
Narrowing the feature down to NoFit would be a severe regression for our users. That was the main reason we had to improve it with this PR: https://github.com/kubernetes-sigs/kueue/pull/10001
In this comment I say that it should be ok for our users to narrow it down to NoFit + Preempt noPreemptionCandidates: https://github.com/kubernetes-sigs/kueue/issues/10005#issuecomment-4175161601

From the code perspective I think we may choose to pass the reason down to requeueIfNotPresent rather than just bool, so that we can choose what to do later

test/integration/singlecluster/scheduler/inadmissible/equivalence_test.go
```diff
@@ -83,5 +84,45 @@ var _ = ginkgo.Describe("Scheduler", func() {
 			ginkgo.By("verifying the fitting workload gets admitted")
 			util.ExpectWorkloadsToHaveQuotaReservation(ctx, k8sClient, cq.Name, fitWl)
 		})
+
+		ginkgo.It("Should not hash workloads that don't fit due to namespace mismatch", func() {
```
mimowo, line 88: I would like to also add an integration test showing what happens when we have Preempt noPreemptionCandidates, see related: https://github.com/kubernetes-sigs/kueue/pull/11097/changes#r3218161372

sohankunkerkar: Added here: https://github.com/kubernetes-sigs/kueue/pull/11097/changes#diff-cc8aebc1a6a7ddf28563b3f878deabf3d3c9fe834f6f76eee92a3207e56a6d0eR128

test/performance/scheduler/configs/baseline/rangespec.yaml
```diff
@@ -9,12 +9,15 @@ cmd:
   maxWallMs: 425_000
 
 clusterQueueClassesMinUsage:
```
sohankunkerkar, line 11: I ran the perf benchmark locally 5 times with hash enabled to calibrate the thresholds:
                                                                                                                                                                                                                   
   | Run | large avgTimeToAdmissionMs | CQ cpuAverageUsage |                                                                                                                                                        
  | --- | --- | --- |
  | 1 | 18335 | 55.4% |                                                                                                                                                                                            
  | 2 | 16988 | couldn't capture the data here |                                   
  | 3 | 14179 | couldn't capture the data here |                                                                                                                                                                                                
  | 4 | 8283 | 56.6% |                                
  | 5 | 16222 | 55.8% |
 
Large workload avg comes out to ~14801ms (was 6666ms with hash off). This is expected since each large workload takes the full CQ quota (request=20, nominal=20), so all equivalents get NoFit and are bulk-moved to inadmissible. They wait for the 1s requeue batch period after capacity frees up. CQ usage drops slightly to ~55.9% (was 58.7%) for the same reason.   Updated the thresholds accordingly: large 11000→23000, CQ usage 55→50%.

test/integration/singlecluster/scheduler/scheduler_test.go
```diff
@@ -632,12 +632,10 @@ var _ = ginkgo.Describe("Scheduler", func() {
 
 				util.UnholdClusterQueue(ctx, k8sClient, prodClusterQ)
 
-				ginkgo.By("checking the workloads with lower priority do not get admitted")
-				util.ExpectWorkloadsToBePending(ctx, k8sClient, wlLow, wlMid1, wlMid2)
```
sohankunkerkar, line 636: Removed this assertion for lower-priority workloads in Hold CQ/LQ and two-small tests. `ExpectPendingWorkloadsMetric` on the next line validates the same invariant. Hold CQ/LQ workloads can get  `Inadmissible` instead of `Pending` because the scheduler processes faster with hash enabled, widening the stale-cache window where the workload controller still sees the CQ/LQ as inactive. In the two-small test, `smallWl2` is bulk-moved by the hash before individual evaluation so it has no condition at all.

test/integration/singlecluster/scheduler/inadmissible/equivalence_test.go
```diff
@@ -83,5 +84,77 @@ var _ = ginkgo.Describe("Scheduler", func() {
...
+
+			ginkgo.By("creating identical workloads that need preemption but have no lower-priority targets")
+			for i := range 10 {
+				util.MustCreate(ctx, k8sClient, utiltestingapi.MakeWorkload(fmt.Sprintf("blocked-%d", i), ns.Name).
+					Queue(kueue.LocalQueueName(queue.Name)).
+					Request(corev1.ResourceCPU, "1").Obj())
+			}
+
+			ginkgo.By("verifying equivalent blocked workloads are bulk-moved to inadmissible")
+			util.ExpectPendingWorkloadsMetric(cq, 0, 10)
```
mimowo, line 159: Is this assert failing without the code changes? I'm wondering because iiuc this could also happen without the code changes if scheduler can check all workloads in-between requeue events. Probably this is why 10 workloads is created, but even then I'm not sure it is enough.

At least it warrants a comment about the testing strategy, and some experimental results suggesting how many workloads is enough in practice to trigger the failure if not for the code changes

test/performance/scheduler/configs/baseline/rangespec.yaml
```diff
@@ -9,12 +9,15 @@ cmd:
...
+  # Average value 55.9 (+/- 1.1%) with SchedulingEquivalenceHashing enabled
+  # (was 58.7 without). Slight decrease due to requeue latency from hash bulk-move.
+  cq: 50 #%
 
 wlClassesMaxAvgTimeToAdmissionMs:
-  # Average value 6666 (+/- 14%), setting at +35% + 2s accounting for observed failures
-  large: 11_000
+  # Average value 16501 (+/- 13%) with SchedulingEquivalenceHashing enabled, setting at +40%.
+  # Previously 6666 with hash disabled; increase is expected as equivalent NoFit workloads
+  # are bulk-moved to inadmissible and wait for capacity events to be re-evaluated.
```
mimowo, line 19: Could we compensate that by a bit more frequent requeue period for the performance tests? I think we could use something like 100ms instead of 1s, wdyt?

mimowo: OTOH, I'  also ok with keeping 1s. 

More importantly, could you maybe check:
1. what is the impact on the total wallTimeMs for all our performance suites
2. what is the impact on the total wallTimeMs for the large scale test in /test/performance/scheduler/configs/large-scale

pkg/cache/queue/cluster_queue.go
```diff
@@ -537,7 +542,8 @@ func (c *ClusterQueue) requeueIfNotPresent(log logr.Logger, wInfo *workload.Info
 	}
 	log.V(2).Info(logMsg, "clusterQueue", c.name, "workload", key)
 
-	if features.Enabled(features.SchedulingEquivalenceHashing) && wInfo.SchedulingHash != workload.SchedulingHashUnknown && !immediate {
+	if features.Enabled(features.SchedulingEquivalenceHashing) && wInfo.SchedulingHash != workload.SchedulingHashUnknown &&
+		(reason == RequeueReasonNoFit || reason == RequeueReasonPreemptionNoCandidates) {
```
mimowo, line 546: Since we already have prod users enabling `SchedulingEquivalenceHashing` I'm wondering it might be safer to introduce anouther FG like `SchedulingEquivalenceHashingRestricted`, could be Beta by default, just so that we have a tool to disable the restriction in case of problems.

sohankunkerkar: The restriction only removes hashing for NamespaceMismatch, PreemptionGated, and Generic, all identified as broken in #10001 (namespace cross-contamination, MultiKueue individual processing). The two reasons prod users care about (NoFit + PreemptionNoCandidates) are both preserved. I don't think a separate FG is needed since there's no valid use case for the removed behavior, but happy to add one if you feel strongly about it.

test/performance/scheduler/configs/baseline/rangespec.yaml
sohankunkerkar: Sure. I did run those tests and here's the Performance impact across suites:                                                                                                                                                                                
                                                                                                                                                                                                                   
  **Baseline (15k workloads):**                                                                                                                                                                                    
  | Metric | hash off | hash on | Delta |           
  | --- | --- | --- | --- |                         
  | wallMs | 368,568 | 354,086 | -3.9% |            
  | large | 11,918 | 15,615 | +31% |                
  | medium | 83,107 | 67,461 | -19% |               
  | small | 233,131 | 217,133 | -6.9% |             

  **Large-scale (50k workloads):**                  
  | Metric | hash off | hash on | Delta |           
  | --- | --- | --- | --- |                         
  | wallMs | 1,148,840 | 1,137,344 | -1.0% |        
  | large | 75,245 | 78,353 | +4.1% |               
  | medium | 233,600 | 231,851 | -0.7% |            
  | small | 684,797 | 676,737 | -1.2% |             

  Overall wallMs improves in both suites. Large workloads regress in baseline (+31%) because each takes 100% of CQ quota (request=20, nominal=20), causing aggressive bulk-move. In large-scale the regression is  
  only +4.1%. Medium and small workloads improve across the board.

test/integration/singlecluster/scheduler/inadmissible/equivalence_test.go
sohankunkerkar: The test passes on main too since workloads end up inadmissible one-by-one even without the hash. With these changes it goes through the bulk-move path instead, same end state. So this test validates correctness of the `PreemptionNoCandidates` path, not the optimization itself. The hash mechanism is directly tested in the unit test (TestRequeueHashTriggerByReason checks noFitSchedulingHashes). Will add a comment in the test explaining this.

test/performance/scheduler/configs/baseline/rangespec.yaml
mimowo: This is consistent with my expectations - we improve the overall thoughput, but underperform for large workloads as they now spend time in the backoff. That could probably be mitigated by using 100ms of the requeue interval, but would not be 1:1 setup that users use. 

Also, I think it is not very realistic test when there are many large workloads which take 100% CQ capacity. Such workloads are rare in my experience, so I don't worry that much here.

pkg/cache/queue/cluster_queue.go
mimowo: My hesitation is because I'm not sure if users only rely on  (NoFit + PreemptionNoCandidates), because maybe there are some other modes which would previously be covered, but not after the change.

I don't think users would be affected by changing any of them from bulk move -> no bulk move for"NamespaceMismatch or PreemptionGated, because (1.) is expception, (2.) feature they don't use.

However, I'm not so sure about Generic, it seems there are multiple scenarios which can lead to that, but IIUC we assume all of them are "anomaly" so they should not trigger any BAU traffic, right?

mimowo:
LGTM, thank you 👍

pkg/cache/queue/cluster_queue.go
sohankunkerkar: Traced all code paths!  After this change, Generic only applies to inadmissibleEntries from nominate() (failed admission checks, invalid resources, limit range violations). These are error conditions, not normal scheduling outcomes. All BAU traffic now has explicit reasons: `NoFit`, `PreemptionNoCandidates`, `NamespaceMismatch`, `PreemptionGated`, or `FailedAfterNomination`. So no prod user would be affected by Generic losing hash support.

mimowo (commented):
/lgtm
/approve
Thank you 👍

mimowo:
/unhold
The concern about supporting Preempt mode, noPreemptionCandidates is addressed

Issue #10848, opened 2026-04-30
Flaky E2E: MultiKueue Sequential when MultiKueueOrchestratedPreemption is enabled should not trigger concurrent preemptions


mbobrovskyi:
**Which test is flaking?**:
End To End MultiKueue Suite: kindest/node:v1.35.0: [It] MultiKueue Sequential when MultiKueueOrchestratedPreemption is enabled should not trigger concurrent preemptions.

**Link to failed CI job or steps to reproduce locally**:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10844/pull-kueue-test-e2e-multikueue-sequential-main/2049759299504181248

**Failure message or logs**:
```
{ failed [FAILED] Timed out after 10.001s.
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/multikueuesequential/e2e_test.go:828 with:
Expected
    <bool>: true
not to equal
    <bool>: true
In [It] at: /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/multikueuesequential/e2e_test.go:841 @ 04/30/26 08:11:34.268
}
```

**Anything else we need to know?**:

mimowo:
cc @Singularity23x0

mimowo:
Use the flake debugger skill here:

 I have investigated the flake reported in https://github.com/kubernetes-sigs/kueue/issues/10848 (https://github.com/kubernetes-sigs/kueue/issues/10848) and identified the root cause.

  Summary of Findings

  The failing test MultiKueue Sequential when MultiKueueOrchestratedPreemption is enabled [It] should not trigger concurrent preemptions fails because it incorrectly assumes that a high-priority workload will only be created in one of the worker
  clusters. However, given the current configuration and MultiKueue's dispatching logic, the workload is correctly synchronized to both worker clusters, causing the test's assertion to fail.

  1. Incorrect Test Assertion
  The failure occurs at test/e2e/multikueuesequential/e2e_test.go:828:

   1 g.Expect(worker1Error == nil).NotTo(gomega.Equal(worker2Error == nil))
  This line expects that exactly one of the worker clusters will have the workload (worker1Error == nil XOR worker2Error == nil). If the workload is present in both clusters, both conditions are true, and the NotTo(Equal(true)) assertion fails with:
[...]

PR #11122, opened 2026-05-11
Increase flaky pull-kueue-test-e2e-multikueue-sequential-main verbosity.

polinasand:
#### What type of PR is this?

/area testing
/area multikueue
/kind flake

This PR adds logging with workloads details inside a flaky test, to provide more data about resources state in case of future flakiness.

Progresses #10848

```release-note
NONE

```

### DIFF
--- a/test/e2e/multikueue/sequential/e2e_test.go
+++ b/test/e2e/multikueue/sequential/e2e_test.go
@@ -838,7 +838,7 @@ var _ = ginkgo.Describe("MultiKueue Sequential", func() {
 						unaffectedWlKey = lowWlKey1
 						unaffectedWorkerClient = k8sWorker1Client
 					}
-				}, util.Timeout, util.Interval).Should(gomega.Succeed())
+				}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
 			})
 
 			ginkgo.By("Checking that the non-evicted workload remains Admitted", func() {
--- a/test/e2e/multikueue/sequential/e2e_test.go
+++ b/test/e2e/multikueue/sequential/e2e_test.go
@@ -19,10 +19,12 @@ package sequential
 import (
 	"fmt"
 	"os/exec"
+	"strings"
 
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
+	"github.com/onsi/gomega/format"
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -825,15 +827,17 @@ var _ = ginkgo.Describe("MultiKueue Sequential", func() {
 					worker1Error := k8sWorker1Client.Get(ctx, highWlKey, worker1HighWorkload)
 					worker2Error := k8sWorker2Client.Get(ctx, highWlKey, worker2HighWorkload)
 
-					g.Expect(worker1Error == nil).NotTo(gomega.Equal(worker2Error == nil))
+					g.Expect(worker1Error == nil).NotTo(gomega.Equal(worker2Error == nil)) //
 
 					if worker1Error == nil {
 						g.Expect(worker1HighWorkload.Status.Conditions).To(utiltesting.HaveConditionStatusTrue(kueue.WorkloadAdmitted))
+						g.Expect(workload.IsAdmitted(worker1HighWorkload)).To(gomega.BeTrue(), assertMsg("Workload not Admitted in worker1", highWlKey))
 						evictedWlKey = lowWlKey1
 						unaffectedWlKey = lowWlKey2
 						unaffectedWorkerClient = k8sWorker2Client
 					} else {
 						g.Expect(worker2HighWorkload.Status.Conditions).To(utiltesting.HaveConditionStatusTrue(kueue.WorkloadAdmitted))
+						g.Expect(workload.IsAdmitted(worker2HighWorkload)).To(gomega.BeTrue(), assertMsg("Workload not Admitted in worker2", highWlKey))
 						evictedWlKey = lowWlKey2
 						unaffectedWlKey = lowWlKey1
 						unaffectedWorkerClient = k8sWorker1Client
@@ -876,3 +880,24 @@ var _ = ginkgo.Describe("MultiKueue Sequential", func() {
 		})
 	})
 })
+
+func assertMsg(msg string, wlKey client.ObjectKey) func() string {
+	return func() string {
+		output := &strings.Builder{}
+		fmt.Fprintln(output, msg)
+		withClusterWl(output, wlKey, "Manager", k8sManagerClient)
+		withClusterWl(output, wlKey, "Worker1", k8sWorker1Client)
+		withClusterWl(output, wlKey, "Worker2", k8sWorker2Client)
+		return output.String()
+	}
+}
+
+func withClusterWl(output *strings.Builder, wlKey client.ObjectKey, cName string, cClient client.Client) {
+	wl := &kueue.Workload{}
+	if err := cClient.Get(ctx, wlKey, wl); err == nil {
+		fmt.Fprintln(output, cName, "Workload:", format.Object(wl.ObjectMeta, 1))
+		fmt.Fprintln(output, cName, "Status:", format.Object(wl.Status, 1))
+	} else {
+		fmt.Fprintln(output, "Workload not present in:", cName, "Error:", err)
+	}
+}
--- a/test/e2e/multikueue/sequential/e2e_test.go
+++ b/test/e2e/multikueue/sequential/e2e_test.go
@@ -827,7 +827,7 @@ var _ = ginkgo.Describe("MultiKueue Sequential", func() {
 					worker1Error := k8sWorker1Client.Get(ctx, highWlKey, worker1HighWorkload)
 					worker2Error := k8sWorker2Client.Get(ctx, highWlKey, worker2HighWorkload)
 
-					g.Expect(worker1Error == nil).NotTo(gomega.Equal(worker2Error == nil)) //
+					g.Expect(worker1Error == nil).NotTo(gomega.Equal(worker2Error == nil))
 
 					if worker1Error == nil {
 						g.Expect(worker1HighWorkload.Status.Conditions).To(utiltesting.HaveConditionStatusTrue(kueue.WorkloadAdmitted))
--- a/test/e2e/multikueue/sequential/e2e_test.go
+++ b/test/e2e/multikueue/sequential/e2e_test.go
@@ -839,7 +839,8 @@ var _ = ginkgo.Describe("MultiKueue Sequential", func() {
 						unaffectedWorkerClient = k8sWorker1Client
 					}
 				}, util.MediumTimeout, util.Interval).Should(gomega.Succeed(), func() string {
-					return util.AssertMsgForMk(ctx, "Workload not Admitted in worker1", worker1HighWorkload, k8sManagerClient, k8sWorker1Client, k8sWorker2Client)() + util.AssertMsgForMk(ctx, "Workload not Admitted in worker2", worker2HighWorkload, k8sManagerClient, k8sWorker1Client, k8sWorker2Client)()
+					return util.AssertMsgForMk(ctx, "Workload not Admitted in worker1", worker1HighWorkload, k8sManagerClient, k8sWorker1Client, k8sWorker2Client)() +
+						util.AssertMsgForMk(ctx, "Workload not Admitted in worker2", worker2HighWorkload, k8sManagerClient, k8sWorker1Client, k8sWorker2Client)()
 				})
 			})
 
--- a/test/e2e/multikueue/sequential/e2e_test.go
+++ b/test/e2e/multikueue/sequential/e2e_test.go
@@ -839,7 +839,7 @@ var _ = ginkgo.Describe("MultiKueue Sequential", func() {
 						unaffectedWorkerClient = k8sWorker1Client
 					}
 				}, util.MediumTimeout, util.Interval).Should(gomega.Succeed(), func() string {
-					return util.AssertMsgForMk("Workload not Admitted in worker1", worker1HighWorkload, ctx, k8sManagerClient, k8sWorker1Client, k8sWorker2Client)() + util.AssertMsgForMk("Workload not Admitted in worker2", worker2HighWorkload, ctx, k8sManagerClient, k8sWorker1Client, k8sWorker2Client)()
+					return util.AssertMsgForMk(ctx, "Workload not Admitted in worker1", worker1HighWorkload, k8sManagerClient, k8sWorker1Client, k8sWorker2Client)() + util.AssertMsgForMk(ctx, "Workload not Admitted in worker2", worker2HighWorkload, k8sManagerClient, k8sWorker1Client, k8sWorker2Client)()
 				})
 			})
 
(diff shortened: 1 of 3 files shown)


test/e2e/multikueue/sequential/e2e_test.go
```diff
@@ -876,3 +880,24 @@ var _ = ginkgo.Describe("MultiKueue Sequential", func() {
 		})
 	})
 })
+
+func assertMsg(msg string, wlKey client.ObjectKey) func() string {
```
mimowo: Move these predefined function https://github.com/kubernetes-sigs/kueue/blob/048c2f01639162068013bf957e9d609b75699940/test/e2e/multikueue/baseline/e2e_test.go#L1995C6-L2015
to test/util/e2e.go and reuse instead of duplicating.

We need to rename it to avoid conflict, maybe `AssertMsgForMk`

test/e2e/multikueue/sequential/e2e_test.go
```diff
@@ -829,11 +831,13 @@ var _ = ginkgo.Describe("MultiKueue Sequential", func() {
 
 					if worker1Error == nil {
 						g.Expect(worker1HighWorkload.Status.Conditions).To(utiltesting.HaveConditionStatusTrue(kueue.WorkloadAdmitted))
+						g.Expect(workload.IsAdmitted(worker1HighWorkload)).To(gomega.BeTrue(), assertMsg("Workload not Admitted in worker1", highWlKey))
 						evictedWlKey = lowWlKey1
 						unaffectedWlKey = lowWlKey2
 						unaffectedWorkerClient = k8sWorker2Client
 					} else {
 						g.Expect(worker2HighWorkload.Status.Conditions).To(utiltesting.HaveConditionStatusTrue(kueue.WorkloadAdmitted))
+						g.Expect(workload.IsAdmitted(worker2HighWorkload)).To(gomega.BeTrue(), assertMsg("Workload not Admitted in worker2", highWlKey))
```
kshalot: Maybe we could move `worker1HighWorkload` and `worker2HighWorkload` outside of the block and use `util.AssertMsg` on the entire `Eventually` case? like:
```go
...
}, util.MediumTimeout, util.Interval).Should(gomega.Succeed(), util.AssertMsg("some good message", worker1HighWorkload, worker2HighWorkload))

Comment (kshalot) [reply test/e2e/multikueue/sequential/e2e_test.go]: If the workload was not found it will remain `nil`. And this won't require adding any new logic.

File: test/e2e/multikueue/sequential/e2e_test.go [hunk 3]
```diff
@@ -829,11 +831,13 @@ var _ = ginkgo.Describe("MultiKueue Sequential", func() {
 
 					if worker1Error == nil {
 						g.Expect(worker1HighWorkload.Status.Conditions).To(utiltesting.HaveConditionStatusTrue(kueue.WorkloadAdmitted))
+						g.Expect(workload.IsAdmitted(worker1HighWorkload)).To(gomega.BeTrue(), assertMsg("Workload not Admitted in worker1", highWlKey))
```
olekzabl: Nit: Looking at these 2 expectations, aren't they *equivalent*, just the latter equipped with a custom message?
How about just this adding the message to the existing expectation?
```suggestion
g.Expect(worker1HighWorkload.Status.Conditions).To(utiltesting.HaveConditionStatusTrue(kueue.WorkloadAdmitted), assertMsg("Workload not Admitted in worker1", highWlKey))
```

test/e2e/multikueue/sequential/e2e_test.go
olekzabl: @kshalot 
Not sure if I'm following. The remark on `nil` is cryptic to me, and I'm not sure what you mean by "new logic" (possibly obsoleted by [this comment](https://github.com/kubernetes-sigs/kueue/pull/11122#discussion_r3219830336)?).

But anyway, given that @polinasand apparently wants to log workload state from all the clusters, I guess it wouldn't hurt to replace your proposed message with sth like `assertWlMsg("sth1", worker1HighWorkload) + assertWlMsg("sth2", worker2HighWorkload)`?

kshalot: My point is that if we initialize the variables outside (maybe we can even assign `&kueue.Workload{}` to them, it doesn't matter, then we could just wrap the entire block with the existing `util.AssertMsg`.

This would mean that wherever it fails, we will get the message. For example, currently this PR does not cover the line that was originally failing:
```go
g.Expect(worker1Error == nil).NotTo(gomega.Equal(worker2Error == nil))
```

My point with `nil` was that if the call fails:
```go
worker1Error := k8sWorker1Client.Get(ctx, highWlKey, worker1HighWorkload)
```
then `worker1HighWorkload` will remain `&kueue.Workload{}`, which conveys the message that something went wrong and we can log it as is, without any additional logic handling "not found" etc. (so I misspoke with `nil`, it's more accurate to say `&kueue.Workload{}`).

One difference is that it won't include the manager workload, which might or might not be useful.

olekzabl: Thanks, now I understand you better.

+1 for wishing to cover more potentially failing lines, in particular `(w1err == nil) != (w2err == nil)` (where, in particular, two non-nil errs could be a misleading case).

> One difference is that it won't include the manager workload, which might or might not be useful.

Agreed, and my intuition is to rather include it, if it's not problematic.
Every iteration of this ping-pong research costs time (to merge the PR, and then to wait for a repro), so I'd lean towards gathering more info whenever we have some nice idea.

So, I think we're converging towards sth like `gomega.Eventually(..., assertWlMsg(w1) + assertWlMsg(w2))`.
(And if so, [my other comment](https://github.com/kubernetes-sigs/kueue/pull/11122#discussion_r3222786352) will become obsolete)

kshalot: Ok, let's include the manager. But I'd consider moving it to utils as something more generic, as proposed in https://github.com/kubernetes-sigs/kueue/pull/11122#discussion_r3219830336.

olekzabl: Oh, yes. 

BTW, to clarify: my understanding of that comment is that we shouldn't move the helpers from this PR to utils.
Instead, we should move pre-existing helpers ([here](https://github.com/kubernetes-sigs/kueue/pull/11122)) to utils, and use them in this PR.
So that this PR does not need to add new logging helpers.

test/util/e2e.go
```diff
@@ -804,3 +804,25 @@ func SetResourceNominalQuota(cq *kueue.ClusterQueue, resourceName corev1.Resourc
 	}
 	return cq
 }
+
+func AssertMsgForMk(msg string, wl *kueue.Workload, ctx context.Context, k8sManagerClient client.Client, k8sWorker1Client client.Client, k8sWorker2Client client.Client) func() string {
```
mimowo: nit, ctx is always the first param by convention https://pkg.go.dev/context "The Context should be the first parameter, typically named ctx:"

mimowo:
/approve
/cherrypick release-0.17
/cherrypick release-0.16
leaving lgtm to @kshalot or @olekzabl for sanity check

mimowo:
Thanks

test/e2e/multikueue/sequential/e2e_test.go
```diff
@@ -838,7 +838,10 @@ var _ = ginkgo.Describe("MultiKueue Sequential", func() {
 						unaffectedWlKey = lowWlKey1
 						unaffectedWorkerClient = k8sWorker1Client
 					}
-				}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
+				}, util.MediumTimeout, util.Interval).Should(gomega.Succeed(), func() string {
+					return util.AssertMsgForMk(ctx, "Workload not Admitted in worker1", worker1HighWorkload, k8sManagerClient, k8sWorker1Client, k8sWorker2Client)() +
+						util.AssertMsgForMk(ctx, "Workload not Admitted in worker2", worker2HighWorkload, k8sManagerClient, k8sWorker1Client, k8sWorker2Client)()
```
mbobrovskyi: I’m actually not sure why we have this duplication. Can we use `AssertMsg()` for `worker1HighWorkload` and `worker2HighWorkload`?

I agree that we need to print the current state for `worker1HighWorkload` and `worker2HighWorkload`, but I think we can introduce one more helper for `k8sManagerClient`, `k8sWorker1Client`, and `k8sWorker2Client` and reuse it in `AssertMsgForMk()`.

test/util/e2e.go
```diff
@@ -804,3 +804,25 @@ func SetResourceNominalQuota(cq *kueue.ClusterQueue, resourceName corev1.Resourc
...
 }
+
+func AssertMsgForMk(ctx context.Context, msg string, wl *kueue.Workload, k8sManagerClient client.Client, k8sWorker1Client client.Client, k8sWorker2Client client.Client) func() string {
+	return func() string {
+		wlKey := client.ObjectKeyFromObject(wl)
+		return strings.Join([]string{
+			AssertMsg(msg, wl)(),
+			AssertMsg("Manager", getWorkload(ctx, k8sManagerClient, wlKey))(),
+			AssertMsg("Worker1", getWorkload(ctx, k8sWorker1Client, wlKey))(),
+			AssertMsg("Worker2", getWorkload(ctx, k8sWorker2Client, wlKey))(),
```
kshalot: I'm not sure I like this. Since you are doing requests in `getWorkload` **after** the flake already occurred, you might fetch the new state that would actually pass the flaky, no? I mean - it's not guaranteed that what you fetch from the cluster here will be consistent with what the test observed.

That's another reason why I liked the approach I described in https://github.com/kubernetes-sigs/kueue/pull/11122#discussion_r3219831812, since it's using exactly what it encountered in the test.

mimowo: Yes, but if the Eventually fails then it means we waited 45s or 90s already. The assumption is that at this point we can be confident the object is not making progress and will not change if added another second. 

Still, I would not object another approach which doesn't have the drawback, but I'm not sure if we have such currently.

mbobrovskyi: We still keep AssertMsg(msg, wl)(), which shows the current state of the workload.

kshalot: True that my concern might not be practical given the long timeouts. But we can consider https://github.com/kubernetes-sigs/kueue/pull/11122#discussion_r3219831812 like I said. It displays what was returned from the workers without any post-flake calls, but it won't log the manager.

olekzabl: So anyway we get the state after, say, either "45s" or "45s + tiny delay" (where I also don't believe that tiny delay matters).
So maybe the largest risk is that the valuable info may have disappeared way before 45s elapsed?

I'm thinking if we can nicely log "intermediate checkpoints" while waiting on the long timeout.
The "ginkgo-idiomatic" way seems to be `PollProgressInterval` + `AttachProgressReporter`; see [here](https://onsi.github.io/ginkgo/#getting-visibility-into-long-running-specs).

But maybe it's overdoing the logs... WDYT?

test/e2e/multikueue/sequential/e2e_test.go
mimowo: +1, I think this will result in printing the same set of workloads twice because `AssertMsgForMk` already handles lookups to all clusters.

I think to make it less confusing `AssertMsgForMk` should only accept the wlKey so that it is agnostic of the client used. The code here could be simplified to just say:

`return util.AssertMsgForMk(ctx, "Workload was expected to be admitted on exactly one worker", highWlKey, k8sManagerClient, k8sWorker1Client, k8sWorker2Client)`

test/util/e2e.go
mimowo: I understand your concern @kshalot, and I agree the proposal in https://github.com/kubernetes-sigs/kueue/pull/11122#discussion_r3222786352 does not have the drawback, but it seems to have another - it is more work on the test side to each time think which information is needed for debugging MK workloads, rather than to have a standardized approach with AssertMsgForMk which could dump the workloads just based on the "wlKey".

Still, I'm wondering if we could have the best of both worlds if we encapsulated the three workloads in a struct like `struct mkWorkload` , say

```golang 
type MKWorkload  struct {
  manager
  worker1
  worker2
  
  managerClinet
  worker1Client
  worker2Client
}
```
Then it could have the `Get` function which would refresh the state of the struct. Then, `mkWorkload.worker1`, `mkWorkload.worker2` , `mkWorkload.manager`  could be used for asserts, and to build the failure message, without the delay.

Then it would make `AssertMsgForMk` obsolete, but this is significantly more work to do well, so I'm leaning to go ahead with `AssertMsgForMk` as pilot and search for alternatives like MKWorkload if this proves confusing / not suffecient. 

BTW: The post-failure dump strategy is used in k8s core extensively with [DumpAllNamespaceInfo](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/framework/debug/dump.go#L66), and has proven to be useful. Still, I agree what we do with `AssertMsgForMk` is somewhere in-between.

polinasand: Agreed to stay with current approach for now.

test/util/e2e.go
```diff
@@ -805,11 +805,9 @@ func SetResourceNominalQuota(cq *kueue.ClusterQueue, resourceName corev1.Resourc
 	return cq
 }
 
-func AssertMsgForMk(ctx context.Context, msg string, wl *kueue.Workload, k8sManagerClient client.Client, k8sWorker1Client client.Client, k8sWorker2Client client.Client) func() string {
+func AssertMsgForMk(ctx context.Context, msg string, wlKey client.ObjectKey, k8sManagerClient client.Client, k8sWorker1Client client.Client, k8sWorker2Client client.Client) func() string {
```
mbobrovskyi, line 808: I’m actually not sure about that. We need to show the current state of the object instead of the newly fetched one. It could be confusing if the original object and the fetched object are different. So I propose reverting these changes and printing the failed object instead.

mbobrovskyi: My concern was different. I think we should print worker1HighWorkload and worker2HighWorkload as failed workloads, and also fetch worker3HighWorkload to provide more details.

polinasand: What is `worker3HighWorkload` here?

mbobrovskyi: > What is worker3HighWorkload here?

Oops, sorry, I mean workload for manager.

mimowo: > t could be confusing if the original object and the fetched object are different. So I propose reverting these changes and printing the failed object instead.

I don't understand, we re-fetch the object in both approaches. The object can be changed in the tiny amount of time anyway, yes this is a known drawback, but it is not getting fixed by passing the entire workload object. Passing the entire workload makes it more confusing imo, because it asks a question: which cluster is the workload coming from? But this question is not relevant, what matters is only the "key" - because the same key is used for all clusters.

mbobrovskyi: No. In this case, we don’t re-fetch the failed object. We print the failed object, so we should see the failed object.

mbobrovskyi: Otherwise, we may fetch an updated object, which could make debugging confusing.

mimowo: I think we already discussed we are ok with this risk, also this risk remains for other clusters.

test/e2e/multikueue/sequential/e2e_test.go
```diff
@@ -818,10 +818,10 @@ var _ = ginkgo.Describe("MultiKueue Sequential", func() {
 			var unaffectedWorkerClient client.Client
 
 			ginkgo.By("Checking that the high-priority workload is admitted in one of the workers", func() {
-				gomega.Eventually(func(g gomega.Gomega) {
-					worker1HighWorkload := &kueue.Workload{}
-					worker2HighWorkload := &kueue.Workload{}
+				worker1HighWorkload := &kueue.Workload{}
+				worker2HighWorkload := &kueue.Workload{}
 
+				gomega.Eventually(func(g gomega.Gomega) {
```
olekzabl, line 824: Super nit: this move should be ideally reverted now.
(So that `worker{1,2}HighWorkload` have the minimal necessary scope).

olekzabl (commented):
/lgtm
though with one super-nit comment

mimowo:
@polinasand please check if it makes sense to CP for 0.16, if so then prepare the CP manually

PR #11125, opened 2026-05-11
[ConcurrentAdmission] Do not allow migration to a flavor below the minPreferredFlavor

PBundyra:
#### What type of PR is this?
/kind bug

When using the ConcurrentAdmission feature, once one Variant is running it prevents from migrating to a more favorable flavor that is below the `minPreferredFlavorName`

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -455,6 +455,9 @@ func (s *Scheduler) processEntry(
 
 	if features.Enabled(features.ConcurrentAdmission) && concurrentadmission.IsVariant(e.Obj) {
 		if lessFavorableSibling := s.findAdmittedLessFavorableSibling(&e.Info, snapshot); lessFavorableSibling != nil {
+			if !s.isMigrationAllowed(cq, e, log) {
+				return
+			}
 			s.issueMigration(ctx, log, e, lessFavorableSibling)
 			return
 		}
@@ -1172,6 +1175,25 @@ func (s *Scheduler) findAdmittedSiblingMatching(wl *workload.Info, cq *schdcache
 		}
 	}
 	return nil
+}
+
+func (s *Scheduler) isMigrationAllowed(cq *schdcache.ClusterQueueSnapshot, e *entry, log klog.Logger) bool {
+	if cq.ConcurrentAdmissionPolicy == nil || cq.ConcurrentAdmissionPolicy.Migration.Constraints == nil || cq.ConcurrentAdmissionPolicy.Migration.Constraints.MinPreferredFlavorName == nil {
+		return true
+	}
+	minPrefFlavor := *cq.ConcurrentAdmissionPolicy.Migration.Constraints.MinPreferredFlavorName
+	flavors := cq.ResourceGroups[0].Flavors
+	minPrefIdx := slices.Index(flavors, minPrefFlavor)
+	wlFlavorIdx, err := getFlavorIndex(&e.Info, flavors)
+	if err != nil || minPrefIdx < 0 {
+		return true
+	}
+	if wlFlavorIdx > minPrefIdx {
+		log.V(3).Info("Skipping migration as target flavor is below MinPreferredFlavorName", "targetFlavor", concurrentadmission.GetVariantFlavor(e.Obj), "minPreferredFlavor", minPrefFlavor)
+		e.inadmissibleMsg = "Target flavor is below MinPreferredFlavorName"
+		return false
+	}
+	return true
 }
 
 func getFlavorIndex(wl *workload.Info, flavors []kueue.ResourceFlavorReference) (int, error) {
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -454,10 +454,10 @@ func (s *Scheduler) processEntry(
 	}
 
 	if features.Enabled(features.ConcurrentAdmission) && concurrentadmission.IsVariant(e.Obj) {
-		if !s.isMigrationAllowed(cq, e, log) {
-			return
-		}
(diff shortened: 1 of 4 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -1174,6 +1177,25 @@ func (s *Scheduler) findAdmittedSiblingMatching(wl *workload.Info, cq *schdcache
 	return nil
 }
 
+func (s *Scheduler) isMigrationAllowed(cq *schdcache.ClusterQueueSnapshot, e *entry, log klog.Logger) bool {
+	if cq.ConcurrentAdmissionPolicy == nil || cq.ConcurrentAdmissionPolicy.Migration.Constraints == nil || cq.ConcurrentAdmissionPolicy.Migration.Constraints.MinPreferredFlavorName == nil {
+		return true
+	}
+	minPrefFlavor := *cq.ConcurrentAdmissionPolicy.Migration.Constraints.MinPreferredFlavorName
```
mwielgus, line 1186: TBH, I have a problem understanding what is minimal here. Is it the first or last flavor? Are we sure this is the best name for that setting?

pkg/scheduler/scheduler.go
```diff
@@ -1174,6 +1177,25 @@ func (s *Scheduler) findAdmittedSiblingMatching(wl *workload.Info, cq *schdcache
...
+	flavors := cq.ResourceGroups[0].Flavors
+	minPrefIdx := slices.Index(flavors, minPrefFlavor)
+	wlFlavorIdx, err := getFlavorIndex(&e.Info, flavors)
+	if err != nil || minPrefIdx < 0 {
+		return true
```
mimowo: Do not silent errors, propagate up, or at the very least log

pkg/scheduler/scheduler.go
```diff
@@ -455,6 +455,9 @@ func (s *Scheduler) processEntry(
 
 	if features.Enabled(features.ConcurrentAdmission) && concurrentadmission.IsVariant(e.Obj) {
 		if lessFavorableSibling := s.findAdmittedLessFavorableSibling(&e.Info, snapshot); lessFavorableSibling != nil {
+			if !s.isMigrationAllowed(cq, e, log) {
```
mimowo, line 458: This function does not use `lessFavorableSibling` as an argument, so why don't we check it before calling `findAdmittedLessFavorableSibling`?

pkg/scheduler/scheduler.go
```diff
@@ -1174,6 +1177,25 @@ func (s *Scheduler) findAdmittedSiblingMatching(wl *workload.Info, cq *schdcache
...
 }
 
+func (s *Scheduler) isMigrationAllowed(cq *schdcache.ClusterQueueSnapshot, e *entry, log klog.Logger) bool {
+	if cq.ConcurrentAdmissionPolicy == nil || cq.ConcurrentAdmissionPolicy.Migration.Constraints == nil || cq.ConcurrentAdmissionPolicy.Migration.Constraints.MinPreferredFlavorName == nil {
+		return true
+	}
+	minPrefFlavor := *cq.ConcurrentAdmissionPolicy.Migration.Constraints.MinPreferredFlavorName
+	flavors := cq.ResourceGroups[0].Flavors
+	minPrefIdx := slices.Index(flavors, minPrefFlavor)
+	wlFlavorIdx, err := getFlavorIndex(&e.Info, flavors)
```
mimowo: nit, this is preexisting, but let's rename it from `getFlavorIndex` to `resolveFlavorIndex` to reflect this is not a simple "get"

pkg/scheduler/scheduler.go
mimowo: I guess this is a good question but is not really relevant for the implementation PR as the API was already reviewed. We may open separate discussion on improving the API naming, but I don't consider this blocking for this PR. Still, if we think this is important to fix, then let's do it before 0.18 is released which is soon.

mwielgus: This is somewhat relevant as the PR is precisely about minPreferredFlavor and if the name is misleading that the whole mechanics of the PR may be wrong.

mimowo: Maybe, but in that case i would suggest a small kep update PR first to align on the name. Im open to better suggestions

pkg/scheduler/scheduler.go
PBundyra: Done

pkg/scheduler/scheduler.go
PBundyra: ~~Done~~

Edit: Reverted

pkg/scheduler/scheduler.go
PBundyra: Done

pkg/scheduler/scheduler.go
PBundyra: Even we don't use `lessFavorableSibling` as an argument, we still use the information that it exists. We only want to check if the Variant is below `minPreferredFlavor` is we know that there's a running Sibling

pkg/scheduler/scheduler.go
mimowo: I had a sync with @PBundyra and @mwielgus and we will try to rename the field in a follow up PR, the currently leading idea is `lastAcceptableFlavor`, cc @tenzen-y

mimowo (commented):
/lgtm
/approve
Let's merge the PR as it fixes an important gap in the mechanism. The naming issue (raised in the [comment](https://github.com/kubernetes-sigs/kueue/pull/11125/changes#r3220780904)) will be addressed as a separate PR along with the KEP update.

PR #11136, opened 2026-05-12
tas: fall back to BestFit when balanced placement fails

### LINKED ISSUE #7494: Better support for LeaderWorkerSet in Balanced TAS
In Balanced Placement algorithm in TAS, there is only basic support for LeaderWorkerSet. The threshold value calculation and pruning logic leaves space for leader on each node, which is suboptimal. 
https://github.com/kubernetes-sigs/kueue/pull/6851#discussion_r2481622745

Also, consider making the pruning reversible to enable fallback to BestFit at later stages of the Balanced algorithm.

**Why is this needed**:
It would result in more balanced placement for LWS.

ShaanveerS:
#### What type of PR is this?
/kind bug
/area tas

If `applyBalancedPlacementAlgorithm` can't satisfy the threshold (e.g. due to a LWS leader consuming node capacity), fall back to BestFit instead of failing the assignment.

Fixes #7494 (second part)

```release-note
TAS: Balanced Placement now also falls back to BestFit when the threshold cannot be satisfied, instead of failing the assignment.
```

### DIFF
--- a/pkg/cache/scheduler/tas_cache_test.go
+++ b/pkg/cache/scheduler/tas_cache_test.go
@@ -3953,6 +3953,111 @@ func TestFindTopologyAssignments(t *testing.T) {
 				},
 			},
 		},
+		//           b1           b2
+		//      /    |    \       |
+		//   r1     r2    r3     r1
+		//   |      |     |      |
+		//   x1:3  x2:3  x3:8   x4:11
+		// request: 7
+		// leaders: 1 (5 GPUs)
+		// expected outcome: x1:3, x2:1, x3:3 + leader
+		"balanced placement; fallback to BestFit when leader worker set cannot meet threshold": {
+			featureGates: map[featuregate.Feature]bool{features.TASBalancedPlacement: true},
+			nodes: []corev1.Node{
+				*testingnode.MakeNode("b1-r1-x1").
+					Label(tasBlockLabel, "b1").
+					Label(tasRackLabel, "r1").
+					Label(corev1.LabelHostname, "x1").
+					StatusAllocatable(corev1.ResourceList{
+						"example.com/gpu":   resource.MustParse("3"),
+						corev1.ResourcePods: resource.MustParse("12"),
+					}).
+					Ready().
+					Obj(),
+				*testingnode.MakeNode("b1-r2-x2").
+					Label(tasBlockLabel, "b1").
+					Label(tasRackLabel, "r2").
+					Label(corev1.LabelHostname, "x2").
+					StatusAllocatable(corev1.ResourceList{
+						"example.com/gpu":   resource.MustParse("3"),
+						corev1.ResourcePods: resource.MustParse("12"),
+					}).
+					Ready().
+					Obj(),
+				*testingnode.MakeNode("b1-r3-x3").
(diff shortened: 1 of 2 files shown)


ShaanveerS:
cc @pajakd

mimowo:
cc @pajakd

pkg/cache/scheduler/tas_cache_test.go
```diff
@@ -3953,6 +3953,111 @@ func TestFindTopologyAssignments(t *testing.T) {
 				},
 			},
 		},
+		//           b1           b2
+		//      /    |    \       |
+		//   r1     r2    r3     r1
+		//   |      |     |      |
+		//   x1:3  x2:3  x3:8   x4:11
+		// request: 7
+		// leaders: 1
```
pajakd: comment here that the leader uses 5 GPUs

pkg/cache/scheduler/tas_cache_test.go
```diff
@@ -3953,6 +3953,111 @@ func TestFindTopologyAssignments(t *testing.T) {
...
+		// expected outcome: x1:3, x2:1, x3:3 + leader
```
pajakd, line 3963: Why does the balanced placement fail here? Its fine if we fall back to Best Fit but I'm just trying to understand.

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -911,7 +911,8 @@ func (s *TASFlavorSnapshot) findTopologyAssignment(
 		if useBalancedPlacement {
 			currFitDomain, fitLevelIdx, reason = applyBalancedPlacementAlgorithm(s, &state.topologyAssignmentParameters, bestThreshold, currFitDomain)
 			if len(reason) > 0 {
-				return nil, reason
+				s.log.V(3).Info("Balanced placement threshold not met, falling back to BestFit", "reason", reason)
```
pajakd: I would even say more strongly something like "Balanced placement algorithm failed, falling back to Best Fit" because we try the Balanced Placement algorithm

ShaanveerS: done!

pkg/cache/scheduler/tas_cache_test.go
ShaanveerS: done!

pkg/cache/scheduler/tas_cache_test.go
ShaanveerS: Balanced placement fails because the leader only fits on `x3`. Once we narrow the placement to `b1`, the 7 workers have to use all three racks and with threshold 3 (calculated earlier while `b2` was still a candidate, since the workers could fit in x3,x4: 7 / 2 = 3), that would take 9 workers, so balanced placement can’t satisfy the threshold and falls back to Best Fit.

pajakd:
/lgtm

Thank you 🙂

mimowo:
/approve
Thank you for the PR, promoting the lgtm from https://github.com/kubernetes-sigs/kueue/pull/11136#issuecomment-4477684180
/lgtm

PR #11137, opened 2026-05-12
Prevent integer overflow in flavorassigner

pajakd:
#### What type of PR is this?
/kind bug

Automatic vulnerability scanner flagged that this integer addition could result in int64 overflow. 

Confirmed that with unit test: in case of two podsets with a total requests exceeding the MaxInt64, the value calculated in the flavorassigner would get negative and the workload would get admitted.

Fixed the issue by checking if the sum would overflow before performing the addition.

Fixes #

```release-note
Fixed vulnerability where two podsets with total requests exceeding max int64 would lead to integer overflow and break quota limits.
```

### DIFF
--- a/pkg/scheduler/flavorassigner/flavorassigner.go
+++ b/pkg/scheduler/flavorassigner/flavorassigner.go
@@ -40,6 +40,7 @@ import (
 	"sigs.k8s.io/kueue/pkg/scheduler/preemption/classical"
 	preemptioncommon "sigs.k8s.io/kueue/pkg/scheduler/preemption/common"
 	utilmaps "sigs.k8s.io/kueue/pkg/util/maps"
+	utilmath "sigs.k8s.io/kueue/pkg/util/math"
 	"sigs.k8s.io/kueue/pkg/util/orderedgroups"
 	"sigs.k8s.io/kueue/pkg/util/tas"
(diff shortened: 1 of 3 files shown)


pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -1072,7 +1072,13 @@ func (a *FlavorAssigner) fitsResourceQuota(log logr.Logger, fr resources.FlavorR
 
 	available := a.cq.Available(fr)
 	maxCapacity := a.cq.PotentialAvailable(fr)
-	val := assumedUsage + requestUsage
+
+	var val int64
```
mimowo: wrap the computations i. a helper function to avoid mixing levels of abtsraction

mimowo:
please also add a release note

pkg/scheduler/flavorassigner/flavorassigner.go
mimowo: also such a function has a potential for reuse

pajakd: Done, moved to util/

mimowo:
Thanks 👍 
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16

PR #11139, opened 2026-05-12
Prevent integer overflow in ResourceValue

pajakd:
#### What type of PR is this?
/kind bug

Automatic vulnerability scanner flagged that conversion to milivalue could result in int64 overflow.

Confirmed that with unit test: in case of a workload requesting a huge amount of CPU, after the conversion the ResourceValue got negative and the workload got admitted.

Fixed the issue by checking if the value would overflow before converting.

Fixes #

```release-note
Fixed vulnerability where workload with CPU requests close to max int64 would lead to integer overflow when converting to milicpu and break quota limits.
```

### DIFF
--- a/pkg/resources/requests.go
+++ b/pkg/resources/requests.go
@@ -99,11 +99,30 @@ func (r Requests) ToResourceList() corev1.ResourceList {
 	return ret
 }
 
+var (
+	maxMilliValue = *resource.NewMilliQuantity(math.MaxInt64, resource.DecimalSI)
+	minMilliValue = *resource.NewMilliQuantity(math.MinInt64, resource.DecimalSI)
+	maxValue      = *resource.NewQuantity(math.MaxInt64, resource.DecimalSI)
+	minValue      = *resource.NewQuantity(math.MinInt64, resource.DecimalSI)
+)
+
 // ResourceValue returns the integer value for the resource name.
 // It's milli-units for CPU and absolute units for everything else.
 func ResourceValue(name corev1.ResourceName, q resource.Quantity) int64 {
 	if name == corev1.ResourceCPU {
+		if q.Cmp(maxMilliValue) > 0 {
+			return math.MaxInt64
+		}
+		if q.Cmp(minMilliValue) < 0 {
+			return math.MinInt64
+		}
 		return q.MilliValue()
+	}
+	if q.Cmp(maxValue) > 0 {
+		return math.MaxInt64
+	}
+	if q.Cmp(minValue) < 0 {
+		return math.MinInt64
 	}
 	return q.Value()
 }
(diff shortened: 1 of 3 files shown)


pkg/resources/requests.go
```diff
@@ -99,12 +99,31 @@ func (r Requests) ToResourceList() corev1.ResourceList {
...
+	maxValue      = *resource.NewQuantity(math.MaxInt64, resource.DecimalSI)
+	minValue      = *resource.NewQuantity(math.MinInt64, resource.DecimalSI)
+)
+
 // ResourceValue returns the integer value for the resource name.
 // It's milli-units for CPU and absolute units for everything else.
 func ResourceValue(name corev1.ResourceName, q resource.Quantity) int64 {
 	if name == corev1.ResourceCPU {
+		if q.Cmp(maxMilliValue) > 0 {
+			return math.MaxInt64
```
mwielgus: TBH, operating on MaxInt / MinInts (or close) is risky, as any math operation on them may trigger overflow. Maybe we should drop workloads that get anywhere close to these values?

mimowo: I would not go into rejecting workloads as a bugfix as this is compatibility breaking. I maybe consider that for future releases, but even that is  tricky. Also, it does not give any guarantees - a workload may have mutlople podSets, then we would need to ensure that `podSet * podSetSIze` is not exceeding maxInt64.

As an alternative fix (my preference) I proposed in the past we could stop using bare int64 and start using considently some custom type like "Amount" that would have the int64 trimming built-in.

However, introducing Amount consistently might be more involving.

pajakd: I also think, it will be hard to define which workloads get close to MaxInt.

mimowo: We could also research how does core k8s deal with overflows, because iirc there are some code paths in the core k8s where requests from different containers and init containers are added.

kshalot: I did a quick Gemini search in the k8s repo and found an approach that is similar to this PR:
https://github.com/kubernetes/kubernetes/blob/f01363b4912ad545ff1a50b690e21bc148271a8c/pkg/controller/podautoscaler/replica_calculator.go#L295-L300

Also here:
https://github.com/kubernetes/kubernetes/blob/f01363b4912ad545ff1a50b690e21bc148271a8c/pkg/scheduler/framework/runtime/framework.go#L598-L601

pkg/resources/requests.go
```diff
@@ -99,12 +99,31 @@ func (r Requests) ToResourceList() corev1.ResourceList {
...
+		}
+		if q.Cmp(minMilliValue) < 0 {
+			return math.MinInt64
+		}
 		return q.MilliValue()
```
mimowo: Let's extact the code to a helper, like`safeMilliValue`, in order to:
- avoid mixing abstraction layers (math operations with business logic)
- avoid duplication - it is done twice in the PR, but likely will be repated in other places too

pajakd: Done.

mimowo (commented):
Thanks 👍 
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16

Issue #10629, opened 2026-04-20
[MultiKueue] Check improvability of remote client caching


olekzabl:
_(I hesitated between "cleanup", "bug" and "feature request" :) Picked the least prominent one)_

MultiKueue controllers (on the manager cluster) fetch K8s objects from workers using remote clients.
Let's focus on Workloads and generic job kinds, as these can grow largest, in terms of QPS or volume.
For those types, MultiKueue does the following data fetches:

* A `.Watch` call [here](https://github.com/kubernetes-sigs/kueue/blob/d97ed7916df34aa9dafd017ca52755d639aa089a/pkg/controller/admissionchecks/multikueue/multikueuecluster.go#L207) (for Workloads when called from [here](https://github.com/kubernetes-sigs/kueue/blob/d97ed7916df34aa9dafd017ca52755d639aa089a/pkg/controller/admissionchecks/multikueue/multikueuecluster.go#L178), and for generic jobs when called from [here](https://github.com/kubernetes-sigs/kueue/blob/d97ed7916df34aa9dafd017ca52755d639aa089a/pkg/controller/admissionchecks/multikueue/multikueuecluster.go#L190)).
  Used to detect any changes, and trigger MK Workload Reconciler on them.
* A `.Get` call [here](https://github.com/kubernetes-sigs/kueue/blob/d97ed7916df34aa9dafd017ca52755d639aa089a/pkg/controller/admissionchecks/multikueue/workload.go#L314) (for Workloads).
  Used to reconcile a MK Workload.
* A `.List` call [here](https://github.com/kubernetes-sigs/kueue/blob/d97ed7916df34aa9dafd017ca52755d639aa089a/pkg/controller/admissionchecks/multikueue/multikueuecluster.go#L292) (for Workloads).
  Used for GC, so not very often (and can be completely [opted out](https://github.com/kubernetes-sigs/kueue/blob/d97ed7916df34aa9dafd017ca52755d639aa089a/pkg/controller/admissionchecks/multikueue/multikueuecluster.go#L667-L670) of via [Kueue Config](https://github.com/kubernetes-sigs/kueue/blob/d97ed7916df34aa9dafd017ca52755d639aa089a/apis/config/v1beta2/configuration_types.go#L305)).

Notably, the `.Get` and the `.List` use a remote client which is **uncached**.
[...]

PR #11141, opened 2026-05-12
Implement KEP-9988 (MultiKueue Manager Quota Automation), Alpha1 stage

### LINKED ISSUE #9988: Support aggregated quotas in MultiKueue manager cluster

A ClusterQueue in a MultiKueue manager cluster should aggregate the capacity and current usage of all connected ClusterQueues in worker clusters, and ensure timely updates.

**Why is this needed**:

This splits to two stories:

1. As a MultiKueue user, I want to understand the quota picture (total limits & usage) without a need to log into individual workers. I want the totals exposed by the manager cluster.

2. As a MultiKueue user relying on finite manager quotas to distribute scheduling load between manager and workers, I want to keep the manager quotas automatically synchronized (*) with total worker capacity.

   (*) Caveat: "synchronized with" may not mean exactly "equal to" (as suggested by feedback from customers & project leads). 
   A relatively simple yet reasonably powerful approach here would be to introduce a configurable _relative multiplier_ (i.e. keep manager quota equal to `customMultiplier * sum(workerQuotas)`).

**Completion requirements**:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

olekzabl:
#### What type of PR is this?

/kind feature
/area multikueue

It adds a feature which automatically updates MultiKueue manager ClusterQueue quotas, based on summing up worker-side quotas.

Fixes #9988 _but only in Alpha1 scope_ - I'm not sure how to track this.

* Heavily vibe-coded, but also heavily supervised.

* **Informers** for remote clusters had to be introduced to avoid triggering the logic by irrelevant remote CQ changes.
  (Plain "Watch" gives no insight into the previous state, hence no insight into "what has changed").
  Given that, I went a step further and created `SelectivelyCachingClient` to achieve [this follow-up](https://github.com/kubernetes-sigs/kueue/tree/main/keps/9988-multikueue-manager-quota-automation#use-cached-remote-clients-for-low-volume-resource-kinds) from the KEP.
  As a result, all **remote** `Get`s and `List`s on CQs and LQs are **cached**, which optimizes the feature even further.

* The test coverage is not yet in the full KEP-planned scope.
  However, I think it's already decent, so I'm sending it for review now.
  More tests can be added in follow-up PRs. (Or, if anyone insists, I can add them also in this PR).
  * Tests currently provided:
    - unit tests for various exit paths from `Reconcile()`
    - unit tests for triggering MK CQ Reconciler by AC / MKConfig / MKCluster changes
    - integration tests verifying MK CQ Reconciliation on remote CQ/LQ changes
    - integration tests verifying that MK CQ Reconciler is inactive on FG disabled
    - basic e2e test "it essentially works"
  * Tests still TODO:
    - e2e tests besides the minimal scope (see [KEP test plan](https://github.com/olekzabl/kueue/blob/d3f1456215d5007321f511254f38ba49a409da1e/keps/9988-multikueue-manager-quota-automation/README.md#e2e-tests))
 
```release-note
MultiKueue: Keep manager ClusterQueue quota equal to the sum of workers' quotas. This is enabled by the `MultiKueueManagerQuotaAutomation` feature gate, and by `.QuotaManagement` in MultiKueueConfig.
```


pkg/controller/admissionchecks/multikueue/clusterqueue.go
```diff
@@ -0,0 +1,460 @@
...
+		helper:      helper,
+		clusters:    clusters,
+		roleTracker: roleTracker,
+	}
+}
+
+func (r *CQReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	if !features.Enabled(features.MultiKueueManagerQuotaAutomation) {
+		return reconcile.Result{}, nil
+	}
```
kshalot: Is this needed? This check is performed before setting up the controller anyway.

pkg/controller/admissionchecks/multikueue/clusterqueue.go
```diff
@@ -0,0 +1,460 @@
...
+
+	log := ctrl.LoggerFrom(ctx).WithValues("clusterQueue", req.Name)
+	log.V(3).Info("Reconcile ClusterQueue event received (in the MultiKueue controller)")
```
kshalot, line 72: I think logs like this are perfomed automatically by controller-runtime.

pkg/controller/admissionchecks/multikueue/clusterqueue.go
```diff
@@ -0,0 +1,460 @@
...
+
+	ac, hasAC, err := r.getMultiKueueAdmissionCheck(ctx, cq)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	if !hasAC {
+		log.V(3).Info("Not a MultiKueue manager ClusterQueue, skipping reconcile.")
+		err := r.removeQuotaAutomationCondition(ctx, cq)
+		return reconcile.Result{}, err
+	}
```
kshalot, line 87: Is this handling the case where a cluster queue stops being a MK cluster queue? I guess it's possible in theory.

I'm thinking about this because this check could be moved to the predicates so avoid doing reconciles for every CQ in the cluster. I don't think it's a big deal though, because there aren't that many and this is an opt-in controller.

pkg/controller/admissionchecks/multikueue/clusterqueue.go
```diff
@@ -0,0 +1,460 @@
...
+
+	lqKeys := sets.New[types.NamespacedName]()
+	for _, llq := range localLQs.Items {
+		lqKeys.Insert(types.NamespacedName{Namespace: llq.Namespace, Name: llq.Name})
+	}
+
+	for _, clusterName := range cfg.Spec.Clusters {
+		rc, found := r.clusters.controllerFor(clusterName)
+		if !found || rc.connecting.Load() {
```
kshalot: I think handling `rc.connecting` is still missing:

> To cover for that, we could increase the manager-side quota by the sum of requests of all workloads which are admitted on the manager but their assigned worker clusters have remote clients which are currently `.connecting`. (Notably, all this info is available on the manager side, without querying the workers).

pkg/controller/admissionchecks/multikueue/clusterqueue.go
```diff
@@ -0,0 +1,460 @@
...
+	if err := r.client.List(ctx, localLQs, client.MatchingFields{indexer.QueueClusterQueueKey: cq.Name}); err != nil {
+		return nil, fmt.Errorf("listing local LocalQueues: %w", err)
+	}
+
+	lqKeys := sets.New[types.NamespacedName]()
+	for _, llq := range localLQs.Items {
+		lqKeys.Insert(types.NamespacedName{Namespace: llq.Namespace, Name: llq.Name})
+	}
+
+	for _, clusterName := range cfg.Spec.Clusters {
```
kshalot: Since this is not cached (https://github.com/kubernetes-sigs/kueue/issues/10629) and actually works over a netowrk, I'd consider aggregating this in parallel.

pkg/controller/admissionchecks/multikueue/clusterqueue.go
olekzabl: Yes - because this is now a part of "Future work ideas" in the KEP.
So postponed deliberately, for simplicity.

apis/kueue/v1beta2/multikueue_types.go
```diff
@@ -155,8 +155,33 @@ type MultiKueueConfigSpec struct {
 	// +kubebuilder:validation:items:MaxLength=256
 	// +required
 	Clusters []string `json:"clusters,omitempty,omitzero"`
+
+	// quotaAutomation specifies the automation behavior of ClusterQueue quotas 
+	// in the manager cluster. 
+	// If unspecified, the default mode depends on the MultiKueueManagerQuotaAutomation feature gate.
+	// +optional
+	QuotaAutomation *QuotaAutomation `json:"quotaAutomation,omitempty"`
```
kshalot: I remember a discussion in the KEP about the where this API should be and some related things like handling constraints that were not met etc.

So a quick question for a sanity check - do we even need to commit to this minimal Manual/Automated API for the alpha? We could just automate for every MK CQ when the fg is enabled to gather initial feedback.

olekzabl: FG is cluster-wide; this flip is configurable per CQ.
I think we want to have it ultimately, so exposing in Alpha enables gathering early feedback.

While the KEP speculates that we may move it (from MKConfig to CQSpec), the field name & values are considered stabilized (once KEP is approved :), so it's even more worth exposing for feedback.

pkg/controller/admissionchecks/multikueue/clusterqueue.go
olekzabl: Great point. Done.
(Btw also changed error handling)

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
```diff
@@ -108,10 +109,11 @@ type remoteClient struct {
 	builderOverride clientWithWatchBuilder
 }
 
-func newRemoteClient(localClient client.Client, wlUpdateCh, watchEndedCh chan<- event.GenericEvent, origin, clusterName string, adapters map[string]jobframework.MultiKueueAdapter) *remoteClient {
+func newRemoteClient(localClient client.Client, wlUpdateCh chan<- event.GenericEvent, cqUpdateCh chan<- event.TypedGenericEvent[kueue.ClusterQueueReference], watchEndedCh chan<- event.GenericEvent, origin, clusterName string, adapters map[string]jobframework.MultiKueueAdapter) *remoteClient {
```
Singularity23x0: You do not have to separate wlUpdateCh and watchEndedCh,  cqUpdateCh can be defined afterwards, reducing signature length.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
```diff
@@ -175,25 +177,35 @@ func (rc *remoteClient) setConfig(watchCtx context.Context, config *clientConfig
 	rc.client = remoteClient
 
 	watchCtx, rc.watchCancel = context.WithCancel(watchCtx)
-	err = rc.startWatcher(watchCtx, kueue.GroupVersion.WithKind("Workload").GroupKind().String(), &workloadKueueWatcher{})
-	if err != nil {
+	
+	handleConnectionError := func(err error) (*time.Duration, error) {
 		rc.failedConnAttempts++
 		return new(retryAfter(rc.failedConnAttempts)), err
 	}
```
Singularity23x0: Cleaner:
```go
func (rc *remoteClient) handleConnectionError(err error) (*time.Duration, error) {
    rc.failedConnAttempts++
    return new(retryAfter(rc.failedConnAttempts)), err
}
```

pkg/controller/admissionchecks/multikueue/clusterqueue.go
olekzabl: > Is this handling the case where a cluster queue stops being a MK cluster queue? I guess it's possible in theory.

Yes. I just checked this in practice - you can remove the MK AC from an existing CQ.

And in such case, I want the quota automation condition removed, which feels clearly not fit for a predicate.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
olekzabl: I considered this before, and my view was different.

For me, `func (rc *remoteClient) handleConnnectionError` sounds like some "general connection error handler for the whole `remoteClient`, which tries fixing stuff, etc.". While here it's just for the internal purpose of this single method.

That's why I deliberately wanted it nested.

WDYT?

pkg/controller/admissionchecks/multikueue/clusterqueue.go
olekzabl: Right.

Let me remove it when I add some test (not unit) verifying the outer guard.
(Or make sure the existing tests verify it already - which may also be the case)
For now, I'm treating the outer guard as not tested, and feeling somewhat safer having another here.

pkg/controller/admissionchecks/multikueue/clusterqueue.go
olekzabl: Looking around, many our `Reconcile()`s have such entry logs also, typically at `V(2)`.
(In this case, it's `V(3)` due to the initial triaging; I'll emit `V(2)` once we know it's a MK CQ).

Maybe, if these auto-logs from controller-runtime exist, such manual logs are still considered more readable?

pkg/controller/admissionchecks/multikueue/clusterqueue.go
olekzabl: Done.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
olekzabl: Done.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
olekzabl: Note: this got obsolete after rebase. (A nice named helper has been added in the meantime).

test/integration/multikueue/setup_test.go
```diff
@@ -1160,3 +1169,93 @@ var _ = ginkgo.Describe("MultiKueue with ClusterProfile", ginkgo.Label("area:mul
...
+				multikueue.WithAdapters(adapters),
+				multikueue.WithDispatcherName(config.MultiKueueDispatcherModeAllAtOnce),
+			)
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+			failedWebhook, err := webhooks.Setup(mgr, nil)
+			gomega.Expect(err).ToNot(gomega.HaveOccurred(), "webhook", failedWebhook)
+		}, framework.WithNewCache(newCacheFunc))
+
+		time.Sleep(3 * time.Second)
```
mimowo: Don't use custom sleeps, if waiting is needed use Eventually to finalize waiting as soon as the condition is met

test/integration/multikueue/setup_test.go
```diff
@@ -1160,3 +1169,93 @@ var _ = ginkgo.Describe("MultiKueue with ClusterProfile", ginkgo.Label("area:mul
...
+
+		gomega.Expect(ic.hasWatch("*v1beta2.ClusterQueue")).To(gomega.Equal(spec.wantWatched))
```
mimowo: Checking the watch registration seems very low level, why not check something user observable, like the state of the new condition?

pkg/controller/admissionchecks/multikueue/indexer.go
```diff
@@ -86,5 +108,11 @@ func SetupIndexer(ctx context.Context, indexer client.FieldIndexer, configNamesp
 	if err := indexer.IndexField(ctx, &kueue.AdmissionCheck{}, AdmissionCheckUsingConfigKey, admissioncheck.IndexerByConfigFunction(kueue.MultiKueueControllerName, configGVK)); err != nil {
 		return fmt.Errorf("setting index on admission checks config: %w", err)
 	}
+	if err := indexer.IndexField(ctx, &kueue.AdmissionCheck{}, AdmissionCheckControllerNameKey, admissionCheckControllerNameIndexerFunc); err != nil {
+		return fmt.Errorf("setting index on admission checks controllerName: %w", err)
+	}
+	if err := indexer.IndexField(ctx, &kueue.ClusterQueue{}, ClusterQueueAdmissionChecksKey, clusterQueueAdmissionChecksIndexerFunc); err != nil {
+		return fmt.Errorf("setting index on cluster queues admissionChecks: %w", err)
+	}
```
mimowo: Add new indexers behind the new FG

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
```diff
@@ -277,6 +299,58 @@ func (rc *remoteClient) startWatcher(ctx context.Context, kind string, w jobfram
...
+
+	return nil
+}
+
+func (rc *remoteClient) queueEventsForCQ(ctx context.Context, remoteCQ *kueue.ClusterQueue) {
+	log := ctrl.LoggerFrom(ctx).WithValues("remoteCQ", remoteCQ.Name)
+
+	// Remote clients aren't indexed, so we fetch all remote LQs.
+	remoteLQs := &kueue.LocalQueueList{}
+	if err := rc.client.List(ctx, remoteLQs); err != nil {
```
mimowo: Hm, I'm wondering if this can be too heavy to call listing all LQ from remote. There might be >1000 of them for sure. Cannot we restrict that in some way?

test/integration/multikueue/setup_test.go
olekzabl: Because this test is intended to verify that "FG = disabled" does not increase even the traffic (no watches).

If I verified "sth user observable", it could still be the case that the watches are there, and the controller is running, just skipping its updates.

test/integration/multikueue/setup_test.go
olekzabl: When expecting the watch to have happened, I could indeed use `Eventually`.

However, when expecting the watch **not** to have happened, when FG = disabled (which is what I really care about in this test), I don't see a way to employ `Eventually` in the right way. 
Initially "watch has happened" will be certainly false, and I don't have a good "hook" to wait until.

From what I see, such cases are often handled with `Consistently`, which in this case is actually equivalent to a custom sleep & check then. (Because, once "watch has happened", it will never "un-have-happened").

pkg/controller/admissionchecks/multikueue/clusterqueue_test.go
```diff
@@ -0,0 +1,670 @@
...
+					}
+					if cond.Message != tc.wantMessage {
+						t.Errorf("expected message: %q, got: %q", tc.wantMessage, cond.Message)
+					}
+				}
+			} else {
+				if cond != nil {
+					t.Errorf("expected status condition to be absent, got: %v", cond)
+				}
+			}
```
mimowo: nit, but I would prefer to use cmp.Diff to make the assert less verbose and more controlled

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
```diff
@@ -212,26 +220,45 @@ func (rc *remoteClient) setConfig(watchCtx context.Context, config *clientConfig
 		if !implementsWatcher {
 			continue
 		}
-		err := rc.startWatcher(watchCtx, kind, watcher)
+		err := rc.startGenericWorkloadWatcher(watchCtx, kind, watcher)
```
mimowo: Something is off here: we call `startGenericWorkloadWatcher` so the name suggests we have "Workloads", but OTOH it seems that "kind" can be any type which is a key in the `rc.adapters` map. So either the function is mis-named, or we call it incorrectly.

olekzabl: Please note the name says "**generic** workload". 
I intended it to be analogous to "**generic** job", i.e. a much broader class than a `Job`.

This function is supposed to be called:
- for `Workload` 
- for any generic job kind that is registered in `rc.adapters`.

For that, "generic workload" was the best naming I could come up with.
Any other ideas?

mimowo: Ah, because it is called "Generic" Workload, so it encompasses both "Workload" and various job Kinds. This makes sense, but I'm wondering if we could make it more explicit.

olekzabl: `startWorkloadOrGenericJobWatcher` ?

(another race condition :) )

mimowo: race condition, I realized soon after posting :) no better idea for now

mimowo: Rather `startWorkloadOrJobWatcher` just, because GenericJob suggests it is for the "GenericJob" interface, and this interface AFAIK does not include currenlty some types like LWS or StatefulSet, whereas the watcher is taken for them as well.

apis/kueue/v1beta2/clusterqueue_types.go
```diff
@@ -22,6 +22,16 @@ import (
...
 )
 
+// ClusterQueue status condition types.
+const (
+	// ClusterQueueActive indicates that the ClusterQueue can admit new workloads and its quota
+	// can be borrowed by other ClusterQueues in the same cohort.
+	ClusterQueueActive string = "Active"
+	// MultiKueueManagerQuotaAutomation indicates that this ClusterQueue is a
+	// MultiKueue manager queue and its quota is automatically managed.
+	MultiKueueManagerQuotaAutomation string = "MultiKueueManagerQuotaAutomation"
```
kshalot, line 32: Nit: One one hand I like the fact that the `MultiKueueManagerQuotaAutomation` type is consistent with the FG, on the other I think it's more natural to word conditions as an adjective, like `AutomatedMultiKueueManagerQuota`. WDYT?

test/integration/multikueue/setup_test.go
mimowo: I see, in that case I would suggest to actually stop using DescribeTable.

When we have two separate tests we can easily use Eventuaully in one, and Consistently in the other without any "if's". 

I know this would be just a single "if" now, so maybe let's keep "DescribeTable" but my practice so far shows that the "DescribeTable" are notoriously hard to evolve when the paths diverge as it results in more and more "ifs".

test/integration/multikueue/setup_test.go
mimowo: I see, this makes sense, let's keep the asserts then, but still I would extend the tests with condition checks.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
kshalot: The relevant interface is called `MultiKueueWatcher` so also potentially `startMultiKueueWatcher`? But that kind of reads as "start some generic watcher for MultiKueue".

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
olekzabl: I think not easily.

The problem: `rc` is a remote client, with:
- no caching (and hence no indexing)
- no support for API-side filtering by `.Spec.ClusterQueue` because it's not listed among `selectableFields` in the LocalQueue CRD.

My preferred way forward would be to make this remote client cached, so that it can be indexed.
However, I'd rather *not* make it cached for all types (given that Workloads volume can be high).
I'd rather make it selective per kind - e.g. for LQs and CQs, but not Workloads. 
This is exactly [this future work idea](https://github.com/kubernetes-sigs/kueue/tree/main/keps/9988-multikueue-manager-quota-automation#use-cached-remote-clients-for-low-volume-resource-kinds) from the KEP.

So - my proposal is to:

- leave it as is for Alpha, where those "maybe heavy" `List`s are hidden behind FG anyway
- plan for kind-selective caching for Beta

WDYT?

mimowo: I see, thanks for explaining. 

For now I would just suggest adding some comment (maybe even TODO) that this is likely sub-optimal and will need to be revist if we observe it becomes a bottlneck.

pkg/controller/admissionchecks/multikueue/remote_fetch.go
kshalot, line 1: This could potentially be a more "global" util in `pkg/util`, since this might be useful for all clients without caching (although I can't think of any case apart from MultiKueue for now.

apis/kueue/v1beta2/clusterqueue_types.go
olekzabl: I think I understand the point.

However, note that your proposal is also, strictly syntactically, a noun phrase :) ("quota" being root of it).

To make it truly "adjectival", we'd need to say `HavingX`, `AutomatedByX`, `WithXEnabled`... but in my eyes it'd make the whole name just too long and harder to parse. While this nominal `...Automation` feels to me easy to grasp: of course it's about "being affected by that thing".

WDYT?

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
mimowo: I think it would be nice to have some kind of alias for string in the context of MultiKueue jobs, like `MultiKueueJob`, similarly as we have ClusterQueueReference etc. Then we could have two functions like `startWorkloadWatcher` and `startMultiKueueJobWatcher` which use strong typing. However, this is largely a pre-existing concern for what I can see, so it is not a blocker to me. Still, some good ideas for long term are welcome ofc.

test/integration/multikueue/setup_test.go
olekzabl: In my eyes, the main value of DescribeTable here is that it verifies that the test for the "disabled" case actually tests the right thing.

That's because the crucial line in this test looks like this:
```golang
gomega.Expect(ic.hasWatch("*v1beta2.ClusterQueue")).To(...)
```

And now, how do we know that `*v1beta2.ClusterQueue` is the right value to be there?
How do we ensure that this test remains meaningful e.g. when `ClusterQueue` gets migrated to `v1beta3`?
The best way to know is to run this test in the "enabled" case, and verify that the result is different.

But now, if I "fork" this into separate "enabled" & "disabled" stories, I expect the "enabled" story to evolve like this: "here, let's check the user-observable outcome, and don't care about this watch existence, why would we care about that".

And that way, we'll leave the "disabled" story without a nice self-verification.

I'm still thinking how to incorporate your feedback without losing this bit.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
kshalot: I also have a hunch that this might be heavy, because IIUC this call will be made for every change in any CQ on any worker. I think those events won't be deduped, only the events in the workqueue (`cqUpdateCh`) will be deduped, which is after the List is called. So this List call would be made constantly.

apis/kueue/v1beta2/clusterqueue_types.go
kshalot: You are right, my proposal is not strictly accurate. It is easier to tie it with the status though (true/false).

`MultiKueueManagerQuotaAutomation: true`
vs
`AutomatedMultiKueueManagerQuota: true`
or even
`MultiKueueManagerQuotaAutomated: true`

It sounds closer to other conditions like `QuotaReserved`. But I don't feel that strongly about it to be honest.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
kshalot: Maybe instead of opening a watch directly we could use an informer here?

test/integration/multikueue/setup_test.go
mimowo: Maybe introduce a helper like hasCQWatch(), and check it `gomega.Expect(hasCQWatch()).To(...)`, expecting once "false", one "true"?

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
mimowo: I think ideally we maintain the CQ and LQ caches locally, based on the "watch" events. However, this might be tricky to implement. Certainly not required for Alpha. 

However, maybe there is an easily to implement approach which would already mitigate the problem with constant heavy Lists, for example a cached list with fixed expiration time, like 1s?

kshalot: I think we can create an Informer using https://pkg.go.dev/k8s.io/client-go/informers and use predicates instead of callbacks for CQs only to start. I never used it directly, but it doesn't look like a huge change in the current implementation.

olekzabl: > the problem with constant heavy Lists

Just to set the stage: I think "**constant**" is the problem (while "heavy" not so much).

@kshalot just pointed out that the current code is going to run the whole reconciling routine on every change of every remote CQ (even its status!), which is clearly sub-optimal.

Instead, we should make sure work is triggered only on a remote CQ quota change. (Or creation, or deletion). I think that, once we ensure that, even "heavy" `List` will become acceptable for Alpha.

So I think Alpha needs some caching. Maybe with Informers.
(And then, once we have this, heavy Lists won't be a heavy problem, and may be solvable with that as well).

test/integration/multikueue/setup_test.go
olekzabl: Yes, for now I did sth similar by externalizing the string constant.

pkg/controller/admissionchecks/multikueue/indexer.go
olekzabl: Done

pkg/controller/admissionchecks/multikueue/clusterqueue_test.go
olekzabl: Done

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
olekzabl: > Rather `startWorkloadOrJobWatcher`

Done.

> The relevant interface is called `MultiKueueWatcher` so also potentially `startMultiKueueWatcher`? 

I made it agree in the other way - renamed `MultiKueueWatcher` to `MultiKueueWorkloadOrJobWatcher`.

olekzabl: > I think it would be nice to have some kind of alias for string (...) strong typing (...)

I'm not following. 

Which `string` would you like to alias?
What kind of "stronger typing" are you hoping to achieve, compared to what we have now?

pkg/controller/admissionchecks/multikueue/remote_fetch.go
olekzabl: Ack. Though my longer-term vision is to get CQ & LQ cached for remotes and get rid of this helper.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
mimowo: I mean that the `kind` is a string parameter which makes it possible to pass to `startWorkloadOrJobWatcher` virtually anything (while only Workload, or Job kind names make sense), but interestingly it is not used for anything except just logging.

olekzabl:
@mimowo @kshalot PTAL

FWIW **summary of main changes** since our yesterday's discussion (specifically since commit [ec6313e](https://github.com/kubernetes-sigs/kueue/pull/11141/commits/ec6313e84a7337729a6e3ae5224634714879f31d)):

1. Made `rc.client` smarter (see [`remote_client.go`](https://github.com/olekzabl/kueue/blob/df78244250c419ccb11ec01ab74c1d95bee3349e/pkg/controller/admissionchecks/multikueue/remote_client.go)) - so that it:
   - Uses informer cache for [past-state-aware watching](https://github.com/olekzabl/kueue/blob/df78244250c419ccb11ec01ab74c1d95bee3349e/pkg/controller/admissionchecks/multikueue/multikueuecluster.go#L408-L414), but [also `Get` and `List` calls](https://github.com/olekzabl/kueue/blob/df78244250c419ccb11ec01ab74c1d95bee3349e/pkg/controller/admissionchecks/multikueue/remote_client.go#L76-L99)
   - Does that only for specified `cachedKinds` (currently: LQ & CQ); otherwise falls back to the old client.

2. Added [integration tests](https://github.com/olekzabl/kueue/blob/df78244250c419ccb11ec01ab74c1d95bee3349e/test/integration/multikueue/setup_test.go#L783-L866) to verify reaction to remote LQ / CQ changes.

3. Equipped `remoteClient` with an internal `RWMutex`; this made new integration tests happier.

4. Reverted some changes which are now not relevant (or less relevant):
   - the refactor around `startWorkloadOrJobWatcher`, `MultiKueueWatcher` (no longer relevant)
   - the `getOrList` helper (much less relevant due to caching)
   - parallelization in quota computation (much less relevant due to caching)

pkg/controller/admissionchecks/multikueue/remote_fetch.go
olekzabl: This got obsolete and removed.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
olekzabl: FYI I reverted this whole renaming, as (after introducing informer cache) the old watchers became unrelated to this PR.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
olekzabl: Done

test/integration/multikueue/setup_test.go
olekzabl: I added other meaningful test cases higher up in this file.
The new cases check reaction to remote LQ / CQ changes, specifically they check the manager-side quotas.
As for conditions - I felt no much value in checking those; this is already covered extensively by unit tests.

I chose to keep those "user-observable effects" tests separately from the "watcher exists or not" tests, due to diverging test setup. ("User-observable effects" could be easily unified with the main Describe() block; "watcher exists or not" feels quite different).

apis/kueue/v1beta2/clusterqueue_types.go
olekzabl: TBH I still like it like it is, and this is also the alignment from KEP (which so far has attracted more reviewers than this impl).

So I'd prefer to stick to it.

pkg/controller/admissionchecks/multikueue/clusterqueue.go
```diff
@@ -0,0 +1,450 @@
...
+	}
+
+	apimeta.SetStatusCondition(&cq.Status.Conditions, newCondition)
+	if err := r.client.Status().Update(ctx, cq); err != nil {
+		return fmt.Errorf("updating ClusterQueue condition status: %w", err)
+	}
+	return nil
+}
+
+func (r *CQReconciler) queueEventsForAC(ctx context.Context, acName string, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
```
mimowo: We already have the function `ClusterQueuesUsingAdmissionCheck`, any reason not to use it?

pkg/controller/admissionchecks/multikueue/clusterqueue.go
```diff
@@ -0,0 +1,450 @@
...
+	}
+
+	for _, ac := range acList.Items {
+		r.queueEventsForAC(ctx, ac.Name, q)
+	}
+}
+
+func (r *CQReconciler) queueEventsForMKCluster(ctx context.Context, clusterName string, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	configList := &kueue.MultiKueueConfigList{}
+	if err := r.client.List(ctx, configList, client.MatchingFields{UsingMultiKueueClusters: clusterName}); err != nil {
```
mimowo, line 291: do not silence errors, at least log them

pkg/controller/admissionchecks/multikueue/clusterqueue.go
```diff
@@ -0,0 +1,450 @@
...
+	}
+
+	for _, cq := range cqList.Items {
+		q.Add(reconcile.Request{NamespacedName: types.NamespacedName{Name: cq.Name}})
+	}
+}
+
+func (r *CQReconciler) queueEventsForMKConfig(ctx context.Context, configName string, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	acList := &kueue.AdmissionCheckList{}
+	if err := r.client.List(ctx, acList, client.MatchingFields{AdmissionCheckUsingConfigKey: configName}); err != nil {
```
mimowo, line 279: do not silence errors, at least log them

pkg/controller/admissionchecks/multikueue/clusterqueue.go
```diff
@@ -0,0 +1,450 @@
...
+			newCQ, okNew := e.ObjectNew.(*kueue.ClusterQueue)
+			if !okOld || !okNew {
+				return true
+			}
+
+			if equality.Semantic.DeepEqual(oldCQ.Spec.AdmissionChecksStrategy, newCQ.Spec.AdmissionChecksStrategy) &&
+				equality.Semantic.DeepEqual(oldCQ.Spec.ResourceGroups, newCQ.Spec.ResourceGroups) {
+				return false
+			}
+			return true
```
mimowo: nit, but maybe for simplicity let's just check Generation? That should work pretty much the same.

pkg/controller/admissionchecks/multikueue/clusterqueue.go
```diff
@@ -0,0 +1,450 @@
...
+		},
+	}
+
+	remoteHandler := handler.TypedFuncs[kueue.ClusterQueueReference, reconcile.Request]{
+		GenericFunc: func(_ context.Context, e event.TypedGenericEvent[kueue.ClusterQueueReference], q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+			q.Add(reconcile.Request{NamespacedName: types.NamespacedName{Name: string(e.Object)}})
```
mimowo: I would consider using AffAfter with 1s delay day one, if we don't use delay then it becomes trcky to add later.

When the delay is used, then the workqueue deduplicates the Reconcile events, so this is a mechnism for batching "Reconcile" events in time. I'm not clear if we need Reconcile per every event. 

The same comment applies to other places.

pkg/controller/admissionchecks/multikueue/clusterqueue.go
```diff
@@ -0,0 +1,450 @@
...
+	oldAC, okOld := event.ObjectOld.(*kueue.AdmissionCheck)
+	newAC, okNew := event.ObjectNew.(*kueue.AdmissionCheck)
+	if !okOld || !okNew {
+		return
+	}
+
+	if oldAC.Spec.ControllerName == newAC.Spec.ControllerName &&
+		equality.Semantic.DeepEqual(oldAC.Spec.Parameters, newAC.Spec.Parameters) {
+		return
+	}
```
mimowo: ACs are changes really really rarely. I think especially if we add batching, then we could drop this complication, until it proves to be a bottlneck.

pkg/controller/admissionchecks/multikueue/clusterqueue.go
```diff
@@ -0,0 +1,450 @@
...
+	oldConfig, okOld := event.ObjectOld.(*kueue.MultiKueueConfig)
+	newConfig, okNew := event.ObjectNew.(*kueue.MultiKueueConfig)
+	if !okOld || !okNew {
+		return
+	}
+
+	if equality.Semantic.DeepEqual(oldConfig.Spec.Clusters, newConfig.Spec.Clusters) &&
+		equality.Semantic.DeepEqual(oldConfig.Spec.QuotaManagement, newConfig.Spec.QuotaManagement) {
+		return
+	}
```
mimowo: Similarly here, this looks like optimization of a non-hot path.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
```diff
@@ -149,15 +169,35 @@ func newRemoteClient(localClient client.Client, wlUpdateCh, watchEndedCh chan<-
...
 	}
-	restConfig, err := clientcmd.RESTConfigFromKubeConfig(config.Kubeconfig)
+
+	directClient, err := client.NewWithWatch(restConfig, options)
 	if err != nil {
 		return nil, err
 	}
-	return client.NewWithWatch(restConfig, options)
+
+	if !features.Enabled(features.MultiKueueManagerQuotaAutomation) {
```
mimowo, line 180: Actually, I think it would be nice to have a separate FG for this like `MultiKueueSelectivelyCachingClient` so that we can drop the NeverCachingClient prior to graduation of `MultiKueueManagerQuotaAutomation`? That could be a follow up PR to me, just raising for now.

mimowo (commented):
LGTM overall, left some comments mostly to make sure if we can simplify the code.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
olekzabl: Maybe, but for now it's anyway set up to cache only for remote LQs & CQs, and I think these aren't touched by anyone except the quota automation feature.

So, once we identify some other "lightweight" resource types which we may to "enroll into" Selective Caching (IMO certainly not Workloads, likely not generic job types), then I'll be happy to come back to this question.

pkg/controller/admissionchecks/multikueue/clusterqueue.go
olekzabl: This opens a general interesting question: should MultiKueue controllers have access to scheduler cache?
Currently, they don't. I always thought I wouldn't dare changing this for such a minor reason.

My problem with "just passing `Cache`" to MK controllers is that `Cache` exposes many mutating methods.
So doing so would boldly cross an (apparently deliberate?) black-boxing barrier between Kueue core & MultiKueue.

Perhaps we could define some `CacheReader` interface, and expose just that?
Though if so, I wouldn't do this in this PR. It feels a large change on its own, very weakly related.

WDYT?

pkg/controller/admissionchecks/multikueue/clusterqueue.go
olekzabl: Done.

pkg/controller/admissionchecks/multikueue/clusterqueue.go
olekzabl: Done.

pkg/controller/admissionchecks/multikueue/clusterqueue.go
olekzabl: Done.
Instead of hardcoding 1s, I reused the injectable `eventsBatchPeriod`, like in MultiKueue Workload Reconciler.

pkg/controller/admissionchecks/multikueue/clusterqueue.go
olekzabl: Done - but I have mixed feelings about this.

* On one hand: true, this removed a few lines of code.
* OTOH: simplicity for me is also about following some conventions. Like "when feasible, trigger on update only when needed, by comparing old vs. new".
  Otherwise, our convention seems to be this: "For every resource kind, discuss how often it tends to change".
  And that doesn't feel simple to me, either :)

pkg/controller/admissionchecks/multikueue/clusterqueue.go
olekzabl: Done - but... (similar as [here](https://github.com/kubernetes-sigs/kueue/pull/11141#discussion_r3269823708))

pkg/controller/admissionchecks/multikueue/clusterqueue.go
olekzabl: Done - but I think I don't like it:

* It effectively removes just 1 line of code.
* It obscures the true intent behind this code. (We lose knowledge which fields are relevant).
* It makes me feel unsafe for the future: `ClusterQuotaSpec` has already plenty of fields, for which I can't assess "what's the likelihood field X will start changing much more frequently than AC/Quotas, and more frequently than we'd like to reconcile here". Plus, new troubling fields might be added in the future.

WDYT?

olekzabl:
/retest

The last test failure is unrelated to this PR, and should be fixed by #11338.

(Edit: indeed, it's fixed).

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
mimowo: Yes, but my point is that since MultiKueueSelectivelyCachingClient is a generalization of the neverCachingClient, then we could use it unconditionally for both cases. We can defer the decision for a bit later.

pkg/controller/admissionchecks/multikueue/clusterqueue.go
mimowo: > This opens a general interesting question: should MultiKueue controllers have access to scheduler cache?

Great comment. Historically we thought that maybe MultiKueue could become a fully independent AC (subproject), and so we decoupled it as much as possible. However, in the last 2 years the coupling has strengthened. My comment was because I now started to assume the coupling is so strong that we already have access to the cache - I didn't check that. However, if we don't have access yet let's keep it that way.

The extra indexer is not too bad, especially as it avoids making significant changes for Alpha1. 

> Perhaps we could define some CacheReader interface, and expose just that?
> Though if so, I wouldn't do this in this PR. It feels a large change on its own, very weakly related.

I like this idea, and we could consider it as a separate PR, but for some another already pre-existing component which we consider "core". 

MultiKueue is still in the gray area of semi-independent project. So, let's not increase the coupling for now.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
mimowo: sgtm

pkg/controller/admissionchecks/multikueue/clusterqueue.go
mimowo: Ok, I can see this perspective, and this is reasonable. I will certainly keep open to this perspective. Feel free to follow up towards this direction in a follow up PR, so that it is easier to review.

pkg/controller/admissionchecks/multikueue/clusterqueue.go
mimowo: This is similar to https://github.com/kubernetes-sigs/kueue/pull/11141#discussion_r3270426163. If you prefer this way, then I'm open to the follow up PR, even before the release, but separate PR so that it is easy to review and see the consequences - up to you really. I can see the trade off, I primarily left the comment to  minimize this PR, and double check on you  - is it really improving readability vs. premature optimization. Assessing that in a follow up PR with a small diff will be much easier.

pkg/controller/admissionchecks/multikueue/clusterqueue.go
mimowo: Yes, same answer as in the other thread.

mimowo:
/lgtm
/approve
Thank you for delivering this important feature in the Alpha1 for 0.18. The remaining dicussions are non-blocking and I'm certainly open to the follow up PRs if you think this is the right direction - raised the comments as "let's take a deep breath and re-evaluate this is really improving readability, or risking premature optimization". Seeing the changes in dedicated small PRs will let us assess the tradeoffs much better - up to you.

PR #11144, opened 2026-05-12
Use ClusterQueueReference wrappers.MakeAdmission instead of a string

### LINKED ISSUE #10133: Use strong typing for wrappers.MakeAdmission to make sure all tests use it correctly
**What happened**:

Some integration tests are buggy because they pass LocalQueue to MakeAdmission rather than CLusterQueue as expected by the signature:
https://github.com/kaisoz/kueue/blob/aca13b5963a15ad2f1a10bf29e6e54b474c68493/pkg/util/testing/v1beta2/wrappers.go#L677

Example test which is passing LocalQueue name instead: https://github.com/kaisoz/kueue/blob/9e82591ef5522a249723b3644590b8f17ee83709/test/integration/singlecluster/controller/jobs/jobset/jobset_controller_test.go#L324

This sometimes works by "luck", but it makes the tests very hard to reason about.

Use strong typing for the MakeAdmission to accept ClusterQueueReference rather than a string, so that all tests use it correctly.

kaisoz:
#### What type of PR is this?

/kind cleanup
/area testing

Some integration tests were passing `LocalQueue.Name` to `MakeAdmission` instead of `ClusterQueueReference` as expected by the function signature. This PR refactors the `MakeAdmission` function to expect a typed variable (`ClusterQueueReference`) to avoid this.

Fixes #10133

```release-note
NONE
```

### DIFF
--- a/pkg/util/testing/v1beta2/wrappers.go
+++ b/pkg/util/testing/v1beta2/wrappers.go
@@ -149,7 +149,7 @@ func (w *WorkloadWrapper) Active(a bool) *WorkloadWrapper {
 
 // SimpleReserveQuota reserves the quota for all the requested resources in one flavor.
 // It assumes one podset with one container.
-func (w *WorkloadWrapper) SimpleReserveQuota(cq, flavor string, now time.Time) *WorkloadWrapper {
+func (w *WorkloadWrapper) SimpleReserveQuota(cq kueue.ClusterQueueReference, flavor string, now time.Time) *WorkloadWrapper {
 	admission := MakeAdmission(cq, w.Spec.PodSets[0].Name)
 	resReq := make(corev1.ResourceList)
 	flavors := make(map[corev1.ResourceName]kueue.ResourceFlavorReference)
@@ -690,9 +690,9 @@ func (p *PodSetWrapper) ResourceClaim(claimName, resourceClaimName string) *PodS
 // AdmissionWrapper wraps an Admission
 type AdmissionWrapper struct{ kueue.Admission }
 
-func MakeAdmission(cq string, podSetNames ...kueue.PodSetReference) *AdmissionWrapper {
+func MakeAdmission(cq kueue.ClusterQueueReference, podSetNames ...kueue.PodSetReference) *AdmissionWrapper {
 	wrap := &AdmissionWrapper{kueue.Admission{
-		ClusterQueue: kueue.ClusterQueueReference(cq),
+		ClusterQueue: cq,
 	}}
 
 	if len(podSetNames) == 0 {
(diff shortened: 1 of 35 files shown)


kaisoz:
The real changes are in `pkg/util/testing/v1beta2/wrappers.go`, the rest  are just adapting the tests.

pkg/util/testing/v1beta2/wrappers.go
```diff
@@ -690,9 +690,9 @@ func (p *PodSetWrapper) ResourceClaim(claimName, resourceClaimName string) *PodS
 // AdmissionWrapper wraps an Admission
 type AdmissionWrapper struct{ kueue.Admission }
 
-func MakeAdmission(cq string, podSetNames ...kueue.PodSetReference) *AdmissionWrapper {
+func MakeAdmission(cq kueue.ClusterQueueReference, podSetNames ...kueue.PodSetReference) *AdmissionWrapper {
```
mbobrovskyi, line 693: I’m actually not sure about this. Do we really need it? Right now, we need to wrap with kueue.ClusterQueueReference in every invocation. For the wrapper, I would prefer to keep it as a string since this is just for tests.

kaisoz: This is to avoid situations [like this one](https://github.com/kaisoz/kueue/blob/9e82591ef5522a249723b3644590b8f17ee83709/test/integration/singlecluster/controller/jobs/jobset/jobset_controller_test.go#L324) where we pass the `localQueue` name instead of the cluster queue name (in this case it works by pure luck... 🤷🏻)  Using a typed name in the function signature ensures that this cannot happen

test/integration/singlecluster/controller/jobs/appwrapper/appwrapper_controller_test.go
```diff
@@ -442,7 +442,7 @@ var _ = ginkgo.Describe("AppWrapper controller", ginkgo.Ordered, ginkgo.Continue
 			})
 
 			ginkgo.By("admit the workload", func() {
-				admission := utiltestingapi.MakeAdmission(clusterQueueAc.Name).
+				admission := utiltestingapi.MakeAdmission(kueue.ClusterQueueReference(clusterQueueAc.Name)).
```
mbobrovskyi, line 445: What about this?

mbobrovskyi: Ah, I see. This is cq with checks. A bit confusing naming.

mbobrovskyi (commented):
Thank you!

/lgtm
/approve

/cherry-pick release-0.17
/cherry-pick release-0.16

mimowo:
@kaisoz please prepare the CPs

PR #11147, opened 2026-05-12
fix: avoid duplicate tolerations when one uses the default Operator

benkermani:
## What does this PR do?

`PodSetInfo.Merge` deduplicates a `ResourceFlavor`'s tolerations against the workload's existing ones via struct `==`, which treats `Operator: ""` and `Operator: "Equal"` as different , while in the k8s API it defaults the empty operator to `Equal`. 

A workload toleration written one way and a flavor toleration written the other slip through dedup and both get appended
```
# e.g
# RF
tolerations:
  - effect: NoSchedule
    key: some-toleration-key
    value: some-toleration-value
    operator: Equal

# Job
tolerations:
  - effect: NoSchedule
    key: some-toleration-key
    value: some-toleration-value
```

which causes the next Pod/Job update to be rejected by `validateOnlyAddedTolerations` in clusters where a mutating webhook collapses the duplicate between Kueue's read and write — leaving pods `scheduleGated`.

Kueue would emit the following error log:
```
Pod "<name>" is invalid: spec.tolerations: Forbidden: existing toleration can not be modified except its tolerationSeconds
```

This PR normalizes the empty operator to `Equal` before comparing, so the two forms are properly deduplicated.

## Testing

Added the UT without the code change in `pkg/podset/podset.go`, and it was failing locally
```
Unexpected template (-want/+got):
          v1.PodTemplateSpec{
          	ObjectMeta: {Labels: {"l0": "l0v"}, Annotations: {"a0": "a0v"}},
          	Spec: v1.PodSpec{
          		... // 21 identical fields
          		Affinity:      nil,
          		SchedulerName: "",
          		Tolerations: []v1.Toleration{
          			{Key: "t0", Operator: "Equal", Value: "t0v", Effect: "NoSchedule", ...},
        + 			{Key: "t0", Value: "t0v", Effect: "NoSchedule"},
          		},
          		HostAliases:       nil,
          		PriorityClassName: "",
          		... // 16 identical fields
          	},
          }
```

```release-note
Scheduling: Fixed a bug where Kueue could inject duplicate tolerations when a
ResourceFlavor toleration and a PodTemplate toleration differed only by `operator: ""`
versus `operator: "Equal"`, which represent the same Kubernetes toleration. This could
cause update rejections and leave Pods `scheduleGated`.
```

### DIFF
--- a/pkg/podset/podset.go
+++ b/pkg/podset/podset.go
@@ -130,7 +130,9 @@ func (podSetInfo *PodSetInfo) Merge(o PodSetInfo) error {
 
 	// make sure we don't duplicate tolerations
 	for _, t := range o.Tolerations {
-		if slices.Index(podSetInfo.Tolerations, t) == -1 {
+		if !slices.ContainsFunc(podSetInfo.Tolerations, func(e corev1.Toleration) bool {
+			return tolerationsEqual(e, t)
+		}) {
 			podSetInfo.Tolerations = append(podSetInfo.Tolerations, t)
 		}
 	}
@@ -141,6 +143,20 @@ func (podSetInfo *PodSetInfo) Merge(o PodSetInfo) error {
 		}
 	}
 	return nil
+}
+
+// tolerationsEqual reports whether two tolerations describe the same toleration
+func tolerationsEqual(a, b corev1.Toleration) bool {
+	// Normalize Operator before comparing: the Kubernetes API defaults an empty
+	// Operator to "Equal", so Operator: "" and Operator: "Equal" describe the
+	// same toleration and must compare as equal here.
+	if a.Operator == "" {
+		a.Operator = corev1.TolerationOpEqual
+	}
+	if b.Operator == "" {
+		b.Operator = corev1.TolerationOpEqual
+	}
+	return a.MatchToleration(&b)
 }
 
 // AddOrUpdateLabel adds or updates the label identified by k with value v
--- a/pkg/podset/podset.go
+++ b/pkg/podset/podset.go
@@ -33,6 +33,7 @@ import (
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
 	"sigs.k8s.io/kueue/pkg/features"
 	utilmaps "sigs.k8s.io/kueue/pkg/util/maps"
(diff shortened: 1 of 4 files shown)


mimowo:
injected the release-note block with TBD, let me propose:
/release-note-edit
```release-note
Scheduling: Fixed a bug where Kueue could inject duplicate tolerations when a
ResourceFlavor toleration and a PodTemplate toleration differed only by `operator: ""`
versus `operator: "Equal"`, which represent the same Kubernetes toleration. This could
cause update rejections and leave Pods `scheduleGated`.
```

pkg/podset/podset_test.go
```diff
@@ -358,6 +358,17 @@ func TestMergeRestore(t *testing.T) {
 				Obj(),
 			wantRestoreChanges: true,
 		},
+		"don't duplicate tolerations that differ only by empty vs Equal Operator": {
```
mimowo, line 372: Maybe also add the test which works the other way round - the podSet has toleration without Operator, but the info has the toleration with it.

pkg/podset/podset.go
```diff
@@ -143,6 +145,20 @@ func (podSetInfo *PodSetInfo) Merge(o PodSetInfo) error {
 	return nil
 }
 
+// tolerationsEqual reports whether two tolerations describe the same toleration
+func tolerationsEqual(a, b corev1.Toleration) bool {
```
pajakd: For reusability, lets move this function to some utility file under `pkg/util/`. You can create a new one for tolerations. WDYT?

mimowo:
@benkermani the fix looks great and it would be good to include it in the releases tomorrow. Do you have cycles to address the remaining comments?

benkermani:
> @benkermani the fix looks great and it would be good to include it in the releases tomorrow. Do you have cycles to address the remaining comments?

@mimowo I'll try to address your comments now/this afternoon at worst

mimowo:
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16
Thank you 👍 I can confirm that empty operator is the same as Equal: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/toleration.go#L63-L65

PR #5059, opened 2025-04-21
add read only root filesystem for kueue deployment

kannon92:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
enable readOnlyRootFilesystem for kueue deployments.
#### Which issue(s) this PR fixes:

Fixes #

#### Special notes for your reviewer:
```
Whether this container has a read-only root filesystem. Default is false. Note that this field cannot be set when spec.os.name is windows.
```
#### Does this PR introduce a user-facing change?

```release-note
add readOnlyRootFilesystem as default option for kueue deployment.
```

PR #11149, opened 2026-05-12
helm: add /tmp emptyDir volume for visibility server self-signed certs

cixuuz:
## What this PR does

Adds a `/tmp` emptyDir volume mount to the manager deployment template so the visibility server can write its self-signed certificates when `readOnlyRootFilesystem: true` (the Helm chart default) and `enableCertManager: false`.

## Why

When `readOnlyRootFilesystem` is `true` (default since v0.17.2) and cert-manager is disabled, the visibility server tries to write `/tmp/apiserver.crt` but fails with:

```
error creating self-signed certificates: open /tmp/apiserver.crt: read-only file system
```

The chart already provides a `/visibility` emptyDir volume, but the server binary writes to `/tmp` which is on the read-only root filesystem.

## Changes

- Add `volumeMount` for `/tmp` → `tmp` volume in the manager container
- Add `tmp` emptyDir volume to the pod spec

```release-note
Fix visibility server crash with "read-only file system" error when readOnlyRootFilesystem is true and enableCertManager is false by adding a /tmp emptyDir volume mount.
```

### DIFF
--- a/charts/kueue/templates/manager/manager.yaml
+++ b/charts/kueue/templates/manager/manager.yaml
@@ -91,6 +91,8 @@ spec:
         - mountPath: /controller_manager_config.yaml
           name: manager-config
           subPath: controller_manager_config.yaml
+        - mountPath: /tmp
+          name: tmp
         {{- if .Values.enableCertManager }}
         - mountPath: /etc/kueue/metrics/certs
           name: metrics-certs
@@ -151,3 +153,5 @@ spec:
       - name: visibility
         emptyDir: {}
       {{- end }}
+      - name: tmp
+        emptyDir: {}


kannon92:
/hold

I'd really like more research into why `/tmp` is being used instead of `/visibility`. 

Can you also narrow down when this problem started to occur?

cixuuz:
@kannon92 Good question — I dug into this more.

**Why `/tmp` instead of `/visibility`?**

The error path depends on the `internalCertManagement` config:

1. **v0.11.0** used `o.SecureServing.ServerCert.CertDirectory = "/tmp"` unconditionally ([server.go:91 in v0.11.0](https://github.com/kubernetes-sigs/kueue/blob/v0.11.0/pkg/visibility/server.go#L91)), which meant `MaybeDefaultWithSelfSignedCerts` wrote to `/tmp/apiserver.crt`.

2. **v0.17.2** refactored this to use `certDir = "/visibility"` and branches on `InternalCertManagement.Enable`:
   - When `true` (default via Helm when `enableCertManager: false`): sets `CertDirectory = "/visibility"` → generates `/visibility/apiserver.crt` on the emptyDir ✓
   - When `false` (cert-manager provides certs): sets `CertFile`/`KeyFile` to `/visibility/tls.{crt,key}` and expects them to already exist

With default Helm values today, the path should indeed be `/visibility/apiserver.crt` (on the emptyDir), not `/tmp`. The `/tmp/apiserver.crt` error in the issue likely comes from an upgrade scenario where the old v0.11.0 binary was still running briefly against the new security context, or possibly a cached image.

**However**, a `/tmp` emptyDir is still good practice here because:
- Go's `os.TempDir()` returns `/tmp` — any transitive dependency writing temp files will fail with `readOnlyRootFilesystem: true`
- The k8s apiserver library itself uses `os.CreateTemp` in some code paths
- It's the standard pattern for hardened containers (distroless/nonroot images always mount `/tmp`)

**When did this start?** The `readOnlyRootFilesystem: true` default was introduced in v0.17.x (not present in v0.11.0). So anyone upgrading from ≤v0.16.x to v0.17.x with default values and without cert-manager would hit filesystem issues.

Happy to adjust the approach if you'd prefer a different fix — e.g., changing the Go code to use the already-mounted `/visibility` dir as a general temp dir, or gating the `/tmp` mount on `!enableCertManager`.

kannon92:
So `readOnlyRootFilesystem` was added in https://github.com/kubernetes-sigs/kueue/pull/5059 which I think was 0.12.0

I don't know if this occured in `0.17.2`.

kannon92:
So this problem only occurs when you update from 0.11 to 0.17.2?

Can you try a fresh install of 0.17.2?

kannon92:
> However, a /tmp emptyDir is still good practice here because:

> Go's os.TempDir() returns /tmp — any transitive dependency writing temp files will fail with readOnlyRootFilesystem: true
   The k8s apiserver library itself uses os.CreateTemp in some code paths
    It's the standard pattern for hardened containers (distroless/nonroot images always mount /tmp)

I don't really agree with this take. This would essentially be dead code for the mount because it sounds like no volume is writing to /tmp. The only case would be a 0.11 binary (maybe) but a fresh install should be writing to `/visibility`.

cixuuz:
I’m fine that not having this trick. But maybe worthy to document this issue somewhere. User have to handle this manually when upgrading from old version like v0.11. The upgrade will cause downtime to the system.

kannon92:
> I’m fine that not having this trick. But maybe worthy to document this issue somewhere. User have to handle this manually when upgrading from old version like v0.11. The upgrade will cause downtime to the system.

Upgrading from v0.11 to v0.17.2 seems quite difficult to support. 

@mimowo @tenzen-y @PBundyra 
do we provide any guidance on what upgrades we expect to work? I think n -1 -> n (0.16 to 0.17).

mimowo:
> do we provide any guidance on what upgrades we expect to work? I think n -1 -> n (0.16 to 0.17).

Yeah, from N-1 they should work, N-2 is nice to have. Going N-5 no one is testing.

I propose to close this issue to avoid distractions. At this point 0.17 is not supported.

Issue #11006, opened 2026-05-07
Exclude kube-system namespace from webhooks targeting common resource types


dkaluza:
**What would you like to be added**:
Namespace selector for webhooks targeting pods, jobs, and deployments that excludes them from targeting objects created in `kube-system` namespace.

It may be also worth it to exclude kueue namespace.

**Why is this needed**:
Currently when `Kueue` is failing, webhooks will block any pod creation in `kube-system` namespace as they have `FailurePolicy: Fail`. This might result in clusters without possibility of failure recovery as various cluster critical pods update or creation will be blocked.

**Completion requirements**:
Addition of namespace exlusions [here](https://github.com/kubernetes-sigs/kueue/blob/3b5147d997cccb8e61ae39032aeddd5c7157ede9/config/components/webhook/manifests.yaml#L188) and in other related webhooks.

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

dkaluza:
I will be happy to work on this issue if there are no objections from the owners.

/assign

mimowo:
I think this is a good idea overall, but might be breaking so we need to communicate this well.

Also, could you check how this is done for Helm - I'm pretty sure we already disable the webhooks from kube-system and kueue-system namespaces for Helm already. Please check if this is by default there. Then we could probably claim this is a consistency bug.

dkaluza:
Yes it is already available in helm chart - [sources](https://github.com/kubernetes-sigs/kueue/blob/main/charts/kueue/templates/webhook/manifests.yaml#L827), however only if `managedJobsNamespaceSelector` is not set (default state, the commented out value also suggests excluding it - [here](https://github.com/kubernetes-sigs/kueue/blob/main/charts/kueue/values.yaml#L141)).

Exclusion for both `kube-system` and `.Release.Namespace` is set for:
- statefulsets
- pods
- deployments

It is not set for jobs, IMO it would be safer to also add it there - it looks like `kubeadm` runs jobs during cluster upgrades ([source](https://github.com/kubernetes/kubeadm/issues/3050)).

mimowo:
[...]

PR #11155, opened 2026-05-13
Improve docs for namespace management

strigazi:
1. Change the comment in helm values to select labeled namespaces and point to the docs.
2. Remove operator: NotIn values: [ kube-system, kueue-system ] from plain_pods.md
3. In opt_in_namespace_management.md add the pattern of having an explicit list of namespaces managed by kueue.
4. In tasks/manage/productization add a link to opt_in_namespace_management.md. Moving the file would break existing links.

/kind documentation

/area website

Related issue: #11006 

```release-note
NONE
```

### DIFF
--- a/charts/kueue/values.yaml
+++ b/charts/kueue/values.yaml
@@ -138,6 +138,8 @@ managerConfig:
     #    backoffBaseSeconds: 60
     #    backoffMaxSeconds: 3600
     #manageJobsWithoutQueueName: true
+    # See "Opt-in Namespace Management" for guidance on namespace management:
+    # https://kueue.sigs.k8s.io/docs/tasks/manage/enforce_job_management/opt_in_namespace_management/
     #managedJobsNamespaceSelector:
     #  matchExpressions:
     #    - key: kubernetes.io/metadata.name
--- a/site/content/en/docs/tasks/manage/enforce_job_management/opt_in_namespace_management.md
+++ b/site/content/en/docs/tasks/manage/enforce_job_management/opt_in_namespace_management.md
@@ -34,19 +34,43 @@ kubectl label namespace my-namespace managed-by-kueue=true
 
 ### Step 2: Configure the Selector
 
-Configure the `managedJobsNamespaceSelector` in your Kueue Configuration to match the labeled namespaces. Use `matchLabels` to select namespaces that you have explicitly labeled:
+Configure the `managedJobsNamespaceSelector` in your Kueue Configuration with `matchLabels` or `matchExpressions`.
 
-```yaml
-apiVersion: config.kueue.x-k8s.io/v1beta2
-kind: Configuration
-metadata:
-  name: config
-  namespace: kueue-system
-manageJobsWithoutQueueName: true
-managedJobsNamespaceSelector:
-  matchLabels:
-    managed-by-kueue: "true"
-```
+1. Use `matchLabels` to select labeled namespaces:
+
+   ```yaml
+   apiVersion: config.kueue.x-k8s.io/v1beta2
+   kind: Configuration
+   metadata:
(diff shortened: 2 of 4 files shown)


mimowo:
/ok-to-test
cc @dkaluza @kannon92 @amy

kannon92:
/approve



This looks good to me. @mimowo can you give this a quick look? You had some comments on the issue so want to check in with you.

site/content/en/docs/tasks/manage/enforce_job_management/opt_in_namespace_management.md
```diff
@@ -34,19 +34,33 @@ kubectl label namespace my-namespace managed-by-kueue=true
...
+
+   ```yaml
+   managedJobsNamespaceSelector:
+     matchExpressions:
+     - key: kubernetes.io/metadata.name
+       operator: In
+       values: [ production, training, inference ]
+   ```
+
+   Alternatively, you can use the `NotIn` operator for the inverse selection.
```
mimowo: Let's keep this still as an option as this is the default setting.

charts/kueue/values.yaml
```diff
@@ -138,11 +138,10 @@ managerConfig:
...
     #    backoffMaxSeconds: 3600
     #manageJobsWithoutQueueName: true
+    # See "Opt-in Namespace Management" https://kueue.sigs.k8s.io/docs/tasks/manage/enforce_job_management/opt_in_namespace_management/
     #managedJobsNamespaceSelector:
-    #  matchExpressions:
-    #    - key: kubernetes.io/metadata.name
-    #      operator: NotIn
-    #      values: [ kube-system, kueue-system ]
+    #  matchLabels:
+    #    managed-by-kueue: "true"
```
mimowo: As much as I like the approach, maybe it is safer to keep the default here, as I assume many users will just want to uncomment that code, and then it will not work for them  OOTB.

charts/kueue/values.yaml
```diff
@@ -138,11 +138,10 @@ managerConfig:
     #    backoffBaseSeconds: 60
     #    backoffMaxSeconds: 3600
     #manageJobsWithoutQueueName: true
+    # See "Opt-in Namespace Management" https://kueue.sigs.k8s.io/docs/tasks/manage/enforce_job_management/opt_in_namespace_management/
```
mimowo:
```suggestion
    # See "Opt-in Namespace Management" for guide on configuring namespace management: https://kueue.sigs.k8s.io/docs/tasks/manage/enforce_job_management/opt_in_namespace_management/
```

dkaluza:
From my point of view this still shouldn't close https://github.com/kubernetes-sigs/kueue/issues/11006 , as this issue is originally about inconsistencies in the default setup not the helm chart values or documentation.
(I will try to create a PR to fix them tomorrow)

charts/kueue/values.yaml
strigazi: done

charts/kueue/values.yaml
strigazi: done

site/content/en/docs/tasks/manage/enforce_job_management/opt_in_namespace_management.md
strigazi: done

strigazi:
> From my point of view this still shouldn't close #11006 , as this issue is originally about inconsistencies in the default setup not the helm chart values or documentation. (I will try to create a PR to fix them tomorrow)

done, changed `Closes` to just linking to it.

mimowo:
/lgtm
/approve
/cherrypick website
Thank you 👍

PR #11156, opened 2026-05-13
Introduce Amount struct to deal with quotas overflow

### LINKED ISSUE #9843: Kueue does not allow to schedule workloads when CQ quota is specified as 1E
**What happened**:

When CQ quota is specified as 1E to simulate "infinity", then effectively it is handled as 0.

To see this is supported check the tests for resource.Quantity https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/api/resource/quantity_test.go#L317

We correctly support 1E since allowed by API. If this is really problematic to support, then we should have validation against it.

1. Create a CQ and specify nominalQuota as 1E.
2. Submit a workload with cpu request.
Issue: the workload fails with `couldn't assign flavors to pod set main: insufficient quota for cpu in flavor default-flavor, previously considered podsets requests (0) + current podset request (3) > maximum capacity (0)`

1P seems to work fine

mszadkow:
#### What type of PR is this?
/kind bug

Introduce a dedicated resources.Amount struct for quota arithmetic and migrate quota storage to Amount in scheduler quota paths (Nominal/BorrowingLimit/LendingLimit and SubtreeQuota).
AmountFromQuantity will treat out-of-range quota quantities (for example 1E CPU in milli units) as Unlimited, and all quota math and comparisons will be done through Amount methods so cohort aggregation cannot overflow.

Fixes #9843 

```release-note
NONE
```

### DIFF
--- a/pkg/cache/scheduler/cache.go
+++ b/pkg/cache/scheduler/cache.go
@@ -961,11 +961,11 @@ func getUsage(frq resources.FlavorResourceQuantities, cq *clusterQueue) []kueue.
 				used := frq[fr]
 				rUsage := kueue.ResourceUsage{
 					Name:  rName,
-					Total: resources.ResourceQuantity(rName, used),
+					Total: resources.ResourceQuantity(rName, used.Int64()),
 				}
 				// Enforce `borrowed=0` if the clusterQueue doesn't belong to a cohort.
 				if cq.HasParent() {
-					borrowed := used - rQuota.Nominal
+					borrowed := used.Sub(rQuota.Nominal).Int64()
 					if borrowed > 0 {
 						rUsage.Borrowed = resources.ResourceQuantity(rName, borrowed)
 					}
@@ -1026,7 +1026,7 @@ func filterLocalQueueUsage(orig resources.FlavorResourceQuantities, resourceGrou
 				fr := resources.FlavorResource{Flavor: fName, Resource: rName}
 				outFlvUsage.Resources = append(outFlvUsage.Resources, kueue.LocalQueueResourceUsage{
 					Name:  rName,
-					Total: resources.ResourceQuantity(rName, orig[fr]),
+					Total: resources.ResourceQuantity(rName, orig[fr].Int64()),
 				})
 			}
 			// The resourceUsages should be in a stable order to avoid endless creation of update events.
--- a/pkg/cache/scheduler/cache_test.go
+++ b/pkg/cache/scheduler/cache_test.go
@@ -1531,10 +1531,9 @@ func TestCacheWorkloadOperations(t *testing.T) {
 					UsedResources: cq.resourceNode.Usage,
 				}
 			}
-			cmpFilter := func(a, b resources.FlavorResourceQuantities) bool { return len(a) != 0 || len(b) != 0 }
 			cmpOpts := []cmp.Option{
 				cmpopts.EquateEmpty(),
-				cmp.FilterValues(cmpFilter, cmp.Comparer(equalFlavorResourceQuantitiesIgnoringZero)),
+				cmp.FilterValues(func(a, b resources.FlavorResourceQuantities) bool { return len(a) != 0 || len(b) != 0 }, cmp.Comparer(equalFlavorResourceQuantitiesIgnoringZero)),
 			}
 			if diff := cmp.Diff(step.wantResults, gotResult, cmpOpts...); diff != "" {
 				t.Errorf("Unexpected clusterQueues (-want,+got):\n%s", diff)
@@ -2622,12 +2621,11 @@ func TestCacheQueueOperations(t *testing.T) {
 					cacheQueues[qKey] = cacheQ
 				}
 			}
-			cmpFilter := func(a, b resources.FlavorResourceQuantities) bool { return len(a) != 0 || len(b) != 0 }
 			cmpOpts := []cmp.Option{
 				cmp.AllowUnexported(LocalQueue{}),
 				cmpopts.EquateEmpty(),
 				cmpopts.IgnoreTypes(sync.RWMutex{}),
-				cmp.FilterValues(cmpFilter, cmp.Comparer(equalFlavorResourceQuantitiesIgnoringZero)),
+				cmp.FilterValues(func(a, b resources.FlavorResourceQuantities) bool { return len(a) != 0 || len(b) != 0 }, cmp.Comparer(equalFlavorResourceQuantitiesIgnoringZero)),
 			}
 			if diff := cmp.Diff(tc.wantLocalQueues, cacheQueues, cmpOpts...); diff != "" {
 				t.Errorf("Unexpected localQueues (-want,+got):\n%s", diff)
@@ -3038,25 +3036,12 @@ func TestIsAddedCheckWorkload(t *testing.T) {
 	}
 }
 
-func normalizeFlavorResourceQuantities(in resources.FlavorResourceQuantities) resources.FlavorResourceQuantities {
-	out := make(resources.FlavorResourceQuantities, len(in))
-	for k, v := range in {
-		if v.CmpInt64(0) != 0 {
-			out[k] = v
-		}
-	}
-	return out
-}
-
 func equalFlavorResourceQuantitiesIgnoringZero(a, b resources.FlavorResourceQuantities) bool {
-	na := normalizeFlavorResourceQuantities(a)
-	nb := normalizeFlavorResourceQuantities(b)
-	if len(na) != len(nb) {
-		return false
+	if len(b) > len(a) {
+		a, b = b, a
 	}
-	for k, av := range na {
-		bv, ok := nb[k]
-		if !ok || !av.Equal(bv) {
+	for k, av := range a {
+		if av.Cmp(b[k]) != 0 {
 			return false
 		}
 	}
--- a/pkg/cache/scheduler/cache_test.go
+++ b/pkg/cache/scheduler/cache_test.go
@@ -451,11 +451,11 @@ func TestCacheClusterQueueOperations(t *testing.T) {
 					FlavorFungibility:             defaultFlavorFungibility,
 					resourceNode: resourceNode{
 						Usage: resources.FlavorResourceQuantities{
-							{Flavor: "default", Resource: corev1.ResourceCPU}: 5000,
+							{Flavor: "default", Resource: corev1.ResourceCPU}: resources.NewAmount(5000),
 						},
 					},
 					AdmittedUsage: resources.FlavorResourceQuantities{
-						{Flavor: "default", Resource: corev1.ResourceCPU}: 5000,
+						{Flavor: "default", Resource: corev1.ResourceCPU}: resources.NewAmount(5000),
 					},
 					Status:     active,
 					Preemption: defaultPreemption,
@@ -903,12 +903,12 @@ func TestCacheClusterQueueOperations(t *testing.T) {
 					AllocatableResourceGeneration: 1,
 					FlavorFungibility:             defaultFlavorFungibility,
 					AdmittedUsage: resources.FlavorResourceQuantities{
-						{Flavor: "f1", Resource: corev1.ResourceCPU}: 1000,
+						{Flavor: "f1", Resource: corev1.ResourceCPU}: resources.NewAmount(1000),
 					},
 					FairWeight: defaultWeight,
 					resourceNode: resourceNode{
 						Usage: resources.FlavorResourceQuantities{
-							{Flavor: "f1", Resource: corev1.ResourceCPU}: 2000,
+							{Flavor: "f1", Resource: corev1.ResourceCPU}: resources.NewAmount(2000),
 						},
 					},
 					Workloads: map[workload.Reference]*workload.Info{
@@ -1179,8 +1179,8 @@ func TestCacheWorkloadOperations(t *testing.T) {
 				"one": {
 					Workloads: sets.New[workload.Reference]("/a", "/b"),
 					UsedResources: resources.FlavorResourceQuantities{
-						{Flavor: "on-demand", Resource: corev1.ResourceCPU}: 10,
-						{Flavor: "spot", Resource: corev1.ResourceCPU}:      15,
+						{Flavor: "on-demand", Resource: corev1.ResourceCPU}: resources.NewAmount(10),
+						{Flavor: "spot", Resource: corev1.ResourceCPU}:      resources.NewAmount(15),
 					},
 				},
 				"two": {
@@ -1204,8 +1204,8 @@ func TestCacheWorkloadOperations(t *testing.T) {
 				"one": {
 					Workloads: sets.New[workload.Reference]("/a", "/b"),
 					UsedResources: resources.FlavorResourceQuantities{
-						{Flavor: "on-demand", Resource: corev1.ResourceCPU}: 10,
-						{Flavor: "spot", Resource: corev1.ResourceCPU}:      15,
+						{Flavor: "on-demand", Resource: corev1.ResourceCPU}: resources.NewAmount(10),
+						{Flavor: "spot", Resource: corev1.ResourceCPU}:      resources.NewAmount(15),
 					},
 				},
 				"two": {
@@ -1228,8 +1228,8 @@ func TestCacheWorkloadOperations(t *testing.T) {
 				"one": {
 					Workloads: sets.New[workload.Reference]("/a", "/b"),
 					UsedResources: resources.FlavorResourceQuantities{
-						{Flavor: "on-demand", Resource: corev1.ResourceCPU}: 10,
-						{Flavor: "spot", Resource: corev1.ResourceCPU}:      15,
+						{Flavor: "on-demand", Resource: corev1.ResourceCPU}: resources.NewAmount(10),
+						{Flavor: "spot", Resource: corev1.ResourceCPU}:      resources.NewAmount(15),
 					},
 				},
 				"two": {
@@ -1252,8 +1252,8 @@ func TestCacheWorkloadOperations(t *testing.T) {
 				"one": {
 					Workloads: sets.New[workload.Reference]("/a"),
 					UsedResources: resources.FlavorResourceQuantities{
-						{Flavor: "on-demand", Resource: corev1.ResourceCPU}: 10,
-						{Flavor: "spot", Resource: corev1.ResourceCPU}:      15,
+						{Flavor: "on-demand", Resource: corev1.ResourceCPU}: resources.NewAmount(10),
+						{Flavor: "spot", Resource: corev1.ResourceCPU}:      resources.NewAmount(15),
 					},
 				},
 				"two": {
@@ -1274,8 +1274,8 @@ func TestCacheWorkloadOperations(t *testing.T) {
 				"one": {
 					Workloads: sets.New[workload.Reference]("/a"),
 					UsedResources: resources.FlavorResourceQuantities{
-						{Flavor: "on-demand", Resource: corev1.ResourceCPU}: 10,
-						{Flavor: "spot", Resource: corev1.ResourceCPU}:      15,
+						{Flavor: "on-demand", Resource: corev1.ResourceCPU}: resources.NewAmount(10),
+						{Flavor: "spot", Resource: corev1.ResourceCPU}:      resources.NewAmount(15),
 					},
 				},
 				"two": {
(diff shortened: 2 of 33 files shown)


mszadkow:
In addition to the changes present in the PR we might want to follow up with `PodSetResources.Requests`, which currently use int64 still.
wdyt @pajakd ?

pkg/resources/amount.go
```diff
@@ -0,0 +1,218 @@
...
+//   - cohort aggregation can't wrap around int64 - Add of an Unlimited yields
+//     Unlimited, and bounded arithmetic saturates rather than overflowing;
+//   - the type system funnels quota arithmetic through methods, so a future
+//     contributor can't accidentally fall back to raw "+" / "-" on quota
+//     values and silently re-introduce the overflow.
+//
+// Keep usage values (which are bounded by what real workloads consume) as
+// int64. Mixed Amount-vs-int64 arithmetic uses the *Int64 methods.
+type Amount struct {
+	unlimited bool
```
mimowo: Actually, I'm thinking if we need it, probably we could represent that just by MaxInt64, similarly as the new functions in math package are doing, wdyt?

pkg/resources/amount.go
```diff
@@ -0,0 +1,218 @@
...
+// saturatingAddInt64 returns a + b, clamping at math.MaxInt64 / math.MinInt64
+// on overflow. Internal-only; quota call sites should use Amount methods.
+func saturatingAddInt64(a, b int64) int64 {
+	if b > 0 && a > math.MaxInt64-b {
+		return math.MaxInt64
+	}
+	if b < 0 && a < math.MinInt64-b {
+		return math.MinInt64
+	}
+	return a + b
```
mimowo: delegate to the pre-existing functions

pkg/cache/scheduler/resource_node.go
```diff
@@ -209,10 +238,17 @@ func updateCohortTreeResourcesIfNoCycle(cohort *cohort) {
 
 func accumulateFromChild(parent *cohort, child flatResourceNode) {
 	for fr, childQuota := range child.getResourceNode().SubtreeQuota {
-		parent.resourceNode.SubtreeQuota[fr] += childQuota - child.getResourceNode().localQuota(fr)
+		// Amount.Add propagates Unlimited and saturates bounded overflow,
+		// so summing across many large-quota children can't wrap to a small
+		// or negative SubtreeQuota at the parent.
```
mimowo: Idon't understand the comment, could you maybe rephrase to make it more clear.
In particular it mentions "Add" but the line below is using `SubInt64` so I'm confused.

pkg/cache/scheduler/clusterqueue.go
```diff
@@ -223,9 +223,12 @@ func (c *clusterQueue) updateQuotasAndResourceGroups(in []kueue.ResourceGroup) b
 	c.resourceNode.Quotas = createResourceQuotas(in)
 
 	// Start at 1, for backwards compatibility.
+	// Use maps.EqualFunc with ResourceQuota.Equal for the Quotas map: it holds
+	// resources.Amount values whose unexported fields cause k8s
+	// equality.Semantic.DeepEqual (a forked reflect-based DeepEqual) to panic.
 	return c.AllocatableResourceGeneration == 0 ||
 		!equality.Semantic.DeepEqual(oldRG, c.ResourceGroups) ||
-		!equality.Semantic.DeepEqual(oldQuotas, c.resourceNode.Quotas)
+		!maps.EqualFunc(oldQuotas, c.resourceNode.Quotas, ResourceQuota.Equal)
```
mimowo, line 232: What are the cases when simple `!equality.Semantic.DeepEqual(oldQuotas, c.resourceNode.Quotas)` is giving wrong results? Do we have them covered with tests?

mimowo: Ahh, I see, this is exaplained above, because Amount has unexported fields. No need to change, lgtm.

pkg/cache/scheduler/clusterqueue.go
```diff
@@ -612,7 +615,14 @@ func (c *clusterQueue) updateWorkloadTASUsage(log logr.Logger, wi *workload.Info
...
 	for fr, q := range newUsage {
-		oldUsage[fr] += q * int64(op.asSignedOne())
+		if op == add {
+			oldUsage[fr] = oldUsage[fr].Add(q)
+		} else {
+			oldUsage[fr] = oldUsage[fr].Sub(q)
+		}
+		if oldUsage[fr].CmpInt64(0) == 0 {
+			delete(oldUsage, fr)
+		}
```
mimowo: Is this a bugfix for pre-existing bug? Because we didn't seem to clean it before?

pkg/cache/scheduler/clusterqueue.go
```diff
@@ -612,7 +615,14 @@ func (c *clusterQueue) updateWorkloadTASUsage(log logr.Logger, wi *workload.Info
 
 func updateFlavorUsage(newUsage resources.FlavorResourceQuantities, oldUsage resources.FlavorResourceQuantities, op usageOp) {
 	for fr, q := range newUsage {
-		oldUsage[fr] += q * int64(op.asSignedOne())
+		if op == add {
+			oldUsage[fr] = oldUsage[fr].Add(q)
+		} else {
+			oldUsage[fr] = oldUsage[fr].Sub(q)
+		}
```
mimowo: Could we inline as before using saturatingMul from https://github.com/kubernetes-sigs/kueue/pull/11182?

test/integration/singlecluster/scheduler/scheduler_test.go
```diff
@@ -1412,6 +1417,30 @@ var _ = ginkgo.Describe("Scheduler", func() {
 			util.ExpectAdmittedWorkloadsTotalMetric(prodCQ, "", 1)
 		})
 
+		ginkgo.It("Should not overflow when sibling ClusterQueues in a cohort have very large nominalQuota", func() {
```
mimowo: Is this test currently failing on main if not for the changes?

pkg/cache/scheduler/clusterqueue.go
mszadkow: not a bugfix, but a memory improvement - suggested by AI
Can be dropped

pkg/resources/amount.go
mszadkow: Yes, we totally can

pkg/cache/scheduler/cache.go
```diff
@@ -898,11 +898,11 @@ func getUsage(frq resources.FlavorResourceQuantities, cq *clusterQueue) []kueue.
 				used := frq[fr]
 				rUsage := kueue.ResourceUsage{
 					Name:  rName,
-					Total: resources.ResourceQuantity(rName, used),
+					Total: resources.ResourceQuantity(rName, used.Int64()),
```
mimowo, line 964: Wouldn't it make the diff smaller if we change the type accepted by resources.ResourceQuantity to Amount, to use it also more consistently to avoid the transitions with Iny64() as much as possible

pkg/cache/scheduler/clusterqueue.go
mimowo: Hmmm, I asked because this looks like a bugfix for memory leak, are you sure there is no memory leak?  
Maybe this is not a permanent memory leak because the maps are short leaving anyway?

IN any case this is separete PR for sure, either bugifix or just cleanup

pkg/resources/amount.go
```diff
@@ -0,0 +1,196 @@
...
+		return Amount{value: q.MilliValue()}
+	}
+	if q.Cmp(maxNonCPUQuantityForAmount) >= 0 {
+		return Unlimited
+	}
+	return Amount{value: q.Value()}
+}
+
+// IsUnlimited reports whether a is the Unlimited sentinel.
+func (a Amount) IsUnlimited() bool { return a.value == math.MaxInt64 }
```
mimowo: why is that exported?

pkg/cache/scheduler/resource_node.go
```diff
@@ -104,56 +108,81 @@ func LocalAvailable(node flatResourceNode, fr resources.FlavorResource) int64 {
...
+func addUsage(node hierarchicalResourceNode, fr resources.FlavorResource, val resources.Amount) {
 	r := node.getResourceNode()
 	localAvailable := LocalAvailable(node, fr)
-	r.Usage[fr] += val
-	if node.HasParent() && val > localAvailable {
-		deltaParentUsage := val - localAvailable
+	r.Usage[fr] = r.Usage[fr].Add(val)
+	if r.Usage[fr].CmpInt64(0) == 0 {
+		delete(r.Usage, fr)
+	}
```
mimowo: Similarly, is this a fx for memory leak or premature optimization?

pkg/cache/scheduler/resource_node.go
```diff
@@ -104,56 +108,81 @@ func LocalAvailable(node flatResourceNode, fr resources.FlavorResource) int64 {
...
-func removeUsage(node hierarchicalResourceNode, fr resources.FlavorResource, val int64) {
+func removeUsage(node hierarchicalResourceNode, fr resources.FlavorResource, val resources.Amount) {
 	r := node.getResourceNode()
-	usageStoredInParent := r.Usage[fr] - r.localQuota(fr)
-	r.Usage[fr] -= val
+	usageStoredInParent := r.Usage[fr].Int64() - r.localQuota(fr)
+	r.Usage[fr] = r.Usage[fr].Sub(val)
+	if r.Usage[fr].CmpInt64(0) == 0 {
+		delete(r.Usage, fr)
+	}
```
mimowo: Similarly, is this a fx for memory leak or premature optimization?

pkg/cache/scheduler/resource_node.go
```diff
@@ -209,10 +238,15 @@ func updateCohortTreeResourcesIfNoCycle(cohort *cohort) {
...
+		delta := childQuota.SubInt64(child.getResourceNode().localQuota(fr))
+		// Add saturates at MaxInt64 on overflow, so large children never wrap the parent SubtreeQuota negative.
+		parent.resourceNode.SubtreeQuota[fr] = parent.resourceNode.SubtreeQuota[fr].Add(delta)
 	}
 	for fr, childUsage := range child.getResourceNode().Usage {
-		parent.resourceNode.Usage[fr] += max(0, childUsage-child.getResourceNode().localQuota(fr))
+		parent.resourceNode.Usage[fr] = parent.resourceNode.Usage[fr].AddInt64(max(0, childUsage.Int64()-child.getResourceNode().localQuota(fr)))
+		if parent.resourceNode.Usage[fr].CmpInt64(0) == 0 {
+			delete(parent.resourceNode.Usage, fr)
+		}
```
mimowo: Similarly, is this a fx for memory leak or premature optimization?

pkg/cache/scheduler/cache.go
mszadkow: Not necessarily, because it would benefit in some places but it will get worse in other places e.g:
`map[ResourceName]int64` - pkg/cache/scheduler/resource_node.go
`resourceFloat` - pkg/cache/scheduler/cache.go
`pendingResource()` - pkg/cache/queue/metrics.go

It can be done ofc, we could probably identify more places where we could switch to Amount, but I think the PR will grow

pkg/cache/scheduler/resource_node.go
```diff
@@ -104,56 +108,81 @@ func LocalAvailable(node flatResourceNode, fr resources.FlavorResource) int64 {
...
+func saturatingAdd(a, b int64) int64 {
+	const maxInt64 = int64(^uint64(0) >> 1)
+	const minInt64 = -maxInt64 - 1
+	if b > 0 && a > maxInt64-b {
+		return maxInt64
+	}
+	if b < 0 && a < minInt64-b {
+		return minInt64
+	}
+	return a + b
```
mimowo: Why is this code duplcated since we already have helpers for that?

pkg/scheduler/flavorassigner/flavorassigner.go
mimowo: I think part of the task is to generalize the previous fixes, and so use the Amount here too. I'm wondering if we could pass amount into the function to avoid: Amoun -> int64 -> Amount

mszadkow: For this to work and make sense we will have to go "all in" for use of Amount in resource calculations.
Otherwise we gain simplification here with cost of pilling up code changes around to make it work

mimowo: Sure, but:
- the end result is "all in" it is just a matter of how we split the work to make the code reviewable
- it is very hard to me to know how big the PR will be without actually trying
- some places make actually be gone from the diff - when we call currently Int64() before passing to function. 

So I would rather propose to make an experimental PR where we go "all in" to see the complete set of changes, and then we can see how to divide it. 

Ofc that could be another experimental PR to see how much it blows up if the places are addressed.

mimowo: Or maybe I'm not fully sure what "all in" means, but the places which were adjusted with the SaturatingAdd and SaturatingMul I think should already be attempted to be generalized with the "Amount"

mszadkow: ok, sounds reasonable to try

test/integration/singlecluster/scheduler/scheduler_test.go
mszadkow: Adjusted and tested, yes, now it does

pkg/cache/scheduler/resource.go
```diff
@@ -43,9 +43,31 @@ func (rg *ResourceGroup) Clone() ResourceGroup {
...
+		return false
+	}
+	return equalAmountPtr(q.LendingLimit, other.LendingLimit)
+}
+
+func equalAmountPtr(a, b *resources.Amount) bool {
+	if a == nil || b == nil {
+		return a == b
+	}
+	return a.Equal(*b)
```
reruno: maybe use ptr.Equal() instead from "k8s.io/utils/ptr"

pkg/resources/amount_test.go
```diff
@@ -0,0 +1,203 @@
...
+			}
+		})
+	}
+}
+
+// TestAmountArithmetic pins the unlimited-propagation and saturation rules
+// that prevent cohort math from overflowing, matching mimowo's request that
+// quota arithmetic be funneled through an abstraction rather than relying on
+// raw int64.
+func TestAmountArithmetic(t *testing.T) {
```
reruno, line 111: // TestAmountArithmetic pins the unlimited-propagation and saturation rules
  // that prevent cohort math from overflowing. Quota arithmetic must funnel
  // through Amount rather than raw int64 so these invariants hold globally.

rephrase :)

pkg/resources/amount_test.go
```diff
@@ -0,0 +1,203 @@
...
+	testCases := []struct {
```
mbobrovskyi:
```suggestion
	testCases := map[string]struct {
```
Can we use this pattern?

pkg/resources/requests.go
```diff
@@ -149,6 +149,16 @@ func ResourceQuantityString(name corev1.ResourceName, v int64) string {
...
 
+// AmountQuantityString formats an Amount as a Kubernetes resource quantity
+// string. Unlimited amounts are formatted as "<unlimited>" rather than as
+// the raw math.MaxInt64 sentinel.
+func AmountQuantityString(name corev1.ResourceName, a Amount) string {
+	if a.Equal(Unlimited) {
+		return Unlimited.String()
+	}
+	return ResourceQuantityString(name, a.Int64())
+}
```
reruno, line 160: Should we account for negative unlimited?

mszadkow: not really, we can't have negative quotas or usage

pkg/resources/amount_test.go
mszadkow: done

mszadkow:
@reruno @mbobrovskyi ptal

pkg/cache/scheduler/cache_test.go
```diff
@@ -3035,6 +3038,31 @@ func TestIsAddedCheckWorkload(t *testing.T) {
...
+		return false
+	}
+	for k, av := range na {
+		bv, ok := nb[k]
+		if !ok || !av.Equal(bv) {
+			return false
+		}
+	}
+	return true
+}
```
mbobrovskyi, line 3049:
```suggestion
func equalFlavorResourceQuantitiesIgnoringZero(a, b resources.FlavorResourceQuantities) bool {
	if len(b) > len(a) {
		a, b = b, a
	}
	for k, av := range a {
		if av.Cmp(b[k]) != 0 {
			return false
		}
	}
	return true
}
```

mbobrovskyi, line 3049:
```suggestion
```

pkg/cache/scheduler/cache_test.go
```diff
@@ -2626,7 +2622,14 @@ func TestCacheQueueOperations(t *testing.T) {
 					cacheQueues[qKey] = cacheQ
 				}
 			}
-			if diff := cmp.Diff(tc.wantLocalQueues, cacheQueues, cmp.AllowUnexported(LocalQueue{}), cmpopts.EquateEmpty(), cmpopts.IgnoreTypes(sync.RWMutex{})); diff != "" {
+			cmpFilter := func(a, b resources.FlavorResourceQuantities) bool { return len(a) != 0 || len(b) != 0 }
```
mbobrovskyi: Can we rename this variable to something more specific, or just propagate it implicitly?

pkg/cache/scheduler/cache_test.go
```diff
@@ -1540,7 +1531,12 @@ func TestCacheWorkloadOperations(t *testing.T) {
 					UsedResources: cq.resourceNode.Usage,
 				}
 			}
-			if diff := cmp.Diff(step.wantResults, gotResult, cmpopts.EquateEmpty()); diff != "" {
+			cmpFilter := func(a, b resources.FlavorResourceQuantities) bool { return len(a) != 0 || len(b) != 0 }
```
mbobrovskyi: Also here.

vladikkuzn:
Please fix linter
Otherwise /lgtm

test/integration/singlecluster/scheduler/quotacheckstrategy/quota_check_strategy_test.go
```diff
@@ -46,7 +47,6 @@ var _ = ginkgo.Describe("Quota check strategy", ginkgo.Ordered, ginkgo.ContinueO
 		)
 
 		ginkgo.BeforeAll(func() {
-			features.SetFeatureGateDuringTest(ginkgo.GinkgoTB(), features.QuotaCheckStrategy, true)
```
mbobrovskyi, line 49: Good catch!

test/integration/singlecluster/scheduler/quotacheckstrategy/quota_check_strategy_test.go
```diff
@@ -133,6 +134,30 @@ var _ = ginkgo.Describe("Quota check strategy", ginkgo.Ordered, ginkgo.ContinueO
...
+			createdWl := &kueue.Workload{}
+			wlKey := client.ObjectKeyFromObject(wl)
+			gomega.Eventually(func(g gomega.Gomega) {
+				g.Expect(k8sClient.Get(ctx, wlKey, createdWl)).To(gomega.Succeed())
+				g.Expect(workload.IsAdmitted(createdWl)).To(gomega.BeTrue(), "expected workload to be admitted with very large nominal quota")
+				cond := apimeta.FindStatusCondition(createdWl.Status.Conditions, kueue.WorkloadQuotaReserved)
+				g.Expect(cond).NotTo(gomega.BeNil())
+				g.Expect(cond.Message).NotTo(gomega.ContainSubstring("maximum capacity (0)"))
+			}, util.Timeout, util.Interval).Should(gomega.Succeed())
+		})
```
mbobrovskyi:
```suggestion
			wlKey := client.ObjectKeyFromObject(wl)
			util.ExpectWorkloadsToBeAdmittedByKeys(wlKey)
```
Is it enough?

mbobrovskyi (commented):
Overall LGTM. Just few nits.

test/integration/singlecluster/scheduler/quotacheckstrategy/quota_check_strategy_test.go
mszadkow: I think so, this was a result of toggling the original test that showed the issue, but overall yes admitted is even better

mbobrovskyi (commented):
Awesome! Thank you!

/lgtm

mimowo (commented):
/lgtm
/approve
Thank you, I think by using the Amount consistently we can prevent the overflow bugs in the future 👍

PR #11181, opened 2026-05-14
Fix finished Workload GC with finalizers

### LINKED ISSUE #11130: Finished workload slices aren't garbage-collected
**What happened**:

I configured my Kueue installation's `.objectRetentionPolicies` to garbage-collect finished workloads after 1 minute. Then I created an elastic Ray cluster. The Ray cluster scaled up a few times which created new workload slices. Old slices were marked as `Finished` and the controller logs stated they were deleted, but they never terminated.

I expected the old Workload objects (which are slices for an elastic Ray cluster) to terminate.

1. Install Kueue and configure `objectRetentionPolicies`:
    ```
    objectRetentionPolicies:
      workloads:
        afterFinished: "1m"
    ```
2. Create an elastic Ray cluster with `inTreeAutoscaling` enabled
3. Trigger scale up such that there is at least one `Finished` `Workload` (slice) that has been replaced by a new `Workload`.
4. Observe Kueue controller logs reporting deletion, but `Finished` workload slice is not deleted.

**Other notes**:

This relates to the change made in https://github.com/kubernetes-sigs/kueue/pull/2686.

**Environment**:
- Kubernetes version (use `kubectl version`): v1.34.6-gke.1154000
- Kueue version (use `git describe --tags --dirty --always`): v0.16.2
- Cloud provider or hardware configuration: Google Cloud Platform (GCP)

ShaanveerS:
#### What type of PR is this?
/kind bug

When finished Workload retention expires, if the Workload still has the `resource-in-use` finalizer attached, Kubernetes marks it for deletion but won't actually remove it.

This was already called out as a caveat in [KEP-1618](https://github.com/kubernetes-sigs/kueue/blob/main/keps/1618-optional-gc-of-workloads/README.md#L108-L115):

> The deletion mechanism assumes that finished Workload objects do not have any finalizers attached to them. After an investigation, the [exact place in the code](https://github.com/kubernetes-sigs/kueue/blob/a128ae3354a56670ffa58695fef1eca270fe3a47/pkg/controller/jobframework/reconciler.go#L332-L340) was found that removes the resource-in-use finalizer when the Workload state transitions to finished. If that behavior ever changes, or if there is an external source of a finalizer being attached to the Workload, it will be marked for deletion by the Kubernetes client but will not actually be deleted until the finalizer is removed.

WorkloadSlices were added after the KEP. 
Replaced WorkloadSlices are marked finished by the scheduler in [`replaceWorkloadSlice`](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/scheduler/scheduler.go#L1047-L1057), and that path only calls [`workload.Finish`](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/workload/workload.go#L1709-L1720), which sets the finished condition but does not remove the finalizer.

The fix is to remove it in the finished Workload GC path, so the Workload can actually be removed once retention has elapsed.
Fixes #11130

```release-note
Fix a bug where finished Workloads could remain stuck after object retention deletion if they still had Kueue's resource-in-use finalizer.
```

### DIFF
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -265,10 +265,20 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		}
 
 		log.V(2).Info("Deleting workload because it has finished and the retention period has elapsed", "retention", *r.workloadRetention.afterFinished)
-		if err := r.client.Delete(ctx, &wl); err != nil {
+
+		// Finished Workloads should no longer need Kueue's resource-in-use finalizer.
+		// However, WorkloadSlices and other paths can mark a Workload as Finished without
+		// going through the job finalization path that normally removes it. Remove only
+		// Kueue's finalizer before deleting; otherwise Kubernetes may only set
+		// deletionTimestamp and the Workload can remain stuck until the finalizer is removed.
+		deleteRequested, err := workload.Delete(ctx, r.client, &wl)
+		if err != nil {
 			return ctrl.Result{}, client.IgnoreNotFound(err)
 		}
-		r.recorder.Eventf(&wl, nil, corev1.EventTypeNormal, "Deleted", "Deleted", "Deleted finished workload due to elapsed retention")
+		if deleteRequested {
+			r.recorder.Eventf(&wl, nil, corev1.EventTypeNormal, "Deleted", "Deleted", "Deleted finished workload due to elapsed retention")
+		}
+
 		return ctrl.Result{}, nil
 	}
 
--- a/pkg/controller/core/workload_controller_test.go
+++ b/pkg/controller/core/workload_controller_test.go
@@ -2724,6 +2724,40 @@ func TestReconcile(t *testing.T) {
 			wantWorkload: nil,
 			wantError:    nil,
 		},
+		"should remove finalizer from owned finished workload already being deleted, object retention configured": {
+			workload: utiltestingapi.MakeWorkload("wl", "ns").
+				Condition(metav1.Condition{
+					Type:   kueue.WorkloadFinished,
+					Status: metav1.ConditionTrue,
+					LastTransitionTime: metav1.Time{
+						Time: now.Add(-2 * util.MediumTimeout),
+					},
+				}).
+				ControllerReference(batchv1.SchemeGroupVersion.WithKind("Job"), "ownername", "owneruid").
+				DeletionTimestamp(now).
+				Finalizers(kueue.ResourceInUseFinalizerName, "example.com/finalizer").
+				Obj(),
+			reconcilerOpts: []Option{
+				WithWorkloadRetention(
+					&workloadRetentionConfig{
+						afterFinished: ptr.To(util.MediumTimeout),
+					},
+				),
+			},
+			wantWorkload: utiltestingapi.MakeWorkload("wl", "ns").
+				Condition(metav1.Condition{
+					Type:   kueue.WorkloadFinished,
+					Status: metav1.ConditionTrue,
+					LastTransitionTime: metav1.Time{
+						Time: now.Add(-2 * util.MediumTimeout),
+					},
+				}).
+				ControllerReference(batchv1.SchemeGroupVersion.WithKind("Job"), "ownername", "owneruid").
+				DeletionTimestamp(now).
+				Finalizers("example.com/finalizer").
+				Obj(),
+			wantError: nil,
+		},
 		"shouldn't try to delete the workload because the retention period hasn't elapsed yet, object retention configured": {
 			workload: utiltestingapi.MakeWorkload("wl", "ns").
 				Condition(metav1.Condition{
--- a/pkg/workload/workload.go
+++ b/pkg/workload/workload.go
@@ -1550,6 +1550,25 @@ func RemoveFinalizer(ctx context.Context, c client.Client, wl *kueue.Workload) e
 	return nil
 }
 
+// Delete removes Kueue's resource-in-use finalizer before deleting the Workload.
+// If deletion is already in progress, finalizer removal is enough for Kubernetes
+// to continue object cleanup, so Delete returns false without issuing another
+// delete request. The returned bool reports whether this call successfully
+// requested deletion.
+func Delete(ctx context.Context, c client.Client, wl *kueue.Workload) (bool, error) {
+	if err := RemoveFinalizer(ctx, c, wl); err != nil {
+		return false, err
+	}
+	if !wl.DeletionTimestamp.IsZero() {
+		log.FromContext(ctx).V(3).Info("Skipping workload delete because deletion is already in progress", "workload", klog.KObj(wl))
+		return false, nil
+	}
+	if err := c.Delete(ctx, wl); err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
 type flavorSet = sets.Set[kueue.ResourceFlavorReference]
 
 // AdmissionChecksForWorkload returns AdmissionChecks that should be assigned to a specific Workload based on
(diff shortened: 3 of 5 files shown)


pkg/controller/core/workload_controller.go
```diff
@@ -264,11 +264,18 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 			return ctrl.Result{RequeueAfter: remainingTime}, nil
 		}
 
+		if err := workload.RemoveFinalizer(ctx, r.client, &wl); err != nil {
```
mwielgus: This is getting more complex and tricky. Please add comments here and above explaining what is being done and why.

pkg/controller/core/workload_controller.go
```diff
@@ -264,11 +264,18 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+			return ctrl.Result{}, client.IgnoreNotFound(err)
+		}
+		if !wl.DeletionTimestamp.IsZero() {
+			return ctrl.Result{}, nil
+		}
+
 		log.V(2).Info("Deleting workload because it has finished and the retention period has elapsed", "retention", *r.workloadRetention.afterFinished)
 		if err := r.client.Delete(ctx, &wl); err != nil {
 			return ctrl.Result{}, client.IgnoreNotFound(err)
 		}
```
mimowo: We already have this pattern in a couple of places where we remove finalizer, then delete a workload, check if there is potential to comminize to avoid bugs in the future

pkg/controller/core/workload_controller.go
```diff
@@ -264,11 +264,18 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 			return ctrl.Result{RequeueAfter: remainingTime}, nil
 		}
 
+		if err := workload.RemoveFinalizer(ctx, r.client, &wl); err != nil {
+			return ctrl.Result{}, client.IgnoreNotFound(err)
+		}
+		if !wl.DeletionTimestamp.IsZero() {
```
mimowo: IIUC this is just an optimization to skip setting the deletion timestamp if already done? I'm ok-ish with that but we should at least log at v(3) that we skip the deletion becuase already in progress.

ShaanveerS: Done!

pkg/controller/core/workload_controller.go
ShaanveerS: Done!

pkg/controller/core/workload_controller.go
ShaanveerS: added workload.Delete helper. Thanks!

pkg/workload/workload.go
```diff
@@ -1550,6 +1550,24 @@ func RemoveFinalizer(ctx context.Context, c client.Client, wl *kueue.Workload) e
 	return nil
 }
 
+// Delete removes Kueue's resource-in-use finalizer before deleting the Workload.
+// If deletion is already in progress, finalizer removal is enough for Kubernetes
+// to continue object cleanup, so Delete returns false without issuing another
+// delete request. The returned bool reports whether this call successfully
+// requested deletion.
+func Delete(ctx context.Context, c client.Client, wl *kueue.Workload) (bool, error) {
+	if err := RemoveFinalizer(ctx, c, wl); err != nil {
```
mimowo, line 1559: Handle the NotFound error gracefully - this means already deleted, so return (true, nil)

mimowo: Ahh, this is already handled externally, let me re-check, that could be ok too

pkg/controller/core/workload_controller.go
```diff
@@ -264,10 +264,21 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+		// However, WorkloadSlices and other paths can mark a Workload as Finished without
+		// going through the job finalization path that normally removes it. Remove only
+		// Kueue's finalizer before deleting; otherwise Kubernetes may only set
+		// deletionTimestamp and the Workload can remain stuck until the finalizer is removed.
+		deleteRequested, err := workload.Delete(ctx, r.client, &wl)
+		if err != nil {
 			return ctrl.Result{}, client.IgnoreNotFound(err)
 		}
+		if !deleteRequested {
+			log.V(3).Info("Skipping workload delete because deletion is already in progress")
```
mimowo: I would rather push this log down inside "delete" so we don't need to handle here

ShaanveerS: Agreed, done!

pkg/workload/workload.go
ShaanveerS: Yeah it was because each caller already handles NotFound in its own way: 
workload GC ignores it, jobframework ignores it but only records an event when delete succeeds, and LWS just returns the error through its existing path.

pkg/controller/core/workload_controller.go
```diff
@@ -264,10 +264,20 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+		// deletionTimestamp and the Workload can remain stuck until the finalizer is removed.
+		deleteRequested, err := workload.Delete(ctx, r.client, &wl)
+		if err != nil {
 			return ctrl.Result{}, client.IgnoreNotFound(err)
 		}
+		if !deleteRequested {
+			return ctrl.Result{}, nil
+		}
+
+		log.V(2).Info("Deleting workload because it has finished and the retention period has elapsed", "retention", *r.workloadRetention.afterFinished)
```
mimowo: Can we move the comment above - where it was as it is saying "deleting" . We will also have an event after the operation anyway.

ShaanveerS: done

ShaanveerS:
the failed test seems another recurrence of https://github.com/kubernetes-sigs/kueue/issues/10298

pkg/controller/core/workload_controller_test.go
```diff
@@ -2724,6 +2724,40 @@ func TestReconcile(t *testing.T) {
 			wantWorkload: nil,
 			wantError:    nil,
 		},
+		"should remove finalizer from owned finished workload already being deleted, object retention configured": {
```
mimowo, line 2727: One more question: why hasn't the bug surfaced earlier in the pre-existing tests that we have in test/e2e/sequential/baseline/objectretentionpolicies_test.go?

ShaanveerS: Because those tests seems to use the path where the job reconciler already removes Kueue's finalizer when the Workload finishes whereas this bug needs a path like replaced WorkloadSlices, where we don't.

pkg/controller/core/workload_controller.go
```diff
@@ -265,9 +265,20 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+		// deletionTimestamp and the Workload can remain stuck until the finalizer is removed.
+		deleteRequested, err := workload.Delete(ctx, r.client, &wl)
+		if err != nil {
 			return ctrl.Result{}, client.IgnoreNotFound(err)
 		}
+		if !deleteRequested {
+			return ctrl.Result{}, nil
+		}
+
 		r.recorder.Eventf(&wl, nil, corev1.EventTypeNormal, "Deleted", "Deleted", "Deleted finished workload due to elapsed retention")
```
mimowo: nit, but  let's move the record under `if deleteRequested`, becasuse the `return` statements don't differ

mimowo: Analogously as it is done here: https://github.com/kubernetes-sigs/kueue/pull/11181/changes#diff-552ab657667de70900b502e2617bcb405fbf4203ff08b2a35d7c87f2d637a75bR976-R980

pkg/controller/core/workload_controller_test.go
mimowo: sgtm

pkg/controller/core/workload_controller.go
ShaanveerS: done

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16
Thank you for fixing the issue, and also commonizing the code to make the code more resilient in the future.

mimowo:
@ShaanveerS please prepare CPs manually

PR #11182, opened 2026-05-14
Prevent integer overflow in TotalRequests

pajakd:
#### What type of PR is this?
/kind bug

Automatic vulnerability scanner flagged that this multiplication could result in int64 overflow.

I confirmed in unit test, that when setting very high requests and a very large number of pods, a total requests would overflow to negative.

Fixes #

```release-note
Fixed vulnerability where workload with a large number of pods and with CPU requests close to max int64 would lead to integer overflow when calculating total requests and break quota limits
```

### DIFF
--- a/pkg/util/math/math.go
+++ b/pkg/util/math/math.go
@@ -56,6 +56,11 @@ func SaturatingMul(a, b int64) int64 {
 	if a == 0 || b == 0 {
 		return 0
 	}
+	// Explicitly catch the MinInt64 * -1 edge case.
+	// In two's complement, the absolute value of MinInt64 is 1 greater than MaxInt64.
+	// If we multiply it by -1, it overflows to MinInt64. Go safely handles MinInt64 / -1
+	// by evaluating it back to MinInt64. Because of this, our division check
+	// (res/b != a) evaluates to false and fails to detect the overflow.
 	if (a == -1 && b == stdmath.MinInt64) || (b == -1 && a == stdmath.MinInt64) {
 		return stdmath.MaxInt64
 	}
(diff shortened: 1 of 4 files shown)


pkg/util/math/math.go
```diff
@@ -49,3 +49,22 @@ func SafeMilliValue(q resource.Quantity) int64 {
...
+	}
+	res := a * b
+	if res/b != a {
+		if (a < 0) == (b < 0) {
+			return stdmath.MaxInt64
+		}
+		return stdmath.MinInt64
+	}
+	return res
+}
```
mimowo, line 75: This function is tricky, and the problem must be well known, can you maybe link some reference which inspired the implementation?

Because the impl is far from obvious, especially the special cases about "== -1"

pajakd: 1. in the case of `== -1` the int64 has values from ` - 2^63 = -9,223,372,036,854,775,808` to `9,223,372,036,854,775,807 = (2^63 - 1)`, so the stdmath.MinInt * (-1) overflows to stdmath.MinInt and we fail to observe the overflow. not sure if needed at all but added for completeness. Added a comment with the explanation.
2. For testing overflow with res/b != a its inspired by https://groups.google.com/g/golang-nuts/c/h5oSN5t3Au4/m/KaNQREhZh0QJ

mimowo: Cool, let's add unit test cases for the function to cover it 100%

mimowo (commented):
Thank you 👍 
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16

PR #11185, opened 2026-05-14
Fix TAS node taint checks to use admitted tolerations

### LINKED ISSUE #11152: TAS NodeHotSwap ignores ResourceFlavor tolerations when classifying node taints
**What happened**:

We noticed that a workload sometimes gets evicted with a `NodeFailure` error, even when the assigned node is healthy.
In our TAS-enabled cluster, the nodes have a GPU taint. The assigned `ResourceFlavor` has the `spec.tolerations` for this taint. This eviction issue happens if the Pod template does not originally include this toleration.

In the node controller, the node is classified by `classifyTaints(ctx, node.Spec.Taints, workload.PodSetsOnNode(wl, nodeName))` func.
However, `workload.PodSetsOnNode` function only checks `Workload.spec.podSets[*].template.spec.tolerations`. It does not check the tolerations injected by the ResourceFlavor, which results in the taint not being tolerated and marks the node as unhealthy.
It only occurs during the time after the workload is admitted but before the kube-scheduler binds the node, and no other kueue workload is running on that node.

TAS NodeHotSwap does not mark a node as unhealthy if the required tolerations are automatically injected by the ResourceFlavor, even if they are missing from the original Pod template.

1. Use Kueue v0.17 (with `TASReplaceNodeOnNodeTaints=true` enabled by default).
2. Set up a node with a taint like `gpu-type=nvidia-a100:NoSchedule`.
3. Create a TAS `ResourceFlavor` with:
    * `spec.nodeTaints` matching that taint.
    * `spec.tolerations` tolerating that taint.
    * `spec.topologyName` using a hostname topology.
4. Create a Workload using that ResourceFlavor. (emulate a binding delay by setting a non-existent scheduler).
6. The workload is terminated by kueue

<details>

```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: tas-rf-toleration-repro
---
kind: Topology
metadata:
  name: tas-rf-toleration-repro
spec:
  levels:
  - nodeLabel: kubernetes.io/hostname
---
kind: ResourceFlavor
metadata:
  name: tas-rf-toleration-repro
spec:
  nodeLabels:
    tas-node: "true"
    gpu-type: nvidia-a100
  nodeTaints:
  - key: gpu-type
    value: nvidia-a100
    effect: NoSchedule
  tolerations:
  - key: gpu-type
    operator: Equal
    value: nvidia-a100
    effect: NoSchedule
  topologyName: tas-rf-toleration-repro
---
kind: ClusterQueue
metadata:
  name: tas-rf-toleration-repro
spec:
  namespaceSelector: {}
  resourceGroups:
  - coveredResources: ["cpu", "memory", "nvidia.com/gpu"]
    flavors:
    - name: tas-rf-toleration-repro
      resources:
      - name: cpu
        nominalQuota: 100
      - name: memory
        nominalQuota: 100Gi
      - name: nvidia.com/gpu
        nominalQuota: 24
---
kind: LocalQueue
metadata:
  namespace: tas-rf-toleration-repro
  name: tas
spec:
  clusterQueue: tas-rf-toleration-repro
```

```yaml
apiVersion: batch/v1
kind: Job
metadata:
  namespace: tas-rf-toleration-repro
  name: tas-rf-toleration-repro
  labels:
    kueue.x-k8s.io/queue-name: tas
spec:
  backoffLimit: 0
  template:
    metadata:
      annotations:
        kueue.x-k8s.io/podset-required-topology: kubernetes.io/hostname
    spec:
      schedulerName: repro-never-scheduler # enmurate pending state by using a non-existent scheduler
      restartPolicy: Never
      containers:
      - name: main
        image: busybox
        args: ["sleep", "3600"]
        resources:
          requests:
            cpu: "100m"
            memory: "64Mi"
            nvidia.com/gpu: "1"
          limits:
            cpu: "100m"
            memory: "64Mi"
            nvidia.com/gpu: "1"
```

</details>

We don't use AdmissionCheck `PodSetUpdates` in our own environment, but it seems like `AdmissionCheckState.PodSetUpdates[*].tolerations` might have the same problem.

**Environment**:
- Kubernetes version (use `kubectl version`): v1.35
- Kueue version (use `git describe --tags --dirty --always`):  v0.17.1
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Ladicle:
#### What type of PR is this?

/kind bug
/area tas

This PR updates TAS node taint checks to use the effective PodSet tolerations for admitted Workloads, including tolerations both from assigned ResourceFlavor and AdmissionCheck `PodSetUpdates`.

TAS NodeHotSwap currently classifies node taints based on `Workload.Spec.PodSets[*].Template.Spec.Tolerations`.
However, the Pod created after admission can also receive tolerations from the assigned ResourceFlavor and AdmissionCheck. In that case, TAS can incorrectly treat a healthy tainted node as unhealthy for the Workload, and eventually evict the Workload with `NodeFailures` even though the effective PodSpec tolerates the taint.

Fixes #11152

```release-note
TAS: Fixed NodeHotSwap with TASReplaceNodeOnNodeTaints enabled to evaluate node taints using effective Workload tolerations, including tolerations from AdmissionCheck PodSetUpdates.
```

### DIFF
--- a/pkg/controller/tas/node_controller.go
+++ b/pkg/controller/tas/node_controller.go
@@ -422,17 +422,7 @@ func (r *nodeReconciler) podSetsWithEffectiveTolerationsOnNode(ctx context.Conte
 	var result []kueue.PodSet
 	for i := range wl.Status.Admission.PodSetAssignments {
 		psa := &wl.Status.Admission.PodSetAssignments[i]
-		if psa.TopologyAssignment == nil || !utiltas.IsLowestLevelHostname(psa.TopologyAssignment.Levels) {
-			continue
-		}
-		assigned := false
-		for val := range utiltas.LowestLevelValues(psa.TopologyAssignment) {
-			if val == nodeName {
-				assigned = true
-				break
-			}
-		}
-		if !assigned {
+		if !utiltas.HasNodeInPodSetAssignment(psa, nodeName) {
 			continue
 		}
 
@@ -446,15 +436,19 @@ func (r *nodeReconciler) podSetsWithEffectiveTolerationsOnNode(ctx context.Conte
 		if err != nil {
 			return nil, err
 		}
-		effective.Template.Spec.Tolerations = append(effective.Template.Spec.Tolerations, info.Tolerations...)
 
 		for _, admissionCheck := range wl.Status.AdmissionChecks {
 			for _, podSetUpdate := range admissionCheck.PodSetUpdates {
 				if podSetUpdate.Name == psa.Name {
-					effective.Template.Spec.Tolerations = append(effective.Template.Spec.Tolerations, podSetUpdate.Tolerations...)
+					if err := info.Merge(podsetinfo.FromUpdate(&podSetUpdate)); err != nil {
+						return nil, fmt.Errorf("in admission check %q: %w", admissionCheck.Name, err)
+					}
 					break
 				}
 			}
+		}
+		if err := podsetinfo.Merge(&effective.Template.ObjectMeta, &effective.Template.Spec, info); err != nil {
+			return nil, err
 		}
 		result = append(result, *effective)
 	}
--- a/pkg/controller/tas/node_controller.go
+++ b/pkg/controller/tas/node_controller.go
@@ -52,8 +52,10 @@ import (
 	"sigs.k8s.io/kueue/pkg/controller/core"
 	"sigs.k8s.io/kueue/pkg/controller/tas/indexer"
 	"sigs.k8s.io/kueue/pkg/features"
+	podsetinfo "sigs.k8s.io/kueue/pkg/podset"
 	utilclient "sigs.k8s.io/kueue/pkg/util/client"
 	utilpod "sigs.k8s.io/kueue/pkg/util/pod"
+	utilpodset "sigs.k8s.io/kueue/pkg/util/podset"
 	"sigs.k8s.io/kueue/pkg/util/roletracker"
 	utiltas "sigs.k8s.io/kueue/pkg/util/tas"
 	"sigs.k8s.io/kueue/pkg/workload"
@@ -236,6 +238,7 @@ func (r *nodeReconciler) Delete(e event.TypedDeleteEvent[*corev1.Node]) bool {
 //+kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch
 //+kubebuilder:rbac:groups="",resources=pods,verbs=get;list;watch
 //+kubebuilder:rbac:groups="",resources=pods/status,verbs=patch
+//+kubebuilder:rbac:groups=kueue.x-k8s.io,resources=resourceflavors,verbs=get;list;watch
 //+kubebuilder:rbac:groups=kueue.x-k8s.io,resources=workloads,verbs=get;list;watch;patch
 
 func newNodeReconciler(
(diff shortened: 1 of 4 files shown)


mimowo:
/ok-to-test
cc @j-skiba ptal

j-skiba (commented):
/lgtm overall

Initially I was thinking about checking tolerations from the pod spec as a source of truth but it's tricky for situations when pods are not on the node for any reason (taints, any other issues related to node's health) so probably checking them this way is one of the ways. Other possibility I can think of is reading them from parent job but I'm not sure we get any better results from such approach (it would be more complex too).

j-skiba:
cc @amy

pkg/controller/tas/node_controller.go
```diff
@@ -399,6 +408,59 @@ func (r *nodeReconciler) getWorkloadStatus(
...
+		info, err := podsetinfo.FromAssignment(ctx, r.client, psa, ps)
+		if err != nil {
+			return nil, err
+		}
+		effective.Template.Spec.Tolerations = append(effective.Template.Spec.Tolerations, info.Tolerations...)
+
+		for _, admissionCheck := range wl.Status.AdmissionChecks {
+			for _, podSetUpdate := range admissionCheck.PodSetUpdates {
+				if podSetUpdate.Name == psa.Name {
+					effective.Template.Spec.Tolerations = append(effective.Template.Spec.Tolerations, podSetUpdate.Tolerations...)
```
mimowo: Instead of the code speciifc for tolerations could we use the `Merge` function as done here: https://github.com/kubernetes-sigs/kueue/blob/main/pkg/cache/scheduler/tas_flavor_snapshot.go#L829-L834

pkg/controller/tas/node_controller.go
```diff
@@ -399,6 +408,59 @@ func (r *nodeReconciler) getWorkloadStatus(
...
+		assigned := false
+		for val := range utiltas.LowestLevelValues(psa.TopologyAssignment) {
+			if val == nodeName {
+				assigned = true
+				break
+			}
+		}
+		if !assigned {
+			continue
+		}
```
mimowo, line 427: This seems very similar to `HasTASAssignmentOnNode`. I would prefer to introduce a new helper ``HasNodeInPodSetAssignment` by commonizing this inner part:

```golang
		if psa.TopologyAssignment == nil || !IsLowestLevelHostname(psa.TopologyAssignment.Levels) {
			continue
		}
		for domain := range InternalSeqFrom(psa.TopologyAssignment) {
			if len(domain.Values) > 0 && domain.Values[len(domain.Values)-1] == nodeName && domain.Count > 0 {
				return true
			}
		}

```

mimowo (commented):
LGTM overall, just nits

mimowo (commented):
Thank you 👍 
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16

tenzen-y:
/release-note-edit
```release-note
TAS: Fixed TASReplaceNodeOnNodeTaints to use effective Workload tolerations from ResourceFlavors and AdmissionCheck PodSetUpdates.
```

mimowo:
/release-note-edit
```release-note
TAS: Fixed NodeHotSwap with TASReplaceNodeOnNodeTaints enabled to evaluate node taints using effective Workload tolerations, including tolerations from AdmissionCheck PodSetUpdates.
```

Issue #10296, opened 2026-04-06
[Flaky E2E] Retrying due to error: could not find a JWS signature in the cluster-info ConfigMap for token ID "abcdef"


mbobrovskyi:
**Which test is flaking?**:

[discovery] Retrying due to error: could not find a JWS signature in the cluster-info ConfigMap for token ID "abcdef"
ERROR: failed to create cluster: failed to join node with kubeadm: command "docker exec --privileged kind-worker kubeadm join --config /kind/kubeadm.conf --v=6" failed with error: exit status 137

**Link to failed CI job or steps to reproduce locally**:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/periodic-kueue-test-e2e-main-1-33/2040077147321667584

**Failure message or logs**:
```
I0403 14:49:18.562117     183 round_trippers.go:632] "Response" verb="GET" url="https://kind-control-plane:6443/api/v1/namespaces/kube-public/configmaps/cluster-info?timeout=10s" status="200 OK" milliseconds=15
I0403 14:49:18.562482     183 token.go:250] [discovery] Retrying due to error: could not find a JWS signature in the cluster-info ConfigMap for token ID "abcdef"
 ✗ Joining worker nodes 🚜
DEBUG: kubeadmjoin/join.go:147] I0403 14:49:18.544000     184 join.go:423] [preflight] found NodeName empty; using OS hostname as NodeName
I0403 14:49:18.544148     184 joinconfiguration.go:83] loading configuration from "/kind/kubeadm.conf"
W0403 14:49:18.544819     184 joinconfiguration.go:113] [config] WARNING: Ignored configuration document with GroupVersionKind kubeadm.k8s.io/v1beta3, Kind=ClusterConfiguration
W0403 14:49:18.544903     184 joinconfiguration.go:113] [config] WARNING: Ignored configuration document with GroupVersionKind kubeadm.k8s.io/v1beta3, Kind=InitConfiguration
W0403 14:49:18.544922     184 common.go:101] your configuration file uses a deprecated API spec: "kubeadm.k8s.io/v1beta3" (kind: "JoinConfiguration"). Please use 'kubeadm config migrate --old-config old-config-file --new-config new-config-file', which will write the new, similar spec using a newer API version.
[...]

PR #11190, opened 2026-05-14
fix: Return on error in e2e cluster_create instead of continuing execution

ivnovakov:
#### What type of PR is this?

/kind flake
/area testing

After looking into flaky e2e logs, the issue with `cluster_create` in `hack/testing/e2e-common.sh` was found.

`cluster_create` previously masked failures with `|| { echo "unable to start the $cluster cluster"; cat "$ARTIFACTS/$cluster-create.log"; }`, which returns 0 to the shell.
So, when `kind create cluster` failed, the function kept going to `kubectl config use-context`, which then errored with a misleading `error: no context exists with the name: "kind-kind"`.
Meanwhile, the EXIT-trap cleanup ran kubectl/kind commands against the non-existent or half-built cluster, producing a cascade of `error: no context exists`, `unknown cluster`, and `connection refused` errors.

This PR adds following changes in:
- **`cluster_collect_artifacts`** — prevents errors in case there is no cluster or apiserver.
    - `kind_cluster_exists` is false, skip everything; prints
      `Skipping artifact collection for '<name>': cluster does not exist.`
    - apiserver unreachable (containers exist, but `kubeadm init` never finished),
      skip `kubectl describe pods`; prints
      `Skipping pod descriptions for '<name>': API server is not reachable.`
- **`cluster_create`** — `kind create cluster` failure now returns 1 immediately.
  This replaces the vague `unable to start the kind cluster `
  prefix with `ERROR: Unable to create kind cluster '<name>'.`, and stops
  the function from continuing into `kubectl config use-context`.

Verified locally on macOS/colima against two reproducers; in each case the main script output ends with a clear single cause failure block and the EXIT-trap cleanup is quiet.

- **`kind create` returns non-zero**
  (`E2E_KIND_VERSION=kindest/node:v0.0.0-bogus make test-e2e`):
  output ends with `ERROR: Unable to create kind cluster 'kind'.` + dumped kind log, followed by
  `Skipping artifact collection for 'kind': cluster does not exist.`
- **Compound failure matching the CI run** — foreground `docker pull` fails while background `kind create` is still running
  (`KUBERAY_VERSION=does-not-exist make test-e2e`):
  main script ends with the pull-failure message, then
  `Skipping pod descriptions for 'kind': API server is not reachable.`
  (the API-reachability guard, since kind got far enough to create containers). The background subshell's deferred kind log dump appears after
  `make: *** Error 1` but is now prefixed by `ERROR: Unable to create kind cluster 'kind'.`
  instead of the misleading `unable to start the kind cluster `.

```release-note
NONE
```

### DIFF
--- a/hack/testing/e2e-common.sh
+++ b/hack/testing/e2e-common.sh
@@ -230,13 +230,23 @@ function cluster_collect_artifacts {
     local name=$1
     local kubeconfig=${2:-}
 
+    if ! kind_cluster_exists "$name"; then
+        echo "Skipping artifact collection for '$name': cluster does not exist."
+        return 0
+    fi
+
+    $KIND export logs "$ARTIFACTS" --name "$name" || true
+
     local -a kubectl_args=()
     if [[ -n "${kubeconfig}" ]]; then
         kubectl_args=(--kubeconfig="${kubeconfig}")
     fi
 
-    kubectl config ${kubectl_args[@]+"${kubectl_args[@]}"} use-context "kind-${name}" || true
-    $KIND export logs "$ARTIFACTS" --name "$name" || true
+    if ! kubectl ${kubectl_args[@]+"${kubectl_args[@]}"} version --request-timeout=30s >/dev/null 2>&1; then
+        echo "Skipping pod descriptions for '$name': API server is not reachable."
+        return 0
+    fi
+
     kubectl describe pods ${kubectl_args[@]+"${kubectl_args[@]}"} -n kueue-system > "$ARTIFACTS/${name}-kueue-system-pods.log" || true
     kubectl describe pods ${kubectl_args[@]+"${kubectl_args[@]}"} > "$ARTIFACTS/${name}-default-pods.log" || true
 }
@@ -368,12 +378,17 @@ function cluster_create {
         cat "$kind_config"
     fi
 
-    $KIND create cluster --name "$cluster" --image "$E2E_KIND_VERSION" --config "$kind_config" --kubeconfig="$kubeconfig" --wait 1m -v 5  > "$ARTIFACTS/$cluster-create.log" 2>&1 \
-    ||  { echo "unable to start the $cluster cluster "; cat "$ARTIFACTS/$cluster-create.log" ; }
+    if ! $KIND create cluster --name "$cluster" --image "$E2E_KIND_VERSION" \
+            --config "$kind_config" --kubeconfig="$kubeconfig" --wait 5m -v 5 \
+            > "$ARTIFACTS/$cluster-create.log" 2>&1; then
+        echo "ERROR: Unable to create kind cluster '$cluster'." >&2
+        cat "$ARTIFACTS/$cluster-create.log" >&2
+        return 1
+    fi
 
     kubectl config --kubeconfig="$kubeconfig" use-context "kind-$cluster"
     # wait for nodes to become ready before loading images or deploying components
-    kubectl wait --kubeconfig="$kubeconfig" --for=condition=Ready node --all --timeout=300s
+    kubectl wait --kubeconfig="$kubeconfig" --for=condition=Ready node --all --timeout=5m
     kubectl get nodes --kubeconfig="$kubeconfig" > "$ARTIFACTS/$cluster-nodes.log" || true
     kubectl describe pods --kubeconfig="$kubeconfig" -n kube-system > "$ARTIFACTS/$cluster-system-pods.log" || true
 }


ivnovakov:
Related to https://github.com/kubernetes-sigs/kueue/issues/10296.

hack/testing/e2e-common.sh
```diff
@@ -368,12 +380,17 @@ function cluster_create {
         cat "$kind_config"
     fi
 
-    $KIND create cluster --name "$cluster" --image "$E2E_KIND_VERSION" --config "$kind_config" --kubeconfig="$kubeconfig" --wait 1m -v 5  > "$ARTIFACTS/$cluster-create.log" 2>&1 \
-    ||  { echo "unable to start the $cluster cluster "; cat "$ARTIFACTS/$cluster-create.log" ; }
+    if ! $KIND create cluster --name "$cluster" --image "$E2E_KIND_VERSION" \
+            --config "$kind_config" --kubeconfig="$kubeconfig" --wait 1m -v 5 \
```
mimowo: @ivnovakov is this fixing https://github.com/kubernetes-sigs/kueue/issues/10296, and the issue wasn't about the timeout as we initially assumed?

hack/testing/e2e-common.sh
```diff
@@ -368,12 +380,17 @@ function cluster_create {
...
+            > "$ARTIFACTS/$cluster-create.log" 2>&1; then
+        echo "ERROR: Unable to create kind cluster '$cluster'." >&2
+        cat "$ARTIFACTS/$cluster-create.log" >&2
+        return 1
```
mbobrovskyi, line 386: Where we handle this result?

hack/testing/e2e-common.sh
ivnovakov: Yes, it wasn't about timeout. Added this comment on the issue clarifying that https://github.com/kubernetes-sigs/kueue/issues/10296#issuecomment-4450978106.

mbobrovskyi: I still propose increasing it to 5m, even if the issue was not related to a timeout.

ivnovakov: Agree, I don't see a reason why not. If cluster is ready it will proceed right away.

hack/testing/e2e-common.sh
ivnovakov: It's catched by `set -o errexit`.

hack/testing/e2e-common.sh
```diff
@@ -230,13 +230,25 @@ function cluster_collect_artifacts {
...
+
     local -a kubectl_args=()
     if [[ -n "${kubeconfig}" ]]; then
         kubectl_args=(--kubeconfig="${kubeconfig}")
     fi
 
     kubectl config ${kubectl_args[@]+"${kubectl_args[@]}"} use-context "kind-${name}" || true
-    $KIND export logs "$ARTIFACTS" --name "$name" || true
+
+    if ! kubectl ${kubectl_args[@]+"${kubectl_args[@]}"} version --request-timeout=5s >/dev/null 2>&1; then
```
mbobrovskyi: Why 5s?

ivnovakov: Just to fail fast if not present. Should I use 1m or 5m then?

hack/testing/e2e-common.sh
```diff
@@ -230,13 +230,25 @@ function cluster_collect_artifacts {
     local name=$1
     local kubeconfig=${2:-}
 
+    if ! kind_cluster_exists "$name"; then
+        echo "Skipping artifact collection for '$name': cluster does not exist."
+        return 0
+    fi
+
+    $KIND export logs "$ARTIFACTS" --name "$name" || true
```
mbobrovskyi, line 238: Do we need to specify kubeconfig for this command?

mbobrovskyi: Previously we use `use-context` command.

hack/testing/e2e-common.sh
mbobrovskyi: Why we actually need this if?

hack/testing/e2e-common.sh
```diff
@@ -230,13 +230,25 @@ function cluster_collect_artifacts {
...
+
     local -a kubectl_args=()
     if [[ -n "${kubeconfig}" ]]; then
         kubectl_args=(--kubeconfig="${kubeconfig}")
     fi
 
     kubectl config ${kubectl_args[@]+"${kubectl_args[@]}"} use-context "kind-${name}" || true
```
mbobrovskyi: I think we can remove this command. We are using `--kubeconfig` flag in all commands.

hack/testing/e2e-common.sh
ivnovakov: To remove such errors during cleanup, when cluster exists but apiserver isn't up.

```
E0403 14:48:49.594259   32912 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: Get \"http://localhost:8080/api?timeout=32s\": dial tcp [::1]:8080: connect: connection refused"
E0403 14:48:49.596420   32912 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: Get \"http://localhost:8080/api?timeout=32s\": dial tcp [::1]:8080: connect: connection refused"
The connection to the server localhost:8080 was refused - did you specify the right host or port?

```

hack/testing/e2e-common.sh
ivnovakov: There is no need, only name specification. Internally it uses `docker logs` to fetch logs.

hack/testing/e2e-common.sh
ivnovakov: Agree, removed.

hack/testing/e2e-common.sh
mbobrovskyi: Ah, sorry, you’re right. This is the `kind` command.

mbobrovskyi (commented):
Thank you!

/lgtm
/approve

/hold if @mimowo want to add something

hack/testing/e2e-common.sh
mimowo: Fail fast makes sense, but 5s sounds very aggressive on a loaded CI, maybe we can use 30s since the timeout used by memcache.go seems to be 32s.

mimowo:
lgtm, just proposed to extend the timeout for the fail fast check

hack/testing/e2e-common.sh
ivnovakov: Done, changed to 30s.

mimowo (commented):
/lgtm
/unhold
/cherrypick release-0.17
/cherrypick release-0.16

PR #11192, opened 2026-05-14
Exclude kube-system from webhooks targeting common resources

### LINKED ISSUE #11006: Exclude kube-system namespace from webhooks targeting common resource types

dkaluza:
#### What type of PR is this?
/kind feature
/area integrations

Adds `kube-system` and `kueue-system` (or other release namespace in case of helm installation) namespace excludes to webhooks targeting common resource types.
(Custom resources that are not added by Kueue)

This makes sure Kueue deployment does not suspend crucial system components, e.g. jobs run by `kubeadm`.

Fixes #11006

I haven't modified webhook selectors for Kueue's own resources as those shouldn't be run by other crucial system components. But if you think it is worth to also exclude those, just let me know.

Prepared with Antigravity, but carefully reviewed to not miss any resource.

```release-note
Kueue integration webhooks now consistently exclude `kube-system` and the Kueue
installation namespace. For manifest-based installs this is `kueue-system`; for Helm
installs this is the Helm release namespace.

Previously, Pod, Deployment, and StatefulSet integrations already excluded these
namespaces, while other workload integrations did not.

ACTION REQUIRED: If you run Kueue-managed workloads in `kube-system` or in the namespace
where Kueue is installed, move them to another namespace before upgrading.

If this is not possible, update `managedJobsNamespaceSelector` and the relevant webhook
`namespaceSelector`s to include those namespaces. This setup is discouraged, as it may
affect system workloads and disrupt cluster upgrades.
```

### DIFF
--- a/charts/kueue/templates/webhook/manifests.yaml
+++ b/charts/kueue/templates/webhook/manifests.yaml
@@ -39,10 +39,17 @@ webhooks:
         path: /mutate-workload-codeflare-dev-v1beta2-appwrapper
     failurePolicy: Fail
     name: mappwrapper.kb.io
-    {{- if (hasKey $managerConfig "managedJobsNamespaceSelector") }}
     namespaceSelector:
-      {{- toYaml $managerConfig.managedJobsNamespaceSelector | nindent 6 }}
-    {{- end }}
+      {{- if (hasKey $managerConfig "managedJobsNamespaceSelector") }}
+        {{- toYaml $managerConfig.managedJobsNamespaceSelector | nindent 6 }}
+      {{- else }}
+      matchExpressions:
+        - key: kubernetes.io/metadata.name
+          operator: NotIn
+          values:
+            - kube-system
+            - '{{ .Release.Namespace }}'
+      {{- end }}
     rules:
       - apiGroups:
           - workload.codeflare.dev
@@ -119,10 +126,17 @@ webhooks:
         path: /mutate-kubeflow-org-v1-jaxjob
     failurePolicy: Fail
     name: mjaxjob.kb.io
-    {{- if (hasKey $managerConfig "managedJobsNamespaceSelector") }}
     namespaceSelector:
-      {{- toYaml $managerConfig.managedJobsNamespaceSelector | nindent 6 }}
-    {{- end }}
+      {{- if (hasKey $managerConfig "managedJobsNamespaceSelector") }}
+        {{- toYaml $managerConfig.managedJobsNamespaceSelector | nindent 6 }}
+      {{- else }}
+      matchExpressions:
+        - key: kubernetes.io/metadata.name
+          operator: NotIn
+          values:
+            - kube-system
+            - '{{ .Release.Namespace }}'
+      {{- end }}
     rules:
       - apiGroups:
           - kubeflow.org
@@ -143,10 +157,17 @@ webhooks:
         path: /mutate-batch-v1-job
     failurePolicy: Fail
     name: mjob.kb.io
-    {{- if (hasKey $managerConfig "managedJobsNamespaceSelector") }}
     namespaceSelector:
-      {{- toYaml $managerConfig.managedJobsNamespaceSelector | nindent 6 }}
-    {{- end }}
+      {{- if (hasKey $managerConfig "managedJobsNamespaceSelector") }}
+        {{- toYaml $managerConfig.managedJobsNamespaceSelector | nindent 6 }}
+      {{- else }}
+      matchExpressions:
+        - key: kubernetes.io/metadata.name
+          operator: NotIn
+          values:
+            - kube-system
+            - '{{ .Release.Namespace }}'
+      {{- end }}
     rules:
       - apiGroups:
           - batch
@@ -167,10 +188,17 @@ webhooks:
         path: /mutate-jobset-x-k8s-io-v1alpha2-jobset
     failurePolicy: Fail
     name: mjobset.kb.io
-    {{- if (hasKey $managerConfig "managedJobsNamespaceSelector") }}
     namespaceSelector:
-      {{- toYaml $managerConfig.managedJobsNamespaceSelector | nindent 6 }}
-    {{- end }}
+      {{- if (hasKey $managerConfig "managedJobsNamespaceSelector") }}
+        {{- toYaml $managerConfig.managedJobsNamespaceSelector | nindent 6 }}
+      {{- else }}
+      matchExpressions:
+        - key: kubernetes.io/metadata.name
+          operator: NotIn
+          values:
+            - kube-system
+            - '{{ .Release.Namespace }}'
+      {{- end }}
     rules:
       - apiGroups:
           - jobset.x-k8s.io
@@ -191,10 +219,17 @@ webhooks:
         path: /mutate-leaderworkerset-x-k8s-io-v1-leaderworkerset
     failurePolicy: Fail
     name: mleaderworkerset.kb.io
-    {{- if (hasKey $managerConfig "managedJobsNamespaceSelector") }}
     namespaceSelector:
-      {{- toYaml $managerConfig.managedJobsNamespaceSelector | nindent 6 }}
-    {{- end }}
+      {{- if (hasKey $managerConfig "managedJobsNamespaceSelector") }}
+        {{- toYaml $managerConfig.managedJobsNamespaceSelector | nindent 6 }}
+      {{- else }}
+      matchExpressions:
+        - key: kubernetes.io/metadata.name
+          operator: NotIn
+          values:
+            - kube-system
+            - '{{ .Release.Namespace }}'
+      {{- end }}
     rules:
       - apiGroups:
           - leaderworkerset.x-k8s.io
@@ -216,10 +251,17 @@ webhooks:
         path: /mutate-kubeflow-org-v2beta1-mpijob
     failurePolicy: Fail
     name: mmpijob.kb.io
-    {{- if (hasKey $managerConfig "managedJobsNamespaceSelector") }}
     namespaceSelector:
-      {{- toYaml $managerConfig.managedJobsNamespaceSelector | nindent 6 }}
-    {{- end }}
+      {{- if (hasKey $managerConfig "managedJobsNamespaceSelector") }}
+        {{- toYaml $managerConfig.managedJobsNamespaceSelector | nindent 6 }}
+      {{- else }}
+      matchExpressions:
+        - key: kubernetes.io/metadata.name
+          operator: NotIn
+          values:
+            - kube-system
+            - '{{ .Release.Namespace }}'
+      {{- end }}
     rules:
       - apiGroups:
           - kubeflow.org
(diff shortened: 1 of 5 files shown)


hack/processing-plan.yaml
```diff
@@ -397,234 +397,416 @@ files:
       - type: INSERT_TEXT
         key: .webhooks.[].name
         value: |
+          namespaceSelector:
             {{- if (hasKey $managerConfig "managedJobsNamespaceSelector") }}
-            namespaceSelector:
               {{- toYaml $managerConfig.managedJobsNamespaceSelector | nindent 6 }}
+            {{- else }}
```
dkaluza: I do not like the fact that we have to modify 200+ lines to make such small change, but it looks like we want to be really explicit about the changes here, so didn't want to change this.

config/components/webhook/kustomization.yaml
```diff
@@ -50,19 +20,58 @@ patches:
...
+        - webhooks.[name=mpod.kb.io].namespaceSelector
+        - webhooks.[name=mpytorchjob.kb.io].namespaceSelector
+        - webhooks.[name=mraycluster.kb.io].namespaceSelector
+        - webhooks.[name=mrayjob.kb.io].namespaceSelector
+        - webhooks.[name=mrayservice.kb.io].namespaceSelector
+        - webhooks.[name=msparkapplication.kb.io].namespaceSelector
+        - webhooks.[name=mstatefulset.kb.io].namespaceSelector
+        - webhooks.[name=mtfjob.kb.io].namespaceSelector
+        - webhooks.[name=mtrainjob.kb.io].namespaceSelector
+        - webhooks.[name=mxgboostjob.kb.io].namespaceSelector
```
dkaluza, line 51: This unfortunately still requires registration of new webhook here when there is a new integration.

This is not the case if we just do `*` but then we will also target Kueue crds... This can be solved by splitting those in some way, but I'm not convinced this is worth it right now.

config/components/webhook/kustomization.yaml
```diff
@@ -6,37 +6,7 @@ configurations:
...
-    - name: mstatefulset.kb.io
-      namespaceSelector:
-        matchExpressions:
-        - key: kubernetes.io/metadata.name
-          operator: NotIn
-          values:
-            - kube-system
-            - kueue-system
-- patch: |-
+ - patch: |-
```
olekzabl:
```suggestion
- patch: |-
```

mimowo:
/test pull-kueue-test-e2e-main-1-34
look like infra, we are fixing currently similar issues by retry: https://github.com/kubernetes-sigs/kueue/pull/11238#issuecomment-4460782452

hack/processing-plan.yaml
```diff
@@ -277,8 +277,8 @@ files:
           failurePolicy: Ignore
           {{- end }}
           namespaceSelector:
-            {{- if (hasKey $managerConfig "managedJobsNamespaceSelector") -}}
-              {{- toYaml $managerConfig.managedJobsNamespaceSelector | nindent 6 -}}
+            {{- if (hasKey $managerConfig "managedJobsNamespaceSelector") }}
```
mimowo, line 275: this par looks duplicated for all crds:

```
          namespaceSelector:
            {{- if (hasKey $managerConfig "managedJobsNamespaceSelector") -}}
              {{- toYaml $managerConfig.managedJobsNamespaceSelector | nindent 6 -}}
            {{- else }}
            matchExpressions:
              - key: kubernetes.io/metadata.name
                operator: NotIn
                values:
                  - kube-system
                  - '{{ .Release.Namespace }}'
            {{- end }}
```
i'm wondering if there would be a way to introduce some kind of placeholder for that in processing-plain.yaml, wdyt?

This is not a blocker, but it would be good to research this topic on this occasion so that we can open follow up issue to maybe introduce that capability if not existing currently.

olekzabl: Would [named templates](https://helm.sh/docs/chart_template_guide/named_templates/) help?

config/components/webhook/kustomization.yaml
```diff
@@ -50,19 +20,58 @@ patches:
...
+      fieldPaths:
+        - webhooks.[name=vappwrapper.kb.io].namespaceSelector
+        - webhooks.[name=vdeployment.kb.io].namespaceSelector
+        - webhooks.[name=vjaxjob.kb.io].namespaceSelector
+        - webhooks.[name=vjob.kb.io].namespaceSelector
+        - webhooks.[name=vjobset.kb.io].namespaceSelector
+        - webhooks.[name=vleaderworkerset.kb.io].namespaceSelector
+        - webhooks.[name=vmpijob.kb.io].namespaceSelector
+        - webhooks.[name=vpaddlejob.kb.io].namespaceSelector
+        - webhooks.[name=vpod.kb.io].namespaceSelector
```
olekzabl, line 66: FYI this target path is exactly the source path.

An AI bot told me it'll still work, even with `create: true` (just silently ignoring this target).
So maybe this is desired, for the completeness of the target list.
Or maybe it could be considered confusing/misleading.

I'm leaving the choice to you; I just wanted to make sure you're aware of this.

hack/processing-plan.yaml
dkaluza: I believe it should be generally doable with just a better yq selector(condition here), no need for named template yet, but generally I agree named templates also look like a viable way to achieve deduplication in similar cases. 

I got some strange behavior of our yaml processor with new condition(not pushed yet), I will investigate on Monday.

config/components/webhook/kustomization.yaml
dkaluza: Thanks for pointing this out.

Generally I consider this beneficial, as it provides mentioned completeness. In this way the number of webhooks in both selects is the same. Also they can be easier matched against expected number, you do not have to always remember this special case.

> An AI bot told me it'll still work, even with create: true (just silently ignoring this target).

According to [kustomize docs](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/transformerconfigs/README.md#fieldspec) `create: true` just means to create the target path if it does not exists instead of throwing an error. 
Generally this is a replacement of the same content with the same content, so I wouldn't expect any problems here.

mimowo:
/release-note-edit
```release-note
Kueue integration webhooks will no longer target kube-system or kueue-system namespace.

ACTION REQUIRED: If you are running Kueue-managed workloads in kube-system namespace or
namespace in which  Kueue is installed please move those workloads to a different namespace. 

If  this is not feasible then you may need to adjust your `managedJobsNamespaceSelector` and
webhook configuration to also cover these namespaces. However such a deployment is
discouraged as may result in disruptions during cluster upgrades.
```

charts/kueue/templates/webhook/manifests.yaml
```diff
@@ -39,10 +39,17 @@ webhooks:
         path: /mutate-workload-codeflare-dev-v1beta2-appwrapper
     failurePolicy: Fail
     name: mappwrapper.kb.io
-    {{- if (hasKey $managerConfig "managedJobsNamespaceSelector") }}
```
mimowo, line 42: qq: does this change:
```
  {{- if (hasKey $managerConfig "managedJobsNamespaceSelector") }}
    namespaceSelector:
      {{- toYaml $managerConfig.managedJobsNamespaceSelector | nindent 6 }}
    {{- end }}
```
to 
```
    namespaceSelector:
      {{- if (hasKey $managerConfig "managedJobsNamespaceSelector") }}
        {{- toYaml $managerConfig.managedJobsNamespaceSelector | nindent 6 }}
      {{- else }}
      matchExpressions:
        - key: kubernetes.io/metadata.name
          operator: NotIn
          values:
            - kube-system
            - '{{ .Release.Namespace }}'
      {{- end }}
```
impact the default configuration by Helm? IIUC it does because by default we would keep managedJobsNamespaceSelector as nil, and so by default Helm would cover kueue-system and kube-system, but I want to double check.

config/components/webhook/kustomization.yaml
mimowo: Cool, do I understand that this change only then impacts all workloads other than Pod, Deployment, StatefulSet? Because Pod, Deployment, StatefulSet already had the exclusions in the deleted code here:
```
- patch: |-
    apiVersion: admissionregistration.k8s.io/v1
    kind: MutatingWebhookConfiguration
    metadata:
      name: mutating-webhook-configuration
    webhooks:
    - name: mpod.kb.io
      namespaceSelector:
        matchExpressions:
        - key: kubernetes.io/metadata.name
          operator: NotIn
          values:
            - kube-system
            - kueue-system
    - name: mdeployment.kb.io
      namespaceSelector:
        matchExpressions:
        - key: kubernetes.io/metadata.name
          operator: NotIn
          values:
            - kube-system
            - kueue-system
    - name: mstatefulset.kb.io
      namespaceSelector:
        matchExpressions:
        - key: kubernetes.io/metadata.name
          operator: NotIn
          values:
            - kube-system
            - kueue-system
```

dkaluza: That's correct.

hack/tools/yaml-processor/yamlproc/txtinserter.go
```diff
@@ -73,15 +72,14 @@ func (ti *TextInserter) insertBelowKey(yamlData []byte, opts InsertOptions) ([]b
 		trimmedLine := strings.TrimSpace(line)
 		buffer.WriteString(line + "\n")
 
-		if slices.Contains(keyLines, i+offset) {
+		if slices.Contains(keyLines, i) {
```
dkaluza, line 75: As keyLines are from the original yaml, we shouldn't be skipping lines if we add a new snippet.

hack/processing-plan.yaml
dkaluza: Done, needed a small hotfix to make this feasible, added a test for that.

charts/kueue/templates/webhook/manifests.yaml
dkaluza: Right now default helm deployment has namespace excludes only for the mentioned:
* statefulsets
* pods
* deployments

After this PR the namespaces will be excluded for all of the workload source types by default . Which makes those consistent across types and aligned with the release manifests.

mimowo: Cool, let's capture that in the release note

mimowo:
/release-note-edit
```release-note
Kueue integration webhooks now consistently exclude `kube-system` and the Kueue
installation namespace. For manifest-based installs this is `kueue-system`; for Helm
installs this is the Helm release namespace.

Previously, Pod, Deployment, and StatefulSet integrations already excluded these
namespaces, while other workload integrations did not.

ACTION REQUIRED: If you run Kueue-managed workloads in `kube-system` or in the namespace
where Kueue is installed, move them to another namespace before upgrading.

If this is not possible, update `managedJobsNamespaceSelector` and the relevant webhook
`namespaceSelector`s to include those namespaces. This setup is discouraged, as it may
affect system workloads and disrupt cluster upgrades.
```

mimowo:
/lgtm
/approve
Thank you, nice to see closing the inconsistencies between configuration for different Jobs 👍  I'm even wondering if we should cherrypick this as a consistency bug.

Any opinion @tenzen-y ?

mimowo:
> Thank you, nice to see closing the inconsistencies between configuration for different Jobs 👍 I'm even wondering if we should cherrypick this as a consistency bug.
> Any opinion @tenzen-y ?

I think we can skip the cherrypick as of now because no user requested the change as a bugfix for now. We will reconsider when this is needed to CP another bugfix, or if a user requests.

PR #11195, opened 2026-05-14
Finish old workload slice only after new slice is admitted

### LINKED ISSUE #9015: Elastic Workloads: During scale-up, old WorkloadSlice is finished but new WorkloadSlice is not admitted, leaving Job active and unsuspended
During elastic scale-up, the controller may mark the existing WorkloadSlice as finished before the replacement slice is successfully admitted. If the admission of the new slice fails, for example due to transient API or scheduling errors, the Job can remain active and unsuspended while no slice is admitted and capacity accounting is not restored. This results in an inconsistent state where the Job appears to be running despite having no admitted WorkloadSlice, which should instead trigger suspension and quota restoration.

**What happened**:

This surfaced in #5897

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

sohankunkerkar:
#### What type of PR is this?
/kind bug

Fixes #9015 

This implements the "admit first, finish old after" approach discussed by @ichekrygin and @mimowo in #9015. The old slice is now only finished inside the admission success path, so if admission fails the old slice retains its quota and the job keeps running. There is a brief window where both slices are in the cache (old not yet finished, new just admitted).                                                                                                            This is conservative over-counting (blocks other admits, never allows over-commitment) and resolves within one API round-trip.

```release-note
ElasticJobsViaWorkloadSlices: Fix quota leak during elastic workload scale-up where old slice was finished before replacement slice was admitted.
```

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -449,10 +449,9 @@ func (s *Scheduler) processEntry(
 
 	s.waitForPodsReadyIfBlocked(ctx, log, e)
 
+	// Copy ClusterName from old slice before admission (needed for MultiKueue).
 	if features.Enabled(features.ElasticJobsViaWorkloadSlices) && oldWorkloadSlice != nil {
-		if err := s.replaceOldWorkloadSlice(ctx, log, e, oldWorkloadSlice); err != nil {
-			return
-		}
+		e.Obj.Status.ClusterName = oldWorkloadSlice.WorkloadInfo.Obj.Status.ClusterName
 	}
 
 	if features.Enabled(features.ConcurrentAdmission) && concurrentadmission.IsVariant(e.Obj) {
@@ -463,7 +462,7 @@ func (s *Scheduler) processEntry(
 	}
 
 	e.markNominated()
-	if err := s.admit(ctx, e, cq); err != nil {
+	if err := s.admit(ctx, e, cq, oldWorkloadSlice); err != nil {
 		e.inadmissibleMsg = fmt.Sprintf("Failed to admit workload: %v", err)
 	}
 }
@@ -531,18 +530,14 @@ func (s *Scheduler) waitForPodsReadyIfBlocked(ctx context.Context, log logr.Logg
 	log.V(5).Info("Finished waiting for all admitted workloads to be in the PodsReady condition")
 }
 
-// replaceOldWorkloadSlice deactivates the old slice and finalizes its status.
-// The new slice's clusterName is copied from the old one so that MultiKueue
-// placement stays consistent across the replacement. If the subsequent admit
-// fails, the old slice may end up Finished while the new one is not admitted;
-// downstream controllers handle suspension/eviction.
-func (s *Scheduler) replaceOldWorkloadSlice(ctx context.Context, log logr.Logger, e *entry, oldWorkloadSlice *preemption.Target) error {
-	e.Obj.Status.ClusterName = oldWorkloadSlice.WorkloadInfo.Obj.Status.ClusterName
+// replaceOldWorkloadSlice finishes the old slice after the new slice has been
+// admitted. Called inside the admit success path so the old slice is only
+// finished when the new one is confirmed. If this fails, the job reconciler's
+// EnsureWorkloadSlices detects both slices admitted and finishes the old one.
+func (s *Scheduler) replaceOldWorkloadSlice(ctx context.Context, log logr.Logger, e *entry, oldWorkloadSlice *preemption.Target) {
 	if err := s.replaceWorkloadSlice(ctx, oldWorkloadSlice.WorkloadInfo.ClusterQueue, e.Obj, oldWorkloadSlice.WorkloadInfo.Obj.DeepCopy()); err != nil {
-		log.Error(err, "Failed to replace workload slice")
-		return err
+		log.Error(err, "Failed to finish old workload slice after admitting replacement; job reconciler will handle recovery")
 	}
-	return nil
 }
 
 type entryStatus string
@@ -788,7 +783,7 @@ func updateAssignmentForTAS(log logr.Logger, snapshot *schdcache.Snapshot, cq *s
 // the entry, and asynchronously updates the object in the apiserver after
 // assuming it in the cache.
 // Note: this does not necessarily make the workload "admitted".
-func (s *Scheduler) admit(ctx context.Context, e *entry, cq *schdcache.ClusterQueueSnapshot) error {
+func (s *Scheduler) admit(ctx context.Context, e *entry, cq *schdcache.ClusterQueueSnapshot, oldWorkloadSlice *preemption.Target) error {
 	log := ctrl.LoggerFrom(ctx)
 	admission := &kueue.Admission{
 		ClusterQueue:      e.ClusterQueue,
@@ -816,6 +811,9 @@ func (s *Scheduler) admit(ctx context.Context, e *entry, cq *schdcache.ClusterQu
 			s.recordWorkloadAdmissionMetrics(log, newWorkload, e.Obj, admission, consideredStr)
 
 			log.V(2).Info("Workload successfully admitted and assigned flavors", "assignments", admission.PodSetAssignments)
+			if features.Enabled(features.ElasticJobsViaWorkloadSlices) && oldWorkloadSlice != nil {
+				s.replaceOldWorkloadSlice(ctx, log, e, oldWorkloadSlice)
+			}
 			return
 		}
 		// Ignore errors because the workload or clusterQueue could have been deleted
--- a/test/integration/singlecluster/controller/jobs/job/job_controller_test.go
+++ b/test/integration/singlecluster/controller/jobs/job/job_controller_test.go
@@ -4326,6 +4326,52 @@ var _ = ginkgo.Describe("Job with elastic jobs via workload-slices support", gin
 		util.ExpectFinishedWorkloadsTotalMetric(clusterQueue, highPriorityClass.Name, 0)
 		util.ExpectLQFinishedWorkloadsTotalMetric(localQueue, highPriorityClass.Name, 0)
 	})
+
+	ginkgo.It("Should keep old workload-slice unfinished while replacement slice is pending", func() {
+		features.SetFeatureGateDuringTest(ginkgo.GinkgoTB(), features.ElasticJobsViaWorkloadSlices, true)
+
+		elasticJob := testingjob.MakeJob("job-pending-replacement", ns.Name).
+			SetAnnotation(workloadslicing.EnabledAnnotationKey, workloadslicing.EnabledAnnotationValue).
+			Queue(kueue.LocalQueueName(localQueue.Name)).
+			Request(corev1.ResourceCPU, "1000m").
+			Parallelism(3).
+			Completions(int32(cpuNominalQuota + 1)).
+			Obj()
+
+		ginkgo.By("creating an elastic job")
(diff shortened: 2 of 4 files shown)


mimowo:
@yaroslava-serdiuk @ichekrygin ptal

pkg/scheduler/scheduler.go
```diff
@@ -531,18 +530,14 @@ func (s *Scheduler) waitForPodsReadyIfBlocked(ctx context.Context, log logr.Logg
...
-	e.Obj.Status.ClusterName = oldWorkloadSlice.WorkloadInfo.Obj.Status.ClusterName
+// replaceOldWorkloadSlice finishes the old slice after the new slice has been
+// admitted. Called inside the admit success path so the old slice is only
+// finished when the new one is confirmed. If this fails, the job reconciler's
+// EnsureWorkloadSlices detects both slices admitted and finishes the old one.
+func (s *Scheduler) replaceOldWorkloadSlice(ctx context.Context, log logr.Logger, e *entry, oldWorkloadSlice *preemption.Target) {
 	if err := s.replaceWorkloadSlice(ctx, oldWorkloadSlice.WorkloadInfo.ClusterQueue, e.Obj, oldWorkloadSlice.WorkloadInfo.Obj.DeepCopy()); err != nil {
-		log.Error(err, "Failed to replace workload slice")
-		return err
+		log.Error(err, "Failed to finish old workload slice after admitting replacement; job reconciler will handle recovery")
```
mimowo, line 539: I think this is ok, but IIUC if this fails, then we may observe double counting of the quota for the brief moment - before the JobReconciler finishes the old slices. 

The quota bump may result in temporary excessive preemptions of other workloads, or blocking workloads that could be admitted otherwise.

So, I think ideally if the scheduler cache was aware of the fact that both slices are admitted at the same time, and only count capacity from the "max" of them, rather than "sum".

However, that quota bump is very rare (requires request failure), and the consequence is also limited (excessive preemptions for a brief moment), so my comment is mostly to confirm my understanding rather than to request changes.

We may re-evaluate if this is good enough or not before GA graduation.

sohankunkerkar: Yup, that's right. The double-count only happens if `replaceOldWorkloadSlice` fails after admission succeeds i.e. the old slice stays admitted in the cache alongside the new one until the job reconciler's `EnsureWorkloadSlices` finishes the old slice on its next reconcile. This is conservative (over-reports usage, blocks other admits temporarily) and resolves within one reconcile cycle. I agree it's fine for now, and the cache-level max-not-sum optimization is a good candidate for GA if it shows up in practice.

pkg/scheduler/scheduler.go
```diff
@@ -788,7 +783,7 @@ func updateAssignmentForTAS(log logr.Logger, snapshot *schdcache.Snapshot, cq *s
 // the entry, and asynchronously updates the object in the apiserver after
 // assuming it in the cache.
 // Note: this does not necessarily make the workload "admitted".
-func (s *Scheduler) admit(ctx context.Context, e *entry, cq *schdcache.ClusterQueueSnapshot) error {
+func (s *Scheduler) admit(ctx context.Context, e *entry, cq *schdcache.ClusterQueueSnapshot, oldWorkloadSlice *preemption.Target) error {
```
ichekrygin, line 786: Nit: I understand why `replaceOldWorkloadSlice` needs to happen only after successful admission, but is there a reason it has to be inside `admit` rather than immediately after a successful `admit` call?

For example, we could keep `admit` generic and do:

```go
if err := s.admit(ctx, e, cq); err != nil {
    e.inadmissibleMsg = fmt.Sprintf("Failed to admit workload: %v", err)
    return
}

if features.Enabled(features.ElasticJobsViaWorkloadSlices) && oldWorkloadSlice != nil {
    s.replaceOldWorkloadSlice(ctx, log, e, oldWorkloadSlice)
}
```

That would preserve the important ordering, finish the old slice only after the replacement was successfully admitted, while avoiding exposing `admit` to the workload-slice replacement concept. Most workloads are not elastic slice replacements, so passing `oldWorkloadSlice` through `admit` feels a bit leaky to me.

pkg/scheduler/scheduler.go
```diff
@@ -531,18 +530,14 @@ func (s *Scheduler) waitForPodsReadyIfBlocked(ctx context.Context, log logr.Logg
...
-// placement stays consistent across the replacement. If the subsequent admit
-// fails, the old slice may end up Finished while the new one is not admitted;
-// downstream controllers handle suspension/eviction.
-func (s *Scheduler) replaceOldWorkloadSlice(ctx context.Context, log logr.Logger, e *entry, oldWorkloadSlice *preemption.Target) error {
-	e.Obj.Status.ClusterName = oldWorkloadSlice.WorkloadInfo.Obj.Status.ClusterName
+// replaceOldWorkloadSlice finishes the old slice after the new slice has been
+// admitted. Called inside the admit success path so the old slice is only
+// finished when the new one is confirmed. If this fails, the job reconciler's
+// EnsureWorkloadSlices detects both slices admitted and finishes the old one.
+func (s *Scheduler) replaceOldWorkloadSlice(ctx context.Context, log logr.Logger, e *entry, oldWorkloadSlice *preemption.Target) {
```
ichekrygin, line 537: Nit: After the refactoring, `replaceOldWorkloadSlice` no longer appears to have meaningful control-flow value beyond local error reporting.

Would it make sense to make it fire-and-forget and move the error logging into the `processEntry` call site instead?

Something along the lines of:

```go
if err := s.admit(ctx, e, cq); err != nil {
    e.inadmissibleMsg = fmt.Sprintf("Failed to admit workload: %v", err)
    return
}

if features.Enabled(features.ElasticJobsViaWorkloadSlices) && oldWorkloadSlice != nil {
    if err := s.replaceOldWorkloadSlice(...); err != nil {
        log.Error(err, "Failed to finish old workload slice after admitting replacement")
    }
}
```

This would keep `admit` generic and avoid exposing it to workload-slice replacement semantics for the common non-elastic workload path.

ichekrygin (approved):
Changes are LGTM overall, aside from two small nits, nothing blocking from my side.

Thank you for updating this!

pkg/scheduler/scheduler.go
ichekrygin: We had discussed earlier that it could be helpful to "teach" the queue/cache how to exclude the unfinished old slice from quota accounting, so we don't observe an artificial quota increase in the case where updating the old slice fails.

It would be great if we could address that in this PR, but I also think it is reasonable as a follow-up.

In that case, since you now have all the context around this flow 🙂, would you mind creating a follow-up issue and linking it to this PR?

pkg/scheduler/scheduler.go
sohankunkerkar: I think admit() is async. The `admissionRoutineWrapper.Run()`  launches a goroutine and admit() returns nil immediately before `PatchAdmissionStatus` completes. If we finish the old slice after admit() returns in processEntry(), we'd be finishing it before admission is actually confirmed. It's the same race we're fixing. That's why it has to live inside the goroutine's success path.

pkg/scheduler/scheduler.go
sohankunkerkar: Same reason as above — since this runs inside the async goroutine, there's no synchronous call site to bubble the error up to. Logging inside the function is the only option here.

pkg/scheduler/scheduler.go
sohankunkerkar: How about I document this as a GA consideration in the ElasticJobsViaWorkloadSlices KEP instead? It's a pretty narrow failure path (API write failing right after a successful one) and @mimowo was also leaning toward revisiting at GA.

test/integration/singlecluster/controller/jobs/job/job_controller_test.go
```diff
@@ -4326,6 +4326,52 @@ var _ = ginkgo.Describe("Job with elastic jobs via workload-slices support", gin
...
+		ginkgo.By("creating an elastic job")
+		util.MustCreate(ctx, k8sClient, elasticJob)
+
+		ginkgo.By("the original workload slice is admitted")
+		workloads := util.ExpectWorkloadsInNamespace(ctx, k8sClient, elasticJob.Namespace, 1)
+		oldWorkloadSlice := &workloads[0]
+		util.ExpectWorkloadsToBeAdmitted(ctx, k8sClient, oldWorkloadSlice)
+		util.ExpectJobUnsuspendedWithNodeSelectors(ctx, k8sClient, client.ObjectKeyFromObject(elasticJob), nil)
+
+		ginkgo.By("scaling the job beyond the remaining quota")
```
yaroslava-serdiuk, line 4350: IFIUC If the newWorkload is beyond the remaining quota the newWorkload would not reserve the quota, and so Kueue won't try to admit the newWorkload and return in line 401. 
So I don't think the added changes are tested here.

pkg/scheduler/scheduler.go
yaroslava-serdiuk: +1 to updating the KEP, could you do it in this PR, please?

test/integration/singlecluster/controller/jobs/job/job_controller_test.go
sohankunkerkar: The scheduler test framework can only fail all workload patches at once, not just the new slice's, so we can't simulate a selective admission failure here. The scheduler fix is structural: `replaceOldWorkloadSlice` only fires after `PatchAdmissionStatus` succeeds. This integration test covers the other side, making sure the job reconciler doesn't finish the old slice while the replacement is still pending.

pkg/scheduler/scheduler.go
sohankunkerkar: Added here: https://github.com/kubernetes-sigs/kueue/pull/11242

mimowo:
/release-note-edit
```release-note
ElasticJobViaWorkloadSlices: Fix quota leak during elastic workload scale-up where old slice was finished before replacement slice was admitted.
```

mimowo:
/lgtm
/approve
I think this is a step in the right direction. I will also be good for ElasticJobs + TAS integration where we need to be sure we can copy the assignment from the "old" slice.

I reviewed and checked the discussions and I think we area good to merge as is.
/cherrypick release-0.17
/cherrypick release-0.16
Thank you 👍

test/integration/singlecluster/controller/jobs/job/job_controller_test.go
yaroslava-serdiuk: > This integration test covers the other side, making sure the job reconciler doesn't finish the old slice while the replacement is still pending.

I don't think it's actually tested here, because the replacement doesn't happening here since the new workload is beyond the quota. 
Additionally, the test passes when run without the changes.

mimowo: Thank you @yaroslava-serdiuk for spotting that. I think this test is wrong indeed. If the test is not testing the behavior change (passes before and after the change), then let's drop it. 

We may consider a test which explicitly tests the invariant that OldSlice it not finished when transitioning the new slice to Admitted. This can be done with the watch pattern.

mimowo: Opened the follow up: https://github.com/kubernetes-sigs/kueue/issues/11283

tenzen-y:
/release-note-edit
```release-note
ElasticJobsViaWorkloadSlices: Fix quota leak during elastic workload scale-up where old slice was finished before replacement slice was admitted.
```

PR #11199, opened 2026-05-14
Filter kueueviz local queue workloads by spec.queueName

### LINKED ISSUE #6613: Kueueviz local queue workloads is using invalid field label
**What happened**: 
Local Queue details page does not show workloads. The backend pod shows the following error: 
```
2025/08/18 21:47:41 ERROR Error fetching data %v error="error fetching workloads for local queue q-testing-np-2: field label not supported: spec.queueName"
```

The local queues page should show the list of admitted workloads.

Simplest way to reproduce the issue is to just use kubectl
```
kubectl get workloads.kueue.x-k8s.io -n <NAMESPACE> --field-selector spec.queueName=<QUEUE_NAME>
```
This would throw the same error as the queueName on workload CRD is not marked selectable. 

I did some basic digging and selectable fields can be used to index for fields on CRDs. Ref: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#crd-selectable-fields

Workload CRD might need to be updated. However this would only work for K8s 1.31 and above. 

Ideally if workloads can be filtered by labels, it would be more generic. 

Code link for local queue workload list logic:  https://github.com/kubernetes-sigs/kueue/blame/main/cmd/kueueviz/backend/handlers/local_queue_workloads.go#L40

**Environment**:
- Kubernetes version (use `kubectl version`): 1.32
- Kueue version (use `git describe --tags --dirty --always`): v0.13.1
- Cloud provider or hardware configuration: EKS
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

ManthanNimodiya:
## What type of PR is this?
/kind bug
/area dashboard

## What this PR does / why we need it:
The local queue details page in kueueviz was not filtering workloads by queue, it returned every workload in the namespace regardless of which local queue was requested. The root cause was that `fetchLocalQueueWorkloads` accepted a `queueName` parameter but never applied it when building the response list.

The fix adds a client-side filter on `spec.queueName` in the existing loop, consistent with how kueuectl list workload localqueue handles the same case.

## Which issue(s) this PR fixes:
Fixes #6613

## Special notes for your reviewer:
Using a field selector on `spec.queueName` was considered but that field is not registered as selectable in the Workload CRD, which is the exact error the issue reports. 
Client-side filtering avoids CRD `changesand` works across all supported Kubernetes versions.

```release-note
KueueViz: Fixed the LocalQueue details page to show only workloads from the selected queue.
```

### DIFF
--- a/cmd/kueueviz/backend/handlers/local_queue_workloads.go
+++ b/cmd/kueueviz/backend/handlers/local_queue_workloads.go
@@ -46,7 +46,9 @@ func (h *Handlers) fetchLocalQueueWorkloads(ctx context.Context, namespace, queu
 
 	var workloads []any
 	for _, item := range wql.Items {
-		workloads = append(workloads, item)
+		if string(item.Spec.QueueName) == queueName {
+			workloads = append(workloads, item)
+		}
 	}
 	return workloads, nil
 }
--- a/cmd/kueueviz/backend/handlers/local_queue_workloads_test.go
+++ b/cmd/kueueviz/backend/handlers/local_queue_workloads_test.go
@@ -108,7 +108,7 @@ func TestFetchLocalQueueWorkloads(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 			h := &Handlers{client: &fakeWorkloadClient{workloads: tc.workloads, err: tc.listErr}}
 
-			got, err := h.fetchLocalQueueWorkloads(context.Background(), tc.namespace, tc.queueName)
+			got, err := h.fetchLocalQueueWorkloads(t.Context(), tc.namespace, tc.queueName)
 			if tc.wantErr {
 				if err == nil {
 					t.Fatal("expected error, got nil")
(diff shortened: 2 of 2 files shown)


mimowo:
/ok-to-test
cc @samzong

cmd/kueueviz/backend/handlers/local_queue_workloads.go
```diff
@@ -46,7 +46,9 @@ func (h *Handlers) fetchLocalQueueWorkloads(ctx context.Context, namespace, queu
 
 	var workloads []any
 	for _, item := range wql.Items {
-		workloads = append(workloads, item)
+		if string(item.Spec.QueueName) == queueName {
```
mimowo, line 49: Let's try to add some unit or e2e test for the change

ManthanNimodiya: sure, currently working on it

ManthanNimodiya:
@mimowo - added unit tests, let me know if anything else is needed

kannon92 (commented):
/lgtm

mimowo:
Thank you 👍 
/approve
/cherrypick release-0.17
/cherrypick release-0.16

mimowo:
/release-note-edit
```release-note
KueueViz: Fixed the LocalQueue details page to show only workloads from the selected queue.
```

PR #10990, opened 2026-05-06
Add reconnect backoff guardrail to suppress redundant cluster reconciles

reruno:
#### What type of PR is this?
/kind bug

#### What this PR does / why we need it:
Add reconnect backoff guardrail to suppress redundant cluster reconciles
#### Which issue(s) this PR fixes:

Discovered when working on #10926 
Related to #10950 

#### Does this PR introduce a user-facing change?

```release-note
MultiKueue: Add reconnect backoff guardrail to suppress redundant cluster reconciles for the MultiKueueCluster reconciler.
```

Issue #9145, opened 2026-02-12
Introduce a test to reproduce conversion webhooks on 0.15


mimowo:
**What would you like to be added**:

An automated test which would reproduce the issues with conversion webhooks at scale encountered in 0.15

See the scenario: https://github.com/kubernetes/kubernetes/issues/136950#issuecomment-3889610473

**Why is this needed**:

To help to guide the graduation of the ConcurrentWatchObjectDecode feature gate in k8s core: https://github.com/kubernetes/kubernetes/issues/136950

mimowo:
@tenzen-y @gabesaba @IrvingMg cc @mbobrovskyi

Jefftree:
Hey @mimowo @mykysha @mbobrovskyi, just want to ask about the current status of this? I'm evaluating https://github.com/kubernetes/kubernetes/issues/136950 on the k/k side and would love to learn about your testing scenario.

mykysha:
Hi @Jefftree! I am currently working on this, have managed to reproduce the issue locally using the script that mostly follows the scenario described in the commend linked above, and have verified that enabling ConcurrentWatchObjectDecode and re-running the script resolves the problem. The main obstacle was not the script itself, but finding the right variables for kueue and workload configurations.

Currently working on focusing the script on the conversion webhook only and cleaning the script up, after which will be creating a PR. Will also be making a scenario for a cloud cluster setup.

PR #11207, opened 2026-05-15
MultiKueue: prevent a hung remote watch from stopping all-cluster admission

### LINKED ISSUE #11206: [MultiKueue] One hung remote watch blocks dispatch to every other cluster
## What happens

When one MultiKueue remote cluster's watch hangs, all other clusters become unusable for dispatch too — not just the affected one.

The `multikueuecluster` controller has one worker goroutine. It sets up each remote serially, opening watches with `client.Watch()` and not marking the cluster ready until all watches return. If one cluster's `client.Watch()` hangs (for example, when the remote apiserver is behind a proxy that buffers empty chunked responses), the worker is stuck on that cluster and never moves on to the rest. The remaining clusters stay in the "still connecting" state forever; the dispatcher then excludes them from the active set and admission for every queue stops, even for queues that point at completely different clusters.

## How we hit it

kueue v0.13.3 with one remote behind Cloudflare Tunnel. Cloudflare buffers chunked HTTP responses with no payload, so an empty watch (label selector matched zero objects) never streams a single event back. `client.Watch()` never returns. The reconciler's worker is stuck. Other clusters never get their watches set up, and admission cluster-wide stops.

Roughly 90 minutes of zero-admission across all our queues before we reverted the offending change.

## Proposed fix

Add a 60s bound on the `client.Watch()` *establishment* phase only. On timeout, propagate an error to the existing `failedConnAttempts` / `retryAfter()` backoff. The successful-watch stream lifetime is unchanged.

PR: #11207.

This is defense in depth on top of #9968 (which fixed the specific Cloudflare-empty-watch trigger by enabling watch bookmarks). Any other future condition that hangs `client.Watch()` would hit the same head-of-line blocking.

Adjacent: #10428.

trilamsr:
#### What type of PR is this?

/kind bug
/area multikueue

A hung `client.Watch()` against one remote MultiKueueCluster could block the single `multikueuecluster` reconciler worker indefinitely, preventing every other cluster behind it from being reconciled. Those clusters keep `remoteClient.connecting=true`, the dispatcher excludes them, and admission stops cluster-wide.

This bounds the Watch establishment phase with `watchEstablishTimeout` (default 60s, package-level var overridable in tests). On timeout the in-flight Watch is canceled and `errWatchEstablishTimeout` is returned, falling back to the existing `failedConnAttempts` / `retryAfter` backoff. The successful-watch stream lifetime is unchanged — the returned watcher continues to use a context derived from the caller's `ctx`, and its cancel is owned by the watcher's `Stop()` method via a small wrapper (no goroutine leak, no lostcancel warning).

Follow-up to #9968, which fixed the most common trigger (Cloudflare Tunnel buffering empty chunked watch responses) by enabling `AllowWatchBookmarks`. This change defends against any future condition that hangs `client.Watch()` similarly.

Fixes #11206

- Helper extracted as `watchWithEstablishTimeout` for direct unit testing.
- New unit test `TestWatchWithEstablishTimeout` covers three paths: hung Watch (timeout), Watch error (propagated immediately), Watch success (no timeout wait).
- Tested locally with `go test ./pkg/controller/admissionchecks/multikueue/` — all package tests pass.

```release-note
NONE
```

### DIFF
--- a/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
+++ b/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
@@ -72,6 +72,15 @@ const (
 	// this set will provide waiting time between 0 to 5m20s
 	retryIncrement = 5 * time.Second
 	retryMaxSteps  = 7
+
+	// Bounds how long startWatcher waits for client.Watch() to return,
+	// so a hung remote cannot head-of-line block the reconciler. See #11206.
+	defaultWatchEstablishTimeout = 60 * time.Second
+)
+
+var (
+	watchEstablishTimeout    = defaultWatchEstablishTimeout
+	errWatchEstablishTimeout = errors.New("watch establishment timed out")
 )
 
 // retryAfter returns an exponentially increasing interval between
@@ -226,12 +235,56 @@ func (rc *remoteClient) setConfig(watchCtx context.Context, config *clientConfig
 	return nil, nil
 }
 
+// cancelOnStopWatcher carries the establishment context's cancel func so it
+// runs when Stop is called, satisfying govet's lostcancel check without
+// shortening the watch's stream lifetime on the success path.
+type cancelOnStopWatcher struct {
+	watch.Interface
+	cancel context.CancelFunc
+}
+
+func (cw *cancelOnStopWatcher) Stop() {
+	cw.Interface.Stop()
+	cw.cancel()
+}
+
+// establishWatch opens a MultiKueue remote watch, bounded by
+// watchEstablishTimeout. On timeout the in-flight Watch is canceled and
+// errWatchEstablishTimeout is returned so the caller falls back to the
+// standard failedConnAttempts / retryAfter backoff in setConfig.
+func establishWatch(ctx context.Context, c client.WithWatch, obj client.ObjectList, origin string) (watch.Interface, error) {
+	type result struct {
+		w   watch.Interface
+		err error
+	}
+	resultCh := make(chan result, 1)
+	establishCtx, cancel := context.WithCancel(ctx)
+
+	go func() {
+		w, err := c.Watch(establishCtx, obj,
+			client.MatchingLabels{kueue.MultiKueueOriginLabel: origin},
+			&client.ListOptions{Raw: &metav1.ListOptions{AllowWatchBookmarks: true}},
+		)
+		resultCh <- result{w: w, err: err}
+	}()
+
+	select {
+	case r := <-resultCh:
+		if r.err != nil {
+			cancel()
+			return nil, r.err
+		}
+		return &cancelOnStopWatcher{Interface: r.w, cancel: cancel}, nil
+	case <-time.After(watchEstablishTimeout):
+		cancel()
+		<-resultCh
+		return nil, errWatchEstablishTimeout
+	}
+}
+
 func (rc *remoteClient) startWatcher(ctx context.Context, kind string, w jobframework.MultiKueueWatcher) error {
 	log := ctrl.LoggerFrom(ctx).WithValues("watchKind", kind)
-	newWatcher, err := rc.client.Watch(ctx, w.GetEmptyList(),
-		client.MatchingLabels{kueue.MultiKueueOriginLabel: rc.origin},
-		&client.ListOptions{Raw: &metav1.ListOptions{AllowWatchBookmarks: true}},
-	)
+	newWatcher, err := establishWatch(ctx, rc.client, w.GetEmptyList(), rc.origin)
 	if err != nil {
 		return err
 	}
--- a/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
+++ b/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
@@ -75,7 +75,20 @@ const (
 
 	// Bounds how long startWatcher waits for client.Watch() to return,
 	// so a hung remote cannot head-of-line block the reconciler. See #11206.
-	defaultWatchEstablishTimeout = 60 * time.Second
+	//
+	// The bound has to cover the slowest legitimate case, not just the common
+	// one. client.Watch() returns as soon as the apiserver sends HTTP 200
+	// headers, but the apiserver can delay sending those headers while it
+	// warms its watch cache. When the served version differs from the storage
+	// version and a conversion webhook is in play, that warmup serializes a
+	// per-object conversion and has been observed to take ~8 min at ~50k
+	// Workloads (kubernetes/kubernetes#136950). 10 min leaves headroom above
+	// the documented worst case while still ringing the alarm on a remote
+	// that is truly hung. Today Kueue's served and storage versions match
+	// (v1beta2), so the warmup path is not exercised, but we keep the
+	// generous bound as a guard against future version skew and unknown
+	// apiserver behavior.
+	defaultWatchEstablishTimeout = 10 * time.Minute
 )
 
 var (
--- a/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
+++ b/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
@@ -277,7 +277,9 @@ func establishWatch(ctx context.Context, c client.WithWatch, obj client.ObjectLi
 		return &cancelOnStopWatcher{Interface: r.w, cancel: cancel}, nil
 	case <-time.After(watchEstablishTimeout):
 		cancel()
-		<-resultCh
+		if r := <-resultCh; r.w != nil {
+			r.w.Stop()
+		}
 		return nil, errWatchEstablishTimeout
 	}
 }
--- a/pkg/controller/admissionchecks/multikueue/multikueuecluster_test.go
+++ b/pkg/controller/admissionchecks/multikueue/multikueuecluster_test.go
@@ -1160,3 +1160,68 @@ func TestClustersReconcilerEventFilters(t *testing.T) {
 		})
 	}
 }
+
+func TestEstablishWatch(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pkg/controller/admissionchecks/multikueue/multikueuecluster_test.go
```diff
@@ -925,3 +925,68 @@ func TestValidateKubeconfig(t *testing.T) {
 		})
 	}
 }
+
+func TestEstablishWatch(t *testing.T) {
```
kshalot, line 1164: Nit: Usually in the codebase unit tests are table-driven.

In this case maybe it would look something like this:
```go
cases := map[string]struct {
    interceptor interceptor.Funcs
    wantErr     error
}{ ... }
```

kshalot (commented):
/lgtm

I'm thinking about e2e testing this, although it might be tricky to simulate in an easy way (I was thinking of maybe using `DROP` in `iptables`).

trilamsr:
Thanks @kshalot!
- Inline nit: `TestEstablishWatch` is now table-driven.
- e2e via `iptables -j DROP`: interesting idea, that does seem like the cleanest way to simulate an unresponsive remote apiserver

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
```diff
@@ -70,6 +70,15 @@ const (
 	// this set will provide waiting time between 0 to 5m20s
 	retryIncrement = 5 * time.Second
 	retryMaxSteps  = 7
+
+	// Bounds how long startWatcher waits for client.Watch() to return,
+	// so a hung remote cannot head-of-line block the reconciler. See #11206.
```
mimowo, line 77: Exploratory question to confirm my understanding: could this issue be also mitigated by increasing the `GroupKindConcurrency` level for MultiKueueCluster for the controller, wdyt? Code pointer to the configuration: https://github.com/kubernetes-sigs/kueue/blob/main/apis/config/v1beta2/configuration_types.go#L264C2-L264C22
In that case when one Reconcile is stuck the other may continue

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
```diff
@@ -70,6 +70,15 @@ const (
...
+	defaultWatchEstablishTimeout = 60 * time.Second
```
mimowo: I'm wondering if the timeout at 60s is safe in practice. IIUC this watch is used for all types, that includes Workloads, which might be numerous. And when Watch is opened IIUC all workloads are Listed, which may take a lot of time at the api-server side to prepare the response, in particular if the Workload objects need to go via conversion webhooks.

Recently we had an issue in Kueue (single cluster) that at scale of 50k workloads when conversion webhooks are running it may take around 8min, whilst the etcd compaction timeout is only around 2.5min. The test we are working on here is to prevent that in the future: https://github.com/kubernetes-sigs/kueue/issues/9145

Also, it should not be that long when conversion webhooks are not necessary.

If my suspection is correct then maybe we should:
- use increased timeout for Workload objects (a bit hacky but maybe works well in practice (say 10min)
- use exponential timeout which is increased on retries (we already have the retry mechanism recently improved here: https://github.com/kubernetes-sigs/kueue/pull/10990

kshalot: OTOH, in this particular case, will it actually list all the workloads? I think here it will just make na HTTP call, get the headers and then the events will start streaming (IIUC it will stream an ADDED event for every resource because we are not passing any resource version in the options).

AFAIR the 50k workloads issue happened because the Informer first does a List and then a Watch.

mimowo: > OTOH, in this particular case, will it actually list all the workloads? I think here it will just make na HTTP call, get the headers and then the events will start streaming (IIUC it will stream an ADDED event for every resource because we are not passing any resource version in the options).

Possible, I'm not sure about the exact semantics, if this is the case then we are good. 

Let's research that more and support the claim  (that opening the watch does not entail blocking on listing the workloads) with references or experiments.

trilamsr: > Let's research that more…

Did some digging. TLDR: @kshalot is right on the client side, @mimowo is right on the server side, and the server side is the one that bites.

**Client side**: traced controller-runtime → client-go → `newStreamWatcher`. `c.Watch()` returns as soon as HTTP 200 headers arrive; there's no client-side `List`, no full-body buffering. So the 50k-workloads Informer scenario doesn't transfer directly. (Happy to drop in file:line refs if useful.)

**Server side**: `c.Watch()` still blocks until the apiserver *sends* those headers — and that's exactly the case @mimowo flagged. From [kubernetes/kubernetes#136950](https://github.com/kubernetes/kubernetes/issues/136950):

> If the API server does not have a "warm" cache […] before returning to Kueue it tries to convert them sequentially to v1beta2 […] estimated to take around 8min.

With 60s we'd cancel + retry on every cycle and never get past warmup, arguably worse than the pre-PR "wait forever, eventually succeed" for that path.

**Proposal**: bump `defaultWatchEstablishTimeout` to **10 min** in this PR (matching @mimowo's "say 10min"). Tiny diff, preserves the hung-remote fix, avoids the cold-start regression.

For the more refined version — exponential timeout on retries that stacks with the `failedConnAttempts` / `retryAfter` machinery from #10990. I'll open a follow-up so this PR stays focused. WDYT?

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
trilamsr: > could this issue be also mitigated by increasing the `GroupKindConcurrency` level…

I don't think it actually buys us anything in this code path. The knob is wired through correctly (the controller doesn't override `MaxConcurrentReconciles`), so in principle you'd get N parallel workers. But Reconcile calls `setRemoteClientConfig`, which grabs `clustersReconciler.lock` (a single controller wide write mutex at `multikueuecluster.go:394`) and holds it through the entire `setConfig` → `establishWatch` → `c.Watch()` chain. So when one worker is parked inside `c.Watch()`, every other Reconcile that wants to do anything with a different cluster sits waiting on that same lock.

In practice concurrency=N collapses back to 1 whenever any remote is hung. Bumping the knob would only help after we also split that lock per cluster, or released it before establishing the watch. Probably a separate PR. Happy to file a follow up issue for it if you want.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
mimowo: @trilamsr thank you for all the digging and super precise descriptions. We need to capture that with comments. 

> Proposal: bump defaultWatchEstablishTimeout to 10 min in this PR (matching @mimowo's "say 10min"). Tiny diff, preserves the hung-remote fix, avoids the cold-start regression.

I think 10min should be ok, espectially since we don't have now the discrepancy between the storage and serving versions, which bite us in 0.15. We may need to introduce the discrepancy when transitioning to v1beta3 one day, but at that point we will probably need to fix larger problems with https://github.com/kubernetes/kubernetes/issues/136950

trilamsr: Done in 9bd2d79e8. Bumped `defaultWatchEstablishTimeout` to `10 * time.Minute` and expanded the doc block above it to capture the cold-cache + conversion-webhook reasoning and the link to kubernetes/kubernetes#136950. PTAL.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
mimowo: I see, thank you for the summary, I now understand why this would not work currently, and why fixing the problem is not straighforward.

Having said that, I think this renders `GroupKindConcurrency` useless for `MultiKueueCluster` due to very technical reasons which can be considered a bug by a user.

So, please file an issue for that. Feel free to work on it, or just keep it posted for some contributors.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
mimowo: sgtm

mimowo:
/release-note-edit
```release-note
MultiKueue: bound the establishment phase of remote watches with a 10min timeout so a hung `client.Watch()` against one remote cannot head-of-line block the multikueuecluster reconciler worker and stop admission cluster-wide.
```

mimowo:
@trilamsr release note proposal:

/release-note-edit
```release-note
MultiKueue: Fixed a bug where a hung watch connection to one remote cluster could block
reconciliation of other MultiKueueClusters, leaving them inactive and preventing workload
admission. Kueue now applies a 10-minute circuit-breaking timeout while establishing
remote-cluster watches, allowing reconciliation to recover instead of blocking indefinitely.
```

mimowo:
Thank you for fixing the issue upstream (in Kueue), nice contribution 👍 
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16

mimowo:
Please also work on CPs in case the robot fails due to conflicts

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
trilamsr: Filed #11297 to track the lock refactor. Leaving it unassigned for now so it's open for any contributor to pick up.

mimowo:
Clearing the release note as we will have a combined release note in the follow up PR: https://github.com/kubernetes-sigs/kueue/pull/11304
/release-note-edit
```release-note
NONE
```

PR #10657, opened 2026-04-21
DO NOT MERGE: Reproduce the TAS over-subscription bugs

tenzen-y:
#### What type of PR is this?

/kind bug
/area tas

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #11210, opened 2026-05-15
TAS: Stop implicitly assuming ResrouceFlavor:Node = 1:N

### LINKED ISSUE #10659: TAS: Kueue over-subscribes Node capacities when a Node belongs to multiple ResourceFlavors
**What happened**:

When a Node belongs to the multiple ResourceFlavors like the following, Kueue TAS over-subscribes the Node capacities for Workloads while the Quota Reservation is correct.

```yaml
# This Node belongs to the following both ResourceFlavors.
kind: Node
metadata:
  labels:
    example.com/machine: standard
spec:
  taints:
  - effects: NoSchedule
    key: example.com/machine
    value: standard
  # The Node doesn't have example.com/instance-type taints.
  ...
---
kind: ResourceFlavor
...
spec:
  topologyName: flat
  nodeLabels:
    example.com/machine: standard
  nodeTaints:
  - effect: NoSchedule
    key: example.com/machine
    value: standard
  - effect: NoSchedule
    key: example.com/instance-type
    value: partial-reserved
---
kind: ResourceFlavor
...
spec:
  topologyName: flat
  nodeLabels:
    example.com/machine: standard
  nodeTaints:
  - effect: NoSchedule
    key: example.com/machine
    value: standard
```

There is not over subscription in TAS.

I reproduced this problem in the following scheduler UT case.
In this scenario, the `partial-reserved-pending` wl should get Node x2 Topology assignment because Node x1 has already been occupied by the `ondemand-admitted-a` and `ondemand-admitted-b` wls.

However, the current `partial-reserved-pending` wl topology assignment will be Node x2 (oversubscription).

https://github.com/kubernetes-sigs/kueue/pull/10657
https://github.com/tenzen-y/kueue/blob/6523d77d8534932e91255d25c7333478103d3254/pkg/scheduler/scheduler_tas_test.go#L2391-L2569

Again, the Quota Reservation is correct.

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

tenzen-y:
#### What type of PR is this?

/kind bug
/area tas

I fixed a TAS over-subscription bug where multiple `ResourceFlavor` referencing the same Topology and overlapping on physical nodes each independently track `usage`, so sibling flavors believe a shared node is only partially used.

This PR refined the TAS cache so that sibling flavors aggregate usage.

Fixes https://github.com/kubernetes-sigs/kueue/issues/10659

https://github.com/kubernetes-sigs/kueue/pull/10657 is problem reproduced PR.

```release-note
TAS: fix over-subscription of nodes that belong to multiple ResourceFlavors sharing the same hostname-leaf Topology.
```

### DIFF
--- a/pkg/cache/scheduler/clusterqueue_snapshot.go
+++ b/pkg/cache/scheduler/clusterqueue_snapshot.go
@@ -19,6 +19,7 @@ package scheduler
 import (
 	"iter"
 	"maps"
+	"slices"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -198,13 +199,23 @@ func (c *ClusterQueueSnapshot) FindTopologyAssignmentsForWorkload(
 		option(opts)
 	}
 
+	var aggregatedDomainUsages map[utiltas.TopologyDomainID]resources.Requests
+	if features.Enabled(features.TASHandleOverlappingFlavors) {
+		aggregatedDomainUsages = make(map[utiltas.TopologyDomainID]resources.Requests)
+	}
+
 	result := make(TASAssignmentsResult)
-	for tasFlavor, flavorTASRequests := range tasRequestsByFlavor {
+	for _, tasFlavor := range slices.Sorted(maps.Keys(tasRequestsByFlavor)) {
+		flavorTASRequests := tasRequestsByFlavor[tasFlavor]
 		// We assume the `tasFlavor` is already in the snapshot as this was
 		// already checked earlier during flavor assignment, and the set of
 		// flavors is immutable in snapshot.
 		tasFlavorCache := c.TASFlavors[tasFlavor]
-		flvResult := tasFlavorCache.FindTopologyAssignmentsForFlavor(flavorTASRequests, options...)
+		flvOpts := options
+		if features.Enabled(features.TASHandleOverlappingFlavors) && tasFlavorCache.isLowestLevelNode {
+			flvOpts = append(slices.Clone(options), WithAggregatedDomainUsages(aggregatedDomainUsages))
+		}
+		flvResult := tasFlavorCache.FindTopologyAssignmentsForFlavor(flavorTASRequests, flvOpts...)
 		maps.Copy(result, flvResult)
 	}
 	return result
--- a/pkg/cache/scheduler/snapshot.go
+++ b/pkg/cache/scheduler/snapshot.go
@@ -33,8 +33,10 @@ import (
 	"sigs.k8s.io/kueue/pkg/cache/hierarchy"
 	queueafs "sigs.k8s.io/kueue/pkg/cache/queue/afs"
 	"sigs.k8s.io/kueue/pkg/features"
+	"sigs.k8s.io/kueue/pkg/resources"
 	afs "sigs.k8s.io/kueue/pkg/util/admissionfairsharing"
 	utilmaps "sigs.k8s.io/kueue/pkg/util/maps"
+	utiltas "sigs.k8s.io/kueue/pkg/util/tas"
 	"sigs.k8s.io/kueue/pkg/workload"
 )
 
@@ -192,8 +194,27 @@ func (c *Cache) Snapshot(ctx context.Context, options ...SnapshotOption) (*Snaps
 	}
 	tasSnapshots := make(map[kueue.ResourceFlavorReference]*TASFlavorSnapshot)
 	if features.Enabled(features.TopologyAwareScheduling) {
-		for flavor, cache := range c.tasCache.Clone() {
-			tasSnapshots[flavor] = cache.snapshot(log, c.tasCache.nodesCache.find(cache.flavor.NodeLabels, cache.topology.Levels))
+		var aggregatedDomainUsages map[utiltas.TopologyDomainID]resources.Requests
+		flvTASCache := c.tasCache.Clone()
+
+		if features.Enabled(features.TASHandleOverlappingFlavors) {
+			aggregatedDomainUsages = make(map[utiltas.TopologyDomainID]resources.Requests)
+			for _, cache := range flvTASCache {
+				c.snapshotTopologyDomainUsages(cache, aggregatedDomainUsages)
+			}
+		}
+		for flavor, cache := range flvTASCache {
+			// Only when this flavor is aggregation targets,
+			// we should propagate aggregated domain usages to snapshot constructions.
+			var aggregatedDomainUsagesForFlavor map[utiltas.TopologyDomainID]resources.Requests
+			if features.Enabled(features.TASHandleOverlappingFlavors) && utiltas.IsLowestLevelHostname(cache.topology.Levels) {
+				aggregatedDomainUsagesForFlavor = aggregatedDomainUsages
+			}
+			tasSnapshots[flavor] = cache.snapshot(
+				log,
+				c.tasCache.nodesCache.find(cache.flavor.NodeLabels, cache.topology.Levels),
+				aggregatedDomainUsagesForFlavor,
+			)
 		}
 	}
 	for _, cq := range cqNames {
@@ -219,6 +240,24 @@ func (c *Cache) Snapshot(ctx context.Context, options ...SnapshotOption) (*Snaps
 	// Shallow copy is enough
 	maps.Copy(snap.ResourceFlavors, c.resourceFlavors)
 	return &snap, nil
+}
+
+func (c *Cache) snapshotTopologyDomainUsages(
+	tasFlvCache *TASFlavorCache, aggregatedDomainUsages map[utiltas.TopologyDomainID]resources.Requests,
+) {
+	tasFlvCache.RLock()
+	defer tasFlvCache.RUnlock()
+
+	if len(tasFlvCache.topology.Levels) == 0 || !utiltas.IsLowestLevelHostname(tasFlvCache.topology.Levels) {
+		return
+	}
+	for domainID, domainUsage := range tasFlvCache.usage {
+		if _, ok := aggregatedDomainUsages[domainID]; ok {
+			aggregatedDomainUsages[domainID].Add(domainUsage)
+		} else {
+			aggregatedDomainUsages[domainID] = domainUsage.Clone()
+		}
+	}
 }
 
 // skipInactiveCQReason reports why the CQ should not be considered for
--- a/pkg/cache/scheduler/snapshot.go
+++ b/pkg/cache/scheduler/snapshot.go
@@ -202,6 +202,7 @@ func (c *Cache) Snapshot(ctx context.Context, options ...SnapshotOption) (*Snaps
 			for _, cache := range flvTASCache {
 				c.snapshotTopologyDomainUsages(cache, aggregatedDomainUsages)
 			}
+			log.V(4).Info("Aggregated TAS usage across flavors")
 		}
 		for flavor, cache := range flvTASCache {
 			// Only when this flavor is aggregation targets,
--- a/pkg/cache/scheduler/tas_cache_test.go
+++ b/pkg/cache/scheduler/tas_cache_test.go
@@ -18,6 +18,7 @@ package scheduler
 
 import (
 	"fmt"
+	"maps"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
@@ -36,6 +37,7 @@ import (
 	utiltestingapi "sigs.k8s.io/kueue/pkg/util/testing/v1beta2"
 	testingnode "sigs.k8s.io/kueue/pkg/util/testingjobs/node"
 	testingpod "sigs.k8s.io/kueue/pkg/util/testingjobs/pod"
+	"sigs.k8s.io/kueue/pkg/workload"
 )
 
 // PodSetTestCase defines a test case for a single podset in the consolidated test.
@@ -382,12 +384,15 @@ func TestFindTopologyAssignments(t *testing.T) {
 	}
 
 	cases := map[string]struct {
-		featureGates map[featuregate.Feature]bool
-		nodes        []corev1.Node
-		pods         []corev1.Pod
-		levels       []string
-		nodeLabels   map[string]string
-		podSets      []PodSetTestCase
+		featureGates           map[featuregate.Feature]bool
+		nodes                  []corev1.Node
+		pods                   []corev1.Pod
+		levels                 []string
+		nodeLabels             map[string]string
+		aggregatedDomainUsages map[tas.TopologyDomainID]resources.Requests
+		priorFlavorUsage       []workload.TopologyDomainRequests
+		priorOwnUsage          []workload.TopologyDomainRequests
+		podSets                []PodSetTestCase
 	}{
 		"minimize the number of used racks before optimizing the number of nodes; BestFit": {
 			// Solution by optimizing the number of racks then nodes: [r3]: [x1,x6,x2,x4]
@@ -7119,6 +7124,121 @@ func TestFindTopologyAssignments(t *testing.T) {
 				},
 			},
 		},
+		"sibling-flavor usage on the shared hostname reduces remaining capacity": {
+			levels: []string{tasBlockLabel, corev1.LabelHostname},
+			// Two nodes only, so the discrimination is unambiguous.
+			// The x1's effective capacity is 0, so x2 is the only
+			// candidate that can fit the request.
+			nodes: []corev1.Node{
+				*testingnode.MakeNode("b1-x1").
+					Label(tasBlockLabel, "b1").Label(corev1.LabelHostname, "x1").
+					StatusAllocatable(corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("1"), corev1.ResourcePods: resource.MustParse("10")}).
+					Ready().Obj(),
+				*testingnode.MakeNode("b1-x2").
+					Label(tasBlockLabel, "b1").Label(corev1.LabelHostname, "x2").
+					StatusAllocatable(corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("1"), corev1.ResourcePods: resource.MustParse("10")}).
+					Ready().Obj(),
+			},
+			nodeLabels: map[string]string{},
+			priorFlavorUsage: []workload.TopologyDomainRequests{
+				{
+					Values:            []string{"x1"},
+					SinglePodRequests: resources.Requests{corev1.ResourceCPU: 1000},
+					Count:             1,
+				},
+			},
+			podSets: []PodSetTestCase{
+				{
+					podSetName: "main",
+					topologyRequest: &kueue.PodSetTopologyRequest{
+						Required: new(corev1.LabelHostname),
+					},
+					requests: resources.Requests{corev1.ResourceCPU: 1000},
+					count:    1,
+					wantAssignment: &tas.TopologyAssignment{
+						Levels:  []string{corev1.LabelHostname},
+						Domains: []tas.TopologyDomainAssignment{{Count: 1, Values: []string{"x2"}}},
+					},
+				},
+			},
+		},
+		"non-hostname-leaf topology: nil aggregatedDomainUsages preserves per-flavor behavior": {
+			// Topology lowest level is rack (not hostname). In the new model
+			// topologyInformation.Usage is nil for such topologies, so
+			// NewTASFlavorCache gives each flavor a private map. The harness
+			// leaves tc.aggregatedDomainUsages nil. The snapshot() call then takes the
+			// per-flavor branch and iterates c.usage (empty here).
+			levels:     []string{tasBlockLabel, tasRackLabel},
+			nodes:      defaultNodes,
+			nodeLabels: map[string]string{},
+			podSets: []PodSetTestCase{
+				{
+					podSetName: "main",
+					topologyRequest: &kueue.PodSetTopologyRequest{
+						Required: ptr.To(tasRackLabel),
+					},
+					requests: resources.Requests{corev1.ResourceCPU: 1000},
+					count:    1,
+					wantAssignment: &tas.TopologyAssignment{
+						Levels:  []string{tasBlockLabel, tasRackLabel},
+						Domains: []tas.TopologyDomainAssignment{{Count: 1, Values: []string{"b1", "r1"}}},
+					},
+				},
+			},
+		},
+		"non-hostname-leaf topology: prior c.usage must be applied even when aggregatedDomainUsages is non-nil empty": {
+			// Reproduces the integration regression: with TASHandleOverlappingFlavors on,
+			// Cache.Snapshot allocates aggregatedDomainUsages as a non-nil empty map and
+			// snapshotTopologyDomainUsages returns early for non-hostname-leaf flavors,
+			// leaving the map empty. snapshot() must still apply c.usage in this case.
+			//
+			//      b1            b2
+			//     /  \          /  \
+			//   r1    r2      r1    r2
+			//   |     |       |     |
+			//  b1-r1 b1-r2   b2-r1 b2-r2   (each 1 CPU)
+			//
+			// b1/r1 and b1/r2 are fully consumed by prior workloads of this flavor.
+			// A 1-CPU, required-rack request must land on b2.
+			levels: []string{tasBlockLabel, tasRackLabel},
+			nodes: []corev1.Node{
+				*testingnode.MakeNode("b1-r1").
+					Label(tasBlockLabel, "b1").Label(tasRackLabel, "r1").
+					StatusAllocatable(corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("1"), corev1.ResourcePods: resource.MustParse("10")}).
+					Ready().Obj(),
+				*testingnode.MakeNode("b1-r2").
+					Label(tasBlockLabel, "b1").Label(tasRackLabel, "r2").
+					StatusAllocatable(corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("1"), corev1.ResourcePods: resource.MustParse("10")}).
+					Ready().Obj(),
+				*testingnode.MakeNode("b2-r1").
+					Label(tasBlockLabel, "b2").Label(tasRackLabel, "r1").
+					StatusAllocatable(corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("1"), corev1.ResourcePods: resource.MustParse("10")}).
+					Ready().Obj(),
+				*testingnode.MakeNode("b2-r2").
+					Label(tasBlockLabel, "b2").Label(tasRackLabel, "r2").
+					StatusAllocatable(corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("1"), corev1.ResourcePods: resource.MustParse("10")}).
+					Ready().Obj(),
+			},
+			nodeLabels: map[string]string{},
+			priorOwnUsage: []workload.TopologyDomainRequests{
+				{Values: []string{"b1", "r1"}, SinglePodRequests: resources.Requests{corev1.ResourceCPU: 1000}, Count: 1},
+				{Values: []string{"b1", "r2"}, SinglePodRequests: resources.Requests{corev1.ResourceCPU: 1000}, Count: 1},
+			},
+			podSets: []PodSetTestCase{
+				{
+					podSetName: "main",
+					topologyRequest: &kueue.PodSetTopologyRequest{
+						Required: new(tasRackLabel),
+					},
+					requests: resources.Requests{corev1.ResourceCPU: 1000},
+					count:    1,
+					wantAssignment: &tas.TopologyAssignment{
+						Levels:  []string{tasBlockLabel, tasRackLabel},
+						Domains: []tas.TopologyDomainAssignment{{Count: 1, Values: []string{"b2", "r1"}}},
+					},
+				},
+			},
+		},
 	}
 	for name, tc := range cases {
 		t.Run(name, func(t *testing.T) {
@@ -7153,7 +7273,29 @@ func TestFindTopologyAssignments(t *testing.T) {
 				tasCache.Update(&pod, log)
 			}
 			tasFlavorCache := tasCache.NewTASFlavorCache(topologyInformation, flavorInformation)
-			snapshot := tasFlavorCache.snapshot(log, tasCache.nodesCache.find(tasFlavorCache.flavor.NodeLabels, tasFlavorCache.topology.Levels))
+			if len(tc.priorOwnUsage) > 0 {
+				tasFlavorCache.addUsage("prior-wl", tc.priorOwnUsage)
+			}
+
+			if features.Enabled(features.TASHandleOverlappingFlavors) {
+				tc.aggregatedDomainUsages = aggregatedDomainUsagesForPriorFlavorUsage(
+					topologyInformation,
+					flavorInformation,
+					tc.priorFlavorUsage,
+					&tasCache,
+					tc.aggregatedDomainUsages,
+				)
+			}
+
+			var aggregatedDomainUsage map[tas.TopologyDomainID]resources.Requests
+			if features.Enabled(features.TASHandleOverlappingFlavors) && tas.IsLowestLevelHostname(tasFlavorCache.topology.Levels) {
+				aggregatedDomainUsage = tc.aggregatedDomainUsages
+			}
+			snapshot := tasFlavorCache.snapshot(
+				log,
+				tasCache.nodesCache.find(tasFlavorCache.flavor.NodeLabels, tasFlavorCache.topology.Levels),
+				aggregatedDomainUsage,
+			)
 			flavorTASRequests := make([]TASPodSetRequests, 0, len(tc.podSets))
 			wantResult := make(TASAssignmentsResult)
 			for _, ps := range tc.podSets {
@@ -7213,16 +7355,19 @@ func TestFindTopologyAssignmentsMultiLayerReplacement(t *testing.T) {
 	podSetName := kueue.PodSetReference("main")
 
 	cases := map[string]struct {
-		levels          []string
-		nodes           []corev1.Node
-		pods            []corev1.Pod
-		existingTA      *kueue.TopologyAssignment
-		admissionCount  int32
-		unhealthyNode   string
-		topologyRequest *kueue.PodSetTopologyRequest
-		count           int32
-		wantAssignment  *tas.TopologyAssignment
-		wantReason      string
+		featureGates           map[featuregate.Feature]bool
+		levels                 []string
+		nodes                  []corev1.Node
+		pods                   []corev1.Pod
+		existingTA             *kueue.TopologyAssignment
+		admissionCount         int32
+		unhealthyNode          string
+		topologyRequest        *kueue.PodSetTopologyRequest
+		count                  int32
+		aggregatedDomainUsages map[tas.TopologyDomainID]resources.Requests
+		priorFlavorUsage       []workload.TopologyDomainRequests
+		wantAssignment         *tas.TopologyAssignment
+		wantReason             string
 	}{
 		"replace unhealthy node in incomplete rack slice": {
 			//       b1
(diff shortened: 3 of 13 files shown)


tenzen-y:
/hold
for reviewers

mimowo:
cc @pajakd @PBundyra ptal

pkg/cache/scheduler/snapshot.go
```diff
@@ -185,8 +187,39 @@ func (c *Cache) Snapshot(ctx context.Context, options ...SnapshotOption) (*Snaps
 	}
 	tasSnapshots := make(map[kueue.ResourceFlavorReference]*TASFlavorSnapshot)
 	if features.Enabled(features.TopologyAwareScheduling) {
-		for flavor, cache := range c.tasCache.Clone() {
-			tasSnapshots[flavor] = cache.snapshot(log, c.tasCache.nodesCache.find(cache.flavor.NodeLabels, cache.topology.Levels))
+		flavorClone := c.tasCache.Clone()
```
pajakd: Can we move it closer to its first usage?

pkg/cache/scheduler/clusterqueue_snapshot.go
```diff
@@ -197,13 +198,32 @@ func (c *ClusterQueueSnapshot) FindTopologyAssignmentsForWorkload(
 		option(opts)
 	}
 
+	// One shared assumedUsage map per TopologyName for this workload. PodSets
+	// landing on different sibling flavors (same Topology, hostname leaf)
+	// reserve against the same map, preventing intra-workload self-overlap on
+	// a shared physical node. Cache-write-time aggregation does not cover
+	// in-flight reservations because pending workloads have not hit addUsage.
```
pajakd: This is tested in the test "multi-PodSet workload across sibling flavors must not self-overlap on shared node", right?

test/integration/singlecluster/tas/tas_test.go
```diff
@@ -4894,6 +4894,218 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
...
+				util.ExpectObjectToBeDeleted(ctx, k8sClient, clusterQueue, true)
+				util.ExpectObjectToBeDeleted(ctx, k8sClient, tasFlavorA, true)
+				util.ExpectObjectToBeDeleted(ctx, k8sClient, tasFlavorB, true)
+				util.ExpectObjectToBeDeleted(ctx, k8sClient, topology, true)
+				for _, node := range nodes {
+					util.ExpectObjectToBeDeleted(ctx, k8sClient, &node, true)
+				}
+			})
+
+			ginkgo.It("should admit the pending workload to tas-flavor-b on the free node when tas-flavor-a is full", func() {
```
pajakd, line 5311: How would this test fail without the fix?

pkg/util/maps/maps_test.go
```diff
@@ -292,3 +293,225 @@ func TestFilterKeys(t *testing.T) {
...
+			seed:          map[string]int{"k": 1},
+			wantCalls:     1,
+			wantCollected: map[string]int{"k": 1},
+		},
+		"multiple entries: every entry visited": {
+			seed:          map[string]int{"a": 1, "b": 2, "c": 3},
+			wantCalls:     3,
+			wantCollected: map[string]int{"a": 1, "b": 2, "c": 3},
+		},
+		"early termination: stops on first false": {
```
pajakd: shouldn't we check the value of `wantCollected` in this test? This is the only test where it should differ from `seed`, no?

pkg/cache/scheduler/snapshot.go
tenzen-y: Yes, sure.

test/integration/singlecluster/tas/tas_test.go
tenzen-y: This is repro PR: https://github.com/kubernetes-sigs/kueue/pull/10657/changes#r3118973170

pkg/cache/scheduler/clusterqueue_snapshot.go
tenzen-y: Yes, exactly, right.

pkg/util/maps/maps_test.go
tenzen-y: Oh, good point. You're right.

pkg/util/maps/maps.go
```diff
@@ -132,3 +132,20 @@ func (dwc *SyncMap[K, V]) Swap(k K, v V) (V, bool) {
 	dwc.m[k] = v
 	return old, existed
 }
+
+func (dwc *SyncMap[K, V]) Range(f func(key K, value V) bool) {
```
pajakd: Since this function holds only rlock, if someone would try to call it with f(k,v) that modifies the underlying map, it could be dangerous (deadlock?). Shouldn't we add a comment warning about this?

pkg/cache/scheduler/snapshot.go
```diff
@@ -185,8 +187,39 @@ func (c *Cache) Snapshot(ctx context.Context, options ...SnapshotOption) (*Snaps
...
+				return true
+			})
+			sharedSnapshotUsage[topologyName] = clonedUsage
+		}
+		c.tasCache.RUnlock()
+
+		for flavor, cache := range flavorClone {
+			nodes := c.tasCache.nodesCache.find(cache.flavor.NodeLabels, cache.topology.Levels)
+			var sharedUsage map[utiltas.TopologyDomainID]resources.Requests
+			if cache.topology.Usage != nil {
```
pajakd: In the loop above we don't initialize the `sharedSnapshotUsage[topologyName]` if `usageLength == 0`. But here we don't check for that. But if a topology has 0 usage, it might still need the shared map (for example if the first workload has multiple podsets)?

pkg/cache/scheduler/clusterqueue_snapshot.go
```diff
@@ -197,13 +198,32 @@ func (c *ClusterQueueSnapshot) FindTopologyAssignmentsForWorkload(
...
+	// iteration order is randomized otherwise.
+	for _, tasFlavor := range slices.Sorted(maps.Keys(tasRequestsByFlavor)) {
+		flavorTASRequests := tasRequestsByFlavor[tasFlavor]
 		// We assume the `tasFlavor` is already in the snapshot as this was
 		// already checked earlier during flavor assignment, and the set of
 		// flavors is immutable in snapshot.
 		tasFlavorCache := c.TASFlavors[tasFlavor]
-		flvResult := tasFlavorCache.FindTopologyAssignmentsForFlavor(flavorTASRequests, options...)
+		flvOpts := options
+		if tasFlavorCache != nil && tasFlavorCache.isLowestLevelNode {
```
pajakd: if tasFlavorCache would be nil it would panic in the line below this if. The comment above says that "`tasFlavor` is already in the snapshot" so I think `tasFlavorCache != nil` should either be removed entirely or changed to a guard clause.

pkg/util/maps/maps.go
tenzen-y: Yes, good point.
I'm wondering if we should take RW lock instead of RLock here.

Any preference adding comments vs taking RW lock?

pkg/cache/scheduler/snapshot.go
tenzen-y: Good catch, we should initialize that when `sharedSnapshotUsage[topologyName]` was not initialized in the previous loop block due to the `usageLength`.

pkg/cache/scheduler/clusterqueue_snapshot.go
tenzen-y: That totally makes sense. Let's remove `tasFlavorCache != nil` because we can expect that it has already been checked as described in the above code comment.

pkg/util/maps/maps_test.go
tenzen-y: I fixed this in https://github.com/kubernetes-sigs/kueue/pull/11210/commits/70b843439f7d4ee4a432486c215b2c332b480f32

The `seed` is map. So, the expectation is not stable. Hence, I added a verification if the `collected` map key-value pairs match with seed's one.

pkg/cache/scheduler/snapshot.go
tenzen-y: I addressed it in https://github.com/kubernetes-sigs/kueue/pull/11210/commits/52901477a36fc902f9e0c852628e4f833380cc75

pkg/cache/scheduler/snapshot.go
tenzen-y: I addressed this in https://github.com/kubernetes-sigs/kueue/pull/11210/commits/1be5bd2b238c7a11654e0b620fc71d3d612b9cea

The `sharedSnapshotUsage[topologyName]` is always initialized even when the length of usageLength is zero.

pkg/cache/scheduler/clusterqueue_snapshot.go
tenzen-y: I address it in https://github.com/kubernetes-sigs/kueue/pull/11210/commits/3afdebb6e3209d1aed12d7464c73d8d25c73833d

pkg/util/maps/maps.go
tenzen-y: I just added a comment in https://github.com/kubernetes-sigs/kueue/pull/11210/commits/8bfa984968f186595fd95713d8e06aa9979489de

Because Writable `Range` will cause performance regressions for syncMap.

tenzen-y:
@pajakd Thank you for your review, I addressed all your comments, PTAL, thank you 🙏

mimowo:
@tenzen-y please rebase and I think it would be good to squeeze the commits in case we cherrypick

pkg/cache/scheduler/snapshot.go
```diff
@@ -184,16 +178,43 @@ func (c *Cache) Snapshot(ctx context.Context, options ...SnapshotOption) (*Snaps
...
+			}
+			clonedUsage := make(map[utiltas.TopologyDomainID]resources.Requests, info.Usage.Len())
+
+			info.Usage.Range(func(domainID utiltas.TopologyDomainID, req resources.Requests) bool {
+				clonedUsage[domainID] = req.Clone()
+				return true
+			})
+			sharedSnapshotUsage[topologyName] = clonedUsage
+		}
+		c.tasCache.RUnlock()
```
mimowo: Exract this code to a helper function so that we can use the Lock, defer UnLock pattern

mimowo (commented):
Given the amount of non-trivial changes I would consider a feature gate for the fix, like `TASHandleOverlappingFlavors`

mimowo:
cc @Ladicle maybe could also give it a pass who already have pretty solid understanding of TAS

tenzen-y:
> Given the amount of non-trivial changes I would consider a feature gate for the fix, like `TASHandleOverlappingFlavors`

No objections at all.

tenzen-y:
> Given the amount of non-trivial changes I would consider a feature gate for the fix, like `TASHandleOverlappingFlavors`

Sure, let's do that.

tenzen-y:
In terms of squashing, I will squash all commits after @pajakd checked all changes. Because I tagged each commit with each of his review comments. Each commit can help understand fixes for @pajakd, I think.

pkg/cache/scheduler/snapshot.go
tenzen-y: That sounds good to me.

tenzen-y:
I rebased this PR on top of the main branch.

pkg/cache/scheduler/snapshot.go
tenzen-y: I addressed this in https://github.com/kubernetes-sigs/kueue/pull/11210/commits/5387a99a9f205ed43ca4870f6c2df7d39e903761

tenzen-y:
I'm currently working on the dedicated FG introduction, `TASHandleOverlappingFlavors`.

tenzen-y:
@pajakd Please let me know once you have checked the changes for your comments. After that, I will squach all commits into one for backports.

tenzen-y:
> I'm currently working on the dedicated FG introduction, `TASHandleOverlappingFlavors`.

@mimowo I'm assuming that we will introduce `TASHandleOverlappingFlavors` with Beta stage in the main branch, but we degrade the FG phase with alpha when we backport it to release branches, the same as previously, right?

pajakd:
> @pajakd Please let me know once you have checked the changes for your comments. After that, I will squach all commits into one for backports.

Checked all the changes. All looks good.

mimowo:
> @mimowo I'm assuming that we will introduce TASHandleOverlappingFlavors with Beta stage in the main branch, but we degrade the FG phase with alpha when we backport it to release branches, the same as previously, right?

Yes, it can be beta Imo  (I will yet double check), because it is bugfix. However, since TAS is already used in prod at scale we need to be careful about changes which have a risk of a performance hit. So, introducing the FG as a bailout is good.

tenzen-y:
> > @pajakd Please let me know once you have checked the changes for your comments. After that, I will squach all commits into one for backports.
> 
> Checked all the changes. All looks good.

Thank you for checking that. Let me squash all commits into one from now.

tenzen-y:
> > @mimowo I'm assuming that we will introduce TASHandleOverlappingFlavors with Beta stage in the main branch, but we degrade the FG phase with alpha when we backport it to release branches, the same as previously, right?
> 
> Yes, it can be beta Imo (I will yet double check), because it is bugfix. However, since TAS is already used in prod at scale we need to be careful about changes which have a risk of a performance hit. So, introducing the FG as a bailout is good.

That sounds reasonable to me, thank you for confirming that 👍

tenzen-y:
I turned this into a draft state to add FG.

tenzen-y:
@pajakd @mimowo I added feature gate guards as we discussed.
PTAL, thanks!

pkg/cache/scheduler/snapshot.go
```diff
@@ -192,8 +194,14 @@ func (c *Cache) Snapshot(ctx context.Context, options ...SnapshotOption) (*Snaps
 	}
 	tasSnapshots := make(map[kueue.ResourceFlavorReference]*TASFlavorSnapshot)
 	if features.Enabled(features.TopologyAwareScheduling) {
+		sharedSnapshotUsage := c.snapshotTopologyUsages()
```
mimowo: IIUC this is ungated currently, I'm wondering if we can gate this too

tenzen-y: Good point, let me guard this by TASHandleOverlappingFlavors.

tenzen-y: Done in https://github.com/kubernetes-sigs/kueue/pull/11210/commits/52e4733bfd58781a5b04f044f944fb56f2b8a135

pkg/cache/scheduler/tas_flavor.go
```diff
@@ -160,12 +194,33 @@ func (c *TASFlavorCache) updateUsage(topologyRequests []workload.TopologyDomainR
...
-			c.usage[domainID].Add(resources.Requests{corev1.ResourcePods: int64(tr.Count)})
+			c.usage[domainID].Add(total)
+			c.usage[domainID].Add(podDelta)
+		}
+		// For hostname-leaf topologies, mirror the delta to the shared
+		// SyncMap so sibling TASFlavorCache instances and Cache.Snapshot's
+		// clone path observe the cross-flavor aggregate. SyncMap.Update is
+		// atomic so concurrent sibling writers serialize on its lock.
+		if features.Enabled(features.TASHandleOverlappingFlavors) && c.topology.Usage != nil {
+			c.topology.Usage.Update(domainID, func(prev resources.Requests, _ bool) resources.Requests {
```
sohankunkerkar: I think the integration test only exercises the create→admit path. Could we add a test case that admits a workload via flavor A, finishes/deletes it, and then verifies flavor B can use the freed node? That would cover the subtract side of the shared accounting.

tenzen-y: That sounds reasonable, let me expand integration test case

test/integration/singlecluster/tas/tas_test.go
```diff
@@ -5186,6 +5186,334 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
...
+									Count:  1,
+									Values: []string{"x2"},
+								},
+							},
+						}),
+					))
+				})
+			})
+
+			ginkgo.It("should admit a sibling-flavor workload on the freed node once the primary workload finishes", func() {
```
tenzen-y, line 5428: @sohankunkerkar This is for https://github.com/kubernetes-sigs/kueue/pull/11210#discussion_r3307108634

pkg/cache/scheduler/tas_flavor.go
tenzen-y: Done.

tenzen-y:

due to https://github.com/kubernetes-sigs/kueue/issues/11631

pkg/cache/scheduler/clusterqueue_snapshot.go
```diff
@@ -198,13 +199,36 @@ func (c *ClusterQueueSnapshot) FindTopologyAssignmentsForWorkload(
 		option(opts)
 	}
 
+	// One shared assumedUsage map per TopologyName for this workload. PodSets
+	// landing on different sibling flavors (same Topology, hostname leaf)
+	// reserve against the same map, preventing intra-workload self-overlap on
+	// a shared physical node. Cache-write-time aggregation does not cover
+	// in-flight reservations because pending workloads have not hit addUsage.
+	var sharedAssumed map[kueue.TopologyReference]map[utiltas.TopologyDomainID]resources.Requests
```
mimowo: Why is the sharedAssumed key-ed by the topologyName? I think it should not really matter, as long as two Pods use the same domain (node) which toplogyName they were assigned using, wdyt?

pkg/scheduler/scheduler_tas_test.go
```diff
@@ -2448,6 +2448,243 @@ func TestScheduleForTAS(t *testing.T) {
 				utiltesting.MakeEventRecord("default", "foo", "Admitted", corev1.EventTypeNormal).Obj(),
 			},
 		},
+		// Two TAS ResourceFlavors share the same physical nodes (same NodeLabel and
+		// Topology). tas-flavor-b has an extra NodeTaint that only the pending
```
mimowo, line 2452: Do we have test case where the two ResourceFlavors overlap on nodes, but using different topology? Is that supported?

pkg/cache/scheduler/clusterqueue_snapshot.go
tenzen-y: Yes, that's right. Actually, I was taking another approach in the POC phase.
This `topoloyName` key style is not needed in this selected approach.

tenzen-y: Done in https://github.com/kubernetes-sigs/kueue/pull/11210/commits/fa0431dabfa18aaeca0d5e30fb2298f98bce6f85

pkg/cache/scheduler/snapshot.go
```diff
@@ -221,6 +232,30 @@ func (c *Cache) Snapshot(ctx context.Context, options ...SnapshotOption) (*Snaps
...
+		if info.Usage == nil {
+			continue
+		}
+		clonedUsage := make(map[utiltas.TopologyDomainID]resources.Requests, info.Usage.Len())
+
+		info.Usage.Range(func(domainID utiltas.TopologyDomainID, req resources.Requests) bool {
+			clonedUsage[domainID] = req.Clone()
+			return true
+		})
+		topologyUsages[topologyName] = clonedUsage
```
mimowo: Here I'm also not sure why this code is tracking the usage per topologyName, conceptually I think we just have a cluster-wide TAS usage per domain, coming from admitted workloads.

pkg/scheduler/scheduler_tas_test.go
tenzen-y: We don't have such a test case because I didn't imagine such a case, as both resource flavors point to the same nodes. 

If we support it, we allow them to assign a different topology to a single Node. It sounds curious from my PoV.

Do you see the request from your side?

tenzen-y:
> @tenzen-y: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> pull-kueue-test-integration-shard-2-main	[fa0431d](https://github.com/kubernetes-sigs/kueue/commit/fa0431dabfa18aaeca0d5e30fb2298f98bce6f85)	[link](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/11210/pull-kueue-test-integration-shard-2-main/2059689283937112064)	true	`/test pull-kueue-test-integration-shard-2-main`
> [Full PR test history](https://prow.k8s.io/pr-history?org=kubernetes-sigs&repo=kueue&pr=11210). [Your PR dashboard](https://prow.k8s.io/pr?query=is%3Apr%20state%3Aopen%20author%3Atenzen-y). Please help us cut down on flakes by [linking to](https://git.k8s.io/community/contributors/devel/sig-testing/flaky-tests.md#filing-issues-for-flaky-tests) an [open issue](https://github.com/kubernetes-sigs/kueue/issues?q=is:issue+is:open) when you hit one in your PR.
> 
> Details

opened https://github.com/kubernetes-sigs/kueue/issues/11694

pkg/features/kube_features.go
```diff
@@ -664,10 +669,12 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 	WorkloadPriorityClassDefaulting: {
 		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Alpha},
 	},
-
 	MetricsForCohorts: {
 		{Version: version.MustParse("0.18"), Default: true, PreRelease: featuregate.Beta},
 	},
+	TASHandleOverlappingFlavors: {
+		{Version: version.MustParse("0.18"), Default: true, PreRelease: featuregate.Beta},
+	},
```
tenzen-y, line 677: As we discussed with @mimowo offline, we've decided to start this FG from the alpha phase.
Because this change is slightly performance sensitive (like changing the usage calculation mechanism).

So, we disable this by default.

tenzen-y: I will replace this feature phase with the alpha tomorrow.

pkg/cache/scheduler/snapshot.go
```diff
@@ -192,8 +194,17 @@ func (c *Cache) Snapshot(ctx context.Context, options ...SnapshotOption) (*Snaps
...
+			sharedSnapshotUsage = c.snapshotTopologyUsages()
+		}
 		for flavor, cache := range c.tasCache.Clone() {
-			tasSnapshots[flavor] = cache.snapshot(log, c.tasCache.nodesCache.find(cache.flavor.NodeLabels, cache.topology.Levels))
+			nodes := c.tasCache.nodesCache.find(cache.flavor.NodeLabels, cache.topology.Levels)
+			var sharedUsage map[utiltas.TopologyDomainID]resources.Requests
+			if features.Enabled(features.TASHandleOverlappingFlavors) && cache.topology.Usage != nil {
+				sharedUsage = sharedSnapshotUsage[cache.flavor.TopologyName]
+			}
+			tasSnapshots[flavor] = cache.snapshot(log, nodes, sharedUsage)
```
mimowo: IIUC the code aims computing globalTASUsage and pass it inside `snapshot`, is that correct? If so then I would call it globalTASUsage.

Second, I think we may not have the whole usage at hand collected from all flavors when calling snapshot as we are building it greedily.

Maybe I'm missing something but I'm wondering if we could make it more readable by the two-pass approach:

0. tasCacheClone = c.tasCache.Clone()
1. compute the aggregated globalTASUsage based on  tasCacheClone
2. trim the globalTASUsage to the flavor based on `nodes`

This way I think we can signifficantly simplify:
- `snapshotTopologyUsages` does not need to track per topologyName
- `Usage` could be dropped under topologyInformation
- https://github.com/kubernetes-sigs/kueue/pull/11210/changes#diff-aa793cfc2528f289b2b59725f714db6e6f6b2c6e9edf4c75925e656e0cc055b7R206-R223 - could be dropped

pkg/features/kube_features.go
mimowo: If I'm not mistaken in this comment https://github.com/kubernetes-sigs/kueue/pull/11210/changes#r3313550771 then we can signifficantly simplify the code. If this is the case then I'm open to re-consider Beta.

I was just hesitant going streight to beta seeing the complications. I'm happy to also sync if I'm missing something.

pkg/cache/scheduler/snapshot.go
tenzen-y: Yes, we can simplify code by using the alternative approach. Indeed, I implemented a similar one as my PoC PR: https://github.com/kubernetes-sigs/kueue/pull/10688

But, this has scheduling performance degrading risks, so I am moving this #11210 approach.

#10688 has to aggregate all usages at snapshot timing, which takes a longer time than computing aggregation at usage change timing. It could lead to longer TAS scheduling duration every scheduling cycle because #10688 doesn't preserve aggregation information anywhere.

mimowo: But the aggregation does not increase asymptotic complexity, because we need to copy the usage for all flavors anyway. It just has another pass to aggregate the data. I think it is also using the same memory.

I think I'm more worried about performance regressions due to a bug due to a too complicated code.

tenzen-y: > But the aggregation does not increase asymptotic complexity,

Yes, for each scheduling cycle, but from the sum of scheduling cycles during the arbitrary time PoV, it will be different because #11210 can reuse already computed aggregation information.

> I think I'm more worried about performance regressions due to a bug due to a too complicated code.

Isn't it always? Adding any computing mechanism to snapshotting timing is easy because we can just add our business logic, but the cache mechanism's advantage will not be performed. 

The feature gate is responsible for guarding Kueue by breaking existing behavior. So, I think that we can avoid worrying about bugs.

mimowo: > Yes, for each scheduling cycle, but from the sum of scheduling cycles during the arbitrary time PoV, it will be different because https://github.com/kubernetes-sigs/kueue/pull/11210 can reuse already computed aggregation information.

Maybe, but it still does not improve the asymptotic complexity of a single cycle.

> Isn't it always? Adding any computing mechanism to snapshotting timing is easy because we can just add our business logic, but the cache mechanism's advantage will not be performed.

I want to avoid premature optimization. My gut feeling is that in this case we really don't gain much by the optimization given how much TAS is processing overall. 

I would start simple, and complicate code when we prove this is a bottlneck. One idea to prove would be to run our tas perf tests with CPU profiling enabled and confirm the results. I don't expect visible difference. 

I know it is late for the experiments in this cycle, so I still propose simple Alpha, and we can consider more sothisticated code for Beta if we prove bottlneck.

pkg/cache/scheduler/tas_flavor.go
```diff
@@ -115,19 +116,41 @@ func (c *TASFlavorCache) TopologyLevels() []string {
...
+			snapshot.leaves[domainID].tasUsage = aggregatedResources
+		}
 	}
 	snapshot.initialize()
-	for domainID, usage := range c.usage {
-		snapshot.addTASUsage(domainID, usage)
+	if aggregatedDomainUsages == nil || !features.Enabled(features.TASHandleOverlappingFlavors) {
+		for domainID, usage := range c.usage {
+			snapshot.addTASUsage(domainID, usage)
+		}
```
mimowo: I'm wondering if we could commonize adding the TAS usage to snaphot if we introduced a variable like

tasUsageToAdd := c.usage
if features.Enabled(features.TASHandleOverlappingFlavors)  {
  tasUsageToAdd = // some conputation 
}
snapshot.initialize()

for domainID, usage := tasUsageToAdd {
	snapshot.addTASUsage(domainID, usage)
}

pkg/cache/scheduler/tas_flavor.go
```diff
@@ -115,19 +116,41 @@ func (c *TASFlavorCache) TopologyLevels() []string {
...
 	nodeToDomain := make(map[string]utiltas.TopologyDomainID)
 	for _, node := range nodes {
-		nodeToDomain[node.Name] = snapshot.addNode(node)
+		domainID := snapshot.addNode(node)
+		nodeToDomain[node.Name] = domainID
+		if features.Enabled(features.TASHandleOverlappingFlavors) && aggregatedDomainUsages != nil {
+			aggregatedResources, ok := aggregatedDomainUsages[domainID]
+			if !ok {
+				aggregatedResources = resources.Requests{}
+				aggregatedDomainUsages[domainID] = aggregatedResources
```
mimowo: Why mutating the map in here, it would be cleaner if we could fully pre-compute it before calling snapshot(), is there some issue with that?

pkg/cache/scheduler/tas_flavor.go
tenzen-y: Sorry, I wasn't sure where do we want to commonize?

mimowo: I mean this two paths:

1. new: `snapshot.leaves[domainID].tasUsage = aggregatedResources` L146
2. old `snapshot.addTASUsage(domainID, usage)` L152

Because IIUC they both do the same just slightly differently. So I wanted to reuse the pre-existing code, just feed it with `tasUsageToAdd` rather than c.usage

pkg/cache/scheduler/tas_flavor.go
tenzen-y: Good call.
I think that we can pre-compute. This was preserved from my previous PoC codes.
So, there is really no need.

tenzen-y: I moved this node calculation to `snapshot()` in https://github.com/kubernetes-sigs/kueue/pull/11210/commits/f68c1c75f1734506386b4509f471261bfe866581

Please let me know if I am still missing anything, thanks.

pkg/cache/scheduler/tas_flavor.go
tenzen-y: I see, that makes sense. Let me try to consolidate such duplicated mechanisms into one.

pkg/cache/scheduler/tas_flavor.go
tenzen-y: After https://github.com/kubernetes-sigs/kueue/pull/11210#discussion_r3319204994 change, we might not need this change. Let me check what we should do.

pkg/cache/scheduler/tas_flavor.go
tenzen-y: I addressed this in https://github.com/kubernetes-sigs/kueue/pull/11210/commits/a07fa048ddf56b4652f7ccfd84a86173e32a562b

pkg/cache/scheduler/tas_flavor.go
tenzen-y: https://github.com/kubernetes-sigs/kueue/pull/11210/commits/a07fa048ddf56b4652f7ccfd84a86173e32a562b is the final version.

tenzen-y:
I'm currently checking an integration test failure. I will push fix if it is not flake.
Give me some time.

tenzen-y:
> I'm currently checking an integration test failure. I will push fix if it is not flake. Give me some time.

I got the bug point. I will push bug fix code.

pkg/cache/scheduler/snapshot.go
```diff
@@ -221,6 +236,24 @@ func (c *Cache) Snapshot(ctx context.Context, options ...SnapshotOption) (*Snaps
...
+	tasFlvCache.RLock()
+	defer tasFlvCache.RUnlock()
+
+	if len(tasFlvCache.topology.Levels) == 0 || !utiltas.IsLowestLevelHostname(tasFlvCache.topology.Levels) {
+		return
+	}
+	for domainID, domainUsage := range tasFlvCache.usage {
+		if _, ok := aggregatedDomainUsages[domainID]; ok {
+			aggregatedDomainUsages[domainID].Add(domainUsage)
+			continue
```
mimowo: nit, this continue is a bit tricky to follow, maybe instead use `else` for the outer "if"?

pkg/cache/scheduler/tas_cache_test.go
```diff
@@ -7535,12 +7676,24 @@ func TestFindTopologyAssignmentsMultiLayerReplacement(t *testing.T) {
...
+			if features.Enabled(features.TASHandleOverlappingFlavors) && len(tc.priorFlavorUsage) > 0 {
+				siblingCache := tasCache.NewTASFlavorCache(topologyInfo, flvInfo)
+				siblingCache.addUsage("prior-wl", tc.priorFlavorUsage)
+				if tc.aggregatedDomainUsages == nil {
+					tc.aggregatedDomainUsages = make(map[tas.TopologyDomainID]resources.Requests, len(siblingCache.usage))
+				}
+				for domainID, usage := range siblingCache.usage {
+					tc.aggregatedDomainUsages[domainID] = usage.Clone()
+				}
+			}
```
mimowo: This code looks copy-pasted in a couple places, let's commonize

tenzen-y:
> > I'm currently checking an integration test failure. I will push fix if it is not flake. Give me some time.
> 
> I got the bug point. I will push bug fix code.

I pushed the bug fix code. Let's see if this can resolve that.

pkg/cache/scheduler/clusterqueue_snapshot.go
```diff
@@ -198,15 +199,26 @@ func (c *ClusterQueueSnapshot) FindTopologyAssignmentsForWorkload(
...
 		tasFlavorCache := c.TASFlavors[tasFlavor]
-		flvResult := tasFlavorCache.FindTopologyAssignmentsForFlavor(flavorTASRequests, options...)
+		flvOpts := options
+		if features.Enabled(features.TASHandleOverlappingFlavors) && tasFlavorCache.isLowestLevelNode {
+			flvOpts = append(slices.Clone(options), WithAggregatedDomainUsages(aggregatedDomainUsages))
+		}
+		flvResult := tasFlavorCache.FindTopologyAssignmentsForFlavor(flavorTASRequests, flvOpts...)
 		maps.Copy(result, flvResult)
 	}
+
```
mimowo: nit, drop empty line insertion

pkg/cache/scheduler/snapshot.go
tenzen-y: I fixed this in https://github.com/kubernetes-sigs/kueue/compare/f5f810f6dc7a8778df9985ac1cfe2f9aca8f91f4..940ab1232aff067991bc5f9c2665d6e6805ccb37

pkg/cache/scheduler/tas_cache_test.go
tenzen-y: Yes, right. Are you ok to introduce prod commonized codes to UT?

pkg/util/maps/maps.go
```diff
@@ -132,3 +132,24 @@ func (dwc *SyncMap[K, V]) Swap(k K, v V) (V, bool) {
...
+			return
+		}
+	}
+}
+
+func (dwc *SyncMap[K, V]) Update(k K, fn func(prev V, exists bool) V) {
+	dwc.lock.Lock()
+	defer dwc.lock.Unlock()
+	prev, exists := dwc.m[k]
+	dwc.m[k] = fn(prev, exists)
```
mimowo: Are these in use in the new impl? If not then drop

tenzen-y: No need. That is dead code. Let's remove it.

tenzen-y:
> @tenzen-y: The following tests **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> pull-kueue-test-integration-shard-0-main	[a07fa04](https://github.com/kubernetes-sigs/kueue/commit/a07fa048ddf56b4652f7ccfd84a86173e32a562b)	[link](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/11210/pull-kueue-test-integration-shard-0-main/2060044168830390272)	true	`/test pull-kueue-test-integration-shard-0-main`
> pull-kueue-test-integration-shard-1-main	[940ab12](https://github.com/kubernetes-sigs/kueue/commit/940ab1232aff067991bc5f9c2665d6e6805ccb37)	[link](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/11210/pull-kueue-test-integration-shard-1-main/2060051729759604736)	true	`/test pull-kueue-test-integration-shard-1-main`
> [Full PR test history](https://prow.k8s.io/pr-history?org=kubernetes-sigs&repo=kueue&pr=11210). [Your PR dashboard](https://prow.k8s.io/pr?query=is%3Apr%20state%3Aopen%20author%3Atenzen-y). Please help us cut down on flakes by [linking to](https://git.k8s.io/community/contributors/devel/sig-testing/flaky-tests.md#filing-issues-for-flaky-tests) an [open issue](https://github.com/kubernetes-sigs/kueue/issues?q=is:issue+is:open) when you hit one in your PR.
> 
> Details

infra

pkg/cache/scheduler/clusterqueue_snapshot.go
tenzen-y: removed in https://github.com/kubernetes-sigs/kueue/pull/11210/commits/bbd278f9cff14f9b9ac95859482b0eb6bbd8326c

pkg/cache/scheduler/tas_cache_test.go
mimowo: yes, why not.

pkg/util/maps/maps.go
tenzen-y: Removed those in https://github.com/kubernetes-sigs/kueue/pull/11210/commits/05e0d2754e9252fc41b9209b2bb1a2b19a5e7e83

pkg/cache/scheduler/tas_cache_test.go
tenzen-y: just checked that. let me confirm how we can commonize those into comprehensive func.

tenzen-y: Uhm, it seems that it's not easy to commonize this mechanism in prod code.
If we want to commonize this, we should have test specific util library.

mimowo: We have util for package for TAS, maybe place it there, or at least commonize the two similar places in unit tests

tenzen-y: The core reason why this is just a copy-paste style is `TestFindTopologyAssignmentsMultiLayerReplacement` and `TestFindTopologyAssignments` are the mostly same. `TestFindTopologyAssignmentsMultiLayerReplacement` has just enabled specific FG.

https://github.com/kubernetes-sigs/kueue/blob/851a097d19c4326bdfdf0276492625e17a08495e/pkg/cache/scheduler/tas_cache_test.go#L7206

https://github.com/kubernetes-sigs/kueue/blob/851a097d19c4326bdfdf0276492625e17a08495e/pkg/cache/scheduler/tas_cache_test.go#L56

tenzen-y: > We have util for package for TAS, maybe place it there, or at least commonize the two similar places in unit tests

That meant that we have to consolidate `TestFindTopologyAssignmentsMultiLayerReplacement` and `TestFindTopologyAssignments` into one...

mimowo: I see , we can keep it for follow up, but maybe we could consolidate the two tests?

tenzen-y: My proposal is that 

1) adding a temporarily consolidated function for line 7685-7694 in this file, then
2) consolidate TestFindTopologyAssignmentsMultiLayerReplacement and TestFindTopologyAssignments into one
3) In the same PR as 2), we should remove 1) the tmp commonized function.

mimowo: Sgtm

tenzen-y: done in https://github.com/kubernetes-sigs/kueue/pull/11210/commits/d22e18bdae2e368b985fd6963fb8a0869e87139b

pkg/features/kube_features.go
```diff
@@ -675,10 +669,12 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 	WorkloadPriorityClassDefaulting: {
 		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Alpha},
 	},
-
 	MetricsForCohorts: {
 		{Version: version.MustParse("0.18"), Default: true, PreRelease: featuregate.Beta},
 	},
+	TASHandleOverlappingFlavors: {
+		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Alpha},
+	},
```
tenzen-y, line 677: I disabled this FG by default in https://github.com/kubernetes-sigs/kueue/pull/11210/commits/ae1db16ca849bff7a47b1435f76146935f4a937d

pkg/cache/scheduler/snapshot.go
```diff
@@ -192,8 +194,27 @@ func (c *Cache) Snapshot(ctx context.Context, options ...SnapshotOption) (*Snaps
...
-			tasSnapshots[flavor] = cache.snapshot(log, c.tasCache.nodesCache.find(cache.flavor.NodeLabels, cache.topology.Levels))
+		var aggregatedDomainUsages map[utiltas.TopologyDomainID]resources.Requests
+		flvTASCache := c.tasCache.Clone()
+
+		if features.Enabled(features.TASHandleOverlappingFlavors) {
+			aggregatedDomainUsages = make(map[utiltas.TopologyDomainID]resources.Requests)
+			for _, cache := range flvTASCache {
+				c.snapshotTopologyDomainUsages(cache, aggregatedDomainUsages)
+			}
+		}
```
mimowo, line 206: Maybe we could add here a log line, so that users can tell how long roughly different phases of TAS take

log.V(4).Info("Aggregated TAS usage across flavors")

mimowo (commented):
LGTM, just left a small comment to consider enriching the logs for the new code.

Seeing how much the code is simplified I'm good to go to Beta in 0.18. I'm also ok to cherrypick, probably as alpha.

pkg/features/kube_features.go
mimowo: Ok, I just said that I'm ok with Beta in 0.18 https://github.com/kubernetes-sigs/kueue/pull/11210#pullrequestreview-4383838161

I'm happy of the current code shape, up to you. I would just like to keep it alpha if for the release branches so that we can have more time before it lands on prod actually.

tenzen-y:
1. I reverted to making TASHandleOverlappingFlavors alpha (new code enables it by default)
2. I squashed all commits into one
3. I added logs as you mentioned.

Thanks for the review 👍

pkg/cache/scheduler/snapshot.go
tenzen-y: Added in https://github.com/kubernetes-sigs/kueue/pull/11210/commits/c7efbca3d9129caee9f53c27ee13a66959100ac7

pkg/features/kube_features.go
tenzen-y: I turned this TASHandleOverlappingFlavors to Beta in this branch.

I will manually open cherry-pick PRs to make TASHandleOverlappingFlavors Alpha after this PR is mergd.

mimowo (commented):
/lgtm
/approve
Thank you for closing the long-standing gap in TAS 👍  I propose to CP as alpha being careful about the prod use of TAS.

tenzen-y:
> /lgtm /approve Thank you for closing the long-standing gap in TAS 👍 I propose to CP as alpha being careful about the prod use of TAS.

I will open PRs before tomorrow's release.

pkg/cache/scheduler/tas_cache_test.go
tenzen-y: @coderabbitai Please open an issue to track 2 and 3 described in https://github.com/kubernetes-sigs/kueue/pull/11210#discussion_r3319835282 with kind/cleanup label.

PR #11218, opened 2026-05-15
Use Kueue config for populator namespace selector

### LINKED ISSUE #7852: Kueue-populator: re-use Kueue configuration to restrict the namespaces

I would like the `kueue-populator` to use the configuration used by Kueue for `managedJobsNamespaceSelector` instead of defining its own. This would allow the populator to respect the same namespace restrictions as the main Kueue manager without requiring duplicate configuration.

**Why is this needed**:

- Consistency: Ensures that both kueue-populator and the main Kueue controller target the exact same set of namespaces.
- Simplicity

**Completion requirements**:

- [ ] Code changes to make populator read Kueue config
- [ ] Update `kueue-populator` helm chart values.yaml
- [ ] Docs update (if applicable)

MatteoFari:
#### What type of PR is this?
/kind feature

`kueue-populator` now uses the same `managedJobsNamespaceSelector` as Kueue.

The Helm chart copies `managedJobsNamespaceSelector` from Kueue’s manager config into the existing populator config, instead of keeping a separate Helm value.

Fixes #7852

```release-note
kueue-populator: Supports Kueue's managedJobsNamespaceSelector from the Kueue manager configuration instead of a separate Helm value.
```

### DIFF
--- a/cmd/experimental/kueue-populator/charts/kueue-populator/templates/manager.yaml
+++ b/cmd/experimental/kueue-populator/charts/kueue-populator/templates/manager.yaml
@@ -1,3 +1,24 @@
+{{- $kueueConfig := dict -}}
+{{- with .Values.kueue }}
+{{- with .managerConfig }}
+{{- with .controllerManagerConfigYaml }}
+{{- $kueueConfig = . | fromYaml -}}
+{{- end }}
+{{- end }}
+{{- end }}
+{{- $kueueNamespace := .Release.Namespace -}}
+{{- with $kueueConfig.namespace }}
+{{- $kueueNamespace = . -}}
+{{- end }}
+{{- $managedJobsNamespaceSelector := dict "matchExpressions" (list (dict "key" "kubernetes.io/metadata.name" "operator" "NotIn" "values" (list "kube-system" $kueueNamespace))) -}}
+{{- if and (hasKey $kueueConfig "managedJobsNamespaceSelector") (ne $kueueConfig.managedJobsNamespaceSelector nil) }}
+{{- $managedJobsNamespaceSelector = $kueueConfig.managedJobsNamespaceSelector -}}
+{{- end }}
+{{- with .Values.kueuePopulator.config }}
+{{- if and (hasKey . "managedJobsNamespaceSelector") (ne .managedJobsNamespaceSelector nil) }}
+{{- $managedJobsNamespaceSelector = .managedJobsNamespaceSelector -}}
+{{- end }}
+{{- end }}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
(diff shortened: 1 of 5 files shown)


cmd/experimental/kueue-populator/charts/kueue-populator/templates/manager.yaml
```diff
@@ -1,3 +1,10 @@
+{{- $defaultKueueConfig := "apiVersion: config.kueue.x-k8s.io/v1beta2\nkind: Configuration\n" -}}
+{{- $kueueConfig := $defaultKueueConfig -}}
+{{- with .Values.kueue }}
+{{- with .managerConfig }}
+{{- $kueueConfig = .controllerManagerConfigYaml | default $defaultKueueConfig -}}
```
kannon92: Honestly this is quite a lot of API to bring in just for one field.

Should we just expose the managedNamespaceSelector in the config and read it from the config map for Kueue?

MatteoFari: Yeah it's better, the chart now reads Kueue’s manager config and writes its `managedJobsNamespaceSelector` into the existing populator config field

mimowo (commented):
cc @j-skiba as the original author of kueue-populator: https://github.com/kubernetes-sigs/kueue/pull/11218

cmd/experimental/kueue-populator/charts/kueue-populator/tests/manager_test.yaml
```diff
@@ -30,20 +30,76 @@ tests:
...
+    release:
+      namespace: custom-kueue-system
+    set:
+      kueue.managerConfig.controllerManagerConfigYaml: |
+        apiVersion: config.kueue.x-k8s.io/v1beta2
+        kind: Configuration
+    asserts:
+      - matchRegex:
+          path: data["controller_manager_config.yaml"]
+          pattern: "managedJobsNamespaceSelector:\n\\s+matchExpressions:\n\\s+- key: kubernetes.io/metadata.name\n\\s+operator: NotIn\n\\s+values:\n\\s+- kube-system\n\\s+- custom-kueue-system"
```
j-skiba: nit: is this possible to split these `pattern` fields into multiline fields?
something like:
```yaml
pattern: |
  managedJobsNamespaceSelector:
  \s+matchExpression:
...
```

j-skiba:
/lgtm
one nit, looks good overall

mimowo (commented):
Thank you, reading the namespace selector from Kueue's configuration sounds great, because they should be equal most of the time. However, do we support still setting the  `managedJobsNamespaceSelector` field explicitily in the Kueue populator config? There might still exist a use case when the Kueue-populator wants to restrict the set of namespaces it operates on. If this remains supported, do we have a test to demonstrate that?

MatteoFari:
> Thank you, reading the namespace selector from Kueue's configuration sounds great, because they should be equal most of the time. However, do we support still setting the `managedJobsNamespaceSelector` field explicitily in the Kueue populator config? There might still exist a use case when the Kueue-populator wants to restrict the set of namespaces it operates on. If this remains supported, do we have a test to demonstrate that?


Yeah that's more flexible. I restored `kueuePopulator.config.managedJobsNamespaceSelector` as an explicit override over the Kueue selector and added a Helm unit test for it.

mimowo (commented):
/lgtm
/approve
Thank you for working on this feature to respect the Kueue configuration without the need to configure kueue-populator, at the same time respecting the populator local config when present.

tenzen-y:
/release-note-edit

```release-note
kueue-populator: Supports Kueue's managedJobsNamespaceSelector from the Kueue manager configuration instead of a separate Helm value.
```

PR #11251, opened 2026-05-15
Switch to using kind build node-image for e2e tests

### LINKED ISSUE #11249: Switch to using `kind build node-images`

https://github.com/kubernetes-sigs/kind/issues/4157

Kind would prefer that we build the nodes images on releases without relying on official images.

**Why is this needed**:

This can put us in control of when we do e2e testing and not rely on kind to have an official release.

**What we need to do**:

I think we can switch our CI to do `kind build node-image v1.36.0` and that will pull latest release artifacts. And then we can just use that cluster instead of relying on kind to host official images.

kannon92:
#### What type of PR is this?

/kind cleanup
/area testing

Switches e2e tests to build kind node images locally from K8s release
artifacts instead of pulling pre-built `kindest/node` images from
Docker Hub.

This adds a `build_kind_node_image` function in `e2e-common.sh` that
runs `kind build node-image v<version>` before cluster creation. The
command downloads release tarballs (does not compile K8s from source),
so the cost is comparable to pulling the pre-built image.

Behavior:
- Non-standard images (e.g. `k8s-main:latest`) are skipped since they
  have their own build targets
- In dev mode (`E2E_MODE=dev`), the build is skipped when the image
  already exists locally

Fixes #11249

The existing `kind-k8s-main-image-build` Makefile target is unchanged
since it builds from CI artifacts, not releases.

```release-note
NONE
```

---

### DIFF
--- a/hack/testing/e2e-common.sh
+++ b/hack/testing/e2e-common.sh
@@ -43,6 +43,23 @@ export E2E_SKIP_IMAGE_RELOAD="${E2E_SKIP_IMAGE_RELOAD:-false}"
 
 export KIND_VERSION="${E2E_KIND_VERSION/"kindest/node:v"/}"
 
+function build_kind_node_image {
+    if [[ "$E2E_KIND_VERSION" != kindest/node:v* ]]; then
+        echo "Skipping kind node image build for non-standard image: $E2E_KIND_VERSION"
+        return 0
+    fi
+
+    E2E_KIND_VERSION="kueue/kind-node:v${KIND_VERSION}"
+
+    if [[ "${E2E_MODE}" == "dev" ]] && docker image inspect "$E2E_KIND_VERSION" &>/dev/null; then
+        echo "Reusing existing node image: $E2E_KIND_VERSION (E2E_MODE=dev)"
+        return 0
+    fi
+
+    echo "Building kind node image: $E2E_KIND_VERSION (K8s v$KIND_VERSION)"
+    $KIND build node-image "v$KIND_VERSION" --image "$E2E_KIND_VERSION"
+}
+
 function e2e_is_truthy {
     case "${1:-}" in
         1|true|TRUE|True|yes|YES|Yes|y|Y|on|ON|On) return 0 ;;
--- a/hack/testing/e2e-multikueue-test.sh
+++ b/hack/testing/e2e-multikueue-test.sh
@@ -90,7 +90,8 @@ function kueue_deploy {
 }
 
 trap cleanup EXIT
-startup 
+build_kind_node_image
+startup
 prepare_docker_images
 for job in $(jobs -p); do 
     wait "$job" || { echo "Cluster creation failed!"; exit 1; } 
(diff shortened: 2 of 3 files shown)


hack/testing/e2e-common.sh
```diff
@@ -43,6 +43,21 @@ export E2E_SKIP_IMAGE_RELOAD="${E2E_SKIP_IMAGE_RELOAD:-false}"
...
+        return 0
+    fi
+
+    if [[ "${E2E_MODE}" == "dev" ]] && docker image inspect "$E2E_KIND_VERSION" &>/dev/null; then
+        echo "Reusing existing node image: $E2E_KIND_VERSION (E2E_MODE=dev)"
+        return 0
+    fi
+
+    echo "Building kind node image: $E2E_KIND_VERSION (K8s v$KIND_VERSION)"
+    $KIND build node-image "v$KIND_VERSION" --image "$E2E_KIND_VERSION"
```
mbobrovskyi, line 60: How much longer does it take compared to just downloading it? I’m wondering if we can first check whether the image already exists, and only build it if it doesn’t.

mimowo: cc @kannon92

mimowo: Doesn't seem too bad, the longest taking suite took 26min, which seems normal for the suite: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/11251/pull-kueue-test-e2e-sequential-baseline-main/2055376057527177216

mimowo:
/test all
checking if this doesn't introduce any flakiness on downloading the packages in a new way
try (1/5)

mimowo:
cc @BenTheElder ptal

hack/testing/e2e-common.sh
BenTheElder: Either way you have to download the base image layer (squashed) and the kubernetes binaries (Either via a node-image or via the release tarball) so it's similar.

hack/testing/e2e-common.sh
```diff
@@ -43,6 +43,21 @@ export E2E_SKIP_IMAGE_RELOAD="${E2E_SKIP_IMAGE_RELOAD:-false}"
 
 export KIND_VERSION="${E2E_KIND_VERSION/"kindest/node:v"/}"
 
+function build_kind_node_image {
+    if [[ "$E2E_KIND_VERSION" != kindest/node:v* ]]; then
```
BenTheElder, line 47: Same comments as https://github.com/kubernetes-sigs/jobset/pull/1234#discussion_r3251113398

mbobrovskyi:
/test all

try (2/5)

hack/testing/e2e-common.sh
kannon92: yea it seems similar.

hack/testing/e2e-common.sh
kannon92: addressed.

mimowo:
/lgtm
/approve
Thank you, let's give it a try 👍 
/cherrypick release-0.17
/cherrypick release-0.16

PR #11252, opened 2026-05-15
metrics: add workload-level preemption count gauge

### LINKED ISSUE #11171: metrics: add workload-level preemption count gauge
## Summary

Kueue currently tracks preemptions at the **ClusterQueue level** via `kueue_preempted_workloads_total` (a Counter). There is no way to track how many times a specific workload has been preempted.

This makes it impossible to detect and alert on workloads that are repeatedly preempted — a pattern that indicates starvation or misconfigured priority/quota.

## Current State

Existing preemption metrics:

| Metric | Type | Labels | Granularity |
|---|---|---|---|
| `kueue_preempted_workloads_total` | Counter | `preempting_cluster_queue`, `reason`, `replica_role` | ClusterQueue |
| `kueue_admission_cycle_preemption_skips` | Gauge | `cluster_queue` | ClusterQueue |

Neither metric carries workload identity (`namespace`, `name`).

## Problem with `_total` Counter for Sparse Events

`kueue_preempted_workloads_total` requires `rate()` or `increase()` in PromQL to be useful for alerting. This works well for high-frequency events, but workload preemptions can be very sparse — days or weeks apart. For sparse events:

- `rate()` over any practical window (e.g. `[1h]`, `[24h]`) will return `0` or a near-zero float even when a workload has been preempted multiple times.
- Threshold-based alerts (`> N`) are not expressible with a counter alone.

## Request

Add a new **Gauge** metric tracking cumulative preemption count per workload:

```
kueue_workload_preemptions_count
```

- **Type:** Gauge
- **Labels:** `namespace`, `workload` (workload name), `cluster_queue`, `reason`
- **Semantics:** incremented each time the workload is preempted; deleted (with matching labels) when the workload completes or is deleted
- **Measurement point:** inside `metrics.ReportPreemption()` or equivalent, where the target workload is known

This enables straightforward alerting:

```promql
# Alert: workload preempted more than 3 times
kueue_workload_preemptions_count > 3
```

which is not achievable with the current counter-based metric.

## Workaround
We can "extract" workload requeue and/or eviction count from the Workload CRD via Kube-state-metrics(KSM)
```
requeueState:
  count: 6
schedulingStats:
  evictions:
    - count: 6
      reason: PodsReadyTimeout
      underlyingCause: WaitForStar
```

## Environment

- Kueue version:
- Kubernetes version:

gyliu513:
#### What type of PR is this?
/kind bug

Fixes #11171 

```release-note
metrics: add workload-level preemption count gauge
```


mimowo:
cc @amy ptal

pkg/metrics/metrics.go
```diff
@@ -188,6 +188,10 @@ var (
 	// +metricsdoc:labels=preempting_cluster_queue="the ClusterQueue executing preemption",reason="eviction or preemption reason",replica_role="one of `leader`, `follower`, or `standalone`"
 	PreemptedWorkloadsTotal *prometheus.CounterVec
 
+	// +metricsdoc:group=workload
+	// +metricsdoc:labels=namespace="the namespace of the preempted workload",workload="the name of the preempted workload",cluster_queue="the ClusterQueue of the preempted workload",reason="eviction or preemption reason"
+	WorkloadPreemptionsCount *prometheus.GaugeVec
```
amy: If you want to model similar to `PreemptedWorkloadsTotal`, please rename count to total.

pkg/metrics/metrics.go
```diff
@@ -651,6 +655,21 @@ The label 'reason' can have the following values:
 		}, append([]string{"preempting_cluster_queue", "reason", "replica_role"}, extraLabels...),
 	)
 
+	WorkloadPreemptionsCount = prometheus.NewGaugeVec(
```
amy: The Help text says: "Uses a Gauge (rather than a Counter) so that the time series can be deleted when the workload finishes or is removed, keeping cardinality bounded."That rationale is incorrect. CounterVec supports DeletePartialMatch exactly the same way. In fact, ClearClusterQueueMetrics at pkg/metrics/metrics.go:1038 already deletes PreemptedWorkloadsTotal (a CounterVec) via DeletePartialMatch.

Using a Gauge for a value that only ever increments is a Prometheus anti-pattern: it breaks rate()/increase() and the reset-detection semantics those functions rely on. Operators who try to chart preemption rate per workload will get wrong answers.

Please switch to CounterVec and update the help text. The variable should also be renamed accordingly (WorkloadPreemptionsTotal, with metric name workload_preemptions_total), matching both Prometheus naming convention and PreemptedWorkloadsTotal for ClusterQueue.

pkg/metrics/metrics.go
```diff
@@ -651,6 +655,21 @@ The label 'reason' can have the following values:
...
+			Help: `The cumulative number of times each workload has been preempted.
+The label 'reason' can have the following values:
+- "InClusterQueue" means that the workload was preempted by a workload in the same ClusterQueue.
+- "InCohortReclamation" means that the workload was preempted by a workload in the same cohort due to reclamation of nominal quota.
+- "InCohortFairSharing" means that the workload was preempted by a workload in the same cohort Fair Sharing.
+- "InCohortReclaimWhileBorrowing" means that the workload was preempted by a workload in the same cohort due to reclamation of nominal quota while borrowing.
+Uses a Gauge (rather than a Counter) so that the time series can be deleted when the workload finishes or is removed, keeping cardinality bounded.
+This gauge is deleted when the workload is completed or deleted.`,
+		}, []string{"namespace", "workload", "cluster_queue", "reason"},
+	)
```
amy, line 678: Please mirror PreemptedWorkloadsTotal. Add replica_role and extraLabels.

pkg/metrics/metrics.go
```diff
@@ -1321,6 +1352,7 @@ func Register() {
 		EvictedWorkloadsTotal,
 		EvictedWorkloadsOnceTotal,
 		PreemptedWorkloadsTotal,
+		WorkloadPreemptionsCount,
```
amy: As an example, per-LocalQueue metrics are gated by LocalQueueMetrics: pkg/metrics/metrics.go:1344, RegisterLQMetrics(). 

Per workload has an even higher cardinality. Please add a featuregate for per workload metrics.

pkg/controller/core/workload_controller.go
```diff
@@ -252,6 +252,7 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 
 	finishedCond := apimeta.FindStatusCondition(wl.Status.Conditions, kueue.WorkloadFinished)
 	if finishedCond != nil && finishedCond.Status == metav1.ConditionTrue {
+		metrics.ClearWorkloadPreemptionMetrics(wl.Namespace, wl.Name)
```
amy: When workloadRetention.afterFinished is configured, the workload object is retained for some duration after it finishes, but this call wipes its preemption history immediately on the first finished reconcile. An operator inspecting a retained finished workload (which is the whole point of the retention feature) will see zero preemptions even if it was preempted many times.

Move the metrics.ClearWorkloadPreemptionMetrics(...) call into the branch where we actually delete the workload (line 268, just before/after r.client.Delete), and rely on the Delete event handler at line 1141 for the no-retention path. That way the metric lifetime matches the object lifetime.

pkg/metrics/metrics.go
gyliu513: Makes sense to be consistent with PreemptedWorkloadsTotal.
Will rename WorkloadPreemptionsCount to WorkloadPreemptionsTotal.

pkg/metrics/metrics.go
gyliu513: Done

pkg/metrics/metrics.go
gyliu513: Will add replica_role and extraLabels to match PreemptedWorkloadsTotal's label set.

pkg/metrics/metrics.go
gyliu513: Agreed, per-workload metrics have higher cardinality than per-LocalQueue. Will add a new feature gate (WorkloadMetrics) following the same pattern as LocalQueueMetrics, starting Alpha/off by default, and gate WorkloadPreemptionsTotal registration behind it.

pkg/controller/core/workload_controller.go
gyliu513: Good catch. The metric lifetime should match the object lifetime , clearing immediately on the first finished reconcile defeats the purpose of workloadRetention.

gyliu513 (commented):
Thanks @amy , comments are addressed.

pkg/features/kube_features.go
```diff
@@ -414,6 +414,9 @@ const (
 	// (pod-group-name and prebuilt-workload-name) via annotations. When the gate is
 	// disabled, the label-based identifiers are used instead.
 	WorkloadIdentifierAnnotations featuregate.Feature = "WorkloadIdentifierAnnotations"
+
+	// Enable gathering of per-workload metrics.
+	WorkloadMetrics featuregate.Feature = "WorkloadMetrics"
```
amy, line 448: https://github.com/kubernetes-sigs/kueue/pull/11252#discussion_r3285358805
> Agreed, per-workload metrics have higher cardinality than per-LocalQueue. Will add a new feature gate (WorkloadMetrics) following the same pattern as LocalQueueMetrics, starting Alpha/off by default, and gate WorkloadPreemptionsTotal registration behind it.

cc/ @mimowo Can you check that featuregate is sufficient enough to gate this? Workload name is unbounded as part of the metric labels.

mimowo (commented):
cc @mszadkow ptal

pkg/features/kube_features.go
amy: also tagging @mszadkow here ^

sohankunkerkar:
@gyliu513 do you have cycles to push this to the finish line?

gyliu513:
@sohankunkerkar sorry for late, I may not able to help for this recently, please feel free to pick this up.

sohankunkerkar:
@gyliu513 thanks! I will work on this part.

sohankunkerkar:
@MaysaMacedo will take over this PR and push it to finish line.

gyliu513:
I added @MaysaMacedo to my repo https://github.com/gyliu513/kueue so that she can have permission to update this PR

Issue #8871, opened 2026-01-29
☂️ Integration with WAS in kube-scheduler


mimowo:
**What would you like to be added**:

Make sure Kueue is compatible with WAS, and can delegate node-level scheduling to kube-scheduler.

This is an umbrella issue with a lot of work we need to identify as WAS is graduating.

Some of the related items:
- https://github.com/kubernetes-sigs/kueue/issues/3755
- https://github.com/kubernetes-sigs/kueue/issues/8858

**Why is this needed**:

To leverage kube-scheduler primitives for:
- Gang Scheduling
- Dynamic Resource Allocation
- Topology-Aware Scheduling

mimowo:
cc @tenzen-y @gabesaba @sohankunkerkar @kannon92

kannon92:
For those that are interested, there is a weekly call on Mondays at 1 pm EST to sync on WAS.

Right now my hope is to get workload controller support for gangs but there is a lot of changes for WAS.

kannon92:
One area we may want to explore is using WAS to enforce gang scheduling for bare pods integration (https://kueue.sigs.k8s.io/docs/tasks/run/plain_pods/#running-a-group-of-pods-to-be-admitted-together).

I think most controllers should handle gang scheduling in their own way but this is the one integration that kueue may need to create its own PodGroups/Workloads so that the grouping of pods can be gang scheduled together.

tenzen-y:
> One area we may want to explore is using WAS to enforce gang scheduling for bare pods integration (https://kueue.sigs.k8s.io/docs/tasks/run/plain_pods/#running-a-group-of-pods-to-be-admitted-together).
> 
> I think most controllers should handle gang scheduling in their own way but this is the one integration that kueue may need to create its own PodGroups/Workloads so that the grouping of pods can be gang scheduled together.

Yes, exactly. But the discussion point is how to handle if controllers create PodGroups / Workloads.
We might be able to just skip handling such a group of Pods if we are not missing any stuff.

44past4:
I've drafted a proposal outlining the next steps for the WAS and Kueue integration: https://docs.google.com/document/d/13I3bMY-abBtsWmDhCl4RdDHKchHeB3G8zMSTpplmxHA/edit?usp=sharing

Please take a look and leave your feedback directly in the document.

mimowo:
cc @tenzen-y @gabesaba @kannon92 @sohankunkerkar ptal

mimowo:
[...]

PR #11255, opened 2026-05-16
Rename DRA feature gates to avoid conflict with upstream Kubernetes gates

### LINKED ISSUE #11248: Rename DynamicResourceAllocation to avoid conflict with k8s FGs

The name of the [feature gate](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/features/kube_features.go#L179C2-L179C27) is conflicting with [k8s core](https://github.com/kubernetes/kubernetes/blob/master/pkg/features/kube_features.go#L333C2-L333C27) which was observed as problematic when we started to prototype the integration via scheduler-library as part of https://github.com/kubernetes-sigs/kueue/issues/8871 (IIUC).

So I would propose to rename to `DRAIntegration` or `DynamicResourceAllocationIntegration`. 

We should probably also rename other FGs with `DRAIntegration` prefix rather than just `DRA` because k8s core is also using the DRA as prefix, see [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/features/kube_features.go#L192).

Another alternative is to have some generic approach, for example prefix all Kueue FGs with "Kueue" or all k8s core with "Kubernetes", but it might be tricky. For now, we probably need to solve the already known conflicts.

**Why is this needed**:

To avoid future breaking changes when we graduate the FG to Beta and want to start supporting DRA

sohankunkerkar:
#### What type of PR is this?
/kind cleanup

Renames DRA feature gates to avoid name collisions with upstream Kubernetes feature gates
(see #8871 for the scheduler-library integration where this became problematic):

| Old (deprecated in 0.18) | New | 
|---|---|
| `DynamicResourceAllocation` | `KueueDRAIntegration` |
| `DRAExtendedResources` | `KueueDRAIntegrationExtendedResource` |

#### Upgrade scenarios

| From | User config | What happens |
|---|---|---|
| **0.14–0.16** | `DynamicResourceAllocation: true` | Deprecation warning logged. Migrated → `KueueDRAIntegration=true`. DRA works. |
| **0.17** | `DynamicResourceAllocation: true, DRAExtendedResources: true` | Both deprecated warnings logged. Both migrated to new names. DRA works. |
| **0.17** | Updates config to `KueueDRAIntegration: true, KueueDRAIntegrationExtendedResource: true` | No migration needed. No warnings. DRA works. |
| **0.18 fresh** | `KueueDRAIntegration: true` | No migration needed. DRA works. |
| **Mixed** | `DynamicResourceAllocation: true, KueueDRAIntegration: false` (explicit) | Canonical gate wins (`ExplicitlySet` check). DRA disabled per user intent. |
| **Invalid** | `KueueDRAIntegrationExtendedResource: true` (no base gate) | `Validate()` rejects: "requires KueueDRAIntegration to be enabled". Startup fails. |

Fixes https://github.com/kubernetes-sigs/kueue/issues/11248

The e2e DRA config (`test/e2e/config/dra/controller_manager_config.yaml`) intentionally
keeps old gate names to exercise the migration path. The DRA e2e tests pass with this config.

```release-note
NONE
```

### DIFF
--- a/cmd/kueue/main.go
+++ b/cmd/kueue/main.go
@@ -299,7 +299,7 @@ func main() {
 		cacheOptions = append(cacheOptions, schdcache.WithResourceTransformations(cfg.Resources.Transformations))
 		queueOptions = append(queueOptions, qcache.WithResourceTransformations(cfg.Resources.Transformations))
 	}
-	if features.Enabled(features.DynamicResourceAllocation) && cfg.Resources != nil && len(cfg.Resources.DeviceClassMappings) > 0 {
+	if features.Enabled(features.KueueDRAIntegration) && cfg.Resources != nil && len(cfg.Resources.DeviceClassMappings) > 0 {
 		if err := dra.CreateMapperFromConfiguration(cfg.Resources.DeviceClassMappings); err != nil {
 			setupLog.Error(err, "Failed to initialize DRA mapper from configuration")
 			os.Exit(1)
--- a/pkg/features/kube_features.go
+++ b/pkg/features/kube_features.go
@@ -175,7 +175,10 @@ const (
 	// owner: @alaypatel07
 	// kep: https://github.com/kubernetes-sigs/kueue/tree/main/keps/2941-DRA
 	//
-	// Enable quota accounting for Dynamic Resource Allocation (DRA) devices in workloads
+	// Enable quota accounting for Dynamic Resource Allocation (DRA) devices in workloads.
+	KueueDRAIntegration featuregate.Feature = "KueueDRAIntegration"
+
+	// Deprecated: planned to be removed in 0.19. Use KueueDRAIntegration instead.
 	DynamicResourceAllocation featuregate.Feature = "DynamicResourceAllocation"
 
 	// owner: @sohankunkerkar
@@ -183,6 +186,9 @@ const (
 	//
 	// Enable extended resources support for DRA. Allows workloads to request DRA devices
 	// via standard resources.requests using DeviceClass extendedResourceName.
+	KueueDRAIntegrationExtendedResource featuregate.Feature = "KueueDRAIntegrationExtendedResource"
+
+	// Deprecated: planned to be removed in 0.19. Use KueueDRAIntegrationExtendedResource instead.
 	DRAExtendedResources featuregate.Feature = "DRAExtendedResources"
 
 	// owner: @MaysaMacedo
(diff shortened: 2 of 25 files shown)


cmd/kueue/main.go
```diff
@@ -171,6 +171,13 @@ func main() {
 		}
 	}
 
+	// Migrate deprecated DRA feature gates before validation so dependency
+	// checks see the migrated state. Remove in 0.19.
+	if err := features.MigrateDeprecatedDRAFeatureGates(setupLog); err != nil {
```
mimowo: LGTM, but since all DRA feature gates are Alpha we don't have any compatibility commitment to keep them. So, I would probably just rename them without worrying about the migration. wdyt?

Specifically the k8s guide says: https://kubernetes.io/docs/reference/using-api/deprecation-policy/

"Alpha API versions may be removed in any release without prior deprecation notice" I think this also counts for FGs.

cc @tenzen-y

sohankunkerkar: Sure, I'm happy to change this if @tenzen-y onboards with this idea.

tenzen-y: I agree with @mimowo .

I don't want to make such precedents (keeping backward compatibility for the alpha feature and API), which can cause misdirected user expectations.

mimowo (commented):
/lgtm
/approve
Thank you 👍

pkg/features/kube_features.go
```diff
@@ -484,11 +490,19 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 	TASBalancedPlacement: {
 		{Version: version.MustParse("0.15"), Default: false, PreRelease: featuregate.Alpha},
 	},
+	KueueDRAIntegration: {
+		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Alpha},
+	},
 	DynamicResourceAllocation: {
 		{Version: version.MustParse("0.14"), Default: false, PreRelease: featuregate.Alpha},
+		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Deprecated},
```
tenzen-y:
```suggestion
		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Deprecated, LockToDefault: true}, // remove in 0.19
```

pkg/features/kube_features.go
```diff
@@ -484,11 +490,19 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
...
+	},
+	KueueDRAIntegrationExtendedResource: {
+		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Alpha},
 	},
 	DRAExtendedResources: {
 		{Version: version.MustParse("0.17"), Default: false, PreRelease: featuregate.Alpha},
+		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Deprecated},
```
tenzen-y:
```suggestion
		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Deprecated, LockToDefault: true}, // remove in 0.19
```

tenzen-y (commented):
Can you lock deprecated FG and add a comment for the removal version?

tenzen-y (commented):
Thank you!
/lgtm
/approve

/hold cancel

tenzen-y:
I confirmed that the new commit has only formatting fix.
/lgtm

PR #11256, opened 2026-05-16
Fix stale skillsaw path so moved skills stay linted

weizhoublue:
#### What type of PR is this?

/kind bug

issue
 the skills was moved from cmd/experimental/agent/skills/ to cmd/experimental/skills/, any skillsaw lint run that relies  on .skillsaw.yaml still used the old glob in content-paths.
 
The skills files were no longer matched by the lint config, so make verify-skills-lint or make skills-lint could silently miss problems in the moved  skills directory.
  

Fixes #

```release-note

```

### DIFF
--- a/.skillsaw.yaml
+++ b/.skillsaw.yaml
@@ -187,7 +187,7 @@ exclude:
 
 # Additional markdown files to run content rules on (glob format)
 content-paths:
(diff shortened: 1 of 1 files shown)


mimowo (commented):
/lgtm
/approve
Thank you for keeping this aligned with reality

PR #11260, opened 2026-05-17
Account for KubeRay Redis cleanup resources

### LINKED ISSUE #10946: The Workload corresponding to RayService/RayCluster should reserve quota for RedisCleanup job
**What happened**:

The Workload corresponding to RedisCleanup Job does not allocate resources for the Job.

The quota allocated in the Workload covers all operations. Similarly as we currently cover the k8s submitter Job.

I haven't tested this myself, but there is configuration which triggers the rediscleanup job creation in RayCluster, eg.: https://github.com/ray-project/kuberay/blob/e3790d555b59c3a7a59f41fc0c6f8616ec5b9a45/ray-operator/apis/ray/v1/raycluster_types.go#L110

nerdeveloper:
#### What this PR does / why we need it
When KubeRay GCS fault tolerance is enabled, KubeRay creates a Redis cleanup Job during RayCluster deletion. Kueue previously built Workload PodSets only from the Ray head and workers, so the cleanup Job resources were not reserved.

This accounts for the cleanup Job's hardcoded 200m CPU and 256Mi memory requests by folding those requests into the Ray head PodSet whenever KubeRay GCS fault tolerance is enabled. This keeps the Workload PodSet count aligned with KubeRay's head/worker PodSets while reserving enough quota for the cleanup Job.

#### Which issue(s) this PR fixes
Fixes #10946

#### Special notes for your reviewer
Verified with a local Kind repro that the original Workload under-accounted Redis cleanup resources when GCS fault tolerance was enabled. The PR now has unit coverage for RayCluster, RayJob, and RayService GCS fault-tolerance accounting.

AI assistance was used in preparing this PR.

```release-note
RayJob, RayCluster, and RayServe integrations: Fixed missing quota accounting for Redis cleanup resources when GCS fault tolerance is enabled. Kueue accounts for the Redis cleanup Job resources for workloads by folding the cleanup Job requests into the Ray head PodSet.
```

#### Additional documentation
None

## Summary by CodeRabbit

* **New Features**
  * Added support for GCS fault tolerance configuration in Ray clusters with automatic Redis cleanup resource management.

* **Improvements**
  * Enhanced PodSet validation logic for improved accuracy in resource scheduling.
  * Strengthened test coverage for fault tolerance scenarios.

### DIFF
--- a/pkg/controller/jobs/raycluster/common.go
+++ b/pkg/controller/jobs/raycluster/common.go
@@ -19,11 +19,15 @@ package raycluster
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strconv"
 
 	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	rayutils "github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/validation/field"
@@ -36,6 +40,7 @@ import (
 	"sigs.k8s.io/kueue/pkg/features"
 	"sigs.k8s.io/kueue/pkg/podset"
 	utilpodset "sigs.k8s.io/kueue/pkg/util/podset"
+	utilresource "sigs.k8s.io/kueue/pkg/util/resource"
 	"sigs.k8s.io/kueue/pkg/workloadslicing"
 )
 
@@ -51,8 +56,8 @@ const (
 	RayClusterGenerationAnnotation = "kueue.x-k8s.io/raycluster-generation"
 )
 
-// BuildPodSets builds PodSets from RayClusterSpec
-func BuildPodSets(rayClusterSpec *rayv1.RayClusterSpec) ([]kueue.PodSet, error) {
+// BuildPodSets builds PodSets from RayClusterSpec.
+func BuildPodSets(rayClusterSpec *rayv1.RayClusterSpec, annotations map[string]string) ([]kueue.PodSet, error) {
 	podSets := make([]kueue.PodSet, 0)
 
 	// head
@@ -68,6 +73,11 @@ func BuildPodSets(rayClusterSpec *rayv1.RayClusterSpec) ([]kueue.PodSet, error) 
 			return nil, err
 		}
 		headPodSet.TopologyRequest = topologyRequest
+	}
+	if shouldAccountForRedisCleanup(rayClusterSpec, annotations) {
+		if err := accountForRedisCleanupInHeadPodSet(&headPodSet); err != nil {
+			return nil, err
+		}
 	}
 	podSets = append(podSets, headPodSet)
 
@@ -97,6 +107,33 @@ func BuildPodSets(rayClusterSpec *rayv1.RayClusterSpec) ([]kueue.PodSet, error) 
 	}
 
 	return podSets, nil
+}
+
+func accountForRedisCleanupInHeadPodSet(headPodSet *kueue.PodSet) error {
+	if len(headPodSet.Template.Spec.Containers) <= rayutils.RayContainerIndex {
+		return errors.New("cannot account for Redis cleanup resources: head pod template must include the Ray container")
+	}
+
+	headContainer := &headPodSet.Template.Spec.Containers[rayutils.RayContainerIndex]
+	headContainer.Resources.Requests = utilresource.MergeResourceListKeepMax(headContainer.Resources.Requests, redisCleanupResourceRequests())
+	return nil
+}
+
+func redisCleanupResourceRequests() corev1.ResourceList {
+	// KubeRay hardcodes the Redis cleanup Job CPU and memory requests/limits:
+	// https://github.com/ray-project/kuberay/blob/24442570686d81b9e056315bd08df689887a0d8c/ray-operator/controllers/ray/raycluster_controller.go#L1481
+	return corev1.ResourceList{
+		corev1.ResourceCPU:    resource.MustParse("200m"),
+		corev1.ResourceMemory: resource.MustParse("256Mi"),
+	}
+}
+
+func shouldAccountForRedisCleanup(rayClusterSpec *rayv1.RayClusterSpec, annotations map[string]string) bool {
+	return rayutils.IsGCSFaultToleranceEnabled(rayClusterSpec, annotations)
+}
+
+func ExpectedPodSetsCount(rayClusterSpec *rayv1.RayClusterSpec) int {
+	return len(rayClusterSpec.WorkerGroupSpecs) + 1
 }
 
 func UpdatePodSets(ctx context.Context, podSets []kueue.PodSet, c client.Client, object client.Object, enableInTreeAutoscaling *bool, rayClusterName string) ([]kueue.PodSet, error) {
@@ -216,9 +253,9 @@ func ValidateCreate(object client.Object, rayClusterSpec *rayv1.RayClusterSpec, 
 		)
 	}
 
-	// Should limit the worker count to max PodSets minus the cluster head.
-	if len(rayClusterSpec.WorkerGroupSpecs) > jobframework.MaxPodSets-1 {
-		allErrors = append(allErrors, field.TooMany(rayClusterSpecPath.Child("workerGroupSpecs"), len(rayClusterSpec.WorkerGroupSpecs), jobframework.MaxPodSets-1))
+	// Should limit the generated PodSet count to the maximum supported by Workloads.
+	if podSetsCount := ExpectedPodSetsCount(rayClusterSpec); podSetsCount > jobframework.MaxPodSets {
+		allErrors = append(allErrors, field.TooMany(rayClusterSpecPath.Child("workerGroupSpecs"), podSetsCount, jobframework.MaxPodSets))
 	}
 
 	// None of the workerGroups should be named "head"
--- a/pkg/controller/jobs/raycluster/common_test.go
+++ b/pkg/controller/jobs/raycluster/common_test.go
@@ -25,8 +25,10 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	rayutils "github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/validation/field"
@@ -36,6 +38,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
 
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+	"sigs.k8s.io/kueue/pkg/controller/jobframework"
 	"sigs.k8s.io/kueue/pkg/features"
 	"sigs.k8s.io/kueue/pkg/podset"
 	utiltestingapi "sigs.k8s.io/kueue/pkg/util/testing/v1beta2"
@@ -46,6 +49,7 @@ import (
 func TestBuildPodSets(t *testing.T) {
 	testCases := map[string]struct {
 		rayClusterSpec *rayv1.RayClusterSpec
+		annotations    map[string]string
 		wantPodSets    []kueue.PodSet
 		wantErr        bool
 	}{
@@ -199,11 +203,115 @@ func TestBuildPodSets(t *testing.T) {
 					Obj(),
 			},
 		},
+		"spec with gcs fault tolerance": {
+			rayClusterSpec: &rayv1.RayClusterSpec{
+				GcsFaultToleranceOptions: &rayv1.GcsFaultToleranceOptions{
+					RedisAddress: "redis:6379",
+				},
+				HeadGroupSpec: rayv1.HeadGroupSpec{
+					Template: corev1.PodTemplateSpec{
+						ObjectMeta: metav1.ObjectMeta{
+							Labels: map[string]string{"ray.io/cluster": "raycluster"},
+						},
+						Spec: corev1.PodSpec{
+							Containers: []corev1.Container{{
+								Name:  "head",
+								Image: "rayproject/ray:2.0.0",
+								Resources: corev1.ResourceRequirements{
+									Requests: corev1.ResourceList{
+										corev1.ResourceCPU: resource.MustParse("1"),
+									},
+								},
+							}},
+						},
+					},
+				},
+				WorkerGroupSpecs: []rayv1.WorkerGroupSpec{
+					{
+						GroupName: "workers",
+						Replicas:  ptr.To[int32](1),
+						Template: corev1.PodTemplateSpec{
+							Spec: corev1.PodSpec{
+								Containers: []corev1.Container{{Name: "worker"}},
+							},
+						},
+					},
+				},
+			},
+			wantPodSets: []kueue.PodSet{
+				*utiltestingapi.MakePodSet(headGroupPodSetName, 1).
+					Labels(map[string]string{"ray.io/cluster": "raycluster"}).
+					PodSpec(corev1.PodSpec{
+						Containers: []corev1.Container{{
+							Name:  "head",
+							Image: "rayproject/ray:2.0.0",
+							Resources: corev1.ResourceRequirements{
+								Requests: corev1.ResourceList{
+									corev1.ResourceCPU:    resource.MustParse("1"),
+									corev1.ResourceMemory: resource.MustParse("256Mi"),
+								},
+							},
+						}},
+					}).
+					Obj(),
+				*utiltestingapi.MakePodSet("workers", 1).
+					PodSpec(corev1.PodSpec{
+						Containers: []corev1.Container{{Name: "worker"}},
+					}).
+					Obj(),
+			},
+		},
+		"spec with gcs fault tolerance annotation": {
+			annotations: map[string]string{rayutils.RayFTEnabledAnnotationKey: "true"},
+			rayClusterSpec: &rayv1.RayClusterSpec{
+				HeadGroupSpec: rayv1.HeadGroupSpec{
+					Template: corev1.PodTemplateSpec{
+						ObjectMeta: metav1.ObjectMeta{
+							Labels: map[string]string{"ray.io/cluster": "raycluster"},
+						},
+						Spec: corev1.PodSpec{
+							Containers: []corev1.Container{{
+								Name:  "head",
+								Image: "rayproject/ray:2.0.0",
+							}},
+						},
+					},
+				},
+			},
+			wantPodSets: []kueue.PodSet{
+				*utiltestingapi.MakePodSet(headGroupPodSetName, 1).
+					Labels(map[string]string{"ray.io/cluster": "raycluster"}).
+					PodSpec(corev1.PodSpec{
+						Containers: []corev1.Container{{
+							Name:  "head",
+							Image: "rayproject/ray:2.0.0",
+							Resources: corev1.ResourceRequirements{
+								Requests: corev1.ResourceList{
+									corev1.ResourceCPU:    resource.MustParse("200m"),
+									corev1.ResourceMemory: resource.MustParse("256Mi"),
+								},
+							},
+						}},
+					}).
+					Obj(),
+			},
+		},
+		"spec with gcs fault tolerance and missing head container": {
+			rayClusterSpec: &rayv1.RayClusterSpec{
+				GcsFaultToleranceOptions: &rayv1.GcsFaultToleranceOptions{
+					RedisAddress: "redis:6379",
+				},
+				HeadGroupSpec: rayv1.HeadGroupSpec{
+					Template: corev1.PodTemplateSpec{},
+				},
+			},
+			wantErr: true,
+		},
 	}
 
 	for name, tc := range testCases {
 		t.Run(name, func(t *testing.T) {
-			gotPodSets, err := BuildPodSets(tc.rayClusterSpec)
+			gotPodSets, err := BuildPodSets(tc.rayClusterSpec, tc.annotations)
 
 			if tc.wantErr {
 				if err == nil {
@@ -824,7 +932,7 @@ func TestValidateCreateRayClusterSpec(t *testing.T) {
 				},
 			},
 			wantErrors: field.ErrorList{
-				field.TooMany(field.NewPath("spec", "workerGroupSpecs"), 10, 9),
+				field.TooMany(field.NewPath("spec", "workerGroupSpecs"), 11, jobframework.MaxPodSets),
 			},
 		},
 		"worker group named 'head'": {
@@ -865,7 +973,7 @@ func TestValidateCreateRayClusterSpec(t *testing.T) {
 			},
 			wantErrors: field.ErrorList{
 				field.Invalid(field.NewPath("spec", "enableInTreeAutoscaling"), new(true), "a kueue managed job should only use autoscaling when workload slicing is enabled"),
-				field.TooMany(field.NewPath("spec", "workerGroupSpecs"), 10, 9),
+				field.TooMany(field.NewPath("spec", "workerGroupSpecs"), 11, jobframework.MaxPodSets),
 				field.Forbidden(field.NewPath("spec", "workerGroupSpecs").Index(0).Child("groupName"), fmt.Sprintf("%q is reserved for the head group", headGroupPodSetName)),
 			},
 		},
--- a/pkg/controller/jobs/rayjob/rayjob_controller_test.go
+++ b/pkg/controller/jobs/rayjob/rayjob_controller_test.go
@@ -85,6 +85,50 @@ func TestPodSets(t *testing.T) {
 			},
 			featureGates: map[featuregate.Feature]bool{features.TopologyAwareScheduling: false},
 		},
+		"with gcs fault tolerance": {
+			rayJob: func() *RayJob {
+				job := testingrayutil.MakeJob("rayjob", "ns").
+					WithHeadGroupSpec(
+						rayv1.HeadGroupSpec{
+							Template: corev1.PodTemplateSpec{
+								Spec: corev1.PodSpec{Containers: []corev1.Container{{Name: "head_c"}}},
+							},
+						},
+					).
+					WithWorkerGroups(
+						rayv1.WorkerGroupSpec{
+							GroupName: "group1",
+							Template: corev1.PodTemplateSpec{
+								Spec: corev1.PodSpec{Containers: []corev1.Container{{Name: "group1_c"}}},
+							},
+						},
+					).
+					Obj()
+				job.Spec.RayClusterSpec.GcsFaultToleranceOptions = &rayv1.GcsFaultToleranceOptions{RedisAddress: "redis:6379"}
+				return (*RayJob)(job)
+			}(),
+			wantPodSets: func(rayJob *RayJob) []kueue.PodSet {
+				return []kueue.PodSet{
+					*utiltestingapi.MakePodSet(headGroupPodSetName, 1).
+						PodSpec(corev1.PodSpec{
+							Containers: []corev1.Container{{
+								Name: "head_c",
+								Resources: corev1.ResourceRequirements{
+									Requests: corev1.ResourceList{
+										corev1.ResourceCPU:    resource.MustParse("200m"),
+										corev1.ResourceMemory: resource.MustParse("256Mi"),
+									},
+								},
+							}},
+						}).
+						Obj(),
+					*utiltestingapi.MakePodSet("group1", 1).
+						PodSpec(*rayJob.Spec.RayClusterSpec.WorkerGroupSpecs[0].Template.Spec.DeepCopy()).
+						Obj(),
+				}
+			},
+			featureGates: map[featuregate.Feature]bool{features.TopologyAwareScheduling: false},
+		},
 		"with required topology annotation": {
 			rayJob: (*RayJob)(testingrayutil.MakeJob("rayjob", "ns").
 				WithHeadGroupSpec(
--- a/pkg/controller/jobs/raycluster/raycluster_controller.go
+++ b/pkg/controller/jobs/raycluster/raycluster_controller.go
@@ -102,54 +102,11 @@ func (j *RayCluster) PodLabelSelector() string {
 }
 
 func (j *RayCluster) PodSets(ctx context.Context, _ client.Client) ([]kueue.PodSet, error) {
-	// len = workerGroups + head
-	podSets := make([]kueue.PodSet, len(j.Spec.WorkerGroupSpecs)+1)
-
-	// head
-	podSets[0] = kueue.PodSet{
-		Name:     headGroupPodSetName,
-		Template: *j.Spec.HeadGroupSpec.Template.DeepCopy(),
-		Count:    1,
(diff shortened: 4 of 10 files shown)


nerdeveloper:
Hi @mimowo, could you please take a look when you have a chance?

pkg/controller/jobs/raycluster/common.go
```diff
@@ -96,9 +103,66 @@ func BuildPodSets(rayClusterSpec *rayv1.RayClusterSpec) ([]kueue.PodSet, error)
...
+		},
+		Requests: corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("200m"),
+			corev1.ResourceMemory: resource.MustParse("256Mi"),
+		},
+	}
+}
+
+func hasGCSFaultTolerance(rayClusterSpec *rayv1.RayClusterSpec) bool {
+	return rayClusterSpec.GcsFaultToleranceOptions != nil
```
mimowo: Could you check with the KubeRay code is this is a sufficient check? I remember that there was also a knob at the kuberay deployment that would be the global enabler, which is disabled would make this check irrelevant.

pkg/controller/jobs/raycluster/common.go
```diff
@@ -96,9 +103,66 @@ func BuildPodSets(rayClusterSpec *rayv1.RayClusterSpec) ([]kueue.PodSet, error)
 		podSets = append(podSets, workerPodSet)
 	}
 
+	if hasGCSFaultTolerance(rayClusterSpec) {
```
mimowo: Since this is a delicate change let's introduce a feature gate as a bailout option, say `KubeRayAccountForRedisCleanup` which is Beta, and add a comment o GA in 0.21 in Kueue.

mimowo (commented):
cc @yaroslava-serdiuk ptal

pkg/controller/jobs/raycluster/common.go
nerdeveloper: Checked against KubeRay v1.6.1. The cleanup path is gated by `ENABLE_GCS_FT_REDIS_CLEANUP != "false"` and `utils.IsGCSFaultToleranceEnabled(&instance.Spec, instance.Annotations)`. I updated Kueue to mirror the observable part of that predicate by using `rayutils.IsGCSFaultToleranceEnabled(spec, annotations)`, and added the Kueue-side feature gate as a bailout for clusters where the KubeRay operator env disables cleanup.

pkg/controller/jobs/raycluster/common.go
nerdeveloper: Done in the latest push. I added `KubeRayAccountForRedisCleanup` as Beta, default-on, with the GA-in-0.21 comment and regenerated the feature-gate YAML. I also added coverage for disabling the gate, plus the annotation-based GCS FT path. Local verification: `go test ./pkg/features ./pkg/controller/jobs/raycluster ./pkg/controller/jobs/rayservice ./pkg/controller/jobs/rayjob` and `make verify-ci-lint` both pass.

yaroslava-serdiuk:
So here we represent a RedisCleanupJob as an additional PodSet in the Workload and so decreasing the max number of workers group in the RayCluster from 7 to 6. I'm not sure if this is an issue, but it might be. 
Since the RedisCleanupJob requests are small (CPU 200m and Memory 265Mi) I propose to just add those request to the head pod. 
@mimowo WDYT?

pkg/controller/jobs/raycluster/common.go
```diff
@@ -40,6 +44,9 @@ import (
 )
 
 const (
+	redisCleanupPodSetName = kueue.PodSetReference(rayv1.RedisCleanupNode)
+	maxRayClusterPodSets   = 8
```
yaroslava-serdiuk: It's a kueue limitation, not a raycluster, so let's put it in the constants file and name it MaxPodSetsSize or something similar.

pkg/controller/jobs/raycluster/common.go
```diff
@@ -96,9 +103,66 @@ func BuildPodSets(rayClusterSpec *rayv1.RayClusterSpec) ([]kueue.PodSet, error)
...
+
+func ExpectedPodSetsCount(rayClusterSpec *rayv1.RayClusterSpec, annotations map[string]string) int {
+	count := len(rayClusterSpec.WorkerGroupSpecs) + 1
+	if shouldAccountForRedisCleanup(rayClusterSpec, annotations) {
+		count++
+	}
+	return count
+}
+
+func maxWorkerGroupsForSpec(rayClusterSpec *rayv1.RayClusterSpec, annotations map[string]string) int {
```
yaroslava-serdiuk: I don't think we need both ExpectedPodSetsCount and maxWorkerGroupsForSpec, since they are counting the same thing.
For example, instead of limiting workers in 284, we can just verify that ExpectedPodSetsCount is <= maxPodSetsSize (8).

pkg/controller/jobs/raycluster/common_test.go
```diff
@@ -224,6 +306,118 @@ func TestBuildPodSets(t *testing.T) {
...
+			}).
+			Obj(),
+	}
+
+	if diff := cmp.Diff(wantPodSets, gotPodSets, cmpopts.IgnoreFields(kueue.PodSet{}, "TopologyRequest")); diff != "" {
+		t.Errorf("Unexpected podSets (-want +got):\n%s", diff)
+	}
+}
+
+func TestBuildPodSetsWithGCSFaultToleranceAnnotation(t *testing.T) {
```
yaroslava-serdiuk: Let's merge those two cases and have one table-driven test that verifies combinations for feature gate enabled/disabled and GCSFaultTolerance present/not present .

pkg/controller/jobs/raycluster/common_test.go
```diff
@@ -884,6 +1123,34 @@ func TestValidateCreateRayClusterSpec(t *testing.T) {
 	}
 }
 
+func TestValidateCreateRayClusterSpecWithRedisCleanupFeatureGateDisabled(t *testing.T) {
```
yaroslava-serdiuk: Instead of introducing a separate test, let's include this test as a case of the test above.

yaroslava-serdiuk (changes requested):

nerdeveloper:
> So here we represent a RedisCleanupJob as an additional PodSet in the Workload and so decreasing the max number of workers group in the RayCluster from 7 to 6. I'm not sure if this is an issue, but it might be. Since the RedisCleanupJob requests are small (CPU 200m and Memory 265Mi) I propose to just add those request to the head pod. @mimowo WDYT?

@mimowo what do you think?

mimowo:
> > So here we represent a RedisCleanupJob as an additional PodSet in the Workload and so decreasing the max number of workers group in the RayCluster from 7 to 6. I'm not sure if this is an issue, but it might be. Since the RedisCleanupJob requests are small (CPU 200m and Memory 265Mi) I propose to just add those request to the head pod. @mimowo WDYT?
> 
> @mimowo what do you think?

Interesting idea, but the split feels more natural to me, it is also the approach we took for the SubmitterJob. Since we introduce the fix behind a FG we could mitigate the problem.

Let me also check with @ns-sundar and @hiboyang - can you share how many PodSets you use in practice?

I think 8 PodSets is pretty constraining and I would rather expand this.

mimowo:
We are currently working on getting the PodSet count to 10 so this should mitigate the concern : https://github.com/kubernetes-sigs/kueue/pull/11388

 let's finalize the other remaining comments

nerdeveloper:
@mimowo we are waiting for @ns-sundar and @hiboyang right/

mimowo:
> @mimowo we are waiting for @ns-sundar and @hiboyang right/

I was asking them to explore if they assess that 5 PodSets for workers would be enough, but since we already increase the PodSet count 8->10, then the relevance of the question is decreased. I think we can proceed without the answer, but I will ping them on slack to take a look also asynchronously.

pkg/controller/jobs/raycluster/common.go
ns-sundar: +1

pkg/controller/jobs/raycluster/common.go
```diff
@@ -96,9 +103,66 @@ func BuildPodSets(rayClusterSpec *rayv1.RayClusterSpec) ([]kueue.PodSet, error)
...
 }
 
+func buildRedisCleanupPodSet(rayClusterSpec *rayv1.RayClusterSpec) kueue.PodSet {
+	template := *rayClusterSpec.HeadGroupSpec.Template.DeepCopy()
+	template.Labels = maps.Clone(template.Labels)
+	if template.Labels == nil {
+		template.Labels = make(map[string]string, 1)
+	}
+	template.Labels[rayutils.RayNodeTypeLabelKey] = string(rayv1.RedisCleanupNode)
+	if len(template.Spec.Containers) > 0 {
```
ns-sundar: Why would the length be not > 0 in normal conditions?
If the length is 0, we should treat it as an error, log it, and probably not return a regular podset.

pkg/controller/jobs/raycluster/common.go
nerdeveloper: Done. Removed `maxWorkerGroupsForSpec`; validation now checks `ExpectedPodSetsCount(...)` against `constants.MaxPodSetsSize`.

pkg/controller/jobs/raycluster/common_test.go
nerdeveloper: Done. Folded these cases into the `TestBuildPodSets` table, covering GCS absent/present, the feature gate enabled/disabled path, the annotation path, and the missing Ray container error.

pkg/controller/jobs/raycluster/common_test.go
nerdeveloper: Done. Moved the disabled feature gate validation coverage into `TestValidateCreateRayClusterSpec`.

pkg/controller/jobs/raycluster/common.go
nerdeveloper: Done. Moved the shared PodSet cap to `pkg/controller/constants.MaxPodSetsSize` and updated the Ray validation paths to use it.

pkg/controller/jobs/raycluster/common.go
nerdeveloper: Done. `buildRedisCleanupPodSet` now returns an error when the head template does not include the Ray container, so `BuildPodSets` does not return a regular cleanup PodSet for that malformed input. Added coverage for this case.

pkg/controller/constants/constants.go
```diff
@@ -26,22 +26,16 @@ const (
...
-	// When enabled, we use the value of pre-built workload from the annotation by default;
-	// if it is not present, we fall back to the label.
-	PrebuiltWorkloadAnnotation = "kueue.x-k8s.io/prebuilt-workload-name"
-
 	// JobUIDLabel is the label key in the workload resource, that holds the UID of
 	// the owner job.
 	JobUIDLabel = "kueue.x-k8s.io/job-uid"
 
+	// MaxPodSetsSize is the maximum number of PodSets supported for a Workload.
+	MaxPodSetsSize = 8
```
mimowo: iiuc when we merge https://github.com/kubernetes-sigs/kueue/pull/11388 we can bump this to 10

hiboyang:
> > @mimowo we are waiting for @ns-sundar and @hiboyang right/
> 
> I was asking them to explore if they assess that 5 PodSets for workers would be enough, but since we already increase the PodSet count 8->10, then the relevance of the question is decreased. I think we can proceed without the answer, but I will ping them on slack to take a look also asynchronously.

Thanks for ping me on slack! We do not use large number of podsets. Since we increase from 8 to 10, I think it should be good.

pkg/controller/jobs/raycluster/common.go
```diff
@@ -96,9 +104,63 @@ func BuildPodSets(rayClusterSpec *rayv1.RayClusterSpec) ([]kueue.PodSet, error)
...
+	template.Spec.RestartPolicy = corev1.RestartPolicyNever
+
+	return kueue.PodSet{
+		Name:     redisCleanupPodSetName,
+		Template: template,
+		Count:    1,
+	}, nil
+}
+
+func redisCleanupResourceRequirements() corev1.ResourceRequirements {
```
hiboyang: Better add a reference link from KubeRay, so people understand the cpu/memory values are hardcoded in KubeRay: https://github.com/ray-project/kuberay/blob/24442570686d81b9e056315bd08df689887a0d8c/ray-operator/controllers/ray/raycluster_controller.go#L1481

hiboyang:
Hi @nerdeveloper, thanks working on this! Would you help to double check when a RayJob terminates, will it delete pods before create the Redis cleanup job? If yes, the extra cpu/memory of the Redis cleanup job may not be needed to be counted into the RayJob, since the deleted pod will release resources first.

pkg/controller/jobs/raycluster/common.go
nerdeveloper: Done. Added the KubeRay source link next to the hardcoded Redis cleanup resource requirements.

nerdeveloper:
@hiboyang I checked KubeRay v1.6.1 for the RayJob path. When a RayJob reaches a terminal state with `ShutdownAfterJobFinishes`, KubeRay deletes the associated RayCluster. The RayCluster GCS FT finaliser then deletes the head and worker pods, waits for the head pod to be terminated, and only after that creates the Redis cleanup Job.

So I updated this PR to skip Redis cleanup resource accounting for RayJob only. RayCluster/RayService still accounts for the Redis cleanup PodSet. I also added the KubeRay source link for the hardcoded `200m`/`256Mi` values.

hiboyang:
> @hiboyang I checked KubeRay v1.6.1 for the RayJob path. When a RayJob reaches a terminal state with `ShutdownAfterJobFinishes`, KubeRay deletes the associated RayCluster. The RayCluster GCS FT finaliser then deletes the head and worker pods, waits for the head pod to be terminated, and only after that creates the Redis cleanup Job.
> 
> So I updated this PR to skip Redis cleanup resource accounting for RayJob only. RayCluster/RayService still accounts for the Redis cleanup PodSet. I also added the KubeRay source link for the hardcoded `200m`/`256Mi` values.

Hi @nerdeveloper , thanks for checking! For RayCluster/RayService, is head/worker pod deletion happening before or after Redis cleanup job?

mimowo:
> Hi @nerdeveloper , thanks for checking! For RayCluster/RayService, is head/worker pod deletion happening before or after Redis cleanup job?
 
 Since users can mutate the requests for Head and Worker they may be requesting other resources than the redis job anyway, for example gpu, while the redis job needs cpu.

pkg/controller/constants/constants.go
yaroslava-serdiuk: Also, the constant is already defined in pkg/controller/jobframework/constants.go in https://github.com/kubernetes-sigs/kueue/pull/11388, so let's remove the constant here

yaroslava-serdiuk:
> > @hiboyang I checked KubeRay v1.6.1 for the RayJob path. When a RayJob reaches a terminal state with `ShutdownAfterJobFinishes`, KubeRay deletes the associated RayCluster. The RayCluster GCS FT finaliser then deletes the head and worker pods, waits for the head pod to be terminated, and only after that creates the Redis cleanup Job.
> > So I updated this PR to skip Redis cleanup resource accounting for RayJob only. RayCluster/RayService still accounts for the Redis cleanup PodSet. I also added the KubeRay source link for the hardcoded `200m`/`256Mi` values.
> 
> Hi @nerdeveloper , thanks for checking! For RayCluster/RayService, is head/worker pod deletion happening before or after Redis cleanup job?

The Redis Cleanup pod cleanup the RayCluster data from a Redis instance, so the behavior should be the same for RayJob, RayService and RayCluster. 
I agree that technically we can ignore Redis cleanup job resources, since it works already. But I think the purpose of this PR is to be accurate in resource counting. I suggest to use max function to be sure, that the RedisCleanupJob doesn't request extra resources.

mimowo:
Maybe we could combine the RaySubmitter job and RayRedisCleanup into One PodSet using the max, and call the PodSet say "AuxiliaryJobs", wdyt?

yaroslava-serdiuk:
> Maybe we could combine the RaySubmitter job and RayRedisCleanup into One PodSet using the max, and call the PodSet say "AuxiliaryJobs", wdyt?


I'd lean against combining them. They serve different scopes and run at different times: RaySubmitter belongs to the RayJob (it runs when the RayJob is created and IIUC it runs alongside head and workers), whereas RayRedisCleanup belongs to the RayCluster (it runs when the RayCluster resources are deleted).

In short, for Submitter job we should add resources to the RayJob/RayCluster.
For RedisCleanupJob we should use max of its resources and RayCluster resources if we want to be accurate.

mimowo:
> For RedisCleanupJob we should use max of its resources and RayCluster resources if we want to be accurate.

What do you mean by "RayCluster resources" here, can you rephrase / provide some pointers? Are we already accounting somehow for the RayCluster resources?

yaroslava-serdiuk:
By RayCluster resources I mean head pod and working pods

mimowo:
> By RayCluster resources I mean head pod and working pods

I see, so your proposal is to merge RedisCleanup job as part of the Head? This could technically work, but I think the Head is more likely to be one of the workers requesting GPU. The RedisCleanup job seems to me conceptually as SubmitterJob -small auxiliary job which only needs cpu / mem.

mimowo:
Let me summarize the options:
1. head, raySubmitter, redisCleanup - are all separate podsets (as currently)
2. max(head, redisCleanup), raySubtmitter
3. max(head, redisCleanup, raySubmitter)  
4. head, max(redisCleanup + raySubmitter)

My preferences are: 
1. **pros**: clear separation, **cons**:  potentially declaring slightly too much cpu mem, because we cannot use max
2. **cons**:  because it assumes a different strategy for redisCleanup and raySubtmitter which are both auxiliary jobs
3. **pros**: best resource usage, cons: mixes head with auxiliary jobs
4. **pros**: separation of auxiliary jobs from head, **cons**: potentially declaring slightly too much cpu and mem

My preference is either (1) as the simplest model.

Actually all options (2) - (4.) are performance optimizations which could only be used if we can prove that the job lifecycle overlaps or not. So for them we need to have an analysis of lifecycle.

yaroslava-serdiuk:
> Let me summarize the options:
> 
> 1. head, raySubmitter, redisCleanup - are all separate podsets (as currently)
> 2. max(head, redisCleanup), raySubtmitter
> 3. max(head, redisCleanup, raySubmitter)
> 4. head, max(redisCleanup + raySubmitter)
> 
> My preferences are:
> 
> 1. **pros**: clear separation, **cons**:  potentially declaring slightly too much cpu mem, because we cannot use max
> 2. **cons**:  because it assumes a different strategy for redisCleanup and raySubtmitter which are both auxiliary jobs
> 3. **pros**: best resource usage, cons: mixes head with auxiliary jobs
> 4. **pros**: separation of auxiliary jobs from head, **cons**: potentially declaring slightly too much cpu and mem
> 
> My preference is either (1) as the simplest model.
> 
> Actually all options (2) - (4.) are performance optimizations which could only be used if we can prove that the job lifecycle overlaps or not. So for them we need to have an analysis of lifecycle.

I'm fine with option 1) and I suggested option 2) as an optimisation since in option 1 we overcounting requests a bit. 
Option 3) is not correct, because raySubmitter runs next to head pod, so we should count the raySubmitter requests. 

There is also option 5) head, raySubmitter. (ignoring redis cleanup job). That's what we have now and it works fine.

mimowo:
im not sure it works fine, this is the point. it works most of the time, but will this work well if head and workers declare gpu. Then we would not allocate cpu in any podset currently.

hiboyang:
Just brainstorm idea for another option, e.g. treating this like an autoscaling scenario... when the Redis cleanup job starts, check whether current workload for the RayJob/RayService/RayCluster is good to accommodate the job? at that time, head/worker pods may be already deleted, and some resources are freed from the workload, if the freed resources are larger than the Redis cleanup job, Kueue could just admit it directly.

mimowo:
Actually, today I synced with @yaroslava-serdiuk on this and we reached alignment which is basically to follow Yarsolava's suggestion and combine the resources for Redis into the head, as max(). 

The reason is because the head is suspended before the Redis is started. Also, the RaySubmitted job only works for RayJobs while the RedisCleanup job may work also work for RayService or RayCluster.

mimowo:
@nerdeveloper would you have cycles to adjust the PR according to the new recommendation?

nerdeveloper:
yes, i will take a look

nerdeveloper:
Updated the PR based on the latest review thread.

Changes made:
- Rebased on current `main` and resolved the merge conflict.
- Folded Redis cleanup CPU/memory requests into the Ray head PodSet with max-request accounting instead of adding a separate Redis cleanup PodSet.
- Switched the PodSet limit checks to the shared `jobframework.MaxPodSets` constant from main.
- Kept the KubeRay source link next to the hardcoded cleanup resource values.

nerdeveloper:
The two MultiKueue e2e failures were not test failures. Both jobs failed before running e2e because Prow invoked Make targets that are missing after the recent MultiKueue split:

- `test-multikueue-baseline-e2e-main`
- `test-multikueue-extended-e2e-main`

I added small compatibility aliases that delegate to the current targets:
- `test-multikueue-e2e-baseline`
- `test-multikueue-e2e-extended`

let me know if this makes sense @mimowo

mimowo:
@nerdeveloper please rebase so that we can see if the tests are passing

mimowo:
@nerdeveloper do you have cycles to finalize the PR?

nerdeveloper:
@mimowo, should I follow this CodeRabbit review?

mimowo:
> @mimowo, should I follow this CodeRabbit review?

This is pretty new so I don't have yet a good intuition about the quality of the suggestions, but I wouldn't follow it blindly. I will try to review EOW.

pkg/features/kube_features.go
```diff
@@ -662,6 +668,9 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 	ConcurrentAdmission: {
 		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Alpha},
 	},
+	KubeRayAccountForRedisCleanup: {
+		{Version: version.MustParse("0.18"), Default: true, PreRelease: featuregate.Beta}, // GA in 0.21
```
mimowo: Update to 0.19 as we already released 0.18

Makefile-deps.mk
```diff
@@ -112,7 +112,8 @@ yq: ## Download yq locally if necessary.
 
 .PHONY: helm
 helm: ## Download helm locally if necessary.
-	@GOBIN=$(BIN_DIR) GO111MODULE=on $(GO_CMD) install helm.sh/helm/v4/cmd/helm@$(HELM_VERSION)
+	@$(PROJECT_DIR)/hack/testing/retry.sh --attempts 4 --delay 5 --exponential --stream -- \
+		env GOBIN=$(BIN_DIR) GO111MODULE=on $(GO_CMD) install helm.sh/helm/v4/cmd/helm@$(HELM_VERSION)
```
mimowo: This does not seem related directly to the PR, but might be useful, please open a dedidcated PR.

Makefile-test.mk
```diff
@@ -151,10 +151,19 @@ test-e2e-baseline-helm: test-e2e-baseline
...
+
+.PHONY: test-multikueue-baseline-e2e-main
+test-multikueue-baseline-e2e-main: test-multikueue-e2e-baseline
+
 .PHONY: test-multikueue-e2e-baseline
 test-multikueue-e2e-baseline: E2E_NPROCS := 5
 test-multikueue-e2e-baseline: setup-e2e-env run-test-multikueue-e2e-baseline-$(E2E_KIND_VERSION:kindest/node:v%=%)
 
+.PHONY: test-multikueue-extended-e2e-main
+test-multikueue-extended-e2e-main: test-multikueue-e2e-extended
```
mimowo: These changes seem unrelated, please revert if this is the case

pkg/controller/jobs/raycluster/common.go
```diff
@@ -99,6 +109,33 @@ func BuildPodSets(rayClusterSpec *rayv1.RayClusterSpec) ([]kueue.PodSet, error)
...
+		corev1.ResourceCPU:    resource.MustParse("200m"),
+		corev1.ResourceMemory: resource.MustParse("256Mi"),
+	}
+}
+
+func shouldAccountForRedisCleanup(rayClusterSpec *rayv1.RayClusterSpec, annotations map[string]string) bool {
+	return features.Enabled(features.KubeRayAccountForRedisCleanup) && rayutils.IsGCSFaultToleranceEnabled(rayClusterSpec, annotations)
+}
+
+func ExpectedPodSetsCount(rayClusterSpec *rayv1.RayClusterSpec, _ map[string]string) int {
```
mimowo: Do not pass annotations as this param is unsued in the latest version

pkg/controller/jobs/raycluster/common_test.go
```diff
@@ -824,9 +1009,82 @@ func TestValidateCreateRayClusterSpec(t *testing.T) {
 				},
 			},
 			wantErrors: field.ErrorList{
-				field.TooMany(field.NewPath("spec", "workerGroupSpecs"), 10, 9),
+				field.TooMany(field.NewPath("spec", "workerGroupSpecs"), 11, jobframework.MaxPodSets),
 			},
 		},
+		"too many worker groups with gcs fault tolerance": {
```
mimowo: TBH, I don't think this is needed -  we don't really change the number of PodSets with the latest approach. So, I think we can drop the test extensions. Let me know if I'm missing something, but it looks like legacy code from the previous attempt.

mimowo (commented):
LGTM, left minor comments 👍

nerdeveloper:
@mimowo, PTAL

mimowo:
@nerdeveloper I will check the PR tomorrow, and likely @yaroslava-serdiuk will review it too.

In the meanwhile it would help if you could make sure all the comments are addressed - for simple comments just mark them as resolved with a short note. For less obvious comments feel free to keep them open, but answer what was done, what wasn't and why. This helps reviewers a lot. Otherwise we need to reverse engineer essentially.

mimowo:
/release-note-edit
```release-note
KubeRay: Fixed missing quota accounting for Redis cleanup resources when GCS fault tolerance is enabled. Kueue
now accounts for the Redis cleanup Job resources for RayCluster, RayJob, and RayService workloads through the beta 
KubeRayAccountForRedisCleanup feature gate, which is enabled by default.
```

pkg/controller/jobs/raycluster/common.go
nerdeveloper: Resolved. The code now uses KubeRay's GCS fault-tolerance helper path and keeps the Kueue-side feature gate as the bailout mechanism.

pkg/controller/jobs/raycluster/common.go
nerdeveloper: Resolved. Added KubeRayAccountForRedisCleanup as a beta, default-on feature gate, with the GA-in-0.21 note.

pkg/controller/jobs/raycluster/common.go
nerdeveloper: Resolved. This PR now uses the shared jobframework.MaxPodSets constant from main and no longer introduces a Ray-specific limit constant.

pkg/controller/jobs/raycluster/common.go
nerdeveloper: Resolved. Removed the extra max-worker-groups helper; validation now uses ExpectedPodSetsCount against jobframework.MaxPodSets.

pkg/controller/jobs/raycluster/common_test.go
nerdeveloper: Resolved. The Redis cleanup coverage is now in the BuildPodSets table tests rather than separate test functions.

pkg/controller/jobs/raycluster/common_test.go
nerdeveloper: Resolved as outdated. The separate validation test is gone because Redis cleanup no longer changes PodSet count; feature-gate coverage remains in the BuildPodSets table.

pkg/controller/jobs/raycluster/common.go
nerdeveloper: Resolved. The current implementation returns an error if the head template does not include the Ray container, and coverage was added for that path.

pkg/controller/constants/constants.go
nerdeveloper: Resolved. This PR now uses the shared jobframework.MaxPodSets constant from main and no longer introduces a Ray-specific limit constant.

pkg/controller/jobs/raycluster/common.go
nerdeveloper: Resolved. Added the KubeRay source link next to the hardcoded Redis cleanup CPU/memory values.

pkg/features/kube_features.go
nerdeveloper: Resolved. Updated KubeRayAccountForRedisCleanup to start at 0.19 in kube_features.go and both feature-gate YAML files.

Makefile-deps.mk
nerdeveloper: Resolved. Reverted the unrelated Makefile-deps.mk change from this PR.

Makefile-test.mk
nerdeveloper: Resolved. Reverted the unrelated Makefile-test.mk changes from this PR.

pkg/controller/jobs/raycluster/common.go
nerdeveloper: Resolved. ExpectedPodSetsCount no longer takes annotations, and the RayCluster/RayJob/RayService call sites were updated.

pkg/controller/jobs/raycluster/common_test.go
nerdeveloper: Resolved. Dropped the legacy GCS fault-tolerance PodSet-count test extensions since Redis cleanup no longer changes PodSet count.

nerdeveloper:
@mimowo PTAL responded to the comments

pkg/features/kube_features.go
```diff
@@ -662,6 +668,9 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 	ConcurrentAdmission: {
 		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Alpha},
 	},
+	KubeRayAccountForRedisCleanup: {
```
mimowo: With the latest approach of folding the capacity under head I'm fine without the FG, because it significantly simplifies the design.

pkg/controller/jobs/rayjob/rayjob_controller_test.go
```diff
@@ -684,6 +728,30 @@ func TestNodeSelectors(t *testing.T) {
 	}
 }
 
+func TestRunWithPodSetsInfoDoesNotAddRedisCleanupPodSet(t *testing.T) {
```
yaroslava-serdiuk: nit: It doesn't seem we are testing anything here, lets drop

yaroslava-serdiuk (commented):
In general LGTM, left one nit
/lgtm

mimowo:
@nerdeveloper please address the remaining comments , seems like ready to go otherwise

pkg/features/kube_features.go
nerdeveloper: Done. Removed `KubeRayAccountForRedisCleanup` and the versioned feature-gate entries. Redis cleanup accounting now runs whenever `rayutils.IsGCSFaultToleranceEnabled` is true, so the fold-into-head approach is unconditional.

AI assistance was used to prepare this update.

mimowo (commented):
/lgtm
/approve
Thank you for hardening KubeRay integration 👍 

I will try to use robot to CP, but please prepare the CPs manually if the robot fails on conflcts

/cherrypick release-0.18
/cherrypick release-0.17

mimowo:
@nerdeveloper please check if we could CP to 0.17.

tenzen-y:
/release-note-edit

```release-note
RayJob, RayCluster, RayServe integrations: Fixed missing quota accounting for Redis cleanup resources when GCS fault tolerance is enabled. Kueue accounts for the Redis cleanup Job resources for workloads by folding the cleanup Job requests into the Ray head PodSet.
```

tenzen-y:
/release-note-edit

```release-note
RayJob, RayCluster, and RayServe integrations: Fixed missing quota accounting for Redis cleanup resources when GCS fault tolerance is enabled. Kueue accounts for the Redis cleanup Job resources for workloads by folding the cleanup Job requests into the Ray head PodSet.
```

PR #11263, opened 2026-05-17
Kueue triage ci skill

kannon92:
#### What type of PR is this?
/kind cleanup
/area agents

Provide a CI skill for figuring out why tests are failing.

The integration tests are quite difficult to parse for AI so I usually find myself teaching AI to do this.

Fixes #

```release-note
NONE
```

### DIFF
--- a/cmd/experimental/skills/README.md
+++ b/cmd/experimental/skills/README.md
@@ -6,7 +6,12 @@
 | [kueue-lineage](kueue-lineage.md) | "what pods are running for my workload", "trace workload to pods", "show me the jobs for this workload", lineage/ownership questions |
 | [kueue-flake-debugger](kueue-flake-debugger.md) | "debug a flake", "investigate test failure", "test timed out", "CI flake" |
 | [kueue-release-notes](kueue-release-notes/SKILL.md) | Review or propose a concise PR release note focused on the user-observable change. |
+| [reviewer/](reviewer/README.md) | "review this PR", "review this code", any code review of kueue changes |
+| [kueue-triage-ci](kueue-triage-ci/SKILL.md) | "why is CI failing", "triage CI", "integration tests failing", "verify failing", "CI red", "fix CI" |
+
 @kueue-who-preempted.md
 @kueue-lineage.md
 @kueue-flake-debugger.md
+@kueue-release-notes/SKILL.md
 @reviewer/README.md
+@kueue-triage-ci/SKILL.md
--- a/cmd/experimental/skills/kueue-triage-ci/SKILL.md
+++ b/cmd/experimental/skills/kueue-triage-ci/SKILL.md
@@ -1,0 +1,191 @@
+---
+name: kueue-triage-ci
+description: Triage failing CI on a Kueue PR. Use when CI is failing, integration tests fail, verify fails, lint fails, or a user asks "why is CI red" for a PR.
+license: Apache-2.0
+metadata:
+  copyright: The Kubernetes Authors
+---
+
+You are an expert in the Kueue project CI system which uses Prow for test automation. Kueue CI jobs store artifacts in GCS buckets accessible via `gsutil` and `https://storage.googleapis.com/`.
+
+## Input
+
+The argument is a PR number, a PR URL, or empty (defaults to the current branch's PR).
+
+If no argument is provided, find the PR for the current branch:
+
+```bash
+gh pr list --head "$(git branch --show-current)" --json number --jq '.[0].number'
+```
+
+If a full URL is provided, extract the PR number from it.
+
+## Step 1: Get failing checks
+
+```bash
+gh pr checks <PR_NUMBER> 2>&1
+```
+
+Filter to only jobs with `fail` status. Ignore `pending`, `pass`, and `skipping` jobs.
+
+For each failing job, extract the Prow URL. Convert Prow URLs to GCS and HTTPS base paths:
+
+- Prow URL: `https://prow.k8s.io/view/gs/<bucket>/<path>/<job-name>/<build-id>`
+- GCS base: `gs://<bucket>/<path>/<job-name>/<build-id>`
+- HTTPS base: `https://storage.googleapis.com/<bucket>/<path>/<job-name>/<build-id>`
+
+## Step 2: Categorize failures
+
+Group failing jobs into categories:
+
+| Category | Job name pattern | Triage approach |
+|---|---|---|
+| Verify | `pull-kueue-verify-*` | Fetch build-log.txt, look for lint/fmt/typecheck errors |
+| Integration | `pull-kueue-test-integration-*` | Fetch JUnit XML artifacts, parse test failures |
+| E2E | `pull-kueue-test-e2e-*` | Fetch JUnit XML artifacts, check for infra issues |
+| Unit | `pull-kueue-test-unit-*` | Fetch build-log.txt, look for test failures |
+| Other | everything else | Fetch build-log.txt for general analysis |
+
+Triage each category using the appropriate steps below.
+
+## Step 3: Triage verify failures
+
+For `verify-*` jobs, fetch the build log:
+
+```bash
+# Use WebFetch to get the build log
+# URL: <HTTPS_BASE>/build-log.txt
+```
+
+Look for these common Kueue verify targets and their errors:
+
+- **`verify-ci-lint`** (golangci-lint): typecheck errors, gci import ordering, modernize suggestions, unused variables
+- **`verify-lint-api`** (API linter): typecheck errors in API types or test files that use API types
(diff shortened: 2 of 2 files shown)


amy:
@kannon92 Looks good! Do you have a sample PR you tried this on?

kannon92:
> @kannon92 Looks good! Do you have a sample PR you tried this on?

Yes I used it on a few of the PRs I have. I'll dig up one with test failure and tag on it.

I fixed the CI failures already.

kannon92:
@amy 

I have an example:

CI Triage: PR #10479                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                          
 ### Failing Jobs                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                          
 ┌───────────────────────────────────────────┬─────────┐                                                                                                                                                                                                                                  
 │ Job                                       │ Status  │                                                                                                                                                                                                                                  
 ├───────────────────────────────────────────┼─────────┤                                                                                                                                                                                                                                  
 │ pull-kueue-test-unit-main                 │ ❌ FAIL │                                                                                                                                                                                                                                  
 ├───────────────────────────────────────────┼─────────┤                                                                                                                                                                                                                                  
 │ pull-kueue-test-integration-baseline-main │ ❌ FAIL │                                                                                                                                                                                                                                  
 └───────────────────────────────────────────┴─────────┘                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                          
 All other jobs (E2E, extended integration, verify, build, perf) ✅ pass.                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                          
 ────────────────────────────────────────────────────────────────────────────────                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                          
 ### Root Cause                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                          
 File: pkg/controller/jobframework/reconciler.go, EquivalentToWorkload() (~line 1271)                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                          
 There are two MaximumExecutionTimeSeconds checks — the first one is incorrect and contradicts the second:                                                                                                                                                                                
                                                                                                                                                                                                                                                                                          
 ```go                                                                                                                                                                                                                                                                                    
   defaultDuration := int32(-1)                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                          
   // ❌ BUG — this fires first, returns false when the workload has a CQ-inherited                                                                                                                                                                                                       
   // timeout but the job has NO explicit timeout label (MaximumExecutionTimeSeconds(job) == nil)                                                                                                                                                                                         
   if ptr.Deref(wl.Spec.MaximumExecutionTimeSeconds, defaultDuration) != ptr.Deref(MaximumExecutionTimeSeconds(job), defaultDuration) {                                                                                                                                                   
       return false, nil  // 3600 != -1 → incorrectly says "not equivalent"                                                                                                                                                                                                               
   }                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                          
   // ✅ CORRECT intent — only flag a mismatch if the job *explicitly* carries the label                                                                                                                                                                                                  
   // (dead code because the first check already returned false)                                                                                                                                                                                                                          
   if jobMaxExecTime := MaximumExecutionTimeSeconds(job); jobMaxExecTime != nil {                                                                                                                                                                                                         
       if ptr.Deref(wl.Spec.MaximumExecutionTimeSeconds, defaultDuration) != *jobMaxExecTime {                                                                                                                                                                                            
           return false, nil                                                                                                                                                                                                                                                              
       }                                                                                                                                                                                                                                                                                  
   }                                                                                                                                                                                                                                                                                      
 ```                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                          
 When a job has no max-exec-time-seconds label, MaximumExecutionTimeSeconds(job) returns nil → ptr.Deref(nil, -1) = -1. But the workload may have MaximumExecutionTimeSeconds=3600 (inherited from the ClusterQueue). The first check sees 3600 != -1 and incorrectly returns false, nil  
 — marking the workload as non-equivalent and triggering an UpdatedWorkload event.                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                          
 ### Failing Tests & Symptoms                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                          
 ┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┬────────────────────────────────────────────────────────────────────────────┐                                                                             
 │ Test                                                                                                                        │ Error                                                                      │                                                                             
 ├─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────┤                                                                             
 │ Unit: workload_equivalency_passes_with_ClusterQueue-sourced_timeout (×2 for merge patch variants)                           │ Got unexpected UpdatedWorkload event; expected none                        │                                                                             
 ├─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────┤                                                                             
 │ Integration: should admit and run a job with ClusterQueue-sourced timeout without equivalency mismatch                      │ Admitted workload gets evicted due to false equivalency mismatch           │                                                                             
 ├─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────┤                                                                             
 │ Integration: should keep the original timeout when a job moves to a different LocalQueue backed by a different ClusterQueue │ Equivalency check fails → reconciler overwrites timeout when queue changes │                                                                             
 └─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┴────────────────────────────────────────────────────────────────────────────┘                                                                             
                                                                                                                                                                                                                                                                                          
 ### Fix                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                          
 Remove the incorrect first check. The second check (which only compares when the job explicitly carries the label) is both correct and sufficient:

cmd/experimental/skills/kueue-triage-ci/SKILL.md
```diff
@@ -0,0 +1,191 @@
...
+### 4a: Check JUnit XML for test failures
+
+```bash
+gsutil ls "<GCS_BASE>/artifacts/"
+```
+
+Fetch the top-level `integration-junit.xml`:
+
+```bash
+gsutil cat "<GCS_BASE>/artifacts/integration-junit.xml" | grep -oP 'failures="[^"]*"' | sort | uniq -c | sort -rn
```
amy, line 87: Did a review pass with ai and it caught this:  

> grep -oP is not portable to macOS
> 
>   grep -oP 'failures="[^"]*"' | sort | uniq -c | sort -rn
>   -P is GNU-grep only; BSD grep on macOS will fail. Either switch to grep -oE (extended regex works for this pattern) or do the parse entirely in the Python block below it.

cmd/experimental/skills/kueue-triage-ci/SKILL.md
```diff
@@ -0,0 +1,191 @@
...
+
+If the failure looks like a flake (intermittent timeout, transient network error), cross-reference with the [kueue-flake-debugger](../kueue-flake-debugger/SKILL.md) skill for deeper analysis including kube-scheduler logs, kubelet logs, and pod placement.
+
+## Step 6: Distinguish PR-specific vs pre-existing failures
+
+For each failure, determine if it was introduced by this PR:
+
+```bash
+# Get files changed in this PR
+git diff --name-only origin/main...HEAD
```
amy, line 158: Add a `git fetch origin main` before this

cmd/experimental/skills/README.md
```diff
@@ -6,7 +6,12 @@
 | [kueue-lineage](kueue-lineage.md) | "what pods are running for my workload", "trace workload to pods", "show me the jobs for this workload", lineage/ownership questions |
 | [kueue-flake-debugger](kueue-flake-debugger.md) | "debug a flake", "investigate test failure", "test timed out", "CI flake" |
 | [kueue-release-notes](kueue-release-notes/SKILL.md) | Review or propose a concise PR release note focused on the user-observable change. |
+| [reviewer/](reviewer/README.md) | "review this PR", "review this code", any code review of kueue changes |
+| [kueue-triage-ci](kueue-triage-ci/SKILL.md) | "why is CI failing", "triage CI", "integration tests failing", "verify failing", "CI red", "fix CI" |
```
amy, line 10: Given this has overlap with the flake debugger, you may want to add more clarification on when to use one over the other in the table of context.

kannon92: Hmm. good call here.

I'm not sure. The skill "flake debugger" is really just a general way to deduce CI failures.

The code is the same to get logs so maybe this is not actually needed.

PR #11002, opened 2026-05-07
test: split e2e TAS tests into baseline and extended

ekam-walia:
What type of PR is this:
/kind cleanup
/area tas
/area testing

Why we need it:
Splits the TAS e2e test suite into baseline and extended shards 
to reduce CI wall-clock time for PRs.

- tas-baseline: tests using only built-in Kubernetes resources (Job, Pod, StatefulSet)
- tas-extended: tests requiring custom integrations (JobSet, LeaderWorkerSet, 
  KubeRay, AppWrapper, TrainJob, MPIJob, PyTorch)

Which issue this PR fixes:
Part of #10995

Follows the same pattern as #10472.

User Inerface Change : NONE

```release-note
NONE
```

Issue #10995, opened 2026-05-06
Optimize the e2e tests by sharing in baseline and extended


mimowo:
**What would you like to be cleaned**:

Shorten the time for PRs to build by optimizing the longest taking e2e suites - see the analysis below.

**Why is this needed**:

I have LLMed a script to fetch the runtime of prow jobs:
```py
#!/usr/bin/env python3
from __future__ import annotations

import argparse
import csv
import datetime as dt
import json
import re
import statistics
import sys
import time
import urllib.error
import urllib.parse
import urllib.request
from dataclasses import dataclass
from typing import Any, Iterable

DEFAULT_BUCKET = "kubernetes-ci-logs"
KUEUE_PERIODICS_URL = (
    "https://raw.githubusercontent.com/kubernetes/test-infra/master/config/jobs/kubernetes-sigs/kueue/kueue-periodics-main.yaml"
)

@dataclass(frozen=True)
class BuildRuntime:
    job: str
    build_id: str
    result: str
    start: dt.datetime
    finish: dt.datetime
    duration_seconds: int
    prow_url: str

def http_get(url: str, *, tries: int = 4, timeout: int = 30) -> bytes:
    last_exc: Exception | None = None
    for attempt in range(tries):
        try:
            req = urllib.request.Request(url, headers={"User-Agent": "kueue-prow-runtime/1.0"})
            with urllib.request.urlopen(req, timeout=timeout) as resp:
                return resp.read()
        except urllib.error.HTTPError as e:
            if e.code == 404:
                raise
            last_exc = e
        except urllib.error.URLError as e:
            last_exc = e
        time.sleep(0.5 * (2**attempt))
    raise last_exc  # type: ignore[misc]

def http_json(url: str) -> dict[str, Any]:
    return json.loads(http_get(url).decode("utf-8"))

def parse_gs_prefix(job_or_prefix: str, bucket: str = DEFAULT_BUCKET) -> tuple[str, str, str]:
    value = job_or_prefix.strip().rstrip("/")

    if value.startswith("gs://"):
        rest = value.removeprefix("gs://")
        bucket_name, _, path = rest.partition("/")
        return path.rstrip("/").split("/")[-1], bucket_name, path.rstrip("/")

    if value.startswith("https://storage.googleapis.com/"):
        rest = value.removeprefix("https://storage.googleapis.com/").rstrip("/")
        bucket_name, _, path = rest.partition("/")
[...]

PR #11267, opened 2026-05-18
test: split hotswap_test.go into TAS baseline and extended

ekam-walia:
What type of PR is this:
/kind cleanup
/area tas
/area testing

Splits test/e2e/tas/extended/hotswap_test.go into two files:

- test/e2e/tas/baseline/hotswap_test.go: contains tests using only 
  built-in Job resources (requires TASReplaceNodeOnNodeTaints feature gate)

- test/e2e/tas/extended/hotswap_test.go: retains only tests using 
  JobSet (custom integration)

This aligns with the baseline/extended split pattern established in #11002.

Which issue(s) this PR fixes:

Part of #10995

Follow-up to #11002 which split the TAS e2e suite into baseline and extended.
The 4 Job-based hotswap tests are moved to baseline.
The 2 JobSet-based hotswap tests remain in extended.

Does this PR introduce a user-facing change:

```release-note
NONE
```


mimowo:
Thank you 👍 This PR will improve the split between the "baseline" and "extended" following the pattern of no-intragration/with integrations.

mbobrovskyi:
It looks like netlify stuck.

test/e2e/tas/baseline/hotswap_test.go
```diff
@@ -0,0 +1,608 @@
...
+			util.ExpectObjectToBeDeleted(ctx, k8sClient, localQueue, true)
+			util.ExpectObjectToBeDeleted(ctx, k8sClient, clusterQueue, true)
+			util.ExpectObjectToBeDeleted(ctx, k8sClient, tasFlavor, true)
+			util.ExpectObjectToBeDeleted(ctx, k8sClient, topology, true)
+			util.ExpectAllPodsInNamespaceDeleted(ctx, k8sClient, ns)
+		})
+
+		ginkgo.Context("with TASReplaceNodeOnNodeTaints enabled", func() {
+			ginkgo.BeforeAll(func() {
+				util.UpdateKueueConfigurationAndRestart(ctx, k8sClient, defaultKueueCfg, kindClusterName, func(cfg *configapi.Configuration) {
```
mbobrovskyi, line 124: Hmm. Shouldn't it be a part of sequential tests?

mimowo: tas is already sequential

mbobrovskyi:
/lgtm

Thank you!

test/e2e/tas/baseline/hotswap_test.go
ekam-walia: deploy/netlify issue is resolved

mimowo: However, actually from 0.17+ I think `TASReplaceNodeOnNodeTaints` is already enabled by default, so we can drop the configuration change and restart from the tests: https://github.com/kubernetes-sigs/kueue/blob/b4cc4f091f67428f5326800c8a378314429afdc4/pkg/features/kube_features.go#L540-L542

at least in 0.17+. This would also save us 30s. I would say this is a follow up PR.

mbobrovskyi: Actually, we have parallelization here https://github.com/kubernetes-sigs/kueue/blob/b4cc4f091f67428f5326800c8a378314429afdc4/Makefile-test.mk#L186.

mbobrovskyi: /hold

mbobrovskyi: > However, actually from 0.17+ I think TASReplaceNodeOnNodeTaints is already enabled by default, so we can drop the configuration change and restart from the tests:

@ekam-walia could you please apply this?

mimowo: Interesting, I don't think this  `test-e2e-baseline: E2E_NPROCS := 4 ` works well, because in different tests we verify speciifc kind nodes. Maybe this works ok in practice because the tests are waiting before the job from previous test is deleted - to be verified.

mbobrovskyi: I’m worries that this could cause some flaky issues. I propose removing the configuration to be safe.

mimowo: Ok, let's follow up with two PRs:
- drop the `UpdateKueueConfigurationAndRestart` from TAS, and cherrypick to 0.17
- set the parallelism for tas-baseline and tas-extended to 1

mimowo: Left a comment under the issue also so that we don't forget: https://github.com/kubernetes-sigs/kueue/issues/10995#issuecomment-4475966279

ekam-walia: > > However, actually from 0.17+ I think TASReplaceNodeOnNodeTaints is already enabled by default, so we can drop the configuration change and restart from the tests:
> 
> @ekam-walia could you please apply this?

Yes, I will apply this

mbobrovskyi:
@ekam-walia could you please prepare cherry-picks?

PR #11280, opened 2026-05-18
Setting parallelism to 1 for TAS baseline and TAS extended tests

ekam-walia:
#### What type of PR is this?

/kind cleanup
/area tas

Sets parallelism to 1 for TAS baseline and TAS extended tests

Part of #11267

```release-note
NONE
```

### DIFF
--- a/Makefile-test.mk
+++ b/Makefile-test.mk
@@ -177,7 +177,7 @@ test-multikueue-e2e-helm: test-multikueue-e2e
 ## Examples:
 ##   Run only AppWrapper tests: GINKGO_ARGS="--label-filter=feature:appwrapper" make test-e2e-extended
 .PHONY: test-e2e-extended
-test-e2e-extended: E2E_NPROCS := 4
+test-e2e-extended: E2E_NPROCS := 1
 test-e2e-extended: setup-e2e-env kind-ray-project-mini-image-build run-test-e2e-extended-$(E2E_KIND_VERSION:kindest/node:v%=%)
 
 ## Label Taxonomy:
@@ -186,7 +186,7 @@ test-e2e-extended: setup-e2e-env kind-ray-project-mini-image-build run-test-e2e-
 ## Examples:
 ##   Run only job tests: GINKGO_ARGS="--label-filter=feature:job" make test-e2e-baseline
 .PHONY: test-e2e-baseline
-test-e2e-baseline: E2E_NPROCS := 4
+test-e2e-baseline: E2E_NPROCS := 1
 test-e2e-baseline: setup-e2e-env kueuectl run-test-e2e-baseline-$(E2E_KIND_VERSION:kindest/node:v%=%)
 
 .PHONY: test-tas-e2e-baseline


mimowo:
/lgtm
/approve
Thank you for aligning the TAS tests with the assumptions. We may in the future consider running them on multiple threads but at the moment I don't think we are ready for this.

j-skiba:
we could also move the `hotswap_test.go` to a separate dir (e.g. `test/e2e/tas/extended/hotswap/hotswap_test.go`) and leave procs > 1 unless we want to verify other tests too.

mimowo:
> we could also move the `hotswap_test.go` to a separate dir (e.g. `test/e2e/tas/extended/hotswap/hotswap_test.go`) and leave procs > 1 unless we want to verify other tests too.

We are changing currently hotswap so that it is no longer using custom configuration: https://github.com/kubernetes-sigs/kueue/pull/11281. That was only needed for 0.16 when the Taints FG was alpha.

However, I think the tests still need to be reviewed for enabling parallelism on them, because:
- they don't use the NS-prefix / suffix to avoid conflicts, for example here we have hard-coded Flavor name: https://github.com/kubernetes-sigs/kueue/blob/main/test/e2e/tas/baseline/job_test.go#L61-L63
- they check specific nodes like here: https://github.com/kubernetes-sigs/kueue/blob/main/test/e2e/tas/baseline/job_test.go#L180-L193
however, the nodes might be blocked due to other jobs occupying them.

So, I would say it is a separate (possible) effort. We should probably assign higher amount of the exended resources per node, so that the jobs don't block each other, and start using NS-prefix / suffix approach. 

I will open an issue to enable parallelism for TAS tests careful of these issues so that we don't run into flakes.

Makefile-test.mk
```diff
@@ -183,7 +186,7 @@ test-e2e-extended: setup-e2e-env kind-ray-project-mini-image-build run-test-e2e-
 ## Examples:
 ##   Run only job tests: GINKGO_ARGS="--label-filter=feature:job" make test-e2e-baseline
 .PHONY: test-e2e-baseline
-test-e2e-baseline: E2E_NPROCS := 4
+test-e2e-baseline: E2E_NPROCS := 1
```
mbobrovskyi, line 189: /hold

mbobrovskyi: This is a singlecluster test. Please revert it.

Makefile-test.mk
```diff
@@ -174,7 +177,7 @@ test-multikueue-e2e-helm: test-multikueue-e2e
 ## Examples:
 ##   Run only AppWrapper tests: GINKGO_ARGS="--label-filter=feature:appwrapper" make test-e2e-extended
 .PHONY: test-e2e-extended
-test-e2e-extended: E2E_NPROCS := 4
+test-e2e-extended: E2E_NPROCS := 1
```
mbobrovskyi, line 180:
```suggestion
```
I think we can remove it, since E2E_NPROCS defaults to 1.

mimowo: but this is also non-TAS tests, thanks for spotting.

mimowo: Actually our TAS already is using single thread: https://github.com/kubernetes-sigs/kueue/blob/main/Makefile-test.mk#L189-L193

Makefile-test.mk
mimowo: our TAS tests already are using single thread https://github.com/kubernetes-sigs/kueue/blob/main/Makefile-test.mk#L189-L193

mimowo:
/close
I think there is nothing to do, the TAS tests already correctly set single thread: https://github.com/kubernetes-sigs/kueue/blob/main/Makefile-test.mk#L189-L193

Makefile-test.mk
mbobrovskyi: Ah, you right. Thank you!

Issue #11062, opened 2026-05-08
MultiKueue when Preemption with a multikueue admission check Should re-do admission process when workload gets evicted in the worker


mimowo:
**Which test is flaking?**:
 MultiKueue when Preemption with a multikueue admission check Should re-do admission process when workload gets evicted in the worker 

**Link to failed CI job or steps to reproduce locally**:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/periodic-kueue-test-e2e-multikueue-release-0-17/2052761352430161920

**Failure message or logs**:
```
End To End MultiKueue Suite: kindest/node:v1.35.0: [It] MultiKueue when Preemption with a multikueue admission check Should re-do admission process when workload gets evicted in the worker expand_less	1m35s
{ failed [FAILED] Timed out after 90.000s.
The function passed to Eventually failed at /home/prow/go/src/kubernetes-sigs/kueue/test/e2e/multikueue/e2e_test.go:1439 with:
Workload not Admitted in worker2
Manager Workload:     <v1.ObjectMeta>: {
        Name: "job-job-e64f9",
        GenerateName: "",
        Namespace: "multikueue-cc7c4",
        SelfLink: "",
        UID: "48f736bd-d81c-446c-ad36-6ed226167bf3",
        ResourceVersion: "5017",
        Generation: 1,
        CreationTimestamp: {
            Time: 2026-05-08T15:01:13Z,
        },
        DeletionTimestamp: nil,
        DeletionGracePeriodSeconds: nil,
        Labels: {
            "kueue.x-k8s.io/job-uid": "b1832137-e40e-42c3-af85-f3bf927789c7",
        },
        Annotations: nil,
        OwnerReferences: [
            {
                APIVersion: "batch/v1",
                Kind: "Job",
                Name: "job",
                UID: "b1832137-e40e-42c3-af85-f3bf927789c7",
                Controller: true,
                BlockOwnerDeletion: true,
            },
        ],
        Finalizers: [
            "kueue.x-k8s.io/resource-in-use",
        ],
        ManagedFields: [
            {
                Manager: "kueue-admission",
                Operation: "Apply",
                APIVersion: "kueue.x-k8s.io/v1beta2",
                Time: {
                    Time: 2026-05-08T15:01:15Z,
                },
                FieldsType: "FieldsV1",
                FieldsV1: {
[...]

PR #5361, opened 2025-05-26
Fix - Workloads requesting TAS cannot run via MultiKueue

IrvingMg:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:
It removes a validation that was preventing TAS from running with Multikueue.

#### Which issue(s) this PR fixes:

Fixes #5313 

#### Special notes for your reviewer:
Added test files used for verifying this change on GKE.

#### Does this PR introduce a user-facing change?

```release-note
Fix: Removed a validation that was preventing TAS from running with MultiKueue.
```

PR #11286, opened 2026-05-18
Make the read-only contract on Snapshot explicit

mimowo:
#### What type of PR is this?

/kind bug

Snapshot() is conceptually a read-only operation, but the current implementation calls a mutating helper, which makes the locking correctness very hard to reason about. To make the intended contract explicit, I am proposing this PR: Snapshot() no longer tries to repair TAS usage; instead, it skips TAS ClusterQueues whose TAS usage is not synced yet. The cache update paths remain responsible for performing TAS sync under the cache write lock.

I’ve analyzed this case, and I think this is a false-positive bug detection from the system-level perspective.

The concern is valid locally: `Snapshot()` currently calls `ensureTASIsSynced()` while holding only the cache read lock, and `ensureTASIsSynced()` is a mutating helper. 

However, due to the interaction between the scheduler cache update paths and the controllers, `ensureTASIsSynced()` called from Scheduler does not perform the mutating repair.

The problematic state would be:

```
cq.isTASInitialized() == true
cq.isTASSynced == false
```

This is the only state where `ensureTASIsSynced()` would perform the mutating repair path and re-account existing workloads for TAS usage.

When a Topology is added or updated, the scheduler cache handles it under the cache write lock:
```
Cache.AddOrUpdateTopology
  -> Cache.Lock()
  -> tasCache.AddTopology(...)
  -> updateClusterQueues(...)
      -> cq.UpdateWithFlavors(...)
          -> cq.updateQueueStatus(...)
              -> cq.ensureTASIsSynced(...)
  -> Cache.Unlock()
```

The same applies when a ResourceFlavor is added or updated:
```
Cache.AddOrUpdateResourceFlavor
  -> Cache.Lock()
  -> tasCache.AddResourceFlavor(...)
  -> updateClusterQueues(...)
      -> cq.UpdateWithFlavors(...)
          -> cq.updateQueueStatus(...)
              -> cq.ensureTASIsSynced(...)
  -> Cache.Unlock()
```

So if adding or updating the Topology or ResourceFlavor makes TAS usage sync performed under the cache write lock, blocking `Snapshot()` which needs to take the read lock.

```release-note
TAS: ensure that `Snapshot()` does not perform update the list of workloads under the read-lock.
```

### DIFF
--- a/pkg/cache/scheduler/snapshot.go
+++ b/pkg/cache/scheduler/snapshot.go
@@ -38,6 +38,14 @@ import (
 	"sigs.k8s.io/kueue/pkg/workload"
 )
 
+type inactiveCQReason string
+
+const (
+	inactiveCQReasonNotActive         inactiveCQReason = "NotActive"
+	inactiveCQReasonHasCycle          inactiveCQReason = "HasCycle"
+	inactiveCQReasonTASUsageNotSynced inactiveCQReason = "TASUsageNotSynced"
+)
+
 type Snapshot struct {
 	hierarchy.Manager[*ClusterQueueSnapshot, *CohortSnapshot]
 	ResourceFlavors          map[kueue.ResourceFlavorReference]*kueue.ResourceFlavor
@@ -176,12 +184,11 @@ func (c *Cache) Snapshot(ctx context.Context, options ...SnapshotOption) (*Snaps
 	log := ctrl.LoggerFrom(ctx)
 	cqNames := c.hm.ClusterQueues()
 	for _, cq := range cqNames {
-		if !cq.Active() || (cq.HasParent() && hierarchy.HasCycle(cq.Parent())) {
+		if reason := skipInactiveCQReason(cq); reason != "" {
+			log.V(3).Info("Skipping ClusterQueue", "clusterQueue", cq.Name, "reason", reason)
 			snap.InactiveClusterQueueSets.Insert(cq.Name)
 			continue
 		}
-		// ensure all workloads are accounted for TAS before building TAS snapshots
-		cq.ensureTASIsSynced(log)
 	}
 	tasSnapshots := make(map[kueue.ResourceFlavorReference]*TASFlavorSnapshot)
 	if features.Enabled(features.TopologyAwareScheduling) {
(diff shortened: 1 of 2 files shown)


mimowo:
cc @PBundyra ptal

pkg/cache/scheduler/snapshot.go
```diff
@@ -214,6 +221,23 @@ func (c *Cache) Snapshot(ctx context.Context, options ...SnapshotOption) (*Snaps
...
+	if !cq.Active() {
+		return inactiveCQReasonNotActive
+	}
+	if cq.HasParent() && hierarchy.HasCycle(cq.Parent()) {
+		return inactiveCQReasonHasCycle
+	}
+	if features.Enabled(features.TopologyAwareScheduling) && len(cq.tasFlavors) > 0 && !cq.isTASSynced {
+		// The CQ uses a TAS flavor, but TAS usage is not synced yet.
+		return inactiveCQReasonTASUsageNotSynced
+	}
```
kshalot, line 237: I think in the case of MultiKueue, the topology is not created on the manager cluster, right? I just wanted to check in whether `cq.isTASSynced` won't be always false in this case. There's explicit handling in `isTASViolated` for example:
https://github.com/kubernetes-sigs/kueue/blob/cbbb65b4fc16dd9c633682fdaeb5ced8b42d54b6/pkg/cache/scheduler/clusterqueue.go#L349-L352

mimowo:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/11286/pull-kueue-test-e2e-multikueue-main/2056343144110231552
This looks like the known flake: https://github.com/kubernetes-sigs/kueue/issues/11062

pkg/cache/scheduler/snapshot.go
mimowo: Currently the design assumes creation of the Topology on the management cluster, as created here: https://github.com/kubernetes-sigs/kueue/blob/main/test/e2e/multikueue/baseline/tas_test.go#L120-L121

That was added in this PR: https://github.com/kubernetes-sigs/kueue/pull/5361 and here is where we defer the actual TAS computations of MultiKueue management cluster: https://github.com/kubernetes-sigs/kueue/pull/5361/changes#diff-5a39cea8316e94ece1214aef72728e4476f4531d9714f30c2b00f33b00b47e4fR80-R82

This skipping of validation for MK is something we may need to revisit, this looks too strong IMO, but the design of the MultiKueue + TAS story is not ideal for sure. I'm hoping we can improve on the design in the future.

That being said I'm not yet sure how this impacts correctness of the code changes, I don't think it does. I think regardless if MultiKueue, or not MultiKueue then key think is that the event handlers for Topology, and ResourceFlavor will trigger `cq.ensureTASIsSynced`, and so there is nothing to be done for Snapshot().

pkg/cache/scheduler/snapshot_test.go
```diff
@@ -768,6 +769,90 @@ func TestSnapshot(t *testing.T) {
 				),
 			},
 		},
+		"clusterQueue with unsynced TAS usage is excluded from snapshot": {
```
PBundyra:
```suggestion
		"clusterQueue with unsynced TAS usage due to missing topology is excluded from snapshot": {
```

PBundyra:
/lgtm
/hold
Thanks, I just left a small nit, feel free to unhold if you think it's insignificant

pkg/cache/scheduler/snapshot_test.go
mimowo: Done, added in brackets so that if reads better

mimowo:
/unhold
as the remark is addressed

mimowo:
I think the current Snapshot() calling ensureTASIsSynced is so misleading that we can categorize this as a bug for cherrypicking even if there is no system-level issue
/cherrypick release-0.17
/cherrypick release-0.16

PBundyra:
/lgtm
Thanks

PR #11292, opened 2026-05-18
Fix integration test for workload slice replacement ordering

### LINKED ISSUE #11283: Fix the integration test for the replacement contract for WorkloadSlices
**What happened**:

We have disovered that the test added in https://github.com/kubernetes-sigs/kueue/pull/11195 is not testing the behavior change - is passing before and after the change.: https://github.com/kubernetes-sigs/kueue/pull/11195/changes#r3248009882

The test is demonstrating the behavioral change. We want to make sure the old workload slice is not marked Finished when transionining the new slice to Admitted.

sohankunkerkar:
#### What type of PR is this?
/kind bug

Fixes https://github.com/kubernetes-sigs/kueue/issues/11283

```release-note
NONE
```

### DIFF
--- a/test/integration/singlecluster/controller/jobs/job/job_controller_test.go
+++ b/test/integration/singlecluster/controller/jobs/job/job_controller_test.go
@@ -4327,15 +4327,15 @@ var _ = ginkgo.Describe("Job with elastic jobs via workload-slices support", gin
 		util.ExpectLQFinishedWorkloadsTotalMetric(localQueue, highPriorityClass.Name, 0)
 	})
 
-	ginkgo.It("Should keep old workload-slice unfinished while replacement slice is pending", func() {
+	ginkgo.It("Should not finish old workload-slice before new slice is admitted", func() {
 		features.SetFeatureGateDuringTest(ginkgo.GinkgoTB(), features.ElasticJobsViaWorkloadSlices, true)
 
-		elasticJob := testingjob.MakeJob("job-pending-replacement", ns.Name).
+		elasticJob := testingjob.MakeJob("job-slice-ordering", ns.Name).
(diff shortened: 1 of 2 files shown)


test/integration/singlecluster/controller/jobs/job/job_controller_test.go
```diff
@@ -4347,30 +4348,52 @@ var _ = ginkgo.Describe("Job with elastic jobs via workload-slices support", gin
...
 
-		ginkgo.By("scaling the job beyond the remaining quota")
+		ginkgo.By("setting up a watch on workloads before scaling")
+		watchClient, ok := k8sClient.(client.WithWatch)
+		gomega.Expect(ok).Should(gomega.BeTrue(), "k8sClient must implement client.WithWatch")
+		watcher, err := watchClient.Watch(ctx, &kueue.WorkloadList{}, &client.ListOptions{
+			Namespace: elasticJob.Namespace,
+		})
+		gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+		defer watcher.Stop()
```
mimowo, line 4357: I'm wondering if we could encapsulate the timeout handling behind a helper, similarly as we do for events with `ExpectEventsForObjectsWithTimeout`, just to avoid mixing the abstraction layers or asserting on business logic and managing the watch state, wdyt?

sohankunkerkar: I considered extracting it but held off since there was only one callsite. But you're right that separating the watch plumbing from the business assertion is cleaner regardless

mimowo:
/lgtm
/approve
Thank you for fixing the test so that it can pick up regressions in the mechanism 👍

PR #11293, opened 2026-05-18
  fix: account assumed pod usage in TAS multi-podset placement

### LINKED ISSUE #11287: TAS workload with multiple PodSets may exceed node pods capacity
**What happened**:
When a Node has capacity to host only one pod, and a workload has two podsets (each with one pod), it can happen that the workload is admitted by TAS.

I would expect this unit test [c49b9bc](https://github.com/kubernetes-sigs/kueue/commit/c49b9bcf16252422255c82f8314bf9fc874f090d) to pass but its failing (the workload is admitted).

I think we probably should account the for these already scheduled pods of the first podset in the `addAssumedUsage` function (see the commented line in the above commit, if we uncomment it, the test will pass).

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

yuluo-yx:
#### What type of PR is this?

/kind bug
/area tas
/area testing

Fixes #11287

```release-note
TAS: Fixed a scheduling bug where a workload with multiple PodSets could be admitted even when the combined PodSets exceeded node pod capacity.
```


mimowo:
@yuluo-yx since this is a bug please post release-note

pkg/cache/scheduler/tas_elastic_workloads.go
```diff
@@ -82,14 +82,8 @@ func (s *TASFlavorSnapshot) handleScaleUp(
...
-	// Previous pods consume capacity
-	prevAssumedUsage := utiltas.ComputeUsagePerDomain(prevAssignment, workers.SinglePodRequests)
-	for domainID, usage := range prevAssumedUsage {
-		if assumedUsage[domainID] == nil {
-			assumedUsage[domainID] = resources.Requests{}
-		}
-		assumedUsage[domainID].Add(usage)
-	}
+	// Previous pods consume capacity.
+	addAssumedUsage(assumedUsage, prevAssignment, &workers)
```
mimowo, line 86: Do we have also a test checking this fix?

yuluo-yx: yep, this is covered by the new scheduler-level regression test in 

`pkg/scheduler/scheduler_tas_test.go`: `TestScheduleForTASCohorts/workload with two PodSets exceeds node pods capacity`

It reproduces the issue from #11287 with one node whose `pods` capacity is `1` and a workload with two PodSets, each requesting one pod. The expected result is that the workload stays, pending with:

`topology "tas-single-level" doesn't allow to fit any of 1 pod(s). Total nodes: 1; excluded: resource "pods": 1`

and i also added lower-level coverage:

- `pkg/cache/scheduler/tas_cache_test.go`: `multiple PodSets account assumed pod usage against allocatable pods; BestFit`;
- `pkg/cache/scheduler/tas_flavor_snapshot_test.go`: `TestAddAssumedUsageIncludesPodCount`.

mimowo:
/release-note-edit
```release-note
TAS: Fixed a scheduling bug where a workload with multiple PodSets could be admitted even when the combined PodSets exceeded node pod capacity.
```

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -656,12 +656,15 @@ func (s *TASFlavorSnapshot) findReplacementAssignment(
 }
 
 func addAssumedUsage(assumedUsage map[utiltas.TopologyDomainID]resources.Requests, ta *utiltas.TopologyAssignment, tr *TASPodSetRequests) {
-	for _, domain := range ta.Domains {
-		domainID := utiltas.DomainID(domain.Values)
+	addUsagePerDomain(assumedUsage, utiltas.ComputeUsagePerDomain(ta, tr.SinglePodRequests))
```
mimowo, line 664: This function ComputeUsagePerDomain is an excellent find to de-duplicate the code

pkg/cache/scheduler/tas_flavor_snapshot_test.go
```diff
@@ -625,6 +625,46 @@ func TestComputeAssumedUsageFromAssignment(t *testing.T) {
 	}
 }
 
+func TestAddAssumedUsageIncludesPodCount(t *testing.T) {
+	assumedUsage := map[tas.TopologyDomainID]resources.Requests{
```
mimowo: Let's try to add another test case, and turn the test into table-driven test, so that future addiions of new scenarios don't require re-formatting

mimowo: And generalize the tested function, say `TestAddAssumedUsage`

yuluo-yx: done

mimowo (commented):
/lgtm
/approve
Thank you for the TAS bugfix for multi-podset workloads which are quite important.
/cherrypick release-0.17
/cherrypick release-0.16
Please prepare manual CPs in case the robot fails on conflicts

mimowo:
@yuluo-yx please prepare CP for 0.16

PR #11296, opened 2026-05-18
Prevent finishing workload immediately after creation.

### LINKED ISSUE #10901: Workloads corresponding to Jobs/Jobsets may be marked finished immediately after creation

mbobrovskyi:
#### What type of PR is this?
/kind bug

Prevent finishing workload immediately after creation.

Fixes #10901

```release-note
Workloads: Fixed a bug where, with the `FinishOrphanedWorkloads` feature gate enabled,
Workloads could be marked `Finished` immediately after owner Job or JobSet creation.
`FinishOrphanedWorkloads` is now enabled by default as a beta feature, allowing orphaned
Workloads to be immediately finished and release quota after owner deletion.
```

### DIFF
--- a/pkg/controller/jobframework/reconciler.go
+++ b/pkg/controller/jobframework/reconciler.go
@@ -45,6 +45,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	configapi "sigs.k8s.io/kueue/apis/config/v1beta2"
@@ -688,6 +689,24 @@ func (r *JobReconciler) finalizeWorkloads(ctx context.Context, key types.Namespa
 	}
 	for i := range workloads {
 		wl := &workloads[i]
+
+		// Finish the orphaned workload. This can happen when the workload is readmitted
+		// after eviction (e.g., waitForPodsReady) due to the workload waiting retention policy.
+		// We need to finish the workload to prevent such a situation.
+		// For example, a Deployment-owned pod deleted after PodsReady timeout eviction.
+		if features.Enabled(features.FinishOrphanedWorkloads) && controllerutil.HasControllerReference(wl) {
+			log.V(2).Info("Workload is orphaned; finishing to release quota")
+			err := workload.Finish(ctx, r.client, wl, kueue.WorkloadFinishedReasonOwnerNotFound,
+				"The workload's owner no longer exists", r.clock)
+			if err != nil {
+				if client.IgnoreNotFound(err) != nil {
+					log.Error(err, "Finishing orphaned workload")
+					return err
+				}
+				return nil
+			}
+		}
+
 		if err := workload.RemoveFinalizer(ctx, r.client, wl); client.IgnoreNotFound(err) != nil {
 			log.Error(err, "Removing finalizer")
 			return err
@@ -1240,9 +1259,11 @@ func expectedRunningPodSets(ctx context.Context, c client.Client, wl *kueue.Work
 // EquivalentToWorkload checks if the job corresponds to the workload
 func EquivalentToWorkload(ctx context.Context, c client.Client, job GenericJob, wl *kueue.Workload) (bool, error) {
 	owner := metav1.GetControllerOf(wl)
-	// Indexes don't work in unit tests, so we explicitly check for the
-	// owner here.
 	if owner.Name != job.Object().GetName() {
+		return false, nil
+	}
+
+	if features.Enabled(features.FinishOrphanedWorkloads) && owner.UID != job.Object().GetUID() {
 		return false, nil
 	}
 
--- a/pkg/controller/jobs/pod/pod_controller_test.go
+++ b/pkg/controller/jobs/pod/pod_controller_test.go
@@ -3734,8 +3734,8 @@ func TestReconciler(t *testing.T) {
 				},
 			},
 		},
-		"finalize workload for non existent pod": {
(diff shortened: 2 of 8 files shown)


pkg/controller/jobframework/reconciler.go
```diff
@@ -319,10 +319,26 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
...
 		}
 		for i := range workloads.Items {
-			err := workload.RemoveFinalizer(ctx, r.client, &workloads.Items[i])
+			wl := &workloads.Items[i]
+
+			// Finish the orphaned workload. This can happen when the workload is readmitted
+			// after eviction (e.g., waitForPodsReady) due to the workload waiting retention policy.
+			// We need to finish the workload to prevent such a situation.
+			// For example, a Deployment-owned pod deleted after PodsReady timeout eviction.
+			if features.Enabled(features.FinishOrphanedWorkloads) {
```
mwielgus: Why we may have orphaned an orphaned workload here?

mbobrovskyi: When a pod is deleted (e.g., a Deployment pod after a waitForPodsReady timeout), there can be a gap before the workload is deleted by the garbage collector. During that time, the workload may be readmitted by the scheduler. We need to transition it to Finished to ensure this does not happen.

pkg/features/kube_features.go
```diff
@@ -601,7 +601,8 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Alpha},
 	},
 	FinishOrphanedWorkloads: {
-		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Alpha},
+		// TODO: Revert
+		{Version: version.MustParse("0.18"), Default: true, PreRelease: featuregate.Beta},
```
mbobrovskyi, line 629: /hold

pkg/features/kube_features.go
```diff
@@ -626,7 +626,8 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Alpha},
 	},
 	FinishOrphanedWorkloads: {
-		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Alpha},
+		// TODO: Revert
```
mimowo: I think we can actually go to Beta with the fix. 

For cherrypicking we can just keep in Alpha, but for 0.18 this is reasonable to move to beta as this is fixing deployments behavior

pkg/controller/jobs/pod/pod_controller_test.go
```diff
@@ -3734,8 +3734,8 @@ func TestReconciler(t *testing.T) {
 				},
 			},
 		},
-		"finalize workload for non existent pod": {
-			featureGates: map[featuregate.Feature]bool{features.WorkloadIdentifierAnnotations: false},
```
mimowo, line 3738: why modifying the pre-existing test?

mimowo:
Release note proposal:
/release-note-edit
```release-note
Workloads: Fixed a bug where, with the `FinishOrphanedWorkloads` feature gate enabled,
Workloads could be marked `Finished` immediately after owner Job or JobSet creation.
`FinishOrphanedWorkloads` is now enabled by default as a beta feature, allowing orphaned
Workloads to be immediately finished and release quota after owner deletion.
```
We can drop the second sentence "`FinishOrphanedWorkloads` is now enabled by default" when CPicking .

pkg/features/kube_features.go
mbobrovskyi: Done

mbobrovskyi: /unhold

pkg/controller/jobs/pod/pod_controller_test.go
mbobrovskyi: Actually, not sure why we need to disable this feature.

mimowo: we probably don't need to disable it indeed, but we disable it in many places, so this should be an ortoghonal PR, don't mix this change with this PR. cc @ivnovakov

ivnovakov: I'll take look into that.

mimowo:
/lgtm
/approve
Thank you for the fix 👍 With this fix we can again move FinishOrphanedWorkloads to Beta. 

I think we should cherrypick this fix for users who need FinishOrphanedWorkloads working properly on 0.17, and 0.16 but I don't think we need to graduate it there.

PR #11304, opened 2026-05-19
MultiKueue: detect a hung remote in 1 minute, not 10

### LINKED ISSUE #11303: MultiKueue: hung remote takes 10 minutes to detect, should be 1
**What happened**:

PR #11207 set `defaultWatchEstablishTimeout` to a static 10 minutes. That covers the apiserver cold cache plus conversion webhook scenario from kubernetes/kubernetes#136950 (about 8 min at 50k Workloads), but it also means a truly hung remote keeps a reconciler worker parked for 10 full minutes before the timeout fires.

A tight initial timeout (around 1 minute) so hung remotes fail fast, growing on consecutive failures so a slow but recovering remote still gets the time it needs.

**Proposal**:

Drive the timeout off the existing `rc.failedConnAttempts` counter.

Schedule: 1m, 2m, 4m, 8m, 10m, 10m, ...

`failedConnAttempts` already resets to zero on a successful establishment and resets on config change, so there is no new state to manage. It stacks naturally with the `retryAfter(rc.failedConnAttempts)` reconnect backoff from #10990: both grow together when the same remote keeps failing.

**Anything else**:

Committed to in PR #11207 (https://github.com/kubernetes-sigs/kueue/pull/11207#discussion_r3260697405).

Today's Kueue (v1beta2 served = stored) does not actually hit the slow apiserver cold cache path, so practical impact is limited until version skew returns. Filing this so the design discussion can happen on its own thread.

**Environment**:

N/A. Code level enhancement.

trilamsr:
#### What type of PR is this?

/kind feature
/area multikueue

Follow up to #11207. Replaces the static 10 min `watchEstablishTimeout` with a schedule that starts at 1 min and doubles on each consecutive `failedConnAttempts`, capped at the existing 10 min.

Schedule: 1m, 2m, 4m, 8m, 10m, 10m, ...

`failedConnAttempts` already resets to zero on a successful establishment and on config change, so there is no new state. The new schedule also stacks naturally with the `retryAfter(rc.failedConnAttempts)` reconnect backoff from #10990: both grow together when the same remote keeps failing.

Why bother: under the static 10 min, a truly hung remote keeps a reconciler worker parked for 10 full minutes before the timeout fires. With the new schedule, the first attempt fails fast (1 min), so hung remotes get caught quickly while slow but recovering remotes still get the time they need.

Today Kueue's served and storage versions match (v1beta2), so the apiserver cold cache path that motivated the generous 10 min cap is not exercised. The cap stays as a guard against future version skew and unknown apiserver behavior (kubernetes/kubernetes#136950).

Fixes #11303

- Refactors `establishWatch` to take an explicit `timeout` parameter instead of reading from a package var. Tests now pass `testTimeout` directly, which is cleaner than the previous var mutation pattern.
- The schedule comes from `pkg/util/wait.NewBackoff` (the kueue-wide helper) rather than a hand rolled function, per review feedback.
- New `TestEstablishBackoffSchedule` pins the 1m/2m/4m/8m/10m schedule produced by our configured `establishBackoff`. The helper itself is tested upstream.
- No behavior change for the success path or for already healthy clusters. Only the worst case (hung remote, no prior failures) changes: was 10 min, now 1 min.

```release-note
MultiKueue: Fixed a bug where a hung watch connection to one remote cluster could block
reconciliation of other MultiKueueClusters, leaving them inactive and preventing workload
admission. Kueue now applies a circuit-breaking timeout while establishing remote-cluster
watches: the timeout starts at 1 minute and backs off exponentially on consecutive failures,
up to 10 minutes.
```

### DIFF
--- a/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
+++ b/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
@@ -64,6 +64,7 @@ import (
 	"sigs.k8s.io/kueue/pkg/controller/jobframework"
 	"sigs.k8s.io/kueue/pkg/features"
 	"sigs.k8s.io/kueue/pkg/util/roletracker"
+	utilwait "sigs.k8s.io/kueue/pkg/util/wait"
 )
 
 const (
@@ -78,13 +79,14 @@ const (
 	// covers the apiserver cold cache + conversion warmup path documented in
 	// kubernetes/kubernetes#136950 (~8 min at 50k Workloads). v1beta2 does
 	// not exercise that path today; the cap is a guard. See #11206.
-	initialEstablishTimeout  = 1 * time.Minute
-	maxEstablishTimeout      = 10 * time.Minute
-	establishTimeoutMaxSteps = 4
+	initialEstablishTimeout = 1 * time.Minute
+	maxEstablishTimeout     = 10 * time.Minute
 )
 
 var (
 	errWatchEstablishTimeout = errors.New("watch establishment timed out")
+
+	establishBackoff = utilwait.NewBackoff(initialEstablishTimeout, maxEstablishTimeout, 2, 0)
 )
 
 // retryAfter returns an exponentially increasing interval between
@@ -94,16 +96,6 @@ func retryAfter(failedAttempts uint) time.Duration {
 		return 0
 	}
 	return (1 << (min(failedAttempts, retryMaxSteps) - 1)) * retryIncrement
-}
-
-// establishTimeoutFor doubles initialEstablishTimeout per failure, capped at
-// maxEstablishTimeout. Schedule: 1m, 2m, 4m, 8m, 10m, 10m, ...
-func establishTimeoutFor(failedAttempts uint) time.Duration {
-	t := initialEstablishTimeout << min(failedAttempts, establishTimeoutMaxSteps)
-	if t > maxEstablishTimeout {
-		return maxEstablishTimeout
-	}
-	return t
 }
 
 type clientWithWatchBuilder func(config *clientConfig, options client.Options) (client.WithWatch, error)
@@ -300,7 +292,7 @@ func establishWatch(ctx context.Context, c client.WithWatch, obj client.ObjectLi
 
 func (rc *remoteClient) startWatcher(ctx context.Context, kind string, w jobframework.MultiKueueWatcher) error {
 	log := ctrl.LoggerFrom(ctx).WithValues("watchKind", kind)
-	newWatcher, err := establishWatch(ctx, rc.client, w.GetEmptyList(), rc.origin, establishTimeoutFor(rc.failedConnAttempts))
+	newWatcher, err := establishWatch(ctx, rc.client, w.GetEmptyList(), rc.origin, establishBackoff.WaitTime(int(rc.failedConnAttempts)+1))
 	if err != nil {
 		return err
 	}
--- a/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
+++ b/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
@@ -73,26 +73,17 @@ const (
 	retryIncrement = 5 * time.Second
 	retryMaxSteps  = 7
 
-	// Bounds how long startWatcher waits for client.Watch() to return,
-	// so a hung remote cannot head-of-line block the reconciler. See #11206.
-	//
-	// The bound has to cover the slowest legitimate case, not just the common
-	// one. client.Watch() returns as soon as the apiserver sends HTTP 200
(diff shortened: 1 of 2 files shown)


trilamsr:
@mimowo here's the exponential timeout follow up I promised on #11207. Drops first attempt to 1 min so a hung remote gets caught fast, keeps the 10 min cap on retries via `failedConnAttempts`. WDYT?

mimowo:
cc @kshalot @olekzabl ptal

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
```diff
@@ -105,6 +96,16 @@ func retryAfter(failedAttempts uint) time.Duration {
...
 
+// establishTimeoutFor doubles initialEstablishTimeout per failure, capped at
+// maxEstablishTimeout. Schedule: 1m, 2m, 4m, 8m, 10m, 10m, ...
+func establishTimeoutFor(failedAttempts uint) time.Duration {
+	t := initialEstablishTimeout << min(failedAttempts, establishTimeoutMaxSteps)
+	if t > maxEstablishTimeout {
+		return maxEstablishTimeout
+	}
+	return t
+}
```
mimowo: This looks ok-ish, but could we instead use our helper for exponential backoffs inside: https://github.com/kubernetes-sigs/kueue/blob/main/pkg/util/wait/backoff.go#L63

I'm not saying the code is complex, but the helpers have already been tested pretty well.

mimowo: Also, I would like to give the consistent picture of the codebase so that we don't re-implement the same thing over again. Re-implementation may be simple in this case, but might be more challenging in another place.

kshalot: +1, even if not `UntilWithBackoff`, there's already a `Backoff` struct that calculates the timeout given the attempt:
https://github.com/kubernetes-sigs/kueue/blob/91387fe1625d1c4250555497bde1add4b40f04d0/pkg/util/wait/backoff.go#L33-L58

In our case we'd have sth like:
```go
b := NewBackoff(1m, 10m, 2, 0)
timeout := b.WaitTime(rc.failedConnAttempts)
```

trilamsr: Done. Swapped the `establishTimeoutFor` for `wait.NewBackoff(initialEstablishTimeout, maxEstablishTimeout, 2, 0)` and call `establishBackoff.WaitTime(int(rc.failedConnAttempts)+1)` at the call site. PTAL

mimowo:
I'm tempted to merge this as a cherrypick as part of the fix for https://github.com/kubernetes-sigs/kueue/pull/11207, as this feels closely connected, wdyt @trilamsr ?

My reason is that this PR isn't really a feature on its own, it is just an improved version of the bugfix we have.

The previous PR hasn't been yet released so I would just turn its release-note to NONE, and describe the change here.

trilamsr:
> I'm tempted to merge this as a cherrypick as part of the fix for #11207, as this feels closely connected, wdyt @trilamsr ?
> 
> My reason is that this PR isn't really a feature on its own, it is just an improved version of the bugfix we have.
> 
> The previous PR hasn't been yet released so I would just turn its release-note to NONE, and describe the change here.

Yeah, agreed. They're really one fix in two commits. Combined release note:

```release-note
MultiKueue: a hung watch connection to one remote cluster no longer blocks reconciliation of other MultiKueueClusters. The controller bounds remote-watch establishment with an exponential timeout starting at 1 min and growing up to 10 min on consecutive failures, so a stuck remote is detected quickly while a slow but recovering one still gets the time it needs.
```

I'll flip #11207's note to NONE. /cherrypick release-0.17 release-0.16 here when you're ready?

mimowo:
/lgtm
/approve
Thank you for hardening the MultiKueue implementation 👍 
/cherrypick release-0.17
/cherrypick release-0.16
/remove-kind feature
/kind bug

mimowo:
> Yeah, agreed. They're really one fix in two commits.

Ok, cool.
> Combined release note:

That sounds ok, but let me propose something a bit more explictily indicating it is a bugfix:

/release-note-edit
```release-note
MultiKueue: Fixed a bug where a hung watch connection to one remote cluster could block
reconciliation of other MultiKueueClusters, leaving them inactive and preventing workload
admission. Kueue now applies a circuit-breaking timeout while establishing remote-cluster
watches: the timeout starts at 1 minute and backs off exponentially on consecutive failures,
up to 10 minutes.
```

mimowo:
> I'll flip https://github.com/kubernetes-sigs/kueue/pull/11207's note to NONE. /cherrypick release-0.17 release-0.16 here when you're ready?

Let me do that now.

PR #11305, opened 2026-05-19
MultiKueue: one slow remote no longer stalls every other cluster

### LINKED ISSUE #11297: MultiKueue: GroupKindConcurrency has no effect because of a controller wide lock
**What happened**:

The `GroupKindConcurrency` knob is wired up for the `MultiKueueCluster` controller, but setting it above 1 does not actually buy you parallelism. Any one slow or hung remote stalls reconciles for every other MultiKueueCluster.

The reason: `Reconcile` calls `clustersReconciler.setRemoteClientConfig`, which takes the controller wide write lock `clustersReconciler.lock` ([`pkg/controller/admissionchecks/multikueue/multikueuecluster.go:394`](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/admissionchecks/multikueue/multikueuecluster.go#L394)) and holds it through the entire synchronous `setConfig` → `startWatcher` → `establishWatch` → `client.Watch()` chain ([line 460](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/admissionchecks/multikueue/multikueuecluster.go#L460)). While one worker is parked inside `client.Watch()`, every other reconcile that wants to do anything with a different cluster blocks on the same lock.

So concurrency=N collapses back to 1 whenever any remote is slow or hung.

Setting `controller.groupKindConcurrency["MultiKueueCluster.kueue.x-k8s.io"]` to N should give up to N reconciles running in parallel. When one cluster's reconcile is slow or hung, the other N-1 workers should keep going on other clusters.

1. Run Kueue with `controller.groupKindConcurrency["MultiKueueCluster.kueue.x-k8s.io"]` set to something greater than 1 (e.g. 5).
2. Add two or more `MultiKueueCluster` objects, with one of them pointing at an unreachable or unresponsive remote apiserver.
3. Watch the healthy clusters: they also stall (no `Active=True`, no status updates) for as long as the unhealthy one is hung inside watch establishment.

With #11207 merged this stall is bounded to 10 minutes per cycle instead of indefinite, but the head of line blocking across clusters still happens during that window.

Surfaced during review of #11207. Mimowo flagged it as "this renders `GroupKindConcurrency` useless for `MultiKueueCluster` due to very technical reasons which can be considered a bug by a user" ([PR thread](https://github.com/kubernetes-sigs/kueue/pull/11207#discussion_r3248735173)).

The 10 minute timeout from #11207 only addresses the worst case (indefinite block). The underlying serialization is still there.

Filing this to track the lock refactor.

**Environment**:

N/A. Architectural bug, reproducible on `main` at the time of #11207.

trilamsr:
#### What type of PR is this?

/kind bug
/area multikueue

Fixes #11297. One slow or unresponsive remote used to stall reconciles for every other MultiKueueCluster because `clustersReconciler.lock` was held across the synchronous remote watch establishment in `setConfig`. Bumping `GroupKindConcurrency` did not help, since every worker ended up waiting on that same lock.

This narrows `c.lock` to just the `remoteClients` map find or insert, and adds a per cluster `setConfigLock` on `remoteClient` for the slow path. Different clusters now reconcile in parallel under their own locks.

Fixes #11297

Test driven. `TestSetRemoteClientConfigDoesNotBlockOtherClusters` was written first and failed on `main` (hits the 2 second timeout). After the fix it passes in about 40ms. Race detector clean on the full multikueue suite.

`stopAndRemoveCluster`, `controllerFor`, and `getRemoteClients` already only hold `c.lock` briefly, so they did not need to change. Same key reconciles still serialize through the workqueue dedup contract, so `setRemoteClientConfig` and `stopAndRemoveCluster` cannot race for the same cluster name. The per cluster lock is mostly defensive.

Pairs with #11304 (exponential watch establish timeout). Same head of line scenario, different angle.

```release-note
MultiKueue: Fixed a bug where one slow or unresponsive remote cluster could stall
reconciliation for other MultiKueueClusters, even when
`controller.groupKindConcurrency["MultiKueueCluster.kueue.x-k8s.io"]` was set above 1.
This could delay or block admission through other healthy clusters.
```

### DIFF
--- a/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
+++ b/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
@@ -120,6 +120,10 @@ type remoteClient struct {
 	failedConnAttempts   uint
 	retryConnNextAttempt metav1.Time
 
+	// Held during setConfig. Without it, one stuck remote would stall every
+	// other cluster's reconcile via clustersReconciler.lock. See #11297.
+	setConfigLock sync.Mutex
+
 	clock clock.Clock
 
 	// For unit testing only. There is now need of creating fully functional remote clients in the unit tests
@@ -490,8 +494,6 @@ func (c *clustersReconciler) stopAndRemoveCluster(clusterName string) {
 
 func (c *clustersReconciler) setRemoteClientConfig(ctx context.Context, clusterName string, config *clientConfig, origin string) (*time.Duration, error) {
 	c.lock.Lock()
-	defer c.lock.Unlock()
-
 	client, found := c.remoteClients[clusterName]
 	if !found {
 		client = newRemoteClient(c.localClient, c.wlUpdateCh, c.watchEndedCh, origin, clusterName, c.adapters)
@@ -500,6 +502,10 @@ func (c *clustersReconciler) setRemoteClientConfig(ctx context.Context, clusterN
 		}
 		c.remoteClients[clusterName] = client
 	}
+	c.lock.Unlock()
+
+	client.setConfigLock.Lock()
+	defer client.setConfigLock.Unlock()
 
 	clientLog := ctrl.LoggerFrom(c.rootContext).WithValues("clusterName", clusterName)
 	clientCtx := ctrl.LoggerInto(c.rootContext, clientLog)
--- a/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
+++ b/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
@@ -492,8 +492,12 @@ func (c *clustersReconciler) stopAndRemoveCluster(clusterName string) {
 	}
 }
 
-func (c *clustersReconciler) setRemoteClientConfig(ctx context.Context, clusterName string, config *clientConfig, origin string) (*time.Duration, error) {
+// findOrCreateRemoteClient returns the remoteClient for clusterName, creating
+// one if absent. Only the brief map operation runs under c.lock.
+func (c *clustersReconciler) findOrCreateRemoteClient(clusterName, origin string) *remoteClient {
 	c.lock.Lock()
+	defer c.lock.Unlock()
+
 	client, found := c.remoteClients[clusterName]
 	if !found {
 		client = newRemoteClient(c.localClient, c.wlUpdateCh, c.watchEndedCh, origin, clusterName, c.adapters)
@@ -502,7 +506,11 @@ func (c *clustersReconciler) setRemoteClientConfig(ctx context.Context, clusterN
 		}
 		c.remoteClients[clusterName] = client
 	}
-	c.lock.Unlock()
+	return client
+}
+
+func (c *clustersReconciler) setRemoteClientConfig(ctx context.Context, clusterName string, config *clientConfig, origin string) (*time.Duration, error) {
+	client := c.findOrCreateRemoteClient(clusterName, origin)
 
 	client.setConfigLock.Lock()
 	defer client.setConfigLock.Unlock()
--- a/pkg/controller/admissionchecks/multikueue/multikueuecluster_test.go
+++ b/pkg/controller/admissionchecks/multikueue/multikueuecluster_test.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


trilamsr:
@mimowo took a shot at the lock refactor from #11297 since I was already in the file. TDD path: failing concurrency test on `main` (2 second head of line block), passes after narrowing the controller wide lock. Pairs with #11304, same head of line scenario from a different angle. Happy for you or anyone else to pick it apart.

mimowo:
/ok-to-test
cc @kshalot @olekzabl ptal

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
```diff
@@ -497,8 +501,6 @@ func (c *clustersReconciler) stopAndRemoveCluster(clusterName string) {
 
 func (c *clustersReconciler) setRemoteClientConfig(ctx context.Context, clusterName string, config *clientConfig, origin string) (*time.Duration, error) {
```
mimowo: Let's split this function into two, so that the we can use the `c.lock.Lock(); defer c.lock.Unlock()` for both of them

trilamsr: Done in 37cac3f22. Factored the map find or insert out into `findOrCreateRemoteClient`, so both functions use the simple `Lock(); defer Unlock()` pattern. PTAL.

pkg/controller/admissionchecks/multikueue/multikueuecluster_test.go
```diff
@@ -1244,3 +1245,76 @@ func TestEstablishWatch(t *testing.T) {
 		}
 	})
 }
+
+// Regression for #11297. A remote stuck inside Watch must not stop
+// reconciles of other clusters.
+func TestSetRemoteClientConfigDoesNotBlockOtherClusters(t *testing.T) {
```
mimowo, line 1274: Nice test, thanks for adding 👍

pkg/controller/admissionchecks/multikueue/multikueuecluster_test.go
```diff
@@ -1244,3 +1245,76 @@ func TestEstablishWatch(t *testing.T) {
...
+	reconciler.rootContext = ctx
+	reconciler.builderOverride = gatedBuilder
+
+	slowConfig := &clientConfig{Kubeconfig: []byte(testKubeconfig("slow-user"))}
+	fastConfig := &clientConfig{Kubeconfig: []byte(testKubeconfig("fast-user"))}
+
+	slowDone := make(chan struct{})
+	go func() {
+		defer close(slowDone)
+		_, _ = reconciler.setRemoteClientConfig(ctx, "cluster-slow", slowConfig, defaultOrigin)
```
mimowo: I'm wondering about using the Reconcile as top-level function for the verification. Maybe in the future we would like to start the watches from another place in code, so using the public interface which is never going to change will be more future proof for incoming refactorings.

pkg/controller/admissionchecks/multikueue/multikueuecluster_test.go
```diff
@@ -1244,3 +1245,76 @@ func TestEstablishWatch(t *testing.T) {
...
+	}()
+
+	select {
+	case <-slowReached:
+	case <-time.After(2 * time.Second):
```
mimowo: Let's name the magic consts for 2s to something meaningfull, like `slowUserTimeout`, or `stuckWatchTimeout`.

Same for other timeouts.

Let's also use 5s for the timeouts so that we don't risk the tests failing on a loaded CI.

mimowo (commented):
Thank you, the code changes at the test look great. Left rather minor remarks to make the test more readable and future proof.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
```diff
@@ -507,6 +513,14 @@ func (c *clustersReconciler) setRemoteClientConfig(ctx context.Context, clusterN
...
 		c.remoteClients[clusterName] = client
 	}
+	return client
+}
+
+func (c *clustersReconciler) setRemoteClientConfig(ctx context.Context, clusterName string, config *clientConfig, origin string) (*time.Duration, error) {
+	client := c.findOrCreateRemoteClient(clusterName, origin)
+
+	client.setConfigLock.Lock()
+	defer client.setConfigLock.Unlock()
```
kshalot, line 516: IIUC, in theory this lock is redundant, because I think there's only one reconcile running per `MultiKueueCluster` at any given point, so in theory there's at most one thread accessing `c.remoteClients[clusterName]`.

But this would be very implicit, so having this lock just in case sounds good to me.

kshalot (commented):
/lgtm

Thanks for taking the time to fix this! I don't have any additional comments.

pkg/controller/admissionchecks/multikueue/multikueuecluster_test.go
trilamsr: Done in d0303a74b. Switched the test to drive through `Reconcile` so it survives any future refactor of where watches get triggered from.

pkg/controller/admissionchecks/multikueue/multikueuecluster_test.go
trilamsr: Done in d0303a74b. Named the budget `stuckWatchTimeout = 5 * time.Second` with a doc note about CI flakiness.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
trilamsr: Agreed, the workqueue dedup makes it redundant in production. Kept as a local safety net because that guarantee is invisible from this file, and tests can bypass the workqueue (the one in this PR does).

mimowo:
Trying to make the release note a bit more explicit that this is a bugfix:
/release-note-edit
```release-note
MultiKueue: Fixed a bug where one slow or unresponsive remote cluster could stall
reconciliation for other MultiKueueClusters, even when
`controller.groupKindConcurrency["MultiKueueCluster.kueue.x-k8s.io"]` was set above 1.
This could delay or block admission through other healthy clusters.
```

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
mimowo: lgtm

mimowo:
/lgtm
/approve
Thank you for the quick follow up on this issue 👍 Blocking all reconciles on the lock wasn't good for sure.
/cherrypick release-0.17
/cherrypick release-0.16

mimowo:
This one requires rebase now

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
mimowo: Good point, I missed that. Having said that I'm ok with either option, both have some merit: explicit check + a bit more robust tests, vs. less production code. 

I'm ok to keep as is.

mimowo:
/lgtm
assuming the tests will pass

PR #11309, opened 2026-05-19
test : Removed the deprecated --visibility-server-port in favour of visibilityServer.bindPort

### LINKED ISSUE #11274: Optimize `kueue-test-e2e-sequential-baseline-main`

This is a follow up to https://github.com/kubernetes-sigs/kueue/issues/10995 as `kueue-test-e2e-sequential-baseline-main` is currently one of the longest suites

I reviewed the existing tests and I think the following optimizations can be made:
1. [x] Remove the deprecated `--visibility-server-port` in favor of `visibilityServer.bindPort`, see deprecation note [here](https://kueue.sigs.k8s.io/docs/tasks/manage/monitor_pending_workloads/pending_workloads_on_demand/#monitor-pending-workloads-on-demand)

Expected saving: ~30s

2. [x] Consolidate Failure Recovery tests to share one kubelet outage

The two Failure Recovery specs currently stop and restart kubelet independently. Refactor them so both scenarios use pods on the same worker node and share a single kubelet stop/start cycle. 

Expected saving: ~2min

PR: https://github.com/kubernetes-sigs/kueue/pull/11339

3. [x] Make default-config assumptions explicit and remove redundant visibility config restore

The visibility config API test currently restores the original Kueue configuration in cleanup, costing ~27–30s.

Instead, make the default-configuration invariant explicit: group tests that assume the default config, such as certs and HA failover, under a default-config setup that restores `defaultKueueCfg` once in `BeforeAll`.

Then remove the redundant visibility-test cleanup:

- `ginkgo.By("Restoring original configuration")`
- `UpdateKueueConfigurationAndRestart(... defaultKueueCfg ...)`

Expected saving: ~30s.

**Why is this needed**:

It would be great to have all suites running below 25min.

ekam-walia:
#### What type of PR is this?
/kind cleanup
/area testing

This PR helps to reduce the CI clock time by approximately 30 seconds.
This PR removes the deprecated --visibility-server-port in favour of visibilityServer.bindPort

Fixes #11274
Part of #11274

I have removed the --visibility-server-port from visibility_server_test.go file in baseline folder of sequential tests 

```release-note
Removed the deprecated `--visibility-server-port` flag from the Kueue controller manager.

ACTION REQUIRED: If your installation passes `--visibility-server-port` to the
controller manager, remove the flag before upgrading and set the same port via 
Kueue Configuration API in `visibilityServer.bindPort` instead.
```

### DIFF
--- a/cmd/kueue/main.go
+++ b/cmd/kueue/main.go
@@ -132,9 +132,6 @@ func main() {
 	var featureGates string
 	flag.StringVar(&featureGates, "feature-gates", "", "A set of key=value pairs that describe feature gates for alpha/experimental features.")
 
-	var visibilityServerPort int
-	flag.IntVar(&visibilityServerPort, "visibility-server-port", configapi.DefaultVisibilityBindPort, "The port the visibility server binds to.")
-
 	customLogProcessor := zaplog.WrapCore(utillogging.NewCustomLogProcessor)
 
 	zapOptions := zap.Options{
@@ -373,7 +370,7 @@ func main() {
 
 	if features.Enabled(features.VisibilityOnDemand) {
 		go func() {
-			if err := visibility.CreateAndStartVisibilityServer(ctx, queues, &cfg, kubeConfig, visibilityServerPort, parsedTLSConfig); err != nil {
+			if err := visibility.CreateAndStartVisibilityServer(ctx, queues, &cfg, kubeConfig, parsedTLSConfig); err != nil {
 				setupLog.Error(err, "Unable to create and start visibility server")
 				os.Exit(1)
 			}
--- a/cmd/kueue/main.go
+++ b/cmd/kueue/main.go
@@ -370,7 +370,7 @@ func main() {
 
 	if features.Enabled(features.VisibilityOnDemand) {
 		go func() {
-			if err := visibility.CreateAndStartVisibilityServer(ctx, queues, &cfg, kubeConfig, parsedTLSConfig); err != nil {
+			if err := visibility.CreateAndStartVisibilityServer(ctx, queues, &cfg, kubeConfig, configapi.DefaultVisibilityBindPort, parsedTLSConfig); err != nil {
 				setupLog.Error(err, "Unable to create and start visibility server")
 				os.Exit(1)
 			}
--- a/test/e2e/sequential/baseline/visibility_server_test.go
+++ b/test/e2e/sequential/baseline/visibility_server_test.go
@@ -18,7 +18,6 @@ package baseline
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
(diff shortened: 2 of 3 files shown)


mimowo:
/ok-to-test
We should also drop the impl and the docs about it

test/e2e/sequential/baseline/visibility_server_test.go
```diff
@@ -245,42 +244,6 @@ var _ = ginkgo.Describe("Visibility Server", ginkgo.Label("feature:visibility"),
 		}, util.Timeout, util.Interval).Should(gomega.Succeed())
 	})
 
-	ginkgo.It("Should use the custom port from the --visibility-server-port flag", func() {
```
mimowo, line 248: The post is now read from configuration here: https://github.com/kubernetes-sigs/kueue/blob/main/pkg/visibility/server.go#L118-L119

So we can drop the deprecated `--visibility-server-port` flag complete, starting from main.go: https://github.com/kubernetes-sigs/kueue/blob/main/cmd/kueue/main.go#L135-L136

and do not pass it down

mimowo, line 248: Also drop documentation here: https://github.com/kubernetes-sigs/kueue/blob/edc0bc5c4d17081fc30a13165d42f542b592539b/site/content/en/docs/tasks/manage/monitor_pending_workloads/pending_workloads_on_demand.md?plain=1#L79

mimowo:
/release-note-edit
```release-note
Removed the deprecated `--visibility-server-port` flag from the Kueue controller manager.

ACTION REQUIRED: If your installation passes `--visibility-server-port` to the
controller manager, remove the flag before upgrading and set the same port via 
Kueue Configuration API in `visibilityServer.bindPort` instead.
```

cmd/kueue/main.go
```diff
@@ -373,7 +370,7 @@ func main() {
 
 	if features.Enabled(features.VisibilityOnDemand) {
 		go func() {
-			if err := visibility.CreateAndStartVisibilityServer(ctx, queues, &cfg, kubeConfig, visibilityServerPort, parsedTLSConfig); err != nil {
+			if err := visibility.CreateAndStartVisibilityServer(ctx, queues, &cfg, kubeConfig, configapi.DefaultVisibilityBindPort, parsedTLSConfig); err != nil {
```
mimowo, line 373: Do not pass the default here.

Instead let's adjust the code in https://github.com/kubernetes-sigs/kueue/blob/main/pkg/visibility/server.go#L115-L122
```golang
	// Apply visibility overrides from Configuration API if present and non-default.
	// If the API value is non-default, it takes precedence.
	// Otherwise, use the flag.
	if cfg.VisibilityServer != nil && cfg.VisibilityServer.BindPort != nil && *cfg.VisibilityServer.BindPort != configapi.DefaultVisibilityBindPort {
		o.SecureServing.BindPort = int(*cfg.VisibilityServer.BindPort)
	} else {
		o.SecureServing.BindPort = port
	}
```
to something like:
```golang
	if cfg.VisibilityServer != nil  {
		o.SecureServing.BindPort = ptr.Deref(cfg.VisibilityServer.BindPort, configapi.DefaultVisibilityBindPort)
	}
```

mbobrovskyi:
> Fixes https://github.com/kubernetes-sigs/kueue/issues/11274

Please remove the `fixes` entry from the PR description to avoid accidentally closing the issue.

mimowo:
/lgtm
/approve
Let me merge as this is already offloading our tests, the [comment](https://github.com/kubernetes-sigs/kueue/pull/11309#discussion_r3264199219) can be addressed in a follow up.

PR #11323, opened 2026-05-19
visibility: follow-up cleanup for bind port defaulting

ekam-walia:
#### What type of PR is this?
/kind cleanup

Follow-up cleanup for #11309, which removed the deprecated --visibility-server-port flag in favor of visibilityServer.bindPort.

This PR centralises the visibility server bind port defaulting logic inside pkg/visibility/server.go instead of passing the default value from cmd/kueue/main.go.

The visibility server now internally defaults the bind port using ptr.Deref() with configapi.DefaultVisibilityBindPort.

Follow up for #11309
Part of  #11274

Removed the port parameter from:
1) CreateAndStartVisibilityServer
2) applyVisibilityServerOptions
3) Added bind port defaulting directly in server.go
4) Updated visibility server startup call sites

```release-note
NONE
```

### DIFF
--- a/cmd/kueue/main.go
+++ b/cmd/kueue/main.go
@@ -370,7 +370,7 @@ func main() {
 
 	if features.Enabled(features.VisibilityOnDemand) {
 		go func() {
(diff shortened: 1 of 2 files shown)


mimowo:
/lgtm
/approve
Thank you for the follow up, this concludes the first point from https://github.com/kubernetes-sigs/kueue/issues/11274

PR #11327, opened 2026-05-19
Fix duplicate replacement workload slices stuck during rapid scaling

sohankunkerkar:
#### What type of PR is this?
/kind bug

Fixes #

This bug was discovered during stress testing of the `ElasticJobsViaWorkloadSlices` beta graduation work. The integration test rapidly cycles parallelism 10 times on an elastic job and asserts convergence to exactly 1 active workload. Without the fix, the test fails ~1/20 runs with 2 admitted workloads permanently stuck.

I looped the below test 100 times and it fixes the issue.

```release-note
ElasticJobsViaWorkloadSlices: Fixed a bug where rapid elastic Job scaling could leave
duplicate replacement Workload slices admitted indefinitely, causing quota to remain
reserved.
```

### DIFF
--- a/pkg/workloadslicing/workloadslicing.go
+++ b/pkg/workloadslicing/workloadslicing.go
@@ -19,7 +19,6 @@ package workloadslicing
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"slices"
 
@@ -27,6 +26,7 @@ import (
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/klog/v2"
 	"k8s.io/utils/clock"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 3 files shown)


test/integration/singlecluster/controller/jobs/job/job_controller_test.go
```diff
@@ -4369,6 +4369,64 @@ var _ = ginkgo.Describe("Job with elastic jobs via workload-slices support", gin
...
+		gomega.Eventually(func(g gomega.Gomega) {
+			wlList := &kueue.WorkloadList{}
+			g.Expect(k8sClient.List(ctx, wlList, client.InNamespace(testJob.Namespace))).Should(gomega.Succeed())
+			admittedCount := 0
+			for i := range wlList.Items {
+				if workload.HasQuotaReservation(&wlList.Items[i]) && !workload.IsFinished(&wlList.Items[i]) {
+					admittedCount++
+				}
+			}
+			g.Expect(admittedCount).Should(gomega.Equal(1))
```
mimowo: Instead of assering on count it is generally better to assert on the slice by checking gomega.HasLen(1). In this way we get a better failure message immediately indicating the set of admitted workloads, not just count which can be helpful for debugging based on logs.

test/integration/singlecluster/controller/jobs/job/job_controller_test.go
```diff
@@ -4369,6 +4369,64 @@ var _ = ginkgo.Describe("Job with elastic jobs via workload-slices support", gin
...
+		}, util.Timeout, util.Interval).Should(gomega.Succeed())
+
+		gomega.Eventually(func(g gomega.Gomega) {
+			g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(testJob), testJob)).Should(gomega.Succeed())
+			g.Expect(ptr.Deref(testJob.Spec.Suspend, false)).Should(gomega.BeFalse())
+		}, util.Timeout, util.Interval).Should(gomega.Succeed())
+
+		ginkgo.By("rapidly cycling parallelism 10 times")
+		for i := range 10 {
+			targetParallelism := int32(1 + (i % 4))
```
mimowo, line 4501: That feels a bit magical, iiuc this is to cycle the job sizes, 1->2->3->4->1...

Let's just add a comment to make it immediately clear.

mimowo:
I have a question: Is this PR marked as regression, because https://github.com/kubernetes-sigs/kueue/pull/11195 introduced the regression? If this is the case then maybe don't need to call it regression or have a release note, because the previous PR hasn't been released yet. Would you be able to check if the test is passing prior to https://github.com/kubernetes-sigs/kueue/pull/11195?

test/integration/singlecluster/controller/jobs/job/job_controller_test.go
sohankunkerkar: The formula cycles parallelism through 2→3→4→1→2→... to alternate scale-up and scale-down events. I will add a comment to specify that

sohankunkerkar:
> I have a question: Is this PR marked as regression, because #11195 introduced the regression? If this is the case then maybe don't need to call it regression or have a release note, because the previous PR hasn't been released yet. Would you be able to check if the test is passing prior to #11195?

Yeah, it's a bug and not a regression. Thanks for the pointers!

pkg/workloadslicing/workloadslicing.go
```diff
@@ -226,11 +226,15 @@ func EnsureWorkloadSlices(
 		//    admitted the new slice but failed to finish the old slice.
 		replacementKey := ReplacementForKey(&newWorkload)
 		oldWorkloadKey := workload.Key(&oldWorkload)
+		oldReplacementKey := ReplacementForKey(&oldWorkload)
 		newWorkloadAdmittedAsReplacement := workload.HasQuotaReservation(&newWorkload) &&
 			replacementKey != nil && *replacementKey == oldWorkloadKey
+		bothReplaceTheSameSlice := replacementKey != nil && oldReplacementKey != nil &&
+			*replacementKey == *oldReplacementKey
```
mimowo: This feels a bit ad-hoc, and makes me wonder: is the chain of workload slices not always linear? 

could it be that we get: 

Scenario 1:
```
a1 <- a2 
a1 <- a3
```
and here oldWorkload=a2, and newWorkload=a3 ? 

However, then could it be that we have:

Scenario 2:
```
a1 <- a2 
a1 <- a3
a2 <- a4
```
In that case if oldWorkload=a3, and newWorkload=a4, then they don't have the same previous slice.

Also, even in scenario 1, if we have a2 and a3 - then the JobReconciler will race with scheduler..

So, I'm thinking we should try to make sure the history is linear, if we cannot make the history linear, then I suppose we need a more generic fix.

sohankunkerkar: Yup, agreed! I replaced the whole case 2 + default with a single `normalizeActiveSlices` that enforces a generic invariant: one non-evicted admitted workload, at most one pending replacement pointing to it, evicted workloads always finished. The tricky part was the eviction dimension as a workload can be admitted and evicted at the same time, so LA selection must exclude evicted workloads to preserve the preemption flow.

sohankunkerkar:
failure addressed here: https://github.com/kubernetes-sigs/kueue/pull/11338

pkg/workloadslicing/workloadslicing.go
```diff
@@ -209,72 +208,107 @@ func EnsureWorkloadSlices(
...
+// The input slice must be sorted oldest-first.
+func normalizeActiveSlices(
+	ctx context.Context,
+	clnt client.Client,
+	clk clock.Clock,
+	workloads []kueue.Workload,
+) (*kueue.Workload, error) {
+	log := ctrl.LoggerFrom(ctx)
+
+	var latestAdmitted, pendingReplacement, latestNonEvicted *kueue.Workload
```
mimowo: let's log the workloads, by each category; I think this should be V3 probably, because scaling is a complex operation.

pkg/workloadslicing/workloadslicing_test.go
```diff
@@ -994,6 +1001,150 @@ func TestEnsureWorkloadSlices(t *testing.T) {
 	}
 }
 
+func TestNormalizeActiveSlices(t *testing.T) {
```
mimowo, line 1004: Do the tests cover 100% of the NormalizeActiveSlices function?

pkg/workloadslicing/workloadslicing.go
mimowo: maybe let's rename latestAdmitted -> latestWithQuotaReservetion to reflect the condition we are checking

pkg/workloadslicing/workloadslicing_test.go
```diff
@@ -994,6 +1001,150 @@ func TestEnsureWorkloadSlices(t *testing.T) {
...
+		"admitted + pending replacement, keep replacement": {
+			workloads: []kueue.Workload{
+				*admitted(utiltestingapi.MakeWorkload("wl-a", "ns").ResourceVersion("1").Creation(now.Add(-time.Minute)).
+					PodSets(*utiltestingapi.MakePodSet(kueue.DefaultPodSetName, 1).Request(corev1.ResourceCPU, "1").Obj())).Obj(),
+				*utiltestingapi.MakeWorkload("wl-b", "ns").ResourceVersion("1").Creation(now).
+					Annotation(WorkloadSliceReplacementFor, "ns/wl-a").
+					PodSets(*utiltestingapi.MakePodSet(kueue.DefaultPodSetName, 2).Request(corev1.ResourceCPU, "1").Obj()).Obj(),
+			},
+			want: want{survivor: "wl-b", keptAdmitted: "wl-a"},
+		},
```
mimowo, line 1042: Is this actually the correct behavior? I think we discussed that now we will wait for Admission of the replacement before marking the previously admitted workload as finished?

pkg/workloadslicing/workloadslicing_test.go
sohankunkerkar:
```sh
$ go test ./pkg/workloadslicing/... -run TestNormalizeActiveSlices -coverprofile=/tmp/cover.out -count=1 
ok      sigs.k8s.io/kueue/pkg/workloadslicing   0.126s  coverage: 27.9% of statements
$ go tool cover -func=/tmp/cover.out | grep normalizeActiveSlices
sigs.k8s.io/kueue/pkg/workloadslicing/workloadslicing.go:252:   normalizeActiveSlices           97.1%

```
The 8 test cases cover: two admitted (keep newest), admitted + pending replacement, evicted + pending, forked DAG (both replace same target), all non-admitted, all evicted (nil survivor), orphan pending without annotation, and 3-workload scenario. The uncovered ~3% is the Finish error return path

pkg/workloadslicing/workloadslicing_test.go
sohankunkerkar: Yes, in this case both workloads are already admitted, so finishing the older one is the recovery path. This matches the scheduler's `replaceOldWorkloadSlice` behavior from PR #11195: the new slice was admitted first (that's how it got quota), and the old slice should have been finished by the scheduler but wasn't. `normalizeActiveSlices` cleans up what the scheduler missed.

mimowo: sgtm

pkg/workloadslicing/workloadslicing_test.go
mimowo: sgtm

mimowo:
/release-note-edit
```release-note
ElasticJobsViaWorkloadSlices: Fixed a bug where rapid elastic Job scaling could leave
duplicate replacement Workload slices admitted indefinitely, causing quota to remain
reserved.
```

/lgtm
/approve

Thank you for improving stability of the elastic workloads.
/cherrypick release-0.17
/cherrypick release-0.16

PR #11339, opened 2026-05-20
test: consolidate failure recovery scenarios under a single kubelet outage

### LINKED ISSUE #11274: Optimize `kueue-test-e2e-sequential-baseline-main`

ekam-walia:
#### What type of PR is this?
/kind cleanup
/area testing

This PR consolidates the Failure Recovery Policy e2e scenarios to share a single kubelet stop/start cycle instead of restarting kubelet independently for each spec.

Part of #11274
Fixes #11274

Consolidate Failure Recovery tests to share one kubelet outage
The two Failure Recovery specs currently stop and restart kubelet independently. 
Refactored them so both scenarios use pods on the same worker node and share a single kubelet stop/start cycle.

```release-note
NONE
```

### DIFF
--- a/test/e2e/sequential/baseline/failure_recovery_policy_test.go
+++ b/test/e2e/sequential/baseline/failure_recovery_policy_test.go
@@ -177,11 +177,7 @@ var _ = ginkgo.Describe("Failure Recovery Policy", ginkgo.Label("feature:failure
 			})
 		})
 
-		ginkgo.It("should delete pods running on an unreachable node", func() {
-			util.ExpectObjectToBeDeletedWithTimeout(ctx, k8sClient, pod, false, nodeMonitorGracePeriod+unhealthyNodeforcefulTerminationCheckTimeout+util.MediumTimeout)
-		})
-
-		ginkgo.It("should unblock the stuck pod's parents that are being deleted with foreground propagation", func() {
+		ginkgo.It("should handle failure recovery scenarios during a single kubelet outage", func() {
 			ginkgo.By("waiting for pods to be marked for termination", func() {
 				gomega.Eventually(func(g gomega.Gomega) {
 					g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(pod), pod)).To(gomega.Succeed())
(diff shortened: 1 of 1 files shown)


test/e2e/sequential/baseline/failure_recovery_policy_test.go
```diff
@@ -177,18 +177,24 @@ var _ = ginkgo.Describe("Failure Recovery Policy", ginkgo.Label("feature:failure
 			})
 		})
 
-		ginkgo.It("should delete pods running on an unreachable node", func() {
-			util.ExpectObjectToBeDeletedWithTimeout(ctx, k8sClient, pod, false, nodeMonitorGracePeriod+unhealthyNodeforcefulTerminationCheckTimeout+util.MediumTimeout)
-		})
-
-		ginkgo.It("should unblock the stuck pod's parents that are being deleted with foreground propagation", func() {
+		ginkgo.It("should handle failure recovery scenarios during a single kubelet outage", func() {
```
mimowo: This will now create only one job, I doubt we can reflect both test scenarios this way, because now the pod is deleted and so the foreground deletion of the job will be too simplified.

In order to reflect them more closely we would need to create a pod say: stuckPodToBeDeleted, and a job jobToBeForegroundDeleted.

However I'm wondering if we need the complication maybe now the  E2e test for the more challenging scenario of job foreground deletion is enough, and we can rely on integration tests for the rest. 

Let's assume this direction for now and just skip awaiting for the pod to be deleted before sending foreground deletion to the job.

Wdyt @kshalot?

test/e2e/sequential/baseline/failure_recovery_policy_test.go
mimowo: Inline this, and drop the Pod form L109 - here we only need to get the nodeName for the Node hosting the Pod.

mimowo: This is a bit unrelated, but let's give the Pod more cpu and memory so that it can become "ready" quicker. Let just say `RequestAndLimit(corev1.ResourceCPU, "1").` and `RequestAndLimit(corev1.ResourceMemory, "40Mi").`. This test is not running in parallal, so there is no risk of stealing cpu / mem from other tests.

mimowo: This is now unused, let's drop

ekam-walia: Done, made the changes

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16
Thank you, excited to see the runtime of our CI tests getting under control again 👍

test/e2e/sequential/baseline/failure_recovery_policy_test.go
mimowo: Let me merge as is after the recent corrections. To my knowledge this testing in e2e should be enough as foreground deletion proved to be the most challening test, so the other one is kind of redundant and can be covered by integration tests.

If we notice some scenario missed then we can backport the lost test case scenario as a new Job in the test (as I initially assumed).

PR #11345, opened 2026-05-20
WIP: Use GetOwnerObject() to fetch owner object.

### LINKED ISSUE #10901: Workloads corresponding to Jobs/Jobsets may be marked finished immediately after creation

mbobrovskyi:
#### What type of PR is this?
/kind bug

Use GetOwnerObject() to fetch owner object.

Fixes #10901

```release-note
NONE
```

### DIFF
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -55,6 +55,7 @@ import (
 	schdcache "sigs.k8s.io/kueue/pkg/cache/scheduler"
 	"sigs.k8s.io/kueue/pkg/constants"
 	"sigs.k8s.io/kueue/pkg/controller/core/indexer"
+	"sigs.k8s.io/kueue/pkg/controller/jobframework"
 	"sigs.k8s.io/kueue/pkg/dra"
 	"sigs.k8s.io/kueue/pkg/features"
 	"sigs.k8s.io/kueue/pkg/metrics"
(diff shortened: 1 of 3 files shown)


pkg/features/kube_features.go
```diff
@@ -601,7 +601,7 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Alpha},
 	},
 	FinishOrphanedWorkloads: {
-		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Alpha},
+		{Version: version.MustParse("0.18"), Default: true, PreRelease: featuregate.Beta},
```
mbobrovskyi: /hold

mbobrovskyi: Reverted

mbobrovskyi: /unhold

PR #11347, opened 2026-05-20
Reorganize drop finalizers logic.

mbobrovskyi:
#### What type of PR is this?
/kind cleanup

Reorganize drop finalizers logic.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/controller/jobframework/reconciler.go
+++ b/pkg/controller/jobframework/reconciler.go
@@ -287,12 +287,9 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
 		err = r.ignoreUnretryableError(log, err)
 	}()
 
-	dropFinalizers := false
-	if cJob, isComposable := job.(ComposableJob); isComposable {
-		dropFinalizers, err = cJob.Load(ctx, r.client, &req.NamespacedName)
-	} else {
-		err = r.client.Get(ctx, req.NamespacedName, object)
-		dropFinalizers = apierrors.IsNotFound(err) || !object.GetDeletionTimestamp().IsZero()
+	shouldFinalize, err := r.loadJob(ctx, &req.NamespacedName, job)
+	if err != nil {
+		return ctrl.Result{}, err
 	}
 
 	if jws, implements := job.(JobWithSkip); implements {
@@ -301,42 +298,11 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
 		}
 	}
 
-	if dropFinalizers {
-		// Remove workload finalizer
-		workloads := &kueue.WorkloadList{}
-
-		if cJob, isComposable := job.(ComposableJob); isComposable {
-			var err error
-			workloads, err = cJob.ListChildWorkloads(ctx, r.client, req.NamespacedName)
-			if err != nil {
-				log.Error(err, "Removing finalizer")
-				return ctrl.Result{}, err
-			}
-		} else {
-			if err := r.client.List(ctx, workloads, client.InNamespace(req.Namespace), OwnerReferenceIndexFieldMatcher(job.GVK(), req.Name)); err != nil {
-				log.Error(err, "Unable to list child workloads")
-				return ctrl.Result{}, err
-			}
-		}
-		for i := range workloads.Items {
-			err := workload.RemoveFinalizer(ctx, r.client, &workloads.Items[i])
-			if client.IgnoreNotFound(err) != nil {
-				log.Error(err, "Removing finalizer")
-				return ctrl.Result{}, err
-			}
-		}
-
-		// Remove job finalizer
-		if !object.GetDeletionTimestamp().IsZero() {
-			if err = r.finalizeJob(ctx, job); err != nil {
-				return ctrl.Result{}, err
-			}
+	if shouldFinalize {
+		if err := r.finalize(ctx, req.NamespacedName, job); err != nil {
+			return ctrl.Result{}, err
 		}
 		return ctrl.Result{}, nil
-	}
-
-	if err != nil {
-		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
 	if features.Enabled(features.ManagedJobsNamespaceSelectorAlwaysRespected) {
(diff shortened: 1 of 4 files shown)


pkg/controller/jobframework/reconciler.go
```diff
@@ -287,12 +287,9 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
...
 
-	dropFinalizers := false
-	if cJob, isComposable := job.(ComposableJob); isComposable {
-		dropFinalizers, err = cJob.Load(ctx, r.client, &req.NamespacedName)
-	} else {
-		err = r.client.Get(ctx, req.NamespacedName, object)
-		dropFinalizers = apierrors.IsNotFound(err) || !object.GetDeletionTimestamp().IsZero()
+	finalize, err := r.loadJob(ctx, job)
+	if err != nil {
+		return ctrl.Result{}, nil
```
mimowo: why silencing the error here?

mbobrovskyi: Good catch! THank you!

reruno: maybe `return ctrl.Result{}, err`

mbobrovskyi: Yeah, fixed! Thank you!

pkg/controller/jobframework/reconciler.go
```diff
@@ -657,6 +626,75 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
...
+	workloads, err := r.getWorkloads(ctx, job)
+	if err != nil {
+		return err
+	}
+	for i := range workloads {
+		wl := &workloads[i]
+		err := workload.RemoveFinalizer(ctx, r.client, wl)
+		if client.IgnoreNotFound(err) != nil {
+			log.Error(err, "Removing finalizer")
+			return client.IgnoreNotFound(err)
```
mbobrovskyi:
```suggestion
			return err
```

mbobrovskyi: Let's wait for tests.

pkg/controller/jobframework/reconciler.go
```diff
@@ -301,38 +298,10 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
...
-		// Remove job finalizer
-		if !object.GetDeletionTimestamp().IsZero() {
-			if err = r.finalizeJob(ctx, job); err != nil {
-				return ctrl.Result{}, err
-			}
+	if finalize {
+		if err := r.finalize(ctx, job); err != nil {
+			return ctrl.Result{}, err
 		}
-		return ctrl.Result{}, nil
```
reruno: maybe revert

in origin code after dropFinalizers we return in if statement

pkg/controller/jobframework/reconciler.go
```diff
@@ -657,6 +626,75 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
...
+// loadJob retrieves and loads the specified job resource into memory.
+// Returns true if the job should be finalized and an error if loading fails.
+func (r *JobReconciler) loadJob(ctx context.Context, job GenericJob) (bool, error) {
+	obj := job.Object()
+
+	if cJob, isComposable := job.(ComposableJob); isComposable {
+		return cJob.Load(ctx, r.client, client.ObjectKeyFromObject(obj))
+	}
+
+	err := r.client.Get(ctx, client.ObjectKeyFromObject(obj), obj)
```
reruno:
```
func (r *genericReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
	return r.jr.ReconcileGenericJob(ctx, req, r.newJob())
}

func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Request, job GenericJob) (result ctrl.Result, err error) {
<...>
finalize, err := r.loadJob(ctx, job)
<...>
```

I think this obj is empty, because r.newJob() is passed to loadJob() 

and here `err := r.client.Get(ctx, client.ObjectKeyFromObject(obj), obj)` we try to get from client empty object

pkg/controller/jobframework/reconciler.go
```diff
@@ -301,38 +298,10 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
...
-				return ctrl.Result{}, err
-			}
-		}
-
-		// Remove job finalizer
-		if !object.GetDeletionTimestamp().IsZero() {
-			if err = r.finalizeJob(ctx, job); err != nil {
-				return ctrl.Result{}, err
-			}
+	if finalize {
```
reruno: nit: maybe revert back to dropFinalizers, because there is already finalize symbol in available in the scope  (r.finalize)

mbobrovskyi: Maybe shouldFinalize? WDYT?

pkg/controller/jobframework/reconciler.go
mbobrovskyi: Good catch! Thank you!

pkg/controller/jobframework/reconciler.go
reruno: sounds good, 

also because it is bool value, consider isFinalize if it makes sense

pkg/controller/jobframework/interface.go
```diff
@@ -126,7 +126,7 @@ type JobWithCustomValidation interface {
 type ComposableJob interface {
 	// Load loads all members of the composable job. If removeFinalizers is true,
 	// workload and job finalizers should be removed.
-	Load(ctx context.Context, c client.Client, key *types.NamespacedName) (removeFinalizers bool, err error)
+	Load(ctx context.Context, c client.Client, key types.NamespacedName) (removeFinalizers bool, err error)
```
mbobrovskyi: Oh, I see. We change the key inside pod_controller.

pkg/controller/jobframework/reconciler.go
mbobrovskyi: Yeah, you totally right. Fixed.

mimowo (commented):
/lgtm
/approve
Thank you , I think this is much more readable indeed 👍

PR #11341, opened 2026-05-20
Fix bindAddress regression

weizhoublue:
#### What type of PR is this?

 
/kind bug
 

#### What this PR does / why we need it:

 Trigger Scenario

  A user configures the visibility server with a custom bind address:
  visibilityServer:
    bindAddress: 127.0.0.1  <--- example
    bindPort: 8082

  This happens when users want the embedded visibility API server to listen only on a specific interface, such as loopback, an internal network IP, or a  restricted node interface.

  The configuration is accepted and passes validation, but the visibility server ignores visibilityServer.bindAddress.
  Instead of listening on the configured IP, it keeps the default bind address:  0.0.0.0

  Root Cause
  PR #11323 refactored visibility server port defaulting and moved bindPort handling into pkg/visibility/server.go.
  During that refactor, the existing BindAddress assignment was accidentally removed:

  o.SecureServing.BindAddress = net.ParseIP(*cfg.VisibilityServer.BindAddress)

btw, if the user does not config  visibilityServer. bindPort,   the problem does not appeaer

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #11348, opened 2026-05-20
Generalize visibility server option tests 

weizhoublue:
… config

#### What type of PR is this?

/area testing

This pr is a follow up for PR https://github.com/kubernetes-sigs/kueue/pull/11341  , strengthening the test 

The visibility server options are configured through applyVisibilityServerOptions, especially visibilityServer.bindAddress, visibilityServer.bindPort, and TLS options.

The previous unit test only exercised the helper applyVisibilityServerSecureServingOptions. It did not verify that the full applyVisibilityServerOptions path applies those values into the final RecommendedConfig.

Tests should cover the same path used by the real visibility server startup flow, so regressions in option application are caught at the final config level.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/visibility/server_test.go
+++ b/pkg/visibility/server_test.go
@@ -17,73 +17,180 @@ limitations under the License.
 package visibility
 
 import (
-	"net"
+	"flag"
+	"os"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	genericoptions "k8s.io/apiserver/pkg/server/options"
 
 	configapi "sigs.k8s.io/kueue/apis/config/v1beta2"
 )
 
-func TestApplyVisibilityServerSecureServingOptions(t *testing.T) {
+func TestCreateVisibilityServerOptions(t *testing.T) {
+	enableInternalCertManagement := true
+	disableInternalCertManagement := false
 	bindAddress := "127.0.0.1"
-	bindPort := int32(9444)
+	bindPort := int32(8080)
 
-	tests := map[string]struct {
-		cfg         *configapi.Configuration
-		wantOptions *genericoptions.SecureServingOptionsWithLoopback
+	tests := []struct {
+		name string
+		cfg  *configapi.Configuration
+		want wantVisibilityServerOptions
 	}{
-		"default visibility server config": {
-			cfg:         &configapi.Configuration{},
-			wantOptions: wantSecureServingOptions("", int(configapi.DefaultVisibilityBindPort)),
-		},
-		"custom bind port": {
+		{
+			name: "default visibility server config",
 			cfg: &configapi.Configuration{
-				VisibilityServer: &configapi.VisibilityServerConfiguration{
-					BindPort: &bindPort,
+				InternalCertManagement: &configapi.InternalCertManagement{
+					Enable: &enableInternalCertManagement,
 				},
 			},
-			wantOptions: wantSecureServingOptions("", 9444),
+			want: wantVisibilityServerOptions{
+				BindAddress:   "0.0.0.0",
+				BindPort:      int(configapi.DefaultVisibilityBindPort),
+				CertDirectory: true,
+			},
 		},
-		"custom bind address": {
+		{
+			name: "custom bind address with default port",
 			cfg: &configapi.Configuration{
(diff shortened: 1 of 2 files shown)


weizhoublue:
@mimowo  hi, as you suggested in pr  https://github.com/kubernetes-sigs/kueue/pull/11341 ,  I submit this as a follow-up

pkg/visibility/server_test.go
```diff
@@ -17,73 +17,325 @@ limitations under the License.
...
 
-	tests := map[string]struct {
-		cfg         *configapi.Configuration
-		wantOptions *genericoptions.SecureServingOptionsWithLoopback
+	tests := []struct {
+		name    string
+		cfg     *configapi.Configuration
+		tls     *tlsconfig.TLS
+		prepare func(t *testing.T)
+		want    wantVisibilityServerOptions
```
mimowo: instead of that could we try `wantConfig *genericapiserver.RecommendedConfig` ? 
That should be a more generic approach, but maybe there is an issue asserting on the RecommendedConfig directly?

pkg/visibility/server_test.go
```diff
@@ -17,73 +17,325 @@ limitations under the License.
...
+	}
+	if err := os.WriteFile(certDir+"/tls.crt", cert, 0600); err != nil {
+		t.Fatalf("Writing serving certificate file: %v", err)
+	}
+	if err := os.WriteFile(certDir+"/tls.key", key, 0600); err != nil {
+		t.Fatalf("Writing serving key file: %v", err)
+	}
+}
+
+func writeKubeConfig(t *testing.T, serverURL string) string {
```
mimowo: that feels verbose, could we use TestKubeconfigWrapper instead to construct the kubeconfig?

pkg/visibility/server_test.go
```diff
@@ -17,73 +17,325 @@ limitations under the License.
...
+		t.Fatalf("Setting kubeconfig flag: %v", err)
+	}
+	t.Cleanup(func() {
+		if err := kubeConfigFlag.Value.Set(oldKubeConfigPath); err != nil {
+			t.Fatalf("Restoring kubeconfig flag: %v", err)
+		}
+	})
+}
+
+func newTestKubeAPIServer(t *testing.T) *httptest.Server {
```
mimowo: Actually this introduces a lot of complication to the test. Are the requests to the API server sent from `ApplyTo`? 

In that case I would propose to refactor the prod code and split applyVisibilityServerOptions into two parts, the first part would be this one: https://github.com/kubernetes-sigs/kueue/blob/main/pkg/visibility/server.go#L103-L128

Then only write the unit tests for the first part which doesn't require running api server, wdyt?

pkg/visibility/server_test.go
mimowo: It may not be needed if we simplify the approach with the refactor: https://github.com/kubernetes-sigs/kueue/pull/11348/changes#r3273621990

weizhoublue: Yep — with the split above, the custom kubeconfig helper and fake API server are gone. Only `TestKubeconfigWrapper` + temp file remain for supplying  `kubeConfigPath` to the builder.

pkg/visibility/server_test.go
weizhoublue: Agreed

pkg/visibility/server_test.go
weizhoublue: Done

pkg/visibility/server_test.go
```diff
@@ -17,73 +17,181 @@ limitations under the License.
...
 
-	for name, tc := range tests {
-		t.Run(name, func(t *testing.T) {
-			options := genericoptions.NewSecureServingOptions().WithLoopback()
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			setVisibilityServerTestGlobals(t)
+			if tc.prepare != nil {
+				tc.prepare(t)
+			}
```
mimowo: this does not seem used by any tests so we can drop

weizhoublue: ok

mimowo (commented):
/lgtm
/approve
Thank you for the follow up, this looks much more future proof way for testing 👍

PR #11357, opened 2026-05-20
fix(site): replace deprecated Hugo language direction usage

### LINKED ISSUE #11354: Hugo deprecated warning: .Language.LanguageDirection was deprecated in v0.158.0
When building or running the documentation site, Hugo prints the following deprecation warning:

```text
WARN  deprecated: .Language.LanguageDirection was deprecated in Hugo v0.158.0 and will be
removed in a future release. Use .Language.Direction instead.
… +20 lines
```

The documentation site should no longer use the deprecated `.Language.LanguageDirection` field. It should use `.Language.Direction` instead, so future Hugo versions do not break the build when the deprecated field is removed.

yuluo-yx:
#### What type of PR is this?

/kind cleanup
/area website

Fixes #11354

```release-note
None
```


mimowo:
/ok-to-test
cc @mbobrovskyi

mbobrovskyi:
Oops, not this PR.

site/hugo.toml
```diff
@@ -17,8 +17,8 @@ enableMissingTranslationPlaceholders = true
 disableLanguages = []
 # disable taxonomies
 disableKinds = ["taxonomy"]
-# deprecated directories
-ignoreFiles = []
+# Ignore obsolete Docsy render hook templates removed by newer Hugo versions.
+ignoreFiles = ["td-render-heading\\.html$"]
```
mbobrovskyi: It doesn’t work. I still see the warning.

site/hugo.toml
```diff
@@ -53,24 +53,20 @@ ignoreFiles = []
 ###############################################################################
 # Docsy - Services configuration
 ###############################################################################
-[services]
-  [services.googleAnalytics]
-  id = "UA-135379910-1"
```
mbobrovskyi, line 58: Why this changes?

site/layouts/blog/baseof.html
mbobrovskyi, line 1: Do we have usage of blog?

site/layouts/swagger/baseof.html
mbobrovskyi, line 1: Do we have usage of swagger?

yuluo-yx: my bad, deleted

site/layouts/blog/baseof.html
yuluo-yx: my bad, deleted

site/hugo.toml
yuluo-yx: because the previous Universal Analytics property is retired and makes Hugo emit a GA4 migration warning. Re-enable it only after, configuring a new GA4 property and measurement ID.

site/hugo.toml
```diff
@@ -177,3 +174,72 @@ ignoreFiles = []
   [[module.imports]]
     path = "github.com/google/docsy"
     disable = false
+    [[module.imports.mounts]]
```
yuluo-yx, line 177: I added explicit module mounts for the Docsy import so that we only mount the Docsy layouts/assets/static/i18n paths we need, while avoiding layouts/_default/_markup/td-render-heading.html. This removes the unrecognized render hook template warning that was still reproducible locally.

That said, this may not be the best long-term solution. It makes the Docsy import more explicit and therefore a bit more brittle if Docsy changes its layout structure in the future. I applied it for now to unblock the warning, but I’m open to a cleaner approach if there is a preferred way to exclude this obsolete template.

mimowo:
/lgtm
/approve
/cherrypick website
Thiank you for making sure we don't use depreacated APIs.

/hold
In case @mbobrovskyi has more comments

site/hugo.toml
```diff
@@ -177,6 +174,75 @@ ignoreFiles = []
...
+      target = "layouts/_default/_markup/render-codeblock-math.html"
+    [[module.imports.mounts]]
+      source = "layouts/_default/_markup/render-codeblock-chem.html"
+      target = "layouts/_default/_markup/render-codeblock-chem.html"
+    [[module.imports.mounts]]
+      source = "layouts/_default/_markup/render-codeblock-mermaid.html"
+      target = "layouts/_default/_markup/render-codeblock-mermaid.html"
+    [[module.imports.mounts]]
+      source = "layouts/blog"
+      target = "layouts/blog"
```
mbobrovskyi: Probably blog and swagger we don't need, right?

yuluo-yx: yep, deleted

mbobrovskyi (commented):
/lgtm

Thank you!

PR #6790, opened 2025-09-11
Fix preemption in MultiKueue worker cluster

eric-higgins-ai:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

When using MultiKueue, it's possible for a workload to be preempted in the worker cluster but not the manager cluster. For example, this can happen if there is 5 CPU available in Cluster 1 and 5 in Cluster 2, and someone schedules a workload requesting 6 CPU. The manager cluster would see no issue with this, but one of the worker clusters would need to preempt to admit the new workload.

Currently in this case the manager cluster assumes it's lost connection to the workers. It waits for a timeout and then submits the workload to the worker clusters again.

This PR updates the behavior to instead sync the worker cluster's status back to the manager cluster, and then delete the workload in the worker cluster. This has the nice property that the preempted workload can be readmitted on any worker cluster (not just the one it was previously running on), potentially allowing it to restart sooner.

#### Special notes for your reviewer:
I didn't add unit tests yet because I'm not very confident in this approach. I just wanted to check if this is the right approach and then I'll add unit tests

#### Does this PR introduce a user-facing change?
```release-note
Reflect workload eviction statuses in the MultiKueue manager cluster.
```

PR #11376, opened 2026-05-21
KEP-11375: MultiKueue Cross-Cluster Preemption (alpha)

### LINKED ISSUE #11375: [MultiKueue] Cross-cluster preemption — reclaim cohort quota from sibling-cluster borrowers

A new MultiKueue dispatcher mode (alpha, default-off feature gate
`MultiKueueCrossClusterPreemption`) that, when an incoming workload at
the manager finds sibling clusters full of borrowed work, evicts a
borrower on a sibling worker so the incoming workload can be admitted
there. Mirrors single-cluster Kueue's `ReclaimWithinCohort` cross-cluster
via the External-dispatcher extension point.

The load-bearing semantic is **quota-based reclaim**: a sibling
ClusterQueue is only an eligible victim when it is currently borrowing
cohort capacity (`Status.FlavorsUsage > NominalQuota`). Reclaim is
gated by the victim's `Spec.Preemption.ReclaimWithinCohort`
(Never / LowerPriority / Any). Borrow-while-preempting is gated by the
preemptor's `Spec.Preemption.BorrowWithinCohort` plus
`MaxPriorityThreshold` (victim-side cap, matching the upstream API
godoc and single-cluster `isAboveBorrowingThreshold`). Cohort
membership is the shared `ClusterQueue.spec.cohortName` string;
no new CRDs.

Out of scope (potential follow-ons): cross-cluster FairSharing weights /
DRS, hierarchical cohorts via `Cohort.spec.parentName`, cross-cluster
aggregation of `LendingLimit` / `BorrowingLimit`. Worker-local
schedulers continue to honor each of these at admission time as today.

**Why is this needed**:

Operators that split workloads across per-tenant or per-environment
worker clusters need per-tenant quota guarantees that hold across
cluster boundaries. Today MultiKueue treats each worker as a black
box: a workload admits at the manager only if some worker has free
quota, and there is no mechanism to evict a sibling cluster's borrower
— even when single-cluster Kueue would do it locally via
`ReclaimWithinCohort`. A guaranteed-tier customer's workload can be
blocked indefinitely while a borrower on a sibling cluster finishes
naturally.

The closest related work, after auditing existing KEPs and issues:

- **KEP-8692 (#10766 / #10784)** — composable MultiKueue dispatcher.
  Explicitly lists *Cross-cluster preemption scheduling* under
  Non-Goals; this is the gap that work leaves open.
- **#6790** — addresses the worker → manager eviction propagation
  half. This proposal is the manager → worker initiation half. They
  coexist; neither subsumes the other.
- **KEP-8303** — independent feature gate; coordinates concurrent
  worker-side preemptions. Joint mode is supported (validated by
  integration test).
- **KEP-1337** — single-cluster `BorrowWithinCohort` +
  `MaxPriorityThreshold`. Reused as-is; same field, same semantics,
  applied across worker clusters.

**Completion requirements**:

- [ ] Design doc (KEP)
- [ ] API change: new dispatcher mode constant in
      `Configuration.MultiKueue.DispatcherName`; new feature gate
      `MultiKueueCrossClusterPreemption`.
- [ ] Docs update.

Working alpha prototype with 25 unit tests + 5 integration scenarios:
[`up/multikueue-cross-cluster-preemption`](https://github.com/ningziwen/kueue/tree/up/multikueue-cross-cluster-preemption).

/kind feature
/area multikueue
/area kep

ningziwen:
#### What type of PR is this?

/kind kep
/area multikueue
/area kep

Adds `keps/11375-multikueue-cross-cluster-preemption/` for a new
MultiKueue dispatcher mode (alpha, default-off feature gate
`MultiKueueCrossClusterPreemption`) that, when an incoming workload at
the manager finds sibling clusters full of *borrowed* work, evicts a
borrower on a sibling worker so the incoming workload can be admitted
there. Mirrors single-cluster Kueue's `ReclaimWithinCohort` cross-cluster
via the External-dispatcher extension point.

The load-bearing semantic is **quota-based reclaim** (`cqIsBorrowing`):
a sibling ClusterQueue is only an eligible victim source when it is
currently borrowing cohort capacity (`Status.FlavorsUsage > NominalQuota`).
Priority is one configurable value of the `ReclaimWithinCohort` policy
gate, not the primary driver. With `ReclaimWithinCohort=Any`, an owner
reclaims at *equal* priority — the ownership-of-nominal-quota use case
that motivates the KEP.

The KEP covers:

- `cqIsBorrowing` filter + `ReclaimWithinCohort` policy gate +
  `BorrowWithinCohort` borrow-while-preempting gate (with
  `MaxPriorityThreshold` as a victim-side cap, matching the upstream
  API godoc and single-cluster `isAboveBorrowingThreshold`).
- Priority + creation-time + name victim ordering across the cohort.
- Flat cohorts via the shared `ClusterQueue.spec.cohortName` string;
  the dispatcher does not read `Cohort` CRs.
- Joint operation with [KEP-8303](https://github.com/kubernetes-sigs/kueue/tree/main/keps/8303-multikueue-orchestrated-preemption)
  OrchestratedPreemption — both gates may be enabled simultaneously
  without coordination.

No new CRDs. No changes to existing CRDs other than a new dispatcher
mode constant.

Fixes #11375

This PR contains the **KEP only**. A working alpha prototype with
25 unit tests + 5 integration scenarios lives at
[`up/multikueue-cross-cluster-preemption`](https://github.com/ningziwen/kueue/tree/up/multikueue-cross-cluster-preemption);
the implementation PR will follow once the KEP is `implementable`.

Relationship to existing work (audited before opening):

- Builds on the umbrella in [#10766](https://github.com/kubernetes-sigs/kueue/issues/10766)
  / [KEP-8692](https://github.com/kubernetes-sigs/kueue/pull/10784)
  (composable MultiKueue dispatcher), which explicitly lists
  *Cross-cluster preemption scheduling* under **Non-Goals**. This KEP
  fills that gap as a separate dispatcher mode.
- Originating bug report: [#6790](https://github.com/kubernetes-sigs/kueue/pull/6790)
  ("Fix preemption in MultiKueue worker cluster"). That PR addresses
  the worker-to-manager eviction propagation half; this KEP is the
  manager-to-worker initiation half. Both can coexist.
- Independent of [KEP-8303](https://github.com/kubernetes-sigs/kueue/tree/main/keps/8303-multikueue-orchestrated-preemption)
  (separate feature gate; joint mode validated by integration test in
  the prototype).
- Reuses single-cluster [KEP-1337](https://github.com/kubernetes-sigs/kueue/tree/main/keps/1337-preempt-within-cohort-while-borrowing)
  (BorrowWithinCohort + MaxPriorityThreshold) API and semantics
  unchanged.

Out of scope for this KEP (potential follow-ons if a need surfaces):
[KEP-1714](https://github.com/kubernetes-sigs/kueue/tree/main/keps/1714-fair-sharing)
(FairSharing weights / DRS) cross-cluster, [KEP-79](https://github.com/kubernetes-sigs/kueue/tree/main/keps/79-hierarchical-cohorts)
(hierarchical cohorts) cross-cluster, and cross-cluster aggregation of
`LendingLimit` / `BorrowingLimit`. Worker-local schedulers continue to
honor each of these at admission time as today.

Open questions I'd specifically appreciate maintainer input on:

1. Forward-compat with [KEP-8692](https://github.com/kubernetes-sigs/kueue/pull/10784):
   if the composable dispatcher framework lands first, would
   maintainers prefer the cross-cluster preemption logic restructured
   as plugins within KEP-8692's pipeline at beta, or kept as a sibling
   dispatcher mode?
2. Joint operation with [KEP-8303](https://github.com/kubernetes-sigs/kueue/tree/main/keps/8303-multikueue-orchestrated-preemption):
   alpha keeps the two paths independent (the prototype has an
   integration test asserting no `MultiKueuePreemptionGate` activates
   on workloads our dispatcher touches). Is independent operation OK
   long-term, or should beta unify them?

Happy to discuss in `#sig-scheduling` / `#wg-batch` or at a Kueue
community meeting in parallel with PR review.

```release-note
NONE
```

### DIFF
--- a/keps/11375-multikueue-cross-cluster-preemption/README.md
+++ b/keps/11375-multikueue-cross-cluster-preemption/README.md
@@ -12,9 +12,14 @@
   - [Notes/Constraints/Caveats](#notesconstraintscaveats)
   - [Risks and Mitigations](#risks-and-mitigations)
 - [Design Details](#design-details)
-  - [Cohort membership: the shared `cohortName` string](#cohort-membership-the-shared-cohortname-string)
-  - [Cohort-level properties via the single-cluster `Cohort` CR](#cohort-level-properties-via-the-single-cluster-cohort-cr)
+  - [Cohort membership: the shared <code>cohortName</code> string](#cohort-membership-the-shared-cohortname-string)
+  - [Cohort-level properties via the single-cluster <code>Cohort</code> CR](#cohort-level-properties-via-the-single-cluster-cohort-cr)
   - [Worked example: per-tenant cohort across worker clusters](#worked-example-per-tenant-cohort-across-worker-clusters)
+    - [Topology](#topology)
+    - [Activation (prerequisite)](#activation-prerequisite)
+    - [Setup YAML](#setup-yaml)
+    - [What you observe (step-by-step)](#what-you-observe-step-by-step)
+    - [Why each knob matters](#why-each-knob-matters)
   - [Dispatcher mode](#dispatcher-mode)
   - [Reconcile algorithm](#reconcile-algorithm)
   - [Per-CQ preemption policy](#per-cq-preemption-policy)
@@ -37,10 +42,19 @@
 Today MultiKueue dispatches workloads across worker clusters but treats each
 worker as a black box for scheduling decisions: a workload is admitted only
 if a worker has free quota. There is no way for a workload arriving at the
-manager to reclaim its cohort's quota from a borrower running on a sibling
-worker — even when single-cluster Kueue would already do exactly that
-locally via `ReclaimWithinCohort`. Owners wait indefinitely while
-borrowers finish naturally.
+manager to reclaim its cohort's quota from a cohort-sibling tenant's
+workload that is borrowing it on a worker — even when single-cluster Kueue
+would already do exactly that locally via `ReclaimWithinCohort`. Owners
+wait indefinitely while those borrowing workloads finish naturally.
+
+Note on terminology: "borrowing" here is single-cluster Kueue's local
+cohort math, applied independently inside each worker cluster. A tenant's
+ClusterQueue on a worker is borrowing whenever its
+`Status.FlavorsUsage > NominalQuota`, which means it is using cohort
+capacity that another tenant on that same worker is not currently
+consuming. Worker clusters do not borrow from each other; the "cross-
+cluster" part is the manager's fleet-wide visibility and its ability to
+target the eviction at the right worker.
 
 This KEP introduces a cross-cluster preemption capability for MultiKueue:
 a new dispatcher mode (`kueue.x-k8s.io/multikueue-dispatcher-cross-cluster-preemption`).
@@ -66,12 +80,12 @@ Operators that split workloads across per-tenant or per-environment worker
 clusters need a way to enforce per-tenant quota guarantees across cluster
 boundaries. Without cross-cluster preemption:
 
-- A guaranteed-tier customer's workload can be blocked indefinitely when a
-  sibling cluster's borrower is occupying the customer's nominal quota,
-  because the manager has no mechanism to evict that borrower across the
-  cluster boundary.
-- Wait time is bounded only by the borrower's natural completion time,
-  which may be hours or longer.
+- A guaranteed-tier customer's workload can be blocked indefinitely when
+  a cohort-sibling tenant's workload is borrowing the customer's nominal
+  quota on a worker, because the manager has no mechanism to reach into
+  that worker and evict the borrowing workload.
+- Wait time is bounded only by how long that borrowing workload runs
+  naturally, which may be hours.
 
 Priority-driven preemption between tenants is a *secondary* use case
 this KEP enables (via `ReclaimWithinCohort=LowerPriority`), but the
@@ -139,10 +153,10 @@ A platform offers guaranteed-tier customers a bounded wait time as long as
 they remain within their per-tenant quota. Workers are split per-tenant for
 security isolation, but workers share an on-demand capacity reservation
 behind the scenes. When a guaranteed customer submits a workload within
-its quota and finds a sibling cluster's borrower occupying that quota,
-the borrower should be reclaimed within seconds, not hours — even when
-the borrower is at *equal* priority (the relevant signal is ownership of
-the nominal quota, not priority delta).
+its quota and finds that a cohort-sibling tenant's workload is borrowing
+that quota on a worker, the borrowing workload should be reclaimed within
+seconds, not hours — even at *equal* priority (the relevant signal is
+ownership of the nominal quota, not priority delta).
 
 #### Story 2 — Burst absorption across environments
 
--- a/keps/11375-multikueue-cross-cluster-preemption/README.md
+++ b/keps/11375-multikueue-cross-cluster-preemption/README.md
@@ -1,0 +1,821 @@
+# KEP-11375: MultiKueue Cross-Cluster Preemption
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1 — Guaranteed-tier customer SLA](#story-1--guaranteed-tier-customer-sla)
+    - [Story 2 — Burst absorption across environments](#story-2--burst-absorption-across-environments)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Cohort membership: the shared `cohortName` string](#cohort-membership-the-shared-cohortname-string)
+  - [Cohort-level properties via the single-cluster `Cohort` CR](#cohort-level-properties-via-the-single-cluster-cohort-cr)
+  - [Worked example: per-tenant cohort across worker clusters](#worked-example-per-tenant-cohort-across-worker-clusters)
+  - [Dispatcher mode](#dispatcher-mode)
+  - [Reconcile algorithm](#reconcile-algorithm)
+  - [Per-CQ preemption policy](#per-cq-preemption-policy)
+  - [Single-writer lock: victim-claim annotation](#single-writer-lock-victim-claim-annotation)
+  - [Eviction primitive](#eviction-primitive)
+  - [Interaction with KEP-8303 (Orchestrated Preemption)](#interaction-with-kep-8303-orchestrated-preemption)
+  - [Test Plan](#test-plan)
+    - [Prerequisite testing updates](#prerequisite-testing-updates)
+    - [Unit tests](#unit-tests)
+    - [Integration tests](#integration-tests)
+    - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+Today MultiKueue dispatches workloads across worker clusters but treats each
+worker as a black box for scheduling decisions: a workload is admitted only
+if a worker has free quota. There is no way for a workload arriving at the
+manager to reclaim its cohort's quota from a borrower running on a sibling
+worker — even when single-cluster Kueue would already do exactly that
+locally via `ReclaimWithinCohort`. Owners wait indefinitely while
+borrowers finish naturally.
+
+This KEP introduces a cross-cluster preemption capability for MultiKueue:
+a new dispatcher mode (`kueue.x-k8s.io/multikueue-dispatcher-cross-cluster-preemption`).
+The dispatcher's load-bearing semantic is **quota-based reclaim**: a
+sibling CQ is only an eligible victim when it is currently borrowing
+cohort capacity (`Status.FlavorsUsage > NominalQuota`). Reclaim is then
+gated by the victim CQ's `Spec.Preemption.ReclaimWithinCohort` policy
+(`Never` / `LowerPriority` / `Any`) — so priority is one configurable
+gate, not the primary driver. With `Any`, an owner reclaims at *equal*
+priority; with `LowerPriority`, a strict priority delta is required.
+**No new CRDs are introduced**: cross-cluster cohort identity is the shared
+`ClusterQueue.spec.cohortName` string across the manager and the workers,
+exactly the same mechanism single-cluster Kueue uses. Cohort-level
+properties (parent, additional shared quota, fair sharing) live on the
+standard single-cluster `Cohort` CR on the manager (and on each worker
+that needs them), and the dispatcher inherits the result via each remote
+ClusterQueue's `Status.FlavorsUsage` snapshot. Activation is gated by
+alpha feature gate `MultiKueueCrossClusterPreemption` (default off).
+
+## Motivation
+
+Operators that split workloads across per-tenant or per-environment worker
+clusters need a way to enforce per-tenant quota guarantees across cluster
+boundaries. Without cross-cluster preemption:
+
+- A guaranteed-tier customer's workload can be blocked indefinitely when a
+  sibling cluster's borrower is occupying the customer's nominal quota,
+  because the manager has no mechanism to evict that borrower across the
+  cluster boundary.
+- Wait time is bounded only by the borrower's natural completion time,
+  which may be hours or longer.
+
+Priority-driven preemption between tenants is a *secondary* use case
+this KEP enables (via `ReclaimWithinCohort=LowerPriority`), but the
+primary motivation is honoring per-tenant nominal quota the way
+single-cluster Kueue already does within a cluster.
+
+KEP-693 (the original MultiKueue KEP) explicitly listed "synchronize
+configuration across the clusters" as a Non-Goal. KEP-8303 (Orchestrated
+Preemption) addresses a different problem: avoiding redundant simultaneous
+preemptions across replicas of *one* workload dispatched via AllAtOnce.
+Neither covers cross-cluster victim selection.
+
+### Goals
+
+- Provide cross-cluster preemption with **the same API surface as
+  single-cluster Cohort** so users have one mental model.
+- Honor `ClusterQueue.spec.cohortName` for membership, the full per-CQ
+  preemption policy (`ReclaimWithinCohort` / `BorrowWithinCohort`), and
+  the per-CQ borrowing state (`Status.FlavorsUsage > NominalQuota`)
+  exactly as in single-cluster: a victim is preempted iff its home CQ
+  is currently borrowing AND the priority/policy gate allows it.
+- Stay additive: do not modify the in-tree scheduler or change behavior
+  when the feature gate is off.
+- Reuse existing primitives (admission checks, the External dispatcher
+  extension point, `workload.Evict`, the multikueue origin label) instead of
+  introducing parallel infrastructure.
+
+### Non-Goals
+
+- No new CRD. Cross-cluster cohort identity is the shared
+  `ClusterQueue.spec.cohortName` string, exactly the mechanism
+  single-cluster Kueue uses.
+- The cross-cluster dispatcher does not read `Cohort` CRs. Cohort-level
+  shared quota, fair-sharing weights, parent chains, and lending/
+  borrowing limits remain enforced by each cluster's local scheduler at
+  admission time. The dispatcher only consumes the resulting
+  `ClusterQueue.Status.FlavorsUsage` snapshot.
+- The dispatcher is inherently cross-CQ. Same-CQ preemption stays in the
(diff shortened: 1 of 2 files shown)


mimowo:
/ok-to-test
cc @olekzabl @yaroslava-serdiuk

yaroslava-serdiuk:
>when an incoming workload at the manager finds sibling clusters full of borrowed work

Could you please clarify what do you mean here? I don't understand how the working clusters can borrow from each other.

ningziwen:
@yaroslava-serdiuk Revised to clarify.

olekzabl:
@ningziwen 
First of all, let me praise the depth of your work!
Besides the KEP, I could find a draft implementation, with test coverage, and rich comments. This has been helpful!

I started reviewing today and TBH I'm still not yet done.
However, let me try to start the key topic right away:

### Key topic

I claim that MultiKueue **already has** functionality **like you need**. 
(Though, admittedly, I'm not sure if it's *exactly* what you need. We'll need to discuss).

### Evidence

To kick off the discussion, please look at [this branch](https://github.com/olekzabl/kueue/blob/br44/test/integration/multikueue/scheduler/cross_cluster_preemption_test.go). 
It's my experimenting on top of your implementation. (Specifically, my changes as of now are [these](https://github.com/olekzabl/kueue/commit/c5cee81f405907ae7a37296269ae4911b246df8c)).
_(disclaimer: AI helped me in this)_
I *essentially* made your integration tests pass **without cross-cluster preemption**, by changing the following bits:

1. Use `AllAtOnce` instead of your new dispatcher.
2. Use `OrchestratedPreemption` to make sure reclaiming in cohorts won't cause more evictions than necessary.
3. **Flip** `ReclaimWithinCohort` settings between the 2 tenants A & B.
    The meaning of this field is "how you can reclaim"; **not** "how you can be reclaimed from".
    Cf. [Kueue docs](https://kueue.sigs.k8s.io/docs/concepts/cluster_queue/#preemption):

    > `reclaimWithinCohort` determines whether a pending Workload *[in this ClusterQueue]* can preempt Workloads from **other** ClusterQueues in the cohort that are using more than their nominal quota.

4. Finally, some technical glue to keep the tests meaningful:

   * change the way of verifying that a workload was evicted
      (in the current behavior, the manager-side workload does not become `Evicted` but rather blocked-again on the MultiKueue AdmissionCheck transitioned to `Retry` and then back to `Pending`)
      * OTOH, to make things more convincing, I added checks for `Evicted` condition on the workers
   * remove Scenario 4 (which now became a duplicate of Scenario 1)
   * remove Scenario 3 (which was very specific to the cross-cluster preemption mechanics)

### Conclusions

At this point, I'm **not sure** how much of your work (KEP, impl) remains relevant for your needs.

Let me share my early thoughts on this:

1. `AllAtOnce` + `OrchestratedPreemption` is admittedly not perfect.
   It's quite "noisy". Will dispatch the candidate to _every worker_, and then just make sure not to _evict_ in many places.
   I'm assuming your new dispatcher can do better, in making a smart worker selection and leaving the other workers undisturbed.
   That feels valuable in general.
   How valuable is it from your perspective? Maybe that gain alone is a good reason to introduce a new dispatcher?

2. If we choose to continue this KEP, I'm blindly hoping that my findings above will allow **simplifying** it.
   For example, I imagine the dispatcher could just nominate the best worker - however, it should not need to deal with evicting workloads.
   Here, the existing logic behind `ReclaimWithinCohort` should do.
   (Of course, unless there are some bugs which I haven't yet realized).

3. Given my experiments above, I'm feeling that your background ("Summary" and "Motivation" in this KEP, also the phrasing of #11375) are a bit misleading.
   It's not that MultiKueue does not support quota reclaiming on workers at all. It can :) Possibly a bit imperfectly.
   Once we align here, let's comment-or-update on #11375 to make sure it's not misleading for future readers.
   Same will apply to this KEP if we choose to continue it.

---

Please let me know your thoughts!

keps/11375-multikueue-cross-cluster-preemption/README.md
```diff
@@ -0,0 +1,835 @@
...
+The dispatcher targets the **manager-side** Workload (not the remote)
+when evicting. The manager client is the single source of truth for the
+workload's admission state, and MultiKueue already has a clean cleanup
+path for "manager workload no longer has quota reservation".
+
+The evictor calls `workload.PatchAdmissionStatus` against the manager
+client with a function that:
+
+- Calls `workload.SetEvictedCondition` with reason
+  `WorkloadEvictedByPreemption` and a message identifying the incoming
```
ArangoGutierrez, line 618: Single-cluster preemption writes a fixed message onto the victim: `Preempted to accommodate a workload (UID: ..., JobUID: ...) due to ...; preemptor path: ...; preemptee path: ...` (pkg/scheduler/preemption/preemption.go:179). Our kueue-who-preempted runbook parses exactly that out of the `Evicted` condition. Here the primitive only says the message "identifies the incoming workload and the freed cluster." Since eviction targets the manager-side victim, the incoming (preemptor) manager workload still carries its UID and JobUID label, so the same schema is reachable. Can the KEP commit to reusing that message format? Also, single-cluster sets a separate `Preempted` condition via `SetPreemptedCondition` (preemption.go:235) in addition to `Evicted`. Worth stating whether the cross-cluster path sets it too, so downstream condition consumers behave the same.

keps/11375-multikueue-cross-cluster-preemption/README.md
```diff
@@ -0,0 +1,835 @@
...
+     `victim.Priority < incoming.Priority` AND (when
+     `MaxPriorityThreshold` is set) `victim.Priority <= MaxPriorityThreshold`.
+     The threshold is a victim-side cap: victims above the threshold are
+     protected even if the incoming workload outranks them. Mirrors the
+     upstream API godoc on `BorrowWithinCohort.MaxPriorityThreshold` and
+     single-cluster `isAboveBorrowingThreshold`.
+9. Filter out victims with an existing `victim-of` annotation belonging to
+   a different incoming workload.
+10. Sort candidates: priority ascending, creation timestamp ascending,
+    name. Pick the lowest.
```
ArangoGutierrez, line 557: Picking a single lowest-priority victim by (priority, creation, name) is not how single-cluster selects victims when FairSharing is enabled. That path orders targets by dominant resource share (pkg/scheduler/preemption/fairsharing/ordering.go), not raw priority. If a worker CQ has FairSharing on locally, the manager can evict a victim the worker's own scheduler would have ranked differently. Can the KEP state what happens when a member CQ runs FairSharing, even if cross-cluster DRS is out of scope? At minimum I would call out the divergence explicitly so operators are not surprised.

ArangoGutierrez (commented):
Read this with a GPU fleet operator hat on, since cross-cluster eviction of accelerator workloads is the kind of thing that burns real money when victim selection is even slightly off. The core design holds up: reuse of `cohortName`, `ReclaimWithinCohort` / `BorrowWithinCohort`, the External dispatcher seam, and manager-side eviction so MultiKueue's existing "delete remote when no quota reservation" path does the cleanup. Dropping the separate `MultiKueueCohort` CRD is the right call and the Alternatives section argues it well. The `BorrowWithinCohort.MaxPriorityThreshold` description also matches `isAboveBorrowingThreshold` in the single-cluster code, so the borrow-preemption semantics are faithful.

Three things I want the KEP to answer rather than the implementation PR:

1. Observability parity. Single-cluster preemption stamps a specific message on the victim (preemptor UID, JobUID, cohort paths) that our own who-preempted runbook parses out of the `Evicted` condition. The eviction primitive here only says the message identifies the incoming workload and the freed cluster. If the schema drifts, cross-cluster preemption becomes the one case operators cannot trace with existing tooling. Can we commit to the same message schema and set the `Preempted` condition too?

2. Victim selection vs FairSharing and TAS. The dispatcher picks a single lowest-priority victim by priority, then creation time, then name. Single-cluster selection is not that simple once FairSharing is on (DRS ordering) or TAS is involved (it confirms a topology assignment exists before committing the preemption). FairSharing cross-cluster is out of scope, understood, but a worker CQ can run FairSharing or TAS locally today. The manager's naive pick can evict a victim the worker itself would not have chosen, or free quota the worker still cannot place because of topology fragmentation. On expensive nodes that is a wasted eviction. What is the story there?

3. Convergence of the `victim-of` claim. The claim annotation lives on the remote workload, but eviction targets the manager workload. If the claiming workload disappears between claim and evict, what clears the stale annotation? As written, a leaked claim looks like it permanently protects that victim from future reclaim.

Details inline. None of these look fatal to the approach.

keps/11375-multikueue-cross-cluster-preemption/README.md
```diff
@@ -0,0 +1,835 @@
...
+11. ClaimAndEvict the chosen victim. On `ErrVictimAlreadyClaimed`, retry
+    with the next candidate.
+12. On successful eviction, write `wl.Status.NominatedClusterNames` to the
```
ArangoGutierrez, line 560: Nominating only the freed cluster assumes freeing quota there makes the incoming workload placeable there. With TAS that is not guaranteed. Single-cluster preemption calls `FindTopologyAssignmentsForWorkload` (pkg/scheduler/preemption/preemption.go:633) to confirm the freed capacity yields a valid topology assignment before it commits the eviction. This dispatcher frees quota, then nominates, with no topology feasibility check. If the worker cannot place the incoming workload due to topology fragmentation, we have evicted an expensive victim for nothing and the incoming workload bounces back to pending. For a GPU fleet that is the exact failure operators will complain about. Should the dispatcher verify placement feasibility on the freed worker before evicting, or at least document this as a known alpha limitation with a mitigation path?

keps/11375-multikueue-cross-cluster-preemption/README.md
```diff
@@ -0,0 +1,835 @@
...
+  The threshold is a victim-side cap (matches the upstream API godoc on
+  `BorrowWithinCohort.MaxPriorityThreshold` and single-cluster
+  `isAboveBorrowingThreshold`); victims above the threshold are
+  protected from borrow-preemption even when the incoming workload
+  outranks them on `Spec.Priority`.
+
+`Spec.Preemption.WithinClusterQueue` is honored locally by each worker's own
+Kueue scheduler; cross-cluster preemption does not change its semantics.
+
+### Single-writer lock: victim-claim annotation
```
ArangoGutierrez, line 595: The single-writer lock claims the victim on the remote workload, but eviction happens on the manager workload. Walk the abort case: dispatcher patches `victim-of` on the remote, then the manager crashes or the incoming workload is deleted before the manager-side eviction lands. The remote stays admitted with a stale `victim-of` pointing at a workload that no longer exists. Step 9 of the reconcile then treats that victim as claimed-by-another and skips it, so the victim is protected forever. What clears a stale claim? A TTL, an owner-exists check at reconcile, or clearing on incoming-workload deletion would all work, but the KEP needs to name one.

keps/11375-multikueue-cross-cluster-preemption/README.md
```diff
@@ -0,0 +1,835 @@
...
+   and the priority filter when policy is `LowerPriority`.
+4. **Same-CQ skip**: skip cohort members whose CQ name equals the
+   incoming workload's home CQ name. Cross-cluster preemption is for
+   cross-CQ reclaim within a cohort; within-CQ preemption (e.g. a
+   higher-priority workload in tenant-a-cq displacing another tenant-a-cq
+   workload on a sibling worker) is the worker's local
+   `WithinClusterQueue` policy concern, not this dispatcher's.
+5. For each remaining member, check `Spec.Preemption.ReclaimWithinCohort != Never`.
+   If `Never`, skip — that CQ has opted out of being a victim source.
+6. **Quota-based reclaim filter**: skip members where
```
ArangoGutierrez, line 533: This step says the filter mirrors single-cluster cqIsBorrowing gating, but the borrowing state the doc defines earlier (README:107) is the raw `Status.FlavorsUsage > NominalQuota` comparison. Single-cluster cqIsBorrowing runs `cq.Borrowing(fr)` on a cohort-aware snapshot (pkg/scheduler/preemption/preemption.go:609), which accounts for lending and borrowing limits and the cohort tree, and the KEP keeps LendingLimit and BorrowingLimit local, not aggregated. A CQ can read as borrowing by the raw check while its LendingLimit says the owner is not entitled to reclaim that slice. Can the KEP confirm the raw comparison is intended, and note where it can disagree with the worker's own borrowing math?

keps/11375-multikueue-cross-cluster-preemption/README.md
```diff
@@ -0,0 +1,835 @@
...
+### Dispatcher mode
+
+A new value for `Configuration.MultiKueue.DispatcherName`:
+
+```
+kueue.x-k8s.io/multikueue-dispatcher-cross-cluster-preemption
+```
+
+When set, the new dispatcher reconciler is registered alongside the
+existing built-ins. It activates only when the
```
ArangoGutierrez, line 506: "Activates only when the feature gate is enabled" leaves the gate-off path underspecified. Today nothing rejects an unknown DispatcherName at config load. The only validation touching it is the incrementalDispatcherConfig check (pkg/config/validation.go:214), and any name that is not AllAtOnce or Incremental falls through to the external dispatcher branch (pkg/controller/admissionchecks/multikueue/workload.go:868), which waits for something else to populate Status.NominatedClusterNames. So if an operator sets this new mode while the gate is off, workloads sit with no nominations and no error anywhere. Please spell out the gate-off behavior: reject at config load, or fall back to AllAtOnce. This is the feature-gate-interaction case that tends to ship broken.

keps/11375-multikueue-cross-cluster-preemption/README.md
```diff
@@ -0,0 +1,835 @@
...
+- `crosspreemption_quota_test.go` — 13 table-driven cases pinning the
+  upstream semantics of `canBorrowAgainstVictim`. Cover the nil-spec
+  and `Policy=Never` short-circuits, the standard `LowerPriority`
+  check, and every `MaxPriorityThreshold` case using Apple Ray's
+  priority tiers (`p0=-100`, `p1=-500`, `p2=-1000`) — including the
+  boundary case where the victim is at the threshold (eligible) and
+  the protection case where the victim is above the threshold even
+  though it is below the incoming priority.
+
+#### Integration tests
```
ArangoGutierrez, line 721: The integration scenarios cover happy path, priority gate, already-claimed, joint mode with KEP-8303, and the threshold cap. The highest-risk path is missing: manager-worker partition mid-eviction, and recovery from a stale/leaked `victim-of` claim. Those are the failure modes the risks section leans on idempotency for, so they deserve a test. And with no automated e2e (README:762, manual kind only), a feature that evicts workloads across clusters is landing with no CI coverage of the cross-cluster teardown-and-redispatch flow. Can we get at least one e2e, or a written justification for why the integration harness is sufficient?

keps/11375-multikueue-cross-cluster-preemption/README.md
```diff
@@ -0,0 +1,835 @@
...
+No automated e2e tests beyond the integration suite. The feature has
+been manually validated end-to-end on a kind-based harness with 1
+manager + 2 worker clusters running the modified Kueue manager binary;
+the prototype is referenced from the
+[Implementation History](#implementation-history) section.
+
+### Graduation Criteria
+
+**Alpha → Beta:**
+- Field experience from at least one production deployment.
```
ArangoGutierrez, line 770: "Field experience from at least one production deployment" is a heavy alpha-to-beta gate for a default-off cross-cluster feature. It may be hard to source and hard to verify. Consider pairing it with concrete, checkable criteria: scale numbers for the per-reconcile cohort listing, and metrics on the new path. Related: the KEP adds a whole eviction path but defines no metrics. Single-cluster has `ReportPreemption` (pkg/scheduler/preemption/preemption.go:259). A cross-cluster preemptions-issued / victims-evicted counter would let operators see the feature working without reading logs.

keps/11375-multikueue-cross-cluster-preemption/kep.yaml
```diff
@@ -0,0 +1,32 @@
+title: MultiKueue Cross-Cluster Preemption
+kep-number: 11375
+authors:
+  - "@ziwen-ning"
```
ArangoGutierrez, line 4: nit: `@ziwen-ning` 404s on GitHub. Your login is `ningziwen`, so I think the handle in authors just needs fixing.

keps/11375-multikueue-cross-cluster-preemption/README.md
```diff
@@ -0,0 +1,835 @@
...
+  outranks them on `Spec.Priority`.
+
+`Spec.Preemption.WithinClusterQueue` is honored locally by each worker's own
+Kueue scheduler; cross-cluster preemption does not change its semantics.
+
+### Single-writer lock: victim-claim annotation
+
+Concurrent reconciles for two different incoming workloads must not pick the
+same victim. The dispatcher claims a victim by patching annotation
+`kueue.x-k8s.io/multikueue-cross-preemption-victim-of` =
```
ArangoGutierrez, line 599: nit: the claim annotation is named three ways in the doc: `victim-of` (README:183), the full key `...multikueue-cross-preemption-victim-of` here, and `MultiKueueCrossClusterPreemptionVictimAnnotation` (README:747). It is the load-bearing lock key, so please pick one canonical name and reference it consistently.

PR #11388, opened 2026-05-21
feat: increase Workload PodSet limit to 10

### LINKED ISSUE #11379: Increase the number of PodSets per Job

Currently we have a limit of only 8 PodSets per workload which I would like to increase to 16.

**Why is this needed**:

* some integrations consume multiple-PodSets just for their operation, for example Ray has 1 head, 1 submitter job, 1 redis cleanup job, this leaves only 5 PodSets for workers, see discussion here: https://github.com/kubernetes-sigs/kueue/pull/11260/#issuecomment-4487814710
* some users prefer to split a big JobSet into multiple smaller ReplicatedJobs with different PodSets so that they are admitted to different flavors independently, increasing the granularity of admission

yuluo-yx:
#### What type of PR is this?

/kind feature

Fixes #11379

```release-note
Increased the maximum number of PodSets per Workload from 8 to 10. As a result, RayCluster, RayJob, and RayService integrations now allow up to 9 worker groups, since one PodSet is reserved for the Ray head group.
```


tenzen-y:
/hold

I'd like to figure out if increasing the number of PodSets degrades TAS performance.
Because we can not revert this change afterward.

yuluo-yx:
> /hold
> 
> I'd like to figure out if increasing the number of PodSets degrades TAS performance. Because we can not revert this change afterward.

Agree. 👀

mimowo:
> I'd like to figure out if increasing the number of PodSets degrades TAS performance.
> Because we can not revert this change afterward.

The performance of TAS (or general scheduler) does not degrade unless a user requests to have more PodSets. Also, users can scale up Kueue (add more CPU/mem to the kueue deployment), but they cannot workaround validation except for forking Kueue.

Maybe to mitigate the issue we can start with some smaller number like 10 or 12, at the same time to mitigate the additional "technical" PodSets for Ray: https://github.com/kubernetes-sigs/kueue/pull/11260 - we want to introduce there a new PodSet for RedisCleanup job and that would mean one less PodSet for "Workers".

wdyt?

tenzen-y:
> > I'd like to figure out if increasing the number of PodSets degrades TAS performance.
> > Because we can not revert this change afterward.
> 
> The performance of TAS (or general scheduler) does not degrade unless a user requests to have more PodSets. Also, users can scale up Kueue (add more CPU/mem to the kueue deployment), but they cannot workaround validation except for forking Kueue.
> 
> Maybe to mitigate the issue we can start with some smaller number like 10 or 12, at the same time to mitigate the additional "technical" PodSets for Ray: #11260 - we want to introduce there a new PodSet for RedisCleanup job and that would mean one less PodSet for "Workers".
> 
> wdyt?

The computation costs are N × M in flavor assignment at worst, where N is the number of flavors and M is the number of PodSets. And in the TAS calculation pays more computation costs because it considers the tree topology structure.

But, I am assuming that 10 has less impact. 
Surely, it will degrade TAS performance by increasing the number of pattern matches, but 10 sounds like a reasonable balance between performance and user capabilities.

mimowo:
@yuluo-yx could you adjust the PR to 10?

yuluo-yx:
> @yuluo-yx could you adjust the PR to 10?

got it, I'll update tonight.

mimowo:
/retitle feat: increase Workload PodSet limit to 10
As per discussion

mimowo:
/unhold
as per https://github.com/kubernetes-sigs/kueue/pull/11388#issuecomment-4510224328

mimowo:
/lgtm
/approve
Thank you 👍

tenzen-y:
/release-note-edit
```release-note
Increased the maximum number of PodSets per Workload from 8 to 10. As a result, RayCluster, RayJob, and RayService integrations now allow up to 9 worker groups, since one PodSet is reserved for the Ray head group.
```

PR #11399, opened 2026-05-21
Include effective requests in scheduling equivalence hash

sohankunkerkar:
#### What type of PR is this?
/kind cleanup

Fixes https://github.com/kubernetes-sigs/kueue/pull/10973#discussion_r3282460697

```release-note
NONE
```

### DIFF
--- a/pkg/workload/workload.go
+++ b/pkg/workload/workload.go
@@ -329,13 +329,16 @@ func computeSchedulingHash(log logr.Logger, wl *kueue.Workload, totalRequests []
 	podSetShapes := make([]map[string]any, 0, len(wl.Spec.PodSets))
 	for i, ps := range wl.Spec.PodSets {
 		effectiveCount := ps.Count
+		var effectiveRequests resources.Requests
 		if i < len(totalRequests) {
 			effectiveCount = totalRequests[i].Count
+			effectiveRequests = totalRequests[i].Requests
 		}
 		podSetShapes = append(podSetShapes, map[string]any{
 			"name":            ps.Name,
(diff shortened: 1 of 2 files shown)


pkg/workload/workload.go
```diff
@@ -329,13 +329,16 @@ func computeSchedulingHash(log logr.Logger, wl *kueue.Workload, totalRequests []
...
 	for i, ps := range wl.Spec.PodSets {
 		effectiveCount := ps.Count
+		var effectiveRequests resources.Requests
 		if i < len(totalRequests) {
 			effectiveCount = totalRequests[i].Count
+			effectiveRequests = totalRequests[i].Requests
 		}
 		podSetShapes = append(podSetShapes, map[string]any{
 			"name":            ps.Name,
 			"spec":            utilpod.SpecShape(&ps.Template.Spec),
```
mwielgus, line 339: Why is SpecShape (which includes container requests via ContainerShape) is not enough?

sohankunkerkar: Good question! You are right about `SpecShape`, but DRA preprocessing changes `TotalRequests` without modifying the spec as  it replaces `example.com/gpu` with the DeviceClass-mapped logical resource (gpu). After translation, the spec still hashes the same but the workload's actual resource demands changed, so including `effectiveRequests` makes the hash reflect what the scheduler will actually evaluate.

mimowo:
/lgtm
/approve
Thank you 👍 I think this adds virtually no perf hit compared to what we already do, and if this helps to fix some DRA workloads, then great.

PR #11400, opened 2026-05-21
[MultiKueue] add RayService permissions to MultiKueue worker ClusterRole

kevin85421:
#### What type of PR is this?
/kind documentation
/area multikueue

The MultiKueue worker ClusterRole generated by create-multikueue-kubeconfig.sh was missing rayservices/* rules even though Kueue supports RayService as a MultiKueue-managed integration (since v0.17.0) (https://github.com/kubernetes-sigs/kueue/blame/v0.17.0/charts/kueue/values.yaml#L155). Without these permissions, RayService workloads dispatched to worker clusters fail at the API level.

Add rayservices and rayservices/status rules alongside the existing rayjobs and rayclusters rules. The script is embedded into the MultiKueue setup docs via Hugo include, so the rendered doc picks up the change automatically.

```release-note
NONE
```

### DIFF
--- a/site/static/examples/multikueue/create-multikueue-kubeconfig.sh
+++ b/site/static/examples/multikueue/create-multikueue-kubeconfig.sh
@@ -240,6 +240,22 @@ rules:
   verbs:
   - get
 - apiGroups:
+  - ray.io
+  resources:
+  - rayservices
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ray.io
+  resources:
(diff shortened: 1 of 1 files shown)


mimowo (commented):
/lgtm
/approve
Thank you for improving the documentation about the KubeRay 👍 
/cherrypick website

mszadkow:
We didn't spot this in the test: "Should run a RayService on worker if admitted", because we do it programmatically - `test/util/multikueue.go`
/lgtm

PR #11420, opened 2026-05-21
E2E tests for Partitionable Devices

PannagaRao:
#### What type of PR is this?
/kind feature

Fixes #

This PR has a dependency on https://github.com/kubernetes-sigs/kueue/pull/12097

```release-note
None
```

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end test suite for DRA partitionable GPU devices.
  * Validates GPU memory quota management, multi-partition configurations, and resource allocation scenarios.

* **Chores**
  * Enhanced testing infrastructure with extended DRA test configurations and automation support.


test/e2e/dra/counter/dra_test.go
```diff
@@ -0,0 +1,350 @@
...
+)
+
+func logWorkloadConditions(wl *kueue.Workload) {
+	for _, c := range wl.Status.Conditions {
+		ginkgo.GinkgoLogr.Info(fmt.Sprintf("Workload %s condition: type=%s status=%s reason=%s message=%s",
+			wl.Name, c.Type, c.Status, c.Reason, c.Message))
+	}
+}
+
+var _ = ginkgo.Describe("DRA Partitionable Devices", func() {
```
sohankunkerkar, line 53: can you also add a test for multiple workloads sharing counter quota?

sohankunkerkar: Also, we need to cover the config removed at runtime case.

PannagaRao: Added a test for multiple workloads.

hack/testing/e2e-common.sh
```diff
@@ -1249,19 +1254,26 @@ function install_dra_example_driver {
         && mv "$dra_driver_temp_dir/Makefile.tmp" "$dra_driver_temp_dir/Makefile"
     docker build -t "${dra_image_repo}:${dra_image_tag}" \
         --build-arg GOLANG_VERSION="${go_version}" \
+        --build-arg GO_VERSION="${go_version}" \
```
sohankunkerkar, line 1403: Any reason we need this?

PannagaRao: v0.3.0 renamed the Dockerfile build arg from [GOLANG_VERSION](https://github.com/kubernetes-sigs/dra-example-driver/blob/v0.2.1/deployments/container/Dockerfile) to [GO_VERSION](https://github.com/kubernetes-sigs/dra-example-driver/blob/v0.3.0/deployments/container/Dockerfile#L15). Kept both to support the two versions, but since we've now bumped to v0.3.0, we only need GO_VERSION. Removing GOLANG_VERSION.

PannagaRao:
The dra/extended e2e tests passed in CI: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/11420/pull-kueue-test-e2e-dra-main/2060833302889107456
  
 This was validated by temporarily redirecting the existing test-e2e-dra prow job to run the extended suite. The test-e2e-dra-extended Makefile target is ready to be wired to a dedicated prow job.

kannon92:
When you get a chance can you rebase

kannon92 (commented):
Sorry I thought the large diff was because it was on top of PD.

I realize it's because we bring in a large depencency. Can we look into ways to not need to vendor DRA-example-driver.

PannagaRao:
> Sorry I thought the large diff was because it was on top of PD.
> 
> I realize it's because we bring in a large depencency. Can we look into ways to not need to vendor DRA-example-driver.

We actually don't need to vendor DRA-example-driver at all. I'm even importing a package which isn't used in the code just to keep the module in go mod file.

Following this to get the value of DRA_EXAMPLE_DRIVER_VERSION using go list instead of passing the hard coded value since that is the pattern followed. Thoughts?

kannon92:
> > Sorry I thought the large diff was because it was on top of PD.
> > 
> > I realize it's because we bring in a large depencency. Can we look into ways to not need to vendor DRA-example-driver.
> 
> We actually don't need to vendor DRA-example-driver at all. I'm even importing a package which isn't used in the code just to keep the module in go mod file.
> 
> Following this to get the value of DRA_EXAMPLE_DRIVER_VERSION using go list instead of passing the hard coded value since that is the pattern followed. Thoughts?

One option could be to move this to hack/tools and use that to grab version if you don't want to hard code version

But you do import flags from DRA-example driver. I think that's why the vendor list grows so much.

PannagaRao:
> > > Sorry I thought the large diff was because it was on top of PD.
> > > I realize it's because we bring in a large depencency. Can we look into ways to not need to vendor DRA-example-driver.
> > 
> > 
> > We actually don't need to vendor DRA-example-driver at all. I'm even importing a package which isn't used in the code just to keep the module in go mod file.
> > Following this to get the value of DRA_EXAMPLE_DRIVER_VERSION using go list instead of passing the hard coded value since that is the pattern followed. Thoughts?
> 
> One option could be to move this to hack/tools and use that to grab version if you don't want to hard code version
> 
> But you do import flags from DRA-example driver. I think that's why the vendor list grows so much.

I'm importing flags only to keep the module pinned in go.mod — it's not used in code. Without it, go mod tidy removes the module and go list -m can't resolve the version.

Currently DRA_EXAMPLE_DRIVER_VERSION uses go list -m against the main go.mod, which is why we need the module there. I'll move it to hack/ools/go.mod. That should prevent the vendoring altogether.

kannon92:
> > > > Sorry I thought the large diff was because it was on top of PD.
> 
> > > > I realize it's because we bring in a large depencency. Can we look into ways to not need to vendor DRA-example-driver.
> 
> > > 
> 
> > > 
> 
> > > We actually don't need to vendor DRA-example-driver at all. I'm even importing a package which isn't used in the code just to keep the module in go mod file.
> 
> > > Following this to get the value of DRA_EXAMPLE_DRIVER_VERSION using go list instead of passing the hard coded value since that is the pattern followed. Thoughts?
> 
> > 
> 
> > One option could be to move this to hack/tools and use that to grab version if you don't want to hard code version
> 
> > 
> 
> > But you do import flags from DRA-example driver. I think that's why the vendor list grows so much.
> 
> 
> 
> I'm importing flags only to keep the module pinned in go.mod — it's not used in code. Without it, go mod tidy removes the module and go list -m can't resolve the version.
> 
> 
> 
> Currently DRA_EXAMPLE_DRIVER_VERSION uses go list -m against the main go.mod, which is why we need the module there. I'll move it to hack/ools/go.mod. That should prevent the vendoring altogether.

Sounds good. That is exactly what I mean.

PannagaRao:
@kannon92 I have moved the module to hack/tools. Could you please take a look? Thanks!

kannon92:
/assign

Going to review this week.

test/e2e/dra/baseline/dra_test.go
```diff
@@ -638,7 +637,7 @@ var _ = ginkgo.Describe("DRA", func() {
 		ginkgo.It("Should track ResourceClaimTemplate and Extended Resource requests separately", func() {
 			ginkgo.By("Creating ResourceClaimTemplate requesting 2 GPUs")
 			rct := utiltesting.MakeResourceClaimTemplate("both-gpu-template", ns.Name).
-				DeviceRequest("gpu-request", draconsts.DriverName, 2).
+				DeviceRequest("gpu-request", "gpu.example.com", 2).
```
kannon92: Could we declare a constant and reuse that here?

`gpu.example.com` is used in a lot of places.

test/e2e/dra/extended/dra_test.go
```diff
@@ -0,0 +1,402 @@
...
+				Name:      workloadjob.GetWorkloadNameForJob(job.Name, job.UID),
+				Namespace: ns.Name,
+			}
+
+			ginkgo.By("Verifying workload is admitted with counter charge of 20Gi")
+			createdWorkload := &kueue.Workload{}
+			gomega.Eventually(func(g gomega.Gomega) {
+				g.Expect(k8sClient.Get(ctx, wlLookupKey, createdWorkload)).To(gomega.Succeed())
+				if !workload.HasQuotaReservation(createdWorkload) {
+					logWorkloadConditions(createdWorkload)
```
kannon92: I'm not sure I would include these changes. This eventually polls so it would make the logs quity noisy and difficult to find the actual errors you are looking for.

WDYT @mbobrovskyi?

mbobrovskyi: I think we should use util.AssertMsg(). It will include the object in the error message, which makes debugging easier.

test/e2e/dra/baseline/dra_test.go
PannagaRao: Done

test/e2e/dra/extended/dra_test.go
PannagaRao: Changed the code to use util.AssertMsg()

test/e2e/dra/extended/dra_test.go
```diff
@@ -0,0 +1,390 @@
...
+
+			wlLookupKey := types.NamespacedName{
+				Name:      workloadjob.GetWorkloadNameForJob(job.Name, job.UID),
+				Namespace: ns.Name,
+			}
+
+			ginkgo.By("Verifying workload is admitted with counter charge of 20Gi")
+			createdWorkload := &kueue.Workload{}
+			gomega.Eventually(func(g gomega.Gomega) {
+				g.Expect(k8sClient.Get(ctx, wlLookupKey, createdWorkload)).To(gomega.Succeed())
```
kannon92: nit:
these eventuallys all look very similar.

Can we refactor this into a function that just depends on the few differences (memory).

test/e2e/dra/extended/dra_test.go
```diff
@@ -0,0 +1,390 @@
...
+
+				assignment := createdWorkload.Status.Admission.PodSetAssignments[0]
+				memUsage := assignment.ResourceUsage["gpu.memory"]
+				g.Expect(memUsage.Cmp(resource.MustParse("40Gi"))).To(gomega.Equal(0))
+			}, util.Timeout, util.Interval).Should(gomega.Succeed())
+
+			util.ExpectWorkloadToFinishWithTimeout(ctx, k8sClient, wlLookupKey, util.LongTimeout)
+		})
+
+		ginkgo.It("Should charge MAX when CEL matches multiple device types", func() {
```
kannon92: can we just say 
```suggestion
		ginkgo.It("Should charge 80Gi when CEL matches multiple device types", func() {
```

Reading this I am not sure what MAX is. Is it a special keyword for driver or just short hand to say full mig gpu?

kannon92:
@sohankunkerkar @PannagaRao 
I am surprised to see that [pull-kueue-test-e2e-dra-main](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/11420/pull-kueue-test-e2e-dra-main/2062738042547867648) is running only the PD tests.

Where are the other e2e tests for DRA?

I'd expect to see pull-kueue-test-e2e-dra-baseline-main and pull-kueue-test-e2e-dra-extended-main.

go.mod
```diff
@@ -42,7 +42,6 @@ require (
 	k8s.io/utils v0.0.0-20260507154919-ff6756f316d2
 	sigs.k8s.io/cluster-inventory-api v0.0.0-20251028164203-2e3fabb46733
 	sigs.k8s.io/controller-runtime v0.24.0
-	sigs.k8s.io/dra-example-driver v0.2.1
```
kannon92: @sohankunkerkar @PannagaRao @mimowo 

Could we consider carrying this change in our release branches?

It would be nice to decouple our release branches from this dependency. So I think maybe splitting this go.mod removal as a separate PR would be a good idea.

kannon92: A quick check, this dependency is in release-0.17 and release-0.18.

PannagaRao:
> @sohankunkerkar @PannagaRao 
> 
> I am surprised to see that [pull-kueue-test-e2e-dra-main](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/11420/pull-kueue-test-e2e-dra-main/2062738042547867648) is running only the PD tests.
> 
> 
> 
> Where are the other e2e tests for DRA?
> 
> 
> 
> I'd expect to see pull-kueue-test-e2e-dra-baseline-main and pull-kueue-test-e2e-dra-extended-main.


Yeah that is due to the 3rd commit (hack) I have used in this PR to make PD tests run to check whether they pass as there is no infra created yet to run e2e-dra-extended target. I will be dropping this hack commit before the merge.

kannon92:
> Yeah that is due to the 3rd commit (hack) I have used in this PR to make PD tests run to check whether they pass as there is no infra created yet to run e2e-dra-extended target. I will be dropping this hack commit before the merge.

Okay.

test/e2e/dra/extended/dra_test.go
PannagaRao: You are right. Refactored it.

test/e2e/dra/extended/dra_test.go
PannagaRao: Changed it to `Should charge largest counter value when CEL matches multiple device types`

Makefile-test.mk
```diff
@@ -465,8 +444,9 @@ run-test-e2e-dra-%:
 		E2E_MODE=$(E2E_MODE) \
 		E2E_SKIP_REINSTALL=$(E2E_SKIP_REINSTALL) \
 		E2E_ENFORCE_OPERATOR_UPDATE=$(E2E_ENFORCE_OPERATOR_UPDATE) \
-		KIND_CLUSTER_FILE="kind-cluster.yaml" E2E_TARGET_FOLDER="dra/baseline" \
+		KIND_CLUSTER_FILE="kind-cluster.yaml" E2E_TARGET_FOLDER="dra/extended" \
 		DRA_EXAMPLE_DRIVER_VERSION=$(DRA_EXAMPLE_DRIVER_VERSION) \
+		DRA_GPU_PARTITIONS=4 \
```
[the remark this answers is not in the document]
PannagaRao: I will change it back after I verify the test results

go.mod
PannagaRao: Will split the dra-example-driver removal into a separate PR. This PR will depend on it. Does that sound good?

PannagaRao: I have created a separate PR for the driver version bump. https://github.com/kubernetes-sigs/kueue/pull/12097

PannagaRao:
Latest reference of CI that ran the Partitionable Devices E2E from this PR. https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/11420/pull-kueue-test-e2e-dra-main/2064901042931240960

kannon92 (commented):
/lgtm

test/e2e/config/dra/counter/controller_manager_config.yaml
```diff
@@ -0,0 +1,36 @@
...
+    LocalQueue.kueue.x-k8s.io: 1
+    ClusterQueue.kueue.x-k8s.io: 1
+    ResourceFlavor.kueue.x-k8s.io: 1
+clientConnection:
+  qps: 50
+  burst: 100
+integrations:
+  frameworks:
+  - "batch/job"
+  - "pod"
```
mimowo, line 21: Since we don't use any custom frameworks what about putting the tests under "baseline" job? That is pretty much the convention for other CI e2e jobs we have. 

cc @sohankunkerkar wdyt?

kannon92: https://github.com/kubernetes-sigs/kueue/pull/11625#issuecomment-4543910489

Seems that PD was going to go into extended.

kannon92: I like the idea of extended for alpha features as it may be harder to test/support more advanced DRA features.

I'm not really sure we need to do a lot of testing of different workloads for DRA.

mimowo: yeah. but I'm basically trying to challenge that to see if we really need that. My main concern is consistency, currently extended= requires extra integrations. If "baseline" does not fit, then maybe we could come up with another name, like "subfeatures", to reflect this is focused on sub-features of DRA

mimowo: alternative is to rename our cuurent convention:
- baseline -> core
- extended -> integrations

then extended for dra could probably be a good name.

sohankunkerkar: I think @mimowo's point about naming consistency is valid. How about naming after the DRA allocation model instead? `dra/whole-device/` for the current baseline and `dra/counter/` for partitionable devices. When consumable capacity (KEP-5075) lands, it naturally becomes `dra/capacity/`. Each name maps to the allocation model being tested

test/e2e/dra/extended/dra_test.go
```diff
@@ -0,0 +1,364 @@
...
+				Namespace: ns.Name,
+			}
+
+			for _, wlKey := range []types.NamespacedName{wlLookupKey1, wlLookupKey2} {
+				createdWorkload := &kueue.Workload{}
+				gomega.Eventually(func(g gomega.Gomega) {
+					g.Expect(k8sClient.Get(ctx, wlKey, createdWorkload)).To(gomega.Succeed())
+					g.Expect(workload.HasQuotaReservation(createdWorkload)).To(gomega.BeTrue())
+
+					assignment := createdWorkload.Status.Admission.PodSetAssignments[0]
```
sohankunkerkar: Can we use `expectWorkloadCounterCharge` instead?

PannagaRao: Yes, fixed

test/e2e/config/dra/counter/controller_manager_config.yaml
PannagaRao: Renamed the DRA e2e suites to reflect the allocation model:
- dra/extended → dra/counter 
- dra/baseline → dra/whole-device (added as a separate commit in this PR)

Please let me know if it looks good?

test/e2e/dra/whole-device/dra_test.go
sohankunkerkar: can you add a constant for this here: https://github.com/kubernetes-sigs/kueue/blob/main/test/util/constants.go#L31? You can then reuse that for both `whole-device` and `counter`.

PannagaRao: Done

sohankunkerkar (approved):
Thanks! 
/lgtm
/hold in case @mimowo wants to add anything here.

mimowo:
/unhold
Thank you, would you like to also CP the target? If so, the free free to do so.

PR #11497, opened 2026-05-22
MultiKueue: avoid unnecessary workload updates from AllAtOnce dispatcher due to reordered nominatedClusterNames

### LINKED ISSUE #11453: MultiKueue: possible "useless" update of the order of nominatedClusterNames in AllAtOnce
**What happened**:

The AllAtOnce dispatcher compares `group.local.Status.NominatedClusterNames, nominatedWorkers` and if they differ then it sends an update to the `group.local.Status.NominatedClusterNames`.

However, I think since `nominatedWorkers` are determined based on iterating over the map keys then there is no guarantee on their order. From the golang docs: "When iterating over a map with a range loop, the iteration order is not specified and **is not guaranteed to be the same from one iteration to the next.**": https://go.dev/blog/maps

So, I think we should sort the list of `nominatedWorkers` before comparing or setting.

Requests are only set if the "set" of workers is changed.

mszadkow:
#### What type of PR is this?
/kind cleanup

Sort lists of current and previously stored nominated clusters before comparison. 

Fixes #11453 

```release-note
MultiKueue: Fixed unnecessary `status.nominatedClusterNames` updates from the AllAtOnce
dispatcher when the set of nominated clusters did not change.
```

### DIFF
--- a/pkg/controller/admissionchecks/multikueue/workload.go
+++ b/pkg/controller/admissionchecks/multikueue/workload.go
@@ -696,6 +696,14 @@ func (w *wlReconciler) syncToSingleCluster(ctx context.Context, log klog.Logger,
 	return reconcile.Result{}, errors.Join(errs...)
 }
 
+// nominatedClusterSetsEqual reports whether stored and current contain the same set of cluster names,
+// independent of order.
+func nominatedClusterSetsEqual(stored, current []string) bool {
+	slices.Sort(stored)
+	slices.Sort(current)
+	return slices.Equal(stored, current)
+}
+
 func (w *wlReconciler) nominateAndSynchronizeWorkers(ctx context.Context, group *wlGroup) (reconcile.Result, error) {
 	log := ctrl.LoggerFrom(ctx).WithValues("op", "nominateAndSynchronizeWorkers")
 	log.V(3).Info("Nominate and Synchronize Worker Clusters")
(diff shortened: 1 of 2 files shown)


mimowo:
I think this is actually a bugfix, let me know if I'm missing something, labeling as bug tentatively:

/remove-kind cleanup
/kind bug
/release-note-edit
```release-note
MultiKueue: Fixed unnecessary `status.nominatedClusterNames` updates in AllAtOnce
mode when the nominated cluster set had not changed.
```

mimowo:
/release-note-edit
```release-note
MultiKueue: Fixed unnecessary `status.nominatedClusterNames` updates from the AllAtOnce
dispatcher when the set of nominated clusters did not change.
```

mimowo:
Thank you for the quick follow up fix 👍 
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16

mimowo:
/retitle MultiKueue: avoid unnecessary workload updates from AllAtOnce dispatcher due to reordered nominatedClusterNames

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -764,7 +772,7 @@ func (w *wlReconciler) nominateAndSynchronizeWorkers(ctx context.Context, group
 			nominatedWorkers = append(nominatedWorkers, workerName)
 		}
 
-		if !equality.Semantic.DeepEqual(group.local.Status.NominatedClusterNames, nominatedWorkers) {
+		if !nominatedClusterSetsEqual(group.local.Status.NominatedClusterNames, nominatedWorkers) {
```
mwielgus, line 775: Wouldn't that help not only with AllAtOnce but also other strategies?

mszadkow: Not really, others are incremental or external and they simply copy paste what's in the nominated list:
```
} else {
		// Incremental dispatcher and External dispatcher path
		nominatedWorkers = group.local.Status.NominatedClusterNames
	}
```

mimowo: I think the question is rather about code which is responsible for sending the patches - from within dispatchers

PR #11502, opened 2026-05-22
Ensure admitted workload is not present in `preemptionExpectations`

### LINKED ISSUE #11480: Race condition in scheduler causing stuck preemptions and quota oversubscription
### What happened
When the scheduler admits a workload, it optimistically sets `QuotaReserved=True` for that workload in the cache, while the API server update is handled asynchronously via a [goroutine](https://github.com/kubernetes-sigs/kueue/blob/6a7157c339312e8271711e8a35ea861a25274724/pkg/scheduler/scheduler.go#L803). This introduces indeterminism, as the admission might happen after the next scheduler cycle already started or way later, after several cycles already treated the workload as admitted.

#### Scenario 1 - Infinitely Stuck Preemption
Consider the following sequence of events:
1. **Scheduling Cycle 1**: Workload1 is admitted - it is optimistically added to the cache. The goroutine to update it in the API server is spawned.
2. **Scheduling Cycle 2**: Workload2 preempts Workload1. It sends a synchronous update to the API server to [add the `Evicted` condition](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/workload/workload.go#L1765) to Workload1 and [adds Workload1 to `preemptionExpectation`](https://github.com/kubernetes-sigs/kueue/blob/6a7157c339312e8271711e8a35ea861a25274724/pkg/scheduler/preemption/preemption.go#L221).
3. **Workload1 Admission Goroutine**: The goroutine [sends a request](https://github.com/kubernetes-sigs/kueue/blob/6a7157c339312e8271711e8a35ea861a25274724/pkg/scheduler/scheduler.go#L805) to the API server to admit the workload. But the update is based off the [state of Workload1 in **Scheduling Cycle 1**](https://github.com/kubernetes-sigs/kueue/blob/6a7157c339312e8271711e8a35ea861a25274724/pkg/scheduler/scheduler.go#L802). **This overwrites the Evicted condition**.
4. **Scheduling Cycle 3**: Workload2 is evaluated again. It picks Workload1 as it's preemption target. Since Workload1 does not have the Evicted condition anymore, [it won't hit the branch](https://github.com/kubernetes-sigs/kueue/blob/6a7157c339312e8271711e8a35ea861a25274724/pkg/scheduler/preemption/preemption.go#L204-L209) that would observe the eviction in `preemptionExpectations`. Instead, because Workload1 was never observed in `preemptionExpectations`, it will [fall into a fast exist](https://github.com/kubernetes-sigs/kueue/blob/6a7157c339312e8271711e8a35ea861a25274724/pkg/scheduler/preemption/preemption.go#L210-L216). Workload2 is not admitted.
5. **Scheduling Cycle 4**: Is exactly the same as the third cycle.
6. **Scheduling Cycle N**: This continues until Workload1 is deleted or finishes.

#### Scenario 2 - Quota Oversubcription
If step 3 in the previous scenario happens **after Scheduling Cycle 3**, the `Evicted` condition will not be overwritten. Instead:
1. **Scheduling Cycle 1**: Workload1 is admitted - it is optimistically added to the cache. The goroutine to update it in the API server is spawned.
2. **Scheduling Cycle 2**: Workload2 preempts Workload1. It sends a synchronous update to the API server to [add the `Evicted` condition](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/workload/workload.go#L1765) to Workload1 and [adds Workload1 to `preemptionExpectation`](https://github.com/kubernetes-sigs/kueue/blob/6a7157c339312e8271711e8a35ea861a25274724/pkg/scheduler/preemption/preemption.go#L221).
3. **Scheduling Cycle 3**: Sees that Workload1 is `Evicted` and [observes it](https://github.com/kubernetes-sigs/kueue/blob/6a7157c339312e8271711e8a35ea861a25274724/pkg/scheduler/preemption/preemption.go#L204-L209).
4. **Job Controller**: The controller for Workload1's job evicts it.
5. **Scheduling Cycle 4**: Workload2 is admitted.
6. **Workload1 Admission Goroutine**: The goroutine admits the workload without any quota checks, causing oversubcription.

---

I will admit, I did not trace the second scenario very closely, I only reproduced it. Out of the two, it seems like a less severe one.

Scenario 1 was observed on production and the impact is a **complete lack of admission** for some ClusterQueues in the cluster.

### How to reproduce it (as minimally and precisely as possible)
I wrote a very hacky reproduction by editing the Kueue code to force a race condition from Scenario 1. It's structured as a E2E test on my local fork ([link](https://github.com/kubernetes-sigs/kueue/commit/6db3082613caa25111f5d72a625855bd85b22f05)).

I achieved the race condition by forcing a sequence via the use of channels.

It also reproduces the flood of:
```
Preemption already issued, waiting for observation
```
logs, which is consistent with the production investigation.

### Anything else we need to know?
This was found during an investigation of a production issue.

kshalot:
#### What type of PR is this?
/kind bug

This addresses a race condition, where a workload might be kept in `preemptionExpectations` despite being Admitted and not having the Evicted condition.

Fixes #11480

The reproduction I linked in #11480 is itself flaky - the processing order of some events seems to be relevant. Without this fix, the reproduction succeeds within minutes. I ran it with this fix over the weekend and it did not fail once.

```release-note
Scheduling: Fix a race condition within the admission process that could cause workloads waiting indefinitely for a preemption, causing head-of-line blocking of the affected ClusterQueues.
```

### DIFF
--- a/test/e2e/singlecluster/baseline/admission_routine_race_condition_test.go
+++ b/test/e2e/singlecluster/baseline/admission_routine_race_condition_test.go
@@ -25,6 +25,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
 	workloadjob "sigs.k8s.io/kueue/pkg/controller/jobs/job"
 	utiltestingapi "sigs.k8s.io/kueue/pkg/util/testing/v1beta2"
--- a/test/e2e/singlecluster/baseline/admission_routine_race_condition_test.go
+++ b/test/e2e/singlecluster/baseline/admission_routine_race_condition_test.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2024 The Kubernetes Authors.
+Copyright The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -30,13 +30,13 @@ import (
 	workloadjob "sigs.k8s.io/kueue/pkg/controller/jobs/job"
 	utiltestingapi "sigs.k8s.io/kueue/pkg/util/testing/v1beta2"
 	testingjob "sigs.k8s.io/kueue/pkg/util/testingjobs/job"
-	workload "sigs.k8s.io/kueue/pkg/workload"
+	"sigs.k8s.io/kueue/pkg/workload"
 	"sigs.k8s.io/kueue/test/util"
 )
 
 // This test is the e2e reproduction of the issue described in https://github.com/kubernetes-sigs/kueue/issues/11480.
 // In practice, it should never fail (even without the fix), as the issue is extremely rare.
-// The following commit containts the diff that has to be injected into Kueue in order to make this test fail non-deterministically:
+// The following commit contains the diff that has to be injected into Kueue in order to make this test fail non-deterministically:
 // https://github.com/kshalot/kueue/commit/78f5d0217f55663d8c6ecc8c486704e9f0ae6570
 var _ = ginkgo.Describe("Admission Routine Race Condition", ginkgo.Label("area:singlecluster", "feature:scheduler"), func() {
 	var (
--- a/test/e2e/singlecluster/baseline/admission_routine_race_condition_test.go
+++ b/test/e2e/singlecluster/baseline/admission_routine_race_condition_test.go
@@ -24,7 +24,6 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
-
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
 	workloadjob "sigs.k8s.io/kueue/pkg/controller/jobs/job"
--- a/test/e2e/singlecluster/baseline/admission_routine_race_condition_test.go
+++ b/test/e2e/singlecluster/baseline/admission_routine_race_condition_test.go
@@ -142,6 +142,6 @@ var _ = ginkgo.Describe("Admission Routine Race Condition", ginkgo.Label("area:s
 
 			g.Expect(workload.HasQuotaReservation(wl1)).To(gomega.BeFalse())
 			g.Expect(workload.HasQuotaReservation(wl2)).To(gomega.BeTrue())
-		}, util.Timeout, util.Interval).Should(gomega.Succeed())
+		}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
 	})
 })
--- a/test/e2e/singlecluster/baseline/admission_routine_race_condition_test.go
+++ b/test/e2e/singlecluster/baseline/admission_routine_race_condition_test.go
@@ -1,0 +1,147 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package baseline
+
+import (
+	"time"
+
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+	workloadjob "sigs.k8s.io/kueue/pkg/controller/jobs/job"
+	utiltestingapi "sigs.k8s.io/kueue/pkg/util/testing/v1beta2"
+	testingjob "sigs.k8s.io/kueue/pkg/util/testingjobs/job"
+	workload "sigs.k8s.io/kueue/pkg/workload"
+	"sigs.k8s.io/kueue/test/util"
+)
+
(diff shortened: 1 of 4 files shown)


test/integration/singlecluster/scheduler/inadmissible/suite_test.go
```diff
@@ -89,7 +89,7 @@ func managerAndSchedulerSetup(ctx context.Context, mgr manager.Manager) {
 	err = workloadjob.SetupIndexes(ctx, mgr.GetFieldIndexer())
 	gomega.Expect(err).NotTo(gomega.HaveOccurred())
 
-	sched := scheduler.New(queues, cCache, mgr.GetClient(), mgr.GetEventRecorder(constants.AdmissionName))
+	sched := scheduler.New(queues, cCache, mgr.GetClient(), mgr.GetEventRecorder(constants.AdmissionName), scheduler.WithPreemptionExpectations((preemptionExpectations)))
```
pajakd, line 92: Is this fixing a bug in tests?

pajakd:
You mention that you were testing it with an e2e test that is flaky without the fix and reliable with the fix. Is there are reason not to add this test?

mimowo:
or integeation tets. integration test might be better as we can use interceptor to control timings

test/integration/singlecluster/scheduler/inadmissible/suite_test.go
kshalot: this change would break the test, since it requires the scheduler to have `preemptionExpectations`. I'm not sure it's a bug per-se, as maybe it was not required for this suite up until now. Now it is required.

kshalot:
> You mention that you were testing it with an e2e test that is flaky without the fix and reliable with the fix. Is there are reason not to add this test?

@pajakd The e2e repro relies on changing the binary, it's not possible to reliably introduce this race condition without injecting sleeps.

> or integeation tets. integration test might be better as we can use interceptor to control timings

@mimowo The integration test was my first attempt, but I encountered some nuance that prevented it. I'll look back to it to confirm.

kshalot:
@mimowo, as discussed, I added the e2e repro with a comment explaining on how to make it work (non-deterministically). As a follow-up, I'd continue looking into the integration repro and remove the e2e one if there is one.

test/e2e/singlecluster/baseline/admission_routine_race_condition_test.go
```diff
@@ -0,0 +1,147 @@
...
+		ginkgo.By("Verifying the race condition occurred")
+		gomega.Consistently(func(g gomega.Gomega) {
+			// Because wl1's admission patch used SSA and lacked the Evicted condition,
+			// it overwrote job2's eviction attempt on wl1. wl1 retains the quota, and wl2 gets stuck.
+			g.Expect(k8sClient.Get(ctx, wl1LookupKey, wl1)).To(gomega.Succeed())
+			g.Expect(k8sClient.Get(ctx, wl2LookupKey, wl2)).To(gomega.Succeed())
+
+			g.Expect(workload.HasQuotaReservation(wl1)).To(gomega.BeTrue(), "wl1 should have quota reservation")
+			g.Expect(workload.HasQuotaReservation(wl2)).To(gomega.BeFalse(), "wl2 should be blocked")
+		}, util.Timeout, util.Interval).Should(gomega.Succeed())
```
mimowo: Use constants for Consistently

kshalot: it has to be changed to Eventually in this case anyway. Fixed.

test/e2e/singlecluster/baseline/admission_routine_race_condition_test.go
```diff
@@ -0,0 +1,147 @@
...
+	testingjob "sigs.k8s.io/kueue/pkg/util/testingjobs/job"
+	"sigs.k8s.io/kueue/pkg/workload"
+	"sigs.k8s.io/kueue/test/util"
+)
+
+// This test is the e2e reproduction of the issue described in https://github.com/kubernetes-sigs/kueue/issues/11480.
+// In practice, it should never fail (even without the fix), as the issue is extremely rare.
+// The following commit contains the diff that has to be injected into Kueue in order to make this test fail non-deterministically:
+// https://github.com/kshalot/kueue/commit/78f5d0217f55663d8c6ecc8c486704e9f0ae6570
+var _ = ginkgo.Describe("Admission Routine Race Condition", ginkgo.Label("area:singlecluster", "feature:scheduler"), func() {
```
mimowo, line 41: Yeah, I think we need to replace the e2e test with an integration test in a follow up.

test/e2e/singlecluster/baseline/admission_routine_race_condition_test.go
```diff
@@ -0,0 +1,147 @@
...
+		ginkgo.By("Verifying the race condition occurred")
+		gomega.Eventually(func(g gomega.Gomega) {
+			// Because wl1's admission patch used SSA and lacked the Evicted condition,
+			// it overwrote job2's eviction attempt on wl1. wl1 retains the quota, and wl2 gets stuck.
+			g.Expect(k8sClient.Get(ctx, wl1LookupKey, wl1)).To(gomega.Succeed())
+			g.Expect(k8sClient.Get(ctx, wl2LookupKey, wl2)).To(gomega.Succeed())
+
+			g.Expect(workload.HasQuotaReservation(wl1)).To(gomega.BeFalse())
+			g.Expect(workload.HasQuotaReservation(wl2)).To(gomega.BeTrue())
+		}, util.Timeout, util.Interval).Should(gomega.Succeed())
```
mimowo:
```suggestion
		}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
```

mimowo:
Let me merge using the e2e test even though it does not repro the problem reliably, because:
- I was involved in debugging and can confirm the bug that occasionally preemptions may get stuck currently
- @kshalot reproduced the problem locally with code modifications with sleep injection. After the code change the bug didn't repro
- I think the bugfix makes sense

Let's follow up trying to turn the test into an  integration test with a higher repro chance. I believe using intereptor should allow us to inject sliip in the admit() and preempt() routines to make the flake reproducible reliably,

mimowo:
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16
Thank you for fixing the important bug 👍 

cc @sohankunkerkar who was also initially working on the preemptionExpectations.

mimowo:
/lgtm
Opened: https://github.com/kubernetes-sigs/kueue/issues/11643

kshalot:
Manual cherry-picks:
`release/0.16` - https://github.com/kubernetes-sigs/kueue/pull/11647
`release/0.17` - https://github.com/kubernetes-sigs/kueue/pull/11648

mimowo:
/release-note-edit
```release-note
Scheduling: Fix a race condition within the admission process that could cause workloads waiting indefinitely for a preemption, causing head-of-line blocking of the affected ClusterQueues.
```

PR #11504, opened 2026-05-22
Add kueue_flavor_admitted_active_workloads metric

polinasand:
#### What type of PR is this?

/kind feature

This PR adds new metric `kueue_falvor_admitted_active_workloads` which contains admitted workloads count in CQ broken down by flavor.

Progresses #11644

```
NONE
```


pkg/cache/scheduler/clusterqueue.go
```diff
@@ -582,6 +588,7 @@ func (c *clusterQueue) updateWorkloadUsage(log logr.Logger, wi *workload.Info, o
 	c.updateWorkloadTASUsage(log, wi, op)
 	if admitted {
 		updateFlavorUsage(frUsage, c.AdmittedUsage, op)
+		updateAdmittedWorkloadsCountByFlavor(frUsage, c.admittedWorkloadsCountByFlavor)
```
olekzabl: Take `op` into account; there must be a way to decrease our counters as well.
Please look how it's done in `updateFlavorUsage`, one line above.

pkg/features/kube_features.go
```diff
@@ -428,6 +428,13 @@ const (
 	// issue: https://github.com/kubernetes-sigs/kueue/issues/7539
 	// Enable reporting of Cohort related metrics (also including ClusterQueueInfo metric).
 	MetricsForCohorts featuregate.Feature = "MetricsForCohorts"
+
+  // owner: @polinasand
+	// kep: https://github.com/kubernetes-sigs/kueue/issues/11644
+	//
+	// Whern enabled, metrics for Kueue active admitted and pending workloads count by flavor
```
olekzabl:
```suggestion
	// When enabled, metrics for Kueue active admitted and pending workloads count by flavor
```

pkg/features/kube_features.go
```diff
@@ -428,6 +428,13 @@ const (
...
+	// are populated by clusterqueue.
+	KueueAdmittedAndPendingWorkloadsCountByFlavor featuregate.Feature = "KueueAdmittedAndPendingWorkloadsCountByFlavor"
```
olekzabl: I'd keep admitted & pending behind 2 separate FGs.
Both the implementation & the potential "controversies" about metric semantics are going to be independent.
(In particular, I imagine we could ship "admitted" noticeably before "pending"; in that case, I wouldn't like the FG name to be over-promising)

pkg/cache/scheduler/clusterqueue.go
```diff
@@ -630,6 +637,15 @@ func updateFlavorUsage(newUsage resources.FlavorResourceQuantities, oldUsage res
 	}
 }
 
+func updateAdmittedWorkloadsCountByFlavor(frUsage resources.FlavorResourceQuantities, admittedWorkloadsCountByFlavor resources.FlavorResourceQuantities) {
+	for fr := range frUsage {
+		if _, ok := admittedWorkloadsCountByFlavor[fr]; !ok {
+			admittedWorkloadsCountByFlavor[fr] = 0
+		}
```
olekzabl: I think this guard is unnecessary.
Golang maps return zero values for missing keys.
See https://go.dev/play/p/5eskYsQE4E_d

pkg/metrics/metrics.go
```diff
@@ -1370,6 +1391,7 @@ func Register() {
 		WorkloadEvictionLatencySeconds,
 		ReservingActiveWorkloads,
 		AdmittedActiveWorkloads,
+		FlavorAdmittedActiveWorkloads,
```
olekzabl: How about registering behind our FG?
See how it's done for `WorkloadCreationLatency` below.

olekzabl:
The code now LGTM, thank you!

Now, thinking of what more we should change, I guess these places:

- integration tests in [this file](https://github.com/kubernetes-sigs/kueue/blob/main/test/integration/singlecluster/scheduler/fairsharing/metrics_test.go) (for inspiration search for `AdmittedActiveWorlkoads`) - maybe not all places but at least one
- [this doc page](https://github.com/kubernetes-sigs/kueue/blob/main/site/content/en/docs/reference/metrics.md) (search for `admitted_active_workloads`, or just look at #10004 for inspiration)

kshalot:
/close

In favor of #11888

PR #11515, opened 2026-05-22
Fix multikueue remote workload cleanup when worker is unavailable

### LINKED ISSUE #11346: MultiKueue when an additional admission check covering only some flavors supported by the cluster queue is present should keep multikueue admission check reference and properly clean up the remote workload when the connection to an admitting worker is lost
**Which test is flaking?**:

MultiKueue when an additional admission check covering only some flavors supported by the cluster queue is present should keep multikueue admission check reference and properly clean up the remote workload when the connection to an admitting worker is lost
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/11343/pull-kueue-test-integration-multikueue-release-0-16/2057018080189485056

**Failure message or logs**:
```
slow] expand_less	1m39s
{ failed [FAILED] Timed out after 90.001s.
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/kueue/test/integration/multikueue/jobs_test.go:2509 with:
Error matcher expects an error.  Got:
    <nil>: nil
In [It] at: /home/prow/go/src/sigs.k8s.io/kueue/test/integration/multikueue/jobs_test.go:2510 @ 05/20/26 08:48:43.461
}
```

vamsikrishna-siddu:
#### What type of PR is this?
/kind bug
/kind flake
/area multikueue

When a worker cluster is disconnected or reconnecting, `remoteClientsForAC` silently excludes it from the available clients map. If the local workload loses its quota reservation during this window, step 2 of `reconcileGroup` deletes remote workloads only on reachable clusters and returns with no `RequeueAfter`, leaving the unreachable cluster's remote workload orphaned.
This PR tracks whether any configured clusters were unavailable when building the workload group and, when so, returns `RequeueAfter` so cleanup is retried once the cluster is back.

Fixes #11346

The same gap exists in the `isDeleted` path of `Reconcile` (deleted workload cleanup also skips unavailable clusters). This PR fixes that path too.

```release-note
MultiKueue: Fixed a bug where obsolete remote Workloads could remain on temporarily unavailable worker clusters when the manager Workload lost its reservation or was deleted. Kueue now retries cleanup after worker clusters reconnect.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of remote workloads when a cluster is temporarily disconnected or reconnecting.
  * Reconciliation now waits and retries cleanup when unavailable clusters are detected, helping prevent premature failures.
  * Orphaned remote workloads are now removed correctly after connectivity is restored.
* **Tests**
  * Added and expanded coverage for reconnect scenarios, including requeue behavior and cleanup across multiple workers.
  * Added an integration test covering job completion and remote workload deletion after reconnect.

### DIFF
--- a/pkg/controller/admissionchecks/multikueue/workload.go
+++ b/pkg/controller/admissionchecks/multikueue/workload.go
@@ -103,13 +103,14 @@ type wlReconciler struct {
 var _ reconcile.Reconciler = (*wlReconciler)(nil)
 
 type wlGroup struct {
-	local         *kueue.Workload
-	localClient   client.Client
-	remotes       map[string]*kueue.Workload
-	remoteClients map[string]*remoteClient
-	acName        kueue.AdmissionCheckReference
-	jobAdapter    jobframework.MultiKueueAdapter
-	controllerKey types.NamespacedName
+	local               *kueue.Workload
+	localClient         client.Client
+	remotes             map[string]*kueue.Workload
+	remoteClients       map[string]*remoteClient
+	acName              kueue.AdmissionCheckReference
+	jobAdapter          jobframework.MultiKueueAdapter
+	controllerKey       types.NamespacedName
+	unavailableClusters []string
 }
 
 type Option func(reconciler *wlReconciler)
@@ -258,6 +259,8 @@ func (w *wlReconciler) Reconcile(ctx context.Context, req reconcile.Request) (re
 				return reconcile.Result{}, err
 			}
 		}
+		// Remote workloads on unavailable clusters will be cleaned up by
+		// the per-cluster GC once the cluster reconnects.
 		w.deletedWlCache.Delete(req.String())
 		return reconcile.Result{}, nil
 	}
@@ -274,26 +277,20 @@ func (w *wlReconciler) updateACS(ctx context.Context, wl *kueue.Workload, acs *k
 	})
 }
 
-func (w *wlReconciler) remoteClientsForAC(ctx context.Context, acName kueue.AdmissionCheckReference) (map[string]*remoteClient, error) {
+func (w *wlReconciler) remoteClientsForAC(ctx context.Context, acName kueue.AdmissionCheckReference) (availableClients map[string]*remoteClient, unavailableClusters []string, err error) {
 	cfg, err := w.helper.ConfigForAdmissionCheck(ctx, acName)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
-	clients := make(map[string]*remoteClient, len(cfg.Spec.Clusters))
+	availableClients = make(map[string]*remoteClient, len(cfg.Spec.Clusters))
 	for _, clusterName := range cfg.Spec.Clusters {
-		if client, found := w.clusters.controllerFor(clusterName); found {
-			// Skip the client if its reconnect is ongoing or it is disconnected.
-			if !client.connecting.Load() && !client.disconnected.Load() {
-				clients[clusterName] = client
-			}
+		if client, found := w.clusters.controllerFor(clusterName); found && !client.connecting.Load() && !client.disconnected.Load() {
+			availableClients[clusterName] = client
+		} else {
+			unavailableClusters = append(unavailableClusters, clusterName)
 		}
 	}
-	// Return an empty map (rather than an error) when all configured clusters
-	// are disconnected or reconnecting. This allows readGroup to construct an
-	// empty group, so reconcileGroup can apply the workerLostTimeout delay via
-	// the existing "reserving remote lost" branch instead of failing with
-	// ErrNoActiveClusters and retrying via controller-runtime backoff.
-	return clients, nil
+	return availableClients, unavailableClusters, nil
 }
 
 func (w *wlReconciler) adapter(local *kueue.Workload) (jobframework.MultiKueueAdapter, *metav1.OwnerReference) {
@@ -322,19 +319,20 @@ func (w *wlReconciler) adapter(local *kueue.Workload) (jobframework.MultiKueueAd
 }
 
 func (w *wlReconciler) readGroup(ctx context.Context, local *kueue.Workload, acName kueue.AdmissionCheckReference, adapter jobframework.MultiKueueAdapter, controllerName string) (*wlGroup, error) {
-	rClients, err := w.remoteClientsForAC(ctx, acName)
+	rClients, unavailable, err := w.remoteClientsForAC(ctx, acName)
 	if err != nil {
 		return nil, fmt.Errorf("admission check %q: %w", acName, err)
 	}
 
 	grp := wlGroup{
-		local:         local,
-		localClient:   w.client,
-		remotes:       make(map[string]*kueue.Workload, len(rClients)),
-		remoteClients: rClients,
-		acName:        acName,
-		jobAdapter:    adapter,
-		controllerKey: types.NamespacedName{Name: controllerName, Namespace: local.Namespace},
+		local:               local,
+		localClient:         w.client,
+		remotes:             make(map[string]*kueue.Workload, len(rClients)),
+		remoteClients:       rClients,
+		acName:              acName,
+		jobAdapter:          adapter,
+		controllerKey:       types.NamespacedName{Name: controllerName, Namespace: local.Namespace},
+		unavailableClusters: unavailable,
 	}
 
 	for remote, rClient := range rClients {
@@ -374,6 +372,10 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
 				errs = append(errs, err)
 				log.V(2).Error(err, "Deleting remote workload", "workerCluster", rem)
 			}
+		}
+		if len(group.unavailableClusters) > 0 {
+			log.V(3).Info("Retrying remote workload cleanup, some clusters are unavailable", "unavailableClusters", group.unavailableClusters, "retryAfter", w.workerLostTimeout)
+			return reconcile.Result{RequeueAfter: w.workerLostTimeout}, errors.Join(errs...)
 		}
 		return reconcile.Result{}, errors.Join(errs...)
 	}
--- a/test/integration/multikueue/jobs_test.go
+++ b/test/integration/multikueue/jobs_test.go
@@ -1485,6 +1485,94 @@ var _ = ginkgo.Describe("MultiKueue", ginkgo.Label("area:multikueue", "feature:m
 		})
 	})
 
+	ginkgo.It("should clean up the remote workload on a reconnected worker after the job finishes", framework.SlowSpec, func() {
+		job := testingjob.MakeJob("job", managerNs.Name).
+			Queue(kueue.LocalQueueName(managerLq.Name)).
+			Obj()
+		util.MustCreate(managerTestCluster.ctx, managerTestCluster.client, job)
+		jobLookupKey := client.ObjectKeyFromObject(job)
+
+		createdWorkload := &kueue.Workload{}
+		wlLookupKey := types.NamespacedName{Name: workloadjob.GetWorkloadNameForJob(job.Name, job.UID), Namespace: managerNs.Name}
+
+		admission := utiltestingapi.MakeAdmission(kueue.ClusterQueueReference(managerCq.Name)).
+			PodSets(utiltestingapi.MakePodSetAssignment(kueue.DefaultPodSetName).Flavor(corev1.ResourceCPU, multikueueTestFlavor).Obj()).
+			Obj()
+
+		ginkgo.By("setting workload reservation in the management cluster", func() {
+			util.SetQuotaReservation(managerTestCluster.ctx, managerTestCluster.client, wlLookupKey, admission)
+		})
+
+		ginkgo.By("checking the workload creation in the worker clusters", func() {
+			managerWl := &kueue.Workload{}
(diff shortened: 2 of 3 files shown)


mimowo:
@mszadkow @olekzabl @kshalot ptal

mszadkow:
The solution looks pretty good
what would be the best addition here is to have actual test showing that orphaned workloads get cleaned

vamsikrishna-siddu:
> The solution looks pretty good what would be the best addition here is to have actual test showing that orphaned workloads get cleaned

Thanks for the review! I've added a new test TestOrphanedRemoteWorkloadCleanedAfterReconnect that shows the full cleanup flow.

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -81,13 +81,14 @@ type wlReconciler struct {
...
-	jobAdapter    jobframework.MultiKueueAdapter
-	controllerKey types.NamespacedName
+	local                  *kueue.Workload
+	localClient            client.Client
+	remotes                map[string]*kueue.Workload
+	remoteClients          map[string]*remoteClient
+	acName                 kueue.AdmissionCheckReference
+	jobAdapter             jobframework.MultiKueueAdapter
+	controllerKey          types.NamespacedName
+	hasUnavailableClusters bool
```
mimowo: Thank you for the PR 👍 

This looks to me as an alternative, totally reasonable, bugfix to the issue fixed in parallel by https://github.com/kubernetes-sigs/kueue/pull/11559

If this is so, then I think we can close, but let me know in case this is covering some scenarios not covered by the other PR. 

cc @yuluo-yx @olekzabl

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -250,10 +255,10 @@ func (w *wlReconciler) updateACS(ctx context.Context, wl *kueue.Workload, acs *k
 	})
 }
 
-func (w *wlReconciler) remoteClientsForAC(ctx context.Context, acName kueue.AdmissionCheckReference) (map[string]*remoteClient, error) {
+func (w *wlReconciler) remoteClientsForAC(ctx context.Context, acName kueue.AdmissionCheckReference) (map[string]*remoteClient, bool, error) {
```
Singularity23x0: The signature fails to clarify what the bool value signifies. Naming the return params would be of great help here.
E.g.:

```suggestion
func (w *wlReconciler) remoteClientsForAC(ctx context.Context, acName kueue.AdmissionCheckReference) (availableClients map[string]*remoteClient, hasUnavailable bool, e error) {
```

Singularity23x0: Alternatively: we could return the list of unavailable clients to log in reconcileGroup(), instead of just "**some** clusters are unavailable".

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -234,6 +235,10 @@ func (w *wlReconciler) Reconcile(ctx context.Context, req reconcile.Request) (re
 				return reconcile.Result{}, err
 			}
 		}
+		if grp.hasUnavailableClusters {
+			log.V(3).Info("Retrying cleanup of deleted workload, some clusters are unavailable", "retryAfter", w.workerLostTimeout)
+			return reconcile.Result{RequeueAfter: w.workerLostTimeout}, nil
+		}
 		w.deletedWlCache.Delete(req.String())
 		return reconcile.Result{}, nil
```
Singularity23x0: I am worried this could loop infinitely if we loose the worker for a long time.
What if:
1) we schedule X workloads, all dispatch to worker w1
2) we loose connection to w1
3) all workloads X will finish and are deleted
4) for all of those workloads we will end up here, rescheduling each time to try to clean up w1 after workerLostTimeout

Is the code guarded form attempting it indefinitely?
If not, we could accumulate a large backlog of reconciles all retrying to clear their remote workloads from w1.

pkg/controller/admissionchecks/multikueue/workload.go
Singularity23x0: The two prs seem to differ in the scenarios they address:
* this PR covers cleanup when we are evicted and become pending or we finish and are deleted, but a worker reconnects afterwards - in thsese cases we will make sure to retry and clean up obsolete remotes on said workers; it will still throw an error when no clusters remain connected.
* the other PR makes sure that when when all workers are lost, we do not throw an error but rather proceed - it would allow the delete function to happily say there are no workers (hence no remotes), so nothing to clean up, then finish and leave the workers that reconnect with potential remotes that were not cleaned up.

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -265,9 +270,9 @@ func (w *wlReconciler) remoteClientsForAC(ctx context.Context, acName kueue.Admi
 		}
 	}
 	if len(clients) == 0 {
-		return nil, admissioncheck.ErrNoActiveClusters
+		return nil, false, admissioncheck.ErrNoActiveClusters
```
Singularity23x0: Technically speaking when there are no active clusters, hasUnavailable should be true if we expect some:

```suggestion
		return nil, len(cfg.Spec.Clusters) > 0, admissioncheck.ErrNoActiveClusters
```

pkg/controller/admissionchecks/multikueue/workload.go
yuluo-yx: #11559 fixed an underlying issue by removing lost worker clients from the `remoteClients` mapping and covering `workerLostTimeout`/reserved remote lost processes.

This PR is related to #11559, but the scenario is slightly different. The remaining issue is the cleanup path: when only a portion of the configured cluster is unavailable, `readGroup` only contains reachable remote nodes. If the local workload has lost its quota reservation, or we are cleaning up workloads from `deletedWlCache`,

then `reconcile` currently only removes remote objects from the reachable cluster and then returns without performing `RequeueAfter`. This can cause remote workloads to remain on the unavailable cluster until other events occur after reconnection.

The GC can handle cases where local workloads no longer exist or are being deleted, but it cannot handle cases where "local workloads still exist but no longer have quota reservations" because the GC will treat the remote workload as still relevant. 🤔

mimowo:
cc @ivnovakov ptal

pkg/controller/admissionchecks/multikueue/workload.go
vamsikrishna-siddu: Good suggestion, done. Updated to (availableClients map[string]*remoteClient, hasUnavailable bool, err error).

pkg/controller/admissionchecks/multikueue/workload.go
vamsikrishna-siddu: Good catch. Removed the requeue from the isDeleted path .the per-cluster GC already handles cleanup of remote workloads whose local counterpart no longer exists, so there's no need to retry from here.

pkg/controller/admissionchecks/multikueue/workload.go
vamsikrishna-siddu: Went with this approach remoteClientsForAC now returns unavailableClusters []string instead of a bool. The log in reconcileGroup includes the cluster names:
"Retrying remote workload cleanup, some clusters are unavailable", "unavailableClusters", ["worker2"]

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -250,26 +253,25 @@ func (w *wlReconciler) updateACS(ctx context.Context, wl *kueue.Workload, acs *k
...
+func (w *wlReconciler) remoteClientsForAC(ctx context.Context, acName kueue.AdmissionCheckReference) (availableClients map[string]*remoteClient, unavailableClusters []string, err error) {
 	cfg, err := w.helper.ConfigForAdmissionCheck(ctx, acName)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
-	clients := make(map[string]*remoteClient, len(cfg.Spec.Clusters))
+	availableClients = make(map[string]*remoteClient, len(cfg.Spec.Clusters))
 	for _, clusterName := range cfg.Spec.Clusters {
 		if client, found := w.clusters.controllerFor(clusterName); found {
```
Singularity23x0: Potential simplification.

```suggestion
		if client, found := w.clusters.controllerFor(clusterName); found && !client.connecting.Load() && !client.disconnected.Load() {
```

vamsikrishna-siddu: done.

Singularity23x0:
Looks good overall.
I think it would be good to add an e2e test checking this specific scenario to the MK suite where we:
1. dispatch to both workers, 
2. disconnect w1, 
3. (a) backoff / (b) finish / (c) delete wl on manager,
4. reconnect w1 and see if it cleans up.

vamsikrishna-siddu:
> Looks good overall. I think it would be good to add an e2e test checking this specific scenario to the MK suite where we:
> 
> 1. dispatch to both workers,
> 2. disconnect w1,
> 3. (a) backoff / (b) finish / (c) delete wl on manager,
> 4. reconnect w1 and see if it cleans up.

Added an integration test that covers this dispatches to both workers, disconnects w1, finishes the job on w2, reconnects w1, and verifies the orphaned remote workload on w1 gets cleaned up. The delete and backoff paths were already covered by existing tests.

Singularity23x0 (approved):
/lgtm

vamsikrishna-siddu:
@mimowo  Can you please review this pr when you get some free time?

test/integration/multikueue/jobs_test.go
```diff
@@ -1482,6 +1474,94 @@ var _ = ginkgo.Describe("MultiKueue", ginkgo.Label("area:multikueue", "feature:m
 		})
 	})
 
+	ginkgo.It("should clean up the remote workload on a reconnected worker after the job finishes", framework.SlowSpec, func() {
```
mimowo, line 1488: Is this test failing on the main branch? If so, then which assert it failing specifically, is it "the worker1 remote workload is cleaned up after reconnect"?

mimowo:
@vamsikrishna-siddu since this is a bugfix, updating the production code then we need a release note. It will also improve my understanding of the user-observable impact.

test/integration/multikueue/jobs_test.go
vamsikrishna-siddu: This test is new, it's not on main. The flaky test from the issue is the existing one `should keep multikueue admission check reference and properly clean up the remote workload when the connection to an admitting worker is lost`. I added this new test to cover a slightly different case: a non-admitting worker gets disconnected and its remote workload is left behind. Without the fix, whether that gets cleaned up depends on watch replay timing after reconnect. The RequeueAfter removes that timing dependency.

vamsikrishna-siddu:
> @vamsikrishna-siddu since this is a bugfix, updating the production code then we need a release note. It will also improve my understanding of the user-observable impact.

Done.

mimowo:
>      Fix orphaned remote workloads not being cleaned up in MultiKueue when a worker cluster is temporarily unreachable.

sgtm, but I have two questions:
1.  is the cleanup now immediate after reconnection, or how long the latency could be?
2. before the change would there be the safety net of cleaning the workload by MK garbage collector?

vamsikrishna-siddu:
> > ```
> >  Fix orphaned remote workloads not being cleaned up in MultiKueue when a worker cluster is temporarily unreachable.
> > ```
> 
> sgtm, but I have two questions:
> 
> 1. is the cleanup now immediate after reconnection, or how long the latency could be?
> 2. before the change would there be the safety net of cleaning the workload by MK garbage collector?

**1. is the cleanup now immediate after reconnection, or how long the latency could be?**
Cleanup is not immediate. While a worker is unavailable, we requeue periodically (default workerLostTimeout = 5m). After reconnection, cleanup happens on the next retry, and in some cases earlier if reconnect/watch events trigger reconcile.

**2. before the change would there be the safety net of cleaning the workload by MK garbage collector?**
Before this change, GC was only a partial safety net: it cleans remotes when the local workload is missing or being deleted. If the local workload still exists but has lost quota reservation, GC treats the remote as still relevant and does not remove it. This PR covers that missing cleanup path.

mimowo:
sgtm, I will merge it after the patch release so that it goes in the next one

mimowo:
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Ups, totally missed this one, merging, thank you 👍

mimowo:
/release-note-edit
```release-note
MultiKueue: Fixed a bug where obsolete remote Workloads could remain on temporarily unavailable worker clusters when the manager Workload lost its reservation or was deleted. Kueue now retries cleanup after worker clusters reconnect.
```

mimowo:
ups: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/11515/pull-kueue-test-unit-shard-0-main/2073085697278873600

now this requires 
@rebase
@vamsikrishna-siddu do you have cycles to rebase the PR?

mimowo (commented):
/lgtm
Thank you 👍

mimowo:
@vamsikrishna-siddu please prepare the CP for CPs manually because their build fail, https://github.com/kubernetes-sigs/kueue/pull/12800#issuecomment-4890080478

vamsikrishna-siddu:
> @vamsikrishna-siddu please prepare the CP for CPs manually because their build fail, [#12800 (comment)](https://github.com/kubernetes-sigs/kueue/pull/12800#issuecomment-4890080478)

@mimowo  Done.

PR #11540, opened 2026-05-22
apis: add shortNames to Kueue CRDs

gyliu513:
#### What type of PR is this?

/kind feature

Several Kueue CRDs were missing `shortNames`, making them verbose to use with `kubectl`. This PR adds short names to the remaining resources so operators can use concise aliases like `kubectl get ac` or `kubectl get topo`.

  | Resource | Short Name(s) added |
  |---|---|
  | `AdmissionCheck` | `ac` |
  | `Cohort` | `co` |
  | `MultiKueueCluster` | `mkc` |
  | `MultiKueueConfig` | `mkconf` |
  | `ProvisioningRequestConfig` | `prc` |
  | `Topology` | `topo` |
  | `WorkloadPriorityClass` | `wpc` |

  Resources that already had short names (`ClusterQueue`→`cq`, `LocalQueue`→`lq`, `ResourceFlavor`→`rf`, `Workload`→`kwl`) are unchanged.

  Changes are applied to both `v1beta1` and `v1beta2` API type files, the generated CRD YAML under `config/components/crd/bases/`, and the Helm chart CRD templates under `charts/kueue/templates/crd/`.

Fixes #

```release-note
Add kubectl short names for AdmissionCheck (ac), Cohort (co), MultiKueueCluster (mkc), MultiKueueConfig (mkconf), ProvisioningRequestConfig (prc), Topology (topo), and WorkloadPriorityClass (wpc).
```

/cc @mimowo

### DIFF
--- a/apis/kueue/v1beta1/admissioncheck_types.go
+++ b/apis/kueue/v1beta1/admissioncheck_types.go
@@ -105,7 +105,7 @@ const (
 // +kubebuilder:object:root=true
 // +kubebuilder:deprecatedversion:warning="This version is deprecated. Use v1beta2 instead."
 // +kubebuilder:subresource:status
-// +kubebuilder:resource:scope=Cluster
+// +kubebuilder:resource:scope=Cluster,shortName={ac}
 
 // AdmissionCheck is the Schema for the admissionchecks API
 type AdmissionCheck struct {
--- a/apis/kueue/v1beta1/cohort_types.go
+++ b/apis/kueue/v1beta1/cohort_types.go
@@ -82,7 +82,7 @@ type CohortStatus struct {
 // +genclient:nonNamespaced
 // +kubebuilder:object:root=true
 // +kubebuilder:deprecatedversion:warning="This version is deprecated. Use v1beta2 instead."
-// +kubebuilder:resource:scope=Cluster
+// +kubebuilder:resource:scope=Cluster,shortName={co}
 // +kubebuilder:subresource:status
 
 // Cohort defines the Cohorts API.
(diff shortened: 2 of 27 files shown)


mimowo:
LGTM, I'm just wondering if there are any where we can run in conflicts with the core k8s as we did in the past with "wl" so now we rename to "kwl". 

The potentially risky are "topo" if k8s core introduced Topology CRD, and WorkloadPriorityClass if k8s core introduces the WorkloadPriorityClass class. 

I'm not sure if there are such plans, wdyt @tenzen-y @kannon92 @wojtek-t?  

If there s a collision this is not a big deal - we can rename later as we did for "wl" so this is not a big deal, just doing "due diligence".

gyliu513:
Good point, thanks @mimowo for the due diligence! Following the same pattern as wl -> kwl, I prefer rename all short names by adding `k` as prefix, like topo -> ktopo, wpc -> kwpc to avoid potential conflicts if upstream Kubernetes introduces Topology or WorkloadPriorityClass CRDs in the future. 

Comments are welcomed and I can follow up if we come to a conclusion.  @tenzen-y @kannon92 @wojtek-t @mimowo

mimowo:
> I prefer rename all short names by adding k as prefix, like topo -> ktopo, wpc -> kwpc to avoid potential conflicts if upstream Kubernetes introduces Topology or WorkloadPriorityClass CRDs in the future.

I'm not sure about that, we still like to use "cq" and "lq" for ClusterQueue and LocalQueue. It is very unlikely they would be introduced in the core k8s any time soon, so I like the preferrable names for them. 

On top of that, there is no guarantee really k8s core will no have some CRD which resolves to "kcq" so adding "k" proactively may not even benefit us in all cases. I would use the prefix only when needed.

kannon92:
> The potentially risky are "topo" if k8s core introduced Topology CRD, and WorkloadPriorityClass if k8s core introduces the WorkloadPriorityClass class.

I'm not aware of any plans to add CRDs for this.

mimowo:
> I'm not aware of any plans to add CRDs for this.

I see, ok let's go with the proposed prefixes, we can always adjust if needed as we did with "wl -> kwl"

mimowo:
/lgtm
/approve
Thank you for improving the UX for Kueue.

PR #11547, opened 2026-05-22
Promote ElasticJobsViaWorkloadSlices to beta

### LINKED ISSUE #8651: Path to Beta for ElasticJobsViaWorkloadSlices - Ray Autoscaling
A path to graduate the ElasticJobsViaWorkloadSlices feature gate to beta for Ray (RayCluster/RayJob) autoscaling support.

**Why is this needed**:
  We have customer demand for autoscaling Ray workloads managed by Kueue. The Ray-specific implementation has already landed:
  - RayCluster autoscaling: #6662 (v0.14)
  - RayJob autoscaling: #7605 / #8082 (backported to v0.14 via #8229)

 However, this functionality is gated behind ElasticJobsViaWorkloadSlices, which is alpha and disabled by default.

The current https://github.com/kubernetes-sigs/kueue/tree/main/keps/77-dynamically-sized-jobs bundles requirements that span multiple integrations (GC policies, metrics, JobSet, PyTorchJob, etc.). This creates a situation where Ray autoscaling may be ready but blocked waiting on unrelated work.

  I had an initial conversation with @mimowo about this. Some points that came up:

  - The current ElasticWorkloads implementation has minimal validation, meaning users can hit issues with unsupported combinations (e.g., TAS + Elastic) late in the process. For beta, we'd need to either support these combinations or add validation to reject them upfront.
  - One option discussed was introducing scoped feature gates for specific integrations that aren't ready (e.g., `ElasticJobsViaWorkloadSlicesForTAS`), allowing the core functionality to graduate while deferring others.
  - Strengthening validation now is preferred since it can be relaxed later, whereas the reverse is harder.

The suggestion was to revisit after 0.17 to assess progress and decide on the graduation path. Sharing this for context. I would love to hear other perspectives on how to approach this.

Looking for feedback on how we can move this forward. I'm willing to actively work on the remaining pieces once we align on an approach.

**Completion requirements**:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

sohankunkerkar:
#### What type of PR is this?
/kind feature

Fixes https://github.com/kubernetes-sigs/kueue/issues/8651

```release-note
Promote ElasticJobsViaWorkloadSlices feature gate to beta
```

### DIFF
--- a/keps/77-dynamically-sized-jobs/README.md
+++ b/keps/77-dynamically-sized-jobs/README.md
@@ -546,14 +546,14 @@ Here’s a structured and detailed **Graduation Criteria** section for KEP-77: *
 
 #### Beta
 
-* [ ] Feature is enabled by default in Kueue, still guarded by the feature gate for opt-out.
-* [ ] Webhook validation to reject elastic annotation on unsupported frameworks.
-* [ ] Webhook validation to reject PartialAdmission and Elastic on the same job.
+* [x] Feature is enabled by default in Kueue, still guarded by the feature gate for opt-out.
+* [x] Webhook validation to reject elastic annotation on unsupported frameworks.
+* [x] Webhook validation to reject PartialAdmission and Elastic on the same job.
 * [x] Dedicated controller for removing scheduling gates on admitted workload slices.
 * [x] At least one additional framework beyond `batch/v1.Job` integrates the WorkloadSlice flow.
 * [x] Metrics for slice lifecycle transitions (`replaced_workload_slices_total`).
 * [ ] Documentation with examples for users and integrators.
-* [ ] E2e tests for scale-up, scale-down, slice replacement, sticky flavor.
+* [x] E2e tests for scale-up, scale-down, slice replacement, sticky flavor.
 * [x] Old workload slice finished only after replacement slice is admitted.
 
 #### GA (Stable)
--- a/pkg/controller/core/indexer/indexer.go
+++ b/pkg/controller/core/indexer/indexer.go
@@ -261,11 +261,6 @@ func Setup(ctx context.Context, indexer client.FieldIndexer) error {
 		if err := indexer.IndexField(ctx, &corev1.Pod{}, WorkloadSliceNameKey, IndexPodWorkloadSliceName); err != nil {
 			return fmt.Errorf("setting index on workloadSliceName for Pod: %w", err)
 		}
-		// OwnerReferenceUID index for Pod is for backwards compatibility only.
-		// TODO(sohankunkerkar): remove in 0.18
-		if err := indexer.IndexField(ctx, &corev1.Pod{}, OwnerReferenceUID, IndexOwnerUID); err != nil {
-			return fmt.Errorf("setting index on ownerReferences.uid for Pod: %w", err)
-		}
 	}
 	// Index DeviceClasses by extendedResourceName for fast lookup during extended resource translation.
 	// The index is skipped if the DeviceClass API is not available on the cluster.
--- a/pkg/controller/jobframework/validation_test.go
+++ b/pkg/controller/jobframework/validation_test.go
@@ -459,7 +459,8 @@ func TestValidateJobOnCreate(t *testing.T) {
 			job: utiltestingjob.MakeJob("test-job", "ns1").
 				SetAnnotation(workloadslicing.EnabledAnnotationKey, workloadslicing.EnabledAnnotationValue).
 				Obj(),
-			gvk: schema.GroupVersionKind{Group: "jobset.x-k8s.io", Version: "v1alpha2", Kind: "JobSet"},
+			gvk:          schema.GroupVersionKind{Group: "jobset.x-k8s.io", Version: "v1alpha2", Kind: "JobSet"},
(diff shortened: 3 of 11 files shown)


tenzen-y:
> /test pull-kueue-test-e2e-baseline-main-1-33

@sohankunkerkar What was the reason for the failed CI Job?

sohankunkerkar:
> > /test pull-kueue-test-e2e-baseline-main-1-33
> 
> @sohankunkerkar What was the reason for the failed CI Job?

https://github.com/kubernetes-sigs/kueue/pull/11561

This was a flake.

mimowo:
/lgtm
/approve
Thank you for driving this forward and all the recent fixes 👍

mimowo:
/hold
in case @tenzen-y has additional comments

pkg/controller/jobframework/reconciler.go
```diff
@@ -1589,10 +1589,7 @@ func getPodSetsInfoFromStatus(ctx context.Context, c client.Client, w *kueue.Wor
 		if features.Enabled(features.TopologyAwareScheduling) {
 			info.Annotations[kueue.WorkloadAnnotation] = w.Name
 		}
-		// Set workload slice name annotation on pods for elastic workloads.
-		// This enables pod lookup by annotation instead of owner reference,
-		// supporting JobSet and other workloads where pods are not immediate children.
-		if features.Enabled(features.ElasticJobsViaWorkloadSlices) {
+		if workloadslicing.IsElasticWorkload(w) {
 			info.Annotations[kueue.WorkloadSliceNameAnnotation] = workloadslicing.SliceName(w)
 		}
```
tenzen-y: @sohankunkerkar Is this the reason that we want to add workload slice name annotations only when the workload has `kueue.x-k8s.io/elastic-job: true`  annotations, right?

If yes, isn't this a bug in the alpha phase? Because in the previous alpha stage, Elastic Jobs, workload slice name annotation is added even when the workload doesn't have `kueue.x-k8s.io/elastic-job: true`.

Could we decouple this bug fix from this PR, then create a separate PR for the back ports?

https://github.com/kubernetes-sigs/kueue/blob/e2004ca27b7525308157862333fe64e4d65c114d/pkg/workloadslicing/workloadslicing.go#L60-L76

tenzen-y (commented):
@sohankunkerkar Thank you for moving this forward!
I left a small checking comments.

pkg/controller/jobframework/reconciler.go
mimowo: Yes, I think this is a bug in the alpha phase.  Otoh, this bug would only affect users who used the alpha feature, and also the bug is not disruptive, just polluting workloads with extra information.

Ideally we de-couple any fix from the actual graduation, so +1 for spotting this one.

tenzen-y: Yes, I don't think that this is a blocker for beta graduation.
I just hope to decouple this fix from the beta graduation PR.

I believe that we still can graduate ElasticJobsViaWorkloadSlices to beta in Kueue 0.18.

tenzen-y:
@sohankunkerkar can you also update https://github.com/kubernetes-sigs/kueue/blob/8078525295edb6717ccbfa9343c06834eb675a3b/site/content/en/docs/concepts/elastic_workload.md?plain=1#L9?

Additionally, what about https://github.com/kubernetes-sigs/kueue/blob/8078525295edb6717ccbfa9343c06834eb675a3b/pkg/workloadslicing/workloadslicing.go#L343-L345 and https://github.com/kubernetes-sigs/kueue/blob/8078525295edb6717ccbfa9343c06834eb675a3b/pkg/controller/core/indexer/indexer.go#L264-L265

pkg/controller/jobframework/reconciler.go
sohankunkerkar: Yup, makes sense. Addressed this concern here: https://github.com/kubernetes-sigs/kueue/pull/11620

sohankunkerkar:
> @sohankunkerkar can you also update
> 
> https://github.com/kubernetes-sigs/kueue/blob/8078525295edb6717ccbfa9343c06834eb675a3b/site/content/en/docs/concepts/elastic_workload.md?plain=1#L9
> 
> ?
> Additionally, what about
> 
> https://github.com/kubernetes-sigs/kueue/blob/8078525295edb6717ccbfa9343c06834eb675a3b/pkg/workloadslicing/workloadslicing.go#L343-L345
> 
> and
> https://github.com/kubernetes-sigs/kueue/blob/8078525295edb6717ccbfa9343c06834eb675a3b/pkg/controller/core/indexer/indexer.go#L264-L265

Good call, The owner reference fallback in `StartWorkloadSlicePods` was added for backwards compatibility with pods created before the `WorkloadSliceNameAnnotation` was introduced in v0.16.0. Since the minimum supported  upgrade path to v0.18 is from v0.16+, all elastic workload pods in supported configurations already have the annotation. The fallback and its Pod-specific OwnerReferenceUID index are dead code and can be  safely removed.

mimowo:
/lgtm 
/unhold
This is now good to go as the dependency PR is merged

tenzen-y (commented):
Look great! Thank you!

/lgtm
/approve

PR #11566, opened 2026-05-23
fix: prevent crash when error is not a string in ErrorMessage

### LINKED ISSUE #11565: ErrorMessage.jsx crashes when error is not a string
I was going through the kueueviz frontend code and tested a few error states. When a WebSocket error event gets passed to ErrorMessage.jsx the whole component crashes. I looked into it and the component is calling error.split('\n') directly which only works if error is a plain string. If anything else gets passed in like an actual Error object or a WebSocket event, split is not a function on it and it throws.
I think the fix is to check what type error is first and convert it to a string before calling split. I would like to pick this up if no one else is working on it.

YadavAkhileshh:
Closes #11565 

Currently, `ErrorMessage.jsx` calls `error.split('\n')` assuming that `error` is always a primitive string. 

If the component receives an actual `Error` object or a standard WebSocket event (which is common during connection disruptions), it throws `TypeError: error.split is not a function` and crashes the entire error display container.

This PR adds type-checking and safe string coercion (`error.message || JSON.stringify(error)`) to convert non-string errors to strings safely before splitting.

```release-note
KueueViz: Fixed frontend crashes caused by non-string errors in the error display.
```

### DIFF
--- a/cmd/kueueviz/frontend/src/ErrorMessage.jsx
+++ b/cmd/kueueviz/frontend/src/ErrorMessage.jsx
@@ -20,13 +20,23 @@ import KeyboardArrowDownIcon from '@mui/icons-material/KeyboardArrowDown';
 import KeyboardArrowUpIcon from '@mui/icons-material/KeyboardArrowUp';
 import './App.css';
 
+/**
+ * Converts an error of any type to a displayable string.
+ * Handles plain strings, Error objects, and arbitrary objects/events.
+ */
+export const errorToString = (error) => {
+  if (typeof error === 'string') return error;
+  if (error && typeof error.message === 'string' && error.message.length > 0) return error.message;
+  return JSON.stringify(error);
+};
+
 const ErrorMessage = ({ error }) => {
   const [expanded, setExpanded] = useState(false);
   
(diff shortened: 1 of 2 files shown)


mimowo:
/release-note-edit
```release-note
KueueViz: Fixed frontend crashes caused by non-string errors in the error display.
```

cmd/kueueviz/frontend/src/ErrorMessage.jsx
```diff
@@ -26,7 +26,8 @@ const ErrorMessage = ({ error }) => {
   if (!error) return null;
 
   // Extract the first line as the summary
-  const lines = error.split('\n');
+  const errorString = typeof error === 'string' ? error : (error.message || JSON.stringify(error));
+  const lines = errorString.split('\n');
```
mimowo: Could you add some unit test for it?

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16
Thank you for hardening KueueViz 👍

mimowo:
The 0.16 branch requires manual prep

PR #11568, opened 2026-05-23
Allow quota reserving workloads to be filtered directly in the kube API server

### LINKED ISSUE #11567: Support server-side filtering of Workloads by status using kubectl field selectors

The ability to filter Workloads server-side using kubectl --field-selector. For example, listing all workloads that currently have quota reserved, or all workloads reserved in a specific ClusterQueue, without fetching every workload and filtering client-side.

**Why is this needed**:

- Today, filtering workloads by status requires either `kueuectl list workload --status` (which fetches all workloads then filters client-side). There is no way to use native `kubectl --field-selector` on Workload beyond `metadata.name` and `metadata.namespace`.
- At scale, listing all workloads and discarding most of them is inefficient. Server-side filtering reduces API traffic and response size.
- kubectl is the standard tool for interacting with Kubernetes resources. Operators and tooling (dashboards, scripts, monitoring) should be able to query workloads by reservation or admission status using standard Kubernetes mechanisms.

**Completion requirements**:

- [ ] Design doc
- [ ] API change
- [x] Docs update

mukund-wayve:
#### What type of PR is this?

/kind feature

Adds `status.admission.clusterQueue` as a selectable field on the Workload CRD (v1beta2), so users can filter workloads server-side with `--field-selector`.

This allows users to efficiently query QuotaReserving workloads by ClusterQueue:
- Filter for a specific CQ: `kubectl get workloads --field-selector status.admission.clusterQueue=my-cq`
- Filter across all CQs: `kubectl get workloads --field-selector status.admission.clusterQueue!=`

Only added to `v1beta2` because when both versions have the same `selectableFields`, Kubernetes moves them to the top-level CRD spec, which breaks validation when the per-version schemas differ.

This also works with `kueuectl` since it passes through the field selectors to kubectl. Example command: `kubectl kueue list workloads --field-selector status.admission.clusterQueue=my-cq`

Partially Fixes #11567 

This only partially fixes the issue since it allows filtering for a subset of statuses. The ideal fix IMO would be to add a `status.phase` field in Workload CRD and expose it as a field selector, similar to how pods work. But that's a lot more work (new field, controller changes, likely a KEP). This seemed like a quick win — at least for filtering all quota-reserving and admitted workloads which is my immediate usecase.

For context: workload status today is derived from `status.conditions` by a `workload.Status()` helper. kueuectl uses this client-side for its `--status` flag, but `status.conditions` can't be used as a CRD field selector.

Needs `CustomResourceFieldSelectors` feature gate (on by default since K8s 1.30, GA since 1.31).

```release-note
Workload CRD now supports `--field-selector status.admission.clusterQueue=<name>` to filter by ClusterQueue server-side. `status.admission.clusterQueue` is set when a workload reserves quota in the CQ.
```

### DIFF
--- a/apis/kueue/v1beta2/workload_types.go
+++ b/apis/kueue/v1beta2/workload_types.go
@@ -1113,6 +1113,7 @@ const (
 // +kubebuilder:printcolumn:name="Finished",JSONPath=".status.conditions[?(@.type=='Finished')].status",type="string",description="Workload finished"
 // +kubebuilder:printcolumn:name="Age",JSONPath=".metadata.creationTimestamp",type="date",description="Time this workload was created"
 // +kubebuilder:resource:shortName={kwl,kueueworkload,kueueworkloads}
+// +kubebuilder:selectablefield:JSONPath=.status.admission.clusterQueue
 
 // Workload is the Schema for the workloads API
 // +kubebuilder:validation:XValidation:rule="has(self.status) && has(self.status.conditions) && self.status.conditions.exists(c, c.type == 'QuotaReserved' && c.status == 'True') && has(self.status.admission) ? size(self.spec.podSets) == size(self.status.admission.podSetAssignments) : true", message="podSetAssignments must have the same number of podSets as the spec"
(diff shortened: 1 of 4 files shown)


kannon92:
> This only partially fixes the issue since it allows filtering for a subset of statuses. The ideal fix IMO would be to add a status.phase field in Workload CRD and expose it as a field selector, similar to how pods work. But that's a lot more work (new field, controller changes, likely a KEP). This seemed like a quick win — at least for filtering all quota-reserving and admitted workloads which is my immediate usecase.

Phase is considered an antipattern in Kubernetes api-conventions.

> Some resources in the v1 API contain fields called phase, and associated message, reason, and other status fields. The pattern of using phase is deprecated. Newer API types should use conditions instead. Phase was essentially a state-machine enumeration field, that contradicted [system-design principles](https://git.k8s.io/design-proposals-archive/architecture/principles.md#control-logic) and hampered evolution, since [adding new enum values breaks backward compatibility](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api_changes.md). Rather than encouraging clients to infer implicit properties from phases, we prefer to explicitly expose the individual conditions that clients need to monitor. Conditions also have the benefit that it is possible to create some conditions with uniform meaning across all resource types, while still exposing others that are unique to specific resource types. See [#7856](http://issues.k8s.io/7856) for more details and discussion.

https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md

mimowo (commented):
/lgtm
/approve
Thank you for the useful feature 👍
Would you like to maybe follow up with documentation update on how to use this? I think that could go to https://kueue.sigs.k8s.io/docs/tasks/troubleshooting/troubleshooting_jobs/, wdyt?

Issue #11055, opened 2026-05-08
AgentSkills: extend the flake debugger skill with the ability to compare the failed run with successful


mimowo:
**What would you like to be added**:

The skill should have a step to compare the failed run with successful, as proposed in https://github.com/kubernetes-sigs/kueue/issues/11029#issuecomment-4405761464

The skill is added as part of https://github.com/kubernetes-sigs/kueue/pull/11054

**Why is this needed**:

This is what developers often do to root cause some difficult cases.

andrewseif:
I would like to work on this ticket.

What would be an acceptable AC?
Maybe I should take a flake test and have the agent fix it using the new skill, would that be acceptable? (ofc ill review the code myself 😄)

/assign

PR #11569, opened 2026-05-24
[Fix][Flaky] Scheduler when Reserving Resources in Cohort foundation

### LINKED ISSUE #10363: [Flaky] Scheduler when Reserving Resources in Cohort foundation-queue blocks capacity when it is not sure that it can reclaim
**Which test is flaking?**:
Scheduler Suite: [It] Scheduler when Reserving Resources in Cohort foundation-queue blocks capacity when it is not sure that it can reclaim 

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10348/pull-kueue-test-integration-baseline-release-0-17/2042478052260188160

**Failure message or logs**:
```
{ failed [FAILED] Timed out after 10.000s.
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/kueue/test/util/metrics.go:299 with:
pending_workloads with status=active
Expected
    <float64>: 0
to equal
    <float64>: 2
In [It] at: /home/prow/go/src/sigs.k8s.io/kueue/test/integration/singlecluster/scheduler/scheduler_test.go:2784 @ 04/10/26 06:03:25.449
}
```

andrewseif:
#### What type of PR is this?
/kind bug
/kind flake

Fix Scheduler Suite: [It] Scheduler when Reserving Resources in Cohort foundation-queue blocks capacity when it is not sure that it can reclaim

Fixes #10363

I have used an Agent to fix this flaky test and perform thorough analysis, as part of the AC of this issue #11055, ill attach the investigation.md to the comments below

```release-note
NONE
```

### DIFF
--- a/test/integration/singlecluster/scheduler/scheduler_test.go
+++ b/test/integration/singlecluster/scheduler/scheduler_test.go
@@ -2852,7 +2852,7 @@ var _ = ginkgo.Describe("Scheduler", func() {
 
 			ginkgo.By("best-effort workloads not admitted as foundation workload has priority")
 			util.ExpectReservingActiveWorkloadsMetric(cq2, 0)
-			util.ExpectPendingWorkloadsMetric(cq2, 2, 0)
+			util.ExpectPendingWorkloadsMetric(cq2, 0, 2)
 
 			ginkgo.By("second foundation workload admitted")
 			util.ExpectReservingActiveWorkloadsMetric(cq1, 1)


andrewseif:
Here is the analysis/investigation. 
It is worth noting that I also asked the agent about the "intent" of the original test writer, incase they were trying to check the first cycle before scheduler runs where both workloads are still considered active before moving to Inadmissible.

[10363-analysis.md](https://github.com/user-attachments/files/28186716/10363-analysis.md)

andrewseif:
I was thinking if we want to standardize an analysis template, so that it can be used, and increase determinism, when the agents are doing the analysis.

What do you guys think? I followed the one suggested [here(comment)](https://github.com/kubernetes-sigs/kueue/issues/11029#issuecomment-4405761464)

test/integration/singlecluster/scheduler/scheduler_test.go
```diff
@@ -2852,7 +2852,7 @@ var _ = ginkgo.Describe("Scheduler", func() {
 
 			ginkgo.By("best-effort workloads not admitted as foundation workload has priority")
 			util.ExpectReservingActiveWorkloadsMetric(cq2, 0)
-			util.ExpectPendingWorkloadsMetric(cq2, 2, 0)
+			util.ExpectPendingWorkloadsMetric(cq2, 0, 2)
```
mimowo, line 2855: The section "Why the test was failing (plain version)" from the analysis makes sense to me, in particular:

The test asserted `(active=2, inadmissible=0)`. That state exists for a few milliseconds — the brief window right after the workloads are created, before the scheduler has run a cycle on them alone. As soon as the scheduler evaluates them with no other workload competing, it correctly says "no fit" and moves them to the **inadmissible** bucket. They sit there for the rest of the test.

However,  I think the same is true for the assert in L2848, could you check that one too?

andrewseif: **Why isn't the same change needed at line 2848?**

Both assertions look identical, so it's natural to suspect both are flaky.But, because foundation-queue's heap is in a different state at each call site, and that changes the scheduler path best-effort workloads take.

**Line 2848  foundation has a stuck head, so cycles are multi-entry**

At this point, `workload-7f9t2` (the second foundation workload, 2 CPU) is queued and cannot admit (cohort has only 1 CPU free). foundation-queue is `StrictFIFO`, and the StrictFIFO requeue rule in [`pkg/cache/queue/cluster_queue.go`](pkg/cache/queue/cluster_queue.go#L870-L871) is:

```go
if c.queueingStrategy == kueue.StrictFIFO {
    immediate = reason != RequeueReasonNamespaceMismatch
}
```

Any other reason (including the Generic reason from NoFit) is `immediate=true`, so `workload-7f9t2` goes straight back into foundation's heap every cycle.

Because foundation has a head every cycle, `heads()` returns `count=2` and the scheduler runs a **multi-entry** cycle. When best-effort is processed *after* foundation, it goes through the "skipped" path (`"Workload no longer fits after processing another workload"`, status=`skipped`). In `requeueAndUpdate` ([`pkg/scheduler/scheduler.go`](pkg/scheduler/scheduler.go#L947-L949)):

```go
if e.status != notNominated && e.requeueReason == qcache.RequeueReasonGeneric {
    e.requeueReason = qcache.RequeueReasonFailedAfterNomination
}
```

`skipped != notNominated`, so the reason is upgraded to `FailedAfterNomination`, which **is** in BestEffortFIFO's immediate set ([`pkg/cache/queue/cluster_queue.go`](pkg/cache/queue/cluster_queue.go#L873-L876)) → workload stays in the heap → `(active=2, inadmissible=0)` is stable.

**Line 2855 foundation has admitted, so cycles are single-entry**

By this point, `finish first foundation workload` has run and the second foundation admits. foundation-queue's heap is empty. `heads()` returns `count=1` (best-effort only). The single-entry path goes through `FlavorAssigner assigned NoFit mode` at [`pkg/scheduler/scheduler.go:302`](pkg/scheduler/scheduler.go#L302), which sets `status=""` (notNominated). The `requeueReason` stays at `Generic` — and Generic is **not** in BestEffortFIFO's immediate set → workload goes to `inadmissibleWorkloads` → metric reports `(active=0, inadmissible=2)`.

**Side-by-side**

| | Line 2848 (foundation queued) | Line 2855 (foundation admitted) |
|---|---|---|
| foundation in heap? | yes (StrictFIFO + Generic → immediate) | no |
| cycle entries | 2 | 1 |
| best-effort scheduler path | "skipped" (after another workload) | "notNominated" (FlavorAssigner NoFit) |
| requeueReason | `FailedAfterNomination` | `Generic` |
| BestEffortFIFO treatment | immediate → heap | not immediate → `inadmissibleWorkloads` |
| stable state | `(2, 0)` (this) | `(0, 2)` |

**Empirical confirmation from the failing run** ([Prow #2042478052260188160](https://gcsweb.k8s.io/gcs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10348/pull-kueue-test-integration-baseline-release-0-17/2042478052260188160/build-log.txt)): line 2848 *passed* — the test successfully reached `STEP: finish first foundation workload @ 06:03:15.372` and `STEP: best-effort workloads not admitted @ 06:03:15.447`. Cycles 201–206 in that window were all `entries=2` with best-effort taking the skipped → FailedAfterNomination → heap path. The failure is only at line 2855, where cycle 207 was the first `entries=1` cycle and immediately put workload-7wbbt into `inadmissibleWorkloads`.

So line 2848 is correct as-is; the bug really is one literal at line 2855.

mimowo: This is actually correct

mimowo:
/lgtm 
/approve
/cherrypick release-0.17
/cherrypick release-0.16
Thanks for helping to stabilize the build

PR #11571, opened 2026-05-24
hack/testing: skip install-time operator deployment waits by default

ekam-walia:
What type of PR is this?

/kind cleanup
/area testing

Skip the blocking kubectl wait --for=condition=available after each external operator install. These waits serialize setup unnecessarily — BeforeSuite already gates on operator readiness before any test runs.

The waits can be re-enabled via E2E_WAIT_FOR_OPERATORS_IN_INSTALL=true.
The Kubeflow Trainer webhook wait is intentionally kept unconditional, as the runtime resources applied immediately after depend on it.

Which issue(s) this PR fixes:
Fixes #

Only install_kubeflow_trainer retains a plain kubectl wait — it gates the overlays/runtimes apply that follows it.

```release-note
NONE
```

### DIFF
--- a/hack/testing/e2e-common.sh
+++ b/hack/testing/e2e-common.sh
@@ -103,6 +103,21 @@ function e2e_docker_pull_if_needed {
     return 1
 }
 
+function e2e_wait_for_operator_in_install {
+  local kubeconfig=$1
+  local ns=$2
+  local deployment_name=$3
+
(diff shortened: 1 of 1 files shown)


mimowo:
/test all
double checking stability before merge

mimowo:
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16
Thank you, let's see if this can help with the build times realistically 👍 The helper function is useful anyway to avoid duplication.

PR #11575, opened 2026-05-24
fix: wrap JSON.parse in try-catch in useWebSocket

### LINKED ISSUE #11574: useWebSocket.js crashes when server sends non-JSON message
I was looking at useWebSocket.js and noticed the onmessage handler calls JSON.parse(event.data) with nothing around it. I tested this by sending a plain text message and the UI just froze completely. If the server sends anything that is not valid JSON like a ping or an empty frame it throws and takes down the whole UI with it.
I think wrapping it in a try catch and setting an error state in the catch block would fix this. Happy to open a PR if this is not being worked on.

YadavAkhileshh:
Closes #11574

The `onmessage` handler in `useWebSocket.js` calls `JSON.parse(event.data)` without any error handling. If the server sends a non-JSON frame (plain text, empty message, heartbeat ping), the parse throws an unhandled exception that freezes the entire UI.

Wraps the call in a try-catch block and sets an error state on parse failure so the UI can display it gracefully instead of crashing.

```release-note
KueueViz: Fixed a bug where non-JSON WebSocket messages from the backend could crash
the frontend. KueueViz now reports such messages as errors instead of freezing the UI.
```

### DIFF
--- a/cmd/kueueviz/frontend/src/useWebSocket.js
+++ b/cmd/kueueviz/frontend/src/useWebSocket.js
@@ -43,8 +43,13 @@ const useWebSocket = (url) => {
     };
 
     ws.onmessage = (event) => {
-      const message = JSON.parse(event.data);
-      setData(message);
+      try {
+        const message = JSON.parse(event.data);
+        setData(message);
+      } catch (err) {
+        console.error('Failed to parse WebSocket message:', err);
+        setError('Received malformed data from the server.');
+      }
     };
 
     ws.onerror = (err) => {
--- a/cmd/kueueviz/frontend/src/useWebSocket.js
+++ b/cmd/kueueviz/frontend/src/useWebSocket.js
@@ -25,6 +25,21 @@ const WS_TOKEN_PROTOCOL_PREFIX = 'kueueviz.auth.';
 const encodeTokenForProtocol = (token) =>
   btoa(token).replace(/\+/g, '-').replace(/\//g, '_').replace(/=+$/g, '');
 
+/**
+ * Parses a raw WebSocket message string as JSON.
+ * Returns { data, error } so callers can handle failures gracefully.
+ */
+export const parseWebSocketMessage = (raw) => {
+  try {
+    if (typeof raw !== 'string') {
+      throw new TypeError('expected string');
+    }
+    return { data: JSON.parse(raw), error: null };
+  } catch {
+    return { data: null, error: 'Received malformed data from the server.' };
+  }
+};
+
 const useWebSocket = (url) => {
   const [data, setData] = useState(null);
   const [error, setError] = useState(null);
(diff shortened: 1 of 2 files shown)


kannon92:
Please follow the pull request template.

mimowo:
/release-note-edit
```release-note
KueueViz: Fixed a bug where non-JSON WebSocket messages from the backend could crash
the frontend. KueueViz now reports such messages as errors instead of freezing the UI.
```

mimowo:
injected the release note block manually, please follow the PR template next time

cmd/kueueviz/frontend/src/useWebSocket.js
```diff
@@ -43,8 +43,13 @@ const useWebSocket = (url) => {
     };
 
     ws.onmessage = (event) => {
-      const message = JSON.parse(event.data);
-      setData(message);
+      try {
```
mimowo: Can we add some unit test for it?

mimowo:
@YadavAkhileshh do you have cycles to conclude the PR with unit tests?
Same question for https://github.com/kubernetes-sigs/kueue/pull/11566

YadavAkhileshh:
@mimowo thanks for the review feedback!

I've pushed a follow-up commit that:

- Extracts the JSON parsing into a standalone `parseWebSocketMessage()` function exported from `useWebSocket.js`, keeping it pure and testable without needing browser globals or React context.
- Adds `useWebSocket.test.js` with 10 unit tests covering valid JSON, malformed messages, empty strings, truncated payloads, and non-string inputs.

All 38 tests pass (28 existing + 10 new) and the build is clean.

I'll also add tests for #11566 shortly working on it now.

mimowo:
@kannon92 it seems the job  `pull-kueue-test-e2e-k8s-main-was` is running on the PR which didn't opted-in for it, should we temporarily revert?

kannon92:
> @kannon92 it seems the job `pull-kueue-test-e2e-k8s-main-was` is running on the PR which didn't opted-in for it, should we temporarily revert

Hmm, we don't need to revert as its not a required job. This will not block merges.

But we can revert if you want.

kannon92:
https://github.com/kubernetes/test-infra/pull/37106

This should fix that the pull-kueue-test-e2e-k8s-main-was does not run. 

You can approve/lgtm this PR and it should merge anyway.

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16
Thank you for hardening KueueViz 👍

mimowo:
> Hmm, we don't need to revert as its not a required job. This will not block merges.

I see, I didn't know it is non blocking for merge - good to know. 

At the same time, it is a resource intensive Job which will run in parallel with other precious CI jobs. This is slightly reducing the pool of machines which can be used at the same time for required CI jobs - I don't know how much buffer we have. I think ideally if this job was run on PRs "on-demand" - I think for that we should just comment out skip_if_only_changed.

I will open an issue for that: https://github.com/kubernetes-sigs/kueue/issues/11664

mimowo:
@YadavAkhileshh would you be able to prepare the cherrypick for 0.16?

PR #11576, opened 2026-05-25
Experiment: test kueue on 1.36

mimowo:
#### What type of PR is this?

/kind cleanup

Experiment for https://github.com/kubernetes-sigs/kueue/pull/10936

```release-note
NONE
```


mimowo:
Cc @kannon92 the multikueue e2e tests are failing because the change in core k8s is breaking. We can workaround in kueue but we should also fix upstream

mimowo:
old branch

PR #11598, opened 2026-05-25
MultiKueue: generalize nomination skip during eviction

### LINKED ISSUE #11452: MultiKueue: generalize skipping cluster nomination for all dispatchers while eviction is ongoing

When eviction is ongoing, ie. status.clusterName is still assigned (and the status.admissionCheck.state=Retry) we should skip the cluster nomination phase completely.

**Why is this needed**:

* Not skipping the nomination leads to subtle race conditions as this one fixed here: https://github.com/kubernetes-sigs/kueue/pull/11378
* currently different dispatchers need to re-implement the skip by themselves which is code duplication, and also requires non-obvious checks from the authors of external disaptchers, 
- [AllAtOnce](https://github.com/kubernetes-sigs/kueue/pull/11378/changes#diff-a09b1d2caac9c60d2112a557a72bb2495f52f4c9ac6d161c3b30f1a3a6be24b6R771-R773)
- [Incremental](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/workloaddispatcher/incrementaldispatcher.go#L103-L112)

I think this code should be moved into workload go and skip the nomination phase (copied from the incremental dispatcher):

```golang
	if mkAc == nil || mkAc.State != kueue.CheckStatePending {
		log.V(3).Info("AdmissionCheckState is not in Pending, skip the reconciliation")
	}

	// The workload is already assigned to a cluster, no need to nominate workers.
	if wl.Status.ClusterName != nil {
		log.V(3).Info("The workload is already assigned to a cluster, no need to nominate workers")
	}
```

vamsikrishna-siddu:
#### What type of PR is this?
/kind cleanup
/area multikueue

Centralizes the cluster nomination skip logic into a shared `ShouldSkipClusterNomination` helper in `pkg/workload/workload.go`, called from `reconcileGroup` so all dispatchers (AllAtOnce, Incremental,
External) benefit without duplicating the check. Previously each dispatcher had to independently guard against nominating
clusters while eviction is ongoing (`ClusterName` still assigned, admission check not yet `Pending`). This was error-prone and required non-obvious checks from external dispatcher authors.

Fixes #11452

- The skip gate is placed **after** preemption-gate handling in `reconcileGroup`, so preemption orchestration is not affected.
- The stale-cache guard removed from `nominateAndSynchronizeWorkers` (AllAtOnce path) is now covered by the common check upstream.
- The Incremental dispatcher retains its own call to the shared helper since it is a separate controller.

```release-note
     NONE

```

### DIFF
--- a/pkg/controller/admissionchecks/multikueue/workload.go
+++ b/pkg/controller/admissionchecks/multikueue/workload.go
@@ -529,6 +529,13 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
 		requeueAfterSynchronize = requeueIn
 	}
 
+	// 8. Skip nomination while eviction is still in progress or the
+	// admission check is not yet Pending.
+	if workload.ShouldSkipClusterNomination(acs, group.local, group.IsElasticWorkload()) {
+		log.V(3).Info("Skipping cluster nomination phase")
+		return reconcile.Result{RequeueAfter: requeueAfterSynchronize}, nil
+	}
+
 	res, err := w.nominateAndSynchronizeWorkers(ctx, group)
 	if err == nil && (res.RequeueAfter == 0 || requeueAfterSynchronize < res.RequeueAfter) {
 		res.RequeueAfter = requeueAfterSynchronize
@@ -773,12 +780,6 @@ func (w *wlReconciler) nominateAndSynchronizeWorkers(ctx context.Context, group 
 		}
 
 		if !nominatedClusterSetsEqual(group.local.Status.NominatedClusterNames, nominatedWorkers) {
-			// ClusterName != nil indicates possibly stale cache (eviction just cleared ClusterName
-			// but the informer hasn't caught up yet). Avoid creating remote workloads without a
-			// confirmed nomination — wait for the cache to sync.
-			if group.local.Status.ClusterName != nil {
-				return reconcile.Result{}, nil
-			}
 			if err := workload.PatchAdmissionStatus(ctx, w.client, group.local, w.clock, func(wl *kueue.Workload) (bool, error) {
 				wl.Status.NominatedClusterNames = nominatedWorkers
 				return true, nil
--- a/pkg/workload/workload.go
+++ b/pkg/workload/workload.go
@@ -1901,6 +1901,22 @@ func ClusterName(wl *kueue.Workload) string {
 	return ptr.Deref(wl.Status.ClusterName, "")
 }
 
+// ShouldSkipClusterNomination returns true if cluster nomination should be
+// skipped. This covers the case when eviction is ongoing (ClusterName is still
+// assigned while the admission check transitions through Retry), as well as
+// any other state where the admission check is not Pending.
+// Elastic workloads are exempt from the stale-ClusterName check because they
+// intentionally set ClusterName on new slices to target the same worker cluster.
+func ShouldSkipClusterNomination(acs *kueue.AdmissionCheckState, wl *kueue.Workload, isElastic bool) bool {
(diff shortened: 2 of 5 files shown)


mimowo:
/ok-to-test
nice one

vamsikrishna-siddu:
Hi @mimowo 
The initial push had the ClusterName check centralized in reconcileGroup without an elastic workload exemption, which caused the `MultiKueue Should run an ElasticJob on worker if admitted ` integration test to fail.

The issue is that ClusterName != nil means two different things:
Stale cache during eviction: ClusterName hasn't been cleared yet. nomination should be skipped
Elastic workload scale-up: ClusterName is intentionally set on the new slice so it targets the same worker cluster .nomination should proceed
To fully centralize the skip while handling both cases, I added an IsElasticWorkload() exemption to the ClusterName check in reconcileGroup. The AllAtOnce-specific guard in nominateAndSynchronizeWorkers is now removed.both checks (AC state and ClusterName) are fully centralized.

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/11598/pull-kueue-test-integration-multikueue-main/2058899988095700992

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -529,6 +529,20 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
...
+		log.V(3).Info("Skipping cluster nomination phase")
+		return reconcile.Result{RequeueAfter: requeueAfterSynchronize}, nil
+	}
+	// Stale cache: ClusterName still set after eviction but informer hasn't
+	// caught up. Elastic workloads are exempt because they intentionally set
+	// ClusterName on new slices to target the same worker cluster.
+	if group.local.Status.ClusterName != nil && !group.IsElasticWorkload() {
+		log.V(3).Info("Skipping cluster nomination phase, ClusterName still set")
+		return reconcile.Result{RequeueAfter: requeueAfterSynchronize}, nil
+	}
```
mimowo: Could we roll this logic under ShouldSkipClusterNomination, and call the function here?

vamsikrishna-siddu: done.

pkg/workload/workload.go
```diff
@@ -1901,6 +1901,19 @@ func ClusterName(wl *kueue.Workload) string {
...
+// skipped. This covers the case when eviction is ongoing (ClusterName is still
+// assigned while the admission check transitions through Retry), as well as
+// any other state where the admission check is not Pending.
+// Elastic workloads are exempt from the stale-ClusterName check because they
+// intentionally set ClusterName on new slices to target the same worker cluster.
+func ShouldSkipClusterNomination(acs *kueue.AdmissionCheckState, wl *kueue.Workload, isElastic bool) bool {
+	if acs == nil || acs.State != kueue.CheckStatePending {
+		return true
+	}
+	return wl.Status.ClusterName != nil && !isElastic
```
Singularity23x0: Nit: Separating isElastic check would improve redability of the method:

```go
    if isElastic {
		return false
    }
	return wl.Status.ClusterName != nil
```

Singularity23x0 (commented):
/lgtm

pkg/workload/workload.go
vamsikrishna-siddu: done. @Singularity23x0  can you please review it once?

mimowo:
/lgtm
/approve
Thank you for commonizing the code, that will make the code easier to maintain in the long run 👍

PR #11621, opened 2026-05-25
fix: skip markPodsFailed for unhealthy TAS workloads pending second-pass scheduling

### LINKED ISSUE #11617: Flake: Hotswap for Topology Aware Scheduling when Creating a Job with slices with TASReplaceNodeOnNodeTaints enabled Should evict the workload if replacement is not possible due to taint
**Which test is flaking?**:

Hotswap for Topology Aware Scheduling when Creating a Job with slices with TASReplaceNodeOnNodeTaints enabled Should evict the workload if replacement is not possible due to taint 

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/11615/pull-kueue-test-e2e-tas-baseline-main/2058968675913830400

**Failure message or logs**:
```
End To End TAS Baseline Suite: kueue/kind-node:v1.36.1: [It] Hotswap for Topology Aware Scheduling when Creating a Job with slices with TASReplaceNodeOnNodeTaints enabled Should evict the workload if replacement is not possible due to taint expand_less	56s
{ failed [FAILED] Timed out after 45.000s.
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/tas/baseline/hotswap_test.go:548 with:
Expected
    <[]string | len:2, cap:2>: ["kind-worker", "kind-worker2"]
to be equivalent to
    <[]string | len:2, cap:2>: ["kind-worker3", "kind-worker4"]
In [It] at: /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/tas/baseline/hotswap_test.go:200 @ 05/25/26 17:59:28.742
}
```

gyliu513:
#### What type of PR is this?

/kind bug
/kind failing-test
/kind flake
/area tas
/area testing

Fixes a race in TAS node replacement when `TASReplaceNodeOnNodeTaints` is enabled.

When a node is tainted and a pod is replaced, the node controller may call `markPodsFailed` on the pending replacement pod (old node selector, unschedulable) on a later reconcile, after the node is already listed in `status.unhealthyNodes`. That consumes the Job's `BackoffLimit` before second-pass scheduling can evict
the workload and reschedule it (e.g. to another rack).

This change skips `markPodsFailed` only when the unhealthy node was recorded in a prior reconcile **and** the workload still has progressing pods on other nodes. All other cases (including all pods still pending on a failed node) are unchanged.

Repro: e2e `Should evict the workload if replacement is not possible due to taint` (flaky in #11617).

Fixes #11617

/cc @mimowo 

```release-note
NONE
```

### DIFF
--- a/pkg/controller/tas/node_controller.go
+++ b/pkg/controller/tas/node_controller.go
@@ -557,6 +557,12 @@ func (r *nodeReconciler) reconcileWorkloadsOnNode(
 	unhealthyWorkloads := sets.New[types.NamespacedName]()
 	notUnhealthyWorkloads := sets.New[types.NamespacedName]()
 	podsToTerminateMap := make(map[types.NamespacedName][]*corev1.Pod)
+	// pendingSecondPassWorkloads tracks unhealthy workloads where this node was
+	// already recorded in a prior reconcile and other pods are still progressing
+	// on different nodes. In that case second-pass scheduling is in flight, so we
+	// must not fail retry pods on the unhealthy node — doing so would consume
+	// BackoffLimit before the second-pass eviction runs.
+	pendingSecondPassWorkloads := sets.New[types.NamespacedName]()
 
 	for wlKey := range allTASWorkloads {
 		var wl kueue.Workload
@@ -572,6 +578,15 @@ func (r *nodeReconciler) reconcileWorkloadsOnNode(
 		switch result.status {
 		case workloadUnhealthy:
 			unhealthyWorkloads.Insert(wlKey)
+			if workload.HasUnhealthyNode(&wl, nodeName) {
+				progressingElsewhere, err := r.hasProgressingPodsOnOtherNodes(ctx, &wl, nodeName)
+				if err != nil {
+					return ctrl.Result{}, err
+				}
+				if progressingElsewhere {
+					pendingSecondPassWorkloads.Insert(wlKey)
+				}
+			}
 		case workloadHealthy:
 			notUnhealthyWorkloads.Insert(wlKey)
 		}
@@ -593,7 +608,7 @@ func (r *nodeReconciler) reconcileWorkloadsOnNode(
 	}
 
 	for wlKey, pods := range podsToTerminateMap {
-		if evictedWorkloads.Has(wlKey) {
+		if evictedWorkloads.Has(wlKey) || pendingSecondPassWorkloads.Has(wlKey) {
 			continue
 		}
 		if err := r.markPodsFailed(ctx, pods); err != nil {
@@ -715,6 +730,20 @@ func (r *nodeReconciler) hasProgressingPods(ctx context.Context, wl *kueue.Workl
 		return pod.DeletionTimestamp.IsZero() && !utilpod.IsTerminated(pod)
 	})
 	return hasProgressingPods, nil
+}
+
+func (r *nodeReconciler) hasProgressingPodsOnOtherNodes(ctx context.Context, wl *kueue.Workload, nodeName string) (bool, error) {
+	sliceName := workloadslicing.SliceName(wl)
+	pods, err := ListPodsForWorkloadSlice(ctx, r.client, wl.Namespace, sliceName)
+	if err != nil {
+		return false, err
+	}
+	return slices.ContainsFunc(pods, func(pod *corev1.Pod) bool {
+		if pod.Spec.NodeName == "" || pod.Spec.NodeName == nodeName {
+			return false
+		}
+		return pod.DeletionTimestamp.IsZero() && !utilpod.IsTerminated(pod)
+	}), nil
 }
 
 func checkTaintTolerations(logger logr.Logger, taint *corev1.Taint, podSets []kueue.PodSet) taintToleration {
--- a/pkg/controller/tas/node_controller_test.go
+++ b/pkg/controller/tas/node_controller_test.go
@@ -194,6 +194,7 @@ func TestNodeFailureReconciler(t *testing.T) {
 		wantRequeue        time.Duration
 		wantEvictedCond    *metav1.Condition
 		wantPatchedPods    []string
+		wantNotPatchedPods []string
 		featureGates       map[featuregate.Feature]bool
 		// ignoreUnhealthyNodes is used only when we expect the unhealthy nodes list to be cleared.
 		// Patching the status in tests (via fake client and strategic merge interceptor)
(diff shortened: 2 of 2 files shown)


pkg/controller/tas/node_controller.go
```diff
@@ -593,7 +608,7 @@ func (r *nodeReconciler) reconcileWorkloadsOnNode(
 	}
 
 	for wlKey, pods := range podsToTerminateMap {
-		if evictedWorkloads.Has(wlKey) {
```
j-skiba, line 596: will this work properly in the following situation?:

Job with 3 pods - 1 pod per node

- there is a pod running on the node
- node gets tainted
- pod fails and a new pod gets created that would be a stray (pending forever)
- node ctrl reconcile starts
- reconcile removes the stray pod and updates `unhealthyNodes` list in the workload
- another stray pod gets created
- node ctrl reconcile runs again but it doesn't remove the stray pod because there are other pods running for this Job
- new scheduling cycle runs and replaces the unhealthy node and clears the `unhealthyNodes` list

Wouldn't the stray pod stay forever on the unhealthy node?

j-skiba:
could you also add some description or reproduction (if possible) how it affected https://github.com/kubernetes-sigs/kueue/issues/11617?

pkg/controller/tas/node_controller.go
gyliu513: Thanks for the detailed scenario, this helped me tighten the condition.

A few clarifications on what the controller actually does:

- On the first reconcile after the node becomes unhealthy, `unhealthyNodes` does not yet contain this node, so we still call `markPodsFailed` for pending pods on that node (same as before). That is the reconcile that fails the first replacement/stray pod and records the node in unhealthyNodes.

- We only skip `markPodsFailed` on a later reconcile when both are true:
  - the node is already in `status.unhealthyNodes`, and
  - the workload still has progressing pods on other nodes (`hasProgressingPodsOnOtherNodes`).

That matches the flake in #11617: one pod is still running elsewhere while a replacement pod is pending on the tainted node with the old node selector. Skipping `markPodsFailed` there avoids consuming BackoffLimit before second-pass scheduling / eviction can run.

- For the second stray pod in your sequence: while siblings are still running, we intentionally do not fail it immediately. It should not stay forever, because cleanup happens when either:
  - Second pass / eviction completes, topology is updated or the workload is evicted; pods with a node selector for a node that is no longer in the topology assignment are handled as strays (`!hasTASAssignment` path in `checkPodsOnNode`) and are still eligible for `markPodsFailed` (they are not in `pendingSecondPassWorkloads`, which only applies to `workloadUnhealthy`), or
  - No progressing pods remain on other nodes, `hasProgressingPodsOnOtherNodes` becomes false, so we stop skipping and `markPodsFailed` runs again.

I ran `make test-tas-e2e-baseline` locally (14/14 passed, including the hotswap tests). Happy to add an integration test for this sequence if you think it would help. Do you think we need to add this integration test?

gyliu513:
> could you also add some description or reproduction (if possible) how it affected https://github.com/kubernetes-sigs/kueue/issues/11617?

@j-skiba Updated at https://github.com/kubernetes-sigs/kueue/pull/11621#issue-4519854691

pkg/controller/tas/node_controller.go
j-skiba: I think the described issue might still happen.

> For the second stray pod in your sequence: while siblings are still running, we intentionally do not fail it immediately. It should not stay forever, because cleanup happens when either:
Second pass / eviction completes, topology is updated or the workload is evicted; pods with a node selector for a node that is no longer in the topology assignment are handled as strays (!hasTASAssignment path in checkPodsOnNode) and are still eligible for markPodsFailed (they are not in pendingSecondPassWorkloads, which only applies to workloadUnhealthy), or
No progressing pods remain on other nodes, hasProgressingPodsOnOtherNodes becomes false, so we stop skipping and markPodsFailed runs again.

Okay, so the second reconcile skipped the procedure of marking the stray pod as failed, but in order to mark the pod in the future, we would need a third reconcile loop. I don't see what would actually trigger that third reconcile on this unhealthy node.

The node reconciler is triggered by `Node` or `Pod` changes. If we assume that the second reconcile was triggered by an event from the second stray pod itself, we don't have any guarantee that a third reconcile will be triggered for that node later on to finish the cleanup.

> Do you think we need to add this integration test?

Well, after giving this more thought, I think this scenario might be too tricky - or maybe even impossible -to reliably reproduce in an integration test.

mimowo: I will let @j-skiba assess if we can capture the scenario in an integration test.

However, we certainly need unit testing at least for the Reconcile function which will step into the new code and confirm it is needed.

j-skiba: I also have a question about the flake.

I don't see how this PR and the flake are connected.

The outcome of this bug we have here is that Job reaches `backoffLimit` quicker than it should and might move to `Failed` state as a result.

Here are logs from the worker https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/11615/pull-kueue-test-e2e-tas-baseline-main/2058968675913830400/artifacts/run-test-tas-e2e-baseline-1.36.1/kind-worker8/pods/kueue-system_kueue-controller-manager-c5c97b5d4-lstx6_e34feb0d-4041-4dd6-bc16-004f7fc09545/manager/0.log

There's a lot of `field is immutable` job reconcile errors from here - https://github.com/kubernetes-sigs/kueue/blob/e4db3801f6926d4f73b8f6d3f328169d4ad8bd35/pkg/controller/jobs/job/job_controller.go#L199

so the job cannot be evicted and we reach timeout. How does this bug cause this? I don't understand the correlation.

gyliu513: Thanks @j-skiba @mimowo for the discussion here.

The connection goes through BackoffLimit exhaustion:
1. The bug: markPodsFailed was called on each replacement/stray pod on the tainted node even after the node was already recorded in unhealthyNodes. Each call consumes one unit of BackoffLimit.
2. Once BackoffLimit is exhausted, Kubernetes moves the Job to Failed state.
3. Kueue then tries to evict the workload: StopJob → RestorePodSetsInfo → Patch(spec.template) to strip the TAS node selectors.
4. Kubernetes rejects that patch on a Failed Job with "field is immutable" — the patch loop retries forever.
5. The test times out waiting for eviction to complete.

This PR prevents step 1 by skipping markPodsFailed when the node was already recorded in unhealthyNodes and sibling pods are still progressing. The Job never exhausts BackoffLimit, so it never enters Failed state, so the eviction patch succeeds.

I will add a new test case based on @mimowo 's comment.

j-skiba: I see now, thank you for detailed explanation.

The only part I'm hesitant about is the scenario from here https://github.com/kubernetes-sigs/kueue/pull/11621#discussion_r3308922194 where we might be missing the third trigger for reconciliation.

Basically the thing that tells us if removal (and then recreation) of stuck/stray pods is safe is change of workload's TAS assignment after scheduler chooses replacement node or completely new assignment after eviction. By depending on that we would have guarantee that replacement pods will be created on a correct node set. I'm wondering if we should add such behavior to handle these scenarios without incrementing backoff counter for jobs.

There might be some corner cases for pods that would be assigned (via `nodeSelector`) to a node but the assignment is already different, but I should be possible to handle.

wdyt? @gyliu513

mimowo:
@gyliu513 since this is changing the production behavior we need a release note

mimowo:
@gyliu513 I synced with @j-skiba who explained me:
1. the issue this PR is trying to solve is to minimize the number of marking of Pods of Failed to minimize the impact on the Job's backoffLimit
2. the tests have backoffLimit only 1 which make them volnurable to the excessive marking of Failed
3. the current approach has a gap found by Jakub and so we need to fix it before productazing as a fix
4.  I consider the issue relatively low priority, because users can mitigate with larger backoffLimit, otoh, introducing a bug that Jobs get stuck could have very bad consequences

So, the next steps:
1. @j-skiba is going to open a deflake PR to bump the backoffLimit for the Jobs
2. I will open a dedicated issue to minimize that
3. We will carefully think how to fix it

Opened: https://github.com/kubernetes-sigs/kueue/issues/11789

pkg/controller/tas/node_controller.go
gyliu513: @j-skiba You're right, there's no guaranteed third reconcile trigger from the node/pod event stream alone. I see the gap now.

Agreed that tying stray pod removal to changes in the workload's TAS assignment is the cleaner and more correct approach.

Thanks for the thorough review.

gyliu513:
Thanks @mimowo and @j-skiba for the offline sync and the clear summary.

Agreed with the plan, bumping backoffLimit to unblock the flake in the short term, and then tackling the proper optimization in #11789 with the right invariants in place.

Shall I close this PR?

j-skiba:
@gyliu513 I think we can close this PR and continue the discussion there https://github.com/kubernetes-sigs/kueue/issues/11789

Issue #11536, opened 2026-05-22
Optimize the integration tests by using an addional shard


mimowo:
**What would you like to be cleaned**:

I would like to improve execution time for CI for integration tests.

Currently we have two shards "baseline" and "extended", each taking around 25min. The proposal is to introduduce a third shard. That would improve the runtime of a single shard to around 18min probably (TBD)

**Why is this needed**:

Integration tests are currently the longest for CI jobs, their performance is important because:
1. practically speaking for most features they are the most useful type of tests to be executed on CI. For unit tests you can quickly run all in IDE, for e2e they are rarely affected, but integration tests are picking often on non-trivial interactions - so quick feedback is very useful for developers
2. as the longest raking jobs they impact the release process

Current output from the prow_runtimes.py tool restricted to CI jobs which also run as presubmits, for jobs above 20min.

```
| Rank | Job | Avg | Last runtimes | Build IDs | Results |
|---:|---|---:|---|---|---|
| 3 | `periodic-kueue-test-integration-extended-main` | 25:00 | 24:28, 24:36, 25:02, 24:45, 26:08 | 2057838048695029760, 2057656853092896768, 2057475657947942912, 2057294461309816832, 2057113069535891456 | SUCCESS, SUCCESS, SUCCESS, SUCCESS, SUCCESS |
| 4 | `periodic-kueue-test-integration-baseline-main` | 24:44 | 25:20, 23:42, 23:34, 25:47, 25:15 | 2057838048606949376, 2057656853017399296, 2057475657822113792, 2057294461204959232, 2057113069435228160 | SUCCESS, SUCCESS, SUCCESS, SUCCESS, SUCCESS |
| 5 | `periodic-kueue-test-e2e-extended-main-1-35` | 22:57 | 20:43, 24:14, 24:18, 21:58, 23:31 | 2057842075566608384, 2057660879255638016, 2057479683989049344, 2057298236728676352, 2057116850151493632 | SUCCESS, SUCCESS, SUCCESS, SUCCESS, SUCCESS |
| 6 | `periodic-kueue-test-e2e-multikueue-main` | 22:47 | 24:00, 21:56, 21:56, 23:42, 22:21 | 2057789732603039744, 2057608534413021184, 2057427338534064128, 2057245891835727872, 2057064499231330304 | SUCCESS, SUCCESS, SUCCESS, SUCCESS, SUCCESS |
[...]

PR #11629, opened 2026-05-26
Dynamic Sharding for Integration Tests

gabesaba:
#### What type of PR is this?
/kind cleanup
/area testing

Introduce sharding for integration tests. Adding a new shard will
require a small change to [test-infra](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/kueue/kueue-presubmits-main.yaml). We can discontinue sharding based on baseline/extended, and
simply increase the number of shards when the test runs become too slow.

This change is backwards-compatible: no-op until `TOTAL_SHARDS` is set.

Data from running locally. Each shard was run sequentially, so there is no contention. Baseline is make test-integration, including both baseline and extended suite. 
| #Shards | Shard Run Times | Wall Time | Improvement vs Baseline |
|---|---|---|---|
| **1** | 20m34s | 20m34s | 0.0% |
| **2** | 9m43s, 10m42s | 10m42s | 48.0% |
| **3** | 7m51s, 6m30s, 5m42s | 7m51s | 61.8% |
| **4** | 4m37s, 7m18s, 4m38s, 3m16s | 7m18s | 64.5% |
| **5** | 3m51s, 2m06s, 3m35s, 7m49s, 2m27s | 7m49s | 62.0% |

Part of #11536

Mostly LLM generated. I endorse the changes.

```release-note
NONE
```

### DIFF
--- a/Makefile-test.mk
+++ b/Makefile-test.mk
@@ -35,7 +35,14 @@ TEST_LOG_LEVEL ?= -3
 INTEGRATION_NPROCS ?= 4
 INTEGRATION_NPROCS_MULTIKUEUE ?= 3
 # Folder where the integration tests are located.
+ifdef INTEGRATION_TOTAL_SHARDS
+INTEGRATION_TARGET := $(shell ./hack/testing/shard-integration-tests.sh $(INTEGRATION_SHARD_INDEX) $(INTEGRATION_TOTAL_SHARDS))
+ifeq ($(INTEGRATION_TARGET),)
+$(error Aborting execution due to invalid sharding parameters (INTEGRATION_SHARD_INDEX / INTEGRATION_TOTAL_SHARDS))
+endif
+else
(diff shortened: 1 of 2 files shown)


Makefile-test.mk
```diff
@@ -35,7 +35,14 @@ TEST_LOG_LEVEL ?= -3
 INTEGRATION_NPROCS ?= 4
 INTEGRATION_NPROCS_MULTIKUEUE ?= 3
 # Folder where the integration tests are located.
+ifdef TOTAL_SHARDS
```
mimowo: nit, but I think it would be better to add the `INTEGRATION_` prefix for the params, for consistency, but also readability

gabesaba: Done.

mimowo:
Awesome, let's give it a try 👍 
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16

PR #11615, opened 2026-05-25
Fix MultiKueue to work on 1.36

mimowo:
#### What type of PR is this?

/kind bug
/area multikueue

#### Which issue(s) this PR fixes:

Fixes #11614

#### Special notes for your reviewer:

The issue is fixed by making sure the JobSuspended=True condition is synced from the worker to the manager cluster.

Once the manager has the condition, then the  k8s validation allows to inject scheduling directives like nodeSelectors.

#### Does this PR introduce a user-facing change?

```release-note
MultiKueue: fix MultiKueue to work on k8s 1.36. Additionally now MultiKueue does not skip syncing status from the
worker cluster when the JobSuspended=True is set.
```

PR #11649, opened 2026-05-26
[v2] Fix MultiKueue to work on 1.36

### LINKED ISSUE #11614: MultiKueue is not working on k8s 1.36
**What happened**:

MultiKueue is not working on k8s 1.36

MultiKueue works well on k8s 1.36

Here is the test: https://github.com/kubernetes-sigs/kueue/pull/11576

This is due to regression in the core k8s: https://github.com/kubernetes/kubernetes/issues/139281

olekzabl:
#### What type of PR is this?

/kind bug
/area multikueue

Follow-up for @mimowo's #11615.
It adjusts MultiKueue lifecycle of a Job to be compatible with K8s 1.36+ validation changes. (See https://github.com/kubernetes/kubernetes/issues/139281).

Fixes #11614

The old code skipped `syncElasticJob` (i.e. adjusting worker spec to manager spec) in cases when chose to skip syncing statuses. 
There seems to be no good reason behind this, so this PR removes this skipping.

```release-note
MultiKueue: Fixed admission for Kubernetes Jobs on Kubernetes 1.36 clusters by ensuring all Job status updates comply with the updated Kubernetes Job validation rules. See kubernetes/kubernetes#139281 for more details.
```


olekzabl:
The e2e test failure on commit [4629083](https://github.com/kubernetes-sigs/kueue/pull/11649/commits/46290832aee4c91a6cf2e3a9c710f8267f3fbfef) seems to have been very unlucky:
I just ran this 20 times on another branch (#11688) and got all successes

mimowo:
/test pull-kueue-test-e2e-baseline-main-1-33
/test pull-kueue-test-e2e-extended-main-1-33
look like unrelated infra failures

pkg/controller/jobs/job/job_multikueue_adapter.go
```diff
@@ -103,6 +102,54 @@ func (b *multiKueueAdapter) SyncJob(ctx context.Context, localClient client.Clie
 	return remoteClient.Create(ctx, &remoteJob)
 }
 
+func determineStatusUpdate(ctx context.Context, log logr.Logger, localJob, remoteJob *batchv1.Job) *batchv1.JobStatus {
+	localJobInfo := fromObject(localJob)
+
+	log.V(2).Info("Checking shouldSkipSyncForSuspendedLocalJob", "localJob", localJob, "remoteJob", remoteJob)
```
mimowo:
```suggestion
	log.V(3).Info("Checking determineStatusUpdate", "localJob", localJob, "remoteJob", remoteJob)
```

mimowo: or improve the messaging

pkg/controller/jobs/job/job_multikueue_adapter.go
```diff
@@ -103,6 +102,54 @@ func (b *multiKueueAdapter) SyncJob(ctx context.Context, localClient client.Clie
...
+	newLocalStatus.Conditions = setJobStatusCondition(newLocalStatus.Conditions,
+		batchv1.JobCondition{
+			Type:               batchv1.JobSuspended,
+			Status:             corev1.ConditionTrue,
+			Reason:             "MultiKueueAdapter",
+			Message:            "Set by MultiKueue adapter",
+			LastTransitionTime: metav1.Now(),
+			LastProbeTime:      metav1.Now(),
+		})
+	log.V(2).Info("Updating the localJob suspended Job to set the JobSuspended=True condition")
```
mimowo: IIUC this is only needed for k8s 1.36.0 and 1.36.1 as https://github.com/kubernetes/kubernetes/pull/139287 is merging. Since we can de-comission the code in the future let's add a comment. I'm also thinking about a feature gate like

MultiKueueInjectJobSuspendedOnManager beta by default. OTOH maybe this would be an overkill, any opinion?

Makefile-test.mk
mimowo, line 51: Revert the changes

mimowo (commented):
LGTM, thank you 👍

Makefile-test.mk
mimowo: This might be the reason for tbe failing 1.33 jobs

pkg/controller/jobs/job/job_multikueue_adapter.go
mimowo: I think just a comment should be enough for now.

pkg/controller/jobs/job/job_multikueue_adapter.go
olekzabl: Done.
(BTW I also did some other changes in logs).

pkg/controller/jobs/job/job_multikueue_adapter.go
olekzabl: I added a comment for this, and BTW added more comments on various cases in the new logic.

olekzabl: > I'm also thinking about a feature gate (...) OTOH maybe this would be an overkill, any opinion?

I'm feeling a cleaner thing than an FG would be to "peek" which K8s version we're on, and do the trick only for the versions where it's needed.
Then, the trick would "auto-disable itself" as soon as no longer needed.

Though I'm not sure if, and how, we can do such "peeking".
I see [`FetchServerVersion`](https://github.com/kubernetes-sigs/kueue/blob/13975fe00f669ce369f5992e0f9bbcdc1822a9d5/pkg/util/kubeversion/kubeversion.go#L73) though its only non-test usage seems to reach as far as to [this field](https://github.com/kubernetes-sigs/kueue/blob/13975fe00f669ce369f5992e0f9bbcdc1822a9d5/pkg/controller/jobframework/reconciler.go#L115), which is then... unused?
So this whole util seems dead code which I'd be afraid to happily use in this PR.
Maybe we can try this a follow-up. WDYT?

Makefile-test.mk
olekzabl: Oops, it was my intent to revert them but apparently I mishandled it.

Reverted now, and indeed tests are greener.

pkg/controller/jobs/job/job_multikueue_adapter.go
```diff
@@ -103,6 +102,70 @@ func (b *multiKueueAdapter) SyncJob(ctx context.Context, localClient client.Clie
...
+	newLocalStatus.Active = 0
+	newLocalStatus.StartTime = nil
+	newLocalStatus.Conditions = setJobStatusCondition(newLocalStatus.Conditions,
+		batchv1.JobCondition{
+			Type:               batchv1.JobSuspended,
+			Status:             corev1.ConditionTrue,
+			Reason:             "MultiKueueAdapter",
+			Message:            "Set by MultiKueue adapter",
+			LastTransitionTime: metav1.Now(),
+			LastProbeTime:      metav1.Now(),
```
mimowo: I think this risks a hot-loop updates, because we will always have `!equality.Semantic.DeepEqual(localJob.Status, *statusUpdate)`

So, I think we need to either exclude the timestamps from the DeepEqual or check if Active > 0 || !JobSuspended=True || StartTime != nil. Similarly as tested by @sohankunkerkar here: https://github.com/kubernetes-sigs/kueue/pull/11699/changes#diff-59bc2e31a8d561766b6c4d6e22449ead320ea709ea5e85cf857d0ba81c8c8539R126-R129

olekzabl: Done.
I chose to only check the condition before playing with the condition.
That way controlling non-triviality of updates to `Active` & `StartTime` is deferred to existing `DeepEqual`. So we win a simpler `if`.

mimowo: Lgtm

pkg/controller/jobs/job/job_multikueue_adapter.go
```diff
@@ -103,6 +102,82 @@ func (b *multiKueueAdapter) SyncJob(ctx context.Context, localClient client.Clie
...
+	//    (remote admitted -> MultiKueue AdmissionCheck ready -> local admitted -> local unsuspended).
+	// 3. However, to avoid clashing with K8s 1.36 validation rules
+	//    (since https://github.com/kubernetes/kubernetes/pull/135104 until https://github.com/kubernetes/kubernetes/pull/139287)
+	//    we manually patch a "JobSuspended=True" condition and "Active=0" to unblock further spec updates.
+	//    Similarly, to make 1.35 happy, we patch "StartTime = nil".
+	// As long as we're on a K8s version unaffected by #3, we could just do:
+	//   return &localJob.Status
+	newLocalStatus := localJob.Status.DeepCopy()
+	newLocalStatus.Active = 0
+	newLocalStatus.StartTime = nil
```
mimowo, line 145: Actually, I don't think we need these updates: `newLocalStatus.Active = 0` and `newLocalStatus.StartTime = nil`.

 I suggested them thinking about the k8s validation, but I think this is "defensive programming" which can have a negative impact on code simplicity, and correctness in some cases.

The relevant validation in the k8s for starting a job (which requires injection of scheduling directives like nodeSelectors):
- 1.36 requires `status.Active=0`: https://github.com/kubernetes/kubernetes/blob/release-1.36/pkg/registry/batch/job/strategy.go#L198
- 1.35 requires `status.startTime=nil`: https://github.com/kubernetes/kubernetes/blob/release-1.35/pkg/registry/batch/job/strategy.go#L192-L193

Now, while with this updates we will make the validation pass we can see these are defensive programming when we consider how can we step into the code.

The stepping into the code happens when:
- Job on manager is supended, spec.suspend=True
- Job on worker is running, spec.suspend=false
- the job is not finished

Now, to get into this state I can see the following scenarios:
1. start of a newly created Job (worker starts and manager remains stopped)
2. start of a Job after stop (worker starts and manager is still stopped)
3. stop is called on manager (worker remains running)

**Let's first think about the need to set `status.active=0`.**
1. not needed because we have it already for new job
2. not needed because we awaited on active=0 when stopping the previously running Job
3. not needed, because once worker is also stopped, then we will start syncing the status between stopped Jobs using the code above

Actually, (3.) is the slightly problematic case, because we may set status.active=0 which will be updated to >0 just a moment later. It may also result in premature quota freeing.

**Let's first think about the need to set `status.startTime=nil`.**

1. not needed because we have status.startTime=nil for new jobs
2. not needed because it was cleaned already by our job_controller code [here](https://github.com/kubernetes-sigs/kueue/blob/a393561c2072b1b1366a0bd5ec915be897219373/pkg/controller/jobs/job/job_controller.go#L187)
3. not needed because again it will be cleaned by the Job controller integration

At the same time here I'm less sure it is a problem, because k8s core is clearing the field when the Job is suspended in 1.36+. Since we take the ownership of adding the JobSuspended=True it is more subtle here, ptal: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/job/job_controller.go#L1168-L1170

mimowo:
/test all
check stability again

pkg/controller/jobs/job/job_multikueue_adapter.go
mimowo: Yes, FetchServerVersion is some experiment from the past we introduced for a similar case. Later it was not needed, but we kept it "just in case" to avoid re-implementing from scratch.

However, in this case I wouldn't go for it, I think it adds to much complexity. 

I would rather merge as is since we tested the code works both on 1.35 and 1.36. 

My intention was to keep the version intependent code, but have a mechanism to decomission the code gracefully when no longer needed. The FG could allow us to make the workaround enabled now,  but disabled (alpha) in when k8s 1.36 is EOL, which is probably like a year from now.

mimowo:
/test all
check stability again

pkg/controller/jobs/job/job_multikueue_adapter.go
mimowo: We can introduce the FG at the moment we start decomisioning the code

mimowo:
Thank you for working together on fixing this 👍 

/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16

Please prepare CPs in case the robot fails on conflicts

mimowo:
/release-note-edit
```release-note
MultiKueue: Fixed Job admission on Kubernetes 1.36 clusters by keeping newly created
suspended manager Jobs compatible with Kubernetes validation.
```

mimowo:
/release-note-edit
```release-note
MultiKueue: Fixed admission for Kubernetes Jobs on Kubernetes 1.36 clusters by ensuring all Job status updates comply with the updated Kubernetes Job validation rules. See kubernetes/kubernetes#139281 for more details.
```

PR #11676, opened 2026-05-27
fix: PodsReady for SparkApplication framework

### LINKED ISSUE #11675: SparkApplication framework: PodsReady ignores executor state, and Stop leaves orphan executor pods across requeue cycles
**What happened**:
Two independent bugs in the v0.18 SparkApplication integration framework (`pkg/controller/jobs/sparkapplication/`) which together break Kueue's requeue cycle for Spark workloads.

### Bug 1 — `PodsReady` returns true while executors are still pending

Current implementation:

```go
// pkg/controller/jobs/sparkapplication/sparkapplication_controller.go
func (j *SparkApplication) PodsReady(ctx context.Context, _ client.Client) bool {
    return j.Status.AppState.State == sparkv1beta2.ApplicationStateRunning
}
```

`AppState.State == Running` transitions to `Running` as soon as the **driver** is up, regardless of executor state. When executors are unschedulable (resource shortage, missing node label, etc.), `PodsReady` is permanently `true`, so `waitForPodsReady.timeout` never fires. The Workload sits forever with the driver running and 0 executors ready.

### Bug 2 — `Stop` only sets `spec.suspend=true`, leaving orphan executor pods across requeue cycles

The framework relies on the default `stopJob` flow which only patches `spec.suspend=true` on the SparkApplication and assumes Spark Operator cleans up. In practice Spark Operator's suspend handling only kills the driver; executor pods are owned by the SparkApplication CR via a non-controller `OwnerReference` and do not get GC'd when the driver dies.

Result — observed across two requeue cycles:

```
$ kubectl get pods -n default
NAME                               READY   STATUS    RESTARTS   AGE
pythonpi-7f25689e671b40fd-exec-1   0/1     Pending   0          4m58s   <- cycle 1
pythonpi-7f25689e671b40fd-exec-2   0/1     Pending   0          4m58s   <- cycle 1
pythonpi-db71229e671def82-exec-1   0/1     Pending   0          2m3s    <- cycle 2
pythonpi-db71229e671def82-exec-2   0/1     Pending   0          2m3s    <- cycle 2
```

Each evict -> re-admit cycle leaks N executor pods. With `waitForPodsReady.requeuingStrategy.backoffLimitCount: 5` and `executor.instances: 2`, the cluster accumulates 10 zombie executor pods before the Workload is finally deactivated.
- `PodsReady` returns `true` only when the driver **and** all requested executors are running (or completed), so `waitForPodsReady.timeout` actually fires when executors cannot be scheduled.
- After eviction, `Stop` tears down driver and all executor pods so the next requeue cycle starts clean. After 5 cycles, 0 pods remain (not 10).

---
1. Cluster with Kueue v0.18 + Spark Operator >= v2.4.0, feature gate `SparkApplicationIntegration=true`.
2. ClusterQueue + LocalQueue using `default-flavor`, with quota sufficient for the driver. Bug 2 manifests on any eviction; Bug 1 needs the executor to be unschedulable so the Workload waits for the timeout that never fires.
3. Configure Kueue with `waitForPodsReady.timeout: 2m` and `requeuingStrategy.backoffLimitCount: 5`.
4. Apply the following SparkApplication. The fake `nodeSelector` deliberately makes executors unschedulable so Bug 1 is reachable:
```yaml
apiVersion: sparkoperator.k8s.io/v1beta2
kind: SparkApplication
metadata:
  name: spark-bug-repro
  namespace: default
  labels:
    kueue.x-k8s.io/queue-name: spark-test
spec:
  type: Python
  mode: cluster
  image: "docker.io/apache/spark:3.5.3-python3"
  imagePullPolicy: IfNotPresent
  mainApplicationFile: "local:///opt/spark/examples/src/main/python/pi.py"
  sparkVersion: "3.5.3"
  driver:
    cores: 1
    memory: "512m"
  executor:
    cores: 1
    instances: 2
    memory: "512m"
    nodeSelector:
      non-existent-node-label: "true"   # forces executors to stay Pending
  restartPolicy:
    type: Never
```

**Observed (Bug 1)**: The Workload never gets a `PodsReadyTimeout` eviction. Driver stays `Running`, executors stay `Pending`, indefinitely. With Bug 1 patched out (`PodsReady` also checks executor state), `Evicted=True / reason=PodsReadyTimeout` fires at the configured 2-minute mark.

**Observed (Bug 2)** — using a local patch to make Bug 1 trigger evictions, or by manually requeuing the workload: each cycle accumulates 2 zombie executor pods. After 5 cycles -> 10 zombie pods; cycle 1's executors are still `Pending` when cycle 5 starts.

---

### Why these two bugs travel together

Bug 1 prevents the requeue cycle from ever firing (timeout never triggers), so Bug 2 normally does not manifest in production. They were both discovered while testing the requeue cycle end-to-end with a local workaround for Bug 1 in place.

### `Stop` ordering matters — two "obvious" implementations are incorrect

| Attempt | Why it fails |
|---|---|
| `spec.suspend=true` then immediately `Delete` executors | Driver's Spark scheduler can issue `create executor` API calls right up until its container terminates -> a trailing executor (owner=SparkApplication, no driver pod ref) is created **after** our Delete and lingers as an orphan. |
| `Delete` driver first (force) then `suspend` then `Delete` executors | Force-deleting the driver makes Spark Operator transition `AppState` to `Failed`, which the framework's `Finished()` interprets as `Workload.Finished=True` — the requeue cycle is prematurely terminated. |

Correct ordering: (1) `spec.suspend=true` (Spark Operator gracefully tears down the driver, `AppState` stays transitionable to `Running` on resume); (2) bounded poll (<= 5s) until driver pod is gone; (3) delete all executor pods. This needs to live in a framework custom `Stop` since `JobWithCustomStop` bypasses the default `Suspend()` + `RestorePodSetsInfo()` flow in `pkg/controller/jobframework/reconciler.go`.
**Environment**:
- Kubernetes version: 1.32.0
- Kueue version: v0.18
- Spark Operator version: 2.5.0

hahahaheihei:
#### What type of PR is this?
/kind bug

this pr fixed sparkapplication can't requeue when driver pod or executor unready timeout 

Fixes #11675

```release-note
SparkApplication: Fixed a bug where `PodsReady` returned true based on driver state alone, so workloads
with stuck executors never reached `waitForPodsReady.timeout`.
```

### DIFF
--- a/pkg/controller/jobs/sparkapplication/sparkapplication_controller.go
+++ b/pkg/controller/jobs/sparkapplication/sparkapplication_controller.go
@@ -299,7 +299,26 @@ func (j *SparkApplication) Finished(ctx context.Context) (message string, succes
 }
 
 func (j *SparkApplication) PodsReady(ctx context.Context, _ client.Client) bool {
-	return j.Status.AppState.State == sparkv1beta2.ApplicationStateRunning
+	// Driver must be running.
+	if j.Status.AppState.State != sparkv1beta2.ApplicationStateRunning {
+		return false
+	}
+	// And every requested executor must have reported a Running/Completed state.
+	// AppState.State alone goes to Running as soon as the driver starts even if
+	// executors are stuck (e.g. unschedulable), which would let the
+	// waitForPodsReady timeout never fire on heterogeneous resource shortages.
+	expected := int(ptr.Deref(j.Spec.Executor.Instances, 0))
+	if expected == 0 {
+		return true
+	}
+	ready := 0
+	for _, st := range j.Status.ExecutorState {
+		switch st {
+		case sparkv1beta2.ExecutorStateRunning, sparkv1beta2.ExecutorStateCompleted:
+			ready++
+		}
+	}
+	return ready >= expected
 }
 
 func SetupIndexes(ctx context.Context, indexer client.FieldIndexer) error {
--- a/pkg/controller/jobs/sparkapplication/sparkapplication_controller_test.go
+++ b/pkg/controller/jobs/sparkapplication/sparkapplication_controller_test.go
@@ -17,8 +17,10 @@ limitations under the License.
 package sparkapplication
 
 import (
+	"fmt"
 	"maps"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
@@ -28,11 +30,16 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/component-base/featuregate"
+	testingclock "k8s.io/utils/clock/testing"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	configapi "sigs.k8s.io/kueue/apis/config/v1beta2"
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+	schdcache "sigs.k8s.io/kueue/pkg/cache/scheduler"
 	"sigs.k8s.io/kueue/pkg/controller/jobframework"
 	"sigs.k8s.io/kueue/pkg/features"
 	"sigs.k8s.io/kueue/pkg/podset"
@@ -48,6 +55,7 @@ var (
 	}
 	workloadCmpOpts = cmp.Options{
 		cmpopts.EquateEmpty(),
+		cmpopts.SortSlices(func(a, b metav1.Condition) bool { return a.Type < b.Type }),
 		cmpopts.IgnoreFields(kueue.Workload{}, "TypeMeta"),
 		cmpopts.IgnoreFields(metav1.ObjectMeta{}, "Name", "Labels", "ResourceVersion", "OwnerReferences", "Finalizers"),
 		cmpopts.IgnoreFields(kueue.WorkloadSpec{}, "Priority"),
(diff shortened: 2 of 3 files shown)


mimowo:
Please add the release-note block

hahahaheihei:
> Please add the release-note block
i have completed the release-note，please review the pr

pkg/controller/jobs/sparkapplication/sparkapplication_controller.go
```diff
@@ -298,8 +302,115 @@ func (j *SparkApplication) Finished(ctx context.Context) (message string, succes
...
+	// move on — the next eviction cycle will clean up any orphan executor.
+	var driverList corev1.PodList
+	if err := c.List(ctx, &driverList,
+		client.InNamespace(j.GetNamespace()),
+		client.MatchingLabels{
+			sparkcommon.LabelSparkAppName: j.GetName(),
+			sparkcommon.LabelSparkRole:    sparkcommon.SparkRoleDriver,
+		},
+	); err == nil && len(driverList.Items) > 0 {
+		driverKeys := make([]client.ObjectKey, 0, len(driverList.Items))
```
mimowo: do not silence errors, propagate up

pkg/controller/jobs/sparkapplication/sparkapplication_controller.go
```diff
@@ -298,8 +302,115 @@ func (j *SparkApplication) Finished(ctx context.Context) (message string, succes
...
+				err := c.Get(ctx, key, &p)
+				if err == nil {
+					return false, nil
+				}
+				if !apierrors.IsNotFound(err) {
+					return false, err
+				}
+			}
+			return true, nil
+		})
```
mimowo: The periodic poll is not a good pattern, and is potentially racy. 

I'm wondering - why not just return error if there are any pods which are not terminating? Then the job reconciler should re-enter Stop, and kill some more pods

hahahaheihei: in my test ，we shoul wait for driver pod be cleaned,then kill pending executor pod ,otherwise, the driver pod may occasionally be re-launched before being cleaned, and when a new driver pod starts up, it will launch a new executor pod as well.  if return err 
```
	if ptr.Deref(j.Spec.Suspend, false) {
		return false, nil
	}
```
this code ,will never kill executor pod

pkg/controller/jobs/sparkapplication/sparkapplication_controller.go
hahahaheihei: You're right; thank you for your suggestion.

pkg/controller/jobs/sparkapplication/sparkapplication_controller.go
hahahaheihei: Perhaps a more elegant implementation would be to delegate both the executor cleanup and driver cleanup to the Spark-operator, while the kueue only needs to manage the suspend operation at critical moments. @mimowo

mimowo: Oh yes, this is how the integration k8s job, JobSet or other CRDs work - we just flip the suspend, and wait for termination in isActive

hahahaheihei: I have submitted the corresponding issue to the Spark community. https://github.com/kubeflow/spark-operator/issues/2961

mimowo: yeah but still waiting 5s seems arbitrary and racy im wondering if we can fix it better even if the code is in kueue

hahahaheihei: I have removed the stop-related logic. I believe that with proper configuration of ownerReferences, driver pod and executor pod should be cascaded for deletion.

mimowo: sgtm it would be great if you can confirm this is working well on a live cluster, plus let's adjust the release note.

hahahaheihei: done

pkg/controller/jobs/sparkapplication/sparkapplication_controller.go
```diff
@@ -299,7 +299,26 @@ func (j *SparkApplication) Finished(ctx context.Context) (message string, succes
 }
 
 func (j *SparkApplication) PodsReady(ctx context.Context, _ client.Client) bool {
-	return j.Status.AppState.State == sparkv1beta2.ApplicationStateRunning
+	// Driver must be running.
+	if j.Status.AppState.State != sparkv1beta2.ApplicationStateRunning {
```
mimowo, line 303: Could you add a unit test for the Reconcile, IIUC it will call PodsReady

hahahaheihei: sure ，i have added some unit test

mimowo (commented):
/lgtm
/approve
Thank you for bugfixing the integration 👍 
/cherrypick release-0.17

mimowo:
@hahahaheihei please prepare the cherrypick if you would like the fix to be included in patch release

hahahaheihei:
> @hahahaheihei please prepare the cherrypick if you would like the fix to be included in patch release

sure , once i resolve the conflict , i will submit a new pr

mimowo:
Awesome. we are planning the patch releases today, around 11 am Polish time.

Otherwise it will go to the next patch release, in two weeks.

PR #11680, opened 2026-05-27
test: add webhook scale conversion repro script

### LINKED ISSUE #9145: Introduce a test to reproduce conversion webhooks on 0.15

mykysha:
#### What type of PR is this?
/kind feature

Add a script used to reproduce the scale webhook conversion issue appearing on apiserver restart

Fixes #9145

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added an experimental reproduction tool to generate large configurable workloads with concurrency, QPS, payload padding, retries, and progress/failure reporting.
  * Added automation and a containerized runtime to run single- or two-pass experiments (local or cloud) including cluster setup, gate toggling, apiserver restart, and log-based instability detection.

* **Documentation**
  * Added a README with usage, configuration guidance, run modes, and tuning recommendations.


mykysha:
During local reproduction attempts, I discovered that the webhook conversion instability is triggered not just by workload volume, but also by workload weight. To reliably reproduce the issue, this PR introduces a configurable populator where overall volume is controlled by a workload count, and individual object weight is artificially inflated by specifying the number of containers and padded environment variables. Because aggressively creating heavy workloads puts immense pressure on the cluster, the populator also exposes QPS and worker controls to throttle creation and avoid triggering validation webhook failures. Another populator-related adjustment is an option to run it in-cluster to reduce the pressure from creating workloads externally.

To successfully force the failure locally in kind, the orchestrator script introduces environment constraints like reducing the etcd compaction interval and throttling Kueue’s available CPU, forcing the controller to take longer to parse the heavy workloads. This execution is fully automated in local environments, though running it against a cloud cluster requires manual intervention since the script cannot directly restart or access logs from a managed API server.

Using this script, I successfully reproduced the issue locally with the following configuration: `--workloads 2000 --containers 60  --envs 20 --etcd-interval 10s`. The script can test the cluster with the `ConcurrentWatchObjectDecode` feature gate both disabled and enabled, demonstrating how toggling this option helps kueue overcome large amount of workloads on apiserver restart.

mykysha:
The script is independent from kueue codebase, and can be stored anywhere, the test/repro forlder is just a suggestion, a place to put the script in in order to create a PR

cmd/experimental/webhook-conversion-repro/Dockerfile
mbobrovskyi, line 1: Should we move it to cmd/experimental?

mbobrovskyi (commented):
Could you please add README.md with instruction how to run it?

cmd/experimental/webhook-conversion-repro/Dockerfile
mykysha: Moved

cmd/experimental/webhook-conversion-repro/Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+WORKDIR /src
+COPY go.mod go.sum ./
+RUN go mod download
+COPY populator.go .
+RUN CGO_ENABLED=0 go build -o /out/populator populator.go
+
+FROM gcr.io/distroless/static:nonroot
+COPY --from=build /out/populator /populator
+USER 65532:65532
+ENTRYPOINT ["/populator"]
```
mimowo: nit, but maybe let's name it main.go, which clearly indicates this is the entry point.

This is just brainstoring, feel free to ignore.

cmd/experimental/webhook-conversion-repro/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM golang:latest AS build
```
mimowo: nit, IIRC we use pinned golang version in other projects so let's use that for consistency, and also so that search works well

cmd/experimental/webhook-conversion-repro/populator.go
```diff
@@ -0,0 +1,167 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+func main() {
```
mimowo: This function seems very heavy, maybe let's split it into functions, one for parsingAndLoading the configuration, the other for actual running. 

Then we could write unit tests, not that we need unit tests now. Up to you,

cmd/experimental/webhook-conversion-repro/populator.go
```diff
@@ -0,0 +1,167 @@
...
+	}()
+
+	for i := *start; i < *start+*count; i++ {
+		idx := i
+		sem <- struct{}{}
+		eg.Go(func() error {
+			defer func() { <-sem }()
+
+			// 100-char padding for env values when --envs > 0.
+			pad := "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ_padding_padding_padding_padding_padding"
```
mimowo: Use some helper to generate the string.

cmd/experimental/webhook-conversion-repro/repro.sh
```diff
@@ -0,0 +1,388 @@
+#!/bin/bash
```
mimowo, line 1: It would be great if another pair of eyes could confirm the repro is working before merging, maybe @vladikkuzn or @mszadkow ?

cmd/experimental/webhook-conversion-repro/Dockerfile
mimowo: I think we could also have a Makefile target for that, that could be invoked with different paramters.

That could be used via a dedicated on-demand CI job.

We recently introduced on-demand CI job for WAS and it seems like a good idea, ptal: pull-kueue-test-e2e-k8s-main-was

cmd/experimental/webhook-conversion-repro/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+### 2. The Orchestrator (`repro.sh`)
+An end-to-end automation bash script that builds the environment, deploys Kueue, runs the populator, reboots the API server to clear the cache, and extracts the etcd compaction logs to determine if the instability successfully triggered.
+
+**Environment Constraints:**
+To successfully force the failure locally in `kind`, the script introduces specific cluster constraints:
+- `--etcd-interval`: Artificially lowers the etcd compaction time to reduce the time window the controller has to parse workloads before a compaction storm forces an infinite relist loop.
+- `--kueue-cpu`: Artificially throttles Kueue's available CPU, forcing the controller to take longer to parse the generated heavy workloads.
+
+## Usage
```
mimowo, line 35: If we make it fully automated, and with an on-demand CI job then maybe we also have a param like "expected-result: failure/success" so that we can make the job green when the tests fail.

This can certainly be a follow up

mimowo:
I left comments, some of them could be addressed as follow ups, but overall it looks great even as is.

/approve
Before merging I would like someone to confirm this is working, two scenarios:
- it reproduces the problem when ConcurrentWatchObjectDecode is disabled
- with the same parameters it passes when ConcurrentWatchObjectDecode is enabled
/hold
Maybe @vladikkuzn or @mszadkow could double check that.

Approved and leaving lgtm to @mbobrovskyi in case this is ready when I'm OOO next week.

cmd/experimental/webhook-conversion-repro/repro.sh
```diff
@@ -0,0 +1,388 @@
...
+  name: local-queue
+  namespace: default
+spec:
+  clusterQueue: cluster-queue
+EOF
+
+  if [ "$POPULATOR_MODE" = "cluster" ]; then
+    echo "=== ${LOG_PREFIX}Step 3: Build + load populator image ==="
+    if ! docker image inspect populator:local >/dev/null 2>&1; then
+      docker build -t populator:local -f "$REPRO_DIR/Dockerfile.populator" "$REPRO_DIR" >/dev/null
```
Jefftree: `s/Dockerfile.populator/Dockerfile`?

cmd/experimental/webhook-conversion-repro/repro.sh
```diff
@@ -0,0 +1,388 @@
...
+      echo "  --kueue-version <version>  (default: v0.15.0)"
+      echo "  --populator-mode <mode>    local or cluster (default: local)"
+      echo "  --workloads <count>        Number of workloads (default: 50000)"
+      echo "  --qps <qps>                Populator QPS (default: 5)"
+      echo "  --workers <count>          Populator workers (default: 10)"
+      echo "  --containers <count>       Containers per workload (default: 1)"
+      echo "  --envs <count>             Padded env vars per container (default: 0)"
+      echo "  --gate <mode>              false, true, or both (default: both)"
+      echo "  --etcd-interval <time>     Compaction interval (default: 5m)"
+      echo "  --kueue-cpu <cpu>          Kueue CPU limit (e.g., 100m) (default: unthrottled)"
```
Jefftree: Is there a reasonable value for this that's not uncapped? For scalability of `ConcurrentWatchObjectDecode`, we would really like to confirm that we don't overwhelm the conversion webhook with the feature enabled.

Jefftree (commented):
Really awesome to see this effort!

cmd/experimental/webhook-conversion-repro/repro.sh
mykysha: Sorry, forgot to rename everywhere, will include with other changes

cmd/experimental/webhook-conversion-repro/repro.sh
mykysha: I think I'll change this to be the "default" rather than "unthrottled" for better transparency. Basically without passing the parameter kueue will run with default value which I believe to be 500m, and the lowest value I recorded to be able to pass the script is 100m.

cmd/experimental/webhook-conversion-repro/main.go
```diff
@@ -0,0 +1,211 @@
...
+		if home := os.Getenv("HOME"); home != "" {
+			kubeconfig = filepath.Join(home, ".kube", "config")
+		}
+	}
+
+	if kubeconfig != "" {
+		config, err = clientcmd.BuildConfigFromFlags("", kubeconfig)
+	}
+	if err != nil || config == nil {
+		config, err = rest.InClusterConfig()
```
mbobrovskyi: I think we will rewrite err. Maybe we should join errors?

cmd/experimental/webhook-conversion-repro/repro.sh
```diff
@@ -0,0 +1,395 @@
...
+    else
+      echo "Image populator:local already exists, skipping build."
+    fi
+    kind load docker-image populator:local --name "$CLUSTER"
+
+    echo "=== ${LOG_PREFIX}Step 4: Populate $WORKLOAD_COUNT v1beta1 workloads (CLUSTER) ==="
+    kubectl create clusterrolebinding populator-admin --clusterrole=cluster-admin \
+      --serviceaccount=default:default 2>/dev/null || true
+    kubectl delete job kueue-populator --ignore-not-found >/dev/null
+    kubectl apply -f - <<EOF >/dev/null
```
mbobrovskyi, line 212: Should we wait after apply?

mykysha: Added the wait condition afterwards

cmd/experimental/webhook-conversion-repro/main.go
mykysha: Adjusted so that both errors are preserved

cmd/experimental/webhook-conversion-repro/Dockerfile
mykysha: Done

cmd/experimental/webhook-conversion-repro/populator.go
mykysha: Done

cmd/experimental/webhook-conversion-repro/populator.go
mykysha: Done

cmd/experimental/webhook-conversion-repro/repro.sh
```diff
@@ -0,0 +1,397 @@
...
+    echo "Skipping automated log measurement."
+    RESULT=1
+  else
+    # Wait for apiserver to restart (up to 180s), then watch for compaction storm.
+    # With gate=false the serial conversion takes longer than the compaction
+    # interval, so etcd compacts the revision mid-relist → Kueue retries → compacted
+    # again → infinite loop. With gate=true the parallel worker pool finishes all
+    # conversions well within the window → clean sync.
+    echo "Waiting for apiserver to restart..."
+  CAME_BACK=0
```
mbobrovskyi: nit: It looks like not aligned

cmd/experimental/webhook-conversion-repro/main.go
```diff
@@ -0,0 +1,224 @@
...
+
+	go func() {
+		ticker := time.NewTicker(10 * time.Second)
+		defer ticker.Stop()
+		for range ticker.C {
+			s := atomic.LoadInt32(&success)
+			f := atomic.LoadInt32(&failed)
+			fmt.Printf("\r\033[KProgress: %d / %d (Failed: %d)", s, cfg.Count, f)
+		}
+	}()
```
mbobrovskyi, line 192: The goroutine of progress never stops.

cmd/experimental/webhook-conversion-repro/repro.sh
mykysha: Aligned now, thanks

cmd/experimental/webhook-conversion-repro/main.go
mykysha: Added a closing channel

mimowo:
> /hold
> Maybe @vladikkuzn or @mszadkow could double check that.

@vladikkuzn @mszadkow have you had a chance to play with the repro script?

mimowo:
FYI we have this effort already ongoing in api-server: https://github.com/kubernetes/enhancements/pull/6179
it would be good to have the tooling to prove that 1.37 works better than before

mykysha:
I have compared the script running locally both with and without the concurrent object decode feature-gate, and according to the results, based on the recommended run configuration from the readme, throughput has increased from ~1000 workloads with the feature disabled to ~2500 with it enabled, showing a ~2.5x improvement with the highly restricted etcd configuration. Will try to run the script with the less restricted etcd (10s -> 30s compaction interval for example) to see whether this improvement is consistent or compaction-interval-dependent and will share the results.

Another observation that has occurred in running the test locally would be that with the feature gate enabled, the population stage uses way more ram so the worker/qps configuration had to be lowered to account for that. IIUC there is no way to restrict the number of decode workers, right?

cc @mimowo @Jefftree

Jefftree:
>  showing a ~2.5x improvement with the highly restricted etcd configuration

Nice. built-in resources show less improvement (1.7x ish) so it's cool to see that the effect is more noticeable when adding a conversion webhook. https://github.com/kubernetes/kubernetes/pull/139679

> Another observation that has occurred in running the test locally would be that with the feature gate enabled, the population stage uses way more ram so the worker/qps configuration had to be lowered to account for that. IIUC there is no way to restrict the number of decode workers, right?

Are you saying that write throughput decreases with the gate enabled in the population stage? If so that's a regression we need to be careful about. If this is just an observation about increased resource consumption, it is expected that CPU and memory will be higher because higher throughput. So if we're observing lower throughput that raises some questions. 

>  IIUC there is no way to restrict the number of decode workers, right?

Correct, we are not intending for this option to be configurable, though we do want to tune it for the best user experience.

mykysha:
> Are you saying that write throughput decreases with the gate enabled in the population stage? If so that's a regression we need to be careful about. If this is just an observation about increased resource consumption, it is expected that CPU and memory will be higher because higher throughput. So if we're observing lower throughput that raises some questions.

If we compare gate-enabled and gate-disabled population using the same worker and qps values, throughput stays the same-ish, did not test deeper to draw conclusions. But given the higher cpu and memory usage, the populator should account for that, so qps and worker count had to be lowered in order to adjust for the new available resources.

mykysha:
I have been testing this script against a cloud cluster on scale of 50k single container workloads. With the default etcd compaction interval of 5 minutes some runs failed and on average it took 4 minutes and 47 seconds to load workloads, close to the 5 min failure mark.

This shows that running the script against a cloud cluster adds additional constraints and significantly reduces the possible throughput compared to local runs, where using the same etcd and workload configuration allowed me to operate 200k workloads without failing and the local failing threshold is even higher than that.

cc @Jefftree @mimowo

mykysha:
Creating a cloud cluster using k8s alpha and enabling ConcurrentWatchObjectDecode showed some serious improvements. The same 50k workloads loaded in just 1 minute and 24 seconds on average on the new cluster, showing a 3.4x improvement compared to a cluster with the feature disabled. If 50k was failing before, the new theoretical limit would be closer to 150k with the ConcurrentWatchObjectDecode, if the load time impact is linear.

cc @Jefftree @mimowo

ivnovakov:
> - it reproduces the problem when ConcurrentWatchObjectDecode is disabled
> - with the same parameters it passes when ConcurrentWatchObjectDecode is enabled

Both confirmed locally on kind v1.36.1 with Kueue v0.15.0, same parameters for both runs:

```
--workloads 600 --qps 2 --workers 2 --containers 60 --envs 20 --etcd-interval 5s
```

Each run populated 600 v1beta1 workloads, then restarted kube-apiserver on a cold cache.

Gate disabled: within ~60s of the restart the apiserver logged 4 etcd compaction events (the relist loop) and the Kueue informers never re-synced. Instability reproduced, matching #9145.

Gate enabled: 0 compaction events over the same window and the informers synced normally. Clean recovery, the apiserver comes back and Kueue re-syncs without the relist loop.

So the script distinguishes the two states as intended: instability with the gate off, clean recovery with it on.

Jefftree:
Do we have measurements on the webhook peak load (cpu/memory/etc) change? is kueue memory/cpu capped?

mimowo:
> Do we have measurements on the webhook peak load (cpu/memory/etc) change? is kueue memory/cpu capped?

We could instument Kueue with pprof, good idea - not sure if already implemented

mykysha:
Did not measure resource usage yet - can do and will share the findings

mimowo:
Yeah, similarly as we enable pprof for performance integration tests, this is an example output https://github.com/kubernetes-sigs/kueue/issues/12580

mykysha:
Sharing the measurements across multiple runs both with the regular cloud cluster and an alpha one with the feature enabled:

With ConcurrentWatchObjectDecode disabled, each restart peaked at 75.4% of a CPU core usage on average (up to 122% at its highest) and burned 102.2 CPU-seconds of work. Average memory allocated per restart was 6.87 GB, and the goroutine count was in the range of 572-583.

With the gate enabled the numbers were higher which was expected: 88.9% average peak CPU, 118.3 CPU-seconds of work per restart, and 8.29 GB allocated per run which translates to roughly 16 to 21% more resource use per run. Goroutines stayed consistent in the range of 591-592.

Will also add this info to the readme.

@mimowo @Jefftree

mimowo:
/lgtm
/approve
/unhold
Thank you 👍 We can follow up in dedicated PRs to make the diffs readable. Thank you @ivnovakov for also trying this out 👍

Jefftree:
Awesome work!

PR #11695, opened 2026-05-27
Extend nominal-first preemptioin to cohort-level nominal quotas in hierarchical setups

### LINKED ISSUE #9466: [Fair Sharing] Extend nominal-first priority to cohort-level nominal quotas in hierarchical setups
**What happened**:

PR #9407 (fixing #9406) ensures that workloads fitting within their ClusterQueue's nominal quota are prioritized in fair sharing admission ordering and preemption, regardless of the CQ's DRS from borrowing on other flavors. 

However, this fix only operates at the CQ level.
In hierarchical cohort setups, the same class of bug exists at the cohort level. A cohort has its own nominal quota in addition to it's ClusterQueues' quota, and descendants of that cohort should have preferential access to resources within that cohort's nominal allocation. Today, even with #9407, the DRS comparison at an ancestor cohort can deprioritize a workload whose entire subtree is within nominal — solely because of borrowing on a different flavor.

The nominal-first principle should apply at every level of the cohort hierarchy: if a workload's usage (including the incoming workload) stays within the nominal quota of its ancestor cohort for the contested flavor-resources, that workload should be preferred at that level of comparison — regardless of the ancestor's aggregate DRS from borrowing on other flavors.

Setup — two flavors (h100-reserved, a10-spot), hierarchical cohort tree:

```
Root
├── Cohort1 (own nominalQuota: 40 h100-reserved GPUs — shared pool for its members)
│   ├── CQ-A  (nominal: 10 h100-reserved, using 8; borrowing 200 a10-spot from Root)
│   └── CQ-B  (nominal: 20 h100-reserved, using all 20 — fully utilized, nothing to lend)
│
└── Cohort2 (no own quota)
    └── CQ-C  (nominal: 30 h100-reserved, using 28)
```

CQ-A submits a workload needing 5 h100-reserved GPUs:
- CQ-A usage becomes 13, exceeding CQ-A's nominal of 10 — #9407 does not apply.
- CQ-B is fully utilizing its 20 GPUs — there is nothing to borrow from CQ-B.
- The extra 3 GPUs come from Cohort1's own shared pool of 40.
- Cohort1 SubtreeQuota = 40 (own) + 10 (CQ-A) + 20 (CQ-B) = 70. Cohort1 usage = 13 + 20 = 33, within 70. Cohort1 is not borrowing from Root.

CQ-C submits a workload needing 5 h100-reserved GPUs:
- CQ-C usage becomes 33, exceeding CQ-C's nominal of 30.
- Cohort2 SubtreeQuota = 30 (CQ-C only, no own quota). Usage = 33 > 30. Cohort2 must borrow from Root.

Observed: At the Root-level comparison, Cohort1 has higher DRS (inflated by 200 a10-spot borrowing), so CQ-C's workload is ordered first — even though CQ-C's cohort needs to borrow from Root while CQ-A's cohort is using its own shared pool.

Expected: CQ-A's workload should be preferred. CQ-A is using resources from Cohort1's own shared pool — not borrowing from CQ-B (which has nothing to spare), and not borrowing from Root. Cohort1's a10-spot borrowing should not block its members from accessing Cohort1's own h100-reserved quota.

The same issue applies to fair sharing preemption: if CQ-A needs to preempt a workload from CQ-C to fit, the DRS strategy gate may block the preemption because Cohort1's DRS exceeds Cohort2's DRS — even though Cohort1 is within its own nominal quota for the contested resource.

This was identified by @pajakd [during review of PR #9407](https://github.com/kubernetes-sigs/kueue/pull/9407#issuecomment-3943059429):
> This PR looks good but will work for using nominal quota of CQ. But in more complex hierarchies, there could be a nominal quota of a cohort for which the descendants of that cohort should have a preferential access, right?
>
> CQ1 -> Cohort1 -> Root
> CQ2 -> Cohort2 -> Root
>
> In the above example if CQ1 and CQ2 were competing for nominal quota in Cohort1, then CQ1 should be preferred, regardless if it is borrowing more in some other flavor.

**Environment**: 

Kubernetes version (use kubectl version): v1.33.5
Kueue version (use git describe --tags --dirty --always): v0.15.2
Cloud provider or hardware configuration: Azure Kubernetes Service
OS (e.g: cat /etc/os-release): Linux

mukund-wayve:
#### What type of PR is this?

/kind bug

Extends the CQ level nominal quota prioritisation to hierarchical cohorts even when DRS is high. An [earlier PR](https://github.com/kubernetes-sigs/kueue/pull/10149) already handled admission ordering. This also handles preemptions which is the final part required to close the issue.

Fixes #9466 

```release-note
FairSharing: In fair sharing preemption, bypass DRS strategy gates when the preemptor ClusterQueue or an ancestor cohort is within subtree nominal on contested resources, allowing preemption even if the preemptor CQ has high aggregate DRS from borrowing on other flavors.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fair-sharing preemption decisions across hierarchical queue cohorts.
  * Prevented workloads in protected nominal cohorts from being preempted unnecessarily.
  * Corrected preemption reasons for workloads reclaimed within the same cohort.

* **Tests**
  * Added coverage for sibling queues under a shared nominal cohort.
  * Updated expected preemption behavior and reasons for affected scenarios.

### DIFF
--- a/pkg/scheduler/preemption/preemption.go
+++ b/pkg/scheduler/preemption/preemption.go
@@ -379,14 +379,21 @@ func runFirstFsStrategy(preemptionCtx *preemptionCtx, candidates []*workload.Inf
 	var targets []*Target
 	var retryCandidates []*workload.Info
 
-	// If the preemptor CQ stays within nominal quota for the contested
-	// resources (including the incoming workload, already simulated),
-	// preemption is allowed regardless of DRS (nominal entitlement).
-	// When true, all cross-CQ candidates are preempted unconditionally
-	// (bypassing the strategy check), so no retryCandidates are produced
-	// and runSecondFsStrategy has nothing to do.
-	preemptorWithinNominal := features.Enabled(features.FairSharingPreemptWithinNominal) &&
-		queueWithinNominalInResourcesNeedingPreemption(preemptionCtx)
+	// FairSharingPreemptWithinNominal uses a two-tier bypass:
+	// 1. Preemptor CQ within nominal on contested FRs → global bypass (#9494).
+	// 2. Closest ancestor cohort within nominal → bypass only for candidates
+	//    outside that cohort subtree (prevents sibling flapping, #9466).
+	var preemptorCQWithinNominal bool
+	var protectionCohort *schdcache.CohortSnapshot
+	if features.Enabled(features.FairSharingPreemptWithinNominal) {
+		frs := preemptionCtx.frsNeedPreemption
+		cq := preemptionCtx.preemptorCQ
+		if schdcache.IsWithinNominalInResources(cq, frs) {
+			preemptorCQWithinNominal = true
+		} else {
+			protectionCohort = closestAncestorWithinNominal(cq, frs)
+		}
+	}
 	for candCQ := range ordering.Iter() {
 		if candCQ.InClusterQueuePreemption() {
 			candWl := candCQ.PopWorkload()
@@ -402,7 +409,7 @@ func runFirstFsStrategy(preemptionCtx *preemptionCtx, candidates []*workload.Inf
 			continue
 		}
 
-		if preemptorWithinNominal {
+		if preemptorCQWithinNominal || (protectionCohort != nil && !cqInCohortSubtree(candCQ.GetTargetCq(), protectionCohort)) {
 			candWl := candCQ.PopWorkload()
 			preemptionCtx.snapshot.RemoveWorkload(candWl)
 			targets = append(targets, &Target{
@@ -656,19 +663,22 @@ func queueUnderNominalInResourcesNeedingPreemption(preemptionCtx *preemptionCtx)
 	return true
 }
 
-// queueWithinNominalInResourcesNeedingPreemption checks whether the preemptor
-// or any ancestor cohort is within subtree nominal quota for all contested
-// flavor-resources. This allows fair-sharing preemption to bypass DRS when a
-// CQ borrows from its parent cohort's pool while that cohort is not borrowing
-// upward.
-func queueWithinNominalInResourcesNeedingPreemption(preemptionCtx *preemptionCtx) bool {
-	frs := preemptionCtx.frsNeedPreemption
-	cq := preemptionCtx.preemptorCQ
-	if schdcache.IsWithinNominalInResources(cq, frs) {
-		return true
-	}
+// closestAncestorWithinNominal returns the first (lowest) ancestor cohort that is
+// within subtree nominal on all contested flavor-resources, or nil if none is
+// found before reaching root.
+func closestAncestorWithinNominal(cq *schdcache.ClusterQueueSnapshot, frs sets.Set[resources.FlavorResource]) *schdcache.CohortSnapshot {
 	for ancestor := range cq.PathParentToRoot() {
 		if schdcache.IsWithinNominalInResources(ancestor, frs) {
+			return ancestor
+		}
+	}
+	return nil
+}
+
+// cqInCohortSubtree returns true if cohort appears on candidateCQ's path to root.
+func cqInCohortSubtree(candidateCQ *schdcache.ClusterQueueSnapshot, cohort *schdcache.CohortSnapshot) bool {
+	for ancestor := range candidateCQ.PathParentToRoot() {
+		if ancestor == cohort {
 			return true
 		}
 	}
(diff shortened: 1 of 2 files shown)


pkg/scheduler/preemption/preemption.go
```diff
@@ -656,18 +656,23 @@ func queueUnderNominalInResourcesNeedingPreemption(preemptionCtx *preemptionCtx)
 	return true
 }
 
-// queueWithinNominalInResourcesNeedingPreemption checks whether the
-// preemptor CQ's usage is at or below nominal quota (usage <= nominal)
-// for all flavor-resources needing preemption.
-// The difference from queueUnderNominalInResourcesNeedingPreemption is
-// that this treats usage exactly equal to nominal as "within nominal."
+// queueWithinNominalInResourcesNeedingPreemption checks whether the preemptor
```
mimowo: I think this is a complex change to the complex feature, so we need a FG.

Either a new one or roll the change under FairSharingPrioritizeNonBorrowing. Wdyt?

mukund-wayve: This is actually enabled only when `FairSharingPreemptWithinNominal` is set which was introduced [here](https://github.com/kubernetes-sigs/kueue/pull/9494/changes#diff-71e3b98f9a6bbf5b8421e26a7ba0c079f397cd8d49abacdad943c66a4f44f03dR61). The linked PR enabled preemption when a CQ was within nominal despite high-DRS from other flavors. This change extends it to hierarchical cohorts as suggested in [the review](https://github.com/kubernetes-sigs/kueue/pull/9494#pullrequestreview-3860098859).

mimowo: Makes sense, thank you for explaining. Let me see if @pajakd or @gabesaba would have cycles to review

pkg/scheduler/preemption/preemption.go
```diff
@@ -656,18 +656,23 @@ func queueUnderNominalInResourcesNeedingPreemption(preemptionCtx *preemptionCtx)
...
+// or any ancestor cohort is within subtree nominal quota for all contested
+// flavor-resources. This allows fair-sharing preemption to bypass DRS when a
+// CQ borrows from its parent cohort's pool while that cohort is not borrowing
+// upward.
```
pajakd: This is tricky. We would like to bypass the DRS comparison between preemptor and candidate if the cohort to which preemptor belongs is not borrowing. But in the following case:
`preemptor` -> CQ1  -> cohort1
`candidate` -> CQ2  -> cohort1
what if CQ1 is borrowing, CQ2 is borrowing and cohort1 is **not borrowing**. 

In that case we should compare the DRS values of preemptor and candidate, because
1. they both have the same "right to use" the quota at cohort1.
2. if we preempt the `candidate`, and it comes back to the queue it will be allowed to preempt the preemptor -- we are going to have an infinite preemption cycle.

Lets say we have a (smallest) subtree of the cohort hierarchy sT in which the preemptor fits within a nominal quota. Basically root of this subtree is the node at which you return true in this function. Then, it makes a difference if the candidate is from that subtree (we should compare the DRS) or its outside (we may skip that comparison as we are "protected" by the nominal quota).  

So, its (in my opinion) not a true/false signal for the preemptor, but rather for each candidate.

EDIT: I had a similar type of problems when working on the classical preemption. There I divided the candidates into two lists depending on where the preemptor and candidates are in the hierarchy (https://github.com/kubernetes-sigs/kueue/blob/a393561c2072b1b1366a0bd5ec915be897219373/pkg/scheduler/preemption/classical/hierarchical_preemption.go#L149-L174)

pkg/scheduler/preemption/preemption.go
```diff
@@ -656,18 +656,23 @@ func queueUnderNominalInResourcesNeedingPreemption(preemptionCtx *preemptionCtx)
...
+	if schdcache.IsWithinNominalInResources(cq, frs) {
+		return true
+	}
+	for ancestor := range cq.PathParentToRoot() {
+		if schdcache.IsWithinNominalInResources(ancestor, frs) {
+			return true
 		}
 	}
-	return true
+	return false
```
pajakd, line 685: I don't think the root cohort can borrow from anywhere, so it might be impossible to reach here, no? Unless I'm missing something this function will always return true.

pkg/scheduler/preemption/preemption.go
mimowo: /hold
To avoid accidental merge, we need to determine if the fix does not have "gaps" which will later be hard to iron out.

mukund-wayve:
This is now fixed by https://github.com/kubernetes-sigs/kueue/pull/13307

PR #11701, opened 2026-05-27
fix(kueueviz): check error before null data on detail pages

### LINKED ISSUE #11700: Detail pages show infinite spinner instead of error when WebSocket fails
**What happened**:
When the WebSocket connection fails on the workload detail or local queue detail pages, the page gets stuck on a loading spinner forever. The error message never appears.

This happens because the component checks whether data has loaded before it checks for errors. If the connection fails, data stays empty so the spinner keeps rendering and the error check is never reached.

The page should show the error message when the WebSocket connection fails instead of spinning indefinitely.

1. Run the kueueviz frontend locally
2. Navigate to any workload detail or local queue detail page without a backend running
3. The page shows a spinner forever instead of the connection error

The fix is to check for errors before checking for missing data so the error message renders when the connection fails. Happy to open a PR if this is not being worked on.

YadavAkhileshh:
#### What type of PR is this?
/kind bug
/area dashboard

Currently, the `WorkloadDetail.jsx` and `LocalQueueDetail.jsx` components check if the resource is null/undefined to render a loading spinner *before* checking if there is an active fetch or WebSocket error.

Because a connection failure prevents the resource state from ever being populated, it remains null indefinitely. Consequently, the component remains stuck in an infinite `CircularProgress` loading state, and the user is never shown the error boundary page.

This PR swaps the conditional logic to evaluate error states first, ensuring the error banner displays correctly upon fetch failure.

Fixes #11700

None. Extremely simple logic swap.

Yes

```release-note
KueueViz: Fixed a bug where workload and local queue detail pages would get stuck on a loading spinner instead of showing connection/fetching error messages on failure.
```

### DIFF
--- a/cmd/kueueviz/frontend/src/LocalQueueDetail.jsx
+++ b/cmd/kueueviz/frontend/src/LocalQueueDetail.jsx
@@ -42,9 +42,9 @@ const LocalQueueDetail = () => {
     if (workloadsData) setWorkloads(workloadsData);
   }, [workloadsData]);
 
-  if (!queue) return <CircularProgress />;
   if (queueError) return <ErrorMessage error={queueError} />;
   if (workloadsError) return <ErrorMessage error={workloadsError} />;
+  if (!queue) return <CircularProgress />;
 
   return (
     <Paper className="parentContainer">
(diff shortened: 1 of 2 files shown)


kannon92:
The template seems odd still.

mimowo:
I fixed the release-note block by manually injecting the ``` block

mimowo:
/approve
/cherrypick release-0.17
/cherrypick release-0.16
Thank you for KueueViz hardening 👍

PR #11711, opened 2026-05-28
fix(kueueviz): prevent flashing empty states on list pages

### LINKED ISSUE #11708: kueueviz: List pages flash empty state message briefly on load
**What happened**:
When navigating to any of the list pages (ClusterQueues, Cohorts, LocalQueues, ResourceFlavors, Workloads) in KueueViz, the UI briefly renders the "No resources found" empty state message on every page load or refresh. A split second later, once the WebSocket connection completes and receives the data, the message disappears and is replaced by the actual resource table.

This happens because the initial array state is set to `[]`. Before the WebSocket connects and streams the data, the render function instantly detects a list length of `0` and renders the empty state UI, causing a jarring visual flash.

The UI should display a loading indicator (like a spinner) while waiting for the WebSocket to establish a connection and stream initial data, rather than prematurely declaring the list is empty.

1. Run KueueViz locally.
2. Navigate to the "Cluster Queues" or "Workloads" page.
3. Refresh the page. You will see a flash of "No Cluster Queues found" or "No Workloads found" before the actual rows populate.

The fix is to differentiate between the initial uninitialized state (raw WebSocket data is `null`) and a genuinely empty list (data is fetched but contains 0 items). We should show a loading spinner when data is `null`, and only render the empty list text when data is received and empty. Happy to open a PR for this!

YadavAkhileshh:
#### What type of PR is this?
/kind bug
/area dashboard

In `ClusterQueues.jsx`, `Cohorts.jsx`, `LocalQueues.jsx`, `ResourceFlavors.jsx`, and `Workloads.jsx`, the components initialized state as an empty array `[]`.

Before the WebSocket establishes a connection and streams down the initial cluster data, the render conditions immediately evaluated a length of 0, resulting in a flash of "No resources found" or similar empty-state banners. Once the stream delivered the resource list, the empty-state banner disappeared and the table rendered. This created a highly jarring visual flash on every refresh.

This PR differentiates between the uninitialized state (raw WebSocket data is `null`) and a genuinely empty list (WebSocket data is fetched and has 0 items). We show a clean `CircularProgress` loading spinner during the initialization window, and only display the empty text if the list is confirmed empty.

Fixes #11708

None.

Yes

```release-note
KueueViz: Fixed a bug where list pages briefly flashed "No resources found" empty state messages before data finished loading over the WebSocket connection.
```

### DIFF
--- a/cmd/kueueviz/frontend/src/ClusterQueues.jsx
+++ b/cmd/kueueviz/frontend/src/ClusterQueues.jsx
@@ -25,12 +25,9 @@ import UsageBar, { aggregateResourceForQueue, computeEffectiveQuota, discoverRes
 
 const ClusterQueues = () => {
   const { data: clusterQueues, error } = useWebSocket('/ws/cluster-queues');
-  const [queues, setQueues] = useState([]);
-
-  useEffect(() => {
-    if (clusterQueues && Array.isArray(clusterQueues)) {
-      setQueues([...clusterQueues].sort((a, b) => (a.name || '').localeCompare(b.name || '')));
-    }
+  const queues = React.useMemo(() => {
+    if (!clusterQueues || !Array.isArray(clusterQueues)) return [];
+    return [...clusterQueues].sort((a, b) => (a.name || '').localeCompare(b.name || ''));
   }, [clusterQueues]);
 
   const resourceNames = discoverResourceNames(queues);
--- a/cmd/kueueviz/frontend/src/ClusterQueues.jsx
+++ b/cmd/kueueviz/frontend/src/ClusterQueues.jsx
@@ -40,7 +40,11 @@ const ClusterQueues = () => {
   return (
     <Paper className="parentContainer">
       <Typography variant="h4" gutterBottom>Cluster Queues</Typography>
-      {queues.length === 0 ? (
+      {clusterQueues === null ? (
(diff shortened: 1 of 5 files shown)


cmd/kueueviz/frontend/src/ClusterQueues.jsx
```diff
@@ -40,7 +40,11 @@ const ClusterQueues = () => {
   return (
     <Paper className="parentContainer">
       <Typography variant="h4" gutterBottom>Cluster Queues</Typography>
-      {queues.length === 0 ? (
+      {clusterQueues === null ? (
```
mimowo, line 40: Is it feasible to add unit tests for the fixes?

mimowo:
@YadavAkhileshh seems like the KueueViz e2e test failure is related to the code changes, ptal

YadavAkhileshh:
@mimowo, pushed a commit that refactors all 5 list components to use `useMemo` for sorting instead of `useState`/`useEffect`. 

This completely removes the one-frame delay that caused a brief empty-state flash, which resolves the Cypress E2E failure.

Since the frontend doesn't have a component testing library (like `@testing-library/react`), writing unit tests for rendering is not feasible. However, this logic is fully covered by the Cypress E2E tests, which will now pass successfully!

mimowo:
Excellent, thank you for the summary 👍

Issue #10852, opened 2026-04-30
Inadmissible workloads metrics


j-skiba:
**What would you like to be added**:
Currently we have two metrics for monitoring pending workloads related to local and cluster queues: `kueue_pending_workloads` and `kueue_local_queue_pending_workloads` (see [metrics documentation](https://kueue.sigs.k8s.io/docs/reference/metrics/)). Both of them use one of 2 values of `status` label: `active` or `inadmissible`.

There are two problems with this:
1. Workloads with missing/non-existing local queue defined in `kueue.x-k8s.io/local-queue-name` label are not counted in `status=inadmissible` label.
2. `Reason`s defined in workloads' `.Status.Conditions` are kind of misleading compared to `status=inadmissible` label defined in the mentioned metrics. In [API](https://github.com/kubernetes-sigs/kueue/blob/9a434bf1e825875a5af480515288385dfff3a511/apis/kueue/v1beta2/workload_types.go#L1004) we have `Inadmissible` defined as "LocalQueue or ClusterQueue doesn't exist or inactive" which relates to the problem no. 1.

It would be beneficial for Kueue users to define alerting based on metrics in case of wrong configuration of the workload itself so manual checking of each workload would not be necessary.

Examples for alerts:
- A workload points to a non-existing `LocalQueue` in `kueue.x-k8s.io/local-queue-name` label
- A workload points to correct `LocalQueue` name but that `LocalQueue` points to non-existing `ClusterQueue`
- A workload is defined with incorrect tolerations, meaning no `ResourceFlavor` can match it.

Initially, I considered simply counting such misconfigured workloads under `status=inadmissible`. However, this isn't very helpful because `status=inadmissible` would become a massive, general group of 'not runnable' workloads, making further investigation difficult.

Furthermore, including workloads that were not initially counted changes the existing behavior of the metric. Despite it sounding natural to count these as `status=inadmissible`, the drawbacks led me to my current proposal.

I'm leaning towards another solution which would be a new set of metrics:
- `kueue_queued_workloads`
- `kueue_local_queue_queued_workloads`

Both of these would use `reason` label and the reason label would map 1:1 to the `Reason` field of `.Status.Conditions` of a workload. The only problem I currently see with this approach is that workloads that are not admitted due to flavor mismatch are  currently given `Pending` reason - the same as workloads that are waiting for resources / cannot fit.

[...]

PR #11720, opened 2026-05-28
KEP: Improve observability of inadmissible workloads

j-skiba:
#### What type of PR is this?
/kind kep

Currently, Kueue marks workloads that fail admission due to configuration or structural errors (such as resource flavor mismatches) with a generic `Pending` reason. This makes it difficult for operators, dashboards, and programmatic tools to distinguish between a workload that is simply waiting for cluster capacity versus one that is completely misconfigured.

To resolve this, the proposal introduces:
*   **Tiered Priority Reasons:** A strict 6-tier hierarchy for the `QuotaReserved` condition (when `False`) to separate structural bugs (e.g., `Misconfigured`, `ExceedsMaxQuota`) from normal capacity wait states (`WaitingForQuota`).
*   **Granular Metrics:** New Prometheus metrics (`kueue_unadmitted_workloads` and `kueue_local_queue_unadmitted_workloads`) tracking fine-grained reasons and underlying causes for unadmitted workloads.
*   **Explicit Status Initialization:** An optional feature gate to initialize conditions to `False` on the first reconciliation loop for continuous lifetime tracking.
*   **Decoupled Lifecycles:** Simplification of status reasoning by removing the hybrid `NoReservationUnsatisfiedChecks` reason.

These improvements will allow for precise alerting, better dashboards, and automated recovery tasks by programmatic clients.

Part of #10852

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Feature-gated observability for inadmissible workloads: tiered Prometheus metrics, clearer admission/reservation reason categories, and a stricter precedence model for reservation-related status.
  * Optional explicit initialization of admission-related conditions to False so missing/unset conditions have defined semantics and metrics reflect explicit states.

* **Documentation**
  * Full KEP with troubleshooting, before/after status examples, test plan, graduation criteria, and design rationale.

### DIFF
--- a/keps/10852-inadmissible-workloads-observability/README.md
+++ b/keps/10852-inadmissible-workloads-observability/README.md
@@ -246,16 +246,19 @@ To separate different blocks and keep the status field highly structured, when
 the `QuotaReserved` status is `False`, the controller will resolve the reason
 field according to a strict priority hierarchy. Lower tier numbers represent
 structural failures and block scheduling early, thus taking absolute precedence
-over higher tiers:
+over higher tiers.
 
-| Tier | Classification | Reason Token | Description / Scenario | Location in Memory |
-| :--- | :--- | :--- | :--- | :--- |
-| **Tier 1** | Workload Deactivation | `Deactivated` | Workload is explicitly deactivated (`spec.active: false`). | None |
-| **Tier 2** | Structural Blockers | `Misconfigured` | Permanent structural error preventing admission (e.g., non-existent/inactive LocalQueue or ClusterQueue, resource flavor mismatch, DRA misconfiguration). | Missing local/cluster queue - None<br>ResourceFlavor mismatch - `ClusterQueue.inadmissibleWorkloads`<br>DRA issues - None |
-| **Tier 3** | Structural Quota Blockers | `ExceedsMaxQuota` | Workload requests resource quantities exceeding ClusterQueue or Cohort maximum limits. | `ClusterQueue.inadmissibleWorkloads` |
-| **Tier 4** | Orchestration & Administrative Holds | `Suspended`, `AdmissionGated`, `WaitingForPodsReady` | Workload is structurally valid, but admission is halted due to:<br>- Administrative hold (the queue's StopPolicy is active).<br>- Gated state (AdmissionGatedBy annotation).<br>- Scheduling hold (waiting for previously admitted workloads to reach PodsReady under waitForPodsReady configuration). | AdmissionGated - None<br>StopPolicy - None<br>WaitingForPodsReady - blocks and waits |
-| **Tier 5** | Resource Deficits | `WaitingForQuota` | Workload fits within the maximum limits of the ClusterQueue or Cohort, but the cluster currently lacks sufficient unreserved capacity (including when preemption is required but is blocked by preemption gates). | `ClusterQueue.inadmissibleWorkloads` |
-| **Tier 6** | Active Queueing | `PendingEvaluation` | Workload is submitted, structurally valid, resides actively in the queue, and is awaiting its capacity evaluation. Set also after eviction. | `ClusterQueue.heap` |
+Tiers are evaluated either at the **workload-wide** level (independent of flavor
+assignments) or at the **per-flavor** assignment level:
+
+| Tier | Scope | Classification | Reason Token | Description / Scenario | Location in Memory |
+| :--- | :--- | :--- | :--- | :--- | :--- |
+| **Tier 1** | Workload-wide | Workload Deactivation | `Deactivated` | Workload is explicitly deactivated (`spec.active: false`). | None |
+| **Tier 2** | Mixed | Structural Blockers | `Misconfigured` | Permanent structural error preventing admission:<br>- Missing local/cluster queue (Workload-wide)<br>- Resource flavor mismatch (Per-flavor)<br>- DRA misconfiguration (Workload-wide) | Missing local/cluster queue - None<br>ResourceFlavor mismatch - `ClusterQueue.inadmissibleWorkloads`<br>DRA issues - None |
+| **Tier 3** | Per-flavor | Structural Quota Blockers | `ExceedsMaxQuota` | Workload requests resource quantities exceeding ClusterQueue or Cohort maximum limits. | `ClusterQueue.inadmissibleWorkloads` |
+| **Tier 4** | Workload-wide | Orchestration & Administrative Holds | `Suspended`, `AdmissionGated`, `WaitingForPodsReady` | Workload is structurally valid, but admission is halted due to:<br>- Administrative hold (the queue's StopPolicy is active).<br>- Gated state (AdmissionGatedBy annotation).<br>- Scheduling hold (waiting for previously admitted workloads to reach PodsReady under waitForPodsReady configuration). | AdmissionGated - None<br>StopPolicy - None<br>WaitingForPodsReady - blocks and waits |
+| **Tier 5** | Per-flavor | Resource Deficits | `WaitingForQuota` | Workload fits within the maximum limits of the ClusterQueue or Cohort, but the cluster currently lacks sufficient unreserved capacity (including when preemption is required but is blocked by preemption gates). | `ClusterQueue.inadmissibleWorkloads` |
+| **Tier 6** | Workload-wide | Active Queueing | `PendingEvaluation` | Workload is submitted, structurally valid, resides actively in the queue, and is awaiting its capacity evaluation. Set also after eviction. | `ClusterQueue.heap` |
 
 *Note: Precedence Resolution examples:*
 - *If a workload requests resource quantities exceeding the ClusterQueue
@@ -268,29 +271,16 @@ over higher tiers:
 
 #### Multi-Flavor Precedence Resolution
 
-When a workload is evaluated against multiple `ResourceFlavor` paths
-within a ClusterQueue, different flavors may yield different admission blocks.
-For example:
-- **Flavor A**: The workload has a structural affinity or taint mismatch
-  (resolves to Tier 2 `Misconfigured`).
-- **Flavor B**: The workload is structurally compatible but requests resource
-  quantities that exceed the ClusterQueue or Cohort maximum limits (resolves
-  to Tier 3 `ExceedsMaxQuota`).
-
-In scenarios where multiple flavors are checked:
-1. A structurally compatible flavor path (even if it currently has quota or
-   limits deficits like Tier 3 `ExceedsMaxQuota` or Tier 5 `WaitingForQuota`)
-   is a closer candidate for scheduling than a structurally incompatible
-   flavor path (Tier 2 `Misconfigured`).
-2. Therefore, when combining diagnostics across multiple flavors, the reasons
-   from compatible flavor paths (e.g., `ExceedsMaxQuota` or `WaitingForQuota`)
-   take precedence and are reported in the status, instead of reporting the
-   structural incompatibility (`Misconfigured`) of the other flavors.
+When a workload is evaluated against multiple `ResourceFlavor` paths within a
+ClusterQueue, Kueue attempts to find the most schedulable assignment. If
+scheduling fails, the blocker tier and reason reported in the workload status
+are emitted from the most schedulable flavor assignment candidate (i.e., the
+candidate that blocks at the highest tier number).
 
 This ensures that operators receive actionable alerts regarding the closest
 viable scheduling path (e.g., a limits or capacity block on a compatible
-flavor) rather than getting generic configuration mismatch reports from
-incompatible flavors.
+flavor) rather than generic configuration mismatch reports from incompatible
+flavors.
 
 ### Admitted Condition Initialization and Lifecycle
 
--- a/keps/10852-inadmissible-workloads-observability/README.md
+++ b/keps/10852-inadmissible-workloads-observability/README.md
@@ -253,18 +253,18 @@ over higher tiers:
 | **Tier 1** | Workload Deactivation | `Deactivated` | Workload is explicitly deactivated (`spec.active: false`). | None |
 | **Tier 2** | Structural Blockers | `Misconfigured` | Permanent structural error preventing admission (e.g., non-existent/inactive LocalQueue or ClusterQueue, resource flavor mismatch, DRA misconfiguration). | Missing local/cluster queue - None<br>ResourceFlavor mismatch - `ClusterQueue.inadmissibleWorkloads`<br>DRA issues - None |
 | **Tier 3** | Structural Quota Blockers | `ExceedsMaxQuota` | Workload requests resource quantities exceeding ClusterQueue or Cohort maximum limits. | `ClusterQueue.inadmissibleWorkloads` |
-| **Tier 4** | Orchestration & Administrative Holds | `Suspended`, `AdmissionGated`, `WaitingForPodsReady` | Workload is structurally valid, but admission is halted due to:<br>- Administrative hold (workload.spec.active is true or the queue's StopPolicy is active).<br>- Gated state (AdmissionGatedBy annotation).<br>- Scheduling hold (waiting for previously admitted workloads to reach PodsReady under waitForPodsReady configuration). | AdmissionGated - None<br>StopPolicy - None<br>WaitingForPodsReady - blocks and waits |
+| **Tier 4** | Orchestration & Administrative Holds | `Suspended`, `AdmissionGated`, `WaitingForPodsReady` | Workload is structurally valid, but admission is halted due to:<br>- Administrative hold (the queue's StopPolicy is active).<br>- Gated state (AdmissionGatedBy annotation).<br>- Scheduling hold (waiting for previously admitted workloads to reach PodsReady under waitForPodsReady configuration). | AdmissionGated - None<br>StopPolicy - None<br>WaitingForPodsReady - blocks and waits |
 | **Tier 5** | Resource Deficits | `WaitingForQuota` | Workload fits within the maximum limits of the ClusterQueue or Cohort, but the cluster currently lacks sufficient unreserved capacity (including when preemption is required but is blocked by preemption gates). | `ClusterQueue.inadmissibleWorkloads` |
 | **Tier 6** | Active Queueing | `PendingEvaluation` | Workload is submitted, structurally valid, resides actively in the queue, and is awaiting its capacity evaluation. Set also after eviction. | `ClusterQueue.heap` |
 
->*Note: Precedence Resolution examples:*
->- *If a workload requests resource quantities exceeding the ClusterQueue
->  maximum limits (Tier 3) but is also blocked by an unsatisfied
->  admission gate (Tier 4), the resolved reason is written as
->  `ExceedsMaxQuota` due to Tier 3 priority.*
->- *If a workload is waiting for capacity (Tier 5) but also has unsatisfied
->  admission gates (Tier 4), the resolved reason is written as
->  `AdmissionGated` due to Tier 4 priority.*
+*Note: Precedence Resolution examples:*
+- *If a workload requests resource quantities exceeding the ClusterQueue
+  maximum limits (Tier 3) but is also blocked by an unsatisfied
+  admission gate (Tier 4), the resolved reason is written as
+  `ExceedsMaxQuota` due to Tier 3 priority.*
+- *If a workload is waiting for capacity (Tier 5) but also has unsatisfied
+  admission gates (Tier 4), the resolved reason is written as
+  `AdmissionGated` due to Tier 4 priority.*
 
 #### Multi-Flavor Precedence Resolution
 
@@ -366,7 +366,10 @@ tracking:
 This ensures that pending workloads are fully tracked from the moment of their
 creation, regardless of whether explicit status initialization is active.
 
-The exact label mapping matrix is detailed below:
+When the `reason` label is `NoReservation` (the workload lacks a quota
+reservation), the `underlying_cause` label maps 1:1 to the `QuotaReserved`
+condition's reason (covering all 6 precedence tiers). Representative examples
+of these mapping combinations are detailed below:
 
 | Admitted Reason | QuotaReserved Reason | `reason` Label | `underlying_cause` Label | Description / Scenario |
 | :--- | :--- | :--- | :--- | :--- |
@@ -500,7 +503,7 @@ status:
 ### Test Plan
 
 [x] I/we understand the owners of the involved components may require updates to
-existing tests to make this code solid enough prior to committing the changes
+existing tests to make this code solid enough before committing the changes
 necessary to implement this enhancement.
 
 #### Unit Tests
--- a/keps/10852-inadmissible-workloads-observability/README.md
+++ b/keps/10852-inadmissible-workloads-observability/README.md
@@ -19,6 +19,7 @@
   - [Upgrade Path](#upgrade-path)
   - [Downgrade Path](#downgrade-path)
 - [Design Details](#design-details)
+  - [API Changes](#api-changes)
   - [Tiered Reason Precedence &amp; Resolution](#tiered-reason-precedence--resolution)
     - [1. Flavor-Independent (Workload-wide) Reasons](#1-flavor-independent-workload-wide-reasons)
     - [2. Nominated Flavor Reasons](#2-nominated-flavor-reasons)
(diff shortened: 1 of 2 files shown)


keps/10852-inadmissible-workloads-observability/README.md
```diff
@@ -0,0 +1,446 @@
...
+### Non-Goals
+
+- Modifying the underlying workload admission state machine or scheduling
+  decisions.
+- Automatically correcting or mutating workload specs when configurations are
+  invalid.
+- Exposing scheduling queue internals beyond standard condition transitions and
+  aggregated metrics.
+
+## Proposal
```
mwielgus, line 145: Could you add a couple of yaml snippets with before/after?

j-skiba: added `Concrete Status Scenarios (Before & After)` section

keps/10852-inadmissible-workloads-observability/README.md
```diff
@@ -0,0 +1,551 @@
...
+
+To ensure safe operational rollouts, the functionality is divided behind two
+separate feature gates:
+- `UnadmittedWorkloadsObservability`: Controls the overall feature.
+  It enables the granular Prometheus metrics (`kueue_unadmitted_workloads`
+  and `kueue_local_queue_unadmitted_workloads`) and updates status reasons for
+  the `QuotaReserved` and `Admitted` conditions during reconciliations and
+  scheduler cycles to use the new tiered reasons instead of the generic
+  "Pending".
+- `UnadmittedWorkloadsExplicitStatus`: Controls whether both status conditions
```
amy: This sentence is difficult to parse. I dont understand what it is gating or what it holistically means.

keps/10852-inadmissible-workloads-observability/README.md
```diff
@@ -0,0 +1,551 @@
...
+controllers. With the new `kueue_unadmitted_workloads` metric, we can define
+alerts targeting the `Misconfigured` underlying cause, notifying tenants to fix
+their configurations immediately.
+
+#### Story 2: Platform Team Diagnostics and Dashboards
+
+As a platform team, we want to construct a unified Grafana dashboard to track
+queue queueing characteristics in real-time. We need to distinguish between
+workloads waiting for normal quota capacity (a standard queueing event) versus
+workloads blocked because they exceed maximum bounds or fail administrative holds.
```
amy: What do these mean "exceed maximum bounds or fail administrative holds". Please provide examples.

j-skiba: I meant a workload that requests above nominalQuota + borrowingLimit or a workload that is created in a cluster/local queue that has StopPolicy set.

I updated the description.

keps/10852-inadmissible-workloads-observability/README.md
j-skiba: I updated the description

mimowo:
/assign
I plan to review this week, probably early feedback today or tomorrow. I already endorse the direction of the changes, thank you for working on that one 👍

kshalot:
/assign

Same here - I want to review this as soon as I have the capacity (likely by EOD tomorrow).

keps/10852-inadmissible-workloads-observability/kep.yaml
```diff
@@ -0,0 +1,38 @@
...
+  - "@kshalot"
+  - "@mimowo"
+  - "@mwielgus"
+  - "@amy"
+approvers:
+  - "@mimowo"
+  - "@tenzen-y"
+
+# The target maturity stage in the current dev cycle for this KEP.
+stage: beta
```
amy, line 17: Just checking. Is this beta or alpha?

keps/10852-inadmissible-workloads-observability/README.md
```diff
@@ -0,0 +1,561 @@
...
+*Note: Precedence Resolution examples:*
+- *If a workload requests resource quantities exceeding the ClusterQueue
+  maximum limits (Tier 3) but is also blocked by an unsatisfied
+  admission gate (Tier 4), the resolved reason is written as
+  `ExceedsMaxQuota` due to Tier 3 priority.*
+- *If a workload is waiting for capacity (Tier 5) but also has unsatisfied
+  admission gates (Tier 4), the resolved reason is written as
+  `AdmissionGated` due to Tier 4 priority.*
+
+#### Multi-Flavor Precedence Resolution
```
amy: You might want to specify that some tiers are computer per flavor whereas others are workload-wide. 

Separately, in place of this section: `In scenarios where multiple flavors are checked:` you're ultimately picking the most schedulable flavor and emitting the blocker tier from the most schedulable? (I think?)

If that's the true, simplify the entire section.

j-skiba: That's correct,

- updated the tiers table to include scope (per-flavor, workload-wide, mixed) of the reason
- simplified `Multi-Flavor Precedence Resolution`

keps/10852-inadmissible-workloads-observability/kep.yaml
j-skiba: It's beta for v0.19 and would be backported to v0.18 as opt-in.

I also updated description of `UnadmittedWorkloadsExplicitStatus` FG - previously it was saying the flag will be disabled by default but it's not true for beta. I removed that from description. The graduation plan is already under `### Graduation Criteria` and mentions enabling both flags by default.

j-skiba:
based on these two issues:
- https://github.com/kubernetes-sigs/kueue/issues/11843
- https://github.com/kubernetes-sigs/kueue/issues/11842
I updated the KEP and added `BorrowingLimitReached`, `TopologyPlacementFailed` and `PreemptionPending` reasons. The tiers were updated too.

/cc @varunsyal

keps/10852-inadmissible-workloads-observability/kep.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. This can be the current (upcoming) milestone, if it is being actively
+# worked on.
+latest-milestone: "v0.19"
+
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  beta: "v0.19"
+  stable: "v0.20"
```
mimowo: Leave it as TBD

keps/10852-inadmissible-workloads-observability/README.md
```diff
@@ -0,0 +1,553 @@
...
+```yaml
+status:
+  conditions:
+  # QuotaReserved condition is ONLY initialized/present on creation if the
+  # UnadmittedWorkloadsExplicitStatus gate is ENABLED. If disabled, it is absent.
+  - type: QuotaReserved
+    status: "False"
+    reason: PendingEvaluation # Tier 8 Reason: Awaiting initial scheduler evaluation
+    message: "Workload is pending evaluation in the scheduling queue"
+    observedGeneration: 1
```
mimowo: Drop the `observedGeneration` from the examples - it does not help, but eats space.

keps/10852-inadmissible-workloads-observability/README.md
```diff
@@ -0,0 +1,553 @@
...
+| **Tier 5** | Per-flavor | Policy Quota Limits | `BorrowingLimitReached` | Total unused capacity in the Cohort is sufficient, but allocating resources would cause the ClusterQueue to exceed its borrowing limit policy constraint (`val > NominalQuota + BorrowingLimit`). | `ClusterQueue.inadmissibleWorkloads` |
+| **Tier 6** | Per-flavor | Resource Deficits | `WaitingForQuota`, `TopologyPlacementFailed` | Parallel capacity-wait reasons:<br>- `WaitingForQuota`: General capacity wait (total unused nominal capacity in the ClusterQueue/Cohort is less than the workload request: `val > available`).<br>- `TopologyPlacementFailed`: TAS workload only. Nominal capacity exists (`val <= available`), but capacity is fragmented across domains preventing contiguous placement. | `ClusterQueue.inadmissibleWorkloads` |
+| **Tier 7** | Workload-wide | Admission Progressing | `PreemptionPending` | Evictions have been issued; waiting for victims to terminate and release their capacity. | None |
```
mimowo: I think PendingPreemption is rather per-flavor because we might be pending preemptions in one flavor, but in the next scheduling cycle we may potentially choose to pick another flavor. This means the preemption may get wasted, but this is how Kueue operates.

keps/10852-inadmissible-workloads-observability/README.md
```diff
@@ -0,0 +1,553 @@
...
+structural failures and block scheduling early, thus taking absolute precedence
+over higher tiers.
+
+Tiers are evaluated either at the **workload-wide** level (independent of flavor
+assignments) or at the **per-flavor** assignment level:
+
+| Tier | Scope | Classification | Reason Token | Description / Scenario | Location in Memory |
+| :--- | :--- | :--- | :--- | :--- | :--- |
+| **Tier 1** | Workload-wide | Workload Deactivation | `Deactivated` | Workload is explicitly deactivated (`spec.active: false`). | None |
+| **Tier 2** | Mixed | Structural Blockers | `Misconfigured` | Permanent structural error preventing admission:<br>- Missing local/cluster queue (Workload-wide)<br>- Resource flavor mismatch (Per-flavor)<br>- DRA misconfiguration (Workload-wide) | Missing local/cluster queue - None<br>ResourceFlavor mismatch - `ClusterQueue.inadmissibleWorkloads`<br>DRA issues - None |
```
mimowo: Should we rather split that into two sub-groups: Miconfigured (global), and Miconfigured (per-flavor)? 

Maybe the per-flavor we should called just FlavorMismatch.

The reason is that they should probably be in different tiers. Consider that we have two flavors:
Flavor A is FlavorMismatch
Flavor B is PendingPreemptions

Then we should probably report PendingPreemptions because this is the "closest to make it runnable".

keps/10852-inadmissible-workloads-observability/README.md
```diff
@@ -0,0 +1,553 @@
...
+| **Tier 8** | Workload-wide | Active Queueing | `PendingEvaluation` | Workload is submitted, structurally valid, resides actively in the queue, and is awaiting its capacity evaluation. Set also after eviction. | `ClusterQueue.heap` |
+
+*Note: Precedence Resolution examples:*
+- *If a workload requests resource quantities exceeding the ClusterQueue
+  maximum limits (Tier 3) but is also blocked by an unsatisfied
+  admission gate (Tier 4), the resolved reason is written as
+  `ExceedsMaxQuota` due to Tier 3 priority.*
+- *If a workload is waiting for capacity (Tier 6) but also has unsatisfied
+  admission gates (Tier 4), the resolved reason is written as
+  `AdmissionGated` due to Tier 4 priority.*
```
mimowo: I think here the rules should be more formalized, what I'm thinking about:

1. Any cluster-wide blocker is reported at the workload level, the highest (PendingEvaluation only is not a blocker)
2. If we only have per-flavor reasons then we report the highest tier -the one that is the closest to make the workload runnable
3. only if we don't have any blockers, and any per-flavor blockers then we report PendingEvaluation

Let me know if I'm missing something.

mimowo (commented):
Overall, the KEP is in good direction. I think the tiers and the rules still require a bit of care. Posted comments to clarify the remaining doubts.

mimowo:
I synced with @j-skiba regarding the yesterday's comments, and we can address them by the following updates:
1. rename "per flavor" as "nominated flavor" to avoid suggesting that the property is computed for each flavor
2. split the table into three categories (flavor independent, nominated flavor, pending evaluation), make sure that all workload-wide are above "nominated flavor"
3. for the current "Misconfigured" FlavorMismatch create sub-category like "NoMatchingFlavor"
4. update the rules to say which one is reported when there are multiple PodSets, potentially choosing other flavors - it should choose the worst tier for the workload

keps/10852-inadmissible-workloads-observability/kep.yaml
j-skiba: done

keps/10852-inadmissible-workloads-observability/README.md
j-skiba: done

keps/10852-inadmissible-workloads-observability/README.md
j-skiba: I agree, I split the reasons and changed the scope as described here https://github.com/kubernetes-sigs/kueue/pull/11720#issuecomment-4667939302

keps/10852-inadmissible-workloads-observability/README.md
j-skiba: I added new reasons precedence rules and reason groups

keps/10852-inadmissible-workloads-observability/README.md
j-skiba: new precedence rules were described in the newest change as described here https://github.com/kubernetes-sigs/kueue/pull/11720#issuecomment-4667939302

mimowo (commented):
/lgtm
/approve
Thank you, I think implementing this KEP will greatly improve the observability.

Currently, it is a mess, primarily because:
- there is no granularity for the QuotaReserved reasons
- the term "inadmissible workloads" is overloaded - workloads for non-existing queues have `reason: Inadmissible`, but are not included in the `kueue_pending_workloads { status=inadmissible}`

The KEP solves both issues 👍 

/hold
in case @tenzen-y would like to review

keps/10852-inadmissible-workloads-observability/README.md
```diff
@@ -0,0 +1,644 @@
...
+scheduler. Lower numbers represent more severe blocks.
+
+| Precedence | Reason Token | Description / Scenario | Location in Memory |
+| :--- | :--- | :--- | :--- |
+| **1** | `ExceedsMaxQuota` | Workload requests resource quantities exceeding ClusterQueue nominal limits (when no cohort) or the Cohort's total maximum capacity limits. | `ClusterQueue.inadmissibleWorkloads` |
+| **2** | `BorrowingLimitReached` | Total unused capacity in the Cohort is sufficient, but allocating resources would cause the ClusterQueue to exceed its borrowing limit policy constraint. | `ClusterQueue.inadmissibleWorkloads` |
+| **3** | `WaitingForQuota` | General capacity wait (total unused nominal capacity in the ClusterQueue/Cohort is less than the workload request). | `ClusterQueue.inadmissibleWorkloads` |
+| **4** | `TopologyPlacementFailed` | TAS workload only. Nominal capacity exists, but capacity is fragmented across domains preventing contiguous placement. | `ClusterQueue.inadmissibleWorkloads` |
+| **5** | `PendingPreemption` | Evictions have been issued; waiting for victims to terminate and release their capacity. | `ClusterQueue.heap` |
```
j-skiba: Maybe it would be better to change this to something like `WaitingForPreemptedWorkloads`? I'm afraid  `PendingPreemption` might be misleading that the workload with this reason is being preempted.

mimowo: Good point, I think the newly proposed name is better than the previous.

j-skiba: renamed

kshalot (commented):
/lgtm

I did a few passes, but nothing jumps out at me, LGTM! 👍

kshalot:
/hold

But I see that @mimowo had a bunch of comments so I'll hold just in case he needs another pass.

tenzen-y (commented):
@j-skiba Could you also add API proposal?
IIUC, you will add condition reasons to API package like https://github.com/kubernetes-sigs/kueue/blob/1ba1b7fc37246aa40c756e2ebe5b661d1346c1d7/apis/kueue/v1beta2/workload_types.go#L975-L1104

mimowo:
cc @sohankunkerkar

keps/10852-inadmissible-workloads-observability/README.md
```diff
@@ -0,0 +1,644 @@
...
+If a workload defines multiple podsets, flavor assignment is evaluated
+sequentially by podset groups.
+
+If a podset group fails to be assigned a flavor, the scheduler exits early,
+skipping subsequent groups. Consequently, the representative reason for the
+workload is resolved as the most severe reason among the first failing podset
+group's reason and any preceding groups' reasons (as shown in Example 2
+below).
+
+#### Resolution Algorithm
```
sohankunkerkar, line 359: This assumes every workload goes through the scheduler for reason evaluation. But with `SchedulingEquivalenceHashing` enabled (default), workloads whose hash is in `noFitSchedulingHashes` are inserted directly into inadmissible, bypassing the scheduler entirely. These workloads keep whatever reason they had from their last full evaluation until `queueInadmissibleWorkloads` clears the hash set. Worth documenting this as a known staleness window for the reason field.

keps/10852-inadmissible-workloads-observability/README.md
```diff
@@ -0,0 +1,644 @@
...
+##### Scenario 1: Newly Created Workload (Initial Reconcile)
+
+Newly submitted workloads in the scheduling queue awaiting their first evaluation.
+
+* **Before KEP-10852:** Both conditions are completely absent from the status:
+```yaml
+status: {}
+```
+
+* **After KEP-10852:** Explicitly updated status:
```
sohankunkerkar, line 532: When the controller restarts, it will reconcile all existing workloads. The KEP should clarify that initialization only happens when the condition is absent, not on every reconcile. Otherwise a restart would overwrite existing conditions and reset their timestamps.

keps/10852-inadmissible-workloads-observability/README.md
```diff
@@ -0,0 +1,644 @@
...
+scheduler. Lower numbers represent more severe blocks.
+
+| Precedence | Reason Token | Description / Scenario | Location in Memory |
+| :--- | :--- | :--- | :--- |
+| **1** | `ExceedsMaxQuota` | Workload requests resource quantities exceeding ClusterQueue nominal limits (when no cohort) or the Cohort's total maximum capacity limits. | `ClusterQueue.inadmissibleWorkloads` |
+| **2** | `BorrowingLimitReached` | Total unused capacity in the Cohort is sufficient, but allocating resources would cause the ClusterQueue to exceed its borrowing limit policy constraint. | `ClusterQueue.inadmissibleWorkloads` |
+| **3** | `WaitingForQuota` | General capacity wait (total unused nominal capacity in the ClusterQueue/Cohort is less than the workload request). | `ClusterQueue.inadmissibleWorkloads` |
+| **4** | `TopologyPlacementFailed` | TAS workload only. Nominal capacity exists, but capacity is fragmented across domains preventing contiguous placement. | `ClusterQueue.inadmissibleWorkloads` |
+| **5** | `WaitingForPreemptedWorkloads` | Evictions have been issued; waiting for victims to terminate and release their capacity. | `ClusterQueue.heap` |
```
sohankunkerkar: Note: Between issuing evictions and victims actually terminating, the scheduler's next cycle may see the same workload as `WaitingForQuota` (quota not yet freed) rather than `WaitingForPreemptedWorkloads`. The reason can oscillate until victims release quota. Worth noting this as expected behavior or defining which reason takes precedence during this transition window.

keps/10852-inadmissible-workloads-observability/README.md
```diff
@@ -0,0 +1,644 @@
...
+  admission checks.
+- Once quota reservation is successfully obtained (`QuotaReserved` status is
+  `True`), if admission checks are still pending, the `Admitted` condition
+  will transition its reason to `UnsatisfiedChecks`.
+
+### Prometheus Metrics Schema
+
+A set of metrics is introduced to track unadmitted workloads when the
+`Admitted` condition status is `False`. The metrics are:
+- `kueue_unadmitted_workloads`: Tracks unadmitted workloads at the
```
sohankunkerkar, line 455: Does this overlap with the `existing kueue_pending_workloads` metric?

j-skiba: yes, here's the root issue https://github.com/kubernetes-sigs/kueue/issues/10852

keps/10852-inadmissible-workloads-observability/README.md
j-skiba: good catch, added `*Note on Scheduling Equivalence Hashing (Reason Staleness)*:`

keps/10852-inadmissible-workloads-observability/README.md
j-skiba: I updated the description of `### Admitted Condition Initialization and Lifecycle` to include this

keps/10852-inadmissible-workloads-observability/README.md
j-skiba: I added `*Note on preemption transition dynamics*` below the table and left the order of the reasons as it is

j-skiba:
> @j-skiba Could you also add API proposal? IIUC, you will add condition reasons to API package like
> 
> https://github.com/kubernetes-sigs/kueue/blob/1ba1b7fc37246aa40c756e2ebe5b661d1346c1d7/apis/kueue/v1beta2/workload_types.go#L975-L1104

@tenzen-y that's correct.

I added `### API Changes` section.

keps/10852-inadmissible-workloads-observability/README.md
```diff
@@ -0,0 +1,715 @@
...
+	// because no resource flavor matches its nodeSelector or taints.
+	WorkloadNoMatchingFlavor = "NoMatchingFlavor"
+
+	// WorkloadExceedsMaxQuota indicates that the workload requests resources
+	// exceeding the maximum capacity limits of the ClusterQueue or Cohort.
+	WorkloadExceedsMaxQuota = "ExceedsMaxQuota"
+
+	// WorkloadBorrowingLimitReached indicates that allocating resources would
+	// cause the ClusterQueue to exceed its borrowing limit constraint.
+	WorkloadBorrowingLimitReached = "BorrowingLimitReached"
```
tenzen-y: Why do we need to have these 2 reasons? It seems that "BorrowingLimitReached" is a subset of "ExceedsMaxQuota". Why do we specialize only "BorrowingLImit"?

If we want to tell this "borrowing" based reasons, shouldn't we handle this as underlying cause like "ExceedsMaxQuotaDueToBorrowingLimitReached"?

Otherwise, either "ExceedsMaxQuota" or "BorrowingLimitReached" should not be introduced.

We have 3 approaches to avoid overlapping between reasons:

1. Introduce the `BorrowingLimitReached` as an underlying cause for the `ExceedsMaxQuota` which means `ExceedsMaxQuotaDueToBorrowingLimitReached`.
2. Introduce only `ExceedsMaxQuota`
3. Introduce only `BorrowingLimtReached` and other reasons for quota exceeding.

keps/10852-inadmissible-workloads-observability/README.md
```diff
@@ -0,0 +1,715 @@
...
+	WorkloadSuspended = "Suspended"
+
+	// WorkloadPendingEvaluation indicates that the workload is pending evaluation in the scheduling queue.
+	WorkloadPendingEvaluation = "PendingEvaluation"
+
+	// WorkloadAdmittedReasonNoReservation indicates that the workload has no reservation.
+	WorkloadAdmittedReasonNoReservation = "NoReservation"
+
+	// WorkloadAdmittedUnsatisfiedChecks indicates that the workload has not all checks ready.
+	WorkloadAdmittedUnsatisfiedChecks = "UnsatisfiedChecks"
```
tenzen-y:
```suggestion
	// WorkloadAdmittedUnsatisfiedChecks indicates that the workload has not all checks ready.
	WorkloadAdmittedUnsatisfiedAdmissionChecks = "UnsatisfiedAdmissionChecks"
```

The API should be descriptive. The `Checks` is a common word. So, we should use Kueue Kueue-specific `AdmissionCheck` terminology here.

keps/10852-inadmissible-workloads-observability/README.md
j-skiba: Good point.

My initial approach to make them two separate reasons was that `ExceedsMaxQuota` was a reason that needs administrative actions to resolve while `BorrowingLimitReached` could be resolved when some other workloads finish their work.

Now I'm wondering if we even benefit from having `BorrowingLimitReached`, I think we could simply use `WaitingForQuota` instead. This way we don't introduce some kind of overlapping and keep the number of reasons shorter. WDYT?

keps/10852-inadmissible-workloads-observability/README.md
j-skiba: sounds good, renamed

keps/10852-inadmissible-workloads-observability/README.md
mimowo: +1 for reusing `WaitingForQuota`.

tenzen-y: +1 for reusing `WaitingForQuota`. We can revisit more detailed reasons like borrowing limit, fair sharing, and more... with `DueTo` style or another way.

tenzen-y (commented):
Thank you!
/lgtm
/approve

/hold cancel

PR #11721, opened 2026-05-28
Fix ObjectRetentionPolicy for orphaned Workloads.

mbobrovskyi:
#### What type of PR is this?
/kind bug

Fix ObjectRetentionPolicy for orphaned Workloads.

Fixes #

```release-note
ObjectRetentionPolicies: Fixed a bug that doesn't allow the deletion of orphaned finished workloads.
```

### DIFF
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -224,7 +224,11 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 
 	if isOrphanedWorkload(&wl) {
 		err := workload.FinalizeOrphanedWorkload(ctx, r.client, r.clock, &wl, true)
-		return ctrl.Result{}, err
+		// If it was deleted, there is nothing to do.
+		// Otherwise, we still need to handle finished workload logic.
+		if !features.Enabled(features.FinishOrphanedWorkloads) || !wl.DeletionTimestamp.IsZero() {
+			return ctrl.Result{}, err
+		}
 	}
 
 	if features.Enabled(features.MultiKueueOrchestratedPreemption) {
--- a/pkg/controller/core/workload_controller_test.go
+++ b/pkg/controller/core/workload_controller_test.go
@@ -2798,6 +2798,28 @@ func TestReconcile(t *testing.T) {
 				},
 			},
 		},
+		"should handle finished workload logic for orphaned workloads when FinishOrphanedWorkloads enabled": {
+			featureGates: map[featuregate.Feature]bool{features.FinishOrphanedWorkloads: true},
+			workload: utiltestingapi.MakeWorkload("wl", "ns").
+				Finalizers(kueue.ResourceInUseFinalizerName).
+				FinishedAt(now).
+				Obj(),
+			reconcilerOpts: []Option{
+				WithWorkloadRetention(
+					&workloadRetentionConfig{
+						afterFinished: new(util.MediumTimeout),
+					},
+				),
+			},
+			wantWorkload: utiltestingapi.MakeWorkload("wl", "ns").
+				Finalizers(kueue.ResourceInUseFinalizerName).
+				FinishedAt(now).
+				Obj(),
+			wantError: nil,
+			wantResult: reconcile.Result{
+				RequeueAfter: util.MediumTimeout,
+			},
+		},
 		"shouldn't delete the workload because, object retention not configured": {
 			workload: utiltestingapi.MakeWorkload("wl", "ns").
 				Condition(metav1.Condition{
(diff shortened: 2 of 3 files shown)


pkg/controller/core/workload_controller.go
```diff
@@ -224,7 +224,10 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 
 	if isOrphanedWorkload(&wl) {
 		err := workload.FinalizeOrphanedWorkload(ctx, r.client, r.clock, &wl, true)
-		return ctrl.Result{}, err
+		// If it was deleted, nothing to do.
+		if !wl.DeletionTimestamp.IsZero() {
+			return ctrl.Result{}, err
+		}
```
mimowo: Can you explain a bit why this is fixing the bug? Seemingly it has nothing to do with ObjectRetentionPolicy. 

Please also update PR description with the notes to reviewers to explain. This is far from obvious.

mbobrovskyi: We have the ObjectRetentionPolicy logic a bit later. This logic allows us to handle it. Otherwise, if we just return nil, we will never handle the ObjectRetentionPolicy logic.

mimowo: Ah, I see could we maybe move that logic up so that it is next to handling orphan workloads, they both seem very similar conceptualy? Or you wanted to minize diff?

mbobrovskyi: Yeah, I tried to move it up, but I ran into other issues in the unit tests. We probably still need to handle the workload finishing logic.

mbobrovskyi: I think if we want to move this up, we need to split the logic for orphaned workloads. We still need to handle orphan workloads where `!wl.DeletionTimestamp.IsZero()`, even if the workload is already finished.

mbobrovskyi: Or maybe we should move the logic for removing finalizers into the Finish logic.

pkg/controller/core/workload_controller.go
```diff
@@ -224,7 +224,11 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 
 	if isOrphanedWorkload(&wl) {
 		err := workload.FinalizeOrphanedWorkload(ctx, r.client, r.clock, &wl, true)
-		return ctrl.Result{}, err
+		// If it was deleted, there is nothing to do.
```
mimowo, line 227: Let's also cover that by unit test for TestReconcile for the ease of testing

pkg/controller/core/workload_controller.go
mimowo: I don't understand this "// Otherwise, we still need to handle the ObjectRetentionPolicy logic." 

Could you somehow clarify that, I'm not sure which code your are talking about, tried to search the file by ObjectRetentionPolicy, but didn't find anything.

mbobrovskyi: Imagine we have a workload that was already finished by the job reconciler. In that case, the workload still has finalizers that should be removed once it becomes orphaned. However, we can't remove them because the current logic for finishedWorkloads (https://github.com/kubernetes-sigs/kueue/blob/38166941f5cb274ae1643e86e6541dd7538d8ae5/pkg/controller/core/workload_controller.go#L250-L280) prevents it.

So I see two possible solutions: either we split the orphaned workload logic (separating deleted and non-deleted cases), or we move finalizer removal to the finish logic.

mimowo: I still don't understand, sorry.

> However, we can't remove them because the current logic for finishedWorkloads (...) prevents it

- "them" I assume is finalizers
- "it" I have no clue

Also, can you point a line which prevents "it"? Sorry, I'm really confused.

mbobrovskyi: > Also, can you point a line which prevents "it"? Sorry, I'm really confused.

Yes, this one (https://github.com/kubernetes-sigs/kueue/blob/38166941f5cb274ae1643e86e6541dd7538d8ae5/pkg/controller/core/workload_controller.go#L250-L280) prevents the finalizing logic if we move it up.

pkg/controller/core/workload_controller.go
mbobrovskyi: Done

pkg/controller/core/workload_controller.go
mimowo: Ok, I synced with @mbobrovskyi and there seem to be two aporaches to fix it more cleanly:
1. split FinalizeOrphanedWorkload
2. remove finalizers from within Finish

I think I'm ok with both approaches, maybe slightly leaning to (1.) as it allows more modularity

mimowo:
Thank you 👍 Merging as is since the fix looks sensible and has been exercised by @mbobrovskyi . We may re-think  alternative ways to organize the workload Finish and Finalize in the future.
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.16

tenzen-y:
/release-note-edit

```release-note
ObjectRetentionPolicies: Fixed a bug that doesn't allow the deletion of orphaned finished workloads.
```

PR #8877, opened 2026-01-29
promote multikueue batch job to GA

kannon92:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
promote multikueue batch job to GA
#### Which issue(s) this PR fixes:

Fixes #

#### Special notes for your reviewer:
https://github.com/kubernetes-sigs/kueue/issues/8855#issuecomment-3812483101
#### Does this PR introduce a user-facing change?

```release-note
MultiKueue: Promote MultiKueueBatchJobWithManagedBy to stable
```

PR #6445, opened 2025-08-04
Introduce multi-Kueue support for ElasticJobs, with specific support for `batch/v1.Job` workloads when ElasticJob functionality is enabled.

ichekrygin:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

Now that the core ElasticJobsViaWorkloadSlices functionality has been introduced in Kueue as part of [KEP-77](https://github.com/kubernetes-sigs/kueue/tree/main/keps/77-dynamically-sized-jobs), this PR introduces MultiKueue support for ElasticJobsViaWorkloadSlices.

#### Which issue(s) this PR fixes:

Fixes #6335 

#### Does this PR introduce a user-facing change?

```release-note
Add: Multikueue support for ElasticJobsViaWorklaodSlices with `batchv1/Job` support.
```

PR #11729, opened 2026-05-28
Remove a duplicated unit test case in job_multikueue_adapter_test.go

olekzabl:
#### What type of PR is this?

/kind cleanup
/area multikueue

It removes [this test case](https://github.com/kubernetes-sigs/kueue/blob/67d6dbd04055aea01ca74436615f98e2c9aa40a6/pkg/controller/jobs/job/job_multikueue_adapter_test.go#L405-L422) which is currently exactly identical to [this one](https://github.com/kubernetes-sigs/kueue/blob/67d6dbd04055aea01ca74436615f98e2c9aa40a6/pkg/controller/jobs/job/job_multikueue_adapter_test.go#L368-L385) (except the description).

How did that happen? Originally (in #6445 ), they differed by the value of a feature gate: explicitly true

https://github.com/kubernetes-sigs/kueue/blob/831b2d6fa0bbede10648dfee74aaa6b765f32685/pkg/controller/jobs/job/job_multikueue_adapter_test.go#L432-L436

vs. implicitly false

https://github.com/kubernetes-sigs/kueue/blob/831b2d6fa0bbede10648dfee74aaa6b765f32685/pkg/controller/jobs/job/job_multikueue_adapter_test.go#L474-L475

When promoting FG to stable (in #8877 ), setting FG to true was removed but the other test case remained intact, and so became identical.

none

```release-note
NONE
```

### DIFF
--- a/pkg/controller/jobs/job/job_multikueue_adapter_test.go
+++ b/pkg/controller/jobs/job/job_multikueue_adapter_test.go
@@ -402,24 +402,6 @@ func Test_multiKueueAdapter_SyncJob(t *testing.T) {
 					Obj(),
 			},
 		},
-		"RemoteJobInProgress_LocalIsNotManaged": {
-			args: args{
-				localClient: fake.NewClientBuilder().WithScheme(schema).WithObjects(newJob().Obj()).Build(),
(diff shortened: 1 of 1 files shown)


mimowo:
/lgtm
/approve
Thanks for the careful check 👍

PR #11730, opened 2026-05-28
Fix the flaky MultiKueue integration test

mimowo:
#### What type of PR is this?
/kind cleanup
/kind flake

Prepare for https://github.com/kubernetes-sigs/kueue/pull/10936

When retrying the tests I encountered the flake, when the `status.Active=0` instead of the expected.

However, this makes because MultiKueue does not guarantee all status updates will be synced. This may happen if we update the worker while the manager is suspended. 

This issue was possible before the PR but very unlikely, because it is racy in nature. Now it is more likely because it takes more time to start syncing the status, due to the extra validation in the core k8s.

The fix is to just plumb the test making sure we start updating the worker only once the manager is ready for it - so we are sure our update is not skipped.

```release-note
NONE
```

### DIFF
--- a/test/integration/multikueue/jobs_test.go
+++ b/test/integration/multikueue/jobs_test.go
@@ -348,6 +348,14 @@ var _ = ginkgo.Describe("MultiKueue", ginkgo.Label("area:multikueue", "feature:m
 			}, util.Timeout, util.Interval).Should(gomega.Succeed())
 		})
 
+		ginkgo.By("waiting for the Job on the manager to be unsuspended", func() {
+			gomega.Eventually(func(g gomega.Gomega) {
+				createdJob := batchv1.Job{}
+				g.Expect(managerTestCluster.client.Get(managerTestCluster.ctx, client.ObjectKeyFromObject(job), &createdJob)).To(gomega.Succeed())
+				gomega.Expect(createdJob.Spec.Suspend).To(gomega.Equal(new(false)))
+			}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
+		})
+
 		ginkgo.By("updating the worker job status", func() {
 			startTime := metav1.NewTime(time.Now().Truncate(time.Second))
 			gomega.Eventually(func(g gomega.Gomega) {
--- a/test/integration/multikueue/jobs_test.go
+++ b/test/integration/multikueue/jobs_test.go
@@ -352,7 +352,7 @@ var _ = ginkgo.Describe("MultiKueue", ginkgo.Label("area:multikueue", "feature:m
 			gomega.Eventually(func(g gomega.Gomega) {
 				createdJob := batchv1.Job{}
 				g.Expect(managerTestCluster.client.Get(managerTestCluster.ctx, client.ObjectKeyFromObject(job), &createdJob)).To(gomega.Succeed())
-				gomega.Expect(createdJob.Spec.Suspend).To(gomega.Equal(new(false)))
+				g.Expect(createdJob.Spec.Suspend).To(gomega.Equal(new(false)))
 			}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
 		})
 


sohankunkerkar:
Do you have the link to the failure?

olekzabl:
I'm a bit anxious if this doesn't reveal some issue which might also occur in real Kueue operation?

As I understand, the root cause of flakiness was such a scenario:

1. The manager decides to dispatch a job.
2. The job quickly becomes dispatched and running on the worker.
3. The remote workload becomes admitted, this change is quickly detected & reconciled by manager-side MK controllers.
4. Only then, the manager marks the local Job as unsuspended.
5. Only following step 4, `SyncJob()` would copy `Active` (> 0) from remote Job status to local Job status... but it's too late. All the "startup" invocations of `SyncJob()` have already happened, and a new one will happen only on some new *unrelated* change of some object. 
   (Because `SyncJob()` is only called by the MK workload controller, and it only watches local workloads, remote workloads and remote jobs. But not local jobs!)

Now, how realistic is such a scenario? AI told me that it's realistic as long as `WaitForPodsReady` is disabled. Because:

- with `WaitForPodsReady` enabled, unsuspending a local Job leads to setting `PodsReady` condition on the local workload.
- with `WaitForPodsReady` disabled, there may be no such a trigger?

Moreover, this test (according to AI :) has indeed `WaitForPodsReady` disabled.

So, my takeaway is that, as long as `WaitForPodsReady` is disabled, we may have a real bug.

WDYT?

mimowo:
> Do you have the link to the failure?

No, I think I lost it, would need to dig the history of the build here https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/pr-logs/directory/pull-kueue-test-integration-multikueue-main

mimowo:
I think `WaitForPodsReady` is a distraction here, not a good day for AI :) 

However, indeed one can consider it a "bug" - some updates of the Jobs running on the worker may be skipped. This was true almost forever, see this code in 0.17: https://github.com/kubernetes-sigs/kueue/blob/3753b2981c8909e1cd077821f242778b81760481/pkg/controller/jobs/job/job_multikueue_adapter.go#L68

Now I think the issue got slighlty more popular because previously it took only one request on the manager to unblock the sync: (spec.suspend=true -> spec.suspend=false). Now, two requests are required. So, naturally the window of dropped syncs is extended.

Is this a "real problem"? Not really, because:
1.  Job running on a worker usually create Poids, and update their status over probably minutes before it starts the Pods, and "warms up" to reach the steady equilibrium state
2. the terminal condition Failed/Complete will be synced anyway

In summary:
1. Yes, teoretically the "visbility" is compromised, but it always was slightly
2. Now the extra request unblocking the sync makes it more likely to happen in integration tests
3. This is nowhere near (imo) causing real prod issues when the Job status is not synced due to stable running state of a Job - unless the user has a Job with a single Pod, which has a very lightweight image which starts super fast, and then the Job is running in "steady state"
4. the terminal conditions will be synced anyway

We could aim to improve by somehow "memorizing" the skip update and issuing it later, but I'm not sure about ROI for that.

olekzabl:
> I think `WaitForPodsReady` is a distraction here, not a good day for AI :)

For my benefit of learning, why a distraction? This code looks convincing to me:

https://github.com/kubernetes-sigs/kueue/blob/0487ada92bff6b178812c986038a43889ba35b07/pkg/controller/jobframework/reconciler.go#L514-L519

> We could aim to improve by somehow "memorizing" the skip update and issuing it later, but I'm not sure about ROI for that.

I imagine a simpler way could be to somehow "bump" a condition of the Workload once a Job gets unsuspended.
For example, append `The corresponding Job is unsuspended` to the message of the `Admitted` condition unless it's already contained in that message.
And then, naturally MK wl reconciler will be triggered, doing `SyncJob()` one more time.

But if you say the gain to win is low, I won't insist.

---

EDIT: instead of "bumping" the condition and playing with its message, a possibly cleaner way could be to make sure another reconciliation of the local Job is triggered.
And now, it's been independently proposed in #11867.
(Though this was also my thinking, which I forgot to write down here).

mimowo:
> For my benefit of learning, why a distraction?

because the test failed on this assert: 

https://github.com/kubernetes-sigs/kueue/pull/11730/changes#diff-bb6596d076b1ff4bf28f422644861e21a95ff282555c8746809739995addb020R374 (L374)

which asserts a Job status.Ready, not Workload condition.

WairForPodsReady is updating the PodsReady condition on workload, so I cannot see how they are related.

mimowo:
> But if you say the gain to win is low, I won't insist.

We could consider fixing it, but I wouldn't couple it to the flake fix. The issue existed before, just the time window of "no sync" was shorter. I'm ok to fix it. I just wanted to say will not have capacity to fix it in reasonable amount of time, as I'm focusing on the release by tomorrow, and OOO next week. 

I'm not sure how to fix it either,

sohankunkerkar (commented):
LGTM. The fix is correct. It ensures the manager job is unsuspended before updating worker status eliminates the race. One thing I noticed while reviewing: the suspension guard in SyncJob is unique to batch/Job. The other MK adapters (JobSet, RayJob, PyTorchJob, etc.) all unconditionally copy remote status without checking suspension state. So this specific race only affects batch/Job. I agree with the assessment that the production impact is negligible given the self-healing through continued remote Job status updates and the periodic `workerLostTimeout` requeue.                                                                                                                                                                   
                                                                                                                                                                                                            
                                                                                                                                                                             
  /lgtm

test/integration/multikueue/jobs_test.go
```diff
@@ -348,6 +348,14 @@ var _ = ginkgo.Describe("MultiKueue", ginkgo.Label("area:multikueue", "feature:m
 			}, util.Timeout, util.Interval).Should(gomega.Succeed())
 		})
 
+		ginkgo.By("waiting for the Job on the manager to be unsuspended", func() {
+			gomega.Eventually(func(g gomega.Gomega) {
+				createdJob := batchv1.Job{}
+				g.Expect(managerTestCluster.client.Get(managerTestCluster.ctx, client.ObjectKeyFromObject(job), &createdJob)).To(gomega.Succeed())
+				g.Expect(createdJob.Spec.Suspend).To(gomega.Equal(new(false)))
```
mimowo, line 355: I had to fix this line to use `g.Expect` instead of `gomega.Expect`, because it was failing this way: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10936/pull-kueue-test-integration-multikueue-main/2060013791545921536

ptal; https://github.com/kubernetes-sigs/kueue/pull/10936#issuecomment-4566416499

olekzabl:
> > For my benefit of learning, why a distraction?
> 
> because the test failed on this assert:
> 
> https://github.com/kubernetes-sigs/kueue/pull/11730/changes#diff-bb6596d076b1ff4bf28f422644861e21a95ff282555c8746809739995addb020R374 (L374)
> 
> which asserts a Job status.Ready, not Workload condition.
> 
> WairForPodsReady is updating the PodsReady condition on workload, so I cannot see how they are related.

FTR I think they are related, as explained in my [longer comment](https://github.com/kubernetes-sigs/kueue/pull/11730#issuecomment-4565781099), step 5.

> We could consider fixing it, but I wouldn't couple it to the flake fix.

OK

/lgtm
/unhold

PR #11744, opened 2026-05-28
Return error on FinalizeOrphanedWorkload.

mbobrovskyi:
#### What type of PR is this?
/kind bug

Return error on FinalizeOrphanedWorkload.

Follow-up fix https://github.com/kubernetes-sigs/kueue/pull/11721.

```release-note
NONE
```

### DIFF
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -224,6 +224,9 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 
 	if isOrphanedWorkload(&wl) {
 		err := workload.FinalizeOrphanedWorkload(ctx, r.client, r.clock, &wl, true)
+		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/core/workload_controller.go
```diff
@@ -224,6 +224,9 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 
 	if isOrphanedWorkload(&wl) {
 		err := workload.FinalizeOrphanedWorkload(ctx, r.client, r.clock, &wl, true)
+		if err != nil {
```
mimowo, line 227: We need some test which will show the previously made bug is fixed

mbobrovskyi: Totally agree! Added.

mimowo (commented):
/lgtm
/approve
Thank you, please follow up with CPs for both PRs

Issue #11606, opened 2026-05-25
☂️ Bring down CI job times to 18min


mimowo:
**Do not assign the umbrella issue. This place is to set the goal, collect pointers, discuss and track the progress. Assign individual tasks/issues.**

**What would you like to be cleaned**:

I want to bring down the CI testing to 10min. 

Here is the list of CI jobs which are above the threshold currently:

```
| Rank | Job | Avg | Last runtimes | Build IDs | Results |
|---:|---|---:|---|---|---|
| 2 | `periodic-kueue-test-integration-extended-main` | 25:47 | 24:45, 26:08, 26:15, 25:49, 25:56 | 2057294461309816832, 2057113069535891456, 2056931874105724928, 2056750678138687488, 2056569481508950016 | SUCCESS, SUCCESS, SUCCESS, SUCCESS, SUCCESS |
| 4 | `periodic-kueue-test-integration-baseline-main` | 25:02 | 25:47, 25:15, 24:56, 24:19, 24:53 | 2057294461204959232, 2057113069435228160, 2056931874021838848, 2056750678058995712, 2056569481412481024 | SUCCESS, SUCCESS, SUCCESS, SUCCESS, SUCCESS |
| 5 | `periodic-kueue-test-e2e-sequential-baseline-main` | 24:50 | 22:44, 24:38, 24:54, 25:33, 26:19 | 2057165863055790080, 2056984471026012160, 2056803274887008256, 2056622079305846784, 2056440882705469440 | SUCCESS, SUCCESS, SUCCESS, SUCCESS, SUCCESS |
| 6 | `periodic-kueue-test-e2e-tas-extended-main` | 23:34 | 23:21, 24:12, 23:20, 23:55, 23:03 | 2057226513702457344, 2057045121723011072, 2056863673560862720, 2056682477677711360, 2056501029922410496 | SUCCESS, SUCCESS, SUCCESS, SUCCESS, SUCCESS |
| 7 | `periodic-kueue-test-e2e-extended-main-1-35` | 22:59 | 21:58, 23:31, 23:01, 23:42, 22:45 | 2057298236728676352, 2057116850151493632, 2056935648941576192, 2056754452924207104, 2056573256801980416 | SUCCESS, SUCCESS, SUCCESS, SUCCESS, SUCCESS |
[...]

PR #11750, opened 2026-05-28
Move core integrations from extended to baseline

mimowo:
#### What type of PR is this?

/kind cleanup

To move the e2e tests for the native resources (k8s Job and Pod) to the "baseline" suite which does not require installing any CRDs.

This should offload the "extended" e2e test suite, baseline is much faster thanks to no CRD installation.

Part of  #11606

```release-note
NONE
```


mbobrovskyi (commented):
/lgtm

Thank you!

PR #11764, opened 2026-05-28
test: log tolerated controller restarts during e2e tests

Mostafahassen1:
#### What type of PR is this?
/kind bug
/kind flake

Under 4-thread parallel load, the API server delays leader election renewals, causing the MPI operator to fatally restart.
The test suite enforces a strict zero-restart rule, so this single restart immediately aborts the entire e2e test run.

**Solution 1 (Prevent the restart):**
We patch the MPI operator post-install to increase leader election timeouts and API server connection limits.
This gives the operator enough patience to wait for delayed server responses instead of immediately crashing.

**Solution 2 (Tolerate the restart):**
We change the checkNoRestarts flag to false in the test suite specifically for the external MPI operator.
This ensures the test suite tolerates transient startup restarts and only requires the operator to eventually become available.

Fixes #10298

```release-note
NONE
```

## Summary by CodeRabbit

* **Tests**
  * Relaxed deployment readiness checks to tolerate pod container restarts and improve test reliability; restarted containers are now logged during checks rather than causing unconditional failures.

* **Chores**
  * Minor formatting change in installation/test scripts (whitespace-only edit).


hack/testing/e2e-common.sh
```diff
@@ -962,8 +962,30 @@ function install_mpi {
     cluster_kind_load_image "${name}" "${KUBEFLOW_MPI_IMAGE/#v}"
     curl -sSL "${KUBEFLOW_MPI_MANIFEST}" \
         | kubectl apply --kubeconfig="${kubeconfig}" --server-side -f -
-    e2e_wait_for_operator_in_install "${kubeconfig}" "${ns}" "${deployment_name}"
 
+    kubectl patch --kubeconfig="${kubeconfig}" deployment/"${deployment_name}" \
```
mbobrovskyi: Instead, can we apply these changes as a Kustomize patch?

Mostafahassen1:
@mimowo @mbobrovskyi I want support with this issue because I can't solve it.

config/mpi-operator/patch.yaml
```diff
@@ -0,0 +1,16 @@
...
+  namespace: mpi-operator
+spec:
+  template:
+    spec:
+      containers:
+        - name: mpi-operator
+          args:
+            - --leader-elect-lease-duration=60s
+            - --leader-elect-renew-deadline=40s
+            - --leader-elect-retry-period=10s
```
ShaanveerS: where did you find these flags? 
they dont seem defined in mpi operator's: https://github.com/kubeflow/mpi-operator/blob/101db144e019e1de65aaded07745e9d805c2852a/cmd/mpi-operator/app/options/options.go#L53-L87

test/util/e2e.go
```diff
@@ -357,7 +357,7 @@ func WaitForSparkOperatorAvailability(ctx context.Context, k8sClient client.Clie
 
 func WaitForKubeFlowMPIOperatorAvailability(ctx context.Context, k8sClient client.Client) {
 	kftoKey := types.NamespacedName{Namespace: "mpi-operator", Name: "mpi-operator"}
-	waitForDeploymentAvailability(ctx, k8sClient, kftoKey, true)
+	waitForDeploymentAvailability(ctx, k8sClient, kftoKey, false)
```
ShaanveerS: could you add logging here in case any restart occurs?

config/mpi-operator/patch.yaml
ShaanveerS: I think we can skip Solution 1 entirely. The test fails because it treats any MPI operator restart as fatal, even when the operator becomes available again.
We could avoid changing the deployment and just keep the restarts logged.

cc @mbobrovskyi

test/util/e2e.go
```diff
@@ -357,7 +365,7 @@ func WaitForSparkOperatorAvailability(ctx context.Context, k8sClient client.Clie
...
```
mimowo: > @mimowo @mbobrovskyi I want support with this issue because I can't solve it.

I think we can keep the logging added, but we don't need to flip this value for now, because  the issue is likely attributed to api-server getting temporarily overloaded which affected the "leader election", and:
1. since then we split out the heavy KubeRay integration out to a dedicated shard "pull-kueue-test-e2e-tas-extended-shard-0-main", so likely the number of requests to the API server is now lower
2. we haven't seen this issue re-occur in 2months since creation

So, I would suggest to:
1. close the issue for now
2. keep this PR only to log the number of restarts only - still might be informative for this case: https://github.com/kubernetes-sigs/kueue/blob/d968cfde57b256eefc93e564306222446ce92b87/test/util/e2e.go#L608, or future development,

hack/testing/e2e-common.sh
```diff
@@ -962,8 +962,8 @@ function install_mpi {
     cluster_kind_load_image "${name}" "${KUBEFLOW_MPI_IMAGE/#v}"
     curl -sSL "${KUBEFLOW_MPI_MANIFEST}" \
         | kubectl apply --kubeconfig="${kubeconfig}" --server-side -f -
-    e2e_wait_for_operator_in_install "${kubeconfig}" "${ns}" "${deployment_name}"
 
+    e2e_wait_for_operator_in_install "${kubeconfig}" "${ns}" "${deployment_name}"
```
mimowo, line 966: revert unrelated changes

mimowo (commented):
/lgtm
/approve
Thank you for adjusting the PR to only make it logging 👍 

We will re-evaluate the approach when the issue re-occurs. Currently it does not look like impacting us.

mimowo:
/retitle test: log tolerated controller restarts during e2e tests

PR #11770, opened 2026-05-29
Use k8s Job and move to Baseline suite TAS NodeHotSwap tests

### LINKED ISSUE #11751: Re-write TAS NodeHotSwap tests to use k8s Job and move to Baseline suite

I would like to rewrite the existing e2e tests for TAS Node HotSwap to use k8s Job instead of JobSet and move them to the baseline suite.

These test anyway are using only 1 replica, and only change parallelism.

The tests: https://github.com/kubernetes-sigs/kueue/blob/main/test/e2e/tas/extended/hotswap_test.go

**Why is this needed**:

* The "extended" suite for TAS is one of the longest suites, while "baseline" is very fast. This is because extended requires installation of all integrations. This way we can have better time balance.
* Testing locally is faster.

mbobrovskyi:
#### What type of PR is this?
/kind cleanup

Use k8s Job and move to Baseline suite TAS NodeHotSwap tests.

Fixes #11751

```release-note
NONE
```


mimowo:
/lgtm
/approve
/cherrypick release-0.17
/cherrypick release-0.1
I reviewed the diff between the tests and it lgtm: https://www.diffchecker.com/voZtUkxf/
Thank you 👍

mimowo:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/11770/pull-kueue-test-integration-shard-2-main/2060250993957801984
probably flake, opening issue

PR #11791, opened 2026-05-29
Drop spec update RBAC from Kueue manager

### LINKED ISSUE #11270: Drop spec updating from RBAC requests for the kueue-controller-manager

kueue-controller-manager currently is requesting RBAC update to spec which it is not using, and should not be using unless good reasons

Example, for localQueues we have rbac to delete/create/update/patch, but `kueue-controller-manager` should only be updating finalizers and status, see [here](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/core/localqueue_controller.go#L172)

Analogous controls we give to Topology, ResourceFlavor, ClusterQueue etc.

The only exception where we should keep is the update to ClusterQueue as this is going to be temporarily used  (alpha1) by the MultiKueue global quotas. Later it will be dropped too.

Another exception is Workload, but this is a semi-internal CRD.

**Why is this needed**:

* To make it clear that kueue-controller-manager does not update spec for the Kueue CRDs. These are left to be managed by users / admins.
* Currently it is easy to miss during review that a controller may start using update on LQ or CQ, because the RBAC would remain unchanged.

MatteoFari:
#### What type of PR is this?
 /kind cleanup
 /area tas

Reduces Kueue manager RBAC on Kueue API resources to remove unnecessary spec write permissions for configuration resources. Workload broad writes and ClusterQueue update are preserved as explicit exceptions, and finalizer update permissions are kept only for resources whose controllers mutate finalizers.

Fixes #11270

```release-note
Reduced Kueue manager RBAC permissions for several Kueue configuration resources to avoid unnecessary spec write access.
```

### DIFF
--- a/pkg/controller/core/localqueue_controller.go
+++ b/pkg/controller/core/localqueue_controller.go
@@ -169,9 +169,8 @@ func (r *LocalQueueReconciler) NotifyWorkloadUpdate(oldWl, newWl *kueue.Workload
 }
 
 // +kubebuilder:rbac:groups=events.k8s.io,resources=events,verbs=create;watch;update;patch
-// +kubebuilder:rbac:groups=kueue.x-k8s.io,resources=localqueues,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=kueue.x-k8s.io,resources=localqueues,verbs=get;list;watch
 // +kubebuilder:rbac:groups=kueue.x-k8s.io,resources=localqueues/status,verbs=get;update;patch
-// +kubebuilder:rbac:groups=kueue.x-k8s.io,resources=localqueues/finalizers,verbs=update
 
 func (r *LocalQueueReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	var queueObj kueue.LocalQueue
--- a/pkg/controller/core/resourceflavor_controller.go
+++ b/pkg/controller/core/resourceflavor_controller.go
@@ -80,7 +80,7 @@ func (r *ResourceFlavorReconciler) logger() logr.Logger {
 	return roletracker.WithReplicaRole(ctrl.Log.WithName(r.logName), r.roleTracker)
 }
 
-// +kubebuilder:rbac:groups=kueue.x-k8s.io,resources=resourceflavors,verbs=get;list;watch;update;delete
+// +kubebuilder:rbac:groups=kueue.x-k8s.io,resources=resourceflavors,verbs=get;list;watch;update
 // +kubebuilder:rbac:groups=kueue.x-k8s.io,resources=resourceflavors/finalizers,verbs=update
 
 func (r *ResourceFlavorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
--- a/charts/kueue/templates/rbac/role.yaml
+++ b/charts/kueue/templates/rbac/role.yaml
@@ -240,15 +240,11 @@ rules:
     resources:
       - admissionchecks
       - clusterqueues
-      - cohorts
-      - localqueues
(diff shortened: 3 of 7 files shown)


pkg/controller/core/resourceflavor_controller.go
```diff
@@ -96,28 +95,28 @@ func (r *ResourceFlavorReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	if flavor.DeletionTimestamp.IsZero() {
 		// Although we'll add the finalizer via webhook mutation now, this is still useful
 		// as a fallback.
+		patch := client.MergeFromWithOptions(flavor.DeepCopy(), client.MergeFromWithOptimisticLock{})
 		if controllerutil.AddFinalizer(&flavor, kueue.ResourceInUseFinalizerName) {
-			if err := r.client.Update(ctx, &flavor); err != nil {
+			if err := r.client.Patch(ctx, &flavor, patch); err != nil {
```
mimowo: ElWhy changing Update to Patch? My idea for the issue was to just redtrict RBAC to what is currently used, no need to change prod code.
From RBAC perspective it makes very little difference Update vs Patch. Same for Get and List

pkg/controller/core/resourceflavor_controller.go
```diff
@@ -96,28 +95,28 @@ func (r *ResourceFlavorReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	if flavor.DeletionTimestamp.IsZero() {
 		// Although we'll add the finalizer via webhook mutation now, this is still useful
 		// as a fallback.
+		patch := client.MergeFromWithOptions(flavor.DeepCopy(), client.MergeFromWithOptimisticLock{})
```
mszadkow: to avoid wasted deepCopy we could move patch past the if

pkg/controller/tas/topology_controller.go
```diff
@@ -117,17 +116,21 @@ func (r *topologyReconciler) Reconcile(ctx context.Context, req reconcile.Reques
...
+			if err := r.client.Patch(ctx, topology, patch); err != nil {
 				return ctrl.Result{}, err
 			}
 			log.V(5).Info("Removed finalizer")
 		}
-	} else if controllerutil.AddFinalizer(topology, kueue.ResourceInUseFinalizerName) {
-		if err := r.client.Update(ctx, topology); err != nil {
-			return ctrl.Result{}, err
+	} else {
+		patch := client.MergeFromWithOptions(topology.DeepCopy(), client.MergeFromWithOptimisticLock{})
```
mszadkow: same here

pkg/controller/core/localqueue_controller.go
```diff
@@ -169,7 +169,7 @@ func (r *LocalQueueReconciler) NotifyWorkloadUpdate(oldWl, newWl *kueue.Workload
 }
 
 // +kubebuilder:rbac:groups=events.k8s.io,resources=events,verbs=create;watch;update;patch
-// +kubebuilder:rbac:groups=kueue.x-k8s.io,resources=localqueues,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=kueue.x-k8s.io,resources=localqueues,verbs=get;list;watch
 // +kubebuilder:rbac:groups=kueue.x-k8s.io,resources=localqueues/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=kueue.x-k8s.io,resources=localqueues/finalizers,verbs=update
```
mszadkow: this can be removed as well, localqueue controller never calls AddFinalizer or RemoveFinalizer

tenzen-y (commented):
Thank you!
/lgtm
/approve

tenzen-y:
/retest

due to infra

PR #11824, opened 2026-05-29
Ignore charts temp files

### LINKED ISSUE #11819: Release automation: after building the release artifacts for 0.18.0 kueue-prioriy-booster/Chart.lock is left
**What happened**:
Release automation: after building the release artifacts for 0.18.0 kueue-prioriy-booster/Chart.lock is left

the file is deleted as for other projecs

It was hit during the release process for 0.18.0, I don't remember which specific command produced the file. 

https://github.com/kubernetes-sigs/kueue/issues/10861

mbobrovskyi:
#### What type of PR is this?
/kind cleanup

Ignore charts temp files

Fixes #11819

```release-note
NONE
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,4 @@ node_modules
 test/e2e/kueueviz/cypress/screenshots
 test/e2e/kueueviz/cypress/videos
 
-cmd/experimental/kueue-populator/charts/kueue-populator/charts
-cmd/experimental/kueue-populator/charts/kueue-populator/Chart.lock
 
--- a/cmd/experimental/kueue-populator/.gitignore
+++ b/cmd/experimental/kueue-populator/.gitignore
@@ -1,0 +1,2 @@
+charts/kueue-populator/charts
+charts/kueue-populator/Chart.lock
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -41,6 +41,4 @@ node_modules
 test/e2e/kueueviz/cypress/screenshots
 test/e2e/kueueviz/cypress/videos
 
-cmd/experimental/kueue-populator/charts/kueue-populator/charts
-cmd/experimental/kueue-populator/charts/kueue-populator/Chart.lock
```
tenzen-y, line 45: Why didn't this root gitignore work well?

tenzen-y (commented):
/lgtm
/approve

tenzen-y:
/hold 
during release process

.gitignore
mbobrovskyi: It works for `kueue-populator`, but we also need to add it for `kueue-priority-booster`. I can add this to the root .gitignore as well, but I think it's clearer to keep a separate .gitignore file.

tenzen-y: I see. thank you for clarification.
I'm ok with keeping thesse ignore parameters only inside of kueue-populator.

tenzen-y:
@mbobrovskyi can you manually cherry pick without priority-boost controller one into release 0.17 branch?

PR #11832, opened 2026-05-29
KEP-8190: add curve support to kueue

### LINKED ISSUE #10698: Add TLSCurves support to Kueue
https://github.com/kubernetes/kubernetes/pull/137115

Once 1.36 is supported in Kueue, I want to add support for this.

_Originally posted by @kannon92 in https://github.com/kubernetes-sigs/kueue/issues/10696#issuecomment-4296662228_

kannon92:
#### What type of PR is this?
/kind feature

Add curve support for TLSOptions for Kueue.

Fixes #10698 

Implementation closely follows https://github.com/kubernetes/kubernetes/pull/137115.

```release-note
Security: Added `curvePreferences` to `TLSOptions`, allowing administrators to
restrict the TLS key-exchange groups used by Kueue's TLS-enabled servers to an
approved set. Values are specified as numeric IANA TLS Supported Group IDs; when
the option is omitted, Go's default selection is used.
```

## Summary by CodeRabbit

* **New Features**
  * Added `TLSOptions.curvePreferences` to restrict allowed TLS key-exchange curves via numeric IANA TLS Supported Group IDs.
  * Curve preferences are applied to server TLS configuration, including alongside configured cipher suites.
* **Documentation**
  * Updated the v1beta1/v1beta2 API references and the TLS Security Profiles KEP to describe `curvePreferences` and note Go defaults when omitted.
* **Validation & Tests**
  * Extended configuration validation and unit tests for valid/invalid curve preferences.


apis/config/v1beta2/configuration_types.go
```diff
@@ -443,6 +443,13 @@ type TLSOptions struct {
 	// The default would be to not set this value and inherit golang settings.
 	// +optional
 	CipherSuites []string `json:"cipherSuites,omitempty"`
+
+	// curvePreferences is the list of allowed TLS key exchange mechanisms (curves)
+	// for the server, specified as numeric IANA TLS Supported Group IDs.
+	// See https://pkg.go.dev/crypto/tls#CurveID for values supported by the current Go version.
+	// If omitted, Go defaults are used.
+	// +optional
+	CurvePreferences []int32 `json:"curvePreferences,omitempty"`
```
mwielgus, line 467: Is there a way to have string names here instead of ints? Typing/copying values like 4587 is very error-prone.

kannon92: cc @damdo @liggitt 

The code is directly using the API that is in k/k. This was brought up in a sig-apimachinery but the main concern is putting the burden on k/k to maintain this list.

There isn't a nice function or library in Golang AFAIK to get the list.

kannon92:
@coderabbit review

keps/8190-tls-security-profiles/README.md
```diff
@@ -201,6 +219,11 @@ To maintain backward compatibility and provide secure defaults:
   - For TLS 1.3, cipher suites are not configurable (Go uses the built-in secure set)
   - Valid cipher suite values are defined in Go's crypto/tls package constants
 
+- **CurvePreferences**: nil (uses Go's default curve selection)
+  - Values are numeric IANA TLS Supported Group IDs (e.g., 23 for X25519, 24 for P-256, 29 for P-384)
+  - See https://pkg.go.dev/crypto/tls#CurveID for values supported by the current Go version
```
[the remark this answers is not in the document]
kannon92: good call. done.

kannon92:
@sohankunkerkar 
could you take a look?

pkg/util/tlsconfig/tlsconfig.go
```diff
@@ -50,12 +51,23 @@ func ParseTLSOptions(cfg *config.TLSOptions) (*TLS, error) {
 	if len(cfg.CipherSuites) > 0 {
 		cipherSuites, err := convertCipherSuites(cfg.CipherSuites)
 		if err != nil {
-			errRet = errors.Join(err)
+			errRet = errors.Join(errRet, err)
```
sohankunkerkar: I think this looks like a bug fix. Please create a separate PR for this.

keps/8190-tls-security-profiles/README.md
```diff
@@ -110,6 +115,12 @@ As a security engineer, I need to disable weak cipher suites and enforce strong
 services in my cluster. I want to configure Kueue's TLS settings to align with our organization's security baseline,
 which requires specific cipher suites approved by our security team.
 
+#### Story 3
+
+As a platform administrator, I need to restrict TLS key exchange to specific elliptic curves (e.g., X25519, P-256)
+to comply with organizational cryptographic policies. I want to configure Kueue's curve preferences so that only
+approved key exchange mechanisms are used across all its endpoints.
+
 ### Risks and Mitigations
```
sohankunkerkar, line 124: One question: is there any risk of misconfigured curve preferences causing webhook<->apiserver TLS handshake failures? If the admin sets curves that the apiserver doesn't support, the webhook silently breaks. Same gap exists for cipher suites, but worth a note here.

pkg/util/tlsconfig/tlsconfig.go
```diff
@@ -50,12 +51,23 @@ func ParseTLSOptions(cfg *config.TLSOptions) (*TLS, error) {
...
 		}
 		if err == nil && len(cipherSuites) > 0 {
 			ret.CipherSuites = cipherSuites
 		}
 	}
+
+	if len(cfg.CurvePreferences) > 0 {
```
sohankunkerkar: nit: we could extract a shared helper here. cipher suites block at line 51-57

pkg/util/tlsconfig/tlsconfig.go
```diff
@@ -104,3 +117,14 @@ func convertCipherSuites(cipherSuites []string) ([]uint16, error) {
 	}
 	return suites, nil
 }
+
+func convertCurvePreferences(curveIDs []int32) ([]tls.CurveID, error) {
+	if len(curveIDs) == 0 {
```
sohankunkerkar: This is redundant as `cliflag.TLSCurvePreferences` already handles empty input and returns nil.

sohankunkerkar (commented):
JFYI, [this](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/config/config.go#L185) silently drops `ParseTLSOptions` errors for the webhook server while [main.go](https://github.com/kubernetes-sigs/kueue/blob/main/cmd/kueue/main.go#L195) exits on the same error for metrics. which means invalid config would start the webhook with Go defaults silently

keps/8190-tls-security-profiles/README.md
kannon92: If the configuration is not correct we don't start the kueue-controller-manager. So I think this would cause a failure and an admin would need to configure correctly right?

pkg/util/tlsconfig/tlsconfig.go
kannon92: https://github.com/kubernetes-sigs/kueue/pull/12292

kannon92:
> JFYI, [this](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/config/config.go#L185) silently drops `ParseTLSOptions` errors for the webhook server while [main.go](https://github.com/kubernetes-sigs/kueue/blob/main/cmd/kueue/main.go#L195) exits on the same error for metrics. which means invalid config would start the webhook with Go defaults silently

https://github.com/kubernetes-sigs/kueue/pull/12293

keps/8190-tls-security-profiles/README.md
sohankunkerkar: Right, the concern isn't about invalid config. Kueue already catches that at startup. It's about a valid curve config that doesn't overlap with what the apiserver supports. Kueue starts fine, but webhook TLS handshakes fail at runtime. Worth noting in the Risks section so admins know to check compatibility.

pkg/util/tlsconfig/tlsconfig.go
kannon92: updated.

pkg/util/tlsconfig/tlsconfig.go
kannon92: updated.

kannon92:
/hold

Let's merge https://github.com/kubernetes-sigs/kueue/pull/12293 first and then I'll update this PR to reflect latest cleanups.

keps/8190-tls-security-profiles/README.md
kannon92: I changed the code in a separate PR to fail if webhooks have bad values also.

So I think this problem won't occur anymore.

kannon92:
/hold cancel

All prep PRs are merged.

kannon92:
cc @sohankunkerkar @mimowo @tenzen-y

pkg/util/tlsconfig/tlsconfig.go
sohankunkerkar: hmm... this hasn't been updated yet.

pkg/util/tlsconfig/tlsconfig_test.go
```diff
@@ -19,13 +19,30 @@ package tlsconfig
...
+// compareAllErrors checks that every error in wantErrs is present in got.
+func compareAllErrors(wantErrs []error, got error) bool {
+	if len(wantErrs) == 0 && got == nil {
+		return true
+	}
+	if len(wantErrs) == 0 || got == nil {
+		return false
+	}
+	for _, want := range wantErrs {
+		if !strings.Contains(got.Error(), want.Error()) {
```
sohankunkerkar:
```suggestion
		if !errors.Is(got, want) {
```

kannon92: updated.

pkg/util/tlsconfig/tlsconfig.go
kannon92: For real, updated this now.

keps/8190-tls-security-profiles/README.md
sohankunkerkar, line 163: we don't need to do it now but wanted to note here: The actual API fields are `MinVersion` and `CipherSuites`

sohankunkerkar (commented):
Thanks! 
/lgtm

mimowo (commented):
/approve
Looks great, thank you for maintaining the TLS settings 👍 

/hold
in case @tenzen-y has some comments

kannon92:
@tenzen-y 
can you take a quick look? Want to see if I can remove the hold to merge.

pkg/util/tlsconfig/tlsconfig_test.go
```diff
@@ -330,13 +389,71 @@ func TestConvertTLSMinVersion(t *testing.T) {
 	}
 }
 
+func TestConvertCurvePreferences(t *testing.T) {
+	tests := []struct {
```
tenzen-y, line 393:
```suggestion
	tests := map[string]struct {
```

Making this map allows us to prove that each test case doesn't rely on other test cases.

pkg/util/tlsconfig/tlsconfig_test.go
```diff
@@ -330,13 +389,71 @@ func TestConvertTLSMinVersion(t *testing.T) {
...
+		name     string
+		input    []int32
+		expected []tls.CurveID
+		wantErr  []error
```
tenzen-y, line 397:
```suggestion
		wantErr  error
```

It seems that we don't need have a slice error.

pkg/util/tlsconfig/tlsconfig_test.go
```diff
@@ -330,13 +389,71 @@ func TestConvertTLSMinVersion(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := convertCurvePreferences(tt.input)
+			if !compareAllErrors(tt.wantErr, err) {
+				t.Errorf("convertCurvePreferences() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
```
tenzen-y, line 436: We should just perform `if diff := cmp.Diff(tc.wantErr, err, cmpopts.EquateErrors(); len(diff) != 0`

pkg/util/tlsconfig/tlsconfig_test.go
```diff
@@ -330,13 +389,71 @@ func TestConvertTLSMinVersion(t *testing.T) {
...
+
+			if err != nil {
+				return
+			}
+
```
tenzen-y, line 441:
```suggestion
```

We should verify `got` even when error is not nil.

pkg/util/tlsconfig/tlsconfig_test.go
```diff
@@ -330,13 +389,71 @@ func TestConvertTLSMinVersion(t *testing.T) {
...
+			if !cmp.Equal(got, tt.expected) {
+				t.Errorf("convertCurvePreferences() diff: %s", cmp.Diff(tt.expected, got))
```
tenzen-y, line 443:
```suggestion
			if diff := cmp.Equal(got, tt.expected); len(diff) != 0 {
				t.Errorf("Unexpected convertCurvePreferences() (-want, +got):\n%s", diff)
```

pkg/util/tlsconfig/tlsconfig_test.go
```diff
@@ -434,6 +575,10 @@ func TestBuildTLSOptions(t *testing.T) {
 			if !cmp.Equal(cfg.CipherSuites, tt.wantCiphers) {
 				t.Errorf("CipherSuites diff: %s", cmp.Diff(tt.wantCiphers, cfg.CipherSuites))
 			}
+
+			if !cmp.Equal(cfg.CurvePreferences, tt.wantCurves) {
+				t.Errorf("CurvePreferences diff: %s", cmp.Diff(tt.wantCurves, cfg.CurvePreferences))
+			}
```
tenzen-y, line 581:
```suggestion
			if diff := cmp.Equal(cfg.CurvePreferences, tt.wantCurves); len(diff) != 0 {
				t.Errorf("Unexpected CurvePreferences (-want,+got):\n%s", diff)
			}
```

pkg/util/tlsconfig/tlsconfig_test.go
```diff
@@ -26,6 +26,22 @@ import (
...
+		return false
+	}
+	for _, want := range wantErrs {
+		if !errors.Is(got, want) {
+			return false
+		}
+	}
+	return true
+}
+
```
tenzen-y, line 44:
```suggestion
```

I don't see that this is needed.

tenzen-y (commented):
If you will address my comments in a follow-up PR, feel free to unhold this PR.
/lgtm
/approve

kannon92:
> If you will address my comments in a follow-up PR, feel free to unhold this PR. /lgtm /approve

I will work on this today.

/hold cancel

pkg/util/tlsconfig/tlsconfig_test.go
kannon92: https://github.com/kubernetes-sigs/kueue/pull/12861

pkg/util/tlsconfig/tlsconfig_test.go
kannon92: https://github.com/kubernetes-sigs/kueue/pull/12861

pkg/util/tlsconfig/tlsconfig_test.go
kannon92: removed.

mimowo:
/release-note-edit
```release-note
Security: Added `curvePreferences` to `TLSOptions`, allowing administrators to
restrict the TLS key-exchange groups used by Kueue's TLS-enabled servers to an
approved set. Values are specified as numeric IANA TLS Supported Group IDs; when
the option is omitted, Go's default selection is used.
```

PR #11853, opened 2026-05-31
fix(kueueviz): prevent flashing zero counts on dashboard and fix dependencies

### LINKED ISSUE #11852: kueueviz: Dashboard flashes zero counts briefly on page load
**What happened**:
When loading or refreshing the KueueViz dashboard, all resource and workload counters (ClusterQueues, LocalQueues, Workloads) briefly display `0` before flashing to their actual values a split-second later.

This happens because `Dashboard.jsx` unconditionally sets `loading` to `false` during the initial mount, before the WebSocket connection has retrieved any data (`kueueData` is still `null`). Additionally, `namespacesError` is used within this `useEffect` block but is missing from its dependency array.

The dashboard should display a loading spinner while the WebSocket connection is pending, and only show the widgets once the data has finished loading.

1. Run KueueViz locally.
2. Navigate or refresh the Dashboard page.
3. Observe all widgets briefly show `0` before updating to their actual counts.

The fix is to only set `loading` to `false` once `kueueData` is successfully received or if an error is encountered, and to properly add `namespacesError` to the dependency array. Happy to open a PR for this!

YadavAkhileshh:
#### What type of PR is this?
/kind bug
/area dashboard

In `Dashboard.jsx`, the component unconditionally sets `loading` to `false` at the end of the `useEffect` block, even when `kueueData` is `null` on mount. This causes the UI to immediately transition away from the loading spinner and render the dashboard widgets with `0` values. Once the WebSocket connects and streams the data, the widgets flash to their correct counts.

Additionally, `namespacesError` was used inside the `useEffect` hook but was missing from its dependency array.

This PR:
1. Swaps `setLoading(false)` to only execute when `kueueData` is successfully received, or when an error (`kueueError` or `namespacesError`) is captured.
2. Adds `namespacesError` to the `useEffect` dependency array.

This ensures a clean loading transition with absolutely no zero-value flashes on mount.

Fixes #11852

None.

Yes

```release-note
KueueViz: Fixed a bug where the dashboard briefly displayed zero counts for all metrics on page load before the WebSocket connection finished loading.
```

### DIFF
--- a/cmd/kueueviz/frontend/src/Dashboard.jsx
+++ b/cmd/kueueviz/frontend/src/Dashboard.jsx
@@ -66,11 +66,13 @@ const Dashboard = () => {
           toast.error(`Workload ${workload.metadata?.name} was preempted: ${workload.preemption.reason}`);
         }
       });
+      setLoading(false);
     }
-    if (kueueError || namespacesError) setError(kueueError || namespacesError);
-
-    setLoading(false);
-  }, [kueueData, kueueError]);
+    if (kueueError || namespacesError) {
+      setError(kueueError || namespacesError);
+      setLoading(false);
+    }
+  }, [kueueData, kueueError, namespacesError]);
 
   if (error) {
     return <ErrorMessage error={error} />;


mbobrovskyi:
I'm wondering if it's possible to add a test case.

YadavAkhileshh:
@mbobrovskyi thanks for the review! Currently, the frontend uses cypress E2E tests (`app.cy.js`) instead of component unit tests. Asserting the brief loading spinner state is difficult in E2E tests because the WebSocket connects almost instantly in the test environment. However, this change prevents race conditions on mount, making the existing dashboard E2E tests much more stable and reliable!

mimowo:
/test pull-kueue-test-e2e-extended-shard-1-main-1-36
/test pull-kueue-test-e2e-extended-shard-1-main-1-35
/test pull-kueue-test-e2e-extended-shard-1-main-1-34
/test pull-kueue-test-e2e-extended-shard-1-main-1-33
As the failures are likely unrelated

mimowo:
> Currently, the frontend uses cypress E2E tests (app.cy.js) instead of component unit tests. 

Is that accurate? We have examples of unit tests for example here : https://github.com/kubernetes-sigs/kueue/blob/main/cmd/kueueviz/frontend/src/useWebSocket.test.js

Do you mean they are hard to use for the testing the bug, or maybe not executed on CI?

YadavAkhileshh:
@mimowo sorry for the confusion! You're right, unit tests do exist and run on ci.

What I meant is that testing this specific loading-state transition is tricky we'd need to mock `useWebSocket` in a `Dashboard.test.js`, assert the spinner renders when data is `null`, then simulate data arriving and verify it disappears. 
Happy to add that if it would help the pr move forward!

mimowo:
> What I meant is that testing this specific loading-state transition is tricky we'd need to mock useWebSocket in a Dashboard.test.js, assert the spinner renders when data is null, then simulate data arriving and verify it disappears.

I'm wondering maybe the reasonable test would be to ensure the loading indicator is displayed while data is null. IIUC part of the issue is that currently it will not, because we call setLoading(false) unconditionally.

At the same time the new code is very clear, so this is not a strict requirement, feel free to unblock the PR if you think the gain / effort is particularly good.

mimowo:
Thank you for the bugfix 👍 
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17

in case you would like to add the minimal unit test for ensuring `setLoading(false)` is not called while data is null.
/hold

mimowo:
/unhold
let's consider it for a follow up cleanup

Issue #4803, opened 2025-03-26
Add `waitForPodsReady` to `Workload` API


KPostOffice:
**What would you like to be added**:
`waitForPodsReady` per workload

**Why is this needed**:
I think it would be useful to be able to specify `waitForPodsReady` on a per Workload basis. Certain workloads may take longer to get to a ready state than others, so configuring this on at a cluster-wide scope feels awkward. These values can be propagated to the workload via annotations or labels maybe.

**Completion requirements**:

This enhancement requires the following artifacts:

- [x] Design doc
- [x] API change
- [x] Docs update

The artifacts should be linked in subsequent comments.

kannon92:
https://github.com/kubernetes-sigs/kueue/tree/main/keps/349-all-or-nothing#more-granular-configuration-to-enable-the-mechanism

This was discussed briefly when this API was introduced.

kannon92:
One issue is that workloads are not really exposed by Kueue so I think it would be labels/annotations on the Job objects that would get passed down to workload object once workload is created.

kannon92:
@mimowo @tenzen-y @PBundyra @amy 

WDYT of this? This came up as a feature request recently for us.

amy:
Not opposed. Probably also means you need a default number configured for the cluster though and not rely on users to remember to configure the number.

mimowo:
I'm supportive for that 👍 

However, we need to iron out the details in the KEP. Some questions which come up to my mind:
1. should we enable that even if the global waitForPodsReady is disabled?
2. the API - probably a JSON-based annotation

kannon92:
cc @MaysaMacedo 
I followed up offline and she is interested in looking into this.

PBundyra:
> I'm supportive for that 👍
> 
> However, we need to iron out the details in the KEP. Some questions which come up to my mind:
> 
> 1. should we enable that even if the global waitForPodsReady is disabled?
> 2. the API - probably a JSON-based annotation

+1 to all the above

edmund-shi-arm:
really looking forward to this feature!!! what is the progress?

kimminw00:
[...]

PR #11855, opened 2026-05-31
config: enable waitForPodsReady by default

### LINKED ISSUE #11366: Enable waitForPodsReady by default as a safety net

Enable wait for Pods ready by default as a safety net.

The main consideration I think is backwards compatibility, but maybe with a suffiently large timeout this would be acceptable. This is still a beta feature, I think this could go by default:

```yaml
waitForPodsReady:
  timeout: 1h
  recoveryTimeout: 1h
  blockAdmission: false
  requeuingStrategy:
    backoffBaseSeconds: 60
    backoffMaxSeconds: 3600
    backoffLimitCount: null // endless
    timestamp: Eviction
```
This is our recommendation basically with an extended timeout.

I wanted to post it long time ago, but the KEP DRA was the main trigger: https://github.com/kubernetes-sigs/kueue/pull/10283/changes#diff-9992bde183489d9f804057ec1cebd8b81b2470c8a9430abffee5ad8db5278e93R321

**Why is this needed**:

We have multiple incidents where waitForPodsReady could have mitigated the issue, and so we now always recommend it to all users, but sometimes only after an incident.

amirialy:
#### What type of PR is this?

/kind feature

This PR enables WaitForPodsReady by default by applying default configuration values.

The default configuration now sets:

- timeout: 30m
- recoveryTimeout: 30m
- blockAdmission: false
- default requeuing strategy values

Fixes #11366

This change updates the defaulting behavior for WaitForPodsReady and adjusts the corresponding defaulting and configuration tests.

```release-note
ACTION REQUIRED: WaitForPodsReady is now enabled by default. New Kueue installations and existing installations that do not explicitly configure `waitForPodsReady` will use the default WaitForPodsReady configuration (30 minute timeout, 30 minute recovery timeout).
```

## Release Notes

* **New Features**
  * Added consistent defaulting for pod readiness waiting behavior, including a 30-minute default timeout and related readiness/recovery settings when omitted.
* **Bug Fixes**
  * Ensures the readiness default settings apply reliably across configurations, even when the readiness block is not pre-populated.
* **Tests**
  * Updated and expanded configuration, encoding/serialization, and defaulting tests to cover the new pod readiness default values.


kannon92:
Please follow our template on creating PRs. The release note is critical for us.

kannon92 (commented):
/ok-to-test

kannon92 (commented):
We should probably update the site for WaitForPodsReady to mention this is default on for now.

https://kueue.sigs.k8s.io/docs/tasks/manage/setup_wait_for_pods_ready/

Maybe we mention that the default settings and modify this page to mean configuration waitForPodsReady

apis/config/v1beta2/defaults.go
```diff
@@ -94,6 +95,17 @@ func SetDefaults_Configuration(cfg *Configuration) {
 	cfg.ClientConnection.QPS = cmp.Or(cfg.ClientConnection.QPS, new(DefaultClientConnectionQPS))
 	cfg.ClientConnection.Burst = cmp.Or(cfg.ClientConnection.Burst, new(DefaultClientConnectionBurst))
 
+	cfg.WaitForPodsReady = cmp.Or(
+		cfg.WaitForPodsReady,
+		&WaitForPodsReady{},
+	)
+
+	if cfg.WaitForPodsReady.Timeout.Duration == 0 {
```
mwielgus, line 103: How do I turn off WaitForPodsReady after this change?

apis/config/v1beta2/defaults.go
```diff
@@ -45,6 +45,7 @@ const (
 	DefaultClientConnectionBurst          int32   = 30
 	defaultJobFrameworkName                       = "batch/job"
 	DefaultMultiKueueGCInterval                   = time.Minute
+	DefaultWaitForPodsReadyTimeout                = time.Hour
```
mwielgus: 1 hour seems to be a bit large. Why this value? What about something smaller like 20-30 min?

mwielgus (commented):
Probably it would be good to put the default explicitly in the default configuration file, for visibility.

apis/config/v1beta2/defaults.go
amirialy: Before this change, waitForPodsReady being nil meant the feature was disabled, while a non-nil WaitForPodsReady configuration enabled it.

With this change, the defaulting logic initializes WaitForPodsReady with default values when the field is omitted, and there is no explicit opt-out for it.

I thought of something like an enable flag but that would change API type definition.

apis/config/v1beta2/defaults.go
amirialy: That is based off the issue discussion in #11366

mimowo: 1h was proposed by me initially for extra safety so that it does not break any real installations. However, let me check with some users what they use in practice. If none of them would be impacted by 30min then I think it makes more sense than 1h.

mimowo: I have synced with some users and 30min sounds like a reasonable value here.

site/content/en/docs/tasks/manage/setup_wait_for_pods_ready.md
```diff
@@ -13,113 +13,119 @@ if the physical availability of resources do not match the configured quotas in
...
-        backoffMaxSeconds: 3600
+waitForPodsReady:
+  timeout: 1h
+  recoveryTimeout: 1h
+  blockAdmission: false
+  requeuingStrategy:
+    timestamp: Eviction | Creation
+    backoffLimitCount: 5
+    backoffBaseSeconds: 60
+    backoffMaxSeconds: 3600
```
mimowo: I think no need to change the snippet. We may stay with the old one

site/content/en/docs/tasks/manage/setup_wait_for_pods_ready.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: "Setup All-or-nothing with ready Pods"
+title: "Configure All-or-nothing Scheduling with ready Pods"
```
mimowo: Let's try to minimize the docs changes. Feel free to submit a preparetory or follow up documentation update.

amirialy: OK. I reverted back to the old docs for now, I will submit a follow-up doc in another PR.

kannon92 (commented):
/lgtm

apis/config/v1beta2/defaults.go
```diff
@@ -94,6 +95,17 @@ func SetDefaults_Configuration(cfg *Configuration) {
 	cfg.ClientConnection.QPS = cmp.Or(cfg.ClientConnection.QPS, new(DefaultClientConnectionQPS))
 	cfg.ClientConnection.Burst = cmp.Or(cfg.ClientConnection.Burst, new(DefaultClientConnectionBurst))
 
+	cfg.WaitForPodsReady = cmp.Or(
```
MaysaMacedo, line 98: I think this always enable it, right?
Should we consider some way of disabling this feature if needed?
cc @kannon92

kannon92: Yea I think you call out a bug.

I'm fine to default on but with this change we are saying that if someone doesn't specify `WaitForPodsReady` then we will use default. Maybe we should consider default this on in the helm chart and kustomize so if folks want to disable this they can not specify this config. That does mean that on upgrades the old disabled config will still persist though.

How can someone disable the default setting?

cc @tenzen-y @mimowo

apis/config/v1beta2/defaults.go
kannon92: https://github.com/kubernetes-sigs/kueue/pull/11855#discussion_r3475811314

@MaysaMacedo also calls this out.

We do have users that disable this field AFAIK as they handle gang scheduling in other ways (scheduler-plugins).

apis/config/v1beta2/defaults.go
mimowo: Do we see anyone would need to disable the feature really? I think moving the timeout to 24h is probably effectively disabling. We could also have a feature gate for the migration period, but it seems to me we could enable the feature for all installations without the way to disable. Do you see use cases for disablling ?

kannon92: @dgrove-oss's team is the one I see where they disable this feature.

https://github.com/project-codeflare/mlbatch/blob/main/setup.k8s/kueue/controller_manager_config.yaml#L26

From my understanding it was because they use coscheduling or sakara for gang scheduling.

mimowo: Interesting, I would like to learn more why disabling, and if maybe 1h would be enough to enable for this use case.

Making waitForPodsREady would also clarify the semantics for https://github.com/kubernetes-sigs/kueue/issues/4803. Otherwise we may have weird cases where the feature is globally disbaled, but enabled per workload which sounds confusing to me.

dgrove-oss: Sorry to be slow responding.

We made the decision to disable `waitForPodsReady` quite a while ago (circa kueue 0.7).  There may have been some interactions with gang-scheduling via coscheduler, but the bigger issue was it interacted badly with the fault tolerance capabilities we had in AppWrapper (https://project-codeflare.github.io/appwrapper/arch-fault-tolerance/#detailed-description).   These provide the ability to have job-specific timeouts and also break the "not ready" time into two phases:  `AdmissionGracePeriod` and `WarmupGracePeriod`.

mimowo: @dgrove-oss what do you usually set for `WarmupGracePeriod`? IIUC this is the equivalent of the PodsReady timeout? 

>  but the bigger issue was it interacted badly with the fault tolerance capabilities we had in AppWrapper (https://project-codeflare.github.io/appwrapper/arch-fault-tolerance/#detailed-description)

Can you elaborate on that? If the PodsReadyTimeout is exceeded then Kueue evicts the entire workload (recycles). Is it that AppWrapper does not support well the eviction cycle? 

I think since this was 0.7 when the decision was made maybe we should re-evaluate the status quo. Do you have some ideas for checks we should perform?

In any case we may document that admins who install AppWrapper may need to "disable" waitForPods ready by setting an arbitrarily high timeout like 1 year. However, I think most of the deployments could still benefit from the safety net.

dgrove-oss: > In any case we may document that admins who install AppWrapper may need to "disable" waitForPods ready by setting an arbitrarily high timeout like 1 year. However, I think most of the deployments could still benefit from the safety net.

I think this is reasonable.  AppWrapper provides a configuration that admins can use to enforce upper limits on its grace periods.  Adding instructions to the AppWrapper documentation that tells admins that they should also set the Kueue default PodsReadyTimeout to be marginally larger than either of those values should be enough.

mimowo (commented):
/lgtm
/approve
Thank you for making that changes, I think waitForPodsReady is a very useful feature. I surveyed some users and they either don't enable the feature, or set it below 30min.  Those users who do not specify waitForPodsReady seem mostly due to not being aware of the option. 

I believe we can mitigate the risks by release note.  However, keeping Kueue without the protection is also a risk which can result in Oncall tickets.

/hold
in case @tenzen-y or @kannon92 have some comments or objections.

kannon92:
I don't have any objections to this. 

One thing we are curious on is how many users actually configure WaitForPodsReady other than timeout.

```golang
		RecoveryTimeout: &metav1.Duration{
			Duration: 30 * time.Minute,
		},
		RequeuingStrategy: &RequeuingStrategy{
			Timestamp:          new(EvictionTimestamp),
			BackoffBaseSeconds: ptr.To[int32](DefaultRequeuingBackoffBaseSeconds),
			BackoffMaxSeconds:  ptr.To[int32](DefaultRequeuingBackoffMaxSeconds),
		},
```

The RecoveryTimeout and RequeuingStrategy seem complicated to explain and I'm not sure what guidance we provide on how to configure this.

mimowo:
> The RecoveryTimeout and RequeuingStrategy seem complicated to explain and I'm not sure what guidance we provide on how to configure this.

I agree, they are tricky. Generally for `RequeuingStrategy` we don't have any recommendations other than defaults. 

For static clusters we generally recomment:
```yaml
timeout: 10min
recoveryTimeout: 3min
requeuingStrategy: (default)
```
For autoscaled clusters timeout needs to be larger, but 30min should generally be enough. Also, this time does not include ProvisioningRequests. So if the scaling goes via ProvisioningRequests then 10min is also enough.

kannon92:
> > The RecoveryTimeout and RequeuingStrategy seem complicated to explain and I'm not sure what guidance we provide on how to configure this.
> 
> I agree, they are tricky. Generally for `RequeuingStrategy` we don't have any recommendations other than defaults.
> 
> For static clusters we generally recomment:
> 
> ```yaml
> timeout: 10min
> recoveryTimeout: 3min
> requeuingStrategy: (default)
> ```
> 
> For autoscaled clusters timeout needs to be larger, but 30min should generally be enough. Also, this time does not include ProvisioningRequests. So if the scaling goes via ProvisioningRequests then 10min is also enough.

Is it a problem that our defaults don't match what you recommend for `recoveryTimeout`?

mimowo:
> Is it a problem that our defaults don't match what you recommend for recoveryTimeout?

I don't have actually a view here.  Setting `recoveryTimeout` is more of an optimization than a safety net. 
I would be ok to mark it as 1/3 of the full timeout, so around 10min.

mimowo:
Or for a bit of more conservative approach: 30min.

Our recommendation: timeout: 10min / recoveryTimeout: 3min, is rather for the known environments. I'm cannot tell this will be good for all OSS deployements. So, we may be a bit more conservative in OSS: 30min / 15min wdyt?

kannon92:
For us we are just taking timeout = recoveryTimeout.

This is what the defaults are now that this merged.

@MaysaMacedo and I are thinking through https://github.com/kubernetes-sigs/kueue/pull/12776 so we are curious if we should have two different API fields for timeout and RecoveryTimeout. Or we ignore RecoveryTimeout.

kannon92:
If we don't really have a stance I say let's leave the conservative approach.

tenzen-y (commented):
Thank you for moving this forward. I believe that this Kueue featue is must-to-have for the almost all clusters. And I don't have any objections for the default value.

But, it looks like slightly user-facing changes. So, I would recommend adding ACTION REQUIRED section to release note block.

@Amir380-A Also, can you update the API documentation in a follow-up PR?
https://github.com/kubernetes-sigs/kueue/blob/218f723c6e9979a8449472de767df7c5e2be7e78/apis/config/v1beta1/configuration_types.go#L275-L307

tenzen-y:
> If we don't really have a stance I say let's leave the conservative approach.

@kannon92 

In my cluster, we explicitly specify 30min (timeout) / 30 min (recovery timeout) since the Workload typically consumes start-up time due to large image pulling and sidecar health checks, .... Even after the Pods started, the researchers sometimes are trying to restart main training / inference server process w/o Pod failure.

I understand that the public cloud cluster 's infrastructure layer is stable like fast and lazy image loading, fast volume provisioning, and more. And they continuously improve such stability. However, in bare-metal on-prem environemt is not the same which is more fragile and less-performance rather than public cloud.

So, I would argue 30 min / 30 min as a default.

Issue #11862, opened 2026-06-01
MultiKueue granular automated quota aggregation


Singularity23x0:
**What would you like to be added**:
Consider ways to group workers' resource flavors together when aggregated on the manager cluster to allow for more efficient quota management at the Manager Cluster level.

**Why is this needed**:
A follow-up to https://github.com/kubernetes-sigs/kueue/issues/9988.
We want to consider possible ways to make MultiKueue quota assignment more efficient when automation is enabled, instead of defaulting to a single, global resource flavor.

**Completion requirements**:

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

Issue #11823, opened 2026-05-29
EffectiveResourceGroups


Singularity23x0:
**What would you like to be added**:
A new ClusterQueue status field: EffectiveResourceGroups.
The field will be added alongside spec.ResourceGroups with the split of responsibilities:
* spec.ResourceGroups - manually set by the user to define ClusterQueue quota in default configuration,
* status.EffectiveResourceGroups - a field representing the de-facto quota as interpreted by the Kueue system, to be used in automated quota management configurations.

**Why is this needed**:
To accommodate MultiKueue automated quota management configuration https://github.com/kubernetes-sigs/kueue/issues/9988.

**Completion requirements**:
status.EffectiveResourceGroups implemented and used by Kueue as the source of truth on effective quota **at least for** automated quota management configurations.

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

PR #11807, opened 2026-05-29
9988 effective resource groups proposal 1

Singularity23x0:
#### What type of PR is this?
/kind feature
/area multikueue

#### What this PR does / why we need it:

EXAMPLE IMPLEMENTATION:
Implements Alpha2 stage of https://github.com/kubernetes-sigs/kueue/issues/9988, based on the design in https://github.com/kubernetes-sigs/kueue/pull/11865.

#### Which issue(s) this PR fixes:

Progresses #9988
Progresses #11823

#### Special notes for your reviewer:
* syncAggregateFlavors currently implements support for a single resource flavor. The design was made with possible future expansion of a more granular resource aggregation (based on flavor). **Discussion on the matter will be covered separately as a follow-up in https://github.com/kubernetes-sigs/kueue/issues/11862.**

#### Does this PR introduce a user-facing change?

```release-note
MultiKueue Automation changes:
* when setting up automation, spec.ResourceGroups is now expected to be empty,
* automatically aggregated quota is now collected under "multikueue-aggregated-flavor"; the flavor is automatically generated and maintained; its spec can be specified via MultiKueue Config.

MultiKueue Config changes: 
* renamed QuotaManagement to QuotaManagementMode, 
* added AutomatedQuotaManagementConfig to specify the default aggregated flavor spec.

ClusterQueue changes: 
* added EffectiveResourceGroups status showing effective quota as interpreted by Kueue,
* ClusterQueue AdmissionCheck flavors (onFlavors) can now specify flavors not present in the ResourceGroup when EffectiveResourceQuota is enabled. This should not affect the logic since flavors not set in the ResourceGroups will not be assigned to workloads and thus the ACs will not utilize them.
```

PR #10745, opened 2026-04-23
KEP-10270: Automatic ClusterQueue quota limits based on node selectors

gyliu513:
#### What type of PR is this?

/kind kep
/kind feature

#### Which issue(s) this PR fixes:

Related #10270 

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

## Summary by CodeRabbit

* **Documentation**
  * Added a KEP README detailing automatic quota limits based on node selectors, including alpha/beta behavior, conflict detection, status/conditions, and apply vs. recommend modes.
  * Added KEP release-planning metadata for the feature, including stage and milestone information tied to the relevant feature gate.

PR #11865, opened 2026-06-01
KEP-11823: EffectiveResourceGroups

Singularity23x0:
#### What type of PR is this?
/kind kep

See https://github.com/kubernetes-sigs/kueue/issues/11823 .

Progresses #https://github.com/kubernetes-sigs/kueue/issues/11823

ON HOLD until #12382 is resolved.

* Per-flavor auota aggregation will be discussed in https://github.com/kubernetes-sigs/kueue/issues/11862.
* For MultiKueue automation details see: keps/9988-multikueue-manager-quota-automation
* **Implementation draft**: https://github.com/kubernetes-sigs/kueue/pull/11807

## Summary by CodeRabbit

* **Documentation**
  * Added a KEP describing a new ClusterQueue status field for effective resource groups, controller ownership, rollout stages, test plan, graduation criteria, drawbacks, and alternatives.
  * Updated MultiKueue quota automation docs with automated quota-management modes (manual vs dynamic aggregation), new validation requirements, opt-out examples, and a dynamic resource-flavor alternative.

### DIFF
--- a/keps/11823-effective-resource-groups/README.md
+++ b/keps/11823-effective-resource-groups/README.md
@@ -55,6 +55,27 @@ To resolve this issue and allow the aforementioned initiatives to be implemented
 We introduce the new `status.EffectiveResourceGroups` field alongside the `spec.ResourceGroups` field.
 The `status.EffectiveResourceGroups` will add a layer between the `spec.ResourceGroups` (set directly by users) and Kueue logic. The field will always be populated. By default it will reflect the value set in spec.
 
+```go
+type ClusterQueueStatus struct {
+	
+	// [...]
+
+	// effectiveResourceGroups describes the groups of resources as seen by Kueue controllers.
+	// Each resource group defines the list of resources and a list of flavors
+	// that provide quotas for these resources.
+	// Each resource and each flavor can only form part of one resource group.
+	// By default, it's equal to spec.resourceGroups. However, in some scenarios (e.g. MultiKueue)
+	// it may differ from spec.resourceGroups.
+	// +listType=atomic
+	// +kubebuilder:validation:MaxItems=16
+	// +optional
+	EffectiveResourceGroups []ResourceGroup `json:"effectiveResourceGroups,omitempty"`
+
+	// [...]
+
+}
+```
+
 All internal Kueue logic will use either the `spec.ResourceGroups` or the `status.EffectiveResourceGroups` as a source of truth for the ClusterQueue's quota, based on whether the `EffectiveResourceQuotas` featureGate is enabled.
 
 Setting the value of the effective quota will be handled by the Core-ClusterQueue-Controller and MultiKueue-ClusterQueue-Controller depending on the configuration:
--- a/keps/11823-effective-resource-groups/README.md
+++ b/keps/11823-effective-resource-groups/README.md
@@ -89,14 +89,6 @@ Mitigating this requires an expansive sweep of existing logic, ensuring the code
 
 The `status.EffectiveResourceGroups` will be used to provide resource quota only when the **EffectiveResourceQuotas** feature is enabled. Otherwise we will default to using the spec. To make this feasible, all effective calls to retrieve the ResourceGroups will be handled by a dedicated method:
 
-```go
-func GetEffectiveResourceGroups(cq *kueue.ClusterQueue) []kueue.ResourceGroup {
-	if features.Enabled(features.EffectiveResourceQuotas) {
-		return cq.Status.EffectiveResourceGroups
-	}
-	return cq.spec.ResourceGroups
-}
-```
 
 MultiKueue quota automation will, since its [Alpha2](https://github.com/kubernetes-sigs/kueue/tree/main/keps/9988-multikueue-manager-quota-automation#graduation-criteria) stage, require the **EffectiveResourceQuotas** feature to be enabled alongside its dedicated feature gate.
 
--- a/keps/11823-effective-resource-groups/README.md
+++ b/keps/11823-effective-resource-groups/README.md
@@ -55,7 +55,7 @@ To resolve this issue and allow the aforementioned initiatives to be implemented
 We introduce the new `status.EffectiveResourceGroups` field alongside the `spec.ResourceGroups` field.
 The `status.EffectiveResourceGroups` will add a layer between the `spec.ResourceGroups` (set directly by users) and Kueue logic. The field will always be populated. By default it will reflect the value set in spec.
 
-All internal Kueue logic will use either the `spec.ResourceGroups` or the `status.EffectiveResourceGroups` as a source of truth for the ClusterQueue's quota, based on whether the feature is enabled.
+All internal Kueue logic will use either the `spec.ResourceGroups` or the `status.EffectiveResourceGroups` as a source of truth for the ClusterQueue's quota, based on whether the `EffectiveResourceQuotas` featureGate is enabled.
 
 Setting the value of the effective quota will be handled by the Core-ClusterQueue-Controller and MultiKueue-ClusterQueue-Controller depending on the configuration:
 
--- a/keps/11823-effective-resource-groups/README.md
+++ b/keps/11823-effective-resource-groups/README.md
@@ -57,7 +57,7 @@ The `status.EffectiveResourceGroups` will add a layer between the `spec.Resource
 
 ```go
 type ClusterQueueStatus struct {
-	
+
 	// [...]
 
 	// effectiveResourceGroups describes the groups of resources as seen by Kueue controllers.
@@ -92,9 +92,7 @@ The MultiKueue ClusterQueue Controller will be created only when the following f
 
 We consider MultiKueue Automated Quota Management **enabled** and **configured correctly** when:
 1. The Quota Management Strategy value is **"Automated"**.
-1. **spec.ResourceGroups is empty.**
-
-We treat spec.ResourceGroup with priority since the parameter is an established, user defined value. If set, it takes precedence over the MultiKueue Manager Queue quota automation. As such, to allow for MultiKueue automated quota management we expect it to be empty.
+2. **spec.ResourceGroups has a valid format.** For the details what is considered a valid format see [KEP 8826](../keps/9988-multikueue-manager-quota-automation#planned-changes-in-alpha2-version).
 
 ### Risks and Mitigations
 
@@ -131,11 +129,10 @@ When the controller detects a reason for special handling of quota sync, it will
 
 When feature.MultiKueue and feature.MultiKueueManagerQuotaAutomation are enabled, the MultiKueue Controller will be created. We will delegate the handling of MultiKueue ClusterQueues when it is set up. The Controller itself will:
 1. Check if the ClusterQueue has the MultiKueue admission check assigned. If not: it will skip the CQ as it is not considered an MK CQ.
-1. Check if:
-    1. the ClusterQueue's MultiKueue Config has the **Quota Management Strategy** set to "Automated"; if not - set the `MultiKueueManagerQuotaAutomation` condition to **false**, reason: "Not Requested",
-    1. `spec.ResourceGroups` is empt; if not - set the `MultiKueueManagerQuotaAutomation` condition to **false**,dreason: "Unsupported Configuration".
-1. If any of the above conditions are false: it will perform the **Default Behavior**.
-1. If both conditions are true: the controller will calculate the aggregated quota across the Manager Queue's Workers and set it as the value of `status.EffectiveResourceGroups`.
+2. Check if:
+    1. the ClusterQueue's MultiKueue Config has the **Quota Management Strategy** set to "Automated"; if not - set the `MultiKueueManagerQuotaAutomation` condition to **false**, reason: "Not Requested" and perform **Default Behavior**,
+    2. `spec.ResourceGroups` has a valid format; if not - set the `MultiKueueManagerQuotaAutomation` condition to **false**, reason: "Unsupported Configuration" and return error,
+4. If both conditions are true: the controller will calculate the aggregated quota across the Manager Queue's Workers and set it as the value of `status.EffectiveResourceGroups`.
 
 ### Test Plan
 
--- a/keps/11823-effective-resource-groups/README.md
+++ b/keps/11823-effective-resource-groups/README.md
@@ -104,16 +104,6 @@ MultiKueue quota automation will, since its [Alpha2](https://github.com/kubernet
 
 By default `status.EffectiveResourceGroups` will be synced to the current value `spec.ResourceGroups`.
 
-```go
-// pkg/util/queue/cluster_queue.go
-
-func SyncEffectiveResourceGroupsToSpec(cq *kueue.ClusterQueue) (needsUpdate bool) {
-	if needsUpdate = !equality.Semantic.DeepEqual(cq.Status.EffectiveResourceGroups, cq.Spec.ResourceGroups); needsUpdate {
-		cq.Status.EffectiveResourceGroups = cq.Spec.ResourceGroups
-	}
-	return
-}
-```
 
 ### Core Controller
 
--- a/keps/11823-effective-resource-groups/README.md
+++ b/keps/11823-effective-resource-groups/README.md
@@ -1,0 +1,203 @@
+# KEP-11823: EffectiveResourceGroups
+
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+- [Proposal](#proposal)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Default behavior](#default-behavior)
+  - [Core Controller](#core-controller)
+  - [MultiKueue quota automation](#multikueue-quota-automation)
+  - [Test Plan](#test-plan)
+    - [Prerequisite testing updates](#prerequisite-testing-updates)
+    - [Unit tests](#unit-tests)
+    - [Integration tests](#integration-tests)
+    - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Treat MultiKueue automation as priority](#treat-multikueue-automation-as-priority)
+  - [Let spec override effective quota](#let-spec-override-effective-quota)
+<!-- /toc -->
+
+## Summary
+
+This KEP extends the current resource group definition mechanism of Kueue to accommodate automated quota management.
+
+The proposal is to introduce a new ClusterQueue Status field: `EffectiveResourceGroups`, used to represent and track the ClusterQueue quota as interpreted by the Kueue scheduler.
+
+## Motivation
+
+We need a way to track the ClusterQueue quota. In the default case the quota is static and described by the user in `spec.ResourceGroups`. There are however two ongoing initiatives aiming to automate ClusterQueue quota management:
+1. [KEP 8826](https://github.com/kubernetes-sigs/kueue/pull/8864)
+1. [KEP 9988](https://github.com/kubernetes-sigs/kueue/tree/main/keps/9988-multikueue-manager-quota-automation#move-the-aggregated-quota-out-of-clusterqueuespec)
+
+When automating effective quota calculation, using `spec.ResourceGroups` to store the value would mean allowing Kueue to modify spec of objects it's supposed to reconcile (specifically, ClusterQueues). This may be perceived as a design smell because:
+* It blurs the line between configuration (managed by users) and state (managed by controllers).
+* It brings a risk of an infinite reconciling loop.
+
+To resolve this issue and allow the aforementioned initiatives to be implemented in a robust and clean way, we propose introducing a new ClusterQueue Status field: `EffectiveResourceGroups` to store the automated quota value.
+
+### Goals
+
+1. Implement the `EffectiveResourceGroups` status field in ClusterQueue API.
+2. Define a clear contract on how `EffectiveResourceGroups` is managed and utilized by Kueue in cases of:
+    * Default setup: MultiKueue disabled; this case will be the basis for any possible future additions of more automated ClusterQueue quota management schemes,
+    * MultiKueue setup without Quota automation: MultiKueue enabled, MultiKueueManagerQuotaAutomation feature disabled.
+    * MultiKueue with MultiKueueManagerQuotaAutomation enabled.
+
+## Proposal
+
+We introduce the new `status.EffectiveResourceGroups` field alongside the `spec.ResourceGroups` field.
+The `status.EffectiveResourceGroups` will add a layer between the `spec.ResourceGroups` (set directly by users) and Kueue logic. The field will always be populated. By default it will reflect the value set in spec.
+
+All internal Kueue logic will use either the `spec.ResourceGroups` or the `status.EffectiveResourceGroups` as a source of truth for the ClusterQueue's quota, based on whether the feature is enabled.
+
+Setting the value of the effective quota will be handled by the Core-ClusterQueue-Controller and MultiKueue-ClusterQueue-Controller depending on the configuration:
+
+1. By default, `EffectiveResourceGroups` will always mirror `spec.ResourceGroups`. `EffectiveResourceGroups` will be synced to `spec.ResourceGroups` during reconciliation in the Core ClusterQueue Controller.
+This covers both the cases when the MultiKueue ClusterQueue Controller is not created, and when it is but the ClusterQueue is not a MultiKueue CQ (does not have the MK AdmissionCheck).
+2. When the MultiKueue ClusterQueue Controller is active, it will handle all ClusterQueues with the Multikueue Admission Check enabled. 
+When the automation is **enabled** and **configured correctly** for a given MK CQ, `EffectiveResourceGroups` will be set directly to a value representing the aggregated quota across the Manager Queue's Workers, calculated as decided upon in [issue 11862](https://github.com/kubernetes-sigs/kueue/issues/11862). Otherwise it will be synced to `spec.ResourceGroups` as well.
+
+The MultiKueue ClusterQueue Controller will be created only when the following feature gates are enabled:
+1. MultiKueue
+1. MultiKueueManagerQuotaAutomation
+1. **EffectiveResourceQuotas**
+
+We consider MultiKueue Automated Quota Management **enabled** and **configured correctly** when:
+1. The Quota Management Strategy value is **"Automated"**.
+1. **spec.ResourceGroups is empty.**
+
+We treat spec.ResourceGroup with priority since the parameter is an established, user defined value. If set, it takes precedence over the MultiKueue Manager Queue quota automation. As such, to allow for MultiKueue automated quota management we expect it to be empty.
+
+### Risks and Mitigations
+
+The major risks here are allowing `spec.ResourceGroups` to be empty and allowing a period of time where `status.EffectiveResourceGroups` might be empty as well.
+
+Mitigating this requires an expansive sweep of existing logic, ensuring the code is made secure against:
+* referencing an empty `spec.ResourceGroups` or `status.EffectiveResourceGroups` properties,
+* a misalignment/delay of `status.EffectiveResourceGroups` being synced to `spec.ResourceGroups`.
+
+## Design Details
+
+`status.EffectiveResourceGroups` will be managed by either the Core ClusterQueue Controller or the MultiKueue ClusterQueue Controller (mutually exclusive) as described in the [Proposal](#proposal) section.
+
+The `status.EffectiveResourceGroups` will be used to provide resource quota only when the **EffectiveResourceQuotas** feature is enabled. Otherwise we will default to using the spec. To make this feasible, all effective calls to retrieve the ResourceGroups will be handled by a dedicated method:
+
+```go
+func GetEffectiveResourceGroups(cq *kueue.ClusterQueue) []kueue.ResourceGroup {
+	if features.Enabled(features.EffectiveResourceQuotas) {
+		return cq.Status.EffectiveResourceGroups
+	}
+	return cq.spec.ResourceGroups
+}
+```
+
+MultiKueue quota automation will, since its [Alpha2](https://github.com/kubernetes-sigs/kueue/tree/main/keps/9988-multikueue-manager-quota-automation#graduation-criteria) stage, require the **EffectiveResourceQuotas** feature to be enabled alongside its dedicated feature gate.
+
+### Default behavior
+
+By default `status.EffectiveResourceGroups` will be synced to the current value `spec.ResourceGroups`.
+
+```go
+// pkg/util/queue/cluster_queue.go
+
+func SyncEffectiveResourceGroupsToSpec(cq *kueue.ClusterQueue) (needsUpdate bool) {
+	if needsUpdate = !equality.Semantic.DeepEqual(cq.Status.EffectiveResourceGroups, cq.Spec.ResourceGroups); needsUpdate {
+		cq.Status.EffectiveResourceGroups = cq.Spec.ResourceGroups
+	}
+	return
+}
+```
+
+### Core Controller
+
+The Core Controller is the default controller handling the new parameter.
+The reconciliation logic will check if the ClusterQueue is subject to the default handling scheme and perform the [Default Behavior](#default-behavior) if applicable.
+
+As of right now the only reason for special handling is the MultiKueue Manager ClusterQueue quota automation scheme, but more can be added in the future.
+When the controller detects a reason for special handling of quota sync, it will log it and skip the default behavior.
+
+```go
+  // pkg/controller/core/clusterqueue_controller.go
+
+  if isMK, err := admissioncheck.QuotaManagedByMultiKueue(ctx, r.client, &cqObj); err != nil {
+		return ctrl.Result{}, err
+	} else if isMK {
+		log.V(2).Info("Skipping EffectiveResourceGroups sync: MultiKueue Manager ClusterQueue quota is managed by a dedicated MultiKueue controller.")
+	} else if queue.SyncEffectiveResourceGroupsToSpec(&cqObj) {
+		log.V(2).Info("Syncing EffectiveResourceGroups to spec.")
+		if err := r.client.Status().Update(ctx, &cqObj); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+```
+
+### MultiKueue quota automation
+
+When feature.MultiKueue and feature.MultiKueueManagerQuotaAutomation are enabled, the MultiKueue Controller will be created. We will delegate the handling of MultiKueue ClusterQueues when it is set up. The Controller itself will:
+1. Check if the ClusterQueue has the MultiKueue admission check assigned. If not: it will skip the CQ as it is not considered an MK CQ.
+1. Check if:
+    1. the ClusterQueue's MultiKueue Config has the **Quota Management Strategy** set to "Automated"; if not - set the `MultiKueueManagerQuotaAutomation` condition to **false**, reason: "Not Requested",
+    1. `spec.ResourceGroups` is empt; if not - set the `MultiKueueManagerQuotaAutomation` condition to **false**,dreason: "Unsupported Configuration".
+1. If any of the above conditions are false: it will perform the **Default Behavior**.
+1. If both conditions are true: the controller will calculate the aggregated quota across the Manager Queue's Workers and set it as the value of `status.EffectiveResourceGroups`.
+
+### Test Plan
+
+[ ] I/we understand the owners of the involved components may require updates to
+existing tests to make this code solid enough prior to committing the changes necessary
+to implement this enhancement.
+
+#### Prerequisite testing updates
+
+- **Beta** - Existing tests must be adjusted to use `status.EffectiveResourceGroups` instead of `spec.ResourceGroups` as the de-facto source-of-truth for effective quota.
+
+#### Unit tests
+
+- **Alpha** - Tests added/adjusted in pkg/controller/admissionchecks/multikueue/clusterqueue_test.go.
+
+#### Integration tests
+
+- **Alpha** - Tests added to:
+	- test/integration/multikueue
+	- test/integration/singlecluster/controller/core/clusterqueue_controller_test.go
+
+#### e2e tests
+
+- **Alpha** - Dedicated tests are added in:
+	- test/e2e/multikueue/baseline
+	- test/e2e/singlecluster/baseline
+
+
+### Graduation Criteria
+
+- **Alpha** - `status.EffectiveResourceGroups` is added and tested. *`EffectiveResourceGroups`* feature is set to **disabled** by default.
+- **Beta** - all existing tests are adjusted to work with the feature enabled. The feature is set to **enabled** by default.
+- **Stable** - Kueue and MultiKueue are confirmed to work as expected with the feature enabled and no issues are reported in the field. The feature gate is removed.
+
+## Implementation History
+
+- 01.06.2026 - KEP created
+
+## Drawbacks
+
+Adding an additional layer of resource group tracking requires reworking core kueue logic and updating most existing tests to take the distinction between the spec and status values into account. Mitigating the main risks will be an extensive effort, prone to error by virtue of being such a fundamental change.
+
+## Alternatives
+
+### Treat MultiKueue automation as priority
+
+Instead of using the spec when automation is enabled (and notifying the user that automation is misconfigured) we could instead override the authority of the spec, setting the effective groups to the automated value.
+
+The main drawback here is that the users are likely to expect the spec they set manually to take precedence as it is a user-facing, user-configured and well established field.
+
+### Let spec override effective quota
+
+In code, we could alternatively allow for the method used to retrieve effective quota to override the value of the `status.EffectiveResourceGroups` field when spec is set.
+This could be counter intuitive as the users would see both the `spec.ResourceGroups` and the `status.EffectiveResourceGroups` reflecting different values, and Kueue using the value of spec being the actual effective value for resource groups. To avoid this it is recommended for `status.EffectiveResourceGroups` to reflect the true effective value at all times.
--- a/keps/11823-effective-resource-groups/README.md
+++ b/keps/11823-effective-resource-groups/README.md
@@ -123,20 +123,6 @@ The reconciliation logic will check if the ClusterQueue is subject to the defaul
 As of right now the only reason for special handling is the MultiKueue Manager ClusterQueue quota automation scheme, but more can be added in the future.
 When the controller detects a reason for special handling of quota sync, it will log it and skip the default behavior.
 
-```go
-  // pkg/controller/core/clusterqueue_controller.go
-
-  if isMK, err := admissioncheck.QuotaManagedByMultiKueue(ctx, r.client, &cqObj); err != nil {
-		return ctrl.Result{}, err
-	} else if isMK {
-		log.V(2).Info("Skipping EffectiveResourceGroups sync: MultiKueue Manager ClusterQueue quota is managed by a dedicated MultiKueue controller.")
-	} else if queue.SyncEffectiveResourceGroupsToSpec(&cqObj) {
-		log.V(2).Info("Syncing EffectiveResourceGroups to spec.")
-		if err := r.client.Status().Update(ctx, &cqObj); err != nil {
-			return ctrl.Result{}, err
-		}
-	}
-```
 
 ### MultiKueue quota automation
 
--- a/keps/11823-effective-resource-groups/README.md
+++ b/keps/11823-effective-resource-groups/README.md
@@ -66,8 +66,8 @@ When the automation is **enabled** and **configured correctly** for a given MK C
 
 The MultiKueue ClusterQueue Controller will be created only when the following feature gates are enabled:
 1. MultiKueue
-1. MultiKueueManagerQuotaAutomation
-1. **EffectiveResourceQuotas**
+2. MultiKueueManagerQuotaAutomation
+3. **EffectiveResourceQuotas**
 
 We consider MultiKueue Automated Quota Management **enabled** and **configured correctly** when:
 1. The Quota Management Strategy value is **"Automated"**.
--- a/keps/11823-effective-resource-groups/kep.yaml
+++ b/keps/11823-effective-resource-groups/kep.yaml
@@ -1,0 +1,33 @@
+title: Effective Resource Groups
+kep-number: 11823
+authors:
+  - "@dpastuszka"
+status: implementable
+creation-date: 2026-05-29
+reviewers:
+  - TBD
+approvers:
+  - TBD
+
+see-also:
+  - "keps/9988-multikueue-manager-quota-automation"
+
+# The target maturity stage in the current dev cycle for this KEP.
+stage: alpha
+
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. This can be the current (upcoming) milestone, if it is being actively
+# worked on.
+latest-milestone: "v0.19"
+
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  alpha: "v0.19"
+  beta: "v0.19"
+  stable: "v0.20"
+
+# The following PRR answers are required at alpha release
+# List the feature gate name and the components for which it must be enabled
+feature-gates:
+  - name: EffectiveResourceQuotas
+disable-supported: true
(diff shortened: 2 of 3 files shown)


Singularity23x0:
@tenzen-y, @olekzabl, @mwielgus, @mimowo PTAL

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,366 @@
...
+#### Story 1 (Optional)
+
+#### Story 2 (Optional)
+
+### Notes/Constraints/Caveats (Optional) -->
+
+
+### Risks and Mitigations
+
+The major risk here is allowing spec.ResourceGroups to be empty.
```
olekzabl: Nit: AFAIK this is allowed already.

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,366 @@
...
+Mitigating this requires an expansive sweep of existing logic, ensuring the code is proofed against the cases of:
+* referencing an empty spec.ResourceGroups of spec.EffectiveResourceGroups properties,
```
olekzabl: I couldn't parse this. Did you mean "or" instead of "of"?

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,366 @@
...
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+This KEP extends the current resource group definition mechanism of Kueue to accommodate automated quota management.
+
+The proposal is to introduce a new ClusterQueue Status field: EffectiveResourceGroups, used to represent and track the ClusterQueue quota as interpreted by the Kueue scheduler.
+
+## Motivation
```
olekzabl, line 33: While it reads nice, I'm missing two bits of context here:

1. (Definitely) [KEP-8826](https://github.com/kubernetes-sigs/kueue/pull/8864), as another motivation for having this mechanism in place, and a potential source of another plugin hooking into it.
2. (Preferably) [This section](https://github.com/kubernetes-sigs/kueue/tree/main/keps/9988-multikueue-manager-quota-automation#move-the-aggregated-quota-out-of-clusterqueuespec) of KEP-9988, just for easier "archeology" (what's the context behind this KEP).

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,366 @@
...
+* a misalignment/delay of spec.EffectiveResourceGroups being synced to spec.ResourceGroups.
+
+## Design Details
+
+status.EffectiveResourceGroups will be managed by either the Core ClusterQueue Controller or the MultiKueue ClusterQueue Controller (mutually exclusive):
+1. status.EffectiveResourceGroups will be managed by the MultiKueue Controller when:
+    * features.MultiKueue is enabled AND
+    * features.MultiKueueManagerQuotaAutomation is enabled AND
+    * ClusterQueue has the MultiKueue admission check assigned.
+2. Otherwise, status.EffectiveResourceGroups will be managed by the Core Controller.
```
olekzabl: Hmm, doesn't this contradict the content of "Proposal" section?

There, you gave 5 necessary conditions to let the MK CQ Controller manage effective quotas.
Here, you listed only 3 out of those 5.

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,366 @@
...
+
+The status.EffectiveResourceGroups will be used to provide resource quota only when the UseEffectiveResourceGroupsAsSourceOfTruth feature is enabled. Otherwise we will default to using the spec. To make this feasible, all effective calls to retrieve the ResourceGroups will be handled by a dedicated method:
```
olekzabl: Nit: How about just `EffectiveResourceQuotas` as the FG name?
I don't see how the additional words help. (And I'm feeling they do clutter).
In particular, `Use` feels somewhat out of convention, and irrelevant.

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,374 @@
...
+    * features.MultiKueueManagerQuotaAutomation is enabled
+  * ClusterQueue has the MultiKueue admission check assigned
+2. Otherwise, *status.EffectiveResourceGroups* will be managed by the Core Controller.
+
+The *status.EffectiveResourceGroups* will be used to provide resource quota only when the UseEffectiveResourceGroupsAsSourceOfTruth feature is enabled. Otherwise we will default to using the spec. To make this feasible, all effective calls to retrieve the ResourceGroups will be handled by a dedicated method:
+
+```go
+func GetEffectiveResourceGroup(cq *kueue.ClusterQueue) []kueue.ResourceGroup {
+	if features.Enabled(features.UseEffectiveResourceGroupsAsSourceOfTruth) {
+		return cq.Status.UseEffectiveResourceGroupsAsSourceOfTruth
```
olekzabl: Search-replace leftover?

```suggestion
		return cq.Status.EffectiveResourceGroups
```

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,374 @@
...
+	}
+	return cq.*spec.ResourceGroups*
```
olekzabl: Nit: if you tried to use emphasized font, it seems not to render.

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,374 @@
...
+}
+```
+
+MultiKueue quota automation will require the UseEffectiveResourceGroupsAsSourceOfTruth feature to be enabled alongside its dedicated feature gate.
```
olekzabl:
```suggestion
MultiKueue quota automation will, since its Alpha2 stage, require the UseEffectiveResourceGroupsAsSourceOfTruth feature to be enabled alongside its dedicated feature gate.
```

Plus optionally link to some section of KEP-9988 (maybe "Graduation Criteria"?) to help the reader understand why.

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,374 @@
...
+
+### Default behavior
+
+By default *status.EffectiveResourceGroups* will be synced to the current value *spec.ResourceGroups*.
+
+```go
+// pkg/util/queue/cluster_queue.go
+
+func SyncEffectiveResourceGroupToSpec(cq *kueue.ClusterQueue) (needsUpdate bool) {
```
olekzabl:
```suggestion
func SyncEffectiveResourceGroupsToSpec(cq *kueue.ClusterQueue) (needsUpdate bool) {
```

olekzabl: Why taking `needsUpdate` as param? The body computes it anyway.

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,374 @@
...
+	return
+}
+```
+
+### Core Controller
+
+The Core Controller is the default controller handling the new parameter.
+The reconciliation logic will check if the ClusterQueue is subject to the default handling scheme and perform the **Default Behavior** if applicable.
+
+As of right now the only reason for special handling is the MultiKueue Manager ClusterQueue quota automation scheme, but more can be added in the future without modifying the Core Controller.
```
olekzabl: "without modifying the Core Controller" feels like over-statement; I'd expect a new `else if` branch below would be needed, or so?

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,374 @@
...
+	isMK = false
+	err = nil
+	if features.Enabled(features.MultiKueue) && features.Enabled(features.MultiKueueManagerQuotaAutomation) && features.Enabled(features.EffectiveResourceGroups) {
+		_, isMK, err = GetMultiKueueAdmissionCheck(ctx, c, cq)
+	}
+	return
+}
+```
+
+### MultiKueue quota automation
```
olekzabl, line 128: Warning: during the review of my KEP-9988, I got feedback that I went too deep into technical details about the controllers. Even though I stayed far from long code snippets.

For this KEP, my initial gut feeling is that:
- the snippets above this point are fine; they're short and potentially disambiguating
- the short snippet right below this point feels to repeat what was already said; I'd consider removing
- the long snippet does not belong to a KEP; you can bring same value by linking to a draft impl PR

olekzabl (commented):
A partial review; sending a few comments I managed to gather so far.
Will come back later.

Though overall looking very nice!

keps/11823-effective-resource-groups/README.md
Singularity23x0: I'll reword.

keps/11823-effective-resource-groups/README.md
Singularity23x0: Fixed.

keps/11823-effective-resource-groups/README.md
Singularity23x0: In 
```go
func (r *cqReconciler) syncEffectiveQuotaToSpec(ctx context.Context, cq *kueue.ClusterQueue, reason, message string) error {
	needsUpdate := queue.SyncEffectiveResourceGroupToSpec(cq)
         // ...
}

```

keps/11823-effective-resource-groups/README.md
Singularity23x0: That line should have ended before this section. I think this part of the sentence was added by accident. Removing.

keps/11823-effective-resource-groups/README.md
Singularity23x0: Removed.

keps/11823-effective-resource-groups/README.md
Singularity23x0: IIUC there are places, like the scheduler AC populating and checking logic, which could be affected by RG being empty when reached by Kueue. To be investigated in the second stage of implementation, as we adjust code and tests for ERG feature gate to be enabled by default.

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,277 @@
...
+
+<!--
+A table of contents is helpful for quickly jumping to sections of a KEP and for
+highlighting any additional information provided beyond the standard KEP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
```
olekzabl: And similarly below

```suggestion
```

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,277 @@
...
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+This KEP extends the current resource group definition mechanism of Kueue to accommodate automated quota management.
+
+The proposal is to introduce a new ClusterQueue Status field: *EffectiveResourceGroups*, used to represent and track the ClusterQueue quota as interpreted by the Kueue scheduler.
```
olekzabl: Why rendering *EffectiveResourceGroups* and *spec.ResourceGroups* in italics?

It's way more standard to write `EffectiveResourceGroups` and `Spec.ResourceGroups`, with backticks.

(BTW in the KEP title I'd leave it unformatted).

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,277 @@
...
+When automating effective quota calculation, using *spec.ResourceGroups* to store the value would mean allowing Kueue to modify spec of objects it's supposed to reconcile (specifically, ClusterQueues). This may be perceived as a design smell because:
+* It blurs the line between configuration (managed by users) and state (managed by controllers).
+* It brings a risk of an infinite reconciling loop.
+
+There are two ongoing initiatives aiming to automate ClusterQueue quota management:
+1. [KEP 8826](https://github.com/kubernetes-sigs/kueue/pull/8864)
+1. [KEP 9988](https://github.com/kubernetes-sigs/kueue/tree/main/keps/9988-multikueue-manager-quota-automation#move-the-aggregated-quota-out-of-clusterqueuespec)
+
+To resolve this issue and allow the aforementioned initiatives to be implemented in a robust and clean way, we propose introducing a new ClusterQueue Status field: *EffectiveResourceGroups* to store the automated quota value.
+
```
olekzabl: I'm feeling this will make things read more "in order", without surprises on the way.
(Sadly, the displayed diff is too large)

```suggestion
We need a way to track the ClusterQueue quota. In the default case the quota is static and described by the user in *spec.ResourceGroups*. However, there are two ongoing initiatives aiming to automate ClusterQueue quota management ([KEP 8826](https://github.com/kubernetes-sigs/kueue/pull/8864)
and [KEP 9988](https://github.com/kubernetes-sigs/kueue/tree/main/keps/9988-multikueue-manager-quota-automation#move-the-aggregated-quota-out-of-clusterqueuespec)).

When automating effective quota calculation, using *spec.ResourceGroups* to store the value would mean allowing Kueue to modify spec of objects it's supposed to reconcile (specifically, ClusterQueues). This may be perceived as a design smell because:
* It blurs the line between configuration (managed by users) and state (managed by controllers).
* It brings a risk of an infinite reconciling loop.

To resolve this issue and allow the aforementioned initiatives to be implemented in a robust and clean way, we propose introducing a new ClusterQueue Status field: *EffectiveResourceGroups* to store the automated quota value.

```

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,277 @@
...
+2. When the MultiKueue ClusterQueue Controller is active, it will handle all ClusterQueues with the Multikueue Admission Check enabled. 
+When the automation is **enabled** and **configured correctly** for a given MK CQ, *EffectiveResourceGroups* will be set directly to a value representing the aggregated quota across the Manager Queue's Workers, calculated as decided upon in [issue 11862](https://github.com/kubernetes-sigs/kueue/issues/11862). Otherwise it will be synced to *spec.ResourceGroups* as well.
+
+The MultiKueue ClusterQueue Controller will be created only when the following features are enabled:
+1. feature.MultiKueue
+1. feature.MultiKueueManagerQuotaAutomation
+1. feature.**EffectiveResourceQuotas**
+
+We consider MultiKueue Automated Quota Management **enabled** and **configured correctly** when:
+1. The Quota Management Strategy is present (non-nil) and set to **"Automated"**.
```
olekzabl: That way, we'll abstract from the problem of the default value, which belongs to the scope of KEP-9988.

Whatever treatment for defaulting we define there - the phrasing here will remain correct.

```suggestion
1. The Quota Management Strategy value is **"Automated"**.
```

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,277 @@
...
+1. **spec.ResourceGroup is empty.**
```
olekzabl:
```suggestion
1. **spec.ResourceGroups is empty.**
```

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,277 @@
...
+
+We introduce the new *status.EffectiveResourceGroups* field alongside the *spec.ResourceGroups* field.
+The *status.EffectiveResourceGroups* will add a layer between the *spec.ResourceGroups* (set directly by users) and Kueue logic. The field will always be populated. By default it will reflect the value set in spec.
+
+All internal Kueue logic will use either the *spec.ResourceGroups* or the *status.EffectiveResourceGroups* as a source of truth for the ClusterQueue's quota, based on whether the feature is enabled.
+
+Setting the value of the effective quota will be handled by the Core-ClusterQueue-Controller and MultiKueue-ClusterQueue-Controller depending on the configuration:
+
+1. By default, *EffectiveResourceGroups* will always mirror *spec.ResourceGroups*. *EffectiveResourceGroups* will be synced to *spec.ResourceGroups* during reconciliation in the Core ClusterQueue Controller.
+This coveres both the cases when the MultiKueue ClusterQueue Controller is not created, and when it is but the ClusterQueue is not a MultiKueue CQ (does not have the MK AdmissionCheck).
```
olekzabl:
```suggestion
This covers both the cases when the MultiKueue ClusterQueue Controller is not created, and when it is but the ClusterQueue is not a MultiKueue CQ (does not have the MK AdmissionCheck).
```

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,277 @@
...
+2. When the MultiKueue ClusterQueue Controller is active, it will handle all ClusterQueues with the Multikueue Admission Check enabled. 
+When the automation is **enabled** and **configured correctly** for a given MK CQ, *EffectiveResourceGroups* will be set directly to a value representing the aggregated quota across the Manager Queue's Workers, calculated as decided upon in [issue 11862](https://github.com/kubernetes-sigs/kueue/issues/11862). Otherwise it will be synced to *spec.ResourceGroups* as well.
+
+The MultiKueue ClusterQueue Controller will be created only when the following features are enabled:
+1. feature.MultiKueue
+1. feature.MultiKueueManagerQuotaAutomation
+1. feature.**EffectiveResourceQuotas**
```
olekzabl:
```suggestion
The MultiKueue ClusterQueue Controller will be created only when the following feature gates are enabled:
1. MultiKueue
1. MultiKueueManagerQuotaAutomation
1. **EffectiveResourceQuotas**
```

keps/11823-effective-resource-groups/README.md
olekzabl: You'll say this again in "Design details", so maybe no need to repeat here?
Maybe you could just link to there, possibly separating out a subsection, "Responsible controller" or so?

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,277 @@
...
+
+## Design Details
+
+*status.EffectiveResourceGroups* will be managed by either the Core ClusterQueue Controller or the MultiKueue ClusterQueue Controller (mutually exclusive):
+1. *status.EffectiveResourceGroups* will be managed by the MultiKueue Controller when (AND):
+	* The MultiKueue ClusterQueue Controller is active:
+    * features.MultiKueue is enabled
+    * features.MultiKueueManagerQuotaAutomation is enabled
+    * features.**EffectiveResourceQuotas** is enabled
+  * ClusterQueue has the MultiKueue admission check assigned
```
olekzabl: This mis-renders; I guess you need to fix indents for the `*`-items.

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,277 @@
...
+		cq.*status.EffectiveResourceGroups* = cq.*spec.ResourceGroups*
+	}
+	return
+}
+```
+
+### Core Controller
+
+The Core Controller is the default controller handling the new parameter.
+The reconciliation logic will check if the ClusterQueue is subject to the default handling scheme and perform the **Default Behavior** if applicable.
```
olekzabl:
```suggestion
The reconciliation logic will check if the ClusterQueue is subject to the default handling scheme and perform the [Default Behavior](#default-behavior) if applicable.
```

keps/11823-effective-resource-groups/README.md
olekzabl: Ahh, `needsUpdate` is output, not input. I see.

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,277 @@
...
+```
+
+### MultiKueue quota automation
+
+When feature.MultiKueue and feature.MultiKueueManagerQuotaAutomation are enabled, the MultiKueue Controller will be created. We will delegate the handling of MultiKueue ClusterQueues when it is set up. The Controller itself will:
+1. Check if the ClusterQueue has the MultiKueue admission check assigned. If not: it will skip the CQ as it is not considered an MK CQ.
+1. Check if:
+    1. the ClusterQueue's MultiKueue Config has the **Quota Management Strategy** set to "Automated",
+    1. *spec.ResourceGroups* is empty.
+1. If any of the above conditions are false: it will perform the **Default Behavior**.
```
olekzabl: Yes, but the 2 cases differ in the treatment of quota management Condition.
(In the former case, the status is "not requested"; in the latter, it's "unsupported configuration").

I feel it might be worth mentioning - but up to you.

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,277 @@
...
+    * features.MultiKueue is enabled
+    * features.MultiKueueManagerQuotaAutomation is enabled
+    * features.**EffectiveResourceQuotas** is enabled
+  * ClusterQueue has the MultiKueue admission check assigned
+2. Otherwise, *status.EffectiveResourceGroups* will be managed by the Core Controller.
+
+The *status.EffectiveResourceGroups* will be used to provide resource quota only when the **EffectiveResourceQuotas** feature is enabled. Otherwise we will default to using the spec. To make this feasible, all effective calls to retrieve the ResourceGroups will be handled by a dedicated method:
+
+```go
+func GetEffectiveResourceGroup(cq *kueue.ClusterQueue) []kueue.ResourceGroup {
```
olekzabl:
```suggestion
func GetEffectiveResourceGroups(cq *kueue.ClusterQueue) []kueue.ResourceGroup {
```

keps/11823-effective-resource-groups/README.md
olekzabl: I'm still feeling this exact risk is relatively minor.
_Maybe_ there are some places in the code currently assuming otherwise (i.e. that `Spec.ResourceGroups` is not empty). However:

- they shouldn't be too numerous, as generally the validation rules allow that
- they'll need to be refactored into `GetEffectiveResourceGroups` anyway

TBH I'm more worried about some related-but-still-different risks, that is:

1. The need to refactor `.Spec.ResourceGroups` -> `GetEffectiveResourceGroups` everywhere.
   This feels doable; yet, if we introduce bugs, they'll touch the foundations of Kueue.
   Here, the main mitigation is flipping behind a feature gate.

2. The fact that `.Status.EffectiveResourceGroups` (our _source of truth_) will be _transiently_ empty following CQ creation.
   And it's here where I imagine we'll need to carefully check things don't explode, or misbehave.
   Here I think the main mitigation is to make sure that:
   - `SyncEffectiveResourceGroupsToSpec` happens early enough in the controllers' logic that it can't be derailed by any potential error caused by "effective quotas being empty"
   - the status change triggered by `SyncEffectiveResourceGroupsToSpec` triggers another round of relevant reconcilers (which I hope it will out of the box, but we'll need to verify)

Thoughts?

keps/9988-multikueue-manager-quota-automation/README.md
```diff
@@ -414,12 +414,45 @@ lastTransitionTime: 2026-01-01T00:00:00Z
 
 #### Planned changes in Alpha2 version
 
-Starting from Alpha2 version, we plan to introduce a **new field** in ClusterQueueStatus (provisionally named `effectiveResourceGroups`), which would act as the **new source of truth** about a ClusterQueue capacity (taking over that role from `.spec.resourceGroups`). This is discussed in more detail in [this future work idea](#move-the-aggregated-quota-out-of-clusterqueuespec).
+Starting from Alpha2 version, we plan to introduce a **new field** in ClusterQueueStatus (provisionally named `effectiveResourceGroups`), which would act as the **new source of truth** about a ClusterQueue capacity (taking over that role from `.spec.resourceGroups`). This is discussed in more detail in [KEP 11823](keps/11823-effective-resource-groups).
 
-While this move is mostly independent from the [API for Alpha1](#alpha1-version), it may lead to some slight changes in that API:
+With the introduction of the new requirement for MultiKueue ClusterQueues with quota automation enabled, namely that the spec.ResourceGroups be empty in that case, two major change will be introduced to the quota automation condition:
```
olekzabl:
```suggestion
With the introduction of the new requirement for MultiKueue ClusterQueues with quota automation enabled, namely that the spec.ResourceGroups be empty in that case, two major changes will be introduced to the quota automation condition:
```

keps/9988-multikueue-manager-quota-automation/README.md
```diff
@@ -414,12 +414,45 @@ lastTransitionTime: 2026-01-01T00:00:00Z
...
+	// automatedQuotaManagementConfig is the configuration for the automated quota management.
+	// It is only relevant when quotaManagementMode is set to `Automated`.
+	// +optional
+	AutomatedQuotaManagementConfig *AutomatedQuotaManagementConfig `json:"automatedQuotaManagementConfig,omitempty"`
+}
+
+// AutomatedQuotaManagementConfig is the configuration for the automated quota management in MultiKueue ClusterQueues.
+type AutomatedQuotaManagementConfig struct {
+	// aggregateResourceFlavorSpec is the specification of the resource flavor MultiKueue creates and maintains for the auto-aggregated resource flavors 
+	AggregateResourceFlavorSpec ResourceFlavorSpec `json:"aggregateResourceFlavorSpec,omitempty"`
```
olekzabl: Why not just a ResourceFlavorReference?

In my feeling, it would:
- be more in the spirit of the original `ResourceGroups` format
- when actually reusing existing flavors, make the configuration simpler (use a single string, don't rewrite topology label names, etc.)

keps/9988-multikueue-manager-quota-automation/README.md
```diff
@@ -414,12 +414,45 @@ lastTransitionTime: 2026-01-01T00:00:00Z
 
 #### Planned changes in Alpha2 version
 
-Starting from Alpha2 version, we plan to introduce a **new field** in ClusterQueueStatus (provisionally named `effectiveResourceGroups`), which would act as the **new source of truth** about a ClusterQueue capacity (taking over that role from `.spec.resourceGroups`). This is discussed in more detail in [this future work idea](#move-the-aggregated-quota-out-of-clusterqueuespec).
+Starting from Alpha2 version, we plan to introduce a **new field** in ClusterQueueStatus (provisionally named `effectiveResourceGroups`), which would act as the **new source of truth** about a ClusterQueue capacity (taking over that role from `.spec.resourceGroups`). This is discussed in more detail in [KEP 11823](keps/11823-effective-resource-groups).
```
olekzabl: In `#move-the-aggregated-quota-out-of-clusterqueuespec`, please refer to your KEP as well.

keps/11823-effective-resource-groups/README.md
Singularity23x0: Re. 1: a proposal (including the use of the resource flag) is available at https://github.com/kubernetes-sigs/kueue/pull/11807 (WIP)

Re. 2: I am currently working out how to do that safely.

Added a mention on the overall issue in risks and drawbacks.

keps/9988-multikueue-manager-quota-automation/README.md
Singularity23x0: 1. My worry is about https://github.com/kubernetes-sigs/kueue/issues/11862. In the future, we would have to consider defining multiple flavors on the manager end if we go withe the reference ide. Providing the spec allows to do a copy-paste for all flavors in that case.
2. Yes, if the user has flavors already in mind for MK then it would be simpler. But if the user wants to create a dedicated flavor for the automated quota, one with a specific specification, we don't need to require the creation of the flavor, just providing us with the bare minimum data they need and automation will handle the rest on its own. In that case the programmatic part of the users job is the same, but we remove the added necessity of users also creating and managing said flavor.

All in all until we have to address [1] the matter is one of preference and I am okay with either approach based on feedback, but for the sake of considering per flavor approach in the followup to this kep I would lean towards the current proposal.

Singularity23x0: On the other hand - there remains the matter of admission checks and their supported flavors.
If the user does not setup their Flavors, and they are added later on (by MK), the MK AC cannot have any flavors set, not even by default (since they do not exist as of yet). This means the logic of the AC setup would have to be updated as well, to account for "dynamic flavors". I will add an explicit mention of this issue to the KEP.

keps/9988-multikueue-manager-quota-automation/README.md
```diff
@@ -414,12 +416,64 @@ lastTransitionTime: 2026-01-01T00:00:00Z
...
+
+```go
+// AutomatedQuotaManagementConfig is the configuration for the automated quota management in MultiKueue ClusterQueues.
+type AutomatedQuotaManagementConfig struct {
+	// aggregateResourceFlavorRef is the reference to the resource flavor MultiKueue will use as an aggregated flavor for the quotas gathered from Worker Clusters
+	AggregateResourceFlavorRef *ResourceFlavorReference `json:"aggregateResourceFlavorRef,omitempty"`
+}
+```
+
+###### Dynamic Resource Flavor
```
olekzabl: I suppose I don't understand your thinking behind this Dynamic option.

1. You mention it being somehow more appealing for per-flavor aggregation; yet, the definition below still deals with just one flavor. Typo in definition? Or how is it supposed to work?

2. In the discussion of this variant, I see you mentioned some added complexity ("all components (...) will have to be made aware(...)") - but I don't see any benefit explicitly called out. (Neither can I clearly see myself what this benefit is :)

keps/9988-multikueue-manager-quota-automation/README.md
olekzabl: I'm sorry, I'm not following, in a few ways :|
(cf. my other comment on the new version, with some other questions)

> In the future, we would have to consider defining multiple flavors on the manager end if we go withe the reference ide. Providing the spec allows to do a copy-paste for all flavors in that case.

By "providing the spec", you mean putting a `ResourceFlavorSpec` somewhere in the quota automation config? If yes, how is it noticeably easier or simpler than defining a new `ResourceFlavor`?

Also, what do you mean by "allows to do a copy-paste"? Copy-paste from where, from workers? If so, then the option with `ResourceFlavor` is the one truly copy-pasting; the "`ResourceFlavorSpec` in MK config" approach would require extracting `Spec`s from worker-side Flavors.

> we don't need to require the creation of the flavor, just providing us with the bare minimum data they need and automation will handle the rest on its own

What is "the bare minimum", and what is "the rest"?

> users also creating and managing said flavor.

What's the "managing" that would be required in "Manual" but not in "Dynamic"?

> If the user does not setup their Flavors, and they are added later on (by MK), the MK AC cannot have any flavors set

The MK AC cannot have any flavors set anyway; it's [enforced by validation](https://github.com/kubernetes-sigs/kueue/blob/d968cfde57b256eefc93e564306222446ce92b87/pkg/cache/scheduler/clusterqueue.go#L331).

keps/9988-multikueue-manager-quota-automation/README.md
Singularity23x0: Updated these sections, ptal.

keps/9988-multikueue-manager-quota-automation/README.md
```diff
@@ -414,12 +415,49 @@ lastTransitionTime: 2026-01-01T00:00:00Z
...
+1. The misconfiguration based on difference between the resources reported in spec and gathered from remote workers will be removed. With spec.ResourceGroups expected to be empty, the resources reported by workers will be considered the source of truth by default.
+2. The "MultiKueue manager quota automation does not support ClusterQueues with multiple ResourceFlavors." condition will be replaced with "ResourceGroups set manually in ClusterQueue spec." when any resources are set in spec.
+
+Another matter to address will be the aggregated quota flavors. As of Alpha2 we expect only one, collective resource flavor to be used. To allow users to set up the details of said flavor we will add an automated quota config field to the MultiKueue Config: AutomatedQuotaManagementConfig.
+
+```go
+// MultiKueueConfigSpec defines the desired state of MultiKueueConfig
+// +kubebuilder:validation:XValidation:rule="has(obj.quotaManagementMode) && obj.quotaManagementMode == 'Automated' ? has(obj.automatedQuotaManagementConfig) : true",message="automatedQuotaManagementConfig is only relevant when quotaManagementMode is set to 'Automated'"
```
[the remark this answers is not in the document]
Singularity23x0: It can be defined in the other case, it will simply not be used.

keps/11823-effective-resource-groups/kep.yaml
```diff
@@ -0,0 +1,33 @@
+title: Effective Resource Groups
+kep-number: 11823
+authors:
+  - "@dpastuszka"
+status: implementable
+creation-date: 2026-05-29
+reviewers:
+  - TBD
```
tenzen-y:
```suggestion
  - @olekzabl 
  - @tenzen-y
```

nit: update KEP reviewers.

keps/11823-effective-resource-groups/kep.yaml
```diff
@@ -0,0 +1,33 @@
...
+approvers:
+  - TBD
```
tenzen-y:
```suggestion
  - @tenzen-y
```

keps/11823-effective-resource-groups/kep.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. This can be the current (upcoming) milestone, if it is being actively
+# worked on.
+latest-milestone: "v0.19"
+
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  alpha: "v0.19"
+  beta: "v0.19"
```
tenzen-y, line 27: Does this mean that you are planning to skip alpha phase?

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,203 @@
...
+The `status.EffectiveResourceGroups` will be used to provide resource quota only when the **EffectiveResourceQuotas** feature is enabled. Otherwise we will default to using the spec. To make this feasible, all effective calls to retrieve the ResourceGroups will be handled by a dedicated method:
+
+```go
+func GetEffectiveResourceGroups(cq *kueue.ClusterQueue) []kueue.ResourceGroup {
+	if features.Enabled(features.EffectiveResourceQuotas) {
+		return cq.Status.EffectiveResourceGroups
+	}
+	return cq.spec.ResourceGroups
+}
+```
```
tenzen-y:
```suggestion
```

Let's focus only on specifications in KEPs to make the KEP long-living reference doc.

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,203 @@
...
+    * Default setup: MultiKueue disabled; this case will be the basis for any possible future additions of more automated ClusterQueue quota management schemes,
+    * MultiKueue setup without Quota automation: MultiKueue enabled, MultiKueueManagerQuotaAutomation feature disabled.
+    * MultiKueue with MultiKueueManagerQuotaAutomation enabled.
+
+## Proposal
+
+We introduce the new `status.EffectiveResourceGroups` field alongside the `spec.ResourceGroups` field.
+The `status.EffectiveResourceGroups` will add a layer between the `spec.ResourceGroups` (set directly by users) and Kueue logic. The field will always be populated. By default it will reflect the value set in spec.
+
+All internal Kueue logic will use either the `spec.ResourceGroups` or the `status.EffectiveResourceGroups` as a source of truth for the ClusterQueue's quota, based on whether the feature is enabled.
```
tenzen-y:
```suggestion
All internal Kueue logic will use either the `spec.ResourceGroups` or the `status.EffectiveResourceGroups` as a source of truth for the ClusterQueue's quota, based on whether the `EffectiveResourceQuotas` featureGate is enabled.
```

Does this mean that status will be considered only when EffectiveResourceQuotas feature gate is enabled? Or does this mean you are planning to add another knob to enable this feature?

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+1. By default, `EffectiveResourceGroups` will always mirror `spec.ResourceGroups`. `EffectiveResourceGroups` will be synced to `spec.ResourceGroups` during reconciliation in the Core ClusterQueue Controller.
+This covers both the cases when the MultiKueue ClusterQueue Controller is not created, and when it is but the ClusterQueue is not a MultiKueue CQ (does not have the MK AdmissionCheck).
+2. When the MultiKueue ClusterQueue Controller is active, it will handle all ClusterQueues with the Multikueue Admission Check enabled. 
+When the automation is **enabled** and **configured correctly** for a given MK CQ, `EffectiveResourceGroups` will be set directly to a value representing the aggregated quota across the Manager Queue's Workers, calculated as decided upon in [issue 11862](https://github.com/kubernetes-sigs/kueue/issues/11862). Otherwise it will be synced to `spec.ResourceGroups` as well.
+
+The MultiKueue ClusterQueue Controller will be created only when the following feature gates are enabled:
+1. MultiKueue
+1. MultiKueueManagerQuotaAutomation
+1. **EffectiveResourceQuotas**
```
tenzen-y:
```suggestion
1. MultiKueue
2. MultiKueueManagerQuotaAutomation
3. **EffectiveResourceQuotas**
```

I know that GH markdown viewer automatically increment item numbers, but I think that it would be better to avoid depending on the non-std markdown specifications.

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+We consider MultiKueue Automated Quota Management **enabled** and **configured correctly** when:
+1. The Quota Management Strategy value is **"Automated"**.
+1. **spec.ResourceGroups is empty.**
```
tenzen-y:
```suggestion
1. The Quota Management Strategy value is **"Automated"**.
2. **spec.ResourceGroups is empty.**
```

ditto

keps/11823-effective-resource-groups/README.md
tenzen-y: What if all FGs are enabled, but MultiKueue is not enabled? Which resource groups (spec vs status) are effective? 

I'm wondering if we should decouple this EffectiveResourceGroups feature from MultiKueue usages, because if the external controller can reflect their requirement (e.g., add or subtract some system overhead), it would be useful. I know such stories in some companies.

But, I'm ok with excluding non-mk usage for this EffectiveResourceGroups in the alpha stage. If you agree with revisiting this non-mk usage in beta stage, please add it to future works section.

keps/11823-effective-resource-groups/README.md
tenzen-y: Does this mean that we will add webhook or marker based valiadations so that users or controllers can not specify `.spec.resourceGroups` under this situation?

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,203 @@
...
+```go
+// pkg/util/queue/cluster_queue.go
+
+func SyncEffectiveResourceGroupsToSpec(cq *kueue.ClusterQueue) (needsUpdate bool) {
+	if needsUpdate = !equality.Semantic.DeepEqual(cq.Status.EffectiveResourceGroups, cq.Spec.ResourceGroups); needsUpdate {
+		cq.Status.EffectiveResourceGroups = cq.Spec.ResourceGroups
+	}
+	return
+}
+```
```
tenzen-y:
```suggestion
```

ditto: implementations in KEP.

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,203 @@
...
+		return ctrl.Result{}, err
+	} else if isMK {
+		log.V(2).Info("Skipping EffectiveResourceGroups sync: MultiKueue Manager ClusterQueue quota is managed by a dedicated MultiKueue controller.")
+	} else if queue.SyncEffectiveResourceGroupsToSpec(&cqObj) {
+		log.V(2).Info("Syncing EffectiveResourceGroups to spec.")
+		if err := r.client.Status().Update(ctx, &cqObj); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+```
```
tenzen-y:
```suggestion
```

ditto: implementations in KEP.

keps/11823-effective-resource-groups/README.md
tenzen-y, line 1: Can you add user stories for your current situation?

tenzen-y, line 1: Can you add .status.resourceGroups API proposal to the design section?

tenzen-y (commented):
Thank you for moving this forward, I left some comments.

keps/11823-effective-resource-groups/kep.yaml
Singularity23x0: The idea here was to split the impl based on the compatibility of the feature gate locked logic with default use cases.
The main idea was to give us a milestone that would be a failsafe allowing us to ship the feature required by MK Quota Automation and move the work forward there.
If we manage time-wise - I would like to ship beta by 0.19 as well.
So, in essence: form the user (release) perspective - yes.

keps/11823-effective-resource-groups/README.md
Singularity23x0: Yes, the EffectiveResourceQuotas fg is there to lock all access (set and get alike) to the status.EffectiveResourceGroups field.

keps/11823-effective-resource-groups/README.md
Singularity23x0: `What if all FGs are enabled, but MultiKueue is not enabled?`
Then the status field will be used via the `Default` behavior.

In short

The FG tells us whether the effective rgs are in the spec or in status (and makes kueue populate them accordingly).

If the FG is enabled, we will also allow the users of MK to enable automation. The EffectiveResourceQuota FG is independent. I is only the MK Automation feature that will become dependent on the ERQ FG.

keps/11823-effective-resource-groups/README.md
Singularity23x0: The idea was to use the existing validation scheme, a.k.a. we will refuse to auto aggregate in MK and set the Condition of the CQ to automation false because UnsupportedConfiguration.

keps/11823-effective-resource-groups/README.md
Singularity23x0: On it

Singularity23x0: On it

keps/9988-multikueue-manager-quota-automation/README.md
olekzabl, line 222: This part of the example will need to be updated.
(Not just in place. We'll have to offer 2 variants of the output, depending on whether ERG FG is enabled).

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,189 @@
...
+	// +kubebuilder:validation:MaxItems=16
+	// +optional
+	EffectiveResourceGroups []ResourceGroup `json:"effectiveResourceGroups,omitempty"`
+
+	// [...]
+
+}
+```
+
+All internal Kueue logic will use either the `spec.ResourceGroups` or the `status.EffectiveResourceGroups` as a source of truth for the ClusterQueue's quota, based on whether the `EffectiveResourceQuotas` featureGate is enabled.
```
olekzabl, line 79: Huh, there's one tricky case for this rule, namely... MultiKueue quota automation :)

Currently, we fetch remote `ClusterQueue` objects and check their `.Spec.NominalQuota`.

Now, we'll sometimes need to switch to effective quotas on the workers.
(Not even due to MK automation, as MK can't be chained - but possibly due to node capacity reader ([KEP-10270](https://github.com/kubernetes-sigs/kueue/pull/10745)) or hero jobs ([KEP-8826](https://github.com/kubernetes-sigs/kueue/pull/8864)) operating on the workers).

And now fun part: we'll need to use worker's effective quotas if effective quotas are enabled... where?
I guess on that worker?

I roughly see 2 options:

1. Let the manager check the FG state on each worker, and act accordingly.
   * I'm not sure how feasible, or how easy, this is.
     Keep in mind FGs can be switched not just in the `Configuration` object but also via command-line args.
     (Also, even `Configuration` lives outside `etcd`, right?)
     * Would we need to make a list of FGs an annotation on the `kueue-controller-manager` Pod?
       (Updated by that Pod itself, on its startup).
       So that the manager can read it from workers' `etcd` objects?
       Would such a trick be viable if that Pod is managed e.g. by a Deployment?

2. Let the manager check if `effectiveResourceGroups` field exists on the remote CQ, and fall back to spec if it doesn't.
   * That's easier to implement - but a bit hacky, and potentially unsafe.
     For example, every newly created remote CQ will have no `effectiveResourceGroups` for a short initial period, even if the FG is enabled.
     Then, if [KEP-10270](https://github.com/kubernetes-sigs/kueue/pull/10745) goes for its "Option C" (see [here](https://github.com/kubernetes-sigs/kueue/pull/10745#discussion_r3403835468)), it could lead to a burst of over-booking on the manager side!

keps/11823-effective-resource-groups/kep.yaml
tenzen-y: One of the feature gate objectives is avoiding enabling premature features by default.
It seems that this is not a trivial effort, and obviously has an effect on the user's production.

So, I think that we should start from the alpha stage, and spec should be used by default.
In 0.20, we can graduate it with MultiKueueManagerQuotaAutomation to beta.

amy:
@Singularity23x0 interested in reviewing this kep. Adding this comment here so we don't merge before I take a pass. Thank you!

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+This KEP extends the current resource group definition mechanism of Kueue to accommodate automated quota management.
+
+The proposal is to introduce a new ClusterQueue Status field: `EffectiveResourceGroups`, used to represent and track the ClusterQueue quota as interpreted by the Kueue scheduler.
+
+## Motivation
+
+We need a way to track the ClusterQueue quota. In the default case the quota is static and described by the user in `spec.ResourceGroups`. There are however two ongoing initiatives aiming to automate ClusterQueue quota management:
+1. [KEP 8826](https://github.com/kubernetes-sigs/kueue/pull/8864)
+1. [KEP 9988](https://github.com/kubernetes-sigs/kueue/tree/main/keps/9988-multikueue-manager-quota-automation#move-the-aggregated-quota-out-of-clusterqueuespec)
```
amy, line 37: Make the visible text for these links to say the name of the keps. Otherwise I have to click something else to get full context. ty!

keps/11823-effective-resource-groups/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+Setting the value of the effective quota will be handled by the Core-ClusterQueue-Controller and MultiKueue-ClusterQueue-Controller depending on the configuration:
+
+1. By default, `EffectiveResourceGroups` will always mirror `spec.ResourceGroups`. `EffectiveResourceGroups` will be synced to `spec.ResourceGroups` during reconciliation in the Core ClusterQueue Controller.
+This covers both the cases when the MultiKueue ClusterQueue Controller is not created, and when it is but the ClusterQueue is not a MultiKueue CQ (does not have the MK AdmissionCheck).
+2. When the MultiKueue ClusterQueue Controller is active, it will handle all ClusterQueues with the Multikueue Admission Check enabled. 
+When the automation is **enabled** and **configured correctly** for a given MK CQ, `EffectiveResourceGroups` will be set directly to a value representing the aggregated quota across the Manager Queue's Workers, calculated as decided upon in [issue 11862](https://github.com/kubernetes-sigs/kueue/issues/11862). Otherwise it will be synced to `spec.ResourceGroups` as well.
+
+The MultiKueue ClusterQueue Controller will be created only when the following feature gates are enabled:
```
amy, line 88: As a high level comment... I think we should go back to the drawing board and build out a full CRD API surface that interacts with this status with an easy override flow.

Looking at: https://github.com/kubernetes-sigs/kueue/issues/12382#issue-4701134859

You have 4 scenarios that need to interact with some "internal" idea of nominalquota that differs from what is set in the user facing nominalQuota: https://github.com/kubernetes-sigs/kueue/issues/12382#issuecomment-4781764553

For MultiKueue and DRA, we should consider overrides. For UberQueue and Temporary Quota Overrides, I assume admins shouldn't be able to interact with this outside of the dedicated new CRD. 

I also want to understand if we turn all of these on, how they interact with each other. So you'll need provenance and interaction descriptions. 

Overall, what kueue does with quota is generally confusing to users. So I'd also like the design to surface dedicate events or conditions where its easy to see from the CQ object that the quota is now being managed by something else.

amy: Also. We need a (short) dedicated metrics section. That needs to come out of the box. Because most people I interact with have dashboards built out already that are extremely dependent on today's nominalquota.

Singularity23x0: Hi Amy, will do.

For context: this KEP was created before #12382 (and was part of what sparked the considerations leading to the creation of that issue).
It will be adjusted based on the design we decide upon in #12382. Will make sure to take all feedback from here into account when that happens.

amy: solid! Please tag me when its ready.

PR #11867, opened 2026-06-01
fix status update for local job

### LINKED ISSUE #11115: MultiKueue when Preemption with a multikueue admission check Should preempt a running low-priority workload on a worker when a high-priority workload is admitted to it by manager
**Which test is flaking?**:

MultiKueue when Preemption with a multikueue admission check Should preempt a running low-priority workload on a worker when a high-priority workload is admitted to it by manager

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10357/pull-kueue-test-e2e-multikueue-main/2053781622058127360
**Failure message or logs**:
```
End To End MultiKueue Suite: kindest/node:v1.35.0: [It] MultiKueue when Preemption with a multikueue admission check Should preempt a running low-priority workload on a worker when a high-priority workload is admitted to it by manager expand_less	17s
{ failed [FAILED] Timed out after 10.001s.
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/multikueue/baseline/e2e_test.go:1638 with:
Expected
    <int32>: 0
to equal
    <int32>: 1
In [It] at: /home/prow/go/src/sigs.k8s.io/kueue/test/e2e/multikueue/baseline/e2e_test.go:1639 @ 05/11/26 10:33:11.632
}
```

andrewseif:
#### What type of PR is this?
/kind failing-test
/kind flake
/kind bug

Fix the "remaining case" (local Job is suspended, remote Job is unsuspended but not finished) in pkg/controller/jobs/job/job_multikueue_adapter.go:149, where the worker sends exactly one status update while the the local is still suspended.

Fixes #11115 

ill attach my analysis and logs used in identifying the bug.
I understand that this can but classified as a not-bug but intended use case, in that situation we might want to change the test parameters? 

```release-note
MultiKueue: Fixed a bug that could leave stale status for Kubernetes Jobs in the manager
cluster when the worker-cluster Job reached steady state quickly and stopped getting
updates while the manager-cluster Job was still suspended.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved multi-queue job status synchronization when a local job is temporarily suspended: the controller now treats this as an intentional deferred sync and retries status reconciliation on a shorter interval instead of waiting for the usual timeout.
  * Updated eviction/reservation reconciliation paths to correctly handle deferred sync behavior without surfacing it as an error.

* **Tests**
  * Added regression coverage to confirm deferred sync results in the shorter requeue timing.

### DIFF
--- a/pkg/controller/admissionchecks/multikueue/externalframeworks/adapter.go
+++ b/pkg/controller/admissionchecks/multikueue/externalframeworks/adapter.go
@@ -57,13 +57,13 @@ func NewAdapter(gvk schema.GroupVersionKind) jobframework.MultiKueueAdapter {
 // SyncJob synchronizes a job resource between the local and remote clusters.
 // It ensures that the remote cluster has a corresponding object for the local job,
 // creating it if necessary, and updates the status of the local object based on the remote object.
-func (a *Adapter) SyncJob(ctx context.Context, localClient client.Client, remoteClient client.Client, key types.NamespacedName, workloadName, origin string) error {
+func (a *Adapter) SyncJob(ctx context.Context, localClient client.Client, remoteClient client.Client, key types.NamespacedName, workloadName, origin string) (deferred bool, err error) {
 	// Get the local object
 	localObj := &unstructured.Unstructured{}
 	localObj.SetGroupVersionKind(a.gvk)
-	err := localClient.Get(ctx, key, localObj)
+	err = localClient.Get(ctx, key, localObj)
 	if err != nil {
-		return err
+		return false, err
 	}
 
 	// Check if remote object already exists
@@ -71,16 +71,16 @@ func (a *Adapter) SyncJob(ctx context.Context, localClient client.Client, remote
 	remoteObj.SetGroupVersionKind(a.gvk)
 	err = remoteClient.Get(ctx, key, remoteObj)
 	if err != nil && !apierrors.IsNotFound(err) {
-		return err
+		return false, err
 	}
 
 	if apierrors.IsNotFound(err) {
 		// Create new remote object
-		return a.createRemoteObject(ctx, remoteClient, localObj, workloadName, origin)
+		return false, a.createRemoteObject(ctx, remoteClient, localObj, workloadName, origin)
 	}
 
 	// Update existing remote object status
-	return a.syncStatus(ctx, localClient, localObj, remoteObj)
+	return false, a.syncStatus(ctx, localClient, localObj, remoteObj)
 }
 
 func (a *Adapter) createRemoteObject(ctx context.Context, remoteClient client.Client, localObj *unstructured.Unstructured, workloadName, origin string) error {
--- a/pkg/controller/admissionchecks/multikueue/workload.go
+++ b/pkg/controller/admissionchecks/multikueue/workload.go
@@ -63,6 +63,25 @@ var (
 	singleClusterPreemptionTimeout = 5 * time.Minute
 )
 
+// syncDeferredRequeueAfter is the delay used to re-reconcile a workload after
+// the Job adapter reports a deferred sync: the local Job was still suspended
+// when SyncJob ran, so the remote status.Active was not propagated. The local
+// Job is unsuspended shortly afterwards by the jobframework reconciler, but
+// that transition does not wake this reconciler (we don't watch local Jobs),
+// so without a short requeue the next reconcile could be up to workerLostTimeout
+// away and status.Active would never catch up.
+//
+// Both directions are soft: the requeue loop self-terminates as soon as the
+// local Job is unsuspended, and any value here is bounded above by
+// workerLostTimeout, so neither too-short nor too-long is a correctness
+// problem. 2 s is just a comfortable middle ground: above the sub-second
+// unsuspend latency (so we don't spin), and well below anything a user would
+// notice. See https://github.com/kubernetes-sigs/kueue/issues/11115.
+//
+// This is the canonical explanation of the deferred-sync handling; other sites
+// in this package point here rather than repeating it.
+const syncDeferredRequeueAfter = 2 * time.Second
+
 type wlReconciler struct {
 	client            client.Client
 	helper            *admissioncheck.MultiKueueStoreHelper
@@ -360,7 +379,9 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
 		// it should not be problematic, but the "From remote xxxx:" could be lost ....
 
 		if group.jobAdapter != nil {
-			if err := group.jobAdapter.SyncJob(ctx, w.client, group.remoteClients[remote].client, group.controllerKey, group.local.Name, w.origin); err != nil {
+			// The deferred flag is irrelevant here: the remote workload is
+			// Finished, so determineStatusUpdate always syncs (never defers).
+			if _, err := group.jobAdapter.SyncJob(ctx, w.client, group.remoteClients[remote].client, group.controllerKey, group.local.Name, w.origin); err != nil {
 				log.V(2).Error(err, "copying remote controller status", "workerCluster", remote)
 				// we should retry this
 				return reconcile.Result{}, err
@@ -384,10 +405,17 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
 
 		// workload evicted on manager cluster
 		if workload.IsEvicted(group.local) {
-			if err := group.jobAdapter.SyncJob(ctx, w.client, remoteCl, group.controllerKey, group.local.Name, w.origin); err != nil {
+			deferred, err := group.jobAdapter.SyncJob(ctx, w.client, remoteCl, group.controllerKey, group.local.Name, w.origin)
+			if err != nil {
 				log.Error(err, "Syncing remote controller object")
 				// We'll retry this in the next reconciling.
 				return reconcile.Result{}, err
+			}
+			if deferred {
+				// Eviction re-suspended the local Job; requeue short to retry the
+				// status sync instead of dropping it. See syncDeferredRequeueAfter.
+				log.V(3).Info("Sync deferred until local job is unsuspended; requeuing on short timer")
+				return reconcile.Result{RequeueAfter: syncDeferredRequeueAfter}, nil
 			}
 			return reconcile.Result{}, nil
 		}
@@ -493,7 +521,8 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
 			return reconcile.Result{}, err
 		}
 
-		if err := group.jobAdapter.SyncJob(ctx, w.client, remoteCl, group.controllerKey, group.local.Name, w.origin); err != nil {
+		syncDeferred, err := group.jobAdapter.SyncJob(ctx, w.client, remoteCl, group.controllerKey, group.local.Name, w.origin)
+		if err != nil {
 			log.Error(err, "Syncing remote controller object")
 			// We'll retry this in the next reconciling.
 			return reconcile.Result{}, err
@@ -502,7 +531,14 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
 		if err := w.syncReservingRemoteState(ctx, group, reservingRemote, acs); err != nil {
 			return reconcile.Result{}, err
 		}
-		return reconcile.Result{RequeueAfter: w.workerLostTimeout}, nil
+		requeueAfter := w.workerLostTimeout
+		if syncDeferred {
+			// The local Job is still suspended; requeue short so the next SyncJob
+			// can propagate the remote status. See syncDeferredRequeueAfter.
+			log.V(3).Info("Sync deferred until local job is unsuspended; requeuing on short timer")
+			requeueAfter = syncDeferredRequeueAfter
+		}
+		return reconcile.Result{RequeueAfter: requeueAfter}, nil
 	} else if acs.State == kueue.CheckStateReady {
 		// If there is no reserving and the AC is ready, the connection with the reserving remote might
 		// be lost, keep the workload admitted for keepReadyTimeout and put it back in the queue after that.
--- a/pkg/controller/admissionchecks/multikueue/workload_test.go
+++ b/pkg/controller/admissionchecks/multikueue/workload_test.go
@@ -29,6 +29,7 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/component-base/featuregate"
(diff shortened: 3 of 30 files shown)


andrewseif:
Here is my analysis 
[issue-11115-analysis.md](https://github.com/user-attachments/files/28460288/issue-11115-analysis.md)

and here are the fail logs used for reference
[mgr-kueue-ml9gk-0.log](https://github.com/user-attachments/files/28460300/mgr-kueue-ml9gk-0.log)
[mgr-kueue-w9r8h-0.log](https://github.com/user-attachments/files/28460302/mgr-kueue-w9r8h-0.log)
[w2-kueue-wl8kx-0.log](https://github.com/user-attachments/files/28460310/w2-kueue-wl8kx-0.log)
[w2-kueue-wzvnq-0.log](https://github.com/user-attachments/files/28460312/w2-kueue-wzvnq-0.log)
[w2-sched-0.log](https://github.com/user-attachments/files/28460315/w2-sched-0.log)
[w2cp-kubelet.log](https://github.com/user-attachments/files/28460316/w2cp-kubelet.log)
[11115-second-fail.txt](https://github.com/user-attachments/files/28460305/11115-second-fail.txt)
[w2-kcm-0.log](https://github.com/user-attachments/files/28460306/w2-kcm-0.log)
[w2-kubelet.log](https://github.com/user-attachments/files/28460307/w2-kubelet.log)


and this is the good run logs for comparison
[build-log.txt](https://github.com/user-attachments/files/28460335/build-log.txt)
[mgr-apiserver.log](https://github.com/user-attachments/files/28460336/mgr-apiserver.log)
[mgr-kueue-z7r8g-0.log](https://github.com/user-attachments/files/28460359/mgr-kueue-z7r8g-0.log)
[mgr-kueue-zjvpj-0.log](https://github.com/user-attachments/files/28460360/mgr-kueue-zjvpj-0.log)
[w2-apiserver.log](https://github.com/user-attachments/files/28460368/w2-apiserver.log)
[w2-kcm-0.log](https://github.com/user-attachments/files/28460377/w2-kcm-0.log)
[w2-kueue-b5kd4-0.log](https://github.com/user-attachments/files/28460379/w2-kueue-b5kd4-0.log)
[w2-kueue-xzcmd-0.log](https://github.com/user-attachments/files/28460385/w2-kueue-xzcmd-0.log)

mimowo (commented):
cc @olekzabl ptal

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -63,6 +63,15 @@ var (
 	singleClusterPreemptionTimeout = 5 * time.Minute
 )
 
+// syncDeferredRequeueAfter is the delay used to re-reconcile a workload after
+// the Job adapter returned jobframework.ErrSyncDeferred (because the local Job
+// was still suspended at the moment SyncJob was called). It must be short
+// enough that the local Job's eventual unsuspend is observed within a typical
+// e2e/operator window, but long enough to not hot-loop on the few hundred ms
+// the unsuspend PATCH normally takes. 2 s sits comfortably between the two.
+// See https://github.com/kubernetes-sigs/kueue/issues/11115.
```
olekzabl: > It must be short enough (...) but long enough (...)

How strict are these "must"s?

If taking literally, this would feel alarming. (I blindly suspect various K8s latencies like this are quite variable. So if it was indeed a "must", to be longer than "variable around sth" but shorter than "variable around sth else"... I'd hesitate to LGTM such PR and rather start thinking about safer alternatives).

However, in this case, I think actually both "bounds" are quite soft, right?
Will Kueue seriously misbehave if this duration is a bit too short or too long?

* If too long: users will wait for the "correct" status of the local Job _a bit too long_ - but still shorter than they'd need to wait without this fix. So still an improvement of the current state. (And the current state isn't too bad either; see the discussion in #11730).

* If too short: I understand we may hit this requeueing a few times, but it won't become a deadlocked loop. The loop will stop as soon as the local Job gets unsuspended.

Am I getting this right?

---

To be clear: I'm not challenging your solution. Nor your choice of 2 seconds.
I'm challenging only _the comment_ because I think it's _too drastic_ :)

(And **if** the comment isn't too drastic, then I'll be challenging the solution).

pkg/controller/jobframework/multikueue.go
```diff
@@ -27,10 +28,22 @@ import (
...
 
+// ErrSyncDeferred is returned by MultiKueueAdapter.SyncJob when the adapter
+// intentionally skipped propagating the remote Job's status to the local Job
+// (typically because the local Job is still suspended and a status patch
+// would violate K8s suspend-validation rules). The caller should treat this
+// as a transient condition and requeue the workload on a short timer so the
+// sync can be retried once the local Job is unsuspended; without that
+// requeue the next reconcile may be up to `workerLostTimeout` (15 min) away
+// and the local Job's status.Active will not catch up — see #11115.
+var ErrSyncDeferred = errors.New("multikueue sync deferred until local job is unsuspended")
```
olekzabl: I'm hesitating about passing this via an `error`.
This feels a bit "abusing" (what we encoutered is not really an error).
I guess you also felt it a bit misleading, judging by the number of longer code comments added.

How about making `SyncJob` return a pair `reconcile.Result, err`? Or maybe `*Duration, err`, like [here](https://github.com/kubernetes-sigs/kueue/blob/a67bac84be2d53bf654c849681c3437290cbcdeb/pkg/controller/admissionchecks/multikueue/multikueuecluster.go#L733-L740)?
Then, with appropriate variable naming on the caller side, maybe the code would become more self-explanatory.
(I understand this might involve touching additional lines for non-Job adapters, though in an easy way).

---

Side note: for my intuition, it'd be nicest to have an auxiliary "combiner" for two `reconcile.Result`s, taking the "more aggressive" requeue spec of either side. (I.e. the shorter `RequeueAfter`, though ignoring `0`). Though I couldn't find any existing use case for that, so maybe it's a poor idea. Just thinking loud.

pkg/controller/jobframework/multikueue.go
```diff
@@ -27,10 +28,22 @@ import (
...
 )
 
+// ErrSyncDeferred is returned by MultiKueueAdapter.SyncJob when the adapter
+// intentionally skipped propagating the remote Job's status to the local Job
+// (typically because the local Job is still suspended and a status patch
+// would violate K8s suspend-validation rules). The caller should treat this
+// as a transient condition and requeue the workload on a short timer so the
+// sync can be retried once the local Job is unsuspended; without that
+// requeue the next reconcile may be up to `workerLostTimeout` (15 min) away
+// and the local Job's status.Active will not catch up — see #11115.
```
olekzabl: Nit: #11115, as I'm looking into it, does not contain much written context directly explaining this PR.

Instead, you can consider referring to [this discussion](https://github.com/kubernetes-sigs/kueue/pull/11730#issuecomment-4566063844), which has more context which I think is relevant.

But I'm leaving this to your judgement.

pkg/controller/jobs/job/job_multikueue_adapter.go
```diff
@@ -102,7 +112,15 @@ func (b *multiKueueAdapter) SyncJob(ctx context.Context, localClient client.Clie
...
-func determineStatusUpdate(ctx context.Context, log logr.Logger, localJob, remoteJob *batchv1.Job) *batchv1.JobStatus {
+// determineStatusUpdate returns the JobStatus that should be written to the
+// local Job, plus a "deferred" flag. deferred is true only for the "local
+// suspended, remote unsuspended, not finished" branch — the one case where
+// the real remote status.Active is intentionally NOT propagated to the local
+// Job. Callers use the flag to schedule a short re-reconcile so the sync can
+// be retried once the local Job is unsuspended (see #11115). Keeping both
+// values in this function ensures the deferred condition can't drift out of
+// sync with the branches it describes.
+func determineStatusUpdate(ctx context.Context, log logr.Logger, localJob, remoteJob *batchv1.Job) (*batchv1.JobStatus, bool) {
```
olekzabl: Nit: consider naming the result variables, so that the meaning of `bool` is clearer.

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -493,16 +502,32 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
 			return reconcile.Result{}, err
 		}
 
+		syncDeferred := false
 		if err := group.jobAdapter.SyncJob(ctx, w.client, remoteCl, group.controllerKey, group.local.Name, w.origin); err != nil {
```
olekzabl: This is one of 3 calls to `SyncJob` in this file. How about the other 2 calls?

* Can we hit the "deferred" case in them?
  * In [this case](https://github.com/kubernetes-sigs/kueue/blob/fd81a84b0080869090c61b4212ebbe221c88cc36/pkg/controller/admissionchecks/multikueue/workload.go#L363), I guess we can't, because "deferred" requires that the remote is not `Finished`.
  * In [this case](https://github.com/kubernetes-sigs/kueue/blob/fd81a84b0080869090c61b4212ebbe221c88cc36/pkg/controller/admissionchecks/multikueue/workload.go#L387), I guess we can?

* What's the outcome of not handling the "deferred" (encoded as an error)?
  I guess not too bad; we'll fail the reconcile and reconcile again after some time.
  Still, in the logs, it may produce some distracting false-alarm messages.
  So I think it'd be nicer to handle "deferred" also there.

WDYT?

olekzabl (commented):
Thank you, I like it! 
In fact, it's a change [I hoped to see](https://github.com/kubernetes-sigs/kueue/pull/11730#issuecomment-4566063844) one day :)

Leaving a few minor comments.

---

I'd also have one general remark, about the code comments:

* On one hand, I'm grateful for your effort to explain this logic comprehensively.
  Kudos for that, as I think Kueue code is often missing such explanations :)

* OTOH, I'm feeling such a rich comment would be great to keep in 1 place (or maybe 2).
  While in this PR it's sort-of-repeated 7 times, and this feels too much.
  (I mean, one PR like this doesn't hurt yet - but I'm anxious about some visual clutter which would build up if we encouraged such a trend generally).

To tackle this, I have two ideas:

* Keep a single "mother comment", e.g. next to `syncDeferredRequeueAfter` (or in [#11115](https://github.com/kubernetes-sigs/kueue/issues/11115) ?) and in all other places just say "see the comment on syncDeferredRequeueAfter".

* Use a more consistent language across the layers (cf. [this comment](https://github.com/kubernetes-sigs/kueue/pull/11867/changes#r3374812809))

WDYT?

pkg/controller/admissionchecks/multikueue/workload_test.go
```diff
@@ -2006,3 +2007,111 @@ func TestConfigHandlerDelete(t *testing.T) {
...
+
+	gotResult, gotErr := reconciler.reconcileGroup(ctx, group)
+	if gotErr != nil {
+		t.Fatalf("reconcileGroup returned unexpected error: %v", gotErr)
+	}
+
+	// The bug under test is that the requeue would otherwise be workerLostTimeout
+	// (15 minutes). A short requeue (≤ 30 s) is the contract we want.
+	const maxAcceptableRequeue = 30 * time.Second
+	if gotResult.RequeueAfter <= 0 || gotResult.RequeueAfter > maxAcceptableRequeue {
```
olekzabl: Nit: why not expect exactly `syncDeferredRequeueAfter`?

pkg/controller/admissionchecks/multikueue/workload_test.go
```diff
@@ -2006,3 +2007,111 @@ func TestConfigHandlerDelete(t *testing.T) {
 		t.Errorf("expected workload wl1 to be queued, got %s", mockQ.Items[0].Name)
 	}
 }
+
+// deferredSyncStubAdapter is a minimal MultiKueueAdapter used by
+// TestReconcileGroup_SyncDeferred_ShortRequeue. SyncJob returns whatever the
+// configured syncErr is; the other methods are no-ops or trivial.
```
olekzabl: Nit: I'd remove this comment; I think this matter is quite simple and in such cases we tend to leave the code speak for itself.
```suggestion
```

pkg/controller/jobs/job/job_multikueue_adapter_test.go
```diff
@@ -644,3 +650,44 @@ func Test_multiKueueAdapter_SyncJob(t *testing.T) {
...
+
+	// Local Job: suspended (manager admission check still being processed).
+	localJob := newJob().Obj() // Suspend defaults to true on JobWrapper
+	// Remote Job: unsuspended, NOT finished, with Status.Active=1 (the value
+	// that the failing test in issue #11115 wanted propagated to the manager
+	// Job's status.Active and never got).
+	remoteJob := newJob().
+		Suspend(false).
+		Active(1).
+		Obj()
```
olekzabl: Nit: I'd also remove these comments.
```suggestion
	localJob := newJob().Obj() // Suspend defaults to true on JobWrapper
	remoteJob := newJob().
		Suspend(false).
		Active(1).
		Obj()
```

mimowo:
@andrewseif @olekzabl let's see if we can finalize the bugfix for the next patch. It is also causing flakes https://github.com/kubernetes-sigs/kueue/issues/11115. If we have doubts about the fix we should probably plumb the tests for now to make the flakes not impacting other branches.

mimowo:
/test pull-kueue-test-e2e-extended-shard-1-main-1-35
/test pull-kueue-test-e2e-extended-shard-1-main-1-33 
/test pull-kueue-test-e2e-extended-shard-0-main-1-35
These are unrelated failures, retrying

andrewseif:
ill try to go through all comments and the discussion in #11730, and then see where we go, hopefully will get it done today or tomorrow at the maximum

pkg/controller/admissionchecks/multikueue/workload.go
andrewseif: To answer your question I believe the limits are between 0 -5 minutes, so anything below 5 minutes is an upgrade to the user.

why 2? 2s is a soft preference comfortably above the sub-second unsuspend latency, comfortably below anything a user would notice, and bounded below the 5-min fallback either way, I feel like a too-short value (below 2) just burns a few extra workqueue iterations, without gaining value.

However, I changed the wording abit to reflect that.

pkg/controller/jobframework/multikueue.go
andrewseif: I agree with you, I wasn't sure about the sentinel error approach myself, but the other option I thought of was overkill, but now that you are questioning this approach, I know it's bad. 

This is in-band signaling, one channel carrying two kinds of message is a smell and I do agree we should change it, however this "duration" approach would like to challenge it:

Putting a timing decision inside the adapter, while the adapter (pkg/controller/jobs/job/) is a low-level translator "given local + remote Job, copy status correctly." It has no business deciding how often the multikueue reconciler should wake up. That cadence (the syncDeferredRequeueAfter = 2s constant) is a policy owned by the reconciler (pkg/controller/admissionchecks/multikueue/workload.go). (feel free to correct me if I am wrong here, or anywhere for that matter)

If the adapter returns reconcile.Result{RequeueAfter: 2*time.Second} or *Duration(2s), then:

- the "2 seconds" magic number leaks into the job package, where it doesn't belong.
- every adapter now has to produce a reconcile timing, even though only the Job adapter ever defers;
- to later tune the requeue cadence, you'd edit a jobs/job/ adapter instead of the reconciler, which is a layering inversion(?)

My recommendation a minimal, correctly-layered signal is a boolean:
`SyncJob(...) (deferred bool, err error)`

This way the adapter reports a fact about what it did ("I deferred the status sync"). 

The reconciler decides what to do about that fact (requeue after its own 2s constant). Facts in the adapter, policy in the reconciler. This also kills "it's not an error" objection (which I agree with): `deferred` is a plain bool, `err` is reserved for real failures, and the `errors.Is` dance disappears at all three call sites.

I skipped the Result-combiner idea since with a bool there's nothing to merge.

I could also implement it so you can see the change incase this feels too theoretical. 

WDYT?

pkg/controller/admissionchecks/multikueue/workload.go
andrewseif: I agree we should handle it, if we go with the `(bool, error)` signature, each site becomes a clean `deferred` check instead of an `errors.Is` dance repeated twice.

pkg/controller/admissionchecks/multikueue/workload_test.go
andrewseif: Done, I think exact assertion is better, didn't cross my mind 😄

pkg/controller/jobs/job/job_multikueue_adapter.go
andrewseif: done

pkg/controller/jobframework/multikueue.go
andrewseif: since we have few discussions here, I would augment this comment and add the discussion url aswell.

andrewseif:
@olekzabl 

> Keep a single "mother comment", e.g. next to syncDeferredRequeueAfter (or in https://github.com/kubernetes-sigs/kueue/issues/11115 ?) and in all other places just say "see the comment on syncDeferredRequeueAfter".

I've consolidated it down to two canonical homes, with everything else reduced to one-line pointers:

syncDeferredRequeueAfter → the canonical explanation within the multikueue package (the race, why a short requeue is needed, and the soft-bounds reasoning). The eviction and main call sites now just say "…See syncDeferredRequeueAfter."

MultiKueueAdapter.SyncJob interface doc  → the contract for the deferred return value, carrying the #11115 + discussion links.

The one place I deviated from a strict single "mother comment": I didn't point the jobframework interface doc or the job adapter at syncDeferredRequeueAfter, because that's an unexported symbol in a downstream package (multikueue imports jobframework, not the reverse), referencing it from there would be a backwards, hard-to-navigate cross-package pointer. So within job, the comments point at MultiKueueAdapter.SyncJob / determineStatusUpdate instead, and the determineStatusUpdate godoc keeps only its own local logic (which branch → deferred) plus the anti-drift note.

Net effect: ~5 full restatements → 2 detailed homes + short pointers. Let me know if you'd prefer the canonical text to live somewhere else (e.g. in [#11115](https://github.com/kubernetes-sigs/kueue/issues/11115) directly) and I'll adjust

pkg/controller/jobframework/multikueue.go
andrewseif: I took the liberty of doing the change, if its not acceptable, I can always revert and do it with the duration and combiner 😃

pkg/controller/admissionchecks/multikueue/externalframeworks/adapter.go
```diff
@@ -57,30 +57,30 @@ func NewAdapter(gvk schema.GroupVersionKind) jobframework.MultiKueueAdapter {
 // SyncJob synchronizes a job resource between the local and remote clusters.
 // It ensures that the remote cluster has a corresponding object for the local job,
 // creating it if necessary, and updates the status of the local object based on the remote object.
-func (a *Adapter) SyncJob(ctx context.Context, localClient client.Client, remoteClient client.Client, key types.NamespacedName, workloadName, origin string) error {
+func (a *Adapter) SyncJob(ctx context.Context, localClient client.Client, remoteClient client.Client, key types.NamespacedName, workloadName, origin string) (bool, error) {
```
olekzabl: [nit, feel free to ignore]

You could name the output vars also here.

olekzabl:
Thank you!

/lgtm

Leaving one comment but it's optional

olekzabl (approved):

olekzabl:
> I've consolidated it down to two canonical homes, with everything else reduced to one-line pointers:

Actually I found 3 homes :) (`SyncJob`, `determineStatusUpdate`, and the new duration).

But I like it now. LGTM. Thank you for the rewording.

pkg/controller/admissionchecks/multikueue/externalframeworks/adapter.go
andrewseif: ill do it, makes the code more readable and make more sense

andrewseif: done

mimowo:
@andrewseif I think you need to rebase the PR and fix the compilation issues which is likely causing the liner errors

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -493,16 +521,26 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
...
+		syncDeferred, err := group.jobAdapter.SyncJob(ctx, w.client, remoteCl, group.controllerKey, group.local.Name, w.origin)
+		if err != nil {
 			log.Error(err, "Syncing remote controller object")
 			// We'll retry this in the next reconciling.
 			return reconcile.Result{}, err
 		}
+		if syncDeferred {
+			// The local Job is still suspended; requeue short so the next SyncJob
+			// can propagate the remote status. See syncDeferredRequeueAfter.
+			log.V(3).Info("Sync deferred until local job is unsuspended; requeuing on short timer")
```
mimowo: I would rather move the code below around L541, because this is not necessarily true if `syncReservingRemoteState` ends with error

andrewseif:
@mimowo yah I noticed I need a rebase, thank you!

mimowo:
As a production code bugfix it warrants a release note, proposal:

/release-note-edit
```release-note
MultiKueue: Fixed a bug that could leave stale status for Kubernetes Jobs in the manager
cluster when the worker-cluster Job reached steady state quickly and stopped getting
updates while the manager-cluster Job was still suspended.
```

mimowo:
Thank you for nailing down this remaining flake. I think our build is rock solid now ;) 
/approve
lgtm, I would just wait just for the nit comment.

i will try to use the robot to CP but I expect conflicts so please prepare the CPs manually in case the robot fails

/cherrypick release-0.18
/cherrypick release-0.17

pkg/controller/admissionchecks/multikueue/workload.go
andrewseif: done

andrewseif:
@mimowo I think the PR is ready now

mimowo:
Please squash the commits to make CPs easier

mimowo:
@andrewseif please prepare the CPs manually

PR #11872, opened 2026-06-01
Add preemptionGates support to ConcurrentAdmission

### LINKED ISSUE #10920: [ConcurrentAdmission] Only one Variant per Parent should be able to issue evictions
Currently when using the ConcurrentAdmission feature, each Variant Workload is treated independently and can issue evictions as any other Workload. We'd like to limit that capability to prevent excessive disruptions on all flavors in a ClusterQueue. 

Only one Variant per Parent should be able to issue evictions. If there's a Sibling Workload that has already issued eviction, the consecutive ones shouldn't be able to do so.   

**Why is this needed**:

**Completion requirements**:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

reruno:
#### What type of PR is this?
/kind feature

Fixes #10920 

```release-note
ConcurrentAdmission: make sure there is at most one preemption variant issuing preemptions at any given time.
This is achieved using the "preemption gates" mechanism.
```

## Summary by CodeRabbit

* **New Features**
  * Variants now include a concurrent-admission preemption gate; existing workloads missing the gate are backfilled during reconciliation.
  * Preemption gate “ungating” is serialized so only one variant opens at a time; the controller schedules retries based on ungating timing and emits an event when a gate opens.
* **Bug Fixes**
  * Scheduler now consistently treats closed concurrent-admission gates as preemption-gated when concurrent admission is enabled.
* **Documentation**
  * Updated implementation history with new dated milestones and behavior notes.
* **Tests**
  * Expanded unit, integration, and e2e coverage for gate stamping, selection/ungating semantics, and admission ordering under preemption.

### DIFF
--- a/keps/8691-concurrent-admission/README.md
+++ b/keps/8691-concurrent-admission/README.md
@@ -747,6 +747,11 @@ Major milestones might include:
 - when the KEP was retired or superseded
 -->
 
+- 2026-04-20: Initial KEP merged.
+- 2026-05-19: API field `minPreferredFlavorName` renamed to `lastAcceptableFlavorName`.
+- 2026-05-22: `RetainFirstAdmission` migration mode added to pin a Workload to its first admitted flavor.
+- 2026-06-01: Serialized preemptions across sibling Variants implemented by reusing the `preemptionGates` mechanism from [KEP 8303](/keps/8303-multikueue-orchestrated-preemption/README.md). Each Variant is created with a closed `kueue.x-k8s.io/concurrent-admission` gate, which the Variant Controller opens for at most one Variant at a time.
+
 ## Drawbacks
 
 ### Increased API Object Count
--- a/pkg/controller/concurrentadmission/controller.go
+++ b/pkg/controller/concurrentadmission/controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"slices"
+	"time"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
@@ -51,10 +52,13 @@ import (
 )
 
 const (
-	ConcurrentAdmissionController = "concurrent-admission-controller"
-	ReasonCreatedVariant          = "CreatedVariant"
-	ReasonActivatedVariant        = "ActivatedVariant"
-	ReasonDeactivatedVariant      = "DeactivatedVariant"
+	ConcurrentAdmissionController  = "concurrent-admission-controller"
+	ReasonCreatedVariant           = "CreatedVariant"
+	ReasonActivatedVariant         = "ActivatedVariant"
+	ReasonDeactivatedVariant       = "DeactivatedVariant"
+	ReasonPreemptionUngatedVariant = "PreemptionUngatedVariant"
+
+	preemptionTimeout = 5 * time.Minute
 )
 
 type variantReconciler struct {
@@ -177,7 +181,14 @@ func (r *variantReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		log.Error(err, "Failed to sync admission status")
 		return ctrl.Result{}, err
 	}
-	return ctrl.Result{}, nil
+
+	candidate, timeToWait := r.selectVariantToOpenPreemptionGate(ctx, variants)
+	if candidate != nil {
+		if err := r.openPreemptionGate(ctx, candidate); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+	return ctrl.Result{RequeueAfter: timeToWait}, nil
 }
 
 func (r *variantReconciler) getFamilyAndClusterQueue(ctx context.Context, req ctrl.Request) (*kueue.Workload, []kueue.Workload, *kueue.ClusterQueue, error) {
@@ -267,6 +278,8 @@ func generateVariant(parent *kueue.Workload, flavor kueue.ResourceFlavorReferenc
 		Spec:   parent.Spec,
 		Status: parent.Status,
 	}
+	variant.Spec.PreemptionGates = slices.Clone(variant.Spec.PreemptionGates)
+	workload.EnsurePreemptionGateOnSpec(variant, constants.ConcurrentAdmissionPreemptionGate)
 	delete(variant.Labels, constants.ConcurrentAdmissionParentLabelKey)
 	metav1.SetMetaDataAnnotation(&variant.ObjectMeta, constants.WorkloadAllowedResourceFlavorAnnotation, string(flavor))
 	return variant
@@ -629,6 +642,87 @@ func (r *variantReconciler) shouldReconcile(wl *kueue.Workload) bool {
 	}
 	log.V(3).Info("Workload is neither a parent variant nor a variant, ignoring", "workload", klog.KObj(wl))
 	return false
+}
+
+func (r *variantReconciler) selectVariantToOpenPreemptionGate(ctx context.Context, variants []kueue.Workload) (*kueue.Workload, time.Duration) {
+	log := ctrl.LoggerFrom(ctx).WithValues("op", "selectVariantToOpenPreemptionGate")
+
+	admissible := admissibleVariants(variants)
+	candidate := firstCandidateVariant(log, admissible)
+	if candidate == nil {
+		log.V(4).Info("No candidate variant found for ungating preemptions")
+		return nil, 0
+	}
+
+	lastUngateTime := latestOpenGateTime(log, admissible)
+	if lastUngateTime == nil {
+		log.V(4).Info("Found variant to ungate, no gate is currently open", "variantToUngate", klog.KObj(candidate))
+		return candidate, preemptionTimeout
+	}
+	timeToWait := preemptionTimeout - r.clock.Since(lastUngateTime.Time)
+	if timeToWait > 0 {
+		log.V(4).Info("Single preemption timeout did not expire", "timeToWait", timeToWait, "nextVariantToUngate", klog.KObj(candidate))
+		return nil, timeToWait
+	}
+	log.V(4).Info("Found variant to ungate", "variantToUngate", klog.KObj(candidate))
+	return candidate, preemptionTimeout
+}
+
+func firstCandidateVariant(log logr.Logger, admissibleVariants []*kueue.Workload) *kueue.Workload {
+	for _, wl := range admissibleVariants {
+		if workload.HasOpenPreemptionGate(wl, constants.ConcurrentAdmissionPreemptionGate) {
+			continue
+		}
+		wlLog := log.WithValues("candidateVariant", klog.KObj(wl), "flavor", concurrentadmission.GetVariantFlavor(wl))
+		if workload.BlockedOnPreemptionGatesCondition(wl) == nil {
+			wlLog.V(4).Info("Variant does not require preemption")
+			continue
+		}
+		wlLog.V(4).Info("Variant is the candidate for ungating")
+		return wl
+	}
+	return nil
+}
+
+func latestOpenGateTime(log logr.Logger, admissibleVariants []*kueue.Workload) *metav1.Time {
+	var lastUngateTime *metav1.Time
+	for _, wl := range admissibleVariants {
+		openGate := workload.FindPreemptionGate(wl, constants.ConcurrentAdmissionPreemptionGate)
+		if openGate == nil || openGate.Position != kueue.PreemptionGatePositionOpen {
+			continue
+		}
+		if lastUngateTime == nil || openGate.LastTransitionTime.After(lastUngateTime.Time) {
+			log.V(4).Info("Variant has the latest preemption ungating time", "openPreemptionGateVariant", klog.KObj(wl), "flavor", concurrentadmission.GetVariantFlavor(wl))
+			lastUngateTime = &openGate.LastTransitionTime
+		}
+	}
+	return lastUngateTime
+}
+
+func admissibleVariants(variants []kueue.Workload) []*kueue.Workload {
+	var ret []*kueue.Workload
+	for i := range variants {
+		if workload.IsAdmissible(&variants[i]) {
+			ret = append(ret, &variants[i])
+		}
+	}
+	return ret
+}
+
+func (r *variantReconciler) openPreemptionGate(ctx context.Context, variant *kueue.Workload) error {
+	log := ctrl.LoggerFrom(ctx)
+	log.V(2).Info("Opening preemption gate for variant", "variant", klog.KObj(variant), "flavor", concurrentadmission.GetVariantFlavor(variant))
+	var opened bool
+	if err := workload.PatchAdmissionStatus(ctx, r.client, variant, r.clock, func(wl *kueue.Workload) (bool, error) {
+		opened = workload.OpenPreemptionGate(wl, constants.ConcurrentAdmissionPreemptionGate, metav1.NewTime(r.clock.Now()))
+		return opened, nil
+	}, workload.WithRetryOnConflictForPatch()); err != nil {
+		return fmt.Errorf("opening preemption gate on variant %s: %w", klog.KObj(variant), err)
+	}
+	if opened {
+		r.recorder.Eventf(variant, nil, corev1.EventTypeNormal, ReasonPreemptionUngatedVariant, ReasonPreemptionUngatedVariant, "Opened preemption gate for variant workload %q", klog.KObj(variant))
+	}
+	return nil
 }
 
 func getAdmittedVariant(variants []kueue.Workload) *kueue.Workload {
--- a/test/integration/singlecluster/controller/concurrentadmission/concurrent_admission_test.go
+++ b/test/integration/singlecluster/controller/concurrentadmission/concurrent_admission_test.go
@@ -17,6 +17,8 @@ limitations under the License.
 package concurrentadmission
 
 import (
+	"fmt"
+
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
@@ -497,6 +499,100 @@ var _ = ginkgo.Describe("Concurrent Admission", func() {
 					g.Expect(parentWl.Status.AdmissionChecks).To(gomega.BeEmpty())
 					g.Expect(parentWl.Status.Admission.PodSetAssignments[0].Flavors[corev1.ResourceCPU]).To(gomega.Equal(kueue.ResourceFlavorReference("provreq-flavor")))
 				}, util.Timeout, util.Interval).Should(gomega.Succeed())
+			})
+		})
+	})
+
+	ginkgo.When("Variants require preemption to be admitted", func() {
+		const (
+			lowPriority  int32 = 0
+			highPriority int32 = 100
+		)
+
+		var cq *kueue.ClusterQueue
+		var lq *kueue.LocalQueue
+		var flavorReservation *kueue.ResourceFlavor
+		var flavorSpot *kueue.ResourceFlavor
+
+		ginkgo.BeforeEach(func() {
+			flavorReservation = utiltestingapi.MakeResourceFlavor("reservation").Obj()
+			util.MustCreate(ctx, k8sClient, flavorReservation)
+
+			flavorSpot = utiltestingapi.MakeResourceFlavor("spot").Obj()
+			util.MustCreate(ctx, k8sClient, flavorSpot)
+
+			// reservation is the most-preferred flavor (index 0). Each flavor holds
+			// exactly 1 CPU, so a single low-priority workload fully occupies it and
+			// any higher-priority variant must preempt to be admitted.
+			cq = utiltestingapi.MakeClusterQueue("cq-preemption-gate").
+				ConcurrentAdmissionPolicy(kueue.ConcurrentAdmissionTryPreferredFlavors).
+				ResourceGroup(
+					*utiltestingapi.MakeFlavorQuotas(flavorReservation.Name).Resource(corev1.ResourceCPU, "1").Obj(),
+					*utiltestingapi.MakeFlavorQuotas(flavorSpot.Name).Resource(corev1.ResourceCPU, "1").Obj(),
+				).
+				Preemption(kueue.ClusterQueuePreemption{
+					WithinClusterQueue: kueue.PreemptionPolicyLowerPriority,
+				}).Obj()
+			util.CreateClusterQueuesAndWaitForActive(ctx, k8sClient, cq)
+
+			lq = utiltestingapi.MakeLocalQueue("lq", ns.Name).ClusterQueue(cq.Name).Obj()
+			util.CreateLocalQueuesAndWaitForActive(ctx, k8sClient, lq)
+		})
+
+		ginkgo.AfterEach(func() {
+			gomega.Expect(util.DeleteWorkloadsInNamespace(ctx, k8sClient, ns)).To(gomega.Succeed())
+			util.ExpectObjectToBeDeleted(ctx, k8sClient, lq, true)
+			util.ExpectObjectToBeDeleted(ctx, k8sClient, cq, true)
+			util.ExpectObjectToBeDeleted(ctx, k8sClient, flavorSpot, true)
+			util.ExpectObjectToBeDeleted(ctx, k8sClient, flavorReservation, true)
+		})
+
+		ginkgo.It("should open the preemption gate for only the most-preferred variant at a time", func() {
+			ginkgo.By("Occupying both flavors with low-priority workloads", func() {
+				for i := range 2 {
+					lowWl := utiltestingapi.MakeWorkload(fmt.Sprintf("low-%d", i), ns.Name).
+						Queue(kueue.LocalQueueName(lq.Name)).
+						Priority(lowPriority).
+						Request(corev1.ResourceCPU, "1").
+						Obj()
+					util.MustCreate(ctx, k8sClient, lowWl)
+					util.ExpectWorkloadsToHaveQuotaReservation(ctx, k8sClient, cq.Name, lowWl)
+				}
+			})
+
+			parentWl := utiltestingapi.MakeWorkload("parent-wl-preemption-gate", ns.Name).
+				Queue(kueue.LocalQueueName(lq.Name)).
+				Priority(highPriority).
+				Request(corev1.ResourceCPU, "1").
+				ParentVariant().
+				Obj()
+
+			ginkgo.By("Creating the high-priority parent workload", func() {
+				util.MustCreate(ctx, k8sClient, parentWl)
+			})
+
+			ginkgo.By("Verifying the preemption gate is opened for the reservation variant", func() {
+				gomega.Eventually(func(g gomega.Gomega) {
+					list := &kueue.WorkloadList{}
+					g.Expect(k8sClient.List(ctx, list, client.InNamespace(ns.Name))).To(gomega.Succeed())
+
+					variantReservation := getVariantByFlavor(list, parentWl.Name, flavorReservation.Name)
+					g.Expect(variantReservation).ToNot(gomega.BeNil(), "Variant for reservation not found")
+					g.Expect(workload.HasOpenPreemptionGate(variantReservation, controllerconstants.ConcurrentAdmissionPreemptionGate)).
+						To(gomega.BeTrue(), "reservation variant should have an open preemption gate")
+				}, util.Timeout, util.Interval).Should(gomega.Succeed())
+			})
+
+			ginkgo.By("Verifying the spot variant gate stays closed while reservation holds the open gate", func() {
+				gomega.Consistently(func(g gomega.Gomega) {
+					list := &kueue.WorkloadList{}
+					g.Expect(k8sClient.List(ctx, list, client.InNamespace(ns.Name))).To(gomega.Succeed())
+
+					variantSpot := getVariantByFlavor(list, parentWl.Name, flavorSpot.Name)
+					g.Expect(variantSpot).ToNot(gomega.BeNil(), "Variant for spot not found")
+					g.Expect(workload.HasOpenPreemptionGate(variantSpot, controllerconstants.ConcurrentAdmissionPreemptionGate)).
+						To(gomega.BeFalse(), "spot variant gate must stay closed until the reservation variant's preemption timeout elapses")
+				}, util.ConsistentDuration, util.Interval).Should(gomega.Succeed())
 			})
 		})
 	})
--- a/pkg/controller/concurrentadmission/controller_test.go
+++ b/pkg/controller/concurrentadmission/controller_test.go
@@ -18,6 +18,7 @@ package concurrentadmission
 
 import (
 	"testing"
+	"time"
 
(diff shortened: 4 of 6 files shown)


keps/8691-concurrent-admission/README.md
```diff
@@ -557,7 +557,9 @@ While we only issue preemptions coming from one Workload per CQ, what happens is
 3. A sibling Variant is picked up by the scheduler and is preempting some other Workloads
 
 We want to disallow other Variants to preempt if one of the Variants has already preempted some Workloads.
-During the Alpha implementation phase we will evaluate if the mechanism is done based on in-memory tracking, or we use the preemptionGates mechanism [KEP 8303](/keps/8303-multikueue-orchestrated-preemption/README.md).
+This is implemented by reusing the preemptionGates mechanism [KEP 8303](/keps/8303-multikueue-orchestrated-preemption/README.md).
+Every Variant is created with a closed `kueue.x-k8s.io/concurrent-admission` preemption gate, so the scheduler does not let a Variant issue preemptions while its gate is closed.
+The Variant Controller opens the gate for at most one Variant at a time, starting from the most preferred flavor that requires preemption. Before opening the gate for the next Variant it waits for a fixed timeout (5 minutes), giving the ungated Variant a chance to complete its preemptions and get admitted.
```
kshalot: I feel that this should also be mentioned in the implementation history section.

pkg/controller/concurrentadmission/controller.go
```diff
@@ -177,7 +181,14 @@ func (r *variantReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 	}
-	return ctrl.Result{}, nil
+
+	candidate, timeToWait := r.variantToOpenPreemptionGate(ctx, variants)
+	if candidate != nil {
+		if err := r.openPreemptionGate(ctx, candidate); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+	return ctrl.Result{RequeueAfter: timeToWait}, nil
```
kshalot, line 191: I think that in the case where there is no candidate, we should not requeue the reconcile, no? Since a variant signaling the need to preempt will trigger a reconciliation anyway.

Right now we'll just have reconciliations spinning every 5 minutes regardless of whether a preemption will happen or not.

pkg/controller/concurrentadmission/controller.go
```diff
@@ -267,6 +279,7 @@ func generateVariant(parent *kueue.Workload, flavor kueue.ResourceFlavorReferenc
 		Spec:   parent.Spec,
 		Status: parent.Status,
 	}
+	variant.Spec.PreemptionGates = append(slices.Clone(variant.Spec.PreemptionGates), kueue.PreemptionGate{Name: constants.ConcurrentAdmissionPreemptionGate})
```
kshalot: I might be missing something, but do we need to `slices.Clone` the preemption gates?

pkg/controller/concurrentadmission/controller.go
```diff
@@ -631,6 +644,83 @@ func (r *variantReconciler) shouldReconcile(wl *kueue.Workload) bool {
...
+			wlRequiresPreemption := preemptionSignalCond != nil && preemptionSignalCond.Status == metav1.ConditionTrue
+			if !wlRequiresPreemption {
+				wlLog.V(4).Info("Variant does not require preemption")
+				continue
+			}
+			// variants are sorted by priority,
+			// we want to open preemption gate for high preference flavors first
+			if candidate == nil {
+				wlLog.V(4).Info("Variant is the candidate for ungating")
+				candidate = wl
```
kshalot: If the variants are sorted by priority, then I guess we could break here:

```suggestion
			// variants are sorted by priority,
			// we want to open preemption gate for high preference flavors first
			if candidate == nil {
				wlLog.V(4).Info("Variant is the candidate for ungating")
				candidate = wl
				break
```

kshalot (commented):
First, quick pass.

pkg/controller/concurrentadmission/controller.go
reruno: If there are no candidate timeToWait=0, and when requeueAfter: 0 then requeue will not happen, it is written in the docs for ctrl.Result

pkg/controller/concurrentadmission/controller.go
reruno: variant.Spec is shallow copy of parent, so variant.Spec.PreemptionGates is basically parent array of preemption gates, so we clone it to not change the parent slice

pkg/controller/concurrentadmission/controller.go
reruno: > RequeueAfter if greater than 0, tells the Controller to requeue the reconcile key after the Duration. Implies that Requeue is true, there is no need to set Requeue to true at the same time as RequeueAfter.

reruno:
```
	if candidate == nil {
		log.V(4).Info("No candidate variant found for ungating preemptions")
		return nil, 0
	} else if lastUngateTime == nil {
		log.V(4).Info("Found variant to ungate, no gate is currently open", "variantToUngate", klog.KObj(candidate))
		return candidate, preemptionTimeout
	}
	timeToWait := preemptionTimeout - r.clock.Since(lastUngateTime.Time)
	if timeToWait > 0 {
		log.V(4).Info("Single preemption timeout did not expire", "timeLeft", timeToWait, "nextVariantToUngate", klog.KObj(candidate))
		return nil, timeToWait
	}
	log.V(4).Info("Found variant to ungate", "variantToUngate", klog.KObj(candidate))
	return candidate, preemptionTimeout
```

If there is no candidate return timeToWait=0 (ie no requeue), 
if candidate was found and no variant was ungated before, return candidate + 5min requeue. 
If candidate found, there was ungate before check if 5 min elapsed since, if yes then return no candidate + timeToWait until 5 min elapse, if no then another candidate to ungate + 5min timeout 

let me know if this logic needs correction

pkg/controller/concurrentadmission/controller.go
reruno: I think if we break here it would be breaking too early, because in the code 
```
wlLog.V(4).Info("Variant has an open preemption gate")
			wlOpenedGate := wl.Status.PreemptionGates[wlOpenedGateIdx]
			if lastUngateTime == nil || wlOpenedGate.LastTransitionTime.After(lastUngateTime.Time) {
				wlLog.V(4).Info("Variant has the latest preemption ungating time")
				lastUngateTime = &wlOpenedGate.LastTransitionTime
			}
```

we are searching for the last ungate time in the variants, so to know that we go through each variant item

mimowo:
cc @kshalot who worked on the PreemptionGates mechanism

pkg/controller/concurrentadmission/controller.go
```diff
@@ -631,6 +649,116 @@ func (r *variantReconciler) shouldReconcile(wl *kueue.Workload) bool {
...
+// serialization. Without it, such Variants are never gated and can issue
+// preemptions in parallel, violating the single-preemptor-per-Parent guarantee.
+//
+// TODO: remove once ConcurrentAdmission graduates beyond alpha and no
+// pre-gate Variants can remain across a supported upgrade skew.
+func (r *variantReconciler) backfillPreemptionGates(ctx context.Context, variants []kueue.Workload) error {
+	log := ctrl.LoggerFrom(ctx)
+	for i := range variants {
+		v := &variants[i]
+		gates, added := ensureCAPreemptionGate(slices.Clone(v.Spec.PreemptionGates))
```
kshalot: When creating variants, the clone is indeed necessary for safety. But here, we are already dealing with independent memory, no? Since `variants` here come from `client.List` and is not a shallow clone of the parent.

test/e2e/singlecluster/baseline/concurrent_admission_test.go
```diff
@@ -228,6 +228,122 @@ var _ = ginkgo.Describe("ConcurrentAdmission", ginkgo.Label("area:singlecluster"
...
+			gomega.Expect(util.DeleteAllJobsInNamespace(ctx, k8sClient, ns)).Should(gomega.Succeed())
+			gomega.Expect(util.DeleteWorkloadsInNamespace(ctx, k8sClient, ns)).Should(gomega.Succeed())
+			util.ExpectObjectToBeDeleted(ctx, k8sClient, highWPC, true)
+			util.ExpectObjectToBeDeleted(ctx, k8sClient, lq, true)
+			util.ExpectObjectToBeDeleted(ctx, k8sClient, cq, true)
+			util.ExpectObjectToBeDeleted(ctx, k8sClient, onDemandRF, true)
+			util.ExpectObjectToBeDeleted(ctx, k8sClient, spotRF, true)
+		})
+
+		ginkgo.It("Should open the preemption gate for only the most-preferred Variant, preempting a single occupant", func() {
```
kshalot: A general concern - is this actually deterministic? In principle, wouldn't it be possible for the less preferred variant to run through a scheduling cycle first, trigger the preemption gate via `kueue.WorkloadBlockedOnPreemptionGates`, be ungated and only then the more preferred variant will get signal the need to preempt?

So it's a race that might result in the more preferred variant waiting to be ungated for 5 minutes. Maybe this is acceptable for alpha, but anyway it could cause flakes of this e2e. cc @mimowo

test/integration/singlecluster/controller/concurrentadmission/concurrent_admission_test.go
```diff
@@ -500,6 +503,98 @@ var _ = ginkgo.Describe("Concurrent Admission", func() {
 			})
 		})
 	})
+
+	ginkgo.When("Variants require preemption to be admitted", func() {
```
kshalot, line 506: One case I think is missing in the tests - whether we ungate the second variant after the timeout elapses. This should be easy to add to the integration test.

test/e2e/singlecluster/baseline/concurrent_admission_test.go
reruno: https://github.com/kubernetes-sigs/kueue/pull/11895

keps/8691-concurrent-admission/README.md

>The VariantsController leverages the existing implementation of scheduling logic in Kueue, where Workloads are sorted
by priorities and creation timestamps and then picked-up by the scheduler one per scheduling cycle (per ClusterQueue).
This means when creating Variants the controller should ensure more favorable Variants are ahead of less favorable ones in the queue.

mimowo:
/assign
I will take a look EOW, but another review pass from @kshalot or @mbobrovskyi and lgtm would certainly improve my confidence in the PR

kshalot:
I'm also planning to take another look, but from what I see some comments are not addressed yet (e.g. https://github.com/kubernetes-sigs/kueue/pull/11872#discussion_r3379780844), so I'll wait for the changes to be pushed

test/integration/singlecluster/controller/concurrentadmission/concurrent_admission_test.go
reruno: pkg/controller/concurrentadmission/controller_test.go
"ungates the next-preferred variant once the open gate's preemption timeout elapses":

Added new unit test case for `TestReconcile` that tests this case. 

Added to unit test because I think we don't use fake clocks in integration tests, and I don't want to manually subtracting LastTransitionTime in the object for integration tests

reruno:
/retest-required

If pull-kueue-test-e2e-multikueue-baseline-main pass I will create a flake test issue

Upd: https://github.com/kubernetes-sigs/kueue/issues/12286

pkg/controller/concurrentadmission/controller.go
```diff
@@ -631,6 +649,116 @@ func (r *variantReconciler) shouldReconcile(wl *kueue.Workload) bool {
...
+			if lastUngateTime == nil || wlOpenedGate.LastTransitionTime.After(lastUngateTime.Time) {
+				wlLog.V(4).Info("Variant has the latest preemption ungating time")
+				lastUngateTime = &wlOpenedGate.LastTransitionTime
+			}
+		}
+	}
+	if candidate == nil {
+		log.V(4).Info("No candidate variant found for ungating preemptions")
+		return nil, 0
+	} else if lastUngateTime == nil {
```
mykysha: Can we drop `else` as we return in the previous if statement?

pkg/controller/concurrentadmission/controller.go
```diff
@@ -631,6 +649,116 @@ func (r *variantReconciler) shouldReconcile(wl *kueue.Workload) bool {
...
+			if !wlRequiresPreemption {
+				wlLog.V(4).Info("Variant does not require preemption")
+				continue
+			}
+			// variants are sorted by priority,
+			// we want to open preemption gate for high preference flavors first
+			if candidate == nil {
+				wlLog.V(4).Info("Variant is the candidate for ungating")
+				candidate = wl
+			} else {
```
mykysha: I believe we can move the candidate check up in this if statement to skip the evaluation after the first candidate was found, but this way we will trade some observability in logs for exiting the checking process earlier

reruno: I double checked this piece of code and moving now I think moving candidate to `if wlOpenedGateIdx == -1 {` like this `if wlOpenedGateIdx == -1 && candidate == nil {` is not ideal, because we have `else {` block for `if wlOpenedGateIdx == -1 {` check, we can restructure the boolean checks like that

```
if wlOpenedGateIdx == -1 && candidate == nil{

else if wlOpenedGateIdx != -1 { 
```  

but I am not sure, if we should change the current code to that. 

I landed on this solution 
```
if wlOpenedGateIdx == -1 {
			wlLog.V(4).Info("Variant does not have an open preemption gate")
			// variants are sorted by priority,
			// if candidate is already exists then this candidate is higher priority than current wl
			if candidate != nil {
				wlLog.V(4).Info("A higher-preference variant is already the candidate for ungating")
				continue
			}
			preemptionSignalCond := apimeta.FindStatusCondition(wl.Status.Conditions, kueue.WorkloadBlockedOnPreemptionGates)
			wlRequiresPreemption := preemptionSignalCond != nil && preemptionSignalCond.Status == metav1.ConditionTrue
			if !wlRequiresPreemption {
				wlLog.V(4).Info("Variant does not require preemption")
				continue
			}
			wlLog.V(4).Info("Variant is the candidate for ungating")
			candidate = wl
		} else {
			wlLog.V(4).Info("Variant has an open preemption gate")
			wlOpenedGate := wl.Status.PreemptionGates[wlOpenedGateIdx]
			if lastUngateTime == nil || wlOpenedGate.LastTransitionTime.After(lastUngateTime.Time) {
				wlLog.V(4).Info("Variant has the latest preemption ungating time")
				lastUngateTime = &wlOpenedGate.LastTransitionTime
			}
		}
```

mimowo:
Let's make it more user/admin oriented, proposal:
/release-note-edit
```release-note
ConcurrentAdmission: make sure there is at most one preemption variant issuing preemptions at any given time.
This is achieved using the "preemption gates" mechanism.
```

pkg/controller/concurrentadmission/controller.go
```diff
@@ -631,6 +649,117 @@ func (r *variantReconciler) shouldReconcile(wl *kueue.Workload) bool {
...
+}
+
+// backfillPreemptionGates adds the ConcurrentAdmission preemption gate to Variants
+// that were created by a controller version predating gate-based preemption
+// serialization. Without it, such Variants are never gated and can issue
+// preemptions in parallel, violating the single-preemptor-per-Parent guarantee.
+//
+// TODO: remove once ConcurrentAdmission graduates beyond alpha and no
+// pre-gate Variants can remain across a supported upgrade skew.
+func (r *variantReconciler) backfillPreemptionGates(ctx context.Context, variants []kueue.Workload) error {
```
mimowo: Do we have this as an explicit requirement raised by a user? Generally we don't provide any compatibility for Alpha-level features. I think we could drop this code. Providing such a smooth migration could be considered a separete PR and feature if requested by users.

pkg/controller/concurrentadmission/controller.go
```diff
@@ -631,6 +649,117 @@ func (r *variantReconciler) shouldReconcile(wl *kueue.Workload) bool {
...
 
+func (r *variantReconciler) variantToOpenPreemptionGate(ctx context.Context, variants []kueue.Workload) (*kueue.Workload, time.Duration) {
+	log := ctrl.LoggerFrom(ctx).WithValues("op", "variantToOpenPreemptionGate")
+
+	var candidate *kueue.Workload
+	var lastUngateTime *metav1.Time
+	for i := range variants {
+		wl := &variants[i]
+		wlLog := log.WithValues("variant", klog.KObj(wl), "flavor", concurrentadmission.GetVariantFlavor(wl))
+		if !workload.IsActive(wl) || workload.IsFinished(wl) || workload.IsAdmitted(wl) {
```
mimowo: I think we could check IsAdmissible, which would have a better coverage actually.

For example, there is no point using the mechanism when QuotaReservation=True, but Admitted=False, because when we have QuotaReservation then preemptions are not needed anyway.

pkg/controller/concurrentadmission/controller.go
```diff
@@ -631,6 +649,117 @@ func (r *variantReconciler) shouldReconcile(wl *kueue.Workload) bool {
...
+				continue
+			}
+			preemptionSignalCond := apimeta.FindStatusCondition(wl.Status.Conditions, kueue.WorkloadBlockedOnPreemptionGates)
+			wlRequiresPreemption := preemptionSignalCond != nil && preemptionSignalCond.Status == metav1.ConditionTrue
+			if !wlRequiresPreemption {
+				wlLog.V(4).Info("Variant does not require preemption")
+				continue
+			}
+			wlLog.V(4).Info("Variant is the candidate for ungating")
+			candidate = wl
```
mimowo: Why not just break the loop at the moment we find the candidate, it seems all the downstream looping is no-op at this point, see: https://github.com/kubernetes-sigs/kueue/pull/11872/changes#diff-6fc941f3717c5f64124852f07071a6b02e3e48072b879069fc9bbb4a8d850ddbR671-R674

test/integration/singlecluster/controller/concurrentadmission/concurrent_admission_test.go
```diff
@@ -500,6 +503,98 @@ var _ = ginkgo.Describe("Concurrent Admission", func() {
...
+
+			ginkgo.By("Verifying the spot variant gate stays closed while reservation holds the open gate", func() {
+				gomega.Consistently(func(g gomega.Gomega) {
+					list := &kueue.WorkloadList{}
+					g.Expect(k8sClient.List(ctx, list, client.InNamespace(ns.Name))).To(gomega.Succeed())
+
+					variantSpot := getVariantByFlavor(list, parentWl.Name, flavorSpot.Name)
+					g.Expect(variantSpot).ToNot(gomega.BeNil(), "Variant for spot not found")
+					g.Expect(hasOpenConcurrentAdmissionGate(variantSpot)).To(gomega.BeFalse(), "spot variant gate must stay closed until the reservation variant's preemption timeout elapses")
+				}, util.ConsistentDuration, util.Interval).Should(gomega.Succeed())
```
mimowo, line 595: Could we use the Short consistent duration and that would be also enough to fail if there was bug?

pkg/controller/concurrentadmission/controller.go
reruno:
```
		} else {
			wlLog.V(4).Info("Variant has an open preemption gate")
			wlOpenedGate := wl.Status.PreemptionGates[wlOpenedGateIdx]
			if lastUngateTime == nil || wlOpenedGate.LastTransitionTime.After(lastUngateTime.Time) {
				wlLog.V(4).Info("Variant has the latest preemption ungating time")
				lastUngateTime = &wlOpenedGate.LastTransitionTime
			}
		}
```

We don't break from for loop after finding candidate, because we go through all variants to find the latest ungate time

mimowo: hmmm that is tricky, could we do this in two passes then? it would not change the asymptotic complexity but will help to make the code more readble I think

test/integration/singlecluster/controller/concurrentadmission/concurrent_admission_test.go
reruno: Yes we can use Short consistent duration (100ms), but it would be safer to check for Consistent duration (1s) if no changes were made on spot variant gate

mimowo: Hmmm, I think we don't need 1s duration really. wdyt about suing 250ms, this seems like a better default choice. So feel free to keep 1s now, but I will open an issue to reduce the default.

pkg/controller/concurrentadmission/controller.go
```diff
@@ -631,6 +653,87 @@ func (r *variantReconciler) shouldReconcile(wl *kueue.Workload) bool {
...
+	// gate, so we can serialize ungating to one variant per preemptionTimeout.
+	var lastUngateTime *metav1.Time
+	for i := range variants {
+		wl := &variants[i]
+		openGateIdx := slices.IndexFunc(wl.Status.PreemptionGates, func(gate kueue.PreemptionGateState) bool {
+			return gate.Name == constants.ConcurrentAdmissionPreemptionGate && gate.Position == kueue.PreemptionGatePositionOpen
+		})
+		if !workload.IsAdmissible(wl) || openGateIdx == -1 {
+			continue
+		}
```
mimowo: Can we de-duplicate the two passes, we could create a list of workload we iterate over, 
`candidateWorkloads` which would return the filtered list of workloads

reruno: At start I used pkg/util/slices/slices.go Pick() func that is basically filter, so code looked like that 
```
candidates := utilslice.Pick(variants, candidateCheck)
candidate := candidates[0]
```
But we need only first candidate, we don't need all candidates, that is why I go through variants without helpers. I think `candidateWorkloads` or  `utilslice.Pick(variants, candidateCheck)` is fine too

mimowo: If we only need the first candidate then let's add a helper `firstCandidate`. It is then encapsulated how this is achieved. basially forllowing the Single Layer of Abstraction principle.

reruno: Created two helper func `latestOpenGateTime` and `firstCandidateVariant`

pkg/controller/concurrentadmission/controller.go
mimowo: This looks too similar with the other place, I would like to have 
```golang
admissibleVariants := // some function or one liner to determine that
for _, wl := range admissibleVariants {
  _, found := findOpenPreeptionGate(wl)
  if !found {
    continue
  }
  // logic
}
```
Then in the other place:
```golang
admissibleVariants := // some function or one liner to determine that
for _, wl := range admissibleVariants {
  openGate, found := findOpenPreeptionGate(wl)
  if found {
    continue
  }
  // logic
}
```

pkg/controller/concurrentadmission/controller.go
```diff
@@ -177,7 +181,14 @@ func (r *variantReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		log.Error(err, "Failed to sync admission status")
 		return ctrl.Result{}, err
 	}
-	return ctrl.Result{}, nil
+
+	candidate, timeToWait := r.variantToOpenPreemptionGate(ctx, variants)
```
mimowo:
```suggestion
	candidate, timeToWait := r.selectVariantToOpenPreemptionGate(ctx, variants)
```

pkg/controller/concurrentadmission/controller.go
```diff
@@ -631,6 +653,103 @@ func (r *variantReconciler) shouldReconcile(wl *kueue.Workload) bool {
...
+	}
+
+	lastUngateTime := latestOpenGateTime(log, variants)
+	if lastUngateTime == nil {
+		log.V(4).Info("Found variant to ungate, no gate is currently open", "variantToUngate", klog.KObj(candidate))
+		return candidate, preemptionTimeout
+	}
+	timeToWait := preemptionTimeout - r.clock.Since(lastUngateTime.Time)
+	if timeToWait > 0 {
+		log.V(4).Info("Single preemption timeout did not expire", "timeLeft", timeToWait, "nextVariantToUngate", klog.KObj(candidate))
```
mimowo:
```suggestion
		log.V(4).Info("Single preemption timeout did not expire", "timeToWait", timeToWait, "nextVariantToUngate", klog.KObj(candidate))
```
nit, use consistent vocab as much as possible

reruno:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/11872/pull-kueue-verify-main/2067271243471523840

It was a flake, created issue https://github.com/kubernetes-sigs/kueue/issues/12323

pkg/controller/concurrentadmission/controller_preemption_gate_test.go
```diff
@@ -0,0 +1,354 @@
...
+
+func caGateState(position kueue.PreemptionGatePosition, t time.Time) kueue.PreemptionGateState {
+	return kueue.PreemptionGateState{
+		Name:               constants.ConcurrentAdmissionPreemptionGate,
+		Position:           position,
+		LastTransitionTime: metav1.NewTime(t),
+	}
+}
+
+func TestVariantToOpenPreemptionGate(t *testing.T) {
```
mimowo: Instead of testing the individual non-exported functions could we transfer the tests to the Reconcile function? 

Because building tests around internal functions starts to make future refactorings more difficult.

test/e2e/singlecluster/baseline/concurrent_admission_test.go
```diff
@@ -228,6 +228,122 @@ var _ = ginkgo.Describe("ConcurrentAdmission", ginkgo.Label("area:singlecluster"
 			})
 		})
 	})
+
+	ginkgo.When("Variants must preempt to be admitted", func() {
```
mimowo: I don't see a need for this e2e test. I think it can be fully reliably tested by integration test, because this test does not depend on nodes or pods, or any core k8s controllers.

pkg/controller/concurrentadmission/controller.go
```diff
@@ -51,10 +52,13 @@ import (
...
 const (
-	ConcurrentAdmissionController = "concurrent-admission-controller"
-	ReasonCreatedVariant          = "CreatedVariant"
-	ReasonActivatedVariant        = "ActivatedVariant"
-	ReasonDeactivatedVariant      = "DeactivatedVariant"
+	ConcurrentAdmissionController   = "concurrent-admission-controller"
+	ReasonCreatedVariant            = "CreatedVariant"
+	ReasonActivatedVariant          = "ActivatedVariant"
+	ReasonDeactivatedVariant        = "DeactivatedVariant"
+	ReasonOpenPreemptionGateVariant = "OpenPreemptionGateVariant"
```
mimowo:
```suggestion
	ReasonPreempionGateOpen = "PreempionGateOpen"
```
I think `OpenPreemptionGateVariant` does not read well, it is basically a verb "open" which is weird for an event type or reason for something that already happened. 

I think "OpenedPreempionGateForVariant" reads better, but then we lose the "Variant" wording, I think this is fine, but maybe we 

or maybe, `OpenedPreempionGateForCAVariant` , but then maybe we should also say "CAVariant" for other cases.

pkg/controller/concurrentadmission/controller.go
```diff
@@ -631,6 +653,103 @@ func (r *variantReconciler) shouldReconcile(wl *kueue.Workload) bool {
...
+	if timeToWait > 0 {
+		log.V(4).Info("Single preemption timeout did not expire", "timeLeft", timeToWait, "nextVariantToUngate", klog.KObj(candidate))
+		return nil, timeToWait
+	}
+	log.V(4).Info("Found variant to ungate", "variantToUngate", klog.KObj(candidate))
+	return candidate, preemptionTimeout
+}
+
+func firstCandidateVariant(log logr.Logger, variants []kueue.Workload) *kueue.Workload {
+	for _, wl := range admissibleVariants(variants) {
```
kshalot: AFAICS both `firstCandidateVariant` and `latestOpenGateTime` call `admissibleVariants(variants)` and iterate over that slice.

So a tiny optimization would be to calculate the `admissibleVariants` outside, in `variantToOpenPreemptionGate` and then simply pass it down into those calls.

I think it also makes it easier to see that `variantToOpenPreemptionGate` only considers admissible variants.

pkg/controller/concurrentadmission/controller.go
```diff
@@ -631,6 +653,103 @@ func (r *variantReconciler) shouldReconcile(wl *kueue.Workload) bool {
...
+		if workload.IsAdmissible(&variants[i]) {
+			ret = append(ret, &variants[i])
+		}
+	}
+	return ret
+}
+
+func findOpenPreemptionGate(wl *kueue.Workload) *kueue.PreemptionGateState {
+	idx := slices.IndexFunc(wl.Status.PreemptionGates, func(gate kueue.PreemptionGateState) bool {
+		return gate.Name == constants.ConcurrentAdmissionPreemptionGate && gate.Position == kueue.PreemptionGatePositionOpen
```
kshalot: We already have `HasClosedPreemptionGate` defined. So at least an analogous `HasOpenPreemptionGate` could be added instead of the lambda here. It's also defined in the integration test, so we could get rid of that as well.

pkg/controller/concurrentadmission/controller.go
kshalot: Or `PreemptionUngatedVariant` to try and keep the "Variant" the object/target of the reason, rather than the gate.

reruno: `PreemptionUngatedVariant` looks good to me

mimowo: sgtm

pkg/controller/concurrentadmission/controller.go
mimowo: Moreover, I'm pretty sure a set of similar helper functions has been developed for MultiKueue. I would really like to extract the helper functions to utils, say `pkg/util/preemption_gates` and use from both places. 

We could extract there HasPreemptionGateOpen(gateName), OpenPreemptionGate(gateName), ClosePreemptionGate(gateName).

@reruno let's maybe review the places which could be commonize between the two features so that we can make an informed decision if we do it in this PR or a follow up.

reruno: I think this is the set of helper functions that will be used in MK and CA
 
 I will create a prep PR, where we will use these helper function in MK, and then I will rebase on it for CA and reuse them for CA 
```
# in pkg/workload/
  FindPreemptionGate(wl, gateName) *kueue.PreemptionGate                                                                                                                
  HasOpenPreemptionGate(wl, gateName) bool                                                                                                                        
  OpenPreemptionGate(wl, gateName string))                                                                                                                                                                                                                                                                            
  EnsurePreemptionGateOnSpec(w *kueue.Workload, gateName string) bool                                                                                             
  RequiresPreemption(w *kueue.Workload) bool
```

mimowo (commented):
/lgtm
/approve
Thank you for working on this important improvement 👍 Happy to see the PreemptionGates getting another use case. Hope we can Beta them soon @kshalot

PR #9201, opened 2026-02-13
Fix race in RemoveRemoteObjects where remote LWS is not deleted

sohankunkerkar:
When MultiKueueGC deletes the remote workload before the reconciler processes the deletion, RemoveRemoteObjects would early-return on `remWl == nil` without deleting the remote controller object (e.g., LWS).

Move `DeleteRemoteObject` before the nil check to ensure the remote controller object is always deleted regardless of remote workload state.

#### What type of PR is this?
/kind bug

#### Which issue(s) this PR fixes:

Fixes #9178

#### Does this PR introduce a user-facing change?

```release-note
MultiKueue: Fix a bug that the remote Job object was occasionally left by MultiKueue GC, 
even when the corresponding Job object on the management cluster was deleted.
This issue was observed for LeaderWorkerSet.
```

Issue #5704, opened 2025-06-20
MultiKueue: Support cluster role sharing (worker and manager inside one cluster)


tenzen-y:
**What would you like to be added**:

I would like to support the cluster role sharing functionality for MultiKueue.

**Why is this needed**:

This allows us to reduce the number of managed clusters, which mitigates the operation costs for admins.

Indeed, the blocker for this feature request was mostly resolved by JobManagedBy feature and almost major Job integrations like Kubeflow and Ray support the managedBy functionality as well.

> Support for cluster role sharing (worker & manager inside one cluster) is out of scope for this KEP. We will get back to the topic once https://github.com/kubernetes/enhancements/pull/4370 is merged and becomes a wider standard.

https://github.com/kubernetes-sigs/kueue/tree/main/keps/693-multikueue#non-goals

**Completion requirements**:

This enhancement requires the following artifacts:

- [x] Design doc
- [ ] API change
- [x] Docs update

The artifacts should be linked in subsequent comments.

tenzen-y:
cc @mimowo @mwielgus @gabesaba @PBundyra

mimowo:
> Indeed, the blocker for this feature request was mostly resolved by JobManagedBy feature and almost major Job integrations like Kubeflow and Ray support the managedBy functionality as well.

Yeah, it feels that on clusters with 1.32+, with managedBy, it is already supported. We can have one CQ using MK and another CQ not.

Unless I'm missing something it feels like just about adding tests and updating the KEP.

tenzen-y:
> Yeah, it feels that on clusters with 1.32+, with managedBy, it is already supported. We can have one CQ using MK and another CQ not.
> 
> Unless I'm missing something it feels like just about adding tests and updating the KEP.

I hope that. Let us check the actual behavior!

mimowo:
I think @IrvingMg @mszadkow has a lot of experience setting up MultiKueue so can easily check that

Basically, we can have one CQ using MultiKueue and on the same cluster another CQ using the local cluster.

tenzen-y:
> I think [@IrvingMg](https://github.com/IrvingMg) [@mszadkow](https://github.com/mszadkow) has a lot of experience setting up MultiKueue so can easily check that
> 
> Basically, we can have one CQ using MultiKueue and on the same cluster another CQ using the local cluster.

If @IrvingMg or @mszadkow has enough bandwidth, feel free to take this issue.
[...]

PR #11877, opened 2026-06-01
[Fix] Remove MultiKueue owned Jobs only

### LINKED ISSUE #11849: MultiKueue may delete an existing worker-local Job when a dispatched remote Job has the same NamespacedName
**What happened**:

In a MultiKueue environment, I observed that the existing worker-local Jobs will be deleted when a Job is created on the manager cluster and a Job with the same NamespacedName already exists on the worker clusters. 
This occurs regardless of whether the existing worker-local Jobs are managed by kueue.

Before `wlReconciler` on manager cluster create remote workload onto worker clusters, 
it seems to delete all Jobs with same NamespacedName by `DeleteRemoteObject()` in `RemoveRemoteObjects()`:

https://github.com/kubernetes-sigs/kueue/blob/563f449618940b39805dcdec289c8a4d6e522d26/pkg/controller/admissionchecks/multikueue/workload.go#L132-L138

Is that intentional?

**How to reproduce it**
- Setup [MultiKueue](https://kueue.sigs.k8s.io/docs/tasks/manage/setup_multikueue/), and manager and worker cluster
- Create a job named `sample-job-a` on worker cluster.(call this job `Job(a)`)
- Create a job which has the same name of worker's(`sample-job-a`) on manager cluster specifying `kueue.x-k8s.io/queue-name` so that it is dispatched to worker cluster.(call this job `Job(b)`)

After `Job(b)` is created, `Job(a)` will be deleted and remote job `Job(b)` is created on worker cluster.
You can check this by comparing uid of Job object on worker cluster.

This change seems to have been introduced in https://github.com/kubernetes-sigs/kueue/pull/9201, but this problem might not been considered.

**Environment**:
- Kubernetes version: v1.35.1
- Kueue version: v0.18.0

mszadkow:
#### What type of PR is this?
/kind bug
/area multikueue

When a Job is created on the manager cluster while a Job with the same NamespacedName already exists on a worker cluster, the pre-existing worker-local Job is deleted.

Fixes #11849 

```release-note
MultiKueue: Fixed a bug where creating a Job on the manager cluster could
delete a pre-existing worker-local Job with the same namespace and name.
MultiKueue now deletes remote Jobs only when they are owned by MultiKueue.
```

## Summary by CodeRabbit

* **New Features**
  * Ownership validation added: remote resources are checked for the expected origin before status sync.
  * If a remote resource isn’t owned by MultiKueue, sync/deletion is skipped; retries now honor a sync timeout.

* **Bug Fixes**
  * Garbage collection avoids deleting worker-local resources that aren’t owned by the controller.

* **Tests**
  * Added/updated unit tests for ownership validation and conditional deletion.
  * Added an integration regression test to ensure pre-existing worker-local jobs are preserved.


mochizuki875:
@mszadkow 
Thank you for your fix.

By adding `DeleteRemoteObjectIfOwnedByMultiKueue()`, it will not delete pre-existing worker-local job with the same NamespacedName and return nil.
However, after skipping deletion, remote workloads may be created onto worker clusters following dispatching algorithm, and one of them will be admitted and remain on the worker cluster without `managedBy` field.

Is this intentional?
IMO, should we prevent remote workload from being created and admitted on the cluster where the job with same NamespacedName exist on?

mochizuki875:
By adding `DeleteRemoteObjectIfOwnedByMultiKueue()`, the possibility of performing `SyncJob()` on Job resources with the same NamespacedName has been reintroduced.
Upon further investigation, I've noticed two problems caused by this:

1. Pre-existing worker-local job will be mistakenly identified as remote Job
It may be the same behavior of previous version before https://github.com/kubernetes-sigs/kueue/pull/9201 was merged.

`jobSync()` is implemented in each multiKueueAdapter.
For example, in case of job, the status of pre-existing worker-local jobs on worker cluster will be reflected in the status of local jobs on the manager cluster.

https://github.com/kubernetes-sigs/kueue/blob/1744552f93d8a886c7265dd81aee3198d79424b6/pkg/controller/jobs/job/job_multikueue_adapter.go#L65-L83

2. The manager cluster may choose an inappropriate worker cluster.
Please assume that, there are two worker cluster, one has the local job resource which has same NamespacedName(call worker-1) and the other does not(call worker-2).
While the remote job should be dispatched on to worker-2 avoiding worker-1, the manager cluster selects the worker cluster containing the first admitted workload from among the worker clusters that contain the admitted workload.
So there is a possibility that worker-1 will be selected.

https://github.com/kubernetes-sigs/kueue/blob/1744552f93d8a886c7265dd81aee3198d79424b6/pkg/controller/admissionchecks/multikueue/workload.go#L113-L130

mszadkow:
> @mszadkow Thank you for your fix.
> 
> By adding `DeleteRemoteObjectIfOwnedByMultiKueue()`, it will not delete pre-existing worker-local job with the same NamespacedName and return nil. However, after skipping deletion, remote workloads may be created onto worker clusters following dispatching algorithm, and one of them will be admitted and remain on the worker cluster without `managedBy` field.
> 
> Is this intentional? IMO, should we prevent remote workload from being created and admitted on the cluster where the job with same NamespacedName exist on?

Good question, I think we should reference the [Cluster Role Sharing](https://github.com/kubernetes-sigs/kueue/issues/5704).

We can either:
1. Put the pressure on the customer to not create the same named jobs and update the documentation.
2. Prevent it completely by checking if the job or workload already exists (which I assume is not what we want).
3. Allow for the same named jobs to exist, but add all the guards.

@mimowo @tenzen-y  any thoughts?

pkg/controller/jobframework/multikueue.go
```diff
@@ -27,6 +28,22 @@ import (
...
+// otherwise deletes it. The Delete is guarded by a UID precondition pinned to the object
+// observed by Get, so a concurrent delete-and-recreate between Get and Delete cannot
+// cause us to remove an unrelated replacement. deleteOpts are forwarded to the Delete call.
+func DeleteRemoteObjectIfOwnedByMultiKueue(ctx context.Context, remoteClient client.Client, key types.NamespacedName, obj client.Object, deleteOpts ...client.DeleteOption) error {
+	if err := remoteClient.Get(ctx, key, obj); err != nil {
+		return client.IgnoreNotFound(err)
+	}
+	if _, owned := obj.GetLabels()[kueue.MultiKueueOriginLabel]; !owned {
+		return nil
+	}
```
mimowo: I think we should not just check if by "MultiKueue" , but also if it matches "our" origin name. so that we don't delete jobs owner by another multiKueue. I know this is a corner case, but it will make the logic more robust.

mimowo:
@mszadkow please update the release note, the prefix should always be the feature / area, not "fix" . The fix is annotated by the PR kind. Updated to "bug"

mimowo:
@mochizuki875 thank you for reviewing and your comment and findings in https://github.com/kubernetes-sigs/kueue/pull/11877#issuecomment-4643280638

> Pre-existing worker-local job will be mistakenly identified as remote Job
It may be the same behavior of previous version before https://github.com/kubernetes-sigs/kueue/pull/9201 was merged.

Could this be mitigated by checking if the MultiKueueOriginLabel corresponding to "our" MK? 

> jobSync() is implemented in each multiKueueAdapter.
For example, in case of job, the status of pre-existing worker-local jobs on worker cluster will be reflected in the status of local jobs on the manager cluster.

This is a good point, but here we could make the sync only if (1.) MultiKueueOriginLabel corresponds to "our" MK? 

Additionally we could strengthten the check by saving the UUID of the job on the manager in the job on the worker, so that later we can check the UUIDs for skip of the sync. Would that be needed in your case? If not, then maybe we can defer for later, but I think this is needed in the long term.

> The manager cluster may choose an inappropriate worker cluster.
Please assume that, there are two worker cluster, one has the local job resource which has same NamespacedName(call worker-1) and the other does not(call worker-2).
While the remote job should be dispatched on to worker-2 avoiding worker-1, the manager cluster selects the worker cluster containing the first admitted workload from among the worker clusters that contain the admitted workload.
So there is a possibility that worker-1 will be selected.

Yes, but this probably also could be fixed by checking "MultiKueueOriginLabel" for your needs, right?

Similarly as in the other points we could check if the job on worker-1 has the matching UUID (from the manager clusteR) imprinted in the annotation.

mimowo:
So, I would propose in this PR to check MultiKueueOriginLabel consistently in all places identified by @mochizuki875, and re-evaluate the need for checking also the UUIDs, but that would require imprinting the UUID on the worker workloads, and probably can be done in a follow up, because the scenario is a bit different - it is not about a job submitted to worker outside of MK, but re-creation of a job on the manager cluster.

mochizuki875:
@mimowo 
Thank you for your idea.
I agree your idea to check completely MultiKueueOriginLabel key-value.

> @mochizuki875 thank you for reviewing and your comment and findings in [#11877 (comment)](https://github.com/kubernetes-sigs/kueue/pull/11877#issuecomment-4643280638)
> 
> > Pre-existing worker-local job will be mistakenly identified as remote Job
> > It may be the same behavior of previous version before #9201 was merged.
> 
> Could this be mitigated by checking if the MultiKueueOriginLabel corresponding to "our" MK?
> 
> > jobSync() is implemented in each multiKueueAdapter.
> > For example, in case of job, the status of pre-existing worker-local jobs on worker cluster will be reflected in the status of local jobs on the manager cluster.
> 
> This is a good point, but here we could make the sync only if (1.) MultiKueueOriginLabel corresponds to "our" MK?

As you say, I think this miss-identified problem may be mitigated by identifying remote job using not only NamespacedName, but also MultiKueueOriginLabel when manager cluster will sync the status of the worker-local job.

> > The manager cluster may choose an inappropriate worker cluster.
> > Please assume that, there are two worker cluster, one has the local job resource which has same NamespacedName(call worker-1) and the other does not(call worker-2).
> > While the remote job should be dispatched on to worker-2 avoiding worker-1, the manager cluster selects the worker cluster containing the first admitted workload from among the worker clusters that contain the admitted workload.
> > So there is a possibility that worker-1 will be selected.
> 
> Yes, but this probably also could be fixed by checking "MultiKueueOriginLabel" for your needs, right?

On the other hand, I don't know about second one.
To select appropriate worker in this case, manager need to know which worker has same NamespacedName jobs and try to avoid them.
I can't imagine it can be done by checking MultiKueueOriginLabel.
So could you please tell me your idea more detail?

mimowo:
> To select appropriate worker in this case, manager need to know which worker has same NamespacedName jobs and try to avoid them.
> So could you please tell me your idea more detail?

Oh I see, I didn't understand fully the problem at the beginning so my idea isn't clear yet.

One relatively simple idea is to use Get just before dispatching to check if the Job already exists or not - and avoid dispatching workloads to such workers in the first place.

However, this may still be racy for TOCTOU as someone may create the workload in the meanwhile.

In that case we probably should mark workloads corresponding to pre-existing Jobs as Finished. We would not delete them to avoid re-creation, but just mark finished so that they don't get admitted.

mszadkow:
@mimowo @mochizuki875 
As this PR will be accepted I am happy to followup with UUID propagation to tackle "the second problem" with dispatching

pkg/controller/jobframework/multikueue.go
```diff
@@ -27,13 +30,47 @@ import (
...
+// ValidateRemoteObjectOwnership validates that the object is managed by this MultiKueue controller.
+// A MultiKueue-managed object must have the MultiKueueOriginLabel set to the controller origin.
+func ValidateRemoteObjectOwnership(obj client.Object, origin string) error {
+	if origin == "" {
+		return nil
+	}
+
+	if objOrigin, owned := obj.GetLabels()[kueue.MultiKueueOriginLabel]; owned && objOrigin == origin {
+		return nil
+	}
```
mimowo: These are rather unexpected scenarios so deserve logging at V2 at least.

pkg/controller/admissionchecks/multikueue/externalframeworks/adapter.go
```diff
@@ -145,12 +148,10 @@ func (a *Adapter) copyStatusFromRemote(localObj, remoteObj *unstructured.Unstruc
...
 // DeleteRemoteObject deletes the remote object identified by the given key from the remote cluster.
 // It first attempts to retrieve the object, and if found, deletes it with background propagation policy.
-func (a *Adapter) DeleteRemoteObject(ctx context.Context, _ client.Client, remoteClient client.Client, key types.NamespacedName) error {
+func (a *Adapter) DeleteRemoteObject(ctx context.Context, _ client.Client, remoteClient client.Client, key types.NamespacedName, origin string) error {
 	obj := &unstructured.Unstructured{}
 	obj.SetGroupVersionKind(a.gvk)
-	obj.SetName(key.Name)
-	obj.SetNamespace(key.Namespace)
-	return client.IgnoreNotFound(remoteClient.Delete(ctx, obj, client.PropagationPolicy(metav1.DeletePropagationBackground)))
+	return jobframework.DeleteRemoteObjectIfOwnedByMultiKueue(ctx, remoteClient, key, origin, obj, client.PropagationPolicy(metav1.DeletePropagationBackground))
```
mimowo: Have you considered a solution when we check the predonditions for deletion before `DeleteRemoteObject` is called? Doing the checks from within Adapters works, but it makes the adapaters more complex, and ultimtimatly we want other developers to maintain in-house adapters.

pkg/controller/jobframework/multikueue.go
```diff
@@ -27,13 +30,47 @@ import (
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
 )
 
+var ErrRemoteObjectNotOwnedByMultiKueue = errors.New("remote object is not owned by MultiKueue")
+
+// ValidateRemoteObjectOwnership validates that the object is managed by this MultiKueue controller.
+// A MultiKueue-managed object must have the MultiKueueOriginLabel set to the controller origin.
+func ValidateRemoteObjectOwnership(obj client.Object, origin string) error {
+	if origin == "" {
+		return nil
```
mochizuki875: `ValidateRemoteObjectOwnership()` will return `nil` when object satisfies label check.
Is it appropriate to treat the case where `origin == ""` as equivalent to this?

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -133,7 +133,7 @@ func (g *wlGroup) bestMatchByCondition(conditionType string) (*metav1.Condition,
 // The controller object is deleted first to handle cases where GC has already removed
 // the remote workload.
 func (g *wlGroup) RemoveRemoteObjects(ctx context.Context, cluster string) error {
-	if err := g.jobAdapter.DeleteRemoteObject(ctx, g.localClient, g.remoteClients[cluster].client, g.controllerKey); err != nil {
+	if err := g.jobAdapter.DeleteRemoteObject(ctx, g.localClient, g.remoteClients[cluster].client, g.controllerKey, g.remoteClients[cluster].origin); err != nil {
```
mochizuki875: Related to [this](https://github.com/kubernetes-sigs/kueue/pull/11877/changes#r3379018906), can we check the condition of the remote object before this line?
If it can, we may not need to implement `DeleteRemoteObjectIfOwnedByMultiKueue` and call it in each adapters.

pkg/controller/jobframework/multikueue.go
mszadkow: added

pkg/controller/admissionchecks/multikueue/externalframeworks/adapter.go
mszadkow: That sounds reasonable, I have updated

pkg/controller/admissionchecks/multikueue/workload.go
mszadkow: Yes, we can, I have added that

pkg/controller/jobframework/multikueue.go
```diff
@@ -18,21 +18,86 @@ package jobframework
...
+	owned, err := isRemoteObjectOwnedByMultiKueue(ctx, remoteClient, key, origin, obj)
+	if err != nil {
+		return err
+	}
+	if !owned {
+		return nil
+	}
+
+	return adapter.DeleteRemoteObject(ctx, localClient, remoteClient, key)
+}
```
mochizuki875: Should we need to implement `isRemoteObjectOwnedByMultiKueue()` because it seems to be used only in `DeleteRemoteObjectIfOwned()`?
How about using `ValidateRemoteObjectOwnership()` directly in `DeleteRemoteObjectIfOwned()`?


```suggestion
// DeleteRemoteObjectIfOwned checks whether the remote object represented
// by adapter.GVK() is owned by this MultiKueue origin and invokes adapter delete only
// when ownership matches.
func DeleteRemoteObjectIfOwned(ctx context.Context, localClient client.Client, remoteClient client.Client, adapter MultiKueueAdapter, key types.NamespacedName, origin string) error {
	obj := &unstructured.Unstructured{}
	obj.SetGroupVersionKind(adapter.GVK())

	if err := remoteClient.Get(ctx, key, obj); err != nil {
		if client.IgnoreNotFound(err) == nil {
			log.V(2).Info("Skipping remote object deletion because remote object was not found")
			return nil
		}
		return err
	}
	if err := ValidateRemoteObjectOwnership(obj, origin); err != nil {
		objOrigin, owned := obj.GetLabels()[kueue.MultiKueueOriginLabel]
		log.V(2).Info("Skipping remote object deletion because remote object is not owned by this MultiKueue origin", "hasOriginLabel", owned, "objectOrigin", objOrigin)
		return nil
	}

	return adapter.DeleteRemoteObject(ctx, localClient, remoteClient, key)
}
```

pkg/controller/jobframework/multikueue.go
mszadkow: This one is problematic, this should return an error probably as the empty originLabel

pkg/controller/jobframework/multikueue.go
mszadkow: done

mszadkow:
@mochizuki875 ptal

mimowo:
cc @olekzabl @kshalot ptal

pkg/controller/jobs/appwrapper/appwrapper_multikueue_adapter.go
```diff
@@ -54,6 +54,9 @@ func (b *multiKueueAdapter) SyncJob(ctx context.Context, localClient client.Clie
 
 	// if the remote exists, just copy the status
 	if err == nil {
+		if err := jobframework.ValidateRemoteObjectOwnership(ctx, &remoteAppWrapper, origin); err != nil {
+			return err
```
kshalot: Do I see correctly that the reconciler for the workload group will just return this error:
https://github.com/kubernetes-sigs/kueue/blob/1ba1b7fc37246aa40c756e2ebe5b661d1346c1d7/pkg/controller/admissionchecks/multikueue/workload.go#L496-L500

So it will just keep looping until you notice the log until the remote-local object is removed? Since this workload already has quota reserved, this would lock resources.

Maybe in such cases we should consider something more final. If so, this could be a follow-up, since preventing the unowned objects from being deleted is the priority in my eyes.

pkg/controller/jobframework/multikueue.go
```diff
@@ -18,15 +18,70 @@ package jobframework
...
+		return nil
+	}
+
+	return fmt.Errorf("%w: expected %q=%q on %T %q", ErrRemoteObjectNotOwnedByMultiKueue, kueue.MultiKueueOriginLabel, origin, obj, client.ObjectKeyFromObject(obj))
+}
+
+// DeleteRemoteObjectIfOwned fetches the remote object for the given adapter's GVK and key,
+// skips deletion if the object does not exist or is not owned by this MultiKueue origin,
+// and otherwise delegates to adapter.DeleteRemoteObject.
+func DeleteRemoteObjectIfOwned(ctx context.Context, localClient client.Client, remoteClient client.Client, adapter MultiKueueAdapter, key types.NamespacedName, origin string) error {
```
kshalot: This helper has to do a little ping pong with the adapter, i.e. take the GVK and then call `DeleteRemoteObject`.

But I'm wondering if it wouldn't be cleaner to just pass `origin` to the `DeleteRemoteObject` function on the adapter, like in `SyncJob`?  From what I see, we are checking the origin in all cases, so it might as well be part of the adapter contract

pkg/controller/jobs/appwrapper/appwrapper_multikueue_adapter.go
mszadkow: Yes, right now it will loop.
We could apply the same thing as the `workerLostTimeout`, then after time expires we could re-admit.

pkg/controller/jobframework/multikueue.go
mszadkow: We could do that but IIUC @mimowo wanted something to be called before the `DeleteRemoteObject` func, to  separate this logic from the one that is adapters' owned.
https://github.com/kubernetes-sigs/kueue/pull/11877#discussion_r3379018906

mochizuki875:
@mszadkow 

Thank you for your addressing.

By the way, we are currently working to prevent unowned job resources from being deleted, but where should we ultimately land to resolve  #11849 ?
In other words, I think we need to prepare the answer for this question(I should have told sooner, but I'm sorry for the delay.):

"What will happen each on the manager cluster and the worker clusters if user create a job resource when the job resources with same NamespacedName are exist on the worker clusters?"

As we are working now, preventing the pre-existing worker-local Jobs from being deleted is definitely key point of this, but I think we should define how it looks from the user's perspective.
Especially defining how a user of manager cluster notice the situation where remote job cannot be dispatched to any worker cluster by the same NamespacedName job existing is also important.

mszadkow:
> "What will happen each on the manager cluster and the worker clusters if user create a job resource when the job resources with same NamespacedName are exist on the worker clusters?"

Let me translate it into concrete scenario:
```
Job xyz/job1 is running on the worker cluster (but not MultiKueue owned), now Multikueue wants to dispatch workload for job xyz/job1 into the worker cluster.
```

Then we have:

Worker side:
 - MultiKueue does not delete that pre-existing non-owned Job.
It does not take ownership of it.

Manager side:
 - SyncJob reads remote object.
 - Ownership validation fails because origin label is missing or mismatched.
- Reconcile retries; if admission check is Ready:
   * currently it will loop until manual intervention happens
   * or if we go [with](https://github.com/kubernetes-sigs/kueue/pull/11877#discussion_r3402179670): after sync timeout it transitions to Retry with timed-out message
   
   
 Or if you ask the other way around:
 ```
 Multikueue dispatched workload admission on the worker cluster creates Job xyz/job1 (multikueue owned)
Now user creates Job xyz/job1 on the worker while the job is slready running.
 ```
 
 Worker side:
 - API server returns AlreadyExists, so no second Job is created
 Manager side:
 - nothing special changes just because of that failed create
 - Sync continues with the existing owned remote Job

mochizuki875:
@mszadkow 
Thank you for explanation.

I've got understand pre-existing non-owned Job will remain on worker cluster and remote job will not be dispatched on it.
To say extremely, nothing change on worker cluster.

On the other hand, how about the status of job and workload on the manager cluster?
Will the job keeps suspended? Will (or should) the workload  be admitted or quota reserved?

mszadkow:
> @mszadkow Thank you for explanation.
> 
> I've got understand pre-existing non-owned Job will remain on worker cluster and remote job will not be dispatched on it. To say extremely, nothing change on worker cluster.
> 
> On the other hand, how about the status of job and workload on the manager cluster? Will the job keeps suspended? Will (or should) the workload be admitted or quota reserved?

Well, manager cluster job status can't be synced as there is no MK owned remote job.
Manager workload on the other hand will wait for first reserving workload of the remote workers.
If it choose the worker where we have the job already it will stuck on sync validation.
Then we should order re-admission after some timeout.

That means quota could be reserved in the remote worker which already has the job, but the job will remain suspended.
@kshalot am I correct?

pkg/controller/jobs/appwrapper/appwrapper_multikueue_adapter.go
mszadkow: we still have eviction and finish handling using SyncJob, but I can't imagine if something can happen with the origin label of the remote jobs

mochizuki875:
> Manager workload on the other hand will wait for first reserving workload of the remote workers.
If it choose the worker where we have the job already it will stuck on sync validation.
Then we should order re-admission after some timeout.

@mszadkow 
At the first place, should the remote workload on the worker with pre-existing non-owned job be able to reserve quota?
Is this a waste of time and effort?

In addition, I think it should not be admitted because it will be reflected to workload on manager and make users confused.
Users may expect that the remote job will actually run on the worker after some time has passed by checking the manager workload status(QuotaReserved=true, Admitted=true).

@mimowo has suggested a idea related to it(https://github.com/kubernetes-sigs/kueue/pull/11877#issuecomment-4651139178).

mochizuki875:
> In addition, I think it should not be admitted because it will be reflected to workload on manager and make users confused.
Users may expect that the remote job will actually run on the worker after some time has passed by checking the manager workload status(QuotaReserved=true, Admitted=true).

Sorry, I may misunderstood this.
While Remote workload should still not be admitted on worker, remote workload status seems not to be reflected to manager workload status unless `SyncJob()` succeeded.
So  my understanding is that users can recognize the situation where remote job is not running on worker, by checking manager workload status(Admitted != True), right?

https://github.com/kubernetes-sigs/kueue/blob/01e576255b55275520ff4ab0ebfdfa443ae7bcc2/pkg/controller/admissionchecks/multikueue/workload.go#L496-L504

mszadkow:
> > Manager workload on the other hand will wait for first reserving workload of the remote workers.
> > If it choose the worker where we have the job already it will stuck on sync validation.
> > Then we should order re-admission after some timeout.
> 
> @mszadkow At the first place, should the remote workload on the worker with pre-existing non-owned job be able to reserve quota? Is this a waste of time and effort?
> 
> In addition, I think it should not be admitted because it will be reflected to workload on manager and make users confused. Users may expect that the remote job will actually run on the worker after some time has passed by checking the manager workload status(QuotaReserved=true, Admitted=true).
> 
> @mimowo has suggested a idea related to it([#11877 (comment)](https://github.com/kubernetes-sigs/kueue/pull/11877#issuecomment-4651139178)).

I found this comment confusing:

> In that case we probably should mark workloads corresponding to pre-existing Jobs as Finished. We would not delete them to avoid re-creation, but just mark finished so that they don't get admitted.

We should probably mark the workload we dispatched to that remote worker as Finished.
Not the one related to the pre-existing Job - otherwise I don't see how this helps.

I think the same-name pre-existing Job conflict should be detected on the worker side where the cloned workload is admitted/materialized, since that is the authoritative state and avoids manager-side TOCTOU. Then the manager-side MultiKueue reconciler can react by retrying/re-admitting elsewhere rather than treating it as a finished workload.

But that belongs to another PR, I wouldn't over extend this PR.

However yes, as you asked, the full picture after all effort should behave as you described.

mszadkow:
> > In addition, I think it should not be admitted because it will be reflected to workload on manager and make users confused.
> > Users may expect that the remote job will actually run on the worker after some time has passed by checking the manager workload status(QuotaReserved=true, Admitted=true).
> 
> Sorry, I may misunderstood this. While Remote workload should still not be admitted on worker, remote workload status seems not to be reflected to manager workload status unless `SyncJob()` succeeded. So my understanding is that users can recognize the situation where remote job is not running on worker, by checking manager workload status(Admitted != True), right?
> 
> https://github.com/kubernetes-sigs/kueue/blob/01e576255b55275520ff4ab0ebfdfa443ae7bcc2/pkg/controller/admissionchecks/multikueue/workload.go#L496-L504

Yes, exactly, the whole idea of having manager side is also to have one place where you can observe all your jobs statuses.

mszadkow:
@mochizuki875 @kshalot any comments?

pkg/controller/jobframework/multikueue_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+func TestValidateRemoteObjectOwnership(t *testing.T) {
+	makeJob := func(labels map[string]string) *batchv1.Job {
+		return &batchv1.Job{ObjectMeta: metav1.ObjectMeta{Name: "test", Namespace: "default", Labels: labels}}
+	}
+
+	tests := map[string]struct {
+		obj     *batchv1.Job
+		origin  string
+		wantErr bool
```
mykysha: Nit: can we check the exact error returned, instead of just the presence of an error?

pkg/controller/jobframework/multikueue.go
kshalot: I see.

On one hand, it makes the adapter leaner, on the other we are already do a similar check in `SyncJob`, it's also responsible of setting the labels etc. So some of this "low-level" logic is already seeping into the adapters.

But I'm not saying that this is a reason why we should continue going in that direction, I'm okay with keeping this as is. Maybe the contract could be revamped at some point so all the necessary checks and updates are on the "MultiKueue side".

pkg/controller/jobframework/multikueue.go
```diff
@@ -18,15 +18,70 @@ package jobframework
...
+
+	if err := remoteClient.Get(ctx, key, obj); err != nil {
+		if client.IgnoreNotFound(err) == nil {
+			log.V(2).Info("Skipping remote object action because object was not found")
+			return nil
+		}
+		return err
+	}
+
+	objOrigin, owned := obj.GetLabels()[kueue.MultiKueueOriginLabel]
```
kshalot: Could we maybe have a single place for this logic, i.e. re-use `ValidateRemoteObjectOwnership` here? It might require some slight changes though.

kshalot:
> Well, manager cluster job status can't be synced as there is no MK owned remote job.
Manager workload on the other hand will wait for first reserving workload of the remote workers.
If it choose the worker where we have the job already it will stuck on sync validation.
Then we should order re-admission after some timeout.
>
> That means quota could be reserved in the remote worker which already has the job, but the job will remain suspended.
@kshalot am I correct?

IIUC the risk here is that the workload will be admitted in a worker which already has a job with a conflicting name, then the workload will be stuck until the conflicting job is removed. Which in my opinion is an improvement over the status quo, where we remove a non-MK job, but it's still far from ideal.

I'd follow this up with a more user-friendly fix. For example retriggering the whole admission process or deactivating the workload on the manager.

pkg/controller/jobframework/multikueue.go
mszadkow: Yes, I will reuse ValidateRemoteObjectOwnership

mszadkow:
> > Well, manager cluster job status can't be synced as there is no MK owned remote job.
> > Manager workload on the other hand will wait for first reserving workload of the remote workers.
> > If it choose the worker where we have the job already it will stuck on sync validation.
> > Then we should order re-admission after some timeout.
> > That means quota could be reserved in the remote worker which already has the job, but the job will remain suspended.
> > @kshalot am I correct?
> 
> IIUC the risk here is that the workload will be admitted in a worker which already has a job with a conflicting name, then the workload will be stuck until the conflicting job is removed. Which in my opinion is an improvement over the status quo, where we remove a non-MK job, but it's still far from ideal.
> 
> I'd follow this up with a more user-friendly fix. For example retriggering the whole admission process or deactivating the workload on the manager.

I think that sounds like a syncTimeout that I added [here](https://github.com/epam/kubernetes-kueue/blob/1ad6a540932b2263304c022bb8ae04e995c1b2a0/pkg/controller/admissionchecks/multikueue/workload.go#L535).
No need for manual intervention, but logs will show the issue
So when we stuck we can force re-admission, wdyt?

mszadkow:
@coderabbitai review

kshalot:
> I think that sounds like a syncTimeout that I added [here](https://github.com/epam/kubernetes-kueue/blob/1ad6a540932b2263304c022bb8ae04e995c1b2a0/pkg/controller/admissionchecks/multikueue/workload.go#L535). No need for manual intervention, but logs will show the issue So when we stuck we can force re-admission, wdyt?

Yes, I must have missed this when re-reading the diff. Thanks for pointing that out

I have two loose thoughts. I'll write them down here first and then keep thinking:
1. Should this `syncTimeout` be specific to this name conflict issue that we are solving here? I'm thinking about other of classes of issues that might prevent us from creating an object during `SyncJob`. I'm hand-waving here, but I can think of things like `ResourceQuota` being set for the number of objects (say `batch/job`) or maybe there are some cluster-specific validation webhooks. I'm wondering if this should be generalized.
2. We could also flag the conflict more visibly, e.g. by deactivating the workload. This signals to the admin that there are naming conflicts which might be a systemic issue on the user end. We might prefer to push them towards a unique naming scheme or better MultiKueue isolation.

CC @olekzabl who might be interested in chipping in.

pkg/controller/jobs/appwrapper/appwrapper_multikueue_adapter.go
```diff
@@ -54,6 +54,9 @@ func (b *multiKueueAdapter) SyncJob(ctx context.Context, localClient client.Clie
 
 	// if the remote exists, just copy the status
 	if err == nil {
+		if err := jobframework.ValidateRemoteObjectOwnership(ctx, &remoteAppWrapper, origin); err != nil {
+			return false, err
+		}
```
mimowo: Hm, can we move the code outside of SyncJob to avoid duplicating across of adapers? I suppose the code will always need to be the same which is a sign that it should be outside.

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -63,31 +63,38 @@ var (
...
+	// unsuspend latency (so we don't spin), and well below anything a user would
+	// notice. See https://github.com/kubernetes-sigs/kueue/issues/11115.
+	//
+	// This is the canonical explanation of the deferred-sync handling; other sites
+	// in this package point here rather than repeating it.
+	syncDeferredRequeueAfter = 2 * time.Second
+
+	// defaultSyncTimeoutDuration is the maximum time to wait for syncing the remote
+	// controller object before timing out and retrying the admission check.
+	defaultSyncTimeoutDuration = 5 * time.Minute
```
mimowo: I think the name is misleading because it suggests "we will timeout SyncJob if it takes too long". 

An idea for a better name: `delayForSyncJobDueToRemoteJobConflict`

However, I would not do this at all. I think the situation is quite unusual that there are Jobs. I don't think this PR should be oriented on the UX experience for this case. Before we start overtuning UX for this scenario let's make sure this is really needed. I read the Issue more like "let's avoid destructive job deletion". Using controller-runtime error handling already handles that.

Also, assuming the remote job exists say for 1 day, it is better UX to use controller-runtime backoff because the backoff will grow to 15min. So while improving for short living jobs we may be worsening for long living jobs. TLDR, I wouldn't do any of that at all (at least until proven to be a pain point for users).

mimowo: cc @mochizuki875 please let me know if this disagree with your assessment here.

pkg/controller/jobs/appwrapper/appwrapper_multikueue_adapter.go
mszadkow: I will check this

mimowo:
@mszadkow have you had a chance to test the fix on a real cluster? I'm wondering how this behaves in practice, assume you have 2 worker clusters:
- worker 1, empty
- worker 2, with conflicting job and workload

IIUC we will:
1. dispatch the workload to both clusters
2a. admit on worker 1 - then we will delete both workloads, but will keep the pre-existing workload and job in tact
2b. admit on worker 2 - then we will get continues errors?

I guess ideally we would change (1.) and not even dispatch the workload to "worker 2", but IIUC this is out of scope for the PR?

It would be great if you can test and summarize the current behavior because is rather tricky to imagine the scenarios just by static code analysis, at least for me.

mszadkow:
> @mszadkow have you had a chance to test the fix on a real cluster? I'm wondering how this behaves in practice, assume you have 2 worker clusters:
> 
> * worker 1, empty
> * worker 2, with conflicting job and workload
> 
> IIUC we will:
> 
> 1. dispatch the workload to both clusters
>    2a. admit on worker 1 - then we will delete both workloads, but will keep the pre-existing workload and job in tact
>    2b. admit on worker 2 - then we will get continues errors?
> 
> I guess ideally we would change (1.) and not even dispatch the workload to "worker 2", but IIUC this is out of scope for the PR?
> 
> It would be great if you can test and summarize the current behavior because is rather tricky to imagine the scenarios just by static code analysis, at least for me.

OK, ready to report both scenarios went exactly as assumed.

In 2a, job1 gets to the worker1 and does not disturb job1 from worker2.
And in 2b, yes we get continues logs of:
```
{"level":"Level(-2)","ts":"2026-06-18T13:50:49.16925678Z","logger":"multikueue-workload","caller":"multikueue/workload.go:182","msg":"Reconcile Workload","replica-role":"leader","namespace":"default","name":"job-job1-dfe89","reconcileID":"faf5b41b-e385-4ed4-abb9-747eb1fbca26"}
{"level":"error","ts":"2026-06-18T13:50:49.194519343Z","logger":"multikueue-workload","caller":"multikueue/workload.go:537","msg":"validating remote controller object","replica-role":"leader","namespace":"default","name":"job-job1-dfe89","reconcileID":"faf5b41b-e385-4ed4-abb9-747eb1fbca26","op":"reconcileGroup","remote":"multikueue-test-worker2","remoteWorkload":{"name":"job-job1-dfe89","namespace":"default"},"cluster":"multikueue-test-worker2","error":"remote object is not owned by MultiKueue: expected \"kueue.x-k8s.io/multikueue-origin\"=\"multikueue\" on *v1.PartialObjectMetadata \"default/job1\"","stacktrace":"sigs.k8s.io/kueue/pkg/controller/admissionchecks/multikueue.(*wlReconciler).reconcileGroup\n\t/workspace/pkg/controller/admissionchecks/multikueue/workload.go:537\nsigs.k8s.io/kueue/pkg/controller/admissionchecks/multikueue.(*wlReconciler).Reconcile\n\t/workspace/pkg/controller/admissionchecks/multikueue/workload.go:262\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Reconcile\n\t/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:221\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:478\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:437\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func1.1\n\t/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:312"}
{"level":"error","ts":"2026-06-18T13:50:49.194867124Z","logger":"multikueue-workload","caller":"controller/controller.go:494","msg":"Reconciler error","replica-role":"leader","namespace":"default","name":"job-job1-dfe89","reconcileID":"faf5b41b-e385-4ed4-abb9-747eb1fbca26","error":"remote object is not owned by MultiKueue: expected \"kueue.x-k8s.io/multikueue-origin\"=\"multikueue\" on *v1.PartialObjectMetadata \"default/job1\"","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:494\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:437\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func1.1\n\t/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:312"}
```

mimowo:
Excellent news 👍  And when the conflicting job ends, then the MultiKueue job gets running, right?

Certainly the sequence of errors is not the ideal UX, but I think we really don't need to over-optimize the scenario. I think admins could use basically another LQ/CQ on the worker to run local jobs. 

Still, in the ideal UX (2b) does not error, but instead we just never admit on worker2, or skip even disaptching to worker 2. However, fixing that would be more involving, so I'm happy to go ahead with the MVP fix just for the disruptive action of delete.

mochizuki875:
@mimowo @mszadkow 
Sorry I couldn't respond these past few days.
I agree with you.

In this PR, I think it's enough to meet them at least:
- avoiding pre-existing job deletion
- letting the manager cluster the remote job is not being executed(like [this](https://github.com/kubernetes-sigs/kueue/pull/11877#issuecomment-4705975212)) 

IMO: Skip dispatching to worker2 is related to not only UX but also effective use of resources.
So it's better to improve it in another PR after it's merged.

mimowo:
> IMO: Skip dispatching to worker2 is related to not only UX but also effective use of resources.
So it's better to improve it in another PR after it's merged.

Cool, let's track this improvement as a follow up issue. In order to determine the priority we will need more feedback about the use cases for users interacting directly with the worker clusters. Also, I suppose some in-house platform could most of the time fix this issue by suffixing jobs based on the shortened cluster name, etc.

Still, I accept there is a room for improvement. @mochizuki875 would you like to open the follow up issue and present the use cases so that we can triage the issue best?

mimowo:
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for the important bugfix, deleting remote jobs created by users was certainly a disruptive operation 👍 


Just to manually confirm the answer to my question "And when the conflicting job ends, then the MultiKueue job gets running, right?", so that the MK job can happily progress once the Job on the worker cluster is deleted.
/hold

mszadkow:
Yes I confirm, after the conflicting job ends, then the MultiKueue job gets running.
/unhold

mszadkow:
/test kueue-test-e2e-certmanager-upgrade-main

Seems that CI related:
```
There are no nodes that your pod can schedule to - check your requests, tolerations, and node selectors (skip schedule deleting pod: test-pods/405492db-cf3b-47bc-9a23-e8910bf57f6a)
```

mimowo:
@mszadkow please prep CP for 0.17

mochizuki875:
@mszadkow 
Thank you for your work!

mochizuki875:
> > IMO: Skip dispatching to worker2 is related to not only UX but also effective use of resources.
> > So it's better to improve it in another PR after it's merged.
> 
> Cool, let's track this improvement as a follow up issue. In order to determine the priority we will need more feedback about the use cases for users interacting directly with the worker clusters. Also, I suppose some in-house platform could most of the time fix this issue by suffixing jobs based on the shortened cluster name, etc.
> 
> Still, I accept there is a room for improvement. @mochizuki875 would you like to open the follow up issue and present the use cases so that we can triage the issue best?

Sure, I'll submit another issue later( #12404 ).

mimowo:
/release-note-edit
```release-note
MultiKueue: Fixed a bug where creating a Job on the manager cluster could
delete a pre-existing worker-local Job with the same namespace and name.
MultiKueue now deletes remote Jobs only when they are owned by MultiKueue.
```

Issue #11644, opened 2026-05-26
Breakdown kueue_admitted_active_workloads and kueue_pending_workloads metrics by flavor


polinasand:
**What would you like to be added**:
We would like to be able to breakdown by flavor the next metrics:
a) `kueue_admitted_active_workloads`
b) `kueue_pending_workloads`

**Why is this needed**:
This ask comes from customer wanting to measure pending demand on Kueue scheduler stratified by flavor. This is because mixed flavor workloads are sent to Kueue at the moment.

This enhancement requires the following artifacts:

- [ ] Design doc
- [x] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

olekzabl:
/assign

Taking over from @polinasand as she will no longer work on this.

PR #11888, opened 2026-06-02
Add kueue_flavor_admitted_active_workloads metric

olekzabl:
#### What type of PR is this?

/kind feature

Adds a new metric `kueue_flavor_admitted_active_workloads`, reporting admitted workloads count in CQ broken down by flavor.

This is built on top of #11504.

Part of #11644.

```release-note
Add a `kueue_flavor_admitted_active_workloads` metric, counting admitted workloads per flavor.
```


olekzabl:
/kind feature

I placed it in PR description but Prow seems to somehow have missed it.

pkg/features/kube_features.go
```diff
@@ -435,11 +435,19 @@ const (
...
 	// Enable reporting of Cohort related metrics (also including ClusterQueueInfo metric).
 	MetricsForCohorts featuregate.Feature = "MetricsForCohorts"
+
 	// owner: @tenzen-y
 	// kep: https://github.com/kubernetes-sigs/kueue/tree/main/keps/2724-topology-aware-scheduling
 	// issue: https://github.com/kubernetes-sigs/kueue/issues/10659
 	// Enable accurately topology aware scheduling when multiple flavors cover the same Node.
 	TASHandleOverlappingFlavors featuregate.Feature = "TASHandleOverlappingFlavors"
+
+	// owner: @polinasand
```
kshalot:
```suggestion
	// owner: @olekzabl
```

pkg/features/kube_features.go
```diff
@@ -435,11 +435,19 @@ const (
...
+	// kep: https://github.com/kubernetes-sigs/kueue/issues/11644
```
kshalot:
```suggestion
	// issue: https://github.com/kubernetes-sigs/kueue/issues/11644
```

pkg/features/kube_features.go
```diff
@@ -686,6 +694,9 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 	TASHandleOverlappingFlavors: {
 		{Version: version.MustParse("0.18"), Default: true, PreRelease: featuregate.Beta},
 	},
+	KueueAdmittedWorkloadsCountByFlavor: {
+		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Beta},
```
kshalot: It's false, but in beta. If this should be disabled by default, I'd say this should be in alpha.

Although personally I think this can be enabled by default:

```suggestion
		{Version: version.MustParse("0.18"), Default: true, PreRelease: featuregate.Beta},
```

pkg/cache/scheduler/clusterqueue.go
```diff
@@ -584,6 +590,12 @@ func (c *clusterQueue) updateWorkloadUsage(log logr.Logger, wi *workload.Info, o
 		updateFlavorUsage(frUsage, c.AdmittedUsage, op)
 		c.Parent().updateAdmittedWorkloadsCount(op.asSignedOne())
 		c.admittedWorkloadsCount += op.asSignedOne()
+
+		flavors := sets.New[kueue.ResourceFlavorReference]()
+		for fr := range frUsage {
+			flavors.Insert(fr.Flavor)
+		}
+		updateAdmittedWorkloadsCountByFlavor(flavors, c.admittedWorkloadsCountByFlavor, op)
```
kshalot: I'd rather that all new code would be hidden behind the feature gate. It's less wasteful and potentially safer.

pkg/cache/scheduler/clusterqueue.go
```diff
@@ -584,6 +590,12 @@ func (c *clusterQueue) updateWorkloadUsage(log logr.Logger, wi *workload.Info, o
 		updateFlavorUsage(frUsage, c.AdmittedUsage, op)
 		c.Parent().updateAdmittedWorkloadsCount(op.asSignedOne())
 		c.admittedWorkloadsCount += op.asSignedOne()
+
+		flavors := sets.New[kueue.ResourceFlavorReference]()
+		for fr := range frUsage {
+			flavors.Insert(fr.Flavor)
+		}
```
kshalot: Nit: Since this is all private functions, we could skip building this set and just pass `frUsage` to `updateAdmittedWorkloadsCountByFlavor`. Then it would be all encapsulated in the function call.

pkg/metrics/metrics.go
```diff
@@ -706,6 +710,17 @@ The label 'reason' can have the following values:
 	)
 	trackGaugeVec(AdmittedActiveWorkloads, gaugeCleanupScopeClusterQueueCache)
 
+	// +metricsdoc:group=clusterqueue
+	// +metricsdoc:labels=cluster_queue="the name of the ClusterQueue",replica_role="one of `leader`, `follower`, or `standalone`",flavor="the resource flavor name"
+	FlavorAdmittedActiveWorkloads = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: constants.KueueName,
+			Name:      "flavor_admitted_active_workloads",
+			Help:      "The number of admitted Workloads that are active (unsuspended and not finished), per 'cluster_queue' and 'flavor'",
```
kshalot: The help of `AdmittedActiveWorkloads` reads:

> The number of admitted Workloads that are active, per 'cluster_queue'

We could keep them the same so it's clear that it's the same thing (modulo the breakdown by flavor).

Also, it might be good to mention that this metric does not sum to the total workload count because of workloads with multiple flavors.

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
kshalot, line 1: Should this be in the fair sharing suite?

kshalot (commented):
@olekzabl Should be close #11504, which is where most if not all of this diff comes from?

pkg/features/kube_features.go
olekzabl: Done

pkg/cache/scheduler/clusterqueue.go
olekzabl: OK. I initially preferred to pull it out - but you convinced me.

test/integration/singlecluster/scheduler/fairsharing/metrics_test.go
olekzabl: Sticking to the convention. This is the only file testing the original metric ("admitted workloads" with no breakdown).

olekzabl:
> @olekzabl Should be close #11504, which is where most if not all of this diff comes from?

Not all :) See the commits history.

Yes, we should close #11504, though I seem unable to. Will ask Polina.

kshalot:
> Yes, we should close #11504, though I seem unable to. Will ask Polina.

I closed it (with the slashclose command).

pkg/cache/scheduler/clusterqueue.go
```diff
@@ -630,6 +639,17 @@ func updateFlavorUsage(newUsage resources.FlavorResourceQuantities, oldUsage res
...
 }
 
+func updateAdmittedWorkloadsCountByFlavor(frUsage resources.FlavorResourceQuantities, admittedWorkloadsCountByFlavor map[kueue.ResourceFlavorReference]int, op usageOp) {
+	sign := op.asSignedOne()
+	flavors := sets.New[kueue.ResourceFlavorReference]()
+	for fr := range frUsage {
+		flavors.Insert(fr.Flavor)
+	}
+	for flavor := range flavors {
+		admittedWorkloadsCountByFlavor[flavor] += sign
```
kshalot, line 649:
```suggestion
	flavors := sets.New[kueue.ResourceFlavorReference]()
	for fr := range frUsage {
		flavors.Insert(fr.Flavor)
	}
	for flavor := range flavors {
		admittedWorkloadsCountByFlavor[flavor] += op.asSignedOne()
```

Nano-nit: I don't care much whether this is applied or not

pkg/features/kube_features.go
```diff
@@ -435,11 +435,19 @@ const (
...
 	// kep: https://github.com/kubernetes-sigs/kueue/tree/main/keps/2724-topology-aware-scheduling
 	// issue: https://github.com/kubernetes-sigs/kueue/issues/10659
 	// Enable accurately topology aware scheduling when multiple flavors cover the same Node.
 	TASHandleOverlappingFlavors featuregate.Feature = "TASHandleOverlappingFlavors"
+
+	// owner: @polinasand
+	// kep: https://github.com/kubernetes-sigs/kueue/issues/11644
+	//
+	// When enabled, metrics for Kueue active admitted workloads count by flavor
+	// are populated by clusterqueue.
```
kshalot: I don't love this phrasing but I also don't think it's a big deal. Usually the FGs are described as "Enable ..." and not "When enabled...".

Sth like:
```suggestion
	// Enable reporting of the count of active admitted workloads broken down by flavor.
```

I think it should also be clear that this does not change the existing metric, but adds a new one.

pkg/features/kube_features.go
```diff
@@ -686,6 +694,9 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 	TASHandleOverlappingFlavors: {
 		{Version: version.MustParse("0.18"), Default: true, PreRelease: featuregate.Beta},
 	},
+	AdmittedWorkloadsCountByFlavor: {
+		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Beta},
```
kshalot: https://github.com/kubernetes-sigs/kueue/pull/11888/changes#r3341701619

^ Here, I proposed enabling this by default. This comment is now resolved, but I see no change.

pkg/metrics/metrics.go
```diff
@@ -212,6 +212,10 @@ var (
 	// +metricsdoc:labels=cluster_queue="the name of the ClusterQueue",replica_role="one of `leader`, `follower`, or `standalone`"
 	AdmittedActiveWorkloads *prometheus.GaugeVec
 
+	// +metricsdoc:group=clusterqueue
+	// +metricsdoc:labels=cluster_queue="the name of the ClusterQueue",replica_role="one of `leader`, `follower`, or `standalone`",flavor="the resource flavor name"
```
kshalot:
```suggestion
	// +metricsdoc:labels=cluster_queue="the name of the ClusterQueue",flavor="the resource flavor name",replica_role="one of `leader`, `follower`, or `standalone`"
```

Nit: This order makes more sense to me

pkg/metrics/metrics.go
```diff
@@ -706,6 +710,17 @@ The label 'reason' can have the following values:
 	)
 	trackGaugeVec(AdmittedActiveWorkloads, gaugeCleanupScopeClusterQueueCache)
 
+	// +metricsdoc:group=clusterqueue
+	// +metricsdoc:labels=cluster_queue="the name of the ClusterQueue",replica_role="one of `leader`, `follower`, or `standalone`",flavor="the resource flavor name"
```
kshalot: Same comment about ordering.

pkg/metrics/metrics.go
```diff
@@ -212,6 +212,10 @@ var (
 	// +metricsdoc:labels=cluster_queue="the name of the ClusterQueue",replica_role="one of `leader`, `follower`, or `standalone`"
 	AdmittedActiveWorkloads *prometheus.GaugeVec
 
+	// +metricsdoc:group=clusterqueue
+	// +metricsdoc:labels=cluster_queue="the name of the ClusterQueue",replica_role="one of `leader`, `follower`, or `standalone`",flavor="the resource flavor name"
+	FlavorAdmittedActiveWorkloads *prometheus.GaugeVec
```
kshalot: Maybe `AdmittedActiveWorkloadsByFlavor`? When you search for the metrics in some dropdown (if someone has a UI like this or smth), both this new metric and the existing `AdmittedActiveWorkloads` will be listed when you search by prefix.

We even call the FG itself `AdmittedWorkloadsCountByFlavor`

pkg/cache/scheduler/clusterqueue.go
olekzabl: This was deliberate, for consistency with `updateFlavorUsage` above.

pkg/metrics/metrics.go
olekzabl: Yes, I like it, not just for alpha-sorting reasons.

Initially we followed `Flavor...` to stick to the existing `LocalQueue...` names - but I see `...ByStatus` already exists, and I think for the pending workload case we'll need `...By...` anyway. (Otherwise, the name would be too entangled).

pkg/features/kube_features.go
olekzabl: "When enabled" also happens. It's Polina's wording; I didn't care.

As you brought my attention to this, I picked some yet another wording.

pkg/features/kube_features.go
olekzabl: Oh. I did that... but somehow reverted it later. No idea how. Done again.

BTW same about other neighboring changes (owner, KEP -> issue) :)

kshalot (commented):
/lgtm

olekzabl:
/test all

As the last run did not produce explicit failures; rather (infra?) timeouts.

olekzabl:
@mimowo or @tenzen-y could you take a look? This is already lgtm-ed.

pkg/metrics/metrics.go
```diff
@@ -706,6 +710,17 @@ The label 'reason' can have the following values:
...
 	trackGaugeVec(AdmittedActiveWorkloads, gaugeCleanupScopeClusterQueueCache)
 
+	// +metricsdoc:group=clusterqueue
+	// +metricsdoc:labels=cluster_queue="the name of the ClusterQueue",flavor="the resource flavor name",replica_role="one of `leader`, `follower`, or `standalone`"
+	AdmittedActiveWorkloadsByFlavor = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: constants.KueueName,
+			Name:      "admitted_active_workloads_by_flavor",
+			Help:      "The number of admitted Workloads that are active, per 'cluster_queue' and 'flavor' (a single Workload is accounted for every flavor it uses)",
+		}, append([]string{"cluster_queue", "flavor", "replica_role"}, extraLabels...),
```
mimowo, line 720: Have you considered extending the pre-xexisting metric with a new label, if so, any reason not to do it?

Asking because it was already also requested to break down by priority in another issue. Creating metrics _by_flavor, by_priority, by_flavor_and_priiority etc. will lead to many metrics.

olekzabl: The reason not to do it is that a single Workload may be admitted in multiple flavors.
(Because of having multiple PodSets, or multiple ResourceGroups).

Then, if we break down the existing `admitted_active_workloads` metric by flavor, any workload with N flavors would be counted (at least) N times, thus skewing upwards the total (= when reading the metric without specifying the "flavor" label).

I consider this a breaking change, and the way to avoid it is to introduce a new metric.

olekzabl: The "by priority" story is *not* analogous because a workload has a single priority value.

olekzabl:
Update for clarity:

* AFAIK this PR _essentially mergeable_ (of course up to rebasing).
  Not only in my eyes but also e.g. in @mimowo 's.

* Then why haven't we?
  The reason is that we're now envisioning a more general approach for breaking various metrics (not just this one) by custom-specified workload labels, in line with [KEP-7066](https://github.com/kubernetes-sigs/kueue/tree/main/keps/7066-custom-metric-labels).
  _[EDIT: tracking issue for this: #12629]_

  Advantages of that new approach:
  * a more systematic approach to various workload properties, no artificial restricting to flavors
  * no confusion in case when a workload is admitted into multiple flavors (due to multiple PodSets or ResourceGroups) - it would be still emitted once, controlled by the customer-defined label value
  
  Disadvantages:
  * not expressive enough to handle multiple flavors, in case the user really wants that
  * even if flavors:workloads is 1:1 mapping, the user will need to label their workloads (or jobs) manually.

---

Summing up: we're withholding merging it right now, with the aim of reaching a nicer code state with some more effort.

However, if you need exactly this PR merged - without the custom labels system - please let us know. We'll reconsider.

mimowo:
/close
To avoid distractions. We can re-open if needed. For now we believe this is replaced by the custom labels extension for workloads.

Issue #11835, opened 2026-05-29
kueue_cohort_borrowing_limit metric for cohort and fix the metric for CQ


mimowo:
**What would you like to be added**:

kueue_cohort_borrowing_limit metric for cohort and fix the metric for CQ

We have the inflight PR: https://github.com/kubernetes-sigs/kueue/pull/9347

However, during review it turned out the metric for CQ has some critical issue. I'm not sure we want to replicate the same issue for cohorts: when users don't set borrotingLimit their borrowing is **unlimited**, but the metric shows 0. We need to fix this I think.

**Why is this needed**:

To provide the observability for cohorts, requested originally in https://github.com/kubernetes-sigs/kueue/issues/7539

mimowo:
/assign @mszadkow 
who is already working on the inflight PR.

PR #11894, opened 2026-06-02
[Fix] Align ClusterQueue metric with the borrowing limit behavior

mszadkow:
#### What type of PR is this?
/kind bug

ClusterQueue treated `nil` as `0` by defaulting the float and only filling it when the field was non-nil.
Thus `kueue_cluster_queue_borrowing_limit` misreported “unlimited” as `0`.

Relates to #11835 

```release-note
Observability: Fix ClusterQueue Borrowing Limit metric to display infinity if the limit is unset.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed cluster queue borrowing limit metrics to correctly report unbounded capacity (infinity) when not explicitly configured.

* **Documentation**
  * Updated metric documentation to clarify borrowing limit behavior when unset.

* **Tests**
  * Improved quota reporting test coverage with table-driven test approach.


pkg/cache/scheduler/clusterqueue.go
```diff
@@ -542,7 +542,9 @@ func (c *clusterQueue) reportResourceMetrics(fairSharingEnabled bool) {
 		fName, rName := string(fr.Flavor), string(fr.Resource)
 		nominal := resourceFloat(fr.Resource, quota.Nominal.Int64())
 		var borrowing, lending float64
-		if quota.BorrowingLimit != nil {
+		if quota.BorrowingLimit == nil || quota.BorrowingLimit.Equal(resources.Unlimited) {
+			borrowing = math.Inf(1)
```
mimowo, line 546: Did you test how this behaves ootb in Grafana? Also, it would be great if you compare the behavior with checking if we set N/A for the value.

mszadkow: I will come back with the results of comparison

mszadkow:
CQs setup:
```
kind: ClusterQueue
metadata:
  name: "cluster-queue-a"
spec:
  cohortName: "default-cohort"
  namespaceSelector: {} # match all.
  resourceGroups:
  - coveredResources: ["cpu", "memory"]
    flavors:
    - name: "default-flavor"
      resources:
      - name: "cpu"
        nominalQuota: 9
      - name: "memory"
        nominalQuota: 36Gi
        borrowingLimit: 12Gi
---
kind: ClusterQueue
metadata:
  name: "cluster-queue-b"
spec:
  cohortName: "default-cohort"
  namespaceSelector: {} # match all.
  resourceGroups:
  - coveredResources: ["cpu", "memory"]
    flavors:
    - name: "default-flavor"
      resources:
      - name: "cpu"
        nominalQuota: 9
      - name: "memory"
        nominalQuota: 36Gi
        borrowingLimit: 0
---
kind: ClusterQueue
metadata:
  name: "cluster-queue-c"
spec:
  cohortName: "default-cohort"
  namespaceSelector: {} # match all.
  resourceGroups:
  - coveredResources: ["cpu", "memory"]
    flavors:
    - name: "default-flavor"
      resources:
      - name: "cpu"
        nominalQuota: 9
      - name: "memory"
        nominalQuota: 36Gi
```

Prometheus entries:

Without the PR:
```
kueue_cluster_queue_borrowing_limit{cluster_queue="cluster-queue-a", cohort="default-cohort", replica_role="leader", resource="memory", service="kueue-controller-manager-metrics-service"} | 12884901888
kueue_cluster_queue_borrowing_limit{cluster_queue="cluster-queue-b", cohort="default-cohort",  replica_role="leader", resource="memory", service="kueue-controller-manager-metrics-service"} | 0
kueue_cluster_queue_borrowing_limit{cluster_queue="cluster-queue-c", cohort="default-cohort", replica_role="leader", resource="memory", service="kueue-controller-manager-metrics-service"} | 0
```

With the PR:
```
kueue_cluster_queue_borrowing_limit{cluster_queue="cluster-queue-a", cohort="default-cohort", replica_role="leader", resource="memory", service="kueue-controller-manager-metrics-service"} | 12884901888
kueue_cluster_queue_borrowing_limit{cluster_queue="cluster-queue-b", cohort="default-cohort", replica_role="leader", resource="memory", service="kueue-controller-manager-metrics-service"} | 0
kueue_cluster_queue_borrowing_limit{cluster_queue="cluster-queue-c", cohort="default-cohort", replica_role="leader", resource="memory", service="kueue-controller-manager-metrics-service"} | +Inf
```

mimowo:
Yeah, what I'm also interested is:
- how does Grafana handle that
- how would it be handled by prometheus and Grafana if we modified the PR to say NaN rather than +inf

mszadkow:
Well Grafana struggle to interpret that +inf
image


Will check NaN now

mszadkow:
image

Pretty much the same

mimowo:
So, getting NaN or +inf for one curve makes other curves not displayed?

mszadkow:
> So, getting NaN or +inf for one curve makes other curves not displayed?

No, they are normally displayed, I have selected the one with Inf/NaN to show how it's displayed.

mimowo:
Ah, I see, but that is not very relevant I think, and kind of obvious, the relevant checks are:
1. how will it display if you don't filter - attempt to show all in the same graph
2. attempt to use some aggregation like "sum"

mszadkow:
No filter:
image
Both Nan and Inf are simply not displayed

And sum over cohort
image
Well this doesn't look good

mimowo:
> Both Nan and Inf are simply not displayed

This lgtm

> Well this doesn't look good

That is probably ok, I doubt anyone will to the sum actually, mostly ask the question out of curiosity

pkg/cache/scheduler/clusterqueue.go
mimowo: I see, thank you for the display charts. I think they look good - returning Nan or +inf does not mess with other curves. That lgtm

pkg/cache/scheduler/clusterqueue.go
```diff
@@ -542,7 +542,9 @@ func (c *clusterQueue) reportResourceMetrics(fairSharingEnabled bool) {
 		fName, rName := string(fr.Flavor), string(fr.Resource)
 		nominal := resourceFloat(fr.Resource, quota.Nominal.Int64())
 		var borrowing, lending float64
-		if quota.BorrowingLimit != nil {
+		if quota.BorrowingLimit == nil || quota.BorrowingLimit.Equal(resources.Unlimited) {
```
mimowo: Do we need that check actually "quota.BorrowingLimit.Equal(resources.Unlimited)"? If it is "unlimited", then it would anyway result in maxInt, so the same output, no?

mszadkow: I think you are right, this would be extra check, although we can end up with Inf and MaxInt which are not the same in Prometheus and Grafana

mimowo: I see, I think this is fine, I cannot imagine anyway really setting maxInt. Also, it wasn't the intention of the issue to change that.

mimowo: "I see, I think this is fine"  - meant that I'm fine for not addressing the handling when someone sets borrowingLimit=MaxInt64. 

TBH, I still don't understand what breaks in this case, maybe this is a separate but, but I wouldn't couple it here.

So, I propose we norrow down the PR to the known issue, and cherrypick. As a follow up please summarize your finding about the buggy handling of BorrowingLimit=MaxInt64 (if this is the case), and open a follow up issue for that.

mimowo:
@mszadkow could you see if we also have the analogous issue for the lending_limit metric? If so, then we should open a follow up Issue or PR.

pkg/cache/scheduler/clusterqueue.go
mszadkow: ok, I have changed that

mimowo:
/lgtm
/approve
Thank you for the metrix fix 👍 Let's also check if we have the same issue for the lending_limit metric. In any case this should also unblock the work on the new metric for cohorts.

mimowo:
@mszadkow please prep CP for 0.17

PR #11895, opened 2026-06-02
Create variants function in ConcurrentAdmission controller accepts fl…

reruno:
…avors slice instead of flavors map

#### What type of PR is this?
/kind cleanup

Create variants function in ConcurrentAdmission controller accepts flavors slice instead of flavors map

Related to #10920 

keps/8691-concurrent-admission/README.md

|The `VariantsController` leverages the existing implementation of scheduling logic in Kueue, where Workloads are sorted
by priorities and creation timestamps and then picked-up by the scheduler one per scheduling cycle (per ClusterQueue).
This means when creating Variants the controller should ensure more favorable Variants are ahead of less favorable ones in the queue.

origin createVariants goes through map with for range, which is non deterministic. I replaced it with slice where `0` idx is high priority, and `len - 1` is lowest priority, with this change we create highest priority first all the time 

```release-note
NONE
```

### DIFF
--- a/pkg/controller/concurrentadmission/controller.go
+++ b/pkg/controller/concurrentadmission/controller.go
@@ -143,7 +143,7 @@ func (r *variantReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 
 	if len(variants) < len(flavorOrder) {
 		log.V(3).Info("Too few variants, creating new ones", "desired", len(flavorOrder), "actual", len(variants))
-		if err := r.createVariants(ctx, parent, variants, flavorOrder); err != nil {
+		if err := r.createVariants(ctx, parent, variants, cq.Spec.ResourceGroups[0].Flavors); err != nil {
 			return ctrl.Result{}, err
 		}
 	}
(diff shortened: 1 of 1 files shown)


pkg/controller/concurrentadmission/controller.go
```diff
@@ -230,9 +230,10 @@ func (r *variantReconciler) getClusterQueue(ctx context.Context, wl *kueue.Workl
 	return cq, nil
 }
 
-func (r *variantReconciler) createVariants(ctx context.Context, parent *kueue.Workload, variants []kueue.Workload, resourceFlavors map[kueue.ResourceFlavorReference]int) error {
+func (r *variantReconciler) createVariants(ctx context.Context, parent *kueue.Workload, variants []kueue.Workload, resourceFlavors []kueue.FlavorQuotas) error {
 	log := ctrl.LoggerFrom(ctx)
-	for flavor := range resourceFlavors {
+	for _, flavorQuota := range resourceFlavors {
+		flavor := flavorQuota.Name
```
mszadkow: and why is this needed?

reruno: origin createVariants goes through map with for range, which is non deterministic. I replaced it with slice where 0 idx is high priority, and len - 1 is lowest priority, with this change we create highest priority first all the time

mszadkow: make sense, thank you

mbobrovskyi:
Thank you!

/lgtm

pkg/controller/concurrentadmission/controller.go
```diff
@@ -230,9 +230,10 @@ func (r *variantReconciler) getClusterQueue(ctx context.Context, wl *kueue.Workl
...
 		if r.hasVariantWithFlavor(ctx, variants, flavor) {
```
tenzen-y:
```suggestion
if r.hasVariantWithFlavor(ctx, variants, flavorQuota.Name) {
```

nit: I think that we don't need to remap from `flavorQuota.Name` to `flavor` new variable.

tenzen-y (commented):
Thank you for your contribution.
I left a small nit comment.

tenzen-y (commented):
Thank you!
/lgtm
/approve

PR #11920, opened 2026-06-03
Optimize test-e2e-extended taget and test-tas-e2e-extended

mbobrovskyi:
#### What type of PR is this?
/kind cleanup

Optimize test-e2e-extended taget and test-tas-e2e-extended

Fixes #

```release-note
NONE
```

### DIFF
--- a/Makefile-test.mk
+++ b/Makefile-test.mk
@@ -175,40 +175,38 @@ test-multikueue-e2e-sequential-shard-1: test-multikueue-e2e-sequential
 test-multikueue-e2e-helm: E2E_USE_HELM=true
 test-multikueue-e2e-helm: test-multikueue-e2e
 
+# Assign Shard 0 variables to all parent suites AND all shard-0 targets
+TEST_E2E_SHARD_0_TARGETS := test-e2e-extended test-e2e-extended-shard-0 test-tas-e2e-extended test-tas-e2e-extended-shard-0
+$(TEST_E2E_SHARD_0_TARGETS): export KUBERAY_VERSION := $(KUBERAY_VERSION)
+$(TEST_E2E_SHARD_0_TARGETS): export RAY_VERSION := $(RAY_VERSION)
+$(TEST_E2E_SHARD_0_TARGETS): export RAYMINI_VERSION := $(RAYMINI_VERSION)
+
+# Assign Shard 1 variables to all parent suites AND all shard-1 targets
+TEST_E2E_SHARD_1_TARGETS := test-e2e-extended test-e2e-extended-shard-1 test-tas-e2e-extended test-tas-e2e-extended-shard-1
+$(TEST_E2E_SHARD_1_TARGETS): export JOBSET_VERSION := $(JOBSET_VERSION)
+$(TEST_E2E_SHARD_1_TARGETS): export LEADERWORKERSET_VERSION := $(LEADERWORKERSET_VERSION)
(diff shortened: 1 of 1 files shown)


Makefile-test.mk
```diff
@@ -181,7 +181,8 @@ test-multikueue-e2e-helm: test-multikueue-e2e
 ## Examples:
 ##   Run only AppWrapper tests: GINKGO_ARGS="--label-filter=feature:appwrapper" make test-e2e-extended
 .PHONY: test-e2e-extended
-test-e2e-extended: test-e2e-extended-shard-0 test-e2e-extended-shard-1
+test-e2e-extended: E2E_NPROCS := 4
+test-e2e-extended: setup-e2e-env kind-ray-project-mini-image-build run-test-e2e-extended-$(E2E_KIND_VERSION:kindest/node:v%=%)
```
mbobrovskyi: @tenzen-y WDYT? or should we delete this target?

tenzen-y: Sorry, why do we need to revert this change?

mbobrovskyi: In that case it will run faster. Do you think we should keep as it is ?

ivnovakov: I agree that on local machine it's more convenient run it on single cluster.
But we also need to export versions for integrations in this target.

mbobrovskyi: Just for context, we currently have two test targets:

https://github.com/kubernetes-sigs/kueue/blob/57d32942bb83f751880a78da8916f20aaf62d284/Makefile-test.mk#L186-L212

We kept this target to verify that everything worked before updating test-infra (https://github.com/kubernetes/test-infra/pull/37178). After the test-infra update, this target is no longer very useful. However, it might still be helpful for contributors.

tenzen-y: I imagine that developers specify the test case where they want to verify.
So, what is the primary use case for keeping this target?

mbobrovskyi: > I imagine that developers specify the test case where they want to verify.
So, what is the primary use case for keeping this target?

Agree. Let's remove this target.

tenzen-y: Sounds good to me.

mbobrovskyi: Done

mbobrovskyi: /hold

Actually, I tried running tests with sharding, and it doesn't work locally when I run tests individually in dev/debug mode. To make it work, we would need to propagate the shard label every time. So I think we should keep the non-sharded target for development.

@tenzen-y @ivange94 WDYT?

mbobrovskyi: Let's keep it for now. I think it makes sense to have this target for development.

tenzen-y: > Actually, I tried running tests with sharding, and it doesn't work locally when I run tests individually in dev/debug mode. To make it work, we would need to propagate the shard label every time. So I think we should keep the non-sharded target for development.
> 
> 

Thank you for checking that in your local. In that case, let me merge this for now, then can you enhance our tests in a follow up PR?

/lgtm
/approve

tenzen-y: As we discussed with @mbobrovskyi offline, my proposal doesn't improve developer experience because they have to identify shard number and specify it when they perform make command if we apply my proposal to test sysytem.

So, I decline my previous proposal. @mbobrovskyi Thank you for making the developer experience better!

PR #11927, opened 2026-06-04
Fix stale DRA resource requests on requeue after DeviceClass deletion

sohankunkerkar:
#### What type of PR is this?
/kind bug

Fixes #

When a workload is requeued after a DeviceClass deletion, Info.Update keeps the old DRA-translated `TotalRequests` even though the resource is no longer DRA-backed. Rebuild TotalRequests from the raw spec in that case.

```release-note
DRA: Fix workloads retaining DRA-mapped resource names after their DeviceClass is deleted.
```

## Summary by CodeRabbit

* **Refactor**
  * Improved workload update flow: updates to total-request recomputation now support a conditional preserve mode to retain preprocessed DRA totals when requeuing workloads.

* **Tests**
  * Added unit tests verifying rebuild vs. preservation of total resource requests.
  * Updated integration test setup to use a shared DRA-backed resource cache for consistent reconciliation.

### DIFF
--- a/pkg/cache/queue/manager.go
+++ b/pkg/cache/queue/manager.go
@@ -693,7 +693,11 @@ func (m *Manager) RequeueWorkload(ctx context.Context, info *workload.Info, reas
 	}
 	log := ctrl.LoggerFrom(ctx)
 	workload.AdjustResources(ctx, m.client, &w)
-	info.Update(log, &w)
+	if dra.NeedsDRAReconcile(&w, m.draBackedResources) {
+		info.Update(log, &w, workload.WithPreserveTotalRequests())
+	} else {
+		info.Update(log, &w, m.workloadInfoOptions...)
+	}
 	m.addWorkload(info, q)
 
 	cq := m.hm.ClusterQueue(q.ClusterQueue)
--- a/pkg/workload/workload.go
+++ b/pkg/workload/workload.go
@@ -127,6 +127,7 @@ type dra struct {
 type InfoOptions struct {
 	excludedResourcePrefixes []string
 	resourceTransformations  map[corev1.ResourceName]*config.ResourceTransformation
+	preserveTotalRequests    bool
 	dra
 }
 
@@ -145,6 +146,15 @@ func WithExcludedResourcePrefixes(n []string) InfoOption {
 func WithResourceTransformations(transforms []config.ResourceTransformation) InfoOption {
 	return func(o *InfoOptions) {
 		o.resourceTransformations = utilslices.ToRefMap(transforms, func(e *config.ResourceTransformation) corev1.ResourceName { return e.Input })
+	}
+}
+
+// WithPreserveTotalRequests prevents Update from rebuilding TotalRequests.
+// Used when requeuing DRA-backed workloads whose TotalRequests were
+// preprocessed by the workload controller and must survive the requeue.
+func WithPreserveTotalRequests() InfoOption {
+	return func(o *InfoOptions) {
+		o.preserveTotalRequests = true
 	}
 }
 
@@ -289,34 +299,47 @@ func (p *PodSetResources) ScaledTo(newCount int32) *PodSetResources {
 }
 
 func NewInfo(w *kueue.Workload, opts ...InfoOption) *Info {
-	options := defaultOptions
-	for _, opt := range opts {
-		opt(&options)
-	}
-	info := &Info{
-		Obj: w,
-	}
-	if w.Status.Admission != nil {
-		info.ClusterQueue = w.Status.Admission.ClusterQueue
-		info.TotalRequests = totalRequestsFromAdmission(w)
-	} else {
-		info.TotalRequests = totalRequestsFromPodSets(w, &options)
-	}
+	info := &Info{}
+	info.Update(klog.Background(), w, opts...)
 	return info
 }
 
 // UpdateSchedulingHash computes and sets the scheduling hash using the
-// provided contextual logger. Call this after NewInfo in production code.
+// provided contextual logger. Called internally by Update.
 func (i *Info) UpdateSchedulingHash(log logr.Logger) {
 	i.SchedulingHash = computeSchedulingHash(log, i.Obj, i.TotalRequests)
 }
 
-// Update refreshes the object reference and recomputes the scheduling hash
-// to reflect any changes (e.g., priority updates).
-func (i *Info) Update(log logr.Logger, wl *kueue.Workload) {
-	log.V(5).Info("Workload info updated", "workload", klog.KObj(wl))
+// Update refreshes the object reference, rebuilds TotalRequests, and
+// recomputes the scheduling hash. Pass WithPreserveTotalRequests to skip
+// the TotalRequests rebuild (e.g., to retain DRA preprocessing on requeue).
+func (i *Info) Update(log logr.Logger, wl *kueue.Workload, opts ...InfoOption) {
 	i.Obj = wl
+	i.rebuildTotalRequests(opts...)
 	i.UpdateSchedulingHash(log)
+}
+
+// rebuildTotalRequests refreshes ClusterQueue and recomputes TotalRequests
+// from the current workload state. When WithPreserveTotalRequests is set,
+// only TotalRequests recomputation is skipped.
+func (i *Info) rebuildTotalRequests(opts ...InfoOption) {
+	options := defaultOptions
+	for _, opt := range opts {
+		opt(&options)
+	}
+	admitted := i.Obj.Status.Admission != nil
+	if admitted {
+		i.ClusterQueue = i.Obj.Status.Admission.ClusterQueue
+	} else {
+		i.ClusterQueue = ""
+	}
+	if !options.preserveTotalRequests {
+		if admitted {
+			i.TotalRequests = totalRequestsFromAdmission(i.Obj)
+		} else {
+			i.TotalRequests = totalRequestsFromPodSets(i.Obj, &options)
+		}
+	}
 }
 
 // computeSchedulingHash returns a deterministic hash of the workload's
--- a/pkg/workload/workload_test.go
+++ b/pkg/workload/workload_test.go
@@ -621,6 +621,107 @@ func TestNewInfo(t *testing.T) {
 	}
 }
 
+func TestUpdateWithRebuild(t *testing.T) {
+	now := time.Now().Truncate(time.Second)
+	cases := map[string]struct {
+		initial        *kueue.Workload
+		initialOptions []InfoOption
+		updated        *kueue.Workload
+		updateOptions  []InfoOption
+		wantRequests   []PodSetResources
+	}{
+		"pending workload with changed requests": {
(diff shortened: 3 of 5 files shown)


sohankunkerkar:
> @sohankunkerkar: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> pull-kueue-test-integration-shard-0-main	[17bc8db](https://github.com/kubernetes-sigs/kueue/commit/17bc8db1bedb4f9e713be620a2648a02c504fbf3)	[link](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/11927/pull-kueue-test-integration-shard-0-main/2062413892377645056)	true	`/test pull-kueue-test-integration-shard-0-main`
> [Full PR test history](https://prow.k8s.io/pr-history?org=kubernetes-sigs&repo=kueue&pr=11927). [Your PR dashboard](https://prow.k8s.io/pr?query=is%3Apr%20state%3Aopen%20author%3Asohankunkerkar). Please help us cut down on flakes by [linking to](https://git.k8s.io/community/contributors/devel/sig-testing/flaky-tests.md#filing-issues-for-flaky-tests) an [open issue](https://github.com/kubernetes-sigs/kueue/issues?q=is:issue+is:open) when you hit one in your PR.
> 
> Details

Need https://github.com/kubernetes-sigs/kueue/pull/11943 to fix this flake

sohankunkerkar:
@coderabbitai help

sohankunkerkar:
@coderabbitai review

sohankunkerkar:
@kannon92 PTAL

pkg/dra/extended_resources.go
```diff
@@ -44,7 +44,7 @@ func NeedsDRAReconcile(wl *kueue.Workload, erCache *ExtendedResourceCache) bool
 	if workload.HasDRA(wl) {
 		return true
 	}
-	if !features.Enabled(features.KueueDRAIntegrationExtendedResource) {
+	if !features.Enabled(features.KueueDRAIntegrationExtendedResource) || erCache == nil {
```
mimowo: Could erCache be even nil on prod? It seems like code just for testing, no?

pkg/cache/queue/manager.go
```diff
@@ -693,7 +693,11 @@ func (m *Manager) RequeueWorkload(ctx context.Context, info *workload.Info, reas
 	}
 	log := ctrl.LoggerFrom(ctx)
 	workload.AdjustResources(ctx, m.client, &w)
-	info.Update(log, &w)
+	if dra.NeedsDRAReconcile(&w, m.draBackedResources) {
+		info.Update(log, &w)
+	} else {
+		info.UpdateAndRebuildTotalRequests(log, &w, m.workloadInfoOptions...)
```
mimowo: This is a little bit tricky because I don't see actually Update being called from UpdateAndRebuildTotalRequests. 

Also, I'm wondering if could be better to move down the `NeedsDRAReconcile` check so that the manager.go does no need to worry about the intricacies of DRA too much.

mimowo: Or maybe we could simplify and just call UpdateAndRebuildTotalRequests - TBD

pkg/dra/extended_resources.go
sohankunkerkar: You're right. The prod always sets it. The nil was from the test not sharing the cache, which the suite_test.go fix addresses. Happy to drop the nil guard so misconfigured tests fail loudly instead of silently returning false.

mimowo: Yeah, let's drop and adjust tests. I would rather panic in the code than keep such dangerous skips which may later cause failures when we forget to set it from some code place.

pkg/cache/queue/manager.go
sohankunkerkar: `UpdateAndRebuildTotalRequests` reimplements Update instead of calling it. I'll refactor to have it call Update then rebuildTotalRequests to avoid the duplication.
  
The manager already has DRA awareness via `AddLocalQueue`. This follows the same pattern. Happy to refactor DRA out of the manager as a follow-up. On the simplification: always rebuilding would lose DRA-translated TotalRequests since that preprocessing isn't in the manager-level options.

pkg/cache/queue/manager.go
```diff
@@ -693,7 +693,11 @@ func (m *Manager) RequeueWorkload(ctx context.Context, info *workload.Info, reas
...
+	}
```
mimowo, line 700: Could we drop that "if" we just call `info.rebuildTotalRequests(opts...)` from "Update"?

Then we could call Update from inside NewInfo. wdyt?

pkg/workload/workload.go
```diff
@@ -289,19 +289,10 @@ func (p *PodSetResources) ScaledTo(newCount int32) *PodSetResources {
 }
 
 func NewInfo(w *kueue.Workload, opts ...InfoOption) *Info {
```
tenzen-y, line 301: Instead of adding new `rebuildTotalRequests` function, can we add new InfoOption for request update?

sohankunkerkar: Would you prefer `WithRebuildTotalRequests()` (opt-in rebuild, preserve by default) or `WithPreserveTotalRequests()` (opt-out, rebuild by default)? Both eliminate rebuildTotalRequests as a separate path. Just need to know which direction you and @mimowo prefer.

mimowo: `WithRebuildTotalRequests` sounds reasonable, but I still don't understand why we need that conditional path at all. Is it for performance, because we assume `rebuildTotalRequests` is heavy or correctness?

Also asked another comment trying to understand if we can eliminate the conditional logic here: https://github.com/kubernetes-sigs/kueue/pull/11927#discussion_r3389741723

mimowo: If this is a performance optimization, and we think that `rebuildTotalRequests` is heavy, then maybe I would call it `WithSafeSkipForRebuildTotalRequests` - that will clearly indicate we just skip the rebuild, because we consider it "safe"

sohankunkerkar: Correctness, not performance. When the workload is still DRA-backed, rebuildTotalRequests would recompute from the raw spec (nvidia.com/gpu: 1), dropping the controller-preprocessed DRA translation (gpu: 1). The manager-level `workloadInfoOptions` don't carry per-workload DRA state, so rebuild can't re-derive it. I've switched to rebuild-by-default in Update() and now call Update() from NewInfo() per the earlier suggestion. The DRA path uses `WithPreserveTotalRequests()` as the explicit opt-out.

sohankunkerkar: I think opt-out is safer here: if a future caller forgets the opt, they get a rebuild (correct), not stale data.

pkg/workload/workload.go
```diff
@@ -289,36 +299,45 @@ func (p *PodSetResources) ScaledTo(newCount int32) *PodSetResources {
...
+	if options.preserveTotalRequests {
+		return
+	}
+	if i.Obj.Status.Admission != nil {
+		i.ClusterQueue = i.Obj.Status.Admission.ClusterQueue
+		i.TotalRequests = totalRequestsFromAdmission(i.Obj)
+	} else {
+		i.ClusterQueue = ""
+		i.TotalRequests = totalRequestsFromPodSets(i.Obj, &options)
+	}
```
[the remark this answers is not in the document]
sohankunkerkar: I agree with this.

pkg/workload/workload_test.go
```diff
@@ -621,6 +621,107 @@ func TestNewInfo(t *testing.T) {
...
+			},
+			updated: utiltestingapi.MakeWorkload("wl", "ns").
+				Request("example.com/gpu", "1").Obj(),
+			updateOptions: []InfoOption{WithPreserveTotalRequests()},
+			wantRequests: []PodSetResources{{
+				Name:     kueue.DefaultPodSetName,
+				Requests: resources.Requests{"gpu": 1},
+				Count:    1,
+			}},
+		},
```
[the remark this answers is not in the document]
sohankunkerkar: Not agree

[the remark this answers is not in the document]
kannon92: I agree with your assesment but should we have negative testing here?

ie is there any behavior that is gated under this feature gate. we should make sure to test negative case also.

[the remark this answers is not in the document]
sohankunkerkar: The preserve/rebuild logic is gate-independent. `WithPreserveTotalRequests` is pure `InfoOption` mechanics. The DRA gate controls whether `WithPreprocessedDRAResources` transforms requests, which is existing behavior unchanged by this PR.

test/integration/singlecluster/controller/dra/suite_test.go
```diff
@@ -126,7 +126,11 @@ func managerSetup(modifyConfig func(*config.Configuration)) framework.ManagerSet
 
 		cCache := schdcache.New(mgr.GetClient())
 		preemptionExpectations := preemptexpectations.New()
-		queueOptions := []qcache.Option{qcache.WithPreemptionExpectations(preemptionExpectations)}
+		draBackedResources := dra.NewExtendedResourceCache()
```
mimowo, line 129: how are the changes to the ntegration test relevant to the issue? Shouldn't we add some dedicated repro test, or the pre-existing tests can already repro when the setup is modified as here?

sohankunkerkar: This is the fix from the erCache nil [thread](https://github.com/kubernetes-sigs/kueue/pull/11927#discussion_r3380827466) that shares one `draBackedResources` between queue manager and controllers to match prod wiring, so the nil guard can be dropped.

mimowo:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/11927/pull-kueue-test-e2e-certmanager-main/2064790741418250240
Does not seem related, let me ticket as we need to stabilize the build more

Opened: https://github.com/kubernetes-sigs/kueue/issues/12091

test/integration/singlecluster/controller/dra/suite_test.go
mimowo: I see, in that case, would it be feasible to add an integration test which reproduces the issue?

sohankunkerkar: Let me add a test for the visible behavior (stale DRA cleared after DeviceClass deletion). The exact RequeueWorkload path is covered by the unit test (TestUpdateWithRebuild).

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for hardening the DRA integration 👍

mimowo:
/release-note-edit
```release-note
DRA:  Fix workloads retaining DRA-mapped resource names after their DeviceClass is deleted.
```

mimowo:
/release-note-edit
```release-note
DRA: Fix workloads retaining DRA-mapped resource names after their DeviceClass is deleted.
```

PR #11929, opened 2026-06-04
Fix pending workloads not requeued after DeviceClass deletion

sohankunkerkar:
#### What type of PR is this?
/kind bug

This was found during QE testing of the DRA Extended Resources Alpha-2 feature on an OCP 4.21 cluster with NVIDIA T4 GPUs. The test scenario: fill GPU quota with two running jobs, submit a third pending job, delete the DeviceClass. The pending workload's condition and resourceVersion remained unchanged i.e. no requeue occurred.

Fixes #

 When a DeviceClass is deleted, r`econcileWorkloads` only enqueued affected workloads for a delayed Reconcile. But the DRA cache is already cleared at that point, so Reconcile skips DRA processing and never re-adds the workload to the scheduler queue.

When the workload is no longer DRA-backed, add it directly to the scheduler queue instead of relying on the delayed Reconcile.

```release-note
DRA: Fixed a bug where pending Workloads using DRA extended resources were not requeued when their `DeviceClass` was deleted or its `extendedResourceName` changed. Kueue now re-evaluates affected Workloads so they do not remain in stale admission state.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure workloads are re-evaluated and re-enqueued when DeviceClass is deleted or its extended resource is changed, preventing stale admission state.
  * Fix handling so pending workloads are correctly reassessed and marked not-admitted when reserved quota/resource names no longer apply.

* **Tests**
  * Added integration tests covering DeviceClass-driven requeue and admission behavior for extended-resource changes and deletions.

### DIFF
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -1742,7 +1742,16 @@ func (h *deviceClassHandler) reconcileWorkloads(ctx context.Context, q workqueue
 			log.Error(err, "Could not list workloads for extended resource", "resource", name)
 			continue
 		}
-		for _, w := range lst.Items {
+		for i := range lst.Items {
+			w := &lst.Items[i]
+			log.V(3).Info("Requeuing workload due to DeviceClass change", "workload", klog.KObj(w), "resource", name)
+			if !dra.NeedsDRAReconcile(w, h.r.draBackedResources) && workload.IsAdmissible(w) {
+				wlCopy := w.DeepCopy()
+				workload.AdjustResources(ctx, h.r.client, wlCopy)
+				if err := h.r.queues.AddOrUpdateWorkload(log, wlCopy); err != nil {
+					log.Error(err, "Failed to re-add workload to queue after DeviceClass change")
+				}
+			}
 			q.AddAfter(reconcile.Request{
 				NamespacedName: types.NamespacedName{
 					Name:      w.Name,
--- a/test/integration/singlecluster/controller/dra/dra_test.go
+++ b/test/integration/singlecluster/controller/dra/dra_test.go
@@ -23,6 +23,7 @@ import (
 	"github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	resourcev1 "k8s.io/api/resource/v1"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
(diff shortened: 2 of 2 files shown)


sohankunkerkar:
@kannon92 PTAL

pkg/controller/core/workload_controller.go
```diff
@@ -1742,7 +1742,20 @@ func (h *deviceClassHandler) reconcileWorkloads(ctx context.Context, q workqueue
...
-		for _, w := range lst.Items {
+		for i := range lst.Items {
+			w := &lst.Items[i]
+			log.V(3).Info("Requeuing workload due to DeviceClass change", "workload", klog.KObj(w), "resource", name)
+			if !dra.NeedsDRAReconcile(w, h.r.draBackedResources) {
+				wlCopy := w.DeepCopy()
+				workload.AdjustResources(ctx, h.r.client, wlCopy)
+				if workload.IsAdmissible(wlCopy) {
+					if err := h.r.queues.AddOrUpdateWorkload(log, wlCopy); err != nil {
+						log.V(2).Info("Failed to re-add workload to queue after DeviceClass change", "error", err)
```
mimowo:
```suggestion
						log.Error(err, "Failed to re-add workload to queue after DeviceClass change")
```

pkg/controller/core/workload_controller.go
```diff
@@ -1742,7 +1742,20 @@ func (h *deviceClassHandler) reconcileWorkloads(ctx context.Context, q workqueue
...
+					}
+				} else {
+					h.r.cache.AddOrUpdateWorkload(log, wlCopy)
```
mimowo: hmm, but this seems like a bug, what if the workload is not admissible, but also not admitted, then we shouldn't add it into the cache, right?

mimowo: If this is the case let's make sure we have an integration test for that.

sohankunkerkar: Good catch! I think the else part is effectively a dead code.

pkg/controller/core/workload_controller.go
```diff
@@ -1742,7 +1742,18 @@ func (h *deviceClassHandler) reconcileWorkloads(ctx context.Context, q workqueue
 			log.Error(err, "Could not list workloads for extended resource", "resource", name)
 			continue
 		}
-		for _, w := range lst.Items {
+		for i := range lst.Items {
+			w := &lst.Items[i]
+			log.V(3).Info("Requeuing workload due to DeviceClass change", "workload", klog.KObj(w), "resource", name)
+			if !dra.NeedsDRAReconcile(w, h.r.draBackedResources) {
+				wlCopy := w.DeepCopy()
+				workload.AdjustResources(ctx, h.r.client, wlCopy)
```
mimowo, line 1750: nit, but I think we can do these two lines only under the IsAdmissible check, right?

sohankunkerkar: Agree! I addressed that in the subsequent push.

mimowo:
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for hardening the DRA integration 👍

mimowo:
@sohankunkerkar please prep the CP for 0.17 if feasible

sohankunkerkar:
@mimowo I don't think we need this code for 0.17. We added https://github.com/kubernetes-sigs/kueue/pull/10973 in 0.18

mimowo:
sgtm

mimowo:
/release-note-edit
```release-note
DRA: Fixed a bug where pending Workloads using DRA extended resources were not requeued when their `DeviceClass` was deleted or its `extendedResourceName` changed. Kueue now re-evaluates affected Workloads so they do not remain in stale admission state.
```

PR #11943, opened 2026-06-04
Use fixed-rate requeue for DRA RCT processing errors

sohankunkerkar:
#### What type of PR is this?
/kind flake

Fixes https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/11927/pull-kueue-test-integration-shard-0-main/2062413892377645056

Replace error return with RequeueAfter to avoid exponential backoff retry storms after patching the workload as inadmissible.

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Failures when processing certain device/resources now schedule a short automatic retry instead of surfacing an immediate reconcile error, reducing transient error noise and improving stability.
* **Tests**
  * Reconciliation tests updated to expect the short requeue behavior instead of an error.

### DIFF
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -73,6 +73,10 @@ import (
 	"sigs.k8s.io/kueue/pkg/workloadslicing"
 )
 
+const (
+	draRequeueDelay = time.Second
+)
+
 var (
 	realClock = clock.RealClock{}
 )
@@ -351,7 +355,7 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 			if updateErr != nil {
 				return ctrl.Result{}, fmt.Errorf("failed to update workload status for DRA error: %w", updateErr)
 			}
-			return ctrl.Result{}, err
+			return ctrl.Result{RequeueAfter: draRequeueDelay}, nil
 		}
 
 		// Process Extended Resources backed by DRA (new path)
@@ -373,7 +377,7 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 				if updateErr != nil {
 					return ctrl.Result{}, fmt.Errorf("failed to update workload status for DRA extended resources error: %w", updateErr)
 				}
-				return ctrl.Result{}, err
+				return ctrl.Result{RequeueAfter: draRequeueDelay}, nil
 			}
 		}
 
@@ -408,7 +412,7 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 				if updateErr != nil {
 					return ctrl.Result{}, fmt.Errorf("failed to update workload status for DRA counter resources error: %w", updateErr)
 				}
-				return ctrl.Result{}, err
+				return ctrl.Result{RequeueAfter: draRequeueDelay}, nil
 			}
 			for podSetName, resources := range counterResources {
 				if existing, ok := draResources[podSetName]; ok {
--- a/pkg/controller/core/workload_controller_test.go
+++ b/pkg/controller/core/workload_controller_test.go
@@ -669,10 +669,10 @@ func TestReconcile(t *testing.T) {
 				}
 				return wl
 			}(),
-			wantErrorMsg: "not mapped in DRA configuration",
-			wantEvents:   nil,
+			wantResult: reconcile.Result{RequeueAfter: draRequeueDelay},
+			wantEvents: nil,
 		},
-		"reconcile DRA ResourceClaimTemplate not found should return error": {
+		"reconcile DRA ResourceClaimTemplate not found should requeue and mark inadmissible": {
 			featureGates: map[featuregate.Feature]bool{
 				features.KueueDRAIntegration:              true,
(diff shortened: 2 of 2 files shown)


sohankunkerkar:
cc @PannagaRao

sohankunkerkar:
@mbobrovskyi could you please take a look at it?

pkg/controller/core/workload_controller.go
```diff
@@ -351,7 +351,7 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 			if updateErr != nil {
 				return ctrl.Result{}, fmt.Errorf("failed to update workload status for DRA error: %w", updateErr)
 			}
-			return ctrl.Result{}, err
+			return ctrl.Result{RequeueAfter: time.Second}, nil
```
mbobrovskyi: Should we log error?

mbobrovskyi: And maybe add a message indicating that we will requeue in one second.

sohankunkerkar: Good point! The error is already logged at line 343 above. I kept the same pattern as the other RequeueAfter returns in this controller (lines 268, 470, 524) which don't log requeue intent either.

mbobrovskyi (commented):
Ah, I see. Thank you!

/lgtm

pkg/controller/core/workload_controller_test.go
```diff
@@ -669,8 +669,8 @@ func TestReconcile(t *testing.T) {
 				}
 				return wl
 			}(),
-			wantErrorMsg: "not mapped in DRA configuration",
-			wantEvents:   nil,
+			wantResult: reconcile.Result{RequeueAfter: time.Second},
+			wantEvents: nil,
 		},
 		"reconcile DRA ResourceClaimTemplate not found should return error": {
```
tenzen-y:
```suggestion
		"reconcile DRA ResourceClaimTemplate not found should requeue and mark inadmissible": {
```

This test case no longer raises any error.

pkg/controller/core/workload_controller.go
tenzen-y: Can you make the delay time a constant at the top of this file?

pkg/controller/core/workload_controller.go
tenzen-y: Do we need to apply the same pattern to here as well, or not?

tenzen-y: Do we need to apply the same pattern to here as well, or not?

tenzen-y (commented):
Thank you for working on this. Mostly LGTM
I left confirmations.

tenzen-y:
@sohankunkerkar will address my comments in a follow-up PR.

pkg/controller/core/workload_controller.go
sohankunkerkar: Yeah, I will update this in a follow up PR.

pkg/controller/core/workload_controller.go
```diff
@@ -351,7 +355,7 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 			if updateErr != nil {
 				return ctrl.Result{}, fmt.Errorf("failed to update workload status for DRA error: %w", updateErr)
 			}
-			return ctrl.Result{}, err
+			return ctrl.Result{RequeueAfter: draRequeueDelay}, nil
```
mimowo, line 358: Unconditional constant using generic `draRequeueDelay` delay on errors is usually a code smell, because it eliminates the controller-runtime native mechanism for exponential backoff and can result in much more retries over period of time. 

I would like to first understand what is the scenario resulting in the "error storm". If these are conflicts, then we generally handle them with controller-runtime backoff. Introducing a new policy just for DRA is confusing IMO. 

If we decide to handle some kind of errors with a longer period, like 1s, then it might be reasonable but instead of constant factor I would rather consider it as the basis in the workqueue. Such parametrization of the workqueue could be then injected to various controllers we have.

mimowo: cc @tenzen-y @sohankunkerkar I would consider this PR to revert actually as it introduces inconsistency across the codebase. If we need to keep it, there should be some commentary why we handle DRA in so special way.

sohankunkerkar: My reasoning was that after `PatchAdmissionStatus()` succeeds, some DRA cases are already converted into a persisted inadmissible state, so returning err felt like treating a completed state transition as a reconcile failure and causing repeated reprocessing of the same known condition. But looking again, these paths mix deterministic inadmissibility with genuine internal/API failures, and the latter should likely keep the normal controller-runtime backoff/error accounting. So I’m fine to revert this PR, or follow up with a narrower change that only handles the deterministic inadmissible cases differently.

mimowo: I see, yes https://github.com/kubernetes-sigs/kueue/issues/11994 is a much cleaner approach. When there error is permanent we don't need to retry the Reconcile at all, when the error is transient we can just use the default controller-runtime error handling (or parametrized, but with exponential backoff).

PR #11997, opened 2026-06-08
Add check for ElasticWorkloads in TAS topology request

yaroslava-serdiuk:
#### What type of PR is this?
AI-assisted

/kind bug

/area tas
/area elasticworkloads

Scheduler guard is not scoped to elastic jobs, but should be: keps/77-dynamically-sized-jobs/README.md:289-291

Fixes # 

```release-note
ElasticJobsViaWorkloadSlices: Fix the bug that regular (non-elastic) workloads with the required/preferred topology
were rejected when the feature ElasticJobsViaWorkloadSlicesWithTAS is enabled.
```

## Summary by CodeRabbit

* **New Features**
  * Improved handling of elastic jobs for topology-aware scheduling, gated by feature flags and the elastic-job annotation.

* **Tests**
  * Added validation tests for topology requests covering elastic and non-elastic workloads with multiple sub-cases.

* **Chores**
  * Centralized the elastic-job annotation key across the codebase and introduced a workload-level check for elastic jobs.

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -76,4 +76,6 @@ const (
 	//
 	// This annotation is alpha-level and requires the AdmissionGatedBy feature gate, disabled by default.
 	AdmissionGatedByAnnotation = "kueue.x-k8s.io/admission-gated-by"
+
+	ElasticJobAnnotation = "kueue.x-k8s.io/elastic-job"
 )
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -77,5 +77,6 @@ const (
 	// This annotation is alpha-level and requires the AdmissionGatedBy feature gate, disabled by default.
 	AdmissionGatedByAnnotation = "kueue.x-k8s.io/admission-gated-by"
 
+	// ElasticJobAnnotation is an annotation set on the Job to indicate that it is an elastic job.
 	ElasticJobAnnotation = "kueue.x-k8s.io/elastic-job"
 )
--- a/pkg/scheduler/flavorassigner/flavorassigner_test.go
+++ b/pkg/scheduler/flavorassigner/flavorassigner_test.go
@@ -27,6 +27,7 @@ import (
 	"github.com/google/go-cmp/cmp/cmpopts"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/component-base/featuregate"
 	"k8s.io/utils/ptr"
 
@@ -4108,6 +4109,11 @@ func TestWorkloadsTopologyRequests_ElasticJobsValidation(t *testing.T) {
 				}),
 			},
 			workload: *workload.NewInfo(&kueue.Workload{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						"kueue.x-k8s.io/elastic-job": "true",
+					},
+				},
 				Spec: kueue.WorkloadSpec{
 					PodSets: []kueue.PodSet{
 						*utiltestingapi.MakePodSet(kueue.DefaultPodSetName, 2).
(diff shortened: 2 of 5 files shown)


yaroslava-serdiuk:
@sohankunkerkar Could you take a look, please?

pkg/scheduler/flavorassigner/tas_flavorassigner.go
```diff
@@ -58,9 +58,14 @@ func (a *Assignment) WorkloadsTopologyRequests(log logr.Logger, wl *workload.Inf
 			}
 			isTASImplied := isTASImplied(&podSet, cq)
 
-			// Only unconstrained topology is supported with elastic workload slices.
 			var previousAssignment *kueue.TopologyAssignment
-			if features.Enabled(features.ElasticJobsViaWorkloadSlicesWithTAS) {
+
+			// Only unconstrained topology is supported with elastic workload slices.
+			if workload.IsElasticWorkload(wl.Obj) {
```
sohankunkerkar: This block now runs for elastic workloads even when `ElasticJobsViaWorkloadSlicesWithTAS` is off. On main, the gate check was the outer guard and gate off meant the entire block was skipped. Now the annotation check is the outer guard, so elastic workloads enter the block and immediately hit the "gate not enabled" error. They never get a topology assignment.

Please combine both conditions in the outer guard.
```go
if workload.IsElasticWorkload(wl.Obj) && features.Enabled(features.ElasticJobsViaWorkloadSlicesWithTAS) 
```

pkg/workload/workload.go
```diff
@@ -1977,3 +1977,12 @@ func TASAssignedNodeNames(wl *kueue.Workload) []string {
 	}
 	return nodesSet.UnsortedList()
 }
+
+// IsElasticWorkload returns true if ElasticJobsViaWorkloadSlices feature gate is enabled
```
sohankunkerkar, line 1981: Please use `workloadslicing.IsElasticWorkload` instead.

pkg/constants/constants.go
```diff
@@ -76,4 +76,6 @@ const (
 	//
 	// This annotation is alpha-level and requires the AdmissionGatedBy feature gate, disabled by default.
 	AdmissionGatedByAnnotation = "kueue.x-k8s.io/admission-gated-by"
+
+	ElasticJobAnnotation = "kueue.x-k8s.io/elastic-job"
```
sohankunkerkar, line 81: nit: please add a doc for this.

sohankunkerkar (commented):
Thanks for working on this!The fix direction is right. I have left a few comments. I think [this](https://github.com/kubernetes-sigs/kueue/pull/11997/changes#r3376451022) should address your CI failures.

pkg/workload/workload.go
yaroslava-serdiuk: This would introduce a cycle dependancy: 
flavorassignee imports workloadslicing, workloadslicing imports preemption, preemption imports flavorassignee.

pkg/scheduler/flavorassigner/tas_flavorassigner.go
yaroslava-serdiuk: Done

pkg/workload/workload.go
sohankunkerkar: ah, yeah

pkg/scheduler/flavorassigner/flavorassigner_test.go
```diff
@@ -4190,6 +4237,56 @@ func TestWorkloadsTopologyRequests_ElasticJobsValidation(t *testing.T) {
...
+					PodSets: []kueue.PodSet{
+						*utiltestingapi.MakePodSet(kueue.DefaultPodSetName, 2).
+							Request(corev1.ResourceCPU, "1").
+							RequiredTopologyRequest(corev1.LabelHostname).
+							Obj(),
+					},
+				},
+			}),
+		},
+		"preferred topology is accepted for regular jobs with ElasticJobsViaWorkloadSlicesWithTAS": {
```
mimowo, line 4265: Actually, I prefer to split the PR for preferred and required, because for preferred we don't have any clear expectations of results, and so the problem is much easier.

For required I think the problem is quite hard, and we should agree on the algorithm.

In both cases we need KEP update, and also scheduling tests showing the assignments successfully scaled up / scaled down.

yaroslava-serdiuk: Currently if the ElasticWorkloadWithTAS feature is enabled it blocks TAS assignment for **non** elastic workloads with preferred or required topology, but it should not. This PR just fixing it. It doesn't change the behavior for elastic workloads.

mimowo: ahh, good point, thank you for clarifying 👍

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for hardening the ElasticJobs support 👍

mimowo:
/release-note-edit
```release-note
ElasticJobsViaWorkloadSlices: Fix the bug that regular (non-elastic) workloads with the required/preferred topology
were rejected when the feature ElasticJobsViaWorkloadSlicesWithTAS is enabled.
```

PR #12002, opened 2026-06-08
Classify DRA reconcile errors as transient vs deterministic

### LINKED ISSUE #11994: Classify DRA reconcile errors as transient vs deterministic
## Why

DRA error paths in the workload reconciler treat all errors the same i.e. patch inadmissible and retry with backoff. This mixes transient API failures (should backoff) with deterministic conditions like "no matching devices" (should wait for external resource changes, not retry). The unnecessary retries on deterministic errors waste reconciler cycles.
  
  PR #11943 attempted a blanket `RequeueAfter` but was reverte because it suppressed backoff for transient failures too.
  
  ## What needs to change
  
 The DRA helpers return `field.ErrorList` without distinguishing error categories. The reconciler needs a way to tell whether the errors are transient (retry) or deterministic (stop retrying).

One approach: return a typed wrapper alongside the field errors that indicates whether any error requires backoff, so the reconciler can branch accordingly.
  
  Note: for deterministic "wait for external change" cases, the workload must be discoverable by the existing watchers (DeviceClass handler, ResourceSlice reconciler) before `return nil` is safe. The DeviceClass path already has direct API-level requeue; the ResourceSlice path uses `NotifyRetryInadmissible` which requires the workload to be in the queue manager's inadmissible set.

thc1006:
#### What type of PR is this?
/kind cleanup

DRA resolution errors in the workload reconciler were all returned as reconcile
errors, so permanent failures (unsupported DRA features, unmapped device classes,
non-integer extended resources) were retried with controller-runtime backoff even
though retrying cannot change the outcome.

The three DRA helpers now report whether their errors warrant a retry:
- Internal/API failures and cluster-state shortages ("too few matching devices")
  keep the controller-runtime backoff, so the workload still recovers once the
  cluster or its ResourceSlices change.
- Permanent spec/configuration errors leave the workload inadmissible and return
  without requeue.

This supersedes the reverted #11943 (fixed-rate requeue) by distinguishing the
error categories instead of applying one policy to all of them.

Fixes #11994

- Device-shortage errors intentionally stay on backoff: the integration test
  "Should requeue inadmissible workload when ResourceSlice is created" relies on
  that recovery, and there is no ResourceSlice->workload informer yet (KEP-2941
  lists event-driven ResourceSlice requeuing as an alpha graduation item).
- Verified: pkg/dra and pkg/controller/core unit tests, the DRA controller
  integration suite, gofmt, go vet, and golangci-lint.

```release-note
DRA: Fixed hot reconcile loops for inadmissible Workloads with deterministic DRA resolution
failures. Kueue now avoids requeueing permanent DRA spec or configuration errors while still
retrying transient failures with backoff.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling in device resource allocation operations to properly identify and process transient failures
  * Device resource validation failures now correctly classified as recoverable errors
  
* **Tests**
  * Added test coverage for transient failures in resource discovery operations

### DIFF
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -77,6 +77,21 @@ var (
 	realClock = clock.RealClock{}
 )
 
+// hasInternalError reports whether the field error list contains an internal
+// error. DRA resolution reports retryable failures — API errors and cluster-state
+// shortages that clear once ResourceSlices change — as internal errors, so the
+// reconciler retries them with controller-runtime backoff; deterministic spec or
+// configuration errors use other field error types and are left inadmissible
+// without requeue.
+func hasInternalError(errs field.ErrorList) bool {
+	for _, e := range errs {
+		if e.Type == field.ErrorTypeInternal {
+			return true
+		}
+	}
+	return false
+}
+
 type waitForPodsReadyConfig struct {
 	timeout                     time.Duration
 	recoveryTimeout             *time.Duration
@@ -351,7 +366,10 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 			if updateErr != nil {
 				return ctrl.Result{}, fmt.Errorf("failed to update workload status for DRA error: %w", updateErr)
 			}
-			return ctrl.Result{}, err
+			if hasInternalError(fieldErrs) {
+				return ctrl.Result{}, err
+			}
+			return ctrl.Result{}, nil
 		}
 
 		// Process Extended Resources backed by DRA (new path)
@@ -373,7 +391,10 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 				if updateErr != nil {
 					return ctrl.Result{}, fmt.Errorf("failed to update workload status for DRA extended resources error: %w", updateErr)
 				}
-				return ctrl.Result{}, err
+				if hasInternalError(extFieldErrs) {
+					return ctrl.Result{}, err
+				}
+				return ctrl.Result{}, nil
 			}
 		}
 
@@ -408,7 +429,10 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 				if updateErr != nil {
 					return ctrl.Result{}, fmt.Errorf("failed to update workload status for DRA counter resources error: %w", updateErr)
 				}
-				return ctrl.Result{}, err
+				if hasInternalError(counterFieldErrs) {
+					return ctrl.Result{}, err
+				}
+				return ctrl.Result{}, nil
 			}
 			for podSetName, resources := range counterResources {
 				if existing, ok := draResources[podSetName]; ok {
--- a/pkg/controller/core/workload_controller_test.go
+++ b/pkg/controller/core/workload_controller_test.go
@@ -416,6 +416,7 @@ func TestReconcile(t *testing.T) {
 		resourceClaims            []*resourcev1.ResourceClaim
 		resourceClaimTemplates    []*resourcev1.ResourceClaimTemplate
 		patchErr                  error
+		listErr                   error
 		wantDRAResourceTotal      *int64
 		wantWorkloadsInQueue      *int
 		wantWorkload              *kueue.Workload
@@ -669,8 +670,7 @@ func TestReconcile(t *testing.T) {
 				}
 				return wl
 			}(),
-			wantErrorMsg: "not mapped in DRA configuration",
-			wantEvents:   nil,
+			wantEvents: nil,
 		},
 		"reconcile DRA ResourceClaimTemplate not found should return error": {
 			featureGates: map[featuregate.Feature]bool{
@@ -708,6 +708,33 @@ func TestReconcile(t *testing.T) {
 				}).
 				Obj(),
 			wantErrorMsg: "failed to get claim spec",
+			wantEvents:   nil,
+		},
+		"reconcile DRA transient ResourceSlice list failure should back off": {
+			featureGates: map[featuregate.Feature]bool{
+				features.KueueDRAIntegration:              true,
+				features.MultiKueueOrchestratedPreemption: false,
+			},
+			listErr: errTest,
+			workload: utiltestingapi.MakeWorkload("wlListErrDRA", "ns").
+				Queue("lq").
+				PodSets(*utiltestingapi.MakePodSet(kueue.DefaultPodSetName, 1).
+					ResourceClaimTemplate("gpu", "gpu-template").
+					Obj()).
+				Obj(),
+			resourceClaimTemplates: []*resourcev1.ResourceClaimTemplate{
+				utiltesting.MakeResourceClaimTemplate("gpu-template", "ns").
+					DeviceRequest("gpu-request", "gpu.example.com", 1).
+					WithCELSelectors("device.driver == \"test-driver\"").
+					Obj(),
+			},
+			cq: utiltestingapi.MakeClusterQueue("cq").
+				ResourceGroup(
+					*utiltestingapi.MakeFlavorQuotas("flavor1").
+						Resource("gpu", "2").Obj(),
+				).Obj(),
+			lq:           utiltestingapi.MakeLocalQueue("lq", "ns").ClusterQueue("cq").Obj(),
+			wantErrorMsg: "failed to list ResourceSlices",
 			wantEvents:   nil,
 		},
 		"assign Admission Checks from ClusterQueue.spec.AdmissionCheckStrategy": {
@@ -3430,6 +3457,14 @@ func TestReconcile(t *testing.T) {
 								return tc.patchErr
 							}
 							return utiltesting.TreatSSAAsStrategicMerge(ctx, client, subResourceName, obj, patch, opts...)
+						},
+						List: func(ctx context.Context, c client.WithWatch, list client.ObjectList, opts ...client.ListOption) error {
+							if tc.listErr != nil {
+								if _, ok := list.(*resourcev1.ResourceSliceList); ok {
+									return tc.listErr
+								}
+							}
+							return c.List(ctx, list, opts...)
 						},
 					})
 				cl := clientBuilder.Build()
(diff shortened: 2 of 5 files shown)


pkg/controller/core/workload_controller.go
```diff
@@ -337,7 +337,7 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		log.V(3).Info("Processing DRA resources for workload")
 
 		// Process ResourceClaimTemplates (existing DRA path)
-		draResources, fieldErrs := dra.GetResourceRequestsForResourceClaimTemplates(ctx, r.client, r.draMapper, &wl)
+		draResources, fieldErrs, needsBackoff := dra.GetResourceRequestsForResourceClaimTemplates(ctx, r.client, r.draMapper, &wl)
```
mimowo: Hm, I think this is an ok approach, but it is a bit cumbersome to return another parameter just for that. 

Let's start by thinking when `needsBackoff`=true` if there is an internalError on the list of errors.

So here are some alternatives I'm thinking about:
1. introduce a helper function `needsBackoff(fieldErrs) bool ` which would introspect fieldErrs checking if there is an internal error on the list 
2. encapsulate ` fieldErrs, needsBackoff ` into a struct like `DRAValidationResult` that could have a function like NeedsBackoff() bool, Errors(), etc.

I think I'm leaning towards (1.) to make the changes minimal for now, but let me check with @sohankunkerkar

mimowo: (1.) would also be conceptually close to the pre-existing pattern of IsUnretryableError helper

thc1006: Thanks so much for the thoughtful review — and for the `IsUnretryableError` pointer, that's a great precedent. I agree a `needsBackoff(fieldErrs)` helper reads much more cleanly than threading an extra return value through each helper, and I'm happy to move in that direction.

One subtlety I'd like to surface first, since it affects whether the helper can introspect purely by error type: the "insufficient matching devices" / counter-shortage errors are `field.Invalid`, yet they currently need to keep backing off. The integration test `dra_pd_test.go` ("Should requeue inadmissible workload when ResourceSlice is created") relies on the reconcile retry to recover once a matching ResourceSlice appears — there's no ResourceSlice→workload informer yet, which KEP-2941 lists as an alpha-graduation item. A pure `hasInternalError` check makes those `return nil`, and that recovery test fails locally. It's the one case that error type alone doesn't capture.

So to keep your option (1) while preserving today's behavior, I see two small variants:
- **1a** — reclassify the recoverable cluster-state errors (device shortage, unresolved DeviceClass) as `field.InternalError`, so `needsBackoff = hasInternalError(fieldErrs)` matches your model exactly. Trade-off: the inadmissible message becomes "Internal error: insufficient matching devices…".
- **1b** — keep them as `field.Invalid` (clearer user-facing message) and let the result carry the distinction, classified where each error is created (close to your option 2).

Since you're already looping in @sohankunkerkar: the deciding question is really a product one — should "insufficient matching devices" keep retrying (recover via backoff, as today), or become non-retryable now and rely on event-driven ResourceSlice requeuing later? That's what picks between 1a and 1b. Happy to implement whichever you both prefer — just let me know and I'll update the PR.

mimowo: > The integration test dra_pd_test.go ("Should requeue inadmissible workload when ResourceSlice is created") relies on the reconcile retry to recover once a matching ResourceSlice appears — there's no ResourceSlice→workload informer yet, which KEP-2941 lists as an alpha-graduation item

Interesting, could you double check if the test would indeed fail if not for the backoff? It seems we already have the ResourceSlice controller added, see here: https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/core/resourceslice_controller.go. Or maybe there is some channel update missing to trigger reconcile of such workloads? 

So I would like to consider **1c**: keep them as `field.Invalid`, but use the simple logic `needsBackoff = hasInternalError(fieldErrs)`, and make sure we trigger reconcile when the ResourceSlice is added. 

I think the long term plan. If this is tricky to achieve, then I would lean towards **1a** as closer to the end goal, so less changes in the future.
 
 Yeah, let's see what @sohankunkerkar would say.

thc1006: Edit: please disregard this one — it flaked under heavier load (the recovery spec overran its 10s timeout, 4/4), so device-shortage needs to stay on backoff after all. Going with 1a; details in my follow-up below.

thc1006: Following up on my own comment to correct course. I measured it properly this time: under back-to-back load the "requeue when ResourceSlice is created" spec fails 4/4 on its 10s timeout with device-shortage → `nil`, while the same suite on backoff passes 4/4 under identical load. Under lighter load both pass, so `nil` is borderline rather than broken — but it leans on the requeue batch period instead of an active retry, and I'd rather not risk it flaking on busy CI runners.

So I'll keep your `needsBackoff(fieldErrs)` helper but reclassify the recoverable cluster-state errors (device shortage) as `field.InternalError`, so `hasInternalError` keeps them on backoff (your 1a). Reliable, and still the simple helper — the only cost is the inadmissible message reading "Internal error: …". 1c (an active ResourceSlice→workload trigger) is still the nicer long-term fix, but it's the bigger KEP-alpha piece, so I'll leave it as a follow-up unless you'd rather have it now.

Sorry for the zig-zag on this — wanted to actually measure before committing.

sohankunkerkar: LGTM. +1 on @mimowo's `hasInternalError` helper as `InternalError = transient` infrastructure failure, `Invalid = deterministic`, matches how `field.ErrorType` is used across K8s and the existing IsUnretryableError pattern. Can you add tests for the transient error paths (e.g., API list failure)?

mimowo: What if you increase the timeout to 45s (MediumTimeout)?

What we care the most is eventual consistency, I don't think we need to squeeze seconds out of the scenario, at the expense of code complication and slightly degraded error messages.

Basically, I try to avoid premature optimization. We can ofc optimize when this becomes a bottleneck for some usecases, but I highly doubt it would matter in practice.

I'm also wondering how both approaches would impact the volume of logs produced by the test.

thc1006: Sounds good, MediumTimeout it is. Agreed — eventual consistency is the goal here, and I'd rather not reclassify device-shortage to `InternalError` just to shave recovery time (exactly the code complication + "Internal error" message you mentioned). On log volume, `nil` is actually the lighter one: backoff re-reconciles and re-logs the failure on every retry, while `nil` logs once and waits for the watcher. So I'll go with the plain `hasInternalError` helper (`Invalid` → deterministic → `nil`), bump that recovery spec to `MediumTimeout`, and add the transient-path tests @sohankunkerkar asked for. Pushing shortly.

thc1006: Thanks @mimowo — I gave MediumTimeout (45s) a proper try, run as a matched A/B against the backoff variant on the same machine, alternating runs to rule out load drift:

- backoff: 3/3 pass
- return-`nil` @ 45s: 3/3 fail (each times out right at 45.0s on "Should requeue inadmissible workload when ResourceSlice is created")

So the bump alone doesn't get us there — and I don't think it's about squeezing seconds; the timeout was never the bottleneck. It's the point you raised earlier: there's no ResourceSlice→Workload trigger today. When the shortage error is swallowed with `nil`, the workload isn't requeued and isn't in the queue manager's inadmissible set either, so the ResourceSlice controller's `NotifyRetryInadmissible` never reaches it. Recovery then rides whatever incidental reconcile happens to fire — borderline under light load, consistently over 45s under load.

For the genuinely-transient paths (API/list failures → `InternalError`) the `hasInternalError` helper is exactly right, and I'll add the transient tests @sohankunkerkar asked for. The one open point is the device-shortage case: semantically it's deterministic-given-current-state, so `nil` is the "right" label — but until there's an active ResourceSlice→Workload retrigger (your 1c), `nil` has no bounded path back, so it needs to keep backing off to stay green.

I'd suggest either: (1) keep device-shortage on backoff for now, ship the helper + transient tests, and file the retrigger as a follow-up — then flip device-shortage to `nil` cleanly once it lands; or (2) add the trigger in this PR. Leaning towards (1) to keep this focused — any preference?

mimowo: @thc1006 thank you for the summary. Just checking if my understanding of the suggestion is (1.) is correct:
a) this PR would use the simple hasInternalError, and we will return InternalError on "DeviceClass not found" to trigger the backoff (as currently, just a slightly changed message)
b) the follow up PR will change the error returned to InvalidField, but will use the proper trigger for reconcile when a new DeviceClass is created

I'm good with the plan, doing (a) alone will already benefit a couple of other scenarios and prepare the code for (b).

thc1006: Thanks @mimowo — glad we're on the same page. One small clarification on which errors actually change here: the ones moving to `InternalError` are the cluster-state shortages ("insufficient matching devices" on both the CEL and counter paths, and "no consumesCounters"). They're `field.Invalid` today, and they clear on their own once the matching ResourceSlices appear. "failed to get DeviceClass" is already an `InternalError`, and "not mapped in DRA configuration" stays deterministic, so that one still returns without a requeue.

I've pushed (a) — the simple `hasInternalError` helper, plus the transient test for the list failure that @sohankunkerkar asked for. For the follow-up (b) I'll switch those shortages back to `Invalid` once the proper trigger is in place. Thanks to you both!

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you fot the contribution to the Kueue-DRA integration 👍

mimowo:
@thc1006 if feasible please prep the CP for 0.17

thc1006:
Thanks @mimowo. The automated cherry-pick to release-0.17 hit conflicts (0.17 predates the DRA counter and CEL paths), so I've opened a manual, adapted backport as #12094. It keeps the `hasInternalError` classification for the claims and extended-resource paths that exist on 0.17. PTAL when you get a chance 🙏

mimowo:
/release-note-edit
```release-note
DRA: Fixed hot reconcile loops for inadmissible Workloads with deterministic DRA resolution
failures. Kueue now avoids requeueing permanent DRA spec or configuration errors while still
retrying transient failures with backoff.
```

Issue #11990, opened 2026-06-08
Optimize periodic-kueue-test-e2e-sequential-extended-main by sharding


mimowo:
**What would you like to be cleaned**:

Optimize periodic-kueue-test-e2e-sequential-extended-main by sharding 

**Why is this needed**:

To get below 18min with all CI testing: https://github.com/kubernetes-sigs/kueue/issues/11606

apullo777:
Hi, I'd like to work on this :)
  
I looked at a recent run (21 specs, ~328s) and the heaviest part is the Spark setup (apache/spark + spark-operator) are by far the largest images, while the rest of the suite is lighter. So my initial thought is to isolate Spark into its own shard by feature label, so the other shard can skip those image pulls:
- shard-0: feature:spark
- shard-1: the rest (managejobswithoutqueuename, workloadidentifierannotations)

Most of the job time is actually setup rather than test exec, so I'll confirm the wall-clock balance and adjust the split if needed. Does this direction sound reasonable?

/assign

mimowo:
Hi @apullo777 this makes sense to me, I would just suggest making spark `shard-1` so that we don't need to rename the other ones if we delete the shard for spark one day. This is possible because:
1. if we graduate it to beta then we will not need to change configuration and we can just run it as part of regular singlecluster build
2. we may externalize it to a subproject for external integrations (I will work on the issue)

I think (2.) ,might be more likely, cc @everpeace 

Also, I think we could move some tests for the core Job from `managejobswithoutqueuename_test.go` to the sequential baseline suite.

apullo777:
Thanks @mimowo ! Both points make sense.

I will go with spark as shard-1 (so removing it later won't renumber the others).

Two scoping notes:
1. I will shard the presubmit (pull-…-main) alongside the periodic; both on main, mirroring the baseline shards.
2. For moving core-Job tests to baseline, happy to take that as a separate follow-up! I will keep this PR focused on the sharding.

mimowo:
Sgtm

mimowo:
/close
This is already done by https://github.com/kubernetes-sigs/kueue/pull/12004, https://github.com/kubernetes/test-infra/pull/37230 and  https://github.com/kubernetes-sigs/kueue/pull/12069

PR #11964, opened 2026-06-05
Use feature labels for shards.

mbobrovskyi:
#### What type of PR is this?
/kind cleanup

#### What this PR does / why we need it:
Use feature labels for shards.

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

## Summary by CodeRabbit

* **Tests**
  * E2E selection now uses individual feature labels instead of shard-based labels.
  * Extended test targets updated to run feature-specific test sets; KubeRay runs now ensure Ray image prep when selected.
  * Test startup and dependency/image preparation gate operator installs by feature labels only, removing shard gating for more targeted runs.

PR #12004, opened 2026-06-08
Optimize the sequential/extended e2e test suites by sharding

apullo777:
#### What type of PR is this?

/kind cleanup
/area testing

Shards the `sequential/extended` e2e suite so the heavy Spark setup is isolated, following the same feature-label sharding approach as the `singlecluster/extended` suite (#11964). The suite is split into two shards:
  
- shard-0 (`feature:managejobswithoutqueuename`, `feature:workloadidentifierannotations`): 9 specs; installs jobset/lws/appwrapper.
- shard-1 (`feature:spark`): 2 specs; installs only the Spark operator + image.
  
Current pre-shard runtimes (`hack/tools/prow-runtimes`, last 6 successful runs):

| Job | Avg |
|---|---:|
| `periodic-kueue-test-e2e-sequential-extended-main` | 18:27 |
| `pull-kueue-test-e2e-sequential-extended-main` | 21:12 |
 

Part of #11990

- Verified spec allocation with `ginkgo --dry-run`: shard-0 = 19, shard-1 = 2, total = 21 (no specs dropped or duplicated).
- Two pieces remain: (1) the paired test-infra PR and (2) release-0.17/0.18 backports. How would you prefer to sequence these: test-infra `main` right after this, or bundle everything (including release) into one test-infra PR?
    

```release-note
None
```

## Summary by CodeRabbit

* **Tests**
  * Split the extended end-to-end suite into two shard targets so related test groups run separately with shard-specific label filters (managejobs/workload identifiers vs. spark).
  * Assign shard-scoped operator versions and add feature labels to targeted e2e suites to ensure each shard runs with the appropriate configuration for consistent results.

### DIFF
--- a/Makefile-test.mk
+++ b/Makefile-test.mk
@@ -266,12 +266,32 @@ test-e2e-sequential-baseline-shard-1: test-e2e-sequential-baseline
 test-e2e-sequential-baseline-helm: E2E_USE_HELM=true
 test-e2e-sequential-baseline-helm: test-e2e-sequential-baseline
 
+# Assign Shard 0 variables to the shard-0 target
+TEST_E2E_SEQUENTIAL_EXTENDED_SHARD_0_TARGETS := test-e2e-sequential-extended-shard-0
+$(TEST_E2E_SEQUENTIAL_EXTENDED_SHARD_0_TARGETS): export JOBSET_VERSION := $(JOBSET_VERSION)
+$(TEST_E2E_SEQUENTIAL_EXTENDED_SHARD_0_TARGETS): export APPWRAPPER_VERSION := $(APPWRAPPER_VERSION)
+$(TEST_E2E_SEQUENTIAL_EXTENDED_SHARD_0_TARGETS): export LEADERWORKERSET_VERSION := $(LEADERWORKERSET_VERSION)
+
+# Assign Shard 1 variables to the shard-1 target
+TEST_E2E_SEQUENTIAL_EXTENDED_SHARD_1_TARGETS := test-e2e-sequential-extended-shard-1
+$(TEST_E2E_SEQUENTIAL_EXTENDED_SHARD_1_TARGETS): export SPARKOPERATOR_VERSION := $(SPARKOPERATOR_VERSION)
+
 ## Label Taxonomy:
-##   Features: managejobswithoutqueuename, spark
+##   Features: managejobswithoutqueuename, workloadidentifierannotations, spark
 ## Examples:
 ##   Run only Spark Integration tests: GINKGO_ARGS="--label-filter=feature:spark" make test-e2e-sequential-extended
 .PHONY: test-e2e-sequential-extended
-test-e2e-sequential-extended: setup-e2e-env run-test-e2e-sequential-extended-$(E2E_KIND_VERSION:kindest/node:v%=%)
+test-e2e-sequential-extended:
+	$(MAKE) test-e2e-sequential-extended-shard-0 E2E_USE_HELM=$(E2E_USE_HELM)
+	$(MAKE) test-e2e-sequential-extended-shard-1 E2E_USE_HELM=$(E2E_USE_HELM)
+
+.PHONY: test-e2e-sequential-extended-shard-0
+test-e2e-sequential-extended-shard-0: GINKGO_ARGS=--label-filter=feature:managejobswithoutqueuename,feature:workloadidentifierannotations
+test-e2e-sequential-extended-shard-0: setup-e2e-env run-test-e2e-sequential-extended-$(E2E_KIND_VERSION:kindest/node:v%=%)
+
+.PHONY: test-e2e-sequential-extended-shard-1
+test-e2e-sequential-extended-shard-1: GINKGO_ARGS=--label-filter=feature:spark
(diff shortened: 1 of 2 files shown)


Makefile-test.mk
```diff
@@ -270,9 +270,19 @@ test-e2e-sequential-baseline-helm: test-e2e-sequential-baseline
...
 .PHONY: test-e2e-sequential-extended
 test-e2e-sequential-extended: setup-e2e-env run-test-e2e-sequential-extended-$(E2E_KIND_VERSION:kindest/node:v%=%)
 
+.PHONY: test-e2e-sequential-extended-shard-0
+test-e2e-sequential-extended-shard-0: GINKGO_ARGS=--label-filter=feature:managejobswithoutqueuename,feature:workloadidentifierannotations
+test-e2e-sequential-extended-shard-0: test-e2e-sequential-extended
+
+.PHONY: test-e2e-sequential-extended-shard-1
+test-e2e-sequential-extended-shard-1: GINKGO_ARGS=--label-filter=feature:spark
+test-e2e-sequential-extended-shard-1: test-e2e-sequential-extended
```
mbobrovskyi: For the Spark feature, we should install only Spark, right?

Could you please take a look at how we implemented it for the test-e2e-extended shards?

https://github.com/kubernetes-sigs/kueue/blob/f79bd4aa59ea7fe5cc88b3ecdb74b5f40a51c036/Makefile-test.mk#L178-L210

I think we should implement something very similar here.

apullo777: Yes. The spark shard should only install Spark. I think I missed that. 

Thanks a lot for the reference! Reworked it to follow that pattern.

Makefile-test.mk
```diff
@@ -266,13 +266,33 @@ test-e2e-sequential-baseline-shard-1: test-e2e-sequential-baseline
...
+# Assign Shard 1 variables to the parent suite AND all shard-1 targets
+TEST_E2E_SEQUENTIAL_EXTENDED_SHARD_1_TARGETS := test-e2e-sequential-extended test-e2e-sequential-extended-shard-1
+$(TEST_E2E_SEQUENTIAL_EXTENDED_SHARD_1_TARGETS): export SPARKOPERATOR_VERSION := $(SPARKOPERATOR_VERSION)
+
 ## Label Taxonomy:
 ##   Features: managejobswithoutqueuename, spark
 ## Examples:
 ##   Run only Spark Integration tests: GINKGO_ARGS="--label-filter=feature:spark" make test-e2e-sequential-extended
+##   Run only shard 0 tests: make test-e2e-sequential-extended-shard-0
+##   Run only shard 1 tests: make test-e2e-sequential-extended-shard-1
```
mbobrovskyi:
```suggestion
```

I don't think so we need this.

Makefile-test.mk
```diff
@@ -266,13 +266,33 @@ test-e2e-sequential-baseline-shard-1: test-e2e-sequential-baseline
...
 .PHONY: test-e2e-sequential-extended
 test-e2e-sequential-extended: setup-e2e-env run-test-e2e-sequential-extended-$(E2E_KIND_VERSION:kindest/node:v%=%)
```
mbobrovskyi:
```suggestion
test-e2e-sequential-extended: test-e2e-sequential-extended-shard-0 test-e2e-sequential-extended-shard-1
```
We need to ensure that it works as expected.

apullo777: Thanks! I implemented the intent having the parent run both shards. 

I made the parent invoke each shard in its own sub-make instead, which runs both correctly and threads E2E_USE_HELM through. (literal suggestion doesn't work: both shards share the run-test-e2e-sequential-extended-% prerequisite, so make builds it only once and only shard-0 runs)

The parent no longer honors a GINKGO_ARGS override (it'd hit both shards), so I also pointed that example at the shard target.

mbobrovskyi: Ah, I see. Ok.

Makefile-test.mk
```diff
@@ -266,12 +266,32 @@ test-e2e-sequential-baseline-shard-1: test-e2e-sequential-baseline
...
+# Assign Shard 1 variables to the shard-1 target
+TEST_E2E_SEQUENTIAL_EXTENDED_SHARD_1_TARGETS := test-e2e-sequential-extended-shard-1
+$(TEST_E2E_SEQUENTIAL_EXTENDED_SHARD_1_TARGETS): export SPARKOPERATOR_VERSION := $(SPARKOPERATOR_VERSION)
+
 ## Label Taxonomy:
-##   Features: managejobswithoutqueuename, spark
+##   Features: managejobswithoutqueuename, workloadidentifierannotations, spark
 ## Examples:
-##   Run only Spark Integration tests: GINKGO_ARGS="--label-filter=feature:spark" make test-e2e-sequential-extended
+##   Run only Spark tests: make test-e2e-sequential-extended-shard-1
```
mbobrovskyi: Please revert this line. I think it's much better to keep it as it is, since the shard targets are only used for test-infra.

mimowo (commented):
@apullo777 please already prepare the test-infra PR so that we minimize the transition period

mbobrovskyi:
Thank you!

/lgtm
/approve

/hold for addressing https://github.com/kubernetes-sigs/kueue/pull/12004#pullrequestreview-4466061218.

mimowo:
/unhold
As we already have https://github.com/kubernetes/test-infra/pull/37230

PR #12006, opened 2026-06-09
TAS: Use PodSpec resources for node capacity tracking

### LINKED ISSUE #12005: TAS ignores excluded or transformed resources in node capacity tracking
**What happened**:

TAS does not consider resources excluded by `excludeResourcePrefixes` or transformed by `transformations` when calculating node capacity.
TAS may schedule pods to nodes with insufficient remaining resources.

#8865 seems to address this issue, but TAS still ignores resources used by admitted workloads.

related: #8838 

TAS should consider all resource requests of admitted workloads for node capacity tracking, including resources excluded or transformed.

https://github.com/wafrelka/bugbox/tree/e6c312d41f8e72828f50fe28cc3af6f9ac05ef3b

https://github.com/wafrelka/kueue/blob/67ab8032e6b6ef79539d36adc62032ed55623650/test/integration/singlecluster/scheduler/excluderesources/scheduler_test.go#L331-L375

**Environment**:
  - client: v1.35.5
  - server: v1.35.1
  - 0.18.0
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

wafrelka:
#### What type of PR is this?

/kind bug

/area tas

Fixes #12005

I'm not yet sure whether this PR addresses all cases where nominal and actual resource requests are mixed up.

```release-note
TAS: Fix a bug where TAS ignores excluded or transformed resources in node capacity tracking.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Derive topology-aware scheduling (TAS) resource requests from pod templates to improve TAS net-usage accuracy and avoid double-counting.
  * Skip redundant TAS computation when a prior admission already contains a topology assignment.
  * Thread contextual logging into usage and TAS update paths for clearer diagnostics.
  * Ensure excluded resources respect per-node capacity during TAS.

* **Tests**
  * Added unit tests validating TAS net-usage generation and skipping logic.
  * Added integration test for TAS with excluded-resource handling.

### DIFF
--- a/pkg/scheduler/flavorassigner/flavorassigner.go
+++ b/pkg/scheduler/flavorassigner/flavorassigner.go
@@ -79,7 +79,7 @@ type Assignment struct {
 }
 
 // UpdateForTASResult updates the Assignment with the TAS result
-func (a *Assignment) UpdateForTASResult(cq *schdcache.ClusterQueueSnapshot, wl *workload.Info, result schdcache.TASAssignmentsResult) {
+func (a *Assignment) UpdateForTASResult(log logr.Logger, cq *schdcache.ClusterQueueSnapshot, wl *workload.Info, result schdcache.TASAssignmentsResult) {
 	for psName, psResult := range result {
 		psAssignment := a.podSetAssignmentByName(psName)
 		psAssignment.TopologyAssignment = psResult.TopologyAssignment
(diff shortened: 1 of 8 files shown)


pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -78,50 +79,43 @@ type Assignment struct {
...
-			for resourceName, quantity := range psa.Requests {
-				flv := psa.Flavors[resourceName]
-				if flv == nil || cq.TASFlavors[flv.Name] == nil {
-					// This is the quota-only resources, skip when computing TAS usage.
-					continue
-				}
-				tasRequests[resourceName] = quantity
+			podSet := podset.FindPodSetByName(wl.Obj.Spec.PodSets, psa.Name)
+			if podSet == nil {
+				continue
```
mimowo, line 104: I think it would be better to add logging here, and when `err != nil ` for `onlyTASFlavor`, because both are unexpected. 

I know there is currently no logger in the function, but I think it is reasonable to pass it as part of the PR. We could also consider including logger in the Assignment struct. 

I don't have an opinion here, but probably passing as an argument is less changes.

mimowo (commented):
LGTM, this is an excellent finding. Thank you for making TAS more robust 👍

pkg/scheduler/flavorassigner/flavorassigner.go
wafrelka: I agree.
Added `log.Error` in 8550162e299b26ad78cf5c9495a871a832940276.

mimowo:
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for the important TAS bugfix 👍

mimowo:
@wafrelka please prepare the cherrypick for 0.17, you can make this semi-automated by ./hack/cherry_pick_pull.sh script

PR #12011, opened 2026-06-09
Add accessProviders to MultiKueue ClusterProfiles

kahirokunn:
## What type of PR is this?

/kind feature
/kind api-change
/kind deprecation

/area multikueue

## What this PR does / why we need it:

It also adds `accessProviders` to the v1beta2 `ClusterProfile` configuration used by MultiKueue. The existing `credentialsProviders` field is kept for compatibility, marked as deprecated, and still validated with the same rules. When both fields define the same provider name, the `accessProviders` entry takes precedence.

Docs, examples, generated API reference, and tests are updated to show `accessProviders` as the preferred field.

## Which issue(s) this PR fixes:

None

## Special notes for your reviewer:

The dependency bump includes a small MultiKueue source change because `cluster-inventory-api` v0.1.3 moved the provider implementation from `pkg/credentials` to `pkg/access`.

`credentialsProviders` remains accepted for compatibility, but new examples and docs use `accessProviders`. If the same provider name appears in both fields, the `accessProviders` entry is used.

Local verification run:

```sh
go test ./pkg/controller/admissionchecks/multikueue ./pkg/config ./cmd/kueue
make gomod-verify
make generate-apiref
make test-multikueue-integration INTEGRATION_FILTERS='--label-filter=feature:multikueue'
```

AI disclosure: This PR was written in part with the assistance of generative AI.

## Does this PR introduce a user-facing change?

```release-note
MultiKueue: Added `accessProviders` as the preferred `ClusterProfile` field for
configuring cluster access providers. The existing `credentialsProviders` field remains
supported but is deprecated and cannot be used together with `accessProviders`.
```

## Summary by CodeRabbit

* **New Features**
  * Added ClusterProfile.accessProviders to configure cluster access providers.

* **Deprecations**
  * ClusterProfile.credentialsProviders marked deprecated; accessProviders takes precedence on name collisions while legacy entries remain supported for compatibility.

* **Validation**
  * Validation now enforces required exec fields and allowed interactive modes for accessProviders.

* **Documentation**
  * Reference and setup docs/examples updated to use accessProviders and explain deprecation/precedence.

* **Tests**
  * Unit, integration and e2e tests updated for accessProviders and legacy compatibility.

PR #12015, opened 2026-06-09
Document MultiKueue migration to ClusterProfile accessProviders

kahirokunn:
#### What type of PR is this?

/kind kep
/area multikueue

Updates KEP-693 to document MultiKueue's migration from ClusterProfile `credentialsProviders` to `accessProviders`.

The Cluster Inventory API renamed the provider concept because the old `credentialProviders` field actually carries cluster access information rather than credential material. The new `accessProviders` field is now the preferred API, and `credentialProviders` is deprecated:

- kubernetes-sigs/cluster-inventory-api#23 introduced the rename direction.
- kubernetes-sigs/cluster-inventory-api#52 propagated the rename from credentials to access in the Cluster Inventory API code.

The Cluster Inventory API still keeps compatibility for the deprecated field today, but that compatibility is expected to be removed in the future:

- kubernetes-sigs/cluster-inventory-api#65 tracks removal of credential provider compatibility.

Because MultiKueue consumes ClusterProfile access information, this KEP update records the intent to migrate before the deprecated Cluster Inventory API field is removed. The KEP now describes `multiKueue.clusterProfile.accessProviders` as the preferred MultiKueue configuration field, keeps `credentialsProviders` as a deprecated compatibility field, and records the precedence rule when both fields define the same provider name.

None

This is a KEP-only follow-up for the API shape being implemented in #12011.

AI disclosure: This PR was written in part with the assistance of generative AI.

```release-note
NONE
```

## Summary by CodeRabbit

* **Documentation**
  * Updated the MultiKueue KEP-693 ClusterProfile API/config examples to use a configured `accessProviders` list for cluster access provisioning.
  * Added `accessProviders` (`ClusterProfileAccessProvider` with `Name` and `ExecConfig`) and retained `credentialsProviders` as a deprecated compatibility option; `ClusterProfileCredentialsProvider` is now an alias.
  * Clarified precedence and status fields: `accessProviders` is preferred over `credentialsProviders`, and `status.accessProviders` is preferred while `status.credentialProviders` remains supported.

### DIFF
--- a/keps/693-multikueue/README.md
+++ b/keps/693-multikueue/README.md
@@ -276,8 +276,8 @@ type ClusterProfile struct {
 
 	// CredentialsProviders defines a list of providers to obtain credentials of worker clusters
 	// using the ClusterProfile API.
-	// Deprecated: Use AccessProviders instead. If both AccessProviders and CredentialsProviders are provided,
-	// both are used. In case they specify a provider with the same name, the one in AccessProviders is preferred.
+	// Deprecated: Use AccessProviders instead. AccessProviders and CredentialsProviders
+	// are mutually exclusive.
 	CredentialsProviders []ClusterProfileCredentialsProvider `json:"credentialsProviders,omitempty"`
 }
 
@@ -294,7 +294,7 @@ type ClusterProfileAccessProvider struct {
 type ClusterProfileCredentialsProvider = ClusterProfileAccessProvider
 ```
 
-  The `credentialsProviders` field remains accepted for backwards compatibility. If both `accessProviders` and `credentialsProviders` are specified, the controller uses providers from both fields. When both fields specify a provider with the same name, the provider in `accessProviders` takes precedence.
+  The `credentialsProviders` field remains accepted for backwards compatibility when `accessProviders` is not configured. The `accessProviders` and `credentialsProviders` fields are mutually exclusive, so users should migrate existing `credentialsProviders` entries to `accessProviders`. The deprecated `credentialsProviders` field and `ClusterProfileCredentialsProvider` alias will be removed in the `v1beta3` Configuration API.
 
   On the ClusterProfile API side, Kueue uses `status.accessProviders` as the preferred source of cluster access information. The deprecated `status.credentialProviders` field remains supported by the ClusterProfile API compatibility layer.
 
--- a/keps/693-multikueue/README.md
+++ b/keps/693-multikueue/README.md
@@ -258,7 +258,7 @@ The controller obtains cluster credentials based on the `MultiKueueCluster` spec
 
   The authentication flow is as follows:
   1. The controller reads the `ClusterProfile` object referenced by the `MultiKueueCluster`. The `ClusterProfile` contains a list of access providers.
-  2. The controller uses this configuration to match the access provider with the credentials provider and invoke the credential plugin binary. The list of credentials providers is configured in the `CredentialsProviders` section under `ClusterProfile` in the `MultiKueue` in the Configuration API.
+  2. The controller uses this configuration to match the cluster access provider with a configured access provider and invoke the credential plugin binary. The list of access providers is configured in the `accessProviders` section under `clusterProfile` in the `MultiKueue` Configuration API.
 
 ```go
 type MultiKueue struct {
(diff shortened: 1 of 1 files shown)


mimowo:
cc @hdp617 @olekzabl @kshalot

keps/693-multikueue/README.md
```diff
@@ -270,20 +270,33 @@ type MultiKueue struct {
...
+
+// ClusterProfileCredentialsProvider defines a credentials provider in the ClusterProfile API.
+// Deprecated: Use ClusterProfileAccessProvider instead.
+type ClusterProfileCredentialsProvider = ClusterProfileAccessProvider
 ```
 
+  The `credentialsProviders` field remains accepted for backwards compatibility. If both `accessProviders` and `credentialsProviders` are specified, the controller uses providers from both fields. When both fields specify a provider with the same name, the provider in `accessProviders` takes precedence.
+
+  On the ClusterProfile API side, Kueue uses `status.accessProviders` as the preferred source of cluster access information. The deprecated `status.credentialProviders` field remains supported by the ClusterProfile API compatibility layer.
+
```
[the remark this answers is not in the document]
kahirokunn: This has already been implemented in a separate PR.
https://github.com/kubernetes-sigs/kueue/pull/12011

kshalot (commented):
/lgtm

Thank you for aligning this!

kshalot:
Although let's hold off merging this until https://github.com/kubernetes-sigs/kueue/pull/12011 is merged. I doubt the implementation will change much, so the KEP diff will likely stay the same, but just in case.

keps/693-multikueue/README.md
```diff
@@ -270,20 +270,33 @@ type MultiKueue struct {
...
+	// ExecConfig is the exec configuration to obtain credentials.
 	ExecConfig clientcmdapi.ExecConfig `json:"execConfig"`
 }
+
+// ClusterProfileCredentialsProvider defines a credentials provider in the ClusterProfile API.
+// Deprecated: Use ClusterProfileAccessProvider instead.
+type ClusterProfileCredentialsProvider = ClusterProfileAccessProvider
 ```
 
+  The `credentialsProviders` field remains accepted for backwards compatibility. If both `accessProviders` and `credentialsProviders` are specified, the controller uses providers from both fields. When both fields specify a provider with the same name, the provider in `accessProviders` takes precedence.
```
hdp617: I'm just curious why combining providers from both fields instead validate that they should be mutually exclusive.

hdp617:
/lgtm

Thank you for making this change. Just a quick question re. the behavior when both are specified.

keps/693-multikueue/README.md
kahirokunn: Good question. I chose the merge behavior because it follows the compatibility behavior defined by the Cluster Inventory API during the `credentialProviders` -> `accessProviders` migration.

The Cluster Inventory API type comment documents that the deprecated `CredentialProviders` field remains usable together with `AccessProviders`, with `AccessProviders` preferred on same-name conflicts:
https://github.com/kubernetes-sigs/cluster-inventory-api/blob/260ef011204c4091e3aca8fe08786cdf1fd8fe47/apis/v1alpha1/clusterprofile_types.go#L78-L82

The Cluster Inventory API helper implementation follows the same behavior: it loads legacy `CredentialProviders` first, then loads `AccessProviders`, so same-name entries from `AccessProviders` override the legacy
ones:
https://github.com/kubernetes-sigs/cluster-inventory-api/blob/260ef011204c4091e3aca8fe08786cdf1fd8fe47/pkg/access/config.go#L217-L228

Since Kueue is consuming the Cluster Inventory API here, I mirrored that compatibility behavior in Kueue's config instead of making the two fields mutually exclusive.

kahirokunn:
@hdp617 Thank you for your review!
I have sent a reply, so I would appreciate it if you could check it again. 🙏

hdp617:
> @hdp617 Thank you for your review!
> 
> I have sent a reply, so I would appreciate it if you could check it again. 🙏 

That makes sense. Thanks again for driving this change!

kahirokunn:
I'm happy to help make Kueue better! 🙌

mimowo:
/lgtm
/approve
In case @tenzen-y has some comments here
/hold

keps/693-multikueue/README.md
```diff
@@ -270,20 +270,34 @@ type MultiKueue struct {
...
+	// AccessProviders defines a list of providers to obtain access to worker clusters
+	// using the ClusterProfile API.
+	AccessProviders []ClusterProfileAccessProvider `json:"accessProviders,omitempty"`
+
 	// CredentialsProviders defines a list of providers to obtain credentials of worker clusters
 	// using the ClusterProfile API.
+	// Deprecated: Use AccessProviders instead. If both AccessProviders and CredentialsProviders are provided,
+	// both are used. In case they specify a provider with the same name, the one in AccessProviders is preferred.
 	CredentialsProviders []ClusterProfileCredentialsProvider `json:"credentialsProviders,omitempty"`
 }
```
tenzen-y, line 282: The accessProviders and credentialsProviders should be mutually exclusive because merge behavior is implicit behavior from the user's PoV, and the Kueue API is already Beta. We shouldn't introduce such implicit behavior to Beta. Instead of that, we should have a proper deprecation step like

1. Deprecating credentialsProviders at this time
2. (Opt) Stop evaluations in 3 or 4 minor releases later
3. Remove deprecated field in v1beta3 API

tenzen-y (commented):
Almost LGTM

keps/693-multikueue/README.md
kahirokunn: Thank you for the review! Multiple people have pointed out the same thing, and as you said, I have made the corrections as instructed. 🙏

mimowo:
/lgtm
/unhold
seeing the last comment addressed, thank you 👍

PR #11514, opened 2026-05-22
Add contributing guidelines for coding

kannon92:
#### What type of PR is this?
/kind documentaiton
/area website
/area agents

#### What this PR does / why we need it:

Add coding guidelines based on AI analysis of code base. 

#### Which issue(s) this PR fixes:

Fixes #

#### Special notes for your reviewer:

AI was used to craft the guildelines and some editing was done for brevity.

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

## Summary by CodeRabbit

* **Documentation**
  * Added a dedicated coding guidelines page covering conventions for product and test code, including controllers, APIs/webhooks, feature gates, and testing practices.
  * Updated contributor guidance links to new Kueue-specific documentation paths and reorganized the “Getting Started” resources.
  * Updated testing documentation to use `LABEL_FILTER` for integration/e2e filtering examples.
* **Chores**
  * Updated ignore rules to exclude Python bytecode artifacts (`__pycache__/`, `*.pyc`).
* **Documentation (Tests)**
  * Clarified comments for shared timeout and polling interval constants used in tests.

PR #12028, opened 2026-06-09
Add missing event message truncations

olekzabl:
#### What type of PR is this?

/kind cleanup

It adds a few missing invocations to `TruncateEventMessage` (in places where it's not evident that the message fits well within the limit).

(Triggered by reviewing #11514)

none

I didn't go through all `Eventf()` calls manually; I asked AI to do it.
So the findings could be incomplete.

```release-note
Observability: add a safeguard check truncating the event messages to make sure the events can be successfully recorded in the API server.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Event messages are now truncated to fit system limits when emitted across the system — including workload eviction, workload deactivation due to admission checks, admission-check updates, job stop/update events, and Pod composable workload warnings.


mimowo:
I think we should probably cherrypick that, it might be considered a bugfix. 

That makes me wondering about the impact - when a message exceeds the limits, does it make the  `w.recorder.Eventf(` fail causing the downstream code not to execute, or it just makes the asynchronous sending of the event to the API server fail behind the scenes?

For now I assume in the release note the latter, but let me know your findings.

/kind bug

proposal, but feel free to adjust:
/release-note-edit
```release-note
Observability: add a safeguard check truncating the event messages to make sure the events can be successfully recorded in the API server.
```

olekzabl:
> That makes me wondering about the impact - when a message exceeds the limits, does it make the `w.recorder.Eventf(` fail causing the downstream code not to execute, or it just makes the asynchronous sending of the event to the API server fail behind the scenes?
> 
> For now I assume in the release note the latter, but let me know your findings.

It indeed seems so: [code link](https://github.com/kubernetes/client-go/blob/c53b2772ec8b07cf9c7883934cd691fd90b02d41/tools/events/event_recorder.go#L93-L97)
(and, AFAICS, _all_ implementations of `Eventf` in the Kueue repo ultimately rely on this code)

mimowo:
Thanks for checking that 👍 In that case indeed the impact is only limited to "Observability".
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17

mimowo:
@olekzabl please prep a CP for 0.17

PR #12045, opened 2026-06-09
fix: cap elastic job ungating to granted quota

### LINKED ISSUE #11977: ElasticJobsViaWorkloadSlices: scale-up worker pods are ungated past quota (run outside ClusterQueue accounting)
### What happened

With the `ElasticJobsViaWorkloadSlices` feature gate enabled, an elastic
autoscaling RayCluster scaled its workers **up past its ClusterQueue's nominal
quota**, and the surplus worker pods were **ungated and Running** — while the
ClusterQueue's accounted usage stayed flat at the admitted amount. The
replacement WorkloadSlice for the larger worker count was still `Pending`
(correctly held back by quota), yet its pods had already been started.

The net effect is that the `ElasticJobUngater` lets an elastic job exceed its
ClusterQueue quota during scale-up: quota enforcement is silently bypassed for
the extra replicas.

### Environment

- Kueue **v0.18.0** (also reproduces on `main` / `release-0.18` — see below)
- KubeRay v1.6.0, Ray 2.51.1
- Feature gate: `ElasticJobsViaWorkloadSlices=true`
- An elastic RayCluster (`kueue.x-k8s.io/elastic-job: "true"`,
  `enableInTreeAutoscaling: true`) whose `worker` replicas exceed the quota
  granted to the admitted slice.

### Expected

When a scale-up mints a new WorkloadSlice whose worker count does not fit the
ClusterQueue quota, that slice stays `Pending` and its **surplus** worker pods
stay `SchedulingGated` until the replacement slice is admitted. Only the
replicas actually granted quota should ever be ungated.

### Actual

All gated pods carrying the elastic job's (chain-root) workload name are ungated
the moment **any** slice in the chain is admitted/has a quota reservation — with
no cap to the granted replica count. The surplus workers run outside quota.

### Root cause

`pkg/controller/elasticjobs/elastic_job_ungater.go`. `Reconcile` ungates **every**
pod returned by `listGatedPods`, and `listGatedPods` is uncapped:

```go
func (r *elasticJobUngater) listGatedPods(ctx context.Context, wl *kueue.Workload) ([]*corev1.Pod, error) {
	sliceName := workloadslicing.SliceName(wl)
	var podList corev1.PodList
	if err := r.client.List(ctx, &podList,
		client.InNamespace(wl.Namespace),
		client.MatchingFields{indexer.WorkloadSliceNameKey: sliceName},
	); err != nil {
		return nil, fmt.Errorf("listing pods for workload slice: %w", err)
	}

	var gated []*corev1.Pod
	for i := range podList.Items {
		if utilpod.HasGate(&podList.Items[i], kueue.ElasticJobSchedulingGate) &&
			podList.Items[i].Annotations[kueue.WorkloadAnnotation] == wl.Name {
			gated = append(gated, &podList.Items[i])
		}
	}
	return gated, nil
}
```

The list is keyed by the **chain-root** slice name (`SliceName(wl)` returns the
chain root via the `kueue.x-k8s.io/workload-slice-name` annotation, falling back
to `wl.Name`). Every pod of an elastic job — including the scale-up workers
belonging to the *replacement* slice — carries that same chain-root name in its
`kueue.x-k8s.io/workload` annotation. So when the **already-admitted root** slice
reconciles (`shouldUngate` is true for it: it's admitted), `listGatedPods`
returns *all* gated pods in the chain, and `Reconcile` ungates the entire set —
including the surplus workers whose own replacement slice is still `Pending` and
over quota.

There is nothing bounding the ungated count to the replicas the ClusterQueue
actually granted. The reconcile of the admitted slice ungates pods that belong,
quota-wise, to a slice that has not been admitted.

### Reproduction

2. Create a ClusterQueue + LocalQueue with quota for, say, head + **1** worker.
3. Create an elastic, in-tree-autoscaling RayCluster against that queue; it is
   admitted at 1 worker.
4. Drive load so the Ray autoscaler raises `worker.replicas` (e.g. to 3). Kueue
   mints a replacement WorkloadSlice for the larger count; it does **not** fit
   quota and stays `Pending`.
5. Observe: the surplus worker pods are **Running** (ungated), but the
   ClusterQueue's admitted usage is unchanged — the job is over quota.

   Expected: the surplus workers stay `SchedulingGated` until the replacement
   slice is admitted.

### Affected versions

The ungater source is byte-identical across releases — the same blob
(`ee4ac3f2…`) is on `main`, `release-0.18`, and the `v0.18.0` tag — so this is a
live bug on `main`, not only on the release.

### Proposed fix

Cap ungating, per PodSet, to the replicas actually **granted quota** across the
slice chain (the largest count among quota-reserved slices sharing the chain
root), counting already-ungated pods against the cap. Surplus pods stay gated
until their replacement slice is admitted. Sketch:

```go
// In listGatedPods, after the List:
granted, err := r.grantedPodSetCounts(ctx, wl) // max Count per podSet over quota-reserved slices in the chain
if err != nil {
	return nil, err
}

ungatedPerPodSet := make(map[string]int32)
gatedPerPodSet := make(map[string][]*corev1.Pod)
for i := range podList.Items {
	p := &podList.Items[i]
	if p.Annotations[kueue.WorkloadAnnotation] != wl.Name {
		continue
	}
	ps := p.Labels[constants.PodSetLabel]
	if utilpod.HasGate(p, kueue.ElasticJobSchedulingGate) {
		gatedPerPodSet[ps] = append(gatedPerPodSet[ps], p)
	} else {
		ungatedPerPodSet[ps]++ // already-ungated pods consume quota too
	}
}

var gated []*corev1.Pod
for ps, pods := range gatedPerPodSet {
	slices.SortFunc(pods, func(a, b *corev1.Pod) int { return strings.Compare(a.Name, b.Name) })
	room := granted[ps] - ungatedPerPodSet[ps]
	if room <= 0 {
		continue
	}
	if int32(len(pods)) > room {
		pods = pods[:room]
	}
	gated = append(gated, pods...)
}
return gated, nil
```

`grantedPodSetCounts` walks the slice chain (siblings sharing the parent job's
`OwnerReference` UID, matched to the same `SliceName`) and takes, per PodSet, the
max `Count` among slices that `HasQuotaReservation`. We have this fix running on
our cluster; with it, scale-up surplus workers stay `SchedulingGated` until the
replacement slice is admitted, and the ClusterQueue quota is respected. Happy to
open a PR.

mcochner:
#### What type of PR is this?

/kind bug
/area integrations

With `ElasticJobsViaWorkloadSlices` enabled, the `ElasticJobUngater` ungated
**every** gated pod returned by `listGatedPods`, and `listGatedPods` was
uncapped. Because every pod of an elastic job (including scale-up workers
belonging to a still-`Pending` replacement slice) carries the chain-root
workload name in its `kueue.x-k8s.io/workload` annotation, reconciling the
already-admitted root slice ungated the entire chain — including surplus
workers whose replacement slice was over quota. The result: an elastic job
could exceed its ClusterQueue's nominal quota during scale-up, with the
surplus replicas running outside quota accounting.

This PR caps ungating, per PodSet, to the replicas actually **granted quota**
across the slice chain (the max `Count` among quota-reserved slices sharing
the chain root), counting already-ungated pods against the cap. Surplus pods
stay `SchedulingGated` until their replacement slice is admitted.

- `listGatedPods` now buckets candidate pods by their PodSet label and ungates
  at most `granted[podSet] - alreadyUngated[podSet]` gated pods (lowest pod
  name first for determinism).
- A new `grantedPodSetCounts` walks the slice chain via the existing
  `OwnerReferenceUID` workload index (siblings sharing the parent job's owner
  UID and the same slice name) and takes the per-PodSet max `Count` among
  slices that hold a quota reservation.

Fixes #11977

- Unit tests added: surplus pods over quota stay gated during scale-up, and
  are ungated once the replacement slice is admitted (the chain walk raises
  the cap). Existing cases were updated to carry the PodSet label that real
  elastic pods always have.
- AI disclosure: this change was developed with the assistance of an AI coding
  tool, in line with the Kubernetes AI tool usage policy.

```release-note
ElasticJobsViaWorkloadSlices: Fixed a bug where worker pods of an elastic job could be ungated after scale up, 
past the ClusterQueue quota; ungating is now capped to the replicas granted quota across the workload-slice chain.
```

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforces per-replica-group quota caps when ungating, counts already-ungated pods toward quota, and deterministically selects which pods to ungate to avoid over-ungating.

* **Tests**
  * Added scenarios for pending replacement slices, exclusion of finished slices from caps, ungating when replacement slices gain quota, and an error case for workloads missing a controller owner; tests normalize default replica-group labels.

* **Test Config**
  * Increased CPU quota in an end-to-end test setup.

### DIFF
--- a/pkg/controller/elasticjobs/elastic_job_ungater.go
+++ b/pkg/controller/elasticjobs/elastic_job_ungater.go
@@ -79,12 +79,24 @@ func SetupWithManager(mgr ctrl.Manager, cfg *configapi.Configuration, roleTracke
 	podHandler := elasticPodHandler{
 		expectationsStore: r.expectationsStore,
 	}
+	// Reconcile by the stable slice-chain key rather than by an individual
+	// workload, so every slice in a chain (and every pod that names any slice in
+	// it) maps to a single reconcile request. Reconcile then resolves the active
+	// slice from that key.
+	sliceKeyHandler := handler.TypedEnqueueRequestsFromMapFunc(
+		func(_ context.Context, wl *kueue.Workload) []reconcile.Request {
+			return []reconcile.Request{{NamespacedName: types.NamespacedName{
+				Namespace: wl.Namespace,
+				Name:      workloadslicing.SliceName(wl),
+			}}}
+		},
+	)
 	return ControllerName, builder.TypedControllerManagedBy[reconcile.Request](mgr).
 		Named("elastic_job_ungater").
 		WatchesRawSource(source.TypedKind(
 			mgr.GetCache(),
 			&kueue.Workload{},
-			&handler.TypedEnqueueRequestForObject[*kueue.Workload]{},
+			sliceKeyHandler,
 			r,
 		)).
 		Watches(&corev1.Pod{}, &podHandler).
@@ -100,30 +112,25 @@ func (r *elasticJobUngater) Reconcile(ctx context.Context, req reconcile.Request
 	log := ctrl.LoggerFrom(ctx)
 	log.V(2).Info("Reconcile ElasticJobUngater")
 
-	wl := &kueue.Workload{}
-	if err := r.client.Get(ctx, req.NamespacedName, wl); err != nil {
-		if client.IgnoreNotFound(err) != nil {
-			return reconcile.Result{}, err
-		}
-		return reconcile.Result{}, nil
-	}
 	if !r.expectationsStore.Satisfied(log, req.NamespacedName) {
 		return reconcile.Result{}, errPendingUngateOps
 	}
-	// Ungate pods for the slice we are currently reconciling. We skip
-	// non-elastic workloads, pending workloads (no quota), and finished slices
-	// (whose pods are either rescheduled by the replacement slice or being
-	// terminated by the parent job). Skipping finished slices is what makes
-	// "the reconciled workload is the latest admitted slice" hold, so its own
-	// granted PodSet counts are the right cap for ungating.
-	if !workloadslicing.IsElasticWorkload(wl) {
-		return reconcile.Result{}, nil
+
+	// req.Name is the stable slice-chain key shared by every slice and pod in the
+	// chain (see workloadslicing.SliceName). Resolve the active (latest admitted,
+	// non-finished) slice: it is the only one whose granted PodSet counts define
+	// how many pods may be ungated. Finished and pending slices are skipped here,
+	// so the cap is always taken from the live slice regardless of which slice (or
+	// which pod's stamped WorkloadAnnotation) triggered the event.
+	active, err := r.activeSlice(ctx, req.Namespace, req.Name)
+	if err != nil {
+		return reconcile.Result{}, err
 	}
-	if workload.IsFinished(wl) || (!workload.IsAdmitted(wl) && !workload.HasQuotaReservation(wl)) {
+	if active == nil {
 		return reconcile.Result{}, nil
 	}
 
-	pods, err := r.podsToUngate(ctx, wl)
+	pods, err := r.podsToUngate(ctx, active)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
@@ -168,8 +175,8 @@ func (r *elasticJobUngater) podsToUngate(ctx context.Context, wl *kueue.Workload
 	// so the index lookup returns every pod created on behalf of this job.
 	// Although those pods may still carry an older slice's name in their
 	// WorkloadAnnotation (the template is stamped at the slice's admission),
-	// the reconcile guards ensure wl is the currently-admitted slice, so its
-	// granted PodSet counts are the right cap for ungating any of them.
+	// wl is the chain's active slice (resolved in Reconcile), so its granted
+	// PodSet counts are the right cap for ungating any of them.
 	sliceName := workloadslicing.SliceName(wl)
 	var podList corev1.PodList
 	if err := r.client.List(ctx, &podList,
@@ -207,6 +214,39 @@ func (r *elasticJobUngater) podsToUngate(ctx context.Context, wl *kueue.Workload
 		gated = append(gated, pods...)
 	}
 	return gated, nil
+}
+
+// activeSlice returns the active (latest admitted, non-finished) workload slice
+// for the chain identified by sliceName, or nil if none qualifies.
+func (r *elasticJobUngater) activeSlice(ctx context.Context, namespace, sliceName string) (*kueue.Workload, error) {
+	var list kueue.WorkloadList
+	if err := r.client.List(ctx, &list,
+		client.InNamespace(namespace),
+		client.MatchingFields{indexer.WorkloadSliceNameWorkloadKey: sliceName},
+	); err != nil {
+		return nil, fmt.Errorf("listing workload slices: %w", err)
+	}
+	var active *kueue.Workload
+	for i := range list.Items {
+		wl := &list.Items[i]
+		if !shouldUngate(wl) {
+			continue
+		}
+		if active == nil || moreActiveSlice(wl, active) {
+			active = wl
+		}
+	}
+	return active, nil
+}
+
+// moreActiveSlice reports whether slice a is the better cap source than b: the
+// newer slice wins, and on equal creation timestamps the replacement slice
+// (carrying WorkloadSliceReplacementFor) is treated as newer than the root.
+func moreActiveSlice(a, b *kueue.Workload) bool {
+	if c := a.CreationTimestamp.Compare(b.CreationTimestamp.Time); c != 0 {
+		return c > 0
+	}
+	return workloadslicing.ScaledUp(a) && !workloadslicing.ScaledUp(b)
 }
 
 // Workload predicates
@@ -261,18 +301,32 @@ func (h *elasticPodHandler) queueReconcileForPod(ctx context.Context, object cli
 	if !isPod {
 		return
 	}
-	wlName, found := pod.Annotations[kueue.WorkloadAnnotation]
-	if !found {
+	// Enqueue by the stable slice-chain key, not the pod's stamped
+	// WorkloadAnnotation. A pod minted after a scale-up still carries the previous
+	// (now Finished) slice's name, but the chain key is shared by every slice, so
+	// Reconcile can always resolve the active slice from it.
+	sliceName := podSliceName(pod)
+	if sliceName == "" {
 		return
 	}
 	key := types.NamespacedName{
-		Name:      wlName,
+		Name:      sliceName,
 		Namespace: pod.Namespace,
 	}
 	// Mark expectation as observed when the gate has been removed or the pod is deleted.
 	if !utilpod.HasGate(pod, kueue.ElasticJobSchedulingGate) || deleted {
-		log := ctrl.LoggerFrom(ctx).WithValues("pod", klog.KObj(pod), "workload", key.String())
+		log := ctrl.LoggerFrom(ctx).WithValues("pod", klog.KObj(pod), "workloadSlice", key.String())
 		h.expectationsStore.ObservedUID(log, key, pod.UID)
 	}
 	q.AddAfter(reconcile.Request{NamespacedName: key}, constants.UpdatesBatchPeriod)
+}
+
+// podSliceName returns the slice-chain key for a pod: the WorkloadSliceName
+// annotation if present, otherwise the stamped Workload annotation. Mirrors
+// indexer.IndexPodWorkloadSliceName so the key matches the pod index.
+func podSliceName(pod *corev1.Pod) string {
+	if v, found := pod.Annotations[kueue.WorkloadSliceNameAnnotation]; found {
+		return v
+	}
+	return pod.Annotations[kueue.WorkloadAnnotation]
 }
--- a/pkg/controller/elasticjobs/elastic_job_ungater.go
+++ b/pkg/controller/elasticjobs/elastic_job_ungater.go
@@ -20,8 +20,11 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
@@ -168,14 +171,88 @@ func (r *elasticJobUngater) listGatedPods(ctx context.Context, wl *kueue.Workloa
 		return nil, fmt.Errorf("listing pods for workload slice: %w", err)
 	}
 
-	var gated []*corev1.Pod
+	// Cap ungating, per PodSet, to the replicas actually granted quota across the
+	// slice chain. Without this cap, scale-up surplus pods (which carry the
+	// chain-root workload name, like the admitted slice's pods) would be ungated
+	// even though their replacement slice has not been admitted, bypassing the
+	// ClusterQueue quota.
+	granted, err := r.grantedPodSetCounts(ctx, wl)
+	if err != nil {
+		return nil, err
+	}
+
+	gatedPerPodSet := make(map[kueue.PodSetReference][]*corev1.Pod)
+	ungatedPerPodSet := make(map[kueue.PodSetReference]int32)
 	for i := range podList.Items {
-		if utilpod.HasGate(&podList.Items[i], kueue.ElasticJobSchedulingGate) &&
-			podList.Items[i].Annotations[kueue.WorkloadAnnotation] == wl.Name {
-			gated = append(gated, &podList.Items[i])
+		p := &podList.Items[i]
+		if p.Annotations[kueue.WorkloadAnnotation] != wl.Name {
+			continue
+		}
+		ps := kueue.PodSetReference(p.Labels[constants.PodSetLabel])
+		if utilpod.HasGate(p, kueue.ElasticJobSchedulingGate) {
+			gatedPerPodSet[ps] = append(gatedPerPodSet[ps], p)
+		} else {
+			// Already-ungated pods consume quota too.
+			ungatedPerPodSet[ps]++
 		}
 	}
+
+	var gated []*corev1.Pod
+	for ps, pods := range gatedPerPodSet {
+		room := granted[ps] - ungatedPerPodSet[ps]
+		if room <= 0 {
+			continue
+		}
+		// Ungate the lowest-named pods first for deterministic behavior.
+		slices.SortFunc(pods, func(a, b *corev1.Pod) int { return strings.Compare(a.Name, b.Name) })
+		if int32(len(pods)) > room {
+			pods = pods[:room]
+		}
+		gated = append(gated, pods...)
+	}
 	return gated, nil
+}
+
+// grantedPodSetCounts returns, per PodSet, the maximum replica Count among the
+// workload slices in wl's chain that hold a quota reservation. The chain is
+// identified by slices owned by the same parent job (matched via the
+// OwnerReferenceUID index) that share the same slice name.
+func (r *elasticJobUngater) grantedPodSetCounts(ctx context.Context, wl *kueue.Workload) (workload.PodSetsCounts, error) {
+	granted := workload.PodSetsCounts{}
+	mergeMax := func(w *kueue.Workload) {
+		if !workload.HasQuotaReservation(w) {
+			return
+		}
+		for ps, count := range workload.ExtractPodSetCountsFromWorkload(w) {
+			if existing, ok := granted[ps]; !ok || count > existing {
+				granted[ps] = count
+			}
+		}
+	}
+
+	mergeMax(wl)
+
+	owner := metav1.GetControllerOf(wl)
+	if owner == nil {
+		return granted, nil
+	}
+
+	sliceName := workloadslicing.SliceName(wl)
+	var wlList kueue.WorkloadList
+	if err := r.client.List(ctx, &wlList,
+		client.InNamespace(wl.Namespace),
+		client.MatchingFields{indexer.OwnerReferenceUID: string(owner.UID)},
+	); err != nil {
+		return nil, fmt.Errorf("listing slice chain workloads: %w", err)
+	}
+	for i := range wlList.Items {
+		sibling := &wlList.Items[i]
+		if sibling.Name == wl.Name || workloadslicing.SliceName(sibling) != sliceName {
+			continue
+		}
+		mergeMax(sibling)
+	}
+	return granted, nil
 }
 
 // Workload predicates
--- a/pkg/controller/elasticjobs/elastic_job_ungater.go
+++ b/pkg/controller/elasticjobs/elastic_job_ungater.go
@@ -24,6 +24,8 @@ import (
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
@@ -117,12 +119,17 @@ func (r *elasticJobUngater) Reconcile(ctx context.Context, req reconcile.Request
 	}
 
 	// req.Name is the stable slice-chain key shared by every slice and pod in the
-	// chain (see workloadslicing.SliceName). Resolve the active (latest admitted,
-	// non-finished) slice: it is the only one whose granted PodSet counts define
-	// how many pods may be ungated. Finished and pending slices are skipped here,
-	// so the cap is always taken from the live slice regardless of which slice (or
-	// which pod's stamped WorkloadAnnotation) triggered the event.
-	active, err := r.activeSlice(ctx, req.Namespace, req.Name)
+	// chain (see workloadslicing.SliceName); it is the name of the chain's root
+	// slice. Load it to find the owning job, then resolve the active (latest
+	// admitted, non-finished) slice from the job's slice chain: it is the only
+	// one whose granted PodSet counts define how many pods may be ungated, so the
+	// cap is always taken from the live slice regardless of which slice (or which
+	// pod's stamped WorkloadAnnotation) triggered the event.
+	root := &kueue.Workload{}
+	if err := r.client.Get(ctx, req.NamespacedName, root); err != nil {
+		return reconcile.Result{}, client.IgnoreNotFound(err)
+	}
+	active, err := r.activeSlice(ctx, root)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
@@ -216,37 +223,19 @@ func (r *elasticJobUngater) podsToUngate(ctx context.Context, wl *kueue.Workload
 	return gated, nil
 }
 
-// activeSlice returns the active (latest admitted, non-finished) workload slice
-// for the chain identified by sliceName, or nil if none qualifies.
-func (r *elasticJobUngater) activeSlice(ctx context.Context, namespace, sliceName string) (*kueue.Workload, error) {
-	var list kueue.WorkloadList
-	if err := r.client.List(ctx, &list,
-		client.InNamespace(namespace),
-		client.MatchingFields{indexer.WorkloadSliceNameWorkloadKey: sliceName},
-	); err != nil {
-		return nil, fmt.Errorf("listing workload slices: %w", err)
+// activeSlice resolves the active (latest admitted, non-finished) workload slice
+// of the chain that anyWl belongs to, or nil if none qualifies. It looks up the
+// chain through the owning job's workload index, reusing the same slice ordering
+// as the rest of the slicing code (workloadslicing.FindLatestActiveWorkload).
+func (r *elasticJobUngater) activeSlice(ctx context.Context, anyWl *kueue.Workload) (*kueue.Workload, error) {
+	owner := metav1.GetControllerOf(anyWl)
+	if owner == nil {
+		return nil, nil
 	}
-	var active *kueue.Workload
-	for i := range list.Items {
-		wl := &list.Items[i]
-		if !shouldUngate(wl) {
-			continue
-		}
-		if active == nil || moreActiveSlice(wl, active) {
-			active = wl
-		}
+	jobObject := &metav1.PartialObjectMetadata{
+		ObjectMeta: metav1.ObjectMeta{Namespace: anyWl.Namespace, Name: owner.Name},
 	}
-	return active, nil
-}
-
-// moreActiveSlice reports whether slice a is the better cap source than b: the
-// newer slice wins, and on equal creation timestamps the replacement slice
-// (carrying WorkloadSliceReplacementFor) is treated as newer than the root.
-func moreActiveSlice(a, b *kueue.Workload) bool {
-	if c := a.CreationTimestamp.Compare(b.CreationTimestamp.Time); c != 0 {
-		return c > 0
-	}
-	return workloadslicing.ScaledUp(a) && !workloadslicing.ScaledUp(b)
+	return workloadslicing.FindLatestActiveWorkload(ctx, r.client, jobObject, schema.FromAPIVersionAndKind(owner.APIVersion, owner.Kind))
 }
 
 // Workload predicates
--- a/pkg/controller/elasticjobs/elastic_job_ungater.go
+++ b/pkg/controller/elasticjobs/elastic_job_ungater.go
@@ -214,13 +214,16 @@ func (r *elasticJobUngater) listGatedPods(ctx context.Context, wl *kueue.Workloa
 }
 
 // grantedPodSetCounts returns, per PodSet, the maximum replica Count among the
-// workload slices in wl's chain that hold a quota reservation. The chain is
-// identified by slices owned by the same parent job (matched via the
+// workload slices in wl's chain that hold an active quota reservation. The
+// chain is identified by slices owned by the same parent job (matched via the
 // OwnerReferenceUID index) that share the same slice name.
 func (r *elasticJobUngater) grantedPodSetCounts(ctx context.Context, wl *kueue.Workload) (workload.PodSetsCounts, error) {
 	granted := workload.PodSetsCounts{}
 	mergeMax := func(w *kueue.Workload) {
-		if !workload.HasQuotaReservation(w) {
+		// Finished (e.g. replaced) slices retain QuotaReserved=True, so require an
+		// active reservation to avoid inflating the cap from stale slices after a
+		// scale-up-then-scale-down cycle.
+		if !workload.HasActiveQuotaReservation(w) {
 			return
 		}
 		for ps, count := range workload.ExtractPodSetCountsFromWorkload(w) {
--- a/pkg/controller/elasticjobs/elastic_job_ungater.go
+++ b/pkg/controller/elasticjobs/elastic_job_ungater.go
@@ -56,6 +56,7 @@ import (
 const ControllerName = "ElasticJobUngater"
 
 var errPendingUngateOps = errors.New("pending elastic ungate operations")
+var errMissingControllerOwner = errors.New("elastic workload slice has no controller owner reference")
 
 type elasticJobUngater struct {
 	client            client.Client
@@ -235,11 +236,19 @@ func (r *elasticJobUngater) grantedPodSetCounts(ctx context.Context, wl *kueue.W
 
 	mergeMax(wl)
 
+	// Elastic workload slices are always created with a controller owner
+	// reference to their job; without it the slice chain cannot be resolved.
 	owner := metav1.GetControllerOf(wl)
 	if owner == nil {
-		return granted, nil
+		return nil, fmt.Errorf("%w: %s", errMissingControllerOwner, klog.KObj(wl))
 	}
 
+	// Surplus scale-up pods keep the chain-root workload name in their
+	// kueue.WorkloadAnnotation, while the quota granted for them lives on the
+	// replacement slice. Since podsToUngate only matches pods annotated with
+	// wl.Name, the reconcile of the root slice must consult its sibling slices'
+	// reservations; otherwise, once the replacement is admitted (finishing the
+	// root slice), the surplus pods would stay gated forever.
 	sliceName := workloadslicing.SliceName(wl)
 	var wlList kueue.WorkloadList
 	if err := r.client.List(ctx, &wlList,
--- a/pkg/controller/elasticjobs/elastic_job_ungater.go
+++ b/pkg/controller/elasticjobs/elastic_job_ungater.go
@@ -207,18 +207,27 @@ func (r *elasticJobUngater) podsToUngate(ctx context.Context, wl *kueue.Workload
 		}
 	}
 
+	log := ctrl.LoggerFrom(ctx)
 	var gated []*corev1.Pod
-	for ps, pods := range gatedPerPodSet {
+	for ps, candidates := range gatedPerPodSet {
 		room := granted[ps] - ungatedPerPodSet[ps]
-		if room <= 0 {
-			continue
+		var toUngate []*corev1.Pod
+		if room > 0 {
+			// Ungate the lowest-named pods first for deterministic behavior.
+			slices.SortFunc(candidates, func(a, b *corev1.Pod) int { return strings.Compare(a.Name, b.Name) })
+			toUngate = candidates
+			if int32(len(candidates)) > room {
+				toUngate = candidates[:room]
+			}
 		}
-		// Ungate the lowest-named pods first for deterministic behavior.
-		slices.SortFunc(pods, func(a, b *corev1.Pod) int { return strings.Compare(a.Name, b.Name) })
-		if int32(len(pods)) > room {
-			pods = pods[:room]
-		}
-		gated = append(gated, pods...)
+		log.V(4).Info("elastic ungating quota accounting for PodSet",
+			"podSet", ps,
+			"grantedCount", granted[ps],
+			"alreadyUngatedCount", ungatedPerPodSet[ps],
+			"gatedCount", len(candidates),
+			"ungatingCount", len(toUngate),
+		)
+		gated = append(gated, toUngate...)
 	}
 	return gated, nil
 }
--- a/pkg/controller/elasticjobs/elastic_job_ungater.go
+++ b/pkg/controller/elasticjobs/elastic_job_ungater.go
@@ -170,7 +170,7 @@ func (r *elasticJobUngater) Reconcile(ctx context.Context, req reconcile.Request
 		if !ungated {
 			r.expectationsStore.ObservedUID(log, req.NamespacedName, pod.UID)
 		} else {
-			utilpod.RecordPodSchedulingGateRemovalSeconds(r.clock, kueue.ElasticJobSchedulingGate, wl, false)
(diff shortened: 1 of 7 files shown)


kannon92:
Please fix the CLA. You may have to remove the commit with cursor.

mimowo:
cc @sohankunkerkar @yaroslava-serdiuk ptal

mcochner:
**Why the e2e test change in fb91733** (analysis of the `pull-kueue-test-e2e-extended-shard-0` failures on the first push):

All four shard-0 jobs (k8s 1.33–1.36) failed deterministically on the same spec, `Kuberay: Should run a rayjob with multi scale-up steps`, timing out waiting for 5 running worker pods.

The expectation of 5 running workers is not reachable within the suite's quota: head (1 CPU) + submitter (400m) + 5 workers (5×400m) = **3.4 CPU**, but the shared ClusterQueue had **3 CPU**. A 5-worker workload slice can never be admitted under that quota.

Comparing the kueue-controller-manager logs of the failed run against a passing run of the same job from another PR (built on unmodified `main`) shows both runs behaved identically on the scheduler side:

- the 5-worker replacement slice was rejected in both runs with `insufficient unused quota for cpu ... 400m more needed` and requeued with `PreemptionNoCandidates`;
- in both runs that slice was only admitted after Ray's autoscaler downsized it back to 3 workers once the demand window passed — **the ClusterQueue never granted 5 workers in either run**.

The only difference: on `main`, the ungater ungated the 2 surplus worker pods while their slice was still `Pending` (3 admitted + 2 running outside quota — exactly the bug from #11977), which is what satisfied the test's "5 running workers" assertion. With the fix, those pods correctly stay gated, so the assertion can no longer be met under a 3-CPU quota.

Raising the quota to 4 CPU makes the 5-worker slice (3.4 CPU) admissible via the normal slice-replacement path — the same path that admitted the 1→3 scale-up within a second in both runs — so the test now verifies legitimate, quota-respecting elastic scale-up. The sibling tests sharing this ClusterQueue are demand-driven (Ray task parallelism caps their worker counts, not quota), so they are unaffected.

(AI disclosure: this investigation and comment were prepared with the assistance of an AI coding tool.)

pkg/controller/elasticjobs/elastic_job_ungater.go
yaroslava-serdiuk: I suppose we should rename it to something like "podsToUngate"

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -159,16 +162,90 @@ func (r *elasticJobUngater) listGatedPods(ctx context.Context, wl *kueue.Workloa
...
 }
 
+// grantedPodSetCounts returns, per PodSet, the maximum replica Count among the
+// workload slices in wl's chain that hold a quota reservation. The chain is
+// identified by slices owned by the same parent job (matched via the
+// OwnerReferenceUID index) that share the same slice name.
+func (r *elasticJobUngater) grantedPodSetCounts(ctx context.Context, wl *kueue.Workload) (workload.PodSetsCounts, error) {
+	granted := workload.PodSetsCounts{}
+	mergeMax := func(w *kueue.Workload) {
+		if !workload.HasQuotaReservation(w) {
```
sohankunkerkar: I think you should use `workload.HasActiveQuotaReservation(w)` instead bc finished/replaced slices retain `QuotaReserved=True`, so without the !IsFinished check, a scale-up→scale-down cycle inflates the cap from stale slices.

Please also add a test case for the scale-up-then-scale-down scenario to cover this.

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -159,16 +162,93 @@ func (r *elasticJobUngater) listGatedPods(ctx context.Context, wl *kueue.Workloa
...
+				granted[ps] = count
+			}
+		}
+	}
+
+	mergeMax(wl)
+
+	owner := metav1.GetControllerOf(wl)
+	if owner == nil {
+		return granted, nil
```
yaroslava-serdiuk: I think we should return err in this case.

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -159,16 +162,93 @@ func (r *elasticJobUngater) listGatedPods(ctx context.Context, wl *kueue.Workloa
...
+		client.MatchingFields{indexer.OwnerReferenceUID: string(owner.UID)},
+	); err != nil {
+		return nil, fmt.Errorf("listing slice chain workloads: %w", err)
+	}
+	for i := range wlList.Items {
+		sibling := &wlList.Items[i]
+		if sibling.Name == wl.Name || workloadslicing.SliceName(sibling) != sliceName {
+			continue
+		}
+		mergeMax(sibling)
```
yaroslava-serdiuk: Why do we need this?

pkg/controller/elasticjobs/elastic_job_ungater.go
mcochner: Done in 47237a07e — `grantedPodSetCounts` now returns an error when the slice has no controller owner reference, with a unit test case covering it.

(Response drafted with AI assistance.)

pkg/controller/elasticjobs/elastic_job_ungater.go
mcochner: The surplus pods created during scale-up carry the chain-root workload's name in `kueue.WorkloadAnnotation` (they are minted from the pod template stamped at the root slice's admission), while the quota granted for them lives on the replacement slice. Since `podsToUngate` only matches pods annotated with `wl.Name`, it is the reconcile of the *root* slice that has to ungate them once the replacement is admitted — and to know how many it may ungate, it must look at the sibling slices' active reservations.

Without the sibling walk, the cap would be computed from the root slice alone, which loses its active reservation when it is replaced, so the surplus pods would stay gated forever after the replacement is admitted. The `ungate surplus once replacement admitted` unit test covers exactly this scenario.

I expanded the code comment above this loop in 47237a07e to capture this reasoning.

(Response drafted with AI assistance.)

yaroslava-serdiuk: The comment says: `otherwise, once the replacement is admitted (finishing the root slice), the surplus pods would stay gated forever`
Have you noticed that in manual test? 
I suppose when Kueue reconcile the admitted workload slice, the correct number of podSets will be obtained in line 228. I still don't understand why we need iterate over sibling here.

test/e2e/singlecluster/extended/kuberay_test.go
```diff
@@ -126,7 +126,7 @@ var _ = ginkgo.Describe("Kuberay", ginkgo.Label("area:singlecluster", "feature:k
 		cq = utiltestingapi.MakeClusterQueue(clusterQueueName).
 			ResourceGroup(
 				*utiltestingapi.MakeFlavorQuotas(rf.Name).
-					Resource(corev1.ResourceCPU, "3").Obj()).
+					Resource(corev1.ResourceCPU, "4").Obj()).
```
mimowo: Why is this changed? It does not seem related in any obvious way

mcochner: Thank you, for your questions and the reviews! I have noted the reasoning for this change in the commit message in the commit [2532f228fe17f7bd01cf2c89de799d90476f640b](https://github.com/mcochner/kueue/commit/2532f228fe17f7bd01cf2c89de799d90476f640b) 

I am copy/pasting it here for better visibility: 

> The multi scale-up e2e test expects 5 running workers, which needs
> 3.4 CPU (head 1 + submitter 400m + 5x400m workers) but the ClusterQueue
> only had 3 CPU. The test previously passed only because surplus pods
> were ungated past quota (the bug fixed in this PR). With 4 CPU the
> 5-worker slice is admitted legitimately via slice replacement.

mimowo:
/assign
I will try to review EOW, in the meanwhile another pass from @sohankunkerkar or @yaroslava-serdiuk would be great

test/e2e/singlecluster/extended/kuberay_test.go
yaroslava-serdiuk: Could you add an integration test that would verify that surplus pods remaining scheduling gated? So the test should fail without your changes and pass with the fix.

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -159,16 +163,101 @@ func (r *elasticJobUngater) listGatedPods(ctx context.Context, wl *kueue.Workloa
...
+		}
+	}
+
+	mergeMax(wl)
+
+	// Elastic workload slices are always created with a controller owner
+	// reference to their job; without it the slice chain cannot be resolved.
+	owner := metav1.GetControllerOf(wl)
+	if owner == nil {
+		return nil, fmt.Errorf("%w: %s", errMissingControllerOwner, klog.KObj(wl))
```
sohankunkerkar: If there's no controller owner, fall back to ungating only from the current workload's count instead of erroring. It avoids infinite requeue for orphaned or manually created workloads.

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -159,16 +163,101 @@ func (r *elasticJobUngater) listGatedPods(ctx context.Context, wl *kueue.Workloa
 		return nil, fmt.Errorf("listing pods for workload slice: %w", err)
 	}
 
-	var gated []*corev1.Pod
+	// Cap ungating, per PodSet, to the replicas actually granted quota across the
+	// slice chain. Without this cap, scale-up surplus pods (which carry the
+	// chain-root workload name, like the admitted slice's pods) would be ungated
+	// even though their replacement slice has not been admitted, bypassing the
+	// ClusterQueue quota.
+	granted, err := r.grantedPodSetCounts(ctx, wl)
```
sohankunkerkar: can you move this after the pod classification loop and early-returning when `len(gatedPerPodSet) == 0`?. I think it will avoid a List API call on every reconcile when there are no gated pods.

pkg/controller/elasticjobs/elastic_job_ungater.go
sohankunkerkar, line 265: I think finished workloads with `QuotaReserved=True` currently pass and get reconciled every time for nothing. Adding `!IsFinished` skips them.

pkg/controller/elasticjobs/elastic_job_ungater.go
mimowo: > (Response drafted with AI assistance.)

@mcochner, why it is ok to use AI for coding, and responses, I would advise using AI differently: instead of drafting the response, use AI for brainstorming, knowledge collection etc. Then write the response by yourself and again use AI to iterate on it (eg improve on the grammar). See also: https://www.kubernetes.dev/docs/guide/pull-requests/#ai-guidance. Drafting to me is already a substantial part of authorship, editorial corrections less so, tbh.

It may seem like a tiny tweak, but I believe it will improve the reviewers'confidence in the PR - show that you  stand behind the changes, and understand the proposal.

That being said, I'm also using AI these days (and most of the scene), and this is perfectly fine, we are trying to find out how to communicate effectively in the new era.

yaroslava-serdiuk:
I just recently hit the original issue in my test. Really looking forward for this fix!

testinfected:
Eager to get this fix too, we're also impacted

pkg/controller/elasticjobs/elastic_job_ungater.go
mimowo: Let's ignore my previous comment and go to the merit - I didn't have time to review that before.

I agree here with @yaroslava-serdiuk comment  - this should not be needed, and moreover - this is buggy (IIUC). 

You are correct here: "The surplus pods created during scale-up carry the chain-root workload's name in kueue.WorkloadAnnotation".

However, when we build the list of pods to ungate we don't search them by "kueue.WorkloadAnnotation", but by the generic "WorkloadSliceNameKey", see [here](https://github.com/kubernetes-sigs/kueue/pull/12045/changes#diff-41ddd3064525f2b6f99b4178832975ef69dd96e3e7f996d784ef10c70c37ec78R158-R164). So, we have the complete list of Pods, also corresponding to the previous workloads in the chain (previous WorkloadSlices).

Now, when you calculate quota, then only the **last admitted workload should matter**. To determine the last admitted workload we need to sort the list of workloads as done here: https://github.com/kubernetes-sigs/kueue/blob/main/pkg/workloadslicing/workloadslicing.go#L111-L137, so let's extract this function as a helper, OrderNonFinishedWorkloads().

Then we can use the code https://github.com/kubernetes-sigs/kueue/blob/main/pkg/workloadslicing/workloadslicing.go#L258-L272 to find the "latestAdmitted" - please again extract to a helper "FindLatestAdmitted" and reuse.

**Now: why there is a problem?**

Because by doing mergeMax we may assume too much quota when a workload was scaled down. So, assume we have had the following history of workload sizes:
100 -> 1 -> 40 -> ...

Because we had 100 in the past then now effectively we always assume "100" quota.

cc @sohankunkerkar @yaroslava-serdiuk in case I missed something, but that seems to me more robuts.

pkg/controller/elasticjobs/elastic_job_ungater.go
mimowo: Since this is pre-exiting code I wouldn't worry about this as part of the PR, but would you like to maybe send a separate PR?

pkg/controller/elasticjobs/elastic_job_ungater.go
mimowo: I'm not sure I'm following this comment, which List API call you think we can save? We need to list all Pods in order to determine how many of them are gated.

sohankunkerkar: Not the Pod List... I'm talking about the Workload List inside `grantedPodSetCounts` (the client.List with MatchingFields{indexer.OwnerReferenceUID} to resolve the sibling chain). If we move grantedPodSetCounts after the pod classification loop and early-return when `len(gatedPerPodSet) == 0`, we skip that Workload List call entirely on every reconcile where there are no gated pods. The Pod List has to stay where it is.

pkg/controller/elasticjobs/elastic_job_ungater.go
mimowo: Actually, I think we shouldn't complicate code for such cases, unless we can find they are occuring naturally in business relevant cases. Performing actions on orphaned workloads is generally unsafe and we have code to mark such workloads as Finished, with the FinishOrphanedWorkloads FG (still Beta). 

Doing business logic on such workloads is very tricky - some users have easier task do harm to the cluster exploiting vulnerabilities. 

Otherwise we need to extend testing, but we really don't want to support orphaned workloads.

pkg/controller/elasticjobs/elastic_job_ungater.go
sohankunkerkar: sure, will do that

pkg/controller/elasticjobs/elastic_job_ungater.go
mimowo: Actually, I'm not sure if we need to determine even the last admitted workload. We can basically rely on the workload we are reconciling for. So, I agree fully with @yaroslava-serdiuk we can just drop the code, and skip the refactorings I was discussing in my previous comment. We know this one of the latest admitted workloads, because we check if already in the Reconcile guards.

pkg/controller/elasticjobs/elastic_job_ungater.go
mimowo: Ah, ok, but that should not be needed at all. We don't need to list old workloads, just take the current, as pointed out in this thread: https://github.com/kubernetes-sigs/kueue/pull/12045/changes#r3393983204

pkg/controller/elasticjobs/elastic_job_ungater.go
sohankunkerkar: @mimowo Just to clarify, are you suggesting to drop the entire `grantedPodSetCounts` sibling walk and use only wl's own pod set counts as the cap?

mimowo: Yes, read the grantedPodSetCounts based on the capacity provided by the currently reconciled workload.

mimowo:
@mcochner let us know if you have cycles to address the remaining comments. I'm now fully onboarded to the review.

pkg/controller/elasticjobs/elastic_job_ungater.go
sohankunkerkar: Dropping the sibling walk would break the ungate surplus once replacement admitted case. podsToUngate() lists by WorkloadSliceNameKey but then filters to `p.Annotations[kueue.WorkloadAnnotation] == wl.Name`. Surplus pods from scale-up still carry the root workload's annotation because the pod template is only stamped via` startJob()/RunWithPodSetsInfo()`, which only runs while the job is suspended as elastic scale-up happens while the job is already running. 

So the replacement slice's reconcile filters out the surplus pods (wrong WorkloadAnnotation), and the finished root's reconcile is the only one that can see them. Without the sibling walk, the root has no active quota → cap = 0 → surplus pods stay gated forever.

mimowo: Yes and no. This is not needed if we drop this filter  p.Annotations[kueue.WorkloadAnnotation] == wl.Name

Sorry for not being clear enough that we should drop it. Basically we search all pods based on the slice name, but ungater only based on the current workload

mimowo:
@mcochner do you still have an interest in finalizing the PR? I think this request https://github.com/kubernetes-sigs/kueue/pull/12045#discussion_r3393983204 is pretty important for the long term maintainability. Otherwise we may need to look for someone to take over, as we plan to release this Thursday, and it would be great to include this bugfix.

mcochner:
Hi @mimowo, thank you for the patience and guidance regarding the usage of AI as well :-) ...  I cannot promise that I will finish everything this but I will try my best today/tomorrow.

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -159,16 +177,69 @@ func (r *elasticJobUngater) listGatedPods(ctx context.Context, wl *kueue.Workloa
...
+}
+
+// moreActiveSlice reports whether slice a is the better cap source than b: the
+// newer slice wins, and on equal creation timestamps the replacement slice
+// (carrying WorkloadSliceReplacementFor) is treated as newer than the root.
+func moreActiveSlice(a, b *kueue.Workload) bool {
+	if c := a.CreationTimestamp.Compare(b.CreationTimestamp.Time); c != 0 {
+		return c > 0
+	}
+	return workloadslicing.ScaledUp(a) && !workloadslicing.ScaledUp(b)
```
mimowo: We already solve the problem of comparing the slices here https://github.com/kubernetes-sigs/kueue/blob/main/pkg/workloadslicing/workloadslicing.go#L105-L137

I would like to commonize the functions

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -95,27 +109,25 @@ func (r *elasticJobUngater) Reconcile(ctx context.Context, req reconcile.Request
...
-	if !workloadslicing.IsElasticWorkload(wl) {
-		return reconcile.Result{}, nil
+
+	// req.Name is the stable slice-chain key shared by every slice and pod in the
+	// chain (see workloadslicing.SliceName). Resolve the active (latest admitted,
+	// non-finished) slice: it is the only one whose granted PodSet counts define
+	// how many pods may be ungated. Finished and pending slices are skipped here,
+	// so the cap is always taken from the live slice regardless of which slice (or
+	// which pod's stamped WorkloadAnnotation) triggered the event.
+	active, err := r.activeSlice(ctx, req.Namespace, req.Name)
```
mimowo: Why exactly we need to check the activeSlice rather than just the current? Using the current would be much simpler algo.

mimowo: Alternatively we could just skip if we follow the replacement-for annotation, and check if the immediate replacement is admitted or not - if admitted then we can skip this Reconcile, becuase we know it will be handled by the Reconcile for the follower workload.

pkg/controller/elasticjobs/elastic_job_ungater.go
mcochner: Yes, I agree. This is an clear+easy improvement, let's implement that as soon as we clear the more involved questions.

pkg/controller/elasticjobs/elastic_job_ungater.go
mcochner: > Why exactly we need to check the activeSlice rather than just the current? Using the current would be much simpler algo.

Reconcile is triggered from two sources:

**Workload events** — here you're completely right. When the replacement slice is admitted, its own event fires and req already points at the active slice. No resolution needed.

**Pod events** — this is the problem case. The pod handler enqueues based on the pod's kueue.x-k8s.io/workload annotation, which is stamped from the PodTemplate at the slice's admission. A worker pod that the autoscaler mints after a scale-up still carries the previous (now Finished) slice's name. So the event enqueues the Finished root, and "just use the current" means current = Finished root → the guard bails → nothing ever re-reconciles the active replacement → the pod stays scheduling-gated forever. For this case please see the test added in ~https://github.com/kubernetes-sigs/kueue/pull/12045/changes/ebb11487c1fee40ee35c048f8035796b95db8218~ [6e0fcd81b](https://github.com/kubernetes-sigs/kueue/pull/12045/changes/6e0fcd81b07277bc3b2f7d09e3baf600a7c3caf4) (edited)

mimowo: So, I can perfectly well see why serching for the last active is correct, but I think just ungating for the "current admitted" workload will effectively result in the same results, because all old workloads are quickly marked finished.

mcochner: I believe the integration test I added — Should ungate a scale-up worker pod minted after the replacement slice was admitted (in [6e0fcd81b](https://github.com/kubernetes-sigs/kueue/pull/12045/changes/6e0fcd81b07277bc3b2f7d09e3baf600a7c3caf4)) — is a faithful, deterministic reproduction of the failing e2e (e.g. [pull-kueue-test-e2e-extended-shard-2](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12045/pull-kueue-test-e2e-extended-shard-2-main-1-34/2069803592126042112)). It fails on the current code for the same reason the e2e times out, and passes once the late pod's event is routed to the active slice.

I'll be honest that progress here has felt slower than I'd like. The core complication is that the worker pods RayCluster creates carry only the root slice's kueue.x-k8s.io/workload annotation, not the active slice's — so a pod minted after a scale-up points at an already-finished workload, which is what makes the ungate routing tricky.

Given that, would it make sense to bring in someone with deeper Go / Kubernetes / Kueue experience?

mimowo: > The core complication is that the worker pods RayCluster creates carry only the root slice's kueue.x-k8s.io/workload annotation, not the active slice's — so a pod minted after a scale-up points at an already-finished workload, which is what makes the ungate routing tricky.

Ahh, I understand now - this is indeed a bit of a problem, and now I understand the issue with my approach of using the current "workload" annotation. This is a very useful note 👍 

I think we may want to fix this, but let me now re-consider the options we have so that we don't create a complex chain of dependencies. The approach of finding the latest active workload by the slice seems like a reasonable option. I'm putting high in ToDo to recommend some option here.

cc @sohankunkerkar

mimowo: Hmmm. but let me think again we have two phases:
1. we have reconcile for an admitted workload, for the workload we can read the SliceName from the annotation
2. we identify all Pods by the indexer, based on the SliceName annotation on the Pods
3. for the workload we are currently reconciling we only ungate the certain number of Pods.

I think this flow will still work ok even with the issue you are pointing, and is much simpler - no extra indexing. wdyt?

mcochner: IIUC I (and my agent) believe that 
> "we have reconcile for an admitted workload" 

is exactly the missing assumption. For the late pod, we don't get a reconcile for the admitted workload; we get a reconcile for the finished root.

mimowo: Ahhh, this is correct. This will probably be fixed later when we start calling startJob, probably here: https://github.com/kubernetes-sigs/kueue/pull/12421, cc @ikchifo 

However, until that happens you diagnosis is right - we only get the reconciles on the already finished workloads. So. indeed the most pragmantic flow for the fix will be:

1. we have reconcile for a ~an admitted~ workload, for the workload we can read the SliceName from the annotation
2. we identify all Pods by the indexer, based on the SliceName annotation on the Pods
3. we find the last admitted workload in the chain, and for the workload ~we are currently reconciling~ we only ungate the certain number of Pods.

Once we fix the Job PodTemplates injection, then we can revisit the topic of relying on the fact that we get the reconcile for the last workload in the chain.

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -181,6 +252,7 @@ func (r *elasticJobUngater) Update(e event.TypedUpdateEvent[*kueue.Workload]) bo
 
 func shouldUngate(wl *kueue.Workload) bool {
 	return workloadslicing.IsElasticWorkload(wl) &&
+		!workload.IsFinished(wl) &&
 		(workload.IsAdmitted(wl) || workload.HasQuotaReservation(wl))
```
mimowo, line 267: Actually, this is a bug in the previous impl - we should only ungate when the workload is Admitted. QuotaReservation is not enough, because there might be extra admission checks. 

would you like to send a separate PR for the fix here?

cc @sohankunkerkar

pkg/controller/elasticjobs/elastic_job_ungater.go
mimowo: No, I'm still puzzled by that. If the "late pods:" (newest) have the workload annotation pointing to the first workload (oldest in the chain), then why this change: https://github.com/kubernetes-sigs/kueue/pull/12045/changes#diff-41ddd3064525f2b6f99b4178832975ef69dd96e3e7f996d784ef10c70c37ec78R295-R300

I'm wondering this may also be interfering with your assessment .

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -220,18 +292,32 @@ func (h *elasticPodHandler) queueReconcileForPod(ctx context.Context, object cli
 	if !isPod {
 		return
 	}
-	wlName, found := pod.Annotations[kueue.WorkloadAnnotation]
-	if !found {
+	// Enqueue by the stable slice-chain key, not the pod's stamped
+	// WorkloadAnnotation. A pod minted after a scale-up still carries the previous
+	// (now Finished) slice's name, but the chain key is shared by every slice, so
+	// Reconcile can always resolve the active slice from it.
+	sliceName := podSliceName(pod)
```
mimowo, line 310: I think changes here don't really matter - because it should not matter for which workload we reconcile as the other code is already looking for the latest active workload in the chain anyway.

Trigged by thinking in the context of this discussion: https://github.com/kubernetes-sigs/kueue/pull/12045#discussion_r3469426139

mimowo: Or this is meant to improve readability - we explicitly trigger the reconicle only for the root, because the previous behavior was tricky? In that case I can by the motivation, but it is far from clear based on the comment.

pkg/controller/elasticjobs/elastic_job_ungater.go
mimowo: Ok, at this point I'm ok with the proposed approach to find the latest active workload slice. 

I think we could introduce the helper function FindLatestActiveWorkload which uses commonized code with  the FindNotFinishedWorkloads to find the last admitted in the chain.

Then we don't need the new indexer iiuc.

test/integration/singlecluster/controller/jobs/raycluster/raycluster_controller_test.go
```diff
@@ -943,4 +946,210 @@ var _ = ginkgo.Describe("RayCluster with elastic jobs via workload-slices suppor
...
+		ginkgo.By("the cap holds: both granted workers are ungated and no surplus is")
+		gomega.Consistently(func(g gomega.Gomega) {
+			pods := &corev1.PodList{}
+			g.Expect(k8sClient.List(ctx, pods, client.InNamespace(ns.Name))).Should(gomega.Succeed())
+			ungated := 0
+			for i := range pods.Items {
+				if !utilpod.HasGate(&pods.Items[i], kueue.ElasticJobSchedulingGate) {
+					ungated++
+				}
+			}
```
mimowo: Introduce a helper `findUngatedPods() sets.Set[NamespacedName]` - this way we will have much better assert output if we do `g.Expect(ungatedPods).Should(gomega.HasLen(2))`

because we will have the Pod names in the failure message - so easier debugging of flakes.

Also, such a function could likely be reused

test/integration/singlecluster/controller/jobs/raycluster/raycluster_controller_test.go
```diff
@@ -943,4 +946,210 @@ var _ = ginkgo.Describe("RayCluster with elastic jobs via workload-slices suppor
...
+				if workload.IsFinished(wl) {
+					g.Expect(wl.Name).Should(gomega.Equal(rootWorkloadName))
+					foundFinishedRoot = true
+					continue
+				}
+				g.Expect(workload.IsAdmitted(wl)).Should(gomega.BeTrue())
+				g.Expect(wl.Spec.PodSets[1].Count).Should(gomega.BeEquivalentTo(int32(2)))
+				activeWorkload = wl
+				workerPodSet = wl.Spec.PodSets[1].Name
+			}
```
mimowo: Finishing the root is not impacted by PR, I don't think it is relevent here. Still, if you think it is then avoid asserting on bools because they are hard to debug. Instead you may find the root workload, and assert it is fnished using a helper for checking a workload is finished - this is expected to produce a more informative output.

pkg/controller/elasticjobs/elastic_job_ungater.go
mcochner: > I think we could introduce the helper function FindLatestActiveWorkload which uses commonized code with the FindNotFinishedWorkloads to find the last admitted in the chain.
Then we don't need the new indexer iiuc.

Proposed change implemented in [refactor: resolve active elastic slice via owner index](https://github.com/kubernetes-sigs/kueue/pull/12045/commits/5ad2b86c9876f2cb476653de03e1df0e63aa8e15)

test/integration/singlecluster/controller/jobs/raycluster/raycluster_controller_test.go
```diff
@@ -943,4 +946,210 @@ var _ = ginkgo.Describe("RayCluster with elastic jobs via workload-slices suppor
...
+			g.Expect(foundFinishedRoot).Should(gomega.BeTrue())
+			g.Expect(activeWorkload).ShouldNot(gomega.BeNil())
+		}, util.Timeout, util.Interval).Should(gomega.Succeed())
```
mimowo: I would like to decompose the complex wait into smaller ones:

```golang
ginkgo.By("new workload is admitted")
	workloads := &kueue.WorkloadList{}
	g.Expect(k8sClient.List(ctx, workloads, client.InNamespace(testRayCluster.Namespace))).Should(gomega.Succeed())

	activeWorkloads := util.FindNonFinishedWorkloads(workloads.Items)
	g.Expect(activeWorkloads).To(gomega.HaveLen(1), "Expected exactly 1 non-finished workload")

	testRayClusterWorkload = &activeWorkloads[0]
	g.Expect(testRayClusterWorkload.Name).ShouldNot(gomega.Equal(nameBeforeScaleUp))
	g.Expect(workload.IsAdmitted(testRayClusterWorkload)).Should(gomega.BeTrue())

ginkgo.By("old workload is finished")
oldWorkload := &kueue.Workload{}
	g.Expect(k8sClient.Get(ctx, oldWorkloadKey, oldWorkload)).Should(gomega.Succeed())
	g.Expect(workload.IsFinished(oldWorkload)).Should(gomega.BeTrue())
```
or something elong the lines, just to avoid the nested checks, and to have better failure messages when things don't work well

test/integration/singlecluster/controller/jobs/raycluster/raycluster_controller_test.go
mimowo: thinking about something like:

```golang
ginkgo.By("the cap holds: exactly 2 worker pods are ungated")
gomega.Consistently(func(g gomega.Gomega) {
	pods := &corev1.PodList{}
	g.Expect(k8sClient.List(ctx, pods, client.InNamespace(ns.Name))).Should(gomega.Succeed())

	ungatedWorkerPods := sets.New[string]()
	for i := range pods.Items {
		pod := &pods.Items[i]
		if !utilpod.HasGate(pod, kueue.ElasticJobSchedulingGate) {
			ungatedWorkerPods.Insert(pod.Name)
		}
	}

	g.Expect(ungatedWorkerPods).Should(
		gomega.HaveLen(2),
		"Expected exactly 2 ungated worker pods, got %v",
		sets.List(ungatedWorkerPods),
	)
}, util.ConsistentDuration, util.Interval).Should(gomega.Succeed())
```

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -159,16 +184,51 @@ func (r *elasticJobUngater) listGatedPods(ctx context.Context, wl *kueue.Workloa
...
+			continue
+		}
+		// Ungate the lowest-named pods first for deterministic behavior.
+		slices.SortFunc(pods, func(a, b *corev1.Pod) int { return strings.Compare(a.Name, b.Name) })
+		if int32(len(pods)) > room {
+			pods = pods[:room]
+		}
+		gated = append(gated, pods...)
+	}
 	return gated, nil
```
mimowo, line 236: This function is geeting complex, and so we should add logging for the number of granted pods, ungated, and gated, log that pod podSet at V4 I think.

mimowo:
@mcochner please rebase

pkg/controller/elasticjobs/elastic_job_ungater.go
mimowo: Actually, let's consider that a separate issue to support ElasticJobs and admission checks - I think this is currently pretty broken anyway. When we don't have admission checks then QuotaReserved=True iff Admitted=True. We will open a dedicated issue for that so don't worry in this PR, cc @yaroslava-serdiuk @sohankunkerkar

test/integration/singlecluster/controller/jobs/raycluster/raycluster_controller_test.go
```diff
@@ -943,4 +946,210 @@ var _ = ginkgo.Describe("RayCluster with elastic jobs via workload-slices suppor
...
+			pod := testingpod.MakePod(fmt.Sprintf("worker-%d", i), ns.Name).
+				Annotation(kueue.WorkloadAnnotation, rayWorkload.Name).
+				Annotation(kueue.WorkloadSliceNameAnnotation, rayWorkload.Name).
+				Label(pkgconstants.PodSetLabel, string(workerPodSet)).
+				Gate(kueue.ElasticJobSchedulingGate).
+				Obj()
+			util.MustCreate(ctx, k8sClient, pod)
+		}
+
+		countUngated := func(g gomega.Gomega) int {
```
mimowo: Instead of a counter let's return Set. Then you can still formulate the asserts using `gomega.HasLen(set.UnstructuredList(), 0)` , then the error message which contain the ungated pods, which is much easier for later debugging.

mimowo:
@mcochner since this PR is an important bugfix ans is getting cherrypicked let's squash the commits - it usually make the CP easier.

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -98,27 +114,30 @@ func (r *elasticJobUngater) Reconcile(ctx context.Context, req reconcile.Request
...
+	root := &kueue.Workload{}
+	if err := r.client.Get(ctx, req.NamespacedName, root); err != nil {
+		return reconcile.Result{}, client.IgnoreNotFound(err)
+	}
+	active, err := r.activeSlice(ctx, root)
+	if err != nil {
+		return reconcile.Result{}, err
 	}
-	if !workload.IsAdmitted(wl) && !workload.HasQuotaReservation(wl) {
+	if active == nil {
```
mimowo, line 136: Let's log this case at v2, iiuc this is an anomaly

test/integration/singlecluster/controller/jobs/raycluster/raycluster_controller_test.go
```diff
@@ -943,4 +961,189 @@ var _ = ginkgo.Describe("RayCluster with elastic jobs via workload-slices suppor
...
+		// This test mints more gated worker pods than the admitted slice was
+		// granted (KubeRay is not running in the integration manager, so the pod
+		// creation is done by hand) and asserts the cap holds — both immediately
+		// and consistently.
+		const (
+			grantedWorkerCount = 2
+			totalWorkerPods    = 5
+		)
+
+		testRayCluster := testingraycluster.MakeCluster("ungate-cap", ns.Name).
```
mimowo: Since this bug is not specific to Ray I would prefer the test to use regular batch Jobs so that it is lighter to run

mcochner: Excellent point!

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -168,16 +197,60 @@ func (r *elasticJobUngater) listGatedPods(ctx context.Context, wl *kueue.Workloa
...
+		gated = append(gated, toUngate...)
+	}
 	return gated, nil
 }
 
+// activeSlice resolves the active (latest admitted, non-finished) workload slice
+// of the chain that anyWl belongs to, or nil if none qualifies. It looks up the
+// chain through the owning job's workload index, reusing the same slice ordering
+// as the rest of the slicing code (workloadslicing.FindLatestActiveWorkload).
+func (r *elasticJobUngater) activeSlice(ctx context.Context, anyWl *kueue.Workload) (*kueue.Workload, error) {
```
sohankunkerkar, line 243: @mimowo should we drop it and use the reconciled workload's own granted counts directly?

mimowo: Yeah, that was my original thinking, and my question. I think it could work, but maybe proving correctness is harder, so I'm ok with using the currently active slice.

sohankunkerkar (approved):
I'm happy with the changes. Thanks for looking into it.

mimowo:
Thank you for fixing one of the most important ElasticJob bugs 👍
/lgtm
/approve

I will try to use the cherry-pick robot, but I think this is unlikely to succeed as the code evolved a bit. Please prepare the CP manually if the robot fails on conflicts.
/cherrypick release-0.18
/cherrypick release-0.17

mimowo:
@mcochner friendly ping - do you have cycles to prepare the CPs?

mimowo:
/release-note-edit
```release-note
ElasticJobViaWorkloadSlices: Fixed a bug where worker pods of an elastic job could be ungated after scale up, 
past the ClusterQueue quota; ungating is now capped to the replicas granted quota across the workload-slice chain.
```

tenzen-y:
/release-note-edit

```release-note
ElasticJobsViaWorkloadSlices: Fixed a bug where worker pods of an elastic job could be ungated after scale up, 
past the ClusterQueue quota; ungating is now capped to the replicas granted quota across the workload-slice chain.
```

PR #12052, opened 2026-06-10
feat: stable domain sorting for deterministic TAS placement on score ties

### LINKED ISSUE #11910: TAS: Ensure stable domain sorting in balanced placement to improve slice reuse
We would like to introduce stable sorting for Topology Aware Scheduling (TAS) node allocation when TAS scoring yields equal scores for multiple domains/nodes. 

Currently, when multiple equivalent nodes/domains are available, the selection across different TAS placement algorithms can be non-deterministic (e.g., when the `TASBalancedPlacement` feature gate is enabled, nodes with equal capacities and entropies might be selected randomly). 

We should ensure that all TAS sorting paths have a consistent, stable tie-breaker. One potential approach could be to fall back to comparing the domain's `levelValues` (which include the node name or topology values) to break the tie, similar to how it is currently handled in the fallback `sortedDomains` logic.

**Why is this needed**:
In environments where workloads require specific physical topologies (e.g., hardware accelerators interconnected at the rack or block level), unpredictable scheduling reduces the opportunity for infrastructure reuse. When consecutive, identical requests arrive in a cluster with symmetric free capacity, an unstable sort means Kueue might distribute them across completely different, equivalent nodes each time. 

By ensuring a stable sort order across all TAS algorithms, Kueue will predictably prioritize the same nodes when scores are tied. This predictability is highly beneficial for systems relying on external admission checks like dynamic hardware provisioning, as it maximizes the reuse of existing underlying infrastructure and minimizes fragmentation over time.

**Completion requirements**:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

mvanhorn:
#### What type of PR is this?

/kind feature
/area tas

Makes TAS domain selection deterministic when multiple domains tie on score. Three call sites iterated or sorted domains in map order, so equal-capacity (and, with TASBalancedPlacement, equal-entropy) candidates could be picked differently across scheduling cycles for the same cluster state:

- `sortDomainsByCapacityAndEntropy` returned 0 on full ties; it now tie-breaks by `levelValues`.
- `selectOptimalDomainSetToFit` iterated the caller's slice in map order; it now iterates a `levelValues`-sorted clone (sorted by capacity and entropy when the feature gate is on), without mutating the caller's slice.
- `findBestDomainsForBalancedPlacement` collected higher-level domains from a map and iterated them unsorted; they are now sorted before their children are considered.

The lexicographic `levelValues` comparison is the same final ordering `sortedDomains` and `sortedDomainsWithLeader` already use, extracted into shared `compareDomainLevelValues` / `sortDomainsByLevelValues` helpers so the tie-break rule stays in one place. Stable selection means repeated workloads on identical cluster state land on the same set of nodes, which keeps reusable inter-node connections warm as discussed in the issue.

Fixes #11910

`selectOptimalDomainSetToFit` clones the input slice before sorting so the change cannot reorder the caller's view of `domains`. New tests cover tie-breaking in `sortDomainsByCapacityAndEntropy`, deterministic results from `selectOptimalDomainSetToFit` across shuffled input orders, and stable higher-level iteration in `findBestDomainsForBalancedPlacement`. `go test ./pkg/cache/scheduler/...` passes.

```release-note
TAS: domain selection is now deterministic when multiple domains tie on score; ties are broken by the domains' levelValues ordering.
```

## Summary by CodeRabbit

* **Tests**
  * Enhanced test coverage for deterministic domain ordering and placement scenarios.

* **Refactor**
  * Improved consistency in domain ordering by implementing deterministic tie-breaking during scheduling decisions.
  * Consolidated domain sorting logic into shared utility functions for better maintainability.

### DIFF
--- a/pkg/cache/scheduler/tas_balanced_placement.go
+++ b/pkg/cache/scheduler/tas_balanced_placement.go
@@ -84,8 +84,11 @@ func selectOptimalDomainSetToFit(s *TASFlavorSnapshot, domains []*domain, sliceC
 		return nil
 	}
 
+	orderedDomains := slices.Clone(domains)
 	if priorizeByEntropy {
-		sortDomainsByCapacityAndEntropy(domains)
+		sortDomainsByCapacityAndEntropy(orderedDomains)
+	} else {
+		sortDomainsByLevelValues(orderedDomains)
 	}
 
 	// domain_placements[i][j][k] stores a list of domains that uses 'i' domains with
@@ -96,7 +99,7 @@ func selectOptimalDomainSetToFit(s *TASFlavorSnapshot, domains []*domain, sliceC
 	}
 	domainPlacements[0][leaderCount] = map[int32][]*domain{sliceCount * sliceSize: {}}
 
-	for _, d := range domains {
+	for _, d := range orderedDomains {
 		for i := optimalNumberOfDomains; i > 0; i-- {
 			for _, beforeLeader := range slices.Sorted(maps.Keys(domainPlacements[i-1])) {
 				for _, beforeState := range slices.Sorted(maps.Keys(domainPlacements[i-1][beforeLeader])) {
@@ -226,7 +229,7 @@ func sortDomainsByCapacityAndEntropy(domains []*domain) {
 		if bEntropy < aEntropy {
 			return -1
 		}
-		return 0
+		return compareDomainLevelValues(a, b)
 	})
 }
 
@@ -242,7 +245,9 @@ func findBestDomainsForBalancedPlacement(s *TASFlavorSnapshot, params *topologyA
 	if params.requestedLevelIdx == 0 {
 		requestedLevelDomainsToConsider = [][]*domain{slices.Collect(maps.Values(s.domainsPerLevel[0]))}
 	} else {
-		for _, higherLevelDomain := range slices.Collect(maps.Values(s.domainsPerLevel[params.requestedLevelIdx-1])) {
+		higherLevelDomains := slices.Collect(maps.Values(s.domainsPerLevel[params.requestedLevelIdx-1]))
+		sortDomainsByLevelValues(higherLevelDomains)
+		for _, higherLevelDomain := range higherLevelDomains {
 			requestedLevelDomainsToConsider = append(requestedLevelDomainsToConsider, higherLevelDomain.children)
 		}
 	}
--- a/pkg/cache/scheduler/tas_balanced_placement_test.go
+++ b/pkg/cache/scheduler/tas_balanced_placement_test.go
@@ -26,11 +26,19 @@ import (
 	utiltesting "sigs.k8s.io/kueue/pkg/util/testing"
 )
 
+func domainIDs(domains []*domain) []string {
+	ids := make([]string, len(domains))
+	for i, d := range domains {
+		ids[i] = string(d.id)
+	}
+	return ids
+}
+
 func TestSelectOptimalDomainSetToFit(t *testing.T) {
-	d1 := &domain{id: "d1", state: 9, sliceState: 9, leaderState: 1, stateWithLeader: 8, sliceStateWithLeader: 8}
-	d2 := &domain{id: "d2", state: 6, sliceState: 6, leaderState: 0, stateWithLeader: 6, sliceStateWithLeader: 6}
-	d3 := &domain{id: "d3", state: 4, sliceState: 4, leaderState: 1, stateWithLeader: 3, sliceStateWithLeader: 3}
-	d4 := &domain{id: "d4", state: 2, sliceState: 2, leaderState: 0, stateWithLeader: 2, sliceStateWithLeader: 2}
+	d1 := &domain{id: "d1", levelValues: []string{"d1"}, state: 9, sliceState: 9, leaderState: 1, stateWithLeader: 8, sliceStateWithLeader: 8}
+	d2 := &domain{id: "d2", levelValues: []string{"d2"}, state: 6, sliceState: 6, leaderState: 0, stateWithLeader: 6, sliceStateWithLeader: 6}
+	d3 := &domain{id: "d3", levelValues: []string{"d3"}, state: 4, sliceState: 4, leaderState: 1, stateWithLeader: 3, sliceStateWithLeader: 3}
+	d4 := &domain{id: "d4", levelValues: []string{"d4"}, state: 2, sliceState: 2, leaderState: 0, stateWithLeader: 2, sliceStateWithLeader: 2}
 
 	testCases := map[string]struct {
 		domains     []*domain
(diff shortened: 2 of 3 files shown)


mimowo:
/ok-to-test
Thanks!

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1570,6 +1568,14 @@ func (s *TASFlavorSnapshot) lowerLevelDomains(domains []*domain) []*domain {
 	return result
 }
 
+func compareDomainLevelValues(a, b *domain) int {
+	return slices.Compare(a.levelValues, b.levelValues)
+}
+
+func sortDomainsByLevelValues(domains []*domain) {
+	slices.SortFunc(domains, compareDomainLevelValues)
+}
```
mimowo, line 1577: I know the pre-existing code looked this way, but instead of introducing helpers like `sortDomainsBy...` could we rather create helper compare functions `(a, b *domain) int` and then just pass the selected function to slices.SortFunc. wdyt?

pkg/cache/scheduler/tas_balanced_placement.go
```diff
@@ -84,8 +84,11 @@ func selectOptimalDomainSetToFit(s *TASFlavorSnapshot, domains []*domain, sliceC
 		return nil
 	}
 
+	orderedDomains := slices.Clone(domains)
 	if priorizeByEntropy {
-		sortDomainsByCapacityAndEntropy(domains)
+		sortDomainsByCapacityAndEntropy(orderedDomains)
```
mwielgus, line 89: Is the domain name included in both cases (after all other criteria) to ensure stable sorting?

mimowo (commented):
IIUC this PR only affects behavior for BalancedPlacement, right? In that case I think it would be good to reflect that in the release note.

pkg/cache/scheduler/tas_balanced_placement.go
```diff
@@ -84,8 +84,11 @@ func selectOptimalDomainSetToFit(s *TASFlavorSnapshot, domains []*domain, sliceC
...
+	} else {
+		sortDomainsByLevelValues(orderedDomains)
```
mimowo, line 91: This sort was absent before,  so we relied on the order which came from above. I'm wondering if now sorting by levelValues we may break the order by some other more imporant factors, like the capacity.

mimowo (commented):
cc @pajakd ptal

pkg/cache/scheduler/tas_balanced_placement_test.go
```diff
@@ -172,6 +246,24 @@ func TestPlaceSlicesOnDomainsBalanced(t *testing.T) {
 	}
 }
 
+func TestPlaceSlicesOnDomainsBalancedStableTieBreak(t *testing.T) {
```
pajakd, line 249: Lets reformat this test suite to contain testCases (in case we will want to add more cases to it).

pkg/cache/scheduler/tas_balanced_placement.go
pajakd: In the `selectOptimalDomainSetToFit` the order in which the domains are considered does not matter too much. We are always going to find an optimal subset of domains (using dynamic programming instead of greedy as in the original TAS). 

The only issue is that if two domains "look identical" from the perspective of the algorithm (ie can accommodate the same number of pods), we introduced this fancy entropy to prioritize domains whose children are balanced.

So I think adding the LevelValue sorting is not going to break the algorithm.

danylott:
Hi @mvanhorn, thanks for your effort on this issue.

Could you please take a look at the comments from the reviewers, and if possible share when this issue might be resolved?

danylott:
Hi @mvanhorn, will you be able to take a look at this PR?

mimowo:
/test all
Checking is still tests pass

mimowo:
@mvanhorn do you have cycles to address the remaining comments?

mimowo:
I will review in detail later today or tomorrow

pkg/cache/scheduler/tas_balanced_placement_test.go
```diff
@@ -86,12 +94,78 @@ func TestSelectOptimalDomainSetToFit(t *testing.T) {
...
+	}
+
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			_, log := utiltesting.ContextWithLog(t)
+			s := newTASFlavorSnapshot(log, "dummy", []string{}, nil)
+			domains := []*domain{
+				{id: "leaf-a", levelValues: []string{"block-b", "host-a"}, state: 3, sliceState: 3, stateWithLeader: 3, sliceStateWithLeader: 3},
+				{id: "leaf-m", levelValues: []string{"block-b", "host-m"}, state: 3, sliceState: 3, stateWithLeader: 3, sliceStateWithLeader: 3},
+				{id: "leaf-z", levelValues: []string{"block-a", "host-z"}, state: 3, sliceState: 3, stateWithLeader: 3, sliceStateWithLeader: 3},
```
dkaluza, line 116: Can we have one more not leaf with `block-a` after the and host larger lexicographically so the chosen value is not the last in the array?

pkg/cache/scheduler/tas_balanced_placement_test.go
```diff
@@ -86,12 +94,78 @@ func TestSelectOptimalDomainSetToFit(t *testing.T) {
...
+			want: []string{"leaf-z", "leaf-a", "leaf-m"},
+		},
+		"capacity and entropy ranking before tiebreaker": {
+			domains: []*domain{
+				{id: "lower-leader", levelValues: []string{"a"}, leaderState: 0, sliceStateWithLeader: 100, children: []*domain{{state: 50}, {state: 50}}},
+				{id: "lower-capacity", levelValues: []string{"b"}, leaderState: 1, sliceStateWithLeader: 4, children: []*domain{{state: 2}, {state: 2}}},
+				{id: "low-entropy", levelValues: []string{"c"}, leaderState: 1, sliceStateWithLeader: 5, children: []*domain{{state: 4}, {state: 0}}},
+				{id: "high-entropy", levelValues: []string{"d"}, leaderState: 1, sliceStateWithLeader: 5, children: []*domain{{state: 2}, {state: 2}}},
+			},
+			want: []string{"high-entropy", "low-entropy", "lower-capacity", "lower-leader"},
```
dkaluza, line 148: Can we indicate in the test case name what is the exact order of the sorting? Otherwise it is hard to grasp if the result is actually ok.

dkaluza (commented):
IMO it would be best to have a test case to check stable sorting of whole `findTopologyAssignment`, otherwise we will not be sure if the issue with unstable sorting is fixed entirely or not.

pkg/cache/scheduler/tas_flavor_snapshot.go
mimowo: This is more of an exploratory question. We can ignore it for now as it does not change the nature of the PR

danylott:
Hi @mimowo @dkaluza, I've continue working on this issue in https://github.com/kubernetes-sigs/kueue/pull/13027
Fixed comments from you & other reviewers

mimowo:
I would like to handle this as a bug and cherrypick, because deterministic order is quite important for TAS, and the core algorithm (outside of balanced placement already follows the principle AFAIK), since labelValues are used for [fallback](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/cache/scheduler/tas_flavor_snapshot.go#L1652). So, I consider it a consistency bug in the branch of compact placement.

With that I would like to cherrypick this change.

/remove-kind feature
/kind bug

mimowo:
There are some remaining comments but @danylott is addressing them as per https://github.com/kubernetes-sigs/kueue/pull/12052#issuecomment-4967801391

mimowo:
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for the PR 👍

mimowo:
@danylott could you follow up with the cherrypick for 0.17?

danylott:
@mimowo Added the cherry pick for 0.17: https://github.com/kubernetes-sigs/kueue/pull/13068

PR #12055, opened 2026-06-10
Invalidate representative mode cache on psAssignment error

yaroslava-serdiuk:
#### What type of PR is this?
AI assisted

/kind bug
/area tas

If onlyTASFlavor() return error the psAssigment has error. In this case the assignment mode should be NoFit (https://github.com/kubernetes-sigs/kueue/blob/main/pkg/scheduler/flavorassigner/flavorassigner.go#L307) but if the previous a.assignmentMode is Fit, RepresentativeMode() will persist to return Fit mode because the cache is not invalidated, which will result in a wrong scheduler results (quota reservation).

Fixes #

```release-note
TAS: Fixed error handling for TAS topology assignments so Workloads are not considered
`Fit` when topology assignment fails. Kueue now treats such assignment errors as `NoFit`
instead of allowing the Workload to reserve quota.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure topology-request validation updates representative mode reliably for elastic-job/TAS scenarios; clear representative mode when recording TAS-related errors and report conflicts when multiple TAS flavors are assigned.
  * Improve error comparison in test assertions to tolerate additional error suffixes or compatible wrapped errors.

* **Tests**
  * Strengthened tests for multi-TAS conflict behavior and explicit representative-mode initialization/assertions; adjusted an integration test to use an explicit resource group for on-demand flavor.

### DIFF
--- a/pkg/scheduler/flavorassigner/flavorassigner_test.go
+++ b/pkg/scheduler/flavorassigner/flavorassigner_test.go
@@ -43,11 +43,11 @@ import (
 
 var (
 	statusComparer = cmp.Comparer(func(a, b Status) bool {
+		if a.err != nil && b.err != nil {
+			return errors.Is(a.err, b.err) || strings.HasPrefix(a.err.Error(), b.err.Error()) || strings.HasPrefix(b.err.Error(), a.err.Error())
+		}
 		if a.err != nil || b.err != nil {
-			if a.err == nil || b.err == nil {
-				return false
-			}
-			return a.err.Error() == b.err.Error()
+			return false
 		}
 		return cmp.Equal(a.reasons, b.reasons, cmpopts.SortSlices(func(x, y string) bool {
 			return x < y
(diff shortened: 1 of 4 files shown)


mimowo:
it is very unclear what the user-observable impact is (if any). Please add some functional unit (or integration) test to make that clear.

yaroslava-serdiuk:
> it is very unclear what the user-observable impact is (if any). Please add some functional unit (or integration) test to make that clear.

The initial bug was observed when the bug https://github.com/kubernetes-sigs/kueue/pull/11997 took place. I.e the non-elastic workloads has empty tas assignment (main bug), but then reserved the quota (this bug). Since that bug is fixed, we won't have an error in https://github.com/kubernetes-sigs/kueue/blob/main/pkg/scheduler/flavorassigner/tas_flavorassigner.go#L64-L74 section since the non-elastic workload skip this section and elastic workloads are verified via webhook https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/jobs/job/job_webhook.go#L245. 

However, I think the issue still may persist if more than 1 flavor assigned, I added a case for this. 

I also believe this is a pattern that we should follow — any errors during psAssignment should result in NoFit.

test/integration/singlecluster/scheduler/scheduler_test.go
```diff
@@ -2133,7 +2133,8 @@ var _ = ginkgo.Describe("Scheduler", func() {
 
 		ginkgo.It("Should not admit new created workloads", func() {
 			ginkgo.By("Create clusterQueue")
-			cq = utiltestingapi.MakeClusterQueue("cluster-queue").Obj()
+			cq = utiltestingapi.MakeClusterQueue("cluster-queue").
+				ResourceGroup(*utiltestingapi.MakeFlavorQuotas("on-demand").Resource(corev1.ResourceCPU, "5").Obj()).Obj()
```
mimowo, line 2137: Please add a dedicated test, it is very tricky to understand why this is making the test against the new changes. Also, it might be impacting pre-existing tests in non-obvious ways.

A dedicated test will allow us to add good comments explaining the expectations.

yaroslava-serdiuk: I updated this test because it started failing with new changes. It started failing because the CQ with no resources considered as CQ that contains onlyTAS resource flavors.

pkg/scheduler/flavorassigner/tas_flavorassigner.go
```diff
@@ -58,16 +58,17 @@ func (a *Assignment) WorkloadsTopologyRequests(log logr.Logger, wl *workload.Inf
...
 
-			var previousAssignment *kueue.TopologyAssignment
-
 			// Only unconstrained topology is supported with elastic workload slices.
-			if workload.IsElasticWorkload(wl.Obj) && features.Enabled(features.ElasticJobsViaWorkloadSlicesWithTAS) {
+			var previousAssignment *kueue.TopologyAssignment
+			if features.Enabled(features.ElasticJobsViaWorkloadSlicesWithTAS) {
 				if podSet.TopologyRequest != nil && podSet.TopologyRequest.Required != nil {
 					psAssignment.error(errors.New("required topology is not supported with ElasticJobsViaWorkloadSlices"))
+					a.representativeMode = nil
```
mimowo: It would be nice to encapsulate these two lines with a function like `a.error(psAssignment, err)` so that we never forget to set `a.representativeMode = nil`.

pkg/scheduler/flavorassigner/flavorassigner_test.go
```diff
@@ -3311,6 +3315,58 @@ func TestAssignFlavors(t *testing.T) {
 				Usage: workload.Usage{Quota: resources.FlavorResourceQuantities{}},
 			},
 		},
+		"multiple TAS flavors assigned to different resources in the same PodSet leads to NoFit": {
```
mimowo, line 3318: Do we have test cases for all 3 changes in WorkloadsTopologyRequests ?

pkg/scheduler/flavorassigner/tas_flavorassigner.go
mwielgus, line 47: Why don't we invalidate the representativeMode in case of all assignments being skipped/error?

mimowo:
@yaroslava-serdiuk since this is a bugfix please also adjust the release note clarifying the fixed scanario to the users.

pkg/scheduler/flavorassigner/tas_flavorassigner.go
yaroslava-serdiuk: The invalidation happens when the error is set for psAssignment, so I don't think we need to do it here as well

pkg/scheduler/flavorassigner/flavorassigner_test.go
yaroslava-serdiuk: Updated already existing cases

mimowo:
/release-note-edit
```release-note
TAS: Fixed error handling for TAS topology assignments so Workloads are not considered
`Fit` when topology assignment fails. Kueue now treats such assignment errors as `NoFit`
instead of allowing the Workload to reserve quota.
```
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for the hardening of TAS 👍

PR #12063, opened 2026-06-10
[Cleanup] Increase Multikueue lost-worker pending retry assertion

mszadkow:
#### What type of PR is this?
/kind bug
/area multikueue

Expecting the admission check to have exactly RetryCount = 1 is weak. 
If the controller runs one extra reconciliation, the retry count can become 2 and the test fails even though the system is behaving correctly.

```release-note
NONE
```

## Summary by CodeRabbit

* **Tests**
  * Refined test assertions for admission check state validation during connection and admission loss scenarios to ensure more granular validation of admission check references and retry behavior.

### DIFF
--- a/test/integration/multikueue/jobs_test.go
+++ b/test/integration/multikueue/jobs_test.go
@@ -1518,11 +1518,11 @@ var _ = ginkgo.Describe("MultiKueue", ginkgo.Label("area:multikueue", "feature:m
 				createdWorkload := &kueue.Workload{}
 				g.Expect(managerTestCluster.client.Get(managerTestCluster.ctx, wlLookupKey, createdWorkload)).To(gomega.Succeed())
 				acs := admissioncheck.FindAdmissionCheck(createdWorkload.Status.AdmissionChecks, kueue.AdmissionCheckReference(multiKueueAC.Name))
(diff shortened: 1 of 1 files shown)


mimowo:
@mszadkow please reference the failed builds

mimowo:
I have synced with @mszadkow and this PR is fixing a potential flake which hasn't been observed yet. It was found by a static code analysis. I will try to analyze that on my end too to see if we should merge the fix proactively

mimowo:
Still, I cannot understand by static code analysis easily if this may happen or not. 

If you believe it may happen we need better evidence, for example:
1. detailed explanation of the code paths leading to the flake
2. demonstrated flake with tuned parameters, for example wait times

If those cannot be provided, no evidence for the test ever flaking, then I wouldn't relax assertions like this, and close the PR for now.

mimowo:
@mszadkow let me know if you want to address the questions in https://github.com/kubernetes-sigs/kueue/pull/12063#issuecomment-4677972360, or we close the PR

mszadkow:
I am not able to present any evidence of the test flaking due to that issue.
Let's close, we can always bring it up if we encounter that

PR #12067, opened 2026-06-10
cleanup: share image-resolution logic for e2e test images

### LINKED ISSUE #12066: Cache GetAgnHostImage with sync.Once to match GetSparkTestImage pattern
Optimize GetAgnHostImage func used in e2e tests by following the pattern established in GetSparkTestImage func.
**Why is this needed**:
GetAgnHostImage opens and parses the agnhost Dockerfile from disk on every call, whereas GetSparkTestImage already wraps the equivalent lookup in a sync.Once so the image is resolved only once and cached for reuse. Aligning GetAgnHostImage with that pattern avoids repeated file I/O and parsing across e2e test runs and keeps the two image-resolution helpers consistent and easier to maintain.

ivnovakov:
#### What type of PR is this?
/kind cleanup

/area testing
Optimize GetAgnHostImage func used in e2e tests by following the pattern established in GetSparkTestImage func.

GetAgnHostImage opens and parses the agnhost Dockerfile from disk on every call, whereas GetSparkTestImage already wraps the equivalent lookup in a sync.Once so the image is resolved only once and cached for reuse. Aligning GetAgnHostImage with that pattern avoids repeated file I/O and parsing across e2e test runs and keeps the two image-resolution helpers consistent and easier to maintain.

Fixes #12066

```release-note
NONE
```

## Release Notes

* **Refactor**
  * Optimized Docker image resolution for tests with caching to improve performance.

* **Chores**
  * Added support for environment variable overrides in test image configuration, with fallback to Dockerfile parsing.

### DIFF
--- a/test/util/constants.go
+++ b/test/util/constants.go
@@ -105,4 +105,7 @@ const (
 var (
 	sparkTestImageOnce sync.Once
 	sparkTestImage     string
+
+	agnHostImageOnce sync.Once
+	agnHostImage     string
 )
--- a/test/util/e2e.go
+++ b/test/util/e2e.go
@@ -27,6 +27,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 
 	cmv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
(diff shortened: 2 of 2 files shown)


mbobrovskyi (commented):
Thank you!

/lgtm
/approve

Issue #11988, opened 2026-06-08
Optimize CI job `periodic-kueue-test-e2e-multikueue-extended-main`


mimowo:
**What would you like to be cleaned**:

Optimize periodic-kueue-test-e2e-multikueue-extended-main, to split it by sharding

**Why is this needed**:

To get below 18min with all CI jobs: https://github.com/kubernetes-sigs/kueue/issues/11606

mimowo:
This is currently the highest priority for the CI job optimizations based on https://github.com/kubernetes-sigs/kueue/issues/11606#issuecomment-4673603161

apullo777:
Hi, I'd like to work on this :)
 
I pulled a few recent runs of periodic-kueue-test-e2e-multikueue-extended-main from Prow. They average ~20 min (range 18.9–22.6 over the last 10 runs). The bulk of that is cluster/operator setup, not test execution: 

Ginkgo itself only runs ~3 min wall-clock (Ran 10 of 10 Specs in 179.86s, 5 parallel nodes), while building the kind image, bringing up the 3 kind clusters, and pulling/loading the 6 external operators (AppWrapper, JobSet, training-operator, Kubeflow Trainer, MPI, KubeRay) eats the other ~16 min. Within that ~3 min of specs, the Ray group (RayJob/RayCluster/RayService) is the heaviest: ~164s of ~258s serial spec time (63%), and it's the only group that needs KubeRay + the Ray image. 

So my thought is to isolate Ray into its own shard by label, mirroring how sequential-extended was sharded (feature-label + per-shard conditional operator install), so each shard only installs the operators it actually needs: 
- shard-0: LeaderWorkerSet, JobSet, AppWrapper, Kubeflow (PyTorchJob / MPIJob / TrainJob)
- shard-1: the Ray integration tests

/assign

mimowo:
sgtm, yes KubeRay is known to be notoriously heavy

PR #12071, opened 2026-06-10
Optimize test-multikueue-e2e-extended target.

mbobrovskyi:
#### What type of PR is this?
/kind cleanup

#### What this PR does / why we need it:
Optmize periodic-kueue-test-e2e-tas-extended-shard-0-main

#### Which issue(s) this PR fixes:

Fixes #12103

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

## Summary by CodeRabbit

* **Tests**
  * Test scripts now emit verbose, timestamped command traces for clearer logs.
  * Automated end-to-end test execution and top-report generation have been disabled in the test workflow.

PR #12096, opened 2026-06-10
Optimize the multikueue/extended e2e test suite by sharding

apullo777:
#### What type of PR is this?

/kind cleanup
/area testing

Shards the `multikueue/extended` e2e suite to isolate the KubeRay setup into its own shard.
Split into two shards:

  - shard-0 (`feature:leaderworkerset,jobset,appwrapper,pytorchjob,mpijob,trainjob`): 7 specs; installs JobSet, LeaderWorkerSet, AppWrapper, Kubeflow Training-Operator, Kubeflow MPI and Kubeflow Trainer.
  - shard-1 (`feature:kuberay`): 3 specs; installs only KubeRay + the Ray image.

Part of #11988

- This PR also gates the MPI operator install on the `feature:mpijob` label (in `e2e-common.sh`), aligning it with the other operators so sharding is self-consistent; a side effect is that `test-e2e-extended-shard-1` no longer installs an unused MPI operator.

```release-note
None
```

## Summary by CodeRabbit

* **Tests**
  * E2E extended suite split into two sharded subtargets for balanced parallel execution and focused feature filtering.
  * Tests labeled by feature so each shard runs only relevant scenarios; suite setup only waits for optional operators when selected.
  * Ray cleanup now removes all RayService resources across namespaces.

* **Chores**
  * Test env wiring refined to separate Ray-related settings from other operator-version settings.

* **Infrastructure**
  * Added per-shard controller-manager configurations and kustomize overlays to tailor integrations and feature gates.

### DIFF
--- a/Makefile-test.mk
+++ b/Makefile-test.mk
@@ -178,11 +178,13 @@ test-multikueue-e2e-extended:
 .PHONY: test-multikueue-e2e-extended-shard-0
 test-multikueue-e2e-extended-shard-0: E2E_NPROCS := 5
 test-multikueue-e2e-extended-shard-0: GINKGO_ARGS=--label-filter=feature:leaderworkerset,feature:jobset,feature:appwrapper,feature:pytorchjob,feature:mpijob,feature:trainjob
+test-multikueue-e2e-extended-shard-0: E2E_CONFIG_FOLDER=multikueue/extended-shard-0
 test-multikueue-e2e-extended-shard-0: setup-e2e-env run-test-multikueue-e2e-extended-$(E2E_KIND_VERSION:kindest/node:v%=%)
 
 .PHONY: test-multikueue-e2e-extended-shard-1
 test-multikueue-e2e-extended-shard-1: E2E_NPROCS := 5
 test-multikueue-e2e-extended-shard-1: GINKGO_ARGS=--label-filter=feature:kuberay
+test-multikueue-e2e-extended-shard-1: E2E_CONFIG_FOLDER=multikueue/extended-shard-1
 test-multikueue-e2e-extended-shard-1: setup-e2e-env run-test-multikueue-e2e-extended-$(E2E_KIND_VERSION:kindest/node:v%=%)
 
 .PHONY: test-multikueue-e2e-sequential
@@ -396,6 +398,7 @@ run-test-multikueue-e2e-extended-%:
 		E2E_ENFORCE_OPERATOR_UPDATE=$(E2E_ENFORCE_OPERATOR_UPDATE) \
 		USE_RAY_FOR_TESTS=$(USE_RAY_FOR_TESTS) \
 		E2E_TARGET_FOLDER="multikueue/extended" \
+		E2E_CONFIG_FOLDER=$(E2E_CONFIG_FOLDER) \
 		TEST_LOG_LEVEL=$(TEST_LOG_LEVEL) \
 		E2E_RUN_ONLY_ENV=$(E2E_RUN_ONLY_ENV) \
 		E2E_USE_HELM=$(E2E_USE_HELM) \
--- a/Makefile-test.mk
+++ b/Makefile-test.mk
@@ -155,8 +155,35 @@ test-e2e-extended-helm: test-e2e-extended
 test-multikueue-e2e-baseline: E2E_NPROCS := 5
 test-multikueue-e2e-baseline: setup-e2e-env run-test-multikueue-e2e-baseline-$(E2E_KIND_VERSION:kindest/node:v%=%)
 
-test-multikueue-e2e-extended: E2E_NPROCS := 5
-test-multikueue-e2e-extended: setup-e2e-env run-test-multikueue-e2e-extended-$(E2E_KIND_VERSION:kindest/node:v%=%)
+# Assign shard-0 operator versions (all operators except KubeRay) to the shard-0 target
+TEST_MULTIKUEUE_E2E_EXTENDED_SHARD_0_TARGETS := test-multikueue-e2e-extended-shard-0
+$(TEST_MULTIKUEUE_E2E_EXTENDED_SHARD_0_TARGETS): export JOBSET_VERSION := $(JOBSET_VERSION)
+$(TEST_MULTIKUEUE_E2E_EXTENDED_SHARD_0_TARGETS): export LEADERWORKERSET_VERSION := $(LEADERWORKERSET_VERSION)
+$(TEST_MULTIKUEUE_E2E_EXTENDED_SHARD_0_TARGETS): export APPWRAPPER_VERSION := $(APPWRAPPER_VERSION)
+$(TEST_MULTIKUEUE_E2E_EXTENDED_SHARD_0_TARGETS): export KUBEFLOW_VERSION := $(KUBEFLOW_VERSION)
+$(TEST_MULTIKUEUE_E2E_EXTENDED_SHARD_0_TARGETS): export KUBEFLOW_MPI_VERSION := $(KUBEFLOW_MPI_VERSION)
+$(TEST_MULTIKUEUE_E2E_EXTENDED_SHARD_0_TARGETS): export KUBEFLOW_TRAINER_VERSION := $(KUBEFLOW_TRAINER_VERSION)
+
+# Assign shard-1 operator versions (KubeRay + Ray) to the shard-1 target
+TEST_MULTIKUEUE_E2E_EXTENDED_SHARD_1_TARGETS := test-multikueue-e2e-extended-shard-1
+$(TEST_MULTIKUEUE_E2E_EXTENDED_SHARD_1_TARGETS): export KUBERAY_VERSION := $(KUBERAY_VERSION)
+$(TEST_MULTIKUEUE_E2E_EXTENDED_SHARD_1_TARGETS): export RAY_VERSION := $(RAY_VERSION)
+$(TEST_MULTIKUEUE_E2E_EXTENDED_SHARD_1_TARGETS): export RAYMINI_VERSION := $(RAYMINI_VERSION)
+
+.PHONY: test-multikueue-e2e-extended
+test-multikueue-e2e-extended:
+	$(MAKE) test-multikueue-e2e-extended-shard-0
+	$(MAKE) test-multikueue-e2e-extended-shard-1
+
+.PHONY: test-multikueue-e2e-extended-shard-0
+test-multikueue-e2e-extended-shard-0: E2E_NPROCS := 5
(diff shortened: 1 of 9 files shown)


Makefile-test.mk
```diff
@@ -155,10 +155,36 @@ test-e2e-extended-helm: test-e2e-extended
...
+	$(MAKE) test-multikueue-e2e-extended-shard-0
+	$(MAKE) test-multikueue-e2e-extended-shard-1
+
+.PHONY: test-multikueue-e2e-extended-shard-0
+test-multikueue-e2e-extended-shard-0: E2E_NPROCS := 5
+test-multikueue-e2e-extended-shard-0: GINKGO_ARGS=--label-filter=feature:leaderworkerset,feature:jobset,feature:appwrapper,feature:pytorchjob,feature:mpijob,feature:trainjob
+test-multikueue-e2e-extended-shard-0: setup-e2e-env run-test-multikueue-e2e-extended-$(E2E_KIND_VERSION:kindest/node:v%=%)
+
+.PHONY: test-multikueue-e2e-extended-shard-1
+test-multikueue-e2e-extended-shard-1: export USE_RAY_FOR_TESTS="ray"
```
mbobrovskyi: We already have it in run-test-multikueue-e2e-extended-* target, right?

Makefile-test.mk
```diff
@@ -155,10 +155,36 @@ test-e2e-extended-helm: test-e2e-extended
...
+$(TEST_MULTIKUEUE_E2E_EXTENDED_SHARD_0_TARGETS): export KUBEFLOW_TRAINER_VERSION := $(KUBEFLOW_TRAINER_VERSION)
+
+# Assign shard-1 operator versions (KubeRay + Ray) to the shard-1 target
+TEST_MULTIKUEUE_E2E_EXTENDED_SHARD_1_TARGETS := test-multikueue-e2e-extended-shard-1
+$(TEST_MULTIKUEUE_E2E_EXTENDED_SHARD_1_TARGETS): export KUBERAY_VERSION := $(KUBERAY_VERSION)
+$(TEST_MULTIKUEUE_E2E_EXTENDED_SHARD_1_TARGETS): export RAY_VERSION := $(RAY_VERSION)
+$(TEST_MULTIKUEUE_E2E_EXTENDED_SHARD_1_TARGETS): export RAYMINI_VERSION := $(RAYMINI_VERSION)
+
 .PHONY: test-multikueue-e2e-extended
-test-multikueue-e2e-extended: export USE_RAY_FOR_TESTS="ray"
```
mbobrovskyi: Ah, I see. We already had this variable.

mbobrovskyi: Hmm. Than maybe we should delete it to use raymini in presubmits?

mimowo: We want to use raymini in presubmits consistently, but it does not seem the role of the PR to change it. Maybe this is why we observe so long Kuberay builds for presubmits recently? 

In any case fixing this does not seem reponsibility of this PR.

mimowo: @mbobrovskyi do you have cycles maybe to fix that?

mimowo: In any case, I confirmed we have a regression currently, opened https://github.com/kubernetes-sigs/kueue/issues/12103 to track it

apullo777: Thanks for the review and for digging into the raymini/ray topic 🙏

Agreed this is out of scope. I will keep ="ray" on shard-1 to preserve current behavior. Please let me know if anything else is needed.

Makefile-test.mk
apullo777: run-test-multikueue-e2e-extended-% only passes it through but does not set it. USE_RAY_FOR_TESTS only matters for feature:kuberay (shard-1), so I scoped it here with the other ray vars, which also keeps shard-0 from pulling the ray image.

Makefile-test.mk
mimowo: We are merging https://github.com/kubernetes-sigs/kueue/pull/12071 right now, so just rebase when merged.

mimowo:
@apullo777 please rebase and check for the failures reason

apullo777:
Hi, @mimowo @mbobrovskyi

I dug into the failures. The issue seems to be that I treated the MultiKueue e2e setup like the single-cluster one, but missing framework CRDs are not tolerated the same way there.

In MultiKueue, the remote client hard-fails if any enabled integration is missing its CRD, so the clusters stay `Inactive` and the AdmissionCheck remains `NoUsableClusters`. In shard-0 we skip KubeRay, but still deploy Kueue with `ray.io/*` enabled; shard-1 has the inverse issue.

There is also a cleanup issue where `AfterEach` assumes Ray CRDs exist.

I see two possible fixes:

1. Per-shard config folders (similar to `multikueue/baseline`): each shard deploys Kueue with only the integrations it actually installs. This keeps the config consistent with the test environment, but introduces additional config variants that need to stay in sync, and cleanup would still need to tolerate missing CRDs.

2. Install only the missing CRDs (without the operators): smaller change, localized to the multikueue e2e setup, and it also avoids the cleanup issue since the CRDs exist. The downside is that the deployed config would still enable integrations whose operators are not installed, and the CRD list would need to stay aligned with the shard filters.

I’m open to either approach, or another direction if preferred. WDYT?

mbobrovskyi:
> In MultiKueue, the remote client hard-fails if any enabled integration is missing its CRD, so the clusters stay Inactive and the AdmissionCheck remains NoUsableClusters. In shard-0 we skip KubeRay, but still deploy Kueue with ray.io/* enabled; shard-1 has the inverse issue.

Yeah, if we're using MultiKueue, the CRDs must be installed. 

```
2026-06-11T18:32:35.029112234Z stderr F 2026-06-11T18:32:35.028943103Z	ERROR	multikueue/multikueuecluster.go:284	Unable to start the watcher	{"clusterName": "worker1-pclb7", "kind": "ray.io/v1, Kind=RayCluster", "error": "no matches for kind \"RayCluster\" in version \"ray.io/v1\""}
2026-06-11T18:32:35.029137915Z stderr F sigs.k8s.io/kueue/pkg/controller/admissionchecks/multikueue.(*remoteClient).setConfig
2026-06-11T18:32:35.029142895Z stderr F 	/workspace/pkg/controller/admissionchecks/multikueue/multikueuecluster.go:284
2026-06-11T18:32:35.029147815Z stderr F sigs.k8s.io/kueue/pkg/controller/admissionchecks/multikueue.(*clustersReconciler).setRemoteClientConfig
2026-06-11T18:32:35.029151905Z stderr F 	/workspace/pkg/controller/admissionchecks/multikueue/multikueuecluster.go:684
2026-06-11T18:32:35.029156005Z stderr F sigs.k8s.io/kueue/pkg/controller/admissionchecks/multikueue.(*clustersReconciler).Reconcile
2026-06-11T18:32:35.029160065Z stderr F 	/workspace/pkg/controller/admissionchecks/multikueue/multikueuecluster.go:733
2026-06-11T18:32:35.029163955Z stderr F sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Reconcile
2026-06-11T18:32:35.029168345Z stderr F 	/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:221
2026-06-11T18:32:35.029172715Z stderr F sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler
2026-06-11T18:32:35.029176515Z stderr F 	/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:478
2026-06-11T18:32:35.029179755Z stderr F sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem
2026-06-11T18:32:35.029193005Z stderr F 	/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:437
2026-06-11T18:32:35.029196405Z stderr F sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func1.1
2026-06-11T18:32:35.029199735Z stderr F 	/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:312

```

I would propose creating a separate configuration for each shard. You can use E2E_CONFIG_FOLDER to select the configuration you want to use.

@mimowo WDYT?

mimowo:
Yes, per shard configs is better out of the two. We prefer to use in our e2e testing the setup which will be used by users. Most users will expect Kueue works well even if full operators are installed.

Another alternative would be to have a custom config which does not enable the not-installed integrations, right? Then such a config could be working for all integrations.

I don't have a view here, whatever is simpler.

apullo777:
Thanks! Agreed. Going with per-shard config folders selected via `E2E_CONFIG_FOLDER`, following the existing `multikueue/baseline` pattern.

One additional fix turned out to be needed: the suite-level `AfterEach` unconditionally deletes RayService objects on all three clusters, which fails with `NoKindMatchError` on the shard without the Ray CRDs. I made that cleanup tolerant of missing CRDs (reusing the existing `deleteAllObjectsInNamespace` helper pattern).

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for the build improvement for the current longest CI job 👍

mbobrovskyi:
@apullo777 could you please prepare cherry-pick for release-0.17?

Makefile-test.mk
```diff
@@ -155,8 +155,37 @@ test-e2e-extended-helm: test-e2e-extended
...
+# Assign shard-1 operator versions (KubeRay + Ray) to the shard-1 target
+TEST_MULTIKUEUE_E2E_EXTENDED_SHARD_1_TARGETS := test-multikueue-e2e-extended-shard-1
+$(TEST_MULTIKUEUE_E2E_EXTENDED_SHARD_1_TARGETS): export KUBERAY_VERSION := $(KUBERAY_VERSION)
+$(TEST_MULTIKUEUE_E2E_EXTENDED_SHARD_1_TARGETS): export RAY_VERSION := $(RAY_VERSION)
+$(TEST_MULTIKUEUE_E2E_EXTENDED_SHARD_1_TARGETS): export RAYMINI_VERSION := $(RAYMINI_VERSION)
+
+.PHONY: test-multikueue-e2e-extended
+test-multikueue-e2e-extended:
+	$(MAKE) test-multikueue-e2e-extended-shard-0
+	$(MAKE) test-multikueue-e2e-extended-shard-1
```
mbobrovskyi, line 176: @apullo777 could you please prepare PR to revert this target and update [README.md](https://github.com/kubernetes-sigs/kueue/blob/e916236b3235d95906e09e4fd98766762732100c/README.md?plain=1#L80)?

PR #12109, opened 2026-06-11
[Fix] Align ClusterQueue metric with the lending limit behavior

mszadkow:
#### What type of PR is this?
/kind bug

The CQ's lending limit was unset the corresponding metric was set to 0.
However when lendingLimit is unset, the queue cannot lend more than it owns so it should be equal to the nominal quota value.

Followup on #11835 

```release-note
Observability: Fix ClusterQueue Lending Limit metric to display value equal to nominal quota if the limit is unset.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Lending-limit metric now reports +Inf when a lending limit is unset and uses configured values when present.

* **Documentation**
  * Clarified metrics docs to state that an unset lending limit is reported as +Inf.

* **Tests**
  * Extended tests to cover lending-limit metrics across multiple queues, label updates, and the +Inf (unlimited) reporting behavior.


pkg/cache/scheduler/clusterqueue.go
```diff
@@ -547,7 +547,10 @@ func (c *clusterQueue) reportResourceMetrics(fairSharingEnabled bool) {
 		} else {
 			borrowing = resourceFloat(fr.Resource, quota.BorrowingLimit.Int64())
 		}
-		if quota.LendingLimit != nil {
+		if quota.LendingLimit == nil {
+			// Unset lendingLimit means all nominal quota can be lent.
+			lending = nominal
```
mimowo: Hm, maybe it is better to use +inf too, to distiguish the case when this is set explicitly vs unset,

This is similar case as for borrowingLimit - we could compute the precise "borrowable_capacity", and we could also introduce "landable_capacity". However, for now I think it makes sense to keep it undefined.

mszadkow: The idea here was to kind of "auto-set" the value up to the nominal quota to not be misleading.
But I get what you mean with the +inf, because if we don't set on purpose it should be something easy to spot in the metrics.
Make sense to me

mimowo: yeah, I understood the idea, but I would still use +inf, because:
1. to make it consistent with the fix for borrowing_limit, there we also could compute borrowable instead
2. to distinguish two cases: lending_limit=nominal_quota vs lending_limit=null

mszadkow: Sure, done

mimowo:
@mszadkow there was today morning basically the same PR https://github.com/kubernetes-sigs/kueue/pull/12143 so I went with that one. So I think we can close here, but make sure it is cherrypicked.

mszadkow:
closed due to similar PR being approved - https://github.com/kubernetes-sigs/kueue/pull/12143

PR #11981, opened 2026-06-08
bump kubeadm from v1beta3 to v1beta4

pacoxu:
#### What type of PR is this?
/kind failing-test
/area testing

#### Which issue(s) this PR fixes:

Fixes #11979

#### Special notes for your reviewer:

See https://github.com/kubernetes/kubernetes/pull/136016 and https://github.com/kubernetes-sigs/kind/issues/3847

https://github.com/kubernetes/kubeadm/issues/2890 is all about kubeadm v1beta4 

#### Does this PR introduce a user-facing change?

```release-note
bump kubeadm v1beta4 API as kubernetes 1.37 dropped kubeadm v1beta3 API support
```

## Summary by CodeRabbit

* **Chores**
  * Updated KIND test cluster kubeadm patches to apiVersion kubeadm.k8s.io/v1beta4.
  * Converted component extraArgs and kubeletExtraArgs from map-style to ordered name/value lists across control-plane and worker join patches.
  * Preserved runtime flags (e.g., v=3) and enable-aggregator-routing by representing them in the new list format.

PR #12121, opened 2026-06-11
[WIP] Feat/11981 kubeadm 1vbeta4 test

reruno:

related to #11981 

```release-note

```

## Summary by CodeRabbit

* **Chores**
  * Increased graceful shutdown timeout and added a 10s preStop delay for the controller manager to allow extra time for cleanup.
  * Raised manager log verbosity to improve diagnostic output.

* **Tests**
  * Updated test cluster configs to the newer kubeadm schema and list-style extraArgs.
  * Enhanced webhook readiness check to validate API server connectivity via a dry-run before proceeding.

### DIFF
--- a/config/components/manager/manager.yaml
+++ b/config/components/manager/manager.yaml
@@ -45,6 +45,10 @@ spec:
             port: 8081
           initialDelaySeconds: 5
           periodSeconds: 10
+        lifecycle:
+          preStop:
+            sleep:
+              seconds: 10
         # TODO(user): Configure the resources accordingly based on the project requirements.
         # More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
         resources:
@@ -55,4 +59,4 @@ spec:
             cpu: 500m
             memory: 512Mi
       serviceAccountName: controller-manager
-      terminationGracePeriodSeconds: 10
+      terminationGracePeriodSeconds: 30
--- a/test/util/e2e.go
+++ b/test/util/e2e.go
@@ -498,6 +498,14 @@ func waitForKueueControllerReadyWithWebhookEndpoints(ctx context.Context, k8sCli
 		g.Expect(endpointIPs).To(gomega.Equal(readyPodIPs))
 	}, LongTimeout, Interval).Should(gomega.Succeed())
 
+	ginkgo.By("Waiting for apiserver to reach the webhook")
+	gomega.Eventually(func(g gomega.Gomega) {
+		rf := &kueue.ResourceFlavor{
+			ObjectMeta: metav1.ObjectMeta{GenerateName: "webhook-probe-rf-"},
+		}
+		g.Expect(k8sClient.Create(ctx, rf, client.DryRunAll)).To(gomega.Succeed())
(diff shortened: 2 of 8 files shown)


reruno:
round 2
/test all

config/components/manager/manager.yaml
```diff
@@ -55,4 +59,4 @@ spec:
             cpu: 500m
             memory: 512Mi
       serviceAccountName: controller-manager
-      terminationGracePeriodSeconds: 10
+      terminationGracePeriodSeconds: 30
```
mimowo, line 62: Why is this getting added? Could we solve without changing the controller, because this could have user-facing consequences, which are unclear to me. Hard to cherrypick.
So, I'm thinking we should have a solution for our tests which does not require that.

reruno:
Round 3
/test all

reruno:
Round 3
/test all

reruno:
Round 4 
/test all

test/util/e2e.go
```diff
@@ -498,6 +498,14 @@ func waitForKueueControllerReadyWithWebhookEndpoints(ctx context.Context, k8sCli
 		g.Expect(endpointIPs).To(gomega.Equal(readyPodIPs))
```
mimowo, line 498: I'm wondering if there the check is strong enough. Because it allows to still exist some Endpoint IPs and still some terminating Pods. I think we could start to fix the issue by making sure the set of non-ready Pods is empty, and the set of non-empty endpoints is also ready.

Can we try that first before changing the prod manifests for kueue?

reruno: https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12121/pull-kueue-test-e2e-extended-shard-2-main-1-35/2065178188132126720/artifacts/run-test-e2e-extended-1.35.5/kind-create.log

https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12121/pull-kueue-test-e2e-extended-shard-2-main-1-36/2065178191516930048/artifacts/run-test-e2e-extended-1.36.1/kind-create.log

I found interesting difference in kind-create.log between 1.35 and 1.36 k8s versions. 

In 1.36 we have `enable-aggregator-routing=true`, but in 1.35 we don't have `enable-aggregator-routing=true`

1.36
```
DEBUG: config/config.go:96] Using the following kubeadm config for node kind-control-plane:
apiServer:
  certSANs:
  - localhost
  - 127.0.0.1
  extraArgs:
  - name: runtime-config
    value: ""
  - name: enable-aggregator-routing
    value: "true"
  - name: v
    value: "3"
``` 
1.35
```
DEBUG: config/config.go:96] Using the following kubeadm config for node kind-worker2:
apiServer:
  certSANs:
  - localhost
  - 127.0.0.1
  extraArgs:
    runtime-config: ""
```

Maybe because of that flag for apiserver, we have flakes in 1.35 e2e tests, and that is why apiserver has trouble communicating with kueue webhook service

mimowo:
/close
To avoid distractions, the problem is already solved

Issue #11170, opened 2026-05-13
pod integration: scheduling gate removal latency after workload admission + missing metric


ichekrygin:
## Summary

When using Kueue's pod integration, there is an observable and sometimes significant delay (up to minutes) between a workload being admitted and the scheduling gate (`kueue.x-k8s.io/admission`) being removed from the pod. During this window, the pod remains blocked from Kubernetes scheduling despite the workload already being admitted.

Additionally, there is no metric to observe or alert on this latency today.

## Expected Behavior

Once a workload transitions to `Admitted`, the scheduling gate on its associated pod(s) should be removed promptly (within seconds), unblocking Kubernetes scheduling.

## Observed Behavior

- Workload is `Admitted` (visible in workload status).
- Pod retains the `kueue.x-k8s.io/admission` scheduling gate.
- The gate is eventually removed, but delay can reach **minutes** in production clusters.

## Flow (for reference)

```
1. Pod is created
2. Kueue webhook adds kueue.x-k8s.io/admission scheduling gate to the pod
3. kube-scheduler skips the pod due to the gate
4. Pod controller creates a Workload for the pod
5. Kueue scheduler admits the Workload
6. Pod controller detects admitted workload → patches pod to remove the gate
7. Pod proceeds with kube-scheduler scheduling
```

The delay occurs in the transition from step 5 to step 6.

### 1. Controller queue saturation
`MaxConcurrentReconciles` for the Pod controller defaults to `1` unless `GroupKindConcurrency["Pod"]` is explicitly configured. Under load, the reconcile request (enqueued by `workloadHandler.Update()` when the Workload status changes) may sit behind other items.

### 2. Informer/watch event delivery latency
The trigger for gate removal is a watch event on the Workload object (status change to Admitted). Under apiserver or network load, event delivery can lag.

### 3. Controller manager downtime / leader election
Any Kueue controller restart leaves admitted workloads gated until the new leader reconciles.

## Missing Observability

There is currently **no metric** capturing the time between workload admission and scheduling gate removal. Existing metrics do not cover this window:

| Metric | Measures | Covers gate removal? |
|---|---|---|
| `kueue_admission_wait_time_seconds` | Workload creation → admission | No |
| `kueue_admitted_until_ready_wait_time_seconds` | Admission → pods ready | No (far later) |
| `kueue_ready_wait_time_seconds` | Workload creation → ready | No |

## Request

Add a new histogram metric to make this latency observable:

```
[...]

PR #12137, opened 2026-06-11
Add kueue_pod_scheduling_gate_removal_seconds metric.

mbobrovskyi:
Add kueue_pod_scheduling_gate_removal_seconds metric.

Part of #11170

```release-note
Observability: Added the `kueue_pod_scheduling_gate_removal_seconds` histogram metric to
measure the time from Workload admission to Pod scheduling-gate removal, helping operators
track delays before admitted Pods can be scheduled.
```

## Summary by CodeRabbit

* **New Features**
  * Added `kueue_pod_scheduling_gate_removal_seconds` Prometheus histogram to measure latency from workload admission to scheduling gate removal, labeled by `name`, `cluster_queue`, and whether the workload is a pod group (`is_group`).

* **Documentation**
  * Updated the metrics reference to include `kueue_pod_scheduling_gate_removal_seconds` and its labels.

* **Bug Fixes**
  * Improved workload context and metric timing so scheduling-gate removal latency is recorded correctly for both single-pod and pod-group workloads.

* **Tests**
  * Extended unit and integration tests to validate the new metric.


site/content/en/docs/reference/metrics.md
```diff
@@ -25,6 +25,7 @@ Use the following metrics to monitor the health of the kueue controllers:
 | --- | --- | --- | --- |
 | `kueue_admission_attempt_duration_seconds` | Histogram | The latency of an admission attempt.<br>The label 'result' can have the following values:<br>- 'success' means that at least one workload was admitted.,<br>- 'inadmissible' means that no workload was admitted. | `result`: possible values are `success` or `inadmissible`<br> `replica_role`: one of `leader`, `follower`, or `standalone` |
 | `kueue_admission_attempts_total` | Counter | The total number of attempts to admit workloads.<br>Each admission attempt might try to admit more than one workload.<br>The label 'result' can have the following values:<br>- 'success' means that at least one workload was admitted.,<br>- 'inadmissible' means that no workload was admitted. | `result`: possible values are `success` or `inadmissible`<br> `replica_role`: one of `leader`, `follower`, or `standalone` |
+| `kueue_pod_scheduling_gate_removal_seconds` | Histogram | The time between workload admission and scheduling gate removal. | `cluster_queue`: the name of the ClusterQueue<br> `is_group`: whether the gate removal applies to a pod group or a single pod |
```
mwielgus: How are multi-pod metrics handled? Does every pod record a sample or there is an aggregated one for the entire workload? Why is is_group needed?

mbobrovskyi: > How are multi-pod metrics handled? Does every pod record a sample or there is an aggregated one for the entire workload?

We record a single sample when the entire group is unscheduled while running.

> Why is is_group needed?

This is required in issue.

pkg/metrics/metrics.go
```diff
@@ -573,6 +578,14 @@ The label 'underlying_cause' can have the following values:
 		}, append([]string{"job_kind", "replica_role"}, extraLabels...),
 	)
 
+	PodSchedulingGateRemovalSeconds = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: constants.KueueName,
+			Name:      "pod_scheduling_gate_removal_seconds",
+			Help:      "The time between workload admission and scheduling gate removal.",
+		}, append([]string{"cluster_queue", "is_group"}, extraLabels...),
```
mimowo: What about the name label I mentioned in the issue comment? How do you break down by name otherwise?

mbobrovskyi: Yeah, still working on it.

mimowo: Support for other scheduling gates could be done in a follow up PRs but I think the label should already be there so that we don't need to break compatibility

mbobrovskyi: Done

pkg/controller/jobs/pod/pod_controller.go
```diff
@@ -282,46 +286,63 @@ func (p *Pod) Run(ctx context.Context, c client.Client, podSetsInfo []podset.Pod
...
-			recorder.Eventf(pod, nil, corev1.EventTypeNormal, jobframework.ReasonStarted, "Started", msg)
+	if updated.Load() {
+		cond := apimeta.FindStatusCondition(wl.Status.Conditions, kueue.WorkloadAdmitted)
+		if cond == nil || cond.Status != metav1.ConditionTrue {
+			return nil
 		}
-		return nil
-	})
+		latency := p.clock.Now().Sub(cond.LastTransitionTime.Time)
+		metrics.RecordPodSchedulingGateRemovalSeconds(podconstants.SchedulingGateName, wl.Status.Admission.ClusterQueue, p.isGroup, latency)
```
mimowo: Hm, I don't think this is the correct interpretation of the metric. I think each ungated pod should contribute a new datapoint to the histogram.

mbobrovskyi: Actually, I think you're right. I was looking at this metric from the workload perspective, but we need to expose a metric for each Pod. I've updated it accordingly.

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
```diff
@@ -209,6 +213,8 @@ var _ = ginkgo.Describe("Pod controller", ginkgo.Label("job:pod", "area:jobs"),
 					wlConditionCmpOpts...,
 				))
 
+				util.ExpectPodSchedulingGateRemovalSecondsMetricLessOrEqual(podconstants.SchedulingGateName, cqName, false, time.Second)
```
mimowo: Hold on, so we expect the delay to be <= 1s? This feels racy in case the CI build system is loaded. 

I think for integration tests we could assert like util.Timeout.

We can assert more precisely in unit tests where we have a full control over the clock.

pkg/controller/jobs/pod/pod_controller_test.go
```diff
@@ -133,12 +133,14 @@ func TestPodsReady(t *testing.T) {
 
 func TestRun(t *testing.T) {
```
mimowo, line 136: Please add some test cases for the reconcile that assert on the new metric.

pkg/metrics/metrics.go
```diff
@@ -573,6 +578,14 @@ The label 'underlying_cause' can have the following values:
 		}, append([]string{"job_kind", "replica_role"}, extraLabels...),
 	)
 
+	PodSchedulingGateRemovalSeconds = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: constants.KueueName,
+			Name:      "pod_scheduling_gate_removal_seconds",
+			Help:      "The time between workload admission and scheduling gate removal.",
+		}, append([]string{"name", "cluster_queue", "is_group"}, extraLabels...),
```
mimowo, line 586: Hm, I'm not sure about the is_group - I know this was requested in the issue, and it makes **some** sense, but I'm not sure about some cases.

In particular, should we could "is_group" as true when the Pod is owned by a framework which is using PodGroups internally, like LWS; or just user-defined Pod Groups. I suppose is_group is mostly relevant for top-level PodGroups, but let me check if @ichekrygin has an opinion here.

mimowo: In any case, please clarify the current semantics in a comment. I would tentatively assume we only count the user-defined PodGroups. It is only technical detail that LWS is using PodGroups underneath I think

ichekrygin: > I suppose is_group is mostly relevant for top-level PodGroups

Yes, this was the thought behind is_group: to distinguish "simple" pods from "pod groups" in the context of scheduling gate removal.

ichekrygin: But I hear your concern for cases with higher-level Kueue frameworks leveraging pod integration. Although I don't have a strong opinion or suggestion on how to address that in terms of metric labels.

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
mbobrovskyi: I'm also thinking maybe we should use milliseconds for this metrics to be more precise.

@mimowo WDYT?

mimowo: I don't think so, users need the metrics to find our relevant delays, not to measure with sub-second precision. See the motivation from the issue description **The gate is eventually removed, but delay can reach minutes in production clusters.**.

This is also my experience on-calling some of the clusters, sometimes the ungating takes minutes, and it makes users unhappy. In that case we can most often mitigate the issue by increasing the controller concurrency. However, the users expect to have some metric to see the improvement.

pkg/controller/jobs/pod/pod_controller_test.go
mbobrovskyi: Done

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
mbobrovskyi: Done

pkg/controller/jobs/pod/pod_controller_test.go
mbobrovskyi: Done

mimowo:
/release-note-edit
```release-note
Observability: Added the `kueue_pod_scheduling_gate_removal_seconds` histogram metric to
measure the time from Workload admission to Pod scheduling-gate removal, helping operators
track delays before admitted Pods can be scheduled.
```

pkg/controller/jobs/pod/pod_controller.go
```diff
@@ -320,10 +324,22 @@ func (p *Pod) Run(ctx context.Context, c client.Client, podSetsInfo []podset.Pod
 		if recorder != nil {
 			recorder.Eventf(pod, nil, corev1.EventTypeNormal, jobframework.ReasonStarted, "Started", msg)
 		}
+
+		p.recordPodSchedulingGateRemovalSeconds(wl)
```
mimowo, line 328: if I have a workload with 1000 pods the metric should be emitted 1000 times. cc @ichekrygin is that assumption matching your expectation?

IIUC the currently implementation makes it random.

mimowo: Basically, I cannot see if the current approach would transfer to other scheduling gates. Also it is very hard to interpret for users, because the timings between the pods may vary signifficantly within a group. So, I think we need the metric to be per pod effectively.

pkg/controller/jobs/pod/pod_controller_test.go
```diff
@@ -376,6 +380,10 @@ func TestReconciler(t *testing.T) {
 		wantEvents        []utiltesting.EventRecord
 		reconcilerOptions []jobframework.Option
 		featureGates      map[featuregate.Feature]bool
+
+		isGroup                             bool
+		wantPodSchedulingGateRemovalSeconds *float64
+		wantPodSchedulingGateRemovalCount   *uint64
```
mimowo: What is the semantics of these asserts? Please add some comments because I cannot understand by name? Instead of the very dedicated checks, could we have something more generic, as we assert on metrics for example here https://github.com/kubernetes/kubernetes/blob/05e1adfaecf2d4a369f06daaf659944e5ceb54c9/test/integration/job/job_test.go#L1729

I'm also ok to defer it for later if this would be too much changes, trying to understand for now.

pkg/controller/jobs/pod/pod_controller.go
mimowo: Oh, sorry, I think it might be ok as you did - already per pod. Do we have test cases which show that the metric is reported per pod, when the group has two pods?

pkg/controller/jobs/pod/pod_controller_test.go
```diff
@@ -425,9 +433,9 @@ func TestReconciler(t *testing.T) {
 								Assignment(corev1.ResourceCPU, "unit-test-flavor", "1").
 								Obj()).
 							Obj(),
-						now,
+						now.Add(-2*time.Second),
 					).
-					AdmittedAt(true, now).
+					AdmittedAt(true, now.Add(-2*time.Second)).
```
mimowo: Why adjusting pre-existing tests ? I would much rather have a new dedicated test case than change pre-existing ones in solve very non-obvious way.

pkg/metrics/metrics.go
```diff
@@ -573,6 +578,14 @@ The label 'underlying_cause' can have the following values:
 		}, append([]string{"job_kind", "replica_role"}, extraLabels...),
 	)
 
+	PodSchedulingGateRemovalSeconds = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: constants.KueueName,
+			Name:      "pod_scheduling_gate_removal_seconds",
+			Help:      "The time between workload admission and scheduling gate removal.",
```
mimowo:
```suggestion
			Help:      ""Duration from Workload admission to removal of a Pod scheduling gate.",
```

pkg/controller/jobs/pod/pod_controller_test.go
mimowo: We basially need to cover:
1. one workload for pod group with two pods, they remove the scheduling gates with different delays
2. one workload for pod group with two pods, they remove the scheduling gates at the same time
2. one workload with one pod (no group)

mbobrovskyi: Done

pkg/metrics/metrics.go
mbobrovskyi: Done

pkg/controller/jobs/pod/pod_controller_test.go
```diff
@@ -6187,6 +6191,304 @@ func TestReconciler(t *testing.T) {
...
+		clockStep          bool
+		isGroup            bool
+		pods               []corev1.Pod
+		workloads          []kueue.Workload
+		wantPods           []corev1.Pod
+		wantMetricsCount   uint64
+		wantMetricsSeconds float64
+		wantErr            error
+	}{
+		"one workload for pod group with two pods; they remove the scheduling gates with different delays": {
```
mimowo, line 6214: Could you elaborate in a comment how this test is exercising the scheduling gates to be lifefted at different times? 

I would assume one of the Pod is already scheduling ungated in the group, for example using the fast-admission option.

pkg/controller/jobs/pod/pod_controller_test.go
```diff
@@ -6187,6 +6191,304 @@ func TestReconciler(t *testing.T) {
...
+	now := time.Now().Truncate(time.Second)
+
+	const (
+		rfName = "rf"
+		cqName = "cq"
+		lqName = "lq"
+	)
+
+	testCases := map[string]struct {
+		clockStep          bool
```
mimowo: How is this param relevant?

pkg/controller/jobs/pod/pod_controller_test.go
mbobrovskyi: Done

mimowo (commented):
/lgtm
/approve
Thank you for the new metric 👍 Please follow up with the update for the topology and elastic-job gates.

Issue #12124, opened 2026-06-11
Reduce kueue controller-manager restarts in the E2E suites


ikchifo:
**What would you like to be cleaned**:

Reduce how often the e2e suites restart the kueue controller manager, and make each remaining restart cheaper:

1. Drop restarts no test depends on: the `AfterSuite` config resets in the sequential suites (`test/e2e/sequential/baseline/suite_test.go:67`, `test/e2e/sequential/extended/suite_test.go:73`) restore the default configuration moments before the kind cluster is torn down.
2. Parallelize grouped restarts: the `MultiKueueOrchestratedPreemption` block in `test/e2e/multikueue/sequential/e2e_test.go` restarts the manager and both worker clusters sequentially in `BeforeAll` and again in `AfterAll`. Restarting the three clusters concurrently cuts each block from roughly 75s to 25s.
3. Run the e2e deployment with one replica instead of two: `test/e2e/config/default/kustomization.yaml` sets `count: 2`, so every restart pays for a two-pod rolling update. A one-replica default halves the restart cost; a two-replica overlay covers the HA failover specs that need it.

**Why is this needed**:

Restarts dominate the sequential suites. Counting `Kueue restarted` events in recent CI build logs: 17 restarts in sequential-baseline, 6 in sequential-extended, and 12 per cluster (x3 clusters) in multikueue-sequential.

Each restart takes about 25s with two replicas. In a local A/B run of `make test-e2e-sequential-baseline` with `count: 1`, the mean restart dropped from 24.7s (n=15) to 13.8s (n=16) and the whole make target from 792s to 633s (-20%), with no new failures. Together, the three changes could save roughly 5+ minutes per CI cycle across the three suites.

Could contribute to #11606.

mimowo:
1. Great idea, we restart kueue in After suite just to revert the cluster if ran locally, but we can do this only when E2E_MODE=dev because this is useless in CI
2. Great finding
3. Probably yes, but haveing 2 Kueue instances might be testing Kueue against regressions for the HA mode. Maybe we could consider using single instance for extended suites which are longest, and also mostly focused on integrations rather than cache management. For HA cache is moreimportant

mimowo:
@ikchifo would you like to submit PRs?

ikchifo:
/assign

Yes! I'll start working on it.

mimowo:
[...]

PR #12139, opened 2026-06-11
Skip AfterSuite Kueue config resets outside dev mode

ikchifo:
#### What type of PR is this?

/kind cleanup
/area testing

The sequential baseline and extended e2e suites reset the Kueue
configuration to the default in AfterSuite, which restarts the
controller manager and waits for it to become available (~25s each).
The reset exists only to leave a locally kept cluster in a clean state.
In CI the cluster is deleted right after the run, so the restart is
wasted time.

This PR adds `util.IsE2EModeDev()` and gates both AfterSuite resets on
it, as suggested in
https://github.com/kubernetes-sigs/kueue/issues/12124#issuecomment-4683845695.
Local runs with `E2E_MODE=dev` keep the current behavior.

Part of #12124, contributes to #11606.

The multikueue AfterSuites are untouched: they clean up shared secrets,
not configuration, so there is nothing to skip.

```release-note
NONE
```

/cc @mimowo 

## Summary by CodeRabbit

* **Tests**
  * Optimized end-to-end test teardown behavior to conditionally skip cluster restart steps during local development, reducing test cycle time while maintaining full cleanup in CI environments.

### DIFF
--- a/test/e2e/sequential/baseline/suite_test.go
+++ b/test/e2e/sequential/baseline/suite_test.go
@@ -64,5 +64,7 @@ var _ = ginkgo.BeforeSuite(func() {
 })
 
 var _ = ginkgo.AfterSuite(func() {
-	util.UpdateKueueConfigurationAndRestart(ctx, k8sClient, defaultKueueCfg, kindClusterName)
+	if util.IsE2EModeDev() {
+		util.UpdateKueueConfigurationAndRestart(ctx, k8sClient, defaultKueueCfg, kindClusterName)
+	}
 })
--- a/test/util/e2e.go
+++ b/test/util/e2e.go
@@ -83,6 +83,14 @@ const (
 	DefaultMetricsServiceName = "kueue-controller-manager-metrics-service"
 )
 
+// IsE2EModeDev returns true when E2E_MODE is set to "dev".
+// Use this to skip teardown steps that only make sense when the cluster
+// is kept around after the test run (local development), but are
+// unnecessary in CI where the cluster is ephemeral.
+func IsE2EModeDev() bool {
+	return os.Getenv("E2E_MODE") == "dev"
+}
+
 func GetKueueNamespace() string {
 	if ns := os.Getenv("KUEUE_NAMESPACE"); ns != "" {
 		return ns
--- a/test/e2e/sequential/extended/suite_test.go
+++ b/test/e2e/sequential/extended/suite_test.go
@@ -70,5 +70,7 @@ var _ = ginkgo.BeforeSuite(func() {
 })
 
 var _ = ginkgo.AfterSuite(func() {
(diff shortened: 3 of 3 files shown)


test/e2e/sequential/baseline/suite_test.go
```diff
@@ -64,5 +64,7 @@ var _ = ginkgo.BeforeSuite(func() {
 })
 
 var _ = ginkgo.AfterSuite(func() {
-	util.UpdateKueueConfigurationAndRestart(ctx, k8sClient, defaultKueueCfg, kindClusterName)
+	if util.IsE2EModeDev() {
+		util.UpdateKueueConfigurationAndRestart(ctx, k8sClient, defaultKueueCfg, kindClusterName)
+	}
```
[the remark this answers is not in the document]
ikchifo: The panic can't happen. The generated `DeepCopy()` is nil-safe (`zz_generated.deepcopy.go`: `if in == nil { return nil }`). 
Also, this call was previously unconditional in AfterSuite, so the PR narrows the exposure to dev mode rather than adding it.

test/util/e2e.go
```diff
@@ -83,6 +83,14 @@ const (
 	DefaultMetricsServiceName = "kueue-controller-manager-metrics-service"
 )
 
+// IsE2EModeDev returns true when E2E_MODE is set to "dev".
+// Use this to skip teardown steps that only make sense when the cluster
+// is kept around after the test run (local development), but are
+// unnecessary in CI where the cluster is ephemeral.
+func IsE2EModeDev() bool {
+	return os.Getenv("E2E_MODE") == "dev"
+}
```
mbobrovskyi, line 92: What about case with E2E_RUN_ONLY_ENV?

mimowo: The idea of introducing `E2E_MODE` was to be able to drop the confusing `E2E_RUN_ONLY_ENV`. I think we should rather just remove it, do we still have any use cases for it? If not then I think it is time to remove it and cleanup the code.

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Let me merge as is since this will help all builds already. We can follow up on the comment for E2E_RUN_ONLY_ENV, but I would suggest to just drop it.

test/util/e2e.go
mbobrovskyi: Actually, it's useful when we don't need to run the tests and only want to create the environment.

mimowo: But this use-case can be easily done by non-matching --focus. It makes our code simpler, and is easier to remember.

mbobrovskyi: Another question is: what happens if we run these tests manually, even with DEV_MODE enabled? In that case, the config will never be cleaned up. We should think about how to optimize this for local runs.

mimowo: I think every test should cleanup in its own AfterAll  / AfterEach. If we don't do that already, then yet, we could improve that.

PR #12142, opened 2026-06-11
Restart worker Kueues in parallel in preemption E2E test

ikchifo:
#### What type of PR is this?

/kind cleanup
/area testing
/area multikueue

The `MultiKueueOrchestratedPreemption is enabled` block restarts the
kueue-controller-manager on the manager and both worker clusters, one
after another, in both BeforeAll and AfterAll: ~150s of restarts per
run at ~25s each.

The two worker restarts are independent, so this PR runs them in
parallel goroutines (guarded with `ginkgo.GinkgoRecover()` so a failed
assertion still fails the calling Ginkgo node). Each block drops from
~75s to ~50s.

The manager restart stays first and synchronous on purpose. Right
after its rollout, the new leader is still acquiring the lease and
reconnecting the MultiKueue remote clients; a workload reconciled in
that window finds no connected clusters and is never requeued when
they connect, so the test times out (reproduced locally). The worker
restarts give the manager that warm-up window, as the old sequential
ordering did implicitly.

Verified with a focused local run: manager restart ~25s, the two
worker restarts overlapping at ~25s combined, spec passes.

Part of #12124, contributes to #11606

The BeforeAll closure only reads the feature-gate map and DeepCopies
the config before mutating, so the two goroutines share no mutable
state.

```release-note
NONE
```

## Summary by CodeRabbit

* **Tests**
  * Added concurrent execution for restart operations in e2e tests to improve performance while maintaining reliability.
  * Enhanced assertion error handling and recovery mechanisms for parallel test operations.
  * Optimized the setup and teardown sequence for MultiKueueOrchestratedPreemption testing.

### DIFF
--- a/test/e2e/multikueue/sequential/e2e_test.go
+++ b/test/e2e/multikueue/sequential/e2e_test.go
@@ -19,6 +19,7 @@ package sequential
 import (
 	"fmt"
 	"os/exec"
+	"sync"
 
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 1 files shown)


mimowo:
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for the work on speeding up the tests 👍

PR #12192, opened 2026-06-12
Run extended e2e suites with a single Kueue replica

ikchifo:
#### What type of PR is this?

/kind cleanup
/area testing

Every Kueue restart in the E2E suites pays for two controller-manager
pods plus leader election handover. My local runs showed the median
restart dropping from 24.9s with two replicas to 13.8s with one, and
the sequential suite from 638.8s to 535.8s.

From https://github.com/kubernetes-sigs/kueue/issues/12124#issuecomment-4683845695, two replicas stay as the default for HA-regression 
coverage; only the extended suites switch to one replica, since they 
are the longest and mostly exercise integrations rather than cache management.

Contributes to #11606 ., Part of #12124

- The helm-based extended targets (`test-e2e-extended-helm`,
  `test-tas-e2e-extended-helm`, `test-e2e-sequential-extended-helm`)
  still deploy two replicas: the helm path uses
  `test/e2e/config/default/values.yaml` and bypasses
  `E2E_CONFIG_FOLDER`. Left out of scope here to keep the change
  small; happy to follow up if wanted.
- The multikueue extended target is untouched; it dispatches its own
  config via `hack/testing/e2e-multikueue-test.sh`.

```release-note
NONE
```

## Summary by CodeRabbit

* **Tests**
  * Extended e2e test configuration added to support comprehensive end-to-end testing scenarios.


mimowo:
@ikchifo I'm still open to discuss using 1 replica also for "baseline" tests. 

In the past we didn't have the HA Failover suite so using 2 replicas unconditionally was our idea for testing that scenarios. However currently I think using the dedicated test only might be a good idea - it would make a clear indication for where are the HA-related tests.

Could you maybe open another experimental PR to also enable single replica for the baseline tests so that we can assess how much we gain in terms of CI execution times?

ikchifo:
@mimowo Sounds good, I agree the HA Failover suite is now the right home for the 2-replica scenarios. I'll open that experimental PR and drop thee CI time difference here

mimowo:
The timings here are impressive, only 16min: https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/pr-logs/directory/pull-kueue-test-e2e-extended-shard-2-main-1-36

mimowo:
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you, this is an excellent build time improvement 👍

PR #12193, opened 2026-06-12
9988 effective resource groups proposal 2

Singularity23x0:

Fixes #

```release-note

```


pkg/controller/core/quota_manager.go
```diff
@@ -0,0 +1,99 @@
...
+import (
+	"context"
+	"reflect"
+	"sync"
+
+	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+)
+
+// Manager trigger specific updates in order.
+type QuotaUpdateStep = func(ctx context.Context, cq *kueue.ClusterQueue, cache *QuotaCache) (cont bool, err error)
```
olekzabl, line 35: By giving the whole `*QuotaCache` to a `QuotaUpdateStep`, we allow that step to mess with all levels of computation, arbitrarily.

How about:
* replacing a fixed enumeration `spec`, `mkAggregatedQuota`, `effectiveQuota` with a generic array of "stages", each of type `map[kueue.ClusterQueueReference][]kueue.ResourceGroup`
  (i.e. add `[]` for the MK stage, why not)
* passing only a single element of that array as an input to `QuotaUpdateStep`, and expecting it to produce an output of the same format (which the QM will then put into the next array entry)

?

pkg/controller/core/clusterqueue_controller.go
```diff
@@ -174,6 +197,10 @@ func (r *ClusterQueueReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		)
 	}
 
+	if err := r.triggerSpecUpdate(ctx, &cqObj); err != nil {
```
olekzabl: Is there any chance that `r.triggerSpecUpdate` is `nil` at this point?
E.g. because it's Kueue binary startup and we haven't yet managed to "register" all callbacks?
If so, is this going to panic (not just err)?
Please double-check.

(This may be obsoleted, or at least shifted, by part 2 of [my other comment](https://github.com/kubernetes-sigs/kueue/pull/12193#discussion_r3403308579), though)

pkg/controller/core/quota_manager.go
```diff
@@ -0,0 +1,102 @@
...
+func NewQuotaManager(opts *QuotaManagerOpts) *QuotaManager {
+	qm := &QuotaManager{
+		cache: &QuotaCache{
+			spec:              nil,
+			mkAggregatedQuota: nil,
+		},
+		updateChain: make([]QuotaUpdateStep, 0),
+	}
+
+	qm.addUpdateStep(opts.CoreCQRec.updateSpec, &opts.CoreCQRec.triggerSpecUpdate)
```
olekzabl: 1. As discussed offline, I'm not a fan of passing `&triggerSpecUpdate` here; I'd rather let `addUpdateStep` return a result and then do the right thing with it.
  The analogy with `client.Get(&resource, ...)` does not hold in my opinion because `Get` needs that trick for strict (generic-like) typing. Here, it's not needed.

2. I'm also not sure if we need this `triggerSpecUpdate` field _at all_.
  What if the Core CQ Rec just did
   ```golang
   qm.TriggerChain(ctx, r.updateSpec, cq)
   ```
   and then, the `QuotaManager` would find `r.updateSpec` in its `.updateChain`, determine the index, and call `triggerChain()` for that index?
   This might fail technically; IIUC functions are not _comparable_ in Go (cf. [here](https://go.dev/blog/comparable#comparable-types) or [here](https://go.dev/blog/maps#key-types)).
   But then, we should be able to _wrap around_ this, by introducing some pointers, or interfaces, or modeling `UpdateSpec` as a `struct {name someEnum, handler func}`, where we'd only compare the names.

Singularity23x0: 1. Copy. Will revert to previous variant.
2. We could work with aliases (name denoting a stage) or even ids (which eliminates the issue of potential name conflicts as QM generates the ids). It is a matter of semantics. I prefer the callback as it is a single param - instead of having saving the qm and an id, we have one callable with the id baked into it.

PR #12196, opened 2026-06-12
Run baseline e2e suites with a single Kueue replica

ikchifo:
#### What type of PR is this?

/kind cleanup
/area testing

Every Kueue rollout in the e2e suites pays for a second
controller-manager pod and a leader election handover. This PR drops
the singlecluster and TAS baseline suites to one replica, as @mimowo
suggested in #12192, which did the same for the extended suites.

The sequential baseline suite keeps two replicas. It hosts the HA
Failover tests (`test/e2e/sequential/baseline/default_config_test.go`),
which admit workloads into both the leader and follower caches, force a
handover with `util.ForceLeaderFailover`, and check that the former
follower schedules from its warm cache. With one replica the
lease-holder assertion would still pass, but the test would stop
exercising the follower cache.

Part of #11606

- This is the experiment suggested in
  https://github.com/kubernetes-sigs/kueue/pull/12192#issuecomment-4690493555:
  apply the single-replica config to the baseline suites and watch CI times.
- The helm baseline targets still deploy two replicas; the helm path
  reads `test/e2e/config/default/values.yaml` and ignores
  `E2E_CONFIG_FOLDER`. Out of scope here.
- The multikueue targets are untouched; they dispatch their own config
  via `hack/testing/e2e-multikueue-test.sh`.
- Independent of #12192 (different directory); the PRs can merge in either order.

```release-note
NONE
```

## Summary by CodeRabbit

* **Tests**
  * End-to-end test runs now default to the baseline configuration for baseline scenarios; updated E2E metrics expectations to reflect refined cleanup behavior and ensure stable assertions after resource deletions.

* **Chores**
  * Added a baseline Kustomize configuration to provide a consistent test deployment baseline, including controller replica overrides.


ikchifo:
The four baseline job failures were from the same issue. The metrics test expects `LocalQueue` counters (e.g. `kueue_local_queue_quota_reserved_workloads_total`) to disappear after deleting only the `ClusterQueue`, but Kueue clears those series only when the `LocalQueue` itself is deleted.

With two replicas the assertion passed by accident: the test scrapes the metrics Service, which load-balances between pods, and the follower never creates those series, so any scrape landing on it satisfied the "metric absent" check. With one replica every scrape hits the leader, exposing this.

Updated the test in [`034134a` (this PR)](https://github.com/kubernetes-sigs/kueue/pull/12196/commits/034134a38bd05cebc42becef18c3f74719fe3fc4) to expect `LocalQueue` counters and `kueue_local_queue_status` to survive `ClusterQueue` deletion, and added a step that deletes the LocalQueue and verifies its metrics are then removed, so the cleanup path is now actually exercised on the leader.

🤞 that this works

mimowo:
I see think you for the explanation. The proposed change to the assert makes total sense to me, but de-couple to a separate PR.

mimowo:
Actually, I'm also ok to couple, it is all to be cherrypicked anyway. Sorry for confusion if you already started the split

/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17

Issue #12176, opened 2026-06-12
Tune KubeRay test parameters for shorter CI execution times


mimowo:
**What would you like to be cleaned**:

I would like to runt the KubeRay test parameters for shorter CI execution times

For example look here: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12172/pull-kueue-test-e2e-extended-shard-2-release-0-18-1-35/2065360364047962112

We see that "End To End Extended Suite: kueue/kind-node:v1.35.5: [It] Kuberay Should run a rayjob with InTreeAutoscaling [area:singlecluster, feature:kuberay, Serial, shard:kuberay-b]" took 3min 30s. This is too long.

I think the main reason is because we use 60s tasks. I think this number was arbitrary and never tunes. Using 30s should also be save and already improve the build times by 1min.

One approach we could consider - see what is the timeout to make the jobs sometimes fail, and starting from here try to estimante how much we need to extend the timeout to be safe.

**Why is this needed**:

To improve the performence of the KubaRay builds which are very long

Part of https://github.com/kubernetes-sigs/kueue/issues/11606

mimowo:
cc @sohankunkerkar as you gained a lot of experience with KubeRay, any guidance on tuning of the params will be welcome

ikchifo:
/assign

Also fine to work on this. I've worked with KubeRay a bunch

mimowo:
/close 
as the task is completed, we just need to CP to 0.17, but I think we can track it outside the issue

PR #12213, opened 2026-06-12
Tune KubeRay e2e task durations for faster CI

ikchifo:
#### What type of PR is this?

/kind cleanup
/area testing

Two specs dominate the kuberay CI shards, per the junit from the prow
run linked in #12176:

| Spec | Shard | CI duration |
|---|---|---|
| Should run a rayjob with InTreeAutoscaling | kuberay-b | 206s |
| Should run a rayjob with multi scale-up steps | kuberay-a | 179s |

**InTreeAutoscaling**: replace the 3 tasks of 60s with a queue of 20
tasks of 10s (the same ~200 task-seconds of work).

The long tasks make the job end late: the deleted worker's task
restarts on the replacement pod, adding ~85s after replacement is
verified. Shorter sleeps are unsafe: scale-up takes ~40s in CI, so
30s tasks finish before the second worker starts and the autoscaler
(idleTimeoutSeconds=10, minReplicas=1) scales the idle first worker
down under the "2 running workers" assertions. The queue avoids
both: pending tasks hold worker demand through verification, and the
job drains within ~10s once the queue empties.

**Multi scale-up**: cut the trailing keep-alive tasks from 32 to 20.
They only need to outlive scale-down verification (the 10s idle
timeout plus an annotation update); 20 tasks (~25s) do.

Local A/B runs on a kind cluster:

| Spec | main | this PR |
|---|---|---|
| multi scale-up | 120-123s | 107-110s |
| InTreeAutoscaling | 118-121s | 116-117s |

Locally the pod deletion lands early, so main's tail barely shows.
In CI it lands late (the ~85s tail in the 205.7s baseline); the
queue's tail stays bounded no matter when the deletion lands, so the
spec gets faster and more deterministic in CI.

Part of #12176, contributes to #11606

The verification steps and timeouts are unchanged; only the Ray task
payloads are tuned.

```release-note
NONE
```

### DIFF
--- a/test/e2e/singlecluster/extended/kuberay_test.go
+++ b/test/e2e/singlecluster/extended/kuberay_test.go
@@ -242,10 +242,14 @@ def my_task(x, s):
     time.sleep(s)
     return x * x
 
-# 3 parallel tasks with 60s sleep: triggers scale-up to 2 workers
-# (1 task on head + 2 on workers) and keeps them alive through
-# pod replacement verification.
-print(ray.get([my_task.remote(i, 60) for i in range(3)]))`,
+# A queue of 20 short tasks (~200 task-seconds over at most 3 slots:
+# 1 on head + 1 per worker) triggers scale-up to 2 workers and keeps
+# both busy through pod replacement verification: pending tasks hold
+# the demand no matter how long scale-up or replacement takes, and the
+# job drains quickly once the queue is empty. With a few long tasks
+# instead, their expiry races the autoscaler's idle scale-down
+# (idleTimeoutSeconds=10) during verification.
+print(ray.get([my_task.remote(i, 10) for i in range(20)]))`,
 			},
 		}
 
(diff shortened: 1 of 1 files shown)


mimowo:
cc @sohankunkerkar ptal

mimowo:
/test all
retry to see stability

sohankunkerkar (approved):
I ran these tests locally on Kind (in loop), all passed. I confirm the numbers from the PR description with stable pod replacement timing. Further squeezing is possible but risks flakes. This is a good balance.
/lgtm
/hold in case @mimowo has anything to add here

mimowo:
Maybe this is over-optimizing, but could we reduce the task lengths if we make the "idleTimeout" to 1s?

In any case, I'm happy to merge already as is, because this is a great improvement already. We can follow up if you like the idea of decreasing the "idleTimeout".

/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17

Thank you, the results look great 👍

mimowo:
@ikchifo what about CP to 0.17, would that be feasible?

PR #7439, opened 2025-10-30
MultiKueue remote client kubeconfig validation

mszadkow:
#### What type of PR is this?
/kind bug

#### What this PR does / why we need it:
Add MultiKueue remote client kubeconfig validation to reduce the risk of insecure configuration.
Introduce a feature gate: `MultiKueueAllowInsecureKubeconfigs` to allow insecure kubeconfigs until Kueue v0.17.0

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note
MultiKueue: validate remote client kubeconfigs and reject insecure kubeconfigs by default; add feature gate MultiKueueAllowInsecureKubeconfigs to temporarily allow insecure kubeconfigs until v0.17.0.

ACTION REQUIRED: if you are using MultiKueue kubeconfigs which are not passing the new validation please
enable the `MultiKueueAllowInsecureKubeconfigs` feature gate and let us know so that we can re-consider
the deprecation plans for the feature gate.
```

PR #12223, opened 2026-06-12
secure path to avoid path traversal issues

### LINKED ISSUE #12144: Secure the use of Path for KubeConfigs for MultiKueue
MultiKueue has either KubeConfig support via Secret or Path. The safest option is to use ClusterProfile which is the recommendation for production.

The support of Path reads directly from a file and is not recommended for production.

I think we should consider deprecated Path and either require ClusterProfile or Secret.

kannon92:
#### What type of PR is this?
/kind feature
/area multikueue

Fixes #12144 

https://owasp.org/www-community/attacks/Path_Traversal

```release-note
MultiKueue: Fixed a security vulnerability in `locationType=Path` kubeconfig handling
that could allow users with `MultiKueueCluster` create or update access to make the
controller read arbitrary files. Kueue now validates path-based kubeconfigs to stay under
`/etc/multikueue/kubeconfigs`.

ACTION REQUIRED: If you use `locationType=Path`, plan to move kubeconfig files under
`/etc/multikueue/kubeconfigs`, or switch to `locationType=Secret` or `ClusterProfile`.
This prepares your setup for future releases where `MultiKueueKubeConfigPathValidation`
is expected to be enabled by default.
```

## Summary by CodeRabbit

* **New Features**
  * Added the `SafeMultiKueueConfigPath` feature gate (enabled by default) to enforce safe `locationType=Path` kubeconfig handling: rejects empty/relative/`..` paths, blocks symlink escapes, confines reads to `/etc/multikueue/kubeconfigs/`, and avoids filesystem watcher registration for invalid paths. Disable the gate to restore legacy behavior.
* **Documentation**
  * Added security considerations and setup guidance for using `Path` kubeconfigs, including recommended `Secret`/`ClusterProfile` alternatives.
* **Tests**
  * Added kubeconfig path validation coverage and updated integration setup for test-specific path prefixing.

### DIFF
--- a/keps/693-multikueue/README.md
+++ b/keps/693-multikueue/README.md
@@ -152,6 +152,23 @@ deployments with very high job throughput (above 1M jobs per day).
     learning the state of the world from the worker clusters (not covered
     by this KEP).
 
+* **Arbitrary file read via `locationType=Path`**: When `KubeConfig.LocationType`
+  is set to `Path`, the controller reads the file at the user-supplied location
+  using `os.ReadFile`. Without path restrictions, any principal with
+  `create`/`update` access to `MultiKueueCluster` resources could read arbitrary
+  files from the controller pod's filesystem (e.g. the projected service account
+  token). This is mitigated by:
+  * Gating safe path validation behind the `MultiKueueKubeConfigPathValidation` feature
+    gate (Alpha, default disabled). When the gate is on, only
+    paths under the hardcoded prefix `/etc/multikueue/kubeconfigs/` are
+    accepted. When disabled, the controller falls back to the legacy
+    behavior of allowing any path.
+  * Canonicalizing and validating the path: the controller cleans the path,
+    rejects `..` segments, resolves symlinks, and requires the resolved absolute
+    path to reside under the prefix directory.
+  * Recommended deployment practice is to use `locationType=Secret` or
+    `ClusterProfile` instead of `Path` in production environments.
+
 ## Design Details
 MultiKueue will be enabled on a cluster queue using the admission check fields.
 Just like ProvisioningRequest, MultiKueue will have its own configuration, 
--- a/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
+++ b/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
@@ -24,6 +24,7 @@ import (
 	"net"
 	"net/url"
 	"os"
+	"path/filepath"
 	"slices"
 	"strings"
 	"sync"
@@ -45,6 +46,7 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/clock"
@@ -574,9 +576,16 @@ func (rc *remoteClient) runGC(ctx context.Context) {
 
 // clustersReconciler implements the reconciler for all MultiKueueClusters.
 // Its main task being to maintain the list of remote clients associated to each MultiKueueCluster.
+// defaultKubeConfigPathPrefix is the hardcoded directory under which
+// kubeconfig files must reside when the MultiKueueKubeConfigPathValidation feature
+// gate is enabled (the default).
+const defaultKubeConfigPathPrefix = "/etc/multikueue/kubeconfigs"
+
 type clustersReconciler struct {
-	localClient     client.Client
-	configNamespace string
+	localClient          client.Client
+	configNamespace      string
+	kubeConfigPathPrefix string
+	recorder             record.EventRecorder
 
 	lock sync.RWMutex
 	// The list of remote remoteClients, indexed by the cluster name.
@@ -710,6 +719,15 @@ func (c *clustersReconciler) Reconcile(ctx context.Context, req reconcile.Reques
 
 	log := ctrl.LoggerFrom(ctx)
 	log.V(2).Info("Reconcile MultiKueueCluster")
+
+	// Warn about deprecated Path usage when the validation feature gate is off.
+	if cluster.Spec.ClusterSource.KubeConfig != nil &&
+		cluster.Spec.ClusterSource.KubeConfig.LocationType == kueue.PathLocationType &&
+		!features.Enabled(features.MultiKueueKubeConfigPathValidation) {
+		c.recorder.Event(cluster, corev1.EventTypeWarning, "DeprecatedPathUsage",
+			"Using locationType=Path without MultiKueueKubeConfigPathValidation feature gate is deprecated and will be removed in a future release. "+
+				"Enable the MultiKueueKubeConfigPathValidation feature gate and place kubeconfig files under /etc/multikueue/kubeconfigs/.")
+	}
 
 	if err != nil || !cluster.DeletionTimestamp.IsZero() {
 		c.stopAndRemoveCluster(req.Name)
@@ -901,8 +919,70 @@ func (c *clustersReconciler) getKubeConfigFromSecret(ctx context.Context, secret
 	return kconfigBytes, nil
 }
 
-func (c *clustersReconciler) getKubeConfigFromPath(path string) ([]byte, error) {
-	return os.ReadFile(path)
+// errPathNotAllowed is returned when the resolved kubeconfig path escapes the
+// configured allowed prefix directory.
+var errPathNotAllowed = errors.New("kubeconfig path is not under the allowed prefix")
+
+// validateKubeConfigPath resolves symlinks and ensures the resulting absolute
+// path is located under allowedPrefix when the MultiKueueKubeConfigPathValidation
+// feature gate is enabled (the default). When the gate is disabled, any
+// path is accepted (legacy unsafe behavior).
+func validateKubeConfigPath(ctx context.Context, rawPath, allowedPrefix string) (string, error) {
+	if !features.Enabled(features.MultiKueueKubeConfigPathValidation) {
+		// Legacy unsafe behavior: accept any path.
+		log := ctrl.LoggerFrom(ctx)
+		log.V(2).Info("Legacy unsafe behavior detected: this will be deprecated in the future.")
+		return rawPath, nil
+	}
+	if rawPath == "" {
+		return "", errors.New("kubeconfig path must not be empty")
+	}
+
+	cleaned := filepath.Clean(rawPath)
+
+	// Reject paths where any component is exactly ".." after cleaning.
+	if slices.Contains(strings.Split(cleaned, string(filepath.Separator)), "..") {
+		return "", fmt.Errorf("%w: path contains \"..\"", errPathNotAllowed)
+	}
+
+	// Require an absolute path.
+	if !filepath.IsAbs(cleaned) {
+		return "", fmt.Errorf("%w: path must be absolute", errPathNotAllowed)
+	}
+
+	// Resolve symlinks and get the absolute path of the allowed prefix.
+	// If the prefix directory does not exist, no kubeconfig can be under it.
+	resolvedPrefix, err := filepath.EvalSymlinks(allowedPrefix)
+	if err != nil {
+		return "", fmt.Errorf("%w: %q is not under %q", errPathNotAllowed, cleaned, filepath.Clean(allowedPrefix))
+	}
+	allowedAbs, err := filepath.Abs(resolvedPrefix)
+	if err != nil {
+		return "", fmt.Errorf("cannot resolve allowed prefix: %w", err)
+	}
+
+	// Resolve symlinks so an attacker cannot use a symlink to escape.
+	resolved, err := filepath.EvalSymlinks(cleaned)
+	if err != nil {
+		return "", fmt.Errorf("cannot resolve kubeconfig path symlinks: %w", err)
+	}
+
+	// Ensure the prefix directory boundary is respected.
+	// Add a trailing separator so "/etc/kueue" does not match "/etc/kueue-other".
+	prefixWithSep := allowedAbs + string(filepath.Separator)
+	if !strings.HasPrefix(resolved+string(filepath.Separator), prefixWithSep) {
+		return "", fmt.Errorf("%w: %q is not under %q", errPathNotAllowed, resolved, allowedAbs)
+	}
+
+	return resolved, nil
+}
+
+func (c *clustersReconciler) getKubeConfigFromPath(rawPath string) ([]byte, error) {
+	validated, err := validateKubeConfigPath(c.rootContext, rawPath, c.kubeConfigPathPrefix)
+	if err != nil {
+		return nil, err
+	}
+	return os.ReadFile(validated)
 }
 
 func (c *clustersReconciler) updateStatus(ctx context.Context, cluster *kueue.MultiKueueCluster, active bool, reason, message string) error {
@@ -967,10 +1047,13 @@ func newClustersReconciler(
 	adapters map[string]jobframework.MultiKueueAdapter,
 	cpAccessProvider clusterProfileAccessProvider,
 	roleTracker *roletracker.RoleTracker,
+	recorder record.EventRecorder,
 ) *clustersReconciler {
 	return &clustersReconciler{
 		localClient:                  c,
 		configNamespace:              namespace,
+		kubeConfigPathPrefix:         defaultKubeConfigPathPrefix,
+		recorder:                     recorder,
 		remoteClients:                make(map[string]*remoteClient),
 		wlUpdateCh:                   make(chan event.GenericEvent, eventChBufferSize),
 		watchEndedCh:                 make(chan event.GenericEvent, eventChBufferSize),
@@ -1043,8 +1126,9 @@ func (c *clustersReconciler) Create(e event.CreateEvent) bool {
 		log := c.logger().WithValues("multiKueueCluster", klog.KObj(cluster))
 		log.V(5).Info("MultiKueueCluster create event")
 		if cluster.Spec.ClusterSource.KubeConfig != nil && cluster.Spec.ClusterSource.KubeConfig.LocationType == kueue.PathLocationType {
-			err := c.fsWatcher.AddOrUpdate(cluster.Name, cluster.Spec.ClusterSource.KubeConfig.Location)
-			if err != nil {
+			if validated, err := validateKubeConfigPath(c.rootContext, cluster.Spec.ClusterSource.KubeConfig.Location, c.kubeConfigPathPrefix); err != nil {
+				log.Error(err, "Rejecting FS watch for invalid path")
+			} else if err := c.fsWatcher.AddOrUpdate(cluster.Name, validated); err != nil {
 				log.Error(err, "AddOrUpdate FS watch")
 			}
 		}
@@ -1071,8 +1155,9 @@ func (c *clustersReconciler) Update(e event.UpdateEvent) bool {
 	}
 
 	if clusterNewHasKubeConfigPath {
-		err := c.fsWatcher.AddOrUpdate(clusterNew.Name, clusterNew.Spec.ClusterSource.KubeConfig.Location)
-		if err != nil {
+		if validated, err := validateKubeConfigPath(c.rootContext, clusterNew.Spec.ClusterSource.KubeConfig.Location, c.kubeConfigPathPrefix); err != nil {
+			log.Error(err, "Rejecting FS watch for invalid path")
+		} else if err := c.fsWatcher.AddOrUpdate(clusterNew.Name, validated); err != nil {
 			log.Error(err, "AddOrUpdate FS watch")
 		}
 	}
--- a/pkg/controller/admissionchecks/multikueue/multikueuecluster_test.go
+++ b/pkg/controller/admissionchecks/multikueue/multikueuecluster_test.go
@@ -38,6 +38,7 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/utils/clock"
 	testingclock "k8s.io/utils/clock/testing"
 	inventoryv1alpha1 "sigs.k8s.io/cluster-inventory-api/apis/v1alpha1"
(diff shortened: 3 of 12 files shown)


pkg/controller/admissionchecks/multikueue/multikueuecluster_test.go
```diff
@@ -1411,3 +1413,99 @@ func TestSetRemoteClientConfigDoesNotBlockOtherClusters(t *testing.T) {
...
+			wantErr:       true,
+			errContains:   "not under",
+		},
+		"prefix name collision": {
+			// /etc/kueue-other should not match prefix /etc/kueue
+			path:          "/etc/kueue-other/file",
+			allowedPrefix: "/etc/kueue",
+			wantErr:       true,
+			errContains:   "not under",
+		},
```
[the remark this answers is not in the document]
kannon92: We don't really test on windows for kueue.

apis/config/v1beta2/configuration_types.go
```diff
@@ -335,6 +335,14 @@ type MultiKueue struct {
...
 	// +optional
 	ClusterProfile *ClusterProfile `json:"clusterProfile,omitempty"`
+
+	// KubeConfigPathPrefix defines the allowed directory prefix for kubeconfig
+	// files referenced by MultiKueueCluster objects using locationType=Path.
+	// Paths outside this directory are rejected. If empty (the default),
+	// locationType=Path is not allowed and only Secret or ClusterProfile
+	// sources may be used.
+	// +optional
+	KubeConfigPathPrefix *string `json:"kubeConfigPathPrefix,omitempty"`
```
mimowo: I'm proposing to hardcode the path, and using FG for safety. To avoid maintaining the extended API surface for this mode. The path could probably be something like: `/etc/multikueue/kubeconfigs/`


Also commented here: https://github.com/kubernetes-sigs/kueue/issues/12144#issuecomment-4695010023

cc @olekzabl @tenzen-y in case you have some other opinion.

kannon92: The problem with hard coding a path is that its not platform agnostic. `/etc/multikueue/kubeconfigs/` wouldn't work on windows and kube does support windows.

mimowo: We could have a dedicated hard coded  path for windows, and choose the one depending on OS detected, no?

 But tbh I think Windows on k8s makes more sense for running applications which require Windows. I cannot imagine any use case for deploying kueue on windows ;)

So I would also be ok to just limit the path for Linux like systems

kannon92: > But tbh I think Windows on k8s makes more sense for running applications which require Windows. I cannot imagine any use case for deploying kueue on windows ;)

There are HPC software that runs on windows. NVIDIA does also support windows but honestly it is probably a stretch as it doesn't look like we support DRA on windows atm.

kannon92: I'll go with hard coding to a path and we can review that.

kannon92: fixed.

pkg/features/kube_features.go
```diff
@@ -238,6 +238,15 @@ const (
...
 	MultiKueueAllowInsecureKubeconfigs featuregate.Feature = "MultiKueueAllowInsecureKubeconfigs"
 
+	// owner: @kannon92
+	// issue: https://github.com/kubernetes-sigs/kueue/issues/12144
+	//
+	// Enforce safe path validation for locationType=Path MultiKueueCluster
+	// kubeconfig references. When enabled (the default), only paths under
+	// /etc/multikueue/kubeconfigs/ are accepted. When disabled, the
+	// controller falls back to the legacy behavior of allowing any path.
+	SafeMultiKueueConfigPath featuregate.Feature = "SafeMultiKueueConfigPath"
```
```suggestion
	MultiKueueRestrictedKubeconfigLocation featuregate.Feature = "MultiKueueRestrictedKubeconfigLocation"
```
Mostly to make sure all MK features have the consistent prefix

mimowo: Or better name is maybe MultiKueueKubeConfigPathValidation

kannon92:
@mimowo 

The one thing I don't like about this PR is the reliant on the feature gate as a permanent flag.

If we merge this if we GA this feature gate we are essentially deprecating the insecure Path field.

Should we go with a new API field (SecurePath) so that we can not rely on feature gate?

mimowo:
> The one thing I don't like about this PR is the reliant on the feature gate as a permanent flag.

No, we will GA the FG if no-one complains for a couple of releases (which I assume will be the case). That is what we did historically for MultiKueueAllowInsecureKubeconfigs, just forgot to remove it.

> If we merge this if we GA this feature gate we are essentially deprecating the insecure Path field.

No, we will still support the path, but are restricted to the safe prefix.

> Should we go with a new API field (SecurePath) so that we can not rely on feature gate?

I think that was the reason for my original comment - I don't see the need to maintain larger API surface for the custom prefixes.

pkg/controller/admissionchecks/multikueue/multikueuecluster_test.go
```diff
@@ -191,12 +191,13 @@ func TestUpdateConfig(t *testing.T) {
...
-		wantCancelCalled       int
-		wantErr                error
-		skipInsecureKubeconfig bool
+		wantRemoteClients        map[string]*remoteClient
+		wantClusters             []kueue.MultiKueueCluster
+		wantRequeueAfter         time.Duration
+		wantCancelCalled         int
+		wantErr                  error
+		skipInsecureKubeconfig   bool
+		overrideKubeConfigPrefix bool
```
mimowo: Why parametrizing and pumping it trough so many layers. If hardcoded then we only need it inside pkg/controller/admissionchecks/multikueue/multikueuecluster.go as a non-exported value.

kannon92: updated.

pkg/features/kube_features.go
kannon92: done.

mimowo:
/release-note-edit
```release-note
MultiKueue: Fixed a security vulnerability in `locationType=Path` kubeconfig handling
that could allow users with `MultiKueueCluster` create or update access to make the
controller read arbitrary files. Kueue now validates path-based kubeconfigs to stay under
`/etc/multikueue/kubeconfigs`.
```

keps/693-multikueue/README.md
```diff
@@ -152,6 +152,23 @@ deployments with very high job throughput (above 1M jobs per day).
     learning the state of the world from the worker clusters (not covered
     by this KEP).
 
+* **Arbitrary file read via `locationType=Path`**: When `KubeConfig.LocationType`
+  is set to `Path`, the controller reads the file at the user-supplied location
+  using `os.ReadFile`. Without path restrictions, any principal with
+  `create`/`update` access to `MultiKueueCluster` resources could read arbitrary
+  files from the controller pod's filesystem (e.g. the projected service account
+  token). This is mitigated by:
+  * Gating safe path validation behind the `MultiKueueKubeConfigPathValidation` feature
```
mwielgus, line 161: This is potentially a breaking change. How can we ensure that the need for either disabling the feature gate or relocating configs are discoverable for the users who upgrade to the newest version?

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
```diff
@@ -905,8 +913,69 @@ func (c *clustersReconciler) getKubeConfigFromSecret(ctx context.Context, secret
...
+	}
+
+	// Resolve symlinks so an attacker cannot use a symlink to escape.
+	resolved, err := filepath.EvalSymlinks(cleaned)
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			// If the file doesn't exist yet we still validate the cleaned path
+			// against the prefix (the file will be read later and os.ReadFile
+			// will return a proper error).
+			resolved = cleaned
```
olekzabl: Why not return error right away?
Is your intent here to support some cases when "not found" would have been mitigated at the time of `os.ReadFile`?

But if so, it feels risky. What if the "not found, soon found" thing is itself an escaping symlink? We'd fail to catch it then?

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
```diff
@@ -905,8 +913,69 @@ func (c *clustersReconciler) getKubeConfigFromSecret(ctx context.Context, secret
...
+
+	// Reject paths where any component is exactly ".." after cleaning.
+	if slices.Contains(strings.Split(cleaned, string(filepath.Separator)), "..") {
+		return "", fmt.Errorf("%w: path contains \"..\"", errPathNotAllowed)
+	}
+
+	// Require an absolute path.
+	if !filepath.IsAbs(cleaned) {
+		return "", fmt.Errorf("%w: path must be absolute", errPathNotAllowed)
+	}
```
olekzabl, line 951: How about checking this after `EvalSymlinks`?
Then, we wouldn't need to call `.Abs` subsequently.

Is there value in allowing an "absolute path containing with a relative symlink"?

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
```diff
@@ -905,8 +913,69 @@ func (c *clustersReconciler) getKubeConfigFromSecret(ctx context.Context, secret
...
+			// If the file doesn't exist yet we still validate the cleaned path
+			// against the prefix (the file will be read later and os.ReadFile
+			// will return a proper error).
+			resolved = cleaned
+		} else {
+			return "", fmt.Errorf("cannot resolve kubeconfig path symlinks: %w", err)
+		}
+	}
+
+	allowedAbs, err := filepath.Abs(filepath.Clean(allowedPrefix))
```
olekzabl: Nits:

1. Given that `Abs` calls `Clean` on its result, I'm not sure calling `Clean` before `Abs` has value.
 
2. Shall we also `EvalSymLinks` on `allowedPrefix`?
   (Right now, IIUC, if the `allowedPrefix` itself contains a symlink, we'll reject every `path` here - is that intended?)

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
```diff
@@ -905,8 +913,69 @@ func (c *clustersReconciler) getKubeConfigFromSecret(ctx context.Context, secret
...
+	if err != nil {
+		return "", fmt.Errorf("cannot resolve allowed prefix: %w", err)
+	}
+
+	// Ensure the prefix directory boundary is respected.
+	// Add a trailing separator so "/etc/kueue" does not match "/etc/kueue-other".
+	prefixWithSep := allowedAbs + string(filepath.Separator)
+	if !strings.HasPrefix(resolved+string(filepath.Separator), prefixWithSep) && resolved != allowedAbs {
```
olekzabl: Nit: every string is a prefix of itself, so I think the latter condition is redundant.

```suggestion
	if !strings.HasPrefix(resolved+string(filepath.Separator), prefixWithSep) {
```

test/integration/multikueue/suite_test.go
```diff
@@ -386,6 +386,7 @@ func managerAndMultiKueueSetup(
 	gcInterval time.Duration,
 	enabledIntegrations sets.Set[string],
 	dispatcherName string,
+	extraOpts ...multikueue.SetupOption,
```
olekzabl: Is this used?
(I don't see changes affecting any call to `managerAndMultiKueueSetup` in this PR)

olekzabl (commented):
Nice, thank you!

A few minor comments.

keps/693-multikueue/README.md
kannon92: https://github.com/kubernetes-sigs/kueue/issues/12144

This was reported to us as a security issue and we are trying to figure out how to balance securing this insecure option while keeping an option for people to migrate?

It seems the recommendation is to move to use Secret or ClusterProfile. My thought would be to just deprecate this option but we thought maybe we should keep it around.

kannon92: https://github.com/kubernetes-sigs/kueue/issues/12144

This was reported to us as a security issue and we are trying to figure out how to balance securing this insecure option while keeping an option for people to migrate?

It seems the recommendation is to move to use Secret or ClusterProfile. My thought would be to just deprecate this option but we thought maybe we should keep it around.

mimowo: We need to reflect on that in the release note with the ACTION REQUIRED section. Feel free to propose something, I may also propose something later.

mimowo:
/release-note-edit
```release-note
MultiKueue: Fixed a security vulnerability in `locationType=Path` kubeconfig handling
that could allow users with `MultiKueueCluster` create or update access to make the
controller read arbitrary files. Kueue now validates path-based kubeconfigs to stay under
`/etc/multikueue/kubeconfigs`.

ACTION REQUIRED: If you use `locationType=Path`, move kubeconfig files under
`/etc/multikueue/kubeconfigs` before upgrading, or switch to `locationType=Secret` or
`ClusterProfile`. As a temporary measure, you can disable the
`MultiKueueKubeConfigPathValidation` feature gate, but plan to migrate as soon as possible.
```

mimowo:
/release-note-edit
```release-note
MultiKueue: Fixed a security vulnerability in `locationType=Path` kubeconfig handling
that could allow users with `MultiKueueCluster` create or update access to make the
controller read arbitrary files. Kueue now validates path-based kubeconfigs to stay under
`/etc/multikueue/kubeconfigs`.

ACTION REQUIRED: If you use `locationType=Path`, move kubeconfig files under
`/etc/multikueue/kubeconfigs` before upgrading, or switch to `locationType=Secret` or
`ClusterProfile`. As a temporary measure, you can disable the
`MultiKueueKubeConfigPathValidation` feature gate, but plan to migrate to one of these
supported options as soon as possible.
```

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
kannon92: I did try to move this but the tests started failing.

> Is there value in allowing an "absolute path containing with a relative symlink"?

 ### The realistic scenario: Kubernetes Secret volume mounts                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                          
 In production, the most common deployment pattern for MultiKueue kubeconfigs is mounting them via Kubernetes Secrets. When Kubernetes mounts a Secret as a volume, it creates a directory structure like this:                                                                           
                                                                                                                                                                                                                                                                                          
 ```                                                                                                                                                                                                                                                                                      
   /etc/multikueue/kubeconfigs/                                                                                                                                                                                                                                                           
   ├── ..data -> ..2026_06_17_12_00_00.123456789     # relative symlink to timestamped dir                                                                                                                                                                                                
   ├── ..2026_06_17_12_00_00.123456789/                                                                                                                                                                                                                                                   
   │   └── worker.kubeconfig                         # actual file                                                                                                                                                                                                                        
   └── worker.kubeconfig -> ..data/worker.kubeconfig  # relative symlink                                                                                                                                                                                                                  
 ```                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                          
 So when a user specifies the path /etc/multikueue/kubeconfigs/worker.kubeconfig, they're giving an absolute path containing relative symlinks. The chain is:                                                                                                                             
                                                                                                                                                                                                                                                                                          
 1. worker.kubeconfig → ..data/worker.kubeconfig (relative symlink)                                                                                                                                                                                                                       
 2. ..data → ..2026_06_17_12_00_00.123456789 (relative symlink)                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                          
 filepath.EvalSymlinks resolves all of this and produces the final real path, e.g., /etc/multikueue/kubeconfigs/..2026_06_17_12_00_00.123456789/worker.kubeconfig, which is still under the allowed prefix — so it passes validation correctly.                                           
                                                                                                                                                                                                                                                                                          
 ### What the code guards against                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                          
 The distinction is:                                                                                                                                                                                                                                                                      
 - Relative symlink within the prefix (like Kubernetes Secret mounts): ✅ Allowed — EvalSymlinks resolves it and the final path is still under the prefix.                                                                                                                                
 - Symlink that escapes the prefix (e.g., escape-symlink → /tmp/other-dir/secret): ❌ Rejected — EvalSymlinks resolves it to a path outside the prefix, and the strings.HasPrefix check catches it.                                                                                       
                                                                                                                                                                                                                                                                                          
 If you rejected all paths containing symlinks, you'd break the most common real-world deployment pattern (Secret-mounted kubeconfigs). The security goal isn't "no symlinks" — it's "the resolved final path must be under the allowed prefix." That's exactly what EvalSymlinks +       
 prefix check achieves.

olekzabl: > Is there value in allowing an "absolute path containing with a relative symlink"?

Ah, this question from me didn't make sense. 
`EvalSymlinks` will always return an absolute path when given an absolute path on its input. (Right?)

But, if so, then I guess calling `filepath.Abs(resolvedPrefix)` (on the result of `EvalSymlinks`) is a no-op?
(But that's a nit now; feel free to ignore).

pkg/features/kube_features.go
```diff
@@ -574,6 +583,10 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 		{Version: version.MustParse("0.15"), Default: false, PreRelease: featuregate.Alpha},
 		{Version: version.MustParse("0.17"), Default: false, PreRelease: featuregate.Deprecated}, // remove in 0.19
 	},
+
+	MultiKueueKubeConfigPathValidation: {
+		{Version: version.MustParse("0.19"), Default: true, PreRelease: featuregate.Beta},
```
mimowo: @tenzen-y I think during the wg-batch meeting you said you prefer starting with Alpha here, but since this is a vulnerability and we consider it a bugfix, maybe it is better to keep as beta and cherry-pick, wdyt? Yes, it is affecting the ootb and could potentially break some installation, but we could mitigate with the ACTION REQUIRED release note as proposed, wdyt?

tenzen-y: I still think that we should start with the Alpha phase. Otherwise, I think we should have the v1beta3 API.

Yes, this is actually due to less consideration when we introduced that.

But even with bug fixes, we should carefully handle API breaking, since I think this is the most impactful API break since Kueue launched. If we fully follow the K8S API guideline, we should make validations strong regardless of bug or security fix (e.g., gitRepo volume). 

We occasionally do that flexibly, but we should not change the behavior immediately as a default. We should provide a migration period. The webhook warning messages can inform us so dangerously.

mimowo: > I think this is the most impactful API break since Kueue launched. 

Hm, we chose the same approach here (FG: https://github.com/kubernetes-sigs/kueue/pull/7439, and went directly by validating by default. This case I would argue was a stronger break because it impacted KubeConfigs in secrets. Kubeconfigs in path is really niche use, I don't expect any user using it, as mentioned during the meeting I don't know about such a user from my end.

That being said I'm ok to go via Alpha, just wanted to clarify why I think it is ok to go directly with Beta too.

kannon92: Security issues do push us to do things we wouldn't normally do. Deprecating nginx-ingress.

Let me ask some folks in SRC what we do around this area.

tenzen-y: Yes, I understand that security issues do push us to think differently than we would normally. So, I approved validation changes. But, as much as possible, we should handle that carefully.

You can check gitRepo volume handling as an example.

mimowo: As mentioned I'm good with Alpha, I would just like to also cherrypick so that admins have more time to adjust their clusters it case it is affecting their clusters. 

> You can check gitRepo volume handling as an example.

Can you elaborate on this example? Was there a similar vulnerability story?

kannon92:
@mimowo @tenzen-y this is ready for review again.

olekzabl (approved):

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
```diff
@@ -711,6 +720,15 @@ func (c *clustersReconciler) Reconcile(ctx context.Context, req reconcile.Reques
...
 	log.V(2).Info("Reconcile MultiKueueCluster")
 
+	// Warn about deprecated Path usage when the validation feature gate is off.
+	if cluster.Spec.ClusterSource.KubeConfig != nil &&
+		cluster.Spec.ClusterSource.KubeConfig.LocationType == kueue.PathLocationType &&
+		!features.Enabled(features.MultiKueueKubeConfigPathValidation) {
+		c.recorder.Event(cluster, corev1.EventTypeWarning, "DeprecatedPathUsage",
+			"Using locationType=Path without MultiKueueKubeConfigPathValidation feature gate is deprecated and will be removed in a future release. "+
+				"Enable the MultiKueueKubeConfigPathValidation feature gate and place kubeconfig files under /etc/multikueue/kubeconfigs/.")
+	}
```
mimowo, line 730: Hm, I thought we decided not to deprecate the "Path" as such for now, just restrict to the allowed prefix. If we deprecate then it means removal at some point, which I think we wanted to avoid.

Let me know if something has changed, but if so then it should be mentioned in the KEP update too.

kannon92: Hmm this is sorta the issue with having an API field on an option we don't really want to support. Its impossible.

I think we still want to say that Path is insecure and not recommended for production. I hope that if we agree that Path isn't the solution we want to support maybe we should consider removal when we go to V1.

I'm fine to leave it in v1beta2 but I thought we agree that Path is not recommended approach for production and we recommend either Secret or ClusterProfile.

olekzabl: The proposed message here says

> Using locationType=Path **without MultiKueueKubeConfigPathValidation feature gate** is deprecated and will be removed in a future release

So my understanding is that the plan is to keep support for `Path` but eventually only with FG enabled.
So that we'll eventually forbid insecure values of `Path`.
Which feels to make sense and comply with the decision not to deprecate `Path` (completely).

WDYT?

mimowo: Ah, ok, so when the MultiKueueKubeConfigPathValidation is evanetally stable "locked to true", then this block of valiadtion will be removed completely. That makes sense.

mimowo:
/release-note-edit
```release-note
MultiKueue: Fixed a security vulnerability in `locationType=Path` kubeconfig handling
that could allow users with `MultiKueueCluster` create or update access to make the
controller read arbitrary files. Kueue now validates path-based kubeconfigs to stay under
`/etc/multikueue/kubeconfigs`.

ACTION REQUIRED: If you use `locationType=Path`, plan to move kubeconfig files under
`/etc/multikueue/kubeconfigs`, or switch to `locationType=Secret` or `ClusterProfile`.
This prepares your setup for future releases where `MultiKueueKubeConfigPathValidation`
is expected to be enabled by default.
```

mimowo:
/lgtm
/approve
Thank you for the PR, please prepare CP in case the robot fails on conflicts.
/cherrypick release-0.18
/cherrypick release-0.17

kannon92:
> /lgtm /approve Thank you for the PR, please prepare CP in case the robot fails on conflicts. /cherrypick release-0.18 /cherrypick release-0.17

I'm not sure if I really like the idea of cherry-pick an alpha feature gate into release branches.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
kannon92: Ah thanks olekzabl. Missed the negative feature gate in the review.

mimowo:
The ideas is to early notify users by the release note so that they would have more time to adapt their env. 

However, I don't have a strong view here.

kannon92:
The feature is alpha so early notice is in 0.19. I just want to avoid carrying alpha features as that makes it hard to rationalize support of release branches.

PR #12230, opened 2026-06-13
fix transient kueue restarts flake in multikueue BeforeSuite

### LINKED ISSUE #12007: e2e tests: Kueue restarts in SynchronizedBeforeSuite
**Which test is flaking?**:
periodic-kueue-test-e2e-multikueue-extended

https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/periodic-kueue-test-e2e-multikueue-extended-release-0-18/2064103106718732288
**Failure message or logs**:
```

End To End Extended MultiKueue Suite: kueue/kind-node:v1.35.5: [SynchronizedBeforeSuite] expand_less
Run #0: Failed expand_less	1s
{ failed [FAILED] Told to stop trying after 0.037s.
"manager" in "kueue-controller-manager-d895c568d-6kghm" has restarted 1 times
In [SynchronizedBeforeSuite] at: /home/prow/go/src/kubernetes-sigs/kueue/test/e2e/multikueue/extended/suite_test.go:80 @ 06/08/26 22:14:07.067
}
```

TapanManu:
Signed-off-by: TapanManu <tapanmanu2000@gmail.com>

#### What type of PR is this?
/kind bug 
/kind flake

Fixes #12007

```release-note
NONE
```

## Summary by CodeRabbit

* **Tests**
  * Added/updated end-to-end controller-manager configuration for Kueue v1beta2, including cluster-queue metrics, leader-election tuning, per-controller concurrency limits, fair-sharing preemption ordering, and client QPS/burst limits.
  * Updated multi-shard leader-election timing values and refreshed TLS cipher-suite guidance to align with Go 1.26 expectations.
* **Chores**
  * Updated extended Kustomize setup to merge the controller-manager config into the manager ConfigMap.

### DIFF
--- a/pkg/util/cert/cert.go
+++ b/pkg/util/cert/cert.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"fmt"
 	"strings"
-	"time"
 
 	cert "github.com/open-policy-agent/cert-controller/pkg/rotator"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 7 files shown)


pkg/util/cert/cert.go
```diff
@@ -46,57 +47,85 @@ const (
...
-		},
-		HealthProbeBindAddress: cfg.Health.HealthProbeBindAddress,
-		LivenessEndpointName:   cfg.Health.LivenessEndpointName,
-	})
-	if err != nil {
-		return fmt.Errorf("unable to create bootstrap manager: %w", err)
-	}
+	maxAttempts := 5
+	for attempt := 1; attempt <= maxAttempts; attempt++ {
+		log.Info("Attempting certificate bootstrap", "attempt", attempt, "maxAttempts", maxAttempts)
```
mimowo: Wrap the single attempt under a single helper function to avoid increasing indentation for the code

mimowo (commented):
/ok-to-test
I left one comment, but let me ask a more important question - why is this expected to help. What was the root cause for the container restart? Since this PR attempts to modify the production code we need to be sure we have a solid justification for the change. Maybe the same effect (deflake) can be achieved by other means.

/hold
To determine the underlying root cause for the flake and discuss the alternatives.

mimowo:
@TapanManu I propose to close it for now

TapanManu:
> /ok-to-test I left one comment, but let me ask a more important question - why is this expected to help. What was the root cause for the container restart? Since this PR attempts to modify the production code we need to be sure we have a solid justification for the change. Maybe the same effect (deflake) can be achieved by other means.
> 
> /hold To determine the underlying root cause for the flake and discuss the alternatives.

@mimowo one incident which I have noticed was the cert-controller library was unresponsive on the first pass if the API server was too loaded, so i had to attempt a fixed retry approach so that it can be picked up in tests in subsequent passes.

```
The Bottleneck – Because 5 parallel processes and other ML operators are hammering the API server, the background network driver delays the packet transmission. The cert-controller inside Kueue misses its startup window, throws the i/o timeout we saw in your local logs, and crashes.
```
I have noticed the similar issue in my attempt in dev environment

mimowo:
Are you able to repro locally with kuberay moved out?

TapanManu:
> Are you able to repro locally with kuberay moved out?

i will try this and let you know the impact

mimowo:
Please continue with the impact estimation. In any case I'm wondering I would go with the leaderElection parametrization for the "extended" test suites, as proposed here: https://github.com/kubernetes-sigs/kueue/issues/12007#issuecomment-4718135706

What this PR attempts to do is virtually our own layer for the retry-handling which is already present and configurable.

pkg/util/cert/cert.go
```diff
@@ -46,57 +47,85 @@ const (
...
+			time.Sleep(2 * time.Second)
+			continue
+		}
 
-	if err := bootstrapMgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
-		return fmt.Errorf("unable to set up health check for bootstrap manager: %w", err)
-	}
+		if err := bootstrapMgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
+			log.Error(err, "Unable to set up health check for bootstrap manager, retrying...")
+			time.Sleep(2 * time.Second)
```
mszadkow: I don't like those scattered Sleep calls
we could use ExponentialBackoffWithContext() (apimachinery/wait) for whole function that @mimowo mentioned to wrap as helper

mimowo: @mszadkow the PR is going to be re-written so that it only tunes the leaderElection paramters without impacting the business logic. See also discussion in the issue https://github.com/kubernetes-sigs/kueue/issues/12007#issuecomment-4718135706

TapanManu:
@mimowo thanks for the suggestion, i have pushed the changes for changing test suite configurations, could you please review when free and let me know if any changes required ? cc @mszadkow

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for moving this forward. Having the customized tests will be great long term 👍

test/e2e/config/extended/controller_manager_config.yaml
```diff
@@ -0,0 +1,56 @@
+apiVersion: config.kueue.x-k8s.io/v1beta2
```
mimowo, line 1: I'm wondering if we could de-duplicate those configurations by using kustomize more, but we can keep it like that for now, as we already follow this approach.

TapanManu: sure, @mimowo , thanks for your suggestion earlier

mimowo: Cool, thank you for that. Feel free to open an issue to track it. Otherwise I will open later today

TapanManu:
@mimowo @mbobrovskyi Hi both! Since I've been contributing to Kueue and handling these test fixes, the k8s-ci-robot suggested that I consider joining the Kubernetes organization to streamline the CI/testing workflow. 

Would either of you be comfortable sponsoring my membership application once I open the request issue under kubernetes/org? Thanks a lot for your guidance on these PRs!

mimowo:
Hi TapanManu, yes I'm happy to support you joining the k8s and k8s-sigs orgs

mbobrovskyi:
Hi @TapanManu, yes I'm happy to support you joining the k8s and k8s-sigs orgs as well.

TapanManu:
Thanks for your support @mimowo @mbobrovskyi 
I have opened 2 issues for my organisation access request 
* [Kubernetes](https://github.com/kubernetes/org/issues/6458)
* [Kubernetes-sigs](https://github.com/kubernetes/org/issues/6459)

mimowo:
Please do both requests in one issue, you can just comma separate the two orga

test/e2e/config/extended/controller_manager_config.yaml
TapanManu: https://github.com/kubernetes-sigs/kueue/issues/12415 
@mimowo i have opened an issue under this topic

PR #12233, opened 2026-06-14
Fix StatefulSet scale-to-zero workloads holding quota and requeue jitter

### LINKED ISSUE #12232: StatefulSet scale-to-zero can leave admitted Workload reserving quota and block scale-up
**What happened**:

When a Kueue-managed StatefulSet is scaled down to `spec.replicas=0`, the old Workload may remain in the cluster and continue to consume ClusterQueue quota.

In the problematic case we observed:

1. the StatefulSet was scaled down to `0`
2. the corresponding Workload no longer had the StatefulSet ownerReference
3. the Workload still had:
   - `QuotaReserved=True`
   - `Admitted=True`
   - `status.admission != nil`
   - `spec.active=true`
4. the Workload was also marked `Evicted=True` with reason `PodsReadyTimeout`
5. the ClusterQueue still counted the Workload usage

As a result, the workload was logically retired, but its quota reservation was still active and continued blocking queue capacity.

Additionally, scaling the StatefulSet from `0` back to `>0` can be blocked by the webhook with:

```text
workload from previous scale-down is still being deleted
```

So the old Workload can both:

1. keep consuming quota, and
2. prevent the StatefulSet from scaling up again.

When a Kueue-managed StatefulSet is explicitly scaled down to `spec.replicas=0`, the previous admitted Workload should not keep consuming ClusterQueue quota indefinitely.

At minimum, one of the following should happen automatically:

1. the old Workload should release its quota reservation and clear `status.admission`, or
2. the old Workload should be driven to a terminal cleanup state that no longer counts against ClusterQueue usage.

In other words, `replicas=0` should not leave behind an admitted/reserved Workload that remains active in quota accounting.

1. Create a Kueue-managed StatefulSet with `replicas=1` and a queue label so it gets a Workload.
2. Let the Workload become admitted and reserve quota in a ClusterQueue.
3. Make the Workload hit a `PodsReadyTimeout` path, or otherwise keep it in a state where `Evicted=True` is set while the old reservation still exists.
4. Scale the StatefulSet down to `spec.replicas=0`.
5. Observe the old Workload:
   - ownerReference may be removed
   - but `QuotaReserved=True` and `status.admission` may remain
6. Observe the ClusterQueue usage:
   - the old Workload can still be counted as active reserved usage
7. Try scaling the StatefulSet back from `0` to `>0`
   - the webhook may reject it with:
     `workload from previous scale-down is still being deleted`

This looks like a semantic mismatch across multiple parts of the StatefulSet / Workload lifecycle:

1. In `pkg/controller/jobs/statefulset/statefulset_reconciler.go`, when `replicas == 0`, the reconciler removes the StatefulSet ownerReference from the Workload, but does not release quota reservation.
2. In `pkg/controller/core/workload_controller.go`, the `PodsReadyTimeout` eviction path calls `workload.Evict(...)`, but does not clear `status.admission` or unset the quota reservation.
3. In `pkg/workload/workload.go`, `HasActiveQuotaReservation(...)` still treats the Workload as active quota usage as long as:
   - it has quota reservation
   - it is not finished
   - it is active
4. In `pkg/controller/jobs/statefulset/statefulset_webhook.go`, scaling from `0` to `>0` may be rejected if the old Workload still exists:
   - `workload from previous scale-down is still being deleted`

So even though `replicas=0` looks like an explicit owner retirement event, the old Workload can remain both:

1. quota-consuming, and
2. scale-up-blocking

This does not look like a normal delayed cleanup window. It can leave the system in a long-lived residual state that requires manual intervention.

**Environment**:
  - reproducible on our internal Kubernetes environment; exact output can be filled at submission time
  - reproduced on a fork based on `v0.17.3`
  - code inspection also suggests the same behavior still exists in upstream `v0.18.1` and `main`
  - internal GPU cluster
  - Linux
- Kernel (e.g. `uname -a`):
  - Linux kernel
- Install tools:
  - Helm / controller-manager style deployment
- Others:
  - StatefulSet was Kueue-managed
  - the residual Workload remained admitted/reserved after scale-to-zero

gola:
#### What type of PR is this?

/kind bug
/area integrations
/area testing

Fix StatefulSet scale-to-zero behavior so that when `spec.replicas=0`, the old Workload can release its active quota reservation and avoid short-lived re-admission jitter during the terminating-pod window.

Instead of introducing a new condition type, the implementation now uses the existing `WorkloadQuotaReserved=False` condition with reason `OnHold` to represent the temporary scale-to-zero hold, keeping the behavior extensible and aligned with the current workload status model.

Fixes #12232

This PR does not introduce a new Workload condition type. For the StatefulSet scale-to-zero case, the reconciler marks the Workload as `WorkloadQuotaReserved=False` with reason `OnHold`, and the workload controller/helper logic treats on-hold workloads as not eligible for immediate requeue.

The on-hold checks are pushed down into helpers such as `NeedsSecondPass()` and `NeedsDRAReconcile()`, so the behavior remains consistent across different code paths and stays extensible for future integrations that may also need an on-hold style transition.

```release-note
StatefulSet: Fixed a bug where scaling a StatefulSet to zero caused its Workload to be incorrectly requeued for scheduling during the terminating-pod window, competing for quota it should no longer hold.
```

## Summary by CodeRabbit

* **New Features**
  * Added requeue hold mechanism to prevent unnecessary workload requeuing when quota is released during StatefulSet scale-to-zero operations.

* **Bug Fixes**
  * Improved workload re-admission and condition management when scaling StatefulSets back to normal replica counts.

* **Tests**
  * Added comprehensive test coverage for StatefulSet scale-down and scale-up workload transitions, including quota reservation release and re-admission scenarios.

### DIFF
--- a/CHANGELOG/CHANGELOG-0.18.md
+++ b/CHANGELOG/CHANGELOG-0.18.md
@@ -12,7 +12,6 @@ Changes since `v0.18.0`:
 
 ### Bug or Regression
 
-- StatefulSet: Fixed a bug where scaling a StatefulSet to zero caused its Workload to be incorrectly requeued for scheduling during the terminating-pod window, competing for quota it should no longer hold. The core workload controller now uses a generic `WorkloadRequeueHeld` condition instead of StatefulSet-specific logic, making the mechanism extensible for future integrations. (#12233, @gola)
 - DRA: Fixed a bug where pending Workloads using DRA extended resources were not requeued when their `DeviceClass` was deleted or its `extendedResourceName` changed. Kueue now re-evaluates affected Workloads so they do not remain in stale admission state. (#12093, @sohankunkerkar)
 - DRA: Fixed configuration validation to reject `deviceClassMappings[].sources` when the `KueueDRAIntegrationPartitionableDevices` feature gate is disabled, preventing unsupported partitionable-device configuration from being accepted. (#12152, @sohankunkerkar)
 - DRA: Fixed hot reconcile loops for inadmissible Workloads with deterministic DRA resolution
--- a/CHANGELOG/CHANGELOG-0.18.md
+++ b/CHANGELOG/CHANGELOG-0.18.md
@@ -12,6 +12,7 @@ Changes since `v0.18.0`:
 
 ### Bug or Regression
 
+- StatefulSet: Fixed a bug where scaling a StatefulSet to zero caused its Workload to be incorrectly requeued for scheduling during the terminating-pod window, competing for quota it should no longer hold. The core workload controller now uses a generic `WorkloadRequeueHeld` condition instead of StatefulSet-specific logic, making the mechanism extensible for future integrations. (#12233, @gola)
 - DRA: Fixed a bug where pending Workloads using DRA extended resources were not requeued when their `DeviceClass` was deleted or its `extendedResourceName` changed. Kueue now re-evaluates affected Workloads so they do not remain in stale admission state. (#12093, @sohankunkerkar)
 - DRA: Fixed configuration validation to reject `deviceClassMappings[].sources` when the `KueueDRAIntegrationPartitionableDevices` feature gate is disabled, preventing unsupported partitionable-device configuration from being accepted. (#12152, @sohankunkerkar)
 - DRA: Fixed hot reconcile loops for inadmissible Workloads with deterministic DRA resolution
--- a/apis/kueue/v1beta1/workload_types.go
+++ b/apis/kueue/v1beta1/workload_types.go
@@ -665,6 +665,13 @@ const (
 	// WorkloadDeactivationTarget means that the Workload should be deactivated.
 	// This condition is temporary, so it should be removed after deactivation.
 	WorkloadDeactivationTarget = "DeactivationTarget"
+
+	// WorkloadRequeueHeld means that the Workload should not be requeued
+	// after its quota reservation is released. The possible reasons for this
+	// condition are:
+	// - "StatefulSetScaledDown": the owning StatefulSet scaled to zero
+	// Other integrations may use additional reasons in the future.
+	WorkloadRequeueHeld = "RequeueHeld"
 )
 
 // Reasons for the WorkloadPreempted condition.
--- a/apis/kueue/v1beta1/workload_types.go
+++ b/apis/kueue/v1beta1/workload_types.go
@@ -665,13 +665,6 @@ const (
 	// WorkloadDeactivationTarget means that the Workload should be deactivated.
 	// This condition is temporary, so it should be removed after deactivation.
 	WorkloadDeactivationTarget = "DeactivationTarget"
-
-	// WorkloadRequeueHeld means that the Workload should not be requeued
-	// after its quota reservation is released. The possible reasons for this
-	// condition are:
-	// - "StatefulSetScaledDown": the owning StatefulSet scaled to zero
-	// Other integrations may use additional reasons in the future.
-	WorkloadRequeueHeld = "RequeueHeld"
 )
 
 // Reasons for the WorkloadPreempted condition.
@@ -697,6 +690,13 @@ const (
 	// WorkloadInadmissible means that the Workload can't reserve quota
 	// due to LocalQueue or ClusterQueue doesn't exist or inactive.
 	WorkloadInadmissible = "Inadmissible"
+
+	// WorkloadOnHold indicates that the Workload's quota reservation is
+	// intentionally released and the workload should not be requeued.
+	// The possible reasons for this are:
+	// - "StatefulSetScaledDown": the owning StatefulSet scaled to zero
+	// Other integrations may use additional reasons in the future.
+	WorkloadOnHold = "OnHold"
 
 	// WorkloadEvictedByPreemption indicates that the workload was evicted
 	// in order to free resources for a workload with a higher priority.
--- a/apis/kueue/v1beta1/workload_types.go
+++ b/apis/kueue/v1beta1/workload_types.go
@@ -691,12 +691,6 @@ const (
 	// due to LocalQueue or ClusterQueue doesn't exist or inactive.
 	WorkloadInadmissible = "Inadmissible"
 
-	// WorkloadOnHold is the QuotaReserved=False reason used when a Workload's
-	// quota reservation is intentionally released and the workload should not be
-	// requeued. It is currently used by StatefulSet scale-to-zero, and other
-	// integrations may reuse the same reason in the future.
-	WorkloadOnHold = "OnHold"
-
 	// WorkloadEvictedByPreemption indicates that the workload was evicted
 	// in order to free resources for a workload with a higher priority.
 	WorkloadEvictedByPreemption = "Preempted"
--- a/apis/kueue/v1beta1/workload_types.go
+++ b/apis/kueue/v1beta1/workload_types.go
@@ -691,11 +691,10 @@ const (
 	// due to LocalQueue or ClusterQueue doesn't exist or inactive.
 	WorkloadInadmissible = "Inadmissible"
 
-	// WorkloadOnHold indicates that the Workload's quota reservation is
-	// intentionally released and the workload should not be requeued.
-	// The possible reasons for this are:
-	// - "StatefulSetScaledDown": the owning StatefulSet scaled to zero
-	// Other integrations may use additional reasons in the future.
+	// WorkloadOnHold is the QuotaReserved=False reason used when a Workload's
+	// quota reservation is intentionally released and the workload should not be
+	// requeued. It is currently used by StatefulSet scale-to-zero, and other
+	// integrations may reuse the same reason in the future.
 	WorkloadOnHold = "OnHold"
 
 	// WorkloadEvictedByPreemption indicates that the workload was evicted
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -1330,6 +1330,7 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 			backoff = time.Until(e.ObjectNew.Status.RequeueState.RequeueAt.Time)
 		}
 		immediate := backoff <= 0
+		scaledDown := workload.IsStatefulSetScaledDownRelease(wlCopy)
 		log.V(3).Info("Workload transitioned back to pending", "backoff", backoff, "immediate", immediate)
 		// trigger the move of associated inadmissibleWorkloads, if there are any.
 		r.queues.QueueAssociatedInadmissibleWorkloadsAfter(ctx, wlKey, func() {
@@ -1338,6 +1339,13 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 			// the next scheduling cycle.
 			if err := r.cache.DeleteWorkload(log, wlKey); err != nil {
 				log.Error(err, "Failed to delete workload from cache")
+			}
+			// StatefulSet scale-to-zero release has already unreserved quota and
+			// should not be put back into the scheduling queue during the
+			// terminating-pod window. Keep the cache cleanup, but stop here.
+			if scaledDown {
+				r.queues.DeleteSecondPassWithoutLock(wlKey)
+				return
 			}
 			// Here we don't take the lock as it is already taken by the wrapping function.
 			// The delayed requeue is done in the Reconcile function, but we need to
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -1281,6 +1281,8 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 	// We do not handle old workload here as it will be deleted or replaced by new one anyway.
 	workload.AdjustResources(ctrl.LoggerInto(ctx, log), r.client, wlCopy)
 
+	requeueHeld := workload.IsRequeueHeld(wlCopy)
+
 	switch {
 	case status == workload.StatusFinished || !active:
 		if !active {
@@ -1330,7 +1332,6 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 			backoff = time.Until(e.ObjectNew.Status.RequeueState.RequeueAt.Time)
 		}
 		immediate := backoff <= 0
-		scaledDown := workload.IsStatefulSetScaledDownRelease(wlCopy)
 		log.V(3).Info("Workload transitioned back to pending", "backoff", backoff, "immediate", immediate)
 		// trigger the move of associated inadmissibleWorkloads, if there are any.
 		r.queues.QueueAssociatedInadmissibleWorkloadsAfter(ctx, wlKey, func() {
@@ -1340,10 +1341,10 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 			if err := r.cache.DeleteWorkload(log, wlKey); err != nil {
 				log.Error(err, "Failed to delete workload from cache")
 			}
-			// StatefulSet scale-to-zero release has already unreserved quota and
-			// should not be put back into the scheduling queue during the
-			// terminating-pod window. Keep the cache cleanup, but stop here.
-			if scaledDown {
+			// Requeue-held workloads (e.g. StatefulSet scale-to-zero) have already
+			// unreserved quota and should not be put back into the scheduling queue
+			// during the terminating-pod window. Keep the cache cleanup, but stop here.
+			if requeueHeld {
 				r.queues.DeleteSecondPassWithoutLock(wlKey)
 				return
 			}
@@ -1380,7 +1381,9 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 		// and are not supposed to actually change anything.
 		r.cache.AddOrUpdateWorkload(log, wlCopy)
 	}
-	r.queues.QueueSecondPassIfNeeded(ctx, wlCopy, 0)
+	if !requeueHeld {
+		r.queues.QueueSecondPassIfNeeded(ctx, wlCopy, 0)
+	}
 	return true
 }
 
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -1282,7 +1282,6 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 	workload.AdjustResources(ctrl.LoggerInto(ctx, log), r.client, wlCopy)
 
 	onHold := workload.IsOnHold(wlCopy)
-	skipSecondPass := false
 
 	switch {
 	case status == workload.StatusFinished || !active:
@@ -1362,10 +1361,9 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 					r.queues.DeleteSecondPassWithoutLock(wlKey)
 				}
 			}
-			// Clean up second pass for on-hold workloads since NeedsSecondPass
-			// returns false for them and they should not be re-queued.
-			if onHold {
-				skipSecondPass = true
+
+			// Clean up second pass when the updated workload no longer needs it.
+			if !workload.NeedsSecondPass(wlCopy) {
 				r.queues.DeleteSecondPassWithoutLock(wlKey)
 			}
 		})
@@ -1385,9 +1383,7 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 		// and are not supposed to actually change anything.
 		r.cache.AddOrUpdateWorkload(log, wlCopy)
 	}
-	if !skipSecondPass {
-		r.queues.QueueSecondPassIfNeeded(ctx, wlCopy, 0)
-	}
+	r.queues.QueueSecondPassIfNeeded(ctx, wlCopy, 0)
 	return true
 }
 
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -469,7 +469,7 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 			queueOptions = append(queueOptions, workload.WithPreprocessedDRAResources(draResources, replacedExtendedResources))
 		}
 
-		if workload.IsAdmissible(&wl) && !workload.IsRequeueHeld(&wl) {
+		if workload.IsAdmissible(&wl) {
 			if err := r.queues.AddOrUpdateWorkload(log, wl.DeepCopy(), queueOptions...); err != nil {
 				log.V(2).Info("Failed to add DRA workload to queue", "error", err)
 				return ctrl.Result{}, err
@@ -1210,7 +1210,7 @@ func (r *WorkloadReconciler) Create(e event.TypedCreateEvent[*kueue.Workload]) b
 		return true
 	}
 
-	if workload.IsAdmissible(e.Object) && !workload.IsRequeueHeld(wlCopy) {
+	if workload.IsAdmissible(e.Object) {
 		if err := r.queues.AddOrUpdateWorkload(log, wlCopy); err != nil {
 			log.V(2).Info("ignored an error for now", "error", err)
 		}
@@ -1281,7 +1281,7 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 	// We do not handle old workload here as it will be deleted or replaced by new one anyway.
 	workload.AdjustResources(ctrl.LoggerInto(ctx, log), r.client, wlCopy)
 
-	requeueHeld := workload.IsRequeueHeld(wlCopy)
+	onHold := workload.IsOnHold(wlCopy)
 
 	switch {
 	case status == workload.StatusFinished || !active:
@@ -1308,8 +1308,8 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 		})
 	case prevStatus == workload.StatusPending && status == workload.StatusPending:
 		switch {
-		case requeueHeld:
-			log.V(2).Info("Skipping queue update for requeue-held workload")
+		case onHold:
+			log.V(2).Info("Skipping queue update for workload on hold")
 		case dra.NeedsDRAReconcile(e.ObjectNew, r.draBackedResources):
 			log.V(2).Info("Skipping queue update for DRA workload - handled in Reconcile")
 		default:
@@ -1343,10 +1343,10 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 			if err := r.cache.DeleteWorkload(log, wlKey); err != nil {
 				log.Error(err, "Failed to delete workload from cache")
 			}
-			// Requeue-held workloads (e.g. StatefulSet scale-to-zero) have already
+			// Workloads on hold (e.g. StatefulSet scale-to-zero) have already
 			// unreserved quota and should not be put back into the scheduling queue
 			// during the terminating-pod window. Keep the cache cleanup, but stop here.
-			if requeueHeld {
+			if onHold {
 				r.queues.DeleteSecondPassWithoutLock(wlKey)
 				return
 			}
@@ -1383,7 +1383,7 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 		// and are not supposed to actually change anything.
 		r.cache.AddOrUpdateWorkload(log, wlCopy)
 	}
-	if !requeueHeld {
+	if !onHold {
 		r.queues.QueueSecondPassIfNeeded(ctx, wlCopy, 0)
 	}
 	return true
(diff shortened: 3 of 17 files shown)


kannon92:
The repro for this seems pretty simple. Could we add some integration or e2e on this to not regress?

I'd like a e2e for this to be honest.

Do you know if we see a similar problem for LWS?

pkg/controller/core/workload_controller.go
```diff
@@ -1330,6 +1330,7 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 			backoff = time.Until(e.ObjectNew.Status.RequeueState.RequeueAt.Time)
 		}
 		immediate := backoff <= 0
+		scaledDown := workload.IsStatefulSetScaledDownRelease(wlCopy)
```
kannon92: We have a job library so we don't have to change anything in workload controller.

I think we should look into ways to avoid having to add "hacks" for statefulset behavior in the workload controller.

Ideally this logic can go to jobs/statefulset.

kannon92:
Please follow the template. We will need a release note for this bug.

pkg/controller/core/workload_controller.go
gola: Thanks for the review @kannon92! You make a great point — we shouldn't add StatefulSet-specific logic in workload_controller.go. I agree this should live in the job framework layer.

I've been thinking about how to leverage the job framework for this. Here's my assessment of possible approaches, with my preferred one first:

Approach B (Preferred): Job framework callback/hook

The core idea is to let the job integration influence workload requeue behavior without the workload controller knowing about specific job types. There are a few concrete variants:

B1: Add ShouldSkipRequeue callback to IntegrationCallbacks — The workload controller queries the integration manager for the workload's owning integration, and the StatefulSet reconciler implements the callback to return true when the workload has a scale-to-zero release condition. This requires wiring the integration manager into the workload reconciler.

B2: Generic Workload condition (e.g., WorkloadRequeueHeld) — The StatefulSet reconciler sets a generic condition on the Workload when doing scale-to-zero release. The workload controller checks this condition generically (not tied to any job type). Any integration can set this condition to prevent requeue. This is a lighter variant of B that doesn't require the workload controller to reference the integration manager.

B3: Extend JobReconcilerInterface — Add a method like ShouldSkipWorkloadRequeue(wl *kueue.Workload) bool. The workload controller uses the workload's owner reference to find the corresponding reconciler and queries it. This requires all JobReconcilerInterface implementations to add the method (defaulting to false).

I'm leaning toward B2 as a good balance — it's architecturally clean (the job reconciler is responsible for setting the condition, the workload controller checks it generically), it doesn't require major structural changes to the controller wiring, and it's extensible for other integrations (e.g., LWS could set the same condition).

Approach A: Generic annotation — Less preferred because annotations are less structured than conditions and harder to evolve.

Approach C: Mark Workload as inactive — Not recommended because it has unintended side effects on eviction and UI.

What do you think? Is B2 a reasonable direction, or would you prefer one of the other B variants?

Regarding your other points:

e2e/integration test: I'll add one. I see the existing integration test suite at test/integration/singlecluster/controller/jobs/statefulset/ — I'll add a scale-to-zero quota release test there.
LWS: I'll investigate and report back.
Release note: Adding now.

gola:
> The repro for this seems pretty simple. Could we add some integration or e2e on this to not regress?
> 
> I'd like a e2e for this to be honest.
> 
> Do you know if we see a similar problem for LWS?

**On whether LWS has the same scale-to-zero requeue issue:**


I checked the LWS integration and it does **not** have the same problem. The architectures are fundamentally different:


- **StatefulSet**: When `spec.replicas=0`, the StatefulSet reconciler **keeps the Workload** and only releases the quota reservation (setting `QuotaReserved=False`). This creates an intermediate state where the Workload transitions from Admitted → Pending, which triggers the workload_controller's Update handler to requeue it — that's the bug this PR fixes.


- **LWS**: When `spec.replicas=0`, the LWS reconciler's [`filterWorkloads()`](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/jobs/leaderworkerset/leaderworkerset_reconciler.go#L252) places **all existing Workloads into the `toDelete` list** and deletes them directly. There is no "Workload preserved with quota released" intermediate state. The workload_controller's `Delete()` handler correctly cleans up cache and queues.


So LWS never encounters the `QuotaReserved=False` + `Pending` combination that causes the spurious requeue, and therefore does not need the `WorkloadRequeueHeld` condition.

pkg/controller/core/workload_controller.go
gola: @kannon92 Thanks for the follow-up! Here's a summary of all the changes since my last update:

Remove StatefulSet-specific logic from workload_controller.go — Done in previous commit (generic WorkloadRequeueHeld condition).

Add e2e/integration tests — Added both:

Integration test in test/integration/singlecluster/controller/jobs/statefulset/statefulset_controller_test.go: verifies the full RequeueHeld lifecycle (scale-to-zero sets the condition, scale-up clears it and re-admits).
e2e test in test/e2e/singlecluster/baseline/statefulset_test.go: end-to-end test that creates a StatefulSet, scales to zero, verifies QuotaReserved=False + RequeueHeld=True, scales back up, and verifies re-admission with RequeueHeld cleared.
Add release note — Added to CHANGELOG/CHANGELOG-0.18.md. I've also updated the PR description with the standard release-note block.

Idempotency fix — Addressed a review comment from CodeRabbit: if client.Update succeeds but clearRequeueHeld fails, a subsequent reconcile could skip clearing the condition. Now the !shouldUpdate path also retries clearRequeueHeld when replicas > 0 and the condition is still present.

Changes pushed — PTAL when you get a chance.

kannon92: cc @mimowo @tenzen-y 

WDYT?

CHANGELOG/CHANGELOG-0.18.md
```diff
@@ -12,6 +12,7 @@ Changes since `v0.18.0`:
 
 ### Bug or Regression
 
+- StatefulSet: Fixed a bug where scaling a StatefulSet to zero caused its Workload to be incorrectly requeued for scheduling during the terminating-pod window, competing for quota it should no longer hold. The core workload controller now uses a generic `WorkloadRequeueHeld` condition instead of StatefulSet-specific logic, making the mechanism extensible for future integrations. (#12233, @gola)
```
kannon92: Is this a regression for 0.18 or a problem that has been in code for a while?

If its a problem we will probably carry to 0.17 also. 

I don't usually see people modifying the change log in main so I would revert this for now.

gola: This is not a regression in v0.18 — it's a bug that has existed since the StatefulSet integration was first introduced. When a StatefulSet scales to zero, its Workload releases the quota reservation (QuotaReserved=False) but remains as an object. The workload controller's Update handler then sees this transition from Admitted→Pending and requeues the Workload for scheduling, causing it to briefly re-acquire quota before the terminating-pod window closes. This has been the behavior all along; it just wasn't reported until #12232.

I've reverted the CHANGELOG modification as requested. The release note remains in the PR description's release-note block.

mimowo:
/assign 
I will try to review EOW, but I would like to have review from @mbobrovskyi first who worked on the STS integration a lot before.

kannon92:
looks like this PR has conflicts.

sohankunkerkar:
Thanks for raising this, the root cause is real. One concern on the approach: introducing `WorkloadRequeueHeld` adds API surface that's hard to remove, worth considering whether `WorkloadEvicted` with a new reason could do the same within the existing lifecycle. Also the requeue suppression may need to cover all queue-entry paths in the workload controller, not just the admission transition, to fully prevent the held workload from being re-enqueued

gola:
> Thanks for raising this, the root cause is real. One concern on the approach: introducing `WorkloadRequeueHeld` adds API surface that's hard to remove, worth considering whether `WorkloadEvicted` with a new reason could do the same within the existing lifecycle. Also the requeue suppression may need to cover all queue-entry paths in the workload controller, not just the admission transition, to fully prevent the held workload from being re-enqueued

@sohankunkerkar Thanks for the thoughtful review! Let me address both concerns:
1. WorkloadRequeueHeld vs WorkloadEvicted with a new reason
I considered using WorkloadEvicted with a new reason, but it has semantic and behavioral issues:
• WorkloadEvicted triggers eviction-specific logic in the workload controller Reconcile path (e.g., eviction latency metrics, pods-ready timeout handling, admission check eviction). A scale-down-to-zero is not an eviction — the workload should remain active and pending, not enter the evicted lifecycle.
• Using WorkloadEvicted would incorrectly report eviction latency metrics and could interfere with other eviction handlers that check the eviction condition.
• The semantic of RequeueHeld is fundamentally different: the workload is not evicted, it is temporarily held from requeueing while remaining a valid pending workload.
2. Requeue suppression covering all queue-entry paths
You are absolutely right — this was a gap in the initial implementation. I have now added !workload.IsRequeueHeld() checks in all queue-entry paths in the workload controller (Create handler, addLocalQueueLocked, resourceUpdatesHandler, deviceClassHandler, Reconcile DRA path, and Update Pending→Pending path). This is covered by the new test case TestUpdateStatefulSetScaleDown/requeue-held_workload_with_pending-to-pending_update_should_not_be_requeued.

gola:
@mimowo  Gentle ping — all CI checks are now passing and I've addressed the review feedback. PTAL when you get a chance.

mimowo:
Cool putting into my ToDo towards EOD. It would be great if @mbobrovskyi can give it a pass in the meanwhile

pkg/controller/core/workload_controller.go
```diff
@@ -469,7 +469,7 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 			queueOptions = append(queueOptions, workload.WithPreprocessedDRAResources(draResources, replacedExtendedResources))
 		}
 
-		if workload.IsAdmissible(&wl) {
+		if workload.IsAdmissible(&wl) && !workload.IsRequeueHeld(&wl) {
```
mimowo: Why not wind this check `&& !workload.IsRequeueHeld(&wl)` under `workload.IsAdmissible(&wl)`? It makes the workload inadmissible, so it seems to match the definition.

apis/kueue/v1beta1/workload_types.go
```diff
@@ -665,6 +665,13 @@ const (
 	// WorkloadDeactivationTarget means that the Workload should be deactivated.
 	// This condition is temporary, so it should be removed after deactivation.
 	WorkloadDeactivationTarget = "DeactivationTarget"
+
+	// WorkloadRequeueHeld means that the Workload should not be requeued
+	// after its quota reservation is released. The possible reasons for this
+	// condition are:
+	// - "StatefulSetScaledDown": the owning StatefulSet scaled to zero
+	// Other integrations may use additional reasons in the future.
+	WorkloadRequeueHeld = "RequeueHeld"
```
mimowo: Maybe rename to OnHold?

I think this would read better. Seems like "Requeue" is more of a technical detail from user perspective. We already also have a concept of putting CQ / LQ On Hold. So that would feel consistent to me.

apis/kueue/v1beta1/workload_types.go
```diff
@@ -665,6 +665,13 @@ const (
 	// WorkloadDeactivationTarget means that the Workload should be deactivated.
 	// This condition is temporary, so it should be removed after deactivation.
 	WorkloadDeactivationTarget = "DeactivationTarget"
+
+	// WorkloadRequeueHeld means that the Workload should not be requeued
```
mimowo: Any reason for using a new condition rather than just "OnHold" for the QuotaReserved=False? This would feel more consistent with the new approach recently developed by @j-skiba as part of https://github.com/kubernetes-sigs/kueue/issues/10852

pkg/controller/core/workload_controller.go
```diff
@@ -1372,7 +1383,9 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 		// and are not supposed to actually change anything.
 		r.cache.AddOrUpdateWorkload(log, wlCopy)
 	}
-	r.queues.QueueSecondPassIfNeeded(ctx, wlCopy, 0)
+	if !requeueHeld {
+		r.queues.QueueSecondPassIfNeeded(ctx, wlCopy, 0)
+	}
```
mimowo: Hm, I think this is something we could check inside `QueueSecondPassIfNeeded`. Also, I would rather make the check more generic using IsAdmissible, wdyt?

pkg/controller/core/workload_controller.go
```diff
@@ -1339,6 +1343,13 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 			if err := r.cache.DeleteWorkload(log, wlKey); err != nil {
 				log.Error(err, "Failed to delete workload from cache")
 			}
+			// Requeue-held workloads (e.g. StatefulSet scale-to-zero) have already
+			// unreserved quota and should not be put back into the scheduling queue
+			// during the terminating-pod window. Keep the cache cleanup, but stop here.
+			if requeueHeld {
+				r.queues.DeleteSecondPassWithoutLock(wlKey)
+				return
+			}
```
mimowo: Here, I'm also wondering if we could generalize the check using IsAdmissible, or maybe we are missing a helper like `workload.NeedsSecondPass`, if not, then delete. Then, workload.NeedsSecondPass could check the necessary thing.

mimowo (commented):
I think the approach is reasonable, my comments are mostly:
1. do we need an entire new condition, by using reason for QuotaReserved=False we are more consistent with the new granular approach to reasons. It would be tier "Suspended" from the Inadmissible workloads KEP
2. Try to generalize the checks to avoid scattering the custom checks. For example, I'm pretty sure the check could be winded under IsAdmissible, or SecondCheck (...) helpers.

pkg/controller/core/workload_controller.go
mimowo: I think my perspective is captured by the comment https://github.com/kubernetes-sigs/kueue/pull/12233#pullrequestreview-4547406388, but let me know if that does not cover something discussed so far.

pkg/controller/jobs/statefulset/statefulset_reconciler.go
```diff
@@ -214,11 +214,21 @@ func (r *Reconciler) reconcileWorkload(ctx context.Context, sts *appsv1.Stateful
 	}
 
 	var shouldUpdate bool
+	// Initialize retry-idempotent flags early so that a partial failure in a
+	// previous reconcile (e.g. owner-ref update succeeded but status patch
+	// failed) does not leave the workload stuck.
+	shouldReleaseReservation := replicas == 0 && workload.HasActiveQuotaReservation(wl) && !workload.IsFinished(wl)
```
mimowo: That should also check if not deactivated

gola: Thanks @mimowo for the review! Good catch on the deactivated check.

I've addressed the latest comment:

shouldReleaseReservation not checking deactivated workloads: Fixed in the latest push. Added workload.IsActive(wl) to the condition so that a deactivated workload (e.g., one with spec.active=false) will not trigger the OnHold quota-release path:

shouldReleaseReservation := replicas == 0 && workload.HasActiveQuotaReservation(wl) && !workload.IsFinished(wl) && workload.IsActive(wl)

This ensures that if a user deactivates a StatefulSet's workload while it's scaled to zero, we don't interfere with the deactivation lifecycle by setting the OnHold condition.

To summarize, all your review comments have now been addressed:

✅ Use QuotaReserved=False with reason OnHold instead of a new condition type — consistent with the granular reasons approach from #10852
✅ Merge IsOnHold into IsAdmissible() — no scattered custom checks; on-hold workloads are automatically inadmissible
✅ Rename RequeueHeld → OnHold — consistent with CQ/LQ "on hold" terminology
✅ Check IsActive in shouldReleaseReservation — skip deactivated workloads
PTAL when you get a chance.

gola:
Thanks @mimowo for the thorough review! All your feedback has been addressed in the latest commits. Here's a summary of the changes:

Use QuotaReserved=False with reason OnHold instead of a new condition type — Removed WorkloadRequeueHeld condition. Now using QuotaReserved=False with reason WorkloadOnHold, consistent with the granular reasons approach from #10852.

Merge IsOnHold() into IsAdmissible() — On-hold workloads are automatically inadmissible. No scattered IsOnHold checks at queue entry points; IsAdmissible() includes the !IsOnHold(w) check internally.

Rename RequeueHeld → OnHold — Consistent with CQ/LQ "on hold" terminology. All constants, functions, and references updated.

Check IsActive in shouldReleaseReservation — Added workload.IsActive(wl) to prevent setting the OnHold condition on deactivated workloads.

Generalize second-pass and requeue checks — Both the QueueSecondPassIfNeeded and DeleteSecondPassWithoutLock paths now use IsAdmissible() instead of custom OnHold checks.

PTAL when you get a chance.

pkg/controller/core/workload_controller.go
```diff
@@ -1372,7 +1383,9 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 		// and are not supposed to actually change anything.
 		r.cache.AddOrUpdateWorkload(log, wlCopy)
 	}
-	r.queues.QueueSecondPassIfNeeded(ctx, wlCopy, 0)
+	if !onHold {
+		r.queues.QueueSecondPassIfNeeded(ctx, wlCopy, 0)
```
mimowo: I think here we could check the new condition inside the function, wdyt? that would also make the code more resilient if the function is called from another place

pkg/controller/core/workload_controller.go
```diff
@@ -1306,6 +1308,8 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 		})
 	case prevStatus == workload.StatusPending && status == workload.StatusPending:
 		switch {
+		case onHold:
```
mimowo, line 1311: Here I'm also wondering if we could move the check inside the NeedsDRAReconcile, so that we skip from all code paths

pkg/controller/core/workload_controller.go
```diff
@@ -1339,6 +1343,13 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 			if err := r.cache.DeleteWorkload(log, wlKey); err != nil {
 				log.Error(err, "Failed to delete workload from cache")
 			}
+			// Workloads on hold (e.g. StatefulSet scale-to-zero) have already
+			// unreserved quota and should not be put back into the scheduling queue
+			// during the terminating-pod window. Keep the cache cleanup, but stop here.
+			if onHold {
```
mimowo: Maybe here we could have a new helper with the IfNeeded suffic which would check the new condition. Iiuc the only is is then to stop add oOrUpdate, but I think there too we could check inside the function .

mimowo (commented):
Basically it looks good to me, but I'm thinking if we really need any changes to the workload controller. I would like to consider the alternative where the checks are pushed down to the helpers so that the code is resilient no matter from which code path the helpers are called.

pkg/controller/core/workload_controller.go
```diff
@@ -1346,6 +1350,8 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 			// reconciler, we would execute it on every run, which might mess up the state.
 			if immediate {
 				switch {
+				case workload.IsOnHold(wlCopy):
```
mbobrovskyi:
```suggestion
				case onHold:
```

pkg/controller/core/workload_controller.go
```diff
@@ -1355,6 +1361,11 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 					r.queues.DeleteSecondPassWithoutLock(wlKey)
 				}
 			}
+			// Clean up second pass for on-hold workloads since NeedsSecondPass
+			// returns false for them and they should not be re-queued.
+			if workload.IsOnHold(wlCopy) {
```
mbobrovskyi:
```suggestion
			if onHold {
```

pkg/controller/jobs/statefulset/statefulset_reconciler.go
```diff
@@ -244,10 +253,63 @@ func (r *Reconciler) reconcileWorkload(ctx context.Context, sts *appsv1.Stateful
...
+		return err
+	}
+
+	if shouldReleaseReservation {
+		return r.releaseScaleDownReservation(ctx, wl)
+	}
+
+	if shouldClearOnHold {
+		return r.clearOnHold(ctx, wl)
+	}
```
mbobrovskyi, line 267: It looks like we duplicate this logic.

pkg/controller/jobs/statefulset/statefulset_reconciler.go
```diff
@@ -244,10 +253,63 @@ func (r *Reconciler) reconcileWorkload(ctx context.Context, sts *appsv1.Stateful
...
+		}
+		if shouldClearOnHold {
+			return r.clearOnHold(ctx, wl)
+		}
+		return nil
+	}
+
+	if err := r.client.Update(ctx, wl); err != nil {
+		return err
+	}
```
mbobrovskyi:
```suggestion
	if !shouldUpdate {
		return r.client.Update(ctx, wl)
	}
```

gola:
> Basically it looks good to me, but I'm thinking if we really need any changes to the workload controller. I would like to consider the alternative where the checks are pushed down to the helpers so that the code is resilient no matter from which code path the helpers are called.

Thanks for the suggestions — I addressed this direction in the latest updates.

What changed:
- I renamed the user-facing concept from `RequeueHeld` to `OnHold`.
- I reused `QuotaReserved=False` with reason `OnHold` instead of introducing a new condition type.
- I moved the generic on-hold handling into workload helpers:
  - `workload.IsAdmissible()`
  - `workload.NeedsSecondPass()`
  - `dra.NeedsDRAReconcile()`

So on-hold workloads are now treated generically as inadmissible, and the helper-level checks make the behavior more resilient across call sites.

I still keep a small amount of explicit handling in `WorkloadReconciler.Update()`, because the admitted/quotaReserved -> pending transition must first clear the scheduler cache entry before we skip the immediate requeue path. That part is event-path-specific, but the on-hold semantics themselves are now generic.

I also added/updated tests to cover the OnHold lifecycle and the Update-path behavior. PTAL.

pkg/controller/jobs/statefulset/statefulset_reconciler.go
gola: Good point, thanks.

I refactored this to avoid duplicating the release/clear flow:
- compute `shouldReleaseReservation` / `shouldClearOnHold` up front
- do a single `if shouldUpdate { ... }` update block
- then run release / clear unconditionally afterward when needed

That keeps the reconcile path simpler and also preserves retry-idempotency if a partial update succeeds but the status patch fails.

pkg/controller/core/workload_controller.go
gola: @kannon92  Thanks again for pushing on the architecture here.

I aligned the implementation with that direction in the latest revision:
- no new StatefulSet-specific condition type was kept in the workload controller
- the user-facing state is now expressed as `QuotaReserved=False` with reason `OnHold`
- the generic on-hold behavior was moved into shared workload helpers (`IsAdmissible`, `NeedsSecondPass`, `NeedsDRAReconcile`)
- StatefulSet remains responsible for setting/clearing the OnHold state during scale-to-zero / scale-up

So the workload controller now handles the generic semantics, while the StatefulSet integration owns when that state should be applied.

I also added integration/e2e coverage and fixed the retry-idempotency issue around clearing/releasing state. PTAL when you have a chance.

pkg/controller/core/workload_controller.go
mimowo: Why do you still maintain this check `if !skipSecondPass {` since the check IsOnHold is already moved down to QueueSecondPassIfNeeded? Seems like an unnecessary check on both, then we also wouldn't need the  `skipSecondPass` variable.

apis/kueue/v1beta1/workload_types.go
```diff
@@ -691,6 +691,12 @@ const (
 	// due to LocalQueue or ClusterQueue doesn't exist or inactive.
 	WorkloadInadmissible = "Inadmissible"
 
+	// WorkloadOnHold is the QuotaReserved=False reason used when a Workload's
+	// quota reservation is intentionally released and the workload should not be
+	// requeued. It is currently used by StatefulSet scale-to-zero, and other
+	// integrations may reuse the same reason in the future.
+	WorkloadOnHold = "OnHold"
```
mimowo: Is this really in use? We only maintain v1beta2 at this point

pkg/controller/core/workload_controller.go
```diff
@@ -1355,6 +1362,12 @@ func (r *WorkloadReconciler) Update(e event.TypedUpdateEvent[*kueue.Workload]) b
 					r.queues.DeleteSecondPassWithoutLock(wlKey)
 				}
 			}
+			// Clean up second pass for on-hold workloads since NeedsSecondPass
+			// returns false for them and they should not be re-queued.
+			if onHold {
+				skipSecondPass = true
```
mimowo: Instead of OnHold I think we should rather call NeedsSecondPass, so:

if !workload.NeedsSecondPass(w) {
  r.queues.DeleteSecondPassWithoutLock(wlKey)
}

mimowo:
I think the PR is overall in a good shape, but some of the places puzzle me, so I asked questions

gola:
> I think the PR is overall in a good shape, but some of the places puzzle me, so I asked questions

@mimowo Thanks for the careful review and for calling out the places that felt a bit odd.


I addressed the latest feedback by pushing the on-hold handling further down into the helpers, so the behavior is now consistent across code paths:
- `NeedsSecondPass()` now returns false for on-hold workloads, and the controller calls `QueueSecondPassIfNeeded()` unconditionally.
- `NeedsDRAReconcile()` now also skips on-hold workloads directly.
- I also simplified the second-pass cleanup to use `NeedsSecondPass()` and removed the redundant outer guard, and dropped the unnecessary v1beta1-only change.


So the current behavior should be:
- we still clean the workload out of cache when it transitions back to Pending,
- we do not immediately requeue it during the terminating-pod window,
- and we do not keep it in second-pass unless it actually needs it again.


Please let me know if you'd still prefer a different factoring, but I hope this version is closer to the helper-centric approach you had in mind.

apis/kueue/v1beta1/workload_types.go
gola: Done — removed OnHold code from v1beta1 workload_types.go, only v1beta2 has it now.

pkg/controller/core/workload_controller.go
gola: Done in c189c0f — second-pass cleanup now uses NeedsSecondPass() as suggested.

pkg/controller/core/workload_controller.go
gola: Done in c189c0f — removed skipSecondPass variable entirely, QueueSecondPassIfNeeded is now called unconditionally, and second-pass cleanup uses NeedsSecondPass() directly.

pkg/controller/core/workload_controller.go
gola: Done — IsOnHold check is now inside NeedsDRAReconcile() (pkg/dra/extended_resources.go:41), so all code paths calling it automatically skip on-hold workloads.

pkg/controller/core/workload_controller.go
gola: Done — IsOnHold is now checked inside IsAdmissible(), so on-hold workloads are automatically inadmissible.

apis/kueue/v1beta1/workload_types.go
gola: Done — renamed to OnHold, consistent with CQ/LQ On Hold terminology.

apis/kueue/v1beta1/workload_types.go
gola: Done — now using QuotaReserved=False with reason "OnHold" instead of a new condition type, consistent with the granular reasons approach from #10852.

pkg/controller/core/workload_controller.go
gola: Done — on-hold check is now inside NeedsSecondPass() (returns false) and IsAdmissible() (returns false), so the behavior is consistent from all call sites.

pkg/controller/core/workload_controller.go
gola: Done — NeedsSecondPass() helper now exists and includes the IsOnHold check internally, returning false for on-hold workloads.

pkg/controller/core/workload_controller.go
gola: Done — on-hold workloads are caught by the onHold switch case before reaching AddOrUpdate, and IsAdmissible() also returns false for them.

pkg/controller/core/workload_controller.go
gola: Done — applied the suggestion, using case onHold: now.

pkg/controller/core/workload_controller.go
gola: Done — used the switch case form as suggested in your other comment.

pkg/controller/jobs/statefulset/statefulset_reconciler.go
gola: Thanks — I restructured the update logic to use a single shouldUpdate block, computing shouldReleaseReservation/shouldClearOnHold upfront, then doing one Update + conditional release/clear. This achieves the same deduplication goal while preserving the retry-idempotent flow.

gola:
@mimowo @kannon92 @mbobrovskyi  Gentle ping for a final pass when you have a moment.

The latest updates are on commit `c189c0f63`, and from the current PR status I’m not seeing any new actionable CI failures. At this point, it looks like the remaining blocker is approval/LGTM.

PTAL, and thanks again for the review.

mimowo (commented):
/lgtm
/approve
Thank you for the STS bugfix, I think the mechanism is pretty generic - allowing Kueue to put OnHold any workload without necessarily deactivating. 
/cherrypick release-0.18
/cherrypick release-0.17
Please prepare the cherrypicks in case the robot fails on conflicts. You may want to use the ./hack/cherry_pick_pull.sh script

PR #12256, opened 2026-06-15
WIP: Implement the TAS node domain fit check using the WAS `scheduler-library`

kshalot:
#### What type of PR is this?
/kind feature
/area tas

#### What this PR does / why we need it:
To get our foot in the door with the [`scheduler-library`](https://github.com/kubernetes-sigs/scheduler-library) integration.

#### Which issue(s) this PR fixes:
Fixes #
Part of #8871

#### Special notes for your reviewer:
This is not mergeable until all the TAS tests are passing (feature parity with the existing implementation) and until https://github.com/kubernetes-sigs/scheduler-library/pull/2 is merged.

Additionally, I want to benchmark this first (I'll update the PR later today) using our scheduler performance tests. I benchmarked the proof-of-concept I implemented earlier, the results suggested that the CPU usage does not meaningfully differ when this flag is disabled/enabled, but the memory balloons when the topology is large. This branch contains an improved implementation of the fit check, which ideally should be more performant. TBD.

#### Does this PR introduce a user-facing change?
```release-note
WorkloadAwareScheduler: Add the `SchedulerLibraryIntegration` feature gate. Perform the node domain TAS fit check using the `scheduler-library` (https://github.com/kubernetes-sigs/scheduler-library).
```

PR #12258, opened 2026-06-15
Drop the deprecated `DynamicResourceAllocation` feature gate

kshalot:
#### What type of PR is this?
/kind cleanup
/kind deprecation

Fixes #

Required for https://github.com/kubernetes-sigs/kueue/pull/12256.

```release-note
DRA: Remove the deprecated `DynamicResourceAllocation` feature gate. Use `KueueDRAIntegration` instead.
```

## Summary by CodeRabbit

* **Chores**
  * Removed the `DynamicResourceAllocation` feature gate from the versioned feature lists.
  * Updated `KueueDRAIntegration` to use the 0.18 Beta default behavior (enabled by default).

### DIFF
--- a/pkg/features/kube_features.go
+++ b/pkg/features/kube_features.go
@@ -197,7 +197,7 @@ const (
 	// owner: @kannon92
 	// kep: https://github.com/kubernetes-sigs/kueue/tree/main/keps/2941-DRA
 	//
-	// Reject workloads that use DRA resources when the DynamicResourceAllocation feature gate is disabled.
+	// Reject workloads that use DRA resources when the KueueDRAIntegration feature gate is disabled.
 	KueueDRARejectWorkloadsWhenDRADisabled featuregate.Feature = "KueueDRARejectWorkloadsWhenDRADisabled"
 
 	// owner: @PannagaRao
--- a/pkg/features/kube_features.go
+++ b/pkg/features/kube_features.go
@@ -178,9 +178,6 @@ const (
 	// Enable quota accounting for Dynamic Resource Allocation (DRA) devices in workloads.
 	KueueDRAIntegration featuregate.Feature = "KueueDRAIntegration"
 
-	// Deprecated: planned to be removed in 0.19. Use KueueDRAIntegration instead.
-	DynamicResourceAllocation featuregate.Feature = "DynamicResourceAllocation"
-
 	// owner: @sohankunkerkar
 	// kep: https://github.com/kubernetes-sigs/kueue/tree/main/keps/2941-DRA
 	//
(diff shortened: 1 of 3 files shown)


mimowo:
/release-note-edit
```release-note
DRA: Remove the deprecated `DynamicResourceAllocation` feature gate. Use `KueueDRAIntegration` instead.
```
cc @sohankunkerkar @kannon92 , fyi as stated in PR description this is now needed for https://github.com/kubernetes-sigs/kueue/pull/12256
/lgtm
/approve
Thank you 👍

sohankunkerkar (approved):
/lgtm

sohankunkerkar:
The versioned spec is removed but the constant and ~30 usages across pkg/, cmd/, and tests still reference DynamicResourceAllocation.  Is there a follow-up to migrate those to KueueDRAIntegration?

mimowo:
Oh good point, we need to drop all usages completely to avoid future confusion. We don't need any smarter migration than just rename, because k8s deprecation policy allows to drop any Alpha feature without notice.

kshalot:
@sohankunkerkar @mimowo I must have missed something, because I saw that https://github.com/kubernetes-sigs/kueue/pull/11255 changes the usage to reference `KueueDRAIntegration` instead. And the only references to `DynamicResourceAllocation` that I can find are in the context of the k8s config, where the fg name should stay the same.

Let me take another look.

mimowo:
Hm, yeah most of them are not relevant, but this one could be cleaner: https://github.com/kshalot/kueue/blob/04708953fdd83b08dc18750586572ebc79623633/pkg/features/kube_features.go#L181-L182, along with all uses

sohankunkerkar:
> @sohankunkerkar @mimowo I must have missed something, because I saw that #11255 changes the usage to reference `KueueDRAIntegration` instead. And the only references to `DynamicResourceAllocation` that I can find are in the context of the k8s config, where the fg name should stay the same.
> 
> Let me take another look.

Sorry for the false alarm. One more [change](https://github.com/kubernetes-sigs/kueue/pull/12258#issuecomment-4710510273) requested by @mimowo  and this is good to go

kshalot:
> Hm, yeah most of them are not relevant, but this one could be cleaner: https://github.com/kshalot/kueue/blob/04708953fdd83b08dc18750586572ebc79623633/pkg/features/kube_features.go#L181-L182, along with all uses

Yeah I don't know how I missed this one, maybe I did not stage that chunk into git 😩

I also found a reference to the old name in a comment:
https://github.com/kubernetes-sigs/kueue/pull/12258/changes/c50070cf9ca0147b139fe41d40e6f671ff60c302

I aligned that as well.

mimowo:
/lgtm 
/approve
/unhold
Thabks for the preparatory PR to offload the main integration PR which probably needs to wait a bit more anyway.

PR #12274, opened 2026-06-15
Generalize TAS PodSet groups beyond leader-worker (required, alpha)

### LINKED ISSUE #11953: TAS: Allow PodSet groups with more than 2 PodSets and without the single-replica leader constraint (required topology)

Relax the two constraints that TAS PodSet grouping (`kueue.x-k8s.io/podset-group-name`) currently enforces in `ValidatePodSetGroupingTopology` ([`pkg/controller/jobframework/tas_validation.go:214-264`](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/jobframework/tas_validation.go#L214)):

1. A group must contain **exactly 2 PodSets** — otherwise: `"can only define groups of exactly 2 pod sets, got: N pod set(s)"`.
2. At least one of the 2 PodSets must have **count == 1** — otherwise: `"can only define groups where at least one pod set has only 1 replica"`.

We would like to relax **both** so that a PodSet group with a **required** topology constraint can:
- contain **more than 2 PodSets**, and
- contain PodSets that **all have count > 1** (no required single-replica "leader").

Placement semantics: all PodSets sharing a `podset-group-name` with a required topology should be co-located within the same topology domain (e.g., the same spine).

Scope this to the **required** topology case (`podset-required-topology`). The **preferred** case is a harder optimization problem (best-effort balancing across domains, fallback when the whole group doesn't fit) and is left as future work.

The 2-PodSet / leader-worker assumption is also baked into the scheduler and ungater, which would need to be generalized:
- `findLeaderAndWorkers` ([`pkg/cache/scheduler/tas_flavor_snapshot.go:614-627`](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/cache/scheduler/tas_flavor_snapshot.go#L614)) — only reads `trs[0]` and `trs[1]`; a 3rd+ PodSet in a group is ignored during placement.
- topology ungater rank assignment ([`pkg/controller/tas/topology_ungater.go:201-221`](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/tas/topology_ungater.go#L201)) — handles a group only as a 2-element (smaller=leader, larger=workers) pair.

**Why is this needed**:

We have a reinforcement learning (RL) use case where **inference and training pods need to be placed on the same spine** for performance. These are distinct roles with different pod templates, modeled as separate PodSets, each with multiple replicas — and none is a singleton "leader". The current grouping model can't express this, since it violates both the "exactly 2 PodSets" and the "one must be <=1 replica" rules.

Related: #6242 (Add validation for PodSet Grouping)

**Completion requirements**:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

phoenix1712:
#### What type of PR is this?

/kind feature
/kind kep
/area tas

Generalizes TAS PodSet groups beyond the leader-worker pair so that, for the
**required** topology case, a group may contain `N >= 2` PodSets, each with
`count > 1` and no single-replica leader, co-located within a single topology
domain.

The motivating use case is reinforcement learning, where distinct
multi-replica roles (e.g. training and inference/rollout) must share a spine so
that weight-sync collectives over the high-bandwidth fabric do not cross
spines. The existing PodSet-group support requires exactly two PodSets with at
least one single-replica leader, which cannot express this.

The change is split into two commits:

1. **KEP-2724 update** — documents the generalization (required-only): routing,
   placement model (vector bin packing), per-PodSet node eligibility, ungating,
   validation changes, alternatives considered (including why the suggested
   `state(d, k)` DP is not adopted for the heterogeneous multi-resource case),
   scalability limits, and the test plan.
2. **Alpha prototype** behind a new `TASPodSetGroupGeneralization` feature gate
   (off by default):
   - **Routing**: the generalized path (validation, scheduling, ungating) is
     used when the gate is enabled, the group requests required topology, and it
     is **not** the legacy leader-worker shape (which keeps the existing path and
     merged leader/worker rank layout).
   - **Scheduler**: best-fit selection of a single required-level domain, then a
     greedy best-fit First-Fit-Decreasing pass that packs each PodSet's pods
     (with their real per-pod footprints) onto eligible leaves, honoring
     per-PodSet node constraints (taints, node selector, required/preferred
     affinity) via helpers shared with the single-PodSet path.
   - **Validation**: relaxes the two-PodSet/single-leader rules for gate-enabled
     required groups while keeping topology-consistency invariants.
   - **Ungater**: assigns each PodSet in a generalized group its own
     PodSet-local rank space.

This is intended as a prototype to align on the implementation approach for the
"Cross-PodSet Topology Aware scheduling" section of the TAS KEP, per the
discussion on the issue. Feedback on the algorithm and scope is very welcome.

Fixes #11953

- The feature gate `TASPodSetGroupGeneralization` is alpha and disabled by
  default; with it off, behavior is unchanged.
- Placement uses a single greedy best-fit pass (near-optimal, not exact). A
  fragmented-but-packable group may be conservatively reported as not fitting
  (false-negative pend); it is never mis-placed. Rationale and complexity are
  documented in the KEP's "Scalability and limitations" and "Alternatives
  considered" sections, including why the existing balanced-placement DP and the
  `state(d, k)` formulation do not apply to single-domain heterogeneous packing.
- The intersection with elastic workload slices (`ElasticJobsViaWorkloadSlices`)
  is out of scope: that path supports only unconstrained topology, and a group
  is identified by `podSetGroupName`, which elastic worker groups do not use.
- Manually validated on a cluster: co-location within a required domain,
  best-fit domain selection, node packing, and node-eligibility (a tainted node
  is excluded → workload pends → untaint → schedules).

AI assistance disclosure: this change was developed with the assistance of an
AI coding tool. All code, tests, and KEP content were reviewed by the author.

```release-note
TAS: added an alpha `TASPodSetGroupGeneralization` feature gate (disabled by default) that generalizes PodSet groups for the required-topology case to support groups of two or more PodSets, each with multiple replicas and no single-replica leader, co-located within a single topology domain.
```

### DIFF
--- a/keps/2724-topology-aware-scheduling/README.md
+++ b/keps/2724-topology-aware-scheduling/README.md
@@ -60,6 +60,10 @@
     - [Example](#example-2)
   - [Cross-PodSet Topology Aware scheduling](#cross-podset-topology-aware-scheduling)
     - [Ensure leader and workers end up on the same flavor](#ensure-leader-and-workers-end-up-on-the-same-flavor)
+    - [Generalizing PodSet Groups beyond leader-worker (N PodSets, no single-replica leader)](#generalizing-podset-groups-beyond-leader-worker-n-podsets-no-single-replica-leader)
+      - [Out of scope (future work)](#out-of-scope-future-work)
+      - [Alternatives considered](#alternatives-considered)
+      - [Scalability and limitations](#scalability-and-limitations)
   - [Enforcing the assignment](#enforcing-the-assignment)
   - [Support for Elastic Workloads](#support-for-elastic-workloads)
   - [Balanced placement](#balanced-placement)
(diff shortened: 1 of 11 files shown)


mimowo:
please sign CLA

phoenix1712:
Ah mistakenly used a diff email when signing the commits. Let me fix that.

mimowo:
@phoenix1712 is the PR kept still as a draft for early feedback or you are still working actively on some changes?

phoenix1712:
@mimowo I kept it in draft to first align on the approach.

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1667,6 +1671,228 @@ assignment unit to `PodSet Group`.
...
+
+**Use case.** The initial `PodSet Group` implementation (motivated by LeaderWorkerSet, see
+[Story 6](#story-6)) supports exactly **two** PodSets where at least one has `count == 1`
+(the "leader"). This cannot express reinforcement-learning workloads where the
+*inference/rollout* and *training* pods — distinct PodSets, each with multiple replicas and
+no singleton leader — must share a spine so weight-sync collectives do not cross the
+high-bandwidth fabric (see [#11953](https://github.com/kubernetes-sigs/kueue/issues/11953)).
+
+**What we expand.** For the **required** topology case, a group may now contain `N >= 2`
+PodSets, each with `count > 1` (no required single-replica leader). The semantics are
```
dkaluza, line 1684: Why do they have to be strictly greater than 1?
For me it looks like artificial requirement. What about scenarios when we wish to have for example:
1. A single leader for weights collection & distribution between training & inference podsets. n=1
2. Inference  podset with pod number n>1.
3.  Training podset with pod number n>1.
Sure one may want to replicate the leader for better availability but this obviously comes at a consistency/resources costs.

IMO we should design this functionality in a way that it also covers such scenarios.

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1667,6 +1671,228 @@ assignment unit to `PodSet Group`.
...
+unchanged: all PodSets sharing a `podset-group-name` are co-located within a single domain
+at the requested level.
+
+The routing is simple: the new generalized path (validation, scheduling, and ungating) is
+taken when the `TASPodSetGroupGeneralization` feature gate is enabled, the group requests
+required topology, **and** the group is not the legacy leader-worker shape (exactly two
+PodSets, one with a single replica). Everything else — the gate disabled, a group
+requesting preferred topology, or a legacy leader-worker group — uses the existing
```
dkaluza, line 1692: > requesting preferred topology

Is this just for now? Are we aiming to support preferred topology in the future in similar way?

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1667,6 +1671,228 @@ assignment unit to `PodSet Group`.
...
+
+**Placement.** Each PodSet keeps its real per-pod footprint, and its pods are placed only on
+nodes that PodSet is allowed to run on — the packer applies the same per-PodSet node
+filtering as the single-PodSet path (taints vs. tolerations, node selector, required node
+affinity, with `PodSetUpdates` and flavor tolerations merged in), via the shared
+`leafNodeEligibility` / `podSetNodeRequirements` helpers. The group's pods are packed into
+the leaves of one required-level domain. Since the PodSets have different footprints,
+fitting a group into a domain is a **multi-dimensional (vector) bin-packing** problem rather
+than a single "how many pods fit" count. The exact problem is NP-hard (and has no APTAS for
+two or more resource dimensions, Woeginger 1997), so we use a near-optimal heuristic: a
```
dkaluza, line 1705: >near-optimal heuristic 

What do we mean by "near-optimal" here? Is this a sort of K-approximation?

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1667,6 +1671,228 @@ assignment unit to `PodSet Group`.
...
+single **best-fit First-Fit-Decreasing** pass (per the vector-bin-packing literature,
+Panigrahy et al., "Heuristics for Vector Bin Packing", ESA 2011; Grandl et al., "Tetris",
+SIGCOMM 2014). Placing each pod (largest first) on the eligible leaf that leaves the least
+scarcity-weighted residual capacity keeps roomier leaves free for larger pods, consistent
+with the domain-level BestFit and the single-PodSet fit behavior. This is the same
+fragmentation concern tracked for the single-PodSet case in
+[#10574](https://github.com/kubernetes-sigs/kueue/issues/10574); the existing
+balanced-placement pseudo-polynomial DP solves *multi-domain selection over homogeneous
```
dkaluza, line 1713: > multi-domain selection over homogeneous slice counts

Is there a KEP from which I can learn more about current DP algorithm?
Maybe it is worth linking it here for easier reference?

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1667,6 +1671,228 @@ assignment unit to `PodSet Group`.
...
+
+**Algorithm.**
+
+```
+1. Group all PodSets sharing a podset-group-name; sum their per-pod requests to
+   determine the flavor (unchanged from today).
+2. Compute the group's real aggregate demand D_total = sum_i(count_i * footprint_i),
+   plus the total pod count.
+3. Pre-filter required-level domains whose aggregate free capacity (over leaves, minus
+   TAS and assumed usage) is >= D_total. If none qualify, the group does not fit and the
```
dkaluza, line 1725: what do you mean by minus TAS here?

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1667,6 +1671,228 @@ assignment unit to `PodSet Group`.
...
+   workload is not admitted (no fallback to coarser levels -- that is preferred behavior,
+   out of scope here). Rank the survivors tightest-fit first.
+4. For each surviving domain in order, run one greedy best-fit pass: place pods
+   largest-first (FFD) on the eligible leaf (one the PodSet may run on) that leaves the
+   least scarcity-weighted residual capacity. Return the first domain whose pass places
```
dkaluza, line 1730: How do we define "scarcity-weighed residual capacity"?

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1667,6 +1671,228 @@ assignment unit to `PodSet Group`.
...
+   all pods. The pass is O(pods * leaves * resources); it is near-optimal but NOT exact, so
+   a fragmented-but-packable group may be reported as not fitting (a conservative
+   false-negative; pods are never mis-placed).
```
dkaluza, line 1733: >but NOT exact, so
   a fragmented-but-packable group may be reported as not fitting (a conservative
   false-negative; pods are never mis-placed).

An example of not fitting sequence would be nice to clarify what does this mean.

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1667,6 +1671,228 @@ assignment unit to `PodSet Group`.
...
+
+    # 2. Best-fit search over domains at the required level.
+    best <- none ; bestSlack <- +inf
+    for domain in sortedDomains(level):     # deterministic order
+        leaves <- leavesUnder(domain)
+        free   <- [ leaf.capacity - tasUsage(leaf) - assumedUsage(leaf)
+                    for leaf in leaves ]
+        agg    <- sum(free)
+
+        slack <- howManyTimesFits(Dtotal, agg)   # CountIn: necessary pre-filter
```
dkaluza, line 1761: what do you mean by "CountIn" here?
how is this defined? min( agg[resource] /dtotal[resource]  for resource in resources)? or soemthing different?

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1667,6 +1671,228 @@ assignment unit to `PodSet Group`.
...
+        if slack < 1:        continue            # domain can never hold the group
+        if best != none and slack >= bestSlack:  continue   # cannot beat current best
+
+        placement <- PackGroupOntoLeaves(group, leaves, free)
+        if placement != none:
+            best <- placement ; bestSlack <- slack
```
dkaluza, line 1767: How do we know that this placement is really better than previous ones?
Isn't this the case that after Packing group onto leaves we can actually get worse placement? - the previous comparison was just on aggregation, in real placement we can make other resources unusable, e.g. we have done unused RAM on the node, that cannot be allocated anywhere else as we already allocated all of the cpus.

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1667,6 +1671,228 @@ assignment unit to `PodSet Group`.
...
+    totalCap <- sum(free)                       # for scarcity weights
+
+    remaining <- copy(free)
+    for i in 0..len(pods):
+        bestLeaf <- none ; bestScore <- +inf
+        for leaf in leaves:
+            if not eligible(pods[i].podSet, leaf):        continue   # taints / selector / affinity
+            if not fits(pods[i].need, remaining[leaf]):   continue
+            # scarcity-weighted leftover capacity AFTER placing the pod (lower = tighter)
+            score <- sum over r (sorted) of max(0, remaining[leaf][r] - pods[i].need[r]) / totalCap[r]
```
dkaluza, line 1788: what do you mean by "r" here? resource?

dkaluza, line 1788: why do we nead max here? isn't remaining[leaf][r] always >= pods[i].need[r]? How otherwise it fits?

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1667,6 +1671,228 @@ assignment unit to `PodSet Group`.
...
+nodes that PodSet is allowed to run on — the packer applies the same per-PodSet node
+filtering as the single-PodSet path (taints vs. tolerations, node selector, required node
+affinity, with `PodSetUpdates` and flavor tolerations merged in), via the shared
+`leafNodeEligibility` / `podSetNodeRequirements` helpers. The group's pods are packed into
+the leaves of one required-level domain. Since the PodSets have different footprints,
+fitting a group into a domain is a **multi-dimensional (vector) bin-packing** problem rather
+than a single "how many pods fit" count. The exact problem is NP-hard (and has no APTAS for
+two or more resource dimensions, Woeginger 1997), so we use a near-optimal heuristic: a
+single **best-fit First-Fit-Decreasing** pass (per the vector-bin-packing literature,
+Panigrahy et al., "Heuristics for Vector Bin Packing", ESA 2011; Grandl et al., "Tetris",
```
dkaluza, line 1707: can we have a more accurate reference?
I see 2 versions of FFD in the first reference, FFDSum and FFDProd. Proposed approach is sum but with what scoring heuristic, I guess something derived from Norm-based Greedy as it has residuals? Is the inverse of sum of available resources a known one? Do you have some literature reference to this one?
From my point of view it would be best to do some literature/experiments/data driven decision which heuristic we would like to use (e.g. there is an exponent scores transformation mentioned in the paper).

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1667,6 +1671,228 @@ assignment unit to `PodSet Group`.
...
+  two-PodSet/single-leader validation.
+
+##### Out of scope (future work)
+
+- **Preferred topology for generalized groups.** When the group requests `preferred`
+  rather than `required`, the desired degradation behavior (climb to coarser levels, then
+  best-effort spread across multiple domains) and, crucially, *whether the PodSets of a
+  group degrade in lockstep or may be split across domains independently*, is undesigned
+  upstream. The current two-PodSet code treats the group as one summed unit when it falls
+  back, but this has never been specified for groups. Generalized `preferred` groups are
```
dkaluza, line 1821: >  but this has never been specified for groups

Just an idea - probably one can also just threat all summed as one group... Though one would have to investigate into the consequences and performance of that in various scenarios.

I fully support postponing preferred groups for future PR

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1667,6 +1671,228 @@ assignment unit to `PodSet Group`.
...
+**Cost.** Let `P` be the total pods in the group, `N` the number of PodSets, `L` the leaves
+under a candidate domain, `D` the candidate domains at the required level, and `R` the
+resource dimensions (small constant, e.g. CPU/memory/GPU). Per candidate domain the work is:
+
+- per-PodSet node eligibility precompute: `O(N × L × R)` (taints/selector/affinity per
+  leaf, reusing the same filter as the single-PodSet path);
+- one greedy best-fit pass: `O(P × L × R)` (each pod scans the eligible leaves, scoring `R`
+  resources).
+
+Domains are aggregate-pre-filtered (`O(D × L × R)`) and sorted tightest-fit first
```
dkaluza, line 1871: "sorted tightest-fit first"
Aren't they sorted just for consistent ordering?
If this is tightest-fit then it should be probably on residual score sums or would require resource scarcity ordering(which can make it RxD log D if we assume comparison is in time R and special precaution in sorting algorithm is not taken)

from pseudo-algorithm:
>    for domain in sortedDomains(level):     # deterministic order

dkaluza: Also according to the pseudo code the sorting is of pods
>   pods <- [ (psIndex, ps.footprint + {pods:1}) for ps in group, repeated ps.count times ]
    sort pods by footprint descending          # FFD (dominant-resource order)

Assuming you are making this only once beforehand and you duplicate the results per pod count after appropriately it is O(R x N log N) without special sorting algorithm.

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1667,6 +1671,228 @@ assignment unit to `PodSet Group`.
...
+(`O(D log D)`); the packer then runs on candidate domains in order and stops at the first
```
dkaluza, line 1872: >the packer then runs on candidate domains in order and stops at the first that places all pods. 

not true in pseudo-code, unless I'm missing something.

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1667,6 +1671,228 @@ assignment unit to `PodSet Group`.
...
+that places all pods. In the common case (the tightest domain packs) only one domain is
+packed, so the dominant term is `O((P + N) × L × R)` — polynomial and suitable for the
```
dkaluza, line 1874: Why P + N? Is P total number of pods in the whole group of just in the podset? I assume the first one. 
Isn't P >= N, so it is basically O(P x L x R)  - which if I understand correctly should be dominant factor in `PackGroupOntoLeaves` which is run for every domain leading to O(Dx P x L x R)  time complexity for `FindGroupAssignment`.
Assuming sorting is non dominant factor.

dkaluza: If I understand the procedure also has to be run for each distinct group, but this just boils down to the same complexity and different number of pods.

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1667,6 +1671,228 @@ assignment unit to `PodSet Group`.
...
+candidate domain passes the aggregate pre-filter but fails to pack, runs the greedy pass on
+all of them: `O(D × (P + N) × L × R)`. There is no exponential path.
+
+**Not exact.** The single greedy pass is near-optimal but **not** exact: a
+fragmented-but-packable group can be reported as not fitting. Consequences, stated honestly:
+
+- A greedy strand makes the workload **pend** (a false-negative) even if a packing
+  technically exists. It is never mis-placed, and never placed on an ineligible node.
+- Best-fit FFD (largest pods first, onto the tightest eligible leaf) recovers most realistic
+  fragmentation cases, which is why a single greedy pass is acceptable for the alpha.
```
dkaluza, line 1885: >single greedy pass is acceptable for the alpha.

needs clarification - the pseudo-code is not making a single pass unless I'm missing something.

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1667,6 +1671,228 @@ assignment unit to `PodSet Group`.
...
+- This is a deliberate simplicity vs. placement-completeness tradeoff. Possible future
+  improvements (still polynomial): additional greedy orderings, a local-search swap phase,
+  or an exact search bounded to small groups.
+
+**Determinism.** Placement is reproducible across scheduling cycles: leaves under a domain
+are iterated in a stable `levelValues` order, candidate domains are sorted by slack then
+domain ID, the best-fit score iterates resources in sorted order (float addition is not
```
dkaluza, line 1892: > candidate domains are sorted by slack then domain ID,

Again I do not see this ordering in pseudo-code, am I missing something?

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1896,6 +2122,19 @@ well covered. In particular, the following scenarios need coverage:
 - un-gate pods by with the "topology" scheduling gate
 - the Workload can be suspended and unsuspended
 
+For the generalized PodSet groups feature (`TASPodSetGroupGeneralization`), additional
+coverage:
+- validation routing: a required group of `N > 2` PodSets (and a two-PodSet group with no
+  single-replica leader) is accepted with the gate on and rejected with it off; a
+  `preferred` generalized group is rejected via the legacy path; mismatched levels within a
+  group are rejected
+- scheduler packing: heterogeneous-footprint groups that fit / do not fit a domain;
```
dkaluza, line 2131: Some test regarding resource scarcity? That more scarce resources are prioritized during best-fitting?

dkaluza (commented):
Reviewed the KEP adjustments, there is a couple of inconsistencies that concern me. The pseudo-code does not fully match the descriptions. I would prefer to clarify/fix them before delving into the implementation.

Issue #12100, opened 2026-06-11
Support partial scaleup for ElasticWorkloadSlices


yaroslava-serdiuk:
**What would you like to be added**:
We need the ability for Kueue to partially admit an ElasticWorkloadSlice that represents a ScaleUp activity.

Currently, if a massive ScaleUp request exceeds the remaining cohort or flavor quota, the entire request is blocked. However Kueue should be able to admit a smaller, partial slice of the request—up to the maximum available quota limit—instead of taking an all-or-nothing approach.

**Why is this needed**:
For example, a RayCluster might trigger a massive scale-up request to its maximum configured pod count. If that full request slightly exceeds the available quota, Kueue blocks the entire ScaleUp activity.

However, elastic frameworks like Ray can still make great progress and run efficiently with a partial scale-up (e.g., getting 50 out of 100 requested workers). Admitting a partial slice ensures that available cluster capacity doesn't sit idle while waiting for a massive block of quota to clear.

**Completion requirements**:

This enhancement requires the following artifacts:

- [x] Design doc
- [ ] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

mimowo:
@mszadkow let me check on the status here, do we have:
1. support for batch/Job and KubeRay already? KubeRay means Ray RayService primarily I think
2. do we have integration tests for batch/Job and KubeRay for the baseline scenario?
3. do we have integration tests for batch/Job and KubeRay which have scale up -> eviction -> re-admission again?
4. e2e tests for batch/Job and KubeRay?

mszadkow:
> [@mszadkow](https://github.com/mszadkow) let me check on the status here, do we have:
> 
> 1. support for batch/Job and KubeRay already? KubeRay means Ray RayService primarily I think
> 2. do we have integration tests for batch/Job and KubeRay for the baseline scenario?
> 3. do we have integration tests for batch/Job and KubeRay which have scale up -> eviction -> re-admission again?
> 4. e2e tests for batch/Job and KubeRay?

1. KubeRay: yes, implemented. RayCluster and RayJob are exercised end to end at integration level. RayService shares the same code paths, so it should work, but nothing verifies it yet - I will provide it in a follow up 
batch/Job: no. It's explicitly a Beta graduation item
2. As before for Kuberay yes
3. Nothing covers that cycle, maybe worth adding but I can add while working on Give-back PodSetReducer (in progress)
[...]

PR #12281, opened 2026-06-16
Update partial admission kep to support ElasticJobs

yaroslava-serdiuk:
#### What type of PR is this?
/kind kep

Related to #12100 

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Updated partial admission behavior for workloads with multiple PodSets, improving how pod loss is minimized when selecting a fitable combination.
  * Expanded Ray controller partial-preemption eligibility and clarified how worker replica minimums map to PodSet minimums, including scheduling-gated leftovers.

* **Documentation**
  * Revised the multi-PodSet partial admission procedure and worked example, with clearer guidance on the proportional fallback.
  * Updated KEP metadata (authors/latest milestone) and adjusted Ray controller section formatting/notes for implementation history.

### DIFF
--- a/keps/420-partial-admission/README.md
+++ b/keps/420-partial-admission/README.md
@@ -79,7 +79,8 @@ type PodSet struct {
 
 In case the workload proposed for the current scheduling cycle, does not fit, with or without preemption, in the current available quota and any of its PodSets allow partial admission, try to find to find a lower counts combination that fits the available quota with or without borrowing.
 
-The search should be priority based assuming that PodSets are listed accordinly to priorities.
+The search could be priority based or proportianaly. It's a job-level setting, that is set by adding 'kueue.x-k8s.io/partial-admission-policy=priority-based' or 'kueue.x-k8s.io/partial-admission-policy=proportional' annotation. If the value is not set, the proportional policy is used. 
+The priority based policy assume that the PodSets are listed in descending order of priorities.
 
 The accepted number of pods in each PodSet are recorded in `workload.Status.Admission.PodSetAssignments[*].ResourceUsage.Count`
 
@@ -145,7 +146,7 @@ Additional research is needed into the potential usage of multiple variable coun
 
 ### RayJob/RayService/RayCluster controller
 
-The RayCluster with 'kueue.x-k8s.io/partial-preemption=true' annotation are eligible for partial preemption. The
+The RayCluster with 'kueue.x-k8s.io/partial-admission=true' annotation are eligible for partial admission. The
 RayCluster.workerGroupSpec[i].minReplicas will be translated to the PodSet.MinCount for the Worker PodSet.
 There will be no update on RayCluster object. If the number of admitted pods less than requested count, the leftover pods should remain scheduling gated. In order to achieve that the same mechanism as for elastic workloads will be applied – the podTemplate will be initially gated and then the correct amount of pods will be ungated.
 
@@ -176,5 +177,3 @@ The `scheduler` and `controllers/job` should be extended to cover the new capabi
 
 
 ## Alternatives
-
-Using proportional algorithm instead of priority based one, i.e each PodSet should be shrinked proportionaly.
--- a/keps/420-partial-admission/README.md
+++ b/keps/420-partial-admission/README.md
@@ -10,6 +10,8 @@
 - [Design Details](#design-details)
   - [Workload API](#workload-api)
   - [Scheduler / Flavorassignment](#scheduler--flavorassignment)
+    - [Priority-Based Policy (<code>priority-based</code>)](#priority-based-policy-priority-based)
+    - [Proportional Policy (<code>proportional</code> - default)](#proportional-policy-proportional---default)
   - [Jobframework](#jobframework)
   - [batch/Job controller](#batchjob-controller)
   - [kubeflow/MPIJob controller](#kubeflowmpijob-controller)
@@ -80,33 +82,102 @@ type PodSet struct {
 
 In case the workload proposed for the current scheduling cycle does not fit, with or without preemption, in the current available quota and any of its PodSets allow partial admission, try to find a lower counts combination that fits the available quota with or without borrowing.
 
-The search could be priority-based or proportionally. It's a job-level setting, that is set by adding 'kueue.x-k8s.io/partial-admission-policy=priority-based' or 'kueue.x-k8s.io/partial-admission-policy=proportional' annotation. If the value is not set, the proportional policy is used. 
-The priority-based policy assumes that the PodSets are listed in descending order of priorities.
+The policy for shrinking PodSet counts is configured via the `kueue.x-k8s.io/partial-admission-policy` annotation on the Job. The supported policies are:
+- **`proportional` (default)**: Shrinks the counts of all variable PodSets proportionally to their allowable reduction ranges.
+- **`priority-based`**: Shrinks the counts of the PodSets sequentially, assuming they are listed in descending order of priority (i.e., the first PodSet has the highest priority and the last PodSet has the lowest priority).
 
-The accepted number of pods in each PodSet is recorded in `workload.Status.Admission.PodSetAssignments[*].ResourceUsage.Count`
+#### Priority-Based Policy (`priority-based`)
 
-In order to evaluate the potential success of the preemption, the preemption process should be split into:
-- Target selection, which should select the workloads within cohort that will need to be evicted, if the list is empty the assignment will be treated as `NoFit`. This step will take place during nomination.
-- Preemption issuing, will do the actual eviction of the targets previously selected.
+Under the `priority-based` policy, Kueue assumes that the PodSets are listed in descending order of priority. To preserve the counts of higher-priority PodSets, Kueue shrinks the PodSets starting from the end (lowest priority) and moving towards the beginning (highest priority) as needed.
 
-**Partial admission with multiple variable PodSets count**
+Specifically, if multiple PodSets have variable counts, Kueue iterates over them in the order they are defined in the Workload spec, starting from the last one. It decreases the count of the current PodSet down to its `minCount` until the workload fits the available quota. If shrinking the last PodSet to its `minCount` is still not enough to fit, Kueue keeps it at its `minCount` and moves to the second-to-last PodSet, decreasing its count down to its `minCount`, and so on.
 
-If multiple PodSets within a workload have variable counts, Kueue will iterate over the PodSets in the order they are defined in the Workload spec, starting from the end, and will try to decrease their counts until a fit is found. If shrinking the last PodSet still results in a NoFit, Kueue will try to decrease the next PodSet while keeping the last one at its minimum count, and so on.
+**Example:**
+Consider a Job with three PodSets:
+- `ps0` (highest priority): `count: 1`, no `minCount` (cannot be shrunk).
+- `ps1` (medium priority): `count: 4`, `minCount: 2` (can be reduced by up to 2 pods).
+- `ps2` (lowest priority): `count: 20`, `minCount: 10` (can be reduced by up to 10 pods).
+
+Total requested pods: `1 + 4 + 20 = 25` pods.
+
+- **Scenario A: Available quota is 19 pods** (requires a reduction of 6 pods).
+  1. Kueue targets the lowest priority PodSet, `ps2`, and decreases its count by 6 (from 20 to 14).
+  2. The resulting counts are: `ps0: 1`, `ps1: 4`, `ps2: 14` (total 19 pods, fits the quota).
+  3. Admitted counts: `ps0: 1`, `ps1: 4`, `ps2: 14`.
+
+- **Scenario B: Available quota is 13 pods** (requires a reduction of 12 pods).
+  1. Kueue targets the lowest priority PodSet, `ps2`, and decreases its count to its minimum: `10` (reduction of 10 pods). The current total count is now `1 + 4 + 10 = 15`.
+  2. Since it still does not fit the quota of 13, Kueue keeps `ps2` at `10` and moves to the next lowest priority PodSet, `ps1`.
+  3. Kueue decreases `ps1` by the remaining 2 pods (from 4 to 2). The resulting total count is `1 + 2 + 10 = 13` pods.
+  4. Admitted counts: `ps0: 1`, `ps1: 2`, `ps2: 10`.
+
+- **Scenario C: Available quota is 10 pods** (requires a reduction of 15 pods).
+  1. Kueue targets the lowest priority PodSet, `ps2`, and decreases its count to its minimum: `10` (reduction of 10 pods). The current total count is now `1 + 4 + 10 = 15`.
+  2. Since it does not fit the quota of 10, Kueue keeps `ps2` at `10` and moves to the next lowest priority PodSet, `ps1`.
+  3. Kueue decreases `ps1` to its minimum: `2` (reduction of 2 pods). The current total count is now `1 + 2 + 10 = 13`.
+  4. Since it still does not fit the quota of 10, and the remaining PodSet `ps0` does not allow partial admission (has no `minCount`), the search fails.
+  5. The job remains unadmitted.
+
+#### Proportional Policy (`proportional` - default)
+
+Under the `proportional` policy, Kueue reduces the counts of all variable PodSets proportionally to their maximum allowable reduction range (i.e., `count - minCount`).
+
+For each PodSet $j$ that allows partial admission, let:
+- $C_j$ be the maximum requested count.
+- $M_j$ be the minimum acceptable count (`minCount`).
+- $D_j = C_j - M_j$ be the maximum reduction delta for PodSet $j$.
+- $T_D = \sum D_j$ be the sum of all maximum reduction deltas.
+
+To find the optimal fit, Kueue performs a binary search on a scaling index $i$ in the range $[0, T_D]$. For a given index $i$, the count for PodSet $j$ is computed as:
+$$currentCount_j = C_j - \lfloor \frac{D_j \times i}{T_D} \rfloor$$
+
+Kueue finds the smallest index $i$ (corresponding to the minimal reduction) for which the resulting counts fit the available quota.
+
+**Example:**
+Consider a Job with three PodSets:
+- `ps0`: `count: 1`, no `minCount` ($D_0 = 0$).
+- `ps1`: `count: 4`, `minCount: 2` ($D_1 = 2$).
+- `ps2`: `count: 20`, `minCount: 10` ($D_2 = 10$).
+
+Total requested pods: `1 + 4 + 20 = 25` pods.
+Total reduction delta: $T_D = 0 + 2 + 10 = 12$ pods.
+
+- **Scenario A: Available quota is 19 pods** (requires a reduction of at least 6 pods).
+  - For $i = 6$:
+    - `ps0` count: $1 - \lfloor \frac{0 \times 6}{12} \rfloor = 1 - 0 = 1$
+    - `ps1` count: $4 - \lfloor \frac{2 \times 6}{12} \rfloor = 4 - 1 = 3$
+    - `ps2` count: $20 - \lfloor \frac{10 \times 6}{12} \rfloor = 20 - 5 = 15$
+    - Total count: `1 + 3 + 15 = 19` pods. This fits the 19-pod quota.
+  - For $i = 5$:
+    - `ps0` count: $1 - \lfloor \frac{0 \times 5}{12} \rfloor = 1 - 0 = 1$
+    - `ps1` count: $4 - \lfloor \frac{2 \times 5}{12} \rfloor = 4 - 0 = 4$
+    - `ps2` count: $20 - \lfloor \frac{10 \times 5}{12} \rfloor = 20 - 4 = 16$
+    - Total count: `1 + 4 + 16 = 21` pods. This does not fit the 19-pod quota.
+  - Thus, the binary search selects $i = 6$.
+  - Admitted counts: `ps0: 1`, `ps1: 3`, `ps2: 15` (total 19 pods).
+
+- **Scenario B: Available quota is 13 pods** (requires a reduction of at least 12 pods).
+  - For $i = 12$:
+    - `ps0` count: $1 - \lfloor \frac{0 \times 12}{12} \rfloor = 1 - 0 = 1$
+    - `ps1` count: $4 - \lfloor \frac{2 \times 12}{12} \rfloor = 4 - 2 = 2$
+    - `ps2` count: $20 - \lfloor \frac{10 \times 12}{12} \rfloor = 20 - 10 = 10$
+    - Total count: `1 + 2 + 10 = 13` pods. This fits the 13-pod quota.
+  - For $i = 11$:
+    - `ps0` count: $1 - \lfloor \frac{0 \times 11}{12} \rfloor = 1 - 0 = 1$
+    - `ps1` count: $4 - \lfloor \frac{2 \times 11}{12} \rfloor = 4 - 1 = 3$
+    - `ps2` count: $20 - \lfloor \frac{10 \times 11}{12} \rfloor = 20 - 9 = 11$
+    - Total count: `1 + 3 + 11 = 15` pods. This does not fit the 13-pod quota.
+  - Thus, the binary search selects $i = 12$.
+  - Admitted counts: `ps0: 1`, `ps1: 2`, `ps2: 10` (total 13 pods).
+
+- **Scenario C: Available quota is 10 pods** (requires a reduction of at least 15 pods).
+  - Even with the maximum possible index $i = 12$ (maximum reduction), the total count is `1 + 2 + 10 = 13` pods, which does not fit in 10 pods.
+  - Thus, the binary search fails, and the job remains unadmitted.
 
 As an optimization, we will introduce an increased step: when a workload finds a combination that fits the available quota, Kueue could try to find the combination that can be reached with minimum pod loss by checking if any of the PodSets with count == min count can be increased up to the original count. 
 The increased step is not necessary for elastic workloads, but the non-elastic workloads will benefit from it.
 
-Starting with the PodSets:
-
-```go
-[]podSet { {Count: 1} , {Count: 4, MinCount: 2}, {Count: 20, MinCount: 10}}
-```
-
-And only being able to admit 19 pods, it should end up with
-
-```go
-[]podSet { {Count: 1} , {Count: 4}, {Count: 14}}
-```
+The accepted number of pods in each PodSet is recorded in `workload.Status.Admission.PodSetAssignments[*].ResourceUsage.Count`.
 
 ### Jobframework
 
--- a/keps/420-partial-admission/README.md
+++ b/keps/420-partial-admission/README.md
@@ -79,7 +79,7 @@ type PodSet struct {
 
 In case the workload proposed for the current scheduling cycle, does not fit, with or without preemption, in the current available quota and any of its PodSets allow partial admission, try to find to find a lower counts combination that fits the available quota with or without borrowing.
 
-The search should be optimized (binary search) and preserve the proportion of pods lost across the variable count PodSets.
+The search should be priority based assuming that PodSets are listed accordinly to priorities.
 
 The accepted number of pods in each PodSet are recorded in `workload.Status.Admission.PodSetAssignments[*].ResourceUsage.Count`
 
@@ -91,7 +91,8 @@ In order to evaluate the potential success of the preemption, the preemption pro
 
 If multiple PodSets within a workload have variable counts, Kueue will iterate over the PodSets in the order they are defined in the Workload spec, starting from the end, and will try to decrease their counts until a fit is found. If shrinking the last PodSet still results in a NoFit, Kueue will try to decrease the next PodSet while keeping the last one at its minimum count, and so on.
 
-When a workload finds a combination that fits the available quota, it tries to find the combination that can be reached with minimum pod loss by checking if any of the PodSets with count == min count can be increased up to the original count.
+As an optimisation we will introdce an increased step: when a workload finds a combination that fits the available quota, Kueue could try to find the combination that can be reached with minimum pod loss by checking if any of the PodSets with count == min count can be increased up to the original count. 
+The increased step is not nessesary for elastic workloads, but it the non-elastic workloads will benefit from it.
 
 Starting with the PodSets:
 
@@ -104,8 +105,6 @@ And only being able to admit 19 pods, it should end up with
 ```go
 []podSet { {Count: 1} , {Count: 4}, {Count: 14}}
 ```
-
-If the priority based partial admission is not sufficient, the "proportional" partial admission will be introduced.
 
 ### Jobframework
 
@@ -148,7 +147,7 @@ Additional research is needed into the potential usage of multiple variable coun
 
 The RayCluster with 'kueue.x-k8s.io/partial-preemption=true' annotation are eligible for partial preemption. The
 RayCluster.workerGroupSpec[i].minReplicas will be translated to the PodSet.MinCount for the Worker PodSet.
-There will be no update on RayCluster object. If the number of admitted pods less than requested count, the leftover pods should remain scheduling gated. In order to achive that the same mechanism as for elastic workloads will be applied – the podTemplate will be initially gated and then the correct amount of pods will be ungated.
+There will be no update on RayCluster object. If the number of admitted pods less than requested count, the leftover pods should remain scheduling gated. In order to achieve that the same mechanism as for elastic workloads will be applied – the podTemplate will be initially gated and then the correct amount of pods will be ungated.
 
 ### Test Plan
 
@@ -178,3 +177,4 @@ The `scheduler` and `controllers/job` should be extended to cover the new capabi
 
 ## Alternatives
 
+Using proportional algorithm instead of priority based one, i.e each PodSet should be shrinked proportionaly.
--- a/keps/420-partial-admission/README.md
+++ b/keps/420-partial-admission/README.md
@@ -17,6 +17,7 @@
   - [Test Plan](#test-plan)
     - [Unit Tests](#unit-tests)
     - [Integration tests](#integration-tests)
+    - [E2E tests](#e2e-tests)
   - [Graduation Criteria](#graduation-criteria)
 - [Implementation History](#implementation-history)
 - [Drawbacks](#drawbacks)
@@ -31,7 +32,7 @@ Add an optional way of allowing the partial admission of a workload if the full 
 
 In practice, not all Workloads require the parallel execution of all the `count` of a `PodSet`, for such cases having a way to partially reserve the quota in order to prevent starvation.
 
-For example if a batch/Job has parallelism is x and there is only quota available for y < x then the job could still be admitted if it can work with a lower parallelism.
+For example, if a batch/Job has parallelism x and there is only quota available for y < x, then the job could still be admitted if it can work with a lower parallelism.
 
 ### Goals
 
@@ -45,7 +46,7 @@ Kueue will not take any measure to ensure that the parent job respects the assig
 
 ## Proposal
 
-Change the way the flavor assigner work to support decrementing the pods count in order to find a better fit for the current workload.
+Change the way the flavor assigner works to support decrementing the pods count in order to find a better fit for the current workload.
 In case a partial fit is chosen, the jobframework reconciler should provide the admitted pod counts to the parent job before unsuspending it, in a similar fashion as the node selectors are provided. In case the job gets suspended, the original pod counts should be restored in order to allow a potential future admission with its original pod counts.
 
 ### User Stories
@@ -77,14 +78,14 @@ type PodSet struct {
 
 ### Scheduler / Flavorassignment
 
-In case the workload proposed for the current scheduling cycle, does not fit, with or without preemption, in the current available quota and any of its PodSets allow partial admission, try to find to find a lower counts combination that fits the available quota with or without borrowing.
+In case the workload proposed for the current scheduling cycle does not fit, with or without preemption, in the current available quota and any of its PodSets allow partial admission, try to find a lower counts combination that fits the available quota with or without borrowing.
 
-The search could be priority based or proportianaly. It's a job-level setting, that is set by adding 'kueue.x-k8s.io/partial-admission-policy=priority-based' or 'kueue.x-k8s.io/partial-admission-policy=proportional' annotation. If the value is not set, the proportional policy is used. 
-The priority based policy assume that the PodSets are listed in descending order of priorities.
+The search could be priority-based or proportionally. It's a job-level setting, that is set by adding 'kueue.x-k8s.io/partial-admission-policy=priority-based' or 'kueue.x-k8s.io/partial-admission-policy=proportional' annotation. If the value is not set, the proportional policy is used. 
+The priority-based policy assumes that the PodSets are listed in descending order of priorities.
 
-The accepted number of pods in each PodSet are recorded in `workload.Status.Admission.PodSetAssignments[*].ResourceUsage.Count`
+The accepted number of pods in each PodSet is recorded in `workload.Status.Admission.PodSetAssignments[*].ResourceUsage.Count`
 
-In order to evaluate the potential success of the preemption, the preemption process should be split in:
+In order to evaluate the potential success of the preemption, the preemption process should be split into:
 - Target selection, which should select the workloads within cohort that will need to be evicted, if the list is empty the assignment will be treated as `NoFit`. This step will take place during nomination.
 - Preemption issuing, will do the actual eviction of the targets previously selected.
 
@@ -92,8 +93,8 @@ In order to evaluate the potential success of the preemption, the preemption pro
 
 If multiple PodSets within a workload have variable counts, Kueue will iterate over the PodSets in the order they are defined in the Workload spec, starting from the end, and will try to decrease their counts until a fit is found. If shrinking the last PodSet still results in a NoFit, Kueue will try to decrease the next PodSet while keeping the last one at its minimum count, and so on.
 
-As an optimisation we will introdce an increased step: when a workload finds a combination that fits the available quota, Kueue could try to find the combination that can be reached with minimum pod loss by checking if any of the PodSets with count == min count can be increased up to the original count. 
-The increased step is not nessesary for elastic workloads, but it the non-elastic workloads will benefit from it.
+As an optimization, we will introduce an increased step: when a workload finds a combination that fits the available quota, Kueue could try to find the combination that can be reached with minimum pod loss by checking if any of the PodSets with count == min count can be increased up to the original count. 
+The increased step is not necessary for elastic workloads, but the non-elastic workloads will benefit from it.
 
 Starting with the PodSets:
 
@@ -115,7 +116,7 @@ type GenericJob interface {
 
 	
 -    // RunWithNodeAffinity will inject the node affinity extracting from workload to job and unsuspend the job.
-+    // RunWithPodSetsInfo will inject the node affinity and podsSet counts extracted from workload to the job and unsuspend the job.
++    // RunWithPodSetsInfo will inject the node affinity and podSet counts extracted from workload to the job and unsuspend the job.
 -    RunWithNodeAffinity(nodeSelectors []PodSetNodeSelector)
 +    RunWithPodSetsInfo(nodeSelectors []PodSetNodeSelector, podSetCounts []int32)
 -    // RestoreNodeAffinity will restore the original node affinity of job.
@@ -133,8 +134,8 @@ type GenericJob interface {
 Besides adapting `RunWithPodSetsInfo` and `RestorePodSetsInfo` it should also:
 
 - rework `PodSets()` to populate `MinCount` if the job is marked to support partial admission.
-  * jobs supporting partial admission should have a dedicated annotation. eg. `kueue.x-k8s.io/job-min-parallelism`, indicating the minimum `parallelism` acceptable by the job in case of partial admission.
-  * jobs which need the `completions` count kept in sync with `parallelism` should indicate this in a second annotation `kueue.x-k8s.io/job-completions-equal-parallelism`
+  * jobs supporting partial admission should have a dedicated annotation, e.g., `kueue.x-k8s.io/job-min-parallelism`, indicating the minimum `parallelism` acceptable by the job in case of partial admission.
+  * jobs which need the `completions` count kept in sync with `parallelism` should indicate this in a second annotation, `kueue.x-k8s.io/job-completions-equal-parallelism`
 - rework `EquivalentToWorkload` to account for potential differences in `PodSets` spec `Parallelism`.
 
 ### kubeflow/MPIJob controller
@@ -146,13 +147,13 @@ Additional research is needed into the potential usage of multiple variable coun
 
 ### RayJob/RayService/RayCluster controller
 
-The RayCluster with 'kueue.x-k8s.io/partial-admission=true' annotation are eligible for partial admission. The
+RayClusters with 'kueue.x-k8s.io/partial-admission=true' annotation are eligible for partial admission. The
 RayCluster.workerGroupSpec[i].minReplicas will be translated to the PodSet.MinCount for the Worker PodSet.
-There will be no update on RayCluster object. If the number of admitted pods less than requested count, the leftover pods should remain scheduling gated. In order to achieve that the same mechanism as for elastic workloads will be applied – the podTemplate will be initially gated and then the correct amount of pods will be ungated.
+There will be no update on RayCluster object. If the number of admitted pods is less than requested count, the leftover pods should remain scheduling gated. In order to achieve that, the same mechanism as for elastic workloads will be applied – the podTemplate will be initially gated and then the correct amount of pods will be ungated.
 
 ### Test Plan
 
-No regressions in the current test should be observed.
+No regressions in the current tests should be observed.
 
 [X] I/we understand the owners of the involved components may require updates to
 existing tests to make this code solid enough prior to committing the changes necessary
@@ -160,11 +161,41 @@ to implement this enhancement.
 
 #### Unit Tests
 
-The changes in the flavorassignment should be covered by unit tests
+- **Scheduler/Flavor Assignment**:
+  - `pkg/scheduler/scheduler_test.go`:
+    - `partial admission single variable pod set`: verifies flavor assignment with a single variable count PodSet.
+    - `partial admission single variable pod set, preempt first`: verifies preemption behavior when a workload can be admitted using partial admission.
+    - `partial admission single variable pod set, preempt with partial admission`: verifies that preemption triggers when partial admission alone is not enough.
+    - `partial admission multiple variable pod sets, proportional policy`: verifies shrinking order and flavor assignment when multiple variable count PodSets are defined using the default proportional policy.
+    - `partial admission multiple variable pod sets, priority-based policy`: verifies shrinking order when the priority-based policy is set, starting from the last PodSet.
+    - `partial admission disabled, multiple variable pod sets`: verifies that no partial admission is performed if features/annotations are not active.
+  - `pkg/scheduler/scheduler_tas_test.go`:
+    - `TAS workload gets scheduled as trimmed by partial admission`: verifies that Topology Aware Scheduling is compatible with partial admission.
+    - `reclaim within cohort; preempting with partial admission`: verifies that reclaiming quota within a cohort works alongside preemption and partial admission.
+
+- **Webhooks**:
+  - `pkg/webhooks/workload_webhook_test.go`: validates that `minCount` cannot be negative or larger than the base `count`.
+
+- **Controllers**:
+  - `pkg/controller/jobs/job/job_controller_test.go`: verifies job controller's `RunWithPodSetsInfo` and `RestorePodSetsInfo` logic, updating/restoring parallelism correctly when job is unsuspended or suspended.
 
 #### Integration tests
 
-The `scheduler` and `controllers/job` should be extended to cover the new capabilities.
+- **Controllers & Scheduler**:
+  - `test/integration/singlecluster/controller/jobs/job/job_controller_test.go`:
+    - `Should schedule jobs with partial admission`: verifies a complete integration flow where a Job with `kueue.x-k8s.io/job-min-parallelism` is suspended, partially admitted with reduced parallelism, and its original parallelism is restored when the workload is stopped.
+  - `test/integration/singlecluster/controller/jobs/raycluster/raycluster_controller_test.go`:
+    - `Should schedule RayClusters with partial admission priority policy`: verifies a complete integration flow where a RayCluster with partial admission annotation is admitted with reduced worker count according to priority policy.
+- **Workload Webhook**:
+  - `test/integration/singlecluster/webhook/core/workload_test.go`:
+    - `invalid podSet minCount (negative)`: verifies negative minCount values are rejected.
+    - `invalid podSet minCount (too big)`: verifies minCount larger than count is rejected.
+    - `too many variable count podSets`: verifies that workloads with multiple variable count PodSets are rejected.
+
+#### E2E tests
+
+- `test/e2e/singlecluster/baseline/job_test.go`:
+  - `Should partially admit the Job if configured and not fully fits`: verifies that a real Job is successfully admitted with a reduced parallelism count matching the available cluster resources.
 
 ### Graduation Criteria
 
--- a/keps/420-partial-admission/README.md
+++ b/keps/420-partial-admission/README.md
@@ -13,6 +13,7 @@
   - [Jobframework](#jobframework)
   - [batch/Job controller](#batchjob-controller)
   - [kubeflow/MPIJob controller](#kubeflowmpijob-controller)
+  - [RayJob/RayService/RayCluster controller](#rayjobrayserviceraycluster-controller)
   - [Test Plan](#test-plan)
     - [Unit Tests](#unit-tests)
     - [Integration tests](#integration-tests)
@@ -88,7 +89,9 @@ In order to evaluate the potential success of the preemption, the preemption pro
 
 **Partial admission with multiple variable PodSets count**
 
-If multiple PodSets within a workload have variable counts the way the counts should be decrease is highly dependent on the job framework needs, currently the following is proposed:
+If multiple PodSets within a workload have variable counts, Kueue will iterate over the PodSets in the order they are defined in the Workload spec, starting from the end, and will try to decrease their counts until a fit is found. If shrinking the last PodSet still results in a NoFit, Kueue will try to decrease the next PodSet while keeping the last one at its minimum count, and so on.
+
+When a workload finds a combination that fits the available quota, it tries to find the combination that can be reached with minimum pod loss by checking if any of the PodSets with count == min count can be increased up to the original count.
 
 Starting with the PodSets:
 
@@ -98,14 +101,11 @@ Starting with the PodSets:
 
 And only being able to admit 19 pods, it should end up with
 
-
 ```go
-[]podSet { {Count: 1} , {Count: 3}, {Count: 15}}
+[]podSet { {Count: 1} , {Count: 4}, {Count: 14}}
 ```
 
-With both pods that accept partial admission losing 50% of their range.
-
-The presented solution is not taking into account which of the variable count PodSets is causing the "NoFit" status and can potentially decrease the count of PodSets that will otherwise fit. If this behaviour is not suitable for a specific job framework, the integration layer of the framework can take into account limiting the variable count PodSets to 1.
+If the priority based partial admission is not sufficient, the "proportional" partial admission will be introduced.
 
 ### Jobframework
 
@@ -144,6 +144,12 @@ In case of MPIJob `j.Spec.RunPolicy.SchedulingPolicy.MinAvailable` can be used t
 Whether an MPIJob supports partial admission or not can be deduced based on `MinAvailable` without the need of a dedicated annotation.
 Additional research is needed into the potential usage of multiple variable count PodSets.
 
+### RayJob/RayService/RayCluster controller
+
+The RayCluster with 'kueue.x-k8s.io/partial-preemption=true' annotation are eligible for partial preemption. The
+RayCluster.workerGroupSpec[i].minReplicas will be translated to the PodSet.MinCount for the Worker PodSet.
+There will be no update on RayCluster object. If the number of admitted pods less than requested count, the leftover pods should remain scheduling gated. In order to achive that the same mechanism as for elastic workloads will be applied – the podTemplate will be initially gated and then the correct amount of pods will be ungated.
+
 ### Test Plan
 
 No regressions in the current test should be observed.
@@ -164,6 +170,7 @@ The `scheduler` and `controllers/job` should be extended to cover the new capabi
 
 
 ## Implementation History
+- 07/07/2023 - Partial admission for batch.job was added that support only 1 podSet with minCount value.
 
 
 ## Drawbacks
--- a/keps/420-partial-admission/README.md
+++ b/keps/420-partial-admission/README.md
@@ -10,7 +10,7 @@
 - [Design Details](#design-details)
   - [Workload API](#workload-api)
   - [Scheduler / Flavorassignment](#scheduler--flavorassignment)
-    - [Priority-Based Policy (<code>priority-based</code>)](#priority-based-policy-priority-based)
+    - [Order-Based Policy (<code>order-based</code>)](#order-based-policy-order-based)
     - [Proportional Policy (<code>proportional</code> - default)](#proportional-policy-proportional---default)
   - [Jobframework](#jobframework)
   - [batch/Job controller](#batchjob-controller)
@@ -84,15 +84,17 @@ In case the workload proposed for the current scheduling cycle does not fit, wit
 
 The policy for shrinking PodSet counts is configured via the `kueue.x-k8s.io/partial-admission-policy` annotation on the Job. The supported policies are:
 - **`proportional` (default)**: Shrinks the counts of all variable PodSets proportionally to their allowable reduction ranges.
-- **`priority-based`**: Shrinks the counts of the PodSets sequentially, assuming they are listed in descending order of priority (i.e., the first PodSet has the highest priority and the last PodSet has the lowest priority).
+- **`order-based`**: Shrinks the counts of the PodSets sequentially starting from the last one (suits for the cases when the podsets are ordered by priority). The Workload PodSet order is usually the same as the order of the PodSet in the Job spec. For the RayCluster the Workload PodSets starting from Head PodSet and following by WorkerGroupPodSets in the same order as in RayCluster.
 
-#### Priority-Based Policy (`priority-based`)
+#### Order-Based Policy (`order-based`)
 
-Under the `priority-based` policy, Kueue assumes that the PodSets are listed in descending order of priority. To preserve the counts of higher-priority PodSets, Kueue shrinks the PodSets starting from the end (lowest priority) and moving towards the beginning (highest priority) as needed.
-
+Under the `order-based` policy, Kueue shrinks the PodSets starting from the last one and moving towards the beginning as needed.
 Specifically, if multiple PodSets have variable counts, Kueue iterates over them in the order they are defined in the Workload spec, starting from the last one. It decreases the count of the current PodSet down to its `minCount` until the workload fits the available quota. If shrinking the last PodSet to its `minCount` is still not enough to fit, Kueue keeps it at its `minCount` and moves to the second-to-last PodSet, decreasing its count down to its `minCount`, and so on.
+As an optimization, we will introduce a second phase (similar to the preemption algorithm): when a workload finds a combination that fits the available quota, Kueue tries to gradually put the reduced counts back. In this phase, Kueue iterates over all PodSets from the first to the last one. For each PodSet that was reduced, Kueue tries to increase its count back to the original count. If that fits, Kueue keeps it. Otherwise, Kueue performs a binary search on the PodSet's count between the current count and the original count to find the maximum count that fits.
 
-**Example:**
+One example when order-based policy is used, is when the RayCluster has identical WorkerGroupPodSets, that have different node selectors that are tight to different node group capacity — for example, reservation/on-demand/spot. In this case, it is preferable to keep workers to run on reservation nodes than on-demand/spot nodes.
+
+**Examples:**
 Consider a Job with three PodSets:
 - `ps0` (highest priority): `count: 1`, no `minCount` (cannot be shrunk).
 - `ps1` (medium priority): `count: 4`, `minCount: 2` (can be reduced by up to 2 pods).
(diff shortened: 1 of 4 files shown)


keps/420-partial-admission/README.md
sohankunkerkar: I think you might need to update this information as this contradicts new algorithm proposed.

sohankunkerkar, line 397: Ideally, this point should have been filled as part of the beta process.

keps/420-partial-admission/README.md
```diff
@@ -164,6 +170,7 @@ The `scheduler` and `controllers/job` should be extended to cover the new capabi
 
 
 ## Implementation History
+- 07/07/2023 - Partial admission for batch.job was added that support only 1 podSet with minCount value.
 
 
 ## Drawbacks
```
sohankunkerkar, line 417: The same here.

sohankunkerkar: https://github.com/kubernetes-sigs/kueue/pull/12281/changes#r3431196655

keps/420-partial-admission/README.md
sohankunkerkar, line 420: The proportional algorithm (what the code does today) Is the alternative. It should be listed here with the tradeoff: proportional is fairer but less predictable per-PodSet, tail-first is simpler but order-dependent.

keps/420-partial-admission/README.md
```diff
@@ -88,7 +89,9 @@ In order to evaluate the potential success of the preemption, the preemption pro
 
 **Partial admission with multiple variable PodSets count**
 
-If multiple PodSets within a workload have variable counts the way the counts should be decrease is highly dependent on the job framework needs, currently the following is proposed:
+If multiple PodSets within a workload have variable counts, Kueue will iterate over the PodSets in the order they are defined in the Workload spec, starting from the end, and will try to decrease their counts until a fit is found. If shrinking the last PodSet still results in a NoFit, Kueue will try to decrease the next PodSet while keeping the last one at its minimum count, and so on.
```
sohankunkerkar: The change to tail-first reduction makes sense for Ray-style workloads where later PodSets (e.g., GPU workers) are more expendable than earlier ones (e.g., CPU preprocessors), but worth stating explicitly that podSets ordering acts as reduction priority.

keps/420-partial-admission/README.md
```diff
@@ -144,6 +144,12 @@ In case of MPIJob `j.Spec.RunPolicy.SchedulingPolicy.MinAvailable` can be used t
 Whether an MPIJob supports partial admission or not can be deduced based on `MinAvailable` without the need of a dedicated annotation.
 Additional research is needed into the potential usage of multiple variable count PodSets.
 
+### RayJob/RayService/RayCluster controller
+
+The RayCluster with 'kueue.x-k8s.io/partial-preemption=true' annotation are eligible for partial preemption. The
+RayCluster.workerGroupSpec[i].minReplicas will be translated to the PodSet.MinCount for the Worker PodSet.
+There will be no update on RayCluster object. If the number of admitted pods less than requested count, the leftover pods should remain scheduling gated. In order to achive that the same mechanism as for elastic workloads will be applied – the podTemplate will be initially gated and then the correct amount of pods will be ungated.
```
sohankunkerkar:
```suggestion
There will be no update on RayCluster object. If the number of admitted pods less than requested count, the leftover pods should remain scheduling gated. In order to achieve that the same mechanism as for elastic workloads will be applied – the podTemplate will be initially gated and then the correct amount of pods will be ungated.
```

keps/420-partial-admission/README.md
```diff
@@ -98,14 +101,11 @@ Starting with the PodSets:
...
-
 ```go
-[]podSet { {Count: 1} , {Count: 3}, {Count: 15}}
+[]podSet { {Count: 1} , {Count: 4}, {Count: 14}}
 ```
 
-With both pods that accept partial admission losing 50% of their range.
-
-The presented solution is not taking into account which of the variable count PodSets is causing the "NoFit" status and can potentially decrease the count of PodSets that will otherwise fit. If this behaviour is not suitable for a specific job framework, the integration layer of the framework can take into account limiting the variable count PodSets to 1.
+If the priority based partial admission is not sufficient, the "proportional" partial admission will be introduced.
```
sohankunkerkar: Can you please elaborate this point? like what triggers the switching?

keps/420-partial-admission/README.md
sohankunkerkar: Can you also cover the recovery step in more detail? It would help to include a worked example where recovery is actually triggered, as the current example doesn't demonstrate that path.

yaroslava-serdiuk:
@sohankunkerkar thanks for review! I'll be on vacation for (almost) 2 weeks, will address your comments when I'm back.

keps/420-partial-admission/README.md
yaroslava-serdiuk: What do you mean by recovery?

keps/420-partial-admission/README.md
yaroslava-serdiuk: I meant that we can introduce "proportional" partial admission if we have explicit request for it. Actually not sure I need to mention it here. Moved proportional algorithm to the alternatives section.

keps/420-partial-admission/README.md
yaroslava-serdiuk: I added proportional algorithm as an alternative. 
>what the code does today

Not really. The proportional algorithm is mentioned in this kep, but at the moment we support only one PodSet:
https://github.com/kubernetes-sigs/kueue/blob/main/apis/kueue/v1beta2/workload_types.go#L590 
https://github.com/kubernetes-sigs/kueue/blob/main/pkg/webhooks/workload_webhook.go#L116

yaroslava-serdiuk: Agree, but PartialAdmission feature is already in beta: https://github.com/kubernetes-sigs/kueue/blob/main/pkg/features/kube_features.go#L472

keps/420-partial-admission/README.md
sohankunkerkar: By "recovery" I meant the increase step described in the KEP after shrinking from the tail to find a fit (e.g., reaching {1, 2, 10} = 13 pods), showing how Kueue tries increasing each PodSet back toward its original count to reach the minimum-loss combination ({1, 4, 14} = 19 pods). The current example only shows the final result, not that intermediate step.

yaroslava-serdiuk: In the current example:
1. The starting point is {1, 4, 20}.
2. Kueue tries to shrink the last podSet (using the existing approach) and finds the {1, 4, 14} combination.
3. If {1, 4, 10} still doesn't fit, Kueue evaluates {1, 2, 10} and {1, 3, 10}.

We could also add an additional step: once a suitable combination is found, Kueue tries to increase the replicas from the last podSet to find the absolute optimal combination (similar to what we do in preemption). However, I didn't include that here because this change specifically targets elastic workloads. For elastic workloads, we will have a WorkloadSlice that represents the original count and will handle verifying the increased replicas for that slice.

sohankunkerkar: Makes sense! the WorkloadSlice recovery path covers this for elastic. Worth noting that non-elastic multi-PodSet partial admission would benefit from the increase step if we support it later, but fine to defer.

yaroslava-serdiuk: Done

yaroslava-serdiuk:
@sohankunkerkar I tried to reply/address all your comments. Please, let me know if anything is missed.

keps/420-partial-admission/README.md
```diff
@@ -78,7 +79,7 @@ type PodSet struct {
 
 In case the workload proposed for the current scheduling cycle, does not fit, with or without preemption, in the current available quota and any of its PodSets allow partial admission, try to find to find a lower counts combination that fits the available quota with or without borrowing.
 
-The search should be optimized (binary search) and preserve the proportion of pods lost across the variable count PodSets.
+The search should be priority based assuming that PodSets are listed accordinly to priorities.
```
sohankunkerkar:
```suggestion
The search should be priority based assuming that PodSets are listed accordingly to priorities.
```

keps/420-partial-admission/README.md
```diff
@@ -88,7 +89,10 @@ In order to evaluate the potential success of the preemption, the preemption pro
 
 **Partial admission with multiple variable PodSets count**
 
-If multiple PodSets within a workload have variable counts the way the counts should be decrease is highly dependent on the job framework needs, currently the following is proposed:
+If multiple PodSets within a workload have variable counts, Kueue will iterate over the PodSets in the order they are defined in the Workload spec, starting from the end, and will try to decrease their counts until a fit is found. If shrinking the last PodSet still results in a NoFit, Kueue will try to decrease the next PodSet while keeping the last one at its minimum count, and so on.
+
+As an optimisation we will introdce an increased step: when a workload finds a combination that fits the available quota, Kueue could try to find the combination that can be reached with minimum pod loss by checking if any of the PodSets with count == min count can be increased up to the original count.
```
sohankunkerkar:
```suggestion
As an optimization we will introduce an increased step: when a workload finds a combination that fits the available quota, Kueue could try to find the combination that can be reached with minimum pod loss by checking if any of the PodSets with count == min count can be increased up to the original count. 
```

keps/420-partial-admission/README.md
```diff
@@ -88,7 +89,10 @@ In order to evaluate the potential success of the preemption, the preemption pro
 
 **Partial admission with multiple variable PodSets count**
 
-If multiple PodSets within a workload have variable counts the way the counts should be decrease is highly dependent on the job framework needs, currently the following is proposed:
+If multiple PodSets within a workload have variable counts, Kueue will iterate over the PodSets in the order they are defined in the Workload spec, starting from the end, and will try to decrease their counts until a fit is found. If shrinking the last PodSet still results in a NoFit, Kueue will try to decrease the next PodSet while keeping the last one at its minimum count, and so on.
+
+As an optimisation we will introdce an increased step: when a workload finds a combination that fits the available quota, Kueue could try to find the combination that can be reached with minimum pod loss by checking if any of the PodSets with count == min count can be increased up to the original count. 
+The increased step is not nessesary for elastic workloads, but it the non-elastic workloads will benefit from it.
```
sohankunkerkar:
```suggestion
The increased step is not necessary for elastic workloads, but the non-elastic workloads will benefit from it.
```

keps/420-partial-admission/README.md
```diff
@@ -164,10 +169,12 @@ The `scheduler` and `controllers/job` should be extended to cover the new capabi
...
 ## Implementation History
+- 07/07/2023 - Partial admission for batch.job was added that support only 1 podSet with minCount value.
 
 
 ## Drawbacks
 
 
 ## Alternatives
 
+Using proportional algorithm instead of priority based one, i.e each PodSet should be shrinked proportionaly.
```
sohankunkerkar:
```suggestion
Using proportional algorithm instead of priority based one, i.e each PodSet should be shrunk proportionaly.
```

keps/420-partial-admission/README.md
```diff
@@ -144,6 +143,12 @@ In case of MPIJob `j.Spec.RunPolicy.SchedulingPolicy.MinAvailable` can be used t
 Whether an MPIJob supports partial admission or not can be deduced based on `MinAvailable` without the need of a dedicated annotation.
 Additional research is needed into the potential usage of multiple variable count PodSets.
 
+### RayJob/RayService/RayCluster controller
+
+The RayCluster with 'kueue.x-k8s.io/partial-preemption=true' annotation are eligible for partial preemption. The
```
sohankunkerkar: should this be `partial-admission` to match existing terminology (PartialAdmission FG, partialAssignment struct)?

keps/420-partial-admission/README.md
```diff
@@ -144,6 +143,12 @@ In case of MPIJob `j.Spec.RunPolicy.SchedulingPolicy.MinAvailable` can be used t
...
+RayCluster.workerGroupSpec[i].minReplicas will be translated to the PodSet.MinCount for the Worker PodSet.
+There will be no update on RayCluster object. If the number of admitted pods less than requested count, the leftover pods should remain scheduling gated. In order to achieve that the same mechanism as for elastic workloads will be applied – the podTemplate will be initially gated and then the correct amount of pods will be ungated.
+
 ### Test Plan
```
sohankunkerkar, line 352: While we are here, could you also update the test plan?

sohankunkerkar (commented):
Thanks for the update! Just a few nits but overall this looks good to me!

keps/420-partial-admission/README.md
yaroslava-serdiuk: updated test plan by including the existing tests and adding test for priority based algorithm

yaroslava-serdiuk:
@sohankunkerkar I updated the kep. I decided to leave the proportional algorithm since it's already implemented and just have one more priority based algorithm. The partial admission policy will be set by annotation on job, and the appropriate PodSetReducer will be created during scheduler cycle.

keps/420-partial-admission/README.md
```diff
@@ -76,19 +78,23 @@ type PodSet struct {
 
 ### Scheduler / Flavorassignment
 
-In case the workload proposed for the current scheduling cycle, does not fit, with or without preemption, in the current available quota and any of its PodSets allow partial admission, try to find to find a lower counts combination that fits the available quota with or without borrowing.
+In case the workload proposed for the current scheduling cycle does not fit, with or without preemption, in the current available quota and any of its PodSets allow partial admission, try to find a lower counts combination that fits the available quota with or without borrowing.
 
-The search should be optimized (binary search) and preserve the proportion of pods lost across the variable count PodSets.
+The search could be priority-based or proportionally. It's a job-level setting, that is set by adding 'kueue.x-k8s.io/partial-admission-policy=priority-based' or 'kueue.x-k8s.io/partial-admission-policy=proportional' annotation. If the value is not set, the proportional policy is used.
```
mwielgus: Is this a new annotation or has it been defined elsewhere? https://kueue.sigs.k8s.io/docs/reference/labels-and-annotations doesn't mention it.

keps/420-partial-admission/README.md
```diff
@@ -76,19 +78,23 @@ type PodSet struct {
...
 
-In order to evaluate the potential success of the preemption, the preemption process should be split in:
+In order to evaluate the potential success of the preemption, the preemption process should be split into:
 - Target selection, which should select the workloads within cohort that will need to be evicted, if the list is empty the assignment will be treated as `NoFit`. This step will take place during nomination.
 - Preemption issuing, will do the actual eviction of the targets previously selected.
 
 **Partial admission with multiple variable PodSets count**
 
-If multiple PodSets within a workload have variable counts the way the counts should be decrease is highly dependent on the job framework needs, currently the following is proposed:
+If multiple PodSets within a workload have variable counts, Kueue will iterate over the PodSets in the order they are defined in the Workload spec, starting from the end, and will try to decrease their counts until a fit is found. If shrinking the last PodSet still results in a NoFit, Kueue will try to decrease the next PodSet while keeping the last one at its minimum count, and so on.
```
mwielgus: Do we need to support multiple partial admission pod-set? What is the use case?

keps/420-partial-admission/README.md
```diff
@@ -76,19 +78,23 @@ type PodSet struct {
 
 ### Scheduler / Flavorassignment
 
-In case the workload proposed for the current scheduling cycle, does not fit, with or without preemption, in the current available quota and any of its PodSets allow partial admission, try to find to find a lower counts combination that fits the available quota with or without borrowing.
+In case the workload proposed for the current scheduling cycle does not fit, with or without preemption, in the current available quota and any of its PodSets allow partial admission, try to find a lower counts combination that fits the available quota with or without borrowing.
 
-The search should be optimized (binary search) and preserve the proportion of pods lost across the variable count PodSets.
+The search could be priority-based or proportionally. It's a job-level setting, that is set by adding 'kueue.x-k8s.io/partial-admission-policy=priority-based' or 'kueue.x-k8s.io/partial-admission-policy=proportional' annotation. If the value is not set, the proportional policy is used. 
+The priority-based policy assumes that the PodSets are listed in descending order of priorities.
```
mwielgus: Can we get more description on the policies? Ideally with some examples.

keps/420-partial-admission/README.md
yaroslava-serdiuk: This is a new annotation.

keps/420-partial-admission/README.md
yaroslava-serdiuk: Supporting partial admission across only a single PodSet is highly limiting. 
Current partial admission in Kueue relies on the kueue.x-k8s.io/job-min-parallelism annotation. This annotation only allows you to define a single minimum count for the entire job.

The primary use case is support for elastic workloads like an autoscaling RayCluster, which inherently split their work into multiple distinct PodSets (such as a Head node group and one or more elastic Worker node groups).
For example, if a RayCluster needs 1 Head pod (which is non-negotiable) and a variable number of Worker pods (e.g., minimum 5, target 20), a single global parameter (kueue.x-k8s.io/job-min-parallelism annotation that kueue currently support) is not flexible enough to negotiate different thresholds for different pod sets.

keps/420-partial-admission/README.md
yaroslava-serdiuk: Done

keps/420-partial-admission/README.md
```diff
@@ -76,36 +80,104 @@ type PodSet struct {
...
-
-The presented solution is not taking into account which of the variable count PodSets is causing the "NoFit" status and can potentially decrease the count of PodSets that will otherwise fit. If this behaviour is not suitable for a specific job framework, the integration layer of the framework can take into account limiting the variable count PodSets to 1.
+In case the workload proposed for the current scheduling cycle does not fit, with or without preemption, in the current available quota and any of its PodSets allow partial admission, try to find a lower counts combination that fits the available quota with or without borrowing.
+
+The policy for shrinking PodSet counts is configured via the `kueue.x-k8s.io/partial-admission-policy` annotation on the Job. The supported policies are:
+- **`proportional` (default)**: Shrinks the counts of all variable PodSets proportionally to their allowable reduction ranges.
+- **`priority-based`**: Shrinks the counts of the PodSets sequentially, assuming they are listed in descending order of priority (i.e., the first PodSet has the highest priority and the last PodSet has the lowest priority).
```
mwielgus: Explain more clearly how the list is provided - podset importance/priority matches the order they are mentioned in the workload object. Mention how the the order in the podset matches the order in jobset/rayjob/lws/whatever.

mwielgus: What about renaming it to something that doesn't have priority? "Priority" is already booked for workload/pod numerical priority and may cause confusion. Maybe something like `podset-order-based`?

keps/420-partial-admission/README.md
```diff
@@ -76,36 +80,104 @@ type PodSet struct {
...
+
+#### Priority-Based Policy (`priority-based`)
+
+Under the `priority-based` policy, Kueue assumes that the PodSets are listed in descending order of priority. To preserve the counts of higher-priority PodSets, Kueue shrinks the PodSets starting from the end (lowest priority) and moving towards the beginning (highest priority) as needed.
```
mwielgus: Can you mention the real world situation (not just an theoretical example) where this strategy is more suitable than proportional?

keps/420-partial-admission/README.md
```diff
@@ -76,36 +80,104 @@ type PodSet struct {
...
+Under the `proportional` policy, Kueue reduces the counts of all variable PodSets proportionally to their maximum allowable reduction range (i.e., `count - minCount`).
+
+For each PodSet $j$ that allows partial admission, let:
+- $C_j$ be the maximum requested count.
+- $M_j$ be the minimum acceptable count (`minCount`).
+- $D_j = C_j - M_j$ be the maximum reduction delta for PodSet $j$.
+- $T_D = \sum D_j$ be the sum of all maximum reduction deltas.
+
+To find the optimal fit, Kueue performs a binary search on a scaling index $i$ in the range $[0, T_D]$. For a given index $i$, the count for PodSet $j$ is computed as:
+$$currentCount_j = C_j - \lfloor \frac{D_j \times i}{T_D} \rfloor$$
```
mwielgus: How the pod count is rounded?

keps/420-partial-admission/README.md
```diff
@@ -76,36 +80,104 @@ type PodSet struct {
...
+- **`proportional` (default)**: Shrinks the counts of all variable PodSets proportionally to their allowable reduction ranges.
+- **`priority-based`**: Shrinks the counts of the PodSets sequentially, assuming they are listed in descending order of priority (i.e., the first PodSet has the highest priority and the last PodSet has the lowest priority).
+
+#### Priority-Based Policy (`priority-based`)
+
+Under the `priority-based` policy, Kueue assumes that the PodSets are listed in descending order of priority. To preserve the counts of higher-priority PodSets, Kueue shrinks the PodSets starting from the end (lowest priority) and moving towards the beginning (highest priority) as needed.
+
+Specifically, if multiple PodSets have variable counts, Kueue iterates over them in the order they are defined in the Workload spec, starting from the last one. It decreases the count of the current PodSet down to its `minCount` until the workload fits the available quota. If shrinking the last PodSet to its `minCount` is still not enough to fit, Kueue keeps it at its `minCount` and moves to the second-to-last PodSet, decreasing its count down to its `minCount`, and so on.
```
mwielgus, line 189: We have a similar problem in preemption algorithm. First we try to find a set of preemption candidates that allows to fit the considered workload at all, then we gradually try to put the unneeded candidates back, to ensure that the preemption list is somewhat minimal.

Here, if I understand correctly, you try to find a sizing that fits, but you don't try to maximize the number of pods once you find a workable sizing (no phase 2 equivalen of the preemption algorithm). Is this desired?

keps/420-partial-admission/README.md
yaroslava-serdiuk: To the lower or equal int: https://github.com/kubernetes-sigs/kueue/blob/main/pkg/scheduler/flavorassigner/podset_reducer.go#L59

keps/420-partial-admission/README.md
yaroslava-serdiuk: Updated

yaroslava-serdiuk: Done

keps/420-partial-admission/README.md
yaroslava-serdiuk: Added below

keps/420-partial-admission/README.md
yaroslava-serdiuk: Added a phase 2 to the partial admission as well.

keps/420-partial-admission/README.md
```diff
@@ -76,36 +80,106 @@ type PodSet struct {
...
+
+Under the `order-based` policy, Kueue shrinks the PodSets starting from the last one and moving towards the beginning as needed.
+Specifically, if multiple PodSets have variable counts, Kueue iterates over them in the order they are defined in the Workload spec, starting from the last one. It decreases the count of the current PodSet down to its `minCount` until the workload fits the available quota. If shrinking the last PodSet to its `minCount` is still not enough to fit, Kueue keeps it at its `minCount` and moves to the second-to-last PodSet, decreasing its count down to its `minCount`, and so on.
+As an optimization, we will introduce a second phase (similar to the preemption algorithm): when a workload finds a combination that fits the available quota, Kueue tries to gradually put the reduced counts back. In this phase, Kueue iterates over all PodSets from the first to the last one. For each PodSet that was reduced, Kueue tries to increase its count back to the original count. If that fits, Kueue keeps it. Otherwise, Kueue performs a binary search on the PodSet's count between the current count and the original count to find the maximum count that fits.
```
mwielgus, line 190: Can we have an example covering phase 2?

yaroslava-serdiuk: Added Scenario D to cover it

mimowo:
/hold
I don't think this PR is ready to merge. It lacks essential details to be ready for implementation - design details detailing how the mechanism will work, and user stories explaining **why** we do the proportional mechanism - I have reservation against that.

To start addressing the mechanics I propose to move the other PR here https://github.com/kubernetes-sigs/kueue/pull/12890, and work on the "Design Details" section of the Partial Admission KEP. The updates to the ElasticJobs KEP could be mechanical, just for x-referencing and small notes.

I will post more details comments to guide the review in more detail.

keps/420-partial-admission/README.md
```diff
@@ -42,9 +46,11 @@ Since this is an opt-in feature, the parent job should accept the partial admiss
 
 Kueue will not take any measure to ensure that the parent job respects the assigned quota.
 
+Partial admission is not allowed for concurrent admission in order to not increase the complexity. The webhook will reject workloads that support partial admission to be created for a cluster queue with concurrent admission policy.
```
mimowo: Please also add the note to the Concurrent Admission KEP.

keps/420-partial-admission/kep.yaml
mimowo, line 19: Please x-reference the KEP with ElasticJobs KEP and ConcurrentAdmission KEPS, using the syntax as here: https://github.com/kubernetes-sigs/kueue/blob/main/keps/2724-topology-aware-scheduling/kep.yaml#L18-L20

mimowo, line 13: Add me here

mimowo, line 35: Add a new FG we need to introduce, probably PartialAdmissionViaElasticJobs

keps/420-partial-admission/README.md
mimowo, line 75: Please add a user story explaining motivation for the ongoing work.

mimowo:
/hold
Let's merge this content with the other PR https://github.com/kubernetes-sigs/kueue/pull/12281.
The conceptual overlap is too strong to keep them separate.
The design details of the mechanism belong actually to the Partial Admission KEP, as it is the Partial Admission KEP which details the semantic for the MinCount fields: https://github.com/kubernetes-sigs/kueue/pull/12281/changes#diff-90141b55d447a6883a0cf57f2e592640204e21d396a1212df7ae0d4963fa680bR62-R80

In this KEP I imagine the updates are fairly minimal.

The alternative, instead of merging is to have two KEP updates:
1. Enable PartialAdmissionViaElasticJobs
2. Enable PartialAdmission via ElsaticJobs for multiple PodSets

keps/420-partial-admission/README.md
mimowo, line 138: Update for the semantic of the fields is required in the context of the new feature. Specifically we should explain that .spec.count will correspond now to the Job .spec.parallelism as requested by the user.

keps/420-partial-admission/README.md
```diff
@@ -76,36 +82,115 @@ type PodSet struct {
...
-```go
-[]podSet { {Count: 1} , {Count: 3}, {Count: 15}}
-```
-
-With both pods that accept partial admission losing 50% of their range.
-
-The presented solution is not taking into account which of the variable count PodSets is causing the "NoFit" status and can potentially decrease the count of PodSets that will otherwise fit. If this behaviour is not suitable for a specific job framework, the integration layer of the framework can take into account limiting the variable count PodSets to 1.
+In case the workload proposed for the current scheduling cycle does not fit, with or without preemption, in the current available quota and any of its PodSets allow partial admission, try to find a lower counts combination that fits the available quota with or without borrowing.
+
+The policy for shrinking PodSet counts is configured via the `kueue.x-k8s.io/partial-admission-policy` annotation on the Job. The supported policies are:
```
mimowo: Please define the annotation in the Design Details section for the API.

keps/420-partial-admission/README.md
```diff
@@ -76,36 +82,115 @@ type PodSet struct {
...
+- **`proportional` (default)**: Shrinks the counts of all variable PodSets proportionally to their allowable reduction ranges.
```
mimowo: Actually, I think the proportional has signifficant issues:
1.  when handling heterogenous PodSets - it would scale both CPU and GPU.
2. even for homogenous PodSets it cannot find simple solutions like (1,2), or (2, 1) to scale down from (2,2). It will just find (1,1) [assuming we have quota for 3].

So, I would like to request parking the proportional mode. Unless we can expand the User story section with a use case which really requires this policy

keps/420-partial-admission/README.md
```diff
@@ -76,36 +82,115 @@ type PodSet struct {
...
+- **`order-based`**: Shrinks the counts of the PodSets sequentially starting from the last one (suits for the cases when the podsets are ordered by priority). The Workload PodSet order is usually the same as the order of the PodSet in the Job spec. For the RayCluster the Workload PodSets starting from Head PodSet and following by WorkerGroupPodSets in the same order as in RayCluster.
+
+#### Order-Based Policy (`order-based`)
```
mimowo: Note that currently PodSets in our API are not ordered. So this policy needs to first define the order.
One idea is to use alphabetical order, but it should be explicitly noted.

keps/420-partial-admission/README.md
```diff
@@ -144,30 +229,66 @@ In case of MPIJob `j.Spec.RunPolicy.SchedulingPolicy.MinAvailable` can be used t
 Whether an MPIJob supports partial admission or not can be deduced based on `MinAvailable` without the need of a dedicated annotation.
 Additional research is needed into the potential usage of multiple variable count PodSets.
 
+### RayJob/RayService/RayCluster controller
+
+RayClusters with 'kueue.x-k8s.io/partial-admission=true' annotation are eligible for partial admission. The
```
mimowo: Please add the annotation to the Design Details APi section

keps/420-partial-admission/README.md
```diff
@@ -144,30 +229,66 @@ In case of MPIJob `j.Spec.RunPolicy.SchedulingPolicy.MinAvailable` can be used t
...
+RayCluster.workerGroupSpec[i].minReplicas will be translated to the PodSet.MinCount for the Worker PodSet.
+There will be no update on RayCluster object. If the number of admitted pods is less than requested count, the leftover pods should remain scheduling gated. In order to achieve that, the same mechanism as for elastic workloads will be applied – the podTemplate will be initially gated and then the correct amount of pods will be ungated.
```
mimowo: I think we should actually aim with the mechanism to replace the old one also for Job. Maintaining two different mechanism depending on the Job CRD will be nightmare. 

So I want to propose to introduce a new FG like `PartialAdmissionViaElasticJobsForJob` to allow eventual migration of the pre-existing mechanism to the new one also for Job.

keps/420-partial-admission/README.md
```diff
@@ -76,36 +82,115 @@ type PodSet struct {
...
+In case the workload proposed for the current scheduling cycle does not fit, with or without preemption, in the current available quota and any of its PodSets allow partial admission, try to find a lower counts combination that fits the available quota with or without borrowing.
+
+The policy for shrinking PodSet counts is configured via the `kueue.x-k8s.io/partial-admission-policy` annotation on the Job. The supported policies are:
+- **`proportional` (default)**: Shrinks the counts of all variable PodSets proportionally to their allowable reduction ranges.
+- **`order-based`**: Shrinks the counts of the PodSets sequentially starting from the last one (suits for the cases when the podsets are ordered by priority). The Workload PodSet order is usually the same as the order of the PodSet in the Job spec. For the RayCluster the Workload PodSets starting from Head PodSet and following by WorkerGroupPodSets in the same order as in RayCluster.
```
mimowo: Should we name it `greedy-reduce-with-fillback` to better reflect on the mechanism .The fact this is order-based does not really tell much by itself.

keps/420-partial-admission/README.md
mimowo: Also, the good part about dropping proportional is that we would not need to introduce the `kueue.x-k8s.io/partial-admission-policy` annotation at all for now.

keps/420-partial-admission/README.md
mimowo: Please refresh the comments, I think we may already have 0 as minimum, and we have use-cases for that for scale-down to 0

mimowo:
```suggestion
    // count is the number of pods requested by the Job.
```

mimowo:
```suggestion
// minCount is the minimum number of pods acceptable for admission
// in case of partial admission.
//
// If not provided, partial admission for the current PodSet is not
// enabled.
// +optional
```

mimowo, line 165: Let's also add description and comment for Workload `.status.admission.admittedPodSets[].count`

keps/420-partial-admission/README.md
yaroslava-serdiuk: I thought the order is the order of the PodSets in the Job. 
For the RayCluster/RayJob/RayService the PodSets order in Workload is HeadPod, Submitter jobs etc, and then WorkingGroups in the same order as they are in the Job. We are interested only in WorkingGroups, because only they could have min size.

keps/420-partial-admission/README.md
yaroslava-serdiuk: I though the order is defined by user, so the order-based should be a good name.

keps/420-partial-admission/README.md
yaroslava-serdiuk: I verified with the user, that they use only 1 working group, so they don't need a specific policy at the moment — anything will work for them. 

So, I agree to leave only order-based algorithm. Removed proportional algorithm.

keps/420-partial-admission/README.md
yaroslava-serdiuk: Added

yaroslava-serdiuk: Done

yaroslava-serdiuk: I think comment for .status.admission.admittedPodSets[].count is up to date, WDYT?

keps/420-partial-admission/README.md
yaroslava-serdiuk: I added it above API section

keps/420-partial-admission/README.md
yaroslava-serdiuk: I added a feature gate. 
However I would not mix approach for ElasticJobs and regular jobs for now. When we can see that approach for ElasticJobs works fine, we can extend it to non-elastic job

yaroslava-serdiuk:
cc: @kevin85421 
Just for your awareness about this feature related to the RayJob/RayCluster

kevin85421:
The partial admission idea is great! There are a lot of multimodal data-processing workloads in Ray, and some of their stages can be embarrassingly parallel. 

Is it possible for users to opt into or out of this feature? For example, we may want to launch an async checkpoint eval after several steps of training. The model may need multiple GPU nodes to serve, so if we partially admit one Pod, that Pod will sit idle.

keps/420-partial-admission/kep.yaml
```diff
@@ -10,6 +11,11 @@ reviewers:
 approvers:
   - "@alculquicondor"
   - "@ahg-g"
+  - "@mimiwo"
```
mimowo:
```suggestion
  - "@mimowo"
```

keps/420-partial-admission/README.md
```diff
@@ -44,68 +56,135 @@ Kueue will not take any measure to ensure that the parent job respects the assig
...
+
+type Admission struct {
+    // .........
+
+    // podSetAssignments hold the admission results for each of the .spec.podSets entries.
+    // +listType=map
+    // +listMapKey=name
+    // +kubebuilder:validation:MaxItems=10
+    PodSetAssignments []PodSetAssignment `json:"podSetAssignments"`
+}
```
mimowo: I think these can be skipped, let's keep only PodSetAssignment definition here

keps/420-partial-admission/README.md
```diff
@@ -128,13 +207,84 @@ type GenericJob interface {
...
+  4. **ElasticJobUngater Controller**: Detects that `wl-A` is admitted and removes the scheduling gate from the 5 pods.
+  5. **Kube-scheduler**: Schedules the 5 ungated pods, which transition to the Running state.
+* **Quota usage**: 5/7 (2 available).
+
+##### Step 1: Scale Up from 5 to 10
+* **Job spec.parallelism**: 10
+* **Workloads**:
+  * `wl-A` (Finished - aggregated/replaced by `wl-B`)
+  * `wl-B` (Admitted - Partially):
+    * `spec.podSets.count` = 7 (originally requested 10, but updated to 7 during partial admission)
```
mimowo: I think this is incorrect given our recent discussion  - the wl .spec.podSet.count stays as in Job, but the

keps/420-partial-admission/README.md
```diff
@@ -128,13 +207,84 @@ type GenericJob interface {
...
+Also, a new Workload representing the full job will be created and added to the queue, to admit the remaining capacity once it becomes available.
+
+#### Example: Two-Step Scale Up under Quota Constraints (with Partial Admission)
+
+Consider a scenario where:
+1. The ClusterQueue has a total quota of **7** for the requested resource flavor.
+2. The Job is configured for both `ElasticJobs` and `PartialAdmission`.
+3. The user performs a two-step scale up of the Job: starting at **5** replicas, scaling up to **10**, and then to **12**.
+
+##### Step 0: Job Creation (Initial Size: 5)
```
mimowo, line 274: Instead of doing the one chained step-ed examples, could you show multiple transition cases rather than steps. 

The relevant examples:
- partial admission succeeded, partial
- scale up of the Job by user, succeeding fully
- scale up of the Job by user, succeeding partially
- opportunistic scale up when capacity is freed

keps/420-partial-admission/README.md
```diff
@@ -44,68 +56,135 @@ Kueue will not take any measure to ensure that the parent job respects the assig
...
 
 ### User Stories
 
 Kueue issue [420](https://github.com/kubernetes-sigs/kueue/issues/420) provides details on the initial feature details and its applicability for `batch/Job`.
 
+Autoscaled RayJob is running and scaled up from 1000 to 5000 pods, but only capacity for 2000 more pod is available. With partial admission, the Kueue admits a 2000 out of 4000 pods to the cluster instead of rejecting the scale up.
+
 ## Design Details
 
+Jobs with `kueue.x-k8s.io/partial-admission=true` annotation are eligible for partial admission.
```
mimowo: add the definition of rhe annotation in the API section

yaroslava-serdiuk:
> The partial admission idea is great! There are a lot of multimodal data-processing workloads in Ray, and some of their stages can be embarrassingly parallel.
> 
> Is it possible for users to opt into or out of this feature? For example, we may want to launch an async checkpoint eval after several steps of training. The model may need multiple GPU nodes to serve, so if we partially admit one Pod, that Pod will sit idle.

@kevin85421 Sure, the PartialAdmission for ElasticJobs could be enabled/disabled cluster wide via feature gate and also individually on each job by adding/removing the `kueue.x-k8s.io/partial-admission=true` annotation.

keps/420-partial-admission/README.md
```diff
@@ -44,68 +56,135 @@ Kueue will not take any measure to ensure that the parent job respects the assig
...
+    // Value could be missing for Workloads created before this field was added,
+    // in that case spec.podSets[*].count value will be used.
+    //
+    // +optional
+    // +kubebuilder:validation:Minimum=0
+    Count *int32 `json:"count,omitempty"`
+}
 ```
 
 ### Scheduler / Flavorassignment
```
mimowo, line 172: On a sync you also described the "opportunistic scale up" - scale up after partial admission when capacity becomes available. We discussed this could be done by the ElasticWorkloads, by sending the new workload probe. 

Please either include design for this in the doc as a dedicated section, or explicitly make out of scope.

keps/420-partial-admission/kep.yaml
```diff
@@ -27,3 +33,4 @@ milestone:
 
 feature-gates:
   - name: PartialAdmission
+  - name: PartialAdmissionForElasticJobs
```
mimowo, line 36: also add: PartialAdmissionForElasticJobsForBatchJob

keps/420-partial-admission/README.md
```diff
@@ -128,13 +207,84 @@ type GenericJob interface {
...
 ### batch/Job controller
 
 Besides adapting `RunWithPodSetsInfo` and `RestorePodSetsInfo` it should also:
 
 - rework `PodSets()` to populate `MinCount` if the job is marked to support partial admission.
-  * jobs supporting partial admission should have a dedicated annotation. eg. `kueue.x-k8s.io/job-min-parallelism`, indicating the minimum `parallelism` acceptable by the job in case of partial admission.
-  * jobs which need the `completions` count kept in sync with `parallelism` should indicate this in a second annotation `kueue.x-k8s.io/job-completions-equal-parallelism`
+  * jobs supporting partial admission should have a dedicated annotation, e.g., `kueue.x-k8s.io/job-min-parallelism`, indicating the minimum `parallelism` acceptable by the job in case of partial admission.
+  * jobs which need the `completions` count kept in sync with `parallelism` should indicate this in a second annotation, `kueue.x-k8s.io/job-completions-equal-parallelism`
 - rework `EquivalentToWorkload` to account for potential differences in `PodSets` spec `Parallelism`.
```
mimowo, line 342: Please also add a paragraph below that the Job will be migrated to the new mechanism under the PartialAdmissionForElasticJobsForBatchJob FG

keps/420-partial-admission/README.md
```diff
@@ -144,30 +294,68 @@ In case of MPIJob `j.Spec.RunPolicy.SchedulingPolicy.MinAvailable` can be used t
 Whether an MPIJob supports partial admission or not can be deduced based on `MinAvailable` without the need of a dedicated annotation.
 Additional research is needed into the potential usage of multiple variable count PodSets.
 
+### RayJob/RayService/RayCluster controller
+
+The RayCluster.workerGroupSpec[i].minReplicas will be translated to the PodSet.MinCount for the Worker PodSet.
```
mimowo: I think this is inaccurate., because we should probably use `minReplicas * numOfHosts`? 

Analogously as we do for count, ptal: https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/jobs/raycluster/common.go#L69-L72

And please also describe the count so that the KEP also details for "count".

keps/420-partial-admission/README.md
```diff
@@ -144,30 +294,68 @@ In case of MPIJob `j.Spec.RunPolicy.SchedulingPolicy.MinAvailable` can be used t
...
+    - `invalid podSet minCount (negative)`: verifies negative minCount values are rejected.
+    - `invalid podSet minCount (too big)`: verifies minCount larger than count is rejected.
+    - `too many variable count podSets`: verifies that workloads with multiple variable count PodSets are rejected.
+
+#### E2E tests
+
+- `test/e2e/singlecluster/baseline/job_test.go`:
+  - `Should partially admit the Job if configured and not fully fits`: verifies that a real Job is successfully admitted with a reduced parallelism count matching the available cluster resources.
 
 ### Graduation Criteria
```
mimowo, line 398: Please fill in the gradution criteria for: PartialAdmissionForElasticJobs and PartialAdmissionForElasticJobsForBatchJob FGs, divide them in sections for Alpha, Beta and Stable.

keps/420-partial-admission/README.md
```diff
@@ -44,68 +56,135 @@ Kueue will not take any measure to ensure that the parent job respects the assig
...
 
-Change the way the flavor assigner work to support decrementing the pods count in order to find a better fit for the current workload.
+Change the way the flavor assigner works to support decrementing the pods count in order to find a better fit for the current workload.
 In case a partial fit is chosen, the jobframework reconciler should provide the admitted pod counts to the parent job before unsuspending it, in a similar fashion as the node selectors are provided. In case the job gets suspended, the original pod counts should be restored in order to allow a potential future admission with its original pod counts.
 
 ### User Stories
 
 Kueue issue [420](https://github.com/kubernetes-sigs/kueue/issues/420) provides details on the initial feature details and its applicability for `batch/Job`.
 
+Autoscaled RayJob is running and scaled up from 1000 to 5000 pods, but only capacity for 2000 more pod is available. With partial admission, the Kueue admits a 2000 out of 4000 pods to the cluster instead of rejecting the scale up.
```
mimowo: Please add structure to the section like 

Story 1 (partial admission)

/  as before + KubeRay

Story 2 (graceful scale up handling)

// as presented currently

Story 3 (opportunistic scale up)

// write a story if this scenario matters as we discussed on a sync

Story 4 (multi-podset RayJob)

// As a user of multi-podset RayJob .. .

keps/420-partial-admission/README.md
```diff
@@ -44,68 +56,135 @@ Kueue will not take any measure to ensure that the parent job respects the assig
...
+
+    // count is the number of pods taken into account at admission time.
+    // This field will not change in case of quota reclaim.
+    // Value could be missing for Workloads created before this field was added,
+    // in that case spec.podSets[*].count value will be used.
+    //
+    // +optional
+    // +kubebuilder:validation:Minimum=0
+    Count *int32 `json:"count,omitempty"`
+}
```
mimowo, line 164: Please add a section on the "Validation" for the API fields. Some ideas:

minCount <= count
minCount <= admission.count <= count

wdyt?

keps/420-partial-admission/README.md
```diff
@@ -44,68 +56,135 @@ Kueue will not take any measure to ensure that the parent job respects the assig
...
-If multiple PodSets within a workload have variable counts the way the counts should be decrease is highly dependent on the job framework needs, currently the following is proposed:
+- **`order-based` (default)**: Shrinks the counts of the PodSets sequentially starting from the last one (suits for the cases when the podsets are ordered by priority). The Workload PodSet order is usually the same as the order of the PodSet in the Job spec. For the RayCluster the Workload PodSets starting from Head PodSet and following by WorkerGroupPodSets in the same order as in RayCluster.
 
-Starting with the PodSets:
+#### Order-Based policy (`order-based`)
 
-```go
-[]podSet { {Count: 1} , {Count: 4, MinCount: 2}, {Count: 20, MinCount: 10}}
-```
+Under the `order-based` policy, Kueue shrinks the PodSets starting from the last one and moving towards the beginning as needed.
```
mimowo: what determines the order of PodSets being considered in this strategy? Names or declaration order in the array list?

keps/420-partial-admission/kep.yaml
yaroslava-serdiuk: Why? How the BatchJob would differ from the approach to ElasticJobs?

keps/420-partial-admission/README.md
yaroslava-serdiuk: >the Job will be migrated to the new mechanism

Actually I don't think any migration is needed. In order to apply elastic approach, the workload should have elastic-job annotation. If the job has elastic-job annotation, during the partial admission the elastic path will be chosen. 
If the workload doesn't have elastic-job annotation, kueue won't apply workload-slices mechanism.

keps/420-partial-admission/kep.yaml
mimowo: because we need to have the guard to deccomission the old code

yaroslava-serdiuk: Do you mean proportional approach?

keps/420-partial-admission/README.md
mimowo: Basically my point is that end the end of the day we don't want to maintain two mechanisms for partial admission. 

This needs to be clarified in the KEP. 

what happens on RayJob and Job when user sets:
- partial-admission=true & elastic-job-true
- partial-admission=true only
- elastic-job=true only
I was thinking that you want to make `partial-admission=true` imply `partial-admission=true` & `elastic-job-true`, but I'm no longer sure. 

Please clarify the semantics for the combination of the annotations. My only request is to align the semantics for all CRDs.

keps/420-partial-admission/kep.yaml
mimowo: There is no proportional approach for batch Job, because there is only one PodSet.

I meant to decommission the code which modifies Job's parallelism

keps/420-partial-admission/README.md
yaroslava-serdiuk: I think approaches of partial admission for elastic workloads and non-elastic workloads are different and should not be mixed.

mimowo: Sorry, I'm very confused. I would like to request a summary in the KEP the behavior in the three scenarios as I listed above:
1. `partial-admission=true` & `elastic-job-true`
2. `partial-admission=true` only
3. `elastic-job=true`

Then multiply this by Job and KubeRay. 

As a maintainer of Kueue I would like to understand if we need to maintain separate handling for Job and KubeRay. So far I cannot understand the argument.

keps/420-partial-admission/README.md
yaroslava-serdiuk: I added a name for each step, so it should be linked transition cases. Is that what you meant?

keps/420-partial-admission/README.md
yaroslava-serdiuk: declaration order in the array list

keps/420-partial-admission/README.md
yaroslava-serdiuk: added the section in design details of elasticjobs

keps/420-partial-admission/README.md
yaroslava-serdiuk: I added a section of enabling, seems annotation suits there better, wdyt?

mimowo:
cc @ivnovakov @mbobrovskyi ptal
I would like to engage the team early at the KEP phase  so that we minimize the chances of surprises during the implementation.

keps/420-partial-admission/kep.yaml
yaroslava-serdiuk: This is a different effort, I created a follow up for it https://github.com/kubernetes-sigs/kueue/issues/13962

keps/420-partial-admission/README.md
yaroslava-serdiuk: Added graduation criteria for PartialAdmissionForElasticJob

keps/420-partial-admission/README.md
yaroslava-serdiuk: Thanks, updated

keps/420-partial-admission/README.md
yaroslava-serdiuk: Added

keps/420-partial-admission/README.md
yaroslava-serdiuk: Added

keps/420-partial-admission/README.md
yaroslava-serdiuk: I think validation for minCount is the Job responsibility, we can just have an partial admission error

mimowo: we probably should also have a validation at the workload object to prevent malicious settings or just incorrect settings. I think the mechanism we do coukd also be used for custom job crd imementations in general.

Finally defining invartiants on the workload crd helps other developers to understand the code

yaroslava-serdiuk: Added validation

keps/420-partial-admission/README.md
```diff
@@ -128,13 +249,89 @@ type GenericJob interface {
...
+### ElasticJob
+
+For ElasticJobs, updating `job.spec.parallelism` or `job.spec.count` could cause race conditions between partial admission and scaling up/down activity. 
+To avoid this, the `podSets` count won't be updated in `RunWithPodSetsInfo` for elastic jobs. Instead, the workload controller will use the `workload.Status.Admission.PodSetAssignments[*].Count` value to calculate the number of pods from which Kueue should remove scheduling gates. 
+The `minCount` value for an ElasticJob workload will represent the currently admitted value + 1.
+Also, a new Workload representing the full job will be created and added to the queue, to admit the remaining capacity once it becomes available.
+
+#### Opportunistic scale up when capacity is freed
+
+In order to schedule remaining pods after partial admission, workload controller will create a new workload representing the full job and add it to the queue. The scheduler will admit new workload and replace the old workload via workload slice mechanism as the capacity becomes available.
```
mimowo, line 264: This sounds good. One detail is that we need to somehow generate the name for the workload probe that would not collide with the admitted workload. This is because currently ElasticJobs generate workload names based on the "base" of the name, and the Job's generation. However, both will be the same for the admitted workload, and the opportunistic probe. And I'm not totally sure how to solve it, maybe we could mix-in into the hash the "probe" type. 

I think we should at least flag the collision problem. I'm ok to say, that we propose using "opportunistic probe" to mix-into the hash, but details may be determined during implementation.

keps/420-partial-admission/README.md
```diff
@@ -128,13 +249,89 @@ type GenericJob interface {
 
 ```
 
+### ElasticJob
+
+For ElasticJobs, updating `job.spec.parallelism` or `job.spec.count` could cause race conditions between partial admission and scaling up/down activity.
```
mimowo: `job.spec.count` - do you mean here Workload's `.spec.podSet[].count`? I'm not sure

keps/420-partial-admission/README.md
```diff
@@ -128,13 +249,89 @@ type GenericJob interface {
 
 ```
 
+### ElasticJob
+
+For ElasticJobs, updating `job.spec.parallelism` or `job.spec.count` could cause race conditions between partial admission and scaling up/down activity. 
+To avoid this, the `podSets` count won't be updated in `RunWithPodSetsInfo` for elastic jobs. Instead, the workload controller will use the `workload.Status.Admission.PodSetAssignments[*].Count` value to calculate the number of pods from which Kueue should remove scheduling gates. 
+The `minCount` value for an ElasticJob workload will represent the currently admitted value + 1.
```
mimowo: I'm confused here, two questions:
1. by `minCount` do you mean Workload's `.spec.podSets[].minCount`? If so, please clarify
2. `minCount` is `admitted value+1`? 

This sounds counter-intuitive. I would assume `.spec.podSets[].count` `.spec.podSets[].minCount` are derived from the Job specification. Admission happens later, and admission is anywhere between the two values.

keps/420-partial-admission/README.md
```diff
@@ -128,13 +249,89 @@ type GenericJob interface {
...
+
+#### Example: Two-Step Scale Up under Quota Constraints (with Partial Admission)
+
+Consider a scenario where:
+1. The ClusterQueue has a total quota of **7** for the requested resource flavor.
+2. The Job is configured for both `ElasticJobs` and `PartialAdmission`.
+3. The user performs a two-step scale up of the Job: starting at **5** replicas, scaling up to **10**, and then to **12**.
+
+##### Step 0: Job Creation (Initial Size: 5)
+* **Job spec.parallelism**: 5
```
mimowo, line 275: I think this example should also define for the Job the min-count annotation, otherwise it is unclear where `spec.podSets.minCount` is coming from

keps/420-partial-admission/README.md
```diff
@@ -128,13 +249,89 @@ type GenericJob interface {
...
+In order to schedule remaining pods after partial admission, workload controller will create a new workload representing the full job and add it to the queue. The scheduler will admit new workload and replace the old workload via workload slice mechanism as the capacity becomes available.
+
+#### Example: Two-Step Scale Up under Quota Constraints (with Partial Admission)
+
+Consider a scenario where:
+1. The ClusterQueue has a total quota of **7** for the requested resource flavor.
+2. The Job is configured for both `ElasticJobs` and `PartialAdmission`.
+3. The user performs a two-step scale up of the Job: starting at **5** replicas, scaling up to **10**, and then to **12**.
+
+##### Step 0: Job Creation (Initial Size: 5)
```
mimowo, line 274: Instead of "Step" please make there "Scenario" so that they can be analyzed independently

keps/420-partial-admission/README.md
```diff
@@ -144,30 +341,81 @@ In case of MPIJob `j.Spec.RunPolicy.SchedulingPolicy.MinAvailable` can be used t
 Whether an MPIJob supports partial admission or not can be deduced based on `MinAvailable` without the need of a dedicated annotation.
 Additional research is needed into the potential usage of multiple variable count PodSets.
```
mimowo: I think we can update the KEP and drop this section. This is not done. Let's re-design when we do this

keps/420-partial-admission/README.md
```diff
@@ -44,68 +68,165 @@ Kueue will not take any measure to ensure that the parent job respects the assig
...
+| **KubeRay** | `true` | `true` | **Partial Admission + Workload Slicing**: Kueue can shrink the worker groups down to their `minReplicas` (defined in the worker spec). The Ray object's spec remains unmodified. Admission is managed via scheduling gates. A pending workload slice is created to allow KubeRay to scale up (opportunistic scale up) worker pods as more quota becomes available. |
+| **KubeRay** | `true` | `false`/not present | **Not supported / Out of scope**: Kueue does not support partial admission for non-autoscaled KubeRay jobs at this time. |
+| **KubeRay** | `false`/not present | `true` | **Workload Slicing Only**: Kueue cannot shrink the worker groups; they must fit their full requested replicas to be admitted. Once admitted, scheduling gates are used. If the autoscaler scales up the worker groups, Kueue creates a workload slice representing the scale-up request, but this scale-up is only admitted if there is enough quota to satisfy the full scale-up request. |
+
 ### Workload API
 
 ```go
+// +kubebuilder:validation:XValidation:rule="has(self.minCount) ? self.minCount <= self.count : true", message="minCount should be positive and less or equal to count"
```
mimowo, line 137: Insteead of the kubebuilder, please define a section with the validation constraints. I think this is actually more involving that that.

We don't need to have a rule which compiples exactly, what is more important to me is general direction:
- `.spec.podSets.minCount <= .spec.podSets.count`
- `.spec.podSets.minCount <= .status.admission.podSets.count <= .spec.podSets.count`
- decide if minCount= 0 is supported (I think so, but we may state that explicitly)

You may also mention that the exact arythmentics is impacted by the reclamablePods.

cc @amy who is now an expect in reclamablePods :)

keps/420-partial-admission/README.md
```diff
@@ -144,30 +341,81 @@ In case of MPIJob `j.Spec.RunPolicy.SchedulingPolicy.MinAvailable` can be used t
...
+No regressions in the current tests should be observed.
 
 [X] I/we understand the owners of the involved components may require updates to
 existing tests to make this code solid enough prior to committing the changes necessary
 to implement this enhancement.
 
 #### Unit Tests
 
-The changes in the flavorassignment should be covered by unit tests
+- **Scheduler/Flavor Assignment**:
```
mimowo, line 362: I think the test plan needs refreshing, because it looks stale, for example the proporitonal mode is still here,

keps/420-partial-admission/README.md
```diff
@@ -128,13 +249,89 @@ type GenericJob interface {
...
+
 ### batch/Job controller
 
 Besides adapting `RunWithPodSetsInfo` and `RestorePodSetsInfo` it should also:
 
 - rework `PodSets()` to populate `MinCount` if the job is marked to support partial admission.
-  * jobs supporting partial admission should have a dedicated annotation. eg. `kueue.x-k8s.io/job-min-parallelism`, indicating the minimum `parallelism` acceptable by the job in case of partial admission.
-  * jobs which need the `completions` count kept in sync with `parallelism` should indicate this in a second annotation `kueue.x-k8s.io/job-completions-equal-parallelism`
+  * jobs supporting partial admission should have a dedicated annotation, e.g., `kueue.x-k8s.io/job-min-parallelism`, indicating the minimum `parallelism` acceptable by the job in case of partial admission.
+  * jobs which need the `completions` count kept in sync with `parallelism` should indicate this in a second annotation, `kueue.x-k8s.io/job-completions-equal-parallelism`
```
mimowo, line 341: I think this will depend if we enable the elastic-jobs or not, right?
because it we enable the I assume workload's 
```
.spec.podSets.minCount=kueue.x-k8s.io/job-min-parallelism
.spec.podSets.count= Job's spec.parallelism
```

keps/420-partial-admission/README.md
```diff
@@ -44,68 +68,165 @@ Kueue will not take any measure to ensure that the parent job respects the assig
...
+The search for appopriate count value is done using binary search algorithm.
 
-In order to evaluate the potential success of the preemption, the preemption process should be split in:
-- Target selection, which should select the workloads within cohort that will need to be evicted, if the list is empty the assignment will be treated as `NoFit`. This step will take place during nomination.
-- Preemption issuing, will do the actual eviction of the targets previously selected.
+#### Partial Admission for multiple PodSets
 
-**Partial admission with multiple variable PodSets count**
+The are multiple ways how to approach multiple podsets shrinking in case of insufficient quota. For simplisity reason we'll start with the order-based one and will expand option if needed in future.
+Another approch that was considered is proportional. However it was disgarded because of insufficiency for some use cases.
```
mimowo: Move the mention of this to alternatives section

keps/420-partial-admission/README.md
```diff
@@ -44,68 +68,165 @@ Kueue will not take any measure to ensure that the parent job respects the assig
...
-The accepted number of pods in each PodSet are recorded in `workload.Status.Admission.PodSetAssignments[*].ResourceUsage.Count`
+The search for appopriate count value is done using binary search algorithm.
 
-In order to evaluate the potential success of the preemption, the preemption process should be split in:
-- Target selection, which should select the workloads within cohort that will need to be evicted, if the list is empty the assignment will be treated as `NoFit`. This step will take place during nomination.
-- Preemption issuing, will do the actual eviction of the targets previously selected.
+#### Partial Admission for multiple PodSets
 
-**Partial admission with multiple variable PodSets count**
+The are multiple ways how to approach multiple podsets shrinking in case of insufficient quota. For simplisity reason we'll start with the order-based one and will expand option if needed in future.
```
mimowo: simplisity is a typo

keps/420-partial-admission/README.md
```diff
@@ -44,68 +68,165 @@ Kueue will not take any measure to ensure that the parent job respects the assig
...
 ### Scheduler / Flavorassignment
 
-In case the workload proposed for the current scheduling cycle, does not fit, with or without preemption, in the current available quota and any of its PodSets allow partial admission, try to find to find a lower counts combination that fits the available quota with or without borrowing.
+In case the workload proposed for the current scheduling cycle does not fit, with or without preemption, in the current available quota and any of its PodSets allow partial admission, kueue will try to find a lower counts combination that fits the available quota with or without borrowing.
 
-The search should be optimized (binary search) and preserve the proportion of pods lost across the variable count PodSets.
+#### Partial Admission for one PodSets
 
-The accepted number of pods in each PodSet are recorded in `workload.Status.Admission.PodSetAssignments[*].ResourceUsage.Count`
+The search for appopriate count value is done using binary search algorithm.
```
mimowo: appopriate - is a typo, please check the KEP with spellchecker

keps/420-partial-admission/README.md
```diff
@@ -44,68 +68,165 @@ Kueue will not take any measure to ensure that the parent job respects the assig
...
+}
 ```
 
 ### Scheduler / Flavorassignment
 
-In case the workload proposed for the current scheduling cycle, does not fit, with or without preemption, in the current available quota and any of its PodSets allow partial admission, try to find to find a lower counts combination that fits the available quota with or without borrowing.
+In case the workload proposed for the current scheduling cycle does not fit, with or without preemption, in the current available quota and any of its PodSets allow partial admission, kueue will try to find a lower counts combination that fits the available quota with or without borrowing.
 
-The search should be optimized (binary search) and preserve the proportion of pods lost across the variable count PodSets.
+#### Partial Admission for one PodSets
```
mimowo: "one PodSets" - something is wrong here

keps/420-partial-admission/README.md
mimowo: "disgarded" - something is wrong, please check with spellcker

keps/420-partial-admission/README.md
```diff
@@ -44,68 +68,165 @@ Kueue will not take any measure to ensure that the parent job respects the assig
...
+
+#### Story 3 (opportunistic scale up)
+
+As a user running an elastic job (like a RayJob), when my job is admitted with partial capacity due to resource constraints, I want the job to dynamically scale up to its full requested capacity as soon as other workloads complete and resources become available in the cluster, maximizing resource utilization and reducing the job's overall completion time.
+
+#### Story 4 (multi-podset RayJob)
+
+As a user of a multi-podset RayJob (which defines a head pod and multiple worker groups, potentially targeting different resource flavors or node groups like reservation, on-demand, or spot), I want to enable partial admission such that Kueue reduces the worker groups sequentially starting from the least critical (e.g., spot or low-priority worker groups defined last in the spec) while preserving the capacity of the more critical worker groups.
+
 ## Design Details
```
mimowo, line 93: I think there is also a problem (even currently) with ElasticJobs for KubeRay, but for partial admission in particular. 

Because PodsReady requires all pods to be ready, but we should only expect that from the number of admitted pods, ie Workloads `.status.admission.podSets.count` - I think we should actually open an issue for this even today. Would you like to drive fixing this since this is needed for KubeRay elastic jobs anyway?

keps/420-partial-admission/README.md
```diff
@@ -44,68 +68,165 @@ Kueue will not take any measure to ensure that the parent job respects the assig
...
+  // EnabledAnnotationKey refers to the annotation key present on Job's that support
+  // workload slicing.
+  // This annotation is alpha-level.
+  EnabledPartialAdmission = "kueue.x-k8s.io/partial-admission"
+)
+```
+
+#### Scenario Summary
+
+The table below summarizes the scheduling and admission behaviors for different combinations of the `kueue.x-k8s.io/partial-admission` and `kueue.x-k8s.io/elastic-job` annotations for `batch/v1.Job` and KubeRay (`RayJob` / `RayCluster`) objects:
```
mimowo, line 123: Thank you for the table, this clarifies a lot.

keps/420-partial-admission/kep.yaml
```diff
@@ -2,6 +2,7 @@ title: Allow partial admission of jobs
 kep-number: 420
 authors:
   - "@trasc"
+  - "@yaroslava-serdiuk"
```
mimowo, line 5: cc @sohankunkerkar @amy would you find cycles to give it a pass?

keps/420-partial-admission/README.md
yaroslava-serdiuk: Sorry for confusion, I updated the section to be more accurate. 

The Workload's minCount must diverge from the original minCount for already admitted Jobs to ensure we don't admit fewer pods than are currently admitted.

keps/420-partial-admission/README.md
yaroslava-serdiuk: >they can be analyzed independently

They are connected since the previous state was described in previous step (except step 0), so it make sense to process step 0 (previous state) + step 1 (current state).

keps/420-partial-admission/README.md
yaroslava-serdiuk: removed proportional

mimowo:
cc @olekzabl who is considering "opportunistic scale-up mechanism in the context of WAS integration"
cc @ivnovakov as a tentative owner of the implementation

keps/420-partial-admission/README.md
yaroslava-serdiuk:
```
func (j *RayCluster) PodsReady(ctx context.Context, _ client.Client) bool {
	return j.Status.State == rayv1.Ready
}
```

The implementation doesn't rely on podSets.count

mimowo: Can you investigate if this is correct for partial admission?  I'm wondering if partially admitted RayCluster will have Ready=True. If this is the case, then I suppose it is ok.

keps/420-partial-admission/README.md
yaroslava-serdiuk: >jobs which need the `completions` count kept in sync with `parallelism` should indicate this in a second annotation, `kueue.x-k8s.io/job-completions-equal-parallelism`

But this is unrelated to elastic-job. Not sure I understand your comment, can you clarify?

keps/420-partial-admission/README.md
yaroslava-serdiuk: Good point, updated.

keps/420-partial-admission/README.md
yaroslava-serdiuk: >I'm wondering if partially admitted RayCluster will have Ready=True.

RayCluster will have Ready state only if all pods are running. However the state will remain Ready when the RayCluster is scaled up but not all newly created pods are running. 

So I suggest to add it as limitation for now. I.e. when the job is created, it could not be partially admitted, the kueue will partially admit only scale up. We can also recommend to user to set count=minCount, because the main user request was regarding scale ups.

yaroslava-serdiuk: For RayCluster when constructing the initial (first) the minCount will equal replicas and then minCount for following Workloads (scaleups) will equal to admitted count + 1.

keps/420-partial-admission/README.md
yaroslava-serdiuk: Added validation for `.spec.podSets.minCount >= 0` and `.spec.podSets.minCount <= .spec.podSets.count.`
I think this validation should happen upon creation. 

For `.spec.podSets.minCount <= .status.admission.podSets.count <= .spec.podSets.count` i don't see when we should validate it. This is a part of scheduling

keps/420-partial-admission/README.md
```diff
@@ -44,68 +67,169 @@ Kueue will not take any measure to ensure that the parent job respects the assig
...
+| **batch/Job** | `true` | `false`/not present | **Partial Admission (Traditional)**: Kueue shrinks the Job's podset down to its `job-min-parallelism`. Upon admission, Kueue modifies the Job's spec (`.spec.parallelism`) directly to match the admitted count. No scheduling gates are used, and no future scale-up occurs for this workload. |
+| **batch/Job** | `false`/not present | `true` | **Workload Slicing Only**: Kueue cannot shrink the Job's podset during initial scheduling; the job must fit its full requested capacity to be admitted. If admitted, scheduling gates are used. If the user manually scales up the Job, Kueue creates a new workload slice for the scale-up request, but this scale-up will only be admitted if it can be fully satisfied (no partial admission of the scale-up). |
+| **KubeRay** | `true` | `true` | **Partial Admission + Workload Slicing**: Kueue can shrink the worker groups down to their `minReplicas` (defined in the worker spec). The Ray object's spec remains unmodified. Admission is managed via scheduling gates. A pending workload slice is created to allow KubeRay to scale up (opportunistic scale up) worker pods as more quota becomes available. |
```
vladikkuzn, line 130: Does it affect head pods in any way?

yaroslava-serdiuk: Head pod doesn't have minReplicas, so no

keps/420-partial-admission/README.md
kshalot: I did a preliminary test on a RayCluster:
1. During a scale-up, the RayCluster is ready even if the new pods are not ready. I did this by bumping `replicas` for the cluster and adding a scheduling gate to the new pod.
2. During initial admission, if not all workers are ready, the RayCluster will not be ready (and it will have the `RayClusterProvisioned: False` Condition). I did this by using the `podAntiAffinity` on the workers, so one of them gets unschedulable during init.

I'll confirm both further, but what 2. suggests to me is that partial admission might be broken. Because an initial config that explicitly sets `replicas` to something greater than `minReplicas` is a valid configuration, no?

So in the case where we set `minReplicas: 1, replicas: 4`, Kueue will set `MinCount=1` and might do a partial admission of 1-3 workers. But according to finding 2. this RayCluster will not become ready until the 4th pod is ready.

> We can also recommend to user to set count=minCount

Hm, but wouldn't this mean that there really is no partial admission? Because if `count=minCount`, then this is just regular admission for RayCluster, no? We can't admit it partially, we can only scale it up.

yaroslava-serdiuk: > We can't admit it partially, we can only scale it up.

True. However the main user request was regarding partial scale up. So I'm thinking now, that we should not mix partial scale up and partial admission despite both are using the one mechanism.

kshalot: > However the main user request was regarding partial scale up.

Which to me sounds more like a feature for Elastic Jobs, a second mode of operation to differentiate the all-or-nothing scale up that is implemented currently.

In general, to me it seems that partial admission could be treated under the Elastic Job feature, conceptually. But that is not a conversation for right now.

yaroslava-serdiuk: I created a separate kep for partial scaleup: https://github.com/kubernetes-sigs/kueue/pull/14310

mimowo: I would prefer to consolidate this all in one PR. Seems like the topics are too inter-connected to spread in multiple PRs imo.

yaroslava-serdiuk: Sure. I'm closing this PR

yaroslava-serdiuk:
After discussion https://github.com/kubernetes-sigs/kueue/pull/12281#discussion_r3766288636 the changes will be introduces under the new feature PartialScaleUp https://github.com/kubernetes-sigs/kueue/pull/14310 instead of extending the functionality of PartialAdmission

Issue #12160, opened 2026-06-12
Release v0.18.1


mimowo:
## Release Checklist

- [x] [OWNERS](https://github.com/kubernetes-sigs/kueue/blob/main/OWNERS) must LGTM the release proposal.
  At least two for minor or major releases. At least one for a patch release.
- [x] Verify that the changelog in this issue and the CHANGELOG folder is up-to-date
  - [x] Run `./hack/releasing/sync-notes.sh $VERSION` to generate and publish the release notes
- [ ] For major or minor releases (v$MAJ.$MIN.0), create a new release branch.
  - [ ] An OWNER creates a vanilla release branch with
        `git branch release-$MAJ.$MIN main`
  - [ ] An OWNER pushes the new release branch with
        `git push upstream release-$MAJ.$MIN`
- [x] Update the release branch:
  - [x] Ensure there are no unstaged changes in your directory (the script adds everything)
  - [x] Run `./hack/releasing/prepare_pull.sh --target release $VERSION`
  - [x] Wait for this PR to merge #12195 
- [x] An OWNER creates a signed tag
  - [x] pull the release branch after PR from previous step merged
  - [x] run
     `git tag -s $VERSION`
      and inserts the changelog into the tag description.
      To perform this step, you need [a PGP key registered on github](https://docs.github.com/en/authentication/managing-commit-signature-verification/checking-for-existing-gpg-keys).
- [x] An OWNER pushes the tag with
      `git push upstream $VERSION`
  - Triggers prow to build and publish a staging container image
      `us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue:$VERSION`
- [x] An OWNER [prepares a draft release](https://github.com/kubernetes-sigs/kueue/releases)
  - [x] Create the draft release pointing out to the created tag.
  - [x] Write the change log into the draft release.
  - [x] Run
      `make release-artifacts IMAGE_REGISTRY=registry.k8s.io/kueue GIT_TAG=$VERSION`
      to generate the artifacts in the `release-artifacts` folder.
  - [x] Upload the files in the `release-artifacts` folder to the draft release - either
      via UI or `gh release --repo kubernetes-sigs/kueue upload $VERSION release-artifacts/*`.
- [x] Promote images and Helm Charts to production:
  - [x] Run `./hack/releasing/wait_for_images.sh $VERSION` to await for the staging images.
  - [x] Run `./hack/releasing/promote_pull.sh $VERSION` to submit the promotion PR
[...]

PR #12287, opened 2026-06-16
Generate SLSA Attestations with new releases

### LINKED ISSUE #1466: Generate SLSA Attestations with new releases
SLSA Attestation Generated with new releases. 

**Why is this needed**:

SLSA's are resources that show evidence that the release consumers receive has not been tampered with during the supply chain process.

**Completion requirements**:

Implementation of a tool such as https://github.com/kubernetes-sigs/tejolote into the CI process for builds. This will generate the SLSA and attach it to the release.

- [ ] Design doc
- [ ] API change
- [ ] Docs update

ctmphuongg:
#### What type of PR is this?
/kind feature

Adds a SLSA build provenance attestation workflow that downloads release artifacts and generates a signed attestation stored in GitHub's attestation API, verifiable with gh attestation verify.

Fixes #1466 

The workflow is triggered manually rather than automatically on release, because I don't see any CI build workflow in this repo. Release artifacts appear to be built and uploaded outside of GitHub Actions. This can be updated if that changes.

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -244,6 +244,20 @@ image-build:
 		$(IMAGE_BUILD_EXTRA_OPTS) \
 		./
 
+.PHONY: image-pushing
+image-pushing:
+ifeq ($(JOB_TYPE),periodic)
+	$(MAKE) -j3 image-pushing-helper-artifacts
+else
+	$(MAKE) -j5 image-pushing-release-artifacts
+endif
+
+.PHONY: image-pushing-helper-artifacts
+image-pushing-helper-artifacts: debug-image-push importer-image-push ray-project-mini-image-build-push
+
+.PHONY: image-pushing-release-artifacts
+image-pushing-release-artifacts: image-push helm-chart-push kueueviz-image-push kueue-populator-image-push kueue-priority-booster-image-push
+
 .PHONY: image-push
 image-push: PUSH=--push
 image-push: image-build
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -7,14 +7,7 @@ steps:
   - name: 'gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20260127-c1affcc8de'
     entrypoint: make
     args:
-      - image-push
-      - debug-image-push
-      - importer-image-push
-      - helm-chart-push
-      - kueueviz-image-push
-      - kueue-populator-image-push
-      - kueue-priority-booster-image-push
-      - ray-project-mini-image-build-push
+      - image-pushing
     env:
       - DOCKER_BUILDX_CMD=/buildx-entrypoint
       - GOTOOLCHAIN=auto


mimowo:
cc @tenzen-y @kannon92 @amy

mimowo:
> The workflow is triggered manually rather than automatically on release, because I don't see any CI build workflow in this repo. Release artifacts appear to be built and uploaded outside of GitHub Actions. This can be updated if that changes.

Yeah, our releases are semi-manual currently, but the steps are scripted here: https://github.com/kubernetes-sigs/kueue/blob/main/.github/ISSUE_TEMPLATE/NEW_RELEASE.md

This is an example release issue https://github.com/kubernetes-sigs/kueue/issues/12160. It would be good to update the associated release script, otherwise we will forget for sure.

kannon92:
I'm not sure github actions are really going to work well for us.

Our release artifacts are images that we publish to registry.k8s.io.

ctmphuongg:
Ah I see. The current workflow works for attesting the binary artifacts uploaded to GitHub releases via gh release. However, if the primary artifact to attest is the container image on registry.k8s.io, I can try a different approach with  https://github.com/kubernetes-sigs/tejolote but that probably requires modifying the prow build process in another repo (https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/k8s-staging-kueue.yaml ?) Is it okay to proceed that way or would that be out of scope for this issue?

kannon92:
I sent a message to @saschagrunert who mentioned that for sig-release they use goreleaser for some of this.

mimowo:
> I can try a different approach with https://github.com/kubernetes-sigs/tejolote but that probably requires modifying the prow build process in another repo (https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/k8s-staging-kueue.yaml ?) Is it okay to proceed that way or would that be out of scope for this issue?

another potential injection point you may consider is k/k8s.io repo responsible for promotion of release artifacts from staging to k8s registry, wdyt?

if we keep this out of scope them im not clear what remains in the scope of the issue. i think we basically need to atest the images

kannon92:
> if we keep this out of scope them im not clear what remains in the scope of the issue. i think we basically need to atest the images

I think the artifacts on github release would be SLSA. But yes I am not entirely sure the path for images.

saschagrunert:
The primary artifacts to attest here are the container images on `registry.k8s.io`, not the GitHub release binaries. The official Kubernetes image promotion tooling ([promo-tools/kpromo](https://github.com/kubernetes-sigs/promo-tools)) already handles signing and SLSA provenance attestation during promotion (staging -> registry.k8s.io), so that side is covered.

The gap is **build-time provenance** on the staging images. A couple of options:

1. **goreleaser**: what sig-release uses for artifact builds; has built-in SLSA and cosign support. Could replace the current GCB `make image-push` approach.
2. **cosign attest** added as a step in Kueue's `cloudbuild.yaml` to attest images right after they're pushed to staging.
3. **tejolote** (kubernetes-sigs): designed to wrap Kubernetes CI builds with SLSA provenance generation.

Once build-time attestations exist on the staging images, kpromo will automatically verify them during promotion (phase 3, verify-if-present semantics) before signing and attesting the promoted images.

So the recommended path would be: add provenance generation to the existing GCB build pipeline (option 1 or 2), and rely on kpromo for the promotion-side attestation. No changes needed in test-infra or k8s.io for that.

ctmphuongg:
Thanks! I also found out some of these facts when looking at these approaches, but your comment gives much more context. I've just drafted a PR to handle the build-time image attestations here: #12313.

PR #12292, opened 2026-06-16
Fix ParseTLSOptions dropping earlier errors in errors.Join

### LINKED ISSUE #10698: Add TLSCurves support to Kueue

kannon92:
/kind bug

Fixes a bug in `ParseTLSOptions` where `errors.Join(err)` was used instead of `errors.Join(errRet, err)`. This silently dropped any previously accumulated errors. For example, if both the TLS min version and cipher suites were invalid, only the cipher suite error was returned — the TLS version error was lost.

The test case name is updated to reflect it now validates both errors are returned.

This was split out from #11832 per reviewer feedback.

Fixes #10698 (partially)

AI was used to help create this PR.

```release-note
ConfigAPI: Fixed TLS configuration validation to report all detected option errors instead of only the last one.
```

## Bug Fixes
- Fixed TLS options validation to consistently return all relevant errors when multiple inputs are invalid (for example, when both the minimum TLS version and cipher suites are wrong).

## Tests
- Enhanced TLS config tests to validate multi-error outputs by checking that every expected error is present, updating table-driven cases accordingly.

### DIFF
--- a/pkg/util/tlsconfig/tlsconfig.go
+++ b/pkg/util/tlsconfig/tlsconfig.go
@@ -42,7 +42,7 @@ func ParseTLSOptions(cfg *config.TLSOptions) (*TLS, error) {
 	}
 	version, err := convertTLSMinVersion(cfg.MinVersion)
 	if err != nil {
-		errRet = errors.Join(err)
+		errRet = err
 	}
 	ret.MinVersion = version
 
@@ -50,7 +50,7 @@ func ParseTLSOptions(cfg *config.TLSOptions) (*TLS, error) {
 	if len(cfg.CipherSuites) > 0 {
 		cipherSuites, err := convertCipherSuites(cfg.CipherSuites)
 		if err != nil {
-			errRet = errors.Join(err)
+			errRet = errors.Join(errRet, err)
 		}
 		if err == nil && len(cipherSuites) > 0 {
 			ret.CipherSuites = cipherSuites
--- a/pkg/util/tlsconfig/tlsconfig_test.go
+++ b/pkg/util/tlsconfig/tlsconfig_test.go
@@ -27,16 +27,20 @@ import (
 	config "sigs.k8s.io/kueue/apis/config/v1beta2"
 )
 
-// compareErrors compares errors by their string representation
-func compareErrors(want, got error) bool {
-	if want == nil && got == nil {
+// compareAllErrors checks that every error in wantErrs is present in got.
+func compareAllErrors(wantErrs []error, got error) bool {
+	if len(wantErrs) == 0 && got == nil {
 		return true
 	}
-	if want == nil || got == nil {
+	if len(wantErrs) == 0 || got == nil {
 		return false
 	}
-	// Compare error strings, handling wrapped errors
-	return strings.Contains(got.Error(), want.Error())
+	for _, want := range wantErrs {
+		if !strings.Contains(got.Error(), want.Error()) {
+			return false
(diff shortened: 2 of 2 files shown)


mimowo:
Since this is considered a bug with a user (admin) observable change we need release note, proposal:
/release-note-edit
```release-note
ConfigAPI: Fixed TLS configuration validation to report all detected option errors instead of only the last one.
```

pkg/util/tlsconfig/tlsconfig_test.go
```diff
@@ -185,14 +185,14 @@ func TestParseTLSOptions(t *testing.T) {
...
-			name: "tls 1.0 with invalid cipher suites",
+			name: "tls 1.0 with invalid cipher suites returns both errors",
 			cfg: &config.TLSOptions{
 				MinVersion: "VersionTLS10",
 				CipherSuites: []string{
 					"DUMMY",
 				},
 			},
-			wantErr: errInvalidCipherSuite, // Both TLS version and cipher suite are invalid, but cipher suite error is returned last
+			wantErr: errInvalidMinVersionTLS10or11,
```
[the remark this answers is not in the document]
mimowo: I thinks this is a reasoable comment, I was also thinking about it - it seems the change of the assert is just cosmetics in the current form, it would pass before and after, no? Ideally we can assert that both are present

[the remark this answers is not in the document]
kannon92: updated.

pkg/util/tlsconfig/tlsconfig_test.go
```diff
@@ -51,6 +67,7 @@ func TestParseTLSOptions(t *testing.T) {
 		cfg        *config.TLSOptions
 		expectNil  bool
 		wantErr    error
+		wantErrs   []error
```
sohankunkerkar, line 57: nit: can we consolidate wantErr and wantErrs into a single wantErrs []error as every case uses one or the other, never both?

sohankunkerkar (approved):
/lgtm

kannon92:
/assign @mimowo 

Guess we will need your approval here as this has top level approvers.

pkg/util/tlsconfig/tlsconfig_test.go
kannon92: done.

pkg/util/tlsconfig/tlsconfig.go
```diff
@@ -42,15 +42,15 @@ func ParseTLSOptions(cfg *config.TLSOptions) (*TLS, error) {
 	}
 	version, err := convertTLSMinVersion(cfg.MinVersion)
 	if err != nil {
-		errRet = errors.Join(err)
+		errRet = errors.Join(errRet, err)
```
mbobrovskyi: nit: errRet is nil here, right?

mbobrovskyi:
/lgtm

Thank you!

tenzen-y (commented):
Thank you 👍 
/lgtm
/approve

pkg/util/tlsconfig/tlsconfig.go
kannon92: done.

PR #12293, opened 2026-06-16
Fix webhook server silently dropping TLS parse errors

kannon92:
/kind bug

`AddWebhookSettingsTo` silently ignored `ParseTLSOptions` errors (`if err == nil`), causing the webhook server to start with Go TLS defaults on invalid config. This is inconsistent with `main.go` which calls `os.Exit(1)` on the same error for the metrics server.

This changes `AddWebhookSettingsTo` to return an error, and the call site in `main.go` now exits with a clear error message, matching the metrics server behavior.

Found by @sohankunkerkar during review of #11832.

Partial #10698

AI was used to help create this PR.

```release-note
ConfigAPI: TLS: Fixed a bug where invalid webhook server TLS settings could be silently ignored,
causing the webhook server to start with Go TLS defaults. Kueue now fails startup with a
clear configuration error, matching metrics server TLS validation behavior.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved webhook server startup validation by treating webhook settings application as fallible, surfacing TLS parsing errors when the TLS options feature gate is enabled.
  * Startup now fails fast with a clear error message if webhook settings cannot be configured, preventing silent misconfiguration.
* **Tests**
  * Expanded TLS feature-gate coverage with a new invalid TLS fixture and updated assertions to verify error behavior differs correctly between enabled vs. disabled modes.

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -913,12 +913,25 @@ webhook:
 		t.Fatal(err)
 	}
 
+	tlsConfigInvalid := filepath.Join(tmpDir, "tls-invalid.yaml")
+	if err := os.WriteFile(tlsConfigInvalid, []byte(`apiVersion: config.kueue.x-k8s.io/v1beta2
+kind: Configuration
+namespace: kueue-system
+tls:
+  minVersion: InvalidVersion
+webhook:
+  port: 9443
+`), os.FileMode(0600)); err != nil {
+		t.Fatal(err)
+	}
+
 	testcases := []struct {
 		name              string
 		configFile        string
 		featureGates      map[featuregate.Feature]bool
 		wantConfiguration configapi.Configuration
 		verifyTLSApplied  bool
+		wantError         error
 	}{
 		{
 			name:         "TLS config applied when feature gate enabled",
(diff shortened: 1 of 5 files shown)


mimowo:
/release-note-edit
```release-note
ConfigAPI: TLS: Fixed a bug where invalid webhook server TLS settings could be silently ignored,
causing the webhook server to start with Go TLS defaults. Kueue now fails startup with a
clear configuration error, matching metrics server TLS validation behavior.
```

pkg/config/config_test.go
```diff
@@ -1125,7 +1125,9 @@ webhook:
 			}
 
 			// Call AddWebhookSettingsTo to configure webhook server with TLS options
-			AddWebhookSettingsTo(&options, &cfg)
+			if err := AddWebhookSettingsTo(&options, &cfg); err != nil {
```
mimowo: Would it be feasible to add a unit test for the change?

pkg/config/config_test.go
```diff
@@ -913,12 +913,25 @@ webhook:
...
+		t.Fatal(err)
+	}
+
 	testcases := []struct {
 		name              string
 		configFile        string
 		featureGates      map[featuregate.Feature]bool
 		wantConfiguration configapi.Configuration
 		verifyTLSApplied  bool
+		wantErr           bool
```
mimowo: Let's use some more info than just bool, either full error, full message or at least substring

kannon92: done.

sohankunkerkar (approved):
/lgtm

sohankunkerkar:
@mimowo FYI, This PR has been in the merge pool for over an hour now.

mimowo:
Yeah, we have some issue, started yesterday, ptal: https://kubernetes.slack.com/archives/C09QZ4DQB/p1781628001783129

pkg/config/config_test.go
```diff
@@ -913,12 +914,25 @@ webhook:
...
+		t.Fatal(err)
+	}
+
 	testcases := []struct {
 		name              string
 		configFile        string
 		featureGates      map[featuregate.Feature]bool
 		wantConfiguration configapi.Configuration
 		verifyTLSApplied  bool
+		wantErrSubstring  string
```
tenzen-y: This is untipattern. So, can you use error comparision pattern? We can introduce `wantError error` instead of `wantErrSubstring string`.

kannon92: This PR is in the merge pool. I can address as a followup. I'm thinking this will merge any minute now...

tenzen-y: I'm ok with a follow-up, thank you.

kannon92: who knows now. I'll push up a change.

kannon92: updated.

pkg/config/config_test.go
```diff
@@ -1066,6 +1080,58 @@ webhook:
 			},
 			verifyTLSApplied: true,
 		},
+		{
+			name:         "invalid TLS config returns error when feature gate enabled",
+			configFile:   tlsConfigInvalid,
+			featureGates: map[featuregate.Feature]bool{features.TLSOptions: true},
+			wantError:    fmt.Errorf("unable to parse TLS options for webhook server: invalid minVersion: unknown tls version %q. Please use VersionTLS12 or VersionTLS13", "InvalidVersion"),
```
tenzen-y: This is the same as previously. But, I think that we have to extra work for error comparision pattern here.
The root cause is https://github.com/kubernetes-sigs/kueue/blob/5446b2a38eeec3615d95b3e06ecd0ec311ce7d38/pkg/util/tlsconfig/tlsconfig_test.go#L42-L46

We shouldn't define errors in test. Instead of that, we should define these errors in prod codes and use them, and verify it in tests.

But, as I said, it seems extra work from this PR PoV. So, I'm ok with a follow-up.

mimowo:
This needs rebase

pkg/config/config_test.go
kannon92: Pushed up a fix for this.

kannon92:
PR is up dated.

mimowo:
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for improving the error handling, this is very important for prod systems which often need debugging

kannon92:
We don't really have any reason to carry this. I'm fine to fix this forward.

Issue #11143, opened 2026-05-12
kueueviz: Backend service is missing authorization of requests


PBundyra:
**What happened**:
Currently the kueueviz backend only authenticate incoming requests and doesn't authorize them. It leads to a vulnerability where a user with limited access to some namespaces, can send GET requests and retrieve information about Pods in other namespaces

**What you expected to happen**:
Authorize requests

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
- Kueue version (use `git describe --tags --dirty --always`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PBundyra:
@samzong would you be interested in implementing it?

samzong:
Sure! I’m interested in working on this.

/assign @samzong

kshalot:
Hey, @samzong! Quick question - are you actively working on this ticket?

samzong:
@kshalot sorry for late. I have some upgrade.

https://github.com/kubernetes-sigs/kueue/pull/12302#issuecomment-4726191887

kshalot:
@samzong no worries at all, I was just asking to know whether you were still interested! Thanks for your work so far, I'll take a look 🫡

mimowo:
The implementation PR: https://github.com/kubernetes-sigs/kueue/pull/13810

PR #12302, opened 2026-06-17
KEP: KueueViz authorization

samzong:
#### What type of PR is this?

/kind kep
/area dashboard

This PR adds a KEP for alpha KueueViz authorization.

The KEP defines how KueueViz should authorize authenticated users before returning REST, WebSocket, or aggregated resource data. It proposes using Kubernetes `SubjectAccessReview` with the `UserInfo` produced by TokenReview authentication.

The KEP covers:
- Authorization semantics for namespaced and cluster-scoped resources.
- Endpoint-level access policy for REST and WebSocket routes.
- Aggregated view behavior when a user can access only part of the data.
- SubjectAccessReview caching behavior.
- RBAC requirements for the KueueViz backend service account.
- Frontend behavior for authorization failures.
- Test and graduation criteria for alpha.

This is needed because TokenReview proves user identity, but does not by itself prove that the user may read or watch the resources displayed by KueueViz.

This PR was prepared with assistance from OpenAI Codex, following the Kubernetes AI Tool Usage Policy.

KEP for #11143

This PR is documentation-only and does not change runtime behavior.

The corresponding implementation is intended to be submitted separately after the KEP direction is agreed.

Validation:
- `./bin/mdtoc --max-depth=5 keps/11143-kueueviz-authorization/README.md`: passed.

```release-note
NONE
```

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive documentation for KueueViz authorization using Kubernetes RBAC, detailing access control semantics, configuration requirements, error-handling guidelines, and alpha-stage graduation criteria for v0.19.

### DIFF
--- a/keps/11143-kueueviz-authorization/README.md
+++ b/keps/11143-kueueviz-authorization/README.md
@@ -1,0 +1,564 @@
+# KEP-11143: KueueViz Authorization
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+  - [Authorization Semantics](#authorization-semantics)
+  - [Endpoint Authorization Policy](#endpoint-authorization-policy)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Authentication Context](#authentication-context)
+  - [SubjectAccessReview Authorizer](#subjectaccessreview-authorizer)
+  - [Request Handling](#request-handling)
+  - [WebSocket Handling](#websocket-handling)
+  - [Partial Results for Aggregated Views](#partial-results-for-aggregated-views)
+  - [Caching](#caching)
+  - [RBAC Updates](#rbac-updates)
+  - [Helm Configuration](#helm-configuration)
+  - [Frontend Considerations](#frontend-considerations)
+  - [Test Plan](#test-plan)
+    - [Prerequisite testing updates](#prerequisite-testing-updates)
+    - [Unit tests](#unit-tests)
+    - [Integration tests](#integration-tests)
+    - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Only authorize Pod YAML requests](#only-authorize-pod-yaml-requests)
+  - [Use per-user Kubernetes clients](#use-per-user-kubernetes-clients)
+  - [Use impersonation instead of SubjectAccessReview](#use-impersonation-instead-of-subjectaccessreview)
+  - [Rely on frontend filtering](#rely-on-frontend-filtering)
+  - [Keep authorization optional](#keep-authorization-optional)
+<!-- /toc -->
+
+## Summary
+
+KueueViz authentication, introduced by
+[KEP-5993](../5993-kueueviz-authentication/README.md), validates that an
+incoming bearer token belongs to a Kubernetes user. It does not validate that
+the authenticated user is allowed to read the Kubernetes objects returned by the
+KueueViz backend.
+
+This KEP adds Kubernetes RBAC authorization to KueueViz when authentication is
+enabled. KueueViz will use `SubjectAccessReview` to decide whether the
+authenticated user can read each resource represented by a REST or WebSocket
+request. The backend may continue using its service account and shared informer
+cache to read cluster state, but it must filter or reject responses according
+to the requesting user's Kubernetes RBAC permissions before data leaves the
+backend.
+
+## Motivation
+
+From [Issue #11143](https://github.com/kubernetes-sigs/kueue/issues/11143):
+
+> Currently the kueueviz backend only authenticate incoming requests and doesn't
+> authorize them. It leads to a vulnerability where a user with limited access
+> to some namespaces, can send GET requests and retrieve information about Pods
+> in other namespaces.
+
+The current `TokenReview` flow prevents unauthenticated access, but it still
+creates a confused deputy: users authenticate as themselves, then receive data
+read by the KueueViz backend service account. If that service account has
+cluster-wide read permissions, a user with access to only one namespace may be
+able to read Pods, Events, Workloads, LocalQueues, or other resources in other
+namespaces through KueueViz.
+
+KueueViz must enforce the same authorization boundary users would encounter
+when using `kubectl get` directly against the Kubernetes API server.
+
+### Goals
+
+1. Enforce Kubernetes RBAC authorization for KueueViz API and WebSocket
+   responses when `auth.mode` is `TokenReview`.
+2. Use Kubernetes `SubjectAccessReview` with the authenticated user's identity
+   from `TokenReview`.
+3. Deny direct object requests when the user lacks permission to read the
+   requested resource.
+4. Filter aggregated views so users only receive objects and related details
+   they are authorized to read.
+5. Keep the existing shared informer/cache architecture for KueueViz backend.
+6. Fail closed when authorization cannot be determined.
+7. Add tests that demonstrate a user with access to one namespace cannot read
+   resources from another namespace through KueueViz.
+
+### Non-Goals
+
+1. Replacing the KueueViz authentication mechanism from KEP-5993.
+2. Adding OAuth2, OIDC, mTLS, or server-side sessions.
+3. Creating one Kubernetes client or informer cache per end user.
+4. Changing Kubernetes RBAC semantics or introducing KueueViz-specific roles.
+5. Redesigning the KueueViz UI or API schemas beyond necessary authorization
+   failures and filtered aggregate results.
+6. Per-message authentication or token refresh for existing WebSocket
+   connections.
+7. Supporting authorization when `auth.mode` is `Disabled`. Without an
+   authenticated user identity, KueueViz cannot evaluate user RBAC.
+
+## Proposal
+
+KueueViz will add an authorization layer after authentication and before any
+handler returns Kubernetes object data to the client.
+
+At request time:
+
+1. The authentication middleware validates the token using `TokenReview`.
+2. The middleware stores the full authenticated `UserInfo` in the request
+   context.
+3. Handlers build one or more Kubernetes resource access checks from the
+   endpoint semantics.
+4. The authorizer evaluates those checks with `SubjectAccessReview`.
+5. The handler either returns the requested data, returns HTTP 403, closes the
+   WebSocket, or filters unauthorized objects from an aggregate response.
+
+The KueueViz backend service account remains responsible for reading data from
+the Kubernetes API server and maintaining informers. The new rule is that the
+backend must not disclose data unless the requesting user is authorized to read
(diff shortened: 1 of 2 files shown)


samzong:
Hi maintianers! I already have a local implementation, but it is fairly broad: ~27 files, over than 1000 line changes.

https://github.com/user-attachments/assets/7a25d567-e642-4b6b-a8bc-24535d43b710

It touches backend SAR authz, REST/WebSocket authorization, aggregate filtering, RBAC, small frontend UX, and docs.

To reduce review load, I’m thinking about splitting the implementation into:

1. backend authz foundation + REST/RBAC

2. WebSocket authz + aggregate filtering

3. frontend UX + docs

Would maintainers prefer that split, or one cohesive implementation PR after the KEP direction is agreed?

@mbobrovskyi @mimowo @pajakd

keps/11143-kueueviz-authorization/README.md
```diff
@@ -0,0 +1,564 @@
...
+| `/api/resourceflavor/{name}?output=yaml` | `get` `resourceflavors.kueue.x-k8s.io` | 403 |
+| `/api/node/{name}?output=yaml` | `get` `nodes` | 403 |
+| `/ws/workloads?namespace={ns}` | `list` and `watch` `workloads.kueue.x-k8s.io` in `{ns}` | 403 before upgrade |
+| `/ws/workloads` | `list` and `watch` `workloads.kueue.x-k8s.io` per namespace | filtered namespaces |
+| `/ws/workload/{ns}/{name}` | `get` `workloads.kueue.x-k8s.io` named `{name}` in `{ns}`; `watch` `workloads.kueue.x-k8s.io` in `{ns}` | 403 before upgrade |
+| `/ws/workload/{ns}/{name}/events` | `get` `workloads.kueue.x-k8s.io` named `{name}` in `{ns}`; `list` and `watch` `events` in `{ns}` | 403 before upgrade |
+| `/ws/local-queues` | `list` and `watch` `localqueues.kueue.x-k8s.io` per namespace | filtered namespaces |
+| `/ws/local-queue/{ns}/{name}` | `get` `localqueues.kueue.x-k8s.io` named `{name}` in `{ns}`; `watch` `localqueues.kueue.x-k8s.io` in `{ns}` | 403 before upgrade |
+| `/ws/local-queue/{ns}/{name}/workloads` | `list` and `watch` `workloads.kueue.x-k8s.io` in `{ns}` | 403 before upgrade |
+| `/ws/namespaces` | `list` and `watch` `localqueues.kueue.x-k8s.io` per namespace | filtered namespaces |
```
kshalot, line 184: Shouldn't this check the permissions on the `namespaces` resource?

keps/11143-kueueviz-authorization/README.md
```diff
@@ -0,0 +1,564 @@
...
+endpoints are malformed requests and fail with HTTP 400 before the WebSocket
+upgrade.
+
+For endpoints that can stream partial aggregate data, each snapshot is filtered
+before it is written to the WebSocket. The same filtering is applied on the
+initial snapshot and on every informer-triggered update.
+
+KueueViz will not re-authenticate the token for every WebSocket message in this
+KEP. Token lifetime and refresh remain part of the authentication KEP's
+limitations.
```
kshalot, line 309: We can consider some max timeout on the websocket connections, i.e. dropping the websocket after 10 minutes and re-establishing it on the client.

Maybe not in v1, but could be considered as an improvement in subsequent iterations.

kshalot:
@samzong by the way, I'd replace:
> Fixes https://github.com/kubernetes-sigs/kueue/issues/11143

with something like:

> KEP for https://github.com/kubernetes-sigs/kueue/issues/11143

So we don't close the issue before the implementation is merged.

keps/11143-kueueviz-authorization/kep.yaml
```diff
@@ -0,0 +1,30 @@
+title: KueueViz Authorization
+kep-number: 11143
+authors:
+  - "@samzong"
+status: provisional
+creation-date: 2026-05-12
+reviewers:
+  - "@PBundyra"
```
kshalot, line 8:
```suggestion
  - "@kshalot"
```

keps/11143-kueueviz-authorization/README.md
```diff
@@ -0,0 +1,564 @@
...
+## Drawbacks
+
+This KEP adds backend complexity and extra API server calls. Aggregated views
+become more complex because each piece of data can have different
+authorization requirements. Some users may see partial dashboards rather than a
+single complete cluster view.
+
+These costs are necessary for safe multi-tenant use. Without authorization,
+KueueViz cannot safely be exposed to authenticated users with different
+namespace permissions.
```
kshalot, line 525: I think another cost that this KEP should acknowledge is that in large multi-tenant Kueue deployments, a view like "all the workloads" will require to enumerate all namespaces and to create a SAR for each one. 

So 5000 namespaces would require 5000 requests.

keps/11143-kueueviz-authorization/README.md
```diff
@@ -0,0 +1,564 @@
...
+
+## Alternatives
```
kshalot, line 527: How about using [`SelfSubjectRulesReview`](https://kubernetes.io/docs/reference/kubernetes-api/definitions/self-subject-rules-review-v1-authorization/) instead of SAR?

It returns all the permissions in a namespace, so we could retrieve this once per namespace, we don't have to care about specific verbs. This makes it easier to cache and reduces the total number of calls to the API server.

kshalot (commented):
I made a second pass.

cc @olekzabl

keps/11143-kueueviz-authorization/README.md
```diff
@@ -0,0 +1,564 @@
...
+The `/ws/workloads/dashboard` endpoint is namespace-scoped for alpha. An empty
+or missing `namespace` query parameter is rejected with HTTP 400 before the
+WebSocket upgrade. A cluster-wide dashboard would require all-namespace
+filtering across Workloads, Pods, LocalQueues, ClusterQueues, and
+ResourceFlavors, and is out of scope for the first implementation.
+
+The frontend currently initializes the dashboard with an empty namespace
+selection. The alpha implementation must avoid opening
+`/ws/workloads/dashboard?namespace=` by selecting an authorized namespace first
+or by showing an empty-state prompt until the user chooses one.
```
olekzabl, line 223: This is all valid, though a bit invasive for the current UX.
I understand fanning out per namespace is a major change, and it makes sense to postpone it for post-alpha.

However, there's another _partial_ solution: even namespaced resources can be opened cluster-scoped, via a `ClusterRole` and `ClusterRoleBinding`. And if so, we can use cluster-scoped SARs to check this. No fanout.
So, in case when a user was given broad access to Kueue objects, we could save the cluster-scoped view working quite easily.
I'm not saying it _should_ be done in Alpha - but please consider.

(To clarify: I don't propose it _instead of_ per-namespace fanout; the fanout could be added later as another strategy of getting permissions for a view).

keps/11143-kueueviz-authorization/README.md
olekzabl: Reading the description of SSRR

> SelfSubjectRulesReview should be used by UIs to show/hide actions, or to quickly let an end user reason about their permissions. _It should NOT Be used by external systems to drive authorization decisions_

My understanding is that:

- example good use cases of SSRR are:
  - quickly prefetch user permissions to decide which UI elements should be disabled
  - (if desired) quickly prefetch user permissions to speed up data retrieval before SAR responses arrive
- however, the actual transmission of data should be never gated on SSRR alone.

So I think "using SSRR instead of SAR" goes against recommendations.

We could use it "in front of SAR" (so that SAR later verifies if we've indeed computed the "right" view) though in our case I don't see much value of that.

olekzabl (commented):
Wow, this looks solid, thank you!
I'll continue to read in depth but let me post 2 quick comments.

keps/11143-kueueviz-authorization/README.md
kshalot: Both SAR and SSRR go around admission controllers like Kyverno, so both have a gap. But this made me wonder, why don't we just:
1. Send SSRR to populate the UI.
2. Pass the Bearer token/Impersonate user on the actual API calls.

Otherwise it seems we try to reinvert the RBAC logic in an external system

kshalot: Nevermind - it's in the alternatives.

keps/11143-kueueviz-authorization/README.md
danylott: +1 for UX, I believe we should aim for showing user in dashboard useful data by default, not only after some additional actions

keps/11143-kueueviz-authorization/README.md
```diff
@@ -0,0 +1,564 @@
...
+is not authorized. Instead, they should omit unauthorized secondary resources.
+
+Examples:
+
+- Dashboard data may include Workloads for a namespace but omit Pod details if
+  the user lacks `list pods` or `watch pods` in that namespace.
+- ClusterQueue detail may include the ClusterQueue but omit LocalQueues in
+  namespaces where the user lacks `list localqueues` or `watch localqueues`.
+- ResourceFlavor detail may include flavor details but omit matching Nodes if
+  the user lacks `list nodes` or `watch nodes`.
```
danylott, line 323: Could we add requirement to the design to match UX expectations for Partial Results & unauthorized requests?

We should let user know using for ex. banners about missing permissions or partially loaded data.
Without these users will be confused, why they can access some part of data, but then nothing is inside. As KueueViz is UI project - I'd like to include this as a part of requirements for this proposal

keps/11143-kueueviz-authorization/README.md
```diff
@@ -0,0 +1,564 @@
...
+
+The backend should not include object names, counts, or status information for
+unauthorized resources.
+
+### Caching
+
+SubjectAccessReview results may be cached to reduce API server load. The cache
```
danylott, line 330: Given the high volume of SAR calls KueueViz will likely generate per user (especially if we evaluate multiple resources or namespaces for aggregate views), an uncached implementation could easily overwhelm the Kubernetes API server.

Should this be upgraded from 'may' to 'must' or 'should'? Making caching a hard requirement seems critical to protect the control plane.

keps/11143-kueueviz-authorization/README.md
```diff
@@ -0,0 +1,564 @@
...
+- 403 means the user is authenticated but not authorized; show a permission
+  error without clearing the token.
+
+The frontend should not attempt client-side filtering as a security boundary.
+All filtering must happen in the backend.
+
+For WebSocket authorization failures before upgrade, browsers may expose only a
+generic connection failure to JavaScript. The frontend should not clear the
+token for these failures. HTTP 403 responses must remain distinct from 401 and
+must not trigger logout.
```
danylott, line 398: As mentioned higher - Could we please add UX for unauthorized requests & partial results here as a requirement?

keps/11143-kueueviz-authorization/README.md
danylott: +1, I believe for "all the workloads" request we need to use Cluster scoped authorization request, instead of 5000 namespace-based ones - So just reject if Cluster scoped authorization is not present, and then user switched to using the per-namespace ones.

mykysha:
Hi @samzong ! Do you have time to continue working on this? If not I can update the KEP to address the remaining comments, so that we can proceed with the implementation #13810

kannon92:
> Hi @samzong ! Do you have time to continue working on this? If not I can update the KEP to address the remaining comments, so that we can proceed with the implementation #13810

Based on this.

/close

Issue #12303, opened 2026-06-17
Optimize post-kueue-push-images which is currently the longest job blocking merges


mimowo:
**What would you like to be cleaned**:

I would like to optimize the post-kueue-push-images CI job which is currently the longest job blocking merges.

1. build the helper images like raymini, debug, importer in periodics only (remove from postsubmits)
2. build the images concurrently, we may need to push sequentially, but we could do two phases: (1.) concurrent builds, (2.) sequential pushes; currently we sequence everything.

**Why is this needed**:

The job is currently taking >21min (example [latest build taking 21min 44s](https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-kueue-push-images/2067096268827529216)) and is blocking merges, significantly undermining the effort to make builds faster in https://github.com/kubernetes-sigs/kueue/issues/11606

[Job dashboard](https://testgrid.k8s.io/sig-scheduling#post-kueue-push-images)

mimowo:
/assign @mbobrovskyi 
with whom I already discussed on slack tentatively
cc @sohankunkerkar @BenTheElder

mimowo:
cc @tenzen-y

mimowo:
> https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-kueue-push-images/2067161237023625216
> Test started today at 10:23 AM passed after **6m37s.** (more info)

This is a great result and I think it will improve merge time significantly 🚀

PR #12304, opened 2026-06-17
Refactor cloudbuild and introduce parallelism

mbobrovskyi:
#### What type of PR is this?
/kind cleanup

Split cloudbuild for presubmit and periodic.

Part of #12303

```release-note
NONE
```

## Summary by CodeRabbit

* **Chores**
  * Updated build orchestration for publishing container image and chart artifacts, adding dedicated release-artifacts and helper aggregation workflows.
  * Improved build performance by adjusting parallelism based on run type (periodic vs non-periodic).
  * Streamlined the CI image-push stage by invoking a single consolidated `make` target instead of multiple individual image-push targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -244,6 +244,20 @@ image-build:
 		$(IMAGE_BUILD_EXTRA_OPTS) \
 		./
 
+.PHONY: image-pushing-presubmit
+image-pushing-presubmit:
+	$(MAKE) -j5 run-image-pushing-presubmit
+
+.PHONY: run-image-pushing-presubmit
+run-image-pushing-presubmit: image-push helm-chart-push kueueviz-image-push kueue-populator-image-push kueue-priority-booster-image-push
+
+.PHONY: image-pushing-periodic
+image-pushing-periodic:
+	$(MAKE) -j3 run-image-pushing-periodic
+
+.PHONY: run-image-pushing-periodic
+run-image-pushing-periodic: debug-image-push importer-image-push ray-project-mini-image-build-push
+
 .PHONY: image-push
 image-push: PUSH=--push
 image-push: image-build
--- a/Makefile
+++ b/Makefile
@@ -244,19 +244,19 @@ image-build:
 		$(IMAGE_BUILD_EXTRA_OPTS) \
 		./
 
-.PHONY: image-pushing-presubmit
-image-pushing-presubmit:
-	$(MAKE) -j5 run-image-pushing-presubmit
+.PHONY: image-pushing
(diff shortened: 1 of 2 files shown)


cloudbuild-periodic.yaml
```diff
@@ -0,0 +1,23 @@
+# See https://cloud.google.com/cloud-build/docs/build-config
```
mimowo: Do we need another cloudbuild, or we could reuse the same? I think we could just use the JOB_TYPE env variable to discriminate periodic vs. postsubmit, wdyt?

mimowo: For example I see that kubetest2 in `config/jobs/image-pushing/k8s-staging-kubetest2.yaml` is using both postsubmits and periodics, but I could only find one cloudbuild in their repo

mbobrovskyi: Done

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for the PR, this should already substantially improve the merge time 👍

mimowo:
/hold
Actually I will have another comment

Makefile
```diff
@@ -244,6 +244,20 @@ image-build:
...
 
+.PHONY: image-pushing
+image-pushing-presubmit:
+ifeq ($(JOB_TYPE),periodic)
+	$(MAKE) -j3 image-pushing-periodic
+else
+	$(MAKE) -j5 image-pushing-presubmit
+endif
+
+.PHONY: image-pushing-presubmit
```
mimowo:
```suggestion
.PHONY: image-pushing-release-artifacts
```
Actually, I think "presubmit" is misleading because this job will run as postsubmit, because this is how test-infra is configuring it: https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/k8s-staging-kueue.yaml#L1

Makefile
```diff
@@ -244,6 +244,20 @@ image-build:
...
+image-pushing-presubmit: image-push helm-chart-push kueueviz-image-push kueue-populator-image-push kueue-priority-booster-image-push
+
+.PHONY: image-pushing-periodic
```
mimowo:
```suggestion
.PHONY: image-pushing-helper-artifacts
```

Makefile
mbobrovskyi: Done

Makefile
mbobrovskyi: Done

PR #12309, opened 2026-06-17
Adds missing check to importer.

### LINKED ISSUE #12306: Kueue importer not checking for LQ namespace when listing LQs.
**What happened**:
In cmd/importer/cache/cache.go the Load function does not check listed local queues for namespace. 

When loading local queues per namespace, we expect only the LQs within the namespace to be loaded.

Singularity23x0:
#### What type of PR is this?
/kind bug

Adds a check to cmd/importer/cache/cache.go to ensure we list only local queues within the desired namespace. 

Fixes #12306

NONE

```release-note
Importer: Fixed LocalQueue namespace isolation to prevent information leakage between
namespaces when multiple LocalQueues with the same name exist in different namespaces.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed namespace scoping during queue import so local queues are fetched only for the specified namespaces, ensuring correct isolation.

* **Tests**
  * Updated importer integration tests to cover multiple namespaces and cluster queues, including shared local-queue names.
  * Added a new integration test to confirm importer-generated local-queue assignments are correctly mapped per namespace.

### DIFF
--- a/test/integration/singlecluster/importer/importer_test.go
+++ b/test/integration/singlecluster/importer/importer_test.go
@@ -180,7 +180,7 @@ var _ = ginkgo.Describe("Importer", func() {
 		})
 	})
 
-	ginkgo.It("Should import only mathching namespace", framework.SlowSpec, func() {
+	ginkgo.It("Should import only matching namespace", framework.SlowSpec, func() {
 		pod1 := utiltestingpod.MakePod("pod1", ns1.Name).
 			Label("src.lbl", "src-val").
 			Request(corev1.ResourceCPU, "2").
--- a/test/integration/singlecluster/importer/importer_test.go
+++ b/test/integration/singlecluster/importer/importer_test.go
@@ -180,32 +180,27 @@ var _ = ginkgo.Describe("Importer", func() {
 		})
 	})
 
-	ginkgo.It("Should import only matching namespace", framework.SlowSpec, func() {
-		pod1 := utiltestingpod.MakePod("pod1", ns1.Name).
-			Label("src.lbl", "src-val").
-			Request(corev1.ResourceCPU, "2").
-			Obj()
+	ginkgo.It("Should assign local queues to appropriate namespaces", framework.SlowSpec, func() {
+		var mapping *importercache.ImportCache
+		var err error
 
-		ginkgo.By("Creating the pod", func() {
-			util.MustCreate(ctx, k8sClient, pod1)
-		})
-
-		ginkgo.By("Running the import for ns1", func() {
-			mapping, err := importercache.Load(ctx, k8sClient, []string{ns1.Name}, importermapping.RulesForLabel("src.lbl", map[string]string{"src-val": lqName}), nil)
+		ginkgo.By("Importing across all namespaces", func() {
+			mapping, err = importercache.Load(ctx, k8sClient, []string{ns1.Name, ns2.Name}, importermapping.RulesForLabel("src.lbl", map[string]string{"src-val": lqName}), nil)
 			gomega.Expect(err).ToNot(gomega.HaveOccurred())
 			gomega.Expect(mapping).ToNot(gomega.BeNil())
-
-			gomega.Expect(importerpod.Check(ctx, k8sClient, mapping, 8)).To(gomega.Succeed())
-			gomega.Expect(importerpod.Import(ctx, k8sClient, mapping, 8)).To(gomega.Succeed())
 		})
 
-		wl1 := &kueue.Workload{}
-		wl1LookupKey := types.NamespacedName{Name: pod.GetWorkloadNameForPod(pod1.Name, pod1.UID), Namespace: ns1.Name}
+		ginkgo.By("Checking the local queues are assigned correctly", func() {
+			gomega.Expect(mapping.LocalQueues).To(gomega.HaveLen(2))
 
-		ginkgo.By("Checking the Workload for Pod1 is admitted to cq1", func() {
-			gomega.Expect(k8sClient.Get(ctx, wl1LookupKey, wl1)).To(gomega.Succeed())
-			gomega.Expect(wl1.Status.Admission).ToNot(gomega.BeNil())
-			gomega.Expect(wl1.Status.Admission.ClusterQueue).To(gomega.Equal(kueue.ClusterQueueReference(cq1.Name)))
+			gomega.Expect(mapping.LocalQueues[ns1.Name]).To(gomega.HaveLen(1))
+			gomega.Expect(mapping.LocalQueues[ns2.Name]).To(gomega.HaveLen(1))
+
+			gomega.Expect(mapping.LocalQueues[ns1.Name]).To(gomega.HaveKey(lqName))
+			gomega.Expect(mapping.LocalQueues[ns2.Name]).To(gomega.HaveKey(lqName))
+
+			gomega.Expect(mapping.LocalQueues[ns1.Name][lqName].Namespace).To(gomega.Equal(ns1.Name))
+			gomega.Expect(mapping.LocalQueues[ns2.Name][lqName].Namespace).To(gomega.Equal(ns2.Name))
 		})
 	})
 })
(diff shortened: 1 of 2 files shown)


test/integration/singlecluster/importer/importer_test.go
```diff
@@ -157,4 +179,33 @@ var _ = ginkgo.Describe("Importer", func() {
 			})
 		})
 	})
+
+	ginkgo.It("Should import only mathching namespace", framework.SlowSpec, func() {
```
kshalot:
```suggestion
	ginkgo.It("Should import only matching namespace", framework.SlowSpec, func() {
```

olekzabl: AFAIU this test could pass even without your fix.
Because the order of LocalQueues returned from client.List is non-deterministic.
So, even without your fix, it could happen that lq2 is returned first, lq1 later, then lq1 overrides lq2 as the value under `lq.Name`... and everything works, from the perspective of pod1.
(In this scenario, it would fail if we had a pod2 in ns2, purposed for cq2 via lq2).

Hence, I'd suggest to:
- define pod2 in this test as well
- pass both namespaces to `Load`
- verify that pod1 is admitted in cq1 _and_ pod2 is admitted in cq2

And this will make a test case that would fail without your fix in 100% of cases.

WDYT?

olekzabl (commented):
Thank you! 1 comment.

test/integration/singlecluster/importer/importer_test.go
```diff
@@ -157,4 +179,33 @@ var _ = ginkgo.Describe("Importer", func() {
...
+			gomega.Expect(mapping).ToNot(gomega.BeNil())
+
+			gomega.Expect(importerpod.Check(ctx, k8sClient, mapping, 8)).To(gomega.Succeed())
+			gomega.Expect(importerpod.Import(ctx, k8sClient, mapping, 8)).To(gomega.Succeed())
+		})
+
+		wl1 := &kueue.Workload{}
+		wl1LookupKey := types.NamespacedName{Name: pod.GetWorkloadNameForPod(pod1.Name, pod1.UID), Namespace: ns1.Name}
+
+		ginkgo.By("Checking the Workload for Pod1 is admitted to cq1", func() {
```
kshalot: Is this equivalent to "Nothing was imported from the other namespace"? We might need a more explicit check

test/integration/singlecluster/importer/importer_test.go
Singularity23x0: How non-deterministic is the call - if we do "List" for both namespaces are we sure the order will be the same for both calls in the scope of a single execution? If not then we could get reverse orders within the loop and the test will still be prone to succeeding when it should not.

test/integration/singlecluster/importer/importer_test.go
olekzabl: I think the test name does not most faithfully reflect the thing which is being fixed (and which should be tested).

Importing happens on the level of Pods, and that is already namespace-gated well, [here](https://github.com/kubernetes-sigs/kueue/blob/5446b2a38eeec3615d95b3e06ecd0ec311ce7d38/cmd/importer/pod/check.go#L35).

The bug we're fixing is more subtle: opening a way of collision between LocalQueue (with same names but in different namespaces).

My [other comment](https://github.com/kubernetes-sigs/kueue/pull/12309#discussion_r3428587095) was focused on making the test verify that fix.

Now, from this comment my takeaway is that we should ideally adjust the test name.

Singularity23x0: This means "Pod1 was not overwritten by Pod2 due to lq1 overriding lq2 bc of a shared name"

kshalot: Ok, I read the issue report. I see now.

I'd consider adjusting the test names or adding more By() blocks to make the expectation explicit.

Singularity23x0: Rewritten the test to more explicitly test the updated behavior.

test/integration/singlecluster/importer/importer_test.go
Singularity23x0: Rewrote the test to perform a load on both namespaces.
My concerns still stand however. I am not sure we can mitigate the issue unless the load is consistent in the results it provides.

Singularity23x0: Update: it seems the lqs are listed in order based on /registry/kueue.x-k8s.io/localqueues/[namespace]/[name]. So we should be good now.

test/integration/singlecluster/importer/importer_test.go
```diff
@@ -157,4 +179,28 @@ var _ = ginkgo.Describe("Importer", func() {
...
+			gomega.Expect(mapping.LocalQueues).To(gomega.HaveLen(2))
+
+			gomega.Expect(mapping.LocalQueues[ns1.Name]).To(gomega.HaveLen(1))
+			gomega.Expect(mapping.LocalQueues[ns2.Name]).To(gomega.HaveLen(1))
+
+			gomega.Expect(mapping.LocalQueues[ns1.Name]).To(gomega.HaveKey(lqName))
+			gomega.Expect(mapping.LocalQueues[ns2.Name]).To(gomega.HaveKey(lqName))
+
+			gomega.Expect(mapping.LocalQueues[ns1.Name][lqName].Namespace).To(gomega.Equal(ns1.Name))
+			gomega.Expect(mapping.LocalQueues[ns2.Name][lqName].Namespace).To(gomega.Equal(ns2.Name))
```
olekzabl, line 203: Nit: why not compare LQs as whole objects?

```suggestion
			gomega.Expect(mapping.LocalQueues[ns1.Name][lqName]).To(gomega.Equal(lq1))
			gomega.Expect(mapping.LocalQueues[ns2.Name][lqName]).To(gomega.Equal(lq2))
```

(But if this fails for some reason, I'm fine to keep it your way)

test/integration/singlecluster/importer/importer_test.go
olekzabl: lgtm, thanks

test/integration/singlecluster/importer/importer_test.go
kshalot: > Update: it seems the lqs are listed in order based on /registry/kueue.x-k8s.io/localqueues/[namespace]/[name]. So we should be good now.

To confirm - is the order deterministic? Then can we confirm whether this fails without the fix?

olekzabl: > My concerns still stand however. I am not sure we can mitigate the issue unless the load is consistent in the results it provides.

Good catch; it could still pass without the fix if the ordering is random. 
Though IIRC the new version would have 25% chance for passing (while the old had 50%).
Generally, 1/N^2 if we used N namespaces.

> Update: it seems the lqs are listed in order based on /registry/kueue.x-k8s.io/localqueues/[namespace]/[name]. So we should be good now.

Ah, it's not cached, so deterministic. Another good point.

olekzabl: BTW I like your new test version more than what I proposed :) Thanks for this.

mimowo:
@Singularity23x0 please add a release note with the "Importer" because this is a bugfix with user-observable impact

test/integration/singlecluster/importer/importer_test.go
Singularity23x0: Comparing LocalQueues does not work out of the gate. This way is simpler and frankly - more explicit.

mimowo:
I think the information leakage between namespaces i the name problem, but feel to correct me if I'm missing something

/release-note-edit
```release-note
Importer: Fixed LocalQueue namespace isolation to prevent information leakage between
namespaces when multiple LocalQueues with the same name exist in different namespaces.
```

mimowo:
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for the vulnerability fix 👍

PR #12331, opened 2026-06-18
Replace admission race condition repro with integration test

### LINKED ISSUE #11643: Replace the e2e test for the preemptionExpectations into more reliable repro by integration test

I would like to replace the e2e test added in https://github.com/kubernetes-sigs/kueue/pull/11502/ with an integration  test which has a higher chance of reproducing the problem.

**Why is this needed**:

The e2e while teoretically could repro the issue is not sensitive enough to repro the bug. It required careful choice of sleeps injected into the production code.

kshalot:
#### What type of PR is this?
/kind cleanup
/area testing

The non-deterministic, hacky e2e test that requires changes in `pkg/` to reproduce the issue is not ideal, as it mostly served documentation purposes. This integration test is deterministic and always fails without the changes from https://github.com/kubernetes-sigs/kueue/pull/11502.

Fixes #11643

AI Assisted - I told Gemini the flow of the test and generated the code.

```release-note
NONE
```

## Summary by CodeRabbit

* **Tests**
  * Added an integration test that deterministically reproduces a scheduler race between admission updates and within-cluster preemption, validating eviction then successful admission with correct quota reservation behavior.
  * Removed an e2e baseline test that previously covered the same admission race condition scenario.


test/integration/singlecluster/scheduler/scheduler_test.go
```diff
@@ -3256,6 +3256,134 @@ var _ = ginkgo.Describe("Scheduler", func() {
...
+
+			ginkgo.By("Waiting for wl1 to be evicted and simulating the job controller unsetting QuotaReserved after eviction")
+			gomega.Eventually(func(g gomega.Gomega) {
+				wl1Created := &kueue.Workload{}
+				g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(wl1), wl1Created)).To(gomega.Succeed())
+				// Subsequent scheduling cycles for wl2 should re-trigger the eviction after the first one was lost due to the race condition.
+				if meta.IsStatusConditionTrue(wl1Created.Status.Conditions, kueue.WorkloadEvicted) {
+					_ = workload.PatchAdmissionStatus(ctx, k8sClient, wl1Created, nil, func(wl *kueue.Workload) (bool, error) {
+						return workload.UnsetQuotaReservationWithCondition(wl, "Pending", "Evicted", metav1.Now().Time), nil
+					})
```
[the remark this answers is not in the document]
kshalot: SGTM

test/integration/singlecluster/scheduler/scheduler_test.go
```diff
@@ -3256,6 +3256,134 @@ var _ = ginkgo.Describe("Scheduler", func() {
...
+			util.ExpectObjectToBeDeleted(ctx, k8sClient, highWPC, true)
+			util.ExpectObjectToBeDeleted(ctx, k8sClient, q, true)
+		})
+
+		ginkgo.It("Should not deadlock the preempting workloads", func() {
+			admissionPatchStarted := make(chan struct{})
+			allowAdmissionPatch := make(chan struct{})
+			evictionPatchAttempted := make(chan struct{})
+
+			var evictionPatchCount atomic.Int32
```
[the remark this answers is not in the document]
kshalot: SGTM

test/integration/singlecluster/scheduler/scheduler_test.go
```diff
@@ -3257,6 +3256,143 @@ var _ = ginkgo.Describe("Scheduler", func() {
 			}, util.ConsistentDuration, util.ShortInterval).Should(gomega.Succeed())
 		})
 	})
+
+	// See: https://github.com/kubernetes-sigs/kueue/issues/11480
+	ginkgo.When("The admission goroutine races with preemptions in the next scheduling cycle", func() {
```
mimowo: Could you maybe add a bit more context in the comment, specifically:
1. how reliable the repro is when the fix is reverted, at least estimation
2. which assert is expected to fail when the fix is reverted

nit, but also wind the comment under the ginkgo.When (as the first line) so that the comment does not drift away when we move the tests around potentially in the future

test/integration/singlecluster/scheduler/scheduler_test.go
```diff
@@ -3257,6 +3256,143 @@ var _ = ginkgo.Describe("Scheduler", func() {
...
+					return fallThrough, nil
+				}
+
+				return fallThrough, nil
+			}
+			defer func() {
+				if !admissionPatchReleased {
+					close(allowAdmissionPatch)
+				}
+				fakeSubResourcePatchSpec = nil
```
mimowo: nit, This doesn't see to do anything

test/integration/singlecluster/scheduler/scheduler_test.go
```diff
@@ -3257,6 +3256,143 @@ var _ = ginkgo.Describe("Scheduler", func() {
...
+					// Allow subsequent eviction patches
+					return fallThrough, nil
+				}
+
+				return fallThrough, nil
+			}
+			defer func() {
+				if !admissionPatchReleased {
+					close(allowAdmissionPatch)
+				}
```
mimowo, line 3386: I'm not sure if this is relevant here, but it seems to me we might be calling close() twice on the allowAdmissionPatch channel, now? If we enter this code when `admissionPatchReleased=false`, then we will enter the test code in L3371 unconditionally.

mimowo: It might not be able to happen for some non-trivial reason, but maybe it would be safer to do a safeguard using sync.Once, wdyt?

test/integration/singlecluster/scheduler/scheduler_test.go
kshalot: `fakeSubResourcePatchSpec` is global for this suite, so it has to be `nil` to not leak into other tests. It's already set in `BeforeEach`, but this is in the `defer` just in case the `AfterEach` cleanup triggers it.

I think it would work without it... But we can keep it just in case.

test/integration/singlecluster/scheduler/scheduler_test.go
kshalot: Done

test/integration/singlecluster/scheduler/scheduler_test.go
kshalot: Maybe the flow is unintuitive, but I think it will only be called once.

If `defer` is called and `admissionPatchReleased` is false, then it means we did not close the channel yet, since we have this:
```go
			admissionPatchReleased = true
			close(allowAdmissionPatch)
```

So the channel would only be closed if `admissionPatchReleased` was true.

test/integration/singlecluster/scheduler/scheduler_test.go
kshalot: We overwrite it explicitly with `nil` in another place as well:
https://github.com/kubernetes-sigs/kueue/blob/a09a6e1cf2b490e7b5dc0b150126b75d8e3f82b6/test/integration/singlecluster/scheduler/scheduler_test.go#L858-L861

this time in `AfterEach`, which makes less sense than a defer to me.

mimowo: Actually, I also found it in BeforeEach: https://github.com/kubernetes-sigs/kueue/blob/e03d45371ab2cc55be9596086772e036c0f74375/test/integration/singlecluster/scheduler/suite_test.go#L83

I see this is just one harmless line, but I think it is confusing to put it in many places. I would stick to BeforeEach or AfterEach only.

kshalot: I know it's there, see:

> fakeSubResourcePatchSpec is global for this suite, so it has to be nil to not leak into other tests. It's already set in BeforeEach, but this is in the defer just in case the AfterEach cleanup triggers it.

Having said that, I think in this case we can rely on that. I dropped the explicit niling.

test/integration/singlecluster/scheduler/scheduler_test.go
mimowo: Yes, you are right, I somehow assumed the two places could be called from different goroutines

test/integration/singlecluster/scheduler/scheduler_test.go
```diff
@@ -3257,6 +3256,145 @@ var _ = ginkgo.Describe("Scheduler", func() {
...
+			ginkgo.By("Waiting for wl1 to be evicted and simulating the job controller unsetting QuotaReserved after eviction")
+			gomega.Eventually(func(g gomega.Gomega) {
+				wl1Created := &kueue.Workload{}
+				g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(wl1), wl1Created)).To(gomega.Succeed())
+				// Subsequent scheduling cycles for wl2 should re-trigger the eviction after the first one was lost due to the race condition.
+				if meta.IsStatusConditionTrue(wl1Created.Status.Conditions, kueue.WorkloadEvicted) {
+					g.Expect(workload.PatchAdmissionStatus(ctx, k8sClient, wl1Created, nil, func(wl *kueue.Workload) (bool, error) {
+						return workload.UnsetQuotaReservationWithCondition(wl, "Pending", "Evicted", metav1.Now().Time), nil
+					})).To(gomega.Succeed())
+				}
```
mimowo: Can we split this into two Eventually blocks to make it easier to follow?

test/integration/singlecluster/scheduler/scheduler_test.go
```diff
@@ -3257,6 +3256,145 @@ var _ = ginkgo.Describe("Scheduler", func() {
...
+					<-allowAdmissionPatch
+					return fallThrough, nil
+				}
+
+				// Intercept wl1 eviction patch (triggered by wl2)
+				if wl.Name == wl1.Name && meta.IsStatusConditionTrue(wl.Status.Conditions, kueue.WorkloadEvicted) {
+					count := evictionPatchCount.Add(1)
+					if count == 1 {
+						// Drop the first eviction patch to simulate it being immediately overwritten
+						// by the async admission patch before the cache can process it.
```
mimowo: Dropping the request is a bit disruptive to the flow, even if we can reproduce the workload getting stuck, because this would not happen on prod. I think it is ok to manipulate time by delaying some requests, but dropping is a bit too disruptive.

So, before I accept I need to at least understand well if alternatives are feasible or not. 

I would assume (but without thinking very carefully about it yet), that what we need is to intercept admission (this is already done here)https://github.com/kubernetes-sigs/kueue/pull/12331/changes#diff-8fc55df1e8578a6eb073960491a7de3ee974e604745ac3c93a99a068176b9c8bR3334, then proceed with eviction (which would add the preemptionExpectations), and only then resume the admission.

IIUC this is not done this way in order to make the test deterministic (asking to double check my understanding).

If this is the case, then what would be the ratio of reproducing the problem if we didn't drop the Evict request?

test/integration/singlecluster/scheduler/scheduler_test.go
kshalot: done

test/integration/singlecluster/scheduler/scheduler_test.go
kshalot: I can try to run it in a loop without it, but my hunch is that the ratio might be low.

My thinking here was - I want to simulate a scenario, where the admission and eviction requests are spaced so closely in time, that:
1. Eviction arrives.
2. It is **immediately** overwritten by the admission.
3. From Kueue's perspective, the eviction is invisible - when handling the event after the update, Kueue never sees the eviction update because the events got deduped etc.

Dropping the first eviction seemed cleanest to me, but it still warranted a comment, that's why added it.

kshalot: Alternatively, we can try holding off both requests like:
1. Hold admission update until eviction signals that it wants to happen.
2. Eviction signals.
3. In a rapid succession both requests are ungated.

But maybe that's not required, we'll see.

mimowo: Hm, but why we need to override the eviction immediately by the admission to repro the case? So suppose there is a 10s gap between the eviction and the ongoing admission, is Kueue going to clean the preemption expectations during that gap?

kshalot: Yes. Consider the case with two workloads - `lowPrio` and `highPrio`. Let's suppose there is a 10s gap between them so:
1. `lowPrio` is evicted.
2. 10s elapsed.
3. `lowPrio` is admitted.

When `highPrio` evicts `lowPrio`, it sets the expectations and updates `lowPrio` with `Evicted` (Point 1). 

**In the next scheduling cycle**, if the `Evicted` condition was not overriden by the admission (in our scenario it wasn't, because the next scheduling cycle starts faster than the 10s), the preemption expectation will be satisfied here:
https://github.com/kubernetes-sigs/kueue/blob/1c7c8c6a340560c048d0f3b7bdea3316badacbe3/pkg/scheduler/preemption/preemption.go#L209-L214

Now if this happens very quickly, i.e. Kueue never sees `lowPrio` with `Evicted` condition, then we will fall here in the next cycles, forever:
https://github.com/kubernetes-sigs/kueue/blob/1c7c8c6a340560c048d0f3b7bdea3316badacbe3/pkg/scheduler/preemption/preemption.go#L215-L221

It might be helpful to see both those conditions at once:
https://github.com/kubernetes-sigs/kueue/blob/1c7c8c6a340560c048d0f3b7bdea3316badacbe3/pkg/scheduler/preemption/preemption.go#L209-L221

mimowo: I have synced with @kshalot and tried to debug the test to make it reproducing the issue just by time manipulations to avoid silencing a request, so far no success. The plan is to continue the path, but deeper understanding of the interactions is needed.

kshalot: Note to self: to me, the immediate difference between dropping the first eviction request completely and trying to induce a race via timing is that it's missing a transition from "pending" state to "admitted" (the eviction request will set QuotaReserved, Admitted and Evicted to true on wl1), so this code ends up being executed upon eviction:
https://github.com/kubernetes-sigs/kueue/blob/9cd3329e7514719240dc102a6a7d4a0b4aa73063/pkg/controller/core/workload_controller.go#L1316-L1323

Then afterwards, the update with the admission will overwrite Evicted to false, which is an "admitted" -> "admitted" transition, executing this code:
https://github.com/kubernetes-sigs/kueue/blob/9cd3329e7514719240dc102a6a7d4a0b4aa73063/pkg/controller/core/workload_controller.go#L1370-L1373

In our reproduction attempts, the observation was done here:
https://github.com/kubernetes-sigs/kueue/blob/9cd3329e7514719240dc102a6a7d4a0b4aa73063/pkg/controller/core/workload_controller.go#L1352-L1354

When the workload transitioned from QuotaReserved/Admitted to Pending.

kshalot: @mimowo The part that is crucial for the repro to work is simulating/forcing event deduplication in the Kueue informers - we need the events to:
1. Evict wl1.
2. Admit wl2.

Be squished into one event, so that our reconcilers do not fire for the eviction event. That's what the "drop first update" approach is trying to emulate, but I can't think of any other way to do it... yet.

mimowo: This is surprising, because we had a repro using e2e tests by injecting custom sleeps. So, I assume in the integration test we just need to reproduce the same flow of requests. Maybe worth recording what is the sequence of requests in e2e as the initial step.

kshalot: The "meat" of the reproduction were channels, not sleeps. AFAIR the reproduction works without sleeps and it's non deterministic.

The original e2e repro can be seen in full [here](https://github.com/kubernetes-sigs/kueue/commit/6db3082613caa25111f5d72a625855bd85b22f05#diff-cd44f6e671ae7fdc7882224c13986f4b1b98e1ea45f65bdbcc9ae185d8e91496). But in short, it used channels to force the admission patch to happen right after the eviction patch, by closing a channel when issuing the preemption which resumes the admission. So the two requests are sent very close together in time, which sometimes produces this race condition via event deduplication.

That's why the reproduction was non-deterministic. This is my current understanding of the whole issue.

kshalot: ^ Nevermind, @mimowo the approach we tested on Friday worked the entire time. What we missed:

1. The `KeepGoing` signal goes around backoff.
2. `FinishEvictionForWorkloads` was called too soon. We had to wait for the condition to be overwritten first.

Added a new commit to discuss what's the best way to inject a delay into the scheduling cycle (temporary added a sleep for demonstration purposes).

kshalot:
@mimowo A proposal from me: https://github.com/kubernetes-sigs/kueue/pull/12331/changes/4ac6a027cc4ca89c6a2eb118eee1e31f4e486bab

This has 0 impact on the production code and other suites.

test/integration/singlecluster/scheduler/scheduler_test.go
```diff
@@ -3257,6 +3257,146 @@ var _ = ginkgo.Describe("Scheduler", func() {
...
+
+			fakeSubResourcePatchResponseHookSpec = func(obj client.Object, err error) (fakeClientUsage, error) {
+				wl, ok := obj.(*kueue.Workload)
+				if !ok {
+					return fallThrough, nil
+				}
+				if wl.Name == wl1.Name && meta.IsStatusConditionTrue(wl.Status.Conditions, kueue.WorkloadQuotaReserved) && meta.IsStatusConditionTrue(wl.Status.Conditions, kueue.WorkloadEvicted) {
+					// The first request evicting wl1 will set both the quota reservation and eviction, as this status is assumed in the cache.
+					// We sleep here in order to lag the next scheduling cycle.
+					time.Sleep(time.Second * 3)
```
kshalot: Actually, I think we can even replace this sleep with some channels

mimowo: Maybe but I think isn't really necessary

test/integration/singlecluster/scheduler/scheduler_test.go
```diff
@@ -3257,6 +3257,146 @@ var _ = ginkgo.Describe("Scheduler", func() {
 			}, util.ConsistentDuration, util.ShortInterval).Should(gomega.Succeed())
 		})
 	})
+
+	ginkgo.When("The admission goroutine races with preemptions in the next scheduling cycle", func() {
+		// This deterministically reproduces a race condition in the scheduler, where the asynchronous
+		// admission goroutine does not finish before the next scheduling cycle starts.
+		// Reverting https://github.com/kubernetes-sigs/kueue/pull/11502 makes this test fail.
+		// See: https://github.com/kubernetes-sigs/kueue/issues/11480
```
mimowo, line 3263: I would suggest to also provide a detailed description of the scenario (ie. the timline-like of the requests, and commonents issueing them) .It took us quite some time to figure out the scenario so let's record it if someone ever needs to understand this again it will be super useful.

Then, please add also a note why the fix in https://github.com/kubernetes-sigs/kueue/pull/11502 breaks the scenario as if the test was part of that PR and we want to explain to the reader why the scenario is no longer problematic. 

Don't worry about making comment 3x longer, I think it is still a time saver to read it vs debug on your own.

test/integration/singlecluster/scheduler/scheduler_test.go
kshalot: It would look something like this: https://github.com/kubernetes-sigs/kueue/pull/12331/changes/03cb40eb7a16e2b6527b24e656a2a45352307196

I think it's more self-documenting, but it can be potentially flaky (unlikely but possible) if the informer is not updated but the poll to check the status of the workload already passes.

test/integration/singlecluster/scheduler/scheduler_test.go
kshalot: Yes, good point.

WDYT? https://github.com/kubernetes-sigs/kueue/pull/12331/changes/89cb6ae31c98fd2a9387cc9c4c0430ac80dbacfd

mimowo: In general it looks ok, but I would also like to correlate the points to the code (the channels). In particular some questions worth clarifying
1. how `<-continueSchedulingAfterAdmissionPatch` defers the scheduling cycles - this is non-obvious because that is doing at the code level is just deferring the Eviction request for wl1
2. How  is the `quota releasing for wll1 fitting into  the picture. For example iiuc it is important that it is deferred beyond point (8.) so because otherwise it would re-add the wl1 in to the heap again and so remove the preemptionExpectations

kshalot: https://github.com/kubernetes-sigs/kueue/pull/12331/changes/33d85c449116c98dbedfe3d7954bc24e07fec8b0 <- I added comments that describe what blocks where.

It's a little hard to describe this in a textual comment. We can consider adding a link to a mermaid sequence diagram at some point if necessary.

mimowo: image
No need, I followed along with the caveman style :) . Thank you for explaining. I also tied this to the code and it makes perfect sense. I think this level of detail is really needed to build the mental model of the complex interactions.

mimowo:
I think this is probably one of the most complex race conditions we had in Kueue so far, so awesome work in documenting the scenario, wholeheartedly 🥇

Also,  I think it is possible we may return to such deliberations if we extend one day the preemptionExpectations to cover other types of evictions or defrag migrations. 

/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17

mimowo:
That one is weird, all tests passed green: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12331/pull-kueue-test-integration-shard-1-main/2070232659472158720

I'm wondering if this is maybe due to some secondary checks like for leaking goroutines, which makes me wonder - maybe we should use ginkgo.DeferCleanup() rather than "defer".

Let me hold for a couple tries at least, becasue I've never seen this before, and the newly added test was part of the failed suite
/hold

kshalot:
> Let me hold for a couple tries at least, becasue I've never seen this before, and the newly added test was part of the failed suite

AFAICS it's a data race caused by setting the interceptor functions:

```
  WARNING: DATA RACE
  Write at 0x000006057ea8 by goroutine 9216:
    sigs.k8s.io/kueue/test/integration/singlecluster/scheduler.init.func6()
        /home/prow/go/src/sigs.k8s.io/kueue/test/integration/singlecluster/scheduler/suite_test.go:86 +0x53
    github.com/onsi/ginkgo/v2/internal.extractBodyFunction.func3()
        /home/prow/go/src/sigs.k8s.io/kueue/vendor/github.com/onsi/ginkgo/v2/internal/node.go:585 +0x2e
    github.com/onsi/ginkgo/v2/internal.(*Suite).runNode.func3()
        /home/prow/go/src/sigs.k8s.io/kueue/vendor/github.com/onsi/ginkgo/v2/internal/suite.go:946 +0x699

  Previous read at 0x000006057ea8 by goroutine 607:
    sigs.k8s.io/kueue/test/integration/singlecluster/scheduler.newInterceptedClient.fakeSubResourcePatchFrom.func1()
        /home/prow/go/src/sigs.k8s.io/kueue/test/integration/singlecluster/scheduler/suite_test.go:172 +0x197
    sigs.k8s.io/controller-runtime/pkg/client/interceptor.subResourceInterceptor.Patch()
        /home/prow/go/src/sigs.k8s.io/kueue/vendor/sigs.k8s.io/controller-runtime/pkg/client/interceptor/intercept.go:173 +0xfd
    sigs.k8s.io/controller-runtime/pkg/client/interceptor.(*subResourceInterceptor).Patch()
        <autogenerated>:1 +0x1c4
    sigs.k8s.io/kueue/pkg/workload.patchStatus()
        /home/prow/go/src/sigs.k8s.io/kueue/pkg/workload/workload.go:1298 +0x4e6
    sigs.k8s.io/kueue/pkg/workload.PatchAdmissionStatus()
        /home/prow/go/src/sigs.k8s.io/kueue/pkg/workload/workload.go:1320 +0x145
    sigs.k8s.io/kueue/pkg/controller/core.(*WorkloadReconciler).Reconcile()
        /home/prow/go/src/sigs.k8s.io/kueue/pkg/controller/core/workload_controller.go:740 +0x68a4
    sigs.k8s.io/kueue/pkg/controller/core.(*WorkloadReconciler).Reconcile()
        /home/prow/go/src/sigs.k8s.io/kueue/pkg/controller/core/workload_controller.go:618 +0x455d
    sigs.k8s.io/kueue/pkg/controller/core.(*WorkloadReconciler).Reconcile()
        /home/prow/go/src/sigs.k8s.io/kueue/pkg/controller/core/workload_controller.go:233 +0x171
    sigs.k8s.io/kueue/pkg/controller/core.(*leaderAwareReconciler).Reconcile()
    ...
 ```
 
 My guess is that maybe the subresource patch is executed in the workload controller, while simultaneously we `nil` the interceptor in `BeforeEach`:
 https://github.com/kubernetes-sigs/kueue/blob/8ea8b61e68788f42509f36da212c4b61840aafb4/test/integration/singlecluster/scheduler/suite_test.go#L82-L84
 
 If I'm right, then:
 1. This isn't necessarily caused by this PR. But maybe this PR makes it easier to trigger this race, although I'm not certain.
 2. We can fix it by adding a mutex on the BeforeEach and in the interceptor here:
 https://github.com/kubernetes-sigs/kueue/blob/8ea8b61e68788f42509f36da212c4b61840aafb4/test/integration/singlecluster/scheduler/suite_test.go#L155-L173

**Comment by mimowo** [comment]
Please check - I've never seen this error before so I assume the PR indeed makes it more likely to happen. Maybe try looping the test locally 50 times or so to see if you can repro, and if the problem goes away after the mutex

**Comment by kshalot** [comment]
> Please check - I've never seen this error before so I assume the PR indeed makes it more likely to happen. Maybe try looping the test locally 50 times or so to see if you can repro, and if the problem goes away after the mutex

Yep, I'm checking. I pushed https://github.com/kubernetes-sigs/kueue/pull/12331/changes/c548989c584bc453ddd81163a806267863a84533 already, but let's hold to see whether I can repro this

**Comment by kshalot** [comment]
@mimowo

1. I was able to trigger this on this PR without https://github.com/kubernetes-sigs/kueue/commit/c548989c584bc453ddd81163a806267863a84533.
2. After https://github.com/kubernetes-sigs/kueue/commit/c548989c584bc453ddd81163a806267863a84533 I wasn't able to trigger this.
3. I wasn't able to trigger this on main.

This is somewhat expected, because this PR originally added:
```go
		case "status":
			response := baseK8sClient.Status().Patch(ctx, obj, patch, opts...) <- SLOW
			if *g != nil { <- DATA RACE
				fakeUsage, err := (*g)(obj, response)
				if fakeUsage == emitResponse {
					return err
				}
			}
			return response
```
The patch is relatively slow, which gives more time for overlapping changes to `g` before `*g != nil` is called. On main a data race can still happen [here](https://github.com/kubernetes-sigs/kueue/blob/42fc89c89364afad6dc64112ef8bb3d0c60389ca/test/integration/singlecluster/scheduler/suite_test.go#L160), but the chances of that happening are extremely low. That's why it never popped up.

As an alternative "fix", we could just add smth like `gSpec = *g` at the beginning of this function, but that is still _technically_ a data race, just less likely. I'd keep https://github.com/kubernetes-sigs/kueue/commit/c548989c584bc453ddd81163a806267863a84533 to make this robust.

mimowo:
/unhold
/lgtm
/cherrypick release-0.18
/cherrypick release-0.17
Please make sure it gets cherrypicked in case the robot fails

kshalot:
I guess the failures should be resolved by https://github.com/kubernetes-sigs/kueue/commit/7b5e7de0bd5080f03d20949eb3330e30b6654dc7?

/retest

mimowo:
/retest
Due to kuberay temp outage

mimowo:
/retest
Due to kuberay temp outage

PR #12332, opened 2026-06-18
Extract preemption gate helpers into pkg/workload and reuse them in t…

reruno:
…he MultiKueue controller

#### What type of PR is this?
/kind cleanup

Extract preemption gate helpers into pkg/workload and reuse them in the MultiKueue controller

Related to #11872 

```release-note
NONE
```

## Summary by CodeRabbit

* **Refactor**
  * Improved preemption gate handling in MultiKueue with enhanced reliability and fairness in determining which workloads require preemption. System now uses more robust detection mechanisms and selects candidates based on the oldest preemption signals for fairer scheduling decisions.

### DIFF
--- a/pkg/workload/workload.go
+++ b/pkg/workload/workload.go
@@ -1086,6 +1086,51 @@ func SetPreemptionGatePosition(w *kueue.Workload, gateName string, gatePosition 
 	return true
 }
 
+// Finds preemption gate with gateName, and returns pointer to that preemption gate state object
+func FindPreemptionGate(w *kueue.Workload, gateName string) *kueue.PreemptionGateState {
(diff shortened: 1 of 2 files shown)


mimowo:
/lgtm 
/approve
Thank you for the preparatory PR

pkg/workload/workload.go
```diff
@@ -1086,6 +1086,51 @@ func SetPreemptionGatePosition(w *kueue.Workload, gateName string, gatePosition
 	return true
 }
 
+// Finds preemption gate with gateName, and returns pointer to that preemption gate state object
```
kshalot, line 1089: This is a nit, but AFAICS the other function comments are formatted as "<function name> <what it does>". So here, <function name> is missing.

No point in aligning that though IMO

PR #12334, opened 2026-06-18
bug: support for pod level resources

### LINKED ISSUE #4233: Support PodLevelResources
It would be great if we can support [the PodLevelResources feature](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/2837-pod-level-resource-spec/kep.yaml). Indeed, this feature is still alpha stage, but the APIs has already been added to k/k corev1 API.

Indeed, we have already started to use the [`PodRequests`](https://github.com/kubernetes/kubernetes/blob/69ab91a5c59617872c9f48737c64409a9dec2957/staging/src/k8s.io/component-helpers/resource/helpers.go#L120) helper to calculate total Pod resources in https://github.com/kubernetes-sigs/kueue/pull/4177, which means we can calculate the requested Pod resources with PodLevelResources by passing something to the `PodResourceOptions`.

Once we have done the below tasks, we can confirm the PodLevelResources compatibility with Kueue.

- [ ] Add Unit / Integration testings by using Pod with `.spec.resources` (≠ `.spec.containers.resources`)
- [ ] Refine the total Pod resources calculation mechanism to support `.spec.resources` in the following:
    - https://github.com/kubernetes-sigs/kueue/blob/2738e8b05c03fe5dcf38beba1f4769fe2e1888c9/pkg/cache/tas_flavor.go
    - https://github.com/kubernetes-sigs/kueue/blob/main/pkg/util/limitrange/limitrange.go
    - https://github.com/kubernetes-sigs/kueue/blob/main/pkg/workload/workload.go

> [!NOTE]
> We need to enable Kubernetes featureGate, `PodLevelResources` in kube-scheduler, kube-apiseerver, and kubelet to verify the behavior.

**Why is this needed**:
It would be better to support k/k new enhancements.

**Completion requirements**:

- [x] Design doc
- [ ] API change
- [x] Docs update

anuragdalvi:
#### What type of PR is this?
/kind bug

/area website

This PR adds support for pod-level resources ([KEP-2837](https://kubernetes.io/docs/concepts/workloads/pods/#pod-level-resources), GA-track on Kubernetes 1.32+), where a pod declares CPU/memory requests and limits at the pod level via pod.spec.resources instead of, or in addition to, per-container requests. Previously Kueue only looked at container-level resources, so pod-level declarations were ignored in role hashing, request adjustment, and validation. This PR makes Kueue's workload resource handling pod-level-aware across all four of those paths:

Role hashing `pod.SpecShape` : pod-level requests now contribute to the pod-spec shape used to compute role hashes. They are only added to the shape when podSpec.Resources is non-nil, so role hashes for pods without pod-level resources stay stable across upgrades.
LimitRange defaulting `handlePodLimitRange` : LimitRange Default/DefaultRequest values are now also merged into pod.spec.resources, mirroring the existing container behavior.
Limits-as-missing-requests `UseLimitsAsMissingRequestsInPod`: when only pod-level limits are set, they are treated as the pod-level requests, consistent with how container limits are handled.
Validation `ValidateResources` : pod-level requests are validated not to exceed pod-level limits, producing the same requests must not exceed its limits error (scoped to spec.podSets[*].template.spec.resources) already emitted for containers.
All pod-level handling is guarded by an if pod.Resources != nil check, since the field is an optional pointer only populated when the cluster's PodLevelResources feature gate is enabled and used — so behavior is unchanged for clusters not using the feature.

Also adds PodLevelRequest/PodLevelLimit test wrappers, unit tests for the new paths, and documents pod-level resource accounting in the Workload concepts page.

Fixes [#4233](https://github.com/kubernetes-sigs/kueue/issues/4233)

- Not sure if adding limits to `SpecShape()` or `ContainersShape` function would result in fragmentation of multiple PodSets, so did not add it in this PR. 
- Will add Integration tests in a follow up PR. 

```release-note
Scheduling: Fixed resource accounting and validation for Pods using Kubernetes pod-level
resources (`pod.spec.resources`), including LimitRange defaulting and request/limit
validation.
```

## Summary by CodeRabbit

* **New Features**
  * Added Kubernetes 1.32+ pod-level resources (`pod.spec.resources`) support for request/limit adjustment, validation, and quota accounting.
* **Bug Fixes**
  * Ensure stable role hashing when pod-level resources are omitted.
  * Apply LimitRange defaults for pod-type limits, adjust pod requests from pod limits, and validate pod-level requests vs pod-level limits.
* **Tests**
  * Expanded unit, integration, and end-to-end coverage for pod-level quota and gating behavior.
* **Documentation**
  * Updated workload docs with “Pod-level resources” guidance and validation details.

### DIFF
--- a/pkg/controller/core/indexer/indexer.go
+++ b/pkg/controller/core/indexer/indexer.go
@@ -38,7 +38,7 @@ const (
 	WorkloadQueueKey                     = "spec.queueName"
 	WorkloadClusterQueueKey              = "status.admission.clusterQueue"
 	QueueClusterQueueKey                 = "spec.clusterQueue"
-	LimitRangeHasContainerType           = "spec.hasContainerType"
+	LimitRangeHasContainerOrPodType      = "spec.hasContainerOrPodType"
 	WorkloadQuotaReservedKey             = "status.quotaReserved"
 	WorkloadRuntimeClassKey              = "spec.runtimeClass"
 	OwnerReferenceUID                    = "metadata.ownerReferences.uid"
@@ -126,14 +126,14 @@ func IndexWorkloadClusterQueue(obj client.Object) []string {
 	return []string{string(wl.Status.Admission.ClusterQueue)}
 }
 
-func IndexLimitRangeHasContainerType(obj client.Object) []string {
+func IndexLimitRangeHasContainerOrPodType(obj client.Object) []string {
 	lr, ok := obj.(*corev1.LimitRange)
 	if !ok {
 		return nil
 	}
-
 	for i := range lr.Spec.Limits {
-		if lr.Spec.Limits[i].Type == corev1.LimitTypeContainer {
+		t := lr.Spec.Limits[i].Type
+		if t == corev1.LimitTypeContainer || t == corev1.LimitTypePod {
 			return []string{"true"}
 		}
 	}
@@ -277,8 +277,8 @@ func Setup(ctx context.Context, indexer client.FieldIndexer) error {
 	if err := indexer.IndexField(ctx, &kueue.LocalQueue{}, QueueClusterQueueKey, IndexQueueClusterQueue); err != nil {
 		return fmt.Errorf("setting index on clusterQueue for localQueue: %w", err)
 	}
-	if err := indexer.IndexField(ctx, &corev1.LimitRange{}, LimitRangeHasContainerType, IndexLimitRangeHasContainerType); err != nil {
-		return fmt.Errorf("setting index on hasContainerType for limitRange: %w", err)
+	if err := indexer.IndexField(ctx, &corev1.LimitRange{}, LimitRangeHasContainerOrPodType, IndexLimitRangeHasContainerOrPodType); err != nil {
+		return fmt.Errorf("setting index on hasContainerOrPodType for limitRange: %w", err)
 	}
 	if err := indexer.IndexField(ctx, &kueue.Workload{}, OwnerReferenceUID, IndexOwnerUID); err != nil {
 		return fmt.Errorf("setting index on ownerReferences.uid for Workload: %w", err)
--- a/pkg/util/pod/pod_test.go
+++ b/pkg/util/pod/pod_test.go
@@ -23,6 +23,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	testingpod "sigs.k8s.io/kueue/pkg/util/testingjobs/pod"
@@ -247,6 +248,79 @@ func TestIsTerminated(t *testing.T) {
 			got := IsTerminated(tc.pod)
 			if tc.wantTerminated != got {
 				t.Errorf("Unexpected Pod terminal\nwant: %v\ngot: %v\n", tc.wantTerminated, got)
+			}
+		})
+	}
+}
+
+func TestSpecShape(t *testing.T) {
+	podResources := &corev1.ResourceRequirements{
+		Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("1")},
+		Limits:   corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("2")},
+	}
+	cases := map[string]struct {
+		podSpec       *corev1.PodSpec
+		wantResources any
+		wantPresent   bool
+	}{
+		"pod-level resources omitted when unset": {
+			podSpec:     &corev1.PodSpec{Containers: []corev1.Container{{Name: "c"}}},
+			wantPresent: false,
+		},
+		"pod-level resources included when set": {
+			podSpec:       &corev1.PodSpec{Containers: []corev1.Container{{Name: "c"}}, Resources: podResources},
+			wantResources: podResources.Requests,
+			wantPresent:   true,
+		},
+	}
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			got, present := SpecShape(tc.podSpec)["resources"]
+			if present != tc.wantPresent {
+				t.Fatalf("Unexpected presence of \"resources\" key: want %v, got %v", tc.wantPresent, present)
+			}
+			if diff := cmp.Diff(tc.wantResources, got); diff != "" {
+				t.Errorf("Unexpected resources shape (-want,+got):\n%s", diff)
+			}
+		})
+	}
+}
+
+func TestGenerateRoleHash(t *testing.T) {
+	withoutPodResources := &corev1.PodSpec{Containers: []corev1.Container{{Name: "c"}}}
+	withPodResources := &corev1.PodSpec{
+		Containers: []corev1.Container{{Name: "c"}},
+		Resources: &corev1.ResourceRequirements{
+			Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("1")},
+		},
+	}
+
+	baseHash, err := GenerateRoleHash(withoutPodResources)
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+
+	cases := map[string]struct {
+		podSpec      *corev1.PodSpec
+		wantBaseHash bool
+	}{
+		"pod-level resources change the role hash": {
+			podSpec: withPodResources,
+		},
+		"spec without pod-level resources has a stable role hash": {
+			podSpec:      withoutPodResources.DeepCopy(),
+			wantBaseHash: true,
+		},
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			gotHash, err := GenerateRoleHash(tc.podSpec)
+			if err != nil {
+				t.Fatalf("Unexpected error: %v", err)
+			}
+			if gotBaseHash := gotHash == baseHash; gotBaseHash != tc.wantBaseHash {
+				t.Errorf("Unexpected role hash comparison with base hash: want equality %t, base hash %q, got hash %q", tc.wantBaseHash, baseHash, gotHash)
 			}
 		})
 	}
--- a/pkg/util/testing/v1beta2/wrappers.go
+++ b/pkg/util/testing/v1beta2/wrappers.go
@@ -587,6 +587,28 @@ func (p *PodSetWrapper) Limit(r corev1.ResourceName, q string) *PodSetWrapper {
 	return p
 }
 
+func (p *PodSetWrapper) PodLevelRequest(r corev1.ResourceName, q string) *PodSetWrapper {
+	if p.Template.Spec.Resources == nil {
+		p.Template.Spec.Resources = &corev1.ResourceRequirements{}
+	}
+	if p.Template.Spec.Resources.Requests == nil {
+		p.Template.Spec.Resources.Requests = corev1.ResourceList{}
+	}
+	p.Template.Spec.Resources.Requests[r] = resource.MustParse(q)
+	return p
+}
+
+func (p *PodSetWrapper) PodLevelLimit(r corev1.ResourceName, q string) *PodSetWrapper {
+	if p.Template.Spec.Resources == nil {
+		p.Template.Spec.Resources = &corev1.ResourceRequirements{}
+	}
+	if p.Template.Spec.Resources.Limits == nil {
+		p.Template.Spec.Resources.Limits = corev1.ResourceList{}
+	}
+	p.Template.Spec.Resources.Limits[r] = resource.MustParse(q)
+	return p
+}
+
 func (p *PodSetWrapper) Image(image string) *PodSetWrapper {
 	p.Template.Spec.Containers[0].Image = image
 	return p
--- a/pkg/workload/resources.go
+++ b/pkg/workload/resources.go
@@ -67,7 +67,7 @@ func handlePodOverhead(ctx context.Context, cl client.Client, wl *kueue.Workload
 func handlePodLimitRange(ctx context.Context, cl client.Client, wl *kueue.Workload) error {
 	// get the list of limit ranges
 	var limitRanges corev1.LimitRangeList
-	if err := cl.List(ctx, &limitRanges, &client.ListOptions{Namespace: wl.Namespace}, client.MatchingFields{indexer.LimitRangeHasContainerType: "true"}); err != nil {
+	if err := cl.List(ctx, &limitRanges, &client.ListOptions{Namespace: wl.Namespace}, client.MatchingFields{indexer.LimitRangeHasContainerOrPodType: "true"}); err != nil {
 		return err
 	}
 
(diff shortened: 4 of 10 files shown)


sohankunkerkar:
Thanks for the PR. I'd encourage you to open a KEP PR first to discuss the design.

kannon92:
> Thanks for the PR. I'd encourage you to open a KEP PR first to discuss the design.

I don't think a KEP is really needed.

https://github.com/kubernetes-sigs/kueue/issues/4233

Turns out that we have support for this already but there are a few bugs we have to address.

kannon92:
@coderabbitai review

sohankunkerkar:
> > Thanks for the PR. I'd encourage you to open a KEP PR first to discuss the design.
> 
> I don't think a KEP is really needed.
> 
> #4233
> 
> Turns out that we have support for this already but there are a few bugs we have to address.

I see, so the label is misleading here. It should be `/kind bug`?

anuragdalvi:
Thanks, changed label to /kind bug.

kannon92:
> Will add Integration tests in a follow up PR.

I would prefer integration tests in this PR actually.

mimowo:
Yes, for an integration tests I think it makes sense to cover a scanario when the first Pod is scheduled, but for the second we don't have enough quota (due to the PodLevel resources). 

I think we don't need e2e tests for the feature.

mimowo:
injected a proposed release note, since this is certainly a bug

pkg/util/pod/pod_test.go
```diff
@@ -251,3 +252,70 @@ func TestIsTerminated(t *testing.T) {
...
+				t.Fatalf("Unexpected presence of \"resources\" key: want %v, got %v", tc.wantPresent, present)
+			}
+			if diff := cmp.Diff(tc.wantResources, got); diff != "" {
+				t.Errorf("Unexpected resources shape (-want,+got):\n%s", diff)
+			}
+		})
+	}
+}
+
+func TestGenerateRoleHash(t *testing.T) {
```
kannon92, line 289: In pkg/util/pod/pod_test.go, the new TestGenerateRoleHash function covers multiple test scenarios inline as a sequential block (base hash, changed pod-resources hash, identically stable empty hash). Please refactor this to use table-driven style (cases := map[string]struct{...})  
 to match established patterns and make adding future cases easier.

pkg/controller/core/indexer/indexer.go
```diff
@@ -277,8 +291,8 @@ func Setup(ctx context.Context, indexer client.FieldIndexer) error {
 	if err := indexer.IndexField(ctx, &kueue.LocalQueue{}, QueueClusterQueueKey, IndexQueueClusterQueue); err != nil {
 		return fmt.Errorf("setting index on clusterQueue for localQueue: %w", err)
 	}
-	if err := indexer.IndexField(ctx, &corev1.LimitRange{}, LimitRangeHasContainerType, IndexLimitRangeHasContainerType); err != nil {
```
kannon92, line 280: With this change, I think LimitRangeHasContainerType is no longer used. 

Can we remove this function then?

pkg/util/testing/v1beta2/wrappers.go
```diff
@@ -1536,10 +1558,12 @@ type WorkloadPriorityClassWrapper struct {
 
 // MakeWorkloadPriorityClass creates a wrapper for a WorkloadPriorityClass.
 func MakeWorkloadPriorityClass(name string) *WorkloadPriorityClassWrapper {
```
kannon92, line 1566: Can you revert these changes?

anuragdalvi: Apologies—this was an unintended change from running gofumpt. I’ve reverted it.

pkg/controller/core/indexer/indexer.go
anuragdalvi: Done—removed `LimitRangeHasContainerType` and updated the tests to cover the replacement function.

test/e2e/singlecluster/baseline/pod_test.go
```diff
@@ -140,6 +141,64 @@ var _ = ginkgo.Describe("Pod groups", ginkgo.Label("area:singlecluster", "featur
 			})
 		})
 
+		ginkgo.It("should keep a second Pod gated when Pod-level resources exhaust quota", func() {
```
mimowo: I think we don't really need the e2e test, because this feature does not depend on k8s core controllers in a direct way. It only matters what we put into the quota requests.

anuragdalvi: understood. I have it removed.

test/integration/singlecluster/scheduler/scheduler_test.go
```diff
@@ -2258,6 +2258,181 @@ var _ = ginkgo.Describe("Scheduler", func() {
...
+			// Container-level: req=500m, limit=500m (valid on their own). Pod-level: req=2 < min=3 fails.
+			ginkgo.Entry(
+				"pod-level requests under pod LimitRange min",
+				testParams{containerReqCPU: "500m", containerLimitCPU: "500m", podReqCPU: "2", podLimitCPU: "3", minCPU: "3", wantedStatus: "resources didn't satisfy LimitRange constraints:"},
+			),
+			// Container-level: req=500m, limit=500m. Pod-level: req=2, limit=3, within [min=1, max=4] → admitted.
+			ginkgo.Entry(
+				"valid pod-level resources supersede container-level",
+				testParams{containerReqCPU: "500m", containerLimitCPU: "500m", podReqCPU: "2", podLimitCPU: "3", minCPU: "1", maxCPU: "4", shouldBeAdmitted: true},
+			),
```
mimowo: nit, but could you unwind the test cases. Making them programmatic requires sometimes awkward options to them. In my practice table driven tests are great for unit tests, but are hard to follow in the integration context.

test/integration/singlecluster/scheduler/scheduler_test.go
```diff
@@ -2258,6 +2258,181 @@ var _ = ginkgo.Describe("Scheduler", func() {
...
+				gomega.Expect(util.DeleteObject(ctx, k8sClient, wl2)).To(gomega.Succeed())
+			}()
+
+			ginkgo.By("verifying the second workload is pending due to insufficient quota")
+			util.ExpectWorkloadsToBePending(ctx, k8sClient, wl2)
+			createdWl2 := kueue.Workload{}
+			gomega.Consistently(func(g gomega.Gomega) {
+				g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(wl2), &createdWl2)).Should(gomega.Succeed())
+				g.Expect(workload.HasQuotaReservation(&createdWl2)).Should(gomega.BeFalse())
+			}, util.ConsistentDuration, util.ShortInterval).Should(gomega.Succeed())
```
mimowo: Instead of using ConsistentDuration you can check the metric for kueue_pending_worklaods {inadmissible=1}, that would save 1s here, and in other cases I suppose.

test/integration/singlecluster/scheduler/scheduler_test.go
anuragdalvi: Should I do the same for this test as well ?
https://github.com/anuragdalvi/kueue/blob/0254c8e6f034e7d306f75ff9cb6c90293b92db72/test/integration/singlecluster/scheduler/scheduler_test.go#L2173

test/integration/singlecluster/scheduler/workload_controller_test.go
```diff
@@ -746,6 +791,67 @@ var _ = ginkgo.Describe("Workload controller with scheduler", func() {
...
+							Obj(),
+					).
+					Obj()
+				util.MustCreate(ctx, k8sClient, wl2)
+
+				createdWl2 := kueue.Workload{}
+				gomega.Consistently(func(g gomega.Gomega) {
+					g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(wl2), &createdWl2)).Should(gomega.Succeed())
+					g.Expect(workload.HasQuotaReservation(&createdWl2)).Should(gomega.BeFalse())
+				}, util.ConsistentDuration, util.ShortInterval).Should(gomega.Succeed())
```
mimowo: Let's use `util.ExpectPendingWorkloadsMetric(cq, 0, 1)` . It might be a follow up, but it really does not make sense to wait 1s

test/integration/singlecluster/scheduler/scheduler_test.go
```diff
@@ -2258,6 +2258,215 @@ var _ = ginkgo.Describe("Scheduler", func() {
...
+						Limit(corev1.ResourceCPU, "500m").
+						PodLevelRequest(corev1.ResourceCPU, "2").
+						PodLevelLimit(corev1.ResourceCPU, "2").
+						Obj(),
+				).
+				Obj()
+			util.MustCreate(ctx, k8sClient, wl2)
+			defer func() {
+				gomega.Expect(util.DeleteObject(ctx, k8sClient, wl2)).To(gomega.Succeed())
+			}()
```
mimowo: nit, Instead normally we use the helper to delete all workloads and wait for their deletion in AfterEach.

test/integration/singlecluster/scheduler/scheduler_test.go
```diff
@@ -2258,6 +2258,215 @@ var _ = ginkgo.Describe("Scheduler", func() {
...
+						Limit(corev1.ResourceCPU, "500m").
+						PodLevelRequest(corev1.ResourceCPU, "2").
+						PodLevelLimit(corev1.ResourceCPU, "2").
+						Obj(),
+				).
+				Obj()
+			util.MustCreate(ctx, k8sClient, wl1)
+			defer func() {
+				gomega.Expect(util.DeleteObject(ctx, k8sClient, wl1)).To(gomega.Succeed())
+			}()
```
mimowo: Here also, we can rely on the helper to wait for deletion of all workloads: util.DeleteWorkloadsInNamespace (same in other places, just for readability)

mimowo:
LGTM overall, left some rather nit comments.

mimowo:
@anuragdalvi do you have cycles to address the remaining comments?

pkg/workload/resources.go
```diff
@@ -67,30 +67,39 @@ func handlePodOverhead(ctx context.Context, cl client.Client, wl *kueue.Workload
...
+			}
 		}
-		for ci := range pod.Containers {
-			res := &pod.Containers[ci].Resources
-			res.Limits = resource.MergeResourceListKeepFirst(res.Limits, containerLimits.Default)
-			res.Requests = resource.MergeResourceListKeepFirst(res.Requests, containerLimits.DefaultRequest)
+		// Pod-level resources (KEP-2837) are an optional pointer, only set when the
+		// PodLevelResources feature is enabled and used.
+		if pod.Resources != nil && foundPodLimits {
+			pod.Resources.Limits = resource.MergeResourceListKeepFirst(pod.Resources.Limits, podLimits.Default)
```
mimowo, line 101: sanity check: is the new code fully covered by unit tests?

test/integration/singlecluster/scheduler/scheduler_test.go
```diff
@@ -2258,6 +2258,211 @@ var _ = ginkgo.Describe("Scheduler", func() {
...
+
+		ginkgo.It("blocks workloads when pod-level requests exceed the pod LimitRange max", func() {
+			lr := utiltesting.MakeLimitRange("pod-max-limit", ns.Name).
+				WithType(corev1.LimitTypePod).
+				WithValue("Max", corev1.ResourceCPU, "1").
+				Obj()
+			util.MustCreate(ctx, k8sClient, lr)
+			defer func() {
+				gomega.Expect(k8sClient.Delete(ctx, lr)).To(gomega.Succeed())
+			}()
```
mimowo: Move the cleanups to AfterEach consistently. In particular use DeleteWorkloadsInNamespace to cleanup worklaods, or DeleteNamespace to cleanup all namspecad resources.

pkg/workload/resources.go
anuragdalvi: Yes, were missing Added tests.

kannon92:
can you squash your commits

test/integration/singlecluster/scheduler/scheduler_test.go
```diff
@@ -2258,6 +2258,202 @@ var _ = ginkgo.Describe("Scheduler", func() {
...
+						PodLevelRequest(corev1.ResourceCPU, "3").
+						PodLevelLimit(corev1.ResourceCPU, "2").
+						Obj(),
+				).
+				Obj()
+			util.MustCreate(ctx, k8sClient, wl)
+			defer func() {
+				gomega.Expect(util.DeleteObject(ctx, k8sClient, wl)).To(gomega.Succeed())
+			}()
+
```
mimowo: that should not be needed after we added `gomega.Expect(util.DeleteWorkloadsInNamespace(ctx, k8sClient, ns)).Should(gomega.Succeed())` in AfterEach - seme for other places. Let me know if there are some unexpected issues with that. 

It is ok-ish to keep them both, but I would like to cleanup the tests as we often maintain them long term

test/integration/singlecluster/scheduler/scheduler_test.go
```diff
@@ -2258,6 +2258,185 @@ var _ = ginkgo.Describe("Scheduler", func() {
...
+					*utiltestingapi.MakeFlavorQuotas("on-demand").Resource(corev1.ResourceCPU, "5").Obj(),
+				).
+				Obj()
+			util.MustCreate(ctx, k8sClient, cq)
+			queue = utiltestingapi.MakeLocalQueue("queue", ns.Name).ClusterQueue(cq.Name).Obj()
+			util.MustCreate(ctx, k8sClient, queue)
+		})
+
+		ginkgo.AfterEach(func() {
+			gomega.Expect(util.DeleteWorkloadsInNamespace(ctx, k8sClient, ns)).Should(gomega.Succeed())
```
mimowo, line 2279: Call DeleteNamespace actually, to make sure that also LQs are deleted (same for all other AfterEach blocks)

mimowo: Ahhh, this is already covered here: https://github.com/anuragdalvi/kueue/blob/support-pod-level-resources/test/integration/singlecluster/scheduler/scheduler_test.go#L103-L105

mimowo:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12334/pull-kueue-test-integration-shard-1-main/2072671060314034176
This looks like a related flake

mimowo:
PLease run the new tests locally looped 50 times or so to make sure we have no flakes

test/integration/singlecluster/scheduler/scheduler_test.go
anuragdalvi: Fixed — duplicate LocalQueue creation in one test, missing creation in another. Both now match the existing pattern.

anuragdalvi:
Thanks for the nudge. I ran the new tests locally 50 times, and they passed consistently with no flakes observed.

```
 make test-integration \
  INTEGRATION_TARGET=./test/integration/singlecluster/scheduler/... \
  GINKGO_ARGS="--repeat=49"
```

mimowo (commented):
/lgtm
/cherrypick release-0.18
/cherrypick release-0.17
Thank you, please cherrypick in case the robot fails on conflicts 👍

mimowo:
@anuragdalvi do you have cycles to prepare the CP for 0.17?

anuragdalvi:
@mimowo prepared CP for 0.17: https://github.com/kubernetes-sigs/kueue/pull/12780

It targets release-0.17 and cherry-picks #12334. The bot conflict was resolved in the PR.

tenzen-y:
/release-note-edit

```release-note
Scheduling: Fixed resource accounting and validation for Pods using Kubernetes pod-level
resources (`pod.spec.resources`), including LimitRange defaulting and request/limit
validation.
```

Issue #12307, opened 2026-06-17
Flaky E2E: WAS: unable to create ClusterRoleBinding: context deadline exceeded


mbobrovskyi:
**Which test is flaking?**:
error: error execution phase wait-control-plane: cannot obtain client without bootstrap: could not bootstrap the admin user in file admin.conf: unable to create ClusterRoleBinding: context deadline exceeded

**Link to failed CI job or steps to reproduce locally**:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/periodic-kueue-test-e2e-k8s-main-was/2067217561786781696

**Failure message or logs**:
```
I0617 12:16:21.232532     208 round_trippers.go:632] "Response" verb="POST" url="https://172.18.0.3:6443/apis/rbac.authorization.k8s.io/v1/clusterrolebindings?timeout=10s" status="" milliseconds=0
I0617 12:16:21.732815     208 round_trippers.go:632] "Response" verb="POST" url="https://172.18.0.3:6443/apis/rbac.authorization.k8s.io/v1/clusterrolebindings?timeout=10s" status="" milliseconds=0
I0617 12:16:22.233049     208 round_trippers.go:632] "Response" verb="POST" url="https://172.18.0.3:6443/apis/rbac.authorization.k8s.io/v1/clusterrolebindings?timeout=10s" status="" milliseconds=0
I0617 12:16:22.733369     208 round_trippers.go:632] "Response" verb="POST" url="https://172.18.0.3:6443/apis/rbac.authorization.k8s.io/v1/clusterrolebindings?timeout=10s" status="" milliseconds=0
I0617 12:16:23.232586     208 round_trippers.go:632] "Response" verb="POST" url="https://172.18.0.3:6443/apis/rbac.authorization.k8s.io/v1/clusterrolebindings?timeout=10s" status="" milliseconds=0
I0617 12:16:23.733494     208 round_trippers.go:632] "Response" verb="POST" url="https://172.18.0.3:6443/apis/rbac.authorization.k8s.io/v1/clusterrolebindings?timeout=10s" status="" milliseconds=0
I0617 12:16:24.232763     208 round_trippers.go:632] "Response" verb="POST" url="https://172.18.0.3:6443/apis/rbac.authorization.k8s.io/v1/clusterrolebindings?timeout=10s" status="" milliseconds=0
[...]

PR #11586, opened 2026-05-25
test(e2e): retry cluster creation on port collision (#11278)

Mostafahassen1:
#### What type of PR is this?
/kind flake
/kind bug

Fixes #11278

### The Error

* Multiple concurrent CI jobs randomly pull the same ephemeral port from the host OS network.
* KinD cluster creation crashes with `port is already allocated`, causing a 100% infrastructure test failure.

### The Solution

* Wrap the KinD cluster creation step inside a 3-attempt retry loop with a 3-second sleep.
* After a failure, KinD deletes the broken setup and requests a brand new random port.
* Use a `grep` guard to only retry specifically when the "port is already allocated" string appears.
* Avoid retrying real bugs like configuration mistakes, missing dependencies, or bad test code.
* Keep ports completely dynamic so multiple separate Pull Request test jobs can still run in parallel.
* Fix this directly inside `hack/testing/e2e-common.sh` to permanently stabilize all E2E test suites at once.

```release-note
NONE
```

PR #12338, opened 2026-06-18
test: retry kind cluster creation on transient failures

apullo777:
#### What type of PR is this?

/kind flake
/kind bug
/area testing

`cluster_create` already retries `kind create cluster`, but only for the `port is already allocated` case added in #11586.

In #12307, kind failed during kubeadm's `wait-control-plane` phase, so the existing allowlist did not match and `retry.sh` aborted after attempt 1/3.

This keeps the allowlist narrow and adds one known transient cluster start-up signature:

`error execution phase wait-control-plane`

The existing cleanup deletes the partial kind cluster before retrying.

Addresses the observed failure mode in #12307.

An alternative approach (flip to retry-by-default + a `non_retriable_errors` list, mirroring `e2e_docker_pull_if_needed`) is being discussed in #12307. Happy to switch this PR to that shape if reviewer prefers it.

```release-note
NONE
```

## Summary by CodeRabbit

* **Chores**
  * Enhanced test infrastructure to handle a broader range of transient errors during cluster setup, improving overall test reliability and reducing false failures in the CI/CD pipeline.

### DIFF
--- a/hack/testing/e2e-common.sh
+++ b/hack/testing/e2e-common.sh
@@ -550,7 +550,9 @@ function cluster_create {
 
     local log_file="$ARTIFACTS/$cluster-create.log"
     local create_cmd="$KIND create cluster --name \"$cluster\" --image \"$E2E_KIND_VERSION\" --config \"$kind_config\" --kubeconfig=\"$kubeconfig\" --wait 5m -v 5 > \"$log_file\" 2>&1"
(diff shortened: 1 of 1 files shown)


mimowo:
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for stabilizing the build, we are almost there 👍

PR #12344, opened 2026-06-18
[TAS] add feature TASReplaceMultipleFailedNodes feature to allow incremental node replacements without eviction

tg123:
#### What type of PR is this?
 
 /kind feature
 /area tas
 
 Adds a new alpha feature gate `TASReplaceMultipleFailedNodes` (default off)
 that lets Topology-Aware Scheduling replace **multiple** failed nodes for an
 admitted workload, one at a time, instead of evicting the workload.
 
 Today TAS can replace a single failed node in a workload's
 `TopologyAssignment`, but as soon as a **second distinct node** fails while a
 replacement is still in flight, the workload is evicted (by the node
 controller's multi-node-failure path, and/or by the scheduler's fail-fast
 eviction `TASFailedNodeReplacementFailFast`). For large, long-running
 gang-scheduled TAS workloads on unreliable hardware (e.g. GPU fleets), an
 eviction means tearing down and re-admitting the whole workload, which is far
 more disruptive than incrementally swapping out the failed nodes.
 
 When `TASReplaceMultipleFailedNodes` is enabled:
 
 - **Node controller** (`evictWorkloadIfNeeded`): the second-distinct-node
   eviction is suppressed; the newly failed node is appended to
   `Status.UnhealthyNodes` and the workload stays admitted.
 - **Scheduler** (`processEntry`): the fail-fast eviction is suppressed, so a
   single failed replacement attempt does not evict the workload.
 - **Replacement planner** (`findReplacementAssignment`): only the
   head-of-queue unhealthy node is replaced per cycle. Other queued unhealthy
   nodes — which may also be missing from the snapshot — are ignored during the
   stale-assignment check (new `isTopologyAssignmentStaleIgnoring` helper), so a
   stuck tail entry doesn't block head replacement. Tail entries are
   re-discovered and replaced on subsequent reconciles after each successful
   head replacement.
 
 The change also hardens two concurrency edge cases that this multi-failure
 path exposes:
 - `findReplacementAssignment` rejects a replacement candidate domain that
   overlaps an existing assignment domain, avoiding a `mergeTopologyAssignments`
   collapse (`canMergeDomains`, `Count += Count`) that would silently drop a pod
   slot and corrupt pod counts. The cycle retries on a fresh snapshot.
 - `mergeTopologyAssignments`'s sort is made nil-safe for tail domains that have
   already been removed from the snapshot, falling back to ordering by raw
   values instead of nil-panicking.
 
 The gate defaults off; existing single-node-replacement and eviction behavior
 is unchanged when it is disabled.
 
 Fixes #
 
 - No API change. New behavior is entirely gated behind
   `TASReplaceMultipleFailedNodes` (Alpha, default off, introduced in 0.19).
 - The gate is wired so that it overrides `TASFailedNodeReplacementFailFast`
   (Beta, default on): when both are set, no-eviction wins.
 - Tests: Added integration coverage under
   `test/integration/singlecluster/tas/` for incremental multi-node
   replacement, eviction thresholds, topology pod-count consistency, and the
   greedy-assignment fallback path.
 
 ```release-note
 TAS: Added a new alpha feature gate `TASReplaceMultipleFailedNodes` (default off). When enabled, a Topology-Aware Scheduling workload with multiple failed nodes has those nodes replaced incrementally instead of being evicted, keeping the workload admitted.
```

## Summary by CodeRabbit

* **New Features**
  * Added Alpha support for incrementally replacing multiple failed topology-aware scheduling nodes.
  * Added a per-workload annotation to configure the unhealthy-node eviction threshold; `0` allows unlimited failures.
* **Bug Fixes**
  * Prevented unrelated or concurrent failures from blocking replacement.
  * Improved handling of incomplete topology information and repeated replacements.
* **Documentation**
  * Documented multi-node replacement behavior, feature-gate configuration, and eviction thresholds.
* **Tests**
  * Added coverage for replacement workflows, eviction thresholds, and multiple failure scenarios.

### CODE REVIEW

**Comment by mimowo** [comment]
@tg123 thanks for the PR!

Historically we chose only replacing one node because the algorithms for replacing multiple nodes got complicated, but certainly it was something we considered as a possiblity, and thus introduced the API which contains a list of Nodes rather than one.

My main questions:
1. do you foresee a path forward where the generalization is the only supported mode?
2. do you see drawbacks of replacing multiple nodes? In which case we may choose to have a configuration setting for long term?

It is ok if you choose "maybe, I don't know yet", but would love to know your considerations first.

It is great that you have a working prototype, but I would like to make sure we have a KEP update where we align on our current knowledge of the problem space, and the path forward, as much as we can.

**Comment by tg123** [comment]
more context, we verified this change in 0.16.6 for a month running a very big job.

1. Could multi-node replacement become the only supported mode?

I think probably yes at the algorithm level — single-node replacement is just the N == 1 case of the general one, so I'd rather not maintain two code paths long-term.

2. Drawbacks of replacing multiple nodes / case for a long-term config
the main goal is to prevent large job from being requeue, the replacing may wait forever if there is no candidate. 
so current algo is very simple:
```
 while (bad node > 0) { 
    lock  { 
      replace 1 node
      bad node--
    } 
}

for now, we also introduced some tool to manually move stuck slice to another rack if there is room, but need to restart `slice-size` count of pod. maybe, this is another feature we can add in feature

 ```
 
 
 i am also drafting e2e test to fit open source plugin, will add them soon
let me know if anything need to improve and let me know if you want me to draft KEP update

mimowo:
> more context, we verified this change in 0.16.6 for a month running a very big job.

Excellent results then 👍 Thank you for sharing, certainly it helps to build confidence in the change.

In your experiments, do you know how many Pods would require migration at any given moment (how many Nodes would fail?). I'm wondering if we still could cap the algo, but move the cap to say 5 or 10?

> 1. Could multi-node replacement become the only supported mode?
> 
> I think probably yes at the algorithm level — single-node replacement is just the N == 1 case of the general one, so I'd rather not maintain two code paths long-term.

sgtm, I would also want to have one path long term.

> 2. Drawbacks of replacing multiple nodes / case for a long-term config
>    the main goal is to prevent large job from being requeue, the replacing may wait forever if there is no candidate.
>    so current algo is very simple:
> 
> ```
>  while (bad node > 0) { 
>     lock  { 
>       replace 1 node
>       bad node--
>     } 
> }
> 
> for now, we also introduced some tool to manually move stuck slice to another rack if there is room, but need to restart `slice-size` count of pod. maybe, this is another feature we can add in feature
> ```

Isn't this a bit not oversimplifying? Assume you have "topology-required": "rack". Then consider the following:
- each rack can host 2 Pods, and one rack failed. Now, if you do the algo you will fail to transfer the two Pods to the new rack, because moving the first one will fail (IIUC) due to the constraint (the other failed is not moved yet). For example, when you start searching for the rack replacement you need to check "ahead" that the rack will have enough free space to acommodate all the pods from the previous rack.

The proposed algo will work well for "unconstrained" and "preferred". For "required" it may work in many cases, but I think to make it work in all cases requires some complications. Still, it sounds like a pragmatic balance between finding optimal solution and complexity. So, I'm ok with that, but we need to document all the tricky cases in unit tests, and make sure nothing bad happens, like panic, or workload getting stuck when "required" is used.

> i am also drafting e2e test to fit open source plugin, will add them soon let me know if anything need to improve and let me know if you want me to draft KEP update

Great.

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -774,7 +805,23 @@ func (s *TASFlavorSnapshot) requiredReplacementDomain(tr *TASPodSetRequests, ta
...
 func (s *TASFlavorSnapshot) IsTopologyAssignmentStale(ta *utiltas.TopologyAssignment) (bool, string) {
+	return s.isTopologyAssignmentStaleIgnoring(ta, nil)
+}
+
+// isTopologyAssignmentStaleIgnoring returns whether the topologyAssignment contains
+// node-level domains missing from the snapshot, ignoring any node names in
+// ignoreNodes. Used by the head-of-queue replacement path so that other queued
+// unhealthy nodes (which may also be missing from the snapshot) do not poison
+// the stale-check for the head we are actively replacing.
+func (s *TASFlavorSnapshot) isTopologyAssignmentStaleIgnoring(ta *utiltas.TopologyAssignment, ignoreNodes map[string]struct{}) (bool, string) {
```
Singularity23x0: Nit: We could use `sets.Set[string]` here instead of `map[string]struct{}`

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -686,7 +700,24 @@ func (s *TASFlavorSnapshot) findReplacementAssignment(
...
+	// retries with a fresh snapshot.
+	existingDomainIDs := make(map[utiltas.TopologyDomainID]struct{}, len(existingAssignment.Domains))
+	for _, d := range existingAssignment.Domains {
+		existingDomainIDs[utiltas.DomainID(d.Values)] = struct{}{}
+	}
+	for _, d := range replacementAssignment[tr.PodSet.Name].Domains {
+		if _, dup := existingDomainIDs[utiltas.DomainID(d.Values)]; dup {
+			return nil, nil, fmt.Sprintf("replacement candidate %v already present in existing TopologyAssignment; concurrent replacement race - retry on next cycle", d.Values)
+		}
 	}
```
Singularity23x0: Nit
```suggestion
	existingDomainIDs := sets.New[utiltas.TopologyDomainID]()
	for _, d := range existingAssignment.Domains {
		existingDomainIDs.Insert(utiltas.DomainID(d.Values))
	}
	for _, d := range replacementAssignment[tr.PodSet.Name].Domains {
		if existingDomainIDs.Has(utiltas.DomainID(d.Values)) {
			return nil, nil, fmt.Sprintf("replacement candidate %v already present in existing TopologyAssignment; concurrent replacement race - retry on next cycle", d.Values)
		}
}
```

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -654,8 +654,22 @@ func (s *TASFlavorSnapshot) findReplacementAssignment(
...
+	// stale-assignment check below, otherwise head replacement would never
+	// succeed and the workload would be stuck without an automatic eviction
+	// safety net.
+	var ignoreNodes map[string]struct{}
+	if features.Enabled(features.TASReplaceMultipleFailedNodes) && len(wl.Status.UnhealthyNodes) > 1 {
+		ignoreNodes = make(map[string]struct{}, len(wl.Status.UnhealthyNodes)-1)
+		for _, n := range wl.Status.UnhealthyNodes[1:] {
+			ignoreNodes[n.Name] = struct{}{}
+		}
+	}
```
Singularity23x0: Nit: If we switch to sets
```suggestion
	var ignoreNodes sets.Set[string]
	if features.Enabled(features.TASReplaceMultipleFailedNodes) && len(wl.Status.UnhealthyNodes) > 1 {
		ignoreNodes = sets.New(wl.Status.UnhealthyNodes[1:]...)
	}
```

mimowo:
cc @pajakd @j-skiba ptal

tg123:
_Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our release note process to remove it._

worst case 5 nodes in ~20m, the job size is ~2000 nodes


_The proposed algo will work well for "unconstrained" and "preferred". For "required" it may work in many cases, but I think to make it work in all cases requires some complications. Still, it sounds like a pragmatic balance between finding optimal solution and complexity. So, I'm ok with that, but we need to document all the tricky cases in unit tests, and make sure nothing bad happens, like panic, or workload getting stuck when "required" is used._

we use `required`, and allow pod stuck waiting for nodes. human may kick in to help if auto repair does not work

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -686,7 +701,29 @@ func (s *TASFlavorSnapshot) findReplacementAssignment(
...
+		existingDomainIDs := make(map[utiltas.TopologyDomainID]struct{}, len(existingAssignment.Domains))
+		for _, d := range existingAssignment.Domains {
+			existingDomainIDs[utiltas.DomainID(d.Values)] = struct{}{}
+		}
+		for _, d := range replacementAssignment[tr.PodSet.Name].Domains {
+			if _, dup := existingDomainIDs[utiltas.DomainID(d.Values)]; dup {
+				return nil, nil, fmt.Sprintf("replacement candidate %v already present in existing TopologyAssignment; concurrent replacement race - retry on next cycle", d.Values)
+			}
+		}
 	}
```
j-skiba: it might detect non-race scenarios too

topology
`block-1`:
- `node-1` - 4 GPUs
- `node-2` - 6 GPUs
- `node-3` - 2 GPUs

let's say there is workload with:
- 4 pods on `block-1/node1`
- 4 pods on `block-1/node2`

1 pod - 1 GPU

`node1` fails and we start looking for replacement. The only possibility is to put 2 pods on `node-2` and 2 pods on `node-3` but this check would not allow that. The workload would be retried in a loop all the time

mimowo:
> we use required, and allow pod stuck waiting for nodes. human may kick in to help if auto repair does not work

 Currently when two nodes fail, they require eviction, but this happens unassisted. Waiting for an admin reaction to unstuck the job will certainly be considered as a regression by some users. 
 
 So, we need to have a plan, either (1) improve the algo to support multiple nodes without issues, or add some kind of API, or maybe (2.) some kind of annotation to allow "require" with multiple nodes failing getting stuck for manual intervation, but that feels like needs to be opt-in.

tg123:
do you want me to share/doc our dirty tools?
it moves slice from rack to rack via editing kwl
which means 1 slice (2-16 nodes in our case) has bad node and no replacement in same rack, whole slice need to be moved to another rack.

i do not think this is a good algo, however, it is really helpful for repairing by only to restart pods on a small partial of nodes

mimowo:
> do you want me to share/doc our dirty tools?

not really, but I think this is the difference between hacking some solution just for a specific company, rather than making a default for all users. 

> i do not think this is a good algo, however, it is really helpful for repairing by only to restart pods on a small partial of nodes

I can clearly imagine this behavior being useful for workloads with >1k pods.

The question is - could we improve it to work better, or as mentioned above make the behavior opt-in. For example we could by default evict if there are >1 nodes failing, or leave it to an admin if there is `kueue.x-k8s.io/tas-unhealthy-nodes-eviction-threshold` (or something like that). That could be assumed as 1 by default if unspecified, but could be increased on some workloads, wdyt?

tg123:
i agree, `kueue.x-k8s.io/tas-unhealthy-nodes-eviction-threshold` , let me impl it instead of using a feature

pkg/cache/scheduler/tas_flavor_snapshot.go
tg123: testcase added, please check

tg123:
`kueue.x-k8s.io/tas-unhealthy-nodes-eviction-threshold` added
default = 1

N -> evict when len(unhealthy node) > N  
0 -> no ec
others (invalid neg) -> default 


N = 1 is current behavior, i am thinking removing featureflag? and make default a configable value?

mimowo:
Whatever we choose I think we need an alpha FG in order to make it clear to users that this may change.

And I'm still not committed that we go with `kueue.x-k8s.io/tas-unhealthy-nodes-eviction-threshold`, it just felt reasonable.

Another option I'm considering about is to allow for an arbitrary number of nodes to be replaced, but when this is "required" and the simple heursistic is getting stuck, then evict the workload. This seems suboptimal in some cases where we choose the wrong domain by the greedy heuristic, but it will jus result in eviction, as it would result eviction currently anyway. So, I think it might be a reasonable approach, wdyt?

tg123:
Agreed on the alpha FG — it's already behind  TASReplaceMultipleFailedNodes  (alpha, off by default).

 On the design: I don't have a broad enough view of Kueue to have a strong opinion, so either direction works for us. Our concrete need is just to keep large training jobs running across node failures rather than evicting the whole job. A big part  of why we opened this PR is to confirm we understand Kueue's behavior correctly and to contribute the fix back rather than carry it downstream.

Your "replace arbitrarily, evict only when the greedy heuristic gets stuck" approach sounds reasonable to me and matches our goal — it falls back to eviction exactly where today's behavior already would. Happy to drop the threshold annotation and  go that way if you prefer; just let me know which direction you'd like and I'll adjust the PR.

pkg/cache/scheduler/tas_flavor_snapshot.go
j-skiba: it looks good, thanks

pkg/scheduler/scheduler.go
```diff
@@ -804,8 +804,21 @@ func (s *Scheduler) admit(ctx context.Context, e *entry, cq *schdcache.ClusterQu
...
+					// longer generate node events (e.g. deleted nodes), leaving
+					// them stuck in the TopologyAssignment forever.
+					log.V(5).Info("Dropping the replaced head node from the workload recovery field, keeping the remaining unhealthy nodes",
+						"replacedNode", e.Obj.Status.UnhealthyNodes[0].Name,
+						"remainingUnhealthyNodes", len(e.Obj.Status.UnhealthyNodes)-1)
+					wl.Status.UnhealthyNodes = e.Obj.Status.UnhealthyNodes[1:]
+				} else {
+					log.V(5).Info("Clearing the topology assignment recovery field from the workload status after successful recovery")
+					wl.Status.UnhealthyNodes = nil
+				}
```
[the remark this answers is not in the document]
j-skiba: I think it would be safer to remove unhealthy nodes by name

test/integration/singlecluster/tas/tas_test.go
```diff
@@ -2070,6 +2070,258 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
...
+								{Count: 1, Values: []string{node2Name}},
+							},
+						}),
+					))
+				})
+
+				ginkgo.By("deleting the first assigned node: "+node1Name, func() {
+					nodeToDelete := &corev1.Node{ObjectMeta: metav1.ObjectMeta{Name: node1Name}}
+					gomega.Expect(k8sClient.Delete(ctx, nodeToDelete)).Should(gomega.Succeed())
+					util.ExpectObjectToBeDeleted(ctx, k8sClient, nodeToDelete, false)
```
j-skiba: nit
```suggestion
					util.ExpectObjectToBeDeleted(ctx, k8sClient, nodeToDelete, true)
```

also in the tests below

test/integration/singlecluster/tas/tas_test.go
```diff
@@ -2070,6 +2070,258 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
...
+				// With the gate enabled and an eviction threshold of 2, the
+				// workload tolerates up to two unhealthy nodes; once a third
+				// distinct node fails the unhealthy-node count exceeds the
+				// threshold and the workload is evicted. The workload occupies
+				// all four nodes so that the failed nodes cannot be replaced (no
+				// spare capacity) and the unhealthy-node count accumulates
+				// deterministically past the threshold.
+				features.SetFeatureGateDuringTest(ginkgo.GinkgoTB(), features.TASReplaceMultipleFailedNodes, true)
+
+				var wl1 *kueue.Workload
```
j-skiba, line 3411: nit
```suggestion
				var wl *kueue.Workload
```

also in the tests below

pkg/scheduler/scheduler.go
```diff
@@ -414,7 +414,10 @@ func (s *Scheduler) processEntry(
 	usage, fits := s.updateAssignmentIfNeeded(ctx, log, e, snapshot, cq, preemptedWorkloads)
 	mode := e.assignment.RepresentativeMode()
 
-	if features.Enabled(features.TASFailedNodeReplacementFailFast) && workload.HasTopologyAssignmentWithUnhealthyNode(e.Obj) && mode != flavorassigner.Fit {
+	if features.Enabled(features.TASFailedNodeReplacementFailFast) && workload.HasTopologyAssignmentWithUnhealthyNode(e.Obj) &&
+		mode != flavorassigner.Fit &&
+		(!features.Enabled(features.TASReplaceMultipleFailedNodes) ||
+			len(e.Obj.Status.UnhealthyNodes) > workload.UnhealthyNodesEvictionThreshold(e.Obj)) {
```
j-skiba: I'm not sure what's the expected behavior now, should a workload be evicted while `len(unhealthyNodes) == threshold` and we cannot find replacement node or should we keep looking for replacement in future scheduling cycles?

shouldn't this check be
```golang
len(e.Obj.Status.UnhealthyNodes) >= workload.UnhealthyNodesEvictionThreshold(e.Obj)
```
?

It seems like node controller doesn't append to `UnhealthyNodes` if we meet the threshold and evicts, so probably we won't see `len(e.Obj.Status.UnhealthyNodes) > threshold` to be true.

j-skiba: https://github.com/kubernetes-sigs/kueue/pull/12344/changes#diff-c8054619a353d0fc872425de6a22f0df62001353eb75f856d5ba504553b5fcbfR1511

```
- The scheduler fail-fast eviction is suppressed while the number of unhealthy nodes is below
  `N` (overriding `TASFailedNodeReplacementFailFast`); at or above `N` it is allowed to evict.
```

according to this it should be `>=`

pkg/features/kube_features.go
```diff
@@ -144,6 +144,19 @@ const (
...
 
+	// TASReplaceMultipleFailedNodes enables replacing multiple failed nodes for
+	// a TAS workload instead of evicting the workload. Without this gate, TAS
+	// replaces a single failed node, but evicts the workload as soon as a
+	// second distinct node fails while a replacement is still in flight. With
+	// this gate enabled, the workload stays admitted: failed nodes are queued
+	// in Status.UnhealthyNodes and replaced incrementally, head-of-queue first,
+	// as fits become available. It suppresses both the node-controller eviction
+	// on the second distinct failure and the scheduler fail-fast eviction
+	// (TASFailedNodeReplacementFailFast), and makes the replacement planner
```
j-skiba: nit:
```suggestion
	// this gate enabled and an eviction threshold greater than 1 configured via the
	// kueue.x-k8s.io/tas-unhealthy-nodes-eviction-threshold annotation (or 0 for unlimited),
	// the workload stays admitted: failed nodes are queued in Status.UnhealthyNodes
	// and replaced incrementally, head-of-queue first, as fits become available. While
	// within threshold, it suppresses both the node-controller eviction on the second
	// distinct failure and the scheduler fail-fast eviction (TASFailedNodeReplacementFailFast),
```

pkg/controller/tas/node_controller.go
```diff
@@ -486,8 +486,20 @@ func (r *nodeReconciler) checkPodsOnNode(
 
 // evictWorkloadIfNeeded idempotently evicts the workload when the node has failed.
 // It returns whether the node was evicted, and whether an error was encountered.
+//
+// When the TASReplaceMultipleFailedNodes feature gate is enabled, the second-distinct-node
+// eviction is suppressed. The new node is appended via addUnhealthyNode and the
+// workload remains admitted; the scheduler will keep attempting head replacement.
```
j-skiba: nit:
```suggestion
// When the TASReplaceMultipleFailedNodes feature gate is enabled and the number of unhealthy
// nodes is below the threshold configured via the kueue.x-k8s.io/tas-unhealthy-nodes-eviction-threshold
// annotation (default 1), eviction is suppressed. The new node is appended via addUnhealthyNode
// and the workload remains admitted; the scheduler will keep attempting head replacement.
```

j-skiba (commented):
Lgtm in general.

Just 2 nits for documentation and I also think one more integration test for scenario when:
- threshold is > 1
- workload has len(unhealthyNodes) < threshold
- there's currently no enough space to replace the node
- workload is not evicted

would give us good coverage overall

pkg/scheduler/scheduler.go
```diff
@@ -415,7 +415,10 @@ func (s *Scheduler) processEntry(
 	usage, fits := s.updateAssignmentIfNeeded(ctx, log, e, snapshot, cq, preemptedWorkloads)
 	mode := e.assignment.RepresentativeMode()
 
-	if features.Enabled(features.TASFailedNodeReplacementFailFast) && workload.HasTopologyAssignmentWithUnhealthyNode(e.Obj) && mode != flavorassigner.Fit {
+	if features.Enabled(features.TASFailedNodeReplacementFailFast) && workload.HasTopologyAssignmentWithUnhealthyNode(e.Obj) &&
+		mode != flavorassigner.Fit &&
+		(!features.Enabled(features.TASReplaceMultipleFailedNodes) ||
+			len(e.Obj.Status.UnhealthyNodes) >= workload.UnhealthyNodesEvictionThreshold(e.Obj)) {
```
j-skiba: nit:

this and https://github.com/kubernetes-sigs/kueue/pull/12344/changes#diff-4a0cb79ce927ca48906417d823ad3303285b2c08b5a74c0a66d66214578eb654R496-R503 could be extracted to something like `IsWithinUnhealthyNodesEvictionThreshold`

pkg/workload/workload.go
```diff
@@ -1357,6 +1359,50 @@ func HasTopologyAssignmentWithUnhealthyNode(w *kueue.Workload) bool {
...
+				return n
+			}
+		}
+	}
+	return DefaultUnhealthyNodesEvictionThreshold
+}
+
+// IsWithinUnhealthyNodesEvictionThreshold indicates whether the Workload can
+// continue replacing unhealthy nodes without scheduler fail-fast eviction.
+func IsWithinUnhealthyNodesEvictionThreshold(w *kueue.Workload) bool {
```
j-skiba: Actually I'm not sure now if we can reuse that code. Sorry for confusion!

So in node_controller we check in `evictWorkloadIfNeeded` if the current `unhealthyNodes` length is less than threshold and we do that because later we add a node to the list so in edge case after adding the node, `len(unhealthyNodes) == threshold` will be true. If `len(unhealthyNodes) == threshold` was true before adding a node then we should evict because after addition it would be > threshold. That's why we need `len(w.Status.UnhealthyNodes) < UnhealthyNodesEvictionThreshold(w)` comparison in node_controller.

But in scheduler we don't append to `unhealthyNodes` list. We only want to know if at the current moment `len(unhealthyNodes) <= threshold`.

pkg/scheduler/scheduler.go
```diff
@@ -914,14 +917,34 @@ func (s *Scheduler) admit(ctx context.Context, e *entry, cq *schdcache.ClusterQu
 
 	newWorkload := e.Obj.DeepCopy()
 	s.admissionRoutineWrapper.Run(func() {
+		replacedNodeName := ""
+		if len(e.Obj.Status.UnhealthyNodes) > 0 {
+			replacedNodeName = e.Obj.Status.UnhealthyNodes[0].Name
```
j-skiba:
```suggestion
			replacedNodeName = newWorkload.Status.UnhealthyNodes[0].Name
```

pkg/cache/queue/manager.go
```diff
@@ -996,10 +996,24 @@ func (m *Manager) QueueSecondPassIfNeeded(ctx context.Context, w *kueue.Workload
 }
 
 func (m *Manager) queueSecondPass(ctx context.Context, w *kueue.Workload, iteration int) {
+	log := ctrl.LoggerFrom(ctx)
+	if features.Enabled(features.TASReplaceMultipleFailedNodes) {
```
j-skiba: let's add `workload.HasTopologyAssignmentWithUnhealthyNode(w)` here so we don't do Get calls for each workload

pkg/workload/workload.go
tg123: reverted

keps/2724-topology-aware-scheduling/kep.yaml
mimowo: nit adjust to 0.20

pkg/cache/queue/manager.go
```diff
@@ -996,10 +996,24 @@ func (m *Manager) QueueSecondPassIfNeeded(ctx context.Context, w *kueue.Workload
...
 func (m *Manager) queueSecondPass(ctx context.Context, w *kueue.Workload, iteration int) {
+	log := ctrl.LoggerFrom(ctx)
+	if features.Enabled(features.TASReplaceMultipleFailedNodes) && workload.HasTopologyAssignmentWithUnhealthyNode(w) {
+		var latest kueue.Workload
+		if err := m.client.Get(ctx, client.ObjectKeyFromObject(w), &latest); err == nil {
+			w = &latest
+		} else if apierrors.IsNotFound(err) {
+			m.Lock()
+			m.secondPassQueue.deleteByKey(workload.Key(w))
+			m.Unlock()
```
mimowo: Could you encapsulate the locking to a helper

test/compatibility_lifecycle/reference/versioned_feature_list.yaml
```diff
@@ -437,6 +445,12 @@
     lockToDefault: false
     preRelease: Beta
     version: "0.19"
+- name: TASReplaceMultipleFailedNodes
+  versionedSpecs:
+  - default: false
+    lockToDefault: false
+    preRelease: Alpha
+    version: "0.19"
```
mimowo: update

pkg/workload/workload.go
```diff
@@ -1357,6 +1359,44 @@ func HasTopologyAssignmentWithUnhealthyNode(w *kueue.Workload) bool {
...
+//   - "0" means never evict (UnlimitedUnhealthyNodesEvictionThreshold);
+//   - a positive integer N tolerates up to N unhealthy nodes;
+//   - an absent or otherwise invalid value returns
+//     DefaultUnhealthyNodesEvictionThreshold (1).
+func UnhealthyNodesEvictionThreshold(w *kueue.Workload) int {
+	if w == nil {
+		return DefaultUnhealthyNodesEvictionThreshold
+	}
+	if v, ok := w.Annotations[kueue.TASUnhealthyNodesEvictionThresholdAnnotation]; ok {
+		if n, err := strconv.Atoi(v); err == nil {
```
mimowo: Make sure errors are not silenced

pkg/cache/queue/manager.go
```diff
@@ -996,10 +1003,22 @@ func (m *Manager) QueueSecondPassIfNeeded(ctx context.Context, w *kueue.Workload
...
+		if err := m.client.Get(ctx, client.ObjectKeyFromObject(w), &latest); err == nil {
+			w = &latest
+		} else if apierrors.IsNotFound(err) {
+			m.DeleteSecondPass(workload.Key(w))
+			return
+		} else {
+			log.Error(err, "Failed to refresh workload before the second pass", "workload", workload.Key(w))
+		}
+	}
+
```
mimowo, line 1056: Why refreshing the workload is relevant for the multi-node replacement, and not single node?

I sense that there might also be a bug for single node scenarios? If so, I would like to de-couple the fix and cherry-pick. Let me know your research

pkg/scheduler/scheduler.go
```diff
@@ -415,9 +415,21 @@ func (s *Scheduler) processEntry(
...
+			if err != nil {
+				log.Error(err, "Invalid unhealthy nodes eviction threshold")
+			}
+			failFast = len(e.Obj.Status.UnhealthyNodes) > threshold
+		}
+		if failFast {
+			s.handleFailedTASReplacement(ctx, log, e)
+			return
+		}
 	}
```
mimowo: Could you move this code, or at least some part of it to offload the main scheduler function which is critical and so we want to keep it clean and readable

pkg/scheduler/scheduler.go
```diff
@@ -914,14 +926,34 @@ func (s *Scheduler) admit(ctx context.Context, e *entry, cq *schdcache.ClusterQu
...
+					wl.Status.UnhealthyNodes = slices.DeleteFunc(wl.Status.UnhealthyNodes, func(n kueue.UnhealthyNode) bool {
+						return n.Name == replacedNodeName
+					})
+					log.V(5).Info("Dropping the replaced head node from the workload recovery field, keeping the remaining unhealthy nodes",
+						"replacedNode", replacedNodeName,
+						"remainingUnhealthyNodes", len(wl.Status.UnhealthyNodes))
+				} else {
+					log.V(5).Info("Clearing the topology assignment recovery field from the workload status after successful recovery")
+					wl.Status.UnhealthyNodes = nil
+				}
```
mimowo: Please move the node-repalcement related code to a helper to offload reading the critical admit function from the details

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1484,6 +1485,50 @@ Kueue tries to find a replacement for a failed node until success (or until it g
 evicted by e.g. `waitForPodsReady.recoveryTimeout`). One can limit the number of retries
 to only one, by setting the `TASFailedNodeReplacementFailFast` feature gate to `true`.
 
+##### Replacing multiple failed nodes
```
mimowo, line 1488: Make sure the value for the annotation is capped. I think it cannot exceed 8, because this is the current cap on the UnhealthyNodes. However, I would like to explore what it the number you actually would like to use, maybe 4 is enough in practice as max?

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -724,8 +725,32 @@ func (s *TASFlavorSnapshot) findReplacementAssignment(
...
+		// scheduler has already queued this replacement attempt. Treat any
+		// other missing node-level domain as pending replacement too; the
+		// admission patch removes only the recorded head and preserves newer
+		// UnhealthyNodes observed from the API server.
+		for _, domain := range existingAssignment.Domains {
+			if _, found := s.domains[utiltas.DomainID(domain.Values)]; !found {
+				ignoreNodes.Insert(domain.Values[len(domain.Values)-1])
+			}
+		}
+	}
```
mimowo: extract the ignoreNodes computation to a helper

keps/2724-topology-aware-scheduling/README.md
mimowo, line 1488: I would like the KEP to clarify that the algorithm is greedy, node by node, and this means that in case of "required" we may not find a solution which exists with a more sophisticated mechanism. We can leave this subject to change, but it would be good to set the user expectations,

So imagine there are two pods, and you have required=node. Moving nodes sequentially will not fly, but a "smart" algo could move them together to another node conforming to "required".

I would like to consider two integration tests for "required" mode:
1. greedy algo works for two nodes
2. greedy algo fails for two nodes while a smart algo could succeed

pkg/cache/queue/manager.go
tg123: this seems because we internally have the flag on, and patched the code path only.

checked some old day notes: 

a bad workload repeatedly cycled through admission and eviction nine times.
another workload emitted  SecondPassFailed  shortly before the incident.
TA appeared to flip between scheduling cycles.
restarting Kueue immediately cleared the behavior <--- this is good enough to unblock us

tg123: i am working on a separate pr

tg123: https://github.com/kubernetes-sigs/kueue/pull/14078

keps/2724-topology-aware-scheduling/README.md
tg123: our customer prefers unlimited (value 0 now)
i should say i did not even notice there is a cap at `UnhealthyNodes`

will add a cap and remove unlimited, and i think the behavior `unlimited` makes no sense in this case.

then to seek option to make UnhealthyNodes cap to `uint32` or `uint8`, what do you think?

tg123: now [1,8] only

mimowo:
@tg123 do you have cycles to address the build issues?

Overall I like the enhancement. Would you like to propose it at the wg-batch or a planning meeting?

tg123:
Sure will check it tonight

keps/2724-topology-aware-scheduling/README.md
amy: Wait. Also eventually want unlimited threshold. **This can be a followup if it overcomplicates this existing PR.**

As these workload sizes increase, you sorta want to look at proportionality. What is 8 simultaneous failed nodes in the context of like thousands of nodes. 

These days I think user code can also handle some threshold of failure. So I don't want eviction if its just say 8 node failures for a 3K node job as an example. We should allow users to set a higher threshold than that.

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -724,8 +725,10 @@ func (s *TASFlavorSnapshot) findReplacementAssignment(
 	wl *kueue.Workload,
 	assumedUsage map[utiltas.TopologyDomainID]resources.Requests,
 ) (*utiltas.TopologyAssignment, *utiltas.TopologyAssignment, string) {
-	tr.Count = deleteDomain(existingAssignment, wl.Status.UnhealthyNodes[0].Name)
-	if isStale, staleDomain := s.IsTopologyAssignmentStale(existingAssignment); isStale {
+	headNodeName := wl.Status.UnhealthyNodes[0].Name
```
amy, line 694: Replacement only ever attempts `UnhealthyNodes[0]`, which can starve replaceable nodes behind an unreplaceable head.

Example: a workload with two PodSets (e.g. a JobSet with two replicated jobs), each with a `required` rack topology. Replacement is pinned to each PodSet's assigned rack because the healthy pods do not move:

```
        RACK-1 (podset-a)             RACK-2 (podset-b)
  node-1: pod (dies 1st)        node-4: pod (dies 2nd)
  node-2: pod                   node-5: pod
  node-3: other job             node-6: empty

unhealthyNodes: [node-1, node-4]

every cycle: replace node-1? RACK-1 has no free node, fail.
             node-4 is never attempted, though node-6 is a valid fit.
```

The workload runs with node-4's pod missing, even though a valid replacement (node-6) is free the whole time.

Notably, this PR already handles the mirror image of this problem via `isTopologyAssignmentStaleIgnoring`, which keeps a dead tail entry from poisoning the stale check during head replacement.

If strict FIFO is intentional, we need to document the starvation case in the KEP. 

> ai-assisted comment 🤖 beep boop

mimowo (commented):
LGTM overall. Leaving the reminder of the review to @amy and @tenzen-y

keps/2724-topology-aware-scheduling/README.md
tg123: agree, will follow up this size in following PR

pkg/cache/scheduler/tas_flavor_snapshot.go
tg123: You are right, this can leak dead lock

This is intentional for Alpha: replacements are strict FIFO, one node per cycle. It can cause head-of-line blocking, so I’ll document this limitation in the KEP

pkg/cache/queue/manager.go
tg123: i removed this change in the pr, will discuss in 14078

tg123: sorry @mimowo , i found new e2e test failed after remove it, i reverted the `revert commit`

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1485,6 +1486,64 @@ Kueue tries to find a replacement for a failed node until success (or until it g
...
```
mimowo, line 1489: I would like to try to make this more approachable, it seems very verbose currently.

In particular:
1. the intro para "The mechanism described above replaces a single failed node. .." is repeating content of other paragraphs 
2. "For large, long-running gang-scheduled TAS workload" seems to mix motivation with validation
3. "Strict head-of-queue processing can cause head" seems to document the generic mechanism which is not specific to the feature
4. "The default threshold of `1` lets the Workload tolerate one ..." this paragraph is obvious  - we keep 1 as default

I imagine something like:

##### Replacing multiple failed nodes

Without this feature, a second failed node triggers Workload eviction.
Enabling the  `TASReplaceMultipleFailedNodes` feature gate and setting the 
 `kueue.x-k8s.io/tas-unhealthy-nodes-eviction-threshold` allows Kueue to keep the Workload
admitted while incrementally replacing up to `N` unhealthy nodes.

In Alpha, Kueue processes `.status.unhealthyNodes` by attempting one failed-node replacement per Workload per scheduling cycle. Placement is greedy and may miss feasible solutions for `required` topology requests. For example, placing the first replacement Pod in a rack with insufficient room for the second may block recovery, even though another rack could fit both.

We will revisit joint replacement planning for `required` topology requests for Beta or GA, based on user feedback.

apis/kueue/v1beta2/topology_types.go
```diff
@@ -74,6 +74,19 @@ const (
...
+	// instead of having its failed nodes replaced in place. The value is an
+	// integer N in the range [1, 8]: while at most N of the Workload's assigned
+	// nodes are unhealthy, Kueue keeps the Workload admitted and replaces the
+	// failed node(s) incrementally; once a further distinct node failure would
+	// exceed N (or no replacement can be found for the current unhealthy set),
+	// the Workload is evicted. An absent or otherwise invalid value defaults to
+	// 1, which reproduces the default single-node-replacement behavior (evict on
+	// the second distinct node failure). Requires the
+	// TASReplaceMultipleFailedNodes feature gate to be enabled; ignored otherwise.
+	TASUnhealthyNodesEvictionThresholdAnnotation = "kueue.x-k8s.io/tas-unhealthy-nodes-eviction-threshold"
```
mimowo, line 88: nit, This looks ok, but verbose, let's try to capture the essence:
```
// TASMaxUnhealthyNodesAnnotation sets the maximum number of unhealthy nodes
// retained for replacement without triggering workload eviction.
// Valid values are 1–8.
// Ignored unless TASReplaceMultipleFailedNodes is enabled.
```

pkg/features/kube_features.go
```diff
@@ -150,6 +150,20 @@ const (
 	// In TAS, treat node as failed if the node is not ready and the pods assigned to this node terminate.
 	TASReplaceNodeOnPodTermination featuregate.Feature = "TASReplaceNodeOnPodTermination"
 
+	// TASReplaceMultipleFailedNodes enables replacing multiple failed nodes for
```
mimowo, line 153: Add yourself as the owner for the FG.

pkg/features/kube_features.go
```diff
@@ -150,6 +150,20 @@ const (
...
+	// replaces a single failed node, but evicts the workload as soon as a
+	// second distinct node fails while a replacement is still in flight. With
+	// this gate enabled and an eviction threshold in the range [1, 8] configured
+	// via the kueue.x-k8s.io/tas-unhealthy-nodes-eviction-threshold annotation,
+	// the workload stays admitted: failed nodes are queued in Status.UnhealthyNodes
+	// and replaced incrementally, head-of-queue first, as fits become available. While
+	// within threshold, it suppresses both the node-controller eviction on the second
+	// distinct failure and the scheduler fail-fast eviction (TASFailedNodeReplacementFailFast),
+	// ignore other queued unhealthy-node domains during the stale-assignment
+	// check so a stuck tail entry does not block head replacement.
```
mimowo, line 164: // TASReplaceMultipleFailedNodes enables support for the
// kueue.x-k8s.io/tas-unhealthy-nodes-eviction-threshold annotation
// which allows to define the maximal number of unhealhy nodes which are
// retained for replacement without triggering workload eviction.

pkg/scheduler/scheduler.go
```diff
@@ -1082,6 +1091,47 @@ func (s *Scheduler) admit(ctx context.Context, e *entry, cq *schdcache.ClusterQu
...
+	admission *kueue.Admission,
+) error {
+	replacedNodeName := ""
+	if len(wl.Status.UnhealthyNodes) > 0 {
+		replacedNodeName = wl.Status.UnhealthyNodes[0].Name
+	}
+	patchOptions := []workloadpatching.PatchStatusOption{workloadpatching.WithRetryOnConflict()}
+	if !features.Enabled(features.TASReplaceMultipleFailedNodes) {
+		patchOptions = append(patchOptions, workloadpatching.WithLooseOnApply())
+	}
```
mimowo, line 1108: Why is the choice of patch options depending on the feature? This seems pretty unrelated, is it trying to fix another issue maybe?

apis/kueue/v1beta2/topology_types.go
tenzen-y: +1

tenzen-y, line 88:
```suggestion
	UnhealthyNodesConcurrentEvictionThresholdAnnotation = "kueue.x-k8s.io/unhealthy-nodes-concurrent-eviction-threshold"
```

We often use "WHAT" as a prefix, like "podset" and "pod", but "tas" is a concept name that is not a target.
Also, IIUC, this annotation represents how many failed nodes are handled concurrently. 

So, I would recommend using `kueue.x-k8s.io/unhealthy-nodes-concurrent-eviction-threshold`. Also,

keps/2724-topology-aware-scheduling/README.md
tenzen-y, line 1489: Please also mention about Kueue version and feature stage the same as another TAS enhancement section.

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1485,6 +1486,64 @@ Kueue tries to find a replacement for a failed node until success (or until it g
...
+
+The mechanism described above replaces a single failed node. As soon as a *second* distinct
+node fails while a replacement is still in flight, the workload is evicted — both by the
+node-failure controller (the `evictWorkloadIfNeeded` "multiple node failures" path) and, when
+`TASFailedNodeReplacementFailFast` is enabled, by the scheduler's fail-fast eviction. The
+original design chose single-node replacement deliberately, since replacing multiple nodes at
```
tenzen-y, line 1495:
```suggestion
The Node failures original design chose single-node replacement deliberately, since replacing multiple nodes at
```

Please remove duplicated description as another this KEP section.

keps/2724-topology-aware-scheduling/README.md
```diff
@@ -1485,6 +1486,64 @@ Kueue tries to find a replacement for a failed node until success (or until it g
...
+once complicates the assignment algorithm. The `UnhealthyNodes` status field was nonetheless
+modeled as a *list* (rather than a single node) to leave room for this generalization.
+
+For large, long-running gang-scheduled TAS workloads on unreliable hardware (e.g. GPU fleets),
+eviction is expensive: the whole workload is torn down and re-admitted, even though only a
+couple of nodes failed. We introduce the `TASReplaceMultipleFailedNodes` feature gate (Alpha,
+default off) which, when enabled, allows a workload to keep being admitted while its failed
+nodes are replaced incrementally instead of evicting. How many nodes may be unhealthy at once
+before falling back to eviction is controlled per-Workload by the
+`kueue.x-k8s.io/tas-unhealthy-nodes-eviction-threshold` annotation (an integer `N` in the
```
tenzen-y, line 1505:
```suggestion
`kueue.x-k8s.io/unhealthy-nodes-concurrent-eviction-threshold` annotation (an integer `N` in the
```

pkg/workload/workload.go
tenzen-y, line 1: Can we validate the Workload unhealthyNode field and the new annotation in https://github.com/kubernetes-sigs/kueue/blob/83cc2e2a9f029eb8f933f175a598b9cbfb924493/pkg/webhooks/workload_webhook.go instead of here?

Any reason why we can not leverage that?

pkg/workload/workload.go
```diff
@@ -1473,6 +1474,49 @@ func HasTopologyAssignmentWithUnhealthyNode(w *kueue.Workload) bool {
...
+// DefaultUnhealthyNodesEvictionThreshold is used when the
+// TASUnhealthyNodesEvictionThresholdAnnotation is absent or invalid. It
+// reproduces the default single-node-replacement behavior: the Workload is
+// evicted as soon as a second distinct node fails (i.e. it tolerates one
+// unhealthy node while a replacement is in flight).
+const DefaultUnhealthyNodesEvictionThreshold = 1
+
+// MaxUnhealthyNodesEvictionThreshold is the maximum supported threshold,
+// matching the API limit on Workload.Status.UnhealthyNodes.
+const MaxUnhealthyNodesEvictionThreshold = 8
```
tenzen-y, line 1486: I think this should be moved to apis package where we maintain annotations since this actually for the annotation.

pkg/workload/workload_test.go
```diff
@@ -4185,6 +4185,82 @@ func TestShouldSkipClusterNomination(t *testing.T) {
 	}
 }
 
+func TestUnhealthyNodesEvictionThreshold(t *testing.T) {
+	cases := map[string]struct {
+		annotationValue string
+		annotationSet   bool
+		want            int
+		wantErr         bool
```
tenzen-y, line 4193:
```suggestion
		wantErr         error
```

We should compare error instead of err toggle.

pkg/workload/workload_test.go
```diff
@@ -4185,6 +4185,82 @@ func TestShouldSkipClusterNomination(t *testing.T) {
...
+			}
+		})
+	}
+	got, err := UnhealthyNodesEvictionThreshold(nil)
+	if got != DefaultUnhealthyNodesEvictionThreshold {
+		t.Errorf("UnhealthyNodesEvictionThreshold(nil) = %d, want %d", got, DefaultUnhealthyNodesEvictionThreshold)
+	}
+	if err != nil {
+		t.Errorf("UnhealthyNodesEvictionThreshold(nil) error = %v, want nil", err)
+	}
```
tenzen-y, line 4261: Why do we need to place these outside of t.Run?

test/integration/singlecluster/tas/tas_test.go
```diff
@@ -121,6 +122,20 @@ func assignedPodCount(ta *kueue.TopologyAssignment) int32 {
...
+		expected[i].Name = name
+	}
+	gomega.Eventually(func(g gomega.Gomega) {
+		updatedWl := &kueue.Workload{}
+		g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(wl), updatedWl)).To(gomega.Succeed())
+		g.Expect(workload.IsAdmitted(updatedWl)).To(gomega.BeTrue())
+		g.Expect(apimeta.FindStatusCondition(updatedWl.Status.Conditions, kueue.WorkloadEvicted)).To(gomega.BeNil())
+		g.Expect(updatedWl.Status.UnhealthyNodes).To(gomega.Equal(expected))
+	}, util.Timeout, util.Interval).Should(gomega.Succeed())
+}
```
tenzen-y, line 137: Can you move this into the util package? Also can you use ginkgo Helper to issue appropriate debug log?

pkg/features/kube_features.go
tenzen-y: +1

pkg/features/kube_features.go
tenzen-y, line 1: Also, can you add feature gate dependencies?

https://github.com/kubernetes-sigs/kueue/blob/83cc2e2a9f029eb8f933f175a598b9cbfb924493/pkg/features/kube_features.go#L671

pkg/cache/queue/manager.go
```diff
@@ -1040,10 +1047,22 @@ func (m *Manager) QueueSecondPassIfNeeded(ctx context.Context, w *kueue.Workload
...
+		var latest kueue.Workload
+		if err := m.client.Get(ctx, client.ObjectKeyFromObject(w), &latest); err == nil {
+			w = &latest
+		} else if apierrors.IsNotFound(err) {
+			m.DeleteSecondPass(workload.Key(w))
+			return
+		} else {
+			log.Error(err, "Failed to refresh workload before the second pass", "workload", workload.Key(w))
+		}
+	}
```
tenzen-y, line 1061: I suspect this refreshment mechanism has some problem due to non adjustment.
We typically adjust Workload w/ RuntimeClass, LendingLimit, ResourceTransformations, and DRA.

But, it seems that this mechanism doesn't follow anything, IIUC.

I was not totally clear on what we need here. Why do we need this refresh?


```suggestion
	if features.Enabled(features.TASReplaceMultipleFailedNodes) && workload.HasTopologyAssignmentWithUnhealthyNode(w) {
		var latest kueue.Workload
		if err := m.client.Get(ctx, client.ObjectKeyFromObject(w), &latest); err == nil {
			workload.AdjustResources(ctx, m.client, &latest)
			w = &latest
		} else if apierrors.IsNotFound(err) {
			m.DeleteSecondPass(workload.Key(w))
			return
		} else {
			log.Error(err, "Failed to refresh workload before the second pass", "workload", workload.Key(w))
		}
	}
```

pkg/cache/scheduler/tas_flavor_snapshot.go
tenzen-y, line 845: Can this respect multi-level topology requirements like [rack, hostname]?
This seems to be a compute replacement, excluding failure replacement.

1. PodSet was admitted w/ rack r1 and node x1.
2. x1 is broken, but r1 doesn't free capacity. Hence, replacement is pending.
3.  x2 is broken, and it is added to unhealthyNodes.
4. Kueue computes a replacement node for x2 w/ excluding x1.
5. requiredReplacementDomain looks up the replacement only for x2 and doesn't look up x3
6. Replacement searching selects r2 / x4 while expectation is r1 / x3

I guess that this leads to an invalid required topology assignment.

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -2373,7 +2420,7 @@ func (s *TASFlavorSnapshot) mergeTopologyAssignments(a, b *utiltas.TopologyAssig
 		aDomain := s.domainForAssignmentValues(levels, a.Values)
 		bDomain := s.domainForAssignmentValues(levels, b.Values)
 		if aDomain == nil || bDomain == nil {
-			// Defensive: staleness is verified before merging.
+			// Queued unhealthy nodes may already be absent from the snapshot.
 			return cmp.Compare(utiltas.DomainID(a.Values), utiltas.DomainID(b.Values))
```
tenzen-y, line 2424: Shouldn't we have the following fallback, since a part of unhealthyNodes can be dropped due to the failure of replacements?

```suggestion
        aID, bID := utiltas.DomainID(a.Values), utiltas.DomainID(b.Values)
        if aDomain := s.domainForAssignmentValues(levels, a.Values); aDomain != nil {
            aID = utiltas.DomainID(aDomain.levelValues)
        }
        if bDomain := s.domainForAssignmentValues(levels, b.Values); bDomain != nil {
            bID = utiltas.DomainID(bDomain.levelValues)
        }
        return cmp.Compare(aID, bID)
```

tenzen-y (commented):
Thank you for working on this issue, I left some comments.

pkg/cache/queue/manager.go
mimowo: This fix is for pre-existing issue, not de-coupled to https://github.com/kubernetes-sigs/kueue/pull/15571

mimowo:
@tg123 do you have cycles to address the comments? We expect all KEPS to be code complete EOW. Otherwise we can revisit this enhancement for v0.21.

tg123:
Will close it and clean up this week 
Sorry for the delay

tg123:
AI-assisted CI triage: integration shard 3 failed in the unchanged FairSharing test 'should clear metrics for the implicit root Cohort after switching to another root Cohort' (metrics_test.go:1086). The root1 active-workload gauge stayed at 0 instead of 3 before reparenting. Retrying only this shard to check whether the failure reproduces.

/test pull-kueue-test-integration-shard-3-main

pkg/scheduler/scheduler.go
tg123: here

https://github.com/kubernetes-sigs/kueue/pull/12344/changes/BASE..0b5439f0416b38c94920d56c0f674d4bc7935884#diff-fb3d6b36d3cd727e1dd2353fb86c0aee0bfaae1c900e0f005ab913620021d551L1051

Image
moved to `patchWorkloadAdmission`


this is to protect `unhealthy nodes`, i am not 100% sure if it is safe to touch non `TASReplaceMultipleFailedNodes`, thus, preserving existing behavior when the gate is off

tg123:
AI-assisted CI triage: the website-preview link check failed with HTTP 500/503 responses from Netlify, not missing-link errors. The sampled failing URLs now return HTTP 200. Retrying only the website-preview check.

/test pull-kueue-verify-website-links-preview-main

pkg/cache/scheduler/tas_flavor_snapshot.go
tg123: please help double check if 9b9257864dc88bb2d56bb8308291073003160c28 helps

tg123:
@mimowo @tenzen-y please check, i will also push internal upgrade to 0.20

apis/kueue/v1beta2/topology_types.go
```diff
@@ -74,6 +74,20 @@ const (
 	// during the Pod creation.
 	TopologySchedulingGate = "kueue.x-k8s.io/topology"
 
+	// UnhealthyNodesConcurrentEvictionThresholdAnnotation sets the maximum number
+	// of unhealthy nodes retained for replacement without triggering workload eviction.
+	// Valid values are 1-8; absent values default to 1. Invalid values are rejected
+	// on creation or annotation changes. Existing invalid values default to 1.
+	// Ignored unless TASReplaceMultipleFailedNodes is enabled.
+	UnhealthyNodesConcurrentEvictionThresholdAnnotation = "kueue.x-k8s.io/unhealthy-nodes-concurrent-eviction-threshold"
```
mimowo:
```suggestion
	MaxConcurrentUnhealthyNodesAnnotation = "kueue.x-k8s.io/max-concurrent-unhealthy-nodes"
```
I want to propose shorter name which is also descriptive - "max" is shorter than "threshold". Also eviction is implied I think. 

If we want to be super precise, then `max-concurrent-unhealthy-nodes-before-eviction`.

wdyt @tenzen-y ?

pkg/scheduler/scheduler.go
```diff
@@ -1211,6 +1220,47 @@ func (s *Scheduler) admit(ctx context.Context, e *entry, cq *schdcache.ClusterQu
...
+	cq *schdcache.ClusterQueueSnapshot,
+	admission *kueue.Admission,
+) error {
+	replacedNodeName := ""
+	if len(wl.Status.UnhealthyNodes) > 0 {
+		replacedNodeName = wl.Status.UnhealthyNodes[0].Name
+	}
+	patchOptions := []workloadpatching.PatchStatusOption{workloadpatching.WithRetryOnConflict()}
+	if !features.Enabled(features.TASReplaceMultipleFailedNodes) {
+		patchOptions = append(patchOptions, workloadpatching.WithLooseOnApply())
```
mimowo: Could we include `patchOptions = append(patchOptions, workloadpatching.WithLooseOnApply())` unconditionally? This shouldn't really depend on the new FG.

test/integration/singlecluster/tas/tas_test.go
```diff
@@ -3613,6 +3614,478 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
 					}, util.Timeout, util.Interval).Should(gomega.Succeed())
 				})
 			})
+
+			ginkgo.It("should evict when the number of unhealthy nodes exceeds a threshold greater than 1", framework.SlowSpec, func() {
```
mimowo: Could you also add an integration test for an actual job controller running? I'm not clear how the propagation would work of the annotation from Job to workload. I suppose we might be missing the propagation here. Please check how this is done in https://github.com/kubernetes-sigs/kueue/pull/15503

pkg/scheduler/scheduler.go
```diff
@@ -1211,6 +1220,47 @@ func (s *Scheduler) admit(ctx context.Context, e *entry, cq *schdcache.ClusterQu
...
+	ctx context.Context,
+	log logr.Logger,
+	wl *kueue.Workload,
+	cq *schdcache.ClusterQueueSnapshot,
+	admission *kueue.Admission,
+) error {
+	replacedNodeName := ""
+	if len(wl.Status.UnhealthyNodes) > 0 {
+		replacedNodeName = wl.Status.UnhealthyNodes[0].Name
+	}
```
mimowo: Don't we have a helper which returns first unhealthyNode name already?

PR #12289, opened 2026-06-16
Prepare MultiKueue Kind dev setup to use CA-based kubeconfigs

apullo777:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

This is a preparatory PR for #12284.

The documented Kind-based MultiKueue development setup currently generates worker kubeconfigs with `insecure-skip-tls-verify: true` and relies on the deprecated `MultiKueueAllowInsecureKubeconfigs` feature gate.

This PR updates that dev setup to generate CA-based kubeconfigs using `certificate-authority-data` extracted from the Kind worker cluster, so the documented workflow no longer depends on insecure kubeconfigs.

It does not change the feature-gate lifecycle yet.

#### Which issue(s) this PR fixes:

Part of #12284

#### Special notes for your reviewer:

The `setup-kind-multikueue-tas.sh` change is only a stale-comment fix; the script already used `certificate-authority-data`.

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

Issue #12284, opened 2026-06-16
Remove deprecated MultiKueueAllowInsecureKubeconfigs


mimowo:
**What would you like to be cleaned**:

Remove deprecated MultiKueueAllowInsecureKubeconfigs. It was deprecated in 0.17 and we haven't heard any issue with that.

**Why is this needed**:

To keep our code healthy

apullo777:
I'd like to take this :)
/assign

apullo777:
Hi @mimowo while working on this, I found that direct removal is blocked by the feature-gate lifecycle tooling:

`MultiKueueAllowInsecureKubeconfigs` is `Deprecated`, but its last spec is not `LockToDefault: true`, so `make verify` rejects removing it.

Would you prefer this PR to do the required intermediate step instead?

* keep the gate recognized;
* set `LockToDefault: true` with default `false`;
* make the secure kubeconfig validation path unconditional, so the insecure path is no longer reachable;
* leave full removal for a follow-up release/PR.

I also noticed that the MultiKueue dev setup still relies on this gate because it generates kubeconfigs with `insecure-skip-tls-verify: true`. Should this PR also update that workflow to generate a CA-based kubeconfig using `certificate-authority-data`, so the documented Kind setup keeps working after the gate is locked?

mimowo:
The interim step sounds good, we could keep it around with `LockToDefault: true` for another release. 

> I also noticed that the MultiKueue dev setup still relies on this gate because it generates kubeconfigs with insecure-skip-tls-verify: true. Should this PR also update that workflow to generate a CA-based kubeconfig using certificate-authority-data, so the documented Kind setup keeps working after the gate is locked?

This sounds like more of a problem, because maybe it can add a lot of complication to our testing, I'm not sure yet. If this complicates the setup for testing too much then maybe we should consider keeping the FG. I would like to see what that entails in a preparatory PR even

kannon92:
The gate remains registered as a locked no-op for 0.19 and can be removed in 0.20.

PR #12366, opened 2026-06-19
Lock the deprecated MultiKueueAllowInsecureKubeconfigs feature gate to its default

apullo777:
#### What type of PR is this?

/kind cleanup
/area multikueue

`MultiKueueAllowInsecureKubeconfigs` was deprecated in 0.17 but cannot be removed yet because it was never locked. This PR locks it to the default value (`false`) in 0.19, makes `validateKubeconfig` unconditional, and removes the now-unreachable insecure path and tests.

The gate remains registered as a locked no-op for 0.19 and can be removed in 0.20.

Follow-up to #12289, which moved the documented Kind dev setup to `certificate-authority-data`.

Part of #12284

Feature lists regenerated via `make generate-featuregates`

```release-note
MultiKueue: `MultiKueueAllowInsecureKubeconfigs` is now locked to disabled. Remove `--feature-gates=MultiKueueAllowInsecureKubeconfigs=true` and use kubeconfigs with `certificate-authority-data` instead.
```


mbobrovskyi (commented):
Thank you!

/lgtm

mimowo (commented):
/approve
Thank you for the cleanup 👍

PR #12378, opened 2026-06-19
Add flavor assigner diagnostic reasons for NoFit results

j-skiba:
#### What type of PR is this?
/kind feature

This PR implements the scheduler-side logic specified in [KEP-10852](https://github.com/kubernetes-sigs/kueue/blob/main/keps/10852-inadmissible-workloads-observability) to track why a workload cannot fit within any resource flavors. It satisfies the following points from the KEP:

- Distinguish configuration faults vs. normal capacity wait
     - Adds `NoFitReason` tracking to the individual flavor assignment attempts.
     - Classifies attempts as:
          - `NoMatchingFlavor` for structural mismatches (node selectors, taints, or topology constraints).
          - `ExceedsMaxQuota` when exceeding absolute limits (cohort max capacity, or `ClusterQueue` nominal + borrowingLimit).
          - `WaitingForQuota` when waiting for preemption or cohort quota availability.
          - `TopologyPlacementFailed` when TAS placement requirements cannot be satisfied.
- Tiered Reason Precedence & Multi-PodSet Resolution
   - Implements the precedence tiers outlined in [Nominated Flavor Reasons](https://github.com/kubernetes-sigs/kueue/tree/main/keps/10852-inadmissible-workloads-observability#2-nominated-flavor-reasons) using a `reasonSeverity` ranking.
   - Provides `Assignment.resolveNoFitReason()` to calculate the final `NoFitReason` for the workload:
        - Alternatives (within a Resource Group): Resolves to the minimum severity blocker (best chance of success).
        - Co-requisites (across Resource Groups): Resolves to the maximum severity blocker (worst-case blocker).

Part of #10852

```release-note
NONE
```

> AI assistance disclosure: this change was developed with the assistance of an AI coding tool.

## Release Notes

* **New Features**
  * Added the `UnadmittedWorkloadsObservability` feature gate to enable more detailed scheduling “no-fit” diagnostics.

* **Bug Fixes**
  * Expanded “no-fit” reason reporting for quota/resource flavor matching and topology placement failures.
  * Improved flavor attribution in topology assignment failure details so returned diagnostics are labeled with the correct flavor.

* **Tests**
  * Updated scheduling diagnostics tests to validate top-level and per-flavor reason fields, including feature-gate on/off behavior.

### DIFF
--- a/apis/kueue/v1beta2/workload_types.go
+++ b/apis/kueue/v1beta2/workload_types.go
@@ -957,6 +957,34 @@ const (
 	// requirements that cannot be satisfied with the current cluster topology usage.
 	WorkloadQuotaReservedReasonTopologyPlacementFailed = "TopologyPlacementFailed"
 
+	// WorkloadQuotaReservedReasonWaitingForPreemptedWorkloads indicates that the workload is waiting
+	// for preempted workloads to release their quota.
+	WorkloadQuotaReservedReasonWaitingForPreemptedWorkloads = "WaitingForPreemptedWorkloads"
+
+	// WorkloadQuotaReservedReasonMisconfigured indicates that the workload is inadmissible due to
+	// misconfiguration, such as missing LocalQueue or ClusterQueue.
+	WorkloadQuotaReservedReasonMisconfigured = "Misconfigured"
+
+	// WorkloadQuotaReservedReasonSuspended indicates that the workload is inadmissible because
+	// the LocalQueue or ClusterQueue StopPolicy is active.
+	WorkloadQuotaReservedReasonSuspended = "Suspended"
+
+	// WorkloadQuotaReservedReasonPendingEvaluation indicates that the workload is pending evaluation in the scheduling queue.
+	WorkloadQuotaReservedReasonPendingEvaluation = "PendingEvaluation"
+
+	// WorkloadQuotaReservedReasonWaitingForPodsReady indicates that the workload is waiting
+	// for previously admitted workloads to reach PodsReady condition under waitForPodsReady configuration.
+	WorkloadQuotaReservedReasonWaitingForPodsReady = "WaitingForPodsReady"
+
+	// WorkloadAdmittedReasonNoReservation indicates that the workload has no reservation.
+	WorkloadAdmittedReasonNoReservation = "NoReservation"
+
+	// WorkloadAdmittedReasonUnsatisfiedAdmissionChecks indicates that the workload has not all checks ready.
+	WorkloadAdmittedReasonUnsatisfiedAdmissionChecks = "UnsatisfiedAdmissionChecks"
+
+	// WorkloadAdmittedReasonPendingDelayedTopologyRequests indicates that there are pending delayed topology requests.
+	WorkloadAdmittedReasonPendingDelayedTopologyRequests = "PendingDelayedTopologyRequests"
+
 	// WorkloadFinished means that the workload associated to the
 	// ResourceClaim finished running (failed or succeeded).
 	WorkloadFinished = "Finished"
--- a/apis/kueue/v1beta2/workload_types.go
+++ b/apis/kueue/v1beta2/workload_types.go
@@ -938,6 +938,25 @@ const (
 	// WorkloadQuotaReserved means that the Workload has reserved quota a ClusterQueue.
 	WorkloadQuotaReserved = "QuotaReserved"
 
+	// Reasons for the WorkloadQuotaReserved condition.
+
+	// WorkloadQuotaReservedReasonNoMatchingFlavor indicates that the workload cannot be scheduled
+	// because no resource flavor matches its nodeSelector or taints.
+	WorkloadQuotaReservedReasonNoMatchingFlavor = "NoMatchingFlavor"
+
+	// WorkloadQuotaReservedReasonWaitingForQuota indicates that the workload is waiting for
+	// sufficient unused quota to become available in the ClusterQueue/Cohort.
+	WorkloadQuotaReservedReasonWaitingForQuota = "WaitingForQuota"
+
+	// WorkloadQuotaReservedReasonExceedsMaxQuota indicates that the workload requests resources
+	// exceeding the maximum capacity limits of the ClusterQueue or Cohort.
+	// This also includes exceeding ClusterQueue nominal + borrowingLimit constraints.
+	WorkloadQuotaReservedReasonExceedsMaxQuota = "ExceedsMaxQuota"
+
+	// WorkloadQuotaReservedReasonTopologyPlacementFailed indicates that the workload has topology
+	// requirements that cannot be satisfied with the current cluster topology usage.
+	WorkloadQuotaReservedReasonTopologyPlacementFailed = "TopologyPlacementFailed"
+
 	// WorkloadFinished means that the workload associated to the
 	// ResourceClaim finished running (failed or succeeded).
 	WorkloadFinished = "Finished"
--- a/apis/kueue/v1beta2/workload_types.go
+++ b/apis/kueue/v1beta2/workload_types.go
@@ -957,34 +957,6 @@ const (
 	// requirements that cannot be satisfied with the current cluster topology usage.
 	WorkloadQuotaReservedReasonTopologyPlacementFailed = "TopologyPlacementFailed"
 
-	// WorkloadQuotaReservedReasonWaitingForPreemptedWorkloads indicates that the workload is waiting
-	// for preempted workloads to release their quota.
-	WorkloadQuotaReservedReasonWaitingForPreemptedWorkloads = "WaitingForPreemptedWorkloads"
-
-	// WorkloadQuotaReservedReasonMisconfigured indicates that the workload is inadmissible due to
-	// misconfiguration, such as missing LocalQueue or ClusterQueue.
-	WorkloadQuotaReservedReasonMisconfigured = "Misconfigured"
-
-	// WorkloadQuotaReservedReasonSuspended indicates that the workload is inadmissible because
-	// the LocalQueue or ClusterQueue StopPolicy is active.
-	WorkloadQuotaReservedReasonSuspended = "Suspended"
-
-	// WorkloadQuotaReservedReasonPendingEvaluation indicates that the workload is pending evaluation in the scheduling queue.
-	WorkloadQuotaReservedReasonPendingEvaluation = "PendingEvaluation"
-
-	// WorkloadQuotaReservedReasonWaitingForPodsReady indicates that the workload is waiting
-	// for previously admitted workloads to reach PodsReady condition under waitForPodsReady configuration.
-	WorkloadQuotaReservedReasonWaitingForPodsReady = "WaitingForPodsReady"
-
-	// WorkloadAdmittedReasonNoReservation indicates that the workload has no reservation.
-	WorkloadAdmittedReasonNoReservation = "NoReservation"
-
-	// WorkloadAdmittedReasonUnsatisfiedAdmissionChecks indicates that the workload has not all checks ready.
-	WorkloadAdmittedReasonUnsatisfiedAdmissionChecks = "UnsatisfiedAdmissionChecks"
-
-	// WorkloadAdmittedReasonPendingDelayedTopologyRequests indicates that there are pending delayed topology requests.
-	WorkloadAdmittedReasonPendingDelayedTopologyRequests = "PendingDelayedTopologyRequests"
-
 	// WorkloadFinished means that the workload associated to the
 	// ResourceClaim finished running (failed or succeeded).
 	WorkloadFinished = "Finished"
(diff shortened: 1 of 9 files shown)


pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -761,9 +814,57 @@ func (a *FlavorAssigner) assignFlavors(log logr.Logger, counts []int32) Assignme
 			}
 		}
 	}
+	assignment.resolveNoFitReason(a.cq)
 	return assignment
 }
 
+func (a *Assignment) resolveNoFitReason(cq *schdcache.ClusterQueueSnapshot) {
```
mimowo, line 824: This logic is already pretty haevy, could we wind it under the new FG?  Also, the new reasons seem like already part of the KEP.

j-skiba: 👍 sounds good, I wasn't sure if introducing FG when the feature is not completely implemented is ok. I will change that

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -246,9 +249,46 @@ func IgnoreUndeclaredResources(quotaCheckStrategy configapi.QuotaCheckStrategy)
 	return features.Enabled(features.QuotaCheckStrategy) && quotaCheckStrategy == configapi.QuotaCheckIgnoreUndeclared
 }
 
+func worstReason(a, b string) string {
```
kshalot: I'd consider renaming to something like `mostSevereReason`, because `worst` is not very precise.

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -246,9 +249,46 @@ func IgnoreUndeclaredResources(quotaCheckStrategy configapi.QuotaCheckStrategy)
...
+	err         error
+	NoFitReason string
+}
+
+func (s *Status) MarkNoMatchingFlavor() *Status {
+	if s != nil {
+		s.NoFitReason = kueue.WorkloadQuotaReservedReasonNoMatchingFlavor
+	}
+	return s
+}
```
kshalot: Nit: We can consider inlining this in the one call-site where it's used. We are currently doing:
```
			flavorStatus.MarkNoMatchingFlavor()
```
which discards the return value and performs the nil check on something that is never nil. If this is useful in the next steps of implementation that feel free to ignore this comment.

j-skiba: we can drop this function. I removed that

pkg/scheduler/flavorassigner/flavorassigner.go
j-skiba: I agree, it sounds better, renamed

pkg/scheduler/flavorassigner/flavorassigner.go
j-skiba: I added the FGs. I disabled them by default for now as the feature is not fully implemented, is that ok?

mimowo: sgtm, we can update to Beta once the feature is fully done, before 0.19

pkg/features/kube_features.go
```diff
@@ -680,6 +695,12 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 	TASHandleOverlappingFlavors: {
 		{Version: version.MustParse("0.18"), Default: true, PreRelease: featuregate.Beta},
 	},
+	UnadmittedWorkloadsObservability: {
+		{Version: version.MustParse("0.19"), Default: false, PreRelease: featuregate.Beta},
+	},
+	UnadmittedWorkloadsExplicitStatus: {
+		{Version: version.MustParse("0.19"), Default: false, PreRelease: featuregate.Beta},
+	},
```
kshalot: Should we start with alpha or beta?

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -246,9 +249,32 @@ func IgnoreUndeclaredResources(quotaCheckStrategy configapi.QuotaCheckStrategy)
...
+	case kueue.WorkloadQuotaReservedReasonNoMatchingFlavor:
+		return 4
+	case kueue.WorkloadQuotaReservedReasonExceedsMaxQuota:
+		return 3
+	case kueue.WorkloadQuotaReservedReasonWaitingForQuota:
+		return 2
+	case kueue.WorkloadQuotaReservedReasonTopologyPlacementFailed:
+		return 1
+	default:
+		return 0
```
mimowo: Maybe we can use iota for the named constants representing the numbers?

apis/kueue/v1beta2/workload_types.go
```diff
@@ -938,6 +938,53 @@ const (
 	// WorkloadQuotaReserved means that the Workload has reserved quota a ClusterQueue.
 	WorkloadQuotaReserved = "QuotaReserved"
 
+	// Reasons for the WorkloadQuotaReserved condition.
```
mimowo, line 941: Would it be feasible to break it down into smaller PRs to make the review easier? We can keep this full PR, but I'm wondering about such steps:

1. UnadmittedWorkloadsExplicitStatus together with support
2. UnadmittedWorkloadsObservability exdended for flavor nominated reaons
3. UnadmittedWorkloadsObservability introduced for flavor-independent reasons

Another order would also be fine. wdyt? I think it it much easier for bugs to slip in larger PRs

apis/kueue/v1beta2/workload_types.go
```diff
@@ -938,6 +938,53 @@ const (
...
+	// exceeding the maximum capacity limits of the ClusterQueue or Cohort.
+	// This also includes exceeding ClusterQueue nominal + borrowingLimit constraints.
+	WorkloadQuotaReservedReasonExceedsMaxQuota = "ExceedsMaxQuota"
+
+	// WorkloadQuotaReservedReasonTopologyPlacementFailed indicates that the workload has topology
+	// requirements that cannot be satisfied with the current cluster topology usage.
+	WorkloadQuotaReservedReasonTopologyPlacementFailed = "TopologyPlacementFailed"
+
+	// WorkloadQuotaReservedReasonWaitingForPreemptedWorkloads indicates that the workload is waiting
+	// for preempted workloads to release their quota.
```
mwielgus: What about waiting for Topology (for topology-driven preemptions)?

apis/kueue/v1beta2/workload_types.go
j-skiba: I was thinking to do something like this in separate PRs:
1. this branch - just assigning `NoFit` reasons for flavor dependent causes
2. surfacing the reasons introduced in this PR. It will be mostly changes in scheduler.go and adjustment to unit tests of the scheduler. The tests would be run with the `UnadmittedWorkloadsObservability` FG disabled and then enabled.
3. handling the rest of the reasons. That will be mostly changes in workload_controller.go and tests - both unit and integration tests
4. add support for `UnadmittedWorkloadsExplicitStatus`
5. add metrics
6. enabled FGs by default

j-skiba: and
7. backport to 0.18

mimowo: This sounds reasonable. So, IIUC in this PR we just introduce the reason and the logic for their aggregation, but without surfacing them to the user level yet, right?

So my questions would be:
- does the PR have some  user-observable impact already? 
- what is the testing plan for the various cases, could we test them already by asserting on messages?

apis/kueue/v1beta2/workload_types.go
```diff
@@ -938,6 +938,53 @@ const (
...
+	WorkloadQuotaReservedReasonWaitingForPreemptedWorkloads = "WaitingForPreemptedWorkloads"
+
+	// WorkloadQuotaReservedReasonMisconfigured indicates that the workload is inadmissible due to
+	// misconfiguration, such as missing LocalQueue or ClusterQueue.
+	WorkloadQuotaReservedReasonMisconfigured = "Misconfigured"
```
mimowo: This looks unused for now. Let's drop the unused ones out, so that we don't forget to add tests for them once they become used.

apis/kueue/v1beta2/workload_types.go
j-skiba: > This sounds reasonable. So, IIUC in this PR we just introduce the reason and the logic for their aggregation, but without surfacing them to the user level yet, right?

that's correct.

> does the PR have some user-observable impact already?

no, there's no yet

> what is the testing plan for the various cases, could we test them already by asserting on messages?

I have a plan for PRs 2 and 3 to include a separate field `wantConditionsWithUnadmittedWorkloadsFG` (or something like this) for the already existing unit tests we have and run them once with enabled and once with disabled. It would be kind of similar what we do with `WorkloadRequestUseMergePatch`. For integration tests I think we could modify the already existing ones and implement some new ones to improve coverage.

apis/kueue/v1beta2/workload_types.go
j-skiba: yes, it should be there. For now I removed that completely as it wasn't used in the PR. I will update it in the following PRs

apis/kueue/v1beta2/workload_types.go
j-skiba: removed unused reasons and `UnadmittedWorkloadsExplicitStatus` FG

pkg/scheduler/flavorassigner/flavorassigner.go
j-skiba: done

pkg/features/kube_features.go
j-skiba: we start with beta - https://github.com/kubernetes-sigs/kueue/tree/main/keps/10852-inadmissible-workloads-observability#api-and-backwards-compatibility

but later we backport as alpha to 0.18

pkg/scheduler/flavorassigner/flavorassigner_test.go
```diff
@@ -4934,3 +5090,473 @@ func TestAssignFlavorsWithAllowedFlavors(t *testing.T) {
...
+				"flavor-a": "",
+				"flavor-b": "NoMatchingFlavor",
+			},
+		},
+		"insufficient quota": {
+			podSet:          *utiltestingapi.MakePodSet("main", 1).Request(corev1.ResourceCPU, "3").Obj(),
+			wantNoFitReason: "ExceedsMaxQuota",
+			wantFlavorAttempts: map[kueue.ResourceFlavorReference]string{
+				"flavor-a": "ExceedsMaxQuota",
+				"flavor-b": "NoMatchingFlavor",
```
mimowo, line 5177: This makes me wonder there is something off with "NoMatchingFlavor" - because "NoMatchingFlavor" is surfaced per flavor, but it only makes sense as a global property.

Maybe we we need two reasons:
- NonMatchingFlavor - per flavor
- NoMatchingFlavor / Misconfigured - aggregated reason when all are NonMatchingFlavor

wdyt?

pkg/scheduler/flavorassigner/flavorassigner_test.go
```diff
@@ -3369,90 +3415,186 @@ func TestAssignFlavors(t *testing.T) {
...
+					Quota: resources.FlavorResourceQuantities{
+						{Flavor: "one", Resource: corev1.ResourceCPU}: resources.NewAmount(1000),
+					},
 				},
-			})
+			},
+		},
+	}
+	for name, tc := range cases {
+		for _, enabled := range []bool{false, true} {
```
mimowo: Make the name more informative, maybe `unadmittedWorkloadsObservabilityEnabled`

apis/kueue/v1beta2/workload_types.go
mimowo: I see, thanks for explaining, I think the plan is reasonable. 

> I have a plan for PRs 2 and 3 to include a separate field wantConditionsWithUnadmittedWorkloadsFG (or something like this) for the already existing unit tests we have and run them once with enabled and once with disabled. It would be kind of similar what we do with WorkloadRequestUseMergePatch.

This sounds ok, but maybe in case of the new feature we would write the "wantReason" as the new already, but if the FG is disabled, then assert on the old "Pending" / "Inadmissible".

pkg/scheduler/flavorassigner/flavorassigner_test.go
j-skiba: I would say it's ok as it is. First we take the least severe problem within flavors for specific resource group (the closest to succeeding) and then across all resource groups and podsets we choose the strongest reason and set it as a underlying reason of `NoFit` result.

I think `NonMatchingFlavor` doesn't give us any extra info over `NoMatchingFlavor`.

pkg/scheduler/flavorassigner/flavorassigner_test.go
j-skiba: renamed as advised

pkg/scheduler/flavorassigner/flavorassigner_test.go
mimowo: Ack, let's start with that. I makes code simpler and the internal No matching flavor per flavor is not surfacing to users anyway

mimowo:
/lgtm 
/approve
Thank you for the first foundation PR.

PR #2196, opened 2024-05-14
Provide support for ProvisioningRequest's CapacityRevoked condition

PBundyra:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
Provide support for the new ProvisioningRequest's condition `CapacityRevoked`. When user uses a job that allows retries and sets `.spec.backOffLimit` > 0 and the job gets evicted because of `MaxRunDuration` an according AdmissionCheck gets rejected.

After a Workload is finished we no longer delete corresponding ProvisioningRequest. This is because Workload might be finished due to `MaxRunDuration` timeout, before ClusterAutoscaler updates ProvisioningRequest's status to `CapacityRevoked` and the Kueue's controller has no opportunity to react to that.

Additionally I did a small cleanup.

#### Which issue(s) this PR fixes:

Part of #2041 

#### Special notes for your reviewer:
I've tested it e2e manually

#### Does this PR introduce a user-facing change?

```release-note
Kueue now supports ProvisioningRequest's condition `CapacityRevoked`. ProvisioningRequests objects persist until the corresponding Job or the Workload is deleted
```

PR #12401, opened 2026-06-20
test: fix provisioning request retention assertion on workload finish

akshay-pm:

/kind failing-test
/area testing

Corrects the test assertion for the case where a workload reaches the `Finished` state while a ProvisioningRequest is still present.

Due to a race condition (#2196) where a workload can complete due to MaxRunDuration timeout before ClusterAutoscaler sets `CapacityRevoked=true`, ProvisioningRequests are intentionally preserved after a workload finishes so that Kueue can still react to that condition.

The previous test checked for `"wl-check1"` instead of `"wl-check1-1"`, missing the attempt suffix, resulting in the  assertion passing trivially. The test is now updated to assert the correct behavior: the ProvisioningRequest `"wl-check1-1"` is preserved after reconciliation of a finished workload.

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where `ProvisioningRequest` resources were being incorrectly removed when workloads completed. These resources now persist after workload completion.

### DIFF
--- a/pkg/controller/admissionchecks/provisioning/controller_test.go
+++ b/pkg/controller/admissionchecks/provisioning/controller_test.go
@@ -541,7 +541,7 @@ func TestReconcile(t *testing.T) {
 				},
 			},
 		},
-		"request removed on workload finished": {
+		"request not removed on workload finished": {
 			workload: (&utiltestingapi.WorkloadWrapper{Workload: *baseWorkload.DeepCopy()}).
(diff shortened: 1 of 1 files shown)


mimowo:
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for the cleanup, and the helpful explanation in the PR description 👍

PR #12403, opened 2026-06-21
Inject ResourceFlavor nodeLabels and tolerations in ElasticJobUngater

### LINKED ISSUE #12350: ElasticJobsViaWorkloadSlices: ResourceFlavor nodeLabels not injected on elastic job pods
**What happened**:

Elastic jobs (with `ElasticJobsViaWorkloadSlices` enabled) admitted to a ResourceFlavor with `nodeLabels` never receive those labels on their pods. The pods keep only the nodeSelector set by the workload owner. If the owner strips the nodepool selector (to let Kueue pick the flavor), pods land on any node matching the accelerator label, including nodes outside the target pool.

The standard path injects nodeLabels via `startJob` -> `RunWithPodSetsInfo` -> `podset.Merge`, which calls `podset.FromAssignment` to collect labels from each assigned flavor. The elastic path bypasses `startJob`. The job is never suspended, so step 7 of `ReconcileGenericJob` skips it. The `ElasticJobUngater` then removes the scheduling gate from pods but never injects nodeLabels or tolerations from the assigned ResourceFlavor.

This is the same class of bug as #10167 (missing tolerations) and #11977 (ungater bypasses quota). The ungater touches pods but does not apply the full PodSetInfo that `podset.Merge` applies on the standard path.

Pods admitted to a ResourceFlavor should receive that flavor's `nodeLabels` and `tolerations`, whether the job uses the standard path or the elastic path. The elastic ungater should call `podset.FromAssignment` and merge the result into each pod before removing the scheduling gate.

2. Create a ResourceFlavor with two `nodeLabels`: `cloud.google.com/gke-accelerator` and `cloud.google.com/gke-nodepool`.
3. Create a RayCluster with `kueue.x-k8s.io/elastic-job: "true"` and a worker pod template whose `nodeSelector` has only `cloud.google.com/gke-accelerator` (the nodepool label omitted, to let Kueue place it).
4. Submit the RayCluster to a ClusterQueue that uses that flavor, so Kueue admits it.
5. The pods are ungated and scheduled. Their `nodeSelector` lacks the `gke-nodepool` label from the ResourceFlavor. They land on any node pool with the matching accelerator, not just the reserved one.

- Reproduces on v0.16.9, v0.17.5, and v0.18.1. The `ElasticJobUngater` is unchanged in all three.
- If the client sets the nodepool label on the pod template directly, RayCluster workers are unaffected: the label survives pod creation. The bug bites only when the owner omits the nodepool label and relies on Kueue to inject it.
- Related issues: #10167 (elastic jobs missing tolerations from ResourceFlavor), #11977 (ungater bypasses quota by ungating all pods in the slice chain). Both stem from the ungater doing less than `podset.Merge` does on the standard path.

**Environment**:
- Kubernetes version: GKE 1.33
- Kueue version: v0.16.9 (also confirmed on v0.18.1 source)
- Cloud provider: GKE
- Others: KubeRay v1.6.0, `ElasticJobsViaWorkloadSlices=true`

ikchifo:
#### What type of PR is this?

/kind bug
/area integrations

Elastic jobs that bypass `startJob` (RayCluster with autoscaling, batch/v1
Job with the elastic annotation) relied on the `ElasticJobUngater` only to
remove the `ElasticJobSchedulingGate`. The ungater never applied the
`PodSetInfo` that `startJob` injects via `RunWithPodSetsInfo` and
`podset.Merge`. So pods admitted to a ResourceFlavor with `nodeLabels`
never got those labels, and could land outside the target node pool.

The ungater now calls `jobframework.GetPodSetsInfoFromStatus` (exported
for this purpose) and merges the result into each pod before removing the
gate. Reusing the standard path also picks up admission-check
`PodSetUpdates`, the `PodSetLabel`, and slice/queue labels.

A missing ResourceFlavor now makes the reconcile retry, rather than
ungating pods without injection.

Fixes #12350

Pod-to-PodSet mapping: by `kueue.x-k8s.io/podset` label first; falls back
to the single PodSet when the workload has one PodSet and the label is
absent (the issue's RayCluster-worker scenario); otherwise logs and
ungates without injection.

```release-note
`ElasticJobsViaWorkloadSlices`: fixed pods of elastic jobs not receiving
the assigned ResourceFlavor's nodeLabels and tolerations, which could
land them outside the target node pool. A missing ResourceFlavor now
makes the ungater retry rather than silently ungating pods.
```

## Release Notes

* **New Features**
  * Elastic jobs now inject PodSet “flavor” information during pod ungating, applying the correct node labels and tolerations from assigned resource flavors to improve scheduling reliability.

* **Tests**
  * Updated and expanded elastic job ungating coverage for single and multi–PodSet scenarios, including cases without PodSet labeling and situations where required resource flavors are missing (now validated to fail safely).

### DIFF
--- a/pkg/controller/elasticjobs/elastic_job_ungater.go
+++ b/pkg/controller/elasticjobs/elastic_job_ungater.go
@@ -169,7 +169,7 @@ func (r *elasticJobUngater) Reconcile(ctx context.Context, req reconcile.Request
 			}
 			if info, ok := infoFor(pod); ok {
 				if mergeErr := podset.Merge(&pod.ObjectMeta, &pod.Spec, info); mergeErr != nil {
-					log.Error(mergeErr, "failed merging PodSet info before ungating; ungating without flavor injection",
+					log.Error(mergeErr, "failed merging PodSet info after ungating; pod released without flavor injection",
 						"pod", klog.KObj(pod))
 				} else {
 					log.V(3).Info("ungating elastic pod with assigned flavor info",
--- a/pkg/controller/elasticjobs/elastic_job_ungater.go
+++ b/pkg/controller/elasticjobs/elastic_job_ungater.go
@@ -132,8 +132,10 @@ func (r *elasticJobUngater) Reconcile(ctx context.Context, req reconcile.Request
 		return reconcile.Result{}, fmt.Errorf("building PodSet info for elastic ungater: %w", err)
 	}
 	infoByPodSet := make(map[kueue.PodSetReference]podset.PodSetInfo, len(infos))
+	flavorsByPodSet := make(map[kueue.PodSetReference]map[corev1.ResourceName]kueue.ResourceFlavorReference, len(infos))
 	for i := range infos {
 		infoByPodSet[infos[i].Name] = infos[i]
+		flavorsByPodSet[infos[i].Name] = wl.Status.Admission.PodSetAssignments[i].Flavors
 	}
 	var singleInfo *podset.PodSetInfo
 	if len(infos) == 1 {
@@ -167,13 +169,14 @@ func (r *elasticJobUngater) Reconcile(ctx context.Context, req reconcile.Request
 			}
 			if info, ok := infoFor(pod); ok {
 				if mergeErr := podset.Merge(&pod.ObjectMeta, &pod.Spec, info); mergeErr != nil {
-					log.V(2).Info("failed merging PodSet info before ungating; ungating without flavor injection",
-						"pod", klog.KObj(pod), "error", mergeErr)
+					log.Error(mergeErr, "failed merging PodSet info before ungating; ungating without flavor injection",
+						"pod", klog.KObj(pod))
 				} else {
-					log.V(3).Info("ungating elastic pod with assigned flavor info", "pod", klog.KObj(pod))
+					log.V(3).Info("ungating elastic pod with assigned flavor info",
+						"pod", klog.KObj(pod), "flavors", flavorsByPodSet[info.Name])
 				}
 			} else {
-				log.V(2).Info("ungating elastic pod without flavor info; no matching PodSet",
+				log.V(0).Info("ungating elastic pod without flavor info; no matching PodSet",
 					"pod", klog.KObj(pod))
 			}
 			return true, nil
--- a/pkg/controller/elasticjobs/elastic_job_ungater.go
+++ b/pkg/controller/elasticjobs/elastic_job_ungater.go
@@ -40,6 +40,8 @@ import (
 	"sigs.k8s.io/kueue/pkg/constants"
 	"sigs.k8s.io/kueue/pkg/controller/core"
 	"sigs.k8s.io/kueue/pkg/controller/core/indexer"
+	"sigs.k8s.io/kueue/pkg/controller/jobframework"
+	"sigs.k8s.io/kueue/pkg/podset"
 	utilclient "sigs.k8s.io/kueue/pkg/util/client"
 	"sigs.k8s.io/kueue/pkg/util/expectations"
 	"sigs.k8s.io/kueue/pkg/util/parallelize"
@@ -123,6 +125,31 @@ func (r *elasticJobUngater) Reconcile(ctx context.Context, req reconcile.Request
 		return reconcile.Result{}, nil
 	}
 
+	// Elastic jobs bypass startJob, so inject the same PodSetInfo here before
+	// ungating.
+	infos, err := jobframework.GetPodSetsInfoFromStatus(ctx, r.client, wl)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("building PodSet info for elastic ungater: %w", err)
+	}
+	infoByPodSet := make(map[kueue.PodSetReference]podset.PodSetInfo, len(infos))
+	for i := range infos {
+		infoByPodSet[infos[i].Name] = infos[i]
+	}
+	var singleInfo *podset.PodSetInfo
+	if len(infos) == 1 {
+		singleInfo = &infos[0]
+	}
+	infoFor := func(pod *corev1.Pod) (podset.PodSetInfo, bool) {
+		if name, found := pod.Labels[constants.PodSetLabel]; found {
+			info, ok := infoByPodSet[kueue.PodSetReference(name)]
+			return info, ok
+		}
+		if singleInfo != nil {
+			return *singleInfo, true
+		}
+		return podset.PodSetInfo{}, false
+	}
+
 	log.V(2).Info("identified elastic pods to ungate", "count", len(pods))
 	uids := make([]types.UID, len(pods))
 	for i := range pods {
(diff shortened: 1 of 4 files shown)


pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -135,10 +162,21 @@ func (r *elasticJobUngater) Reconcile(ctx context.Context, req reconcile.Request
...
+						"pod", klog.KObj(pod), "error", mergeErr)
+				} else {
+					log.V(3).Info("ungating elastic pod with assigned flavor info", "pod", klog.KObj(pod))
+				}
+			} else {
+				log.V(2).Info("ungating elastic pod without flavor info; no matching PodSet",
+					"pod", klog.KObj(pod))
 			}
-			return ungated, nil
+			return true, nil
```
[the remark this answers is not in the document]
ikchifo: `Merge` failures here are deterministic conflicts (e.g. `ErrDuplicateEntry`, nodeSelector key overlap), not transient errors. Returning the error would retry forever and block workloads that previously ran, and I think that's a worse regression than the bug being fixed.

The common path (where a _merge_ succeeds) fixes the issue, while the rare path (where a _merge_ fails) preserves previous behavior with a `V(2)` log for diagnosis.

mimowo:
The code itself looks ok, but what makes me wonder is this:

> Elastic jobs that bypass startJob (RayCluster with autoscaling, batch/v1
Job with the elastic annotation) relied on the ElasticJobUngater only to
remove the ElasticJobSchedulingGate.

Would it be feasible to **not bypass calling startJob** (just making sure we inject the "elastic-job" scheduling gate into the owner's Pod template)? This is the strategy we take for TAS with the "topology" gates.

If you think this is feasible, but too much work, then I'm ok to ticket this for later, but I want to check, because tmaybe using startJob is not that far off, and would make the injection patterns more coherent.

cc @sohankunkerkar @ichekrygin

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -135,10 +162,21 @@ func (r *elasticJobUngater) Reconcile(ctx context.Context, req reconcile.Request
...
 		e := utilclient.Patch(ctx, r.client, pod, func() (bool, error) {
 			ungated = utilpod.Ungate(pod, kueue.ElasticJobSchedulingGate)
-			if ungated {
-				log.V(3).Info("ungating elastic pod", "pod", klog.KObj(pod))
+			if !ungated {
+				return false, nil
+			}
+			if info, ok := infoFor(pod); ok {
+				if mergeErr := podset.Merge(&pod.ObjectMeta, &pod.Spec, info); mergeErr != nil {
+					log.V(2).Info("failed merging PodSet info before ungating; ungating without flavor injection",
```
mwielgus: When the merge error can happen? Should it warrant a warning instead of an Info?

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -135,10 +162,21 @@ func (r *elasticJobUngater) Reconcile(ctx context.Context, req reconcile.Request
...
+						"pod", klog.KObj(pod), "error", mergeErr)
+				} else {
+					log.V(3).Info("ungating elastic pod with assigned flavor info", "pod", klog.KObj(pod))
+				}
+			} else {
+				log.V(2).Info("ungating elastic pod without flavor info; no matching PodSet",
```
mwielgus: What does it actually mean? When can it happen and should we really ungate in that case?

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -135,10 +162,21 @@ func (r *elasticJobUngater) Reconcile(ctx context.Context, req reconcile.Request
...
-				log.V(3).Info("ungating elastic pod", "pod", klog.KObj(pod))
+			if !ungated {
+				return false, nil
+			}
+			if info, ok := infoFor(pod); ok {
+				if mergeErr := podset.Merge(&pod.ObjectMeta, &pod.Spec, info); mergeErr != nil {
+					log.V(2).Info("failed merging PodSet info before ungating; ungating without flavor injection",
+						"pod", klog.KObj(pod), "error", mergeErr)
+				} else {
+					log.V(3).Info("ungating elastic pod with assigned flavor info", "pod", klog.KObj(pod))
```
mwielgus: Maybe add flavor name for completeness.

ikchifo:
@mimowo Routing elastic jobs through `startJob` does seem like the cleaner long-term design. However, elastic jobs bypass `startJob` [by design](https://github.com/kubernetes-sigs/kueue/blob/38c962f6af181667c8a0440733ed0f83d98448f7/pkg/controller/jobframework/reconciler.go#L608-L613), and reconnecting them is a non-trivial change that would touch admission, the workload controller, and the elastic reconciler (would likely needs its own design review. 

I'm thinking merge this PR as a targeted fix, then open a follow-up issue to evaluate the startJob approach.

pkg/controller/elasticjobs/elastic_job_ungater.go
ikchifo: Agreed on the log level. I'll bump it to a `V(0)` and an error since the pod is being released without expected flavor info and users should see it.

pkg/controller/elasticjobs/elastic_job_ungater.go
ikchifo: "no matching PodSet" means the pod lacks `kueue.x-k8s.io/podset-label` and the workload has multiple PodSets e.g. a RayCluster head pods (since only worker groups are PodSets), ray-autoscaler-created pods that bypass the owner PodTemplate, or even stale pods from a prior slice.

Should we ungate? yes. The workload is already admitted, so blocking the pod forever would be worse than missing flavor info. Plus, the ungater's job is to release pods after admission, not enforce injection.

mimowo:
> @mimowo Routing elastic jobs through startJob does seem like the cleaner long-term design. However, elastic jobs bypass startJob [by design](https://github.com/kubernetes-sigs/kueue/blob/38c962f6af181667c8a0440733ed0f83d98448f7/pkg/controller/jobframework/reconciler.go#L608-L613), and reconnecting them is a non-trivial change that would touch admission, the workload controller, and the elastic reconciler (would likely needs its own design review.

I wouldn't say so, rather an omission. At least I don't remember this being discussed, and it was one of the things which slipped my attention as the PR was large. Also, it might be that originally we only supported Jobs which made it easy to implement Start which ungates the pods directly. As we later started to support KubeRay we reworked that to ungate the pods from a dedicated controller. This now allows for easier decoupling.

 let me cc  @ichekrygin as the original design driver.

> I'm thinking merge this PR as a targeted fix, then open a follow-up issue to evaluate the startJob approach.

Yeah, this is an option. I'm just wondering if then we may have other subtle bugs which we could prevent by the cleaner approach. It also seems less code when done properly. Could you explore this option on an alternative PR?

pkg/controller/elasticjobs/elastic_job_ungater.go
ikchifo: Now looking up the flavor from `wl.Status.Admission.PodSetAssignments[i].Flavors`. Added in b60126fceb0745d94348619e6453e05a253ee67c

ikchifo:
> > I'm thinking merge this PR as a targeted fix, then open a follow-up issue to evaluate the startJob approach.
> 
> Yeah, this is an option. I'm just wondering if then we may have other subtle bugs which we could prevent by the cleaner approach. It also seems less code when done properly. Could you explore this option on an alternative PR?

I'll open a PR to explore this.

sohankunkerkar:
I think the fix is right in its current form. Later workload slices get admitted while the job is already running and the reconciler does not re-enter startJob() for those, so the ungater is the right place for slice-specific flavor injection. Extending the ungater's existing per-pod path (gate removal) to also inject flavor info is reasonable.

On the startJob() unification idea, I think it would not eliminate the ungater injection path by itself, since startJob() only runs once at initial admission and later slices still need per-pod handling.

mimowo:
hmm, but if we start startJob once this is enough to inject the RF specific stuff, because workload slices never migrate between flavors. If we injected the RF node lables into pod template we dont need to do anctions on scale ups or scale downs. Let me know if Im missing something. Also this way we would surface the selected flavor to the user also at the Job level, not just Pods.

sohankunkerkar:
> hmm, but if we start startJob once this is enough to inject the RF specific stuff, because workload slices never migrate between flavors. If we injected the RF node lables into pod template we dont need to do anctions on scale ups or scale downs. Let me know if Im missing something. Also this way we would surface the selected flavor to the user also at the Job level, not just Pods.

Yeah, good point, slices are locked to the same flavor (enforced in the assigner), so startJob injection once would cover all scale-ups. Makes sense to explore that path.

mimowo:
/close
This is no longer needed" https://github.com/kubernetes-sigs/kueue/issues/12350#issuecomment-4876929582

PR #12405, opened 2026-06-22
[ElasticWorkloadSlicing] Account for the autoscaler sidecar in RayCluster head PodSet quota

kevin85421:
#### What type of PR is this?

/kind bug
/area integrations

When a Kueue-managed `RayCluster` enables in-tree autoscaling (`enableInTreeAutoscaling: true`, only allowed for elastic jobs via the `ElasticJobsViaWorkloadSlices` feature gate), KubeRay injects an **[autoscaler sidecar container](https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/configuring-autoscaling.html#autoscaler-sidecar-container)** into the head Pod. This container is added at Pod-build time and is **not** part of `HeadGroupSpec.Template`, so Kueue never saw it when building the head `PodSet`.

As a result the head `PodSet` was under-counted against quota by the autoscaler's resources — KubeRay's [default `500m` CPU / `512Mi` memory (request and limit)](https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/configuring-autoscaling.html#autoscaler-sidecar-container), or `spec.autoscalerOptions.resources` when set. A head Pod that actually needs `head + autoscaler` could be admitted against quota sized only for `head`, letting a ClusterQueue overcommit.

This PR mirrors KubeRay's behavior in `BuildPodSets`: when in-tree autoscaling is enabled, an `autoscaler` container is appended to the head `PodSet` using `AutoscalerOptions.Resources` when provided, otherwise KubeRay's documented defaults (matching `ray-operator/controllers/ray/common/pod.go: BuildAutoscalerContainer`). The change lives in the shared `BuildPodSets`, so both the `RayCluster` and `RayJob` integrations (both call `BuildPodSets`) get the fix from one place.

#### Manual test

* Gist: [autoscaling-raycluster.yaml](https://gist.github.com/kevin85421/5ba40fc3adbf25e9fbb3ddb6f96b668e#file-autoscaling-raycluster-yaml)

* Step 1: Create the RayCluster CR:
  * 1 head Pod (ray container: 1 CPU, autoscaler container: 500m CPU + 512 MB memory) + 1 worker Pod (400m CPU) = 1900m CPU + 512 MB memory
    ```sh
    kubectl apply -f autoscaling-raycluster.yaml
    ```
* Step 2: Check the Workload CR
  image

There is no existing tracked issue for this specific gap.

AI usage disclosure: This change was developed with the assistance of an AI coding tool (Claude Code). All code and tests were reviewed by the author before submission.

```release-note
RayCluster: Fixed a bug where the Ray autoscaler sidecar container's resources were not counted against quota when in-tree autoscaling was enabled, causing the head PodSet to be under-counted. The head PodSet now includes the autoscaler sidecar (KubeRay's default `500m` CPU / `512Mi` memory, or `spec.autoscalerOptions.resources` when set).

ACTION REQUIRED: users with autoscaling-enabled RayClusters may need to increase their ClusterQueue CPU quota by 500m and memory quota by 512Mi per head pod to avoid admission failures after upgrading.
```

## Release Notes

* **New Features**
  * Updated in-tree autoscaling quota calculations for RayCluster head pods to include the injected autoscaler sidecar’s CPU and memory (using default values or configured overrides).

* **Tests**
  * Expanded unit tests to verify autoscaler sidecar injection and resource override behavior in head PodSets.
  * Updated end-to-end and integration tests to include memory in quota/ClusterQueue setup and to confirm autoscaler CPU/memory consistency via drift checks.

### DIFF
--- a/pkg/controller/jobs/rayservice/rayservice_controller_test.go
+++ b/pkg/controller/jobs/rayservice/rayservice_controller_test.go
@@ -36,6 +36,27 @@ import (
 	"sigs.k8s.io/kueue/pkg/workloadslicing"
 )
 
+// headSpecWithAutoscaler returns the head PodSpec with the autoscaler sidecar
+// container appended, matching what BuildPodSets adds when in-tree autoscaling
+// is enabled (KubeRay's default 500m CPU / 512Mi memory).
+func headSpecWithAutoscaler(rayService *RayService) corev1.PodSpec {
+	spec := rayService.Spec.RayClusterSpec.HeadGroupSpec.Template.Spec.DeepCopy()
+	spec.Containers = append(spec.Containers, corev1.Container{
+		Name: "autoscaler",
+		Resources: corev1.ResourceRequirements{
+			Requests: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("500m"),
+				corev1.ResourceMemory: resource.MustParse("512Mi"),
+			},
+			Limits: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("500m"),
+				corev1.ResourceMemory: resource.MustParse("512Mi"),
+			},
+		},
+	})
+	return *spec
+}
+
 func TestPodSets(t *testing.T) {
 	testCases := map[string]struct {
 		rayService   *RayService
@@ -294,7 +315,7 @@ func TestPodSets(t *testing.T) {
 			wantPodSets: func(rayService *RayService) []kueue.PodSet {
 				return []kueue.PodSet{
 					*utiltestingapi.MakePodSet(headGroupPodSetName, 1).
-						PodSpec(*rayService.Spec.RayClusterSpec.HeadGroupSpec.Template.Spec.DeepCopy()).
+						PodSpec(headSpecWithAutoscaler(rayService)).
 						Obj(),
 					*utiltestingapi.MakePodSet("group1", 5). // Updated from RayCluster
 											PodSpec(*rayService.Spec.RayClusterSpec.WorkerGroupSpecs[0].Template.Spec.DeepCopy()).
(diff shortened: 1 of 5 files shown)


mimowo (commented):
/ok-to-test
Thank you 👍

kevin85421:
@mimowo thank you for the fast reply! I am still polishing this PR and I will let you know when it is ready for review.

kevin85421:
Hi @mimowo, this PR should be ready for review. Thanks.

mimowo:
cc @yaroslava-serdiuk ptal

pkg/controller/jobs/rayservice/rayservice_controller_test.go
```diff
@@ -36,6 +36,27 @@ import (
 	"sigs.k8s.io/kueue/pkg/workloadslicing"
 )
 
+// headSpecWithAutoscaler returns the head PodSpec with the autoscaler sidecar
+// container appended, matching what BuildPodSets adds when in-tree autoscaling
+// is enabled (KubeRay's default 500m CPU / 512Mi memory).
+func headSpecWithAutoscaler(rayService *RayService) corev1.PodSpec {
```
sohankunkerkar, line 42: The 500m/512Mi defaults are hardcoded here  and separately in `defaultAutoscalerResources()`.  Consider exporting `AutoscalerContainer` or extracting a shared test helper in test/util/ so the values live in one place.

sohankunkerkar (approved):
I think the changes are LGTM. I would update the release note to something like
```sh
RayCluster: Fixed a bug where the Ray autoscaler sidecar container's resources were not counted against quota when in-tree autoscaling was enabled, causing the head PodSet to be under-counted. The head PodSet now includes the autoscaler sidecar (KubeRay's default `500m` CPU / `512Mi` memory, or `spec.autoscalerOptions.resources` when set). Action required: users with autoscaling-enabled RayClusters may need to increase their ClusterQueue CPU quota by 500m and memory quota by 512Mi per head pod to avoid admission failures after upgrading.
```

I would leave `/lgtm` to @yaroslava-serdiuk

pkg/controller/jobs/rayservice/rayservice_controller_test.go
kevin85421: I intentionally made this helper function private because it's only used in this file. My personal habit is to make a function public only when it's used in multiple places. If Kueue has a different convention, feel free to let me know and I can move it to a util file instead.

kevin85421 (commented):
> https://github.com/kubernetes-sigs/kueue/pull/12405#pullrequestreview-4594129613

Done. I have already updated the release note. Thanks @sohankunkerkar for the review.

yaroslava-serdiuk:
/lgtm
Thanks!

PR #12410, opened 2026-06-22
TAS: Skip flavors with no schedulable topology domains during flavor assignment

GonzaloLuminary:
#### What type of PR is this?

/kind feature 

We have a CQ with a lot of TAS RFs linked to nodes that that may be in unschedulable state. Depending on the RF order in the CQ and the nodes unschedulable state we see significant requeues and scheduling latency since the scheduler tries every flavor in order. The scheduler should not try to use flavors that do not have domains. The check is cheap so I don't think these changes will add performance regressions

Fixes #

```release-note

```

## Release Notes

```
TAS: the flavor assigner now skips TAS flavors without domains to improve scheduling performance.
```

* **Bug Fixes**
  * Improved Topology-Aware Scheduling flavor validation: the scheduler now more reliably rejects TAS flavors that have no schedulable topology domains, including cases where a TAS flavor is implied but placement cannot proceed.
* **Tests**
  * Added unit coverage for detecting when TAS flavor snapshots contain available topology domains.
  * Expanded flavor assignment tests to simulate node availability affecting schedulable TAS placement outcomes.

### DIFF
--- a/pkg/scheduler/flavorassigner/flavorassigner_test.go
+++ b/pkg/scheduler/flavorassigner/flavorassigner_test.go
@@ -40,6 +40,7 @@ import (
 	"sigs.k8s.io/kueue/pkg/util/tas"
 	utiltesting "sigs.k8s.io/kueue/pkg/util/testing"
 	utiltestingapi "sigs.k8s.io/kueue/pkg/util/testing/v1beta2"
+	testingnode "sigs.k8s.io/kueue/pkg/util/testingjobs/node"
 	"sigs.k8s.io/kueue/pkg/workload"
 )
 
@@ -218,6 +219,7 @@ func TestAssignFlavors(t *testing.T) {
 		preemptWorkloadSlice       *workload.Info
 		featureGates               map[featuregate.Feature]bool
 		topologies                 []*kueue.Topology
+		nodes                      []corev1.Node
 	}{
 		"single flavor, fits": {
 			wlPods: []kueue.PodSet{
@@ -3325,6 +3327,10 @@ func TestAssignFlavors(t *testing.T) {
 				utiltestingapi.MakeTopology("tas-topo-a").Levels(corev1.LabelHostname).Obj(),
 				utiltestingapi.MakeTopology("tas-topo-b").Levels(corev1.LabelHostname).Obj(),
 			},
+			nodes: []corev1.Node{
+				*testingnode.MakeNode("node-a").Label(corev1.LabelHostname, "node-a").Ready().Obj(),
+				*testingnode.MakeNode("node-b").Label(corev1.LabelHostname, "node-b").Ready().Obj(),
+			},
 			wlPods: []kueue.PodSet{
 				*utiltestingapi.MakePodSet(kueue.DefaultPodSetName, 1).
 					Request(corev1.ResourceCPU, "1").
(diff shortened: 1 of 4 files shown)


mimowo:
@GonzaloLuminary could you elaborate what is the user-observable benefit - is it performance or less confusing messages, or something else? Understanding the UX change will help a lot.

/ok-to-test

GonzaloLuminary:
>could you elaborate what is the user-observable benefit - is it performance or less confusing messages, or something else? Understanding the UX change will help a lot.

This helps to fix scheduling performance issues and log churn that happens in one of our CQs that has 100s of TAS RFs. Many of those TAS sometimes match unschedulable nodes which don't have domains. In that case, kueue loops over all RFs sequentially and fails scheduling because there are no domains. The latter results in unnecessary requeues.

I think we could further improve these checks by including a cheap comparison of the PodSet total amount of resources vs the resources available in the flavor (ignoring topology) to avoid unnecessary requeues. The latter changes are probably a lot more invasive.

mimowo:
Cool. please update the release-note along the lines

pkg/scheduler/flavorassigner/tas_flavorassigner.go
```diff
@@ -209,6 +218,19 @@ func checkPodSetAndFlavorMatchForTAS(cq *schdcache.ClusterQueueSnapshot, ps *kue
 	return nil
 }
 
+// emptyTASFlavorMessage returns a rejection message when the TAS flavor currently has no
+// schedulable topology domains and TAS placement is not deferred to ProvisioningRequest or
+// MultiKueue. Such a flavor would be selected on quota and then fail topology assignment,
+// forcing a requeue.
+func emptyTASFlavorMessage(cq *schdcache.ClusterQueueSnapshot, s *schdcache.TASFlavorSnapshot, flavorRef kueue.ResourceFlavorReference) *string {
```
mwielgus, line 225: The function name suggest that a message will be returned without any additional check. But inside there is quite extensive check.

mwielgus (commented):
If there is no schedulable domain the RF will be quickly rejected - nothing is running there, nothing can be scheduled, so it Kueue won't spend much time there. What is the gain from of this change other than handling this, relatively rare, situation directly?

GonzaloLuminary:
We have CQs with hundreds of RFs and, as we balance these RFs around at the infra level, many of the RFs stop having domains. Given the scheduler backoff, each requeue due to RFs being empty can take more than 10ms (it was worse in the past with 100ms backoff). We lose a lot of scheduling throughout and log churn due to the scheduler back off. If the flavor assigner knows that the RF doesn't have domains then there's no point for the scheduler to process said flavor.

mimowo:
@GonzaloLuminary please add a release note

pkg/scheduler/flavorassigner/flavorassigner_test.go
```diff
@@ -219,6 +220,7 @@ func TestAssignFlavors(t *testing.T) {
 		preemptWorkloadSlice       *workload.Info
 		featureGates               map[featuregate.Feature]bool
 		topologies                 []*kueue.Topology
+		nodes                      []corev1.Node
```
mimowo, line 223: I think it would also be great to add a unit test in scheduler_tas_test.go for the case where the first TAS flavor is empty, and the workload is moved to the next flavor. Additionally, a test case where there is a single flavor or two empty flavors. The reason is tow also check how this bugfix plays with the new granular reasons for workload observability for inadmissible workloads.

mimowo:
@GonzaloLuminary friendly ping

PR #12414, opened 2026-06-22
MultiKueue: gate JobSuspended workaround behind Alpha feature gate

### LINKED ISSUE #11863: Clean up the JobCondition hack for k8s 1.36 once the k8s fix is released

In order to allow the upgrade of k8s dependencies to 1.36 (#10936), due to a regression in k8s validation semantics (https://github.com/kubernetes/kubernetes/issues/139281), we added a hack to add a `JobSuspended = True` condition to Jobs in MultiKueue adapter (#10946, specifically [this code](https://github.com/kubernetes-sigs/kueue/blob/a4b8a8d65721a4cf47d0d12a69d849e863d359e5/pkg/controller/jobs/job/job_multikueue_adapter.go#L140-L155)).

As already state in the comment above the `if` block, this block can be removed once the k8s-side fix (https://github.com/kubernetes/kubernetes/pull/139329) is released. This fix has been already cherry-picked in k8s (https://github.com/kubernetes/kubernetes/pull/139329), and is expected in 1.36.2.

**Why is this needed**:

To avoid maintaining the hacky code.

k-ngs:
#### What type of PR is this?

/kind feature
/area multikueue

Introduces the `MultiKueueJobSuspendedWorkaroundDisabled` Alpha feature gate to control the workaround that manually patches a `JobSuspended=True` condition on the local Job in the MultiKueue adapter (#11649).

The workaround avoids a K8s 1.36 validation regression (kubernetes/kubernetes#139281), where modifications to scheduling directives on a suspended Job were rejected if the `JobSuspended` condition had not yet been set by the job controller. 
The regression was fixed in kubernetes/kubernetes#139287 and cherry-picked to 1.36.2.

Rather than removing the workaround outright, this PR gates it so that:
- Gate disabled (default): the workaround is applied, keeping users on
  Kubernetes 1.36.0/1.36.1 protected.
- Gate enabled: the workaround is skipped, which is safe on Kubernetes 1.36.2+
  where the underlying issue is fixed.

Changes:
- Add the `MultiKueueJobSuspendedWorkaroundDisabled` Alpha feature gate
- Guard the `JobSuspended=True` patching in `determineStatusUpdate` behind the gate
- Add tests covering both gate states and the already-patched case

Fixes #11863

```release-note
MultiKueue: Added the `MultiKueueJobSuspendedWorkaroundDisabled` Alpha feature gate (disabled by default). 
When disabled, Kueue keeps patching `JobSuspended=True` on the local Job to remain compatible with Kubernetes 1.36.0/1.36.1 (kubernetes/kubernetes#139281).
Enable it on Kubernetes 1.36.2+, where the underlying issue is fixed, to skip the workaround.
```

Note: I investigated and implemented this change with the assistance of an AI tool.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved job synchronization when a locally suspended job becomes unsuspended remotely but remains unfinished.
  * Applied the `JobSuspended=True` compatibility behavior only when needed for Kubernetes 1.36.0 and 1.36.1.

* **New Features**
  * Added the `MultiKueueJobSuspendedWorkaroundDisabled` feature gate to control this compatibility behavior.

* **Tests**
  * Expanded coverage for enabled, disabled, and already-applied workaround scenarios.

### DIFF
--- a/pkg/controller/jobs/job/job_multikueue_adapter.go
+++ b/pkg/controller/jobs/job/job_multikueue_adapter.go
@@ -23,7 +23,6 @@ import (
 
 	"github.com/go-logr/logr"
 	batchv1 "k8s.io/api/batch/v1"
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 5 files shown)


tenzen-y:
I think that we should keep this workaround while we support v1.36. Because we should not enforce users to update their K8s patch version. It will be a blocker for updating Kueue version.

/hold

mimowo:
> I think that we should keep this workaround while we support v1.36. Because we should not enforce users to update their K8s patch version. It will be a blocker for updating Kueue version.

Yes, I think we either keep the workaround until 1.36 is EoL, or we introduce a FG to decomission the code in steps. For example, we could already introduce an Alpha FG in this PR which skips the workaround. Now, users on 1.36.2+ can enable the FG.

mimowo:
/ok-to-test
We can test, but I think Alpha FG is needed

tenzen-y:
> /ok-to-test
> 
> We can test, but I think Alpha FG is needed

The alpha FG (disabled by default) sounds fine as well.

k-ngs:
Sorry...
I didn't consider the users running older patch versions of v1.36.

Is addressing the Cleanup issue by implementing Alpha FG an overengineered solution?

mimowo:
I think this is reasonable, so that users on newer versions will be able to skip the update. Getting 1.36 EoL will be a year from now. In the meanwhile, say in 3 months, most users will be on versions which don't require the workaround.

k-ngs:
OK!

It might take some time, but would it be okay for me to attempt implementing this as an FG?
If that's okay, can I keep this PR open? Or should I convert it to draft?

mimowo:
> It might take some time, but would it be okay for me to attempt implementing this as an FG?

Yes, feel free to work on it.

> If that's okay, can I keep this PR open? Or should I convert it to draft?

I would suggest just updating the PR, don't see a benefit of closing and re-opening another one in this case

k-ngs:
Following your suggestion, I've implemented the code as a feature gate.
I'll add you as reviewers.

/cc @mimowo @tenzen-y

mimowo:
/kind feature
@k-ngs would you like to continue on that one?

tenzen-y:
> Following your suggestion, I've implemented the code as a feature gate. I'll add you as reviewers.
> 
> /cc @mimowo @tenzen-y

Sorry for the delayed response. I found this notification now. @k-ngs can you rebase this PR onto the main branch?

mimowo:
@k-ngs would you like to continue working on this and rebase the branch?

PR #12420, opened 2026-06-22
[WIP] Add placement-level gang scheduling via scheduler-library

sohankunkerkar:
#### What type of PR is this?
/kind feature

 Adds placement-level gang scheduling via scheduler-library. Before admitting a workload, Kueue now runs the scheduler-library to verify all pods across all PodSets can actually be placed on cluster nodes. If any PodSet fails placement, the workload stays pending instead of wasting quota.

Fixes #

Builds on top of the scheduler-library vendor in #12256. The initial implementation was AI-assisted. The architecture decisions were made and reviewed by me. This work is primarily intended to understand the problem space and get a POC out to evaluate feasibility.

```release-note
NONE
```


sohankunkerkar:
cc @alaypatel07 this is the placement-level gang scheduling POC we discussed.

olekzabl:
@sohankunkerkar Would it be feasible that you split this PR into 2 commits, one identical to some commit from #12256 and the other one containing your changes on top of that?
I imagine it could make this PR easier to grasp for everyone.

sohankunkerkar:
> @sohankunkerkar Would it be feasible that you split this PR into 2 commits, one identical to some commit from #12256 and the other one containing your changes on top of that? I imagine it could make this PR easier to grasp for everyone.

Good call! I agree splitting makes it easier to review. The issue is that #12256 is outdated and I couldn't use it as-is, so I had to make changes on top. But fair point to keep the base as a separate commit so the delta is clear. I'll split it out.

PR #12421, opened 2026-06-22
Add e2e regression test for elasticJob ResourceFlavor injection

### LINKED ISSUE #12350: ElasticJobsViaWorkloadSlices: ResourceFlavor nodeLabels not injected on elastic job pods

ikchifo:
#### What type of PR is this?

/kind cleanup
/area testing

Adds an e2e regression test verifying that an elastic Job
(`ElasticJobsViaWorkloadSlices`) receives the admitted ResourceFlavor's
node labels and tolerations on the Job template and its running pods,
both at initial admission and after a scale-up that replaces the
workload slice.

Investigation of #12350 showed flavor injection works correctly in
current Kueue; the reported symptom was caused by an external mutating
webhook in my environment stripping the injected nodeSelector. This
test pins the correct behavior against future regressions.

Fixes #12350

Scoped down from the earlier startJob code change per review
discussion; also supersedes #12403. Test-only.

```release-note
NONE
```

### DIFF
--- a/test/e2e/singlecluster/baseline/job_test.go
+++ b/test/e2e/singlecluster/baseline/job_test.go
@@ -42,6 +42,7 @@ import (
 	"sigs.k8s.io/kueue/pkg/workload"
 	workloadevict "sigs.k8s.io/kueue/pkg/workload/evict"
 	workloadfinish "sigs.k8s.io/kueue/pkg/workload/finish"
+	"sigs.k8s.io/kueue/pkg/workloadslicing"
 	"sigs.k8s.io/kueue/test/util"
 )
 
@@ -841,6 +842,161 @@ var _ = ginkgo.Describe("Kueue", ginkgo.Label("area:singlecluster", "feature:job
 					g.Expect(k8sClient.Get(ctx, jobKey, createdJob)).Should(gomega.Succeed())
 					g.Expect(ptr.Deref(createdJob.Spec.Suspend, false)).Should(gomega.BeTrue())
 				}, util.Timeout, util.Interval).Should(gomega.Succeed())
+			})
+		})
+	})
+
+	ginkgo.When("Creating an elastic Job with a ResourceFlavor that has nodeLabels and tolerations", func() {
+		var (
+			onDemandRF   *kueue.ResourceFlavor
+			localQueue   *kueue.LocalQueue
+			clusterQueue *kueue.ClusterQueue
+		)
+
+		flavorToleration := corev1.Toleration{
+			Key:      "example.com/test-taint",
+			Operator: corev1.TolerationOpExists,
+			Effect:   corev1.TaintEffectNoSchedule,
+		}
+
+		ginkgo.BeforeEach(func() {
+			flavorName := "on-demand-" + ns.Name
+			cqName := "cluster-queue-" + ns.Name
+
+			onDemandRF = utiltestingapi.MakeResourceFlavor(flavorName).
+				NodeLabel("instance-type", "on-demand").
+				Toleration(flavorToleration).
+				Obj()
+			util.MustCreate(ctx, k8sClient, onDemandRF)
+
+			clusterQueue = utiltestingapi.MakeClusterQueue(cqName).
+				ResourceGroup(
+					*utiltestingapi.MakeFlavorQuotas(flavorName).
+						Resource(corev1.ResourceCPU, "4").
+						Resource(corev1.ResourceMemory, "4Gi").
+						Obj(),
+				).
+				Obj()
+			util.CreateClusterQueuesAndWaitForActive(ctx, k8sClient, clusterQueue)
+
+			localQueue = utiltestingapi.MakeLocalQueue("main", ns.Name).ClusterQueue(cqName).Obj()
+			util.CreateLocalQueuesAndWaitForActive(ctx, k8sClient, localQueue)
+		})
+
+		ginkgo.AfterEach(func() {
+			gomega.Expect(util.DeleteAllJobsInNamespace(ctx, k8sClient, ns)).Should(gomega.Succeed())
+			gomega.Expect(util.DeleteWorkloadsInNamespace(ctx, k8sClient, ns)).Should(gomega.Succeed())
+			gomega.Expect(util.DeleteObject(ctx, k8sClient, localQueue)).Should(gomega.Succeed())
+			util.ExpectObjectToBeDeleted(ctx, k8sClient, clusterQueue, true)
+			util.ExpectObjectToBeDeleted(ctx, k8sClient, onDemandRF, true)
+			util.ExpectAllPodsInNamespaceDeleted(ctx, k8sClient, ns)
+		})
+
+		ginkgo.It("should inject flavor nodeLabels and tolerations into the Job and pods at admission and after scale-up", func() {
+			expectedNodeSelector := map[string]string{"instance-type": "on-demand"}
+
+			elasticJob := testingjob.MakeJob("elastic-flavor-injection", ns.Name).
+				SetAnnotation(workloadslicing.EnabledAnnotationKey, workloadslicing.EnabledAnnotationValue).
+				Queue(kueue.LocalQueueName(localQueue.Name)).
+				RequestAndLimit(corev1.ResourceCPU, "200m").
+				RequestAndLimit(corev1.ResourceMemory, "20Mi").
+				Parallelism(2).
+				Completions(10).
+				Image(util.GetAgnHostImage(), util.BehaviorWaitForDeletion).
+				TerminationGracePeriod(1).
+				Obj()
+			util.MustCreate(ctx, k8sClient, elasticJob)
+
+			ginkgo.By("waiting for the workload to be admitted", func() {
+				workloads := util.ExpectWorkloadsInNamespace(ctx, k8sClient, ns.Name, 1)
+				gomega.Eventually(func(g gomega.Gomega) {
+					g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(&workloads[0]), &workloads[0])).Should(gomega.Succeed())
+					g.Expect(workload.IsAdmitted(&workloads[0])).Should(gomega.BeTrue())
+				}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
+			})
+
+			ginkgo.By("verifying the Job template carries the flavor nodeSelector and toleration", func() {
+				util.ExpectJobUnsuspendedWithNodeSelectors(ctx, k8sClient, client.ObjectKeyFromObject(elasticJob), expectedNodeSelector)
+
+				gomega.Eventually(func(g gomega.Gomega) {
+					g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(elasticJob), elasticJob)).Should(gomega.Succeed())
+					g.Expect(elasticJob.Spec.Template.Spec.Tolerations).Should(gomega.ContainElement(flavorToleration))
+				}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
+			})
+
+			ginkgo.By("verifying running pods carry the flavor nodeSelector and toleration", func() {
+				gomega.Eventually(func(g gomega.Gomega) {
+					pods := &corev1.PodList{}
+					g.Expect(k8sClient.List(ctx, pods,
+						client.InNamespace(ns.Name),
+						client.MatchingLabels{"batch.kubernetes.io/job-name": elasticJob.Name},
+					)).Should(gomega.Succeed())
+
+					runningCount := 0
+					for i := range pods.Items {
+						if pods.Items[i].Status.Phase == corev1.PodRunning {
+							runningCount++
+							g.Expect(pods.Items[i].Spec.NodeSelector).Should(gomega.Equal(expectedNodeSelector))
+							g.Expect(pods.Items[i].Spec.Tolerations).Should(gomega.ContainElement(flavorToleration))
+						}
+					}
+					g.Expect(runningCount).Should(gomega.BeNumerically(">=", 2))
+				}, util.LongTimeout, util.Interval).Should(gomega.Succeed())
+			})
+
+			var oldWorkload *kueue.Workload
+			ginkgo.By("recording the current workload before scaling", func() {
+				workloads := util.ExpectWorkloadsInNamespace(ctx, k8sClient, ns.Name, 1)
+				oldWorkload = &workloads[0]
+			})
+
+			ginkgo.By("scaling up parallelism from 2 to 3", func() {
+				gomega.Eventually(func(g gomega.Gomega) {
+					g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(elasticJob), elasticJob)).Should(gomega.Succeed())
+					elasticJob.Spec.Parallelism = ptr.To[int32](3)
+					g.Expect(k8sClient.Update(ctx, elasticJob)).Should(gomega.Succeed())
+				}, util.Timeout, util.Interval).Should(gomega.Succeed())
+			})
+
+			ginkgo.By("waiting for the replacement workload slice to be admitted and old slice finished", func() {
+				newWorkload := util.ExpectNewWorkloadSlice(ctx, k8sClient, oldWorkload)
+				gomega.Eventually(func(g gomega.Gomega) {
+					g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(newWorkload), newWorkload)).Should(gomega.Succeed())
+					g.Expect(workload.IsAdmitted(newWorkload)).Should(gomega.BeTrue())
+				}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
+
+				gomega.Eventually(func(g gomega.Gomega) {
+					g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(oldWorkload), oldWorkload)).Should(gomega.Succeed())
+					g.Expect(workload.IsFinished(oldWorkload)).Should(gomega.BeTrue())
+				}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
+			})
+
+			ginkgo.By("verifying the Job template still carries the flavor nodeSelector and toleration after scale-up", func() {
+				gomega.Eventually(func(g gomega.Gomega) {
+					g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(elasticJob), elasticJob)).Should(gomega.Succeed())
+					g.Expect(elasticJob.Spec.Template.Spec.NodeSelector).Should(gomega.Equal(expectedNodeSelector))
+					g.Expect(elasticJob.Spec.Template.Spec.Tolerations).Should(gomega.ContainElement(flavorToleration))
+				}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
+			})
+
+			ginkgo.By("verifying all 3 pods carry the flavor nodeSelector and toleration after scale-up", func() {
+				gomega.Eventually(func(g gomega.Gomega) {
+					pods := &corev1.PodList{}
+					g.Expect(k8sClient.List(ctx, pods,
+						client.InNamespace(ns.Name),
+						client.MatchingLabels{"batch.kubernetes.io/job-name": elasticJob.Name},
+					)).Should(gomega.Succeed())
+
+					runningCount := 0
+					for i := range pods.Items {
+						if pods.Items[i].Status.Phase == corev1.PodRunning {
+							runningCount++
+							g.Expect(pods.Items[i].Spec.NodeSelector).Should(gomega.Equal(expectedNodeSelector))
+							g.Expect(pods.Items[i].Spec.Tolerations).Should(gomega.ContainElement(flavorToleration))
+						}
+					}
+					g.Expect(runningCount).Should(gomega.BeNumerically(">=", 3))
+				}, util.LongTimeout, util.Interval).Should(gomega.Succeed())
 			})
 		})
 	})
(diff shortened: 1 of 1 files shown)


ikchifo:
This PR keeps per-pod injection for pods that already exist at admission, while also routing admitted elastic jobs through `startJob`. This updates the owner PodTemplate so autoscaler-created pods inherit the ResourceFlavor placement. Replacement slices restore canonical Workload PodSets before applying their new admission.

Locally validated with some targeted Job and RayCluster integration tests.
If the design here looks right, I'll close the other PR  in favor of this.
cc: @mimowo @sohankunkerkar

pkg/controller/jobframework/reconciler.go
```diff
@@ -619,12 +620,85 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
...
+	if runningPodSets != nil && podSetInfoFieldsEqual(jobPodSets, runningPodSets) {
+		return nil
+	}
+
+	log.V(2).Info("Injecting PodSet info into elastic job template after admission")
+	// Workload PodSets are canonical for PodSetInfo-managed fields. Restoring
+	// them intentionally replaces out-of-band changes to those fields.
+	restoreInfo := GetPodSetsInfoFromWorkload(wl)
+	for i := range restoreInfo {
+		restoreInfo[i].SchedulingGates = withoutElasticJobSchedulingGate(restoreInfo[i].SchedulingGates)
```
mimowo: I was rather thinking about injecting the scheduling gate inside start job, probably next to setting the PodSetLavel or injecting the Topology scheduling gate. Any reason to do this outside of startJob?

ikchifo: Ahh, I put the logic outside because already-running elastic jobs bypass the 'suspended' code branch, and I wanted to avoid patching and emitting start events on every reconcile.

I’ll just move that mutation into `startJob` then.

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -135,10 +167,22 @@ func (r *elasticJobUngater) Reconcile(ctx context.Context, req reconcile.Request
...
+				} else {
+					log.V(3).Info("ungating elastic pod with assigned flavor info",
+						"pod", klog.KObj(pod), "flavors", flavorsByPodSet[info.Name])
+				}
+			} else {
+				log.V(0).Info("ungating elastic pod without flavor info; no matching PodSet",
+					"pod", klog.KObj(pod))
 			}
-			return ungated, nil
+			return true, nil
```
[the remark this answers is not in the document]
ikchifo: This is intentional. Merge conflicts are deterministic, so retrying would leave an admitted pod gated indefinitely. We log the error and ungate to avoid that regression; future pods inherit the corrected owner template.

pkg/controller/jobframework/reconciler.go
```diff
@@ -619,12 +620,79 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
...
+	// makes a still-gated owner template differ and trigger injection.
+	for i := range runningPodSets {
+		runningPodSets[i].Template.Spec.SchedulingGates = withoutElasticJobSchedulingGate(runningPodSets[i].Template.Spec.SchedulingGates)
+	}
+	if runningPodSets != nil && podSetInfoFieldsEqual(jobPodSets, runningPodSets) {
+		return nil
+	}
+
+	log.V(2).Info("Injecting PodSet info into elastic job template after admission")
+	if err := r.startJob(ctx, job, object, wl); err != nil {
```
mimowo: Maybe I'm missing something, but I'm still wondering why we need a separete call to startJob, and cannot reuse the existing path which already takes case of the flavor injection. 

I think in this case maybe a short call over slack could help to align (and walk together trough the code), feel free to reach out over slack.

ikchifo:
Also, visit [pull-kueue-test-integration-shard-2-main/2070659213809422336/build-log.txt](https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12421/pull-kueue-test-integration-shard-2-main/2070659213809422336/build-log.txt) and search for this error:
```
field is immutable
```
That's what 53ae383 addresses; when a Batch Job scale-up repeatedly attempts to mutate immutable spec.template

pkg/controller/jobframework/reconciler.go
ikchifo: Thanks for the call. 

Two incorrect assumptions were clarified: 
- ResourceFlavor does not change between slices
- Suspended batch Jobs cannot create Pods before `startJob` injects `PodSetInfo`. 

We agreed to reuse `startJob` for injection and avoid reinjecting on replacement slices. I’ve updated the PR accordingly.

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -135,10 +167,22 @@ func (r *elasticJobUngater) Reconcile(ctx context.Context, req reconcile.Request
 		var ungated bool
 		e := utilclient.Patch(ctx, r.client, pod, func() (bool, error) {
 			ungated = utilpod.Ungate(pod, kueue.ElasticJobSchedulingGate)
-			if ungated {
-				log.V(3).Info("ungating elastic pod", "pod", klog.KObj(pod))
+			if !ungated {
+				return false, nil
+			}
+			if info, ok := infoFor(pod); ok {
+				if mergeErr := podset.Merge(&pod.ObjectMeta, &pod.Spec, info); mergeErr != nil {
```
sohankunkerkar: I think it should return (false, mergeErr) on merge failure so the patch is not applied, or move Merge before Ungate.

pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -135,10 +167,22 @@ func (r *elasticJobUngater) Reconcile(ctx context.Context, req reconcile.Request
...
+					log.Error(mergeErr, "failed merging PodSet info after ungating; pod released without flavor injection",
+						"pod", klog.KObj(pod))
+				} else {
+					log.V(3).Info("ungating elastic pod with assigned flavor info",
+						"pod", klog.KObj(pod), "flavors", flavorsByPodSet[info.Name])
+				}
+			} else {
+				log.V(0).Info("ungating elastic pod without flavor info; no matching PodSet",
```
sohankunkerkar: Consider returning (false, nil) to keep the pod gated rather than releasing it without injection.

pkg/controller/jobframework/reconciler.go
sohankunkerkar, line 1343: `startJob` was designed for the suspended→unsuspended transition, so this "Started"/"Admitted by clusterQueue" event makes sense there. But now `ensureElasticPodSetInfoInjected` reuses `startJob` for template-patching on already-running elastic jobs. Nothing is actually "started". Users will see a spurious Started event in `kubectl describe`. Consider a different event reason when the job is already running, or suppress the event in that case.

pkg/controller/jobframework/reconciler.go
```diff
@@ -1478,6 +1547,14 @@ func prepareWorkloadSlice(ctx context.Context, clnt client.Client, job GenericJo
 	case 1:
 		// Scale-up event - link to old slice and carry origin name.
 		oldSlice := workloadSlices[0]
+		desiredCounts := workload.ExtractPodSetCountsFromWorkload(wl)
```
sohankunkerkar: This changes workload slice replacement behavior (copying old slice's user-authored PodSets to avoid runtime-injected data leaking into the new workload spec). It looks separate from the ungater/startJob fix? if yes, then consider splitting for it into a separate PR.

pkg/controller/elasticjobs/elastic_job_ungater.go
ikchifo: That does make sense, I'll move `Merge` before `Ungate`. So the patch gets aborted so the pod stays gated

pkg/controller/elasticjobs/elastic_job_ungater.go
ikchifo: I was thinking since there's no matching `PodSetInfo` at all (unknown `PodSetLabel` with a >1 `PodSet`, so no single-PodSet fallback), it is an internal inconsistency rather than a user template conflict. 

Gating indefinitely would block an already-admitted pod with no actionable signal. 
also Happy to make it block, just like the template conflict scenario above.

EDIT: Fixed in [0da2b95](https://github.com/kubernetes-sigs/kueue/pull/12421/commits/0da2b95416b4f9afbed25d8777c478b4dbd88c5b)

pkg/controller/jobframework/reconciler.go
ikchifo: will update so no confusing Started/Admitted event is emitted

EDIT: Done here: [c8fb5c6](https://github.com/kubernetes-sigs/kueue/pull/12421/commits/c8fb5c60c661520c1ea5c14d364796214d5bc59b)

pkg/controller/jobframework/reconciler.go
ikchifo: This `prepareWorkloadSlice` block only exists to counter the template-patching in this PR: once `startJob` stamps the Job template with the admitted flavor's _nodeSelector/gate_, a scale-up builds the replacement Workload from that template, so the runtime placement leaks into the new Workload's spec (the request) where it doesn't belong. Restoring the old slice's user-authored PodSets keeps the request clean.


That coupling is what makes splitting awkward, if this PR merges first without the fix, every scale-up regresses (stale flavor/gate leaks into the new slice) until the follow-up lands.

pkg/controller/jobframework/reconciler.go
```diff
@@ -626,12 +627,77 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
...
+
 	// workload is admitted and job is running, nothing to do.
-	// For elastic jobs, pod ungating is handled by the ElasticJobUngater controller.
 	log.V(3).Info("Job running with admitted workload, nothing to do")
 	return ctrl.Result{}, nil
 }
 
+// ensureElasticPodSetInfoInjected patches the parent job's PodTemplate with the
+// admitted PodSetInfo when it differs from the current workload assignment.
+func (r *JobReconciler) ensureElasticPodSetInfoInjected(ctx context.Context, job GenericJob, object client.Object, wl *kueue.Workload, log logr.Logger) error {
```
mimowo: Hm, why do we even attempt to patch jobs which are already running?

ikchifo: For elastic jobs, "running" doesn't mean Kueue started it. We deliberately don't re-suspend a running elastic job that isn't admitted yet, and the schedulng gate holds its pods instead of `suspend`. 
So admission can happen while the job is already 'running', and the normal unsuspend path that stamps the template never fires. 
Without patching here, the template would never get the flavor nodeSelectors/tolerations, and every future pod would rely on per-pod injection by the ungater (the #12403 approach).

When the template is already correct (e.g. during a slice scale-up) this is a no-op. we just compare the stable fields first and skip `startJob`.

sohankunkerkar: I agreed! the ungater is the right place for this, not startJob. Elastic jobs are never re-suspended after the first admission, so routing back through startJob is a category mismatch and the "field is immutable" error confirms it's a dead end for batch/v1. Since the scheduling gate stays in the template, every future pod already flows through the ungater, so injecting PodSetInfo there atomically with gate removal (same pattern as the TAS ungater) is sufficient. That drops the reconciler changes entirely.

mimowo: Hm, I didn't mean exactly that - I meant that calling `startJob` once for the first slice should be enough, and no need to re-run it.

Now, let me have one question: does currently Kueue call startJob once for elasticJobs?

The issue description says that: "The standard path injects nodeLabels via startJob -> RunWithPodSetsInfo -> podset.Merge, which calls podset.FromAssignment to collect labels from each assigned flavor. The elastic path bypasses startJob." However, I studied the code and it seems that the "startJob" is called on the admission of the first workload slice.

Now, my confusion is here: if "startJob" is called on admitting the first workload slice (before pod creation), and it injects the nodeLabels and tolerations into the Job's PodTemplate, then all Pods should already have them (as the workload resize is sticky wrt ResourceFlavor), no need for the ElasticJobUngater to inject the NodeLabels and tolerations. 

So, I cannot even see where the bug is TBH. The flow I imagine (as implemented currently, and correct):
1. RayCluster is created, suspend=true is set by webhook, not Pods created
2. RayCluster created and workload is admitted to a resourceFlavor - nodeLabels and tolerations are injected into the RayCluster PodTemplate
3. pods are created with already injected nodeLabels & tolerations
4. resize skips startJob, but that is fine, resize is sticky to the previous ResourceFlavor, so all previously injected (into the Job's PodTemplate) nodeLabels & tolerations are still correct, and will be injected to the newly created pods.

Now, I would like to first understand where I'm making an inaccurate assumption, because if that flow was accurate then we wouldn't have the bug, or maybe the bug only exists on some old versions?

sohankunkerkar: You're right that startJob runs on first admission and the webhook unconditionally suspends, and when the workload is admitted the reconciler hits the `job.IsSuspended() && workload.IsAdmitted(wl)` path, so RunWithPodSetsInfo injects the flavor nodeLabels. I also verified that stopJob on eviction self-heals because it re-suspends the job, so startJob re-fires on re-admission.

I can't reproduce the issue as described in #12350 from code analysis or envtest. Worth asking @ikchifo for `kubectl describe raycluster` output showing the template state pre/post admission on their GKE cluster. There may be an environment-specific interaction we're not seeing.

mimowo: @sohankunkerkar thank you for double checking (both in code and experimentally)! So, I think we need to return to the drawing board and understand if the issue exists on the latest version of Kueue (and KubeRay). 

In the meanwhile I think it would be great to add an e2e tests for elastic Jobs to confirm that the RF labels and tolerations are injected into the pods correctly.

ikchifo: Root cause found, and it was indeed environment-specific. Our platform runs a mutating webhook on RayJobs that stripped `cloud.google.com/gke-nodepool` from worker nodeSelectors on every CREATE/UPDATE, so Kueue's `startJob` injection was correct but could never persist: the next update (e.g. kuberay-operator's finalizer write) passed through our webhook and removed the label again. Compounding it, the cluster runs Kueue v0.16.9, whose `ElasticJobUngater` removes the gate without injecting anything, so there was no recovery path once the template lost the label.

ikchifo: @mimowo @sohankunkerkar I apologize for the noise on this issue and the time spent reviewing the PR

I'll scope #this PR down to just the regression e2e test, "where an elastic batch Job admitted and scaled up, asserting the ResourceFlavor's nodeLabels and tolerations reach the pod template and the pods"

mimowo: Thank you, no problem I know how tricky are env-specific webhook issues. I have chased a couple in my past. Happy it is all good now, and your env will work well.

test/e2e/singlecluster/baseline/job_test.go
```diff
@@ -843,6 +844,161 @@ var _ = ginkgo.Describe("Kueue", ginkgo.Label("area:singlecluster", "feature:job
...
+
+					runningCount := 0
+					for i := range pods.Items {
+						if pods.Items[i].Status.Phase == corev1.PodRunning {
+							runningCount++
+							g.Expect(pods.Items[i].Spec.NodeSelector).Should(gomega.Equal(expectedNodeSelector))
+							g.Expect(pods.Items[i].Spec.Tolerations).Should(gomega.ContainElement(flavorToleration))
+						}
+					}
+					g.Expect(runningCount).Should(gomega.BeNumerically(">=", 2))
```
mimowo: Instead of doing `runningCount` build the set. So that later you can compare its size. When that fails then the assert reported failure contains more details - list of pot names, not just sizes.

test/e2e/singlecluster/baseline/job_test.go
```diff
@@ -843,6 +844,161 @@ var _ = ginkgo.Describe("Kueue", ginkgo.Label("area:singlecluster", "feature:job
...
+			ginkgo.By("verifying the Job template carries the flavor nodeSelector and toleration", func() {
+				util.ExpectJobUnsuspendedWithNodeSelectors(ctx, k8sClient, client.ObjectKeyFromObject(elasticJob), expectedNodeSelector)
+
+				gomega.Eventually(func(g gomega.Gomega) {
+					g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(elasticJob), elasticJob)).Should(gomega.Succeed())
+					g.Expect(elasticJob.Spec.Template.Spec.Tolerations).Should(gomega.ContainElement(flavorToleration))
+				}, util.MediumTimeout, util.Interval).Should(gomega.Succeed())
+			})
+
+			ginkgo.By("verifying running pods carry the flavor nodeSelector and toleration", func() {
```
mimowo: Let's also check the job size in status.Ready.

test/e2e/singlecluster/baseline/job_test.go
mimowo: That should be exactly two

test/e2e/singlecluster/baseline/job_test.go
```diff
@@ -843,6 +844,161 @@ var _ = ginkgo.Describe("Kueue", ginkgo.Label("area:singlecluster", "feature:job
...
+
+					runningCount := 0
+					for i := range pods.Items {
+						if pods.Items[i].Status.Phase == corev1.PodRunning {
+							runningCount++
+							g.Expect(pods.Items[i].Spec.NodeSelector).Should(gomega.Equal(expectedNodeSelector))
+							g.Expect(pods.Items[i].Spec.Tolerations).Should(gomega.ContainElement(flavorToleration))
+						}
+					}
+					g.Expect(runningCount).Should(gomega.BeNumerically(">=", 3))
```
mimowo: extract finding the running pods to a helper function `findRunningPods` or `findPodsByPhase() to avoid repating the code. the helper returns set of NamepsacedName, and you can check if the set size is as expected.

test/e2e/singlecluster/baseline/job_test.go
```diff
@@ -843,6 +844,161 @@ var _ = ginkgo.Describe("Kueue", ginkgo.Label("area:singlecluster", "feature:job
...
+			util.CreateClusterQueuesAndWaitForActive(ctx, k8sClient, clusterQueue)
+
+			localQueue = utiltestingapi.MakeLocalQueue("main", ns.Name).ClusterQueue(cqName).Obj()
+			util.CreateLocalQueuesAndWaitForActive(ctx, k8sClient, localQueue)
+		})
+
+		ginkgo.AfterEach(func() {
+			gomega.Expect(util.DeleteAllJobsInNamespace(ctx, k8sClient, ns)).Should(gomega.Succeed())
+			gomega.Expect(util.DeleteWorkloadsInNamespace(ctx, k8sClient, ns)).Should(gomega.Succeed())
+			gomega.Expect(util.DeleteObject(ctx, k8sClient, localQueue)).Should(gomega.Succeed())
```
mimowo: nit, to remove the namepsaced objects you can just use DeleteNamespace

mimowo:
@ikchifo I see you probably addressed all the comments, but it would be good if you handle the comments either by:
1. simple suggestions mark as resolved with a small comment like "done"
2. for more complex suggestions you may follow them but keep the comment open for others
3. for some suggestions just  answer in case they need more discussion

That helps reviewers with the process

ikchifo:
@mimowo Understood! will make sure to keep doing that going forward.

mimowo (commented):
/lgtm
/approve
Thank you @ikchifo 👍 
Let me cherrypick so that future cherrypicks to elastic Jobs are easier
/cherrypick release-0.18
/cherrypick release-0.17

PR #12425, opened 2026-06-23
Skip finished workloads in elastic job ungater reconcile

sohankunkerkar:
#### What type of PR is this?
/kind cleanup

related to https://github.com/kubernetes-sigs/kueue/pull/12045#discussion_r3414461956

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined elastic job ungating behavior to only ungates workloads with active quota reservation, preventing unintended ungating of finished workloads that retain reserved quota.

* **Tests**
  * Added test coverage to verify finished workloads with reserved quota remain properly gated.

### DIFF
--- a/pkg/controller/elasticjobs/elastic_job_ungater.go
+++ b/pkg/controller/elasticjobs/elastic_job_ungater.go
@@ -105,13 +105,10 @@ func (r *elasticJobUngater) Reconcile(ctx context.Context, req reconcile.Request
 	if !r.expectationsStore.Satisfied(log, req.NamespacedName) {
 		return reconcile.Result{}, errPendingUngateOps
 	}
-	// Ungate pods for workloads that were admitted (including finished ones
-	// whose pods may still be gated after scale-up). Pending workloads that
-	// were never admitted keep their pods gated.
 	if !workloadslicing.IsElasticWorkload(wl) {
 		return reconcile.Result{}, nil
 	}
-	if !workload.IsAdmitted(wl) && !workload.HasQuotaReservation(wl) {
+	if !workload.HasActiveQuotaReservation(wl) {
 		return reconcile.Result{}, nil
 	}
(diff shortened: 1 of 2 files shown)


pkg/controller/elasticjobs/elastic_job_ungater.go
```diff
@@ -105,13 +105,11 @@ func (r *elasticJobUngater) Reconcile(ctx context.Context, req reconcile.Request
...
 	}
-	// Ungate pods for workloads that were admitted (including finished ones
-	// whose pods may still be gated after scale-up). Pending workloads that
-	// were never admitted keep their pods gated.
+	// Skip finished workloads and pending workloads that were never admitted.
 	if !workloadslicing.IsElasticWorkload(wl) {
 		return reconcile.Result{}, nil
 	}
-	if !workload.IsAdmitted(wl) && !workload.HasQuotaReservation(wl) {
+	if workload.IsFinished(wl) || (!workload.IsAdmitted(wl) && !workload.HasQuotaReservation(wl)) {
```
mimowo: Should we rather check !IsAdmissible() ? Seems to capture the intention better and cover some other cases

sohankunkerkar: I think switching to `HasActiveQuotaReservation` would help here as it covers finished, deactivated, and never-admitted in one check. Also changed `shouldUngate` to avoid enqueueing no-op reconciles

sohankunkerkar: After investigating, the original code is correct. Finished workloads with `QuotaReserved=True` need their pods ungated during elastic slice replacement, so skipping them breaks InTreeAutoscaling e2e tests deterministically. Closing this PR as the cleanup isn't safe

PR #12434, opened 2026-06-23
Add kueue_pod_scheduling_gate_removal_seconds metric for TopologySchedulingGate.

mbobrovskyi:
#### What type of PR is this?
/kind feature

Add kueue_pod_scheduling_gate_removal_seconds metric for TopologySchedulingGate.

Part of #11170

```release-note
NONE
```

### DIFF
--- a/pkg/controller/jobs/pod/pod_controller.go
+++ b/pkg/controller/jobs/pod/pod_controller.go
@@ -49,7 +49,6 @@ import (
 	"sigs.k8s.io/kueue/pkg/controller/jobframework"
 	podconstants "sigs.k8s.io/kueue/pkg/controller/jobs/pod/constants"
 	"sigs.k8s.io/kueue/pkg/features"
-	"sigs.k8s.io/kueue/pkg/metrics"
 	"sigs.k8s.io/kueue/pkg/podset"
 	"sigs.k8s.io/kueue/pkg/util/api"
 	clientutil "sigs.k8s.io/kueue/pkg/util/client"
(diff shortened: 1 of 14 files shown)


pkg/util/pod/pod.go
```diff
@@ -165,3 +173,28 @@ func ContainersShape(containers []corev1.Container) (result []map[string]any) {
...
+
+func RecordPodSchedulingGateRemovalSeconds(cl clock.Clock, name string, wl *kueue.Workload, isGroup bool) {
+	cond := apimeta.FindStatusCondition(wl.Status.Conditions, kueue.WorkloadAdmitted)
+	if cond == nil || cond.Status != metav1.ConditionTrue {
+		return
+	}
+	latency := cl.Now().Sub(cond.LastTransitionTime.Time)
+	metrics.RecordPodSchedulingGateRemovalSeconds(name, wl.Status.Admission.ClusterQueue, isGroup, latency)
+}
+
```
mimowo, line 185: Let's commonize with the other function

pkg/controller/jobs/pod/pod_controller.go
```diff
@@ -325,21 +334,14 @@ func (p *Pod) Run(ctx context.Context, c client.Client, wl *kueue.Workload, podS
 			recorder.Eventf(pod, nil, corev1.EventTypeNormal, jobframework.ReasonStarted, "Started", msg)
 		}
 
-		p.recordPodSchedulingGateRemovalSeconds(wl)
+		for _, gate := range gates {
+			utilpod.RecordPodSchedulingGateRemovalSeconds(p.clock, gate, wl, p.isGroup)
+		}
```
mimowo: Hm, I don't get it, why updating any other gates from pod_controller since it is only responsible by removing the "admission" gate. I think it may generate wrong results if we update the metric for other gates. 

So, 
- "topology" gate is only removed by the TopologyUngater
- "elastic-job" gate is only removed by the ElasticJobUngater

pkg/util/pod/pod.go
mbobrovskyi: Yeah, it was moved from pod_controller.go.

pkg/controller/jobs/pod/pod_controller.go
mbobrovskyi: Fixed.

mimowo (commented):
/lgtm
/approve
Looks great, thank you for extending the metric 👍

PR #12436, opened 2026-06-23
fix(tas): account for elastic workload leaders

### LINKED ISSUE #12409: TAS + Elastic workloads: leader skipped except scale-ups
Suspected bug, based on my static reading of the code (and initial consulting with AI).
Not observed in practice.

In [`handleElasticWorkload`](https://github.com/kubernetes-sigs/kueue/blob/38c962f6af181667c8a0440733ed0f83d98448f7/pkg/cache/scheduler/tas_elastic_workloads.go#L35):

* In the [scale-up case](https://github.com/kubernetes-sigs/kueue/blob/38c962f6af181667c8a0440733ed0f83d98448f7/pkg/cache/scheduler/tas_elastic_workloads.go#L58), we [do account](https://github.com/kubernetes-sigs/kueue/blob/38c962f6af181667c8a0440733ed0f83d98448f7/pkg/cache/scheduler/tas_elastic_workloads.go#L98-L101) for the leader (if it exists), both in `assumedUsage` and in the returned topology assignment.
* However, in the [scale-down](https://github.com/kubernetes-sigs/kueue/blob/38c962f6af181667c8a0440733ed0f83d98448f7/pkg/cache/scheduler/tas_elastic_workloads.go#L60) and [no-change](https://github.com/kubernetes-sigs/kueue/blob/38c962f6af181667c8a0440733ed0f83d98448f7/pkg/cache/scheduler/tas_elastic_workloads.go#L64) cases, the leader is completely ignored.

This seems to be an omission, potentially leading to:
- getting an unrelated Pod stuck because it was assigned leader's capacity falsely assumed to be free
- getting the leader itself stuck, in case it needed a restart and a new placement (which will be missing)

RooobinYe:
#### What type of PR is this?
/kind bug
/area tas

This PR fixes TAS placement for elastic workloads with leader/worker pod sets.

Fixes #12409

Validation:
- `go test ./pkg/cache/scheduler -count=1`

```release-note
TAS: Fixed elastic workload placement to preserve leader pod set assignments and capacity accounting when worker counts stay the same or scale down.
```

## Summary by CodeRabbit

* **Tests**
  * Added coverage for elastic workload pod set recomputation across scale-down and same-count scenarios, verifying correct preservation/reuse of leader and worker pod assignments.

* **Bug Fixes**
  * Fixed elastic workload scheduling to correctly handle leader assignment preservation during worker truncation on scale-down, and to reuse prior topology assignments appropriately when pod counts remain the same.

### DIFF
--- a/pkg/cache/scheduler/tas_elastic_workloads.go
+++ b/pkg/cache/scheduler/tas_elastic_workloads.go
@@ -51,18 +51,15 @@ func (s *TASFlavorSnapshot) handleElasticWorkload(
 	}
 
 	previousCount := utiltas.CountPodsInAssignment(prevAssignment)
-	result := make(map[kueue.PodSetReference]tasPodSetAssignmentResult)
 
 	switch {
 	case workers.Count > previousCount:
 		return s.handleScaleUp(workers, leader, prevAssignment, previousCount, assumedUsage, opts)
 	case workers.Count < previousCount:
-		return s.handleScaleDown(workers, prevAssignment, assumedUsage)
+		return s.handleScaleDown(workers, leader, prevAssignment, assumedUsage)
 	default:
 		// Same count: reuse previous assignment
-		result[workers.PodSet.Name] = tasPodSetAssignmentResult{TopologyAssignment: prevAssignment}
-		addAssumedUsage(assumedUsage, prevAssignment, &workers)
-		return elasticPlacementResult{applied: true, assignments: result}
+		return s.finalizeElasticAssignment(workers, prevAssignment, leader, assumedUsage)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


olekzabl (approved):
Thank you!

/lgtm

olekzabl:
Side note: looking at this now, I have a feeling that all our calls to `addAssumedUsage` duplicate the other logic without a good reason.
I'm suspecting that ideally we could refactor them to a single call, based on the TAS result, placed e.g. [here](https://github.com/kubernetes-sigs/kueue/blob/8b18b7ec3d9c8a9dc4c3079eed2c333ae662263c/pkg/cache/scheduler/tas_flavor_snapshot.go#L638) (or maybe in a go-routine to catch early-exit paths, like [here](https://github.com/kubernetes-sigs/kueue/blob/8b18b7ec3d9c8a9dc4c3079eed2c333ae662263c/pkg/cache/scheduler/tas_flavor_snapshot.go#L616), if desired).

However, this is just code cleanup, out of scope of the original issue behind this PR.
So I'll soon open another cleanup issue for that.

[EDIT: just opened #12485]

pkg/cache/scheduler/tas_elastic_workloads.go
```diff
@@ -57,11 +57,18 @@ func (s *TASFlavorSnapshot) handleElasticWorkload(
...
 		return s.handleScaleUp(workers, leader, prevAssignment, previousCount, assumedUsage, opts)
 	case workers.Count < previousCount:
-		return s.handleScaleDown(workers, prevAssignment, assumedUsage)
+		return s.handleScaleDown(workers, leader, prevAssignment, assumedUsage)
 	default:
 		// Same count: reuse previous assignment
 		result[workers.PodSet.Name] = tasPodSetAssignmentResult{TopologyAssignment: prevAssignment}
 		addAssumedUsage(assumedUsage, prevAssignment, &workers)
+
+		if leader != nil && leader.PreviousAssignment != nil {
```
mwielgus: In what situations leader doesn't have previousassignment?

RooobinYe: @mwielgus

The leader and workers read their `PreviousAssignment` independently from the old admission.

If the old admission was produced by the previous code's scale-down or same-count path — which only wrote workers, not the leader — then `getPreviousTopologyAssignment` returns the workers' assignment but nil for the leader.

The nil guard prevents the subsequent `InternalFrom(nil)` panic.

```go
leaderPrevAssignment := utiltas.InternalFrom(leader.PreviousAssignment)
```

olekzabl: `InternalFrom(nil)` does not panic, see [here](https://github.com/kubernetes-sigs/kueue/blob/a6af4d05e7faf7f0da85756303e512da98cbbe03/pkg/util/tas/tas_assignment.go#L126-L128).

pkg/cache/scheduler/tas_elastic_workloads.go
```diff
@@ -57,11 +57,18 @@ func (s *TASFlavorSnapshot) handleElasticWorkload(
...
+			leaderPrevAssignment := utiltas.InternalFrom(leader.PreviousAssignment)
+			result[leader.PodSet.Name] = tasPodSetAssignmentResult{TopologyAssignment: leaderPrevAssignment}
+			addAssumedUsage(assumedUsage, leaderPrevAssignment, leader)
+		}
```
mimowo: This code looks duplicated with https://github.com/kubernetes-sigs/kueue/pull/12436/changes#diff-642f3e7eead554222faa9ac6647de46f0ba360d718f38c30a197e5f1acd866e3R128-R132. Let's commonize using a helper function

mimowo: We may also exend the scope of the comminized code to cover L63 - L72

RooobinYe: Solved.

I kept the `default` branch content consistent with the `case` branches, also using a single function call.

pkg/cache/scheduler/tas_elastic_workloads.go
RooobinYe: Yes, you are right. I removed `&& leader.PreviousAssignment != nil`.

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for hardening of the elastic jobs 👍

PR #12437, opened 2026-06-23
Drop removed GangScheduling feature gate from WAS e2e kind config

### LINKED ISSUE #12435: WAS e2e on k8s-main fails after GangScheduling feature gate removal
**What happened**:

periodic-kueue-test-e2e-k8s-main-was has failed on every run since Jun 17 ~12:18 UTC.

Job history:
https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/logs/periodic-kueue-test-e2e-k8s-main-was

The kind cluster never finishes bootstrapping. kubeadm waits in wait-control-plane and eventually fails with unable to create ClusterRoleBinding: ... connection refused, because kube-apiserver never starts.

The root cause appears to be that the WAS kind config still enables the removed GangScheduling feature gate:

https://github.com/kubernetes-sigs/kueue/blob/main/hack/testing/e2e-common.sh#L487

That gate was removed upstream in kubernetes/kubernetes#139520 and consolidated into GenericWorkload per KEP-4671. kind propagates the gate to the control-plane components, which now reject it on k8s-main with unrecognized feature gate: GangScheduling.

This is not a transient flake: it fails deterministically on every run and every retry attempt. I am splitting this out from #12307, which tracked a transient bootstrap flake with the same surface symptom.

WAS e2e against k8s-main should build the kind cluster successfully and run the suite.

Run the WAS e2e against k8s-main, for example:

WAS_ENABLED=true make test-e2e-k8s-main-was

Or inspect any recent periodic run, for example:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/periodic-kueue-test-e2e-k8s-main-was/2069211073285722112

Each run rebuilds k8s-main, so it picks up the upstream feature-gate removal.

The timing lines up with the upstream merge:
|Time (UTC) | Build | Result|
|-- | -- | --|
|Jun 17 00:43 | 2067036… | last green|
|Jun 17 07:03 | — | kubernetes/kubernetes#139520 merged|
|Jun 17 12:18 | 2067217… | first red|
|Jun 18–23 | …2069211… | consistently red|

Proposed fix: drop GangScheduling from the WAS patch and keep GenericWorkload=true. I will confirm whether the suite needs any successor gates, such as CompositePodGroup / TopologyAwareWorkloadScheduling, before sending the PR.

/kind bug
/area test
/assign

apullo777:
#### What type of PR is this?

/kind bug
/area testing

The WAS e2e kind config enabled the `GangScheduling` feature gate, which upstream merged into `GenericWorkload` (kubernetes/kubernetes#139520). 

On k8s-main the control-plane components now reject it as an unrecognized feature gate and fail to start, so the kind cluster never bootstraps; `periodic-kueue-test-e2e-k8s-main-was` has failed on every run since Jun 17. 

Fixes #12435

```release-note
NONE
```

## Summary by CodeRabbit

* **Chores**
  * Updated test environment configuration to refine feature gate settings and improve test coverage.

### DIFF
--- a/hack/testing/e2e-common.sh
+++ b/hack/testing/e2e-common.sh
@@ -484,7 +484,6 @@ function patch_kind_config_for_was {
     cp "$1" "$patched_config"
 
     $YQ -i '.featureGates.GenericWorkload = true' "$patched_config"
-    $YQ -i '.featureGates.GangScheduling = true' "$patched_config"
     $YQ -i '(.nodes[] | select(.role == "control-plane")).kubeadmConfigPatches[0] = "kind: ClusterConfiguration
 apiVersion: kubeadm.k8s.io/v1beta4
 scheduler:


mimowo:
we can test using the on-demand CI job, using now:
/test pull-kueue-test-e2e-k8s-main-was

mimowo:
Looks like it works. Let's follow up if more is needed,cc @kannon92 
/lgtm
/approve
Thank you for addressing the issue quickly 👍

PR #12440, opened 2026-06-23
feat: Increase OOTB QPS and concurrency for Kueue

### LINKED ISSUE #12411: Increase OOTB QPS and concurrency for Kueue

I would like to revisit the OOTB QPS and concurrency for Kueue.

In particular, currently we use QPS=50, Burst=100, which is very constraining for users.

For QPS the load can be rate-limited currently at the API -server layer with P&F.

Similarly, the current concurrency level for workload (5), LQ (1), CQ (1) seems constraining.

I propose:
- QPS: 300, Burst=500
- Workload concurrency: 10, LQ and CQ: 5.

**Why is this needed**:

Currently, the 300 - 500 pods/ second scheduling seems like the standard. However, Kueue becomes a bottleneck when TAS is used, because we need to remove scheduling gate from every pod with TopologyUngater. So, users get only 50 pods / second scheduling rate.

yuluo-yx:
#### What type of PR is this?

/kind feature

Fixes #12411

```release-note
Increase OOTB QPS and concurrency for Kueue: QPS: 300, Burst=500, Workload concurrency: 10, LQ and CQ: 5.
```

## Summary by CodeRabbit

* **Chores**
  * Increased default API client connection rate limits (QPS 20→300, Burst 30→500) and raised controller concurrency for Workload, LocalQueue, and ClusterQueue.
  * Updated Helm chart values and e2e/deployment configuration to use the new performance tuning.
* **Documentation**
  * Refreshed the MultiKueue kind development setup example to reflect the higher concurrency and client connection limits.
* **Tests**
  * Updated defaulting tests to expect the new client connection defaults.


tenzen-y (commented):
Thank you!
/lgtm
/approve

mimowo:
/lgtm
Thank you, I can see the last update is just using new(): https://github.com/kubernetes-sigs/kueue/pull/12440/changes/476bc3f5b32ea5313ff81654db7918798fc0ba00

Issue #12442, opened 2026-06-23
Commonize the "run" procedure running scheduling tests


mimowo:
**What would you like to be cleaned**:

I would like to comminize the "run" procedure for scheduling tests (e.g the main test in scheduler_test.go, or scheduler_tas_test.go)

USing the same approach as in https://github.com/kubernetes-sigs/kueue/pull/12441

**Why is this needed**:

To deduplicate the code. The tests could have been put together (they historicially were), but we decided to split only based on the fact that the tests were getting too long.

mimowo:
cc @apullo777

apullo777:
I will take this :) 
/assign

apullo777:
I took a quick look at the Schedule-style run loops in `pkg/scheduler/`. My read is that they split naturally into separate groups: the core scheduling tests, the TAS-specific tests, and possibly AFS

To keep the refactor reviewable, I would handle these in separate PRs. Strting with the core group first:`TestSchedule`, `TestScheduleForFairSharing`, and `TestScheduleConcurrentAdmission`, with a shared `scheduleTestCase` and `runScheduleTestCases` helper.

Then I would handle the TAS tests in a follow-up PR, since they have extra topology/node/pod setup and probably deserve their own runner. Will evaluate AFS last.

Does that split match what you had in mind?

mimowo:
Actually, I haven't analyzed the step by step refactoring, but this proposal sound reasonable. Indeed trying to do all in one go might result in non-reviewable mess, because the tests diverged a lot over time.

apullo777:
I checked the remaining bespoke `t.Run` loops in `pkg/scheduler/*_test.go`, and I do not think further commonization would improve readability. 

The remaining tests each have test-specific setup or scheduling flow, such as TAS admission/resource handling, AFS-specific cache setup, multi-iteration scheduling, concurrent-update interceptors, or single-scenario assertions.

Trying to fold those into the shared helpers would likely require extra knobs and branches, which may go against the goal of this cleanup.

So from my side, I think the high-value commonization is covered. Please feel free to close this if that matches your expectation, or let me know if there is another specific test you would still like me to look at :)

mimowo:
Sounds good, let me close for now. Indeed the most sticking examples are already comkonized. I will open a follow up issue if I see the need in the future.

PR #12445, opened 2026-06-23
Commonize the TAS preemption and cohort scheduling test run procedure

apullo777:
#### What type of PR is this?

/kind cleanup
/area testing

`TestScheduleForTASPreemption` and `TestScheduleForTASCohorts` duplicated the same TAS scheduling run/assert procedure.

This extracts a shared `tasScheduleTestCase` and `runTASScheduleTestCases` helper, then converts both tests to use it.

`TestScheduleForTAS` is left as-is because its run procedure differs materially, so folding it in would add complexity rather than remove duplication.

Part of #12442

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/scheduler_tas_test.go
+++ b/pkg/scheduler/scheduler_tas_test.go
@@ -3470,6 +3470,191 @@ func TestScheduleForTAS(t *testing.T) {
 	}
 }
 
+// tasScheduleTestCase is the shared case definition for the TAS preemption and cohort
+// scheduling tests. Each test sets only the fields it needs; unused fields keep their
+// zero value.
+type tasScheduleTestCase struct {
+	nodes           []corev1.Node
+	pods            []corev1.Pod
+	topologies      []kueue.Topology
+	cohorts         []kueue.Cohort
+	resourceFlavors []kueue.ResourceFlavor
+	clusterQueues   []kueue.ClusterQueue
+	workloads       []kueue.Workload
+	wantWorkloads   []kueue.Workload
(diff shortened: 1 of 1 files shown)


pkg/scheduler/scheduler_tas_test.go
```diff
@@ -7249,142 +7265,11 @@ func TestScheduleForTASCohorts(t *testing.T) {
...
-				if diff := cmp.Diff(tc.wantEvents, recorder.RecordedEvents, append(tc.eventCmpOpts, cmpopts.SortSlices(utiltesting.SortEvents))...); diff != "" {
-					t.Errorf("unexpected events (-want/+got):\n%s", diff)
-				}
-			})
-		}
-	}
+	runTASScheduleTestCases(t, tasScheduleTestConfig{
+		queues:     queues,
+		now:        now,
+		sortEvents: true,
```
mimowo, line 7271: I think we should actually use the sort in both cases to make sure the results are deterministic. This can be a follow up

mimowo (commented):
/lgtm
/approve
Thank you for driving the effort of deduplicating the boilerplate 👍 
/cherrypick release-0.18
/cherrypick release-0.17

mimowo:
Please cherrypick manully if the robot fails on conflicts

pkg/scheduler/scheduler_tas_test.go
apullo777: Okay! Handled this in follow-up PR #12468.

Issue #4106, opened 2025-01-30
Add Support for Plain Pods with Known OwnerReferences


BenHinthorne:
**What would you like to be added**:
I would like to add the ability for pods annotated with `kueue.x-k8s.io/standalone` to be suspended even if their owner is a known Kueue integration (open to suggestions one exact annotation wording). 

Currently, in the [pod_webhook](https://github.com/kubernetes-sigs/kueue/blob/198d276002b12066b19601da9a0a7ca8ae4879b3/pkg/controller/jobs/pod/pod_webhook.go#L172), if the pod has an ownerReference of a `Kind` known to Kueue, Kueue will not suspend this pod and a workload will never be created for it. I would like to add the ability to bypass this check, and allow the pod to be suspended and have a workload be created for it, if the user opts in via an annotation. 

**Why is this needed**:

At a high level, this would give users more flexibility to process plain pods with Kueue. For example, let's consider Spark Applications, which don't yet have native support within Kueue. In short, Spark Apps operate with a Driver pod, which then creates Executor pods. The Executor  owner is the Driver pod, and thus of Kind Pod. Currently, the ownerReference check mentioned above prevents the user from being able to have Kueue process the Executors at all. If users could add the `standalone` annotation to the Executor pods, then they would be able to use Kueue to manage Spark Apps via plain pods. The same concept could extend to any CRD not yet natively integrated for Kueue, that may follow a similar pattern of Pods owning Pods. 

Similarly, this option would also give users the flexibility to process _known workload types_ as plain pods, should they choose to do so. This flexibility would be helpful for users to be able to process dynamically sized jobs (at least until https://github.com/kubernetes-sigs/kueue/tree/main/keps/77-dynamically-sized-jobs#implementation-history is fully supported), as well as give them the choice if they would like gang scheduling behavior or not. For example, let's consider RayJobs. Currently, RayJobs will be processed by Kueue as a single workload, and any pods resulting from the job autoscaling aren't yet supported. If users could allow for pods owned by a RayJob to be processed as plain pods, then they could choose to process RayJobs as single pods (forgoing gang scheduling), and process any dynamically scaled pods in the same way. 

**Completion requirements**:

This enhancement requires the following artifacts:

Will link an example PR in the comments, and wait for feedback before proceeding with any doc updates. 

- [ ] Design doc
- [ ] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

BenHinthorne:
[...]

PR #12447, opened 2026-06-23
Add doc for spark-submit plain-pods kueue integration

zemin-piao:
#### What type of PR is this?

/kind documentation

Add docs on how to do plain spark-submit integration with kueue(non spark operator way), mentioned at https://github.com/kubernetes-sigs/kueue/issues/4106#issuecomment-3281251848

Fixes #

```release-note
NONE
```

## Summary by CodeRabbit

* **Documentation**
  * Added a new guide for running Apache Spark with `spark-submit` directly on Kubernetes while using Kueue’s Plain Pod workflow.
  * Added a linked entry in the external workloads docs navigation.
  * Documented setup requirements, queue selection, example configuration, and current limitations of this community-supported approach.

### DIFF
--- a/site/content/en/docs/tasks/run/external_workloads/_index.md
+++ b/site/content/en/docs/tasks/run/external_workloads/_index.md
@@ -19,4 +19,5 @@ You can use AppWrapper, job-based workloads and pod-based workloads.
 - [Run an Argo Workflow using pod integration](/docs/tasks/run/external_workloads/argo_workflow).
 - [Run a tekton cd pipeline using pod integration](/docs/tasks/run/external_workloads/tektoncd)
 - [Run a Sandbox using pod integration](/docs/tasks/run/external_workloads/sandbox)
+- [Run Spark with spark-submit using pod integration](/docs/tasks/run/external_workloads/spark_submit_plain_pod)
 
--- a/site/content/en/docs/tasks/run/external_workloads/spark_submit_plain_pod.md
+++ b/site/content/en/docs/tasks/run/external_workloads/spark_submit_plain_pod.md
@@ -1,0 +1,89 @@
+---
+title: "Run Spark with spark-submit using Plain Pods"
+linkTitle: "Spark (Plain Pods)"
+date: 2026-06-24
+weight: 9
+description: >
+  Integrate Kueue with Spark applications submitted via spark-submit using the Plain Pod integration.
+---
+
+This page shows how to leverage Kueue's scheduling and resource management capabilities when running
+[Spark](https://spark.apache.org/docs/latest/running-on-kubernetes.html) applications submitted with
+`spark-submit` directly to Kubernetes, i.e. without the [Spark Operator](https://github.com/kubeflow/spark-operator).
+
+This guide is for [batch users](/docs/tasks#batch-user) that have a basic understanding of Kueue. For more information, see [Kueue's overview](/docs/overview).
+
+If you run Spark through the Spark Operator, prefer the first-class
+[SparkApplication integration](/docs/tasks/run/kubeflow/sparkapplications/) instead.
+
+When using `spark-submit` in cluster mode, Spark creates a driver Pod that in turn creates the
+executor Pods. Kueue can manage these Pods through the [Plain Pod](/docs/tasks/run/plain_pods)
+integration, where the driver Pod and each executor Pod are represented as independent Plain Pods.
+
+{{% alert title="Note" color="warning" %}}
+This is not an officially supported Kueue integration for Spark; it is a community workaround that
+works with the current Plain Pod implementation. It relies on explicitly opting executor Pods into
+Kueue's `pod` integration and may change in a future release. See
+[issue #4106](https://github.com/kubernetes-sigs/kueue/issues/4106) for the discussion on a
+first-class solution.
+{{% /alert %}}
+
+## Before you begin
+
+1. Learn how to [install Kueue with a custom manager configuration](/docs/installation/#install-a-custom-configured-released-version).
+
+1. Follow the steps in [Run Plain Pods](/docs/tasks/run/plain_pods/#before-you-begin) to learn how to enable and configure the `pod` integration.
+
+1. Check [Administer cluster quotas](/docs/tasks/manage/administer_cluster_quotas) for details on the initial Kueue setup.
+
+1. Install [Spark](https://spark.apache.org/downloads.html) and learn how to
+[run Spark on Kubernetes](https://spark.apache.org/docs/latest/running-on-kubernetes.html) with `spark-submit`.
+
+## Queue selection
+
+Kueue can manage the Spark driver Pod through the regular Plain Pod flow. Explicitly set the
+`kueue.x-k8s.io/queue-name` label on the driver Pod, or rely on a default LocalQueue if one is
+configured in the namespace.
+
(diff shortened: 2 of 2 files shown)


kannon92:
cc @everpeace

everpeace:
The mentioned method in https://github.com/kubernetes-sigs/kueue/issues/4106#issuecomment-3281251848 works in the current Kueue implementation.

However, I am not sure whether the manual usage of `kueue.x-k8s.io/pod-suspending-parent=pod` annotation and `kueue.x-k8s.io/managed=true` label should be documented as the official way or not. Actually, I still feel it's a kind of hacky way.

So, I think documenting this on the official doc should be discussed with Kueue leads.

mimowo:
cc @kannon92 @amy

kannon92:
I agree with @everpeace take. I'm not really that knowledgeable in this area to comment on this. Is this a pattern we want to encourage?

zemin-piao:
Thank you @everpeace @kannon92 @mimowo for replying! 

I think from the discussion of https://github.com/kubernetes-sigs/kueue/issues/4106 I can see why you folks have doubt on documenting this approach in the official doc. (e.g. such integration skips what the webhook should do about suspending the pods, adding labels + annotations).

To be clear about my motivation on this PR: I'm not trying to make manual usage of `kueue.x-k8s.io/pod-suspending-parent=pod ` and `kueue.x-k8s.io/managed=true` to be the official way on the kueue <-> spark non operator integration. 

The motivation of having such doc from me is about easily figuring out there is a way to do Kueue <-> spark (without spark operator) integration. For me as a new comer to Kueue, it took me quite some time to figure out how to do it (e.g. reading the source code, trial and error) until I found this thread https://github.com/kubernetes-sigs/kueue/issues/4106#issuecomment-3281251848 to actually make it work. Also there are many other organisations using spark-submit (non operator) way of doing spark on k8s hence there can be many others faced with the same puzzle. Within spark ecosystem, spark operator is optional and `spark-submit` (native spark script creating pods on k8s without an operator) is the spark "native" way of running spark applications. With such doc it can help folks quickly know there is an option of doing it. 

Given your your doubts that this is hacky, I'd like to propose **mentioning that this is not an officially supported way for Kueue <-> spark non operator integration but an option that community has tried and works**. WDYT? Happy to rework the PR so you folks are comfortable with it.

kannon92:
Thank you for the context.

I'd be fine to merge this but I'd like us to follow https://kueue.sigs.k8s.io/docs/tasks/run/external_workloads/.

Maybe you can add a section on how to support spark-submit workloads.

site/content/en/docs/tasks/run/plain_pods.md
```diff
@@ -90,6 +90,40 @@ Once the corresponding Workload is admitted, Kueue will patch the pods and remov
...
+Add the queue label, the `kueue.x-k8s.io/pod-suspending-parent` annotation, and
+the `kueue.x-k8s.io/managed: "true"` label to the child Pod:
+
+```yaml
+metadata:
+  labels:
+    kueue.x-k8s.io/queue-name: user-queue
+    kueue.x-k8s.io/managed: "true"
+  annotations:
+    kueue.x-k8s.io/pod-suspending-parent: pod
```
amy: >  kueue.x-k8s.io/pod-suspending-parent: pod

`pod` can be a longer description placeholder. Do like `pod-parent-name` or something.

site/content/en/docs/tasks/run/plain_pods.md
```diff
@@ -90,6 +90,40 @@ Once the corresponding Workload is admitted, Kueue will patch the pods and remov
 - In case of [preemption](/docs/concepts/cluster_queue/#preemption), the Pod will
   be terminated and deleted.
 
+### f. Pods created by another Pod
+
+When a Pod is created by another Pod, Kueue won't manage it through the `pod`
+integration unless you opt it in explicitly. This pattern is useful for Spark
```
amy: `When a Pod is created by another Pod, Kueue won't manage it through the pod integration unless you opt it in explicitly. This pattern is useful for Spark executor Pods created by a Spark driver Pod.` 

This part can be improved. Kueue doesn't manage random pods without expected annotations (its not just pod created by another pod), so I don't think you need the first sentence. Also specify what pattern.

site/content/en/docs/tasks/run/kubeflow/sparkapplications.md
```diff
@@ -59,6 +59,16 @@ metadata:
 SparkApplication integration does not support dynamic allocation. If you set `spec.dynamicAllocation.enabled=true` in SparkApplication, Kueue will reject such resources in the webhook.
 {{% /alert %}}
 
+{{% alert title="Note" color="primary" %}}
+If you need to queue Spark Pods directly, for example when using pod-based
+`spark-submit` flows or driver-created executor Pods, you can use Kueue's
+[plain Pod integration](/docs/tasks/run/plain_pods/). In that mode, executor
+Pods should set `kueue.x-k8s.io/pod-suspending-parent: pod` and the queue label.
```
amy: and the queue label -> add the label example

site/content/en/docs/tasks/run/kubeflow/sparkapplications.md
```diff
@@ -59,6 +59,16 @@ metadata:
...
+Current pod-based Spark flows also set `kueue.x-k8s.io/managed: "true"`
```
amy: What does this mean: `Current pod-based Spark flows also set `kueue.x-k8s.io/managed: "true"``

Like does this flow need this? If it does, remove the "current". Just say "Also set blah blah on executor pods"

zemin-piao: Good catch! 

As per https://github.com/kubernetes-sigs/kueue/pull/12447#issuecomment-4793313293: I'll create a new page for spark-submit plain pod integration, hence gonna rephrase a bit with standalone context within that page

site/content/en/docs/tasks/run/plain_pods.md
zemin-piao: Will do `spark-driver` in this case to make it explicit: spark driver is the "leader" pod within the spark application which creates + owns the executor pods

amy (commented):
/lgtm

everpeace:
> I'd be fine to merge this but I'd like us to follow https://kueue.sigs.k8s.io/docs/tasks/run/external_workloads/.
> Maybe you can add a section on how to support spark-submit workloads.

I still lean to this plan which @kannon92 proposed rather than including this practice in plain pods integration page.


> Given your your doubts that this is hacky, I'd like to propose mentioning that this is not an officially supported way for Kueue <-> spark non operator integration but an option that community has tried and works. WDYT? Happy to rework the PR so you folks are comfortable with it.

Additionally, mentioning this method is NOT a standard would be better as @zemin-piao proposed. (I think "standard" would better than "official" in wording)


Then, sparkapplication page could have a link to the spark-submit workloads page.


WDYT??

site/content/en/docs/tasks/run/kubeflow/sparkapplications.md
zemin-piao: Added 👍

kannon92:
Thank you for moving this guide to external frameworks.

/approve

PR #12449, opened 2026-06-24
Fix flavor selector using union of label keys across resource group

### LINKED ISSUE #12446: FlavorAssigner rejects a flavor for node-affinity on a label key it doesn't declare (resource-group LabelKeys is a union, but matching is per-flavor)
**What happened**:
We have a working setup with Kueue and a few resource flavours created. We want to route new types of jobs through Kueue and add new resource flavours for them to be admitted to. These resource flavours will match on additional labels to the previously existing resource flavours.

We found that within a single resource group, a workload's required node affinity is filtered against the union of all flavors' label keys, but then matched against each flavor individually. So an affinity term on key K that some other flavor in the group declares will cause a flavor that doesn't declare K to be rejected for a given workload even though K is irrelevant to that flavor. 

This has caused some issues on our end and resulted in a need for a complicated unnecessary workaround to be able to add new resource flavors that have specific labels that don't exist on previously existing resource flavors.

Is there a reason things are this way or can it change to do the comparison per resource flavor's label?

We have flavors declaring certain labels and flavors that don't. When a job that has that label declared comes in and is tested for admission against a flavor that doesn't have that label declared, it shouldn't be rejected. Currently it is rejected because the current code builds LabelKeys as the union of labels on all resource flavors here: https://github.com/kubernetes-sigs/kueue/blob/fefb31a40508e4447d509ed77598e1fb15e92e4b/pkg/cache/scheduler/clusterqueue.go#L390

**How to reproduce it**:
Create 2 resource flavours on a resource group. Flavor A declares a label X =a ,  flavor B declares that label X= b and another additional label Y = k.
Create a job that has label X=a, Y=g.
The job will be stuck in pending forever instead of being admitted to flavor A

**Environment**:
- Kubernetes version : v1.33.12-eks-0247562
- Kueue version : v0.17.2

carterpewpew:
#### What type of PR is this?
/kind bug

`flavorSelector` was built once per resource group using the union of all flavors' label keys. This caused a flavor to be rejected when the workload referenced a label key declared by a different flavor in the same group, leaving workloads stuck pending forever. This PR computes the selector per-flavor using only that flavor's own keys.

Fixes #12446

ResourceGroup.LabelKeys is removed entirely; each flavor's keys are extracted at match time inside checkFlavorForPodSets.
Behavioral note: a flavor with no declared node labels will now match any workload (all selector keys are stripped as irrelevant). Previously it inherited keys from sibling flavors via the union, which was an accidental side effect. The upstream test TestIsNoFitDueToCapacityAndLimits/node_affinity_mismatch was updated accordingly.

```release-note
Scheduling: Fixed a bug where a workload could be stuck pending when its node selector referenced a label key declared by a different flavor in the same resource group.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workload-to-flavor matching by evaluating node-affinity selectors using only the currently considered flavor’s label keys, enabling assignments to labelless flavors when keys are irrelevant.
  * Refreshed resource-group flavor metadata used for cohort/snapshot labeling so cohort results reflect updated flavor topology and labeling behavior.

* **Tests**
  * Updated snapshot expectations for cohort scenarios.
  * Expanded flavor-assignment tests to validate selector key scoping and labelless-flavor behavior.

### DIFF
--- a/pkg/scheduler/flavorassigner/flavorassigner.go
+++ b/pkg/scheduler/flavorassigner/flavorassigner.go
@@ -941,13 +941,8 @@ func (a *FlavorAssigner) findFlavorForPodSets(
 	requests = filterRequestedResources(requests, resourceGroup.CoveredResources)
 
 	podSets := make([]*kueue.PodSet, len(psIDs))
-	selectors := make([]nodeaffinity.RequiredNodeAffinity, len(psIDs))
-
 	for idx, psID := range psIDs {
-		ps := &a.wl.Obj.Spec.PodSets[psID]
-		podSets[idx] = ps
-
-		selectors[idx] = flavorSelector(&ps.Template.Spec, resourceGroup.LabelKeys)
+		podSets[idx] = &a.wl.Obj.Spec.PodSets[psID]
 	}
 
 	var bestAssignment ResourceAssignment
@@ -965,7 +960,7 @@ func (a *FlavorAssigner) findFlavorForPodSets(
 			continue
 		}
 
-		if flavorStatus := a.checkFlavorForPodSets(log, fName, psIDs, podSets, selectors, resourceGroup); !flavorStatus.IsFit() {
+		if flavorStatus := a.checkFlavorForPodSets(log, fName, psIDs, podSets, resourceGroup); !flavorStatus.IsFit() {
 			flavorStatus.noFitReason = kueue.WorkloadQuotaReservedReasonNoMatchingFlavor
 			status.reasons = append(status.reasons, flavorStatus.reasons...)
 			consideredFlavors.AddNoFitFlavorAttempt(fName, flavorStatus)
@@ -1075,7 +1070,6 @@ func (a *FlavorAssigner) checkFlavorForPodSets(
 	flavorName kueue.ResourceFlavorReference,
 	psIDs []int,
 	podSets []*kueue.PodSet,
-	selectors []nodeaffinity.RequiredNodeAffinity,
 	rg *schdcache.ResourceGroup,
 ) *Status {
 	status := NewStatus()
@@ -1086,6 +1080,11 @@ func (a *FlavorAssigner) checkFlavorForPodSets(
 		status.appendf("flavor %s not found", flavorName)
 		return status
 	}
+
+	// Use only this flavor's own label keys (not the union across all flavors in
+	// the resource group) so that affinity terms referencing keys from other
+	// flavors are correctly ignored when evaluating this flavor.
+	flavorLabelKeys := sets.KeySet(flavor.Spec.NodeLabels)
 
 	for psIdx, psID := range psIDs {
 		if features.Enabled(features.TopologyAwareScheduling) {
@@ -1104,7 +1103,7 @@ func (a *FlavorAssigner) checkFlavorForPodSets(
 			status.appendf("untolerated taint %s in flavor %s", taint, flavorName)
 			return status
 		}
-		selector := selectors[psIdx]
+		selector := flavorSelector(&podSpec, flavorLabelKeys)
 		if match, err := selector.Match(&corev1.Node{ObjectMeta: metav1.ObjectMeta{Labels: flavor.Spec.NodeLabels}}); !match || err != nil {
 			if err != nil {
 				status.err = err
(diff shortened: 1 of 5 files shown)


mimowo:
/ok-to-test
@carterpewpew please set a release note, since there are user-facing consequences of the change. IIUC this is a bugfix for some scenarios

mimowo:
cc @sohankunkerkar @j-skiba ptal

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -1087,6 +1081,11 @@ func (a *FlavorAssigner) checkFlavorForPodSets(
 		return status
 	}
 
+	// Use only this flavor's own label keys (not the union across all flavors in
+	// the resource group) so that affinity terms referencing keys from other
+	// flavors are correctly ignored when evaluating this flavor.
+	flavorLabelKeys := sets.New(slices.Collect(maps.Keys(flavor.Spec.NodeLabels))...)
```
sohankunkerkar:
```suggestion
	flavorLabelKeys := sets.KeySet(flavor.Spec.NodeLabels)
```

pkg/scheduler/flavorassigner/flavorassigner_test.go
j-skiba: let's update the test name to something like `succeeds to schedule on flavor-b with a tolerated taint`

j-skiba (commented):
sgtm,

It totally makes sense to handle them this way but I wonder what was the idea to validate against that union in the first place and if it's still valid.

pkg/scheduler/flavorassigner/flavorassigner_test.go
```diff
@@ -1137,6 +1139,80 @@ func TestAssignFlavors(t *testing.T) {
...
+					Requests: corev1.ResourceList{
+						corev1.ResourceCPU: resource.MustParse("1"),
+					},
+					Count: 1,
+				}},
+				Usage: workload.Usage{Quota: resources.FlavorResourceQuantities{
+					{Flavor: "label-x-a", Resource: corev1.ResourceCPU}: resources.NewAmount(1_000),
+				}},
+			},
+		},
```
sohankunkerkar, line 1178: Could you add a test case here where both flavors declare the same key but neither value matches the workload?

sohankunkerkar (commented):
Thanks for the fix! The root cause analysis is correct. The union of LabelKeys across all flavors in a resource group caused cross-flavor contamination, rejecting workloads that should have been admitted. The per-flavor approach is the right fix. A few suggestions inline

mimowo:
I like the direction to recompile the selector when needed. The perf cost is minimal and it makes the correctness much easier to achieve. I think the previous code was premature optimization resulting in a subtle bug. Let's celebrate the fix with a release note.

sohankunkerkar:
/release-note-edit Fixed a bug where a workload could be stuck pending when its node selector referenced a label key declared by a different flavor in the same resource group.

mimowo:
image

sohankunkerkar:
/release-note-edit 
```release-note
Scheduling: Fixed a bug where a workload could be stuck pending when its node selector referenced a label key declared by a different flavor in the same resource group.
```

pkg/scheduler/flavorassigner/flavorassigner_test.go
carterpewpew: Updated, thanks. Renamed the test case to: succeeds to schedule on flavor-b with a tolerated taint.

pkg/scheduler/flavorassigner/flavorassigner.go
carterpewpew: Applied, thanks.

pkg/scheduler/flavorassigner/flavorassigner_test.go
carterpewpew: Included a new test case: node affinity mismatch when both flavors declare same key where both flavors have key type but values a/b and workload selector is type=non-existent, asserting NoMatchingFlavor for both.

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for improving the scheduling 👍

mimowo:
@carterpewpew could you please prepare CPs manually? You can make it semi-automated way using the ./hack/cherry_pick_pull.sh script

PR #12452, opened 2026-06-24
Setting UnadmittedWorkloadsObservability reasons in scheduler

j-skiba:
#### What type of PR is this?
/kind feature

This PR integrates the tracking and propagation of granular diagnostic reasons (encompassing `NoFit` failures, preemption/migration states, inactive queues, pods-readiness blocks, and configuration errors) into the core scheduling loop within scheduler under the `UnadmittedWorkloadsObservability` feature gate.

Part of #10852 

```release-note
When the `UnadmittedWorkloadsObservability` feature gate is enabled, workloads that fail to obtain a quota reservation now receive detailed diagnostic reasons in their `QuotaReserved` status condition (such as `WaitingForQuota`, `ExceedsMaxQuota`, `TopologyPlacementFailed`, or `NoMatchingFlavor`) along with an explicit `Admitted: False` condition.
```

## Summary by CodeRabbit

* **New Features**
  * Added more granular `WorkloadQuotaReserved` and `WorkloadAdmitted` reason values (e.g., waiting for preempted workloads, misconfiguration, suspended, pending evaluation, pending delayed topology requests, and no reservation).
  * Improved unadmitted-workload observability so condition reasons better match the exact scheduling/preemption outcome.
* **Bug Fixes**
  * Refined how quota-reservation and admission condition reasons are computed and patched during requeues and syncs, including observability-aware behavior.
* **Tests**
  * Updated/expanded scheduling and admission-check tests to assert the new reason values and run across feature-gate combinations, including adjusted expectations when observability is disabled.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -386,6 +386,7 @@ func (s *Scheduler) processEntry(
 	if features.Enabled(features.ConcurrentAdmission) && concurrentadmission.IsVariant(e.Obj) {
 		if moreFavorableSibling := s.findAdmittedMoreFavorableSibling(&e.Info, snapshot); moreFavorableSibling != nil {
 			log.V(3).Info("Skipping workload as a more favorable variant is already admitted", "moreFavorableVariant", klog.KObj(moreFavorableSibling.Obj))
+			e.quotaReservedReason = kueue.WorkloadQuotaReservedReasonPendingEvaluation
 			e.markSkipped("A more favorable variant is already admitted")
 			return
 		}
@@ -401,18 +402,21 @@ func (s *Scheduler) processEntry(
 	if mode == flavorassigner.NoFit {
 		e.requeueReason = qcache.RequeueReasonNoFit
 		log.V(3).Info("Skipping workload as FlavorAssigner assigned NoFit mode")
+		e.quotaReservedReason = e.assignment.NoFitReason
 		return
 	}
 
 	if mode == flavorassigner.Preempt {
 		if len(e.preemptionTargets) == 0 {
 			e.requeueReason = qcache.RequeueReasonPreemptionNoCandidates
+			e.quotaReservedReason = kueue.WorkloadQuotaReservedReasonWaitingForQuota
 			s.reserveCapacityForUnreclaimablePreempt(log, e, cq)
 			return
 		}
 		if (features.Enabled(features.ConcurrentAdmission) || features.Enabled(features.MultiKueueOrchestratedPreemption)) && workload.HasClosedPreemptionGate(e.Obj) {
 			gatedMsg := "Workload requires preemption, but it's gated"
 			log.V(3).Info(gatedMsg)
+			e.quotaReservedReason = kueue.WorkloadAdmissionGated
 			e.markPreemptionGated(gatedMsg)
 			return
 		}
@@ -421,6 +425,7 @@ func (s *Scheduler) processEntry(
 	// We skip multiple-preemptions per cohort if any of the targets are overlapping
 	if preemptedWorkloads.HasAny(e.preemptionTargets) {
 		e.markSkipped("Workload has overlapping preemption targets with another workload")
+		e.quotaReservedReason = kueue.WorkloadQuotaReservedReasonWaitingForQuota
 		skippedPreemptions[cq.Name]++
 		return
 	}
@@ -428,6 +433,7 @@ func (s *Scheduler) processEntry(
 	usage := e.assignmentUsage(log)
 	if !fits(snapshot, cq, &usage, preemptedWorkloads, e.preemptionTargets) {
 		e.markSkipped("Workload no longer fits after processing another workload")
+		e.quotaReservedReason = kueue.WorkloadQuotaReservedReasonWaitingForQuota
 		if mode == flavorassigner.Preempt {
 			skippedPreemptions[cq.Name]++
 		}
@@ -443,6 +449,7 @@ func (s *Scheduler) processEntry(
 	preemptionTargets, oldWorkloadSlice := workloadslicing.FindReplacedSliceTarget(e.Obj, e.preemptionTargets)
 
 	if mode == flavorassigner.Preempt {
+		e.quotaReservedReason = kueue.WorkloadQuotaReservedReasonWaitingForPreemptedWorkloads
 		s.issuePreemptions(ctx, log, e, preemptionTargets)
 		return
 	}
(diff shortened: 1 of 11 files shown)


pkg/util/testing/observability.go
```diff
@@ -0,0 +1,54 @@
...
+
+// AdjustConditionsForDisabledObservability adjusts a slice of workload status conditions
+// to match the scenario when the UnadmittedWorkloadsObservability feature gate is disabled.
+// Specifically, it performs the following modifications:
+//  1. Removes the 'Admitted' condition if its status is 'False' (as workloads that have no
+//     reservation did not receive an explicit 'Admitted: False' condition prior to the feature gate).
+//  2. Rolls back the reason of the 'QuotaReserved: False' condition to the generic "Pending"
+//     (desurfacing the descriptive reasons like 'WaitingForQuota' or 'WaitingForPreemptedWorkloads'
+//     introduced under the observability feature gate).
+func AdjustConditionsForDisabledObservability(conditions []metav1.Condition) []metav1.Condition {
```
mimowo, line 34: I like the new functions, but it seems they are only used for tests for now. So, I would prefer moving them to /pkg/util/testing

pkg/scheduler/scheduler_concurrent_admission_test.go
```diff
@@ -329,12 +350,30 @@ func TestScheduleConcurrentAdmission(t *testing.T) {
...
 
+	fgScenarios := []struct {
+		workloadRequestUseMergePatch     bool
+		unadmittedWorkloadsObservability bool
+	}{
+		{false, false},
+		{false, true},
+		{true, false},
+		{true, true},
+	}
```
mimowo: this is hard to follow. Could you maybe introduce the featureGates map as in other tests?

pkg/workload/admissionchecks.go
```diff
@@ -41,10 +42,21 @@ func SyncAdmittedCondition(w *kueue.Workload, now time.Time) bool {
 	hasAllChecksReady := HasAllChecksReady(w)
 	isAdmitted := IsAdmitted(w)
 	hasAllTopologyAssignmentsReady := !HasTopologyAssignmentsPending(w)
+	admittedCond := apimeta.FindStatusCondition(w.Status.Conditions, kueue.WorkloadAdmitted)
+	quotaReservedCond := apimeta.FindStatusCondition(w.Status.Conditions, kueue.WorkloadQuotaReserved)
 
-	if isAdmitted == (hasReservation && hasAllChecksReady && hasAllTopologyAssignmentsReady) {
-		return false
+	if !features.Enabled(features.UnadmittedWorkloadsObservability) {
```
mimowo: Let's check the positive case first to avoid negations as much as possible - they make code harder to read

pkg/scheduler/scheduler_afs_test.go
```diff
@@ -635,13 +681,17 @@ func TestScheduleForAFS(t *testing.T) {
 					t.Fatalf("Unexpected list workloads error: %v", err)
 				}
 
+				if !scenario.unadmittedWorkloadsObservability {
+					utiltestingapi.AdjustWorkloadsForDisabledObservability(wantWorkloads)
+				}
```
mimowo: let's put this somehow close to the copying of wantWorkloads so that a reader can easily match the related code, now it is scaterred

mimowo:
cc @kshalot @sohankunkerkar ptal, even if I consider it ready to merge to unblock downstream work, then your comments might bring good ideas for follow ups

pkg/util/testing/observability.go
j-skiba: moved the file to `pkg/util/testing`

pkg/scheduler/scheduler_concurrent_admission_test.go
j-skiba: refactored to

```go
scenarios := []map[featuregate.Feature]bool{
		{
			features.WorkloadRequestUseMergePatch:     false,
			features.UnadmittedWorkloadsObservability: false,
		},
		...
	}
```

pkg/workload/admissionchecks.go
j-skiba: sounds good, done

pkg/scheduler/scheduler_afs_test.go
j-skiba: done

pkg/workload/admissionchecks.go
```diff
@@ -56,19 +68,28 @@ func SyncAdmittedCondition(w *kueue.Workload, now time.Time) bool {
...
+		newCondition.Reason = kueue.WorkloadAdmittedReasonNoReservation
 		newCondition.Message = "The workload has no reservation"
 	case !hasAllChecksReady:
 		newCondition.Status = metav1.ConditionFalse
-		newCondition.Reason = "UnsatisfiedChecks"
+		if features.Enabled(features.UnadmittedWorkloadsObservability) {
+			newCondition.Reason = kueue.WorkloadAdmittedReasonUnsatisfiedAdmissionChecks
+		} else {
+			newCondition.Reason = "UnsatisfiedChecks"
+		}
```
mimowo: This pattern repeats in a couple of places. I think it would be nice to have a helper like
`workload.UnadmittedWorkloadReasonWithFallback(granularReason, fallback)`. I know the name is lengthy, but it would help to convey the intention well, and avoid duplicating the code,

pkg/scheduler/scheduler.go
```diff
@@ -965,8 +986,12 @@ func (s *Scheduler) requeueAndUpdate(ctx context.Context, e entry) {
 		Info("Workload re-queued", "workload", klog.KObj(e.Obj), "clusterQueue", klog.KRef("", string(e.ClusterQueue)), "queue", klog.KRef(e.Obj.Namespace, string(e.Obj.Spec.QueueName)), "requeueReason", e.requeueReason, "added", added, "status", e.status)
 	if e.status == notNominated || e.status == skipped || e.status == preemptionGated {
 		wl := e.Obj.DeepCopy()
+		condReason := "Pending"
+		if features.Enabled(features.UnadmittedWorkloadsObservability) {
+			condReason = e.quotaReservedReason
+		}
```
mimowo: Here I think we could also use the helper proposed in the other comment

pkg/scheduler/scheduler.go
```diff
@@ -525,7 +534,11 @@ func (s *Scheduler) waitForPodsReadyIfBlocked(ctx context.Context, log logr.Logg
 	log.V(5).Info("Waiting for all admitted workloads to be in the PodsReady condition")
 	wl := e.Obj.DeepCopy()
 	if err := workload.PatchAdmissionStatus(ctx, s.client, wl, s.clock, func(wl *kueue.Workload) (bool, error) {
-		return workload.UnsetQuotaReservationWithCondition(wl, "Waiting", "waiting for all admitted workloads to be in PodsReady condition", s.clock.Now()), nil
+		reason := "Waiting"
+		if features.Enabled(features.UnadmittedWorkloadsObservability) {
+			reason = kueue.WorkloadQuotaReservedReasonWaitingForPodsReady
+		}
```
mimowo: Here I think we could also use the helper proposed in the other comment

pkg/workload/admissionchecks.go
j-skiba: I extracted the code to such helper

```go
func UnadmittedWorkloadReasonWithFallback(granularReason, fallback string) string {
    return granularReason
  }	
  return fallback
}
```

pkg/scheduler/scheduler.go
j-skiba: https://github.com/kubernetes-sigs/kueue/pull/12452#discussion_r3465962064 fixed

pkg/scheduler/scheduler.go
j-skiba: fixed https://github.com/kubernetes-sigs/kueue/pull/12452#discussion_r3465962064

mimowo:
/lgtm
/approve
Thanks 👍

PR #12459, opened 2026-06-24
Remove the LocalQueueDefaulting feature gate which is already stable

mbobrovskyi:
#### What type of PR is this?
/kind cleanup

Remove LocalQueueDefaulting feature.

Fixes #

```release-note
NONE
```

## Summary by CodeRabbit

* **Chores**
  * Removed the `LocalQueueDefaulting` feature gate from the supported feature list.
  * Updated generated feature-gate listings and compatibility references to reflect the current set of enabled features.
* **Documentation**
  * Updated published feature-gate metadata to indicate `LocalQueueDefaulting` is removed.
* **New Features**
  * Added a Beta prerelease versioned specification for `TASMultiLayerTopology`.

### DIFF
--- a/pkg/features/kube_features.go
+++ b/pkg/features/kube_features.go
@@ -97,12 +97,6 @@ const (
 	// Enabled gathering of LocalQueue metrics
 	LocalQueueMetrics featuregate.Feature = "LocalQueueMetrics"
 
-	// owner: @yaroslava-serdiuk
-	// kep: https://github.com/kubernetes-sigs/kueue/tree/main/keps/2936-local-queue-defaulting
-	//
-	// Enable to set default LocalQueue.
-	LocalQueueDefaulting featuregate.Feature = "LocalQueueDefaulting"
-
 	// owner: @pbundyra
 	// kep: https://github.com/kubernetes-sigs/kueue/tree/main/keps/2724-topology-aware-scheduling
 	//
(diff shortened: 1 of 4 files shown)


keps/2936-local-queue-defaulting/kep.yaml
```diff
@@ -28,5 +28,4 @@ milestone:
 # The following PRR answers are required at alpha release
 # List the feature gate name and the components for which it must be enabled
 feature-gates:
-  - name: LocalQueueDefaulting
```
mimowo, line 31: Don't delete it from the KEP

mbobrovskyi: Reverted

mimowo:
/lgtm
/approve
Thank you for clearing the legacy FGs, let's try to clean more of  them 👍

mwielgus (commented):
If the PR is only about feature gate flag (not the feature itself), please name the PR/commit accordingly.

mimowo:
Good point, we are not removing the feature but a FG
/retitle Remove the LocalQueueDefaulting feature gate which is already stable

mimowo:
@mbobrovskyi looks like the build is hung on "deploy/netlifyWaiting for status to be reported — Deploy Preview processing." I guess we need to push a no-op update

Issue #12469, opened 2026-06-24
make verify fails on macOS arm64 due to Linux specific SELinux flags in verify-skills-lint


rflorenc:
**What happened**:

`make verify` fails at the `verify-skills-lint` target on macOS with:

```
  Error: preparing container ... for attach: lsetxattr .git/objects/pack/...: permission denied
  make[1]: *** [verify-skills-lint] Error 126
```
  
After getting past that error, the skillsaw container cannot write its output:

  `PermissionError: [Errno 13] Permission denied: '/out/skillsaw-summary.html'`

**What you expected to happen**:

`make skills-lint` and `make verify` should work on macOS (Darwin arm64) the same way they do on Linux.

**How to reproduce it (as minimally and precisely as possible)**:

```bash
  # On macOS with podman or docker running
  make skills-lint
  # or
  make verify
```

**Anything else we need to know?**:
NA

PR #12473, opened 2026-06-24
Fix verify-skills-lint on macOS arm64 by skipping Linux-only SELinux

rflorenc:
#### What type of PR is this?

  /kind bug

  `make verify` fails on macOS arm64 at the `verify-skills-lint` target because the container run command uses Linux only SELinux flags, ie: `:Z` volume suffix and label confinement.

This simple change extends the detection block in `Makefile-verify.mk` with two variables:
  - `VOLUME_FLAGS`: empty on macOS, `:Z` on Linux
  - `CONTAINER_SECURITY_OPTS`: `--security-opt label=disable` on macOS, empty on Linux

On Linux the container command expands to exactly what it was before.

See #12469.

  Tested on macOS Darwin arm64 with podman 5.x. 

`make skills-lint` passes with 24 skills scanned, 34 rules run, all checks passed.

  ```release-note
  NONE
  ```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container-based verification to work more reliably across different platforms, including macOS.
  * Updated mount and security handling so verification runs use the right settings automatically without manual tweaks.

### DIFF
--- a/Makefile-verify.mk
+++ b/Makefile-verify.mk
@@ -22,8 +22,13 @@ VERIFY_NPROCS ?= 8
 # Requires GNU Make 4.0+. Values: target, line, recurse, or empty.
 ifeq ($(shell uname),Darwin)
     VERIFY_OUTPUT_SYNC ?=
+    # SELinux :Z relabeling and label confinement are Linux-only.
+    VOLUME_FLAGS ?=
+    CONTAINER_SECURITY_OPTS ?= --security-opt label=disable
 else
     VERIFY_OUTPUT_SYNC ?= target
+    VOLUME_FLAGS ?= :Z
+    CONTAINER_SECURITY_OPTS ?=
 endif
 # Paths whose content is expected to be fully reproducible from sources.
 # The final step of `make verify` enforces that these paths have:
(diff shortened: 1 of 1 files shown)


mimowo:
cc @mbobrovskyi ptal

kannon92:
Maybe @amy has a macOS around to check this? 

I wasn't aware this job was failing on mac..

mbobrovskyi:
Interesting – this target works for me on the `main` branch (`macOS`, `arm64`).

```
Scanned:
  Repo type: agentskills, coderabbit
  Plugins:   0
  Skills:    24
  Rules run: 34

Summary:
  Errors:   0
  Warnings: 0

✓ All checks passed!
```

amy:
Works for me using docker. Here's what my agent said:

```
Why it works for you but fails for rflorenc
The difference isn't macOS vs. Linux — it's Docker vs. Podman, both on macOS.

Your machine uses Docker (/usr/local/bin/docker, daemon v29.3.1). The issue author tested with podman 5.x (stated in the PR). Both run containers inside a Linux VM, but the VMs differ in one critical way:

Docker Desktop's Linux VM does not enforce SELinux. So the :Z volume suffix in [Makefile-verify.mk:166](vscode-webview:///Makefile-verify.mk#L166) is silently ignored, and there's no label confinement to block the container writing to /out. Everything just works — which is exactly what you see.
Podman's VM (Fedora CoreOS) runs with SELinux enforcing. That triggers both failures in the issue
```

kannon92:
Ah nice catch. I guess this is an issue of podman versus docker on macosx.

mbobrovskyi:
Thank you!

/approve

mbobrovskyi:
Ah, we don't have skills-lint target in release-0.17.

PR #12474, opened 2026-06-24
test: drop MiniJobController in integration test

### LINKED ISSUE #11756: Drop MiniJobController in integration tests once we have k8s fix released

In order to allow the upgrade of k8s dependencies to 1.36 (see [PR](https://github.com/kubernetes-sigs/kueue/pull/10936)) we introduced the MiniJobController, because of the bug: https://github.com/kubernetes/kubernetes/issues/139281

Once the new version of K8s is released (hopefully it will be released in 1.36.2), see cherrypick: https://github.com/kubernetes/kubernetes/pull/139329

**Why is this needed**:

To avoid maintaining the hacky code.

yuluo-yx:
#### What type of PR is this?

/kind cleanup
/area testing

Fixes #11756

```release-note
None
```

## Summary by CodeRabbit

* **Tests**
  * Simplified job-related integration test setup and teardown.
  * Removed the mini job test helper used to manage job suspension behavior during test runs.

* **Chores**
  * Reduced integration harness complexity by eliminating extra lifecycle management.
  * Updated the default envtest Kubernetes version used for test asset downloads (1.36 → 1.36.2).


mimowo:
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thanks, so happy to see it dropped now

Makefile-test.mk
```diff
@@ -19,7 +19,7 @@ endif
 GO_TEST_FLAGS ?= -race
 
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION ?= 1.36
+ENVTEST_K8S_VERSION ?= 1.36.2
```
mimowo: Hm, I think here we need to keep it as 1.36 and that should select whatever is pointed by the go.mod

mimowo:
@yuluo-yx any idea why the tests are failing?

yuluo-yx:
> @yuluo-yx any idea why the tests are failing?

let me check

Makefile-test.mk
yuluo-yx: hi @mimowo 

The PR I submitted removed `MiniJobController`, but that removal relies on `envtest` using a Kubernetes version that includes the upstream fix. **That is why I initially set `ENVTEST_K8S_VERSION` to `1.36.2`.**

After reverting to `1.36`, the CI actually resolved to version `1.36.0`. Consequently, the PR's CI status reflects that the workaround has been removed, yet the test environment is still running the unfixed `1.36.0`. This causes the integration tests to lack both the `MiniJobController` and the `1.36.2` fix; as a result, the API server rejects the attempt to unsuspend the Job because the `spec.template` is being modified.

and, I agree that the `envtest` minor version should align with the Kubernetes dependency in `go.mod`, so using 1.36 is reasonable.

So, in that case, my understanding is that this cleanup should be deferred until `setup-envtest` in the CI actually selects a version that includes the Kubernetes fix.

yuluo-yx: refer: https://github.com/kubernetes-sigs/controller-tools/issues/1443

yuluo-yx:
ptal @mimowo

yuluo-yx:
Hi @mimowo, please take a look at this when you have time.

mimowo:
/lgtm
/approve

Thank you for the nice cleanup 👍  Let me try to cherrypick, that should be possible because 0.18 is already using envtest=1.36
/cherrypick release-0.18

PR #12488, opened 2026-06-25
[release-0.17] Fix flavor selector using per-flavor label keys

### LINKED ISSUE #12446: FlavorAssigner rejects a flavor for node-affinity on a label key it doesn't declare (resource-group LabelKeys is a union, but matching is per-flavor)

carterpewpew:
#### What type of PR is this?

/kind bug

Cherry-pick of #12449 to release-0.17.

`flavorSelector` was built once per resource group using the union of all flavors' label keys. This caused a flavor to be rejected when the workload referenced a label key declared by a different flavor in the same group, leaving workloads stuck pending forever. This PR computes the selector per-flavor using only that flavor's own keys.

Fixes #12446

```release-note
Scheduling: Fixed a bug where a workload could be stuck pending when its node selector referenced a label key declared by a different flavor in the same resource group.
```

### DIFF
--- a/pkg/cache/scheduler/clusterqueue.go
+++ b/pkg/cache/scheduler/clusterqueue.go
@@ -347,25 +347,17 @@ func (c *clusterQueue) isTASViolated() bool {
 // UpdateWithFlavors updates a ClusterQueue based on the passed ResourceFlavors set.
 // Exported only for testing.
 func (c *clusterQueue) UpdateWithFlavors(log logr.Logger, flavors map[kueue.ResourceFlavorReference]*kueue.ResourceFlavor) {
-	c.updateLabelKeys(flavors)
+	c.updateFlavorMetadata(flavors)
 	c.updateQueueStatus(log)
 }
 
-func (c *clusterQueue) updateLabelKeys(flavors map[kueue.ResourceFlavorReference]*kueue.ResourceFlavor) {
+func (c *clusterQueue) updateFlavorMetadata(flavors map[kueue.ResourceFlavorReference]*kueue.ResourceFlavor) {
 	c.missingFlavors = nil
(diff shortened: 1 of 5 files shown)


carterpewpew:
cc @mimowo

PR #12501, opened 2026-06-25
fix: sync ConcurrentAdmission variants on ClusterQueue flavor changes

### LINKED ISSUE #10918: [ConcurrentAdmission] Concurrentadmission Controller should react upon changes to ClusterQueue's `resourceGroups`
**What happened**:
When using ConcurrentAdmission, the concurrentadmission controller should react upon changes to ClusterQueue's `resourceGroups` and create/delete Variants adequately. It should always ensure that it has exactly one Variant per flavor.

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

ivnovakov:

The ConcurrentAdmission controller only created Variants when the parent Workload was reconciled and never deleted them, so it ignored ClusterQueue flavor-list changes at runtime. This PR makes it watch ClusterQueues and reconcile affected parents on flavor add/remove/reorder, keeping exactly one Variant per flavor; removing the admitted Variant's flavor evicts the parent so it re-admits on a remaining one.

Fixes #10918

```release-note
ConcurrentAdmission: Fixed Variants not being created or deleted when a ClusterQueue's resource flavors change.
```

## Summary by CodeRabbit

* **New Features**
  * Workload variants now automatically track the latest `ClusterQueue` flavor configuration: variants are created when flavors are added and when the desired flavor order changes.
* **Bug Fixes**
  * Stale variant workloads are deleted when their flavors are removed or no longer match the `ClusterQueue` flavor order, with improved “variant deleted” event details.
  * If the currently admitted variant flavor is removed, the parent workload is evicted and then re-admitted using an available flavor.
* **Tests**
  * Added unit and integration coverage for flavor-change detection and add/remove scenarios.

### DIFF
--- a/pkg/controller/concurrentadmission/controller.go
+++ b/pkg/controller/concurrentadmission/controller.go
@@ -28,6 +28,7 @@ import (
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/events"
+	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/clock"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -43,7 +44,8 @@ import (
 	configapi "sigs.k8s.io/kueue/apis/config/v1beta2"
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
 	qcache "sigs.k8s.io/kueue/pkg/cache/queue"
-	"sigs.k8s.io/kueue/pkg/controller/constants"
+	"sigs.k8s.io/kueue/pkg/constants"
+	controllerconsts "sigs.k8s.io/kueue/pkg/controller/constants"
 	"sigs.k8s.io/kueue/pkg/controller/core/indexer"
 	"sigs.k8s.io/kueue/pkg/controller/jobframework"
 	"sigs.k8s.io/kueue/pkg/util/roletracker"
@@ -56,6 +58,7 @@ import (
 const (
 	ConcurrentAdmissionController  = "concurrent-admission-controller"
 	ReasonCreatedVariant           = "CreatedVariant"
+	ReasonDeletedVariant           = "DeletedVariant"
 	ReasonActivatedVariant         = "ActivatedVariant"
 	ReasonDeactivatedVariant       = "DeactivatedVariant"
 	ReasonPreemptionUngatedVariant = "PreemptionUngatedVariant"
@@ -112,6 +115,18 @@ func (r *variantReconciler) setupWithManager(mgr ctrl.Manager, cfg *configapi.Co
 			}),
 			r,
 		)).
+		WatchesRawSource(source.TypedKind(
+			mgr.GetCache(),
+			&kueue.ClusterQueue{},
+			handler.TypedFuncs[*kueue.ClusterQueue, reconcile.Request]{
+				UpdateFunc: func(ctx context.Context, e event.TypedUpdateEvent[*kueue.ClusterQueue], q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+					for _, req := range r.parentsForClusterQueue(ctx, e.ObjectNew) {
+						q.AddAfter(req, constants.UpdatesBatchPeriod)
+					}
+				},
+			},
+			clusterQueueFlavorsChanged(),
+		)).
 		WithOptions(controller.Options{
 			NeedLeaderElection:      new(false),
 			MaxConcurrentReconciles: mgr.GetControllerOptions().GroupKindConcurrency[kueue.GroupVersion.WithKind("Workload").GroupKind().String()],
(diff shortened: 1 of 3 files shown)


mimowo:
cc @reruno @mbobrovskyi ptal

pkg/controller/concurrentadmission/controller.go
```diff
@@ -266,6 +318,26 @@ func (r *variantReconciler) createVariants(ctx context.Context, parent *kueue.Wo
 	return nil
 }
 
+func (r *variantReconciler) deleteStaleVariants(ctx context.Context, parent *kueue.Workload, variants []kueue.Workload, flavorOrder map[kueue.ResourceFlavorReference]int) ([]kueue.Workload, error) {
+	log := ctrl.LoggerFrom(ctx)
+	remaining := make([]kueue.Workload, 0, len(variants))
```
reruno: nit: maybe rename remaining to `activeVariants` or something similar

ivnovakov: Agree, but active collides with activate/deactivate workload, may be 'freshVariants', wdyt?

reruno: sgtm

pkg/controller/concurrentadmission/controller_test.go
```diff
@@ -2025,3 +2145,61 @@ func TestReconcile(t *testing.T) {
...
+
+	if pred.Create(event.TypedCreateEvent[*kueue.ClusterQueue]{Object: base}) {
+		t.Error("Create() = true, want false")
+	}
+	if pred.Delete(event.TypedDeleteEvent[*kueue.ClusterQueue]{Object: base}) {
+		t.Error("Delete() = true, want false")
+	}
+	if pred.Generic(event.TypedGenericEvent[*kueue.ClusterQueue]{Object: base}) {
+		t.Error("Generic() = true, want false")
+	}
```
reruno: Looks good, but maybe we could have these cases handled like 	updateCases := map[string]struct {, just for the future.

you can have one item in the maps for these create delete generic tests

ivnovakov: Done

pkg/controller/concurrentadmission/controller.go
```diff
@@ -145,14 +197,14 @@ func (r *variantReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 
 	// TODO: If ConcurrentAdmission is no longer enabled for this CQ, delete parent and variants.
 
-	if len(variants) < len(flavorOrder) {
```
reruno, line 150: Why this condition got removed?

ivnovakov: The count check only covered "too few" and had no delete path, so it couldn't converge when a flavor was removed or replaced 1:1. Now createVariants (skips existing flavors) + deleteStaleVariants (drops gone ones) converge to one variant per flavor declaratively, no count comparison needed.

pkg/controller/concurrentadmission/controller.go
```diff
@@ -266,6 +322,26 @@ func (r *variantReconciler) createVariants(ctx context.Context, parent *kueue.Wo
...
+	freshVariants := make([]kueue.Workload, 0, len(variants))
+	for i := range variants {
+		v := &variants[i]
+		flavor := concurrentadmission.GetVariantFlavor(v)
+		if _, ok := flavorOrder[flavor]; ok {
+			freshVariants = append(freshVariants, *v)
+			continue
+		}
+		log.V(2).Info("Deleting variant whose flavor is no longer in the ClusterQueue", "variant", klog.KObj(v), "flavor", flavor)
+		if _, err := workload.Delete(ctx, r.client, v); err != nil && !apierrors.IsNotFound(err) {
```
reruno, line 346: can you double check the stages of workload

ie what if the workload is in the middle of preemption or already admitted, can we just delete already admitted variant without some additional stuff that needs to be done

ivnovakov: Good question — I double-checked the lifecycle. A variant is a regular Workload, so deleting it goes through the core `WorkloadReconciler.Delete` handler (`workload_controller.go:1226`), which handles the cleanup:

- Admitted variant: `cache.DeleteWorkload` → `forgetWorkload` → `updateWorkloadUsage(..., subtract)` (`clusterqueue.go:494`+) releases its quota.
- Mid-preemption variant: `preemptionExpectations.ObservedUID(...)` (`workload_controller.go:1234`) clears its preemption expectations.
- Plus it removes the variant from the queues and re-queues anything blocked behind it.

It's the same path every regular Workload deletion takes. The `clearWorkloadAdmission` / `UnsetQuotaReservation` steps exist only on the deactivate path, which keeps the object alive — on delete that's handled by the delete handler above.

The job follows the parent, not the variant: deleting the admitted variant evicts the parent, which re-admits on a remaining flavor (covered by the `re-admits the parent on a remaining flavor` test).

reruno: What if we remove workload that has opened preemption gate, will reconciler correctly open the next one preemption gate? Can you check

Maybe we can create a test that checks this case, wdyt?

ivnovakov: Added a test for that case, it works.
The flavor removal triggers a parent reconcile, during that same reconcile the stale variant is deleted and the gate is opened for the next candidate.

pkg/controller/concurrentadmission/controller.go
```diff
@@ -119,9 +125,54 @@ func (r *variantReconciler) setupWithManager(mgr ctrl.Manager, cfg *configapi.Co
...
+		if err := r.client.List(ctx, wlList,
+			client.InNamespace(lq.Namespace),
+			client.MatchingFields{indexer.WorkloadQueueKey: lq.Name},
+			client.MatchingLabels{constants.ConcurrentAdmissionParentLabelKey: "true"},
+		); err != nil {
+			log.Error(err, "Failed to list parent Workloads for LocalQueue", "localQueue", klog.KObj(lq))
+			continue
+		}
+		for j := range wlList.Items {
+			requests = append(requests, reconcile.Request{NamespacedName: client.ObjectKeyFromObject(&wlList.Items[j])})
```
mimowo, line 175: Let's add some batching for the reconcile requests, say with 1s delay to deduplicate some of the reconicle events

pkg/controller/concurrentadmission/controller_test.go
```diff
@@ -233,6 +242,177 @@ func TestReconcile(t *testing.T) {
 				},
 			},
 		},
+		"flavor removed from ClusterQueue; variant is deleted": {
```
mimowo, line 245: Do the new unit tests cover all the new code paths (maybe except for the errors)? You can use a coverage tool

test/integration/singlecluster/controller/concurrentadmission/concurrent_admission_test.go
```diff
@@ -597,6 +596,180 @@ var _ = ginkgo.Describe("Concurrent Admission", func() {
 			})
 		})
 	})
+
+	ginkgo.When("Should react to changes in the ClusterQueue's resource flavors", func() {
```
mimowo, line 601: Let's check if the newly added intgration tests don't flake -loop then 50 times or so locally

mimowo:
LGTM overall, thank you folks for working on it  👍 
I will wait for lgtm from @reruno

reruno:
Thank you!

/lgtm

pkg/controller/concurrentadmission/controller_test.go
ivnovakov: Added `TestParentsForClusterQueue` and a test covering the already-exists create path.

pkg/controller/concurrentadmission/controller.go
ivnovakov: Done, the CQ watch now enqueues via `q.AddAfter(req, constants.UpdatesBatchPeriod)` (1s), so a burst of flavor edits collapses to one reconcile per parent.

test/integration/singlecluster/controller/concurrentadmission/concurrent_admission_test.go
ivnovakov: Yep, 200 runs locally, no failures.

mimowo (commented):
/lgtm
/approve
Thank you for this CA enhancement. Support for mutability is quite important 👍

PR #12510, opened 2026-06-25
Granular status reasons for unadmitted workloads in workload_controller

j-skiba:
#### What type of PR is this?
/kind feature

This PR implements the status condition and granular reason resolution for unadmitted workloads in the workload_controller, as part of the UnadmittedWorkloadsObservability feature gate.

Part of #10852

```release-note
Observability: Added support for the UnadmittedWorkloadsObservability feature gate in the workload controller. When enabled, Kueue populates the QuotaReserved workload condition with granular reasons (such as Misconfigured, Suspended, or AdmissionGated) and detailed messages when a workload cannot be admitted, making it easier for operators to diagnose admission issues.
```

## Summary by CodeRabbit

* **New Features**
  * Added more detailed workload status tracking so pending, blocked, and unadmitted states now surface clearer reasons and messages.
  * Expanded DRA and scheduler scenarios to show why a workload is not ready or cannot proceed.

* **Bug Fixes**
  * Improved how workload conditions are updated when queues are missing, inactive, on hold, or admission is gated.
  * Preserved more specific quota-related status reasons instead of replacing them with generic pending states.

* **Tests**
  * Added broader integration coverage for feature-gated workload behavior and condition transitions.

### DIFF
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -1838,7 +1838,7 @@ func (r *WorkloadReconciler) resolveGranularUnadmittedQuotaReservedCondition(
 		if err != nil {
 			log := ctrl.LoggerFrom(ctx)
 			log.Error(err, "Invalid ClusterQueue NamespaceSelector", "clusterQueue", cq.Name)
-			selector = labels.Nothing()
+			return r.newQuotaReservedCondition(wl, kueue.WorkloadQuotaReservedReasonMisconfigured, fmt.Sprintf("invalid namespace selector: %v", err)), nil
 		}
 		wlInfo := workload.NewInfo(wl)
 		admissibilityErr = workload.ValidateAdmissibility(ctx, r.client, wlInfo, selector)
@@ -1866,10 +1866,6 @@ func (r *WorkloadReconciler) resolveGranularUnadmittedQuotaReservedCondition(
 		return r.newQuotaReservedCondition(wl, kueue.WorkloadQuotaReservedReasonMisconfigured, admissibilityErr.Error()), nil
 	case workload.HasAdmissionGate(wl):
 		return r.newQuotaReservedCondition(wl, kueue.WorkloadAdmissionGated, fmt.Sprintf("Admission is gated by: %s", wl.Annotations[constants.AdmissionGatedByAnnotation])), nil
-	case cond != nil && cond.Status == metav1.ConditionFalse && cond.Reason == kueue.WorkloadQuotaReservedReasonNoMatchingFlavor:
-		return r.newQuotaReservedCondition(wl, cond.Reason, cond.Message), nil
-	case cond != nil && cond.Status == metav1.ConditionFalse && cond.Reason == kueue.WorkloadQuotaReservedReasonWaitingForPodsReady:
-		return r.newQuotaReservedCondition(wl, cond.Reason, cond.Message), nil
 	case cond != nil && cond.Status == metav1.ConditionFalse && cond.Reason != "" &&
 		cond.Reason != kueue.WorkloadQuotaReservedReasonSuspended &&
 		cond.Reason != kueue.WorkloadQuotaReservedReasonMisconfigured:
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -1817,7 +1817,14 @@ func (r *WorkloadReconciler) resolveUnadmittedQuotaReservedCondition(
 	cq *kueue.ClusterQueue,
 ) (*metav1.Condition, error) {
 	if features.Enabled(features.UnadmittedWorkloadsObservability) {
-		return r.resolveGranularUnadmittedQuotaReservedCondition(ctx, wl, lqExists, lqActive, cqOk, cqActive, cq)
+		reason, message, err := r.resolveGranularUnadmittedQuotaReservedCondition(ctx, wl, lqExists, lqActive, cqOk, cqActive, cq)
+		if err != nil {
+			return nil, err
+		}
+		if reason == "" {
+			return nil, nil
+		}
+		return r.newQuotaReservedCondition(wl, reason, message), nil
 	}
 	return r.resolveLegacyUnadmittedQuotaReservedCondition(wl, lqExists, lqActive, cqOk, cqActive), nil
 }
@@ -1827,7 +1834,7 @@ func (r *WorkloadReconciler) resolveGranularUnadmittedQuotaReservedCondition(
 	wl *kueue.Workload,
 	lqExists, lqActive, cqOk, cqActive bool,
 	cq *kueue.ClusterQueue,
-) (*metav1.Condition, error) {
+) (string, string, error) {
 	cond := apimeta.FindStatusCondition(wl.Status.Conditions, kueue.WorkloadQuotaReserved)
 
 	var admissibilityErr error
@@ -1838,45 +1845,46 @@ func (r *WorkloadReconciler) resolveGranularUnadmittedQuotaReservedCondition(
 		if err != nil {
 			log := ctrl.LoggerFrom(ctx)
 			log.Error(err, "Invalid ClusterQueue NamespaceSelector", "clusterQueue", cq.Name)
-			return r.newQuotaReservedCondition(wl, kueue.WorkloadQuotaReservedReasonMisconfigured, fmt.Sprintf("invalid namespace selector: %v", err)), nil
+			return kueue.WorkloadQuotaReservedReasonMisconfigured, fmt.Sprintf("invalid namespace selector: %v", err), nil
 		}
 		wlInfo := workload.NewInfo(wl)
 		admissibilityErr = workload.ValidateAdmissibility(ctx, r.client, wlInfo, selector)
 		if admissibilityErr != nil && errors.Is(admissibilityErr, workload.ErrInternal) {
-			return nil, admissibilityErr
+			return "", "", admissibilityErr
 		}
 	}
 
 	switch {
 	case !workload.IsActive(wl):
-		return r.newQuotaReservedCondition(wl, kueue.WorkloadDeactivated, "The workload is deactivated"), nil
+		return kueue.WorkloadDeactivated, "The workload is deactivated", nil
 	case workload.IsOnHold(wl):
-		return r.newQuotaReservedCondition(wl, cond.Reason, cond.Message), nil
+		return cond.Reason, cond.Message, nil
 	case !lqExists:
-		return r.newQuotaReservedCondition(wl, kueue.WorkloadQuotaReservedReasonMisconfigured, fmt.Sprintf("LocalQueue %s doesn't exist", wl.Spec.QueueName)), nil
+		return kueue.WorkloadQuotaReservedReasonMisconfigured, fmt.Sprintf("LocalQueue %s doesn't exist", wl.Spec.QueueName), nil
 	case !lqActive:
-		return r.newQuotaReservedCondition(wl, kueue.WorkloadQuotaReservedReasonSuspended, fmt.Sprintf("LocalQueue %s is inactive", wl.Spec.QueueName)), nil
+		return kueue.WorkloadQuotaReservedReasonSuspended, fmt.Sprintf("LocalQueue %s is inactive", wl.Spec.QueueName), nil
 	case !cqOk:
 		cqName, _ := r.queues.ClusterQueueForWorkload(wl)
-		return r.newQuotaReservedCondition(wl, kueue.WorkloadQuotaReservedReasonMisconfigured, fmt.Sprintf("ClusterQueue %s doesn't exist", cqName)), nil
+		return kueue.WorkloadQuotaReservedReasonMisconfigured, fmt.Sprintf("ClusterQueue %s doesn't exist", cqName), nil
 	case !cqActive:
 		cqName, _ := r.queues.ClusterQueueForWorkload(wl)
-		return r.newQuotaReservedCondition(wl, kueue.WorkloadQuotaReservedReasonSuspended, fmt.Sprintf("ClusterQueue %s is inactive", cqName)), nil
+		return kueue.WorkloadQuotaReservedReasonSuspended, fmt.Sprintf("ClusterQueue %s is inactive", cqName), nil
 	case admissibilityErr != nil:
-		return r.newQuotaReservedCondition(wl, kueue.WorkloadQuotaReservedReasonMisconfigured, admissibilityErr.Error()), nil
+		return kueue.WorkloadQuotaReservedReasonMisconfigured, admissibilityErr.Error(), nil
 	case workload.HasAdmissionGate(wl):
-		return r.newQuotaReservedCondition(wl, kueue.WorkloadAdmissionGated, fmt.Sprintf("Admission is gated by: %s", wl.Annotations[constants.AdmissionGatedByAnnotation])), nil
+		return kueue.WorkloadAdmissionGated, fmt.Sprintf("Admission is gated by: %s", wl.Annotations[constants.AdmissionGatedByAnnotation]), nil
 	case cond != nil && cond.Status == metav1.ConditionFalse && cond.Reason != "" &&
 		cond.Reason != kueue.WorkloadQuotaReservedReasonSuspended &&
-		cond.Reason != kueue.WorkloadQuotaReservedReasonMisconfigured:
-		return r.newQuotaReservedCondition(wl, cond.Reason, cond.Message), nil
+		cond.Reason != kueue.WorkloadQuotaReservedReasonMisconfigured &&
+		cond.Reason != kueue.WorkloadDeactivated:
+		return cond.Reason, cond.Message, nil
 	default:
 		// Stamping the condition on creation is deferred to the next feature gate:
 		// UnadmittedWorkloadsObservabilityExplicitStatus.
 		if cond == nil {
-			return nil, nil
+			return "", "", nil
 		}
-		return r.newQuotaReservedCondition(wl, kueue.WorkloadQuotaReservedReasonPendingEvaluation, "Workload is pending evaluation in the scheduling queue"), nil
+		return kueue.WorkloadQuotaReservedReasonPendingEvaluation, "Workload is pending evaluation in the scheduling queue", nil
 	}
 }
 
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -667,27 +667,20 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 	if !workload.IsAdmitted(&wl) {
 		var updated bool
 		err := workloadpatching.PatchAdmissionStatus(ctx, r.client, &wl, r.clock, func(wl *kueue.Workload) (bool, error) {
-			if features.Enabled(features.UnadmittedWorkloadsObservability) {
-				if !workload.HasQuotaReservation(wl) {
-					cqActive := cqOk && r.cache.ClusterQueueActive(cqName)
-					cond, err := r.resolveUnadmittedQuotaReservedCondition(ctx, wl, lqExists, lqActive, cqOk, cqActive, cq)
-					if err != nil {
-						return false, err
-					}
-					if cond != nil {
-						if apimeta.SetStatusCondition(&wl.Status.Conditions, *cond) {
-							updated = true
-						}
-					}
-					if wl.Status.Admission != nil {
-						wl.Status.Admission = nil
-						updated = true
-					}
+			if !workload.HasQuotaReservation(wl) {
+				cqActive := cqOk && r.cache.ClusterQueueActive(cqName)
+				var cond *metav1.Condition
+				var err error
+				if features.Enabled(features.UnadmittedWorkloadsObservability) {
+					cond, err = r.resolveUnadmittedQuotaReservedCondition(ctx, wl, lqExists, lqActive, cqOk, cqActive, cq)
+				} else {
+					cond = r.resolveLegacyUnadmittedQuotaReservedCondition(wl, lqExists, lqActive, cqOk, cqActive)
 				}
-				if workload.HasRejectedChecks(wl) {
-					rejectedChecks := workload.RejectedChecks(wl)
-					message := buildAdmissionChecksMessage(rejectedChecks, kueue.CheckStateRejected)
-					if workload.SetDeactivationTarget(wl, kueue.WorkloadEvictedByAdmissionCheck, message) {
+				if err != nil {
+					return false, err
+				}
+				if cond != nil {
+					if apimeta.SetStatusCondition(&wl.Status.Conditions, *cond) {
 						updated = true
 					}
 				}
@@ -699,12 +692,6 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		})
 		if err != nil {
 			return ctrl.Result{}, client.IgnoreNotFound(err)
-		}
-		// If the observability feature gate is enabled and the workload has no quota reservation,
-		// we have already resolved and patched the granular conditions (e.g., missing or inactive
-		// queue) in the block above.
-		if features.Enabled(features.UnadmittedWorkloadsObservability) && !workload.HasQuotaReservation(&wl) {
-			return ctrl.Result{}, nil
 		}
 		isAdmitted := workload.IsAdmitted(&wl)
 		if isAdmitted {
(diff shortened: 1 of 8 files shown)


pkg/controller/core/workload_controller.go
```diff
@@ -1786,3 +1838,61 @@ func (h *deviceClassHandler) reconcileWorkloads(ctx context.Context, q workqueue
...
+		newReason = kueue.WorkloadQuotaReservedReasonMisconfigured
+		newMsg = fmt.Sprintf("ClusterQueue %s doesn't exist", cqName)
+	case cond != nil && cond.Status == metav1.ConditionFalse && cond.Reason == kueue.WorkloadQuotaReservedReasonNoMatchingFlavor:
+		newReason = cond.Reason
+		newMsg = cond.Message
+	case !cqActive:
+		cqName, _ := r.queues.ClusterQueueForWorkload(wl)
+		newReason = kueue.WorkloadQuotaReservedReasonSuspended
+		newMsg = fmt.Sprintf("ClusterQueue %s is inactive", cqName)
+	case features.Enabled(features.AdmissionGatedBy) && workload.HasAdmissionGate(wl):
```
kshalot: AFAICR `HasAdmissionGate` already contains the FG check

j-skiba: you're right, removed the extra check

pkg/cache/queue/manager.go
```diff
@@ -903,10 +903,13 @@ func (m *Manager) QueueSecondPassIfNeeded(ctx context.Context, w *kueue.Workload
 	log := ctrl.LoggerFrom(ctx)
 	wlKey := workload.Key(w)
 	if workload.NeedsSecondPass(w) {
+		if !m.secondPassQueue.prequeueIfNotPresent(w) {
+			log.V(4).Info("Workload is already in the second pass queue, skipping re-queue", "workload", wlKey)
+			return false
+		}
```
j-skiba: conditions updates were triggering this and some of fairsharing integration tests were becoming flaky. I can extract this to a separate PR.

mimowo: What was the problem with the previous code? Depending on that: 
1.  if there is a bug which just wasn't exposed by the preexisting test then we want separate PR and cherrypick the bugfix.
2. if there is no bug, but somehow it is helping to write unit tests - then I would probably not change it at all
3. there was no bug (or we don't know), but due to the recent changes there can be due to new code paths explored ,  then we merge as separete PR so that we have flexibility to cherry pick later
4. ???

pkg/controller/core/workload_controller.go
```diff
@@ -656,13 +661,47 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+					rejectedChecks := workload.RejectedChecks(wl)
+					message := buildAdmissionChecksMessage(rejectedChecks, kueue.CheckStateRejected)
+					if workload.SetDeactivationTarget(wl, kueue.WorkloadEvictedByAdmissionCheck, message) {
+						updated = true
+					}
+				}
+			}
+			if workload.SyncAdmittedCondition(wl, r.clock.Now()) {
+				updated = true
+			}
```
mimowo: I would like to first understand why flow changes are introduced to this file, because it looks more than just more granular conditions. In particular I'm wondering if we could de-couple the preparetory flow changes to the separete PR, independent of the new FG. Then, in the next PR just introduce the more granular conditions. Would that be feasible?

mimowo (commented):
For now I just scanned the code. I think we should consider splitting the PR to de-couple flow changes from the "granular status reasons for unadmitted workloads". In my opinion currently the PR does much more than sold in the title. We can consider the preparatory flow changes to be independent of the. FG, but I need to first understand the risks better

pkg/controller/core/workload_controller.go
j-skiba: that whole branch in `PatchAdmissionStatus`
```go
  ...
}
```
was introduced because the legacy reason resolving doesn't follow the precedence rules we have defined in the KEP.

We could split that in two PRs:
- replace all legacy reasons references in `workload_controller.go` with `workload.UnadmittedWorkloadReasonWithFallback`. It depends if we would like to update unit tests too if yes then we would have to update them in a way that will break reason precedence rules from the KEP.
- add that `if features.Enabled(features.UnadmittedWorkloadsObservability)` branch there and depending on the previous decision about tests update them accordingly.

mimowo: Cc @kshalot ptal

pkg/cache/queue/manager.go
j-skiba: alright, I was trying to reproduce the flakiness I encountered before but I cannot. I reverted the change and rerun integration tests 15 times and don't see the flake no more. I think it might have been some other thing I added along the way I was implementing the feature. Sorry for introducing confusion.

the code has been removed.

j-skiba:
/retest 

it seems like python dependency issue

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12510/pull-kueue-test-e2e-extended-shard-2-main-1-34/2070478448626765824

amy:
@j-skiba can you use the release note skill: https://github.com/kubernetes-sigs/kueue/blob/main/cmd/experimental/skills/kueue-release-notes/SKILL.md

If you did, ignore this comment.

j-skiba:
> @j-skiba can you use the release note skill: https://github.com/kubernetes-sigs/kueue/blob/main/cmd/experimental/skills/kueue-release-notes/SKILL.md
> 
> If you did, ignore this comment.

I updated the release note as requested

pkg/controller/core/workload_controller.go
mimowo: Yes, I think it is better to split. I looked again, and the new code inside `PatchAdmissionStatus` seems to do more than just "precedence rules". IIUC presedence rules is done inside `resolveUnadmittedQuotaReservedCondition`, but aside from that we have:

```golang
					if wl.Status.Admission != nil {
						wl.Status.Admission = nil
						updated = true
					}
				}
				if workload.HasRejectedChecks(wl) {
					rejectedChecks := workload.RejectedChecks(wl)
					message := buildAdmissionChecksMessage(rejectedChecks, kueue.CheckStateRejected)
					if workload.SetDeactivationTarget(wl, kueue.WorkloadEvictedByAdmissionCheck, message) {
						updated = true
					}
				}
```
which is my prerception so far is doing something else, something that warrants a comment message at least.

Additionally, I'm wondering if all of that we could wipe under `SyncAdmittedCondition`,wdyt?

Maybe we just need to rename that to something like `SyncStatusAndConditions`?

pkg/controller/core/workload_controller.go
```diff
@@ -1787,3 +1846,69 @@ func (h *deviceClassHandler) reconcileWorkloads(ctx context.Context, q workqueue
 		}
 	}
 }
+
+func (r *WorkloadReconciler) resolveUnadmittedQuotaReservedCondition(wl *kueue.Workload, lqExists, lqActive, cqOk, cqActive bool) *metav1.Condition {
+	cond := apimeta.FindStatusCondition(wl.Status.Conditions, kueue.WorkloadQuotaReserved)
+
+	var newReason, newMsg string
```
mimowo: instead of that, let's introduce a helper where you just `return` from the different branches rather than mutating the local variables. Mutability proves to be harder to follow and lead to unexpected interactions as the code evolves.

pkg/controller/core/workload_controller.go
```diff
@@ -1787,3 +1846,69 @@ func (h *deviceClassHandler) reconcileWorkloads(ctx context.Context, q workqueue
...
+		newMsg = fmt.Sprintf(clusterQueueIsInactiveMsgFormat, cqName)
+	case workload.HasAdmissionGate(wl):
+		newReason = kueue.WorkloadAdmissionGated
+		newMsg = fmt.Sprintf("Admission is gated by: %s", wl.Annotations[constants.AdmissionGatedByAnnotation])
+	case cond != nil && cond.Status == metav1.ConditionFalse && cond.Reason == kueue.WorkloadQuotaReservedReasonWaitingForPodsReady:
+		newReason = cond.Reason
+		newMsg = cond.Message
+	case cond != nil && cond.Status == metav1.ConditionFalse && cond.Reason == kueue.WorkloadQuotaReservedReasonMisconfigured:
+		cqName, _ := r.queues.ClusterQueueForWorkload(wl)
+		if cond.Message == fmt.Sprintf(localQueueDoesNotExistMsgFormat, wl.Spec.QueueName) ||
```
kshalot: Two points:
1. In principle, I think using human-readable messages for logic is brittle.
2. What happens when someone resolves the misconfiguration here by changing the LQ/CQ name? We now know that `lqExists == true` because of the previous switch statement, so we'll fall here for "recovery". But the condition never matches, so we remain misconfigured forever? We don't have a test for this case, we only have one for the case where the LQ/CQ is created after it didn't exist, not one for the case where the lq name is changed in the wl. This highlights point number 1.

j-skiba: I agree.

I think we could run the same code that we currently run in scheduler

https://github.com/kubernetes-sigs/kueue/blob/d205c10a8a1a76068c6cc9b2146697be7e4b9fc5/pkg/scheduler/scheduler.go#L632

so we don't have to deal with special handling of a reason in workload_controller that was resolved in the scheduler.

We could have a function like this 

https://github.com/kubernetes-sigs/kueue/pull/12659/changes#diff-d6b37b9e32fc2e214d8cb094a9c9f1607df91930402addb50f8691f888888843R193

and reuse the code in both files.

pkg/controller/core/workload_controller.go
```diff
@@ -663,10 +666,31 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+			if !workload.HasQuotaReservation(wl) {
+				cqActive := cqOk && r.cache.ClusterQueueActive(cqName)
+				var cond *metav1.Condition
+				var err error
+				if features.Enabled(features.UnadmittedWorkloadsObservability) {
+					cond, err = r.resolveUnadmittedQuotaReservedCondition(ctx, wl, lqExists, lqActive, cqOk, cqActive, cq)
+				} else {
+					cond = r.resolveLegacyUnadmittedQuotaReservedCondition(wl, lqExists, lqActive, cqOk, cqActive)
+				}
+				if err != nil {
```
mimowo: Let's wrap than into single function call `resolveUnadmittedQuotaReservedCondition`, than could dispatch the logic to `resolveGranularUnadmittedQuotaReservedCondition` and `resolveLegacyUnadmittedQuotaReservedCondition`

pkg/controller/core/workload_controller.go
```diff
@@ -663,10 +666,31 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+					return false, err
+				}
+				if cond != nil {
+					if apimeta.SetStatusCondition(&wl.Status.Conditions, *cond) {
+						updated = true
+					}
+				}
```
mimowo: I would rather wrap the entarity of this into `syncQuotaReservedFalseCondition` which already calls `apimeta.SetStatusCondition` inside, and returns bool indicating if it made changes or not - analogously to `SyncAdmittedCondition`. Then the new function could call `resolveUnadmittedQuotaReservedCondition` internally.

I think this is needed to offload the logic inside the lambda function.

mimowo (commented):
LGTM overall, left some nit readaiblity comments mostly, but another pass from @kshalot would be welcome

pkg/controller/core/workload_controller.go
j-skiba: wrapped into 

```go
func (r *WorkloadReconciler) resolveUnadmittedQuotaReservedCondition(
	ctx context.Context,
	wl *kueue.Workload,
	lqExists, lqActive, cqOk, cqActive bool,
	cq *kueue.ClusterQueue,
) (*metav1.Condition, error) {
		return r.resolveGranularUnadmittedQuotaReservedCondition(ctx, wl, lqExists, lqActive, cqOk, cqActive, cq)
	}
	return r.resolveLegacyUnadmittedQuotaReservedCondition(wl, lqExists, lqActive, cqOk, cqActive), nil
}
```
and moved into another function `syncQuotaReservedFalseCondition` as advised in the next comment

pkg/controller/core/workload_controller.go
j-skiba: wrapped into `syncQuotaReservedFalseCondition` which calls `resolveUnadmittedQuotaReservedCondition` and `apimeta.SetStatusCondition` internally

pkg/controller/core/workload_controller.go
j-skiba: the code has been updated as described above

pkg/controller/core/workload_controller.go
j-skiba: done

pkg/controller/core/workload_controller.go
j-skiba: A new PR has been extracted ([#12659](https://github.com/kubernetes-sigs/kueue/pull/12659)) and merged already.

Some refactoring was also done in these two discussion threads:

- https://github.com/kubernetes-sigs/kueue/pull/12510#discussion_r3511628231

- https://github.com/kubernetes-sigs/kueue/pull/12510#discussion_r3506496284

also the code has changed slightly since those comments were originally posted.

j-skiba:
@coderabbitai let's update PR description and do another round of review

pkg/controller/core/workload_controller.go
```diff
@@ -1814,3 +1793,121 @@ func (h *deviceClassHandler) reconcileWorkloads(ctx context.Context, q workqueue
...
+	var admissibilityErr error
+	if cqOk && cq != nil {
+		var selector labels.Selector
+		var err error
+		selector, err = metav1.LabelSelectorAsSelector(cq.Spec.NamespaceSelector)
+		if err != nil {
+			log := ctrl.LoggerFrom(ctx)
+			log.Error(err, "Invalid ClusterQueue NamespaceSelector", "clusterQueue", cq.Name)
+			selector = labels.Nothing()
+		}
```
mimowo, line 1849: Hm, why doing this differently than in scheduler? In scheduler we run `ValidateAdmissibility` and then pdepending on the error we handle - in this case we should use MisConfigured rather than silently ignore the error. IIUC the error will fail anyway later in scheduler, so we prefer fail fast

pkg/controller/core/workload_controller.go
```diff
@@ -642,10 +644,11 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 	}
 
 	cqName, cqOk := r.queues.ClusterQueueForWorkload(&wl)
+	var cq *kueue.ClusterQueue
```
mimowo, line 647: Why is that changed? I mean the change is not necessarily wrong, but seems to expand the diff for no reason.

pkg/controller/core/workload_controller.go
```diff
@@ -1814,3 +1793,121 @@ func (h *deviceClassHandler) reconcileWorkloads(ctx context.Context, q workqueue
...
+	case workload.HasAdmissionGate(wl):
+		return r.newQuotaReservedCondition(wl, kueue.WorkloadAdmissionGated, fmt.Sprintf("Admission is gated by: %s", wl.Annotations[constants.AdmissionGatedByAnnotation])), nil
+	case cond != nil && cond.Status == metav1.ConditionFalse && cond.Reason == kueue.WorkloadQuotaReservedReasonNoMatchingFlavor:
+		return r.newQuotaReservedCondition(wl, cond.Reason, cond.Message), nil
+	case cond != nil && cond.Status == metav1.ConditionFalse && cond.Reason == kueue.WorkloadQuotaReservedReasonWaitingForPodsReady:
+		return r.newQuotaReservedCondition(wl, cond.Reason, cond.Message), nil
+	case cond != nil && cond.Status == metav1.ConditionFalse && cond.Reason != "" &&
+		cond.Reason != kueue.WorkloadQuotaReservedReasonSuspended &&
+		cond.Reason != kueue.WorkloadQuotaReservedReasonMisconfigured:
+		return r.newQuotaReservedCondition(wl, cond.Reason, cond.Message), nil
```
pajakd: Out of these 3 cases, would anything change if we keep only the last one?

pkg/controller/core/workload_controller.go
j-skiba: actually it's kind of the same here and in scheduler. Scheduler uses CQ from snapshot which has been parsed already and here we operate on a fresh object.

but I agree that we should set `Misconfigured` in some place in such case, it's a little similar to misconfiguration of LQ<->CQ pointers which we already treat as `Misconfigured`. Possibly it might be the only place where we can resolve this. I will update the code

pkg/controller/core/workload_controller.go
```diff
@@ -1814,3 +1793,121 @@ func (h *deviceClassHandler) reconcileWorkloads(ctx context.Context, q workqueue
...
+		wlInfo := workload.NewInfo(wl)
+		admissibilityErr = workload.ValidateAdmissibility(ctx, r.client, wlInfo, selector)
+		if admissibilityErr != nil && errors.Is(admissibilityErr, workload.ErrInternal) {
+			return nil, admissibilityErr
+		}
+	}
+
+	switch {
+	case !workload.IsActive(wl):
+		return r.newQuotaReservedCondition(wl, kueue.WorkloadDeactivated, "The workload is deactivated"), nil
```
kshalot: Nit: I think this `r.newQuotaReservedCondition(wl...)` in each case statement is adding a bunch of visual noise. We always do `case: return`

Wouldn't it be easier to follow this function by introducing variables like `reason` and `message` and assigning values in the branches? Then it's easier to reason about what this function returns, because there is only one "happy path" return. But this is just a hunch.

test/util/util.go
```diff
@@ -477,6 +477,18 @@ func ExpectWorkloadsToBePending(ctx context.Context, k8sClient client.Client, wl
 	ExpectWorkloadsToBePendingByKeys(ctx, k8sClient, wlKeys...)
 }
 
+var pendingQuotaReservedReasons = sets.New(
+	"Pending",
+	"Waiting",
```
kshalot: AFAIR you added constants for this in the prep PR.

test/integration/singlecluster/controller/dra/dra_pd_test.go
```diff
@@ -584,5 +584,62 @@ var _ = ginkgo.Describe("DRA Partitionable Devices Integration", ginkgo.Ordered,
...
+
+				g.Expect(updatedWl.Status.Conditions).To(gomega.ContainElement(gomega.And(
+					gomega.HaveField("Type", kueue.WorkloadQuotaReserved),
+					gomega.HaveField("Status", metav1.ConditionFalse),
+					gomega.HaveField("Reason", kueue.WorkloadQuotaReservedReasonMisconfigured),
+				)))
+				g.Expect(updatedWl.Status.Conditions).To(gomega.ContainElement(gomega.And(
+					gomega.HaveField("Type", kueue.WorkloadAdmitted),
+					gomega.HaveField("Status", metav1.ConditionFalse),
+					gomega.HaveField("Reason", "NoReservation"),
```
kshalot:
```suggestion
					gomega.HaveField("Reason", WorkloadAdmittedReasonNoReservation),
```

I think we have this as well

pkg/controller/core/workload_controller.go
j-skiba: previously `cq` was used just in the if branch below, now it's used outside of that if too. Also we use only pointers everywhere where the var is used.

pkg/controller/core/workload_controller.go
j-skiba: updated

pkg/controller/core/workload_controller.go
j-skiba: yeah, it doesn't have to be in 3 separate branches
I merged them into one

pkg/controller/core/workload_controller.go
```diff
@@ -1814,3 +1793,117 @@ func (h *deviceClassHandler) reconcileWorkloads(ctx context.Context, q workqueue
...
+		cqName, _ := r.queues.ClusterQueueForWorkload(wl)
+		return r.newQuotaReservedCondition(wl, kueue.WorkloadQuotaReservedReasonSuspended, fmt.Sprintf("ClusterQueue %s is inactive", cqName)), nil
+	case admissibilityErr != nil:
+		return r.newQuotaReservedCondition(wl, kueue.WorkloadQuotaReservedReasonMisconfigured, admissibilityErr.Error()), nil
+	case workload.HasAdmissionGate(wl):
+		return r.newQuotaReservedCondition(wl, kueue.WorkloadAdmissionGated, fmt.Sprintf("Admission is gated by: %s", wl.Annotations[constants.AdmissionGatedByAnnotation])), nil
+	case cond != nil && cond.Status == metav1.ConditionFalse && cond.Reason != "" &&
+		cond.Reason != kueue.WorkloadQuotaReservedReasonSuspended &&
+		cond.Reason != kueue.WorkloadQuotaReservedReasonMisconfigured:
+		return r.newQuotaReservedCondition(wl, cond.Reason, cond.Message), nil
```
kshalot: How about Deactivated? Or `AdmissionGated` is better, because it's already handled here:

1. We have the admission gated annotation.
2. We fall into `case workload.HasAdmissionGate(wl)` which sets the admission gated reason.
3. We remove the admission gate.
4. We fall into this clause because `cond.Reason != ""`, which copies the `AdmissionGated` reason again.

Will that make the workload stuck forever? Or will the scheduler overwrite the condition? I'm not sure. But I think we should cleanly transition to pending. WDYT?

test/integration/singlecluster/controller/core/workload_controller_test.go
```diff
@@ -646,6 +646,238 @@ var _ = ginkgo.Describe("Workload controller", ginkgo.Label("controller:workload
...
+					Status:  metav1.ConditionFalse,
+					Reason:  "NoReservation",
+					Message: "The workload has no reservation",
+				},
+			}
+			for _, wantCond := range wantConditions {
+				cond := apimeta.FindStatusCondition(updatedWl.Status.Conditions, wantCond.Type)
+				gomega.Expect(cond).NotTo(gomega.BeNil())
+				gomega.Expect(*cond).To(gomega.BeComparableTo(wantCond, util.IgnoreConditionTimestampsAndObservedGeneration))
+			}
```
kshalot: Maybe we can introduce some `ExpectConditions(conditions)` helper, because this logic is duplicated in the new tests. WDYT?

pkg/controller/core/workload_controller.go
j-skiba: Clearing of reason for `AdmissionGated` case is done in `mayUpdateConditionForAdmissionGatedBy`

but for `Deactivated` you're right, we should add `cond.Reason != Deactivated` check to this list so it goes to default branch.

j-skiba: done,

- updated the switch case to:
```go
case cond != nil && cond.Status == metav1.ConditionFalse && cond.Reason != "" &&
		cond.Reason != kueue.WorkloadQuotaReservedReasonSuspended &&
		cond.Reason != kueue.WorkloadQuotaReservedReasonMisconfigured &&
		cond.Reason != kueue.WorkloadDeactivated:
```
- added two unit test cases for that: `workload that is deactivated should set QuotaReserved condition to Deactivated (observability enabled)` and `workload that is reactivated should clear Deactivated condition and transition to PendingEvaluation (observability enabled)`, we have `workload with AdmissionGatedBy annotation removed should clear the gate and emit event` already for `AdmissionGated` already in the code

test/integration/singlecluster/controller/core/workload_controller_test.go
j-skiba: good idea, done

test/integration/singlecluster/controller/dra/dra_pd_test.go
j-skiba: done

test/util/util.go
j-skiba: replaced

pkg/controller/core/workload_controller.go
j-skiba: yeah, that makes sense. I made it return `(string, string, err)` and then these are consumed to create a condition

test/integration/singlecluster/controller/core/workload_controller_test.go
```diff
@@ -646,6 +646,183 @@ var _ = ginkgo.Describe("Workload controller", ginkgo.Label("controller:workload
...
+					Message: "The workload has no reservation",
+				},
+			)
+		})
+
+		ginkgo.It("should write granular conditions and reasons when clusterQueue is missing", func() {
+			lq := utiltestingapi.MakeLocalQueue("missing-cq-q", ns.Name).ClusterQueue("non-existent-cq").Obj()
+			gomega.Expect(k8sClient.Create(ctx, lq)).To(gomega.Succeed())
+
+			// Wait for LocalQueue to be reconciled and registered in the Queue Manager
```
kshalot: Instead of using comments, maybe we could group certain actions with `ginkgo.By`?  Makes the output of the tests clearer

j-skiba:
@kshalot @mimowo thank you for your reviews

I ran (multiple times) review using the kueue-review skill and found a few issues covered in these commits:

- https://github.com/kubernetes-sigs/kueue/pull/12510/commits/c10e1602c399b4139a0c3436e4749571a10071c2
- https://github.com/kubernetes-sigs/kueue/pull/12510/commits/a1a3a9ed7a9028c61f7e7a7fea6012adeb5ff812
- https://github.com/kubernetes-sigs/kueue/pull/12510/commits/54a7e259e6e6144b8e1d24dd0979acf92ba484ef

along with some cleanups and additional tests

test/integration/singlecluster/controller/core/workload_controller_test.go
j-skiba: sounds good, done

j-skiba:
> @kshalot @mimowo thank you for your reviews
> 
> I ran (multiple times) review using the kueue-review skill and found a few issues covered in these commits:
> 
> * [c10e160](https://github.com/kubernetes-sigs/kueue/commit/c10e1602c399b4139a0c3436e4749571a10071c2)
> * [a1a3a9e](https://github.com/kubernetes-sigs/kueue/commit/a1a3a9ed7a9028c61f7e7a7fea6012adeb5ff812)
> * [54a7e25](https://github.com/kubernetes-sigs/kueue/commit/54a7e259e6e6144b8e1d24dd0979acf92ba484ef)
> 
> along with some cleanups and additional tests

which makes me also think if conderabbit uses that skill?

@coderabbitai do you use that?

pkg/controller/core/workload_controller.go
mimowo: I see, makes sense

mimowo (commented):
/lgtm
/approve
Thank you 👍

PR #12514, opened 2026-06-26
Remove features.* is deprecated exclusion.

mbobrovskyi:
#### What type of PR is this?
/kind cleanup
/area testing

Remove features.* is deprecated exclusion.

Fixes #

```release-note
NONE
```

## Summary by CodeRabbit

* **Chores**
  * Updated code quality checks to better match current project usage.
  * Refined a few warning exclusions for deprecated patterns and test-related context issues.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -89,9 +89,6 @@ linters:
         text: 'SA1019: j.Status.RayClusterStatus.State is deprecated: the State field is replaced by the Conditions field.'
       - linters:
           - staticcheck
-        text: 'SA1019: features.* is deprecated: planned to be removed in .*'
-      - linters:
-          - staticcheck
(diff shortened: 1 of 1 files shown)


.golangci.yaml
```diff
@@ -87,9 +87,6 @@ linters:
       - linters:
           - staticcheck
         text: 'SA1019: j.Status.RayClusterStatus.State is deprecated: the State field is replaced by the Conditions field.'
-      - linters:
-          - staticcheck
-        text: 'SA1019: features.* is deprecated: planned to be removed in .*'
```
mimowo, line 92: I'm not sure exactly what is the benefit of that, or what this was introduced for. Could you explain? 

Also, what should we be doing when we want to deprecate so feature gate after this is merged?

mbobrovskyi: I think you're right. We should keep this exclusion.

PR #12515, opened 2026-06-26
Fix scheme.Builder is deprecated warning.

mbobrovskyi:
#### What type of PR is this?
/kind cleanup

Fix scheme.Builder is deprecated warning.

Fixes #

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated API group/version handling across the app so generated resources, admission checks, and controllers use the same consistent API identifiers.
  * Fixed several tests and runtime paths to match the new API version format.

* **Refactor**
  * Simplified API registration setup for multiple resources and removed redundant per-file initialization hooks.
  * Improved concurrency-key lookup consistency for controllers and scheduling components.


apis/config/v1beta1/groupversion_info.go
```diff
@@ -20,29 +20,30 @@ limitations under the License.
 package v1beta1
 
 import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"sigs.k8s.io/controller-runtime/pkg/scheme"
 )
 
+const GroupName = "config.kueue.x-k8s.io"
```
tenzen-y: Why do we need to have this exposed group name as a constant?

mbobrovskyi: I found it in this comment: https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/scheme/scheme.go#L62-L91. It looks like this is a common pattern. But I can revert it if you think it's not needed.

tenzen-y: The main difference between this and the controller-runtime example is the package.
The controller runtime one is managed under scheme package, but Kueue one is managed under api package.

All things under the api package are respected with strict lifecycle management as much as possible, even if they aren't directly tied into CRD like constant.

So, if we expose this constant, I'd like to have another PR. In other words, I don't have objections to expose this group name, but I think that we should separate the change from this PR at least.

mbobrovskyi: Reverted.

apis/config/v1beta1/groupversion_info.go
```diff
@@ -20,27 +20,33 @@ limitations under the License.
...
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"sigs.k8s.io/controller-runtime/pkg/scheme"
 )
 
 var (
-	// GroupVersion is group version used to register these objects
-	GroupVersion = schema.GroupVersion{Group: "config.kueue.x-k8s.io", Version: "v1beta1"}
+	// SchemeGroupVersion is a group version used to register these objects
+	SchemeGroupVersion = schema.GroupVersion{Group: "config.kueue.x-k8s.io", Version: "v1beta1"}
```
mimowo: I understand `scheme.Builder{` is deprecated and replaced with `runtime.NewSchemeBuilder`, but why this variable is getting renamed?

That is probably ok, but I would like to understand why, because it means a breaking change (not compiling code) for users of Kueue API programmatically.  

It also seems to bloat the diff for an unclear reason.

mbobrovskyi: In the example from controller-runtime (https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/scheme/scheme.go#L62-L91), I don't see a GroupVersion variable being used with runtime.NewSchemeBuilder. I renamed it to SchemeGroupVersion for consistency with that example.

I also noticed that we already have an alias here:

https://github.com/kubernetes-sigs/kueue/blob/f1aeeb4150b0542d65ac3c8ab735282bb0a3756a/apis/kueue/v1alpha1/groupversion_info.go#L28-L33

So I simply removed GroupVersion and renamed it to SchemeGroupVersion to keep it consistent with the rest of the codebase and don't duplicate variables.

mbobrovskyi: If you think we should keep both variables, I can revert this change. It doesn't make any functional difference, but I'm not sure why we need to keep both variables.

Alternatively, I can move this change into a separate PR.

mimowo: I see, avoiding the alias might be a good idea, but I cannot see how is that related to the deprecation of the builder, so yes, I suggest separate PR.

Moreover I think the other PR should have a release note because it will be a breaking change for people who compile their integrations using Kueue .

mbobrovskyi: Created https://github.com/kubernetes-sigs/kueue/pull/12738.

mbobrovskyi: Done

PR #10971, opened 2026-05-05
Migrate to controller-runtime v0.23+ Event Recorder API

vladikkuzn:
## Description

This PR migrates kueue to use the new Event Recorder API introduced in controller-runtime v0.23.

## Background

controller-runtime v0.23 migrates event recording toward the newer `events.k8s.io` API. As part of this migration, `mgr.GetEventRecorderFor(...)` is deprecated in favor of `mgr.GetEventRecorder(...)`.

### API Changes
- Replace `mgr.GetEventRecorderFor(...)` with `mgr.GetEventRecorder(...)`
- Update `Recorder` field type from `record.EventRecorder` to `events.EventRecorder`
- Update imports from `k8s.io/client-go/tools/record` to `k8s.io/client-go/tools/events`

### Event Call Updates
- Update `Event` and `Eventf` calls to new signature:
  - Old: `Event(obj, eventType, reason, message)`
  - New: `Eventf(regarding, related, eventtype, reason, action, note, args...)`

### RBAC Updates
- Update RBAC comments to use `events.k8s.io` API group:
  - Old: `// +kubebuilder:rbac:groups="",resources=events`
  - New: `// +kubebuilder:rbac:groups=events.k8s.io,resources=events`

### Test Updates
- Update test files to use `&utiltesting.EventRecorder{}` instead of `record.NewBroadcaster().NewRecorder()`

### Linter Configuration
- Remove `GetEventRecorderFor` deprecation ignore rule from `.golangci.yaml`

## Testing

All unit tests pass:
```
go test ./pkg/... ./cmd/... -short  # PASS
go build ./...  # PASS
go vet ./...  # PASS
```

## Related Issues

Fixes: https://github.com/kubernetes-sigs/kueue/issues/1481

## Release Note

```release-note
Observability: Migrate from the old event types `core/v1` to use `events.k8s.io/v1`, by replacing the
helper used in controller-runtime.

ACTION REQUIRED: if you compile your own "in house" integration then you need to adjust the code to replace 
GetEventRecorderFor to GetEventRecorder when obtaining the event recorder.
```

PR #12516, opened 2026-06-26
Remove mgr.GetEventRecorderFor is deprecated exclusion.

mbobrovskyi:
#### What type of PR is this?
/kind cleanup
/area testing

Remove mgr.GetEventRecorderFor is deprecated exclusion.

Fixes #

```release-note
NONE
```

## Summary by CodeRabbit

* **Chores**
  * Updated linting exclusion rules to align with current standards, reducing deprecated API warning noise.
  * Refined static analysis suppressions for additional deprecations and narrowed a fatcontext rule for test paths.
* **Tests**
  * Updated multikueue controller tests to use a shared custom event recorder and structured event assertions instead of client-go fake recorders.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -92,9 +92,6 @@ linters:
         text: 'SA1019: features.* is deprecated: planned to be removed in .*'
       - linters:
           - staticcheck
-        text: 'SA1019: mgr.GetEventRecorderFor is deprecated: this uses the old events API and will be removed in a future release. Please use GetEventRecorder instead.'
-      - linters:
-          - staticcheck
         text: 'SA1019: scheme.Builder is deprecated: This helper is only useful in api packages'
       - linters:
           - fatcontext
--- a/pkg/controller/admissionchecks/multikueue/clusterqueue_test.go
+++ b/pkg/controller/admissionchecks/multikueue/clusterqueue_test.go
@@ -27,7 +27,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
-	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -455,7 +454,8 @@ func TestCQReconcile(t *testing.T) {
 				Build()
 
 			adapters, _ := jobframework.GetMultiKueueAdapters(sets.New("batch/job"))
-			cRec := newClustersReconciler(c, TestNamespace, 0, defaultOrigin, nil, adapters, nil, nil, record.NewFakeRecorder(10))
+			recorder := &utiltesting.EventRecorder{}
+			cRec := newClustersReconciler(c, TestNamespace, 0, defaultOrigin, nil, adapters, nil, nil, recorder)
 			cRec.rootContext = ctx
 			for worker, wState := range tc.workers {
 				workerClient := NewNeverCachingClient(utiltesting.NewClientBuilder().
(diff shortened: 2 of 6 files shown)


mimowo:
Does it mean that we forgot to move all places in https://github.com/kubernetes-sigs/kueue/pull/10971 ? 

In that case I think we can handle that as a consistency bug / omission, and cherrypick to 0.18

With the note like 

Observability: Migrate the remaining MultiKueueCluster controller from the old event types `core/v1` to use `events.k8s.io/v1`, by replacing the helper used in controller-runtime.

wdyt?

mbobrovskyi:
Actually, it was introduced in https://github.com/kubernetes-sigs/kueue/pull/12223, which was merged only into the main branch. But yeah, we can cherry-pick it to release-0.18 to remove lint exclusion.

mimowo:
Sorry, I don't get, https://github.com/kubernetes-sigs/kueue/pull/12223 only introduce lint which is irrelevant for users. we care primarly in the release note about user observable behavior. 

I wouldn't bother cherrypicking for lint, but I think there is just a different type of events currently generaed in 0.18 which might be user observable. 

This PR IIUC fixes the inconsistency introduced in https://github.com/kubernetes-sigs/kueue/pull/10971 (at least in my understanding).

mbobrovskyi:
Not exactly. We simply forgot to remove the exclusion from `.golangci.yaml` in the `release-0.18` branch (more precisely, in PR #10971). I tested it locally, and everything works fine without that exclusion on `release-0.18`.

So I think we can cherry-pick this without release notes. The only downside is that it also brings back the older version of the event recorder introduced in PR #12223. However, I don't think that requires release notes either, since `0.19` hasn't been released yet. This would just be a follow-up fix to PR #12223.

mbobrovskyi:
I created the cherry-pick PR: https://github.com/kubernetes-sigs/kueue/pull/12574.

mimowo:
Hold on, so by chending the recorder I thought we are changing also the type of produced events, no?

Let me hold to clarify my confusion
/hold

mimowo:
cc @vladikkuzn who worked on https://github.com/kubernetes-sigs/kueue/pull/10971,

@mbobrovskyi  is the recorder not changing he events indeed? However then I'm confused why the PR required RBAC change https://github.com/kubernetes-sigs/kueue/pull/10971/changes#diff-551fa6e9f4f56c712cbde8e7e176be653d8b14821f0d6d7d2e7fd86dedcd0fb7R139-R147 ?

mimowo:
What is more, I think this PR is actually a regression bugfix introduced to 0.18 in https://github.com/kubernetes-sigs/kueue/pull/10971, becuase that PR took the RBAC: https://github.com/kubernetes-sigs/kueue/pull/10971/changes#diff-551fa6e9f4f56c712cbde8e7e176be653d8b14821f0d6d7d2e7fd86dedcd0fb7L26-L34

but didn't change the recorder for MultiKueueCluster. So if my hypothesis is correct, then on 0.18 we currently have a regression in observability that the events are not visible at all to users.

If this is the case then release note is mandatory and cherrypick. Let me tentatively annotate like that, but please adjust if you disagree

 Could you double check?

kannon92 (commented):
/lgtm

kannon92:
> but didn't change the recorder for MultiKueueCluster. So if my hypothesis is correct, then on 0.18 we currently have a regression in observability that the events are not visible at all to users.

I think I added an events recorder for MultiKueueCLuster for SecurePath. There wasn't one before.

mimowo:
Ahh, ok so the regression was introduced in this PR https://github.com/kubernetes-sigs/kueue/pull/12223 but that was never released. In that case we don't need a release note, and we can handle just as a cleanup.
Still, a CP makes sense to prevent such a regression in the future. Thank you folks for investigating and clarifying my doubts.

/unhold
/remove-kind bug
/remove-kind regression

mbobrovskyi:
/retest

Due to https://github.com/kubernetes-sigs/kueue/issues/12579.

PR #12520, opened 2026-06-26
feat(metrics): add pending scheduling hashes metric

### LINKED ISSUE #12463: metrics: Add a `pending_scheduling_hashes` gauge metric for CQs

With `SchedulingEquivalenceHashing` enabled, we get a notion of workload "classes" that are currently pending in the Kueue, i.e. unique requirements for the workloads.

We could add a metric that would expose how many unique hashes are pending in a queue to see how many of those classes the queue has to consider.

**Why is this needed**:

By dividing the indamissble requeue time _T1_ by the average duration of the scheduling cycle _S1_ we get a notion of more or less how many cycles the scheduler can execute before the inadmissible workloads are requeued.

Without `SchedulingEquivalenceHashing`, we can pit it against the number of pending workloads when trying to estimate whether workloads lower down in the queue are getting starved when the top workloads are inadmissible (i.e. no progress is made in the queue).

With `SchedulingEquivalenceHashing` enabled, the number to compare would have to be the number of pending hashes, since they are moved to inadmissible in bulk.

RooobinYe:
#### What type of PR is this?

/kind feature

Adds a new `kueue_pending_scheduling_hashes` gauge metric that reports the number of unique pending scheduling equivalence hashes per ClusterQueue.

This helps operators evaluate queue progress when `SchedulingEquivalenceHashing` is enabled. In that mode, equivalent pending workloads can be moved to inadmissible in bulk, so the number of pending hashes is the useful comparison point rather than only the number of pending workloads.

Fixes #12463

Validation performed:

```sh
env GOCACHE=/private/tmp/kueue-go-build-cache go test -count=1 ./pkg/metrics ./pkg/cache/queue
ok  	sigs.k8s.io/kueue/pkg/metrics	1.017s
ok  	sigs.k8s.io/kueue/pkg/cache/queue	4.849s
```

```sh
env GOCACHE=/private/tmp/kueue-go-build-cache go test -count=1 ./test/e2e/singlecluster/baseline -run TestDoesNotExist
ok  	sigs.k8s.io/kueue/test/e2e/singlecluster/baseline	0.902s [no tests to run]
```

```sh
cd hack/tools/metricsdoc
env GOCACHE=/private/tmp/kueue-go-build-cache go test -count=1 ./...
ok  	sigs.k8s.io/kueue/hack/tools/metricsdoc	0.498s
```

Manual k3s smoke test:

- Deployed the controller image from this branch.
- Enabled `--feature-gates=SchedulingEquivalenceHashing=true`.
- Created a zero-quota ClusterQueue and three pending Jobs.
- Two Jobs had identical pod specs, and one Job had a different CPU request.
- Scraped the controller metrics endpoint and observed:

```text
kueue_pending_scheduling_hashes{cluster_queue="hash-test-cq",replica_role="leader",status="active"} 0
kueue_pending_scheduling_hashes{cluster_queue="hash-test-cq",replica_role="leader",status="inadmissible"} 2
```

This matches the expected result: three pending workloads represented by two unique scheduling equivalence hashes.

```release-note
Added the `kueue_pending_scheduling_hashes` metric to report the number of unique pending scheduling equivalence hashes per ClusterQueue.
```

## Summary

* **New Features**
  * Added `kueue_pending_scheduling_hashes` metric to report unique pending scheduling-equivalence hashes for ClusterQueues, split by `status` (`active`/`inadmissible`) and labeled by `replica_role`.

* **Bug Fixes**
  * Ensure pending-hash reporting is emitted only when `SchedulingEquivalenceHashing` is enabled.

* **Documentation**
  * Updated metrics reference docs (English and Chinese) to include the new gauge and its labels.

* **Tests**
  * Extended unit, metrics, scheduler integration, and e2e tests to validate reporting, transitions, and cleanup.


mimowo:
/ok-to-test
cc @kshalot

pkg/cache/queue/metrics.go
```diff
@@ -52,6 +52,7 @@ func reportCQPendingWorkloads(m *Manager, cq *ClusterQueue) {
 	}
 	cqCustomLabels := m.customLabels.CQGet(cq.name)
 	metrics.ReportPendingWorkloads(cq.name, active, inadmissible, cqCustomLabels, m.roleTracker)
+	metrics.ReportPendingSchedulingHashes(cq.name, cq.PendingSchedulingHashes(), cqCustomLabels, m.roleTracker)
```
kshalot: We should report this metric only when `SchedulingEquivalenceHashing` is enabled.

pkg/metrics/metrics.go
```diff
@@ -89,6 +89,10 @@ var (
 	// +metricsdoc:labels=cluster_queue="the name of the ClusterQueue",status="status label (varies by metric)",replica_role="one of `leader`, `follower`, or `standalone`"
 	PendingWorkloads *prometheus.GaugeVec
 
+	// +metricsdoc:group=clusterqueue
+	// +metricsdoc:labels=cluster_queue="the name of the ClusterQueue",replica_role="one of `leader`, `follower`, or `standalone`"
+	PendingSchedulingHashes *prometheus.GaugeVec
```
kshalot, line 103: It would probably be good to also break this down by status, like the `PendingWorkloads` metric:
https://github.com/kubernetes-sigs/kueue/blob/42fc89c89364afad6dc64112ef8bb3d0c60389ca/pkg/metrics/metrics.go#L374-L384

pkg/cache/queue/metrics.go
RooobinYe: Fixed. The `pending_scheduling_hashes` metric is now reported only when `SchedulingEquivalenceHashing` is enabled.

pkg/metrics/metrics.go
RooobinYe: Done.

pkg/cache/queue/cluster_queue.go
```diff
@@ -608,6 +608,37 @@ func (c *ClusterQueue) Pending() (int, int) {
...
+
+func (c *ClusterQueue) pendingSchedulingHashes() (int, int) {
+	activeHashes := sets.New[string]()
+	inadmissibleHashes := sets.New[string]()
+	add := func(hashes sets.Set[string], wInfo *workload.Info) {
+		if wInfo.SchedulingHash != "" && wInfo.SchedulingHash != workload.SchedulingHashUnknown {
+			hashes.Insert(wInfo.SchedulingHash)
+		}
+	}
+	for _, wInfo := range c.heap.List() {
```
sohankunkerkar: IIUC, `pendingSchedulingHashes()` is O(active + inadmissible) with a heap.List() allocation on every call. `reportCQPendingWorkloads` calls this on every workload add/requeue/delete/Pop. At 5k+ workloads this regresses the hot path. Maintain hash counts incrementally instead.

pkg/cache/queue/cluster_queue.go
```diff
@@ -608,6 +608,37 @@ func (c *ClusterQueue) Pending() (int, int) {
...
+func (c *ClusterQueue) PendingSchedulingHashes() (int, int) {
+	if !features.Enabled(features.SchedulingEquivalenceHashing) {
+		return 0, 0
+	}
+	c.rwm.RLock()
+	defer c.rwm.RUnlock()
+	return c.pendingSchedulingHashes()
+}
+
+func (c *ClusterQueue) pendingSchedulingHashes() (int, int) {
```
sohankunkerkar: This iterates all workloads to count unique hashes, but `noFitSchedulingHashes` already tracks hashes the equivalence optimization acts on and its Len() is O(1). This metric counts all unique hashes including workloads inadmissible for other reasons, which is broader but less actionable. Consider exposing `noFitSchedulingHashes.Len()` instead.

pkg/cache/queue/cluster_queue.go
RooobinYe: Thanks, done. I changed this to maintain scheduling hash reference counts incrementally instead of rebuilding the set on every report.

The implementation keeps three map[string]int counters:

```go
activeSchedulingHashCounts       map[string]int
inadmissibleSchedulingHashCounts map[string]int
pendingSchedulingHashCounts      map[string]int
```

pkg/cache/queue/cluster_queue.go
RooobinYe: I considered this option, but kept the original pending-hash semantics here.

pkg/cache/queue/cluster_queue.go
```diff
@@ -136,6 +136,17 @@ type ClusterQueue struct {
...
+	// Scheduling hash counters are kept incrementally so reporting the
+	// pending_scheduling_hashes metric doesn't scan the pending workloads.
+	// activeSchedulingHashCounts tracks heap workloads; inflight is folded in at
+	// read time because Pop moves it out of the heap but it is still active.
+	// pendingSchedulingHashCounts is the union counter across active and
+	// inadmissible workloads, used when inactive ClusterQueues report all
+	// pending workloads as inadmissible without double-counting shared hashes.
+	activeSchedulingHashCounts       map[string]int
+	inadmissibleSchedulingHashCounts map[string]int
+	pendingSchedulingHashCounts      map[string]int
```
mimowo: I think we should wrap the helper structures as we did for `inadmissibleWorkloads`, for two reasons:
1. to put together fields responsible for one feature together 
2. to allow injecting a RW lock in case we need to ensure consistency of the structs. For example that workload is not double counted in another goroutine when it is moving between "active" and "pending".
3. when we encapsulate the new logic in a new struct we can move the code alltogether to a dedicated file for scheduling hashes. This will offload the already too loaded cluster_queue.go code.

Something like `schedulingHashCounts`

We may also consider combining that with `hashToBulkMoveReason`, but this is less obvious.

mimowo: cc @kshalot @j-skiba please check the idea for sanity here and how this aligns with the plans for the code structuring for the granular observability feature

mimowo:
This new metric is closely related with two features:
- schedliung hashes - owned by @sohankunkerkar 
- granular workload observability  for unadmitted workloads - owned by @j-skiba and @kshalot 

So I would like to wait for lgtm from the feature owners.

pkg/cache/queue/cluster_queue.go
j-skiba: +1 for extracting these to a separate structs

-------

also as @kshalot noticed before in another PR maybe we should also include the new `QuotaReserved` reasons in these metrics as labels? This data is already in `hashToBulkMoveReason`.

Note: it should be `UnadmittedWorkloadsObservability` FG enabled in such case

j-skiba: I've just read the PR in more detail

> also as @kshalot noticed before in another PR maybe we should also include the new QuotaReserved reasons in these metrics as labels? This data is already in hashToBulkMoveReason.

that's probably not applicable here

pkg/cache/queue/cluster_queue.go
```diff
@@ -658,6 +728,46 @@ func (c *ClusterQueue) Pending() (int, int) {
...
+	return len(c.inadmissibleSchedulingHashCounts)
+}
+
+func (c *ClusterQueue) pendingSchedulingHashCountsLen() int {
+	return c.schedulingHashCountsLenWithInflight(c.pendingSchedulingHashCounts)
+}
+
+func (c *ClusterQueue) schedulingHashCountsLenWithInflight(counts map[string]int) int {
+	result := len(counts)
+	if c.inflight != nil && validSchedulingHash(c.inflight.SchedulingHash) && counts[c.inflight.SchedulingHash] == 0 {
```
j-skiba: I'm not sure what's the expected result but this behavior might be misleading

`inflight` is cleared on explicit workload deletion or new `Pop()`.

Let's say there is a single workload in the heap:
- `Pop()` is called and we try to schedule the workload. Old `inflight` is cleared and the popped workload is assigned to `inflight`.
- `kueue_pending_workloads{status="active"}` reports `0` when a workload is admitted
- `kueue_pending_scheduling_hashes{status="active"}` will be reporting `1` until `Pop()` is called again

pkg/cache/queue/cluster_queue.go
sohankunkerkar: IIUC, `pendingSchedulingHashCounts` is updated on every hot-path mutation but only read in the cold `pendingSchedulingHashesForInactiveClusterQueue` path (stopped/deleted CQ). Consider dropping this map and computing the union lazily on read i.e. iterate `activeSchedulingHashCounts` and count hashes absent from `inadmissibleSchedulingHashCounts`. O(len(active hashes)), bounded and small.

pkg/cache/queue/cluster_queue_test.go
```diff
@@ -1579,7 +1581,194 @@ func TestPushOrUpdateRespectsInadmissibleHashes(t *testing.T) {
 	}
 }
 
+func TestPendingSchedulingHashes(t *testing.T) {
+	features.SetFeatureGateDuringTest(t, features.SchedulingEquivalenceHashing, true)
+
+	ctx, _ := utiltesting.ContextWithLog(t)
+	now := time.Now()
+	cq := newClusterQueueImpl(ctx, nil, defaultOrdering, testingclock.NewFakeClock(now))
```
sohankunkerkar: nit: add a test case where the inflight workload's hash is shared with an inadmissible workload, which exercises the `counts[hash] == 0` branch in `schedulingHashCountsLenWithInflight`.

pkg/cache/queue/inadmissible_workloads.go
```diff
@@ -184,6 +184,8 @@ func queueInadmissibleWorkloads(ctx context.Context, c *ClusterQueue, client cli
 		if err != nil || !c.namespaceSelector.Matches(labels.Set(ns.Labels)) || !c.backoffWaitingTimeExpired(wInfo) {
 			newInadmissibleWorkloads.insert(key, wInfo)
 		} else if c.heap.PushIfNotPresent(wInfo) {
```
sohankunkerkar: If `PushIfNotPresent` returns false (reachable via `AddFromLocalQueue` resync), this workload falls through both branches and dropped from `inadmissibleWorkloads` by `replaceAll` but `removeInadmissibleSchedulingHash` never fires, permanently leaking the hash count.

pkg/cache/queue/cluster_queue.go
```diff
@@ -608,6 +608,33 @@ func (c *ClusterQueue) Pending() (int, int) {
 	return c.pendingActive(), c.pendingInadmissible()
 }
 
+// PendingSchedulingHashes returns the number of unique known scheduling hashes
+// across active, inflight, and inadmissible pending workloads.
+func (c *ClusterQueue) PendingSchedulingHashes() int {
+	c.rwm.RLock()
+	defer c.rwm.RUnlock()
+	return c.pendingSchedulingHashes()
+}
```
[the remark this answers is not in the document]
RooobinYe: Solved. Now it `return 0, 0`

pkg/cache/queue/cluster_queue_test.go
```diff
@@ -1527,6 +1527,41 @@ func TestPushOrUpdateRespectsInadmissibleHashes(t *testing.T) {
 	}
 }
 
+func TestPendingSchedulingHashes(t *testing.T) {
+	ctx, _ := utiltesting.ContextWithLog(t)
+	now := time.Now()
+	cq := newClusterQueueImpl(ctx, nil, defaultOrdering, testingclock.NewFakeClock(now))
```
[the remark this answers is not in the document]
RooobinYe: Solved

pkg/cache/queue/cluster_queue.go
RooobinYe: Solved. I used a structure.

```go
type schedulingHashCounts struct {
	// active counts hashes of heap workloads; the inflight workload is folded
	// in at read time because Pop moves it out of the heap while it is still
	// being scheduled.
	active map[string]int
	// inadmissible counts hashes of workloads in inadmissibleWorkloads.
	inadmissible map[string]int
}
```

pkg/cache/queue/cluster_queue_test.go
RooobinYe: Done. I added a test where the inflight workload shares its scheduling hash with an inadmissible workload. And it passed.

pkg/cache/queue/inadmissible_workloads.go
RooobinYe: Fixed.

mimowo:
@RooobinYe please rebase

mimowo:
@sohankunkerkar @kshalot please check again

pkg/cache/queue/inadmissible_workloads.go
```diff
@@ -183,7 +183,11 @@ func queueInadmissibleWorkloads(ctx context.Context, c *ClusterQueue, client cli
 		err := client.Get(ctx, types.NamespacedName{Name: wInfo.Obj.Namespace}, &ns)
 		if err != nil || !c.namespaceSelector.Matches(labels.Set(ns.Labels)) || !c.backoffWaitingTimeExpired(wInfo) {
 			newInadmissibleWorkloads.insert(key, wInfo)
-		} else if c.heap.PushIfNotPresent(wInfo) {
+		} else {
+			// The workload leaves inadmissibleWorkloads regardless of whether the
+			// push succeeds (it may already be in the heap after an
+			// AddFromLocalQueue resync), so always release its inadmissible state.
+			c.moveInadmissibleToActive(key, wInfo)
 			moved++
```
sohankunkerkar:
```suggestion
			if c.moveInadmissibleToActive(key, wInfo) {
                 moved++			
			}
```

sohankunkerkar: nit: `moved++` is now unconditional. On `main` it only incremented when `PushIfNotPresent` returned true (i.e. the workload actually moved from inadmissible to the heap). With this change, `moved` also counts workloads that were already in the heap (the `AddFromLocalQueue` resync case), where `moveInadmissibleToActive` returns `false` because it only cleaned up the inadmissible state without actually moving anything.

sohankunkerkar (approved):
one nit but overall LGTM

pkg/cache/queue/cluster_queue.go
```diff
@@ -500,12 +499,87 @@ func (c *ClusterQueue) subtractPendingResources(wInfo *workload.Info) {
...
+		wInfo = old
+		c.schedulingHashes.removeInadmissible(old)
+		c.inadmissibleWorkloads.delete(key)
+		if c.heap.PushIfNotPresent(wInfo) {
+			// Keep the existing resource contribution while changing buckets.
+			c.schedulingHashes.addActive(wInfo)
+			return true
+		}
+		// AddFromLocalQueue already counted the heap copy separately.
+		c.subtractPendingResources(old)
```
kshalot: QQ: Was this also somehow called on the main branch? Or did we have a leak after `AddFromLocalQueue`? I'm asking because AFAICS this piece of logic was not present in `queueInadmissibleWorkloads` at all.

sohankunkerkar: Good catch! This is actually fixing a pre-existing double-counting leak on main. When `AddFromLocalQueue` pushes a workload that's already inadmissible, resources get counted twice, and `replaceAll` later drops the inadmissible entry without subtracting. The `subtractPendingResources` here is the fix. But I think it would be helpful to add a test for this for sure.

mimowo: If this is a bugfix for main, I would prefer to extract as a separate bugfix PR, so that we can cherrypick.

mimowo:
@RooobinYe do you have cycles to address the remaining comments?

RooobinYe:
Yes, of course.

pkg/cache/queue/inadmissible_workloads.go
RooobinYe: Fixed

pkg/cache/queue/cluster_queue.go
RooobinYe: I will create a new PR to address this problem.

UPDATE: New PR is #13660

pkg/cache/queue/scheduling_hash_counts.go
```diff
@@ -0,0 +1,137 @@
...
+
+// PendingSchedulingHashes returns the number of unique known scheduling hashes
+// for active and inadmissible pending workloads.
+func (c *ClusterQueue) PendingSchedulingHashes() (int, int) {
+	if !features.Enabled(features.SchedulingEquivalenceHashing) {
+		return 0, 0
+	}
+	c.rwm.RLock()
+	defer c.rwm.RUnlock()
+	return c.schedulingHashes.activeLen(c.inflight), c.schedulingHashes.inadmissibleLen()
```
mimowo: This taking of the high level lock is suspicious. Could we instead add a R/W lock to the schedulingHashes struct

pkg/cache/queue/scheduling_hash_counts.go
```diff
@@ -0,0 +1,137 @@
...
+// pendingSchedulingHashesForInactiveClusterQueue reports all pending hashes
+// as inadmissible, mirroring how inactive ClusterQueues report their pending
+// workloads.
+func (c *ClusterQueue) pendingSchedulingHashesForInactiveClusterQueue() (int, int) {
+	if !features.Enabled(features.SchedulingEquivalenceHashing) {
+		return 0, 0
+	}
+	c.rwm.RLock()
+	defer c.rwm.RUnlock()
+	return 0, c.schedulingHashes.pendingUnionLen(c.inflight)
```
mimowo: Here too, I hope we can skip high level lock if using a dedicated one. Our locks in Kueue are too broad in general. trying to control them in small steps

pkg/cache/queue/scheduling_hash_counts.go
```diff
@@ -0,0 +1,137 @@
...
+		return
+	}
+	hash := wInfo.SchedulingHash
+	counts[hash]--
+	if counts[hash] <= 0 {
+		delete(counts, hash)
+	}
+}
+
+func (s *schedulingHashCounts) addActive(wInfo *workload.Info) {
```
mimowo, line 78: Can we also introduce higher-level functions that would be useful, say moveToActive, or update, or moveToInadmissible.

mimowo:
Left some comments, but an additional pass from @kshalot or @sohankunkerkar would be great

mimowo:
@RooobinYe do you have cycles to address the remaining comments?

pkg/cache/queue/scheduling_hash_counts.go
RooobinYe: Fixed

pkg/cache/queue/scheduling_hash_counts.go
RooobinYe: Sure, I fixed it

pkg/cache/queue/scheduling_hash_counts.go
RooobinYe: I added `moveToActive`、`moveToInadmissible`、`updateActive`、`updateInadmissible`、`moveActiveToInflight`、`clearInflight`.

pkg/cache/queue/cluster_queue.go
RooobinYe: The inflight workload is counted by both pending metrics while it is being scheduled. Once it is successfully admitted and removed from the queue, both metrics are updated to zero.

pkg/cache/queue/scheduling_hash_counts.go
```diff
@@ -0,0 +1,211 @@
...
+		return 0, 0
+	}
+	return c.schedulingHashes.pendingLens()
+}
+
+// pendingSchedulingHashesForInactiveClusterQueue reports all pending hashes
+// as inadmissible, mirroring how inactive ClusterQueues report their pending
+// workloads.
+func (c *ClusterQueue) pendingSchedulingHashesForInactiveClusterQueue() (int, int) {
+	if !features.Enabled(features.SchedulingEquivalenceHashing) {
```
kshalot: Do we have to guard those functions with the FG check? It seems that we are already guarding the metrics code itself from calling this in the first place.

kshalot (commented):
/lgtm overall. One tiny comment.

Thanks! 🎉

mimowo (commented):
/lgtm
/approve
Thank you for extending the mechanism with a metric 👍

RooobinYe:
Hi @mimowo @sohankunkerkar 👋 Thanks again for the thorough reviews on this PR, and sorry for the off-topic ping on a closed PR.

I've now had 4 merged PRs in Kueue (this one, plus #12436, #13574 and #13660), and I'd like to apply for kubernetes-sigs org membership. Since you both reviewed and approved several of my changes, would you be willing to be my two sponsors?

Thanks!

PR #12522, opened 2026-06-26
Clean up ProvisioningRequests on Workload eviction

### LINKED ISSUE #11677: Kueue doesn't clean up ProvisioningRequest after Workload's eviction
**What happened**:
Kueue assigned a flavor to a Workload with ProvisioningRequest as an AdmissionCheck. After the Workload was evicted e.g. by higher priority Workload, or deactivation, the ProvisioningRequest is kept which is a bad design pattern

ProvisioningRequest to be cleaned-up after Workload is evicted

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

MatteoFari:
#### What type of PR is this?
/kind bug

Deletes old `ProvisioningRequest`s when a `Workload` is marked as finished or evicted.

Before, the controller stopped too early and left old requests behind. Retry info is still kept on the `Workload`, and a default-on feature gate can turn off eviction cleanup.

Fixes #11677

```release-note
ProvisioningRequest: Fix a bug where ProvisioningRequest owned by finished or evicted Workloads are not cleaned up. The CleanupProvisioningRequestsOnEviction feature gate allows cleanup on eviction to be enabled by default.
```

## Summary by CodeRabbit

* **New Features**
  * Added a feature gate to automatically clean up provisioning requests when a workload is evicted (enabled by default).
* **Bug Fixes**
  * Improved reconcile behavior to distinguish completion vs. eviction and only clean up on eviction when the setting is enabled.
  * Preserves retry context when creating new provisioning requests after cleanup.
* **Tests**
  * Updated unit and integration coverage for Finished/Evicted lifecycle and retry/backoff scenarios, including validation of deletion when cleanup on eviction is enabled/disabled.

### DIFF
--- a/pkg/controller/admissionchecks/provisioning/controller.go
+++ b/pkg/controller/admissionchecks/provisioning/controller.go
@@ -47,6 +47,7 @@ import (
 
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
 	"sigs.k8s.io/kueue/pkg/constants"
+	"sigs.k8s.io/kueue/pkg/features"
 	"sigs.k8s.io/kueue/pkg/podset"
 	"sigs.k8s.io/kueue/pkg/util/admissioncheck"
 	"sigs.k8s.io/kueue/pkg/util/api"
@@ -128,13 +129,27 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	log := ctrl.LoggerFrom(ctx)
 	log.V(2).Info("Reconcile Workload")
 
-	if !workload.HasQuotaReservation(wl) || workload.IsFinished(wl) || workload.IsEvicted(wl) {
+	isFinished := workload.IsFinished(wl)
+	isEvicted := workload.IsEvicted(wl)
+	if !workload.HasQuotaReservation(wl) && !isFinished && !isEvicted {
 		return reconcile.Result{}, nil
 	}
 
 	provReqs := &autoscaling.ProvisioningRequestList{}
 	if err := c.client.List(ctx, provReqs, client.InNamespace(wl.Namespace), client.MatchingFields{RequestsOwnedByWorkloadKey: wl.Name}); client.IgnoreNotFound(err) != nil {
 		return reconcile.Result{}, err
+	}
+
+	if isFinished || (isEvicted && features.Enabled(features.CleanupProvisioningRequestsOnEviction)) {
+		err = c.deleteUnusedProvisioningRequests(ctx, provReqs.Items, nil)
+		if err != nil {
+			log.V(2).Error(err, "failed to delete stale provisioning requests")
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{}, nil
+	}
+	if isEvicted {
+		return reconcile.Result{}, nil
 	}
 
 	// get the lists of relevant checks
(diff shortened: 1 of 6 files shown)


pkg/controller/admissionchecks/provisioning/controller.go
```diff
@@ -136,6 +144,15 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{}, err
 	}
 
+	if shouldDeleteProvReqs {
```
mimowo: Instead of communiacting the two places in the function by shouldDeleteProvReqs why not just do this new snippet (moved up):

``golang
if features.Enabled(features.CleanupProvisioningRequestsOnEviction) && (IsEvicted() || IsFinished()) {
		err = c.deleteUnusedProvisioningRequests(ctx, provReqs.Items, nil)
		if err != nil {
			log.V(2).Error(err, "failed to delete stale provisioning requests")
		}
	}
```
Seems like it could be more readable, because also the use of negations with !feature() makes it hard to follow

File: pkg/controller/admissionchecks/provisioning/controller.go [hunk 2]
```diff
@@ -251,6 +268,9 @@ func (c *Controller) syncOwnedProvisionRequest(
 
 		req, exists := activeOrLastPRForChecks[checkName]
 		attempt := int32(1)
+		if ac != nil {
+			attempt = ptr.Deref(ac.RetryCount, 0) + 1
+		}
```
mimowo, line 274: how is this related to the fix? It does not seem at the first sight, maybe it is a bugfix for a separate issue, tracked separately. If this is integral required part of the fix then please add some commentary because this is far from obvious

pkg/controller/admissionchecks/provisioning/controller_test.go
```diff
@@ -316,6 +316,8 @@ func TestReconcile(t *testing.T) {
 		wantTemplates        map[string]*corev1.PodTemplate
 		wantRequestsNotFound []string
 		wantEvents           []utiltesting.EventRecord
+
+		cleanupProvReqsOnEviction *bool
```
mimowo: introuce the more generic featureGates map for the feature gate enablement. We go with this approach as  much more scalable. Also, move this up because ":want" section should be at the bottom typically

pkg/controller/admissionchecks/provisioning/controller.go
MatteoFari: once cleanup removes the previous request, the next attempt must come from `Workload.status.admissionChecks[].retryCount`, otherwise we restart at attempt 1.

mimowo: Ah, an in the past we didn't have to use retryCount because we could rely on the `attempt` parsed from the Provreq name?

MatteoFari: Yes exactly

mimowo (commented):
/lgtm
/approve
Thank you for the ProvReq integration improvement, looks great to me. Please prepare the cherrypicks manually, updating the FG to the version corresponding to the branch versions.

mimowo:
/release-note-edit
```release-note
ProvisioningRequests owned by finished or evicted Workloads are now cleaned up by default. Added the CleanupProvisioningRequestsOnEviction feature gate to allow disabling cleanup on eviction.
```

tenzen-y:
/release-note-edit

```release-note
ProvisioningRequest: Fix a bug where ProvisioningRequest owned by finished or evicted Workloads are not cleaned up. The CleanupProvisioningRequestsOnEviction feature gate allows cleanup on eviction to be enabled by default.
```

PR #12562, opened 2026-06-26
TAS: infer flavor for grouped PodSets with no managed resource requests

### LINKED ISSUE #12417: LWS podset TAS regression since v0.17.5 when leader does not request resources managed by CQ
**What happened**:
I have a CQ that only manages one type of precious resources (machines with those in the cluster are homogeneous), e.g. GPUs.
I have a topology specified for the flavor used in this CQ.
I tried to create LWS job and deploy it to the queue.

As leader does not require the precious resources I have not specified appropriate request for it.
Workers do require the resource so I have added appropriate entry.

LWS job never got admitted, although nodes with appropriate resources were idle all the time.
My investigation showed that it fails admission because of this condition:
https://github.com/kubernetes-sigs/kueue/blob/release-0.18/pkg/scheduler/flavorassigner/tas_flavorassigner.go#L156

which in combination with recent [fix](https://github.com/kubernetes-sigs/kueue/pull/12055
) results in workload treated as invalid. 

If additional resource exists on the list of resources managed by CQ validation succeeds (e.g. cpu). 
If "0" resources are requested explicitly by the leader podset the validation succeeds.

Workload getting properly admitted.

0. Deploy Kueue with LWS integration enabled.
1. Have a cluster queue with TAS enabled flavor and just one resource managed.
2. Create a LWS workload with one of the podsets not requesting the managed resource (for example with empty resource requests).

Reproduced on different environments v0.17.5 and v0.18.1

**Environment**:
I believe not connected to particular environment as specific line of code and PR that introduce the regression have bin pinned - omitting to minimize the distraction.

mszadkow:
#### What type of PR is this?
/kind bug
/kind regression
/area tas
/area integrations

When a TAS-enabled Workload has a PodSet that requests none of its ClusterQueue's
managed resources, Kueue's flavor assigner previously left that PodSet's flavor
unresolved and rejected it with `no TAS flavor assigned`.

This PR resolves the flavor for such a PodSet by inheriting the TAS flavor already resolved for its PodSet group, instead of  being filtered down to an empty flavor set.

If this can't be applied the PodSet still correctly fails, since there is no information to determine which flavor it should use.

Fixes #12417

```release-note
TAS: Fixed a bug where a grouped PodSet (e.g. an LWS leader) with no requests for the TAS-managed resource was rejected with "no TAS flavor assigned".
```

## Summary

* **Bug Fixes**
  * Improved TAS flavor assignment for grouped pod sets, including leader/worker behavior when the leader has no managed resource requests but is part of a topology-aware group.
  * Enhanced conflict detection and clearer failure behavior when multiple TAS flavors could be assigned (while preserving worker pod set assignments where applicable).
  * Now rejects required topology requests for ElasticJobs using workload slices, with the error surfaced on the affected pod set.

* **Tests**
  * Added regression coverage for grouped leader/worker TAS resolution and the ElasticJobs required-topology rejection case.

### DIFF
--- a/pkg/scheduler/flavorassigner/flavorassigner.go
+++ b/pkg/scheduler/flavorassigner/flavorassigner.go
@@ -788,15 +788,7 @@ func (a *FlavorAssigner) assignFlavors(ctx context.Context, log logr.Logger, cou
 		finalConsidered := finalizeFlavorAssignmentAttempts(consideredFlavors)
 		atLeastOnePodsAssignmentFailed := false
 		for _, podSet := range podSets {
-			var reqKeys []corev1.ResourceName
-			if podSet.podSet.Requests != nil {
-				podSet.podSet.Requests.ForEach(func(name corev1.ResourceName, _ int64) {
-					reqKeys = append(reqKeys, name)
-				})
-			}
-			podSetFlavors := utilmaps.FilterKeys(groupFlavors, reqKeys)
-
-			podSet.podSetAssignment.Flavors = podSetFlavors
+			podSet.podSetAssignment.Flavors = a.resolvePodSetFlavors(log, podSet, groupFlavors)
 			podSet.podSetAssignment.Status = groupStatus
 			podSet.podSetAssignment.FlavorAssignmentAttempts = finalConsidered
 
@@ -865,6 +857,43 @@ func (a *FlavorAssigner) assignFlavors(ctx context.Context, log logr.Logger, cou
 	return assignment
 }
 
+// resolvePodSetFlavors returns the flavors podSet should be assigned, given the flavors
+// already resolved for its whole PodSet group (groupFlavors). Normally this is just
+// groupFlavors filtered down to the resources podSet itself requests. A PodSet requesting
+// none of the group's managed resources (e.g. an LWS leader) would otherwise end up with no
+// flavor and be rejected from TAS, so such a PodSet instead falls back to the group's TAS
+// flavor(s) if it belongs to a topology group. A ClusterQueue-wide fallback can be revisited
+// later if users request it.
+func (a *FlavorAssigner) resolvePodSetFlavors(log logr.Logger, idxPodSet indexedPodSet, groupFlavors ResourceAssignment) ResourceAssignment {
+	// For PodSets with requests, keep only flavors for resources this PodSet requests.
+	if idxPodSet.podSet.Requests != nil && idxPodSet.podSet.Requests.Len() != 0 {
+		var reqKeys []corev1.ResourceName
+		idxPodSet.podSet.Requests.ForEach(func(name corev1.ResourceName, _ int64) {
+			reqKeys = append(reqKeys, name)
+		})
+		podSetFlavors := utilmaps.FilterKeys(groupFlavors, reqKeys)
+		log.V(5).Info("Resolved PodSet flavors from group flavors",
+			"podSet", idxPodSet.podSet.Name,
+			"requestedResources", idxPodSet.podSet.Requests.Len(),
+			"resolvedFlavors", len(podSetFlavors))
+		return podSetFlavors
+	}
+
+	// For PodSets without requests, reuse TAS flavors from the topology group when available.
+	if groupName := podSetGroupName(&a.wl.Obj.Spec.PodSets[idxPodSet.originalIndex]); groupName != nil {
+		// A PodSet with no resource requests in a topology group (e.g. an LWS leader) still needs a
+		// resolved TAS flavor so it can be placed; keep the group's TAS flavor(s) instead
+		// of filtering the group's resolution down to nothing.
+		podSetFlavors := tasFlavorsOnly(groupFlavors, a.cq.TASFlavors)
+		if len(podSetFlavors) > 0 {
+			log.V(5).Info("Using TAS flavors from topology group for PodSet with no resource requests", "podSet", idxPodSet.podSet.Name, "flavors", podSetFlavors)
+			return podSetFlavors
+		}
+	}
+
+	return nil
+}
+
 func (a *Assignment) resolveNoFitReason(cq *schdcache.ClusterQueueSnapshot) {
 	if a.RepresentativeMode() != NoFit {
 		return
@@ -912,6 +941,17 @@ func (a *Assignment) resolveNoFitReason(cq *schdcache.ClusterQueueSnapshot) {
 		overallReason = mostSevereReason(overallReason, podSetReason)
 	}
 	a.NoFitReason = overallReason
+}
+
+// tasFlavorsOnly returns the subset of resourceAssignment whose flavor is a TAS flavor.
+func tasFlavorsOnly(resourceAssignment ResourceAssignment, tasFlavors map[kueue.ResourceFlavorReference]*schdcache.TASFlavorSnapshot) ResourceAssignment {
+	result := make(ResourceAssignment, len(resourceAssignment))
+	for resName, flavorAssignment := range resourceAssignment {
+		if _, isTAS := tasFlavors[flavorAssignment.Name]; isTAS {
+			result[resName] = flavorAssignment
+		}
+	}
+	return result
 }
 
 func findRGIndicesByFlavor(cq *schdcache.ClusterQueueSnapshot, flavor kueue.ResourceFlavorReference) []int {
--- a/pkg/scheduler/flavorassigner/flavorassigner_test.go
+++ b/pkg/scheduler/flavorassigner/flavorassigner_test.go
@@ -30,6 +30,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/component-base/featuregate"
 	"k8s.io/utils/ptr"
 
@@ -39,9 +40,11 @@ import (
 	"sigs.k8s.io/kueue/pkg/features"
 	"sigs.k8s.io/kueue/pkg/resources"
 	preemptioncommon "sigs.k8s.io/kueue/pkg/scheduler/preemption/common"
+	"sigs.k8s.io/kueue/pkg/util/resourcegroups"
 	"sigs.k8s.io/kueue/pkg/util/tas"
 	utiltesting "sigs.k8s.io/kueue/pkg/util/testing"
 	utiltestingapi "sigs.k8s.io/kueue/pkg/util/testing/v1beta2"
+	testingnode "sigs.k8s.io/kueue/pkg/util/testingjobs/node"
 	"sigs.k8s.io/kueue/pkg/workload"
 )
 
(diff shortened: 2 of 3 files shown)


j-skiba (commented):
I think this fix exposes another issue here 
https://github.com/kubernetes-sigs/kueue/blob/571fa9b2137ec871d15b38c435eec29689a9412b/pkg/resources/requests.go#L226
as `r` will be empty then the returned `result` will be equal to `0`.

Maybe let's add an integration test for that

pkg/scheduler/flavorassigner/tas_flavorassigner.go
```diff
@@ -50,6 +54,11 @@ func (a *Assignment) WorkloadsTopologyRequests(log logr.Logger, wl *workload.Inf
 				log.V(3).Info("Skipping TAS for count=0 podSet", "podSet", podSet.Name)
 				continue
 			}
+			a.normalizeTASPodSetAssignmentForManagedResources(psAssignment, wl, cq, i)
+			if psAssignment.Status.IsError() {
+				a.representativeMode = nil
```
mimowo: why not use the helper we already have here: https://github.com/kubernetes-sigs/kueue/pull/12055/changes#diff-0870ec2c0ea330a7d087a6ef599905bf9b7164a0dd889d1b00ba17c31a1bfd05R244-R247?

pkg/scheduler/flavorassigner/tas_flavorassigner.go
```diff
@@ -85,6 +94,115 @@ func (a *Assignment) WorkloadsTopologyRequests(log logr.Logger, wl *workload.Inf
...
+// PodSet's managed resources that the PodSet itself does not request, inferring the flavor
+// from a same-group peer.
+//
+// This handles formations such as LWS, where a leader PodSet may declare no resource
+// requests of its own and would otherwise end up with no flavor assigned and be rejected
+// from TAS. It only acts when there is something to infer: if the PodSet is not grouped,
+// already has its flavor resolved, or no peer provides one, the assignment is left
+// untouched. The inferred flavors are written back into psAssignment.Flavors so that all
+// downstream consumers act on the same assignment.
+func (a *Assignment) normalizeTASPodSetAssignmentForManagedResources(
```
mimowo: This looks like an explosion of complexity for the relatively simple corner case. I'm also not sure , the peer PodSet checks seem like might be dependent on the order of processing the PodSets, wdyt @mszadkow ? Maybe we should make sure that the PodSets with no restrictions are considered last, but I cannot find easily a change that woutd be responsible for ordering.

mimowo:
/hold
cc @dkaluza ptal

pkg/scheduler/flavorassigner/flavorassigner_test.go
```diff
@@ -5652,3 +5653,295 @@ func TestIsNoFitDueToCapacityAndLimits(t *testing.T) {
...
+				*utiltestingapi.MakePodSet("leader", 1).
+					RequiredTopologyRequest(corev1.LabelHostname).
+					PodSetGroup("group1").
+					Obj(),
+				*utiltestingapi.MakePodSet("worker", 1).
+					Request(corev1.ResourceCPU, "1").
+					Request("example.com/gpu", "1").
+					RequiredTopologyRequest(corev1.LabelHostname).
+					PodSetGroup("group1").
+					Obj(),
```
dkaluza: How does this differ from "regression: leader missing managed request is treated as zero"?
In both leader does not request anything. Is this introducing any code coverage?

pkg/scheduler/flavorassigner/flavorassigner_test.go
```diff
@@ -5652,3 +5653,295 @@ func TestIsNoFitDueToCapacityAndLimits(t *testing.T) {
...
+					"example.com/gpu":  {Name: "tas", Mode: Fit, TriedFlavorIdx: -1},
+				}, Count: 1, Status: *NewStatus()},
+			},
+			wantPodSetErrors: map[kueue.PodSetReference]string{},
+			wantTASRequestsByPodSet: map[kueue.PodSetReference]kueue.ResourceFlavorReference{
+				"leader": "tas",
+				"worker": "tas",
+			},
+		},
+		"standalone non-consuming podset without a group is rejected": {
```
dkaluza: Are we sure this is intended behavior?

IMO nothing in current docs https://kueue.sigs.k8s.io/docs/tasks/run/leaderworkerset/
says you cannot have those without group. From my perspective this shouldn't be rejected - just user will not have any guarantees about the same topology domain.

Probably it would be good to emit some warning in logs about this though.

If this cannot be handled gracefully we should update the docs to indicate this.

pkg/scheduler/flavorassigner/flavorassigner_test.go
```diff
@@ -5652,3 +5653,295 @@ func TestIsNoFitDueToCapacityAndLimits(t *testing.T) {
...
+				{Name: "worker", Flavors: ResourceAssignment{"example.com/gpu": {Name: "tas", Mode: Fit, TriedFlavorIdx: -1}}, Count: 1, Status: *NewStatus()},
+			},
+			wantPodSetErrors: map[kueue.PodSetReference]string{
+				"leader": "no TAS flavor assigned",
+			},
+			wantTASRequestsByPodSet: map[kueue.PodSetReference]kueue.ResourceFlavorReference{
+				"worker": "tas",
+			},
+		},
+		"still fails on truly invalid TAS assignment (no same-group peer)": {
```
dkaluza: Why are we treating this as invalid?

Cannot group2 be admitted to this flavor?
This resource flavor has all of the resources required by this podset (empty set of resources)

pkg/scheduler/flavorassigner/flavorassigner_test.go
```diff
@@ -5652,3 +5653,295 @@ func TestIsNoFitDueToCapacityAndLimits(t *testing.T) {
...
+					Obj(),
+				*utiltestingapi.MakePodSet("completed-worker", 1).
+					Request("example.com/gpu", "1").
+					RequiredTopologyRequest(corev1.LabelHostname).
+					PodSetGroup("group1").
+					Obj(),
+			},
+			psAssignments: []PodSetAssignment{
+				{Name: "leader", Flavors: ResourceAssignment{}, Count: 1, Status: *NewStatus()},
+				{Name: "completed-worker", Flavors: ResourceAssignment{"example.com/gpu": {Name: "tas", Mode: Fit, TriedFlavorIdx: -1}}, Count: 0, Status: *NewStatus()},
```
dkaluza: Is this really possible to have count 0 here during admission when 1 is required by the lws?

pkg/scheduler/flavorassigner/flavorassigner_test.go
```diff
@@ -5652,3 +5653,295 @@ func TestIsNoFitDueToCapacityAndLimits(t *testing.T) {
...
+				TASFlavors:     tasFlavors,
+				ResourceGroups: resourceGroups,
+			}
+
+			tasReqs := assignment.WorkloadsTopologyRequests(testr.New(t), wl, &cq)
+
+			gotErrors := map[kueue.PodSetReference]string{}
+			for _, ps := range assignment.PodSets {
+				if ps.Status.err != nil {
+					gotErrors[ps.Name] = ps.Status.err.Error()
```
dkaluza: Generally it would be better to use structured errors and just check their types & fields.

If you are using string comparison for error checking in tests then tests using those can become "change detectors" which are considered to be harmful for the maintenance.  Just adjustment of the error wording will require X tests update. If you are using error type and fields this would not be the case.

pkg/scheduler/flavorassigner/tas_flavorassigner.go
dkaluza: I agree that this looks complex for such simple corner case.
I'm wondering if this is actually a good direction - maybe we should go back a little bit and investigate why actually we do not have any psAssignments for those podsets? Maybe our filtering is actually imperfect? See my comment in [issue](https://github.com/kubernetes-sigs/kueue/issues/12417#issuecomment-4809361755 )

dkaluza (commented):
For now I have reviewed the tests, I will try to review the rest tomorrow.

pkg/scheduler/flavorassigner/tas_flavorassigner.go
mszadkow: @mimowo  What do you refer to as restrictions?

@dkaluza Actually we have the psAssignment for the podSet, but we don't proceed due to lack of flavors.

dkaluza: But why psAssignment doesn't have any flavors? Have you looked into the root cause of this?

I think mimowo had resource requests in mind here.

pkg/scheduler/flavorassigner/flavorassigner_test.go
mszadkow: This one has one more resource outside of managed 
But yes indeed `onlyTASFlavor` filters by `tasFlavors[flavorAssignment.Name] != nil` — and "quota" is not in `cq.TASFlavors`.
I will drop this one

pkg/scheduler/flavorassigner/flavorassigner_test.go
mszadkow: I see it now, that's a mistake on my side.
I will work on this

pkg/scheduler/flavorassigner/flavorassigner_test.go
mszadkow: The only way psAssignment.Count could be 0 for a TAS PodSet would be if the workload spec itself declared Count: 0, which LWS validation prevents.
I will remove that

pkg/scheduler/flavorassigner/flavorassigner_test.go
mszadkow: Sure I created 2 PR for this:
https://github.com/kubernetes-sigs/kueue/pull/12676 - new skill pointing structured errors as review feedback
fixing existing errors in `tas_flavorassigner.go` - https://github.com/kubernetes-sigs/kueue/pull/12679

pkg/scheduler/flavorassigner/tas_flavorassigner.go
mszadkow: PodSet with zero resource requests ends up with empty Flavors map out of `assignFlavors`, regardless of what its same-group peers resolved to.
However I will follow o @mimowo suggestion to try to not loose the information on first place

mimowo: Just to clarify: I'm not saying the approach is wrong, but there is a significant complexity to the fix, and our test coverage for such cases is very limited, so I'm just careful about adding the complexity for the fix.

I think what would help with decision making is to answer:
1. is the handling of "no requests" the same as "requests=0"? I think in principle it should be, if it is not, then it is a smell to me about the approach (not necessarily a blocker but something which needs to be documented in PR description and tests) 
2. in case the handling differs, could you summarize the differences?
3. is the algorithm robust for different ordering of PodSets in the PodSetGroup? (my original comment, for which I still don't seem to have got an answer). I think it should be robust for any order, if it is not robut, then it is a small (again, not necessarily a blocker, but something that needs to be documented in PR description and tests)

pkg/scheduler/flavorassigner/tas_flavorassigner.go
```diff
@@ -85,6 +94,115 @@ func (a *Assignment) WorkloadsTopologyRequests(log logr.Logger, wl *workload.Inf
...
+		if peerAssignment.Count == 0 {
+			continue
+		}
+		if tasFlavor, err := onlyTASFlavor(peerAssignment.Flavors, cq.TASFlavors); err == nil {
+			peerFlavors.Insert(*tasFlavor)
+		}
+	}
+
+	if peerFlavors.Len() == 0 {
+		return nil, nil
```
dkaluza: I'm a little bit confused, is this non error situation?
Don't we expect to have one flavor from peer assgined?

pkg/scheduler/flavorassigner/tas_flavorassigner.go
```diff
@@ -85,6 +94,115 @@ func (a *Assignment) WorkloadsTopologyRequests(log logr.Logger, wl *workload.Inf
...
+	}
+	if peerFlavors.Len() == 1 {
+		list := sets.List(peerFlavors)
```
dkaluza: Why do we convert the set into the list?
Wouldn't PopAny be enough?
https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#Set.PopAny

pkg/scheduler/flavorassigner/tas_flavorassigner.go
```diff
@@ -85,6 +94,115 @@ func (a *Assignment) WorkloadsTopologyRequests(log logr.Logger, wl *workload.Inf
...
+		return &list[0], nil
+	}
+	list := sets.List(peerFlavors)
```
dkaluza: Maybe https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#Set.UnsortedList ?
Probably no reason to sort it just for the purpose of creating an error message.

pkg/scheduler/flavorassigner/tas_flavorassigner.go
```diff
@@ -85,6 +94,115 @@ func (a *Assignment) WorkloadsTopologyRequests(log logr.Logger, wl *workload.Inf
...
+
+	for _, rgIdx := range findRGIndicesByFlavor(cq, *inferredFlavor) {
+		rg := &cq.ResourceGroups[rgIdx]
+		for resourceName := range rg.CoveredResources {
+			if _, found := normalizedFlavors[resourceName]; found {
+				continue
+			}
+			if requests[resourceName] != 0 {
+				continue
+			}
```
dkaluza: Can you explain why this part has to be so complex?
Why we need this double for loop?

Isn't this the case that we should always have just one flavor assigned from our peer?
Cannot we just use the same assignments that this peer is using?

dkaluza: Isn't this the case that if any of this "skip" conditions occurred we wouldn't even need this peer assignment as we would have a covered resource and therefore a proper TAS assignment even before running this normalization function?

dkaluza (commented):
Went through the rest of code. I'm generally a little worried about the complexity of this code and the fact that by it we are probably only covering a single edge case - when peer podset has an assignment. As mentioned in the original issue, there might be also other edge cases that we should probably think about (e.g. what if there are no peers at all?)

In my opinion we should investigate the root cause, why none of the flavors was assigned to this podset and try to improve there. For example if this is cause because we always require a resource in this assignment maybe it would be a good idea to create an artificial resource entry in the assignment to just handle those in more systematic manner (remark I do not know if this is viable, one should investigate).

pkg/scheduler/flavorassigner/tas_flavorassigner.go
mszadkow: Can be use, I will do that actually

pkg/scheduler/flavorassigner/tas_flavorassigner.go
mszadkow: It is not as if we can't find anything in the neighbours it will fail anyway in the next steps.

pkg/scheduler/flavorassigner/tas_flavorassigner.go
mszadkow: Let me just answer @mimowo questions for completeness (even though solution will be different now):
1. No difference here, both will bring the same outcome.
3. I don't see order dependency as we scanned all PodSets and collected flavors into a set and then we expect single flavor.

I understand the complexity concern though.
I found simpler solution, instead of post-processing `psAssignment` we can make sure we infer `Flavors` for request-less podSet during the `assignFlavors()`.
Now we cover both scenarios for both grouped and un-grouped podSets, with the constraint that CQ must be single flavored.

Why is it better?
Less code, I think readability increased.
Not order dependent too.

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -821,6 +819,50 @@ func (a *FlavorAssigner) assignFlavors(log logr.Logger, counts []int32) Assignme
...
+// resolvePodSetFlavors returns the flavors podSet should be assigned, given the flavors
+// already resolved for its whole PodSet group (groupFlavors). Normally this is just
+// groupFlavors filtered down to the resources podSet itself requests. A PodSet requesting
+// none of the group's managed resources (e.g. an LWS leader) would otherwise end up with no
+// flavor and be rejected from TAS, so such a PodSet instead falls back to: the group's TAS
+// flavor(s) if it belongs to a topology group, or the ClusterQueue's sole flavor if the
+// ClusterQueue is unambiguous.
+func (a *FlavorAssigner) resolvePodSetFlavors(log logr.Logger, podSet indexedPodSet, groupFlavors ResourceAssignment) ResourceAssignment {
+	podSetFlavors := utilmaps.FilterKeys(groupFlavors, slices.Collect(maps.Keys(podSet.podSet.Requests)))
+	tr := a.wl.Obj.Spec.PodSets[podSet.originalIndex].TopologyRequest
```
dkaluza: do you need to go through the workload to get the topology request?
Isn't it just available as `podSet.podSet.TopologyRequest`?

Also can we rename the variable somehow `podset.podet` looks wrong... (Though I do not have better idea than `ipodSet` or `idxPodSet`)

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -821,6 +819,50 @@ func (a *FlavorAssigner) assignFlavors(log logr.Logger, counts []int32) Assignme
...
+		return podSetFlavors
+	}
+
+	if tr.PodSetGroupName != nil {
+		// A request-less PodSet in a topology group (e.g. an LWS leader) still needs a
+		// resolved TAS flavor so it can be placed; keep the group's TAS flavor(s) instead
+		// of filtering the group's resolution down to nothing.
+		podSetFlavors = tasFlavorsOnly(groupFlavors, a.cq.TASFlavors)
+		if len(podSetFlavors) > 0 {
+			log.V(3).Info("Inferred TAS flavor for request-less PodSet from topology group", "podSet", podSet.podSet.Name, "flavors", podSetFlavors)
```
dkaluza: nit: probably if this podset requests some resources not covered by the flavors this will still be the condition branch that we land in, so the message is inaccurate.

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -821,6 +819,50 @@ func (a *FlavorAssigner) assignFlavors(log logr.Logger, counts []int32) Assignme
...
+		if cqFlavors.Len() == 1 {
+			flavor, _ := cqFlavors.PopAny()
+			if a.cq.TASFlavors[flavor] != nil {
+				podSetFlavors = make(ResourceAssignment)
+				for _, rgIdx := range findRGIndicesByFlavor(a.cq, flavor) {
+					for resourceName := range a.cq.ResourceGroups[rgIdx].CoveredResources {
+						podSetFlavors[resourceName] = &FlavorAssignment{Name: flavor, Mode: Fit, TriedFlavorIdx: -1}
+					}
+				}
+				log.V(3).Info("Inferred TAS flavor for request-less PodSet from sole ClusterQueue flavor", "podSet", podSet.podSet.Name, "flavor", flavor)
```
dkaluza: nit: same as above "request-less" might be inaccurate.

dkaluza (commented):
generally the implementation IMO looks much more understandable right now, added a few nitpicks, I didn't go through the tests again - will try to do this tomorrow.

pkg/scheduler/flavorassigner/tas_flavorassigner.go
mimowo: ack, the implementation also now looks very different than before, and seems much easier to follow. 
So maybe the comment, aside from the guiding principle is now outdated.

pkg/scheduler/flavorassigner/flavorassigner.go
mszadkow: those are unfortunately different ones, the one from podSet lacks podSetGroupName, so workload is the only option

dkaluza: Ok, got it, thanks for the explanation.

pkg/scheduler/flavorassigner/flavorassigner_test.go
```diff
@@ -5652,3 +5653,264 @@ func TestIsNoFitDueToCapacityAndLimits(t *testing.T) {
...
+			},
+		},
+	}
+
+	_ = assignment.WorkloadsTopologyRequests(testr.New(t), wl, &cq)
+
+	if assignment.PodSets[0].Status.err == nil {
+		t.Fatalf("expected leader podSet error to be persisted")
+	}
+	if got, want := assignment.PodSets[0].Status.err.Error(), "required topology is not supported with ElasticJobsViaWorkloadSlices"; got != want {
```
dkaluza: nit: comparison with sentinel error would be better.

pkg/scheduler/flavorassigner/flavorassigner_test.go
```diff
@@ -5652,3 +5653,264 @@ func TestIsNoFitDueToCapacityAndLimits(t *testing.T) {
...
+}
+
+func TestTasFlavorsOnly(t *testing.T) {
+	tasFlavors := map[kueue.ResourceFlavorReference]*schdcache.TASFlavorSnapshot{"tas": {}}
+	in := ResourceAssignment{
+		"example.com/gpu":  {Name: "tas", Mode: Fit, TriedFlavorIdx: -1},
+		corev1.ResourceCPU: {Name: "quota", Mode: Fit, TriedFlavorIdx: -1},
+	}
+	want := ResourceAssignment{
+		"example.com/gpu": {Name: "tas", Mode: Fit, TriedFlavorIdx: -1},
```
dkaluza: minor: it would be better to have a couple of more edge cases tested here.
e.g. if this behaves  correctly if all flavors are in TasFlavors,
empty tasFlavors (not sure if it can occur in runtime without investigation)

pkg/scheduler/flavorassigner/flavorassigner_test.go
```diff
@@ -5652,3 +5653,264 @@ func TestIsNoFitDueToCapacityAndLimits(t *testing.T) {
...
+		utiltestingapi.MakeTopology("tas-topo-b").Levels(corev1.LabelHostname).Obj(),
+	}
+
+	cases := map[string]struct {
+		wlPods            []kueue.PodSet
+		clusterQueue      kueue.ClusterQueue
+		wantPodSetErrors  map[kueue.PodSetReference]error
+		wantPodSetFlavors map[kueue.PodSetReference]ResourceAssignment
+	}{
+		"leader without a managed request infers the group's TAS flavor": {
```
dkaluza, line 5679: minor: maybe it is worth to add a test case with multiple groups that leader infers the flavor from correct one?

pkg/scheduler/flavorassigner/flavorassigner.go
mszadkow: done

pkg/scheduler/flavorassigner/flavorassigner.go
mszadkow: done

pkg/scheduler/flavorassigner/tas_flavorassigner.go
mszadkow: thanks!

pkg/scheduler/flavorassigner/flavorassigner_test.go
```diff
@@ -5652,3 +5653,264 @@ func TestIsNoFitDueToCapacityAndLimits(t *testing.T) {
...
+					Obj(),
+				*utiltestingapi.MakePodSet("worker", 1).
+					Request("example.com/gpu", "1").
+					RequiredTopologyRequest(corev1.LabelHostname).
+					PodSetGroup("group1").
+					Obj(),
+			},
+			clusterQueue: *utiltestingapi.MakeClusterQueue("test-clusterqueue").
+				ResourceGroup(
+					*utiltestingapi.MakeFlavorQuotas("tas-a").
```
dkaluza, line 5693: Let's add a second flavor here. Otherwise current success may be a result of unambiguous assignment from the CQ

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -821,6 +819,50 @@ func (a *FlavorAssigner) assignFlavors(log logr.Logger, counts []int32) Assignme
...
+// groupFlavors filtered down to the resources podSet itself requests. A PodSet requesting
+// none of the group's managed resources (e.g. an LWS leader) would otherwise end up with no
+// flavor and be rejected from TAS, so such a PodSet instead falls back to: the group's TAS
+// flavor(s) if it belongs to a topology group, or the ClusterQueue's sole flavor if the
+// ClusterQueue is unambiguous.
+func (a *FlavorAssigner) resolvePodSetFlavors(log logr.Logger, podSet indexedPodSet, groupFlavors ResourceAssignment) ResourceAssignment {
+	podSetFlavors := utilmaps.FilterKeys(groupFlavors, slices.Collect(maps.Keys(podSet.podSet.Requests)))
+	tr := a.wl.Obj.Spec.PodSets[podSet.originalIndex].TopologyRequest
+	if len(podSet.podSet.Requests) != 0 || tr == nil {
+		return podSetFlavors
```
mimowo, line 879: Please log indicating what happens here. Use v2 for anomaly, use V4-V6 for BAU situations. The more common a log line might be the higher the level so that we avoid spamming.

pkg/scheduler/flavorassigner/flavorassigner.go
mimowo: Here I'm confused:
1. the log says "inferred" which is past tense indicating it is already inferred, however it is unclear what is the inference result - is it "podSetFlavors"? If so then indicate that by the log key, maybe "assignedFlavors". 
2. the log says "TAS flavor" - singular, but the results are plural, so I feel some discomfort if I understand well. 
3. if we claim we finished inferring then why not just return from here - that we avoid the risk of overring the results by the code below

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -821,6 +819,50 @@ func (a *FlavorAssigner) assignFlavors(log logr.Logger, counts []int32) Assignme
...
+// already resolved for its whole PodSet group (groupFlavors). Normally this is just
+// groupFlavors filtered down to the resources podSet itself requests. A PodSet requesting
+// none of the group's managed resources (e.g. an LWS leader) would otherwise end up with no
+// flavor and be rejected from TAS, so such a PodSet instead falls back to: the group's TAS
+// flavor(s) if it belongs to a topology group, or the ClusterQueue's sole flavor if the
+// ClusterQueue is unambiguous.
+func (a *FlavorAssigner) resolvePodSetFlavors(log logr.Logger, podSet indexedPodSet, groupFlavors ResourceAssignment) ResourceAssignment {
+	podSetFlavors := utilmaps.FilterKeys(groupFlavors, slices.Collect(maps.Keys(podSet.podSet.Requests)))
+	tr := a.wl.Obj.Spec.PodSets[podSet.originalIndex].TopologyRequest
+	if len(podSet.podSet.Requests) != 0 || tr == nil {
```
mimowo: Do we need the `tr == nil` check really, or this is just a safeguard? It seems to me that if we invoke the downstream logic for non-TAS workloads then it should be no-op

mimowo: If this is the case I would rather drop the safeguard to avoid extra "ifs" which split our testing into two paths, but no strong opinion here. I would just like to make sure the intention is reflected in the log / comment, becuase this is far from obvious.

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -821,6 +819,50 @@ func (a *FlavorAssigner) assignFlavors(log logr.Logger, counts []int32) Assignme
...
+	}
+	if len(podSetFlavors) == 0 {
+		// No group (or no group peer resolved a TAS flavor either). If the whole
+		// ClusterQueue only offers a single flavor, there is no ambiguity about which
+		// one a request-less, TAS-requesting PodSet should use.
+		cqFlavors := schdcache.AllFlavors(a.cq.ResourceGroups)
+		if cqFlavors.Len() == 1 {
+			flavor, _ := cqFlavors.PopAny()
+			if a.cq.TASFlavors[flavor] != nil {
+				podSetFlavors = make(ResourceAssignment)
```
mimowo: I'm confused here,, why do we allocate a new var here - we know from the previous check `if len(podSetFlavors) == 0 {` the already allocated var is empty, so why not just add to it?

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -821,6 +819,50 @@ func (a *FlavorAssigner) assignFlavors(log logr.Logger, counts []int32) Assignme
...
+				for _, rgIdx := range findRGIndicesByFlavor(a.cq, flavor) {
+					for resourceName := range a.cq.ResourceGroups[rgIdx].CoveredResources {
```
mimowo: Let's introduce a reuable helper `coveredResourcesForFlavor()` what would avoid mixing the abstraction layers - business logic  (flavors and resources) with low level index play (rgIdx).

pkg/scheduler/flavorassigner/flavorassigner.go
mszadkow: updated

pkg/scheduler/flavorassigner/flavorassigner.go
mszadkow: Well, it actually can be un-initialized here.
Line 830: podSetFlavors := utilmaps.FilterKeys(...) - returns an empty map when requests are empty, possibly nil
Line 844: We check if len(podSetFlavors) == 0 - this is true whether it's an empty map or nil

So we need to initialize it

pkg/scheduler/flavorassigner/flavorassigner.go
mszadkow: updated

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -821,6 +819,60 @@ func (a *FlavorAssigner) assignFlavors(log logr.Logger, counts []int32) Assignme
...
+				}
+			}
+		}
+	}
+	return podSetFlavors
+}
+
+// coveredResourcesForFlavor returns the set of all resources covered by the given flavor
+// across all resource groups that include that flavor in a ClusterQueue.
+func coveredResourcesForFlavor(cq *schdcache.ClusterQueueSnapshot, flavor kueue.ResourceFlavorReference) sets.Set[corev1.ResourceName] {
```
mimowo: review if we don't do something like that in other places of the codebase

pkg/scheduler/flavorassigner/flavorassigner.go
mimowo: Ah, ok, but this hand over of the variables makes it really hard to follow. I'm not yet sure constructively how to best eliminate that. Maybe we need to wrap the sections into indivudual functions, name them properly to indicate the intention.

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -821,6 +819,60 @@ func (a *FlavorAssigner) assignFlavors(log logr.Logger, counts []int32) Assignme
...
 }
 
+// resolvePodSetFlavors returns the flavors podSet should be assigned, given the flavors
+// already resolved for its whole PodSet group (groupFlavors). Normally this is just
+// groupFlavors filtered down to the resources podSet itself requests. A PodSet requesting
+// none of the group's managed resources (e.g. an LWS leader) would otherwise end up with no
+// flavor and be rejected from TAS, so such a PodSet instead falls back to: the group's TAS
+// flavor(s) if it belongs to a topology group, or the ClusterQueue's sole flavor if the
+// ClusterQueue is unambiguous.
+func (a *FlavorAssigner) resolvePodSetFlavors(log logr.Logger, idxPodSet indexedPodSet, groupFlavors ResourceAssignment) ResourceAssignment {
```
mimowo, line 867: Is the code here fully covered by unit tests?

pkg/scheduler/flavorassigner/flavorassigner_test.go
mszadkow: If I understand correctly "empty tasFlavors"  it can not happen:
```
rule="!has(self.topologyName) || self.nodeLabels.size() >= 1"
```

pkg/scheduler/flavorassigner/flavorassigner.go
mszadkow: I will try

pkg/scheduler/flavorassigner/flavorassigner.go
mszadkow:
```
sigs.k8s.io/kueue/pkg/scheduler/flavorassigner/flavorassigner.go:829:           resolvePodSetFlavors                    100.0%
```
 :)

pkg/scheduler/flavorassigner/flavorassigner.go
mszadkow: We have `allCoveredResources`, but `coveredResourcesForFlavor` only accumulate for resource groups selected by one flavor

pkg/scheduler/flavorassigner/flavorassigner_test.go
mszadkow: added

pkg/scheduler/flavorassigner/flavorassigner_test.go
mszadkow: added

pkg/scheduler/flavorassigner/flavorassigner.go
mszadkow: added

mszadkow:
@dkaluza can you ptal

mszadkow:
ptal @ivnovakov @dkaluza

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -821,6 +819,69 @@ func (a *FlavorAssigner) assignFlavors(log logr.Logger, counts []int32) Assignme
...
+	// Strategy 1: Filter group flavors to requested resources.
+	podSetFlavors := utilmaps.FilterKeys(groupFlavors, slices.Collect(maps.Keys(idxPodSet.podSet.Requests)))
+	if len(idxPodSet.podSet.Requests) != 0 {
```
dkaluza:
```suggestion
	if len(podSetFlavors) != 0 {
```
Probably result is the same for "normal" situations, but this should be more intuitive/bullet proof.

pkg/scheduler/flavorassigner/flavorassigner_test.go
```diff
@@ -5643,3 +5645,361 @@ func TestIsNoFitDueToCapacityAndLimits(t *testing.T) {
...
+				"leader2": {"example.com/gpu-b": {Name: "tas-b", Mode: Fit, TriedFlavorIdx: -1}},
+				"worker2": {"example.com/gpu-b": {Name: "tas-b", Mode: Fit, TriedFlavorIdx: -1}},
+			},
+		},
+		"leader in group where peer resolves to non-TAS flavor is rejected": {
+			// Exercises the Strategy 2 fallback path in resolvePodSetFlavors: tasFlavorsOnly
+			// returns empty because the group's resolved flavor is non-TAS, then Strategy 3
+			// also fails because the sole CQ flavor has no topology. The leader ends up with
+			// empty Flavors; since cq.TASFlavors is empty, podSetTopologyRequest returns
+			// ErrNoTASCacheInformation (checked before onlyTASFlavor).
```
dkaluza: I'm actually not sure this should be the correct behavior. If there is only one why don't we just return it no matter if it is tas or not... This would also make the code easier, the downside that I currently see is that it can block non-tas + tas situations, but I guess in such situation with ambiguity we are returning error anyway @mimowo wdyt?

dkaluza (commented):
Generally looks good to me added one question and one nitpick

dkaluza (approved):

pkg/scheduler/flavorassigner/flavorassigner_test.go
mszadkow: Thanks for the review, I will update the comment so it's more readable.
Are you referring to some specific scenario like mixed group TAS and non-TAS?

I think it won't pass anyway, because we later run `WorkloadsTopologyRequests` -> `podSetTopologyRequest` -> `onlyTASFlavor`
So even if an earlier step allowed a non-TAS flavor to survive, the later TAS request construction still enforces TAS-specific validity.

dkaluza: Hmmm, not sure if I understand the situation fully already.

I'm not even referring to a mixed scenario (the concern that I mentioned was that if we change the behavior we can affect the mixed-scenario), what will happen in fully non-tas scenario right now?

Let's imagine we have 2 podsets in one group - "group1" and CQ with only one non tas flavor.
Podset1 in the group has a request for resource managed by this flavor - e.g. "gpu"
Podset2 in the groupd does not have a request for the "gpu" resource.

What will be the result of the current code?
IIUC it will be rejected because the only resource is non-tas. (The inference from the group will fail)
And for me this looks suspicious/incorrect. Because why we permit the same situation if flavor is tas, and do not permit it if it is non-tas?

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -861,6 +853,73 @@ func (a *FlavorAssigner) assignFlavors(ctx context.Context, log logr.Logger, cou
...
+		// A PodSet with no resource requests in a topology group (e.g. an LWS leader) still needs a
+		// resolved TAS flavor so it can be placed; keep the group's TAS flavor(s) instead
+		// of filtering the group's resolution down to nothing.
+		podSetFlavors = tasFlavorsOnly(groupFlavors, a.cq.TASFlavors)
+		if len(podSetFlavors) > 0 {
+			log.V(5).Info("Using TAS flavors from topology group for PodSet with no resource requests", "podSet", idxPodSet.podSet.Name, "flavors", podSetFlavors)
+		}
+	}
+
+	// Strategy 3: Fallback to sole ClusterQueue flavor if unambiguous.
```
mimowo: I synced with @dkaluza  and I would like to drop Strategy 3 entirely - it is not needed to address the bug we are solving, and is introducing a significant complication. We can leave a comment that this is a possible future improvement if raised by users. However, I don't anticipate for now.

mszadkow: Ok, make sense I will update accordingly

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -865,6 +857,55 @@ func (a *FlavorAssigner) assignFlavors(ctx context.Context, log logr.Logger, cou
...
+	if groupName := podSetGroupName(&a.wl.Obj.Spec.PodSets[idxPodSet.originalIndex]); groupName != nil {
+		// A PodSet with no resource requests in a topology group (e.g. an LWS leader) still needs a
+		// resolved TAS flavor so it can be placed; keep the group's TAS flavor(s) instead
+		// of filtering the group's resolution down to nothing.
+		tasCoveredResources := sets.New[corev1.ResourceName]()
+		for resourceName, assignment := range groupFlavors {
+			tasFlavorName := assignment.Name
+			if _, isTAS := a.cq.TASFlavors[tasFlavorName]; !isTAS {
+				continue
+			}
```
mimowo: Don't we have a helper to select onlyTASFlavors already? Or the issue is that it fails? Maybe introduce FilterTASFlavors or something like this.

mszadkow: That's a different type of helper, but I have one that solves the problem

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -865,6 +857,42 @@ func (a *FlavorAssigner) assignFlavors(ctx context.Context, log logr.Logger, cou
...
+		reqKeys = append(reqKeys, name)
+	})
+	podSetFlavors := utilmaps.FilterKeys(groupFlavors, reqKeys)
+	if idxPodSet.podSet.Requests.Len() != 0 {
+		log.V(5).Info("Resolved PodSet flavors from group flavors",
+			"podSet", idxPodSet.podSet.Name,
+			"requestedResources", idxPodSet.podSet.Requests.Len(),
+			"resolvedFlavors", len(podSetFlavors))
+		return podSetFlavors
+	}
```
mimowo, line 880:
```suggestion
	if idxPodSet.podSet.Requests.Len() != 0 {
  	   // For PodSets with requests, keep only flavors for resources this PodSet requests.
       var reqKeys []corev1.ResourceName
  	   idxPodSet.podSet.Requests.ForEach(func(name corev1.ResourceName, _ int64) {
    		reqKeys = append(reqKeys, name)
     	})
    	podSetFlavors := utilmaps.FilterKeys(groupFlavors, reqKeys)
		log.V(5).Info("Resolved PodSet flavors from group flavors",
			"podSet", idxPodSet.podSet.Name,
			"requestedResources", idxPodSet.podSet.Requests.Len(),
			"resolvedFlavors", len(podSetFlavors))
		return podSetFlavors
	}
```
I might be missing something, but why not just like this, scoping `reqKeys` to where it is needed?

pkg/scheduler/flavorassigner/flavorassigner.go
```diff
@@ -865,6 +857,42 @@ func (a *FlavorAssigner) assignFlavors(ctx context.Context, log logr.Logger, cou
...
+
+	// For PodSets without requests, reuse TAS flavors from the topology group when available.
+	if groupName := podSetGroupName(&a.wl.Obj.Spec.PodSets[idxPodSet.originalIndex]); groupName != nil {
+		// A PodSet with no resource requests in a topology group (e.g. an LWS leader) still needs a
+		// resolved TAS flavor so it can be placed; keep the group's TAS flavor(s) instead
+		// of filtering the group's resolution down to nothing.
+		podSetFlavors = tasFlavorsOnly(groupFlavors, a.cq.TASFlavors)
+		if len(podSetFlavors) > 0 {
+			log.V(5).Info("Using TAS flavors from topology group for PodSet with no resource requests", "podSet", idxPodSet.podSet.Name, "flavors", podSetFlavors)
```
mimowo, line 889: why not just return from there? Leaving the final branch just return nil?

mszadkow: ah right, thanks

pkg/scheduler/flavorassigner/flavorassigner.go
mszadkow: I have followed your suggestion while slightly changing it

dkaluza (commented):
Please update the PR description as we no longer assign the single flavor from the CQ

pkg/scheduler/flavorassigner/flavorassigner_test.go
```diff
@@ -5649,3 +5652,329 @@ func TestIsNoFitDueToCapacityAndLimits(t *testing.T) {
...
+				// Verify both leaders inferred their flavors from their groups
+				"leader1": {"example.com/gpu-a": {Name: "tas-a", Mode: Fit, TriedFlavorIdx: -1}},
+				"worker1": {"example.com/gpu-a": {Name: "tas-a", Mode: Fit, TriedFlavorIdx: -1}},
+				"leader2": {"example.com/gpu-b": {Name: "tas-b", Mode: Fit, TriedFlavorIdx: -1}},
+				"worker2": {"example.com/gpu-b": {Name: "tas-b", Mode: Fit, TriedFlavorIdx: -1}},
+			},
+		},
+		"leader in group where peer resolves to non-TAS flavor is rejected": {
+			// Exercises the Strategy 2 fallback path in resolvePodSetFlavors: tasFlavorsOnly
+			// returns empty because the group's resolved flavor is non-TAS, then Strategy 3
```
dkaluza: nit: there is no strategy 3

Also it is probably best to refer to them by some names instead of numbers - today those are 1 and 2, but maybe we will have yet another one before them in the future

dkaluza (approved):
Code LGTM, just nitpick to fix a comment

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.19
/cherrypick release-0.18
Thank you for the fix  👍

mimowo:
@mszadkow please prepare the CPs if the robot fails

PR #12587, opened 2026-06-26
Treat remote and local preemption gates independently

### LINKED ISSUE #12543: Infinite Reconcile Loop DoS via Workload PreemptionGates in pkg/controller/admissionchecks/multikueue/workload.go

### Impact
A low-privileged tenant can induce a severe Denial of Service (DoS) against the Kueue control plane and the underlying Kubernetes API server. Because the Reconcile function successfully completes without returning an error after re-creating the remote workload, the controller-runtime's internal backoff rate limit is bypassed. The controller will infinitely delete and recreate the remote workload, consuming its own API Priority and Fairness (APF) quota, starving other legitimate reconciliations, and flooding the API server/etcd with write operations.

### Data Flow
- **pkg/controller/admissionchecks/multikueue/workload.go:537**: User-supplied WorkloadSpec is passed as `local`, containing populated `PreemptionGates`.
- **pkg/controller/admissionchecks/multikueue/workload.go:538**: The function nullifies `remote.PreemptionGates` to ignore them during comparison, but fails to nullify `local.PreemptionGates`.
- **pkg/controller/admissionchecks/multikueue/workload.go:539**: `equality.Semantic.DeepEqual(local, remote)` evaluates to false because `local` still contains PreemptionGates while `remote` does not, causing the function to always return true.
- **pkg/controller/admissionchecks/multikueue/workload.go:303**: Because the specs are falsely identified as out-of-sync, the controller deletes the remote Workload.
- **pkg/controller/admissionchecks/multikueue/workload.go:343**: The controller then immediately recreates the remote Workload in the same reconciliation pass, generating new API events that instantly trigger the loop again.

### Reproduction
**Preconditions**:
1. Kueue must be installed and configured with a MultiKueue AdmissionCheck and an active remote worker cluster.
2. The attacker needs RBAC permissions to create Workloads that route to the MultiKueue ClusterQueue.

**Steps**:
1. Create a Workload definition that explicitly defines `spec.preemptionGates` (e.g., a dummy gate).
2. Submit the Workload to a LocalQueue/ClusterQueue that utilizes the MultiKueue AdmissionCheck.
3. Monitor the Kueue controller manager logs and observe the tight infinite loop of deletions and creations.

### Vulnerable Code
```
func isRemoteSpecOutOfSync(local, remote kueue.WorkloadSpec) bool {
	remote.PreemptionGates = nil
	return !equality.Semantic.DeepEqual(local, remote)
}
```

### Remediation
Ensure that `PreemptionGates` are nullified on both the `local` and `remote` WorkloadSpec copies before performing the DeepEqual comparison.

Suggested Fix:
```
func isRemoteSpecOutOfSync(local, remote kueue.WorkloadSpec) bool {
	local.PreemptionGates = nil
	remote.PreemptionGates = nil
	return !equality.Semantic.DeepEqual(local, remote)
}
```

This logical root cause was found at **1** additional call site(s):

- `pkg/controller/jobframework/events.go` (Finding 13590: Infinite MultiKueue Workload Deletion Loop due to Faulty isRemoteSpecOutOfSync)

### Environment
* Install tools: N/A
* Others: N/A

kshalot:
#### What type of PR is this?
/kind bug
/area multikueue

Right now creating a MK workload with a preemption gate will cause an infinite loop of creations and deletions.

Fixes #12543

I posted more context about this in the issue here: https://github.com/kubernetes-sigs/kueue/issues/12543#issuecomment-4810859760. I posted this trivial fix since I already had the context.

```release-note
MultiKueue: Stop considering `spec.PreemptionGates` when syncing workloads. The preemption gates on the manager and worker clusters are treated independently - they are not copied from the manager to the workers and differences between them are not considered as out-of-sync. This fixes an issue where creating a MultiKueue workload with a preemption gate would cause an infinite loop of sync and deletions on the worker clusters.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated MultiKueue orchestrated preemption so preemption-gate changes no longer cause unnecessary remote workload updates or deletions.
  * Ensured worker-side preemption gates are established independently and aren’t copied from manager settings.
* **Tests**
  * Added coverage validating independent manager vs worker preemption-gate behavior.
  * Improved MultiKueue test cluster selection to reliably target the selected worker cluster across DRA and TAS scenarios.
  * Added shared test utilities for selected-worker client/context resolution.

### DIFF
--- a/pkg/controller/admissionchecks/multikueue/workload.go
+++ b/pkg/controller/admissionchecks/multikueue/workload.go
@@ -599,6 +599,9 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
 }
 
 func isRemoteSpecOutOfSync(local, remote kueue.WorkloadSpec) bool {
+	// Preemption gates should be treated independently on the remotes and manager,
+	// so any differences should not be considered as out-of-sync.
+	local.PreemptionGates = nil
 	remote.PreemptionGates = nil
 	return !equality.Semantic.DeepEqual(local, remote)
 }
@@ -1179,6 +1182,9 @@ func cloneForCreate(orig *kueue.Workload, origin string, preemptionGated bool) *
 	orig.Spec.DeepCopyInto(&remoteWl.Spec)
 
 	if features.Enabled(features.MultiKueueOrchestratedPreemption) && preemptionGated {
+		// Preemption gates should be treated independently on the remotes and manager,
+		// so we avoid copying them over to the remote.
+		remoteWl.Spec.PreemptionGates = nil
 		workload.EnsurePreemptionGateOnSpec(remoteWl, constants.MultiKueuePreemptionGate)
 	}
 
--- a/test/e2e/multikueue/sequential/e2e_test.go
+++ b/test/e2e/multikueue/sequential/e2e_test.go
@@ -771,6 +771,67 @@ var _ = ginkgo.Describe("MultiKueue Sequential", func() {
 			})
 		})
 
+		ginkgo.It("Should treat manager-level and worker-level preemption gates independently", func() {
+			job := testingjob.MakeJob("job-with-gates", managerNs.Name).
+				Queue(kueue.LocalQueueName(managerLq.Name)).
+				RequestAndLimit(corev1.ResourceCPU, "0.1").
+				RequestAndLimit(corev1.ResourceMemory, "0.1G").
+				TerminationGracePeriod(1).
+				Image(util.GetAgnHostImage(), util.BehaviorWaitForDeletion).
+				Obj()
+
+			ginkgo.By("Creating the job", func() {
+				util.MustCreate(ctx, k8sManagerClient, job)
+			})
+
+			wlLookupKey := types.NamespacedName{Name: workloadjob.GetWorkloadNameForJob(job.Name, job.UID), Namespace: managerNs.Name}
+
+			managerWl := &kueue.Workload{}
+			ginkgo.By("Waiting for the workload to be created", func() {
+				gomega.Eventually(func(g gomega.Gomega) {
+					g.Expect(k8sManagerClient.Get(ctx, wlLookupKey, managerWl)).To(gomega.Succeed())
+				}, util.Timeout, util.Interval).Should(gomega.Succeed())
+			})
+
+			dummyGateName := "dummy-gate"
+			ginkgo.By("Adding a dummy preemption gate to the workload", func() {
+				gomega.Eventually(func(g gomega.Gomega) {
+					g.Expect(k8sManagerClient.Get(ctx, wlLookupKey, managerWl)).To(gomega.Succeed())
+					managerWl.Spec.PreemptionGates = []kueue.PreemptionGate{{Name: dummyGateName}}
+					g.Expect(k8sManagerClient.Update(ctx, managerWl)).To(gomega.Succeed())
+				}, util.Timeout, util.Interval).Should(gomega.Succeed())
+			})
+
+			workerWl := &kueue.Workload{}
+			var workerClient client.Client
+			var initialUID types.UID
+			ginkgo.By("Waiting for the workload to be created on one of the workers", func() {
+				gomega.Eventually(func(g gomega.Gomega) {
+					if err := k8sWorker1Client.Get(ctx, wlLookupKey, workerWl); err == nil {
+						initialUID = workerWl.UID
+						workerClient = k8sWorker1Client
+					} else {
+						gomega.Expect(k8sWorker2Client.Get(ctx, wlLookupKey, workerWl)).To(gomega.Succeed())
+						initialUID = workerWl.UID
+						workerClient = k8sWorker2Client
+					}
+				}, util.Timeout, util.Interval).Should(gomega.Succeed())
+			})
+
+			ginkgo.By("Checking that the workload is not continuously recreated on workers", func() {
+				gomega.Consistently(func(g gomega.Gomega) {
+					g.Expect(workerClient.Get(ctx, wlLookupKey, workerWl)).To(gomega.Succeed())
+					g.Expect(workerWl.Spec.PreemptionGates).To(gomega.Not(gomega.ContainElement(kueue.PreemptionGate{Name: dummyGateName})))
+					g.Expect(workerWl.UID).To(gomega.Equal(initialUID))
+				}, util.ConsistentDuration, util.Interval).Should(gomega.Succeed())
+			})
+
+			ginkgo.By("Checking that the remote workload does not contain the manager workload's dummy gate", func() {
+				gomega.Expect(workerClient.Get(ctx, wlLookupKey, workerWl)).To(gomega.Succeed())
+				gomega.Expect(workerWl.Spec.PreemptionGates).To(gomega.Not(gomega.ContainElement(kueue.PreemptionGate{Name: dummyGateName})))
+			})
+		})
+
 		ginkgo.It("should not trigger concurrent preemptions", func() {
 			// Fits only in worker1
 			lowJob1 := testingjob.MakeJob("low-job1", managerNs.Name).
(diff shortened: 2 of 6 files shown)


pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -599,6 +599,9 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
 }
 
 func isRemoteSpecOutOfSync(local, remote kueue.WorkloadSpec) bool {
+	// Preemption gates should be treated independently on the remotes and manager,
+	// so any differences should not be considered as out-of-sync.
+	local.PreemptionGates = nil
```
mimowo, line 604: First of all I'm curious how this hasn't surfaced in our tests yet - busy loop creating / deleting a workload seems quite severe. IIUC this is only about pending workloads, but still curious. 

I guess we could add a check to our integration tests for Preemption Gates on MK and verify the UUID of the first observed workload, and the admitted are the same, wdyt?

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -1179,6 +1182,9 @@ func cloneForCreate(orig *kueue.Workload, origin string, preemptionGated bool) *
 	orig.Spec.DeepCopyInto(&remoteWl.Spec)
 
 	if features.Enabled(features.MultiKueueOrchestratedPreemption) && preemptionGated {
+		// Preemption gates should be treated independently on the remotes and manager,
+		// so we avoid copying them over to the remote.
+		remoteWl.Spec.PreemptionGates = nil
```
mimowo, line 1187: Maybe we could cover both palces with unit tests?

pkg/controller/admissionchecks/multikueue/workload.go
kshalot: We didn't have any MultiKueue tests where there are multiple preemption gates. We only had tests for the multi-kueue preemption gate.

pkg/controller/admissionchecks/multikueue/workload.go
kshalot: Yes, I'm looping agents right now to have an e2e repro and test.

test/e2e/multikueue/sequential/e2e_test.go
```diff
@@ -771,6 +771,67 @@ var _ = ginkgo.Describe("MultiKueue Sequential", func() {
 			})
 		})
 
+		ginkgo.It("Should treat manager-level and worker-level preemption gates independently", func() {
```
mimowo: Any reason this is not an integration test? Integration tests are generally more lightweight and this feature does not seem to require full  clusters running to test

kshalot: No, good point. I defaulted to an e2e test, but all the cross-cluster interactions we need can be achieved in the integration test. Changed.

test/integration/multikueue/scheduler/scheduler_test.go
```diff
@@ -539,6 +539,70 @@ var _ = ginkgo.Describe("MultiKueue with scheduler", ginkgo.Label("area:multikue
...
+						workerClient = worker1TestCluster.client
+						workerCtx = worker1TestCluster.ctx
+					} else {
+						gomega.Expect(worker2TestCluster.client.Get(worker2TestCluster.ctx, wlLookupKey, workerWl)).To(gomega.Succeed())
+						initialUID = workerWl.UID
+						workerClient = worker2TestCluster.client
+						workerCtx = worker2TestCluster.ctx
+					}
+				}, util.Timeout, util.Interval).Should(gomega.Succeed())
+			})
```
mimowo, line 594: Hm, why is the expectation that the workload is created on only one of the workers, whereas IIUC it will be created on both? 

If we are going to check what happens while the workload is pending, then we may need to make sure non of the two workload copies get admitted.

If we are going to check what happens when the workload is admitted, then we could wait to obtain the status.clusterName on the manager, and only check this worker. 

Maybe we want to test both situations, but then I would rather have two tests than overload one test with the tricky scenario.

kshalot: I felt that when testing the occurrence of the loop, just checking on one worker would be less flaky and easier to make pass when the issue is fixed.

But now that you raised it, I think we should just test admission, i.e. whether everything works as expected and not whether the loop occurs. PTAL: https://github.com/kubernetes-sigs/kueue/pull/12587/changes/08059ce3caa3e7869ee357636b20db62024b9893

test/integration/multikueue/scheduler/scheduler_test.go
```diff
@@ -539,6 +539,66 @@ var _ = ginkgo.Describe("MultiKueue with scheduler", ginkgo.Label("area:multikue
...
+					g.Expect(clusterName).ToNot(gomega.BeNil())
+					var workerClient client.Client
+					var workerCtx context.Context
+					if *clusterName == "worker1" {
+						workerClient = worker1TestCluster.client
+						workerCtx = worker1TestCluster.ctx
+					} else {
+						workerClient = worker2TestCluster.client
+						workerCtx = worker2TestCluster.ctx
+					}
```
mimowo: I remember I've seen similar code in the part, so I believe it would be useful for other places to have a helper like:
`workerCtx, workerClient := getClientForSelectedCluster(managerWl)`, wdyt?

test/integration/multikueue/scheduler/scheduler_test.go
```diff
@@ -539,6 +539,66 @@ var _ = ginkgo.Describe("MultiKueue with scheduler", ginkgo.Label("area:multikue
...
+
+					// When the gates are not treated independently, a mismatch between the gates will cause
+					// the workload to never get admitted and be continuously recreated.
+					// See: https://github.com/kubernetes-sigs/kueue/issues/12543
```
mimowo, line 590: Have you verified that experimentally locally?

kshalot: yes, I verified that manually to see whether the report is correct + now we have an integration test which fails if the fix is removed.

test/integration/multikueue/scheduler/scheduler_test.go
kshalot: Yeah, I initially did not bother because you have to pass a lot of data into this helper, but I quickly implemented it and it does look like an improvement - https://github.com/kubernetes-sigs/kueue/pull/12587/changes/3037d900f492def0f58ef024bb311a6fefa02697.

I found two other places where it can be re-used and applied it there as well. I can do that in a separate PR if you prefer, but it seemed like a small enough change to just keep here.

test/integration/multikueue/scheduler/scheduler_test.go
mimowo: Looks good but we could also have a wrapper which allows to avoid passing the struct with the clusters each time, because the struct will always look the same. So either introduce a helper util ClustersInfosForE2E() to avoid building the struct each time.

kshalot:
One test failed: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12587/pull-kueue-test-integration-multikueue-main/2071609407178805248
```
{ failed [FAILED] Expected
    <*string | 0x0>: nil
not to be nil
In [It] at: /home/prow/go/src/sigs.k8s.io/kueue/test/integration/multikueue/scheduler/scheduler_test.go:577 @ 06/29/26 15:09:21.584
}
```

Potentially flaky, I'll investigate.

test/util/multikueue.go
```diff
@@ -290,3 +290,45 @@ func ExpectWorkloadsToBeAdmittedAndGetWorkerName(ctx context.Context, k8sClient
...
 	return workerName
 }
+
+type ClusterInfo struct {
+	Name   string
+	Client client.Client
+	Ctx    context.Context
+}
+
+func ClusterInfosForE2E(
```
mimowo:
```suggestion
func DefaultClusterInfosForTests(
```
since used in both e2e and integration

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for fixing bugs 👍

PR #12581, opened 2026-06-26
Test temporarily use Full kuberay for presubmits

mimowo:
#### What type of PR is this?

/kind failing-test

#### Which issue(s) this PR fixes:

Fixes #12579

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

## Summary by CodeRabbit

* **Chores**
  * Updated non-periodic test jobs to use a more reliable test environment, helping keep presubmit checks running smoothly.

PR #12589, opened 2026-06-26
Fix ray-mini image by pinning Python 3.13 and re-enable for presubmits

### LINKED ISSUE #12588: Reenable using of miniray for presubmits

We need to reenable miniray for presubmits 

raymini was temporairly disabled to fix build issues: https://github.com/kubernetes-sigs/kueue/issues/12579

PR: https://github.com/kubernetes-sigs/kueue/pull/12581

**Why is this needed**:

to improve CI times for presubmits, as the fill kuberay is much heavier 2gb vs 200mb

sohankunkerkar:
#### What type of PR is this?
/kind bug

Python 3.14 ships with protobuf v34+ which removed `FieldDescriptor.label`. Ray 2.55.1 uses this API in `serve/_private/config.py`, causing RayService `DEPLOY_FAILED`. The fix exists on Ray main (ray-project/ray#64362) but is unreleased.

This PR pins the ray-mini image to Python 3.13 and reverts #12581 to re-enable raymini for presubmits.

Fixes #12588

```release-note
NONE
```

## Summary by CodeRabbit

* **Chores**
  * Updated the container base image to a newer, pinned Python runtime version.

### DIFF
--- a/hack/testing/ray-mini/Dockerfile
+++ b/hack/testing/ray-mini/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.14-slim@sha256:63a4c7f612a00f92042cbdcc7cdc6a306f38485af0a200b9c89de7d9b1607d15
+FROM python:3.13-slim@sha256:2b7445fb71ca9cb15e9aab053fe8cb3162796f8e1d92ada12a49c766a811bc1e
 
 # Set environment variable to avoid interactive prompts during package installation
 ENV DEBIAN_FRONTEND=noninteractive


hack/testing/ray-mini/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.14-slim@sha256:63a4c7f612a00f92042cbdcc7cdc6a306f38485af0a200b9c89de7d9b1607d15
+FROM python:3.13-slim
```
mimowo: Two questions:
1. why downgrading to 3.13, is 3.14 corrupted, or just incompatible with the KubeRay we use?
2. why dropping the sha256, I guess we could still use it for 3.13

Makefile-test.mk
```diff
@@ -84,10 +84,7 @@ GO_TEST_TARGET ?= .
 ifeq ($(JOB_TYPE),periodic)
     export USE_RAY_FOR_TESTS="ray"
 else
-	export USE_RAY_FOR_TESTS="ray"
-# Use full ray temporarily for presubmit jobs until we figure out the issue with raymini
-# See: https://github.com/kubernetes-sigs/kueue/issues/12579
-# export USE_RAY_FOR_TESTS="raymini"
+    export USE_RAY_FOR_TESTS="raymini"
```
mimowo: We may also add the build depenedcy to the makefile goal, because currently miniray.0.0.3 only built periodically, so this PR will not yet see the new version.

We have two options:
1. only fix the image (the other line) but enable raymini for presubmits in a follow up PR, or
2. enable building miniray on every PR (including this one)

hack/testing/ray-mini/Dockerfile
sohankunkerkar: 1. Not corrupted — incompatible. Python 3.14 ships with protobuf v34+ which [removed `FieldDescriptor.label`](https://protobuf.dev/news/2025-07-14/). Ray 2.55.1's `serve/_private/config.py` still uses `field.label`, causing the `DEPLOY_FAILED`. The fix landed on Ray main yesterday ([ray-project/ray#64362](https://github.com/ray-project/ray/pull/64362)) but isn't in any release yet. Python 3.13 avoids the issue since its protobuf doesn't have the removal.

2. Added it back — force-pushed with the digest pin.

Also dropped the Makefile-test.mk revert per your comment — will re-enable raymini for presubmits in a follow-up after the periodic job rebuilds the image.

mimowo: sgtm

mimowo:
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for the quick fix and clear explanations 👍

mimowo:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12589/pull-kueue-test-e2e-sequential-extended-shard-1-main/2070535290627297280
/test pull-kueue-test-e2e-sequential-extended-shard-1-main
I will open the issue

PR #12590, opened 2026-06-26
KueueViz: Fix missing http server timeouts

### LINKED ISSUE #12556: Denial of Service (DoS) via Missing HTTP Server Timeouts in cmd/kueueviz/backend/main.go

### Impact
An attacker can exhaust the server's resources (file descriptors, memory, goroutines) by keeping many connections open indefinitely (a Slowloris attack), leading to a Denial of Service (DoS) for legitimate users.

### Data Flow
- **cmd/kueueviz/backend/main.go:51**: The HTTP server is initialized without setting `ReadTimeout`, `WriteTimeout`, or `IdleTimeout`.
- **cmd/kueueviz/backend/main.go:81**: The server listens and serves requests using `srv.ListenAndServe()`, relying on the default Go `net/http` configuration which allows connections to remain open indefinitely.

### Reproduction
**Preconditions**:
The kueueviz backend server is running and accessible.

**Steps**:
1. Open a raw TCP connection to the server port.
2. Send a partial HTTP request (e.g., `GET / HTTP/1.1\r\nHost: localhost\r\n`) and keep the connection open without sending the terminating `\r\n\r\n`.
3. Open many such connections concurrently.

### Vulnerable Code
```
	srv := &http.Server{
		Addr:    serverConfig.GetServerAddress(),
		Handler: r.Handler(),
	}
```

### Remediation
Configure appropriate timeouts on the `http.Server` instance.

Suggested Fix:
```
srv := &http.Server{
	Addr:         serverConfig.GetServerAddress(),
	Handler:      r.Handler(),
	ReadTimeout:  10 * time.Second,
	WriteTimeout: 10 * time.Second,
	IdleTimeout:  120 * time.Second,
}
```

### Environment
* Install tools: N/A
* Others: N/A

ABHIGYAN-MOHANTA:
#### What type of PR is this?
/kind bug
/area dashboard

Configures appropriate timeouts on the http.Server instance for KueueViz Backend.

Fixes #12556 

```release-note
KueueViz: backend includes HTTP server timeouts (ReadHeaderTimeout, ReadTimeout, WriteTimeout, IdleTimeout) to prevent connection resource exhaustion.
```

## Summary of Changes

* **Bug Fixes**
  * Improved backend server reliability by adding explicit request and connection timeout settings (header, read, write, and idle) to avoid stalled requests and keep-alive connections lingering.
  * Enhanced WebSocket responsiveness by standardizing write timing and debouncing informer-driven updates, smoothing rapid event bursts and reducing unnecessary update floods.

### DIFF
--- a/cmd/kueueviz/backend/main.go
+++ b/cmd/kueueviz/backend/main.go
@@ -57,13 +57,25 @@ func main() {
 		os.Exit(1)
 	}
 
+	// The timeouts below are configured to prevent Denial-of-Service (DoS) attacks,
+	// such as Slowloris, where attackers can exhaust the server's resources by
+	// keeping many connections open indefinitely, leading to a Denial of Service
+	// (DoS) for legitimate users.
 	srv := &http.Server{
-		Addr:              serverConfig.GetServerAddress(),
-		Handler:           r.Handler(),
-		ReadHeaderTimeout: 3 * time.Second,
-		ReadTimeout:       10 * time.Second,
-		WriteTimeout:      10 * time.Second,
-		IdleTimeout:       120 * time.Second,
+		Addr:    serverConfig.GetServerAddress(),
+		Handler: r.Handler(),
+		// ReadHeaderTimeout is the amount of time allowed to read request headers.
+		// Set to 32s to align with Kubernetes and controller-runtime standards.
+		ReadHeaderTimeout: 32 * time.Second,
+		// ReadTimeout is the maximum duration for reading the entire request, including the body.
+		// Set to 120s to bound standard REST APIs (e.g., /api/*, /healthz). Websockets (e.g., /ws/*) are unaffected.
+		ReadTimeout: 120 * time.Second,
+		// WriteTimeout is the maximum duration before timing out writes of the response.
+		// Set to 120s to bound standard REST APIs (e.g., /api/*, /healthz). Websockets (e.g., /ws/*) are unaffected.
+		WriteTimeout: 120 * time.Second,
+		// IdleTimeout is the maximum amount of time to wait for the next request when keep-alives are enabled.
+		// Set to 90s to align with Kubernetes and controller-runtime standards.
+		IdleTimeout: 90 * time.Second,
 	}
 
 	h := handlers.New(handlers.NewClientFromManager(manager))
--- a/cmd/kueueviz/backend/main.go
+++ b/cmd/kueueviz/backend/main.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
 	"github.com/go-logr/logr"
 	"kueueviz/config"
@@ -57,8 +58,12 @@ func main() {
 	}
 
 	srv := &http.Server{
-		Addr:    serverConfig.GetServerAddress(),
-		Handler: r.Handler(),
+		Addr:              serverConfig.GetServerAddress(),
+		Handler:           r.Handler(),
+		ReadHeaderTimeout: 3 * time.Second,
+		ReadTimeout:       10 * time.Second,
+		WriteTimeout:      10 * time.Second,
+		IdleTimeout:       120 * time.Second,
 	}
 
 	h := handlers.New(handlers.NewClientFromManager(manager))
--- a/cmd/kueueviz/backend/handlers/websocket_handler.go
+++ b/cmd/kueueviz/backend/handlers/websocket_handler.go
@@ -31,6 +31,18 @@ import (
 	"kueueviz/middleware"
 )
 
+const (
+	// writeDeadlineExtension defines the duration added to the current time to establish
+	// an absolute write deadline on the underlying net.Conn before sending a WebSocket message.
+	// This ensures that the server does not block indefinitely when attempting to write
+	// to a slow or unresponsive client.
+	writeDeadlineExtension = 5 * time.Second
+
+	// debounceDelay is the duration to wait before triggering an update to the WebSocket client.
+	// This prevents overwhelming the client and server during rapid informer events (e.g. pod churn).
+	debounceDelay = 200 * time.Millisecond
+)
+
 // WebSocket upgrader
 var upgrader = websocket.Upgrader{
 	CheckOrigin: func(r *http.Request) bool {
@@ -100,7 +112,7 @@ func (h *Handlers) sendData(ctx context.Context, conn *websocket.Conn, dataFetch
 	}
 	slog.Debug("Data marshaled into JSON", "duration", time.Since(marshalStart))
 
-	if err := conn.SetWriteDeadline(time.Now().Add(5 * time.Second)); err != nil {
+	if err := conn.SetWriteDeadline(time.Now().Add(writeDeadlineExtension)); err != nil {
 		return err
 	}
 
@@ -121,7 +133,7 @@ func (h *Handlers) handleInformerUpdates(ctx context.Context, conn *websocket.Co
 	updateChan := make(chan struct{}, 1)
 
 	// Debounce rapid informer events (e.g. pod churn) into a single update
-	debounced := debounce.New(200 * time.Millisecond)
+	debounced := debounce.New(debounceDelay)
 	triggerUpdate := func() {
 		debounced(func() {
 			select {


mimowo:
/retitle KueueViz: Fix missing http server timeouts
please also add a release note since this is considered a bugfix
cc @olekzabl ptal

cmd/kueueviz/backend/main.go
```diff
@@ -57,8 +58,12 @@ func main() {
...
 
 	srv := &http.Server{
-		Addr:    serverConfig.GetServerAddress(),
-		Handler: r.Handler(),
+		Addr:              serverConfig.GetServerAddress(),
+		Handler:           r.Handler(),
+		ReadHeaderTimeout: 3 * time.Second,
+		ReadTimeout:       10 * time.Second,
+		WriteTimeout:      10 * time.Second,
+		IdleTimeout:       120 * time.Second,
```
mimowo: Please add a comment describing the intention for setting the timeouts - DoS prevention.
Also, add comments to the indivudual comments describing the timeouts briefly and proving example of API calls which might be affected by the timeouts (if feasible) .

Also, it would be good if you can find some industry standard for setting the timeouts, maybe controller-runtime, k8s itself,  or other projects spawning servers, so that we get a better grasp on the standards here.

ABHIGYAN-MOHANTA: > Please add a comment describing the intention for setting the timeouts - DoS prevention. Also, add comments to the indivudual comments describing the timeouts briefly and proving example of API calls which might be affected by the timeouts (if feasible) .
> 
> Also, it would be good if you can find some industry standard for setting the timeouts, maybe controller-runtime, k8s itself, or other projects spawning servers, so that we get a better grasp on the standards here.

Thanks @mimowo! Did some digging into how other projects handle this.

Turns out both `kubernetes` (apiserver) and `controller-runtime`'s internal HTTP server only set two things globally: `ReadHeaderTimeout` (32s) and `IdleTimeout` (90s).

- k8s: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/secure_serving.go#L175
- controller-runtime: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/httpserver/server.go#L16

They deliberately skip `ReadTimeout`/`WriteTimeout` globally, because setting those on `http.Server` puts a hard deadline on the connection, which breaks long-lived hijacked connections like watches, execs, and websockets. Since Kueueviz has a bunch of websocket routes, this matters for us too.

If we still want to set `ReadTimeout`/`WriteTimeout` to keep the security scanner happy, a few other projects give us some reference points:

- Traefik uses 60s pretty broadly
  https://github.com/traefik/traefik/blob/master/pkg/config/static/static_config.go#L50-L51
- Prometheus goes up to 5m for its web UI/API, to allow for big queries
  https://github.com/prometheus/prometheus/blob/main/cmd/prometheus/main.go#L441

For us, the REST endpoints (`/api/:resourceType/:name`, `/healthz`, `/auth/status`, etc.) are fast and won't be affected either way. The websocket ones are the concern. There are 14 of them:

- `/ws/namespaces`
- `/ws/workloads`
- `/ws/workloads/dashboard`
- `/ws/workload/:namespace/:workload_name`
- `/ws/workload/:namespace/:workload_name/events`
- `/ws/local-queues`
- `/ws/local-queue/:namespace/:queue_name`
- `/ws/local-queue/:namespace/:queue_name/workloads`
- `/ws/cluster-queues`
- `/ws/cluster-queue/:cluster_queue_name`
- `/ws/cohorts`
- `/ws/cohort/:cohort_name`
- `/ws/resource-flavors`
- `/ws/resource-flavor/:flavor_name`

With a global timeout, all of these would get force-disconnected on that interval. 10s would mean constant reconnects on the frontend, so that's out.

My proposal: set both to 60s (matching Traefik) and fix `websocket_handler.go` so it clears the inherited read deadline right after the upgrade, so the websocket connections aren't affected. 

*(Note: We only need to clear the **ReadDeadline** because Kueueviz [already manually overrides the WriteDeadline](https://github.com/kubernetes-sigs/kueue/blob/main/cmd/kueueviz/backend/handlers/websocket_handler.go#L104) every single time it sends a message, which safely bypasses the global `WriteTimeout`).*

```go
// In GenericWebSocketHandler after upgrading:
conn, err := upgrader.Upgrade(c.Writer, c.Request, nil)
if err != nil {
    slog.Debug("Failed to upgrade to WebSocket", "error", err)
    return
}
// Clear the inherited ReadDeadline so websockets stay alive indefinitely
conn.UnderlyingConn().SetReadDeadline(time.Time{})
```

Happy to go this route, or just drop the global `ReadTimeout`/`WriteTimeout` entirely like k8s/controller-runtime do, if you'd rather keep it simple. Let me know which you prefer.

mimowo: > Turns out both kubernetes (apiserver) and controller-runtime's internal HTTP server only set two things globally: ReadHeaderTimeout (32s) and IdleTimeout (90s).

We can follow them here.

> They deliberately skip ReadTimeout/WriteTimeout globally, because setting those on http.Server puts a hard deadline on the connection, which breaks long-lived hijacked connections like watches, execs, and websockets. Since Kueueviz has a bunch of websocket routes, this matters for us too.

Are websocket connections interrutpted actually by the `ReadTimeout` or `WriteTimeout`, or both? When they are disconnected, then are they re-established by frontend immediately under the hood? 

> (Note: We only need to clear the ReadDeadline because Kueueviz already manually overrides the WriteDeadline every single time it sends a message, which safely bypasses the global WriteTimeout).

Are ReadDeadline / WriteDeadline same as ReadTimeout / WriteTimeout, what is the difference if not?

Also, I would like to extract the deadline from [here](https://github.com/kubernetes-sigs/kueue/blob/fc4a09f2788bb1282ee4b19cdaa355e23dce446d/cmd/kueueviz/backend/handlers/websocket_handler.go#L104) to a const with a comment.

> My proposal: set both to 60s (matching Traefik) and fix websocket_handler.go so it clears the inherited read deadline right after the upgrade, so the websocket connections aren't affected.

Maybe between Traefik and Prometheus - 2min?

cc @danylott

ABHIGYAN-MOHANTA: Thanks @mimowo! 

> Are websocket connections interrutpted actually by the `ReadTimeout` or `WriteTimeout`, or both? When they are disconnected, then are they re-established by frontend immediately under the hood?

I looked into this, tested by running it locally, and surprisingly, neither of them actually interrupt the websockets in our case. Thanks for catching this!

While it's true that setting a global `ReadTimeout` in raw `net/http` breaks hijacked connections, we're using the `gorilla/websocket` library here. It turns out `gorilla/websocket` automatically clears any deadlines set by the HTTP server during the upgrade process:
[gorilla/websocket/blob/v1.5.0/server.go#L250-L251](https://github.com/gorilla/websocket/blob/v1.5.0/server.go#L250-L251)

So we can safely set the global timeouts without breaking the websockets. We don't even need the manual `conn.UnderlyingConn().SetReadDeadline(time.Time{})` fix I was considering earlier.

> Are ReadDeadline / WriteDeadline same as ReadTimeout / WriteTimeout, what is the difference if not?

They are closely related but operate at different layers. `ReadTimeout` / `WriteTimeout` are high-level configuration properties of the Go `http.Server`. `ReadDeadline` / `WriteDeadline` are the low-level, absolute timestamps applied directly to the TCP socket (`net.Conn`). When the `http.Server` accepts a connection, it simply takes your configured `ReadTimeout` and sets a hard socket deadline via `conn.SetReadDeadline(...)`. Similarly for `WriteTimeout` and `WriteDeadline`.

> Also, I would like to extract the deadline from here to a const with a comment.
> Maybe between Traefik and Prometheus - 2min?

Sounds good to me. I have pushed a commit addressing these. Let me know if we decide on something else for the timeouts or if anything else requires changes! I will be happy to make the changes.

Only the REST APIs (`/api/:resourceType/:name`, `/healthz`, and `/auth/status`) will be impacted, since websockets clear the deadline.

Thank you for reviewing!

mimowo: Thank you for checking all of that, summarizing here and in comments. This is very useful for reviewing 👍

mimowo:
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for hardening KueueViz with the connection control 👍

PR #12594, opened 2026-06-26
Reject changing queueName on admitted workload

### LINKED ISSUE #12552: CEL validation bypass allows mutation of immutable Workload fields in apis/kueue/v1beta2/workload_types.go

### Impact
Users can bypass intended field immutability on workloads that are actively admitted, altering their logical association with queues and priority classes without triggering expected re-admission or rejection.

### Data Flow
- **apis/kueue/v1beta2/workload_types.go:1118**: User patches an Admitted workload by removing the 'queueName' field. Because 'has(self.spec.queueName)' evaluates to false, the ternary operator evaluates to 'true'.
- **apis/kueue/v1beta2/workload_types.go:1118**: User patches the workload again, adding a different 'queueName'. Because 'has(oldSelf.spec.queueName)' evaluates to false, the ternary evaluates to 'true', successfully bypassing the immutability check.

### Reproduction
**Preconditions**:
A workload is admitted (QuotaReserved is True).

**Steps**:
1. Submit a JSON patch to the workload removing the 'queueName' field: `[{"op": "remove", "path": "/spec/queueName"}]`
2. Submit a second JSON patch adding a new 'queueName': `[{"op": "add", "path": "/spec/queueName", "value": "new-queue"}]`

### Vulnerable Code
```
has(oldSelf.spec.queueName) && has(self.spec.queueName) ? oldSelf.spec.queueName == self.spec.queueName : true
```

### Remediation
Use default values for missing fields during comparison instead of skipping the comparison.

Suggested Fix:
```
Change the CEL rules to emulate the maximumExecutionTimeSeconds rule. E.g., `((has(oldSelf.spec.queueName)?oldSelf.spec.queueName:"") == (has(self.spec.queueName)?self.spec.queueName:""))`
```

### Environment
* Install tools: N/A
* Others: N/A

kannon92:
Fix: replace the escape-hatch ternary with the same default-value pattern already used by the maximumExecutionTimeSeconds rule, treating an absent field as the empty string on both sides:

With this change:
- Removing queueName from an admitted workload is rejected because 'q1' != ''.
- Adding a new queueName to an admitted workload that has none is rejected because '' != 'q2'.
- Legitimate mutations (e.g. resetting admission before a queue change) are unaffected because the outer QuotaReserved guard still allows changes when the workload is not admitted.

#### What type of PR is this?
/kind bug

Fixes #12552 

```release-note
Workloads: Fixed a validation bug that allowed the `queueName` of a quota-reserved Workload to be changed by first removing and then re-adding the field, bypassing its immutability constraint. Kueue now rejects adding, removing, or changing `queueName` while the Workload's quota remains reserved.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened `Workload.spec.queueName` immutability enforcement when workload quota is reserved (`QuotaReserved=True`), including cases where the field is omitted/added between update steps; validation now requires consistent presence and matching values.
  * Improved consistency of the immutability rule across supported workload CRD versions and charts.
* **Tests**
  * Added a regression test covering an attempted two-step update to ensure `queueName` immutability cannot be bypassed when quota is reserved.


test/integration/singlecluster/webhook/core/workload_test.go
```diff
@@ -1031,6 +1031,43 @@ var _ = ginkgo.Describe("Workload validating webhook", func() {
...
+			util.MustCreate(ctx, k8sClient, workloadNoQueue)
+			util.SetQuotaReservation(ctx, k8sClient, client.ObjectKeyFromObject(workloadNoQueue), utiltestingapi.MakeAdmission("cq").Obj())
+			gomega.Eventually(func(g gomega.Gomega) {
+				var wl kueue.Workload
+				g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(workloadNoQueue), &wl)).To(gomega.Succeed())
+				wl.Spec.QueueName = "queue2" // has(oldSelf.spec.queueName)==false in the old rule → true (bypass)
+				g.Expect(k8sClient.Update(ctx, &wl)).Should(utiltesting.BeInvalidError())
+			}, util.Timeout, util.Interval).Should(gomega.Succeed())
+		})
+
```
mimowo: I think the step (1.) should be enough. I believe for step (2) we should already have integration tests

kannon92: You are correct. Going ahead to remove.

kannon92: removed.

kannon92:
@mimowo 

there was a minor verify issue so I had to push up a fix for the linter.

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Awesome, thank you for hardening Kueue 👍

mimowo:
/release-note-edit
```release-note
Workloads: Fixed a validation bug that allowed the `queueName` of a quota-reserved Workload to be changed by first removing and then re-adding the field, bypassing its immutability constraint. Kueue now rejects adding, removing, or changing `queueName` while the Workload's quota remains reserved.
```

PR #12605, opened 2026-06-26
Re-enable raymini for presubmit e2e tests

sohankunkerkar:
#### What type of PR is this?
/kind cleanup

Fixes #

I went with option 2 from https://github.com/kubernetes-sigs/kueue/pull/12589#discussion_r3482520508

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added support for a lighter Ray test image in non-periodic test runs.
  * Some end-to-end tests are now marked to require the full Ray image, improving test selection.

* **Bug Fixes**
  * End-to-end runs now skip full-Ray-only tests when using the mini image, reducing failures from unavailable images.

* **Chores**
  * Updated the default Ray mini image version used in builds.

### DIFF
--- a/Makefile-test.mk
+++ b/Makefile-test.mk
@@ -96,11 +96,12 @@ endif
 ifeq ($(JOB_TYPE),periodic)
     export USE_RAY_FOR_TESTS="ray"
 else
-	export USE_RAY_FOR_TESTS="ray"
-# Use full ray temporarily for presubmit jobs until we figure out the issue with raymini
-# See: https://github.com/kubernetes-sigs/kueue/issues/12579
-# export USE_RAY_FOR_TESTS="raymini"
+	export USE_RAY_FOR_TESTS="raymini"
 endif
+
+# When using raymini, exclude tests that require the full ray image (e.g. RayService).
+# Workaround until upstream Ray fixes protobuf 7.35+ compatibility (ray-project/ray#64362).
+FULLRAY_EXCLUDE := $(if $(filter "raymini",$(USE_RAY_FOR_TESTS)), && !requires:fullray)
 
 .PHONY: test
 test: gotestsum ## Run tests. Set UNIT_TOTAL_SHARDS and UNIT_SHARD_INDEX to run a specific shard.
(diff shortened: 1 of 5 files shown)


hack/testing/e2e-common.sh
```diff
@@ -613,7 +613,9 @@ function prepare_docker_images {
     if [[ -n ${KUBERAY_VERSION:-} && ("$GINKGO_ARGS" =~ feature:kuberay || ! "$GINKGO_ARGS" =~ "--label-filter") ]]; then
         e2e_docker_pull_if_needed "${KUBERAY_IMAGE}"
         determine_kuberay_ray_image
-        e2e_docker_pull_if_needed "${KUBERAY_RAY_IMAGE}"
+        if [[ "${USE_RAY_FOR_TESTS:-}" == "ray" ]]; then
```
mimowo, line 616: Hm I think the logic should be: for 'ray' pull the image, for miniray check if the image exists ready to pull, if so then use. Otherwise build it locally. Because most of the time the image will already be pushed and ready by the periodic job. If we have this logic then we could solve the today's incident just by fixing the dependency and bumping to 0.0.4

sohankunkerkar: I added some code. Let me know if that addresses your concern.

Makefile-test.mk
```diff
@@ -84,12 +84,13 @@ GO_TEST_TARGET ?= .
...
-	export USE_RAY_FOR_TESTS="ray"
-# Use full ray temporarily for presubmit jobs until we figure out the issue with raymini
-# See: https://github.com/kubernetes-sigs/kueue/issues/12579
-# export USE_RAY_FOR_TESTS="raymini"
+	export USE_RAY_FOR_TESTS="raymini"
 endif
 
+# When using raymini, exclude tests that require the full ray image (e.g. RayService).
+# Workaround until upstream Ray fixes protobuf 7.35+ compatibility (ray-project/ray#64362).
+FULLRAY_EXCLUDE := $(if $(filter "raymini",$(USE_RAY_FOR_TESTS)), && !requires:fullray)
```
mimowo: Why would the RayService require full ray? I think we didn't use that before and it was ok.

sohankunkerkar: RayService worked with raymini before because protobuf 7.35 hadn't been released yet. Now pip install pulls it by default and it crashes Ray Serve (FieldDescriptor.label removed).  I tried pinning protobuf<7.35 in an earlier push and the DEPLOY_FAILED error went away, but RayService still timed out at 300s in that run. It could've been a slow CI node. Pushing the pin again to give it another shot.

mimowo: Do you see some path forward which would allow us to use miniray for RayService in the long term?

sohankunkerkar: Short-term: no as protobuf pin fixed the DEPLOY_FAILED crash but RayService still times out with raymini (likely missing system deps from the slim base that the full ray image has). So we need the requires:fullray skip for now. Periodic still runs them with full ray so coverage isn't lost.

mimowo:
/lgtm 
/approve
/cherry-pick release-0.18
/cherry-pick release-0.17
Thank you for the cleanup, let's see how it works in practice. I'm not sure it makes sense to maintain miniray given the complications but we can try for a little longer

Makefile-test.mk
mimowo: In that case I'm not sure about the benefits of miniray any longer if this is only to benefit RayJob and Ray cluster. If a suite uses both RayJob and Ray service then both images are now pulled anyway?

sohankunkerkar: The main benefit is presubmits don't have to pull the 2GB ray image. raymini builds locally in ~30s. RayJob and RayCluster tests still cover the Kueue integration logic, RayService is the only gap and periodic catches it. But if you'd rather just use full ray for everything and keep full coverage in presubmits, I'm fine dropping this. Thoughts?

PR #12612, opened 2026-06-27
Fix MultiKueue remote client data race during reconnect

### LINKED ISSUE #12557: Denial of Service via Data Race on MultiKueue Remote Client in pkg/controller/admissionchecks/multikueue/multikueuecluster.go

### Impact
A data race on the `client.WithWatch` interface causes a runtime panic, leading to a Denial of Service (controller crash loop). This can be triggered intentionally by a privileged user or occur accidentally during normal operations (e.g., network blips causing automatic reconnection while the GC runs).

### Data Flow
- **pkg/controller/admissionchecks/multikueue/multikueuecluster.go:679**: A background goroutine running `c.runGC` calls `getRemoteClients()`, which locks `c.lock`, returns a copy of the clients slice, and releases `c.lock`.
- **pkg/controller/admissionchecks/multikueue/multikueuecluster.go:292**: The background goroutine iterates the slice and calls `rc.runGC()`, which directly reads the `rc.client` interface to call `.List(ctx, wls, ...)`.
- **pkg/controller/admissionchecks/multikueue/multikueuecluster.go:175**: Concurrently, a Reconcile loop (triggered by a kubeconfig change or network watch reconnect) calls `rc.setConfig()`, which directly writes to `rc.client` without synchronization on `rc`.

### Reproduction
**Preconditions**:
Kueue controller is running with a configured MultiKueueCluster pointing to a remote cluster.

**Steps**:
1. Configure a MultiKueueCluster object pointing to a valid kubeconfig Secret.
2. Configure Kueue's GCInterval to a very small value to increase the race window, or populate many Workloads so `runGC` takes longer.
3. Continuously update the kubeconfig Secret or simulate network disconnects to trigger `rc.setConfig` via the reconciler.
4. Observe the controller manager crash with a Go runtime panic (data race or invalid memory address) when reading the `rc.client` interface.

### Vulnerable Code
```
rc.client = remoteClient
```

### Remediation
Introduce a mutex in `remoteClient` to protect read/write access to `rc.client` and other configuration state.

Suggested Fix:
```
Add a `sync.RWMutex` to the `remoteClient` struct. Acquire `rc.lock.Lock()` in `setConfig` when assigning `rc.client`, and acquire `rc.lock.RLock()` in `runGC` when reading `rc.client`.
```

### Environment
* Install tools: N/A
* Others: N/A

apullo777:
#### What type of PR is this?

/kind bug
/area multikueue

`remoteClient.client` can be replaced by `setConfig` during reconnects or kubeconfig updates while other goroutines read it from the GC loop and the workload/clusterqueue paths. Since this is an unsynchronized read/write of an interface value, a reader may observe an inconsistent client and crash the controller manager.

This PR guards the client swap with the existing `remoteClient.mu` and routes concurrent readers through `getClient()`, which snapshots the client under `RLock`. Reads that happen synchronously inside `setConfig` remain direct.

Also add race-detector coverage for the remote client readers. AI assistance was used for parts of this change.

Fixes #12557

```release-note
MultiKueue: Fixed a data race where reconnecting a remote cluster could swap the remote client while other goroutines were reading it, which could crash-loop the controller manager.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience of remote reconciliation during configuration refreshes by synchronizing access to the remote Kubernetes client.
  * Remote queue and workload operations now consistently use the most up-to-date available connection, with safe handling when it’s not yet available.

* **Tests**
  * Added race-focused concurrency regression coverage that exercises concurrent configuration refresh with remote garbage collection and remote read/list operations.

### DIFF
--- a/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
+++ b/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
@@ -267,7 +267,10 @@ func (rc *remoteClient) setConfig(watchCtx context.Context, config *clientConfig
 		return rc.increaseFailedConnAttempt(), err
 	}
 
+	// Lock the swap so a concurrent getClient() never reads a torn rc.client. See #12557.
+	rc.mu.Lock()
 	rc.client = remoteClient
+	rc.mu.Unlock()
 
 	err = rc.startWatcher(watchCtx, kueue.GroupVersion.WithKind("Workload").GroupKind().String(), &workloadKueueWatcher{})
 	if err != nil {
@@ -444,8 +447,12 @@ func (rc *remoteClient) startQueueWatchers(ctx context.Context) error {
 func (rc *remoteClient) queueEventsForCQ(ctx context.Context, remoteCQ *kueue.ClusterQueue) {
 	log := ctrl.LoggerFrom(ctx).WithValues("remoteCQ", remoteCQ.Name)
 
+	remoteCl := rc.getClient()
+	if remoteCl == nil {
+		return
+	}
 	lqList := kueue.LocalQueueList{}
-	err := rc.client.List(ctx, &lqList, client.MatchingFields{indexer.QueueClusterQueueKey: remoteCQ.Name})
+	err := remoteCl.List(ctx, &lqList, client.MatchingFields{indexer.QueueClusterQueueKey: remoteCQ.Name})
 	if err != nil {
 		log.Error(err, "Failed to list remote LocalQueues from cache")
 		return
@@ -481,6 +488,14 @@ func (rc *remoteClient) getRetryConnNextAttempt() metav1.Time {
 	rc.mu.RLock()
 	defer rc.mu.RUnlock()
 	return rc.retryConnNextAttempt
+}
+
+// getClient reads rc.client under the read lock so callers never observe a torn
+// value while setConfig swaps it on a reconnect. See #12557.
+func (rc *remoteClient) getClient() SelectivelyCachingClient {
+	rc.mu.RLock()
+	defer rc.mu.RUnlock()
+	return rc.client
 }
 
 func (rc *remoteClient) StopWatchers() {
@@ -531,8 +546,13 @@ func (rc *remoteClient) runGC(ctx context.Context) {
 		return
 	}
 
+	remoteCl := rc.getClient()
+	if remoteCl == nil {
+		return
+	}
+
 	wls := &kueue.WorkloadList{}
-	err := rc.client.List(ctx, wls, client.MatchingLabels{kueue.MultiKueueOriginLabel: rc.origin})
+	err := remoteCl.List(ctx, wls, client.MatchingLabels{kueue.MultiKueueOriginLabel: rc.origin})
 	if err != nil {
 		log.Error(err, "Listing remote workloads")
 		return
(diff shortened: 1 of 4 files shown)


pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -382,14 +382,14 @@ func (w *wlReconciler) reconcileGroup(ctx context.Context, group *wlGroup) (reco
...
-			if _, err := jobframework.ValidateRemoteObjectOwnership(ctx, group.remoteClients[remote].client, group.controllerKey, group.jobAdapter.GVK(), w.origin); err != nil {
+			if _, err := jobframework.ValidateRemoteObjectOwnership(ctx, group.remoteClients[remote].getClient(), group.controllerKey, group.jobAdapter.GVK(), w.origin); err != nil {
 				log.Error(err, "validating remote controller object", "workerCluster", remote)
 				return reconcile.Result{}, err
 			}
 
 			// The deferred flag is irrelevant here: the remote workload is
 			// Finished, so determineStatusUpdate always syncs (never defers).
-			if _, err := group.jobAdapter.SyncJob(ctx, w.client, group.remoteClients[remote].client, group.controllerKey, group.local.Name, w.origin); err != nil {
+			if _, err := group.jobAdapter.SyncJob(ctx, w.client, group.remoteClients[remote].getClient(), group.controllerKey, group.local.Name, w.origin); err != nil {
```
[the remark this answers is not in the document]
apullo777: Thanks, good point.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
```diff
@@ -267,7 +267,10 @@ func (rc *remoteClient) setConfig(watchCtx context.Context, config *clientConfig
 		return rc.increaseFailedConnAttempt(), err
 	}
 
+	// Lock the swap so a concurrent getClient() never reads a torn rc.client. See #12557.
+	rc.mu.Lock()
 	rc.client = remoteClient
+	rc.mu.Unlock()
```
mimowo: I would like to wrap this in a simple `secConfig` setter-like function. We should rename the pre-existing function as it is doing so much more, maybe `updateConfigAndRefreshWatchers`

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
```diff
@@ -444,8 +447,12 @@ func (rc *remoteClient) startQueueWatchers(ctx context.Context) error {
 func (rc *remoteClient) queueEventsForCQ(ctx context.Context, remoteCQ *kueue.ClusterQueue) {
 	log := ctrl.LoggerFrom(ctx).WithValues("remoteCQ", remoteCQ.Name)
 
+	remoteCl := rc.getClient()
+	if remoteCl == nil {
+		return
```
mimowo, line 451: This feels like an unusual situation (maybe a race, but could be a memory leak), so it feels it deserves logging at V3, maybe even V2. Same for the other uses of getClient which don't return error currently, ie. https://github.com/kubernetes-sigs/kueue/pull/12612/changes#diff-1340a25c39126b0bdf7ee35d63e6a3def36f8282abd49dee94155eee2e804544R549-R552

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
apullo777: Thanks, makes sense. I will wrap this in a small helper and rename setConfig to better reflect what it does.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
apullo777: Agreed. Will add logging for these nil-client early returns.

mimowo:
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for hardening Kueue👍

apullo777:
I will do the manual cherry-pick :)

PR #12615, opened 2026-06-28
address issue 12537 by removing mpijob file

### LINKED ISSUE #12537: Dependency Confusion in sample-mpijob.py via Unregistered mpijob PyPI Package in Victim Environment

### Impact
An attacker who claims the `mpijob` package on PyPI can achieve arbitrary code execution on the machines of users or CI/CD pipelines that attempt to run the provided example and install the missing dependency. This can lead to the compromise of Kubernetes credentials (`kubeconfig`) and other sensitive local data.

### Data Flow
- **site/static/examples/python/sample-mpijob.py:5**: The example script instructs Python to import the `mpijob` package.
- **site/static/examples/python/requirements.txt**: The `mpijob` package is not listed in `requirements.txt` and is not a built-in Python module.
- **PyPI Package Registry**: The `mpijob` package namespace is unregistered on PyPI and can be claimed by an attacker.
- **Victim Environment**: When a user runs the script, encounters a `ModuleNotFoundError`, and runs `pip install mpijob` to resolve it, the attacker's malicious code is executed.

### Reproduction
**Preconditions**: A Python environment with `pip` installed. The `mpijob` package must not be locally defined or installed.
**Steps**:
1. Run `python site/static/examples/python/sample-mpijob.py`.
2. Observe the `ModuleNotFoundError: No module named 'mpijob'` error.
3. Note that `mpijob` is missing from `requirements.txt`.
4. Verify on PyPI that the `mpijob` package namespace is currently unclaimed (e.g., curling `https://pypi.org/pypi/mpijob/json` returns a 404).
5. An attacker can register this package with a malicious `setup.py` that executes upon installation.

### Vulnerable Code
```
import mpijob.models as models
```

### Remediation
Remove the unregistered `mpijob` import and replace it with the officially maintained SDK package, or vendor the necessary CRD models locally.

Suggested Fix:
```
Replace `import mpijob.models as models` with the correct official SDK (e.g., `from kubeflow.training import models` if applicable) and update `requirements.txt` to include the verified dependency.
```

### Environment
* Install tools: N/A
* Others: N/A

kannon92:
#### What type of PR is this?
/kind documentation

Convert files to use uv and add explict installation for flux operator.

Remove MPIJob as there is no pip package for that.

Plus the hope would be to use kubeflow sdk for TrainJob with MPI if we want to add this back.

Fixes #12537

```release-note
NONE
```

## Summary by CodeRabbit

* **Documentation**
  * Updated Python task docs to use `uv run` for example commands instead of manual virtual environment setup.
  * Added brief installation guidance for `uv` and noted that example dependencies are installed automatically.
  * Removed the MPI Operator example section from the Python task guides.

* **Chores**
  * Removed an outdated Python example script from the published examples.


mimowo:
cc @amy ptal

site/static/examples/python/sample-mpijob.py
```diff
@@ -1,143 +0,0 @@
-#!/usr/bin/env python3
```
andreyvelich, line 1: @mimowo @kannon92 Any thoughts on introducing Python example w/ Kueue and TrainJob using Kubeflow SDK?
Or you prefer doing this in the followup PR?
Maybe @kaisoz @aniket2405 @abhijeet-dhumal @robert-bell @reckless-sherixx can help with that ?

kannon92: Sure. Let's add a follow up.

kannon92: https://github.com/kubernetes-sigs/kueue/issues/12634

mimowo:
/hold
I'm confused here, the PR seems to fix the vulnerability by removing parts of documentation, ignores the recommendation, and does some seemingly unrelated changes. In particular:
1. instead of addressing the vulnerability as per suggestion "Remove the unregistered mpijob import and replace it with the officially maintained SDK package, or vendor the necessary CRD models locally." - I would like to understand what is wrong with the suggestion
2. it removes the entire tutorial page for " MPI Operator Job" - I think this is still valuable, so I'm not sure if we need to drop it entirely
3. It replaces "python" invocations with "uv run" - I don't see how this is related, should this be separate PR?

So, could we somehow focus this PR on just fixing the vulnerability without dropping parts of documentation, and leave the "python" -> "uv run" as a potentially separate PR (I'm still not sure about that)

kannon92:
For UV I figured it's easier then having people create venv. Plus uv seems to be gaining traction as the more popular way of installing python dependencies.


There is no way to fix the vulnerability with MPIJob. The issue is that there isn't a pip package for MPIJob so the issue is that someone could add anything for MPIJob and then it could do something bad.

Andrew suggests switching to the Python sdk which has TrainJob integration and that package is on PyPI so we don't have to worry about package squatting.

mimowo:
> For UV I figured it's easier then having people create venv. Plus uv seems to be gaining traction as the more popular way of installing python dependencies.

Maybe, but it seems totally independent, we could take it as a separate PR.

> There is no way to fix the vulnerability with MPIJob. The issue is that there isn't a pip package for MPIJob so the issue is that someone could add anything for MPIJob and then it could do something bad.

Does that mean the example does not even work currently? I'm wondering if must work somehow when the tutorial was written

kannon92:
I can open up UV as separate PR.

For the example, I think it could work if you built the wheel manually and installed it in the environment.

mimowo:
Ah, indeed the manual compilation of the SDK for mpijob is done here: https://github.com/kubernetes-sigs/kueue/pull/12615/changes#diff-51f3a75f861c6802a8db68585c6715bfa7a1debd62fdfc63b20068b094c23796L283-L290

kannon92:
UV PR: https://github.com/kubernetes-sigs/kueue/pull/12723

kannon92:
> Ah, indeed the manual compilation of the SDK for mpijob is done here: https://github.com/kubernetes-sigs/kueue/pull/12615/changes#diff-51f3a75f861c6802a8db68585c6715bfa7a1debd62fdfc63b20068b094c23796L283-L290

Yea so I don't see a way to resolve this vulnerability since we have to import but we can't define this requirement.

So I think replacing with #12634 seems to be the better choice.

mimowo:
@kannon92 please rebase

kannon92:
/hold cancel

I think removing this in place of kubeflow sdk for TrainJob is the best way to "fix" this.

tenzen-y (commented):
I don't have any objections for this removal.
/approve

Leaving LGTM on @mimowo

mimowo:
Ok,let's do that, another fading flower 🥀  Thank you for taking care of the legacy stuff we have no capacity to maintain 👍 
/lgtm
/cherrypick website

PR #12643, opened 2026-06-30
Sanitize remote metadata for MultiKueue external-framework jobs

### LINKED ISSUE #12540: Infinite Create-Delete Loop (DoS) in MultiKueue External Frameworks via Malicious OwnerReferences in pkg/controller/admissionchecks/multikueue/externalframeworks/adapter.go

### Impact
A malicious or misconfigured tenant can submit an external framework custom job with an invalid ownerReference. This causes the multikueue worker adapter to enter a rapid, infinite Create-Delete-Reconcile loop. This severely degrades the performance of the local Kueue controller and the remote API server, leading to a Denial of Service (DoS) for all multikueue operations.

### Data Flow
- **pkg/controller/admissionchecks/multikueue/externalframeworks/adapter.go:63**: Local custom job containing an untrusted ownerReference is fetched into localObj
- **pkg/controller/admissionchecks/multikueue/externalframeworks/adapter.go:77**: localObj is deep copied into remoteObj, retaining the malicious ownerReference and local UID
- **pkg/controller/admissionchecks/multikueue/externalframeworks/adapter.go:78**: Only ResourceVersion is cleared. The ownerReferences and other sensitive metadata fields remain intact
- **pkg/controller/admissionchecks/multikueue/externalframeworks/adapter.go:90**: remoteClient.Create is called, pushing the object to the remote cluster where its garbage collector will immediately delete it due to the invalid ownerReference

### Reproduction
**Preconditions**:
1. MultiKueue is configured with `MultiKueueAdaptersForCustomJobs` enabled.
2. At least one external framework custom job CRD is configured.
3. A remote worker cluster is linked.

**Steps**:
1. Create an external framework Custom Job with an `ownerReferences` block pointing to a random, non-existent UID.
2. Submit the custom job to a LocalQueue managed by Kueue.
3. Observe the logs of the Kueue controller manager. It will be stuck in a rapid, infinite loop constantly logging 'Creating remote object'.

### Vulnerable Code
```
	// Create a copy of the local object for the remote cluster
	remoteObj := localObj.DeepCopy()
	remoteObj.SetResourceVersion("")
```

### Remediation
When duplicating the local unstructured object for the remote cluster, explicitly strip `metadata.ownerReferences`, `metadata.uid`, `metadata.finalizers`, `metadata.managedFields`, and the `status` block. This mirrors the safe behavior of `api.CloneObjectMetaForCreation` used by built-in job adapters.

Suggested Fix:
```
	// Create a copy of the local object for the remote cluster
	remoteObj := localObj.DeepCopy()
	
	// Strip fields that shouldn't be copied to the remote cluster
	unstructured.RemoveNestedField(remoteObj.Object, "metadata", "resourceVersion")
	unstructured.RemoveNestedField(remoteObj.Object, "metadata", "uid")
	unstructured.RemoveNestedField(remoteObj.Object, "metadata", "ownerReferences")
	unstructured.RemoveNestedField(remoteObj.Object, "metadata", "finalizers")
	unstructured.RemoveNestedField(remoteObj.Object, "metadata", "managedFields")
	unstructured.RemoveNestedField(remoteObj.Object, "status")
```

This logical root cause was found at **3** additional call site(s):

- `pkg/controller/jobframework/indexer.go` (Finding 13596: Nil pointer dereference panic in EquivalentToWorkload via malicious Workload OwnerReference)
- `pkg/controller/jobs/pod/constants/constants.go` (Finding 13651: Cluster Quota and Admission Bypass via Malicious Pod OwnerReferences)
- `pkg/util/api/api.go` (Finding 13722: Infinite Create-Delete Loop (DoS) in MultiKueue External Frameworks via Malicious OwnerReferences)

### Environment
* Install tools: N/A
* Others: N/A

apullo777:
#### What type of PR is this?

/kind bug
/area multikueue

The external-framework MultiKueue adapter created remote objects from `localObj.DeepCopy()` and only cleared `resourceVersion`. This could copy source-cluster metadata such as ownerReferences to the worker cluster.

Since those owner UIDs do not exist on the worker, the worker garbage collector can delete the remote object, causing the adapter to recreate it repeatedly.

This PR builds the remote unstructured object from scratch instead, keeping only GVK, name, namespace, labels, annotations, and spec. This mirrors the `api.CloneObjectMetaForCreation` whitelist used by the typed adapters.

The integration test asserts that the remote object is created without `ownerReferences`, `finalizers`, or `status`. It does not assert the full create-delete loop because envtest does not run a real garbage collector. AI assistance was used for parts of this change.

Fixes #12540

```release-note
MultiKueue: Fixed custom jobs using external-framework adapters being repeatedly created and deleted on worker clusters when source-cluster metadata was copied to the remote object.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Remote objects created for admitted workloads are now sanitized to include only the expected creation-safe metadata (and spec only when present), avoiding unintended carryover such as local ownership details and status.
  * Worker-created resources no longer inherit source-cluster ownership, finalizers, or status metadata.

* **Tests**
  * Added unit tests to verify creation-safe cloning and remote object metadata/spec handling (including replicas/spec behavior).
  * Added an integration scenario ensuring worker-created RayCluster resources start without inherited source metadata.

### DIFF
--- a/pkg/controller/admissionchecks/multikueue/externalframeworks/adapter.go
+++ b/pkg/controller/admissionchecks/multikueue/externalframeworks/adapter.go
@@ -86,9 +86,7 @@ func (a *Adapter) SyncJob(ctx context.Context, localClient client.Client, remote
 func (a *Adapter) createRemoteObject(ctx context.Context, remoteClient client.Client, localObj *unstructured.Unstructured, workloadName, origin string) error {
 	log := ctrl.LoggerFrom(ctx)
 
-	// Create a copy of the local object for the remote cluster
-	remoteObj := localObj.DeepCopy()
-	remoteObj.SetResourceVersion("")
+	remoteObj := cloneUnstructuredForCreation(localObj)
 
 	// Apply default transformation: remove the managedBy field
 	a.removeManagedByField(remoteObj)
@@ -99,6 +97,24 @@ func (a *Adapter) createRemoteObject(ctx context.Context, remoteClient client.Cl
 	// Create the object in the remote cluster
 	log.V(2).Info("Creating remote object", "gvk", a.gvk, "name", remoteObj.GetName(), "namespace", remoteObj.GetNamespace())
 	return remoteClient.Create(ctx, remoteObj)
+}
+
+// cloneUnstructuredForCreation creates a copy of obj containing only the GVK, name,
+// namespace, labels, annotations and spec, the unstructured analogue of
+// api.CloneObjectMetaForCreation; keep the two in sync. Dropping the rest (notably
+// ownerReferences, whose UIDs don't exist remotely) avoids the remote GC deleting the
+// copy in a create-delete loop.
+func cloneUnstructuredForCreation(obj *unstructured.Unstructured) *unstructured.Unstructured {
+	remote := &unstructured.Unstructured{Object: map[string]any{}}
+	remote.SetGroupVersionKind(obj.GroupVersionKind())
+	remote.SetName(obj.GetName())
+	remote.SetNamespace(obj.GetNamespace())
+	remote.SetLabels(obj.GetLabels())
+	remote.SetAnnotations(obj.GetAnnotations())
+	if spec, found, _ := unstructured.NestedFieldCopy(obj.Object, "spec"); found {
+		remote.Object["spec"] = spec
+	}
+	return remote
 }
 
 func (a *Adapter) syncStatus(ctx context.Context, localClient client.Client, localObj, remoteObj *unstructured.Unstructured) error {
--- a/pkg/controller/admissionchecks/multikueue/externalframeworks/adapter.go
+++ b/pkg/controller/admissionchecks/multikueue/externalframeworks/adapter.go
@@ -34,6 +34,7 @@ import (
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
 	"sigs.k8s.io/kueue/pkg/controller/jobframework"
 	"sigs.k8s.io/kueue/pkg/features"
+	"sigs.k8s.io/kueue/pkg/util/api"
 )
 
 // Adapter implements the MultiKueueAdapter interface for external frameworks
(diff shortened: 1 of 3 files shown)


pkg/controller/admissionchecks/multikueue/externalframeworks/adapter.go
```diff
@@ -101,6 +99,24 @@ func (a *Adapter) createRemoteObject(ctx context.Context, remoteClient client.Cl
 	return remoteClient.Create(ctx, remoteObj)
 }
 
+// cloneUnstructuredForCreation creates a copy of obj containing only the GVK, name,
+// namespace, labels, annotations and spec, the unstructured analogue of
+// api.CloneObjectMetaForCreation; keep the two in sync. Dropping the rest (notably
```
Singularity23x0: To be more precise, this replicates the logic of `copySpec func(dst, src PtrT)`, which just so happens to, for every implementation, copy the `spec` and use the `api.CloneObjectMetaForCreation` to copy the ObjectMeta.

To avoid having to sync methods we can extract the part that does indeed focus on mirroring `api.CloneObjectMetaForCreation`  and make one of the methods call the other. Since both cases would require converting Unstructured and ObjectMeta in both directions, choosing where the logic actually lies should mainly be a cosmetic choice. It might be easier to use the `api.CloneObjectMetaForCreation` here instead of the other way around as it is placed in the utils/api package.

apullo777: Thanks, that makes sense. I will update `cloneUnstructuredForCreation` to reuse `api.CloneObjectMetaForCreation` for the metadata part, and keep only the unstructured-specific `spec` copy locally.

That should remove the parallel whitelist / “keep in sync” concern while keeping the behavior covered by the existing `TestCloneUnstructuredForCreation` cases. I will also thread an error return through, since the unstructured to ObjectMeta conversion can fail.

mimowo:
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for hardening of Kueue 👍

apullo777:
Let me do the cherry-pick for 0.17 :)

PR #12644, opened 2026-06-30
Fix RayJob dropping admission constraints for default submitter pod template

### LINKED ISSUE #12567: RayJob Controller drops placement constraints for default Submitter Pods allowing quota/node restriction bypass in pkg/controller/jobs/rayjob/rayjob_controller.go

### Impact
Users can execute arbitrary code in the cluster via the submitter pod on nodes that should be restricted to them (e.g., bypassing node-based tenant isolation or GPU quota restrictions) because Kueue fails to persist the admission constraints.

### Data Flow
- **pkg/controller/jobs/rayjob/rayjob_controller.go:275**: When SubmitterPodTemplate is nil, getSubmitterTemplate returns a new, unattached PodTemplateSpec object.
- **pkg/controller/jobs/rayjob/rayjob_controller.go:219**: RunWithPodSetsInfo calls getSubmitterTemplate and receives this local, unattached struct.
- **pkg/controller/jobs/rayjob/rayjob_controller.go:221**: podset.Merge applies Kueue's admission constraints (nodeAffinity, tolerations, etc.) to the local struct.
- **pkg/controller/jobs/rayjob/rayjob_controller.go:224**: The function returns without assigning the mutated struct back to j.Spec.SubmitterPodTemplate, causing the constraints to be lost.

### Reproduction
**Preconditions**:
1. Kueue and Kuberay are installed.
2. A LocalQueue and ClusterQueue are configured with strict NodeAffinity or Tolerations that restrict where workloads can run.

**Steps**:
1. Submit a RayJob with `submissionMode: K8sJobMode` and leave `submitterPodTemplate` undefined.
2. Wait for Kueue to admit the Workload.
3. Inspect the created RayJob resource and the resulting submitter Job.

### Vulnerable Code
```go
submitterPod := getSubmitterTemplate(j)
info := podSetsInfo[expectedLen-1]
if err := podset.Merge(&submitterPod.ObjectMeta, &submitterPod.Spec, info); err != nil {
	return err
}
```

### Remediation
Persist the generated submitter template back to the RayJob spec.

Suggested Fix:
```go
if j.Spec.SubmissionMode == rayv1.K8sJobMode {
	submitterPod := getSubmitterTemplate(j)
	info := podSetsInfo[expectedLen-1]
	if err := podset.Merge(&submitterPod.ObjectMeta, &submitterPod.Spec, info); err != nil {
		return err
	}
	if j.Spec.SubmitterPodTemplate == nil {
		j.Spec.SubmitterPodTemplate = submitterPod
	}
}
```

### Environment
* Install tools: N/A
* Others: N/A

carterpewpew:
#### What type of PR is this?
/kind bug
/area integrations

When a RayJob uses `submissionMode: K8sJobMode` with no explicit `submitterPodTemplate`, `getSubmitterTemplate()` returns a new, disconnected `PodTemplateSpec`. `podset.Merge` applies Kueue's admission constraints (nodeSelector, tolerations, nodeAffinity) to this local struct, but it is never assigned back to `j.Spec.SubmitterPodTemplate`. The patch callback then serializes a nil `SubmitterPodTemplate`, and the constraints are silently dropped. KubeRay creates the submitter Job with its own unconstrained default template. This PR assigns the merged template back to `j.Spec.SubmitterPodTemplate` when it was originally nil, in both `RunWithPodSetsInfo` and `RestorePodSetsInfo`. This follows the same pattern used by the SparkApplication controller for nil `Driver.Template` / `Executor.Template`.

Fixes #12567

```release-note
RayJob: Fix the integration controller dropping Kueue admission placement constraints (nodeSelector, tolerations, nodeAffinity) for the submitter pod when submitterPodTemplate is not explicitly set and submissionMode is K8sJobMode.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved job handling so submitter pod settings are preserved when running in Kubernetes Job mode.
  * Fixed an issue where admission-related placement constraints could be lost for jobs that start without a submitter pod template.
  * Ensured node selector settings are restored correctly after job suspension and reconciliation.

### DIFF
--- a/pkg/controller/jobs/rayjob/rayjob_controller.go
+++ b/pkg/controller/jobs/rayjob/rayjob_controller.go
@@ -174,6 +174,9 @@ func (j *RayJob) RunWithPodSetsInfo(ctx context.Context, _ client.Client, podSet
 		if err := podset.Merge(&submitterPod.ObjectMeta, &submitterPod.Spec, info); err != nil {
 			return err
 		}
+		if j.Spec.SubmitterPodTemplate == nil {
+			j.Spec.SubmitterPodTemplate = submitterPod
+		}
 	}
 
 	return nil
--- a/pkg/controller/jobs/rayjob/rayjob_controller_test.go
+++ b/pkg/controller/jobs/rayjob/rayjob_controller_test.go
@@ -685,6 +685,140 @@ func TestNodeSelectors(t *testing.T) {
 
 			wantFinal: baseJob.DeepCopy(),
 		},
+		"K8sJobMode with nil SubmitterPodTemplate persists admission constraints": {
+			job: func() *rayv1.RayJob {
+				j := testingrayutil.MakeJob("job", "ns").
+					WithSubmissionMode(rayv1.K8sJobMode).
+					Obj()
+				j.Spec.RayClusterSpec.HeadGroupSpec.Template.Spec.NodeSelector = map[string]string{}
+				j.Spec.RayClusterSpec.WorkerGroupSpecs = baseJob.Spec.RayClusterSpec.WorkerGroupSpecs
+				return j
+			}(),
+			runInfo: []podset.PodSetInfo{
+				{
+					NodeSelector: map[string]string{
+						"newKey": "newValue",
+					},
+				},
+				{
+					NodeSelector: map[string]string{
+						"key-wg1": "value-wg1",
+					},
+				},
+				{
+					NodeSelector: map[string]string{},
+				},
+				{
+					NodeSelector: map[string]string{
+						"submitter-key": "submitter-value",
+					},
+				},
+			},
+			restoreInfo: []podset.PodSetInfo{
+				{
+					NodeSelector: map[string]string{},
+				},
+				{
+					NodeSelector: map[string]string{
+						"key-wg1": "value-wg1",
+					},
+				},
+				{
+					NodeSelector: map[string]string{
+						"key-wg2": "value-wg2",
+					},
+				},
+				{
+					NodeSelector: map[string]string{},
+				},
+			},
+			wantAfterRun: func() *rayv1.RayJob {
+				j := testingrayutil.MakeJob("job", "ns").
+					Suspend(false).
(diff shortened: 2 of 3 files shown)


pkg/controller/jobs/rayjob/rayjob_controller.go
```diff
@@ -196,6 +199,9 @@ func (j *RayJob) RestorePodSetsInfo(podSetsInfo []podset.PodSetInfo) bool {
 		submitterPod := getSubmitterTemplate(j)
 		info := podSetsInfo[expectedLen-1]
 		changed = podset.RestorePodSpec(&submitterPod.ObjectMeta, &submitterPod.Spec, info) || changed
+		if j.Spec.SubmitterPodTemplate == nil {
+			j.Spec.SubmitterPodTemplate = submitterPod
+		}
```
mimowo: Is that needed? I understand why we inject the snippet into `RunWithPodSetsInfo` - in order to inject the nodeSelectors & tolerations into the Job PodTamplete on start, but I'm not clear why we would need to do that on Restore (stop), wdyt?

mimowo (commented):
cc @yaroslava-serdiuk ptal

pkg/controller/jobs/rayjob/rayjob_controller.go
carterpewpew: Good point and thanks for the catch. The guard here is purely defensive in normal operation RunWithPodSetsInfo always sets SubmitterPodTemplate before RestorePodSetsInfo is called, so the nil check is always false. Happy to drop it.

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for hardening of the KubeRay integration 👍

mimowo:
/release-note-edit
```release-note
RayJob: Fix the integration controller dropping Kueue admission placement constraints (nodeSelector, tolerations, nodeAffinity) for the submitter pod when submitterPodTemplate is not explicitly set and submissionMode is K8sJobMode.
```

PR #12647, opened 2026-06-30
Fix SparkApplication node selector clobbering across driver and executor

### LINKED ISSUE #12547: ResourceFlavor Node Affinity Clobbering in SparkApplication Allows Quota Bypass in pkg/controller/jobs/sparkapplication/sparkapplication_controller.go
### Impact
A tenant can exploit this bug to schedule large driver pods onto restricted or expensive node pools (e.g. GPUs) without Kueue accounting for those resources against the expensive node pool's quota. This leads to quota bypass, resource theft, and potential denial of service (OOM/CPU starvation) on the restricted nodes.

### Data Flow
- **pkg/controller/jobs/sparkapplication/sparkapplication_controller.go:167**: User sets `spec.nodeSelector: {}` (or any valid map) on the SparkApplication. Kueue reads this into `nodeSelector = j.Spec.NodeSelector`.
- **pkg/controller/jobs/sparkapplication/sparkapplication_controller.go:216**: mutatePodSetInfoFor(driver) executes first. The driver's ResourceFlavor node selector is computed and injected into the global `j.Spec.NodeSelector`.
- **pkg/controller/jobs/sparkapplication/sparkapplication_controller.go:216**: mutatePodSetInfoFor(executor) executes next. `j.Spec.NodeSelector` is OVERWRITTEN by the executor's ResourceFlavor node selector.
- **pkg/controller/jobs/sparkapplication/sparkapplication_controller.go:225**: The function returns. The Spark Operator consumes the global `j.Spec.NodeSelector` and schedules BOTH the driver and executor pods onto the executor's flavor node pool.

### Reproduction
**Preconditions**:
1. Kueue must be configured with a `ClusterQueue` containing two `ResourceFlavors`: `cpu-flavor` (e.g. nodeSelector: `node-type: cpu`) and `gpu-flavor` (e.g. nodeSelector: `node-type: gpu`).
2. The Spark operator must be installed and the `sparkoperator.k8s.io/sparkapplication` integration enabled in Kueue.

**Steps**:
1. Submit a `SparkApplication` with `spec.nodeSelector: {}`.
2. Define the `driver` pod spec to request CPU resources (which Kueue will match to `cpu-flavor`).
3. Define the `executor` pod spec to request GPU resources (which Kueue will match to `gpu-flavor`).
4. Wait for Kueue to admit the workload.
5. Observe the created `SparkApplication` object's `spec.nodeSelector` field. It will match the GPU flavor.

### Vulnerable Code
```
if j.Spec.NodeSelector != nil {
	j.Spec.NodeSelector = sparkPodSetInfo.NodeSelector
} else {
```

### Remediation
Do not overwrite the global `j.Spec.NodeSelector` during `RunWithPodSetsInfo`. Instead, migrate the injected node selectors into the role-specific `sparkPodSpec.NodeSelector` fields and clear the global `j.Spec.NodeSelector`.

Suggested Fix:
```
Change the assignment in `RunWithPodSetsInfo` to always apply to `sparkPodSpec.NodeSelector` and explicitly clear `j.Spec.NodeSelector`:

go
		sparkPodSpec.NodeSelector = sparkPodSetInfo.NodeSelector
		j.Spec.NodeSelector = nil
```
Apply a similar fix to `RestorePodSetsInfo` to ensure correctness during suspension.

### Environment
* Install tools: N/A
* Others: N/A

carterpewpew:
#### What type of PR is this?
/kind bug
/area integrations

When a SparkApplication uses `spec.nodeSelector` (global), `RunWithPodSetsInfo` overwrites `j.Spec.NodeSelector` on each role iteration inside the `mutatePodSetInfoFor` closure. The driver's merged NodeSelector is written to the global field, then the executor reads it as its merge base. This causes a merge conflict error (`conflict for key=..., value1=..., value2=...`) when flavors inject the same key with different values Silent clobbering when flavors inject non-overlapping keys, causing the Spark Operator to schedule both pods to the wrong node pool. The fix snapshots the original global nodeSelector before the loop, gives each role its own clone as the merge base, always writes the result to per-role `sparkPodSpec.NodeSelector`, and clears `spec.nodeSelector` afterward. Symmetric changes are applied to `RestorePodSetsInfo`.

AI-assisted: This change was developed with AI assistance, but was manually verified and tested.

Fixes: #12547

```release-note
SparkApplication: Fixed a bug where the global spec.nodeSelector could overwrite driver or executor node selectors when they were admitted to different ResourceFlavors.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of Spark application `nodeSelector` settings so global and per-role (`driver`/`executor`) values are merged and applied consistently during updates and restores.
  * Fixed cases where a leftover global `nodeSelector` could persist after processing, potentially affecting scheduling.

* **New Features**
  * Enhanced Spark application test builder utilities with a `NodeSelector(...)` helper for easier setup of global `nodeSelector` values.

### DIFF
--- a/pkg/controller/jobs/sparkapplication/sparkapplication_controller.go
+++ b/pkg/controller/jobs/sparkapplication/sparkapplication_controller.go
@@ -157,6 +157,8 @@ func (j *SparkApplication) RunWithPodSetsInfo(ctx context.Context, _ client.Clie
 
 	j.Spec.Suspend = new(false)
 
+	origGlobalNodeSelector := maps.Clone(j.Spec.NodeSelector)
+
 	mutatePodSetInfoFor := func(role string) error {
 		var podSetInfo podset.PodSetInfo
 		var nodeSelector map[string]string
@@ -170,8 +172,8 @@ func (j *SparkApplication) RunWithPodSetsInfo(ctx context.Context, _ client.Clie
 			}
 			// spec.NodeSelector and spec.driver.nodeSelector is mutually exclusive
 			nodeSelector = j.Spec.Driver.NodeSelector
-			if j.Spec.NodeSelector != nil {
-				nodeSelector = j.Spec.NodeSelector
+			if origGlobalNodeSelector != nil {
+				nodeSelector = maps.Clone(origGlobalNodeSelector)
 			}
 			sparkPodSpec = &j.Spec.Driver.SparkPodSpec
 		case sparkcommon.SparkRoleExecutor:
@@ -182,8 +184,8 @@ func (j *SparkApplication) RunWithPodSetsInfo(ctx context.Context, _ client.Clie
 			}
 			// spec.NodeSelector and spec.executor.nodeSelector is mutually exclusive
 			nodeSelector = j.Spec.Executor.NodeSelector
-			if j.Spec.NodeSelector != nil {
-				nodeSelector = j.Spec.NodeSelector
+			if origGlobalNodeSelector != nil {
+				nodeSelector = maps.Clone(origGlobalNodeSelector)
 			}
 		default:
 			return fmt.Errorf("unknown Spark role: %s", role)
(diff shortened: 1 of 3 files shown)


mimowo:
cc @everpeace ptal

everpeace (approved):
/lgtm

Thanks for the good catch! The change is LGTM! 

I think clearing `spec.nodeSelector` after Kueue admitted it would not be a severe problem because `spec.nodeSelector` and `spec.{driver, executor}.spec.nodeSelector` is mutually exclusive in Spark Operator.

mimowo:
/release-note-edit
```release-note
SparkApplication: Fixed a bug where the global spec.nodeSelector could overwrite driver or executor node selectors when they were admitted to different ResourceFlavors.
```

PR #12648, opened 2026-06-30
fix: apply IMAGE_BUILD_CMD in makefile

weizhoublue:
#### What type of PR is this?

/kind bug

IMAGE_BUILD_CMD was overridden with the retry wrapper in the parent push target, but not forwarded into the recursive make call, so the child image-build target fell back
  to its default docker buildx command

```
  entry: image-local-push     <--------- defines IMAGE_BUILD_CMD = retry.sh + docker buildx build

      -> image-local-build
          $(MAKE) image-build   <------- missing IMAGE_BUILD_CMD forwarding

      -> image-build
          defines IMAGE_BUILD_CMD ?= $(DOCKER_BUILDX_CMD) build
          $(IMAGE_BUILD_CMD) .....
```

Fix:
  Forward IMAGE_BUILD_CMD explicitly, using:

```
  IMAGE_BUILD_CMD="$(subst ",\",$(IMAGE_BUILD_CMD))"
```

  This preserves embedded double quotes while passing the value to the recursive make invocation.

  Example:

  IMAGE_BUILD_CMD = docker build -t "my-app:latest" .
  -->
  IMAGE_BUILD_CMD = docker build -t \"my-app:latest\" .
  
  
  

Fixes #

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image build behavior so custom build arguments/overrides are preserved when pushing images.
  * Made local and specialized image push targets more reliable by ensuring the intended build command is consistently propagated to the underlying build steps.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -228,7 +228,7 @@ image-local-build:
 
 # Build the multiplatform container image locally and push to repo.
 .PHONY: image-local-push
-image-local-push: IMAGE_BUILD_CMD := $(IMAGE_PUSH_RETRY) $(IMAGE_BUILD_CMD)
+image-local-push: export IMAGE_BUILD_CMD := $(IMAGE_PUSH_RETRY) $(IMAGE_BUILD_CMD)
 image-local-push: PUSH=--push
 image-local-push: image-local-build
 
@@ -487,7 +487,7 @@ kueue-populator-image-build:
 		IMAGE_BUILD_EXTRA_OPTS="$(IMAGE_BUILD_EXTRA_OPTS) -t $(IMAGE_REPO_KUEUE_POPULATOR):$(RELEASE_BRANCH)"
 
 .PHONY: kueue-populator-image-push
-kueue-populator-image-push: IMAGE_BUILD_CMD := $(IMAGE_PUSH_RETRY) $(IMAGE_BUILD_CMD)
+kueue-populator-image-push: export IMAGE_BUILD_CMD := $(IMAGE_PUSH_RETRY) $(IMAGE_BUILD_CMD)
 kueue-populator-image-push: PUSH=--push
 kueue-populator-image-push: kueue-populator-image-build
 
@@ -511,7 +511,7 @@ kueue-priority-booster-image-build:
 		IMAGE_BUILD_EXTRA_OPTS="$(IMAGE_BUILD_EXTRA_OPTS) -t $(IMAGE_REPO_KUEUE_PRIORITY_BOOSTER):$(RELEASE_BRANCH)"
 
 .PHONY: kueue-priority-booster-image-push
-kueue-priority-booster-image-push: IMAGE_BUILD_CMD := $(IMAGE_PUSH_RETRY) $(IMAGE_BUILD_CMD)
+kueue-priority-booster-image-push: export IMAGE_BUILD_CMD := $(IMAGE_PUSH_RETRY) $(IMAGE_BUILD_CMD)
 kueue-priority-booster-image-push: PUSH=--push
 kueue-priority-booster-image-push: kueue-priority-booster-image-build
 


mimowo:
Cc @mbobrovskyi ptal

Makefile
mbobrovskyi:
```suggestion
image-local-push: export IMAGE_BUILD_CMD := $(IMAGE_PUSH_RETRY) $(IMAGE_BUILD_CMD)
```
Can we use export instead?

weizhoublue: I think the change is not needed for  image-local-push .

This fix follows the Makefile's established style of passing variables explicitly to sub-makes, rather than relying on export. The execution chain is already complete:

```
    image-local-push          
    # L231: sets IMAGE_BUILD_CMD := $(IMAGE_PUSH_RETRY) $(IMAGE_BUILD_CMD)
      └─→ image-local-build   
         # L226: passes  $(MAKE) image-build ... IMAGE_BUILD_CMD="$(subst ...$(IMAGE_BUILD_CMD)...)" to sub-make <---- added by this commit
            └─→ image-build   # uses the received IMAGE_BUILD_CMD directly
```

The fix in this commit makes  image-local-build  forward that already-mutated value to the sub-make via explicit  IMAGE_BUILD_CMD="..."  argument.   So by the time  image-build  runs, it receives the retry-prefixed command — no extra line needed.

mbobrovskyi: Yes, my idea is not to change `image-local-build`. Instead, just add `export` before `IMAGE_BUILD_CMD`. Will that work?

weizhoublue: Yes, using export to set environment variables is another viable fix. 
I chose a solution that aligns more closely with the coding style of the Makefile. 
If you think using export would be more concise, I can switch to that approach.

mbobrovskyi: I think using `export` is more obvious than replacing it in place.

@mimowo WDYT?

mimowo: i think the current proposal is not changing the value in-place, but indeed tbe weak point is the need for the extra quotation escaping. so iiuc this is the alternative which workarounds the quotation problem:

image-local-push: export IMAGE_BUILD_CMD := $(IMAGE_PUSH_RETRY) $(IMAGE_BUILD_CMD)
image-local-push: PUSH=--push
image-local-push: image-local-build

weizhoublue: updated

mbobrovskyi:
Thank you!

/lgtm
/approve

mimowo:
@weizhoublue please prep the CP for 0.17

mimowo:
@weizhoublue friendly ping

weizhoublue:
> @weizhoublue friendly ping

thanks for reminding.

I feel this fix is not applicable to the branch release-0.17 . 
This variable  IMAGE_PUSH_RETRY set in IMAGE_BUILD_CMD was introduced in the  main  branch after  release-0.17  was cut.
The  release-0.17  Makefile never sets  IMAGE_BUILD_CMD  in any push target, so the issue does not exist there.

mbobrovskyi:
> This variable IMAGE_PUSH_RETRY set in IMAGE_BUILD_CMD was introduced in the main branch after release-0.17 was cut.

Actually, we have this https://github.com/kubernetes-sigs/kueue/blob/release-0.17/Makefile#L222-L223.

mimowo:
> This variable IMAGE_PUSH_RETRY set in IMAGE_BUILD_CMD was introduced in the main branch after release-0.17 was cut.

Yes, but that PR was cherrypicked to ensure build system conistency across all release branches. So, we also need to CP the bugfix.

weizhoublue:
> > This variable IMAGE_PUSH_RETRY set in IMAGE_BUILD_CMD was introduced in the main branch after release-0.17 was cut.
> 
> Actually, we have this https://github.com/kubernetes-sigs/kueue/blob/release-0.17/Makefile#L222-L223.

Thanks for your double-check.  I missed this because my fork was out of sync with upstream.

I submitted  the PR #12753 to apply the fix to 0.17

PR #12512, opened 2026-06-25
Add kueue_pod_scheduling_gate_removal_seconds metric for ElasticJobUngater.

mbobrovskyi:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
Add kueue_pod_scheduling_gate_removal_seconds metric for ElasticJobUngater.

#### Which issue(s) this PR fixes:

Fixes #11170

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scheduling-gate removal timing so it’s recorded more reliably when elastic workloads are ungated.
  * Reconciliation now uses a consistent time source, improving the accuracy of scheduling-gate removal metrics.

* **Tests**
  * Added coverage to validate scheduling-gate removal timing and histogram metric reporting, including use of a deterministic clock to prevent time-related flakiness.

PR #12652, opened 2026-06-30
Fix: Clear Ungate Expectations on No-Op Elastic Pod Patch

weizhoublue:
#### What type of PR is this?
 
/kind bug
 

The scenario is a race around ElasticJob pod ungating. 

The controller records expected pod update events before patching the gated pods. If a pod has already had the ElasticJob scheduling gate removed by another reconcile or external/concurrent update before the patch runs, `utilpod.Ungate` returns `ungated=false`. 
In that path no API update is sent, so no pod update event will be observed later. Without explicitly marking the UID as observed, the expectation stays pending and the ElasticJob workload can remain blocked by pending ungate operations.

```
func (r *elasticJobUngater) Reconcile(..）{
    ...
    r.expectationsStore.ExpectUIDs(log, req.NamespacedName, uids)      <--------- register
    ...
    err = parallelize.Until(ctx, len(pods), func(i int) error {
       ...
        ungated = Patch(pod)
        if error  → ObservedUID(pod.UID)  // fail to patch, manually clear
        if ungated → OnPodUpdate          // patch succeeded, event handler clears
        if !ungated → ???                 // no-op, ObservedUID never called  ← BUG
    }
}
```

The consequence is that the workload may keep requeueing/backing off instead of completing the ungating flow, even though the pod no longer has the scheduling gate.

Expected behavior is to clear the expectation immediately when no patch/update event is expected. This matches the existing logic in  [pkg/controller/tas/topology_ungater.go](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/tas/topology_ungater.go#L303) .

 
Fixes #

```release-note
ElasticJob: Added a defensive safeguard to avoid stale pending expectations in no-op pod ungate paths, following 
the pattern from the TAS topology ungater.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pod scheduling gate handling so gates are removed and tracked more consistently during reconciliation.
  * Fixed the order of internal status updates to better reflect when pods have actually been ungated.

### DIFF
--- a/pkg/controller/elasticjobs/elastic_job_ungater.go
+++ b/pkg/controller/elasticjobs/elastic_job_ungater.go
@@ -148,7 +148,9 @@ func (r *elasticJobUngater) Reconcile(ctx context.Context, req reconcile.Request
 			log.Error(e, "failed ungating elastic pod", "pod", klog.KObj(pod))
 			return e
 		}
-		if ungated {
+		if !ungated {
+			r.expectationsStore.ObservedUID(log, req.NamespacedName, pod.UID)
+		} else {
 			utilpod.RecordPodSchedulingGateRemovalSeconds(r.clock, kueue.ElasticJobSchedulingGate, wl, false)
 		}
 		return nil


weizhoublue:
@mbobrovskyi  hi , I submitted the follow-up optimization for PR #12512
could you help check wether this makes sense, thanks

mimowo:
@weizhoublue could you please provide a bit more context - explain what is the scenario and what are the consequences of the bug. Setting the release note is think is needed too. 

Also unit tests will be great.

Finally, do you think this is a recent regression related to the introduction oRecordPodSchedulingGateRemovalSeconds, or the bug existed before?

cc @mbobrovskyi

weizhoublue:
> @weizhoublue could you please provide a bit more context - explain what is the scenario and what are the consequences of the bug. Setting the release note is think is needed too.
> 
> Also unit tests will be great.
> 
> Finally, do you think this is a recent regression related to the introduction oRecordPodSchedulingGateRemovalSeconds, or the bug existed before?
> 
> cc @mbobrovskyi


OK, I updated the PR description according to my understanding.

On unit tests: the  ungated=false  path cannot be accurately reproduced in the fake-client framework. listGatedPods  and  Ungate  operate on the same in-memory pod pointer, so there is no injection point to simulate the race.

But the fix is verified by symmetry with  [pkg/controller/tas/topology_ungater.go](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/tas/topology_ungater.go#L303)  , which has the identical  !ungated → ObservedUID  pattern

I thinks this may be an original design oversight. Looking at the first commit of this controller, the !ungated  branch was already unhandled from day one

mimowo:
> On unit tests: the ungated=false path cannot be accurately reproduced in the fake-client framework. listGatedPods and Ungate operate on the same in-memory pod pointer, so there is no injection point to simulate the race.
> But the fix is verified by symmetry with [pkg/controller/tas/topology_ungater.go](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/tas/topology_ungater.go#L303) , which has the identical !ungated → ObservedUID pattern

Hmmm, could you check if there are any unit tests failing for the topology ungater if this line is commented out? Maybe that could inspire some unit testing here too.

weizhoublue:
> Hmmm, could you check if there are any unit tests failing for the topology ungater if this line is commented out? Maybe that could inspire some unit testing here too.

Yes, good suggestion. 
The existing TAS unit tests don't cover this path — I verified by commenting out the same `ObservedUID` line in `topology_ungater.go`, all 32 `TestReconcile` cases still pass. 

I couldn't find a practical unit test that reproduces the issue without a contrived client setup, so I'm keeping this as a defensive parity fix with TAS.

If this fix can’t be reproduced by unit tests and is considered risky, we could pend merging this PR, since the likelihood of encountering it in practice appears low.

mimowo:
Actually, have you observed this issue on prod? It looks like a defensive safeguard in both cases (TAS and elastic-jobs), because whenever the Pod has the gate removed or is deleted we already have the code to remove the observation.

If you haven't seen this in prod, then I'm ok with this fix as is, but we may adjust the release note clarifying this is a safeguard. wdyt?

weizhoublue:
> Actually, have you observed this issue on prod? It looks like a defensive safeguard in both cases (TAS and elastic-jobs), because whenever the Pod has the gate removed or is deleted we already have the code to remove the observation.
> 
> If you haven't seen this in prod, then I'm ok with this fix as is, but we may adjust the release note clarifying this is a safeguard. wdyt?

I have not observed this in production. This came from static analysis while comparing ElasticJobUngater with the existing TAS ungater logic.
I agree this is best described as a defensive safeguard.
I updated the release note to make that clear and avoid implying confirmed production impact.

mimowo:
/remove-kind bug
/kind cleanup
/release-note-edit
```release-note
ElasticJob: Added a defensive safeguard to avoid stale pending expectations in no-op pod ungate paths, following 
the pattern from the TAS topology ungater.
```

mimowo:
Thank you for explaining, please prepare the CPs in case the robot fails on conflicts.
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17

Issue #11980, opened 2026-06-08
enh: Additional QueueOrder strategy for kueue-priority-booster


dddwsd:
**What would you like to be added**:
When using the time-sharing fairness mechanism through `kueue-priority-booster`, I would like the queue ordering to be based on a criterion other than `creationTimestamp`.

If this proposal sounds reasonable, I'd be happy to work on an implementation and submit a PR.

**Why is this needed**:
The current behavior of `kueue-priority-booster` is as follows
```
job-sample-job-st6g6-f7381 - create - running
job-sample-job-h9446-f9845 - create - running
job-sample-job-87dhp-fc903 - create - suspend
job-sample-job-r4hkr-a3e27 - create - suspend
job-sample-job-87dhp-fc903 preempt job-sample-job-st6g6-f7381
job-sample-job-st6g6-f7381 clear booster annotation
job-sample-job-st6g6-f7381 preempt job-sample-job-h9446-f9845
job-sample-job-h9446-f9845 clear booster annotation
job-sample-job-h9446-f9845 preempt job-sample-job-87dhp-fc903
```
However, with this approach, `job-sample-job-r4hkr-a3e27` remains suspended and cannot preempt any workload. 
This seems inconsistent with the intended goals of a time-sharing fairness mechanism.

Therefore, I suggest adopting a new queue-ordering strategy that operates as follows:
```
job-sample-job-st6g6-f7381 - create - running
job-sample-job-h9446-f9845 - create - running
job-sample-job-87dhp-fc903 - create - suspend
job-sample-job-r4hkr-a3e27 - create - suspend
job-sample-job-87dhp-fc903 preempt job-sample-job-st6g6-f7381
job-sample-job-st6g6-f7381 clear booster annotation
job-sample-job-r4hkr-a3e27 preempt job-sample-job-h9446-f9845
job-sample-job-h9446-f9845 clear booster annotation
job-sample-job-st6g6-f7381 preempt job-sample-job-87dhp-fc903
```

**Completion requirements**:
- Introduce a new queue-ordering strategy that does not rely solely on `creationTimestamp`.
- Ensure the behavior works consistently with workloads preempted by kueue-priority-booster.

This enhancement requires the following artifacts:

- [v] Design doc
- [v] API change
- [v] Docs update

The artifacts should be linked in subsequent comments.

kannon92:
What do you mean by enh?

dddwsd:
[...]

PR #12653, opened 2026-06-30
KEP: Support Round-Robin Ordering for Pending Workloads in ClusterQueue

dddwsd:
#### What type of PR is this?

/kind documentation

Part of #11980 

```release-note
None
```

## Summary by CodeRabbit

* **Documentation**
  * Added a new design document for optional round-robin ordering of pending workloads in a cluster queue.
  * Documented a new ordering setting with default FIFO behavior, an optional round-robin mode, and a time window before falling back to FIFO.
  * Included rollout details, graduation criteria, and a test plan covering unit, integration, and end-to-end checks.
  * Added the feature definition and lifecycle metadata for the upcoming alpha stage.

### DIFF
--- a/keps/11980-round-robin-ordering/README.md
+++ b/keps/11980-round-robin-ordering/README.md
@@ -1,0 +1,268 @@
+# KEP-11980: Support Round-Robin Ordering for Pending Workloads in ClusterQueue
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories (Optional)](#user-stories-optional)
+    - [Story 1 (Optional)](#story-1-optional)
+    - [Story 2 (Optional)](#story-2-optional)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Test Plan](#test-plan)
+    - [Prerequisite testing updates](#prerequisite-testing-updates)
+    - [Unit tests](#unit-tests)
+    - [Integration tests](#integration-tests)
+    - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+This KEP adds optional `ClusterQueue.spec.ordering` to configure how pending
+workloads with the same effective priority are ordered within a ClusterQueue.
+The default remains FIFO by `creationTimestamp`, while Cluster administrators
+can opt into `RoundRobin` ordering to rotate admission opportunities among
+same-priority pending workloads.
+
+- `FIFO` (default): order by `creationTimestamp` (current behavior).
+- `RoundRobin`: order by the latest `lastTransitionTime` among
+  `Workload.status.conditions`. Workloads with a more recent condition
+  transition are placed behind those with an older latest transition.
+
+When configured, `roundRobinWindow` limits how long `RoundRobin` applies after
+workload creation; older workloads fall back to FIFO by `creationTimestamp`.
+
+## Motivation
+
+By default, Kueue orders pending workloads within a ClusterQueue by
+`creationTimestamp`. This gives the earliest-created workload priority and
+helps it finish sooner under steady load.
+
+However, when same-priority workloads are repeatedly preempted and requeued,
+the workload with the earliest `creationTimestamp` can keep winning admission
+while others wait.
+
+```plaintext
+1. `job-a` (priority=10) is admitted.
+2. `job-b` (priority=10) is queued.
+3. `job-c` (priority=100) arrives and preempts `job-a`.
+4. `job-a` is requeued ahead of `job-b`.
+5. `job-c` completes.
+6. `job-a` is admitted again.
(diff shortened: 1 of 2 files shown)


keps/11980-round-robin-ordering/README.md
```diff
@@ -0,0 +1,268 @@
+# KEP-11980: Support Round-Robin Ordering for Pending Workloads in ClusterQueue
```
mimowo, line 1: Is this about the prioirity booster? If so then we may generalize the KEP a bit to make it about the priority booster itself.

dddwsd: The original issue was motivated by the priority booster. However, as the proposal evolved, I realized that the KEP is fundamentally about the ordering of workloads in the ClusterQueue's internal heap, independent of any priority boost mechanism.

mimowo: Interesting, would you like to maybe present the idea at one the the wg-batch meetings?

keps/11980-round-robin-ordering/README.md
```diff
@@ -0,0 +1,268 @@
...
+## Summary
+
+This KEP adds optional `ClusterQueue.spec.ordering` to configure how pending
+workloads with the same effective priority are ordered within a ClusterQueue.
+The default remains FIFO by `creationTimestamp`, while Cluster administrators
+can opt into `RoundRobin` ordering to rotate admission opportunities among
+same-priority pending workloads.
+
+- `FIFO` (default): order by `creationTimestamp` (current behavior).
+- `RoundRobin`: order by the latest `lastTransitionTime` among
```
mwielgus, line 35: How does it work with best effort FIFO and PreemptionPolicyLowerOrNewerEqualPriority?

keps/11980-round-robin-ordering/README.md
```diff
@@ -0,0 +1,268 @@
...
+- Add `ClusterQueue.spec.ordering` to configure ordering of same-priority
+  pending workloads within a ClusterQueue.
+- Support `spec.ordering.mode`: `FIFO` (default, current behavior) or
+  `RoundRobin`.
+- Support optional `spec.ordering.roundRobinWindow` to apply `RoundRobin` only
+  for a bounded period after workload creation.
+- When `roundRobinWindow` is set, workloads older than
+  `creationTimestamp + roundRobinWindow` fall back to FIFO ordering by
+  `creationTimestamp`.
+
```
mwielgus, line 79: What is the business goal behind this? Why is it better to have a number of half-done tasks instead of predictable and stable order?

keps/11980-round-robin-ordering/README.md
dddwsd: Thank you for the suggestion. Unfortunately, due to personal circumstances, I don't think I'll be able to present at a WG Batch meeting. Would it be possible to continue the discussion through the KEP instead?

keps/11980-round-robin-ordering/README.md
```diff
@@ -0,0 +1,268 @@
...
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories (Optional)](#user-stories-optional)
+    - [Story 1 (Optional)](#story-1-optional)
+    - [Story 2 (Optional)](#story-2-optional)
```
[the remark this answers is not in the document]
dddwsd: Fix in next commit

keps/11980-round-robin-ordering/README.md
dddwsd: The goal is fairness rather than maximizing the number of partially completed workloads. 

In preemption-heavy environments, FIFO can repeatedly favor the same workload after requeue because it retains the oldest creationTimestamp. 

RoundRobin distributes admission opportunities more evenly among same-priority workloads, and roundRobinWindow ensures that older workloads eventually regain priority through FIFO ordering.

keps/11980-round-robin-ordering/README.md
dddwsd: RoundRobin only affects the ordering of workloads in the ClusterQueue pending heap. Therefore, the behavior of BestEffortFIFO and PreemptionPolicyLowerOrNewerEqualPriority remains unchanged.

For example, when a workload is preempted due to BestEffortFIFO or PreemptionPolicyLowerOrNewerEqualPriority, the evicted workload is requeued as it is today. The only difference is that, once requeued, its position within the ClusterQueue pending heap is determined according to the configured ordering mode (FIFO or RoundRobin).

In other words, this proposal does not change when preemption happens or how preemption candidates are selected; it only changes how same-priority pending workloads are ordered after they are placed in the ClusterQueue pending heap.

keps/11980-round-robin-ordering/README.md
dddwsd: @mimowo @mwielgus 

Thought about this more — instead of changing the ClusterQueue's internal ordering logic, I think it'd be simpler for alpha to build this on top of the priority-booster.

Concretely: add a roundRobinWindow to the timeSharing config. Once negativeBoost is set and the workload gets preempted, the annotation isn't cleared immediately — it stays until roundRobinWindow elapses. That keeps the workload's priority lowered for a while after preemption, giving other same-priority pending workloads a chance to get admitted first, before it falls back to normal priority.

Modifying the internal heap felt like it'd need more careful design and review given how central it is to scheduling, so I'd like to avoid that complexity for alpha if possible. This is also a feature we need pretty urgently in our environment, so I'd like to move on it quickly if there's no objection.

Curious what you think about this approach. If it sounds good to you, I'll revise the KEP and push an update.

Issue #12646, opened 2026-06-30
Flaky: AdmissionCheck controller when one clusterQueue references admissionChecks Should delete the admissionCheck when the corresponding clusterQueue is deleted


mbobrovskyi:
**Which test is flaking?**:
Core Controllers Suite: [It] AdmissionCheck controller when one clusterQueue references admissionChecks Should delete the admissionCheck when the corresponding clusterQueue is deleted [controller:admissioncheck, area:core]

**Link to failed CI job or steps to reproduce locally**:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/periodic-kueue-test-integration-shard-1-release-0-17/2071797564218281984

**Failure message or logs**:
```
{ failed [FAILED] Timed out after 10.001s.
The function passed to Eventually failed at /home/prow/go/src/kubernetes-sigs/kueue/test/integration/singlecluster/controller/core/admissioncheck_controller_test.go:116 with:
Expected success, but got an error:
    <*errors.StatusError | 0xc00088c0a0>: 
    admissionchecks.kueue.x-k8s.io "check1" not found
    {
        ErrStatus: {
            TypeMeta: {Kind: "", APIVersion: ""},
            ListMeta: {
                SelfLink: "",
                ResourceVersion: "",
                Continue: "",
                RemainingItemCount: nil,
            },
            Status: "Failure",
            Message: "admissionchecks.kueue.x-k8s.io \"check1\" not found",
            Reason: "NotFound",
            Details: {
                Name: "check1",
                Group: "kueue.x-k8s.io",
                Kind: "admissionchecks",
                UID: "",
                Causes: nil,
                RetryAfterSeconds: 0,
            },
            Code: 404,
        },
    }
In [It] at: /home/prow/go/src/kubernetes-sigs/kueue/test/integration/singlecluster/controller/core/admissioncheck_controller_test.go:118 @ 06/30/26 03:33:33.335
}
```

**Anything else we need to know?**:

apullo777:
This looks like the same flake as #11694, fixed by #11731.
It seems the fix was requested for cherry-pick but does not seem to have landed there.

mimowo:
Yes it does, would you like to submit a PR? You may also proactively search for more cases like these, but I'm not sure how to do that systematically, so plumbing the known ones is reasonable for now.

mimowo:
[...]

PR #12655, opened 2026-06-30
test: wait for Topology finalizer before deletion in TAS tests

apullo777:
#### What type of PR is this?

/kind cleanup
/area testing

Adds a finalizer-wait guard before the two Topology deletions in the TAS re-creation tests.

The Topology in-use finalizer is added asynchronously by the reconciler, so waiting before `Delete` makes these tests consistent with the sibling Topology-deletion test and avoids a delete-before-finalizer race under delayed reconciliation.

This is test hardening rather than an observed CI flake. Related to #12646.

Fixes #

```release-note
NONE
```


mimowo:
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for the update to harden our build system 👍

PR #12657, opened 2026-06-30
[release-0.17] Fix StatefulSet scale-to-zero workloads holding quota and requeue jitter

### LINKED ISSUE #12232: StatefulSet scale-to-zero can leave admitted Workload reserving quota and block scale-up

gola:
#### What type of PR is this?

/kind bug
/area integrations
/area testing

Cherry pick of #12233 on release-0.17.

The automatic cherry-pick failed because of conflicts around release-0.17 differences. This manual backport resolves those conflicts and adds a small compatibility commit for the release-0.17 test/helper APIs.

Fixes #12232

The original automatic cherry-pick to release-0.17 failed in #12233:
https://github.com/kubernetes-sigs/kueue/pull/12233#issuecomment-4840770854

Local verification:

```bash
go test ./pkg/workload ./pkg/dra ./pkg/controller/core ./pkg/controller/jobs/statefulset
KUBEBUILDER_ASSETS="/home/guowei/.local/share/kubebuilder-envtest/k8s/1.35.0-linux-amd64" go test ./test/integration/singlecluster/controller/jobs/statefulset
```

```release-note
StatefulSet: Fixed a bug where scaling a StatefulSet to zero caused its Workload to be incorrectly requeued for scheduling during the terminating-pod window, competing for quota it should no longer hold.
```

### DIFF
--- a/apis/kueue/v1beta1/workload_types.go
+++ b/apis/kueue/v1beta1/workload_types.go
@@ -665,6 +665,13 @@ const (
 	// WorkloadDeactivationTarget means that the Workload should be deactivated.
 	// This condition is temporary, so it should be removed after deactivation.
 	WorkloadDeactivationTarget = "DeactivationTarget"
+
+	// WorkloadRequeueHeld means that the Workload should not be requeued
+	// after its quota reservation is released. The possible reasons for this
+	// condition are:
+	// - "StatefulSetScaledDown": the owning StatefulSet scaled to zero
+	// Other integrations may use additional reasons in the future.
+	WorkloadRequeueHeld = "RequeueHeld"
 )
 
 // Reasons for the WorkloadPreempted condition.
--- a/apis/kueue/v1beta1/workload_types.go
+++ b/apis/kueue/v1beta1/workload_types.go
@@ -665,13 +665,6 @@ const (
 	// WorkloadDeactivationTarget means that the Workload should be deactivated.
 	// This condition is temporary, so it should be removed after deactivation.
 	WorkloadDeactivationTarget = "DeactivationTarget"
-
-	// WorkloadRequeueHeld means that the Workload should not be requeued
-	// after its quota reservation is released. The possible reasons for this
-	// condition are:
-	// - "StatefulSetScaledDown": the owning StatefulSet scaled to zero
-	// Other integrations may use additional reasons in the future.
-	WorkloadRequeueHeld = "RequeueHeld"
 )
 
 // Reasons for the WorkloadPreempted condition.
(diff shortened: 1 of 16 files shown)


mimowo:
injected the release note manually
/kind bug

gola:
Thanks @mimowo for the quick review and approval!

PR #12659, opened 2026-06-30
Preparatory - workload observability flow for UnadmittedWorkloadObservability feature gate

j-skiba:
#### What type of PR is this?

/kind feature

This is a preparatory PR based on https://github.com/kubernetes-sigs/kueue/pull/12510.

This PR introduces dynamic resolution of reasons for QuotaReserved condition based on UnadmittedWorkloadsObservability feature gate and adds some preparatory functions for the following PRs

Part of #https://github.com/kubernetes-sigs/kueue/issues/10852

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Admission checks now validate workloads against the ClusterQueue namespace selector, improving outcomes when namespaces don’t match.

* **Bug Fixes**
  * Unadmitted and requeued workloads now show more precise quota-reservation and admitted-condition reasons (e.g., misconfigured vs suspended; pending vs waiting), including DRA-related and queue-state transitions.
  * “Pending/Waiting” messaging is normalized for quota-reservation unset and admission-gated scenarios, with improved behavior when unadmitted-workloads observability is disabled.

### DIFF
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -1069,7 +1069,10 @@ func (r *WorkloadReconciler) mayUpdateConditionForAdmissionGatedBy(ctx context.C
 	if !hasGatedAnnotation && hasGatedCondition {
 		// This previously gated workload is becoming admissible because its AdmissionGatedBy annotation is cleared
 		err := workloadpatching.PatchAdmissionStatus(ctx, r.client, wl, r.clock, func(wl *kueue.Workload) (bool, error) {
-			reason := workload.UnadmittedWorkloadReasonWithFallback(kueue.WorkloadQuotaReservedReasonPendingEvaluation, kueue.WorkloadPending)
+			reason := workload.UnadmittedWorkloadReasonWithFallback(
+				kueue.WorkloadQuotaReservedReasonPendingEvaluation,
+				kueue.WorkloadPending, //nolint:staticcheck // SA1019: fallback
+			)
 			// Update the condition to indicate the gate is cleared, rather than removing it
 			return workload.UnsetQuotaReservationWithCondition(wl, reason, "AdmissionGatedBy cleared, waiting for quota reservation", r.clock.Now()), nil
 		})
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -1069,7 +1069,7 @@ func (r *WorkloadReconciler) mayUpdateConditionForAdmissionGatedBy(ctx context.C
 	if !hasGatedAnnotation && hasGatedCondition {
 		// This previously gated workload is becoming admissible because its AdmissionGatedBy annotation is cleared
 		err := workloadpatching.PatchAdmissionStatus(ctx, r.client, wl, r.clock, func(wl *kueue.Workload) (bool, error) {
-			reason := workload.UnadmittedWorkloadReasonWithFallback(kueue.WorkloadQuotaReservedReasonPendingEvaluation, "Pending")
+			reason := workload.UnadmittedWorkloadReasonWithFallback(kueue.WorkloadQuotaReservedReasonPendingEvaluation, kueue.WorkloadPending)
 			// Update the condition to indicate the gate is cleared, rather than removing it
 			return workload.UnsetQuotaReservationWithCondition(wl, reason, "AdmissionGatedBy cleared, waiting for quota reservation", r.clock.Now()), nil
 		})
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -707,12 +707,6 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 				)
 			}
 		}
-		// Return early if the status was updated and either:
-		// 1. The observability feature gate is disabled (legacy behavior).
-		// 2. The workload actually transitioned to Admitted.
-		if updated && (!features.Enabled(features.UnadmittedWorkloadsObservability) || isAdmitted) {
-			return ctrl.Result{}, nil
-		}
 	}
 
 	if workload.HasQuotaReservation(&wl) {
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -321,7 +321,8 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		workload.HasDRA(&wl) {
 		log.V(3).Info("Rejecting workload that uses DRA resources because KueueDRAIntegration feature gate is disabled")
 		err := workloadpatching.PatchAdmissionStatus(ctx, r.client, &wl, r.clock, func(wl *kueue.Workload) (bool, error) {
-			updated := workload.UnsetQuotaReservationWithCondition(wl, kueue.WorkloadInadmissible,
+			reason := workload.UnadmittedWorkloadReasonWithFallback(kueue.WorkloadQuotaReservedReasonMisconfigured, kueue.WorkloadInadmissible)
+			updated := workload.UnsetQuotaReservationWithCondition(wl, reason,
 				"Workload uses DRA resources but the KueueDRAIntegration feature gate is not enabled",
 				r.clock.Now())
 			if workload.SetRequeuedCondition(wl, kueue.WorkloadInadmissible,
@@ -340,7 +341,8 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		if workload.HasResourceClaim(&wl) {
 			log.V(3).Info("Workload is inadmissible because it uses resource claims which is not supported")
 			err := workloadpatching.PatchAdmissionStatus(ctx, r.client, &wl, r.clock, func(wl *kueue.Workload) (bool, error) {
-				updated := workload.UnsetQuotaReservationWithCondition(wl, kueue.WorkloadInadmissible, "KueueDRAIntegration feature does not support use of resource claims", r.clock.Now())
+				reason := workload.UnadmittedWorkloadReasonWithFallback(kueue.WorkloadQuotaReservedReasonMisconfigured, kueue.WorkloadInadmissible)
+				updated := workload.UnsetQuotaReservationWithCondition(wl, reason, "KueueDRAIntegration feature does not support use of resource claims", r.clock.Now())
 				if updated && workload.SetRequeuedCondition(wl, kueue.WorkloadInadmissible, "DRA resource claims not supported", false) {
 					updated = true
 				}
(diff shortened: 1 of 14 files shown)


pkg/workload/resources.go
```diff
@@ -182,3 +186,31 @@ func ValidateLimitRange(ctx context.Context, c client.Client, wi *Info) field.Er
 	}
 	return allErrs
 }
+
+// ValidateAdmissibility checks if the workload's namespace matches the ClusterQueue's
+// namespace selector, and if its resource requests are valid and satisfy LimitRanges.
+// Returns the admissibility error message and a boolean indicating if it was a namespace mismatch.
+func ValidateAdmissibility(
```
j-skiba, line 206: https://github.com/kubernetes-sigs/kueue/pull/12510#discussion_r3497535014

pkg/controller/core/workload_controller.go
```diff
@@ -320,7 +320,8 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		workload.HasDRA(&wl) {
 		log.V(3).Info("Rejecting workload that uses DRA resources because KueueDRAIntegration feature gate is disabled")
 		err := workloadpatching.PatchAdmissionStatus(ctx, r.client, &wl, r.clock, func(wl *kueue.Workload) (bool, error) {
-			updated := workload.UnsetQuotaReservationWithCondition(wl, kueue.WorkloadInadmissible,
+			reason := workload.UnadmittedWorkloadReasonWithFallback(kueue.WorkloadQuotaReservedReasonMisconfigured, kueue.WorkloadInadmissible)
```
mwielgus, line 324: Can  you extend the tests to cover the reason?

pkg/workload/resources.go
```diff
@@ -182,3 +186,31 @@ func ValidateLimitRange(ctx context.Context, c client.Client, wi *Info) field.Er
...
+	ctx context.Context,
+	c client.Client,
+	wi *Info,
+	cqNamespaceSelector labels.Selector,
+) (string, bool) {
```
mimowo: Could we return the exact reason instead of the bool, to alllow easier matching outside the function

j-skiba: the bool is used just to decide if we want to set `requeueReason` of an `entry`, and that is necessary just in scheduler.go.

The `quotaReservedReason` is set to `Misconfigured` if `inadmissibleMsg` is non empty.

Alternatively we could refactor it to something like this

```go
func ValidateAdmissibility(...) error {
    // ...
    if cqNamespaceSelector != nil && !cqNamespaceSelector.Matches(labels.Set(ns.Labels)) {
        return ErrNamespaceMismatch
    }
    if errs := ValidateResources(wi); len(errs) > 0 {
        return fmt.Errorf("%s: %v", ErrInvalidWLResources, errs.ToAggregate())
    }
    // ...
    return nil
}
```

and in scheduler we would have 

```go
} else if err := workload.ValidateAdmissibility(ctx, s.client, &w, e.clusterQueueSnapshot.NamespaceSelector); err != nil {
    e.inadmissibleMsg = err.Error()
    e.quotaReservedReason = kueue.WorkloadQuotaReservedReasonMisconfigured
    if errors.Is(err, workload.ErrNamespaceMismatch) {
        e.requeueReason = qcache.RequeueReasonNamespaceMismatch
    }
}
```

pkg/controller/core/workload_controller.go
```diff
@@ -1052,8 +1064,9 @@ func (r *WorkloadReconciler) mayUpdateConditionForAdmissionGatedBy(ctx context.C
 	if !hasGatedAnnotation && hasGatedCondition {
 		// This previously gated workload is becoming admissible because its AdmissionGatedBy annotation is cleared
 		err := workloadpatching.PatchAdmissionStatus(ctx, r.client, wl, r.clock, func(wl *kueue.Workload) (bool, error) {
+			reason := workload.UnadmittedWorkloadReasonWithFallback(kueue.WorkloadQuotaReservedReasonPendingEvaluation, "Pending")
```
kshalot: AFAICS `"Pending"` is the only reason that we hard-code as a string literal in the code instead of using a constant.

So nit: maybe this is the opportunity to align that?

kshalot: Actually, same for `"Waiting"`.

pkg/workload/resources.go
kshalot: Returning error makes sense to me. I was originally going to comment that this function is mixing a lot of different checks, but now I think for re-usability in the other PR this might be necessary.

But the signature is currently pretty cryptic IMO, the bool especially (even more so if we want to re-use this outside the scheduler).

pkg/controller/core/workload_controller.go
j-skiba: sounds good, extracted to consts

pkg/workload/resources.go
j-skiba: I refactored the function to return just an error

pkg/controller/core/workload_controller.go
```diff
@@ -701,7 +707,10 @@ func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 				)
 			}
 		}
-		if updated {
+		// Return early if the status was updated and either:
+		// 1. The observability feature gate is disabled (legacy behavior).
+		// 2. The workload actually transitioned to Admitted.
+		if updated && (!features.Enabled(features.UnadmittedWorkloadsObservability) || isAdmitted) {
```
j-skiba: in the new flow if the workload has some quota reservation issues and didn't have Admitted condition yet then SyncAdmittedCondition

https://github.com/kubernetes-sigs/kueue/pull/12659/changes#diff-60dd240c20adbd6a189d018d1c216c2d296730f446c341d8bf449fa6657964ffL662

would set it and if we didn't have `|| isAdmitted` then the reconcile would exit early and not set reasons that are set outside this if condition branch

mimowo: Hm, two questions:

> if we didn't have || isAdmitted then the reconcile would exit early and not set reasons that are set outside this if condition branch

1. Yes, but that would happen on the next reconcile - we know the next reconcile will happen because we have an update, right? So from user PoV keeping `if updated` will pretty much result in the same behavior. (IIUC)?

2.  Do you know why we need the "if" in the first place - what bad would happen if we deleted it altogether and just let the logic flow?

j-skiba: 1. yes, that would happen on the next reconcile. There's also a thing that currently in unit tests we run `Reconcile` once on a workload and to see the expected final result we would have to run the `Reconcile` twice for some test cases.
2. I don't know exactly, removing this check doesn't break any of the tests locally. Technically if a first update runs then another components might be triggered and modify the object but that could happen even without this check. Maybe we should somehow gather the changes and push them in one API call?

mimowo: Reading the code again I think this check is not necessary -  I'm ok to drop it entirely in this PR. I tried to trace it with git blame, and it seems it got there where the paths for handling admitted and non-admitted workloads were separate.  

Ideally, yes, we would collect all the changes and only issue a single workload update. This pattern is for example implemented in the k8s Job - [syncJob](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/job/job_controller.go#L911) is the Reconcile. That change would be compatible with dropping the "if", but it is much more work, so I would leave it as a follow up.

pkg/controller/core/workload_controller.go
j-skiba: sure, tests added

pkg/controller/core/workload_controller.go
mimowo: For the follow up - gathering all all Workload changes and submitting a single API request at the end.

Feel free to open an issue / or PR (depending on your capacity) towards that refactoring when you see path forward to how this can be executed in practice.

mimowo:
/lgtm
/approve
Thank you for this prep PR, looks great 👍

pkg/controller/core/workload_controller.go
j-skiba: `if` branch removed

j-skiba:
@coderabbitai @mimowo 

```
Inline comments:
In `@pkg/controller/core/workload_controller.go`:
- Around line 1334-1335: The on-hold branch in workload reconciliation is only
skipping AddOrUpdateWorkload, which leaves any previously queued pending
workload entry behind. Update the onHold case in workload_controller.go to
actively remove or cancel the existing queue entry for that workload before
returning, so a workload transitioning to WorkloadOnHold cannot still be
scheduled from stale queue state. Use the existing queue/reconciliation logic
around AddOrUpdateWorkload and the onHold switch branch to locate the fix.
```
I think would be better to fix in a separate PR

j-skiba:
@coderabbitai 

> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> `@j-skiba` Sounds good, that's a reasonable scope call for a prep PR — no need to hold this one up for it. Would you like me to open a follow-up issue to track fixing the on-hold branch's stale queue entry cleanup, so it doesn't get lost? I can link it back to this PR and comment for context.

yes

pkg/scheduler/scheduler.go
```diff
@@ -631,9 +632,13 @@ func (s *Scheduler) nominate(ctx context.Context, workloads []workload.Info, sna
 			e.quotaReservedReason = kueue.WorkloadQuotaReservedReasonMisconfigured
 		} else if err := workload.ValidateAdmissibility(ctx, s.client, &w, e.clusterQueueSnapshot.NamespaceSelector); err != nil {
 			e.inadmissibleMsg = err.Error()
-			e.quotaReservedReason = kueue.WorkloadQuotaReservedReasonMisconfigured
-			if errors.Is(err, workload.ErrNamespaceMismatch) {
-				e.requeueReason = qcache.RequeueReasonNamespaceMismatch
+			if errors.Is(err, workload.ErrAPI) {
+				e.skipStatusUpdate = true
```
mimowo: Is this a change of the semantics of handling API errors compared to what we have released? 

So that currently we would put the API server error message in to the QuotaReserved=False message, but after this change we will skip?

j-skiba: hmm yes, actually I'm wondering what's better now:
- state of main - we set Misconfigured no matter what's wrong during validation
- on this branch - we skip status update as this might be a transient problem

the only thing I'm hesitant about if this skip logic might become a permanent state but I think it shouldn't, missing namespace should not be permanent state.

the code from main might be flapping in some cases but it shouldn't be too frequent

mimowo: I think we shouldn't use Misconfigured - that would be misleading.  If we want to use a pre-existing reason, then probably Pendingevaluation fits better. I would probably use this for the PR, however, I think we need something more detailed reason like `APIError`. Could be a follow up.

> the only thing I'm hesitant about if this skip logic might become a permanent state but I think it shouldn't, missing namespace should not be permanent state.

Hm, but if scheduler considers any of the errors , including missing namespace, as transient APIError, then it should be ok, because the workload_controller which has more detailed reasons will set the dedicated reason, right?

j-skiba: Setting `PendingEvaluation` sounds ok but there's one more thing for this approach, what should we do with precedence rules for now in this PR? Should we just set PendingEvaluation or set only if there is no QuotaReserved condition yet? PendingEvaluation has the lowest priority power.

`APIError` sounds like a solution to this problem. I like the idea.

j-skiba: alternatively: we could skip status update OR set `PendingEvaluation` if there is no `QuotaReserved` condition yet on API error

mimowo: Ok, I checked this code again carefully, and I think:
1. skipping the status update is ok, this is consitent with how we handle an error on admit() which arguably is the same category error - yes, this is a semantic change but I think it can be justified here, as this is a corner case (InternalErrors) handling anyway. The case for admit() error may or may not be internal error.
2. rename it to "InternalError" because the Get calls used there are not really connecting to the API server, but taking data from the cache. 
3. add a log.Error for such type of error - currently seems missing

pkg/scheduler/scheduler.go
mimowo, line 1044: @j-skiba do we have it planned to also update the reasons for the events? I think it would be great to introduce the consistency between Conditions, metrics and also events.

It might be a follow up surely.

pkg/scheduler/scheduler.go
j-skiba: ok, changes:

- renamed `workload.ErrAPI` -> `workload.ErrInternal`
- added error log
   ```go
   log.Error(err, "Failed to validate workload admissibility")
   e.skipStatusUpdate = t
   ```

pkg/scheduler/scheduler.go
j-skiba: yes, that will be covered in a separate PR

PR #12661, opened 2026-06-30
feat: add times for makefile steps

### LINKED ISSUE #12660: Add timestamps to show how long individual makefile steps take

Add timestamps to show how long individual makefile steps take

**Why is this needed**:

Currently we have no visibility how long indivual steps take, for example see the log https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12658/pull-kueue-test-e2e-multikueue-extended-shard-1-main/2071920183638757376/build-log.txt

yuluo-yx:
#### What type of PR is this?

/area testing

Fixes #12660

```release-note
None
```

## Summary by CodeRabbit

* **New Features**
  * Added optional timing output for `make` runs, automatically enabled in CI and configurable with environment variables.
  * Timing information is only shown for slower runs or failures, keeping normal output cleaner.

* **Tests**
  * Added a verification step to cover the new timed shell behavior and keep `make verify` aligned with the updated execution flow.

### DIFF
--- a/hack/make-timed-shell.sh
+++ b/hack/make-timed-shell.sh
@@ -1,0 +1,71 @@
+#!/usr/bin/env bash
+
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o nounset
+set -o pipefail
+
+is_truthy() {
+  case "${1:-}" in
+    1 | true | TRUE | yes | YES | on | ON)
+      return 0
+      ;;
+    *)
+      return 1
+      ;;
+  esac
+}
+
+command_preview() {
+  local command=$1
+
+  command=${command//$'\n'/; }
+  command=${command//$'\t'/ }
+  if [[ "${#command}" -gt 240 ]]; then
+    printf '%s...' "${command:0:237}"
+  else
+    printf '%s' "${command}"
+  fi
+}
+
+if ! is_truthy "${MAKE_TIMING:-}"; then
+  exec /usr/bin/env bash -o pipefail "$@"
+fi
+
+command=""
+if [[ "$#" -gt 0 ]]; then
+  command="${!#}"
+fi
+preview=$(command_preview "${command}")
+min_seconds="${MAKE_TIMING_MIN_SECONDS:-1}"
+if ! [[ "${min_seconds}" =~ ^[0-9]+$ ]]; then
+  min_seconds=1
+fi
+
+start_epoch=$(date +%s)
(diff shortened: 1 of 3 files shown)


hack/make-timed-shell.sh
```diff
@@ -0,0 +1,71 @@
...
+
+/usr/bin/env bash -o pipefail "$@"
+status=$?
+
+end_epoch=$(date +%s)
+end_time=$(date -u '+%Y-%m-%dT%H:%M:%SZ')
+duration=$((end_epoch - start_epoch))
+if [[ "${status}" -ne 0 || "${duration}" -ge "${min_seconds}" ]]; then
+  printf 'make-timing: started=%s finished=%s status=%d duration=%ss command=%s\n' \
+    "${start_time}" "${end_time}" "${status}" "${duration}" "${preview}" >&2
```
mimowo, line 68: This works nicely and already provides some useful information where we might be optimizing. 

I would like to keep it, but I'm wondering maybe the data sometimes is lacking resolution when we call bash scripts, for example the script to run tests executes multiple bash scripts which we don't capture with the current approach.

So I'm wondering if the following would be feasible to  stamp every log line with timestamp, as we do using regular loggers, I think there is the "ts" command, wdyt?

yuluo-yx: While `ts` could solve the per-line log timestamping issue, it is part of `moreutils`; Kueue does not currently have this dependency, nor is its availability guaranteed in local or Prow environments. To avoid introducing an additional tool dependency into the verification/CI pipeline, I prefer to keep the current self-contained wrapper for now.

For finer-grained log timestamps, we can implement a self-contained per-line timestamping mechanism within the wrapper at a later stage.

How about this 🤔

mimowo: > While ts could solve the per-line log timestamping issue, it is part of moreutils; Kueue does not currently have this dependency, nor is its availability guaranteed in local or Prow environments. To avoid introducing an additional tool dependency into the verification/CI pipeline, I prefer to keep the current self-contained wrapper for now.

I see, thanks for the research, that sgtm.

> For finer-grained log timestamps, we can implement a self-contained per-line timestamping mechanism within the wrapper at a later stage.

sgtm

mimowo: Another path to explore then - would it be feasible to measure timings for individual bash commands?

I'm thinking how to break down this entry: `make-timing: started=2026-06-30T15:52:06Z finished=2026-06-30T15:59:25Z status=0 duration=439s command=E2E_KIND_VERSION="kindest/node:v1.36.1" KIND_CLUSTER_NAME=kind \;  ARTIFACTS="/logs/artifacts/run-test-e2e-extended-1.36.1" IMAGE_TAG=us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue:v0.19.0-devel-274-ga0455746e GINKGO_ARGS="--l...`

from https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12661/pull-kueue-test-e2e-extended-shard-2-main-1-36/2071983779773157376/build-log.txt

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you, I think this will be useful to see the room to optimize 👍

mimowo:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12661/pull-kueue-test-e2e-baseline-main-1-34/2072000770030440448
this is not related
/test pull-kueue-test-e2e-baseline-main-1-34

PR #1469, opened 2023-12-15
Added security insight

ashish493:
#### What type of PR is this?

Signed-off-by: Ashish Malik [malikashish493@gmail.com](mailto:malikashish493@gmail.com)
#SecuritySlam
#### What this PR does / why we need it:
This PR adds security insights yml file
#### Which issue(s) this PR fixes:

This commit is part of the December 2023 Security Slam.

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #12669, opened 2026-06-30
Add in-scope/out-of-scope to SECURITY-INSIGHTS.yaml

### LINKED ISSUE #1473: Add in-scope/out-scope subsections to SECURITY-INSIGHTS.yml
`#SecuritySlam`

When the SECURITY-INSIGHTS.yml file was initially created in https://github.com/kubernetes-sigs/kueue/pull/1469, the in-scope / out-scope subsections of the vulnerability-reporting section were omitted, as it was not entirely clear which of the [OWASP Top 10](https://owasp.org/Top10/) were in/out of scope. Someone with more knowledge of the project should update the SECURITY-INSIGHTS.yml file to include this section.

**Why is this needed**:

Adding this section will improve the project's score on the CLOMonitor site.

/sig release

iacker:
#### What type of PR is this?

/kind cleanup
/sig release

The `vulnerability-reporting` section of `SECURITY-INSIGHTS.yaml` has been missing
its `in-scope` and `out-of-scope` subsections since the file was first added in
#1469, where they were intentionally deferred because it was not yet clear which
classes of issues were in or out of scope. This PR fills them in so the Security
Insights document is complete and Kueue's CLOMonitor score improves.

The scope is described in terms of Kueue's actual attack surface:

- **In-scope:** the Kueue controller-manager and its admission/validating
  webhooks, tenant isolation across ClusterQueues/LocalQueues/Cohorts (quota
  borrowing or preemption affecting another tenant's workloads), RBAC defaults in
  the shipped manifests and Helm chart, and the optional KueueViz dashboard
  (backend and frontend).
- **Out-of-scope:** vulnerabilities in Kubernetes core (delegated to the upstream
  Kubernetes security process at https://kubernetes.io/security/), issues that
  only affect unsupported Kueue releases, insecure configuration in a consuming
  cluster, and reports that require pre-existing privileged (cluster-admin)
  access.

While here, this also refreshes `header.expiration-date`, which had expired on
2024-09-28. An expired Security Insights document is flagged by CLOMonitor, so
leaving it stale would partially defeat the purpose of this change. Maintainers
should feel free to adjust the date or the scope wording to match their intent.

Fixes #1473

The scope items reflect my reading of the project's architecture and the
Kubernetes subproject security conventions; please correct anything that does not
match the maintainers' intended reporting scope. This is the kind of section that
benefits from a maintainer's explicit sign-off, per the original issue.

> AI usage disclosure (per the project's AI contribution policy): I used an AI
> assistant to help draft the scope wording and the prose in this PR. The
> technical content reflects my own review of Kueue's architecture, and I take
> responsibility for the change.

```release-note
NONE
```

## Summary by CodeRabbit

* **Documentation**
  * Updated the security policy expiration date to a newer review cycle.
  * Expanded reporting guidance by clearly defining what security issues are considered in-scope versus out-of-scope, including examples of allowed and excluded categories.

### DIFF
--- a/SECURITY-INSIGHTS.yaml
+++ b/SECURITY-INSIGHTS.yaml
@@ -1,6 +1,6 @@
 header:
   schema-version: 1.0.0
-  expiration-date: '2024-09-28T01:00:00.000Z'
+  expiration-date: '2027-06-26T01:00:00.000Z'
   last-updated: '2026-06-26'
   last-reviewed: '2026-06-26'
   commit-hash: "444058af81a08235306ef2a43581692289cd4822"
@@ -57,6 +57,25 @@ vulnerability-reporting:
   Security-policy: 'https://kubernetes.io/security/'
   bug-bounty-available: true
   bug-bounty-url: 'https://hackerone.com/kubernetes'
+  in-scope:
+    - Privilege escalation or workload admission bypass in the Kueue
+      controller-manager and its admission/validating webhooks
+    - Tenant isolation breakout across ClusterQueues, LocalQueues, and Cohorts
+      (e.g. quota borrowing or preemption affecting workloads of another tenant)
+    - Improper RBAC defaults in the shipped manifests and Helm chart that grant
+      excessive permissions to Kueue components
+    - Vulnerabilities in the optional KueueViz dashboard backend and frontend
+      (authentication, access control, or injection)
+  out-of-scope:
+    - Vulnerabilities in Kubernetes core, the kube-apiserver, or other core
+      Kubernetes components, which are handled by the Kubernetes security
+      process at https://kubernetes.io/security/
+    - Issues that only affect unsupported Kueue releases (see the release policy
+      in RELEASE.md); only the latest supported minor versions receive fixes
+    - Insecure configuration in a consuming cluster, such as disabling webhooks,
+      relaxing the provided RBAC, or running components with custom privileges
+    - Reports requiring pre-existing cluster-admin or equivalent privileged
+      access, since such access already permits the reported impact
 dependencies:
   third-party-packages: true
   dependencies-lists:


SECURITY-INSIGHTS.yaml
```diff
@@ -57,6 +57,25 @@ vulnerability-reporting:
   Security-policy: 'https://kubernetes.io/security/'
   bug-bounty-available: true
   bug-bounty-url: 'https://hackerone.com/kubernetes'
+  in-scope:
+    - Privilege escalation or workload admission bypass in the Kueue
+      controller-manager and its admission/validating webhooks
```
mimowo: Maybe worth to split this into two points as this seems mixing two things:
- Privilege escalation or RBAC bypass for viewing or editing the restricted information
// for example we had a case for viewing restricted information via KueueViz)
- Workload quota bypass at admission time, or post-admission time 
// for example cases when a user may admit workload with 1pod, but later increase the size of the workload to 2 pods which still continue to run

mimowo (commented):
Nice update, thank  you👍 

cc @dkaluza @kannon92 @amy

mimowo:
cc @tenzen-y

SECURITY-INSIGHTS.yaml
```diff
@@ -57,6 +57,25 @@ vulnerability-reporting:
...
+    - Tenant isolation breakout across ClusterQueues, LocalQueues, and Cohorts
+      (e.g. quota borrowing or preemption affecting workloads of another tenant)
```
dkaluza: I wonder if maybe it is also worth it to add Quota hijacking or other form of overriding increasing without admin access explicitly. 

This might not be connected to tenant isolation - you do not affect other quotas & their workloads if for example preemption is disabled/they have highest priority.

Admission still works as intended - just you are getting resources that were not granted to you.

SECURITY-INSIGHTS.yaml
iacker: Done — split into two separate items: (1) privilege escalation / RBAC bypass for viewing or editing restricted information (with the KueueViz example), and (2) workload quota bypass at admission or post-admission time (with the pod-count increase example). PTAL.

SECURITY-INSIGHTS.yaml
iacker: Good point — added quota hijacking as an explicit in-scope item, worded to cover getting resources that were not granted even without affecting other tenants' quotas or workloads.

SECURITY-INSIGHTS.yaml
```diff
@@ -57,6 +57,32 @@ vulnerability-reporting:
...
+  in-scope:
+    - Privilege escalation or RBAC bypass allowing viewing or editing of
+      restricted information managed by Kueue (for example, viewing restricted
+      information via KueueViz)
+    - Workload quota bypass at admission time or post-admission (for example,
+      admitting a workload within quota and later increasing its resource usage
+      beyond what was admitted)
+    - Quota hijacking, or otherwise consuming resources that were not granted,
+      without admin access, even when other tenants' quotas and workloads are
+      not affected
```
mimowo: I think these two points overlap strongly, maybe we could combine them into one:

```
- Quota enforcement bypass even when other tenants' quotas and workloads are not affected; for example, admitting a workload beyond its granted quota, increasing its resource usage after admission.
```
wdyt @dkaluza ?

dkaluza: Hmmm, for me this wording wouldn't cover the situation that because of some vulnerability I would be able to override my Quota, which IMO is better covered by current fragment.

I'm fine with merging them, but from my point of view we should cover this case, especially that we are aiming to add complexity to quotas with automatic computation etc

mimowo: So you mean that workloads respect quota, but the quota itself if hijacted (say doubled?)

I didn't think of that, so we may indeed keep these two separated:
    - Workload quota bypass at admission time or post-admission (for example,
      admitting a workload within quota and later increasing its resource usage
      beyond what was admitted)
    - Quota hijacking without admin access, even when other tenants' quotas and workloads are
      not affected; for example assigning yourself more quota than intended by the admin.

I guess "or otherwise consuming resources that were not granted," is more towards the first point then, and we could skip from the second, wdyt?

dkaluza: Yes this is what I meant exactly, sounds good to me!

mimowo:
/test pull-kueue-test-e2e-tas-extended-shard-0-main
infra

SECURITY-INSIGHTS.yaml
mimowo: @iacker I say the latest update: https://github.com/kubernetes-sigs/kueue/compare/09e9b44ccaa9659e61037c4afafd4491b9ee1408..dd53e9cb459e7a5c8d07a1c39e10bf24295f61f9

which is pretty different than we proposed - much more concise and without examples. I'm not sure this is good or bad change because I don't have much experience writing such security guides. 

However, could you please justify that change? 

@dkaluza do you have any opinion here? I personally feel the concise desciption is ok - probably it depends on the target audience. A user of Kueue may not necessarily care about the examples, just about the principles. A contributor may be guided however by more detailed list with examples. 

Maybe worth checking how k8s core does it or other leading OSS projects.

dkaluza: Generally when it comes to definition of scopes from my point of view it is not about exact wording - it is about the intent that we try to communicate - "What we consider in scope".
And because the wording might be sometimes misleading or not fully accurate in my opinion examples are really valuable for the reader to better understand the intent. 

Also the targeted audience are rather contributors & security specialists so I would consider those valuable for them.

Moreover according to the [OpenSSF](https://openssf.org/projects/security-insights/) the aim of this file it to provide machine readable metadata about the project. I wouldn't be surprised if security agents are already scanning those. From my experience with LLMs the more detailed examples we provide the better usually results we will get. 

So taking those into account I would personally strongly vote  for the more descriptive version with examples.

SECURITY-INSIGHTS.yaml
mimowo, line 62:
```suggestion
    - Privilege escalation or RBAC bypass allowing viewing or editing of
      restricted information managed by Kueue (for example, viewing restricted
      information via KueueViz)
    - Workload quota bypass at admission time or post-admission (for example,
      admitting a workload within quota and later increasing its resource usage
      beyond what was admitted)
    - Quota hijacking, or otherwise consuming resources that were not granted,
      without admin access, even when other tenants' quotas and workloads are
      not affected
```
following the discussion  in the thread: https://github.com/kubernetes-sigs/kueue/pull/12669#discussion_r3518498332

mimowo: cc @iacker

iacker:
The branch has diverged significantly from main — rebase now requires resolving hundreds of conflicts across vendor and core files. Given the scope of this PR (adding security-insights scope metadata), the effort/benefit ratio doesn't justify maintaining it.

Closing this PR. If the metadata addition is still wanted, the change can be re-applied cleanly from a fresh branch off current main in a few minutes.

iacker:
Closing due to extensive rebase conflicts. See previous comment.

mimowo:
@iacker friendly ping about the replacement PR

iacker:
Replacement opened cleanly off current main: #13069. It carries the final agreed wording from this thread (the descriptive in-scope/out-scope items with examples). PTAL 🙏

PR #12671, opened 2026-07-01
Fix AFS ConsumedResources CPU truncating to zero on self-requeue

sohankunkerkar:
#### What type of PR is this?

/kind bug

The LocalQueue `Update` predicate always returns `true`, including for status-only updates. When AFS writes `ConsumedResources` to status, the Update event re-enters `Reconcile` within ~40ms instead of the configured `usageSamplingInterval` (5s). The tiny `elapsed` makes the decay alpha near zero, truncating CPU to 0m while memory (much larger in milli-units) keeps rising.

Failing upgrade CI job in downstream (46/49 passed, AFS test failed 3x): https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_kueue-operator/2134/pull-ci-openshift-kueue-operator-release-1.4-upgrade-from-4.22-e2e-operator-upgrade-4-22-kueue-1-3-to-1-4/2072073388884496384

```release-note
AFS: Fixed ConsumedResources CPU truncating to zero when the sampling interval guard was bypassed by informer cache lag during initialization.
```

## Summary

* **Bug Fixes**
  * Improved Admission Fair Sharing to avoid unnecessary reconciles when sampling cache data is recent but fair-sharing status isn’t available yet.
  * Corrected CPU consumed accounting so it remains numerically positive and doesn’t truncate to zero after self-triggered or repeated updates.

* **Tests**
  * Added an integration test ensuring CPU consumed stays positive across sampling intervals.
  * Added a unit test for the “recent cache with missing fair-sharing status” scenario, including expected requeue timing.

### DIFF
--- a/pkg/controller/core/localqueue_controller.go
+++ b/pkg/controller/core/localqueue_controller.go
@@ -44,6 +44,7 @@ import (
 	config "sigs.k8s.io/kueue/apis/config/v1beta2"
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
 	qcache "sigs.k8s.io/kueue/pkg/cache/queue"
+	queueafs "sigs.k8s.io/kueue/pkg/cache/queue/afs"
 	schdcache "sigs.k8s.io/kueue/pkg/cache/scheduler"
 	"sigs.k8s.io/kueue/pkg/constants"
 	"sigs.k8s.io/kueue/pkg/controller/core/indexer"
@@ -211,15 +212,14 @@ func (r *LocalQueueReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 	}
 
 	if afs.Enabled(r.admissionFSConfig) {
-		initNeeded := r.initializeAfsIfNeeded(&queueObj)
 		lqKey := utilqueue.Key(&queueObj)
-		entry, found := r.queues.AfsConsumedResources.Get(lqKey)
-		if !found {
-			log.V(3).Info("AFS cache entry deleted concurrently, skipping reconciliation")
-			return ctrl.Result{}, nil
-		}
+		hadCache, entry := r.initializeAfsIfNeeded(&queueObj)
 		sinceLastUpdate := r.clock.Now().Sub(entry.LastUpdate)
-		if interval := r.admissionFSConfig.UsageSamplingInterval.Duration; !initNeeded && sinceLastUpdate < interval && !r.queues.AfsEntryPenalties.HasPendingFor(lqKey) {
+		// Enforce the sampling interval when the cache already existed
+		// before this reconcile. Without this, self-triggered status
+		// updates cause sub-millisecond reconciles where the decay math
+		// truncates CPU consumed resources to zero.
+		if interval := r.admissionFSConfig.UsageSamplingInterval.Duration; hadCache && sinceLastUpdate < interval && !r.queues.AfsEntryPenalties.HasPendingFor(lqKey) {
 			return ctrl.Result{RequeueAfter: interval - sinceLastUpdate}, nil
 		}
 		if err := r.reconcileConsumedUsage(ctx, &queueObj); err != nil {
@@ -335,17 +335,14 @@ func (r *LocalQueueReconciler) Update(e event.TypedUpdateEvent[*kueue.LocalQueue
 	return true
 }
 
-func (r *LocalQueueReconciler) initializeAfsIfNeeded(lq *kueue.LocalQueue) bool {
+func (r *LocalQueueReconciler) initializeAfsIfNeeded(lq *kueue.LocalQueue) (hadCache bool, entry queueafs.ConsumedResourcesEntry) {
 	if lq.Status.FairSharing == nil {
 		lq.Status.FairSharing = &kueue.LocalQueueFairSharingStatus{}
 	}
 
 	lqKey := utilqueue.Key(lq)
 	hasStatus := lq.Status.FairSharing.AdmissionFairSharingStatus != nil
-	_, hasCache := r.queues.AfsConsumedResources.Get(lqKey)
-	if hasStatus && hasCache {
-		return false
-	}
+	entry, hadCache = r.queues.AfsConsumedResources.Get(lqKey)
 
 	now := r.clock.Now()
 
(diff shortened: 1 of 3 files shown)


pkg/controller/core/localqueue_controller.go
```diff
@@ -332,7 +332,9 @@ func (r *LocalQueueReconciler) Update(e event.TypedUpdateEvent[*kueue.LocalQueue
 		r.resyncLocalQueueGaugeMetrics(e.ObjectNew)
 	}
 
-	return true
+	// Skip reconcile when spec is unchanged to avoid AFS self-requeue
+	// loop where sub-millisecond decay intervals truncate CPU to zero.
+	return e.ObjectOld.Generation != e.ObjectNew.Generation
```
mimowo: could you add a unit test for the Handler? i think we already have some which assert using wantReconcile

mimowo: while this change is good it seems to main the main source of the problem is that calling the Reconcile too early messes up the AFS state. As the set of other feature grows we will not be able to avoid reconciles in the future. 

Maybe one idea woukd be to also keep in memory in the cache for AFS the information about the last refresh and only take action when we exceed the decay interval.

However, Im ok to postpone this for now, just raising that we shoukd not rely on the face that noting calls Reconcile earlier other than AFS.

sohankunkerkar: Yup, that makes sense! I moved the fix from Update predicate into Reconcile (`hadCache` instead of `!initNeeded`), so the sampling interval is enforced regardless of what triggers the reconcile

mimowo: I see, but could you also submit a separate PR to skip calling Reconcile on no-generation change updates?

test/integration/singlecluster/scheduler/fairsharing/fair_sharing_test.go
```diff
@@ -1425,6 +1425,48 @@ var _ = ginkgo.Describe("Scheduler", ginkgo.Label("feature:fairsharing", "featur
...
+				g.Expect(lqObj.Status.FairSharing).NotTo(gomega.BeNil())
+				g.Expect(lqObj.Status.FairSharing.AdmissionFairSharingStatus).NotTo(gomega.BeNil())
+				consumed := lqObj.Status.FairSharing.AdmissionFairSharingStatus.ConsumedResources
+				g.Expect(consumed).To(gomega.HaveKey(corev1.ResourceCPU))
+				cpu := consumed[corev1.ResourceCPU]
+				g.Expect(cpu.Cmp(resource.MustParse("0"))).To(gomega.BeNumerically(">", 0),
+					"ConsumedResources CPU should be > 0, got %v", cpu.String())
+			}, util.Timeout, util.Interval).Should(gomega.Succeed())
+
+			gomega.Consistently(func(g gomega.Gomega) {
```
mimowo, line 1465: Add ginkgo.By steps to explain what is done and being checked. These steps have two roles:
- document the expectations for other readers of the test code
- leave trace in the logs so that in case of a failure it is easier to correlate the failure with the step which failed

Same for the eventually above

mimowo (commented):
LGTM, left minor test comments

pkg/controller/core/localqueue_controller.go
sohankunkerkar: The hadCache guard already enforces the interval for all reconcile triggers, but happy to submit the Generation check as defense-in-depth in a follow-up if you still want it.

mimowo: Nah, I change my mind, let's re-consider that when we are proven to need it.

test/integration/singlecluster/scheduler/fairsharing/fair_sharing_test.go
sohankunkerkar: Addressed

pkg/controller/core/localqueue_controller.go
```diff
@@ -211,15 +211,20 @@ func (r *LocalQueueReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 	if afs.Enabled(r.admissionFSConfig) {
-		initNeeded := r.initializeAfsIfNeeded(&queueObj)
 		lqKey := utilqueue.Key(&queueObj)
+		_, hadCache := r.queues.AfsConsumedResources.Get(lqKey)
+		r.initializeAfsIfNeeded(&queueObj)
 		entry, found := r.queues.AfsConsumedResources.Get(lqKey)
 		if !found {
 			log.V(3).Info("AFS cache entry deleted concurrently, skipping reconciliation")
 			return ctrl.Result{}, nil
 		}
```
mimowo: I think that code could simplify if we extended the r.initializeAfsIfNeeded to return:
```golang
hadCache, hadStatus, currentUsage := r.initializeAfsIfNeeded(lqKey)
```
wdyt? Then in paces where we check initNeeded - which is confusing, we could compute as 

`wasInitNeeded ;= !hadCache || !hadStatus`

sohankunkerkar: Yeah, that's a good idea. This looks like a cleanup fix that can be handled as a follow-up to keep this PR focused on the fix!?

mimowo: i would combine the two because I think the combined PR will still be of reasonable complexity, and future CPs will be easiee

sohankunkerkar: Kept it to just `hadCache` since `hadStatus` and `currentUsage` aren't used by the caller, so returning them would add two `_` ignores at the call site without simplifying the flow. Does that work for you?

mimowo: That is ok-ish, but I was thinking about also returning currentUsage so that we can drop this code entirely:

```golang
		entry, found := r.queues.AfsConsumedResources.Get(lqKey)
		if !found {
			log.V(3).Info("AFS cache entry deleted concurrently, skipping reconciliation")
			return ctrl.Result{}, nil
		}
```

pkg/controller/core/localqueue_controller.go
```diff
@@ -355,12 +352,13 @@ func (r *LocalQueueReconciler) initializeAfsIfNeeded(lq *kueue.LocalQueue) bool
...
 	}
 
-	if !hasCache {
+	if !hadCache {
 		currentUsage := r.getCurrentUsageForLocalQueue(lq.Spec.ClusterQueue, lqKey)
 		r.queues.AfsConsumedResources.Set(lqKey, currentUsage, now)
+		entry, _ = r.queues.AfsConsumedResources.Get(lqKey)
```
mimowo: nit, but instead of calling get we could just return currentUsage, that should be the same value, right?

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for bugfixing AFS 👍

mimowo:
/release-note-edit
```release-note
AdmissionFairSharing: Fixed ConsumedResources CPU truncating to zero when the sampling interval guard was bypassed by informer cache lag during initialization.
```

tenzen-y:
/release-note-edit

```release-note
AFS: Fixed ConsumedResources CPU truncating to zero when the sampling interval guard was bypassed by informer cache lag during initialization.
```

Issue #12580, opened 2026-06-26
Improve TAS performance


j-skiba:
**What would you like to be added**:

Here's an example of pprof CPU flame chart on 2500 nodes cluster with 15 resource flavors with TAS enabled (GC function calls excluded).

<img width="2555" height="907" alt="Image" src="https://github.com/user-attachments/assets/bcb31f16-b678-43f2-89c2-4c1bc08eca1d" />

There are some substeps of the snapshot building that could be improved:
- Node Copying (`addNodes`):
  The node information we store rather doesn't change very frequently:
  ```go
  type nodeInfo struct {
  	Name        string
  	Labels      map[string]string
  	Taints      []corev1.Taint
  	Allocatable corev1.ResourceList
  }
  ```
  So there is room for improvement to not copy them on every snapshot building. We can introduce a `generation` counter on the nodes cache and reuse the pre-built snapshot tree in-place if the generation doesn't change, completely bypassing the tree-building phase.
  
- Tree building (`initialize`):
  We don't have to rebuild the parent-child pointers (`dom.parent` and `dom.children`) on every cycle if the nodes list hasn't changed.
  
- Eliminate `usagePerNode` map cloning:
  Currently, `usagePerNode()` copies `map[string]resources.Requests{}` (a map of maps) to defend against concurrent modifications during iteration. We can replace this with read and subtract in a lock without cloning the maps.

**Why is this needed**:

At large scales (e.g., 2,500+ nodes and multiple resource flavors), rebuilding the TAS topology snapshot from scratch on every scheduling cycle creates significant CPU overhead and heavy garbage collection pressure. 

**Completion requirements**:

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

mimowo:
@akshay-pm as (3.) is merged, are you actively exploring (1.) or (2.)?

akshay-pm:
@mimowo Yes, I’m actively exploring (1.) and (2.) together, since they seem closely related and can share the same generation-based caching mechanism

mimowo:
Awesome, take your time as it does not seem urgent, and I can totally understand that TAS is complex. Just asked to make sure it is not stale.

akshay-pm:
@j-skiba, @mimowo, I have tried and benchmarked two approaches. Their performance is very similar, and I would appreciate feedback on which direction fits the project better.

[...]

PR #12672, opened 2026-07-01
Avoid cloning per-node usage maps when building TAS snapshots

akshay-pm:
#### What type of PR is this?
/kind cleanup
/area tas

Replaces `nonTasUsageCache.usagePerNode()`, which allocated a new map and deep-cloned every per-node `Requests` on each TAS snapshot, with `forEachNodeUsage(fn)`, which reads and subtracts per-node usage in place under the read lock, avoiding the copy. See #12580 for the background on the CPU and GC overhead this addresses at scale.

Benchmark results
Without change:
image

With change:
Screenshot 2026-07-01 at 10 39 46 AM

pprof flame graph results
Without change:
image

With changes:
image

Part of #12580

```release-note
TAS: Reduced the CPU and memory overhead of building the topology snapshot on large clusters by no longer cloning per-node usage maps on every scheduling cycle.
```

## Summary by CodeRabbit

* **Tests**
  * Added a new benchmark to measure snapshot performance across different topology sizes and workload mixes.
  * Updated cache tests to validate node usage through a new iteration path, keeping coverage aligned with current behavior.

* **Performance**
  * Improved how cache data is traversed during snapshot generation, helping support more efficient processing without changing results.

### DIFF
--- a/pkg/cache/scheduler/tas_flavor.go
+++ b/pkg/cache/scheduler/tas_flavor.go
@@ -146,11 +146,11 @@ func (c *TASFlavorCache) snapshot(
 	for domainID, usage := range tasDomainUsages {
 		snapshot.addTASUsage(domainID, usage)
 	}
-	for nodeName, usage := range c.nonTasUsageCache.usagePerNode() {
+	c.nonTasUsageCache.forEachNodeUsage(func(nodeName string, usage resources.Requests) {
 		if domainID, ok := nodeToDomain[nodeName]; ok {
 			snapshot.addNonTASUsage(domainID, usage)
 		}
-	}
+	})
 	return snapshot
 }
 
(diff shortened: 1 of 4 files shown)


mimowo:
cc @j-skiba 
Im wondering if we could repeat the pprof experiment to confirm the improvement.

I think since this is changing production code we need a release note.

akshay-pm:
> cc @j-skiba Im wondering if we could repeat the pprof experiment to confirm the improvement.
> 
> I think since this is changing production code we need a release note.

Sure, I have added a release note. Will also come back with the pprof result comparisions as well.

mimowo:
Thank you 👍 Leaving lgtm to @j-skiba 

/approve
/cherrypick release-0.18
/cherrypick release-0.17

akshay-pm:
@j-skiba, @mimowo Added before/after pprof flame graphs to the description

j-skiba:
/lgtm

Thank you 👍

PR #12676, opened 2026-07-01
[Skills] Add code style skill for structured errors

mszadkow:
#### What type of PR is this?
/kind documentation

Adds a `structured-errors` code-review skill under `cmd/experimental/skills/reviewer/code-style/`
that flags error checks (especially in tests) done via string comparison on `err.Error()`,
and recommends sentinel or typed errors checked with `errors.Is`/`errors.As` instead —
so future reviews suggest this pattern instead of relying on ad hoc feedback.

```release-note
NONE
```

## Summary by CodeRabbit

* **Documentation**
  * Expanded review guidance for identifying improper error handling patterns.
  * Added a new rule covering structured errors, encouraging typed or sentinel error checks instead of comparing error message text.
  * Updated the skill reference list so the new guidance is easier to find.

### DIFF
--- a/cmd/experimental/skills/reviewer/code-style/structured-errors/SKILL.md
+++ b/cmd/experimental/skills/reviewer/code-style/structured-errors/SKILL.md
@@ -1,0 +1,29 @@
+---
+name: structured-errors
+description: Flag error checking done via string comparison on err.Error() instead of structured/typed errors checked with errors.Is/errors.As.
+license: Apache-2.0
+metadata:
+  copyright: The Kubernetes Authors
+---
+
+# Skill: Structured Errors
+
+**Structured errors over string comparison** — error checking (especially in tests)
+done via substring/equality checks on `err.Error()` instead of a sentinel or typed
+error checked with `errors.Is` / `errors.As`. String-matching turns tests into
+"change detector" tests, where a purely cosmetic wording tweak forces updates across
+every test that asserted on it.
+
+- A test asserting `err.Error() == "..."` or `strings.Contains(err.Error(), "...")`
+  against a bare `errors.New(...)` / `fmt.Errorf(...)`. Prefer a sentinel
+  (`var ErrFoo = errors.New(...)`) or a typed error with an `Is` method, checked with
+  `errors.Is`.
+- An error whose data (offending names, counts, a resource reference) only exists in
+  the formatted message (`fmt.Errorf("...: %s", x)`) when a caller already needs that
+  data. Prefer a typed error struct with the data as a field.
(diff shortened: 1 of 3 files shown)


cmd/experimental/skills/reviewer/code-style/structured-errors/SKILL.md
```diff
@@ -0,0 +1,29 @@
...
+- A test asserting `err.Error() == "..."` or `strings.Contains(err.Error(), "...")`
+  against a bare `errors.New(...)` / `fmt.Errorf(...)`. Prefer a sentinel
+  (`var ErrFoo = errors.New(...)`) or a typed error with an `Is` method, checked with
+  `errors.Is`.
+- An error whose data (offending names, counts, a resource reference) only exists in
+  the formatted message (`fmt.Errorf("...: %s", x)`) when a caller already needs that
+  data. Prefer a typed error struct with the data as a field.
+- A `cmp.Comparer` or similar helper that falls back to `strings.HasPrefix` to work
+  around errors not being directly comparable — make the errors comparable instead
+  (sentinel values or an `Is` method), and prefer `cmpopts.EquateErrors()` in tests.
```
mimowo, line 26: I think these descriptions aren't very clear without example snippets for "how-to-do-well" .

Alternatively, maybe instead of the SKILL we could introduce our Error comparator in Kueue and reuse it across all tests giving the good example how to do it? 

From my experience most people and AI agents just copy-paste code in that situations, so the more good examples we have the more of a chance the next example is as we want.

From my experience with errors in golang and k8s in general some kind of better comparator is really needed as an option to cmp.Diff, and currently people tend to workaround that problem

mimowo: In other words to say - maybe instead of the skill what we need is an issue and a PR which refactors known problematic cases with some commonized fashion. 

Once we have this "gold standard" example, and we still see people not following it, then I would create the skill. For now, I think the main problem is the lack of a good example how to do it which could be transferred easily test-to-test.

mimowo:
cc @kannon92 @amy I raised the concern that what we need is rather some good examples how to do it first, and potentially a function to commonize the code: https://github.com/kubernetes-sigs/kueue/pull/12676#discussion_r3504109170

However, I'm also ok to merge if you think the skill is valuable too, I just think it would be valuable more with example snippets.

kannon92:
I was thinking we can have a real contributor guide: https://github.com/kubernetes-sigs/kueue/pull/11514

Some of these read like style guide. I don't mind a skill here but I'm not really sure how well the description would work.

amy:
Dumb question. Is this easy to make a lint rule?

PR #12679, opened 2026-07-01
[Cleanup] Use structured errors in TAS Flavorassigner

mszadkow:
#### What type of PR is this?
/kind cleanup

Refactoring that improves error handling patterns as suggested in [the comment](https://github.com/kubernetes-sigs/kueue/pull/12562/changes#r3498546475)

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized TAS/flavor assignment failures to use consistent typed error values instead of ad-hoc string messages, improving identification and matching.
  * Added semantic error matching for cases like multiple TAS flavors being assigned, with order-insensitive flavor set comparison.
  * Improved clarity for missing TAS cache data, missing TAS flavor assignments, and unsupported elastic required/preferred topology.

* **Tests**
  * Updated assertions to compare errors using typed/sentinel errors rather than string comparisons, including the duplicated-flavor scenario.

### DIFF
--- a/pkg/scheduler/flavorassigner/errors.go
+++ b/pkg/scheduler/flavorassigner/errors.go
@@ -1,0 +1,57 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package flavorassigner
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+)
+
+// Sentinel errors for TAS request building, so callers/tests can identify the
+// failure with errors.Is instead of matching on the message text.
+var (
+	ErrElasticRequiredTopologyNotSupported  = errors.New("required topology is not supported with ElasticJobsViaWorkloadSlices")
+	ErrElasticPreferredTopologyNotSupported = errors.New("preferred topology is not supported with ElasticJobsViaWorkloadSlices")
+	ErrNoTASCacheInformation                = errors.New("workload requires Topology, but there is no TAS cache information")
+	ErrNoTASFlavorAssigned                  = errors.New("no TAS flavor assigned")
+)
+
+// MultipleTASFlavorsAssignedError indicates that a PodSet ended up with more
+// than one TAS flavor assigned across its resources, which TAS cannot handle.
+// The offending flavors are kept as a field rather than only in the message so
+// callers can inspect them without parsing the error text.
+type MultipleTASFlavorsAssignedError struct {
+	Flavors []kueue.ResourceFlavorReference
+}
+
+func (e *MultipleTASFlavorsAssignedError) Error() string {
+	names := make([]string, len(e.Flavors))
+	for i, f := range e.Flavors {
+		names[i] = string(f)
+	}
+	return fmt.Sprintf("more than one TAS flavor assigned: %s", strings.Join(names, ", "))
+}
+
+// Is reports whether target is any MultipleTASFlavorsAssignedError (regardless of Flavors).
+// Callers must compare Flavors separately if exact error matching is required.
+func (e *MultipleTASFlavorsAssignedError) Is(target error) bool {
+	_, ok := target.(*MultipleTASFlavorsAssignedError)
+	return ok
+}
--- a/pkg/scheduler/flavorassigner/errors.go
+++ b/pkg/scheduler/flavorassigner/errors.go
@@ -19,6 +19,7 @@ package flavorassigner
 import (
 	"errors"
 	"fmt"
+	"slices"
 	"strings"
 
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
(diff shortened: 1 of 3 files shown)


mimowo:
/test pull-kueue-test-unit-shard-0-main
/test pull-kueue-test-unit-shard-1-main
fixed now

pkg/scheduler/flavorassigner/errors.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	return fmt.Sprintf("more than one TAS flavor assigned: %s", strings.Join(names, ", "))
+}
+
+// Is reports whether target is any MultipleTASFlavorsAssignedError (regardless of Flavors).
+// Callers must compare Flavors separately if exact error matching is required.
+func (e *MultipleTASFlavorsAssignedError) Is(target error) bool {
+	_, ok := target.(*MultipleTASFlavorsAssignedError)
+	return ok
+}
```
vladikkuzn, line 70: Why not use errors.As/AsType instead?

mszadkow: Right, this might be handled by errors.As too

pkg/scheduler/flavorassigner/errors.go
```diff
@@ -0,0 +1,57 @@
...
+
+func (e *MultipleTASFlavorsAssignedError) Error() string {
+	names := make([]string, len(e.Flavors))
+	for i, f := range e.Flavors {
+		names[i] = string(f)
+	}
+	return fmt.Sprintf("more than one TAS flavor assigned: %s", strings.Join(names, ", "))
+}
+
+// Is reports whether target is any MultipleTASFlavorsAssignedError (regardless of Flavors).
```
dkaluza: Why regardless of Flavors?
This is quite unexpected, default behavior is equality comparison AFAIK.

IMO we should definitely check the flavors matching here, otherwise we might miss incorrect behavior in tests.

pkg/scheduler/flavorassigner/flavorassigner_test.go
```diff
@@ -4352,12 +4352,16 @@ func TestWorkloadsTopologyRequests_ErrorBranches(t *testing.T) {
...
 			}
-			if tc.wantErr != "" && errMsg != tc.wantErr {
-				t.Errorf("Error mismatch (-want +got):\n%s", cmp.Diff(tc.wantErr, errMsg))
+			// For MultipleTASFlavorsAssignedError, also verify Flavors match
+			if wantMFE, ok := tc.wantErr.(*MultipleTASFlavorsAssignedError); ok {
+				if actualMFE, ok := tc.assignment.PodSets[0].Status.err.(*MultipleTASFlavorsAssignedError); ok {
+					if diff := cmp.Diff(wantMFE.Flavors, actualMFE.Flavors); diff != "" {
+						t.Errorf("Flavors mismatch (-want +got):\n%s", diff)
+					}
+				}
```
dkaluza: Why don't we just incloude this in Error Is?

pkg/scheduler/flavorassigner/flavorassigner_test.go
```diff
@@ -4590,14 +4594,20 @@ func TestWorkloadsTopologyRequests_ElasticJobsValidation(t *testing.T) {
...
 				}
-				if tc.assignment.PodSets[0].Status.err == nil {
-					t.Errorf("expected error %q, got nil", tc.wantErr)
-				} else if diff := cmp.Diff(tc.wantErr, tc.assignment.PodSets[0].Status.err.Error()); diff != "" {
-					t.Errorf("Error mismatch (-want +got):\n%s", diff)
+				if !errors.Is(tc.assignment.PodSets[0].Status.err, tc.wantErr) {
+					t.Errorf("got error %v, want error %v", tc.assignment.PodSets[0].Status.err, tc.wantErr)
+				}
+				// For MultipleTASFlavorsAssignedError, also verify Flavors match
+				if wantMFE, ok := tc.wantErr.(*MultipleTASFlavorsAssignedError); ok {
```
dkaluza: Consider just including flavors in error is

mszadkow: I went for errors.As

mszadkow:
@vladikkuzn @dkaluza ptal

pkg/scheduler/flavorassigner/flavorassigner_test.go
```diff
@@ -46,18 +47,30 @@ import (
...
 
+// errEqual reports whether a and b represent the same error.
+// MultipleTASFlavorsAssignedError is unwrapped with errors.As and compared by its Flavors field;
+// every other error is compared with errors.Is, which covers sentinel errors declared with errors.New.
+func errEqual(a, b error) bool {
+	if a == nil || b == nil {
+		return a == nil && b == nil
+	}
+	var am, bm *MultipleTASFlavorsAssignedError
+	aIsMulti, bIsMulti := errors.As(a, &am), errors.As(b, &bm)
```
dkaluza: nit: from https://pkg.go.dev/errors#As
> For most uses, prefer [AsType](https://pkg.go.dev/errors#AsType). As is equivalent to [AsType](https://pkg.go.dev/errors#AsType) but sets its target argument rather than returning the matching error and doesn't require its target argument to implement error.

I would generally still go with `Is` implementation based on the flavors inside, in this way you do not have to implement custom "errEqual" - AFAIK this is exactly the purpose of the `Is` method and why it is supposed to be overriden in structs.  Also this would better extend to future usage. E.g. if sb needs to check that  `MultipleTASFlavorsAssignedError` are equal in other file... `Is` overriding gives us a consistent solution across different types of errors.

dkaluza (approved):
IMO proper implementation of `Is` would be more elegant & easier to maintain, but generally looks fine.

dkaluza (approved):

pkg/scheduler/flavorassigner/flavorassigner_test.go
mszadkow: done

mszadkow:
@ivnovakov ptal

dkaluza:
/lgtm
@mimowo this cleanup looks ready to me, can you take a look when you have a moment?

mimowo:
/approve
Thank you folks

mimowo:
/test pull-kueue-populator-test-e2e-main

looks like infra failure , the test just interrupted :/ This never happened in the past, I suppose some issue with the new machines or the autoscaling of the pool (we have new faster machines after 1st July, but maybe the autoscaler configuration needs improving
@coderabbitai please open the flake issue, ofc remember to add the flake label

PR #12693, opened 2026-07-01
feat: try to measure timings for individual bash commands

yuluo-yx:
#### What type of PR is this?

/area testing

- For https://github.com/kubernetes-sigs/kueue/pull/12661#discussion_r3500177270

```release-note
None
```

## Summary by CodeRabbit

* **New Features**
  * Added optional command-level timing output for timed shell runs, controlled by a new setting (`MAKE_TIMING_COMMANDS`, defaulting to off).
  * When enabled, timing output logs command transitions (including exit-triggered entries) rather than only overall runtime.
  * Timing entries respect the existing minimum-duration threshold, with safe fallback behavior when the value is misconfigured.
  * If disabled, existing timing behavior remains unchanged.


mimowo (commented):
cc @mbobrovskyi ptal

hack/make-timed-shell.sh
```diff
@@ -40,6 +40,100 @@ command_preview() {
   fi
 }
 
+write_command_timing_env() {
```
amy, line 43: please provide sample output in the PR description.

mimowo, line 43: @yuluo-yx this is not mutually exclusive with this proposal, but I tried with LLM and it suggests using "ts" style for timestamps should be possible and provides that snippet: 

```sh
timestamp_stream() {
  local line
  local ts
  local supports_bash_time=0

  if printf -v ts '%(%Y-%m-%dT%H:%M:%SZ)T' -1 2>/dev/null; then
    supports_bash_time=1
  fi

  while IFS= read -r line || [[ -n "${line}" ]]; do
    if [[ "${supports_bash_time}" -eq 1 ]]; then
      printf -v ts '%(%Y-%m-%dT%H:%M:%SZ)T' -1
    else
      ts=$(date -u '+%Y-%m-%dT%H:%M:%SZ')
    fi
    printf '[%s] %s\n' "${ts}" "${line}"
  done
}
```

then when running bash command as a wrapper in `hack/make-timed-shell.sh`
```
if is_truthy "${MAKE_TIMING_LOG_TIMESTAMPS:-0}"; then
  exec > >(timestamp_stream)
  exec 2> >(timestamp_stream >&2)
fi
```
that seems relatively simple, could you evaluate if that make sense and is feasible - I haven't tried, but if that works then it might be pretty useful - I would many times in the past debug locally using "ts" command, so I know its power.

mimowo: You can find examples in the build-log files. This is an example one for e2e tests:
https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12693/pull-kueue-test-e2e-baseline-main-1-36/2072327129072144384/build-log.txt

You can find the timings by searching "make-command-timing", example log:
```sh
make-command-timing: started=2026-07-01T14:32:05Z finished=2026-07-01T14:34:23Z status=0 duration=138s command=docker buildx build -t us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue:v0.19.0-devel-291-gefd42432c -t us-central1-docker.pkg.dev/k8s-staging-images/kueue/kueue:main --platform=linux/amd64 --build-arg BASE_IMAGE=gcr.io/distrole...
```

mimowo: That would probably anyway be another PR

yuluo-yx: Got it. I think this can be done in a single PR — I'll submit it in a bit.

yuluo-yx:
ptal @mimowo

mimowo:
@yuluo-yx so stamping all lines with a timestamp is for later?

mimowo:
/lgtm
/approve
Let's give it a try, thank you 👍 
/cherrypick release-0.18
/cherrypick release-0.17

PR #12697, opened 2026-07-01
Fix race on AFS penalties update

### LINKED ISSUE #12546: Read-Modify-Write Data Race in Fair Sharing Caches Allows Quota/Penalty Evasion in pkg/cache/queue/afs/entry_penalties.go
### Impact
A malicious tenant can exploit this data race to evade AdmissionFairSharing penalties. By spamming the creation/updating of Workloads to force concurrent RMW operations, they can cause the Kueue scheduler to 'lose' their penalties, keeping their LocalQueue artificially prioritized and monopolizing cluster resources (Denial of Service to other tenants).

### Data Flow
- **pkg/scheduler/scheduler.go:812**: Scheduler goroutine calls s.updateEntryPenalty(..., add) when assuming a Workload, which calls AfsEntryPenalties.Push().
- **pkg/controller/core/workload_controller.go:1175**: Informer listener goroutine concurrently calls WorkloadReconciler.Update() -> updateAfsConsumedUsage() -> AfsEntryPenalties.Sub() for another Workload in the same LocalQueue.
- **pkg/cache/queue/afs/entry_penalties.go:36**: Both goroutines perform m.Peek(lqKey) simultaneously, reading the same stale penalty value.
- **pkg/cache/queue/afs/entry_penalties.go:36**: Both goroutines compute MergeResourceListKeepSum based on the stale value.
- **pkg/cache/queue/afs/entry_penalties.go:36**: Both goroutines call m.penalties.Add() sequentially. The second Add overwrites the first, silently discarding the other goroutine's penalty addition or subtraction (SINK).

### Reproduction
**Preconditions**:
AdmissionFairSharing must be enabled. A tenant must have permissions to create Workloads in a LocalQueue.

**Steps**:
1. Tenant creates a Workload (W1) that is immediately admissible (no admission checks).
2. At the exact same time, the tenant creates a second Workload (W2) in the same LocalQueue.
3. The Scheduler goroutine assumes W2 and calls AfsEntryPenalties.Push().
4. Simultaneously, the Informer listener goroutine processes W1's transition to Admitted and calls AfsEntryPenalties.Sub().
5. Due to the data race, the Push from the Scheduler is overwritten by the Sub from the Informer.
6. W2 is successfully scheduled but its fair-sharing penalty is permanently lost, allowing the tenant to bypass fair-sharing limits.

### Vulnerable Code
```
m.penalties.Add(lqKey, resource.MergeResourceListKeepSum(m.Peek(lqKey), penalty))
```

### Remediation
Introduce a mutex to protect Read-Modify-Write operations on the Fair Sharing caches, or use a compute/update function that locks the specific entry.

This logical root cause was found at **4** additional call site(s):

- `pkg/controller/core/workload_controller.go` (Finding 13584: Race conditions in Admission Fair Sharing (AFS) caches lead to permanent tenant starvation or quota bypass)
- `pkg/scheduler/scheduler.go` (Finding 13772: TOCTOU Race Condition in Admission Fair Sharing Caches Causes Unfair LocalQueue Penalization)
- `pkg/util/queue/local_queue.go` (Finding 13815: Data Race and Lost Updates in Admission Fair Sharing Penalty Map)
- `pkg/util/testingjobs/mpijob/wrappers_mpijob.go` (Finding 13840: TOCTOU Race Condition in Admission Fair Sharing Penalty Tracking)

### Environment
* Install tools: N/A
* Others: N/A

MaysaMacedo:
#### What type of PR is this?
/kind bug

Fixes #12546

AI Assistance Disclosure: This pull request was developed with the assistance of AI (Claude Code).

```release-note
AFS: Fixed a race in Admission Fair Sharing penalty updates where concurrent workload operations could lose penalty changes, causing LocalQueues to receive incorrect priority.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Made AFS penalty adjustments fully atomic by merging deltas into stored totals and using conditional merge/delete behavior.
  * Improved penalty subtraction to avoid in-place mutation and reliably clear entries when totals drop to zero.

* **New Features**
  * Added a `UpdateOrDelete` operation to the synchronized map utility to support atomic conditional overwrite/removal.

* **Tests**
  * Added concurrency-focused tests covering `Update` and `UpdateOrDelete` correctness, including no lost updates under contention.

### DIFF
--- a/pkg/cache/queue/afs/entry_penalties.go
+++ b/pkg/cache/queue/afs/entry_penalties.go
@@ -33,18 +33,22 @@ func NewPenaltyMap() *AfsEntryPenalties {
 }
 
 func (m *AfsEntryPenalties) Push(lqKey utilqueue.LocalQueueReference, penalty corev1.ResourceList) {
-	m.penalties.Add(lqKey, resource.MergeResourceListKeepSum(m.Peek(lqKey), penalty))
+	m.penalties.Update(lqKey, func(existing corev1.ResourceList, _ bool) corev1.ResourceList {
+		return resource.MergeResourceListKeepSum(existing, penalty)
+	})
 }
 
 func (m *AfsEntryPenalties) Sub(lqKey utilqueue.LocalQueueReference, penalty corev1.ResourceList) {
+	negated := make(corev1.ResourceList, len(penalty))
 	for k, v := range penalty {
-		v.Neg()
-		penalty[k] = v
+		q := v.DeepCopy()
+		q.Neg()
+		negated[k] = q
 	}
-	m.penalties.Add(lqKey, resource.MergeResourceListKeepSum(m.Peek(lqKey), penalty))
-	if canClearPenalty(m.Peek(lqKey)) {
-		m.penalties.Delete(lqKey)
-	}
+	m.penalties.UpdateOrDelete(lqKey, func(existing corev1.ResourceList) (corev1.ResourceList, bool) {
+		merged := resource.MergeResourceListKeepSum(existing, negated)
+		return merged, canClearPenalty(merged)
+	})
 }
 
 func canClearPenalty(penalty corev1.ResourceList) bool {
--- a/pkg/util/maps/maps.go
+++ b/pkg/util/maps/maps.go
@@ -100,6 +100,17 @@ func (dwc *SyncMap[K, V]) Add(k K, v V) {
 	dwc.m[k] = v
 }
 
+func (dwc *SyncMap[K, V]) UpdateOrDelete(k K, f func(existing V) (V, bool)) {
+	dwc.lock.Lock()
+	defer dwc.lock.Unlock()
+	newVal, canClear := f(dwc.m[k])
+	if canClear {
+		delete(dwc.m, k)
+	} else {
+		dwc.m[k] = newVal
+	}
+}
+
 func (dwc *SyncMap[K, V]) Get(k K) (V, bool) {
 	dwc.lock.RLock()
 	defer dwc.lock.RUnlock()
--- a/pkg/util/maps/maps_test.go
+++ b/pkg/util/maps/maps_test.go
@@ -18,6 +18,7 @@ package maps
 
 import (
 	"maps"
+	"sync"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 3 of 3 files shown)


amy:
@MaysaMacedo noticed you put a hold on this PR. If its not ready for review, can you convert it to a draft or title it with WIP?

mimowo:
cc @apullo777 would you like to review as you are alerady quite familiar with AFS?

mimowo:
cc @kshalot

pkg/cache/queue/afs/entry_penalties.go
kshalot: Isn't this the exact same vulnerability here?

kshalot: Although this case might be more complex, because I think there is a potential TOCTOU if we just replace `Add` with `GetAndUpdate`.

apullo777: +1 The referenced issue is specifically about a scheduler `Push` racing a workload-reconciler `Sub`.

And I think the clear path needs to be atomic too: if `Sub` becomes atomic but `Peek -> canClearPenalty -> Delete` stays separate, a concurrent `Push` can land between the Peek that feeds the zero-check and the `Delete`, and then be removed with the whole key. So the subtract/merge, zero-check, and optional delete probably need to happen in the same penalties-map critical section, perhaps with an `UpdateOrDelete`-style helper.

pkg/util/maps/maps_test.go
```diff
@@ -292,3 +296,57 @@ func TestFilterKeys(t *testing.T) {
 		})
 	}
 }
+
+func TestGetAndUpdate(t *testing.T) {
```
apullo777: Could we also add a concurrent regression test next to this one?

The failure mode here is a lost update / atomicity issue, not necessarily something `go test -race` will catch, since each individual SyncMap operation is already locked. A useful test shape would run many goroutines updating the same key and then assert the final value is exactly the conserved total.

pkg/util/maps/maps.go
```diff
@@ -100,6 +100,22 @@ func (dwc *SyncMap[K, V]) Add(k K, v V) {
...
+}
+
+func (dwc *SyncMap[K, V]) UpdateOrDelete(k K, f func(existing V) (V, bool)) {
+	dwc.lock.Lock()
+	defer dwc.lock.Unlock()
+	newVal, canClear := f(dwc.m[k])
+	dwc.m[k] = newVal
+	if canClear {
+		delete(dwc.m, k)
+	}
```
kshalot, line 111: I think we could have a tiny optimization here:

```suggestion
	newVal, canClear := f(dwc.m[k])
	if canClear {
		delete(dwc.m, k)
	} else {
		dwc.m[k] = newVal
	}
```

Makes no sense to set and then immediately clear the value.

kshalot:
@MaysaMacedo We could use a release note for this one.

Also - is there something blocking this PR from your POV? Or can we remove `do-not-merge/hold`?

pkg/util/maps/maps.go
```diff
@@ -100,6 +100,23 @@ func (dwc *SyncMap[K, V]) Add(k K, v V) {
 	dwc.m[k] = v
 }
 
+func (dwc *SyncMap[K, V]) GetAndUpdate(k K, f func(existing V) V) {
```
sohankunkerkar: I think you can drop this in favor of https://github.com/kubernetes-sigs/kueue/blame/main/pkg/util/maps/maps.go#L136?

pkg/util/maps/maps_test.go
```diff
@@ -292,3 +298,153 @@ func TestFilterKeys(t *testing.T) {
...
+			initial: map[string]corev1.ResourceList{
+				"default/lq1": {corev1.ResourceCPU: resource.MustParse("3")},
+			},
+			k: "default/lq1",
+			f: func(existing corev1.ResourceList) corev1.ResourceList {
+				return r.MergeResourceListKeepSum(existing, add5CPU)
+			},
+			want: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("8")},
+		},
+		"different key is not affected": {
```
sohankunkerkar, line 318: This test sets up `default/lq2` but only asserts the value at `default/lq1`. Can you add an assertion that "default/lq2" is unchanged?

pkg/cache/queue/afs/entry_penalties.go
```diff
@@ -33,18 +33,20 @@ func NewPenaltyMap() *AfsEntryPenalties {
...
 
 func (m *AfsEntryPenalties) Push(lqKey utilqueue.LocalQueueReference, penalty corev1.ResourceList) {
-	m.penalties.Add(lqKey, resource.MergeResourceListKeepSum(m.Peek(lqKey), penalty))
+	m.penalties.GetAndUpdate(lqKey, func(existing corev1.ResourceList) corev1.ResourceList {
+		return resource.MergeResourceListKeepSum(existing, penalty)
+	})
 }
 
 func (m *AfsEntryPenalties) Sub(lqKey utilqueue.LocalQueueReference, penalty corev1.ResourceList) {
 	for k, v := range penalty {
```
sohankunkerkar, line 43: Sub negates penalty in-place before the callback, which mutates the caller's map. So the logs will print the wrong sign.

MaysaMacedo:
@sohankunkerkar @kshalot This PR is ready for another look, whenever you can. Thanks!

pkg/util/maps/maps_test.go
```diff
@@ -292,3 +298,166 @@ func TestFilterKeys(t *testing.T) {
 		})
 	}
 }
+
+func TestUpdate(t *testing.T) {
```
sohankunkerkar, line 297: I  think the rest of this file uses simple Go types (int, string). These tests pull in 3 Kubernetes/Kueue imports to do what's essentially integer addition. Update/UpdateOrDelete are generic, so their atomicity can be proven with SyncMap[string, int] and v+1/v-1 to match the file convention.

sohankunkerkar (approved):
/lgtm

mimowo:
/release-note-edit
```release-note
AFS: Fixes race condition on Admission Fair Sharing penalties, that could be generated by concurrent workloads triggering many read-modify-write operations and making the LocalQueues to be prioritized and monopolize cluster resources.
```

mimowo:
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for hardening AFS 👍 Please cherrypick if the robot fails

mimowo:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12697/pull-kueue-test-e2e-baseline-main-1-36/2079625394578788352
/test pull-kueue-test-e2e-baseline-main-1-36
That is not related, but I suspect a recent regression

mimowo:
@coderabbitai please open issue for https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12697/pull-kueue-test-e2e-baseline-main-1-36/2079625394578788352

mimowo:
/release-note-edit
```release-note
AFS: Fixed a race in Admission Fair Sharing penalty updates where concurrent workload operations could lose penalty changes, causing LocalQueues to receive incorrect priority and potentially monopolize cluster resources.
```

mimowo:
/release-note-edit
```release-note
AFS: Fixed a race in Admission Fair Sharing penalty updates where concurrent workload operations could lose penalty changes, causing LocalQueues to receive incorrect priority.
```

Issue #12633, opened 2026-06-29
KueueViz: set up rich unit tests for the frontend


olekzabl:
**What would you like to be cleaned**:

KueueViz, for its frontend part, has an e2e test suite ([here](https://github.com/kubernetes-sigs/kueue/blob/bee9304a8aa9a83412b5d91516fdd3779ff1b331/test/e2e/kueueviz/cypress/e2e/app.cy.js)) but a very limited coverage of unit tests (for [ErrorMessage](https://github.com/kubernetes-sigs/kueue/blob/bee9304a8aa9a83412b5d91516fdd3779ff1b331/cmd/kueueviz/frontend/src/ErrorMessage.test.js), [UsageBar](https://github.com/kubernetes-sigs/kueue/blob/bee9304a8aa9a83412b5d91516fdd3779ff1b331/cmd/kueueviz/frontend/src/UsageBar.test.js) and [useWebSocket](https://github.com/kubernetes-sigs/kueue/blob/bee9304a8aa9a83412b5d91516fdd3779ff1b331/cmd/kueueviz/frontend/src/useWebSocket.test.js) - but not for most of the components).

The existing unit tests are JS-level only, with no DOM operations.

For healthiest growth, we should add DOM-level unit tests, and coverage for all components.

**Why is this needed**:

This will allow unit-testing any code changes in the components, including edge cases.

While this can be technically covered within the existing framework of Cypress e2e tests, such strategy is suboptimal: e2e tests are inherently less stable and costly to run (incl. latency); developing a project should not rely so heavily on those.

(A direct trigger to create this issue was reviewing #12613)

ekam-walia:
Hi all, I am happy to handle this issue 
/assign

PR #12698, opened 2026-07-01
kueueviz: re-validate bearer token on long-lived WebSocket connections

### LINKED ISSUE #12563: Improper Access Control via Missing Re-authentication on Long-Lived WebSockets in cmd/kueueviz/backend/handlers/websocket_handler.go
### Impact
If a user's access is revoked or their token expires, they will continue to receive real-time, sensitive cluster data via their pre-existing WebSocket connection indefinitely.

### Data Flow
- **cmd/kueueviz/backend/middleware/auth.go:73**: HTTP request is initially authenticated by auth.Middleware()
- **cmd/kueueviz/backend/handlers/websocket_handler.go:48**: Connection is upgraded to a WebSocket. The HTTP context and middleware checks are finalized.
- **cmd/kueueviz/backend/handlers/websocket_handler.go:179**: An infinite loop streams Kubernetes informer updates to the client without ever re-verifying the user's token or enforcing session expiration.

### Reproduction
**Preconditions**:
1. kueueviz server is running with TokenReview auth enabled.
2. You have a valid Kubernetes Bearer token that is about to expire or can be manually revoked.
3. `websocat` tool is installed.

**Steps**:
1. Export your valid token as VALID_TOKEN.
2. Run the reproduction script to establish a WebSocket connection.
3. Once connected, manually revoke the ServiceAccount or wait for the token to expire.
4. Create a new Workload in the Kubernetes cluster.

### Vulnerable Code
```
case <-updateChan:
	if err := h.sendData(ctx, conn, dataFetcher); err != nil {
```

### Remediation
Implement periodic token re-evaluation for open WebSocket connections or enforce a maximum connection lifetime.

Suggested Fix:
```
Add a ticker inside the `handleInformerUpdates` select loop that periodically triggers a re-authentication check using the original token. If the TokenReview fails, close the WebSocket connection.
```

### Environment
* Install tools: N/A
* Others: N/A

Vaishnav88sk:
#### What type of PR is this?

/kind bug
/area dashboard

The kueueviz WebSocket handler authenticated the bearer token only once on the initial HTTP upgrade. After that, `handleInformerUpdates` streamed cluster data in an infinite loop with no further token checks, a user whose token expired or ServiceAccount was revoked continued receiving real-time sensitive cluster data indefinitely.

Fix:
- Store the raw bearer token in gin context (`"token"` key) inside `auth.Middleware`.
- Add `Authenticator.ValidateToken`, reuses the existing TokenReview + TTL-cache.
- Add a `TokenValidator` interface in the handlers package (nil-safe when auth is disabled).
- Add a 30-second re-auth ticker in `handleInformerUpdates`; closes the connection
  with `ClosePolicyViolation` if the token is invalid.

Fixes #12563 

- No behaviour change when `AuthMode != "TokenReview"`, the ticker is never started.
- `ValidateToken` reuses the TTL cache, so no extra API calls between cache expiry.
- This PR was developed with AI assistance, disclosed per the [Kubernetes AI Tool Usage Policy](https://www.kubernetes.dev/docs/guide/pull-requests/#ai-guidance).

```release-note
KueueViz: Fixed a security issue in kueueviz where WebSocket connections continued streaming cluster data after a bearer token expired or was revoked.
Connections are now closed within 30 seconds of token invalidation.
```

## Summary by CodeRabbit

* **New Features**
  * WebSocket streams now periodically re-check bearer token validity (default every 30s) and close active sessions when the token is revoked/expired.
* **Bug Fixes**
  * WebSocket connections now terminate with a clear “token expired or revoked” policy-violation close reason when revalidation fails.
  * Frontend now displays more specific WebSocket error messages based on the close code.
* **Improvements**
  * Token authentication now returns more accurate HTTP status codes (401 for invalid tokens, 503 when token review fails) and preserves auth details in request context.
* **Tests**
  * Added coverage for token revalidation and close behavior.

### DIFF
--- a/cmd/kueueviz/backend/handlers/websocket_handler.go
+++ b/cmd/kueueviz/backend/handlers/websocket_handler.go
@@ -51,6 +51,10 @@ var upgrader = websocket.Upgrader{
 	},
 }
 
+const (
+	tokenValidationTimeout = 5 * time.Second
+)
+
 // GenericWebSocketHandler creates a WebSocket endpoint with informer-based real-time updates
 // Accepts one or more GroupVersionKinds to watch for changes
 func (h *Handlers) GenericWebSocketHandler(dataFetcher func(ctx context.Context) (any, error), gvks ...schema.GroupVersionKind) gin.HandlerFunc {
@@ -214,7 +218,7 @@ func (h *Handlers) handleInformerUpdates(ctx context.Context, conn *websocket.Co
 		case <-authTickerC:
 			// Use a short timeout so a slow or unavailable API server cannot
 			// block this goroutine indefinitely.
-			validateCtx, validateCancel := context.WithTimeout(context.Background(), 5*time.Second)
+			validateCtx, validateCancel := context.WithTimeout(ctx, tokenValidationTimeout)
 			ok, err := h.validator.ValidateToken(validateCtx, token)
 			validateCancel()
 			if err != nil {
@@ -224,7 +228,7 @@ func (h *Handlers) handleInformerUpdates(ctx context.Context, conn *websocket.Co
 			}
 			if err != nil || !ok {
 				// Set a write deadline so the close-frame write cannot block indefinitely.
-				_ = conn.SetWriteDeadline(time.Now().Add(5 * time.Second))
+				_ = conn.SetWriteDeadline(time.Now().Add(writeDeadlineExtension))
 				_ = conn.WriteMessage(
 					websocket.CloseMessage,
 					websocket.FormatCloseMessage(websocket.ClosePolicyViolation, "token expired or revoked"),
--- a/cmd/kueueviz/backend/handlers/websocket_handler.go
+++ b/cmd/kueueviz/backend/handlers/websocket_handler.go
@@ -131,7 +131,6 @@ func (h *Handlers) sendData(ctx context.Context, conn *websocket.Conn, dataFetch
 	return nil
 }
 
-
 // handleInformerUpdates uses Kubernetes informers to stream real-time changes
 // Supports watching multiple resource types by registering handlers for all provided GVKs
 func (h *Handlers) handleInformerUpdates(ctx context.Context, conn *websocket.Conn, token string, dataFetcher func(ctx context.Context) (any, error), gvks ...schema.GroupVersionKind) {
--- a/cmd/kueueviz/backend/handlers/websocket_handler.go
+++ b/cmd/kueueviz/backend/handlers/websocket_handler.go
@@ -135,7 +135,7 @@ func (h *Handlers) sendData(ctx context.Context, conn *websocket.Conn, dataFetch
 // re-verifies the bearer token against the Kubernetes TokenReview API.
 // This bounds the window in which a revoked or expired token can still
 // receive cluster data.
-const tokenRevalidationInterval = 30 * time.Second
+var tokenRevalidationInterval = 30 * time.Second
 
 // handleInformerUpdates uses Kubernetes informers to stream real-time changes
 // Supports watching multiple resource types by registering handlers for all provided GVKs
--- a/cmd/kueueviz/backend/handlers/websocket_handler.go
+++ b/cmd/kueueviz/backend/handlers/websocket_handler.go
@@ -58,6 +58,10 @@ func (h *Handlers) GenericWebSocketHandler(dataFetcher func(ctx context.Context)
 		startTime := time.Now()
 		slog.Debug("WebSocket handler started")
 
+		// Extract the bearer token stored by auth.Middleware (empty when auth is disabled).
+		token, _ := c.Get("token")
+		tokenStr, _ := token.(string)
+
 		// Upgrade the HTTP connection to a WebSocket connection
 		connStart := time.Now()
 		conn, err := upgrader.Upgrade(c.Writer, c.Request, nil)
@@ -91,7 +95,7 @@ func (h *Handlers) GenericWebSocketHandler(dataFetcher func(ctx context.Context)
 		}
 
 		// Use informer-based updates for real-time streaming
-		h.handleInformerUpdates(ctx, conn, dataFetcher, gvks...)
+		h.handleInformerUpdates(ctx, conn, tokenStr, dataFetcher, gvks...)
 
 		slog.Debug("WebSocket handler completed", "duration", time.Since(startTime))
 	}
@@ -127,9 +131,15 @@ func (h *Handlers) sendData(ctx context.Context, conn *websocket.Conn, dataFetch
 	return nil
 }
 
+// tokenRevalidationInterval controls how often an open WebSocket connection
+// re-verifies the bearer token against the Kubernetes TokenReview API.
+// This bounds the window in which a revoked or expired token can still
+// receive cluster data.
+const tokenRevalidationInterval = 30 * time.Second
+
 // handleInformerUpdates uses Kubernetes informers to stream real-time changes
 // Supports watching multiple resource types by registering handlers for all provided GVKs
-func (h *Handlers) handleInformerUpdates(ctx context.Context, conn *websocket.Conn, dataFetcher func(ctx context.Context) (any, error), gvks ...schema.GroupVersionKind) {
+func (h *Handlers) handleInformerUpdates(ctx context.Context, conn *websocket.Conn, token string, dataFetcher func(ctx context.Context) (any, error), gvks ...schema.GroupVersionKind) {
 	// Channel to signal when debounced update should be sent
 	updateChan := make(chan struct{}, 1)
 
@@ -192,11 +202,35 @@ func (h *Handlers) handleInformerUpdates(ctx context.Context, conn *websocket.Co
 		registrations = append(registrations, registrationInfo{gvk: gvk, reg: registration})
 	}
 
+	// Periodic token re-validation: close the connection if the bearer token
+	// is revoked or has expired. Only active when a TokenValidator is wired in.
+	var authTicker *time.Ticker
+	var authTickerC <-chan time.Time
+	if h.validator != nil && token != "" {
+		authTicker = time.NewTicker(tokenRevalidationInterval)
+		defer authTicker.Stop()
+		authTickerC = authTicker.C
+	}
+
 	// Handle updates from the informers
 	for {
 		select {
 		case <-ctx.Done():
 			return
+		case <-authTickerC:
+			ok, err := h.validator.ValidateToken(ctx, token)
+			if err != nil {
+				slog.Warn("Token re-validation failed; closing WebSocket", "error", err)
+			} else if !ok {
+				slog.Info("Bearer token is no longer valid; closing WebSocket")
+			}
+			if err != nil || !ok {
+				_ = conn.WriteMessage(
+					websocket.CloseMessage,
+					websocket.FormatCloseMessage(websocket.ClosePolicyViolation, "token expired or revoked"),
+				)
+				return
+			}
 		case <-updateChan:
 			if err := h.sendData(ctx, conn, dataFetcher); err != nil {
 				slog.Error("Error sending update", "error", err)
--- a/cmd/kueueviz/backend/handlers/websocket_handler.go
+++ b/cmd/kueueviz/backend/handlers/websocket_handler.go
@@ -131,11 +131,6 @@ func (h *Handlers) sendData(ctx context.Context, conn *websocket.Conn, dataFetch
 	return nil
 }
 
-// tokenRevalidationInterval controls how often an open WebSocket connection
-// re-verifies the bearer token against the Kubernetes TokenReview API.
-// This bounds the window in which a revoked or expired token can still
-// receive cluster data.
-var tokenRevalidationInterval = 30 * time.Second
 
 // handleInformerUpdates uses Kubernetes informers to stream real-time changes
 // Supports watching multiple resource types by registering handlers for all provided GVKs
@@ -207,7 +202,7 @@ func (h *Handlers) handleInformerUpdates(ctx context.Context, conn *websocket.Co
 	var authTicker *time.Ticker
 	var authTickerC <-chan time.Time
 	if h.validator != nil && token != "" {
-		authTicker = time.NewTicker(tokenRevalidationInterval)
+		authTicker = time.NewTicker(h.tokenRevalidationInterval)
 		defer authTicker.Stop()
 		authTickerC = authTicker.C
 	}
--- a/cmd/kueueviz/backend/handlers/websocket_handler.go
+++ b/cmd/kueueviz/backend/handlers/websocket_handler.go
@@ -218,13 +218,19 @@ func (h *Handlers) handleInformerUpdates(ctx context.Context, conn *websocket.Co
 		case <-ctx.Done():
 			return
 		case <-authTickerC:
-			ok, err := h.validator.ValidateToken(ctx, token)
+			// Use a short timeout so a slow or unavailable API server cannot
+			// block this goroutine indefinitely.
+			validateCtx, validateCancel := context.WithTimeout(context.Background(), 5*time.Second)
+			ok, err := h.validator.ValidateToken(validateCtx, token)
+			validateCancel()
 			if err != nil {
 				slog.Warn("Token re-validation failed; closing WebSocket", "error", err)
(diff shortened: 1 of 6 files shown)


cmd/kueueviz/backend/handlers/websocket_handler.go
```diff
@@ -179,11 +189,35 @@ func (h *Handlers) handleInformerUpdates(ctx context.Context, conn *websocket.Co
...
+			} else if !ok {
+				slog.Info("Bearer token is no longer valid; closing WebSocket")
+			}
+			if err != nil || !ok {
+				_ = conn.WriteMessage(
+					websocket.CloseMessage,
+					websocket.FormatCloseMessage(websocket.ClosePolicyViolation, "token expired or revoked"),
+				)
+				return
+			}
```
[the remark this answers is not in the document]
Vaishnav88sk: done

cmd/kueueviz/backend/middleware/auth.go
```diff
@@ -123,10 +124,36 @@ func (a *Authenticator) Middleware() gin.HandlerFunc {
...
+	if authenticated {
+		ttl = a.config.CacheTTL
+	}
+	a.cache.Set(key, cacheEntry{
+		authenticated: authenticated,
+		username:      username,
+		expiresAt:     now.Add(ttl),
+	})
+	return authenticated, nil
+}
```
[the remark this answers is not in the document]
Vaishnav88sk: done

mimowo:
/ok-to-test
cc @olekzabl

cmd/kueueviz/backend/handlers/websocket_handler.go
```diff
@@ -179,11 +189,41 @@ func (h *Handlers) handleInformerUpdates(ctx context.Context, conn *websocket.Co
...
+				slog.Info("Bearer token is no longer valid; closing WebSocket")
+			}
+			if err != nil || !ok {
+				// Set a write deadline so the close-frame write cannot block indefinitely.
+				_ = conn.SetWriteDeadline(time.Now().Add(5 * time.Second))
+				_ = conn.WriteMessage(
+					websocket.CloseMessage,
+					websocket.FormatCloseMessage(websocket.ClosePolicyViolation, "token expired or revoked"),
+				)
+				return
```
olekzabl, line 236: When we return here, what will be the end effect for the UI user?

Looking at this frontend code:
https://github.com/kubernetes-sigs/kueue/blob/6485c5729c53a7f63d35708796c96e6fa80729eb/cmd/kueueviz/frontend/src/useWebSocket.js#L80-L82
will it be just an error logged in the console, and besides that, a silent freeze of the UI?

Would it make sense to make it more visible (say, by displaying an error banner?)

olekzabl (commented):
The code looks nice, and I like the refactor around caching, thank you!
However, I'd like to understand how this will render for an UI user (see [this comment](https://github.com/kubernetes-sigs/kueue/pull/12698/changes#r3520419177))

olekzabl:
One more question: how feasible would it be to add unit test coverage for this change?

cmd/kueueviz/backend/handlers/websocket_handler.go
Vaishnav88sk: Good catch! I've updated the `useWebSocket` hook in the frontend to properly handle the `onclose` event. Since the backend sends a `1008 Policy Violation` code when a token expires, the UI will now display an explicit error banner instructing the user to log in again instead of silently freezing.

Vaishnav88sk:
> One more question: how feasible would it be to add unit test coverage for this change?

It is definitely feasible! However, since `useWebSocket.test.js` currently only tests the pure `parseWebSocketMessage` function and doesn't test the React hook itself, we'd need to introduce a bit of testing infrastructure.

Specifically, since we don't currently have React testing utilities in `package.json`, we would need to:
1. Add `@testing-library/react` to our `devDependencies`.
2. Mock the global `WebSocket` object in the vitest environment.
3. Use `renderHook` to mount `useWebSocket`, trigger `ws.onclose({ code: 1008 })`, and assert that the returned `error` state updates correctly.

I'm more than happy to add this testing setup and the unit tests in this PR, or we could handle the React hook testing infrastructure in a separate follow-up PR. Let me know which you prefer!

olekzabl:
> > One more question: how feasible would it be to add unit test coverage for this change?
> 
> It is definitely feasible! However, since `useWebSocket.test.js` currently only tests (...)

Ah, my apologies, I wasn't clear.

By "add unit test coverage for this change", I mainly meant this PR generally. So, its _backend_ part.
(Which I'm hoping to be easier than the frontend part).

As for the frontend part, I understand the current coverage is suboptimal, and ideally we should have the frontend unit tests generally much better developed (cf. #12633). Hence, I consider improving this out of scope of this PR.

> I'm more than happy to add this testing setup and the unit tests in this PR, or we could handle the React hook testing infrastructure in a separate follow-up PR. Let me know which you prefer!

Exactly, I think improving test infrastructure is out of scope of this PR. Especially for the frontend.

If you'd like to work on #12633 - it'll be great!

Vaishnav88sk:
> Ah, my apologies, I wasn't clear.
> 
> By "add unit test coverage for this change", I mainly meant this PR generally. So, its _backend_ part. (Which I'm hoping to be easier than the frontend part).
> 
> As for the frontend part, I understand the current coverage is suboptimal, and ideally we should have the frontend unit tests generally much better developed (cf. #12633). Hence, I consider improving this out of scope of this PR.
> 
> > I'm more than happy to add this testing setup and the unit tests in this PR, or we could handle the React hook testing infrastructure in a separate follow-up PR. Let me know which you prefer!
> 
> Exactly, I think improving test infrastructure is out of scope of this PR. Especially for the frontend.
> 
> If you'd like to work on #12633 - it'll be great!

Ah, that makes perfect sense! Thanks for the clarification. I will go ahead and add the backend unit tests for the revalidation logic in this PR.

Regarding the frontend testing setup (#12633), it looks like that issue already has an active assignee, so I'll leave it to them to avoid stepping on any toes. I'll focus on getting the backend tests pushed here shortly!

cmd/kueueviz/backend/handlers/websocket_handler_test.go
```diff
@@ -226,6 +253,62 @@ func TestWebSocketHandleInformerUpdates(t *testing.T) {
 				}
 			},
 		},
+		"closes connection when token becomes invalid": {
+			run: func(t *testing.T) {
+				origInterval := tokenRevalidationInterval
+				tokenRevalidationInterval = 100 * time.Millisecond
+				defer func() { tokenRevalidationInterval = origInterval }()
```
olekzabl: This feels out of style, even if it isn't causing troubles right now.
And it may cause troubles in the future, e.g. prevent parallelizing this test suite (as it'd lead to races).

How about making "token revalidation interval" a field in `Handlers`? 
Then, this test could set it in a well-scoped way, as it constructs `Handlers{...}` below. 
And non-test code would set it on its side independently.

cmd/kueueviz/backend/handlers/websocket_handler_test.go
```diff
@@ -42,6 +42,33 @@ var closedChan = func() <-chan struct{} {
 	return ch
 }()
 
+type mockTokenValidator struct {
+	mu    sync.Mutex
+	valid bool
+	err   error
```
olekzabl: Is this field used (ever set to non-`nil`)? How about cleaning up?

olekzabl (commented):
Thank you for adding tests!

Just 2 minor comments.

cmd/kueueviz/backend/handlers/websocket_handler_test.go
Vaishnav88sk: You're right, the `err` field was never set from outside and only introduced noise. I've removed it, `ValidateToken` on the mock now always returns `nil` error, and `setValid` takes just a `bool`.

cmd/kueueviz/backend/handlers/websocket_handler_test.go
Vaishnav88sk: Using a package-level `var` for test control is fragile and prevents parallelism. So, I've moved `tokenRevalidationInterval` to a field on `Handlers` (defaulting to 30s via `New()`), so the test now sets it directly on the `Handlers{...}` struct it constructs, no global mutation.

cmd/kueueviz/backend/handlers/websocket_handler.go
```diff
@@ -179,11 +183,41 @@ func (h *Handlers) handleInformerUpdates(ctx context.Context, conn *websocket.Co
...
+
 	// Handle updates from the informers
 	for {
 		select {
 		case <-ctx.Done():
 			return
+		case <-authTickerC:
+			// Use a short timeout so a slow or unavailable API server cannot
+			// block this goroutine indefinitely.
+			validateCtx, validateCancel := context.WithTimeout(context.Background(), 5*time.Second)
```
mimowo: Avoid using the Background context, because this means the gotoutine and the request wil continue even during shutdown. Normally we prefer to shutdown the app as soon as we receive SIGTERM.

cmd/kueueviz/backend/handlers/websocket_handler_test.go
```diff
@@ -226,6 +251,63 @@ func TestWebSocketHandleInformerUpdates(t *testing.T) {
...
+					}
+					var closeErr *websocket.CloseError
+					if errors.As(err, &closeErr) {
+						if closeErr.Code != websocket.ClosePolicyViolation {
+							t.Fatalf("expected close code %d, got %d", websocket.ClosePolicyViolation, closeErr.Code)
+						}
+					} else {
+						t.Fatalf("expected CloseError, got %v", err)
+					}
+				case <-time.After(2 * time.Second):
```
mimowo: Avoid magic constants - declare some kind of timeout constant as 5s

cmd/kueueviz/frontend/src/useWebSocket.js
```diff
@@ -78,7 +78,14 @@ const useWebSocket = (url) => {
...
 
     ws.onclose = (event) => {
-      console.log('WebSocket connection closed', 'code:', event.code);
+      console.log('WebSocket connection closed', 'code:', event.code, 'reason:', event.reason);
+      if (event.code !== 1000 && event.code !== 1001) {
+        if (event.code === 1008) {
+          setError('WebSocket connection closed: Token expired or revoked. Please log in again.');
+        } else {
+          setError(`WebSocket connection closed unexpectedly (code: ${event.code}). Please refresh the page.`);
+        }
```
mimowo: use switch instead, and declare constants to avoid magic numbers. Even better if you can find the constants reusable from some http websocket depenedncy

cmd/kueueviz/backend/handlers/websocket_handler.go
Vaishnav88sk: Switched from `context.Background()` to `ctx` to ensure the token re-validation goroutine respects `SIGTERM` and shuts down correctly.

cmd/kueueviz/backend/handlers/websocket_handler_test.go
Vaishnav88sk: Removed the magic 5s numbers and declared a `tokenValidationTimeout = 5 * time.Second` constant. (I also adopted the newly merged `writeDeadlineExtension` constant from `main` to resolve the conflict).

cmd/kueueviz/frontend/src/useWebSocket.js
Vaishnav88sk: Updated `useWebSocket.js` to use a `switch` statement with explicit constants instead of magic numbers.

Vaishnav88sk:
branch updated with resolved conflicts.

cmd/kueueviz/backend/handlers/websocket_handler_test.go
```diff
@@ -226,6 +251,63 @@ func TestWebSocketHandleInformerUpdates(t *testing.T) {
...
+					tokenRevalidationInterval: 100 * time.Millisecond,
+				}
+				conn, closeServer := newTestWebSocketConnectionWithToken(t, h, "test-token", dataFetcher, gvk)
+				defer closeServer()
+
+				// Read initial message
+				readMessage(t, conn)
+
+				// Token should be validated successfully initially (by the ticker)
+				waitUntil(t, 2*time.Second, 10*time.Millisecond, func() bool {
```
mimowo: Seems like here we also have some magic consts, let's reuse the consts introduced in other cases.

cmd/kueueviz/backend/handlers/websocket_handler_test.go
mimowo: nit, same here

mimowo:
/approve
leaving final lgtm to @olekzabl or @danylott 
/cherrypick release-0.18
/cherrypick release-0.17
Thank you 👍

cmd/kueueviz/backend/handlers/websocket_handler_test.go
Vaishnav88sk: Resolved

cmd/kueueviz/backend/handlers/websocket_handler_test.go
Vaishnav88sk: thanks for nitpick
updated

cmd/kueueviz/frontend/src/useWebSocket.js
```diff
@@ -78,7 +78,14 @@ const useWebSocket = (url) => {
...
+      }
```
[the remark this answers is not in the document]
danylott: @Vaishnav88sk please take a look

cmd/kueueviz/backend/middleware/auth.go
```diff
@@ -88,45 +88,56 @@ func (a *Authenticator) Middleware() gin.HandlerFunc {
...
+		username:      username,
+		expiresAt:     now.Add(ttl),
+	})
+	return authenticated, username, nil
+}
+
+// ValidateToken performs a live TokenReview against the Kubernetes API,
+// intentionally bypassing the shared cache so that WebSocket re-checks always
+// reflect the current revocation state of the token. The result is NOT written
+// back to the shared cache to avoid extending a revoked session.
```
danylott, line 135: Fully omitting cache can cause some performance problem, can we set a short TTL for a cache here?

mimowo: +1 I guess the intention is to make sure persmission are always fresh. Otherwise there might be some issues that  someone may see content they no longer have permission to.

OTOH no matter how much we try avoiding cache with permissions this is a separate request, so we hit TOCTOU anyway. 

So, having some positive time cache sgtm too, but lifetime would probably need to be short like 2s. At this point I'm not sure how much it benefits the KueueViz experience.

I'm ok to defer the optimization for later.  cc @Vaishnav88sk @danylott  wdyt?

danylott (approved):

cmd/kueueviz/frontend/src/useWebSocket.js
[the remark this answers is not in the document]
Vaishnav88sk: Good catch. I'll address this in a follow-up — it's minor UX and doesn't affect the token re-validation behavior in this PR.

cmd/kueueviz/backend/middleware/auth.go
Vaishnav88sk: Agree with deferring the short-TTL cache for now, fresh permissions are the priority here, and the perf impact seems small for KueueViz. Happy to revisit with a ~2s TTL in a follow-up if we see issues.

mimowo: Let's open a follow up issue to track the effort

mimowo: @coderabbitai would you like to open the KueueViz issue to introuduce the positive cache for token review?

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for the contribution. Please CP in case the robot fails

mimowo:
@Vaishnav88sk could you also prepare CP for 0.17?

Vaishnav88sk:
> @Vaishnav88sk could you also prepare CP for 0.17?

CP #13164

PR #12699, opened 2026-07-01
feat: make visibility auth-reader RoleBinding optional

amy:
#### What type of PR is this?

/kind feature

The Helm chart always renders the `kueue-visibility-server-auth-reader` RoleBinding into the `kube-system` namespace, and there is currently no way to disable it (`enableVisibilityAPF` gates only the visibility-apf FlowSchema/PriorityLevelConfiguration, not this RoleBinding).

This is a problem for installs under namespace-scoped GitOps projects that are not permitted to target `kube-system`. 

NONE

```release-note
Helm: Added enableVisibilityAuthReaderRoleBinding Helm value (default: true) to make the visibility server's auth-reader RoleBinding in kube-system optional. Set to false when deploying under a GitOps project that cannot manage resources in kube-system, and create the RoleBinding out-of-band instead.
```

## Summary by CodeRabbit

* **New Features**
  * Added a Helm chart option to control whether the visibility server’s authentication RoleBinding is created.
  * The option is enabled by default, and can be disabled for GitOps setups that can’t manage resources in the system namespace.
  * Updated documentation with guidance for creating the RoleBinding externally when disabled.
* **Tests**
  * Added Helm tests to verify RoleBinding rendering for default, enabled, and disabled configurations.


charts/kueue/README.md
```diff
@@ -143,6 +143,7 @@ The following table lists the configurable parameters of the kueue chart and the
 | enableKueueViz | bool | `false` | Enable KueueViz dashboard |
 | enablePrometheus | bool | `false` | Enable Prometheus |
 | enableVisibilityAPF | bool | `false` | Enable API Priority and Fairness configuration for the visibility API |
+| enableVisibilityAuthReaderRoleBinding | bool | `true` | Enable the visibility server's auth-reader RoleBinding in the kube-system namespace. Disable when deploying under a GitOps project that cannot target kube-system; the RoleBinding must then be created out-of-band. |
```
kannon92: Why would `kube-system` be used here?

The namespace is `{{ .Release.Namespace}}` so it should be whatever the namespace is you are installing your helm chart into.

amy:
```yaml
metadata:
  namespace: kube-system                          # where the RoleBinding lives
roleRef:
  kind: Role
  name: extension-apiserver-authentication-reader # the Role it binds to
subjects:
  - kind: ServiceAccount
    namespace: '{{ .Release.Namespace }}'          # who gets the grant
```

The binding must live in `kube-system` because it references the built-in `extension-apiserver-authentication-reader` Role, and a RoleBinding can only reference a Role in its own namespace That Role only exists in `kube-system`. Using `.Release.Namespace` would point at a Role that isn't there and break visibility auth.

The `.Release.Namespace` is just the `subjects` ServiceAccount (the controller-manager), which does run in the install namespace. So that's correct.

The PR doesn't change any of this; it just adds a flag to skip the binding so projects that can't target `kube-system` can create it out-of-band.

mbobrovskyi:
Thank you!

/lgtm

mbobrovskyi:
Proposal for release notes:

/release-note-edit
```release-note
Helm: Added enableVisibilityAuthReaderRoleBinding Helm value (default: true) to make the visibility server's auth-reader RoleBinding in kube-system optional. Set to false when deploying under a GitOps project that cannot manage resources in kube-system, and create the RoleBinding out-of-band instead.
```

charts/kueue/templates/visibility/role_binding.yaml
```diff
@@ -16,6 +16,7 @@ limitations under the License.
 
 {{/* Code generated by yaml-processor. DO NOT EDIT. */}}
 
+{{- if .Values.enableVisibilityAuthReaderRoleBinding }}
```
mimowo, line 19: It would be nice to add a helm unit test for this change

hack/processing-plan.yaml
```diff
@@ -133,6 +133,16 @@ files:
           {{- end }}
         indentation: 2
         onFileCondition: '.kind == "APIService"'
+      - type: INSERT_TEXT
+        position: START
+        value: |
+          {{- if .Values.enableVisibilityAuthReaderRoleBinding }}
+        onFileCondition: '.kind == "RoleBinding"'
```
mimowo: Hm, IIUC this works because we only have one instance with kind "RoleBinding", but this feels fragile. Could we add an additional check for the name or labels?

mimowo:
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for the helm enhancement

PR #12631, opened 2026-06-29
Update KEP 7066 to include Workload labels in custom metrics

Singularity23x0:
#### What type of PR is this?
/kind kep

#### What this PR does / why we need it:
Updates the Custom Label Metrics KEP to account for workload-level labels.

#### Which issue(s) this PR fixes:

Progresses #12629

#### Special notes for your reviewer:
AI used for formatting and editorial fixes.
#### Does this PR introduce a user-facing change?

```release-note
NONE
```

## Summary by CodeRabbit

* **Documentation**
  * Updated the KEP to support Prometheus custom metric labels sourced from reported workloads (in addition to existing queue-based sources).
  * Expanded configuration guidance with `customWorkloadLabels`, including an example to break down admitted active workloads by `priority_class`.
  * Clarified workload-level metrics behavior and grouping expectations (including pre-aggregation requirements, label naming, and validation timing).
  * Enhanced the integration test checklist to validate workload-level aggregation breakdowns.

Issue #12629, opened 2026-06-29
☂️ Custom Metric Labels - Workload Labels


Singularity23x0:
**What would you like to be added**:
Expand the Custom Metric Labels mechanism (KEP-7066) for workload labels.

**Why is this needed**:
To allow breaking down workload metrics by custom labels. This will allow metrics like: `admitted_active_workloads ` or `kueue_pending_workloads ` to provide summaries broken down by e.g. flavor or priority, by setting them as values of workload labels.

**Completion requirements**:

* [x] KEP-7066 is amended to include Workload Labels,
* [x] POC: Support for workload custom metric labels is added to `kueue_admitted_active_workloads `
* [ ] Cluster queue metrics support:
   - [x] `pending_workloads`: https://github.com/kubernetes-sigs/kueue/issues/13586
   - [ ] `workloads_admitted_total`: https://github.com/kubernetes-sigs/kueue/issues/13589
 * [x] Local queue metrics support:
   - [x] `local_queue_admitted_active_workloads`: https://github.com/kubernetes-sigs/kueue/issues/13587
   - [x] `local_queue_admitted_workloads_total`: https://github.com/kubernetes-sigs/kueue/issues/13587
   - [x] `local_queue_pending_workloads`: https://github.com/kubernetes-sigs/kueue/issues/13588
 * [ ] Cohort metrics support:
   - [ ] `cohort_subtree_admitted_workloads_total`: https://github.com/kubernetes-sigs/kueue/issues/14185
   - [ ] `cohort_subtree_admitted_active_workloads`: https://github.com/kubernetes-sigs/kueue/issues/14186
* [ ] Data leak prevention test: #15202

This enhancement requires the following artifacts:

- [x] Design doc
- [x] API change
- [x] Docs update

The artifacts should be linked in subsequent comments.

amy:
following

Singularity23x0:
A task for creating a util to enhance test development experience:
- [x] https://github.com/kubernetes-sigs/kueue/issues/13860

carmal891:
@Singularity23x0 Please check if it makes sense to cover this metric `kueue_cohort_subtree_pending_workloads`

PR : https://github.com/kubernetes-sigs/kueue/pull/14149

Singularity23x0:
- [ ] #14555

PR #12713, opened 2026-07-02
Implement support for new custom labels API.

Singularity23x0:
#### What type of PR is this?
/kind feature
/kind api-change

Implements the KEP-7066 update ([#12631](https://github.com/kubernetes-sigs/kueue/pull/12631)):
* support for SourceKind field
* support for TrackedValues field

CustomLabels rework:
* consolidated access to custom label caching,
* support for workload labels.

What will be addressed in a follow up PR:
* label and annotation copying for Workload objects
* admitte_active_workloads metric utilization of the added mechanisms

Progresses #12629

Groundwork for expanding the admitted active workloads metric with workload-sourced labels support.

```release-note
Observability: Updated the alpha custom metric labels API with support for source-specific labels, including labels sourced from Workloads, and allowlisting of tracked label values. The feature remains disabled by default and can be enabled using the `CustomMetricLabels` feature gate.
```

## Summary by CodeRabbit

* **Enhancements**
  * Improved custom metric label handling across cluster queues, local queues, and cohorts using source-kind scoped label names and a shared kind-aware lookup for consistent reporting.
  * Metrics wiring now derives the correct label sets up front, ensuring label ordering matches reported values across status, active workload, and quota/resource series.

* **Bug Fixes**
  * Fixed custom-label sourcing for local-queue and cohort metric reporting, including during resync and termination, with improved updates when definitions change.

* **Tests**
  * Updated and expanded custom-label and controller metric resync expectations to match the new source-kind scoping and naming.


site/content/en/docs/reference/kueue-config.v1beta2.md
```diff
@@ -1321,6 +1329,18 @@ for Dynamic Resource Allocation support.</p>
 </tbody>
 </table>
 
+## `SourceKind`     {#config-kueue-x-k8s-io-v1beta2-SourceKind}
```
mwielgus, line 1351: Can we have some real world examples of how to use it in practice?

apis/config/v1beta2/configuration_types.go
```diff
@@ -185,37 +185,62 @@ type ControllerMetrics struct {
...
-	// automatically prepends "custom_" (e.g., name: "team" becomes label "custom_team").
-	// Must follow Prometheus label naming conventions: [a-zA-Z_][a-zA-Z0-9_]*.
+	// Name is the Prometheus metric label name suffix.
+	// Prepended with "custom_" to form the full Prometheus label name
+	// (e.g., "team" becomes "custom_team").
+	// Must contain only [a-zA-Z0-9_] characters and start with a letter.
 	Name string `json:"name"`
 
 	// SourceLabelKey is the Kubernetes label key to read the value from.
-	// Must be a valid Kubernetes qualified name.
```
olekzabl, line 219: Out of curiosity, why removing this sentence?
(and same for `SourceAnnotationKey`)

pkg/cache/scheduler/cache.go
```diff
@@ -399,7 +396,7 @@ func (c *Cache) TerminateClusterQueue(name kueue.ClusterQueueReference) {
 	defer c.Unlock()
 	if cq := c.hm.ClusterQueue(name); cq != nil {
 		cq.Status = terminating
-		metrics.ReportClusterQueueStatus(cq.Name, cq.Status, cq.customMetricLabelValues, c.roleTracker)
+		metrics.ReportClusterQueueStatus(cq.Name, cq.Status, cq.GetCustomLabelValues(), c.roleTracker)
```
olekzabl, line 399: Out of curiosity: 

- Why have you switched from cached `customMetricLabelValues` to dynamic `GetCustomLabelValues()`?
- Why have you done it for CQ but not for LQ?

Even if it's a good move, I'd like to check with you if it should be done in this PR.
Its title suggests just supporting `SourceKinds`. Its description suggests just `laying groundwork`. 
OTOH this change looks like sth beyond that scope.

pkg/metrics/custom_labels.go
```diff
@@ -17,166 +17,214 @@ limitations under the License.
...
+	values     *utilmaps.SyncMap[string, []string]
 }
 
-// Store saves values and returns true if they changed.
-func (s CustomLabelStore[K]) Store(key K, newVals []string) bool {
-	old, existed := s.m.Swap(string(key), newVals)
-	return existed && !slices.Equal(old, newVals)
+func NewCustomLabels(entries []configapi.ControllerMetricsCustomLabel) *CustomLabels {
+	if len(entries) == 0 {
+		return nil
```
olekzabl, line 46: Nit: I'm not a fan of `return nil` here, as it brings a need for all `CustomLabels` methods to do `nil` checks.
While, if you kept returning "a real object with empty content" (as the old impl did), the logic without these checks would work.

(Probably with a few more pointer dereferences, true, but I'd value simpler code over optimizations at that level).

pkg/metrics/custom_labels.go
```diff
@@ -17,166 +17,214 @@ limitations under the License.
...
+	InitMetricVectors(cl)
+	return cl
 }
 
-func (s CustomLabelStore[K]) Get(key K) []string {
-	vals, _ := s.m.Get(string(key))
-	return vals
+// LabelNames returns the computed metric label names (e.g. "custom_team")
+// for the requested set of sources.
+// The labels are ordered by source kind then by the order of definition in the config.
```
olekzabl, line 65: Am I getting this right that the key thing about this ordering is that it must be consistent with the ordering of values returned by `GetFor()`?

(So that, once we reach metrics supporting multiple kinds, they'll first have their labels "declared" via `.LabelNames()` here, and then the actual values emitted using `.GetFor()`, and there'll be no ordering mismatch?)

If so, you could strengthen the code comments to reflect the need for this compatibility.
(E.g. by just saying "it must be consistent with ordering of GetFor()", and maybe same disclaimer on that side).

Unless I'm getting your intent wrong; then I'm curious how it'll work in the end.

pkg/metrics/custom_labels_test.go
```diff
@@ -79,20 +88,81 @@ func TestExtractValues(t *testing.T) {
...
+			annotations: map[string]string{"billing/cost": "high", "cohort": "c1"},
+			sourceKinds: []configapi.SourceKind{configapi.SourceKindClusterQueue, configapi.SourceKindCohort},
+			want:        []string{"ml", "high", "c1"},
+			wantNames:   []string{"custom_team", "custom_cost", "custom_cohort"},
+		},
+		"all values tracked": {
+			entries: []configapi.ControllerMetricsCustomLabel{
+				{Name: "team", TrackedValues: []string{"infra", "platform"}},
+				{Name: "env", SourceLabelKey: "environment", TrackedValues: []string{"prod", "staging"}},
+				{Name: "cost", TrackedValues: []string{}},
```
olekzabl, line 121: The test case name is "all values tracked". Mismatch?

pkg/metrics/custom_labels_test.go
```diff
@@ -79,20 +88,81 @@ func TestExtractValues(t *testing.T) {
...
 			var cl *CustomLabels
 			if tc.entries != nil {
+				// Set default SourceKind for test entries
+				for i := range tc.entries {
+					e := &tc.entries[i]
+					if e.SourceKind == nil {
+						e.SourceKind = ptr.To(configapi.SourceKindClusterQueue)
+					}
+				}
+				// Create the CustomLabels objetc. Defer cleanup.
```
olekzabl: Why do we need it?
AFAICS `NewCustomLabels` includes the defaulting in its logic.

The way it is now, we're not testing this defaulting; instead, we're silently modifying test data. Why?

pkg/metrics/custom_labels_test.go
```diff
@@ -79,20 +88,81 @@ func TestExtractValues(t *testing.T) {
...
 					InitMetricVectors(nil)
 				})
 			}
-			got := cl.ExtractValues(tc.labels, tc.annotations)
-			if diff := cmp.Diff(tc.want, got); diff != "" {
+
+			srcKinds := []configapi.SourceKind{configapi.SourceKindClusterQueue}
+			if len(tc.sourceKinds) > 0 {
+				srcKinds = tc.sourceKinds
+			}
```
olekzabl: This seems like effectively another case of silently modifying test data.
(Instead of always using `tc.SourceKinds`, replace it with `{CQ}` when it's empty).

Wouldn't it be cleaner to update `.SourceKinds` in the test cases and just use that?

apis/config/v1beta2/configuration_types.go
Singularity23x0: This seems like an accident, cleaning up.

pkg/metrics/custom_labels.go
Singularity23x0: You understand correctly. Adding this to the commend sounds like a good idea.

pkg/metrics/custom_labels_test.go
Singularity23x0: The test case is for when all label values are tracked - meaning allowed by the defined TrackedValues lists.

pkg/metrics/custom_labels_test.go
Singularity23x0: Artifact from a previous implementation. Removing.

pkg/metrics/custom_labels_test.go
Singularity23x0: We default to using SourceKindClusterQueue in practice.
This part of the test aims to reflect that - the config sets the value to empty, the test ensures the empty value defaults correctly (since we do not invoke the mechanism doing this in our effective code here).

pkg/cache/scheduler/cache.go
Singularity23x0: > Why have you switched from cached customMetricLabelValues to dynamic GetCustomLabelValues()?
> Why have you done it for CQ but not for LQ?[ ...] if it should be done in this PR.

As part of the "groundwork", the CQ needs to have access to the CustomLabels object (it will be using it in the future when reporting admitted active workloads).
Since we are adding the CL object already, switching form `customMetricLabelValues` to  `GetCustomLabelValues()` was a low hanging fruit change - why hard code values, if we can get them form the source.
LQs did not require this addition of CL, so updating them right away was not a natural consequence of the core changes.

pkg/metrics/custom_labels.go
Singularity23x0: All CL methods already had this check as we allowed for it to be nil if the Feature Gate is off.
This way we unify the approach between FG off and no entries, instead of providing an empty object unnecessarily.

It is mostly a cosmetic change to me, so if the opinion on this is strong, I am okay with updating the CL creation.
To eliminate the need for nil checks in structure methods we would have to also update the logic of FG off - namely creating an empty CL instead of leaving the value at nil.

pkg/metrics/custom_labels_test.go
```diff
@@ -79,20 +88,74 @@ func TestExtractValues(t *testing.T) {
...
+
+			gotNames := cl.LabelNames(srcKinds...)
+			if diff := cmp.Diff(tc.wantNames, gotNames); diff != "" {
+				t.Errorf("LabelNames() mismatch (-want +got):\n%s", diff)
+			}
+
+			gotValues := cl.GetFor(sourcesMap)
+			if diff := cmp.Diff(tc.want, gotValues); diff != "" {
 				t.Errorf("ExtractValues() mismatch (-want +got):\n%s", diff)
 			}
```
olekzabl, line 169: The test is (still) named `TestExtractValues` but it's now testing `GetFor`.

In particular, the `Errorf` below is outdated.

pkg/metrics/custom_labels_test.go
olekzabl: Well, I find it surprising. Per the DAMP philosophy of our testing, 
I'd keep the inputs possibly close to what is actually being processed.

pkg/metrics/custom_labels.go
olekzabl: > All CL methods already had this check

Ah, ok, then it can stay. I thought I checked it - but I must have checked wrong.

pkg/metrics/custom_labels_test.go
Singularity23x0: Renamed to `TestCustomLabels`

pkg/cache/scheduler/cache.go
olekzabl: > why hard code values, if we can get them form the source.

For performance :) Though as I'm looking at it, I'm guessing the performance difference is negligible.

> LQs did not require this addition of CL, so updating them right away was not a natural consequence of the core changes.

OK but at the end of your chain of PRs I'd like LQs & CQs to be treated analogously (i.e. both cached or both dynamically computed). Unless you'll provide a reason for this distinction.

Singularity23x0: >For performance :) Though as I'm looking at it, I'm guessing the performance difference is negligible.

Its effectively a call to a map, so i think perf is not a concern here.

> OK but at the end of your chain of PRs I'd like LQs & CQs to be treated analogously (i.e. both cached or both dynamically computed). Unless you'll provide a reason for this distinction.

Using CustomLabels is my goal. I will add it here.

Singularity23x0: Updated.

site/content/en/docs/reference/kueue-config.v1beta2.md
Singularity23x0: Described in related KEP-7066.

pkg/cache/scheduler/cache.go
```diff
@@ -445,11 +442,9 @@ func (c *Cache) AddClusterQueue(ctx context.Context, cq *kueue.ClusterQueue) err
 			totalReserved:      make(resources.FlavorResourceQuantities),
 			admittedUsage:      make(resources.FlavorResourceQuantities),
 			labels:             q.GetLabels(),
+			customLabels:       c.customLabels,
 		}
-		if features.Enabled(features.CustomMetricLabels) {
-			qImpl.customMetricLabelValues = c.customLabels.ExtractValues(q.Labels, q.Annotations)
-		}
-		qImpl.resetFlavorsAndResources(cqImpl.resourceNode.Usage, cqImpl.AdmittedUsage)
```
mimowo, line 452: Why is resetFlavorsAndResources dropped?

pkg/cache/scheduler/cache.go
```diff
@@ -445,11 +442,9 @@ func (c *Cache) AddClusterQueue(ctx context.Context, cq *kueue.ClusterQueue) err
 			totalReserved:      make(resources.FlavorResourceQuantities),
 			admittedUsage:      make(resources.FlavorResourceQuantities),
 			labels:             q.GetLabels(),
+			customLabels:       c.customLabels,
 		}
-		if features.Enabled(features.CustomMetricLabels) {
-			qImpl.customMetricLabelValues = c.customLabels.ExtractValues(q.Labels, q.Annotations)
-		}
```
mimowo, line 451: Happy to see the duplications eliminated 🥇

pkg/cache/scheduler/cache.go
```diff
@@ -445,11 +442,9 @@ func (c *Cache) AddClusterQueue(ctx context.Context, cq *kueue.ClusterQueue) err
...
-		qImpl.resetFlavorsAndResources(cqImpl.resourceNode.Usage, cqImpl.AdmittedUsage)
+		qImpl.customLabels.LQStore(qKey, q.GetLabels(), q.Annotations)
```
mimowo, line 447: This need to be wrapped in FG

pkg/cache/scheduler/clusterqueue.go
```diff
@@ -111,6 +111,10 @@ func (c *clusterQueue) GetName() kueue.ClusterQueueReference {
 	return c.Name
 }
 
+func (c *clusterQueue) GetCustomLabelValues() []string {
+	return c.customLabels.CQGet(c.Name)
```
mimowo, line 115: Wrap in FG, otherwise return nil

pkg/cache/scheduler/localqueue.go
```diff
@@ -35,9 +35,14 @@ type LocalQueue struct {
...
-	customMetricLabelValues []string
-	labels                  map[string]string
+
+	// allows access to values extracted from K8s labels/annotations, used as custom Prometheus metric labels
+	customLabels *metrics.CustomLabels
+	labels       map[string]string
+}
+
+func (q *LocalQueue) customMetricLabelValues() []string {
+	return q.customLabels.LQGet(q.key)
```
mimowo, line 45: wrap in FG, otherwise nil

pkg/cache/scheduler/cache.go
Singularity23x0: If the FG is disable, customLabels will be nil, leading to calling Store(), which has a nil guard.
This line will already do nothing if the FG is disabled.

pkg/cache/scheduler/cache.go
Singularity23x0: Accident, restoring.

site/content/en/docs/reference/kueue-config.v1beta2.md
mimowo: Yeah, but I think we need to also update documentation. This needs to be a follow up.  Can be done slightly after release though

pkg/cache/scheduler/clusterqueue.go
Singularity23x0: Bc of nil gates, this will return a nil when the FG is disabled.

mimowo: maybe but that conclusion relies on correctness of other places. We want to repeat the gate check in various places of code to isolate the new code well - one can be sure this is nil without studying all entry points which may mutate the field

Singularity23x0:
Regarding gating CustomLabel calls behind the FG - current implementation ustilizies nil gating - if CL is nil, the methods called on it do nothing and return empty results. This allows us to gate only the creation of the CL begind the FG and make the calls to the object freely.

Let me know if there are any objections to such an approach.

pkg/cache/scheduler/cache.go
mimowo: i see, but still i think it will make reading easier if the code short circuits based on the FG. the new code should be well isolated from all access places.

mimowo:
im not questioning correctness but code isolation and thus readablity. some correctness assumptikns also change as the code evolves being developed by multiple developers. FG isolation is useful from all access places

pkg/cache/scheduler/cache.go
Singularity23x0: Should we apply this rule to existing code that does not gate the calls to CustomLabels behind the FG as well?

Singularity23x0: While I do not argue it will improve readability, it will introduce much redundancy to the codebase.
Otoh applying this philosophy only to some calls to the CL does not eliminate the readability issue across the board.

Singularity23x0: Alternatively, to minimize code duplication, while ensuring the logic is properly gated behind the FG, I propose adding a dedicated method returning: `if FG=Enabled && CL != nil`. We can then use it in all exposed CL methods to ensure the logic is isolated.

Lmk your opinion.

mimowo: > Should we apply this rule to existing code that does not gate the calls to CustomLabels behind the FG as well?

Can you list those actually? I'm not sure which places you are referring too.

Generally, we have places which are "write" for the field, and "read" from it. I think it is a good  practice to guard at least the write points, like [this](https://github.com/kubernetes-sigs/kueue/pull/12713/changes#diff-9d50921591b2d9622bc605a02ee4015136a176c9b9f8886ca5b10fd58d0c06dfR447).

Indeed, for "Read" access indeed this is a bit more nuanced, because if we assume all mutation places are guarded, then we can be sure that the "read" places always read nil / empty structs.  

Surely, we can avoid some FG checks and still prove correctness. However, from my experience, bugs happen and people may add new places doing "writes" which are not guarded, then the "reads" suddenly may start to return data.

I don't have a strong view about the "read" spots. Since I'm currently a bit underwater I have no time to carefully check correctness, so I wanted to prioritize code isolation.  If you think it is better to avoid duplication at the expense of isolation, and get lgtm from another reviewer who checks that carefully I'm happy to approve.

mimowo:
/approve
Thank you, this looks great overall. I would love to have another pair of reviewer eyes to leave lgtm. Maybe @kshalot

kshalot:
@Singularity23x0 @mimowo I'll do a pass later today.

pkg/cache/scheduler/cache.go
Singularity23x0: I think we could guard both writes and reads directly in the CustomLabels structure methods.
If we check the flag there, no matter where any writes or reads happen, CL methods will not perform any actions on those guarded methods.
This ensures nothing leaks while reducing the amount of places the FG has to be checked verbetum.

pkg/metrics/custom_labels.go
```diff
@@ -17,166 +17,235 @@ limitations under the License.
...
-		cohort: newCustomLabelStore[kueue.CohortReference](),
-	}
-	if len(entries) > 0 {
-		cl.labelEntries = entries
-		cl.names = make([]string, len(entries))
-		for i, e := range entries {
-			cl.names[i] = "custom_" + e.Name
+func (cl *CustomLabels) labelStoreIter(srcs ...configapi.SourceKind) iter.Seq2[*SourceKindLabelStore, configapi.SourceKind] {
+	return func(yield func(*SourceKindLabelStore, configapi.SourceKind) bool) {
+		slices.Sort(srcs)
```
kshalot: This will mutate the `srcs` slice that you pass as the argument. In our case it doesn't really matter, but it's an antipattern.

I'm okay to save the allocation, but I think cloning will not be a huge price if we feel it might save a bug in the future.

pkg/metrics/custom_labels.go
```diff
@@ -17,166 +17,235 @@ limitations under the License.
...
-		cq:     newCustomLabelStore[kueue.ClusterQueueReference](),
-		lq:     newCustomLabelStore[utilqueue.LocalQueueReference](),
-		cohort: newCustomLabelStore[kueue.CohortReference](),
-	}
-	if len(entries) > 0 {
-		cl.labelEntries = entries
-		cl.names = make([]string, len(entries))
-		for i, e := range entries {
-			cl.names[i] = "custom_" + e.Name
+func (cl *CustomLabels) labelStoreIter(srcs ...configapi.SourceKind) iter.Seq2[*SourceKindLabelStore, configapi.SourceKind] {
```
kshalot: The caller always has a slice, so maybe just use a slice instead of variadic

pkg/controller/core/cohort_controller.go
```diff
@@ -141,16 +140,14 @@ func (r *CohortReconciler) Create(e event.TypedCreateEvent[*kueue.Cohort]) bool
...
-			r.customLabels.CohortGet(kueue.CohortReference(e.ObjectNew.GetName())),
-			r.customLabels.ExtractValues(e.ObjectNew.GetLabels(), e.ObjectNew.GetAnnotations()),
-		)
-	}
+	clUpdateRequired := r.customLabels.UpdateRequired(
+		config.SourceKindCohort,
+		e.ObjectNew.GetName(),
+		e.ObjectNew.GetLabels(),
+		e.ObjectNew.GetAnnotations(),
+	)
```
kshalot: Why isn't this feature gated anymore?

Singularity23x0: Technically - the FG check is baked into UpdateRequired. This does however introduce readability issues, so I will make sure the FG is checked explicitly here.

pkg/cache/scheduler/cache.go
Singularity23x0: Implemented CustomLabels.enabled check.

kshalot: Apologies, this is still unclear to me - are you planning to wrap this with the FG or not?

Maybe this will get too cumbersome to align in this particular FG, but I think explicit checks were usually encouraged during code review.

I'm not arguing against the current implementation, but if there is no convention in this feature currently then I think we should take the time to align it (i.e. either rely on explicit check everywhere or rely on the nil check in the struct everywhere).

test/integration/singlecluster/controller/core/custom_labels_test.go
```diff
@@ -49,7 +50,8 @@ var _ = ginkgo.Describe("CustomMetricLabels", ginkgo.Label("controller:clusterqu
 			features.SetFeatureGateDuringTest(ginkgo.GinkgoTB(), features.LocalQueueMetrics, true)
 			controllersCfg := &config.Configuration{}
 			controllersCfg.Metrics.CustomLabels = []config.ControllerMetricsCustomLabel{
-				{Name: "team"},
+				{Name: "team_cq", SourceLabelKey: "team", SourceKind: ptr.To(config.SourceKindClusterQueue)},
+				{Name: "team_lq", SourceLabelKey: "team", SourceKind: ptr.To(config.SourceKindLocalQueue)},
```
mimowo, line 54: It would be great to have an integration test for SourceKind=Workload

test/integration/singlecluster/controller/core/custom_labels_test.go
```diff
@@ -292,7 +294,8 @@ var _ = ginkgo.Describe("CustomMetricLabels", ginkgo.Label("controller:clusterqu
 			features.SetFeatureGateDuringTest(ginkgo.GinkgoTB(), features.LocalQueueMetrics, false)
 			controllersCfg := &config.Configuration{}
 			controllersCfg.Metrics.CustomLabels = []config.ControllerMetricsCustomLabel{
-				{Name: "team"},
+				{Name: "team_cq", SourceLabelKey: "team", SourceKind: ptr.To(config.SourceKindClusterQueue)},
+				{Name: "team_lq", SourceLabelKey: "team", SourceKind: ptr.To(config.SourceKindLocalQueue)},
```
mimowo, line 298: Why do you split "team" into "team_cq" and "team_lq"? Is it just for readability of tests, ro the reason is deeper?

mimowo:
/lgtm
/approve
I looked again, and I think it is ready to merge as Alpha. So let me proceed to de-risk slipping the release, nothing strikes me. The isolation is improved.  I only missed integration tests for `SourceKind=Workload`, but I think we can add those in a follow up.

Thank you 👍

mimowo:
/release-note-edit
```release-note
Observability: Updated the alpha custom metric labels API with support for source-specific labels, including labels sourced from Workloads, and allowlisting of tracked label values. The feature remains disabled by default and can be enabled using the `CustomMetricLabels` feature gate.
```

PR #12719, opened 2026-07-02
Add UnadmittedWorkloadsExplicitStatus feature gate

j-skiba:
#### What type of PR is this?
/kind feature

This PR implements the `UnadmittedWorkloadsExplicitStatus` feature gate (disabled by default) as part of this [KEP](https://github.com/kubernetes-sigs/kueue/tree/main/keps/10852-inadmissible-workloads-observability).

When enabled, the workload controller immediately writes unadmitted conditions to a newly created workload's status during its initial reconciliation:
- `QuotaReserved = False` (Reason: `PendingEvaluation`, Message: "Workload is pending evaluation in the scheduling queue")
- `Admitted = False` (Reason: `NoReservation`, Message: "The workload has no reservation")

**Why we need it:**
Currently, newly created workloads have no status conditions initialized until they are first evaluated by the scheduling loop. This lack of initial state makes it difficult for users and dashboard tools to distinguish between a workload that is queued and waiting for evaluation versus a workload that Kueue has not reconciled yet. Explicitly initializing these unadmitted conditions immediately upon workload creation significantly improves observability of the queue state.

Part of #10852

it's based on #12510 and it's waiting for #1251 to merge first

```release-note
Scheduling: Added the `UnadmittedWorkloadsExplicitStatus` feature gate. When enabled, newly created workloads immediately receive explicit unadmitted status conditions (`QuotaReserved=False` and `Admitted=False`) during initial reconciliation to improve queue state observability.
```

## Summary by CodeRabbit

* **New Features**
  * Added beta feature gate for explicit status initialization on newly created unadmitted workloads.
  * When enabled (requires observability), workloads now immediately show `QuotaReserved=False` (`PendingEvaluation`) and `Admitted=False` (`NoReservation`).

* **Bug Fixes**
  * Preserves scheduler-provided `WorkloadQuotaReserved` reason/message across reconciliations for selected reasons.
  * Refined condition stamping/suppression behavior based on the feature gates.

* **Tests**
  * Added unit and integration coverage for the new explicit-status scenarios and recovery from misconfiguration.  

* **Documentation / Validation**
  * Updated feature gate lists and added configuration validation enforcing the gate dependency.

### DIFF
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -1868,9 +1868,8 @@ func (r *WorkloadReconciler) resolveGranularUnadmittedQuotaReservedCondition(
 		cond.Reason != kueue.WorkloadWaiting: //nolint:staticcheck // SA1019: legacy reason
 		return cond.Reason, cond.Message, nil
 	default:
-		// Stamping the condition on creation is deferred to the next feature gate:
-		// UnadmittedWorkloadsObservabilityExplicitStatus.
-		if cond == nil {
+		// Stamping the condition on creation is gated by UnadmittedWorkloadsExplicitStatus.
+		if !features.Enabled(features.UnadmittedWorkloadsExplicitStatus) && cond == nil {
 			return "", "", nil
 		}
 		return kueue.WorkloadQuotaReservedReasonPendingEvaluation, "Workload is pending evaluation in the scheduling queue", nil
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -1869,7 +1869,7 @@ func (r *WorkloadReconciler) resolveGranularUnadmittedQuotaReservedCondition(
 		return cond.Reason, cond.Message, nil
 	default:
 		// Stamping the condition on creation is gated by UnadmittedWorkloadsExplicitStatus.
-		if !features.Enabled(features.UnadmittedWorkloadsExplicitStatus) && cond == nil {
+		if cond == nil && !features.Enabled(features.UnadmittedWorkloadsExplicitStatus) {
 			return "", "", nil
 		}
 		return kueue.WorkloadQuotaReservedReasonPendingEvaluation, "Workload is pending evaluation in the scheduling queue", nil
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -82,8 +82,16 @@ var (
 	realClock = clock.RealClock{}
 
 	// schedulerSetReasons contains the reasons for the QuotaReserved=False condition
-	// that are set by the scheduler. These reasons must be preserved by the controller
-	// during reconciliation until the next scheduler cycle.
+	// that are set by the scheduler and must be preserved by the controller during
+	// reconciliation until the next scheduler cycle.
+	//
+	// We explicitly DO NOT include:
+	// - WorkloadQuotaReservedReasonSuspended
+	// - WorkloadQuotaReservedReasonMisconfigured
+	// Even though the scheduler can emit them, they represent active blockages (e.g.
+	// inactive or missing queues) that the controller actively re-evaluates. If we
+	// preserved them, the controller would not be able to transition the workload
+	// back to PendingEvaluation once the blockages are resolved.
 	schedulerSetReasons = sets.New(
 		kueue.WorkloadQuotaReservedReasonWaitingForQuota,
 		kueue.WorkloadQuotaReservedReasonNoMatchingFlavor,
@@ -91,6 +99,7 @@ var (
 		kueue.WorkloadQuotaReservedReasonTopologyPlacementFailed,
 		kueue.WorkloadQuotaReservedReasonWaitingForPreemptedWorkloads,
 		kueue.WorkloadQuotaReservedReasonWaitingForPodsReady,
+		kueue.WorkloadQuotaReservedReasonPendingEvaluation,
 	)
 )
 
--- a/pkg/controller/core/workload_controller.go
+++ b/pkg/controller/core/workload_controller.go
@@ -80,6 +80,18 @@ import (
 
 var (
 	realClock = clock.RealClock{}
+
+	// schedulerSetReasons contains the reasons for the QuotaReserved=False condition
+	// that are set by the scheduler. These reasons must be preserved by the controller
+	// during reconciliation until the next scheduler cycle.
+	schedulerSetReasons = sets.New(
+		kueue.WorkloadQuotaReservedReasonWaitingForQuota,
+		kueue.WorkloadQuotaReservedReasonNoMatchingFlavor,
+		kueue.WorkloadQuotaReservedReasonExceedsMaxQuota,
+		kueue.WorkloadQuotaReservedReasonTopologyPlacementFailed,
+		kueue.WorkloadQuotaReservedReasonWaitingForPreemptedWorkloads,
+		kueue.WorkloadQuotaReservedReasonWaitingForPodsReady,
+	)
 )
 
 // hasInternalError reports whether the field error list contains an internal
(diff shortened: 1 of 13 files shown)


j-skiba:
it's based on #12510 and it's waiting for #1251 to merge first

pkg/controller/core/workload_controller.go
```diff
@@ -1868,9 +1868,8 @@ func (r *WorkloadReconciler) resolveGranularUnadmittedQuotaReservedCondition(
 		cond.Reason != kueue.WorkloadWaiting: //nolint:staticcheck // SA1019: legacy reason
 		return cond.Reason, cond.Message, nil
 	default:
-		// Stamping the condition on creation is deferred to the next feature gate:
-		// UnadmittedWorkloadsObservabilityExplicitStatus.
-		if cond == nil {
+		// Stamping the condition on creation is gated by UnadmittedWorkloadsExplicitStatus.
+		if !features.Enabled(features.UnadmittedWorkloadsExplicitStatus) && cond == nil {
 			return "", "", nil
 		}
```
mimowo: Nagation if "if" is notoriously tricky to read - could we revert the order?

pkg/scheduler/scheduler_test.go
```diff
@@ -8544,27 +8578,41 @@ func TestSchedulerWhenWorkloadModifiedConcurrently(t *testing.T) {
 
 	scenarios := []map[featuregate.Feature]bool{
 		{
-			features.WorkloadRequestUseMergePatch:     false,
-			features.UnadmittedWorkloadsObservability: false,
+			features.WorkloadRequestUseMergePatch:      false,
+			features.UnadmittedWorkloadsObservability:  false,
+			features.UnadmittedWorkloadsExplicitStatus: false,
```
mimowo: Do we really benefit from multiplying every test by that? I feel this is getting very time consuming for our unit testing with limited benefit. So, I'm wondering if we could consider enabling the "false" more selectively - only where we expect it matters, same for "UnadmittedWorkloadsObservability". For example the FGs will not matter if the workload ends up admitted.

It might be considered for a follow up cleanup, but I think we are multiplying too much, this is already 2^3, cc @kshalot, wdyt?

pkg/controller/core/workload_controller.go
j-skiba: yeah, it makes sense. I replaced order here and in admissionchecks.go too

pkg/scheduler/scheduler_test.go
j-skiba: I though about this as a defensive approach but I agree it makes the tests longer too.

I removed that, I think the added unit tests (and the integration test) cover that anyway so it should be safe.

mimowo: Ok. to please just run the unit tests for scheduler to ensure all the newly added code paths are covered (there is go tooling for that)

j-skiba: the newly added code here is 100% covered

j-skiba:
/retest

```
Failed to read log: failed to read log "build-log.txt": error getting pod log size: error getting size of pod log: pods "9b0a984e-5b86-49c1-817f-59af0983ee31" not found
```

j-skiba:
/retest

```
Failed to read log: failed to read log "build-log.txt": error getting pod log size: error getting size of pod log: pods "9b0a984e-5b86-49c1-817f-59af0983ee31" not found
```

j-skiba:
/retest

```
Failed to read log: failed to read log "build-log.txt": error getting pod log size: error getting size of pod log: pods "9b0a984e-5b86-49c1-817f-59af0983ee31" not found
```

pkg/controller/core/workload_controller.go
```diff
@@ -1868,9 +1868,8 @@ func (r *WorkloadReconciler) resolveGranularUnadmittedQuotaReservedCondition(
 		cond.Reason != kueue.WorkloadWaiting: //nolint:staticcheck // SA1019: legacy reason
 		return cond.Reason, cond.Message, nil
 	default:
-		// Stamping the condition on creation is deferred to the next feature gate:
-		// UnadmittedWorkloadsObservabilityExplicitStatus.
-		if cond == nil {
+		// Stamping the condition on creation is gated by UnadmittedWorkloadsExplicitStatus.
+		if cond == nil && !features.Enabled(features.UnadmittedWorkloadsExplicitStatus) {
```
mimowo, line 1887: I think this requires a bit of cleanup, because:
1. the check `cond.Status == metav1.ConditionFalse` is unnecesaey, observing true is impossible I think, because of the outer check 
2. the chained cond.Reason are hard to read, and I'm not sure why we need that, and I sense some bug here. For example when `cond.Reason=Deactivated`, then we skip `case cond != nil && cond.Status == metav1.ConditionFalse && cond.Reason != "" &&`, so go into "default", then we set  `WorkloadQuotaReservedReasonPendingEvaluation` which might be wrong.
3. the `features.Enabled(features.UnadmittedWorkloadsExplicitStatus) ` checks inside the function are no-op because of the outer check which already checks features.Enabled(features.UnadmittedWorkloadsExplicitStatus) : https://github.com/kubernetes-sigs/kueue/pull/12719/changes#diff-60dd240c20adbd6a189d018d1c216c2d296730f446c341d8bf449fa6657964ffL1803

Maybe the code is somehow correct, but I cannot wrap my head around it with 10min of reading, which is not a good sign.

I imagine the structure should be:

```golang
switch {
case admissibilityErr != nil:
// ok
case workload.HasAdmissionGate(wl):
// ok
case cond != nil:
  return cond
default: // cond == nil
  return PendingEvaluation
```
if this check 
```
	case cond != nil && cond.Status == metav1.ConditionFalse && cond.Reason != "" &&
		cond.Reason != kueue.WorkloadQuotaReservedReasonSuspended &&
		cond.Reason != kueue.WorkloadQuotaReservedReasonMisconfigured &&
		cond.Reason != kueue.WorkloadDeactivated &&
		// Exclude legacy reasons so we can transition from them to more granular reasons
		// when the UnadmittedWorkloadsObservability feature gate is enabled.
		cond.Reason != kueue.WorkloadPending && //nolint:staticcheck // SA1019: legacy reason
		cond.Reason != kueue.WorkloadWaiting: //nolint:staticcheck // SA1019: legacy reason
```
is needed , then please add some comment why, and think about creating a named set of the reasons, like `stickyReasons`, so that we can better express the intention.

I know this comment should likely be to the previous PR, but it  missed my attention. Let's try to cleanup this now before we build on this more.

j-skiba: I agree, it can be improved for readability. I updated this list of checks into

```go
case cond != nil && cond.Reason != "" && schedulerSetReasons.Has(cond.Reason):
		// Preserve scheduler feedback reasons until the next scheduler cycle.
		return cond.Reason, cond.Message, nil
```

and removed negation too by changing the set of reasons it checks against.

I also removed the `== False` check, it was redundant.

--------

The `Deactivated` reason is covered for both cases:
- `Deactivated` -> `PendingEvaluation` - `workload that is reactivated should clear Deactivated condition and transition to PendingEvaluation (observability enabled)` unit test case
- `any` -> `Deactivated` - https://github.com/kubernetes-sigs/kueue/pull/12719/changes#diff-60dd240c20adbd6a189d018d1c216c2d296730f446c341d8bf449fa6657964ffL1830 it's a few lines above, also covered in `workload that is deactivated should set QuotaReserved condition to Deactivated (observability enabled)` and `should set status QuotaReserved conditions to False with reason Deactivated if workload is deactivated even if LocalQueue doesn't exist`

--------

> the features.Enabled(features.UnadmittedWorkloadsExplicitStatus)  checks inside the function are no-op because of the outer check which already checks features.Enabled(features.UnadmittedWorkloadsExplicitStatus) : https://github.com/kubernetes-sigs/kueue/pull/12719/changes#diff-60dd240c20adbd6a189d018d1c216c2d296730f446c341d8bf449fa6657964ffL1803

the check at 1803 checks `UnadmittedWorkloadsObservability` and the check here is for `UnadmittedWorkloadsExplicitStatus`, I think it should stay as is

kshalot (commented):
This looks good to me, no comments :shipit: 

/lgtm

mimowo (commented):
/lgtm
/approve
Thank you for this ongoing work, and making the PRs digestible for reviews 👍

PR #12723, opened 2026-07-02
Switch Python examples from pip to uv with PEP 723 inline metadata

kannon92:
#### What type of PR is this?
/kind documentation

- Replace shebang with #!/usr/bin/env -S uv run
- Add PEP 723 inline script metadata declaring dependencies
- Move config.load_kube_config() from module-level into main()
- Update docs (en + zh-CN) to use uv run instead of python/pip
- Delete requirements.txt (deps now inline in each script)
- Add Running section to examples README

Fixes #

```release-note
NONE
```

## Summary by CodeRabbit

* **Documentation**
  * Updated Python job-run guides to use `uv run` for all example commands, removing prior manual virtual environment and `pip install` steps.
  * Refreshed Flux Operator instructions so dependency setup is handled automatically when running the script.
* **New Features**
  * The included Python examples are now runnable directly via `uv run`, with inline dependency metadata enabling automatic environment setup.
* **Chores**
  * Improved example script execution behavior by deferring Kubernetes client setup until runtime.

### DIFF
--- a/site/static/examples/python/install-kueue-queues.py
+++ b/site/static/examples/python/install-kueue-queues.py
@@ -1,4 +1,11 @@
-#!/usr/bin/env python3
+#!/usr/bin/env -S uv run --script
+# /// script
+# requires-python = ">=3.9"
+# dependencies = [
+#     "kubernetes",
+#     "requests",
+# ]
+# ///
 
 from kubernetes import utils, config, client
 import tempfile
@@ -10,9 +17,6 @@ import argparse
 # This example will demonstrate installing Kueue and applying a YAML file (local) to install Kueue
 
 # Make sure your cluster is running!
-config.load_kube_config()
-crd_api = client.CustomObjectsApi()
-api_client = crd_api.api_client
 
 
 def get_parser():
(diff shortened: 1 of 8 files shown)


site/static/examples/python/install-kueue-queues.py
```diff
@@ -1,4 +1,11 @@
-#!/usr/bin/env python3
+#!/usr/bin/env -S uv run
```
mimowo: The documentation of uv proposes uv run --script:

https://docs.astral.sh/uv/guides/scripts/?utm_source=chatgpt.com#using-a-shebang-to-create-an-executable-file

should we rather use that?

mimowo:
cc @amy looks good to me, any thoughts?

site/static/examples/python/install-kueue-queues.py
kannon92: https://github.com/kubernetes-sigs/kueue/pull/12723#pullrequestreview-4617890915

Yea I'll fix that.

kannon92: done.

mimowo:
/lgtm
/approve
/cherrypick website
Let's go with the proposal, I tested "uv" locally (first time) and it works nicely indeed 👍

Issue #2227, opened 2024-05-17
Jobsets having replicas = 0 for their replicatedJob are not handled in Kueue


shrinandj:
**What happened**:
- Created a jobset with two replicatedJobs: the first one had a replica count of 1 and the other one had a replica count of 0.
- The jobset was created in a suspended state and had the required kueue labels/annotations
- The jobset was NOT unsuspended by Kueue and the 1 expected job did not start

**What you expected to happen**:
- Kueue should have unsuspended the jobset and started the job

**How to reproduce it (as minimally and precisely as possible)**:
- As mentioned above, create a jobset was two replicatedJobs: one with replicas > 0 and another with replicas = 0

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`): `v1.28.8`
- Kueue version (use `git describe --tags --dirty --always`):`v0.6`
- Cloud provider or hardware configuration: `aws`
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

** Details **

Jobset CRD shows the default is 1 but it is not a minimum value of 1. So 0 is acceptable.
```
                    replicas:
                      default: 1
                      description: |-
                        Replicas is the number of jobs that will be created from this ReplicatedJob's template.
                        Jobs names will be in the format: <jobSet.name>-<spec.replicatedJob.name>-<job-index>
                      format: int32
                      type: integer
                    template:
```

Kueue's workload CRD shows that the podSet's count has a minimum value of 1:
```
                  properties:
                    count:
                      description: count is the number of pods for the spec.
                      format: int32
                      minimum: 1
                      type: integer
                    minCount:
                      description: |-
                        minCount is the minimum number of pods for the spec acceptable
                        if the workload supports partial admission.


                        If not provided, partial admission for the current PodSet is not
                        enabled.


                        Only one podSet within the workload can use this.


                        This is an alpha field and requires enabling PartialAdmission feature gate.
                      format: int32
                      minimum: 1
                      type: integer
```

Kueue's log show the following error:
```
[...]

PR #12728, opened 2026-07-02
TAS Scheduler PodSetSliceSize division-by-zero hardening

mszadkow:
#### What type of PR is this?
/kind bug
/area tas

This PR hardens TAS slice-size validation and aligns scheduler behavior with stricter invariants:

* Enforces podSetSliceSize validation in workload admission for TAS slice requests.
* Requires podSetSliceSize when podSetSliceRequiredTopology is set.
* Rejects non-positive slice sizes for both legacy slice fields and topology constraint layers.

This PR is specifically for cherry-picks as mentioned in the [comment](https://github.com/kubernetes-sigs/kueue/pull/12728#issuecomment-5103740536).
The strict API validation will be covered in [this PR](https://github.com/kubernetes-sigs/kueue/pull/13634).

```release-note
TAS: Enforce stricter slice-size validation for Workloads. When podSetSliceRequiredTopology is specified, podSetSliceSize must also be specified and must be greater than 0. Non-positive slice sizes in topology constraints are also rejected.

ACTION REQUIRED: If you create Workload objects directly (or via custom controllers), update manifests before upgrade so that:

* podSetSliceRequiredTopology is never set without podSetSliceSize
* podSetSliceSize is always greater than 0
* podSetSliceSize is not set when podSetSliceRequiredTopology is absent
* every podsetSliceRequiredTopologyConstraints entry has size greater than 0

If you need a phased rollout, temporarily disable TASValidateWorkloadSliceSize, clean up invalid Workloads, then re-enable it.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened workload and topology validation so slice-size values must be at least `1` across CRDs and request schemas.
  * Improved slice-size resolution: nil/empty topology requests now default to `1`, and slice topology now reports clearer errors when slice size is missing or non-positive.
* **Tests**
  * Added unit tests covering defaulting and validation for slice sizing, including updated error checks and assertions.

### DIFF
--- a/pkg/features/kube_features.go
+++ b/pkg/features/kube_features.go
@@ -514,6 +514,11 @@ const (
 	//
 	// Rejects Workloads with negative container or pod-level resource requests/limits.
 	WorkloadValidateResourcesAreNonNegative featuregate.Feature = "WorkloadValidateResourcesAreNonNegative"
+
+	// owner: mszadkow
+	//
+	// Rejects Workloads with non-positive TAS slice sizes in PodSet topology requests.
+	TASValidateWorkloadSliceSize featuregate.Feature = "TASValidateWorkloadSliceSize"
 )
 
 func init() {
@@ -789,6 +794,10 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 
 	WorkloadValidateResourcesAreNonNegative: {
 		{Version: version.MustParse("0.20"), Default: true, PreRelease: featuregate.Beta},
+	},
+
+	TASValidateWorkloadSliceSize: {
+		{Version: version.MustParse("0.20"), Default: true, PreRelease: featuregate.Beta}, // GA in 0.21
 	},
 }
 
--- a/test/integration/singlecluster/tas/tas_test.go
+++ b/test/integration/singlecluster/tas/tas_test.go
@@ -3746,7 +3746,8 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
 							*utiltestingapi.MakePodSet("leader", 1).
 								PodSetGroup("group").
 								PreferredTopologyRequest(utiltesting.DefaultRackTopologyLevel).
-								SliceRequiredTopologyRequest(corev1.LabelHostname).
+								// Leader slice settings are not required for grouped leader-worker placement.
+								// Workers drive slice topology, while leader is handled as a reserved slot.
 								Request("nvidia.com/gpu", "1").
 								Obj(),
 							*utiltestingapi.MakePodSet("worker", 30).
--- a/pkg/webhooks/workload_webhook.go
+++ b/pkg/webhooks/workload_webhook.go
@@ -172,6 +172,10 @@ func validatePodSet(ps *kueue.PodSet, path *field.Path) field.ErrorList {
 		allErrs = append(allErrs, validateResourceList(ps.Template.Spec.Resources.Limits, resPath.Child("limits"))...)
 	}
 
+	if features.Enabled(features.TASValidateWorkloadSliceSize) {
+		allErrs = append(allErrs, validateTASSliceSize(ps.TopologyRequest, path.Child("topologyRequest"))...)
+	}
+
 	return allErrs
 }
 
@@ -466,5 +470,38 @@ func validateClusterNameUpdate(newObj, oldObj *kueue.Workload, statusPath *field
 		}
 	}
 
+	return allErrs
+}
+
+// validateTASSliceSize enforces basic topology-slice invariants:
+// - legacy fields must appear together (required-topology <-> slice-size)
+// - slice sizes must be strictly positive (legacy and multi-layer constraints)
+// Kept in the Workload webhook as defense-in-depth for direct Workload writes,
+// including cases where CRDs or producer-side validators are older or bypassed.
+func validateTASSliceSize(tr *kueue.PodSetTopologyRequest, path *field.Path) field.ErrorList {
+	if tr == nil {
+		return nil
+	}
+
+	var allErrs field.ErrorList
+	if tr.PodSetSliceRequiredTopology != nil {
+		switch {
+		case tr.PodSetSliceSize == nil:
+			allErrs = append(allErrs, field.Required(path.Child("podSetSliceSize"), "must be set when podSetSliceRequiredTopology is specified"))
+		case *tr.PodSetSliceSize <= 0:
+			allErrs = append(allErrs, field.Invalid(path.Child("podSetSliceSize"), *tr.PodSetSliceSize, "must be greater than 0"))
+		}
+	}
+
+	if tr.PodSetSliceRequiredTopology == nil && tr.PodSetSliceSize != nil {
+		allErrs = append(allErrs, field.Forbidden(path.Child("podSetSliceSize"), "may not be set when podSetSliceRequiredTopology is not specified"))
+	}
+
+	for i := range tr.PodsetSliceRequiredTopologyConstraints {
+		if tr.PodsetSliceRequiredTopologyConstraints[i].Size <= 0 {
+			allErrs = append(allErrs,
+				field.Invalid(path.Child("podsetSliceRequiredTopologyConstraints").Index(i).Child("size"), tr.PodsetSliceRequiredTopologyConstraints[i].Size, "must be greater than 0"))
+		}
+	}
 	return allErrs
 }
--- a/pkg/webhooks/workload_webhook_test.go
+++ b/pkg/webhooks/workload_webhook_test.go
@@ -255,6 +255,57 @@ func TestValidateWorkload(t *testing.T) {
 				Obj(),
 			wantErr: nil,
 		},
+		"should reject non-positive podSetSliceSize when TASValidateWorkloadSliceSize is enabled": {
+			workload: utiltestingapi.MakeWorkload(testWorkloadName, testWorkloadNamespace).
+				PodSets(
+					*utiltestingapi.MakePodSet("bad", 1).
+						SliceRequiredTopologyRequest("kubernetes.io/hostname").
+						SliceSizeTopologyRequest(0).
+						Obj(),
+				).
+				Obj(),
+			wantErr: field.ErrorList{
+				field.Invalid(podSetsPath.Index(0).Child("topologyRequest", "podSetSliceSize"), nil, ""),
+			}.ToAggregate(),
+		},
+		"should reject podSetSliceSize without podSetSliceRequiredTopology when TASValidateWorkloadSliceSize is enabled": {
+			workload: utiltestingapi.MakeWorkload(testWorkloadName, testWorkloadNamespace).
+				PodSets(
+					*utiltestingapi.MakePodSet("bad", 1).
+						SliceSizeTopologyRequest(1).
+						Obj(),
+				).
+				Obj(),
+			wantErr: field.ErrorList{
+				field.Forbidden(podSetsPath.Index(0).Child("topologyRequest", "podSetSliceSize"), ""),
+			}.ToAggregate(),
+		},
+		"should reject non-positive podsetSliceRequiredTopologyConstraints size when TASValidateWorkloadSliceSize is enabled": {
+			workload: utiltestingapi.MakeWorkload(testWorkloadName, testWorkloadNamespace).
+				PodSets(
+					*utiltestingapi.MakePodSet("bad", 1).
+						SliceRequiredTopologyConstraints(kueue.PodsetSliceRequiredTopologyConstraint{Topology: "kubernetes.io/hostname", Size: 0}).
+						Obj(),
+				).
+				Obj(),
+			wantErr: field.ErrorList{
+				field.Invalid(podSetsPath.Index(0).Child("topologyRequest", "podsetSliceRequiredTopologyConstraints").Index(0).Child("size"), nil, ""),
+			}.ToAggregate(),
+		},
+		"should accept non-positive podSetSliceSize when TASValidateWorkloadSliceSize is disabled": {
+			featureGates: map[featuregate.Feature]bool{
+				features.TASValidateWorkloadSliceSize: false,
+			},
+			workload: utiltestingapi.MakeWorkload(testWorkloadName, testWorkloadNamespace).
+				PodSets(
+					*utiltestingapi.MakePodSet("ok", 1).
+						SliceRequiredTopologyRequest("kubernetes.io/hostname").
+						SliceSizeTopologyRequest(0).
+						Obj(),
+				).
+				Obj(),
+			wantErr: nil,
+		},
 		"empty podSetUpdates": {
 			workload: utiltestingapi.MakeWorkload(testWorkloadName, testWorkloadNamespace).AdmissionChecks(kueue.AdmissionCheckState{}).Obj(),
 			wantErr:  nil,
--- a/site/data/featuregates/versioned_feature_list.yaml
+++ b/site/data/featuregates/versioned_feature_list.yaml
@@ -479,6 +479,12 @@
     lockToDefault: false
     preRelease: Alpha
     version: "0.18"
+- name: TASValidateWorkloadSliceSize
+  versionedSpecs:
+  - default: true
+    lockToDefault: false
+    preRelease: Beta
+    version: "0.20"
 - name: TLSOptions
   versionedSpecs:
   - default: true
--- a/test/compatibility_lifecycle/reference/versioned_feature_list.yaml
+++ b/test/compatibility_lifecycle/reference/versioned_feature_list.yaml
@@ -479,6 +479,12 @@
     lockToDefault: false
     preRelease: Alpha
     version: "0.18"
+- name: TASValidateWorkloadSliceSize
+  versionedSpecs:
+  - default: true
+    lockToDefault: false
+    preRelease: Beta
+    version: "0.20"
 - name: TLSOptions
   versionedSpecs:
   - default: true


apis/kueue/v1beta2/workload_types.go
```diff
@@ -231,6 +231,7 @@ type PodSetTopologyRequest struct {
 	// in `kueue.x-k8s.io/podset-slice-required-topology` annotation.
 	//
 	// +optional
+	// +kubebuilder:validation:Minimum=1
```
mwielgus: What will happen if a workload with PodSetSliceSize = 0 is in the system at the upgrade time?

mszadkow: At upgrade time, pre-existing Workloads are not re-checked by CRD create-time validation.
So if a Workload already stored in the cluster has PodSetSliceSize = 0, it can still exist after upgrade.
The new checks matter when Kueue has to admit it again, such as on changes that trigger a new admission decision

dkaluza: But what about workloads that really have podsetslice size 0?
We allowed jobsets with replicas=0. 
https://github.com/kubernetes-sigs/kueue/issues/2227

IIUC making this required to be minimum 1 would lead to serious regression without additional handling.

mszadkow: Ok, so for now I will provide additional tests that will prove if the validation is backward compatible or not.

mszadkow: I guess we want to allow for  zero replicas jobs still, right?
So the actual fix must be shifted from validation to making sure there is no division by zero caused by that.
@dkaluza thanks for insight

dkaluza: From my point of view yes - we want to allow 0 replicas in jobs. 

Correct, shifting to just avoiding divisions by 0 in the implementation is one of valid approaches.

mszadkow:
OK, added validation in `getSliceSizeWithSinglePodAsDefault()` that:
* rejects negative values with error
* converts zero to one with warning to avoid panics, which is mathematically safe

The issue was both proved and solved with extra integration test, which I am not sure is particularly required in this PR.
However I added couple unit tests to fully cover `getSliceSizeWithSinglePodAsDefault()` function.

apis/kueue/v1beta2/workload_types.go
mszadkow: Updated - https://github.com/kubernetes-sigs/kueue/pull/12728#issuecomment-4914391766

pkg/cache/scheduler/tas_flavor_snapshot.go
ivnovakov: There are 2 calls of this function that ignore reason and  just use the size value.
So, negative `PodSetSliceSize` now returns `(0, "cannot be negative...")`.
Lines 671-672 and 744-745 drop that reason and go straight to `tr.Count % 0` that leads to panic.
Before this PR negative values just passed through harmlessly; now they crash.
Only line 885 (`findTopologyAssignment`) checks `reason` first.

This is on the unhealthy-node/TAS-replacement path, and there's no `recover()` in the scheduler, so it takes down `kueue-controller-manager`, not just the workload.

Should we allow that panic or should we handle that gracefully?

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1232,7 +1232,17 @@ func getSliceSizeWithSinglePodAsDefault(podSetTopologyRequest *kueue.PodSetTopol
 		return 0, "slice topology requested, but slice size not provided"
 	}
 
-	return *podSetTopologyRequest.PodSetSliceSize, ""
+	sliceSize := *podSetTopologyRequest.PodSetSliceSize
```
mbobrovskyi: I'm wondering if we want to add validation on API.

mszadkow: https://github.com/kubernetes-sigs/kueue/pull/12728#discussion_r3518492227

pkg/cache/scheduler/tas_flavor_snapshot.go
mszadkow: Good catch!
Thanks I will look into it

mszadkow: check again

ivnovakov: Fix is valid, thanks!

pkg/cache/scheduler/tas_flavor_snapshot.go
mbobrovskyi: I see. It's possible to have PodSetSliceSize=0. But can we add validation to ensure that it cannot be set to a value less than 0?

mszadkow: Added, this should be neutral for existing workloads
@dkaluza wdyt?

dkaluza: IMO validation that it is not negative is reasonable.

pkg/util/tas/tas.go
```diff
@@ -126,6 +126,11 @@ func PodSetSliceRequiredTopologyConstraints(tr *kueue.PodSetTopologyRequest) []k
 	size := int32(0)
 	if tr.PodSetSliceSize != nil {
 		size = *tr.PodSetSliceSize
+		if size == 0 {
+			// While we allow replicas equal to 0 per #2227, we must prevent sliceSize=0 to avoid division/modulo errors.
```
dkaluza: Doesn't this lead to incorrect computation/assumptions about scheduling and possible admissions in some places?

E.g.  There is maximum of pods per node https://docs.cloud.google.com/kubernetes-engine/docs/how-to/flexible-pod-cidr#configuring_maximum_pods_per_node
If we assume that 0 replicas still "consume" one pod we might get into the state that Kueue might not admit jobs although it could. 

Generally it would be less tricky to really guard the divisions size by returning explicitly desired value in those places.

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -666,7 +666,10 @@ func (s *TASFlavorSnapshot) findReplacementAssignment(
 	if isStale, staleDomain := s.IsTopologyAssignmentStale(existingAssignment); isStale {
 		return nil, nil, fmt.Sprintf("Cannot replace the node, because the existing topologyAssignment is invalid, as it contains the stale domain %v", staleDomain)
 	}
-	requiredReplacementDomain := s.requiredReplacementDomain(tr, existingAssignment)
+	requiredReplacementDomain, reason := s.requiredReplacementDomain(tr, existingAssignment)
+	if reason != "" {
+		return nil, nil, reason
+	}
```
dkaluza: why do we need this change of  `s.requiredReplacementDomain` signature?

Aren't we anyway calling `getSliceSizeWithSinglePodAsDefault(tr.PodSet.TopologyRequest)` in the next line returning reason?

pkg/cache/scheduler/tas_flavor_snapshot_test.go
```diff
@@ -940,3 +940,156 @@ func TestTruncateAssignment(t *testing.T) {
...
+
+func TestRequiredReplacementDomain(t *testing.T) {
+	tests := map[string]struct {
+		name       string
+		request    *kueue.PodSetTopologyRequest
+		ta         *tas.TopologyAssignment
+		wantDomain tas.TopologyDomainID
+		wantReason string
+	}{
+		"negative slice size returns reason": {
```
dkaluza: Don't we lack  tests with PodSetSliceSize 0?

pkg/cache/scheduler/tas_flavor_snapshot_test.go
```diff
@@ -940,3 +940,156 @@ func TestTruncateAssignment(t *testing.T) {
...
+		},
+		"slice size nil - should return 0 with error": {
+			request: &kueue.PodSetTopologyRequest{
+				PodSetSliceRequiredTopology: new(string("kubernetes.io/zone")),
+				PodSetSliceSize:             nil,
+			},
+			wantSize:   0,
+			wantReason: "slice topology requested, but slice size not provided",
+		},
+		"zero slice size - should convert to 1 (division by zero protection)": {
```
dkaluza: From my perspective this is not an intuitive behavior for this function.

We should at least describe it in the function documentation. Even better would be to either make this an explicit conversion in the division place (mentioned in other comment) 
Also I do not see the change in implementation of this function, am I missing something?

pkg/cache/scheduler/tas_flavor_snapshot_test.go
dkaluza: Also there are no  tests that really return any domain id...

Anyway `requiredReplacementDomain` looks like a local helper that may easily get removed or replaced in the future implementation refactors.
 It would be best to add unit tests for actual business logic in some higher level function like `FindTopologyAssignmentsForFlavor`, to make sure that we actually get desired behavior from the business perspective (for example they are no longer panicking but returning desired value)

This `requiredReplacementDomain` actually never did division by 0 so introduction of those test doesn't yet prove that we have fixed the bug.

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1222,6 +1225,7 @@ func getSliceSizeWithSinglePodAsDefault(tr *kueue.PodSetTopologyRequest) (int32,
 	if size <= 0 {
 		return 0, "slice topology requested, but slice size not provided"
 	}
+
```
mimowo: nit, drop formatting changes

pkg/util/tas/tas.go
```diff
@@ -126,6 +126,11 @@ func PodSetSliceRequiredTopologyConstraints(tr *kueue.PodSetTopologyRequest) []k
...
+			// Convert 0 to 1: mathematically safe (x/1=x, x%1=0) and 1 pod per slice is valid scheduling semantics.
```
mimowo: Add a logger at V2 this should not really happen

apis/kueue/v1beta1/workload_types.go
```diff
@@ -148,6 +148,7 @@ type PodSetTopologyRequest struct {
 	// in `kueue.x-k8s.io/podset-slice-required-topology` annotation.
 	//
 	// +optional
+	// +kubebuilder:validation:Minimum=0
```
mimowo: I would suggest to drop schema changes from a PR which needs to be cherry-picked. We can consider as a follow up

pkg/cache/scheduler/tas_flavor_snapshot.go
mszadkow: As suggested by @mimowo will be provided as separate PR

pkg/util/tas/tas.go
mszadkow: Thanks! I got you, the consequences are hard to anticipate here, but we should avoid them if possible.
I will adjust

pkg/cache/scheduler/tas_flavor_snapshot.go
mszadkow: I think we could actually reverse the order of calling those 2 function and pass sliceSize to `s.requiredReplacementDomain` no need to check the reason

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1979,6 +1996,9 @@ func (s *TASFlavorSnapshot) notFitMessage(slicesFitCount, totalRequestsSlicesCou
 
 func countSlicesInSubtree(d *domain, currentLevel, targetLevel int, sliceSize int32) int32 {
 	if currentLevel == targetLevel {
+		if sliceSize <= 0 {
```
mimowo: I can see this code`if sliceSize <= 0 {` repeated many times. Are all these places really reachable by tests, or one guard at the beginning would be enough? I would prefer a single guard and don't bother in downstream code. Let's start by checking if all places are covered with tests.

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1387,7 +1388,10 @@ func (s *TASFlavorSnapshot) findLevelWithFitDomains(
 	sortedDomain := s.sortedDomainsWithLeader(levelDomains, state.unconstrained)
 	topDomain := sortedDomain[0]
 
-	sliceCount := state.count / state.sliceSize
+	sliceCount := int32(0)
+	if state.sliceSize > 0 {
+		sliceCount = state.count / state.sliceSize
+	}
```
mimowo: Hm, this silent handling makes me anxious it may hit issues later, why not return error?

mimowo (commented):
I looked at the code, but I'm lost seeing the checks scattered across. Some return errors, some silently handle, looks like a mess tbh. 

Ideally, we should have a single guard check at the beginning, or maybe in two places if on different layers, and return errors. 

I might be missing some scenarios, but please investiage if we could do fail fast

pkg/cache/scheduler/tas_flavor_snapshot.go
dkaluza: I agree with mimowo here. It would be best to handle this in less number of places - maybe we can track the actual usage of functions here from the public API. And just handle this once on the higher level?

Maybe it is enough to have handling in  `findReplacementAssignment` and `findTopologyAssignment` and return appropriate error for those?

mszadkow: We began with the assumption that 0 as podSetSlice is allowed, due to possibility of `replicas=0`.

We return the reason if the podSetSliceSize is 0 [here already](https://github.com/kubernetes-sigs/kueue/pull/12880/changes) in `getSliceSizeWithSinglePodAsDefault`.

It seems to me that simply checking the reason of `getSliceSizeWithSinglePodAsDefault` in `findReplacementAssignment` and `requiredReplacementDomain` would be enough to prevent divisions by zero.
I think something similar was proposed at [some version of this PR](https://github.com/kubernetes-sigs/kueue/commit/3d1c2ccdaf5ef44a02f311b9449be6525844f817).

dkaluza: The linked one changed `getSliceSizeWithSinglePodAsDefault`, always returning an error on size 0. I agree that this is somehow similar, but in this way we are:
A) Redefining a function that is supposed to return one as default. Personally I would interpret its name as it is returning 1 if something fails, it is not intuitive that it does some validation and returns errors.
B) We were always returning error in this case - which might be wrong. What is the expected behavior for 0 requires careful investigation here, what is the semantics for those higher level functions and how they should behave for 0 replicas.

How about instead of forcing this check to the `getSliceSizeWithSinglePodAsDefault` as this one is a low level helper add the check to mentioned 2 functions and return explicitly "early" what is desired there. It may be error but it may be also some other value, hard to tell without going deeper in code.

mszadkow: Got you!
1. I will rename this function ;)
2. @dkaluza wdyt of plan like that:

We will keep `getSliceSizeWithSinglePodAsDefault` strict for real scheduling work, so slice size <= 0 remains invalid when pods must be placed.
We will add early returns in `findTopologyAssignment` and `findReplacementAssignment` for Count == 0, so zero-replica and zero-replacement paths succeed as no-ops instead of failing slice-size validation.
This preserves legacy/new slice-constraint normalization in `PodSetSliceRequiredTopologyConstraints` while moving zero-count semantics to the higher-level functions that actually know scheduling intent.
We will add focused tests to prove two behaviors: positive counts still fail on invalid slice size, and zero counts return successful empty assignments.

mszadkow: Documentation check:
KEP defines TopologyAssignment as per-domain pod counts, which implies no domain entries when there are no pods to place: [README.md](https://github.com/kubernetes-sigs/kueue/blob/4ce0b760f9001bdacf63acceb4599299bff203fb/keps/2724-topology-aware-scheduling/README.md?plain=1#L1108) and [README.md](https://github.com/kubernetes-sigs/kueue/blob/4ce0b760f9001bdacf63acceb4599299bff203fb/keps/2724-topology-aware-scheduling/README.md?plain=1#1149).

A planned implementation will explicitly encode count-0 as a direct success path.
There is no failure reason, so assignment shape is Levels present + empty Domains, which means “nothing to place.”

mszadkow:
Updated the PR description of the latest changes requested in  - https://github.com/kubernetes-sigs/kueue/pull/12728#discussion_r3628885543 and https://github.com/kubernetes-sigs/kueue/pull/12728#pullrequestreview-4745501163
ptal @mbobrovskyi @reruno

pkg/cache/scheduler/tas_flavor_snapshot_test.go
```diff
@@ -1064,3 +1064,333 @@ func TestTASCachingRemainingResourcesFeatureGate(t *testing.T) {
...
+		wantReason      string
+		wantAssignment  *tas.TopologyAssignment
+	}{
+		"legacy nil slice size": {
+			topologyRequest: &kueue.PodSetTopologyRequest{
+				PodSetSliceRequiredTopology: new(string(corev1.LabelHostname)),
+			},
+			count:      2,
+			wantReason: "slice topology requested, but slice size not provided",
+		},
```
dkaluza: nit: consider defining constant for this reason instead of repeating it in many places.

pkg/cache/scheduler/tas_flavor_snapshot_test.go
```diff
@@ -1064,3 +1064,333 @@ func TestTASCachingRemainingResourcesFeatureGate(t *testing.T) {
...
 }
+
+func TestFindTopologyAssignmentsForFlavorSliceSize(t *testing.T) {
+	tests := map[string]struct {
+		topologyRequest *kueue.PodSetTopologyRequest
+		count           int32
+		wantReason      string
+		wantAssignment  *tas.TopologyAssignment
+	}{
+		"legacy nil slice size": {
```
dkaluza: I'm not sure if legacy is a good word here - AFAIK the constraints are in alpha

pkg/cache/scheduler/tas_flavor_snapshot_test.go
```diff
@@ -1064,3 +1064,333 @@ func TestTASCachingRemainingResourcesFeatureGate(t *testing.T) {
...
+		"constraints zero slice size": {
+			topologyRequest: &kueue.PodSetTopologyRequest{
+				PodsetSliceRequiredTopologyConstraints: []kueue.PodsetSliceRequiredTopologyConstraint{{
+					Topology: corev1.LabelHostname,
+					Size:     0,
+				}},
+			},
+			count:      2,
+			wantReason: "slice topology requested, but slice size not provided",
+		},
```
dkaluza: Can we have one with constraint size 0 and count 0?

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -972,10 +973,23 @@ func (s *TASFlavorSnapshot) findTopologyAssignment(
...
 	}
+	// Zero-pod requests are successful no-ops. We still keep topology levels to
+	// preserve the assignment contract shape for downstream consumers,
+	// while skipping slice validation and capacity/fit computations.
+	if workersTasPodSetRequests.Count == 0 {
+		assignments := map[kueue.PodSetReference]*utiltas.TopologyAssignment{
+			workersTasPodSetRequests.PodSet.Name: s.buildAssignment(nil),
+		}
+		if leaderTasPodSetRequests != nil {
+			assignments[leaderTasPodSetRequests.PodSet.Name] = s.buildAssignment(nil)
```
dkaluza: Is this a correct behavior?

Cannot I have a leader and 0 count worker and they all require sepecific topology?

I guess since I have a request of 1 pod for leader it should be assigned appropriately.
Probably we are also missing test for that.

pkg/cache/scheduler/tas_flavor_snapshot_test.go
mszadkow: thanks, I will find better word

pkg/cache/scheduler/tas_flavor_snapshot.go
mszadkow: I thought of this too and checked that:
The pairing logic in [tas_flavor_snapshot.go:725](https://github.com/kubernetes-sigs/kueue/blob/bcf7ae5e37cb01d28a0f355f68a588dcf2f234e3/pkg/cache/scheduler/tas_flavor_snapshot.go#L725) always keeps workers as the larger-or-equal count and leader as the smaller count, so a positive count is assigned to workers.

In the standard flow, scheduler first normalizes the pair by choosing the higher-count podset as workers and the other as leader, so workers.Count is always greater than or equal to leader.Count. That means a raw state like workers=0 and leader>0 does not propagate to assignment logic.
It gets reordered before findTopologyAssignment runs. The zero-count no-op path is therefore only reached when the effective workers count is actually zero (and leader is zero or absent).

pkg/cache/scheduler/tas_flavor_snapshot_test.go
mszadkow: sure

pkg/cache/scheduler/tas_flavor_snapshot.go
dkaluza: Thanks for checking, maybe it is worth a comment in code, that this part is based on this assumption?
This will make debugging and code reading easier if something diverges.

dkaluza (approved):
@mimowo this one looks good to me in the current form

pkg/cache/scheduler/tas_flavor_snapshot_test.go
```diff
@@ -1064,3 +1064,357 @@ func TestTASCachingRemainingResourcesFeatureGate(t *testing.T) {
 		})
 	}
 }
+
+func TestFindTopologyAssignmentsForFlavorSliceSize(t *testing.T) {
```
mimowo: Would it be feasible to move the tests to tas_cache_test.go? At least some of them. That would show that the code was reachable. If we make tests only for the internal helpers then it is unclear if the code is actually reachable or dead - only covered by low level testing.

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -972,10 +975,23 @@ func (s *TASFlavorSnapshot) findTopologyAssignment(
 			return nil, fmt.Sprintf("invalid podSetUpdate for PodSet %s, error: %s", workersTasPodSetRequests.PodSet.Name, err.Error())
 		}
 	}
+	// Zero-pod requests are successful no-ops. We still keep topology levels to
+	// preserve the assignment contract shape for downstream consumers,
+	// while skipping slice validation and capacity/fit computations.
+	if workersTasPodSetRequests.Count == 0 {
+		assignments := map[kueue.PodSetReference]*utiltas.TopologyAssignment{
```
mimowo: This sounds like an anomaly, let's add V2 logging

mimowo:
> TAS: Validate slice size across all topology constraint layers and treat zero-pod TAS assignment paths as successful no-ops with consistent empty assignments.```

Please formulate the release note to reflect the "user-observable change". I'm still puzzled what are the user-observable changes of the PR, let me ask some clarifying questions:

1. this PR does not introduce a rejecting validation, right? If I was a user this is how I would understand "TAS: Validate slice size across all topology constraint layers". IIUC this PR is a mix of fixing handling when PodSetSlice=0 is legal and internal failing avoiding crashing in some other cases.
2. do we have rejection-validation for PodSetSliceSize = 0, to fail fast in webhooks, was this considered?
3. it seems in some cases the PR rejects PodSetSliceSize=0, and in some cases if handles them as correct.  Could you please consider splitting the PR to make the different scenarios easier to review?

mszadkow:
1. Yes, your understanding is correct: this PR does not introduce a new rejecting validation.
For positive-count requests, invalid slice-size inputs now fail more deterministically during TAS assignment.
For count=0 paths, it treats them as legal no-ops instead of allowing them to reach "division by zero" code.

2. Yes, it can be introduced, but doing it as an immediate hard rejection can break some existing users.
We avoided that change due to this [discussion](https://github.com/kubernetes-sigs/kueue/pull/12728#discussion_r3518492227)

3. Yes, your concern is valid, and the behavior split is intentional but subtle.
Slice size <= 0 with pods to place is treated as invalid and fails.
Slice size <= 0 with count = 0 is treated as a legal no-op and succeeds with an empty assignment.
That’s why it can look like “same input sometimes rejected, sometimes accepted” but the deciding factor is whether there is actual placement work.

Overall splitting the PR for easier review is a reasonable request.
I will make one PR for zero-count/no-op semantics and another one PR for strict invalid-slice handling on positive-count scheduling paths.
Does it sound right?

mimowo:
> Yes, your understanding is correct: this PR does not introduce a new rejecting validation.
For positive-count requests, invalid slice-size inputs now fail more deterministically during TAS assignment.
For count=0 paths, it treats them as legal no-ops instead of allowing them to reach "division by zero" code.

Yeah, the "legal no-ops" is concerning me. I cannot see any useful use of PodSetSliceSize=0 (see below)

> Yes, it can be introduced, but doing it as an immediate hard rejection can break some existing users.
We avoided that change due to this https://github.com/kubernetes-sigs/kueue/pull/12728#discussion_r3518492227

I missed that discussion, but I'm not sure if allowing replicas = 0 makes the argument. replicas = 0 makes sense for deleting all pods without deleting the Job itself - scale down to 0. For PodSetSliceSize annotation we don't really support elasticity, so I cannot see a legitimate use case.

So, I think it is fair to just reject at the creation time, but I wouldn't do that by the schema, but webhook code for the Jobs & workload. Analogously I would add checks inside TAS to fail with errors.

Doing this in code allows us to use FG to decomission the old code.

> Slice size <= 0 with pods to place is treated as invalid and fails.

That is correct.

> Slice size <= 0 with count = 0 is treated as a legal no-op and succeeds with an empty assignment.

That is an unnecessary complication IMO (see above).

> I will make one PR for zero-count/no-op semantics and another one PR for strict invalid-slice handling on positive-count scheduling paths.
> Does it sound right?

I would just introduce a FG like `TASRejectPodSetSliceSizeZero` and reject in TAS code, workload webhook and job webhooks. Let me confirm the approach change with @dkaluza, wdyt?

dkaluza:
@mimowo 
From my perspective we should either support 0 replicas and counts or not. We should be consistent in this decision, otherwise it will result in poor user experience and bugs in the future.
Especially we already had a bug regarding 0 replicas that we considered valid - https://github.com/kubernetes-sigs/kueue/issues/2227

If we are supporting 0 replicas for JobSet and therefore 0 counts for PodSets I do not see why we should block  `PodSetSliceSize` 0. IIUC this would be actually the only "valid" explicit value for PodSet count 0... Or are we allowing for values greater than count?
So I would personally treat rejection of 0 replicas with 0 slice size as a regression. Sure users can "if" this on their side but this leads to inferior user experience.

> I missed that discussion, but I'm not sure if allowing replicas = 0 makes the argument. replicas = 0 makes sense for deleting all pods without deleting the Job itself - scale down to 0. For PodSetSliceSize annotation we don't really support elasticity, so I cannot see a legitimate use case.

So what should be the "correct" PodSetSliceSize defined explicitly for 0 replicas? Are we treating any value as valid?

I can think of other use case for Jobs having replicas 0, - it is before they are scheduled.

Let's imagine we have a strict FIFO Cluster Queue, that is running some ETL jobs. The queue is generally large - days and the jobs are of the same importance - so you do not want to push any of them first by the priority, you are fine with waiting some time.  Let's imagine I have a job that is based on the current "demand" - is replicated by each of the customer requests.
Then I would like to scale it according to "current" situation, not the one from many days ago. So as a user I would submit a job with 0 replicas to save the spot in the strict FIFO queue, and then when I know it will be scheduled "soon" update the replicas based on the current demand. It may be 10, it may be 1, or it may even stay 0 if there were no requests.

Obviously it can have replicas 1 by default and work in similar way, but here 0 is natural for me as this will be no-op if I do not "update it'", because some other component of my system fails.

>  I would just introduce a FG like TASRejectPodSetSliceSizeZero and reject in TAS code, workload webhook and job webhooks. Let me confirm the approach change with @dkaluza, wdyt?

For me it depends what do we permit in situation with 0 replicas. If podset slice size has to be always <= than count of pods in the Podset, then it is not a good idea (and this validation generally also makes some sense from my point of view).
If we permit larger PodSetSlice sizes, so that user can set the slice size to X and then just scale the replicas, then it is fine.

mimowo:
> From my perspective we should either support 0 replicas and counts or not. 

The fields have different use-cases. Just because both are "size" it does not mean we need to follow the same validation automatically. 

The first one was relaxed to accomodate the scale down to 0 use case. 

I don't know about the use case for setting PodSetSliceSize=0. If we do have such a use case, then the PR needs to be accompanied by KEP update with the use-case listed.

However, I think this was just an omission during review, not a deliberate maintainer decision to allow to PodSetSliceSize=0.



> So what should be the "correct" PodSetSliceSize defined explicitly for 0 replicas? Are we treating any value as valid?

I think so, we can keep any PodSetSliceSize>0 as valid. 


Overall summary:
1. My current decision to reject PodSetSliceSize>0 is code simplicity, if we want to relax the validation we need to have a use case. Allowing for replicas=0 is not an argument for me, because both fields has a different purpose and validation. Both cannot be negative for sure.
2. Could you please @mszadkow summarize what currently happens on "main" when someone is setting `PodSetSliceSize<0` or `PodSetSliceSize==0` in the annotation? I mean e2e testing on a live cluster. I want to understand what is the current e2e behavior and what will be. I cannot fully undersnad consequences of the change only based on the low level "ifs"
3. we can always relax validation in the future @dkaluza, strengthening validation is very hard

mimowo:
Oh, I just noticed this might be tricky:
> I think so, we can keep any PodSetSliceSize>0 as valid.

in case we have validation (reasonable) that PodSetSlizeSize is <= replicas (parallelism). 

However, still what I'm missing is the summary of the current validations and behavior we have e2e before we start moving so low-level pieces.

mimowo:
So, I think what would maybe help is:
1. break down the PR into validation against negative values: Job-level and Workload level - this is totally uncontrolvesial
2. handling of scenarios where PodSetSize == 0, here we may indeed take a breath, but I want to first have an integration test which documents the scenario: for Job, and for workload - simulating custom job CRD integration.

I'm still leaning to say we can validate against both. If (2.) does not work currently, then we will not worsen any user by validation, we will just do fail fast. The purpose of this Issue is not to provide new functinality , but avoid crashing Kueue. We can revisit supporting `PodSetSize == 0` as a new feature. It might be reasonable, given the limitation (if exists) that PodSetSize <= repliaces.

mszadkow:
@mimowo ack
I will split this into 2 separate scenarios.
I will double check the behavior for both negative and zero values of PodSetSliceSize.

And the second "controversial" one will be suplemented with additional integration test.

mszadkow:
I did cluster tests:
As expected admission webhook forbid the annotation to be smaller than 1:
```
Error from server (Forbidden): error when creating "site/static/examples/tas/sample-job-slice-size-negative.yaml": admission webhook "vjob.kb.io" denied the request: spec.template.metadata.annotations[kueue.x-k8s.io/podset-slice-size]: Invalid value: "-1": must be greater than or equal to 1
```

```
Error from server (Forbidden): error when creating "site/static/examples/tas/sample-job-slice-size-zero.yaml": admission webhook "vjob.kb.io" denied the request: spec.template.metadata.annotations[kueue.x-k8s.io/podset-slice-size]: Invalid value: "0": must be greater than or equal to 1
```

However that was no the reported issue, the vulnerability analysis pointed that when manually crafted Workload with `podSetSliceSize: 0` can cause division by zero and in result panic.

Direct Workload with `spec.podSets[].topologyRequest.podSetSliceSize: 0` is still accepted by API admission.
But division by zero was prevented thanks to the recent change of the constraint: `slice topology requested, but slice size not provided` in `getSliceSizeWithSinglePodAsDefault`.

In the light of this and lack of concrete use-case scenario I believe we can simply go with API validation:
```
+kubebuilder:validation:Minimum=1
```

mimowo:
Thank you @mszadkow , so I think this validation report:
```
Error from server (Forbidden): error when creating "site/static/examples/tas/sample-job-slice-size-zero.yaml": admission webhook "vjob.kb.io" denied the request: spec.template.metadata.annotations[kueue.x-k8s.io/podset-slice-size]: Invalid value: "0": must be greater than or equal to 1
```

 clearly shows the intention is to only support PodSetSlizeSize>=1. 
 
 Even if "PodSetSlizeSize=0" makes some sense that would be a new feature, not the bugfx.

As for the fix, I agree "+kubebuilder:validation:Minimum=1" would be ideal, but cherrypicking schema changes is tricky. Many users when updating patches would really bump the schema, but just do a rolling update of the new image.

Also, it might be tricky to deal with pre-existing workloads if they are somehow there. 

So, I think it makes sense to have the fix behind a Beta FG, but for 0.20 go with the  proposal and use "+kubebuilder:validation:Minimum=1". That is the approach we took here: https://github.com/kubernetes-sigs/kueue/pull/12838

mszadkow:
This PR becomes fix behind a Beta FG.
For 0.20 I added API validation in this PR - https://github.com/kubernetes-sigs/kueue/pull/13634

pkg/features/kube_features.go
```diff
@@ -514,6 +514,9 @@ const (
 	//
 	// Rejects Workloads with negative container or pod-level resource requests/limits.
 	WorkloadValidateResourcesAreNonNegative featuregate.Feature = "WorkloadValidateResourcesAreNonNegative"
+
+	// Rejects Workloads with non-positive TAS slice sizes in PodSet topology requests.
+	WorkloadValidateTASSliceSize featuregate.Feature = "WorkloadValidateTASSliceSize"
```
```suggestion
	TASValidateWorkloadSliceSize featuregate.Feature = "WorkloadValidateTASSliceSize"
```
nit, also add a comment to make "stable" in 0.21

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -1310,16 +1313,20 @@ func slicesRequested(tr *kueue.PodSetTopologyRequest) bool {
 	return len(utiltas.PodSetSliceRequiredTopologyConstraints(tr)) > 0
 }
 
-func getSliceSizeWithSinglePodAsDefault(tr *kueue.PodSetTopologyRequest) (int32, string) {
+func resolveSliceSizeOrReason(tr *kueue.PodSetTopologyRequest) (int32, string) {
```
mimowo: Hm, IIUC changes to this file aren't really needed assuming the validation is in place, right?

mszadkow: true, although I agreed with @dkaluza that this function name is bad, I can move it to separate PR

mimowo: yeah I just doubt we need the changes to code at all provided we implement the validation, but some safe guarding is ok probably

mszadkow: ok, I removed all unnecessary changes

dkaluza (approved):
New implementation looks good to me - taking into account the decision that we do not wish to support 0 slice size for now.

But please update the FG name according to mimowo suggestion if you agree with it

mszadkow:
@mimowo everything got addressed - ptal

test/integration/singlecluster/tas/tas_test.go
```diff
@@ -3747,6 +3504,7 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
 								PodSetGroup("group").
 								PreferredTopologyRequest(utiltesting.DefaultRackTopologyLevel).
 								SliceRequiredTopologyRequest(corev1.LabelHostname).
+								SliceSizeTopologyRequest(1).
```
mimowo: Hmmm, so I see this is needed  because previously this was assumed 0 and now 0 would be rejected. 

I have two questions:
1. would it work ok if we dropped "SliceRequiredTopologyRequest"?
2. I think the PodSetGroup for leader and worker would be mostly interesting to make leader scheduled together with the workers, so:
- leader: SliceSizeTopologyRequest=5, SliceRequiredTopologyRequest, group="group"
- workers: SliceSizeTopologyRequest=5, SliceRequiredTopologyRequest, group="group"

Then we have: `[leader, w0,w1,w2,w3,w4], [w5,w6,w7,w8,w9], [w10,w11,w12,w13,w14]`. IIUC it doesn't work this way today? In that case I'm ok as is.. My question (2.)  is mostly exploratory for now.


In any case, assuming we go this way we need a release note with "ACTION REQUIRED" because this is a breaking change.

mszadkow: 1. That's interesting to learn that with single leader it can be dropped.
Worker count and slice topology drive the main placement state ([count and sliceSize](https://github.com/kubernetes-sigs/kueue/blob/eb40974cf493b708e527d74ab0fe747e8c49b88c/pkg/cache/scheduler/tas_flavor_snapshot.go#L977)) and for the leader it is treated as one reserved leader slot if leader exists - https://github.com/kubernetes-sigs/kueue/blob/eb40974cf493b708e527d74ab0fe747e8c49b88c/pkg/cache/scheduler/tas_flavor_snapshot.go#L966

2. There is no mechanism that constructs slices like [leader,w0,w1,w2,w3,w4] as one atomic 5-sized slice. 
Slices are effectively worker-driven, while leader is modeled as a reserved leader slot on selected domain.

mimowo: > That's interesting to learn that with single leader it can be dropped.
Worker count and slice topology drive the main placement state ([count and sliceSize](https://github.com/kubernetes-sigs/kueue/blob/eb40974cf493b708e527d74ab0fe747e8c49b88c/pkg/cache/scheduler/tas_flavor_snapshot.go#L977)) and for the leader it is treated as one reserved leader slot if leader exists -

So, we could also modify the integartion test to pass if we dropped "SliceRequiredTopologyRequest" instead of adding "SliceSizeTopologyRequest" ?

mimowo: > There is no mechanism that constructs slices like [leader,w0,w1,w2,w3,w4] as one atomic 5-sized slice.
Slices are effectively worker-driven, while leader is modeled as a reserved leader slot on selected domain.

Yeah, maybe currently, but I claim this is more useful usage, because if you use "group", then most often leader is one of workers. 

(I'm not saying we need to change  that in this PR, I'm just surfacing my findings as maybe someone will find them in the future useful)

mszadkow: > > That's interesting to learn that with single leader it can be dropped.
> > Worker count and slice topology drive the main placement state ([count and sliceSize](https://github.com/kubernetes-sigs/kueue/blob/eb40974cf493b708e527d74ab0fe747e8c49b88c/pkg/cache/scheduler/tas_flavor_snapshot.go#L977)) and for the leader it is treated as one reserved leader slot if leader exists -
> 
> So, we could also modify the integartion test to pass if we dropped "SliceRequiredTopologyRequest" instead of adding "SliceSizeTopologyRequest" ?

Yes, we can do it this way too.
Maybe I will drop it and leave a comment why it's not specified.

mimowo (commented):
/lgtm
/approve
Thank you @mszadkow 👍 Please prepare cherrypicks

PR #12729, opened 2026-07-02
kueueviz: bound token cache with LRU and add global rate limiter

### LINKED ISSUE #12561: Denial of Service via Unbounded TokenReview Amplification in cmd/kueueviz/backend/middleware/auth.go

### Impact
An unauthenticated attacker can trivially exhaust the backend's memory (since invalid tokens are cached in an unbounded map until the 1-minute eviction ticker fires) and simultaneously launch a Denial of Service attack against the Kubernetes API server by forcing it to process thousands of bogus `TokenReview` requests.

### Data Flow
- **cmd/kueueviz/backend/middleware/auth.go:61**: The authenticator extracts an unauthenticated, user-provided token from the request header.
- **cmd/kueueviz/backend/middleware/auth.go:70**: If the token is not found in the local cache, the backend immediately calls `reviewToken`.
- **cmd/kueueviz/backend/middleware/auth.go:106**: The backend submits a synchronous `TokenReview` request to the Kubernetes API server for every unique invalid token.

### Reproduction
**Preconditions**:
KueueViz backend is running with `KUEUEVIZ_AUTH_MODE=TokenReview`.

**Steps**:
1. Send thousands of HTTP requests to a protected KueueViz endpoint, each containing a unique, randomly generated Bearer token.
2. Observe the KueueViz backend forwarding all requests to the Kubernetes API server.

### Vulnerable Code
```
result, err := a.clientset.AuthenticationV1().TokenReviews().Create(ctx, review, metav1.CreateOptions{})
```

### Remediation
Implement rate limiting and bound the token cache size.

Suggested Fix:
```
Add an IP-based or global rate limiter before the authentication middleware to throttle excessive requests with invalid tokens. Additionally, replace the unbounded map in `tokenReviewCache` with an LRU cache (e.g., `hashicorp/golang-lru`) to strictly bound memory consumption.
```

### Environment
* Install tools: N/A
* Others: N/A

Vaishnav88sk:
#### What type of PR is this?

/kind bug
/area dashboard

The `tokenReviewCache` in `kueueviz` used an unbounded `map[string]cacheEntry` with a 1-minute eviction ticker. An unauthenticated attacker sending requests with unique random bearer tokens could:
1. **Exhaust memory** — every unique invalid token is cached until the eviction
   ticker fires, with no upper bound on the map size.
2. **Flood the Kubernetes API** — every cache miss immediately triggers a
   synchronous `TokenReview` request, allowing a trivial DoS against the API server.

Fix:
- Replace the unbounded map with a **fixed-capacity LRU cache** (`container/list`, stdlib — no new dependency) capped at 1024 entries. When full, the least-recently-used entry is evicted before inserting a new one.
- Add a **`RateLimiter()` gin middleware** (`golang.org/x/time/rate`, already in go.mod`) applied to all protected routes (100 rps, burst 50). Requests over the limit receive `429 Too Many Requests` before reaching the auth middleware, preventing TokenReview amplification.
- Tests added for LRU eviction, capacity bounding, and rate limiter behaviour.

Fixes #12561

- LRU implementation uses only stdlib (`container/list` + `sync.Mutex`) — no new dependency added.
- `CacheSize` field added to `AuthConfig` (zero → default 1024); fully backward-compatible.
- Rate limit constants (100 rps / burst 50) are set at the call site in `main.go` and can be made configurable via env/config in a follow-up.
- This PR was developed with AI assistance, disclosed per the [Kubernetes AI Tool Usage Policy](https://www.kubernetes.dev/docs/guide/pull-requests/#ai-guidance).

```release-note
Fixed a Denial of Service vulnerability in kueueviz where an attacker could exhaust backend memory and flood the Kubernetes TokenReview API by sending requests with unique invalid bearer tokens.
The token cache is now bounded by an LRU eviction policy and a global rate limiter throttles excessive authentication requests.
```

## Summary by CodeRabbit

* **New Features**
  * Added global request rate limiting for protected routes, enforcing per-client-IP limits (using `X-Forwarded-For` when available).
  * Improved authentication token-review result caching with a configurable, bounded TTL+LRU cache (including separate handling for negative results).
* **Bug Fixes**
  * Over-limit requests now return **429 Too Many Requests** with a JSON error response (`{"error":"too many requests"}`).
  * Prevented unbounded growth by replacing the previous cache with a TTL-expiring LRU.
* **Tests**
  * Added automated coverage for rate-limiter middleware behavior.

### DIFF
--- a/cmd/kueueviz/backend/main.go
+++ b/cmd/kueueviz/backend/main.go
@@ -84,6 +84,9 @@ func main() {
 	handlers.InitializeUnauthenticatedRoutes(publicRoutes, serverConfig.AuthMode)
 
 	protectedRoutes := r.Group("/")
+	// Apply a global rate limiter to all protected routes to prevent
+	// TokenReview amplification (100 rps sustained, burst of 50).
+	protectedRoutes.Use(middleware.RateLimiter(100, 50))
 	if serverConfig.AuthMode == "TokenReview" {
 		slog.Info("Authentication enabled", "mode", serverConfig.AuthMode)
 		auth := middleware.NewAuthenticator(clientset, serverConfig.AuthConfig)
--- a/cmd/kueueviz/backend/middleware/auth.go
+++ b/cmd/kueueviz/backend/middleware/auth.go
@@ -101,28 +101,25 @@ func (a *Authenticator) Stop() {
 //   - r: steady-state requests per second allowed per IP.
 //   - burst: maximum burst size (peak requests from a single IP at once).
 func RateLimiter(r rate.Limit, burst int) gin.HandlerFunc {
-	var (
-		mu       sync.Mutex
-		limiters = make(map[string]*rate.Limiter)
-	)
+	// A bounded, TTL-based cache mapping IP addresses to *rate.Limiter.
+	// This prevents memory exhaustion from spoofed or highly distributed IP addresses.
+	limiters := utilcache.NewLRUExpireCache(10000)
+	var mu sync.Mutex
+
 	getLimiter := func(ip string) *rate.Limiter {
 		mu.Lock()
 		defer mu.Unlock()
-		if l, ok := limiters[ip]; ok {
-			return l
+		if l, ok := limiters.Get(ip); ok {
+			return l.(*rate.Limiter)
 		}
 		l := rate.NewLimiter(r, burst)
-		limiters[ip] = l
+		limiters.Add(ip, l, 10*time.Minute)
 		return l
 	}
 	return func(c *gin.Context) {
-		// Prefer X-Forwarded-For (set by trusted reverse proxies) over
-		// RemoteAddr so that clients behind a load balancer are keyed
-		// correctly rather than all sharing the proxy's IP.
-		ip := c.GetHeader("X-Forwarded-For")
-		if ip == "" {
-			ip = c.ClientIP()
-		}
+		// Rely on gin's ClientIP() which automatically handles X-Forwarded-For
+		// and securely validates it against trusted proxies.
+		ip := c.ClientIP()
 		if !getLimiter(ip).Allow() {
 			c.AbortWithStatusJSON(http.StatusTooManyRequests, gin.H{"error": "too many requests"})
 			return
--- a/cmd/kueueviz/backend/middleware/auth.go
+++ b/cmd/kueueviz/backend/middleware/auth.go
@@ -17,6 +17,7 @@ limitations under the License.
 package middleware
 
 import (
+	"container/list"
 	"context"
 	"crypto/sha256"
 	"encoding/base64"
@@ -31,6 +32,7 @@ import (
 	authenticationv1 "k8s.io/api/authentication/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
+	"golang.org/x/time/rate"
 )
 
 const (
@@ -38,6 +40,12 @@ const (
 	WebSocketBaseProtocol = "kueueviz.v1"
 
 	webSocketTokenProtocolPrefix = "kueueviz.auth."
+
+	// defaultCacheSize is the maximum number of token entries held in the LRU
+	// cache. When the limit is reached the least-recently-used entry is evicted,
+	// bounding memory regardless of how many unique (invalid) tokens an attacker
+	// sends.
+	defaultCacheSize = 1024
 )
 
 type Clock interface {
@@ -58,6 +66,9 @@ type AuthConfig struct {
 	Audiences        []string
 	CacheTTL         time.Duration
 	NegativeCacheTTL time.Duration
+	// CacheSize is the maximum number of distinct tokens held in the LRU cache.
+	// Zero means defaultCacheSize (1024).
+	CacheSize int
 }
 
 type Authenticator struct {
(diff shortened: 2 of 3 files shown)


mimowo:
/ok-to-test
cc @olekzabl

olekzabl:
Let me ask early: have you considered using [this](https://github.com/kubernetes/apimachinery/blob/514a87090e39d314e50aa3911a5afec5b41ffd15/pkg/util/cache/lruexpirecache.go#L53-L54)?

* TTLs are supported per each entry (see [here](https://github.com/kubernetes/apimachinery/blob/514a87090e39d314e50aa3911a5afec5b41ffd15/pkg/util/cache/lruexpirecache.go#L72-L73))
* periodic cleanup (an analogue of `evictExpired`) is not supported - but probably also no longer needed, given the bounded size?

So it _seems_ like it could fit our needs, without writing much new code - though I admit I didn't check deep in detail.

Vaishnav88sk:
> Let me ask early: have you considered using [this](https://github.com/kubernetes/apimachinery/blob/514a87090e39d314e50aa3911a5afec5b41ffd15/pkg/util/cache/lruexpirecache.go#L53-L54)?
> 
> * TTLs are supported per each entry (see [here](https://github.com/kubernetes/apimachinery/blob/514a87090e39d314e50aa3911a5afec5b41ffd15/pkg/util/cache/lruexpirecache.go#L72-L73))
> * periodic cleanup (an analogue of `evictExpired`) is not supported - but probably also no longer needed, given the bounded size?
> 
> So it _seems_ like it could fit our needs, without writing much new code - though I admit I didn't check deep in detail.

Great suggestion! I wasn't aware of `LRUExpireCache`, but it perfectly matches our requirements. I've updated the PR to replace our custom cache implementation entirely with `k8s.io/apimachinery/pkg/util/cache.LRUExpireCache`. This allowed us to drop the periodic cleanup goroutine and over 100 lines of manual LRU logic. I've also updated the tests accordingly. Thanks for pointing this out!

danylott:
Great update using LRUExpireCache for the tokens!

I noticed a couple of things with the RateLimiter middleware that we might want to adjust. 

1. the limiters map is unbounded and entries are never cleaned up. Could an attacker exhaust memory by sending requests with different IPs to continuously grow this map? Maybe we could use another bounded LRUExpireCache here for the rate limiters to prevent this?
2. I see the comment about preferring X-Forwarded-For over RemoteAddr. Gin's c.ClientIP() actually does this automatically, but it validates the headers against trusted proxies. By reading the header directly with c.GetHeader, we bypass that validation, meaning anyone could spoof their IP to bypass the rate limiter. I think we should just rely entirely on c.ClientIP().

What do you think?

Vaishnav88sk:
> What do you think?

I completely agree with both points. 

1. Using an unbounded map for the rate limiters definitely introduced a memory exhaustion risk if an attacker continuously spoofed IP addresses. I've refactored the `RateLimiter` middleware to store the limiters in a bounded `utilcache.LRUExpireCache` (capped at 10,000 entries) with a 10-minute TTL per entry. This guarantees a firm ceiling on memory usage.
2. You're right about Gin's `c.ClientIP()`, it automatically manages the proxy trust boundaries securely. I've dropped the manual `c.GetHeader("X-Forwarded-For")` call and we now rely entirely on `c.ClientIP()`.

I'll push these updates shortly. Thanks for the thorough review!

olekzabl:
Thank you for applying all the changes!

Still, I'd like to discuss one thing: how should we actually behave in the scenario mentioned by @danylott?

> Could an attacker (...) by sending requests with different IPs (...)

Danylo's concern was about growing `limiters` map, and you addressed that.
But, what will happen to the auth requests?
I think it's tricky.

Let's say the attacker has a pool of 16k IPs, and rotates them every 1ms.
In that scenario, am I right that:

1. With the current code, _all_ attacker's requests will pass our rate limiters, forever?

2. Even if made `limiters` map unbounded, it'd be still the same?
   (Every single IP would circle back as rarely as once per 16 seconds, passing its own rate limiter by a broad margin).

3. To avoid DDoS-ing K8s auth endpoint, we'd need some kind of global rate limiting on the KueueViz side - which in turn would undermine the current gain that the "good citizen IPs" are not punished by attacker's attempts?

I'm not sure what's the best solution here.
Maybe two-level setup of rate-limiters (per IP + global)?
WDYT @Vaishnav88sk @danylott ?

Vaishnav88sk:
> I'm not sure what's the best solution here. Maybe two-level setup of rate-limiters (per IP + global)? WDYT @Vaishnav88sk @danylott ?

You're right! Per-IP limiting alone can't defend against a distributed IP-rotation attack. A two-level setup makes sense: a **global rate limiter** to cap the total number of TokenReview calls kueueviz makes to K8s (protecting the API server), combined with the existing **per-IP limiter** that protects individual users from being throttled by others.

The trade-off of the global limiter is that a coordinated attacker could still exhaust the global budget and slow down legitimate users, but that's likely acceptable compared to DDoS-ing the auth endpoint entirely.

danylott:
I agree with Olek example of 16k different IPs - it will still exhaust the system, the [original issue](https://github.com/kubernetes-sigs/kueue/issues/12561) actually propose implementing either global or IP-based rate limiter, so if we can do both - amazing.

If only one is viable - then let's make sure to create another issue connected to the situation described by Olek, wdyt?

CC @Vaishnav88sk @olekzabl

Vaishnav88sk:
I agree with both of you, @danylott and @olekzabl. Since implementing a robust two-level rate limiter (per-IP + global) adds a bit more complexity and requires careful configuration, I think the most viable path right now is to merge this PR to fix the immediate unbounded memory risk with the per-IP `LRUExpireCache`. 

I'll go ahead and create a follow-up issue tracking the IP-rotation attack vector so we can properly design and implement the global rate limiter there. Does that sound good?

danylott (approved):
Please put a link to an issue for global rate-limiter.
Other than that LGTM

Vaishnav88sk:
> Please put a link to an issue for global rate-limiter. Other than that LGTM

Thanks for the approval! I've opened the follow-up issue to track the global rate limiter implementation here: #13172.

Vaishnav88sk:
I think it is good to merge. 👍🏻

Vaishnav88sk:
Closing this PR as it has been fully superseded by #13173! 

The `LRUExpireCache` and per-IP rate-limiting implementations originally proposed here were integrated into that PR alongside the new global rate limiter, and everything has now been successfully merged into `main`. 

Thanks to everyone for the feedback and reviews! @mimowo

PR #12733, opened 2026-07-02
fix: permanent TAS resource leak during workload deletion with missing topology

### LINKED ISSUE #12545: Permanent TAS Resource Leak when ClusterQueue TAS initialization fails during workload deletion in pkg/cache/scheduler/clusterqueue.go
### Impact
A permanent resource leak in the scheduler's cache. If a ClusterQueue relies on multiple TAS flavors, a temporary failure or misconfiguration of one flavor (e.g., deleted Topology) causes any completing or deleted workloads using the *other* healthy flavors to leak their TAS resources globally. Over time, or via intentional tenant action, this leads to a Denial of Service (DoS) where the global `tasCache` believes all topologies are fully utilized, stalling the cluster's batch scheduling. Recovery requires manually restarting the Kueue controller pod.

### Data Flow
- **pkg/cache/scheduler/cache.go:710**: User deletes an admitted Workload, triggering Cache.DeleteWorkload().
- **pkg/cache/scheduler/clusterqueue.go:490**: c.deleteWorkload calls c.updateWorkloadUsage(..., subtract) to remove the workload's resource usage.
- **pkg/cache/scheduler/clusterqueue.go:530**: updateWorkloadUsage removes normal resources, then calls c.updateWorkloadTASUsage(..., subtract) to remove TAS resources.
- **pkg/cache/scheduler/clusterqueue.go:583**: If the ClusterQueue is temporarily missing ANY of its referenced topologies/flavors, c.isTASInitialized() returns false. The function returns early.
- **pkg/cache/scheduler/clusterqueue.go:498**: The workload is permanently deleted from c.Workloads. Its TAS usage for valid flavors remains stuck in the global tasCache. When the missing topology is restored, ensureTASIsSynced ignores this workload because it was removed from c.Workloads, finalizing the leak.

### Reproduction
**Preconditions**:
TopologyAwareScheduling feature gate must be enabled. A ClusterQueue must use multiple TAS-enabled ResourceFlavors.

**Steps**:
1. Create topologies 'topo-a' and 'topo-b', and flavors 'flavor-a' and 'flavor-b' referencing them.
2. Create a ClusterQueue 'cq-tas' that references both 'flavor-a' and 'flavor-b'.
3. Submit a TAS Workload to 'cq-tas' that requests 'flavor-a' and ensure it gets admitted.
4. Temporarily delete 'topo-b'. The ClusterQueue's isTASInitialized() state becomes false.
5. Delete the admitted Workload while 'topo-b' is missing.
6. Restore 'topo-b'.
7. Observe the TAS metrics or scheduler state; the workload's resource usage for 'flavor-a' is permanently leaked.

### Vulnerable Code
```
if !c.isTASInitialized() {
	log.V(2).Info("Delaying accounting of the TAS usage, because TAS cache is not initialized yet")
	// TAS cache is not synced yet so we defer accounting for TAS usage.
	return
}
```

### Remediation
Process individual flavors unconditionally during subtraction, rather than returning early if the entire ClusterQueue is uninitialized.

Suggested Fix:
```
In `updateWorkloadTASUsage`, remove the `!c.isTASInitialized()` early return. Instead, iterate over `wi.TASUsage()` and apply `tasFlvCache.addUsage` / `removeUsage` for any flavor that exists in the cache, ignoring only those that are truly missing.
```

### Environment
* Install tools: N/A
* Others: N/A

Vaishnav88sk:
# Description
#### What type of PR is this?

/kind bug
/area tas

Fixes a permanent resource leak in the scheduler's TAS cache. When a ClusterQueue relies on multiple TAS flavors, a temporary failure or misconfiguration (e.g., a deleted Topology) causes `isTASInitialized()` to evaluate to false. Previously, this caused an early return in `updateWorkloadTASUsage`, meaning completing or deleted workloads leaked their TAS resource usage globally. 

This PR removes the `!c.isTASInitialized()` early return in `updateWorkloadTASUsage`. Now, it unconditionally iterates over the workload's TAS usage and applies `addUsage` / `removeUsage` for any flavor that currently exists in the cache, safely ignoring missing flavors and allowing proper cleanup of resources during workload deletion.

Fixes #12545

```release-note
TAS: Fixed a bug that permanently leaked Topology-Aware Scheduling (TAS) resources if a workload was deleted while its ClusterQueue was temporarily missing a required Topology.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TAS usage accounting to update immediately, rather than waiting for TAS initialization.
  * TAS flavor usage updates now correctly self-heal when the same workload is re-added, and safely handle missing entries during add/remove to keep accounting accurate.

* **Tests**
  * Added an integration test to ensure TAS domains are not leaked when a workload is deleted while a topology is uninitialized, and that admission recovers correctly afterward.
  * Added a unit test covering TAS flavor usage aggregation and add/remove behavior.

### DIFF
--- a/pkg/cache/scheduler/clusterqueue.go
+++ b/pkg/cache/scheduler/clusterqueue.go
@@ -605,7 +605,7 @@ func (c *clusterQueue) updateWorkloadTASUsage(log logr.Logger, wi *workload.Info
 		case op == add:
 			tasFlvCache.addUsage(log, key, tasUsage)
 		case op == subtract:
-			tasFlvCache.removeUsage(key)
+			tasFlvCache.removeUsage(log, key)
 		}
 	}
 }
--- a/pkg/cache/scheduler/clusterqueue.go
+++ b/pkg/cache/scheduler/clusterqueue.go
@@ -603,7 +603,7 @@ func (c *clusterQueue) updateWorkloadTASUsage(log logr.Logger, wi *workload.Info
 		case tasFlvCache == nil:
 			log.V(2).Info("TAS flavor used by workload not found in cache", "tasFlavor", tasFlavor)
 		case op == add:
-			tasFlvCache.addUsage(key, tasUsage)
+			tasFlvCache.addUsage(log, key, tasUsage)
 		case op == subtract:
 			tasFlvCache.removeUsage(key)
 		}
--- a/pkg/cache/scheduler/clusterqueue.go
+++ b/pkg/cache/scheduler/clusterqueue.go
@@ -597,11 +597,6 @@ func (c *clusterQueue) updateWorkloadTASUsage(log logr.Logger, wi *workload.Info
 	}
 	key := workload.Key(wi.Obj)
 	log = log.WithValues("workload", key)
-	if !c.isTASInitialized() {
-		log.V(2).Info("Delaying accounting of the TAS usage, because TAS cache is not initialized yet")
-		// TAS cache is not synced yet so we defer accounting for TAS usage.
-		return
-	}
 	for tasFlavor, tasUsage := range wi.TASUsage() {
 		tasFlvCache := c.tasCache.Get(tasFlavor)
 		switch {
--- a/pkg/cache/scheduler/tas_flavor.go
+++ b/pkg/cache/scheduler/tas_flavor.go
@@ -157,15 +157,16 @@ func (c *TASFlavorCache) snapshot(
 func (c *TASFlavorCache) addUsage(log logr.Logger, key workload.Reference, topologyRequests []workload.TopologyDomainRequests) {
 	if _, found := c.wlUsage[key]; found {
 		log.V(2).Info("Workload usage already exists in TAS flavor cache, self-healing by replacing it", "workload", key)
-		c.removeUsage(key)
+		c.removeUsage(log, key)
 	}
 	c.wlUsage[key] = slices.Clone(topologyRequests)
 	c.updateUsage(topologyRequests, add)
 }
 
-func (c *TASFlavorCache) removeUsage(key workload.Reference) {
+func (c *TASFlavorCache) removeUsage(log logr.Logger, key workload.Reference) {
 	value, found := c.wlUsage[key]
 	if !found {
+		log.V(2).Info("Workload usage not found during removal from TAS flavor cache", "workload", key)
 		return
 	}
 	c.updateUsage(value, subtract)
--- a/pkg/cache/scheduler/tas_flavor.go
+++ b/pkg/cache/scheduler/tas_flavor.go
@@ -154,7 +154,11 @@ func (c *TASFlavorCache) snapshot(
 	return snapshot
 }
 
-func (c *TASFlavorCache) addUsage(key workload.Reference, topologyRequests []workload.TopologyDomainRequests) {
+func (c *TASFlavorCache) addUsage(log logr.Logger, key workload.Reference, topologyRequests []workload.TopologyDomainRequests) {
+	if _, found := c.wlUsage[key]; found {
+		log.V(2).Info("Workload usage already exists in TAS flavor cache, self-healing by replacing it", "workload", key)
+		c.removeUsage(key)
+	}
 	c.wlUsage[key] = slices.Clone(topologyRequests)
 	c.updateUsage(topologyRequests, add)
 }
--- a/test/integration/singlecluster/tas/tas_test.go
+++ b/test/integration/singlecluster/tas/tas_test.go
@@ -1080,12 +1080,11 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
 				var wl1, wl2 *kueue.Workload
 				ginkgo.By("creating a workload which requires block and can fit", func() {
 					wl1 = utiltestingapi.MakeWorkload("wl1", ns.Name).
-						Queue(kueue.LocalQueueName(localQueue.Name)).
-						PodSets(*utiltestingapi.MakePodSet("main", 1).
-							Request(corev1.ResourceCPU, "4").
-							RequiredTopologyRequest(utiltesting.DefaultBlockTopologyLevel).
-							Obj()).
-						Obj()
+						Queue(kueue.LocalQueueName(localQueue.Name)).Request(corev1.ResourceCPU, "4").Obj()
+					wl1.Spec.PodSets[0].Count = 1
+					wl1.Spec.PodSets[0].TopologyRequest = &kueue.PodSetTopologyRequest{
+						Required: ptr.To(utiltesting.DefaultBlockTopologyLevel),
+					}
 					util.MustCreate(ctx, k8sClient, wl1)
 				})
 
@@ -1135,12 +1134,11 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
 
 					// Request the capacity that was freed by wl1.
 					wl2 = utiltestingapi.MakeWorkload("wl2", ns.Name).
-						Queue(kueue.LocalQueueName(localQueue.Name)).
-						PodSets(*utiltestingapi.MakePodSet("main", 1).
-							Request(corev1.ResourceCPU, "4").
-							RequiredTopologyRequest(utiltesting.DefaultBlockTopologyLevel).
-							Obj()).
-						Obj()
+						Queue(kueue.LocalQueueName(localQueue.Name)).Request(corev1.ResourceCPU, "4").Obj()
+					wl2.Spec.PodSets[0].Count = 1
+					wl2.Spec.PodSets[0].TopologyRequest = &kueue.PodSetTopologyRequest{
+						Required: ptr.To(utiltesting.DefaultBlockTopologyLevel),
+					}
 					util.MustCreate(ctx, k8sClient, wl2)
 				})
 
--- a/test/integration/singlecluster/tas/tas_test.go
+++ b/test/integration/singlecluster/tas/tas_test.go
@@ -1132,11 +1132,9 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
 
 					// Request the capacity that was freed by wl1.
 					wl2 = utiltestingapi.MakeWorkload("wl2", ns.Name).
-						Queue(kueue.LocalQueueName(localQueue.Name)).Request(corev1.ResourceCPU, "4").Obj()
-					wl2.Spec.PodSets[0].Count = 1
-					wl2.Spec.PodSets[0].TopologyRequest = &kueue.PodSetTopologyRequest{
-						Required: ptr.To(utiltesting.DefaultBlockTopologyLevel),
-					}
+						Queue(kueue.LocalQueueName(localQueue.Name)).PodSets(*utiltestingapi.MakePodSet("worker", 2).
+						RequiredTopologyRequest(utiltesting.DefaultBlockTopologyLevel).
+						Obj()).Request(corev1.ResourceCPU, "1").Obj()
 					util.MustCreate(ctx, k8sClient, wl2)
 				})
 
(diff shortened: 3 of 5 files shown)


pkg/cache/scheduler/clusterqueue.go
```diff
@@ -597,11 +597,6 @@ func (c *clusterQueue) updateWorkloadTASUsage(log logr.Logger, wi *workload.Info
 	}
 	key := workload.Key(wi.Obj)
 	log = log.WithValues("workload", key)
-	if !c.isTASInitialized() {
```
mimowo, line 600: I think this is reasonable. Specifically, I think for deletion that should be safe because we added the safe-guard making `tasFlvCache.removeUsage` indempotent with the wlUsage [here]( https://github.com/kubernetes-sigs/kueue/blob/c968bd8b9d60f918b3e26ae3f17423a3f063a31b/pkg/cache/scheduler/tas_flavor.go#L162-L169). 

I think the safety for `tasFlvCache.addUsage` relies on the fact that we would never call `updateWorkloadTASUsage` twice with op=add. We could for safety add also a guard inside addUsage to first call removeUsage if the wlUsage contains the entry already, wdyt?

pkg/cache/scheduler/clusterqueue.go
```diff
@@ -597,11 +597,6 @@ func (c *clusterQueue) updateWorkloadTASUsage(log logr.Logger, wi *workload.Info
...
-		log.V(2).Info("Delaying accounting of the TAS usage, because TAS cache is not initialized yet")
-		// TAS cache is not synced yet so we defer accounting for TAS usage.
-		return
```
mimowo, line 603: Please add an integration test for tas. I think this should be possible with the flow:
1. setup TAS (RF, Topology)
2. admit workload
3. delete the Topology <- isTASInitialized = false
4. delete the Workload <- leak of TAS domain usage
5. attempt to admit another workload <- fails because the domains are occupied by the phantom leak

wdyt?

pkg/cache/scheduler/clusterqueue.go
mimowo: I would at least like to add a log when addUsage encounters pre-existing wlUsage, but I think making the add self-heal is ok.

Vaishnav88sk: Done. I've updated `tasFlvCache.addUsage` to check if the workload usage already exists. If it does, it emits a `V(2)` log and calls `removeUsage` to safely self-heal before re-adding the usage.

pkg/cache/scheduler/clusterqueue.go
Vaishnav88sk: My mistake! I've added the integration test tracking this exact flow to `test/integration/singlecluster/tas/tas_test.go`. It successfully validates that the domain usage doesn't leak during the uninitialized state and correctly admits subsequent workloads.

pkg/cache/scheduler/tas_flavor.go
mimowo, line 167: As we add the logger to addUsage, let's also add to removeUsage - here the situation is also rather exceptional - under !found.

pkg/cache/scheduler/tas_flavor.go
```diff
@@ -154,7 +154,11 @@ func (c *TASFlavorCache) snapshot(
 	return snapshot
 }
 
-func (c *TASFlavorCache) addUsage(key workload.Reference, topologyRequests []workload.TopologyDomainRequests) {
+func (c *TASFlavorCache) addUsage(log logr.Logger, key workload.Reference, topologyRequests []workload.TopologyDomainRequests) {
+	if _, found := c.wlUsage[key]; found {
+		log.V(2).Info("Workload usage already exists in TAS flavor cache, self-healing by replacing it", "workload", key)
```
mimowo, line 159: Let's check if this appears ever in the integration test logs currently on CI arftifacts

pkg/cache/scheduler/tas_flavor.go
```diff
@@ -154,7 +154,11 @@ func (c *TASFlavorCache) snapshot(
...
+		c.removeUsage(key)
```
mimowo: I think we need to make sure this is also covered by unit tests - I guess that could be a new testing suite just for addUsage, and removeUsage functions in tas_flavor_test.go, wdyt?

test/integration/singlecluster/tas/tas_test.go
```diff
@@ -1075,6 +1075,68 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
...
+
+				ginkgo.By("creating the second workload to verify the domain usage was not leaked", func() {
+					// Wait for the ClusterQueue to become active again
+					util.ExpectClusterQueuesToBeActive(ctx, k8sClient, clusterQueue)
+
+					// Request the capacity that was freed by wl1.
+					wl2 = utiltestingapi.MakeWorkload("wl2", ns.Name).
+						Queue(kueue.LocalQueueName(localQueue.Name)).Request(corev1.ResourceCPU, "4").Obj()
+					wl2.Spec.PodSets[0].Count = 1
+					wl2.Spec.PodSets[0].TopologyRequest = &kueue.PodSetTopologyRequest{
```
mimowo: nit, but I think we should be able to use the wrapper pattern here for the PodSets

test/integration/singlecluster/tas/tas_test.go
```diff
@@ -1075,6 +1075,68 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
...
+
+				ginkgo.By("delete the topology so that it becomes uninitialized", func() {
+					gomega.Expect(k8sClient.Delete(ctx, topology)).Should(gomega.Succeed())
+					// Remove the finalizer to force deletion
+					gomega.Eventually(func(g gomega.Gomega) {
+						err := k8sClient.Get(ctx, client.ObjectKeyFromObject(topology), topology)
+						if err != nil {
+							g.Expect(client.IgnoreNotFound(err)).Should(gomega.Succeed())
+							return
+						}
```
mimowo: That could be inlined - we already have tests which delete topology, check how they do it

test/integration/singlecluster/tas/tas_test.go
```diff
@@ -1075,6 +1075,68 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
...
+					util.MustCreate(ctx, k8sClient, wl1)
+				})
+
+				ginkgo.By("verify the workload is admitted", func() {
+					util.ExpectWorkloadsToBeAdmitted(ctx, k8sClient, wl1)
+					util.ExpectAdmittedWorkloadsTotalMetric(clusterQueue, "", 1)
+				})
+
+				ginkgo.By("delete the topology so that it becomes uninitialized", func() {
+					gomega.Expect(k8sClient.Delete(ctx, topology)).Should(gomega.Succeed())
```
mimowo: Before we issue the delete let's wait for the finalizer to be present first - to make sure we don't have a race conditions where the finalizer is being added concurrently with the deletions.

For example, technically speaking we may have:
1. Topology object created without finalizer
2. the Reconcile which is adding the finalizer can be very slow ...
3. in the test code we issue delete and check no finalizer - proceeed
4. now the Reconcile goroutine adds the finalizer

So to prevent that (even if very unlikely) I would suggest to first wait for the finalizer.

test/integration/singlecluster/tas/tas_test.go
```diff
@@ -1075,6 +1075,68 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
...
+					wl2.Spec.PodSets[0].Count = 1
+					wl2.Spec.PodSets[0].TopologyRequest = &kueue.PodSetTopologyRequest{
+						Required: ptr.To(utiltesting.DefaultBlockTopologyLevel),
+					}
+					util.MustCreate(ctx, k8sClient, wl2)
+				})
+
+				ginkgo.By("verify the second workload is admitted", func() {
+					util.ExpectWorkloadsToBeAdmitted(ctx, k8sClient, wl2)
+					util.ExpectAdmittedWorkloadsTotalMetric(clusterQueue, "", 2)
```
mimowo, line 1143: Please do sanity testing to confirm the test fails here when the bugfix is reverted.

pkg/cache/scheduler/tas_flavor.go
Vaishnav88sk: Agreed. I added the logger to `removeUsage` as well so we can track when it's called for a workload that doesn't exist. This helps maintain consistency in our debugging logs.

pkg/cache/scheduler/tas_flavor.go
Vaishnav88sk: I've verified locally and on CI that this self-healing log is not triggered during normal, healthy operations (including all existing integration tests). It correctly only fires as a fail-safe when state gets out of sync, which is exactly the scenario we're fixing here.

pkg/cache/scheduler/tas_flavor.go
Vaishnav88sk: That makes sense. I have created a new testing suite in `tas_flavor_test.go` specifically covering the idempotent behavior of `addUsage` (replacing existing usage) and `removeUsage` (gracefully handling missing keys) to ensure these self-healing paths are explicitly tested.

test/integration/singlecluster/tas/tas_test.go
Vaishnav88sk: Thanks! Refactored the workload creation to use the `utiltestingapi.MakePodSet` wrapper pattern to keep the test code clean and consistent with the rest of the suite.

test/integration/singlecluster/tas/tas_test.go
Vaishnav88sk: Added an `Eventually` block to wait and assert that `kueue.ResourceInUseFinalizerName` is present on the topology before we issue the delete command and strip the finalizer. This will guarantee deterministic test behavior.

test/integration/singlecluster/tas/tas_test.go
Vaishnav88sk: I have confirmed this locally. When the early return `if !c.isTASInitialized() { return }` is added back into `updateWorkloadTASUsage`, this test successfully catches the regression, the second workload fails to admit because the domain usage is leaked by the deleted first workload. The test passes when the fix is applied.

test/integration/singlecluster/tas/tas_test.go
Vaishnav88sk: The existing tests that delete topologies simply assert `k8sClient.Delete(ctx, topology)`. However, because the topology is still referenced by the `tasFlavor` in the cache, the topology controller places a finalizer (`kueue.ResourceInUseFinalizerName`) on it. As a result, the topology isn't actually deleted from the cluster, it just receives a `DeletionTimestamp`, which means `isTASInitialized` remains `true` in the cache, preventing us from simulating the uninitialized state needed to reproduce this bug. 

To truly delete it and trigger the `isTASInitialized = false` state, we have to manually strip the finalizer here. Let me know if there's a cleaner way to force the cache to drop the topology!

mimowo: Yes, I know the finalizer needs to be managed, but we could do this with a bit simpler code, as here: https://github.com/kubernetes-sigs/kueue/blob/f1aeeb4150b0542d65ac3c8ab735282bb0a3756a/test/integration/singlecluster/tas/tas_test.go#L934-L951

Vaishnav88sk: Hm! I got your point. I simplified it. It's definitely much cleaner and consistent with the other tests.

mimowo:
this one is realeted: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12733/pull-kueue-test-integration-shard-0-main/2072961283233157120
the other are infra:
/retest

Vaishnav88sk:
> this one is realeted: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12733/pull-kueue-test-integration-shard-0-main/2072961283233157120 the other are infra: /retest

Done 👍🏻

mimowo:
/lgtm
/approve
Thank you for hardening TAS!
/cherry-pick release-0.18
/cherry-pick release-0.17

mimowo:
/release-note-edit
```release-note
TAS: Fixed a bug that permanently leaked Topology-Aware Scheduling (TAS) resources if a workload was deleted while its ClusterQueue was temporarily missing a required Topology.
```

PR #12734, opened 2026-07-02
kueueviz: fix CSWSH by removing permissive CheckOrigin

Vaishnav88sk:
# Description
#### What type of PR is this?

/kind bug
/area dashboard

#### What this PR does / why we need it:

Fixes a Cross-Site WebSocket Hijacking (CSWSH) vulnerability in the KueueViz Backend where the WebSocket handler allowed connections from any origin. Previously, the `CheckOrigin` function unconditionally returned `true`, which completely bypassed CORS validation for WebSockets and could allow an attacker to silently extract sensitive real-time Kubernetes cluster metadata if a victim visited a malicious external website while running KueueViz.

This PR removes the permissive `CheckOrigin` override in `GenericWebSocketHandler`. By removing this custom function, `gorilla/websocket` automatically falls back to its default secure validation behavior, which strictly compares the `Origin` header against the `Host` header to ensure they match, securing the real-time WebSocket endpoints against CSWSH attacks. 

#### Which issue(s) this PR fixes:

Fixes #12531

#### Does this PR introduce a user-facing change?

```release-note
KueueViz: Fixed a Cross-Site WebSocket Hijacking (CSWSH) vulnerability in the KueueViz Backend by strictly validating WebSocket Origin headers to prevent unauthorized cross-origin data extraction.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * WebSocket connections now validate the request origin before allowing a connection, improving protection against unauthorized cross-site access.
  * Local development access remains supported for localhost and 127.0.0.1, while matching allowed origins continues to work as expected.

Issue #12534, opened 2026-06-26
Denial of Service via non-transitive comparison function in ClusterQueue Snapshot in pkg/cache/queue/cluster_queue.go


cpiekacz:
_Note: This issue was initially flagged by an automated AI security scan and has been manually triaged by a human_

### Impact
When the visibility API is accessed, the controller attempts to generate a snapshot. The invalid sorting function causes a runtime panic, which blocks workload visibility and could disrupt internal scheduling if invoked elsewhere, leading to Denial of Service.

### Data Flow
- **pkg/cache/queue/cluster_queue.go:616**: The function `buildSnapshotSort` creates a custom sort function for fair-sharing that relies on a `usageCache` map.
- **pkg/cache/queue/cluster_queue.go:672**: Not all LocalQueues are added to `usageCache` (e.g., if `afsConsumedResources.Peek(lqKey)` returns nil or `getLQWeight(lqKey)` fails).
- **pkg/cache/queue/cluster_queue.go:680**: If one item is in the cache but the other is not, the code falls back to `baseCmp(a, b)`.
- **pkg/cache/queue/cluster_queue.go:675**: This fallback conditionally violates the strict weak ordering requirement (transitivity). For elements A, B in the cache where A < B, and element C outside the cache, it's possible to create a cycle where A < B, B < C, and C < A. Go's `slices.SortFunc` detects this cycle and panics, crashing the Kueue controller.

### Reproduction
**Preconditions**: Fair sharing must be enabled. The Kueue controller must be managing workloads.
**Steps**:
1. Create LocalQueue A and submit a high-priority workload to establish a usage value (e.g. usage = 20).
2. Create LocalQueue B and submit a low-priority workload to establish a lower usage value (e.g. usage = 10).
3. Create LocalQueue C and submit a medium-priority workload, but ensure the queue has NO consumed resources (so it is omitted from `usageCache`).
4. When `ClusterQueue.Snapshot()` is called (e.g., via the visibility API `PendingWorkloadsInfo`), the `slices.SortFunc` attempts to sort the workloads from A, B, and C.
5. The non-transitive relationship triggers the runtime `panic("slices.SortFunc: comparison function is not a strict weak ordering")`.

### Vulnerable Code
```
if okA && okB {
        if cmpResult := cmp.Compare(usageA, usageB); cmpResult != 0 {
                return cmpResult
        }
}
return baseCmp(a, b)
```

### Remediation
[...]

PR #12736, opened 2026-07-03
Fix data race on stickyWorkload between Snapshot and RequeueIfNotPresent

### LINKED ISSUE #12532: Data Race and Controller DoS via Visibility API and Preemption in pkg/cache/queue/cluster_queue.go

### Impact
A low-privileged user or tenant who is only authorized to read visibility metrics and submit standard batch workloads can intentionally trigger a data race. This leads to a persistent DoS (Denial of Service) via continuous controller panics, disrupting batch scheduling for the entire Kubernetes cluster.

### Data Flow
- **pkg/cache/queue/manager.go:793**: The PendingWorkloadsInfo method calls cq.Snapshot() WITHOUT holding Manager's Lock() or ClusterQueue's rwm.Lock() during the sorting phase.
- **pkg/cache/queue/cluster_queue.go:375**: ClusterQueue.Snapshot() calls snapshotSort(elements) without any locks.
- **pkg/cache/queue/cluster_queue.go:763**: The sorting function baseCompareFunc reads the multi-word string field c.sw.workloadName via the matches() method.
- **pkg/cache/queue/cluster_queue.go:738**: Concurrently, a scheduler preemption triggers RequeueIfNotPresent (holding Manager's Lock, but NOT ClusterQueue's rwm.Lock), which performs an unsynchronized write to the multi-word string c.sw.workloadName.

### Reproduction
**Preconditions**:
1. Kueue must be deployed with a BestEffortFIFO ClusterQueue and the Visibility API enabled.
2. The attacker needs permissions to create/delete jobs/workloads and view the pending workloads via the Visibility API.

**Steps**:
1. Start a background script that repeatedly queries the Visibility API: `watch -n 0.1 kubectl get --raw /apis/visibility.kueue.x-k8s.io/v1beta2/clusterqueues/<cq-name>/pendingworkloads`
2. Create a low-priority workload that fully consumes the ClusterQueue's quota.
3. Create a high-priority workload targeting the same ClusterQueue.
4. The scheduler will trigger a preemption, causing `RequeueIfNotPresent` to be called with `RequeueReasonPendingPreemption`.
5. The concurrent read/write to the `c.sw.workloadName` string field will trigger a data race (torn read), crashing the controller.

### Vulnerable Code
```
c.sw.set(workload.Key(wInfo.Obj))
```

### Remediation
Synchronize access to the `stickyWorkload` string field by protecting it with `ClusterQueue`'s `rwm` lock during both reads and writes.

Suggested Fix:
```
Modify `RequeueIfNotPresent` to hold `c.rwm.Lock()` before modifying `c.sw`. Update `Snapshot()` so that `c.snapshotSort` is called while holding `c.rwm.RLock()`, or precompute the `stickyWorkload` matches inside `totalElements` so the sorting function operates on a safe, immutable snapshot.
```

### Environment
* Install tools: N/A
* Others: N/A

somaz94:
#### What type of PR is this?
/kind bug

`stickyWorkload.workloadName` is read by `ClusterQueue.Snapshot()` through the sort comparator (`baseCompareFunc` -> `matches`) without holding the ClusterQueue's `rwm` lock, while `RequeueIfNotPresent()` writes it via `set()` during a BestEffortFIFO preemption requeue. The read path (Visibility API `PendingWorkloadsInfo` -> `Snapshot`) and the write path run in different lock domains, so access to the field is unsynchronized. That is a data race on a multi-word string, confirmed by `go test -race`.

This makes `stickyWorkload` self-synchronizing: the struct holds `workloadName` as an `atomic.Pointer[workload.Reference]`, and `matches` / `set` / `clear` do a single atomic load/store for every read and write (`nil` means no sticky workload). All access already funnels through those three methods, so the invariant is enforced by the type rather than by caller locking. Each access is a lock-free atomic op, so there is no added lock and lock ordering is unchanged.

I kept the fix on the field's own type instead of holding `rwm.RLock()` across `Snapshot`'s sort. That sort's fair-sharing path calls `client.Get` per LocalQueue, and `totalElements()` already takes and releases `rwm.RLock` internally, so widening the lock would hold it across I/O and nest read locks.

Fixes #12532

Added `TestSnapshotConcurrentWithRequeueNoDataRace`, which runs `Snapshot()` concurrently with preemption requeues. It fails under `-race` without the fix (race at `set` / `matches`) and passes with it. `go test -race ./pkg/cache/queue/...` is green.

```release-note
VisibilityOnDemand: Fixed a data race between the Visibility API pending-workloads endpoint and preemption requeuing that could crash the queue manager for BestEffortFIFO ClusterQueues.
```

---

🤖 AI assistance disclosure (per the [Kubernetes AI Tool Usage Policy](https://www.kubernetes.dev/docs/guide/pull-requests/#ai-guidance)): this change was developed with AI assistance (Claude Code, Opus 4.8). The race was reproduced with `go test -race`, and I reviewed and validated the fix and the test locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved concurrency behavior when taking queue snapshots while workloads are being requeued, reducing the likelihood of race-related issues.
  * Made sticky workload tracking safer under concurrent access.

* **Tests**
  * Added a concurrency test that repeatedly snapshots the queue while simultaneously requeueing workloads, aimed at catching potential data races (including under race detection).

### DIFF
--- a/pkg/cache/queue/cluster_queue.go
+++ b/pkg/cache/queue/cluster_queue.go
@@ -22,6 +22,7 @@ import (
 	"maps"
 	"slices"
 	"sync"
+	"sync/atomic"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
@@ -80,29 +81,24 @@ var (
 // The workloadName field is accessed concurrently: Snapshot sorts a copy of
 // the pending workloads (via baseCompareFunc -> matches) without holding the
 // ClusterQueue's rwm lock, while RequeueIfNotPresent sets it during preemption.
-// It carries its own mutex so every read and write of the field is
-// synchronized regardless of the caller's locking.
+// It is a single field with only whole-value read/store/clear, so an
+// atomic.Pointer synchronizes every access regardless of the caller's locking
+// (nil means no sticky workload).
 type stickyWorkload struct {
-	mu           sync.RWMutex
-	workloadName workload.Reference
+	workloadName atomic.Pointer[workload.Reference]
 }
 
 func (s *stickyWorkload) matches(workload workload.Reference) bool {
-	s.mu.RLock()
-	defer s.mu.RUnlock()
-	return s.workloadName == workload
+	name := s.workloadName.Load()
+	return name != nil && *name == workload
 }
 
 func (s *stickyWorkload) clear() {
-	s.mu.Lock()
-	defer s.mu.Unlock()
-	s.workloadName = ""
+	s.workloadName.Store(nil)
 }
 
 func (s *stickyWorkload) set(workload workload.Reference) {
-	s.mu.Lock()
-	defer s.mu.Unlock()
-	s.workloadName = workload
+	s.workloadName.Store(&workload)
 }
 
 func logStickyWorkloadSelectionIfVerbose(log logr.Logger, wl *kueue.Workload) {
--- a/pkg/cache/queue/cluster_queue.go
+++ b/pkg/cache/queue/cluster_queue.go
@@ -76,19 +76,32 @@ var (
 // workloads from going back to the head of the queue.  A workload is
 // considered sticky until it is admitted, unschedulable, or deleted.
 // See Kueue#6929 and Kueue#7101 for motivation.
+//
+// The workloadName field is accessed concurrently: Snapshot sorts a copy of
+// the pending workloads (via baseCompareFunc -> matches) without holding the
+// ClusterQueue's rwm lock, while RequeueIfNotPresent sets it during preemption.
+// It carries its own mutex so every read and write of the field is
+// synchronized regardless of the caller's locking.
 type stickyWorkload struct {
+	mu           sync.RWMutex
 	workloadName workload.Reference
 }
 
 func (s *stickyWorkload) matches(workload workload.Reference) bool {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
 	return s.workloadName == workload
 }
 
 func (s *stickyWorkload) clear() {
+	s.mu.Lock()
+	defer s.mu.Unlock()
 	s.workloadName = ""
 }
 
 func (s *stickyWorkload) set(workload workload.Reference) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
 	s.workloadName = workload
 }
 
--- a/pkg/cache/queue/cluster_queue_test.go
+++ b/pkg/cache/queue/cluster_queue_test.go
@@ -478,6 +478,58 @@ func TestSnapshotStableWithConcurrentFSUpdates(t *testing.T) {
 	}
 }
 
+// TestSnapshotConcurrentWithRequeueNoDataRace guards against a data race on the
+// sticky workload: Snapshot sorts a copy of the pending workloads through the
+// comparator (which reads stickyWorkload.workloadName) without holding the
+// ClusterQueue lock, while RequeueIfNotPresent writes that field during a
+// BestEffortFIFO preemption requeue. Run with -race to detect regressions.
+func TestSnapshotConcurrentWithRequeueNoDataRace(t *testing.T) {
+	ctx, _ := utiltesting.ContextWithLog(t)
+	cq, err := newClusterQueue(ctx, nil,
+		&kueue.ClusterQueue{
+			Spec: kueue.ClusterQueueSpec{
+				QueueingStrategy: kueue.BestEffortFIFO,
+			},
+		},
+		defaultOrdering,
+		nil, nil, nil)
+	if err != nil {
+		t.Fatalf("failed to create ClusterQueue: %v", err)
+	}
+
+	// At least two workloads so Snapshot's sort actually invokes the
+	// comparator that reads the sticky workload.
+	wls := []*kueue.Workload{
+		utiltestingapi.MakeWorkload("wl1", defaultNamespace).Obj(),
+		utiltestingapi.MakeWorkload("wl2", defaultNamespace).Obj(),
+		utiltestingapi.MakeWorkload("wl3", defaultNamespace).Obj(),
+	}
+	for _, wl := range wls {
+		cq.PushOrUpdate(workload.NewInfo(wl))
+	}
+
+	// Writer: continuously set the sticky workload via a preemption requeue.
+	stop := make(chan struct{})
+	go func() {
+		for {
+			select {
+			case <-stop:
+				return
+			default:
+				for _, wl := range wls {
+					cq.RequeueIfNotPresent(ctx, workload.NewInfo(wl), RequeueReasonPendingPreemption)
+				}
+			}
+		}
+	}()
+	defer close(stop)
+
+	// Reader: Snapshot reads the sticky workload through the comparator.
+	for range 1000 {
+		cq.Snapshot()
+	}
+}
+
 func TestPendingResources(t *testing.T) {
 	ctx, log := utiltesting.ContextWithLog(t)
 	now := time.Now()


somaz94:
Thanks for the `/ok-to-test`, @mimowo! I've flipped this out of Draft — ready for review.

The core unit/integration jobs are green. The remaining red jobs are all Pod scheduling timeouts (Prow infra flakes, unrelated to this change) and should clear on retest.

pkg/cache/queue/cluster_queue.go
```diff
@@ -76,19 +76,32 @@ var (
...
 // considered sticky until it is admitted, unschedulable, or deleted.
 // See Kueue#6929 and Kueue#7101 for motivation.
+//
+// The workloadName field is accessed concurrently: Snapshot sorts a copy of
+// the pending workloads (via baseCompareFunc -> matches) without holding the
+// ClusterQueue's rwm lock, while RequeueIfNotPresent sets it during preemption.
+// It carries its own mutex so every read and write of the field is
+// synchronized regardless of the caller's locking.
 type stickyWorkload struct {
+	mu           sync.RWMutex
```
mimowo: This is non-blocking thought, maybe a premature optimization, but I'm wondering if we could give away with just `atomic.Pointer` which is a bit lighter.

That could matter if any of the functions were called in a hot loop, but they seem to be called only a handful of times per scheduling cycle, so I don't feel strongly here. Still, checking your opinion.

pkg/cache/queue/cluster_queue_test.go
```diff
@@ -478,6 +478,58 @@ func TestSnapshotStableWithConcurrentFSUpdates(t *testing.T) {
 	}
 }
 
+// TestSnapshotConcurrentWithRequeueNoDataRace guards against a data race on the
+// sticky workload: Snapshot sorts a copy of the pending workloads through the
+// comparator (which reads stickyWorkload.workloadName) without holding the
+// ClusterQueue lock, while RequeueIfNotPresent writes that field during a
+// BestEffortFIFO preemption requeue. Run with -race to detect regressions.
+func TestSnapshotConcurrentWithRequeueNoDataRace(t *testing.T) {
```
mimowo, line 486: Nice test, but how long does the unit test execute? asking just to strike the right balance between discovering issues and runtime.

mimowo (commented):
LGTM overall, thank you for the PR. A sanity look from @j-skiba or @kshalot would be great

pkg/cache/queue/cluster_queue.go
somaz94: Good call. Since `stickyWorkload` guards a single field with only whole-value read/store/clear (no compound read-modify-write to hold together), `atomic.Pointer[workload.Reference]` does the job and is lighter than the RWMutex, exactly as you suggested. Switched `matches`/`set`/`clear` over (nil = cleared) and dropped the mutex in fc45db3. Target race test + the full `pkg/cache/queue` suite are still green under `-race`.

pkg/cache/queue/cluster_queue_test.go
somaz94: Thanks! It's fast — `TestSnapshotConcurrentWithRequeueNoDataRace` runs in ~0.01s under `-race` on my machine. (The ~1.6s reported for the package is shared build/init overhead, not this test.) The reader does 1000 `Snapshot()` iterations against the concurrent writer goroutine, which reliably trips the race detector without adding meaningful runtime. Happy to dial the iteration count down if you'd prefer an even tighter budget.

j-skiba: maybe alternatively instead of running snapshots in a loop we could pass some fake client to `newClusterQueue`

```go
func (fc *fakeClient) Get(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
	if _, ok := obj.(*kueue.LocalQueue); ok {
		fc.notify() // notifies some goroutine to run RequeueIfNotPresent
	}
	return fc.Client.Get(ctx, key, obj, opts...)
}
```

and then run 2 goroutines:
1. runs snapshot
2. calls `RequeueIfNotPresent` that will be notified by `fc.notify()`

I think `-race` should be able to detect the race.

j-skiba: oh, but I see you've just posted it's rather fast, so maybe there's no need for that

mimowo (commented):
Thank you for hardening scheduling which is the core job of Kueue👍 

Leaving lgtm to @j-skiba 
/approve
/cherrypick release-0.18
/cherrypick release-0.17

pkg/cache/queue/cluster_queue.go
```diff
@@ -76,20 +77,28 @@ var (
...
+//
+// The workloadName field is accessed concurrently: Snapshot sorts a copy of
+// the pending workloads (via baseCompareFunc -> matches) without holding the
+// ClusterQueue's rwm lock, while RequeueIfNotPresent sets it during preemption.
+// It is a single field with only whole-value read/store/clear, so an
+// atomic.Pointer synchronizes every access regardless of the caller's locking
+// (nil means no sticky workload).
 type stickyWorkload struct {
-	workloadName workload.Reference
+	workloadName atomic.Pointer[workload.Reference]
```
kshalot, line 88: This will fix the data race, but isn't there another underlying issue here? If the sticky workload is changing in an unsynchronized manner and it is being used inside the comparator function for the workload heap and also for the snapshot sort, then wouldn't that lead to corruptions?

Here `RequeueIfNotPresent` is setting the SW without the lock:
https://github.com/kubernetes-sigs/kueue/blob/f1aeeb4150b0542d65ac3c8ab735282bb0a3756a/pkg/cache/queue/cluster_queue.go#L854-L864

kshalot: We had similar issues reported recently: https://github.com/kubernetes-sigs/kueue/issues/12534, but not in regards to sticky workloads.

mimowo: Maybe yes, but if so let's ticket that independently, I think this PR is already useful fixing the crash of Kueue, and ensuring the stability or sticky workload across heap comparisions has been discussed already when we introduced that. 

It is possible that we missed some risks, so thinking about problematic scenarios is not wasted time, but I wouldn't block the PR on that - this is a difficult topic. We brainstormed that at length with Gabe back then. We need some detailed scenarios to attack the problem

kshalot: Okay then. I'll create a follow-up ticket, maybe someone will have some time to look into it or maybe we run into this sorting issue, then we'll have a starting point for debugging.

Overall LGTM apart from this comment. I'll leave the final tag to @j-skiba

mimowo: >  I'll create a follow-up ticket, 

Thanks 👍 I certainly agree it is worth looking into. I guess our discussions with Gabe should have been documented better. In any case, a bug might be lurking there.

somaz94: Good point — I agree the concern is broader than the crash this PR fixes.

The `atomic.Pointer` makes each individual read/write of the sticky field race-free, but as you note the *value* can still change between the comparisons a single `SortFunc` pass makes: a workload can be sticky for one comparison and not for another within the same sort, so the ordering the comparator produces isn't guaranteed to be internally consistent. That's the same class of non-transitive-comparator hazard as #12534 (there it's the fair-sharing usage lookup; here it's the sticky-workload check), which under `slices.SortFunc` can panic on a strict-weak-ordering violation.

Per @mimowo's suggestion I'll keep this PR scoped to fixing the data-race crash and open a separate tracking issue for the "sticky state can mutate mid-sort" question, citing #12534 as a sibling case, so it can be worked with concrete scenarios. I'll link it here once filed.

somaz94:
The three failing jobs are `Pod scheduling timeout` infra flakes — the pods never scheduled (`0/99 nodes available: Insufficient cpu/memory`), no test binary ran, and the same integration jobs are erroring on unrelated PRs (e.g. #12734) in the same window. Retesting.

/retest

j-skiba:
/lgtm

Thank you for fixing the bug!

pkg/cache/queue/cluster_queue.go
somaz94: Ah, I see you've already offered to open the follow-up — happy to leave it to you rather than duplicate it. Whenever the ticket is up I can drop in the concrete scenario I have in mind (the sticky pointer flipping between two comparator calls within a single `SortFunc` pass, yielding a non-transitive order). Thanks!

kshalot: I wrote something down quickly here: https://github.com/kubernetes-sigs/kueue/issues/12740

Feel free to comment there or flag any inconsistencies 😄  Thanks!

mimowo:
@somaz94 could you please CP the PR to 0.17?

tenzen-y:
/release-note-edit

```release-note
VisibilityOnDemand: Fixed a data race between the Visibility API pending-workloads endpoint and preemption requeuing that could crash the queue manager for BestEffortFIFO ClusterQueues.
```

PR #12759, opened 2026-07-03
Add unadmitted workloads metrics

j-skiba:
#### What type of PR is this?
/kind feature

This PR introduces two new Prometheus metrics to track the number of unadmitted workloads in Kueue, allowing operators to monitor queue blockages and their causes:
- `kueue_unadmitted_workloads`: Tracks the count of unadmitted workloads per ClusterQueue.
- `kueue_local_queue_unadmitted_workloads`: Tracks the count of unadmitted workloads per LocalQueue

Both metrics partition the workloads by:
- `reason`: The reason from the `Admitted` condition (e.g., `NoReservation`).
- `underlying_cause`: The granular reason from the `QuotaReserved` condition (e.g., `WaitingForQuota`, `ChecksNotReady`, `PendingEvaluation`, `PendingTopology`).

These metrics are only active when the `UnadmittedWorkloadsObservability` feature gate is enabled.

Part of #10852

```release-note
Observability: Added `kueue_unadmitted_workloads` and `kueue_local_queue_unadmitted_workloads` metrics (gated by `UnadmittedWorkloadsObservability`) to track the count of unadmitted workloads by ClusterQueue/LocalQueue and the underlying blockage cause (e.g., `WaitingForQuota`, `ChecksNotReady`).
```

## Summary by CodeRabbit

* **New Features**
  * Added additional unadmission underlying causes.
  * Introduced Prometheus gauges for unadmitted workloads at cluster-queue and local-queue levels, with labels for queue identity, reason, underlying cause, and replica role.
* **Bug Fixes**
  * Improved end-to-end unadmitted status and metrics tracking, including correct updates and cleanup across queue/workload lifecycle events, and behavior when explicit unadmitted status is disabled.
* **Documentation / Tests**
  * Updated the metrics reference docs and added integration coverage validating reporting and pruning of metrics.


pkg/cache/queue/manager.go
```diff
@@ -990,3 +1087,150 @@ func (m *Manager) NotifyWorkloadUpdateWatchers(oldWorkload, newWorkload *kueue.W
 func (m *Manager) AddWorkloadUpdateWatcher(watcher WorkloadUpdateWatcher) {
 	m.workloadUpdateWatchers = append(m.workloadUpdateWatchers, watcher)
 }
+
+func getUnadmittedWorkloadStatus(wl *kueue.Workload) (unadmittedWorkloadStatus, bool) {
```
mimowo: nit, return *unadmittedWorkloadStatus if it may be nil

pkg/cache/queue/manager.go
```diff
@@ -155,6 +157,11 @@ type Manager struct {
 	workloadAssignedQueues map[workload.Reference]queue.LocalQueueReference
 	finishedWorkloads      map[workload.Reference]queue.LocalQueueReference
 
+	// Tracks unadmitted workload statuses and counts.
+	trackedUnadmittedWorkloadStatuses map[workload.Reference]unadmittedWorkloadStatus
+	unadmittedCQCounts                map[unadmittedCQStatus]int
+	unadmittedLQCounts                map[unadmittedWorkloadStatus]int
```
mimowo: I'm a bit concerned that tracking counts may result bugs when we call delete twice or add. 

We could make the operations idempotent, and so less susceptible to errors if we kept the sets per CQ and per LQ. If we track counts, then we could directly add / delete in the metrics - I'm not sure why duplicating the memory? 

If this is indeed need in this form - please add some comments to explain why

apis/kueue/v1beta2/workload_types.go
```diff
@@ -984,6 +984,15 @@ const (
...
 	// WorkloadAdmittedReasonPendingDelayedTopologyRequests indicates that there are pending delayed topology requests.
 	WorkloadAdmittedReasonPendingDelayedTopologyRequests = "PendingDelayedTopologyRequests"
+
+	// WorkloadUnadmittedCauseChecksNotReady indicates that the workload has reserved quota,
+	// but is blocked by pending admission checks.
+	WorkloadUnadmittedCauseChecksNotReady = "ChecksNotReady"
+
+	// WorkloadUnadmittedCausePendingTopology indicates that the workload has reserved quota,
+	// but is blocked by delayed topology requests.
+	WorkloadUnadmittedCausePendingTopology = "PendingTopology"
```
mimowo: This does not seem specific to adding metrics? Could we de-couple to a separate PR so that we don't mix responsibilities - adding new reasons and exposing in metrics.

mimowo: I imagine another prep PR where these are only added as Condition reason

j-skiba: these are metrics specific underlying_causes - https://github.com/kubernetes-sigs/kueue/tree/main/keps/10852-inadmissible-workloads-observability#prometheus-metrics-schema

pkg/cache/queue/manager.go
mimowo: The new functions in the manager look pretty involving without any logging - please add some logging at V2-V4 levels depending on how much anomaly are the conditions

pkg/cache/queue/manager.go
```diff
@@ -990,3 +1087,150 @@ func (m *Manager) NotifyWorkloadUpdateWatchers(oldWorkload, newWorkload *kueue.W
 func (m *Manager) AddWorkloadUpdateWatcher(watcher WorkloadUpdateWatcher) {
 	m.workloadUpdateWatchers = append(m.workloadUpdateWatchers, watcher)
 }
+
+func getUnadmittedWorkloadStatus(wl *kueue.Workload) (unadmittedWorkloadStatus, bool) {
+	admittedCond := apimeta.FindStatusCondition(wl.Status.Conditions, kueue.WorkloadAdmitted)
+	if (admittedCond != nil && admittedCond.Status == metav1.ConditionTrue) || workload.IsFinished(wl) || !wl.DeletionTimestamp.IsZero() {
```
mimowo: that is hard to parse - please add a comment and a logger message. Is stepping here anomaly? If so then reflect that with the v2 logging.

pkg/cache/queue/manager.go
```diff
@@ -990,3 +1087,150 @@ func (m *Manager) NotifyWorkloadUpdateWatchers(oldWorkload, newWorkload *kueue.W
...
+			return kueue.WorkloadUnadmittedCauseChecksNotReady
+		case kueue.WorkloadAdmittedReasonPendingDelayedTopologyRequests:
+			return kueue.WorkloadUnadmittedCausePendingTopology
+		default:
+			return reason
+		}
+	}
+	if quotaReservedCond != nil {
+		return quotaReservedCond.Reason
+	}
```
mimowo: Maybe combine them by the entry check `if quotaReservedCond != nil {` and only inside dispatch

pkg/cache/queue/manager.go
```diff
@@ -990,3 +1087,150 @@ func (m *Manager) NotifyWorkloadUpdateWatchers(oldWorkload, newWorkload *kueue.W
...
+	return kueue.WorkloadQuotaReservedReasonPendingEvaluation
+}
+
+func (m *Manager) UpdateUnadmittedWorkload(wl *kueue.Workload) {
+	m.Lock()
+	defer m.Unlock()
+	m.updateUnadmittedWorkloadWithoutLock(wl)
+}
+
+func (m *Manager) updateUnadmittedWorkloadWithoutLock(wl *kueue.Workload) {
```
mimowo: This one requires some logging - non trivial logic

pkg/cache/queue/manager.go
```diff
@@ -990,3 +1087,150 @@ func (m *Manager) NotifyWorkloadUpdateWatchers(oldWorkload, newWorkload *kueue.W
...
+
+func (m *Manager) incrementStatusCounts(status unadmittedWorkloadStatus) {
+	m.updateUnadmittedWorkloadMetric(status, 1)
+}
+
+func (m *Manager) decrementStatusCounts(status unadmittedWorkloadStatus) {
+	m.updateUnadmittedWorkloadMetric(status, -1)
+}
+
+func (m *Manager) updateUnadmittedWorkloadMetric(status unadmittedWorkloadStatus, diff int) {
```
mimowo: this one requires some logging too

mimowo (commented):
cc @kshalot ptal

pkg/controller/core/workload_controller.go
```diff
@@ -253,11 +253,24 @@ func (r *WorkloadReconciler) logger() logr.Logger {
 // +kubebuilder:rbac:groups=resource.k8s.io,resources=deviceclasses,verbs=get;list;watch
 // +kubebuilder:rbac:groups=resource.k8s.io,resources=resourceslices,verbs=get;list;watch
 
-func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, retErr error) {
```
kshalot, line 256:
```suggestion
func (r *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```

Do we need those variables? I don't see them used, maybe this is a leftover.

pkg/cache/queue/manager.go
j-skiba: > I'm a bit concerned that tracking counts may result bugs when we call delete twice or add.
> We could make the operations idempotent, and so less susceptible to errors if we kept the sets per CQ and per LQ. 

We get idempotent behavior by checking against `trackedUnadmittedWorkloadStatuses`.

> If we track counts, then we could directly add / delete in the metrics - I'm not sure why duplicating the memory?

I wanted to avoid O(N) paths:
- the current approach is O(1) for workload update event and prometheus scrape
- some kind of push, scan map on update would be O(N) for each workload update
- the add / delete (`Inc()`, `Dec()` ?) I think would leave us with `0` valued metrics, we would have to track them somehow anyway to remove them in the end
- custom collector - we could keep just `trackedUnadmittedWorkloadStatuses` to keep tract of workloads but it leaves us with O(N) on scrape, or O(1) but we would have to keep counts anyway

I don't know if I understood the question correctly

pkg/cache/queue/manager.go
```diff
@@ -955,6 +1026,32 @@ func (m *Manager) resyncLocalQueueGaugeMetricsLocked(lq *LocalQueue) {
...
+		if status.ClusterQueue == cqName {
+			cqCustomLabels := m.customLabels.CQGet(cqName)
+			metrics.ReportUnadmittedWorkload(status.ClusterQueue, status.Reason, status.UnderlyingCause, cqCustomLabels, m.roleTracker, count)
+		}
+	}
+}
+
+func (m *Manager) resyncLocalQueueUnadmittedWorkloadsMetricsLocked(lqRef queue.LocalQueueReference) {
+	lqNamespace, lqName := queue.MustParseLocalQueueReference(lqRef)
+	for status, count := range m.unadmittedLQCounts {
```
kshalot: Instead of performing a scan here and in the CQ case, we could use a nested map like `map[LqName]map[unadmittedWorkloadStatus]int` which would be an O(1) lookup during resync.

pkg/cache/queue/manager.go
```diff
@@ -187,17 +194,42 @@ type Manager struct {
...
-		finishedWorkloads:      make(map[workload.Reference]queue.LocalQueueReference),
+		clock:                             realClock,
+		client:                            client,
+		statusChecker:                     checker,
+		localQueues:                       make(map[queue.LocalQueueReference]*LocalQueue),
+		workloadAssignedQueues:            make(map[workload.Reference]queue.LocalQueueReference),
+		finishedWorkloads:                 make(map[workload.Reference]queue.LocalQueueReference),
+		trackedUnadmittedWorkloadStatuses: make(map[workload.Reference]unadmittedWorkloadStatus),
+		unadmittedCQCounts:                make(map[unadmittedCQStatus]int),
+		unadmittedLQCounts:                make(map[unadmittedWorkloadStatus]int),
```
kshalot: Nit: I don't love those names, because they suggest that we are tracking the counts of "unadmitted CQ/LQs". Maybe smth like `unadmittedWorkloadsPerCQ`?

pkg/controller/core/workload_controller.go
j-skiba: it's used there https://github.com/kubernetes-sigs/kueue/pull/12759/changes#diff-60dd240c20adbd6a189d018d1c216c2d296730f446c341d8bf449fa6657964ffR262

```go
			if getErr == nil && retErr == nil {
				r.queues.UpdateUnadmittedWorkload(&wl)
			} else if apierrors.IsNotFound(getErr) {
				r.queues.RemoveUnadmittedWorkload(wlKey)
			}
```

pkg/cache/queue/manager.go
j-skiba: `unadmittedWorkloadsPerLQ` and `unadmittedWorkloadsPerCQ` sounds less confusing, renamed

pkg/cache/queue/manager.go
```diff
@@ -572,6 +628,11 @@ func (m *Manager) DeleteLocalQueue(log logr.Logger, q *kueue.LocalQueue) {
 		clearLQMetrics(key)
 	}
 	delete(m.localQueues, key)
+	if features.Enabled(features.UnadmittedWorkloadsObservability) {
+		for _, wInfo := range qImpl.items {
+			m.updateUnadmittedWorkloadWithoutLock(wInfo.Obj)
+		}
+	}
```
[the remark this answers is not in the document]
j-skiba: fix applied

pkg/cache/queue/manager.go
j-skiba: fixed

j-skiba: I added some logs in all functions

pkg/cache/queue/manager.go
j-skiba: I refactored the check to
```go
if workload.IsAdmitted(wl) || workload.IsFinished(wl) || !wl.DeletionTimestamp.IsZero() {
```
I think it's less confusing now, also added some logging if the check fails
```go
klog.V(4).InfoS("Workload is not unadmitted",
			"workload", klog.KObj(wl),
			"isAdmitted", workload.IsAdmitted(wl),
             "isFinished", workload.IsFinished(wl),
			"isDeleting", !wl.DeletionTimestamp.IsZero(),
            )

File: pkg/cache/queue/manager.go (hunk 6 above)
Comment (j-skiba) [reply pkg/cache/queue/manager.go]: I added some logging here too

File: pkg/cache/queue/manager.go (hunk 7 above)
Comment (j-skiba) [reply pkg/cache/queue/manager.go]: done

File: site/content/en/docs/reference/metrics.md [hunk 12]
```diff
@@ -56,6 +56,7 @@ Use the following metrics to monitor the status of your ClusterQueues:
 | `kueue_quota_reserved_workloads_total` | Counter | The total number of quota reserved workloads per 'cluster_queue' | `cluster_queue`: the name of the ClusterQueue<br> `priority_class`: the priority class name<br> `replica_role`: one of `leader`, `follower`, or `standalone` |
 | `kueue_replaced_workload_slices_total` | Counter | The number of replaced workload slices per 'cluster_queue' | `cluster_queue`: the name of the ClusterQueue<br> `replica_role`: one of `leader`, `follower`, or `standalone` |
 | `kueue_reserving_active_workloads` | Gauge | The number of Workloads that are reserving quota, per 'cluster_queue' | `cluster_queue`: the name of the ClusterQueue<br> `replica_role`: one of `leader`, `follower`, or `standalone` |
+| `kueue_unadmitted_workloads` | Gauge | The number of unadmitted workloads, per 'cluster_queue', 'reason', and 'underlying_cause'. | `cluster_queue`: the name of the ClusterQueue<br> `reason`: the reason why the workload is not admitted<br> `underlying_cause`: the underlying cause for the quota reservation deficit<br> `replica_role`: one of `leader`, `follower`, or `standalone` |
```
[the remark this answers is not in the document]
j-skiba: it should be fixed now

pkg/cache/queue/manager.go
j-skiba: 👍 done

apis/kueue/v1beta2/workload_types.go
mimowo: Ah, sorry - that escaped my attention on the KEP review, two things we need to change:
1. avoid the translation mapping between `For UnsatisfiedAdmissionChecks: Set to ChecksNotReady.` - such things make the undersanding much harder - for example users will no longer be able to easily map metrics to condition
2. revisit the roles of `reason` and `underlying_cause`.the underlying cause should be adding more granularity to the "reason". If the reason says "UnsatisfiedAdmissionChecks" , then I see no point setting "underlying_cause" ChecksNotReady - this is 1:1 mapping which does not add any granularity.

pkg/cache/queue/manager.go
mimowo: > We get idempotent behavior by checking against trackedUnadmittedWorkloadStatuses.

I see - thanks for exaplining. I didn't check yet closely in the code that you ensure the idempotence by trackedUnadmittedWorkloadStatuses. However, in that case for readability I would like to encapsualte the structers into `unadmittedWorkloads`, so that it is an internal detail how we represent things, and how we ensure idempotence

> I wanted to avoid O(N) paths:

Sure, I was thinking about the same just keeping cqUnadmittedWorkloads cqName -> workloadRefs, so that we can compute `Len() ` etc in 0(1) too. Your approach is working fine too, but I would like to just encapsulate the potentially tricky logic behind a struct (mentioned above).

> I don't know if I understood the question correctly

I think you got it right - I wanted to probe why thigs are done this way to better understand how to improve on readability, while keeping O(1), and indempotence. I think encapsulating behind a strurct is my current thinking.

pkg/cache/queue/manager.go
```diff
@@ -990,3 +1083,177 @@ func (m *Manager) NotifyWorkloadUpdateWatchers(oldWorkload, newWorkload *kueue.W
...
+	return &unadmittedWorkloadStatus{
+		LocalQueueName:      wl.Spec.QueueName,
+		LocalQueueNamespace: wl.Namespace,
+		Reason:              reason,
+		UnderlyingCause:     underlyingCause,
+	}
+}
+
+func resolveUnadmittedUnderlyingCause(reason string, quotaReservedCond *metav1.Condition) string {
+	if quotaReservedCond != nil {
```
kshalot: We could flip this and reduce the indentation:
```go
	if quotaReservedCond == nil {
	    // If the scheduler has not evaluated the workload yet, or if the UnadmittedWorkloadsExplicitStatus
	    // feature gate is disabled (meaning conditions are not populated), the QuotaReserved condition
    	// will be missing. We assume the workload is pending evaluation.
	    return kueue.WorkloadQuotaReservedReasonPendingEvaluation
    }
    ...
```

Because every branch in the if-branch has a return statement anyway

pkg/cache/queue/manager.go
```diff
@@ -990,3 +1083,177 @@ func (m *Manager) NotifyWorkloadUpdateWatchers(oldWorkload, newWorkload *kueue.W
...
+		} else {
+			klog.V(3).InfoS("Failed to resolve ClusterQueue for unadmitted workload", "workload", wlKey, "queue", wl.Spec.QueueName)
+		}
+		if ok && oldStatus == *status {
+			klog.V(5).InfoS("Workload unadmitted status unchanged, skipping update", "workload", wlKey)
+			return
+		}
+	}
+
+	klog.V(4).InfoS("Updating unadmitted workload tracking", "workload", wlKey, "oldStatus", oldStatus, "newStatus", status)
```
kshalot: Hm, I don't think we should use `klog` directly, but `logr.Logger`, which AFAIK allows to swap the underlying logging engine.

pkg/cache/queue/manager.go
```diff
@@ -990,3 +1083,177 @@ func (m *Manager) NotifyWorkloadUpdateWatchers(oldWorkload, newWorkload *kueue.W
...
+
+func (m *Manager) RemoveUnadmittedWorkload(wlKey workload.Reference) {
+	m.Lock()
+	defer m.Unlock()
+	m.removeUnadmittedWorkloadWithoutLock(wlKey)
+}
+
+func (m *Manager) removeUnadmittedWorkloadWithoutLock(wlKey workload.Reference) {
+	oldStatus, ok := m.trackedUnadmittedWorkloadStatuses[wlKey]
+	if ok {
```
kshalot: Nit: I'm a big fan of guard clauses, but this is stylistic preference so feel free to ignore:
```go
if !ok {
    klog.V(4).InfoS("Workload to remove was not tracked as unadmitted", "workload", wlKey)
    return
}
...
```

apis/kueue/v1beta2/workload_types.go
mimowo: unresolve, because it doesn't seem addressed, let's discuss how we can address that first

j-skiba: https://github.com/kubernetes-sigs/kueue/pull/12814 opened PR with proposed update. I think it should be enough to leave the `underlying_reason` label empty in case we don't need to carry any granularity in that field

pkg/cache/queue/unadmitted_workloads.go
```diff
@@ -0,0 +1,282 @@
...
+func newUnadmittedWorkloads() *unadmittedWorkloads {
+	return &unadmittedWorkloads{
+		statuses: make(map[workload.Reference]unadmittedWorkloadStatus),
+		perCQ:    make(map[unadmittedCQStatus]int),
+		perLQ:    make(map[unadmittedWorkloadStatus]int),
+	}
+}
+
+func (u *unadmittedWorkloads) update(wl *kueue.Workload, m *Manager) {
+	log := ctrl.Log.WithName("queue-manager")
```
mimowo: Propagate log you have in the context

pkg/cache/queue/unadmitted_workloads.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+	if status != nil {
+		u.statuses[wlKey] = *status
+		u.incrementStatusCounts(*status, m)
+	}
+}
+
+func (u *unadmittedWorkloads) remove(wlKey workload.Reference, m *Manager) {
+	oldStatus, ok := u.statuses[wlKey]
+	log := ctrl.Log.WithName("queue-manager")
```
mimowo: Propagate log you have in the context

pkg/cache/queue/unadmitted_workloads.go
```diff
@@ -0,0 +1,282 @@
...
+func (u *unadmittedWorkloads) incrementStatusCounts(status unadmittedWorkloadStatus, m *Manager) {
+	u.updateUnadmittedWorkloadMetric(status, 1, m)
+}
+
+func (u *unadmittedWorkloads) decrementStatusCounts(status unadmittedWorkloadStatus, m *Manager) {
+	u.updateUnadmittedWorkloadMetric(status, -1, m)
+}
+
+func (u *unadmittedWorkloads) updateUnadmittedWorkloadMetric(status unadmittedWorkloadStatus, diff int, m *Manager) {
+	log := ctrl.Log.WithName("queue-manager")
```
mimowo: Propagate log you have in the context

pkg/cache/queue/unadmitted_workloads.go
```diff
@@ -0,0 +1,282 @@
+/*
+Copyright The Kubernetes Authors.
```
mimowo, line 2: Thank you for extracting the logic to a new file - this will offload the main files which is great in the long run for readability 🥇

pkg/cache/queue/unadmitted_workloads.go
```diff
@@ -0,0 +1,282 @@
...
+//   - statuses (workload.Reference -> unadmittedWorkloadStatus) records the current unadmitted
+//     status of each workload. Checking against tracked statuses guarantees idempotent behavior:
+//     repeated updates or deletes for the same workload status do not duplicate metric increments/decrements.
+//   - perCQ and perLQ maintain aggregated counts in memory to ensure O(1) processing on workload
+//     events and O(1) lookups during Prometheus metric scrapes, avoiding O(N) scans across workloads.
+type unadmittedWorkloads struct {
+	statuses map[workload.Reference]unadmittedWorkloadStatus
+	perCQ    map[unadmittedCQStatus]int
+	perLQ    map[unadmittedWorkloadStatus]int
+}
```
mimowo, line 64: Sanity check: don't wee need a R/W lock to ensure consistency of the struct in case of concurrent access / modification? I haven't checked that carefully yet, but it seems to me we may be accessing the structure from workload_controller and scheduler, and so the access may not be synced. 

In case the access is safe please add a comment why the lock is not needed.

j-skiba: the locks are currently on `manager.go` level, I think it would be better to add separate locks here. I will update that

mimowo: agree

mimowo: I think the manager lock is too large for the most cases and overused, so we may consider in the future more granular locks. Introducing the lock at this struct is a step in this direction, even if currently the access is protected by manager lock.

apis/kueue/v1beta2/workload_types.go
j-skiba: PR is merged, and the underlying_reasons are set to empty strings now for these Admitted False resons

pkg/cache/queue/manager.go
j-skiba: done

pkg/cache/queue/manager.go
j-skiba: replaced

pkg/cache/queue/manager.go
```diff
@@ -347,7 +347,7 @@ func (m *Manager) AddClusterQueue(ctx context.Context, cq *kueue.ClusterQueue) e
 			cqImpl.addLocalQueue(queue.Key(&q))
 			if features.Enabled(features.UnadmittedWorkloadsObservability) {
 				for _, wInfo := range qImpl.items {
-					m.updateUnadmittedWorkloadWithoutLock(wInfo.Obj)
+					m.updateUnadmittedWorkloadWithoutLock(ctx, wInfo.Obj)
```
mimowo: Any reason for passing the entire context? This is ok-ish but I would prefer to start with just logger to make it more light-weight. That decision has been already discussed and thus in most cases in the cache we pass just the logger.

mimowo: The main reasons:
1. avoids the common repackaging of the ctx int logger
2. full ctx is rarely useful in cache because cache normally shouldn't issue requests

pkg/cache/queue/manager.go
```diff
@@ -435,7 +445,7 @@ func (m *Manager) RebuildClusterQueue(cq *kueue.ClusterQueue, lqName string) err
 	return nil
 }
 
-func (m *Manager) DeleteClusterQueue(cq *kueue.ClusterQueue) {
+func (m *Manager) DeleteClusterQueue(ctx context.Context, cq *kueue.ClusterQueue) {
```
mimowo: Let's pass logger here only for now if ctx is not needed

pkg/cache/queue/manager.go
```diff
@@ -991,3 +1043,32 @@ func (m *Manager) NotifyWorkloadUpdateWatchers(oldWorkload, newWorkload *kueue.W
...
+}
+
+func (m *Manager) updateUnadmittedWorkloadWithoutLock(log logr.Logger, wl *kueue.Workload) {
+	cqName, _ := m.ClusterQueueForWorkloadWithoutLock(wl)
+	m.unadmittedWorkloads.update(log, wl, cqName, m.LocalQueueExistsWithoutLock, m)
+}
+
+func (m *Manager) RemoveUnadmittedWorkload(ctx context.Context, wlKey workload.Reference) {
+	m.unadmittedWorkloads.remove(ctrl.LoggerFrom(ctx), wlKey, m.LocalQueueExists, m)
+}
```
mimowo: Let's update the functions to accept just the logger for now, as ctx is not needed.

pkg/controller/core/workload_controller.go
mimowo, line 256: Would it be hard to avoid declaring the values in the return signature? I know golang supports that, but this is notoriously tricky to read, especially if most of the codebase is not using that. Since we don't use the syntax in Kueue much I would prefer to avoid it for consistency. 

OTOH this is a perfectly valid syntax so let me know if it gets tricky or messy to avoid it.

pkg/cache/queue/manager.go
j-skiba: done

pkg/cache/queue/manager.go
j-skiba: done

pkg/controller/core/workload_controller.go
j-skiba: we would have to make sure all errors that happen inside Reconcile are assigned to some `err` var defined at the beginning of the method. I think that would be error prone. Also the `retError` is used only in one place

pkg/cache/queue/unadmitted_workloads.go
j-skiba: lock added to `unadmittedWorkloads`

pkg/cache/queue/manager.go
```diff
@@ -484,6 +503,10 @@ func (m *Manager) addLocalQueueLocked(ctx context.Context, q *kueue.LocalQueue)
 	for _, w := range workloads.Items {
 		m.assignWorkload(workload.Key(&w), qImpl.Key)
 
+		if features.Enabled(features.UnadmittedWorkloadsObservability) {
+			m.updateUnadmittedWorkloadWithoutLock(ctrl.LoggerFrom(ctx), &w)
```
mimowo, line 507: Include workload in the logger, as done here: https://github.com/kubernetes-sigs/kueue/pull/12759/changes#diff-c7ed976ece2ea8cfd72847c76a214be5920b16aa7770e4d3cef7d8c974a36028L527

pkg/cache/queue/manager.go
```diff
@@ -445,6 +455,15 @@ func (m *Manager) DeleteClusterQueue(cq *kueue.ClusterQueue) {
 	cqName := kueue.ClusterQueueReference(cq.Name)
 	m.hm.DeleteClusterQueue(cqName)
 	clearCQMetrics(cqName)
+	if features.Enabled(features.UnadmittedWorkloadsObservability) {
+		for _, lq := range m.localQueues {
+			if lq.ClusterQueue == cqName {
+				for _, wInfo := range lq.items {
+					m.updateUnadmittedWorkloadWithoutLock(log, wInfo.Obj)
```
mimowo, line 462: Include workload in the logger, as done here: https://github.com/kubernetes-sigs/kueue/pull/12759/changes#diff-c7ed976ece2ea8cfd72847c76a214be5920b16aa7770e4d3cef7d8c974a36028L527

pkg/cache/queue/manager.go
```diff
@@ -341,6 +345,12 @@ func (m *Manager) AddClusterQueue(ctx context.Context, cq *kueue.ClusterQueue) e
 			added := cqImpl.AddFromLocalQueue(qImpl, m.roleTracker, m.customLabels)
 			addedWorkloads = addedWorkloads || added
 			cqImpl.addLocalQueue(queue.Key(&q))
+			if features.Enabled(features.UnadmittedWorkloadsObservability) {
+				log := ctrl.LoggerFrom(ctx)
+				for _, wInfo := range qImpl.items {
+					m.updateUnadmittedWorkloadWithoutLock(log, wInfo.Obj)
```
mimowo, line 351: I'm wondering if we should commonize the iteration with https://github.com/kubernetes-sigs/kueue/pull/12759/changes#diff-c7ed976ece2ea8cfd72847c76a214be5920b16aa7770e4d3cef7d8c974a36028R461

pkg/controller/core/workload_controller.go
mimowo: ack

pkg/cache/queue/manager.go
j-skiba: added `WithValues("workload...` to
```go
func (u *unadmittedWorkloads) update(
	log logr.Logger,
	wl *kueue.Workload,
	cqName kueue.ClusterQueueReference,
	qExists func(queue.LocalQueueReference) bool,
	m *Manager,
) {
```

pkg/cache/queue/manager.go
j-skiba: added `WithValues("workload...` to
```go
func (u *unadmittedWorkloads) update(
	log logr.Logger,
	wl *kueue.Workload,
	cqName kueue.ClusterQueueReference,
	qExists func(queue.LocalQueueReference) bool,
	m *Manager,
) {
```

pkg/cache/queue/manager.go
j-skiba: done

pkg/cache/queue/unadmitted_workloads.go
j-skiba: done

pkg/cache/queue/unadmitted_workloads.go
j-skiba: done

pkg/cache/queue/unadmitted_workloads.go
j-skiba: done

pkg/cache/queue/manager.go
j-skiba: `DeleteClusterQueue` runs loop over local queues from in memory list or all local queues while here in `AddClusterQueue` we get local queues from `List` call with indexer so we don't have to do filtering like in `DeleteClusterQueue`. I think we would have to sacrifice something here or there to commonize the code.

j-skiba:
/test pull-kueue-test-e2e-certmanager-main
/test pull-kueue-test-e2e-dra-counter-main

flaky integration test - https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12759/pull-kueue-test-integration-shard-1-main/2074483999039819776

@coderabbitai create a flaky test issue for this

`Scheduler Suite: [It] Workload controller with scheduler when the queue has admission check strategies the workload should have appropriate AdditionalChecks added [slow]`

pkg/cache/queue/manager.go
j-skiba: resync should not be run that frequently and we would have to move to more nested structure, also it's just a scan so shouldn't be a heavy operation. Do you think it's worth pursuing this approach or is this ok to leave it as it?

j-skiba:
[27d1572](https://github.com/kubernetes-sigs/kueue/pull/12759/commits/27d1572783d46c844a5192200dfc37f7c7e4877c)

I also added a check to not count variant workloads in the metrics. I noticed that when preparing another PR

mimowo (commented):
/lgtm
/approve
Thank you, nice to see this feature landing 👍

PR #12768, opened 2026-07-03
Don't admit workload when WaitingForReplacementPods=True.

### LINKED ISSUE #12656: Workloads with WaitingForReplacementPods=True are re-admitted, holding quota indefinitely
## What happened

  On a Kueue v0.16.9 cluster, using podset integration, we observed that if some of the pods of a workload are in the failed state, and waiting for replacement i.e. workload condition WaitingForReplacement is True, Kueue continues to evict and re-admit the workload (even though all the replacement pods are not created). This happens when the podset controller stops creating replacement pods for some workloads due to various possible reasons or failures, in which case we expect the admissions to eventually stop. However, the workload goes into a sustained cycle of Admission -> PodReadyTimeout -> backoff -> Readmission up to the re-queueing limit.

The cycle with Failed pods is somewhat like this:

  1. Workload is `Admitted=True`, quota reserved.
  2. Pods failed (possibly after a previous preemption)
  3. No replacement pods are created due to controller limit; `WaitingForReplacementPods=True` is set 
  4. `waitForPodsReady.timeout` elapses → eviction (`PodsReadyTimeout`); quota released.
  5. After backoff, the Workload is re-admitted and reserves quota again — despite the absence of some of the replacement pods.
  6. Repeats 1-5

Note: The previous failed pod is stuck in Failed state and not cleaned up due to the Kueue finalizer on the pods.

 The result: a meaningful share of the ClusterQueue's quota was held by Workloads with no running pods, blocking otherwise admissible Workloads in the same flavor. The cycle does not self-terminate until
  `requeuingStrategy.backoffLimitCount` is reached, which represents a substantial window of degraded capacity utilization.

  ## What we expected to happen

  Re-admission should not occur while `WaitingForReplacementPods=True`. The same invariant is already enforced at *initial* admission for the Pod integration — a Workload is not admitted until its pods exist — and applying it on re-admission would prevent the deadlock.

  ## Recommendation

  Gate re-admission on `WaitingForReplacementPods` being `False` or absent. 

  ## Environment

  - Kueue version: v0.16.9
  - Integration: Pod (pod-group)
  - Config: `waitForPodsReady.enable: true`, `timeout: 30m`, `blockAdmission: false`, `requeuingStrategy.backoffLimitCount: 50`

mbobrovskyi:
#### What type of PR is this?
/kind bug

Fixes #12656

```release-note
PodGroup integration: Fixed a bug that allowed Workloads corresponding to PodGroups with the `WaitingForReplacementPods=True` condition to be re-admitted immediately.
```

## Summary by CodeRabbit

* **New Features**
  * Added a new workload condition type for “waiting for replacement pods”.
* **Bug Fixes**
  * Prevented workloads from re-entering queues after backoff when they’re no longer admissible.
  * Blocked admission while “waiting for replacement pods” is true.
  * Improved pod replacement handling for fast-admission scenarios, including more accurate absent-pod counting.
* **Tests**
  * Updated unit, e2e, and integration assertions to use the shared condition type constant.
  * Added a helper to set pod phases by object keys for broader test coverage.

### DIFF
--- a/apis/kueue/v1beta2/workload_types.go
+++ b/apis/kueue/v1beta2/workload_types.go
@@ -1016,6 +1016,10 @@ const (
 	// WorkloadDeactivationTarget means that the Workload should be deactivated.
 	// This condition is temporary, so it should be removed after deactivation.
 	WorkloadDeactivationTarget = "DeactivationTarget"
+
+	// WorkloadWaitingForReplacementPods means that Kueue doesn't observe all
+	// the Pods declared for the group.
+	WorkloadWaitingForReplacementPods = "WaitingForReplacementPods"
 )
 
 // Reasons for the WorkloadPreemptionBlocked condition.
--- a/pkg/controller/jobs/pod/pod_controller.go
+++ b/pkg/controller/jobs/pod/pod_controller.go
@@ -75,10 +75,6 @@ const (
 )
 
 const (
-	// WorkloadWaitingForReplacementPods is True when Kueue doesn't observe all
-	// the Pods declared for the group.
-	WorkloadWaitingForReplacementPods = "WaitingForReplacementPods"
-
 	// WorkloadPodsFailed means that at least one Pod are not runnable or not succeeded.
 	WorkloadPodsFailed = "PodsFailed"
 )
@@ -712,8 +708,13 @@ func (p *Pod) Load(ctx context.Context, c client.Client, key *types.NamespacedNa
 	return !p.isFound, nil
 }
 
+// fastAdmission determines if the pod is configured for fast admission based on specific annotations.
+func (p *Pod) fastAdmission() bool {
+	return p.pod.GetAnnotations()[podconstants.GroupFastAdmissionAnnotationKey] == podconstants.GroupFastAdmissionAnnotationValue
+}
+
 func (p *Pod) constructGroupPodSets() ([]kueue.PodSet, error) {
-	if _, useFastAdmission := p.pod.GetAnnotations()[podconstants.GroupFastAdmissionAnnotationKey]; useFastAdmission {
+	if p.fastAdmission() {
 		tc, err := p.groupTotalCount()
 		if err != nil {
 			return nil, err
@@ -827,9 +828,8 @@ func (p *Pod) validatePodGroupMetadata(r events.EventRecorder, activePods []core
 	}
 
 	originalQueue := jobframework.QueueName(p)
-	_, useFastAdmission := p.pod.GetAnnotations()[podconstants.GroupFastAdmissionAnnotationKey]
 
-	if !useFastAdmission && len(activePods) < groupTotalCount {
+	if !p.fastAdmission() && len(activePods) < groupTotalCount {
 		errMsg := fmt.Sprintf("'%s' group has fewer runnable pods than expected", utilpod.GetPodGroupName(&p.pod))
 		r.Eventf(p.Object(), nil, corev1.EventTypeWarning, jobframework.ReasonErrWorkloadCompose, "ErrWorkloadCompose", errMsg)
 		return jobframework.UnretryableError(errMsg)
@@ -1242,9 +1242,7 @@ func (p *Pod) FindMatchingWorkloads(ctx context.Context, c client.Client, r even
 			return nil, nil, fmt.Errorf("failed to calculate pod role hash: %w", errors.Join(roleHashErrors...))
 		}
 
-		if absentCount := int(ps.Count) - len(roleActivePods); absentCount > 0 {
-			absentPods += absentCount
-		}
+		absentPods += p.countAbsentPods(ps, len(roleActivePods))
 
 		if excessCount := len(roleActivePods) - int(ps.Count); excessCount > 0 {
 			sortActivePods(roleActivePods)
@@ -1291,6 +1289,16 @@ func (p *Pod) FindMatchingWorkloads(ctx context.Context, c client.Client, r even
 		return nil, nil, err
 	}
 	return workload, []*kueue.Workload{}, nil
+}
+
+func (p *Pod) countAbsentPods(ps kueue.PodSet, activePods int) int {
+	if p.fastAdmission() {
+		if ps.Count > 0 && activePods == 0 {
+			return 1
+		}
+		return 0
+	}
+	return max(0, int(ps.Count)-activePods)
 }
 
 func (p *Pod) equivalentToWorkload(wl *kueue.Workload, jobPodSets []kueue.PodSet) bool {
@@ -1384,7 +1392,7 @@ func (p *Pod) waitingForReplacementPodsCondition(wl *kueue.Workload) (*metav1.Co
 		return nil, false
 	}
 
-	replCond := apimeta.FindStatusCondition(wl.Status.Conditions, WorkloadWaitingForReplacementPods)
+	replCond := apimeta.FindStatusCondition(wl.Status.Conditions, kueue.WorkloadWaitingForReplacementPods)
 	replCondStatus := p.absentPods > 0
 
 	// Nothing to change.
@@ -1396,7 +1404,7 @@ func (p *Pod) waitingForReplacementPodsCondition(wl *kueue.Workload) (*metav1.Co
 
 	if replCond == nil {
 		replCond = &metav1.Condition{
-			Type: WorkloadWaitingForReplacementPods,
+			Type: kueue.WorkloadWaitingForReplacementPods,
 		}
 		updated = true
 	} else {
--- a/test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
+++ b/test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
@@ -899,7 +899,7 @@ var _ = ginkgo.Describe("Pod controller", ginkgo.Label("job:pod", "area:jobs"), 
 						g.Expect(k8sClient.Get(ctx, wlLookupKey, createdWorkload)).To(gomega.Succeed())
 						g.Expect(createdWorkload.Status.Conditions).To(gomega.ContainElements(
 							gomega.BeComparableTo(metav1.Condition{
-								Type:    podcontroller.WorkloadWaitingForReplacementPods,
+								Type:    kueue.WorkloadWaitingForReplacementPods,
 								Status:  metav1.ConditionTrue,
 								Reason:  "ByTest",
 								Message: "by test",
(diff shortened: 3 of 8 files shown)


apis/kueue/v1beta2/workload_types.go
```diff
@@ -1016,6 +1016,10 @@ const (
 	// WorkloadDeactivationTarget means that the Workload should be deactivated.
 	// This condition is temporary, so it should be removed after deactivation.
 	WorkloadDeactivationTarget = "DeactivationTarget"
+
+	// WorkloadWaitingForReplacementPods means that Kueue doesn't observe all
+	// the Pods declared for the group.
+	WorkloadWaitingForReplacementPods = "WaitingForReplacementPods"
```
mimowo, line 1022: Could we reuse the OnHold reason? It seems like similar case, I'm not sure if we need more genularity here, wdyt?

mimowo: Ah, waiting for replacement Pods is already a pre-existing condition? In that case I think it is ok. 
What is the reason for the QuotaReserved=False in that case?

I think we may consider either `WaitingForReplacementPods`, or just `OnHold`.

cc @j-skiba who is currently working on improving the observabiilty of the non-admitted workloads, any opinion here?

mbobrovskyi: > Could we reuse the OnHold reason? It seems like similar case, I'm not sure if we need more genularity here, wdyt?

I just moved this reason to the API package to avoid a cyclic dependency.

mbobrovskyi: > I think we may consider either WaitingForReplacementPods, or just OnHold.

I'm actually not sure do we need to change logic in pod-groups.

mimowo: Maybe, I have not investiated the PR in detail yet. 

Maybe, could you just check what is currently the reason for QuotaReserved=False in this case?

mbobrovskyi: From the issue https://github.com/kubernetes-sigs/kueue/issues/12656, I can see that the problem is related to re-admitting a workload when WaitingForReplacementPods=True. This happens because we don't currently handle this case. We should avoid admitting workloads while WaitingForReplacementPods=True.

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
```diff
@@ -2455,6 +2455,134 @@ var _ = ginkgo.Describe("Pod controller interacting with Workload controller whe
 	})
 })
 
+var _ = ginkgo.Describe("Pod controller interacting with scheduler when waitForPodsReady enabled", ginkgo.Label("job:pod", "area:jobs"), ginkgo.Ordered, ginkgo.ContinueOnFailure, func() {
```
mimowo, line 2458: One more thing we may need to test is the fast-admission annotation. That does not need to wait for all pods to be replaced, one is enough.

apis/kueue/v1beta2/workload_types.go
j-skiba: I understand that `QuotaReserved` is kept set to `True` most of the time when `WaitingForReplacementPods` condition is set when Pods simply fail.

When workload is being evicted the `QuotaReserved` is still set to `True`.

Only when pods are fully terminated the generic job controller calls `UnsetQuotaReservationWithCondition` but then I think the reason should be `PendingEvaluation`.

Btw I haven't noticed when `OnHold` was added, we should include that in precedence rules for `QuotaReserved` reasons. I will update the KEP.

mimowo: > From the issue https://github.com/kubernetes-sigs/kueue/issues/12656, I can see that the problem is related to re-admitting a workload when WaitingForReplacementPods=True. This happens because we don't currently handle this case. We should avoid admitting workloads while WaitingForReplacementPods=True.

Yes, but when we have `WaitingForReplacementPods=True`, and we are waiting for re-admission, what is currently the reason for the `QuotaReserved=False` condition?

> I understand that QuotaReserved is kept set to True most of the time when WaitingForReplacementPods condition is set when Pods simply fail.

Ah, yes, but partly - now I also found the answer to my question in the issue descriptoin: https://github.com/kubernetes-sigs/kueue/issues/12656. 

"waitForPodsReady.timeout elapses → eviction (PodsReadyTimeout); quota released." - so I suppose we observe the reason as `PodsReadyTimeout`

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
mbobrovskyi: Done

pkg/controller/jobs/pod/pod_controller.go
```diff
@@ -1242,8 +1242,14 @@ func (p *Pod) FindMatchingWorkloads(ctx context.Context, c client.Client, r even
...
-			absentPods += absentCount
+		if p.fastAdmission() {
+			if ps.Count > 0 && len(roleActivePods) == 0 {
+				absentPods++
+			}
+		} else {
+			if absentCount := int(ps.Count) - len(roleActivePods); absentCount > 0 {
+				absentPods += absentCount
+			}
 		}
```
mimowo: Nice, let's move that to a helper function

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
```diff
@@ -2455,6 +2455,193 @@ var _ = ginkgo.Describe("Pod controller interacting with Workload controller whe
...
+				gomega.Consistently(func(g gomega.Gomega) {
+					wl := utiltestingapi.MakeWorkload(wlKey.Name, wlKey.Namespace).Obj()
+					g.Expect(k8sClient.Get(ctx, wlKey, wl)).Should(gomega.Succeed())
+					g.Expect(workload.IsActive(wl)).Should(gomega.BeTrue())
+					g.Expect(wl.Status.Conditions).Should(utiltesting.HaveConditionStatusFalse(kueue.WorkloadQuotaReserved))
+					g.Expect(wl.Status.Conditions).Should(utiltesting.HaveConditionStatusFalse(kueue.WorkloadAdmitted))
+					g.Expect(wl.Status.Conditions).Should(utiltesting.HaveConditionStatusTrue(kueue.WorkloadEvicted))
+					g.Expect(wl.Status.Conditions).Should(utiltesting.HaveConditionStatusTrue(kueue.WorkloadWaitingForReplacementPods))
+				}, util.ShortConsistentDuration, util.ShortInterval).Should(gomega.Succeed())
+			})
```
mimowo: I think this might be flaky if the test code does not yet see the transition. 

To make it robust let's use Eventually rather than `time.Sleep(backoffBaseSeconds * time.Second)`. 

You can repeat the check after eventualy with consistently.

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
```diff
@@ -2455,6 +2455,193 @@ var _ = ginkgo.Describe("Pod controller interacting with Workload controller whe
...
```
mimowo: Same here, avoid custom sleeps, use Eventually followed by Consistently

mimowo (commented):
LGTM, just minor comments to the code and tests, thank you for the bugfix 👍

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
mbobrovskyi: But we are waiting for a transition in "Checking that workload was evicted due to pods ready timeout". The idea is to wait for the backoff period plus 1 second to confirm that no transition occurs after the backoff period. Alternatively, we can use the consistency timeout: `backoffBaseSeconds * time.Second + util.ShortConsistentDuration`.

mimowo: Sure, but instead of waiting for the transition with 1s you can use Eventually to avoid magic numbers and to eliminate the chance of flaking

mbobrovskyi: Sorry, I didn't understand what we should wait for with Eventually. On the contrary, we should verify that nothing happens.

mbobrovskyi: Ah, I see – it's for the BackoffFinished status. Makes sense.

mimowo: I see, so if you deleted "Waiting for backoff"  the tests would pass too? 

In that case I would either:
1. wait for the backoff by the API observable fields, maybe status.requeueAfter? Or  Requeued=True?
2. use consistently with `backoffBaseSeconds * time.Second + util.ShortConsistentDuration` as proposed above

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
mbobrovskyi: Done

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
mbobrovskyi: Done

pkg/controller/jobs/pod/pod_controller.go
mbobrovskyi: Done

pkg/controller/jobs/pod/pod_controller.go
```diff
@@ -1293,6 +1291,19 @@ func (p *Pod) FindMatchingWorkloads(ctx context.Context, c client.Client, r even
...
+	if p.fastAdmission() {
+		if ps.Count > 0 && activePods == 0 {
+			return 1
+		}
+		return 0
+	}
+	if absentCount := int(ps.Count) - activePods; absentCount > 0 {
+		return absentCount
+	}
+	return 0
```
mimowo: nit, this can be just return `max(0, int(ps.Count) - activePods)`

mimowo:
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
please prepare the CPs in case the robot fails on conflicts
for the nit
/hold

pkg/controller/jobs/pod/pod_controller.go
mbobrovskyi: Done

mimowo:
/test pull-kueue-test-e2e-baseline-main-1-36
This is not related
@codetabbitai please open the flake issue

mimowo:
@coderabbitai please open the flake issue

mbobrovskyi:
@mimowo could you please lgtm this PR?

mimowo:
/release-note-edit
```release-note
PodGroup integration: Fixed a bug that allowed Workloads corresponding to PodGroups with the `WaitingForReplacementPods=True` condition to be re-admitted immediately.
```

PR #12776, opened 2026-07-03
Add KEP for waitForPodsReady per workload

MaysaMacedo:
#### What type of PR is this?

/kind feature
/kind documentation

KEP for #4803 

This pull request was created with the assistance of AI (Claude Code)

```release-note

```

## Summary by CodeRabbit

* **Documentation**
  * Added a KEP specification for workload-level `WaitForPodsReady`, including per-workload timeout configuration, annotation support, precedence rules, validation, immutability, and controller behavior.
  * Documented `blockAdmission` usage, integration details, testing requirements, and graduation criteria for alpha, beta, and stable stages.
* **Chores**
  * Added KEP metadata with provisional status, milestone mappings, feature-gate requirements, and metric requirements for later stages.

### DIFF
--- a/keps/4803-workload-level-wait-for-pods-ready/README.md
+++ b/keps/4803-workload-level-wait-for-pods-ready/README.md
@@ -1,0 +1,212 @@
+# KEP-4803: Workload-level WaitForPodsReady
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Open Questions](#open-questions)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API](#api)
+    - [Workload Spec](#workload-spec)
+    - [Job Annotations](#job-annotations)
+  - [Controller](#controller)
+    - [Workload](#workload)
+    - [Jobs / Jobframework](#jobs--jobframework)
+  - [Webhooks](#webhooks)
+    - [Workload](#workload-1)
+    - [Jobs](#jobs)
+  - [Test Plan](#test-plan)
+    - [Prerequisite testing updates](#prerequisite-testing-updates)
+    - [Integration tests](#integration-tests)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+<!-- /toc -->
+
+## Summary
+
+This proposal introduces a mechanism to allow specification of `WaitForPodsReady` configuration per job.
+
+## Motivation
+
+Currently the `WaitForPodsReady` feature can only be configured at the cluster
+level via the Kueue `ConfigMap`. Since different workloads may take different
+amounts of time to reach a ready state, a single cluster-wide setting may often
+not fit well all workloads.
+
+### Goals
+
+- Introduce per-workload `WaitForPodsReady.Timeout` field in `WorkloadSpec`.
+- Define new Job annotations `kueue.x-k8s.io/pods-ready-timeout` that is
+  propagated to the corresponding `WorkloadSpec` field at Workload creation time.
+- When a workload carries per-workload timeouts, enforce them independently of
+  the cluster-wide `WaitForPodsReady` configuration.
+- Per-workload timeouts take precedence over the global configuration when
+  both are present.
+
+### Non-Goals
+
+- Change the cluster-wide `WaitForPodsReady` behaviour.
+- Change the requeue / backoff strategy; that remains cluster-wide only
+- Introduce per-workload `BlockAdmission` semantics.
+
+## Proposal
+
+One new optional field is added to `WorkloadSpec`. The job object has the annotation
+populated. The workload controller reads the annotation during reconciliation and uses it instead of the global config to determine whether to evict a workload that is taking too long
+to become ready.
+
+### User Stories
+
+#### Story 1
+
+As a batch platform administrator, I want to configure a tight pods-ready
+timeout for data-preprocessing jobs (2 minutes) while allowing large ML
+training jobs a longer timeout (30 minutes), without requiring two separate
+Kueue deployments.
+
+#### Story 2
+
+As a job owner whose workload requires pulling a large container image that takes longer than the cluster-wide WaitForPodsReady timeout, I want to annotate my job with a longer per-job timeout so it is not evicted prematurely, without requiring the platform administrator to relax the global timeout for everyone.
+
+
+### Notes/Constraints/Caveats
+
+- Should we allow combining `WaitForPodsReady` disabled and enforcement of
+  per-workload timeout? Maybe it should always be enabled to ensure the
+  other `WaitForPodsReady` fields could be enforced.
+
+### Risks and Mitigations
+
+- **Conflicting values**: A user may set a per-workload timeout that conflicts
+  with a cluster-wide one.  Mitigation: per-workload values always take
+  precedence; a clear precedence rule is documented and enforced in the webhook.
+
+## Design Details
+
+### API
+
+#### Workload Spec
+
+One new optional field is added to `WorkloadSpec`:
+
+```go
+// WorkloadSpec defines the desired state of Workload
+type WorkloadSpec struct {
+    // ...existing fields...
+
+    // podsReadyTimeout defines the maximum duration the workload may remain
+    // admitted before all pods are in a Ready or Succeeded state.
+    // When elapsed, the workload is evicted with reason PodsReadyTimeout.
+    // If both this field and the cluster-wide WaitForPodsReady.Timeout are set,
+    // this field takes precedence.
+    // +optional
+    PodsReadyTimeout *metav1.Duration `json:"podsReadyTimeout,omitempty"`
+}
+```
+
+#### Job Annotations
+
+One new annotation is defined in `pkg/controller/constants/constants.go`:
+
+```go
+// PodsReadyTimeoutAnnotation sets the per-job pods-ready timeout.
+// Value must be a Go duration string parseable by time.ParseDuration (e.g. "10m").
+PodsReadyTimeoutAnnotation = "kueue.x-k8s.io/pods-ready-timeout"
+```
+
+### Controller
+
+#### Workload
+
+`admittedNotReadyWorkload` in `pkg/controller/core/workload_controller.go` is
+updated to resolve the timeout by merging the per-workload spec with
+the global config.
+
+`reconcileNotReadyTimeout` gains a short-circuit check: if neither the global
+config is set nor the workload carries per-workload fields, the function
+returns immediately (current behaviour preserved).
+
+The check that decides to write the `WorkloadPodsReady` condition in the job reconciler
+(`pkg/controller/jobframework/reconciler.go`) is updated to also check when the
+workload carries per-workload timeout.
+
+#### Jobs / Jobframework
+
+`constructWorkload` (in `pkg/controller/jobframework/reconciler.go`) reads the
+new annotations from the Job and populates the corresponding `WorkloadSpec`
+field.
+
+### Webhooks
+
+#### Workload
+
+- Validate that `podsReadyTimeout`, if set, is greater than zero.
+- Field is immutable while the workload is admitted.
+
+#### Jobs
+
+- Validate that `kueue.x-k8s.io/pods-ready-timeout` is a valid duration
+  string greater than zero.
+- The annotation is immutable while the job is unsuspended.
+- (open question) Should the workload be recreated if the annotation value changes
+or there should be a webhook that doesn't allow the update?
+
+### Test Plan
+
+[x] I/we understand the owners of the involved components may require updates to
+existing tests to make this code solid enough prior to committing the changes
+necessary to implement this enhancement.
+
+#### Prerequisite testing updates
+
+No regressions in the existing `WaitForPodsReady` tests under
+`test/integration/singlecluster/scheduler/podsready/` and
+`test/e2e/sequential/baseline/waitforpodsready_test.go`.
+
+#### Integration tests
+
+##### controller/core/workload
+
+- "The per-workload podsReadyTimeout takes precedence over the cluster-wide
+  timeout."
+
+##### controller/jobs/job
+
+- "A Job annotated with kueue.x-k8s.io/pods-ready-timeout has the timeout
+  propagated to its Workload spec."
+
+### Graduation Criteria
+
+#### Alpha
+
+- Feature gate `WorkloadLevelWaitForPodsReady` introduced, disabled by default.
+- New `WorkloadSpec` fields and Job annotations implemented.
+- Unit and integration tests covering the core controller and job framework
+  propagation added.
+
+#### Beta
+
+- Feature gate enabled by default.
+- E2E tests added.
+- Documentation updated.
+
+#### Stable
+
+- No issues reported for two or more releases.
+- Feature gate removed; behaviour always on.
+
+## Implementation History
+
+- 2026-07-03: KEP created as provisional.
+
+## Drawbacks
+
+- Adds optional field to `WorkloadSpec`, increasing API surface area.
--- a/keps/4803-workload-level-wait-for-pods-ready/README.md
+++ b/keps/4803-workload-level-wait-for-pods-ready/README.md
@@ -1,0 +1,297 @@
+# KEP-4803: Workload-level WaitForPodsReady
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+    - [Story 3](#story-3)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API](#api)
+    - [Workload Spec](#workload-spec)
+    - [Resource Annotations](#resource-annotations)
+  - [Controller](#controller)
+    - [Workload](#workload)
+    - [Jobs / Jobframework](#jobs--jobframework)
+  - [Webhooks](#webhooks)
+    - [Managed resources (Jobs, Deployments, StatefulSets, etc.)](#managed-resources-jobs-deployments-statefulsets-etc)
+  - [Test Plan](#test-plan)
+    - [Prerequisite testing updates](#prerequisite-testing-updates)
+    - [Webhook unit tests](#webhook-unit-tests)
+      - [webhooks/job (controller/jobframework/validation)](#webhooksjob-controllerjobframeworkvalidation)
+      - [webhooks/deployment](#webhooksdeployment)
+    - [Integration tests](#integration-tests)
+      - [controller/core/workload](#controllercoreworkload)
+      - [controller/jobs/job](#controllerjobsjob)
+      - [controller/jobs/deployment (Pod-template exception)](#controllerjobsdeployment-pod-template-exception)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [Stable](#stable)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+<!-- /toc -->
+
+## Summary
+
+This proposal introduces a mechanism to allow specification of `WaitForPodsReady` timeout per workload.
+
+## Motivation
+
+Currently the `WaitForPodsReady` feature can only be configured at the cluster
+level via the Kueue `ConfigMap`. Since different workloads may take different
+amounts of time to reach a ready state, a single cluster-wide setting may often
+not fit well all workloads.
+
+### Goals
+
+- Introduce per-workload `WaitForPodsReady.TimeoutSeconds` field in `WorkloadSpec`.
+- Define a new annotation `kueue.x-k8s.io/wait-for-pods-ready-timeout-seconds` applicable to any
+  Kueue-managed resource (Job, StatefulSet, etc.)
+  that is propagated to the corresponding `WorkloadSpec` field at Workload creation time.
+  For Deployment, the annotation needs to be defined at `spec.template.metadata.annotations`
+  to be propagated to the Pods and then workloads.
+- Per-workload timeouts take precedence over the global `WaitForPodsReady` timeout configuration when
+  both are present.
+- Introduce `MaxTimeoutOnWorkload` field in the configuration.
+
+### Non-Goals
+
+- Change the cluster-wide `WaitForPodsReady` behaviour.
+- Change the recovery / backoff strategy; that remains cluster-wide only
+- Introduce per-workload `BlockAdmission` semantics.
+
+## Proposal
+
+One new optional field is added to `WorkloadSpec`. The
+`kueue.x-k8s.io/wait-for-pods-ready-timeout-seconds` annotation is parsed at Workload creation
+and stored in `wl.Spec.WaitForPodsReady.TimeoutSeconds`. The eviction deadline
+is resolved exclusively from that stored field.
+
+A new optional configuration is added to the `WaitForPodsReady` that allows admins to control
+the maximum values they want to allow users to specify at the `kueue.x-k8s.io/wait-for-pods-ready-timeout-seconds`
+annotation.
+
+The managed-resource webhook (see [Webhooks](#webhooks)) rejects annotation
+changes while the resource is unsuspended. When the job is suspended and the
+annotation is changed, the mismatch is detected and the Workload spec is reconstructed.
+The webhooks also rejects timeouts that exceed the maximum defined by the admin.
+
+### User Stories
+
+#### Story 1
+
+As a batch platform administrator, I want to configure a tight pods-ready
+timeout for data-preprocessing jobs (2 minutes) while allowing large ML
+training jobs a longer timeout (30 minutes), without requiring two separate
+Kueue deployments.
+
+#### Story 2
+
+As a job owner whose workload requires pulling a large container image that takes longer than the cluster-wide WaitForPodsReady timeout, I want to annotate my job with a longer per-job timeout so it is not evicted prematurely, without requiring the platform administrator to relax the global timeout for everyone.
+
+#### Story 3
+
+As a platform administrator, I want to configure Kueue with `blockAdmission: true` and assign per-workload pods-ready timeouts so that each workload is evicted after its own appropriate timeout rather than a single cluster-wide timeout. This ensures that a workload whose pods never become ready is evicted promptly and unblock admission of workloads in other ClusterQueues.
+
+### Notes/Constraints/Caveats
+
+- When a per-workload timeout is enforced and `DisableWaitForPodsReady` feature gate is enabled
+the workload is forbbiden from being admitted.
+
+### Risks and Mitigations
+
+- **Conflicting values**: A user may set a per-workload timeout that conflicts
+  with a cluster-wide one.  Mitigation: per-workload values always take
+  precedence; a clear precedence rule is documented and enforced in the webhook.
+
+## Design Details
+
+### API
+
+#### Workload Spec
+
+The `WaitForPodsReady` struct and its `TimeoutSeconds` field are added to
+`WorkloadSpec` in **v1beta2 only**.
+
+```go
+// WorkloadSpec defines the desired state of Workload
+type WorkloadSpec struct {
+    // ...existing fields...
+
+    WaitForPodsReady *WaitForPodsReady `json:"WaitForPodsReady,omitempty"`
+}
+
+// +kubebuilder:validation:MinProperties=1
+type WaitForPodsReady struct {
+    // timeoutSeconds defines the maximum time the workload may remain
+    // admitted before all pods are in a Ready or Succeeded state.
+    // When elapsed, the workload is evicted with reason PodsReadyTimeout.
+    // If both this field and the cluster-wide WaitForPodsReady.Timeout are set,
+    // this field takes precedence.
+    // +optional
+    // +kubebuilder:validation:Minimum=1
+    TimeoutSeconds *int64 `json:"timeoutSeconds,omitempty"`
+}
+```
+
+**Conversion** (`apis/kueue/v1beta1/workload_conversion.go`):
+
+No manual conversion function is needed. The generated `autoConvert_*` drops the
+field when converting v1beta2→v1beta1.
+
+#### Configuration
+
+The `WaitForPodsReady` struct is has a new field added `MaxTimeoutOnWorkload`:
+
+```go
+type WaitForPodsReady struct {
+  ...
+	// MaxTimeoutOnWorkload defines the upper bound allowed for a per-workload
+	// PodsReady timeout override (set via the `kueue.x-k8s.io/wait-for-pods-ready-timeout-seconds`
+	// annotation). If a workload requests a timeout greater than
+	// MaxTimeoutOnWorkload, the job is rejected by the admission webhook.
+	// When unset, the default maximum of 2 hours is enforced.
+	// It has no effect on workloads that don't set a per-workload override.
+	// +optional
+	MaxTimeoutOnWorkload *metav1.Duration `json:"maxTimeoutOnWorkload,omitempty"`
+}
+```
+
+#### Resource Annotations
+
+One new annotation is defined in `pkg/controller/constants/constants.go`:
+
+```go
+// WaitForPodsReadyTimeoutSecondsAnnotation is the annotation key on any Kueue-managed resource that sets a
+	// per-workload pods-ready timeout, overriding the cluster-wide WaitForPodsReady.Timeout.
+	// Value must be an integer number of seconds.
+	WaitForPodsReadyTimeoutSecondsAnnotation = "kueue.x-k8s.io/wait-for-pods-ready-timeout-seconds"
+```
+
+### Controller
+
+#### Workload
+
+`admittedNotReadyWorkload` in `pkg/controller/core/workload_controller.go` is
+updated to resolve the effective timeout exclusively from `wl.Spec.WaitForPodsReady.TimeoutSeconds`,
+falling back to the cluster-wide config only when that field is nil.
+The eviction deadline is resolved exclusively from `wl.Spec.WaitForPodsReady.TimeoutSeconds`
+and never from the live annotation. `prepareWorkload` does re-read the annotation
+during reconciliation to keep the stored field in sync, but any attempt to
+update `TimeoutSeconds` while the Job is unsuspended is rejected by the
+managed-resource webhook before the change can reach the API server.
+
+#### Jobs / Jobframework
+
+The check that decides to write the `WorkloadPodsReady` condition in the job reconciler
+(`pkg/controller/jobframework/reconciler.go`) is updated to also trigger when
+`wl.Spec.WaitForPodsReady` is non-nil.
+
+`ConstructWorkload` (in `pkg/controller/jobframework/reconciler.go`) reads the
+`kueue.x-k8s.io/pods-ready-timeout` annotation from the managed resource object
+and populates the corresponding `WorkloadSpec` field. This covers most integrations
+(Job, StatefulSet, RayJob, PyTorchJob, JobSet, etc.).
+
+The **Deployment** integration is an exception: it does not construct `Workload` objects directly. Kueue tracks Deployment-owned workloads via the Pod integration, so the annotation must be placed by the user on
+the Pod template (`spec.template.metadata.annotations`). The Pod
+integration then calls `ConstructWorkload` directly and reads the annotation from
+the Pod object, so no additional propagation logic is needed in Kueue.
+
+### Webhooks
+
+#### Managed resources (Jobs, Deployments, StatefulSets, etc.)
+
+- Validate that `kueue.x-k8s.io/wait-for-pods-ready-timeout-seconds` is an int greater than zero
+  and doesn't exceed the maximum value set by the admin at the cluster configuration.
+- The annotation is immutable while the job is unsuspended. Changes are allowed
+  while the job is suspended (i.e. between eviction cycles), which is the
+  intended window for a user to adjust the timeout before re-admission.
+  `EquivalentToWorkload` detects the drift and `prepareWorkload` updates the
+  existing Workload's `spec.waitForPodsReady.TimeoutSeconds` field in place —
+  no delete-and-recreate occurs.
+
+### Test Plan
+
+[x] I/we understand the owners of the involved components may require updates to
+existing tests to make this code solid enough prior to committing the changes
+necessary to implement this enhancement.
+
+#### Prerequisite testing updates
+
+No regressions in the existing `WaitForPodsReady` tests under
+`test/integration/singlecluster/scheduler/podsready/` and
+`test/e2e/sequential/baseline/waitforpodsready_test.go`.
+
+#### Webhook unit tests
+
+##### webhooks/job (controller/jobframework/validation)
+
+- "A Job annotated with kueue.x-k8s.io/pods-ready-timeout set to a valid
+  duration is accepted."
+- "A Job annotated with kueue.x-k8s.io/pods-ready-timeout set to a zero or
+  negative duration is rejected at admission time."
+- "A Job annotated with kueue.x-k8s.io/pods-ready-timeout set to a non-duration
+  string is rejected at admission time."
+
+##### webhooks/deployment
+
+- "A Deployment whose Pod template carries kueue.x-k8s.io/pods-ready-timeout
+  with a valid duration is accepted."
+- "A Deployment whose Pod template carries kueue.x-k8s.io/pods-ready-timeout
+  with an invalid value is rejected at admission time."
+
+#### Integration tests
+
+##### controller/core/workload
+
+- "The per-workload TimeoutSeconds takes precedence over the cluster-wide
+  timeout."
+- "RecoveryTimeout uses only the cluster-wide value even when a per-workload
+  TimeoutSeconds is set."
+
+##### controller/jobs/job
+
+- "A Job annotated with kueue.x-k8s.io/pods-ready-timeout has the timeout
+  propagated to its Workload spec."
+
+##### controller/jobs/deployment (Pod-template exception)
+
+- "A Deployment whose Pod template carries kueue.x-k8s.io/pods-ready-timeout
+  has the timeout propagated to the Workload created by the Pod integration."
+
+### Graduation Criteria
+
+#### Alpha
+
+- Feature gate `WorkloadLevelWaitForPodsReady` introduced, disabled by default.
+- New `WorkloadSpec` fields and resource annotations implemented.
+- Unit tests for annotation parsing, webhook validation (duration, immutability),
+  and the Deployment Pod-template exception added.
+- Integration tests covering precedence, cluster-wide fallback, RecoveryTimeout
+  behaviour, and Job/Deployment propagation added.
+- E2E tests added.
+
+#### Beta
+
+- Feature gate enabled by default.
+- Documentation updated.
+
+#### Stable
+
+- No issues reported for two or more releases.
+- Feature gate removed; behaviour always on.
+
+## Implementation History
+
+- 2026-07-03: KEP created as provisional.
+
+## Drawbacks
+
+- Adds optional field to `WorkloadSpec`, increasing API surface area.
--- a/keps/4803-workload-level-wait-for-pods-ready/README.md
+++ b/keps/4803-workload-level-wait-for-pods-ready/README.md
@@ -14,24 +14,21 @@
   - [Risks and Mitigations](#risks-and-mitigations)
 - [Design Details](#design-details)
   - [API](#api)
-    - [Workload Spec](#workload-spec)
     - [Resource Annotations](#resource-annotations)
   - [Controller](#controller)
     - [Workload](#workload)
     - [Jobs / Jobframework](#jobs--jobframework)
   - [Webhooks](#webhooks)
     - [Managed resources (Jobs, Deployments, StatefulSets, etc.)](#managed-resources-jobs-deployments-statefulsets-etc)
+  - [Future Work](#future-work)
   - [Test Plan](#test-plan)
     - [Prerequisite testing updates](#prerequisite-testing-updates)
     - [Webhook unit tests](#webhook-unit-tests)
       - [webhooks/job (controller/jobframework/validation)](#webhooksjob-controllerjobframeworkvalidation)
-      - [webhooks/deployment](#webhooksdeployment)
-    - [Integration tests](#integration-tests)
-      - [controller/core/workload](#controllercoreworkload)
-      - [controller/jobs/job](#controllerjobsjob)
-      - [controller/jobs/deployment (Pod-template exception)](#controllerjobsdeployment-pod-template-exception)
+    - [e2e](#e2e)
   - [Graduation Criteria](#graduation-criteria)
-    - [Alpha](#alpha)
+    - [Alpha1 (0.20)](#alpha1-020)
+    - [Alpha2 (0.21)](#alpha2-021)
     - [Beta](#beta)
     - [Stable](#stable)
 - [Implementation History](#implementation-history)
@@ -51,37 +48,33 @@ not fit well all workloads.
 
 ### Goals
 
-- Introduce per-workload `WaitForPodsReady.TimeoutSeconds` field in `WorkloadSpec`.
-- Define a new annotation `kueue.x-k8s.io/wait-for-pods-ready-timeout-seconds` applicable to any
-  Kueue-managed resource (Job, StatefulSet, etc.)
-  that is propagated to the corresponding `WorkloadSpec` field at Workload creation time.
-  For Deployment, the annotation needs to be defined at `spec.template.metadata.annotations`
-  to be propagated to the Pods and then workloads.
-- Per-workload timeouts take precedence over the global `WaitForPodsReady` timeout configuration when
+- Introduce the JSON-based annotation: `kueue.x-k8s.io/wait-for-pods-ready` applicable to any
+  Kueue-managed resource (Job, StatefulSet, etc.). For Deployment, the annotation needs to be
+  defined at `spec.template.metadata.annotations` to be propagated to the Pods and then workloads.
+- Per-workload timeouts take precedence over the global `WaitForPodsReady` timeouts configuration when
   both are present.
 - Introduce `MaxTimeoutOnWorkload` field in the configuration.
 
 ### Non-Goals
 
-- Change the cluster-wide `WaitForPodsReady` behaviour.
-- Change the recovery / backoff strategy; that remains cluster-wide only
+- Change the cluster-wide `WaitForPodsReady` behavior.
+- Change the backoff strategy; that remains cluster-wide only
 - Introduce per-workload `BlockAdmission` semantics.
+- Supporting MultiKueue in Alpha. Re-evaluate at Beta.
 
 ## Proposal
 
-One new optional field is added to `WorkloadSpec`. The
-`kueue.x-k8s.io/wait-for-pods-ready-timeout-seconds` annotation is parsed at Workload creation
-and stored in `wl.Spec.WaitForPodsReady.TimeoutSeconds`. The eviction deadline
-is resolved exclusively from that stored field.
+The `kueue.x-k8s.io/wait-for-pods-ready` annotation with the values for `Timeout` and `recoveryTimeout`
+is read from the managed resource and stored in the workload annotation.
 
 A new optional configuration is added to the `WaitForPodsReady` that allows admins to control
-the maximum values they want to allow users to specify at the `kueue.x-k8s.io/wait-for-pods-ready-timeout-seconds`
+the maximum timeout values they want to allow users to specify at the `kueue.x-k8s.io/wait-for-pods-ready`
 annotation.
 
 The managed-resource webhook (see [Webhooks](#webhooks)) rejects annotation
 changes while the resource is unsuspended. When the job is suspended and the
 annotation is changed, the mismatch is detected and the Workload spec is reconstructed.
-The webhooks also rejects timeouts that exceed the maximum defined by the admin.
+The webhooks also rejects timeouts that exceed the maximum timeout or are not valid integers.
 
 ### User Stories
 
@@ -103,11 +96,11 @@ As a platform administrator, I want to configure Kueue with `blockAdmission: tru
 ### Notes/Constraints/Caveats
 
 - When a per-workload timeout is enforced and `DisableWaitForPodsReady` feature gate is enabled
-the workload is forbbiden from being admitted.
+the timeout is ignored and a log warns about incompatible combination.
 
 ### Risks and Mitigations
 
-- **Conflicting values**: A user may set a per-workload timeout that conflicts
+- **Conflicting values**: A user may set per-workload timeouts that conflicts
   with a cluster-wide one.  Mitigation: per-workload values always take
   precedence; a clear precedence rule is documented and enforced in the webhook.
 
@@ -115,16 +108,105 @@ the workload is forbbiden from being admitted.
 
 ### API
 
-#### Workload Spec
+The `WaitForPodsReady` struct has a new field added `MaxTimeoutOnWorkload`:
 
-The `WaitForPodsReady` struct and its `TimeoutSeconds` field are added to
-`WorkloadSpec` in **v1beta2 only**.
+```go
+type WaitForPodsReady struct {
+  ...
+	// MaxTimeoutOnWorkload defines the upper bound allowed for a per-workload
+	// PodsReady timeout and recoveryTimeout override (set via the `kueue.x-k8s.io/wait-for-pods-ready`
+	// annotation). If a workload requests a timeout or recoveryTimeout greater than
+	// MaxTimeoutOnWorkload, the job is rejected by the admission webhook.
+	// When unset, the default maximum of 2 hours is enforced.
+	// It has no effect on workloads that don't set a per-workload override.
+	// +optional
+	MaxTimeoutOnWorkload *metav1.Duration `json:"maxTimeoutOnWorkload,omitempty"` 
+}
+```
+
+#### Resource Annotations
+
+One new annotation is defined in `pkg/controller/constants/constants.go`:
+
+```go
+  // WaitForPodsReadyAnnotation is the annotation key on any Kueue-managed resource that sets
+	// per-workload timeout and recoveryTimeout, overriding those values at cluster-wide WaitForPodsReady.
+  // The value is a JSON containing timeout in seconds and RecoveryTimeout in seconds.
+	// This annotation is alpha-level enabled by the WorkloadLevelWaitForPodsReady.
+	WaitForPodsReadyAnnotation = "kueue.x-k8s.io/wait-for-pods-ready"
+```
+
+Here is an example of the annotation definition:
+
+```yaml
+annotations:
+  kueue.x-k8s.io/wait-for-pods-ready: '{"timeoutSeconds": 30, "recoveryTimeoutSeconds": 40}'
+```
+
+### Controller
+
+#### Workload
+
+The workload controller is updated to resolve both the effective eviction deadline
+and the recovery timeout from the per-workload values carried in the workload's
+own annotations, falling back to the cluster-wide configuration for each field
+independently when no per-workload value is present. This means a workload may override 
+the timeout, the recovery timeout, or both. When only the timeout is specified,
+it is also used as the default for the recovery timeout. When neither field is
+specified, the cluster-wide values are used as the fallback for each.
+The annotation is propagated from the managed resource to the workload at
+construction time and re-evaluated on every reconciliation to keep the workload
+annotation in sync with any changes made while the job is suspended. Any attempt
+to change the annotation on the managed resource while the job is unsuspended is
+rejected by the managed-resource webhook before it can take effect.
+
+#### Jobs / Jobframework
+
+The job reconciler is updated so that the `WorkloadPodsReady` condition is also
+written when the workload carries the `kueue.x-k8s.io/wait-for-pods-ready`
+annotation, in addition to the existing cluster-wide trigger.
+
+When a workload is constructed for a managed resource, the
+`kueue.x-k8s.io/wait-for-pods-ready` annotation is read from that resource and
+copied to the resulting workload's annotations. This covers most integrations
+(Job, StatefulSet, RayJob, PyTorchJob, JobSet, etc.).
+
+The **Deployment** integration is an exception: Kueue tracks Deployment-owned
+workloads via the Pod integration rather than directly from the Deployment object.
+Users must therefore place the annotation on the Pod template
+(`spec.template.metadata.annotations`), from where it is read when the workload
+is constructed for the Pod — no additional propagation logic is needed in Kueue.
+
+### Webhooks
+
+#### Managed resources (Jobs, Deployments, StatefulSets, etc.)
+
+- Validate that the timeouts values in `kueue.x-k8s.io/wait-for-pods-ready` are
+  positive integers and does not exceed the maximum value set by the admin in the
+  cluster configuration.
+- Setting a recoveryTimeout without timeout set is not supported.
+- The annotation is immutable while the job is unsuspended. Changes are allowed
+  while the job is suspended (i.e. between eviction cycles), which is the
+  intended window for a user to adjust the timeout before re-admission. When a
+  change is detected during reconciliation, the workload's own annotation is
+  updated in place — no delete-and-recreate occurs.
+
+### Future Work
+
+The `kueue.x-k8s.io/wait-for-pods-ready` annotation is parsed at Workload creation
+and stored in the Workload spec fields `timeoutSeconds` and `recoveryTimeoutSeconds`
+instead of being carried as a workload annotation.
+
+The following APIs are considered for future releases and will be evaluated.
+
+The `WaitForPodsReady` struct and its `TimeoutSeconds` field are added to `WorkloadSpec`.
 
 ```go
 // WorkloadSpec defines the desired state of Workload
 type WorkloadSpec struct {
     // ...existing fields...
-
+    // WaitForPodsReady ensures the workload is ready within an specific timeout
+    // +optional
     WaitForPodsReady *WaitForPodsReady `json:"WaitForPodsReady,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


keps/4803-workload-level-wait-for-pods-ready/README.md
```diff
@@ -0,0 +1,212 @@
...
+
+### API
+
+#### Workload Spec
+
+One new optional field is added to `WorkloadSpec`:
+
+```go
+// WorkloadSpec defines the desired state of Workload
+type WorkloadSpec struct {
```
kannon92, line 125: Even if we decide to only go with a single field for PodsReady (timeout), I think maybe we should consider a struct. If we get new requests we at least have the API to add them.

Something like
```golang
type WorkloadSpec struct {
WaitForPodsReadyOverride *WaitForPodsReadyOverride `json:"waitForPodsReadyOverride,omitempty"`
}

type WaitForPodsReadyOverride struct {
  PodsReadyTimeout *metav1.Duration `json:"podsReadyTimeout,omitempty"`
}
```

keps/4803-workload-level-wait-for-pods-ready/README.md
```diff
@@ -0,0 +1,212 @@
...
+
+`constructWorkload` (in `pkg/controller/jobframework/reconciler.go`) reads the
+new annotations from the Job and populates the corresponding `WorkloadSpec`
+field.
+
+### Webhooks
+
+#### Workload
+
+- Validate that `podsReadyTimeout`, if set, is greater than zero.
```
kannon92: Both of these validations should be done as cel on the API. This way we can avoid having code for this validation.

kannon92: Maybe immutable while the workload is admitted may be tough to do as CEL. But the first one can be done via CEL expressions for sure.

kaushik229 (commented):
Thanks for driving this KEP! Adding a motivating data point and a couple of design thoughts.
Motivation (cluster-global blast radius). On current main with blockAdmission: true, one admitted-but-not-yet-PodsReady workload blocks admission across all ClusterQueues podsReadyForAllAdmittedWorkloads iterates every CQ and blocks the single scheduling goroutine. 
Repro on kind: wl-a in cq-a/cohort-a (admitted, pod unschedulable → never ready) prevents a fitting wl-b in an empty cq-b in a different cohort (cohort-b) from being admitted; deleting wl-a immediately unblocks wl-b. This is the concrete pain a per-workload timeout addresses.

On the [open questions](https://github.com/kubernetes-sigs/kueue/issues/4803#issuecomment-4442205885):
1. Per-workload when global waitForPodsReady is disabled — from a user POV, opting in per-workload independent of the global setting is the most useful (e.g. enable it only for gang-scheduled jobs). 
Implementation note: WorkloadsNotReady tracking (podsReadyTracking) is currently enabled only when global BlockAdmission=true, so supporting per-workload-when-global-off would mean enabling that tracking unconditionally (or per-CQ) — worth calling out in the KEP.
2. API — a JSON annotation on the Job propagated to the Workload seems reasonable and matches the discussion; a defaulting story (per @amy 's [point](https://github.com/kubernetes-sigs/kueue/issues/4803#issuecomment-4441719532)) would avoid relying on users to set it everywhere. 
Scope dimension. Even with a per-workload Timeout, a workload with a long/absent timeout still freezes unrelated  cohorts while blocked. It might help for the KEP to state explicitly whether per-ClusterQueue/cohort scoping of the block is in or out of scope — [KEP-349 deferred exactly this granularity](https://github.com/kubernetes-sigs/kueue/tree/main/keps/349-all-or-nothing#more-granular-configuration-to-enable-the-mechanism), so a sentence on the chosen direction would clarify.

kannon92:
/release-note-none

This is just a KEP so not release note needed.

anguszzzz:
I want to follow up on this and see whether this is still in-progress, our production is impacted by this and I'm happy to continue drive/help this new feature, thanks!

kannon92:
> I want to follow up on this and see whether this is still in-progress, our production is impacted by this and I'm happy to continue drive/help this new feature, thanks!

This is being actively worked on. I asked this to be included in the 0.20 features so hope is that we get reviews.

kannon92 (commented):
/lgtm

/assign @tenzen-y @mimowo @amy

keps/4803-workload-level-wait-for-pods-ready/kep.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+see-also:
+  - "/keps/349-all-or-nothing"
+
+stage: alpha
+
+latest-milestone: "v0.19"
+
+milestone:
+  alpha: "v0.19"
```
kannon92: can you update these milestones when you get a chance?

MaysaMacedo:
> I want to follow up on this and see whether this is still in-progress, our production is impacted by this and I'm happy to continue drive/help this new feature, thanks!

@anguszzzz hello, thanks for willing to help. I'm currently working on it and proposed a draft of the implementation. It would be great to have your feedback on the KEP/implementation, if you notice anything you want to share.

anguszzzz:
@MaysaMacedo thanks! This is really helpful. I will check design in details

MaysaMacedo:
@kannon92 I have an open question, what do you think should happen if the user set `DisableWaitForPodsReady` Feature Gate, but the user specified the annotation of Timeout on the Resource?

kannon92:
> @kannon92 I have an open question, what do you think should happen if the user set `DisableWaitForPodsReady` Feature Gate, but the user specified the annotation of Timeout on the Resource?

We could forbid the workload from getting admitted. IMO if someone is setting this feature gate then I think we would need to know why. The goal should be that this feature gate is not disabled.

keps/4803-workload-level-wait-for-pods-ready/README.md
MaysaMacedo: The type Duration is serialized as a string, so the CEL cannot make numeric comparisons.
Unless we want to make the timeout to be represented in something like seconds and we only accept an int value.

MaysaMacedo:
/hold
want to add some adjustments

keps/4803-workload-level-wait-for-pods-ready/README.md
```diff
@@ -0,0 +1,273 @@
...
+
+As a job owner whose workload requires pulling a large container image that takes longer than the cluster-wide WaitForPodsReady timeout, I want to annotate my job with a longer per-job timeout so it is not evicted prematurely, without requiring the platform administrator to relax the global timeout for everyone.
+
+#### Story 3
+
+As a platform administrator, I want to configure Kueue with `blockAdmission: true` and assign per-workload pods-ready timeouts so that each workload is evicted after its own appropriate timeout rather than a single cluster-wide timeout. This ensures that a workload whose pods never become ready is evicted promptly and unblock admission of workloads in other ClusterQueues.
+
+### Notes/Constraints/Caveats
+
+- When a per-workload timeout is enforced and `DisableWaitForPodsReady` feature gate is enabled
```
MaysaMacedo, line 105: I would appreciate inputs on this. I'm not sure if this should be the expected behavior.

keps/4803-workload-level-wait-for-pods-ready/kep.yaml
```diff
@@ -0,0 +1,34 @@
+title: Workload-level WaitForPodsReady
+kep-number: 4803
+authors:
+  - "@MaysaMacedo"
+status: provisional
```
kannon92:
```suggestion
status: implementable
```

keps/4803-workload-level-wait-for-pods-ready/README.md
```diff
@@ -0,0 +1,273 @@
...
+the workload is forbiden from being admitted.
```
kannon92:
```suggestion
the workload is forbidden from being admitted.
```

keps/4803-workload-level-wait-for-pods-ready/README.md
kannon92: This sounds good to me.

keps/4803-workload-level-wait-for-pods-ready/README.md
```diff
@@ -0,0 +1,273 @@
...
+- New `WorkloadSpec` fields and resource annotations implemented.
+- Unit tests for annotation parsing, webhook validation (duration, immutability),
+  and the Deployment Pod-template exception added.
+- Integration tests covering precedence, cluster-wide fallback, RecoveryTimeout
+  behaviour, and Job/Deployment propagation added.
+
+#### Beta
+
+- Feature gate enabled by default.
+- E2E tests added.
```
kannon92: I would prefer e2e tests added in alpha.

kannon92 (commented):
I just have some nits.

/lgtm

keps/4803-workload-level-wait-for-pods-ready/kep.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - "/keps/349-all-or-nothing"
+
+stage: alpha
+
+latest-milestone: "v0.20"
+
+milestone:
+  alpha: "v0.20"
+  beta: "v0.21"
+  stable: "v0.22"
```
tenzen-y:
```suggestion
```

Let's make the future stage empty.

keps/4803-workload-level-wait-for-pods-ready/README.md
```diff
@@ -0,0 +1,273 @@
...
+```go
+// +kubebuilder:validation:MinProperties=1
+type WaitForPodsReady struct {
+    // podsReadyTimeout defines the maximum duration the workload may remain
+    // admitted before all pods are in a Ready or Succeeded state.
+    // When elapsed, the workload is evicted with reason PodsReadyTimeout.
+    // If both this field and the cluster-wide WaitForPodsReady.Timeout are set,
+    // this field takes precedence.
+    // +optional
+    PodsReadyTimeout *metav1.Duration `json:"podsReadyTimeout,omitempty"`
```
tenzen-y:
```suggestion
    TimeoutSeconds *int64 `json:"timeoutSeconds,omitempty"`
```

The API name is redundant. The proposed name indicates `podsReadyTimeout ' of ' WaitForPodsReadyTimeout". It should be avoided.

Also, using duration is obviously an anti-pattern in a Custom Resource. We are using it in Configuration since it is not CR.

keps/4803-workload-level-wait-for-pods-ready/README.md
```diff
@@ -0,0 +1,273 @@
...
+}
+
+```go
+// +kubebuilder:validation:MinProperties=1
+type WaitForPodsReady struct {
+    // podsReadyTimeout defines the maximum duration the workload may remain
+    // admitted before all pods are in a Ready or Succeeded state.
+    // When elapsed, the workload is evicted with reason PodsReadyTimeout.
+    // If both this field and the cluster-wide WaitForPodsReady.Timeout are set,
+    // this field takes precedence.
```
tenzen-y: The unlimited timeout is too risky. At least we should add `0 <= timeoutSeconds <= KueueConfiguration.WaitForPodsReady.Timeout` validation.

tenzen-y (commented):
@MaysaMacedo I haven't checked everything. I just left the first review comments.

keps/4803-workload-level-wait-for-pods-ready/README.md
MaysaMacedo: @tenzen-y Setting the maximum allowed value as the cluster wide timeout, might contradict the story 2 scenario.
I thought the idea was to allow values that can be different from the cluster wide one.
For example, the cluster timeout defaults to 30 minutes and the user might need to set a timeout per workload of 1hr. Perhaps we could define a different value for the maximum? Maybe 12/24hrs? Not sure what would be a value that makes sense here.

tenzen-y: I still thinkg cluster-level default should be cap of workload level timeout since cluster admin can specify the safe cap based on their cluster env.

Otherwise, users can specify an unsafe timeout, and the safe timeout fully depends on their env. So, we can not provide the comprehensive safe timeout in upstream.

We can revisit whether we should provide a configuration API for the cap (e.g., waitForPodsReady.maxTimeoutOnWorkload) that users can specify in the Workload timeout.
But, I don't want to introduce it in the first alpha release.

keps/4803-workload-level-wait-for-pods-ready/README.md
```diff
@@ -0,0 +1,273 @@
...
+  that is propagated to the corresponding `WorkloadSpec` field at Workload creation time.
+  For Deployment, the annotation needs to be defined at `spec.template.metadata.annotations`
+  to be propagated to the Pods and then workloads.
+- Per-workload timeouts take precedence over the global `WaitForPodsReady` timeout configuration when
+  both are present.
+
+### Non-Goals
+
+- Change the cluster-wide `WaitForPodsReady` behaviour.
+- Change the recovery / backoff strategy; that remains cluster-wide only
```
sohankunkerkar, line 67: I think `recoveryTimeout` isn't independent of timeout though, `defaults.go` does `cmp.Or(RecoveryTimeout, &Timeout)`. So a workload with a 1h per-workload timeout on a cluster set to 30m gets an hour to start but only 30 minutes to recover. Worth saying whether recovery should follow the per-workload value, or spelling out that it doesn't.

keps/4803-workload-level-wait-for-pods-ready/README.md
```diff
@@ -0,0 +1,273 @@
...
+As a job owner whose workload requires pulling a large container image that takes longer than the cluster-wide WaitForPodsReady timeout, I want to annotate my job with a longer per-job timeout so it is not evicted prematurely, without requiring the platform administrator to relax the global timeout for everyone.
+
+#### Story 3
+
+As a platform administrator, I want to configure Kueue with `blockAdmission: true` and assign per-workload pods-ready timeouts so that each workload is evicted after its own appropriate timeout rather than a single cluster-wide timeout. This ensures that a workload whose pods never become ready is evicted promptly and unblock admission of workloads in other ClusterQueues.
+
+### Notes/Constraints/Caveats
+
+- When a per-workload timeout is enforced and `DisableWaitForPodsReady` feature gate is enabled
+the workload is forbbiden from being admitted.
```
sohankunkerkar, line 106: This flips what the gate does now. `waitForPodsReady()` returns nil when `DisableWaitForPodsReady` is on, so the feature just goes quiet and nothing gets rejected. If we block admission instead, an admin flipping the escape hatch during an incident can't admit anything with the annotation. Could we ignore the field instead?

keps/4803-workload-level-wait-for-pods-ready/README.md
sohankunkerkar: Story 2 is "my image pull takes longer than the cluster timeout", so a cap at the cluster value removes it. But one number can't be both the default and the max. Separate `maxTimeout` in Configuration? Otherwise we should drop Story 2. Whichever way this lands, the Risks section needs updating too, it currently says per-workload values always take precedence.

keps/4803-workload-level-wait-for-pods-ready/README.md
```diff
@@ -0,0 +1,273 @@
...
+#### Story 1
+
+As a batch platform administrator, I want to configure a tight pods-ready
+timeout for data-preprocessing jobs (2 minutes) while allowing large ML
+training jobs a longer timeout (30 minutes), without requiring two separate
+Kueue deployments.
+
+#### Story 2
+
+As a job owner whose workload requires pulling a large container image that takes longer than the cluster-wide WaitForPodsReady timeout, I want to annotate my job with a longer per-job timeout so it is not evicted prematurely, without requiring the platform administrator to relax the global timeout for everyone.
```
sohankunkerkar, line 97: xref: https://github.com/kubernetes-sigs/kueue/pull/12776#discussion_r3901057749

keps/4803-workload-level-wait-for-pods-ready/README.md
MaysaMacedo: @tenzen-y I understand your concern. But, from the initial issue reported we have:

> Certain workloads may take longer to get to a ready state than others, so configuring this on at a cluster-wide scope feels awkward.

But if we define the Maximum as the cluster wide timeout, I think we would end up having the follow statement as the issue that we're addressing:

`Certain workloads may take less time to get to a ready state than others.`

Would that now be the purpose of the timeout on the workload level? 

If we don't use that specific existent cluster Timeout as max and instead define another one as the Max we could still think the workload timeout as a way to only tune workloads that takes longer.

@sohankunkerkar Why would the per-workload values always take precedence need to be updated?

keps/4803-workload-level-wait-for-pods-ready/README.md
MaysaMacedo: @sohankunkerkar Is your idea that we should just ignore the timeout annotation that is set when `DisableWaitForPodsReady` is enabled?

@kannon92 I think Kevin's suggestion initially was to block admission because we would be interested in knowing if anyone is using this feature gate or not.

keps/4803-workload-level-wait-for-pods-ready/README.md
sohankunkerkar: >@sohankunkerkar Why would the per-workload values always take precedence need to be updated?

It's written as the mitigation, but it's the part that creates the risk as anyone can set the annotation, so "user value always wins" is what allows a very long timeout to hold quota.

And if we add a cap, _always_ won't be accurate anymore, since the workload value would only apply up to the admin's limit.

keps/4803-workload-level-wait-for-pods-ready/README.md
sohankunkerkar: Yes, ignore the annotation when the gate is on. `PriorityBoost` already works that way: with the gate off the annotation is ignored, while an invalid value still gets rejected by the webhook. Rejecting because a feature is disabled is a different thing from rejecting bad input.

On Kevin's point: `DisableWaitForPodsReady` is an escape hatch, so if an admin turns it on during an incident, blocking admission means nothing carrying the annotation can be admitted at all, the opposite of what the switch is for. Ignoring costs nothing here, the workload just falls back to the cluster-wide behavior.

@kannon92 if the goal is knowing whether the gate is used, that's probably better as its own issue e.g. exposing feature gate enablement as a metric would cover all gates instead of special-casing this one.

sohankunkerkar:
@coderabbitai can you file a tracking issue for the question I asked to @kannon92 in this https://github.com/kubernetes-sigs/kueue/pull/12776#discussion_r3907069237? Please don't assign that issue to anyone and add a label kind/feature.

keps/4803-workload-level-wait-for-pods-ready/README.md
kannon92: Maybe we could have a warning at least? I was thinking of blocking to inform user of an incompatible change rather than just ignoring.

MaysaMacedo:
@mimowo Hello, could you take a look at this KEP as well? Thanks so much.

mimowo:
Sure, but it would be great if you can get lgtm from @sohankunkerkar

keps/4803-workload-level-wait-for-pods-ready/README.md
mimowo: I assume most people may have timeout 30min. So users will want it to have 3min, but some will need 1h. If we cap by global timeout, then admin needs to set 1h, which is not good as a global default. 

So, I think the per-workload timeout needs to be able to exceed the global timeout.

> waitForPodsReady.maxTimeoutOnWorkload

That feels technically correct, but a bit of an overkill. I think even for Alpha and Beta I would assume capping the max per-workload timeout to 2h, wdyt? It will limit the feature, but ensure the values are safe. We may revisit supporting longer max timeout (or configurable) if such use-cases arise, but I think 2h should be generally acceptable as max.

MaysaMacedo: 2hrs as a CAP for per workload timeout looks good to me. @sohankunkerkar @tenzen-y what do you think?

tenzen-y: Fixed timeout cap (even 1 hours , 2 hours , 3 hours...) doesn't look good to me. The safe wait for pods ready cap depends on the infrastructure. So, I would request cluster-level waitForPodsReady.timeout should be the cap or introducing maxTimeoutOnWorkload.

But I'm wondering if we can start using cluster-level waitForPodsReady.timeout for workload cap.

If we have a concrete use case, I'm ok w/ introducing maxTimeoutOnWorkload in the first phase.

MaysaMacedo: @tenzen-y what do you think about the topic I mentioned here https://github.com/kubernetes-sigs/kueue/pull/12776#discussion_r3906759595 ?

tenzen-y: > If we don't use that specific existent cluster Timeout as max and instead define another one as the Max we could still think the workload timeout as a way to only tune workloads that takes longer.

I think even with shorter timeouts, workload-level timeout annotations can still be leveraged.
I imagine that cluster operator specify timeout for generic workload (say 60% workloads) in Kueue Configuration, and some special workloads (like special infrastracture or accelerators) will specify customizd timeout anntation in their Workload, but it will be capped by cluste level timeout or maxTimeoutOnWorkload field.

tenzen-y: Why I keep claiming the cluster-level configurable cap regardless of this feature maturity is discussed annotation is batch-user facing one and they can hijack and break the multitenant cluster w/ dangerous timeout immidiately.

mimowo: >  batch-user facing one and they can hijack and break the multitenant cluster w/ dangerous timeout immidiately.

This I don't understand - how would setting high value, say 10days make it possible for the user to "hijack and break the multitenant cluster w/ dangerous timeout". Can you elaborate on that?

I understand the workload may stay pending for long, but so what, the workload would also stay blocking quota for long if the user just had a very simple Sleep(10h) code, because the pods would be ready.

Also, bare k8s doesn't have such timeouts and it isn't considered volnurable to hijacking. 

Finally, in a well configured Kueue there is isolation and fair sharing between teams. Even if one team reserves all available quota (no matter if the pods are running or not) the impact on other teams is limited by the fair sharing rules.

The only reason I see for the max-cap is to make the validation showing the usage intention, and I think the 2h cap conveys the intention well.

tenzen-y: I don't mention about exactly timeout value. They can break prod regardless of timeout value which fully depends on infrastructure.

tenzen-y: The point is the cap should be specified by cluster admins based on their infra.

MaysaMacedo: Does it make sense to add a new `maxTimeoutOnWorkload` in the cluster config and set a 2hr default for it or should we go with a different approach?

mimowo: I don't see a need, but I'm ok  with it. I would default it certainly to make the configuration sensible ootb

mimowo: I synced with @tenzen-y offline and we agree to have:
1. for alpha maxTimeoutOnWorkload is hardcoded to 1h
2. for beta maxTimeoutOnWorkload is a new ConfigAPI knob defaulted to 1h, but admins can change

wdyt?

MaysaMacedo: Could we make the maximum allowed at least 2hours? or could we make the configurable API something available on Alpha?

mimowo: I'm ok either with hardcoded 2h for alpha, configurable API for Alpha, any preference @tenzen-y ?

tenzen-y: > I'm ok either with hardcoded 2h for alpha, configurable API for Alpha, any preference @tenzen-y ?

I'm ok with either approach. My request is only to disable the annotation evaluations by default in the alpha phase, regardless of the approach.

MaysaMacedo: @tenzen-y @mimowo Do you have any preferences on where this validation should happen? would it be at the job webhook or workload webhook  level?

mimowo: Job's webhook makes sense to me, but we should de-duplicate the code there between CRDs

tenzen-y: Validating parameters on Job webhook sounds reasonable. We can commonize the validation code in pkg/controller/jobframework.

keps/4803-workload-level-wait-for-pods-ready/README.md
```diff
@@ -0,0 +1,273 @@
...
+// +kubebuilder:validation:MinProperties=1
+type WaitForPodsReady struct {
+    // timeoutSeconds defines the maximum time the workload may remain
+    // admitted before all pods are in a Ready or Succeeded state.
+    // When elapsed, the workload is evicted with reason PodsReadyTimeout.
+    // If both this field and the cluster-wide WaitForPodsReady.Timeout are set,
+    // this field takes precedence.
+    // +optional
+    // +kubebuilder:validation:Minimum=1
+    TimeoutSeconds *int64 `json:"timeoutSeconds,omitempty"`
```
MaysaMacedo, line 140: @mimowo @tenzen-y something I noticed, most of the time related fields under Workload use a type of `int32`, should I use that instead? And also update the annotation to accept a number instead of a Duration type?
I'm just concerned if we add the new cluster configuration of MaxTimeoutOnWorkload of type Duration, there could be some mismatch on the the length accepted in that field and the length accepted on the workload.spec timeout.

mimowo: Hmmm, I see, I think for alpha we could just copy the annotation from Job to workload to avoid strong typing, wdyt @tenzen-y ?

If we go with full fledged API for this then Duration should not be used, because API is exposed to other clients, and some clients is other languages may have problems parsing the syntax, so duration is generally discouraged by the API guidelines.

So, if we go with full API, then TimeoutSeconds, and RecoveryTimeoutSeconds as int32 make sense.

However, to minimize chances of API bugs I would like to consider just copyging the annotation in Alpha.

MaysaMacedo: Note that annotation are represented as a String, so it's a matter of deciding what we convert it to internally, either Duration type or an int. Given the workload.TimeoutSeconds is represented as an `int`, probably I will just convert to an `int` and also very likely I will keep it set as `int64`, given `Duration` is of type `int64`.

mimowo: > Note that annotation are represented as a String, so it's a matter of deciding what we convert it to internally, either Duration type or an int.

We should use int internally, to enable convering the annotation into the Workloads spec fields

kannon92: I think the linter may complain about int versus int64. The reason is that on different architecture the size of int can change. Hence why the recommendation is int32 or int64.

keps/4803-workload-level-wait-for-pods-ready/README.md
MaysaMacedo: I updated the KEP with the idea of adding an optional configuration to control the max on the annotation.

keps/4803-workload-level-wait-for-pods-ready/README.md
MaysaMacedo: I'm setting to use int64.

keps/4803-workload-level-wait-for-pods-ready/README.md
```diff
@@ -0,0 +1,297 @@
...
+  string is rejected at admission time."
+
+##### webhooks/deployment
+
+- "A Deployment whose Pod template carries kueue.x-k8s.io/pods-ready-timeout
+  with a valid duration is accepted."
+- "A Deployment whose Pod template carries kueue.x-k8s.io/pods-ready-timeout
+  with an invalid value is rejected at admission time."
+
+#### Integration tests
```
kannon92, line 250: I notice you are missing e2e tests.

Is the intention that all the testing should be covered by integration?

if so, we should at least mention why no e2e tests are needed. Or if this is an oversight, it would be good to think through what e2e tests we want to add.

keps/4803-workload-level-wait-for-pods-ready/README.md
```diff
@@ -0,0 +1,297 @@
...
+
+- **Conflicting values**: A user may set a per-workload timeout that conflicts
+  with a cluster-wide one.  Mitigation: per-workload values always take
+  precedence; a clear precedence rule is documented and enforced in the webhook.
+
+## Design Details
+
+### API
+
+#### Workload Spec
```
mimowo, line 118: Please also add the section with the new JSON-based annotation format.

mimowo, line 118: As we introduce the JSON-based annotation should we defer the WorkloadSpec API changes from Alpha1 (v0.20), and do them in Alpha2 (v0.21)?

This will give us more time to properly design the API. For the first iteration we could focus on the mechanism, and just copy the annotation from Job to the Workload, wdyt?

keps/4803-workload-level-wait-for-pods-ready/README.md
```diff
@@ -0,0 +1,297 @@
...
+
+The `WaitForPodsReady` struct and its `TimeoutSeconds` field are added to
+`WorkloadSpec` in **v1beta2 only**.
```
tenzen-y, line 121:
```suggestion
```

Please add it to v1beta1 as well, since v1beta1 is still provided

keps/4803-workload-level-wait-for-pods-ready/README.md
```diff
@@ -0,0 +1,297 @@
...
+
+#### Resource Annotations
+
+One new annotation is defined in `pkg/controller/constants/constants.go`:
+
+```go
+// WaitForPodsReadyTimeoutSecondsAnnotation is the annotation key on any Kueue-managed resource that sets a
+	// per-workload pods-ready timeout, overriding the cluster-wide WaitForPodsReady.Timeout.
+	// Value must be an integer number of seconds.
+	WaitForPodsReadyTimeoutSecondsAnnotation = "kueue.x-k8s.io/wait-for-pods-ready-timeout-seconds"
```
mimowo, line 175:
```suggestion
	WaitForPodsReadyTimeoutSecondsAnnotation = "kueue.x-k8s.io/wait-for-pods-ready"
```
Let's make the annotation JSON-based, because it will encode more configuration fields. 
I propose the format similar as we decide here: https://github.com/kubernetes-sigs/kueue/pull/13787#issuecomment-5429503717

So something like:

```
"kueue.x-k8s.io/wait-for-pods-ready" = {
  "timeout": "5m",
  "recoveryTimeout: 3m"
}
```
etc

keps/4803-workload-level-wait-for-pods-ready/README.md
tenzen-y: @MaysaMacedo, based on https://github.com/kubernetes-sigs/kueue/pull/12776/changes#r3934818357 comment, we considered annotation propagation from Job to Workload instead of a dedicated Workload field during alpha stage. Am I missing any discussions?

keps/4803-workload-level-wait-for-pods-ready/README.md
```diff
@@ -0,0 +1,297 @@
...
+```
+
+### Controller
+
+#### Workload
+
+`admittedNotReadyWorkload` in `pkg/controller/core/workload_controller.go` is
+updated to resolve the effective timeout exclusively from `wl.Spec.WaitForPodsReady.TimeoutSeconds`,
+falling back to the cluster-wide config only when that field is nil.
+The eviction deadline is resolved exclusively from `wl.Spec.WaitForPodsReady.TimeoutSeconds`
```
mimowo, line 185: reads like some duplication, no?

keps/4803-workload-level-wait-for-pods-ready/README.md
tenzen-y: SGTM on JSON annotation.

mimowo (commented):
The KEP looks solid overall, and I want to include it, but the freeze is approaching, and so I would propose to:
1. for alpha1 in v0.20:
- only introduce the JSON-based annotation: "kueue.x-k8s.io/wait-for-pods-ready"
- hard-code the max MaxTimeoutOnWorkload to 2h
- no API changes in Workload's spec.waitForPodsReady and ConfigAPI
- implement the support for .timeout and .recoveryTimeout
2. for alpha2 in v0.21
- extend the support for .unscheduledTimeout
- introduce the Workload's spec.waitForPodsReady
- introduce the ConfigAPI for MaxTImeoutOnWorkload
3. for Beta in v0.22
- move to Beta (enabled by default)

Hope this is ok for you, but this will vastly de-risk merging API issues, and will allow you to focus on the implementation of the mechanism to deliver for v0.20, hopefully.

Wdyt @MaysaMacedo @kannon92 @sohankunkerkar @tenzen-y ?

tenzen-y:
> The KEP looks solid overall, and I want to include it, but the freeze is approaching, and so I would propose to:
> 
> 1. for alpha1 in v0.20:
> 
> * only introduce the JSON-based annotation: "kueue.x-k8s.io/wait-for-pods-ready"
> * hard-code the max MaxTimeoutOnWorkload to 2h
> * no API changes in Workload's spec.waitForPodsReady and ConfigAPI
> * implement the support for .timeout and .recoveryTimeout
> 
> 2. for alpha2 in v0.21
> 
> * extend the support for .unscheduledTimeout
> * introduce the Workload's spec.waitForPodsReady
> * introduce the ConfigAPI for MaxTImeoutOnWorkload
> 
> 3. for Beta in v0.22
> 
> * move to Beta (enabled by default)
> 
> Hope this is ok for you, but this will vastly de-risk merging API issues, and will allow you to focus on the implementation of the mechanism to deliver for v0.20, hopefully.
> 
> Wdyt @MaysaMacedo @kannon92 @sohankunkerkar @tenzen-y ?

I also asked a part of things here https://github.com/kubernetes-sigs/kueue/pull/12776#discussion_r3937409259

So, I agree with the proposed graduation steps.

kannon92:
I am not a fan of struct-based json. We have been designing this feature with a customer and the main ask from them was overriding the timeouts. 

I am not confident in the long term plan for `waitForPodsReady` with WAS. So some of these features I can't really say how they will behave over time.

tenzen-y:
> I am not a fan of struct-based json. We have been designing this feature with a customer and the main ask from them was overriding the timeouts.
> 
> I am not confident in the long term plan for `waitForPodsReady` with WAS. So some of these features I can't really say how they will behave over time.

@kannon92 The point is JSON is only way to input any multiple parameters from Jobs.
Or does that mean you want to discontinue this KEP?

kannon92:
>  The point is JSON is only way to input any multiple parameters from Jobs.

No. I mostly want to keep the KEP as is. We asked the customer about the different fields for WaitForPodsReady and the main ask was timeout. 

We haven't really see a lot of use cases for other fields in WaitForPodsReady at the user level so we didn't think to expand to those other fields.

tenzen-y:
> > The point is JSON is only way to input any multiple parameters from Jobs.
> 
> No. I mostly want to keep the KEP as is. We asked the customer about the different fields for WaitForPodsReady and the main ask was timeout.
> 
> We haven't really see a lot of use cases for other fields in WaitForPodsReady at the user level so we didn't think to expand to those other fields.

I understand your customer demand is only for timeout, for sure.
But this is an upstream repository. So, I would avoid to gradually expand the single annotations in the future like

Now `kueue.x-k8s.io/wait-for-pods-ready-timeout-seconds` is added,
In future releases, `kueue.x-k8s.io/wait-for-pods-ready-recovery-timeout-seconds` and `kueue.x-k8s.io/wait-for-pods-ready-unschedulable-pods-timeout-seconds` will be added.

tenzen-y:
> I am not confident in the long term plan for waitForPodsReady with WAS. So some of these features I can't really say how they will behave over time.

For WAS integration, I think wait for pods ready is still needed since kube-scheduler and Kueue uses a separate cluster snapshot which cause the scheduling result friction between 2 components. So, the only way to eliminate wait for pods ready from Kueue is kube-scheduler expose its live-snapshot via RPC, but it was rejected due to kube-scheduler perf regressions.

kannon92:
> I understand your customer demand is only for timeout, for sure.
But this is an upstream repository. So, I would avoid to gradually expand the single annotations in the future like
Now kueue.x-k8s.io/wait-for-pods-ready-timeout-seconds is added,
In future releases, kueue.x-k8s.io/wait-for-pods-ready-recovery-timeout-seconds and kueue.x-k8s.io/wait-for-pods-  ready-unschedulable-pods-timeout-seconds will be added.

This is borderline YANGI. If we get use cases for future reasons to override, let people come with a feature. The `WaitForPodsReady` struct allows folks to add future annotations and we can reason through validation on each field.

If we had a generic json annotation, testing and support becomes challenging. I'm also not sure how json based annotations works for different clients (python, javascript, golang) which is a common use case in Kubeflow. 

Its easier to reason through the validation of annotations when you can validate the explicit type. I guess for json we basically won't validate and pass the values as is?

tenzen-y:
> > I understand your customer demand is only for timeout, for sure.
> > But this is an upstream repository. So, I would avoid to gradually expand the single annotations in the future like
> > Now kueue.x-k8s.io/wait-for-pods-ready-timeout-seconds is added,
> > In future releases, kueue.x-k8s.io/wait-for-pods-ready-recovery-timeout-seconds and kueue.x-k8s.io/wait-for-pods-  ready-unschedulable-pods-timeout-seconds will be added.
> 
> This is borderline YANGI. If we get use cases for future reasons to override, let people come with a feature. The `WaitForPodsReady` struct allows folks to add future annotations and we can reason through validation on each field.

Actually, we already have for v0.21: https://github.com/kubernetes-sigs/kueue/pull/14041#discussion_r3936834345

> > If we had a generic json annotation, testing and support becomes challenging. I'm also not sure how json based annotations works for different clients (python, javascript, golang) which is a common use case in Kubeflow.
> 
> Its easier to reason through the validation of annotations when you can validate the explicit type. I guess for json we basically won't validate and pass the values as is?

We will have validation webhook for the JSON annotations.

tenzen-y:
@kannon92 https://github.com/kubernetes-sigs/kueue/blob/cabf9482e066d208ebf482a7ae9765570c60bde9/pkg/controller/jobframework/tas_validation.go#L201-L214 can help to understand how to validate JSON anntations on Kueue side.

keps/4803-workload-level-wait-for-pods-ready/README.md
MaysaMacedo: Thanks for bringing this to my attention.
I guess I didn't fully get the idea initially of just copying from annotation directly, but now I understand it.
As soon as the current reviewers have a chance to look at the discussions, I will update the KEP with what is agreed.

kannon92:
Thinking back on this, I'm fine to continue with annotations. I do worry how much this project is depending on annotations with no real plan to phase them out for proper apis. Annotations should be alpha-level fields but it seems we are making annotations first class API fields.

Annotations for alpha features are explictly called out in our [Kubernetes api conventions](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#label-selector-and-annotation-conventions).

I'll accept the plan laid out in https://github.com/kubernetes-sigs/kueue/pull/12776#pullrequestreview-5117247421.

tenzen-y:
> Thinking back on this, I'm fine to continue with annotations. I do worry how much this project is depending on annotations with no real plan to phase them out for proper apis. Annotations should be alpha-level fields but it seems we are making annotations first class API fields.
> 
> Annotations for alpha features are explictly called out in our [Kubernetes api conventions](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#label-selector-and-annotation-conventions).
> 
> I'll accept the plan laid out in [#12776 (review)](https://github.com/kubernetes-sigs/kueue/pull/12776#pullrequestreview-5117247421).

I fully agree w/ you. I also don't like annotation / label apporach. Personally, I believe annotations are the final weapon. However, in reality, there is no way to represent such parameters in Kuberentes API. 

For sure, if WAS / TAS become mature, we can push the Kueue annotations to WAS / TAS API fields. But for now, WAS / TAS lacks too many capabilities from Kueue PoV. The delegation will progress slowly rather than the Kueue development cycle.

kannon92:
> For sure, if WAS / TAS become mature, we can push the Kueue annotations to WAS / TAS API fields. But for now, WAS / TAS lacks too many capabilities from Kueue PoV. The delegation will progress slowly rather than the Kueue development cycle.

There hasn't really been any discussion on something like `WaitForPodsReady` for WAS/TAS though. I remember there was one timout field for podgroup scheduling but I don't think we have anything like `WaitForPodsReady` slated in the roadmap.

For some of the TAS features I am fine to continue with annotations as I think we can agree that WAS's TASA should have some of the kueue functionality. I am not so sure on `WaitForPodsReady` though.

tenzen-y:
> > For sure, if WAS / TAS become mature, we can push the Kueue annotations to WAS / TAS API fields. But for now, WAS / TAS lacks too many capabilities from Kueue PoV. The delegation will progress slowly rather than the Kueue development cycle.
> 
> There hasn't really been any discussion on something like `WaitForPodsReady` for WAS/TAS though. I remember there was one timout field for podgroup scheduling but I don't think we have anything like `WaitForPodsReady` slated in the roadmap.
> 
> For some of the TAS features I am fine to continue with annotations as I think we can agree that WAS's TASA should have some of the kueue functionality. I am not so sure on `WaitForPodsReady` though.

I feel bringing wait for pods ready to core sounds hard since the core problem is scheduling result drift in kube-scheduler <-> kueue, autoscaler <-> kueue, and autosclaer <-> kube-scheduler.

So, the straightforward approach is having a centralized source of cluster and future-cluster (for autoscaling) snapshots to make the same scheduling decisions across all components. But I'm not sure if we can realize the tightly coupled approach.

mimowo:
> Annotations for alpha features are explictly called out in our Kubernetes api conventions.

Actually I cannot find it. @kannon92 can you quote the specific fragment saying that?

mimowo:
> I feel bringing wait for pods ready to core sounds hard since the core problem is scheduling result drift in kube-scheduler <-> kueue, autoscaler <-> kueue, and autosclaer <-> kube-scheduler.

Yeah, WaitForPodsReady triggers "requeuing" and AFAIK k8s core is not planning to introduce Queuing any time soon, so I don't see overlap currently.

MaysaMacedo:
> The KEP looks solid overall, and I want to include it, but the freeze is approaching, and so I would propose to:
> 
>     1. for alpha1 in v0.20:
> 
> 
>     * only introduce the JSON-based annotation: "kueue.x-k8s.io/wait-for-pods-ready"
> 
>     * hard-code the max MaxTimeoutOnWorkload to 2h
> 
>     * no API changes in Workload's spec.waitForPodsReady and ConfigAPI
> 
>     * implement the support for .timeout and .recoveryTimeout
> 
> 
>     2. for alpha2 in v0.21
> 
> 
>     * extend the support for .unscheduledTimeout
> 
>     * introduce the Workload's spec.waitForPodsReady
> 
>     * introduce the ConfigAPI for MaxTImeoutOnWorkload
> 
> 
>     3. for Beta in v0.22
> 
> 
>     * move to Beta (enabled by default)
> 
> 
> Hope this is ok for you, but this will vastly de-risk merging API issues, and will allow you to focus on the implementation of the mechanism to deliver for v0.20, hopefully.
> 
> Wdyt @MaysaMacedo @kannon92 @sohankunkerkar @tenzen-y ?

Initially configuring the `recoveryTimeout` at the workload level was set as a non-goal of this KEP. The KEP was just targeting configuring the `Timeout` at the workload level. Will  `recoveryTimeout` configuration also be a goal now?

kannon92:
> Actually I cannot find it. @kannon92 can you quote the specific fragment saying that?

In early versions of Kubernetes, some in-development features represented new API fields as annotations, generally with the form something.alpha.kubernetes.io/name or something.beta.kubernetes.io/name (depending on our confidence in it). This pattern is deprecated. Some such annotations may still exist, but no new annotations may be defined. New API fields are now developed as regular fields.

mimowo:
> > Actually I cannot find it. @kannon92 can you quote the specific fragment saying that?
> 
> In early versions of Kubernetes, some in-development features represented new API fields as annotations, generally with the form something.alpha.kubernetes.io/name or something.beta.kubernetes.io/name (depending on our confidence in it). This pattern is deprecated. Some such annotations may still exist, but no new annotations may be defined. New API fields are now developed as regular fields.

Hm, but my reading of this is not that annotations are deprecated, but the format which has the maturity in the name is deprecated, example something.**alpha**.kubernetes.io/name

kannon92:
> Hm, but my reading of this is not that annotations are deprecated, but the format which has the maturity in the name is deprecated, example something.alpha.kubernetes.io/name

I haven't seen a single kubernetes feature go in with annotations in the 5 years I've worked on the project. In fact, gang scheduling tried to introduce as annotations but that was rejected. Annotations are hard to deprecate and hard to validate.

Goal is to always go with API fields as that is a better long term soluition than relying on annotations.

mimowo:
> I haven't seen a single kubernetes feature go in with annotations in the 5 years I've worked on the project. 

1. I think there were some added, for example:

 I've seen these added 3  years ago without any opposition from API reviewers:

-"job-index-failure-count" and "job-index-ignored-failure-count" were added 3 years ago
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/batch/v1/types.go#L53-L61

- for DRA "resource.kubernetes.io/extended-resource-claim" and "resource.kubernetes.io/pod-claim-name" were added 6 months ago 

https://github.com/kubernetes/kubernetes/blob/5b0aabc9f33aee4dc7c58b728cb77f3e61b45483/pkg/apis/resource/types.go#L35-L43

2. even if this was true I think the paragraph is not saying "introducing annotations is depracted", it is discussing the naming of the annotations IMO.

kannon92:
I guess it isn't that clear. You can propose an update to make it clearer? My conversations with api-reviews is that user facing feature changes should not use annotations. I know that it is a red-flag for any KEP/PRR that I call out for any KEP.

The items you mention are mostly internally and we don't have examples of user facing API changes AFAIK.

mimowo:
> I guess it isn't that clear. You can propose an update to make it clearer? My conversations with api-reviews is that user facing feature changes should not use annotations. I know that it is a red-flag for any KEP/PRR that I call out for any KEP.

Yes, I agree these annotations are not really meant to use by users directly, but they are part of the API to communicate between components. 

There are some more user-facing like "cluster-autoscaler.kubernetes.io/safe-to-evict", but generally I agree - annotations are rare in core k8s.

However, for Kueue annotations are a convenient way to support features in a Job agnostic fashion. I'm not sure how to otherwise introduce the feature, or other Job-agnostic features in Kueue. Let us know if you have a constructive alternative / pattern we could use.

tenzen-y:
> I guess it isn't that clear. You can propose an update to make it clearer? My conversations with api-reviews is that user facing feature changes should not use annotations. I know that it is a red-flag for any KEP/PRR that I call out for any KEP.
> 
> The items you mention are mostly internally and we don't have examples of user facing API changes AFAIK.

@kannon92 I fully understand your concerns about user-facing annotations. But, from an out-of-tree project PoV, we don't have the capability to extend the core API (e.g.,g bringing Kueue-specific parameters to the Pod spec). The common way is having out-of-tree specific Job like KueueJob which have a set of Kueue scheduling prameters as field instead of annotations and a whole of Job spec. But I don't like such patterns since the advantage of Kueue is that users can deploy your Job w/o project-specific CR Job, which allows them to avoid deeply locking into the projects and leverage the pure K8S capabilities easily.

Instead of the project-specific CR Job (KueueJob) pattern, the mitigation is introducing a "Class" or "Parameter" capability to represent project-specific parameters.
I would imagine the following WorkloadSchedulingParameter, but it is just an example.

The WorkloadSchedulingParameter problem is (1 how to provide a migration period for end users and (2 whether we should block all scheduling-related feature development until the WorkloadSchedulingParameter are mature.

```yaml
apiVersion: kueue.x-k8s.io/v1alpha1
kind: WorkloadSchedulingParameter
metadata:
  name: alpha
  namespace: tenant-x
spec:
  waitForPodsReady:
    timeout: 60m
    unscheduledTimeout: 5m
  topologyAwareScheduling:
    topologyConstraints:
    - podSetName: worker
      required: "cloud.provider.com/topology-block"
      preferred: "cloud.provider.com/topology-rack"
---
apiVersion: batch/v1
kind: Job
metadata:
  name: foundation-model-training
  namespace: tenant-x
  labels:
    kueue.x-k8s.io/workload-scheduling-parameter: alpha
spec:
  ...
```

mimowo:
Yeah, WorkloadSchedulingParameter seems like a reasonable alternative, but it does not eliminate the need for labels / annotations - because it still needs to be referenced in a Job-agnostic fashion. However, it can eliminate the need for JSON-based annotations, granted.

I think we can consider it as an alternative pattern for users who want cleaner API, but the need to create another API object may be not ideal for some users.

kannon92:
For `WaitForPodsReady` I'm fine to continue as is. I think over time if there is gang scheduling and we reduce scheduling discrepancies I hope the need for `WaitForPodsReady` goes down. What idea we have floated internally is the use of descheduler for some of these cases (evict if pod is unscheduable etc). 

For `TopologyAwareScheduling` I hope that we continue to engage with WAS folks to make sure the needed fields for Kueue would be in scheduling API.

tenzen-y:
> Yeah, WorkloadSchedulingParameter seems like a reasonable alternative, but it does not eliminate the need for labels / annotations - because it still needs to be referenced in a Job-agnostic fashion. However, it can eliminate the need for JSON-based annotations, granted.
> 
> I think we can consider it as an alternative pattern for users who want cleaner API, but the need to create another API object may be not ideal for some users.

Yes, so this is a just "mitigation" way. For users who don't want to create additional resource, the class mechanism can mitigate their demand like

In this example, once cluster admins create WorkloadSchedulingParameterClass, kueue controller creates WorkloadSchedulingParemter resource in target namespaces automatically. Then batch-user just specify the created WorkloadSchedulingParameter name their Job label.

But these WorkloadScheduling specific resource model development needs lot of engineering effort, though.

```yaml
apiVersion: kueue.x-k8s.io/v1alpha1
kind: WorkloadSchedulingParameter
metadata:
  name: alpha
  namespace: tenant-x
spec:
  parentRef:
    name: class-alpha
---
apiVersion: kueue.x-k8s.io/v1alpha1
kind: WorkloadSchedulingParameterClass
metadata:
  name: class-alpha
  # namespace <- this is cluster-scope resouce
spec:
  ...
```

mimowo:
> For TopologyAwareScheduling I hope that we continue to engage with WAS folks to make sure the needed fields for Kueue would be in scheduling API.

Sure, we have a huge investment into integrating WAS with Kueue using the API. 

We will gradually migrate some features in Kueue to the new APIs from WAS as they arise, but we cannot stop development in Kueue until that happens.

tenzen-y:
> For `WaitForPodsReady` I'm fine to continue as is. I think over time if there is gang scheduling and we reduce scheduling discrepancies I hope the need for `WaitForPodsReady` goes down. What idea we have floated internally is the use of descheduler for some of these cases (evict if pod is unscheduable etc).

I know descheduler has a PodLifeTime plugin which can specify arbitrary eviction criteria (e.g., timeout, condition, ...).
If descheduler can satisfy the problem, why do we need this KEP? You also can leverage descheduler instead of Workload-level waitForPodsReady annotation.

But I know the descheduler can not satisfy both waitForPodsReady timeout and unscheduledTimeout since the descheduler can not trigger Kueue flavor and topology reassignment. It just evicts Pods, then the workload controller (e.g., Deployment) creates replacement Pods immediately with the same assignment.

kannon92:
I'm a little lost on `WorkloadSchedulingParemter`. Where did we have this discussion? 

Maybe we can move that discussion to a separate thread.

kannon92:
> I know descheduler has a PodLifeTime plugin which can specify arbitrary eviction criteria (e.g., timeout, condition, ...).
If descheduler can satisfy the problem, why do we need this KEP? You also can leverage descheduler instead of Workload-level waitForPodsReady annotation.

As of now, long term I'm not sure the right path for `WaitForPodsReady`. Many AI users don't seem to use descheduler and I'm not sure descheduler is aware of workload-level eviction. So I see Kueue's `WaitForPodsReady` as needed still.

I am mostyl saying we could eventually push some of the functionality of `WaitForPodsReady` into descheduler if we are looking for a more permanent home for some of these features.

tenzen-y:
> I'm a little lost on `WorkloadSchedulingParemter`. Where did we have this discussion?
> 
> Maybe we can move that discussion to a separate thread.

I stopped the discussion about WorkloadSchedulingParameters, indeed. It is too large for this KEP.
Yes, we can consider it in a separate space.

tenzen-y:
> > I know descheduler has a PodLifeTime plugin which can specify arbitrary eviction criteria (e.g., timeout, condition, ...).
> > If descheduler can satisfy the problem, why do we need this KEP? You also can leverage descheduler instead of Workload-level waitForPodsReady annotation.
> 
> As of now, long term I'm not sure the right path for `WaitForPodsReady`. Many AI users don't seem to use descheduler and I'm not sure descheduler is aware of workload-level eviction. So I see Kueue's `WaitForPodsReady` as needed still.
> 
> I am mostyl saying we could eventually push some of the functionality of `WaitForPodsReady` into descheduler if we are looking for a more permanent home for some of these features.

I see. So we might be able to implement waitForPodsReady in descheduler as a Kueue integration, since descheduler should issue Kueue-Eviction instead of Pod eviction to trigger Kueue requeueing and topology assignment.

mimowo:
Interesting, is descheduler workload-aware? Indeed, if it is, then it might be possible technically. would be good to know what is the timeline for that.

tenzen-y:
> Interesting, is descheduler workload-aware? Indeed, if it is, then it might be possible technically. would be good to know what is the timeline for that.

No workload-aware for now. But the descheduler has a plugin mechanism. So, I guess that we can add Kueue or WAS plugin into the descheduler.

mimowo:
I see, so that would be a new project and binary say kueue-descheduler which depends on Kueue (so that it can trigger the eviction). Sounds like overkill to me, but we can consider this if there are some people willing to invest in that solution.

tenzen-y:
> I see, so that would be a new project and binary say kueue-descheduler which depends on Kueue (so that it can trigger the eviction). Sounds like overkill to me, but we can consider this if there are some people willing to invest in that solution.

I'm not exactly sure what the right way is to extend descheduler, but we have 2 barriers (1 support workload-aware descheduling and (2 support triggering Kueue eviction. So, it's not an easy way anyway.

keps/4803-workload-level-wait-for-pods-ready/README.md
sohankunkerkar: I don't think we need immutability here. `admittedNotReadyWorkload` recomputes the deadline every reconcile from the Admitted condition's timestamp, and that doesn't move while the workload stays admitted. So changing the annotation mid-flight can't extend anything.

keps/4803-workload-level-wait-for-pods-ready/README.md
sohankunkerkar: A few things on the cap. The webhook checks it on the Job, but the timeout is read from `wl.Spec.WaitForPodsReady.TimeoutSeconds` and nothing checks it there. We validate priority-boost and admission-gated-by on the Workload, so why is this one different? 

MultiKueue copies the spec to the worker, and the worker has its own `MaxTimeoutOnWorkload`. Which one applies?

keps/4803-workload-level-wait-for-pods-ready/README.md
sohankunkerkar: Yup, +1 to it. Warn and ignore the timeout gets you the signal without that.

sohankunkerkar:
> The KEP looks solid overall, and I want to include it, but the freeze is approaching, and so I would propose to:
> 
> 1. for alpha1 in v0.20:
> 
> * only introduce the JSON-based annotation: "kueue.x-k8s.io/wait-for-pods-ready"
> * hard-code the max MaxTimeoutOnWorkload to 2h
> * no API changes in Workload's spec.waitForPodsReady and ConfigAPI
> * implement the support for .timeout and .recoveryTimeout
> 
> 2. for alpha2 in v0.21
> 
> * extend the support for .unscheduledTimeout
> * introduce the Workload's spec.waitForPodsReady
> * introduce the ConfigAPI for MaxTImeoutOnWorkload
> 
> 3. for Beta in v0.22
> 
> * move to Beta (enabled by default)
> 
> Hope this is ok for you, but this will vastly de-risk merging API issues, and will allow you to focus on the implementation of the mechanism to deliver for v0.20, hopefully.
> 
> Wdyt @MaysaMacedo @kannon92 @sohankunkerkar @tenzen-y ?

Phasing works for me, though one thing worth checking first. v1beta1 has been unserved since #14558 landed, so there's one served version and no conversion to write. Does that make the spec field cheap enough for alpha1, rather than building the annotation path and replacing it in alpha2? PreemptionGates went in as a v1beta2 field with its gate alpha and off, so there's precedent either way.

Long side: @kaushik229 showed one not-ready workload blocks admission everywhere with `blockAdmission: true`. He raised it as motivation, but it also means a long per-workload timeout sets how long everyone else waits. A hardcoded 2h is a two-hour cluster-wide stall any tenant can set. It's off by default, and Story 3 is that config.

Short side: `RequeueState.Count` only goes up, and is cleared only on deactivation. With `requeuingBackoffLimitCount` set, a short timeout burns a budget that never refills and the workload gets deactivated. #13502 reuses the same requeuing strategy, so two timers will share one budget. I'd bound both, and use the cluster Timeout as the ceiling when blockAdmission is on.

`recoveryTimeout` in alpha1 is right. RecoveryTimeout defaults to Timeout in defaults.go, so a per-workload start timeout alone gives a 2h workload on a 30m cluster two hours to start and thirty minutes to recover. #13502 already specified that row of the deadline table, so align to it rather than redesigning it. The recovery Non-Goal should go.

Two for the rewrite. I don't think the Pod-group path propagates the way the KEP describes, since Kueue builds the group's Workload on a path that copies no annotations and reads whichever pod triggered the reconcile, and Deployments go through it. And the descheduler discussion above is the clearest justification this KEP has, that descheduler evicts pods but can't trigger flavor or topology reassignment, so replacements land on the same assignment. Worth writing into an Alternatives section.

MaysaMacedo:
@kannon92 @mimowo @tenzen-y:
@sohankunkerkar Brought up some interesting points on the max/min values allowed. Do you guys have any specific inputs for his [last comment](https://github.com/kubernetes-sigs/kueue/pull/12776#issuecomment-5591269586)?

mimowo:
> @sohankunkerkar Brought up some interesting points on the max/min values allowed. Do you guys have any specific inputs for his https://github.com/kubernetes-sigs/kueue/pull/12776#issuecomment-5591269586?

I think we are aligned on all points, except maybe "Story 3"

> Long side: @kaushik229 showed one not-ready workload blocks admission everywhere with blockAdmission: true. He raised it as motivation, but it also means a long per-workload timeout sets how long everyone else waits. A hardcoded 2h is a two-hour cluster-wide stall any tenant can set. It's off by default, and Story 3 is that config.

IIUC the story 3 is to evict workloads waiting due to "blockAdmission: true"? This might be reasonable, but IIUC this is not how current WaitForPodsReady works. Sure, we can evaluate such a proposal, but I would prefer to scope this proposal to reuse the mechanism we have, just scope the timeouts per workload.

keps/4803-workload-level-wait-for-pods-ready/README.md
```diff
@@ -0,0 +1,297 @@
...
+training jobs a longer timeout (30 minutes), without requiring two separate
+Kueue deployments.
+
+#### Story 2
+
+As a job owner whose workload requires pulling a large container image that takes longer than the cluster-wide WaitForPodsReady timeout, I want to annotate my job with a longer per-job timeout so it is not evicted prematurely, without requiring the platform administrator to relax the global timeout for everyone.
+
+#### Story 3
+
+As a platform administrator, I want to configure Kueue with `blockAdmission: true` and assign per-workload pods-ready timeouts so that each workload is evicted after its own appropriate timeout rather than a single cluster-wide timeout. This ensures that a workload whose pods never become ready is evicted promptly and unblock admission of workloads in other ClusterQueues.
```
mimowo, line 94: I'm not sure how is the story supported by the KEP. I would like to scope the KEP to only reuse the current semantics of WaitForPodsReady, but do it "per-workload". It seems to me the story warrants bigger semantic changes to waitForPodsReady, no?

MaysaMacedo: hmm, I don't think so. What is being mentioned is the following scenario:
someone set `blockAdmission:true` on the cluster-wide `waitForPodsReady` config and set the Timeout per workload to the max, which would be 2 hours. Wouldn't this mean that all the other workload would be blocked until the 2 hours if the workload doesn't become ready before?

mimowo: Yeah, I have difficulty parsing the story TBH, maybe it doesn't but some re-phrasing to make it clear would be useful.

IIUC the idea is rather to make the per-workload timeouts short so that they can release the quota ASAP instead of waiting the default 30min. This is particularly important when `blockAdmission: true` is used, because it makes admission sequential. 

If my understanding is correct, then indeed the story can be met without extending the mechanism, but I would love some clarififcation here.

MaysaMacedo: Should the per-workload timeout be shorter than the cluster-wide timeout? 
Previously we have discussed about setting the maximum of the timeout on the workload to be hard-coded to 2 hours https://github.com/kubernetes-sigs/kueue/pull/12776#pullrequestreview-5117247421.

mimowo: I think that depends on the story. 

I actually imagine two stories basically, assuming global is 30min:
1. short timeout (say 5min) - as a user I want my faulty workload to release quota ASAP to make room for other valid workloads in queue 
2. long timeout (say 1h) - I am a platform team running a hero training with 10k pods. My workload cannot start within 30min because of the number of Pods and loading time for each of them. So, I need to have 1h, but that should not impact the global timeout.

sohankunkerkar: One thing on story 2 though -> `should not impact the global timeout` doesn't hold when `blockAdmission: true`. The scheduler waits until every admitted workload is ready or evicted, so the hero training's 1h is 1h of blocked admission for everyone else. I think that's what Story 3 was reaching for. It's a risk sitting inside your story 2 rather than a story of its own.

Story 1 has a cost too. `RequeueState.Count` only goes up and is cleared only on deactivation, so a 5min timeout keeps spending a budget that never refills, until `requeuingBackoffLimitCount` deactivates the workload.

 If both directions stay, the 2h is the only bound left. With blockAdmission: true I'd keep `maxTimeoutOnWorkload` at or below the global timeout.

keps/4803-workload-level-wait-for-pods-ready/README.md
```diff
@@ -0,0 +1,297 @@
...
+#### Workload Spec
+
+The `WaitForPodsReady` struct and its `TimeoutSeconds` field are added to
+`WorkloadSpec` in **v1beta2 only**.
+
+```go
+// WorkloadSpec defines the desired state of Workload
+type WorkloadSpec struct {
+    // ...existing fields...
+
```
mimowo: // +optional

mimowo: And also please add some comment defining the fields

mimowo: but as suggested on another thread I would prefer going with the JSON-based annotation for the Alpha1 phase. We are happy to re-evaluate the API design in next releases.

mimowo:
@MaysaMacedo  do you maybe already have a prototype for the feature?

MaysaMacedo:
> @MaysaMacedo do you maybe already have a prototype for the feature?

@mimowo I had for the previous version https://github.com/kubernetes-sigs/kueue/pull/14640
where we would include APIs.
After all the discussion we had on not including any API since last week, I will need to adapt all of that and I consider opening another draft to not lose that work, which we will eventually want to include in 0.21.

mimowo:
sgtm, I think we could still include the enhancement in 0.20 if the KEP is adjusted and the implementation.

I suppose the decision to drop API will reduce the amount of code.

MaysaMacedo:
> sgtm, I think we could still include the enhancement in 0.20 if the KEP is adjusted and the implementation.
> 
> I suppose the decision to drop API will reduce the amount of code.

@mimowo ack, is there any specific deadline for including KEPs for 0.20?

mimowo:
This was our initial plan to finalize KEPs this week and code next week:
https://github.com/kubernetes-sigs/kueue/blob/main/release-timelines/v0.20.md

mimowo:
I think we want to actually stick to the plan, with max delay of 3-4 days, because we want hard timeline for the release to be end of September.

keps/4803-workload-level-wait-for-pods-ready/README.md
mimowo: > If both directions stay, the 2h is the only bound left. With blockAdmission: true I'd keep maxTimeoutOnWorkload at or below the global timeout.

Not sure I follow, so whould you like to:
1. make `maxTimeoutOnWorkload` configurable in Alpha1
2. hardcode to 2h in Alpha1 (as currently proposed)
3. hardcode to another value in Alpha1

I'm good with any, but I want to understand the proposal

sohankunkerkar: Option 2. Sorry, that was unclear. I wasn't asking to change alpha1. `maxTimeoutOnWorkload` doesn't exist there, so the constraint I mentioned only applies in alpha2 once the field lands: with `blockAdmission: true`, don't let an admin set it above `waitForPodsReady.timeout`.

For alpha1 the hardcoded 2h is fine. Gate's off by default and so is blockAdmission, so it takes opting into both. I'd just note it in Risks saying-> with `blockAdmission: true` a single workload can hold admission cluster-wide for up to 2h, which can be longer than the admin's own timeout.

mimowo: > For alpha1 the hardcoded 2h is fine. 

cool

> Gate's off by default and so is blockAdmission, so it takes opting into both. I'd just note it in Risks saying-> with blockAdmission: true a single workload can hold admission cluster-wide for up to 2h, which can be longer than the admin's own timeout.

This part I still don't get fully - this KEP (either alpha1 or alpha2) in my mind is not changing the behavior of `blockAdmission: true`. And I always considered `blockAdmission: true` as a legacy mode since we have TAS .  `blockAdmission: true`  was introduced historically to prevent deadlocks without node awareness. `blockAdmission: false` breaks deadlocks the hard way.

sohankunkerkar: Yeah, Agreed it doesn't change the mechanism, I'm not asking for anything there. The only shift is who picks the number. Today admission stays blocked until the admin's timeout evicts the workload holding it up, so the admin sets that limit. After this a tenant sets it from an annotation, up to 2h.

mimowo: Ahhh, gotcha! So this is the risk of using the new per-workload feature with unbounded timeout.

A user who sets very high timeout (per-workload) may now block all other users indefinetly (if not for the bound).

So using 2h as hardcoded bound is good. We will need to also introduce the MaxTimeoutOnWorkload then.

sohankunkerkar: >So using 2h as hardcoded bound is good. We will need to also introduce the MaxTimeoutOnWorkload then.

On `MaxTimeoutOnWorkload`, is the 2h the hardcoded value, or the default of the new field? Either's fine, I just want @MaysaMacedo building the right one before the freeze.

mimowo: Since this is ConfigAPI I'm ok with "MaxTimeoutOnWorkload" in Alpha1, are you ok with that @tenzen-y ?

keps/4803-workload-level-wait-for-pods-ready/README.md
```diff
@@ -0,0 +1,297 @@
...
+  that is propagated to the corresponding `WorkloadSpec` field at Workload creation time.
+  For Deployment, the annotation needs to be defined at `spec.template.metadata.annotations`
+  to be propagated to the Pods and then workloads.
+- Per-workload timeouts take precedence over the global `WaitForPodsReady` timeout configuration when
+  both are present.
+- Introduce `MaxTimeoutOnWorkload` field in the configuration.
+
+### Non-Goals
+
+- Change the cluster-wide `WaitForPodsReady` behaviour.
```
tenzen-y:
```suggestion
- Change the cluster-wide `WaitForPodsReady` behavior.
```

nit: typo

keps/4803-workload-level-wait-for-pods-ready/README.md
```diff
@@ -0,0 +1,297 @@
...
+type WaitForPodsReady struct {
+  ...
+	// MaxTimeoutOnWorkload defines the upper bound allowed for a per-workload
+	// PodsReady timeout override (set via the `kueue.x-k8s.io/wait-for-pods-ready-timeout-seconds`
+	// annotation). If a workload requests a timeout greater than
+	// MaxTimeoutOnWorkload, the job is rejected by the admission webhook.
+	// When unset, the default maximum of 2 hours is enforced.
+	// It has no effect on workloads that don't set a per-workload override.
+	// +optional
+	MaxTimeoutOnWorkload *metav1.Duration `json:"maxTimeoutOnWorkload,omitempty"`
```
tenzen-y: I'm ok w/ introducing this in the alpha stage.

Just note: I confirmed metav1.Duration also uses int64. Hence, there is no precision issue between the annotation and the Kueue Configuration param.

keps/4803-workload-level-wait-for-pods-ready/README.md
```diff
@@ -0,0 +1,297 @@
...
+#### Workload
+
+`admittedNotReadyWorkload` in `pkg/controller/core/workload_controller.go` is
+updated to resolve the effective timeout exclusively from `wl.Spec.WaitForPodsReady.TimeoutSeconds`,
+falling back to the cluster-wide config only when that field is nil.
+The eviction deadline is resolved exclusively from `wl.Spec.WaitForPodsReady.TimeoutSeconds`
+and never from the live annotation. `prepareWorkload` does re-read the annotation
+during reconciliation to keep the stored field in sync, but any attempt to
+update `TimeoutSeconds` while the Job is unsuspended is rejected by the
+managed-resource webhook before the change can reach the API server.
```
tenzen-y: Please avoid mentioning impl-level one like variable name and function name, and whatever.

keps/4803-workload-level-wait-for-pods-ready/README.md
tenzen-y: I'm ok with that.
https://github.com/kubernetes-sigs/kueue/pull/12776#discussion_r3972887882

keps/4803-workload-level-wait-for-pods-ready/README.md
MaysaMacedo: @mimowo @tenzen-y do we want to allow the annotation values for the timeouts to be of type Duration or int?
--- EDIT
Ended up proposing int64.

keps/4803-workload-level-wait-for-pods-ready/README.md
MaysaMacedo: Addressed in https://github.com/kubernetes-sigs/kueue/pull/12776/changes/2b97fff93c744383f8f1954de8e934fadbea5da8

keps/4803-workload-level-wait-for-pods-ready/README.md
MaysaMacedo: @sohankunkerkar I added it a Non-goal for alpha and to revisit in Beta.

keps/4803-workload-level-wait-for-pods-ready/README.md
MaysaMacedo: I decided to cover the tests with e2e+unit tests only, given we can set really low values. I believe it's feasible to try with that.

MaysaMacedo:
The reviews were addressed in the last commit https://github.com/kubernetes-sigs/kueue/pull/12776/changes/7de0c7b1b5dbb56520e673a4894b3a578f0c9e82 and it's ready for another look.
Let me know in case something needs changing or was missed. Thank you.

mimowo (commented):
/lgtm
/approve

Thank you for working on this important improvement to WaitForPodsReady. Global timeout was a known limitation since the day one of the feature. Finally we introduce some flexibility here 👍 

/hold
For @tenzen-y

tenzen-y (commented):
LGTM, thank you!
/lgtm
/approve

Issue #11132, opened 2026-05-11
MultiKueue metrics needed


ns-sundar:
**What would you like to be added**:
Multikueue would need a set of new metrics at various levels: cluster connectivity and status, dispatch performance and errors, evictions/preemptions across clusters, etc. (There are probably more, but these seem to be the most obvious ones.)

  * Cluster connectivity and status:
       * Cluster status: Reachable/active, disconnected, retrying. 
       * Last Reachable time by cluster: when was the cluster last reachable?
       * Sync_errors by cluster by reason: Kube API server returned 4xx/5xx, API server not reachable,  etc.
  * Dispatch performance:
       * Number of workloads dispatched to a cluster as candidate
       * Number of workloads admitted by the cluster? (This may be available in the worker cluster, but not sure if the mgmt cluster can provide more info.)
       * Admission latency by cluster (worker clusters' admission latency metric is not the same)
       * Number of dispatch rounds before admission
       * Number of clusters tried per admission 
       * Number of workloads pending that are eligible for a given cluster
  * Dispatch errors:
       * Number of dispatch rejections, faiilure to submit, etc. by cluster and by reason
  * Cross-cluster evictions:
       * Number of evictions for the same workload across clusters 

**Why is this needed**:
For MultiKueue to move from beta to production readiness, observability is important. Cross-cluster metrics will provide a global visibility. 

**Completion requirements**:

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] API change
- [ x] Docs update

The artifacts should be linked in subsequent comments.

Mostafahassen1:
Hi @mimowo, is it okay to start working on this feature? I finished the last feature, and it’s now open for review.

mimowo:
Hi @Mostafahassen1 , sure. We will need some   more detailed proposal for the metrics. 

I think it is a good idea to also give the proposal for discussion during wg-batch meeting

Mostafahassen1:
> We will need some more detailed proposal for the metrics.
> I think it is a good idea to also give the proposal for discussion during wg-batch meeting

Hi @mimowo  ,Sorry for that, this is a new step for me. Do you want me to prepare something like documentation that explains the metric names and why we need these metrics?

mimowo:
Yes , a short doc of post which specifies the type of the metric, the labels and how they will be collected

Mostafahassen1:
[...]

PR #1132, opened 2023-09-18
[test/cache] Don't capture testing.T in operation callbacks.

trasc:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:
 Don't capture testing.T in operation callbacks.
#### Which issue(s) this PR fixes:

Fixes #1112 

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #12782, opened 2026-07-04
Add multikueue_workloads_dispatched_total metric

Mostafahassen1:
#### What type of PR is this?

/kind feature
/area multikueue

Adds a new metric `kueue_multikueue_workloads_dispatched_total` (Counter, labeled by `cluster`) counting remote workloads successfully created by the MultiKueue manager on each worker cluster, giving operators visibility into dispatch distribution.

Fixes part 1 of #11132

```release-note
MultiKueue Observability: Added a new metric `multikueue_workloads_dispatched_total` to count remote workloads successfully created by the MultiKueue manager per worker cluster.
```

## Summary by CodeRabbit

* **New Features**
  * Added a new metric for tracking workloads dispatched to worker clusters.
  * Expanded metrics documentation to include the new counter and labels.

* **Bug Fixes**
  * Improved metric naming so subsystem-based metrics are documented and exported consistently.

* **Tests**
  * Added coverage for reporting dispatched workloads across different roles and cluster assignments.


pkg/metrics/metrics.go
```diff
@@ -927,6 +939,10 @@ func AdmissionAttempt(result AdmissionResult, duration time.Duration, tracker *r
 	admissionAttemptDuration.WithLabelValues(string(result), role).Observe(duration.Seconds())
 }
 
+func MultiKueueWorkloadDispatched(cluster string, tracker *roletracker.RoleTracker) {
```
olekzabl: Nit: most such helpers in this file have names starting with an explicit verb
(usually `Report`, sometimes `Increment`, or `Record`).

I know there are exceptions but I'd stick to the dominant style.

pkg/metrics/metrics.go
```diff
@@ -350,6 +354,14 @@ The label 'result' can have the following values:
 		}, []string{"result", "replica_role"},
 	)
 
+	MultiKueueWorkloadsDispatchedTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: constants.KueueName,
+			Name:      "multikueue_workloads_dispatched_total",
+			Help:      `The total number of remote workloads created by the MultiKueue manager on a worker cluster, per 'cluster'.`,
+		}, []string{"cluster", "replica_role"},
```
mimowo: Would it be feasible to break it down by ClusterQueue? If so then it would be more consistent with other metrics we have.

pkg/metrics/metrics.go
```diff
@@ -350,6 +354,14 @@ The label 'result' can have the following values:
 		}, []string{"result", "replica_role"},
 	)
 
+	MultiKueueWorkloadsDispatchedTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: constants.KueueName,
+			Name:      "multikueue_workloads_dispatched_total",
```
mimowo: This is not necessarily an issue, but I'm wondering how the final name of the metric will look like here, is it?

kueue_multikueue_workloads_dispatched_total?

Could you check if maybe we could achieve `multikueue_workloads_dispatched_total` by introducing `constants.MultiKueueName,` as the subsystem name?

pkg/controller/admissionchecks/multikueue/workload.go
```diff
@@ -872,6 +875,8 @@ func (w *wlReconciler) nominateAndSynchronizeWorkers(ctx context.Context, group
 				if err := group.remoteClients[rem].getClient().Create(ctx, clone); err != nil {
 					log.V(2).Error(err, "creating remote object", "remote", rem)
 					errs = append(errs, err)
+				} else {
```
olekzabl, line 887: You needed to do this in 2 places (which is correct) - but this BTW raises the question if these 2 places could be unified.

I think this would make sense. We could have `syncToClusters` covering creation & deletion of remotes (i.e. `syncToSingleCluster` & lines 871-890 in your current code).
The only caveat would be the 3rd argument passed to `cloneForCreate`, which would need to become another argument of `syncToClusters`.

Would you like to make such de-duping? Ideally in a separate PR (as I recall @mimowo doesn't like mixing cleanups with logic). Can be before or after this one; I don't care in this case.

olekzabl:
Nit on the description:

> Fixes #11132

If you say so, IIUC #1132 is going to be closed once this is merged. So I wouldn't say so.

AI just told me :) that if you rephrase this to

> Fixes part 1 of #11132

it will be enough (as it claims the rules is about some keywords like "fixes" directly neighboring PR ID, ignoring punctuation).

Can you verify & adjust?

olekzabl (commented):
Thank you!

pkg/controller/admissionchecks/multikueue/workload.go
Mostafahassen1: @olekzabl After this PR, I will open a new PR to improve this.

pkg/metrics/metrics.go
Singularity23x0: Note: in multikueue context it would be good to specifically name the type of the cluster queue (since we have manager and worker CQs).
Assuming @mimowo meant the cluster queue on the manager: I would suggest a "manager_cluster_queue" metric label.

mimowo: Yeah, I assumed the `cluster_queue` for the manager cluster. I'm not sure we really need the CQ for the worker, because the pair (CQ on manager, worker) uniquely identifies the CQ on the worker. 

I'm leaning to say we can clarify the ambiguity by the comment and keep the label name short.

olekzabl: > _the pair (CQ on manager, worker) uniquely identifies the CQ on the worker_

That is, in general, not true :)
The pair (LQ on manager, worker) suffices. But there may be different LQs pointing to a single CQ.

However, I've no strong opinion on the matter (i.e. whether it's worth adding `manager_` prefix).

BTW we also have `replica_role`, which could be analogously considered ambiguous.
I had wondered earlier if we shouldn't make this `manager_replica_role` - though I finally gave up given how many metrics have just `replica_role`, and how having `manager_replica_role` would stick out of this trend.

But a comment wouldn't hurt.

How about adjusting the wording of the `// +metricsdoc:labels` explanation? For both `leader_role` and `cluster_queue`?

mimowo: > That is, in general, not true :)
The pair (LQ on manager, worker) suffices. But there may be different LQs pointing to a single CQ.

Good point, thanks for clarifying. Still, I imagine the metric could evolve to also have the "local_queue" as the label rather than introducing "worker_cluster_queue" as the label. This way we would make it non-ambiguous and more intuitive (IMO).

> BTW we also have replica_role, which could be analogously considered ambiguous.
I had wondered earlier if we shouldn't make this manager_replica_role - though I finally gave up given how many metrics have just replica_role, and how having manager_replica_role would stick out of this trend.

Yes, the manager as default is so natural, that I don' t see the need for prefixing with "manager_". I would rather just prefix with "worker_" the labels which depend on the worker.

Mostafahassen1:
I  finished updating code 
 @mimowo @olekzabl

hack/tools/metricsdoc/main.go
```diff
@@ -238,17 +238,40 @@ func metricFromCall(expr ast.Expr) (Metric, bool) {
...
+		constName = v.Name
+	default:
+		if s := stringLiteral(e); s != "" {
+			return s
+		}
+	}
+	if constName == "MultiKueueName" {
+		return "multikueue"
+	}
+	return "kueue"
```
olekzabl, line 272: Nit: this will need changes whenever yet another subsystem name is added.
We could work around this by dynamically looking up `constName` value in `pkg/constants/constants.go`.

Though I'm not sure this is worth blocking this PR. Please follow your judgement.

olekzabl (commented):
/lgtm

mimowo:
Adjusting the release note to align with the recent changes
/release-note-edit
```release-note
Added a new metric `multikueue_workloads_dispatched_total` to count remote workloads successfully created by the MultiKueue manager per worker cluster.
```

hack/tools/metricsdoc/main.go
mimowo: Good point, but I think we could just rename the var `constName` -> `subsystemName` and just return it, right? I'm just not sure about empty values, but those we could handle by fallback: `if len(subsystemName) > 0 { return subsystemName } return "kueue"`

Let's try to get it cleanup before merge

mimowo:
/approve
/hold
For the nit: https://github.com/kubernetes-sigs/kueue/pull/12782#discussion_r3552333465

hack/tools/metricsdoc/main.go
Mostafahassen1: I prototyped the dynamic lookup approach and it turned out to be more involved than a small nit fix.

Currently, there are only two subsystem prefixes (kueue and multikueue), and adding a future one is a one-line change in a single, obvious place. 

Therefore, I'd propose keeping the current explicit mapping for this PR — it's simpler to read and review, and the maintenance cost is trivial (one line, only when a subsystem is actually added). 

If more subsystems accumulate later, we can revisit the dynamic lookup as a focused follow-up.

What's your opinion, @mimowo?

mimowo: what about this option: `if len(subsystemName) > 0 { return subsystemName } return "kueue"`?

mimowo:
```golang
func subsystemValue(e ast.Expr) string {
  subsystemName := readSubsystemValue(e)
  if len(subsystemName) > 0 { 
    return subsystemName 
   } 
   return "kueue"
}

func readSubsystemValue(e ast.Expr) string {
	switch v := e.(type) {
	case *ast.SelectorExpr:
		return v.Sel.Name
	case *ast.Ident:
		return v.Name
	default:
		return stringLiteral(e)
	}
}
```
wdyt?

olekzabl: @mimowo AFAIU this code does parsing of the `metrics.go` code line looking like this:

```golang
			Subsystem: constants.MultiKueueName,
```

So, with your proposal, `readSubsystemValue` would return `"MultiKueueName"`.
And this is not what we want to use as `subsystem` value.

mimowo: ah, thanks for clarifying, so I don't have any better approach

mimowo: I'm ok to keep it hardcoded for now, unless you have some snippet @olekzabl which can be used easily

mimowo:
/unhold
Let's merge, we can always follow up when someone comes up with a better approach to this issue: https://github.com/kubernetes-sigs/kueue/pull/12782#discussion_r3552333465.

mimowo:
/release-note-edit
```release-note
MultiKueue Observability: Added a new metric `multikueue_workloads_dispatched_total` to count remote workloads successfully created by the MultiKueue manager per worker cluster.
```

PR #7780, opened 2025-11-20
[AFS] Implement cache for AFS consumed resources

IrvingMg:
#### What type of PR is this?
Implements cache-based storage for AFS consumed resources, fixing race condition where consumed resources aren't visible to scheduler when entry penalties are deleted.

/kind bug

#### Which issue(s) this PR fixes:

Fixes #6710
Fixes #7688
Fixes #7693

#### Does this PR introduce a user-facing change?

```release-note
AdmissionFairSharing: Fix the bug that occasionally a workload may get admitted from a busy LocalQueue,
bypassing the entry penalties.
```

Issue #12783, opened 2026-07-05
Flaky: Scheduler when Using AdmissionFairSharing at Cohort level should preempt a workload from LQ with higher recent usage


mbobrovskyi:
**Which test is flaking?**:
Scheduler Fair Sharing Suite: [It] Scheduler when Using AdmissionFairSharing at Cohort level should preempt a workload from LQ with higher recent usage [feature:fairsharing, feature:admissionfairsharing] 

**Link to failed CI job or steps to reproduce locally**:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/periodic-kueue-test-integration-shard-1-release-0-18/2073434117982130176

**Failure message or logs**:
```
{ failed [FAILED] Timed out after 10.000s.
Unexpected workloads are preempted
metadata:
  creationTimestamp: "2026-07-04T15:58:08Z"
  generateName: workload-
  generation: 1
  managedFields:
  - apiVersion: kueue.x-k8s.io/v1beta2
    fieldsType: FieldsV1
    fieldsV1:
      f:status:
        f:admission:
          f:clusterQueue: {}
          f:podSetAssignments:
            k:{"name":"main"}:
              .: {}
              f:count: {}
              f:flavors:
                f:cpu: {}
              f:name: {}
              f:resourceUsage:
                f:cpu: {}
        f:conditions:
          k:{"type":"Admitted"}:
            .: {}
            f:lastTransitionTime: {}
            f:message: {}
            f:observedGeneration: {}
            f:reason: {}
            f:status: {}
            f:type: {}
          k:{"type":"QuotaReserved"}:
            .: {}
            f:lastTransitionTime: {}
            f:message: {}
            f:observedGeneration: {}
            f:reason: {}
            f:status: {}
            f:type: {}
    manager: kueue-admission
    operation: Apply
    subresource: status
    time: "2026-07-04T15:58:08Z"
  - apiVersion: kueue.x-k8s.io/v1beta2
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:generateName: {}
      f:spec:
        .: {}
        f:active: {}
        f:podSets:
          .: {}
          k:{"name":"main"}:
            .: {}
            f:count: {}
            f:name: {}
            f:template:
              .: {}
              f:metadata: {}
              f:spec:
                .: {}
                f:containers:
                  .: {}
                  k:{"name":"c"}:
                    .: {}
                    f:name: {}
                    f:resources:
                      .: {}
[...]

PR #12786, opened 2026-07-05
Fix AFS entry penalty leak for workloads admitted via AdmissionChecks

### LINKED ISSUE #12539: Fair Sharing Penalties Leak Permanently for Workloads with Admission Checks in pkg/controller/core/workload_controller.go

### Impact
Legitimate tenants who use AdmissionChecks (e.g., for TopologyAwareScheduling or external provisioning) will permanently leak their fair sharing penalties. Over time, their `LocalQueue`'s `FSUsage` will artificially inflate, pushing them to the absolute bottom of the scheduler's priority queue. This results in a persistent Denial of Service (DoS) for those tenants, as their future workloads will be permanently starved by the scheduler.

### Data Flow
- **pkg/scheduler/scheduler.go:812**: Scheduler assumes the workload and calls `s.updateEntryPenalty(log, e, add)`. The workload transitions to `QuotaReserved`.
- **pkg/controller/core/workload_controller.go:1174**: Informer processes `Pending` -> `QuotaReserved`. `status == workload.StatusAdmitted` is false, so `updateAfsConsumedUsage` is SKIPPED.
- **pkg/controller/core/workload_controller.go:1219**: Later, the workload passes AdmissionChecks and transitions `QuotaReserved` -> `Admitted`. This falls to the `default:` case in the switch statement.
- **pkg/controller/core/workload_controller.go:1222**: `updateAfsConsumedUsage` is never called. The initial penalty added by the scheduler is never subtracted via `AfsEntryPenalties.Sub()`.

### Reproduction
**Preconditions**: AdmissionFairSharing must be enabled. A Workload must be created that requires an AdmissionCheck (e.g., ProvisioningRequest, MultiKueue).
**Steps**:
1. Create a Workload that is bound by an AdmissionCheck.
2. The scheduler processes the Workload, adds the penalty to `AfsEntryPenalties`, and sets its status to `QuotaReserved`.
3. Fulfill the AdmissionCheck so the Workload transitions to `Admitted`.
4. Inspect the Kueue cache (or logs); observe that `AfsEntryPenalties.Sub()` is never called.
5. Delete or finish the Workload.
6. The penalty permanently remains on the LocalQueue until the LocalQueue itself is destroyed.

### Vulnerable Code
```
if afs.Enabled(r.admissionFSConfig) && status == workload.StatusAdmitted && r.cache.ClusterQueueUsesAdmissionFairSharing(wlCopy.Status.Admission.ClusterQueue) {
	r.updateAfsConsumedUsage(log, wlCopy)
}
```

### Remediation
Ensure `updateAfsConsumedUsage` is called when a Workload transitions to `Admitted`, regardless of whether its previous state was `Pending` or `QuotaReserved`.

Suggested Fix:
```
Add a specific switch case for `prevStatus == workload.StatusQuotaReserved && status == workload.StatusAdmitted` that calls `r.updateAfsConsumedUsage(log, wlCopy)` before falling through to the cache update logic.
```

### Environment
* Install tools: N/A
* Others: N/A

apullo777:
#### What type of PR is this?

/kind bug

With AdmissionFairSharing enabled, workloads receive an entry penalty when they are assumed. That penalty was only settled on the `Pending → Admitted` path.

Workloads gated by AdmissionChecks, such as ProvisioningRequest, MultiKueue, or TAS, are admitted through `Pending → QuotaReserved → Admitted`. The final transition did not settle the entry penalty, so the LocalQueue's fair-sharing usage stayed artificially inflated and its later workloads were deprioritized.

This seems to be a regression introduced by the AFS consumed-resources cache refactor (#7780). Before that refactor, settlement happened during the LocalQueue reconciler's recomputation and was transition-independent. The refactor moved settlement into the workload status-transition switch, where it was only wired to the `Pending → Admitted` case.

This PR fixes the leak by settling the entry penalty whenever a workload transitions into `Admitted`. The settle call is centralized after the status-transition switch so it also covers transitions handled by earlier cases, instead of duplicating the call in a dedicated `QuotaReserved → Admitted` case.

It also clamps `AfsEntryPenalties.Sub` at zero, so settling without a matching in-memory penalty, for example after restart, cannot create a permanent usage discount.

Also added regression coverage for the AdmissionCheck admission path. AI assistance was used for parts of this change.

Rollback paths such as `QuotaReserved → Pending` eviction or deletion while QuotaReserved are left for follow-up  #12787, since they have different lifecycle semantics.

Fixes #12539

- On the clamp: the settle amount is recomputed at settlement time rather than recorded at push time, so it can exceed what was pushed. For example, a manager restart can clear the in-memory penalties while the workload keeps its persisted QuotaReserved state. Clamping at zero prevents that mismatch from becoming a permanent usage discount.

- Test isolation: without the settlement fix, the new `QuotaReserved → Admitted` unit case fails and the new integration test times out. Without the clamp, the new over-subtraction `Sub` unit cases fail.

```release-note
AdmissionFairSharing: Fixed a bug where workloads admitted via AdmissionChecks could keep their entry penalty permanently, inflating LocalQueue fair-sharing usage and deprioritizing later workloads.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Admission fair-sharing penalties no longer become negative and are cleared correctly when fully offset.
  * Penalties are settled when workloads become admitted, including after quota reservation, while avoiding inactive or unchanged workloads.

* **Tests**
  * Added coverage for subtraction edge cases and admission-state transitions.
  * Added integration coverage confirming penalties clear after admission checks become ready.

### DIFF
--- a/pkg/cache/queue/afs/entry_penalties.go
+++ b/pkg/cache/queue/afs/entry_penalties.go
@@ -47,6 +47,16 @@ func (m *AfsEntryPenalties) Sub(lqKey utilqueue.LocalQueueReference, penalty cor
 	}
 	m.penalties.UpdateOrDelete(lqKey, func(existing corev1.ResourceList) (corev1.ResourceList, bool) {
 		merged := resource.MergeResourceListKeepSum(existing, negated)
+		// Sub recomputes the amount at settlement time rather than recording it at
+		// push time, so it can exceed what was pushed (e.g. after a restart loses the
+		// in-memory penalties). Clamp at zero: a mismatch must not become a permanent
+		// usage discount for the LocalQueue.
+		for k, v := range merged {
+			if v.Sign() < 0 {
+				v.Set(0)
+				merged[k] = v
+			}
+		}
 		return merged, canClearPenalty(merged)
 	})
 }
--- a/pkg/cache/queue/afs/entry_penalties_test.go
+++ b/pkg/cache/queue/afs/entry_penalties_test.go
@@ -1,0 +1,93 @@
+// Copyright The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package afs
+
+import (
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+
+	utilqueue "sigs.k8s.io/kueue/pkg/util/queue"
+)
+
+func TestSub(t *testing.T) {
+	lqKey := utilqueue.NewLocalQueueReference("ns", "lq")
+
+	cases := map[string]struct {
+		push        corev1.ResourceList
+		sub         corev1.ResourceList
+		wantPending bool
+		wantPenalty corev1.ResourceList
+	}{
+		"subtracting the pushed amount clears the entry": {
+			push:        corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("4")},
+			sub:         corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("4")},
+			wantPending: false,
+			wantPenalty: corev1.ResourceList{},
+		},
+		"partial subtraction keeps the remainder": {
+			push:        corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("4")},
+			sub:         corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("1")},
+			wantPending: true,
+			wantPenalty: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("3")},
+		},
+		"subtracting more than pushed clamps at zero and clears the entry": {
+			push:        corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("2")},
+			sub:         corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("4")},
+			wantPending: false,
+			wantPenalty: corev1.ResourceList{},
+		},
+		"subtracting without a prior push does not leave a negative entry": {
+			sub:         corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("4")},
+			wantPending: false,
+			wantPenalty: corev1.ResourceList{},
+		},
+		"clamping applies per resource": {
+			push: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("4")},
+			sub: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("1"),
+				corev1.ResourceMemory: resource.MustParse("1Gi"),
+			},
+			wantPending: true,
+			wantPenalty: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("3"),
+				corev1.ResourceMemory: resource.MustParse("0"),
+			},
+		},
+	}
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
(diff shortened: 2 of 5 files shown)


pkg/cache/queue/afs/entry_penalties.go
```diff
@@ -41,7 +41,18 @@ func (m *AfsEntryPenalties) Sub(lqKey utilqueue.LocalQueueReference, penalty cor
...
+	// push time, so it can exceed what was pushed (e.g. after a restart loses the
+	// in-memory penalties). Clamp at zero: a mismatch must not become a permanent
+	// usage discount for the LocalQueue.
+	for k, v := range newPenalty {
+		if v.Sign() < 0 {
+			v.Set(0)
+			newPenalty[k] = v
+		}
+	}
+	m.penalties.Add(lqKey, newPenalty)
```
mimowo: This does not seem directly related to the referenced issue. IIUC this is to make sure the usage never drops below 0 in case of concurrent modifications, right?

mimowo (commented):
cc @sohankunkerkar

pkg/cache/queue/afs/entry_penalties.go
apullo777: Thanks for the question! You are right that this part is not the main fix for the issue.

This is not about concurrent map modifications. It is a defensive guard for the settlement path enabled by this PR. `Sub` subtracts an amount recomputed at settlement time, rather than replaying the exact amount that was pushed at reservation time.

With this PR, the `QuotaReserved -> Admitted` path now calls settlement. If the in-memory entry penalty is missing or smaller than the recomputed amount, for example after a manager restart while the Workload was waiting in `QuotaReserved`, the old code could subtract from an empty/smaller entry and leave a negative pending penalty. That would act as a usage discount in AFS ordering.

The clamp makes `Sub` saturate at zero instead. This is covered by the new unit test case `subtracting more than pushed clamps at zero and clears the entry`.

I can also make the code comment more explicit about the restart/mismatch case, or split this out if you prefer keeping this PR limited to the transition fix.

mimowo: sgtm, but let me also align on the direction as I cover this partly in my question here: https://github.com/kubernetes-sigs/kueue/issues/12783#issuecomment-4891164731

kshalot:
Hey @apullo777, I'm doing a quick scan of my GH inbox.

AFAIR we agreed to go with the QR-anchored approach here, right? https://github.com/kubernetes-sigs/kueue/issues/12783#issuecomment-4916930808

I'm not rushing you, I just wanted to make sure you are blocked on a response from somebody etc :)

apullo777:
> AFAIR we agreed to go with the QR-anchored approach here, right? [#12783 (comment)](https://github.com/kubernetes-sigs/kueue/issues/12783#issuecomment-4916930808)
> I just wanted to make sure you are blocked on a response from somebody etc :)

@kshalot, thanks for checking in :)

Yes. QR-anchored accounting is my understanding of the agreed direction. I’m not blocked on review feedback so much as on the implementation boundary and sequencing.

Since this changes the accounting semantics, should we first update KEP-4136 to define QuotaReserved as the accounting start and specify the consolidation/exit-path invariant, or would you prefer a PoC implementation first with the KEP updated alongside it? Also, should #12786 be repurposed for that first implementation slice, or kept as a separate stopgap?

@mimowo, happy to help with either the KEP update or the implementation once we agree on the shape.

Also, given that the discussion is now spread across multiple issues, perhaps it would be great to open a focused umbrella issue to track the KEP-4136 update and the QR-anchored implementation?

mimowo:
I think we have two issues;
1. Entry penalties should be accounted at QR time
2. Entry penalties should not leak

If you think (1.) and (2.) can be solved independently - go for it. I just proposed doing (1.) in case solving (2.) currently requires hacks which would not be needed if we do (1.).

Overall, I think the KEPfor (1.)  is just for track keeping. 

What would be helpful to determine if we couple (1.) and (2.) is a prototype for fixing both as we would like it to be. Then we could be more informed to decide if we want to split the fixes or not.

mimowo:
@apullo777 please rebase

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.19
/cherrypick release-0.18
Thank you for the nice AFS leak fix 👍

mimowo:
@apullo777 please prep the CP for 0.18: https://github.com/kubernetes-sigs/kueue/pull/13794

Issue #3125, opened 2024-09-24
Introduce Job-agnostic API to declare the maximal execution time for a Job


mimowo:
**What would you like to be added**:

A job-agnostic API to set the maximal execution time for a Job. 

There are some open questions:
- name of the API (I would initially propose annotation ` kueue.x-k8s.io/max-exec-time-seconds`)
- semantics of the deadline (measured since last admission or cumulative execution time) - I would initially suggest to measure the cumulative execution time across admissions , and re-use it in the EXEC_TIME command for kueuectl, but it will require API changes
- what happens after exceeding the time  - I would initially propose deactivating the workload, but alternatives might be worth considering

**Why is this needed**:

Different Jobs CRDs have a field with similar semantics, but there is no standard. For example batch/Job has `spec.activeDeadlneSeconds`, while JobSet does not have such an API for now.

We would like to have such an API to use it in kueuectl command as an analog of slurm's "-t" option. Long term (but out-of-scope here) we could use this value to optimize Job scheduling.

**Completion requirements**:

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

kannon92:
Maybe kueue could set activeDeadlineSeconds on all the pod templates?

tardieu:
Very much in favor.

Specifying a deadline for a workload (possibly wall clock time instead of cumulative execution time) would also enable capacity planning and permit developing ordering strategies to, e.g., ensure large workloads have a chance to run without resorting to priorities and preemption.

An alternative to workload deactivation could be to dynamically lower the priority of the workload, hence let it run if there is excess capacity and only evict if necessary.

mimowo:
> (possibly wall clock time instead of cumulative execution time)

Yeah, I was thinking that "cumulative execution time" measures wall time. I say it is  cumulative because it would add up running wall time from all admission times (say if it is admitted (3min) -> suspended (whatever) -> admitted (3min), this would account to 6min, vs activeDeadlineSeconds which accounts only 3min (last admission time).

wdyt?

> would also enable capacity planning and permit developing ordering strategies to, e.g., ensure large workloads have a chance to run without resorting to priorities and preemption.

Yeah, that is the long-term plan. IIUC its counterpart ("-t" in slurm) is used for better scheduling.

[...]

PR #12791, opened 2026-07-05
KEP-9876: Preemption protection — guaranteed minimum runtime for cross-CQ preemption

michael-pryor:
#### What type of PR is this?
/kind kep

Revised version of the KEP-9876 proposal (cross-ClusterQueue guaranteed minimum runtime before preemption), restructured to address all review feedback on #9877.

Part-of #9876
Revision of #9877

```release-note
NONE
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - KEP-9876 documentation introduces a planned alpha **preemption protection** mechanism to enforce a configurable minimum runtime for admitted workloads before they can be targeted for cross-ClusterQueue preemption.
  - Separate protection controls are defined for **fair-sharing rebalancing** and **cross-ClusterQueue reclaim** candidates.
  - The KEP specifies scheduling and simulation behavior, observability expectations (including `V(4)` logs), validation rules, and rollout/testing criteria.

### DIFF
--- a/keps/9876-cross-cq-guaranteed-runtime/README.md
+++ b/keps/9876-cross-cq-guaranteed-runtime/README.md
@@ -79,8 +79,11 @@ for reclaim (where the owner has a legitimate entitlement).
 
 ### Non-Goals
 
-- Within-ClusterQueue time-based preemption (addressed by KEP-8522; see
-  [Alternatives](#rely-on-within-clusterqueue-time-based-preemption-alone))
+- Within-ClusterQueue time-based preemption, currently addressed
+  experimentally by the
+  [priority booster](../../cmd/experimental/kueue-priority-booster/README.md).
+  See
+  [Alternatives](#rely-on-within-clusterqueue-time-based-preemption-alone).
 - Per-ClusterQueue overrides for either duration (see
   [Future Extensibility](#future-extensibility))
 - Per-workload minimum runtime overrides
@@ -151,12 +154,14 @@ Workloads that don't implement checkpointing lose progress when preempted.
 This is not new; workloads already need to handle preemption gracefully.
 Protection reduces how often that happens but does not remove the need.
 
-This feature is complementary to KEP-8522 (time-based same-priority
-preemption within a ClusterQueue). KEP-8522 addresses time-sharing among
-workloads within the same ClusterQueue, while this feature addresses minimum
-runtime guarantees across ClusterQueues. The two mechanisms operate
-independently and do not conflict; the protection rules are grouped so that
-a within-ClusterQueue rule can slot in later (see
+This feature is complementary to within-ClusterQueue time-based preemption,
+currently addressed experimentally by the
+[priority booster](../../cmd/experimental/kueue-priority-booster/README.md).
+The priority booster implements time-sharing within a ClusterQueue by changing
+a workload's effective priority after a configured interval, while this
+feature filters cross-ClusterQueue preemption candidates. The two mechanisms
+operate independently and do not conflict; the protection rules are grouped
+so that a native within-ClusterQueue rule can slot in later (see
 [Future Extensibility](#future-extensibility)).
 
 It is also complementary to the `SchedulerTimestampPreemptionBuffer` feature
@@ -166,13 +171,45 @@ Preemption protection provides an administrator-configurable, cross-CQ
 guarantee measured from admission time rather than queue timestamps.
 
 **Interaction with Fair Sharing strategies**: protected cross-CQ candidates
-are excluded during target classification, before the configured
-`preemptionStrategies` are applied to them; `InClusterQueue` candidates are
-never affected.
+are excluded only after the existing resource, policy, and configured
+`preemptionStrategies` checks pass and the candidate's preemption reason is
+known. This ensures that expiry retries are recorded only for candidates that
+would otherwise be eligible; `InClusterQueue` candidates are never affected.
 
 **Re-admission**: preempted workloads follow the existing eviction and
 re-admission path. On re-admission the workload's admission time is set
 afresh, restarting the protection window.
+
+**Enabling protection for already-admitted workloads**: enabling the feature
+gate and configuration applies to workloads that are already admitted. Their
+persisted `Admitted.LastTransitionTime` is used; enabling protection does not
+start a fresh window. For example, a workload admitted 5 minutes before a
+10-minute rule takes effect is protected for approximately 5 more minutes.
+
+**Maximum execution time**: preemption protection does not override
+[KEP-3125](../3125-maximum-execution-time/README.md) or
+`Workload.spec.maximumExecutionTimeSeconds`. The maximum-execution-time
+controller accounts for execution accumulated in previous admission cycles
+and elapsed time in the current cycle. If that limit expires before the
+protection duration, the workload is deactivated and evicted even though it
+is still protected from the configured cross-ClusterQueue preemption reason.
+
+**Elastic workloads using WorkloadSlices**: scale-down updates the existing
+Workload slice in place and preserves its `Admitted` timestamp. Scale-up of an
+admitted Workload creates a replacement Workload slice. When that replacement
+is admitted it receives a new `Admitted=True` transition time, starts a new
+protection window, and the old slice is marked `Finished`. Protection is
+therefore measured per Workload slice rather than from the parent job's first
+admission.
+
+This is intentional: each admitted scale-up represents a new execution shape
+for the complete workload, and the fresh window gives that expanded workload
+an opportunity to stabilize and make progress. Reusing the parent job's
+original timestamp could make a long-running elastic workload immediately
+eligible for preemption as soon as it scales up, allowing the newly added
+capacity to be removed before it contributes and slowing the workload's
+progress toward completion. Scale-down does not restart protection because it
+does not require admitting additional capacity.
 
 ### Risks and Mitigations
 
@@ -188,13 +225,16 @@ are within their protection window, an owner CQ must wait to get its own
 nominal quota back — and because every newly admitted borrower starts a
 fresh protection window, a sustained stream of short-lived borrower
 admissions can delay reclaim indefinitely, not just by one window. Two
-design elements bound this risk: while an owner's reclaim fails solely due
+design elements mitigate this risk: while an owner's reclaim fails solely due
 to protection, the scheduler reserves the contested capacity within the
 scheduling cycle rather than admitting new borrowers onto it (see
 [Preemption Eligibility](#preemption-eligibility) for the `CanAlwaysReclaim`
 interaction), and the pending owner is retried when the earliest protection
 window expires (see
 [Retrying After Protection Expiry](#retrying-after-protection-expiry)).
+Neither mechanism establishes a cross-cycle upper bound: sustained borrower
+churn can still delay reclaim indefinitely.
+
 Administrators should additionally keep the reclaim protection duration
 short (or unset). An alternative that exempts late-admitted borrowers
 entirely is discussed in
@@ -325,6 +365,16 @@ distinction matters:
   `QuotaReserved` in the first pass and only become `Admitted` after
   topology assignment in the second pass.
 
+`Admitted.LastTransitionTime` is the authoritative start of the current
+admission cycle. Updates that leave `Admitted=True` must preserve it; a real
+transition away from and back to `Admitted=True` starts a new cycle and gets a
+new timestamp. This timestamp already affects behavioral decisions such as
+the `waitForPodsReady` timeout and maximum execution time. This KEP also makes
+it an input to preemption victim selection. Changes to admission flows - for
+example concurrent-admission retries, WorkloadSlice replacement, or TAS
+second-pass handling - must therefore preserve or reset it deliberately and
+cover the resulting protection behavior in tests.
+
 Note that pod scheduling and startup happen after `Admitted`, so pod startup
 time counts against the protection window; administrators should size
 durations accordingly.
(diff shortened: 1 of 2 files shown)


keps/9876-cross-cq-guaranteed-runtime/README.md
```diff
@@ -0,0 +1,702 @@
...
+`reclaimWithinCohort: Any` can always reclaim its nominal quota later, which
+lets the scheduler skip reserving capacity for its pending workloads
+(`reserveCapacityForUnreclaimablePreempt`, `pkg/scheduler/scheduler.go`).
+Reclaim protection breaks that premise while candidates are protected. When
+a reclaim protection duration is configured, `CanAlwaysReclaim` will return
+`false`, so the existing capacity-reservation path prevents new borrowers
+from being admitted onto contested capacity within the same scheduling
+cycle while the owner waits out protection windows.
+
+### Interaction with the Preemption Oracle
```
michael-pryor, line 455: This section and below two sections may be too verbose, happy to remove this if its too much information

keps/9876-cross-cq-guaranteed-runtime/README.md
```diff
@@ -0,0 +1,702 @@
...
+    minAdmitDuration: 30m
+```
+
+While maximally extensible, it is over-general for two or three entries:
+validation must reject duplicate and unknown reasons, merging configurations
+becomes positional, and the schema cannot express per-reason refinements as
+naturally as named structs. The struct-based grouping keeps the same
+extensibility with simpler validation.
+
+### Per-ClusterQueue configuration
```
kannon92: The downside of configuration is its one size fits all.

We have quite a few requests to add timeouts to tune `WaitForPodsReady` because some workloads can take a lot longer and they want a longer time limit.

Could you envision users for AI/ML needing more time due to their init containers (or setup) take a lot longer?

I'm having a hard time thinking through how can admin decides what value to put for guaranteed minimum runtime?

mimowo:
cc @mukund-wayve @kshalot ptal

sohankunkerkar:
I'll take a pass at it later ~today~.

edit: maybe tomorrow :)

keps/9876-cross-cq-guaranteed-runtime/README.md
michael-pryor: > We have quite a few requests to add timeouts to tune WaitForPodsReady because some workloads can take a lot longer and they want a longer time limit.
>  
> Could you envision users for AI/ML needing more time due to their init containers (or setup) take a lot longer?

Yes - I will move per-CQ configuration into the Future Extensibility section. Most of the value would be on the borrower side: "if I borrow, I want a minimum duration of X before it can be reclaimed." But there's a lender-side dimension too; some lower-latency queues may want to reclaim faster than the global default. I do think global configuration is a reasonable first step though.

> I'm having a hard time thinking through how can admin decides what value to put for guaranteed minimum runtime?

We rolled this out across our inference clusters this week with a 30-minute minimum. The framing that worked for us was simple fairness: if the scheduler gives you a GPU and takes it away a minute later, that's not fair - there must be some baseline duration that everyone can agree is reasonable.

Any reasonable duration (10m, 30m, 1h) is an improvement over the status quo where a workload can be preempted after one second. We chose 30 minutes to reduce churn and make scheduling behaviour easier to reason about, and it immediately had a visible impact.

It's worth noting this is different from WaitForPodsReady in one important way: WaitForPodsReady needs to match the slowest possible startup for a workload type, which varies a lot - hence the per-workload pressure. minAdmitDuration is a fairness floor that applies equally regardless of workload type, so global works well. That said, a per CQ override would provide greater flexibility and I think there is some need for that as a future extension.

michael-pryor:
PR with changes that we've deployed internally: https://github.com/kubernetes-sigs/kueue/pull/12996/changes

Once the KEP is approved, I will move that into ready for review.

mimowo:
cc @mwielgus

keps/9876-cross-cq-guaranteed-runtime/kep.yaml
```diff
@@ -0,0 +1,43 @@
...
+stage: alpha
+
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. This can be the current (upcoming) milestone, if it is being actively
+# worked on.
+latest-milestone: "v0.19"
+
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  alpha: "v0.19"
```
mimowo: The KEP matches my expectations from the previous discussions, but the maintainers are too loaded I think to include it in 0.19 which is coming just next week (most likely). If you don't mind I would propose to revisit this for 0.20. 

Also, I know that @mwielgus is thinking about a more general configurable preemptions, so maybe we could cover this use case by the other KEP. 

Let me know your thoughts

mimowo:
cc @mukund-wayve would you like to review?

keps/9876-cross-cq-guaranteed-runtime/README.md
```diff
@@ -0,0 +1,757 @@
...
+quota promptly, while usage negotiated purely by fairness can afford a longer
+protection budget.
+
+### Notes/Constraints/Caveats
+
+Workloads that don't implement checkpointing lose progress when preempted.
+This is not new; workloads already need to handle preemption gracefully.
+Protection reduces how often that happens but does not remove the need.
+
+This feature is complementary to KEP-8522 (time-based same-priority
```
sohankunkerkar: I would describe the problem space ("within-CQ time-based preemption, currently addressed experimentally by the priority booster") rather than mentioning KEP-8522

keps/9876-cross-cq-guaranteed-runtime/README.md
```diff
@@ -0,0 +1,757 @@
...
+
+Eviction does not flip `Admitted` immediately: an evicted workload carries
+`Evicted=True` alongside `Admitted=True` until its quota reservation is
+released, and such workloads are deliberately the *preferred* preemption
+victims (preempting them costs nothing extra). Protection therefore applies
+only to candidates with `Admitted=True` and not `Evicted=True`. When the
+workload is later re-admitted, `Admitted` transitions back to `True` with a
+fresh timestamp, so the protection window always refers to the current
+uninterrupted run. Because the condition and its timestamp are persisted in
+the workload status, protection windows survive controller restarts with no
```
sohankunkerkar, line 396: IIUC, this is the first feature that depends on `Admitted.LastTransitionTime` for a scheduling decision. Today nothing reads that timestamp as it's just bookkeeping. Once protection ships, any future change that shifts when Admitted transitions (concurrent admission retry, elastic re-admission, TAS second-pass) will silently change protection windows. Worth calling out as a new invariant so future authors know the timestamp is load-bearing.

keps/9876-cross-cq-guaranteed-runtime/README.md
```diff
@@ -0,0 +1,757 @@
...
+protection budget.
+
+### Notes/Constraints/Caveats
+
+Workloads that don't implement checkpointing lose progress when preempted.
+This is not new; workloads already need to handle preemption gracefully.
+Protection reduces how often that happens but does not remove the need.
+
+This feature is complementary to KEP-8522 (time-based same-priority
+preemption within a ClusterQueue). KEP-8522 addresses time-sharing among
```
sohankunkerkar: When the feature gate is enabled on a running cluster, already-admitted workloads get retroactive protection based on their existing Admitted timestamp. A workload      admitted 5 minutes ago with a 10m window suddenly becomes protected for 5 more minutes. Probably fine, but worth noting so admins know enabling the gate mid-flight immediately affects the preemption candidate pool.

keps/9876-cross-cq-guaranteed-runtime/README.md
```diff
@@ -0,0 +1,757 @@
...
+runtime guarantees across ClusterQueues. The two mechanisms operate
+independently and do not conflict; the protection rules are grouped so that
+a within-ClusterQueue rule can slot in later (see
+[Future Extensibility](#future-extensibility)).
+
+It is also complementary to the `SchedulerTimestampPreemptionBuffer` feature
+gate, which adds a fixed 5-minute buffer preventing near-simultaneous
+equal-priority workloads from preempting each other within a ClusterQueue.
+Preemption protection provides an administrator-configurable, cross-CQ
+guarantee measured from admission time rather than queue timestamps.
```
sohankunkerkar, line 171: If the cluster also configures a CQ max execution time ([KEP-3125](https://github.com/kubernetes-sigs/kueue/issues/3125)), a workload whose execution limit is shorter than the protection duration gets evicted by the time limit before protection matters. Not a conflict, just worth a sentence so admins don't set a 1h protection window and wonder why their 30m jobs still get evicted.

keps/9876-cross-cq-guaranteed-runtime/README.md
```diff
@@ -0,0 +1,757 @@
...
+
+**Interaction with Fair Sharing strategies**: protected cross-CQ candidates
+are excluded during target classification, before the configured
+`preemptionStrategies` are applied to them; `InClusterQueue` candidates are
+never affected.
+
+**Re-admission**: preempted workloads follow the existing eviction and
```
sohankunkerkar, line 179: What about elastic workloads? If a workload grows via WorkloadSlices after admission, the new pods become preemption candidates too. They'd inherit the original Admitted timestamp since the condition doesn't change on elastic growth. That seems correct and the workload already got its guaranteed runtime but it would help to state this explicitly so the interaction isn't left ambiguous.

keps/9876-cross-cq-guaranteed-runtime/README.md
```diff
@@ -0,0 +1,757 @@
...
+workloads that have exceeded their protection window. This is the intended
+trade-off; documentation will include tuning guidance (typically: slightly
+above the workloads' checkpoint interval).
+
+**Reclaim can be starved under borrower churn**: if all reclaim candidates
+are within their protection window, an owner CQ must wait to get its own
+nominal quota back — and because every newly admitted borrower starts a
+fresh protection window, a sustained stream of short-lived borrower
+admissions can delay reclaim indefinitely, not just by one window. Two
+design elements bound this risk: while an owner's reclaim fails solely due
```
sohankunkerkar: `bound` is too strong here. `CanAlwaysReclaim` only prevents new borrower admission within the same scheduling cycle. Between cycles, new borrowers can still be admitted and start fresh protection windows. In sustained churn (many short-lived borrowers), the owner could wait well beyond one window. I'd rephrase it to "mitigate" instead of "bound," and note sustained churn as a known limitation.

keps/9876-cross-cq-guaranteed-runtime/kep.yaml
michael-pryor: > The KEP matches my expectations from the previous discussions, but the maintainers are too loaded I think to include it in 0.19 which is coming just next week (most likely). If you don't mind I would propose to revisit this for 0.20.

Ok, I will update to 0.20

> Also, I know that @mwielgus is thinking about a more general configurable preemptions, so maybe we could cover this use case by the other KEP.

Which KEP is this? If its possible to do I'd prefer to get my feature in in isolation, but keep it compatible with the broader vision.

keps/9876-cross-cq-guaranteed-runtime/README.md
michael-pryor: ok I've made a change.

keps/9876-cross-cq-guaranteed-runtime/README.md
michael-pryor: Ok I've made a change to reflect this

keps/9876-cross-cq-guaranteed-runtime/README.md
michael-pryor: ok I have added some text about this

keps/9876-cross-cq-guaranteed-runtime/README.md
michael-pryor: ok i've changed this

keps/9876-cross-cq-guaranteed-runtime/README.md
michael-pryor: I've added a section on this. Note that its not the first time its being used; it is used for decisions like waitForPodsReady, but it is the first time its used for victim selection.

michael-pryor:
Additional corrections included in this update:

- Use “admission cycle” instead of “uninterrupted run.”
- Clarify that fair-sharing protection is checked only after normal strategy eligibility is established.
- Define clock sampling per target-selection operation, including oracle simulation.
- Make the oracle wiring explicit: it uses the same rules and time, without scheduling retries.
- Clarify that expiry rebroadcasts the root Cohort tree, covering inadmissible and heap-resident StrictFIFO workloads.

keps/9876-cross-cq-guaranteed-runtime/README.md
michael-pryor: This is an interesting point. I’ve added a section describing the current behavior, but I think the trade-off is worth discussing.

As implemented in #12996, scaling up creates a replacement WorkloadSlice representing the workload’s new total size. When that replacement is admitted, it receives a new Admitted timestamp, so the protection window restarts for the whole workload.

For example, with a 30-minute protection period:
- T0: admitted with 2 pods; protected until T30
- T29: scales to 6 pods; protected until T59
- T58: scales to 10 pods; protected until T88
- T88: the 10-pod workload becomes eligible for cross-ClusterQueue preemption

Consequently, the original 2-pod allocation remains part of a protected workload for 88 minutes, while the capacity added at T29 is protected for 59 minutes.

I think this behavior has a useful property: after granting a scale-up, Kueue gives the complete, expanded workload an opportunity to stabilize and make progress. Otherwise, newly granted capacity could be removed almost immediately, creating avoidable churn. Both scaling and preemption can incur startup costs, reliability risk, and lost work.

The trade-off is that repeated scale-ups can extend protection beyond the configured 30 minutes and delay reclaim or fair-sharing convergence. I think that may be reasonable for the initial semantics, but it should be an explicit decision.

PR #12795, opened 2026-07-05
Fix populator error swallowing

### LINKED ISSUE #12569: Unchecked Error Condition in LocalQueue Creation Leads to Persistent Queue Provisioning Failure in cmd/experimental/kueue-populator/pkg/controller/controller.go

RepoInspector found a vulnerability in https://github.com/kubernetes-sigs/kueue.

### Impact
A transient error or a mild race condition permanently disrupts the primary function of the `kueue-populator` for a given namespace. Workloads in the affected namespace will be unable to schedule because the required `LocalQueue` was never provisioned, leading to a persistent, localized Denial of Service.

### Data Flow
- **cmd/experimental/kueue-populator/pkg/controller/controller.go:127**: Reconcile calls r.ensureLocalQueueExists(ctx, cq, &ns) for each matching ClusterQueue.
- **cmd/experimental/kueue-populator/pkg/controller/controller.go:193**: Inside ensureLocalQueueExists, r.client.Create(ctx, newLQ) fails due to a transient API error, rate limit, or a TOCTOU race condition (e.g., an attacker creating a conflicting LocalQueue after the initial Get check).
- **cmd/experimental/kueue-populator/pkg/controller/controller.go:195**: ensureLocalQueueExists returns the error back to Reconcile.
- **cmd/experimental/kueue-populator/pkg/controller/controller.go:128**: Reconcile logs the error but completely ignores the error state, continuing to the next iteration.
- **cmd/experimental/kueue-populator/pkg/controller/controller.go:132**: Reconcile returns 'ctrl.Result{}, nil', successfully completing the reconciliation loop without informing the controller-runtime manager that an error occurred.

### Reproduction
**Preconditions**:
The `kueue-populator` controller is running. The attacker must have RBAC permissions to create `LocalQueue` objects in a target namespace.

**Steps**:
1. Identify a target namespace and wait for a `ClusterQueue` to be created that matches its labels.
2. Trigger a transient failure precisely when the populator attempts to create the `LocalQueue`. This can be achieved by running a loop that constantly creates and deletes a conflicting `LocalQueue`.
3. The populator's `Get` call will return `NotFound`, but its subsequent `Create` call will fail with `AlreadyExists`.
4. Observe that the populator logs the failure but returns a successful reconciliation result.
5. Terminate the race condition script and ensure the target namespace is empty.
6. Observe that the populator never retries creating the `LocalQueue`.

### Vulnerable Code
```go
if err := r.ensureLocalQueueExists(ctx, cq, &ns); err != nil {
	log.Error(err, "Failed to ensure LocalQueue exists in namespace", "clusterQueue", klog.KObj(cq), "namespace", klog.KObj(&ns))
}
```

### Remediation
Accumulate errors encountered during `ensureLocalQueueExists` and return an aggregate error at the end of the `Reconcile` function to ensure that controller-runtime requeues the request on failure.

Suggested Fix:
```go
var errs []error
for i := range cqs.Items {
	cq := &cqs.Items[i]
	// ...
	if err := r.ensureLocalQueueExists(ctx, cq, &ns); err != nil {
		log.Error(err, "Failed to ensure LocalQueue exists in namespace", "clusterQueue", klog.KObj(cq), "namespace", klog.KObj(&ns))
		errs = append(errs, err)
	}
}

if len(errs) > 0 {
	import utilerrors "k8s.io/apimachinery/pkg/util/errors"
	return ctrl.Result{}, utilerrors.NewAggregate(errs)
}

return ctrl.Result{}, nil
```

### Environment
* Install tools: N/A
* Others: N/A

NasitSony:
#### What type of PR is this?
/kind bug
/area kueue-populator

In `KueuePopulatorReconciler.Reconcile`, errors returned from
`ensureLocalQueueExists` were logged but not propagated. Since
`Reconcile` unconditionally returned `ctrl.Result{}, nil`, controller-runtime
never requeued the request on failure — so a transient API error, rate
limit, or a TOCTOU race (a conflicting `LocalQueue` created between the
`Get` check and the `Create` call) would silently and permanently prevent
a `LocalQueue` from being provisioned for that namespace, with no retry.

This PR accumulates errors encountered while ensuring `LocalQueue`s exist
across all matching `ClusterQueue`s in a reconcile pass, and returns them
as an aggregated error via `utilerrors.NewAggregate`, so controller-runtime
requeues the request when something goes wrong. Behavior is unchanged
when there are no errors (`NewAggregate(nil)` returns `nil`).

Added a unit test that injects a `Create` failure via the fake client's
interceptor and asserts that `Reconcile` now returns a non-nil error
(previously it did not).

Fixes #12569

Verified via unit tests (`go test ./cmd/experimental/kueue-populator/pkg/controller/...`),
which pass, including the new regression test. I wasn't able to run the
local integration/e2e suites for `kueue-populator` due to local envtest/CRD
generation setup, but they don't exercise this reconcile path directly and
should be covered by CI.

```release-note
kueue-populator: Fixed a bug where an error creating a LocalQueue was logged but not returned from Reconcile, 
preventing controller-runtime from retrying. LocalQueue creation failures are now aggregated and returned so the request is requeued.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Reconciliation now returns an error when LocalQueue creation fails (instead of silently succeeding).
  * When multiple matching ClusterQueues fail, reconciliation returns a combined error so all failures are surfaced.
* **Tests**
  * Added deterministic test coverage for transient LocalQueue creation failures.
  * Verifies reconciliation returns the injected error, does not create LocalQueue objects, and emits the expected `LocalQueueCreationFailed` warning event.

### DIFF
--- a/cmd/experimental/kueue-populator/pkg/controller/controller.go
+++ b/cmd/experimental/kueue-populator/pkg/controller/controller.go
@@ -26,7 +26,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
-
 	"k8s.io/client-go/tools/events"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
--- a/cmd/experimental/kueue-populator/pkg/controller/controller.go
+++ b/cmd/experimental/kueue-populator/pkg/controller/controller.go
@@ -25,6 +25,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/tools/events"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -147,6 +148,7 @@ func (r *KueuePopulatorReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		}
 	}
 
+	var errs []error
 	for i := range cqs.Items {
 		cq := &cqs.Items[i]
 		if !cq.DeletionTimestamp.IsZero() {
@@ -154,10 +156,11 @@ func (r *KueuePopulatorReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		}
 		if err := r.ensureLocalQueueExists(ctx, cq, &ns); err != nil {
 			log.Error(err, "Failed to ensure LocalQueue exists in namespace", "clusterQueue", klog.KObj(cq), "namespace", klog.KObj(&ns))
+			errs = append(errs, err)
 		}
 	}
 
-	return ctrl.Result{}, nil
+	return ctrl.Result{}, utilerrors.NewAggregate(errs)
 }
 
 func (r *KueuePopulatorReconciler) SetupWithManager(mgr ctrl.Manager) error {
--- a/cmd/experimental/kueue-populator/pkg/controller/controller.go
+++ b/cmd/experimental/kueue-populator/pkg/controller/controller.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"context"
+	"errors"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
@@ -25,7 +26,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
-	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+
 	"k8s.io/client-go/tools/events"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -148,7 +149,7 @@ func (r *KueuePopulatorReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		}
 	}
 
-	var errs []error
+	var errs error
 	for i := range cqs.Items {
 		cq := &cqs.Items[i]
 		if !cq.DeletionTimestamp.IsZero() {
@@ -156,11 +157,11 @@ func (r *KueuePopulatorReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		}
 		if err := r.ensureLocalQueueExists(ctx, cq, &ns); err != nil {
 			log.Error(err, "Failed to ensure LocalQueue exists in namespace", "clusterQueue", klog.KObj(cq), "namespace", klog.KObj(&ns))
-			errs = append(errs, err)
+			errs = errors.Join(errs, err)
 		}
 	}
 
-	return ctrl.Result{}, utilerrors.NewAggregate(errs)
+	return ctrl.Result{}, errs
 }
 
 func (r *KueuePopulatorReconciler) SetupWithManager(mgr ctrl.Manager) error {
--- a/cmd/experimental/kueue-populator/pkg/controller/controller_test.go
+++ b/cmd/experimental/kueue-populator/pkg/controller/controller_test.go
@@ -48,12 +48,12 @@ type Event struct {
 	Message   string
 }
 
-// errCreateFailedForTest simulates a transient API error (or a TOCTOU race
-// where a conflicting object is created between Get and Create) when
-// injected via the fake client's interceptor.
-var errCreateFailedForTest = errors.New("create failed")
-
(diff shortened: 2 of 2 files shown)


j-skiba:
/ok-to-test

Thank you for your contribution!

also please sign the CLA

cmd/kueueviz/frontend/src/utils/urlHelper.test.js
```diff
@@ -0,0 +1,225 @@
+/*
```
j-skiba: It seems like unrelated change, let's remove that

NasitSony:
@j-skiba Good catch — that file got pulled in by branching off the wrong base. Rebased onto main and force-pushed; this PR now only touches controller.go/controller_test.go.

j-skiba (commented):
/lgtm

cc @mimowo

cmd/experimental/kueue-populator/pkg/controller/controller.go
```diff
@@ -147,17 +148,19 @@ func (r *KueuePopulatorReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		}
 	}
 
+	var errs []error
```
tenzen-y:
```suggestion
	var errs error
```

I think that we have more standardized approach. Please see my following comment as well

cmd/experimental/kueue-populator/pkg/controller/controller.go
```diff
@@ -147,17 +148,19 @@ func (r *KueuePopulatorReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 	for i := range cqs.Items {
 		cq := &cqs.Items[i]
 		if !cq.DeletionTimestamp.IsZero() {
 			continue
 		}
 		if err := r.ensureLocalQueueExists(ctx, cq, &ns); err != nil {
 			log.Error(err, "Failed to ensure LocalQueue exists in namespace", "clusterQueue", klog.KObj(cq), "namespace", klog.KObj(&ns))
+			errs = append(errs, err)
```
tenzen-y:
```suggestion
			errs = erros.Join(errs, err)
```

cmd/experimental/kueue-populator/pkg/controller/controller.go
```diff
@@ -147,17 +148,19 @@ func (r *KueuePopulatorReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 		}
 	}
 
-	return ctrl.Result{}, nil
+	return ctrl.Result{}, utilerrors.NewAggregate(errs)
```
tenzen-y:
```suggestion
	return ctrl.Result{}, errs
```

cmd/experimental/kueue-populator/pkg/controller/controller_test.go
```diff
@@ -44,6 +48,11 @@ type Event struct {
 	Message   string
 }
 
+// errCreateFailedForTest simulates a transient API error (or a TOCTOU race
+// where a conflicting object is created between Get and Create) when
+// injected via the fake client's interceptor.
+var errCreateFailedForTest = errors.New("create failed")
+
 func TestKueuePopulatorReconciler(t *testing.T) {
```
tenzen-y:
```suggestion
func TestKueuePopulatorReconciler(t *testing.T) {
	// errCreateFailedForTest simulates a transient API error (or a TOCTOU race
	// where a conflicting object is created between Get and Create) when
	// injected via the fake client's interceptor.
	var errCreateFailedForTest = errors.New("create failed")

```

We shouldn't expose such error in this test package. Any reason?

tenzen-y (commented):
Thank you for contributing kueue populator!

NasitSony:
@tenzen-y hanks for the review — switched to errors.Join and scoped errCreateFailedForTest to inside the test function. No particular reason it was package-level originally, just didn't think about scope. Pushed.

tenzen-y:
@NasitSony Could you address lint errors?

NasitSony:
@tenzen-y Fixed — gci was flagging a stray blank line inside the k8s.io/* import group (leftover from an earlier edit). Ran golangci-lint run --fix locally, confirmed 0 issues, and re-ran the unit tests, all passing. Pushed.

tenzen-y (commented):
Thank you!
/lgtm
/approve

tenzen-y:
@NasitSony can you manually submit cherry pick PR to release-0.17 branch? You can use hack/cherry_pick_pull.sh script.

NasitSony:
@tenzen-y  Got it — will do the manual cherry-pick to release-0.17 this afternoon using hack/cherry_pick_pull.sh and follow up here.

mimowo:
/release-note-edit
```release-note
kueue-populator : Fixed a bug where an error creating a LocalQueue was logged but not returned from Reconcile, 
preventing controller-runtime from retrying. LocalQueue creation failures are now aggregated and returned so the request is requeued.
```

mimowo:
/release-note-edit
```release-note
kueue-populator: Fixed a bug where an error creating a LocalQueue was logged but not returned from Reconcile, 
preventing controller-runtime from retrying. LocalQueue creation failures are now aggregated and returned so the request is requeued.
```

PR #12797, opened 2026-07-06
Keep the sticky workload consistent during ClusterQueue heap and snapshot sorts

### LINKED ISSUE #12740: Unsynchronized writes to the sticky workload might make heap and snapshot sorting non-transitive
**What happened**:
We fixed https://github.com/kubernetes-sigs/kueue/issues/12532 by using an atomic pointer in the `stickyWorkload` struct (https://github.com/kubernetes-sigs/kueue/pull/12736). This fixes a data race caused by concurrent reads and writes to that part of memory.

While it's an improvement, I have a hunch that this was a symptom of another issue - the fact that `stickyWorkload` is being written to without a "broader" lock (https://github.com/kubernetes-sigs/kueue/pull/12736#discussion_r3518672605).

`RequeueIfNotPresent` is setting the `stickyWorkload` without the lock:
https://github.com/kubernetes-sigs/kueue/blob/f1aeeb4150b0542d65ac3c8ab735282bb0a3756a/pkg/cache/queue/cluster_queue.go#L854-L864

The `stickyWorkload` is used to sort the CQ heap and the CQ snapshot used in the visibility server. If it changes mid-sort, it might cause corruptions/panics. 

This has to be carefully considered/reproduced. An intuitive fix would be to make sure the `stickyWorkload` is not written to without acquiring the lock.

somaz94:
#### What type of PR is this?
/kind bug

#12736 made `stickyWorkload` an atomic pointer, which fixed the data race but
not the underlying ordering problem it was a symptom of (#12740): the sticky
workload drives both the CQ heap ordering and the visibility-server Snapshot
ordering, and it can change mid-sort, making the comparison non-transitive and
corrupting the resulting order.

This fixes both sort paths:

- Heap: the sticky-workload write is moved from `RequeueIfNotPresent` (which
  held no lock) into `requeueIfNotPresent` under `rwm.Lock()`, so heap
  operations, which also hold the lock, never observe it changing mid-sort.
- Snapshot: this sort runs without the lock by design (FS usage is
  pre-computed), so instead of locking it, the comparator now captures the
  sticky workload once per sort via a new `stickyWorkload.matcher()`, keeping a
  single sort transitive while preserving the lock-free snapshot.

Fixes #12740

The reporter suggested acquiring the broader lock before writing the sticky
workload. That covers the heap, but the Snapshot sort is intentionally
lock-free, so locking the write alone would not make it transitive; hence the
per-sort capture for Snapshot. If you would rather hold the read lock around the
Snapshot sort instead, I am happy to switch to that.

Validation (local):
- `go test -race ./pkg/cache/queue/...` passes
- the new test `TestSnapshotConsistentUnderConcurrentStickyChange` fails
  deterministically on main (non-transitive order) and passes with this fix
- `gofmt` clean, `golangci-lint` reports 0 issues on the package

This PR was written with AI assistance (Claude Code), per the Kubernetes AI
contribution guidance; I have reviewed and validated every change.

```release-note
Scheduling: Fixed a concurrency bug in BestEffortFIFO ClusterQueues where the
sticky Workload could change while pending Workloads were being sorted, making
the comparison non-transitive and potentially corrupting the scheduler queue or
visibility snapshot ordering.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ClusterQueue sticky-workload selection so heap admission remains transitive and consistent even as sticky signals change concurrently.
  * Stabilized Snapshot ordering by capturing sticky state once per snapshot sort to prevent mid-sort inconsistencies.
  * Updated requeue handling for BestEffortFIFO so sticky state is applied under the proper lock for specific pending conditions.
* **Tests**
  * Added a concurrency-focused test to verify Snapshot order remains consistent while sticky scheduling changes during snapshot creation.

### DIFF
--- a/pkg/cache/queue/cluster_queue.go
+++ b/pkg/cache/queue/cluster_queue.go
@@ -98,8 +98,8 @@ var (
 //   - Writes (set/clear) happen under the ClusterQueue's rwm lock, so heap
 //     operations, which also hold the lock, never observe it changing mid-sort.
 //   - Snapshot sorts a copy of the pending workloads without holding the lock,
-//     so it captures the sticky workload once per sort via matcher() instead of
-//     re-reading it on every comparison.
+//     so it captures the sticky workload once per sort via capturedMatcher()
+//     instead of re-reading it on every comparison.
 //
 // The field holds a single whole value (nil means no sticky workload), so an
 // atomic.Pointer keeps individual reads and writes memory-safe on top of the
@@ -113,11 +113,11 @@ func (s *stickyWorkload) matches(workload workload.Reference) bool {
 	return name != nil && *name == workload
 }
 
-// matcher captures the current sticky workload once and returns a predicate
-// bound to that fixed value. A sort that compares through the returned predicate
-// stays transitive even if set/clear runs concurrently, because every
+// capturedMatcher captures the current sticky workload once and returns a
+// predicate bound to that fixed value. A sort that compares through the returned
+// predicate stays transitive even if set/clear runs concurrently, because every
 // comparison in that sort observes the same sticky workload. See Kueue#12740.
-func (s *stickyWorkload) matcher() func(workload.Reference) bool {
+func (s *stickyWorkload) capturedMatcher() func(workload.Reference) bool {
 	name := s.workloadName.Load()
 	if name == nil {
 		return func(workload.Reference) bool { return false }
@@ -575,17 +575,17 @@ func resolveQuotaReservedReason(reason QuotaReservedReason) QuotaReservedReason 
 func (c *ClusterQueue) requeueIfNotPresent(log logr.Logger, wInfo *workload.Info, immediate bool, reason RequeueReason, quotaReservedReason QuotaReservedReason) bool {
 	c.rwm.Lock()
 	defer c.rwm.Unlock()
+	key := workload.Key(wInfo.Obj)
 	// When preemptions are in-progress, keep re-attempting the same workload at
 	// the head for BestEffortFIFO queues (see documentation of stickyWorkload).
 	// The sticky workload is set under the lock so heap operations, which also
 	// hold the lock, never observe it changing mid-sort. See Kueue#12740.
 	if (reason == RequeueReasonPendingPreemption || reason == RequeueReasonPendingMigration) && c.queueingStrategy == kueue.BestEffortFIFO {
 		if logV := log.V(5); logV.Enabled() {
-			logV.Info("Setting sticky workload", "clusterQueue", wInfo.ClusterQueue, "workload", workload.Key(wInfo.Obj))
+			logV.Info("Setting sticky workload", "clusterQueue", wInfo.ClusterQueue, "workload", key)
 		}
-		c.sw.set(workload.Key(wInfo.Obj))
+		c.sw.set(key)
 	}
-	key := workload.Key(wInfo.Obj)
 	// wasTracked records whether the workload is already counted in pendingResourcesTotal before inflight state is cleared.
 	// Guards addPendingResources below to prevent double-counting if the workload was concurrently re-added to the heap or inadmissible.
 	wasTracked := c.inadmissibleWorkloads.hasKey(key) || c.heap.GetByKey(key) != nil
@@ -826,7 +826,7 @@ func (c *ClusterQueue) Snapshot() []*workload.Info {
 
 // buildSnapshotSort returns a function that sorts workload elements for Snapshot().
 // The sort runs without holding the ClusterQueue lock, so it captures the sticky
-// workload once per sort (via stickyWorkload.matcher) to keep the comparison
+// workload once per sort (via stickyWorkload.capturedMatcher) to keep the comparison
 // transitive even if the sticky workload changes concurrently. See Kueue#12740.
 // When fair-sharing is enabled, it also pre-computes FS usage per LocalQueue from
 // deep-copied AFS state to avoid inconsistent comparisons from concurrent updates.
@@ -843,7 +843,7 @@ func buildSnapshotSort(
 	log := ctrl.LoggerFrom(ctx)
 	if !enableAdmissionFs {
 		return func(elements []*workload.Info) {
-			slices.SortFunc(elements, baseCompareFunc(log, wo, sw.matcher()))
+			slices.SortFunc(elements, baseCompareFunc(log, wo, sw.capturedMatcher()))
 		}
 	}
 
@@ -863,7 +863,7 @@ func buildSnapshotSort(
 	return func(elements []*workload.Info) {
 		// Capture the sticky workload once so the sort stays transitive without
 		// holding the lock. See Kueue#12740.
-		baseCmp := baseCompareFunc(log, wo, sw.matcher())
+		baseCmp := baseCompareFunc(log, wo, sw.capturedMatcher())
 		usageCache := make(map[utilqueue.LocalQueueReference]float64)
 		for _, wInfo := range elements {
 			lqKey := utilqueue.KeyFromWorkload(wInfo.Obj)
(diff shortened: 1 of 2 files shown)


somaz94:
Marked ready for review — CI is fully green (all e2e/integration/unit/perf jobs passing). Thanks for the `/ok-to-test`, @mimowo.

One open design decision for reviewers: the Snapshot sort is intentionally lock-free, so I made the comparator capture the sticky workload once per sort via `stickyWorkload.matcher()` (keeping a single sort transitive) rather than holding `rwm` around it. If you'd prefer holding the read lock around the Snapshot sort instead, I'm happy to switch — let me know which you'd rather have.

mimowo:
> One open design decision for reviewers: the Snapshot sort is intentionally lock-free, so I made the comparator capture the sticky workload once per sort via stickyWorkload.matcher() (keeping a single sort transitive) rather than holding rwm around it. If you'd prefer holding the read lock around the Snapshot sort instead, I'm happy to switch — let me know which you'd rather have.

This design choice is sensible to me and preferred to freeze the matcher rather than using extended locking. I would love to have a pass from @kshalot

pkg/cache/queue/cluster_queue.go
```diff
@@ -107,6 +113,17 @@ func (s *stickyWorkload) matches(workload workload.Reference) bool {
...
 
+// matcher captures the current sticky workload once and returns a predicate
+// bound to that fixed value. A sort that compares through the returned predicate
+// stays transitive even if set/clear runs concurrently, because every
+// comparison in that sort observes the same sticky workload. See Kueue#12740.
+func (s *stickyWorkload) matcher() func(workload.Reference) bool {
+	name := s.workloadName.Load()
+	return func(key workload.Reference) bool {
+		return name != nil && *name == key
+	}
```
kshalot, line 128: Since we capture the name once, we could also perform the nil check and pointer dereference once, instead of doing it NlogN times during sorting. Something like:

```suggestion
	name := s.workloadName.Load()
	if name == nil {
		return func(workload.Reference) bool { return false }
	}
	captured := *name
	return func(key workload.Reference) bool {
		return captured == key
	}
```

kshalot:
> One open design decision for reviewers: the Snapshot sort is intentionally lock-free, so I made the comparator capture the sticky workload once per sort via `stickyWorkload.matcher()` (keeping a single sort transitive) rather than holding `rwm` around it.

I agree with this direction. We don't want calls to the visibility server to induce a big overhead on the Kueue scheduler.

pkg/cache/queue/cluster_queue.go
somaz94: Good call — applied in `2b8990238`. `matcher()` now does the `nil` check and dereference once up front (returning a constant-`false` predicate when there's no sticky workload), so each comparison in the sort just does the `captured == key` check instead of re-checking `nil` and dereferencing `NlogN` times. `go test -race ./pkg/cache/queue/...` still green.

somaz94:
Thanks @kshalot — glad the lock-free direction lands well. That was exactly the goal: freezing the sticky matcher once per sort (`matcher()` captures the workload key and returns a constant-`false` predicate when there's none) keeps each comparison to a single `captured == key` check and avoids any visibility-server / locking overhead on the scheduler hot path.

CI is fully green (all e2e/integration/unit/perf), and the earlier inline nit is applied (`2b8990238`). Nothing outstanding on my side — happy to take a full review pass / lgtm whenever you and @mimowo are ready. 🙏

pkg/cache/queue/cluster_queue.go
```diff
@@ -107,6 +113,21 @@ func (s *stickyWorkload) matches(workload workload.Reference) bool {
 	return name != nil && *name == workload
 }
 
+// matcher captures the current sticky workload once and returns a predicate
+// bound to that fixed value. A sort that compares through the returned predicate
+// stays transitive even if set/clear runs concurrently, because every
+// comparison in that sort observes the same sticky workload. See Kueue#12740.
+func (s *stickyWorkload) matcher() func(workload.Reference) bool {
```
kshalot: Tiny thought about the name, although I don't have a very strong opinion - `matcher` seems a little generic. Maybe we can consider `snapshotMatcher` or `capturedMatcher`.

Because if you have the `sw.matches` and `sw.matcher()` references right next to each other it looks slightly confusing.

pkg/cache/queue/cluster_queue.go
```diff
@@ -552,6 +575,16 @@ func resolveQuotaReservedReason(reason QuotaReservedReason) QuotaReservedReason
...
 	defer c.rwm.Unlock()
+	// When preemptions are in-progress, keep re-attempting the same workload at
+	// the head for BestEffortFIFO queues (see documentation of stickyWorkload).
+	// The sticky workload is set under the lock so heap operations, which also
+	// hold the lock, never observe it changing mid-sort. See Kueue#12740.
+	if (reason == RequeueReasonPendingPreemption || reason == RequeueReasonPendingMigration) && c.queueingStrategy == kueue.BestEffortFIFO {
+		if logV := log.V(5); logV.Enabled() {
+			logV.Info("Setting sticky workload", "clusterQueue", wInfo.ClusterQueue, "workload", workload.Key(wInfo.Obj))
+		}
+		c.sw.set(workload.Key(wInfo.Obj))
```
kshalot: Nit: You can extract `key := workload.Key(wInfo.Obj)` to a variable so it's not evaluated twice.

pkg/cache/queue/cluster_queue_test.go
```diff
@@ -530,6 +531,96 @@ func TestSnapshotConcurrentWithRequeueNoDataRace(t *testing.T) {
...
+				for _, k := range keys {
+					cq.sw.set(k)
+				}
+				cq.sw.clear()
+			}
+		}
+	}()
+	defer close(stop)
+
+	for i := range 2000 {
```
kshalot, line 617: Maybe this is not an issue at all, but I'm wondering if we can have a situation where this 2000 iteration loop finishes before the setter goroutine gets woken up 🤔

If we want to be extra safe, we could close some `started` channel in that goroutine and block the main thread on it to have 100% guarantee.

pkg/cache/queue/cluster_queue.go
somaz94: Good call — renamed to `capturedMatcher`. It also lines up with the wording already in `baseCompareFunc`'s doc comment ("a live matcher ... or a captured one"), so `matches` = live, `capturedMatcher` = captured now reads consistently.

pkg/cache/queue/cluster_queue.go
somaz94: Done — hoisted `key := workload.Key(wInfo.Obj)` above the sticky block and reused it for both the log line and `sw.set()`.

pkg/cache/queue/cluster_queue_test.go
somaz94: Good catch — added. The writer now closes a `started` channel and the main thread blocks on it before the loop.

For what it's worth, I checked whether the loop was actually finishing early: reverting the fix (swapping `capturedMatcher()` back to `matches`) makes the test fail on **call 1**, so the writer was already churning before the first Snapshot. But that's a scheduling accident rather than a guarantee, so the `started` channel is worth having regardless.

kshalot (commented):
/lgtm

Thanks for working on this!

mimowo:
/approve
Thank you for the bugfix, the code is also more understandable because mutability is really tricky👍 
/cherrypick release-0.18
/cherrypick release-0.17

mimowo:
/retest
infra overload

mimowo:
@somaz94 would you like to follow up with a CP?

mimowo:
/release-note-edit
```release-note
Scheduling: Fixed a concurrency bug in BestEffortFIFO ClusterQueues where the
sticky Workload could change while pending Workloads were being sorted, making
the comparison non-transitive and potentially corrupting the scheduler queue or
visibility snapshot ordering.
```

somaz94:
@mimowo Yes — opened both manually since the bot couldn't apply:

- `release-0.18`: #13428
- `release-0.17`: #13429

`release-0.18` was a one-line conflict (the `quotaReservedReason` signature that only exists on `main`). `release-0.17` needed a bit more adaptation — its private `requeueIfNotPresent` predates the `reason` parameter, so I threaded it through, kept the `PendingPreemption`-only sticky condition that matches that branch, and excluded the `main`-only `wasTracked` / `pendingResources` code and tests the 3-way merge tried to pull in. Details are in the #13429 description. Both are `go build` / `vet` / `test` green (incl. the new regression test), and `-race` clean on the sticky paths.

PR #12807, opened 2026-07-06
TAS: prevent tas-ungater divide-by-zero/nil panic on invalid subGroupCount

reruno:
#### What type of PR is this?
/kind bug

TAS: prevent tas-ungater divide-by-zero/nil panic on invalid subGroupCount

Fixes [RepoInspector] HIGH: Denial of Service (Panic) via Unvalidated TAS SubGroup Annotations in pkg/controller/tas/topology_ungater.go

```release-note
TAS: Fixed a bug where a PodSet with `subGroupIndexLabel` set but a missing or zero `subGroupCount` could crash the tas-ungater controller. Kueue now falls back to greedy domain assignment for these pods instead of panicking.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed scheduling/ungating when a subgroup index label is set but the subgroup count is missing or invalid, so reconciliation completes safely and node selection falls back to the expected topology-based greedy mapping.
  * Preserves existing job labeling (including completion index and job index labels).

* **Tests**
  * Added regression tests for single-pod workloads covering omitted and zero subgroup count cases.

* **API / CRD Validation**
  * Enforced non-negative values for `subGroupCount` via CRD validation (minimum `0`).

### DIFF
--- a/pkg/controller/tas/topology_ungater.go
+++ b/pkg/controller/tas/topology_ungater.go
@@ -59,6 +59,7 @@ import (
 var (
 	errPendingUngateOps      = errors.New("pending ungate operations")
 	errParseOffsetAnnotation = errors.New("failed to parse offset annotation")
+	errInvalidSubGroupCount  = errors.New("invalid subgroup count for podset with subgroup index label")
 )
 
 type topologyUngaterOptions struct {
@@ -454,9 +455,12 @@ func readRanksIfAvailable(log logr.Logger,
 	}
 	result, err := readRanksForLabels(psa, pods, psReq, offset, maxRank)
 	if err != nil {
-		if errors.Is(err, utilpod.ErrLabelNotFound) {
+		switch {
+		case errors.Is(err, utilpod.ErrLabelNotFound):
 			log.V(5).Info("pods missing index label for rank ordering", "error", err)
-		} else {
+		case errors.Is(err, errInvalidSubGroupCount):
+			log.V(3).Info("invalid subGroupCount, falling back to greedy assignment", "error", err)
+		default:
 			log.Error(err, "failed to read rank information from pods")
 		}
 		return nil, false
@@ -491,6 +495,9 @@ func readRanksForLabels(
 	podSetSize := int(*psa.Count)
 	singleJobSize := podSetSize
 	if psReq.SubGroupIndexLabel != nil {
+		if psReq.SubGroupCount == nil || *psReq.SubGroupCount <= 0 {
+			return nil, fmt.Errorf("%w %q: must be set and greater than 0", errInvalidSubGroupCount, *psReq.SubGroupIndexLabel)
+		}
 		singleJobSize = podSetSize / int(*psReq.SubGroupCount)
 	}
 
--- a/pkg/controller/tas/topology_ungater_test.go
+++ b/pkg/controller/tas/topology_ungater_test.go
@@ -166,6 +166,111 @@ func TestReconcile(t *testing.T) {
 				},
 			},
 		},
+		"ungate single pod with sub group index label but no sub group count": {
+			// Regression test: a PodSet with SubGroupIndexLabel set but SubGroupCount
+			// nil (e.g. from an unvalidated user-supplied annotation) must not panic
+			// the controller; it should fall back to greedy domain assignment.
+			workloads: []kueue.Workload{
+				*utiltestingapi.MakeWorkload("unit-test", "ns").Finalizers(kueue.ResourceInUseFinalizerName).
+					PodSets(*utiltestingapi.MakePodSet(kueue.DefaultPodSetName, 1).
+						Request(corev1.ResourceCPU, "1").
+						PodIndexLabel(new(batchv1.JobCompletionIndexAnnotation)).
+						SubGroupIndexLabel(new(jobset.JobIndexKey)).
+						Obj()).
+					ReserveQuotaAt(
+						utiltestingapi.MakeAdmission("cq").
+							PodSets(utiltestingapi.MakePodSetAssignment(kueue.DefaultPodSetName).
+								Assignment(corev1.ResourceCPU, "unit-test-flavor", "1").
+								TopologyAssignment(utiltestingapi.MakeTopologyAssignment(defaultTestLevels).
+									Domains(utiltestingapi.MakeTopologyDomainAssignment([]string{"b1", "r1"}, 1).Obj()).
+									Obj()).
+								Obj()).
+							Obj(), now,
+					).
+					AdmittedAt(true, now).
+					Obj(),
+			},
+			pods: []corev1.Pod{
+				*testingpod.MakePod("pod", "ns").
+					Annotation(kueue.WorkloadAnnotation, "unit-test").
+					Label(constants.PodSetLabel, string(kueue.DefaultPodSetName)).
+					Label(batchv1.JobCompletionIndexAnnotation, "0").
+					Label(jobset.JobIndexKey, "0").
+					TopologySchedulingGate().
+					Obj(),
+			},
+			nodeSelectorAssertMode: nodeSelectorAssertCountsOnly,
+			wantPods: []corev1.Pod{
+				*testingpod.MakePod("pod", "ns").
+					Annotation(kueue.WorkloadAnnotation, "unit-test").
+					Label(constants.PodSetLabel, string(kueue.DefaultPodSetName)).
+					Label(batchv1.JobCompletionIndexAnnotation, "0").
+					Label(jobset.JobIndexKey, "0").
+					Obj(),
+			},
+			wantCounts: []counts{
+				{
+					NodeSelector: map[string]string{
+						tasBlockLabel: "b1",
+						tasRackLabel:  "r1",
+					},
+					Count: 1,
+				},
+			},
+		},
+		"ungate single pod with sub group index label but zero sub group count": {
+			// Regression test: a PodSet with SubGroupIndexLabel set but SubGroupCount
+			// zero (the literal divide-by-zero) must not panic the controller; it should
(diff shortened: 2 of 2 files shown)


apis/kueue/v1beta2/workload_types.go
```diff
@@ -208,6 +208,7 @@ type PodSetTopologyRequest struct {
 
 	// subGroupCount indicates the count of replicated Jobs (groups) within a PodSet.
 	// For example, in the context of JobSet this value is read from jobset.sigs.k8s.io/replicatedjob-replicas.
+	// +kubebuilder:validation:Minimum=1
```
dkaluza: Wouldn't this cause regression for https://github.com/kubernetes-sigs/kueue/issues/2227

IMO generally we should allow 0 replicas in JobSets and other workload sources. We just should handle those in graceful way - avoid divisions by replica count

reruno: Thanks for the comment. Removed the validation and additional check in build() for appwrapper, and only left the guard against divide by zero panic.

pkg/controller/tas/topology_ungater.go
```diff
@@ -491,6 +491,9 @@ func readRanksForLabels(
 	podSetSize := int(*psa.Count)
 	singleJobSize := podSetSize
 	if psReq.SubGroupIndexLabel != nil {
+		if psReq.SubGroupCount == nil || *psReq.SubGroupCount <= 0 {
```
ArangoGutierrez, line 498: Good, returning here also covers the second *psReq.SubGroupCount deref in the ReadUIntFromLabelBelowBound call lower in the loop, so one check handles both. Note this path is a fallback to greedy assignment, not a rejection, so the pod is still ungated and scheduled. That is the bit the release note should reflect (top-level comment).

pkg/controller/tas/topology_ungater_test.go
```diff
@@ -166,6 +166,58 @@ func TestReconcile(t *testing.T) {
 				},
 			},
 		},
+		"ungate single pod with sub group index label but no sub group count": {
```
ArangoGutierrez, line 169: nit: I confirmed this case panics on the base branch without the guard (nil deref at the division) and that it asserts the greedy placement rather than only that no panic happened, so it is a real regression test. It only exercises the nil branch though. The guard also rejects subGroupCount <= 0, which is the literal divide-by-zero, and that branch has no coverage. Could you add a second case with SubGroupCount(ptr.To[int32](0)) so both halves of the check are tested?

ArangoGutierrez (commented):
Took a close look at this one. The fix is small and sits in the right place. The guard in readRanksForLabels stops both failure modes: the nil deref and the divide-by-zero. It returns before the second read of *psReq.SubGroupCount further down in the loop, so both dereferences are covered by the single check.

I traced the callers. assignGatedPodsToDomains falls back to greedy assignment when readRanksIfAvailable returns false, so an invalid subGroupCount now means the pods are placed by greedy domain assignment, keeping the admitted topology domains but dropping the rank ordering, instead of crashing the controller. I also checked where the bad state can come from: the workload webhook (validatePodSet) does not validate topology-request fields, the CRD has no minimum on subGroupCount, and the AppWrapper adapter keeps subGroupIndexLabel even when the count annotation is malformed and dropped. So the invalid combination can reach the ungater from more than one path, and guarding here catches all of them. Rejecting at the webhook would not, and it would regress the zero-replica case in #2227, which is the direction dkaluza already asked for. Agree with that call.

One thing to fix before this merges: the release-note block and the "What this PR does" line still say Kueue "validates and rejects such PodSets." The code does not reject anything now. It admits and schedules the workload with greedy placement. The CodeRabbit summary in the body actually describes the real behavior. Can you update the note to something like "Kueue now falls back to greedy domain assignment for these pods instead of panicking"? That is also how your commit message describes it. Otherwise the release note tells users the opposite of what happens.

Optional: on a workload that stays invalid, readRanksForLabels returns this error every reconcile and readRanksIfAvailable logs it at Error level each time (topology_ungater.go:460, unchanged by this PR). Existing errors on that path are transient during rollout; this one is permanent for the life of the workload, so it will repeat. Not blocking, but worth a thought on whether it should log once or at a lower verbosity.

This looks good to me once the release note is corrected. Nits inline.

pkg/controller/tas/topology_ungater_test.go
reruno: Added another testcase  "ungate single pod with sub group index label but zero sub group count"

pkg/controller/tas/topology_ungater.go
reruno: Thanks for noticing, updated the release notes and added new error case in `readRanksIfAvailable` with lower log level for this kind of error 
```
		case errors.Is(err, errInvalidSubGroupCount):
			log.V(3).Info("invalid subGroupCount, falling back to greedy assignment", "error", err)
```

pkg/controller/tas/topology_ungater.go
```diff
@@ -491,6 +495,9 @@ func readRanksForLabels(
...
+			return nil, fmt.Errorf("%w %q: must be set and greater than 0", errInvalidSubGroupCount, *psReq.SubGroupIndexLabel)
```
dkaluza, line 499: nit: Consider a structured error if we need to have values like `SubGroupIndexLabel` or the count inside at some point. This will allow for easier retrieval in tests if ever needed. However we can also add those when needed in the future.

dkaluza (approved):

pkg/controller/tas/topology_ungater.go
```diff
@@ -491,6 +495,9 @@ func readRanksForLabels(
 	podSetSize := int(*psa.Count)
 	singleJobSize := podSetSize
 	if psReq.SubGroupIndexLabel != nil {
+		if psReq.SubGroupCount == nil || *psReq.SubGroupCount <= 0 {
```
mbobrovskyi, line 498: Could you please also add validation in API?

pkg/controller/tas/topology_ungater_test.go
```diff
@@ -166,6 +166,111 @@ func TestReconcile(t *testing.T) {
...
+			// Regression test: a PodSet with SubGroupIndexLabel set but SubGroupCount
+			// nil (e.g. from an unvalidated user-supplied annotation) must not panic
+			// the controller; it should fall back to greedy domain assignment.
+			workloads: []kueue.Workload{
+				*utiltestingapi.MakeWorkload("unit-test", "ns").Finalizers(kueue.ResourceInUseFinalizerName).
+					PodSets(*utiltestingapi.MakePodSet(kueue.DefaultPodSetName, 1).
+						Request(corev1.ResourceCPU, "1").
+						PodIndexLabel(ptr.To(batchv1.JobCompletionIndexAnnotation)).
```
mbobrovskyi:
```suggestion
						PodIndexLabel(new(batchv1.JobCompletionIndexAnnotation)).
```
nit: use new instead

mbobrovskyi: Also in other places

pkg/controller/tas/topology_ungater.go
dkaluza: IMO we should not block 0 counts on the API level. They are generally considered valid and in many systems computed automatically.
This would lead to regression on bugs like:
https://github.com/kubernetes-sigs/kueue/issues/2227

reruno: I can add validation like
Minimum=0

To make sure 0 is possible and negative numbers are rejected 

@dkaluza wdyt?

dkaluza: Makes sense to me

reruno: Added the Minimum=0 to subgroupcount like that 
```
// subGroupCount indicates the count of replicated Jobs (groups) within a PodSet.
	// For example, in the context of JobSet this value is read from jobset.sigs.k8s.io/replicatedjob-replicas.
	// +kubebuilder:validation:Minimum=0
	// +optional
	SubGroupCount *int32 `json:"subGroupCount,omitempty"`
```

pkg/controller/tas/topology_ungater.go
reruno: I think if we have the use case for that, if the logic of error handling requires to know the number of invalid subgroupcount number we will add, but for now we need just the error message for logs, so I think we should keep it as is

mbobrovskyi:
Thank you!

/lgtm

apis/kueue/v1beta2/workload_types.go
```diff
@@ -208,6 +208,7 @@ type PodSetTopologyRequest struct {
 
 	// subGroupCount indicates the count of replicated Jobs (groups) within a PodSet.
 	// For example, in the context of JobSet this value is read from jobset.sigs.k8s.io/replicatedjob-replicas.
+	// +kubebuilder:validation:Minimum=0
```
mimowo: Can we drop the schema update? It is generally better to avoid cherrypicking schema changes unless really needed. 

The reason is that very often users during rolling update of a new patch release just replace the image, without re-uploadiing CRDs. If we change schema this will create inconsistency on some environments, I'm pretty sure. 

So, I would like to consider this as a follow up for 0.19+.

reruno: I will separate schema update into different PR then

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.1&
thank you for hardening of Kueue 👍

apis/kueue/v1beta2/workload_types.go
reruno: https://github.com/kubernetes-sigs/kueue/pull/13061

schema update PR

PR #12810, opened 2026-07-06
API: update ControllerMetricsCustomLabel in line with KEP-7066

Singularity23x0:
#### What type of PR is this?
/kind api-change

This PR adds the new API fields to ControllerMetricsCustomLabel:
* SourceKind, denoting what type of objects the labels are sourced from,
* TrackedValues, denoting the list of allowed values for the declared label.

We are adding this field to allow specifying the source type for the label.
This will allow us to use multiple source objects for custom label values in a single metric, as described in KEP-7066.

A concrete example of such a metric is the AdmittedActiveWorkloads metric, where we want to break the counts down not only by ClusterQueue labels/annotations, but also by Workload labels/annotations.

Progresses #12629

NONE

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added `sourceKind` and `trackedValues` to Prometheus custom-label configuration, including a default `sourceKind`.
  * Supports representing values that aren’t included in `trackedValues`.

* **Validation**
  * Enforced new limits on total custom labels, per-`sourceKind` counts, and `trackedValues` (with additional required/limit rules for `Workload`).

* **Documentation**
  * Updated the v1beta2 configuration reference with new schema fields, allowed naming behavior, and count/value-limit semantics.

* **API Compatibility**
  * Updated version conversion for the custom-label schema.


apis/config/v1beta2/configuration_types.go
```diff
@@ -216,6 +225,14 @@ type ControllerMetricsCustomLabel struct {
 	// Mutually exclusive with SourceLabelKey.
 	// +optional
 	SourceAnnotationKey string `json:"sourceAnnotationKey,omitempty"`
+
+	// SourceKind is the object kind from which the label value should be sourced.
+	// Defaults to ClusterQueue.
```
tenzen-y, line 233: Doesn't this break the exsting env? Because previously, `ControllerMetricsCustomLabel` strategy is applied into all Cohort;LocalQueue;ClusterQueue resources.

But after this PR, the only ClusterQueue is evaluated when they don't specify anything.

apis/config/v1beta2/configuration_types.go
```diff
@@ -216,6 +225,14 @@ type ControllerMetricsCustomLabel struct {
...
+	//
+	// +kubebuilder:default:=ClusterQueue
+	// +kubebuilder:validation:Enum=Cohort;LocalQueue;ClusterQueue;Workload
+	// +optional
+	SourceKind *SourceKind `json:"sourceKind,omitempty"`
```
tenzen-y, line 240:
```suggestion
	SourceKinds []SourceKind `json:"sourceKinds,omitempty"`
```

Shouldn't we make this multiple? Because previously, a single set of this strategy is applied into cohort, lq, and cq resources. But, after this PR, they should specify all explicitly. In that case, I guess that the number of needed `ControllerMetricsCustomLabel` will be increased, but the limit of the amout of custom labels are still 8.

apis/config/v1beta2/configuration_types.go
tenzen-y: I don't say that we should introduce this default value since this is still alpha feature IIUC, but if we will break this behavior, we should carefully annouce that like via RELEASE NOTE.

Singularity23x0: 1. Yes, this is a breaking change.
2. Adding the default value still makes it a breaking change, but as it is the most common source of label values, it minimizes the amount of adjustments users would be required to make.
3. Added the mention to the release notes.

apis/config/v1beta2/configuration_types.go
Singularity23x0: The decision to make it a single value is deliberate, to avoid the issue of label source conflicts (i.e. multiple source kinds having their own, different values for a specific label). This has been described in more detail in the alternatives section of the related KEP.

It is true however that this is a problem when considering the max amount of custom labels.
I propose updating the limit to 8 labels per source Kind - this would amount the same amount of memory being used by internal caching.
The downside: it allows for metrics with multiple sources to have up to [amount of distinct source kinds] * 8 added labels.
Still, introducing multi-source-kind metrics would hit that problem even if we kept the original limit, if all 8 labels are defined for all SourceKinds, so that downside does not make this proposal compare worse to the other.

apis/config/v1beta2/configuration_types.go
```diff
@@ -216,6 +228,24 @@ type ControllerMetricsCustomLabel struct {
...
+	// SourceKind is the object kind from which the label value should be sourced.
+	// Up to 2 labels are allowed for Workloads and up to 6 for other source kinds.
+	// Defaults to ClusterQueue.
+	// The possible values are:
+	// - Cohort
+	// - LocalQueue
+	// - ClusterQueue
+	// - Workloads
+	// +optional
+	SourceKind *SourceKind `json:"sourceKind,omitempty"`
```
mimowo, line 240: do we have a test case for the scenario when an unknown sourceKind is specified? (scenario we talked about briefly about offline)

pkg/config/validation.go
```diff
@@ -795,7 +803,27 @@ func validateCustomLabels(c *configapi.Configuration) field.ErrorList {
...
+
+	for kind, count := range countPerSourceKind {
+		if count > maxCustomLabelsPerSourceKind[kind] {
+			allErrs = append(allErrs, field.Invalid(
+				customLabelsPath,
+				c.Metrics.CustomLabels,
+				fmt.Sprintf("too many custom labels for source kind %s: found %d, expected <= %d", kind, count, maxCustomLabelsPerSourceKind[kind]),
+			))
+		}
+	}
```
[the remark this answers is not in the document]
mimowo: That is a good point, let's try to make sure the validation is deterministic. This makes test writing easier without too much cost.

apis/config/v1beta2/configuration_types.go
```diff
@@ -185,23 +185,35 @@ type ControllerMetrics struct {
 	EnableClusterQueueResources bool `json:"enableClusterQueueResources,omitempty"`
 
 	// CustomLabels is a list of entries whose values will be added as extra
-	// Prometheus labels on ClusterQueue, LocalQueue, and Cohort metrics.
-	// Requires the CustomMetricLabels feature gate.
+	// Prometheus labels on supported metrics.
+	// A maximum of 6 labels are allowed per SourceKind (2 for Workloads), with up to 20 labels defined in total.
 	// +optional
-	// +kubebuilder:validation:MaxItems=8
+	// +kubebuilder:validation:MaxItems=20
```
mimowo: Let's drop this marker - it is irrelevant for ConfigAPI anyway. Instead we can rely on the validation we have in code.

apis/config/v1beta2/configuration_types.go
Singularity23x0: Added along a larger test case update.

mimowo:
/release-note-none
We don't need release note just for the API change PR.

mimowo:
/release-note-edit
```release-note
NONE
```

apis/config/v1beta2/configuration_types.go
```diff
@@ -216,6 +227,24 @@ type ControllerMetricsCustomLabel struct {
...
+	// SourceKind is the object kind from which the label value should be sourced.
+	// Up to 2 labels are allowed for Workloads and up to 6 for other source kinds.
+	// Defaults to ClusterQueue.
+	// The possible values are:
+	// - Cohort
+	// - LocalQueue
+	// - ClusterQueue
+	// - Workloads
+	// +optional
+	SourceKind *SourceKind `json:"sourceKind,omitempty"`
```
olekzabl, line 240: I remember you once planned `// +kubebuilder:validation:Enum=...`, what happened to it?

mimowo:
/lgtm
/approve
Thank you. Looking forward to see the full impl

apis/config/v1beta2/configuration_types.go
Singularity23x0: This was removed since, as pointed out by others, these markers do not work on config files.

apis/config/v1beta2/configuration_types.go
```diff
@@ -216,6 +227,24 @@ type ControllerMetricsCustomLabel struct {
...
 	SourceAnnotationKey string `json:"sourceAnnotationKey,omitempty"`
+
+	// SourceKind is the object kind from which the label value should be sourced.
+	// Up to 2 labels are allowed for Workloads and up to 6 for other source kinds.
+	// Defaults to ClusterQueue.
+	// The possible values are:
+	// - Cohort
+	// - LocalQueue
+	// - ClusterQueue
+	// - Workloads
```
kshalot:
```suggestion
	// - Workload
```

apis/config/v1beta2/configuration_types.go
```diff
@@ -216,6 +227,24 @@ type ControllerMetricsCustomLabel struct {
...
+	// +optional
+	SourceKind *SourceKind `json:"sourceKind,omitempty"`
+
+	// TrackedValues is a list of the label's allowed values.
+	// When SourceKind is Workload, a closed list of 1-12 TrackedValues is required.
+	// Non-workload source kinds can have 0-16 TrackedValues. If the list is empty, any value is allowed.
+	// Label values not allowed by this field will be reported as "kueue.x-k8s.io/_UNTRACKED_VALUE_".
+	// +optional
+	TrackedValues []string `json:"trackedValues,omitempty"`
```
kshalot, line 247: Nit: Uniqueness of the values could also be validated.

kshalot (commented):
I see the PR is already merging, so feel free to ignore my review. It's just nits anyway.

apis/config/v1beta2/configuration_types.go
mimowo: Good point, let's check and once done @kshalot please stamp it

Singularity23x0: Done.

apis/config/v1beta2/configuration_types.go
```diff
@@ -216,6 +227,24 @@ type ControllerMetricsCustomLabel struct {
...
 	SourceAnnotationKey string `json:"sourceAnnotationKey,omitempty"`
+
+	// SourceKind is the object kind from which the label value should be sourced.
+	// Up to 2 labels are allowed for Workloads and up to 6 for other source kinds.
+	// Defaults to ClusterQueue.
+	// The possible values are:
+	// - Cohort
+	// - LocalQueue
+	// - ClusterQueue
+	// - Workload
```
kshalot, line 238: docs have to be regenerated after the "workloads -> "workload" change, otherwise LGTM

kshalot (commented):
/lgtm

PR #12814, opened 2026-07-06
KEP: Set empty underlying_cause for non QuotaReserved admission issues for inadmissible workloads

j-skiba:
#### What type of PR is this?
/kind kep

As discussed here https://github.com/kubernetes-sigs/kueue/pull/12759#discussion_r3520095719

We don't need to set `underlying_reason` label explicitly if it doesn't bring any value, we can assume that empty value means that the reason is the root cause of the problem as we do for other metrics:
- https://github.com/kubernetes-sigs/kueue/blob/8771f2d5e3e39014e4f822d9534b824f9ecb7bf4/pkg/metrics/metrics.go#L607
- https://github.com/kubernetes-sigs/kueue/blob/8771f2d5e3e39014e4f822d9534b824f9ecb7bf4/pkg/metrics/metrics.go#L636

Part of #10852

```release-note
NONE
```

## Summary by CodeRabbit

* **Documentation**
  * Clarified the Prometheus metrics schema for unadmitted workloads.
  * Updated guidance so `underlying_cause` is left empty when a workload is not blocked by `QuotaReserved`-related reasons.
  * Adjusted the listed `Admitted` reason values to use an empty `underlying_cause` in the affected cases.

### DIFF
--- a/keps/10852-inadmissible-workloads-observability/README.md
+++ b/keps/10852-inadmissible-workloads-observability/README.md
@@ -476,9 +476,7 @@ A set of metrics is introduced to track unadmitted workloads when the
     `AdmissionGated`).
 
 If a workload has successfully obtained a quota reservation (`QuotaReserved` is `True`),
-the `underlying_cause` label is populated as follows:
-- For `UnsatisfiedAdmissionChecks`: Set to `ChecksNotReady`.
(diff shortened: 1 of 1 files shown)


mimowo:
/lgtm
/approve
Yes, this is matching my expectations about the shape of the metrics. Thank  you for the cleanup 👍

PR #12819, opened 2026-07-06
Increase Workload PodSet limit to 18

### LINKED ISSUE #12818: Increase the number of PodSets per Workload
### What would you like to be added

Raise the per-Workload `MaxPodSets` limit from the current 10 to 16 (and ideally even higher).

This follows on from #11379, which moved the limit from 8 to 10. We'd like to continue to 16 (the original suggestion in #11379).

### Why is this needed

We run RayClusters, and the effective cap on worker groups is what bites us. Because integrations reserve technical PodSets (the Ray head, the submitter job, and Redis cleanup — see #11260), the limit of 10 leaves only a handful of PodSets for actual worker groups, and our clusters routinely need more distinct worker-group shapes than that headroom allows.

The core difficulty is that, in our case, the shape of a RayCluster is not known ahead of time. Across our platform we have on the order of ~40 distinct worker/pod types (different accelerators, memory profiles, CPU-only helpers, and so on), and any single RayCluster only ever uses a small subset of these. Currently we have no clean workaround short of forking Kueue — which we do, setting `MaxPodSets` to 40. If the limit were raised to 16 we might be able to build a proper workaround, but the current limit of 10 is too restrictive even for that.

### Notes on performance / feasibility

We have tried to gather some performance measurements (driven by bots): https://github.com/mcochner/kueue-maxpodsets-study/blob/main/REPORT.md. So far we haven't found any significant issues with increasing this limit.

mcochner:
## What type of PR is this?

/kind feature

Raises the per-Workload `MaxPodSets` limit from 10 to 18.

This bumps the `jobframework.MaxPodSets` constant and the `MaxItems` validation on the five podSet-indexed arrays in both the `v1beta1` and `v1beta2` Workload APIs, and regenerates the CRDs, Helm chart, apply configurations, and reference docs.

Integrations reserve technical podSets (for example, Ray uses a head, a submitter job, and a Redis cleanup job), which leaves too few podSets for actual worker groups. We run RayClusters whose shape is not known ahead of time and draw from a large catalog of worker/pod types, so a higher limit removes the need to fork Kueue as a workaround. This continues the direction of #11379 / #11388 (which moved the limit from 8 to 10).

Per discussion in #12818, 18 is the value maintainers aligned on for now: it gives users meaningfully more room while leaving headroom to revisit higher values in a future release, once there is feedback on how the increased count interacts with features like MultiKueue and ConcurrentAdmission.

Fixes #12818

## Special notes for your reviewer

* All limit-dependent tests and the pod-group role-count error message derive from `jobframework.MaxPodSets`, so they won't drift on future bumps.
* The oversized Ray worker-group fixtures now come from a shared `testingjobs/raycluster.MakeWorkerGroups` helper (addressing @dkaluza's review nit).
* Validation-cost note: the `Workload` CRD carries CEL rules on the podSet-indexed subtrees whose estimated cost scales with `maxItems`; 18 stays well within the apiserver's CEL cost budget.

```release-note
Workloads: Increase the maximum number of PodSets per Workload from 10 to 18.
```

## Summary by CodeRabbit

* **New Features**
  * Increased the maximum allowed sizes for several workload-related list fields (including `podSets`) from **10 to 18**.
* **Bug Fixes**
  * Updated RayJob/RayCluster/RayService documentation and validation logic to match the new PodSet capacity (worker group specs capped at **17**).
  * Refreshed validation messaging, test cases, and API reference docs to use the new maximum consistently.

### DIFF
--- a/pkg/controller/jobs/rayjob/rayjob_webhook_test.go
+++ b/pkg/controller/jobs/rayjob/rayjob_webhook_test.go
@@ -127,7 +127,11 @@ func TestDefault(t *testing.T) {
 
 func TestValidateCreate(t *testing.T) {
 	worker := rayv1.WorkerGroupSpec{}
-	bigWorkerGroup := []rayv1.WorkerGroupSpec{worker, worker, worker, worker, worker, worker, worker, worker, worker, worker}
+	// head + jobframework.MaxPodSets worker groups exceeds the per-Workload podSet limit.
+	bigWorkerGroup := make([]rayv1.WorkerGroupSpec, jobframework.MaxPodSets)
+	for i := range bigWorkerGroup {
+		bigWorkerGroup[i] = worker
+	}
 
 	testcases := map[string]struct {
 		job          *rayv1.RayJob
@@ -203,7 +207,7 @@ func TestValidateCreate(t *testing.T) {
 				WithWorkerGroups(bigWorkerGroup...).
 				Obj(),
 			wantErr: field.ErrorList{
-				field.TooMany(field.NewPath("spec", "rayClusterSpec", "workerGroupSpecs"), 11, jobframework.MaxPodSets),
+				field.TooMany(field.NewPath("spec", "rayClusterSpec", "workerGroupSpecs"), jobframework.MaxPodSets+1, jobframework.MaxPodSets),
 			}.ToAggregate(),
 		},
 		"valid managed - max worker groups with gcs fault tolerance": {
--- a/pkg/controller/jobs/rayjob/rayjob_webhook_test.go
+++ b/pkg/controller/jobs/rayjob/rayjob_webhook_test.go
@@ -36,6 +36,7 @@ import (
 	"sigs.k8s.io/kueue/pkg/features"
 	utiltesting "sigs.k8s.io/kueue/pkg/util/testing"
 	utiltestingapi "sigs.k8s.io/kueue/pkg/util/testing/v1beta2"
+	testingraycluster "sigs.k8s.io/kueue/pkg/util/testingjobs/raycluster"
 	testingrayutil "sigs.k8s.io/kueue/pkg/util/testingjobs/rayjob"
 	"sigs.k8s.io/kueue/pkg/workloadslicing"
 )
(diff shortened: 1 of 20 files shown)


pkg/controller/jobs/rayjob/rayjob_webhook_test.go
```diff
@@ -127,7 +127,11 @@ func TestDefault(t *testing.T) {
 
 func TestValidateCreate(t *testing.T) {
 	worker := rayv1.WorkerGroupSpec{}
-	bigWorkerGroup := []rayv1.WorkerGroupSpec{worker, worker, worker, worker, worker, worker, worker, worker, worker, worker}
+	// head + jobframework.MaxPodSets worker groups exceeds the per-Workload podSet limit.
```
dkaluza: Nit: We are repeating this comment and in loop creation many times across tests. Consider creating just a common helper in test utils.

dkaluza (approved):

mimowo:
@mcochner I have synced with @tenzen-y and we aligned on 18. We both read your excellent doc, but since these are in-silico experiments we would like to still give a bit more sink time for the increased values to hear from users who may use the increased values in combination with other features like MK or ConcurrentAdmission. We are open to revisit increased the PodSet count in future releases after we give some time for the users to provide feedback.

mcochner:
Thanks @mimowo @tenzen-y — aligning on 18 works for us. I've pushed two commits:

- `feat: set Workload PodSet limit to 18` — dials the limit back from 40 to 18
- `test: extract shared MakeWorkerGroups helper for Ray tests` — addresses @dkaluza's nit by replacing the duplicated make+loop across the Ray tests with a single `testingjobs/raycluster.MakeWorkerGroups` helper.

Please let me know if you are happy with the shape of this PR, looking forward to see this accepted in the main.

mimowo:
/lgtm
/approve
Thank you for letting us know about your interesting use case 
/unhold

tenzen-y:
> @mcochner I have synced with @tenzen-y and we aligned on 18. We both read your excellent doc, but since these are in-silico experiments we would like to still give a bit more sink time for the increased values to hear from users who may use the increased values in combination with other features like MK or ConcurrentAdmission. We are open to revisit increased the PodSet count in future releases after we give some time for the users to provide feedback.

Thank you for summarizing our internal discussions. Yes, we can revisit the limitations for the number of PodSet again in 0.20 and 0.21 development cycles.

mimowo:
/release-note-edit
```release-note
Workloads: Increase the maximum number of PodSets per Workload from 10 to 18.
```

PR #12821, opened 2026-07-06
Handle SchedulingEquivalenceHashing skipped workloads in UnadmittedWorkloadsObservability

### LINKED ISSUE #12769: Flaky: Workload controller interaction with scheduler with UnadmittedWorkloadsObservability feature gate enabled should dynamically transition between reasons based on precedence
**Which test is flaking?**:
Workload controller interaction with scheduler with UnadmittedWorkloadsObservability feature gate enabled should dynamically transition between reasons based on precedence

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12748/pull-kueue-test-integration-shard-2-main/2073088960673878016

**Failure message or logs**:
```
Core Controllers Suite: [It] Workload controller interaction with scheduler with UnadmittedWorkloadsObservability feature gate enabled should dynamically transition between reasons based on precedence expand_less	13s
{ failed [FAILED] Timed out after 10.001s.
The function passed to Eventually failed at /home/prow/go/src/sigs.k8s.io/kueue/test/integration/singlecluster/controller/core/workload_controller_test.go:1219 with:
Expected
    <string>: PendingEvaluation
to equal
    <string>: WaitingForQuota
In [It] at: /home/prow/go/src/sigs.k8s.io/kueue/test/integration/singlecluster/controller/core/workload_controller_test.go:1220 @ 07/03/26 17:01:40.989
}
```

j-skiba:
#### What type of PR is this?
/kind feature

This PR improves visibility for workloads that are bypassed during scheduling evaluation because an equivalent workload recently failed (when the `SchedulingEquivalenceHashing` feature gate is enabled).

Before this change, bypassed workloads would remain in the `PendingEvaluation` reason (or have no condition), hiding the actual reason why they couldn't be scheduled.

This PR propagates the granular failure reason (e.g., `WaitingForQuota`, `NoMatchingFlavor`) from the queue manager's equivalence cache to the workload's `QuotaReserved` condition. The condition message is updated to `"Bypassed scheduling evaluation because an equivalent workload recently failed"`.

If a workload already has a detailed scheduler-set condition (e.g., listing specific insufficient resources), it is preserved and not overwritten.

Fixes #12769

1. Renamed `noFitSchedulingHashes` to `noFitSchedulingHashReasons` in `ClusterQueue` to store the string NoFit reason instead of a boolean.
2. Added a helper `isPendingSchedulingOrStale` in the workload controller to identify when it's appropriate to apply the bypassed status (i.e. only when the workload lacks active scheduler diagnostics, to prevent overwriting detailed resource breakdowns).
3. Consolidated the test hook `beforeReconcile` in `workload_controller_test.go` and added comprehensive unit tests covering the bypass logic under different feature gate combinations.

```release-note
Scheduling: Workloads bypassed by the scheduling equivalence cache now receive the granular failure reason (e.g., `WaitingForQuota`) and a bypass message in their `QuotaReserved` condition, improving visibility into why the workload was unadmitted.
```

## Summary by CodeRabbit

* **New Features**
  * Added an API to retrieve the stored “no fit”/bypass reason for workloads currently marked inadmissible after an equivalent recent failure.
* **Bug Fixes**
  * Improved tracking and clearing of inadmissible bulk-move reasons to ensure “skip to inadmissible” decisions use the correct, consistent rationale.
  * Extended requeue/scheduling and controller status handling to preserve the full quota-reserved reason through transitions.
* **Tests**
  * Updated unit and integration tests to cover reason lookup, bypass flows, and bulk-move reason reset behavior.

### DIFF
--- a/pkg/cache/queue/cluster_queue.go
+++ b/pkg/cache/queue/cluster_queue.go
@@ -117,9 +117,10 @@ type ClusterQueue struct {
 	// inadmissibleWorkloads are workloads that have been tried at least once and couldn't be admitted.
 	inadmissibleWorkloads inadmissibleWorkloads
 
-	// noFitSchedulingHashReasons tracks scheduling equivalence classes and their reasons.
+	// hashToBulkMoveReason tracks scheduling equivalence classes and the reason
+	// why workloads with that hash were bulk-moved to inadmissibleWorkloads.
 	// Cleared when queueInadmissibleWorkloads runs.
-	noFitSchedulingHashReasons map[string]string
+	hashToBulkMoveReason map[string]string
 
 	finishedWorkloads sets.Set[workload.Reference]
 
@@ -246,19 +247,19 @@ func newClusterQueueImpl(ctx context.Context, client client.Client, wo workload.
 		options.afsEntryPenalties, options.afsConsumedResources,
 	)
 	return &ClusterQueue{
-		heap:                       *heap.New(workloadKey, lessFunc),
-		inadmissibleWorkloads:      make(inadmissibleWorkloads),
-		noFitSchedulingHashReasons: make(map[string]string),
-		finishedWorkloads:          sets.New[workload.Reference](),
-		queueInadmissibleCycle:     -1,
-		compareFunc:                compareFunc,
-		snapshotSort:               snapshotSort,
-		rwm:                        sync.RWMutex{},
-		clock:                      clock,
-		afsEntryPenalties:          options.afsEntryPenalties,
-		localQueuesInClusterQueue:  make(map[utilqueue.LocalQueueReference]bool),
-		sw:                         &sw,
-		pendingResourcesTotal:      make(map[corev1.ResourceName]int64),
+		heap:                      *heap.New(workloadKey, lessFunc),
+		inadmissibleWorkloads:     make(inadmissibleWorkloads),
+		hashToBulkMoveReason:      make(map[string]string),
+		finishedWorkloads:         sets.New[workload.Reference](),
+		queueInadmissibleCycle:    -1,
+		compareFunc:               compareFunc,
+		snapshotSort:              snapshotSort,
+		rwm:                       sync.RWMutex{},
+		clock:                     clock,
+		afsEntryPenalties:         options.afsEntryPenalties,
+		localQueuesInClusterQueue: make(map[utilqueue.LocalQueueReference]bool),
+		sw:                        &sw,
+		pendingResourcesTotal:     make(map[corev1.ResourceName]int64),
 	}
 }
 
@@ -377,10 +378,10 @@ func (c *ClusterQueue) PushOrUpdate(wInfo *workload.Info) {
 		c.insertInadmissible(key, wInfo)
 		return
 	}
-	// Skip to inadmissible if the workload's equivalence class is already known to be NoFit
+	// Skip to inadmissible if the workload's equivalence class was already bulk-moved to inadmissible
 	// (only for BestEffortFIFO; StrictFIFO preserves strict ordering).
 	if c.queueingStrategy == kueue.BestEffortFIFO && c.heap.GetByKey(key) == nil && wInfo.SchedulingHash != workload.SchedulingHashUnknown {
-		if _, has := c.noFitSchedulingHashReasons[wInfo.SchedulingHash]; has {
+		if _, has := c.hashToBulkMoveReason[wInfo.SchedulingHash]; has {
 			c.insertInadmissible(key, wInfo)
 			return
 		}
@@ -410,13 +411,14 @@ func draRequestsChanged(oldInfo, newInfo *workload.Info) bool {
 	return !equality.Semantic.DeepEqual(oldInfo.TotalRequests, newInfo.TotalRequests)
 }
 
-func (c *ClusterQueue) GetNoFitReason(wlKey workload.Reference, hash string) (string, bool) {
+func (c *ClusterQueue) GetNoFitReason(wl workload.Reference) (string, bool) {
 	c.rwm.RLock()
 	defer c.rwm.RUnlock()
-	if !c.inadmissibleWorkloads.hasKey(wlKey) {
+	wlInfo := c.inadmissibleWorkloads.get(wl)
+	if wlInfo == nil {
 		return "", false
 	}
-	reason, ok := c.noFitSchedulingHashReasons[hash]
+	reason, ok := c.hashToBulkMoveReason[wlInfo.SchedulingHash]
 	return reason, ok
 }
 
@@ -512,6 +514,13 @@ func (c *ClusterQueue) DeleteFromLocalQueue(log logr.Logger, q *LocalQueue, role
 	reportCQFinishedWorkloads(c, roleTracker, cl)
 }
 
+func resolveQuotaReservedReason(statusReason string) string {
+	if statusReason == "" {
+		return kueue.WorkloadQuotaReservedReasonPendingEvaluation
+	}
+	return statusReason
+}
+
 // requeueIfNotPresent inserts a workload that cannot be admitted into
 // ClusterQueue, unless it is already in the queue. If immediate is true
 // or if there was a call to QueueInadmissibleWorkloads after a call to Pop,
@@ -565,11 +574,7 @@ func (c *ClusterQueue) requeueIfNotPresent(log logr.Logger, wInfo *workload.Info
 
 	if features.Enabled(features.SchedulingEquivalenceHashing) && wInfo.SchedulingHash != workload.SchedulingHashUnknown &&
 		(reason == RequeueReasonNoFit || reason == RequeueReasonPreemptionNoCandidates) {
-		quotaReservedReason := statusReason
-		if quotaReservedReason == "" {
-			quotaReservedReason = kueue.WorkloadQuotaReservedReasonPendingEvaluation
-		}
-		if moved := c.handleInadmissibleHash(wInfo.SchedulingHash, quotaReservedReason); moved > 0 {
+		if moved := c.handleInadmissibleHash(wInfo.SchedulingHash, resolveQuotaReservedReason(statusReason)); moved > 0 {
 			log.V(2).Info("Bulk-moved equivalent workloads to inadmissible", "hash", wInfo.SchedulingHash, "movedCount", moved)
 		}
 	}
@@ -591,7 +596,7 @@ func (c *ClusterQueue) handleInadmissibleHash(hash string, reason string) int {
 	if c.queueingStrategy != kueue.BestEffortFIFO {
 		return 0
 	}
-	c.noFitSchedulingHashReasons[hash] = reason
+	c.hashToBulkMoveReason[hash] = reason
 	moved := 0
 	for _, wInfo := range c.heap.List() {
 		if wInfo.SchedulingHash == hash {
--- a/pkg/cache/queue/cluster_queue.go
+++ b/pkg/cache/queue/cluster_queue.go
@@ -514,6 +514,12 @@ func (c *ClusterQueue) DeleteFromLocalQueue(log logr.Logger, q *LocalQueue, role
 	reportCQFinishedWorkloads(c, roleTracker, cl)
 }
 
+// resolveQuotaReservedReason returns statusReason if set, or defaults to
+// WorkloadQuotaReservedReasonPendingEvaluation when empty.
+// We do not feature-gate the fallback here because this function only computes
+// the reason string stored in the ClusterQueue's in-memory hashToBulkMoveReason map.
+// Whether status conditions are actually written to the Workload API object is
+// governed by the workload controller (shouldCheckEquivalenceHash).
 func resolveQuotaReservedReason(statusReason string) string {
 	if statusReason == "" {
 		return kueue.WorkloadQuotaReservedReasonPendingEvaluation
@@ -529,7 +535,7 @@ func resolveQuotaReservedReason(statusReason string) string {
 // When SchedulingEquivalenceHashing is enabled and the reason is NoFit or
 // PreemptionNoCandidates, equivalent workloads in the heap are bulk-moved
 // to inadmissible.
-func (c *ClusterQueue) requeueIfNotPresent(log logr.Logger, wInfo *workload.Info, immediate bool, reason RequeueReason, statusReason string) bool {
+func (c *ClusterQueue) requeueIfNotPresent(log logr.Logger, wInfo *workload.Info, immediate bool, reason RequeueReason, quotaReservedReason string) bool {
 	c.rwm.Lock()
 	defer c.rwm.Unlock()
 	key := workload.Key(wInfo.Obj)
@@ -574,7 +580,7 @@ func (c *ClusterQueue) requeueIfNotPresent(log logr.Logger, wInfo *workload.Info
 
 	if features.Enabled(features.SchedulingEquivalenceHashing) && wInfo.SchedulingHash != workload.SchedulingHashUnknown &&
 		(reason == RequeueReasonNoFit || reason == RequeueReasonPreemptionNoCandidates) {
-		if moved := c.handleInadmissibleHash(wInfo.SchedulingHash, resolveQuotaReservedReason(statusReason)); moved > 0 {
+		if moved := c.handleInadmissibleHash(wInfo.SchedulingHash, resolveQuotaReservedReason(quotaReservedReason)); moved > 0 {
 			log.V(2).Info("Bulk-moved equivalent workloads to inadmissible", "hash", wInfo.SchedulingHash, "movedCount", moved)
 		}
 	}
@@ -880,8 +886,9 @@ func (c *ClusterQueue) Active() bool {
 // in the latter case the implementation may choose to keep the workload in "inadmissible workloads"
 // where it doesn't compete with other workloads, until cluster events free up quota.
 // The workload should not be reinserted if it's already in the ClusterQueue.
+// quotaReservedReason represents the WorkloadQuotaReserved condition reason computed by the scheduler.
 // Returns true if the workload was inserted.
-func (c *ClusterQueue) RequeueIfNotPresent(ctx context.Context, wInfo *workload.Info, reason RequeueReason, statusReason string) bool {
+func (c *ClusterQueue) RequeueIfNotPresent(ctx context.Context, wInfo *workload.Info, reason RequeueReason, quotaReservedReason string) bool {
 	// when preemptions are in-progress, we keep attempting to
 	// schedule the same workload for BestEffortFIFO queues. See
 	// documentation of stickyWorkload for more details
@@ -902,7 +909,7 @@ func (c *ClusterQueue) RequeueIfNotPresent(ctx context.Context, wInfo *workload.
 			reason == RequeueReasonPendingMigration ||
 			reason == RequeueReasonPreemptionFailed
 	}
-	return c.requeueIfNotPresent(log, wInfo, immediate, reason, statusReason)
+	return c.requeueIfNotPresent(log, wInfo, immediate, reason, quotaReservedReason)
 }
 
 // baseCompareFunc orders workloads by sticky status, priority, timestamp, and UID.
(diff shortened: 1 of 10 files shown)


j-skiba:
/test all

run reproduction of the issue

j-skiba:
issue reproduced - https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12821/pull-kueue-test-integration-shard-2-main/2074116382693265408

j-skiba:
/test all

test the fix

j-skiba:
cc @mimowo @kshalot

pkg/cache/queue/manager.go
```diff
@@ -624,6 +624,27 @@ func (m *Manager) ClusterQueueForWorkloadWithoutLock(wl *kueue.Workload) (kueue.
...
+		return "", false
+	}
+	q := m.localQueues[qKey]
+	if q == nil {
+		return "", false
+	}
+	cq := m.hm.ClusterQueue(q.ClusterQueue)
+	if cq == nil {
+		return "", false
+	}
```
mimowo: I think we should already have some helper for this with proper logging - could you check available options? If there is nothing yet, let's extract some function like 

`ClusterQueueForWorkload`

pkg/cache/queue/cluster_queue.go
```diff
@@ -553,7 +565,11 @@ func (c *ClusterQueue) requeueIfNotPresent(log logr.Logger, wInfo *workload.Info
 
 	if features.Enabled(features.SchedulingEquivalenceHashing) && wInfo.SchedulingHash != workload.SchedulingHashUnknown &&
 		(reason == RequeueReasonNoFit || reason == RequeueReasonPreemptionNoCandidates) {
-		if moved := c.handleInadmissibleHash(wInfo.SchedulingHash); moved > 0 {
+		quotaReservedReason := statusReason
+		if quotaReservedReason == "" {
+			quotaReservedReason = kueue.WorkloadQuotaReservedReasonPendingEvaluation
```
mimowo: Hm, shouldn't WorkloadQuotaReservedReasonPendingEvaluation be guarded by the dedicated feature gate?
Please also delegate the computation to helper, like `resolveQuotaReservedReason`

pkg/cache/queue/cluster_queue.go
```diff
@@ -117,9 +117,9 @@ type ClusterQueue struct {
 	// inadmissibleWorkloads are workloads that have been tried at least once and couldn't be admitted.
 	inadmissibleWorkloads inadmissibleWorkloads
 
-	// noFitSchedulingHashes tracks scheduling equivalence classes that received
-	// NoFit or PreemptionNoCandidates. Cleared when queueInadmissibleWorkloads runs.
-	noFitSchedulingHashes sets.Set[string]
+	// noFitSchedulingHashReasons tracks scheduling equivalence classes and their reasons.
+	// Cleared when queueInadmissibleWorkloads runs.
+	noFitSchedulingHashReasons map[string]string
```
mimowo: Actually, this is not well named even currently. The name comes from the times we thought "noFit" is a good idea, but we later extended the behavior for the Preempt, noPreemptionCandidates scenario.

so I think we should name it differently, maybe `hashToBulkMoveReason`. Because this is not for any "requeue-reason", just the ones supporting "bulk move".

pkg/cache/queue/manager.go
```diff
@@ -624,6 +624,27 @@ func (m *Manager) ClusterQueueForWorkloadWithoutLock(wl *kueue.Workload) (kueue.
...
+	wInfo := workload.NewInfo(wl, m.workloadInfoOptions...)
+	reason, ok := cq.GetNoFitReason(wlKey, wInfo.SchedulingHash)
```
kshalot: In `cq.GetNoFitReason` we are already accessing `c.inadmissibleWorkloads` which is a map of wlInfos which already have the scheduling hash. Maybe we could change it's signature to accept just `workload.Reference` and extract the hash out of the `inadmissibleWorkloads`?

pkg/cache/queue/cluster_queue.go
kshalot: Small nit, feel free to ignore or contest - we could just reassign `statusReason` instead of defining a variable.

pkg/controller/core/workload_controller.go
```diff
@@ -1879,6 +1879,15 @@ func (r *WorkloadReconciler) resolveGranularUnadmittedQuotaReservedCondition(
 		return kueue.WorkloadQuotaReservedReasonMisconfigured, admissibilityErr.Error(), nil //nolint:nilerr // admissibility validation failure does not require retry
 	case workload.HasAdmissionGate(wl):
 		return kueue.WorkloadAdmissionGated, fmt.Sprintf("Admission is gated by: %s", wl.Annotations[constants.AdmissionGatedByAnnotation]), nil
+	}
+
+	if features.Enabled(features.SchedulingEquivalenceHashing) && isPendingSchedulingOrStale(cond) {
+		if reason, ok := r.queues.GetNoFitReason(wl); ok {
+			return reason, "Bypassed scheduling evaluation because an equivalent workload recently failed", nil
+		}
+	}
```
kshalot: Hm, but bulk moving to inadmissible based on the scheduling hash does not trigger any API calls, right? So in many cases, this reconcile will never be triggered or will only be triggered for a small subset of workloads that share a hash. Or am I missing something?

(Btw. I'm not saying that we should make an API call for each skipped workload, that would be a disaster most likely.)

j-skiba: bulk moving doesn't trigger any api calls.

Yes, in many cases that will not be triggered at all, it will be mostly triggered when NEW workloads are created that match equivalence hash - this way we will avoid concurrency between controllers that was observed in the flake.

Also, I updated the code to respect `UnadmittedWorkloadsObservabilityExplicitStatus` FG and extracted logic to `shouldCheckEquivalenceHash` separate method to improve readability

pkg/cache/queue/manager.go
j-skiba: 👍  yes, `ClusterQueueForWorkload` was already there, replaced

pkg/cache/queue/manager.go
j-skiba: yes, your idea looks cleaner. I refactored the cq method to
```go
func (c *ClusterQueue) GetNoFitReason(wl workload.Reference) (string, bool) {
```

pkg/cache/queue/cluster_queue.go
j-skiba: done

j-skiba: it doesn't matter here, values from the map are used conditionally in workload_controller that handles the observability FGs

pkg/cache/queue/cluster_queue.go
j-skiba: renamed to `hashToBulkMoveReason`

pkg/cache/queue/cluster_queue.go
j-skiba: also extracted computation to `resolveQuotaReservedReason`

j-skiba:
> @j-skiba: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> pull-kueue-test-integration-shard-2-main	[a186ddf](https://github.com/kubernetes-sigs/kueue/commit/a186ddf36146e406a52c64369feb822199098a36)	[link](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12821/pull-kueue-test-integration-shard-2-main/2074400575402807296)	true	`/test pull-kueue-test-integration-shard-2-main`
> [Full PR test history](https://prow.k8s.io/pr-history?org=kubernetes-sigs&repo=kueue&pr=12821). [Your PR dashboard](https://prow.k8s.io/pr?query=is%3Apr%20state%3Aopen%20author%3Aj-skiba). Please help us cut down on flakes by [linking to](https://git.k8s.io/community/contributors/devel/sig-testing/flaky-tests.md#filing-issues-for-flaky-tests) an [open issue](https://github.com/kubernetes-sigs/kueue/issues?q=is:issue+is:open) when you hit one in your PR.
> 
> Details

@coderabbitai create issue for this test flake

pkg/controller/core/workload_controller.go
kshalot: Fixing the flake aside, I'm wondering how "surprising" this behavior would be from the user's perspective. I guess it's nice that some workloads will get the granular reason, but I'm thinking whether it couldn't lead someone to draw a wrong conclusion by looking at the metric. Maybe this is not a big deal, but I'm raising it because I'm also curious what you think.

Btw. I recently proposed a `pending_scheduling_hashes` metric in https://github.com/kubernetes-sigs/kueue/issues/12463. This granular reason should probably be added there as well (or instead?), but it's still WIP.

j-skiba: > Fixing the flake aside, I'm wondering how "surprising" this behavior would be from the user's perspective. I guess it's nice that some workloads will get the granular reason, but I'm thinking whether it couldn't lead someone to draw a wrong conclusion by looking at the metric. Maybe this is not a big deal, but I'm raising it because I'm also curious what you think.

the bulk moved workloads are already kind of blind spot in the observability I would say and due to performance reasons we can't do a lot there but I think it's still worth to bring some insight into the state of kueue when possible.

In scenario when workloads are created one by one the I wouldn't expect any surprise, the workloads would be nice granular reasons.

but in case of mass admission of tens of workloads at one time possibly only one workload might be marked with some granular status (depending on explicitstatus fg) and the rest would be bulk moved (with or without `PendingEvaluation` reason depending on FG). In this case I agree there might be some confusion when someone starts seeing hundreds of workloads in `PendingEvaluation` state.

Given this maybe we should try counting these workloads in the new metrics (https://github.com/kubernetes-sigs/kueue/pull/12759)? During bulk move we will know what the underlying_reason is so maybe it would be ok(?). We won't have 1:1 mapping for these workloads but maybe it's something that is worth to do?

Alternatively maybe we could add a third label to make make them distinguishable to avoid missing 1:1 mapping confusion.

wdyt? cc @mimowo 

> Btw. I recently proposed a `pending_scheduling_hashes` metric in #12463. This granular reason should probably be added there as well (or instead?), but it's still WIP.

maybe or the idea above would be doable. Also maybe we should have some metric to show how many bulk moved workloads we have.

pkg/cache/queue/cluster_queue.go
```diff
@@ -500,6 +514,13 @@ func (c *ClusterQueue) DeleteFromLocalQueue(log logr.Logger, q *LocalQueue, role
 	reportCQFinishedWorkloads(c, roleTracker, cl)
 }
 
+func resolveQuotaReservedReason(statusReason string) string {
+	if statusReason == "" {
+		return kueue.WorkloadQuotaReservedReasonPendingEvaluation
```
mimowo: Shouldn't this branch be protected by the UnadmittedWorkloadsObservability FG?

j-skiba: I think it doesn't have to. It's the same as here https://github.com/kubernetes-sigs/kueue/pull/12821#discussion_r3534571089

mimowo: If no need, because for example when the FG is disabled then it is never called with "", then clarify that in the comment

pkg/cache/queue/manager.go
```diff
@@ -673,7 +689,7 @@ func (m *Manager) AddOrUpdateWorkloadWithoutLock(log logr.Logger, w *kueue.Workl
 // RequeueWorkload requeues the workload ensuring that the queue and the
 // workload still exist in the client cache and not admitted. It won't
 // requeue if the workload is already in the queue (possible if the workload was updated).
-func (m *Manager) RequeueWorkload(ctx context.Context, info *workload.Info, reason RequeueReason) bool {
+func (m *Manager) RequeueWorkload(ctx context.Context, info *workload.Info, reason RequeueReason, statusReason string) bool {
```
mimowo: I think the extra parameter requires a comment, plus I would like to consider strong typing as we do for `RequeueReason`, so that we can list all the states and describe their semantrics. Currently "statusReason" is unclear for me. Is this QuotaReservedReason maybe?

mimowo: Also, do we need to pass it as a separate parameter - couldn't this be already derived from the info (which contains the full workload) structure?

pkg/cache/queue/cluster_queue.go
j-skiba: added comment to the function

pkg/cache/queue/manager.go
j-skiba: renamed the argument to `quotaReservedReason`.

> Also, do we need to pass it as a separate parameter - couldn't this be already derived from the info (which contains the full workload) structure?

it can't be. At this point the reason is not updated yet.

pkg/cache/queue/cluster_queue.go
```diff
@@ -117,9 +117,10 @@ type ClusterQueue struct {
 	// inadmissibleWorkloads are workloads that have been tried at least once and couldn't be admitted.
 	inadmissibleWorkloads inadmissibleWorkloads
 
-	// noFitSchedulingHashes tracks scheduling equivalence classes that received
-	// NoFit or PreemptionNoCandidates. Cleared when queueInadmissibleWorkloads runs.
-	noFitSchedulingHashes sets.Set[string]
+	// hashToBulkMoveReason tracks scheduling equivalence classes and the reason
+	// why workloads with that hash were bulk-moved to inadmissibleWorkloads.
+	// Cleared when queueInadmissibleWorkloads runs.
+	hashToBulkMoveReason map[string]string
```
mimowo: Actually, maybe it should be a dedicated PR, but could we introduce strong typing for the keys and values? It would make reading easier.

pkg/cache/queue/manager.go
```diff
@@ -673,7 +689,10 @@ func (m *Manager) AddOrUpdateWorkloadWithoutLock(log logr.Logger, w *kueue.Workl
 // RequeueWorkload requeues the workload ensuring that the queue and the
 // workload still exist in the client cache and not admitted. It won't
 // requeue if the workload is already in the queue (possible if the workload was updated).
-func (m *Manager) RequeueWorkload(ctx context.Context, info *workload.Info, reason RequeueReason) bool {
+// The quotaReservedReason parameter represents the WorkloadQuotaReserved condition reason
+// computed by the scheduler during this cycle. It must be passed explicitly because info.Obj
+// has not yet been patched with the updated condition when RequeueWorkload is called.
+func (m *Manager) RequeueWorkload(ctx context.Context, info *workload.Info, reason RequeueReason, quotaReservedReason string) bool {
```
mimowo: What about the strong typing for `quotaReservedReason`, analogously to the `RequeueReason`?

pkg/cache/queue/manager.go
```diff
@@ -624,6 +624,22 @@ func (m *Manager) ClusterQueueForWorkloadWithoutLock(wl *kueue.Workload) (kueue.
...
+	m.RLock()
+	defer m.RUnlock()
+	cqName, ok := m.ClusterQueueForWorkloadWithoutLock(wl)
+	if !ok {
+		return "", false
+	}
+	cq := m.hm.ClusterQueue(cqName)
+	if cq == nil {
+		return "", false
+	}
```
mimowo: I would like to have this encapsulated into `ClusterQueueForWorkloadWithoutLock`. Maybe the other function should be renamed to `ClusterQueueNameForWorkloadWithoutLock`.

pkg/cache/queue/manager.go
j-skiba: maybe we could this and the other https://github.com/kubernetes-sigs/kueue/pull/12821#discussion_r3535416119 in separate PRs?

mimowo: Actually, since the reason is introduced in this PR, I would rather introduce the strong typing in this PR. 

For `hashToBulkMoveReason`, the same, maybe except for the keys which were pre-existing. The keys could be strong-typed either in prep or follow up.

pkg/cache/queue/manager.go
j-skiba: extracted and renamed the function

pkg/cache/queue/manager.go
j-skiba: introduced

```go
// QuotaReservedReason represents the reason for the WorkloadQuotaReserved condition
// computed by the scheduler or queue manager during evaluation.
type QuotaReservedReason string

const (
	QuotaReservedReasonGeneric                      QuotaReservedReason = ""
	QuotaReservedReasonPendingEvaluation            QuotaReservedReason = kueue.WorkloadQuotaReservedReasonPendingEvaluation
	QuotaReservedReasonWaitingForQuota              QuotaReservedReason = kueue.WorkloadQuotaReservedReasonWaitingForQuota
	QuotaReservedReasonWaitingForPreemptedWorkloads QuotaReservedReason = kueue.WorkloadQuotaReservedReasonWaitingForPreemptedWorkloads
	QuotaReservedReasonSuspended                    QuotaReservedReason = kueue.WorkloadQuotaReservedReasonSuspended
	QuotaReservedReasonMisconfigured                QuotaReservedReason = kueue.WorkloadQuotaReservedReasonMisconfigured
	QuotaReservedReasonAdmissionGated               QuotaReservedReason = kueue.WorkloadAdmissionGated
)
```

and made `hashToBulkMoveReason` use them

pkg/controller/core/workload_controller.go
mimowo: Let me first make sure I understand - the issue is that the PR creates a bit of inconsistency where the metric emits a granular reason, while the workloads have the QuotaReserved=False with the `PendingEvaluation` reason?

Let's confirm if we are on the same page wrt understading of the problem.

If this is the case, I think overall the approach is ok:
1. we update the metric using the granular reasons
2. we don't send API requests to update the bulk-moved workloads (that would not be feasible for large deployments)

We could mitigate it by a better documentation to PendingEvaluation  - that it could also be inadmissible, and some way of reading the reason from the evaluated workload. For example, if every workload had the "status.SchedulingHash" field value, then we could search quickly all workloads with the same hash, and show the last evaluated.

j-skiba: ### Scenario A
when workloads are created one by one the workloads are created with `PendingEvaluation` or some other reasons saved in `hashToBulkMoveReason` that matches workload's equivalence hash for `QuotaReserved` (if explicit status FG is enabled). 

As soon as they get that reason they will be counted by `kueue_unadmitted_workloads` with the underlying_reasons they received.

### Scenario B
The only problem might be when tens of workloads are created at once and they land in cluster queue when bulk move happens, then the workloads that were bulk moved will remain being counted as `PendingEvaluation` because that's what they are counted as when are reconciled for the first time. And theoretically they are `PendingEvaluation` (that's what Condition reason they have) but realistically they were bulk moved because of `WaitingForQuota` but we don't update their condition reasons due to performance.

No API calls are made in both scenarios.

--------------------

So I think that for Scenario A the current changes made in the PR make sense. We simply get more detailed information instead of leaving such workloads in `PendingEvaluation`

Scenario B is untouched by the changes in this PR. Here we can decide what's the approach we would like to have.

j-skiba: > For example, if every workload had the "status.SchedulingHash" field value, then we could search quickly all workloads with the same hash, and show the last evaluated.

that would be valuable but wouldn't we have to perform additional API calls to update this anyway?

mimowo: Not necessarily, we could set status.schedulingHash when setting the first "PendingEvaluation" condition, so I don't think we need any extra API call

mimowo: I was thinking mostly about scenario B. I think this is out of scope for the PR, but we may need to think about it. I think for now I would just update documentation for the PendingEvaluation reason to educate users about the interpretation.

mimowo (commented):
/lgtm
/approve
Thank you, this looks great 👍 Let's consider how to address the bulk-move separately.

PR #12824, opened 2026-07-06
Fix race condition in multikueue watcher establishment

### LINKED ISSUE #12801: MultiKueue flake: should keep multikueue admission check reference and properly clean up the remote workload when the connection to an admitting worker is lost
**Which test is flaking?**:
MultiKueue when an additional admission check covering only some flavors supported by the cluster queue is present should keep multikueue admission check reference and properly clean up the remote workload when the connection to an admitting worker is lost [area:multikueue, feature:multikueue,

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12780/pull-kueue-test-integration-multikueue-release-0-17/2074020247760801792

**Failure message or logs**:
```
{ failed [FAILED] Timed out after 90.001s.
Expected success, but got an error:
    <*errors.errorString | 0xc00c8b3c40>: 
    worker2 workload still exists and deletion has not started: uid=1a2320bc-6714-4845-9e03-7c6607340c62 finalizers=[] deletionTimestamp=<nil>
    {
        s: "worker2 workload still exists and deletion has not started: uid=1a2320bc-6714-4845-9e03-7c6607340c62 finalizers=[] deletionTimestamp=<nil>",
    }
In [It] at: /home/prow/go/src/sigs.k8s.io/kueue/test/integration/multikueue/jobs_test.go:2229 @ 07/06/26 06:46:46.638
}
```

Vaishnav88sk:
#### What type of PR is this?
/kind bug

When a worker cluster's connection is restored, `updateConfigAndRefreshWatchers` establishes new watchers via `startWatcher()`. This spawns a goroutine that immediately begins queuing `Added` events for remote workloads. Previously, `rc.connecting.Store(false)` was executed at the very end of the function. If the `wlReconciler` processed a queued workload event before `connecting=false` was set, it would incorrectly assume the cluster was still unavailable, skip the deletion of the orphaned remote workload, and requeue for 15 minutes, causing the test's `Eventually` block to time out.

This PR moves `rc.connecting.Store(false)` to execute immediately after setting the client, but before starting the watchers. This guarantees that when the `wlReconciler` processes the watch events, the cluster is correctly evaluated as available and remote workloads are cleaned up immediately.

Fixes #12801 

```release-note
MultiKueue: Fixed a bug where a remote could be marked connected too late, causing early workload events to be handled incorrectly (as if the remote was unreachable).
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved multi-cluster startup handling by marking the remote connection as connected immediately after switching clients, reducing races that could trigger missed or out-of-order events.
  * Added safer recovery when watcher initialization fails: the system now disconnects cleanly before backing off and retrying.
  * Treated adapter and queue watcher startup failures as fatal during setup to prevent watchers from continuing in a partially connected state.

### DIFF
--- a/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
+++ b/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
@@ -270,8 +270,15 @@ func (rc *remoteClient) updateConfigAndRefreshWatchers(watchCtx context.Context,
 
 	rc.setClient(remoteClient)
 
+	// Mark as connected before starting watchers to avoid a race condition.
+	// If a watcher queues an event and wlReconciler processes it before connecting=false,
+	// the workload will be treated as unavailable and requeued after 15m.
+	rc.connecting.Store(false)
+	rc.disconnected.Store(false)
+
 	err = rc.startWatcher(watchCtx, kueue.SchemeGroupVersion.WithKind("Workload").GroupKind().String(), &workloadKueueWatcher{})
 	if err != nil {
+		rc.disconnect()
 		return rc.increaseFailedConnAttempt(), err
 	}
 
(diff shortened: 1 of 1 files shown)


mimowo:
cc @olekzabl 
/area multikueue

olekzabl (approved):
Thank you!

/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17

olekzabl:
/release-note-edit

```release-note
MultiKueue: Fixed a bug where a remote could be marked connected too late, causing early workload events to be handled incorrectly (as if the remote was unreachable).
```

PR #11559, opened 2026-05-23
fix: fix MultiKueue workload requeue delay when worker connection is lost

yuluo-yx:
#### What type of PR is this?

/kind bug
/kind flake
/area testing

#### Which issue(s) this PR fixes:

Fixes #11496

#### Special notes for your reviewer:

```
cause: When a cluster connection is lost, `stopAndRemoveCluster()` completely removes the `remoteClient` from the map, causing the workload reconciler to be unable to detect the disconnected cluster, returning an `ErrNoActiveClusters` error and triggering the controller-runtime backoff instead of the `workerLostTimeout` delay mechanism.

Fix: Instead, mark disconnected clients as disconnected and retain them in the map, enabling the workload reconciler to:

	- Detect cluster disconnections
	- Apply the `workerLostTimeout` delay even if all clusters are disconnected
	- Properly handle cleanup when connections are restored
```

#### Does this PR introduce a user-facing change?

```release-note
MultiKueue: Fixed a bug where a lost connection to a worker cluster failed to trigger the "workerLostTimeout" delay mechanism for workload requeuing.
```

PR #12837, opened 2026-07-06
[MultiKueue] Simplify worker connection state to a single bool

olekzabl:
#### What type of PR is this?

/kind cleanup
/area multikueue

`disconnected atomic.Bool` was added in #11559  alongside `connecting atomic.Bool`, which IMO deserves a cleanup for two reasons:

- The overall semantics is confusing, at least for me. When a client is "disconnected", is it "connecting" or not?
  (Answer: It's considered "connecting". The combination of "disconnected, not connecting" does not exist).
- Both vars were typically used in concord; that alone suggested to turn them into one.

Actually, it seems a single `connected` bool can do.

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved worker cluster connectivity handling so quotas, reconciliation, and cleanup now only run when a cluster is fully connected.
  * Made reconnect behavior more reliable after disconnects or configuration updates.
  * Reduced cases where workloads or quotas could be treated as available before a remote cluster is actually ready.

### DIFF
--- a/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
+++ b/pkg/controller/admissionchecks/multikueue/multikueuecluster.go
@@ -129,8 +129,7 @@ type remoteClient struct {
 	origin       string
 	adapters     map[string]jobframework.MultiKueueAdapter
 
-	connecting           atomic.Bool
-	disconnected         atomic.Bool
+	connected            atomic.Bool
 	failedConnAttempts   uint
 	retryConnNextAttempt metav1.Time
 
@@ -166,7 +165,7 @@ func newRemoteClient(
 		adapters:     adapters,
 		clock:        clock.RealClock{},
 	}
-	rc.connecting.Store(true)
+	rc.connected.Store(false)
 	return rc
 }
 
@@ -238,9 +237,8 @@ func (rc *remoteClient) increaseFailedConnAttempt() *time.Duration {
 // If the encountered error is not permanent the duration after which a retry should be done is returned.
 func (rc *remoteClient) updateConfigAndRefreshWatchers(watchCtx context.Context, config *clientConfig) (*time.Duration, error) {
 	configChanged := !equality.Semantic.DeepEqual(config, rc.config)
-	connecting := rc.connecting.Load()
-	disconnected := rc.disconnected.Load()
-	if !configChanged && !connecting && !disconnected {
+	connected := rc.connected.Load()
+	if !configChanged && connected {
 		return nil, nil
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/util/atomic/value.go
```diff
@@ -0,0 +1,45 @@
...
+*/
+
+package atomic
+
+import (
+	"sync/atomic"
+)
+
+type Value[T any] struct {
+	v atomic.Value
```
kshalot: I think we'll have to use `atomic.Pointer` or smth. Because this can explode on runtime type checks:
https://go.dev/play/p/UC_4d_44IYp

^ in that example, if you comment one of the `err.Store` lines, the code will work. But if they are both uncommented, it will compile (because both errors implement `error`), but panic in runtime because their concrete type is different.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
```diff
@@ -117,6 +117,27 @@ func (c *clientConfig) toRESTConfig() (*rest.Config, error) {
...
 
+type clientState int32
+
+const (
+	clientConnecting clientState = iota
+	clientConnected
+	clientDisconnected
+)
+
+func (s clientState) String() string {
```
kshalot: We can also use string constants instead of ints.

pkg/util/atomic/value.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func (a *Value[T]) Load() T {
+	if val := a.v.Load(); val != nil {
+		return val.(T)
+	}
+	var zero T
+	return zero
+}
```
Singularity23x0: Nit: something like this should work as well.

```suggestion
func (a *Value[T]) Load() (t T) {
	if val := a.v.Load(); val != nil {
		t = val.(T)
	}
	return
}
```

Singularity23x0: Similarly for Swap.

pkg/util/atomic/value.go
olekzabl: Obsoleted by simplification.

pkg/controller/admissionchecks/multikueue/multikueuecluster.go
olekzabl: Obsoleted by simplification.

pkg/util/atomic/value.go
olekzabl: Obsoleted by simplification.

mimowo:
/lgtm
Looks great, thank you for keeping the code tidy 👍

mimowo:
@coderabbitai ignore

PR #12838, opened 2026-07-07
Reject negative resource requests and limits in Workloads

vladikkuzn:
A Workload PodSet template embeds a PodSpec that bypasses core Pod validation, so the validating webhook previously accepted negative container resource requests/limits (e.g. cpu: "-100"). Negative values propagated unclamped into ClusterQueue usage accounting, deflating tracked usage and inflating available quota.

Reject negative requests and limits in the Workload validating webhook (reusing validateResourceQuantity), and floor negatives to zero at the request-computation chokepoint in totalRequestsFromPodSets as defense-in-depth for any pre-existing objects.

#### What type of PR is this?
/kind cleanup

Fixes #

```release-note
Scheduling: Fixed a bug where negative container resource requests or limits
could create artificial ClusterQueue quota credit, allowing Workloads to bypass
configured quota limits. Kueue now floors negative values to zero during quota
accounting and rejects them during Workload validation by default. The
validation is controlled by the Beta
`WorkloadValidateResourcesAreNonNegative` feature gate.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workload admission validation to reject the reserved `pods` resource key and to enforce negative-value rules for both **requests** and **limits** across containers and init containers.
  * Total request calculation now clamps negative resource requests to **0**, preventing incorrect quota computations.

* **Tests**
  * Added unit tests for clamping negative requests to zero.
  * Expanded webhook/workload/integration test coverage for negative CPU/memory scenarios and reserved `pods` handling.

### DIFF
--- a/pkg/features/kube_features.go
+++ b/pkg/features/kube_features.go
@@ -529,6 +529,11 @@ const (
 	// VectorizedResourceRequests enables slice-based indexing for resource requests in TAS snapshots,
 	// replacing map lookups for higher performance during scheduling and preemption.
 	VectorizedResourceRequests featuregate.Feature = "VectorizedResourceRequests"
+
+	// owner: @vladikkuzn
+	//
+	// Rejects Workloads with negative container or pod-level resource requests/limits.
+	WorkloadValidateResourcesAreNonNegative featuregate.Feature = "WorkloadValidateResourcesAreNonNegative"
 )
 
 func init() {
@@ -812,7 +817,12 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 	SchedulerLibraryIntegration: {
 		{Version: version.MustParse("0.19"), Default: false, PreRelease: featuregate.Alpha},
 	},
+
 	VectorizedResourceRequests: {
+		{Version: version.MustParse("0.19"), Default: true, PreRelease: featuregate.Beta},
+	},
+
+	WorkloadValidateResourcesAreNonNegative: {
 		{Version: version.MustParse("0.19"), Default: true, PreRelease: featuregate.Beta},
 	},
 }
--- a/pkg/features/kube_features.go
+++ b/pkg/features/kube_features.go
@@ -823,7 +823,7 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 	},
 
 	WorkloadValidateResourcesAreNonNegative: {
-		{Version: version.MustParse("0.19"), Default: true, PreRelease: featuregate.Beta},
+		{Version: version.MustParse("0.20"), Default: true, PreRelease: featuregate.Beta},
 	},
 }
 
--- a/pkg/resources/requests.go
+++ b/pkg/resources/requests.go
@@ -100,6 +100,16 @@ func (r MapRequests) IsEmpty() bool {
 	return len(r) == 0
 }
 
+// FloorToZero replaces negative resource values with zero.
+// Defense-in-depth for pre-existing negative Workloads while
+// WorkloadValidateResourcesAreNonNegative rolls out.
+// TODO: remove ~2 releases after WorkloadValidateResourcesAreNonNegative locks to GA.
+func (r MapRequests) FloorToZero() {
+	for k, v := range r {
+		r[k] = max(v, 0)
+	}
+}
+
 func (r MapRequests) Add(other Requests) {
 	other.ForEach(func(k corev1.ResourceName, v int64) {
 		r[k] += v
--- a/pkg/webhooks/workload_webhook.go
+++ b/pkg/webhooks/workload_webhook.go
@@ -179,16 +179,31 @@ func validatePodSet(ps *kueue.PodSet, path *field.Path) field.ErrorList {
 	for ci := range ps.Template.Spec.Containers {
 		allErrs = append(allErrs, validateContainer(&ps.Template.Spec.Containers[ci], cPath.Index(ci))...)
 	}
+	// validate pod-level resources
+	if ps.Template.Spec.Resources != nil {
+		resPath := path.Child("template", "spec", "resources")
+		allErrs = append(allErrs, validateResourceList(ps.Template.Spec.Resources.Requests, resPath.Child("requests"))...)
+		allErrs = append(allErrs, validateResourceList(ps.Template.Spec.Resources.Limits, resPath.Child("limits"))...)
+	}
 
 	return allErrs
 }
 
 func validateContainer(c *corev1.Container, path *field.Path) field.ErrorList {
+	requestErrors := validateResourceList(c.Resources.Requests, path.Child("resources", "requests"))
+	limitErrors := validateResourceList(c.Resources.Limits, path.Child("resources", "limits"))
+	return append(requestErrors, limitErrors...)
+}
+
+// validateResourceList rejects the reserved pods key and, when enabled, negative quantities.
+func validateResourceList(resources corev1.ResourceList, path *field.Path) field.ErrorList {
 	var allErrs field.ErrorList
-	rPath := path.Child("resources", "requests")
-	for name := range c.Resources.Requests {
+	for name, quantity := range resources {
 		if name == corev1.ResourcePods {
-			allErrs = append(allErrs, field.Invalid(rPath.Key(string(name)), corev1.ResourcePods, "the key is reserved for internal kueue use"))
+			allErrs = append(allErrs, field.Invalid(path.Key(string(name)), corev1.ResourcePods, "the key is reserved for internal kueue use"))
+		}
+		if features.Enabled(features.WorkloadValidateResourcesAreNonNegative) {
+			allErrs = append(allErrs, validateResourceQuantity(quantity, path.Key(string(name)))...)
 		}
 	}
 	return allErrs
--- a/pkg/webhooks/workload_webhook_test.go
+++ b/pkg/webhooks/workload_webhook_test.go
@@ -23,6 +23,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/component-base/featuregate"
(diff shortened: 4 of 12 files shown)


dkaluza (approved):
Is this ready? Or still work in progress?

Generally looks good to me

vladikkuzn:
> Is this ready? Or still work in progress?

@dkaluza It's now

pkg/resources/requests.go
```diff
@@ -81,6 +81,17 @@ func (r Requests) Mul(f int64) {
...
 }
 
+// FloorToZero replaces any negative resource value with zero. Requests are never
+// legitimately negative; this guards quota accounting against malformed PodSet
+// templates, which bypass core Pod validation. See webhook validateContainer.
+func (r Requests) FloorToZero() {
+	for k, v := range r {
+		if v < 0 {
+			r[k] = 0
+		}
```
mszadkow: r[k] = max(v, 0) could be a little bit cleaner

pkg/webhooks/workload_webhook_test.go
```diff
@@ -105,6 +106,64 @@ func TestValidateWorkload(t *testing.T) {
...
+				PodSets(
+					*utiltestingapi.MakePodSet("ok", 1).
+						Containers(utiltesting.SingleContainerForRequest(map[corev1.ResourceName]string{
+							corev1.ResourceCPU: "0",
+						})...).
+						Obj(),
+				).
+				Obj(),
+			wantErr: nil,
+		},
```
mszadkow, line 257: nit: maybe it's worth to see more extended example, like multiple resource request/limits which some are negative

dkaluza:
Can we add a release note? This is a hotfix, but additional validation may still be disruptive for some users if they have negative resource requests right now.

pkg/resources/requests.go
vladikkuzn: Done in 4fa67cd91 — switched to `r[k] = max(v, 0)`.

pkg/webhooks/workload_webhook_test.go
vladikkuzn: Added a case with multiple resources where only some are negative (request + limit) in 4fa67cd91.

vladikkuzn:
Added a release note describing the behavior change, with `ACTION REQUIRED` since the new validation can reject Workloads that currently specify negative resource requests/limits.

dkaluza (approved):

dkaluza:
@mimowo this one looks good to me when you have a moment

pkg/webhooks/workload_webhook.go
```diff
@@ -169,10 +169,15 @@ func validatePodSet(ps *kueue.PodSet, path *field.Path) field.ErrorList {
...
-	for name := range c.Resources.Requests {
+	for name, quantity := range c.Resources.Requests {
 		if name == corev1.ResourcePods {
 			allErrs = append(allErrs, field.Invalid(rPath.Key(string(name)), corev1.ResourcePods, "the key is reserved for internal kueue use"))
 		}
+		allErrs = append(allErrs, validateResourceQuantity(quantity, rPath.Key(string(name)))...)
+	}
+	lPath := path.Child("resources", "limits")
+	for name, quantity := range c.Resources.Limits {
+		allErrs = append(allErrs, validateResourceQuantity(quantity, lPath.Key(string(name)))...)
```
mimowo: Should we also commonize the check with `if name == corev1.ResourcePods {`? Maybe we can push the check inside `validateResourceQuantity`, wdyt?

dkaluza: Hmm, I wouldn't push this to `validateResourceQuantity` this would change its semantic and it is used also in other places.

From my point of view it would also a little bit violate single responsibility principle of this function - as it should just validate the quantity according to the name.

But IMO we can create another helper like `validateResourceList`, then this function would more or less look like:
```
requestErrors := validateResourceList(c.Resources.Requests, path.Child("resources", "requests"))
limitErrors := validateResourceList(c.Resources.Limits, path.Child("resources", "limits"))

return append(requestErrors, limitErrors...)
```

Where `validateResourceList` will loop through the resources, check the name against `corev1.ResourcePods` and call validate  `validateResourceQuantity`. Wdyt?

mimowo: > Hmm, I wouldn't push this to validateResourceQuantity this would change its semantic and it is used also in other places.

The main question - what is the correct semantic. Not checking `corev1.ResourcePods` for limits, if we check it for "resources" feels buggy,  We may de-couple that indeed, leaving "resources" and "limits" inconsistent smells like a bug :).

dkaluza: I do not argue with checking it - I fully agree that we should if we are checking this in requests :)
In fact if we introduce a helper like the one that I suggested above we can be pretty sure the validation for both will be the same. Just didn't spot it in my initial pass as it is not main topic of the PR

It looks like we are using quantity validation in cluster queues. I guess it generally makes sense to not add the pods nam validation there, but I lack context to be sure
https://github.com/kubernetes-sigs/kueue/blob/b7a968c2211780d6e43aee7a148fa293fbf4d21f/pkg/webhooks/clusterqueue_webhook.go#L342

mimowo: Oh, I see, your point was that  validateResourceQuantity is also reused in other places, in CQ validation specifically. It is a valid use case for users to specifiy quota for CQs for "pods", so indeed we cannot fold the check into validateResourceQuantity. I think your proposal makes sense 👍

mimowo (commented):
/lgtm
/approve
Thank you for fixing the validation 👍 
/cherrypick release-0.18
/cherrypick release-0.17

mimowo:
/hold
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12838/pull-kueue-test-unit-shard-0-main/2079113128603291648
This is likely related ptal

pkg/webhooks/workload_webhook.go
```diff
@@ -184,12 +184,21 @@ func validatePodSet(ps *kueue.PodSet, path *field.Path) field.ErrorList {
 }
 
 func validateContainer(c *corev1.Container, path *field.Path) field.ErrorList {
+	requestErrors := validateResourceList(c.Resources.Requests, path.Child("resources", "requests"))
+	limitErrors := validateResourceList(c.Resources.Limits, path.Child("resources", "limits"))
+	return append(requestErrors, limitErrors...)
```
mimowo, line 195: Hm, one about validating also Pod-level resources?

pkg/resources/requests.go
```diff
@@ -96,6 +96,15 @@ func (r Requests) Mul(f int64) {
 	}
 }
 
+// FloorToZero replaces any negative resource value with zero. Requests are never
+// legitimately negative; this guards quota accounting against malformed PodSet
+// templates, which bypass core Pod validation. See webhook validateContainer.
+func (r Requests) FloorToZero() {
```
mimowo: Why is this needed at all if we will have validation now?

pkg/webhooks/workload_webhook.go
mimowo, line 195: I think this is ok, because users cannot use negative values in k8s, except for hihacking quota.

Because of the validation in k8s: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/v1/validation/validation.go#L50-L58

However, let's add a FG (in Beta) `WorkloadValidateResourcesAreNonNegative`

pkg/resources/requests.go
dkaluza: Generally from my perspective this one is still useful for the transition period - what if there alreadu exists some workload with negatives requests in the cluster? It is even more valuable if we are going to FG the validation.

mimowo: > what if there alreadu exists some workload with negatives requests in the cluster? 

In that case Kueue will start rejecting any Workload object updates. So, basically admin will need to delete it manually. With this code we will collapse the values to "0". 

> It is even more valuable if we are going to FG the validation.

I don't have a strong view here. The code is very simple so we can probably afford it.

dkaluza: > In that case Kueue will start rejecting any Workload object updates. So, basically admin will need to delete it manually. With this code we will collapse the values to "0".

True, but without it it will still give a user additional quota, wouldn't it?

mimowo: Yes, until the workload is deleted. So, yeah, I'm ok with zero-ing. It does not cost much complexity. 

Though, we could probably remove the code something like 2 releases after the FG for validation is locked. So that we don't maintain the code indefinetly.

vladikkuzn: Keeping `FloorToZero` for the transition (pre-existing negative workloads + defense while the FG rolls out). Added a TODO in 97927a0a69a to remove it ~2 releases after `WorkloadValidateResourcesAreNonNegative` is locked to GA.

pkg/webhooks/workload_webhook.go
vladikkuzn: Done in 97927a0a69a — also validate pod-level `spec.resources` requests/limits in `validatePodSet`.

vladikkuzn: Done in 97927a0a69a — added Beta FG `WorkloadValidateResourcesAreNonNegative` (default enabled in 0.19) and gated the non-negative quantity checks behind it.

mimowo (commented):
/lgtm
/approve
/unhold
Thank you @vladikkuzn , please prepare CPs

pkg/workload/workload.go
mimowo, line 631: Ups, we should be using the factory here, could you follow up @j-skiba ?

pkg/features/kube_features.go
```diff
@@ -812,9 +817,14 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 	SchedulerLibraryIntegration: {
 		{Version: version.MustParse("0.19"), Default: false, PreRelease: featuregate.Alpha},
 	},
+
 	VectorizedResourceRequests: {
 		{Version: version.MustParse("0.19"), Default: true, PreRelease: featuregate.Beta},
 	},
+
+	WorkloadValidateResourcesAreNonNegative: {
+		{Version: version.MustParse("0.19"), Default: true, PreRelease: featuregate.Beta},
```
```suggestion
		{Version: version.MustParse("0.20"), Default: true, PreRelease: featuregate.Beta},
```
Please update the PR as we already relased 0.19

vladikkuzn: Done — bumped `WorkloadValidateResourcesAreNonNegative` intro version to 0.20 in `kube_features.go` and both feature-list YAMLs.

pkg/workload/workload.go
vladikkuzn: Ack — `PodSetResources.Requests` is still `MapRequests`, so this call site stays on `NewMapRequests`. Migrating it to the `Requests` factory (`NewRequestsFrom*`) belongs with the VectorizedResourceRequests follow-up for @j-skiba; out of scope for this PR.

mimowo:
I checked the k8s rejects pods with negative resources, so such a workload would not run anyway.

So, the only use is malicious:  the user creates two workloads:
1. A with negative requests - it will not run anyway due to k8s validation, but it inflates the available quota
2. B with positive requests which normally wouldn't fit, but can fit due to inflating quota by the workload A

Since this is malicious use I'm not sure we should be saying about ACTION REQUIRED. At least the current action required makes it sound like a regular use-case. 

Any opinion @tenzen-y ?

/release-note-edit
```release-note
Scheduling: Fixed a bug where negative container resource requests or limits
could create artificial ClusterQueue quota credit, allowing Workloads to bypass
configured quota limits. Kueue now floors negative values to zero during quota
accounting and rejects them during Workload validation by default. The
validation is controlled by the Beta
`WorkloadValidateResourcesAreNonNegative` feature gate.
```

mimowo:
/lgtm
/approve
Thank you, please prepare the cherrypicks for 0.19 and 0.18

PR #12840, opened 2026-07-07
Fix the flaky perf tas test

mimowo:
#### What type of PR is this?
/kind bug
/kind flake

We have the perf tas test flaking which is probably due to new stronger machines rather than a code change.

https://testgrid.k8s.io/sig-scheduling#periodic-kueue-test-tas-scheduling-perf-main

image

from https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/periodic-kueue-test-tas-scheduling-perf-main/2073757754912149504
```
{Failed  === RUN   TestScalability/ClusterQueueClasses/cq
    checker_test.go:98: Usage 57.73% is less then expected 58.00%
--- FAIL: TestScalability/ClusterQueueClasses/cq (0.00s)
}
```
and from https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/periodic-kueue-test-tas-scheduling-perf-main/2073576302073352192
```
{Failed  === RUN   TestScalability/ClusterQueueClasses/cq
    checker_test.go:98: Usage 57.83% is less then expected 58.00%
--- FAIL: TestScalability/ClusterQueueClasses/cq (0.00s)
}
```

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted a performance test threshold to reduce flaky failures in scheduler runs.
  * Updated the related note to reflect the lowered usage limit.

### DIFF
--- a/test/performance/scheduler/configs/tas/rangespec.yaml
+++ b/test/performance/scheduler/configs/tas/rangespec.yaml
@@ -17,7 +17,7 @@ cmd:
 # Thresholds: mean − 7%, rounded down to 0.1%
 # Class  min usage%    mean / stddev / CV%
 clusterQueueClassesMinUsage:
-  cq:  58.0          # 62.41 / 0.30 / 0.48%
+  cq:  57.0          # 62.41 / 0.30 / 0.48% (reduced due to flakes)
 
 # Workload admission time limits
 # Thresholds: mean + 20% rounded up to 1000ms


mimowo:
cc @mbobrovskyi

test/performance/scheduler/configs/tas/rangespec.yaml
```diff
@@ -17,7 +17,7 @@ cmd:
 # Thresholds: mean − 7%, rounded down to 0.1%
 # Class  min usage%    mean / stddev / CV%
 clusterQueueClassesMinUsage:
-  cq:  58.0          # 62.41 / 0.30 / 0.48%
+  cq:  57.0          # 62.41 / 0.30 / 0.48% (reduced due to flakes)
```
mbobrovskyi, line 20: Do you know what does it mean this comment (`# 62.41 / 0.30 / 0.48%`)?

mimowo: 62.41 is probably the mean, 0.30 is probably the std dev, 0.48% - I don't know (maybe how much it was reduced from the mean initially).

mimowo: I think it would be good to write / generate a script that would set those values automatically based on historical data. I always wanted to have a script which would fit the distrubutions to mean and std dev, and compute the thresold according to 0.001 percentile (or so).

mimowo: Opened: https://github.com/kubernetes-sigs/kueue/issues/12841

kshalot: > 0.48% - I don't know

It's the [coefficient of variation](https://en.wikipedia.org/wiki/Coefficient_of_variation).

The data fits: `0.3/62.41 * 100 = 0.48`

kshalot: Oh, it's documented above btw:
```yaml
# Thresholds: mean − 7%, rounded down to 0.1%
# Class  min usage%    mean / stddev / CV%           <- HERE
clusterQueueClassesMinUsage:
```

mimowo: @mbobrovskyi are you ok with the current diff? Indeed I think the numbers may need refreshing, but without a script it is very tedious. So, I would like to defer it until we have https://github.com/kubernetes-sigs/kueue/issues/12841, just leaving the free-form comment for now.

All the numbers need refreshing actually because the machines which run our CI changed on 1st of July.

mbobrovskyi: Yeah, I'm good with this changes. Thank you!

/lgtm

Issue #12763, opened 2026-07-03
☂️ The release team and release process automation


mimowo:
**What would you like to be added**:

We are starting the idea of the "release team" in Kueue. The team will perform the releases caring about the cadence and the process.

Currently we execute the scripted release process steps, example. https://github.com/kubernetes-sigs/kueue/issues/12427

The process is semi-automated, but still some actions are manual and require **write access** to the repo - for example creating the release tags. Part of the challenge is to automate such steps so that the release team can perform releases autonomously. This is ticketed here: https://github.com/kubernetes-sigs/kueue/issues/12316

We will start with shadowing sessions so that the release team learns what exactly are the steps and challenges. The expectation in the long term is that the release team also maintains the process.

**Why is this needed**:

To offload the top level maintainers, similarly as this is done in the core k8s. The process itself takes around 3h for patches and 5h for full releases, so this will be a lot of time saving.

mimowo:
cc @mbobrovskyi @Singularity23x0 who we choose for the release team 

cc @tenzen-y @amy @kannon92 @mwysokin @sohankunkerkar

mimowo:
cc @olekzabl

amy:
🎉

Singularity23x0:
Release issue template - pasted here for ease of reference.

## Release Checklist

- [ ] [OWNERS](https://github.com/kubernetes-sigs/kueue/blob/main/OWNERS) must LGTM the release proposal.
  At least two for minor or major releases. At least one for a patch release.
- [ ] Verify that the changelog in this issue and the CHANGELOG folder is up-to-date
  - [ ] Run `./hack/releasing/sync-notes.sh $VERSION` to generate and publish the release notes
- [ ] For major or minor releases (v$MAJ.$MIN.0), create a new release branch.
  - [ ] An OWNER creates a vanilla release branch with
        `git branch release-$MAJ.$MIN main`
  - [ ] An OWNER pushes the new release branch with
        `git push upstream release-$MAJ.$MIN`
- [ ] Update the release branch:
  - [ ] Ensure there are no unstaged changes in your directory (the script adds everything)
  - [ ] Run `./hack/releasing/prepare_pull.sh --target release $VERSION`
  - [ ] Wait for this PR to merge  
- [ ] An OWNER creates a signed tag
  - [ ] pull the release branch after PR from previous step merged
  - [ ] run
     `git tag -s $VERSION`
      and inserts the changelog into the tag description.
[...]

PR #12848, opened 2026-07-07
Release team and release automation - sync release notes step.

Singularity23x0:
#### What type of PR is this?
/kind documentation

A proposal on how the release team could be given the capabilities of independently creating and managing Kueue releases via ChatOps.

Currently covers the step of creating and syncing release notes for the release issue.

Contributes to #12763

Generated with the use of AI.

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added an automated issue-comment command to sync release notes for a specified version.
  * The command now strictly validates the comment format and the expected release title pattern before running.
  * Release note syncing is gated by a maintained allowlist configured in the repository.
  * Improved end-to-end CI automation for syncing release notes with authenticated repository access and non-interactive execution.


mimowo:
Cc @mbobrovskyi ptal

tenzen-y (commented):
Thank you for working on this. 
Can you check if this workflow works well in your forked repository?

release-team.json
```diff
@@ -0,0 +1,6 @@
+[
+  "mimowo",
+  "tenzen-y",
+  "mbobrovskyi",
+  "Singularity23x0"
+]
```
mbobrovskyi:
```suggestion
]

```
nit: new line

.github/workflows/release-utils.yml
```diff
@@ -0,0 +1,66 @@
...
+          go install k8s.io/release/cmd/release-notes@latest
+          echo "$(go env GOPATH)/bin" >> $GITHUB_PATH
+
+      - name: Run Sync Notes
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          UPSTREAM_REMOTE: origin
+        run: |
+          BODY="${{ github.event.comment.body }}"
+          if [[ "$BODY" =~ ^/sync-release-notes[[:space:]]*$ ]]; then
```
mbobrovskyi: I don't think this is necessary. We already skip this Job without this comment.

.github/workflows/release-utils.yml
```diff
@@ -0,0 +1,66 @@
...
+            TITLE="${{ github.event.issue.title }}"
+            if [[ "$TITLE" =~ ^Release\ (v[0-9]+\.[0-9]+\.[0-9]+)$ ]]; then
```
mbobrovskyi: I think we should skip the job if the issue doesn't have a title. So we should probably move this condition to line 10.

mbobrovskyi:
I tested it in my local fork: https://github.com/mbobrovskyi/kueue/actions/runs/28921710503/job/85800184580. The CI is running as expected. I only noticed one issue:

```
Run BODY="/sync-release-notes"
VERSION=v0.18.3
github.com
  ✓ Logged in to github.com account github-actions[bot] (GH_TOKEN)
  - Active account: true
  - Git operations protocol: https
  - Token: ghs_15***F_wkhikm5jlD_jjlqXVMPx0wBurYZL43WPYDjz6RicniMkuMCuHp1AlkqvESlmXtN_*****
+++ HEAD_BRANCH=release-0.18
+++ PREVIOUS_VERSION=v0.18.0
+++ Running release-notes:
release-notes --org "mbobrovskyi" \
              --repo "kueue" \
              --branch "release-0.18" \
              --start-sha "f26d35cb3d8a39967839909c66899c86399ce226" \
              --end-sha "f26d35cb3d8a39967839909c66899c86399ce226" \
              --dependencies=false \
              --required-author="" 
              --output="/tmp/tmp.Eii8Pso6Jq" 
level=warning msg="No subcommand specified, running \"generate\" "
level=fatal msg="unknown flag: --required-author"
yes: standard output: Broken pipe
Error: Process completed with exit code 1.
```

I think it happens because we use `k8s.io/release/cmd/release-notes@latest`. Maybe we should pin it to a specific version?

mimowo:
> I think it happens because we use k8s.io/release/cmd/release-notes@latest. Maybe we should pin it to a specific version?

I don't have an opinion here actually. The release-notes script is released rarely (I think on-demand), so using the "latest" could make sense in this case, or we need to ping more often the maintainers for releases

However, one worry with "latest" is supply chain injection, because IIUC for fixed versions we could pin them to hash, but for latest pinning to hash makes it not useful anyway.

release-team.json
kshalot, line 1: Should this file be in the root of the repository? To me it belongs in `hack/releasing` for example

.github/workflows/release-utils.yml
Singularity23x0: From what I understand this is to ensure we trigger only if the comment matches exactly to "/sync-release-notes" with the possibility of adding trailing whitespaces.

The initial "if" only checks if the comment starts with that phrase. We do not want to do an exact match there, so that the command does not fail silently when running "/sync-release-notes v0.18.3", but rather explicitly states that the command should be ran without args, and what the commenter is trying to do may not be what the command will do.

.github/workflows/release-utils.yml
```diff
@@ -0,0 +1,66 @@
...
+        run: |
+          go install k8s.io/release/cmd/release-notes@latest
+          echo "$(go env GOPATH)/bin" >> $GITHUB_PATH
+
+      - name: Run Sync Notes
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          UPSTREAM_REMOTE: origin
+        run: |
+          BODY="${{ github.event.comment.body }}"
```
kshalot: I'm not 100% sure how this gets interpolated, but I think this might be very dangerous.

What if I set the comment to something like:
```sh
/sync-release-notes"; hack_kueue; #
```
then wouldn't this be interpolated into:
```sh
BODY="/sync-release-notes"; hack_kueue; #"
```

IIUC right now this can only be triggered by the members of the release team, not arbitrary users. Nevertheless we shouldn't allow for ACE.

But maybe I'm paranoid and this is not exploitable. I'd appreciate if you checked though

kshalot: There's also this check for prefix... So maybe the area to exploit is small. But it smelled to me so I'm flagging it

.github/workflows/release-utils.yml
Singularity23x0: From what I understand, the if statement in line 10 does not support regex, so we would have to simplify it.
Also, failing the "if" statement causes the command to fail silently. I would prefer the command to notify us itf there is a problem with the issue name rather than just do nothing.

Singularity23x0: As an alternative - i proposed extracting this part of the logic to be a separate, earlier check, done right after authorizing the actor.

release-team.json
Singularity23x0: The scripting had to be placed in the `.github/workflows/` dir fr ChatOps to work.

As for the release-team file - it is not a hack script but rather a field documenting the members of the release team - I think it makes more sense to be placed in the root, akin to how OWNERS files are handled.

If the format of the file is misleading in that case, we can change it. The file is technically a configuration file for the repo however, and I don't see it fitting natrually in the hack folder.

.github/workflows/release-utils.yml
Singularity23x0: Good catch, adjusted the logic to put the command body into an env variable.

Singularity23x0:
Can confirm the updated version works and updates the issue correctly.

.github/workflows/release-utils.yml
```diff
@@ -0,0 +1,72 @@
...
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+
+      - name: Install release-notes tool
+        run: |
+          go install k8s.io/release/cmd/release-notes@v0.17.0
```
mimowo: Isn't the latest release 0.21.1? https://github.com/kubernetes/release/releases/tag/v0.21.1

.github/workflows/release-utils.yml
```diff
@@ -0,0 +1,72 @@
...
+          if [[ ! "$TITLE" =~ ^Release\ (v[0-9]+\.[0-9]+\.[0-9]+)$ ]]; then
+            echo "Error: Issue title '$TITLE' does not match 'Release vX.Y.Z'"
+            exit 1
+          fi
+
+          VERSION="${BASH_REMATCH[1]}"
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
```
mimowo: pin all versions to hashes to prevent delivery chain attack, as done here: https://github.com/kubernetes-sigs/kueue/blob/45d895f5d5a3b848eb08ce56e8b423b03434bc44/.github/workflows/sbom.yaml#L24

Singularity23x0: Updated version and pinned.

.github/workflows/release-utils.yml
Singularity23x0: Updated to 0.20.1. Using the latest version requires adjusting our script, as the `--required-author` flag is no longer supported as of the next release after, specifically 0.21.0.

mimowo:
Leaving lgtm to @mbobrovskyi or @tenzen-y

mbobrovskyi (commented):
Thank you!

/lgtm

PR #12849, opened 2026-07-07
Defer workload finalization until owner job is gone.

### LINKED ISSUE #12820: The workload considered as orphan, while the job is still running.
If a Job managed by Kueue is deleted while it still has a finalizer, Kueue prematurely removes the associated Workload even though the underlying Job pod might still be running.

This issue was discovered in #12778.

**Root Cause**
The bug was introduced in #11596 with the implementation of FinalizeOrphanedWorkload. Currently, the logic triggers orphan finalization too early.

**Proposed Fix**
We should modify the logic so that a Workload is only considered an orphan if it has an owner reference, but that owner object _no longer exists_ in the cluster.

mbobrovskyi:
#### What type of PR is this?

/kind bug

Defer workload finalization until owner job is gone.

Fixes #12820

```release-note
Workloads: Fixed a bug that caused Workloads to be finalized as orphaned while their owner Job was being deleted. Kueue now waits until the owner Job no longer exists before finalizing the Workload.

ACTION REQUIRED: If you implement the ComposableJob interface in a custom job integration, change your Load method to return (*jobframework.LoadResult, error) instead of (bool, error). Use jobframework.NewLoadResult(shouldFinalize, found) where found is false when the job is absent and true otherwise.
```

## AI summary

Kueue now defers Workload orphan finalization until the owner Job no longer exists. This prevents premature quota release while the Job or its pod may still be running. Tests cover the deferred finalization flow.

### Suggested release note

Workloads: Fixed a bug that caused Workloads to be finalized as orphaned while their owner Job was being deleted. Kueue now waits until the owner Job no longer exists before finalizing the Workload.

### DIFF
--- a/pkg/controller/jobframework/interface.go
+++ b/pkg/controller/jobframework/interface.go
@@ -124,9 +124,10 @@ type JobWithCustomValidation interface {
 // ComposableJob is an optional interface that should be implemented by generic jobs
 // composed of multiple API objects.
 type ComposableJob interface {
-	// Load loads all members of the composable job. If removeFinalizers is true,
-	// workload and job finalizers should be removed.
-	Load(ctx context.Context, c client.Client, key *types.NamespacedName) (removeFinalizers bool, err error)
+	// Load loads all members of the composable job.
+	// Returns a boolean indicating whether to remove finalizers,
+	// another boolean indicating if the jobs are not found, and an error.
+	Load(ctx context.Context, c client.Client, key *types.NamespacedName) (removeFinalizers bool, jobNotFound bool, err error)
 
 	// Run unsuspends all members of the ComposableJob and injects node affinity
 	// with pod set counts extracted from the workload into all members of the job.
--- a/pkg/controller/jobframework/interface.go
+++ b/pkg/controller/jobframework/interface.go
@@ -134,12 +134,24 @@ type JobWithCustomValidation interface {
 	ValidateOnUpdate(ctx context.Context, oldJob GenericJob) (field.ErrorList, error)
 }
 
+// LoadResult contains the result of loading a job.
+type LoadResult struct {
+	// ShouldFinalize indicates whether the Workload corresponding to the Job should be finalized,
+	// i.e. have "resource-in-use" removed.
+	ShouldFinalize bool
+	// Found indicates whether the Job exists, or at least one member of the composable job exists.
+	Found bool
+}
+
+func NewLoadResult(shouldFinalize bool, found bool) *LoadResult {
+	return &LoadResult{ShouldFinalize: shouldFinalize, Found: found}
+}
+
 // ComposableJob is an optional interface that should be implemented by generic jobs
 // composed of multiple API objects.
 type ComposableJob interface {
-	// Load loads all members of the composable job. If removeFinalizers is true,
-	// workload and job finalizers should be removed.
-	Load(ctx context.Context, c client.Client, key *types.NamespacedName) (removeFinalizers bool, err error)
+	// Load loads all members of the composable job.
+	Load(ctx context.Context, c client.Client, key *types.NamespacedName) (*LoadResult, error)
 
 	// Run unsuspends all members of the ComposableJob and injects node affinity
 	// with pod set counts extracted from the workload into all members of the job.
--- a/pkg/controller/jobframework/reconciler.go
+++ b/pkg/controller/jobframework/reconciler.go
@@ -292,7 +292,7 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
 		err = r.ignoreUnretryableError(log, err)
 	}()
 
-	shouldFinalize, err := r.loadJob(ctx, &req.NamespacedName, job)
+	shouldFinalize, jobNotFound, err := r.loadJob(ctx, &req.NamespacedName, job)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
@@ -304,7 +304,7 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
 	}
 
 	if shouldFinalize {
-		if err := r.finalize(ctx, req.NamespacedName, job); err != nil {
+		if err := r.finalize(ctx, req.NamespacedName, job, jobNotFound); err != nil {
 			return ctrl.Result{}, err
 		}
 		return ctrl.Result{}, nil
@@ -633,38 +633,32 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
 	return ctrl.Result{}, nil
 }
 
-// loadJob retrieves and loads the specified job resource into memory.
-// Returns true if the job should be finalized and an error if loading fails.
-func (r *JobReconciler) loadJob(ctx context.Context, key *types.NamespacedName, job GenericJob) (bool, error) {
+// loadJob loads a job from the Kubernetes cluster and determines
+// if it is deleted or should be treated as absent.
+func (r *JobReconciler) loadJob(ctx context.Context, key *types.NamespacedName, job GenericJob) (bool, bool, error) {
 	if cJob, isComposable := job.(ComposableJob); isComposable {
 		return cJob.Load(ctx, r.client, key)
 	}
 	obj := job.Object()
 	if err := r.client.Get(ctx, *key, obj); err != nil {
 		if client.IgnoreNotFound(err) != nil {
-			return false, err
+			return false, false, err
 		}
-		return true, nil
+		return true, true, nil
 	}
-	return !obj.GetDeletionTimestamp().IsZero(), nil
+	return !obj.GetDeletionTimestamp().IsZero(), false, nil
 }
 
 // finalize removes finalizers from workloads and the job itself,
 // ensuring proper cleanup during object deletion.
-func (r *JobReconciler) finalize(ctx context.Context, key types.NamespacedName, job GenericJob) error {
-	// Remove workloads finalizer
-	if err := r.finalizeWorkloads(ctx, key, job); client.IgnoreNotFound(err) != nil {
-		return err
+func (r *JobReconciler) finalize(ctx context.Context, key types.NamespacedName, job GenericJob, jobNotFound bool) error {
+	if !jobNotFound {
+		// Remove job finalizer for job with finalizer.
+		return client.IgnoreNotFound(r.finalizeJob(ctx, job))
 	}
 
-	// Remove job finalizer
-	if !job.Object().GetDeletionTimestamp().IsZero() {
-		if err := r.finalizeJob(ctx, job); client.IgnoreNotFound(err) != nil {
-			return err
-		}
-	}
-
-	return nil
+	// Remove the workload finalizer only if the Job was deleted.
+	return r.finalizeWorkloads(ctx, key, job)
 }
 
 // getWorkloads retrieves a list of workloads associated with the specified job.
@@ -697,7 +691,8 @@ func (r *JobReconciler) finalizeWorkloads(ctx context.Context, key types.Namespa
 	}
 	for i := range workloads {
 		wl := &workloads[i]
-		if err := workload.FinalizeOrphanedWorkload(ctx, r.client, r.clock, wl, controllerutil.HasControllerReference(wl)); err != nil {
+		err := workload.FinalizeOrphanedWorkload(ctx, r.client, r.clock, wl, controllerutil.HasControllerReference(wl))
+		if client.IgnoreNotFound(err) != nil {
 			return err
 		}
 	}
--- a/pkg/controller/jobframework/reconciler.go
+++ b/pkg/controller/jobframework/reconciler.go
@@ -322,7 +322,7 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
 		err = r.ignoreUnretryableError(log, err)
 	}()
 
-	shouldFinalize, err := r.loadJob(ctx, &req.NamespacedName, job)
+	loadResult, err := r.loadJob(ctx, &req.NamespacedName, job)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
@@ -333,8 +333,8 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
 		}
 	}
 
-	if shouldFinalize {
-		if err := r.finalize(ctx, req.NamespacedName, job); err != nil {
+	if loadResult.ShouldFinalize {
+		if err := r.finalize(ctx, req.NamespacedName, job, loadResult.Found); err != nil {
 			return ctrl.Result{}, err
 		}
 		return ctrl.Result{}, nil
@@ -710,38 +710,31 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
 	return ctrl.Result{}, nil
 }
 
-// loadJob retrieves and loads the specified job resource into memory.
-// Returns true if the job should be finalized and an error if loading fails.
-func (r *JobReconciler) loadJob(ctx context.Context, key *types.NamespacedName, job GenericJob) (bool, error) {
+// loadJob loads a job from the Kubernetes cluster and determines
+// if it is deleted or should be treated as absent.
+func (r *JobReconciler) loadJob(ctx context.Context, key *types.NamespacedName, job GenericJob) (*LoadResult, error) {
 	if cJob, isComposable := job.(ComposableJob); isComposable {
 		return cJob.Load(ctx, r.client, key)
 	}
 	obj := job.Object()
 	if err := r.client.Get(ctx, *key, obj); err != nil {
 		if client.IgnoreNotFound(err) != nil {
-			return false, err
+			return nil, err
 		}
-		return true, nil
+		return NewLoadResult(true, false), nil
 	}
-	return !obj.GetDeletionTimestamp().IsZero(), nil
+	return NewLoadResult(!obj.GetDeletionTimestamp().IsZero(), true), nil
 }
 
 // finalize removes finalizers from workloads and the job itself,
 // ensuring proper cleanup during object deletion.
-func (r *JobReconciler) finalize(ctx context.Context, key types.NamespacedName, job GenericJob) error {
-	// Remove workloads finalizer
-	if err := r.finalizeWorkloads(ctx, key, job); client.IgnoreNotFound(err) != nil {
-		return err
-	}
-
-	// Remove job finalizer
-	if !job.Object().GetDeletionTimestamp().IsZero() {
-		if err := r.finalizeJob(ctx, job); client.IgnoreNotFound(err) != nil {
+func (r *JobReconciler) finalize(ctx context.Context, key types.NamespacedName, job GenericJob, jobFound bool) error {
+	if jobFound {
+		if err := client.IgnoreNotFound(r.finalizeJob(ctx, job)); err != nil {
 			return err
 		}
 	}
-
-	return nil
+	return r.finalizeWorkloads(ctx, key, job, jobFound)
 }
 
 // getWorkloads retrieves a list of workloads associated with the specified job.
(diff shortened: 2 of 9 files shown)


pkg/controller/jobframework/reconciler.go
```diff
@@ -695,6 +695,12 @@ func (r *JobReconciler) finalizeWorkloads(ctx context.Context, key types.Namespa
 	if err != nil {
 		return err
 	}
+	// A workload is only considered orphaned if its owner no longer exists in the cluster.
+	// If the owner was found with a DeletionTimestamp, it is still present and its pods may
+	// still be running — defer finalization until the owner is truly gone from the cluster.
+	if !job.Object().GetDeletionTimestamp().IsZero() {
+		return nil
+	}
```
mimowo: I don't understand what is the mechanism of deferring here. I'm rather sceptical. My hunch is that we should check if the object is gone or not by Get, from within the workload_controller.

Have you been able to test if this fix really works for RayService on a live cluster?

mimowo:
/assign @yaroslava-serdiuk 
cc @kevin85421 ptal

pkg/controller/jobframework/reconciler.go
mbobrovskyi: > My hunch is that we should check if the object is gone or not by Get, from within the workload_controller.

I'm not sure this is the correct approach. I don't see an issue in `workload_controller.go`. In `workload_controller.go`, we only handle Workloads that don't have an `ownerReference`.

https://github.com/kubernetes-sigs/kueue/blob/8bf23d956b685c994c0942e9e8e702f9c2c9cf94/pkg/controller/core/workload_controller.go#L788-L791

The problem occurs in `reconciler.go`. That's where we finish and finalize Jobs. We're triggering orphan finalization too early.

The `shouldFinalize` is `true` if job is not found or job deleted.

https://github.com/kubernetes-sigs/kueue/blob/8bf23d956b685c994c0942e9e8e702f9c2c9cf94/pkg/controller/jobframework/reconciler.go#L295-L299

So if `true` we are finalizing workloads and job.

https://github.com/kubernetes-sigs/kueue/blob/8bf23d956b685c994c0942e9e8e702f9c2c9cf94/pkg/controller/jobframework/reconciler.go#L306-L311

https://github.com/kubernetes-sigs/kueue/blob/8bf23d956b685c994c0942e9e8e702f9c2c9cf94/pkg/controller/jobframework/reconciler.go#L654-L668

So I think we should prevent finalizing Workloads if the Job still exists. The easiest way to check this is whether DeletionTimestamp is set. If it is nil, it means the Job is not found.

mimowo: My intution is that the correct solution is to check if the owner exists or not here: FinalizeOrphanedWorkload. 

> The easiest way to check this is whether DeletionTimestamp is set. If it is nil, it means the Job is not found.

TBH, i don't understand this. Why DeletionTimestamp == nil indicates the Job not found? That is very weird,even if correct? Why not just explicite check the presence of the job?

kevin85421: I think the solution should be very simple: just change `return !obj.GetDeletionTimestamp().IsZero(), nil` to `return false, nil` so that the workload is only finalized when the object is not found.

http://github.com/kubernetes-sigs/kueue/blob/c968bd8b9d60f918b3e26ae3f17423a3f063a31b/pkg/controller/jobframework/reconciler.go#L648

mbobrovskyi: Actually, not exactly, because we still need to clean up the Job finalizer. I think this function should return two boolean values: isDeleted and isNotFound. That way, we can determine when it's safe to clean up the finalizer.

kevin85421: I am not familiar with the code, so maybe I'm missing something, but my intuition would be:

With the solution I mentioned in https://github.com/kubernetes-sigs/kueue/pull/12849#discussion_r3538503055,

* User deletes the RayService CR.
* The RayService CR is not deleted immediately because of the finalizer, but the deletion timestamp becomes non-zero.
* `shouldFinalize` will be false even though the deletion timestamp is non-zero.
* The RayCluster is deleted after the Redis cleanup job finishes, and thus KubeRay's RayService controller removes the RayService CR's finalizer, so the RayService CR is deleted.
* The RayService CR is deleted, so the Workload CR is deleted because the RayService CR is the workload's owner.
* I am not sure what the "Job finalizer" is, but I guess it's something added to the workload CR. Therefore, the workload CR will not be deleted immediately, and the workload controller will reconcile again due to the workload's deletion event.
* At the moment, `loadJob` will return `shouldFinalize=true` because the RayService CR has already been deleted.
* Remove the job finalizer.

mimowo:
@kevin85421 @mbobrovskyi could one of you put together the new e2e test from this PR together with the fix in https://github.com/kubernetes-sigs/kueue/pull/12849. We don't necessarily need to merge it together, but I would like to make sure the test is passing after the changes

pkg/controller/jobframework/reconciler.go
```diff
@@ -635,36 +635,30 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
 
 // loadJob retrieves and loads the specified job resource into memory.
 // Returns true if the job should be finalized and an error if loading fails.
-func (r *JobReconciler) loadJob(ctx context.Context, key *types.NamespacedName, job GenericJob) (bool, error) {
+func (r *JobReconciler) loadJob(ctx context.Context, key *types.NamespacedName, job GenericJob) (bool, bool, error) {
```
mimowo, line 638: The solution starts to look complex, and is job dependent. Why not just add a check if the owner job exists from [FinalizeOrphanedWorkload](https://github.com/kubernetes-sigs/kueue/blob/f5fea343add2163abde5168cfeb71ae8242bc097/pkg/workload/workload.go#L1312). We could check the cache similarly as we do in FindAncestorJob.

mimowo:
I would like to put together the new e2e test from https://github.com/kubernetes-sigs/kueue/pull/12778  the fix  (once ready(. We don't necessarily need to merge it together, but I would like to make sure the test is passing after the changes.

kevin85421:
> @kevin85421 @mbobrovskyi could one of you put together the new e2e test from this PR together with the fix in https://github.com/kubernetes-sigs/kueue/pull/12849. 

@mimowo: Would you mind elaborating? "This PR" and #12849 refer to the same thing. Thanks.

mimowo:
Sorry I meant putting together the e2e test from https://github.com/kubernetes-sigs/kueue/pull/12778 and the fix from https://github.com/kubernetes-sigs/kueue/pull/12849

kevin85421:
> Sorry I meant putting together the e2e test from #12778 and the fix from #12849

@mbobrovskyi Would you mind porting the e2e test from #12778 to this PR? I may not have the access to push commits to your branch.

mimowo:
We may also use another experimental PR just to check that

kevin85421:
> We may also use another experimental PR just to check that

Got it. I can do that after we reach a consensus on the solution. The current solution seems a bit more complex than I expected. My intuitive solution would be https://github.com/kubernetes-sigs/kueue/pull/12849#discussion_r3541896265. I am not sure whether I miss anything or not.

pkg/controller/jobframework/reconciler.go
mimowo: We already have a mechanism for dealing with "orphaned" workloads - we should mark them finished. However, in this case we mark the Workload as finished prematurely, when the RayService has the deletionTimestamp rather than when it is actually deleted.

So, my proposal is to just add a guard inside [FinalizeOrphanedWorkload](https://github.com/kubernetes-sigs/kueue/blob/f5fea343add2163abde5168cfeb71ae8242bc097/pkg/workload/workload.go#L1312) and check if the owner is actually existing or not. Posted similar note here: https://github.com/kubernetes-sigs/kueue/pull/12849#discussion_r3541957432 

wdyt?

yaroslava-serdiuk: Actually looking at the code a bit more — I don't think we should call FinalizeOrphanedWorkload() from reconcile().
I suggest to leave Workload removal in FinalizeOrphanedWorkload() and have a dedicated function RemoveFinalizers that removes workload finalizers. In finalizeWorkloads() we should call RemoveFinalizers() only.

mimowo: > I don't think we should call FinalizeOrphanedWorkload() from reconcile().

why?

yaroslava-serdiuk: Ah, disregard—I forgot we actually need to fix the --cascade=orphan case here.

That said, I still think it's worth splitting this functionality. RemoveFinalizers is needed for all Workloads when a Job is deleting, and keeping it inside FinalizeOrphanedWorkload() is a bit misleading if we're calling it for non-orphaned workloads too.

kevin85421: > my proposal is to just add a guard inside [FinalizeOrphanedWorkload](https://github.com/kubernetes-sigs/kueue/blob/f5fea343add2163abde5168cfeb71ae8242bc097/pkg/workload/workload.go#L1312) and check if the owner is actually existing or not. Posted similar note here: https://github.com/kubernetes-sigs/kueue/pull/12849#discussion_r3541957432

I spent some time taking a look at the related code. My current understanding is that a Workload CR can be "finalized" in two cases:

* case 1: [workload_controller.go](https://github.com/kubernetes-sigs/kueue/blob/77b87aaf1dc58d77e85acc8b2cc5059ee897ef67/pkg/controller/core/workload_controller.go#L267) 
* case 2: [jobframework/reconciler.go](https://github.com/kubernetes-sigs/kueue/blob/77b87aaf1dc58d77e85acc8b2cc5059ee897ef67/pkg/controller/jobframework/reconciler.go#L700)

I think we can't assume the order of reconciliation between case 1 and case 2, so we need to ensure that both cases finalize a Workload CR only when its owner is gone. 

Maybe case 1 doesn't need to check the owner: a Workload CR becomes an orphan after its owner is removed from etcd, and if users delete with `--cascade=orphan`, waiting until the owner is deleted doesn't seem required since the users have decided to opt out. That said, adding a safeguard is still safer.

With the safeguard, we can cover both cases, but I would suggest 

* For case 1, add a safeguard in [isOrphanedWorkload](https://github.com/kubernetes-sigs/kueue/blob/77b87aaf1dc58d77e85acc8b2cc5059ee897ef67/pkg/controller/core/workload_controller.go#L792) instead, because it's a bit weird to call FinalizeOrphanedWorkload when the workload isn't actually finalized.
* For case 2, apply the change (https://github.com/kubernetes-sigs/kueue/pull/12849#discussion_r3538503055) because it's weird that a Workload is not finalized when shouldFinalize is true.

`FinalizeOrphanedWorkload` can also solve the issue, and we only need to change the code in one place. The suggestion above is just my personal preference, so feel free to decide whatever works best for you.

hack/testing/e2e-test.sh
```diff
@@ -70,9 +70,12 @@ STARTUP_PID=""
...
+echo "Test"
+read
 
-run_e2e_ginkgo --json-report=e2e.json --output-dir="$ARTIFACTS" -v ./test/e2e/"${E2E_TARGET_FOLDER}"/...
-"$ROOT_DIR/bin/ginkgo-top" -i "$ARTIFACTS/e2e.json" > "$ARTIFACTS/e2e-top.yaml"
+#if [[ -n ${PROMETHEUS_OPERATOR_VERSION:-} && ("$GINKGO_ARGS" =~ feature:prometheus || ! "$GINKGO_ARGS" =~ "--label-filter") ]]; then
+#    deploy_kueue_prometheus_config ""
+#fi
+
+#run_e2e_ginkgo --json-report=e2e.json --output-dir="$ARTIFACTS" -v ./test/e2e/"${E2E_TARGET_FOLDER}"/...
```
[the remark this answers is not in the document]
mbobrovskyi: Reverted

pkg/controller/jobframework/reconciler.go
mbobrovskyi: We can't do that because we can't subscribe to job events from workload_controller.

pkg/controller/jobframework/interface.go
```diff
@@ -137,9 +137,10 @@ type JobWithCustomValidation interface {
 // ComposableJob is an optional interface that should be implemented by generic jobs
 // composed of multiple API objects.
 type ComposableJob interface {
-	// Load loads all members of the composable job. If removeFinalizers is true,
-	// workload and job finalizers should be removed.
-	Load(ctx context.Context, c client.Client, key *types.NamespacedName) (removeFinalizers bool, err error)
+	// Load loads all members of the composable job.
+	// Returns a boolean indicating whether to remove finalizers,
+	// another boolean indicating if the jobs are not found, and an error.
+	Load(ctx context.Context, c client.Client, key *types.NamespacedName) (removeFinalizers bool, jobNotFound bool, err error)
```
mimowo: Why passing the result by the bool? Cannot we check existence of the owner (RayCluster) in the controller?

mimowo:
/hold
Let's first evaluate if the fix allows to cleanup the hacky code we introduced here: https://github.com/kubernetes-sigs/kueue/pull/12778/changes#diff-421fcbeca6fa06cf9dfc39890b166df9703bb43b593b809ca44b4cca640b0c6cR153-R155

pkg/controller/jobframework/interface.go
mbobrovskyi: It won't work for PodGroups. For PodGroups, `jobNotFound` is returned when all pods in the group are not found.

mbobrovskyi:
> /hold Let's first evaluate if the fix allows to cleanup the hacky code we introduced here: https://github.com/kubernetes-sigs/kueue/pull/12778/changes#diff-421fcbeca6fa06cf9dfc39890b166df9703bb43b593b809ca44b4cca640b0c6cR153-R155

It works. I tested it in this experimental PR: https://github.com/kubernetes-sigs/kueue/pull/15328.

pkg/controller/jobframework/interface.go
mimowo: I'm not convinced we need solving the issue for PodGroups. AFAIK this is only a problem for RayService.

But I'm ok with a generic solution, but in that case I want to avoid suspicious bools in the interface. 

So I would like to propose

```golang
type LoadResult interface {
    // ShouldFinalize indicates the Workload corresponding to the Job should be finalized, ie.
    // have the "resource-in-use" removed.
    ShouldFinalize() bool
    // Found indicates the Job exists, or at least on member of the composable job exists
    Found() bool
}

Load(ctx context.Context, c client.Client, key *types.NamespacedName) (LoadResult, error)
```
This will also make the interface to evolve in the future.

mimowo:
Nice, in this case, should we extend the scope of the PR and make sure the hack for Ray is cleaned up?

pkg/controller/jobframework/interface.go
mbobrovskyi: Done

mbobrovskyi:
> Nice, in this case, should we extend the scope of the PR and make sure the hack for Ray is cleaned up?

Deprecate the `DeferRayServiceFinalizationForRedisCleanup` feature gate. The current implementation should now be used by default.

pkg/controller/jobframework/interface.go
```diff
@@ -134,13 +134,43 @@ type JobWithCustomValidation interface {
...
+	shouldFinalize bool
+	found          bool
+}
+
+func NewDefaultLoadResult(shouldFinalize, found bool) *DefaultLoadResult {
+	return &DefaultLoadResult{
+		shouldFinalize: shouldFinalize,
+		found:          found,
+	}
+}
```
mimowo: Can we scope this to where used rather than exposing? If we need here then probably struct is better than interface actually

mbobrovskyi: Fixed

pkg/features/kube_features.go
```diff
@@ -907,6 +907,7 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 
 	DeferRayServiceFinalizationForRedisCleanup: {
 		{Version: version.MustParse("0.19"), Default: true, PreRelease: featuregate.Beta},
+		{Version: version.MustParse("0.20"), Default: false, PreRelease: featuregate.Deprecated}, // remove in 0.22
```
mimowo, line 910: Awesome, thank you 👍

mimowo (commented):
/lgtm
/approve
Awesome, thank you for cleaning up after the quick fix 👍

mimowo:
I don't have a strong view about CP. We only observed the bug for KubeRay, and we have a fix there (workaround).


Also, backporting breaks the interfaces, so I would suggest backporting only if we discover some issue by a user. Otherwise I think we can continue as is.

wdyt @mbobrovskyi @tenzen-y ?

mbobrovskyi:
> I don't have a strong view about CP. We only observed the bug for KubeRay, and we have a fix there (workaround).
> 
> Also, backporting breaks the interfaces, so I would suggest backporting only if we discover some issue by a user. Otherwise I think we can continue as is.
> 
> wdyt @mbobrovskyi @tenzen-y ?

SGTM

mimowo:
@tenzen-y I think we will need to CP this PR (breaking interface) to fix this issue: https://github.com/kubernetes-sigs/kueue/pull/15154

I think we could mitigate by ACTION REQUIRED, wdyt? It will only affect users who implement the PodGroup interface themselves, so they are pretty expert.

mimowo:
cc @mbobrovskyi wdyt?

mbobrovskyi:
Sure. Let me prepare the cherry-picks.

mimowo:
Thank you 👍

tenzen-y:
> @tenzen-y I think we will need to CP this PR (breaking interface) to fix this issue: #15154
> 
> I think we could mitigate by ACTION REQUIRED, wdyt? It will only affect users who implement the PodGroup interface themselves, so they are pretty expert.

@mimowo @mbobrovskyi I'm ok w/ backport, but this PR seems to have another MK bug since when they execute `kubectl delete --cascade=orphan` in the manager cluster. So, I would recommend using a feature gate w/ disabling by default.

mimowo:
@tenzen-y what is the bug? is it introduced by this fix or pre-existing? What are the consequences of the bug? These are important questions to know before adding the complexity for the FG.

Also, if simple to fix, then maybe we just fix the bug than introduce the FG anyway.

tenzen-y:
> @tenzen-y what is the bug? is it introduced by this fix or pre-existing? What are the consequences of the bug? These are important questions to know before adding the complexity for the FG.
> 
> Also, if simple to fix, then maybe we just fix the bug than introduce the FG anyway.

I think it's not exsting bug, new bug. Let me leave code comment.

pkg/controller/jobframework/reconciler.go
```diff
@@ -767,14 +760,23 @@ func (r *JobReconciler) getWorkloads(ctx context.Context, key types.NamespacedNa
...
 	workloads, err := r.getWorkloads(ctx, key, job)
 	if err != nil {
 		return err
 	}
 	for i := range workloads {
 		wl := &workloads[i]
-		if err := workload.FinalizeOrphanedWorkload(ctx, r.client, r.clock, wl, controllerutil.HasControllerReference(wl)); err != nil {
+		if jobFound && wl.DeletionTimestamp.IsZero() {
+			continue
+		}
```
tenzen-y, line 777: In case of `--cascade=orphan` mode, I guess the problematic point is this `continue` statement when `jobFound=ture1.

1. This process allows us to keep holding ownerReferences in Workload regardless of deletionTimestamp
2. K8s GC controller removes Workload ownerReferences due to orphan cascade deletion mode.
3. MK AC state will be Rejected https://github.com/kubernetes-sigs/kueue/blob/74a6a4b07e91c5ccf47777b1adb6cf12de301b06/pkg/controller/admissionchecks/multikueue/workload.go#L229-L238 via https://github.com/kubernetes-sigs/kueue/blob/74a6a4b07e91c5ccf47777b1adb6cf12de301b06/pkg/controller/admissionchecks/multikueue/workload.go#L310-L333
4. Finally, the remote Job (worker cluster Job) will not be handled by the manager, which has never been cleaned up.

tenzen-y (commented):
@mimowo I'm suspecting the following new problem. Please let me know If I am missing anything.

pkg/controller/jobframework/reconciler.go
mimowo: Sorry, I cannot understand the link. I think the code here is correct actually, we should defer Workload finalization until the job is gone. I wouldn't complicate the code here because of MK.

Similarly I think point (2.) and (3,) are correct things to happen - I think rejecting a workload which lost owner is the right thing to do.

Intuitively I agree we might have a bug wrt `--cascade=orphan`, but this seems to me rather at (4) that MK fails to perform worker cleanup for rejected jobs.

I would suggest we open an issue for MK specifically and investigate how to fix it there.

mimowo: I also cannot understand why this is a new bug or pre-existing, this seems disconnected to me. 

The handling of MK workloads without Job is inside MK code, not here. Maybe there is some weird timing issue that could make this less / more likely, but I cannot understand the design link between the two.

I would consider opening a MK issue to investigate that further with integraion tests.

mimowo: Opened: https://github.com/kubernetes-sigs/kueue/issues/15488

PR #12850, opened 2026-07-07
test : Deleting an entire block in job_test.go

### LINKED ISSUE #12785: E2E test "Should readmit preempted job with priorityClass into a separate flavor" can cause cross-test kube-scheduler preemption pollution
**What happened**:
While investigating #12765, we found that `test/e2e/singlecluster/baseline/job_test.go` has a second
occurrence of a native Kubernetes `PriorityClass` (`scheduling.k8s.io`) being used in an e2e test:

https://github.com/kubernetes-sigs/kueue/blob/main/test/e2e/singlecluster/baseline/job_test.go#L334-L370

```go
ginkgo.It("Should readmit preempted job with priorityClass into a separate flavor", func() {
    ...
    highPriorityClass := utiltesting.MakePriorityClass("high-" + ns.Name).PriorityValue(100).Obj()
    util.MustCreate(ctx, k8sClient, highPriorityClass)
    ...
    job := testingjob.MakeJob("high", ns.Name).
        Queue("main").
        Image(util.GetAgnHostImage(), util.BehaviorWaitForDeletion).
        PriorityClass(highPriorityClass.Name).
        RequestAndLimit(corev1.ResourceCPU, "1").
        NodeSelector("instance-type", "on-demand").
        TerminationGracePeriod(1).
        Obj()
    ...
})
```

As established in #12765, a real (cluster-scoped) `PriorityClass` is visible to kube-scheduler
itself, not just to Kueue. kube-scheduler performs its own native preemption cluster-wide, with
no awareness of Kueue namespaces/queues. When e2e tests run concurrently (in parallel namespaces),
this can cause kube-scheduler to preempt low-priority pods belonging to *other, unrelated* e2e
tests, as we saw happen to a pod_test.go pod in #12765.

Per @mimowo's suggestion in #12765: *"Basically any e2e test which is using PriorityClass
should be either converted to WorkloadPriorityClass, or moved to the sequential suite."*

For this specific test, since it needs to keep exercising real `PriorityClass` semantics,
moving it to the sequential e2e suite (so it doesn't run concurrently with other namespaces
and can't cause cross-test pollution) looks like the right option, rather than converting it
to `WorkloadPriorityClass`.

Not currently flaking in observed CI runs, but same underlying mechanism as #12765 applies:
run this test concurrently with any other e2e test that creates a low/no-priority pod, and
kube-scheduler's native preemption can target the wrong namespace's pod.

Unlike the `"Should deduplicate env variables"` test fixed in #12765 (where the
`PriorityClass` usage was incidental to what was being tested), this test is *specifically*
about native-`PriorityClass` preemption behavior. Its immediate neighbor test,
`"Should readmit preempted job with workloadPriorityClass into a separate flavor"` (line 372),
already covers the equivalent `WorkloadPriorityClass` path. If we simply swap this test's
`PriorityClass` for a `WorkloadPriorityClass`, we'd:
- make it redundant with its neighbor, and
- lose test coverage of the native-`PriorityClass` code path entirely.

ekam-walia:
#### What type of PR is this?

/kind cleanup
/area testing
/kind flake

Fixes #12785

```release-note
NONE
```

## Summary by CodeRabbit

* **Tests**
  * Updated end-to-end coverage for job preemption and re-admission behavior.
  * The scenario now verifies re-admission using workload-based priority, while still confirming the expected shift from on-demand to spot placement after preemption.

### DIFF
--- a/test/e2e/singlecluster/baseline/job_test.go
+++ b/test/e2e/singlecluster/baseline/job_test.go
@@ -333,44 +333,6 @@ var _ = ginkgo.Describe("Kueue", ginkgo.Label("area:singlecluster", "feature:job
 			})
 		})
 
-		ginkgo.It("Should readmit preempted job with priorityClass into a separate flavor", func() {
-			util.MustCreate(ctx, k8sClient, sampleJob)
-
-			highPriorityClass := utiltesting.MakePriorityClass("high-" + ns.Name).PriorityValue(100).Obj()
-			util.MustCreate(ctx, k8sClient, highPriorityClass)
-			ginkgo.DeferCleanup(func() {
-				gomega.Expect(k8sClient.Delete(ctx, highPriorityClass)).To(gomega.Succeed())
-			})
-
-			ginkgo.By("Job is admitted using the first flavor", func() {
-				util.ExpectJobUnsuspendedWithNodeSelectors(ctx, k8sClient, jobKey, map[string]string{
-					"instance-type": "on-demand",
-				})
-			})
-
-			ginkgo.By("Job is preempted by higher priority job", func() {
-				job := testingjob.MakeJob("high", ns.Name).
-					Queue("main").
-					Image(util.GetAgnHostImage(), util.BehaviorWaitForDeletion).
-					PriorityClass(highPriorityClass.Name).
-					RequestAndLimit(corev1.ResourceCPU, "1").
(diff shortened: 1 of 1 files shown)


ekam-walia:
cc @mimowo

mimowo:
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for the cleanup 👍 I think this test is legacy and we can drop it.

PR #12855, opened 2026-07-07
Add e2e test for visibility API pending-workloads namespace isolation

ivnovakov:
#### What type of PR is this?

/area testing

Adds a baseline e2e test asserting the tenancy boundary of the pending-workloads
visibility endpoints when two namespaces share a ClusterQueue.

A subject bound to `kueue-batch-user-role` in one namespace:
- sees only its own namespace's pending workload via the LocalQueue endpoint, never the other namespace's workload;
- is Forbidden from a LocalQueue in a different namespace;
- is Forbidden from the cluster-scoped ClusterQueue endpoint.

The existing e2e tests cover the authorization *errors* (Forbidden/NotFound) but
never assert the *contents* returned to a namespace-scoped tenant. This closes the
RBAC/tenancy coverage described in KEP-168-2 with a positive data-isolation
assertion, confirming the ClusterQueue endpoint is admin-only by design and the
LocalQueue endpoint is correctly scoped to the caller's namespace.

Fixes #

```release-note
NONE
```

## Summary by CodeRabbit

* **Tests**
  * Added an end-to-end RBAC visibility scenario for users with only a namespace-scoped role binding (no admin role).
  * Verified pending-workload visibility is limited to the caller’s own namespace when using a shared cluster queue with multiple local queues.
  * Confirmed requests for a non-existent local queue return “not found” instead of “forbidden”.
  * Ensured access is denied for cross-namespace local queue reads and for cluster-wide pending workload queries.

### DIFF
--- a/test/e2e/singlecluster/baseline/visibility_test.go
+++ b/test/e2e/singlecluster/baseline/visibility_test.go
@@ -233,36 +233,11 @@ var _ = ginkgo.Describe("Kueue visibility server", ginkgo.Label("area:singleclus
 
 		ginkgo.It("Should allow fetching information about position of pending workloads in ClusterQueue", func() {
 			ginkgo.By("Schedule three different jobs with different priorities and two different LocalQueues", func() {
-				jobCases := []struct {
-					JobName          string
-					JobPrioClassName string
-					LocalQueueName   string
-				}{
-					{
-						JobName:          "lq-a-high-prio",
-						JobPrioClassName: highPriorityClass.Name,
-						LocalQueueName:   localQueueA.Name,
-					},
-					{
-						JobName:          "lq-b-mid-prio",
-						JobPrioClassName: midPriorityClass.Name,
-						LocalQueueName:   localQueueB.Name,
-					},
-					{
-						JobName:          "lq-b-low-prio",
-						JobPrioClassName: lowPriorityClass.Name,
-						LocalQueueName:   localQueueB.Name,
-					},
-				}
-				for _, jobCase := range jobCases {
-					job := testingjob.MakeJob(jobCase.JobName, nsA.Name).
-						Queue(kueue.LocalQueueName(jobCase.LocalQueueName)).
-						Image(util.GetAgnHostImage(), util.BehaviorExitFast).
-						RequestAndLimit(corev1.ResourceCPU, "1").
-						WorkloadPriorityClass(jobCase.JobPrioClassName).
-						Obj()
-					util.MustCreate(ctx, k8sClient, job)
-				}
+				createPendingJobs([]pendingJobCase{
+					{JobName: "lq-a-high-prio", JobPrioClassName: highPriorityClass.Name, LocalQueueName: localQueueA.Name, nsName: nsA.Name},
+					{JobName: "lq-b-mid-prio", JobPrioClassName: midPriorityClass.Name, LocalQueueName: localQueueB.Name, nsName: nsA.Name},
+					{JobName: "lq-b-low-prio", JobPrioClassName: lowPriorityClass.Name, LocalQueueName: localQueueB.Name, nsName: nsA.Name},
+				})
 			})
 
 			ginkgo.By("Verify their positions and priorities", func() {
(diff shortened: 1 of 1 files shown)


test/e2e/singlecluster/baseline/visibility_test.go
```diff
@@ -601,6 +601,89 @@ var _ = ginkgo.Describe("Kueue visibility server", ginkgo.Label("area:singleclus
 				}, util.Timeout, util.Interval).Should(gomega.Succeed())
 			})
 		})
+
+		ginkgo.When("A subject bound to kueue-batch-user-role reads a LocalQueue sharing a ClusterQueue across Namespaces", func() {
```
dkaluza: Can we be more accurate? In the one of the next whens we have: `	ginkgo.When("A subject is bound to kueue-batch-user-role, but not to kueue-batch-admin-role"`, can we also indicate the fact that user is not an admin here?

test/e2e/singlecluster/baseline/visibility_test.go
```diff
@@ -601,6 +601,89 @@ var _ = ginkgo.Describe("Kueue visibility server", ginkgo.Label("area:singleclus
...
+				roleBinding = utiltesting.MakeRoleBinding("read-pending-workloads", nsA.Name).
+					RoleRef(rbacv1.GroupName, "ClusterRole", "kueue-batch-user-role").
+					Subject(rbacv1.ServiceAccountKind, "default", kueueNS).
+					Obj()
+				util.MustCreate(ctx, k8sClient, roleBinding)
+				ginkgo.By("Wait for ResourceNotFound error instead of Forbidden to make sure the role binding works", func() {
+					gomega.Eventually(func(g gomega.Gomega) {
+						_, err := impersonatedVisibilityClient.LocalQueues(nsA.Name).GetPendingWorkloadsSummary(ctx, "non-existent", metav1.GetOptions{})
+						g.Expect(err).Should(utiltesting.BeNotFoundError())
+					}, util.Timeout, util.Interval).Should(gomega.Succeed())
```
dkaluza: this is a repetition of the role binding defined below, maybe we can have a local helper for this?

dkaluza (approved):
Generally looks good, added a small improvement suggestions.

test/e2e/singlecluster/baseline/visibility_test.go
ivnovakov: Agreed, added `readPendingWorkloadsUserRoleBinding` func.

ivnovakov:
@dkaluza, thanks for review!
Comments are addressed, ptal when you have time.

test/e2e/singlecluster/baseline/visibility_test.go
ivnovakov: Yep, updated.

test/e2e/singlecluster/baseline/visibility_test.go
```diff
@@ -601,6 +601,86 @@ var _ = ginkgo.Describe("Kueue visibility server", ginkgo.Label("area:singleclus
...
+						util.MustCreate(ctx, k8sClient, job)
+					}
+				})
+
+				roleBinding = readPendingWorkloadsUserRoleBinding(nsA.Name)
+				util.MustCreate(ctx, k8sClient, roleBinding)
+				ginkgo.By("Wait for ResourceNotFound error instead of Forbidden to make sure the role binding works", func() {
+					gomega.Eventually(func(g gomega.Gomega) {
+						_, err := impersonatedVisibilityClient.LocalQueues(nsA.Name).GetPendingWorkloadsSummary(ctx, "non-existent", metav1.GetOptions{})
+						g.Expect(err).Should(utiltesting.BeNotFoundError())
```
dkaluza: nit: this is still a duplication, and could be inside this common helper. Also is the the fact that the binding is named `read-pending-workloads` important here so we have to indicate this in the name? Maybe something more like `mustCreateBatchUserRoleBinding`

test/e2e/singlecluster/baseline/visibility_test.go
```diff
@@ -601,6 +601,86 @@ var _ = ginkgo.Describe("Kueue visibility server", ginkgo.Label("area:singleclus
...
+							Priority:             lowPriorityClass.Value,
+							PositionInLocalQueue: 0,
+							LocalQueueName:       kueue.LocalQueueName(localQueueA.Name),
+						},
+					}
+					// PositionInClusterQueue is relative to the other Namespace's equal-priority workload
+					// and is therefore not deterministic; the isolation guarantee under test is that only
+					// the caller's own Namespace workload is returned, never the workload from nsB.
+					cmpOpts := append(cmp.Options{}, pendingWorkloadsCmpOpts...)
+					cmpOpts = append(cmpOpts, cmpopts.IgnoreFields(visibility.PendingWorkload{}, "PositionInClusterQueue"))
```
dkaluza: nit: you can just have the second workload with different priority, e.g. greater, and then this will be always deterministic and known

ivnovakov: Done

test/e2e/singlecluster/baseline/visibility_test.go
ivnovakov: Done

test/e2e/singlecluster/baseline/visibility_test.go
```diff
@@ -601,6 +522,68 @@ var _ = ginkgo.Describe("Kueue visibility server", ginkgo.Label("area:singleclus
...
+				ginkgo.By("Wait for ResourceNotFound error instead of Forbidden to make sure the role binding works", func() {
+					gomega.Eventually(func(g gomega.Gomega) {
+						_, err := impersonatedVisibilityClient.LocalQueues(nsA.Name).GetPendingWorkloadsSummary(ctx, "non-existent", metav1.GetOptions{})
+						g.Expect(err).Should(utiltesting.BeNotFoundError())
+					}, util.Timeout, util.Interval).Should(gomega.Succeed())
+				})
+			})
+
+			ginkgo.AfterEach(func() {
+				gomega.Expect(k8sClient.Delete(ctx, roleBinding)).To(gomega.Succeed())
```
mimowo: Can we use the retrying ExpectObjectToBeDeleted just to prevent transient failures in cleanup

mimowo:
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for improving the test coverage 👍

test/e2e/singlecluster/baseline/visibility_test.go
ivnovakov: Done. Switched this and two other places in this file to match the rest of the suite.

mimowo:
/lgtm
Thank you 👍

Issue #11133, opened 2026-05-12
Visibility server fails with "read-only file system" when readOnlyRootFilesystem: true (default in Helm chart)


cixuuz:
### What happened

After upgrading from Kueue v0.11.0 to v0.17.2 using the official Helm chart with default values, the controller crashes in a loop with:
```
{"level":"error","logger":"visibility-server","caller":"visibility/server.go:65","msg":"Unable to apply VisibilityServerOptions","error":"error creating self-signed certificates: open /tmp/apiserver.crt: read-only file system"}
```

### What did you expect to happen

The visibility server should be able to write its self-signed certificates when `enableCertManager: false` (the default).

### Environment

- Kueue version: v0.17.2
- Kubernetes version: 1.30 (EKS)
- Install method: Helm chart with default values
- `enableCertManager`: false (default)

### Root cause

The v0.17.2 Helm chart defaults `controllerManager.manager.containerSecurityContext.readOnlyRootFilesystem` to `true` (new in v0.17.2, not present in v0.11.0). 

When `enableCertManager: false`, the chart provides a `/visibility` emptyDir volume. However, the visibility server binary (`pkg/visibility/server.go:65`) writes its self-signed cert to `/tmp/apiserver.crt` — which is on the read-only root filesystem, not the mounted volume.

### Suggested fix

Add a `/tmp` emptyDir volume mount to the manager deployment template in `charts/kueue/templates/manager/manager.yaml`:

```yaml
volumeMounts:
- mountPath: /tmp
  name: tmp

volumes:
- name: tmp
  emptyDir: {}

**Comment by kannon92** [comment]
I agree with this suggested fix. 

Would you like to apply the patch?

**Comment by kannon92** [comment]
Hmm. 

Thinking on this, I'm not entirely sure if we test self managed certificates for visibility service.

https://github.com/kubernetes-sigs/kueue/pull/6798/changes#r2349334227

cc @MaysaMacedo

**Comment by mimowo** [comment]
> I agree with this suggested fix.

Hold on, why is this change needed - I'm not following. Is it required for users on 0.17.2, or to gracefully migrate users from older versions?

**Comment by kannon92** [comment]
```
    containerSecurityContext:
      readOnlyRootFilesystem: true
      allowPrivilegeEscalation: false
      capabilities:
        drop:
          - ALL
```

[...]

PR #12859, opened 2026-07-07
chart: disable internal cert management when cert-manager is enabled

meln5674:
#### What type of PR is this?

/kind regression

Disable internal certificate management when cert-manager is enabled in the helm chart. Without this, the manager pod attempts to write the key/cert pair to a read-only file it has mounted from a secret and fails.

```release-note
Helm: Fix helm chart failing to install with a manager CrashLoopBackoff when cert-manager integration is enabled.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved default certificate management settings in the controller manager configuration.
  * The internal certificate management block is now applied consistently even when cert-manager support is enabled, ensuring the manager starts with the expected certificate behavior.
  * Updated chart tests to validate the new default behavior when custom manager configuration is provided.

### DIFF
--- a/charts/kueue/templates/manager/manager-config.yaml
+++ b/charts/kueue/templates/manager/manager-config.yaml
@@ -24,13 +24,14 @@ metadata:
 data:
   controller_manager_config.yaml: |
 {{- $config := .Values.managerConfig.controllerManagerConfigYaml | fromYaml }}
-{{- if not .Values.enableCertManager }}
-  {{- $defaultInternalCert := dict
-        "enable" true
-        "webhookServiceName" (printf "%s-webhook-service" (include "kueue.fullname" .))
-        "webhookSecretName" (printf "%s-webhook-server-cert" (include "kueue.fullname" .)) }}
-  {{- if not $config.internalCertManagement }}
-    {{- $_ := set $config "internalCertManagement" $defaultInternalCert }}
-  {{- end }}
+{{- $defaultInternalCert := dict
+      "enable" true
+      "webhookServiceName" (printf "%s-webhook-service" (include "kueue.fullname" .))
+      "webhookSecretName" (printf "%s-webhook-server-cert" (include "kueue.fullname" .)) }}
+{{- if .Values.enableCertManager }}
+  {{- $defaultInternalCert = dict "enable" false }}
+{{- end }}
+{{- if not $config.internalCertManagement }}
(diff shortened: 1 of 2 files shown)


tenzen-y:
/hold

As I check linked issue, we are now still discussing. So, meanwhile I will hold this PR.

meln5674:
To clarify, is that discussion regarding the `readOnlyRootFilesystem` and `emptyDir`s? Meaning no disrespect, the original poster of that issue got the diagnosis wrong, the readability of the underlying filesystem is immaterial to the issue, the manager shouldn't have been writing to the file to begin with. This is a simple case of the `internalCertManagement.enable` field having a different default value depending on if cert-manager integration is enabled or not.

mimowo:
@meln5674 I think you PR looks like a legit bugfix for `enableCertManager=true` in Helm.

However, I'm not sure the PR is really fixing https://github.com/kubernetes-sigs/kueue/issues/11133 at all.
This is because the issue claims problems for `enableCertManager=false`: "The visibility server should be able to write its self-signed certificates when enableCertManager: false (the default).".

So, I think we could merge this PR as a bugfix (without dedicated issue), but keep the issue and that discussion separeate. Let me know if I'm missing something.

mimowo:
> Fix helm chart failing to install when cert-manager integration is enabled.

In this release note - could you expand a bit to clarify how it would fail (at which stage of installation), or maybe post-installation fail to run?

meln5674:
Updated, let me know if this is sufficient.

mimowo:
Yes, but should we also remove `Fixes #11133` from PR description? That would also confirm my understanding the two are actually unrelated - trying to solve the opposite configuration for `enableCertManager`

meln5674:
it seems I didn't read the issue closely enough, and assumed it was the same issue I was encountering because it had the same error message. The original issue occurred on a closed system that I didn't have available to cross reference. Apologies for the confusion, I've removed it from the PR description.

mimowo (commented):
Np, thank you for fixing the Helm installation when cert-manager is used 👍 

/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17

/unhold

mimowo:
/release-note-edit
```release-note
Helm: Fix helm chart failing to install with a manager CrashLoopBackoff when cert-manager integration is enabled.
```

PR #12885, opened 2026-07-08
MultiKueue: propagate manager-driven elastic RayCluster worker replicas

jiaoew1991:
#### What type of PR is this?

/kind feature
/area multikueue
/area integrations

The Ray MultiKueue adapter creates the remote object once and then only copies
status back, so an in-place worker-replica change of an elastic
(`ElasticJobsViaWorkloadSlices`) RayCluster on the management cluster never
reaches the worker cluster. `batch/Job` already handles this via
`needElasticJobSync` / `syncElasticJob`; this PR gives the shared Ray adapter the
same capability.

- An optional `ElasticReplicaSync` hook is added to the generic Ray adapter and
  wired only for `RayCluster`. When the feature gate is on and the workload is
  slicing-enabled, `SyncJob` detects a worker-replica change on the manager and
  patches the remote object's per-worker-group replicas (and the prebuilt
  workload label), reusing `batch/Job`'s stale scale-up guard.
- `RayJob` and `RayService` leave the hook unset: their worker replicas live on a
  child `RayCluster` on the worker cluster that has no representation on the
  manager, so they keep the create-once behavior.
- Because scaling is manager-driven, the remote copy must not run the in-tree Ray
  autoscaler (which would otherwise fight the manager by editing replicas on the
  worker cluster); the adapter clears `enableInTreeAutoscaling` on the remote copy
  of an elastic RayCluster.
- `NumOfHosts` is synced alongside `Replicas`, since `needElasticSync` compares the
  effective per-group pod count (`Replicas * NumOfHosts`).

Covered by unit, integration, and e2e tests for the manager-driven scale up/down
path.

This is a new capability; there is no linked issue.

AI disclosure: per the [Kubernetes AI Tool Usage Policy](https://www.kubernetes.dev/docs/guide/pull-requests/#ai-guidance), this change was developed with the assistance of an AI coding tool. All code was authored, reviewed, and tested by the submitter.

```release-note
MultiKueue: Elastic RayCluster worker-group replica changes made on the management cluster (via the `ElasticJobsViaWorkloadSlices` feature gate) now propagate to the RayCluster on the admitting worker cluster. Previously the remote RayCluster was created once and never resized.
```

## Summary by CodeRabbit

* **New Features**
  * Added optional elastic worker replica synchronization for Ray workloads under MultiKueue, gated by the elastic workload-slicing feature and propagating admitted scale changes to worker clusters.
  * Keeps the remote elastic workload aligned with the active elastic workload slice, including updating the workload identity label.

* **Bug Fixes**
  * Prevents propagating stale slice/replica state during elastic scale-up.
  * Ensures in-tree Ray autoscaling remains disabled on destination clusters while workload slicing is enabled.

* **Tests**
  * Expanded unit, integration, and E2E coverage for elastic scale-up/down, stale-slice skipping, and correct behavior for non-elastic RayClusters.

### DIFF
--- a/pkg/controller/jobs/ray/ray_multikueue_adapter.go
+++ b/pkg/controller/jobs/ray/ray_multikueue_adapter.go
@@ -19,6 +19,7 @@ package ray
 import (
 	"context"
 	"fmt"
+	"maps"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -26,11 +27,14 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/ptr"
+	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
 	"sigs.k8s.io/kueue/pkg/controller/jobframework"
+	"sigs.k8s.io/kueue/pkg/features"
 	clientutil "sigs.k8s.io/kueue/pkg/util/client"
+	"sigs.k8s.io/kueue/pkg/workloadslicing"
 )
 
 type objAsPtr[T any] interface {
@@ -46,6 +50,44 @@ type adapter[PtrT objAsPtr[T], T any] struct {
 	gvk          schema.GroupVersionKind
 	getManagedBy func(PtrT) *string
 	setManagedBy func(PtrT, *string)
+	// elastic is optional. When set, the adapter propagates manager-driven,
+	// in-place worker replica changes of an elastic workload to the remote copy
+	// on the worker cluster. It is left unset for job types that do not support
+	// this (see ElasticReplicaSync).
+	elastic *ElasticReplicaSync[PtrT, T]
+}
+
+// ElasticReplicaSync carries the type-specific hooks that let the MultiKueue
+// adapter propagate manager-driven, in-place worker replica changes of an
+// elastic workload (the ElasticJobsViaWorkloadSlices feature) to the remote
+// copy on the worker cluster.
+//
+// Only job types whose worker replicas live directly on the Kueue-managed
+// object wire this. RayCluster does. RayJob and RayService do not: their worker
+// replicas belong to a child RayCluster on the worker cluster that has no
+// representation on the manager, so they keep the create-once behavior and
+// leave this unset.
+type ElasticReplicaSync[PtrT objAsPtr[T], T any] struct {
+	// SyncReplicas copies the worker replica counts from src into dst, returning
+	// whether dst changed.
+	SyncReplicas func(dst, src PtrT) bool
+	// WorkerReplicas returns the per-worker-group replica counts keyed by PodSet
+	// reference. Used to detect a replica change and its direction.
+	WorkerReplicas func(PtrT) map[kueue.PodSetReference]int32
+	// WorkloadNameExtraPart mirrors ElasticWorkloadNameProvider for the type; it
+	// is used to compute the workload name of the object's current slice.
+	WorkloadNameExtraPart func(PtrT) string
+}
+
+// Option configures a Ray MultiKueue adapter.
+type Option[PtrT objAsPtr[T], T any] func(*adapter[PtrT, T])
+
+// WithElasticReplicaSync enables manager-driven elastic replica propagation
+// over MultiKueue for job types that support it (see ElasticReplicaSync).
+func WithElasticReplicaSync[PtrT objAsPtr[T], T any](e *ElasticReplicaSync[PtrT, T]) Option[PtrT, T] {
+	return func(a *adapter[PtrT, T]) {
+		a.elastic = e
+	}
 }
 
 type fullInterface interface {
@@ -64,8 +106,9 @@ func NewMKAdapter[PtrT objAsPtr[T], T any](
 	gvk schema.GroupVersionKind,
 	getManagedBy func(PtrT) *string,
 	setManagedBy func(PtrT, *string),
+	opts ...Option[PtrT, T],
 ) fullInterface {
-	return &adapter[PtrT, T]{
+	a := &adapter[PtrT, T]{
 		copySpec:     copySpec,
 		copyStatus:   copyStatus,
 		emptyList:    emptyList,
@@ -73,6 +116,10 @@ func NewMKAdapter[PtrT objAsPtr[T], T any](
 		getManagedBy: getManagedBy,
 		setManagedBy: setManagedBy,
 	}
+	for _, opt := range opts {
+		opt(a)
+	}
+	return a
 }
 
 func (a *adapter[PtrT, T]) GVK() schema.GroupVersionKind {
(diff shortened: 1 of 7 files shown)


mimowo:
/ok-to-test
cc @yaroslava-serdiuk @olekzabl

pkg/controller/jobs/raycluster/raycluster_multikueue_adapter.go
```diff
@@ -36,6 +42,80 @@ func copyJobSpec(dst, src *rayv1.RayCluster) {
 		ObjectMeta: api.CloneObjectMetaForCreation(&src.ObjectMeta),
 		Spec:       *src.Spec.DeepCopy(),
 	}
+	// An elastic RayCluster over MultiKueue is scaled by the manager: the
+	// manager's worker replica counts are the source of truth and are propagated
+	// to this remote copy on each sync. The remote must therefore not run the
+	// in-tree Ray autoscaler, which would otherwise fight the manager by editing
+	// worker replicas on the worker cluster.
+	if workloadslicing.Enabled(src) {
+		dst.Spec.EnableInTreeAutoscaling = nil
```
ArangoGutierrez, line 51: EnableInTreeAutoscaling is cleared here on the create path, but the elastic sync path (syncElastic) only patches replicas and the workload label, so it never re-asserts this. The manager is the sole replica driver, which is what makes disabling the remote autoscaler load-bearing, so I would expect that invariant to be reconciled rather than set once at create. If the remote spec is genuinely never re-copied after create then this is safe, but I want to be sure that is the intent. Can you either re-assert it in syncElastic when slicing is enabled, or add a one-line comment explaining why create-time clearing is sufficient?

pkg/controller/jobs/raycluster/raycluster_multikueue_adapter.go
```diff
@@ -36,6 +42,80 @@ func copyJobSpec(dst, src *rayv1.RayCluster) {
...
+	return &ray.ElasticReplicaSync[*rayv1.RayCluster, rayv1.RayCluster]{
+		SyncReplicas:          syncWorkerReplicas,
+		WorkerReplicas:        workerReplicaCounts,
+		WorkloadNameExtraPart: func(rc *rayv1.RayCluster) string { return GetWorkloadNameExtraPart(rc) },
+	}
+}
+
+// workerReplicaCounts returns the effective worker pod count per worker group,
+// matching how BuildPodSets derives PodSet counts (replicas scaled by NumOfHosts).
+func workerReplicaCounts(rc *rayv1.RayCluster) map[kueue.PodSetReference]int32 {
```
ArangoGutierrez, line 67: nit: this effective-count derivation (replicas, times NumOfHosts when greater than 1) is now the third copy of the same logic, alongside BuildPodSets in common.go:96 and UpdatePodSets in common.go:226. The elastic guard's correctness depends on it matching BuildPodSets exactly, so a future change to how counts are derived would have to touch all three or the guard silently desyncs from the podset count. Worth extracting a shared effectiveWorkerCount(wgs) helper in common.go and calling it from all three.

pkg/controller/jobs/ray/ray_multikueue_adapter.go
```diff
@@ -132,6 +186,61 @@ func (a *adapter[PtrT, T]) SyncJob(
...
+	// observed state is stale and must not be propagated to the worker cluster.
+	newWorkloadName := jobframework.GenerateWorkloadNameWithExtra(
+		localJob.GetName(), localJob.GetUID(), a.gvk, a.elastic.WorkloadNameExtraPart(localJob))
+	if totalReplicas(oldCounts) < totalReplicas(newCounts) && workloadName != newWorkloadName {
+		ctrl.LoggerFrom(ctx).V(2).Info("Skipping stale ElasticWorkload sync",
+			"observedWorkloadName", workloadName, "currentWorkloadName", newWorkloadName)
+		return false
+	}
+
+	return !maps.Equal(oldCounts, newCounts) || jobframework.PrebuiltWorkloadNameFor(remoteJob) != workloadName
```
ArangoGutierrez, line 222: Detection compares effective per-group counts (replicas times NumOfHosts). Two spec-shape cases fall outside that: an inverse change that keeps the product equal (say replicas 4 / NumOfHosts 1 becoming 2 / 2) would not trigger a sync, and a worker group that exists on the manager but not on the remote is skipped by the name match in syncWorkerReplicas. Both are probably out of scope for replica scaling, and the podset count is what Kueue cares about, so this is not a bug. But are inverse NumOfHosts changes and worker-group additions expected to be supported for elastic RayCluster? If not, a sentence on the type or in the PR description scoping this to per-group replica counts of existing groups would save the next reader the same trace I just did.

pkg/controller/jobs/ray/ray_multikueue_adapter.go
```diff
@@ -132,6 +186,61 @@ func (a *adapter[PtrT, T]) SyncJob(
...
 }
 
+// needElasticSync reports whether the remote object must be updated to reflect
+// a manager-driven worker replica change of an elastic workload. It mirrors the
+// batch/Job adapter, including the stale scale-up guard.
+func (a *adapter[PtrT, T]) needElasticSync(ctx context.Context, workloadName string, localJob, remoteJob PtrT) bool {
+	if a.elastic == nil {
+		return false
+	}
+	if !features.Enabled(features.ElasticJobsViaWorkloadSlices) || !workloadslicing.Enabled(localJob) {
```
ArangoGutierrez, line 202: nit, optional: workloadslicing.Enabled already checks features.Enabled(features.ElasticJobsViaWorkloadSlices) internally, so the leading features.Enabled here is redundant. It mirrors needElasticJobSync in job_multikueue_adapter.go:239, so I am fine leaving it for consistency, but flagging in case you want to drop it in both places.

ArangoGutierrez (commented):
Read through the whole change. The shape is good. Making the shared Ray adapter carry an optional ElasticReplicaSync hook and wiring it only for RayCluster is the right call, and the doc comment on why RayJob and RayService stay create-once is genuinely helpful for the next reader.

Two things I checked closely and they hold up:

- The stale scale-up guard computes newWorkloadName through the WorkloadNameExtraPart hook, so for RayCluster it includes the RayClusterGenerationAnnotation part and matches what newWorkloadName in reconciler.go actually assigns. That is the subtle bit the batch/Job version could not reuse directly, and it looks correct here.
- Scale-up replicas only reach the worker once SyncJob is reconciling with the new admitted slice name, so quota is re-reserved before the larger spec propagates. The integration test exercises exactly that path (new slice, admit, finish old), which is the coverage I most wanted to see.

The one design question I want an answer to before I lgtm is the autoscaler invariant. copySpec clears EnableInTreeAutoscaling only on the create path. The elastic sync path patches replicas and the workload label but never re-asserts it. That is fine as long as the remote spec is never re-copied, but it means the "manager owns replicas, autoscaler is off" invariant is create-time only rather than reconciled. Can you confirm that is intended, or should syncElastic also keep EnableInTreeAutoscaling cleared? See the inline note.

The rest are nits and edge-case questions inline. Nothing blocking.

AI disclosure noted in the description, thanks for including it. Same disclosure from my side: I used an AI assistant while preparing this review and verified the findings myself.

mimowo:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12885/pull-kueue-test-e2e-multikueue-extended-shard-1-main/2075335250845634560
@jjacobelli ptal, maybe the failure is related

jiaoew1991:
Hi @ArangoGutierrez , I fixed the comments, please continue the review

mimowo (commented):
/lgtm
/approve
Thank you for working on this important capability for MK + RayCluster autoscaling👍 

Let me ask a follow up question: have you considered updating the "manager" cluster usage based on the "worker" autoscalings? This would be the missing piece we are looking into with @yaroslava-serdiuk and @olekzabl

mimowo:
/hold
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12885/pull-kueue-test-integration-multikueue-main/2079110106246549504
This looks likely related, ptal

mimowo:
/unhold
/lgtm
I see the diff is clear https://github.com/kubernetes-sigs/kueue/pull/12885/commits/f1bd4168e42dec8326ba83cc57e8b2428ce3139d, related to other rename changes.

mimowo:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12885/pull-kueue-test-integration-shard-2-main/2079265648038383616

This is unrelated. 
@cideeabbitai please open a flake issue. Remember to add the flake lable. Attach also the link to the failed build in the issue description

mimowo:
@coderabbitai please follow the comment above

kannon92:
This is unrelated.
@coderabbitai please open a flake issue. Remember to add the flake label. Attach also the link to the failed build in the issue description

PR #12889, opened 2026-07-09
fix: add nil check for Exactly in dra

mykysha:
#### What type of PR is this?
/kind bug

Add a nil check for Exactly to ensure it is present in case FirstAvailable is nil.

Fixes #

```release-note
DRA: introduce a safeguard for invalid parameter combinations to prevent nil dereference crashes
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for resource claim device requests so incomplete configurations are rejected earlier with a clear error.
  * Prevented requests from continuing when neither supported device selection option is set.

* **Tests**
  * Added coverage for the missing-configuration case to verify the correct validation error is returned.

### DIFF
--- a/pkg/dra/claims.go
+++ b/pkg/dra/claims.go
@@ -74,6 +74,11 @@ func countDevicesPerClass(claimSpec *resourcev1.ResourceClaimSpec) (resources.Re
 			return nil, allErrs
 		}
 
+		if req.Exactly == nil {
+			allErrs = append(allErrs, field.Invalid(devicesRequestsPath.Index(i), nil, "Exactly must be set if FirstAvailable is nil"))
+			return nil, allErrs
+		}
+
 		selectorsPath := devicesRequestsPath.Index(i).Child("exactly", "selectors")
 		if err := validateCELSelectors(req.Exactly.Selectors, selectorsPath); err != nil {
 			allErrs = append(allErrs, field.Invalid(selectorsPath, nil, err.Error()))
--- a/pkg/dra/claims_test.go
+++ b/pkg/dra/claims_test.go
@@ -526,6 +526,32 @@ func Test_GetResourceRequests(t *testing.T) {
 			},
 		},
 		{
+			name: "Exactly and FirstAvailable are nil returns error",
+			extraObjects: []runtime.Object{
+				&resourcev1.ResourceClaimTemplate{
+					ObjectMeta: metav1.ObjectMeta{Name: "claim-tmpl-empty", Namespace: "ns1"},
+					Spec: resourcev1.ResourceClaimTemplateSpec{
+						Spec: resourcev1.ResourceClaimSpec{
+							Devices: resourcev1.DeviceClaim{
+								Requests: []resourcev1.DeviceRequest{
+									{Name: "req"},
+								},
+							},
+						},
+					},
+				},
+			},
+			modifyWL: func(w *kueue.Workload) {
+				w.Spec.PodSets[0].Template.Spec.ResourceClaims = []corev1.PodResourceClaim{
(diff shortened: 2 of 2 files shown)


pkg/dra/claims.go
```diff
@@ -74,6 +74,11 @@ func countDevicesPerClass(claimSpec *resourcev1.ResourceClaimSpec) (resources.Re
 			return nil, allErrs
 		}
 
+		if req.Exactly == nil {
```
kannon92, line 77: Kube API should prevent this from occuring.

Are you adding this based on the code or you saw this in production?

kannon92: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/resource/v1/types.go#L1145

I am curious on what this means since these fields are under a feature gate in 1.37 so it could be possible that both are nil if features are disabled.

kannon92:
/hold

https://github.com/kubernetes-sigs/kueue/pull/12889#discussion_r3553426115

I would like to better understand the motivation and figure out how to handle this API when there are different k8s versions.

pkg/dra/claims.go
mykysha: This was mostly based on the code, not a production issue. Based on some further research I don't think that disabling the gate will allow for such resource to be created, imo this check should serve as more as a safeguard. We do have another `Exactly == nil` check in the same file on line 248 in the `extractCELRequests` function already, and it is called after the `FirstAvailable` requests are discarded.

mimowo: @kannon92 I think the safe-guards in the Kueue code make sense anyway, because these are different codebases which can evolve independently, and are versioned independently. Even if currently there are some assumptions in k8s validation that guarantee req.Exactly != nil this may not always be true on all k8s versions, including future. For example, if a future version of k8s relaxes validation (and relaxing is possible) then we may get panic here.

The situation is different if you edit k8s code (same codebase, same binary). 

Let me know if I'm missing something.

kannon92: I'm mostly trying to understand if I use k8s 1.34 or 1.35 with current Kueue, will we hit a case where both fields are nill since they both don't exist or are turned off.

kannon92: Looks like DRA e2es only runs on one version of Kube.

mimowo: i see, but why is that relevant for the kueue code? 

since kueue is separate binary it should have its own safeguards against nils because assuming external validation is error prone.

For example nothing guarantees the validation is not relaxed in 1.38 and then kueue may panic.

kannon92: I'm more wondering how does this work if on K8s 1.34 I create a resource claim with neither of these fields specified. They didn't exist there or they are alpha turned off will the validation kick off?

mykysha: I have tried creating the resource claim with the feature disabled, it states that there is no resource matching the kind existing. Technically it is possible to create one using `"api-approved.kubernetes.io": "unapproved, experimental-only"`, but I am unsure that we need to prepare for such cases

kannon92: I think this is fine.

It looks like both of these fields exist on the API in 1.34 and we only support v1 DRA in 1.34 and above.

kannon92 (commented):
/lgtm

mimowo:
@mykysha please add a release note. If we don't have a concrete failure scenario, then saying "DRA: introduce a safeguard for invalid parameter cominations to prevent nil dereference crashes is perfectly ok". If we have a concrete failure scenario then even better

mimowo:
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for hardening Kueue and DRA integration 👍

Issue #12787, opened 2026-07-05
AFS entry penalty can accumulate when Workloads leave QuotaReserved without admission


apullo777:
**What happened**:

With `AdmissionFairSharing` enabled, the scheduler pushes an entry penalty for a Workload's LocalQueue at assume time (`AfsEntryPenalties.Push`). That penalty is normally settled when the Workload transitions into `Admitted`: the Workload's usage is folded into consumed usage and the entry penalty is subtracted.

#12539 tracks a missing settlement path into `Admitted` (`QuotaReserved -> Admitted`; fix in progress). However, there are still paths where a Workload can leave `QuotaReserved` without ever reaching `Admitted`, and the entry penalty does not appear to be handled:

1. `QuotaReserved -> Pending`: for example AdmissionCheck `Retry`, preemption, or requeue. The relevant update path appears to be `workload.FromQuotaReservedOrAdmittedToPending(prevStatus, status)`. This path deletes the Workload from the cache and may requeue it, but I do not see any corresponding `AfsEntryPenalties` rollback for the `QuotaReserved -> Pending` subcase. If the Workload is assumed/reserved again later, another entry penalty can be pushed on top of the old one.

2. Workload deletion while `QuotaReserved`: for example, a user deletes a Job while it is waiting for a ProvisioningRequest. I do not see penalty handling in the deletion/deactivation path either.

As far as I can find, the only `AfsEntryPenalties.Sub` call sites are the scheduler's assume-failure rollback and the admission settlement path. Since entry penalties do not decay and are otherwise cleared only with the LocalQueue, these paths can leave the LocalQueue with accumulated pending entry penalties that are never subtracted.

**What you expected to happen**:

The entry-penalty lifecycle should have an explicit contract for every exit from `QuotaReserved`.

There seem to be two possible designs:

- Keep the penalty after eviction/requeue: this would treat the penalty as an intentional anti-churn cost for occupying a scheduling slot, in the spirit of the entry-penalty rationale in KEP-4136. If this is the intended behavior, it should probably be documented, and the repeated evict/re-admit case may need an explicit bound. The deletion case would still need a decision.

- Roll back the penalty on exits that do not reach Admitted: this would make eviction/requeue and deletion symmetric with assume-failure rollback, since the Workload never reached admitted consumed usage.

Either answer would make the lifecycle explicit. Today it is not clear to me whether the current behavior is intentional or an accounting leak.

**How to reproduce it (as minimally and precisely as possible)**:

[...]

PR #12891, opened 2026-07-09
Fix AFS cache initialization and warm-start recovery

### LINKED ISSUE #12783: Flaky: Scheduler when Using AdmissionFairSharing at Cohort level should preempt a workload from LQ with higher recent usage

### LINKED ISSUE #12911: AFS: LocalQueue historical usage can be lost after manager restart when settlement precedes LocalQueue reconcile
**What happened**:

Follow-up to the discussion in #12891.

The in-memory AFS consumed-resources cache is rebuilt after a manager restart by seeding from the persisted LocalQueue status (#12891). This recovery is currently best-effort: if the workload-controller settlement path creates the in-memory entry first, before the LocalQueue's first reconcile, the LocalQueue seed backs off and the persisted historical EMA is not applied.

The cache then keeps only the post-restart settlement entry. On the next sampling tick, that lower value can be written back to status, so the historical EMA is dropped from the warm start.

This predates #12891: the previous live-usage seed was also skipped whenever settlement created the entry first.

The LocalQueue's historical fair-sharing usage should survive a manager restart regardless of reconcile ordering.

For example, with a persisted EMA of `100` and a post-restart admission adding `penalty`, the recovered in-memory value should be `100 + penalty`; today it can end up as just `penalty`.

This is a reconcile-ordering race, so it is timing-dependent:

1. Enable AdmissionFairSharing.
2. Run workloads through a LocalQueue until `status.fairSharing.admissionFairSharingStatus.consumedResources` is non-zero.
3. Restart the kueue-controller-manager.
4. Immediately after restart, have a workload admitted to that LocalQueue, so the workload reconciler observes the admission and creates the AFS cache entry before the LocalQueue's first reconcile.
5. The LocalQueue's consumed resources can restart from roughly the new admission's entry penalty instead of the persisted status value, visible after the next sampling interval.

The fix is being added in #12891.

The fix direction is to merge the persisted status into an already-existing entry exactly once, tracked with a marker such as `statusAccounted`. Once we support merging, entry existence alone no longer tells us whether the persisted history has already been applied.

**Design constraints from that discussion**:

- the marker must be preserved by every writer of `AfsConsumedResources`. Settlement and sampling currently build a fresh entry when they `Set`, so a marker stored on the entry could be dropped. Since sampling also writes the merged value back to status, repeated re-merges could compound rather than add a fixed offset;
- this points to a shared `Update(lqKey, func(entry) entry)` primitive, so writers update the entry under one lock and preserve extra state by construction. The `SetIfAbsent` added in #12891 is the insert-if-absent case of that primitive
- the `Update` closure should not call back into the scheduler cache, to avoid lock-ordering issues. Live usage should be read before taking the entry lock, with only the fold/update done inside
- one accounting caveat for the merge: a second settlement before the LocalQueue's first reconcile can fold a small alpha-weighted slice of live usage that, after restart, may overlap the persisted status. This should be bounded by the short pre-reconcile window, but should be documented at the merge site.

The same primitive would also close related check-then-act windows on `AfsConsumedResources`:

- settlement's read-modify-write can replace a just-seeded entry, producing the same history-loss end state as above
- the sampling tick's read -> status-update API call -> write sequence can overwrite a concurrent settlement, dropping an already-subtracted penalty

The second case can be split into its own bug report if preferred; it is listed here because the same primitive would close it.

**Impact**: under-count side. The affected LocalQueue can temporarily look less used than it is, gaining admission/preemption advantage until the EMA re-converges. This depends on manager restart and reconcile ordering, so it does not look directly tenant-triggerable.

AI assistance was used for parts of the analysis.

apullo777:
#### What type of PR is this?

/kind bug

With AdmissionFairSharing enabled, the in-memory AFS consumed-resources cache is currently seeded from live admitted usage. If a LocalQueue is seeded after a Workload is assumed into the scheduler cache but before its pending entry penalty is settled, the same admission can be counted twice: once in the seed, and again through the entry penalty. This can inflate a LocalQueue's recent usage and affect cohort preemption victim selection, as seen in #12783.

This PR changes the LocalQueue seed path to initialize the cache from the persisted LocalQueue fair-sharing status instead of live usage.

This keeps the two important cases separate:

- a LocalQueue with persisted fair-sharing status can warm-start from that history;
- a LocalQueue with no persisted status seeds empty, so its first admissions are represented by their entry penalties and then converge through the normal accounting path.

The seeded entry is stamped with `LastUpdate=now`, so the sampling pass in the same reconcile sees ~zero elapsed time and does not immediately fold live usage back into the cache.

AI assistance was used for parts of the change.

Fixes #12783
Fixes #12911

Also related to the broader AFS accounting discussion (#12539, #12786, #12787)

Small PR to make the seed-from-status approach concrete, as discussed in https://github.com/kubernetes-sigs/kueue/issues/12783#issuecomment-4892474975.

This should remain compatible with the QR-anchored direction discussed there.

The main alternative discussed was clearing or subtracting pending penalties at seed time. Seeding from persisted status avoids needing to split the per-LocalQueue pending-penalty aggregate, and avoids using an instantaneous snapshot that cannot distinguish a warm restart from a genuinely new LocalQueue.

One boundary: this PR only changes the LocalQueue seed source. It does not guarantee persisted-status warm start if another path creates the cache entry first, and it does not make seed-vs-consolidation atomic. The narrower guarantee
is that the LocalQueue seed path no longer over-counts by starting from live admitted usage.

```release-note
AFS: Fixed consumed-resources cache initialization and warm-start recovery so LocalQueue usage is not over-counted during cache seeding, and persisted historical usage is preserved after manager restarts when workload settlement runs before LocalQueue reconciliation.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Fair Sharing consumed-usage handling on startup/restart, including safer initialization when caches are empty.
  * Prevents double-counting by seeding from persisted fair-sharing status when available, otherwise starting at zero.
  * Hardened consumed-usage updates to correctly account for decay timing and concurrent updates, while preserving accounted-state.

* **Tests**
  * Expanded unit tests for consumed-resource cache update/seeding and decay/accounting behavior.
  * Added an integration scenario verifying LocalQueue historical usage is preserved across manager restarts.

### DIFF
--- a/pkg/controller/core/localqueue_controller.go
+++ b/pkg/controller/core/localqueue_controller.go
@@ -52,6 +52,7 @@ import (
 	"sigs.k8s.io/kueue/pkg/metrics"
 	afs "sigs.k8s.io/kueue/pkg/util/admissionfairsharing"
 	utilqueue "sigs.k8s.io/kueue/pkg/util/queue"
+	utilresource "sigs.k8s.io/kueue/pkg/util/resource"
 	"sigs.k8s.io/kueue/pkg/util/roletracker"
 )
 
@@ -357,14 +358,30 @@ func (r *LocalQueueReconciler) initializeAfsIfNeeded(lq *kueue.LocalQueue) (hadC
 	// correct for a fresh LocalQueue: entry penalties cover its early admissions.
 	// LastUpdate=now keeps the first sampling tick at ~zero elapsed, so it folds
 	// no live usage and the seed stays independent of in-flight admissions.
-	// SetIfAbsent lets a concurrently settled entry win over the seed; settlement's
-	// own read-modify-write may still replace a fresh seed (bounded history loss,
-	// never a double count).
+	// If settlement created the entry first (post-restart admission), the persisted
+	// history is merged into it exactly once per process, tracked by StatusAccounted.
+	// The merge can slightly over-count: a second settlement before this reconcile
+	// folds alpha(elapsed) of live usage, which after a restart overlaps the persisted
+	// status; bounded by the short pre-reconcile window and decaying within a half-life.
 	seeded := corev1.ResourceList{}
 	if hasStatus {
 		seeded = lq.Status.FairSharing.AdmissionFairSharingStatus.ConsumedResources.DeepCopy()
 	}
-	entry, hadCache = r.queues.AfsConsumedResources.SetIfAbsent(lqKey, seeded, now)
+	entry = r.queues.AfsConsumedResources.Update(lqKey, func(old queueafs.ConsumedResourcesEntry, found bool) queueafs.ConsumedResourcesEntry {
+		hadCache = found
+		switch {
+		case !found:
+			return queueafs.ConsumedResourcesEntry{Resources: seeded, LastUpdate: now, StatusAccounted: true}
+		case !old.StatusAccounted:
+			return queueafs.ConsumedResourcesEntry{
+				Resources:       utilresource.MergeResourceListKeepSum(old.Resources, seeded),
+				LastUpdate:      old.LastUpdate,
+				StatusAccounted: true,
+			}
+		default:
+			return old
+		}
+	})
 
 	return hadCache, entry
 }
@@ -380,7 +397,9 @@ func (r *LocalQueueReconciler) reconcileConsumedUsage(ctx context.Context, lq *k
 			log.V(2).Info("Failed to reset LocalQueue status", "namespace", lq.Namespace, "name", lq.Name, "error", err)
 			return err
 		}
-		r.queues.AfsConsumedResources.Set(lqKey, corev1.ResourceList{}, now)
+		r.queues.AfsConsumedResources.Update(lqKey, func(old queueafs.ConsumedResourcesEntry, found bool) queueafs.ConsumedResourcesEntry {
+			return queueafs.ConsumedResourcesEntry{Resources: corev1.ResourceList{}, LastUpdate: now, StatusAccounted: old.StatusAccounted || !found}
+		})
 		log.V(2).Info("Reset AFS consumed resources cache", "namespace", lq.Namespace, "name", lq.Name)
 		return nil
 	}
@@ -401,8 +420,30 @@ func (r *LocalQueueReconciler) reconcileConsumedUsage(ctx context.Context, lq *k
 		log.V(2).Info("Failed to update LocalQueue status", "namespace", lq.Namespace, "name", lq.Name, "error", err)
 		return err
 	}
-	r.queues.AfsConsumedResources.Set(lqKey, newConsumed, now)
-	log.V(2).Info("Updated AFS consumed resources cache", "namespace", lq.Namespace, "name", lq.Name, "consumedResources", newConsumed)
+	// Recompute inside the lock rather than storing newConsumed: a settlement can
+	// land between the Get above and this write (the status update is an API call),
+	// and its folded penalty must not be overwritten. The persisted status may lag
+	// the stored value by one interval in that case; the next tick converges it.
+	stored := r.queues.AfsConsumedResources.Update(lqKey, func(old queueafs.ConsumedResourcesEntry, found bool) queueafs.ConsumedResourcesEntry {
+		if !found {
+			return queueafs.ConsumedResourcesEntry{Resources: newConsumed, LastUpdate: now, StatusAccounted: true}
+		}
+		// A settlement during the status update above can stamp a LastUpdate
+		// later than now; clamp the elapsed time so alpha stays within [0, 1],
+		// and keep the stored timestamp monotonic so the next decay does not
+		// over-elapse.
+		elapsed := max(0, now.Sub(old.LastUpdate).Seconds())
+		storedLastUpdate := now
+		if old.LastUpdate.After(now) {
+			storedLastUpdate = old.LastUpdate
+		}
+		return queueafs.ConsumedResourcesEntry{
+			Resources:       afs.CalculateDecayedConsumed(old.Resources, newUsage, elapsed, halfLifeTime),
+			LastUpdate:      storedLastUpdate,
+			StatusAccounted: old.StatusAccounted,
+		}
+	})
+	log.V(2).Info("Updated AFS consumed resources cache", "namespace", lq.Namespace, "name", lq.Name, "consumedResources", stored.Resources)
 	return nil
 }
 
--- a/pkg/controller/core/localqueue_controller.go
+++ b/pkg/controller/core/localqueue_controller.go
@@ -353,20 +353,22 @@ func (r *LocalQueueReconciler) initializeAfsIfNeeded(lq *kueue.LocalQueue) (hadC
 	}
 
 	if !hadCache {
-		currentUsage := r.getCurrentUsageForLocalQueue(lq.Spec.ClusterQueue, lqKey)
-		r.queues.AfsConsumedResources.Set(lqKey, currentUsage, now)
-		entry = queueafs.ConsumedResourcesEntry{Resources: currentUsage, LastUpdate: now}
+		// Seed only from persisted state, never from live admitted usage: counting a
+		// concurrently-admitted Workload here would double it against its still-pending
+		// entry penalty, inflating the LocalQueue and inverting preemption ordering
+		// (#12783). Stamping LastUpdate to now keeps the first sampling tick at ~zero
+		// elapsed so it folds no live usage, leaving the seed independent of any
+		// in-flight admission; the EMA converges over later ticks. An empty seed is
+		// correct for a fresh LocalQueue: its entry penalty accounts for early admissions.
+		seeded := corev1.ResourceList{}
+		if hasStatus {
+			seeded = lq.Status.FairSharing.AdmissionFairSharingStatus.ConsumedResources.DeepCopy()
+		}
+		r.queues.AfsConsumedResources.Set(lqKey, seeded, now)
+		entry = queueafs.ConsumedResourcesEntry{Resources: seeded, LastUpdate: now}
 	}
 
 	return hadCache, entry
-}
-
-func (r *LocalQueueReconciler) getCurrentUsageForLocalQueue(cqName kueue.ClusterQueueReference, lqKey utilqueue.LocalQueueReference) corev1.ResourceList {
-	cacheLq, err := r.cache.GetCacheLocalQueue(cqName, lqKey)
-	if err != nil {
-		return corev1.ResourceList{}
-	}
-	return cacheLq.GetAdmittedUsage()
 }
 
 func (r *LocalQueueReconciler) reconcileConsumedUsage(ctx context.Context, lq *kueue.LocalQueue) error {
--- a/pkg/controller/core/localqueue_controller.go
+++ b/pkg/controller/core/localqueue_controller.go
@@ -342,7 +342,6 @@ func (r *LocalQueueReconciler) initializeAfsIfNeeded(lq *kueue.LocalQueue) (hadC
 
 	lqKey := utilqueue.Key(lq)
 	hasStatus := lq.Status.FairSharing.AdmissionFairSharingStatus != nil
-	entry, hadCache = r.queues.AfsConsumedResources.Get(lqKey)
 
 	now := r.clock.Now()
 
@@ -352,21 +351,20 @@ func (r *LocalQueueReconciler) initializeAfsIfNeeded(lq *kueue.LocalQueue) (hadC
 		}
 	}
 
-	if !hadCache {
-		// Seed only from persisted state, never from live admitted usage: counting a
-		// concurrently-admitted Workload here would double it against its still-pending
-		// entry penalty, inflating the LocalQueue and inverting preemption ordering
-		// (#12783). Stamping LastUpdate to now keeps the first sampling tick at ~zero
-		// elapsed so it folds no live usage, leaving the seed independent of any
-		// in-flight admission; the EMA converges over later ticks. An empty seed is
-		// correct for a fresh LocalQueue: its entry penalty accounts for early admissions.
-		seeded := corev1.ResourceList{}
-		if hasStatus {
-			seeded = lq.Status.FairSharing.AdmissionFairSharingStatus.ConsumedResources.DeepCopy()
-		}
-		r.queues.AfsConsumedResources.Set(lqKey, seeded, now)
-		entry = queueafs.ConsumedResourcesEntry{Resources: seeded, LastUpdate: now}
+	// Seed only from persisted state, never from live admitted usage: a live
+	// snapshot can include a concurrently-admitted Workload that its still-pending
+	// entry penalty already prices, double-counting it (#12783). An empty seed is
+	// correct for a fresh LocalQueue: entry penalties cover its early admissions.
+	// LastUpdate=now keeps the first sampling tick at ~zero elapsed, so it folds
+	// no live usage and the seed stays independent of in-flight admissions.
+	// SetIfAbsent lets a concurrently settled entry win over the seed; settlement's
+	// own read-modify-write may still replace a fresh seed (bounded history loss,
+	// never a double count).
+	seeded := corev1.ResourceList{}
+	if hasStatus {
+		seeded = lq.Status.FairSharing.AdmissionFairSharingStatus.ConsumedResources.DeepCopy()
 	}
+	entry, hadCache = r.queues.AfsConsumedResources.SetIfAbsent(lqKey, seeded, now)
 
 	return hadCache, entry
 }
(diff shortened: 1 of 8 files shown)


apullo777:
Opened this to make the seed-from-status direction concrete. Sharing here in case it is useful for the ongoing discussion.

cc @mimowo @kshalot

pkg/controller/core/localqueue_controller.go
```diff
@@ -353,22 +353,24 @@ func (r *LocalQueueReconciler) initializeAfsIfNeeded(lq *kueue.LocalQueue) (hadC
...
+		// entry penalty, inflating the LocalQueue and inverting preemption ordering
+		// (#12783). Stamping LastUpdate to now keeps the first sampling tick at ~zero
+		// elapsed so it folds no live usage, leaving the seed independent of any
+		// in-flight admission; the EMA converges over later ticks. An empty seed is
+		// correct for a fresh LocalQueue: its entry penalty accounts for early admissions.
+		seeded := corev1.ResourceList{}
+		if hasStatus {
+			seeded = lq.Status.FairSharing.AdmissionFairSharingStatus.ConsumedResources.DeepCopy()
+		}
+		r.queues.AfsConsumedResources.Set(lqKey, seeded, now)
```
mimowo: This looks quite reaosnable, let me just ask a sanity check if we don't have a TOCTOU problem here, what if `r.queues.AfsConsumedResources` is set by another path (maybe it cannot?). In that case instead of set we may have a function like `SetOrAdd` (or just `Update`.

kshalot: I was wondering about the exact same thing. This is also set from the workload controller when a workload transitions from pending to admitted:
https://github.com/kubernetes-sigs/kueue/blob/6c0e4cde0090247ce6b75b0075beb50b66b058d5/pkg/controller/core/workload_controller.go#L1438

So in theory at least, there might be a TOCTOU

mimowo:
LGTM overall, qq from description:

> Related to #12783

Why not "Fixes", is it because we need more test time to confirm, or it will not fix "in principle"?

pkg/controller/core/localqueue_controller.go
kshalot: Now that I look at it, it might be a problem during restarts.

If the WL Reconcile runs first, the LQ will then see `hadCache` as `true`. This will cause us to skip the seeding phase and we'll lose the historic usage. But this was also the case before this fix, we can treat it as a separate change probably.

mimowo: How we could solve it? Maybe `AfsConsumedResources` should keep a boolean like `seededFromStatus` to avoid accounting twice? Since this PR claims to solve the restart problem maybe worth combining.

We can test the restart (and try to repro Karol's idea for the bug) in integration tests. We already have a number of tests which actually restart Kueue controller manager.

apullo777: @mimowo  I think we can make the seed path idempotent without adding another state bit.

I will change this to SetIfAbsent / LoadOrStore-style semantics for `AfsConsumedResources`: seed only if no entry exists, otherwise leave the existing entry untouched.

That should handle both interleavings: if settlement wins first, its entry is newer and the seed backs off; if the seed wins first, settlement proceeds on top of the seeded entry as usual.

So I think we do not need a `seededFromStatus` flag here?

apullo777: @kshalot Agreed. That restart-recovery case predates this fix, so I think it is better as a follow-up.

With the SetIfAbsent change, the behavior becomes explicit: if the workload-controller path creates the entry first, that entry wins and the persisted warm-start is skipped. 

That is still on the under-count / convergence side, and depends on restart + reconcile ordering.

I will keep this PR focused on avoiding the live-usage seed over-count, and can track guaranteed warm-start recovery separately.

apullo777:
> Why not "Fixes", is it because we need more test time to confirm, or it will not fix "in principle"?

It should fix #12783 in principle. I used `Related` mostly out of caution thinking there might be adjacent follow-ups, but I think this PR removes the live-usage seed source behind this flake.

I will update it to `Fixes #12783` :)

pkg/controller/core/localqueue_controller.go
kshalot: > That restart-recovery case predates this fix, so I think it is better as a follow-up.

Honestly, I'm fine with either.

> With the SetIfAbsent change, the behavior becomes explicit: if the workload-controller path creates the entry first, that entry wins and the persisted warm-start is skipped.

I think I'm not grasping one part yet:
1. LocalQueue has a historical EMA of 100 CPU before the restart.
2. Workload A is admitted, it requires 10 CPU.
4. Workload Controller handles admission, it defaults `oldUsage` to 0 because the cache is not seeded yet. So it sets it to `0 + penalty`.
5. LocalQueue Controller is finally triggered. It uses `SetIfAbsent`, which skips the seeding because the cache already exists.
6. So the AFS consumed usage cache stays at `penalty`, while I think we would expect it to rather be `100 + penalty`.

I think @mimowo's idea of `seededFromStatus` solves this, because we could use it to essentially do `historicalUsageSeed + r.queues.AfsConsumedResources.Get(lqKey)`. Or we could make the workload controller apply the seed itself when it encounters a cache miss, but that seems worse to me.

mimowo: Or better name would be `statusAccounted` even if it was seeded by workload controller, then the LQ controller may want to still add the status and record `statusAccounted` to avoid re-doing.

mimowo: > Honestly, I'm fine with either.

I'm ok too, but I'm wondering - let's try to get to the bottom of both problems first, because maybe fixing one impact the approach for the other.

apullo777: @kshalot Yes, that scenario is exactly the restart-recovery boundary I had in mind. If the persisted EMA is 100 and the workload-controller path creates the entry first, then with the current SetIfAbsent shape we would keep only the post-restart penalty entry, not `100 + penalty`.

I agree the desired recovery shape is to merge the persisted status with the existing entry. The first settlement on a cache miss is fine: elapsed is zero, so it folds no live usage and the entry starts as just the penalty. One caveat is a second settlement on the same LQ before its first LQ reconcile: that can fold a small alpha-weighted slice of live admitted usage, which after restart may include relisted pre-restart workloads and overlap the persisted status. That overlap should be bounded by the short pre-reconcile window, but it is worth calling out in the merge comment.

@mimowo I also agree the two problems interact. Re-reading your suggestion, I think `statusAccounted` is the right tool for the recovery problem. My earlier "no flag needed" point was only about the seed-overwrite race, where insert-if-absent is enough.

Once we merge persisted history into an existing entry, the entry's existence no longer tells us whether that history has already been applied, so we need to record that explicitly. One thing to watch is that every writer of `AfsConsumedResources` then needs to preserve the flag. Today settlement and sampling both build a fresh entry when they `Set`, so a flag stored on the entry could be dropped and the persisted status could be merged again later. Since sampling also writes the merged value back to status, repeated re-merges could compound rather than just add a fixed offset.

So I think the cleaner shape is probably an `Update(lqKey, func(entry) entry)` primitive for `AfsConsumedResources`: seed/merge/status-accounted updates would happen under one lock, and other writers would preserve the extra state. One constraint: the `Update` closure should not call back into the scheduler cache, to avoid lock ordering issues. Live usage should be read before taking the entry lock, with only the fold/update done inside. SetIfAbsent would just be the simpler insert-if-absent case of that primitive.

Scope-wise, I’m happy either way: keep this PR focused on the #12783 seed over-count and do the guaranteed warm-start / `Update` rework as a follow-up, or fold the full design into this PR.

kshalot: If we are confident that both issues are orthogonal, then IMO making this a separate PR makes sense, but I'm leaving it up to you @apullo777. If you choose to extract it, maybe it would also be good to write a quick summary of this in a separate issue, if you have the time. This way it would be little easier to track what is what for people in the future.

mimowo: I also dont have an opinion - both are conceptually related bugfixes and both need cherrypicking. otoh If there is no code overlap Im good to separate.

One note is that we do patch release tomorrow and it would be nice to include them.

apullo777: @mimowo @kshalot pushed the rework discussed above.

On closer look there was real code overlap: the warm-start recovery fix generalizes the seed-race fix. So I folded the recovery work into this PR, but the last commit is still self-contained enough to extract if you prefer.

The PR now has three commits:

1. seed the AFS cache from persisted LocalQueue status instead of live usage
2. add the SetIfAbsent seed hardening discussed above
3. add the shared `Update` path + `statusAccounted` recovery rework

In the last commit, the known `AfsConsumedResources` writers (seed, settlement, sampling tick, and half-life reset) update the entry under one lock and preserve `statusAccounted`. The seed can now merge persisted history into a settlement-created entry exactly once per controller process; after a restart, the fresh cache is intentionally eligible to seed/merge from status again.

Two behavior notes:
- absent entries initialize `statusAccounted` differently by design: settlement creates `false` because persisted history may still be owed, while seed and sampling create `true`
- if settlement interleaves with the sampling tick's status update, status can lag the in-memory entry by up to one sampling interval before converging. A restart in that window restores the lagging value, which is a bounded under-count; previously that interleave could drop the settled penalty permanently.

This keeps the scope to the consumed-resources cache RMW races; the broader cross-ledger atomicity between consumed resources and entry penalties remains a separate follow-up.

I also added unit coverage for the restart-merge case and the no-re-merge case. Live usage is read before taking the entry lock, per the lock-ordering constraint.

Also opened #12911 to track the warm-start recovery bug for release notes / cherry-pick bookkeeping.

pkg/controller/core/localqueue_controller.go
```diff
@@ -352,23 +352,40 @@ func (r *LocalQueueReconciler) initializeAfsIfNeeded(lq *kueue.LocalQueue) (hadC
...
+	// no live usage and the seed stays independent of in-flight admissions.
+	// If settlement created the entry first (post-restart admission), the persisted
+	// history is merged into it exactly once per process, tracked by StatusAccounted.
+	// The merge can slightly over-count: a second settlement before this reconcile
+	// folds alpha(elapsed) of live usage, which after a restart overlaps the persisted
+	// status; bounded by the short pre-reconcile window and decaying within a half-life.
+	seeded := corev1.ResourceList{}
+	if hasStatus {
+		seeded = lq.Status.FairSharing.AdmissionFairSharingStatus.ConsumedResources.DeepCopy()
+	}
```
mimowo, line 369: It would be great to add an integration test for the restart scenario

pkg/controller/core/localqueue_controller.go
mimowo: Cool, please reflect on both behavior fixes in the release note, prefix that with "AFS:" or "AdmissionFairSharing:"

test/integration/singlecluster/scheduler/fairsharing/fair_sharing_test.go
```diff
@@ -1426,6 +1426,86 @@ var _ = ginkgo.Describe("Scheduler", ginkgo.Label("feature:fairsharing", "featur
 		})
 	})
 
+	ginkgo.When("Restarting the manager", func() {
```
mimowo, line 1429: Can you try to loop the test 30 times locally to confirm no flakes?

apullo777: Done. looped the focused integration test 30 times locally, 30/30 green with no flakes (8m40s total):

```
INTEGRATION_FILTERS='--focus=retain.the.LocalQueue.s.historical.usage' \
GINKGO_ARGS='--repeat=29' \
make test-integration
```

pkg/controller/core/workload_controller.go
```diff
@@ -1418,28 +1419,40 @@ func (r *WorkloadReconciler) updateAfsConsumedUsage(log logr.Logger, wl *kueue.W
...
+			lastUpdate = now
+		}
+		// A concurrent writer can stamp a LastUpdate later than now; clamp the
+		// elapsed time so alpha stays within [0, 1], and keep the stored
+		// timestamp monotonic so the next decay does not over-elapse.
+		elapsed := max(0, now.Sub(lastUpdate).Seconds())
+		storedLastUpdate := now
+		if lastUpdate.After(now) {
+			storedLastUpdate = lastUpdate
+		}
```
mimowo, line 1444: IIUC these are a bit of drive by fix? Do we have a unit test for that?

mimowo (commented):
/lgtm
/approve
This is an excellent contribution @apullo777, I'm impressed 🥇

Please prepare the CPs.

@kshalot I decide to the PR merge today as we release tomorrow. I reviewed the code and haven't noticed any blockers. However, feel free to provide feedback post-merge. I'm sure it will get addressed,

mimowo:
This is known unrelated flake

mimowo:
That is a known issue which is ticketed

pkg/controller/core/workload_controller.go
apullo777: Ah, I think partly yes. The tick-side guard came from this PR's `Update` rework, but the settlement-side guard is a drive-by.

I will open a small follow-up to add that, and also cover the same clamp for the tick's pre-lock status computation.

kshalot:
> /lgtm /approve This is an excellent contribution @apullo777, I'm impressed 🥇
> 
> Please prepare the CPs.
> 
> @kshalot I decide to the PR merge today as we release tomorrow. I reviewed the code and haven't noticed any blockers. However, feel free to provide feedback post-merge. I'm sure it will get addressed,

@mimowo It looks good to me! 

@apullo777 Thanks for the change!

PR #12893, opened 2026-07-09
Fix: Make race condition fails more descriptive in singlecluster tests

### LINKED ISSUE #8070: Make race condition fails more descriptive

Currently, race condition errors look like this:

```
There were failures detected in the following suites:
  dra ./test/integration/singlecluster/controller/dra
```

This is very unclear and doesn’t show what’s actually happening.

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/8063/pull-kueue-test-integration-baseline-main/1996549770306392064

**Why is this needed**:

To display the full trace for race conditions.

vic-comm:
What type of PR is this?
/kind cleanup /area testing

This PR adds the --output-interceptor-mode=none flag to the Ginkgo command in the test-integration-run Makefile target (used for singlecluster integration tests).

Currently, when the Go race detector triggers during these tests, it crashes the test process abruptly. Because Ginkgo's output interceptor is enabled by default, the race detector's stack trace gets swallowed in the crash, leaving developers with a generic and unhelpful suite failure message (e.g., "There were failures detected in the following suites...").

Disabling the output interceptor forces the race trace to be written directly to the terminal's standard error stream, ensuring it is always visible in CI/local logs. This makes debugging race conditions significantly easier and matches the behavior already applied to the test-multikueue-integration target.

Which issue(s) this PR fixes:
Fixes #8070

None

```release-note
NONE
```

## Summary by CodeRabbit

* **Chores**
  * Updated the integration test run configuration to disable output interception for single-cluster integration tests, improving consistency during execution.

### DIFF
--- a/Makefile-test.mk
+++ b/Makefile-test.mk
@@ -137,7 +137,7 @@ test-integration-run:
 	ENVTEST_K8S_VERSION=$(ENVTEST_K8S_VERSION) \
 	ARTIFACTS=$(ARTIFACTS) \
 	TEST_LOG_LEVEL=$(TEST_LOG_LEVEL) API_LOG_LEVEL=$(INTEGRATION_API_LOG_LEVEL) \
-	$(GINKGO) $(INTEGRATION_FILTERS) $(GINKGO_ARGS) $(GOFLAGS) -procs=$(INTEGRATION_NPROCS) --race --json-report=integration.json --output-dir=$(ARTIFACTS) -v $(INTEGRATION_TARGET)
+	$(GINKGO) $(INTEGRATION_FILTERS) $(GINKGO_ARGS) $(GOFLAGS) -procs=$(INTEGRATION_NPROCS) --race --json-report=integration.json --output-interceptor-mode=none --output-dir=$(ARTIFACTS) -v $(INTEGRATION_TARGET)
 	$(BIN_DIR)/ginkgo-top -i $(ARTIFACTS)/integration.json > $(ARTIFACTS)/integration-top.yaml
 
 .PHONY: test-integration-baseline


mimowo:
cc @mbobrovskyi ptal

mimowo:
cc @mbobrovskyi

Makefile-test.mk
mbobrovskyi, line 160: Actually, we need to handle this somehow.

Makefile-test.mk
```diff
@@ -137,7 +137,7 @@ test-integration-run:
 	ENVTEST_K8S_VERSION=$(ENVTEST_K8S_VERSION) \
 	ARTIFACTS=$(ARTIFACTS) \
 	TEST_LOG_LEVEL=$(TEST_LOG_LEVEL) API_LOG_LEVEL=$(INTEGRATION_API_LOG_LEVEL) \
-	$(GINKGO) $(INTEGRATION_FILTERS) $(GINKGO_ARGS) $(GOFLAGS) -procs=$(INTEGRATION_NPROCS) --race --json-report=integration.json --output-dir=$(ARTIFACTS) -v $(INTEGRATION_TARGET)
+	$(GINKGO) $(INTEGRATION_FILTERS) $(GINKGO_ARGS) $(GOFLAGS) -procs=$(INTEGRATION_NPROCS) --race --json-report=integration.json --output-interceptor-mode=none --output-dir=$(ARTIFACTS) -v $(INTEGRATION_TARGET)
```
mbobrovskyi, line 140: If I remember correctly, it's actually the other way around: `--output-interceptor-mode=none` disables the race detector logs.

vic-comm: I think --output-interceptor-mode=none is the needed fix, but testing it locally is a bit tricky due to how Ginkgo v2 and halt_on_error interact.

Here's what i observed when testing with -procs=4:

1- Interceptor ON (default): The race trace writes to an internal pipe. If the process crashes abruptly (exit 66), the pipe is destroyed before it can flush over RPC. The logs are permanently lost.
2- Interceptor OFF (none): The trace writes directly to the Ginkgo coordinator's in-memory buffer instead.
The local testing catch: Locally, Go defaults to halt_on_error=0. The worker prints the warning, finishes the tests, sends a SuiteDone message, and then exits with 66. Since the coordinator receives SuiteDone, it skips its timeout logic and never prints the buffer (which makes it look like the logs are disabled).
   However, if you test it with GORACE="halt_on_error=1", the process crashes instantly. The coordinator never gets the SuiteDone message, hits a timeout, and correctly dumps the buffer to the terminal, revealing the full stack trace!

mbobrovskyi:
/release-note-edit
```release-note
NONE
```

mbobrovskyi:
Thank you!

/lgtm
/approve

Makefile-test.mk
mbobrovskyi: Sounds reasonable.

PR #12897, opened 2026-07-09
dra: reject device counts whose sum overflows int64

### LINKED ISSUE #12896: DRA device count overflow admits over-quota Workloads and corrupts ClusterQueue used-quota
### What happened

With the DRA integration enabled (on by default), a ResourceClaimTemplate whose device requests use very large `count` values causes Kueue's per-device-class count to overflow int64 to a negative value. The Workload is admitted even though it far exceeds quota, and the ClusterQueue's used-quota for that resource is left negative.

### Reproduction (kueue v0.18.2, verified on a live cluster)

1. DRA integration on; a `resources.deviceClassMappings` entry maps a DeviceClass to a quota resource; a ClusterQueue with `nominalQuota` for it (e.g. `gpu.nvidia.com: 4`).
2. Create a ResourceClaimTemplate with two requests, each `exactly.count: 9223372036854775807` (MaxInt64), same DeviceClass, `allocationMode: ExactCount`.
3. Create a suspended Job referencing it, labeled with the queue.

Observed: the Workload is **Admitted**; `ClusterQueue.status.flavorsUsage[].resources[<res>].total = -2` (MaxInt64 + MaxInt64 wraps to -2 in int64); the controller logs `resourceUsage {"<res>":"-2"}` with no warning.

### Root cause

`pkg/dra/claims.go` `countDevicesPerClass` does `out[dc] += q` (`q = req.Exactly.Count`, int64) with no overflow guard, and there is no upper bound on `count` (the apiserver only requires `count > 0`).

### Impact

I checked the enforcement paths and they are otherwise sound: with the negative usage present, a legitimate over-quota Workload still stayed Pending (both direct admission and cohort borrowing correctly rejected it), so I did not find a way to over-admit other workloads or steal resources. So this is a correctness/robustness issue (a malformed claim is admitted and the used-quota shown in the ClusterQueue status is corrupted to a negative value), not a demonstrated quota bypass for real workloads.

### Fix

Accumulate the per-device-class counts with a saturating add (matching the `Amount` arithmetic the scheduler and cache already use) so the sum caps at MaxInt64 instead of wrapping negative. See #12897.

### Environment
- Kueue v0.18.2, `KueueDRAIntegration` enabled (default).

thc1006:
#### What type of PR is this?

/kind bug

`countDevicesPerClass` sums per-request device counts into an int64 with no overflow guard, and device counts are unbounded (the apiserver only requires `count > 0`, so it accepts up to MaxInt64). A ResourceClaimTemplate whose counts sum past MaxInt64 wraps to a negative value: the quota usage goes negative, which admits a Workload that exceeds quota and leaves a negative used-quota in the ClusterQueue status.

This accumulates with `utilmath.SaturatingAdd` (matching the `Amount` arithmetic the scheduler and cache already use), so the sum saturates at MaxInt64 instead of wrapping negative. A huge count then stays a large positive that cannot fit quota, rather than being admitted on a negative. The count is turned into a `resource.Quantity` right after this, so the saturated value is handled by overflow-safe Quantity arithmetic downstream.

A unit test covers both the normal sum and the saturation case.

Fixes #12896

Switched from an explicit validation error to `SaturatingAdd` per review feedback, to match the resilient `Amount` arithmetic used across the scheduler and cache. The extended-resource path already uses Quantity arithmetic and is unaffected.

```release-note
DRA: Fix an integer overflow in device-count quota accounting where a ResourceClaimTemplate with very large device counts could be admitted over quota and leave a negative used-quota in the ClusterQueue status.
```

### DIFF
--- a/pkg/dra/claims.go
+++ b/pkg/dra/claims.go
@@ -34,6 +34,7 @@ import (
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
 	"sigs.k8s.io/kueue/pkg/features"
 	"sigs.k8s.io/kueue/pkg/resources"
+	utilmath "sigs.k8s.io/kueue/pkg/util/math"
 	utilresource "sigs.k8s.io/kueue/pkg/util/resource"
 )
 
@@ -109,7 +110,11 @@ func countDevicesPerClass(claimSpec *resourcev1.ResourceClaimSpec) (resources.Re
 		if dc == "" {
 			continue
 		}
-		out[dc] += q
+		// Device counts are user-controlled and effectively unbounded (the
+		// apiserver accepts up to MaxInt64), so accumulate with a saturating add
+		// (matching the scheduler's Amount arithmetic) rather than letting the
+		// sum wrap to a negative count.
+		out[dc] = utilmath.SaturatingAdd(out[dc], q)
 	}
 	return out, nil
 }
--- a/pkg/dra/claims_test.go
+++ b/pkg/dra/claims_test.go
@@ -18,6 +18,7 @@ package dra
 
 import (
 	"errors"
+	"math"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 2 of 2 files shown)


pkg/dra/claims.go
```diff
@@ -109,6 +110,13 @@ func countDevicesPerClass(claimSpec *resourcev1.ResourceClaimSpec) (resources.Re
 		if dc == "" {
 			continue
 		}
+		// Device counts are user-controlled and effectively unbounded (the
+		// apiserver accepts up to MaxInt64), so guard the running sum against
+		// int64 overflow instead of letting it wrap to a negative count.
+		if q > math.MaxInt64-out[dc] {
```
mimowo: Wdyt about using the "Amount" structure we use for scheduling. This strucutre instead of validation error caps at int64. I'm ok with both, but maybe worth to consider using Amount more in DRA code for better resilience instead of checking individual operations, wdyt @sohankunkerkar ?

Validation is also ok, just a matter of preference, because values near 64 don't have much of practical use anyway.

thc1006: Good point. `SaturatingAdd` is what the scheduler and cache code already use everywhere via `Amount`, and `countDevicesPerClass` is really the outlier still doing raw int64. Aligning it with that pattern is cleaner than a one-off validation error, and I agree counts near MaxInt64 have no practical use.

I've switched the accumulation to `utilmath.SaturatingAdd(out[dc], q)` and updated the test to assert it saturates rather than wraps negative. Happy to go back to the validation error if you or @sohankunkerkar would rather keep an explicit signal.

mimowo: Yeah, both approaches have some merit, but I think Amount is ok and more consistent with the rest of the codebase

mimowo:
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for hardening Kueue 👍

pkg/dra/claims.go
thc1006: Thanks for confirming. The latest revision (f0fef9b2c) already accumulates with SaturatingAdd, which is the same helper that Amount.AddInt64 wraps, so countDevicesPerClass now uses the scheduler saturating arithmetic while keeping the Requests int64 return type unchanged. If you would prefer this call site to use the Amount type directly, I am glad to switch it.

mimowo:
/release-note-edit
```release-note
DRA: Fix an integer overflow in device-count quota accounting where a ResourceClaimTemplate with very large device counts could be admitted over quota and leave a negative used-quota in the ClusterQueue status.
```

PR #12898, opened 2026-07-09
Force delete pods in TAS integration suite to improve cleanup

kshalot:
#### What type of PR is this?
/kind cleanup
/area tas
/area was
/area testing

In https://github.com/kubernetes-sigs/kueue/pull/12256, I added a wrapper to run each TAS integration test twice - once with `scheduler-library` integration enabled, once with it disabled.

There is a problem in the current suite, in that it never actually deletes the pods after running `util.DeleteNamespace`. This is because in `envtest` there is no kubelet, so nothing is actually cleaning up the pod objects.

Fixes #

```release-note
NONE
```

### DIFF
--- a/test/integration/singlecluster/tas/tas_test.go
+++ b/test/integration/singlecluster/tas/tas_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package tas
 
 import (
+	"context"
 	"fmt"
 	"slices"
 	"time"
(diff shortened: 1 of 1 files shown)


test/integration/singlecluster/tas/tas_test.go
```diff
@@ -95,6 +96,20 @@ func createPodsForWorkload(wl *kueue.Workload, nsName string, withTopologyReques
 	})
 }
 
+func forceDeleteNamespace(ctx context.Context, c client.Client, ns *corev1.Namespace) error {
```
kshalot, line 99: I also considered making this a util (but I'm not sure how useful it will be) or injecting this logic into `util.DeleteNamespace` (but I don't want to affect other tests).

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for the cleanup 👍

test/integration/singlecluster/tas/tas_test.go
mimowo: Let's consider that for the future, I'm ok scoping this where needed  for now. I think long term we may have some options to util.DeleteNamespace, like {force:true}

Issue #8079, opened 2025-12-04
CornerCase: Kueue blocks scheduling of workloads if they request "0" or resource not present in CQ definition


mimowo:
**What happened**:

Kueue would not admit workloads which request "0" of a resource which is not defined in the ClusterQueue, for example

**What you expected to happen**:

Kueue would admit workloads which request "0" of some resource, even if not declared in CQ.

I expect that because the vanilla kubernetes schedules Pods which request 0 nvidia.com/gpu, even if Node does not provide GPU.

**How to reproduce it (as minimally and precisely as possible)**:

```yaml
apiVersion: kueue.x-k8s.io/v1beta1
kind: ClusterQueue
metadata:
  name: "cluster-queue"
spec:
  namespaceSelector: {} # match all.
  resourceGroups:
  - coveredResources: ["memory"]
    flavors:
    - name: "default-flavor"
      resources:
      - name: "memory"
        nominalQuota: 36Gi
```
and Job 
```yaml
apiVersion: batch/v1
kind: Job
metadata:
  generateName: sample-job
  namespace: default
  labels:
    kueue.x-k8s.io/queue-name: user-queue
spec:
  parallelism: 3
  completions: 3
  completionMode: Indexed
  template:
    spec:
      containers:
      - name: dummy-job
        image: gcr.io/k8s-staging-perf-tests/sleep:v0.1.0
        args: ["10s"]
        resources:
          limits:
            "nvidia.com/gpu": 0
          requests:
            "nvidia.com/gpu": 0
            memory: "200Mi"
      restartPolicy: Never
```
Will fail workload scheduling with:

```yaml
  status:
    conditions:
    - lastTransitionTime: "2025-12-04T16:23:57Z"
      message: 'couldn''t assign flavors to pod set main: resource nvidia.com/gpu
        unavailable in ClusterQueue'
      observedGeneration: 1
      reason: Pending
      status: "False"
      type: QuotaReserved
```

mimowo:
cc @tenzen-y @kannon92  as I may lack context if this is a bug or a deliberate decision, but I think we should aim for compatibility with the kube-scheduler, wdyt?

kannon92:
This reminds me of https://github.com/kubernetes-sigs/kueue/issues/7513.

kannon92:
0 value is weird but typically Kueue blocks on anything that isn't in CQ or excludeResourcePrefixes.

[...]

PR #12924, opened 2026-07-09
Correct Flavor Assignment in Pod Importer

mszadkow:
#### What type of PR is this?
/kind bug

This PR improves behavior of Flavor Assignment in Pod Importer and test coverage for additional resources.

It introduces structured error handling for uncovered resources, keeps resource-to-flavor selection deterministic, and consolidates importer tests into a cleaner table-driven structure.

```release-note
Importer: Fixed a bug where the Pod importer picked a single ResourceFlavor for the whole Pod, so Pods whose resources map to different flavors could be imported with a wrong flavor assignment. Flavors are now resolved per requested resource.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Pod imports now assign admission flavors per requested resource, including mixed CPU/GPU workloads.
  * Imports reject missing flavors, uncovered resources, conflicting queue labels, empty requests, and unknown priority classes.
  * Invalid workloads no longer create resources or mutate the original Pod.
  * Admission retries now handle conflicts and cancellation reliably, with terminal errors after repeated failures.
  * Pod validation consistently checks queues, flavors, resource coverage, workload requests, and priority metadata.

* **Documentation**
  * Updated importer validation guidance and the local dry-run command.

* **Tests**
  * Added coverage for GPU workloads, queue conflicts, priority classes, missing flavors, uncovered resources, and deterministic errors.

### DIFF
--- a/cmd/importer/README.md
+++ b/cmd/importer/README.md
@@ -30,8 +30,9 @@ The importer will perform following checks:
 - The target LocalQueue exists.
 - The LocalQueues involved in the import are using an existing ClusterQueue.
 - The ClusterQueues involved have ResourceGroups that reference existing ResourceFlavors.
-- For each Pod, the check validates that every non-zero requested resource is covered by the target ClusterQueue ResourceGroups.
 - For each Pod, the check validates that Workload construction succeeds (using the same construction path used by import) before any Pod mutation.
+- For each Pod, the check validates that every non-zero resource request in the constructed Workload is covered by the target ClusterQueue ResourceGroups.
+- Pass the configured prefixes with `--exclude-resource-prefixes` so excluded resources are ignored during validation and admission.
 - If a Pod specifies a PriorityClass, the check validates that the PriorityClass exists.
 
 There are two ways the mapping from a pod to a LocalQueue can be specified:
@@ -80,6 +81,7 @@ Usage:
 
 Flags:
       --add-labels stringToString     additional label=value pairs to be added to the imported pods and created workloads (default [])
+      --exclude-resource-prefixes strings  resource name prefixes ignored by Kueue during workload quota accounting
       --burst int                     client Burst, as described in https://kubernetes.io/docs/reference/config-api/apiserver-eventratelimit.v1alpha1/#eventratelimit-admission-k8s-io-v1alpha1-Limit (default 50)
   -c, --concurrent-workers uint       number of concurrent import workers (default 8)
       --dry-run                       don't import, check the config only (default true)
--- a/cmd/importer/README.md
+++ b/cmd/importer/README.md
@@ -30,9 +30,9 @@ The importer will perform following checks:
 - The target LocalQueue exists.
 - The LocalQueues involved in the import are using an existing ClusterQueue.
 - The ClusterQueues involved have ResourceGroups that reference existing ResourceFlavors.
-- For each Pod, the dry-run validates that every non-zero requested resource is covered by the target ClusterQueue ResourceGroups.
-- For each Pod, the dry-run validates that Workload construction succeeds (using the same construction path used by import) before any Pod mutation.
-- If a Pod specifies a PriorityClass, the dry-run validates that the PriorityClass exists.
+- For each Pod, the check validates that every non-zero requested resource is covered by the target ClusterQueue ResourceGroups.
+- For each Pod, the check validates that Workload construction succeeds (using the same construction path used by import) before any Pod mutation.
+- If a Pod specifies a PriorityClass, the check validates that the PriorityClass exists.
 
 There are two ways the mapping from a pod to a LocalQueue can be specified:
 
--- a/cmd/importer/README.md
+++ b/cmd/importer/README.md
@@ -29,7 +29,10 @@ The importer will perform following checks:
 - For every Pod a  mapping to a LocalQueue is available.
 - The target LocalQueue exists.
 - The LocalQueues involved in the import are using an existing ClusterQueue.
-- The ClusterQueues involved have at least one ResourceGroup using an existing ResourceFlavor. This ResourceFlavor is used when the importer creates the admission for the created workloads.
+- The ClusterQueues involved have ResourceGroups that reference existing ResourceFlavors.
+- For each Pod, the dry-run validates that every non-zero requested resource is covered by the target ClusterQueue ResourceGroups.
+- For each Pod, the dry-run validates that Workload construction succeeds (using the same construction path used by import) before any Pod mutation.
+- If a Pod specifies a PriorityClass, the dry-run validates that the PriorityClass exists.
 
 There are two ways the mapping from a pod to a LocalQueue can be specified:
 
@@ -172,7 +175,7 @@ Note: `dry-run` is set to `false` by default.
 
 5. (Optional) Check your config by dry-running it locally e.g.
 ```bash
-./bin/importer --dry-run=true <your-custom-flags> --queuemapping-file=cmd/importer/run-in-cluster/mapping.yaml
+./bin/importer import --dry-run=true <your-custom-flags> --queuemapping-file=cmd/importer/run-in-cluster/mapping.yaml
 ```
 
 6. Deploy the configuration:
--- a/cmd/importer/cache/cache.go
+++ b/cmd/importer/cache/cache.go
@@ -24,10 +24,12 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	schedulingv1 "k8s.io/api/scheduling/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
 	"sigs.k8s.io/kueue/cmd/importer/mapping"
+	"sigs.k8s.io/kueue/pkg/util/resourcegroups"
 	utilslices "sigs.k8s.io/kueue/pkg/util/slices"
 )
 
@@ -47,6 +49,10 @@ type ImportCache struct {
 	ResourceFlavors map[kueue.ResourceFlavorReference]*kueue.ResourceFlavor
 	PriorityClasses map[string]*schedulingv1.PriorityClass
 	AddLabels       map[string]string
+
+	// Derived from ClusterQueues and ResourceFlavors at Load time; read-only after that.
+	FlavorValidation  map[string]error
+	FlavorsByResource map[string]map[corev1.ResourceName]kueue.ResourceFlavorReference
 }
 
 func Load(ctx context.Context, c client.Client, namespaces []string, mappingRules mapping.Rules, addLabels map[string]string) (*ImportCache, error) {
@@ -84,10 +90,67 @@ func Load(ctx context.Context, c client.Client, namespaces []string, mappingRule
 		return nil, fmt.Errorf("loading priority classes: %w", err)
 	}
 	ret.PriorityClasses = utilslices.ToRefMap(pcList.Items, func(pc *schedulingv1.PriorityClass) string { return pc.Name })
+
+	ret.FlavorValidation = make(map[string]error, len(cqList.Items))
+	ret.FlavorsByResource = make(map[string]map[corev1.ResourceName]kueue.ResourceFlavorReference, len(cqList.Items))
+	for i := range cqList.Items {
+		cq := &cqList.Items[i]
+		rgs := resourceGroupsFrom(cq)
+		ret.FlavorsByResource[cq.Name] = flavorsByResourceFrom(rgs)
+		ret.FlavorValidation[cq.Name] = validateFlavors(cq.Name, rgs, ret.ResourceFlavors)
+	}
+
 	return &ret, nil
 }
 
-func (ic *ImportCache) LocalQueue(p *corev1.Pod) (*kueue.LocalQueue, bool, error) {
+// resourceGroupsFrom converts cq's API resource groups to the ResourceGroup
+// representation used by pkg/util/resourcegroups.
+func resourceGroupsFrom(cq *kueue.ClusterQueue) []resourcegroups.ResourceGroup {
+	rgs := make([]resourcegroups.ResourceGroup, 0, len(cq.Spec.ResourceGroups))
+	for _, rg := range cq.Spec.ResourceGroups {
+		if len(rg.Flavors) == 0 {
+			continue
+		}
+		flavors := make([]kueue.ResourceFlavorReference, len(rg.Flavors))
+		for i, f := range rg.Flavors {
+			flavors[i] = f.Name
+		}
+		rgs = append(rgs, resourcegroups.ResourceGroup{
+			CoveredResources: sets.New(rg.CoveredResources...),
+			Flavors:          flavors,
+		})
+	}
+	return rgs
+}
+
+// flavorsByResourceFrom returns, for every resource covered by rgs, the first
+// flavor listed in its resource group.
+func flavorsByResourceFrom(rgs []resourcegroups.ResourceGroup) map[corev1.ResourceName]kueue.ResourceFlavorReference {
+	m := make(map[corev1.ResourceName]kueue.ResourceFlavorReference)
+	for _, rg := range rgs {
+		for resource := range rg.CoveredResources {
+			if _, exists := m[resource]; !exists {
+				m[resource] = rg.Flavors[0]
+			}
+		}
+	}
+	return m
+}
+
+// validateFlavors checks that every ResourceFlavor referenced by rgs is known.
+func validateFlavors(cqName string, rgs []resourcegroups.ResourceGroup, resourceFlavors map[kueue.ResourceFlavorReference]*kueue.ResourceFlavor) error {
+	// Sorted for a deterministic error message; AllFlavors returns a set.
+	flavors := resourcegroups.AllFlavors(rgs).UnsortedList()
+	slices.Sort(flavors)
+	for _, flavor := range flavors {
+		if _, found := resourceFlavors[flavor]; !found {
+			return fmt.Errorf("%q flavor %q: %w", cqName, flavor, ErrCQInvalid)
+		}
+	}
+	return nil
+}
+
+func (ic *ImportCache) LocalQueueForPod(p *corev1.Pod) (*kueue.LocalQueue, bool, error) {
 	queueName, skip, found := ic.MappingRules.QueueFor(p.Spec.PriorityClassName, p.Labels)
 	if !found {
 		return nil, false, mapping.ErrNoMapping
@@ -107,17 +170,4 @@ func (ic *ImportCache) LocalQueue(p *corev1.Pod) (*kueue.LocalQueue, bool, error
 		return nil, false, fmt.Errorf("%s: %w", queueName, ErrLQNotFound)
 	}
 	return lq, false, nil
-}
-
-func (ic *ImportCache) ClusterQueue(p *corev1.Pod) (*kueue.ClusterQueue, bool, error) {
-	lq, skip, err := ic.LocalQueue(p)
-	if skip || err != nil {
-		return nil, skip, err
-	}
-	queueName := string(lq.Spec.ClusterQueue)
-	cq, found := ic.ClusterQueues[queueName]
-	if !found {
-		return nil, false, fmt.Errorf("cluster queue: %s: %w", queueName, ErrCQNotFound)
-	}
-	return cq, false, nil
 }
--- a/cmd/importer/cache/cache.go
+++ b/cmd/importer/cache/cache.go
@@ -35,6 +35,7 @@ var (
 	ErrLQNotFound = errors.New("localqueue not found")
 	ErrCQNotFound = errors.New("clusterqueue not found")
 	ErrCQInvalid  = errors.New("clusterqueue invalid")
+	ErrPodInvalid = errors.New("pod invalid")
 	ErrPCNotFound = errors.New("priorityclass not found")
 )
 
--- a/cmd/importer/cache/cache.go
+++ b/cmd/importer/cache/cache.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"slices"
 
 	corev1 "k8s.io/api/core/v1"
@@ -50,9 +51,9 @@ type ImportCache struct {
 	PriorityClasses map[string]*schedulingv1.PriorityClass
 	AddLabels       map[string]string
 
-	// Derived from ClusterQueues and ResourceFlavors at Load time; read-only after that.
-	FlavorValidation  map[string]error
-	FlavorsByResource map[string]map[corev1.ResourceName]kueue.ResourceFlavorReference
+	// Derived from ClusterQueues and ResourceFlavors at Load time.
+	flavorValidation  map[kueue.ClusterQueueReference]error
+	flavorsByResource map[kueue.ClusterQueueReference]map[corev1.ResourceName]kueue.ResourceFlavorReference
 }
 
 func Load(ctx context.Context, c client.Client, namespaces []string, mappingRules mapping.Rules, addLabels map[string]string) (*ImportCache, error) {
@@ -91,13 +92,14 @@ func Load(ctx context.Context, c client.Client, namespaces []string, mappingRule
 	}
 	ret.PriorityClasses = utilslices.ToRefMap(pcList.Items, func(pc *schedulingv1.PriorityClass) string { return pc.Name })
 
-	ret.FlavorValidation = make(map[string]error, len(cqList.Items))
-	ret.FlavorsByResource = make(map[string]map[corev1.ResourceName]kueue.ResourceFlavorReference, len(cqList.Items))
+	ret.flavorValidation = make(map[kueue.ClusterQueueReference]error, len(cqList.Items))
+	ret.flavorsByResource = make(map[kueue.ClusterQueueReference]map[corev1.ResourceName]kueue.ResourceFlavorReference, len(cqList.Items))
 	for i := range cqList.Items {
 		cq := &cqList.Items[i]
 		rgs := resourceGroupsFrom(cq)
-		ret.FlavorsByResource[cq.Name] = flavorsByResourceFrom(rgs)
-		ret.FlavorValidation[cq.Name] = validateFlavors(cq.Name, rgs, ret.ResourceFlavors)
+		cqRef := kueue.ClusterQueueReference(cq.Name)
+		ret.flavorsByResource[cqRef] = flavorsByResourceFrom(rgs)
+		ret.flavorValidation[cqRef] = validateFlavors(cq.Name, rgs, ret.ResourceFlavors)
 	}
 
 	return &ret, nil
@@ -140,12 +142,15 @@ func flavorsByResourceFrom(rgs []resourcegroups.ResourceGroup) map[corev1.Resour
 // validateFlavors checks that every ResourceFlavor referenced by rgs is known.
 func validateFlavors(cqName string, rgs []resourcegroups.ResourceGroup, resourceFlavors map[kueue.ResourceFlavorReference]*kueue.ResourceFlavor) error {
 	// Sorted for a deterministic error message; AllFlavors returns a set.
-	flavors := resourcegroups.AllFlavors(rgs).UnsortedList()
-	slices.Sort(flavors)
+	flavors := sets.List(resourcegroups.AllFlavors(rgs))
+	missing := make([]kueue.ResourceFlavorReference, 0)
 	for _, flavor := range flavors {
 		if _, found := resourceFlavors[flavor]; !found {
-			return fmt.Errorf("%q flavor %q: %w", cqName, flavor, ErrCQInvalid)
+			missing = append(missing, flavor)
 		}
+	}
+	if len(missing) > 0 {
+		return fmt.Errorf("%q missing flavors %v: %w", cqName, missing, ErrCQInvalid)
 	}
 	return nil
 }
@@ -170,4 +175,20 @@ func (ic *ImportCache) LocalQueueForPod(p *corev1.Pod) (*kueue.LocalQueue, bool,
 		return nil, false, fmt.Errorf("%s: %w", queueName, ErrLQNotFound)
 	}
 	return lq, false, nil
+}
+
+func (ic *ImportCache) FlavorValidationForClusterQueue(cqName kueue.ClusterQueueReference) error {
+	return ic.flavorValidation[cqName]
+}
+
+func (ic *ImportCache) FlavorsByResourceForClusterQueue(cqName kueue.ClusterQueueReference) map[corev1.ResourceName]kueue.ResourceFlavorReference {
+	flavorsByResource := ic.flavorsByResource[cqName]
+	if flavorsByResource == nil {
+		return nil
+	}
+
+	ret := make(map[corev1.ResourceName]kueue.ResourceFlavorReference, len(flavorsByResource))
+	maps.Copy(ret, flavorsByResource)
+
+	return ret
 }
--- a/cmd/importer/cache/cache.go
+++ b/cmd/importer/cache/cache.go
@@ -39,7 +39,6 @@ var (
 	ErrLQNotFound = errors.New("localqueue not found")
 	ErrCQNotFound = errors.New("clusterqueue not found")
 	ErrCQInvalid  = errors.New("clusterqueue invalid")
-	ErrPodInvalid = errors.New("pod invalid")
 	ErrPCNotFound = errors.New("priorityclass not found")
 )
 
--- a/cmd/importer/cache/cache.go
+++ b/cmd/importer/cache/cache.go
@@ -32,6 +32,7 @@ import (
 	"sigs.k8s.io/kueue/cmd/importer/mapping"
 	"sigs.k8s.io/kueue/pkg/util/resourcegroups"
 	utilslices "sigs.k8s.io/kueue/pkg/util/slices"
+	"sigs.k8s.io/kueue/pkg/workload"
 )
 
 var (
@@ -43,25 +44,27 @@ var (
 )
 
 type ImportCache struct {
-	Namespaces      []string
-	MappingRules    mapping.Rules
-	LocalQueues     map[string]map[string]*kueue.LocalQueue
-	ClusterQueues   map[string]*kueue.ClusterQueue
-	ResourceFlavors map[kueue.ResourceFlavorReference]*kueue.ResourceFlavor
-	PriorityClasses map[string]*schedulingv1.PriorityClass
-	AddLabels       map[string]string
+	Namespaces          []string
+	MappingRules        mapping.Rules
+	LocalQueues         map[string]map[string]*kueue.LocalQueue
+	ClusterQueues       map[string]*kueue.ClusterQueue
+	ResourceFlavors     map[kueue.ResourceFlavorReference]*kueue.ResourceFlavor
+	PriorityClasses     map[string]*schedulingv1.PriorityClass
+	AddLabels           map[string]string
+	workloadInfoOptions []workload.InfoOption
 
 	// Derived from ClusterQueues and ResourceFlavors at Load time.
 	flavorValidation  map[kueue.ClusterQueueReference]error
 	flavorsByResource map[kueue.ClusterQueueReference]map[corev1.ResourceName]kueue.ResourceFlavorReference
 }
 
-func Load(ctx context.Context, c client.Client, namespaces []string, mappingRules mapping.Rules, addLabels map[string]string) (*ImportCache, error) {
+func Load(ctx context.Context, c client.Client, namespaces []string, mappingRules mapping.Rules, addLabels map[string]string, workloadInfoOptions []workload.InfoOption) (*ImportCache, error) {
 	ret := ImportCache{
-		Namespaces:   slices.Clone(namespaces),
-		MappingRules: mappingRules,
-		LocalQueues:  make(map[string]map[string]*kueue.LocalQueue),
-		AddLabels:    addLabels,
+		Namespaces:          slices.Clone(namespaces),
+		MappingRules:        mappingRules,
+		LocalQueues:         make(map[string]map[string]*kueue.LocalQueue),
+		AddLabels:           addLabels,
+		workloadInfoOptions: slices.Clone(workloadInfoOptions),
 	}
 
 	cqList := &kueue.ClusterQueueList{}
@@ -103,6 +106,10 @@ func Load(ctx context.Context, c client.Client, namespaces []string, mappingRule
 	}
 
 	return &ret, nil
+}
+
+func (ic *ImportCache) WorkloadInfoOptions() []workload.InfoOption {
+	return slices.Clone(ic.workloadInfoOptions)
 }
 
 // resourceGroupsFrom converts cq's API resource groups to the ResourceGroup
--- a/cmd/importer/cache/cache_test.go
+++ b/cmd/importer/cache/cache_test.go
@@ -17,11 +17,13 @@ limitations under the License.
 package cache
 
 import (
+	"errors"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
 	utiltestingapi "sigs.k8s.io/kueue/pkg/util/testing/v1beta2"
@@ -65,7 +67,7 @@ func TestFlavorsByResourceFrom(t *testing.T) {
 				corev1.ResourceName("nvidia.com/gpu"): "on-demand",
 			},
 		},
-		"omits resources not covered by any resource group": {
+		"returns only resources covered by resource groups": {
 			clusterQueue: utiltestingapi.MakeClusterQueue("cq").
 				ResourceGroup(
 					*utiltestingapi.MakeFlavorQuotas("cpu-flavor").
@@ -86,4 +88,45 @@ func TestFlavorsByResourceFrom(t *testing.T) {
 			}
 		})
 	}
+}
+
+func TestValidateFlavors(t *testing.T) {
+	cq := utiltestingapi.MakeClusterQueue("cq").
+		ResourceGroup(
+			*utiltestingapi.MakeFlavorQuotas("spot").
+				Resource(corev1.ResourceCPU, "10", "0").
+				Obj(),
+			*utiltestingapi.MakeFlavorQuotas("on-demand").
+				Resource(corev1.ResourceCPU, "10", "0").
+				Obj(),
+			*utiltestingapi.MakeFlavorQuotas("reserved").
+				Resource(corev1.ResourceCPU, "10", "0").
+				Obj(),
+		).Obj()
+
+	rgs := resourceGroupsFrom(cq)
+
+	t.Run("returns nil when all referenced flavors exist", func(t *testing.T) {
+		err := validateFlavors(cq.Name, rgs, map[kueue.ResourceFlavorReference]*kueue.ResourceFlavor{
+			"on-demand": {ObjectMeta: metav1.ObjectMeta{Name: "on-demand"}},
+			"reserved":  {ObjectMeta: metav1.ObjectMeta{Name: "reserved"}},
+			"spot":      {ObjectMeta: metav1.ObjectMeta{Name: "spot"}},
+		})
+		if err != nil {
+			t.Fatalf("validateFlavors() = %v, want nil", err)
+		}
+	})
+
+	t.Run("returns all missing flavors in deterministic order", func(t *testing.T) {
+		err := validateFlavors(cq.Name, rgs, map[kueue.ResourceFlavorReference]*kueue.ResourceFlavor{
+			"spot": {ObjectMeta: metav1.ObjectMeta{Name: "spot"}},
+		})
+		if !errors.Is(err, ErrCQInvalid) {
+			t.Fatalf("validateFlavors() error = %v, want wrapped %v", err, ErrCQInvalid)
+		}
+		want := "\"cq\" missing flavors [on-demand reserved]: clusterqueue invalid"
+		if got := err.Error(); got != want {
+			t.Fatalf("validateFlavors() error = %q, want %q", got, want)
+		}
+	})
 }
--- a/cmd/importer/cache/cache_test.go
+++ b/cmd/importer/cache/cache_test.go
@@ -1,0 +1,89 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cache
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	corev1 "k8s.io/api/core/v1"
+
+	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+	utiltestingapi "sigs.k8s.io/kueue/pkg/util/testing/v1beta2"
+)
+
+func TestFlavorsByResourceFrom(t *testing.T) {
+	cases := map[string]struct {
+		clusterQueue *kueue.ClusterQueue
+		want         map[corev1.ResourceName]kueue.ResourceFlavorReference
+	}{
+		"returns the flavor from each matching resource group": {
+			clusterQueue: utiltestingapi.MakeClusterQueue("cq").
+				ResourceGroup(
+					*utiltestingapi.MakeFlavorQuotas("cpu-flavor").
+						Resource(corev1.ResourceCPU, "10", "0").
+						Resource(corev1.ResourceMemory, "10Gi", "0").
+						Obj(),
+				).
+				ResourceGroup(
+					*utiltestingapi.MakeFlavorQuotas("gpu-flavor").
+						Resource(corev1.ResourceName("nvidia.com/gpu"), "10", "0").
+						Obj(),
+				).Obj(),
+			want: map[corev1.ResourceName]kueue.ResourceFlavorReference{
+				corev1.ResourceCPU:                    "cpu-flavor",
+				corev1.ResourceMemory:                 "cpu-flavor",
+				corev1.ResourceName("nvidia.com/gpu"): "gpu-flavor",
+			},
+		},
+		"returns the first flavor from a resource group with several flavors": {
+			clusterQueue: utiltestingapi.MakeClusterQueue("cq").
+				ResourceGroup(
+					*utiltestingapi.MakeFlavorQuotas("on-demand").
+						Resource(corev1.ResourceName("nvidia.com/gpu"), "10", "0").
+						Obj(),
+					*utiltestingapi.MakeFlavorQuotas("spot").
+						Resource(corev1.ResourceName("nvidia.com/gpu"), "10", "0").
+						Obj(),
+				).Obj(),
+			want: map[corev1.ResourceName]kueue.ResourceFlavorReference{
+				corev1.ResourceName("nvidia.com/gpu"): "on-demand",
+			},
+		},
+		"omits resources not covered by any resource group": {
+			clusterQueue: utiltestingapi.MakeClusterQueue("cq").
+				ResourceGroup(
+					*utiltestingapi.MakeFlavorQuotas("cpu-flavor").
+						Resource(corev1.ResourceCPU, "10", "0").
+						Obj(),
+				).Obj(),
+			want: map[corev1.ResourceName]kueue.ResourceFlavorReference{
+				corev1.ResourceCPU: "cpu-flavor",
+			},
+		},
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			got := flavorsByResourceFrom(resourceGroupsFrom(tc.clusterQueue))
+			if diff := cmp.Diff(tc.want, got, cmpopts.EquateEmpty()); diff != "" {
+				t.Errorf("Unexpected flavors (-want/+got)\n%s", diff)
+			}
+		})
+	}
+}
--- a/cmd/importer/pod/check.go
+++ b/cmd/importer/pod/check.go
@@ -134,7 +134,7 @@ func checkPodWorkload(ctx context.Context, c client.Client, importCache *cache.I
 	}
 	maps.Copy(wl.Labels, importCache.AddLabels)
 
-	info := workload.NewInfo(wl)
+	info := workload.NewInfo(wl, importCache.WorkloadInfoOptions()...)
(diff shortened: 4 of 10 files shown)


cmd/importer/pod/import.go
```diff
@@ -186,8 +204,19 @@ func admitWorkload(ctx context.Context, c client.Client, wl *kueue.Workload, cq
 				},
 			},
 		}
-		flv := cq.Spec.ResourceGroups[0].Flavors[0].Name
+
+		// sort requestedResources to ensure deterministic behavior in the admission process and toavoid flakiness in tests.
```
ArangoGutierrez: nit: typo, `toavoid` should be `to avoid`. Separately: the Flavors map you build is order independent, so the sort does not change the assignment itself. Where it does matter is which ResourceNotCoveredError surfaces first when more than one resource is uncovered. Might be worth saying that in the comment so the next reader knows why the sort is here.

cmd/importer/pod/import.go
```diff
@@ -229,3 +258,19 @@ func admitWorkload(ctx context.Context, c client.Client, wl *kueue.Workload, cq
...
+// that covers the requested resource. It skips groups with no flavors and returns
+// an empty string when no matching group exists.
+func resourceFlavorForResource(cq *kueue.ClusterQueue, resource corev1.ResourceName) kueue.ResourceFlavorReference {
+	for _, rg := range cq.Spec.ResourceGroups {
+		if len(rg.Flavors) == 0 {
+			continue
+		}
+
+		if slices.Contains(rg.CoveredResources, resource) {
+			return rg.Flavors[0].Name
```
ArangoGutierrez: You return the first flavor of the covering group. If a group lists several flavors (on-demand, spot), a pod that is really on spot would be imported as on-demand. Is first-flavor intended here, or should this match against the pod's node labels? Not asking you to change it in this PR, just want the decision written down somewhere.

cmd/importer/pod/import.go
```diff
@@ -44,6 +45,23 @@ import (
 
 var realClock = clock.RealClock{}
 
+type ResourceNotCoveredError struct {
```
ArangoGutierrez: nit: this type is only referenced inside the package today. It does travel out through Import's error return, so if you expect callers to errors.As on it, exporting is correct. If not, lowercasing keeps the importer's API surface smaller. Your call, just flagging.

ArangoGutierrez (commented):
Read through this. The core fix is right: each requested resource is mapped to the flavor of the resource group that actually covers it, instead of assigning ResourceGroups[0].Flavors[0] to every resource. The two new table cases pin the exact per-resource flavors (cpu-flavor for cpu, gpu-flavor for gpu) and the uncovered-resource error, so they fail against the old code. That is the right discriminator.

Two questions before I approve, neither blocking:

- Within a covered group you always take Flavors[0]. For a group with several flavors, say on-demand and spot, a pod actually running on spot gets imported as on-demand. Is first-flavor a deliberate simplification, or should the importer match the flavor to the pod's node labels? This is pre-existing behavior, I just want the choice recorded.
- Re-import: if a pod was already imported and its resource is no longer covered by the target CQ, admitWorkload now returns ResourceNotCoveredError and the earlier admission is left in place. That is fine for a one-shot run, but worth confirming it is the intended behavior.

Scope and tests look good otherwise. Nits inline.

cmd/importer/pod/import.go
mszadkow: Like you said this was pre-exisitng, it's seems to be a shortcut.
However as you brought it up, I think I can do a follow up for matching the flavor, I think with the match it doesn't require a documentation of any sort.

cmd/importer/pod/import.go
mszadkow: thanks, that will go lower case

mszadkow:
@ArangoGutierrez ptal

mszadkow:
The plan is the follow up with exact flavor match.
@vladikkuzn @ivnovakov any further thoughts?

vladikkuzn:
>The plan is the follow up with exact flavor match.
@vladikkuzn @ivnovakov any further thoughts?

Nothing from my side

cmd/importer/pod/import.go
```diff
@@ -196,11 +214,21 @@ func admitWorkload(ctx context.Context, c client.Client, wl *kueue.Workload, cq
...
-			})
+
+		var forEachErr error
+		info.TotalRequests[0].Requests.ForEach(func(name corev1.ResourceName, _ int64) {
+			if forEachErr != nil {
+				return
+			}
+			flv := resourceFlavorForResource(cq, name)
+			if flv == "" {
+				forEachErr = &ResourceNotCoveredError{Resource: name, ClusterQueue: cq.Name}
```
ArangoGutierrez: This is the one I would fix before merge. `admitWorkload` can now fail, but `Check` in check.go still validates only `ResourceGroups[0]` and its first flavor, so `--dry-run` (the default) reports success for exactly the pods the real import will reject. Before this PR both files used the same `ResourceGroups[0].Flavors[0].Name` expression, so a green check genuinely did guarantee the import would complete.

The failure now lands after `addLabels` and `createWorkload` have already run, which leaves a running pod carrying the queue and managed-by-kueue labels with a Workload that has no Admission. The jobframework reconciler reads that as a running job that is not admitted and calls `stopJob(StopReasonNotAdmitted)`, and `Pod.Stop` sets TerminationTarget and deletes the pod. So a green dry run can be followed by an import that terminates the pods it failed on. Mirroring the per-resource coverage check into `Check` would keep the failure in the read-only phase.

cmd/importer/pod/import.go
```diff
@@ -241,3 +269,19 @@ func admitWorkload(ctx context.Context, c client.Client, wl *kueue.Workload, cq
...
+// that covers the requested resource. It skips groups with no flavors and returns
+// an empty string when no matching group exists.
+func resourceFlavorForResource(cq *kueue.ClusterQueue, resource corev1.ResourceName) kueue.ResourceFlavorReference {
+	for _, rg := range cq.Spec.ResourceGroups {
+		if len(rg.Flavors) == 0 {
+			continue
+		}
+
+		if slices.Contains(rg.CoveredResources, resource) {
+			return rg.Flavors[0].Name
```
ArangoGutierrez: Same root cause, second consequence: this can now return a flavor from any resource group, but `Check` only resolves `ResourceGroups[0].Flavors[0].Name` against `importCache.ResourceFlavors`. A flavor from a later group is never verified to exist, so an Admission can reference a ResourceFlavor object that was never created and the `ErrCQInvalid` guard is silently bypassed. Worth extending the check to every flavor the import can select.

cmd/importer/pod/import.go
```diff
@@ -196,11 +214,21 @@ func admitWorkload(ctx context.Context, c client.Client, wl *kueue.Workload, cq
...
 			},
 		}
-		flv := cq.Spec.ResourceGroups[0].Flavors[0].Name
-		if info.TotalRequests[0].Requests != nil {
-			info.TotalRequests[0].Requests.ForEach(func(name corev1.ResourceName, val int64) {
-				admission.PodSetAssignments[0].Flavors[name] = flv
-			})
+
+		var forEachErr error
+		info.TotalRequests[0].Requests.ForEach(func(name corev1.ResourceName, _ int64) {
```
ArangoGutierrez: Swapping the sorted loop for `ForEach` drops the determinism the old comment called out. Order here depends on which `Requests` implementation the `VectorizedResourceRequests` gate picks: I ran both, and `*SliceRequests` gives a single stable order over 200 iterations while `MapRequests` gives five. The gate defaults to true and the importer has no `--feature-gates` flag, so nothing is broken today. But which `ResourceNotCoveredError` surfaces when two resources are uncovered now depends on that default rather than on anything in this file. Sorting, or a one-line comment noting the dependency, would pin it back down. Non-blocking.

cmd/importer/pod/import.go
```diff
@@ -45,6 +46,23 @@ import (
 
 var realClock = clock.RealClock{}
 
+type ResourceNotCoveredError struct {
```
ArangoGutierrez: Follow-up from the last round: you mentioned this would go lowercase and it is still exported. Keeping it exported is fine if you expect callers to `errors.As` on it, just noting the change did not land.

cmd/importer/pod/import_test.go
```diff
@@ -176,22 +170,8 @@ func TestImportNamespace(t *testing.T) {
...
-					ToLocalQueue: "lq1",
-				},
-			},
-			localQueues: []kueue.LocalQueue{
-				*baseLocalQueue.Obj(),
-			},
-			clusterQueues: nil, // intentionally omit cluster queues to trigger ErrCQNotFound
-
+			localQueue:   *baseLocalQueue.Obj(),
+			clusterQueue: kueue.ClusterQueue{}, // intentionally omit cluster queues to trigger ErrCQNotFound
```
ArangoGutierrez: The comment says cluster queues are omitted, but the harness always builds `Items` from `tc.clusterQueue`, so this creates a ClusterQueue with an empty name rather than none at all. The case still reaches `ErrCQNotFound` because the LocalQueue points at cq1, so the coverage is real, but the comment describes a setup that is not what runs.

ArangoGutierrez (commented):
Second pass, now on 877d050c.

The first-flavor question from the last round is resolved well: the behavior is documented on `resourceFlavorForResource` and pinned by a unit test, so the decision lives in the code rather than in a review thread. The new `ErrCQNotFound` guard and the nil-labels guard are both good additions. `ResourceNotCoveredError` is still exported.

One new issue is worth looking at before merge. Turning the uncovered-resource case into a hard failure is the right call, but the failure is not mirrored into the dry-run check, so `--dry-run` passes for exactly the pods the real import will reject, and it rejects them after it has already labeled the running pod. Details inline.

Following the repo AI policy: I used an AI assistant while preparing this review and verified the findings myself against the tree.

cmd/importer/pod/import.go
mszadkow: Thanks for spotting this!
This will require for `Check` to be a little bit more like `Import` but it's a good change.

cmd/importer/pod/import.go
mszadkow: I will bring back the sorting

cmd/importer/pod/import.go
mszadkow: Thanks for a reminder

cmd/importer/pod/import.go
mszadkow: Updated, now `Check` verifies the resources the sam way as `Import`.

cmd/importer/pod/check.go
```diff
@@ -49,15 +52,30 @@ func Check(ctx context.Context, c client.Client, importCache *cache.ImportCache,
...
+		info := workload.NewInfo(wl)
+		if len(info.TotalRequests) == 0 {
+			return false, fmt.Errorf("workload has no total requests: %w", cache.ErrCQInvalid)
+		}
+
+		flavors, err := flavorAssignmentsForRequests(cq, info.TotalRequests[0].Requests)
+		if err != nil {
+			return false, err
+		}
+		if err := validateKnownFlavorAssignments(cq.Name, flavors, importCache.ResourceFlavors); err != nil {
```
ArangoGutierrez: Flavor validation is now driven by the pod's own requests, so a pod with no container requests yields an empty map and this loop never iterates. I ran it: a request-less pod against a ClusterQueue whose only flavor is `does-not-exist`, with no ResourceFlavor objects present, now logs `Successfully checked`. The line this replaced validated `ResourceGroups[0].Flavors[0]` unconditionally and caught that. Keeping a ClusterQueue-level flavor-existence check alongside the per-resource one would stop BestEffort pods from getting a clean dry run against a queue Kueue will mark Inactive.

cmd/importer/pod/import.go
```diff
@@ -185,23 +203,22 @@ func admitWorkload(ctx context.Context, c client.Client, wl *kueue.Workload, cq
 		// make its admission and update its status
 		info := workload.NewInfo(wl)
 
+		flavors, err := flavorAssignmentsForRequests(cq, info.TotalRequests[0].Requests)
```
ArangoGutierrez: This failure lands after `addLabels` has stamped the pod with queue-name and managed-by, and after `createWorkload` has persisted the Workload, so an uncovered resource leaves a Kueue-managed pod plus a never-admitted Workload behind. `Check` running first narrows the window rather than closing it, since `Check` and `Import` each do their own `ListPods` and a pod created between the two passes is imported unchecked. The same point was made on this function in #13176 for the ClusterQueue lookup, which is why that lookup now sits ahead of `createWorkload`. This check only needs `cq` and the pod's requests, so it could move up next to it.

cmd/importer/pod/check.go
```diff
@@ -49,15 +52,30 @@ func Check(ctx context.Context, c client.Client, importCache *cache.ImportCache,
 		if len(cq.Spec.ResourceGroups) == 0 {
 			return false, fmt.Errorf("%q has no resource groups: %w", cq.Name, cache.ErrCQInvalid)
 		}
+		for _, rg := range cq.Spec.ResourceGroups {
```
ArangoGutierrez: This rejects a ClusterQueue when any resource group has zero flavors, but `resourceFlavorForResource` deliberately skips those groups, so `Check` ends up stricter than the import it mirrors. I confirmed the split: with group 0 empty and group 1 covering cpu, `Import` assigns cpu to rf1 while `Check` returns `ErrCQInvalid`. `ResourceGroup.Flavors` is `+required` with `MinItems=1`, so the API server will not accept such a ClusterQueue and neither branch is reachable in practice. Raising it only because the comment three lines down says the two paths are aligned and the new test case pins the divergence. Non-blocking.

ArangoGutierrez (commented):
Third pass, now on 25fde6b4. All five points from the last round are resolved: `Check` now runs the same per-resource flavor resolution as `Import`, the flavor-existence guard covers every group, `flavorAssignmentsForRequests` sorts before iterating, `resourceNotCoveredError` is unexported, and the test comment matches the fixture. I mutation-checked the new coverage: deleting the `validateKnownFlavorAssignments` call fails `unknown ResourceFlavor assignment is invalid`, and reverting `resourceFlavorForResource` to the old first-group behavior fails the cpu+gpu import case and two `TestResourceFlavorForResource` cases. The tests discriminate. Three things left, one of which I would fix before merge.

1. Flavor validation is now driven by the pod's own requests, so a pod with no container requests yields an empty map and this loop never iterates. I ran it: a request-less pod against a ClusterQueue whose only flavor is `does-not-exist`, with no ResourceFlavor objects present, now logs `Successfully checked`. The line this replaced validated `ResourceGroups[0].Flavors[0]` unconditionally and caught that. Keeping a ClusterQueue-level flavor-existence check alongside the per-resource one would stop BestEffort pods from getting a clean dry run against a queue Kueue will mark Inactive. (cmd/importer/pod/check.go:77)
2. This failure lands after `addLabels` has stamped the pod with queue-name and managed-by, and after `createWorkload` has persisted the Workload, so an uncovered resource leaves a Kueue-managed pod plus a never-admitted Workload behind. `Check` running first narrows the window rather than closing it, since `Check` and `Import` each do their own `ListPods` and a pod created between the two passes is imported unchecked. The same point was made on this function in #13176 for the ClusterQueue lookup, which is why that lookup now sits ahead of `createWorkload`. This check only needs `cq` and the pod's requests, so it could move up next to it. (cmd/importer/pod/import.go:206)
3. This rejects a ClusterQueue when any resource group has zero flavors, but `resourceFlavorForResource` deliberately skips those groups, so `Check` ends up stricter than the import it mirrors. I confirmed the split: with group 0 empty and group 1 covering cpu, `Import` assigns cpu to rf1 while `Check` returns `ErrCQInvalid`. `ResourceGroup.Flavors` is `+required` with `MinItems=1`, so the API server will not accept such a ClusterQueue and neither branch is reachable in practice. Raising it only because the comment three lines down says the two paths are aligned and the new test case pins the divergence. Non-blocking. (cmd/importer/pod/check.go:55)

mszadkow:
> Third pass, now on [25fde6b](https://github.com/kubernetes-sigs/kueue/commit/25fde6b46893a9ed226af3cd826a470a8d57f33a). All five points from the last round are resolved: `Check` now runs the same per-resource flavor resolution as `Import`, the flavor-existence guard covers every group, `flavorAssignmentsForRequests` sorts before iterating, `resourceNotCoveredError` is unexported, and the test comment matches the fixture. I mutation-checked the new coverage: deleting the `validateKnownFlavorAssignments` call fails `unknown ResourceFlavor assignment is invalid`, and reverting `resourceFlavorForResource` to the old first-group behavior fails the cpu+gpu import case and two `TestResourceFlavorForResource` cases. The tests discriminate. Three things left, one of which I would fix before merge.
> 
> 1. Flavor validation is now driven by the pod's own requests, so a pod with no container requests yields an empty map and this loop never iterates. I ran it: a request-less pod against a ClusterQueue whose only flavor is `does-not-exist`, with no ResourceFlavor objects present, now logs `Successfully checked`. The line this replaced validated `ResourceGroups[0].Flavors[0]` unconditionally and caught that. Keeping a ClusterQueue-level flavor-existence check alongside the per-resource one would stop BestEffort pods from getting a clean dry run against a queue Kueue will mark Inactive. (cmd/importer/pod/check.go:77)
> 2. This failure lands after `addLabels` has stamped the pod with queue-name and managed-by, and after `createWorkload` has persisted the Workload, so an uncovered resource leaves a Kueue-managed pod plus a never-admitted Workload behind. `Check` running first narrows the window rather than closing it, since `Check` and `Import` each do their own `ListPods` and a pod created between the two passes is imported unchecked. The same point was made on this function in #13176 for the ClusterQueue lookup, which is why that lookup now sits ahead of `createWorkload`. This check only needs `cq` and the pod's requests, so it could move up next to it. (cmd/importer/pod/import.go:206)
> 3. This rejects a ClusterQueue when any resource group has zero flavors, but `resourceFlavorForResource` deliberately skips those groups, so `Check` ends up stricter than the import it mirrors. I confirmed the split: with group 0 empty and group 1 covering cpu, `Import` assigns cpu to rf1 while `Check` returns `ErrCQInvalid`. `ResourceGroup.Flavors` is `+required` with `MinItems=1`, so the API server will not accept such a ClusterQueue and neither branch is reachable in practice. Raising it only because the comment three lines down says the two paths are aligned and the new test case pins the divergence. Non-blocking. (cmd/importer/pod/check.go:55)

1. Check now validates ClusterQueue flavor references independently of pod request-derived assignments, so request-less pods no longer slip through with missing flavors.

2. We moved import-side validation earlier so uncovered-resource and missing-flavor failures happen before any mutation.

3. It was solved by removing the special-case rejection in Check that treated any empty-flavor resource group as invalid, and relying on the same flavor-resolution path Import uses.

mszadkow:
@coderabbitai review

mszadkow:
@coderabbitai resume

cmd/importer/pod/import.go
```diff
@@ -241,3 +271,39 @@ func admitWorkload(ctx context.Context, c client.Client, wl *kueue.Workload, cq
...
+
+	resourceNames := make([]corev1.ResourceName, 0, requests.Len())
+	requests.ForEach(func(name corev1.ResourceName, _ int64) {
+		resourceNames = append(resourceNames, name)
+	})
+	slices.Sort(resourceNames)
+
+	for _, name := range resourceNames {
+		flv := resourceFlavorForResource(cq, name)
+		if flv == "" {
```
ArangoGutierrez: The `ForEach` above drops the quantity, so a resource requested at `0` is treated like a real request and hard-fails when the ClusterQueue does not cover it. The scheduler deliberately skips exactly that case (`pkg/scheduler/flavorassigner/flavorassigner.go:752-758`, added for #8079), so the importer now rejects pods Kueue itself would admit. Since `main.go:210` returns on the first `Check` error, one such pod aborts the whole import run, not just that pod. Mirroring the `quantity == 0` skip here would keep the two paths aligned.

ArangoGutierrez (commented):
Round 4 looks good: Check now mirrors Import, and moving validation ahead of addLabels/createWorkload cleanly fixes the orphaned-pod case. Two things left.

1. The `ForEach` above drops the quantity, so a resource requested at `0` is treated like a real request and hard-fails when the ClusterQueue does not cover it. The scheduler deliberately skips exactly that case (`pkg/scheduler/flavorassigner/flavorassigner.go:752-758`, added for #8079), so the importer now rejects pods Kueue itself would admit. Since `main.go:210` returns on the first `Check` error, one such pod aborts the whole import run, not just that pod. Mirroring the `quantity == 0` skip here would keep the two paths aligned. (cmd/importer/pod/import.go:302)
2. `constructPodSets` always returns exactly one PodSet for a single pod, so this cannot trigger today. If it ever did, the cause would be the pod rather than the ClusterQueue, and `ErrCQInvalid` would point the operator at the wrong object. Same guard in `check.go:67`. Either drop it or give it a sentinel that names the pod. (cmd/importer/pod/import.go:101)

cmd/importer/pod/import.go
```diff
@@ -77,6 +94,25 @@ func Import(ctx context.Context, c client.Client, importCache *cache.ImportCache
 		if err != nil {
 			return false, fmt.Errorf("construct workload: %w", err)
 		}
+		// The pod might not carry the queue label yet; set the resolved queue explicitly.
+		wl.Spec.QueueName = kueue.LocalQueueName(lq.Name)
+
+		info := workload.NewInfo(wl)
+		if len(info.TotalRequests) == 0 {
```
ArangoGutierrez: `constructPodSets` always returns exactly one PodSet for a single pod, so this cannot trigger today. If it ever did, the cause would be the pod rather than the ClusterQueue, and `ErrCQInvalid` would point the operator at the wrong object. Same guard in `check.go:67`. Either drop it or give it a sentinel that names the pod.

cmd/importer/pod/import.go
```diff
@@ -77,6 +94,25 @@ func Import(ctx context.Context, c client.Client, importCache *cache.ImportCache
 		if err != nil {
 			return false, fmt.Errorf("construct workload: %w", err)
 		}
+		// The pod might not carry the queue label yet; set the resolved queue explicitly.
+		wl.Spec.QueueName = kueue.LocalQueueName(lq.Name)
```
ArangoGutierrez: Restoring `QueueName` here covers one of several spec fields that `jobframework.NewWorkload` derives from the pod's labels. It also reads `MaximumExecutionTimeSecondsForObject` (`kueue.x-k8s.io/max-exec-time-seconds`), and `PrebuiltWorkloadNameFor` picks the Workload *name* from `kueue.x-k8s.io/prebuilt-workload-name` - but the pod is not labelled yet here, because `addLabels` now runs further down. I ran both trees with `--add-labels kueue.x-k8s.io/max-exec-time-seconds=3600`: before this PR the Workload got `MaximumExecutionTimeSeconds=3600`, after it gets `nil`, while the pod and the Workload metadata both still carry the label. Setting the labels on the in-memory pod before `pod.FromObject(p)` and persisting them only once validation passes would keep the new ordering and the old construction inputs.

cmd/importer/pod/import_test.go
```diff
@@ -235,9 +290,80 @@ func TestImportNamespace(t *testing.T) {
...
+			if tc.wantAdmissionOnlyCompare {
+				workloads := make([]kueue.Workload, 0, len(wlList.Items))
+				for _, wl := range wlList.Items {
+					workloads = append(workloads, kueue.Workload{Status: kueue.WorkloadStatus{Admission: wl.Status.Admission}})
+				}
+				if diff := cmp.Diff(
+					tc.wantWorkloads,
+					workloads,
+					cmpopts.EquateEmpty(),
+					cmpopts.IgnoreFields(kueue.PodSetAssignment{}, "ResourceUsage", "Count"),
```
ArangoGutierrez: This is the only multi-resource case in the suite, so ignoring `ResourceUsage` leaves the per-resource accounting unverified. I deleted every non-CPU key from the `ResourceUsage` that `admitWorkload` builds and the package stayed green. Projecting down to `Status.Admission` also means this case never notices the `wl.Spec.QueueName` line the PR adds - deleting that line fails `create one` but not this one. The expected values are plain literals, so a full expected Workload built from `baseWlWrapper` would remove the boolean and both blind spots at once.

cmd/importer/pod/import.go
```diff
@@ -241,3 +271,39 @@ func admitWorkload(ctx context.Context, c client.Client, wl *kueue.Workload, cq
...
+}
+
+func flavorAssignmentsForRequests(cq *kueue.ClusterQueue, requests resources.Requests) (map[corev1.ResourceName]kueue.ResourceFlavorReference, error) {
+	flavors := make(map[corev1.ResourceName]kueue.ResourceFlavorReference)
+
+	resourceNames := make([]corev1.ResourceName, 0, requests.Len())
+	requests.ForEach(func(name corev1.ResourceName, _ int64) {
+		resourceNames = append(resourceNames, name)
+	})
+	slices.Sort(resourceNames)
```
ArangoGutierrez: Closing the loop on the determinism point from the last round: the sort is back, but nothing guards it. Every case has at most one uncovered resource, so deleting this line leaves the whole package green. A case with two uncovered resources, asserting the lexicographically first one surfaces, would lock the behaviour in.

cmd/importer/pod/import.go
```diff
@@ -77,6 +94,25 @@ func Import(ctx context.Context, c client.Client, importCache *cache.ImportCache
...
 			return false, fmt.Errorf("construct workload: %w", err)
 		}
+		// The pod might not carry the queue label yet; set the resolved queue explicitly.
+		wl.Spec.QueueName = kueue.LocalQueueName(lq.Name)
+
+		info := workload.NewInfo(wl)
+		if len(info.TotalRequests) == 0 {
+			return false, fmt.Errorf("workload has no total requests: %w", cache.ErrCQInvalid)
+		}
+		if _, err := flavorAssignmentsForRequests(cq, info.TotalRequests[0].Requests); err != nil {
```
ArangoGutierrez: This computes the assignment map, discards it, and `admitWorkload` then recomputes the identical map to actually use it. They agree today, so this is latent rather than live. But if they ever diverge, the failure lands after the pod is labelled and the Workload created - the partial state this reordering was meant to avoid. Computing it once here and passing the map into `admitWorkload` removes the possibility.

cmd/importer/pod/check_test.go
```diff
@@ -119,21 +101,61 @@ func TestCheckNamespace(t *testing.T) {
 			},
 			wantError: cache.ErrCQInvalid,
 		},
-		"all found": {
+		"unknown ResourceFlavor assignment is invalid": {
```
ArangoGutierrez: This returns at `validateKnownClusterQueueFlavors` before any workload is built, so it never reaches flavor assignment and ends up asserting the same guard as the request-less case below it. Removing that call from `Check` fails exactly these two subtests together and nothing else. Either rename it to say it covers flavor validation, or give it a known flavor so it reaches `flavorAssignmentsForRequests`.

ArangoGutierrez (commented):
Deeper pass this round with Go and test-focused review, including mutation checks against a local build of the branch. One regression worth fixing before merge, plus a few gaps in the new tests.

Smaller and non-blocking, so not inlined: the `len(rg.Flavors) == 0` skip in `resourceFlavorForResource` looks unreachable, since `flavors` is required with `MinItems=1`, so it and the matching sentence in the godoc could both go. And the new `return false, err` at lines 88 and 105, plus `check.go:56` and `:72`, do not wrap while their neighbours do.

1. Restoring `QueueName` here covers one of several spec fields that `jobframework.NewWorkload` derives from the pod's labels. It also reads `MaximumExecutionTimeSecondsForObject` (`kueue.x-k8s.io/max-exec-time-seconds`), and `PrebuiltWorkloadNameFor` picks the Workload *name* from `kueue.x-k8s.io/prebuilt-workload-name` - but the pod is not labelled yet here, because `addLabels` now runs further down. I ran both trees with `--add-labels kueue.x-k8s.io/max-exec-time-seconds=3600`: before this PR the Workload got `MaximumExecutionTimeSeconds=3600`, after it gets `nil`, while the pod and the Workload metadata both still carry the label. Setting the labels on the in-memory pod before `pod.FromObject(p)` and persisting them only once validation passes would keep the new ordering and the old construction inputs. (cmd/importer/pod/import.go:98)
2. This is the only multi-resource case in the suite, so ignoring `ResourceUsage` leaves the per-resource accounting unverified. I deleted every non-CPU key from the `ResourceUsage` that `admitWorkload` builds and the package stayed green. Projecting down to `Status.Admission` also means this case never notices the `wl.Spec.QueueName` line the PR adds - deleting that line fails `create one` but not this one. The expected values are plain literals, so a full expected Workload built from `baseWlWrapper` would remove the boolean and both blind spots at once. (cmd/importer/pod/import_test.go:302)
3. Closing the loop on the determinism point from the last round: the sort is back, but nothing guards it. Every case has at most one uncovered resource, so deleting this line leaves the whole package green. A case with two uncovered resources, asserting the lexicographically first one surfaces, would lock the behaviour in. (cmd/importer/pod/import.go:298)
4. This computes the assignment map, discards it, and `admitWorkload` then recomputes the identical map to actually use it. They agree today, so this is latent rather than live. But if they ever diverge, the failure lands after the pod is labelled and the Workload created - the partial state this reordering was meant to avoid. Computing it once here and passing the map into `admitWorkload` removes the possibility. (cmd/importer/pod/import.go:104)
5. This returns at `validateKnownClusterQueueFlavors` before any workload is built, so it never reaches flavor assignment and ends up asserting the same guard as the request-less case below it. Removing that call from `Check` fails exactly these two subtests together and nothing else. Either rename it to say it covers flavor validation, or give it a known flavor so it reaches `flavorAssignmentsForRequests`. (cmd/importer/pod/check_test.go:104)

cmd/importer/pod/import.go
mszadkow: Adding UT for this gap

mszadkow:
> Round 4 looks good: Check now mirrors Import, and moving validation ahead of addLabels/createWorkload cleanly fixes the orphaned-pod case. Two things left.
> 
> 1. The `ForEach` above drops the quantity, so a resource requested at `0` is treated like a real request and hard-fails when the ClusterQueue does not cover it. The scheduler deliberately skips exactly that case (`pkg/scheduler/flavorassigner/flavorassigner.go:752-758`, added for #8079), so the importer now rejects pods Kueue itself would admit. Since `main.go:210` returns on the first `Check` error, one such pod aborts the whole import run, not just that pod. Mirroring the `quantity == 0` skip here would keep the two paths aligned. (cmd/importer/pod/import.go:302)
> 2. `constructPodSets` always returns exactly one PodSet for a single pod, so this cannot trigger today. If it ever did, the cause would be the pod rather than the ClusterQueue, and `ErrCQInvalid` would point the operator at the wrong object. Same guard in `check.go:67`. Either drop it or give it a sentinel that names the pod. (cmd/importer/pod/import.go:101)

Thanks!
I have updated the code and added couple important test cases!
Also I have updated README.md so it better describes what is happening.

cmd/importer/pod/import.go
```diff
@@ -62,39 +81,65 @@ func Import(ctx context.Context, c client.Client, importCache *cache.ImportCache
...
+				return false, fmt.Errorf("cannot add queue label: %w", err)
+			}
+		} else if oldLq != lq.Name {
+			return false, fmt.Errorf("another local queue name is set %q expecting %q", oldLq, lq.Name)
+		}
+
 		if pc, found := importCache.PriorityClasses[p.Spec.PriorityClassName]; found {
 			wl.Spec.PriorityClassRef = kueue.NewPodPriorityClassRef(pc.Name)
 			wl.Spec.Priority = &pc.Value
 		}
```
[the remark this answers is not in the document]
mszadkow: Sounds like a valid point, though import never did check that.
Can the error be a part of follow up? 
I have a feeling we aggregated a lot of changes for single PR.

mszadkow:
@ArangoGutierrez ptal

mszadkow:
@reruno @mykysha hey guys ptal

cmd/importer/README.md
```diff
@@ -29,7 +29,10 @@ The importer will perform following checks:
 - For every Pod a  mapping to a LocalQueue is available.
 - The target LocalQueue exists.
 - The LocalQueues involved in the import are using an existing ClusterQueue.
-- The ClusterQueues involved have at least one ResourceGroup using an existing ResourceFlavor. This ResourceFlavor is used when the importer creates the admission for the created workloads.
+- The ClusterQueues involved have ResourceGroups that reference existing ResourceFlavors.
+- For each Pod, the dry-run validates that every non-zero requested resource is covered by the target ClusterQueue ResourceGroups.
+- For each Pod, the dry-run validates that Workload construction succeeds (using the same construction path used by import) before any Pod mutation.
+- If a Pod specifies a PriorityClass, the dry-run validates that the PriorityClass exists.
```
dkaluza: This looks like misplaced or is by mistake referencing the dry run, the documentation fragment is about check not the dry-run import.


If this was on puprose we should clarify the relation between the check and dry-run, as right now it is not clear.

cmd/importer/pod/check.go
```diff
@@ -49,25 +52,32 @@ func Check(ctx context.Context, c client.Client, importCache *cache.ImportCache,
 		if len(cq.Spec.ResourceGroups) == 0 {
 			return false, fmt.Errorf("%q has no resource groups: %w", cq.Name, cache.ErrCQInvalid)
 		}
+		if err := validateKnownClusterQueueFlavors(cq, importCache.ResourceFlavors); err != nil {
+			return false, err
+		}
```
dkaluza: IIUC in this way we will revalidate CQ for each imprted/checked pod.

This can result in 1k revalidations of the same CQ if 1k pods are targeting the same one.

Moreover in this way we are introducing duplication between import and check, which are run one after the other IIUC, so we will be repeating this twice without any purpose.

IMO this should be run once for all CQs not per pod.

cmd/importer/pod/check.go
```diff
@@ -49,25 +52,32 @@ func Check(ctx context.Context, c client.Client, importCache *cache.ImportCache,
...
 		var pv int32
 		if pc, found := importCache.PriorityClasses[p.Spec.PriorityClassName]; found {
 			pv = pc.Value
 		} else if p.Spec.PriorityClassName != "" {
 			return false, fmt.Errorf("%q: %w", p.Spec.PriorityClassName, cache.ErrPCNotFound)
 		}
 
-		log.V(2).Info("Successfully checked", "clusterQueue", klog.KObj(cq), "resourceFlavor", klog.KObj(rf), "priority", pv)
+		log.V(2).Info("Successfully checked", "clusterQueue", klog.KObj(cq), "priority", pv)
```
dkaluza: IMO we should mention that we have checked requests of the workload if we do so.

cmd/importer/pod/check.go
```diff
@@ -49,25 +52,32 @@ func Check(ctx context.Context, c client.Client, importCache *cache.ImportCache,
...
-			return false, fmt.Errorf("%q flavor %q: %w", cq.Name, rfName, cache.ErrCQInvalid)
+		info := workload.NewInfo(wl)
+		if len(info.TotalRequests) == 0 {
+			return false, fmt.Errorf("workload has no total requests: %w", cache.ErrPodInvalid)
 		}
 
+		if _, err := flavorAssignmentsForRequests(cq, info.TotalRequests[0].Requests); err != nil {
+			return false, err
+		}
 		var pv int32
```
dkaluza: From my point of view we are getting considerable duplication with the import command here.

IMO we should have a helper that "checks the workload" -  constructs, checks requests, priority classes.
And the "check" command should be basically just run of this helper, when Import should also call this helper and then if it succeeds proceed with additional needed steps.

Otherwise those can diverge easily.

cmd/importer/pod/import.go
```diff
@@ -62,39 +81,65 @@ func Import(ctx context.Context, c client.Client, importCache *cache.ImportCache
...
+		podForWorkload := preparePodForWorkload(p, lq.Name, importCache.AddLabels)
+
+		kp := pod.FromObject(podForWorkload)
 		// Note: the recorder is not used for single pods, we can just pass nil for now.
 		wl, err := kp.ConstructComposableWorkload(ctx, c, nil, nil, nil)
 		if err != nil {
 			return false, fmt.Errorf("construct workload: %w", err)
 		}
-
+		if prebuiltWorkloadName := jobframework.PrebuiltWorkloadNameFor(podForWorkload); prebuiltWorkloadName != "" {
```
dkaluza: what happens if it is empty?

cmd/importer/pod/import.go
```diff
@@ -62,39 +81,65 @@ func Import(ctx context.Context, c client.Client, importCache *cache.ImportCache
...
+		}
+		var flavors map[corev1.ResourceName]kueue.ResourceFlavorReference
+		if flavors, err = flavorAssignmentsForRequests(cq, info.TotalRequests[0].Requests); err != nil {
+			return false, err
+		}
+
+		// Validate and persist importer-managed labels against the real Pod state only
+		// after the workload construction inputs have passed import validation.
+		oldLq, found := p.Labels[controllerconstants.QueueLabel]
+		if !found {
```
dkaluza: if it is found, shouldn't we also add the labels?

cmd/importer/pod/import.go
```diff
@@ -241,3 +279,57 @@ func admitWorkload(ctx context.Context, c client.Client, wl *kueue.Workload, cq
...
 	return nil
 }
+
+func preparePodForWorkload(p *corev1.Pod, queue string, addLabels map[string]string) *corev1.Pod {
+	podForWorkload := p.DeepCopy()
+	if podForWorkload.Labels == nil {
+		podForWorkload.Labels = make(map[string]string)
+	}
+	maps.Copy(podForWorkload.Labels, addLabels)
+	if _, found := podForWorkload.Labels[controllerconstants.QueueLabel]; !found {
```
dkaluza: in this way we are checking condition that pod has queue label multiple times. 

Cannot we just have a condition that if they differ to return error even before calling preparePodForWorkload?
In this way we will make the rest of the code less complex as we can just always assign it.

cmd/importer/pod/import.go
```diff
@@ -241,3 +279,57 @@ func admitWorkload(ctx context.Context, c client.Client, wl *kueue.Workload, cq
...
+		podForWorkload.Labels[controllerconstants.QueueLabel] = queue
+	}
+	return podForWorkload
+}
+
+// resourceFlavorForResource returns the first flavor from the first resource group
+// that covers the requested resource. It skips groups with no flavors and returns
+// an empty string when no matching group exists.
+func resourceFlavorForResource(cq *kueue.ClusterQueue, resource corev1.ResourceName) kueue.ResourceFlavorReference {
+	for _, rg := range cq.Spec.ResourceGroups {
```
dkaluza: we are possibly going  through this list multiple times. As we will do this for each pod and each cluster queue and each resource name in pod.

Maybe we should just materialize it once whole import?

cmd/importer/pod/import.go
```diff
@@ -241,3 +279,57 @@ func admitWorkload(ctx context.Context, c client.Client, wl *kueue.Workload, cq
...
+	}
+	return ""
+}
+
+func flavorAssignmentsForRequests(cq *kueue.ClusterQueue, requests resources.Requests) (map[corev1.ResourceName]kueue.ResourceFlavorReference, error) {
+	flavors := make(map[corev1.ResourceName]kueue.ResourceFlavorReference)
+
+	resourceNames := make([]corev1.ResourceName, 0, requests.Len())
+	requestQuantities := make(map[corev1.ResourceName]int64, requests.Len())
+	requests.ForEach(func(name corev1.ResourceName, quantity int64) {
```
dkaluza: I'm assuming we do not care about quotas here, am I right?

dkaluza (commented):
Looked into the implementation for now - I will take a look into  the tests when we consolidate on the implementation, as those may still change.

mimowo:
@mszadkow ptal

mimowo:
/lgtm cancel
so that I get notified again with lgtm when the comments are addressed

cmd/importer/pod/check.go
mszadkow: That's good point, thanks

cmd/importer/pod/import.go
mszadkow: Yes, quotas are not checked anywhere in this path, these pods are already running and consuming real capacity.

cmd/importer/pod/import.go
mszadkow: NewWorkload set the name already, we simply keep it

cmd/importer/pod/import.go
mszadkow: thanks for spotting the gap

cmd/importer/pod/import.go
mszadkow: yep, will be added to the importCache

cmd/importer/cache/cache.go
```diff
@@ -46,6 +50,15 @@ type ImportCache struct {
...
 	AddLabels       map[string]string
+
+	flavorValidationMu sync.Mutex
+	flavorValidation   map[string]error
+
+	flavorsByResourceMu sync.Mutex
+	flavorsByResource   map[string]map[corev1.ResourceName]kueue.ResourceFlavorReference
+
+	resourceGroupsMu sync.Mutex
+	resourceGroups   map[string][]resourcegroups.ResourceGroup
```
mimowo: Either wrap the cache in one big R/W lock, or encapsulate the locking per struct

mszadkow: yes, that will be better, thanks
OTOH we could prep it all on Load and use it later

mszadkow:
@dkaluza all addressed, I have changed the approach and now all that can be is calculated at `Load`

cmd/importer/cache/cache.go
```diff
@@ -46,6 +49,10 @@ type ImportCache struct {
 	ResourceFlavors map[kueue.ResourceFlavorReference]*kueue.ResourceFlavor
 	PriorityClasses map[string]*schedulingv1.PriorityClass
 	AddLabels       map[string]string
+
+	// Derived from ClusterQueues and ResourceFlavors at Load time; read-only after that.
+	FlavorValidation  map[string]error
+	FlavorsByResource map[string]map[corev1.ResourceName]kueue.ResourceFlavorReference
```
dkaluza: nit: can we have information here in a comment that first key is CQ name or incorporate this somehow in the field name?

Right now it is hard to guess this without looking at usage.

Same for `FlavorValidation`

cmd/importer/cache/cache.go
```diff
@@ -46,6 +49,10 @@ type ImportCache struct {
 	ResourceFlavors map[kueue.ResourceFlavorReference]*kueue.ResourceFlavor
 	PriorityClasses map[string]*schedulingv1.PriorityClass
 	AddLabels       map[string]string
+
+	// Derived from ClusterQueues and ResourceFlavors at Load time; read-only after that.
```
dkaluza: nit: "read-only after that."
I guess this is an intention, but you actually cannot force this when making this field public.
Consider adding appropriate getter and making this private.

cmd/importer/cache/cache.go
```diff
@@ -83,10 +90,67 @@ func Load(ctx context.Context, c client.Client, namespaces []string, mappingRule
...
+		}
+	}
+	return m
+}
+
+// validateFlavors checks that every ResourceFlavor referenced by rgs is known.
+func validateFlavors(cqName string, rgs []resourcegroups.ResourceGroup, resourceFlavors map[kueue.ResourceFlavorReference]*kueue.ResourceFlavor) error {
+	// Sorted for a deterministic error message; AllFlavors returns a set.
+	flavors := resourcegroups.AllFlavors(rgs).UnsortedList()
+	slices.Sort(flavors)
```
dkaluza: nit: isn't this enough to just use List() set method?

cmd/importer/cache/cache.go
```diff
@@ -83,10 +90,67 @@ func Load(ctx context.Context, c client.Client, namespaces []string, mappingRule
...
+	for _, flavor := range flavors {
+		if _, found := resourceFlavors[flavor]; !found {
+			return fmt.Errorf("%q flavor %q: %w", cqName, flavor, ErrCQInvalid)
```
dkaluza: this will lead to error on first incorrect flavor, consider accumulating those - otherwise UX is impacted - user has to potentially run import multiple times to find all not present flavors.

cmd/importer/cache/cache_test.go
```diff
@@ -0,0 +1,89 @@
...
+			clusterQueue: utiltestingapi.MakeClusterQueue("cq").
+				ResourceGroup(
+					*utiltestingapi.MakeFlavorQuotas("cpu-flavor").
+						Resource(corev1.ResourceCPU, "10", "0").
+						Obj(),
+				).Obj(),
+			want: map[corev1.ResourceName]kueue.ResourceFlavorReference{
+				corev1.ResourceCPU: "cpu-flavor",
+			},
+		},
```
dkaluza: I do not get it, what are we omitting here?  Memory resource? or there  is some other intention?

cmd/importer/pod/check.go
```diff
@@ -78,3 +81,122 @@ func Check(ctx context.Context, c client.Client, importCache *cache.ImportCache,
...
+		return nil, nil, false, fmt.Errorf("cluster queue not found in cache: %s: %w", lq.Spec.ClusterQueue, cache.ErrCQNotFound)
+	}
+	return lq, cq, false, nil
+}
+
+// checkPodWorkload validates that p can be imported into cq and builds the
+// Workload it would produce. Check calls this and only reports the outcome;
+// Import calls this and, on success, persists Pod labels and creates/admits
+// the Workload. Keeping the validation in one place means the two commands
+// cannot silently disagree on whether a Pod is importable.
```
dkaluza: IMO we shouldn't describe the behavior of the caller in helpers documentation. This can get quickly out of date . Business requirements are evolving,  somebody will be changing the import and will not look here at all.

cmd/importer/README.md
```diff
@@ -29,7 +29,10 @@ The importer will perform following checks:
 - For every Pod a  mapping to a LocalQueue is available.
 - The target LocalQueue exists.
 - The LocalQueues involved in the import are using an existing ClusterQueue.
-- The ClusterQueues involved have at least one ResourceGroup using an existing ResourceFlavor. This ResourceFlavor is used when the importer creates the admission for the created workloads.
+- The ClusterQueues involved have ResourceGroups that reference existing ResourceFlavors.
+- For each Pod, the check validates that every non-zero requested resource is covered by the target ClusterQueue ResourceGroups.
+- For each Pod, the check validates that Workload construction succeeds (using the same construction path used by import) before any Pod mutation.
```
dkaluza, line 33: What if resource is excluded from the Kueue managed resources by configuration?  Are we wishing to support it? (otherwise this change can be treated as regression in specific circumstances...)

dkaluza (commented):
Sorry didn't manage to go fully through the implementation today, will try to follow up tomorrow with the remaining files.

mimowo:
/approve
Leaving lgtm to @dkaluza. Thank you for working on that 👍 
/cherrypick release-0.19
/cherrypick release-0.18

cmd/importer/cache/cache.go
mszadkow: good point, `ClusterQueueReference` can be used to show the info

cmd/importer/cache/cache.go
mszadkow: right, thanks

cmd/importer/cache/cache.go
mszadkow: Yes, as the List() call is sorted, it's cleaner, thanks!

cmd/importer/cache/cache_test.go
mszadkow: It means do not include keys for resources that are not covered by any resourceGroup in the ClusterQueue spec.
But I will update test case name.

cmd/importer/README.md
mszadkow: Thanks for that point!
I think we should decide explicitly which behavior we want:
1. Keep the new stricter importer validation and document that it does not honor excludeResourcePrefixes.
2. Align importer behavior with Kueue resource handling and ignore excluded resources during this validation.
@dkaluza @mimowo Which direction would you prefer?

dkaluza: IMO generally 2 is better from the user perspective.
But I have not looked into how problematic this is from the implementation perspective.

mszadkow: Not much problematic, it is however required to pass the excludedResourcePrefixes as a flag.
So it's just wiring, because this will be used as an option to the Workload Info

mszadkow: I keep changes in separate commit, hope that helps reviewing ;)

mszadkow:
@dkaluza ptal :)

cmd/importer/pod/check.go
```diff
@@ -78,3 +81,120 @@ func Check(ctx context.Context, c client.Client, importCache *cache.ImportCache,
...
+	if wl.Labels == nil {
+		wl.Labels = make(map[string]string)
+	}
+	maps.Copy(wl.Labels, importCache.AddLabels)
+
+	info := workload.NewInfo(wl, importCache.WorkloadInfoOptions()...)
+	if len(info.TotalRequests) == 0 {
+		return nil, fmt.Errorf("workload has no total requests: %w", cache.ErrPodInvalid)
+	}
+	flavors, err := flavorAssignmentsForRequests(importCache.FlavorsByResourceForClusterQueue(kueue.ClusterQueueReference(cq.Name)), cq.Name, info.TotalRequests[0].Requests)
```
dkaluza, line 138: I'm assuming total requests from the workload info does not include excluded resources, is this right?

cmd/importer/pod/check_test.go
```diff
@@ -100,40 +98,135 @@ func TestCheckNamespace(t *testing.T) {
...
+					ResourceGroup(*utiltestingapi.MakeFlavorQuotas("rf1").Resource(corev1.ResourceCPU, "1").Obj()).
+					Obj(),
+			},
+			flavors: []kueue.ResourceFlavor{
+				*utiltestingapi.MakeResourceFlavor("rf1").Obj(),
+			},
+			wantError: &resourceNotCoveredError{Resource: corev1.ResourceName("nvidia.com/gpu"), ClusterQueue: "cq1"},
+		},
+		"excluded resource request is ignored": {
+			pods:        []corev1.Pod{*basePodWrapper.Clone().Request(corev1.ResourceName("vendor.com/special"), "1").Obj()},
```
dkaluza, line 145: how does this work?

Shouldn't this lead to `ErrPodInvalid` since there are no requests then?

Or there is an entry with empty requests and this is why this passes but flavors will be empty?
Is workload with empty flavors permitted? I'm guessing it is not. Is the current behavior intended?

Btw. can we have a test with `ErrPodInvalid` - I couldn't find it right now...

cmd/importer/pod/import.go
```diff
@@ -120,37 +109,65 @@ func checkError(err error) (retry, reload bool, timeout time.Duration) {
 	return false, false, 0
 }
 
+// waitForRetry blocks for timeout, or returns early with an error if ctx is
+// done first. A negative timeout returns immediately.
+func waitForRetry(ctx context.Context, timeout time.Duration) error {
+	if timeout < 0 {
```
dkaluza:
```suggestion
	if timeout <= 0 {
```

Maybe it is worth it to return immediately for non-positive timeout?

IIUC 0 still means retry immediately

cmd/importer/pod/import.go
```diff
@@ -120,37 +109,65 @@ func checkError(err error) (retry, reload bool, timeout time.Duration) {
 	return false, false, 0
 }
 
+// waitForRetry blocks for timeout, or returns early with an error if ctx is
+// done first. A negative timeout returns immediately.
```
dkaluza:
```suggestion
// done first. A non-positive timeout returns immediately.
```

cmd/importer/pod/import.go
```diff
@@ -165,78 +182,72 @@ func createWorkload(ctx context.Context, c client.Client, wl *kueue.Workload) er
...
+		})
 		return true, nil
 	}
 
-	for {
+	const maxAttempts = 5
+	for range maxAttempts {
 		err := workloadpatching.PatchAdmissionStatus(ctx, c, wl, realClock, update, workloadpatching.WithForceApply())
-		retry, _, timeout := checkError(err)
+		retry, reload, timeout := checkError(err)
```
dkaluza, line 239: What happens if PatchAdmissionStatus returned `nil`? Is it handled gracefully by `checkError`?

This is not trivial right now for me that this code returns nil if there was no error in status patching at all.
Can we add an explicit return nil when error is nil?

cmd/importer/pod/import_test.go
```diff
@@ -240,3 +446,61 @@ func TestImportNamespace(t *testing.T) {
 		})
 	}
 }
+
+func TestFlavorAssignmentsForRequests(t *testing.T) {
```
dkaluza: Why this test is in import_test when tested function is not in import.go?

dkaluza:
@mszadkow reviewed, generally looks good, but I'm still concerned about this ErrPodInvalid, see https://github.com/kubernetes-sigs/kueue/pull/12924#discussion_r3796883210

cmd/importer/pod/check.go
mszadkow: yes, the "filtering" happens inside workload.NewInfo()

cmd/importer/pod/check_test.go
mszadkow: > Shouldn't this lead to ErrPodInvalid since there are no requests then?

No. ErrPodInvalid is only returned when len(info.TotalRequests) == 0, i.e. when the workload has zero PodSets. It is not returned when a PodSet exists but its Requests map is empty.

> Or there is an entry with empty requests and this is why this passes but flavors will be empty?

Yes — this is exactly what happens. The pod becomes one PodSet, so TotalRequests has length 1 (len(info.TotalRequests) == 1, not 0). That one entry's Requests map is empty because the only requested resource (vendor.com/special) was dropped by the exclusion filter. So the ErrPodInvalid check is skipped, flavorAssignmentsForRequests iterates over zero requests, and it returns an empty flavors map with no error.

> Is workload with empty flavors permitted?

It's a hard question, it's not directly forbidden - a Workload can be admitted with an empty Flavors map for a PodSet, but it varies with outcome.

> Btw. can we have a test with ErrPodInvalid - I couldn't find it right now...

Yes, but it needs a case where wl.Spec.PodSets is empty, which a single plain pod cannot produce (a pod always yields exactly one PodSet, regardless of what it requests). To hit ErrPodInvalid you'd need to go through ConstructComposableWorkload with an input that results in zero constructed PodSets


However, it seems to be a dead code condition anyways.
I will drop it
AI suggests:
Importer is a pod-by-pod tool, entirely blind to Kueue's pod-group feature. 
Each Pod is treated in isolation via pod.FromObject (which never sets isGroup), which is precisely why ConstructComposableWorkload always takes the single-pod branch and always yields exactly one PodSet.
 The `ErrPodInvalid` (zero-PodSets) guard in check.go:138-140 isn't guarding against some importer-specific edge case — it's dead code for a more general path that the importer's current single-pod design never exercises.

cmd/importer/pod/import.go
mszadkow: right, I will add it

cmd/importer/pod/import_test.go
mszadkow: good point, thanks

mszadkow:
@dkaluza as explained [here](https://github.com/kubernetes-sigs/kueue/pull/12924#discussion_r3797454318), I decided to drop ErrPodInvalid completely as it was unreachable code path.

cmd/importer/pod/check_test.go
dkaluza: nit: Then it may be worth a comment in code that pods are guaranteed to have at least one request.

dkaluza (approved):
/lgtm

mszadkow:
I will do CPs manually

tenzen-y:
/release-note-edit
```release-note
Importer: Fixed a bug where the Pod importer picked a single ResourceFlavor for the whole Pod, so Pods whose resources map to different flavors could be imported with a wrong flavor assignment. Flavors are now resolved per requested resource.
```

PR #12933, opened 2026-07-10
[release-0.17] Fix AFS cache initialization and warm-start recovery (#12891)

apullo777:
* Seed AFS consumed-resources cache from persisted status instead of live usage

* Seed AFS cache with SetIfAbsent to avoid overwriting concurrent settlement

* Guarantee AFS warm-start recovery via atomic Update and StatusAccounted

(cherry picked from commit 6f08701fc15de200a91c6c3e3f305832aba15dd6)

Backport adaptations for release-0.17:
- localqueue_controller.go: reuse the existing "resource" import alias instead of the "utilresource" alias added on main (same util/resource package).
- Test helpers: use ptr.To(resource.MustParse(...)) instead of the Go 1.26 new(expr) form, since release-0.17 targets go 1.25.
- workload_controller_test.go: dropped the unused setupDRACache helper pulled in by the merge; it references dra.ExtendedResourceCache, which does not exist on release-0.17, and the backported test does not use it.
- fair_sharing_test.go: inline the AdmissionFairSharing config at the call site (release-0.17 has no afsConfig suite helper), matching the existing managerAndSchedulerSetup callers on this branch.

#### What type of PR is this?

/kind bug

cheery-pick of https://github.com/kubernetes-sigs/kueue/pull/12891

Fixes #

```release-note
AFS: Fixed consumed-resources cache initialization and warm-start recovery so LocalQueue usage is not over-counted during cache seeding, and persisted historical usage is preserved after manager restarts when workload settlement runs before LocalQueue reconciliation.
```

### DIFF
--- a/pkg/cache/queue/afs/consumed_resources.go
+++ b/pkg/cache/queue/afs/consumed_resources.go
@@ -25,10 +25,15 @@ import (
 	utilqueue "sigs.k8s.io/kueue/pkg/util/queue"
 )
 
-// ConsumedResourcesEntry stores the consumed resources for a LocalQueue with timestamp
+// ConsumedResourcesEntry stores the consumed resources for a LocalQueue with timestamp.
+// StatusAccounted records whether the LocalQueue's persisted fair-sharing status has
+// been folded into this entry: once history is merged, the entry's existence alone no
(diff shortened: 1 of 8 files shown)


mimowo:
/lgtm
/approve
Thank you 👍

PR #12939, opened 2026-07-10
Clamp AFS tick status computation and cover future LastUpdate

apullo777:
#### What type of PR is this?

/kind bug

This is a follow-up to the review discussion in https://github.com/kubernetes-sigs/kueue/pull/12891#discussion_r3554510089, which asked for unit coverage of the future-`LastUpdate` / elapsed-clamp guard added on the settlement side.

This PR has two commits:

1. add unit coverage for the AFS settlement-side future-`LastUpdate` guard;
2. fix one remaining call site with the same shape: the sampling tick's pre-lock
   status computation.

#12891 clamps elapsed time in the workload-settlement path and in the sampling tick's in-lock cache recompute. While adding the settlement-side test, I noticed that the tick's pre-lock status computation still used raw elapsed time. That value is persisted to the LocalQueue fair-sharing status.

If a concurrent writer stamps `LastUpdate` later than the tick's captured `now`, the raw elapsed time becomes negative and the decay factor can go outside the expected range. This PR applies the same `max(0, elapsed)` clamp there.

The new tests assert the settlement-side guard and the tick status-computation guard, including persisted status, cached entry, and timestamp monotonicity.

Related to #12891

Two cases intentionally do not get separate test rows:

- the settlement cache-miss branch cannot have a future `LastUpdate`, because `!found` initializes `lastUpdate` from `now`
- the half-life-zero reset path is unaffected by negative elapsed, because the decay factor is constant there

```release-note
AFS: Fixed a race where a sampling tick running concurrently with workload settlement could persist a skewed ConsumedResources value in LocalQueue fair-sharing status.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed consumed-usage decay when timestamps from concurrent updates are in the future.
  * Prevented inflated usage calculations and preserved the latest valid update timestamp.
  * Ensured pending usage penalties are incorporated correctly in these scenarios.

* **Tests**
  * Added coverage for future timestamp handling during queue reconciliation and workload usage updates.

### DIFF
--- a/pkg/controller/core/localqueue_controller.go
+++ b/pkg/controller/core/localqueue_controller.go
@@ -415,7 +415,14 @@ func (r *LocalQueueReconciler) reconcileConsumedUsage(ctx context.Context, lq *k
 
 	oldUsage := entry.Resources
 	newUsage := cacheLq.GetAdmittedUsage()
-	elapsed := now.Sub(entry.LastUpdate).Seconds()
+	// A concurrent settlement can stamp an entry's LastUpdate later than now.
+	// A negative elapsed would drive the decay alpha outside [0, 1] and inflate
+	// consumed usage, so every elapsed derived from a stored LastUpdate is
+	// clamped to a non-negative value (here, the in-lock recompute below, and
+	// updateAfsConsumedUsage). Those cache writes also keep the stored timestamp
+	// monotonic so the next decay does not over-elapse; this pre-lock path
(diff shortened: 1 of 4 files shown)


mimowo:
/lgtm
/approve
Thank you for the quick follow up to sanitize the timestamps used for AFS 👍 

Hope the robot can CP, but please prep manually if it doesn't
/cherrypick release-0.18
/cherrypick release-0.17

PR #12909, opened 2026-07-09
dra: saturate counter-charge multiplication to avoid int64 overflow

thc1006:
#### What type of PR is this?
/kind bug

#### What this PR does / why we need it:
`computeCounterCharges` (`pkg/dra/counters.go`) multiplied a device's counter value by the request count with a plain int64 multiply (`intVal*count`), which can wrap to a negative quantity and under-charge the counter quota resource. This uses `utilmath.SaturatingMul`, consistent with the `SaturatingAdd` applied to `countDevicesPerClass` in #12897.

This is a defense-in-depth / consistency fix. Unlike #12896, it is not reachable from the tenant trust boundary: `count` is bounded upstream by the matched-device shortage check and `intVal` is a driver-published counter value, so overflow would require a driver publishing a near-`MaxInt64` counter value together with `count > 1`. Details in #12908.

#### Which issue(s) this PR fixes:
Fixes #12908

#### Special notes for your reviewer:
Follow-up to #12897 / #12896. Adds a regression test case (`count multiplication saturates instead of overflowing int64`) to `TestComputeCounterCharges`.

#### Does this PR introduce a user-facing change?
```release-note
DRA: fixed a potential int64 overflow in the counter-based device quota charge computation that could under-count quota when a driver publishes very large counter values.
```

PR #12945, opened 2026-07-10
dra: clamp driver counter values to [0, MaxInt64] before charging quota

thc1006:
#### What type of PR is this?
/kind bug

Follow-up to #12909. `computeCounterCharges` reads a counter value straight from a ResourceSlice (`ConsumesCounters[].Counters[name].Value`), which the apiserver does not validate — `validateDeviceCounter` accepts "any parsed quantity" — so a driver can publish a negative counter, or one whose magnitude exceeds `int64`.

The `SaturatingMul` added in #12909 does not cover two paths:
- `maxValue.Value()` truncates an out-of-`int64` quantity to garbage (`1e19` → `0`, `2^63` → a negative) **before** the multiply, so saturating the product cannot recover it.
- The saturating multiply is gated on `count > 1`, so a single-device request passes an out-of-`int64` or negative quantity straight through as the charge.

This adds `utilmath.SafeValue` (the twin of `SafeMilliValue`) to clamp `Value()` to `[MinInt64, MaxInt64]`, and in `computeCounterCharges` drops negatives so the charge is always in `[0, MaxInt64]` for every count, including `count == 1`. Regression tests cover the truncation, the `count == 1` path, and the negative case; each fails on the pre-fix code.

This is defense-in-depth for the driver-published counter path; it is not reachable from the tenant trust boundary.

A note on why a negative counter clamps to `0` (rather than being rejected or clamped to `MaxInt64`): a DRA driver is a node-level component inside the trust boundary — it already controls its ResourceSlice and could even misreport capacity — so a negative counter is a driver bug, not a tenant attack. Clamping to `0` fails safe (the buggy device contributes no counter charge), whereas clamping to `MaxInt64` would let a single harmless driver bug stall admission for the entire ClusterQueue.

Follow-up to #12909.

```release-note
DRA: Fixed incorrect quota charging for invalid driver-published device counters by clamping them to the non-negative 
int64 range before computing quota charges.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected handling of extreme counter values to prevent truncation and overflow.
  * Negative counter values are now treated as zero.
  * Counter values outside the supported 64-bit range are safely clamped.
  * Large multiplied counter charges now saturate instead of overflowing.

* **Tests**
  * Added coverage for oversized, undersized, negative, and multiplied counter values.

### DIFF
--- a/pkg/dra/counters.go
+++ b/pkg/dra/counters.go
@@ -303,11 +303,18 @@ func computeCounterCharges(
 		return nil
 	}
 
+	// A driver-published counter value is an unvalidated resource.Quantity: the
+	// apiserver accepts any parseable value, including negatives and magnitudes
+	// beyond int64 (maxValue.Value() would then truncate to garbage before the
+	// saturating multiply could clamp it). Clamp with SafeValue and drop
+	// negatives so the charge stays in [0, MaxInt64] for every count, including
+	// count == 1.
+	intVal := max(utilmath.SafeValue(maxValue), 0)
 	if count > 1 {
-		intVal := maxValue.Value()
 		// Saturate rather than let intVal*count wrap to a negative quantity,
 		// consistent with countDevicesPerClass after #12897.
-		maxValue = *resource.NewQuantity(utilmath.SaturatingMul(intVal, count), maxValue.Format)
+		intVal = utilmath.SaturatingMul(intVal, count)
 	}
+	maxValue = *resource.NewQuantity(intVal, maxValue.Format)
 	return corev1.ResourceList{quotaResource: maxValue}
 }
--- a/pkg/dra/counters_test.go
+++ b/pkg/dra/counters_test.go
@@ -172,6 +172,44 @@ func TestComputeCounterCharges(t *testing.T) {
 				"gpu.memory": *resource.NewQuantity(math.MaxInt64, resource.DecimalSI),
 			},
 		},
+		{
+			// 1e19 > MaxInt64; Value() truncates it to 0, so without SafeValue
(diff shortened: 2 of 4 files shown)


mimowo (commented):
/lgtm
/approve
Nice one, thank you for hardening Kueue-DRA integration 👍 
/cherrypick release-0.18
/cherrypick release-0.17

thc1006:
Thank you @mimowo, I really appreciate the quick review and the kind words 🙏

One small heads-up on the backports: the counter-charge code (`pkg/dra/counters.go`) only exists from release-0.18 onward, so the release-0.17 cherry-pick won't apply here, same as #12909. I believe 0.18 is the only backport needed. Thanks again!

mimowo:
```
level=warning msg="[runner/exclusion_rules] Skipped 0 issues by rules: [Text: \"SA1019: features.* is deprecated: planned to be removed in .*\", Linters: \"staticcheck\"]"
pkg/dra/counters.go:313:5: minmax: if statement can be modernized using max (modernize)
	if intVal < 0 {
```

mimowo:
> Thank you @mimowo, I really appreciate the quick review and the kind words 🙏

You are welcome 🌷 would you like to consider adding your company to the adopters list https://kueue.sigs.k8s.io/community/adopters/? This would be especially interesting to also mention DRA use

mimowo:
/release-note-edit
```release-note
DRA: Fixed incorrect quota charging for invalid driver-published device counters by clamping them to the non-negative 
int64 range before computing quota charges.
```

PR #12892, opened 2026-07-09
Release process - tagging

Singularity23x0:
#### What type of PR is this?
/kind documentation

#### What this PR does / why we need it:
Adds a script to automate release tag cutting and pushing.
Enhances logging by Actiosn via issue comments.

#### Which issue(s) this PR fixes:

Fixes #12974
Contributes to #12763

#### Special notes for your reviewer:
Generated with the use of AI.

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added a ChatOps `/tag-release` command to automate creating and pushing release tags from a release proposal issue.
  * Updated the release checklist to use `/tag-release`, generating the release tag message from the issue’s changelog.
* **Improvements**
  * Release workflow now recognizes `/tag-release` and routes it to the tagging step to trigger the downstream staging image build/publish flow.
  * Added clearer feedback for missing changelogs and tag/push failures (`❌`/`✅`).
  * Kept existing release coordination commands working as before.

PR #12950, opened 2026-07-10
Add release-team to OWNERS_ALIASES.

mbobrovskyi:
#### What type of PR is this?
/kind cleanup

Add release-team to OWNERS_ALIASES.

Part of #12763

```release-note
NONE
```

## Summary by CodeRabbit

* **Chores**
  * Updated release authorization to use maintained ownership aliases for approved release contributors.
  * Added designated approvers for CHANGELOG updates.
  * Added release-team alias entries and removed the legacy release-team listing.
  * Unauthorized release actions now fail verification with an authorization error.

### DIFF
--- a/.github/workflows/release-utils.yml
+++ b/.github/workflows/release-utils.yml
@@ -19,13 +19,19 @@ jobs:
       - name: Verify Actor Authorization
         env:
           ACTOR: ${{ github.actor }}
+          OWNERS_FILE: OWNERS_ALIASES
         run: |
-          TEAM_FILE="release-team.json"
-          if [ ! -f "$TEAM_FILE" ]; then
-            echo "Error: Allowed actors list file $TEAM_FILE not found."
+          if [ ! -f "$OWNERS_FILE" ]; then
+            echo "Error: $OWNERS_FILE not found."
             exit 1
           fi
-          if ! jq -e --arg actor "$ACTOR" '. | index($actor)' "$TEAM_FILE" > /dev/null; then
+          if ! python3 -c "
+          import os, yaml
+
+          aliases = yaml.safe_load(open(os.environ['OWNERS_FILE'])).get('aliases', {})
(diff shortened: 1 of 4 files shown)


mbobrovskyi:
I tested it in my local fork https://github.com/mbobrovskyi/kueue/actions/runs/29074964753.

CHANGELOG/OWNERS
mbobrovskyi, line 1: We should probably add permissions in more places.

mimowo:
/lgtm
/approve
Thank you for the cleanup, I will leave one follow up comment: https://github.com/kubernetes-sigs/kueue/pull/12950#discussion_r3558156989

.github/workflows/release-utils.yml
```diff
@@ -19,13 +19,19 @@ jobs:
...
+            echo "Error: $OWNERS_FILE not found."
             exit 1
           fi
-          if ! jq -e --arg actor "$ACTOR" '. | index($actor)' "$TEAM_FILE" > /dev/null; then
+          if ! python3 -c "
+          import os, yaml
+
+          aliases = yaml.safe_load(open(os.environ['OWNERS_FILE'])).get('aliases', {})
+          authorized = set(aliases.get('release-team', [])) | set(aliases.get('kueue-approvers', []))
+          raise SystemExit(0 if os.environ['ACTOR'] in authorized else 1)
```
mimowo, line 33: I imagine this logic will need to be repeated for multiple GH actions, so ultimately it would be good to extract it to some common place like a helper python script, to avoid duplication

mbobrovskyi: IIUC, we'll have a separate job for this: https://github.com/kubernetes-sigs/kueue/pull/12892/changes#diff-19e37a3b45ae52d8b17c6cac4dea9ef5df473839c2b2f9c34a08ad5f1e1a8473R6-R35.

I think, if we end up duplicating this code in the future, we can extract it into a separate action.

Issue #12750, opened 2026-07-03
Review & Tidy CPU usage in CI jobs


mimowo:
**What would you like to be cleaned**:

I would like to revisit all our CI jobs to check if we could reduce the CPU requested usage in k8s/test-infra.

Please use grafana monitoring dashboards to assess if we really use the CPU we request: 
https://monitoring-eks.prow.k8s.io/d/96Q8oOOZk/builds?orgId=1&from=now-7d&to=now&var-org=kubernetes-sigs&var-repo=kueue&var-job=pull-kueue-test-e2e-main-1-31&var-build=All&refresh=30s

This was suggested in slack discussion: https://kubernetes.slack.com/archives/CCK68P2Q2/p1782979216407899

Prior related work: https://github.com/kubernetes/k8s.io/pull/9678

**Why is this needed**:

to help mitigating the problem: https://github.com/kubernetes-sigs/kueue/issues/12739

mimowo:
The stats on the CPU consumption by our Jobs which can guide the optimizations (from the slack https://kubernetes.slack.com/archives/CCK68P2Q2/p1782979216407899):
<img width="3020" height="1614" alt="Image" src="https://github.com/user-attachments/assets/55849729-c634-4330-82f3-6f079debd99d" />

<img width="3020" height="1614" alt="Image" src="https://github.com/user-attachments/assets/9d90e73a-8452-4e86-a6f7-b56eaf9ef4a4" />

reruno:
/assign 

cc @mbobrovskyi

reruno:
<img width="1430" height="1430" alt="Image" src="https://github.com/user-attachments/assets/9a031303-fc43-4659-9f23-e34de651937b" />
<img width="1430" height="1430" alt="Image" src="https://github.com/user-attachments/assets/0e1f3670-bc37-469a-8295-7fd9ca0b0ab0" />

for pull-kueue-test-e2e-multikueue-extended-shard-1-main_30d_30s

I downloaded data for 2 weeks for pull-kueue-test-e2e-multikueue-extended-shard-1-main job, and created histograms (attached diagrams). I think we can lower requested cpu based on p99 of average cpu usage of the builds and leave limit as it is based on peaks of the jobs.

I will download all job data, make analysis and prepare suggestions for all pull-* jobs

mimowo:
>  I think we can lower requested cpu based on p99 of average cpu usage of the builds and leave limit as it is based on peaks of the jobs.

[...]

PR #12951, opened 2026-07-10
Add hack/infra/stats tooling to right-size Kueue prow CI jobs

reruno:
#### What type of PR is this?
/kind feature

Add hack/infra/stats tooling to right-size Kueue prow CI jobs

Related to #12750 

`collects_stats.sh ` should be called from root of kueue project, it saves the data and diagrams to `artifacts/infra-stats/` folder 

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added an end-to-end CI right-sizing workflow to fetch CPU/memory (and best-effort network) metrics and generate per-job diagnostics plus a recommendation.
  * Added job selection (exact/regex), resumable runs with a force option, concurrency, retries, and configurable time ranges.
  * Added headless plot generation (usage distributions, throttling/pressure and network timelines) and an aggregated Markdown table comparing current vs recommended CPU and memory.
* **Documentation**
  * Added a guide describing the workflow steps, required settings, collected metrics, and how to interpret the diagrams.

### DIFF
--- a/hack/infra/stats/aggregate_reco.py
+++ b/hack/infra/stats/aggregate_reco.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python3
+
+# Copyright The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+aggregate_reco.py — combine every <workdir>/*/recommendation.json (written by
+plot.py) into a single recommendation.md table in the work directory.
+
+The table shows, per job, the current vs recommended CPU/memory request and limit,
+the CPU cores saved by the recommendation, and a TOTAL row summing the savings.
+
+Example:
+  ./aggregate_reco.py ./artifacts/infra-stats
+"""
+import argparse, glob, json, os
+
+
+def num(v):
+    return "?" if v is None else f"{v:g}"
+
+
+def main():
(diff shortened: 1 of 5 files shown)


hack/infra/stats/aggregate_reco.py
```diff
@@ -0,0 +1,99 @@
...
+    files = sorted(glob.glob(os.path.join(args.workdir, "*", "recommendation.json")))
+    rows = []
+    for f in files:
+        with open(f) as fh:
+            rows.append(json.load(fh))
+    if not rows:
+        raise SystemExit(f"no recommendation.json under {args.workdir}/*/ — run the plot scripts first")
+
+    header = ("| job | builds | avg dur (min) | cur cpu req | cur cpu lim | cur mem req | cur mem lim "
+              "| rec cpu req | rec cpu lim | rec mem req | rec mem lim |")
```
mimowo, line 49: Let's make the table more readable
- only produce column for requests, because prow assumes req=lim to it makes the table
- only include columns for CPU by default
- use full names rather than abbreviations

Now, feel free to make the additional coumns optional, if the script is parametrized like `--with-mem` or `--relax-limit-request-constraint`.

I don't have a strong view, but disable by default.

hack/infra/stats/collect_stats.sh
```diff
@@ -0,0 +1,123 @@
...
+    case "$a" in "$1"=*) return 0 ;; esac
+  done
+  return 1
+}
+
+defaults=()
+# --job / --job-regex are a mutually-exclusive group: default only if neither was given
+if ! contains --job && ! contains --job-regex; then defaults+=(--job-regex '^pull-'); fi
+contains --days        || defaults+=(--days 14)
+contains --step        || defaults+=(--step 30s)
```
mimowo, line 88: Can we get more precise reading about peaks if we reduce the step to 5s?

hack/infra/stats/collect_stats.sh
```diff
@@ -0,0 +1,123 @@
...
+contains --out-dir     || defaults+=(--out-dir artifacts/infra-stats)
+contains --concurrency || defaults+=(--concurrency 2)
```
mimowo, line 90: This feels pretty constraining on modern CPUs, also IIUC this is network bound, so maybe let's use 8?

mimowo:
/lgtm
/approve
Let me merge as is. I think my comments are nice usability improvements rather than blocker. The diff will also be easier to review when we have the initial PR merged.

PR #6983, opened 2025-09-24
parse fair weight into float64 upon update

gabesaba:
#### What type of PR is this?
/kind cleanup

#### What this PR does / why we need it:
Follow-up to #6963, to fix #6960

Upon update to ClusterQueue or Cohort, we convert the resource.Quantity representation of FairWeight into float64

#### Which issue(s) this PR fixes:
Fixes #

#### Does this PR introduce a user-facing change?
```release-note
NONE
```

PR #12953, opened 2026-07-10
Fix milli value handling in resource transformations

### LINKED ISSUE #12952: Resource transformations round up milli values before multiplication
**What happened**:

[Resource transformations](https://kueue.sigs.k8s.io/docs/tasks/manage/administer_cluster_quotas/#transform-resources-for-quota-management) round up milli values before multiplication.

With the following config, `300m` CPUs are transformed to `3000` CPU credits, instead of `300`.

```
transformations:
- input: cpu
  outputs:
    example.com/cpu-credits: 1000
  strategy: Retain
```

Resource transformations account for decimal quantity values during multiplication.

- `300m * 1000` -> `300`
- `300M * 1m` -> `300k`

https://github.com/wafrelka/kueue/commit/e9a795905ed531528e4bc5d96e2df2e92c84f35c#diff-a1c9ac01d392160cf78127805c49aad2b49ce36fbe1c1a82984cccd9e8c0fa3bR649-R688

Without the fix, the test fails:

```text
--- FAIL: TestNewInfo (0.00s)
    --- FAIL: TestNewInfo/transformMilliValues (0.00s)
        workload_test.go:697: NewInfo(_) = (-want,+got):
              &workload.Info{
                ... // 1 ignored field
                TotalRequests: []workload.PodSetResources{
                        {
                                Name: "",
                                Requests: resources.Requests{
            -                           s"example.com/cpu-credits":    3000,
            +                           s"example.com/cpu-credits":    300,
                                        s"example.com/memory-credits": 300000,
                                },
                                Count:           1,
                                TopologyRequest: nil,
                                ... // 2 identical fields
                        },
                },
                ClusterQueue:   "",
                LastAssignment: nil,
                ... // 1 ignored and 4 identical fields
              }
FAIL
FAIL    sigs.k8s.io/kueue/pkg/workload  0.011s
FAIL
```

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

wafrelka:
#### What type of PR is this?

/bug

Fixes #12952

```release-note
ResourceTransformations: Fixed a bug where milli-valued quantities were rounded before
resource transformation multiplication. For example, multiplying `300m` CPU by `1000`
now correctly produces `300` instead of `3000`.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource transformation calculations for quantities using milli and mixed units.
  * Ensured transformed CPU and memory values are calculated accurately while preserving their formats.

* **Tests**
  * Added coverage for resource transformations involving milli-valued quantities.

### DIFF
--- a/pkg/workload/workload.go
+++ b/pkg/workload/workload.go
@@ -27,9 +27,11 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	"gopkg.in/inf.v0"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	resourcehelpers "k8s.io/component-helpers/resource"
@@ -528,13 +530,12 @@ func applyResourceTransformations(input corev1.ResourceList, transforms map[core
 			// the value of the resource specified in MultiplyBy.
 			if mapping.MultiplyBy != "" {
 				if q, ok := input[mapping.MultiplyBy]; ok {
-					inputQuantity.Mul(q.Value())
+					inputQuantity = multiplyResourceQuantities(inputQuantity, q)
 				}
 			}
 
 			for outputName, baseFactor := range mapping.Outputs {
-				outputQuantity := baseFactor.DeepCopy()
-				outputQuantity.Mul(inputQuantity.Value())
+				outputQuantity := multiplyResourceQuantities(inputQuantity, baseFactor)
 				if accumulated, ok := output[outputName]; ok {
 					outputQuantity.Add(accumulated)
 				}
@@ -548,6 +549,14 @@ func applyResourceTransformations(input corev1.ResourceList, transforms map[core
 		}
 	}
 	return output
+}
+
+func multiplyResourceQuantities(value, mul resource.Quantity) resource.Quantity {
+	value = value.DeepCopy()
+	mul = mul.DeepCopy()
+	product := inf.Dec{}
+	product.Mul(value.AsDec(), mul.AsDec())
+	return *resource.NewDecimalQuantity(product, value.Format)
 }
 
 func CanBePartiallyAdmitted(wl *kueue.Workload) bool {
--- a/go.mod
+++ b/go.mod
@@ -25,6 +25,7 @@ require (
 	go.uber.org/mock v0.6.0
 	go.uber.org/zap v1.28.0
 	golang.org/x/sync v0.22.0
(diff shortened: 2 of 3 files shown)


wafrelka:
I forgot to run `go mod tidy` earlier, which caused a conflict. I've resolved it and force-pushed.

pkg/workload/workload.go
```diff
@@ -550,6 +551,14 @@ func applyResourceTransformations(input corev1.ResourceList, transforms map[core
 	return output
 }
 
+func multiplyResourceQuantities(value, mul resource.Quantity) resource.Quantity {
+	value = value.DeepCopy()
+	mul = mul.DeepCopy()
```
kshalot, line 556: Are those copies necessary? IIUC `product.Mul` will set `product`, but won't mutate the arguments

pkg/workload/workload.go
```diff
@@ -550,6 +551,14 @@ func applyResourceTransformations(input corev1.ResourceList, transforms map[core
...
+	product := inf.Dec{}
+	product.Mul(value.AsDec(), mul.AsDec())
+	return *resource.NewDecimalQuantity(product, value.Format)
```
kshalot, line 559: I don't think this matters, but in the old code I think the value used the format of the `baseFactor`:
```go
outputQuantity := baseFactor.DeepCopy()
outputQuantity.Mul(inputQuantity.Value())
```
and not the `inputQuantity`. But this shouldn't impact us in this case.

pkg/workload/workload.go
wafrelka: `AsDec` mutates the receiver.
It seems harmless, but just being conservative here.

kshalot: I see. This makes sense. Thanks!

kshalot:
/lgtm

Thanks and good catch!

pkg/workload/workload.go
mimowo: They might be necessary, resource.Quantity has some weird semantics that some operations which look seemingly like "read" could perform mutations on its iternal state, see here: https://github.com/kubernetes-sigs/kueue/pull/6983/changes#diff-94a0ecf880b4ffc718449fd8ae642ff0c5bed4136b5458ac33e95c307d76b43bR174-R180

I don't know if this is the case for sure, but it might be.

pkg/workload/workload.go
mimowo: Here too, I think the DeepCopy is an ok safeguard cutting us from the resource.Quantity semantics. 

Actually, in a follow up we could consider using our Amount type.

mimowo:
/approve
/cherrypick release-0.18
/cherrypick release-0.17
Thank you for hardening of the ResourceTransformations feature 👍

mimowo:
/release-note-edit
```release-note
ResourceTransformations: Fixed a bug where milli-valued quantities were rounded before
resource transformation multiplication. For example, multiplying `300m` CPU by `1000`
now correctly produces `300` instead of `3000`.
```

PR #12954, opened 2026-07-10
dra: test count==1 rounding and log the negative driver-counter clamp

thc1006:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it:**

Follow-up to #12945, which clamped driver-published counter values to `[0, MaxInt64]` before charging quota. Two small gaps around that clamp remained:

- **The `count == 1` fractional-rounding path was untested.** That path charges the counter through `Value()` without the saturating multiply, so a fractional counter (for example `500m`) rounds up to `1`, which is conservative and consistent with the `count > 1` path, but nothing pinned it. This adds that case, plus companion `count == 1` negative and fractional-negative cases.
- **The negative-counter clamp was silent.** #12945 already clamped negatives to `0` but gave no signal. A driver publishing a negative counter is misbehaving and should never happen, so this logs it at `V(0)`, matching how the scheduler cache surfaces an unexpected negative pod count (`pkg/cache/scheduler/tas_non_tas_pod_cache.go`), and adds a `funcr`-based test that the clamp is surfaced. The negative check uses `maxValue.Sign()` so it reflects the driver's value directly; the charge itself is unchanged from #12945.

No change to the charge, which #12945 already clamped to `[0, MaxInt64]`. This is test coverage plus observability.

**Which issue(s) this PR fixes:**

Follow-up to #12945 (no separate issue).

**Special notes for your reviewer:**

`computeCounterCharges` gains a `logr.Logger` parameter rather than a `context.Context`, because it is a pure computation with no I/O, so a logger advertises exactly what it needs. Passing `ctx` only to extract a logger would be misleading. Happy to switch if you would prefer consistency with the `ctx`-taking helpers.

/cc @mimowo

**Does this PR introduce a user-facing change?**

```release-note
NONE
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected negative device counter handling so charges are explicitly set to zero.
  - Improved computation for edge cases, including fractional values and counters equal to one (with correct rounding behavior).

- **Diagnostics**
  - Added verbosity-0 logging when a driver reports an unexpected negative device counter, including driver/counter context and the original counter value.

- **Tests**
  - Expanded coverage for negative and edge-case counter values, including assertions that the expected log message is emitted.

### DIFF
--- a/pkg/dra/counters.go
+++ b/pkg/dra/counters.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	resourcev1 "k8s.io/api/resource/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -159,7 +160,7 @@ func processCounterCharge(
 		)}
 	}
 
-	charges := computeCounterCharges(counterConfig, quotaResource, matched, count)
+	charges := computeCounterCharges(log, counterConfig, quotaResource, matched, count)
 	if len(charges) > 0 {
 		return charges, nil
 	}
@@ -278,6 +279,7 @@ func groupSlicesByPool(slices []resourcev1.ResourceSlice, driver string) map[str
 }
 
 func computeCounterCharges(
+	log logr.Logger,
 	counterConfig *deviceClassCounterConfig,
 	quotaResource corev1.ResourceName,
 	matched []resourcev1.Device,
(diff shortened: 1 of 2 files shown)


mimowo:
> computeCounterCharges gains a logr.Logger parameter rather than a context.Context, because it is a pure computation with no I/O, so a logger advertises exactly what it needs. Passing ctx only to extract a logger would be misleading. Happy to switch if you would prefer consistency with the ctx-taking helpers.

This decision matches my expectations - we can extend to pass ctx if proven to be needed. This also matches our approach for scheduler cache functions

mimowo (commented):
/lgtm
/approve
/cherrypick release-0.18
Thank you for the quick follow up 👍 Logging anomalous values set by the driver makes sense to me.

thc1006:
Thanks a lot, and for the quick reviews today. Glad the driver-anomaly logging made sense to you; that was the part I went back and forth on the most, so it's good to have your read on it.

Issue #11858, opened 2026-06-01
TAS: ungater leaks pods pinned to topology domains removed by a later admission


rahmanitbhu:
### What happened

When a TAS-admitted Workload undergoes a second-pass admission that rewrites `Workload.Status.Admission.PodSetAssignments[*].TopologyAssignment`, for example after a node-failure-driven eviction and re-scheduling, pods that were ungated against the *previous* `TopologyAssignment` can be left pinned (via `pod.spec.nodeSelector`) to topology domains that are no longer present in the *current* `TopologyAssignment`. This is timing-dependent, it only fires when the ungater observes the replacement pod before the TA rewrite.

In our setup the topology's lowest level is `kubernetes.io/hostname`, so each leaf domain corresponds to a single node and "pinned to a domain" is literally "pinned to a hostname". When the failed host is removed from the TA, an already-ungated pod that targeted it remains nailed to that exact hostname.

Once the underlying host recovers, such a pod runs on a domain no longer tracked by Kueue. It consumes capacity invisible to TAS's per-domain accounting, distorting subsequent scheduling decisions for other Workloads on that ClusterQueue/Cohort. If the host does not recover, the pod stays unschedulable until `PodReadyTimeout` fires,  which is recoverable, but still a stranded slot.

### What you expected to happen

Every ungated, non-terminated pod belonging to a `PodSetAssignment` *P* with `P.TopologyAssignment != nil` should be pinned to a domain that is in `P.TopologyAssignment.Domains`. If a `TopologyAssignment` is rewritten in a way that strands previously-ungated pods, the ungater should reconcile the invariant, e.g., by deleting the orphans so the owner controller (Job/JobSet/LWS/...) recreates them, after which the ungater will pin them against the fresh TA.

### How to reproduce

> **Note on reproducibility:** this is **not deterministic** in a simple cluster. The bug depends on the order in which the ungater processes two events: the newly-created replacement pod and the Workload TA update. If the TA update is reconciled *first* and the replacement pod *second*, the pod is gated by the webhook and the (now-current) TA is used to pin it: no orphan, no bug. The orphan only persists when the replacement pod is reconciled (and ungated) *before* the TA update is observed, so the pod ends up pinned against the stale TA and is never revisited. Note that this race can't be eliminated by event ordering alone, because Pods and Workloads are watched independently. In production this fires often enough to leak hosts; in a small cluster you may need to script the failure injection and retry, or instrument the ungater to delay one of the two informer paths.

[...]

PR #5861, opened 2025-07-03
[TAS] Add FailFast mode to TAS on node failure

PBundyra:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
Adds a FailFast mode to TAS on node failure. After Kueue tries to find a replacement for a failed node and fails it evicts the Workload 

#### Which issue(s) this PR fixes:

Fixes #5862

#### Special notes for your reviewer:
Currently the feature is behind feature gate, we will add dedicated API based on customers feedback

#### Does this PR introduce a user-facing change?

```release-note
TAS: Add FailFast on Node's failure handling mode
```

PR #12980, opened 2026-07-10
Fix handling of single-pod owned workloads; add SkipReassignmentForPodOwnedWorkloads FG

### LINKED ISSUE #12735: TAS: node-failure reassignment of admitted pod-owned Workloads causes permanent placement drift
**What happened**:

We run Kueue v0.17.x, using hostname-level TAS for multi-GPU placement (`kueue.x-k8s.io/podset-required-topology:
kubernetes.io/hostname`). Our workloads are
**Deployments managed through the pod integration** (i.e., workloads are pod-owned).

When a node hosting Running TAS pods goes `NotReady` for longer than
`NodeFailureDelay` (hardcoded 30s), the TAS node controller adds the node to
each affected Workload's `status.unhealthyNodes`, and the scheduler's second
pass rewrites the Workload's `topologyAssignment` to a replacement node — while
the pod keeps running on the original node (its injected
`kubernetes.io/hostname` nodeSelector is immutable). The stored assignment now
permanently disagrees with reality. Two damage windows:

1. **Transient outage — `30s < NotReady < ~5m` (pod-eviction timeout).** The
   node recovers before the taint manager evicts the pod, so the pod keeps
   running and **the drift persists indefinitely** — nothing ever reconciles a
   Running pod's assignment back (there is no recovery path for this; see
   #11858). It clears only when something else deletes the pod. For us this is
   the common case: node-network blips of 1–3 minutes rewrite the assignments
   of every Running pod on the node, every time.

2. **Long outage — `NotReady > ~5m`.** The taint manager sets the pods'
   deletionTimestamp at ~300s, but the unreachable kubelet cannot confirm
   deletion, so the pods stick `Terminating` and their Workloads — with
   rewritten assignments — live on until the node returns. The drift
   self-corrects at node recovery (the pod object finally deletes and the
   Workload is garbage-collected with it), but the accounting is wrong for the
   entire outage and black-holes GPUs on healthy hosts until recovery.

The drift corrupts TAS per-node accounting in both directions:

- The pod's **real node is phantom-freed**: Kueue over-admits new multi-GPU
  pods onto it, which then sit `Pending` with `Insufficient nvidia.com/gpu` on
  a node that is actually full.
- The **replacement node is phantom-filled**: a healthy node's GPUs are
  reserved for pods that are not there and will never come. Critically, in
  window 2 the ReplicaSet creates replacement pods as soon as the
  deletionTimestamp is set (a Terminating pod is inactive to the RS) — and
  when the phantom-filled node holds the flavor's only free capacity, **the
  replacement pods cannot be admitted for the whole outage**, even though
  physical capacity exists.

A related side effect on capacity-tight flavors: when the replacement *search*
fails (no free domain), `TASFailedNodeReplacementFailFast` (Beta, default on)
**evicts the Running pod** instead. For a pod-owned Workload this can kill a
healthy pod early — the Deployment's replacement cannot be placed anywhere
either (the flavor is full and the failed node is excluded), so it waits
Pending and re-admits to the same node once it recovers, now with a cold start.

An admitted Workload owned by a single Pod should never have its
`topologyAssignment` rewritten once the pod is placed. Node replacement is
meaningful for Workloads whose controller recreates pods *within the same
Workload* (Job, JobSet, LeaderWorkerSet) — the mechanism the KEP describes
("the pods are re-created by the Job controller for placement on other
nodes"). A pod-owned Workload cannot outlive its pod and a Running pod cannot
move, so reassignment can never relocate anything; recovery for these
workloads only ever happens via the owning controller creating a new pod = a
new Workload with a fresh admission. The rewrite is therefore pure downside:
it cannot help the affected workload, and it corrupts per-node accounting for
every other workload in the flavor.

Hostname-level TAS ClusterQueue → admit Deployment-managed
pods (pod integration) onto node A → make node A NotReady for >30s but <5m,
with free capacity in the flavor → observe the assignments move to node B
while the pods keep running on A; the divergence never heals.

Proposed fix (validated on a fork; happy to open a PR): **skip the stage-2
replacement for Workloads owned by exactly one Pod, behind a new Alpha feature
gate** (e.g. `TASSkipReplacementForPlacedPod`). In
`TASFlavorSnapshot.FindTopologyAssignmentsForFlavor`, when the Workload has
unhealthy nodes and `len(ownerReferences)==1 && kind==Pod`, return the
existing `topologyAssignment` unchanged on the normal success path instead of
calling `findReplacementAssignment`. Everything else stays intact:
`status.unhealthyNodes` handling, the fail-fast path for other workload types,
and the admit flow (which clears `unhealthyNodes`, marking recovery complete
without moving the pod). ~30 lines + unit test. With the gate on, in our
kind+kwok harness: no drift at any outage length, no fail-fast eviction of
Running pods on saturated flavors, and in the long-outage case the
ReplicaSet's replacement pods admit into the genuinely-free capacity instead
of being blocked by the phantom-fill — while stock reproduces the drift 100%
deterministically in the same scenarios. Kueue's unit and integration suites
pass with the gate compiled in (default off).

Alternatives considered:

- **Making `NodeFailureDelay` configurable** and raising it toward the
  pod-eviction timeout (we validated this too — it suppresses window-1 drift
  when the delay exceeds the blip length). But it is a workaround, not a fix:
  it trades drift for slower legitimate replacement of everything else on the
  cluster (or requires a separate knob from the fail-fast delay to avoid
  that), it is guesswork against blip-length distributions, and it does
  nothing for window 2 (phantom-fill during long outages). Avoiding the
  reassignment for workloads that structurally cannot benefit from it
  addresses the cause rather than the timing.
- **`--feature-gates=TASFailedNodeReplacement=false`** disables replacement
  fleet-wide — acceptable only for clusters with exclusively pod-owned
  workloads, and loses hot-swap for any batch workloads sharing the cluster.

Related issues:
- #11858 — the capacity-stranding symptom of this drift; maintainers there propose 
  deleting running "late" pods behind a gate, which conforms the pods to the 
  rewritten assignment — complementary
  for multi-pod Workloads, but for pod-owned Workloads it deletes a healthy
  pod to justify a rewrite that was futile to begin with.
- #9615 — stage-1 handling of not-yet-running pods on failed nodes; its model
  ("the job controller can recreate these pods") is exactly what pod-owned
  Workloads lack.
- #12528 — the elastic scale-up variant of the same invariant (preserve a
  running leader pod's existing assignment).
- #10929 — Deployment/pod-integration double-accounting in TAS; background on
  pod-owned workloads being an under-exercised path in TAS.

**Environment**:
- Kubernetes version (use `kubectl version`): v1.32 (k3s) in production; kind
  v1.33 in the reproducer
- Kueue version (use `git describe --tags --dirty --always`): v0.17.2 (stock
  feature gates); replacement path inspected on `main` as well
- Cloud provider or hardware configuration: bare-metal GPU hosts (8×GPU
  NVLink) across multiple providers; kwok-simulated GPU nodes in the
  reproducer
- OS (e.g: `cat /etc/os-release`): Ubuntu 22.04 (production)
- Kernel (e.g. `uname -a`): n/a
- Install tools: Helm (production); release manifests (reproducer)
- Others: hostname-level Topology (`kubernetes.io/hostname` lowest level),
  Deployments via the pod integration with
  `kueue.x-k8s.io/pod-suspending-parent: deployment`

/kind bug
/area tas

yakticus:
#### What type of PR is this?

/kind feature
/area tas

A Workload owned by a single Pod (the pod integration: bare pods and e.g. Deployment replicas) is deleted together with its pod, so no future pod can ever consume a recomputed topology assignment. Kueue recomputes such assignments in two places today, and both make the stored assignment diverge from the node the still-running pod occupies, corrupting TAS per-node capacity accounting in both directions — the pod's real node is treated as free and over-admitted (new pods stick Pending with `Insufficient nvidia.com/gpu` on a full node), while the newly assigned node is blocked for admissions its capacity cannot actually serve:

1. **Failed-node replacement**: after `NodeFailureDelay` the node is marked unhealthy and the scheduler computes a replacement assignment, while the pod keeps running on the original node through a transient NotReady blip. On our bare-metal fleet, 1–3 minute kubelet/network blips with surviving pods are far more common than real node deaths.
2. **Post-eviction requeue**: an evicted (e.g. preempted) Workload is requeued and re-admitted with a freshly computed assignment within milliseconds, while its pod is still draining a long termination grace period (up to 15 minutes for our inference servers) on the original node.

With the `TASSkipReassignmentForPodOwnedWorkloads` gate enabled, for Workloads owned by exactly one Pod, Kueue instead:

1. keeps the existing topology assignment during node replacement — the node stays tracked in `.status.unhealthyNodes` and is cleared on the next scheduling pass;
2. sets `Requeued=False` on eviction, parking the Workload so it finishes together with its pod.

Recovery happens through the pod lifecycle: the owning controller (e.g. ReplicaSet) creates a replacement pod, which arrives as a new Workload admitted with a fresh assignment. Workloads with any other owner — Jobs, JobSets, and pod groups (including StatefulSet and LeaderWorkerSet, which can receive replacement pods into the same Workload) — keep the existing behavior.

We have been running this patch for several days in production on a sizeable GPU cluster: it eliminated both drift classes, validated through natural node flaps, a full multi-node site outage, and dozens of preemptions — zero placement drift, no regressions noted.

Fixes #12735

Part of #11858 (the single-Pod-owner case; the holistic multi-pod direction is under discussion on #12735)

- The check is scoped by **ownership**, not podset count: `workload.OwnedBySinglePod` requires exactly one ownerReference of `Kind=Pod, apiVersion=v1`. A size-1 Job must keep the replacement behavior because its controller re-creates pods within the same Workload; a pod-owned Workload can never receive another pod.
- The node-replacement skip returns the existing assignment as a Fit so admission clears `.status.unhealthyNodes` without moving the workload. The eviction-path skip parks the Workload via `Requeued=False` (the same parking mechanism used by `WorkloadDeactivated`).
- KEP-2724 and the TAS concepts page are updated in this PR, following the precedent of gate-introducing PRs #5861/#5931.
- Gate spec: Alpha (default off) at 0.18, Beta (default on) at 0.19, per @mimowo's guidance on #12735. Happy to adjust release targeting.
- Integration coverage: TAS suite (assignment kept for pod-owned vs replaced for Job-owned, gate on) and pod suite (pod-owned Workload evicted by preemption is parked `Requeued=False` gate-on / requeued `Requeued=True` gate-off).

```release-note
TAS & Scheduling: Fixed a bug where Workloads owned by a single Pod could be reassigned after eviction or during TAS node hot swap, even though the existing Pod could not consume the new assignment. The fix applies when the SkipReassignmentForPodOwnedWorkloads feature gate is enabled. The gate is Beta and enabled by default in 0.19+, and Alpha and disabled by default in the 0.17 and 0.18 release branches.
```

## Summary by CodeRabbit

* **New Features**
  * Added beta feature gate **SkipReassignmentForPodOwnedWorkloads** (default on in v0.19) to skip topology assignment recomputation for workloads owned by exactly one Pod.

* **Behavior Changes**
  * During failed-node handling, pod-owned workloads reuse their existing topology assignment while unhealthy node tracking remains until the next successful scheduling pass.
  * For eviction/preemption, pod-owned workloads are requeued with **Requeued=False** during pod termination grace.

* **Documentation**
  * Updated topology-aware scheduling hot-swap docs and design notes for the new gate.

* **Tests**
  * Added unit and integration coverage validating pod-owned vs job-owned behavior and requeue outcomes.

### DIFF
--- a/pkg/cache/scheduler/tas_cache_test.go
+++ b/pkg/cache/scheduler/tas_cache_test.go
@@ -25,11 +25,13 @@ import (
 	"github.com/google/go-cmp/cmp"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/component-base/featuregate"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	kueue "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+	podconstants "sigs.k8s.io/kueue/pkg/controller/jobs/pod/constants"
 	tasindexer "sigs.k8s.io/kueue/pkg/controller/tas/indexer"
 	"sigs.k8s.io/kueue/pkg/features"
 	"sigs.k8s.io/kueue/pkg/resources"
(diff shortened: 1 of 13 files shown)


pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -588,6 +588,19 @@ func (s *TASFlavorSnapshot) FindTopologyAssignmentsForFlavor(flavorTASRequests F
...
+				// diverging from the node the pod actually runs on.
+				for _, tr := range trs {
+					psa := findPSA(opts.workload, tr.PodSet.Name)
+					if psa == nil || psa.TopologyAssignment == nil {
+						continue
+					}
+					result[tr.PodSet.Name] = tasPodSetAssignmentResult{TopologyAssignment: utiltas.InternalFrom(psa.TopologyAssignment)}
+				}
+				continue
+			}
```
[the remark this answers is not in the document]
yakticus: A single-Pod-owned Workload has exactly one domain in its assignment — the unhealthy node itself — so after `deleteDomain()` there are no remaining domains for the stale check to validate. Applying `IsTopologyAssignmentStale` to the kept assignment instead would flag every transient NotReady (the snapshot excludes NotReady nodes) and fall back to replacement, reintroducing the divergence this gate prevents. On outright node deletion the pod GC deletes the pod and the Workload with it, and a missing domain holds no snapshot capacity, so nothing is blocked.

yakticus:
The unit-shard failure was a real catch: a pod group of size 1 also has exactly one Pod ownerReference, so it satisfied the ownership check despite being able to receive replacement pods. Fixed by excluding Workloads carrying the `kueue.x-k8s.io/is-group-workload` annotation (stamped unconditionally by `NewGroupWorkload`, and already used as the group discriminator in the pod integration's event handlers). The failing test's fixture now sets the annotation, consistent with the other group fixtures in that file.

pkg/cache/scheduler/tas_cache_test.go
```diff
@@ -7960,3 +7961,115 @@ func aggregatedDomainUsagesForPriorFlavorUsage(
...
+			want := &tas.TopologyAssignment{
+				Levels:  levels,
+				Domains: []tas.TopologyDomainAssignment{{Count: 1, Values: []string{tc.wantHost}}},
+			}
+			if diff := cmp.Diff(want, psResult.TopologyAssignment); diff != "" {
+				t.Errorf("unexpected topology assignment (-want,+got):\n%s", diff)
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
yakticus: Added: "gate on, size-1 pod group (is-group-workload annotation): replaced" now covers exactly the regression, alongside a rename of the test to match the gate.

pkg/features/kube_features.go
```diff
@@ -565,6 +574,10 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 		{Version: version.MustParse("0.13"), Default: false, PreRelease: featuregate.Alpha},
 		{Version: version.MustParse("0.14"), Default: true, PreRelease: featuregate.Beta},
 	},
+	TASSkipReassignmentForPodOwnedWorkloads: {
+		{Version: version.MustParse("0.18"), Default: false, PreRelease: featuregate.Alpha},
```
mimowo: Let's drop this line. We will adjust on backport

test/integration/singlecluster/tas/tas_test.go
```diff
@@ -1589,6 +1589,108 @@ var _ = ginkgo.Describe("Topology Aware Scheduling", ginkgo.Ordered, func() {
 					}, util.Timeout, util.Interval).Should(gomega.Succeed())
 				})
 			})
+
+			ginkgo.It("should keep the TopologyAssignment of a single-Pod-owned workload when node fails and TASSkipReassignmentForPodOwnedWorkloads is enabled", framework.SlowSpec, func() {
```
mimowo: Since we want to decommission the "NotReady 30s" I would like to cover this scenario using TASReplaceNodeOnPodTermination as well

pkg/controller/jobframework/reconciler.go
```diff
@@ -559,6 +559,11 @@ func (r *JobReconciler) ReconcileGenericJob(ctx context.Context, req ctrl.Reques
 				err := workloadpatching.PatchAdmissionStatus(ctx, r.client, wl, r.clock, func(wl *kueue.Workload) (bool, error) {
 					// The requeued condition status set to true only on EvictedByPreemption
 					setRequeued := (evCond.Reason == kueue.WorkloadEvictedByPreemption) || (evCond.Reason == kueue.WorkloadEvictedDueToNodeFailures)
+					// A pod-owned Workload dies with its pod; requeuing it would
+					// recompute an assignment nothing can consume (placement drift).
+					if features.Enabled(features.TASSkipReassignmentForPodOwnedWorkloads) && workload.OwnedBySinglePod(wl) {
```
mimowo: Hm, actually this seems to be also a bugfix outside of TAS. A deployment-owned pod has workload and without this fix it may get re-admitted immediately consuming a little bit of quota for a short while. Would it be reproducible in an integration test? Makes we wonder if we should split the PR into two, or generalize the FG name as `SkipReadmissionForPodOwnedWorkloads`, because TASNotHotSwap is a form of re-admisson.  wdyt?

pkg/features/kube_features.go
yakticus: Done — spec is `{0.19 Beta/true}` only now.

test/integration/singlecluster/tas/tas_test.go
yakticus: Added: "should keep the TopologyAssignment of a single-Pod-owned workload when the node is marked via TASReplaceNodeOnPodTermination" — node goes NotReady with no NodeFailureDelay wait, marking happens via the pod-termination path, and the assignment is kept.

pkg/controller/jobframework/reconciler.go
yakticus: Agreed it's not TAS-specific — the requeue path fires for any pod-owned Workload. In fact the integration specs in this PR already cover the non-TAS case: the pod-suite specs (parked on preemption / NodeFailures, requeued with the gate off) run without any topology objects. I'd rather keep it as one PR than split during release week, and generalize the name as you suggest — though I'd propose `SkipReassignmentForPodOwnedWorkloads` rather than `SkipReadmission…`, since the node-failure path rewrites the assignment in place without un-admitting, so "reassignment" describes both doors. Happy to go with your preference either way.

mimowo: SkipReassignmentForPodOwnedWorkloads sgtm

yakticus: Renamed throughout: `SkipReassignmentForPodOwnedWorkloads`.

mimowo:
/retitle Fix handling of single-pod owned workloads; add SkipReassignmentForPodOwnedWorkloads FG

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
```diff
@@ -427,6 +449,105 @@ var _ = ginkgo.Describe("Pod controller", ginkgo.Label("job:pod", "area:jobs"),
...
+						g.Expect(createdWorkload.Status.Conditions).Should(gomega.ContainElement(gomega.BeComparableTo(
+							metav1.Condition{
+								Type:   kueue.WorkloadRequeued,
+								Status: metav1.ConditionTrue,
+								Reason: kueue.WorkloadEvictedByPreemption,
+							},
+							wlConditionCmpOpts...,
+						)))
+					}, util.Timeout, util.Interval).Should(gomega.Succeed())
+				})
```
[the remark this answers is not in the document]
mimowo: @yakticus actually this is a good point, wlConditionCmpOpts ignores Reason, so declaring it in asserts is misleading. Let's follow with the `util.IgnoreConditionTimestampsAndObservedGeneration` suggestion.

site/data/featuregates/versioned_feature_list.yaml
```diff
@@ -389,6 +389,12 @@
     lockToDefault: true
     preRelease: GA
     version: "0.18"
+- name: SkipReassignmentForPodOwnedWorkloads
+  versionedSpecs:
+  - default: true
+    lockToDefault: false
+    preRelease: Beta
+    version: "0.19"
```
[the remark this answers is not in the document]
mimowo: We will adjust the version to 0.18 / 0.17 on the cherrypick PRs where we also adjust it to Alpha.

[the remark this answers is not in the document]
yakticus: The 0.18 entry was removed at the maintainer's request (https://github.com/kubernetes-sigs/kueue/pull/12980#discussion_r3573481246) — backport branches carry their own lifecycle entries.

test/integration/singlecluster/controller/jobs/pod/pod_controller_test.go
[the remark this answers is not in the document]
yakticus: Switched the four Requeued assertions to `util.IgnoreConditionTimestampsAndObservedGeneration` and added the expected Message — Reason and Message are now exercised.

pkg/cache/scheduler/tas_cache_test.go
```diff
@@ -7960,3 +7962,125 @@ func aggregatedDomainUsagesForPriorFlavorUsage(
 	}
 	return aggregatedDomainUsages
 }
+
+func TestFindTopologyAssignmentsSkipReassignmentForPodOwnedWorkloads(t *testing.T) {
```
mimowo: nit, but it feels this unit test could be relatively easily swiped under the pre-existing test which exercises many scenarios, including hot swap. I'm thinking this way to avoid duplicating the setup.

I know this means duplicating the params, but these are largely declarative rather than imperative, while the setup is imperative.

Let me know if there are some other considerations.

yakticus: Done — folded into `TestFindTopologyAssignments` as five declarative cases; the runner gained an optional per-case `workload` passed via `WithWorkload` (it previously never exercised the `UnhealthyNodes` path). Standalone test removed. Note the stricter-matcher change you flagged on the pod-suite assertions was already in the previous push.

pkg/cache/scheduler/tas_flavor_snapshot.go
```diff
@@ -588,6 +588,19 @@ func (s *TASFlavorSnapshot) FindTopologyAssignmentsForFlavor(flavorTASRequests F
...
```
mimowo: this loop
```golang
	for _, tr := range trs {
		psa := findPSA(opts.workload, tr.PodSet.Name)
		if psa == nil || psa.TopologyAssignment == nil {
			continue
		}
``` 
looks identical as below, maybe instead of dedicated handling we could just inject the check below, ie:

```golang
for _, tr := range trs {
	// In case of looking for Node replacement, TopologyRequest has only
	// PodSets with the Node to replace, so we match PodSetAssignment
	psa := findPSA(opts.workload, tr.PodSet.Name)
	if psa == nil || psa.TopologyAssignment == nil {
		continue
	}
   if features.Enabled(features.SkipReassignmentForPodOwnedWorkloads) && workload.OwnedBySinglePod(opts.workload) {
	// The pod cannot relocate and the Workload cannot outlive it; keep
	// the existing assignment.
     result[tr.PodSet.Name] = tasPodSetAssignmentResult{TopologyAssignment: utiltas.InternalFrom(psa.TopologyAssignment)}
    continue
   }
```
We could also consider moving down the code under `findReplacementAssignment` to further commonize the handling.

yakticus: Done — merged into the replacement loop as suggested. I left it at the call site rather than inside `findReplacementAssignment`: the skip selects *whether* to compute a replacement at all, so keeping it where the mode is chosen seemed clearer than having `findReplacementAssignment` sometimes return the unmodified input. Happy to move it if you prefer.

mimowo: I'm ok as is 👍

mimowo:
/lgtm
/approve
Thank you for the bugfix 👍 

Please prepare the cherrypicks adjusting the version "from" to correspond to the release branch. Plus, turn them into "Alpha" .

mimowo:
/release-note-edit
```release-note
TAS & Scheduling: Fixed a bug where Workloads owned by a single Pod could be reassigned after eviction or during 
TAS node hot swap, even though the terminating Pod could not consume the new assignment. The fix is guarded by 
the `SkipReassignmentForPodOwnedWorkloads` feature gate, which is Beta and enabled by default in 0.19+, and Alpha 
and disabled by default in the 0.17 and 0.18 release branches.
```

mimowo:
/release-note-edit
```release-note
TAS & Scheduling: Fixed a bug where Workloads owned by a single Pod could be reassigned after eviction or during TAS node hot swap, even though the existing Pod could not consume the new assignment. The fix applies when the SkipReassignmentForPodOwnedWorkloads feature gate is enabled. The gate is Beta and enabled by default in 0.19+, and Alpha and disabled by default in the 0.17 and 0.18 release branches.
```

Issue #11726, opened 2026-05-28
Support DRA Consumable Capacity (KEP-5075) in Kueue


sohankunkerkar:
**What would you like to be added**:
When a DRA driver publishes devices with device.Capacity fields ([KEP-5075](https://github.com/kubernetes/enhancements/issues/5075)), the kube-scheduler tracks capacity consumption to prevent over-allocation of shared resources. Kueue needs capacity-based quota management for these devices. For example, a GPU that supports time-slicing or MPS can serve multiple workloads concurrently, each consuming a fraction of the device's compute and memory capacity. Device-count quota cannot express as one device counted as "1 GPU" might serve 4 workloads at 25% each, or 2 workloads at 50% each. Capacity-based quota charges workloads based on the actual capacity they consume from the device.

**Why is this needed**:
Fractional GPU sharing via time-slicing and MPS is widely used to improve GPU utilization for inference and small training workloads. Without capacity-based quota, admins cannot distinguish between a workload using 25% of a GPU and one using 100%. Both consume "1 GPU" in device-count quota, leading to either over-admission (cluster runs out of capacity) or under-utilization (quota blocks workloads that would fit). This is the complement to Partitionable Devices (#10108): PD handles static partitioning (MIG), while Consumable Capacity handles dynamic sharing (time-slicing, MPS).

**Completion requirements**:

This enhancement requires the following artifacts:

- [x] Design doc: https://github.com/kubernetes-sigs/kueue/pull/12983
- [x] API change: https://github.com/kubernetes-sigs/kueue/pull/13152
- [x] Docs update: https://github.com/kubernetes-sigs/kueue/pull/13309

The artifacts should be linked in subsequent comments.

PR #12983, opened 2026-07-11
KEP-2941: add Consumable Capacity design

sohankunkerkar:
#### What type of PR is this?
/kind kep

related to #11726 

The content was produced with the help of AI tools, but it was reviewed and manually curated by me.

```release-note
NONE
```

## Summary

- **New Features**
  - Added documentation for **DRA Consumable Capacity** as an additional Kueue quota approach.
  - Included the **`KueueDRAIntegrationConsumableCapacity`** feature gate in the KEP guidance.

- **Documentation**
  - Introduced a unified device-class resource source model supporting both **counter** and **capacity** quota sources.
  - Updated configuration validation, including per-source feature gating, required fields, `deviceSelector` validation, and source consistency rules.
  - Expanded the consumable-capacity design, including charging behavior and supported request interactions.

- **Tests**
  - Updated test plan and graduation criteria with consumable-capacity-specific scenarios.

### DIFF
--- a/keps/2941-DRA/README.md
+++ b/keps/2941-DRA/README.md
@@ -63,6 +63,14 @@ tags, and then generate with `hack/update-toc.sh`.
     - [Path Interactions](#path-interactions)
     - [Counter Lifecycle Scenarios](#counter-lifecycle-scenarios)
     - [Validation](#validation)
+  - [Consumable Capacity](#consumable-capacity)
+    - [ResourceSlice Structure](#resourceslice-structure-1)
+    - [User Workload](#user-workload-1)
+    - [Configuration](#configuration-2)
+    - [Processing Flow](#processing-flow-2)
+    - [Path Interactions](#path-interactions-1)
+    - [Capacity Lifecycle Scenarios](#capacity-lifecycle-scenarios)
+    - [Validation](#validation-1)
   - [Architecture Details](#architecture-details)
     - [Queue Manager Extensions](#queue-manager-extensions)
   - [Integration with Admission Fair Sharing](#integration-with-admission-fair-sharing)
@@ -78,10 +86,12 @@ tags, and then generate with `hack/update-toc.sh`.
       - [KueueDRAIntegrationExtendedResource (v0.17)](#kueuedraintegrationextendedresource-v017)
       - [KueueDRAIntegrationExtendedResource (v0.18)](#kueuedraintegrationextendedresource-v018)
       - [KueueDRAIntegrationPartitionableDevices (v0.18)](#kueuedraintegrationpartitionabledevices-v018)
+      - [KueueDRAIntegrationConsumableCapacity (v0.19)](#kueuedraintegrationconsumablecapacity-v019)
     - [Beta](#beta)
       - [KueueDRAIntegration (v0.18)](#kueuedraintegration-v018)
       - [KueueDRAIntegrationExtendedResource](#kueuedraintegrationextendedresource)
       - [KueueDRAIntegrationPartitionableDevices](#kueuedraintegrationpartitionabledevices)
+      - [KueueDRAIntegrationConsumableCapacity](#kueuedraintegrationconsumablecapacity)
     - [GA](#ga)
       - [KueueDRAIntegration](#kueuedraintegration)
       - [KueueDRAIntegrationExtendedResource](#kueuedraintegrationextendedresource-1)
@@ -101,7 +111,8 @@ tags, and then generate with `hack/update-toc.sh`.
   - [Device-Count Quota with Dual Tracking](#device-count-quota-with-dual-tracking)
   - [Auto-discovery of Counters Without Configuration](#auto-discovery-of-counters-without-configuration)
 - [Appendix](#appendix)
-  - [Consumable Capacity Compatibility](#consumable-capacity-compatibility)
+  - [KEP-5941 Shared Consumable Capacity](#kep-5941-shared-consumable-capacity)
+  - [KEP-5963 Device Compatibility Groups](#kep-5963-device-compatibility-groups)
 <!-- /toc -->
 
 ## Summary
@@ -109,13 +120,17 @@ tags, and then generate with `hack/update-toc.sh`.
 [Dynamic Resource Allocation (DRA)](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/)
 is a major effort to improve device support in Kubernetes. It changes how one can request resources in a myriad of ways.
 
-This KEP supports three approaches for DRA integration with Kueue:
+This KEP supports four approaches for DRA integration with Kueue:
 1. **ResourceClaimTemplates**: Pods explicitly reference ResourceClaimTemplates that specify device requests.
 2. **Extended Resources**: Pods request DRA devices via standard `resources.requests` (e.g., `example.com/gpu: 1`), and kube-scheduler automatically creates ResourceClaims when the DeviceClass has an `extendedResourceName` field set.
 3. **Partitionable Devices**: Counter-based quota for devices that can be dynamically
    partitioned (e.g., NVIDIA MIG). Instead of counting devices, Kueue tracks counter
    consumption (e.g., GPU memory) from the `SharedCounters` and `ConsumesCounters` fields
    defined by [KEP-4815](https://github.com/kubernetes/enhancements/issues/4815).
+4. **Consumable Capacity**: Capacity-based quota for devices that allow software-level
+   sharing. Kueue tracks consumed capacity dimensions such as GPU memory and compute
+   cores from the device's `Capacity` field as defined by
+   [KEP-5075](https://github.com/kubernetes/enhancements/issues/5075).
 
 ## Motivation
 
@@ -217,6 +232,8 @@ a simple device class named `gpu.example.com`. This will be the way to enforce q
 - Admins can enforce the quota for number of devices for a given DeviceClass.
 - Admins can enforce counter-based quota for partitionable devices (e.g., GPU memory quota
   instead of device count quota for MIG profiles).
+- Admins can enforce capacity-based quota for devices that allow software-level sharing
+  (e.g., GPU memory and compute cores quota for time-sliced or fractional GPU devices).
 
 ### Non-Goals
 
@@ -226,11 +243,6 @@ a simple device class named `gpu.example.com`. This will be the way to enforce q
 - Support for DRA features like DRADeviceTaints is not included.
 - Multi-host partitionable devices (e.g., NVLink fabrics spanning multiple nodes) are not
   supported.
-- Support for DRAConsumableCapacity (alpha in K8s 1.34, beta targeting K8s 1.36) is not
-  included. Consumable capacity enables software-level device sharing (e.g., MPS, time-slicing)
-  where multiple claims share one device. This requires different quota semantics as Kueue
-  would need to read the cost from the claim's capacity request rather than the device's
-  counters. No GPU driver ships consumable capacity yet.
 - This design does not work with Topology Aware Scheduling feature of Kueue. It is a significant
   amount of work, will be addressed in the future with a separate body of work.
 
@@ -251,6 +263,10 @@ scheduling. This includes:
    partitionable devices (requires Kubernetes `DRAPartitionableDevices` feature gate, beta in
    K8s 1.36). This builds on CEL expression support which adds ResourceSlice access and
    device matching to Kueue.
+7. Supporting capacity-based quota for devices that allow multiple allocations via the
+   `capacity` source type on `deviceClassMappings` (requires Kubernetes
+   `DRAConsumableCapacity` feature gate, beta in K8s 1.36). Kueue charges the workload's
+   `capacity.requests` rounded per the device's `RequestPolicy`.
 
 More details are documented in [Design Details](#design-details)
 
@@ -374,24 +390,36 @@ is documented here:
    - Post-scheduling quota reconciliation will be evaluated for Beta.
    - TAS + DRA is the longer-term path to closing this admission-scheduling gap.
 
+**Consumable capacity under-charge on exclusive devices**: if the `deviceSelector` matches
+devices without `AllowMultipleAllocations`, the scheduler consumes the entire device while
+Kueue charges only the partial `capacity.requests` amount. This is specific to capacity
+sources because the charge comes from the workload's request, not the device (counter
+sources charge from the device's `consumesCounters` and do not have this issue).
+Mitigation: include `device.allowMultipleAllocations == true` in the `deviceSelector`.
+The general admission-scheduling timing gap applies equally to all DRA source types.
+`waitForPodsReady` catches scheduling failures and evicts the workload so quota is
+released.
+
 ## Design Details
 
-Three feature gates control DRA support in Kueue:
-- `DynamicResourceAllocation`: gates ResourceClaimTemplate-based DRA quota accounting.
-  Uses `deviceClassMappings` for DeviceClass-to-quota-resource mapping.
-- `KueueDRAIntegrationExtendedResource`: gates extended resources support, including DeviceClass
-  auto-discovery via `extendedResourceName`. Does not use `deviceClassMappings`.
-  Requires `DynamicResourceAllocation` to also be enabled.
-- `KueueDRAIntegrationPartitionableDevices`: gates counter-based quota for partitionable
-  devices. Enables the `sources` field on `deviceClassMappings` entries. Requires
-  `DynamicResourceAllocation` to also be enabled. Also requires the Kubernetes
-  `DRAPartitionableDevices` feature gate (beta in K8s 1.36).
-- `KueueDRARejectWorkloadsWhenDRADisabled` (default: enabled, Beta since v0.18): rejects workloads that
-  use DRA resources (ResourceClaimTemplates or ResourceClaims) when the
-  `DynamicResourceAllocation` feature gate is disabled. Without this gate, DRA workloads
-  submitted while `DynamicResourceAllocation` is off are silently admitted with zero
-  device resource usage, bypassing quota enforcement entirely. See
-  [Workload Rejection When DRA Is Disabled](#workload-rejection-when-dra-is-disabled).
+Feature gates controlling DRA support in Kueue:
+- `KueueDRAIntegration` (Beta, default on since v0.18): gates ResourceClaimTemplate-based
+  DRA quota accounting. Uses `deviceClassMappings` for DeviceClass-to-quota-resource mapping.
+- `KueueDRAIntegrationExtendedResource` (Alpha): gates extended resources support, including
+  DeviceClass auto-discovery via `extendedResourceName`. Requires `KueueDRAIntegration`.
+- `KueueDRAIntegrationPartitionableDevices` (Alpha): gates counter-based quota for
+  partitionable devices. Enables the `counter` source type on `deviceClassMappings` entries.
+  Requires `KueueDRAIntegration`. Also requires the Kubernetes `DRAPartitionableDevices`
+  feature gate (beta in K8s 1.36).
+- `KueueDRAIntegrationConsumableCapacity` (Alpha): gates capacity-based quota for devices
+  that allow multiple allocations. Enables the `capacity` source type on
+  `deviceClassMappings` entries. Requires `KueueDRAIntegration`. Also requires the
+  Kubernetes `DRAConsumableCapacity` feature gate (beta in K8s 1.36).
+- `KueueDRARejectWorkloadsWhenDRADisabled` (Beta, default on since v0.18): rejects workloads
+  that use DRA resources (ResourceClaimTemplates or ResourceClaims) when `KueueDRAIntegration`
+  is disabled. Without this gate, DRA workloads submitted while `KueueDRAIntegration` is off
+  are silently admitted with zero device resource usage, bypassing quota enforcement entirely.
+  See [Workload Rejection When DRA Is Disabled](#workload-rejection-when-dra-is-disabled).
 
 The following sections will explain the design in detail.
 
@@ -542,7 +570,13 @@ data:
         - fast-gpus.example.com      # Different device class
 ```
 
-This validation approach eliminates ambiguity at configuration time rather than requiring complex runtime resolution logic, ensuring predictable and efficient workload admission.
+This validation approach eliminates ambiguity at configuration time rather than requiring
+complex runtime resolution logic, ensuring predictable and efficient workload admission.
+
+**Note**: A single mapping can have multiple capacity sources that sum into one quota
+resource, which does not violate this constraint. Tracking independent capacity
+dimensions as separate quota resources (same DeviceClass, different resource names)
+requires relaxing this uniqueness constraint and is deferred to beta.
 
 ### RBAC Requirements
 
@@ -1089,10 +1123,10 @@ type DeviceClassMapping struct {
 
     // Sources configures resource accounting sources for this mapping.
     // Each source defines how quota is tracked for this DeviceClass.
-    // Currently only counter sources are supported (for partitionable devices).
+    // Counter sources require KueueDRAIntegrationPartitionableDevices.
+    // Capacity sources require KueueDRAIntegrationConsumableCapacity.
     // Extended resource requests that resolve to a DeviceClass with sources
     // configured are marked inadmissible.
-    // Requires the KueueDRAIntegrationPartitionableDevices feature gate.
     // +optional
     Sources []DeviceClassSourceConfig `json:"sources,omitempty"`
 }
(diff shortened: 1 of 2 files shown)


kannon92 (commented):
/assign @alaypatel07 

Do we know if NVIDIA will have consumable capacity implemented?

If not I don't know if we need to push this for 0.19. Idk of any other drivers that would have this yet either.

sohankunkerkar:
> /assign @alaypatel07
> 
> Do we know if NVIDIA will have consumable capacity implemented?

Not yet, the work is in progress!

> 
> If not I don't know if we need to push this for 0.19. Idk of any other drivers that would have this yet either.

I don't think we need to wait for the NVIDIA to implement things. The feature itself is vendor agnostic and we can use the latest released [version](https://github.com/kubernetes-sigs/dra-example-driver/releases/tag/v0.4.0) of dra-example-driver to test this feature.

kannon92:
> I don't think we need to wait for the NVIDIA to implement things. The feature itself is vendor agnostic and we can use the latest released [version](https://github.com/kubernetes-sigs/dra-example-driver/releases/tag/v0.4.0) of dra-example-driver to test this feature.

It is but in reality we would get no feedback until vendors actually implement this feature.

sohankunkerkar:
> > I don't think we need to wait for the NVIDIA to implement things. The feature itself is vendor agnostic and we can use the latest released [version](https://github.com/kubernetes-sigs/dra-example-driver/releases/tag/v0.4.0) of dra-example-driver to test this feature.
> 
> It is but in reality we would get no feedback until vendors actually implement this feature.

This is alpha. We ship it, get it in front of early adopters, and wait for real vendor feedback before graduating to beta. That's the standard alpha contract.

kannon92:
> This is alpha. We ship it, get it in front of early adopters, and wait for real vendor feedback before graduating to beta. That's the standard alpha contract.

Do we have in mind how early adopters will test and verify this if we don't have device vendors who support this? Is your idea that they can test/verify with dra-example-driver?

sohankunkerkar:
> > This is alpha. We ship it, get it in front of early adopters, and wait for real vendor feedback before graduating to beta. That's the standard alpha contract.
> 
> Do we have in mind how early adopters will test and verify this if we don't have device vendors who support this? Is your idea that they can test/verify with dra-example-driver?

The idea is that we don't wait for vendors to implement new APIs before shipping support on the Kueue side. We always use the `dra-example-driver` as the baseline for testing DRA features, and this is no different. The point of shipping alpha support now is that once vendors adopt the Consumable Capacity API, users won't have to wait for Kueue to catch up. The quota accounting will already be implemented and tested.

kannon92:
> The idea is that we don't wait for vendors to implement new APIs before shipping support on the Kueue side. We always use the dra-example-driver as the baseline for testing DRA features, and this is no different.

Every feature we added support for already had corresponding drivers backing it (Dynamic MIG, Structured Parameters). We were able to test/confirm the  real driver with the Kueue integration.

If we know the timeline of Consumable Capacity for NVIDIA or at least have another driver in mind, I'm fine to merge this.

When I check [NVIDIA repo](https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/issues/850), I don't really see any active work on consumable capacity so I hope we are putting the cart before the horse on the Kueue integration.

sohankunkerkar:
>When I check https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/issues/850, I don't really see any active work on consumable capacity so I hope we are putting the cart before the horse on the Kueue integration.

https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/pull/1231

keps/2941-DRA/README.md
```diff
@@ -1103,27 +1132,30 @@ type DeviceClassSourceConfig struct {
     // Counter configures counter-based quota for partitionable devices.
     // Maps a DRA driver counter to the parent DeviceClassMapping's Kueue quota resource.
     // +optional
-    Counter *DeviceClassCounterSource `json:"counter,omitempty"`
+    Counter *DeviceClassResourceSource `json:"counter,omitempty"`
```
tenzen-y: Why do we need to consolidate `DeviceClassResourceSource` and `DeviceClassCounterSource` into `DeviceClassResourceSource`?
In practice, the API types should be duplicated rather than a compact and reusable one for future extensions. I can understand that this consolidation allows us to compress some implementation costs, but we should prioritize API stability in this case.

- https://github.com/kubernetes-sigs/kueue/pull/10283#discussion_r3289384329
- https://github.com/kubernetes-sigs/kueue/pull/10283#discussion_r3283112646

keps/2941-DRA/README.md
```diff
@@ -1355,6 +1389,273 @@ borrowing only operates within the same resource.
...
+This section is gated behind the `KueueDRAIntegrationConsumableCapacity` Kueue feature gate.
+
+Kueue supports capacity-based quota for DRA devices that allow software-level sharing as
+defined by Kubernetes [KEP-5075](https://github.com/kubernetes/enhancements/issues/5075).
+Kueue tracks consumed capacity dimensions such as GPU memory and compute cores from the
+device's `Capacity` field and the workload's `capacity.requests` on `ExactDeviceRequest`.
+
+Only `ExactDeviceRequest` with `count` is supported. `FirstAvailable` subrequests with
+`capacity.requests` are not supported, consistent with the existing exclusion for
+partitionable devices.
```
tenzen-y, line 1432: @sohankunkerkar Do you want to revisit `FirstAvailable` support in the next Beta stage?
If yes, can you add it to graduation criteria? If not, can you add future work section?

keps/2941-DRA/README.md
```diff
@@ -1355,6 +1389,273 @@ borrowing only operates within the same resource.
...
+          selectors:
+          - cel:
+              expression: "device.attributes['gpu.example.com'].model == 'A100-80GB'"
+          capacity:
+            requests:
+              gpu.example.com/memory: "4Gi"
+              gpu.example.com/cores: "30"
+```
+
+If `capacity.requests` is omitted, the scheduler applies the `RequestPolicy.Default`
```
```suggestion
If `capacity.requests` is omitted, the kube-scheduler applies the `RequestPolicy.Default`
```

I'd like to emphasize that this `scheduler` indicates `kube-scheduler` for clarifications.

keps/2941-DRA/README.md
```diff
@@ -1355,6 +1389,273 @@ borrowing only operates within the same resource.
 - `KueueDRAIntegrationPartitionableDevices` requires `KueueDRAIntegration` to be enabled. Validated
   at startup in `pkg/config/validation.go`.
 
+### Consumable Capacity
```
tenzen-y, line 1421: What if the consumableCapacity compatibilityGroups restrictions and the deviceSelector CEL expressions restrictions conflict?

tenzen-y (commented):
I left some initial my comments.
Thank you 👍

keps/2941-DRA/README.md
```diff
@@ -1355,6 +1389,273 @@ borrowing only operates within the same resource.
...
+type DeviceClassSourceConfig struct {
+    // ...existing field (Counter)...
+
+    // Capacity configures capacity-based quota for devices that allow
+    // multiple allocations (consumable capacity). Uses the same
+    // DeviceClassResourceSource type defined above in the counter section.
+    // +optional
+    Capacity *DeviceClassResourceSource `json:"capacity,omitempty"`
+}
+```
```
tenzen-y: https://github.com/kubernetes-sigs/kueue/pull/12983#discussion_r3581613858

keps/2941-DRA/README.md
sohankunkerkar: I checked the upstream code and both `device.Capacity` and `counter` names use the same key type (`QualifiedName`). Our config just points at a dimension in the same Device struct with the same three fields. The only difference is which map we read from in code, not in config shape. Happy to split if you see a concrete case where they'd need different fields though.

keps/2941-DRA/README.md
sohankunkerkar: compatibilityGroups (KEP-5963) lives on consumesCounters entries and it's a partitionable devices concern. CC devices use `Device.Capacity` without consumesCounters, so there's no conflict path. This is documented in the appendix as well.

keps/2941-DRA/README.md
sohankunkerkar: Its already in beta graduation criteria (line 1907). The blocker is that Kueue can't know which alternative the scheduler picks without running it. scheduler-library integration should close this gap.

keps/2941-DRA/README.md
mimowo: @tenzen-y I think consolidating sgtm is we don't have any divergance atm, and we don't foresee that need.

Sure, we may not foresee all API evolutions, but if we need to diverage, then we can IMO introduce another API type. That change will not be breaking from the schema perspective (type names are not imprinted in the schema), only for the users of goclients-, but I think this is acceptable.

We took the analogous approach for ResourceGroup type reuse for Cohorts - taken from CQs. 1

mimowo: However, I want to check with @sohankunkerkar if this claim is factually correct: "I checked the upstream code and both device.Capacity and counter names use the same key type (QualifiedName). ", because I can see different structures as values: https://github.com/kubernetes/kubernetes/blob/9fc098b4ab93c9ba6ae752dd715b958e8af7b643/staging/src/k8s.io/api/resource/v1/types.go#L330-L339

wdyt?

sohankunkerkar: You're right that the upstream value types differ (Counter has only Value, DeviceCapacity adds RequestPolicy). The key types also use different naming conventions i.e. counter keys are DNS labels like memory, capacity keys are qualified like gpu.example.com/memory. But QualifiedName is type QualifiedName string, so from Go's perspective both are plain strings. Our config Name field is used as a map key in both paths and the cast to QualifiedName is a no-op. The value type difference is handled in the processing code, not the config.

mimowo: So maybe we introduce a pair of aliases *DeviceClassCounterSource and *DeviceClassapacitySource  to the currently common type. This way we avoid code duplication now, but we already make it future proof in case we decide to diverge one day. Keeping the alias definition seems like minimal cost for the benefits


```golang
type DeviceClassResourceSource struct {
    Name           string
    Driver         string
    DeviceSelector resourcev1.DeviceSelector
}

type DeviceClassCounterSource DeviceClassResourceSource
type DeviceClassCapacitySource DeviceClassResourceSource
```

sohankunkerkar: That works well actually. `type` aliases give us separate names without code duplication, and we can promote to full types later if they diverge.
@tenzen-y thoughts?

alaypatel07:
Planning to review this PR this week, apologies for the delay

keps/2941-DRA/README.md
mimowo: I asked LLM if type renames happened in the core k8s and it provided me some examples, even for stable APIs:
- [IPFamilyPolicyType => IPFamilyPolicy
](https://github.com/kubernetes/kubernetes/pull/110990)
- [ServiceExternalTrafficPolicyTypeCluster -> ServiceExternalTrafficPolicyCluster](https://github.com/kubernetes/kubernetes/pull/110991)
- [PodDisruptionsAllowed -> DisruptionsAllowed](https://github.com/kubernetes/kubernetes/pull/85863) (v1beta1)

So such types of breaks for go-clients isn't strictly enforced AFAIK, sure better to avoid.

So, the sensible options I see:
1. DeviceClassResourceSource with the risk of go-client breaking changes (current code)
2. aliases to provide future-compatibility, but commonize using a struct
3. separate the types day one

We did due diligence and we cannot see a use-case to separate. If the use-case was coming in the next release I would be leaning for (3.). 

So, I'm leaning towards option (2.) as a sensible middle-ground, also as most readable (concise code, aliases reflect the intention nicely). 

@sohankunkerkar maybe let's adjust the code tentatively assuming @tenzen-y is ok with that. Feel free to adjust if @tenzen-y prefers (3.). With already introduced aliases it will be still less work than now .

tenzen-y: I'd still like to take (3) in this version. The primary reason is https://github.com/kubernetes-sigs/kueue/pull/12983#discussion_r3581613858, but TBH, I have another hidden reason based on the previous failures.

Previously, we defined a commonized API type in the case of AFS (very similar commonization as this DRA KEP), then we accidentally added the field to the commonized type even though the new field can be used by only one parent field, which means other parent fields can not use the additional field.

I don't think that we should iterate such failures. We should protect the API with a mechanism like a separate type.

So, I would go with (3).

keps/2941-DRA/README.md
```diff
@@ -374,24 +390,35 @@ is documented here:
...
    - TAS + DRA is the longer-term path to closing this admission-scheduling gap.
 
+**Consumable capacity under-charge on exclusive devices**: if the `deviceSelector` matches
+devices without `AllowMultipleAllocations`, the scheduler consumes the entire device while
+Kueue charges only the partial `capacity.requests` amount. This is specific to capacity
+sources because the charge comes from the workload's request, not the device (counter
+sources charge from the device's `consumesCounters` and do not have this issue).
+Mitigation: include `device.allowMultipleAllocations == true` in the `deviceSelector`.
+The general admission-scheduling timing gap applies equally to all DRA source types and
+is addressed long-term by the scheduler-library integration.
```
tenzen-y: I think that the scheduler library cannot resolve this problem. Because the `device.allowMultipleAllocations == true` could be defined in Kueue Configuration API. But, kube-scheduler can not read Kueue Configuratoin.

Or, do you mean that we enforce to specify the same device selector in DeviceClass or ResourceClaimTemplate? Especially, `device.allowMultipleAllocations == true`?

keps/2941-DRA/README.md
```diff
@@ -1355,6 +1389,281 @@ borrowing only operates within the same resource.
...
+     `RequestPolicy.Default` if set, otherwise the maximum `Capacity.Value` across
+     matched devices
+   - `RequestPolicy` is read from the device with the maximum `Capacity.Value`.
+     This assumes homogeneous policies across matched devices within a
+     `deviceSelector` scope. The `deviceSelector` should narrow to a specific
+     device model to ensure policy consistency
+   - Rounds per `RequestPolicy`: `ValidValues` rounds up to the smallest valid value
+     >= base; `ValidRange` rounds up to `Min` if below, aligns to `Min + n*Step` if
+     `Step` is set. If the rounded value exceeds `Max` or all valid values, the workload
+     is marked inadmissible
```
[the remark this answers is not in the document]
tenzen-y: @sohankunkerkar I think that this AI evaluation seems right. Why can we ignore this limitation? Technically, workload can occupy the device capacities over Kueue reservation.

keps/2941-DRA/README.md
```diff
@@ -1355,6 +1398,270 @@ borrowing only operates within the same resource.
...
+| B | `memory: 20Gi` | ValidRange(min=1Mi, max=80Gi) | `20Gi` | `gpu.memory: 20Gi` |
+| C | (omitted) | Default=80Gi | `80Gi` | `gpu.memory: 80Gi` |
+| D | `memory: 90Gi` | ValidRange(min=1Mi, max=80Gi) | inadmissible | (rejected) |
+
+When `sources` is absent on a mapping, the mapping behaves exactly as today (device-count
+quota). This is backward compatible.
+
+A single mapping can have multiple capacity sources, which are summed into the mapping's
+quota resource. This is useful for composite capacity scores but not for independent
+dimension tracking (where each dimension needs its own quota resource).
```
tenzen-y: Does this mean just summarization like 4Gi memory + 30 cores = 34?
Shouldn't we do any normalization before summarization?

keps/2941-DRA/README.md
```diff
@@ -1355,6 +1398,270 @@ borrowing only operates within the same resource.
...
+track partitioned devices via `consumesCounters`, capacity sources track shared devices
+via `Capacity`. These serve different device types and are not combined on a single
+mapping entry.
+
+Admins can create a unified quota pool across both device types by using the same
+resource name on separate mappings with different DeviceClass names. For example,
+partitioned and time-sliced GPUs can both charge `gpu.memory` so that total GPU
+memory consumption is capped regardless of which sharing model workloads use.
+The same DeviceClass cannot appear in two different mappings (one counter, one
+capacity) due to the DeviceClass uniqueness constraint across mappings.
```
tenzen-y: We have a validation for the duplicated resource name in https://github.com/kubernetes-sigs/kueue/blob/2b494fec38e1afd615d706697c75340a7b76c0db/pkg/config/validation.go#L504-L508

Will we relax this validation?
If yes, can you guard the validation relaxsatioin by DRAConsumableCapacity FG, and then mention about that here? Especially, we want to emphasize which version has stricter validation and which version for relaxation. You can check how to emphasize version-oriented limitation description in TAS Kep.

keps/2941-DRA/README.md
```diff
@@ -1104,26 +1137,34 @@ type DeviceClassSourceConfig struct {
...
+// DeviceClassResourceSource identifies which resource dimension to track
+// for quota.
+type DeviceClassResourceSource struct {
+    // Name identifies the resource dimension to track for quota.
     // +required
     Name string `json:"name"`
 
     // Driver is the DRA driver name used to filter relevant ResourceSlices.
     // Must match the spec.driver field on ResourceSlice objects.
+    // The total length must not exceed 253 characters.
```
tenzen-y: Is this right? The right limitation is 63, right?
https://github.com/kubernetes/kubernetes/blob/v1.36.0/pkg/apis/resource/validation/validation.go#L648-L655

-> 

https://github.com/kubernetes/kubernetes/blob/v1.36.0/pkg/apis/core/validation/validation.go#L1810-L1834

-> 

https://github.com/kubernetes/kubernetes/blob/v1.36.0/pkg/apis/resource/validation/validation.go#L69-L73

keps/2941-DRA/README.md
tenzen-y, line 1396: nit: is this still valid description? IIUC, we introduce a resource slice controller.

keps/2941-DRA/README.md
```diff
@@ -1532,6 +1850,14 @@ tracks adding this). This follows the same pattern as upstream K8s integration t
 - CEL expression validation against ResourceSlice devices
 - event-driven requeuing of inadmissible workloads on ResourceSlice changes
 
+##### KueueDRAIntegrationConsumableCapacity (v0.19)
+
+- support for consumable capacity devices via capacity-based quota (KEP-5075, beta in k8s 1.36)
+- rounding per RequestPolicy (ValidValues and ValidRange with Step) to prevent quota gaming
+- multiple capacity sources on a single mapping (summed into one quota resource)
+- event-driven requeuing of inadmissible workloads on ResourceSlice changes
```
tenzen-y: Just checking: does this mean that we will expand resource slice controller for capacity based DRA?

In that case, can you share a rough expansion plan in this KEP? Like, when will the controller be triggered, and what will do for the cluster?

keps/2941-DRA/README.md
mimowo: Yuki, I suspect there may be some confusion about how the scheduler library would be used.

The scheduler library would be linked into and invoked by Kueue; it would not require the standalone kube-scheduler process to read the Kueue configuration API. Kueue can translate its configuration into the inputs used when running the scheduling simulation.

There are at least two possible ways the library API could support this:

1. The library could provide extension points for registering plugins implementing predefined interfaces. Kueue could then implement a plugin whose construction or runtime context includes the relevant Kueue configuration.

2. The library could expose options on its scheduling API, for example, an option controlling whether multiple allocations are allowed. Kueue would set that option based on its configuration before invoking the library.

The current scheduler-library API may not yet provide the necessary extension point or option, but it is only at "v0.1.0-alpha1". I do not see a fundamental architectural reason why the library could not support this.

keps/2941-DRA/README.md
sohankunkerkar: ok, I will go with that change.

keps/2941-DRA/README.md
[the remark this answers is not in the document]
sohankunkerkar: Yeah, good point! Let me update the KEP to add that this point specifically.

keps/2941-DRA/README.md
sohankunkerkar: As @mimowo clarified, the scheduler-library runs inside Kueue. Kueue links in the library and simulates device allocation at admission time using its own configuration. The kube-scheduler doesn't need to read Kueue config. Kueue would translate its deviceSelector  into the scheduling simulation inputs. I've updated the KEP text to make this explicit.

keps/2941-DRA/README.md
sohankunkerkar: No, it's not cross-dimension summation. Multiple capacity sources on one mapping should track the same dimension (e.g., gpu.memory) with different deviceSelector scopes,like aggregating memory from two GPU models into one pool. Mixing different dimensions (memory + cores) into one mapping would produce a meaningless sum. I've updated the KEP to make this explicit. Independent dimensions need separate mappings with distinct resource names.

keps/2941-DRA/README.md
sohankunkerkar: You're right. The resource name uniqueness validation is not relaxed in Alpha, so cross-type unified pools (counter + capacity mappings sharing the same resource name) are not supported. I've corrected the KEP to state this as an Alpha limitation and added it to the Beta graduation criteria for re-evaluation.

keps/2941-DRA/README.md
sohankunkerkar: Good catch. I used 253 treating the driver name as a DNS subdomain (RFC 1123 FQDN limit), but the DRA API explicitly constrains it to 63 via DriverNameMaxLength (same as CSI driver names). Fixed it now.

keps/2941-DRA/README.md
sohankunkerkar: Updated. The text was stale from before the ResourceSlice controller was introduced. It now describes the current behavior: the controller watches for create/update/delete events on ResourceSlices matching configured drivers and triggers requeuing of inadmissible workloads.

keps/2941-DRA/README.md
sohankunkerkar: No expansion needed. The existing ResourceSlice controller from PD already covers it. CC adds capacity source drivers to the watched driver set at startup. The controller behavior is unchanged: it's triggered on create/update/delete of ResourceSlices matching configured drivers, and it requeues inadmissible workloads so they get re-evaluated with updated device availability. Updated the KEP to reflect this informaton.

keps/2941-DRA/README.md
tenzen-y: Alright, that sounds good to me. Thank you.

keps/2941-DRA/README.md
tenzen-y: Oh, I see. Thank you.

keps/2941-DRA/README.md
tenzen-y: I'm resolving this thread as we discussed in another thread.

keps/2941-DRA/README.md
tenzen-y: @mimowo @sohankunkerkar That's exactly the same as my understanding. I am doubting if (1 or (2 approach is realistic one.

Because both approaches affect on kube-scheduler scheduling performance since (1 indicates the external system like Kueue access to kube-scheduler hot cache (2 indicates bypassing the kube-scheduler scheduling cycle. 

So, even in the long term, we probably need to rely on the waitForPodsReady mechanism if we keep expecting kube-scheduler or scheduler-library side improvements.

Hence, I'm wondering if we should consider another realistic mitigation in the next beta stage.

tenzen-y: So, my recommendation is to drop this mitigation approach and just add revisiting the mitigation process to the beta graduation criteria.

mimowo: > @mimowo @sohankunkerkar That's exactly the same as my understanding. I am doubting if (1 or (2 approach is realistic one.

The integration effort is staffed and prioritized on our timeline. We are going to work with the kube-scheduler folks to make the API usable and friendly to Kueue as possible. I cannot share detailed timeline like :"this will be in Feb 2027", but certainly the integration is planned to be continued and strengthened.

> So, even in the long term, we probably need to rely on the waitForPodsReady mechanism if we keep expecting kube-scheduler or scheduler-library side improvements.

waitForPodsReady we can mention as short-term mitigation for sure. Long-term I think we don't have an exact plan, but mentioning scheduler-library to be re-evaluated for GA sounds reasonable for me.

> Hence, I'm wondering if we should consider another realistic mitigation in the next beta stage.

Wait for pods ready is a mitigation which is already working. I don't get why we would need to add it to the beta stage.

keps/2941-DRA/README.md
tenzen-y: Thank you for updating that. I confirmed that there is no concern at all. Let's revisit the independent dimension tracking in the next graduation phase.

---
> A single mapping can have multiple capacity sources, whose charges are summed into the
mapping's quota resource. Each source should track the same dimension (e.g., `gpu.memory`)
with different `deviceSelector` scopes, for example, aggregating memory from two GPU
models into one pool. Mixing different dimensions (e.g., memory + cores) into one mapping
produces a meaningless sum and is not a supported use case. Independent dimensions require
separate mappings with distinct resource names.

keps/2941-DRA/README.md
tenzen-y: Alright, that sounds good to me.

keps/2941-DRA/README.md
tenzen-y: Thank you for fixing that.

keps/2941-DRA/README.md
tenzen-y: Yeah, I had to review the updated KEP more carefully in the previous DRA extended resource beta graduation.

Anyway, the latest KEP looks great to me.

keps/2941-DRA/README.md
tenzen-y: Thank you for updating that.
I checked "Processing Flow" section". That looks great to me.

tenzen-y:
@sohankunkerkar I think that the only point is https://github.com/kubernetes-sigs/kueue/pull/12983#discussion_r3605981620

Also, can you tell me where you addressed https://github.com/kubernetes-sigs/kueue/pull/12983#discussion_r3582235532?

keps/2941-DRA/README.md
tenzen-y: > The integration effort is staffed and prioritized on our timeline. We are going to work with the kube-scheduler folks to make the API usable and friendly to Kueue as possible. I cannot share detailed timeline like :"this will be in Feb 2027", but certainly the integration is planned to be continued and strengthened.

I didn't know that. As I discussed integration at KubeCon Amsterdam, the "extension for external controller" was lower-priority, and I assumed that the proposed approaches might not be realized.

tenzen-y: In conclusion based on @mimowo 's response, I think that we can emphasize this risk probably will not be resolved by the this FG GA graduation, which will be mitigated by waitForPodsReady for a while. We can fully eliminate this risk by scheduler-lib graduation after this FG GA grarduation.

@mimowo @sohankunkerkar Any thoughts?

mimowo: The integration is outlined (high-level) in https://docs.google.com/document/d/1Zen3R4FefiS51EjZOVwv7tuMy-G-SymzKpdwJtIBNFw

and the design of the scheduler library focuses on Kueue integration:
https://docs.google.com/document/d/1MASczp_dvKXjrbhExePxMZfasZN8ukaHopP7y_bNom4, specifically you can find the "Kueue usage example" section.

Sure, we don't have every detail yet planned, but we have plans to iterate on the integration. 

Additionally we have now this umbrella issue https://github.com/kubernetes-sigs/kueue/issues/8871 which specifically links https://github.com/kubernetes-sigs/kueue/issues/13149, which discusses DRA integration as one of the motivation points.

mimowo: > In conclusion based on @mimowo 's response, I think that we can emphasize this risk probably will not be resolved by the this FG GA graduation, which will be mitigated by waitForPodsReady for a while. We can fully eliminate this risk by scheduler-lib graduation after this FG GA grarduation.

I would rather add it to graduation points for GA to "re-evaluate". However, I don't have a strong opinion on that. We can always adjust the KEP, for now we are targeting Alpha, so planning so much ahead (probably a year is hard).

sohankunkerkar:
> Also, can you tell me where you addressed [#12983 (comment)](https://github.com/kubernetes-sigs/kueue/pull/12983#discussion_r3582235532)?

https://github.com/kubernetes-sigs/kueue/pull/12983/changes#diff-9992bde183489d9f804057ec1cebd8b81b2470c8a9430abffee5ad8db5278e93R1598
Let me know if this addresses your concern.

tenzen-y:
> > Also, can you tell me where you addressed [#12983 (comment)](https://github.com/kubernetes-sigs/kueue/pull/12983#discussion_r3582235532)?
> 
> https://github.com/kubernetes-sigs/kueue/pull/12983/changes#diff-9992bde183489d9f804057ec1cebd8b81b2470c8a9430abffee5ad8db5278e93R1598 Let me know if this addresses your concern.

Great, awesome 👍

keps/2941-DRA/README.md
sohankunkerkar: Yeah, I addressed this in the subsequent push.

tenzen-y: > The integration is outlined (high-level) in https://docs.google.com/document/d/1Zen3R4FefiS51EjZOVwv7tuMy-G-SymzKpdwJtIBNFw
> 
> and the design of the scheduler library focuses on Kueue integration:
> https://docs.google.com/document/d/1MASczp_dvKXjrbhExePxMZfasZN8ukaHopP7y_bNom4, specifically you can find the "Kueue usage example" section.
> 
> Sure, we don't have every detail yet planned, but we have plans to iterate on the integration.
> 
> Additionally we have now this umbrella issue https://github.com/kubernetes-sigs/kueue/issues/8871 which specifically links https://github.com/kubernetes-sigs/kueue/issues/13149, which discusses DRA integration as one of the motivation points.

Thank you for sharing that. I will carefully check those later. OTOH, it seems that I don't have read permission for the first doc while I can read the second doc. @mimowo can you request adding permission to non-googler for the first doc?

tenzen-y: > I would rather add it to graduation points for GA to "re-evaluate". However, I don't have a strong opinion on that. We can always adjust the KEP, for now we are targeting Alpha, so planning so much ahead (probably a year is hard).

I'm also fine with that.

tenzen-y (commented):
@sohankunkerkar This looks great, thank you 👍 

/lgtm
/approve

PR #12988, opened 2026-07-11
Use binary search for picking preemption set

olekzabl:
#### What type of PR is this?

/kind bug
/area tas

This PR refactors the `for` loops determining "how many workloads we need to preempt" (in `preemption.go`) into binary search.

This is intended as a quick mitigation for a customer who's just recently observed **700+ iterations** inside one of these loops in practice. Given the "rich" TAS fit check inside each iteration, this slowed down the whole scheduling cycle to take multiple seconds, which is not aacceptable.

Intended as a quick mitigation for the problem described above (which I think doesn't yet have a corresponding issue).

1. Binary search may not be best as a long-term solution, for a number of reasons:

   1. It may be considered overly complex.
   2. When the number of necessary preemptees is *way below* the number of all candidates, binary search can actually *slow down* the search, compared to naive iteration.
      * To remedy that, this PR actually initiates a binary search only after we've considered (unsuccessfully) the first 10 preemption candidates.
        As a result, as long as there's < 1M workloads, this PR is guaranteed to slow down the search by at most 3x.
       (While, in scenarios like the abovementioned "700 iterations", it should speed it up by >30x).
   3. [cc @brejman] In general, whether a new Pod can fit does **not** depend "monotonically" on the set of preempted Pods.
   This is caused by the Kubernetes inter-pod affinity constraints.
      * Kueue currently does not understand such constraints and effectively ignores them when making scheduling decisions.
      * However, once the proposed integration with Scheduling Library happens, things may change.
     In that case, using binary search would become non-trivial (if not incorrect) for Kueue behavior, and we may be forced to revert it.

2. The proposed strategy of acting:

   1. For the start, use binary search in order to substantially reduce the user-observed scheduler slowness.
   2. Emit logs to gather more information for fine-tuning the fix, for example:
      * Did the 700+ loop iterations go over _all_ preemption candidates, or just some?
      * How many loop iterations happen typically?
   3. Fine-tune in a follow-up.
      (Where one alternative is: drop binary search completely; instead, just quickly make a check of "would I fit if all *valid* preemption candidates were removed".
      Even though we currently have an analogous check on an "empty cluster" without TAS usage, that one seems not accurate enough to prevent long iterating).

3. This draft PR has currently *no test coverage* - as test are left (almost) intact, and the functionality is hidden behind an Alpha FG.
   However, on commit [`c767c99`](https://github.com/kubernetes-sigs/kueue/commit/c767c9904c5b61f5d85ac62a82e2c7032a6d57db) tests did pass while the FG was marked Beta (and so enabled). 
   (The only test failing then was a linter complaining about the FG being Beta :)
   So, this is rudimentarily tested.

```release-note
TBD
```

### DIFF
--- a/pkg/features/kube_features.go
+++ b/pkg/features/kube_features.go
@@ -474,6 +474,11 @@ const (
 	// TODO(#12820): temporary workaround in the RayService integration. Remove once
 	// the generic FinishOrphanedWorkloads owner-deletion check lands.
 	DeferRayServiceFinalizationForRedisCleanup featuregate.Feature = "DeferRayServiceFinalizationForRedisCleanup"
+
+	// owner: @olekzabl
+	//
+	// Use binary search for determining the preemption set when more than 10 preemptees are necessary.
+	PreemptorBinarySearch featuregate.Feature = "PreemptorBinarySearch"
 )
 
 func init() {
@@ -730,6 +735,10 @@ var defaultVersionedFeatureGates = map[featuregate.Feature]featuregate.Versioned
 
 	DeferRayServiceFinalizationForRedisCleanup: {
 		{Version: version.MustParse("0.19"), Default: true, PreRelease: featuregate.Beta},
+	},
+
+	PreemptorBinarySearch: {
+		{Version: version.MustParse("0.19"), Default: false, PreRelease: featuregate.Alpha},
 	},
 }
 
--- a/pkg/scheduler/preemption/preemption.go
+++ b/pkg/scheduler/preemption/preemption.go
@@ -21,8 +21,10 @@ import (
 	"context"
 	"fmt"
 	"slices"
+	"sort"
 	"strings"
 	"sync/atomic"
+	"time"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
@@ -55,7 +57,9 @@ import (
 	workloadevict "sigs.k8s.io/kueue/pkg/workload/evict"
 )
 
-const parallelPreemptions = 8
+const (
+	parallelPreemptions = 8
+)
 
 type Preemptor struct {
 	clock clock.Clock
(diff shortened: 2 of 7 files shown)


mimowo:
> [cc @brejman] In general, whether a new Pod can fit does not depend "monotonically" on the set of preempted Pods.

I think the problem is  rather that the  check is monotonical even wrt Kueue logic. Consider we have two nodes:
`a` (full, capacity of 2) and `z` (full, capacity of 2), and a pending workload with two PodSets, each podset if of size `1`. The first PodSet can be placed on any node, the second only on node `z`.

Now after one preemption:
- after one preemption we have: `a` - 1 free, `b` -  1 free. So we  can fit the workload
- after two preemptions we have: `2` - 2 free, `b` -  1 free. So we will place the first PodSet on "a", but the second PodSet remains unschedulable.

Using binary search against non-monotonic functions may be problematic for correctness and misleading in long term. Even if the consequences now are not large the code evolves and I cannot predict the consequences of breaking the monotonicity assumptions.

So, I would prefer to first repro the issue, grap pprof and see if this is really the best way for the fix.

mimowo:
The alternative I think worth considering would be to skip doing full TAS pass if we haven't accumulated enough of preempted workloads. Or short-circuit the  TAS checks as fast as possible. I think this would work pretty well in cases where the preemptor is a large workload say 100GPU, and we want to preempt many small ones (1GPU). In this case we can only start doing TAS when we calculate we released enough space in principle (so skip TAS after preempting the first 1..99 workloads. To facilitate the short-circuit we could tract the amount of free resources total, and fail quicky if not enough (skipping the propagation of the  numbers in the tree).

olekzabl:
@mimowo 

1. For now, I think, the **binary** question "will this wl fit?" behaves monotonically.
   And as long as that holds, the proposed binary search is not a semantic change.
   (Because it won't break searching when it finds "a fit by preempting more than needed".
   It will still find the *minimal* number of preemptees, just testing in another order).
   It will become a semantic change only following sufficiently rich WAS integration.

2. The thing you proposed here

   > To facilitate the short-circuit we could tract the amount of free resources total, and fail quicky if not enough

   is already done. [EDIT: Actually I'm not sure]

3. The next thing like this we could do (and AFAIK it's not yet done) is:
   _"After considering first N (=3?) candidates, check preempting **all** eligible candidates, and if that's not enough, break"_
   I can imagine this being the final state of this enhancement.
   However, this is going to help **only if** the 700+ iteration loop indeed goes through its whole range.
   OTOH if the loop did e.g. 720 out of 900 iterations, then such a fix will be no help.
   That's why I'd like to **start** from the binary search to:
   * quickly get a mitigation for any case
   * gather logs
   * and then decide whether to evolve into this short-circuit or some other kind of.

mimowo:
> For now, I think, the binary question "will this wl fit?" behaves monotonically.

Check the counter-example with [a,z] nodes and two PodSets. I agree it might be a corner case, but we should not mislead the reader this holds.

mimowo:
- I think this PR is a bit misleading because it suggest an assymptotic change of the algo, but the algo remains linear due to the second phase of optimizing the preemption set, so 2x improvement is max
- I'm wondering what happened to the idea we discussed internally of doing SimulatePreemptions{quotaOnly} from fitsResourceQuota

I think we need to have a repro and pprof to analyze in more detail how to approach the problem
/hold

olekzabl:
> I think this PR is a bit misleading because it suggest an assymptotic change of the algo, but the algo remains linear due to the second phase of optimizing the preemption set, so 2x improvement is max

Ohh. That's a very good catch. Apologies for my misleading.

BUT - this made me realize we have more info :) Because:

* In the user's logs, the whole "nomination phase" of the scheduler cycle took similarly as long as the discussed `for` loop of adding new preemption candidates.
* If the loop had ended with success ("we found a sufficiently large set"), then - as you noted - we should have entered `fillBackWorkloads` which should have taken similarly as long. But it did not happen!
* Conclusion: in the considered scenario, we know that even preempting all candidates did not make a TAS fit.

So, by doing item 3 from [this comment](https://github.com/kubernetes-sigs/kueue/pull/12988#issuecomment-4955979639), we should actually achieve speed-up way above 2x in their case.

PR #12989, opened 2026-07-11
Format counter-based DRA resources with BinarySI units

### LINKED ISSUE #12982: Counter-based DRA resource usage displayed as raw bytes instead of human-readable units
<!-- Please use this template while reporting a bug and provide as much info as possible. Not doing so may result in your bug not being addressed in a timely manner. Thanks!

If the matter is security related, please disclose it privately via https://kubernetes.io/security/
-->


**What happened**:
When using counter-based quota for partitionable devices (e.g., gpu.memory), the workload resourceUsage and ClusterQueue usage display raw byte values instead of human-readable BinarySI units.                 
                                                                                                                                                                                                            
$ oc get workloads -o jsonpath='{.items[0].status.admission.podSetAssignments[0].resourceUsage}'                                                                                                               {"cpu":"100m","gpu.memory":"10468982784","memory":"100Mi"}
                                                                                                                                                                                                                                                                                                                                                                        Expected: "gpu.memory":"9984Mi" (matching the BinarySI format the driver publishes and the admin uses in nominalQuota).
        
Counter-based resources should display in the same format the DRA driver publishes them (BinarySI), consistent with how nominalQuota is specified (e.g., 800Gi).  

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

amarkdotdev:
## Summary

- Register counter-based DRA logical resources for BinarySI formatting at mapper init
- Serialize byte-valued resources like `gpu.memory` as human-readable quantities in workload and ClusterQueue status
- Add unit and integration assertions for `.String()` output

## Context

#12982 reports counter-based DRA usage appears as raw byte integers (for example `10468982784`) instead of values like `9984Mi`. Values were correct internally; `ResourceQuantity()` used DecimalSI for custom resources.

## Test plan

- [x] `go test ./pkg/resources/... ./pkg/dra/... -count=1`
- [ ] `make test-integration INTEGRATION_FILTERS="--focus='DRA Partitionable Devices'"`

Fixes #12982

```release-note
DRA: Fixed a bug where byte-valued Partitionable Devices (counter-based) resources were displayed as raw byte integers in Workload and ClusterQueue status. 
These resources are formatted using human-readable BinarySI units, such as Mi and Gi.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved formatting for device resource quantities backed by byte-based counters, ensuring consistent BinarySI display (e.g., Mi/Gi).
  * GPU memory values now consistently render using binary units for both direct and counter-derived scenarios.

* **Tests**
  * Added coverage for counter-based device resources to verify expected BinarySI output (including `gpu.memory`).
  * Expanded unit and integration validations with human-readable `String()` assertions for admission, request multiplication, and cohort borrowing.

### DIFF
--- a/pkg/dra/mapper.go
+++ b/pkg/dra/mapper.go
@@ -69,12 +69,13 @@ func (m *ResourceMapper) PopulateFromConfiguration(mappings []configapi.DeviceCl
 	dcToResource := make(map[corev1.ResourceName]corev1.ResourceName)
 	dcCounters := make(map[corev1.ResourceName]*deviceClassCounterConfig)
 	for _, mapping := range mappings {
-		if len(mapping.Sources) > 0 && mapping.Sources[0].Counter != nil {
+		isCounter := len(mapping.Sources) > 0 && mapping.Sources[0].Counter != nil
+		if isCounter {
 			resources.RegisterBinaryFormattedResource(mapping.Name)
 		}
 		for _, deviceClassName := range mapping.DeviceClassNames {
 			dcToResource[deviceClassName] = mapping.Name
-			if len(mapping.Sources) > 0 && mapping.Sources[0].Counter != nil {
+			if isCounter {
 				c := mapping.Sources[0].Counter
 				dcCounters[deviceClassName] = &deviceClassCounterConfig{
 					driver:         c.Driver,
--- a/pkg/dra/mapper.go
+++ b/pkg/dra/mapper.go
@@ -21,6 +21,7 @@ import (
 	resourcev1 "k8s.io/api/resource/v1"
 
 	configapi "sigs.k8s.io/kueue/apis/config/v1beta2"
+	"sigs.k8s.io/kueue/pkg/resources"
 )
 
 // deviceClassCounterConfig holds counter configuration for a specific DeviceClass.
@@ -68,6 +69,9 @@ func (m *ResourceMapper) PopulateFromConfiguration(mappings []configapi.DeviceCl
 	dcToResource := make(map[corev1.ResourceName]corev1.ResourceName)
 	dcCounters := make(map[corev1.ResourceName]*deviceClassCounterConfig)
 	for _, mapping := range mappings {
+		if len(mapping.Sources) > 0 && mapping.Sources[0].Counter != nil {
+			resources.RegisterBinaryFormattedResource(mapping.Name)
+		}
 		for _, deviceClassName := range mapping.DeviceClassNames {
 			dcToResource[deviceClassName] = mapping.Name
 			if len(mapping.Sources) > 0 && mapping.Sources[0].Counter != nil {
--- a/pkg/resources/requests.go
+++ b/pkg/resources/requests.go
@@ -20,6 +20,7 @@ import (
 	"maps"
 	"math"
 	"strings"
+	"sync"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -28,6 +29,20 @@ import (
 
 	utilmath "sigs.k8s.io/kueue/pkg/util/math"
 )
+
+var binaryFormattedResources sync.Map
+
+// RegisterBinaryFormattedResource marks a resource name as byte-valued for display.
+// Counter-based DRA logical resources (for example gpu.memory) should be registered
+// at startup so quantities serialize with BinarySI units.
+func RegisterBinaryFormattedResource(name corev1.ResourceName) {
+	binaryFormattedResources.Store(name, struct{}{})
+}
+
+func usesBinaryFormat(name corev1.ResourceName) bool {
+	_, ok := binaryFormattedResources.Load(name)
+	return ok
+}
 
 // The following resources calculations are inspired on
 // https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/types.go
@@ -117,7 +132,7 @@ func ResourceQuantity(name corev1.ResourceName, v int64) resource.Quantity {
 	case corev1.ResourceMemory, corev1.ResourceEphemeralStorage:
 		return newCanonicalQuantity(v, resource.BinarySI)
 	default:
-		if strings.HasPrefix(string(name), corev1.ResourceHugePagesPrefix) {
+		if strings.HasPrefix(string(name), corev1.ResourceHugePagesPrefix) || usesBinaryFormat(name) {
 			return newCanonicalQuantity(v, resource.BinarySI)
 		}
 		return *resource.NewQuantity(v, resource.DecimalSI)
--- a/pkg/resources/requests_test.go
+++ b/pkg/resources/requests_test.go
@@ -19,6 +19,7 @@ package resources
 import (
 	"encoding/json"
 	"math"
+	"sync"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
@@ -317,6 +318,10 @@ func TestGreaterKeysRL(t *testing.T) {
 	}
 }
 
+func resetBinaryFormattedResources() {
+	binaryFormattedResources = sync.Map{}
+}
+
 func TestResourceQuantityRoundTrips(t *testing.T) {
 	cases := map[string]struct {
 		resource corev1.ResourceName
@@ -386,6 +391,7 @@ func TestResourceQuantityRoundTrips(t *testing.T) {
 	}
 	for name, tc := range cases {
 		t.Run(name, func(t *testing.T) {
+			t.Cleanup(resetBinaryFormattedResources)
 			if tc.resource == corev1.ResourceName("gpu.memory") {
 				RegisterBinaryFormattedResource(tc.resource)
 			}
--- a/pkg/resources/requests_test.go
+++ b/pkg/resources/requests_test.go
@@ -378,9 +378,17 @@ func TestResourceQuantityRoundTrips(t *testing.T) {
 			value:    10000000000,
 			expected: "9765625Ki",
 		},
+		"counter-based DRA resource": {
+			resource: corev1.ResourceName("gpu.memory"),
+			value:    9984 * 1024 * 1024,
+			expected: "9984Mi",
+		},
 	}
 	for name, tc := range cases {
 		t.Run(name, func(t *testing.T) {
+			if tc.resource == corev1.ResourceName("gpu.memory") {
+				RegisterBinaryFormattedResource(tc.resource)
+			}
 			quantity := ResourceQuantity(tc.resource, tc.value)
 			initial := quantity.String()
 
--- a/pkg/dra/mapper_test.go
+++ b/pkg/dra/mapper_test.go
@@ -22,6 +22,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 
 	configapi "sigs.k8s.io/kueue/apis/config/v1beta2"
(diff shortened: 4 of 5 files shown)


tenzen-y:
@amarkdotdev Thank you for opening this PR. OTOH, I'd always like to respect the first assignees https://github.com/kubernetes-sigs/kueue/issues/12982#issuecomment-4946706860

So, before moving this PR forward, let me confirm with @Amir380-A
@Amir380-A Are you still working on https://github.com/kubernetes-sigs/kueue/issues/12982? If yes, I'd like to check your PR instead of this PR.

amirialy:
> @amarkdotdev Thank you for opening this PR. OTOH, I'd always like to respect the first assignees [#12982 (comment)](https://github.com/kubernetes-sigs/kueue/issues/12982#issuecomment-4946706860)
> 
> So, before moving this PR forward, let me confirm with @Amir380-A @Amir380-A Are you still working on #12982? If yes, I'd like to check your PR instead of this PR.

Thanks for checking! I was working on a different approach that preserves the original resource.Quantity format through the scheduling pipeline. After comparing it with #12989, I think my approach is more complex than needed for this issue, and #12989 is a better fit. I'll unassign myself so it can move forward.

amarkdotdev:
@tenzen-y Thanks for flagging this -I saw @Amir380-A posted `/unassign` on #12982. I'm ready to address any review feedback here.

tenzen-y:
> > @amarkdotdev Thank you for opening this PR. OTOH, I'd always like to respect the first assignees [#12982 (comment)](https://github.com/kubernetes-sigs/kueue/issues/12982#issuecomment-4946706860)
> > So, before moving this PR forward, let me confirm with @Amir380-A @Amir380-A Are you still working on #12982? If yes, I'd like to check your PR instead of this PR.
> 
> Thanks for checking! I was working on a different approach that preserves the original resource.Quantity format through the scheduling pipeline. After comparing it with #12989, I think my approach is more complex than needed for this issue, and #12989 is a better fit. I'll unassign myself so it can move forward.

Thank you for confirming that. So, let's move this approach on.

sohankunkerkar:
@amarkdotdev Could you keep this PR to just the fix commit? You can open a separate PR for the first commit.

amarkdotdev:
@sohankunkerkar Done. Idropped the website link-check docs commit. This PR now contains only the #12982 BinarySI formatting change (pkg/resources, pkg/dra, and the DRA integration test).

pkg/dra/mapper.go
```diff
@@ -68,6 +69,9 @@ func (m *ResourceMapper) PopulateFromConfiguration(mappings []configapi.DeviceCl
 	dcToResource := make(map[corev1.ResourceName]corev1.ResourceName)
 	dcCounters := make(map[corev1.ResourceName]*deviceClassCounterConfig)
 	for _, mapping := range mappings {
+		if len(mapping.Sources) > 0 && mapping.Sources[0].Counter != nil {
```
sohankunkerkar:
```suggestion
		isCounter := len(mapping.Sources) > 0 && mapping.Sources[0].Counter != nil{
		   if isCounter {
```

pkg/dra/mapper.go
```diff
@@ -68,6 +69,9 @@ func (m *ResourceMapper) PopulateFromConfiguration(mappings []configapi.DeviceCl
...
+			resources.RegisterBinaryFormattedResource(mapping.Name)
+		}
 		for _, deviceClassName := range mapping.DeviceClassNames {
 			dcToResource[deviceClassName] = mapping.Name
 			if len(mapping.Sources) > 0 && mapping.Sources[0].Counter != nil {
```
sohankunkerkar:
```suggestion
			if isCounter {
```

pkg/resources/requests_test.go
```diff
@@ -378,9 +378,17 @@ func TestResourceQuantityRoundTrips(t *testing.T) {
...
+		"counter-based DRA resource": {
+			resource: corev1.ResourceName("gpu.memory"),
+			value:    9984 * 1024 * 1024,
+			expected: "9984Mi",
+		},
 	}
 	for name, tc := range cases {
 		t.Run(name, func(t *testing.T) {
+			if tc.resource == corev1.ResourceName("gpu.memory") {
+				RegisterBinaryFormattedResource(tc.resource)
```
sohankunkerkar, line 396: Add a test-only `resetBinaryFormattedResources()` and call it in t.Cleanu. The sync.Map leaks across test cases.

pkg/resources/requests.go
```diff
@@ -29,6 +30,20 @@ import (
 	utilmath "sigs.k8s.io/kueue/pkg/util/math"
 )
 
+var binaryFormattedResources sync.Map
```
sohankunkerkar, line 33: I see the reason behind this change but I'd like to understand if @mimowo is fine with it.

sohankunkerkar (commented):
/approve
Have a few nits
will leave lgtm to @mimowo

pkg/resources/requests.go
mimowo: I'm ok-ish with that - it does not break anything and it helps for the most commonly used resource in the context of DRA - memory.

However, I suspect it may not work best for some other resources, even managed by DRA, because DRA profiles may contain any resources. For example the MIG profiles define profiles for things like "Multiprocessors" or "Copy Engines": https://github.com/kubernetes/enhancements/blob/master/keps/sig-scheduling/4815-dra-partitionable-devices/README.md#dynamic-allocation-of-multi-instance-gpus-mig-on-nvidia-hardware

Additionally IIUC DRA is a pluggable open-world for resources.

So, we move forward I would like to at least explore the alternatives / answer questions:
1. how does the core k8s DRA determine what is the serialization unit used? Maybe we can reuse the algo?
2. maybe determine the resource based on DRA && ""mem" in the name? To exclude other resources managed by DRA? OTOH this might be too much complication.
3. API opt-in enum to say "serialization: Decimal" / "Binary"? Default could be Binary

mimowo, line 33: This global map is not the best approach. Could we instead introduce a DRA-aware resource formatter, which would inject the DRA mapper?

sohankunkerkar: >how does the core k8s DRA determine what is the serialization unit used? Maybe we can reuse the algo?

Core k8s DRA has no serialization algorithm. `resource.Quantity.Format` is set by the driver at construction time (NVIDIA hardcodes BinarySI) and there's nothing to reuse.

>maybe determine the resource based on DRA && ""mem" in the name? To exclude other resources managed by DRA? OTOH this might be too much complication.

This approach is little fragile. MIG profiles publish multiprocessors, copy-engines, decoders, `mem` heuristic would need a growing exclusion list. But more importantly, it doesn't matter for non-memory counters: values < 1024 render identically regardless of format ("28" is "28" in both BinarySI and DecimalSI). The format distinction only matters for large byte values. So the heuristic solves a problem that only exists for one resource type, while creating maintenance burden for all of them.

> API opt-in enum to say "serialization: Decimal" / "Binary"? Default could be Binary

I thought about this option but its the most expensive option. We already have most of the information i.e. the driver-published resource.Quantity.Format in ResourceSlices.

sohankunkerkar: The mapper only lives in the workload controller and the call sites that need formatting (cache/scheduler/, flavorassigner/, workload/, metrics) can't see it. A startup-populated format map that `ResourceQuantity` reads directly is the simplest way to bridge that gap without a cross-package refactor.

mimowo: > Core k8s DRA has no serialization algorithm. resource.Quantity.Format is set by the driver at construction time (NVIDIA hardcodes BinarySI) and there's nothing to reuse.

I see, in that case I'm ok to make "BinarySI"  as the default. We can revisit the API if we ever need to.

mimowo: I would like to explore a bit more how such a refactoring could look like in more detail. I'm ok to move forward as is, but plan for the follow up (and check if it would be feasible).

mimowo:
@sohankunkerkar @amarkdotdev would you like to cherrypick the PR as a bugfix? maybe we could justify the current handling is inconsistent with the industry standard by nvidia?

amarkdotdev:
Yeah, I think this qualifies as a bug fix. Right now, counter-based DRA memory is displayed as raw bytes, which doesn't match the BinarySI units NVIDIA uses for gpu.memory. I'll cherry-pick it to release-0.18 once it merges, and release-0.17 too if we want it there.

pkg/resources/requests.go
sohankunkerkar: @amarkdotdev can you check this part please?

mimowo:
Ack, let me mark the PR accordingly:
/kind bug

pkg/resources/requests.go
amarkdotdev: @sohankunkerkar I took another look, and I think your read on the global map is right.

`ResourceQuantity()` is a package-level helper used by cache/scheduler, flavorassigner, workload, webhooks, and metrics. Those call sites don't have access to the DRA mapper since it's only initialized during controller startup. Registering the counter-based logical resource names in `PopulateFromConfiguration()` seemed like the smallest change to fix the display issue without threading the mapper through every caller.

I agree it's not the ideal long-term design. A DRA-aware formatter with an injected mapper (or passing formatting metadata with the resource name) would be cleaner, but that feels like a bigger refactor affecting most `ResourceQuantity()` callers.

I'm okay with this as is and opening a follow-up issue for that refactor if that works for @mimowo.

mimowo: Ok, I will open a follow up issue. One approach would be to introduce a new stateful ResourceFormatter and inject it into the controllers which need it. This formatter would be initialized on startup to have the set of BinaryResources initialized. It does not need to know about DRA as such, the code which would marry DRA with resourceFormatter could be in main.go.

mimowo:
/lgtm
Thank you for improving the integration👍

mimowo:
/lgtm
Thank you for improving the integration👍

sohankunkerkar:
@amarkdotdev  can you address the nits?

pkg/resources/requests.go
mimowo: https://github.com/kubernetes-sigs/kueue/issues/13037

mimowo:
/unhold
please prepare the CPs

amarkdotdev:
@mimowo Addressed Sohan's nits and pushed to #12989. Backport for release-0.18 is here: https://github.com/kubernetes-sigs/kueue/pull/13038

I looked at release-0.17 too but counter-based DRA isn't on that branch yet, so there's nothing to cherry-pick there.

mimowo:
/release-note-edit
```release-note
DRA: Fixed a bug where byte-valued counter-based DRA resources were displayed
as raw byte integers in Workload and ClusterQueue status. These resources are
now formatted using human-readable BinarySI units, such as Mi and Gi.
```

tenzen-y:
/release-note-edit

```release-note
DRA: PartitionableDevices resources are displayed with binary units (e.g., `9984Mi`) instead of raw byte integers in workload and ClusterQueue status.
```

tenzen-y:
/release-note-edit
```release-note
DRA: Fixed a bug where byte-valued Partitionable Devices (counter-based) resources were displayed as raw byte integers in Workload and ClusterQueue status. 
These resources are formatted using human-readable BinarySI units, such as Mi and Gi.
```

